PR #1213, opened 2017-02-15
uhd_rx_cfile --normalized-gain flag now works.

courtarro:
I was running `uhd_rx_cfile` and tried using the `--normalized-gain` flag with a gain value of 0.75. The output file was incredibly quiet because the gain was still being set as dB. Turns out that the script doesn't use the flag, even though it exists. I fixed that.


jmcorgan:
This was cherry-picked onto the maint branch.

PR #1214, opened 2017-02-15
Modified add_custom_command to not try to execute ""

marcusmueller:
former custom command called ""; this works fine on build systems were
this is expanded by shell, but fails on some of my systems where
add_custom_command's COMMAND parameter is directly executed (because
there's no executable "").


marcusmueller:
Notice that this still feels ugly and we might want to review the whole UseSWIG.cmake code. We shouldn't have a dummy command just to let CMake know where the SWIG-generated files came from – they come from some call to swig, and that *should* just have the appropriate OUTPUT parameter.

However, the code is sufficiently convoluted for me to prefer polishing a dirty hack rather than risking breakage in a lot of unforseen places by rewriting existing commands.

PR #1218, opened 2017-02-24
uhd: Always query current frequency when outputting tags

mbr0wn:
This fixes two issues:
- When updating the frequency via messages, the new tags now reflect the
  actual frequency
- When different channels have different center frequencies, every
  channel reports its own frequency accurately


mbr0wn:
Technically a bugfix, but also a behavioural change. I thus feel more comfortable putting this onto master.

PR #1221, opened 2017-03-09
gr-fec: switch possible C++11 code from 'constexpr' to just 'const'.

michaelld:
Subject says it all. Simple fix to not require C++11 tweaks in GNU Radio when building with Thrift (which does require C++11). Discussed http://lists.gnu.org/archive/html/discuss-gnuradio/2017-03/msg00054.html .

### DIFF
--- a/gr-fec/include/gnuradio/fec/polar_decoder_common.h
+++ b/gr-fec/include/gnuradio/fec/polar_decoder_common.h
@@ -28,10 +28,6 @@
 #include <gnuradio/fec/generic_decoder.h>
 #include <gnuradio/fec/polar_common.h>
 
-#ifndef BOOST_CONSTEXPR_OR_CONST
-#define BOOST_CONSTEXPR_OR_CONST const
-#endif
-
 namespace gr {
   namespace fec {
     namespace code {
@@ -68,7 +64,7 @@ namespace gr {
         bool set_frame_size(unsigned int frame_size){return false;};
 
       private:
-        static BOOST_CONSTEXPR_OR_CONST float D_LLR_FACTOR;
+        static const float D_LLR_FACTOR;
         unsigned int d_frozen_bit_counter;
 
       protected:


jmcorgan:
@gnieboer @guruofquality Can you weigh in on this?

guruofquality:
Sure. I track the latest boost and this ended up being a problem for me as of my last recent built attempt. I came up with the [same fix](https://github.com/pothosware/PothosSDR/blob/master/patches/gr_fec_polar_llr_factor.diff) as @michaelld  Thats the only patch I needed BTW to build maint, which is really cool.

I regret not patching this in a portable way on my previous attempt, I was going for minimum code change and I didnt want to basically undo [this commit](
https://github.com/gnuradio/gnuradio/commit/a1cf1193), and I didnt realize how that boost macro would play out. BOOST_CONSTEXPR_OR_CONST is strange because you cant really portably switch between const and constexpr, at least not in a static member variable.

So for the rundown on these sort of things:

* This generally causes linker error on various compilers, not just msvc:
```
//in the header file
class foo {
static const float SOME_FLOAT = 1234.0f;
```

* This would be fine BTW, integers and some other types are exempt from this rule apparently:
```
//in the header file
class foo {
static const int SOME_INT = 1234;
```

* This is the **portable** way to handle this, we use it throughout most of the gr codebase:
```
//in the header file
class foo {
static const float SOME_FLOAT;

//in the cpp file
const float foo::SOME_FLOAT = 1234.0f;
```

* And this is perfectly fine if you are willing switch to c++11:
```
//in the header file
class foo {
static constexpr SOME_FLOAT = 1234.0f;
```

And finally, D_LLR_FACTOR is a private member variable thats used as a constant exactly inside of one C++ file. We probably should have never put it in a header file in the first place :-)

jmcorgan:
@michaelld Can you rebase on maint for this?

michaelld:
@jmcorgan sure. will do shortly.

michaelld:
Have to close down this PR & start another; GitHub doesn't allow me to rebase to a different branch than that of the PR (which is master). Sigh. I'll tag that PR here once it's up.

guruofquality:
> Have to close down this PR & start another; GitHub doesn't allow me to rebase to a different branch than that of the PR (which is master). Sigh. I'll tag that PR here once it's up.

You can force push the branch. No?

michaelld:
@guruofquality you can force push updates or changes, but the PR will still be off of whatever it was when created. One of the major defects in how GitHub does things, IMHO.

michaelld:
https://github.com/gnuradio/gnuradio/pull/1233

guruofquality:
@michaelld If you click the edit button near the title of the pull request, you get a edit for the title and a drop down to select the branch.

michaelld:
@guruofquality Had never seen that option; thx! Too late for this PR, but I'll remember!

gnieboer:
Yes, I commented in the initial discussions on the mailing list, I think
the change is a good idea and doesn't break what the original changes were
for in the first place.  Concur with Josh's comments just now.

Geof

On Fri, Mar 10, 2017 at 11:44 AM, Johnathan Corgan <notifications@github.com

Issue #1222, opened 2017-03-09
Update SWIG dependency from 1.3.3 to at least 2.0.7


marcusmueller:
SWIG is a bugger, and we should proactively make sure we reduce the number of possible versions that users use. 

<hr>

###  Reasoning 

Revisiting https://bugzilla.redhat.com/show_bug.cgi?id=1143914 . 

Gist of it:

runtime/thread/thread.h uses `int` where it should use `size_t`. We didn't adjust that (back then) because **we weren't sure it'd break older SWIGs**, though it would've been the _right thing to do_<sup>TM</sup>. It breaks build on a few "exotic" archs (s390), but not having that patch around just because it _might_ break something on really old systems makes little to no sense, imho. Patch is really minimal, attached (straight from the Fedora package).

[gnuradio-3.7.10-size_t.txt](https://github.com/gnuradio/gnuradio/files/831620/gnuradio-3.7.10-size_t.txt)

If we upstream that, GR builds unpatched on all architectures supported by RedHat, CentOS, Fedora.

If we bump the dependency, I'll make a PR for said patch against master and next.

marcusmueller:
For perspective: SWIG versions offered by old distros; in **bold** the pre-2.0 SWIGs.


| Distro | EOL | Version of SWIG |
|------------------------|--------------|-------------------------|
| CentOS 5/RHEL 5 |2013-01-08| **1.3.29-2.el5** |
| Debian 6 (squeeze) | 2014-05-31 (full) / 2016-02-29 (LTS) | **1.3.40-3** |
| Ubuntu 10.04 LTS| 2013-05-09 (Desktop) / 2015-04-30 (Server) | **1.3.40-2ubunut1**|
| CentOS 6/RHEL 6 | 2016-05-10| **1.3.40-6.el6** |
| **EOL LINE**| <hr> | <hr>|
| Ubuntu 12.04 LTS| 2017-04-XX | 2.0.4+really2.0.4-4ubuntu2 (gotta love Canonical's versioning)|
| **EOL @ GR 3.8 Release LINE** | <hr> | <hr>|
| Debian 7 (wheezy/*oldstable*) | 2016-04-26 (full) / 2018-05-XX (LTS)) | 2.0.7 |
| Ubuntu 14.04 LTS | 2019-04 | 2.0.11-1ubuntu2|
| CentOS 7/RHEL 7 | 2019-Q4 | 2.0.10-5.el7 |
[...]

PR #1215, opened 2017-02-15
Update dependency versions on next

mbr0wn:
This PR will probably linger a bit until we've figured out the actual versions we want.

PR #1223, opened 2017-03-10
Bumped SWIG minimal SWIG version to 2.0.4

marcusmueller:
This is the pull for next, see #1222


mbr0wn:
There is some overlap with #1215.

marcusmueller:
@mbr0wn indeed, and #1215 does the right thing by removing old versions and making the minimal SWIG version a CMake variable and so on.

We should just also do it for the 3.7.X line, as I presume it won't disappear overly soon, and as community, we won't be rid of the "support burden" the next two years; bumping dependencies on master (and also, maint) will make our lives easier. Frankly, we've not been doing a terrible good job at testing our fixes against the versions we "promise" things should work on – I've been guilty of that several times myself, but if you want a recent example, #1227. I'd rather actually spell out what we **need** on maint than not touching the dependencies. In fact, with our dependency list, GR should build cleanly on RH 6.8, according to the RPMs I've been trying to build yesterday. It does not.

Problem is that testing this kind of thing is pretty intense, and to make matters worse, you basically have to work yourself through about 5 different versions of the Boost docs if you use a single boost function to make sure it works on 1.35 – 1.63; SWIG is worse, by far, in that most of the things where it behaves erratically are not covered by documentation, and oh, SWIG documentation.

I think the approach of taking each of the dependencies individually, looking specifically at it, filling in a table of what we perceive are distros we need to support over the next years might be an easier thing (and easier to test) than to bump all "core" dependencies at once.

jmcorgan:
@marcusmueller I'm pretty sure that the merge up from master in #1224 covers this on next, making this PR redundant.  Can you verify?

mbr0wn:
@jmcorgan Yep you're right. I'll abuse my powers and close this.

PR #1225, opened 2017-03-10
Fix types: pthread_XX(size_t cpuset), not int

marcusmueller:
Patch by Jaroslav, https://bugzilla.redhat.com/show_bug.cgi?id=1143914 , straight from the Fedora package.
Minor modification only.

Since the reasoning back then was "it's totally OK if Fedora has this
patch, but we don't know whether ancient SWIG on other distros works
with that", and we're now bumping the SWIG version (#1222), it's time ot
upstream this patch.


jmcorgan:
I'd rather see this on master than maint, as it does change an API visible data type.  Normally, this would even be cause for going on to next, but in this case the change is unlikely to cause any compatibility problems.

marcusmueller:
can do that, no problem, but it *is* a bug fix – though it's a benign bug, and the change is actually API-changing.

jmcorgan:
This is failing QA tests.  The below is on Ubuntu 14.04, with SWIG 2.0.11: 

```
ERROR: test_000 (__main__.test_affinity)
  File "/opt/gnuradio/src/gr-blocks/python/blocks/qa_affinity.py", line 40, in test_000
    src.set_processor_affinity([0,])
  File "/opt/gnuradio/build/gr-blocks/python/blocks/../../swig/blocks_swig1.py", line 6655, in set_processor_affinity
    return _blocks_swig1.vector_source_f_sptr_set_processor_affinity(self, *args, **kwargs)
TypeError: in method 'vector_source_f_sptr_set_processor_affinity', argument 2 of type 'std::vector< std::size_t,std::allocator< std::size_t > > const &'
'''
```

jmcorgan:
Obsoleted by #1232

PR #1226, opened 2017-03-10
Adding the uniform_int distribution include

marcusmueller:
Seems like this was forgotten when adding that distro to random.cc; breaks
build on non-recent Boosts. (i.e., especially on CentOS 6)


marcusmueller:
argh, typo & fixed wrong problem.

PR #1232, opened 2017-03-10
Changed the affinity/CPU masks from int to size_t (master)

marcusmueller:
Master-based version of #1225: 

Follows discussion on
https://bugzilla.redhat.com/show_bug.cgi?id=1143914 .

Since the reasoning back then was "it's totally OK if Fedora has this
patch, but we don't know whether ancient SWIG on other distros works
with that", and we're now bumping the SWIG version (#1222), it's time ot
upstream this patch.

### DIFF
--- a/docs/doxygen/other/thread_affinity.dox
+++ b/docs/doxygen/other/thread_affinity.dox
@@ -50,7 +50,7 @@ Each block has two new data members:
 A block can set and unset its affinity at any time using the
 following member functions:
 
-- gr::block::set_processor_affinity(const std::vector<int> &mask)
+- gr::block::set_processor_affinity(const std::vector<std::size_t> &mask)
 - gr::block::unset_processor_affinity()
 
 Where \p mask is a vector of core numbers to set the thread's affinity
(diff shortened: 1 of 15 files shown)


jmcorgan:
So, this has the same QA test failures described in #1225:

The below is on Ubuntu 14.04, with SWIG 2.0.11: 

```
ERROR: test_000 (__main__.test_affinity)
  File "/opt/gnuradio/src/gr-blocks/python/blocks/qa_affinity.py", line 40, in test_000
    src.set_processor_affinity([0,])
  File "/opt/gnuradio/build/gr-blocks/python/blocks/../../swig/blocks_swig1.py", line 6655, in set_processor_affinity
    return _blocks_swig1.vector_source_f_sptr_set_processor_affinity(self, *args, **kwargs)
TypeError: in method 'vector_source_f_sptr_set_processor_affinity', argument 2 of type 'std::vector< std::size_t,std::allocator< std::size_t > > const &'
'''
```
I'm going to close #1225 as we'll definitely be doing this on master/next only.

marcusmueller:
Closing this PR because I really can't figure out what I was thinking at the time. And neither can @bastibl or @jmcorgan.

PR #1233, opened 2017-03-10
gr-fec: switch possible C++11 code from 'constexpr' to just 'const'.

michaelld:
should be on maint now.


michaelld:
Prior discussion: https://github.com/gnuradio/gnuradio/pull/1221

PR #1235, opened 2017-03-11
uhd: Removed stray ifdefs

mbr0wn:


jmcorgan:
I think this was from me borking up a merge conflict, thanks.

PR #1246, opened 2017-03-21
blocks: Add exponentiate const int block

mhostetter:
The Complex to Complex^M block takes the M-th power of the input complex stream. The exponent parameter must be at least 1. There is also a callback function so the exponent can be changed at runtime.

### DIFF
--- a/gr-blocks/grc/blocks_exponentiate_const_cci.xml
+++ b/gr-blocks/grc/blocks_exponentiate_const_cci.xml
@@ -1,0 +1,45 @@
+<?xml version="1.0"?>
+<block>
+  <name>Exponentiate Const Int</name>
+  <key>blocks_exponentiate_const_cci</key>
+  <import>from gnuradio import blocks</import>
+  <make>blocks.exponentiate_const_cci($exponent, $vlen)</make>
+  <callback>set_exponent($exponent)</callback>
+
+  <param>
+    <name>Num Ports</name>
+    <key>num_ports</key>
+    <value>1</value>
+    <type>int</type>
+  </param>
+  <param>
+    <name>Vec Length</name>
+    <key>vlen</key>
+    <value>1</value>
+    <type>int</type>
+  </param>
+  <param>
+    <name>Exponent</name>
+    <key>exponent</key>
+    <value>1</value>
+    <type>int</type>
+  </param>
+  
+  <check>$num_ports &gt; 0</check>
+  <check>$vlen &gt; 0</check>
+  <check>$exponent &gt; 0</check>
+  
+  <sink>
+    <name>in</name>
+    <type>complex</type>
+    <vlen>$vlen</vlen>
+    <nports>$num_ports</nports>
+  </sink>
+
+  <source>
+    <name>out</name>
+    <type>complex</type>
+    <vlen>$vlen</vlen>
+    <nports>$num_ports</nports>
+  </source>
+</block>
--- a/gr-blocks/include/gnuradio/blocks/exponentiate_const_cci.h
+++ b/gr-blocks/include/gnuradio/blocks/exponentiate_const_cci.h
@@ -1,0 +1,66 @@
+/* -*- c++ -*- */
+/*
+ * Copyright 2017 Free Software Foundation, Inc.
+ *
+ * This file is part of GNU Radio
+ *
+ * GNU Radio is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 3, or (at your option)
+ * any later version.
+ *
+ * GNU Radio is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with GNU Radio; see the file COPYING.  If not, write to
+ * the Free Software Foundation, Inc., 51 Franklin Street,
+ * Boston, MA 02110-1301, USA.
+ */
+
+
+#ifndef INCLUDED_BLOCKS_EXPONENTIATE_CONST_CCI_H
+#define INCLUDED_BLOCKS_EXPONENTIATE_CONST_CCI_H
+
+#include <gnuradio/blocks/api.h>
+#include <gnuradio/sync_block.h>
+
+namespace gr {
+  namespace blocks {
+
+    /*!
+     * \brief Exponentiates a complex stream with an integer exponent
+     * \ingroup blocks
+     *
+     * \details
+     * This block raises a complex stream to an integer exponent. The exponent
+     * must be at least 1. There is a callback function so the exponent can be
+     * changed at runtime.
+     *
+     * NOTE: The algorithm uses iterative multiplication to achieve exponentiation, 
+     * hence it is O(exponent). Therefore, this block could be inefficient for large
+     * exponents.
+     */
+    class BLOCKS_API exponentiate_const_cci : virtual public gr::sync_block
+    {
+     public:
+      typedef boost::shared_ptr<exponentiate_const_cci> sptr;
+
+      /*
+       * \param exponent Exponent the input stream is raised to, which must be an integer.
+       * The algorithm uses iterative multiplication to achieve exponentiation, hence it is
+       * O(exponent). Therefore, this block could be inefficient for large exponents.
+       * \param vlen Vector length of input/output stream
+       */
+      static sptr make(int exponent, size_t vlen = 1);
+
+      virtual void set_exponent(int exponent) = 0;
+    };
+
+  } // namespace blocks
+} // namespace gr
+
+#endif /* INCLUDED_BLOCKS_EXPONENTIATE_CONST_CCI_H */
+
--- a/gr-blocks/lib/exponentiate_const_cci_impl.cc
+++ b/gr-blocks/lib/exponentiate_const_cci_impl.cc
@@ -1,0 +1,102 @@
+/* -*- c++ -*- */
+/*
+ * Copyright 2017 Free Software Foundation, Inc.
+ *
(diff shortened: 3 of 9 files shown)


ThomasHabets:
Is there value to having this block be able to exponentiate to a complex number, or if there's need should that be a separate block?

mhostetter:
I guess I don't follow you completely. What do you mean by exponentiate?

jmcorgan:
Also, I suggest this block have a variable number of input/output stream pairs (but still using a common exponent parameter.)  There are a few examples like this in gr-blocks that you can draw from.

As far as complex exponents, the algorithm used in this block appears to exponentiation through repeated multiplies, which requires the exponent to be an integer.  I'm not sure if this is the best way from a numerical precision perspective, but the optimization would required factoring the exponent into primes and doing a tree of multiplies.

jmcorgan:
This block isn't a type converter, it is a math operator.

mhostetter:
I'll start working on those changes.

ThomasHabets:
@mhostetter http://mathworld.wolfram.com/ComplexExponentiation.html

I don't know enough about DSP to know if non-integer exponents would ever be useful, but the code here only supports non-complex integers.

Which could be fine. Depending on what you expect the exponent to be set to. If this is only expected to be used to detect BPSK vs QPSK, then the values will be small integers and it probably doesn't matter.

But probably a future user would be surprised if this involved O(n) multiplies if they ever used it for anything else. And if they'd need it they probably expect it to support complex exponents too.

jmcorgan:
I agree in that I'd expect most people to use this for small integer-N cases, so the complex exponentiation case should go in another block (if anyone every cares to do that.)

Also, the O(N) complexity should be documented, however.

mhostetter:
@ThomasHabets I was mainly thinking about PSK detection where the exponent would be an integer, but I see your point.

@jmcorgan regarding the variable number of input/outputs, do you want a parameter like "Num Ports" to configure N inputs going to N outputs? I know there are other blocks like Add/Subtract that take "Num Inputs", but still produce a fixed set of outputs. In this case each input/output pair would be totally independent, right? Is there value in combining them in one block? There would be no performance benefit, right? I'm willing to do it if you think it's a good idea. Just talking. :smile: 

Also, I will move the block to Math Operators. Should the name of this block be something like "Exponentiate Const", similar to the "Add Const" and "Multiply Const" blocks?

And finally, @jmcorgan would you like me to force push my changes so they are all captured in 1 commit, or is the history useful?

Thanks!

jmcorgan:
What I'm suggesting is to allow multiple streams to be handled in the same block, all using the same exponent. That way if someone has, say, five streams they want to cube, they can use one block instead of five blocks to do it.

So, the io_signature would have 1, -1, and sizeof(gr_complex) for input and output, and you'd add a check_topology() that ensures the user connected number of input ports equals the user connected number of output ports.  See the implementation in gr::blocks::copy for an example of port handling for this ability (ignore the itemsize stuff, that's for handling multiple types as well, but you wouldn't be doing that in your block.)

Bonus points for implementing a vlen parameter so that the block can work on items of vlen size at a time, like the output of a FFT operation, without having to resort to using vector_to_stream or the reverse as a gasket block. Many of the math already blocks do this.

I wish we had this pattern (optional multiple streams at once and vlen) for many of the simple operations in gr-blocks, just for consistency, so when accepting new ones, it makes sense to start off correctly.

"Exponentiate Const Int" and gr::blocks::exponentiate_const_cci would make sense for the naming, but it's a mouthful.

Finally, I usually prefer pull requests as single commits based of a recent branch tip, since this makes it easier to ensure that every commit in the tree is a compile-able for automated tools like git bisect.

awalls-cx18:
I can't help but think there is a breakpoint for some value of M, where mag(z)^M * exp(j*M*arg(z)) would be a more efficient way to compute this.

awalls-cx18:
BTW, a multiply block with M inputs, all fed the same stream, covers the small integer M case for one stream.

mhostetter:
@awalls-cx18 that's true about the multiply block with M inputs, but I wanted this block to have a configurable exponent at runtime. That you wouldn't get with a hardwired multiply block.

awalls-cx18:
Ah, good point.

mhostetter:
I've implemented and tested the features requested. In all candor, this is my first time using the vector length feature in GNU Radio. I believe I implemented it correctly, but I'd appreciate feedback/review. Thanks.

jmcorgan:
Thanks for being so responsive to suggestions.  Now we can do the nitpicky line-by-line review :)

mhostetter:
@jmcorgan just curious, did you receive a notification every time I pushed test code to my branch (maybe 10 times in the last 2 hrs)? I'm hoping the answer is no...

jmcorgan:
No, only when you made your comment.

mhostetter:
Phew :relieved:

gr-blocks/grc/blocks_exponentiate_const_cci.xml
```diff
@@ -0,0 +1,49 @@
...
+  </sink>
+
+  <source>
+    <name>out</name>
+    <type>complex</type>
+    <vlen>$vlen</vlen>
+    <nports>$num_ports</nports>
+  </source>
+
+  <doc>
```
mbr0wn: Shouldn't this be auto-generated from the Doxygen?

gr-blocks/include/gnuradio/blocks/exponentiate_const_cci.h
```diff
@@ -0,0 +1,72 @@
...
+     * the algorithm uses iterative multiplies to achieve the exponentiation. Therefore,
+     * the algorithm is O(exponent), which may be inefficient for large exponents.
+     */
+    class BLOCKS_API exponentiate_const_cci : virtual public gr::sync_block
+    {
+     public:
+      typedef boost::shared_ptr<exponentiate_const_cci> sptr;
+
+      /*!
+       * \brief Return a shared_ptr to a new instance of blocks::exponentiate_const_cci.
```
mbr0wn: It's my preference to document the `\param`s here, and then I generally leave out the `\brief`, because it would always be the same. But that's just preference!

mhostetter:
I incorporated your suggestions @mbr0wn. Thanks for the feedback!

gr-blocks/include/gnuradio/blocks/exponentiate_const_cci.h
```diff
@@ -0,0 +1,64 @@
...
+    class BLOCKS_API exponentiate_const_cci : virtual public gr::sync_block
+    {
+     public:
+      typedef boost::shared_ptr<exponentiate_const_cci> sptr;
+
+      /*
+       * \param vlen Vector length of input/output stream
+       * \param exponent Exponent input stream is raised to. Must be an integer.
+       */
+      static sptr make(size_t vlen, int exponent);
```
jmcorgan: The vlen argument should come last, and have a default value set to 1.  This allows someone to just specify the exponent and get the default expected behavior.  Note, this change must propagate down through the impl class, but the default argument should only occur here in the API class.

gr-blocks/python/blocks/qa_exponentiate_const_cci.py
```diff
@@ -0,0 +1,58 @@
...
+                v2s = blocks.vector_to_stream(gr.sizeof_gr_complex, vlen)
+                sink = blocks.vector_sink_c(1)
+
+                self.tb.connect(source, s2v)
+                self.tb.connect(s2v, exponentiate_const_cci)
+                self.tb.connect(exponentiate_const_cci, v2s)
+                self.tb.connect(v2s, sink)
+                self.tb.run()
+
+                self.assertEqual(sink.data(), out_data)
```
jmcorgan: This should use assertTuplesAlmostEqual.  While in this QA with integer coefficients and small n, the results will be precise, in general, QA code for floating point values should have a tolerance for numerical imprecision.

gr-blocks/python/blocks/qa_exponentiate_const_cci.py
```diff
@@ -0,0 +1,58 @@
...
+            out_data = (in_data[0]**exponent, in_data[1]**exponent, in_data[2]**exponent, in_data[3]**exponent)
+
+            for vlen in [1, 2, 4]:
+                source = blocks.vector_source_c(in_data, False, 1)
+                s2v = blocks.stream_to_vector(gr.sizeof_gr_complex, vlen)
+                exponentiate_const_cci = blocks.exponentiate_const_cci(vlen, exponent)
+                v2s = blocks.vector_to_stream(gr.sizeof_gr_complex, vlen)
+                sink = blocks.vector_sink_c(1)
+
+                self.tb.connect(source, s2v)
```
mbr0wn: You don't need to change this, but in Python, you can just string up blocks in the connect call.

mhostetter:
Thanks all for the review. I've incorporated and tested the changes you requested. Let me know if there are other changes you'd like.

jmcorgan:
Almost ready! Check you block in GRC and let me know what you find.

mhostetter:
Well that's embarrassing. I forgot to change the most important `make()` call. Sorry. I've tested this via `make test` and a GRC flowgraph now. Hopefully all is fixed. :smile:

gr-blocks/lib/exponentiate_const_cci_impl.cc
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    int
+    exponentiate_const_cci_impl::work(int noutput_items,
+        gr_vector_const_void_star &input_items,
+        gr_vector_void_star &output_items)
+    {
+      gr::thread::scoped_lock guard(d_setlock);
+
+      for(int i = 0; i < input_items.size(); i++) {
```
jmcorgan: I missed this in my earlier review.  The call to size() returns an unsigned int, but the for loop index is signed, causing a compiler warning. I know we're inconsistent about this elsewhere, but since you're a shiny new contributor, best not to copy our bad habits.

mhostetter: Good catch. I changed it to an `unsigned int`.

PR #1252, opened 2017-03-26
cmake/docs: Move min versions into a single spot, put them all into vars

mbr0wn:
- Autogen docs appropriately
- Make it easier to maintain min versions

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -25,16 +25,18 @@ endif(${CMAKE_SOURCE_DIR} STREQUAL ${CMAKE_BINARY_DIR})
 ########################################################################
 # Project setup
 ########################################################################
+# Make sure this version matches ${GR_CMAKE_MIN_VERSION} (a variable can't be
+# used here).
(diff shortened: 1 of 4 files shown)


mbr0wn:
Same as before, but on master. I think I got the versions right.

mbr0wn:
@jmcorgan This PR is for master, so 3.7, not 3.8. It'll merge-conflict into next, and #1215 will fix that.

CMakeLists.txt
```diff
@@ -283,15 +343,12 @@ include(GrBoost)
 ########################################################################
 # Enable python component
 ########################################################################
-find_package(PythonLibs 2)
+find_package(PythonLibs ${GR_PYTHON_MIN_VERSION))
```
jmcorgan: This has a ) where a } should be, but I'll fix it in the merge.

jmcorgan:
Ah, I can't change anything in the merge without redoing #1215.  Can you fix and re-push to that and this one?

mbr0wn:
OK, I fixed both PRs.

PR #1254, opened 2017-03-28
blocks: Add stream to PDU block

mhostetter:
The Stream to PDU block takes an item stream and slices a fixed- or variable-length packet starting with a tag of a particular key. This block is very similar to the Tagged Stream to PDU block, however it doesn't require a full tagged stream (i.e. consecutive `packet_len` tags), instead just tags when signals are present.

I made this block because I find myself often working with "bursty" data. I only want to create PDUs for my bursts/packets and not all other times. I tried to make this block general enough to support any use case.

This block has two modes:

* **Fixed-Length Mode**: The Stream to PDU block will slice a fixed number of items starting at the tag

image

* **Variable-Length Mode**: The Stream to PDU block will slice starting at the tag with length equal to the tag's value (similar to the Tagged Stream to PDU block)

image

The block will also create a dictionary out of additional tags and store it in the PDU metadata. Additionally the absolute offset of the first packet item will be stored in the dictionary with key "Offset Key".

There is an example flowgraph included. I welcome all your feedback.


jmcorgan:
I'll eventually review this myself, but please contact other project members to ask for their own review on it here.  My first impression is that there could be some confusion with the tagged stream version of this, but I haven't looked in detail yet.

mhostetter:
@mbr0wn, @n-west, @marcusmueller, @bastibl I would appreciate any review/feedback on this new block.

Here were my thoughts when making the block:

* I often find myself wanting to slice an item stream into a PDU, starting at some tagged sample and for some fixed length. One use case is a burst (of known length/duration) whose starting sample is tagged by a detector. The burst would be packaged into a PDU for further entire-burst style DSP. Another use case is a byte stream where a `correlate_access_code_bb` block tagged the beginning of a packet.
* I know Johnathan discussed potentially deprecating Tagged Stream Blocks in 3.8 and removing them in 3.9, which I think is a good idea. I find them difficult to use. And I don't see how they fit "bursty" style comms, as you have to tag the no-signal parts of the stream to make the TSBs happy. That seems non-ideal as the downstream blocks' `work()` functions are called with noise, not just signal. I'm hoping this block will fill that void. It will slice the stream only when a signal is present, as specified by a tag.

There are two modes I made for this block.

* **Fixed-Length Mode:** The block will slice a fixed number of samples, starting with the sample tagged with a specific Tag Key.
* **Variable-Length Mode:** The block will slice starting at the tagged sample with length equal to the tag's value. This mimics the TSB implementation where the key (e.g. `packet_len`) triggers the slicer and the value indicates the slice length.

One open thought I have:

* I originally made this block with one provision: packets cannot overlap. For example. if a tag at offset 100 has length 50, a subsequent tag at offset 110 with any length will be ignored. This was to avoid handling many PDUs simultaneously (for complexity and memory reasons). In hindsight, I think I could easily implement it removing that provision, allowing any type of overlap. I can envision a few scenarios where someone might want to do this. What are your thoughts?

Thanks in advance.

osh:
Hi Matt,

This was essential the reason gr-eventstream (https://github.com/osh/gr-eventstream) was written which serves the same purpose of bridging

(information about when events occur in a stream + a stream )   to produce -->  ( async pdus )

There are a number of modes you will likely want to consider (es.sink tries to abstract this away such that different "trigger" behaviors can result in different patterns of extraction within the data stream) -- such as periodic in wall time, periodic in sample time, based on a rising edge/threshold, etc -- which all have nicely abstracted trigger blocks to try to avoid lumping all the logic into some monolithic monstrosity.

Another thing to consider is, will your processed 'events' based on PDUs ever result in wanting to extract something else back from the original stream itself. (creating a circular dependency).  For instance, header information might inform you of how long a packet is, what modulation is used in a payload field, etc --- this is the reason eventstream has 'events' instead of just PDUs.   The difference here is two fold --- the lifetime of an event is tracked, such that you can satisfy any next-event extraction dependency back into the stream (i.e. holding up consumption such that the event is still in the accessible history range), and secondly, allowing 'event handlers' which don't incur the overhead of the PDU system and scheduler -- which have been measured to run faster ---- for highly involved stateful/bursty PHYs, with a lot of these event -> stream feedback paths, applications I have seen do a lot better off using these more primitive C based event handlers.    However for most simple applications its easiest to just use eventstream to produce PDUs, and this is what I do for virtually every burst system now without these kinds of feedback paths, i.e. all extraction from the stream is performed periodically or based on some kind of triggering criterion (detection statistic) etc.

As you mention below, stream tags are then translated into the pdu meta tag, etc.

You might see if this meets your needs before trying to reinvent it --- all of these operations can be pretty tedious to implement.   I would love to have this cleaned up and merged some day ---  We do already have about 10 different data processing and duplicated blocks in GNU Radio now, I'm not sure if re-introducing this model under another name will help to clean up the ground that has already been covered, or duplicate effort and confuse people with yet-another-block. 

Cheers,
Tim

mhostetter:
Hey Tim. I am familiar with your work on Eventstream, but candidly I haven't used it. I agree that there is duplication of approaches in GNU Radio currently. I remember a year or two ago there was discussion about settling on a path forward -- tagged stream blocks, message passing, or Eventstream. @jmcorgan has the dev team decided on a path forward?

I'm all in favor of a unified approach. I can see how this block could confuse things if Eventstream or a derivative is eventually merged in-tree. I'm totally fine with whatever approach the core devs want to take. And whatever the path forward is, I'm on board to help integrate it. I think a unified, in-tree, burst-processing mechanism is much needed...and I think everyone agrees TSBs are not the best solution.

jmcorgan:
I think @mhostetter is comparing apples and oranges. There are a variety of ways to solve problems using GNU Radio, and each of them has trade-offs that allows them to be used in different situations. Different parts of an application naturally segment into unbounded streaming flows, flows with embedded boundaries at arbitrary points, and discrete, well-bounded bundles of data (PDUs.) Since applications are so unique, there is no one true way to address them.

The specific issue here is the @osh has been pretty far down the road in solving the particular problem you are addressing here with your PR, and has raised good points about the various corner cases that come up and difficulties he's solved along the way, for the problem domain he's working in.  Eventstream is available as an OOT module that requires no changes to GNU Radio itself.

mhostetter:
I guess what I meant was if a user wanted to grab a burst of data from a stream and format it into a PDU at non-periodic intervals (something that's common in my use of GNU Radio when receiving bursty packets), there's no way of doing that currently in tree, as far as I'm aware. That was the intention of this block. I realize now that Eventstream can do this, albeit out of tree.

PR #1256, opened 2017-03-30
Fix cmake issues in python3 branch

djanderson:
Fixes a few outstanding issues with the python3 cmake process.

@sdh11, please verify 3256314 fixes your issue with the prefix
@skoslowski, I touched a file in the GRC dir to fix a cmake error (something about expecting a string instead of bytes) and tested it with both py2 and py3, but please review (5815a58)
@jmcorgan, c274b4e should take care of all the outstanding GR_*_MIN_VERSION strings, as requested

### DIFF
--- a/grc/blocks/variable_struct.xml.py
+++ b/grc/blocks/variable_struct.xml.py
@@ -93,5 +93,5 @@ if __name__ == '__main__':
 
     data = make_xml(MAX_NUM_FIELDS)
 
-    with open(filename, 'w') as fp:
+    with open(filename, 'wb') as fp:
         fp.write(data.encode())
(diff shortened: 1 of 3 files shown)


grc/blocks/variable_struct.xml.py
```diff
@@ -93,5 +93,5 @@ def make_xml(num_fields):
 
     data = make_xml(MAX_NUM_FIELDS)
 
-    with open(filename, 'w') as fp:
+    with open(filename, 'wb') as fp:
```
skoslowski: That's fine. In general, we should probably request utf-8 encoding explicitly. makes no difference here, though

sdh11:
I believe it does fix that issue. I will test it on a couple of other systems as well.

jmcorgan:
@djanderson Let me know when you think this is good to go.

djanderson:
@jmcorgan, should be good to go.

PR #1259, opened 2017-04-03
Flow graphs for BBC reference streams and Germany DVB-T2 profiles.

drmpeg:


jmcorgan:
This can be merged as-is, however, the examples/CMakeLists.txt does not install these nor many other example files into the $PREFIX.

drmpeg:
Yes, I forgot to do that. Additional commit shortly.

PR #1262, opened 2017-04-05
grc: Add window icon for main window

pwicks86:
The main grc icon gets set to the default question mark (at least on linux). GNU Radio has a nice icon, so it seems like it would be nice if it used it. I've used the icon png from the official gnu radio logo repository on github here. 

This is my first PR for gnuradio, so if I've missed anything in the process, please let me know. Thanks.


drmpeg:
Just FYI, there's a way to install an icon already. See:

https://wiki.gnuradio.org/index.php/GNURadioCompanion#Installing_Icons.2C_Mime_Type.2C_and_Menu_Items

pwicks86:
Dang, missed that somehow. 

Would it make sense to integrate ``grc_setup_freedesktop`` into the standard ``make install``?

skoslowski:
I would, I suppose. This has been around longer than I have. @jmcorgan , any reason for this script not to run at install time?

jmcorgan:
Well, it's specific to a particular desktop manager, which the user might not have in their OS.  Typically, these things are run in a postinst script by a package manager on systems that support it.

jmcorgan:
If we want to have this automated at the application level, we could do it the way it is implemented here, but the the [FHS](http://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch04s11.html) location for the icon file would be under $PREFIX/share/icons/gnuradio.

jmcorgan:
Please let me know if there will be further movement on this, or if it should be closed out.

skoslowski:
I'll formally review this soon, but it looks good

PR #1266, opened 2017-04-07
docs: Python has dynamic but strong typing....

skoslowski:


jmcorgan:
This looks like it could be on maint instead.

PR #1272, opened 2017-04-10
qa: fix qa_fast_atan2f

noc0lour:
Try and fix the using std::isnan issue in ubuntu14.04


noc0lour:
Now the ubuntu14.04 fail at some other test, but at least it builds the `qa_fast_atan2f`. I changed the case of the macro to uppercase as well.

jmcorgan:
I'll merge this as is, but we should probably consider moving stuff like this into an arch/distro/compiler specific place, for reuse and clarity.

PR #1275, opened 2017-04-11
Fix a segfault issue for QT time sink

rubund:
When running with trigger mode set to "tag", there is
a corner case where d_start becomes -1, and the flow
graph crashes.

Somebody else should please run some regression tests.

On my side, this does at least eliminate an annoying segfault - which occurs quite often when there are MANY tags..

### DIFF
--- a/gr-qtgui/lib/time_sink_c_impl.cc
+++ b/gr-qtgui/lib/time_sink_c_impl.cc
@@ -563,12 +563,15 @@ namespace gr {
                         nr, nr + nitems + 1,
                         d_trigger_tag_key);
       if(tags.size() > 0) {
-        d_triggered = true;
         trigger_index = tags[0].offset - nr;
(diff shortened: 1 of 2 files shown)


jmcorgan:
Thanks for submitting this. While we still need to solve the underlying corner case issue, this does at least permit operation to continue when it happens.

However, a bug fix, we like to merge these into our 'maint' branch and then up to master and next. Could you please  rebase this branch off maint, and force push it to your repo, then edit this PR to change the requested base to maint?

rubund:
> Thanks for submitting this. While we still need to solve the underlying corner case issue, this does at least permit operation to continue when it happens.

It does not look like the underlying corner case is a problem to worry too much about.

> However, a bug fix, we like to merge these into our 'maint' branch and then up to master and next. Could you please  rebase this branch off maint, and force push it to your repo, then edit this PR to change the requested base to maint?

Done


Cheers

PR #1277, opened 2017-04-14
runtime: add barrier to sync block startup

noc0lour:
Juts want to see how the QA tests turn out now. Definitively not ready to merge

### DIFF
--- a/gnuradio-runtime/include/gnuradio/thread/thread.h
+++ b/gnuradio-runtime/include/gnuradio/thread/thread.h
@@ -27,6 +27,8 @@
 #include <boost/thread/mutex.hpp>
 #include <boost/thread/locks.hpp>
 #include <boost/thread/condition_variable.hpp>
+#include <boost/thread/barrier.hpp>
+#include <boost/shared_ptr.hpp>
 #include <vector>
 
 #if defined(_WIN32) || defined(__WIN32__) || defined(WIN32)
@@ -46,6 +48,8 @@ namespace gr {
     typedef boost::mutex                     mutex;
     typedef boost::unique_lock<boost::mutex> scoped_lock;
     typedef boost::condition_variable        condition_variable;
+    typedef boost::barrier                   barrier;
+    typedef boost::shared_ptr<barrier>       barrier_sptr;
 
     /*! \brief a system-dependent typedef for the underlying thread type.
      */
--- a/gnuradio-runtime/lib/scheduler_tpb.cc
+++ b/gnuradio-runtime/lib/scheduler_tpb.cc
@@ -26,6 +26,7 @@
 #include "scheduler_tpb.h"
 #include "tpb_thread_body.h"
 #include <gnuradio/thread/thread_body_wrapper.h>
+#include <boost/make_shared.hpp>
 #include <sstream>
 
 namespace gr {
@@ -34,14 +35,15 @@ namespace gr {
   {
     block_sptr d_block;
     int d_max_noutput_items;
+    thread::barrier_sptr d_start_sync;
 
   public:
-    tpb_container(block_sptr block, int max_noutput_items)
-      : d_block(block), d_max_noutput_items(max_noutput_items) {}
+    tpb_container(block_sptr block, int max_noutput_items, thread::barrier_sptr start_sync)
+      : d_block(block), d_max_noutput_items(max_noutput_items), d_start_sync(start_sync) {}
 
     void operator()()
     {
-      tpb_thread_body body(d_block, d_max_noutput_items);
+      tpb_thread_body body(d_block, d_start_sync, d_max_noutput_items);
     }
   };
 
@@ -71,6 +73,9 @@ namespace gr {
       blocks[i]->detail()->set_done(false);
     }
 
+    thread::barrier_sptr start_sync =
+      boost::make_shared<thread::barrier>(blocks.size()+1);
+
     // Fire off a thead for each block
 
     for(size_t i = 0; i < blocks.size(); i++) {
@@ -84,12 +89,11 @@ namespace gr {
       else {
         block_max_noutput_items = max_noutput_items;
       }
-
-      d_threads.create_thread(
-	    gr::thread::thread_body_wrapper<tpb_container>
-            (tpb_container(blocks[i], block_max_noutput_items),
-             name.str()));
+      d_threads.create_thread(thread::thread_body_wrapper<tpb_container>(
+          tpb_container(blocks[i], block_max_noutput_items, start_sync),
+          name.str()));
     }
+    start_sync->wait();
   }
 
   scheduler_tpb::~scheduler_tpb()
--- a/gnuradio-runtime/lib/tpb_thread_body.cc
+++ b/gnuradio-runtime/lib/tpb_thread_body.cc
@@ -32,7 +32,7 @@
 
 namespace gr {
 
-  tpb_thread_body::tpb_thread_body(block_sptr block, int max_noutput_items)
+  tpb_thread_body::tpb_thread_body(block_sptr block, gr::thread::barrier_sptr start_sync, int max_noutput_items)
     : d_exec(block, max_noutput_items)
   {
     //std::cerr << "tpb_thread_body: " << block << std::endl;
@@ -92,6 +92,7 @@ namespace gr {
     // make sure our block isnt finished
     block->clear_finished();
 
+    start_sync->wait();
     while(1) {
       tpb_loop_top:
       boost::this_thread::interruption_point();
--- a/gnuradio-runtime/lib/tpb_thread_body.h
+++ b/gnuradio-runtime/lib/tpb_thread_body.h
@@ -24,6 +24,7 @@
 #include <gnuradio/api.h>
 #include <gnuradio/block.h>
 #include <gnuradio/block_detail.h>
+#include <gnuradio/thread/thread.h>
 #include "block_executor.h"
 
 namespace gr {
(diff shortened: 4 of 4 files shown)


jmcorgan:
So I know you are in still in debug phase, but eventually before merge we'll have to figure out a way to get those std::cout calls into the logging system.

noc0lour:
sure, the prints are just for looking at the test output

mbr0wn:
See also: https://github.com/gnuradio/gnuradio/pull/316

mbr0wn:
The two PRs are literally the same, and were developed independently. Just a fun fact :)

jmcorgan:
But now with added German!

bastibl:
I wonder what would be the difference to this approach
https://github.com/gnuradio/gnuradio/pull/1278/commits/fa7c8ec6ab9ccd235fbf0b4e658d99441adcb7ca

jmcorgan:
Also, meta-lol on the fact the the QA for this PR is failing because it doesn't have this PR merged in.

jmcorgan:
Wait, never mind.

noc0lour:
It's only passing one, the QA on 14.04 fails for next because our `make* is too old :D

mbr0wn:
@bastibl You're not synchronizing anything in your approach. The barriers kick off the blocks in sync.

noc0lour:
By default nothing happens on a start call (it returns true). I wonder how many blocks actually implemented that method. So the work calls are still not sync'ed.

bastibl:
@mbr0wn not sure what you are syncing here. I thought it is about start()-ing all blocks before any block calls work.

jmcorgan:
The start(), stop() calls on blocks were intended for hardware that needed kicking and unkicking for streaming operation. They can be used for anything, but they have a different lifetime than constructor/destructor, in that the latter only execute once each, while a flowgraph can be started/stopped/started... indefinitely.

jmcorgan:
Not many blocks override start, though.

bastibl:
Some interleaved discussions, but @noc0lour , @mbr0wn, I still don't understand why it was implemented like that. Why can't you just start() the blocks before creating the threads (see the linked commit).

> @bastibl You're not synchronizing anything in your approach.

Seems like I miss something, but I thought start() is a blocking call. So if the scheduler iterates over all blocks and start()s them before it iterates over them again to create the threads, then, I thought, it would be synchronized. Are you trying to synchronize something else or am I misinterpreting the code?

noc0lour:
@bastibl `start()` is doing nothing in `99%` of blocks. `start()` is meant for hw blocks who initialize/deinitialize hardware. Anyway we have to synchronize threads. Once they are kicked off your only controls are synchronizing points like barriers, latches and mutexes. Currently the threads are started in serial and go directly into the `thread_body` loop and that results in a staggered start. If we sync the block threads before they go into the execution loop we have a guarantee that all the things before than are finished by all blocks. Which can mitigate some nasty race conditions.

bastibl:
I know and I agree. (Even though I think that start() might be a bit more relevant than you think. For example, blocks that need another thread use start() to run it (see [message strobe](https://github.com/gnuradio/gnuradio/blob/3a2c67c67639e0a43ec90e69199c0e4b0cf1b05b/gr-blocks/lib/message_strobe_impl.cc#L68) for example)). So it's not just about USRPs/HW.

As I said:
> I thought it is about start()-ing all blocks before any block calls work.

This is what you are trying to do, istn't it? Or what do you mean with

> we have a guarantee that all the things before than are finished by all blocks.

If I understood you correctly, I just wonder why one cannot just iterate over the blocks to start() them before creating the threads. I linked to [this commit](https://github.com/gnuradio/gnuradio/pull/1278/commits/fa7c8ec6ab9ccd235fbf0b4e658d99441adcb7ca), but @mbr0wn  told me that I'm "not synchronizing anything ".

Looks like I'm pretty stupid and miss an obvious point... Is it not possible to just start() the blocks before starting any thread?

jmcorgan:
By initializing and starting the thread, then calling start/stop from within the block executor, start/stop execute in the same thread context as the work function and msg handlers. Also, start is not just called once at the beginning and stop at the end.

_From the perspective of a block_, it is:

- constructor() called in master thread once when block is created
   - start() called in new block specific thread
      - handle_msg() called in block specific thread (if messages waiting)
      - work() called in block specific thread (if stream processing possible)
   - stop() called in block specific thread, then thread exits
- destructor() called in master thread once when block is deleted

The innermost handle_msg/work repeats indefinitely while the flowgraph is running. The outer start/stop may also happen repeatedly if the top_block has its start/stop called in the master thread, or when there is dynamic flowgraph reconfiguration. (There have been several bugs in blocks related to not understanding the latter.)

It is important to have all the start/msg/work/stop calls into the block happen in the same thread context, so that they can freely share data members without synchronization, as well as any thread-specific data that might be needed for what the block does.

Currently (by deliberate intent and design, not a bug), there are no interactions between block threads, so the wall-clock order of execution of the above sequence in each thread is indeterminate. The only point of synchronization is after stop is called, the master thread joins each of the block threads as they exit (this happens in wait), so when wait returns in the master thread, everything is guaranteed to be complete.

The issue we're facing is that in QA code, this indeterminacy results in race conditions, especially under load, when test results require state to be passed all the way through the flowgraph before shutdown. For pure streaming tests, this is easy, as the start/stop/wait sequence in the master thread guarantees this.  However, when tests use message passing, there are race conditions where senders/receivers are not ready at the same time.  A similar situation happens when using ZMQ, as this doesn't guarantee delivery.

The solution to the above may or may not involve using barriers to synchronize global flowgraph state changes, like start()/stop().  It could be a bug in the message passing system, or missing logic for handling message delivery during shutdown after the stream processing is done.

TL;DR - multi-threaded programming is complex and hard, a lot of thought went into the current design, we have flaky QA tests that use message passing, and the fix could involve this PR, something else, or both.

bastibl:
> Also, start is not just called once at the beginning and stop at the end.

Yes, but for every top_block start()/stop or lock()/unlock() cycle a new scheduler is created. This scheduler could (like in the commit) just start() all blocks before branching off the worker threads.

Since, at this stage, the top_block is still single threaded, there are no problems with shared data members, etc.

The only difference would be, as you mentioned, that the blocks start() method would be called in the context of the main thread. For the blocks work() and message handlers it's clear, but I don't see why start() has to be called in the context of the blocks worker thread.

jmcorgan:
>The only difference would be, as you mentioned, that the blocks start() method would be called in the context of the main thread. For the blocks work() and message handlers it's clear, but I don't see why start() has to be called in the context of the blocks worker thread.

One area this has come up in the past (though I'd have to research the details) is that some hardware and some software libraries require initialization from the same thread that will later call into it. ISTR that some DMA controllers have thread-specific initialization for memory buffers, but again, I'd have to back and find the context (no pun intended) where this was discussed.

jmcorgan:
>I'd have to go back and find the context (no pun intended) where this was discussed.

All this was hashed out back when Eric wrote the thread-per-block scheduler (8 years ago, 9?). We uncovered a bunch of issues because prior to that, there was a single thread that executed all block functions (construct/start/work/stop/destruct), on all blocks, and in a topology sorted way.

mbr0wn:
I took another look at @bastibl 's commit. It would probably achieve the same. However, I'm not sure serializing `start()` calls is a good idea. There's not reason they can't run in parallel; and for large flow graphs, that seems desirable. So if you don't call `start()` serially, ahead of time, you're back in concurrency-land, where you need to sync these things. @noc0lour and I took the approach that blocks call `start()`, then wait for the other `start()`s to finish, then go into `work()`.

bastibl:
So after I had my private lessons in how to start() a block, I don't want to question your approach any longer. It looks OK. I tested it. It didn't blow up my PC.
I assume the patch does what it is supposed to, but all I can tell is that there was not obvious regression.

gnuradio-runtime/include/gnuradio/thread/thread.h
```diff
@@ -46,6 +48,9 @@ namespace gr {
     typedef boost::mutex                     mutex;
     typedef boost::unique_lock<boost::mutex> scoped_lock;
     typedef boost::condition_variable        condition_variable;
+    typedef boost::barrier                   barrier;
+    typedef boost::shared_ptr<barrier>       barrier_sptr;
+
```
noc0lour: unnecessary newline here

gnuradio-runtime/lib/scheduler_tpb.cc
```diff
@@ -84,12 +89,12 @@ namespace gr {
       else {
         block_max_noutput_items = max_noutput_items;
       }
-
       d_threads.create_thread(
 	    gr::thread::thread_body_wrapper<tpb_container>
-            (tpb_container(blocks[i], block_max_noutput_items),
+      (tpb_container(blocks[i], block_max_noutput_items, start_sync),
```
noc0lour: alignment borked

jmcorgan:
Thanks @bastibl for your review and comments, at this point I think this is mergeable with @noc0lour changes above.

jmcorgan:
Let me know when you have these cleaned up, also, your thoughts on maint/master/next for merge.

noc0lour:
probably it would be beneficial to change the base to `master` and then merge it up to next. Don't know if this should go into `maint` though. Since it is a "bugfix" even maint would be legit.

jmcorgan:
I'm most wondering about the surprise factor; that is, does it introduce any unexpected behavior changes. I don't see how, and I am leaning toward maint, but could be convinced otherwise.

jmcorgan:
@noc0lour If you could cleanup those minor issues above and rebase on maint, I'll get this in.

PR #1278, opened 2017-04-14
runtime: merge scheduler and scheduler_tpb

bastibl:
Since there is only one scheduler left, this merges `scheduler` and `scheduler_tpb`.
It also changes `make_scheduler()` to `scheduler::make()` and `scheduler_sptr` to `scheduler::sptr`.


bastibl:
not really a pull request, but an RFC.

PR #1283, opened 2017-04-17
gr-uhd: dont assign back to args

noc0lour:
don't assign values back to args since that breaks apps who use the same
args to instantiate multiple usrp objects (e.g. tx and rx) with the same
arguments. 

The args object will be untouched after the instantiation.


mbr0wn:
:+1:

PR #1285, opened 2017-04-17
cmake: fix off-by-one on version check for swig

jmcorgan:


CMakeLists.txt
```diff
@@ -347,7 +347,8 @@ find_package(SWIG)
 
 if(SWIG_FOUND)
     set(SWIG_VERSION_CHECK FALSE)
-    if("${SWIG_VERSION}" VERSION_GREATER ${GR_SWIG_MIN_VERSION})
+    if("${SWIG_VERSION}" VERSION_EQUAL ${GR_SWIG_MIN_VERSION} OR
+       "${SWIG_VERSION}" VERSION_GREATER ${GR_SWIG_MIN_VERSION})
```
mbr0wn: I think `GR_SWIG_MIN_VERSION` was adjusted to make the old code work. This is cleaner, but you'll also need to update the min version then.

jmcorgan: Let me look, at least on next it was set to 2.0.11, which was causing the fail.  I didn't check on master.

jmcorgan: Yeah, master is at 2.0.3, I'll change it to 2.0.4.  Its more clear as documentation to make this the actual version instead of off by one.  I myself would have forgotten then after only a few days...

PR #1292, opened 2017-04-21
digital: use GR_LOG for debug/error in correlate_access_code_* blocks

nowls:
Replace "if(VERBOSE)" output with GR_LOG_DEBUG statements. Report errors using GR_LOG_ERROR before throwing in constructor.


noc0lour:
I guess we found another flaky QA test \o/

PR #1295, opened 2017-04-23
blocks: Add concatenate block

rubund:
This block takes the input from several sources and put all the items
after each other. All the items from the source connected to the
first input are produced first, then all the items from the second
source, and so on. The upstream blocks need to return -1 in the
general_work method to make this block aware of when they have
completed producing items.

It can be useful for appending a vector (vector_source) to the end of a file
(file_source), appending one file to another file etc.

This is an example of a flow graph:
flow_graph
The number of inputs has been set to 4. The file /tmp/mydata.cfile is prepended with 1000 zeros from the vector_source. Then 1000 zeros are added between the file /tmp/mydata.cfile and /tmp/mydata2.cfile.

settings

The block allows any data type and vector length. The type is common to all inputs and the output.

### DIFF
--- a/gr-blocks/grc/blocks_block_tree.xml
+++ b/gr-blocks/grc/blocks_block_tree.xml
@@ -170,6 +170,7 @@
    </cat>
    <cat>
       <name>Stream Operators</name>
+      <block>blocks_concatenate</block>
       <block>blocks_deinterleave</block>
       <block>blocks_interleave</block>
       <block>blocks_keep_m_in_n</block>
--- a/gr-blocks/grc/blocks_concatenate.xml
+++ b/gr-blocks/grc/blocks_concatenate.xml
@@ -1,0 +1,65 @@
+<?xml version="1.0"?>
+<block>
+  <name>Concatenate</name>
+  <key>blocks_concatenate</key>
+  <import>from gnuradio import blocks</import>
+  <make>blocks.concatenate($type.size*$vlen, $ninputs)</make>
+
+  <param>
+        <name>N input streams</name>
+        <key>ninputs</key>
+        <value>2</value>
+        <type>int</type>
+  </param>
+  <param>
+       <name>Type</name>
+       <key>type</key>
+       <type>enum</type>
+       <option>
+           <name>Complex</name>
+           <key>complex</key>
+           <opt>size:gr.sizeof_gr_complex</opt>
+       </option>
+       <option>
+           <name>Float</name>
+           <key>float</key>
+           <opt>size:gr.sizeof_float</opt>
+       </option>
+       <option>
+           <name>Int</name>
+           <key>int</key>
+           <opt>size:gr.sizeof_int</opt>
+       </option>
+       <option>
+           <name>Short</name>
+           <key>short</key>
+           <opt>size:gr.sizeof_short</opt>
+       </option>
+       <option>
+           <name>Byte</name>
+           <key>byte</key>
+           <opt>size:gr.sizeof_char</opt>
+       </option>
+  </param>
+  <param>
+        <name>Vec Length</name>
+        <key>vlen</key>
+        <value>1</value>
+        <type>int</type>
+  </param>
+  <check>$vlen &gt; 0</check>
+
+
+  <sink>
+    <name>in</name>
+    <type>$type</type>
+    <vlen>$vlen</vlen>
+    <nports>$ninputs</nports>
+  </sink>
+
+  <source>
+    <name>out</name>
+    <type>$type</type>
+    <vlen>$vlen</vlen>
+  </source>
+</block>
--- a/gr-blocks/grc/blocks_concatenate.xml
+++ b/gr-blocks/grc/blocks_concatenate.xml
@@ -3,7 +3,9 @@
   <name>Concatenate</name>
   <key>blocks_concatenate</key>
   <import>from gnuradio import blocks</import>
-  <make>blocks.concatenate($type.size*$vlen, $ninputs)</make>
+  <make>blocks.concatenate($type.size*$vlen, $ninputs)
+self.$(id).set_tag_parts($tag_parts)</make>
+  <callback>self.$(id).set_tag_parts($tag_parts)</callback>
 
   <param>
         <name>N input streams</name>
@@ -47,6 +49,22 @@
         <value>1</value>
         <type>int</type>
   </param>
+
+  <param>
+      <name>Tag parts</name>
+      <key>tag_parts</key>
+      <value>False</value>
+      <type>bool</type>
+      <option>
+          <name>Yes</name>
+          <key>True</key>
+      </option>
+      <option>
+          <name>No</name>
+          <key>False</key>
+      </option>
+  </param>
+
   <check>$vlen &gt; 0</check>
 
 
--- a/gr-blocks/include/gnuradio/blocks/CMakeLists.txt
+++ b/gr-blocks/include/gnuradio/blocks/CMakeLists.txt
@@ -95,6 +95,7 @@ install(FILES
     complex_to_mag_squared.h
     complex_to_arg.h
     conjugate_cc.h
+    concatenate.h
     copy.h
     deinterleave.h
     delay.h
--- a/gr-blocks/include/gnuradio/blocks/concatenate.h
+++ b/gr-blocks/include/gnuradio/blocks/concatenate.h
@@ -1,0 +1,61 @@
+/* -*- c++ -*- */
+/* 
+ * Copyright 2017 Free Software Foundation, Inc.
+ * 
+ * This file is part of GNU Radio
+ * 
+ * GNU Radio is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 3, or (at your option)
+ * any later version.
+ * 
+ * GNU Radio is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ * 
+ * You should have received a copy of the GNU General Public License
+ * along with GNU Radio; see the file COPYING.  If not, write to
+ * the Free Software Foundation, Inc., 51 Franklin Street,
+ * Boston, MA 02110-1301, USA.
+ */
+
+
+#ifndef INCLUDED_BLOCKS_CONCATENATE_H
+#define INCLUDED_BLOCKS_CONCATENATE_H
+
+#include <gnuradio/blocks/api.h>
+#include <gnuradio/block.h>
+
+namespace gr {
+  namespace blocks {
+
+    /*!
+     * \brief Concatenate full streams
+     * \ingroup stream_operators_blk
+     *
+     * \details
+     * This block takes the input from several sources and put all the items
+     * after each other. All the items from the source connected to the
+     * first input are produced first, then all the items from the second
+     * source, and so on. The upstream blocks need to return -1 in the
+     * general_work method to make this block aware of when they have
+     * completed producing items.
+     */
+    class BLOCKS_API concatenate : virtual public gr::block
+    {
+     public:
+      typedef boost::shared_ptr<concatenate> sptr;
+
+      /*!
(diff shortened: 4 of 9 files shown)


jmcorgan:
I have some concerns about the methodology use here.  The pointer chasing through d_detail is intended to be an internal mechanism used by the scheduler, and is only public instead of private to make the class inheritance is-a/has-a hierarchy simpler.  We may make changes to how this works and don't want to be constrained for compatibility reasons by having this baked in in this block.

I could see a mechanism where the runtime adds an EOF stream tag when a block returns DONE from its work function, then having this block pay attention to those.

bastibl:
I didn't have a detailed look yet, but I think without further support from runtime that's the way to implement it. The question might be, whether this is the only case where a block has to know if one of its input is finished (in that case, it might be OK to have a bit a hackish solution)  or if there are other use-cases. At least in the latter case, we could think of extending block_detail with a helper function, which could, for example, return -1 if the input is still active and the number of items left when it's done.

The idea with adding an EOF tag sounds good, but I guess blocks can also mark their output buffers as done without producing. When downstream blocks already consumed all samples, it might be too late to add a stream tag and signal the end of the stream.

jmcorgan:
The whole point of having the d_detail has-a relationship is to make those opaque, and ideally, this member variable would have been private and not even accessible in derived classes.  The work function of a block should communicate through its parent class functions.

I agree the EOF tag idea is problematic for the reason you give. Another idea would be to add an is_done(int port) => bool method to the block interface that provides this information while maintaining clean separation of block user code and runtime internal implementation.

rubund:
> The question might be, whether this is the only case where a block has to know if one of its input is finished (in that case, it might be OK to have a bit a hackish solution)  or if there are other use-cases. At least in the latter case, we could think of extending block_detail with a helper function, which could, for example, return -1 if the input is still active and the number of items left when it's done.

I have a repeat_stream block here on my side which I am on my way to prepare a
pull request for. It repeats the entire input stream (configurable maximum
number of items) N times, where N can be infinite. Together, the concatenate
block followed by the repeat_stream block have been proven to be a very
powerful combination for me. I guess you can already imagine all the convenient
possibilities it opens up for, when having them available in
gnuradio-companion.

So I guess the problem with the approach is that if lots of out-of-tree developers
start copying the approach to detect the end, it will become messy. As long as
everything is in the core GNU radio code (gr-block), I guess it would be maintainable?

> The idea with adding an EOF tag sounds good, but I guess blocks can also mark their output buffers as done without producing. When downstream blocks already consumed all samples, it might be too late to add a stream tag and signal the end of the stream.

I agree on that it would be nice to have a more generalized and obvious way to
do it. It is important that the EOF tag comes on the "last sample", and
not on an additional sample, if it gets added. It would be great if the
runtime could take care of this.


Is it possible to let the concatenate block in now as it is, and if the EOF tag
approach is later added to the runtime, we could modify the concatenate block
at that time?

n-west:
The other big question: should this functionality be included in GNU Radio at all? It's not immediately clear how this is useful other than this one use case which requires jumping through hoops to do one-off file manipulations which is hardly GNU Radio's raison d'etre (why not cat file1.32fc >> file2.32fc, which is a far easier method).

I don't mean to be a downer here, but not all blocks created belong in GR (otherwise we'd never finish compiling). I seem to remember a "what block's get merged in to GR" philosophy/paragraph somewhere that might be useful to point to if I could find it-- or maybe I'm making this up and it never existed.

rubund:
> I agree the EOF tag idea is problematic for the reason you give. Another idea would be to add an is_done(int port) => bool method to the block interface that provides this information while maintaining clean separation of block user code and runtime internal implementation. 

I started on something similar here the other day:
  https://github.com/rubund/gnuradio/commit/6ee90a111da74e69582364931c9960cc7432a81a

, but that is maybe the wrong approach, although it works. It changes the ABI for gr::block, so I found it easier
to just access the detail from inside the block.

rubund:
> The other big question: should this functionality be included in GNU Radio at all? It's not immediately clear how this is useful other than this one use case which requires jumping through hoops to do one-off file manipulations which is hardly GNU Radio's raison d'etre (why not cat file1.32fc >> file2.32fc, which is a far easier method).

So GNU Radio is used for simulation, testing and education a lot. Do you really
not think there should be a straightforward way to append zeros to the end of a
file - without having to open up the shell (and generate a new file which
wastes disk space)? With the concatenate block, every DSP student can do it all
from gnuradio-companion. Pick out one file_source, one vector_source and one
concatenate block. Connect them together, and there it is. Same goes for two files.
Let us say the two files are 1 GB each. Should the user be forced to got to the
shell and create a new file of 2 GB just to get the content of the files loaded
into a flow graph in GNU radio.

> I don't mean to be a downer here, but not all blocks created belong in GR (otherwise we'd never finish compiling). I seem to remember a "what block's get merged in to GR" philosophy/paragraph somewhere that might be useful to point to if I could find it-- or maybe I'm making this up and it never existed.

This is a technology independent block, which is very general in nature (in my
opinion). Blocks which are specific to a specific radio standard etc, I suppose
would make more sense to not let into GR.


Ruben

bastibl:
> Another idea would be to add an is_done(int port) => bool method to the block interface that provides this information while maintaining clean separation of block user code and runtime internal implementation.

So `block` would be the better place for the function, but I think the function signature should be something like items_left(int port) => int. With -1 for "not done" and a positive number for the items_left. Otherwise we need 2 functions that would have to be called in a particular order, i.e., 
```
items = items_available(i);
d = is_done(i);
```
as used in the PR would not work since we don't hold a mutex for the input and there might have been items added between the calls.

sdh11:
> So GNU Radio is used for simulation, testing and education a lot. Do you really
> not think there should be a straightforward way to append zeros to the end of a
> file - without having to open up the shell (and generate a new file which
> wastes disk space)? With the concatenate block, every DSP student can do it all
> from gnuradio-companion. Pick out one file_source, one vector_source and one
> concatenate block. Connect them together, and there it is. Same goes for two files.
> Let us say the two files are 1 GB each. Should the user be forced to got to the
> shell and create a new file of 2 GB just to get the content of the files loaded
> into a flow graph in GNU radio.

A simpler solution may be to extend the File Source to support the zero padding and multiple file inputs. If it is set to repeat, then it repeats all the files in order rather than a single file continuously.

rubund:
> 
> > So GNU Radio is used for simulation, testing and education a lot. Do you really
> > not think there should be a straightforward way to append zeros to the end of a
> > file - without having to open up the shell (and generate a new file which
> > wastes disk space)? With the concatenate block, every DSP student can do it all
> > from gnuradio-companion. Pick out one file_source, one vector_source and one
> > concatenate block. Connect them together, and there it is. Same goes for two files.
> > Let us say the two files are 1 GB each. Should the user be forced to got to the
> > shell and create a new file of 2 GB just to get the content of the files loaded
> > into a flow graph in GNU radio.
> 
> A simpler solution may be to extend the File Source to support the zero padding and multiple file inputs. If it is set to repeat, then it repeats all the files in order rather than a single file continuously. 

I agree it is a simple solution. However, it requires more work, it blows up the file_source block complexity and it is less general. (in my view)

And also, what about all the other application specific source blocks I have in various out-of-tree modules? With the concatenation block, I can add a head block, and a file source, and I can combine to get a stream with my specific stimuli generator data followed by the content of a file.

Another example of an application: Let us say you want to make (for some reason) a flow graph which for 1 second transmits GFSK modulation, then switches to AM modulation for 0.5 second, then back again. With a few head blocks, the concatenation block, the soon-to-be repeat_stream block, and the two different signal generator blocks, you can get this from inside gnuradio-companion. It will loop infinitely. You can then go on and add a gaussian noise source to the full stream, put in some QT GUI sliders to alter parameters live, set up triggering on the "repeat" stream tag from the repeat_stream block, and you will see live how the parameters controlled by the sliders affect the simulation.

I can come up with a number of other examples where a concatenation block (and a repeat_stream block) may be useful.

sdh11:
> I agree it is a simple solution. However, it requires more work, it blows up the file_source block complexity and it is less general. (in my view)

True, it would make that block more complex, but I don’t know about less general. The default would still just be a single file and no padding. 

> And also, what about all the other application specific source blocks I have in various out-of-tree modules? With the concatenation block, I can add a head block, and a file source, and I can combine to get a stream with my specific stimuli generator data followed by the content of a file.
> 
> Another example of an application: Let us say you want to make (for some reason) a flow graph which for 1 second transmits GFSK modulation, then switches to AM modulation for 0.5 second, then back again. With a few head blocks, the concatenation block, the soon-to-be repeat_stream block, and the two different signal generator blocks, you can get this from inside gnuradio-companion. It will loop infinitely. You can then go on and add a gaussian noise source to the full stream, put in some QT GUI sliders to alter parameters live, set up triggering on the "repeat" stream tag from the repeat_stream block, and you will see live how the parameters controlled by the sliders affect the simulation.

Definitely a useful application, but I don’t see the advantage of concatenate for that use case. You can already build this flow graph with core blocks like stream_mux and head. You possibly could add the tag_strobe block to periodically inject tags.

I do see where concatenate is more useful over stream_mux for large file captures since you don’t have to calculate length manually.  But, by relying on the upstream block to return DONE, I think that is a less flexible solution. You wouldn’t be able to use the file source in repeat mode.

Another implementation is to look for a tag on each input specifying the number of samples to pull from each input before switching to the next one. File_source output could output that tag along with your file_begin tag. With this solution, you could make any number of repeating file/vector/signal combinations without requiring a repeat_stream block in the flow graph.

rubund:
> Definitely a useful application, but I don’t see the advantage of concatenate for that use case. You can already build this flow graph with core blocks like stream_mux and head. You possibly could add the tag_strobe block to periodically inject tags.

The advantage of the concatenate block for these use cases, is that it does not require you to keep careful track of stream lengths and enter the same number several places. Just hook up the source, and a pure concatenate block, and the source decides the length.
 
> Another implementation is to look for a tag on each input specifying the number of samples to pull from each input before switching to the next one. .....   With this solution, you could make any number of repeating file/vector/signal combinations without requiring a repeat_stream block in the flow graph.

This is indeed a very good idea. If such a block does not already exist in GR, I would very much like to have something like that as well. If no number is present in the tag you mention, it could also just pull samples until the next time the tag comes. I do not think this replaces the need for a concatenate and a repeat_stream block though..

It is true that the repeat_stream block has one drawback when it comes to memory usage. It will naturally need to buffer the whole input stream in RAM, but in many situations, this is completely acceptable (the repeat_stream will have a configuration for max length).

n-west:
I won't spend too much effort on this stance, but I don't see much real utility in concatenating streams. If the GR API did provide a nice interface for checking if an upstream port is finished then this functionality might be a nice demo of that.

It looks like you have a large OOT module you're getting good utility out of full of blocks that operate on finite-length streams. Perhaps I'm missing why these blocks should live in GNU Radio and not in your OOT with a pybombs recipe.

jmcorgan:
I think the general consensus here is that this particular block functionality is more appropriate to stay as an OOT module vs. in-tree. However, it has triggered a more general discussion of GNU Radio runtime behavior and possible changes to make this sort of thing easier to do. 

@bastibl, please take ownership of the latter and either open an issue to discuss or kick off a mailing list thread for more input.

@rubund, please continue submitting your work for evaluation for in-tree use--these things get judged on a individual basis. It may be useful to start a thread on discuss-gnuradio beforehand to get community feedback.

rubund:
> I think the general consensus here is that this particular block functionality is more appropriate to stay as an OOT module vs. in-tree. However, it has triggered a more general discussion of GNU Radio runtime behavior and possible changes to make this sort of thing easier to do. 

Very cool if an easier way of doing this is implemented.


I had been trying to find out for quite some time how to get done what the concatenate block does (refer to mailing list thread about "How to find out when the last sample has arrived"). For me, it would have been very useful if this block had been available in GNU Radio in the first place, since I would have saved several days getting complete what I was working on. My investigation went through several iterations (among them using tags to mark end item), before I finally figured out how to do it one way which works very well and does not requires tags. Then I also wanted to share this functionality, so that nobody else had to go through the same pain again.

Among all the other blocks I have in my huge OOT module, I found the concatenate and repeat blocks to be the most general ones, and probably useful for others.

I will try to gather the most useful and general blocks I have made, and make them available in a public OOT module. The problem is that it is hard to find a name which makes it easy for people to find it when they need it (such as myself 3-4 weeks ago when I was really searching for such functionality..)

Cheers

PR #1298, opened 2017-04-25
Fix additive scrambler off by one error when reset is tag triggered

v1bri:
The additive scrambler's internal LFSR should be reset at/before the byte associated with the reset tag instead of after it. This will allow for proper operation in packet centric modes where the trigger tag is "packet_len" or something similar.

Per @jmcorgan this PR's first commit is a QA test that fails with current behavior, with the follow-up commit containing the fix.


noc0lour:
:+1: for the QA test (:

jmcorgan:
@v1bri As a bug fix, can you rebase this off maint and edit the PR? Thanks.

jmcorgan:
This was cherry-picked onto maint.

v1bri:
Sorry for the delay. Thanks for pulling in the fix!

PR #1299, opened 2017-04-25
grc: Converted actions to Gio.Action instead of Gtk.Action

sdh11:
The Gtk.Action class in Gtk3 has been deprecated in favor of Gio.Action. This converts all of the actions and rebuilds the menus/toolbar structures to handle the newer style.

### DIFF
--- a/grc/gui/DrawingArea.py
+++ b/grc/gui/DrawingArea.py
@@ -23,6 +23,7 @@ from gi.repository import Gtk, Gdk
 
 from .canvas.colors import FLOWGRAPH_BACKGROUND_COLOR
 from . import Constants
+from . import Actions
 
 
 class DrawingArea(Gtk.DrawingArea):
@@ -42,6 +43,7 @@ class DrawingArea(Gtk.DrawingArea):
         Gtk.DrawingArea.__init__(self)
 
         self._flow_graph = flow_graph
+        self.set_property('can_focus', True)
 
         self.zoom_factor = 1.0
         self._update_after_zoom = False
@@ -66,6 +68,11 @@ class DrawingArea(Gtk.DrawingArea):
             # Gdk.EventMask.FOCUS_CHANGE_MASK
         )
 
+        # This may not be the correct place to be handling the user events
+        # Should this be in the page instead?
+        # Or should more of the page functionality move here?
+        self.connect('key_press_event', self._handle_key_press)
+
         # setup drag and drop
         self.drag_dest_set(Gtk.DestDefaults.ALL, [], Gdk.DragAction.COPY)
         self.connect('drag-data-received', self._handle_drag_data_received)
@@ -78,11 +85,13 @@ class DrawingArea(Gtk.DrawingArea):
 
         def _handle_notify_event(widget, event, focus_flag):
             self._focus_flag = focus_flag
+
         self.connect('leave-notify-event', _handle_notify_event, False)
         self.connect('enter-notify-event', _handle_notify_event, True)
         # todo: fix
 #        self.set_flags(Gtk.CAN_FOCUS)  # self.set_can_focus(True)
 #        self.connect('focus-out-event', self._handle_focus_lost_event)
+
 
     ##########################################################################
     # Handlers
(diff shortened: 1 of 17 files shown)


jmcorgan:
@skoslowski Do you mind if I go ahead and merge this into python3 so we can just evaluate it there?  Right now it is a fast forward.

skoslowski:
I will review over the weekend. I am sure Seth is happy to rebase it, if
something else changes until then

grc/gui/DrawingArea.py
```diff
@@ -155,6 +164,41 @@ def _handle_mouse_motion(self, widget, event):
...
+            Gtk.accelerator_parse('Add'): Actions.PORT_CONTROLLER_INC,
+            Gtk.accelerator_parse('Subtract'): Actions.PORT_CONTROLLER_DEC,
+            Gtk.accelerator_parse('Return'): Actions.BLOCK_PARAM_MODIFY,
+            Gtk.accelerator_parse('<Shift>t'): Actions.BLOCK_VALIGN_TOP,
+            Gtk.accelerator_parse('<Shift>m'): Actions.BLOCK_VALIGN_MIDDLE,
+            Gtk.accelerator_parse('<Shift>b'): Actions.BLOCK_VALIGN_BOTTOM,
+            Gtk.accelerator_parse('<Shift>l'): Actions.BLOCK_HALIGN_LEFT,
+            Gtk.accelerator_parse('<Shift>c'): Actions.BLOCK_HALIGN_CENTER,
+            Gtk.accelerator_parse('<Shift>r'): Actions.BLOCK_HALIGN_RIGHT,
+        }
```
skoslowski: this should probably be a module variable (not recreated for each key-press event)

grc/main.py
```diff
@@ -65,16 +62,23 @@ def main():
...
 
     #formatter = utils.log.ConsoleFormatter()
     console.setFormatter(formatter)
     log.addHandler(console)
 
-    log.debug("Loading platform")
+    log.debug("Running main")
 
+    # Delay importing until the logging is setup
+    # Otherwise, the decorators could not use logging.
```
skoslowski: which decorators?

grc/gui/Notebook.py
```diff
@@ -99,20 +129,12 @@ def __init__(self, main_window, flow_graph, file_path=''):
...
 
-    def _handle_scroll_window_key_press(self, widget, event):
-        """forward Ctrl-PgUp/Down to NotebookPage (switch fg instead of horiz. scroll"""
-        is_ctrl_pg = (
-            event.state & Gdk.ModifierType.CONTROL_MASK and
-            event.keyval in (Gdk.KEY_Page_Up, Gdk.KEY_Page_Down)
-        )
-        if is_ctrl_pg:
-            return self.get_parent().event(event)
+
```
skoslowski: is this no longer required?

grc/gui/MainWindow.py
```diff
@@ -58,18 +59,29 @@ def __init__(self, app, platform, action_handler_callback):
         log.debug("__init__()")
 
         self._platform = platform
+        self.app = app
         self.config = platform.config
 
+        # Add all "win" actions to the local
+        win_actions = filter(lambda x: x.startswith("win."), Actions.get_actions())
+        map(lambda x: self.add_action(Actions.actions[x]), win_actions)
```
skoslowski: style: lambda with side effect only is better expressed as a for loop ... with if clause

grc/gui/DrawingArea.py
```diff
@@ -155,6 +164,41 @@ def _handle_mouse_motion(self, widget, event):
...
+        # Not sold on this.
+        if (key, mod) in accels:
+            accels[(key, mod)]()
+            return True
```
skoslowski: maybe some like
```
try:
    action = accels[(key, mod)]
except KeyError:
    return
else:
    action()
```
if there is no gtk way to do the binding this is okay...

grc/gui/MainWindow.py
noc0lour: map and lambda, Guido van Rossum would hate you (: 
@skoslowski wouldn't a list comprehension work as well, or is that what you meant?

skoslowski: map and lambda have valid use cases.

As no list is build here, a list comprehension isn't much better here IMO.

in python3 map() is a generator. I'm not really if this even had the intended function

jmcorgan:
I'm going to squash this branch into a single commit, to make the history cleaner, then put that on the tip of the current python3 branch. @sdh11 this will mean that you'll need to regenerate your local development branch on top of this.

grc/gui/MainWindow.py
sdh11: Nope, it doesn't have the intended function; don't know how I missed that.

grc/gui/Notebook.py
sdh11: I'm not sure why I deleted that. I'll put it back.

grc/gui/DrawingArea.py
sdh11: At the moment, I think this is the best way to do it.

PR #1300, opened 2017-04-26
blocks: tag_gate (improvements)

rubund:
- Make it possible to specify a single tag key to look for instead of all tags.
- Extended unit test for new functionality
- Actually fix a bug which made the unit test do nothing earlier

### DIFF
--- a/gr-blocks/grc/blocks_tag_gate.xml
+++ b/gr-blocks/grc/blocks_tag_gate.xml
@@ -44,7 +44,7 @@ self.$(id).set_single_key($single_key))</make>
 	  <type>int</type>
   </param>
   <param>
-    <name>Propagate_tags</name>
+    <name>Propagate Tags</name>
     <key>propagate_tags</key>
     <value>False</value>
     <type>enum</type>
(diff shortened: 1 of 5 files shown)


gr-blocks/grc/blocks_tag_gate.xml
```diff
@@ -54,6 +57,13 @@
       <key>False</key>
     </option>
   </param>
+
```
mhostetter: Could we also take this opportunity to change the name of the propagate tags parameter from `<name>Propagate_tags</name>` to `<name>Propagate Tags</name>` on line 47? The old name seems kinda wonky.

mhostetter:
I wonder if it's useful to allow multiple (not just single) tag keys to be gated by this block. Or if assumed usage would just be to use multiple Tag Gates, each gating a single key.

bastibl:
I guess since this [commit](https://github.com/gnuradio/gnuradio/commit/080bbf3dfcef968a2141714852bf691ba442bb2f) `setup_rpc` is not called for all blocks

rubund:
@mhostetter I am sure it would be useful. However, I am not so sure how it then should be implemented. Can we provide a list of strings through the swig interface to c++? How would the GRC configuration look like in order to facilitate providing multiple strings? (maybe just have a field which accepts a list)

PR #1301, opened 2017-04-26
blocks: file_source ("file_begin" stream tag)

rubund:
- Add a configurable "file_begin" stream tag
  which is especially useful in repeat mode.
- Added unit tests for new functionality

### DIFF
--- a/gr-blocks/python/blocks/qa_file_source_sink.py
+++ b/gr-blocks/python/blocks/qa_file_source_sink.py
@@ -107,23 +107,22 @@ class test_file_source_sink(gr_unittest.TestCase):
         src_data = range(1000)
         expected_result = range(1000)
 
-        filename = "tmp.32f"
-        src = blocks.vector_source_f(src_data)
-        snk = blocks.file_sink(gr.sizeof_float, filename)
-        snk.set_unbuffered(True)
-
-        src2 = blocks.file_source(gr.sizeof_float, filename)
-        src2.set_begin_tag(True)
         snk2 = blocks.vector_sink_f()
 
-        self.tb.connect(src, snk)
-        self.tb.run()
+        with tempfile.NamedTemporaryFile() as temp:
+            src = blocks.vector_source_f(src_data)
+            snk = blocks.file_sink(gr.sizeof_float, temp.name)
+            snk.set_unbuffered(True)
 
-        self.tb.disconnect(src, snk)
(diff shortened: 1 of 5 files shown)


gr-blocks/python/blocks/qa_file_source_sink.py
```diff
@@ -104,6 +105,67 @@ def test_file_source_can_seek_after_open(self):
 
         os.remove(filename)
 
+    def test_begin_tag(self):
+        src_data = range(1000)
+        expected_result = range(1000)
+
+        filename = "tmp.32f"
```
jmcorgan: You can use Python's tempfile module and ``with`` construct here, which will create a unique filename and remove it automatically.

rubund: Good idea. I just did it the same way as for the other tests in qa_file_source_sink.py. Do you mean I should change it for the "old" tests as well ? Would it not be outside the scope of the pull request then ?

jmcorgan: We have lots of bad habits strewn throughout the tree, which unfortunately get copied by newcomers.  So, go ahead and do it the correct way for this new test, and for extra credit, submit a new PR with changes to the other ones.

rubund: I have now done what you suggested. There is a new PR for the old parts of the test here: https://github.com/gnuradio/gnuradio/pull/1311

jmcorgan:
I'd like to see the begin tag configurable instead of statically set.  This would also eliminate the need for two different bools to track config/state.  The tag id would start out as NIL; the set_begin_tag() call would take a pmt and set the key id (which would become a member variable instead of static). Finally, the work function would test if the key id was NIL to see if the feature was turned on.

This also would mean the GRC XML would need to test if the param was supplied or not and only conditionally call the set_begin_tag() method.

rubund:
I pushed a commit where set_begin_tag has been converted to accept a pmt::pmt_t instead bool. Please let me know if I understood your suggestion correctly.

> This also would mean the GRC XML would need to test if the param was supplied or not and only conditionally call the set_begin_tag() method.

I do actually not know how to do this, nor have I been able to found this being done for any other blocks. Can you give me an example?

Currently, the GRC model is non-working if no argument is supplied.

PR #1303, opened 2017-04-30
cmake: add Coverage target and CMake module to generate coverage data

noc0lour:
This is an example how we could add a `coverage` target to cmake. 
I'm not entirely sure about the change in the toplevel cmake but I didn't find a better place for it.

This will enable generation of test coverage reports.
Requirements: 
 - Installed lcov (and in $PATH)

how to run:
`cmake -DCMAKE_BUILD_TYPE=Coverage $src_dir && make && make coverage` will build GNU Radio with the required flags and run the testsuite collecting coverage data afterwards and placing results in `build/coverage`. 
Example: https://files.gnuradio.org/dev/coverage/


noc0lour:
This will also enable analyzing change in coverage for each PR with the help of https://codecov.io

PR #1304, opened 2017-04-30
gr-fft: add locking mechanism to fftw_wisdom

noc0lour:
Add a lockfile for access to the fftw wisdom. 
Unfortunately in Linux the lockfile will be locked for all threads in one process so there has to be a mutex to lock other threads out as well.

I didn't test running `all` GR tests with a high job number.

### DIFF
--- a/gr-fft/lib/fft.cc
+++ b/gr-fft/lib/fft.cc
@@ -46,10 +46,12 @@ static int my_fftw_read_char(void *f) { return fgetc((FILE *) f); }
 
 #include <boost/filesystem/operations.hpp>
 #include <boost/filesystem/path.hpp>
+#include <boost/interprocess/sync/file_lock.hpp>
 namespace fs = boost::filesystem;
 
 namespace gr {
   namespace fft {
+    static boost::mutex wisdom_thread_mutex;
 
     gr_complex *
     malloc_complex(int size)
@@ -89,6 +91,30 @@ namespace gr {
       static fs::path path;
       path = fs::path(gr::appdata_path()) / ".gr_fftw_wisdom";
       return path.string();
+    }
+
+    static void
+    lock_wisdom()
+    {
+      const std::string wisdom_lock_file = wisdom_filename() + ".lock";
+      int fd = open(wisdom_lock_file.c_str(),
+                    O_WRONLY|O_CREAT|O_NOCTTY|O_NONBLOCK,
+                    0666);
+      if (fd < 0){
+        throw std::exception();
+      }
+      boost::interprocess::file_lock wisdom_lock(wisdom_lock_file.c_str());
+      wisdom_lock.lock();
+      wisdom_thread_mutex.lock();
+    }
+
+    static void
+    unlock_wisdom()
+    {
+      const std::string wisdom_lock_file = wisdom_filename() + ".lock";
+      boost::interprocess::file_lock wisdom_lock(wisdom_lock_file.c_str());
+      wisdom_lock.unlock();
+      wisdom_thread_mutex.unlock();
     }
 
     static void
@@ -162,6 +188,7 @@ namespace gr {
 
       d_nthreads = nthreads;
       config_threading(nthreads);
+      lock_wisdom();
       import_wisdom();	// load prior wisdom from disk
 
       d_plan = fftwf_plan_dft_1d (fft_size,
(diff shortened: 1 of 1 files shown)


jmcorgan:
I haven't looked through the code yet, but a quick trial merge and parallel ctest resulted in the below:

```
$ ctest -j214 --output-on-failure

94% tests passed, 12 tests failed out of 214

Total Test time (real) =  24.85 sec

The following tests FAILED:
	 97 - qa_fft (Failed)
	100 - qa_fft_filter (Failed)
	104 - qa_freq_xlating_fft_filter (Failed)
	107 - qa_pfb_synthesizer (Failed)
	117 - qa_pfb_channelizer (Failed)
	119 - qa_pfb_interpolator (Failed)
	120 - qa_pfb_decimator (Failed)
	144 - qa_ofdm_txrx (Failed)
	150 - qa_correlate_and_sync (Failed)
	152 - qa_ofdm_sync_sc_cfb (Failed)
	165 - qa_ofdm_serializer_vcc (Failed)
	197 - qa_qtgui (Failed)
Errors while running CTest
```
I think this is every test for something that uses gr::fft inside. All of the failures look like this:
```
1/214 Test  #97: qa_fft ...............................***Failed    7.85 sec
fftw: planner.c:891: assertion failed: flags.u == u
Aborted (core dumped)
```
Full log: [ctest.txt](https://github.com/gnuradio/gnuradio/files/967926/ctest.txt)

jmcorgan:
This may have been a result of a corrupted fft_wisdom file from previous testing, rechecking now.

jmcorgan:
Unfortunately, even with this PR, the wisdom file is getting corrupted by parallel tests.  However, setting .gr_fftw_wisdom to a 0 byte read-only file results in all tests passing ctest -j214, ten times in a row.

noc0lour:
hm, that's strange, probably there is still something wrong in the code,I can put some debug prints in sometime this week and run parallel tests somewhere.

jmcorgan:
@darek-kawamoto wrote an interface class to allow better coordination between GNU Radio and OOT modules for managing FFTW wisdom/planner sharing. I collected his commits and made a branch off current maint [here](https://github.com/jmcorgan/gnuradio/commit/743063e5cc4ff5477cd74f61aae66a5270798bf8).

While this might be overkill for what we need, within it is a similar locking mechanism to what you wrote in the PR.

@darek-kawamoto Can you do a quick summary of the changes and how they might be applicable here?

darek-kawamoto:
boost::interprocess::file_lock attempts interprocess-synchronization of wisdom file access.

The other problem is that currently, every gr-fft block creation results in a hit to the wisdom file (a read and a write).  A cached wisdom file manager only reads the wisdom file once, and is only allowed to write once every 100 ms.

A cached-concurrent manager combines both of the above.

At library load, an initializer in the anonymous namespace sets up the basic manager and queues up the teardown using std::atexit().

jmcorgan:
@noc0lour Given that the lock/unlock wisdom calls are only ever paired with calls to import/export wisdom, it would be simpler to just add the lock/unlock logic to the import/export functions themselves.

noc0lour:
Putting the locking and unlocking in the import/export is possible as well. I thought putting it in the top level function is clearer for other uses of the fftw-wisdom.

jmcorgan:
I've done some more testing and I'll go ahead and merge this as-is.  It may not be the complete solution but it's clear it is part of it.

gnieboer:
FYI, on an MSVC build this is hanging on the test_multithreaded test, it appears that lock is being called on the thread_mutex locked twice and never unlocked.  It'll take a bit more time to set up a specific test scenario to step into.  All I can say for the moment is that the second lock attempt happens during the make() call.

gnieboer:
OK, found the problem.  A new wisdom_lock object is being created in the wisdom_unlock() routine, which then throws an exception when unlock() is called because the new object isn't locked.  This then aborts the make() routine without unlocking the original object.  Then when wisdom_lock() gets called in the next test, it's already locked and the application hangs.  To fix just create a single wisdom_lock object and reuse it for both calls.  Pull request forthcoming.

noc0lour:
@gnieboer thanks for debugging (: I was sure there is still some flaw hidden somwhere ;)

gnieboer:
#1332 is the pull request.  No problems!

PR #1310, opened 2017-05-11
qtgui: Tweaked the QT Label to be more friendly to floating point num…

SaikWolf:
…bers with custom formatter. Previously this failed due to the use of eng_notation.

This prevents eng_notation from wrapping around the custom formatting.
Error can be seen here: http://gnuradio.4.n7.nabble.com/How-do-I-use-formatter-in-QT-GUI-Label-td59929.html


jmcorgan:
This was cherry-picked onto the maint branch.

PR #1312, opened 2017-05-16
gnuradio-runtime: Update pkgconfig to reflect use of the log4cpp libr…

balister:
…ary.

Signed-off-by: Philip Balister <philip@balister.org>


bastibl:
Shouldn't this also add `-L<path of log4cpp>`?

balister:
Probably :) But I suspect that  Lot's of pkgconfig files don't account for this.

PR #1313, opened 2017-05-17
runtime: fix missing link requirement in pkg-config file

jmcorgan:


jmcorgan:
This is for 3.7 (maint/master), where log4cpp is optional. #1312 is still needed for next, where log4cpp is required.

PR #1321, opened 2017-06-03
gr-qtgui: allows time_sink to have up to 24 inputs instead of 12

### LINKED ISSUE #1317: qtgui_time_sink can't display more than 12 inputs
I noticed that qtgui_time_sink segfaults when I provide more than 12 inputs.  I was able to fix the issue and allow more inputs by going into TimeDomainDisplayPlot.cc and adding more QColors to the "colors" list (I added the last line to allow 3 more inputs, pushing it to 15 total):

QList<QColor> colors;
colors << QColor(Qt::blue) << QColor(Qt::red) << QColor(Qt::green)
	 << QColor(Qt::black) << QColor(Qt::cyan) << QColor(Qt::magenta)
	 << QColor(Qt::yellow) << QColor(Qt::gray) << QColor(Qt::darkRed)
	 << QColor(Qt::darkGreen) << QColor(Qt::darkBlue) << QColor(Qt::darkGray) 
         << QColor(Qt::darkGray) << QColor(Qt::darkGray) << QColor(Qt::darkGray); 

I guess those are the default colors?  Either way, we should probably allow for more than 12 inputs if desired, maybe more like 20 or 30, even if the colors are all the same.  If you're curious, the reason I am using so many damn inputs is because I only display a few at a time, but I need to be able to choose from a large number.  I've never put in a pull request before, but I am happy to if someone can point me to instructions for doing it properly.

777arc:
fixes #1317

### DIFF
--- a/gr-qtgui/lib/TimeDomainDisplayPlot.cc
+++ b/gr-qtgui/lib/TimeDomainDisplayPlot.cc
@@ -170,8 +170,14 @@ TimeDomainDisplayPlot::TimeDomainDisplayPlot(int nplots, QWidget* parent)
   colors << QColor(Qt::blue) << QColor(Qt::red) << QColor(Qt::green)
 	 << QColor(Qt::black) << QColor(Qt::cyan) << QColor(Qt::magenta)
 	 << QColor(Qt::yellow) << QColor(Qt::gray) << QColor(Qt::darkRed)
+	 << QColor(Qt::darkGreen) << QColor(Qt::darkBlue) << QColor(Qt::darkGray)
+	 // cycle through all colors again to increase time_sink_f input limit
+	 // from 12 to 24, otherwise you get a segfault
+	 << QColor(Qt::blue) << QColor(Qt::red) << QColor(Qt::green)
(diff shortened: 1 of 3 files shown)


jmcorgan:
I'm not averse to merging this per se, but I'd rather see the limits documented and enforced, then we can consider what the limit should be.

777arc:
Oh I totally agree, although there will be a bit of changes to do that.  One reason is that on the GRC side you are limited to 10 inputs because of the way the labels, widths, colors, etc. lists are generated (they are only 10 elements).  So without editing python code you can only do 10 inputs, although once again there are no enforced limits for that either (the flowgraph just fails to run if you do more than 10).

777arc:
I added code to time_sink_c and time_sink_f to catch the user providing more than 12 or 24 inputs respectively.  This enforces the limits for someone writing a python flowgraph.  On the GRC side, a lot will have to be changed if we want the 12/24 input limit instead of just the current limit of 5/10, and I'm not quite comfortable enough messing with those files.

jmcorgan:
I think we'll need to take a deeper look at how to handle this with GRC at the same time. I'm closing this for now; the issue will remain open to track that.

PR #1322, opened 2017-06-05
sort input files

bmwiedemann:
when building packages (e.g. for openSUSE Linux)
(random) filesystem order of input files
influences ordering of functions in the output,
thus without the patch, builds (in disposable VMs) would usually differ.

See https://reproducible-builds.org/ for why this matters.

such diffs affected volk-config-info volk_profile volk.h volk_typedefs.h gnuradio/xml/index.xml gnuradio/html/volk_8h_source.html etc
http://rb.zq1.de/compare.factory-20170531/gnuradio-compare.out


bmwiedemann:
note: this one is related: https://github.com/gnuradio/volk/pull/116

PR #1327, opened 2017-06-09
uhd: Fixed error case where one multi-sub-dev spec was given on a single device

mbr0wn:


mbr0wn:
Just FYI, I'm not splitting up bug fixes just to increase the activity :)

mbr0wn:
Actually, I have a better fix for this. Let me revise.

PR #1328, opened 2017-06-09
uhd: Properly normalized selection of options for multi-device ops

mbr0wn:
Before, there was an issue where things could go awry with multiple
devices and only specifying a single --spec option.


mbr0wn:
Intentionally on master; on maint, this code looks different.

Issue #1337, opened 2017-06-20
multiply_matrix_XX: can't simply use magic value in place of `enum`, C++ doesn't allow that


marcusmueller:
https://github.com/gnuradio/gnuradio/blob/master/gr-blocks/lib/multiply_matrix_cc_impl.cc#L182 👍
```
[793/2135] Building CXX object gr-blocks/lib/CMakeFiles/gnuradio-blocks.dir/multiply_matrix_cc_impl.cc.o
../gr-blocks/lib/multiply_matrix_cc_impl.cc:184:15: warning: comparison of constant 999 with expression of type 'gr::block::tag_propagation_policy_t' is always false [-Wtautological-constant-out-of-range-compare]
      if (tpp == TPP_SELECT_BY_MATRIX) {
          ~~~ ^  ~~~~~~~~~~~~~~~~~~~~
1 warning generated.
```

Simply because you might in practice be able to cast the `tag_propagation_policy_t` argument to `set_tag_propagation_policy` to an `int` doesn't mean you can just set it to `999`. That's not how `enum`s work (C++ standard, highlight mine):

> [expr.static.cast]/10

>    A value of integral or enumeration type can be explicitly converted to an enumeration type. The value is unchanged if the original value is within the range of the enumeration values (7.2). **Otherwise, the resulting value is unspecified (and might not be in that range).**

So, yay, undefined behaviour occurs whenever one tries to actually pass `TPP_SELECT_BY_MATRIX` to `set_tag_propagation_policy` (which isn't that surprising, considering enums were meant to suppress the usage of numeric literals as magic numbers without compile-time range checking).

Another issue is that `tag_propagation_policy()` will return `TPP_DONT` instead of `TPP_SELECT_BY_MATRIX` after `set_tag_propagation_policy(TPP_SELECT_BY_MATRIX)`, which is a breach of contract.

So, I'm not sure how to proceed. Options:

1. extend `tag_propagation_policy_t` to include `TPP_SELECT_BY_MATRIX`. That demonstrates that we can't keep the number of possible TPPs limited.
2. extend `tag_propagation_policy_t` to include `TPP_CUSTOM`, and change `multiply_matrix` + docs to react to that
3. ???

gigabyte1023:
i need help

mbr0wn:
We actually had a discussion about this when this block was first merged. The `TPP_CUSTOM` solution was the one that I preferred at the time.

marcusmueller:
[...]

PR #1341, opened 2017-06-24
fix: can't pass non-enum integer as enum parameter (#1337)

marcusmueller:
Adresses #1337. Much leet. such wow.


marcusmueller:
Should I be doing separate PRs for master and next, too?

jmcorgan:
@marcusmueller The only reason you'd need master/next PRs is if this didn't merge up cleanly. It looks like it does.

mbr0wn:
Is this really a maint fix though?

marcusmueller:
@mbr0wn well, considering you *can't* use TPP_SELECT_BY_MATRIX with my compiler -O2, since it simply gets rid of the complete branch since it's C++-wise OK to declare it impossible, yes, that is broken functionality, thus, this is a bug fix.

PR #1343, opened 2017-06-24
Fixed storing references to temporary values in RPC infrastructure

marcusmueller:
The RPC aggregator and friends were taking callback objects passed by
value, and then internally called, again by value, constructors which
initialized reference fields to these parameters.

This has the effect of storing references to temporary objects. Chances
are, no-one is using that code, or compiler UB saved our collective
behinds.

In any case, we might need to more carefully dissect the RPC code. There
be dragons.


jmcorgan:
This technically an API change but existing code will still compile okay, so no worries.

jmcorgan:
This has been reverted.

Issue #1348, opened 2017-06-27
Filter window calculation illegal calculation.


cor-mt:
This bug get triggered on an "AMD FX(tm)-8150 Eight-Core Processor" when gnuradio is compiled with `CFLAGS="-march=native -O2"`.

In window.cc line 265, with `i == ntaps - 1` gives `temp = 1.0000000000000000002` that cause `sqrt (0)` witch return "-nan" on the next line for the last value of the vector. This value get used for the gain calculation in the low_pass filter and thus screwing all the rest of the calculations.

marcusmueller:
Thanks for documenting/reporting this here 👍 

For future readers: http://lists.gnu.org/archive/html/discuss-gnuradio/2017-06/msg00277.html

marcusmueller:
Just to be clear: this is a serious bug in the way we do floating point math, and I (hopefully) have a patch coming up. We need to be vigilant, though – this is most likely not the only mistake of this type.

marcusmueller:
```
diff --git a/gr-fft/lib/window.cc b/gr-fft/lib/window.cc
index 126b28978..e80d469e9 100644
--- a/gr-fft/lib/window.cc
+++ b/gr-fft/lib/window.cc
@@ -30,9 +30,10 @@
 namespace gr {
   namespace fft {
 
-#define IzeroEPSILON 1E-21               /* Max error acceptable in Izero */
+#define BESSEL_EPSILON 1E-21               /* Max error acceptable in bessel_i_zero */
 
-    static double Izero(double x)
+    /*! \brief Iterative first kind Bessel function approximation */
+    static double bessel_i_zero(double x)
     {
       double sum, u, halfx, temp;
       int n;
@@ -45,7 +46,7 @@ namespace gr {
         temp *= temp;
         u *= temp;
         sum += u;
-      } while (u >= IzeroEPSILON*sum);
+      } while (u >= BESSEL_EPSILON*sum);
       return(sum);
     }
 
@@ -257,13 +258,17 @@ namespace gr {
 
       std::vector<float> taps(ntaps);
 
-      double IBeta = 1.0/Izero(beta);
+      double bessel_i_beta = 1.0/bessel_i_zero(beta);
       double inm1 = 1.0/((double)(ntaps-1));
       double temp;
 
-      for(int i = 0; i < ntaps; i++) {
+      /* dragging taps[0] out of the loop, because ((double)(-1))*((double)(-1))
[...]

PR #1349, opened 2017-06-27
Fixed floating point math bug (#1348)

marcusmueller:
https://github.com/gnuradio/gnuradio/issues/1348

We were doing floating point math wrong.

### DIFF
--- a/gr-fft/lib/window.cc
+++ b/gr-fft/lib/window.cc
@@ -261,10 +261,19 @@ namespace gr {
       double inm1 = 1.0/((double)(ntaps-1));
       double temp;
 
-      for(int i = 0; i < ntaps; i++) {
+      /* extracting first and last element out of the loop, since
+         sqrt(1.0-temp*temp) might trigger unexpected floating point behaviour
+         if |temp| = 1.0+epsilon, which can happen for i==0 and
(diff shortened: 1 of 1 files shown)


jmcorgan:
It's more difficult to look at the actual change that fixes the bug when it is buried in a variable name change. Can you split this into two commits, one that fixes the bug and a second that renames the variable?

marcusmueller:
done!

marcusmueller:
hold this. Apparently, though I still stand by the math improvements done here, it's not solving the issue.

marcusmueller:
Cor confirms fix actually fixes bug. @jmcorgan if you could review these changes? I guess I don't have to clean up the commits, since you'll probably squash them, anyways?

jmcorgan:
I'd sill like to have one commit for the bug fix and one commit for the renaming.

jmcorgan:
...and with better commit messages :-)

marcusmueller:
Since the order in which you wanted the commits always required manual rewrite (I renamed first, then fixed, you want the opposite), will not do the rename. I want this critical fix to go through, not be pretty.

jmcorgan:
:+1:

PR #1350, opened 2017-06-27
Fix: made callbacks references in rpcmanager_base, forgot Thrift

marcusmueller:


jmcorgan:
The original PR #1343 was reverted; you'll need to start over with a fresh PR.

PR #1351, opened 2017-06-28
Fix compilation under MinGW

pcercuei:
This set of two commits fix the compilation of gnuradio-runtime and gr-fft under mingw-w64, on 32-bit and 64-bit versions of Windows.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -151,6 +151,10 @@ endif(MSVC)
 
 if(WIN32)
 	add_definitions(-D_USE_MATH_DEFINES)
+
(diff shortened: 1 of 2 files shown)


pcercuei:
Good point, I didn't see it.

jmcorgan:
> I think there's already a patch for this in master. Maybe we should cherry-pick it for maint instead of adding another fix?

Hmm, yeah, we should be able to do that. I'll be working the queue in a bit and will take care of it.

pcercuei:
OK, I'll drop the commit from this PR then.

PR #1354, opened 2017-06-30
BUGFIX: Removed persistent logging to console present in correlate_access_code_bb_impl.cc

LaughingCabbage:
Removed persistent logging to console present in packet decode as a result of new_flag calls in correlate_access_code_bb_impl.cc. Test flow: File source -> packet encoder -> dpsk mod -> dpsk demod -> packet decoder -> time sink.

### DIFF
--- a/gr-digital/lib/correlate_access_code_bb_impl.cc
+++ b/gr-digital/lib/correlate_access_code_bb_impl.cc
@@ -109,6 +109,8 @@ namespace gr {
 	// test for access code with up to threshold errors
 	new_flag = (nwrong <= d_threshold);
 
+/*
+	//Removed due to persistent logging to console of access code on each call to new_flag
 	if(new_flag) {
 	  GR_LOG_DEBUG(d_logger, boost::format("access code found: %llx") % d_access_code);
 	}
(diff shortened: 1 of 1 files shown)


LaughingCabbage:
@jmcorgan would you like me to repull this into maint? Really, the packet decode block is pretty useless without removing the logging.

jmcorgan:
I'm trying to understand why this is a problem now and not before.

LaughingCabbage:
@jmcorgan So I've just retested running a build off maint and I have no issue. However, looking at the main branch again, it appears that the 'if #VERBOSE' directive is missing from the main branch.

jmcorgan:
Can you highlight the two areas of code on each branch?

LaughingCabbage:
@jmcorgan [The line from the master branch](https://github.com/gnuradio/gnuradio/blob/master/gr-digital/lib/correlate_access_code_bb_impl.cc#L111) and [the line from the maint branch](https://github.com/gnuradio/gnuradio/blob/maint/gr-digital/lib/correlate_access_code_bb_impl.cc#L112)

jmcorgan:
Ok, that's strange as master is usually a strict superset of maint. I'll look into it.

jmcorgan:
Ok, this was a result of #1292. I'm merging this, but actually snipping out the code altogether.

PR #1360, opened 2017-07-05
packet.py _packet_encoder_thread run() buffer clips payload data

LaughingCabbage:
The current run() function in the _packet_encoder_thread clips the end of samples who's sample length modulo payload_length is greater than zero. The hack applied here is to append padding characters to satisfy the payload_length and allow for the remaining sample to be flushed on the next cycle. Current testing transmits a .jpg using a flow as follows: File source -> packet encoder -> dpsk mod -> dpsk demod -> packet decoder -> file sink. The input file is 37,075 bytes in size. The output file is 36,864 bytes in size. A 211 byte difference still needs to be tracked and patched.

### DIFF
--- a/gr-digital/lib/correlate_access_code_bb_impl.cc
+++ b/gr-digital/lib/correlate_access_code_bb_impl.cc
@@ -109,6 +109,8 @@ namespace gr {
 	// test for access code with up to threshold errors
 	new_flag = (nwrong <= d_threshold);
 
+/*
+	//Removed due to persistent logging to console of access code on each call to new_flag
 	if(new_flag) {
 	  GR_LOG_DEBUG(d_logger, boost::format("access code found: %llx") % d_access_code);
 	}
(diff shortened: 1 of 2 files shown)


jmcorgan:
Thanks for your pull request, we appreciate you taking the time to submit fixes to the project. Since it looks like you might be doing this regularly, let me go over some good practices when submitting a PR. 

First, your commit history for this PR includes 5d120ad, which is already part of #1354, which is independent of this.

Second, for clean repository history we like to consolidate commits into a small number of discrete changes, to make it easier to review the impact of each one. In this case, cd24e81 and 63e7efa should be collapsed into a single bug fix commit.

Third, it's very helpful in a large code base to prefix commit messages with the component so we can filter on that later. In this case it would be gr-digital: as a prefix.

Finally, we maintain a separate maintenance branch for bug fixes that gets merged up into our master branch, so for bugfixes we like have them based off 'maint' and pull requested back into 'maint'.

LaughingCabbage:
I have repulled this commit with a branch based off maint in #1370 . Apologies for the mess.

PR #1363, opened 2017-07-07
gr-digital: reformat file_source_impl for readability

LaughingCabbage:
Modify to eliminate alternating tab and spaces which makes readability difficult depending on the dev editor. Add end of function comments.


LaughingCabbage:
It appears that I've based this pull off maint and the build has failed. I will work to correct this.

PR #1364, opened 2017-07-07
File source impl reformat

LaughingCabbage:
Modify to eliminate alternating tab and spaces which makes readability difficult depending on the dev editor. Add end of function comments.


jmcorgan:
It looks like you branched off of master to create this instead of off maint.

PR #1355, opened 2017-06-30
uhd: Fixed tx streamer creation in usrp_sink

sugandhagupta:
Fix for #1353

Issue #1353, opened 2017-06-29
uhd: unlocking flowgraph in gnuradio disconnects USRP-Sink


sugandhagupta:
@noc0lour reported: 

Running a simple flowgraph with a USRP-sink connected and performing a lock() and unlock() of the flowgraph triggers an error in the transmitter and it stop transmitting.
This error is encountered with latest uhddev/master 1b70b3e and gnuradio/master but not with uhddev/3.9-LTS fdec013 and gnuradio/master.

stdout:
```
linux; GNU C++ version 4.9.3; Boost_105600; UHD_003.011.000.git-95-g811444b1

-- X300 initialization sequence...
-- Determining maximum frame size... 1472 bytes.
-- Setup basic communication...
-- Loading values from EEPROM...
-- Setup RF frontend clocking...
-- Radio 1x clock:200
-- Detecting internal GPSDO.... No GPSDO found
-- [DMA FIFO] Running BIST for FIFO 0... pass (Throughput: 1177.5MB/s)
-- [DMA FIFO] Running BIST for FIFO 1... pass (Throughput: 1182.5MB/s)
-- [RFNoC Radio] Performing register loopback test... pass
-- [RFNoC Radio] Performing register loopback test... pass
-- [RFNoC Radio] Performing register loopback test... pass
-- [RFNoC Radio] Performing register loopback test... pass
-- Performing timer loopback test... pass
-- Performing timer loopback test... pass
locking....
unlocking...
Press Enter to quit: thread[thread-per-block[1]: <block gr uhd usrp sink (1)>]: RuntimeError: On node 0/DmaFIFO_0, input port 1 is already connected.
```

flowgraph:
```
#!/usr/bin/env python2
# -*- coding: utf-8 -*-
##################################################
# GNU Radio Python Flow Graph
# Title: Top Block
# Generated: Fri Oct  7 16:36:43 2016
##################################################

from gnuradio import analog
from gnuradio import eng_notation
from gnuradio import gr
from gnuradio import uhd
from gnuradio.eng_option import eng_option
from gnuradio.filter import firdes
from optparse import OptionParser
import time


class top_block(gr.top_block):

    def __init__(self):
        gr.top_block.__init__(self, "Top Block")

        ##################################################
        # Variables
        ##################################################
        self.samp_rate = samp_rate = 1e6

        ##################################################
        # Blocks
        ##################################################
        self.uhd_usrp_sink_0 = uhd.usrp_sink(
[...]

PR #1365, opened 2017-07-07
uhd: Fixed tx streamer creation in usrp_sink

sugandhagupta:
Fix for #1353 
rebased to maint: https://github.com/gnuradio/gnuradio/pull/1355


sugandhagupta:
@mbr0wn Rx_streamer was correct. Had checked it too.

Issue #1362, opened 2017-07-07
File Sink/Source clipping data


LaughingCabbage:
Data is clipped during a file source / file sink flow. Clipping is varies according to file sink "Unbuffered" setting. With unbuffered OFF: Input file is 37,075 bytes and output file is 36,864 bytes leaving a 211 byte loss from input to output. With unbuffered ON: Input file is 37,075 bytes and output file is 37,072 bytes leaving a 3 byte loss. Bug discovered in testing of pull [#1360](https://github.com/gnuradio/gnuradio/pull/1360). Additionally, bug was noted on email list where users could only pipe data correctly with file source set to repeat. The repeat could potentially cause a flush of the buffer holding the missing data. TEST FLOW: File Source(repeat off) -> File Sink.

mbr0wn:
I took the liberty of renaming the issue; clipping usually indicates that the samples go through but get modified.

jmcorgan:
There's a recurring, difficult to reproduce bug introduced several years ago with finite flowgraph shutdown where some parts of the graph are shutdown prematurely. Not sure if this is related but but worth considering.

LaughingCabbage:
So I face-palmed pretty hard here. The flow graph I used was set to pass float values where as my file should have been read in as a byte format. In doing so I told the file source to consume 4 bytes at a time. However, my test file's size did not divide evenly by 4 and a 3 byte remainder was never read. I've added a warning in #1381 to hopefully make this a bit more clear for users.

noc0lour:
So this is still an issue? Or is the warning enough since GNU Radio can't do anything about truncated data?

LaughingCabbage:
@noc0lour this shouldn't be an issue anymore. The warning should suffice.

frestuc:
Hi, the bug is still there. Still the sink does not dump every received byte...

noc0lour:
So this might be expected behaviour if the total size of the file is not a multiple of your item size I think?

frestuc:
No, it's truncating a lot, like 50kBytes out of 380 kBytes. Also tried to directly dump via file write but still truncating...

noc0lour:
Can you elaborate a bit? If I plug a file into my source and stream it will not read everything out of that file? Or if I stream into a file it will not write everything to file?

frestuc:
Source and sink are byte-size items, so it should be fine right?
I am trying with tx_ofdm and rx_ofdm.py connected together.

frestuc:
[...]

PR #1370, opened 2017-07-12
gr-digital: packet.py _packet_encoder_thread run() buffer clips payload data

LaughingCabbage:
This is a repull of #1360 to base the commit off maint. The current run() function in the _packet_encoder_thread clips the end of samples who's sample length modulo payload_length is greater than zero. The hack applied here is to append padding characters to satisfy the payload_length and allow for the remaining sample to be flushed on the next cycle. Current testing transmits a .jpg using a flow as follows: File source -> packet encoder -> dpsk mod -> dpsk demod -> packet decoder -> file sink. The input file is 37,075 bytes in size. The output file is 36,864 bytes in size. A 211 byte difference still exists from the file source / sink and is noted in #1362 .

### DIFF
--- a/gr-digital/python/grc_gnuradio/blks2/packet.py
+++ b/gr-digital/python/grc_gnuradio/blks2/packet.py
@@ -54,13 +54,21 @@ class _packet_encoder_thread(_threading.Thread):
         self.start()
 
     def run(self):
-        sample = '' #residual sample
         while self.keep_running:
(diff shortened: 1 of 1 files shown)


jmcorgan:
This PR will get through, but as a regular contributor we're going to have to get a copyright license agreement from you for the Free Software Foundation. Please email me or Ben to get this started.

PR #1372, opened 2017-07-12
gr-blocks: Reformat file_source_impl.cc for readability.

LaughingCabbage:
Fix inconsistent whitespace to make this file readable.

### DIFF
--- a/gr-blocks/lib/file_source_impl.cc
+++ b/gr-blocks/lib/file_source_impl.cc
@@ -51,143 +51,143 @@
 #endif
 
 namespace gr {
-  namespace blocks {
+	namespace blocks {
 
-    file_source::sptr file_source::make(size_t itemsize, const char *filename, bool repeat)
-    {
-      return gnuradio::get_initial_sptr
-	(new file_source_impl(itemsize, filename, repeat));
-    }
+file_source::sptr file_source::make(size_t itemsize,
+									const char *filename,
+									bool repeat)
+{
+	file_source_impl *ptr = new file_source_impl(itemsize,
+												  filename,
+												  repeat);
+	return gnuradio::get_initial_sptr(ptr);
+} //end make
 
-    file_source_impl::file_source_impl(size_t itemsize, const char *filename, bool repeat)
-      : sync_block("file_source",
-		      io_signature::make(0, 0, 0),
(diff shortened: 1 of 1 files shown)


jmcorgan:
It's not clear what you've changed here, other than to shift the inner namespaced code to the left margin, which is not really something that's a problem. Can you elaborate on what you did?

mbr0wn:
All the indents were changed to 4 spaces it seems. Now, 4 spaces is my personal preference, but for historical reasons, GR has been using 2 spaces since forever. This diff would obliterate any history on the file (i.e., all git blames go to @LaughingCabbage now), and also introduce a different indentation style.

LaughingCabbage:
@jmcorgan The file has quite a few if statements that are indented using spaces but their children are then indented using single tabs. This leads to the statement's children appearing in-line or before the original if statements [as in this](https://pastebin.com/pPu0DY3X).  I've also modified a few lengthier lines that exceed the character limit and added braces to avoid confusion on single line if statements as per @drmpeg .

 @mbr0wn I have used strictly a tabs as this gives the developer the option to adjust tab sizes in their editor. Feel free to modify as you see fit as the tab/space war is not a battle I intend to fight. As for the blame history, I understand how this is problematic but I have no solution. This pull feels nit-picky but I've identified some issues with the file source/sink in #1362 and I feel readability is important moving forward.

jmcorgan:
I think we should address these types of issues codebase-wide; I'm going to close this in favor of doing something like that in the run up to the 3.8 release.

LaughingCabbage:
I think that would be super useful @jmcorgan . It's hard to want to understand and improve code you can't read at a glance. Out of curiosity, is there a good way to do this without obliterating the blame history like @mbr0wn mentioned?

jmcorgan:
Not that I'm aware of, but I wouldn't be surprised if there was an "ignore whitespace changes" option somewhere that would allow that. I'll need to look into it.

LaughingCabbage:
@jmcorgan It seems git blame has this. The -w flag ignores whitespace changes and the -M flag detects line shifts. [Source](https://coderwall.com/p/x8xbnq/git-don-t-blame-people-for-changing-whitespaces-or-moving-code)

mbr0wn:
git can do everything. Excellent!

PR #1373, opened 2017-07-13
gr-dtv: optimize bbscrambler

reisub-de:
Because kbch is always a multiple of 8, we can xor 8 bytes at once in the scrambler by casting to an uint64_t. This gives a small speed improvement.


jmcorgan:
I'd like to see this merged before 3.7.12, but with no response from the submitter, this is looking unlikely.

marcusmueller:
@reisub-de hey, I haven't heard from you, but I'd really love to merge this code, and I guess as the new maintainer, I'm just a bit confused about the current state of whose code is whose, what copyright you assigned to what and how to proceed. If you need to reach out in private, please do so: mmueller `AT` projectname.org

mbr0wn:

PR #1375, opened 2017-07-13
gr-dtv: optimize the bitinterleaver

reisub-de:
Compute a lookup table for the bitinterleaver as the interleaving and twisting positions are the same for each FECframe.


mbr0wn:

PR #1377, opened 2017-07-13
gr-dtv: move init_pilots from work to init

reisub-de:
Instead of recomputing the pilots for each symbol, generate a list in the constructor and just read the pilots from this list


noc0lour:
So, the first commit has my CLA on it, the second I don't know about, probably the work of @reisub-de there

marcusmueller:
First commit is challenge-specific and not an appropriate specialization for upstreaming; so while we have CLA for the first commit, we don't have for the second, and thus, not for the overall PR.

mbr0wn:

mbr0wn:
@reisub-de My bad, I was cleaning up PRs, but we actually did have permission to merge this. I'll combine them into their own new PR for going through BBot again, then we can merge.

Issue #1378, opened 2017-07-13
types.h: get rid of anachronisms now that we can


marcusmueller:
Considering we can now rely on halfway recent MSVC:

we really shouldn't be doing things like

```
typedef long long               gr_int64;
```

these days. There's `cstdint` for a reason, and that is portability. Also, we<sup>TM</sup> should probably check all these types for actual usage – we can afford to break that one OOT that is strange enough to use `gr_int64` instead of `int64_t` in the 3.8 changeover, if we can avoid lunging around completely useless/unused `typedef`s.

This is even worse in `pmt`, where we really need the unambiguity of types like `uint64_t` instead of `unsigned long long` (by the way, the fact that `long long` is standardized is a C++11 extension, so "backwards compat" doesn't 100% fly).

jmcorgan:
I'm all for cstdint cleanliness, even in 3.7 if possible.

mbr0wn:
`cstdint` requires C++11, but `int32_t` and friends do not. We can include `stdint.h`, which is compatible.

jmcorgan:
@marcusmueller Will you take this on?

marcusmueller:
@jmcorgan yes. Please assign.

jmcorgan:
Great, thanks.

jmcorgan:
@mbr0wn For some reason I continue to have a problem with assigning from the menu drop down; I get a list of 10 people but @marcusmueller is not on it.

marcusmueller:
This change leaves gnuradio/types.h a hollow shell, its pale shadow of utility being #include <stdint.h> ; should I clean up inclusion of types.h in the tree and replace with stdint.h where actually used for that? It's included without reason in several places so I'm going to remove these #includes, anyway.

marcusmueller:
Files that `#include <gnuradio/types.h>`, but don't actually mention any `[u]int{width}_t`: 

```
%> ag -L 'int[[:digit:]]*_t' $(ag -l "<gnuradio/types")
gr-filter/lib/hilbert_fc_impl.h
gr-filter/lib/qa_fir_filter_with_buffer.cc
gr-filter/include/gnuradio/filter/hilbert_fc.h
gr-filter/include/gnuradio/filter/pm_remez.h
gnuradio-runtime/lib/misc.h
gnuradio-runtime/include/gnuradio/runtime_types.h
gnuradio-runtime/include/gnuradio/expj.h
gnuradio-runtime/include/gnuradio/message.h
[...]

PR #1380, opened 2017-07-14
Removed superfluous typedefs of gr_?int*; included stdint.h uncond. (#1378)

marcusmueller:
Elsewhere in tree, we never check for #ifdef HAVE_STDINT_H, so this
can't make a difference here. And, we nowhere use any gr_int types, so
we don't need to drag these around.

### DIFF
--- a/gnuradio-runtime/include/gnuradio/types.h
+++ b/gnuradio-runtime/include/gnuradio/types.h
@@ -30,6 +30,8 @@
 
 #include <gnuradio/gr_complex.h>
 
+#include <stdint.h>
+
 typedef std::vector<int>			gr_vector_int;
 typedef std::vector<unsigned int>		gr_vector_uint;
 typedef std::vector<float>			gr_vector_float;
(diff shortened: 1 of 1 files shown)


jmcorgan:
It doesn't record what the problem is, but there is a merge up failure in types.h going from master->next.

marcusmueller:
uuuh, interesting. Will look into that.

jmcorgan:
Ah, on next branch types.h already had a stdint.h include, I'll take care of it when I do the merge up, no worries.

marcusmueller:
Ah, figured it out. b1dac97691bf721857b02b4a998580f4ae3aed82 already did
what I did. So, merge next's types.h to master instead of merging this.

mbr0wn:
`#ifdef HAVE_STDINT_H` is implicitly covered by our choice of supported compilers (at least on next).

gnuradio-runtime/include/gnuradio/types.h
```diff
@@ -30,36 +30,14 @@
...
-#include <stdint.h>
-typedef int16_t			gr_int16;
-typedef int32_t			gr_int32;
-typedef int64_t			gr_int64;
-typedef uint16_t		gr_uint16;
-typedef uint32_t 		gr_uint32;
-typedef uint64_t		gr_uint64;
-#else
-/*
- * Note: these defaults may be wrong on 64-bit systems
```
mbr0wn: I love this comment. It's a perfect comment why this PR is so necessary.

PR #1381, opened 2017-07-14
gr-blocks: add warning to file_source if file will not be fully consumed

LaughingCabbage:
Add a warning to notify the user if the file passed will be fully consumed. Currently, the file source reads in d_itemsize bytes at a time. If the file size % output type size (d_itemsize) is not zero then the final few bytes of the file are never read.

### DIFF
--- a/gr-blocks/lib/file_source_impl.cc
+++ b/gr-blocks/lib/file_source_impl.cc
@@ -110,6 +110,16 @@ namespace gr {
 	throw std::runtime_error("can't open file");
       }
 
+      //Check to ensure the file will be consumed according to item size
+      fseek(d_new_fp, 0, SEEK_END);
+      int file_size = ftell(d_new_fp);
+      rewind (d_new_fp);
+
+      //Warn the user if part of the file will not be consumed.
+      if(file_size % d_itemsize){
+        GR_LOG_WARN(d_logger, "WARNING: File will not be fully consumed with the current output type");
+      }
+
       d_updated = true;
       d_repeat = repeat;
     }


gr-blocks/lib/file_source_impl.cc
```diff
@@ -110,6 +110,16 @@ namespace gr {
 	throw std::runtime_error("can't open file");
       }
 
+			//Check to ensure the file will be consumed according to item size
```
jmcorgan: See the conversation [here](https://stackoverflow.com/questions/5907162/file-size-lookup-in-c) about using fseek/ftell vs. stat for file size. The performance difference here is negligible as its not inside a loop, bit it is a more direct way to accomplish what you want.

gr-blocks/lib/file_source_impl.cc
```diff
@@ -110,6 +110,16 @@ namespace gr {
...
+			fseek(d_new_fp, 0, SEEK_END);
+			int file_size = ftell(d_new_fp);
+			rewind (d_new_fp);
+
+			//Warn the user if part of the file will not be consumed.
+			if(file_size % d_itemsize){
+				printf("WARNING: File will not be fully consumed with current output type of size %d bytes.\n", d_itemsize);
```
jmcorgan: Please see our logging facilities [here](https://gnuradio.org/doc/doxygen/page_logger.html) and use that instead of printf.

jmcorgan: Also, your tab/space settings are off. As shown in the test of the code, we use 2 spaces per scope indent, no tabs.

LaughingCabbage: Great, I've made some changes for you to review.

PR #1382, opened 2017-07-14
fix: fixed a performance bug in buffer::get_tags_in_range.

lexknuther:
buffer::get_tags_in_range called v.resize(0) instead of v.clear() to clear out the input vector.  This causes needless memory allocation of the default value for the second argument to resize.  In applications that make heavy use of tags, this can cause significant performance problems as get_tags_in_range performs an extra allocation and deallocation.

### DIFF
--- a/gnuradio-runtime/lib/buffer.cc
+++ b/gnuradio-runtime/lib/buffer.cc
@@ -346,7 +346,7 @@ namespace gr {
   {
     gr::thread::scoped_lock guard(*mutex());
 
-    v.resize(0);
+    v.clear();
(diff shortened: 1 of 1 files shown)


lexknuther:
I just noticed that there are several other resize(0)s with complex objects.  Should I fix those too?

jmcorgan:
Probably wouldn't hurt, but some of them might not make any difference.

lexknuther:
Any clue why 2 of the checks failed?  The failed in different unrelated tests.

n-west:
Those checks randomly fail sometimes because of race condiions in them. Not related to your change.

lexknuther:
Good, I couldn't see any obvious reason whey they'd fail due to my change.

PR #1384, opened 2017-07-15
gr-dtv: Fix hardcoded parameter in dvbt_reed_solomon_enc

Flamewires:
Removed hardcoded parameter in dtvb_reed_solomon encoder.
The same value is calculated with the defaults in the block,
and this will allow the encoder to be used generically.


jmcorgan:
@drmpeg

jmcorgan:
This has been cherry-picked onto the maint branch as a bug fix. In the future, please submit bugfix as branches off maint, thanks.

PR #1393, opened 2017-07-20
grc: Terminate the process instead of Killing with Terminate button (F7)

kartikp1995:
Currently, GRC terminate button (F7) sends SIGKILL to the ongoing Python program.

This PR is an attempt to send SIGTERM signal in order to allow cleanup before termination. Also, sends SIGKILL signal to the process after 4 seconds if Python program couldn't terminate with SIGTERM.

Since, many hardware takes around 2 seconds to disconnect, 4 seconds should be more than enough for cleaning up.

### DIFF
--- a/grc/gui/NotebookPage.py
+++ b/grc/gui/NotebookPage.py
@@ -20,6 +20,7 @@ Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA
 import pygtk
 pygtk.require('2.0')
 import gtk
+import gobject
 import Actions
 from StateCache import StateCache
 from Constants import MIN_WINDOW_WIDTH, MIN_WINDOW_HEIGHT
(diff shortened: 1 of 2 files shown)


jmcorgan:
Can you confirm the gobject import does add any new package dependencies?

kartikp1995:
@jmcorgan It does not any extra dependency. The `gobject` is already being used in `grc/gui/Executor.py` file.

grc/gui/NotebookPage.py
```diff
@@ -152,6 +153,30 @@ def set_proc(self, process):
...
+            """
+            Kill process if not already terminated
+
+            Called by gobject.timeout_add
+
+            Returns:
+                False to stop timeout_add periodic calls
+            """
+            is_terminated = process.poll()
+            if is_terminated == None:
```
skoslowski: Always use ```is None```

kartikp1995:
@skoslowski 4 seconds was just my guess about general shutdown time of the hardware. I have updated it to 2 seconds. I hope it's not too short!

Also, updated `== None` with `is None`. Thanks for the suggestion.

grc/gui/NotebookPage.py
```diff
@@ -152,6 +153,30 @@ def set_proc(self, process):
         """
         self.process = process
 
+    def term_proc(self):
+        """
+        Terminate the subprocess object
+
+        Add a callback to kill the process
+        after 4 seconds if not already terminated
```
primercuervo: Might not be at all important, but as you changed the time in the code, you might as well change it in the docstring.

kartikp1995: Oh silly me! Thanks! :+1:

PR #1294, opened 2017-04-23
Add new Symbol Synchronizer blocks

awalls-cx18:
Add new symbol synchronizer blocks to gr-digital and their required supporting classes to gr-digital and gr-filter.

These blocks are a superset of the Clock Recovery M&M blocks, MSK Timing Recovery Block, and PFB Clock Synch blocks.

These blocks address many bugs found in those existing GNU Radio blocks.  These blocks also propagate tags correctly and will resync their tracking on time_est and clock_est tags.

The user can select the timing error detector to use.  Decision directed TEDs (like M&M) require the user to specify a slicer constellation (usually digital.constellation_bpsk.base() or digital.constellation_qpsk.base()).

The user can select which interpolating resampler to use.  The "PFB with MF" interpolating resampler requires the user to provide taps for matched filtering.

The user can also set, within reason, an arbitrary number of output samples per symbol (1, 2, 3, 4, 5,...).

-Andy

PR #1394, opened 2017-07-24
digital: Add expected TED gain for computing symbol sync loop gains

awalls-cx18:
The computation of proportional (alpha) and intergal (beta) loop
gains in the symbol synchronizer blocks neglected to account for
the timing error detector gain.  Add expected timing error
detector gain as a user configurable parameter.

### DIFF
--- a/gr-digital/examples/demod/symbol_sync_test_complex.grc
+++ b/gr-digital/examples/demod/symbol_sync_test_complex.grc
@@ -239,7 +239,7 @@
     </param>
     <param>
       <key>_coordinate</key>
-      <value>(2304, 572)</value>
+      <value>(2288, 588)</value>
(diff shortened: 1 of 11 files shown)


awalls-cx18:
Silly build bot failed to even pull the source code. ☹️

jmcorgan:
So retriggering it succeeded. It probably was something internal to Github that didn't finish in time for the incoming pull. Happens.

awalls-cx18:
Would it make sense to add this to the 3.7.12 milestone, since this is an API change to the symbol sync blocks, and that's the milestone that PR #1294 is slated for?

jmcorgan:
Yes, I just hadn't gotten around to it yet.

Issue #1383, opened 2017-07-14
logger.h change causes heap corruption in certain instances


gnieboer:
Commit 08ab03d781135481f15afc1ef0754c60d4ccad42 introduced a heap corruption issue for OOT modules in certain build configurations and certain compilers.

logger.h lines 59-61 cause heap corruption on systems where sizeof(std::string) != sizeof(void *) and logging is not requested by a module.

Compilers affected appear to include GCC 5.0 and MSVC, or in general any compiler that uses short string storage on the stack.

The change in this commit impacts scenarios where GNURadio was built with logging support (but not log4cpp) but the module did not explicitly set ENABLE_GR_LOG flag.  Prior to the commit, it appears ENABLE_GR_LOG would be defined, essentially inheriting from the GNURadio configuration.  Post commit this flag is undefined.

This then results in an issue later in logger.h where logger_ptr is defined as a void* instead of a std::string for the OOT module, a mismatch from the GNURadio library

This then results in a  different definition of gr::block, specifically a difference in sizeof(block), so calls to malloc() and free() from different libraries result in different memory allocations and end up in heap corruption.

A workaround is to explicitly define ENABLE_GR_LOG for all OOT modules.
Two possible fixes I see:
1- Refactor logger.h to keep ENABLE_GR_LOG defined if GNURadio was built with it
2- Refactor logger_ptr and usage thereof to ensure that it always represents something the size of a system pointer so a public api isn't affected by a build flag.

jmcorgan:
Without thinking deeply about it I'd prefer #1. Also, all this conditional code has been removed already and is always "on", with log4cpp now a dependency, on the next branch.

michaelld:
Great sleuthing work! Isn't the actual issue the use of std::string, since the other 2 variants use a pointer -- which will be of the same size no matter which is used? Seems like replacing the std::string version with std::string* or something like that would be the natural way to go to guarantee sizes are the same. This would, of course, be for just maint and master, not next -- since, as JC writes this issue has already been resolved on next.

gnieboer:
I agree with #1 as well for 3.7.x, and leave fix #2 for 3.8.

jmcorgan:
In 3.8 there is no conditional code--logging is always enabled and log4cpp is a required dependency. Doesn't the issue go away then?

gnieboer:
Yes, it sounds like what's already planned implements fix #2 in 3.8.

michaelld:
[...]

PR #1398, opened 2017-07-28
various: fix logging to be great again

michaelld:
[Lots of words below for so few changes! - MLD]

1: change logger_ptr when not ENABLE_GR_LOG to a pointer, to match the type use for the other typedefs. Addresses issue https://github.com/gnuradio/gnuradio/issues/1383 .

2: Fix FindLog4Cpp to (a) allow external setting of LOG4CPP variables, (b) use HINTS with potential provided variables, (c) correctly set cached variables to be what we expect them to be, no matter what, and (d) do LOG4CPP found variable printout only wnen LOG4CPP is enabled.

3: Reduce HAVE_LOG4CPP and LOG4CPP_FOUND down to just the latter, except when CMake has a #cmakedefine to replace.

4: Only use LOG4CPP_INCLUDES and LOG4CPP_LIBRARIES when actually required, and especially not LOG4CPP_INCLUDES in top-level variables such as GNURADIO_RUNTIME_INCLUDE_DIRS, which broken heading finding since already-installed header could be found before project ones.

5: in c++ when using a logger function, use the absolute namespace.

6: a little whitespace here and there.


gnuradio-runtime/CMakeLists.txt
```diff
@@ -124,14 +123,14 @@ install(
     COMPONENT "runtime_runtime"
 )
 
-if(ENABLE_GR_LOG AND HAVE_LOG4CPP)
+if(ENABLE_GR_LOG AND LOG4CPP_FOUND)
 install(FILES
```
mbr0wn: Maybe indent this?

michaelld: It wasn't indented before but it's a short section so I'll include indenting it in the next push I do for this PR. It's not quite "there" yet if you followed the discussion. I'm almost there with overloading CMake's default `include_directories` ... very cool to be able to make a 'smart' version.

jmcorgan:
See compile issue on buildbot:
https://ci.gnuradio.org/buildbot/#/builders/6/builds/154/steps/4/logs/warnings__1469_

jmcorgan:
@michaelld Can you give me an ETA on addressing this?

michaelld:
Per suggestion in https://github.com/gnuradio/gnuradio/issues/1383, I'm going to split off the logger_ptr part of this PR into a new PR. I'll create an overloaded `include_directories` macro & then tweak the use of it. Plenty of small changes, but I think we can fix the include ordering once and for all with this forthcoming change!

jmcorgan:
Let me know the status of this now that we have #1438 and the recent include macro merged.

michaelld:
This PR will require a little re-working now with those fixes in place. I'll redo & push the changes here & ping you when they are ready to go.

jmcorgan:
I'm going to close this pending your rework, which can be done in a new PR off the latest master.

michaelld:
works for me. not enough time in the day!

PR #1399, opened 2017-07-30
grc: fix attr-style enum param lookup in Cheeteh

skoslowski:
This apparently fixes issue @guruofquality was having with enum opts in cheetah templates.

Reported here:  https://github.com/gnuradio/gnuradio/commit/2708f33f4531b33c212c60c3cc1fc3d3a92b5ae1#commitcomment-23327018

I can't seem to reproduce this: current maint branch, same Python version (I guess same Cheetah version too?)
I tried a simple fg with a bunch of null sinks/sources that use any vary this kind of parameter, always get correct string interpolation. 

I don't mind merging this anyway, I am curious though, why this works for me. @guruofquality, do you maybe have an example for me?

### DIFF
--- a/grc/core/Param.py
+++ b/grc/core/Param.py
@@ -135,6 +135,14 @@ class TemplateArg(object):
     def __getitem__(self, item):
         return str(self._param.get_opt(item)) if self._param.is_enum() else NotImplemented
 
+    def __getattr__(self, item):
+        if not self._param.is_enum():
+            raise AttributeError()
+        try:
+            return str(self._param.get_opt(item))
+        except KeyError:
+            raise AttributeError()
+
     def __str__(self):
         return str(self._param.to_code())
 


guruofquality:
@skoslowski Unfortunately I'm the only one seeing it. I was surprised when someone else using the my installer wasn't seeing the issue. And the null sink/source is certainly enough to produce it. My generated top block actually had the template error message embedded into the code. So I think its a combination of python and cheetah version.

The cheetah templates let you use attribute style access for dictionary objects for convenience. So what would have been `type['size']` in regular python can be `type.size` when used within a cheetah template. I'm not sure what cheetah does to implement this: does it check `isinstance(dict)`, does it check for `__getitem__`, does it try attribute access first and fall back to dictionary access when it raises an exception, etc.... I don't know. But if cheetah implemented this differently across versions that might explain why its harder to reproduce it.

My fix was to implement attribute access with `__getattr__` which I think its more strait forward given that we actually want attribute access anyway. But other possible fixes may have included inheriting from UserDict or dict types like it was before the merge because we know that worked. Ideally we pick one to be safe, rather than sorry.

I don't know if it helps, but I can share the versions of python and cheetah:
```
c:\Python27\python.exe -c "import Cheetah; print Cheetah.Version"
2.4.4

c:\Python27\python.exe -c "import sys; print sys.version"
2.7.13 (v2.7.13:a06454b1afa1, Dec 17 2016, 20:53:40) [MSC v.1500 64 bit (AMD64)]
```

skoslowski:
Thanks. I am fine with adding this. @jmcorgan, this is ready to be merged. I can't seem to formally approve my own changes.

PR #1415, opened 2017-08-07
logpwrfft did `ref_scale/2`, which led to integer div in Py2

marcusmueller:
That is especially hurt as the GRC template puts defaults to
`ref_scale=2` (int literal!) and if someone uses `1`, they'll get zero
internally in logpwrfft, and that leads to errors.


jmcorgan:
I cherry-picked this onto the maint branch.

PR #1418, opened 2017-08-08
qt-gui: Inconsistent qt-gui hinting fix

### LINKED ISSUE #1297: qtgui: Inconsistent GUI hints
In my experience the GUI Hints (I'm using QT GUI), seem a bit inconsistent. In particular, the column spans. I found when adding two GUI items and specifying one to have a column span of 1 and the other something different, the GUI arrangement seems to be the reverse of what was expected. An example is included below.

I created a flowgraph to test various GUI hints. I'm using two QT GUI Tab Widgets as an example GUI containers. I added the GUI hint that is used in the tab name. I attached the test flowgraph if anyone else would like to try. [tab_test.grc.zip](https://github.com/gnuradio/gnuradio/files/955807/tab_test.grc.zip)

Things work fine for column spans of 1 and 1
image

If I put a column span of 2 in the first item, the result seems to be the reverse of what's intended.
image

The same is true if I put a span of 2 in the second item.
image

Things seem to work themselves out with larger spans. For instance, column spans of 2 and 3.
image

Or column spans of 3 and 2.
image

Any idea what might be causing this?

BrashEndeavours:
This commit fixes #1297 

`QTGridLayouts` have their rowStretch and columnStretch values initialized to 0 for all grid elements added.
A stretch factor of 0 results in QT not correctly allowing a first row or column to stretch to fill available space equally as other row/column members, when other members are initialized with a row span or column span of 1 grid element.

This leads to inverted stretching, as:
```
|----------|--------------------|
| 0,0,1,2  |      0,2,1,1       |        
|----------|--------------------|
```
Should appear as:
```
|--------------------|----------|
|     0,0,1,2        | 0,2,1,1  |        
|--------------------|----------|
```

This fix parses the gui_hint values and produces code generation to apply `rowStretch()` and `columnStretch()` to all grid elements occupied by the gui element.
qt-gui gui_hint fix

The `gui_hint` syntax of `[tab_id@tab:][row,col,rowSpan,colSpan]` has been made more robust and can now handle just `[row,col]`, as well as providing a default `tab` value in the case of one being omitted.

This PR also addresses collision detection of GUI elements.
qt-gui collision reporting

### DIFF
--- a/gr-qtgui/grc/qtgui_ber_sink_b.xml
+++ b/gr-qtgui/grc/qtgui_ber_sink_b.xml
@@ -47,7 +47,7 @@ for i in xrange($num_curves):
     self.$(id).set_line_alpha(i, alphas[i])
 
 self._$(id)_win = sip.wrapinstance(self.$(id).pyqwidget(), Qt.QWidget)
-$(gui_hint()($win))</make>
+$(gui_hint() % $win)</make>
 
   <param>
     <name>esno</name>
--- a/gr-qtgui/grc/qtgui_check_box.xml
+++ b/gr-qtgui/grc/qtgui_check_box.xml
@@ -20,7 +20,8 @@ self._$(id)_choices_inv = dict((v,k) for k,v in self._$(id)_choices.iteritems())
 self._$(id)_callback = lambda i: Qt.QMetaObject.invokeMethod($(win), "setChecked", Qt.Q_ARG("bool", self._$(id)_choices_inv[i]))
 self._$(id)_callback(self.$id)
 $(win).stateChanged.connect(lambda i: self.set_$(id)(self._$(id)_choices[bool(i)]))
-$(gui_hint()($win))</make>
+$(gui_hint() % $win)</make>
+
 	<callback>self.set_$(id)($value)</callback>
 	<callback>self._$(id)_callback($id)</callback>
 	<param>
--- a/gr-qtgui/grc/qtgui_chooser.xml
+++ b/gr-qtgui/grc/qtgui_chooser.xml
@@ -86,7 +86,8 @@ self._$(id)_callback(self.$id)
 self._$(id)_button_group.buttonClicked[int].connect(
 	lambda i: self.set_$(id)(self._$(id)_options[i]))
 #end if
-$(gui_hint()($win))</make>
+$(gui_hint() % $win)</make>
+
 	<callback>self.set_$(id)($value)</callback>
 	<callback>self._$(id)_callback($id)</callback>
 	<param>
--- a/gr-qtgui/grc/qtgui_const_sink_x.xml
+++ b/gr-qtgui/grc/qtgui_const_sink_x.xml
@@ -51,7 +51,8 @@ for i in xrange(#if $type.t == 'message' then 1 else $nconnections#):
     self.$(id).set_line_alpha(i, alphas[i])
 
 self._$(id)_win = sip.wrapinstance(self.$(id).pyqwidget(), Qt.QWidget)
-$(gui_hint()($win))</make>
+$(gui_hint() % $win)</make>
+
   <callback>set_resize($width, $height)</callback>
   <callback>set_update_time($update_time)</callback>
   <callback>set_title($which, $title)</callback>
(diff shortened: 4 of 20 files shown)


jmcorgan:
Oh, very nice. 

Unfortunately, this amount of code crosses the threshold for needing a copyright license assignment to FSF. Please contact @bhilburn to get the process started.

jmcorgan:
After looking through this, I think this would be better on master than maint. While it is a fix, it's also adding some new features and complexity beyond that. I'll check if this can be manually merged into master, and if so, I'll close this out; otherwise, I'll need you to redo the branch off master.

jmcorgan:
Yeah, that works.

jmcorgan:
I'm happy with this, but there is some conflict the the python3 changes that have already been made in Param.py. Can you get together with @skoslowski to work out how to fixup?

BrashEndeavours:
Will do!

BrashEndeavours:
Changed base to gnuradio:master.
@jmcorgan, how best to communicate changes against gnuradio:python3? There will be significantly different than this. 

Should I just complete another PR for that branch?

BrashEndeavours:
Okay.
I'll work on moving this code into a seperate function, and moving the __call__ logic into the templates.

BrashEndeavours:
Sidebar: Should we look at promoting qt logic out of <param> and make it it's own element, since this project is dropping wx?

skoslowski:
Sorry for the delay. The GRC part look much cleaner now. I think that function doesn't even need to be part of the class. I am not really sure about the QTgui part right now. Its been a while, but I recall this is used e.g. in the Tab Widgets

BrashEndeavours:
I kept it as part of the class, so it can access `self.get_parent()` in python2, and `parent_block, parent_flowgraph` in python3.

I thought that too about the qt tab widgets, but it seems that the text replacement for `addWidget/addLayout` is unnecessary, as it happens now in the `<make>` block. 

Have tested this with every qt widget that comes with GR, on my end. Maybe you can give it a swing and see if it breaks anything of yours?

skoslowski:
Ah, yes, I missed that. Yes, will do.

jmcorgan:
@skoslowski Let me know where things stand and what the merge path is master->next->python3.

jmcorgan:
@skoslowski Now that we've got the other PRs merged up in python3, this last one here has a conflict between next->python3. Can you resolve this?

skoslowski:
Terribly sorry for dropping the ball here.

I finally took a closer look at this and ended up refactoring the code a bit, see
https://github.com/skoslowski/gnuradio/commit/5479831e2b2db452d5c292b84e5910e6cb2bee6b

I moved things around a bit to make it more python3 friendly and speed up the collision check a bit
--> for larger flowgraphs with lots of qt widgets this is quite a lot of checking. 

I think, it would make sense to centralize the hostage cells in the flowgraph object. Every param would add itself to that set or raise if some other block has claimed the cell already.

That would also help with my other (very minor) issue: Every block produces code for two loops setting setRowStretch and setColumnStretch of its cells. If we had a central "cell map" that code could be generated in the main template, with way fewer code lines. Now QT block produce a lot of code as is, so this really doesn't matter too much. I find the extra lines in the tooltip of the param entry box a bit confusing....

@BrashEndeavours sorry again for delaying this review so long. Please let me know what you think. I am generally okay with keep this as it is. In this case I would do the proposed enhancements in the python3 branch only...

BrashEndeavours:
@skoslowski Looks good to me! 
So my commit will be on master, and [yours](https://github.com/skoslowski/gnuradio/commit/88eab8366f49a269b0e34c54e69892824c2c0f30) will be on python3?

skoslowski:
Good. I will merge yours then with my modifications on top.
(further changes from my side like a centralized cell map would be next/python3 only)

marcusmueller:
@skoslowski does this mean this PR is now obsolete, as it's part of py3 or next (and yes, of which branch specifically?)

dkozel:
@marcusmueller I can replicate on current HEAD of master (76f8abeb0). These commits are not present.

marcusmueller:
So, I'd agree with @dkozel, this looks ripe for merge, but @skoslowski's silence actually is a blocker for me here. Calling him once more (also on IRC)! Say but one world, and I shall merge this PR.

marcusmueller:
After quite a bit of interpreting remarks on IRC, and trusting this won't *break* anything (we don't have GUI testing!), I'm merging this.

marcusmueller:
Manually merged into maint.

marcusmueller:
@sbmueller I just pushed #1723 to maint-3.7. Does that help?

PR #1419, opened 2017-08-10
Fix invalid Asio buffer usage in tcp_connection

v1bri:
Asio requires that the underlying buffer passed to `async_write()` remain valid valid until the handler was called. The previous version was allocating a vector on the stack which gets destroyed once the `send()` method returns. Added a test case for the TCP server.

I'm on the fence about the raw pointer usage for `txbuf`. I considered using `boost::shared_ptr` (though that has the overhead of a ref count). I can switch this to a `std::unique_ptr` if Gnuradio can use C++11 features.

I also discovered the interesting behavior wrt PDUs larger than the configured MTU and left a corresponding comment for developers of the future.

### DIFF
--- a/gr-blocks/lib/tcp_connection.cc
+++ b/gr-blocks/lib/tcp_connection.cc
@@ -54,16 +54,25 @@ namespace gr {
     tcp_connection::send(pmt::pmt_t vector)
     {
       size_t len = pmt::blob_length(vector);
+
+      // Asio async_write() requires the buffer to remain valid until the handler is called. It will be
+      // deleted in tcp_connection::handle_write.
+      char* txbuf = new char[len];
+
+      size_t temp = 0;
+      memcpy(txbuf, pmt::uniform_vector_elements(vector, temp), len);
+
       size_t offset = 0;
-      std::vector<char> txbuf(std::min(len, d_buf.size()));
       while (offset < len) {
-        size_t send_len = std::min((len - offset), txbuf.size());
-        memcpy(&txbuf[0], pmt::uniform_vector_elements(vector, offset), send_len);
+        // Limit the size of each write() to the MTU.
+        // FIXME: Note that this has the effect of breaking a large PDU into several smaller PDUs, each
+        // containing <= MTU bytes. Is this the desired behavior?
+        size_t send_len = std::min((len - offset), d_buf.size());
+        boost::asio::async_write(d_socket, boost::asio::buffer(txbuf + offset, send_len),
+          boost::bind(&tcp_connection::handle_write, this, txbuf,
+            boost::asio::placeholders::error,
+            boost::asio::placeholders::bytes_transferred));
         offset += send_len;
-        boost::asio::async_write(d_socket, boost::asio::buffer(txbuf, send_len),
-			       boost::bind(&tcp_connection::handle_write, this,
-					   boost::asio::placeholders::error,
-					   boost::asio::placeholders::bytes_transferred));
       }
     }
 
--- a/gr-blocks/lib/tcp_connection.cc
+++ b/gr-blocks/lib/tcp_connection.cc
@@ -55,12 +55,11 @@ namespace gr {
     {
       size_t len = pmt::blob_length(vector);
 
-      // Asio async_write() requires the buffer to remain valid until the handler is called. It will be
-      // deleted in tcp_connection::handle_write.
-      char* txbuf = new char[len];
+      // Asio async_write() requires the buffer to remain valid until the handler is called.
+      boost::shared_ptr<char[]> txbuf(new char[len]);
 
       size_t temp = 0;
-      memcpy(txbuf, pmt::uniform_vector_elements(vector, temp), len);
+      memcpy(txbuf.get(), pmt::uniform_vector_elements(vector, temp), len);
 
       size_t offset = 0;
       while (offset < len) {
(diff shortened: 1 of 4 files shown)


gr-blocks/lib/tcp_connection.cc
```diff
@@ -54,16 +54,25 @@ namespace gr {
     tcp_connection::send(pmt::pmt_t vector)
     {
       size_t len = pmt::blob_length(vector);
+
+      // Asio async_write() requires the buffer to remain valid until the handler is called. It will be
+      // deleted in tcp_connection::handle_write.
+      char* txbuf = new char[len];
```
jmcorgan: I'm fairly sure this is okay, but not knowing asio internals, it's not clear if there are times that async_write() might be called but without a following handle_write being invoked.

gr-blocks/python/blocks/qa_socket_pdu.py
```diff
@@ -100,6 +100,35 @@ def test_003 (self):
...
+        self.pdu_recv = blocks.socket_pdu("TCP_CLIENT", "localhost", port, mtu)
+        self.pdu_sink = blocks.message_debug()
+
+        self.tb.msg_connect(self.pdu_source, "strobe", self.pdu_send, "pdus")
+        self.tb.msg_connect(self.pdu_recv, "pdus", self.pdu_sink, "store")
+
+        self.tb.start()
+        while self.pdu_sink.num_messages() < 1:
+            time.sleep(0.1)
+        self.tb.stop()
```
jmcorgan: This has a chance to hang if for some reason the message does not arrive.

v1bri: That's fair. I was trying to avoid another 1 second sleep but adding timeout code is even more unwieldy.

gr-blocks/lib/tcp_connection.cc
```diff
@@ -54,16 +54,25 @@ namespace gr {
...
+      memcpy(txbuf, pmt::uniform_vector_elements(vector, temp), len);
+
       size_t offset = 0;
-      std::vector<char> txbuf(std::min(len, d_buf.size()));
       while (offset < len) {
-        size_t send_len = std::min((len - offset), txbuf.size());
-        memcpy(&txbuf[0], pmt::uniform_vector_elements(vector, offset), send_len);
+        // Limit the size of each write() to the MTU.
+        // FIXME: Note that this has the effect of breaking a large PDU into several smaller PDUs, each
+        // containing <= MTU bytes. Is this the desired behavior?
```
v1bri: True however on the receive side we will get 2 or more PDUs where there was once 1 which might lead to unintended behavior. The comment was added to reflect this.

gr-blocks/lib/tcp_connection.cc
v1bri: The handler is invoked if `io_service::run()` is called, which is part of this class' implicit contract since data is received via `async_read_some()`. As mentioned in the comment above, this could also be implemented with `boost::shared_ptr` or `std::unique_ptr` to guarantee destruction.

jmcorgan: I think the shared_ptr route would work. We can't do unique_ptr until 3.8.

gr-blocks/lib/tcp_connection.cc
jmcorgan: Are you talking about GNU Radio PDUs or TCP segments? TCP can always fragment and deliver a stream over multiple calls to recv(), and it is up to the receiver to reassemble any boundaries in the data.

gr-blocks/python/blocks/qa_socket_pdu.py
jmcorgan: I agree; it points to a bigger issue we have with our QA code in general. But I think this is fine.

gr-blocks/lib/tcp_connection.cc
v1bri: Sounds good.

gr-blocks/lib/tcp_connection.cc
v1bri: GNU Radio PDUs. The `tcp_connection` class has an MTU parameter in the constructor that it uses to decide when to fragment PDUs. The corresponding receive code is located in [socket_pdu_impl.cc](https://github.com/gnuradio/gnuradio/blob/master/gr-blocks/lib/socket_pdu_impl.cc#L149) and it doesn't attempt to reassemble the fragmented PDU. Does the FIXME comment need a little clarification?

gr-blocks/lib/tcp_connection.cc
```diff
@@ -55,20 +55,19 @@ namespace gr {
...
+      memcpy(txbuf.get(), pmt::uniform_vector_elements(vector, temp), len);
 
       size_t offset = 0;
       while (offset < len) {
         // Limit the size of each write() to the MTU.
         // FIXME: Note that this has the effect of breaking a large PDU into several smaller PDUs, each
         // containing <= MTU bytes. Is this the desired behavior?
         size_t send_len = std::min((len - offset), d_buf.size());
-        boost::asio::async_write(d_socket, boost::asio::buffer(txbuf + offset, send_len),
+        boost::asio::async_write(d_socket, boost::asio::buffer(txbuf.get() + offset, send_len),
```
jmcorgan: I don't think this will work the way you expect; you are passing a pointer to the internal contents of a locally created shared pointer, so the actual shared pointer goes out of scope at the end of the function, and the internal raw buffer gets deleted. To take advantage of the shared pointer semantics, you need to pass the shared pointer itself to the async_write call, which will increment its reference count, then when the local 'txbuf' goes out of scope the shared pointer itself remains alive.

v1bri: You're correct we need to make sure the shared pointer stays alive until the async_write handler is called. So I make a copy of the shared pointer with `boost::bind` when creating the handler. The shared pointer's refcount will not go to zero until the write handler returns.

jmcorgan: Ah, got it.

PR #1421, opened 2017-08-12
gr-digital: Correlate access code callbacks

mrjacobagilbert:
Added callbacks to the correlate_access_code_tag_xx block and fixed a minor bug with the block tree for gr_digital related to the same block (it had the earlier _bb suffix in the block tree XML).

### DIFF
--- a/gr-digital/grc/digital_correlate_access_code_tag_xx.xml
+++ b/gr-digital/grc/digital_correlate_access_code_tag_xx.xml
@@ -9,6 +9,10 @@
 	<key>digital_correlate_access_code_tag_xx</key>
 	<import>from gnuradio import digital</import>
 	<make>digital.correlate_access_code_tag_$(type.fcn)($access_code, $threshold, $tagname)</make>
+	<callback>set_access_code($access_code)</callback>
+	<callback>set_threshold($threshold)</callback>
+	<callback>set_tagname($tagname)</callback>
+
 	<param>
 		<name>IO Type</name>
 		<key>type</key>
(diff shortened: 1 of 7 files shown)


jmcorgan:
A few things. Can you break this into one commit that fixes the _bb -> _xx issue, another that adds the callbacks, and a third that updates the work function with the temp_access_code logic? 

I'm not sure I understand the rational for this last, by the way--can you elaborate on why?

Finally, our normal practice is to preface commits messages with the top-level component affected; in this case, it would be "gr-digital: fixed...."

Thanks.

mrjacobagilbert:
Sure, I can do that. Forgot about the commit syntax; I'll update the other PR as well.

Not sure the last part is actually necessary; I don't know enough about all the things that can go wrong so this seemed safest and low impact. It seemed potentially problematic to have d_access_code changing iteratively in a loop, on the off chance something accessed this value in the middle of a change. Please let me know if this is not necessary and I will omit the last commit.

jmcorgan:
Ah, I see what you were trying to do. Yes, your instinct is correct, but we have a "proper" way of solving that at least for now, which is to acquire a mutex in the setter function and in the work function, to ensure exclusive access. There is a member variable 'mutex' defined in basic_block.h for this purpose.

mrjacobagilbert:
Ok sounds good, I'll attempt to emulate another example of this...

I assume this should be rolled into the same commit as the callback update?

jmcorgan:
Yes. I'm being a little pedantic; I'm going to cherry-pick the _bb fix onto maint and put the callbacks/mutex on master, but am not cruel enough to make you do two PRs :-)

mrjacobagilbert:
I'll remove the _bb fix from this commit so you can merge this PR directly.

jmcorgan:
Go ahead.

mrjacobagilbert:
I believe the new commit is working correctly.

PR #1423, opened 2017-08-14
cmake: switch cascaded elses to elseifs

guruofquality:
After updating to cmake 3.9.1, these particular set of elses caused an error. Perhaps cmake was permissive about this in prior versions, but the code should really read elseif on subsequent cases regardless.

* Fixes configuration  error on cmake 3.9.1


mrjacobagilbert:
This PR was made against master as I figured this would be considered a new feature... Let me know if you want me to resubmit against maint.

PR #1424, opened 2017-08-14
cmake: Fix for CppUnit and C++11

michaelld:
If CppUnit version >= 1.14.0, try to switch to C++11. Otherwise, use C++98 as before. Requires that FindCppUnit return the CppUnit version if found.

This PR is on maint, and the top-level CMakeLists.txt on master and next are somewhat different .. so do you want me to a separate PR for those?


jmcorgan:
The easiest way to deal with the merge conflicts would be to create a branch off master, _then merge this branch into it_, with fixups as needed, then make a PR for master. If that then will have merge conflicts with next, do the same--make a new branch off next, merge the PR for master into it, then add another PR for next. I know it is extra work, but that way we can have clean PR merges on each branch and maintain continuity of maint->master->next.  Thanks.

michaelld:
OK. I think I fixed up #1425 for maint -> master. I'm working on -> next now.

michaelld:
-> next is in #1429

noc0lour:
For future reference this breaks on older GCC because of this: https://stackoverflow.com/questions/35090479/why-does-gcc-std-c99-treat-the-keyword-asm-as-a-function 

But this won't be an issue in future versions anymore

michaelld:
frickin' gcc ... thinks it knows best! well ppppppppllllllll to it!

PR #1425, opened 2017-08-14
cmake: Fix for CppUnit and C++11 on master

michaelld:
If CppUnit version >= 1.14.0, try to switch to C++11. Otherwise, use C++98 as before. Requires that FindCppUnit return the CppUnit version if found.

Same same as https://github.com/gnuradio/gnuradio/pull/1424 except on master instead of maint.


jmcorgan:
See my comments on #1424. You can force push the master branch created that way to this PR.

michaelld:
OK. I think this PR is correct for what you wanted: maint -> master, with fixups after the merge.

PR #1427, opened 2017-08-16
gr-qtgui: time_sink: adds nconnections check to grc and fixes #1317

### LINKED ISSUE #1317: qtgui_time_sink can't display more than 12 inputs

777arc:
now fixes #1317 and adds a check to grc so that when a user enters more than 5 (complex) or 10 (float) inputs, an error occurs letting them know what is wrong.  The reason for not allowing 24 inputs when using grc is pretty clear if you look at how qtgui_time_sink_x.xml is written, the whole thing should probably be rewritten using for loops.

on a side note, once a PR is closed, is there a way to open it again and add a commit?  I created this new PR because I wasn't sure how to do that or if it was possible.  The old PR was #1321

### DIFF
--- a/gr-qtgui/grc/qtgui_time_sink_x.xml
+++ b/gr-qtgui/grc/qtgui_time_sink_x.xml
@@ -931,6 +931,8 @@ $(gui_hint()($win))</make>
     <hide>#if ((int($nconnections()) >= 10 or ($type() == "complex" and int($nconnections()) >= 5)) and not $type.t == "message") then 'part' else 'all'#</hide>
   </param>
 
+  <check>$nconnections &lt;= #if $type() == "complex" then 5 else 10#</check>
+
   <sink>
     <name>in</name>
     <type>$type.t</type>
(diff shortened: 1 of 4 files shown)


jmcorgan:
I've squashed merged this to a single commit for the maint branch, so it won't automatically show up here as closed; I'll close it manually once this is pushed.

PR #1432, opened 2017-08-18
renamed fractional_{resampler|interpolator} to mmse_ (#1428)

### LINKED ISSUE #1428: fractional_resampler is a misnomer, should be renamed in next
The blocks that we call `fractional_resampler_{cc|ff}` really shouldn't be called "fractional", as it suggests a rational in/output rate ratio. That's explicitly not the case with these. We have `rational_resampler_` for that!

This has caused a lot of confusion among students, and newbies, in general. I'm not excluding myself here.

It's clear how that name "happened" – we have a (slightly confusingly named, but still in range of sensibility) `fractional_interpolator`. The `fractional_resampler` is basically the same thing.

I propose: we call that `mmse_arb_resampler`, to use the same naming logic than for `pfb_arb_resampler` and list them as "MMSE arbitrary resampler" in GRC.

Renaming the `fractional_interpolator` to `mmse_interpolator` would make sense, too.

marcusmueller:
fixes #1428


awalls-cx18:
Note that mmse_interpolator_xx is easily confused with mmse_fir_interpolator_xx.  You've traded possible end user confusion for possible developer confusion.  Probably an acceptable trade, though.

PR #1434, opened 2017-08-20
Modifications in GRC to allow BokehGUI

kartikp1995:
This includes all modifications required in GRC to allow BokehGUI.

Major changes:
Added BokehGUI `main()` in `flow_graph.tmpl`
Added an option for BokehGUI in Options block

### DIFF
--- a/grc/core/generator/flow_graph.tmpl
+++ b/grc/core/generator/flow_graph.tmpl
@@ -38,7 +38,7 @@ import threading
 
 ## Call XInitThreads as the _very_ first thing.
 ## After some Qt import, it's too late
-#if $generate_options in ('wx_gui', 'qt_gui')
+#if $generate_options in ('wx_gui', 'qt_gui', 'bokeh_gui')
 if __name__ == '__main__':
     import ctypes
     import sys
@@ -119,6 +119,15 @@ class $(class_name)(gr.top_block, Qt.QWidget):
 
         self.settings = Qt.QSettings("GNU Radio", "$class_name")
         self.restoreGeometry(self.settings.value("geometry").toByteArray())
+
+#elif $generate_options == 'bokeh_gui'
+
+class $(class_name)(gr.top_block):
+    def __init__(self, doc):
+        gr.top_block.__init__(self, "$title")
+        self.doc = doc
+        self.plot_lst = []
+        self.widget_lst = []
 #elif $generate_options == 'no_gui'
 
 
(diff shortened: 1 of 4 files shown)


grc/core/generator/flow_graph.tmpl
```diff
@@ -378,6 +403,37 @@ def main(top_block_cls=$(class_name), options=None):
...
+            tb.start()
+            signal.signal(signal.SIGTERM, functools.partial(killProc, tb=tb))
+            session.loop_until_closed()
+        finally:
+            print "Exiting the simulation. Stopping Bokeh Server"
+            tb.stop()
+            tb.wait()
+    finally:
+        serverProc.terminate()
+        serverProc.kill()
```
skoslowski: Is it possible to query if
- a server instance is already running (before creating our own - possibly failing due to the tcp port being used already)
- this is the only document being served - if not the server instance should probably live on...

grc/core/utils/is_bokeh_gui_installed.py
```diff
@@ -0,0 +1,28 @@
+# Copyright 2008-2015 Free Software Foundation, Inc.
```
skoslowski: fix year

grc/core/generator/flow_graph.tmpl
```diff
@@ -378,6 +403,37 @@ def main(top_block_cls=$(class_name), options=None):
...
+        #else
+        doc.title = "$title"
+        #end if
+        session = push_session(doc, session_id="$flow_graph.get_option('id')",
+                               url = "http://localhost:5006/bokehgui")
+        \# Create Top Block instance
+        tb = top_block_cls(doc)
+        try:
+            tb.start()
+            signal.signal(signal.SIGTERM, functools.partial(killProc, tb=tb))
```
skoslowski: Not sure if I missed the import of functools?

grc/core/generator/flow_graph.tmpl
kartikp1995: So, I am not sure what exactly are you suggesting. Is it that instead of the new server, we can connect the new process in existing server?

As of now: Every time the server starts, it checks if the particular port is being used. If yes, it checks next one. It creates new server instance if any port in the sequence is free! And for each server instance, a unique Document is being served.

grc/core/generator/flow_graph.tmpl
kartikp1995: There is `import functools` in `options.xml`

skoslowski: ah okay

jmcorgan:
@skoslowski This will have conflicts on next and python3 that I'd like to to resolve. Can you prepare a merge for next and one for python3, that includes both this and #1418? That would let me merge both this PR and #1418, then merge your two fixup branches.

skoslowski:
My python3 branch https://github.com/skoslowski/gnuradio/commits/python3 not contains commits that merge this into master, from there into next and finally into python3. 
The last merge contains significant changes as the template engine as well as some internal APIs haved changed in the python3 branch. @kartikp1995 please review my changes and make sure I didn't miss anything on the way...

kartikp1995:
@skoslowski Reviewed and verified. Thanks!

PR #1435, opened 2017-08-21
Fix typos from switch to Python 3

hauk142:
Removed eval() from eval(input()).

### DIFF
--- a/docs/exploring-gnuradio/dial_tone.py
+++ b/docs/exploring-gnuradio/dial_tone.py
@@ -41,5 +41,5 @@ def build_graph():
 if __name__ == '__main__':
     tb = build_graph()
     tb.start()
-    eval(input('Press Enter to quit: '))
+    input('Press Enter to quit: ')
     tb.stop()
--- a/docs/exploring-gnuradio/fm_demod.py
+++ b/docs/exploring-gnuradio/fm_demod.py
@@ -82,7 +82,7 @@ class build_graph(gr.top_block):
 def main(args):
     tb = build_graph()
     tb.start()        # fork thread and return
-    eval(input('Press Enter to quit: '))
+    input('Press Enter to quit: ')
     tb.stop()
 
 if __name__ == '__main__':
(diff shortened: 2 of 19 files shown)


docs/exploring-gnuradio/dial_tone.py
```diff
@@ -41,5 +41,5 @@ def build_graph():
 if __name__ == '__main__':
     tb = build_graph()
     tb.start()
-    eval(input('Press Enter to quit: '))
+    input('Press Enter to quit: ')
```
mbr0wn: In Python 2, this doesn't work anymore.

mbr0wn: Unless, of course, if you `from builtins import input`. In that case though, you added 'future' as a dependency for GNU Radio on Py2k.

hauk142:
Yes of course, you're absolutely right. :)

Could do something like this:
`try: input = raw_input
except NameError: pass`

mbr0wn:
...but then it's a nop on Py2k :)

docs/exploring-gnuradio/dial_tone.py
jmcorgan: I think this is okay for 3.8.

jmcorgan:
I'm ok with having this left as 'input()' but the future module import needs to be added for py2 operation.

hauk142:
Could also use input() from six

jmcorgan:
That's an even better idea.

mbr0wn:
Note that `six` is also a new dependency (if it's not already one). It's not part of the vanilla Python libs.

hauk142:
`six` is already used in several modules :)

mbr0wn:
Argh, that's even worse, it means we have an untracked dependency. @jmcorgan do you know what to do to add Python dependencies? `six` needs tracking.

jmcorgan:
I thought it already was. The entire python3 port depends on it.

skoslowski:
six is listed in the "Release 3.8 Dependency Analysis" document

mbr0wn:
OK good, false alarm them. I tried figuring out which CMake build step checks for six, couldn't find it with in a few seconds.

PR #1447, opened 2017-08-27
cmake: add CMake Overloads file and use it at the top level

michaelld:
[This PR is clean for me on maint, master, and next. It needs to be
tested thoroughly, as it overloads a critical CMake function! - MLD]

For now, the CMakeOverloads file just overloads the CMake function
"include_directories", to make it a little smarter. When parsing the
provided directories, the overloaded macro now checks to see whether
each directory is in the SOURCE or BUILD, or neither. If so, then the
directory is added BEFORE others, and if not, then AFTER. The keywords
BEFORE and AFTER are ignored. In this manner, all internal directories
will come before external directories, which allows builds to work
even if a prior version of GNU Radio is already installed into the
CMAKE_INSTALL_PREFIX (e.g., as is often found when building from
source, such as in MacPorts).

By adding this macro at the top-level CMakeLists.txt file, it
overloads all uses of "include_directories" in any subdirectory.

As noted in the file: Moving all include directories to either BEFORE
(internal to source or build) or AFTER (external to source or build)
will work in general. The primary time it could fail is when include
ordering is required to find a specific version of a header when
multiple of the same name are available in the various include
directories. This situation would be poor header naming practice, and
is unlikely to occur in real life. As it does not occur in GNU Radio,
so we ignore it.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -118,6 +118,12 @@ set(EXPORT_FILE ${CMAKE_BINARY_DIR}/ImportExecutables.cmake)
 file(WRITE ${EXPORT_FILE}) #blank the file (subdirs will append)
 
 ########################################################################
+# Incorporate CMake function/macros overloading.
+########################################################################
+
+include(CMakeOverloads)
+
+########################################################################
 # Compiler specific setup
 ########################################################################
 include(GrMiscUtils) #compiler flag check
--- a/cmake/Modules/CMakeOverloads.cmake
+++ b/cmake/Modules/CMakeOverloads.cmake
@@ -1,0 +1,70 @@
+# Copyright 2017 Free Software Foundation, Inc.
+#
+# This file is part of GNU Radio
+#
+# GNU Radio is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 3, or (at your option)
+# any later version.
+#
+# GNU Radio is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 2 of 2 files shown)


jmcorgan:
Looks like there is a version problem with CMake on 14.04.

marcusmueller:
Should something analogous happen in the OOT module template?

marcusmueller:
confirming successful build and beautiful tests on CMake 3.9.0/Fedora26

michaelld:
So CMake isn't liking the "continue" command. Hmmm .... I'll test and rewrite to deal with this.

michaelld:
"continue()" was introduced in CMake 3.2, on March 11, 2015. I can't see from the build log what version of CMake is on the Ubuntu_14_04_64_0 bot, but the top-level CMakeLists.txt file states that the minimum CMake version is 2.6.0. So ... looks like I need to rewrite the statements to be compatible with older CMake.

michaelld:
it'll also be interesting to see if `_include_directories` works on the older CMake ...

michaelld:
@marcusmueller : OOT module template: probably yes. Let's get this working first.

michaelld:
@marcusmueller nice! -> "preexisting prefix that I intentionally poisoned with empty headers" && Thanks for the feedback!

jmcorgan:
Thanks for the quick update, buildbot is very happy.

michaelld:
Excellent! I still wish the build logs had the option of being VERBOSE=on ... so that we can see the actual commands being issued ... but since I can't I'll just believe that what I get locally is the same (since CMake does work pretty consistently across OSs at least in this regard).

PR #1448, opened 2017-08-28
Add non-interned string PMT type 'pmt_string'

samfpetersen:
For my application, I use many symbol PMTs (pmt::string_to_symbol, pmt::intern) that are decoded from stream data. These strings are long and unique and they are created frequently, so over time the PMT symbol hash table fills up - old symbols cannot be removed. This causes increasing hash table access times (iterating long linked lists of collisions) and unbounded memory usage growth. 

In this case, it makes more sense to just work with the string itself rather than dealing with string interning. My solution is to add a non-interned PMT string type and appropriate functions to the PMT API - not interfering with the existing pmt_symbol API at all. 

If there are enough applications that could benefit from this functionality, I think it should be added. 

### Usage
C++:
```
// Convert string to non-interned string PMT (2nd arg indicates whether to intern - default is false)
pmt::pmt_t foo_pmt = pmt::from_string("foo", false);

// Convert non-interned string PMT to string
std::string foo_str = pmt::to_string(foo_pmt);
```
Python:
```
# Convert string to non-interned string PMT
# Use string subclass 'str_noninterned' to indicate non-interned string
foo_str = pmt.str_noninterned("foo")
foo_pmt = pmt.to_pmt(foo_str)

# Convert non-interned string PMT to string
foo_str2 = pmt.to_python(foo_pmt)
```

### DIFF
--- a/gnuradio-runtime/include/pmt/pmt.h
+++ b/gnuradio-runtime/include/pmt/pmt.h
@@ -154,6 +154,26 @@ PMT_API pmt_t intern(const std::string &s);
  */
 PMT_API const std::string symbol_to_string(const pmt_t& sym);
 
+
+/*
+ * ------------------------------------------------------------------------
+ *                 Non-interned strings
+ * ------------------------------------------------------------------------
+ */
+
+//! Return true if obj is a non-interned string PMT, else false.
+PMT_API bool is_string(const pmt_t& p);
+
+/*!
+ * Return an interned or non-interned string PMT containing the given string.
+ */
+PMT_API pmt_t from_string(std::string str, bool interned=false);
+
+/*!
+ * If \p is a string PMT (interned or non-interned), return the contained string
+ */
+PMT_API std::string to_string(const pmt_t& p);
+
 /*
  * ------------------------------------------------------------------------
  *           Numbers: we support integer, real and complex
--- a/gnuradio-runtime/include/pmt/pmt.h
+++ b/gnuradio-runtime/include/pmt/pmt.h
@@ -161,18 +161,19 @@ PMT_API const std::string symbol_to_string(const pmt_t& sym);
  * ------------------------------------------------------------------------
  */
 
-//! Return true if obj is a non-interned string PMT, else false.
+//! Return true if obj is a string PMT, else false.
 PMT_API bool is_string(const pmt_t& p);
 
 /*!
- * Return an interned or non-interned string PMT containing the given string.
+ * Return a string PMT containing the given string - if the string is already in the
+ * symbol table, the PMT will be interned, otherwise it will be non-interned.
  */
-PMT_API pmt_t from_string(std::string str, bool interned=false);
+PMT_API pmt_t from_string(const std::string &str);
 
 /*!
  * If \p is a string PMT (interned or non-interned), return the contained string
  */
-PMT_API std::string to_string(const pmt_t& p);
+PMT_API const std::string to_string(const pmt_t& p);
 
 /*
  * ------------------------------------------------------------------------
--- a/gnuradio-runtime/include/pmt/pmt.h
+++ b/gnuradio-runtime/include/pmt/pmt.h
@@ -24,6 +24,7 @@
 #define INCLUDED_PMT_H
 
 #include <pmt/api.h>
+#include <gnuradio/attributes.h>
 #include <boost/intrusive_ptr.hpp>
 #include <boost/shared_ptr.hpp>
 #include <boost/any.hpp>
@@ -139,6 +140,7 @@ PMT_API bool to_bool(pmt_t val);
  */
 
 //! Return true if obj is a symbol, else false.
+__GR_ATTR_DEPRECATED
 PMT_API bool is_symbol(const pmt_t& obj);
 
 //! Return the symbol whose name is \p s.
--- a/gnuradio-runtime/lib/basic_block.cc
+++ b/gnuradio-runtime/lib/basic_block.cc
@@ -92,7 +92,7 @@ namespace gr {
   void
   basic_block::message_port_register_in(pmt::pmt_t port_id)
   {
-    if(!pmt::is_symbol(port_id)) {
+    if(!pmt::is_string(port_id)) {
       throw std::runtime_error("message_port_register_in: bad port id");
     }
     msg_queue[port_id] = msg_queue_t();
@@ -115,7 +115,7 @@ namespace gr {
   void
   basic_block::message_port_register_out(pmt::pmt_t port_id)
   {
-    if(!pmt::is_symbol(port_id)) {
+    if(!pmt::is_string(port_id)) {
       throw std::runtime_error("message_port_register_out: bad port id");
     }
     if(pmt::dict_has_key(d_message_subscribers, port_id)) {
--- a/gnuradio-runtime/lib/block_detail.cc
+++ b/gnuradio-runtime/lib/block_detail.cc
@@ -192,7 +192,7 @@ namespace gr {
   void
   block_detail::add_item_tag(unsigned int which_output, const tag_t &tag)
   {
-    if(!pmt::is_symbol(tag.key)) {
+    if(!pmt::is_string(tag.key)) {
       throw pmt::wrong_type("block_detail::add_item_tag key", tag.key);
     }
     else {
@@ -204,7 +204,7 @@ namespace gr {
   void
   block_detail::remove_item_tag(unsigned int which_input, const tag_t &tag, long id)
   {
-    if(!pmt::is_symbol(tag.key)) {
+    if(!pmt::is_string(tag.key)) {
       throw pmt::wrong_type("block_detail::add_item_tag key", tag.key);
     }
     else {
--- a/gnuradio-runtime/lib/controlport/thrift/rpcpmtconverters_thrift.cc
+++ b/gnuradio-runtime/lib/controlport/thrift/rpcpmtconverters_thrift.cc
@@ -34,7 +34,7 @@ rpcpmtconverter::from_pmt(const pmt::pmt_t& knob)
     result.value.__set_a_double(pmt::to_double(knob));
     return result;
   }
-  else if(pmt::is_symbol(knob)) {
+  else if(pmt::is_string(knob)) {
     std::string value = pmt::symbol_to_string(knob);
     GNURadio::Knob result;
     result.type = GNURadio::BaseTypes::STRING;
--- a/gnuradio-runtime/lib/hier_block2.cc
+++ b/gnuradio-runtime/lib/hier_block2.cc
@@ -91,7 +91,7 @@ namespace gr {
   hier_block2::msg_connect(basic_block_sptr src, pmt::pmt_t srcport,
                            basic_block_sptr dst, pmt::pmt_t dstport)
   {
-    if(!pmt::is_symbol(srcport)) {
+    if(!pmt::is_string(srcport)) {
       throw std::runtime_error("bad port id");
     }
     d_detail->msg_connect(src, srcport, dst, dstport);
@@ -108,7 +108,7 @@ namespace gr {
   hier_block2::msg_disconnect(basic_block_sptr src, pmt::pmt_t srcport,
                               basic_block_sptr dst, pmt::pmt_t dstport)
   {
-    if(!pmt::is_symbol(srcport)) {
+    if(!pmt::is_string(srcport)) {
       throw std::runtime_error("bad port id");
     }
     d_detail->msg_disconnect(src, srcport, dst, dstport);
--- a/gnuradio-runtime/lib/pmt/pmt-serial-tags.scm
+++ b/gnuradio-runtime/lib/pmt/pmt-serial-tags.scm
@@ -36,6 +36,8 @@
 (define pst-uint64	#x0b)
 (define pst-tuple	#x0c)
 
+(define pst-string        #x0d)   ; untagged-int16 n; followed by n bytes of string value
+
 ;; u8, s8, u16, s16, u32, s32, u64, s64, f32, f64, c32, c64
 ;;
 ;;   untagged-uint8  tag
--- a/gnuradio-runtime/lib/pmt/pmt.cc
+++ b/gnuradio-runtime/lib/pmt/pmt.cc
@@ -285,6 +285,12 @@ hash_string(const std::string &s)
 bool
 is_symbol(const pmt_t& p)
 {
+  return p->is_string();
+}
+
+bool
+is_interned_string(const pmt_t& p)
+{
   return p->is_string() && _string(p)->is_interned();
 }
 
--- a/gnuradio-runtime/lib/pmt/pmt.cc
+++ b/gnuradio-runtime/lib/pmt/pmt.cc
@@ -122,6 +122,12 @@ _symbol(pmt_t x)
   return dynamic_cast<pmt_symbol*>(x.get());
 }
 
+static pmt_string *
+_string(pmt_t x)
+{
+  return dynamic_cast<pmt_string*>(x.get());
+}
+
 static pmt_integer *
 _integer(pmt_t x)
 {
@@ -332,7 +338,40 @@ symbol_to_string(const pmt_t& sym)
   return _symbol(sym)->name();
 }
 
+////////////////////////////////////////////////////////////////////////////
+//                             Non-interned strings
+////////////////////////////////////////////////////////////////////////////
 
+pmt_string::pmt_string(std::string value) : d_value(value) {}
+
+bool
+is_string(const pmt_t& p)
+{
+  return p->is_string();
+}
+
+
+pmt_t
+from_string(std::string str, bool interned)
+{
+  if (interned) {
+    return string_to_symbol(str);
+  } else {
+    return pmt_t(new pmt_string(str));
+  }
+}
+
+std::string
+to_string(const pmt_t& p)
+{
+  if (p->is_symbol()) {
+    return symbol_to_string(p);
+  } else if (p->is_string()) {
+    return _string(p)->value();
+  }
+
+  throw wrong_type("pmt_to_string", p);
+}
 
 ////////////////////////////////////////////////////////////////////////////
 //                             Number
@@ -1080,6 +1119,9 @@ eqv(const pmt_t& x, const pmt_t& y)
   if (x == y)
     return true;
 
+  if (x->is_string() && y->is_string())
+    return _string(x)->value() == _string(y)->value();
+
   if (x->is_integer() && y->is_integer())
     return _integer(x)->value() == _integer(y)->value();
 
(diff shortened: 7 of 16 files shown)


marcusmueller:
Great stuff!

Nearly agreeing with @bastibl here, I've got a few observations from the PMT code, however:

# Overly long version

* I believe PMTs were invented and hand-optimized very much with performance in mind, but we might need to reevaluate that with modern compilers and machines.
  * Although a lot of rather intense design deviations from the rest of GR were made, no benchmarks of why these were made have been passed down (that I knew of)
* As far as I can tell, the purpose of having interning PMTs is that you then get *unique* objects with fixed addresses which 
  * saves memory (that might not even be true for small strings), 
  * avoids calculating a content hash to find the right port in a `map` or so if you re-use the symbol
* We have many places in the source code where things like `if(pmt::dict_has_key(msg, pmt::intern("resamp_ratio")))` call the symbol constructor in-place.
  * Presumable even far more common OOT.
  * That defeats half of the purpose of having a type that would avoids repeated strcmp or hashing, if you first need to have a hash or a string comp to see whether that string has been interned before.
  * not a surprise – symbol do's and dont's were never prominently hammered into user's heads.
    * not a surprise either – doing it right by PMT is *not* pretty at all.
  * the performance of the `pmt::symbol` constructor is hence critical in practice, if internalizing had any advantage in the first place
* The fact that we're locking the hash table for interned strings on creation of new ones is actually a bit of a problem, as it requires double-iterating before and after locking, should an existing incarnation of the symbol not have been found
  * the mutex is only protecting the table from being simultaneously modified. It does **not** protect against iterating through the table while a new symbol is added. The table is a std::vector of fixed size.
* The table is a std::vector of fixed size to hold linked lists of pmt::symbols that have the same hash.
  * std::vector for fixed size is inefficient, but who cares.
  * instead of a clever tree structure that'd allow us to bound the lookup time, we use a linked list
  * The hash table size is 701, hardwired, without any reasoning. The hashing function is undocumented. "Obscure" is usually not a design goal of core components.
  * hash collisions are expensive here; far more than let's say having a hash table of a few kBs more, is my feeling. 701 seems oddly small. See: birthday paradox below 
why a 701 hashtable is bad The probability of a collision crosses 50% after only 30 different internized strings, and 99% after 79%.
    * And that is *assuming* that  the hashes are equally distributed. Which they don't really seem to be: 
distribution of hashes for english words This figure was made based on nearly half a million different English strings. The variance should definitely not be 4%.
* Interned Strings are not transportable as such, and thus, it's a bad idea that a PMT library has them if transportability is going to be a core feature in the future.

<hr>

# Overly short version

* Benefit `pmt::intern`: Speed when looking up (pointer comparison)
* Cost of `pmt::intern`: hashing + table storage on creation
* Rule "never create repeatedly" is not well-known
* Real-world-usage benefit hence questionable since not observed in the wild
* PMT impl not 100% thread safe
* hash table in `pmt::intern` far too small, collisions non-negligible already at small numbers
* hash function suboptimal
* hash table contains suboptimal linked lists

<hr>

# Conclusion

I'm very happily going to review this PR in depth. I'd recommend that we compare the CPU consumption  of the interning code and a boring `std::map<std::string, pmt_t>`.

# Consequences of this for the whole project

* if above benchmark doesn't yield significant advantages for the usage `pmt::intern`, replace by `map` impl on master
* for next: use PMT-heavy flow graph, completely remove the internalization (i.e. alias `pmt::symbol` to `pmt::string`), actually measure time spent in PMT functionality (`perf` will be sufficient for this). @osh has a lot of message-passing-intense FGs, he probably has a good idea of what a good measurement parcour would be

<hr>

# Must-haves for **this** pull request

Luckily, @samfpetersen doesn't have to implement any of the above before we can make use of his feature. However, as @bastibl pointed out, people *will* want to use `pmt::string`, but I don't see the downside of that, *iff* we can avoid that this causes errors when users pass strings (it would currently do). I don't see any downsides of internally converting to `pmt::symbol` in general (am I missing somethin, @jmcorgan, @bastibl ?)

* Make `pmt::string` the default thing to do return doing `pmt::mp(std::string)`. Clean up GNU Radio tree itself, and use `pmt::symbol` explicit where `mp` is used now to name ports and
* add code to `basic_block`'s port registering functions to convert pmt::strings to pmt::interns internally
* add code to the `stream_tag` type to automatically cast `pmt::string` to `pmt::intern` for keys
* implement `operator==` or `eqv` or whatever that thing is called that compares two PMTs so that strings and symbols of the same content are equal.

marcusmueller:
Code for graphs above: https://github.com/marcusmueller/birthday_paradox

marcusmueller:
Oh, and by the way, if we have a string type that is not just a wrapper for null-terminate char[] in `pmt::uvector8`, let's define these are *actually* strings, as in: strings are to be UTF8-encoded. That'd save us ambiguity later.

bastibl:
So, AFAIS, the options are:

- leave everything as is and recommend to use blobs for that use case
- stick with interned symbols and increase the hash table size
- stick with interned symbols and use a different data structure (maybe something that adapts its size dynamically once it reaches a certain utilization).
- extend symbols to allow non-interned strings

Even in the last case, I wouldn't change the defaults (which would require adapting core components).

marcusmueller:
> leave everything as is and recommend to use blobs for that use case

I'd say that we also should be stressing that interning things that aren't actually symbols is bad, somewhere in documentation. But I don't know where.

> stick with interned symbols and increase the hash table size

That won't work without restrictions. The modulo 701 seems to be part of the hashing, actually, even if it's not in the actual hash function (why?), because if I increase 701 to let's say 1024, it becomes obvious that the hashing function isn't uniformly distributed (caveat: at least for English words)

distribution of hashes for english words

It is to be assumed that whatever the hash table size should be, it shouldn't be a power of 2; the only special property I could figure out about 701 is that it's prime, but clearly, number theory isn't what I'm normally good at on a Friday night. So, I just tried with 2039 (which happens to be the largest prime number < 2048, and for some reason, that sounded like a good idea at the point), and that kind of worked (the variance of distribution increased by *roughly* a factor of (2039/701)², which is plausible, but at that point, I'd have to derive a term for the variance of the variance of realization of uniform random variables. Not doing that today.) The variance of a discrete random variable with uniform distribution over N values is (N²-1)/12, and while figuring that out was a nice short exercise, I realize I shouldn't be the one that abuses low-order stochastic moments to reason about hashes.

There's decades of Algebra and CS to design good hashes that are uniform, so either we can trust this hash (for example, because @jmcorgan says its good, assuming he was around when it was introduced and simply knows more than I do about it), or we just replace it with whatever hash standard libraries use in their hashmaps for strings. And while we're doing that, we can just as well just use that hashmap.

> stick with interned symbols and use a different data structure (maybe something that adapts its size dynamically once it reaches a certain utilization).

I'd say, yes, maybe that's a good idea. And, yet another reason to hope for C++11, as that has `unordered_map`, whose implementations pretty certainly have some B-Tree underneath, considering the standard talks of "buckets". 

Our current average complexity is (SIZE_TABLE+  N_COLLISIONS/SIZE_TABLE)/2 (we do a linear search through the hash table, and then jump along the list of colliding strings; the latter through being indirect is probably a lot more expensive). So, even assuming there'd be no hash collisions, we have O(N); the basic B tree has O(log N) worst-case.

> extend symbols to allow non-interned strings (I wouldn't change the defaults, which would require adapting core components)

I see that you wouldn't unnecessarily want to modify core components, but I'd argue that there's really just tags and the msg port registrars as well as msg_connect to be changed in the core – and by changing the default, we'd use the (less expensive if not needed) string type, which is probably what people would want, right? (That's a serious question – if 99% of string usage of PMT is actually going to be symbols, then, oh well)

bastibl:
I think first we have to decide whether we want to support this use case. I can't think of an argument against just using blobs. If you receive some bytes, you could just as well move them around in blobs. Or is there a reason to immediately interpret them as strings?

Only if we really, really want to support that use case, we should think of how to do it properly. In my opinion, there should be only one string class. So the options would be to add support for non-interned strings or intern everything and add garbage collection. GC is important because it's not only about hash collisions but also about memory usage. Currently, the strings would stay around forever, wasting memory and causing collisions. 

How to maintain interned symbols might be an independent question. With
> stick with interned symbols and increase the hash table size

I didn't mean to blindly increase the hash table size and hope for the best. If we want to stick with the current approach, we'll find a way to hash a string to an array of arbitrary size.
But if we want to change something, we might first want to think if the problem here is so specific that we need a custom implementation. There is an unorder_map in boost, which might be an alternative until we have C++11.

> I'd say, yes, maybe that's a good idea. And, yet another reason to hope for C++11, as that has unordered_map, whose implementations pretty certainly have some B-Tree underneath, considering the standard talks of "buckets".

Looking at `bucket_count` and `rehash` I don't think that they use B-trees. I guess it's just the same as we have (except for automatic adaptation of the hash table size).

marcusmueller:
> can't think of an argument against just using blobs. If you receive some bytes, you could just as well move them around in blobs. Or is there a reason to immediately interpret them as strings?

If we wanted to push around raw bytes, why would we have PMTs in the first place? I think the difference between a PMT and a buffer with a size-defining header is that the PMT includes type info, which, to some extent, is even semantic (we make and break those semantics all the time, ie. a list is a tuple of element, tuple(element,tuple)…, a dictionary is a list of (key,value), and such). 

If user desire is to have a portable string wrapper, then we need a portable string wrapper, so we at least should define that strings in a blob have a specific encoding. It might be clever to formalize that with a PMT specialization.

Assuming we're not going the UTF-32 route, that would mean len(BLOB) / sizeof(glyph) !=  len(string), so that the string length is something that the receiver needs to parse from the payload. I'm currently arguing with myself whether having the string length (as something that actually results from the data) as field in a PMT is a bad or a good idea. Probably handy, but probably a security hazard (yeah, sure, parse my 2³² byte long string).

bastibl:
So after we introduce the new string semantic, we immediately worry about the internal byte representation :-) I think that highlights the problem.

In all use-cases I can think of, I'd more worry about the bytes then the string semantic. In my opinion, it only complicates stuff (can I fit that 3 character string in a 4 byte frame? What is the string length? What encoding does GNU Radio use? How can I convert my UTF-16 string in a GNU Radio string?) 

It's complicated and GNU Radio cannot help here. Some applications might use ASCII, others support Unicode with UTF-8/16/32, and again others something custom... So how to convert a string to and from bytes is nothing we could do. That means the user would have to figure out how to make a GNU Radio string out of his chunk of bytes and vice versa.

Maybe I just don't get the advantage of a string semantic. I'd just avoid the hassle and pass them around as blobs. 

> If we wanted to push around raw bytes, why would we have PMTs in the first place?

And don't see why the fact that there are valid use-cases for a blob, questions the existence of PMTs in general.

Anyhow, since people want to do this, I'd recommend using std::string internally and report lengths, etc similar to it. If we try to invent our own better strings, we might end up like Boost. I also don't think that there are any security issues. AFAIK, GNU Radio does not implement any security concept that protects you from bogus data.

According to my understanding, the plan would be to
- rename symbols to string
- add a parameter to specify if this string is interned or not (i.e., whether it is used as string or identifier)
- make the new strings the default and change the API

I'm not sure how you planned to handle the last point

> I see that you wouldn't unnecessarily want to modify core components, but I'd argue that there's really just tags and the msg port registrars as well as msg_connect to be changed in the core 

I don't see how this can be done without changing all blocks (and adpat OOTs). How is, for example, `add_item_tag` supposed to work. Either we change its use in the blocks, to
```add_item_tag(_, _, pmt::string_to_symbol("wifi_start", false), _, _);```
or
```add_item_tag(_, _, "wifi_start", _, _);```
If we stick to the current
```add_item_tag(_, _, pmt::string_to_symbol("wifi_start"), _, _);```
the non-interned string is already created.

**Comment by marcusmueller** [comment]
Hm, you raise a very fine point with your "let's not worry about application payload representation", indeed. GNU Radio's not going to solve string representation.

> And don't see why the fact that there are valid use-cases for a blob, questions the existence of PMTs in general.

That's true, too. So yeah, I have to  retract that argument.

> *  make the new strings the default and change the API
>
> I'm not sure how you planned to handle the last point

In the case of tag keys, I think using interned strings is the right thing – so no need to change things there, especially if people are already using the explicit "string_to_symbol". 

What I meant was that `pmt::mp` (which is the "do the right thing" method, but to be honest, I don't know what `mp` stands for) should give you a *non-interned* string, that the methods that actually want an interned string can convert to an interned string internally.

So, to keep in the examples, if your tags contained payload data, e.g. the currently playing song via RDS, that payload shouldn't be interned by default.

**Comment by bastibl** [comment]
OK, was a bad example, because I actually expected that the whole 'symbol' naming gets dropped, i.e., that we only stick to `mp` and `string`, which, AFAIS, would require to change from `pmt::mp("foo")` to `pmt::mp("foo", false)`

Anyhow, I think these are details. If you still think that this is a good idea, someone would have to implement it. Who's in @marcusmueller, @samfpetersen ? @samfpetersen, I've seen you already worked in that direction in another branch.

**Comment by samfpetersen** [comment]
I've been out of the loop for a couple weeks, but before I left I tried to put together an [implementation](https://github.com/samfpetersen/gnuradio/tree/pmtstring) that doesn't alter the existing API, but that uses a single pmt_string class for both interned and non-interned strings. My main guidelines were

- Existing block code should not have to be changed, API calls should still work the same under the hood
- Interning should be an optional feature but should be recommended for efficiency in many cases -  `pmt::string_to_symbol` and `pmt::from_string` should be interchangeable from a behavior standpoint, the only difference should be performance (speed, memory usage, etc.).
- Serialization should preserve whether the string is interned or non-interned 
- Python string <-> string PMT conversion should not require the user to think about interning, but it should be possible to create interned and non-interned string PMTs from Python

In this implementation, calling `pmt::string_to_symbol` and `pmt::intern` will always create interned string PMTs, keeping existing functionality. Calling `pmt::from_string` will return an interned string PMT if the string has already been interned, otherwise it will return a non-interned string PMT. Although this involves computing the hash even when not interning, it makes the pmt<->python conversion simpler, a feature that I use often. Also, I left the syntactic sugar `pmt::mp` functions as is, calling `pmt::string_to_symbol` (maybe these should call `pmt::from_string` as per @marcusmueller's last comment).

With this implementation, the user would have to explicitly intern strings using `pmt::intern` or `pmt::string_to_symbol`. I agree with @bastibl's initial assessment that providing the `pmt::from_string` method will mean most users won't use interning at all, but I think the performance impact of the string comparisons will be negligible for these type of users anyway. I think if there was some info about when to use `pmt::intern` vs `pmt::from_string` in the documentation, more serious users would pick up on it quite easily. Plus, all the example code already uses `pmt::intern`/`pmt::string_to_symbol` for things like dictionary keys and tag keys.

Let me know what you guys think - I think this differs from what has been discussed lately but a lot of the same ideas are there.

**Comment by bastibl** [comment]
Looks good to me. However, one problem seems to be that

```
import pmt
a = pmt.to_pmt('foo')
pmt.is_symbol(a)
-> False
```

At least it breaks `qa_tag_share`. Following this approach, I think symbols should be the default everywhere.

@marcusmueller, what do you think about this approach?

jmcorgan:
@bastibl  @marcusmueller let me know where this stands.

samfpetersen:
@bastibl the PMT<->Python implications are tricky for this feature in general. Boils down to the fact that when a PMT string is converted to a Python string, whether or not it was intended to be interned is lost.

For example, a few Python blocks I use receive data from a message port as a PMT dictionary. They basically convert the PMT dictionary to a Python dictionary, do some work with that data (often modifying or adding dictionary entries), and then send the data on an output message port as a PMT dictionary again. In this case, if symbols were the default when converting Python->PMT, even if I had previously chosen to not intern a string in that PMT dictionary, it would be interned after I convert it to Python and back. That's why in my implementation, I made Python->PMT return symbols only if the string has already been interned, otherwise it returns a non-interned string. 

I couldn't think of a way to keep symbols as the default while keeping the PMT<->Python functionality intuitive to the user.

bastibl:
But now we have inconsistent defaults or am I getting this wrong? AFAIS, you create symbols by default in C++ and strings by default in Python. I don't think that's a good idea.
I think that either we:
- Create symbols by default, which doesn't require changes to the core. You would have to create a utility function to create a PMT dictionary with strings from a Python dictionary.
- Create strings by default, which would change the semantic of the API.
@marcusmueller what do you think. AFAIS, you favored the second option.

mbr0wn:
Is there an option for a compromise here, where we add some convenience methods / aliases to create blobs from strings and vice versa? Python has the concept of byte strings, too, which map directly to `std::string` (UTF-* strings don't have that simple conversion).

samfpetersen:
Hmm, so I was stuck on how to proceed with this and I thought of a [totally different solution](https://github.com/gnuradio/gnuradio/compare/master...samfpetersen:pmt-symbol-gc) that involves garbage collecting symbols that are both unused and old. This involves no change at all to the API, and I believe the added cost of garbage collection is fairly low (have not actually benchmarked). It does seem a little bit hacky to me, but it is thread safe. 

Before a symbol is added to the hash table, all the symbols in that hash bin are checked for garbage collection. The GC removes a symbol from the hash table if 1) the symbol is not referenced except for in the hash table, and 2) the symbol "hasn't been accessed in a while". Number 1 works by inspecting the reference count used by `boost::intrusive_ptr`. Number 2 works by keeping a running total of the number of symbols created, comparing the current total with the value of the `d_expiration` field in the symbol class. Every time the symbol is retrieved the `d_expiration` field is updated by setting it to the current total plus an arbitrary constant (`SYMBOL_EXPIRATION`). 

It might be a good idea to enable advanced users to enable/disable GC, possibly even change the `SYMBOL_EXPIRATION` constant to a user-configurable value. 

Let me know if this solution sounds any better

bastibl:
It's getting a bit confusing. My current understanding is this:

- one question is whether we really want the string semantic in PMTs. I think that's what @marcusmueller wants. The initial PR really differentiated between string and symbol. The current one with garbage collection does not.

- if it is really only about performance and we don't care about string semantics, I agree with @mbr0wn. I would not change the core, but use blobs to transport strings and use some convenience methods to convert back and forth.

jmcorgan:
I'm onboard with @bastibbl @mbr0wn thinking here.

Ka-zam:
Like samfpetersen I use interned strings to send decoded data around and need long term stability in my system. Has there been any decisions made regarding this issue yet?

bastibl:
I think, most people would like to stick to the current implementation and use blobs in this use-case. I'd recommend to close this PR if nobody has strong feelings against that.

samfpetersen:
I'm fine with that, but I think there should at least be more information in the documentation about this problem. Key points would be:

- Don't use strings to represent data, only use strings for names of things
- `pmt::intern()` / `pmt::string_to_symbol()` are relatively expensive operations, so save the result for reuse

A little bit of info in the documentation about string interning might also be useful to shine some light on these points.

Also, I'm still in favor of increasing the number of hash bins, in light of @marcusmueller's hash collision analysis.

marcusmueller:
In this case, closing the PR! Thanks for the good code, and great discussion. As I type this, @mbr0wn is creating an issue (see below) so that we don't lose the points that are in @samfpetersen last comment here

PR #1453, opened 2017-08-30
Renamed doc to docs so it would render nicely on GitHub

bogdanr:
Also it makes it all consistent because some of these have the directory named docs.


noc0lour:
@jmcorgan Yes, no? I'd recommend squashing the commits and force pushing on your branch, but it's a trivial improvement.

jmcorgan:
I'm a little confused at the goal of this PR, can you elaborate?

bogdanr:
This module has the dir named docs:
https://github.com/gnuradio/gnuradio/tree/master/gr-zeromq
and this one doc:
https://github.com/gnuradio/gnuradio/tree/master/gr-vocoder

The naming is not consistent and as you can see, in the first case github renders documentation out of the box, which makes it a bit more user-friendly.

noc0lour:
And it would make naming consisten across all modules (:

mbr0wn:
Manually merged.

PR #1455, opened 2017-09-01
Add threshold to plateau and S&C OFDM sync

michaelld:
Adding a threshold callback and set / get allows more dynamic use of the S&C OFDM sync block. With a default threshold of 0.9, the S&C OFDM sync block has a lower noise threshold than the header or payload (e.g., if encoded using FEC). These changes allow selection of the mid-level threshold (e.g., 0.5) such that FEC can be used with benefits. In my testing, the GRC blocks are backward compatible with the current ones, as they set default values for the threshold and otherwise change nothing.

### DIFF
--- a/gr-blocks/include/gnuradio/blocks/plateau_detector_fb.h
+++ b/gr-blocks/include/gnuradio/blocks/plateau_detector_fb.h
@@ -61,6 +61,9 @@ namespace gr {
 	* \param threshold Anything above this value is considered a plateau
 	*/
        static sptr make(int max_len, float threshold=0.9);
+
+       virtual void set_threshold(float threshold) = 0;
+       virtual float threshold() const = 0;
     };
 
   } // namespace blocks
--- a/gr-blocks/lib/plateau_detector_fb_impl.cc
+++ b/gr-blocks/lib/plateau_detector_fb_impl.cc
@@ -86,6 +86,16 @@ namespace gr {
       return i;
     }
 
+    void plateau_detector_fb_impl::set_threshold(float threshold)
+    {
+      d_threshold = threshold;
+    }
+
+    float plateau_detector_fb_impl::threshold() const
+    {
+      return d_threshold;
+    }
+
   } /* namespace blocks */
 } /* namespace gr */
 
--- a/gr-blocks/lib/plateau_detector_fb_impl.cc
+++ b/gr-blocks/lib/plateau_detector_fb_impl.cc
@@ -61,6 +61,9 @@ namespace gr {
                       gr_vector_const_void_star &input_items,
                       gr_vector_void_star &output_items)
     {
+      // thread-safe protection from ::set_threshold
+      gr::thread::scoped_lock l (d_setlock);
+
       const float *in = (const float *) input_items[0];
       unsigned char *out = (unsigned char *) output_items[0];
       int flank_start;
@@ -88,6 +91,8 @@ namespace gr {
 
     void plateau_detector_fb_impl::set_threshold(float threshold)
     {
+      // thread-safe protection from ::set_threshold
+      gr::thread::scoped_lock l (d_setlock);
       d_threshold = threshold;
     }
 
--- a/gr-blocks/lib/plateau_detector_fb_impl.h
+++ b/gr-blocks/lib/plateau_detector_fb_impl.h
@@ -44,6 +44,9 @@ class plateau_detector_fb_impl : public plateau_detector_fb
                       gr_vector_int &ninput_items,
                       gr_vector_const_void_star &input_items,
                       gr_vector_void_star &output_items);
+
+  virtual void set_threshold(float threshold);
+  virtual float threshold() const;
 };
 
   } // namespace blocks
--- a/gr-digital/include/gnuradio/digital/ofdm_sync_sc_cfb.h
+++ b/gr-digital/include/gnuradio/digital/ofdm_sync_sc_cfb.h
@@ -71,8 +71,13 @@ namespace gr {
        *  \param use_even_carriers If true, the carriers in the sync preamble are occupied such
        *                     that the even carriers are used (0, 2, 4, ...). If you use all
        *                     carriers, that would include the DC carrier, so be careful.
+       *  \param threshold detection threshold. Default is 0.9.
        */
-      static sptr make(int fft_len, int cp_len, bool use_even_carriers=false);
+      static sptr make(int fft_len, int cp_len, bool use_even_carriers=false,
+                       float threshold=0.9);
+
+      virtual void set_threshold(float threshold) = 0;
+      virtual float threshold() const = 0;
     };
 
   } // namespace digital
--- a/gr-digital/lib/ofdm_sync_sc_cfb_impl.cc
+++ b/gr-digital/lib/ofdm_sync_sc_cfb_impl.cc
@@ -27,7 +27,6 @@
 #include <gnuradio/io_signature.h>
 #include "ofdm_sync_sc_cfb_impl.h"
 
-#include <gnuradio/blocks/plateau_detector_fb.h>
 #include <gnuradio/blocks/complex_to_arg.h>
 #include <gnuradio/blocks/complex_to_mag_squared.h>
 #include <gnuradio/blocks/conjugate_cc.h>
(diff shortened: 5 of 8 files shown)


gr-blocks/include/gnuradio/blocks/plateau_detector_fb.h
```diff
@@ -61,6 +61,9 @@ namespace gr {
 	* \param threshold Anything above this value is considered a plateau
 	*/
        static sptr make(int max_len, float threshold=0.9);
+
+       virtual void set_threshold(float threshold) = 0;
+       virtual float get_threshold() = 0;
```
jmcorgan: Our coding style for getters is to omit the leading get_ and just use a public function name that is the same as the internal variable name.

jmcorgan:
As this isn't a bug fix, it should be on the master branch.

gr-digital/include/gnuradio/digital/ofdm_sync_sc_cfb.h
```diff
@@ -71,8 +71,13 @@ namespace gr {
...
        *                     that the even carriers are used (0, 2, 4, ...). If you use all
        *                     carriers, that would include the DC carrier, so be careful.
+       *  \param threshold detection threshold. Default is 0.9.
        */
-      static sptr make(int fft_len, int cp_len, bool use_even_carriers=false);
+      static sptr make(int fft_len, int cp_len, bool use_even_carriers=false,
+                       float threshold=0.9);
+
+      virtual void set_threshold(float threshold) = 0;
+      virtual float get_threshold() = 0;
```
jmcorgan: Same issue here, the getters shouldn't include get_.

gr-blocks/include/gnuradio/blocks/plateau_detector_fb.h
michaelld: Ah; OK. So just "threshold" to get the value. NP will do for both.

gr-blocks/include/gnuradio/blocks/plateau_detector_fb.h
```diff
@@ -61,6 +61,9 @@ namespace gr {
 	* \param threshold Anything above this value is considered a plateau
 	*/
        static sptr make(int max_len, float threshold=0.9);
+
+       virtual void set_threshold(float threshold) = 0;
```
mbr0wn: `const float`

gr-blocks/lib/plateau_detector_fb_impl.cc
```diff
@@ -86,6 +86,16 @@ namespace gr {
       return i;
     }
 
+    void plateau_detector_fb_impl::set_threshold(float threshold)
```
mbr0wn: `const float`

gr-blocks/lib/plateau_detector_fb_impl.h
```diff
@@ -44,6 +44,9 @@ class plateau_detector_fb_impl : public plateau_detector_fb
                       gr_vector_int &ninput_items,
                       gr_vector_const_void_star &input_items,
                       gr_vector_void_star &output_items);
+
+  virtual void set_threshold(float threshold);
```
mbr0wn: `const float`

marcusmueller:
Hey, I'd really love to merge this; can you do a `it commit --fixup` and submit the proposed changes (both @jmcorgan rename request as well as @mbr0wn's constness)? Thanks!

michaelld:
LOL I'd forgotten about this PR. Sure I'll fix it up & rebase to current.

gr-blocks/lib/plateau_detector_fb_impl.h
michaelld: I can certainly change these to "const float" ... but I don't see why.

The "const" here just indicates that the "threshold" variable cannot be modified, right? It's a compiler directive & likely won't change the resulting assembly code, and hence I greatly doubt it'll change the runtime performance. I'm pretty sure this variable is not modified in the method implementation, and it's simple enough to change the code to not modify it (just copy to a non-const variable). Even without the const, it doesn't matter if the variable is changed, since it's value is not returned to the caller.

Hence, I don't see a good reason to add the "const" .... unless this is a GREP thing ... ? I haven't had time to look through the coding GREP yet.

michaelld:
OK I've read through GREP 0 draft, and it doesn't cover adding a "const" here ... really not sure the benefit in doing so. If this were a reference, then I'd be all for it. But it's a type float, which is 4 bytes and likely passed via a register anyway (in compiled form). Thoughts?

mbr0wn:
The `const`s are not a hard requirement, it's just generally good practice to use `const` and `constexpr` generously. We don't have a coding guideline GREP yet (GREP 0 doesn't include that), but the C++ Core Guidelines for example call it out explicitly: https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#S-const

In this PR, the `const`s would do absolutely nothing, but it's just a hint to the compiler that come in useful in avoiding bugs later on. That said, there are definitely cases where using `const` is not necessary and will in fact make code unnecessarily complicated, but const-correctness is one good thing, and giving the compiler as many hints as possible as to the intended use of the code is another one.

mbr0wn:
...for example, the compiler could decide to inline stuff that it wouldn't otherwise.

michaelld:
Of course, the next paragraph is the "Exception" https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#exception-50
{{{
Function arguments are rarely mutated, but also rarely declared const. To avoid confusion and lots of false positives, don't enforce this rule for function arguments.
}}}

michaelld:
OK so these here are methods, not functions. Whatever. Same applies IMHO. I don't care one way or the other, really, if these are const or not const. I don't think it helps, but if y'all want them const then I'll make them so.

marcusmueller:
Hey, @michaelld, thank you for that. :) I'm really just trying to get this into the release.

I'd agree with you both (and honestly, I think you and Martin aren't even disagreeing); `const` is not strictly useful here to the human reader, it might make inlining easier for the compiler (but then again, a compiler that inlines a setter is probably smart enough to see that it doesn't need to save the old value because the *code* doesn't touch that register). 

Anyway, it's just super nice to have you on boat with this, so really: thanks for having GR's back here, once again :)

michaelld:
OK I redid this commit to be off the current master & fixed the added methods to remove the "get_" part. Tested building in MacPorts with no warnings (or errors).

I did not change the method arguments to "const" since that's contrary to the way the other methods' arguments are declared &, from our discussion, not really necessary. Further, instead of changing such arguments piecemeal, I would suggest a massive change that does them all at the same time (or, a few commits that do them all subsequently) so that their declarations are uniformly the same across all components. This change would be a massive undertaking, of course, but it would start and set a particular coding style -- if this is truly what we want for GR (I'm not convinced it is).

Hmmm ... related: I would agree that the (get) method itself should be const, since it doesn't change the underlying class instantiation. Maybe this is what @mbr0wn wanted in the first place (really)?

michaelld:
@marcusmueller YW & I'm doing what I can. I've got 1 more PR to get in place, after some more testing on my part. It's an almost trivial change, but critical for specific functionality. Saw yesterday that there might already be QA in place that passes, which maybe it shouldn't (or maybe I'm using the block incorrectly). Anyway I'll try to get that going this week still.

michaelld:
Changed (again) so that the (get) method is "const" ... along with the renaming of the (get) method per the above. Now I like what I see & it makes sense.

gr-blocks/lib/plateau_detector_fb_impl.cc
```diff
@@ -86,6 +86,16 @@ namespace gr {
...
+    {
+      d_threshold = threshold;
```
marcusmueller: We should think about thread safety here. What happens if we change the threshold midst-`general_work`?

For now, this is as fine as "any other block<sup>TM</sup>", but if I'm brutally honest, "any other block<sup>TM</sup>" is still broken.

Anyway, not anything that I plan to fix within the scope of this PR, so this is fine to me :)

gr-digital/lib/ofdm_sync_sc_cfb_impl.cc
```diff
@@ -106,6 +108,15 @@ namespace gr {
     {
     }
 
+    void ofdm_sync_sc_cfb_impl::set_threshold(float threshold)
+    {
+      d_plateau_detector->set_threshold(threshold);
+    }
+
+    float ofdm_sync_sc_cfb_impl::threshold() const
```
marcusmueller: yep, excellent.

michaelld:
Indeed thread safety wasn't paramount. Adding in the mutex from gr::block is pretty trivial if you want me to do that.

gr-blocks/lib/plateau_detector_fb_impl.cc
michaelld: Indeed thread safety wasn't paramount. Adding in the mutex from gr::block is pretty trivial if you want me to do that.

michaelld: But, yes, many (most?) GR blocks aren't truly thread-safe when setting variables. "there has got to be a better way"™ ...

marcusmueller: yep, exactly, I have Ideas<sup>TM</sup> about what that better way<sup>TM</sup> should look like, but we can't proactively fix before we've discussed that approach.

If you feel it's really not much work, do add the mutex here. It's the Right Thing<sup>TM</sup> to do :)

michaelld:
OK I just pushed the thread safe changes. Small, quick, simple to understand, and even commented!

marcusmueller:
Narf, now the test builder fails (not your fault). I can't triggeer a rebuild (not in the right group for that), so I'll have to wait till @noc0lour takes care of the rebuild.

mbr0wn:
Failing on unrelated tests now, can thus be merged.

PR #1456, opened 2017-09-01
tuntap_pdu: use MTU+14 byte receive buffer for TAP

mmaroti:
The MTU for TUN/TAP interfaces is always the IP frame MTU (1500 bytes). Currently the receive buffer is of size MTU both for TUN and TAP, but for TAP it should be MTU+14 to be able to hold the extra Ethernet header. We fix this and allow the reception of MTU sized IP frames through TAP.


jmcorgan:
This was cherry-picked onto the maint branch as a bug fix.

PR #1458, opened 2017-09-02
doc: improved msg passing section of doxygen usage manual

777arc:
Several changes to msg_passing.dox including:
1. Adding Python versions of message_port_register and pub commands right after the C++ versions
2. Moving examples to the bottom, to improve flow
3. There were a couple portions of redundant explanations
4. I tried to make it more clear that message_port_sub() is not something we usually call ourselves, but is internal to msg_connect(), so as to not confused new people
5. General proofing

### DIFF
--- a/docs/doxygen/other/msg_passing.dox
+++ b/docs/doxygen/other/msg_passing.dox
@@ -14,7 +14,7 @@
 GNU Radio was originally a streaming system with no other mechanism to
 pass data between blocks. Streams of data are a model that work well
 for samples, bits, etc., but are not really the right mechanism for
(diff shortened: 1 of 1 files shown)


docs/doxygen/other/msg_passing.dox
```diff
@@ -57,20 +57,33 @@ to register a new port are:
   void message_port_register_out(pmt::pmt_t port_id)
 \endcode
 
+In Python:
+
+\code
```
mbr0wn: You can flag this as python and doxygen will syntax highlight.

PR #1462, opened 2017-09-08
Added a vector operation mode to the moving average

marcusmueller:
This allows users to average per-element.

Might be very useful for spectrum "condensing" and such.

### DIFF
--- a/gr-blocks/lib/moving_average_XX_impl.cc.t
+++ b/gr-blocks/lib/moving_average_XX_impl.cc.t
@@ -1,6 +1,6 @@
 /* -*- c++ -*- */
 /*
- * Copyright 2008,2010,2013 Free Software Foundation, Inc.
+ * Copyright 2008,2010,2013,2017 Free Software Foundation, Inc.
  *
  * This file is part of GNU Radio
  *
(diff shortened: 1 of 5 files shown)


gr-blocks/lib/moving_average_XX_impl.cc.t
```diff
@@ -95,18 +101,37 @@ namespace gr {
...
+
+        for(unsigned int i = 0; i < num_iter; i++) {
+          sum += in[i+d_length-1];
+          out[i] = sum * d_scale;
+          sum -= in[i];
+        }
+
+      } else { // d_vlen > 1
+        //TODO: VOLKize this zeroing?
+        std::fill(d_sum.begin(), d_sum.end(), (@I_TYPE@)0);
```
mbr0wn: Having spent some time with the C++ standard library, I don't trust these algorithms to be as high-performance as low-level system calls. Although I hate mixing C and C++ calls, a memset here will will have a better chance of being optimal performance.

gr-blocks/lib/moving_average_XX_impl.cc.t
```diff
@@ -95,18 +101,37 @@ namespace gr {
...
+
+        for(unsigned int i = 0; i < d_length - 1; i++) {
+          for(unsigned int elem = 0; elem < d_vlen; elem++) {
+            d_sum[elem] += in[i*d_vlen + elem];
+          }
+        }
+
+        for(unsigned int i = 0; i < num_iter; i++) {
+          for(unsigned int elem = 0; elem < d_vlen; elem++) {
```
mbr0wn: I would not consider this a blocking issue, but these loops seem like they might have more VOLKizing potential.

gr-blocks/lib/moving_average_XX_impl.h.t
```diff
@@ -25,6 +25,8 @@
 #ifndef @GUARD_NAME_IMPL@
 #define @GUARD_NAME_IMPL@
 
+#include <vector>
```
mbr0wn: I don't think we have a standard for this, but it seems like the C++ world in general is standardizing on including standard headers last, not first. This is to avoid the case where you accidentally hide a missing include by pre-including stuff early.

mbr0wn: I would also not consider this a blocking issue.

gr-blocks/python/blocks/qa_moving_average.py
```diff
@@ -87,5 +87,41 @@ def test_02(self):
...
+    def test_03(self):
+        tb = self.tb
+
+        print "setting up"
+        vlen = 5
+        N = 10*vlen
+        seed = 0
+        data = make_random_float_tuple(N, 2**10)
+        data = [int(d*1000) for d in data]
+        print "making blocks"
```
mbr0wn: These print statements should go -- they're not really helpful, they look like someone was debugging segfaults, and they're not Py3k compatible.

marcusmueller:
I don't trust them to be high-performing, either, but I trust them to be
right.

Just a reminder that `float` doesn't mean IEEE754 32-bit float. It might
be some DSP 20 bit floating point for which 0x00000 might not be 0.0f.

Now, realistically, that's never the case. But before I start with
optimizing the zerosetting, I'm just now realizing that I should simply
skip the zeroing, and just initialize the vector with the first summand.
'doh! Should've been doing that form the start!

gr-blocks/python/blocks/qa_moving_average.py
jmcorgan: @marcusmueller please take care of this

marcusmueller: @jmcorgan took care. Please let me know if changeset is OK now; I will then squash commits. By the way, I currently do not like github's PR work flow. Feels like the whole PR-review-update branch-review-… cycle is less clean than the usual "patch via mailing list" approach.

mbr0wn: Let's not have a discussion on the tooling. It is, however, a bit cleaner if you you `--fixup` commits for easier tracking of updates in the git history.

marcusmueller:
I tend to disagree. If I force-push changes to the branch we're
discussing here, then the Github Pull-Request discussion and review UI
becomes completely unusable, since there's no source code state it can
refer to.

mbr0wn:
@marcusmueller Using `--fixup` commits implies *not* force-pushing until reviews are over.

PR #1467, opened 2017-09-18
grc/qtgui: various Python 3 fixes

hauk142:


grc/core/generator/top_block.py
```diff
@@ -227,7 +228,7 @@ def make_port_sig(port):
                 key = port.key
 
             if not key.isdigit():
-                key = repr(key)
+                key = re.findall(r'\d+', key)[0]
```
skoslowski: @hauk142 what is the reasoning for this? it looks like your breaking message connections.

hauk142: Trying to run a generated flowgraph without this fix gives me a NotImplementedError. I've included the error message in the commit message :)

This fix takes for example 'in2' and changes it to 2.

skoslowski: Yes, that should have been in a WIP commit. Didn't realize that got merged. Checkout _init_ports() in block.py. Removing in and out there allows you to revert this change here

hauk142: Right! I'll have a look

PR #1469, opened 2017-09-21
cmake: fix build issue with introduction with -std=c99 define

noc0lour:
This PR fixes an issue introduced with #1424.
volk uses the `asm` keyword which is only defined in GNU extensions and is not present in builds with `-std=c99` for older GCC versions.

@michaelld Will this break cppunit again?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -71,19 +71,15 @@ ELSE()
 ENDIF()
 
 IF(CMAKE_C_COMPILER_ID STREQUAL "GNU")
-    SET(CMAKE_C_STANDARD 99)
+    SET(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -std=gnu99")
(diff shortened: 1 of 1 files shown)


michaelld:
This change won't break CppUnit; it relies solely on CMAKE_CXX_STANDARD & related variables.

That said, why not instead add a 2nd variable which is the "c" or "gnu" part. Set it to "gnu" by default & to "c" for MSVC. Then use it in the setting line at the end for if cmake version is < 3.1.

Setting CMAKE_C_STANDARD has a much higher chance of working compared with appending "-std=..." to CFLAGS, and is the preferred method to set the C standard to use since CMake 3.1.0.

noc0lour:
ok, now only the QA test is flaky. I'll try to apply your suggestions tomorrow.

jmcorgan:
Don't worry if this doesn't merge up cleanly--whatever changes are needed for maint I'll merge in and then skip those when merging up. The next/python3 branches work as-is; let me know if master needs this fix or not. And thanks again.

PR #1473, opened 2017-09-27
grc: python3 bug fixes

hauk142:
Please note that if you try to open a .grc file created before the port-related fix (https://github.com/gnuradio/gnuradio/pull/1473/commits/d8607abd1c32c5ff983a18e080b98e584444d200) with GRC built after this fix, you will have to change all the connections in the .grc file or else you will get an error. 

For example, you will have to replace 'in0' with '0' in the .grc file. I suppose enabling GRC to read "old-style" .grc files shouldn't be a problem though.

### DIFF
--- a/grc/compiler.py
+++ b/grc/compiler.py
@@ -70,7 +70,7 @@ def main(args=None):
         Messages.send('\n')
 
         flow_graph, file_path = platform.load_and_generate_flow_graph(
-            os.path.abspath(grc_file), os.path.abspath(out_dir))
+            os.path.abspath(grc_file), os.path.abspath(out_dir) + "/")
         if not file_path:
             exit('Compilation error')
     if file_path and args.run:
--- a/grc/compiler.py
+++ b/grc/compiler.py
@@ -70,7 +70,7 @@ def main(args=None):
         Messages.send('\n')
 
         flow_graph, file_path = platform.load_and_generate_flow_graph(
-            os.path.abspath(grc_file), os.path.abspath(out_dir) + "/")
+            os.path.abspath(grc_file), os.path.abspath(out_dir))
         if not file_path:
             exit('Compilation error')
     if file_path and args.run:
--- a/grc/core/blocks/block.py
+++ b/grc/core/blocks/block.py
@@ -127,11 +127,8 @@ class Block(Element):
         port_factory = self.parent_platform.make_port
         port_ids = set()
 
-        def make_stream_port_id(_pool=itertools.count()):
-            return {'sink': 'in', 'source': 'out'}[direction] + str(next(_pool))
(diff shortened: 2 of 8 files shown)


grc/compiler.py
```diff
@@ -70,7 +70,7 @@ def main(args=None):
         Messages.send('\n')
 
         flow_graph, file_path = platform.load_and_generate_flow_graph(
-            os.path.abspath(grc_file), os.path.abspath(out_dir))
+            os.path.abspath(grc_file), os.path.abspath(out_dir) + "/")
```
skoslowski: I'd rather use os.path.join on the consumers of this value

grc/core/blocks/block.py
```diff
@@ -127,11 +127,8 @@ def _init_ports(self, ports_n, direction):
         port_factory = self.parent_platform.make_port
         port_ids = set()
 
-        def make_stream_port_id(_pool=itertools.count()):
-            return {'sink': 'in', 'source': 'out'}[direction] + str(next(_pool))
-
         for i, port_data in enumerate(ports_n):
-            port_id = port_data.setdefault('id', make_stream_port_id())
+            port_id = port_data.setdefault('id', str(next(itertools.count())))
```
skoslowski: perfect

grc/core/utils/hide_bokeh_gui_options_if_not_installed.py
```diff
@@ -20,5 +20,8 @@ def hide_bokeh_gui_options_if_not_installed(options_blk):
     try:
         import bokehgui
     except ImportError:
-        generate_options = options_blk.params['generate_options']
-        del generate_options.options['bokeh_gui']
+        for param in options_blk.parameters_data:
+            if param['id'] == 'generate_options':
+                ind = param['options'].index('bokeh_gui')
+                del param['options'][ind]
+                del param['option_labels'][ind]
```
skoslowski: thanks

skoslowski:
there were a few more fixes in the mako template, that weren't merged in together with the bokeh gui stuff.
this branch readds them on top your PR
https://github.com/skoslowski/gnuradio/tree/python3_bugfix6
put the commit on yours then we can fix it in one step. thanks

hauk142:
@skoslowski will do! Please see my correction on the GRCC bug. 

The reason why I edited Notebook.py is that GRC previously called TopBlockGenerator with the entire path to the actual .grc file, while GRCC only passed the output directory. With this fix they both call with the directory as `file_path`, not the file. In hindsight it might be a good idea to rename `file_path` to `dirname`?

jmcorgan:
Merged via skoslowski/gnuradio

Issue #1302, opened 2017-04-26
Msg input ports not connected to anything are not exposed to Ctrlport


777arc:
While debugging why the example flowgraph **simple_copy.grc** (located in gr-blocks/examples/ctrlport/) wouldn't work I discovered that message input ports that aren't connected to any other block do not become exposed and accessible via ControlPort, so commands like postMessage() won't reach them.  So to get **simple_copy.grc** to work I had to connect a block containing a message output port to the Copy block, which allowed the simple_copy_controller.py to actually enable and disable the block through ControlPort.  Otherwise this example just doesn't work the way it's intended.   

I assume that when this example was written, msg ports without anything connected would still exist and be exposed to ControlPort, but something must have changed since then.  So I think the two options to solve this issue are:

1) Change the flowgraph generator so unconnected message inputs are still active
2) Create a block that has an output message port but doesn't actually produce any messages, so it can be tacked on to this example, and any other flowgraphs that won't work due to this issue.  Here is an example of what I mean:
```python
# Block that doesn't do anything, used to get a msg input port on another block exposed to ControlPort
from gnuradio import gr
import pmt
class blk(gr.basic_block):
    def __init__(self): 
        gr.basic_block.__init__(self,name='Stupid Block',in_sig=[],out_sig=[])
        self.message_port_register_out(pmt.intern("fake_output"))
```

mhostetter:
I think this is related to what I saw in #1239.

777arc:
I don't quite see how it's related to #1239

mhostetter:
What I experienced in #1239 was a source block with an optional message output. When the block was instantiated in GRC but the message output was not connected, the block didn't seem to run properly. I suspected that because the block wasn't connected to the flowgraph via a `msg_connect()` call, that the block wasn't being "included" in the top block somehow, but I wasn't sure.

This seemed similar to me when you said:
> I assume that when this example was written, msg ports without anything connected would still exist and be exposed to ControlPort, but something must have changed since then.

777arc:
Oh ok.  Well with my issue, the block was definitely added to top block and functioned like normal, except for the message input port not working.

mhostetter:
Yeah, they may not be related at all, but my ears perked up when I read the comment I quoted.

bastibl:
[...]

PR #1476, opened 2017-10-02
runtime: Fix issue where setup_rpc() is not called for all blocks in a flowgraph

mrjacobagilbert:
This PR fixes an issue introduced by PR https://github.com/gnuradio/gnuradio/pull/1119 and referenced in issue https://github.com/gnuradio/gnuradio/issues/1302 related to how setup_rpc() is called for blocks. Originally this was done for the source and destination of each streaming connection, which excluded blocks that message only ports. PR1119 made this work for message ports but broke streaming ports.

This PR moves this to the end of the function that interprets flowgraphs and iterates explicitly over each connected block calling setup_rpc() if it is not setup already. This still does not setup RPC for blocks that are not connected to anything else, which may or may not be an issue.

This also removes the workaround for this bug in PR https://github.com/gnuradio/gnuradio/pull/1324 which is no longer necessary.

### DIFF
--- a/gnuradio-runtime/lib/hier_block2_detail.cc
+++ b/gnuradio-runtime/lib/hier_block2_detail.cc
@@ -787,13 +787,6 @@ namespace gr {
     std::vector<basic_block_sptr>::const_iterator b;   // Because flatten_aux is const
     for(b = d_blocks.begin(); b != d_blocks.end(); b++) {
       tmp.push_back(*b);
-      // for every block, attempt to setup RPC
-      if(ctrlport_on) {
-        if(!(*b)->is_rpc_set()) {
-          (*b)->setup_rpc();
-          (*b)->rpc_set();
-        }
-      }
     }
 
     // Now add the list of connected input blocks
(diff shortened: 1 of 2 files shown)


bastibl:
Looks good to me. Some time ago I proposed a very [similar fix](https://github.com/gnuradio/gnuradio/commit/080bbf3dfcef968a2141714852bf691ba442bb2f) for the problem.

Regarding the comment about the blocks without connections, that seems to be a more general problem, not specific to RPC setup. AFAIK, if blocks are not connected at all, they are not handled by the flow graph, i.e., they are also not started.

mrjacobagilbert:
Yep, that looks functionally identical. Was that PR'd? If so why was it not accepted?

Given my confusion with the function (and the previous two failed attempts at doing this correctly) I figure its best to leave this in its own dedicated loop for readability.

After talking w/ Jonathan and your comment I'm going to forget about disconnected blocks. This specifically sounds like an insignificantly small component of supporting that should it be desired.

bastibl:
I did that when I was relocating. That was a bit chaotic, so I think I just forgot about that commit.

Maybe you could remove the comment to avoid confustion.

mrjacobagilbert:
Good idea, I removed that from the commit.

PR #1477, opened 2017-10-03
runtime: expose set_max_noutput_items to python

777arc:
It looks like set_max_noutput_items was exposed to python via swig, but wasn't actually added in the .h file.  Am I missing something here or was it just never added?  I mean if you were going to call this it would be from the python flowgraph right?

### DIFF
--- a/gnuradio-runtime/include/gnuradio/block_gateway.h
+++ b/gnuradio-runtime/include/gnuradio/block_gateway.h
@@ -152,6 +152,10 @@ namespace gr {
       return gr::block::set_min_output_buffer(size);
     }
 
+    void block__set_max_noutput_items(int m) {
+      return gr::block::set_max_noutput_items(m);
+    }
+    
     int block__output_multiple(void) const {
       return gr::block::output_multiple();
     }


jmcorgan:
The swig contains the functions that can be called *on* a block as public functions; block_gateway.h contains functions that can called from *within* a Python block.

777arc:
Wait so then why did I have to make this change in order to be able to do the following in my python flowgraph:

self.my_block.set_max_noutput_items(10) 

Is it because my_block was a python block?

jmcorgan:
Ah, I misunderstood your PR message. Yes, this is needed for python blocks.

mrjacobagilbert:
It might be a good idea to update this to also export the related `set_max_noutput_items()` function as well. It would be quite unexpected to have access to one and not the other.

777arc:
set_min_noutput_items() does not seem to "work", or at least I tried set_min_noutput_items(10000) and I still got the typical 4096 items coming out.  Not sure exactly why, but I might have a clue- if you search the code for "d_max_noutput_items" you get 7 hits, but "d_min_noutput_items" only gets 2 hits, and those 2 locations are only where it's defined, so perhaps it's not really using the parameter.

noc0lour:
@777arc The max (or rather default?) buffer size is set at top block level and defaults to 4096 items. So it is using the maximum allowed buffer size possible.

777arc:
Well I had also tried doing set_min_output_buffer(100000) before calling set_min_noutput_items(10000).  I guess my point is I don't want to expose a function that doesn't do anything, or at least I want to first prove to myself that it can make a difference to the flowgraph operation, before exposing it.  Can anyone make a minimal flowgraph showing that calling set_min_output_buffer() in the top_block on a Python block works?

mrjacobagilbert:
I just tried to do this, and I can't seem to get it to make a difference for C++ blocks...

It was a blind suggestion, as I've ran into similar annoyances before, but perhaps it is not necessary.

777arc:
Yeah, if you search the code for "d_min_noutput_items" you only get 2 hits and neither one seems to be "using" the value to make decisions.  I would say make it a new issue if you want, and meanwhile we can leave it off this PR.

mhostetter:
I was searching around before posting an issue and found this one. I am also having difficulty calling `self.set_min_noutput_items()` from inside a Python block. I know Tom Rondeau tried to fix this in a386edc635ea4a5b0c1db3569bcc03a34cf7493e. I also found this [forum](https://www.ruby-forum.com/topic/4425988) discussing the issue. Just wanted to bring this back up and add a +1 to the issue. Thanks, Matt

mrjacobagilbert:
Current status seems to be that even exposing this call to python doesn't have the generally expected behavior. See issue https://github.com/gnuradio/gnuradio/issues/1483 - I haven't had time to look into it further.

marcusmueller:
As this PR doesn't have the desired effect, I'm closing it. That doesn't mean it's dead! It is still (back)referenced by the issue mentioned above. Note that it's really our goal to expose all core functionality to Pythonland.

PR #1486, opened 2017-10-15
blocks: remove redundant braces in udp source block

talih0:


marcusmueller:
Thank you for observing, reporting and fixing this!

After much deliberation, we've decided that `{ scoped_lock; code; }` is a pattern, even if it's surprising to see unnecessarily nested braces; that element of surprise is _intentional_ to make sure a future developer doesn't forget that he's holding a mutex and mustn't do stuff that conflicts with that.

Issue #1442, opened 2017-08-27
filters and other stateful blocks need to be able to react to some `reset` tag


marcusmueller:
From a discussion with @awalls-cx18 on IRC:

* FFT filters and such are nice, but unusable for bursty operation, since there's no way to clear internal state
* this leads to own implementations and thus, code duplication

Solution I propose:

* Add a GR-wide standard `reset` tag
* On that tag, the scheduler nulls out the input buffer history of a block before calling it. This might require only calling work with items up to the tag position, then modifying the buffer, then calling work again.
* Blocks with internal state SHALL reset that state on the sample they encounter it.
* We *document* that in a GRIPE.

mrjacobagilbert:
Followup on IRC: This will have some impact on performance as all blocks adhering to this will now need to check for and act on tags. The actual impact of this is not clear to me but should be noted.

BrashEndeavours:
The performance offset could be mitigated for non reset-checking blocks by using a *reset enable* flag in the constructor/grc dialog.

marcusmueller:
Hm, that sounds much more like a flag that should have much more
universal application.

I'm thinking of something like a tag filter that you can set to
different values:

* ALL (which would be current behaviour)
* NONE (which means that this block will never call get_tags_in_range;
thus, input tags could directly be handed over to the next block(s)'s
tag container, according to the tag_propagation_policy_t that we already
have)
* LIST (+ a list of tags that your block is interested in. Tags not in
this list would be handled like in the NONE case)

This way, zero tag handling overhead for non-tag-interested blocks, only
tags that are of relevance to a specific block would be copied into the
block's tag map, and blocks that just need all tags for debug, transport
or storage reasons (tagged file sink, ZMQ sinks, tag debug) would still
work.

nowls:
For the FFT filter case, I believe a first requirement is writing a flush/clear/reset method. Second, there's a question of implementation. In the case of the FFT filter, it contains its own history/state whereas FIR filters use the normal GNU Radio buffer history. I just wanted to point out that blocks can be made to implement reset methods (or accept an inherited version that does nothing), whereas there's no good way to impose reset method on kernels, which are generally just stand alone classes.

PR #1489, opened 2017-10-18
gr-filter: Add more filter reset methods

storborg:
This adds ``reset()`` methods to zero the state of various filters. At this point it includes reset methods on both kernels and blocks for FFT filters and single-pole IIR filters, with tests. I'd like to add more filter types before this is merged, but wanted to open a pull request early in case there's feedback.

Related to #1442.

### DIFF
--- a/gr-filter/include/gnuradio/filter/fft_filter.h
+++ b/gr-filter/include/gnuradio/filter/fft_filter.h
@@ -134,6 +134,11 @@ namespace gr {
 	 * \param output  The result of the filter operation
 	 */
 	int filter(int nitems, const float *input, float *output);
+
+        /*!
+         * \brief Reset filter to initial (zero) state.
+         */
+        void reset();
       };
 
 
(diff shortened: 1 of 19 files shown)


storborg:
Also, I'm not sure if it makes sense to implement this for FIR filters or not. If @marcusmueller 's suggestion in #1442 is implemented, it will be redundant. But if it definitely seems like it makes sense to implement reset for FFT and IIR filters, which maintain state independently from the block's buffer.

Also, would it be a welcome change to add tag and message handling to call the reset() method?

marcusmueller:
So, I'm closing this, but actually am pretty thankful for your contribution, without which we wouldn't have gotten this discussion rolling (even if that's been a while back):

Having the `reset()` wouldn't hurt, but we would have to be triple sure about thread safety, because we'd be changing block state from a potentially foreign thread context *while* work is accessing the block.

Thank you also for backing my idea of having a common `reset` tag. Maybe, in the future, let's even having a standardized way for blocks to "register" state that needs to be reset to make `start`/`stop` more smooth. Please keep up the good work!

PR #1491, opened 2017-10-24
Replaced some "import *" imports

aidanbush:
within gr-digital/python/digital
replaced "import *" imports in psk_constellations.py, qam_constellations.py, and test_soft_descisions.py


mbr0wn:
Merged manually (beaa82a7e92a03e8be4ebef823f25983660b1284). Had to fix the commit message.

PR #1497, opened 2017-10-26
grc: make props dialog dark-theme friendly

skoslowski:
simply not set the default text color. 

@jmcorgan, will merge this myself.


skoslowski:
merged into my staging area

PR #1499, opened 2017-10-27
Fix tooltip display for category names

pwicks86:
Category names in the tooltip display don't display correctly if they contain unescaped xml entities, e.g. `&`. They show up as blank tooltips and grc spits out a bunch of errors:

```
gnuradio/grc/main.py:54: GtkWarning: Failed to set text from markup due to error parsing markup: Error on line 1: Entity did not end with a semicolon; most likely you used an ampersand character without intending to start an entity - escape ampersand as &amp;
  gtk.main()
```

This PR escapes the category names so they display correctly.


skoslowski:
Thanks! 
cherrypicked onto merges/maint, will appear on the main repo soon

PR #1500, opened 2017-10-29
BBFrame manipulation

racerxdl:
I was trying to make a Modulator for the GRB Signal (so I could test my software) and I found out that BBheader block only supports TS stream.

So I decided to create a BBHeader block that supports RAW frame input and change of parameters (so it could be used to simulate streams).

I'm not really sure where these changes should be: new component, or same component and different settings (like advanced mode or so). But so far it works for creating a GRB stream.

image

image


drmpeg:
That's a lot of duplicate code. I don't think it can be accepted under the guidelines. Can you test this branch to see if it meets your needs?

https://github.com/drmpeg/gnuradio/tree/bbheader-mode

If it's okay, then I suggest you delete this PR and I'll submit mine instead.

Also, it doesn't make much sense to include all of those parameters unless you can control ISSY in real-time.

racerxdl:
Sure I will give a try @drmpeg , sorry I haven't saw your work. I will give a try and let you know. For now I will close this PR.

PR #1502, opened 2017-10-30
grc: Added the Delete key as a way to delete blocks and connections

haakov:
This feature was present in the `master` branch. I suppose we'd like to keep it?


skoslowski:
Sure, cherry-picked on my staging branch

PR #1503, opened 2017-10-31
New Read ME

bhargav-dindukurthi:
better readability of documentation


777arc:
1. The title and description of your pull request is a bit misleading, you just reworded one phrase
2. The new wording sounds worse in my opinion, because that link goes to a list of dependencies, not a "checker" like a "check if your Java is working" kind of thing

PR #1504, opened 2017-10-31
Fix unpacked behaviour of crc32_bb fixes #1407

### LINKED ISSUE #1407: crc32_bb: broken unpacked processing
As Eric Bandwart [notes](https://lists.gnu.org/archive/html/discuss-gnuradio/2017-08/msg00031.html):

> After some debugging I discovered that the calculated CRC is different depending on whether or not the source data is packed. This can be observed using the attached flow graph. The result of the packed stream appears to be correct. I looked at the code for the block (crc32_impl_bb.cc) and that see that regardless of the Packed setting in the block the CRC calculated with:  `d_crc_impl.process_bytes(in, packet_length);`

>I believe that for the unpacked case, either the bits need be repacked before the calculation or repeated calls to process_bits should be used in place of the the call to process_bytes.

>My version of gnuradio is:  3.7.12git-126-g37d373ac

crc_test grc

annotated plots:

error pattern

noc0lour:
This requires repacking unpacked binary data (currently only LSB) and calculating the CRC.
Previous behaviour generated a mixed packed/unpacked stream of bytes.

I added a new test for the correct behaviour and fixed a test relying on the wrong/old behaviour.
fixes #1407

### DIFF
--- a/gr-digital/lib/crc32_bb_impl.cc
+++ b/gr-digital/lib/crc32_bb_impl.cc
@@ -65,6 +65,26 @@ namespace gr {
       }
     }
 
+    unsigned int
+    crc32_bb_impl::calculate_crc32(const unsigned char* in, size_t packet_length){
+      unsigned int crc = 0;
+      d_crc_impl.reset();
+      if (!d_packed){
+        const size_t n_packed_length = 1 + ((packet_length - 1) / 8);
+        unsigned char packed_buffer[n_packed_length];
+        memset(packed_buffer, 0, n_packed_length);
+        for (size_t bit = 0; bit < packet_length; bit++){
+          packed_buffer[bit/8] |= (in[bit] << (bit % 8));
+        }
+        d_crc_impl.process_bytes(packed_buffer, n_packed_length);
+        crc = d_crc_impl();
+      } else{
+        d_crc_impl.process_bytes(in, packet_length);
+        crc = d_crc_impl();
+      }
+      return crc;
+    }
+
     int
     crc32_bb_impl::work(int noutput_items,
                         gr_vector_int &ninput_items,
@@ -72,14 +92,13 @@ namespace gr {
                         gr_vector_void_star &output_items) {
       const unsigned char *in = (const unsigned char *) input_items[0];
       unsigned char *out = (unsigned char *) output_items[0];
-      long packet_length = ninput_items[0];
+      size_t packet_length = ninput_items[0];
       int packet_size_diff = d_check ? -d_crc_length : d_crc_length;
       unsigned int crc;
 
       if (d_check) {
-        d_crc_impl.reset();
         d_crc_impl.process_bytes(in, packet_length - d_crc_length);
-        crc = d_crc_impl();
+        crc = calculate_crc32(in, packet_length - d_crc_length);
         if (d_packed) {
           if (crc != *(unsigned int *) (in + packet_length - d_crc_length)) { // Drop package
             d_nfail++;
@@ -97,9 +116,7 @@ namespace gr {
         d_npass++;
         memcpy((void *) out, (const void *) in, packet_length - d_crc_length);
       } else {
-        d_crc_impl.reset();
-        d_crc_impl.process_bytes(in, packet_length);
-        crc = d_crc_impl();
+        crc = calculate_crc32(in, packet_length);
         memcpy((void *) out, (const void *) in, packet_length);
         if (d_packed) {
           memcpy((void *) (out + packet_length), &crc, d_crc_length); // FIXME big-endian/little-endian, this might be wrong
--- a/gr-digital/lib/crc32_bb_impl.cc
+++ b/gr-digital/lib/crc32_bb_impl.cc
@@ -45,15 +45,14 @@ namespace gr {
       d_crc_length = 4;
       if (!d_packed) {
         d_crc_length = 32;
-        d_unpacked_crc = new char[d_crc_length];
+        d_buffer = std::vector<char>(d_crc_length);
+      }else{
+        d_buffer = std::vector<char>(4096);
       }
       set_tag_propagation_policy(TPP_DONT);
     }
 
     crc32_bb_impl::~crc32_bb_impl() {
-      if (!d_packed){
-        delete[] d_unpacked_crc;
-      }
     }
 
     int
@@ -71,12 +70,14 @@ namespace gr {
       d_crc_impl.reset();
       if (!d_packed){
         const size_t n_packed_length = 1 + ((packet_length - 1) / 8);
-        unsigned char packed_buffer[n_packed_length];
-        memset(packed_buffer, 0, n_packed_length);
+        if (n_packed_length > d_buffer.size()){
+          d_buffer.resize(n_packed_length);
+          }
+        d_buffer.clear();
         for (size_t bit = 0; bit < packet_length; bit++){
-          packed_buffer[bit/8] |= (in[bit] << (bit % 8));
+          d_buffer[bit/8] |= (in[bit] << (bit % 8));
         }
-        d_crc_impl.process_bytes(packed_buffer, n_packed_length);
+        d_crc_impl.process_bytes(&d_buffer[0], n_packed_length);
         crc = d_crc_impl();
       } else{
         d_crc_impl.process_bytes(in, packet_length);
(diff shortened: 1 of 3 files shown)


gr-digital/lib/crc32_bb_impl.cc
```diff
@@ -65,21 +65,40 @@ namespace gr {
       }
     }
 
+    unsigned int
+    crc32_bb_impl::calculate_crc32(const unsigned char* in, size_t packet_length){
+      unsigned int crc = 0;
+      d_crc_impl.reset();
+      if (!d_packed){
+        const size_t n_packed_length = 1 + ((packet_length - 1) / 8);
```
marcusmueller: Are we 100.000001% sure that `packet_length != 0` at this point?

gr-digital/lib/crc32_bb_impl.cc
```diff
@@ -65,21 +65,40 @@ namespace gr {
...
-      long packet_length = ninput_items[0];
+      size_t packet_length = ninput_items[0];
       int packet_size_diff = d_check ? -d_crc_length : d_crc_length;
       unsigned int crc;
 
       if (d_check) {
-        d_crc_impl.reset();
         d_crc_impl.process_bytes(in, packet_length - d_crc_length);
-        crc = d_crc_impl();
+        crc = calculate_crc32(in, packet_length - d_crc_length);
```
marcusmueller: looks like if `ninput_items[0] == d_crc_length`, you call `calculate_crc32` with an input argument of packet_length = 0

gr-digital/lib/crc32_bb_impl.cc
```diff
@@ -65,21 +65,40 @@ namespace gr {
       }
     }
 
+    unsigned int
+    crc32_bb_impl::calculate_crc32(const unsigned char* in, size_t packet_length){
+      unsigned int crc = 0;
+      d_crc_impl.reset();
+      if (!d_packed){
+        const size_t n_packed_length = 1 + ((packet_length - 1) / 8);
+        unsigned char packed_buffer[n_packed_length];
```
marcusmueller: The gods of C99 do smile upon you, while C++'s olymp frowns and shakes its heads:

This isn't valid C++. You'll need to `new` the hell out of this buffer, and free it at the end, or use a `std::vector`. Both would also inherently `memset` it to 0.

Oh, by the way, don't do that. Memory allocation is expensive, and you really don't need to reallocate every time `work` is called. Just give the class a member buffer of some fixed length (a page?), allocate that in the ctor, deallocate in dtor, and call `d_crc_impl.process_bytes()` repeatedly.

If you're in a hurry, go the `std::vector` route. It's ugly, and that's why we do it all over the place.

gr-digital/lib/crc32_bb_impl.cc
```diff
@@ -65,21 +65,40 @@ namespace gr {
...
+        memset(packed_buffer, 0, n_packed_length);
+        for (size_t bit = 0; bit < packet_length; bit++){
+          packed_buffer[bit/8] |= (in[bit] << (bit % 8));
```
marcusmueller: We should really have a VOLK kernel for this. **should.** Feels like I thought we had a "packer" kernel; @n-west ?

Nothing wrong with this code, though! 😄 

Thinking about that, we should probably just use a different CRC polynomial (this polynomial g(x) "interpolated" by doing g<sup>8</sup>(x), maybe?) and not pack before CRC. Anyway, that's math for another day. (Which doesn't mean I wouldn't encourage you to do that math or just try. Just saying my own brain's too mushy right now.)

gr-digital/lib/crc32_bb_impl.cc
```diff
@@ -65,21 +65,40 @@ namespace gr {
...
+
     int
     crc32_bb_impl::work(int noutput_items,
                         gr_vector_int &ninput_items,
                         gr_vector_const_void_star &input_items,
                         gr_vector_void_star &output_items) {
       const unsigned char *in = (const unsigned char *) input_items[0];
       unsigned char *out = (unsigned char *) output_items[0];
-      long packet_length = ninput_items[0];
+      size_t packet_length = ninput_items[0];
```
marcusmueller: :+1: replacing signed with unsigned integer types where right is definitely a good idea wherever one encounters an opportunity to do so. Reduces potential for bugs that the compiler would tell you about.

gr-digital/python/digital/qa_crc32_bb.py
```diff
@@ -193,6 +193,35 @@ def test_007_crc_equal (self):
...
+                unpack1,
+                sink1
+        )
+        self.tb.connect(
+            src,
+            blocks.stream_to_tagged_stream(gr.sizeof_char, 1, len(data), self.tsb_key),
+            unpack2,
+            crc_unpacked,
+            sink2
+        )
```
marcusmueller: pedantic, but: indents not identical in the two `connect` calls.

gr-digital/lib/crc32_bb_impl.cc
marcusmueller: Ah jetzt ja. (eine Insel.)

So, indeed, 

```
    boost::crc_optimal<32,
                       boost::crc_32_type::truncated_polynominal * boost::crc_32_type::truncated_polynominal *
                           boost::crc_32_type::truncated_polynominal * boost::crc_32_type::truncated_polynominal *
                           boost::crc_32_type::truncated_polynominal * boost::crc_32_type::truncated_polynominal *
                           boost::crc_32_type::truncated_polynominal * boost::crc_32_type::truncated_polynominal,
                       0xFFFFFFFF, 0xFFFFFFFF, true, true>
        padded_crc;
```

at least in my stupid test case works directly on unpacked data and delivers the same output as `crc_32_type` (which is what the crc32_impl uses) on packed data.

Why the ugly unrolled potentiation? The polynomial is a template argument, i.e. needs to be a compile-time constant. Prior to C++11, there's not `constexpr` to let me define a compile-time evaluated function, and prior to C++14, you couldn't use a `for` loop in that function.

gr-digital/python/digital/qa_crc32_bb.py
noc0lour: shouldn't copy stuff from other places with random indentation.

gr-digital/lib/crc32_bb_impl.cc
noc0lour: hm, yeah we could limit the unpacked operation to multiple of 8 items. then the call to process bytes on a single byte might work.

gr-digital/lib/crc32_bb_impl.cc
noc0lour: Hm, we could test that and throw? or use some less efficient algorithm? From below: if packet_length==crc_length && d_check this might actually do harm, but is this realistic?

noc0lour:
Alternatively we could mandate packed operation and remove the alternative to operate on unpacked data.

marcusmueller:
I'd really rather use a CRC that works on unpacked than that :)

marcusmueller:
wait, are we not setting an io multiple of 8 in unpacked mode?

noc0lour:
Yes we do, this is TSB mode, so we can actually rely on that?

marcusmueller:
damned if I knew what we can trust in TSB mode. So yeah, in TSB mode we can't simply let the scheduler guarantee that we get multiples of 8, since that is incompatible with the TSB contract. Sooooooo yeah, we need the ability to deal with a 35 bit unpacked packet. Can you try whether https://github.com/gnuradio/gnuradio/pull/1504#discussion_r148253777 works without the introduction of the "packed" 1/8-sized buffer?

noc0lour:
I'll put that into a unit test, I guess, and see if this fails or not.

marcusmueller:
You're a personal hero of mine :) at least for today

marcusmueller:
Hey! You promised us a unit test, and I pinky promise a speedy merge if you do that (I wouldn't know who I'd rather ask for testing than my personal CI hero, @noc0lour)

noc0lour:
Hm, I forgot this PR even exists... Now I have to figure out what I did there

gr-digital/lib/crc32_bb_impl.cc
noc0lour: Will this ever happen for a valid crc32 packet? If yes will it explode?

PR #1506, opened 2017-10-31
Python3

OfficialLucky:
octoberfest


n-west:
I added the invalid label to signal @digitalocean that this is spam according to hacktoberfest guidelines. On the off-chance it's not spam, merge away-- left it open just in case

mbr0wn:
No, it's not :) Thanks for flagging. I'll close it.

mbr0wn:
I meant "Yes it is". Or, "no, it's not valid". It's still spam.

PR #1509, opened 2017-11-03
Fixed 16-QAM decoding

gvanhoy:
This caused a decoding error with 16-QAM using the hard decoder. Symbol "10" always mapped to "0".


marcusmueller:
Hey, thanks for the fix! There was really a moment of disbelief when @mbr0wn and I read this. For the sake of simplicity, @mbr0wn is now cherry-picking your commit and will slightly reword the commit msg. 

Thank you  so much, and keep up that level of vigility.

mbr0wn:
Merged manually: 4b4d7b4b76b380cecddd95b44e994ad732411dcb

Had to fix the commit message. Thanks!

PR #1519, opened 2017-11-12
grc: Expose qt time sink stem plot option

dkozel:
Stem plots are available in the qt time sink already but not exposed through GRC. This adds a simple control.

I'm not sure if the control is better placed on the main tab or in config. Additionally there is a behavior difference between using this control and the middle click context menu option. The middle click both sets the plot type to stem and changes the point marker to circle. I can add the conditional behavior in the xml to override the marker type if stem is true, but since that is overriding the user's settings I wanted to ask for an opinion before doing that.

Current PR behavior:
no_circle

Usual context menu result (marker overridden to circle):
screenshot from 2017-11-12 23-47-11

### DIFF
--- a/gr-qtgui/grc/qtgui_time_sink_x.xml
+++ b/gr-qtgui/grc/qtgui_time_sink_x.xml
@@ -29,6 +29,7 @@ self.$(id).enable_autoscale($autoscale)
 self.$(id).enable_grid($grid)
 self.$(id).enable_axis_labels($axislabels)
 self.$(id).enable_control_panel($ctrlpanel)
+self.$(id).enable_stem_plot($stemplot)
 
 if not $legend:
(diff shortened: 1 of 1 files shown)


mbr0wn:
@dkozel's suggestions to alter the marker defaults are good, but can go into a separate PR.

dkozel:
This PR does not include appropriate `hide` logic. I'll update it tonight but keep the other changes Martin refers to in a separate PR since they're deeper code changes.

skoslowski:
Would you mind changing the tag in your commit message from grc to
qtguibwhile you are at it?

dkozel:
Hide added and the commit message edited. I believe this is ready for merge.

marcusmueller:
I believe that too. So I'll take action. Thanks!

PR #1522, opened 2017-11-15
Change input by output in brief descriptions of block.h

antonioramosdet:


marcusmueller:
Hm, this looks like a reasonable commit; why did you close it?

marcusmueller:
This is short enough to skip the CLA hassle, but in case you plan to contribute more in the future: get in touch with me or @bhilburn early, so that we can get the copyright signoff off the take quickly.
Thank you for this excellent fix!

PR #1527, opened 2017-11-19
docs: Add missing information on usage of produce()/consume()

rear1019:
See [1] for reference.

[1] [https://lists.gnu.org/archive/html/discuss-gnuradio/2017-11/msg00152.html](https://lists.gnu.org/archive/html/discuss-gnuradio/2017-11/msg00152.html)


noc0lour:
Looks good to me. :+1:

PR #1528, opened 2017-11-20
Making gr_modtool start faster by avoiding to iterate over all globals

marcusmueller:
All subclasses of object have a __subclasses__ method, which gives the
subclasses. That's a lot fewer objects than what's in globals().

### DIFF
--- a/gr-utils/python/utils/gr_modtool
+++ b/gr-utils/python/utils/gr_modtool
@@ -1,6 +1,6 @@
 #!/usr/bin/env python
 #
-# Copyright 2012 Free Software Foundation, Inc.
+# Copyright 2012, 2017 Free Software Foundation, Inc.
 #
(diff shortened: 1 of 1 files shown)


noc0lour:
lol'd at the branch name

marcusmueller:
for reference: [Python 2.7 docs](https://docs.python.org/2.7/library/stdtypes.html?highlight=__subclasses__#class.__subclasses__)

skoslowski:
...better yet, use explicit registration via a decorator =)

marcusmueller:
@skoslowski without having thought about that, I guess that `object` having a `subclasshook` that... registers subclasses is kind of the same. I'm daily reminded that "explicit is better than implicit" is a Pythonic mantra (someone put up a poster),  but I feel that inheritance is explicit in itself!

PR #1529, opened 2017-11-20
Adding readline capabilities to gr_modtool

marcusmueller:
* main advantage: better line editing
* tab-completion for gr_modtool add


noc0lour:
Looks good to me, steady improvement of gr_modtool 👍

PR #1530, opened 2017-11-21
grc: Fix YAML hier blocks (next)

haakov:
Switched to YAML for hierarchical blocks

### DIFF
--- a/grc/core/generator/top_block.py
+++ b/grc/core/generator/top_block.py
@@ -183,7 +183,7 @@ class TopBlockGenerator(object):
         blocks_make = []
         for block in blocks:
             make = block.templates.render('make')
-            if not block.is_variable:
+            if not (block.is_variable or block.is_virtual_or_pad):
                 make = 'self.' + block.name + ' = ' + make
             if make:
                 blocks_make.append((block, make))
--- a/grc/core/generator/flow_graph.py.mako
+++ b/grc/core/generator/flow_graph.py.mako
@@ -114,17 +114,16 @@ class ${class_name}(gr.hier_block2, Qt.QWidget):
 % else:
 class ${class_name}(gr.hier_block2):
 % endif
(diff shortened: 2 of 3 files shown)


grc/core/generator/top_block.py
```diff
@@ -234,15 +234,20 @@ def make_port_sig(port):
...
+        for connection in virtual_source_connections:
             sink = connection.sink_port
             for source in connection.source_port.resolve_virtual_source():
                 resolved = connection_factory(fg.orignal_flowgraph, source, sink)
                 connections.append(resolved)
             # Remove the virtual connection
             connections.remove(connection)
 
+        for connection in virtual_sink_connections:
+            connections.remove(connection)
```
haakov: This prevents lines like `self.connect((self.analog_sig_source_x_0, 0), (self.virtual_sink_0, 0))` from appearing in the Python files. Attempting to use connect() with a virtual source/sink creates an `AttributeError: 'top_block_sptr' object has no attribute 'virtual_sink_0'`.

marcusmueller:
so, following the discussion on #1886, we should probably be merging this on next, right? Or did we already?

haakov:
This hasn’t been merged on next yet. I actually don’t think this PR has been touched since I created it in November, so because of that I’d like a couple of days to make sure it works if that’s okay. I’ll notify you when it’s been tested!

haakov:
Okay so the first commit looks fine, but the one regarding virtual connections does not work with `next`. I'll have a look at it.

haakov:
@marcusmueller It looks like the virtual connections are going to take some time to figure out, so I removed the commit that didn't work and squashed the others. This way you can merge this PR whenever you'd like and don't have to wait for me to figure out how to fix the virtual connections :)

marcusmueller:
Thanks for the incredible patience! And, even more so, the incredible work :)

PR #1537, opened 2017-12-12
gnuradio-runtime: Optimize some buffer[_reader]_sptr grabs and releases

awalls-cx18:
Optimize some reference counted buffer[_reader]_sptr grabs and releases
that show up as minor CPU wasters in profiling when there are a lot of
tags or some blocks that "return 0" often.

At high sample rates (e.g. 160 Msps), this can save ~2% CPU on
blocks that propagate a fair number of tags.

### DIFF
--- a/gnuradio-runtime/lib/block_detail.cc
+++ b/gnuradio-runtime/lib/block_detail.cc
@@ -313,17 +313,19 @@ namespace gr {
       d_total_noutput_items = noutput_items;
       d_pc_start_time = (float)gr::high_res_timer_now();
       for(size_t i=0; i < d_input.size(); i++) {
-	gr::thread::scoped_lock guard(*d_input[i]->mutex());
-        float pfull = static_cast<float>(d_input[i]->items_available()) /
-          static_cast<float>(d_input[i]->max_possible_items_available());
+        buffer_reader_sptr in_buf = d_input[i];
+	gr::thread::scoped_lock guard(*in_buf->mutex());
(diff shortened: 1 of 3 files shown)


noc0lour:
Looks fine to me.

awalls-cx18:
On Sat, 2018-01-06 at 07:20 -0800, Andrej Rode wrote:
> Looks fine to me.

Note that this change is included as the first change in pull request
#1543 

https://github.com/gnuradio/gnuradio/pull/1543

noc0lour:
Even if #1543 isn't merged immediately we can probably accept this PR without further issues.

awalls-cx18:
On Sat, 2018-01-06 at 16:14 +0000, Andrej Rode wrote:
> Even if #1543 isn't merged immediately we can probably accept this PR
> without further issues.

Correct.

Just didn't know if there would be a merge conflict or not.  Probably
not.

marcusmueller:
This looks a) very benign, b) highly desirable and c) like something we should probably let become GNU Radio hivemind knowledge: If possible, avoid changing the refcount on sptrs by reusing them.

As I write this @mbr0wn's fixing the indents in your commit (not your fault, you use a sane setting, but we need to settle on a single style) and then we're good to go :)

mbr0wn:
Manually merged to eee29fbe8ff4c58ae68af89384328f23f0a8e7b6, some indents were broken.

noc0lour:
`git rerere` is your friend if @mbr0wn is using that.

Issue #1548, opened 2018-01-14
blocks: vector_sink reserve (min) data length in constructor


noc0lour:
In the `vector_sink` block data is allocated dynamically on every call. To improve performance a user could supply a minimum number of items he plans to stream into the vector sink, so memory is already allocated. Could improve speed during testing as well.

marcusmueller:
Idea: Make vector_sink have a "total number of items to store": Vector sink is in most cases used directly after a head block, and that's very a) annoying, b) a copying of perfectly fine samples and c) would be an excellent size to allocate first.

noc0lour:
So, basically have a `num_items` ?

lshariprasad:
Hi, I'm Hari Prasad L S, a GSoC 2026 applicant. I'd like to work on this issue and plan to submit a PR within 5 days.

marcusmueller:
Hi @lshariprasad, I'm afraid we forgot to close this issue after merging the feature (see the linked PR above).

So, there's nothing to do here.

Generally, you not seeing the feature yet means you haven't checked out the Vector Sink e.g. in GNU Radio companion. If that's because you're not already quite familiar with GNU Radio: Friendly reminder to get comfortable with the software before opening Pull Requests for it!

PR #1549, opened 2018-01-15
blocks: reserve memory in vector_sink constructor

noc0lour:
Might bring some performance benefits. Changes public interface to vector_sink.

### DIFF
--- a/gr-blocks/grc/blocks_vector_sink_x.xml
+++ b/gr-blocks/grc/blocks_vector_sink_x.xml
@@ -8,7 +8,7 @@
 	<name>Vector Sink</name>
 	<key>blocks_vector_sink_x</key>
 	<import>from gnuradio import blocks</import>
-	<make>blocks.vector_sink_$(type.fcn)($vlen, $min_items)</make>
+	<make>blocks.vector_sink_$(type.fcn)($vlen, $reserve_items)</make>
 	<param>
 		<name>Input Type</name>
 		<key>type</key>
@@ -47,7 +47,7 @@
 	</param>
 	<param>
 		<name>Minimum data items</name>
-		<key>min_items</key>
+		<key>reserve_items</key>
 		<value>100</value>
 		<type>int</type>
 	</param>
--- a/gr-blocks/grc/blocks_vector_sink_x.xml
+++ b/gr-blocks/grc/blocks_vector_sink_x.xml
@@ -46,9 +46,9 @@
 		<type>int</type>
 	</param>
 	<param>
-		<name>Minimum data items</name>
+		<name>Reserve memory for items</name>
 		<key>reserve_items</key>
-		<value>100</value>
+		<value>1024</value>
 		<type>int</type>
 	</param>
 	<check>$vlen &gt; 0</check>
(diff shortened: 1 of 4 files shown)


willcode:
How do we know whether blocks like vector_sink are subclassed by OOT projects? I'm asking mainly because I started working on file_source offset/len and was wondering the same thing for that block?

noc0lour:
We don't know and if a certain OOT uses a block. Though in this case by using a default constructor this shouldn't be the case. (Most certainly a lot of OOTs are using a vector_sink for QA).

noc0lour:
Adds features proposed in #1548

marcusmueller:
Hey, @noc0lour, so, this is a nice, short, clear-benefit PR, but:

This would do a lot better with a sensible default (maybe a page size. Who needs 100 samples?), and honestly: I think you should combine with the proposed *max* feature, as saving a gr::blocks::head definitely is as much a perf improvement.

noc0lour:
Will do that, The value is super application specific, so taking `100` was just merely a PoC than a `real` number for inclusion.

noc0lour:
Was looking at the `get_page` function in runtime, unfortunately this function is not in the public headers and I object to copying it or cross linking through the tree.

marcusmueller:
Tbh, I'll be content with you having `4096 / sizeof(element)` in the code and a comment that says `//4kB is a good assumption on page sizes`.

Other than that, if anything, `gnuradio-runtime` is the component that is OK to be cross-used all over the tree... but I agree, let's not unnecessarily expose inner API. However, that `gr::getpagesize` thing was definitely *meant* to be externally accessible, so I'm not that critical there. The `fprintf(err` would have to go, though...

mbr0wn:
Even a value like 1024 would be fine. I would suggest *not* combining this with a max feature, but rather make that a separate PR.

marcusmueller:
I thought about this. The name "minimum items" suggest something breaks when the user doesn't supply at least as many – and that's really not the case. Instead, I'd recommend a "Items of Memory Reservation" as GRC property name, and "reserve_items" as parameter name. With these wording corrections, I'd be willing to give this a rather immediate merge – there's nothing to be lost here, I guess.

noc0lour:
Added 1024 items and rename min_items -> reserve_items

gr-blocks/grc/blocks_vector_sink_x.xml
```diff
@@ -45,6 +45,12 @@
 		<value>1</value>
 		<type>int</type>
 	</param>
+	<param>
+		<name>Minimum data items</name>
```
marcusmueller: Should read `Reserve Memory for items` or something

gr-blocks/grc/blocks_vector_sink_x.xml
```diff
@@ -45,6 +45,12 @@
...
+		<key>reserve_items</key>
+		<value>100</value>
```
marcusmueller: 1024

marcusmueller:
Ah! I'm taking the freedom of updating the copyright year myself while merging, @noc0lour.

noc0lour:
sure, go ahead

marcusmueller:
Manually merged (b52726077244d2b8439c663144e5f013fa3d8557). Closing this PR! Thanks for the great work 👍

Issue #953, opened 2016-09-16
Make file_source take an offset and length to read


mbr0wn:
I'd like to be able to peek at small intervals from large recordings. I propose adding two parameters to file_source: offset and readlen

Offset n = 0 would preserve the current behavior; offset n > 0 would seek n items into the file (n_1 byte for chars, n_2 bytes for shorts, n*4 bytes for floats)
Readlen = 0 would preserve the current behavior; readlen n > 0 would effectively return a substring from the file (1 = give me 1 item, 1e6 = give me the next million items), then do whatever happens when EOF is reached; repeat or end.

I'd rather do this with a source block than by slicing up my source file with dd so that i can change the part of the file I'm reading at runtime, and so that I don't accidentally split and interesting signal in the middle.

mbr0wn:
Author: Nathan West
Date: Fri, 03 October 2014  8:59:35

mbr0wn:
Author: Chris Kuethe
Date: Fri, 03 October 2014 18:54:55

mbr0wn:
Author: Jeff Long
Date: Sat, 04 October 2014 16:32:58

mbr0wn:
Author: Chris Kuethe
Date: Sat, 04 October 2014 18:18:58

mbr0wn:
Author: Chris Kuethe
Date: Sat, 04 October 2014 22:24:31

mbr0wn:
Author: Kevin Reid
Date: Sun, 05 October 2014 14:26:16

mbr0wn:
Author: Jeff Long
Date: Sun, 05 October 2014 15:40:41

mbr0wn:
Author: Chris Kuethe
Date: Sun, 05 October 2014 16:45:39

mbr0wn:
Author: Chris Kuethe
Date: Sun, 05 October 2014 17:04:10

mbr0wn:
Author: Jeff Long
Date: Sun, 05 October 2014 17:35:42

mbr0wn:
Author: Chris Kuethe
Date: Sun, 05 October 2014 19:32:42

mbr0wn:
Author: Jeff Long
Date: Sun, 05 October 2014 21:47:51

mbr0wn:
Author: Chris Kuethe
Date: Mon, 06 October 2014  5:54:23

mbr0wn:
Author: Johnathan Corgan
[...]

PR #1550, opened 2018-01-16
Add offset and len to file source

willcode:
Addresses https://github.com/gnuradio/gnuradio/issues/953.

API change. New parameters have defaults (offset=0, len=0), where len==0 means "rest of file".


willcode:
Previous (and current) behavior was the throw an exception if a file could not be opened. This makes sense when a flowgraph is first started, but open() can also be called at run time. Crashing on a bad file name seems drastic in that case, but it would be an API/semantic change to return a status from open() to do otherwise. But, you could argue that the caller should expect something bad to happen if a bad file name is used.

Now, with the addition of offset/len, an exception is thrown based on range checks. There is no "correct" behavior if the block is asked to play samples that don't exist, but out-of-bounds offset/len seem more likely, especially since the caller has no way of knowing the file size, especially through GRC.

Opinions on how this should work? I didn't want to break the API more than necessary to add the feature. Should a file_source produce zeros if asked to do something impossible?

gr-blocks/lib/file_source_impl.cc
```diff
@@ -86,12 +89,32 @@ namespace gr {
     bool
     file_source_impl::seek(long seek_point, int whence)
     {
-      return fseek((FILE*)d_fp, seek_point *d_itemsize, whence) == 0;
+      seek_point += d_start_offset_items;
+
+      if (whence == SEEK_SET)
```
noc0lour: I'm a fan of using braces throughout the if/else clause if use them anywhere in a if/else clause. I know the GNU Radio Project has no strict guidelines for this.

gr-blocks/lib/file_source_impl.cc
```diff
@@ -118,15 +141,43 @@ namespace gr {
...
+      // Make sure there will be at least one item available
+      if ((file_size / d_itemsize) < (start_offset_items+1)) {
+        if (start_offset_items) {
+          GR_LOG_WARN(d_logger, "WARNING: file is too small for start offset");
+        }
+        else {
+          GR_LOG_WARN(d_logger, "WARNING: file is too small");
+        }
+        fclose(d_new_fp);
+        throw std::runtime_error("file size too small");
```
noc0lour: Currently throwing is the only viable option at this point.

gr-blocks/lib/file_source_impl.cc
```diff
@@ -179,47 +229,50 @@ namespace gr {
...
-          d_file_begin = true;
-          d_repeat_cnt++;
+        size_t nitems_to_read = std::min((size_t)size, d_items_remaining);
+
+        // Since the bounds of the file are know, unexpected nitems is an error
+	if (nitems_to_read != fread(o, d_itemsize, nitems_to_read, (FILE*)d_fp))
+          throw std::runtime_error("fread error");
+
+	size -= nitems_to_read;
+        d_items_remaining -= nitems_to_read;
```
noc0lour: looks like a whitespace error in the browser

noc0lour:
Also technically not your fault: I have a feeling the file we read from should be filled with data without the help of GNU Radio. Otherwise unlikely but possible: Bugs in file_sink/file_source might cancel out and we won't see tests failing. You could fill the file with samples in a `numpy.ndarray` you like using `numpy.ndarray.tofile()`.

willcode:
I think I addressed all your comments. The unit tests needed a lot of reworking - see commit log.

noc0lour:
Looks good to me. One thing I saw now as well: adding 2018 to the copyright headers of files you touched is would be nice as well.
Maybe someone else has also comments (: ?

willcode:
OK

noc0lour:
Not to be super picky, you should just add 2018 like `2012,2018` that's what I grasp from other headers and copyright boilerplate in the project.

willcode:
Updating copyrights on individual files doesn't scale. How about a common
license and a SPDX statement at the beginning of each file.

noc0lour:
I'm with you on this one. Will need discussion and then enforcement by the project as a whole.

mbr0wn:
Let's ignore the copyright discussion for now. I'm also a fan of SPDX headers.

gr-blocks/include/gnuradio/blocks/file_source.h
```diff
@@ -1,6 +1,6 @@
 /* -*- c++ -*- */
 /*
- * Copyright 2012 Free Software Foundation, Inc.
+ * Copyright 2018 Free Software Foundation, Inc.
```
marcusmueller: The important part here really is the *first* copyright date, and to a much lesser extent the last; so, never replace OLDYEAR with NEWYEAR, but just append (as @mbr0wn said)

gr-blocks/lib/file_source_impl.cc
```diff
@@ -1,6 +1,6 @@
 /* -*- c++ -*- */
 /*
- * Copyright 2012 Free Software Foundation, Inc.
+ * Copyright 2018 Free Software Foundation, Inc.
```
marcusmueller: Same: OLDYEAR, NEWYEAR, never only NEWYEAR

gr-blocks/lib/file_source_impl.cc
```diff
@@ -86,12 +89,35 @@ namespace gr {
...
+        seek_point = d_length_items - seek_point;
+      }
+      else {
+        GR_LOG_WARN(d_logger, "file_source: bad seek mode");
+        return 0;
+      }
+
+      if ((seek_point < (long)d_start_offset_items)
+          || (seek_point > (long)(d_start_offset_items+d_length_items-1))) {
+        GR_LOG_WARN(d_logger, "file_source: bad seek point");
```
marcusmueller: love your usage of the proper logger :) :+1:

gr-blocks/lib/file_source_impl.cc
```diff
@@ -100,33 +126,61 @@ namespace gr {
 
       // we use "open" to use to the O_LARGEFILE flag
       if((fd = ::open(filename, O_RDONLY | OUR_O_LARGEFILE | OUR_O_BINARY)) < 0) {
-	perror(filename);
-	throw std::runtime_error("can't open file");
+        perror(filename);
```
marcusmueller: nope, let's not do `perror` if we could also `GR_LOG_ERROR(d_logger, boost::format("file_source(%s): %s") % filename % strerror(errno))`, or so.

gr-blocks/lib/file_source_impl.cc
```diff
@@ -100,33 +126,61 @@ namespace gr {
...
-	d_new_fp = 0;
+        fclose(d_new_fp);
+        d_new_fp = 0;
       }
 
       if((d_new_fp = fdopen (fd, "rb")) == NULL) {
-	perror(filename);
-	::close(fd);	// don't leak file descriptor if fdopen fails
-	throw std::runtime_error("can't open file");
+        perror(filename);
```
marcusmueller: as said above: let's avoid logging to stderr by avoiding perror.

willcode:
Made changes, as requested.

marcusmueller:
@mbr0wn @noc0lour I'd rebase-merge this now, but you two asked for amendments, that I _think_ are in there now, but I'd still like your brief OK before I merge.

mbr0wn:
Merged and squashed manually: 6ae02d1d6d04a5f529ab3fea496c589f8d14229d

Also fixed a typo in a comment while I was at it.

Issue #1542, opened 2017-12-20
fastnoise_source uses non thread-safe rand() function


mgw93:
The fast noise source block uses the rand() function to sample the precalculated values.
As rand() is not thread safe, this leads to non-reproducible behaviour.
Also, the specified seed value is not used with srand.
It might be better to use the random number functions in boost or in C++11.

To easily show this behaviour, connect two fast noise sources with the same seed to a subtract block and plot the output.

PR #1562, opened 2018-01-27
replaced thread-unsafe rand() in fastnoise_source by XOROSHIRO128+

marcusmueller:
Closes gnuradio/gnuradio#1542.

### DIFF
--- a/gr-analog/lib/fastnoise_source_X_impl.cc.t
+++ b/gr-analog/lib/fastnoise_source_X_impl.cc.t
@@ -27,6 +27,7 @@
 #endif
 
 #include "@IMPL_NAME@.h"
+#include <gnuradio/xoroshiro128p.h>
 #include <gnuradio/io_signature.h>
 #include <stdexcept>
 
@@ -46,13 +47,13 @@ namespace gr {
 		    io_signature::make(1, 1, sizeof(@TYPE@))),
       d_type(type),
 #if @IS_COMPLEX@	// complex?
-      d_ampl(ampl/sqrtf(2.0f)),
+      d_ampl(ampl/sqrtf(2.0f))
 #else
-      d_ampl(ampl),
+      d_ampl(ampl)
 #endif
-      d_rng(seed)
     {
       d_samples.resize(samples);
+      xoroshiro128p_seed(d_state, (uint64_t) seed);
       generate();
     }
 
@@ -144,30 +145,21 @@ namespace gr {
 
     @TYPE@ @IMPL_NAME@::sample()
     {
-#ifdef HAVE_RAND48
-        size_t idx = lrand48() % d_samples.size();
-#else
-        size_t idx = rand() % d_samples.size();
-#endif
+        size_t idx = xoroshiro128p_next(d_state) % d_samples.size();
         return d_samples[idx];
     }
 
-#ifndef FASTNOISE_RANDOM_SIGN
-#ifndef HAVE_RAND48
-#define FASTNOISE_RANDOM_SIGN       ((rand()%2==0)?1:-1)
-#else
-#define FASTNOISE_RANDOM_SIGN       ((lrand48()%2==0)?1:-1)
-#endif
-#endif
-
     @TYPE@ @IMPL_NAME@::sample_unbiased()
     {
+        uint64_t random_int = xoroshiro128p_next(d_state);
 #if @IS_COMPLEX@
         gr_complex s(sample());
-        return gr_complex(FASTNOISE_RANDOM_SIGN * s.real(),
-                          FASTNOISE_RANDOM_SIGN * s.imag());
+        float re = (random_int & (UINT64_C(1)<<23)) ? (- s.real()) : (s.real());
+        float im = (random_int & (UINT64_C(1)<<42)) ? (- s.real()) : (s.real());
+        return gr_complex(re, im);
 #else
-        return FASTNOISE_RANDOM_SIGN * sample();
+        float s = sample();
+        return (random_int & (1<<23)) ? (-s) : s;
 #endif
     }
 
--- a/gr-analog/python/analog/qa_fastnoise.py
+++ b/gr-analog/python/analog/qa_fastnoise.py
@@ -20,7 +20,8 @@
 # Boston, MA 02110-1301, USA.
 #
 
-from gnuradio import gr, gr_unittest, analog
+from gnuradio import gr, gr_unittest, analog, blocks
+import numpy
 
 class test_fastnoise_source(gr_unittest.TestCase):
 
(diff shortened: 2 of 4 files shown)


marcusmueller:
They are the best kind of not flaky:

*Stochastically* not flaky. Which you would know if you sat in my Stoch101 Exercise last week. Tschebyscheff inequality bounds P(failure of unittest) pretty directly when your unittest averages a million samples to measure their mean and compare it to the right value (same goes for variance). 

So, now I'll start a discussion that I should be having with Martin (and his unit tests for gr-digital): 

There's of course a non-zero P that they'll fail. That is, if these values would be random.

They're not random. This is a pseudo-random number generator. It's designed to give exactly the same sequence of numbers given the same seed. And with the removal of global state (which came in through the abuse of `rand()`), there's nothing that can change that.

So, unless someone touches the code of the RNG, it will always not fail. And there's nothing flaky about that, unless computers are flaky. If you ran fastnoise_source->file_sink on two completely different computers and compared the files, they should be identical.

So, with this thing behaving correctly, I could just as well generate lets say 1000 values now and just run a unit test whether the generated values are identical.

But, considering the chance that someone *does* touch the code, or something in a compiler breaks, I've devised a unit test that checks the stochastic properties of the generated number sequence. So, people are now free to change the code, and get hit on the head when something actually breaks what the noise source was made for: generating random numbers of given distribution.

What one should much rather ask is "how likely is that the test passes but the noise source is actually broken" and the answer is that it's "damn likely" (and that's the mathematically correct way of measuring that probability), but "damn less likely than before", where the unit test actually only tested instantiability and whether a getter was able to get a member variable.

TL;DR,  math || GTFO:

PRNG is deterministic. It works here. Thus:

P(fail | computation environment sane ∩ PRNG code unchanged) = 0 and

P(fail | PRNG code changed ∩ PRNG still correct) = ε ≪ 1

and that's the reason I wrote the test like this. If someone changes the PRNG and sees the unit test fail, they'd be in unlikely bad luck, but hopefully be able to figure that out.

1 > P(passes | PRNG broken ∩ new unit test) ≫ P(passes | PRNG broken ∩ old unit test) > 0

because

P(passes | PRNG broken ∩ old unit test) = P(passes | PRNG broken) · P(passes | old unit test) 

as the old unit test didn't give a hoot about the numbers generated.

noc0lour:
Then don't put `actually kind of` in the commit description. Looked through the unit test and it's sound.

gr-analog/lib/fastnoise_source_X_impl.cc.t
```diff
@@ -88,37 +89,68 @@ namespace gr {
...
+          uint64_t rand  = xoroshiro128p_next(d_state);
+          d_samples[i] = uint64_to_f(rand);
+        }
 	break;
 
       case GR_GAUSSIAN:
-	for(int i = 0; i < noutput_items; i++)
-	  d_samples[i] = (@TYPE@)(d_ampl * d_rng.gasdev());
+        for(int i = 0; i < noutput_items; i++) {
+          d_samples[i] = xoroshiro128p_cltf(d_state, 32);
```
noc0lour: Specific target/reason to pick 32 here? It's fine, just curious why..

marcusmueller:
So, what do you want me to do? Claim that this test actually tests a significant part of the features? It does not.

gr-analog/lib/fastnoise_source_X_impl.cc.t
marcusmueller: Three possible answers:

* The plot from [my xoroshiro OOT](https://github.com/marcusmueller/gr-xoroshiro/blob/master/examples/Kolmogorov-Smirnov-Test.ipynb) suggests 32 is kind of a sweet spot
* The Berry-Esseen Theorem bounds the Kolmogorov-Smirnov distance to ca 1؉ for an average of 32 i.i.d. U(0,1) variables
* Gut feeling.

noc0lour:
so this might describe your changes better `added unit test for stochastic properties to fastnoise source`  ?

marcusmueller:
`…test for statistical properties…`, to nitpick, but maybe, yes.

Hm, doesn't @jmcorgan squash all the PRs without retaining the git commit messages, anyway?

noc0lour:
Usually not, I think. So please continue to use good commit messages (:

marcusmueller:
I'll try to do my very best! But: changing the commit message now would require a force-push. Are you OK with that?

noc0lour:
Why do I have to be OK? it's your branch :) :+1:

marcusmueller:
Well, it will break causality of every review and comment you made on this branch, and you've been tremendously helpful, so I'd rather ask :) doing this in a minute.

gnuradio-runtime/include/gnuradio/xoroshiro-variates.h
```diff
@@ -0,0 +1,139 @@
+/*
+  Copyright (C) 2017 Marcus Müller
```
mbr0wn: Did you copy this from another of your projects?

gr-analog/lib/fastnoise_source_X_impl.cc.t
mbr0wn: Still, do you want to make it a local `constexpr` instead of a magic number? Those comments you just posted could also be code comments.

gnuradio-runtime/include/gnuradio/xoroshiro-variates.h
marcusmueller: yes.

Um, stupid question, so I have a CLA that says that code contributed to the GR project is inherently copyright-assigned to the FSF. So, no matter what, I'll have to change this line.

But: my other project is LGPL; I'd rather not dual-license it, but then again my guess is that I can't just "work around" the fact that GNU Radio is GPL by first having the same code in a toy python module elsewhere, right?

gr-analog/lib/fastnoise_source_X_impl.cc.t
marcusmueller: Definitely, @mbr0wn. That's how it should've been done.

gnuradio-runtime/include/gnuradio/xoroshiro-variates.h
mbr0wn: If this is all your code, you can simply assign everything to FSF same with any other code you contribute to GNU Radio. That doesn't affect your own projects.

mbr0wn:
For better reviewability, it might have been a good idea to first add the new code, then have a separate PR that uses it.

marcusmueller:
Agreeing on that. Will rework this PR.

marcusmueller:
Completely overhauling this PR: Now *only* the bug fix, with separate QA introducing, include-adding and bug-fixing commit.

The improvements in terms of replacing the current (slow) random with the same RNG (xoroshirop128+) used to do the random number pool generation have been removed. Might be worth doing that on master and in the shape of replacing boost::mt19937 alltogether in gr::math::random.

marcusmueller:
@noc0lour could you please re-review? Thank you?

gr-analog/python/analog/qa_fastnoise.py
```diff
@@ -30,21 +31,58 @@ def setUp (self):
     def tearDown (self):
         self.tb = None
 
-    def test_001(self):
-        # Just confirm that we can instantiate a noise source
-        op = analog.fastnoise_source_f(analog.GR_GAUSSIAN, 10, 10)
+    def test_001_test_moments(self):
```
noc0lour: In a perfect world one would have a unit test for each type of source or at least one for real and one for complex sources. But this is fine for now.

Issue #1561, opened 2018-01-27
gr::random::laplacian, ::rayleigh and ::impulse use `log` where they should use `logf`


marcusmueller:
performance issue, essentially.

PR #1564, opened 2018-01-28
Use type-generic or float versions of cmath functions where appropriate.

marcusmueller:
Complements gnuradio/gnuradio#1563, larger-scope version of
the issue gnuradio/gnuradio#1561 .

### DIFF
--- a/gnuradio-runtime/include/gnuradio/fxpt.h
+++ b/gnuradio-runtime/include/gnuradio/fxpt.h
@@ -53,7 +53,7 @@ namespace gr {
       float_to_fixed(float x)
     {
       // Fold x into -PI to PI.
-      int d = (int)floorf(x/TAU+0.5);
+      int d = (int)std::floor(x/TAU+0.5);
       x -= d*TAU;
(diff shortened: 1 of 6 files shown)


gnuradio-runtime/include/gnuradio/fxpt.h
```diff
@@ -45,15 +45,16 @@ namespace gr {
...
     static const float TWO_TO_THE_31;
 
   public:
     static int32_t
       float_to_fixed(float x)
     {
       // Fold x into -PI to PI.
-      int d = (int)floor(x/2/PI+0.5);
-      x -= d*2*PI;
+      int d = (int)floorf(x/TAU+0.5);
```
noc0lour: I'm wondering if `std::floor` isn't essentially the same or better?

marcusmueller: well, pretty much everywhere else I used `std::<math function>`. Here I didn't; had a reason, but it was yesterday night and I can't remember. Lemmechecksomething.

PR #1566, opened 2018-01-29
cmake: use gnu${CMAKE_C_STANDARD} for cmake version < 3.1

### LINKED ISSUE #1565: cmake: cmake_c_flags are set wrong on next for ubuntu 14.04
Apparently this: 
```
# Configure C++ and C standards
IF(CMAKE_VERSION VERSION_LESS "3.1")
        SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11")
    ELSEIF(CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
        SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11")
    ELSEIF(CMAKE_CXX_COMPILER_ID STREQUAL "MSVC")
        SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11")
    ELSE()
        MESSAGE(warning "C++ standard could not be set because compiler is not GNU, Clang or MSVC.")
    ENDIF()

    IF(CMAKE_C_COMPILER_ID STREQUAL "GNU")
        SET(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -std=gnu99")
    ELSEIF(CMAKE_C_COMPILER_ID STREQUAL "Clang")
        SET(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -std=gnu99")
    ELSEIF(CMAKE_C_COMPILER_ID STREQUAL "MSVC")
        SET(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -std=c11")
    ELSE()
        MESSAGE(warning "C standard could not be set because compiler is not GNU, Clang or MSVC.")
    ENDIF()
ELSE()
    SET(CMAKE_C_STANDARD 99)
    SET(CMAKE_CXX_STANDARD 11)
ENDIF()
```
has no effect on `Ubuntu 14.04` and leads to: 
```
-- Compiler Flags: /usr/lib/ccache/cc:::-O3 -DNDEBUG  -std=c11 -fvisibility=hidden -Wsign-compare -Wall -Wno-uninitialized
/usr/lib/ccache/c++:::-O3 -DNDEBUG  -fvisibility=hidden -Wsign-compare -Wall -Wno-uninitialized
```

No idea what is happening there.... Maybe the `VERSION_LESS` with "3.1" isn't working and it has to be "3.1.0". But that's clearly a problem for cmake 2.8 people to solve.

noc0lour:
closes #1565

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -157,7 +157,7 @@ ENDIF()
 
 # if cmake version is < 3.1, explicitly set C standard to use.
 IF(${CMAKE_VERSION} VERSION_LESS "3.1")
-    SET(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -std=c${CMAKE_C_STANDARD}")
+    SET(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -std=gnu${CMAKE_C_STANDARD}")
 ENDIF()
 
 ########################################################################


mbr0wn:
I'd rather do the opposite: Disable gnu extensions, always. New CMake is buggy in that respect.

marcusmueller:
So, as this is a non-fix, and we should really rather go ahead and replace/redefine that single constant (`M_SQRT`), we're closing this.

noc0lour:
K, so we have to find a way to fail the build on newer cmake by telling it not to use `gnu` always and then fix it for all cmake versions altogether?

Issue #1568, opened 2018-01-30
[Socket PDU]: Socket connections not destroyed when closed by client 


juan0fran:
Hi, in gr-blocks at socket_pdu_impl, it seams that no-socket is destroyed in TCP-Server mode when the connection is closed by the client. It can be easily seen by creating the Socket PDU and connect-close the socket with an external application. After 1128 (in my case) of connect-close actions, the socket pdu throws boost::system::system_error(error); with error code = 24, that is:

- asio::error::no_descriptors 24

willcode:
At first glance, it looks like newly accepted connections are added to d_tcp_connections and there is no way for the reference to be released. I can take a look at it if the fix isn't obvious to someone else.

willcode:
socket_pdu_impl keeps a list of (reference to) tcp_connections so it can broadcast on the send side. But, tcp_connection handles the recv side, and also knows when the connection closes. There is no way currently for socket_pdu_impl to know the connection has closed, so it maintains the reference, which keeps the tcp_connection destructor from getting called. I think it will take an API change to fix this. Nothing else in-tree is using tcp_connection, so changing its constructor to take a "closing" callback would probably not have much impact. I spent a bit of time looking through boost::asio to find a way to set a "closing" callback at that level (that could be set up during accept(), but found nothing. Opinions?

mbr0wn:
@willcode does this get fixed by #1570?

willcode:
Yes - the pull was a response to this issue.

marcusmueller:
In that case, closing this issue as fixed by #1570.

PR #1570, opened 2018-02-01
socket_pdu_impl: garbage collect closed tcp_connections

willcode:
This patch addresses two problems. First tcp_connections were not being
closed, leading to a bug report about socket exhaustion [1]. Second,
socket_pdu_impl tracks tcp_connections in a vector so that it can
broadcast outgoing data to clients. Items in this vector were never
freed.

This patch does a shutdown and close on the tcp_connection when there
is a read failure (signaling a socket close). On each new accept, the
tcp_connection list in socket_pdu_impl is scanned for closed connections,
which are then removed.

Possible improvements:
- Make sure the read error is eof.
- Check sends as well as reads.

More direct signalling could be done between tcp_connection and
socket_pdu_impl, but that would require an API change. No change
is required, here.

[1] https://github.com/gnuradio/gnuradio/issues/1568

### DIFF
--- a/gr-blocks/lib/socket_pdu_impl.cc
+++ b/gr-blocks/lib/socket_pdu_impl.cc
@@ -184,6 +184,15 @@ namespace gr {
     socket_pdu_impl::handle_tcp_accept(tcp_connection::sptr new_connection, const boost::system::error_code& error)
     {
       if (!error) {
+        // Garbage collect closed sockets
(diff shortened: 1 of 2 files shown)


willcode:
There is a potential race accessing d_tcp_connections between handle_tcp_accept (io_socket thread) and tcp_server_send (scheduler thread). This wasn't exposed before because the send side would fail silently on closed sockets.

PR #1571, opened 2018-02-01
gr-digital: The usage of 'and', 'or' and 'not' logical statements breaks compilation under MSVC

ptrkrysik:
i.e. with Geof Nieboer's scripts.

Replacing them with '&&', '||' and '!' respectively.

### DIFF
--- a/gr-digital/lib/clock_tracking_loop.cc
+++ b/gr-digital/lib/clock_tracking_loop.cc
@@ -251,7 +251,7 @@ namespace gr {
     void
     clock_tracking_loop::set_nom_avg_period(float period)
     {
-        if (period < d_min_avg_period or
+        if (period < d_min_avg_period ||
             period > d_max_avg_period   ) {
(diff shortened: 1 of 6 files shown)


noc0lour:
Testing failures are due to other bugs not related to this change.

ptrkrysik:
Not supporting them is MSCV's 'feature'. Even more weird, as someone
pointed on the #gnuradio IRC channel, is that after disabling MS
"language extensions"
(https://msdn.microsoft.com/en-us/library/0k0w269d.aspx) MSVC compiles
these constructs just fine. I don't know if it compiles whole GNU Radio
correctly with this option though because I didn't try it.

mbr0wn:
In any case this change is harmless and I see no harm merging.

marcusmueller:
We're merging this, changing the commit msg on the way to be action-describing.

marcusmueller:
in there as a212beb816781416d24d1c2b7ea31265ad702f33

PR #1572, opened 2018-02-01
cleanup: remove dynamic exception specifications

noc0lour:
Dynamic exception specifications are deprecated in C++11. Definining function with `noexcept()` is the way to go from C++11 on. `throw()` will be redefined to `noexecept(true)` in C++17.

We only have to define `noexcept(false)` for:
 - destructors
 - {default,copy,move} constructors
 - {copy,move} assignment operators
 - deallocation functions

Otherwise all other functions in C++ are potentially throwing by definition.
I wasn't sure about the swig templates, so I replaced throw(`something`) with noexcept(false) to give swig a hint that those functions are potentially throwing.

### DIFF
--- a/gnuradio-runtime/swig/hier_block2.i
+++ b/gnuradio-runtime/swig/hier_block2.i
@@ -34,7 +34,7 @@ namespace gr {
   make_hier_block2(const std::string name,
                    gr::io_signature::sptr input_signature,
                    gr::io_signature::sptr output_signature)
-  throw (std::runtime_error);
+    noexcept(false);
 }
 
 // Rename connect and disconnect so that we can more easily build a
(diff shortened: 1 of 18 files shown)


noc0lour:
So, something is wrong with `qa_hierblock2`, maybe a side effect caused by replacing the `throw()` clauses.

noc0lour:
Apparently the `throw($type)` specification was propagating the `$type` error to python. This now changes to `RuntimeError` in this case which resembles the error type in `C++`.

marcusmueller:
With me and @mbr0wn both approving, and now that you've fixed the qa_ issue: merging this. Keep up the good work, and thank you, @noc0lour!

PR #1581, opened 2018-02-06
cmake: Next stab at fixing the cmake_c_standard issue

noc0lour:
First part: convert viterbi.c to valid C++ and add required MATH constants to gnuradio/math.h.
Second part: set `-std=c{STANDARD}` flags. Though I don't know if that's correct for MSVC, need to reconfirm that.


noc0lour:
Ok, now only flaky zeromq & volk with GCC 7.2 are failing tests...

CMakeLists.txt
```diff
@@ -134,32 +134,29 @@ ENDIF()
...
 
 IF(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
-    SET(CMAKE_CXX_STANDARD 11)
+    SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++${GR_CXX_STANDARD}")
 ELSEIF(CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
-    SET(CMAKE_CXX_STANDARD 11)
+    SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++${GR_CXX_STANDARD}")
 ELSEIF(CMAKE_CXX_COMPILER_ID STREQUAL "MSVC")
-    SET(CMAKE_CXX_STANDARD 11)
+    SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++${GR_CXX_STANDARD}")
```
marcusmueller: I'll really need an explanation why setting this ourselves is better than letting CMake handle this? I know you're not doing it for the fun of it, but at this point, my understanding of the problem just doesn't explain why *not* setting the `CMAKE_CX*_STANDARD` is better here.

gr-fec/lib/ccsds_encoder_impl.cc
```diff
@@ -28,9 +28,7 @@
 #include <gnuradio/fec/generic_encoder.h>
 #include <cstdio>
 
-extern "C" {
```
marcusmueller: um, wut? viterbi.h pretty much starts out with a `namespace` statement. Thanks for spotting this one!

gr-fec/include/gnuradio/fec/viterbi.h
```diff
@@ -24,30 +24,40 @@
...
 
-struct viterbi_state {
-  unsigned long path;	/* Decoded path to this state */
-  long metric;		/* Cumulative metric to this state */
-};
+namespace gr {
+  namespace fec {
+
+    struct FEC_API viterbi_state {
+      unsigned long path;   /* Decoded path to this state */
```
marcusmueller: Reading the "it fits so nicely into a 32-bit word" comment above: this should explicitly be an `uint32_t`.

marcusmueller: actually, since we have him, could @SpectreJan comment on what the appropriate type sizes would be here.

marcusmueller: What I take away, `sizeof(metric)` should be `sizeof(**mettab)`, and reading decode.cc:

<pre><font color="#4E9A06"><b>lib/viterbi/decode.cc</b></font>
<font color="#C4A000"><b>48</b></font>:  int <span style="background-color:#C4A000"><font color="#2E3436">amp</font></span> = 100;
<font color="#C4A000"><b>52</b></font>:  gen_met(mettab, <span style="background-color:#C4A000"><font color="#2E3436">amp</font></span>, esn0, 0.0, 4);
<font color="#C4A000"><b>67</b></font>:      unsigned char sym = syms[i] &gt; 0 ? -<span style="background-color:#C4A000"><font color="#2E3436">amp</font></span> : <span style="background-color:#C4A000"><font color="#2E3436">amp</font></span>;
</pre>

I'd argue that `amp` should be uchar (or, while we're at it, `uint8_t`), generally.

gr-fec/lib/ccsds_encoder_impl.cc
```diff
@@ -28,9 +28,7 @@
 #include <gnuradio/fec/generic_encoder.h>
 #include <cstdio>
```
marcusmueller: as far as I can see, we should kill this `#include`, since it's not used within this file. Not your job (within this PR); but I certainly don't mind if you do.

CMakeLists.txt
noc0lour: see the discussion with @mbr0wn about cmake setting `gnu$version` flags, which is not really helpful.

gr-fec/lib/ccsds_encoder_impl.cc
noc0lour: I added the namespace to convert it to real C++...

gr-fec/include/gnuradio/fec/viterbi.h
SpectreJan: I would need to get back into the algorithm and implementation. At first glance uint32_t makes sense, since everything seems to be hardcoded for the 32bit words. But at this point I am not sure enough to make recommendation.

SpectreJan: As for the second...yes uchar makes much more sense here. But seein as line 67 has a fixme comment. Maybe we can get two birds with one stone?

marcusmueller:
@mbr0wn just hypothesized that CMake uses the wrong flags when setting CMAKE_C_STANDARD, i.e. that it

Luckily, he's wrong there:

 https://github.com/Kitware/CMake/blob/master/Modules/Compiler/GNU-C.cmake

https://github.com/Kitware/CMake/blob/v3.1.0/Modules/Compiler/GNU-C.cmake

--std=gnu11 is only used when extensions are enabled, even with CMake 3.1 which afaik introduced CMAKE_C_STANDARD.

noc0lour:
Hm, but do you have an explanation why the use of `M_SQRT2` did break on `cmake` < 3.1 where we set `-std=c$version` but everything is fine on `cmake` >= 3.1 where `cmake_c_standard 11` is used?

marcusmueller:
Wish I had! Probably something somewhere enables the use of aforementioned extensions. We need to track that down, and hurt it really badly.

marcusmueller:
I've tried to, but my smart phone is really ill-equipped to do that. So, shall we meet up and do that together later?

marcusmueller:
We've found a resolution. After extensive `cmake --trace`ing, reading CMake distro Modules and quite a bit of headscratching, we followed the old wisdom of @zod, only that we didn't `grep 2`, but `ag gnu11` in a build directory, found the CMake property .rst documentation that is somewhy copied into to `build/CMake/Help/`, and then found:

> C_EXTENSIONS
> ------------
> 
> Boolean specifying whether compiler specific extensions are requested.
> 
> This property specifies whether compiler specific extensions should be
> used.  For some compilers, this results in adding a flag such
> as ``-std=gnu11`` instead of ``-std=c11`` to the compile line.  This
> property is ``ON`` by default. The basic C standard level is
> controlled by the :prop_tgt:`C_STANDARD` target property.
> 
> See the :manual:`cmake-compile-features(7)` manual for information on
> compile features and a list of supported compilers.
> 
> This property is initialized by the value of
> the :variable:`CMAKE_C_EXTENSIONS` variable if it is set when a target
> is created.
> 

So, @noc0lour disabled the variable, and now things build without the compiler-specific extensions.

gr-fec/lib/viterbi/metrics.c
```diff
@@ -39,6 +39,8 @@
  */
 #define	OFFSET	128
 
+#include<gnuradio/math.h>
```
mbr0wn: Add space please `#include <gnuradio/math.h>`

gr-fec/lib/viterbi/metrics.c
```diff
@@ -51,76 +53,81 @@ extern double erf(double x);
...
-
-  for(s=1;s<255;s++){
-    /* P(s|1), prob of receiving s given 1 transmitted */
-    p1 = normal(((s-OFFSET+0.5)/amp - 1)/noise) -
-      normal(((s-OFFSET-0.5)/amp - 1)/noise);
-
-    /* P(s|0), prob of receiving s given 0 transmitted */
-    p0 = normal(((s-OFFSET+0.5)/amp + 1)/noise) -
-      normal(((s-OFFSET-0.5)/amp + 1)/noise);
+    void
```
mbr0wn: Argh fixing whitespace in the same commit makes reading reviews hard. We should really do a mega-whitespace fix soon.

gr-fec/lib/viterbi/viterbi.c
```diff
@@ -86,270 +86,275 @@
...
-    for(i=7;i>=0;i--){
-      encstate = (encstate << 1) | ((*data >> i) & 1);
-      *symbols++ = Partab[encstate & POLYA];
-      *symbols++ = Partab[encstate & POLYB];
+namespace gr {
+  namespace fec {
+
+    /* Convolutionally encode data into binary symbols */
+    unsigned char
+      encode(unsigned char *symbols,
```
mbr0wn: This looks like an auto-indenter went wrong somewhere.

mbr0wn:
I'm a bit surprised at the lengths you eventually had to go to... 😮

gnuradio-runtime/include/gnuradio/math.h
```diff
@@ -31,6 +31,9 @@
 #include <gnuradio/api.h>
 #include <gnuradio/gr_complex.h>
 
+#define M_SQRT2	1.41421356237309504880	/* sqrt(2) */
```
mbr0wn: Seems safer to add `#ifndef M_SQRT2` here, in case we run into more non-standard behaviour elsewhere.

noc0lour: Was thinking about that, but thought this shouldn't break, ever. My compiler didn't even complain about redefining this, when running with `-std=gnu(++)11`, or am I wrong?

mbr0wn: Let me quote Martin's law: "If you say it should, it won't". No harm in adding those lines.

marcusmueller: Hm, don't fully agree with here. If this **redefines** (as in: sets a _different_ value) `M_SQRT2`, I'd rather have *my* my √2 in my tests than some weird other people's not-quite-√2. If I was to feel paranoid about anything, it would be mathematical inconsistency, and then I'd favor @noc0lour's solution.

marcusmueller: notice that this doesn't express any opinion of me as a maintainer. I don't really feel worried about it, either way. I see that @mbr0wn's solution will probably save us a warning on `${strange platform}`, but that @noc0lour is closer to what one mathematically wants. So: `\shrug`.

mbr0wn: Our comments are orthogonal. You can #ifdef FOO -> #undef FOO, or you can #ifndef FOO ->#define FOO. In any case, you're adding a check for previous existence instead of just running `#define` and hoping for the best.

Note that `M_SQRT2` is a not-quite-but-almost-standard symbol. Things would be different if we were using our own symbols (`gnuradio::math::SQRT2` or something like that). Let's say you're redefining `M_PI`, which (I think) is always defined in math.h. Even if your definition is "better" than the original one, you could end up with mathematical inconsistencies between your code and other libs that use `M_PI` because you're redefining it in the middle of a source code flow.

noc0lour: Ok then let's define `GR_M_*` to save us hassle to put in ifdef guards for every constant we might or might not define?

mbr0wn: Perfect.

noc0lour: I was looking at other math defines e.g. `M_PI` which is used all over the place. It isn't defined in neither the c++11 nor the c11 standard. Yet it all works out?

Quick testing with `C++11` vs `C11` reveals that at least for `C++` I don't need to define `M_PI`. For `C` I do otherwise it doesn't know of `M_PI`. Both compiling with `-std=c{++}11`. I'm pretty sure I just have to include `<cmath>` here and be done, without redefining anything....

CMakeLists.txt
```diff
@@ -158,6 +161,7 @@ ENDIF()
 # if cmake version is < 3.1, explicitly set C standard to use.
 IF(${CMAKE_VERSION} VERSION_LESS "3.1")
     SET(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -std=c${CMAKE_C_STANDARD}")
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++${CMAKE_CXX_STANDARD}")
```
jdemel: This `set` should be capitalized.

CMakeLists.txt
```diff
@@ -378,19 +382,6 @@ GR_REGISTER_COMPONENT("testing-support" ENABLE_TESTING
...
-
-if(CPPUNIT_FOUND AND NOT "${CPPUNIT_VERSION}" VERSION_LESS "1.14.0")
-    message(WARNING "\nWarning: CppUnit version is ${CPPUNIT_VERSION} which requires C++11 for building. Trying to set CMake internally to use C++11 ...")
-    SET(CMAKE_CXX_STANDARD 11)
-endif()
-
-# if cmake version is < 3.1, explicitly set C++ standard to use.
-
-if(${CMAKE_VERSION} VERSION_LESS "3.1")
-    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++${CMAKE_CXX_STANDARD}")
```
jdemel: Capitalize for consistency

noc0lour: Err, that's a removed `set` (:

CMakeLists.txt
noc0lour: :+1: thanks, right :)

CMakeLists.txt
jdemel: It's hard to distinguish red and green sometimes. There are a few more places where my comment applies, though.

noc0lour: Only one `set(` was in my PR. I pushed a seperate commit fixing every other `set(` in the file for sake of consistency. "Current" cmake coding style for a copule of years now is having everything lowercase though. Emacs users even have a "unscreamify" macro :)

gr-fec/lib/viterbi/metrics.c
noc0lour: I think most of the whitespacing "issue" stems from inserting the full block into a namespace and then I just did a autoindent to have them in the correct indentation...

gnuradio-runtime/include/gnuradio/math.h
noc0lour: My gut feeling tells me, there is still something wrong, I pushed changes with `GR_M_{sqrt2,log2e}`. But other constants `M_` remain and everything builds correctly with `cmake_c{xx}_extensions` turned off.

noc0lour:
I discovered some strange compiler behaviour (with GCC 6.4.0/clang 5.0.1):

with `math_test.c` as follows
```
#include <math.h>

int main(){
  return M_PI;
}
```

gcc:
```
╰─ /usr/bin/gcc -std=c11 math_test.c -o math_test                                                                                                                                     1 ↵  14:30:52 
math_test.c: In function ‘main’:
math_test.c:4:10: error: ‘M_PI’ undeclared (first use in this function)
   return M_PI;
          ^~~~
math_test.c:4:10: note: each undeclared identifier is reported only once for each function it appears in
```

g++:
```
╭─ ~/test  
╰─ g++ -std=c++11 math_test.c -o math_test                                                                                                                                            1 ↵  14:31:26 
╭─ ~/test  
╰─ ./math_test                                                                                                                                                                          ✓  14:31:28 
```

marcusmueller:
Yay. OK, we can either now take the time to figure out wth math.h does
if included in C++11 mode, or we just recognize that we probably
shouldn't be trusting M_SQRT2 to be uniform across all compilers and
modes that we might expect it to exist (a space that's actually not
that small!), and go for the `#define GR_…` solution.

I'd favor the second. Opinions?

noc0lour:
A commit changing `M_SQRT2` and `M_LOG2E` to `GR_` is already in this PR, since they are affected in `C++`. `M_PI` isn't affected in `C++`. That's what's baffling me.

marcusmueller:
err, sorry, I meant `PI`. Anyway, yeah, um, I'd rather not treat that any special here, so even if this is redundant: if we do a `GR_M_SQRT2` and `_LOG2E`, let's also have a `GR_M_PI`, and, for the love of all that's enlightening, add a doxygen `\brief` to the former two that says "not available in ...", and two the latter "for consistency reasons with ...".

marcusmueller:
Manually rebased! So long, and thanks for all the fish!

gvanem:
I'll comment on `decode.cc` here instead of opening yet another PR. 

Anybody tried this program under Windows? It fails to read binary data from `stdin` correctly 
(since it's in `O_TEXT` mode by default). Patching it like:
```diff
@@ -55,6 +55,11 @@
   unsigned char viterbi_in[16];
   gr::fec::viterbi_chunks_init(state0);

+#ifdef _WIN32
+  setmode (fileno(stdin), O_BINARY);
+  setmode (fileno(stdout), O_BINARY);
+#endif
+
```
works better for me.

michaelld:
I would -greatly- prefer that the CAPS changes be in a separate PR since when alone in that manner they are really easy to decide on. Having them be integrated with other commits that make substantial changes makes the overall PR more difficult to determine viability, at least IMHO.

noc0lour:
Hm, This PR is fully unrelated to changes to `decode.cc` I think?  And it was merged way back already?

PR #1582, opened 2018-02-06
zeromq: change ports used in zeromq_sub QA

noc0lour:
Running zeromq QA tests in parallel can result in using the same port twice for different tests
which would show as failing tests.


marcusmueller:
I wholeheartedly support the idea that 

- tests shouldn't be influencing each other and
- things should generally be able to run in parallel.

Now, for tests the second part is actually a goal that I wasn't aware of it being a formal requirement yet. 

@noc0lour should we actually make it a rule that test *must* be able to run in parallel? How many of the tests actually *do* work when run in parallel?

While for port numbers, adequate workarounds do exist (see this), there are other blocks that _should_ have tests, but might have only one device to work with (e.g. audio sources, gr-uhd). (Hardware-including CI is a whole new story, though.)

PR #1596, opened 2018-02-08
gr_modtool: python QA code for C++ blocks will attempt non swig module import

mrjacobagilbert:
Change allows python QA code for C++ blocks to attempt to import the module directly if import of module_swig fails. This allows for running the QA code explicitly which is helpful for non-build time tests (and efficient writing of QA code): `python qa_my_cpp_block.py`

### DIFF
--- a/gr-utils/python/modtool/templates.py
+++ b/gr-utils/python/modtool/templates.py
@@ -494,7 +494,10 @@ ${str_to_python_comment($license)}
 from gnuradio import gr, gr_unittest
 from gnuradio import blocks
 #if $lang == 'cpp'
-import ${modname}_swig as ${modname}
+try:
+    import ${modname}_swig as ${modname}
+except ImportError:
+    import ${modname} as ${modname}
 #else
 from ${blockname} import ${blockname}
 #end if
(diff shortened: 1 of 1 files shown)


noc0lour:
Formally your fix is correct, I'm wondering if there is anything that might lead to a red herring kind of situation if the first import fails because of a legit bug in the wrapped library?

mrjacobagilbert:
Hmm... Now that you mention it, this could cause a couple issues such as the module being built and tested failing import, and a previously installed module being imported and the failure going unnoticed. 

It would be nice to have the functionality I'm after though...

mbr0wn:
Your tests could also fail in random ways if the catch-clause imports a previously installed module.

What exactly is the use case you're thinking of? Maybe we can enable rapid-iteration features otherwise, e.g., by checking an env variable or something.

mrjacobagilbert:
I am in the process of writing QA tests for several general purpose PDU and timing based GR modules I am about to release. One annoyance is that I cannot install a module and then execute test code I am developing in the `python python/qa_block-name.py` way for C++ blocks as they `import module-name_swig` instead of `import module-name`. 

So yeah, maybe there is something different I should be doing to accomplish this...open to recommendations.

mrjacobagilbert:
So I think the proper way is to adjust how I am doing this. I think running `ctest -V -R <blockname>` allows me to do what I need in a better way so I'll withdraw this.

PR #1597, opened 2018-02-08
pmt: fix for error in pmt::equal() when comparing uniform vectors

mrjacobagilbert:
PMT uniform vector comparisons with pmt::equal() do not fail for different vector elements provided the vectors are the same size and type:

before:

```
>>> import pmt
>>> uv1 = pmt.init_u8vector(4,[1,2,3,4])
>>> uv2 = pmt.init_u8vector(4,[5,6,7,8])
>>> uv3 = pmt.init_u8vector(4,[5,6,7,8])
>>> pmt.equal(uv1, uv2)
True
>>> pmt.equal(uv2, uv3)
True
```

after:

```
>>> import pmt
>>> uv1 = pmt.init_u8vector(4,[1,2,3,4])
>>> uv2 = pmt.init_u8vector(4,[5,6,7,8])
>>> uv3 = pmt.init_u8vector(4,[5,6,7,8])
>>> pmt.equal(uv1, uv2)
False
>>> pmt.equal(uv2, uv3)
True
```

### DIFF
--- a/gnuradio-runtime/lib/pmt/pmt.cc
+++ b/gnuradio-runtime/lib/pmt/pmt.cc
@@ -1163,7 +1163,7 @@ equal(const pmt_t& x, const pmt_t& y)
 	       len_x) == 0)
       return true;
 
-    return true;
+    return false;
   }
 
   // FIXME add other cases here...
--- a/gnuradio-runtime/lib/pmt/pmt.cc
+++ b/gnuradio-runtime/lib/pmt/pmt.cc
@@ -1158,9 +1158,9 @@ equal(const pmt_t& x, const pmt_t& y)
       return false;
 
     size_t len_x, len_y;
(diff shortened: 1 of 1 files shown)


noc0lour:
Can you add a unit test that fails before your fix and passes now? So next time someone breaks it we will see it.

noc0lour:
So, there is a grave bug in `pmt.cc` using non-initialized `len_y` `len_x` ....

marcusmueller:
@noc0lour please formally document by opening an issue, thanks :)

noc0lour:
So, if @mrjacobagilbert is OK with that I can push the fix on top of his PR here, since fixing one thing will break the other...

mrjacobagilbert:
Yeah you can push a fix into this PR.

noc0lour:
So the failing tests are known flakiness & `volk` with GCC 7 ...

PR #1601, opened 2018-02-08
docs: first generate pages & suppress latex output

noc0lour:
This allows to use \ref page_* from the code without doxygen warnings.
References to classes and functions from page_* documentation are possible
without specifying \ref, thus suppressing more warnings about non-existent
references

LaTeX output is suppressed using batchmode


noc0lour:
No idea why this isn't helping with the doxygen generation in the CI system...

noc0lour:
Ok, with parallel builds this is useless

Issue #1585, opened 2018-02-06
objective_fct.c and whatever is not needed from the gen_interpolator_taps directory should be removed


marcusmueller:
@awalls-cx18 pointed out there might be licensing problems surrounding that code.

awalls-cx18:
Technically simpson.[ch] needs to be removed.  objective_fct.c can stay, but it won't work until it is ported to use numerical integration routines from the GSL library.

I recall looking at praxis.f and thinking it was OK.  Someone should take another look at it and the associated README file.

marcusmueller:
closed by #1605, as far as I can see. A **huge** thank goes out to the ever competent @awalls-cx18!

PR #1605, opened 2018-02-09
gr-filter: Remove simpson's rule integration and replace with GSL call

awalls-cx18:
This change set addresses issue #1585 .


awalls-cx18:
The two files in attached zip archive are the taps generated before and after the change.  One can use vimdiff to see that the files don't differ in any meaningful way.  (Some digits in the 5th decimal place differ by +/- 1.)

[interpolator_taps.h.zip](https://github.com/gnuradio/gnuradio/files/1709396/interpolator_taps.h.zip)

gr-filter/lib/gen_interpolator_taps/objective_fct.c
```diff
@@ -49,7 +51,7 @@ double		global_B  = B;
  */
 
 static double
-integrand (double omega)
+integrand (double omega, void *params)
```
mbr0wn: Won't compilers complain about the unused param? Can you just do `integrand(double omega, void*)`?

mbr0wn:
For other reviewers: https://www.gnu.org/software/gsl/manual/html_node/QAG-adaptive-integration.html

gr-filter/lib/gen_interpolator_taps/objective_fct.c
```diff
@@ -89,10 +91,19 @@ integrand (double omega)
...
+
+  F.function = integrand;
+  F.params = NULL;
+
   assert ((n & 1) == 0);	/* assert n is even */
   global_n = n;
   global_h = x;
-  return qsimp (integrand, -2 * M_PI * global_B, 2 * M_PI * global_B);
+  gsl_integration_qag(&F, -2 * M_PI * global_B, 2 * M_PI * global_B,
+                      0.0, 1e-12, 1000, GSL_INTEG_GAUSS61,
```
mbr0wn: This is the only line I'm not clear about: How did you choose those parameters? To be clear, I totally trust your judgment,but from the diff it's not obvious how you reached this. It might improve readability if you declare your epsabs, epsrel, limit as local `const`s. Then again, this is not critical.

awalls-cx18: Well this is an offline utility that is run once every 15 years, so being time efficient does matter.  Keeping that in mind:
- GSL_INTEG_GAUSS61 was the option for best accuracy that this function call had to offer
- epsabs was set to 0.0, so that only epsrel would matter (keeping it simple).
( https://www.gnu.org/software/gsl/manual/html_node/Numerical-Integration-Introduction.html#Numerical-Integration-Introduction )
- epsrel was set to "a really freaking small number" for relative error
- limit was set to "more iterations than we should ever need" to converge
(which wasn't quite right as it specifies the maximum number of subintervals, but the function is so smooth, the number of subintervals shouldn't be large either.)

TL;DR Numerical accuracy mattered; speed didn't.  Let's throw in some extreme numbers.  Oh look, the numbers came out almost the same, and the GSL optimized routines worked faster than the simpson's rule code too. Done. :)

awalls-cx18: I just checked the estimated errors reported by the integration function after it has run.  The largest is 8.136707e-14; the smallest is 1.783098e-24.

I also can't make it better by pushing epsrel smaller (say to 1e-14), since the integration routine will abort() if it can't make the tolerance.

So hey, 1e-12 was a nice lucky guess. :)

mbr0wn: You have me convinced.

gr-filter/lib/gen_interpolator_taps/objective_fct.c
awalls-cx18: Yeah probably, but I didn't set the compiler switches in generate.sh to gripe about uninitialized variables, since most of the code was existing.

A better thing to do would be to use the "params" pointer to pass in a structure of data to eliminate the global_n and global_h variables.  Probably more work than its worth though.

PR #745, opened 2016-02-10
Rayleigh fading

bastibl:
If I see that correctly, the rayleigh fader

- uses only K-1 sinusoids
- and normalizes the power to K-1/K *2 instead of 1

I think that a power level of 1 is also assumed when weighing NLOS and LOS components.

PR #1606, opened 2018-02-09
Update selective_fading_model_impl.cc

jan-safar:
The multiplication by 2 in the sinc() function argument appears to be a mistake. It, for example, results in signal paths with a delay of 0.5 sample time being completely ignored.

https://en.wikipedia.org/wiki/Sinc_function

The same change may be required in 'selective_fading_model2_impl.cc'.

### DIFF
--- a/gr-channels/lib/selective_fading_model_impl.cc
+++ b/gr-channels/lib/selective_fading_model_impl.cc
@@ -103,7 +103,7 @@ namespace gr {
                 //gr_complex ff_H(d_faders[j]->next_sample());
                 for(size_t k=0; k<d_taps.size(); k++){
                     float dist = k-d_delays[j];
-                    float interpmag = d_sintable.sinc(2*M_PI*dist);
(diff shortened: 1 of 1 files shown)


mbr0wn:
Not yet reviewed, but this would have to go onto maint.

mbr0wn:
(With the current branch model)

jan-safar:
I only used 'next' because the flat fading model in 'master/maint' has other issues (as you're probably aware):
https://oshearesearch.com/index.php/2016/03/28/gr-channel-model-enhancements/
https://github.com/gnuradio/gnuradio/pull/745

marcusmueller:
I'll cherry-pick this onto maint, as this really is a bug fix; it does make far more sense to let `dist` interpolate between the maximum at 0 of a sinc and its first zero-crossing at M_PI.

marcusmueller:
Reworded the commit message to be meaningful, committ 7d19bf6 on maint.

PR #1612, opened 2018-02-12
python gateway: permit a variable number of ports

willcode:
Here is an initial cut at allowing Python blocks to have a variable number of ports. Backward compatibility is maintained by checking the type of io args passed to the constructor. The test program below shows how this works. A new py_io_signature class is used. This class is similar to io_signature, but keeps track of the numpy type for each port.

Performance is within 5% of the previous version.

```
-----
EXAMPLE OF USAGE (implements multi-port add block)
-----

#!/usr/bin/env python

from gnuradio import gr, blocks
import numpy as np

# Set this to test py_io_signature
NEW=1

class TestBasic(gr.basic_block):

    def __init__(self):

        # Using py_io_signature, variable number of inputs, fixed number of outputs
        if NEW:
            gr.basic_block.__init__(self, "Test",
                                    gr.py_io_signature(2,-1, ['float32', 'float32']),
                                    gr.py_io_signature(1, 1, ['float32']))

        # Backward compatible, fixed number of I/O ports
        else:
            gr.basic_block.__init__(self, "Test", ['float32', 'float32'], ['float32'])

    def general_work(self, input_items, output_items):
        n = len(output_items[0])
        output_items[0][:] = input_items[0][:n]
        for i in range(1,len(input_items)):
            output_items[0][:] += input_items[i][:n]

        self.consume_each(n)
        return n

class TestSync(gr.sync_block):

    def __init__(self):

        # Using py_io_signature, variable number of inputs, fixed number of outputs
        if NEW:
            gr.sync_block.__init__(self, "Test",
                                   gr.py_io_signature(2,-1, ['float32', 'float32']),
                                   gr.py_io_signature(1, 1, ['float32']))

        # Backward compatible, fixed number of I/O ports
        else:
            gr.sync_block.__init__(self, "Test", ['float32', 'float32'], ['float32'])

    def work(self, input_items, output_items):
        n = len(output_items[0])
        output_items[0][:] = input_items[0][:n]
        for i in range(1,len(input_items)):
            output_items[0][:] += input_items[i][:n]

        self.consume_each(n)
        return n

###
## MAIN
#
tb = gr.top_block()
src = blocks.vector_source_f(range(1000), repeat=True)
head = blocks.head(4, 100000000)
dst = blocks.null_sink(4)
test = TestBasic()
tb.connect(src, (test,0))
tb.connect(src, (test,1))

# Add more ports at run time
if NEW:
    tb.connect(src, (test,2))
    tb.connect(src, (test,3))
    tb.connect(src, (test,4))

tb.connect(test, head, dst)
tb.run()
```

### DIFF
--- a/gnuradio-runtime/python/gnuradio/gr/gateway.py
+++ b/gnuradio-runtime/python/gnuradio/gr/gateway.py
@@ -1,5 +1,5 @@
 #
-# Copyright 2011-2012 Free Software Foundation, Inc.
+# Copyright 2011-2012, 2018 Free Software Foundation, Inc.
 #
 # This file is part of GNU Radio
 #
@@ -78,35 +78,61 @@ class msg_handler(gr.feval_p):
         return 0
 
 ########################################################################
+# io_signature for Python
+########################################################################
+class py_io_signature(object):
+
+    # Minimum and maximum number of ports, and a list of numpy types.
+    def __init__(self, min_ports, max_ports, type_list):
+        self.__min_ports = min_ports
+        self.__max_ports = max_ports
+        self.__types = map(numpy.dtype, type_list)
+
+    # Make/return a gr.io_signature. A non-empty list of sizes is
+    # required, even if there are no ports.
+    def gr_io_signature(self):
+        return io_signaturev(self.__min_ports, self.__max_ports,
+                             [t.itemsize for t in self.__types] or [0])
+
+    # Return data types for the first nports ports. If nports is
+    # smaller than the provided type list, return a truncated list. If
+    # larger, fill with the last type.
+    def port_types(self, nports):
+        ntypes = len(self.__types)
+        if ntypes == 0:
+            return []
+        if nports <= ntypes:
+            return self.__types[:nports]
+        return self.__types + [self.__types[-1]]*(nports-ntypes)
+
+########################################################################
 # The guts that make this into a gr block
 ########################################################################
 class gateway_block(object):
 
     def __init__(self, name, in_sig, out_sig, work_type, factor):
 
-        #ensure that the sigs are iterable dtypes
-        def sig_to_dtype_sig(sig):
-            if sig is None: sig = ()
-            return map(numpy.dtype, sig)
-        self.__in_sig = sig_to_dtype_sig(in_sig)
-        self.__out_sig = sig_to_dtype_sig(out_sig)
+        # Normalize the many Python ways of saying 'nothing' to '[]'
+        in_sig = in_sig or []
+        out_sig = out_sig or []
 
-        #cache the ranges to iterate when dispatching work
-        self.__in_indexes = range(len(self.__in_sig))
-        self.__out_indexes = range(len(self.__out_sig))
-
-        #convert the signatures into gr.io_signatures
-        def sig_to_gr_io_sigv(sig):
-            if not len(sig): return io_signature(0, 0, 0)
-            return io_signaturev(len(sig), len(sig), [s.itemsize for s in sig])
-        gr_in_sig = sig_to_gr_io_sigv(self.__in_sig)
-        gr_out_sig = sig_to_gr_io_sigv(self.__out_sig)
+        # Backward compatibility: array of type strings -> py_io_signature
+        if type(in_sig) is py_io_signature:
+            self.__in_sig = in_sig
+        else:
+            self.__in_sig = py_io_signature(len(in_sig), len(in_sig), in_sig)
+        if type(out_sig) is py_io_signature:
+            self.__out_sig = out_sig
+        else:
+            self.__out_sig = py_io_signature(len(out_sig), len(out_sig), out_sig)
 
         #create internal gateway block
         self.__handler = gateway_handler()
         self.__handler.init(self.__gr_block_handle)
         self.__gateway = block_gateway(
-            self.__handler, name, gr_in_sig, gr_out_sig, work_type, factor)
+            self.__handler, name,
+            self.__in_sig.gr_io_signature(), self.__out_sig.gr_io_signature(),
+            work_type, factor)
         self.__message = self.__gateway.block_message()
 
         #dict to keep references to all message handlers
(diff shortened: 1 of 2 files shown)


giladbeeri:
@willcode I think it fails with Python sink blocks, they don't have out_sig.

``` File "/home/gilad/gr/as/lib/python2.7/dist-packages/gnuradio/gr/gateway.py", line 263, in __init__
    factor=1,
  File "/home/gilad/gr/as/lib/python2.7/dist-packages/gnuradio/gr/gateway.py", line 122, in __init__
    self.__out_sig = py_io_signature(len(out_sig), len(out_sig), out_sig)
TypeError: object of type 'NoneType' has no len()
```

I assume it will also fail for source blocks but haven't checked.

willcode:
Changed to accept any form of Python 'nothing' (None, [], 0, etc.) to indicate that there are no ports. I didn't know there was a qa module for gateway - it's over in gr-blocks instead of gnuradio-runtime. That should probably be moved. Now, qa passes. @giladbeeri try it out again?

giladbeeri:
@willcode I tested it with a sync block and `in_sig=gr.py_io_signature(1, -1, [numpy.float32])` (where previously I had `in_sig=[numpy.float32] * num_ports`), it works as expected. Thank you!

willcode:
Updated copyright dates, squashed, and added a better commit message. @marcusmueller please consider merging.

gnuradio-runtime/python/gnuradio/gr/gateway.py
```diff
@@ -77,36 +77,62 @@ def eval(self, arg):
...
 
-        #ensure that the sigs are iterable dtypes
-        def sig_to_dtype_sig(sig):
-            if sig is None: sig = ()
-            return map(numpy.dtype, sig)
-        self.__in_sig = sig_to_dtype_sig(in_sig)
-        self.__out_sig = sig_to_dtype_sig(out_sig)
+        # Normalize the many Python ways of saying 'nothing' to '[]'
+        in_sig = in_sig or []
+        out_sig = out_sig or []
```
noc0lour: Do we need a mutable type here? Would a tuple `()` be sufficient?

gnuradio-runtime/python/gnuradio/gr/gateway.py
```diff
@@ -128,36 +154,47 @@ def __gr_block_handle(self):
...
+            out_types = self.__out_sig.port_types(noutputs)
             self.__message.general_work_args_return_value = self.general_work(
 
                 input_items=[pointer_to_ndarray(
                     self.__message.general_work_args_input_items[i],
-                    self.__in_sig[i],
+                    in_types[i],
                     self.__message.general_work_args_ninput_items[i]
-                ) for i in self.__in_indexes],
+                ) for i in xrange(ninputs)],
```
noc0lour: As much as I like `xrange()` for speed improvements in Py2k this will break in Py3k. Consider using `range()` even if it's a bit slower in Py2k. As this is probably good work we want to have in the `next` branch.

gnuradio-runtime/python/gnuradio/gr/gateway.py
```diff
@@ -128,36 +154,47 @@ def __gr_block_handle(self):
...
 
                 output_items=[pointer_to_ndarray(
                     self.__message.general_work_args_output_items[i],
-                    self.__out_sig[i],
+                    out_types[i],
                     self.__message.general_work_args_noutput_items
-                ) for i in self.__out_indexes],
+                ) for i in xrange(noutputs)],
```
noc0lour: Same `xrange()` comment here.

gnuradio-runtime/python/gnuradio/gr/gateway.py
```diff
@@ -128,36 +154,47 @@ def __gr_block_handle(self):
...
+            out_types = self.__out_sig.port_types(noutputs)
             self.__message.work_args_return_value = self.work(
 
                 input_items=[pointer_to_ndarray(
                     self.__message.work_args_input_items[i],
-                    self.__in_sig[i],
+                    in_types[i],
                     self.__message.work_args_ninput_items
-                ) for i in self.__in_indexes],
+                ) for i in xrange(ninputs)],
```
noc0lour: `xrange()` -> `range()`

gnuradio-runtime/python/gnuradio/gr/gateway.py
```diff
@@ -128,36 +154,47 @@ def __gr_block_handle(self):
...
 
                 output_items=[pointer_to_ndarray(
                     self.__message.work_args_output_items[i],
-                    self.__out_sig[i],
+                    out_types[i],
                     self.__message.work_args_noutput_items
-                ) for i in self.__out_indexes],
+                ) for i in xrange(noutputs)],
```
noc0lour: `xrange()` -> `range()`

gnuradio-runtime/python/gnuradio/gr/gateway.py
willcode: mutable is not needed

gnuradio-runtime/python/gnuradio/gr/gateway.py
willcode: range in Python 3 is xrange in Python 2. Since GR 3.7 is Python 2 only, can we keep it xrange here and change it to range for the new branch?

willcode: Performance difference is trivial in testing - changing to range.

marcusmueller:
So, regarding Py2/3: Yes, 3.7 is Python 2 only. Yes, 3.8 will be Py3.

`xrange` might be the right thing to do in general, but here, I'd
concur with @noc0lour: The potential memory overhead of a temporary
list (instead of a generator for ints) is outshadowed by the fact that
we're making an array of pointers to ndarrays of the same size. I'm not
foreseeing Python blocks with millions of output ports anytime soon,
anyway.

Does that answer the question?

marcusmueller:
Merged manually. Thank you!

Issue #1608, opened 2018-02-11
gr_logger: OOT modules do not generate log messages


mrjacobagilbert:
I am in the process of changing std::cout messages in an OOT I maintain to use the GR_LOG API and cannot get any log message output. In-tree logger messages print [as described](https://gnuradio.org/doc/doxygen/page_logger.html) when configured to use stdout and stderr or file based logging, but nothing in the OOT will print. 

It's entirely possible I am not using this properly; I am assuming d_logger is something that GR already has for every block...

Module has no changes to CMake files, no extra includes, just put some statements like this into a C++ block:

`GR_LOG_INFO(d_logger, "******** THIS IS A GR LOG INFO MESSAGE ********");`
`std::cout << "++++++++ THIS IS A STD COUT MESSAGE ++++++++" << std::endl;`

The second line prints but not the first. The module also compiles regardless of what I put for the logger; `d_logger`, `d_logger2`, `asdfasdf` which seems suspect. GR seems to pre-define `d_logger` and `d_debug_logger` but no other loggers, so this is a bit odd to me.

GR was built with liblog4cpp, snip from cmake:

```
...
-- Using install prefix: /opt/gnuradio
--
-- Configuring volk support...
--   Enabling volk support.
--   Override with -DENABLE_VOLK=ON/OFF
--   Override with -DENABLE_INTERNAL_VOLK=ON/OFF
-- ENABLE_GR_LOG set to ON.
-- HAVE_LOG4CPP set to True.
-- LOG4CPP_LIBRARIES set to /usr/lib/liblog4cpp.so.
-- Could NOT find Sphinx (missing:  SPHINX_EXECUTABLE)
--
-- Configuring doxygen support...
--   Dependency DOXYGEN_FOUND = YES
...
```

Posting this here to see if someone else can confirm what is going on - if this is just user error, I'll see if I can clarify the documentation.

mrjacobagilbert:
I did a bit more exploring, and I am pretty sure these GR_LOG_* messages are skipped entirely (not compiled?). Putting nonsense in there, or things like divide by zeros compile and run fine (except no log output):

```
d_x = 0;
GR_LOG_INFO(d_logger, boost::format("asdf %1%") % (4/d_x));
GR_LOG_INFO(d_logger, boost::format("asdf %1%") % "string?");
GR_LOG_INFO(d_logger, whateveriputheredoesntmatter@#$234);
```

noc0lour:
So, did you try calling first:
[...]

PR #1614, opened 2018-02-15
logging: updated doxygen with how to use logging in OOT

mrjacobagilbert:
Updated documentation for using logging in OOT modules. Related to issue #1608.


mbr0wn:
I think mid-term we should make it easier to log in OOTs, though. For now, this is a great addition, thanks!

PR #1618, opened 2018-02-16
gr-channels: Fixing the formula behind the IQ_imbal block

SaikWolf:
Fixing the formula behind the IQ_imbal block to be true mathematically.
X' = (1+mag_lin)·cos(theta)·real(X) + j·((1+mag_lin)·sin(theta)·real(X) + imag(X))
is the proper equation for handling the IQ imblance purely within the Inphase component.

Previously it was:
X' = (1+mag_lin)·real(X) + j·(sin(theta)·real(X) + imag(X))

### DIFF
--- a/gr-channels/grc/channels_iqbal_gen.xml
+++ b/gr-channels/grc/channels_iqbal_gen.xml
@@ -8,7 +8,7 @@
   <name>IQ Imbalance Generator</name>
   <key>channels_iqbal_gen</key>
   <import>from gnuradio import channels</import>
-  <make>channels.iqbal_gen($mag, $phase)</make>
+  <make>channels.iqbal_gen($mag, $phase, $mode)</make>
   <callback>set_magnitude($mag)</callback>
   <callback>set_phase($phase)</callback>
   <param>
@@ -22,6 +22,20 @@
     <key>phase</key>
     <value>0</value>
     <type>float</type>
+  </param>
+  <param>
+    <name>Origin</name>
+    <key>mode</key>
+    <value>0</value>
+    <type>enum</type>
+		<option>
+			<name>Reciever</name>
+			<key>1</key>
+		</option>
+		<option>
+			<name>Transmitter</name>
+			<key>0</key>
+		</option>
   </param>
   <sink>
     <name>in</name>
--- a/gr-channels/grc/channels_iqbal_gen.xml
+++ b/gr-channels/grc/channels_iqbal_gen.xml
@@ -45,4 +45,35 @@
     <name>out</name>
     <type>complex</type>
   </source>
+  <doc>
+This block implements the single branch IQ imbalance
+transmitter and receiver models.
+
+Developed from source (2014):
+"In-Phase and Quadrature Imbalance: Modeling, Estimation, and Compensation"
+
+    TX Impairment:
+                                      
+                                                        {R}--|Multiply: 10**(mag/20)|--+--|Multiply: cos(pi*degree/180)|--X1
+Input ---|Complex2Float|---|                                                               |--|Multiply: sin(pi*degree/180)|--X2
+                                                        {I}--|   Adder  |
+                                                       X2--|       (+)      |--X3
+
+                          X1--{R}--| Float 2     |--- Output
+                          X3---{I}--| Complex |
+
+    RX Impairment:
+                                      
+                                                        {R}--|Multiply: cos(pi*degree/180)|-------|               |
+Input ---|Complex2Float|---|                                                                                  | Adder |--X1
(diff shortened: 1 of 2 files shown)


SaikWolf:
Double checking the formula.

marcusmueller:
:) thanks! I just didn't get around doing that myself. If you have trouble, I hear `#gnuradio` on [Freenode](https://webchat.freenode.net/) or the [mailing list](https://lists.gnu.org/mailman/listinfo/discuss-gnuradio) is full of people to discuss with :)

SaikWolf:
So the change I made makes it a transmitter IQ imbalance model. I'll edit it a little more and add a toggle for Transmitter/Receiver and make the adjustments needed for receiver IQ imbalance, and then reopen.

SaikWolf:
Ok, the changes now reflect the origin of the IQ Imbalance while assuming a single component error.
Logic provided below for ease in analyzing.

Transmitter Effects:
X = Xi + j*Xq
X' @ w = Real{(Xi+j*Xq)*(mag_lin*cos(w*t + theta) + j*sin(w*t))}
X' @ w = (mag_lin*cos(theta)*Xi)*cos(w*t) - (mag_lin*sin(theta)*Xi + Xq)*sin(w*t)
X' @ w = Xi'*cos(w*t) - Xq'*sin(w*t)
X' @ baseband = Xi' + j*Xq'
Xi' = mag_lin*cos(theta)*Xi
Xq' = mag_lin*sin(theta)*Xi + Xq

Receiver Effects:
X = Xi*cos(w*t) - Xq*sin(w*t)
Xi' @ baseband = Filtered{ X*2*mag_lin*cos(w*t+theta) }
Xi' = mag_lin*(cos(theta)*Xi + sin(theta)*Xq)
Xq' @ baseband = Filtered{ X*-2*sin(w*t) }
Xq' = Xq
X' = Xi'+ j*Xq'
X' = mag_lin*(cos(theta)*Xi + sin(theta)*Xq) + j*Xq

awalls-cx18:
Hi @SaikWolf and @marcusmueller,

I've taken a look at the receiver model and the math and implementation match.  The math is correct.  But it seems like there some arbitrary choices:
a. The incoming signal had a negative sign on the modulated quadrature component (i.e. -Xq sin(wt))
b. The receiver quadrature mixer mixes in -2sin(wt) vs. +2sin(wt)
c. The gain imbalance is manifested (wholly) after the mixing and filtering vs. before.

The above choices make the math nice in that everything comes out with a positive sign and you only need one multiply block to handle the gain imbalance.

I guess a. and b. are OK, as long as one understands the transmitter modulator is behaving in a complementary manner so that -Xq sin(wt) is the quadrature component going over the air.
However, an uninformed user who didn't read the block code or work through the math might not understand that about the model.

Regarding c., in my limited experience, I've usually seen gain imbalance modeled before the mixing.  I haven't seen it modeled afterward.  It's still valid, just not what I assumed was going on before reading through the math.

This block could really use a block diagram in the comments and/or documentation to document the receiver model pictorially. (This could likely be said for many blocks in gr-channel.)  Something as simple what's in figure 3.4 here  [http://www.eit.lth.se/sprapport.php?uid=764] would be nice.

I'll check the transmitter model next.  I'm guessing @SaikWolf's math will be correct, the model won't match my going in assumptions, and I'll suggest a block diagram would be nice. :)

-Andy

noc0lour:
@SaikWolf probably GRC is not the best choice for creating ASCII arts if you don't use a monospace font. Using my editor (which uses monospace font) I can't see the ASCII art.

awalls-cx18:
If the block diagram doesn't work out, I'm good with a reference.

@SaikWolf is your reference the Springer Brief by Li and Yabo?

noc0lour:
I didn't test with GRC here, just looking at the xml in my editor. GRC might work out for everyone (but I doubt that due to font rendering issues).

SaikWolf:
@noc0lour I've added the monospace version of the block diagram to the Python script. Should hopefully work for all now.

@awalls-cx18 Yes, the Springer brief is what was used. Although that focuses on the balanced approach rather than the single branch approach. I simply adapted what was needed following their explanation.

noc0lour:
@SaikWolf Thanks a lot for fixing the documentation w.r.t to ASCII art :+1:

marcusmueller:
So, albeit it's taken me far too long to follow up on this: 

@SaikWolf, this is grade-A exquisite nice work! Thank you and: Keep that up :) 

@awalls-cx18 Thank you so much for doing the in-depth math analysis. I simply wouldn't be able to put as much math rigor into this, and I'm really thankful for having you on board :)
 
@noc0lour thanks for handling the tagging, and the doc feedback; always great to have you help the project, too!

PR #1628, opened 2018-02-20
Removing calls to PMT::mp("string") at run time wherever possible

marcusmueller:
typical usage:

     message_port_pub(pmt::mp("out_port"), …)

which is bad, as it implies hashing of a string, allocation of memory,
deallocation, finding the hashed string in the table of interned strings and
returning a unique pointer (which for reasons of PMT awesomeness isn't even
unique) to the interned port name.

Replacing all these port name ad hoc ::mp() calls by reusing one, private, port
name member. When feasible, even static const.

PMT::intern is a sometimes less-than-optimal implementation, but things like
`#define`s that put `pmt::mp("string")` where people thought they were actually
using a constant is straight out harmful.

### DIFF
--- a/gnuradio-runtime/include/gnuradio/block.h
+++ b/gnuradio-runtime/include/gnuradio/block.h
@@ -684,6 +684,8 @@ namespace gr {
     bool                  d_update_rate;           // should sched update rel rate?
     bool d_finished;    // true if msg ports think we are finished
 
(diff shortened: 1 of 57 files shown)


marcusmueller:
@mbr0wn If you'd want to review the `gr-uhd` part of this, please? (And maybe digital.) (and while you're at it... I'm sorry :disappointed: )

PR #1633, opened 2018-02-22
POC: Convert custom templating to valid C++ templates

noc0lour:
I did try out to use regular C++ templating for blocks with multiple 
types. Just as Proof of Concept I got the `or_XX` block converted to `or_blk_XX`
by using c++ templates and having a new swig magic wrapping templated classes.

One downside: I had do explicitly instantiate used types in `_impl.cc`. 
This would make our code cleaner since we could stop using custom build time templating.

This is just a show off. Not merge ready. Please discuss in here.


marcusmueller:
Hm, idea: This should be a GREP? It's

* something that describes a (perceived) enhancement to GNU Radio
* something that has to function as general development guide and
* something that has more than a couple of files of scope.

noc0lour:
OK, I'll submit a GREP for that as well. First I want to have some feedback about that topic. Maybe C++ templates are bad and we have custom templates for a reason?

michaelld:
Wow this brings back memories! We worked up a proof of concept for using C++ templates -- roughly same-same as this -- quite some years ago now; maybe 2009 or 2010. It worked nicely, and (as here) required explicit instantiation for any template to be usable. Unfortunately (in my opinion), "the powers that be" decided to stick with the script-based templates. I really don't recall the reason any longer.

michaelld:
I'm all for using C++ templates, by the way. I think they are convenient, easy enough to implement, and robust -- if the standard / generic template won't work for some specific type or class, one can always overload that specific type / class with its own specific code.

michaelld:
I once had an OOT that "replaced" some of standard GR templates with C++ templates. It would have worked as a drop-in replacement for the GR versions from an API / ABI perspective. I found no downsides to using C++ templates.

noc0lour:
Gonna read some ML then as well, thanks @michaelld !

marcusmueller:
@michaelld heh, I actually think that that was among the first things I ever asked on the IRC channel – must've been around the same time, too :)

If I'm a power that is, then that power (which is) is of the opinion that SWIGs that don't support real C++ templates must've gone extinct by now, and that we get rid of gengen where possible – I'm really not a big fan of maintaining .t files, and the introduction of VOLK with it's data-type specific function calls has rendered a lot of the gengen advantage (over copy&pasted code) obsolete.
 
I'd much prefer clean C++ templates with specialization as need be than having more intermediate code.

noc0lour:
Also unlikely I'd convert every other block manually. I'm thinking of a tool which will parse those lines and do it for me.

mbr0wn:
I vaguely remember the discussions around templates, I can't remember the arguments, but I think it was at least partly the state of SWIG at the time.

This approach looks very clean; much nicer than the old approach. I'd hit the 'approve' button if you hadn't flagged it as PoC. I'm also a fan of C++ templates, and they really are supposed to do what we're doing much more clunkily right now.

noc0lour:
Also I picked the worst example block there is, since `or` is a keyword in C++ and we definitely don't want an `or` class in the `gr::blocks` namespace ...

mbr0wn:
Oh that's why it's `or_blk` :smile: Should have figured that out myself.

michaelld:
Only question I've come up with thus far is: How does moving to C++ templates change including those blocks in C++ heir blocks?

noc0lour:
This will change the block signature so C++ blocks will have to switch to template syntax. For Python we can keep most block signatures but will have to adapt where needed. See `or_blk` or something similar. For GRC we just have to adjust the make syntax in `yaml`.

marcusmueller:
couldn't we just actually `typedef` something like `funky_block_ss` to `funky_block<uint16_t, uint16_t>`?

noc0lour:
We can, and then also add a deprecated attribute to it, so users get yelled at by the compiler.

gr-blocks/include/gnuradio/blocks/or_blk.h
```diff
@@ -37,17 +35,18 @@ namespace gr {
...
-    class BLOCKS_API @NAME@ : virtual public sync_block
+    template<class T>
+    class BLOCKS_API or_blk : virtual public sync_block
     {
     public:
 
-      // gr::blocks::@NAME@::sptr
-      typedef boost::shared_ptr<@NAME@> sptr;
+      // gr::blocks::or_blk::sptr
+      typedef boost::shared_ptr<or_blk> sptr;
```
michaelld: I'll admit that I haven't tried your code changes yet, nor tried compiling this change ... but shouldn't the "or_blk" here be something more like "or_blk<T>"? -Something- needs to reflect the template class here in the changes, otherwise why bother with making it a template?

michaelld:
So, using this POC as an example, is this a correct interpretation: Leave in the old "or" block but mark it as deprecated, and then move to using "or_blk<T>" as the new way of doing "or", with a typedef for specific instantiations of the template? Hmmm ... not sure that's what y'all are meaning. Seems like we should be able to do a 1:1 swap-out from the current "template" generation version to this new C++ template version. No deprecation necessary.

noc0lour:
@michaelld We will have to add `typedef` to do something along the lines of 
``` typedef add_ff add<float>;```
For the special case of `or` and `and` blocks we currently have 
`class or_xx` and `class and_xx` which would be `class or` and `class and` which is a big no-no.
Thus I just named it `class or_blk` to avoid overloading `or` in the `gr::blocks` namespace. 

If we want to keep around those `typedef`s forever I'm fine with not deprecating them.

gr-blocks/include/gnuradio/blocks/or_blk.h
noc0lour: Oh, I might have make it `shared_ptr<or_blk<T> >` good catch.

michaelld:
@noc0lour Oh, OK I see what you're getting at: we can't use C++ keywords as class names, so while most block names could be reused (as they are not C++ keywords), some such as `or` and `and` require slight renaming to avoid the C++ keyword issue. Got it.

So we change the class name for these specific cases to `template<class T> class or_blk` etc to avoid this keyword issue. Then we do `typedef or_ff or_blk<float>` in the main header for these blocks to define the supported instantiations of this block & to make sure the API is maintained / unchanged.

Yes, I can see how this could be made to work.

I'm still not sure what needs to be deprecated though. Seems like if the API is maintained, then any code using the `or_ff` class will have expand it to `or_blk<float>` and all will be well from the compiler's perspective at least. I mean clearly when debugging the class name for this block will be different, but I think that's a separate issue.

noc0lour:
I found the original mail thread from back then: 
https://lists.gnu.org/archive/html/discuss-gnuradio/2007-08/msg00361.html

Back then everyone agreed that converting the templating to C++ templates is a good idea. Unfortunately nobody has come around to actually implement a full conversion.

michaelld:
Wow that really goes back. If I recall (after reading through those emails), there was also some off-list email discussing this topic. I doubt I can find those any longer, but I'll take a look. It's more likely that I can track down the original code, since I do still have my archives from my ND work. Of course much would have to be reworked for today's GR coding structure, but the basic underlying code (SWIG & C++ templates) should still be valid.

mbr0wn:
@noc0lour wrote a great GREP for this: https://github.com/gnuradio/greps/pull/10

michaelld:
That GREP is a good start. Are GREPs counted in binary or base-10? Looks like we've seen 0, 1, and now 10 ... so maybe the former LOL? That'll challenge some folks ...

noc0lour:
The first 10 are reserved for special things, like what's a GREP and how are coding guidelines.

marcusmueller:
So, just a gut feeling: since we've "migrated" this to a grep, shouldn't we be closing this PR here?

noc0lour:
@marcusmueller This is basically the corresponding implementation. I can  close this if you want to have it moved out of the PR queue. This is both a reminder, and a work branch I will push to when continuing to work on the conversion tool.

marcusmueller:
Yeah, I think it's better if we close this now, and give it a warm re-open when we've got a significant branch to merge :) This is definitely not a "let's not do that" (quite the contrary!) but a "it's not yet a PR".

Issue #1545, opened 2018-01-10
Scipy needed for make test; not checked for, hence inexplicable test failures


c0debreaker:
The `make` was fine. Running `make test` failed on few items

Here is the output

```
The following tests FAILED:
	 91 - qa_polar_decoder_sc (Failed)
	 92 - qa_polar_decoder_sc_list (Failed)
	 93 - qa_polar_decoder_sc_systematic (Failed)
	 94 - qa_polar_encoder (Failed)
	 95 - qa_polar_encoder_systematic (Failed)
	209 - qa_zeromq_pub (Failed)
Errors while running CTest
make: *** [test] Error 8
```
What am I missing?

c0debreaker:
I fixed it after installing scipy. :)

marcusmueller:
Changing title, reopening: this is actually an issue, and unless you submitted a pull request with a fix for that issue (namely, need for scipy not tracked for `make test`), it shouldn't be closed.

noc0lour:
Replacing scipy with something else already present would be a viable solution imho.

marcusmueller:
I'd argue that if someone builds GNU Radio, then having scipy as *build-time* dependency (just like CMake, SWIG, …) really isn't much of a burden. Packaging systems track these separately from run-time deps, anyway.

The question is whether it should be a run-time dependency, too, is a different one. None of our `gr_plot_**` tools work without, nor does `gr_filter_design`, and a lot of the gr-digital and -dtv and -filter examples depend on it, too. In many cases, I'd agree, all that's used is `fftpack`.... which does come with numpy.

jdemel:
SciPy use should actually be reduced to a minimum. Here in paticular, polar codes actually use specific SciPy function. 
I agree that making SciPy a build/run-time requirement does not hurt. Especially because almost everyone will need it and it is quite tightly bundled with NumPy anyways.

marcusmueller:
Just having had a night of rest about this: don't underestimate the effort that goes into packaging Python packages for embedded platforms. If we do make scipy a runtime req, we should really very clearly define the affected components -maybe even in finer granularity than "gr-utils".

Am 18. Februar 2018 23:44:19 MEZ schrieb jdemel <notifications@github.com>:
>SciPy use should actually be reduced to a minimum. Here in paticular,
>polar codes actually use specific SciPy function. 
>I agree that making SciPy a build/run-time requirement does not hurt.
>Especially because almost everyone will need it and it is quite tightly
[...]

Issue #1539, opened 2017-12-16
error: make test failed


Alhamzi:
hallo
while building source file of GNU radio following error occurred during "make test"
98% tests passed, 5 tests failed out of 211

Total Test time (real) = 177.80 sec

The following tests FAILED:
	 85 - qa_polar_decoder_sc_systematic (Failed)
	 87 - qa_polar_decoder_sc_list (Failed)
	 89 - qa_polar_encoder_systematic (Failed)
	 90 - qa_polar_encoder (Failed)
	 92 - qa_polar_decoder_sc (Failed)
Errors while running CTest
Makefile:61: recipe for target 'test' failed
make: *** [test] Error 8

will this error create any problem in future while using gnuradio, henceforth should i continue or should i look to rectify this error
Any suggestion will be appreciated,
Thanks

drmpeg:
Most likely, you're missing scipy.

sudo apt-get install python-scipy

justtennis:
I see similar error messages after running: make test
...

_99% tests passed, 5 tests failed out of 363

Total Test time (real) = 152.15 sec

The following tests FAILED:
	243 - qa_polar_decoder_sc (Failed)
	244 - qa_polar_decoder_sc_list (Failed)
	245 - qa_polar_decoder_sc_systematic (Failed)
	246 - qa_polar_encoder (Failed)
	247 - qa_polar_encoder_systematic (Failed)
Errors while running CTest
Makefile:85: recipe for target 'test' failed
make: *** [test] Error 8_

I ran the suggested install for python-scipy (sudo apt-get install python-scipy) but still get these errors.

I am running a new install of Ubuntu 18.04.  
Any help would be greatly appreciated.

michaelld:
What does `ctest -V -R qa_polar_decoder_sc_list` return? Hopefully whatever it shows will make the solution obvious ...

drmpeg:
If you're building GNU Radio 3.8 or the master branch, you'll need the Python3 version instead.

sudo apt-get install python3-scipy

justtennis:
short answer: that fixed it for me.  Thanks!

@michaelld good call.  your suggested command returned "No module named 'scipy'"
@drmpeg I installed Python3-scipy as you suggested:
$sudo apt-get install python3-scipy
I reran $make test
100% tests passed (from $make test)
Looks like that fixed it for me.  Thanks for the quick and helpful responses.

michaelld:
[...]

PR #1639, opened 2018-02-25
fec: make SciPy an optional dependency

jdemel:
This patch fixes issues like #1545 and #1539.
Basically if a module in fec.polar requires SciPy and fails to perform `import scipy`,
it overwrites the missing function with an approximate version

### DIFF
--- a/gr-fec/python/fec/polar/channel_construction.py
+++ b/gr-fec/python/fec/polar/channel_construction.py
@@ -31,7 +31,7 @@ from helper_functions import *
 try:
     from channel_construction_awgn import tal_vardy_tpm_algorithm
 except ImportError:
-    print "SciPy missing. Overwrite Tal-Vardy algorithm with BEC approximation"
+    print("SciPy missing. Overwrite Tal-Vardy algorithm with BEC approximation")
     def tal_vardy_tpm_algorithm(block_size, design_snr, mu):
         return bhattacharyya_bounds(design_snr, block_size)
 
@@ -123,7 +123,7 @@ def load_z_parameters(block_size, design_snr, mu):
 
 def main():
     np.set_printoptions(precision=3, linewidth=150)
-    print 'channel construction Bhattacharyya bounds by Arikan'
+    print('channel construction Bhattacharyya bounds by Arikan')
     n = 10
     m = 2 ** n
     k = m // 2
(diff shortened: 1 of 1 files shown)


gr-fec/python/fec/polar/channel_construction.py
```diff
@@ -27,8 +27,14 @@
 from channel_construction_bec import calculate_bec_channel_capacities
 from channel_construction_bec import design_snr_to_bec_eta
 from channel_construction_bec import bhattacharyya_bounds
-from channel_construction_awgn import tal_vardy_tpm_algorithm
 from helper_functions import *
+try:
+    from channel_construction_awgn import tal_vardy_tpm_algorithm
+except ImportError:
+    print "SciPy missing. Overwrite Tal-Vardy algorithm with BEC approximation"
```
mbr0wn: This print statement won't work with Py3k. Can you just add parens -- it's not quite the correct fix (you'd normally need to `from __future__ import print_function` as well) but it'll mean this particular hunk won't break Py3k.

jdemel: I just went through this file.
The print statements are mixed. Some are `print '...'` and some are `print('...')`.
I'd fix that file by consistently changing it to `print('...')`. Without `from __future__ import print_function` some prints might look weird, but I'd argue this is not a problem.

mbr0wn: Agree it's not a problem. We'll probably drop Py2k at some time anyway. However, if you want to change *all* the print statements in this file, I'd ask you to make 2 commits, one with the prints, and then the second with your fix. Thanks!

jdemel: The lines with print statements are: 34 (the one we discuss), 75 (already `print('...')`), 126 (needs to be fixed.), 135 (already `print('...')`).
All in all, I'd fix line 34 and 126 to follow Py3k `print_function` convention. That would be one new commit. I don't intend to squash my initial PR commit with this second one. Would that be the way to go?

jdemel:
All `print` statements should now be replaced by `print_function`.

marcusmueller:
I'm fine with the code, waiting on @mbr0wn to clarify whether his code expectations were met.

One thing I'm going to nitpick about: I don't understand the first commit's message body. Can you reword, maybe? Feels like you're missing a word or something.

Regarding conversion of prints on python3: that's already happened on skoslowski/merges/python3

jdemel:
I know. In the first commit message there's a word missing. I had to learn about not using ```` in commit messages the hard way.
The missing word is `tal_vardy_tpm_algorithm`.

mbr0wn:
@jdemel In preparation of the Py3k merge, I'm trying to make sure we don't merge stuff that is harder than necessary to fix. I'm OK with the current state. Thanks for the patch!

PR #1640, opened 2018-02-25
fec: update VOLK API call

jdemel:
VOLK kernel polarbutterfly has an updated API. This patch updates the corresponding call into VOLK.

This patch should only be applied after the next VOLK release. Or after the VOLK subrepo pointer is updated.

### DIFF
--- a/gr-fec/lib/polar_decoder_common.cc
+++ b/gr-fec/lib/polar_decoder_common.cc
@@ -125,7 +125,7 @@ namespace gr {
       polar_decoder_common::butterfly_volk(float* llrs, unsigned char* u, const int stage,
                                            const int u_num, const int row)
       {
-        volk_32f_8u_polarbutterfly_32f(llrs, u, block_size(), block_power(), stage, u_num, row);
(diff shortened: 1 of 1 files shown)


noc0lour:
Actually the subrepo pointer was updated and your patch is very much appreciated!

noc0lour:
Ah, was just updated on `next`. Probably need your patch there.

marcusmueller:
@noc0lour indeed, I only upgraded VOLK on next, because on Master, I'd right now only do it as soon as @n-west makes a new VOLK release.

so, @n-west: how far-fetched is the possibility of you tagging a VOLK release?

marcusmueller:
This got cherry-picked onto next, but we still need to merge it in master to release 3.7.12. @n-west, I'd really need that VOLK release this or next week?

noc0lour:
The volk 1.3.1 release does not contain the necessary API update. Does that mean we can defer this for the `3.7` release series?

marcusmueller:
no, this is still a bug that needs to be fixed in 3.7.12. I was expecting the release to contain that API update :(

PR #1646, opened 2018-02-27
gr-fec: fix use of Volk API for volk_32f_8u_polarbutterfly_32f

michaelld:
subject says it all. simple fix.


noc0lour:
Actually the same as #1640

michaelld:
OK thx will close this one.

PR #1655, opened 2018-02-28
blocks: Keep 1 in N: raise on n<1, fix docs, tabs to spaces

fwunsch:


gr-blocks/include/gnuradio/blocks/keep_one_in_n.h
```diff
@@ -30,7 +30,7 @@ namespace gr {
   namespace blocks {
 
     /*!
-     * \brief decimate a stream, keeping one item out of every \p n.
+     * \brief decimate a stream, keeping the last item out of every \p n.
```
mbr0wn: Wow, is this what it does? Might want to consider picking the first, for consistency with keep M in N and probably most people's intuition.

PR #1657, opened 2018-02-28
qtgui: Fix some -Wreorder warnings

mbr0wn:


gr-qtgui/lib/sink_c_impl.cc
```diff
@@ -58,13 +58,13 @@ namespace gr {
...
-	d_fftsize(fftsize),
-	d_wintype((filter::firdes::win_type)(wintype)),
-	d_center_freq(fc), d_bandwidth(bw), d_name(name),
-	d_plotfreq(plotfreq), d_plotwaterfall(plotwaterfall),
-	d_plottime(plottime), d_plotconst(plotconst),
-  d_parent(parent),
-  d_port(pmt::mp("freq"))
+    d_fftsize(fftsize),
+    d_wintype((filter::firdes::win_type)(wintype)),
+    d_center_freq(fc), d_bandwidth(bw), d_name(name),
```
marcusmueller: Split that line while you're at it?

noc0lour:
@marcusmueller I'd only enforce the one-line-per-initializer-list-element if we want to keep that for future use with `.clang-format`. https://github.com/noc0lour/gnuradio/blob/be9d6a4713bba86f8d2783d0f4f4ec218efdc129/gr-qtgui/lib/sink_c_impl.cc#L63
Which (with the current `.clang-format` is not the case).

mbr0wn:
I like the 1lpile (it's a personal preference of mine, too) and I don't mind changing the formatting. But how about we agree on a clang-format first, then we can do the reformatting in different commit? Then this commit is purely about fixing those warnings.
I would also update the .clang-format from my PR to add this rule.

noc0lour:
Sure, I just wanted to point out that manual formatting shouldn't be the number one priority right now :)

mbr0wn:
@noc0lour That's what I was trying to say, too. I agree with the formatting comments, but I would like to push them into our formatting/GREP discussion.

marcusmueller:
Agreed. This doesn't hurt more than it hurt before, so, that's fine with me.

drmpeg:
Just FYI, you missed a few files with -Wreorder warnings.

gr-qtgui/lib/freq_sink_f_impl.cc
gr-qtgui/lib/waterfall_sink_f_impl.cc
gr-qtgui/lib/vector_sink_f_impl.cc

PR #1658, opened 2018-03-01
Add astylerc file to dtools

mbr0wn:
This file encodes the formatting standards for GNU Radio as an astyle
options file.

### DIFF
--- a/dtools/astyle/astylerc
+++ b/dtools/astyle/astylerc
@@ -1,0 +1,22 @@
+# Indent style: 1TBS, 4 spaces per indent
+--style=1tbs
+--indent=spaces=4
+# Attach brackets to namespace statements and classes
+--attach-namespaces
+--attach-classes
+# Don't unnecessarily pad parens...
+--unpad-paren
+# ...except:
+# Pad around operators (i+1 => i + 1)
+--pad-oper
+# Pad between header and follow-up paren (if(x) => if (x))
+--pad-header
+# Don't leave if statements unfollowed by {}
+--add-brackets
+# Keep complex statements on one line, the author probably had a reason
+--keep-one-line-statements
+# Templates: vector< vector<int> > => vector<vector<int>>
+# Not using for now, requires C++11
+#--close-templates
+# Force line breaks
+--max-code-length=100


mbr0wn:
Like [GREP1](https://github.com/gnuradio/greps/blob/master/grep-0001-coding-guidelines.md), this is supposed to be a suggestion. I picked values that seemed sensible and in line with the current state of the GREP. BTW, astyle output seems pretty benign and I would suggest simply running it against the entire code tree once we've agreed on one style.

marcusmueller:
From the top of my head, I like your style. Is it possible to tell astyle not to indent on `namespace {`? I feel like at a tabstop of four spaces, were losing a lot of line real estate to the redundant info that yes, the contents of gr-foo/include/foo/bar.h is actually a member of the `gr::foo` name space

noc0lour:
I'd like to defer this to [GREP 00XX: Code reformatting](https://github.com/gnuradio/greps/pull/9). Code reformatting in a large scale fashion has to be done with some forethought. Running `astyle` or `clang-format` once on a codebase and then manually fixing all the mishaps is not worth the trouble.

I think this `astylerc` is something to start with and maybe even convert it to a similar `.clang-format'. I'd prefer `clang-format` over `astyle` due to clang being aware of `C++` syntax. `astyle` is a lot more flexible, but I've got the impression we don't have a lot of bike shedding on the topic of code formatting so adopting some coding style we can achieve with `.clang-format` is very desirable.

mbr0wn:
@marcusmueller From what I figured out, astyle will just keep the current indenting.
@noc0lour I'd like to get some movement here, but I have no stakes in astyle. It took me a few minutes to make this (FYI, this was after I submitted the #1657 PR, which drove me completely nuts because of its crazy formatting).

As for GREPs, there is @noc0lour's reformatting GREP, which is about the process, and GREP1, which will include formatting guidelines. If we can express our formatting guidelines in code (e.g., a clang-format config file) that seems to be superior than expressing it in English.

And yes, everyone is agreeing on the important stuff here and not futzing around about the less important stuff... I'm very happy about this. :heart: working with you d00ds!

mbr0wn:
Don't think we need this anymore.

PR #1660, opened 2018-03-02
fix: purge boost/math/special_functions/erf.hpp

jdemel:
`boost/math/special_functions/erf.hpp` causes compile errors. Also, it is only used for two functions `erf` and `erfc`. Both functions are available in `cmath`. Including this boost header back in 2012 might have fixed an issue, but now it causes huge trouble. A 1500+ line compile error with:
`/usr/include/boost/math/special_functions/detail/erf_inv.hpp:355:50: error: invalid suffix Q on floating constant
             boost::math::erfc_inv(static_cast<T>(BOOST_MATH_BIG_CONSTANT(T, 64, 1e-900)), Policy());`

### DIFF
--- a/gr-analog/lib/cpm.cc
+++ b/gr-analog/lib/cpm.cc
@@ -24,13 +24,10 @@
 #include "config.h"
 #endif
 
+#include <gnuradio/analog/cpm.h>
 #include <cmath>
 #include <cfloat>
-#include <gnuradio/analog/cpm.h>
 
-//gives us erf on compilers without it
-#include <boost/math/special_functions/erf.hpp>
-namespace bm = boost::math;
 
 namespace gr {
   namespace analog {
@@ -183,8 +180,8 @@ namespace gr {
       double alpha = 5.336446256636997 * bt;
       for(unsigned i = 0; i < samples_per_sym * L; i++) {
 	double k =  i - Ls/2; // Causal to acausal
-	taps_d[i] = (bm::erf(alpha * (k / samples_per_sym + 0.5)) -
-		     bm::erf(alpha * (k / samples_per_sym - 0.5)))
+        taps_d[i] = (erf(alpha * (k / samples_per_sym + 0.5)) -
+                     erf(alpha * (k / samples_per_sym - 0.5)))
 	  * 0.5 / samples_per_sym;
 	taps[i] = (float) taps_d[i];
       }
--- a/gr-analog/lib/cpm.cc
+++ b/gr-analog/lib/cpm.cc
@@ -36,14 +36,14 @@ namespace gr {
 #  define M_TWOPI (2*M_PI)
 #endif
 
-    //! Normalised sinc function, sinc(x)=sin(pi*x)/pi*x
+    //! Normalised sinc function, sinc(x)=std::sin(pi*x)/pi*x
     inline double
     sinc(double x)
     {
       if(x == 0) {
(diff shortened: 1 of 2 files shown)


marcusmueller:
changing this to go into master. We can't have cmath's erfc in master – that's 3.7, and that means no C++11!

marcusmueller:
With that being done, now your PR simply doesn't work anymore :( @jdemel, can you fix this by cherrypicking from next?

gr-analog/lib/cpm.cc
```diff
@@ -183,8 +180,8 @@ namespace gr {
       double alpha = 5.336446256636997 * bt;
       for(unsigned i = 0; i < samples_per_sym * L; i++) {
 	double k =  i - Ls/2; // Causal to acausal
-	taps_d[i] = (bm::erf(alpha * (k / samples_per_sym + 0.5)) -
-		     bm::erf(alpha * (k / samples_per_sym - 0.5)))
+        taps_d[i] = (erf(alpha * (k / samples_per_sym + 0.5)) -
```
mbr0wn: Shouldn't this be `std::erf()`?

jdemel: First, it does compile with GCC5.4 without and with the `std` namespace.
Second, http://www.cplusplus.com/reference/cmath/erfc/ claims `'erf` is correct.
Third, http://en.cppreference.com/w/cpp/numeric/math/erf claims `std::erf` is correct.

All in all, this is really tricky. I'd go for the `erf` version since it works. If it breaks on other systems, we need to investigate more.

jdemel:
I updated the base branch from `master` to `next`. This is also reflected by the fact that this is actually a 1 commit PR. Unlike the 40 or so commits that were previously mentioned.

gr-analog/lib/cpm.cc
marcusmueller: Cplusplus.com is known for misleading examples. The example on that page is C99, not C++11. It probably only works because we're including `math.h` or `tgmath` somewhere else. We really ought to go directly for `std::erf`– it's the overloaded, type safe, C++ way of doing this :)

marcusmueller:
Thanks! I will quickly do the `std::`modification that I'd like to see here, and if you agree with my version, this'll get merged quickly :)

noc0lour:
But then we have to use <math> not <cmath>, right?

marcusmueller:
I'm getting a segfault on qa_qtgui; is that special to this patch or does that happen on `next` generally, @noc0lour ?

noc0lour:
Hm, shouldn't happen I tnink, but this patch isn't really touching qtgui stuff...

marcusmueller:
… aside from ber_sink...

noc0lour:
hm, didn't saw the log change before... qtgui is excluded from testing right now since testing X is currently not possible, due to bugs in xvfbwrapper

marcusmueller:
In the meantime, I've buiit next without the patch in two diff configurations, and get the same trouble. So, that's a bug, but it's not in this patchset.

marcusmueller:
Fixed the commit msg, merged manually. Closing! Thanks everyone :) 

Note: `clang-tidy` can do a lot good with improving usage of libc math functions:

[performance-type-promotion-in-math-fn](http://clang.llvm.org/extra/clang-tidy/checks/performance-type-promotion-in-math-fn.html)

PR #1662, opened 2018-03-03
add missing includes to headers + reorder qtgui includes

noc0lour:
Add includes which are missing and will be detected upon correct reordering of includes in implementation files.

### DIFF
--- a/gr-qtgui/lib/sink_c_impl.h
+++ b/gr-qtgui/lib/sink_c_impl.h
@@ -1,6 +1,6 @@
 /* -*- c++ -*- */
 /*
- * Copyright 2008,2009,2011,2012,2014 Free Software Foundation, Inc.
+ * Copyright 2008,2009,2011,2012,2014,2018 Free Software Foundation, Inc.
  *
  * This file is part of GNU Radio
  *
@@ -24,6 +24,7 @@
 #define INCLUDED_QTGUI_SINK_C_IMPL_H
 
 #include <gnuradio/qtgui/sink_c.h>
+
 #include <gnuradio/filter/firdes.h>
 #include <gnuradio/fft/fft.h>
 #include <gnuradio/high_res_timer.h>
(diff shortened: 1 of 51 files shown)


marcusmueller:
I'm fine with the spirit of these changes, but an ever so slightly more descriptive commit msg wouldn't hurt? So, don't call things that aren't made with `git commit --fixup=` "fixup:", and maybe something like "headers: apply include-what-you-use principle" or something. Let your creativity flow.

noc0lour:
Thanks, I'll name my commit appropriately afterwards, these changes are just a precursor to automatic formatting which may or may not come soon.

marcusmueller:
are you doing this automatically or manually? If manually, have you heard of the iwyu tool? I find it cumbersome at times, but it is a nice tool after all.

gr-qtgui/lib/sink_c_impl.h
```diff
@@ -1,6 +1,6 @@
 /* -*- c++ -*- */
 /*
- * Copyright 2008,2009,2011,2012,2014 Free Software Foundation, Inc.
+ * Copyright 2008,2009,2011,2012,2014,2018 Free Software Foundation, Inc.
```
noc0lour: Not sure if I have to fix all copyright headers of files I touched ...

noc0lour:
Headers in qtgui did not have any particular grouping. This was pretty bad especially as the public interface usually included `Python.h` which _has_ to be the first include due to missing include guards.
Using any automated tool will sort includes lexicographic and move the `Python.h` include further back.

This will produce a lot of new warnings due to redifinitions.

That's the rationale for changing a lot of files manually in qtgui.

gr-qtgui/lib/sink_c_impl.h
marcusmueller: Luckily, especially for minor or automated fixes, "latest" copyright dates are of lesser importance.

PR #1663, opened 2018-03-03
Add clang_format_tool to tools/

noc0lour:
Imported the tools used by MongoDB developers to clean up their tree with some tweaks already for the GR source tree.

Using #1662 as base it will format all (.h|cc|c) files according to the `.clang-format` defined it the root directory of the source tree.

I'm not yet happy with the commandline interface and some intermediate functions have to be improved to pass my arbitrary "I can live with
that type of Python" standard.

I want to show in which direction repo tooling can go and get you guys involved/informed early.

### DIFF
--- a/tools/clang_format.py
+++ b/tools/clang_format.py
@@ -1,0 +1,883 @@
+#!/usr/bin/env python
+"""
+A script that provides:
+1. Ability to grab binaries where possible from LLVM.
+2. Ability to download binaries from MongoDB cache for clang-format.
+3. Validates clang-format is the right version.
+4. Has support for checking which files are to be checked.
+5. Supports validating and updating a set of files to the right coding style.
+"""
+from __future__ import print_function, absolute_import
+
+import Queue
+import difflib
+import glob
+import itertools
+import os
+import re
+import shutil
+import string
+import subprocess
+import sys
+import tarfile
+import tempfile
+import threading
+import time
+import urllib
+from distutils import spawn
+from optparse import OptionParser
+from multiprocessing import cpu_count
+
+# Get relative imports to work when the package is not installed on the PYTHONPATH.
+if __name__ == "__main__" and __package__ is None:
+    sys.path.append(os.path.dirname(os.path.dirname(os.path.abspath(os.path.realpath(__file__)))))
+
+from buildscripts import moduleconfig
+
+
+##############################################################################
+#
+# Constants for clang-format
+#
+#
+
+# Expected version of clang-format
+CLANG_FORMAT_VERSION = "3.8.0"
+CLANG_FORMAT_SHORT_VERSION = "3.8"
+
+# Name of clang-format as a binary
+CLANG_FORMAT_PROGNAME = "clang-format"
+
+# URL location of the "cached" copy of clang-format to download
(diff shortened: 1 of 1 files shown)


noc0lour:
Need to test this tool with Python2 now, since I'm developing on a Python3-centric machine

noc0lour:
Also I'd like to have a `tools/` directory with tools not only useful for binary packaging GNU Radio but all sorts of tools. (e.g. maintainance, check-patch, ...)

tools/clang_format.py
```diff
@@ -42,27 +50,15 @@
 #
 
 # Expected version of clang-format
-CLANG_FORMAT_VERSION = "3.8.0"
-CLANG_FORMAT_SHORT_VERSION = "3.8"
+CLANG_FORMAT_VERSION = "5.0.1"
+CLANG_FORMAT_SHORT_VERSION = "5.0"
```
marcusmueller: So, this is OK because you know we're running this only on machines with cool distros, right?

marcusmueller:
Are you sure https://github.com/gnuradio/gnuradio/pull/1663/commits/5969e380c99d1934ef793fdd2defdfef679947c3 should be part of this PR?

tools/clang_format.py
noc0lour: I need to figure out which version of clang-format we want to dictate, so formatting will be the same for all contributors.

noc0lour:
The commit is necessary if you want to run the tool and compile the result afterwards.

tools/clang_format.py
marcusmueller: well, that's then going to be a problem – different packagers package different versions. I wonder if we can have something like a "reformatting" push remote, which lets you push, then wait a couple second, then lets you pull, giving you a `--fixup` commit with the reformatted / `clang-tidy`ied files.

noc0lour: Across all major distributions you have access to at least two or more compiler major versions.

noc0lour: I's 4.0.1 now, which I believe we can expect people to have available.

noc0lour:
Using `python3` it should be sane now to run `python3 ./tools/clang_format.py format` using this branch and look at the diff. clang-format needs to run twice due to grave issues in qtgui code. 
Then `python3 ./tools/clang_format lint` should remain silent.

If python3 is your default python no prefixing of python3 is needed for executing the tool.

noc0lour:
Branch reformatting (updating a branch created before tree reformatting happened) is not yet working.

noc0lour:
This tool has two major modes:
 `format` - will format files in place
 `lint` - will show what is not up to the standard defined in `.clang-format`

`format` also has a special  "rebase" mode `-b` which will update a branch based on the commit before the codebase reformatting to base on the commit which did the formatting. `-b $hash-before $hash-after`.

tools/clang_format.py
mbr0wn: Agreed. FYI, we did the same for UHD.

mbr0wn:
Once #1664 is merged, this can be rebased and merged too, right? I'm fine with the state. It's all new code, it can't break anything.

PR #1664, opened 2018-03-03
formatting: Import clang-format -style=llvm -dump-config & add agreed on values

noc0lour:
Similar the previous of  @mbr0wn . This time import the default llvm style and add already modified values. Now comment on values/discuss what we would like to change compared to the default.

It's possible these list is not complete, in this case please add more flags to the discussion.

Information about possible values and meaning can be found here: 
http://releases.llvm.org/3.9.0/tools/clang/docs/ClangFormatStyleOptions.html

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,20 +1,104 @@
 ---
-BasedOnStyle: LLVM
-IndentWidth: 4
----
 Language: Cpp
-BasedOnStyle: LLVM
-IndentWidth: 4
-AccessModifierOffset: 0
+# BasedOnStyle: LLVM
+AccessModifierOffset: -4
 AlignAfterOpenBracket: Align
-#AlignEscapedNewlines: Left
+AlignConsecutiveAssignments: false
+AlignConsecutiveDeclarations: false
+AlignEscapedNewlinesLeft: true
+AlignOperands:   true
 AlignTrailingComments: true
+AllowAllParametersOfDeclarationOnNextLine: true
+AllowShortBlocksOnASingleLine: false
+AllowShortCaseLabelsOnASingleLine: false
+AllowShortFunctionsOnASingleLine: All
+AllowShortIfStatementsOnASingleLine: false
+AllowShortLoopsOnASingleLine: false
+AlwaysBreakAfterDefinitionReturnType: None
+AlwaysBreakAfterReturnType: None
+AlwaysBreakBeforeMultilineStrings: false
 AlwaysBreakTemplateDeclarations: true
 BinPackArguments: false
 BinPackParameters: false
-ColumnLimit: 100
-#FixNamespaceComments: true
+BreakBeforeBraces: Custom
+BraceWrapping:
+  AfterClass:      true
+  AfterControlStatement: false
+  AfterEnum:       false
+  AfterFunction:   true
+  AfterNamespace:  false
+  AfterObjCDeclaration: false
+  AfterStruct:     false
+  AfterUnion:      false
+  BeforeCatch:     false
+  BeforeElse:      false
+  IndentBraces:    false
+BreakBeforeBinaryOperators: None
+BreakBeforeTernaryOperators: true
+BreakConstructorInitializersBeforeComma: false
+BreakAfterJavaFieldAnnotations: false
+BreakStringLiterals: true
+ColumnLimit:     90
+CommentPragmas:  '^ IWYU pragma:'
+ConstructorInitializerAllOnOneLineOrOnePerLine: true
+ConstructorInitializerIndentWidth: 4
+ContinuationIndentWidth: 4
+Cpp11BracedListStyle: false
+DerivePointerAlignment: false
+DisableFormat:   false
+ExperimentalAutoDetectBinPacking: false
+ForEachMacros:
+  - foreach
+  - Q_FOREACH
+  - BOOST_FOREACH
+IncludeCategories:
+  - Regex:           '^"(gnuradio)/'
+    Priority:        1
+  - Regex:           '^<(gnuradio)/'
+    Priority:        2
+  - Regex:           '^<(boost)/'
+    Priority:        98
+  - Regex:           '^<[a-z]*>$'
+    Priority:        99
+  - Regex:           '^".*"$'
+    Priority:        0
+  - Regex:           '.*'
+    Priority:        10
+
+IncludeIsMainRegex: '(Test)?$'
+IndentCaseLabels: false
+IndentWidth: 4
+IndentWrappedFunctionNames: false
+JavaScriptQuotes: Leave
+JavaScriptWrapImports: true
+KeepEmptyLinesAtTheStartOfBlocks: true
+MacroBlockBegin: ''
+MacroBlockEnd:   ''
 MaxEmptyLinesToKeep: 2
-#CompactNamespaces: true
 NamespaceIndentation: None
+ObjCBlockIndentWidth: 2
+ObjCSpaceAfterProperty: false
+ObjCSpaceBeforeProtocolList: true
+PenaltyBreakBeforeFirstCallParameter: 19
+PenaltyBreakComment: 300
+PenaltyBreakFirstLessLess: 120
+PenaltyBreakString: 1000
+PenaltyExcessCharacter: 1000000
+PenaltyReturnTypeOnItsOwnLine: 60
 PointerAlignment: Left
+ReflowComments:  true
+SortIncludes:    true
+SpaceAfterCStyleCast: false
+SpaceAfterTemplateKeyword: true
+SpaceBeforeAssignmentOperators: true
+SpaceBeforeParens: ControlStatements
+SpaceInEmptyParentheses: false
+SpacesBeforeTrailingComments: 1
+SpacesInAngles:  false
+SpacesInContainerLiterals: true
+SpacesInCStyleCastParentheses: false
+SpacesInParentheses: false
+SpacesInSquareBrackets: false
+Standard:        Cpp11
+TabWidth:        8
+UseTab:          Never


.clang-format
```diff
@@ -4,17 +4,106 @@ IndentWidth: 4
...
+BreakBeforeInheritanceComma: false
+BreakBeforeTernaryOperators: true
+BreakConstructorInitializersBeforeComma: false
+BreakConstructorInitializers: BeforeColon
+BreakAfterJavaFieldAnnotations: false
+BreakStringLiterals: true
+ColumnLimit:     100
+CommentPragmas:  '^ IWYU pragma:'
+CompactNamespaces: true
+ConstructorInitializerAllOnOneLineOrOnePerLine: false
```
mbr0wn: Suggest: `true`

.clang-format
```diff
@@ -4,17 +4,106 @@ IndentWidth: 4
 ---
 Language: Cpp
 BasedOnStyle: LLVM
-IndentWidth: 4
 AccessModifierOffset: 0
```
mbr0wn: Suggest: `-4`

.clang-format
```diff
@@ -4,17 +4,106 @@ IndentWidth: 4
...
+BreakBeforeBraces: Attach
+BreakBeforeInheritanceComma: false
+BreakBeforeTernaryOperators: true
+BreakConstructorInitializersBeforeComma: false
+BreakConstructorInitializers: BeforeColon
+BreakAfterJavaFieldAnnotations: false
+BreakStringLiterals: true
+ColumnLimit:     100
+CommentPragmas:  '^ IWYU pragma:'
+CompactNamespaces: true
```
mbr0wn: This failed on my version of clang-format

.clang-format
```diff
@@ -4,17 +4,106 @@ IndentWidth: 4
...
+Cpp11BracedListStyle: true
+DerivePointerAlignment: false
+DisableFormat:   false
+ExperimentalAutoDetectBinPacking: false
+FixNamespaceComments: true
+ForEachMacros:
+  - foreach
+  - Q_FOREACH
+  - BOOST_FOREACH
+IncludeCategories:
```
mbr0wn: If we want IncludeCategories, we should fix this list.

.clang-format
```diff
@@ -4,17 +4,106 @@ IndentWidth: 4
...
+SpaceBeforeAssignmentOperators: true
+SpaceBeforeParens: ControlStatements
+SpaceInEmptyParentheses: false
+SpacesBeforeTrailingComments: 1
+SpacesInAngles:  false
+SpacesInContainerLiterals: true
+SpacesInCStyleCastParentheses: false
+SpacesInParentheses: false
+SpacesInSquareBrackets: false
+Standard:        Cpp11
```
mbr0wn: I'd love to backport this to older branches and then we must remember to change this.

.clang-format
marcusmueller: I second that suggestion.

.clang-format
mbr0wn: For clarity, this will make `public` etc. look like this:
```cpp
class Foo
{
public:
    Foo();
private:
    void f();
};
```

noc0lour: Note: This was about `AccessModifierOffset`.

.clang-format
```diff
@@ -1,20 +1,102 @@
...
+    Priority:        1
+  - Regex:           '^<(gnuradio)/'
+    Priority:        2
+  - Regex:           '^<(boost)/'
+    Priority:        98
+  - Regex:           '^<[a-z]*>$'
+    Priority:        99
+  - Regex:           '.*'
+    Priority:        10
+
```
noc0lour: Include Categories now sort: 
 - "gnuradio/something.h"
 - <gnuradio/something_else.h>
 - <everythingelse.h(pp)>
 - <boost/boost_module.hpp>
 - <mycstdheader>

noc0lour:
I added your suggestions in and made `.clang-format` clang4 compatible. I'm thinking it is sane to expect core developers to be able to get a (binary or compiled) clang-format version 4.x.x

.clang-format
noc0lour: The Code in `Format.cpp` looks good though and my toy examples with used functions conclude it should detect the header as main include ...

noc0lour: Ok, figured out that the main header is only detected if it is in the first block. Unfortunately `config.h` is the only header in the first block most of the time.

noc0lour:
If we implement #1680 we can actually remove `#include config.h` from any source file.

.clang-format
```diff
@@ -1,20 +1,102 @@
...
+  - Q_FOREACH
+  - BOOST_FOREACH
+IncludeCategories:
+  - Regex:           '^"(gnuradio)/'
+    Priority:        1
+  - Regex:           '^<(gnuradio)/'
+    Priority:        2
+  - Regex:           '^<(boost)/'
+    Priority:        98
+  - Regex:           '^<[a-z]*>$'
```
noc0lour: Probably `[a-z]+` will be better

.clang-format
```diff
@@ -1,20 +1,102 @@
...
+AlignOperands:   true
 AlignTrailingComments: true
+AllowAllParametersOfDeclarationOnNextLine: true
+AllowShortBlocksOnASingleLine: false
+AllowShortCaseLabelsOnASingleLine: false
+AllowShortFunctionsOnASingleLine: All
+AllowShortIfStatementsOnASingleLine: false
+AllowShortLoopsOnASingleLine: false
+AlwaysBreakAfterDefinitionReturnType: None
+AlwaysBreakAfterReturnType: None
```
noc0lour: Specifying `AlwaysBreakAfter*Type` will allow having  all fucntions & declarations with the `ReturnType` on a separate line. 
With the current settings it will be determined automatically taking function name length etc into acconut.

.clang-format
```diff
@@ -1,20 +1,102 @@
...
+AlwaysBreakBeforeMultilineStrings: false
 AlwaysBreakTemplateDeclarations: true
 BinPackArguments: false
 BinPackParameters: false
-ColumnLimit: 100
-#FixNamespaceComments: true
+BraceWrapping:
```
noc0lour: note: BraceWrapping block is not used due to `BreakBeforeBraces: Attach`.

.clang-format
```diff
@@ -1,20 +1,102 @@
...
+  AfterFunction:   false
+  AfterNamespace:  false
+  AfterObjCDeclaration: false
+  AfterStruct:     false
+  AfterUnion:      false
+  BeforeCatch:     false
+  BeforeElse:      false
+  IndentBraces:    false
+BreakBeforeBinaryOperators: None
+BreakBeforeBraces: Attach
```
noc0lour: We don't want braces on a separate line, right? Then this setting is the one to choose.

.clang-format
```diff
@@ -1,20 +1,102 @@
...
+BreakBeforeTernaryOperators: true
+BreakConstructorInitializersBeforeComma: false
+BreakAfterJavaFieldAnnotations: false
+BreakStringLiterals: true
+ColumnLimit:     100
+CommentPragmas:  '^ IWYU pragma:'
```
noc0lour: If we start using IWYU to figure out which includes are useful and necessary this is good.

.clang-format
```diff
@@ -1,20 +1,102 @@
...
+ConstructorInitializerIndentWidth: 4
+ContinuationIndentWidth: 4
+Cpp11BracedListStyle: true
+DerivePointerAlignment: false
+DisableFormat:   false
+ExperimentalAutoDetectBinPacking: false
+ForEachMacros:
+  - foreach
+  - Q_FOREACH
+  - BOOST_FOREACH
```
noc0lour: Do we have more foreach Macros we want to detect?

.clang-format
```diff
@@ -1,20 +1,102 @@
...
 NamespaceIndentation: None
+ObjCBlockIndentWidth: 2
+ObjCSpaceAfterProperty: false
+ObjCSpaceBeforeProtocolList: true
+PenaltyBreakBeforeFirstCallParameter: 19
+PenaltyBreakComment: 300
+PenaltyBreakFirstLessLess: 120
+PenaltyBreakString: 1000
+PenaltyExcessCharacter: 1000000
+PenaltyReturnTypeOnItsOwnLine: 60
```
noc0lour: If we want to modify behaviour of auto line breaks we need to fiddle with Penalty settings.

.clang-format
```diff
@@ -1,20 +1,102 @@
...
 PointerAlignment: Left
+ReflowComments:  true
```
noc0lour: We nee do take care of not misformatting doxygen documentation. By disabling formatting at comments which contain special formatting.

mbr0wn: Yeah good point. Although people should also keep their comment strings below 100 characters... I have no strong opinion on this option.

.clang-format
mbr0wn: Or fewer? `BOOST_FOREACH` should be outlawed on master once we have regular `foreach` (along with many other Boost-isms :skull_and_crossbones: )

.clang-format
mbr0wn: Whatever the LLVM default is.

mbr0wn:
I think this is good to go.

noc0lour:
I'm fine with the state of this proposal as is. Only thing we have to figure out is the `config.h` and main include header situation.

.clang-format
mbr0wn: Actually, almost all of our class definitions have the class braces on a separate line, i.e., this is one of the few things we've done consistently so far. I vote for keeping the current behaviour (i.e., change this to "Linux" or something like that).

.clang-format
marcusmueller: So, we change this to `false`, right?

.clang-format
marcusmueller: @mbr0wn that we can achieve by using `BraceWrapping`'s `AfterClass`, and I'd agree, we should be doing that.

.clang-format
marcusmueller: Can we change that and use `Custom` instead of `Attach`?

.clang-format
marcusmueller: Also, the current setting does that, but  we need to change `BreakBeforeBraces` to `Custom`.

mbr0wn:
FYI, I find it very useful to actually run this against large portions of the source tree and see if I like the outcome. If the script changes stuff that we have some kind of consistency for, then I think we should keep the original consistency. That will, somewhat, reduce the diff of the formatting mega-commit, whenever that comes.

.clang-format
```diff
@@ -1,20 +1,102 @@
...
+BreakBeforeTernaryOperators: true
+BreakConstructorInitializersBeforeComma: false
+BreakAfterJavaFieldAnnotations: false
+BreakStringLiterals: true
+ColumnLimit:     100
+CommentPragmas:  '^ IWYU pragma:'
+ConstructorInitializerAllOnOneLineOrOnePerLine: true
+ConstructorInitializerIndentWidth: 4
+ContinuationIndentWidth: 4
+Cpp11BracedListStyle: true
```
dkozel: I found setting this to false caused a significant improvement in cases where arrays are being initialized with long lists.
For example: https://stackoverflow.com/questions/38058627/clang-format-array-initialisers

dkozel: clang-format version 4.0.1-6 works. I don't know about earlier.

noc0lour: I created this dump with clang-format 4.x.x so all configuration options should apply across clang-format 4.x.x

marcusmueller:
I've integrated the further recommendations into the "add suggested" commit. @noc0lour, please review. Will definitely merge this Monday.

noc0lour:
After looking through the clang-format docs and comments here I refined the `BreakBeforeBrace` Setting: Now it is set to `Custom` with breaks configured for `namespace` and `class`. 
Also I added a regex which will sort local includes with quotes to the top.

.clang-format
```diff
@@ -1,20 +1,104 @@
...
+  AfterUnion:      false
+  BeforeCatch:     false
+  BeforeElse:      false
+  IndentBraces:    false
+BreakBeforeBinaryOperators: None
+BreakBeforeTernaryOperators: true
+BreakConstructorInitializersBeforeComma: false
+BreakAfterJavaFieldAnnotations: false
+BreakStringLiterals: true
+ColumnLimit:     100
```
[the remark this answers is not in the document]
noc0lour: I'm fine with really anything else. So 90?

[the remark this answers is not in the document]
mbr0wn: No one complained, so ... sure.

mbr0wn:
Merge, merge, merge!

.clang-format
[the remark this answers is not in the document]
noc0lour: OK 90 it is.

PR #1672, opened 2018-03-05
fix #1587: edit copyright in gr_modtool generated files

### LINKED ISSUE #1587: gr-newmod: Should not contain "This file is part of GNU Radio"
Instead, something like @awalls-cx18 does: "This file was generated by gr_modtool, a tool from the GNU Radio framework", or so.

swap-nil7:
Replaced "This file is part of GNU Radio" with "This file was generated by gr_modtool, a tool from the GNU Radio framework" in the copyright license of gr_modtool generated files.


swap-nil7:
@mbr0wn,  @marcusmueller, can you review this pull request please?

gr-utils/python/modtool/modtool_add.py
```diff
@@ -110,7 +110,9 @@ def setup(self, options, args):
         if not options.license_file:
             self._info['copyrightholder'] = options.copyright
             if self._info['copyrightholder'] is None:
-                self._info['copyrightholder'] = '<+YOU OR YOUR COMPANY+>'
+                self._info['copyrightholder'] = raw_input("Please specify the copyright holder: ")
+                if(len(self._info['copyrightholder'])==0):
```
noc0lour: you could improve the check for an empty (or only space characters) using: 
`if(not self._info['copyrightholder'] or self._info['copyrightholder'].isspace()):`

gr-utils/python/modtool/modtool_add.py
```diff
@@ -110,7 +110,9 @@ def setup(self, options, args):
         if not options.license_file:
             self._info['copyrightholder'] = options.copyright
             if self._info['copyrightholder'] is None:
-                self._info['copyrightholder'] = '<+YOU OR YOUR COMPANY+>'
+                self._info['copyrightholder'] = raw_input("Please specify the copyright holder: ")
```
marcusmueller: Can you surround this with

```
with SequenceCompleter(list_of_sensible_candidates):
```

please, so that people can just "<tab>" to get predefined answers?

This would require you to generate a `list_of_sensible_candidates`, for example by `import getpass; getpass.getuser()` (to get the login user), and by asking `git config --global user.name`.

Does that sound feasible?

marcusmueller:
I quite like what you're doing here, but wouldn't it be cleverer to do that once for `gr_modtool newmod` instead of for every `add`?

noc0lour:
@marcusmueller In that case gr-modtool needs to store some sort of configuration file in the tree.

gr-utils/python/modtool/modtool_add.py
swap-nil7: Can you please suggest me the members for the `list_of_sensible_candidates`?
And shall I make it mandatory for the user to add a copyright holder from the same list?

marcusmueller:
@noc0lour true. Looking back, I remember liking the fact that gr_modtool doesn't add a secondary source of information to the tree, but now I'd see it as a design shortcoming.

@swap-nil7: He's right. We can't save this anywhere in the module right now, so your approach is the only one feasible.

swap-nil7:
@marcusmueller: As of now I have understood that I need to add `list_of_sensible_candidates` for autocompletion of predefined answers. Please correct me if I am wrong.

swap-nil7:
@marcusmueller @noc0lour : I guess we can add `getpass.getuser()` as getpass is an in-built module. Shall I add the feature for auto-completion of login username?

noc0lour:
@swap-nil7 both suggestions (login user & git user) are fine, you can add either one (I guess login user is easier for a start). And if getting the git user looks to daunting someone else can add that too.

swap-nil7:
@noc0lour: I will create a temporary file in which I'll store the output of the OS,  read the file and get the git user and then delete the file. Is this a right approach or shall I implement it using a subprocess?

swap-nil7:
@noc0lour @marcusmueller : Can you please review the pull request?

mbr0wn:
@swap-nil7 Please stop asking for reviews. They'll come as soon as we get around to them.

gr-utils/python/modtool/modtool_add.py
```diff
@@ -110,7 +112,13 @@ def setup(self, options, args):
         if not options.license_file:
             self._info['copyrightholder'] = options.copyright
             if self._info['copyrightholder'] is None:
-                self._info['copyrightholder'] = '<+YOU OR YOUR COMPANY+>'
+                user = getpass.getuser()
+                git_user = subprocess.check_output('git config --global user.name', shell=True).strip()
```
mbr0wn: This assumes the existence of git. It's a fair assumption for GNU Radio devs, but not good practice for code in general.

gr-utils/python/modtool/modtool_add.py
```diff
@@ -110,7 +112,13 @@ def setup(self, options, args):
...
+                copyright_candidates = (user, git_user, 'GNU Radio')
```
mbr0wn: If this is the only place `git_user` is used, I'd recommend to fail gracefully and silently when the git command fails.

gr-utils/python/modtool/modtool_add.py
mbr0wn: ...and maybe it's a better idea to put this log into `scm.py`

gr-utils/python/modtool/modtool_newmod.py
```diff
@@ -90,6 +90,7 @@ def run(self):
                 s = open(f, 'r').read()
                 s = s.replace('howto', self._info['modname'])
                 s = s.replace('HOWTO', self._info['modname'].upper())
+                s = s.replace('This file is part of GNU Radio', 'This file was generated by gr_modtool, a tool from the GNU Radio framework \n * This file is a part of gr-'+self._info['modname'])
```
mbr0wn: Just change the `grlicense` template in templates.py

swap-nil7: I guess just changing the license in grlicense will not solve the issue. We will have to change all files in gr-newmod directory because the script recursively opens all files and rename howto and HOWTO to respective module name. Please correct me if I am wrong.

mbr0wn: You're right, just changing `grlicense` does not suffice. But if we want a different header, we should then change all the headers in gr-newmod -- after all, they too, are templates.

mbr0wn: ...which you already did.

mbr0wn:
CLA received.

gr-utils/python/modtool/scm.py
```diff
@@ -81,6 +81,13 @@ def is_active(self):
         """ Returns true if this repository manager is operating on an active, source-controlled directory. """
         return self.is_empty
 
+    def get_gituser(self):
+        """ Gets the git user """
+        try:
+            return subprocess.check_output('git config --global user.name', shell=True).strip()
+        except (OSError, subprocess.CalledProcessError):
+            return 0
```
marcusmueller: My only complaint: this should be `None` instead of `0`. `0` wouldn't be a string (which is what we're expecting a git user to be), but still semantically different from `None`, even if both evaluate to `False` in an `if` clause :) Can you please replace 0 by `None`?

PR #1689, opened 2018-03-14
Throwing a std::invalid_argument instead of letting SIGFPE happen for…

### LINKED ISSUE #1648: OFDM carrier allocator causes invisible SIGFPE with certain input
Using twice () to signal, that no pilot carriers and no pilot symbols are desired causes a SIGFPE, which is not even visible in grc. Long-term a solution might be to make the constructor saver, but shot-term changing 
`<value>()<\value>`
to 
`<value>((),)<\value>`

in the XML file should suffice to not mislead users to trust the presetting.

m3x1m0m:
… wrong input in case of the OFDM carrier allocator.

Fixes #1648

### DIFF
--- a/gr-digital/lib/ofdm_carrier_allocator_cvc_impl.cc
+++ b/gr-digital/lib/ofdm_carrier_allocator_cvc_impl.cc
@@ -1,6 +1,6 @@
 /* -*- c++ -*- */
 /*
- * Copyright 2013-2018 Free Software Foundation, Inc.
+ * Copyright 2013, 2018 Free Software Foundation, Inc.
  *
  * This file is part of GNU Radio
  *
@@ -74,9 +74,8 @@ namespace gr {
 	d_output_is_shifted(output_is_shifted)
     {
       // Sanity checks
-      // Since C++11: Get pointer to underlying array
       // If that is is null, the input is wrong -> force user to use ((),) in python
-      if (d_occupied_carriers.data() == nullptr) {
+      if (d_occupied_carriers.empty()) {
         throw std::invalid_argument("Occupied carriers must be of type vector of vector i.e. ((),).");
       }
       for (unsigned i = 0; i < d_occupied_carriers.size(); i++) {
@@ -92,7 +91,7 @@ namespace gr {
 	  }
 	}
       }
-      if (d_pilot_carriers.data() == nullptr) {
+      if (d_pilot_carriers.empty()) {
         throw std::invalid_argument("Pilot carriers must be of type vector of vector i.e. ((),).");
       }
       for (unsigned i = 0; i < d_pilot_carriers.size(); i++) {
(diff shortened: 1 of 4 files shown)


gr-digital/lib/ofdm_carrier_allocator_cvc_impl.cc
```diff
@@ -73,6 +73,12 @@ namespace gr {
 	d_symbols_per_set(0),
 	d_output_is_shifted(output_is_shifted)
     {
+      // Sanity checks
+      // Since C++11: Get pointer to underlying array
+      // If that is is null, the input is wrong -> force user to use ((),) in python
+      if (d_occupied_carriers.data() == nullptr) {
```
jdemel: `nullptr` was introduced with C++11. Thus, this commit must go into next aka GR3.8.

gr-digital/lib/ofdm_carrier_allocator_cvc_impl.cc
```diff
@@ -86,6 +92,9 @@ namespace gr {
 	  }
 	}
       }
+      if (d_pilot_carriers.data() == nullptr) {
```
jdemel: `nullptr` C++11 only.

gr-digital/lib/ofdm_carrier_allocator_cvc_impl.h
```diff
@@ -1,19 +1,19 @@
 /* -*- c++ -*- */
-/* 
- * Copyright 2013 Free Software Foundation, Inc.
- * 
+/*
+ * Copyright 2013-2018 Free Software Foundation, Inc.
```
jdemel: I suppose this should be `Copyright 2013, 2018`.
Also, there seems to be no code changes in this file. Why update the copyright notice then?

gr-digital/python/digital/qa_ofdm_carrier_allocator_cvc.py
```diff
@@ -208,6 +208,32 @@ def test_003_t (self):
                 self.assertEqual(correct_offsets[key], tag.offset)
         self.assertTrue(all(tags_found.values()))
 
+    @gr_unittest.unittest.skip("Skipping test with wrong input (caused SIGFPE earlier, now throws a simple std::invalid_argument with useful message)")
```
jdemel: Instead of skippping this test, could you call the constructor with invalid arguments and then catch the correct exception? Also, are these arguments interdependend? This should be checked in the constructor as well. Then write some tests that try to call the constructor with invalid arguments and check if the appropriate exception is thrown.
`unittest.TestCase.assertRaises` seems to be the correct check here.

gr-digital/lib/ofdm_carrier_allocator_cvc_impl.cc
m3x1m0m: Is there no C++11 code at all up to 3.12.7?

noc0lour: 3.7.12, but yes: no C++11 until 3.8 due to compatibility with GCC 4.x which lacks reasonable C++11 support.

m3x1m0m:
@jdemel Thank you for the feedback. I appreciated it very much. The SIGFPE is now avoided with code compliant to older C++ versions and the test case checks for the three different cases with wrong input that earlier caused the SIGFPE.

jdemel:
The PR looks really good now. Thanks for your effort!

PR #1692, opened 2018-03-15
blocks: replace rand48 with xoroshiro closes #1671

### LINKED ISSUE #1671: Get rid of non-threadsafe `rand48`.
`rand48` is as non-threadsafe as `rand`. And I should feel bad for not having gotten rid of it when I purged `rand` from `fastnoise_source`.  This applies to "gr-blocks/lib/test_tag_variable_rate_ff_impl.cc"

This includes the feature test. Bonus: this makes the CMake step faster (if only by a little degree. Every little counts here, as it's my nerves :) ).

noc0lour:
closes #1671

### DIFF
--- a/gr-blocks/lib/ConfigChecks.cmake
+++ b/gr-blocks/lib/ConfigChecks.cmake
@@ -89,10 +89,3 @@ CHECK_CXX_SOURCE_COMPILES("
     " HAVE_COSF
 )
 GR_ADD_COND_DEF(HAVE_COSF)
-
-CHECK_CXX_SOURCE_COMPILES("
-    #include <stdlib.h>
-	int main(){srand48(0); drand48(); lrand48(); return 0;}
-	" HAVE_RAND48
-)
-GR_ADD_COND_DEF(HAVE_RAND48)
--- a/gr-blocks/lib/test_tag_variable_rate_ff_impl.cc
+++ b/gr-blocks/lib/test_tag_variable_rate_ff_impl.cc
@@ -26,10 +26,12 @@
 
 #include "test_tag_variable_rate_ff_impl.h"
 #include <gnuradio/io_signature.h>
+#include <gnuradio/xoroshiro128p.h>
 #include <string.h>
 #include <iostream>
 #include <iomanip>
 #include <stdexcept>
+#include <cstdint>
 
 using namespace pmt;
 
@@ -57,11 +59,7 @@ namespace gr {
       d_new_in = 0;
       d_last_out = 0;
 
-#ifdef HAVE_RAND48
-      srand48(time(NULL));
-#else
-      srand(time(NULL));
-#endif
+      xoroshiro128p_seed(d_rng_state, 4 /* chosen by fair dice roll */);
     }
 
     test_tag_variable_rate_ff_impl::~test_tag_variable_rate_ff_impl()
@@ -82,11 +80,7 @@ namespace gr {
       GR_LOG_DEBUG(d_logger, boost::format("noutput_items: %1%") % noutput_items);
 
       if(d_update_once) {
-#ifdef HAVE_RAND48
-        if(drand48() > 0.5) {
-#else
-        if (rand() > RAND_MAX / 2) {
-#endif
+        if (xoroshiro128p_next(d_rng_state) > (UINT64_MAX / 2)){
           d_rrate += d_update_step;
         }
         else {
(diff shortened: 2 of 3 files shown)


gr-blocks/lib/ConfigChecks.cmake
```diff
@@ -89,10 +89,3 @@ CHECK_CXX_SOURCE_COMPILES("
     " HAVE_COSF
 )
 GR_ADD_COND_DEF(HAVE_COSF)
-
-CHECK_CXX_SOURCE_COMPILES("
-    #include <stdlib.h>
-	int main(){srand48(0); drand48(); lrand48(); return 0;}
-	" HAVE_RAND48
-)
-GR_ADD_COND_DEF(HAVE_RAND48)
```
marcusmueller: Yes, every deleted line is a good line.

gr-blocks/lib/test_tag_variable_rate_ff_impl.cc
```diff
@@ -57,11 +59,7 @@ namespace gr {
       d_new_in = 0;
       d_last_out = 0;
 
-#ifdef HAVE_RAND48
-      srand48(time(NULL));
-#else
-      srand(time(NULL));
-#endif
+      xoroshiro128p_seed(d_state, (uint64_t) time(NULL));
```
marcusmueller: Nah. Is there any advantage to random seeding?

gr-blocks/lib/test_tag_variable_rate_ff_impl.cc
```diff
@@ -82,11 +80,7 @@ namespace gr {
       GR_LOG_DEBUG(d_logger, boost::format("noutput_items: %1%") % noutput_items);
 
       if(d_update_once) {
-#ifdef HAVE_RAND48
-        if(drand48() > 0.5) {
-#else
-        if (rand() > RAND_MAX / 2) {
-#endif
+        if (xoroshiro128p_next(&d_state[0]) > (UINT64_MAX / 2)){
```
marcusmueller: 1. why `&d_state[0]`? `type[size]` gets automatically converted to `*type` on calling a C-Calling-Convention function. 
2. don't do a comparison with "half-scale" to get a bit of output. Take an arbitrary but the lowest and highest bit and use that as bit.

gr-blocks/lib/test_tag_variable_rate_ff_impl.cc
```diff
@@ -103,11 +97,7 @@ namespace gr {
       while(i < ninput_items[0]) {
 
         if(!d_update_once) {
-#ifdef HAVE_RAND48
-          if(drand48() > 0.5) {
-#else
-          if (rand() > RAND_MAX / 2) {
-#endif
+          if (xoroshiro128p_next(&d_state[0]) > (UINT64_MAX / 2)){
```
marcusmueller: same as above

gr-blocks/lib/test_tag_variable_rate_ff_impl.h
```diff
@@ -36,6 +36,7 @@ namespace gr {
       double d_accum;
       double d_rrate;
       uint64_t d_old_in, d_new_in, d_last_out;
+      uint64_t d_state[2];
```
marcusmueller: maybe call it `d_rngstate` or so, just to avoid confusion?

gr-blocks/lib/test_tag_variable_rate_ff_impl.cc
```diff
@@ -26,10 +26,12 @@
 
 #include "test_tag_variable_rate_ff_impl.h"
 #include <gnuradio/io_signature.h>
+#include <gnuradio/xoroshiro128p.h>
 #include <string.h>
 #include <iostream>
 #include <iomanip>
 #include <stdexcept>
+#include <cstdint>
```
marcusmueller: oh, shoot, that one is C++11 or later.

Go for stdint.h, or do an `#if` that checks for C++11 and switch between `cstdint` and `stdint.h`.

marcusmueller:
yeah, that's a kind courtesy of log4cpp (which we really should get rid off) requiring c++11, your local machine in question hence using that, and then the test CI NOT enabling that.

Hopefully, b6550d7 fixes that.

marcusmueller:
great. 

It doesn't fix it. So, for some reason, `__cplusplus>=201103L` is true, but `cstdint` doesn't work.

Greaaaaaat.

So, hardfixing that to `stdint.h`. Hate.

marcusmueller:
merged (486cd82)

PR #1695, opened 2018-03-16
Fix typos in PFB docs.

argilo:
This fixes some typos I noticed while reading over the Polyphase Filterbank documentation.

### DIFF
--- a/gr-filter/include/gnuradio/filter/pfb_arb_resampler_fff.h
+++ b/gr-filter/include/gnuradio/filter/pfb_arb_resampler_fff.h
@@ -86,7 +86,7 @@ namespace gr {
      *      attenuation_dB=ATT, window=filter.firdes.WIN_BLACKMAN_hARRIS)</EM></B>
      *
      * The theory behind this block can be found in Chapter 7.5 of the
-     * following book.
+     * following book:
      *
      *   <B><EM>f. harris, "Multirate Signal Processing for Communication
      *      Systems", Upper Saddle River, NJ: Prentice Hall, Inc. 2004.</EM></B>
--- a/gr-filter/include/gnuradio/filter/pfb_channelizer_ccf.h
+++ b/gr-filter/include/gnuradio/filter/pfb_channelizer_ccf.h
@@ -42,8 +42,8 @@ namespace gr {
      * <EM>fs</EM> divided by the number of channels, <EM>M</EM>.
      *
      * The PFB channelizer code takes the taps generated above and builds
-     * a set of filters. The set contains <EM>M</EM> number of filters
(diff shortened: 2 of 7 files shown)


gr-filter/include/gnuradio/filter/pfb_arb_resampler_fff.h
```diff
@@ -1,6 +1,6 @@
 /* -*- c++ -*- */
 /*
- * Copyright 2009-2012 Free Software Foundation, Inc.
+ * Copyright 2009-2018 Free Software Foundation, Inc.
```
dkozel: This would be 2009-2012, 2018.
However the changes to this file are so small I don't think we need to update the copyright.

gr-filter/include/gnuradio/filter/pfb_channelizer_ccf.h
```diff
@@ -77,12 +77,12 @@ namespace gr {
      * bandwidth of <EM>TB</EM>. We can also specify the out-of-band
      * attenuation to use, <EM>ATT</EM>, and the filter window
      * function (a Blackman-harris window in this case). The first input
-     *  is the gain of the filter, which we specify here as unity.
+     * is the gain of the filter, which we specify here as unity.
      *
-     *      <B><EM>self._taps = filter.firdes.low_pass_2(1, fs, BW, TB,
-     *           attenuation_dB=ATT, window=filter.firdes.WIN_BLACKMAN_hARRIS)</EM></B>
+     *    <B><EM>self._taps = filter.firdes.low_pass_2(1, fs, BW, TB,
+     *       attenuation_dB=ATT, window=filter.firdes.WIN_BLACKMAN_hARRIS)</EM></B>
```
dkozel: This is not relevant to your changes, but it's odd that the 'h' is not capitalized.

argilo:
@dkozel I removed the copyright changes.

gr-filter/include/gnuradio/filter/pfb_channelizer_ccf.h
argilo: I presume it's that way because harris prefers his name in lower case. :smile: https://en.wikipedia.org/wiki/Fredric_J._Harris

I see an alias has been available for some time: https://github.com/gnuradio/gnuradio/commit/f0c9a22db5442ff6029a0561bb2178de851c6b77

dkozel: That's beautiful. I knew in the back of my mind about his preference, but would not have thought to have inverted it for the comment.

marcusmueller:
Thanks all for the great work here :+1:

Issue #1691, opened 2018-03-15
new fastnoise implementation fails unpredictably in tests


noc0lour:
Building af84fcdca7f4f2123fd0da0a2fb5747e726ff0ea and running test on fastnoise show unpredictable failures in `qa_fastnoise`

```
╭─ /tmp/build-gr  
╰─ ctest --output-on-failure -R fastnoise                       ✔  02:19:06 
Test project /tmp/build-gr
    Start 126: qa_fastnoise
1/1 Test #126: qa_fastnoise .....................   Passed    2.45 sec

100% tests passed, 0 tests failed out of 1

Total Test time (real) =   2.45 sec
╭─ /tmp/build-gr  
╰─ ctest --output-on-failure -R fastnoise                       ✔  02:19:42 
Test project /tmp/build-gr
    Start 126: qa_fastnoise
1/1 Test #126: qa_fastnoise .....................***Failed    2.49 sec
F
======================================================================
FAIL: test_001_test_moments (__main__.test_fastnoise_source)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/andrej/git/gnuradio/gnuradio/gr-analog/python/analog/qa_fastnoise.py", line 85, in test_001_test_moments
    self.assertAlmostEqual(data.var(), 2, places=2)
AssertionError: 1.9934497894857957 != 2 within 2 places

----------------------------------------------------------------------
Ran 1 test in 1.056s

FAILED (failures=1)


0% tests passed, 1 tests failed out of 1

Total Test time (real) =   2.50 sec

The following tests FAILED:
	126 - qa_fastnoise (Failed)
Errors while running CTest
╭─ /tmp/build-gr  
```

noc0lour:
I feel checking on correct places is not what we want ....
`2.007` -> correct 
`1.993` -> wrong 

δ is `.007` for both values....

noc0lour:
Also I feel this is not the only place where this thing is going wrong. 
What you want to test for is probably the number places for the difference (absolute) to the desired value compared to `0.f`.

marcusmueller:
closes by manually merging #1696 .

PR #1696, opened 2018-03-16
qa_fastnoise: split testcases and test for reproducibility

noc0lour:
The unittests were quite hard to comprehend. Splitting in subcases based on 
the PDF to test failures are easier to interpret. (If something changes)

Also adding a reproducibility testcase which compares the vector of samples of two runs
with the same seed. 
Which must result in the same samples if generated on the same machine (due to different FPUs).

The current implementation sometimes yields slightly different results even if run on the same machine.

### DIFF
--- a/gnuradio-runtime/lib/math/random.cc
+++ b/gnuradio-runtime/lib/math/random.cc
@@ -73,9 +73,12 @@ namespace gr {
   void
   random::reseed(unsigned int seed)
   {
-    if(seed==0) d_seed = static_cast<unsigned int>(std::time(0));
-    else d_seed = seed;
-    d_rng->seed(d_seed);
+    d_seed = seed;
+    if (d_seed == 0){
+      d_rng->seed();
+    } else {
+      d_rng->seed(d_seed);
+    }
     // reinstantiate generators. Otherwise reseed doesn't take effect.
     delete d_generator;
     d_generator = new boost::variate_generator<boost::mt19937&, boost::uniform_real<float> > (*d_rng,*d_uniform); // create number generator in [0,1) from boost.random
--- a/gnuradio-runtime/lib/math/random.cc
+++ b/gnuradio-runtime/lib/math/random.cc
@@ -141,9 +141,12 @@ namespace gr {
   float
   random::laplacian()
   {
-    float z = ran1()-0.5;
-    if(z>0) return -logf(1-2*z);
-    else return logf(1+2*z);
+    float z = ran1();
+    if (z > 0.5){
+      return -logf(2*(1-z));
+    }else{
+      return logf(2*z);
+    }
   }
 
   /*
--- a/gnuradio-runtime/lib/math/random.cc
+++ b/gnuradio-runtime/lib/math/random.cc
@@ -141,14 +141,12 @@ namespace gr {
   float
   random::laplacian()
   {
-    float z = ran1();
-    if (z > 0.5){
-      return -logf(2*(1-z));
-    }else{
-      return logf(2*z);
-    }
+          float z = ran1();
+          if (z > 0.5f){
+                  return -logf(2.0f * (1.0f - z) );
+          }
+          return logf(2 * z);
   }
-
   /*
    * Copied from The KC7WW / OH2BNS Channel Simulator
    * FIXME Need to check how good this is at some point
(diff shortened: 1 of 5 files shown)


noc0lour:
And it's failing/passing unpredictable in the CI system as I've seen it on my machines.

noc0lour:
Heh, removing `std::time(0)` as default seed seems to solve a couple of issues of this pseudo-random-number-generator.

gnuradio-runtime/lib/math/random.cc
```diff
@@ -141,9 +141,12 @@ namespace gr {
...
   random::laplacian()
   {
-    float z = ran1()-0.5;
-    if(z>0) return -logf(1-2*z);
-    else return logf(1+2*z);
+    float z = ran1();
+    if (z > 0.5){
+      return -logf(2*(1-z));
+    }else{
+      return logf(2*z);
```
noc0lour: I know I can move this out of else...

marcusmueller: spacing around `else`, please :) I know we're about to do automated formatting in the near future, but let's not be savages until then. Or, really, omit the `else` alltogether.

marcusmueller: Also, this is an algorithmic change, and I can't comment on unchanged lines but: please update the copyright year.

gr-analog/include/gnuradio/analog/fastnoise_source_X.h.t
```diff
@@ -29,6 +29,8 @@
 #include <gnuradio/analog/noise_type.h>
 #include <gnuradio/sync_block.h>
 
+#include <vector>
```
marcusmueller: YESSS. ICWY!

gr-analog/lib/fastnoise_source_X_impl.h.t
```diff
@@ -50,6 +50,7 @@ namespace gr {
       void set_type(noise_type_t type);
       void set_amplitude(float ampl);
       void generate();
+      const std::vector<@TYPE@>& samples() const { return d_samples; };
```
marcusmueller: Can we honestly *not* do that code in the header? I know it's less code here, but I really don't see any upside to this.

noc0lour: In public headers definitely, in this case I can move everything out into implementation in a separate commit.

marcusmueller: well, are you planning to do that?

marcusmueller: also, I'm not sure if one-line functions on the same line are OK per our style guidelines, but if they are, we might want to change that.

noc0lour: Can you comment the one-line functions in the `.clang-format` PR if it is applicable? I'm not opposed to changing this and move the implementation. But this won't happen tonight.

marcusmueller:
This should fix #1691, right, @noc0lour?

noc0lour:
It should, since real randomness is removed and the output now only depends on the supplied seed.

gr-analog/lib/fastnoise_source_X_impl.h.t
mbr0wn: Yeah... and I'm not a fan of one-line functions either :grin:

gr-analog/python/analog/qa_fastnoise.py
```diff
@@ -23,67 +23,117 @@
...
+        # It's pseudoramdo thus must be equal
+        self.assertTrue(numpy.array_equal(data1, data2))
+
+    def test_002_real_gaussian_reproducibility(self):
+        data1 = self.run_test_real(analog.GR_GAUSSIAN)
+        data2 = self.run_test_real(analog.GR_GAUSSIAN)
+
+        self.assertTrue(numpy.array_equal(data1, data2))
+
+    def test_003_real_uniform_pool(self):
```
mbr0wn: Did you autogen this section? The formatting looks different from before.

mbr0wn:
This is a great change overall. @marcusmueller has all the comments. If he approves, so do I!

gr-analog/python/analog/qa_fastnoise.py
noc0lour: No autogen on my part, don't know about the formatting, Did I format the full file by accident?

marcusmueller: at this point, since the original maint test of pre-my-kind-of-fix is nowhere to be found in this test anymore, I'd be OK with that.

marcusmueller:
Merged (b2724ae9c084906fe105c7eb248ba37d7b58153f)

Issue #1688, opened 2018-03-14
gr-uhd: command tags are broken


mrjacobagilbert:
Part of the effort to reduce runtime creation of interned strings has caused UHD CMD_xxx PMTs to be uninitialized causing runtime segmentation fault when used by the command handler. It looks like they are declared in the header, initialized in the implementation class file but for whatever reason the don't seem to be when used.

marcusmueller:
Closed by #1697, right, @mrjacobagilbert?

mrjacobagilbert:
@marcusmueller sorry! close this one out it's good.

PR #1697, opened 2018-03-16
fix broken gr-uhd command tags

catemarie:
These constant pmt variables are not initialized when included from the header file.  I figure they can be initialized in the header file itself.  This corresponds to the bug in issue #1688.


mrjacobagilbert:
Thanks for getting this PR in. For what it's worth, this is the same way we have addressed the issue.

marcusmueller:
Hey @catemarie, thanks for contributing this nice (both as in "accurate, precise" and "good to read") fix!

I'd love to see you do more of that! Which also means that you should never hesitate to reach out to us in any way if you need help with something that needs fixing. Also: If you want, always feel free to head over to the open pull requests and pick one to review.

PR #1490, opened 2017-10-21
[gr-fec] include GSL_LDFLAGS in CMakeLists to avoid linking errors in dependent in-trees

sbmueller:
These changes are required in order to build GR on macOS 10.13 with MacPorts. Also tested on Fedora 25, but further testing e.g. on Ubuntu would be sensible. If these changes are not provided, I get a linker error: `ld: library not found for -lgsl`. @bastibl hope that doesn't interfere with #1247, where you actually deleted this.

PR #1698, opened 2018-03-16
[gr-fec] Define GSL_LDFLAGS in MSVC build

gnieboer:
Pull request #1490 breaks the windows build because GSL_LDFLAGS isn't defined in MSVC windows.  

Assuming the use of GSL_LDFLAGS is needed in Linux instead of the more common XXX_LIBRARIES pattern, this will specifically define GSL_LDFLAGS for WIN32/MSVC builds only.

That allows gr-fec/lib/CMakeText.txt to find the correct libraries and add the dependencies.


marcusmueller:
What can I say: Thank you, guys, and really, keep up the great work!

PR #1699, opened 2018-03-17
grc: Fixed pasting blocks from clipboard

haakov:
Fixed a bug where blocks didn't show up when pasting them. I believe it works like in the `master` branch now. 

I have removed the if statement regarding epy_blocks. Looked like it was redundant, but please correct me if I'm wrong! :)

### DIFF
--- a/grc/gui/canvas/flowgraph.py
+++ b/grc/gui/canvas/flowgraph.py
@@ -243,45 +243,34 @@ class FlowGraph(CoreFlowgraph, Drawable):
         v_adj = scroll_pane.get_vadjustment()
         x_off = h_adj.get_value() - x_min + h_adj.get_page_size() / 4
         y_off = v_adj.get_value() - y_min + v_adj.get_page_size() / 4
+
         if len(self.get_elements()) <= 1:
             x_off, y_off = 0, 0
-        #create blocks
+
+        # create blocks
         for block_n in blocks_n:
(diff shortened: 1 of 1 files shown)


skoslowski:
The paste-function and block.import() share a lot of code/functionality.
I never got around to fixing this --> great that someone finally did!

The epy-block part was surely not redundant - its parameters only exist after the block's source code was loaded and parsed (the rewrite call). Using the standard import() call is better (no more block-type specific code) and *should* work the same. Needs testing though...

haakov:
Right! I tried duplicating an epy_block by copying and pasting it. Looks like it worked, at least the code was preserved.

How about the `_io_cache` parameter? I couldn't find it in an epy_block on this branch.

skoslowski:
Yes, its a state now - like enabled/disabled, position, rotation, ...
https://github.com/gnuradio/gnuradio/blob/python3/grc/core/blocks/embedded_python.py#L96

haakov:
Oh okay :+1: 

Yep, looks like the parameters and their values are kept :)

marcusmueller:
Hey, guys, as you notice I've started merging PR's into python3 again. This one especially was merged very much on a "looks proper, and @haakov as well as @skoslowski probably know better than I what they're doing", so: keep up the great work!  We're moving steady again towards a merge of master/next, after which the master/python3 merge will happen :)

PR #1700, opened 2018-03-19
fft: fix FFTW wisdom file locking for fft_real_fwd and fft_real_rev

balint256:
FFTW wisdom file locking was only taking place from gr::fft::fft_complex.
Generalise lock init and also enable it from 'fft_real_fwd' and 'fft_real_rev'.


balint256:
Sweet as, thanks @noc0lour.

marcusmueller:
yeah, this is one of the many multithreading problems that GR runs into – thanks, @balint256, for being universally Awesome (with a capital A).

PR #1702, opened 2018-03-21
zeromq: adds test delay to allow sub to init prior to data send

gnieboer:
This should only affect testing.

In this test, immediately after tb.start() is called asynchronously, the publisher sends the data stream.  If the flowgraph has not yet had time to initialize the subscriber, no data will be received.  So this adds a .25 second delay to ensure that works.

While this caused test fails on Win64, I assume this issue would cause test fails on Linux builds as well.

### DIFF
--- a/gr-zeromq/python/zeromq/qa_zeromq_sub.py
+++ b/gr-zeromq/python/zeromq/qa_zeromq_sub.py
@@ -50,7 +50,7 @@ class qa_zeromq_sub (gr_unittest.TestCase):
         self.tb.connect(zeromq_sub_source, sink)
 
         self.tb.start()
-        time.sleep(0.25)
+        time.sleep(0.05)
         self.pub_socket.send(src_data.tostring())
         time.sleep(0.25)
         self.tb.stop()
@@ -68,7 +68,7 @@ class qa_zeromq_sub (gr_unittest.TestCase):
         self.tb.connect(zeromq_sub_source, sink)
 
         self.tb.start()
-        time.sleep(0.25)
+        time.sleep(0.05)
         self.pub_socket.send_multipart(src_data)
         time.sleep(0.25)
         self.tb.stop()
--- a/gr-zeromq/python/zeromq/qa_zeromq_sub.py
+++ b/gr-zeromq/python/zeromq/qa_zeromq_sub.py
@@ -50,6 +50,7 @@ class qa_zeromq_sub (gr_unittest.TestCase):
         self.tb.connect(zeromq_sub_source, sink)
 
         self.tb.start()
+        time.sleep(0.25)
         self.pub_socket.send(src_data.tostring())
         time.sleep(0.25)
         self.tb.stop()
@@ -67,6 +68,7 @@ class qa_zeromq_sub (gr_unittest.TestCase):
         self.tb.connect(zeromq_sub_source, sink)
 
         self.tb.start()
+        time.sleep(0.25)
         self.pub_socket.send_multipart(src_data)
         time.sleep(0.25)
         self.tb.stop()


noc0lour:
Indeed these tests sporadically fail on linux as well, though `start()` blocks until every block is spawned and constructed using a barrier. Thus due to the usage of the barrier we shouldn't see failing tests (or at least a fix that looks like yours).

gnieboer:
Odd.  I see the fail consistently.  Perhaps there's a system call that's internally async in windows only that is allowing start() to return early?  A winsock issue perhaps?

noc0lour:
This https://github.com/gnuradio/gnuradio/blob/master/gnuradio-runtime/lib/scheduler_tpb.cc#L96 should prevent start from returning before every block spawned.

I'm checking if the zeromq block might return early (or the network stack just needs a brief moment to propagate the change)

noc0lour:
Reading this: https://msdn.microsoft.com/en-us/library/windows/desktop/ms740476(v=vs.85).aspx

Includes a comment: ```
Note  When issuing a blocking Winsock call such as setsockopt, Winsock may need to wait for a network event before the call can complete. Winsock performs an alertable wait in this situation, which can be interrupted by an asynchronous procedure call (APC) scheduled on the same thread. Issuing another blocking Winsock call inside an APC that interrupted an ongoing blocking Winsock call on the same thread will lead to undefined behavior, and must never be attempted by Winsock clients. ```

marcusmueller:
So, two conflicting feelings in my chest:

1. Make this work reliably on Windows. So, add that delay
2. don't add a delay to mask a problem in testing that a user might see in usage.

@gnieboer am I right in assuming that there will be cases where this also fails outside testing, and a user would have to add the delay themselves?

gnieboer:
That's a good question.  Honestly, I don't think this will be a problem in real life, as the delay is for a transmission that is occurring outside a flowgraph.  I can't image a scenario where a user needs to start a flowgraph perfectly in sync with an outside source.  Normal use would be to fire up the flowgraph and wait to receive, right?

BTW, I also just did some quick testing, and sleep(.25) is overkill.  Looks like the thread just needs to yield very briefly ... sleep(.001) did the trick as well.

marcusmueller:
darn, I was hoping this issue was clearer cut and less easy to be solved by 1 ms of waiting. 

I have my doubts regarding @noc0lour's WINSOCK citation since we sporadically see it on linux, too. So, maybe a zeromq pubsub pattern issue? I mean, you need to be subscribed *before* you can get messages, right?

noc0lour:
@marcusmueller In this unit test only the sub source is tested. The zeromq pub endpoint is created before anything else happens. 

The sub source is setting the right context in its constructor. The only thing that might mess up is the time zeromq actually needs to subscribe to the created socket.  So we probably want the delay in the unit test.

marcusmueller:
I'd argue that, no, we don't **want** to do that, but we shall, in the end. We just want to SUB first, then PUB, <br>http://zguide.zeromq.org/page:all#Missing-Message-Problem-Solver

> shamelessly hotlinked from the zmq guide 
> …
>
> * Even if you synchronize a SUB and PUB socket, you may still lose messages. It's due to the fact that internal queues aren't created until a connection is actually created. If you can switch the bind/connect direction so the SUB socket binds, and the PUB socket connects, you may find it works more as you'd expect.

Just to make matters *worse*: http://zguide.zeromq.org/page:all#Getting-the-Message-Out
(emphasis *not* even by me)
> There is one more important thing to know about PUB-SUB sockets: you do not know precisely when a subscriber starts to get messages. Even if you start a subscriber, wait a while, and then start the publisher, **the subscriber will always miss the first messages that the publisher sends**. This is because as the subscriber connects to the publisher (something that takes a small but non-zero time), the publisher may already be sending messages out.
>…
> There is a simple and stupid way to delay the publisher, which is to sleep. Don't do this in a real application, though, because it is extremely fragile as well as inelegant and slow

So, to avoid the TCP connection time farkup, can't we not fix a TCP-specific issue by adding delay states, but by not using TCP? Inproc should work, connect instantly, and to top it all off, is process-local, so that parallel tests with the same socket ID are trivial, as long as they run in separate processes.

Short answer is: No, we can't, because gr-zeromq won't let us.

So, yeah, add that delay here; make it as short as feasible, localhost TCP connections really shouldn't take quarter of a second to establish, so, go down to 50 ms and let's be done with this :)

Long answer:

gr-zeromq creates a new ZMQ context for every socket. That basically is a no-no for zmq to start with, but I'll do a separate issue to tackle that – there's more about the context creation that doesn't feel well for me. So, let's do this delay here, and later change the way we're doing zeromq, and then come back to this test and make it work without the delay. 

TL;DR:

Yeah, go for it.

gr-zeromq/python/zeromq/qa_zeromq_sub.py
```diff
@@ -50,6 +50,7 @@ def test_001 (self):
         self.tb.connect(zeromq_sub_source, sink)
 
         self.tb.start()
+        time.sleep(0.25)
```
marcusmueller: reduce to `0.05`, please.

gr-zeromq/python/zeromq/qa_zeromq_sub.py
```diff
@@ -67,6 +68,7 @@ def test_002 (self):
...
```
marcusmueller: same.

gr-zeromq/python/zeromq/qa_zeromq_sub.py
```diff
@@ -67,6 +68,7 @@ def test_002 (self):
...
         self.pub_socket.send_multipart(src_data)
         time.sleep(0.25)
```
marcusmueller: same.

marcusmueller:
can immediately be merged after delays have been reduced.

marcusmueller:
We can't let the sub socket bind, that's not architecturally possible in current gr-zeromq

gnieboer:
Reduced sleep to 50ms.

marcusmueller:
merged (78a42dd18b0a88925fc84e766a7207877a629c8f)

PR #1703, opened 2018-03-21
Test: adds swig dirs to python tests to allow ctest to run on win builds

gnieboer:
Running unit tests on windows has previously required running tests individually because the dependencies are not set up correctly for the build structure.

This adds the appropriate directories and allows ctest to run within the run_gr.bat environment.

Only affects test environment, and while Linux may not require the additional paths, I don't think they'll be hurt by them either.


marcusmueller:
I concur. And thus, I merge. And thus, I thank you once again!

PR #1713, opened 2018-03-30
Fix the potential for Static Initialization Order Fiasco

### LINKED ISSUE #1711: Move const initializations of PMTs into headers
as #1697 showed, having these in the .cc files can seemingly, under conditions I don't fully grok, lead to hazard. Anyway, having them static const in the header is probably what one wants, anyway. So let's do that.

marcusmueller:
Found thanks to #1697.

Closes #1711.

### DIFF
--- a/gr-uhd/include/gnuradio/uhd/usrp_block.h
+++ b/gr-uhd/include/gnuradio/uhd/usrp_block.h
@@ -30,6 +30,24 @@
 namespace gr {
   namespace uhd {
 
+    GR_UHD_API const pmt::pmt_t cmd_chan_key();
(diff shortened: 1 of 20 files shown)


marcusmueller:
I'll remove this comment once complete: please don't review before.

gr-uhd/include/gnuradio/uhd/usrp_block.h
```diff
@@ -30,6 +30,24 @@
 namespace gr {
   namespace uhd {
 
+    const pmt::pmt_t cmd_chan_key();
```
noc0lour: Exposing those keys with `GR_UHD_API` fixes the missing symbols in the lib.

PR #1718, opened 2018-04-04
Update volk submodule instructions to work correctly

pwicks86:
The instructions that cmake gives to check out the VOLK submodule don't actually check it out.


marcusmueller:
Every bit of helpful documentation counts! Thank you!

PR #1721, opened 2018-04-05
Made OFDM Cyclic Prefixer ready for 3GPP LTE.

m3x1m0m:
3GPP LTE defines two different lengths for the Cyclic Prefix (CP). For instance, when sampling with 15.36 MHz, the first CP has a length of 40, then 6 following CPs have a length of 36 and this pattern keeps repeating (with an interval of 7). The proposed patch makes it possible to use the OFDM cyclic prefixer together with 3GPP protocols, which are based on this mechanism.

### DIFF
--- a/gr-digital/examples/ofdm/tx_ofdm.grc
+++ b/gr-digital/examples/ofdm/tx_ofdm.grc
@@ -1,5 +1,5 @@
 <?xml version='1.0' encoding='utf-8'?>
-<?grc format='1' created='3.7.10'?>
+<?grc format='1' created='3.7.12'?>
 <flow_graph>
   <timestamp>Wed Jul  9 15:49:47 2014</timestamp>
   <block>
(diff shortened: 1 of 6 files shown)


willcode:
This adds a specific policy to a generic block. If another protocol came along that had 3 different CPs, or if they were sequenced differently, then the block would need to change again.

Instead, you could provide the ability to define a list of CPs (by whatever parameters are appropriate - is it just length?) and a sequence of CPs to use. If it's just length, then it could be defined by a single list, e.g., (3,2,2).

m3x1m0m:
@willcode Sounds reasonable to me. I will look into it.

gr-blocks/lib/message_strobe_impl.cc
```diff
@@ -91,7 +91,7 @@ namespace gr {
     void message_strobe_impl::run()
     {
       while(!d_finished) {
-        boost::this_thread::sleep(boost::posix_time::milliseconds(d_period_ms));
+        boost::this_thread::sleep(boost::posix_time::milliseconds(static_cast<long>(d_period_ms)));
```
michaelld: This is already in master too: https://github.com/gnuradio/gnuradio/commit/42ea667f9ae27e377a010df7050b203276412d2b .

PR #1722, opened 2018-04-05
cmake: use MATCHES for "Clang", not STREQUAL

michaelld:
Clang can be either "Clang" or "AppleClang", so use regex match instead of exact string comparison.

Simple change, but important for folks using recent Mac OS X / macOS.


marcusmueller:
True! Great change.

marcusmueller:
cherry-picked onto maint-3.7

PR #1723, opened 2018-04-05
Grid pos fix

skoslowski:
Eval and collision detection for `grid_pos` parameters (WX GUI) is currently broken.
See https://github.com/pothosware/PothosSDR/issues/47

My refactoring commit https://github.com/gnuradio/gnuradio/commit/760c147523f277f9dd5e1fe6c4c4a1c8447ecc31 on top of https://github.com/gnuradio/gnuradio/pull/1418 caused this. Backporting from next I totally forgot about WX GUI...

As always, I recommend someone double-check this...

### DIFF
--- a/grc/core/Param.py
+++ b/grc/core/Param.py
@@ -541,6 +541,7 @@ class Param(Element):
         # Grid Position Type
         #########################
         elif t == 'grid_pos':
+            self.hostage_cells.clear()
             if not v:
                 # Allow for empty grid pos
                 return ''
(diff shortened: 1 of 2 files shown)


marcusmueller:
I'm currently not in a position to run `tig` to figure this out: does that fix a bug that is present in 3.7.12.0? We don't need to fix WX on master. Master leads to 3.8 (and will be merged with next very soon).

marcusmueller:
After discussing these things with @skoslowski yesterday, and pondering them in his heart, this maintainer has decided that this PR really has no use being on master, and will rebase onto maint-3.7.

This is not a breach in "all commits that apply to master should go onto master, and if they're bugfixes, be cherrypicked/backported onto maint-3.X" policy. This commit only _mechanically_ applies to master.

We don't do WX on master, so it doesn't apply contextually. Also, by not merging onto master, I just avoid yet another conflict when merging next into master (where WX is already gone).

To put this more than clearly: The fact that I'm having this inner and outer dialogue is actually kind of a maintainer fail on my side, I wish I had already done the next->master merge; then there would be no lack in clarity :) So, once more, thank you, @skoslowski, for the awesome work, and might many people buy you many hot and cold drinks, or throw money in your general direction, for writing and factually maintaining the most popular SDR design GUI on planet earth.

marcusmueller:
4f5cda08e971a0ddaf1daa5b2bce16fd596f4bc2 will be cherry-picked onto master

PR #1724, opened 2018-04-05
grc: Fixed the "Show Msg Ports" option (python3)

haakov:
6 of the blocks in the block tree have the ability to hide their message ports. Without this fix, the message ports are shown regardless of the "Show Msg Ports" option.


marcusmueller:
hey,  @noc0lour can you give this another QA run?

marcusmueller:
This still has to be merged into `next`, right? Or did we somehow end up with this already in next?

haakov:
This is not yet a part of next, these fixes are for XML files. I can replace this commit with the fixes for YAML or I could just add it to the XML -> YAML pull request. What do you prefer?

marcusmueller:
If possible, keep separate :) 

So, XML->yaml: I thought that happened on next now?

haakov:
I've created a new PR against `next` with this fix (for YAML). Seemed to be the tidiest thing to do. See: https://github.com/gnuradio/gnuradio/pull/1849

PR #1725, opened 2018-04-06
Restore ability to use absolute thresholding with corr_est

### LINKED ISSUE #1207: Correlation Estimator regression
After trying to use the Correlation Estimator without success, I looked at the examples provided to understand how to make it work.

Unfortunately, the [example flowgraph](https://github.com/gnuradio/gnuradio/blob/master/gr-digital/examples/demod/test_corr_est.grc) seems to be out of sync with the last changes made to the [Correlation Estimator block](https://github.com/gnuradio/gnuradio/blob/master/gr-digital/lib/corr_est_cc_impl.cc), and is completely unusable because almost every point of the output gets tagged. 

Increasing the threshold makes it better, but there is still some correlations where the |corr|^2 curve clearly shows zero, which looks like a bug.

I attached the output of |corr|^2 with a threshold of 0.9999999 :

image

btashton:
This extends the corr_est block to take a threshold method as an argument.

* THRESHOLD_ABSOLUTE restores the previous function of this block.
* THRESHOLD_DYNAMIC provides the outlier detection algorithm that was later added.

This should resolve #1207

### DIFF
--- a/gr-digital/grc/digital_corr_est_cc.xml
+++ b/gr-digital/grc/digital_corr_est_cc.xml
@@ -32,6 +32,20 @@
     <type>float</type>
   </param>
 
+  <param>
+    <name>Threshold Method</name>
+    <key>threshold_method</key>
+    <type>enum</type>
+    <option>
+      <name>Absolute</name>
+      <key>digital.corr_est_cc.THRESHOLD_ABSOLUTE</key>
+    </option>
+    <option>
+      <name>Dynamic</name>
+      <key>digital.corr_est_cc.THRESHOLD_DYNAMIC</key>
+    </option>
+  </param>
+
   <sink>
     <name>in</name>
     <type>complex</type>
--- a/gr-digital/include/gnuradio/digital/corr_est_cc.h
+++ b/gr-digital/include/gnuradio/digital/corr_est_cc.h
@@ -87,20 +87,28 @@ namespace gr {
     public:
       typedef boost::shared_ptr<corr_est_cc> sptr;
 
+      enum tm_type {
+        THRESHOLD_DYNAMIC,
+        THRESHOLD_ABSOLUTE,
+      };
+
       /*!
        * Make a block that correlates against the \p symbols vector
        * and outputs a phase and symbol timing estimate.
        *
-       * \param symbols    Set of symbols to correlate against (e.g., a
-       *                   sync word).
-       * \param sps        Samples per symbol
-       * \param mark_delay tag marking delay in samples after the
-       *                   corr_start tag
-       * \param threshold  Threshold of correlator, relative to a 100%
-       *                   correlation (1.0). Default is 0.9.
+       * \param symbols           Set of symbols to correlate against (e.g., a
+       *                          sync word).
+       * \param sps               Samples per symbol
+       * \param mark_delay        tag marking delay in samples after the
+       *                          corr_start tag
+       * \param threshold         Threshold of correlator, relative to a 100%
+       *                          correlation (1.0). Default is 0.9.
+       * \param threshold_method  Method for computing threshold.
+       *
        */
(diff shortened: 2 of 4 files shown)


marcusmueller:
Hey! I'm jmcorgan's successor as project maintainer, and just wanted to chime in, albeit I'm still on vacation and can't really properly review code in-depth for another 10 days:

This looks good to me, from a coding point of view. As said, haven't had a chance to analyze functionally.

Contributions like this are always highly welcome!

Let me ask this, since this is a changeset significant enough to require copyright assignment:

Have you been in touch with @bhilburn (bhilburn@gburadio.org)? We'll need a tiny bit of paperwork with the Free Software Foundation, so that they can protect your and our rights to your code.

btashton:
Thanks, I just reached out to him.  The part I was most unsure of for the fix was the enum parameter I added in the XML file.

marcusmueller:
@btashton yeah, I saw the paperwork, but I'm as lost as you are on the current state at the FSF. Maybe @bhilburn will enlighten us :)

marcusmueller:
my bad, it's been sorted out for weeks now :(

marcusmueller:
So, regarding the `enum` trouble: you've been spot-on. Generally, it's not a problem to have an `enum` in a class and have SWIG wrap that C++ property to a Python type, but: 

GNU Radio block classes are special in that we hide the actual class behind a reference to its `make()` function of the same name. That means we can't do that. Would you agree if I was to just take that `enum` out of class scope and just put in the same `gr::digital` namespace?

btashton:
@marcusmueller I can update the PR to make that change.  Also I was not really sure what to call the non-fixed mode since there are multiple ways the dynamic thresholding can be done and AFAIK no one seems to know where the current algorithm came from.  Also comments on the `threshold` method? This did not return anything useful  in the dynamic mode, so I changed it to return `d_pfa` but I am not clear if that is actually useful either.

btashton:
@marcusmueller any thoughts on my questions? If you think it is good enough I'll just update it to resolve the namespace issue and hopefully we can get this merged in. This has been verified with real hardware on some flow graphs that were broken.

marcusmueller:
I agree, this is definitely mature enough for merging - please do the namespace update.

Regarding naming: I'll finish this coffee and then decide whether "dynamic" is specific enough.

marcusmueller:
So, I'd recommend you name it dynamic_energy_normalized or so, but if you don't, I'd still merge.

One thing I'm a bit concerned about is that the aggressiveness is defined by how many items a call to work sees, right? Because that defines how much the energy gets averaged (impl.cc:259), or am I missing something? That would introduce signal processing that would depend on how the previous block "chunks up" the samples, and that would not be in the spirit of GNU Radio. It'd be easy to work around that, wouldn't it?

marcusmueller:
@noc0lour something's strange with the builder on this PR… it states it can't find the ref, even after I just re-kicked it.

marcusmueller:
@noc0lour  just re-re-kicked it, works, it seems :)

marcusmueller:
hey @btashton, sorry to have let this slip for so long.

I just revisited your code, and I *really* want to merge all this, but I have my doubts is a good idea that we're doing something different depending on how many input items we get per work() call, but I might simply be lacking the math overview here. Could you elaborate, if you find time, on what's the idea of normalizing by noutput_items?

btashton:
@marcusmueller I totally agree with you, but I have no idea how the dynamic threshold math works out, I just moved the logic to fit alongside the original fixed mode. The commit that changed the logic to be dynamic had no comments on where it came from or several of the magic numbers. I'm traveling for the next two weeks, but I'm happy to make any changes that you think would make the situation better.

marcusmueller:
Agreeing! OK; since this more than a sensible thing to do: merging.

PR #1728, opened 2018-04-11
cmake: fix CTRLPORT includes and testing

michaelld:
Currently, CTRLPORT_BACKENDS is not declared if ENABLE_GR_CTRLPORT is FALSE, so testing this variable makes no sense. All testing in gr-* is done using ENABLE_GR_CTRLPORT, so move to that in gr-blocks.

That said, having CTRLPORT_BACKENDS set to 0 could be useful, so rearrange CTRLPORT determination to set this variable correctly.

If ENABLE_CTRLPORT_THRIFT, then correctly use include_directories for header directories, -after- CTRLPORT is determined, not before. We can also combine some of these declarations to simplify the code.

### DIFF
--- a/gnuradio-runtime/lib/CMakeLists.txt
+++ b/gnuradio-runtime/lib/CMakeLists.txt
@@ -44,7 +44,7 @@ configure_file(
 list(APPEND gnuradio_runtime_sources ${CMAKE_CURRENT_BINARY_DIR}/constants.cc)
 
 ########################################################################
-# Setup the include and linker paths
+# Setup globally used include paths
 ########################################################################
 include_directories(${GNURADIO_RUNTIME_INCLUDE_DIRS}
                     ${CMAKE_CURRENT_SOURCE_DIR}
@@ -52,12 +52,6 @@ include_directories(${GNURADIO_RUNTIME_INCLUDE_DIRS}
                     ${Boost_INCLUDE_DIRS}
                     ${MPLIB_INCLUDE_DIRS}
 )
-
-if(ENABLE_CTRLPORT_THRIFT)
-    list(APPEND include_directories
-        ${THRIFT_INCLUDE_DIRS}
-    )
-endif(ENABLE_CTRLPORT_THRIFT)
 
 ########################################################################
 # Include subdirs rather to populate to the sources lists.
@@ -67,6 +61,13 @@ GR_INCLUDE_SUBDIRECTORY(messages)
 GR_INCLUDE_SUBDIRECTORY(thread)
 GR_INCLUDE_SUBDIRECTORY(math)
 GR_INCLUDE_SUBDIRECTORY(controlport)
+
+########################################################################
+# Setup CTRLPORT include path, if using
+########################################################################
+if(ENABLE_CTRLPORT_THRIFT)
+    include_directories(${THRIFT_INCLUDE_DIRS})
+endif(ENABLE_CTRLPORT_THRIFT)
 
 ########################################################################
 # Setup library
--- a/gnuradio-runtime/lib/controlport/CMakeLists.txt
+++ b/gnuradio-runtime/lib/controlport/CMakeLists.txt
@@ -17,15 +17,18 @@
 # the Free Software Foundation, Inc., 51 Franklin Street,
 # Boston, MA 02110-1301, USA.
 
-if(ENABLE_GR_CTRLPORT)
-
 # Keep track of the number of backends ControlPort supports
 SET(CTRLPORT_BACKENDS 0)
+
+if(ENABLE_GR_CTRLPORT)
 
 # Add definition so we can compile in ControlPort to the blocks.
 add_definitions(-DGR_CTRLPORT)
 
-include_directories(${CMAKE_CURRENT_SOURCE_DIR})
+include_directories(
+  ${CMAKE_CURRENT_SOURCE_DIR}
+  ${CMAKE_CURRENT_BINARY_DIR}
+)
 
 list(APPEND gnuradio_ctrlport_sources
   ${CMAKE_CURRENT_SOURCE_DIR}/rpcmanager.cc
@@ -33,7 +36,6 @@ list(APPEND gnuradio_ctrlport_sources
   ${CMAKE_CURRENT_SOURCE_DIR}/rpcserver_booter_aggregator.cc
   ${CMAKE_CURRENT_SOURCE_DIR}/rpcserver_selector.cc
 )
-
 
 OPTION(ENABLE_CTRLPORT_THRIFT "Enable ControlPort Thrift support" ON)
 
@@ -88,17 +90,10 @@ install(
 
 endif(THRIFT_FOUND)
 endif(ENABLE_CTRLPORT_THRIFT)
-
-########################################################################
-# Add controlport stuff to gnuradio-runtime
-########################################################################
-
-include_directories(${CMAKE_CURRENT_BINARY_DIR})
+endif(ENABLE_GR_CTRLPORT)
 
 # Save the number of backends for testing against later
 set(
   CTRLPORT_BACKENDS ${CTRLPORT_BACKENDS}
   CACHE INTERNAL "Number of ControlPort backends available"
 )
-
-endif(ENABLE_GR_CTRLPORT)
--- a/gr-blocks/python/blocks/CMakeLists.txt
+++ b/gr-blocks/python/blocks/CMakeLists.txt
@@ -45,13 +45,13 @@ if(ENABLE_TESTING)
   file(GLOB py_qa_test_files "qa_*.py")
 
   # Force out the controlport QA tests if we have no backends to use.
-  if(CTRLPORT_BACKENDS EQUAL 0)
+  if(NOT ENABLE_GR_CTRLPORT)
     list(REMOVE_ITEM py_qa_test_files
       ${CMAKE_CURRENT_SOURCE_DIR}/qa_cpp_py_binding.py
       ${CMAKE_CURRENT_SOURCE_DIR}/qa_cpp_py_binding_set.py
       ${CMAKE_CURRENT_SOURCE_DIR}/qa_ctrlport_probes.py
       )
-  endif(CTRLPORT_BACKENDS EQUAL 0)
+  endif(NOT ENABLE_GR_CTRLPORT)
 
   foreach(py_qa_test_file ${py_qa_test_files})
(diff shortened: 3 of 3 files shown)


marcusmueller:
hey @noc0lour, since this builds on my machines: does re-kicking off the builds help, or is there something strange going on here?

marcusmueller:
@michaelld in essence, this feels like an obvious thing to merge, but as you can see, something goes wrong on the build machines. What's more important to me: I'd guess we want this on 3.7.X, too?

michaelld:
@marcusmueller yes: 3.7.x too. I've no idea why the build machines don't like the change. I've been using it in MacPorts now since I submitted this PR, and it works nicely on all of my test systems (10.5 PPC, 10.6 Intel through 10.3).

gnuradio-runtime/lib/CMakeLists.txt
```diff
@@ -68,6 +63,13 @@ GR_INCLUDE_SUBDIRECTORY(thread)
 GR_INCLUDE_SUBDIRECTORY(math)
 GR_INCLUDE_SUBDIRECTORY(controlport)
 
+########################################################################
+# Setup CTRLPORT include path, if using
+########################################################################
+if(ENABLE_CTRLPORT_THRIFT)
+    include_directories(${THRIFT_INCLUDE_DIRS})
```
noc0lour: Apparently `${THRIFT_INCLUDE_DIRS}` is not defined on any Linux distro. Maybe it's a naming issue?
That's what the build machines are trying to tell you.

michaelld: Interesting. Let me look into the naming of this variable. Do the build machines succeed with the original code (which also uses this variable name)? That would be quite odd  ...

michaelld: So unless GR's CMakeLists.txt scripts are not using the local cmake/Modules/FindThrift.cmake, then this variable will be used. See, e.g., https://github.com/gnuradio/gnuradio/blob/master/cmake/Modules/FindThrift.cmake#L84 . This usage is OS-independent.

noc0lour: What's the reasoning behind changing the `list(APPEND ..)` to a simple `include_directories(..)` ?

michaelld: 2 answers: (1) If you search through the whole of the GR repo for "include_directories", this is the only instance (that I can find) where `include_directories()` as a function is not used; I found that very odd & moved to the "correct" usage. That's because (2) `include_directories` is a function that sets a directory property; it is not a list that is supposed to be directly accessible in the manner used here.

noc0lour: Haven't had a lot of time lately, sorry. In the buildbot containers I have `thrift` installed. 
If you have docker for Linux available you can locally checkout a build on them: https://github.com/gnuradio/gnuradio-buildbot/tree/master/worker

I have a kinda working thrift installation on my machine, I'll checkout configuration there.

noc0lour: Ok, running cmake on your branch works on my machine. So it's something special about the compiler/cmake versions used in the build containers...

michaelld:
Thoughts on this PR?

noc0lour:
Not really, I try to add more differentiated containers to the test environment, this might help nailing it down.

michaelld:
Ping here ... any updates on the CI ... from recent PRs, looks like the CI functionality is a "mixed bag" ...

michaelld:
Also: Do you prefer this PR to be on Next instead of Master? I can redo if desired ...

marcusmueller:
This is fine, just confused why this still happens; I'm trying to rebase this onto current master, maybe that helps :/

marcusmueller:
#1992

marcusmueller:
um, @noc0lour: all builds (not only the ones of this PR) say:

<pre>
-- Checking for module 'thrift'
--   No package 'thrift' found
-- Binary 'thrift' not found.
</pre>

wat?

michaelld:
... hmmm ... I don't think that's supposed to happen. Interesting ... strange ...

noc0lour:
Is `thrift` available/installed in all containers?

noc0lour:
Hm, checking again on the containers there is no thrift installed. Does this PR mean thrift will be a hard requirement for building?

gnuradio-runtime/lib/CMakeLists.txt
noc0lour: So, would it make sense to put it after the ` if(THRIFT_FOUND)` stanza?

michaelld: We're already using `OPTION(ENABLE_CTRLPORT_THRIFT "Enable ControlPort Thrift support" ON)` which will be TRUE if Thrift is found and matches whatever specs are required for it. We don't want to allow any old Thrift, after all :)

noc0lour: Reading: https://cmake.org/cmake/help/v3.0/command/option.html We don't match any thrift version before the find_package(thrift) call, or can you enlighten me what I'm missing here. Anyway the build should pass if thrift is not available on a build machine, right?

noc0lour: the `ENABLE_CTRLPORT_THRIFT` is a pure user defined variable which is set to ON by default (which then enables the search for thrift). instead of ENABLE_CTRLPORT_THRIFT the right check should be like in the other file `if(FOUND_THRIFT)`

noc0lour:
getting `thrift` on the CI machines is a real pain. Debian/Ubuntu do not have ready-made binary packages (with C++ library & header). Did not check on Fedora yet. So this definitely needs to work on machines without thrift anyway.

gnuradio-runtime/lib/CMakeLists.txt
michaelld: The build should pass if Thrift is not found: true! Let me verify locally, but a quick read through the CMake code looks OK from this perspective to me.

michaelld: But OK yeah I get your point about ENABLE_CTRLPORT_THRIFT versus FOUND_THRIFT. Note that there's also ENABLE_GR_CTRLPORT, CTRLPORT_BACKENDS, GR_RPCSERVER_ENABLED, and GR_RPCSERVER_THRIFT ... and maybe more! Holy crap ctrlport's CMake stuff is complicated!

michaelld: And CTRLPORT_BACKENDS too.

michaelld: My recommendation is to leave FOUND_THRIFT alone & not use it except for setting ENABLE_CTRLPORT_THRIFT. Then, change the CMake code in gnuradio-runtime/lib/controlport/CMakeLists.txt such that if FOUND_THRIFT isn't TRUE, then ENABLE_CTRLPORT_THRIFT will be set to false. If there's a good way to figure out if the user selected ENABLE_CTRLPORT_THRIFT to be TRUE, then have CMake error out with a reasonable debug printout -- like what it does with ENABLE_GR_* if it doesn't meet enabling criteria. I think this will take care of the immediate issue you found, @noc0lour .

marcusmueller:
OK, I'm not really happy about missing thrift support in our CI :( but oh well, that means we'll merge this after the next merge.

Yes, @noc0lour, Fedora has trivially installable thrift, python2-thrift packages. Don't know about python3; I don't even know whether python3 bindings exist on thrift's side. Uh-oh. I'd like to have ctrlport under Py3, too for 3.8, but I don't want a panic-induced replacement of thrift (because panic-replacing brought us thrift, and it's been honestly the worst dependency we've picked so far).

michaelld:
Holy cow the CTRLPORT CMake stuff is a mess! I'm cleaning it up a little; enough to make some sense of things. Testing various options with and without Thrift / CTRLPORT enabled. Probably take the rest of today for testing. Will try to push the update to this PR late today or tomorrow.

michaelld:
Actually I'll wait until after @marcusmueller does the next -> master merge, then rebase on the new GIT master HEAD & push whatever remains to this PR. That way it'll be ready to go, which might be the case now, of course, but it's always nice to know a PR can be committed without further ado.

noc0lour:
@michaelld I rebased on current master and pushed on your branch

michaelld:
awesome thanks @noc0lour !

michaelld:
I think this PR is good to go now. I'll evaluate it a little more today, but in local testing it does what it's supposed to do for CTRLPORT cmake behavior. Big question is whether allowing CTRLPORT without a backend works -- since the cmake code allows for it, we should verify the behavior.

michaelld:
Interestingly, the CI fails because "make test" fails for CTRLPORT QA ... which should be disabled by this PR! No idea what the CI thinks should be going on here ...

michaelld:
I've done what I can here to verify this change. I'm ready to rebase & merge. Anyone object? I'll do this Thursday morning (CA time) unless I hear otherwise.

noc0lour:
At least one of the errors is solved by putting that  to gnuradio-runtime/lib/CMakeLists.txt:
```
# Setup CTRLPORT include path, if using
if(ENABLE_CTRLPORT_THRIFT)
  if(THRIFT_FOUND)
    include_directories(${THRIFT_INCLUDE_DIRS})
  endif(THRIFT_FOUND)
endif(ENABLE_CTRLPORT_THRIFT)
```

noc0lour:
The other one is related to `GR_ENABLE_CTRLPORT` being enabled without a single backend. Which apparently fails the python tests.

michaelld:
Yes I see there's an issue for when CTRLPORT is enabled without a backend (Thrift). Let me look into it.

gnuradio-runtime/lib/CMakeLists.txt
michaelld: So the issue here is if Thrift is enabled but not found, then the variable ENABLE_CTRLPORT_THRIFT will be set and TRUE, while THRIFT_INCLUDE_DIRS will not be set and/or undefined. Are @noc0lour noted in discussion, we should be checking for both here. I'll fix this check to do both.

michaelld:
For whatever reason, I was comparing the patch I had for the final 3.7 GIT master commit and this PR's patch, and ... they are so very different! I think my other patch is a lot better than this one, but it needs a few fixes per our work here of recent. So ... I'm going to close this PR, fix up the patch locally, and create a new PR with the new changes. Hopefully starting with a clean slate will be useful.

PR #1730, opened 2018-04-11
modtool_rename: Add list of sensible blocknames

swap-nil7:
Added the script for autocompletion of sensible blocknames for renaming.

### DIFF
--- a/gr-utils/python/modtool/modtool_rename.py
+++ b/gr-utils/python/modtool/modtool_rename.py
@@ -22,8 +22,11 @@
 
 import os
 import re
+import glob
+
 from optparse import OptionGroup
 from modtool_base import ModTool, ModToolException
+from util_functions import SequenceCompleter
 
 
 class ModToolRename(ModTool):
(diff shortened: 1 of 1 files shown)


gr-utils/python/modtool/modtool_rename.py
```diff
@@ -38,6 +41,17 @@ def __init__(self):
         self._skip_cmakefiles = False
         self._license_file = None
 
+    def get_block_candidates(self):
+        cpp_blocks = filter(lambda x: not (x.startswith('qa_') or
+                            x.startswith('test_')),
+                            glob.glob1("lib", "*.cc"))
```
mbr0wn: Maybe add `*.cpp` to the glob. It's rarely used, and not created by modtool, but it's the more common ending for C++ files and someone might be using it.
Or not. I don't feel strongly about this.

swap-nil7: I have updated the same for the inclusion of `cpp` files.

mbr0wn:
Manually rebased. I'm not sure it's working properly, though.

mbr0wn:
(It's not breaking anything, either).

PR #1733, opened 2018-04-17
Boost 1.67 compatibility

ilovezfs:
Fix regression due to https://github.com/boostorg/date_time/commit/f9f2aaf5216c.

`uhd` may be fixing these in some other way. See https://github.com/EttusResearch/uhd/pull/170#issuecomment-381749132.

### DIFF
--- a/gr-blocks/lib/message_strobe_impl.cc
+++ b/gr-blocks/lib/message_strobe_impl.cc
@@ -91,7 +91,7 @@ namespace gr {
     void message_strobe_impl::run()
     {
       while(!d_finished) {
-        boost::this_thread::sleep(boost::posix_time::milliseconds(d_period_ms));
+        boost::this_thread::sleep(boost::posix_time::milliseconds(static_cast<long>(d_period_ms)));
         if(d_finished) {
           return;
         }
--- a/gr-blocks/lib/message_strobe_random_impl.cc
+++ b/gr-blocks/lib/message_strobe_random_impl.cc
@@ -110,7 +110,7 @@ namespace gr {
     void message_strobe_random_impl::run()
     {
       while(!d_finished) {
-        boost::this_thread::sleep(boost::posix_time::milliseconds(std::max(0.0f,next_delay())));
+        boost::this_thread::sleep(boost::posix_time::milliseconds(static_cast<long>(std::max(0.0f,next_delay()))));
         if(d_finished) {
           return;
         }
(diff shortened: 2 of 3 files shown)


gr-blocks/lib/message_strobe_impl.cc
```diff
@@ -91,7 +91,7 @@ namespace gr {
     void message_strobe_impl::run()
     {
       while(!d_finished) {
-        boost::this_thread::sleep(boost::posix_time::milliseconds(d_period_ms));
+        boost::this_thread::sleep(boost::posix_time::milliseconds(static_cast<long>(d_period_ms)));
```
marcusmueller: This highlights a mistake in message_strobe_impl!

d_period_ms is a `float`, but really, only integer milliseconds work.

So, that's a separate bug. I'm for fixing this like you propose here on both master and maint-3.7, and later change the constructor signature on master to accept `long`s instead of float. Much saner.

gr-blocks/lib/message_strobe_random_impl.cc
```diff
@@ -110,7 +110,7 @@ namespace gr {
     void message_strobe_random_impl::run()
     {
       while(!d_finished) {
-        boost::this_thread::sleep(boost::posix_time::milliseconds(std::max(0.0f,next_delay())));
+        boost::this_thread::sleep(boost::posix_time::milliseconds(static_cast<long>(std::max(0.0f,next_delay()))));
```
marcusmueller: same as above – nothing here should actually be floats.

gr-uhd/lib/usrp_block_impl.cc
```diff
@@ -207,7 +207,7 @@ bool usrp_block_impl::_wait_for_locked_sensor(
     else {
       first_lock_time = boost::system_time(); //reset to 'not a date time'
 
-      if (boost::get_system_time() > (start + boost::posix_time::seconds(LOCK_TIMEOUT))){
+      if (boost::get_system_time() > (start + boost::posix_time::seconds(static_cast<long>(LOCK_TIMEOUT)))){
```
marcusmueller: whereas the instances above a mere annoyances on a logic level, this is a real bug, because there's a serious difference between locking for 1.5s and 1.0s. CC: @mbr0wn

marcusmueller:
applied to both master and maint-3.7.

need to open issues about message_strobe[_random] and gr-uhd, as outlined above.

PR #1743, opened 2018-04-24
grc: make values of param type gui_hint callable again

skoslowski:
The templating for gui_hint was changed in #1418 breaking some oot modules.

### DIFF
--- a/grc/core/Param.py
+++ b/grc/core/Param.py
@@ -76,6 +76,17 @@ def num_to_str(num):
         return str(num)
 
 
+class CallableString(str):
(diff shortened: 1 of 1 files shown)


marcusmueller:
Nate said it fixed his issue, so I'll be applying this patch now.

marcusmueller:
also backported to maint-3.7. This is an API-breaking change. (This is not strictly an API-breaking thing as things being callable or not is not inherently truly "API" in Python, but let's face it: it changes an API breakage, and thus should be treated as API change).

Issue #1704, opened 2018-03-23
GRC Options should allow specifying copyright holder


dkozel:
The GRC Options block currently allows setting the Author which results in a comment at the top of the file. I suggest that the header always include the GPL-3.0 SPDX header and a copyright line if a copyright holder is specified.

https://spdx.org/licenses/GPL-3.0-only.html

``` python
#!/usr/bin/env python2
# -*- coding: utf-8 -*-
#
# SPDX-License-Identifier: GPL-3.0
#
##################################################
# GNU Radio Python Flow Graph
# Title: Example Title
# Author: Derek Kozel
# Copyright: 2018 Ettus Research, a National Instruments Company
# Generated: Fri Mar 23 11:16:05 2018
# GNU Radio Version: 3.7.12git-337-g76f8abeb
##################################################
```

marcusmueller:
I like that it's a sane default and helpful. Can we also include a "generated using GRC {version}" thing?

dkozel:
Yes! I had that in my mind this morning but it didn't make it into this issue. That would be very useful information for anyone running the flowgraph three years later.

haakov:
I submitted a pull request for this issue above.

How does this look?

```
#!/usr/bin/env python2
# -*- coding: utf-8 -*-
#
# SPDX-License-Identifier: GPL-3.0
#
##################################################
# GNU Radio Python Flow Graph
# Title: Example Title
# Author: Example Author
# Copyright: Example Copyright
# Description: Example Description
#
# Generated: Wed Apr 25 21:29:30 2018
# Generated using GRC version 3.7.12.0
##################################################
```

dkozel:
Yes, that output looks good to me. Thanks for the PR!

@marcusmueller There isn't actually a difference between GR version and GRC version right? I want to make sure there's a git hash (when relevant), but looking at the code it should be correct.

haakov:
https://github.com/gnuradio/gnuradio/pull/1745 has been merged, I believe this issue can be closed.

mbr0wn:
Well then....

PR #1745, opened 2018-04-25
grc: Added copyright holder to Options block

haakov:
As suggested in https://github.com/gnuradio/gnuradio/issues/1704.

### DIFF
--- a/grc/core/generator/Generator.py
+++ b/grc/core/generator/Generator.py
@@ -123,6 +123,7 @@ class TopBlockGenerator(object):
         output = list()
 
         fg = self._flow_graph
+        platform = fg.get_parent()
         title = fg.get_option('title') or fg.get_option('id').replace('_', ' ').title()
         imports = fg.get_imports()
         variables = fg.get_variables()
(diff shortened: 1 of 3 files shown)


grc/core/generator/Generator.py
```diff
@@ -246,6 +247,7 @@ def uses_var_id():
             'msgs': msgs,
             'generate_options': self._generate_options,
             'callbacks': callbacks,
+            'version': gr.version
```
skoslowski: GRC has its own version field. Check-out grc/core/Config.py
This is, of course, the same as gr.version, but it serves a the single source within grc code.
Please use that instead

grc/core/generator/flow_graph.tmpl
```diff
@@ -15,22 +15,31 @@
...
 #end if
+#if $flow_graph.get_option('copyright')
+# Copyright: $flow_graph.get_option('copyright')
+#end if
 #if $flow_graph.get_option('description')
 # Description: $flow_graph.get_option('description')
+$SINGLE_HASH
 #end if
 # Generated: $time.ctime()
+# Generated using GRC version $version()
```
skoslowski: to match style with the other fields, I'd prefer 'GNU Radio version: ...' instead.

haakov:
Sure! Fixed :+1:

dkozel:
@marcusmueller This PR is ready to be merged with your approval.

marcusmueller:
Thanks, everyone!

PR #1747, opened 2018-04-27
fix incorrect strings passed to the log console appender

catemarie:
The log console appender looks for "stdout"/"stderr", not "cout"/"cerr".  The strings used are incorrect in various files, causing log output to often be directed to stderr even when the config file specifies stdout.


m3x1m0m:
Another option would be to design logger_add_console_appender in a rather sane way and throw and exception if it is neither stdcout nor stdcerr or/and to accept cerr and cout.

PR #1749, opened 2018-04-28
Improved documentation for FFT block.

### LINKED ISSUE #1694: FFT and IFFT: Actual formula should be documented.
It is quite common to define FFT and IFFT as follows: 

```math 
X(a) = \sum_{k=0}^{N-1} x(a) e^(\frac{-j 2\pi k a}{N})
```
```math 
Y(b) = \frac{1}{N} \sum_{k=0}^{N-1} y(b) e^(\frac{j 2\pi k b}{N})
```
where N is the data length.

A nice property of this definition is
```math
X(Y) = x \neq N \cdot x
```
However in GNURadio [the output is scaled by a factor of N](https://gist.github.com/m3x1m0m/9963fe5fd764e44cae1c0a6b4d1b055d). I think it might be an enrichment to have this documented at least.

m3x1m0m:
Fixes #1694

### DIFF
--- a/gr-fft/include/gnuradio/fft/fft_vcc.h
+++ b/gr-fft/include/gnuradio/fft/fft_vcc.h
@@ -1,6 +1,6 @@
 /* -*- c++ -*- */
 /*
- * Copyright 2004,2007,2008,2012 Free Software Foundation, Inc.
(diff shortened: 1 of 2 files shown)


marcusmueller:
hey,  @m3x1m0m, thanks for the great work!!

PR #1750, opened 2018-04-29
gr-blocks: fix fast multiply block to allow vector input

karel:
This block ignored the 'Vec Length' parameter and when trying to actually input a vector and set the 'Vec Length' accordingly it'd complain "ValueError: itemsize mismatch: multiply_const_cc0:0 using 8, complex_to_mag_squared0:0 using 8192".

### DIFF
--- a/gr-blocks/grc/blocks_multiply_const_xx.xml
+++ b/gr-blocks/grc/blocks_multiply_const_xx.xml
@@ -19,25 +19,25 @@
 			<name>Complex</name>
 			<key>complex</key>
 			<opt>fcn:cc</opt>
-      <opt>const_type:complex</opt>
+			<opt>const_type:complex</opt>
 		</option>
 		<option>
 			<name>Float</name>
 			<key>float</key>
 			<opt>fcn:ff</opt>
-      <opt>const_type:float</opt>
+			<opt>const_type:float</opt>
 		</option>
 		<option>
 			<name>Int</name>
 			<key>int</key>
 			<opt>fcn:ii</opt>
-      <opt>const_type:int</opt>
+			<opt>const_type:int</opt>
 		</option>
 		<option>
 			<name>Short</name>
 			<key>short</key>
 			<opt>fcn:ss</opt>
-      <opt>const_type:int</opt>
+			<opt>const_type:int</opt>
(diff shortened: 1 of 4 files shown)


drmpeg:
I was the one that enabled this block with this commit.

https://github.com/gnuradio/gnuradio/commit/9566915e48d66bb4eff75d5bd136127499245224

But I have to admit I cheated a little. The vlen parameter in the .xml file is missing on purpose. That's because it's not present in the int and short versions of the block. If you add the vlen parameter like you've done in this PR, the block will fail with this error with ints and shorts.

```
  File "/home/re/xfer/top_block.py", line 114, in <module>
    main()
  File "/home/re/xfer/top_block.py", line 102, in main
    tb = top_block_cls()
  File "/home/re/xfer/top_block.py", line 68, in __init__
    self.blocks_multiply_const_xx_0 = blocks.multiply_const_ii(0, 1)
TypeError: make() takes exactly 1 argument (2 given)
```

So if you want to have vector I/O, this PR needs more work.

BTW, my original intent was that if you need vector I/O, you would use the Multiply Const block instead.

karel-:
Thanks for pointing this problem out, @drmpeg.

I opted to change this block because the Multiply Const block is noticeably slower than this one, given that I want to multiply all elements in the input vector with a single constant. I will look into the problem with ints and shorts.

willcode:
The block/file naming here is tricky!

Would you add a vlan (with default 1) to multiply_const_XX_impl.{h,cc} to fix ii and ss?

karel-:
Apparently this block doesn't work for shorts regardless of my proposed changes

```
Param - Constant(const):
	Type "short" is not a possible type.

Param - Constant(const):
	Type "short" not handled
```

@willcode, I have added the `vlen` parameter to the template files. I will commit as soon as I get this "short" error sorted out.

bhilburn:
Hello!

@karel - Can you send me an e-mail, `bhilburn@gnuradio.org`? As soon as we get the CLA sorted out, we'll be set =)

marcusmueller:
Got the CLA; merging!

Thank you, @karel, for your contribution, as well as your saint-like patience :)

PR #1754, opened 2018-04-30
gr-blocks: use libvolk to speed up nlog10_ff block

karel:
Using libvolk instead of the C double precision `log10()` is faster by about an order of magnitude.

### DIFF
--- a/gr-blocks/lib/nlog10_ff_impl.cc
+++ b/gr-blocks/lib/nlog10_ff_impl.cc
@@ -27,6 +27,7 @@
 #include "nlog10_ff_impl.h"
 #include <gnuradio/io_signature.h>
 #include <volk/volk.h>
+#include <limits>
 
 namespace gr {
   namespace blocks {
@@ -40,7 +41,8 @@ namespace gr {
       : sync_block("nlog10_ff",
 		      io_signature::make (1, 1, sizeof(float)*vlen),
 		      io_signature::make (1, 1, sizeof(float)*vlen)),
-        d_n_log2_10(n/log2f(10.0f)), d_vlen(vlen), d_k(k)
+        d_n_log2_10(n/log2f(10.0f)), d_float_min(std::numeric_limits<float>::min()),
+        d_vlen(vlen), d_k(k)
     {
       const int alignment_multiple =
         volk_get_alignment() / sizeof(float);
@@ -69,7 +71,7 @@ namespace gr {
       int noi = noutput_items * d_vlen;
 
       for (int i = 0; i < noi; i++) {
-          out[i] = in[i] + 1e-18;
+          out[i] = in[i] + d_float_min;
       }
 
       volk_32f_log2_32f(out, out, noi);
--- a/gr-blocks/lib/nlog10_ff_impl.cc
+++ b/gr-blocks/lib/nlog10_ff_impl.cc
@@ -68,13 +68,9 @@ namespace gr {
       float *out = (float *) output_items[0];
       int noi = noutput_items * d_vlen;
 
-      volk_32f_log2_32f(out, in, noi);
-      volk_32f_s32f_multiply_32f(out, out, d_prefactor, noi);
-      if(d_k != 0.0f) {
-        for(int i = 0; i < noi; ++i) {
-          out[i] += d_k;
-        }
-      }
+      for (int i = 0; i < noi; i++)
+        out[i] = n * log10f(std::max(in[i], (float) 1e-18)) + k;
+
       return noutput_items;
     }
 
--- a/gr-blocks/lib/nlog10_ff_impl.cc
+++ b/gr-blocks/lib/nlog10_ff_impl.cc
@@ -48,18 +48,6 @@ namespace gr {
       set_alignment(std::max(1,alignment_multiple));
     }
 
-    void
-    nlog10_ff_impl::setk(float k)
-    {
-      d_k = k;
-    }
-
-    void
-    nlog10_ff_impl::setn(float n)
-    {
-      d_prefactor = n / log2f(10.0f);
-    }
-
     int
     nlog10_ff_impl::work(int noutput_items,
 			      gr_vector_const_void_star &input_items,
--- a/gr-blocks/lib/nlog10_ff_impl.cc
+++ b/gr-blocks/lib/nlog10_ff_impl.cc
@@ -40,7 +40,7 @@ namespace gr {
       : sync_block("nlog10_ff",
 		      io_signature::make (1, 1, sizeof(float)*vlen),
 		      io_signature::make (1, 1, sizeof(float)*vlen)),
-        d_n(n), d_vlen(vlen), d_k(k)
+        d_n_log2_10(n/log2f(10.0f)), d_vlen(vlen), d_k(k)
     {
       const int alignment_multiple =
         volk_get_alignment() / sizeof(float);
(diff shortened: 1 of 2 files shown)


gr-blocks/lib/nlog10_ff_impl.cc
```diff
@@ -54,8 +58,17 @@ namespace gr {
       float n = d_n;
       float k = d_k;
 
-      for (int i = 0; i < noi; i++)
-        out[i] = n * log10f(std::max(in[i], (float) 1e-18)) + k;
+      volk_32f_log2_32f(out, in, noi);
+      volk_32f_s32f_multiply_32f(out, out, n/log2f(10.0f), noi);
```
awalls-cx18: The value "n/log2f(10.0f)" can be precomputed and kept in a class variable d_n_log2_10 which would be updated when d_n is updated.

gr-blocks/lib/nlog10_ff_impl.cc
```diff
@@ -51,11 +55,18 @@ namespace gr {
       const float *in = (const float *) input_items[0];
       float *out = (float *) output_items[0];
       int noi = noutput_items * d_vlen;
-      float n = d_n;
-      float k = d_k;
 
-      for (int i = 0; i < noi; i++)
-	out[i] = n * log10(std::max(in[i], (float) 1e-18)) + k;
+      for (int i = 0; i < noi; i++) {
+          out[i] = in[i] + 1e-18;
```
mbr0wn: Hm, so this does change behaviour, though, I'm just not sure if it's significant. If you did
```cpp
out[i] = std::max<float>(in[i], 1e-18);
```
...you'd retain the behaviour, but obv that would incur a new performance penalty. I would assume it's still faster than the current behaviour.

mbr0wn: I think the loss of accuracy is fine. We handle noisy data usually, after all. However, I would like it if we mention this in the documentation for this block.

awalls-cx18:
On May 6, 2018 12:28:19 AM EDT, Martin Braun <notifications@github.com> wrote:
>mbr0wn commented on this pull request.
>
>
>
>>  
>-      for (int i = 0; i < noi; i++)
>-	out[i] = n * log10(std::max(in[i], (float) 1e-18)) + k;
>+      for (int i = 0; i < noi; i++) {
>+          out[i] = in[i] + 1e-18;
>
>I think the loss of accuracy is fine. We handle noisy data usually,
>after all. However, I would like it if we mention this in the
>documentation for this block.

And the added number just has to be *something* larger than 0.0f.  It could be 1e-24 or something smaller, as long as the log2 function returns a real number.

-Andy

karel-:
@awalls-cx18 indeed, and the loss of precision is also smaller than it was originally.

gr-blocks/lib/nlog10_ff_impl.cc
```diff
@@ -51,11 +57,18 @@ namespace gr {
       const float *in = (const float *) input_items[0];
       float *out = (float *) output_items[0];
       int noi = noutput_items * d_vlen;
-      float n = d_n;
-      float k = d_k;
 
-      for (int i = 0; i < noi; i++)
-	out[i] = n * log10(std::max(in[i], (float) 1e-18)) + k;
+      for (int i = 0; i < noi; i++) {
+          out[i] = in[i] + d_float_min;
```
marcusmueller: That's not a solution! For example,`in[i] == -1.0f` is still negative after adding the minimum float.

No, we *have* to do `max(in,float_min)` in some way. I wish we had a VOLK kernel like `volk_32f_s32f_max_32f`, but we don't, do we, @n-west ? We can only compare two vectors of equal size and take the element-wise maximum.

gr-blocks/lib/nlog10_ff_impl.h
```diff
@@ -30,7 +30,8 @@ namespace gr {
 
     class BLOCKS_API nlog10_ff_impl : public nlog10_ff
     {
-      float  d_n;
+      float  d_n_log2_10;
+      float  d_float_min;
```
marcusmueller: don't add instance properties just to hold constants! (these constants can't be different for different instances of the block, right?)

Use `static const float` here, and call them `s_float_min`, for example.

But: let's be honest: rather go ahead and don't have it as class member at all. A `float min==std::numeric_limits<float>::min()` inside your `work` would achieve the same (as any compiler would see the constness of that, it's a `constexpr` starting with C++11!).

gr-blocks/lib/nlog10_ff_impl.cc
```diff
@@ -51,11 +57,18 @@ namespace gr {
...
+      }
+
+      volk_32f_log2_32f(out, out, noi);
+      volk_32f_s32f_multiply_32f(out, out, d_n_log2_10, noi);
+      if (d_k != 0.0f) {
+        for (int i = 0; i < noi; i++) {
+          out[i] += d_k;
```
marcusmueller: Idea: instead of adding `d_k` here, we could just multiply the input with 2<sup>d_k</sup> and check whether it's truly > 0. Multiplication of floats is actually not any slower or less accurate than addition (quite the contrary, actually).

marcusmueller: Bonus: we have a VOLK multiply const kernel, `volk_32f_s32f_multiply_32f` :)

marcusmueller: (this is just an idea on how to speed things up further, and bears absolutely no criteria for acceptance of this PR! )

karel:
I've reintroduced the ```std::max``` comparison to fix negative input issues, removed the useless instance property of a constant, and also implemented @marcusmueller's idea of doing multiplication instead of addition (I believe it's 10<sup>(k/n)</sup> not 2<sup>k</sup>). Multiplication via libvolk kernel does seem to be faster indeed, thanks! However, anything from libvolk seems to be causing large bursts of instability at the moment.

mbr0wn:
@karel By instability, you mean reliability with respect to building? Or actual run-time instability?

karel:
Run-time instability. Seemed like whole blocks of samples were messed up, which is curious because this block doesn't have history and only a single input sample is used for each output sample. I was thinking maybe it's an alignment issue with libvolk.

marcusmueller:
With quite a heavy heart, I put the "don't merge" tag onto this until we've sorted out why it fails :(

karel:
Apparently it does not fail. The instability appeared when I was using the following flowgraph. Visually the noise floor made some 40+ dB jumps with the proposed implementation. Comparing the outputs of current and proposed implementations in code, however, doesn't indicate such mismatch. Difference between the outputs of those implementations is negligible.

latest-screenshot

karel:
I rebased the branch.

noc0lour:
@karel thanks :) One possible "fix" for this block would be setting the output_multiple to something every volk implementation is capable of handling with SIMD-only. That would at least provide consistent results for SIMD. 
Other than that I think we have to fix the nlog10_ff implementation in volk.

gr-blocks/lib/nlog10_ff_impl.cc
mbr0wn: Final version has the `std::max()` in it.

gr-blocks/lib/nlog10_ff_impl.cc
mbr0wn: Now using `std::max()`.

mbr0wn:
@marcusmueller Added "Don't Merge" because @karel said there were issues, but then there weren't any actually. Removing that label.

mbr0wn:
This PR is ripe for merging. I did another rebase, retest, and modified some small things:
- Changed commit messages according to our format
- Updated copyright headers
- Added `constexpr` to be explicit

Opened a new PR for merging: #2165

PR #1759, opened 2018-05-04
uhd: fix freq_hopping example - missing 'time'

ptrkrysik:
Freq_hopping example wasn't working correctly. Symptom of it was that frequency changes were happening in random moments. The reason was that time of execution of tune commands wasn't specified and so they were executed as non-timed commands.

The fix was to add missing 'time' specified as pmt.cons to 'tx_command' tag value.

In case of of full tuning (dsp_tuning == false) a bit more changes were needed. 'tx_freq' tag can't be executed as timed command at all, so I changed it to 'tx_command' with 'freq' element and also added missing 'time'.

I found out about the issue but credit of fixing it by adding 'time' element to 'tx_command' tag value goes to Sylvain Munaut, as this change was suggested by him.

### DIFF
--- a/gr-uhd/examples/python/freq_hopping.py
+++ b/gr-uhd/examples/python/freq_hopping.py
@@ -109,6 +109,10 @@ class FrequencyHopperSrc(gr.hier_block2):
         gain_tag.value = pmt.to_pmt({'gain': tx_gain})
         tag_list = [gain_tag,]
         for i in xrange(len(self.hop_sequence)):
+            time = pmt.cons(
+                    pmt.from_uint64(int(base_time + i * hop_time+0.01)),
+                    pmt.from_double((base_time + i * hop_time+0.01) % 1),
+            )
             tune_tag = gr.tag_t()
             tune_tag.offset = i * burst_length
             if i > 0 and post_tuning and not dsp_tuning: # TODO dsp_tuning should also be able to do post_tuning
@@ -116,9 +120,11 @@ class FrequencyHopperSrc(gr.hier_block2):
             if dsp_tuning:
                 tune_tag.key = pmt.string_to_symbol('tx_command')
                 tune_tag.value = pmt.to_pmt({'lo_freq': base_freq, 'dsp_freq': base_freq - self.hop_sequence[i]})
+                tune_tag.value = pmt.dict_add(tune_tag.value, pmt.intern("time"),time)
             else:
-                tune_tag.key = pmt.string_to_symbol('tx_freq')
-                tune_tag.value = pmt.to_pmt(self.hop_sequence[i])
+                tune_tag.key = pmt.string_to_symbol('tx_command')
+                tune_tag.value = pmt.to_pmt({'freq': self.hop_sequence[i]})
+                tune_tag.value = pmt.dict_add(tune_tag.value, pmt.intern('time'), time)
             tag_list.append(tune_tag)
             length_tag = gr.tag_t()
             length_tag.offset = i * burst_length
@@ -129,8 +135,8 @@ class FrequencyHopperSrc(gr.hier_block2):
             time_tag.offset = i * burst_length
             time_tag.key = pmt.string_to_symbol('tx_time')
             time_tag.value = pmt.make_tuple(
-                    pmt.from_uint64(int(base_time + i * hop_time)),
-                    pmt.from_double((base_time + i * hop_time) % 1),
+                    pmt.car(time),
+                    pmt.cdr(time)
             )
             tag_list.append(time_tag)
         tag_source = blocks.vector_source_c((1.0,) * n_samples_total, repeat=False, tags=tag_list)
@@ -217,4 +223,3 @@ if __name__ == '__main__':
         main()
     except KeyboardInterrupt:
         pass
-


gr-uhd/examples/python/freq_hopping.py
```diff
@@ -109,16 +109,22 @@ def __init__(
...
+                    pmt.from_uint64(int(base_time + i * hop_time+0.01)),
+                    pmt.from_double((base_time + i * hop_time+0.01) % 1),
+            )
             tune_tag = gr.tag_t()
             tune_tag.offset = i * burst_length
             if i > 0 and post_tuning and not dsp_tuning: # TODO dsp_tuning should also be able to do post_tuning
                 tune_tag.offset -= 1 # Move it to last sample of previous burst
             if dsp_tuning:
                 tune_tag.key = pmt.string_to_symbol('tx_command')
                 tune_tag.value = pmt.to_pmt({'lo_freq': base_freq, 'dsp_freq': base_freq - self.hop_sequence[i]})
```
mbr0wn: Would look much nicer if the time gets rolled into the initial dictionary.

gr-uhd/examples/python/freq_hopping.py
```diff
@@ -129,8 +135,8 @@ def __init__(
             time_tag.offset = i * burst_length
             time_tag.key = pmt.string_to_symbol('tx_time')
             time_tag.value = pmt.make_tuple(
-                    pmt.from_uint64(int(base_time + i * hop_time)),
-                    pmt.from_double((base_time + i * hop_time) % 1),
+                    pmt.car(time),
```
mbr0wn: Can't you simply do `time_tag.value = time`?

gr-uhd/examples/python/freq_hopping.py
ptrkrysik: I don't know yet how to do the squash without making another github pull request. I'll do that as soon as I learn that.

Regarding rolling it into initial dictionary - I did it this way because:
* the initial dictionary is a Python's dictionary that is converted into pmt object,
* I don't know how to confine types so after the conversion from python objects to pmt we get exactly pmt.cons with uint64 as car and double as cdr.

Therefore I preferred to create 'time' the pmt way. I can make this part more consistent by creating whole tune_tag.value in pmt way, instead of doing part by converting from Python and part using directly pmt.

ptrkrysik:
So to summarize:
1. I have been looking how to squash the commits myself without starting another pull request - without success yet. Can you do the squash during merge or rebase?
2. Regarding long lines - the lines I added are shorter than other lines in the file. Breaking only them would make the file look inconsistent in my opinion.
3. Regarding adding 'time' to python dictionary - I don't think it is possible as conversion of pmt cons to python and then back to pmt gives object that isn't pmt cons. Should I create the dictionary pmt way to make this part look more consistent?

noc0lour:
@ptrkrysik you should be able to squash and force push your branch, this will result in an updated Pullrequest here. Though only do that if you have resolved open issues with it.

ptrkrysik:
@noc0lour thanks, I did the squash. Should I change the way the 'tx_command' values are created from using both pmt.to_pmt(...) and pmt.dict_add(...) to only pmt.dict_add(...)? For pmt.to_pmt(...) only I don't know how to do it and assure that it will be done well. I think it might be risky (i.e. cause lose of precision) to define it this way even if it's possible.

ptrkrysik:
Knock, knock, I changed what needed changing and presented rationale for other things that I didn't change. Now it's your move :).

marcusmueller:
@mbr0wn update^^?

gr-uhd/examples/python/freq_hopping.py
mbr0wn: Yeah that makes sense.

dkozel:
@marcusmueller Martin and I spent some time yesterday on this PR. It has a mostly correct fix but there is some additional supporting work needed in gr-uhd to fully address the problem. Hopefully I can give a meaningful update tomorrow or Monday.

ptrkrysik:
@dkozel, can you explain what might be missing in gr-uhd? I found out that the code to control the frequency in both directions  from single block with "direction" key is not fully correct. Currently only one direction can be controlled at a time and I'm trying to figure out some clean solution. But I'm not aware of anything else.

ptrkrysik:
Hi guys, I did one mistake when submitting this issue - I made the pull request from the master branch of my forked repository of GNU Radio. It was probably because I thought it will be a fast issue to resolve and didn't pay enough attention to problems that might arise by creating pull request this way.

Now the pull request is hanging for 5 moths and I can't update my master branch. There are some changes I would try to make to uhd blocks (among others) but it would create even more mess in my forked repo.

Can I resolve this by closing this pull request and creating another one (from another branch than master) and link in the description to this pull request?

dkozel:
@ptrkrysik I think you can even change what your upstream branch is without closing the PR. If that doesn't work then yes, go a head and close this PR and re-open using a side branch in your repo.

By the way, here's the branch that @mbr0wn and I were working on to address the problems in making the freq_hopping example work. The additional issues are in the gr-uhd code which deals how timed commands are executed.
https://github.com/mbr0wn/gnuradio/commits/uhd_freq_hopping

The issue (as I remember) is that the "tx_command" is always used and currently gr-uhd does not support it unless it is on the first sample of the burst. This means that post-tuning is broken in the freq_hopping example. Since post-tuning is virtually always preferable that's definitely worth fixing.
https://github.com/gnuradio/gnuradio/blob/master/gr-uhd/lib/usrp_sink_impl.cc#L605

@mbr0wn thoughts on how to close up this PR?

ptrkrysik:
I can edit only the target branch, the source repository and branch not editable. This is what I see:
https://blog.github.com/2016-08-15-change-the-base-branch-of-a-pull-request/

As I have some other gr-uhd related changes, I'll be closing this pull request and I'll reopen it with new 'from' branch. I'll also check what other changes were made to this file as currently there is some conflict.

ptrkrysik:
In the master the gr-uhd/examples/python/freq_hopping.py file is gone. I'll try to make a pull request to main-3.7 and if it won't be correct you can just close the new pull request.

dkozel:
@mbr0wn Can I get an opinion on Piotr's question above? I think getting this merged with an error thrown in the post-tuning case and filing an issue specifically addressing post-tuning is the best start here.

mbr0wn:
@ptrkrysik, sorry, I really wanted to address this PR before I leave on holidays but I probably won't get around to.

I'm not sure why the python examples aren't on master. I'm pretty sure that's a mistake of some kind. 21 files were deleted from gr-uhd and I don't think that's supposed to happen.

Like @dkozel said, there were some issues with this PR. I can't remember the details, but a few months ago he and I tried this code, and it was making some bad assumptions or something like that, although overall, this fix is absolutely relevant.

mbr0wn:
...actually, the examples were removed as part of the wxgui removal. That seems wrong...

mbr0wn:
OK, I had to find the original example and merge that back. Then I had to rebase your example, which is why I'm moving this to a separate branch and replacing this PR.

PR #1765, opened 2018-05-07
RFC: Replace cppunit with Boost UTF

mbr0wn:


mbr0wn:
I'm hoping that '+181 −675 lines' and the removal of a whole dependency makes this case compelling.

mbr0wn:
See also: https://github.com/gnuradio/greps/pull/13

cmake/Modules/GrTest.cmake
```diff
@@ -141,3 +141,26 @@ function(GR_ADD_TEST test_name)
...
+
+########################################################################
+# Add a C++ unit test and setup the environment for a unit test.
+# Takes the same arguments as the ADD_TEST function.
+#
+# test_name -- An identifier for your test, for usage with ctest -R
+# test_source -- Path to the .cc file
+#
+# Before calling set the following variables:
+# GR_TEST_TARGET_DEPS  - built targets for the library path
```
marcusmueller: why not make this a parameter and make it more "functionally" programmed?

marcusmueller: (I'd accept a "because it's not the way we deal with this kind of thing around our CMake code, especially in this particular file" as valid explanation. So, this is not a blocker.)

gr-filter/lib/qa_mmse_fir_interpolator_ff.cc
```diff
@@ -61,7 +59,7 @@ namespace gr {
 	  float expected = test_fcn((i + 3) + imu * inv_nsteps);
 	  float actual = intr.interpolate(&input[i], imu * inv_nsteps);
 
-	  CPPUNIT_ASSERT_DOUBLES_EQUAL(expected, actual, 0.004);
+	  BOOST_CHECK(std::abs(expected - actual) <= 0.004);
 	  // printf ("%9.6f  %9.6f  %9.6f\n", expected, actual, expected - actual);
```
marcusmueller: A comment, not a request for change in your patch:

Generally, when fundamentally touching code, don't hesitate to get rid of such commented out lines of code that don't contribute anything in terms of e.g. illustrating how something was meant to be used.

marcusmueller:
In fact, in the medium run we want to do this to *all* modules, not just digital, analog, blocks and filter.

PR #1766, opened 2018-05-08
grc: Evaluate gtk icon in generated code instead of template compile …

anuraaga:
…ve so it is only written when gui is enabled.

A cheetah `#import` statement causes the import to be placed at the top of the generated module, not at the location of the statement. This means that even when `nogui` is specified, flow graphs have an unnecessary dependency on `gtk` existing when compiling the Cheetah template. So instead of evaluating `icon` during template compilation, it evaluates it at runtime so the code is only run when `generate_options == wx_gui`. This seems like a good idea either way since it means the generated file could be moved to another computer which may not have the icon available.

### DIFF
--- a/grc/core/generator/flow_graph.tmpl
+++ b/grc/core/generator/flow_graph.tmpl
@@ -78,18 +78,17 @@ $imp
 #set $class_name = $flow_graph.get_option('id')
 #set $param_str = ', '.join(['self'] + ['%s=%s'%(param.get_id(), param.get_make()) for param in $parameters])
 #if $generate_options == 'wx_gui'
-    #import gtk
-    #set $icon = gtk.IconTheme().lookup_icon('gnuradio-grc', 32, 0)
+import gtk
 
 
 class $(class_name)(grc_wxgui.top_block_gui):
 
     def __init__($param_str):
         grc_wxgui.top_block_gui.__init__(self, title="$title")
-    #if $icon
-        _icon_path = "$icon.get_filename()"
-        self.SetIcon(wx.Icon(_icon_path, wx.BITMAP_TYPE_ANY))
-    #end if
+        icon = gtk.IconTheme().lookup_icon('gnuradio-grc', 32, 0)
+        if icon
+            _icon_path = icon.get_filename()
+            self.SetIcon(wx.Icon(_icon_path, wx.BITMAP_TYPE_ANY))
 #elif $generate_options == 'qt_gui'
 from gnuradio import qtgui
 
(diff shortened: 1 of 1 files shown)


grc/core/generator/flow_graph.tmpl
```diff
@@ -78,18 +78,17 @@ $imp
 #set $class_name = $flow_graph.get_option('id')
 #set $param_str = ', '.join(['self'] + ['%s=%s'%(param.get_id(), param.get_make()) for param in $parameters])
 #if $generate_options == 'wx_gui'
-    #import gtk
-    #set $icon = gtk.IconTheme().lookup_icon('gnuradio-grc', 32, 0)
+import gtk
```
skoslowski: This introduces a runtime-dep on gtk into every wxgui flowgraph

grc/core/generator/flow_graph.tmpl
```diff
@@ -78,18 +78,17 @@ $imp
...
 class $(class_name)(grc_wxgui.top_block_gui):
 
     def __init__($param_str):
         grc_wxgui.top_block_gui.__init__(self, title="$title")
-    #if $icon
-        _icon_path = "$icon.get_filename()"
-        self.SetIcon(wx.Icon(_icon_path, wx.BITMAP_TYPE_ANY))
-    #end if
+        icon = gtk.IconTheme().lookup_icon('gnuradio-grc', 32, 0)
+        if icon
```
skoslowski: pretty sure a colon is missing here.

grc/core/generator/flow_graph.tmpl
anuraaga: Oof sorry was copying this change in from hacks in a docker container

grc/core/generator/flow_graph.tmpl
anuraaga: Oof sorry was copying this change in from hacks in a docker container

anuraaga:
I'm not sure what the icon looks like since always use `nogui` :) To avoid the `gtk` dependency, does it make sense to just remove the icon completely until `wxgui` itself is too?

grc/core/generator/flow_graph.tmpl
marcusmueller: I honestly don't know whether there's no inherent gtk dependency in wx or its python bindings. If there isn't, we can't do this, I'm afraid.

marcusmueller: @anuraaga  not quite sure what you mean with "compile time dependency on gtk for all graphs": that doesn't reflect my understanding of what GNU Radio is. Can you elaborate, please?

anuraaga: I mean that when compiling a graph, for example using `grcc`, the system must have `gtk` installed, regardless of what the content of the graph is. This is even when compiling a `nogui` graph. After compilation, the resulting python will not depend on `gtk` but it's bad that we can't even compile the graphs without the dependency installed IMO.

marcusmueller:
it took me a while, but I've made up my mind: we can't depend on gtk in WX flow graphs (e.g. windows, OS X typically don't have WxGTK, but other backends), so this fix won't do.

Furthermore, WX will be gone on master, so you'd need to backport it to maint-3.7. 

That means that of this PR, there's little left to salvage :( I'm sorry!

Issue #1757, opened 2018-05-01
set_output_multiple missing in block.i


willcode:
Reported by Jessica Iwamoto. Need to add:

  void set_output_multiple(int multiple)

to gnuradio-runtime/swig/block.i.

PR #1773, opened 2018-05-12
swig block.i: add void set_output_multiple(int multiple)

willcode:
Resolve https://github.com/gnuradio/gnuradio/issues/1757


dkozel:
Hi @willcode Sorry for the delay here. I've just triggered a rebuild of the CI system. Did you have a test or application which used this function in Python?

willcode:
@dkozel - no application, this is just a missing Python wrapper pointed out/requested by a user.

marcusmueller:
Nice! Thanks.

marcusmueller:
Still, if you'd have a separate test case for this, and be it really minimal, I'd like to have that, too!

PR #1779, opened 2018-05-16
move declaration to case

### LINKED ISSUE #1774: Error with "Float to Complex" block if only one input port is connected
I used this sample flowgraph to reproduce:
grafik

I get this error message:
`/usr/include/c++/8/bits/stl_vector.h:932: std::vector<_Tp, _Alloc>::reference std::vector<_Tp, _Alloc>::operator[](std::vector<_Tp, _Alloc>::size_type) [with _Tp = const void*; _Alloc = std::allocator<const void*>; std::vector<_Tp, _Alloc>::reference = const void*&; std::vector<_Tp, _Alloc>::size_type = long unsigned int]: Assertion '__builtin_expect(__n < this->size(), true)' failed.`

If the "Null Source" on the second input of the "Float to Complex" block is enabled, it works without any problems.

I use Fedora 28 and GNU Radio Companion 3.7.11 (installed with dnf)

om3gafett:
In order to fix  #1774

### DIFF
--- a/gr-blocks/lib/float_to_complex_impl.cc
+++ b/gr-blocks/lib/float_to_complex_impl.cc
@@ -54,8 +54,7 @@ namespace gr {
 				gr_vector_const_void_star &input_items,
 				gr_vector_void_star &output_items)
     {
-      float        *r = (float *)input_items[0];
-      float        *i = (float *)input_items[1];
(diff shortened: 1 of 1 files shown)


marcusmueller:
@noc0lour please kick the builder!

PR #1785, opened 2018-05-18
Short documentation for some GRC blocks

Notou:
I've begun adding a doc to the grc xml block files so I don't need to always search through the codebase for implementation details to understand what a block does with its parameters.

I started with some really simple blocks that I was using at the moment since I needed to go through their source code anyways to use them.

I will certainly add more blocks later as I get to use them. Should I do a pull request for each separate block or as batches?

The philosophy I decided to follow was to do a short description of what the block does, then the effect of the parameters and finally the path to the source file because It's always a pain to search through the different modules for the source file that doesn't always have the same name as the block.

If you have some advice on the formalism I should use to doc the blocks, or anything, I'll take it.

### DIFF
--- a/gr-blocks/grc/blocks_moving_average_xx.xml
+++ b/gr-blocks/grc/blocks_moving_average_xx.xml
@@ -74,4 +74,31 @@
 		<type>$type</type>
 		<vlen>$vlen</vlen>
 	</source>
+	<doc>
+Computes a moving average of the input:
+Output[i] = scale * sum(input[i-length: i])
+
+In its default parameters, the block actually computes a moving sum.
+
+Length:
+The size of the moving average window to used.
+
+Scale:
+Factor to scale the sum of the last (Length) samples.
+To get an actual moving average, the scale should then be set to 1/Length.
+
+Max Iter:
+The maximum number of samples the block will treat in one call of its work function.
+
+vlen:
+Used if the input samples are vectors and correspond to the length of those vectors.
+When operating on vectors, the average is done using only the numbers from the same vector index
+Output[i][vector_index] = scale * sum(Input[i-length: i][vector_index])
+
+
+Implementation file:
+gr-blocks/lib/moving_average_XX_impl.cc.t
+GRC file:
+gr-blocks/grc/blocks_moving_average_xx.xml
+	</doc>
 </block>
--- a/gr-blocks/grc/blocks_nlog10_ff.xml
+++ b/gr-blocks/grc/blocks_nlog10_ff.xml
@@ -39,4 +39,19 @@
 		<type>float</type>
 		<vlen>$vlen</vlen>
 	</source>
+    <doc>
+        Computes this formula on each input sample:
+output = n*log10(input) + k
+
+Actual formula is output = n*log10( max(input, 1e-18) ) + k to ensure things do not break when input <= 0.
+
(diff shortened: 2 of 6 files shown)


dkozel:
Hi @Notou Thanks for the PR. The documentation of many blocks can certainly use improvement.

The text of the documentation tab is generated both from the XML (soon to be YAML!) files and from the doxygen comments of the C++ files. I took a look at the log10 block's documentation and here's what currently appears in GRC.

```
blocks_nlog10_ff
make(float n=1.0, size_t vlen=1, float k=0.0) -> nlog10_ff_sptr
        output = n*log10(input) + k
        Constructor Specific Documentation:
        Make an instance of an nlog10_ff block.
        Args:
            n : Scalar multiplicative constant
            vlen : Input vector length
            k : Scalar additive constant
```

This is technically correct, but doesn't contain a few of the details you you've added such as the `vlen` or the actual formula.

> Actual formula is output = n*log10( max(input, 1e-18) ) + k to ensure things do not break when input <= 0.

I think your time would be best spent improving the doxygen comments in the C++ headers. @777arc is looking into improving how the documents are formatted and where various types of documentation would best live.

We are on the verge of moving from XML files to yaml. Although comments are being carried over automatically, many of the files have already been converted and it would be unfortunate to make extra work. If you do find that documentation in the GRC block definition files is the best option then working on the yaml files would be better.

dkozel:
I do quite like the idea of having the path to the XML/YAML file and the C++ source file included, but would strongly prefer that they be auto-generated during build. It could probably be done at the same time as the doxygen comment inclusion.

Notou:
Hi, I also think It would be better to improve the doxygen comments but in the version that I have, which is the latest from the main branch, the documentation tab only shows the prototype of the function. Even for the nlog10 block (I actually started because of that one).
Is that something that is present in the next branch?

Notou:
Hi @dkozel, I'm using the main branch at the moment and GRC doesn't use the Doxygen comments. Is this a feature that has been introduced in the next or python3 branch? Or maybe it's an issue in the latest commits?

777arc:
@Notou Yeah I'm also only seeing:

777arc:
@Notou I'm also not seeing that full description Derek showed.  If I look at the file blocks/nlog10_ff.h, I definitely see those comments, and the file was last modified in 2014 so it's not a recent change.  Let me track down what's going on here

Notou:
Hi, thank you for reminding me of this thread.
It would seem that Doxygen was not activated by default in Cmake when building from source (at least with pybombs).

777arc:
Oh thats not good

777arc:
@Notou @dkozel  I put in a PR in pybombs that will make building doxygen ON by default, I'm not sure why it was disabled by default.  I verified that it fixes the issue, hopefully it doesn't cause any bugs related to OOTs with broken doxygen code.

777arc:
Closing.  Summary of what happened here: @Notou built GR without building doxygen turned on (not his fault, for many months it was disabled by default when you used PyBOMBS to install GR), and thus it appeared as though these blocks did not have documentation, so @Notou was nice enough to add documentation to the XML.  Since then we have made sure build doxygen is set to true by default, [see this PR](https://github.com/gnuradio/pybombs/pull/508/).  @Notou 's additions are largely redundant info with the documentation in the docstrings.

PR #1787, opened 2018-05-18
Replace uses of std::auto_ptr with std::unique_ptr

pwicks86:
auto_ptr is deprecated in C++11 and will be removed in C++17. Compiling GNURadio in GCC 7 is quite noisy due to this. I'm pretty sure that the behavior of unique_ptr should be identical to auto_ptr in all of GNURadio's uses.

This PR probably can't be applied to master, since it requires C++11.


marcusmueller:
Isn't this already reality on `next`?

pwicks86:
Oops, yeah, looks like it. I'll go ahead and close this.

PR #1788, opened 2018-05-19
modtool: modify str.translate() for python3 comapatibility

swap-nil7:
str.translate() accepts just one argument in python3.x, the transition table rather than 2 in python2.x and str.maketrans() is a static method that returns a translation table usable for str.translate() in python3.


primercuervo:
This looks about right. You have a typo in the commit message, though.

swap-nil7:
@primercuervo: Updated the commit message.

PR #1790, opened 2018-05-22
Including issue_stream_cmd_on_start in arguments to usrp_source

bpkempke:
The static list of parameters passed to the usrp_source swig block in `uhd/__init__.py` does not include issue_stream_cmd_on_start, which is used by the usrp_source block.


marcusmueller:
This is @mbr0wn territory.

mbr0wn:
@bpkempke Thanks a lot for this fix! I had to rebase, and then fix other stuff, which in the end made it easier just to rewrite it (I usually prefer to keep people's contributions flagged as such, but in this one-word change it was difficult). I'm merging this manually along with a couple of other updates.

PR #1792, opened 2018-05-22
Fixing the MIN_INT to be actual MIN_INT

soggysec:
There was an off-by-one in the MIN_INT

### DIFF
--- a/gr-blocks/lib/float_array_to_int.cc
+++ b/gr-blocks/lib/float_array_to_int.cc
@@ -30,7 +30,7 @@
 #include <stdint.h>
 
 static const int64_t MAX_INT =  2147483647; //  (2^31)-1
-static const int64_t MIN_INT = -2147483647; // -(2^31)-1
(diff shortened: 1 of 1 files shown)


pwicks86:
cplusplus.com is a bad reference, it's incorrect in this case. You can check the actual values in either the climits header in C or via `std::numeric_limits` and verify the actual value. See http://cpp.sh/6yuyx

noc0lour:
cplusplus.com is actually correct `* the actual value depends on the particular system and library implementation, but shall reflect the limits of these types in the target platform.` 
And every `_MIN` is suffixed with a `or less*` stanza. Nevertheless we should be using 
`<cstdint>` with `INT*N*_MIN` macros here. Since this is not 3.7 anymore `C++11` should be allowed, right?

noc0lour:
Err, just use `INT32_MIN`, no C++11 involved.

marcusmueller:
backporting this to maint-3.7, too.

PR #1794, opened 2018-05-23
Add support for serializing PMTs containing longs outside of 32-bit range

storborg:
Currently, there is a limitation in PMT serialization and deserialization which prevents the serialization of PMT longs which are outside of the expressible range of a 32-bit signed int. This PR removes that limitation, by adding a new PST (PMT serial tag) for signed 64-bit ints.

This change retains partial backwards compatibility: PMTs which are serialized by older versions of GNU Radio will be deserialized correctly by this version, but if a PMT containing a large 64-bit int is serialized by this version, it will not be deserializeable by older versions.

### DIFF
--- a/gnuradio-runtime/lib/pmt/pmt-serial-tags.scm
+++ b/gnuradio-runtime/lib/pmt/pmt-serial-tags.scm
@@ -35,6 +35,7 @@
 (define pst-uniform-vector	#x0a)
 (define pst-uint64	#x0b)
 (define pst-tuple	#x0c)
+(define pst-int64	#x0d)
 
 ;; u8, s8, u16, s16, u32, s32, u64, s64, f32, f64, c32, c64
 ;;
--- a/gnuradio-runtime/lib/pmt/pmt_serialize.cc
+++ b/gnuradio-runtime/lib/pmt/pmt_serialize.cc
@@ -285,14 +285,16 @@ serialize(pmt_t obj, std::streambuf &sb)
     }
     else {
       if(is_integer(obj)) {
-	long i = to_long(obj);
-	if(sizeof(long) > 4) {
-	  if(i < -2147483647 || i > 2147483647)
-	    throw notimplemented("pmt::serialize (64-bit integers)", obj);
-	}
-	ok = serialize_untagged_u8(PST_INT32, sb);
-	ok &= serialize_untagged_u32(i, sb);
-	return ok;
+        long i = to_long(obj);
+        if((sizeof(long) > 4) && ((i < -2147483647 || i > 2147483647))) {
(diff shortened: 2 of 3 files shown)


marcusmueller:
I like this! Especially the fact that it comes with unit tests :)

I'll merge this, but let's discuss whether we should deprecate `to_long`, as it's pretty stupid for a serialization library to have different return types (32 vs 64 bit ints) on different machines for the same method.

Would you agree with the following statement:

> `from_long` should be instantly deprecated, as it deals differently with input, based on the platform you're on. In a serialization library, the serialized object should be bitwise identical for the same constructing function call, no matter what platform the user is on.

giladbeeri:
@marcusmueller I'd agree, I think a serializer should be platform-agnostic.

storborg:
I definitely agree with that statement and think that both ``from_long`` and ``to_long`` should be deprecated.

The future rules for PMT could look something like this (which nearly maps to existing behavior, but is more specific):

- Have both type-explicit functions like ``from_uint64`` and more general functions like ``from_integer``.
- General ``from_integer`` and ``from_real`` functions should always serialize to the smallest type which is still sufficient to represent the passed value.
- Type-explicit ``to_x`` functions would attempt to coerce types, and throw exceptions if the target datatype can't represent the stored value.

E.g.
```
# Python
foo = pmt.from_integer(2**60)  # gets serialized to a uint64
bar = pmt.from_integer(2**17)  # gets serialized to a uint32

# C++
int64_t val1 = pmt::to_int64(foo)
int32_t val2 = pmt::to_int32(bar)
```

That said, in my really ideal world, PMT would be dropped entirely in favor of [msgpack](https://msgpack.org/).

noc0lour:
I'm all in favour for mspack.

noc0lour:
@marcusmueller you promised am merge, you want to do it?

michaelld:
Nice PR! I really like the ideas here. Should the code internally still use `from_long` and `to_long`? Seems like if we're deprecating it, we shouldn't be using it!

storborg:
I think we'd have to add e.g. ``from_int64`` and ``to_int64`` in order to stop using ``from_long`` and ``to_long``, and at that point we might as well do a more thorough reinventing.

michaelld:
Yeah OK. So the code changes don't deprecate anything -- doing that was just mentioned here in the comments. My vote would be to get this code merged as-is to get the `int64` change & then fix things to actually be able to mark `to_long` and `from_long` as deprecated.

noc0lour:
@storborg just rebased your branch and pushed on my remote. Since the "generation" code was removed I had to add the `PST_INT64` in the corresponding header.

PR #1796, opened 2018-05-24
A couple small tweaks

michaelld:
2 almost trivial tweaks that I came upon recently.


marcusmueller:
While I like it when my compiler is silent: I'm not happy about the commented out `mask` parameter name; it breaks autocompletion in IDEs, it will omit the parameter name in ctags… Let's not be subjects of our compilers, but true masters, as it was meant to be.

Would

```
void function(vector<int>& mask) {
    (void)mask;
    throw...
} 
```
do the trick of silencing your compiler?

michaelld:
I see your point. yes that change works. I'll update the PR.

marcusmueller:
Thanks!

On 26 May 2018 13:12:23 GMT+02:00, Michael Dickens <notifications@github.com> wrote:
>I see your point. yes that change works. I'll update the PR.
>
>-- 
>You are receiving this because you commented.
>Reply to this email directly or view it on GitHub:
>https://github.com/gnuradio/gnuradio/pull/1796#issuecomment-392254436

-- 
Diese Nachricht wurde von meinem Android-Gerät mit K-9 Mail gesendet.

michaelld:
done!

PR #1797, opened 2018-05-28
Modtool: Shift CLI from Argparse to Click for Plugin Architecture and External Plugin Support

swap-nil7:
The following changes to the codebase are covered in this Pull Request:-
* Shift of CLI parsing from `Argparse` to `Click`
* A plug-in architecture for gr_modtool for ease of adding in-tree modules
* Support external plug-ins for extending gr_modtool with user-specific commands.
* Add simplicity where every module works independently of others and the CLI for a particular feature is 
  managed by its own module.

Added Dependencies: `click`, `click-plugins`


gr-utils/python/modtool/modtool_add.py
```diff
@@ -26,17 +26,20 @@
 
 import os
 import re
+import click
+import sys
```
primercuervo: PyLint will most probably complain about this import order. Which means that you could still probably lint all the files you modified.

gr-utils/python/modtool/modtool_makexml.py
```diff
@@ -176,3 +169,21 @@ def _get_blockdata(fname_cc):
...
         return (parser.read_params(), parser.read_io_signature(), blockname)
 
+
+### COMMAND LINE INTERFACE ###
+@click.command('makexml', short_help=ModToolMakeXML().description)
+@ModTool.common_params
+@ModTool.block_name
+def cli(**kwargs):
+    """
+    \b
```
primercuervo: This is probably just formatting but I miss the point of this backspace here.

gr-utils/python/modtool/util_functions.py
```diff
@@ -136,3 +138,29 @@ def ask_yes_no(question, default):
...
+    """ A simple completer function wrapper to be used with readline, e.g.
+    option_iterable = ("search", "seek", "destroy")
+    readline.set_completer(SequenceCompleter(option_iterable).completefunc)
+    """
+
+    def __init__(self, sequence):
+        self._seq = sequence
+        self._tmp_matches = []
+
+    def completefunc(self, text, state):
```
primercuervo: Missing docstrings

gr-utils/python/modtool/modtool_add.py
```diff
@@ -47,43 +50,26 @@ def __init__(self):
...
+                    self._info['blocktype'] = input("Enter block type: ")
+                    if self._info['blocktype'] not in self._block_types:
+                        print ('Must be one of ' + str(self._block_types))
+
         # Allow user to specify language interactively if not set
         self._info['lang'] = options.lang
         if self._info['lang'] is None:
-            while self._info['lang'] not in ['c++', 'cpp', 'python']:
-                self._info['lang'] = input("Language (python/cpp): ")
+            language_candidates = ('cpp', 'python')
```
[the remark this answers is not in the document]
swap-nil7: @primercuervo: I did not remove that intentionally. I just cherry-picked the PR and applied the changes for py3k compatibility. I have amended the same in the other PR.
Moreover, `c++` and `cpp` as different candidates for autocompletion does not make much sense when user types `c` `tab` `tab`

gr-utils/python/modtool/modtool_base.py
```diff
@@ -25,17 +25,79 @@
...
+        self.cmd_folder = os.path.abspath(os.path.dirname(__file__))
+
+    def add_command(self, cmd, name=None):
+        name = name or cmd.name
+        if name is None:
+            raise TypeError('Command has no name.')
+        _check_multicommand(self, name, cmd, register=True)
+        self.commands[name] = cmd
+
+    def command(self, *args, **kwargs):
```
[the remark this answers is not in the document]
swap-nil7: @primercuervo: This is basically an extension of the class `click.Group`. Actually, we can add an entire group of commands or simply a command to the present `gr_modtool`.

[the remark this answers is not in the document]
swap-nil7: @primercuervo: I have removed those methods. They will be inherited from the parent class themselves. Thanks for pointing them out.

gr-utils/python/modtool/modtool_add.py
swap-nil7: @primercuervo: Yes, this is incorrect. The standard imports should be placed first. Corrected.

gr-utils/python/modtool/modtool_add.py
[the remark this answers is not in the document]
swap-nil7: @primercuervo: I have added `c++` as a candidate.

gr-utils/python/modtool/modtool_base.py
```diff
@@ -25,17 +25,79 @@
 from __future__ import unicode_literals
 
 import os
+import sys
 import re
-from argparse import ArgumentParser, RawDescriptionHelpFormatter
+import click
+import functools
+from click import *
```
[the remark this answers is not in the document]
swap-nil7: @primercuervo: Yes, that's not necessary. Removed the wildcard import statement.

gr-utils/python/modtool/modtool_base.py
```diff
@@ -25,17 +25,79 @@
...
+from importlib import import_module
+from pkg_resources import iter_entry_points
+from click_plugins import with_plugins
 
 from gnuradio import gr
 from .util_functions import get_modname
 from .scm import SCMRepoFactory
 
+
+class CommandCLI(click.Group):
```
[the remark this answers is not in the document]
swap-nil7: @primercuervo: Added the document strings to the `class` as well as `methods`.

gr-utils/python/modtool/modtool_add.py
```diff
@@ -47,43 +50,25 @@ def __init__(self):
...
         # Allow user to specify language interactively if not set
         self._info['lang'] = options.lang
         if self._info['lang'] is None:
-            while self._info['lang'] not in ['c++', 'cpp', 'python']:
-                self._info['lang'] = input("Language (python/cpp): ")
+            language_candidates = ('cpp', 'python', 'c++')
+            with SequenceCompleter(language_candidates):
+                while self._info['lang'] not in language_candidates:
+                    self._info['lang'] = input("Language (python/cpp): ")
         if self._info['lang'] == 'c++':
```
primercuervo: I assume that now that 'c++' is in 'language_candidates', these lines can be removed. Right?

gr-utils/python/modtool/util_functions.py
primercuervo: That doesn't move aside the fact that having the docstrings would be nice :slightly_smiling_face:

gr-utils/python/modtool/modtool_add.py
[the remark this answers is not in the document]
primercuervo: One can argue if it makes little sense as it kinda depends on how the individual likes to refer to this language. The point is to make it diverse without losing generality.

gr-utils/python/modtool/modtool_add.py
swap-nil7: @primercuervo: I am not sure, how?

primercuervo: No, forget it. I think I confused the purpose of these lines. There is no change needed here

primercuervo:
@skoslowski could you please have a look at this PR?

gr-utils/python/modtool/modtool_rename.py
```diff
@@ -194,3 +189,23 @@ def _run_file_replace(self, filename, old, new):
...
+
+
+### COMMAND LINE INTERFACE ###
+@click.command('rename', short_help=ModToolRename().description)
+@ModTool.common_params
+@ModTool.block_name
+@click.argument('new-name', metavar="NEW-BLOCK-NAME", nargs=1, required=False)
+def cli(**kwargs):
+    """
+    \b
```
skoslowski: ?

gr-utils/python/modtool/modtool_rm.py
```diff
@@ -169,3 +166,17 @@ def _run_subdir(self, path, globs, makefile_vars, cmakeedit_func=None):
...
+@ModTool.common_params
+@ModTool.block_name
+def cli(**kwargs):
+    """ Remove block (delete files and remove Makefile entries) """
+    args = DictToObject(kwargs)
+    try:
+        ModToolRemove().run(args)
+    except ModToolException as err:
+        print(err, file=sys.stderr)
+        exit(1)
```
skoslowski: this part is really repetitive, consider moving it to base as part of the core/cli spilt

gr-utils/python/modtool/modtool_base.py
```diff
@@ -52,40 +100,32 @@ def __init__(self):
...
-                help="Don't do anything in the python/ subdirectory.")
-        parser.add_argument("--skip-grc", action="store_true",
-                help="Don't do anything in the grc/ subdirectory.")
-        parser.add_argument("--scm-mode", choices=('yes', 'no', 'auto'),
-                default=gr.prefs().get_string('modtool', 'scm_mode', 'no'),
-                help="Use source control management [ yes | no | auto ]).")
-        parser.add_argument("-y", "--yes", action="store_true",
-                help="Answer all questions with 'yes'. This can overwrite and delete your files, so be careful.")
-        return parser
+    def common_params(func):
```
skoslowski: I don't see a reason why this has to be a staticmethod, lets make a function, like the other CLIs

gr-utils/python/modtool/modtool_newmod.py
```diff
@@ -102,3 +98,23 @@ def run(self, options):
...
+def cli(**kwargs):
+    """
+    \b
+    Create a new out-of-tree module
+
+    The argument MODULE-NAME overrides the current module's name (normally is autodetected).
+    """
+    args = DictToObject(kwargs)
+    try:
+        ModToolNewModule().run(args)
```
skoslowski: just as a heads-up, the core-classes will need to take the actual params instead of a namespace object to make their use as API classes easier

mbr0wn:
To be clear, I haven't even read the code. I stopped after 'added dependencies'.

gr-utils/python/modtool/modtool_rename.py
swap-nil7: @skoslowski: Actually, `\b` is used to prevent rewrapping in a click command.
Documentation Link: http://click.pocoo.org/5/documentation/#preventing-rewrapping

skoslowski:
@mbr0wn, I agree with you to some extend. The only reason this was even considered was a go-ahead by @marcusmueller beforehand. So you guys hash this out...

noc0lour:
Really, imho GNU Radio does not have a problem with the amount of dependencies. Most of the time the quailty of dependencies is the bigger problem.

marcusmueller:
@mbr0wn: yes, we can :)

I've discussed this in the context of `gr_modtool` before: we can easily add Python dependencies without upsetting the maintainer to it. `gr_modtool` is not a runtime dependency, and as long as it's sane to install on somewhat modern systems and not GB in size, and we're not adding dead libraries, there's nothing stopping us.

Maintainer perspective: I actually consider GNU Radio est omnis divisa in partes tres:

* GNU Radio librarium runtimis, which is basically all the .so and Python runtime and block API, among things like gnuradio-config-info
* GNU Radio capita developmentis, which contains everything you need to link your source code against GR (`*.h`, pkg-config, CMake stuff)
* GNU Radio tools, which contains the stuff you'd want to have if you're *developing* GNU Radio applications. That means GRC, modtool, the funky gr_plot_stuff scripts (which could use some overhauling, too, and replacing Matplotlib with something even remotely well-performing would be pure winnage)

For the first category, clearly, don't add dependencies. In fact, we should be making it easy to even further compartimentalize that down, to avoid the whole "I installed GNU Radio run time, but that includes gr-qtgui, and now I've got a desktop environment on my headless number cruncher" business

The third category can be handled with much larger leniency when it comes to restricting  dependency influx: These things will happen on workstation PCs, anyway, so having a dependency on python GTK / gobject bindings doesn't hurt me (even if GRC is the only thing in the GR ecosystem to use GTK), nor does having a couple kB of Python library that is both easy to install under non-optimal conditions (I don't like anaconda / pip prefixes nearly as much as @skoslowski does – we simply don't agree on how great a concept that is) and natively on systems that have something resembling a package manager.

Obviously, you raise a fair point here: I can't be indiscriminately merging PRs that introduce new dependencies until I make that distinction easier.

This requires two things:

* Talking to our Debian and Gentoo and Windows package maintainers
* Taking GNU Radio packaging out of the hand of the current Fedora maintainer and redoing it

before we release 3.8.

swap-nil7:

Issue #959, opened 2016-09-16
skip head block should drop corresponding tags


mbr0wn:
Currently it appears that tags are passed through skip head unmodified, whereas the first N should be dropped along with samples and the remaining offsets should be shifted back by -N

pwicks86:
Made PR for this https://github.com/gnuradio/gnuradio/pull/1804.

devnulling:
@mbr0wn @marcusmueller can this be closed out?

mbr0wn:
Yeah.

PR #1800, opened 2018-05-29
Implement fix for gnuradio bug

pwicks86:
This fixes bug https://github.com/gnuradio/gnuradio/issues/959

Once this gets approved here, I'll open a PR upstream too.


pwicks86:
Oops, meant to open this against our local copy first.

PR #1804, opened 2018-05-31
Correctly handle tag offsets in the skiphead block

pwicks86:
This PR addresses https://github.com/gnuradio/gnuradio/issues/959. In the current skiphead block, tags have incorrect offsets, this PR adjusts the offsets of all tags that pass through by the number of items skipped to ensure that tags stay correctly aligned.

### DIFF
--- a/gr-blocks/lib/skiphead_impl.cc
+++ b/gr-blocks/lib/skiphead_impl.cc
@@ -44,6 +44,12 @@ namespace gr {
                  io_signature::make(1, 1, itemsize)),
         d_nitems_to_skip(nitems_to_skip), d_nitems(0)
     {
+      // Reserve space for a few tags to avoid constant re-allocation
+      // in the call to get_tags_in_window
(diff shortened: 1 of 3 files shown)


pwicks86:
Another issue to consider with this block is that some tags should probably be handled specially. For example, it would make sense if the `rx_freq` tags emitted by gr-uhd were shifted forward to the first sample emitted by skiphead if they occur during the skipped period so that downstream blocks can discover any parameter changes in the stream. Currently, I'm planning to build a custom variant of this block to handle this in [gr-sigmf](https://github.com/skysafe/gr-sigmf), but if there's interest, much of that work could probably be added to this skiphead block or a seperate skiphead block that lives in gr-uhd.

marcusmueller:
Fully agree, and I especially like the fact that you've also added QA!

marcusmueller:
@bhilburn can you work your CLA magic here?

PR #1806, opened 2018-06-02
C++98 (surprisingly) doesn't have <cstdint>, cond. #include

marcusmueller:
This was a mistake when cherry-picking a fix from master (which is C++11).


noc0lour:
Looks good, the failing unit test is bogus ...

PR #1810, opened 2018-06-05
Ensure that stdint is found in float_array_to_int

horvathp:


dkozel:
@horvathp Thanks for the patch. The problem is indeed that C++11 is not being forced yet on the master branch and so cstdint can't be guaranteed to exist. @marcusmueller we can either merge this or merge a change to the root CMakeLists file to require C++11.

marcusmueller:
After I thought I could sort this out over the last few days by the sheer means of progress (progressing to next being merged into master, that is), I will prefer to merge this _now_ and do the C++11 switchover later. I'll have to announce we're moving away from C++98 later.

PR #1815, opened 2018-06-08
grc_xml_generator: lxml.etree.tostring returns bytes instead of string

swap-nil7:
The outfiles must be binary written in python3.x

### DIFF
--- a/gr-utils/python/modtool/grc_xml_generator.py
+++ b/gr-utils/python/modtool/grc_xml_generator.py
@@ -52,14 +52,14 @@ class GRCXMLGenerator(object):
     def _lxml_prettyprint(self):
         """ XML pretty printer using lxml """
         return lxml.etree.tostring(
-                   lxml.etree.fromstring(ET.tostring(self.root, encoding="UTF-8")),
-                   pretty_print=True
+                   lxml.etree.fromstring(ET.tostring(self.root, encoding='unicode')),
+                   pretty_print=True, encoding='unicode'
                )
 
     def _manual_prettyprint(self):
         """ XML pretty printer using xml_indent """
         xml_indent(self.root)
-        return ET.tostring(self.root, encoding="UTF-8")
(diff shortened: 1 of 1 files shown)


marcusmueller:
I'm surprised. I wasn't aware of any change in behaviour with respect to open(,"wb") in py3k. Also, the "b" traditionally doesn't do anything on Linux, just on windows. How did you come across this, and what does it fix?

marcusmueller:
Interesting! But looking at the etree.tostring:

```python
def tostring(element, encoding=None, method=None, *,
             short_empty_elements=True):
    """Generate string representation of XML element.

    All subelements are included.  If encoding is "unicode", a string
    is returned. Otherwise a bytestring is returned.

    *element* is an Element instance, *encoding* is an optional output
    encoding defaulting to US-ASCII, *method* is an optional output which can
    be one of "xml" (default), "html", "text" or "c14n".

    Returns an (optionally) encoded string containing the XML data.

    """
    stream = io.StringIO() if encoding == 'unicode' else io.BytesIO()
    ElementTree(element).write(stream, encoding, method=method,
                               short_empty_elements=short_empty_elements)
    return stream.getvalue()
```

shouldn't you just be using `encoding="unicode"` instead of `"UTF-8"` instead of then re-decoding the bytes?

swap-nil7:
Actually, I tried for the version independent compatibility. But I guess, `encoding = "unicode"` will be better for the python3 branch.

marcusmueller:
oh wait, I was about to merge this, but:

do we still need the xml generator in next now that that's py3+yaml?

swap-nil7:
I have replaced XML generator by YAML generator in another PR but if we wish we can just add the XML generating files and they'll work as usual due to the plug-in architecture. 
Currently, imho we don't really need to merge this PR if we don't need XML generator.
Thanks for the approval though. :)

swap-nil7:
Closing this since the XML generation is not required.

PR #1816, opened 2018-06-08
gr-newmod: update the directory parameter for user

swap-nil7:
As a user, the `directory` parameter `-d` should ask for the directory in which the new module directory will be created rather than the entire directory name including the new module name.

I mean the command `gr_modtool -d /home/swapnil/Documents/modtool/gr-sqr_ff newmod sqr_ff` should change to `gr_modtool -d /home/swapnil/Documents/modtool newmod sqr_ff`.

### DIFF
--- a/gr-utils/python/modtool/modtool_newmod.py
+++ b/gr-utils/python/modtool/modtool_newmod.py
@@ -44,7 +44,7 @@ class ModToolNewModule(ModTool):
         parser.add_argument("--srcdir",
                 help="Source directory for the module template.")
         parser.add_argument("module_name", metavar='MODULE-NAME', nargs='?',
-                help="Override the current module's name (normally is autodetected).")
+                help="Name of the module to be added.")
 
     def setup(self, options):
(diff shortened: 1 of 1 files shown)


swap-nil7:
The current `help` page is slightly misleading. It says `Override the current module's name (normally is autodetected).` as the help for positional argument `MODULE-NAME` even though nothing is overridden.

gr-utils/python/modtool/modtool_newmod.py
```diff
@@ -57,8 +57,7 @@ def setup(self, options):
         if not re.match('[a-zA-Z0-9_]+$', self._info['modname']):
             raise ModToolException('Invalid module name.')
         self._dir = options.directory
-        if self._dir == '.':
-            self._dir = './gr-%s' % self._info['modname']
+        self._dir = self._dir + '/gr-{}'.format(self._info['modname'])
```
mbr0wn: Prefer `os.path.join()`

swap-nil7:

Issue #1817, opened 2018-06-09
fmdet_cf: wrong coefficient, totally insufficient testing


marcusmueller:
* As Eugene now brought up the third time, we still haven't fixed fmdet_cf (working on that)
* There's no end-to-end test of FM modulated signals getting decoded correctly in a zero-noise scenario. That's the least we should do.

Solve in two separate PRs. Backport fix to maint-3.7.

PR #1819, opened 2018-06-09
Fixed wrong coefficient in Taylor series; cleanup (3.7 backport)

marcusmueller:
To even edit this file, indents had to be sanitized a bit. Also,
commented-out code was removed. Also, 8.f is a constand and not really
longer than `d_8`.

This is the first in two changes necessary for #1817. The other would be
adding a unit test that is *slightly* more holistic than just doing a
single-tone frequency detection.

Note that there's much about this code that could be refactored,
starting with Sdot being a FIR-filtered version of the input.


marcusmueller:
Tried the github merge conflict resolution tool. Not convinced. Will do manually.

PR #1827, opened 2018-06-14
ModToolRename: Remove erroneous lines

swap-nil7:
I am not sure why these lines are included in `modtool_rename.py`. Maybe I am missing something, but there is no option `lang` for the command `rename`. So, this command raises `KeyError` when passed with flags like `skip-lib`.
I guess this portion should be removed.


mbr0wn:
Probably copy pasta when creating the module.

swap-nil7:

PR #1832, opened 2018-06-20
tests: fixed python template - import error and style

giladbeeri:


dkozel:
@swap-nil7 Can you please take a look at this pull request and let us know if your changes to the modtool already have these changes or conflict with them?

gr-utils/python/modtool/templates.py
```diff
@@ -494,22 +494,22 @@ class qa_${blockname} : public CppUnit::TestCase
 from gnuradio import gr, gr_unittest
 from gnuradio import blocks
 #if $lang == 'cpp'
-import ${modname}_swig as ${modname}
+import ${modname}.${modname}_swig as ${modname}
```
mbr0wn: Is that really correct? Does it make some assumptions on Python version or something like that?

giladbeeri: With Python 2.7:

Before change: `ImportError: No module named mymodule_swig`
After change: works.

With Python 3:
I tested the general import syntax. It seems that Python3 will require `import ${modname}` just before, otherwise, it will fail with an `ImportError: no module named ...`.
I can add it because it doesn't hurt Python 2 but solves 3. Please tell me what you think.

By the way, I imported the module since this seems to be the original author's intention. In my own tests, I import the block directly: `from mymodule.mymodule_swig import myblock`

swap-nil7: I guess `import .${modname}_swig as ${modname}` will work for python3. 
I guess we should also change `from ${blockname} import ${blockname}` to `from .${blockname} import ${blockname}` and all the usual imports in the `__init__.py`.

giladbeeri: That doesn't work with Python 2.7.

What are the current guidelines about Python 2.7 / Python 3 support on master?

swap-nil7: I really don't understand why the current code does not work for py2k. The installation of swig file is in the same directory as that of the py_module.

noc0lour: Relative imports are supported in python2 starting with Python 2.5, right? https://docs.python.org/2/tutorial/modules.html#intra-package-references

noc0lour: I guess if you do `import ${modname}.${modname}_swig as ${modname}` it's possible you're pulling in the globally installed python module and its swig.

giladbeeri: When I cd to the project directory and run `python python/qa_my_module.py`,  "it doesn't have its usual name, but has "__main__" as its name instead. So relative imports don't work." (quoted from https://stackoverflow.com/a/11537218/3285914)

giladbeeri: To sum it up:

1. Tests are executed directly, they don't support relative imports.
1. The context is a C++ block with SWIG, so building and installing the project is required.
1. The package is taken from the prefix installation path.
1. The current code doesn't work because the installation put the swig file under `site-packages/${modname}`, so `import ${modname}_swig as ${modname}` doesn't work but `import ${modname}.${modname}_swig as ${modname}` does.

giladbeeri: Indeed, is it the only official way to run tests (vs. executing the python files)?
If so, I think the `if __name__ == '__main__'` part should be removed from the template.

noc0lour: @giladbeeri 
The direct invocation of test scripts is a good point. I actually did this, in the `python` dir of some modules. Though it has a serious drawback: None of the PYTHONPATH magic is applied. We need to check if tests are run in `ctest` as script. If that's the case we can't really use relative imports, since this only makes sense in the package context. 
It looks like my previous comment only applies to modules which are contained in a package.

giladbeeri: Just tested the different variations with `make test`:

1. Original `import ${modname}_swig as ${modname}` works 
1. Suggested relative import using the dot syntax doesn't work
1. `import ${modname}.${modname}_swig as ${modname}` works (but reads files from the install location).

So it seems that from the suggestions so far, only the last one works both with `make test` and direction execution, but has the downside of requiring `make install`.

swap-nil7: I am totally surprised with relative imports not working.

giladbeeri: Read a bit, relative imports are supported only in the format `from . import a`, not `import .a as b` as was suggested above. `import X` is always absolute.

swap-nil7: Yes, that's right. :)

PR #1835, opened 2018-06-21
uhd: qa: Improve time_spec tests

mbr0wn:
The time spec test could fail if Python cached objects in a certain way.
This also adds some more small tests.

### DIFF
--- a/gr-uhd/python/uhd/qa_uhd.py
+++ b/gr-uhd/python/uhd/qa_uhd.py
@@ -38,10 +38,12 @@ class test_uhd(gr_unittest.TestCase):
     def test_time_spec_t (self):
         seconds = 42.0
         time = uhd.time_spec_t(seconds)
-        twice_time = time + time;
-        zero_time = time - time;
-        self.assertEqual(time.get_real_secs() * 2,  seconds * 2 )
-        self.assertEqual(time.get_real_secs() - time.get_real_secs() , 0.0)
+        self.assertEqual(time.get_real_secs() * 2, seconds * 2)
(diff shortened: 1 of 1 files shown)


noc0lour:
The test in general is there to verify if addition/subtraction with `uhd.time_spec_t` works?

marcusmueller:
@mbr0wn I must admit I don't understand the failure mode we're avoiding here. The PR is fine; I'm just worried I'd run into the same problem as the formerly failing test

dkozel:
The first test is a simple check on the constructor and get_real_secs, the second set of tests are for the sub and add operators of the time_spec.

mbr0wn:
The failure mode I was seeing that after the addition, the original `time` variable had a different value. Not sure where the bug was, because I've only seen this once. However, I find this more readable. Also, I just realized there are still ; in this code... let me remove those.

marcusmueller:
(I'm kinda persistent on this one, because I remember manually fiddling with the SWIG surrounding UHD types in GR, and I'm afraid I've broken that a while back myself!)

marcusmueller:
OK, the unit test in the previous shape *does* detect an actual bug.

Point being the following: 

It's standards-conforming that in gr-uhd's swig the custom `__add__` function does a *copy elision* here:

```
time_spec_t temp = *self;
```

which means that `temp` __is__ `self` afterwards, and we're modifying that. That's a true bug!

So, no, let's not change that unit test. Instead, let's fix the `__add__` implementation. I believe @noc0lour has an approach to reliably avoid copy elusion?

mbr0wn:
OK you're right. Closing this.

Issue #1842, opened 2018-06-23
python3_fix -> python3 outstanding brokenness


marcusmueller:
- [ ] bin_statistics: message.i: to_string doesn't give Python3 compatible string. This comes from `msg_queue`. 
- [x] qa_pmt unit test segfault && pmt.deserialize_string: PY3 string vs std::string conflict (it's the same issue, see #1852)
- [x] test_cpp_py_binding: test_001: seemingly a null deref? `RPC_get_string_get`
- [x] -"- `import exceptions` fails (in many places)
- [x] same for cpp_py_binding_set
- [x] `vector_source_b_make`: in python3, all our seemingly magic wrapping of numpy to std::vector breaks with swig. https://github.com/swig/swig/issues/888 we'll probably have to do our own typemapping or so. (Wrapping of python tuple/list has to be verified to be compatible with std::vector)
- [x] `qa_mpsk_snr_est` does fail with wrong out/input values. might simply be a bug that didn't show in py2land
- [x] same `qa_scrambler.py`. Is it Py3 (std::vector wrappage as above?) or a scrambler bug?
- [x] python `video_sdl` not importable.
- [ ] Check QtGUI blocks Label/Line/Color configurations in `yml` definitions @haakov 
- [ ] `if not label` in QtGUI YAML definition throws: @haakov 
- [x] GRC doesn't work at all

noc0lour:
Volk compile issue: https://github.com/gnuradio/volk/issues/180

mbr0wn:
Suggestion for qa_pmt: https://github.com/gnuradio/gnuradio/pull/1852

This might help with the polar codes: http://scipy-cookbook.readthedocs.io/items/SWIG_NumPy_examples.html

mbr0wn:
Scrambler was a simple mistake in making the QA code Py3k compatible. Fix here: #1853

mbr0wn:
These are unit tests that fail and not yet have a PR to fix them:
```
The following tests FAILED:
        182 - qa_bin_statistics (Failed)
        196 - qa_hier_block2 (Failed)
        214 - qa_file_metadata (Failed)
        221 - qa_polar_decoder_sc_list (Failed)
        222 - qa_fecapi_dummy (Failed)
        224 - qa_polar_decoder_sc (Failed)
        225 - qa_polar_encoder_systematic (Failed)
        226 - qa_polar_decoder_sc_systematic (Failed)
        229 - qa_polar_encoder (Failed)
[...]

PR #1843, opened 2018-06-23
Merge Python3

noc0lour:


noc0lour:
These build errors are new though...

marcusmueller:
Andrej and I sat together today and did the following
1. Merge Andrej's patchset necessary to make the python3 branch compile successfully and start GRC into the python3 branch
2. Merge python3 into next

Now, 1. entails the YAML conversion. 2. is now manifest in https://github.com/gnuradio/gnuradio/pull/1843 ; I'll merge that into next tomorrow or Monday, depending on level of objection :slightly_smiling_face:
We've only tested with Py3, but so far we've been able to identify the regressions documented in https://github.com/gnuradio/gnuradio/issues/1842

The worst issues before next works "about as well" as master are, as far as I can see right now:

* SWIG + Python3 doesn't support automatic wrapping of Python Lists nor numpy.array into std::vector. That basically breaks `vector_source_*` in most applications, as far as I can see. Tuple of numbers seems to work
* we've got a serious architectural problem with methods that return a `std::string` but mean to return an array of chars, not a (textual) string. In Python3, the string type `str` represents text, abstracting the encoding to bytes, and `bytes` doesn't represent text. We can't set SWIG to wrap std::string to one or the other "target type" without breaking a solid part of our usage. The truth is we shouldn't be using `std::string` in C++ for things that aren't essentially text, but it's how it is, and we've got the breakage
* GRC starts, but is largely dysfunctional under Py3 (edited)

mbr0wn:
Failed with SWIG 2.0.12.

PR #1844, opened 2018-06-24
Remove CppUnit dependency

mbr0wn:


mbr0wn:
And now for the killing blow! CppUnit is now gone.

mbr0wn:
@marcusmueller I've also removed the hack that enabled C++11 for master when we have a newer CppUnit. However, I would suggest that we start enabling C++11 on master all the time.

marcusmueller:
Fully Agreed. We shall have C++11 on master.

marcusmueller:
heyyyy you smuggled a bugfix in there, didn't you ;) ! Would the missing FindPkgConfig Cmake statements be something I'd want to cherry-pick onto maint-3.7?

mbr0wn:
@marcusmueller Good question, I think on maint the bug will never occur because the `find_package()` call for CppUnit always includes the other module. It's up to you, it's definitely more accurate with the bugfix.

PR #1847, opened 2018-06-25
ModTool: Proper check on blockname before renaming

### LINKED ISSUE #1727: gr_modtool rename: Renaming of pattern instead of entire blockname
While using ```gr_modtool rename``` the script searches for instances of the same pattern in all the file names in different directories instead of searching for the exact block name. For example:-
```
Enter name of block/code to rename (without module name prefix): mul
Block/code to rename identifier: mul
Enter name of block/code (without module name prefix): sqr
```
will also rename the files corresponding to block name ```multiplexer``` which is undesirable!

swap-nil7:
While renaming the `blockname`, there must be a valid check on the existence of `blockname`. This can be done by traversing the `lib` and `python` directories, getting a list of all `blocknames` and checking if this `blockname` exists. If it does not exists, we can give the user a list of `suggested alternatives` for his ease.
Moreover, this same list of blocknames can be beneficial for the `Sequence Completer`. 
This method of generating the list of `blocknames` is more efficient than the one existing on the `master` branch script `modtool_rm.py` as it is python3 compatible. `filter` returns an object in python 3.
Fixes #1727

### DIFF
--- a/gr-utils/python/modtool/modtool_add.py
+++ b/gr-utils/python/modtool/modtool_add.py
@@ -27,11 +27,12 @@ from __future__ import unicode_literals
 import os
 import re
 
-from .util_functions import append_re_line_sequence, ask_yes_no, SequenceCompleter
 from .cmakefile_editor import CMakeFileEditor
 from .modtool_base import ModTool, ModToolException
 from .templates import Templates
 from .code_generator import render_template
+from .util_functions import append_re_line_sequence, ask_yes_no, SequenceCompleter
+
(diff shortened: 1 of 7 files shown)


swap-nil7:
And getting the first character of `options.newname` is not desirable.

noc0lour:
@swap-nil7 can you rebase locally on `origin/next`, force push on your PR branch and rebase this PR?

swap-nil7:

PR #1849, opened 2018-06-25
grc: Fixed the "Show Msg Ports" option

haakov:
6 of the blocks in the block tree have the ability to hide their message ports. Without this fix, the message ports are shown regardless of the "Show Msg Ports" option.


skoslowski:
bool-typed params have been around before I started working on GRC. Haven't seen any uses though. 

Anyway, not sure if options and option labels make sense in this case. And we choose to use real bool-params, it might make sense to use those switch-widgets instead of a drop-down.

If not, I believe parsing the enum value correctly is the better fix.

haakov:
Right!
So these, I suppose: 
Image of a switch widget

Yeah using that does makes sense

mbr0wn:
Merged manually.

PR #1851, opened 2018-06-25
blocks: Fix QA for file_sink on Py3k 

mbr0wn:


gr-blocks/python/blocks/qa_file_sink.py
```diff
@@ -47,11 +46,11 @@ def test_file_sink(self):
...
 
             # Check file length (float: 4 * nsamples)
-            st = os.stat(temp.name)
-            self.assertEqual(st.st_size, 4 * len(data))
+            file_size = os.stat(temp.name).st_size
+            self.assertEqual(file_size, 4 * len(data))
 
             # Check file contents
-            datafile = open(temp.name, 'r')
+            datafile = open(temp.name, 'rb')
```
mbr0wn: This is the actual fix.

mbr0wn:
Note: I did not test on Python 2.

noc0lour:
Tests on the F26 builders suggest it works for both, py2 and py3k (:

PR #1852, opened 2018-06-26
pmt: Serialize PMTs to byte strings in Python (and vice versa)

mbr0wn:


mbr0wn:
This is a suggestion for a workaround for the PMT serialization, and maybe we can use this elsewhere. I tried forcing a bytes string in Python 3, but failed. However, I could force a tuple and then convert that.

gnuradio-runtime/swig/pmt_swig.i
```diff
@@ -302,7 +319,21 @@ namespace pmt{
...
 } //namespace pmt
+
+%pythoncode %{
+  def serialize_str(pmt_obj):
+    import array
+    tup = _serialize_str_u8(pmt_obj)
+    return array.array('B', tup).tobytes()
+
+  def deserialize_str(pmt_str):
+    tup = tuple([x for x in pmt_str])
```
noc0lour: it should be possible to drop `[]` and do the list comprehension inside `tuple()`

noc0lour:
Looks like the py3k test `qa_pmt` works now, but fails on py2

gnuradio-runtime/swig/pmt_swig.i
mbr0wn: I don't get it. There is a list comprehension inside `tuple()`.

noc0lour: But you should be able to do this `tuple(x for x in pmt_str)` as well (:

mbr0wn: Oh you mean *don't* do a list comprehension :grin:

Yeah you're right; in Py3k it's almost muscle memory to use the [] to force unrolling of iterators, but that `tuple()` will unroll the iterator as well.

noc0lour: Ok, sorry, I meant generator expressions https://www.python.org/dev/peps/pep-0289/

gnuradio-runtime/python/pmt/qa_pmt.py
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env python
 #
-# Copyright 2011 Free Software Foundation, Inc.
+# Copyright 2011,2018 Free Software Foundation, Inc.
```
marcusmueller: Feels good man

gnuradio-runtime/swig/pmt_swig.i
```diff
@@ -302,7 +319,21 @@ namespace pmt{
...
+  std::vector<uint8_t> _serialize_str_u8(pmt_t obj);
+  pmt_t _deserialize_str_u8(std::vector<uint8_t> py_str);
 
 } //namespace pmt
+
+%pythoncode %{
+  def serialize_str(pmt_obj):
+    import array
+    tup = _serialize_str_u8(pmt_obj)
+    return array.array('B', tup).tobytes()
```
marcusmueller: I'm afraid `array.array.tobytes` is Py3-only.

the base problem here being that `bytes` in Py2 gives a string, whereas it's a separate type in Py3k. 
Hm. I don't have a portable solution myself :(

mbr0wn:
OK this one works for me both for Py2 and 3.

PR #1853, opened 2018-06-26
digital: Fix scrambler QA

mbr0wn:


gr-digital/python/digital/qa_scrambler.py
```diff
@@ -122,7 +122,7 @@ def test_additive_scrambler_tags_oneway(self):
         reset_tag3 = gr.tag_t()
         reset_tag3.key = pmt.string_to_symbol(reset_tag_key)
         reset_tag3.offset = 20
-        src = blocks.vector_source_b([s * 3 for s in src_data], False, 1, (reset_tag1, reset_tag2, reset_tag3))
+        src = blocks.vector_source_b(src_data * 3, False, 1, (reset_tag1, reset_tag2, reset_tag3))
```
marcusmueller: confused: the old code under python2 would take every element from the list of numbers in [0,10[, and multiply it by three, but the new code under Python2 would take the list and repeat it three times?

marcusmueller: ah well, that seems what is checked in the end, anyway. I was just confused by the fact that this unit test used to work – but I just realized we broke it ourselves just days ago. Thanks!

mbr0wn:
Even the old code would repeat the list three times (see master branch).

PR #1857, opened 2018-06-29
ModTool: Relative imports in python templates

swap-nil7:
In my opinion, imports should be relative in python scripts.

### DIFF
--- a/gr-utils/python/modtool/gr-newmod/swig/howto_swig.i
+++ b/gr-utils/python/modtool/gr-newmod/swig/howto_swig.i
@@ -2,7 +2,7 @@
 
 #define HOWTO_API
 
-%include "gnuradio.i"			// the common stuff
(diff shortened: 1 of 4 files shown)


gr-utils/python/modtool/gr-newmod/swig/howto_swig.i
```diff
@@ -2,7 +2,7 @@
 
 #define HOWTO_API
 
-%include "gnuradio.i"			// the common stuff
+%include "gnuradio.i"           // the common stuff
```
noc0lour: Was this accidental?

swap-nil7: No, I have replaced `tab spaces` which look not so good using `vim`.

swap-nil7:

PR #1858, opened 2018-06-30
gr-qtgui: Fail less noisily if QT5 is not installed

mbr0wn:
When QT5 is not installed, find_package() will also fail. That's normal.
In the current state however, it throws a huge scary warning.


noc0lour:
I'd like to drop Qt4 for `next` if possible. I don't know of any place where Qt5 is not available. RHEL seems like the last bastion still shipping Qt4 for their customers.

mbr0wn:
Agreed.

Issue #1738, opened 2018-04-18
message_strobe[_random]: change API and impl to actually only accept integer milliseconds


marcusmueller:
without #1733, we wouldn't have noticed that this block acts as if it can wait for float milliseconds, while it's actually only waiting for integer milliseconds.

Since I'd argue that on a PC, integer milliseconds is probably granular enough, I'd recommend just changing the signature from `float` to `long`; but I'm open for opinions.

marcusmueller:
closed via #1861

PR #1861, opened 2018-06-30
Unbreak boost 1.67. Constrain the delay in message_strobe

horvathp:
Apply the fixes contained in master to next in order to unbreak boost 1.67 by static_casting the delay values to long. As suggested before, change the API of message_strobe such that only integer milliseconds can be specified. 

I'm not sure whether constraining the parameters of the delay distributions in message_strobe_random to integer values would be a good idea. This PR leaves the delay distribution values as floats and only rounds the random delays to integer millisecond values.

### DIFF
--- a/gr-blocks/grc/blocks_message_strobe.block.yml
+++ b/gr-blocks/grc/blocks_message_strobe.block.yml
@@ -8,7 +8,7 @@ parameters:
     default: pmt.intern("TEST")
 -   id: period
     label: Period (ms)
-    dtype: real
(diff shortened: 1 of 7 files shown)


marcusmueller:
IIRC the point was that the underlying sleep function really takes integer timeunits, and that for some reasons the block's API pretends that's not the case. So, this closes https://github.com/gnuradio/gnuradio/issues/1738 .

marcusmueller:
Merged per 5d17aec.

PR #1862, opened 2018-06-30
digital: mpsk_snr_est: Fix QA for Py2/3 compat

mbr0wn:
The test was relying on randint() to return the same integers, but since
Python 3.2 it doesn't. Now, we create bits using random.random(), which
will create reproducible bits across Python versions.


noc0lour:
Is this a proactive merge to `master` to ease the merge?

mbr0wn:
@noc0lour Yes, exactly.

marcusmueller:
Since this mirrors #1863: merging it

PR #1863, opened 2018-06-30
digital: mpsk_snr_est: Fix QA for Py2/3 compat

mbr0wn:
The test was relying on randint() to return the same integers, but since
Python 3.2 it doesn't. Now, we create bits using random.random(), which
will create reproducible bits across Python versions.

### DIFF
--- a/gr-digital/python/digital/qa_mpsk_snr_est.py
+++ b/gr-digital/python/digital/qa_mpsk_snr_est.py
@@ -1,6 +1,6 @@
 #!/usr/bin/env python
 #
-# Copyright 2011-2013 Free Software Foundation, Inc.
+# Copyright 2011-2013,2018 Free Software Foundation, Inc.
 #
 # This file is part of GNU Radio
 #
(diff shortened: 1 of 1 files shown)


gr-digital/python/digital/qa_mpsk_snr_est.py
```diff
@@ -19,40 +19,44 @@
...
 class test_mpsk_snr_est(gr_unittest.TestCase):
     def setUp(self):
         self.tb = gr.top_block()
-
         random.seed(0) # make repeatable
         N = 10000
-        self._noise = [get_n_cplx() for i in range(N)]
-        self._bits = [get_cplx() for i in range(N)]
+        self._noise = [get_n_cplx() for _ in range(N)]
+        self._bits = [get_cplx() for _ in range(N)]
```
noc0lour: I like the Python style improvements (:

gr-digital/python/digital/qa_mpsk_snr_est.py
```diff
@@ -19,40 +19,44 @@
...
+""" Test digital.mpsk_snr_est_cc """
 
 import random
 
 from gnuradio import gr, gr_unittest, digital, blocks
 
+def random_bit():
+    """Create random bits using random() rather than randint(). The latter
+    changed for Python 3.2."""
+    return random.random() > .5
```
marcusmueller: My `import random` gives me a `random.randint()` on both Python2 and 3. Why the detour? You say "since 3.2, `randint` doesn't return the same integers. What do you mean with "not the same"?

marcusmueller: I've looked at the docs, and I see no trace of any change to randint. So, I really need clarification!

marcusmueller: Do you perhaps mean to do `version=1` when seeding, so that "older python number generatio compatibility" is restored? https://docs.python.org/3.6/library/random.html#random.seed

marcusmueller:
Wait, are we **relying** on a random number generator to always yield the same sequence, and that's bad, because nothing in the python docs says it will stay stable; it just says "it relies on MT". I'm not even sure I should be advertising using our own random generator here – even if we're in control of that, I don't think we'd want to define its values to be reliably the same "for ever".

So, I think we should not further go in this direction; it's unreliable, and there's nothing worse than a unit test depending on unreliable things. It's really not that much noise and bits we're generating: can't we just generate on Py2 once, and hard-code the values instead?

noc0lour:
So you are suggesting we generate a "standard noise" for QA and load it for every QA test?

mbr0wn:
From the manual (https://docs.python.org/3.5/library/random.html#random.randrange):

> Changed in version 3.2: randrange() is more sophisticated about producing equally distributed values. Formerly it used a style like int(random()*n) which could produce slightly uneven distributions.

mbr0wn:
@marcusmueller In general, the unit test is not a great test. It even says so:
```py
    # Test various SNR estimators; we're not using a Gaussian
    # noise source, so these estimates have no real meaning;
    # just a sanity check.
```




I would have written it differently: I would have used the *actual* SNR as the expected value, and then used the length of the test vector and the quality of the estimator to gauge a variance on the result. Then, as long as the random number generator is doing something halfway sensible, you would have a controllably low failure in QA due to statistical variance.

I still think my fix is OK for now. The current QA code relies on a certain sequence of random numbers, or your results are off. My fix makes sure the sequence of numbers doesn't change between Py2 and 3. And boom, another QA code back to its original state (questionable implementation, but it *does* catch regressions, and it'll pass on all versions). In any case, the new QA code is an improvement of the old.

gr-digital/python/digital/qa_mpsk_snr_est.py
mbr0wn: No I mean the behavioural change in `randrange()`. See my comment below.

mbr0wn: Basically, we can't use `randint()` or `randrange()` if we need the same numbers across Py2 and 3.

marcusmueller:
Yep, OK, convinced me: It's an improvement, and it fixes the issue for now. 

Still, yes, this is a bad unit test. Opening an issue to track that.
To address @noc0lour's exact question:

> So you are suggesting we generate a "standard noise" for QA and load it for every QA test?

Yes! A unit test **must not** be random. There should be zero things in it that depend on environment.
This can be implemented by saved random sequences, or by having a reliable sequence generator, a.k.a. PRNG.

A reliably working PRNG would just do exactly that. Give us the same noise, every time, on every platform, insensitive to any threading/timing/moon phase/number of geckos. The Python `random` thing broke that contract (but it really didn't, since the sequences generated by a PRNG are really just API, and when changing major version number, you can change whatever you want about that API).

Point being: This unit test depends on an assumed, but undocumented, assumption that `random` doesn't change. It changed. We still continue to depend on it. Fool me once...

PR #1870, opened 2018-07-04
Maint 3.7  Fix the fact that, depending on file order, block_tree.xml overrode categories in block files.

marcusmueller:
backport of #1869


mbr0wn:
This has been at "Don't Merge" for almost a year, closing. Can be reopened upon activity.

PR #1877, opened 2018-07-08
Adding a CMake option -DENABLE_GLIBCXX_ASSERTIONS

marcusmueller:
This allows setting of the _GLIBCXX_ASSERTIONS macro, which enables
"cheap" runtime checks for things like vector access bounds.

This is related to #1774.

It allows for implementation of https://github.com/gnuradio/gnuradio-buildbot/issues/6

Note that Fedora 28 introduced the aformentioned macro as set-by-default, so that we now see GNU Radio fail in production Fedora maintainer binary packages.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -300,6 +300,8 @@ endif(ENABLE_PERFORMANCE_COUNTERS)
 OPTION(ENABLE_STATIC_LIBS "Enable building of static libraries" OFF)
 message(STATUS "Building Static Libraries: ${ENABLE_STATIC_LIBS}")
 
+OPTION(ENABLE_GLIBCXX_ASSERTIONS "GCC/libstdc++ cheap assertions (_GLIBCXX_ASSERTIONS)" OFF)
+
 ########################################################################
 # Variables replaced when configuring the package config files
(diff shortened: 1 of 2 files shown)


noc0lour:
I don't see a reason to add a switch to the GR cmake since one could specify this `CMAKE_CXX_FLAGS` on the cmdline or set the `CXXFLAGS` enviroment correctly. That's how fedora is enabling those checks (probably) https://cmake.org/cmake/help/latest/variable/CMAKE_LANG_FLAGS.html#variable:CMAKE_%3CLANG%3E_FLAGS

marcusmueller:
Point being you can't use `CMAKE_CXX_FLAGS`, because that's *overriding*, not *extending* the build-type-specific compiler flags. (very useful feature, much wow)


`export CXXFLAGS=-D_GLIBCXX_FLAGS` does work, too, but I was assuming that was less elegant to buildbot. So if that's not the case, closing this.

marcusmueller:
Hm, well on the other hand, I *do* regularly want to be able to enable these things on a command line with a single switch. I find relying on environent variables for switches unsatisfactory :/

PR #1881, opened 2018-07-08
polar_sc: Fix vector out of bound accesses

marcusmueller:
Unlike most of the other #1774-triggered fixes, this actually was an
effect of reading from unitialized memory. I'm surprised this worked as
intended!

Needs content-wise review from @jdemel!

### DIFF
--- a/gr-fec/lib/polar_common.cc
+++ b/gr-fec/lib/polar_common.cc
@@ -108,14 +108,13 @@ namespace gr {
         d_volk_temp = (unsigned char*) volk_malloc(sizeof(unsigned char) * block_size(), volk_get_alignment());
         d_volk_frozen_bit_mask = (unsigned char*) volk_malloc(sizeof(unsigned char) * block_size(), volk_get_alignment());
         d_volk_frozen_bits = (unsigned char*) volk_malloc(sizeof(unsigned char) * nfrozen, volk_get_alignment());
-        for(int i = 0; i < nfrozen; i++){
-          d_volk_frozen_bits[i] = d_frozen_bit_values[i];
-        }
+        std::copy(d_frozen_bit_values.begin(), d_frozen_bit_values.end(), d_volk_frozen_bits);
+        std::fill(d_volk_frozen_bits + d_frozen_bit_values.size(), d_volk_frozen_bits + nfrozen, 0);
 
         int nfbit = 0;
         for(int i = 0; i < block_size(); i++){
           unsigned char m = 0x00;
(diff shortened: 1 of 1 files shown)


gr-fec/lib/polar_common.cc
```diff
@@ -108,14 +108,13 @@ namespace gr {
         d_volk_temp = (unsigned char*) volk_malloc(sizeof(unsigned char) * block_size(), volk_get_alignment());
         d_volk_frozen_bit_mask = (unsigned char*) volk_malloc(sizeof(unsigned char) * block_size(), volk_get_alignment());
         d_volk_frozen_bits = (unsigned char*) volk_malloc(sizeof(unsigned char) * nfrozen, volk_get_alignment());
-        for(int i = 0; i < nfrozen; i++){
-          d_volk_frozen_bits[i] = d_frozen_bit_values[i];
-        }
+        std::copy(d_frozen_bit_values.begin(), d_frozen_bit_values.end(), d_volk_frozen_bits);
```
jdemel: It takes a while to figure out, why these lines are a good idea. This just asks for a comment. `d_frozen_bit_values` might be an empty std::vector. Fill it with default values ('0').

gr-fec/lib/polar_common.cc
```diff
@@ -108,14 +108,13 @@ namespace gr {
...
+        std::fill(d_volk_frozen_bits + d_frozen_bit_values.size(), d_volk_frozen_bits + nfrozen, 0);
 
         int nfbit = 0;
         for(int i = 0; i < block_size(); i++){
           unsigned char m = 0x00;
-          if(d_frozen_bit_positions[nfbit] == i){
+          if(nfbit < d_frozen_bit_positions.size() && d_frozen_bit_positions[nfbit] == i){
```
jdemel: I'd suggest to update this line to make it more C++ic.
It should state: `d_frozen_bit_positions.at(nfbit) == i`.
This would also throw an error if indices are incorrect.

marcusmueller: no, that's not more C++ic, at all; C++ lives its overloaded operators, and the fact that I enabled bounds checking on that (which `at` would do inherently, I agree) was what led us to this, but it's definitely not more C++ to use a method when you could use an index operator :)

gr-fec/lib/polar_common.cc
marcusmueller: Then we have a bug somewhere else! Our code shouldn't be dealing with more frozen bits than we have, right?

PR #1885, opened 2018-07-08
py3k conv: file_metadata_sink::make now accepts unserialized PMT

marcusmueller:
This was necessitated by #1852

### DIFF
--- a/gr-blocks/include/gnuradio/blocks/file_meta_sink.h
+++ b/gr-blocks/include/gnuradio/blocks/file_meta_sink.h
@@ -87,17 +87,20 @@ namespace gr {
        * \param complex (bool): If data stream is complex
        * \param max_segment_size (size_t): Length of a single segment
        *    before the header is repeated (in items).
-       * \param extra_dict (string): a serialized PMT dictionary of extra
-       *    information. Currently not supported.
+       * \param extra_dict: a PMT dictionary of extra
+       *    information.
        * \param detached_header (bool): Set to true to store the header
        *    info in a separate file (named filename.hdr)
        */
-      static sptr make(size_t itemsize, const std::string &filename,
-		       double samp_rate=1, double relative_rate=1,
-		       gr_file_types type=GR_FILE_FLOAT, bool complex=true,
-		       size_t max_segment_size=1000000,
-		       const std::string &extra_dict="",
-		       bool detached_header=false);
+      static sptr make(size_t itemsize,
+                       const std::string &filename,
+                       double samp_rate=1,
+                       double relative_rate=1,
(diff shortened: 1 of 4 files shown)


gr-blocks/include/gnuradio/blocks/file_meta_sink.h
```diff
@@ -88,15 +88,15 @@ namespace gr {
...
+       *    information.
        * \param detached_header (bool): Set to true to store the header
        *    info in a separate file (named filename.hdr)
        */
       static sptr make(size_t itemsize, const std::string &filename,
 		       double samp_rate=1, double relative_rate=1,
 		       gr_file_types type=GR_FILE_FLOAT, bool complex=true,
 		       size_t max_segment_size=1000000,
-		       const std::string &extra_dict="",
+           const std::vector<uint8_t> &extra_dict=std::vector<uint8_t>(),
```
mbr0wn: This is an API change (which is OK, we're on master). However, when do we express serialized `pmt_t` as `std::vector<uint8_t>` in C++-land? I thought it was always `string`.

mbr0wn: When I tried directly exporting `std::vector<uint8_t>`, it converted it to tuples in Python, which is not what we wanted. My wrappers goes to std::vector<uint8_t>` intermediately because I need to avoid the UTF8 conversion, then my custom Python will turn that into a bytes string.

marcusmueller: Exactly. That's the logical step to take – we don't want the output of pmt.serialize to be interpreted as unicode – it's just an array of 8 bit numbers.

Now, this change makes the complementary operation – feeding a serialized PMT back into C++-land – possible. It happens that `vector<uint8_t>` is something that SWIG is able to convert `bytes` to. That's why it was chosen (I discovered that you were converting through `vector<uint8_t>` only after figuring out how to accept `bytes` from pythonland in c++land).

marcusmueller: So, between a rock and a hard place:

* C++ `pmt::serialize` gives us `std::string`
* We can't accept `std::string` here, because we can't teach SWIG that it can convert `bytes` to `std::string`.

mbr0wn: Actually, yeah, I'm not sure why the block wasn't already doing that :)

Have I recently mentioned my feelings for SWIG? :grimacing:

marcusmueller:
@mbr0wn  changed block to take unserialized PMT, fixes unit tests. Please, re-review :)

mbr0wn:
This is the right way to do this, IMO.

PR #1891, opened 2018-07-10
Forward porting Out-Of-Bound access fixes to allow for more precise testing

marcusmueller:
On several systems, `_GLIBCXX_ASSERTIONS` leads to runtime aborts, e.g. when getting the pointer to the second output stream in a block that might be configured to only have one.
This patchset just remedies that.

### DIFF
--- a/gr-blocks/lib/float_to_complex_impl.cc
+++ b/gr-blocks/lib/float_to_complex_impl.cc
@@ -54,8 +54,7 @@ namespace gr {
 				gr_vector_const_void_star &input_items,
 				gr_vector_void_star &output_items)
     {
-      float        *r = (float *)input_items[0];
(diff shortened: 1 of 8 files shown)


gr-blocks/lib/float_to_complex_impl.cc
```diff
@@ -65,10 +64,13 @@ namespace gr {
 	break;
 
       case 2:
+      {
 	//for (size_t j = 0; j < noutput_items*d_vlen; j++)
 	//  out[j] = gr_complex (r[j], i[j]);
```
noc0lour: One could actually remove the commented lines in a separate commit.

marcusmueller: yep, but let's not do that pre-merge, should we?

PR #1893, opened 2018-07-10
Remove catch statements from thread_body_wrapper.h

pwicks86:
This commit removes the catch statements from thread_body_wrapper.h. Currently, a single block in a flowgraph may throw an exception and stop functioning, but the only indication
of this to the user is printing the exception's `what()` to stderr. This leaves a zombie flowgraph with at least one block which is nonfunctional, which seems like behavior that is never desirable. As far as I can tell, the only mechanism for recovering from such a state is to lock the flowgraph and manually replace the broken block, which is a difficult pattern to implement.

With this patch, the process will terminate immediately, which I think will make "production" usage of GNU Radio more practical, as it allows for other process management tools like systemd to restart a crashed application.

Additionally, the uncaught exception will call `std::terminate`, which by default will still print the exception's `what()`, so debugging information is not lost. I've got another [PR](https://github.com/gnuradio/gnuradio/pull/1894) following this one that adds a `terminate_handler` to ensure that any exceptions get information about them printed and optionally print a backtrace.

### DIFF
--- a/gnuradio-runtime/include/gnuradio/thread/thread_body_wrapper.h
+++ b/gnuradio-runtime/include/gnuradio/thread/thread_body_wrapper.h
@@ -38,20 +38,42 @@ namespace gr {
     private:
       F d_f;
       std::string d_name;
+      bool d_catch_exceptions;
 
     public:
-      explicit thread_body_wrapper(F f, const std::string &name="")
-        : d_f(f), d_name(name) {}
+      explicit thread_body_wrapper(F f, const std::string &name="",
+                                   bool catch_exceptions=true)
+        : d_f(f), d_name(name), d_catch_exceptions(catch_exceptions) {}
 
       void operator()()
       {
         mask_signals();
 
-        try {
-          d_f();
-        }
-        catch(boost::thread_interrupted const &)
-        {
+        if(d_catch_exceptions) {
+          try {
+            d_f();
+          }
+          catch(boost::thread_interrupted const &)
+          {
+          }
+          catch(std::exception const &e)
+          {
+            std::cerr << "thread[" << d_name << "]: "
+                      << e.what() << std::endl;
+          }
+          catch(...)
+          {
+            std::cerr << "thread[" << d_name << "]: "
+                      << "caught unrecognized exception\n";
+          }
+
+        } else {
+          try {
+            d_f();
+          }
+          catch(boost::thread_interrupted const &)
+          {
+          }
         }
       }
     };
--- a/gnuradio-runtime/include/gnuradio/thread/thread_body_wrapper.h
+++ b/gnuradio-runtime/include/gnuradio/thread/thread_body_wrapper.h
@@ -51,18 +51,8 @@ namespace gr {
           d_f();
         }
         catch(boost::thread_interrupted const &)
-          {
-          }
-        catch(std::exception const &e)
-          {
-            std::cerr << "thread[" << d_name << "]: "
-                      << e.what() << std::endl;
-          }
-        catch(...)
-          {
-            std::cerr << "thread[" << d_name << "]: "
-                      << "caught unrecognized exception\n";
-          }
+        {
+        }
       }
     };
 
--- a/gnuradio-runtime/include/gnuradio/top_block.h
+++ b/gnuradio-runtime/include/gnuradio/top_block.h
@@ -30,7 +30,7 @@ namespace gr {
 
   class top_block_impl;
 
-  GR_RUNTIME_API top_block_sptr make_top_block(const std::string &name);
+  GR_RUNTIME_API top_block_sptr make_top_block(const std::string &name, bool catch_exceptions=true);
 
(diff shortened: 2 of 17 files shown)


mbr0wn:
Lol, you already thought of that in #1894 :+1:

mbr0wn:
This is not an API change in the usual sense, but it does change behaviour so I'm flagging it as such.

marcusmueller:
Agreeing with @mbr0wn's API Change flag.

marcusmueller:
So, having thought about this, it's clearly a pretty big change!

The current behaviour, just more-or-less silently failing, will allow the block executors of any downstream block to function, and hence allow the flow graph to flush out existing samples. If I grok the consequences of this change correctly, we'll see an instant termination of the program.

So, that'd actually be a functional regression to a lot of people, because it means that "in flight" samples will no longer e.g. be written to file sinks etc.

So, I'm afraid I can't fully approve of this; I fully embrace that a single block failing leaing to the program stalling in a zombie state is unacceptable, but now introducing unconditional abort isn't feasible, either.

I think we should have both:

* a block in trouble leading to timely termination tremendously helps tools that test code,  because a unit test that triggers total trashing of tpb state now leads to blocking, but:
* a block crash merely means that the flow graph should orderly shut down.

So, what do you think of the following:

* keep the exception handler, add proper logging
* make it do something configurable:
  * in its default behaviour, notify the runtime that  this block is done 
  * with a debug flag set, trigger an `abort` or just raise the exception further, so that a debugger might inspect the overall state of the program as close as possible as we get to the instant the block crashed

storborg:
I'm jumping in here since @pwicks86 is on vacation and might be slow to respond (we work together).

We've actually discussed adding a flag to control this behavior, but weren't sure if the extra complexity to pass the parameter around was justified by the use cases. What you've proposed sounds good, and certainly satisfies our motivations behind this change. I've pushed another commit which adds that flag, enabled by passing ``catch_exceptions=False`` to the top block. That commit includes an additional test to ensure that ``catch_exceptions=True`` results in the old behavior.

There are a couple details to note:

1. A fairly desirable feature is the ability to print a stack trace inside a ``std::terminate`` handler, as added in #1894 (we thought it best to do a separate PR since that change adds a ``libunwind`` dependency). In order to be able to access the stack trace at all, exceptions from within the thread body function cannot ever be caught in the first place. This is why I didn't use a conditional re-raise in the thread body wrapper's catch.

One negative side effect of this is that the thread ID is not printed on an exception, so if you have multiple instances of a given block you don't know which one raised the exception. An informal poll of the GNU Radio users around the office suggested that the stack trace was more important than the thread ID, particularly because log statements can often tell you which instance was responsible in any case.

2. The ``catch_exceptions`` behavior is not currently exposed to GRC: I think it should be, and either @pwicks86 or I can add that on this PR as well.

3. Not sure what you mean by "add proper logging". Do you mean switching the "catch exceptions mode" statements from stderr to log macros? Happy to make whatever changes there you think are appropriate.

I'm also about to merge this change over to the other PR (#1894) as well.

pwicks86:
Just to to add a couple things:

- The reason I didn’t log the output from the terminate handler using the gnuradio logger is that I was concerned about situations where the handler was called and the logger wasn’t yet available or had already been destroyed, whereas `std::cerr` is guaranteed to always be available. 

- When you say that a block crash should result in orderly shut down, do you mean that in the case where exceptions are caught by the thread body wrapper that we should additionally signal the block is done? If this is done, it seems there are a lot of flow graph topologies in which the flow graph will continue to run forever.

- As for the thread id/name, would it be possible to store it somewhere in thread local storage so then the terminate handler could print it as well?

pwicks86:
Are the updates we've made to this PR and #1894 are workable, or is there a different tack we need to take? No worries if it hasn't been reviewed again yet, just got back from vacation and wanted to make sure this didn't get lost in the shuffle.

noc0lour:
After the big merge this needs a rebase as well.

michaelld:
I like the idea behind this PR, especially that one can control the FG behavior when an exception happens in a given block. Looks like it needs a rebase, but otherwise if it works as desired I think it's a good addition.

noc0lour:
Currently I'm in the process of rebasing to get this PR out of the queue. I'm wondering if the change in the constructor is really necessary? A default value + a setter function should be equally well suited for the job? 
I'm pointing that out because something like:
```
     def __init__(${param_str}):
-        gr.top_block.__init__(self, "${title}")
+        gr.top_block.__init__(self, "${title}", catch_exceptions=${catch_exceptions})
``` 
doesn't really work with the mako templates since catch_exceptions isn't exposed to the template generator. Other options use `flowgraph.get_option` but that's not really cool for constructors.

noc0lour:
@pwicks86 I have rebased the branch on master if you want you can check it out here and force-push on your branch: https://github.com/noc0lour/gnuradio/tree/feature/remove-catch

PR #1894, opened 2018-07-10
Feature/add terminate handler

pwicks86:
This patch adds a terminate_handler function that prints information about any uncaught exceptions and also attempts to print a backtrace for said exception after which it calls std::abort. The terminate handler is hooked in at top_block initialization, as that seemed the most sensible place to put it. The backtrace is printed via libunwind, which is a new optional dependency. If libunwind is not found/installed, then the terminate handler will still print information about the exception and then exit. 

This PR relies on a previous [PR](https://github.com/gnuradio/gnuradio/pull/1893).

### DIFF
--- a/gnuradio-runtime/lib/terminate_handler.cc
+++ b/gnuradio-runtime/lib/terminate_handler.cc
@@ -1,3 +1,25 @@
+/* -*- c++ -*- */
+/*
+ * Copyright 2018 Free Software Foundation, Inc.
+ *
+ * This file is part of GNU Radio
+ *
+ * GNU Radio is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 3, or (at your option)
+ * any later version.
+ *
+ * GNU Radio is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with GNU Radio; see the file COPYING.  If not, write to
+ * the Free Software Foundation, Inc., 51 Franklin Street,
+ * Boston, MA 02110-1301, USA.
+ */
+
 #include <stdexcept>
 #include <iostream>
 #include <exception>
--- a/gnuradio-runtime/lib/terminate_handler.cc
+++ b/gnuradio-runtime/lib/terminate_handler.cc
@@ -1,0 +1,137 @@
+#include <stdexcept>
+#include <iostream>
+#include <exception>
+#include <regex>
+#include <pmt/pmt.h>
+#include <boost/thread.hpp>
+
+#include <cxxabi.h>
+
+#ifdef HAVE_LIBUNWIND
+#define UNW_LOCAL_ONLY
+#include <libunwind.h>
+#include <cstdio>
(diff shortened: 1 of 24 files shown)


gnuradio-runtime/lib/terminate_handler.cc
```diff
@@ -0,0 +1,137 @@
+#include <stdexcept>
```
mbr0wn: This file needs the copyright header.

gnuradio-runtime/lib/terminate_handler.cc
```diff
@@ -0,0 +1,137 @@
...
+#include <iostream>
```
mbr0wn: We have some order of includes in the coding guidelines, but I'm not sure they've been finalized yet.

gnuradio-runtime/lib/terminate_handler.cc
```diff
@@ -0,0 +1,137 @@
...
+                std::free(demangled);
+            } else {
+                std::fprintf(stderr, " -- error: unable to obtain symbol name for this frame\n");
+            }
+        }
+    }
+#endif
+
+    void terminate_handler_impl() {
+        std::cerr << "terminate reached from: " << boost::this_thread::get_id();
```
mbr0wn: Adding "thread ID" here would make the error message more useful, since it may not be obvious what's getting printed after the colon.

pwicks86:
> What happens if there are multiple top blocks in a process, will that work?

If there’s multiple top_blocks in a process, it should work the same. `std::set_terminate` is thread-safe, so each top_block will reset the terminate handler to the same function.

storborg:
Actually, the issue here is just that you the ``libunwind-dev`` package is needed in addition to ``libunwind8``. The stack trace printing works fine when both are present.

noc0lour:
So one would need the `-dev` package during runtime? Or is it sufficient to have it around only during build time?

storborg:
Sufficient to have it around only during build time. If it is not present at build time, the terminate handler simply won't offer stack trace printing.

storborg:
Another note: this PR requires ``SET(CMAKE_CXX_STANDARD 11)``, which used to conditionally exist (if cppunit was present), but was removed in 884b30809830bb6e5df04253533e9e7482e6047b

One simple approach to dealing with that might be to just add this, which is kind of consistent with how cppunit inclusion used to behave:

```
if(libunwind_FOUND)
  SET(CMAKE_CXX_STANDARD 11)
endif()
```

storborg:
A related question: would it be more palatable to merge this functionality if it were provided via Boost.Stacktrace rather than libunwind?

noc0lour:
@storborg  on `master` we should have c++11 support unconditionally enabled.

marcusmueller:
no, right now, C++11 is on next. I was about to enforce c++11 as standard on master, but then realized I'd do a major merge of next functionality. To avoid doing that twice, and generating subtle conflicts, I decided that master becomes C++11 the moment we merge next.

noc0lour:
After the big merge this needs a rebase.

noc0lour:
And we have C++11 now on master.

michaelld:
I like the idea behind this PR (and how it couples into the related one); wondering if now that we're requiring C++11 this could be done in just C++ without the extra dependency.

noc0lour:
Rebased on my rebased `feature/remove-catch` branch can be found here: https://github.com/noc0lour/gnuradio/tree/feature/add-terminate-handler

mbr0wn:
Continueing here: https://github.com/gnuradio/gnuradio/pull/3051

PR #1897, opened 2018-07-12
removed several readmes

777arc:
either had no useful info, or outdated info that isn't worth keeping up to date


mbr0wn:
Base branch is maint, though. Why is this even a branch?

777arc:
Oh I just made the changes like I am use to doing, using a separate branch and merging into maint.  What is the proper way?

mbr0wn:
Everything goes into master. Stable branch maintainers can backport. However, the actual `maint` branch was superseded by `maint-3.7` and shouldn't even exist any more (that was more of a comment to @marcusmueller ).

marcusmueller:
@mbr0wn we can't simply delete branches. There might be folks with build scripts that pull these, and there's definitely people that have their own branches based off maint, and those need to be able to correlate what they did in the past with the current state, no matter how dead maint is!

Deleting a branch changing history, so it'll never happen :( as much as I'd like to clean up this mess. I might want to change the README on that branch and add a big CMake warning, good catch.

marcusmueller:
In the meantime, let me recreate this commit on master.

marcusmueller:
manually cherry-picked per dea45c14abb5c4a76a507376198f954d01a74f9c

PR #1905, opened 2018-07-18
ctrlport: Remove superfluous 'import exceptions'

mbr0wn:
It's not needed, not even with Python 2.7. It will fail with Python 3,
though.


mbr0wn:
@marcusmueller This is a fine commit, the only problem is that the binding unit test still fails, but doesn't crash immediately. Instead, it times out which is kind of annoying. Still, this particular fix is correct AFAICT.

PR #1906, opened 2018-07-18
Fixed out-of-bound vector acces in PMT

marcusmueller:
empty `vec`.


marcusmueller:
@noc0lour looks like none of the builders is set to C++11. That would be a mistake?

marcusmueller:
Not doing this on master (requires C++11) but on next

marcusmueller:
reopening after downgrade to maint-3.7 pre-c++11 version

PR #1909, opened 2018-07-18
Fixed out-of-bound vector acces in PMT

marcusmueller:
empty `vec`.


marcusmueller:
merging after a month

PR #1910, opened 2018-07-19
Add a clear warning to README and CMake that `maint` is a dead end.

marcusmueller:
This should steer people towards using `maint-3.7` or `master`.

The changes are:

* Prologue in the README, including scary dragon
* WARNING file, hoping people read that (includes a scary dragon)
* CMake message (with a _scary_ dragon) and warning

Also, as soon as we adapt this, I'll tag the maint branch so that it's called `3.7-deadend` or so in a build.


mbr0wn:
Bit late, sorry, but this is a good change!

PR #1920, opened 2018-07-26
cmake: Bump required version for thrift to 0.11.0

johschmitz:
We require TRecursive.py from thrift and that only exists
starting with thrift 0.11.0 onwards.

### DIFF
--- a/cmake/Modules/FindThrift.cmake
+++ b/cmake/Modules/FindThrift.cmake
@@ -1,7 +1,7 @@
 INCLUDE(FindPkgConfig)
 PKG_CHECK_MODULES(PC_THRIFT thrift)
 
-set(THRIFT_REQ_VERSION "0.9.2")
+set(THRIFT_REQ_VERSION "0.11.0")
 
 # If pkg-config found Thrift and it doesn't meet our version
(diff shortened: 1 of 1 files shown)


johschmitz:
I went through all the thrift versions to get gr-perf-monitorx running.
TRecursive was the last error that made me move up to 0.11.0
The remaining problem then is that gnuradio doesn't compile with 0.11.0,
kind of a deadlock situation that I wasn't able to solve because I am not
familiar with thrift.

Marcus Müller <notifications@github.com> schrieb am Fr., 27. Juli 2018,
09:28:

marcusmueller:
So, I think we should be chasing the bug that leads to not being able to work with thrift 0.10.0, instead of pushing the requirements. Agree?

marcusmueller:
anyway, the perf monitor runs without problem on my thrift 0.10.0 machine, so I'm closing this, as it's not solving a problem, as far as I can tell.

@johschmitz don't let this issue die, though! Please do an in-depth bug report as an issue here (if you haven't already), and we'll try to sort things out. If we really need to push the dependency version, we might even do that, but as you say yourself, pushing for 0.11 breaks other things, so this is nothing I can merge, I'm afraid.

Issue #1912, opened 2018-07-22
multiply_const_vxx: render type for const correctly


dl1ksv:
Porting a flow graph from 3.7 to 3.8 the block multiply const fails with
Assertion "len(const) ==  vlen" did not evaluate. No idea why.

Replacing this block by 'Fast Multiply Const' works ,
but this block only checks Vlen > 0.

haakov:
Thanks! This is caused by `const` being either complex, float, int or short. Therefore you can't do `len()` on it. Not quite sure how this worked in 3.7, because you couldn't do `len()` on those types in Python 2 either, but I suppose the fix would be either to remove `len()` or to cast the int as a string? I'm not quite sure about the purpose of this assert.

dl1ksv:
It seems that 3.7 handled it the same way:

	</param>
	<check>len($const) == $vlen</check>
	<check>$vlen &gt; 0</check>
	<sink>

I don't know the yml syntax, so could it be possible that this is just malformed in 3.8 ?

noc0lour:
@dl1ksv Did you use the vector version of multiply const? In that case `const` is a list which has to be the same length as the input vector.

dl1ksv:
@noc0lour You are right. I took the block blocks_multiply_const_vxx. I think v stands for vector.
Changing the constant value into a list, did it.
But it was a direct port from 3.7.12 and there it worked without a list.

noc0lour:
Oh, probably this yaml file needs some overhaul @haakov. So that the enum you can choose from sets the type of the const variable to `$type_vector` like it did in the xml case. And then a single 0 should be rendered into a list upon creating the flowgraph.

haakov:
Sure!

So, to clarify, does [this gist](https://gist.github.com/haakov/fb381bf3de1dd6a02b697c9d92d7dae5#file-blocks_multiply_const_vxx-block-yml-L39) reflect what we want to achieve? 

(This was an annoyingly tricky bug to fix in an eye pleasing way :) )

haakov:
Pull request created: https://github.com/gnuradio/gnuradio/pull/1925

mbr0wn:
This will also fix it, although it's not as complete as @haakov's PR: https://github.com/gnuradio/gnuradio/pull/2194

haakov:
This one can be closed, yes?

mbr0wn:
Yep.

PR #1925, opened 2018-07-28
grc: Un-hide option_attributes from rest of parameters

haakov:
Following the discussion from https://github.com/gnuradio/gnuradio/issues/1912 and [this gist](https://gist.github.com/haakov/fb381bf3de1dd6a02b697c9d92d7dae5).

This fixes the problem with a lot of blocks having some of their parameters handled as `raw` while the parameter's `dtype` is in fact dependent on the block's chosen type. This happened because the namespace didn't contain the `option_attributes` when evaluating the block. 

I've given this fix its own PR because I'm only 90% sure this doesn't break anything.

### DIFF
--- a/grc/core/blocks/_build.py
+++ b/grc/core/blocks/_build.py
@@ -20,9 +20,11 @@ from __future__ import absolute_import
 import collections
 import itertools
 import re
+import six
 
 from ..Constants import ADVANCED_PARAM_TAB
 from ..utils import to_list
+from ..Messages import send_warning
 
 from .block import Block
 from ._flags import Flags
@@ -51,7 +53,7 @@ def build(id, label='', category='', flags='', documentation='',
     cls.inputs_data = _build_ports(inputs, 'sink') if inputs else []
     cls.outputs_data = _build_ports(outputs, 'source') if outputs else []
     cls.parameters_data = _build_params(parameters or [],
-                                        bool(cls.inputs_data), bool(cls.outputs_data), cls.flags)
+                                        bool(cls.inputs_data), bool(cls.outputs_data), cls.flags, block_id)
     cls.extra_data = kwargs
 
     templates = templates or {}
@@ -86,7 +88,7 @@ def _build_ports(ports_raw, direction):
     return ports
 
 
-def _build_params(params_raw, have_inputs, have_outputs, flags):
+def _build_params(params_raw, have_inputs, have_outputs, flags, block_id):
     params = []
 
     def add_param(**data):
@@ -114,6 +116,9 @@ def _build_params(params_raw, have_inputs, have_outputs, flags):
         if param_id in params:
             raise Exception('Param id "{}" is not unique'.format(param_id))
 
+        if 'option_attributes' in param_data:
+            _validate_option_attributes(param_data, block_id)
+
         base_key = param_data.get('base_key', None)
         param_data_ext = base_params_n.get(base_key, {}).copy()
         param_data_ext.update(param_data)
@@ -133,3 +138,15 @@ def _single_mako_expr(value, block_id):
     if not (value.startswith('${') and value.endswith('}')):
         raise ValueError('{} is not a mako substitution in {}'.format(value, block_id))
     return value[2:-1].strip()
+
+
+def _validate_option_attributes(param_data, block_id):
+    if param_data['dtype'] != 'enum':
+        send_warning('{} - option_attributes are for enums only, ignoring'.format(block_id))
+        del param_data['option_attributes']
+    else:
+        for key in list(param_data['option_attributes'].keys()):
+            if key in dir(str):
+                del param_data['option_attributes'][key]
+                send_warning('{} - option_attribute "{}" overrides str, ignoring'.format(block_id, key))
+
--- a/grc/core/io/yaml.py
+++ b/grc/core/io/yaml.py
@@ -22,6 +22,7 @@ from collections import OrderedDict
 import six
 import yaml
 
+from ..params.param import attributed_str
 
 class GRCDumper(yaml.SafeDumper):
     @classmethod
@@ -79,6 +80,7 @@ GRCDumper.add_representer(ListFlowing, GRCDumper.represent_list_flowing)
 GRCDumper.add_representer(tuple, GRCDumper.represent_list)
 GRCDumper.add_representer(MultiLineString, GRCDumper.represent_ml_string)
 GRCDumper.add_representer(yaml.nodes.ScalarNode, lambda r, n: n)
+GRCDumper.add_representer(attributed_str, GRCDumper.represent_str)
 
 
 def dump(data, stream=None, **kwargs):
--- a/grc/core/params/param.py
+++ b/grc/core/params/param.py
@@ -31,6 +31,7 @@ from ..utils.descriptors import Evaluated, EvaluatedEnum, setup_names
 from . import dtypes
 from .template_arg import TemplateArg
 
+attributed_str = type('attributed_str', (str,), {})
 
 @setup_names
 class Param(Element):
@@ -178,6 +179,10 @@ class Param(Element):
         # ID and Enum types (not evaled)
         #########################
         if dtype in ('id', 'stream_id') or self.is_enum():
+            if self.options.attributes:
+                expr = attributed_str(expr)
+                for key, value in self.options.attributes[expr].items():
+                    setattr(expr, key, value)
             return expr
 
         #########################
--- a/grc/core/blocks/virtual.py
+++ b/grc/core/blocks/virtual.py
@@ -45,7 +45,7 @@ class VirtualSink(Block):
         self.inputs_data = _build_ports(self.inputs, 'sink') if self.inputs else []
         self.outputs_data = _build_ports(self.outputs, 'source') if self.outputs else []
         self.parameters_data = _build_params(self.parameters or [],
-                                bool(self.inputs), bool(self.outputs), self.flags)
+                                bool(self.inputs), bool(self.outputs), self.flags, self.key)
 
         super(VirtualSink, self).__init__(parent, **kwargs)
         self.params['id'].hide = 'all'
@@ -78,7 +78,7 @@ class VirtualSource(Block):
         self.inputs_data = _build_ports(self.inputs, 'sink') if self.inputs else []
         self.outputs_data = _build_ports(self.outputs, 'source') if self.outputs else []
         self.parameters_data = _build_params(self.parameters or [],
-                                bool(self.inputs), bool(self.outputs), self.flags)
+                                bool(self.inputs), bool(self.outputs), self.flags, self.key)
 
         super(VirtualSource, self).__init__(parent, **kwargs)
         self.params['id'].hide = 'all'
--- a/grc/core/params/dtypes.py
+++ b/grc/core/params/dtypes.py
@@ -97,6 +97,9 @@ def validate_scalar(param):
 def validate_vector(param):
     # todo: check vector types
 
+    if param.get_evaluated() is None:
+        raise ValidateError('Expression {!r} is invalid for type{!r}.'.format(param.get_evaluated(), param.dtype))
+
     valid_types = Constants.PARAM_TYPE_MAP[param.dtype.split('_', 1)[0]]
     if not all(isinstance(item, valid_types) for item in param.get_evaluated()):
         raise ValidateError('Expression {!r} is invalid for type {!r}.'.format(


haakov:
Right!

Hm I'm not sure I'm quite following, this means changing both the GRC code and the YAML block files, yes?

marcusmueller:
Closes #1912.

marcusmueller:
@skoslowski if you can spare a minute, I think @haakov  and I both need clarification on this?

skoslowski:
We're doing a hangout tonight. PM me if you want to join.

marcusmueller:
updates?

haakov:
Sure! So, the approach we've decided on will not impact the YAML files, only the GRC code. :)

marcusmueller:
@haakov so, that means that you'll change this PR? A bit confused about what to do with this :)

grc/core/params/param.py
```diff
@@ -178,6 +178,10 @@ def evaluate(self):
         # ID and Enum types (not evaled)
         #########################
         if dtype in ('id', 'stream_id') or self.is_enum():
+            if self.options.attributes:
+                expr = collections.UserString(expr)
+                for key, value in self.options.attributes[expr].items():
+                    setattr(expr, key, value)
```
skoslowski: to make this save, we should probably prevent str methods from being overwritten here

grc/core/params/param.py
```diff
@@ -178,6 +178,10 @@ def evaluate(self):
         # ID and Enum types (not evaled)
         #########################
         if dtype in ('id', 'stream_id') or self.is_enum():
+            if self.options.attributes:
+                expr = collections.UserString(expr)
```
skoslowski: since we're not trying to override any methods (only add attributes), defining a new class `mstr = type('mstr', (str,), {})` would have been my first choice,  but this works as well.

grc/core/utils/descriptors/evaluated.py
```diff
@@ -17,8 +17,7 @@
 
 from __future__ import absolute_import
 
-import six
-
+import six, collections
```
skoslowski: please don't do this, each import shall be in a separate statement. especially stdlib and 3rd party libs should be separated.

haakov: Right! Will fix

haakov:
@skoslowski Okay, I've revised it. Still works as far as I can tell. :+1: 

Not sure if the `attributed_str` type declaration should be moved somewhere else, or if it's okay where it is. About non-enum params and `option_attributes`, is it even possible for non-enums to have options?

grc/core/params/param.py
```diff
@@ -178,6 +179,10 @@ def evaluate(self):
         # ID and Enum types (not evaled)
         #########################
         if dtype in ('id', 'stream_id') or self.is_enum():
+            if self.options.attributes:
+                expr = attributed_str(expr)
+                for key, value in self.options.attributes[expr].items():
+                    setattr(expr, key, value)
```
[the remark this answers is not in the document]
haakov: Right!

dl1ksv:
I just detected a problem with your pr. Applying it leads to problems generating hier type flowgraphs.

I get 

     File "/usr/lib/python3/dist-packages/yaml/representer.py", line 229, in represent_undefined
    raise RepresenterError("cannot represent an object: %s" % data)
    yaml.representer.RepresenterError: cannot represent an object: complex

or 

      File "/usr/lib/python3/dist-packages/yaml/representer.py", line 229, in represent_undefined
    raise RepresenterError("cannot represent an object: %s" % data)
     yaml.representer.RepresenterError: cannot represent an object: float
 
errors depending of the type of the pad source/sink .

The complete traceback is:

Generating: '/home/schroer/.grc_gnuradio/hier2Msg.py'
Generate Error: cannot represent an object: float
>>> Failure
  File "/usr/local/gnuradio/lib/python3.6/dist-packages/gnuradio/grc/gui/Application.py", line 691, in _handle_action
    generator.write()
  File "/usr/local/gnuradio/lib/python3.6/dist-packages/gnuradio/grc/core/generator/hier_block.py", line 39, in write
    data = yaml.dump(self._build_block_n_from_flow_graph_io())
  File "/usr/local/gnuradio/lib/python3.6/dist-packages/gnuradio/grc/core/io/yaml.py", line 87, in dump
    return yaml.dump_all([data], **config)
  File "/usr/lib/python3/dist-packages/yaml/__init__.py", line 188, in dump_all
    dumper.represent(data)
  File "/usr/lib/python3/dist-packages/yaml/representer.py", line 26, in represent
    node = self.represent_data(data)
  File "/usr/lib/python3/dist-packages/yaml/representer.py", line 47, in represent_data
    node = self.yaml_representers[data_types[0]](self, data)
  File "/usr/local/gnuradio/lib/python3.6/dist-packages/gnuradio/grc/core/io/yaml.py", line 43, in represent_ordered_mapping
    node_value = self.represent_data(item_value)
  File "/usr/lib/python3/dist-packages/yaml/representer.py", line 47, in represent_data
    node = self.yaml_representers[data_types[0]](self, data)
  File "/usr/lib/python3/dist-packages/yaml/representer.py", line 197, in represent_list
    return self.represent_sequence('tag:yaml.org,2002:seq', data)
  File "/usr/lib/python3/dist-packages/yaml/representer.py", line 91, in represent_sequence
    node_item = self.represent_data(item)
  File "/usr/lib/python3/dist-packages/yaml/representer.py", line 47, in represent_data
    node = self.yaml_representers[data_types[0]](self, data)
  File "/usr/local/gnuradio/lib/python3.6/dist-packages/gnuradio/grc/core/io/yaml.py", line 43, in represent_ordered_mapping
    node_value = self.represent_data(item_value)
  File "/usr/lib/python3/dist-packages/yaml/representer.py", line 57, in represent_data
    node = self.yaml_representers[None](self, data)
  File "/usr/lib/python3/dist-packages/yaml/representer.py", line 229, in represent_undefined
    raise RepresenterError("cannot represent an object: %s" % data)
yaml.representer.RepresenterError: cannot represent an object: float

haakov:
@dl1ksv Thanks! I'll push a fix for it shortly.

dl1ksv:
@haakov Thanks, it works for me !

grc/core/io/yaml.py
```diff
@@ -39,6 +40,8 @@ def represent_ordered_mapping(self, data):
             self.represented_objects[self.alias_key] = node
 
         for item_key, item_value in six.iteritems(data):
+            if type(item_value) == attributed_str:
+                item_value = str(attributed_str)
```
skoslowski: The whole point of this class is to make custom types dump-able. Instead of converting, I suggest to add a dumper for attributed_str. This would not only solve this issue, but also any conversion to yaml that might happen somewhere else or in future code...

dl1ksv:
Well, going further on testing:
The hier2 block and it's yml representation is generated, but now I can't load a flowgraph containing this block.

No blocks are displayed and the Error window show the message for each pad:

Type "<class 'gnuradio.grc.core.params.param.attributed_str'>" is not a possible type

haakov:
@dl1ksv oh wow haha, my bad! :)

I agree with @skoslowski though, it does make sense to create a new representer.

dl1ksv:
Just some more problems with hier2 blocks:

Copying a variable of type vector  form a flow graph into a  ( new ) hier2 block leads to 

      File "/usr/local/gnuradio/lib/python3.6/dist-packages/gnuradio/grc/gui/Application.py", line 224, in _handle_action
    page.flow_graph.update()
      File "/usr/local/gnuradio/lib/python3.6/dist-packages/gnuradio/grc/gui/canvas/flowgraph.py", line 175, in update
    self.validate()
      File "/usr/local/gnuradio/lib/python3.6/dist-packages/gnuradio/grc/core/base.py", line 40, in validate
    child.validate()
      File "/usr/local/gnuradio/lib/python3.6/dist-packages/gnuradio/grc/core/blocks/block.py", line 136, in validate
    Element.validate(self)
      File "/usr/local/gnuradio/lib/python3.6/dist-packages/gnuradio/grc/core/base.py", line 40, in validate
    child.validate()
      File "/usr/local/gnuradio/lib/python3.6/dist-packages/gnuradio/grc/core/params/param.py", line 158, in validate
    validator(self)
      File "/usr/local/gnuradio/lib/python3.6/dist-packages/gnuradio/grc/core/params/dtypes.py", line 101, in validate_vector
    if not all(isinstance(item, valid_types) for item in param.get_evaluated()):
    TypeError: 'NoneType' object is not iterable


The definition of the vector looks like

\- name: variable_0
   id: variable
  parameters:
    value: (-1,-1,-1,-1,1,1,-1,-1,1,1,-1,-1,1,1,-1,-1,1,1,-1,-1,1,1,-1,-1,1,1,-1,-1,1,1,-1,-1,-1,-1,-1,-1,-1,-1,-1)
   states:
    coordinate: [624, 76.0]
    rotation: 0
    state: true

which works in a qtgui based flowgrah.

@haakov is there any progress on this issue ?

dl1ksv:
Well, the problem was not the vector itself. But I copied an fir filter block that referenced this vector.

But copying changed the id's. So the copied fir filter now referenced a non existing vector.
From this point you have no chance to edit the flowgraph further. So I saved it to a file and restarted grc.

Now  the flowgraph seemed to be empty ! The only way to get the flowgraph editable was to edit the grc file by hand.

So it might by rather a problem of the copy / paste mechanism between flowgraphs than a problem of the hier2 block.

But I would wish the chance to hanlde such an error inside grc !

haakov:
@dl1ksv I've been a little caught up with some lab report deadlines lately, but I should be able to get some work done in the next couple of days. 

I'm not quite able to reproduce the bug you mentioned with copying/pasting though, could you elaborate? :)

dl1ksv:
At the first sight, these patches look well.
Only the blocking behaviour of grc referencing a non existing vector remains.

I found a simple fix for this issue

Just add

    if param.get_evaluated() is None:
        raise ValidateError('In Parameter {!r} variable {!r} is undefined.'.format(
        param.name,param.value))

into params/dtypes.py in line 99 ff. just behind

    def validate_vector(param):
        # todo: check vector types
Perhaps you can add this to your pr.

haakov:
@dl1ksv Thanks! I believe it'd be best if a separate PR is created for the changes that are not related to `option_attributes` :)

I haven't been able to reproduce the bug with GRC hanging, but I agree that the user probably should be informed when the flowgraph is referring to nonexistent variables! Right now they are just evaluated to `None`, as far as I'm concerned.

About the `Ignoring extra key 'value'` warning, you're right. It should be `default`.

dl1ksv:
@haakov I forgot to mention that the blocking behaviour only exists after applying your patch. Sorry , my fault.

In the master branch I can  enter a non existing variable name for taps and the flowgraph evalutes to valid and I can generate the flowgraph.

Running the flowgraph ends in
TypeError: make() missing 1 required positional argument: 'taps'

I think it makes no sense to fix this in the master branch, as the dtypes are not handled correctly. Most of them are treated as raw. So  I'll focus on your pr.

haakov:
Oh okay, yeah then I agree it should be fixed here.

777arc:
Looking forward to this PR, I also had a bunch of flowgraphs break in 3.8 because of the Multiply Const block not accepting a scalar anymore.

dl1ksv:
:+1: Works for me, should be merged soon.

777arc:
I just tested this PR and it fixed my issue with Multiply Const, and didn't seem to introduce any new bugs as far as my test flowgraphs.

grc/core/blocks/_build.py
```diff
@@ -50,7 +52,7 @@ def build(id, label='', category='', flags='', documentation='',
 
     cls.inputs_data = _build_ports(inputs, 'sink') if inputs else []
     cls.outputs_data = _build_ports(outputs, 'source') if outputs else []
-    cls.parameters_data = _build_params(parameters or [],
+    cls.parameters_data = _build_params(block_id, parameters or [],
```
skoslowski: I suggest to move this to the end of the params

grc/core/blocks/_build.py
```diff
@@ -108,12 +110,29 @@ def add_param(**data):
...
+        if 'option_attributes' in param_data:
+            if param_data['dtype'] != 'enum':
+                send_warning('{} - option_attributes are for enums only, ignoring'.format(block_id, param_data.pop('option_attributes')))
+            else:
+                invalid_key = check_if_attr_overrides_str(param_data['option_attributes'])
+
+                if invalid_key:
+                    del param_data['option_attributes'][invalid_key]
+                    send_warning('{} - option_attribute "{}" overrides str, ignoring'.format(block_id, invalid_key, param_id))
+
```
skoslowski: How about we move both these added sections into a separate function - to keep things neat.

777arc:
what's the status on this one?

haakov:
@777arc Sorry about the delay, I'll implement @skoslowski's suggestions tomorrow!

grc/core/blocks/_build.py
haakov: @skoslowski I'm afraid "params" has several meanings here. Do you mean moving `block_id` to the end of the function parameters? Or moving `block_id` to the end of the function parameter called `parameters`?

grc/core/blocks/_build.py
haakov: Done!

noc0lour:
I think there is only the final review of @skoslowski missing.

grc/core/blocks/_build.py
skoslowski: I meant, to change to order of parameters on `_build_params`. A parameter that is only used if something goes wrong in log message, should not go first (but rather last) =)

grc/core/blocks/_build.py
```diff
@@ -114,6 +116,8 @@ def add_param(**data):
         if param_id in params:
             raise Exception('Param id "{}" is not unique'.format(param_id))
 
+        _validate_option_attributes(param_data)
```
skoslowski: I think you missed to forward `block_id`

grc/core/blocks/_build.py
haakov: That's what I thought :) Fixed!

grc/core/blocks/_build.py
haakov: Fixed

haakov:
Ok so I believe this is ready now. 

It might be worth noting that the validation for option_attributes is located in `_build.py`, which means that the warnings for a block with "invalid" option_attributes will occur only the first time GRC discovers this block. GRC will then amend the block by removing the offending parts of option_attributes and add the amended block to the `cache.json` file. Thus, GRC will not complain about this again unless `cache.json` is deleted.

dl1ksv:
Just found this bug
The dtypes of input and output pads now look as follows

inputs:
-   label: in
    dtype: <class 'gnuradio.grc.core.params.param.attributed_str'>
    vlen: 1

Same is for outputs.

I did not find the corresponding commit
but  
https://github.com/gnuradio/gnuradio/blob/f0bb3134e1215a72f3022fe6996e55ca3eccd57a/grc/core/io/yaml.py#L43-L44

should be removed, then it works for me.

haakov:
@dl1ksv Right, of course :) Those two lines should've been removed after the representer was added. :man_facepalming: 

Fixed!

mbr0wn:
This looks ripe for merging.

mbr0wn:
Merging manually. Thanks @haakov!

PR #1940, opened 2018-08-05
swig: fix numpy int conversion on py3k

noc0lour:
In Python2 numpy derives from the built-in fixed width integer type. In Py3k the integer type is variable width and thus NumPy does not derive its fixed-width types from it. 
As a result the numpy objects don't pass the PyLong_Check() necessary for conversion from Python to C/C++. 

Enabling SWIG_PYTHON_CAST_MODE enables casting by adding this code fragment: 
```
%#ifdef SWIG_PYTHON_CAST_MODE
  {
    int dispatch = 0;
    long v = PyInt_AsLong(obj);
    if (!PyErr_Occurred()) {
      if (val) *val = v;
      return SWIG_AddCast(SWIG_OK);
    } else {
      PyErr_Clear();
    }
    if (!dispatch) {
      double d;
      int res = SWIG_AddCast(SWIG_AsVal(double)(obj,&d));
      if (SWIG_IsOK(res) && SWIG_CanCastAsInteger(&d, LONG_MIN, LONG_MAX)) {
	if (val) *val = (long)(d);
	return res;
      }
    }
  }
%#endif
```

Withouth further investigation of sideeffects this skips the PyLong_Check by just using the PyInt_AsLong cast opportunistically on the given Python Object. Which should be ok since it checks for PyErr_Occurred() afterwards.

### DIFF
--- a/gnuradio-runtime/swig/gnuradio.i
+++ b/gnuradio-runtime/swig/gnuradio.i
@@ -83,6 +83,7 @@
 
 %begin %{
 #define SWIG_PYTHON_2_UNICODE
+#define SWIG_PYTHON_CAST_MODE
 %}
 
 #ifdef SWIGPYTHON


noc0lour:
If someone is interested in a minimal test case for that https://github.com/noc0lour/swig_vector

Running `meson & ninja` from a build dir and then  meson test --setup (py2|py3) will execute for py2/py3

noc0lour:
OK, definitely this is a API/behavioural change so `to_string()` always returns a bytes object (in Py2 this does not make a difference, but in Py3k)

gnuradio-runtime/swig/message.i
```diff
@@ -63,6 +69,8 @@ namespace gr {
   long message_ncurrently_allocated();
 }
 
+
+
```
dkozel: Mistaken whitespace diff?

noc0lour: Yes, will remove. A lot of copying around chunks of code has happened.

gnuradio-runtime/swig/message.i
```diff
@@ -22,6 +22,12 @@
 
 namespace gr {
 
+  %pythonappend message::to_string %{
+    import sys
+    if sys.version_info[0] > 2:
+      val =  val.encode("utf8", errors="surrogateescape")
```
marcusmueller: This is cool, but document that shit in docs/

gr-blocks/python/blocks/qa_message.py
```diff
@@ -58,8 +58,8 @@ def test_100(self):
         self.assertEquals(0, msg.length())
 
     def test_101(self):
-        s = 'This is a test'
-        msg = gr.message_from_string(s)
+        s = b'This is a test'
+        msg = gr.message_from_string(s.decode('utf8'))
         self.assertEquals(s, msg.to_string())
```
marcusmueller: Could you also test the other way around, throw in `s=unicode(…)`, and compare the result after encoding to utf8? This feels like something that will fall on our feet if we don't check stuff. And also, there's so many nice characters in UTF, why stick with ASCII? Why not 🦀?

gnuradio-runtime/swig/message.i
marcusmueller: I'll also put the idea out there that we deprecate all PMT to_string methods. Either you want a serialization, or you don't.

noc0lour: So, currently `to_string` is serializing the memory we have to a char* array?

gr-blocks/python/blocks/qa_message.py
noc0lour: :+1: I'll throw in more tests for "nice" unicode characters :)

gnuradio-runtime/swig/message.i
noc0lour: Not a question, but a statement.

gr-blocks/python/blocks/qa_message.py
noc0lour: s = unicode() is Py2 only code. `u"mystring"` is how you do unicode (with unicode_literals imported from __future__)

noc0lour:
hm, now a new error ocurred:  `SyntaxError: Non-ASCII character '\xe2' in file /data/src/pull/1940/next/gr-blocks/python/blocks/qa_message.py on line 66, but no encoding declared; see http://python.org/dev/peps/pep-0263/ for details`

gr-blocks/python/blocks/qa_message.py
noc0lour: Now testing is in place and should check all boxes (Had to explictely tell python about the utf-8 encoding of the file and import unicode_literals to have the same code for py2/py3). If people wish to use unicode in py2 they definitely have to do the same thing.

noc0lour:
Removed the message/unicode commits, this should be easy to merge now.

PR #1942, opened 2018-08-05
python: replace unfunctional old_div with functional code

noc0lour:
a previous change to py3k introduced syntactically incorrect
`old_div` function calls. A replacement with `//` and `/` where applicable is
more appropriate.
Eventually `from __future__ import division` needs to be added as well to have
the "real" division also for integer values in python

### DIFF
--- a/gr-trellis/examples/python/test_tcm.py
+++ b/gr-trellis/examples/python/test_tcm.py
@@ -99,8 +99,8 @@ def main():
     Es = 0
     for i in range(len(constellation)):
         Es = Es + constellation[i]**2
-    Es = Es / (old_div(len(constellation,dimensionality)))
-    N0=Es / pow(10.0,old_div(esn0_db,10.0)); # calculate noise variance
+    Es = Es / (len(constellation)//dimensionality)
+    N0=Es / pow(10.0,esn0_db/10.0); # calculate noise variance
 
     tot_b=0 # total number of transmitted bits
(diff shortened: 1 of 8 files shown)


gr-trellis/examples/python/test_tcm.py
```diff
@@ -112,7 +112,7 @@ def main():
         terr_p=terr_p+(e!=0)
         if ((i+1)%100==0) : # display progress
             print(i+1,terr_p, '%.2e' % ((1.0*terr_p) / (i+1)),tot_b,terr_b, '%.2e' % ((1.0*terr_b) / tot_b))
-	if e!=0:
+        if e!=0:
```
marcusmueller: This looks like a syntax / program flow change. Was that intended?

noc0lour: if it is a real change no, but most importantly this piece of code does not run without fixing the mixed tab/space indentation (python will even refuse to import the file)

noc0lour:
@mbr0wn I looked at the code before `old_div` was put in place and tried to figure out which `div` might be appropriate and makes sense in that piece of code. 
E.g. `delay` wants to have a whole number, but `bw` has to be a float/fraction. Also Py2k can be forced to do real division by having one float in it. 

The `old_div` code was broken and we don't have a dependency on the `future` module (and really don't need it). 
Can you review at least part of the changes if they make sense?

Issue #1944, opened 2018-08-06
gr-uhd: usrp_sink_impl::start breaks two first bursts


ptrkrysik:
When transmitting bursts (either having time_spec attached or not) there are problems with transmission of two first bursts:
* just after the first burst there is 'Underrun' reported by UHD,
* the second burst is transmitted just after the first one or not transmitted at all (depending if it is in USRP's buffer or not when the 'Underrun' happens):
https://imgur.com/a/8OUzSSa

I investigated the issue and found out that it is caused by 'send' function call in usrp_sink_impl::start:
https://github.com/gnuradio/gnuradio/blob/219eae9a9c2ef7644450e71d19f8f54c12e1f9cc/gr-uhd/lib/usrp_sink_impl.cc#L611

It is probably needed for stream transmission but it doesn't play well with bursty transmission.
Seems that executing this function only when 'pmt::is_null(_length_tag_key)' is true (which means we want to work with streams) would fix the problem.

But it would be good to look closely at what usrp_sink_impl::start function does. In example it can set '_metadata.has_time_spec' to 'false' and then set _metadata.time_spec value anyway:

https://github.com/gnuradio/gnuradio/blob/219eae9a9c2ef7644450e71d19f8f54c12e1f9cc/gr-uhd/lib/usrp_sink_impl.cc#L607

which doesn't seem to be reasonable and it might indicate that quality of this function need improvement in general.

I produced two minimal examples presenting the issue:
* using GNU Radio and transmitting 10 bursts (https://github.com/ptrkrysik/usrp_sink_burst_tx_bug/tree/master/minimal_example_gnuradio),
* using only UHD and mimicking what is done in the usrp_sink::start - with the same result (Underrun) (https://github.com/ptrkrysik/usrp_sink_burst_tx_bug/blob/master/minimal_example_uhd_only/tx_single_burst_like_gnuradio.cpp)

There is also one example showing that disabling superfluous 'send' call fixes the issue (https://github.com/ptrkrysik/usrp_sink_burst_tx_bug/blob/master/minimal_example_uhd_only/tx_single_burst.cpp).

mbr0wn:
Thanks for reporting this. Also, which USRP were you using? You're right we need to carefully review this. I think the gist of your PR is fine.

ptrkrysik:
I observed the same effect for USRP B210 and X310.

mbr0wn:
Thanks, that's helpful.

dkozel:
[...]

PR #1947, opened 2018-08-10
gr-uhd: fix first bursts transmitted by usrp_sink

ptrkrysik:
This commit is intended to fix issue #1944, where superfluous
call to UHD's 'send' function causes problems with beginning
of bursty transmission.

Care is needed during the review as I changed some part
that seemed a bit strange (more in #1944) - so there is slight possibility
that maybe I didn't understand intention of the author.

Similar change will be needed also in the usrp_sink_impl::stop
function. However there it will be a bit more complicated as when stop
is called in the middle of the burst, the empty message ending the
transmission should actually be sent.

### DIFF
--- a/gr-uhd/lib/usrp_sink_impl.cc
+++ b/gr-uhd/lib/usrp_sink_impl.cc
@@ -593,28 +593,33 @@ namespace gr {
       if (not _tx_stream)
         _tx_stream = _dev->get_tx_stream(_stream_args);
 #endif
-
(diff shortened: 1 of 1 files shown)


mbr0wn:
I'm pretty certain this will cause merge conflicts with `next` (this is not your fault). @marcusmueller, would it make sense to rebase this onto next, fix it there, and then have it when we merge next into master?

Issue #1937, opened 2018-08-01
grc 3.8 requires the xterm variable to be set 


dl1ksv:
Trying to execute a flowgraph from grc requires the xterm variable to be set even if the variable is set.

Application.py line 690 -699

                Actions.FLOW_GRAPH_GEN()
                xterm = self.platform.config.xterm_executable
                Dialogs.show_missing_xterm(main, xterm)
                if self.config.xterm_missing() != xterm:
                    if not os.path.exists(xterm):
                        Dialogs.show_missing_xterm(main, xterm)
                    self.config.xterm_missing(xterm)

Dialogs.show_missing_xterm(main, xterm) is called before xterm is checked.

So just remove line 692.

noc0lour:
:+1: thanks for the report. sounds like a bug.

marcusmueller:
Closed via #1937, right @dl1ksv ?

PR #1953, opened 2018-08-15
Update Application.py

dl1ksv:
Fixing 
     NameError: name 'SaveScreenShotDialog' is not defined
in Application.py

and removing unneeded call to set xterm .

See [#1937](https://github.com/gnuradio/gnuradio/issues/1937)


marcusmueller:
Nice one! Thanks!

PR #1954, opened 2018-08-16
message: fix unicode behaviour in Py2k/Py3k

noc0lour:
This PR is the second half of #1940
Due to quite different topics it's a separate PR now.


noc0lour:
Looks like qa_message is not failing anymore :tada:

PR #1961, opened 2018-08-18
Get rid of MSVC impl of gettimeofday that's used nowhere

### LINKED ISSUE #1594: get rid of `gettimeofday` implementation
... we don't need that anywhere. Also, @guruofquality on #1586 indicates there might be license trouble around that code (even if Apache Thrift [uses](https://github.com/apache/thrift/blob/master/lib/cpp/src/thrift/windows/GetTimeOfDay.cpp) the very same code and didn't mention the [original source](https://social.msdn.microsoft.com/Forums/vstudio/en-US/430449b3-f6dd-4e18-84de-eebd26a8d668/gettimeofday?forum=vcgeneral), and Thrift, as we all know, is code of the highest usability).

marcusmueller:
and is of unknown origin and quality.

Closes #1594.

### DIFF
--- a/cmake/msvc/sys/time.h
+++ b/cmake/msvc/sys/time.h
@@ -5,67 +5,13 @@
 #ifndef _MSC_SYS_TIME_H_
 #define _MSC_SYS_TIME_H_
 
-//http://social.msdn.microsoft.com/Forums/en/vcgeneral/thread/430449b3-f6dd-4e18-84de-eebd26a8d668
 #include < time.h >
-#include <windows.h> //I've omitted this line.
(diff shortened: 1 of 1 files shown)


marcusmueller:
@gnieboer I'm always a bit timid when touching MSVC-only headers in GNU Radio, since I feel like I'm touching a system I couldn't rebuild. Could you assess these changes?

marcusmueller:
I'm taking a bit of a chance here by merging this, but this change looks really benign.

gnieboer:
Apologies I was away.  Yes, change looks benign.  I will keep an eye out for any issues when I do the next build.

PR #1963, opened 2018-08-18
Fixed out-of-bound vector acces in PMT (pre-C++11 version)

marcusmueller:
empty `vec`.

This commit, unlike the similar commit of the same message, uses `NULL`
in case of `nullptr` (maint-3.7 doesn't require C++11).

This will lead to merge conflicts of `NULL` vs `nullptr` on merging of next.

These are easy to resolve. Use `nullptr` when merging next, becaus that's the
moment we can rely on C++11.


marcusmueller:
Since this is already merged on maint-3.7, this will take the expressway to mergeland once the CI has had its way with it.

Issue #1874, opened 2018-07-05
next branch! grc yml nconnections definitions of qtgui's conflict with save_flow_graph in platform.py


dl1ksv:
grc is not able to generate flowgraphs containing qtgui instrumentation windows.
This is due to
platform.py containing the following replacements (line 359 -365 ) when saving grc files.
      replace = [
            ('blocks:', '\nblocks:'),
            ('connections:', '\nconnections:'),
            ('metadata:', '\nmetadata:'),
        ]
      for r in replace:
            out = out.replace(*r)

This results in
nconnections: 1
 mapped to
n
connections: 1

which is syntactilly wrong.
Proposal to fix:
Replace nconnections by ninputs in gr-qtgui/grc/*.yml

dkozel:
@dl1ksv Thanks for the report.

@haakov @skoslowski Thoughts?

haakov:
Thanks for the report! There is, however, already a pull request with a fix for this bug. See [this](https://github.com/gnuradio/gnuradio/pull/1481/commits/d4eada228d5e513e033814b3cecc9e9657debbd8) commit :)

dl1ksv:
Yes, this pull request fixes this bug. 
Continuing and trying to generate the flowgraph leads to the following:

[...]

PR #1964, opened 2018-08-18
Use proper relative imports in gr-audio's __init__.py

marcusmueller:
This adresses parts of what @dl1kv noticed in #1874


marcusmueller:
Since this is identical to what we do in all other `__init__.py`s, merging this without a second set of eyes.

PR #1965, opened 2018-08-19
Executor.py: decode stdout as UTF-8

dl1ksv:
If the executed script sends some ( error ) messages you get a lot of error messages like

      File "/usr/local/gnuradio/lib/python3.6/dist-packages/gnuradio/grc/core/Messages.py", line 117, in send_verbose_exec
        send(verbose)
       File "/usr/local/gnuradio/lib/python3.6/dist-packages/gnuradio/grc/core/Messages.py", line 52, in send
    messenger(_indent + message)
    TypeError: must be str, not bytes

But you don't find out the reason of this error. The return type of os.read seems to have changed between python 2.7 and python 3.
The decode function fixes this issue and you get valuable messages.

### DIFF
--- a/grc/gui/Executor.py
+++ b/grc/gui/Executor.py
@@ -83,7 +83,7 @@ class ExecFlowGraphThread(threading.Thread):
         r = "\n"
         while r:
             GLib.idle_add(Messages.send_verbose_exec, r)
-            r = os.read(self.process.stdout.fileno(), 1024)
+            r = self.process.stdout.read(1024).decode('utf-8','ignore')
         self.process.poll()
         GLib.idle_add(self.done)
 


noc0lour:
The error stems from os read being `bytes` in both Python2 and Python3. But in Python2 the `str` and `bytes` type is the same. In Python3 the we need unicode which is not `bytes`. 

Using stdout.read() will always return a string-type value.

marcusmueller:
Agreeing here (Popen.stdout.read()). I like @dl1ksv patch, but I'll supplement my own in this PR.

marcusmueller:
hold that thought.

@noc0lour, you're wrong! (this surprises me as much as it surprises you):
```python
import subprocess
p = subprocess.Popen("yes", stdout=subprocess.PIPE)
x = p.read(1000)
type(x)
>> bytes
```

so, no, albeit this being far more logical than calling `os.read()` it doesn't solve the issue.

marcusmueller:
So, @dl1ksv, to explain the error first: we send what we read from the Popen process' `stdout` to a kind of message distributor. In `grc/core/Messages.py`, we register `sys.stdout` as acceptor of such messages.

So, in Python3, `sys.stdout` is different from a `Popen.stdout` and `stdin` in that `sys.stdout.write` only accepts strings, not `bytes` (which *really* makes no sense to me as coming from the OS side of things... but whatevs). However, Py3's Popen.stdout (imho correctly) gives `bytes` when read – nothing says that stdio needs to be text.

So, yes, we should be doing both:

* switch from `os.read(…)` to `self.process.stdout.read()`
* decode the result

Since you clearly did the important thing here, I'll just be modifying your patch to be "pretty" and keep you as the author.

marcusmueller:
@dl1ksv can you look at the changes and confirm they're OK to commit under your name, please?

I'll also change the title of your commit message to describe what you did rather than the filename – the file name is part of the commit's changes, what's interesting is to understand what that commit does. In this case, I'd change the title to "Executor: Decode data read from fg process using utf-8"; but only if that's OK for you, you're the author of the original commit.

dl1ksv:
@marcusmueller  I just compared the result of both solutions and they don't differ much.

Here is the result of 

r = self.process.stdout.read(1024)

    Generating: '/home/schroer/gnuradiocomponents/gr-fcdproplus/example/FMRadio.py'

    Executing: /usr/bin/python3 -u /home/schroer/gnuradiocomponents/gr-fcdproplus/example/FMRadio.py
    /usr/lib/python3.6/subprocess.py:596: ResourceWarning: unclosed file <_io.BufferedWriter name=9>
     _cleanup()
    /usr/lib/python3.6/subprocess.py:596: ResourceWarning: unclosed file <_io.BufferedReader name=10>
      _cleanup()
    /usr/lib/python3.6/subprocess.py:596: ResourceWarning: unclosed file <_io.BufferedReader name=12>
      _cleanup()

    gr::log :INFO: audio source - Audio source arch: alsa
    ALSA lib pcm_hw.c:1820:(_snd_pcm_hw_open) Invalid value for card
    gr::log :ERROR: audio_alsa_source0 - [hw:2,0]: Datei oder Verzeichnis nicht gefunden
    Could not open device: hw:2,0
    gr::log :INFO: fcdproplus0 - Could not open device: hw:2,0
      File "/home/schroer/gnuradiocomponents/gr-fcdproplus/example/FMRadio.py", line 296, in <module>
       main()
      File "/home/schroer/gnuradiocomponents/gr-fcdproplus/example/FMRadio.py", line 284, in main
       tb = top_block_cls()
     File "/home/schroer/gnuradiocomponents/gr-fcdproplus/example/FMRadio.py", line 148, in __init__
        self.fcdproplus_fcdproplus_0 = fcdproplus.fcdproplus('hw:2,0',1000)
      File "/usr/local/gnuradio/lib/python3.6/dist-packages/fcdproplus/fcdproplus_swig.py", line 126, in 
    make
        return _fcdproplus_swig.fcdproplus_make(*args, **kwargs)
    RuntimeError: No FunCube Dongle  V2.0 found.

   >>> Done (return code 1)
      /usr/local/gnuradio/lib/python3.6/dist-packages/gnuradio/grc/main.py:93: ResourceWarning: unclosed file <_io.TextIOWrapper name=9 encoding='UTF-8'>
      sys.exit(app.run())

And with  r = os.read(self.process.stdout.fileno(), 1024).decode('utf-8') you get

    Generating: '/home/schroer/gnuradiocomponents/gr-fcdproplus/example/FMRadio.py'

    Executing: /usr/bin/python3 -u /home/schroer/gnuradiocomponents/gr-fcdproplus/example/FMRadio.py
    /usr/lib/python3.6/subprocess.py:596: ResourceWarning: unclosed file <_io.BufferedWriter name=9>
      _cleanup()
    /usr/lib/python3.6/subprocess.py:596: ResourceWarning: unclosed file <_io.BufferedReader name=11>
      _cleanup()
    /usr/lib/python3.6/subprocess.py:596: ResourceWarning: unclosed file <_io.BufferedReader name=13>
      _cleanup()

    gr::log :INFO: audio source - Audio source arch: alsa
    ALSA lib pcm_hw.c:1820:(_snd_pcm_hw_open) Invalid value for card
    gr::log :ERROR: audio_alsa_source0 - [hw:2,0]: Datei oder Verzeichnis nicht gefunden
    Could not open device: hw:2,0
    gr::log :INFO: fcdproplus0 - Could not open device: hw:2,0
      File "/home/schroer/gnuradiocomponents/gr-fcdproplus/example/FMRadio.py", line 296, in <module>
        main()
     File "/home/schroer/gnuradiocomponents/gr-fcdproplus/example/FMRadio.py", line 284, in main
       tb = top_block_cls()
      File "/home/schroer/gnuradiocomponents/gr-fcdproplus/example/FMRadio.py", line 148, in __init__
        self.fcdproplus_fcdproplus_0 = fcdproplus.fcdproplus('hw:2,0',1000)
      File "/usr/local/gnuradio/lib/python3.6/dist-packages/fcdproplus/fcdproplus_swig.py", line 126, in 
    make
        return _fcdproplus_swig.fcdproplus_make(*args, **kwargs)
    RuntimeError: No FunCube Dongle  V2.0 found.

   >>> Done
   /usr/lib/python3.6/subprocess.py:766: ResourceWarning: subprocess 9903 is still running
    ResourceWarning, source=self)
  /usr/local/gnuradio/lib/python3.6/dist-packages/gnuradio/grc/main.py:93: ResourceWarning: unclosed file <_io.TextIOWrapper name=9 encoding='UTF-8'>
     sys.exit(app.run())

So the only difference is the report of the return code. This may be usefull if the code  will be interpreted

But I'm wondering about the resource warnings. I'm unsure if I saw them before. But can we ignore them. I prefer to avoid them. Any ideas ?

dl1ksv:
Our postings just crossed. The changes are ok for me.
But what about the resource warnings ? I'll try to figure out.

dl1ksv:
At this point I'm not able to find the reason for the resource warnings. I tested with different flow graphs and the warnings make no trouble.

I see similar warnings on startup and closing grc, without noticing any problems.

So I think this fix could be merged.

marcusmueller:
I was just about to merge this, but then it hit me:

Nothing says the stdout of a process needs to be validly utf-8 decodable. @noc0lour, what do we do?

dl1ksv:
@marcusmueller The additional parameter in decode avoids this problem.

marcusmueller:
I still feel like maybe one day we should rewrite this whole console thing (seriously, printing to the console eats CPU like a hungry goat eats biscuits from a schoolkid's bagpack), but yeah, since we'll be only printing things that hopefully are actually text... let's do this!

argilo:
This seems to have broken things. I get the following when I try to run a flowgraph:
```
Exception in thread Thread-2:
  File "/usr/lib/python3.6/threading.py", line 916, in _bootstrap_inner
    self.run()
  File "/home/argilo/prefix_next/lib/python3.6/dist-packages/gnuradio/grc/gui/Executor.py", line 86, in run
    r = self.process.stdout.read(1024).decode('utf-8','ignore')
AttributeError: 'str' object has no attribute 'decode'
```
Removing the `decode` fixes it.

According to the Python documentation, using `universal_newlines=True` causes stdout to be opened in text mode.

dl1ksv:
See discussion in [https://github.com/gnuradio/gnuradio/pull/1989](url)

Should be solved if [pr2024]https://github.com/gnuradio/gnuradio/pull/2024l)

is applied.

PR #1967, opened 2018-08-19
Explicitly set the size of the gengen templates

marcusmueller:
This leads to the following typemap:

s: int16_t
i: int32_t
f: float
c: gr_complex
b: uint8_t

### DIFF
--- a/gnuradio-runtime/python/build_utils_codes.py
+++ b/gnuradio-runtime/python/build_utils_codes.py
@@ -1,6 +1,6 @@
 from __future__ import unicode_literals
 #
-# Copyright 2004 Free Software Foundation, Inc.
+# Copyright 2004,2018 Free Software Foundation, Inc.
 #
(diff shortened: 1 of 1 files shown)


marcusmueller:
@noc0lour certainly, we can, but I'd actually find it cooler if your templates had types like `int32_t` instead of `int`, and `int8_t` instead of `char`. I find the explicit naming of signedness and size a desirable quality here. But yes, definitely, let's hold this off.

noc0lour:
Oh, I'm totally with y'all on the type side. (In the gengen removal PR the gr-analog stuff has the "old" type definitions in it, but I'll update them as well).

michaelld:
@marcusmueller For my knowledge / edification: what is the "swig vector<char> confusion" of which you write?

marcusmueller:
Since @noc0lour's anti-gengen PR does this, closing.

PR #1972, opened 2018-08-23
Force to update all affected elements

dl1ksv:
On a parameter change of block , the block will not be updated correctly.

Example:
     Open a new Flowgraph tab.
     Add an  math add block. The default of the io ports is complex.
     Open the property dialog and change the type of the io ports from complex to float and select OK.
     Only the first input and the output port change from 'complex' to 'float'. The second input port remains 'complex'.
     The only way to change the type of the second input port is to reopen the property dialog and select quit !!!!
     Now the second port has the correct type.

The proposed change fixes this issue. But I'm not sure if there isn't a more elegant way to handle this issue.

### DIFF
--- a/grc/gui/Application.py
+++ b/grc/gui/Application.py
@@ -542,8 +542,12 @@ class Application(Gtk.Application):
                 while response == Gtk.ResponseType.APPLY:  # rerun the dialog if Apply was hit
                     response = self.dialog.run()
                     if response in (Gtk.ResponseType.APPLY, Gtk.ResponseType.ACCEPT):
-                        flow_graph_update()
                         page.state_cache.save_new_state(flow_graph.export_data())
+                        ### Following  lines force an complete update of io ports
+                        n = page.state_cache.get_current_state()
+                        flow_graph.import_data(n)
+                        flow_graph_update()
+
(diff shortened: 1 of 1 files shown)


grc/gui/Application.py
```diff
@@ -542,8 +542,12 @@ def flow_graph_update(fg=flow_graph):
                 while response == Gtk.ResponseType.APPLY:  # rerun the dialog if Apply was hit
                     response = self.dialog.run()
                     if response in (Gtk.ResponseType.APPLY, Gtk.ResponseType.ACCEPT):
-                        flow_graph_update()
                         page.state_cache.save_new_state(flow_graph.export_data())
+                        ### Following  lines force an complete update of io ports
```
marcusmueller: Updating all ports on param modification is, at the very least, a safe choice! Nice, thank you.

One question: Why did this migrate from before to after the state saving?

dl1ksv: I just immated the action after 'quit' as this was the only way to update all ports. Only moving the flow_graph_update 
behind
    page.state_cache.save_new_state(flow_graph.export_data())

didn't make it.
But the combination of the three lines did it !

PR #1978, opened 2018-08-24
Replace scipy/pylab where numpy/pyplot is sufficient

marcusmueller:
This should reduce the number of times users are prompted to install
pylab || scipy when they'd actually get away with functionality fully
contained in numpy and matplotlib.

This only solves the obvious cases. There's some usage of `pylab.mlab`
that would need more than 20s of consideration.

### DIFF
--- a/docs/doxygen/other/pfb_intro.dox
+++ b/docs/doxygen/other/pfb_intro.dox
@@ -96,7 +96,7 @@ Hz. The code then plots the PSD of the original signal to see the
 signals in the origina spectrum and then makes 9 plots for each of the
 channels.
 
-NOTE: you need the Scipy and Matplotlib Python modules installed to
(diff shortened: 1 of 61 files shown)


marcusmueller:
Closes 1611.

marcusmueller:
The main user benefit is that you don't need scipy anymore to run any of the plot utils

marcusmueller:
@mbr0wn hey, thanks! But in this case, I'd like someone to at least punctually run a few of the affected programs. I tried the same, but I inherently had some systematicality in choosing the samples. So, since @balister in #coredevs said he was fond of this PR, I thought he was reviewing that.

marcusmueller:
PR matured enough. Merging, well aware of residual risk.

PR #1980, opened 2018-08-24
remove deleted mpsk_demod.grc from install list

marcusmueller:
This fixes what 2345321 missed (remove the mpsk_demod from the installation
list)


marcusmueller:
non-dangerous hotfix. Merging.

PR #1982, opened 2018-08-25
Remove examples depending on removed dxpsk_(de)mod

marcusmueller:


michaelld:
A quick `grep -lr dxpsk` in GR next reveals that these 3 files are the only ones of interest, so I think this PR does the right thing. LGTM, if that counts for anything ;)

marcusmueller:
@michaelld yes, that counts for something :)

In fact, it counts as a review :+1:

PR #1984, opened 2018-08-25
Removed unused `malloc16Aligned` and consorts

marcusmueller:
This was Code by the great Phil Karn, but whoever needs aligned memory
will go through volk or their OS' tools, anyways. None of this code was
used in-tree anywhere.

Also removing the posix_memalign implementation, which the CMakeFile itself says has no known necessity, but is only included on PowerPC.  Taking that chance.

`Complex_vec_test` is part of libgnuradio, but it's not used anywhere (especially not in a unit test that I could find).


marcusmueller:
Also removing unused headers from being checked at CMake time.

michaelld:
OK so GR next doesn't even use `posix_memalign` nor any of the `16Align` functions ... so it might with on older OS / PPC out of the box. We'll see once this commit is in place!

marcusmueller:
hey, thanks @michaelld! So, would you prefer I wait till Monday? Can do that :) But honestly, since it's not used, I'd expect the build to break for other reasons.

I'll take your "should be two separate things" criticism seriously – let me merge that in two separate merges, you're right, it's clearer and we can more easily pinpoint the fault if something breaks later on :+1:

marcusmueller:
I've split these out to #1988

michaelld:
Awesome; thx! L(really)GTM now!

PR #1989, opened 2018-08-27
Removing unneeded decode

dl1ksv:
There was a misunderstanding in [pr 1965]https://github.com/gnuradio/gnuradio/pull/1965l)
No we get:
      File "/usr/local/gnuradio/lib/python3.6/dist-packages/gnuradio/grc/gui/Executor.py", line 86, in run
         r = self.process.stdout.read(1024).decode('utf-8','ignore')
     AttributeError: 'str' object has no attribute 'decode'

self.process.stdout.read is a string object and must not be decoded.

os.read(self.process.stdout.fileno(), 1024) delivers bytes that have to be decoded to strings.

After removing the decode everything works fine.

### DIFF
--- a/grc/gui/Executor.py
+++ b/grc/gui/Executor.py
@@ -83,7 +83,7 @@ class ExecFlowGraphThread(threading.Thread):
         r = "\n"
         while r:
             GLib.idle_add(Messages.send_verbose_exec, r)
-            r = self.process.stdout.read(1024).decode('utf-8','ignore')
+            r = self.process.stdout.read(1024)
         self.process.poll()
         GLib.idle_add(self.done)
 


marcusmueller:
OK, so, under which python did you test this new code? I really thought we had this sorted out :( But I seem to have messed something up?

marcusmueller:
seriously, I just ran

```
ipython3
import subprocess
p = subprocess.Popen(args=["yes"], stdout=subprocess.PIPE)
x = p.stdout.read(100)
print (type(x))
> bytes
```

so, even if this bug fixes something, I really don't think we're pushing `str` about here, but `bytes`. And that would need decoding to be printed.

dl1ksv:
@marcusmueller Marcus, now I got it.

If you add universal_newlines=True to your Popen call  like is done in Executor.py the type of x is str.

This confirms to the python documentation.

See [Popen.stdout](https://docs.python.org/3/library/subprocess.html#popen-objects)

mfischer:
I know I haven't been around for a while, so not trying to step on anyone's feet here :)
If the commit message would have a reference to the commit say a

```Fixes commit hash ("Commit message")``` it allows me to figure these things out without pulling up a browser to lookup stuff on github, by just looking at git log.
For those interested in the kernel we do it as described in [1].

Again, not trying to step on anyone's feet here, just sharing what works well for us in the kernel. Added bonus you can pull metrics on how long something was broken ;-)

[1] https://www.kernel.org/doc/html/v4.17/process/submitting-patches.html

dl1ksv:
Thank you for your recommendation. I understand your concern and I'll try to follow in my next pr.

dl1ksv:
Setting bufsize=0 does not change anything.

Changing the read call to

 r = self.process.stdout.read(1)

did it.

So I made a new [pr2024](https://github.com/gnuradio/gnuradio/pull/2024) which replaces this pr

PR #1992, opened 2018-08-27
Trying to merge @michaelld's PR

marcusmueller:
cmake: fix CTRLPORT includes and testing

Currently, CTRLPORT_BACKENDS is not declared if ENABLE_GR_CTRLPORT is FALSE, so testing this variable makes no sense. All testing in gr-* is done using ENABLE_GR_CTRLPORT, so move to that in gr-blocks.

That said, having CTRLPORT_BACKENDS set to 0 could be useful, so rearrange CTRLPORT determination to set this variable correctly.

If ENABLE_CTRLPORT_THRIFT, then correctly use include_directories for header directories, -after- CTRLPORT is determined, not before. We can also combine some of these declarations to simplify the code.

### DIFF
--- a/gnuradio-runtime/lib/controlport/CMakeLists.txt
+++ b/gnuradio-runtime/lib/controlport/CMakeLists.txt
@@ -17,15 +17,18 @@
 # the Free Software Foundation, Inc., 51 Franklin Street,
 # Boston, MA 02110-1301, USA.
 
-if(ENABLE_GR_CTRLPORT)
-
 # Keep track of the number of backends ControlPort supports
 SET(CTRLPORT_BACKENDS 0)
+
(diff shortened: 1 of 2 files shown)


michaelld:
Heck yeah! Sorry I forgot about this PR... I'm good with using it if it's been updated for the current (post-next-merge) master ... which it looks like it is ... yay!

marcusmueller:
@michaelld now, on all builders, the cpp bindings qa fail – is this expected?

michaelld:
I'm redoing this PR, since it has issues. I think there was another PR on this topic -- which I closed. I'll do the same here, and just move to a new PR with the fixed code once I get there.

PR #2000, opened 2018-08-31
Add missing header file (fir_filter_blk.h)

carlesfernandez:
Install the missing header ```fir_filter_blk.h```, so other modules using these blocks can be easily updated.


noc0lour:
I apparently overlooked that, thanks!

marcusmueller:
superseded and implemented by #2038

PR #2019, opened 2018-09-19
Use site-packages vs dist-packages for python libs

### LINKED ISSUE #2006: dist-packages is Debian specific and not a general "unix thing"

tomberek:
Fixes #2006

This also helps on non-debian packaging.

### DIFF
--- a/cmake/Modules/GrPython.cmake
+++ b/cmake/Modules/GrPython.cmake
@@ -96,7 +96,7 @@ execute_process(COMMAND ${PYTHON_EXECUTABLE} -c "
 import os
 import sys
 if os.name == 'posix':
-    print(os.path.join('lib', 'python' + sys.version[:3], 'dist-packages'))
+    print(os.path.join('lib', 'python' + sys.version[:3], 'site-packages'))
 if os.name == 'nt':
     print(os.path.join('Lib', 'site-packages'))
 " OUTPUT_VARIABLE GR_PYTHON_DIR OUTPUT_STRIP_TRAILING_WHITESPACE


noc0lour:
Can't we go back to using the information from python distutils?

tomberek:
That would work too. I'm not sure exactly why that section was changed.
@djanderson ?

djanderson:
My commit message for 7c32a9220 says I made that change so that install would also work inside a prefix. My memory is fuzzy but I seem to recall that I had some issues with the default install location of python modules when installing inside a prefix. 

The original code was actually:

```python
from distutils import sysconfig
print sysconfig.get_python_lib(plat_specific=True, prefix='')
``` 

That produces a relative path that I think I was actually not valid for python3 on my system. I seem to recall that py2 virtual envs had a `lib/python2.7/site-packages` but py3 had only `lib/python3/site-packages` without the subversion number. I'm looking at virtualenvs on my Ubuntu 18.04 system right now and that is no longer the case, but it's worth double checking on Ubuntu 16.04 (which I would have been using at the time) if we're going to support that OS just to make sure.

The output of
```python
import site
print(site.getsitepackages()[0])
```
was correct on my system for py3, but it appears that the function `getsitepackages` was not part of the py2 `site` module. So we can't just go back to that.

My guess with the current solution (since it has windows stuff as well) is that I either found that snippet online or in another large project, but I'm not sure where.

So, unfortunately it may or may not be just as simple a fix as this PR. We'd have to test on a few systems in and out of a prefix. At the time I was testing py2 and py3, in and out of prefixes, but obviously I missed non-Ubuntu systems here. :/

djanderson:
So looking a bit more, on my Ubuntu 18.04 system, I have:

```bash
$ ls /usr/lib/ |grep python
python2.7
python3
python3.6
python3.7
```
Notice there's no `python2` directory, but there is a `python3` directory, which contains a `dist-packages` subdir. Python stuff installed by `apt` is all installed in there (pybombs for example will install python dependencies via `apt`, not `pip`), and I think the issue was that with the previous code, all the python dependencies were being installed in `/usr/lib/python3/dist-packages` but gnuradio itself was being installed in `/usr/lib/python3.*/site-packages`, and there was some issue there.

marcusmueller:
>  but it appears that the function getsitepackages was not part of the py2 site module. So we can't just go back to that.

We can! 

Quoting the Python Docs:

> site.getsitepackages()
> Return a list containing all global site-packages directories (and possibly site-python).
> 
> New in version 2.7.

We don't support anything < 2.7 anymore so that is fine.

We've got a superseeding PR #2109, so although this PR is definitely a helpful improvement over the current state, I'm opting to use that.

PR #2024, opened 2018-09-21
Removing unneeded decode.

dl1ksv:
This pr replaces [https://github.com/gnuradio/gnuradio/pull/1989](url)

Modifing read to make sure messages are printed at once.

Tested on python2.7 and python3.6

### DIFF
--- a/grc/gui/Executor.py
+++ b/grc/gui/Executor.py
@@ -83,12 +83,13 @@ class ExecFlowGraphThread(threading.Thread):
         r = "\n"
         while r:
             GLib.idle_add(Messages.send_verbose_exec, r)
-            r = self.process.stdout.read(1024).decode('utf-8','ignore')
+            r = self.process.stdout.read(1)
         self.process.poll()
         GLib.idle_add(self.done)
 
     def done(self):
         """Perform end of execution tasks."""
         Messages.send_end_exec(self.process.returncode)
+        self.page.process.wait()
         self.page.process = None
         self.update_callback()


777arc:
This PR fixes a bug that prevents me from running *any* flowgraph in 3.8, so I'm 100% for it.

phansel:
This issue remains unresolved in Python 3.5.2 as of abd594ae. Buildbot is observing some timeout failures on [qa_constellation](https://github.com/gnuradio/gnuradio/blob/master/gr-digital/python/digital/qa_constellation.py) when this patch is applied, but it builds fine for me on 16.04. Can anyone explain why a change in the flowgraph executor code would affect a DSP test like qa_constellation?

noc0lour:
@phansel in general `qa_constellation` takes quite some time to execute due to building full constellation tables in python. Depending on the load on the build machine the test might hit the timeout limit.

btashton:
Can we get this merged in soon, I keep having to apply this fix to work on other bugs.

777arc:
Yeah same

btashton:
I would actually like to recommend #2150 over this one.  It has some additions around the same code to deal with thread termination and cleaning up the thread resources.

I'm happy to split this one line off if people want to keep this PR, but there is a bug in this PR that could cause a deadlock.
`self.page.process.wait()` This cannot be used with `stdout=subprocess.PIPE`  see https://docs.python.org/2/library/subprocess.html#subprocess.Popen.wait

PR #2038, opened 2018-09-24
gengen replacement: fix for not installed header files

lukaskuzmiak:
Some header files are not being installed, seems they were forgotten during gengen replacement, here: https://github.com/gnuradio/gnuradio/commit/1043222e0de551f00683a1cc9060f194adc4fc85 and https://github.com/gnuradio/gnuradio/commit/7588f103b9e75eea56d241868d161d0f04f3fd79

Perhaps @noc0lour will be able to tell quickly if this is correct.

I noticed this when fixing gqrx to compile against latest master.

### DIFF
--- a/gr-blocks/include/gnuradio/blocks/CMakeLists.txt
+++ b/gr-blocks/include/gnuradio/blocks/CMakeLists.txt
@@ -25,6 +25,7 @@ install(FILES
     add_blk.h
     add_const_v.h
     and_blk.h
+    and_const.h
     api.h
     argmax.h
     control_loop.h
@@ -73,6 +74,7 @@ install(FILES
     char_to_short.h
     check_lfsr_32k_s.h
     complex_to_interleaved_short.h
+    complex_to_interleaved_char.h
     complex_to_float.h
     complex_to_magphase.h
     complex_to_imag.h
--- a/gr-filter/include/gnuradio/filter/CMakeLists.txt
+++ b/gr-filter/include/gnuradio/filter/CMakeLists.txt
@@ -24,10 +24,12 @@ install(FILES
     api.h
     firdes.h
     fir_filter.h
+    fir_filter_blk.h
     fir_filter_with_buffer.h
     fft_filter.h
     iir_filter.h
     interpolator_taps.h
+    interp_fir_filter.h
     mmse_fir_interpolator_cc.h
     mmse_fir_interpolator_ff.h
     mmse_interp_differentiator_cc.h


lukaskuzmiak:
I did check other headers too, wrote a little script for that - it matches `CMakeLists.txt` to actual file list, this is the output:

```
$ ./check_includes.sh 
processing gnuradio-runtime/include/gnuradio/messages/ ...
processing gnuradio-runtime/include/gnuradio/thread/ ...
processing gr-analog/include/gnuradio/analog/ ...
processing gr-audio/include/gnuradio/audio/ ...
0a1,4
> api.h
> osx_impl.h
> sink.h
> source.h
processing gr-blocks/include/gnuradio/blocks/ ...
processing gr-channels/include/gnuradio/channels/ ...
1a2
> cfo_model.h
3a5
> dynamic_channel_model.h
6a9
> sro_model.h
processing gr-comedi/include/gnuradio/comedi/ ...
processing gr-digital/include/gnuradio/digital/ ...
processing gr-dtv/include/gnuradio/dtv/ ...
processing gr-fec/include/gnuradio/fec/ ...
processing gr-fft/include/gnuradio/fft/ ...
processing gr-filter/include/gnuradio/filter/ ...
22a23
> interp_differentiator_taps.h
processing gr-qtgui/include/gnuradio/qtgui/ ...
4a5
> SpectrumGUIClass.h
29a31
> spectrumdisplayform.h
processing gr-trellis/include/gnuradio/trellis/ ...
processing gr-uhd/include/gnuradio/uhd/ ...
processing gr-video-sdl/include/gnuradio/video_sdl/ ...
processing gr-vocoder/include/gnuradio/vocoder/ ...
processing gr-wavelet/include/gnuradio/wavelet/ ...
processing gr-zeromq/include/gnuradio/zeromq/ ...
```

The `gr-audio` is incorrect, it uses different format for the `CMakeLists.txt`. The rest is whatever is left after my commits in the pull request. I went back to your gengen changes and none of these seem to have been affected by it so i just left them alone as they were not installed before maybe that was intentional but I'm not able to judge that very well. So anyway if any of these seem relevant to you those are the differences.

bastibl:
I also ran into this problem when trying to port my modules. It seems to be common. There is another open PR that would be superseded by this ones (https://github.com/gnuradio/gnuradio/pull/2000).

Is there a reason to not just use `glob *.h`? These unsorted and manually maintained lists are somewhat error-prone.

noc0lour:
afaik using `glob` in CMake might lead to unexpected results if one adds/deletes header files since CMake won't detect changes without wiping the build dir or touching the appropriate `CMakeLists.txt`
```
 (We do not recommend using GLOB to collect a list of source files from your source tree. If no CMakeLists.txt file changes when a source is added or removed then the generated build system cannot know when to ask CMake to regenerate.)```

**Comment by bastibl** [comment]
You're right, it's not perfect. But same goes, for example, for all glob uses to install yml files.

```
>> grep -Ri "file(glob" * | grep yml
gr-analog/grc/CMakeLists.txt:file(GLOB yml_files "*.yml")
gr-audio/grc/CMakeLists.txt:file(GLOB yml_files "*.yml")
gr-blocks/grc/CMakeLists.txt:file(GLOB yml_files "*.yml")
grc/blocks/CMakeLists.txt:file(GLOB yml_files "*.yml")
gr-channels/grc/CMakeLists.txt:file(GLOB yml_files "*.yml")
gr-digital/grc/CMakeLists.txt:file(GLOB yml_files "*.yml")
gr-fec/grc/CMakeLists.txt:file(GLOB yml_files "*.yml")
gr-qtgui/grc/CMakeLists.txt:file(GLOB yml_files "*.yml")
```

And for headers, it would even be more relaxed, since they are only needed when an OOT wants to subclass or something.  (For block definitions it is more annoying, since blocks won't show up in GRC).

So, I think, it would only make sense to either remove all globs or use them everywhere. Maintaining header lists by hand and not having reliable rebuild without wiping the build dir sounds like the worst of both worlds.

marcusmueller:
Hey, thanks guys!

Yep, this was an oversight on my side. Yes, this supersedes #2000. I'll close that!

PR #2043, opened 2018-09-25
Widen the id column in the variable editor.

argilo:
The "Id" column in GRC's variable editor is a bit too narrow. I've increased the default width so that the `samp_rate` variable is fully visible in a new flowgraph.

Before:
screenshot from 2018-09-25 08-26-54

After:
screenshot from 2018-09-25 08-25-44


marcusmueller:
@noc0lour slight doubt in a table you'd want "magically jumping" widths. So merging this as more than a temporary fix ;)

PR #2047, opened 2018-09-26
Fix invalid escape sequences.

argilo:
Python produces the following warning when processing many of GNU Radio's .py files:

> DeprecationWarning: invalid escape sequence

I've fixed all these, mostly by using raw strings. In a couple cases, it was simpler to stick with an ordinary string and replace `\` with `\\`.

I compared the strings in Python before & after, so these changes shouldn't break anything.

### DIFF
--- a/docs/doxygen/other/doxypy.py
+++ b/docs/doxygen/other/doxypy.py
@@ -97,22 +97,22 @@ class Doxypy(object):
     def __init__(self):
         string_prefixes = "[uU]?[rR]?"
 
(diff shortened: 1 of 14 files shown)


argilo:
I tracked these down using the following command:

```
find . -name "*.py" | grep -v "^./build/" | grep -v "^./volk/" | xargs python3 -Wd -m py_compile 2>&1 | grep -A 1 "invalid escape sequence"
```

argilo:
Rebased to fix a merge conflict.

PR #2048, opened 2018-09-26
Convert tabs to spaces and fix syntax errors.

argilo:
Scanning Python code for deprecation warnings turned up a bunch of "Sorry: TabError: inconsistent use of tabs and spaces in indentation" and a couple "SyntaxError: EOL while scanning string literal". I've fixed those here. Together with #2047 this fixes all the deprecation warnings and errors identified by the following command:

```
find . -name "*.py" | grep -v "^./build/" | grep -v "^./volk/" | xargs python3 -Wd -m py_compile
```


marcusmueller:
Since this is a functional fix (and not just mere style): here we go! Thank you :)

PR #2054, opened 2018-09-27
Clean up resources allocated by run_worker.

argilo:
When GRC is started, a "subprocess is still running" warning appears, and a zombie process remains. This happens because `run_worker` terminates its subprocess, but does not call `wait()`.

https://github.com/gnuradio/gnuradio/blob/007923c64d2505b37dffb8e11cb1e3ecf4e489eb/grc/core/utils/extract_docs.py#L169

Adding a `wait()` call reveals further warnings because `run_worker` does not close the file handles it uses to communicate with the subprocess.

In this PR, I've added `close()` and `wait()` calls to eliminate the warnings.

The issue can also be reproduced by running `extract_docs.py` directly:

```
python3 -Wd grc/core/utils/extract_docs.py
```

### DIFF
--- a/grc/core/utils/extract_docs.py
+++ b/grc/core/utils/extract_docs.py
@@ -166,7 +166,11 @@ class SubprocessLoader(object):
                 break  # normal termination, return
             finally:
                 if self._worker:
+                    self._worker.stdin.close()
+                    self._worker.stdout.close()
+                    self._worker.stderr.close()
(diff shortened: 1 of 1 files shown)


noc0lour:
Looks good. Depending on the Targeted Python version we might consider moving this piece to multiprocessing after 3.8.

PR #2057, opened 2018-10-01
Replacing srandom with XOROSHIRO128+

marcusmueller:
This is a continuation of the work we did on fastnoise_source. This should especially fix the flaky `qa_header_format`.

### DIFF
--- a/gr-digital/lib/qa_header_format.cc
+++ b/gr-digital/lib/qa_header_format.cc
@@ -29,6 +29,7 @@
 #include <gnuradio/digital/header_format_default.h>
 #include <gnuradio/blocks/unpack_k_bits.h>
 #include <gnuradio/expj.h>
+#include <gnuradio/xoroshiro128p.h>
 #include <volk/volk.h>
 #include <boost/test/unit_test.hpp>
 #include <stdio.h>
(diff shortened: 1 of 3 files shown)


gr-digital/lib/qa_header_format.cc
```diff
@@ -29,25 +29,28 @@
...
 
   std::string ac = "1010101010101010"; //0xAAAA
   unsigned char *data = (unsigned char*)volk_malloc(N*sizeof(unsigned char),
                                                     volk_get_alignment());
-  srand (time(NULL));
-  for(unsigned int i = 0; i < N; i++) {
-    data[i] = rand() % 256;
+  uint64_t *data_ptr = reinterpret_cast<uint64_t*>(data);
+  xoroshiro128p_seed(rng_state, 42);
+  for(unsigned int i = 0; i < N / 8; ++i) {
```
noc0lour: Was gonna complain for cases N not being multiples of 8, maybe actually tell people N *has* to be a multiple of 8, not only that it's convenient.

marcusmueller: True. Will revise comment above.

Maybe I should be writing a helper function in the xoroshiro sources to fill a buffer with random bytes. hm.

marcusmueller: True. Will revise comment above.

Maybe I should be writing a helper function in the xoroshiro sources to fill a buffer with random bytes. hm.

marcusmueller:
@noc0lour done. Please review, and if ok, squash'n'merge.

gr-digital/lib/qa_header_format.cc
```diff
@@ -29,25 +29,40 @@
...
 #include <volk/volk.h>
 #include <boost/test/unit_test.hpp>
 #include <stdio.h>
 #include <cmath>
 
+static void xoroshiro_fill_buffer(uint8_t *buffer, unsigned int length, uint64_t seed = 42) {
+  uint64_t rng_state[2];
+  xoroshiro128p_seed(rng_state, seed);
+  uint64_t *data_ptr = reinterpret_cast<uint64_t*>(buffer);
+  for(unsigned int i = 0; i < length / 8; ++i) {
```
mbr0wn: I would have done `sizeof(uint64_t)` instead of 8, but that's no reason not to merge.

noc0lour:
Merged with 94af848ebe70450a655685d74a75b1977fa120c9

Issue #2050, opened 2018-09-26
Constellation Object and Constellation Rect. Object are broken


argilo:
On the master branch, the Constellation Object and Constellation Rect. Object blocks (from the "Modulators" category) cannot be added to a flow graph. When adding the Constellation Object, I see the following errors at the console:
```
Traceback (most recent call last):
  File "/home/argilo/prefix_next/lib/python3.6/dist-packages/gnuradio/grc/core/blocks/_templates.py", line 59, in _get_template
    return self._template_cache[str(text)]
KeyError: '% if str(type) == "calcdist":\nself.${id} = ${id} = digital.constellation_calcdist(${const_points}, ${sym_map},\n${rot_sym}, ${dims}).base()\n% else:\nself.${id} = ${id} = digital.constellation_${type}().base()\n% endif\n% if str(soft_dec_lut).lower() == \'\'"auto"\'\' or str(soft_dec_lut).lower() == "\'\'auto\'\'":\nself.${id}.gen_soft_dec_lut(${precision})\n% elif str(soft_dec_lut) != \'\'None\'\':\nself.${id}.set_soft_dec_lut(${soft_dec_lut}, ${precision})\n% endif'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/mako/pyparser.py", line 41, in parse
    return _ast_util.parse(code, '<unknown>', mode)
  File "/usr/lib/python3/dist-packages/mako/_ast_util.py", line 84, in parse
    return compile(expr, filename, mode, PyCF_ONLY_AST)
  File "<unknown>", line 2
    elif str(soft_dec_lut) != ''None'':pass
                                   ^
SyntaxError: invalid syntax

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/argilo/prefix_next/lib/python3.6/dist-packages/gnuradio/grc/core/blocks/_templates.py", line 50, in compile
    template = Template(text)
  File "/usr/lib/python3/dist-packages/mako/template.py", line 320, in __init__
    (code, module) = _compile_text(self, text, filename)
  File "/usr/lib/python3/dist-packages/mako/template.py", line 705, in _compile_text
    generate_magic_comment=template.disable_unicode)
  File "/usr/lib/python3/dist-packages/mako/template.py", line 685, in _compile
    node = lexer.parse()
  File "/usr/lib/python3/dist-packages/mako/lexer.py", line 243, in parse
[...]

PR #2058, opened 2018-10-01
gr-digital: Un-break Constellation Object and Constellation Rect Object

haakov:
Fixes https://github.com/gnuradio/gnuradio/issues/2050

### DIFF
--- a/gr-digital/grc/digital_constellation.block.yml
+++ b/gr-digital/grc/digital_constellation.block.yml
@@ -49,9 +49,9 @@ templates:
         % else:
         self.${id} = ${id} = digital.constellation_${type}().base()
         % endif
-        % if str(soft_dec_lut).lower() == ''"auto"'' or str(soft_dec_lut).lower() == "''auto''":
+        % if str(soft_dec_lut).lower() == '"auto"' or str(soft_dec_lut).lower() == "'auto'":
         self.${id}.gen_soft_dec_lut(${precision})
-        % elif str(soft_dec_lut) != ''None'':
+        % elif str(soft_dec_lut) != "None":
         self.${id}.set_soft_dec_lut(${soft_dec_lut}, ${precision})
         % endif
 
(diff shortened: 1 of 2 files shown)


gr-digital/grc/digital_constellation.block.yml
```diff
@@ -49,9 +49,9 @@ templates:
         % else:
         self.${id} = ${id} = digital.constellation_${type}().base()
         % endif
-        % if str(soft_dec_lut).lower() == ''"auto"'' or str(soft_dec_lut).lower() == "''auto''":
+        % if str(soft_dec_lut).lower() == "auto" or str(soft_dec_lut).lower() == "auto":
```
argilo: This doesn't look right. It's the same comparison twice in a row.

gr-digital/grc/digital_constellation_rect.block.yml
```diff
@@ -49,9 +49,9 @@ templates:
     var_make: |-
         self.${id} = ${id} = digital.constellation_rect(${const_points}, ${sym_map},
         ${rot_sym}, ${real_sect}, ${imag_sect}, ${w_real_sect}, ${w_imag_sect}).base()
-        % if str(soft_dec_lut).lower() == ''"auto"'' or str(soft_dec_lut).lower() == "''auto''":
+        % if str(soft_dec_lut).lower() == "auto" or str(soft_dec_lut).lower() == "auto":
```
argilo: Same here.

gr-digital/grc/digital_constellation.block.yml
haakov: Haha wow you're right

haakov:
Okay, this makes more sense :)

argilo:
The code looks like it probably reproduces the behaviour of the old template now (although I haven't tested whether it generates the right Python code).

It seems like it would be better to just use `soft_dec_lut()` though to access the underlying Python object (once that's working properly), instead of working with the string version produced by `repr()`.

I'm okay leaving the code as it is though, assuming it works properly.

PR #2063, opened 2018-10-02
Coverity fixes

balister:


balister:
Yeah, since he block life tie is the same as the flowgraph, the leak is not apparent. But if we every have flowgraphs create and destroy block while running, this class of leak will bite us.

gr-blocks/lib/tuntap_pdu_impl.cc
```diff
@@ -168,9 +168,12 @@ namespace gr {
       ifr.ifr_mtu = MTU;
 
       /* try to set MTU */
-      if ((err = ioctl(sfd, SIOCSIFMTU, (void *) &ifr)) < 0)
-        return err;
+      if ((err = ioctl(sfd, SIOCSIFMTU, (void *) &ifr)) < 0) {
```
marcusmueller: Couldn't we just do 
<pre>
err = ioctl(sfd, SIOCSIFMTU, (void *) &ifr);
close(sfd);
return (err < 0) ? err : MTU;
</pre>

marcusmueller:
Cherry-picked the memleak patch onto master, nothing to improve! (will do the same for maint-3.7)

gr-blocks/lib/tuntap_pdu_impl.cc
marcusmueller: also, and that's what I'm actually worried about: Shouldn't we be returning `errno` instead of `err` (which will always be -1 in case of error)?

PR #2069, opened 2018-10-06
Change vlen variables to unsigned datatype. Fixes #1459

### LINKED ISSUE #1459: integrate has `int vlen` parameter; should be `unsigned`; check other block's
There's a lot of blocks that take `vlen` parameters. This is a reminder to myself (please assign to me) to go through the tree and identify all places where we should've used unsigned int instead of int.

This might look cosmetic (it's unlikely someone would want a vector half of MAX_INT in size, and the scheduler and buffer allocator might reasonably break on that), but it actually gives compiler warnings when trying to parameterize a block with a signed vlen, so it's actually a code quality issue.

sbmueller:
Hi, I had a look at an issue from last year's hacktoberfest which did not seem to be resolved so far. In accordance with GREP 11 I did some code cleanup on the files I touched.

### DIFF
--- a/gr-blocks/include/gnuradio/blocks/integrate.h
+++ b/gr-blocks/include/gnuradio/blocks/integrate.h
@@ -43,7 +43,7 @@ template<class T>
       // gr::blocks::integrate::sptr
       typedef boost::shared_ptr< integrate<T> > sptr;
 
-      static sptr make(int decim, int vlen = 1);
+      static sptr make(int decim, unsigned int vlen = 1);
     };
 
 typedef integrate<std::int16_t> integrate_ss;
--- a/gr-blocks/include/gnuradio/blocks/stream_to_tagged_stream.h
+++ b/gr-blocks/include/gnuradio/blocks/stream_to_tagged_stream.h
@@ -55,7 +55,7 @@ namespace gr {
        */
       static sptr make(
 	  size_t itemsize,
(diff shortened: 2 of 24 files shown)


noc0lour:
@sbmueller Hi, thanks for the effort. Unfortunately The first (and most important part) of GREP 11 hasn't been implemented: Reformatting the full source tree. Thus mixing formatting & bug fixes is undesirable. Sorry for not having been able to have the reformatting rampage done. According to @marcusmueller the reformatting may take place once 3.8 is released.

sbmueller:
@noc0lour 
> Thus mixing formatting & bug fixes is undesirable.

So does that mean I should split formatting and bug fixing commits or do no formatting changes at all?

@michaelld I tried to solve #1459 here. Don't know what happened to the whitespaces. It should be spaces, of course. I will follow up soon.

noc0lour:
Personally I'm fine with separate commits with the formatting in place. We shouldn't clutter history with a lot of formatting commits though. The idea is to have a big formatting commit which everyone could skip.

marcusmueller:
Very nice indeed. Note, however, that this **intentionally** introduces a lot of new warnings to our build. 

We'll need to fix those sooner or later; these are exactly the reason I wanted this to happen. Now, we're notified of any place where we might be doing something stupid.

PR #2070, opened 2018-10-07
modtool: Bump min Boost version for OOTs to 1.54

mbr0wn:
This matches the version we use for GNU Radio itself.


marcusmueller:
Thanks for highlighting this!

Um, I must admit that as far as I can see, cmake/Modules/GrBoost.cmake still contains the line of additional versions starting with "1.45.0", but, tbh, that's probably bad!

So, I don't see how we're saying that the minimum supported Boost version is 1.54 and then accept anything below that. 

Then I tried to figure out _why exactly_ we're requiring 1.54.

Can you help me out? Feels like **I** was the one advocating moving to 1.54. But: The Redhat guys are building pre-grcon master (I think it's master) with Boost 1.53... 

Can you remember what brought on the need for 1.54 specifically? Or was it just that we said "OK, we're going to make a cut here, so that we don't support ancient Boost forever"?

mbr0wn:
@marcusmueller I don't know where the move to 1.54 came from. In UHD we use 1.53; it has a whole bunch of interesting features, and is available in some older Ubuntu. This PR is to align OOTs with mainline.

mbr0wn:
Prior to 1.54 in GNU Radio we had something really old, which was causing issues somewhere.

PR #2072, opened 2018-10-10
uhd: Re-enable GRC bindings

mbr0wn:
This reverts 331ccf8a. The GRC bindings were disabled temporarily
because of Cheetah/Mako incompatibilities which were resolved.


mbr0wn:
Note: The blocks work, but they need some polishing that got lost in the master->next merge. Will work on that separately, this is just to unblock gr-uhd.

PR #2073, opened 2018-10-10
uhd: Minor fixes to GRC bindings

mbr0wn:
- Show/Hide LO settings now works the right way round
- Gain Type changed from bool to enum
- Fix Pylint warning on the YML generation code

### DIFF
--- a/gr-uhd/grc/gen_uhd_usrp_blocks.py
+++ b/gr-uhd/grc/gen_uhd_usrp_blocks.py
@@ -1,5 +1,5 @@
 """
-Copyright 2010-2011,2014 Free Software Foundation, Inc.
+Copyright 2010-2011,2014,2018 Free Software Foundation, Inc.
 
 This file is part of GNU Radio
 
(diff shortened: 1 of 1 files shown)


mbr0wn:
@marcusmueller Sorry for the PR spam -- I wasn't sure which order you wanted to merge this and #2072, so I just made two separate PRs.

haakov:
@mbr0wn Thanks a lot for taking care of the UHD blocks! 

About the `bool` parameter type (I noticed you didn't change the "Show/Hide LO controls" parameter back to `enum`):

Since PR https://github.com/gnuradio/gnuradio/pull/2028 was merged, all parameters with a `bool` dtype are no longer represented by a dropdown list, but a switch button, like this (depending on the theme):

deepinscreenshot_gnuradio-companion_20181010233138

This means that the following two following lines are ignored and can be removed:

 https://github.com/gnuradio/gnuradio/blob/30c9ad0870269707e5d39cbba9646e6c9beb03e9/gr-uhd/grc/gen_uhd_usrp_blocks.py#L338-L339

The `default` field still works like before, though :)

mbr0wn:
@haakov Removed those lines.

marcusmueller:
Hey, looks like you adapted USRP source's "Show LO Control fields", but not "Show Command Port":

controls


(and yes, that's the way the colorscheme plays out on my machine. I need to fix that.)

PR #2074, opened 2018-10-10
modtool: Move to Boost.UTF

mbr0wn:


marcusmueller:
Love this! Will amend your second commit's message "Fix bug in templates for pre-3.7 OOTs", as I think we'd all like to see from the git history what was fixed when, and "bug" is a bit below our abilities in describing the issue that was fixed.

marcusmueller:
Also, I think you should be credited for more: The first commit not only moves to Boost.utf, but also makes all the gr-newmod cmake 3.8-aware. So, I'll change that commit message, too (laters, though, since high-tech country Germany's train wifi is too crappy to `git fetch mbr0wn`).

marcusmueller:
Superseeded  by #2080

marcusmueller:
superseeded by code-wise identical #2080

PR #2076, opened 2018-10-11
modtool: Update cmake/Modules files

mbr0wn:
This imports various fixes that were already applied to cmake/Modules/*
in GNU Radio itself, but weren't forward-ported to gr-newmod.


mbr0wn:
...actually, this fixes a bug when using Python 3.

marcusmueller:
> ...actually, this fixes a bug when using Python 3.

Can you elaborate which one?

mbr0wn:
@marcusmueller The CMake files execute Python code for various reasons, and the code in the `.cmake` files wasn't Py3k-compatible.

PR #2090, opened 2018-10-17
gr-qtgui: 3.8- time raster sink: set default values for ncols and nrows

777arc:
Without these defaults set the block errors out when you run the flowgraph


marcusmueller:
Even without what this fixes, those are very reasonable defaults to have. Thank you!

PR #2091, opened 2018-10-17
gr-qtgui: 3.8- fixed time sink error

777arc:
Recent commit 1f857d30d5d73576070d70ad785cba3d505b2a73 changed variable "which" to be unsigned, so the default can't be -1.  I set it to 0, seems to work...

### DIFF
--- a/gr-qtgui/grc/qtgui_time_sink_x.block.yml
+++ b/gr-qtgui/grc/qtgui_time_sink_x.block.yml
@@ -516,7 +516,7 @@ templates:
 
         self.${id}.set_y_label(${ylabel}, ${yunit})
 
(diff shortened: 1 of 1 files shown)


dkozel:
Yuck! It looks like it was always being cast to `size_t` so there should be no change in behavior here.
https://github.com/gnuradio/gnuradio/commit/1f857d30d5d73576070d70ad785cba3d505b2a73#diff-4888af43004d97ec6ac54c1b483cbd5aR612

noc0lour:
I think the idea was to overload the function to take no `which` to apply to all ports.

mbr0wn:
Hand-merged. Had to modify commit message.

PR #2093, opened 2018-10-17
gr-uhd 3.8- fixes some of the many bugs in usrp yaml

777arc:
Even with these fixes there is still a major block of code missing from this new yaml, and it's the code that tunes and sets the gain, etc.  Here is the old xml version of the block of code I'm referring to:

```
#for $n in range($max_nchan)
\#if \$nchan() > $n
self.\$(id).set_center_freq(\$center_freq$(n), $n)
\#if \$norm_gain${n}()
self.\$(id).set_normalized_gain(\$gain$(n), $n)
\#else
self.\$(id).set_gain(\$gain$(n), $n)
\#end if
	\#if \$ant$(n)()
self.\$(id).set_antenna(\$ant$(n), $n)
	\#end if
	\#if \$bw$(n)()
self.\$(id).set_bandwidth(\$bw$(n), $n)
	\#end if
#if $sourk == 'source'
	\#if \$lo_export$(n)() and not \$hide_lo_controls()
self.\$(id).set_lo_export_enabled(\$lo_export$(n), uhd.ALL_LOS, $n)
	\#end if
    \#if \$lo_source$(n)() and not \$hide_lo_controls()
self.\$(id).set_lo_source(\$lo_source$(n), uhd.ALL_LOS, $n)
	\#end if
	\#if \$dc_offs_enb$(n)()
self.\$(id).set_auto_dc_offset(\$dc_offs_enb$(n), $n)
	\#end if
	\#if \$iq_imbal_enb$(n)()
self.\$(id).set_auto_iq_balance(\$iq_imbal_enb$(n), $n)
	\#end if
#end if
\#end if
#end for
```

### DIFF
--- a/gr-uhd/grc/gen_uhd_usrp_blocks.py
+++ b/gr-uhd/grc/gen_uhd_usrp_blocks.py
@@ -41,8 +41,8 @@ parameters:
 -   id: stream_args
     label: Stream args
     dtype: string
-    options: [peak=0.003906]
-    option_labels: [peak=0.003906]
+    options: [none, peak=0.003906]
+    option_labels: [Default, peak=0.003906]
     hide: ${'$'}{ 'none' if stream_args else 'part'}
 -   id: stream_chans
     label: Stream channels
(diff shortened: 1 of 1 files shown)


mbr0wn:
Hand-merged. Had to edit the commit message. You're right about the missing UHD settings.

PR #2071, opened 2018-10-08
cmake: Error out if Log4Cpp is not found

noc0lour:

PR #2094, opened 2018-10-17
Log4cpp is required. Check for it with cmake

btashton:
The cmake module existed for log4cpp, but it was not being used.  At one point it looks like LOG4CPP_FOUND was used in the code to make logging optional, but this is a hard requirement.  This change adds it as a requirement to "gnuradio-runtime" where the logging code is stored.


drmpeg:
I believe this is essentially a dupe of https://github.com/gnuradio/gnuradio/pull/2071

btashton:
Ah yes, I did not see that other PR.  This outside the requirement against the component that actually needs it instead of globally, but it really does not matter since you always need the runtime.

michaelld:
The PRs are different. They do different things that are important! +1 from me to commit this one too.

noc0lour:
Having this is probably a good idea until we might decide to mix things up (if modern CMake helps us somewhere).

mbr0wn:
@michaelld approved, I merged on his behalf. @btashton: I had to edit the commit message.

michaelld:
Thanks @mbr0wn ... I approved before I had commit access, but it's good to get other eyes / brains on the matter too ;)

PR #2095, opened 2018-10-17
Wrap map with list for in various block yml files. Fixes #2036

### LINKED ISSUE #2036: Trying to use a Random source fails
Using a simple flowgraph with a random source of type int  fails with:

    File "/usr/local/gnuradio/lib/python3.6/dist-packages/gnuradio/blocks/blocks_swig1.py", line 6302, in make
    return _blocks_swig1.vector_source_i_make(*args, **kwargs)
    TypeError: in method 'vector_source_i_make', argument 1 of type 'std::vector< std::int32_t,std::allocator< std::int32_t > > const &'

btashton:
Map now returns a generator in python3 instead of a list.
vector_source_x_make expects a list or tuple.

http://python3porting.com/differences.html#map

### DIFF
--- a/gr-analog/grc/analog_random_source_x.block.yml
+++ b/gr-analog/grc/analog_random_source_x.block.yml
@@ -36,8 +36,8 @@ templates:
     imports: |-
         from gnuradio import blocks
         import numpy
-    make: blocks.vector_source_${type.fcn}(map(int, numpy.random.randint(${min}, ${max},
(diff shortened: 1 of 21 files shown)


btashton:
All of these blocks in gr-fec have now been fixed.

marcusmueller:
Definitely the right thing to do! Thank you :)

PR #2108, opened 2018-10-24
Add files via upload

1fancyhat:
Changed gen_uhd_usrp_blocks.py to rebuild uhd_usrp_source.xml. Fixed FE Correction tab/ dc_offset iq_balance.


1fancyhat:
I originally put this in master. This belongs in a location the creates the .xml file type not the one that creates .yml file types.

PR #2109, opened 2018-10-25
cmake: fix finding Python packages

### LINKED ISSUE #2006: dist-packages is Debian specific and not a general "unix thing"

karel:
Removes the hard-coded 'dist-packages' because it's Debian specific and leads to problems on distributions which use 'site-packages' (e.g. Arch and Gentoo). This commit still retains the current behavior in virtual environments without 'getsitepackages'.

This fixes #2006 and is an alternative to #2019.

### DIFF
--- a/cmake/Modules/GrPython.cmake
+++ b/cmake/Modules/GrPython.cmake
@@ -93,12 +93,12 @@ endmacro(GR_PYTHON_CHECK_MODULE)
 ########################################################################
 if(NOT DEFINED GR_PYTHON_DIR)
 execute_process(COMMAND ${PYTHON_EXECUTABLE} -c "
-import os
-import sys
-if os.name == 'posix':
-    print(os.path.join('lib', 'python' + sys.version[:3], 'dist-packages'))
-if os.name == 'nt':
-    print(os.path.join('Lib', 'site-packages'))
+try:
+    from site import getsitepackages
+    print(getsitepackages()[0])
+except ImportError:
+    from distutils.sysconfig import get_python_lib
+    print(get_python_lib(prefix=''))
 " OUTPUT_VARIABLE GR_PYTHON_DIR OUTPUT_STRIP_TRAILING_WHITESPACE
 )
 endif()
--- a/cmake/Modules/GrPython.cmake
+++ b/cmake/Modules/GrPython.cmake
@@ -98,7 +98,7 @@ try:
     print(getsitepackages()[0])
 except ImportError:
     from distutils.sysconfig import get_python_lib
-    print(get_python_lib(prefix=''))
+    print(get_python_lib())
 " OUTPUT_VARIABLE GR_PYTHON_DIR OUTPUT_STRIP_TRAILING_WHITESPACE
 )
 endif()


cmake/Modules/GrPython.cmake
```diff
@@ -93,12 +93,12 @@ endmacro(GR_PYTHON_CHECK_MODULE)
...
 execute_process(COMMAND ${PYTHON_EXECUTABLE} -c "
-import os
-import sys
-if os.name == 'posix':
-    print(os.path.join('lib', 'python' + sys.version[:3], 'dist-packages'))
-if os.name == 'nt':
-    print(os.path.join('Lib', 'site-packages'))
+try:
+    from site import getsitepackages
+    print(getsitepackages()[0])
```
michaelld: At least for me: This line returns the whole path, while the other one returns a relative path. I'm not sure it makes a difference to CMake, but I think it's better for the value of GR_PYTHON_DIR to be consistently set: full or relative, pick one & go with it.

michaelld:
I agree that this is a better fix for the issue than #2019. So please get the path fixed & we'll go from there.

cmake/Modules/GrPython.cmake
karel: I am not sure which would be the better choice so that everything worked 
> in and out of prefixes #2019

maybe @djanderson can provide some insight.

michaelld: Yes I think that will do the trick.

michaelld: The prior code was relative, but from testing absolute paths work too. So, we can go either way. I like your solution to just remove the `prefix=""` part.

karel: Very well, then.

cmake/Modules/GrPython.cmake
```diff
@@ -93,12 +93,12 @@ endmacro(GR_PYTHON_CHECK_MODULE)
...
+except ImportError:
```
marcusmueller: under which circumstances can this fail? isn't `site` part of Python 2.7+?

karel: The `getsitepackages` is missing from `virtualenv` (https://github.com/pypa/virtualenv/issues/228).

marcusmueller:
squashing and merging this. Thanks!

marcusmueller:
Nope, this patch does *not* work. It breaks the ability to use `CMAKE_INSTALL_PREFIX`. Had to revert on master.

marcusmueller:
@michaelld any idea how to make this respect `CMAKE_INSTALL_PREFIX` and still use the correct `$PREFIX/lib[64]/python{version}/{dist,site}-packages/` path?

michaelld:
I think if we take this change and augment it to remove the leading $CMAKE_INSTALL_PREFIX + "/" to make the GR_PYTHON_DIR a relative path ... and, check to make sure the leading "/" is removed ... should work ... assuming that CMake and Python are installed into the same top-level $PREFIX ... if not then the value of GR_PYTHON_DIR needs to be set as a cmake argument.

michaelld:
So if we leave this PR's code as is, then add some CMake code after it and before "endif()" -- that tries to strip the PREFIX from the Python output "GR_PYTHON_DIR" .. something like:
{{{
STRING(REPLACE ${CMAKE_INSTALL_PREFIX} "" GR_PYTHON_DIR ${GR_PYTHON_DIR})
STRING(SUBSTRING ${GR_PYTHON_DIR} 0 1 FIRST_CHAR)
STRING(COMPARE EQUAL ${FIRST_CHAR} "/" IS_ABSOLUTE)
IF(${IS_ABSOLUTE})
  ERROR OUT SOMEHOW
ENDIF()
}}}
This might do the trick. Both of the strings returned by this PR's code are absolute ... just need to strip off the starting CMAKE_INSTALL_PREFIX, if it exists.

marcusmueller:
ah! yeah, that might work, but it caused something in my head (think it's called an idea):

What if we basically 

```python
import os
import site
import sys
absprefix = os.path.abspath('${CMAKE_INSTALL_PREFIX}')
potential_paths = [site.getsitepackages()] + site.getusersitepackages() + sys.path
abs_paths = [os.path.abspath(p) for p in potential_paths]
prefix_paths = [p if os.path.commonprefix(p, absprefix) == absprefix for p in abs_paths]
if prefix_paths:
    print(prefix_paths[0])
else:
    print(os.path.join(absprefix, potential_paths[0]))
```

marcusmueller:
wait, do we have a variable to explicitly set the path we should be installing to?

michaelld:
Yes there is a variable (GR_PYTHON_DIR) that can be set on the cmake commandline that if set bypasses this whole shebang.

michaelld:
BTW> `site.getsitepackages()[0]` returns the "site-packages" directory for Python, so we don't want to use it to get the top-level CMAKE_INSTALL_DIR ...

michaelld:
Otherwise I like your idea in that code snippet.

karel:
Indeed, setting `GR_PYTHON_DIR` works.

Issue #1236, opened 2017-03-11
Impossibility to disable DC filtering for USRP source in GRC


pelijah:
There is an error in the script _gr-uhd/grc/gen_uhd_usrp_blocks.py_. To disable DC filtering user needs to enter _False_ in the field  _$dc_offs_enbXX_. In this case _set_auto_dc_offset_ will not be called because condition "if False" is always false.
```
	\#if \$dc_offs_enb$(n)()
self.\$(id).set_auto_dc_offset(\$dc_offs_enb$(n), $n)
	\#end if
```
It needs to be changed on something like this
	`\#if \$dc_offs_enb$(n)() != ""`

mbr0wn:
Yep, that's legit. @pelijah do you want to submit a PR for this?

pelijah:
@mbr0wn No

marcusmueller:
closing this two years after a fix was merged (if it wasn't fixed by that fix, the bug reporter would probably have spoken out?)

PR #2110, opened 2018-10-25
modified gen_uhd_usrp_blocks.py

1fancyhat:
Removed the if check around self.\$(id).set_auto_dc_offset(\$dc_offs_enb$(n), $n) and self.\$(id).set_auto_iq_balance(\$iq_imbal_enb$(n), $n). In the FE corrections tab,  if False was entered, the if check would be set to false and prevent the DC filter from being turned off. Now the FE correction tab can be used to turn dc_offset and IQ_balance on or off (On is the default value).

this solves the issue from this link
#1236


mbr0wn:
We're not touching `maint` any more -- the other PR might be relevant though.

PR #2111, opened 2018-10-25
modified gen_uhd_usrp_blocks.py

1fancyhat:
Removed the if check around self.$(id).set_auto_dc_offset($dc_offs_enb$(n), $n) and self.$(id).set_auto_iq_balance($iq_imbal_enb$(n), $n). In the FE corrections tab, if False was entered, the if check would be set to false and prevent the DC filter from being turned off. Now the FE correction tab can be used to turn dc_offset and IQ_balance on or off (On is the default value).

this solves the issue from this link
gnuradio#1236


marcusmueller:
Needs review from @mbr0wn, who's got full merging competences for this. Go wild!

Also, second commit (160e813) looks like it should be squashed into first, maybe?

@1fancyhat this fix is short enough for us to _not_ require any paperwork from you, but if you happen to plan fixing more stuff, or even contributing your own functionality, which I'd very much love to see happen, we can arrange that you get the one form we need you to complete. That would be the CLA, which would allow us to use even complex / long code additions from you in our code base. If you want that, ping me or even better @bhilburn.

Issue #2101, opened 2018-10-21
grc warns on ignoring extra key on startup


dl1ksv:
grc warns 
WARNING:gnuradio.grc.core.platform:variable_struct.block.yml                block: warn: Ignoring extra key 'var_value'
on startup.
At the moment I have no example using a struct variable , but I suppose this will have efffects on using this variable.

PR #2114, opened 2018-10-27
grc: Add missing keys to schema checker to avoid warnings

haakov:
Fixes https://github.com/gnuradio/gnuradio/issues/2101


skoslowski:
I believe I didn't include this when I wrote the schema checker because I wasn't sure about the naming. But I say let's keep them, might not be intuitive but those are the ones we're used to

PR #2116, opened 2018-10-29
grc: Restore right click menu. Fixes #2062

### LINKED ISSUE #2062: GRC's right click popup menu doesn't work
Probably caused by the switch from Gtk.Actions to Gio.Actions. 

Steps to reproduce:
- Right-click anywhere in the GRC canvas

Traceback: 
File "/usr/local/lib/python3.7/dist-packages/gnuradio/grc/gui/DrawingArea.py", line 161, in handle_mouse_button_press 
`event=event,`
File "/usr/local/lib/python3.7/dist-packages/gnuradio/grc/gui/canvas/flowgraph.py", line 664, in handle_mouse_context_press
`self._context_menu.popup(None, None, None, None, event.button, event.time)`
AttributeError: 'ContextMenu' object has no attribute 'popup'

btashton:
When the change was made with gio actions the context menu broke. We need to create a Gtk.Menu from the Gio.Menu to support popup.  Doing this also means that we also need to get access to a Gtk.Widget that has the action group otherwise the context menu will be shown, but the context menu will be show but the states of the menu items and the triggers will not be associated.

We do not cleanly pass down the ApplicationWindow or a parent Widget to these classes, so I had to look for the main window attached to the default application.  We already have code that relies to using the default application, so I am not too worried about it, but it would be nice to look at re-factoring this at some point, maybe even moving the context menu creation to the MainWindow.

### DIFF
--- a/grc/gui/canvas/flowgraph.py
+++ b/grc/gui/canvas/flowgraph.py
@@ -26,16 +26,34 @@ from distutils.spawn import find_executable
 from itertools import count
 
 import six
-from gi.repository import GLib
+from gi.repository import GLib, Gtk
 from six.moves import filter
 
 from . import colors
 from .drawable import Drawable
 from .connection import DummyConnection
-from .. import Actions, Constants, Utils, Bars, Dialogs
+from .. import Actions, Constants, Utils, Bars, Dialogs, MainWindow
 from ..external_editor import ExternalEditor
 from ...core import Messages
 from ...core.FlowGraph import FlowGraph as CoreFlowgraph
+
+
+class _ContextMenu(object):
+    """
+    Help with drawing the right click context menu
+    """
+
+    def __init__(self, main_window):
+        self._menu = Gtk.Menu.new_from_model(Bars.ContextMenu())
+        self._menu.attach_to_widget(main_window)
+
+        # In GTK 3.22 Menu.popup was deprecated, we want to popup at the
+        # pointer, so use that new function instead if we can.
+        if Gtk.check_version(3,22,0) is None:
+            self.popup = self._menu.popup_at_pointer
+
+    def popup(self, event):
+        self._menu.popup(None, None, None, None, event.button, event.time)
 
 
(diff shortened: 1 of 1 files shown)


marcusmueller:
Looks fine to me. Testing manually under Py2 and 3 just to be sure.

btashton:
Interesting, something is off with that import then between python2 and 3. I'll resolve that today, my testing was on python3.

marcusmueller:
Just tested on Py3, same thing!

Maybe something has changed in the meantime; make sure to `git fetch origin; git rebase origin/master`.

btashton:
@marcusmueller I just looked and it appears that I dropped the import of MainWindow when I was cleaning the patch up.  I am building it now, but I think what I just pushed up should be working.

marcusmueller:
confirm: works under Py3. Testing under Py2 now.

btashton:
I did want to leave another note here.
this is depricated as of 3.22:
`self._context_menu.popup(None, None, None, None, event.button, event.time)`
with 3.22 they introduced popup_at_pointer, which I tested and it works:
`self._context_menu.popup_at_pointer(event)`

Do you want to just live with this deprecation warning or should I figure out how to do a conditional on this.  This change was made to better support Wayland, although in our use case I can confirm that the old method works fine in Wayland.

marcusmueller:
so, first of all: works great under Py2.

I'd merge this in its current shape now, and say "of course I'd like to minimize deprecation warnings, but I have to preserve my contributor's time", and ask you whether you'd want to fix that warning in a separate PR. However, if you'd rather fix it now, I can wait with the merging!

btashton:
It looks like F29 is shipping with this (I just upgraded to 29 from 28) so it's probably worth it to make the fix.  Can you take a look at the revised PR? Its a little ugly, but I could not think of a better way to shim the function out.

marcusmueller:
I kinda agree with the shim, testing Py3 now.

marcusmueller:
Works, looks fine, fixes the bug. Merged! Thank you :)

PR #2122, opened 2018-11-02
Removing Scipy / Pylab usage where numpy is sufficient

marcusmueller:
This is a rebase of an old PR against next, #1978


marcusmueller:
requesting review from Marc specifically for the documentation aspect, and from Michael for the general dependency experience.

PR #2125, opened 2018-11-02
docs: moved usage manual to wiki

777arc:
Might be a little confusing what is going on here when looking at the diff.  Let me summarize things:

- Usage manual pages within doxygen have been moved to the wiki, see https://wiki.gnuradio.org
- As such, I have removed the pages (.dox files) and modified the Usage Manual page in doxygen to explain how the pages are now on the wiki
- In order for offline users to have a copy of the usage manual, I created a script that exports the wiki pages into separate text files, and the intention is for that script to be run periodically, and the exported files will be committed to this repo
- There is a new directory, gnuradio/docs/usage-manual that these exported snapshots will reside in, as well as the script to do the exporting automatically. 
- The second commit represents the first wiki page export.  It pulled the usage manual pages off the wiki and saved them to files that all start with "(exported from wiki)".  I'm open to different ideas for that filename prefix.  I just wanted to make it clear to people to not edit those files, edit the wiki instead. 
- To test this PR you will have to blow away your build directory, because otherwise the .dox files I removed will still be there, and show up in the root level of doxygen. 
- For those who want to preview this change without pulling and building the PR, I took a screenshot of what the doxygen looks like after this change:

Why change things?  Well for starters, our usage manual within the doxygen has been barely touched over the past few years; there was a ton of outdated info and spelling/grammar errors before I did an editing pass through the pages a few months ago. 

The motivation behind having it in the wiki is

1. it's very quick and easy for anyone to edit,
2. the wiki is easier to organize, the doxygen usage manual has never been easy for new people to find/use, and
3. it gets us closer to having all documentation that is not specific to a certain block/class/function in the wiki.  That way we can explain to new people that all documentation and tutorials are in the wiki, unless it's specific to block/class/function, in which case it is stored in docstrings, and can be found in the doxygen manual as well as the block's Documentation tab in GRC.

I'm open to suggestions.  I ran this plan by several people at grcon18 and the overall response was that any re-org of our documentation is welcome, as long as there is someone willing to put in the effort.


mbr0wn:
I'm assigning to @marcusmueller to  block @777arc from merging. I'd like to have a good discussion on what the plan is for docs.

mbr0wn:
...if that discussion has already occurred, that's also fine. Then we can re-assign @777arc.

mbr0wn:
I'm not a fan of moving these files out of the codebase. I'm also not the documentation officer :smile: 

But here's my reasons:

## Tracking of multiple versions

For the next few months, we'll need at least two different manuals: One for the 3.7 and one for the 3.8 release cycle. The wiki only lets us have one version. Putting them into the code lets us have as many versions of the manual as there are branches.

## Linking code and docs

It's much easier to link code and docs when they're both in the same repo (and manual). Doxygen will auto-resolve links to blocks or wherever. External links to the manual are not guaranteed to stay valid.

## Proximity of code and docs

Leaving the manual pages in the repo means code and docs are closer together. You can distribute files however you want: You can have manual pages in block docs, or vice versa. You can include snippets from a block doc in a manual page, etc. You can also make a comment in the code "when you change this code, don't remember to also edit this file". Then, in the same commit, you update code and manual.

## Inevitable bitrot

Whenever we remove a component from the repo (e.g., gr-pager, gr-noaa) we are sending it to its death. Sure, we don't phrase it like that. We say things like "someone can maintain this if they want". But that almost never happens. When it's not in the repo, it will die from bitrot, a few select OOTs are the exception.
This might not be true here, after all, you're the guardian of the docs, and we've never had one of those before. But will you be here forever? And which triggers do you use to keep the code and manual in sync?

## Ergonomics

This is purely subjective, but I actually prefer editing MD files rather than mediawiki.

mbr0wn:
I'm not a fan of moving these files out of the codebase. I'm also not the documentation officer :smile: 

But here's my reasons:

## Tracking of multiple versions

For the next few months, we'll need at least two different manuals: One for the 3.7 and one for the 3.8 release cycle. The wiki only lets us have one version. Putting them into the code lets us have as many versions of the manual as there are branches.

## Linking code and docs

It's much easier to link code and docs when they're both in the same repo (and manual). Doxygen will auto-resolve links to blocks or wherever. External links to the manual are not guaranteed to stay valid.

## Proximity of code and docs

Leaving the manual pages in the repo means code and docs are closer together. You can distribute files however you want: You can have manual pages in block docs, or vice versa. You can include snippets from a block doc in a manual page, etc. You can also make a comment in the code "when you change this code, don't remember to also edit this file". Then, in the same commit, you update code and manual.

## Inevitable bitrot

Whenever we remove a component from the repo (e.g., gr-pager, gr-noaa) we are sending it to its death. Sure, we don't phrase it like that. We say things like "someone can maintain this if they want". But that almost never happens. When it's not in the repo, it will die from bitrot, a few select OOTs are the exception.
This might not be true here, after all, you're the guardian of the docs, and we've never had one of those before. But will you be here forever? And which triggers do you use to keep the code and manual in sync?

## Ergonomics

This is purely subjective, but I actually prefer editing MD files rather than mediawiki.

noc0lour:
I have to disagree partly. I agree with @mbr0wn on the API side of things. That type of documentation is best kept as close as possible to the code since API changes have a big influence on the documentation. 
On the other hand usage/tutorial/best practice is heavily bitrotting in tree **right** **now**. (Same for the guided tutorial in the wiki.) There is no real benefit of having it in tree if nobody is really looking for it. Since API changes are not immediately reflected in changing `*.dox` files.
It is way better if someone (also including myself) is looking regulary at current tutorial & guides we provide.

777arc:
- Do we need two different usage manuals for 3.7 and 3.8?  What changed that is in the usage manual?  I can't imagine it's much.  
- There are very few code links in the usage manual, and I already went through the wiki and manually added the important links.  If you look at the type of info in the usage manual, it's less about finding the right page in the C++ API and more about *how* to use stuff like tags/PMTs/msgs etc
- The current usage manual doesn't pull from any docstrings or vice versa
- @noc0lour is right that the usage manual has been bitrotting hard in tree over the last few years.  I think it's because the type of people who use the usage manual are those learning GNU Radio, and they are likely more comfortable making a wiki account and editing the docs than putting in a PR.
- Most people don't edit mediawiki files directly, they use the web gui editor.

Lets make sure this conversation is limited to the Usage Manual (as it exists at the moment), not any other portion of the doxygen.  I don't plan on adding more "Usage Manual" pages, but rather organizing the material so that we don't have a tutorial on Stream Tags across 3 different wiki pages (one of those being the usage manual page)

dkozel:
The comment I deleted was a duplicate.

777arc:
I think we can all agree on one thing- that the Usage Manual is "tutorial in nature" and isn't tightly coupled to any specific block, but rather it's more of a tutorial about how to use stuff like stream tags, messaging, etc.  I think it's very much part of the "user" documentation, it's the first place new users should go if they want to become a GNU Radio power user.  Improvements/fixes made to documentation usually happen when a user/developer goes to reference the docs, and sees an error, and then decides to fix it right then and there.  Well most people using the Usage Manual are newer folks, not as much the core devs, and they will probably be more likely to edit a wiki page than figure out what file to edit and put in a PR.  I'm not 100% sure about that, all I know is that the Usage Manual has barely been edited over the last 3 years if you don't include my big editing effort last year. 

FYI, I converted the entire Usage Manual to the wiki, see [here](https://wiki.gnuradio.org/index.php/Category:Usage_Manual), and I created a script that exports all the pages into text files that closely resemble the .dox files I'm proposing be removed from tree and replaced with these periodic exports.  The idea is to run that script and commit the files to tree before every major release.  It's not a perfect system, but it's mainly meant for folks who need offline use of these docs.

@mbr0wn What do you think Martin?

bhilburn:
I am in favor of moving the usage manual to the wiki.

bhilburn:
You should proceed, @777arc 👍

marcusmueller:
I'm merging this: We discussed this extensively, and while I do agree that the wiki in the past has tended to outdate very uglily, this is clearly a usability advantage. People being able to edit the wiki is probably a good thing!

I'm not 100% clear what the copyright of wiki edits is. Can't be auto-assigned to FSF, because not all wiki users have CLAs. @bhilburn

bhilburn:
Copyright is by owner, and all content submitted is reproduced under GNU
FDL.

In short, no CLA necessary.

777arc:
I forget, is it OK to use the web GUI's merge?  And if so, which option?

bhilburn:
The web UI works fine - there's no difference in the end result. Since there's only one commit here, anyway, not much point in a squash. Rebasing & merging should be fine 👍

Issue #2118, opened 2018-10-30
docs: remove deprecated & removed blocks from sphinx documentation


noc0lour:
Quite some blocks are still referenced in the sphinx documentation and thus the sphinx build fails. 
Cleaning up of `docs/sphinx/` is necessary to fix that.

777arc:
I'll take a look

777arc:
To anyone else trying to build the sphinx docs, make sure to apt-get install python3-sphinx, then go into the build directory and run ./run_sphinx_build.sh

Editing run_sphinx_build.sh and adding -vvv to the sphinx-build will show what went wrong

777arc:
To get it to build I had to:

- index.rst- delete the whole "Video Blocks" section
- delete the video_sdl_blocks.rst file

marcusmueller:
Closed per #2222

PR #2126, opened 2018-11-02
docs: sphinx: removed video_sdl blocks so sphinx builds

777arc:
Probably other stuff that needs to be cleaned up in the sphinx rst files (e.g. blocks missing), but at least now the sphinx docs will build.  Addresses issue #2118

### DIFF
--- a/docs/sphinx/source/index.rst
+++ b/docs/sphinx/source/index.rst
@@ -1117,16 +1117,6 @@ UHD Blocks
    gnuradio.uhd.usrp_source
 
 
-Video Blocks
-------------
-
-.. autosummary::
-   :nosignatures:
-
-   gnuradio.video_sdl.sink_s
-   gnuradio.video_sdl.sink_uc
-
-
 
 Waveform Generator Blocks
 -------------------------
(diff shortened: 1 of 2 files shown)


marcusmueller:
I like when Sphinx works¹! 

So, are we only removing linking to blocks that don't exist anymore, anyways, or are there some blocks that are still there, but now don't appear in sphinx? (I know this is a moot point if Sphinx didn't build before, I just want to keep track of what we need to fix)

<hr>
¹for now – just in tune with your other docs endeavours, I think we need to harmonize documentation and put all API docs in one place. If we could, in the long term, I'd like to integrate the python API docs with our C++ doxygen, but we'll first need to come up with a maintainable way of doing so.

777arc:
@marcusmueller  100% agree, I think the first step is to determine how people actually use the sphinx docs, and then decide how to organize that information, and lastly how to incorporate it into the doxygen manual.

btashton:
I would love to see the doxygen stuff move into Sphinx. At least I find it personally a lot more enjoyable to work with. 

Looks like this actually helps make that more possible

https://breathe.readthedocs.io/en/latest/

777arc:
@btashton yes there has been others who have suggested doing that.  Let me learn more about it.

PR #2129, opened 2018-11-03
gr-zeromq: Add last_endpoint function to zmq source and sink blocks

### LINKED ISSUE #2124: gr-zeromq: There is no way to extract the actual endpoint address from the ZMQ block
Some times it is very useful to let the OS determine the port to bind on by supplying port 0.
`tcp://127.0.0.1:0`  right now if we do this there is no way to then read out the assigned port.  The ZMQ socket provides the getsockopt interface which can be supplied with `ZMQ_LAST_ENDPOINT`

From the docs

> ZMQ_LAST_ENDPOINT: Retrieve the last endpoint set
> The ZMQ_LAST_ENDPOINT option shall retrieve the last endpoint bound for TCP and IPC transports. The returned value will be a string in the form of a ZMQ DSN. Note that if the TCP host is INADDR_ANY, indicated by a *, then the returned address will be 0.0.0.0 (for IPv4).

I also suspect that the hard coded ports in the tests may be causing the CI failures, but that is just a hunch right now.

I'm open to interface suggestions, I was just going to add the interface to return the endpoint string, but would there be a benefit to exposing the whole getsockopt interface instead?

btashton:
This adds a `last_endpoint()` function to the zeromq blocks to allow reading out the address bound to the socket.  This lets us supply an address like `tcp://127.0.0.1:0` and have the OS sort out the port. The ZMQ tests all use this, so multiple copies of a test can run at the same time.

> The ZMQ_LAST_ENDPOINT option shall retrieve the last endpoint bound for TCP and IPC transports. The returned value will be a string in the form of a ZMQ DSN. Note that if the TCP host is INADDR_ANY, indicated by a *, then the returned address will be 0.0.0.0 (for IPv4). 

I added this to the source and sink blocks, although it is most useful to blocks that are performing a bind rather than a connect.

This should resolve #2124

### DIFF
--- a/gr-zeromq/lib/base_impl.cc
+++ b/gr-zeromq/lib/base_impl.cc
@@ -54,6 +54,15 @@ namespace gr {
         delete d_context;
     }
 
+    std::string
+    base_impl::last_endpoint()
+    {
+      size_t addr_len = 256;
+      char addr[addr_len];
+      d_socket->getsockopt(ZMQ_LAST_ENDPOINT, addr, &addr_len);
+      return std::string(addr, addr_len-1);
+    }
+
 
     base_sink_impl::base_sink_impl(int type, size_t itemsize, size_t vlen, char *address, int timeout, bool pass_tags, int hwm)
         : base_impl(type, itemsize, vlen, timeout, pass_tags)
--- a/gr-zeromq/lib/pub_msg_sink_impl.h
+++ b/gr-zeromq/lib/pub_msg_sink_impl.h
@@ -41,6 +41,12 @@ namespace gr {
       ~pub_msg_sink_impl();
 
       void handler(pmt::pmt_t msg);
+      std::string last_endpoint() override {
+        size_t addr_len = 256;
+        char addr[addr_len];
+        d_socket->getsockopt(ZMQ_LAST_ENDPOINT, addr, &addr_len);
+        return std::string(addr, addr_len-1);
+      }
     };
 
   } // namespace zeromq
--- a/gr-zeromq/include/gnuradio/zeromq/pub_msg_sink.h
+++ b/gr-zeromq/include/gnuradio/zeromq/pub_msg_sink.h
@@ -52,6 +52,11 @@ namespace gr {
        * \param timeout  Receive timeout in milliseconds, default is 100ms, 1us increments
        */
       static sptr make(char *address, int timeout=100);
+
+      /*!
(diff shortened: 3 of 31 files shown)


gr-zeromq/lib/base_impl.cc
```diff
@@ -54,6 +54,15 @@ namespace gr {
         delete d_context;
     }
 
+    std::string
+    base_impl::last_address()
+    {
+      size_t addr_len = 256;
+      char addr[addr_len];
+      d_socket->getsockopt(ZMQ_LAST_ENDPOINT, addr, &addr_len);
+      return std::string(addr, addr_len-1);
```
marcusmueller: Thinking about this, doesn't this form of the ‘basic_string‘ ctor only create a std::string *referring* to the char array? I.e. if the same char buffer is reused, the returned std::string would change?

gr-zeromq/lib/pub_msg_sink_impl.h
```diff
@@ -41,6 +41,12 @@ namespace gr {
       ~pub_msg_sink_impl();
 
       void handler(pmt::pmt_t msg);
+      std::string last_address() override {
```
marcusmueller: A :+1: for the use of override!

marcusmueller:
hm, I still see failures on the Ubuntu builds (re-kicked one of the two fails). Wasn't this supposed to fix these? 

Just for clarification: I still 100% endorse these changes, but I'd like to understand what still goes wrong.

noc0lour:
The other zeromq change on master fixes the other half of zeromq issues (and really helps).

btashton:
@marcusmueller yeah @noc0lour  is correct this specific fix is unlikely to help too much with the failures that normally show up, the other PR fixes most of those.  What this does help with on the testing front is lets the same tests run in parallel, and we do not have to "allocate" a port per network test.

I did realized I kind of mixed up the naming convention here.  Might be worth me updating this to use "last_endpoint" instead of "last_address".

btashton:
rebased and renamed last_address to last_endpoint to line up with the zmq api better.

marcusmueller:
I thought I've kicked off the F28py3qt5 build wednesday, but it seems I didn't (or compilation crashed at the exact same point, which I find unlikely and/or alarming enough to do another build). Will do another local build, too.

btashton:
Looks like it passed this time

btashton:
@marcusmueller can we merge this in?

Issue #2051, opened 2018-09-26
Struct Variable block is broken


argilo:
On the master branch, dropping a Struct Variable block (from the "Variables" category) into a flowgraph in GRC results in the following error:
```
Traceback (most recent call last):
  File "/home/argilo/prefix_next/lib/python3.6/dist-packages/gnuradio/grc/core/blocks/_templates.py", line 75, in render
    return template.render(**namespace)
  File "/usr/lib/python3/dist-packages/mako/template.py", line 462, in render
    return runtime._render(self, self.callable_, args, data)
  File "/usr/lib/python3/dist-packages/mako/runtime.py", line 838, in _render
    **_kwargs_for_callable(callable_, data))
  File "/usr/lib/python3/dist-packages/mako/runtime.py", line 873, in _render_context
    _exec_template(inherit, lclcontext, args=args, kwargs=kwargs)
  File "/usr/lib/python3/dist-packages/mako/runtime.py", line 899, in _exec_template
    callable_(context, *args, **kwargs)
  File "memory:0x7fbe94529cf8", line 33, in render_body
TypeError: 'Undefined' object is not callable

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/argilo/prefix_next/lib/python3.6/dist-packages/gnuradio/grc/gui/Application.py", line 422, in _handle_action
    flow_graph_update()
  File "/home/argilo/prefix_next/lib/python3.6/dist-packages/gnuradio/grc/gui/Application.py", line 132, in flow_graph_update
    fg.update()
  File "/home/argilo/prefix_next/lib/python3.6/dist-packages/gnuradio/grc/gui/canvas/flowgraph.py", line 174, in update
    self.rewrite()
  File "/home/argilo/prefix_next/lib/python3.6/dist-packages/gnuradio/grc/core/FlowGraph.py", line 181, in rewrite
    self.renew_namespace()
  File "/home/argilo/prefix_next/lib/python3.6/dist-packages/gnuradio/grc/core/FlowGraph.py", line 212, in renew_namespace
    for variable_block in self.get_variables():
  File "/home/argilo/prefix_next/lib/python3.6/dist-packages/gnuradio/grc/core/FlowGraph.py", line 79, in get_variables
    return expr_utils.sort_objects(variables, attrgetter('name'), methodcaller('get_var_make'))
[...]

Issue #2007, opened 2018-09-14
grc file <var_make> and <var_value> documentation


darsor:
I can't find any documentation on the <var_make> and <var_value> xml tags in block grc files.

I understand <var_make> can be used to define a variable if the variable's name is the block id, and the block key starts with "variable". Are there other possible use cases? For example, is it possible to have a regular block define a variable named something other than the block ID?

I've seen the <var_value> tag used in several variable blocks, but there is no description of its purpose or usage anywhere. Could the documentation be updated?

[This wiki page](https://wiki.gnuradio.org/index.php/GNURadioCompanion#Creating_the_XML_Block_Definition) might be a good place for it.

777arc:
I'll update that wiki page, but does anyone know the answer to @darsor 's questions?  I'm not sure if there are any other use cases.

PR #2131, opened 2018-11-03
grc: Struct Variable template bug convering from Cheetah to Mako

btashton:
Template was still using Cheetah getVar instead of Mako's context
accessor.  Fixes: #2051

### DIFF
--- a/grc/blocks/variable_struct.block.yml.py
+++ b/grc/blocks/variable_struct.block.yml.py
@@ -17,10 +17,11 @@ templates:
         self.${{id}} = ${{id}} = struct({{
             % for i in range({0}):
             <%
-                field = getVar('field' + str(i))
+                field = context.get('field' + str(i))
+                value = context.get('value' + str(i))
             %>
             % if len(str(field)) > 2:
-            ${{field}}: getVar('value' + str(i)),
+            ${{field}}: ${{value}},
             % endif
             % endfor
         }})
@@ -28,10 +29,10 @@ templates:
         struct({{
             % for i in range({0}):
             <%
-                field = getVar('field' + str(i))
+                field = context.get('field' + str(i))
             %>
             % if len(str(field)) > 2:
-            ${{field}}: getVar('field' + str(i)),
+            ${{field}}: ${{field}},
             % endif
             % endfor
         }})
@@ -42,7 +43,7 @@ FIELD0 = """\
     label: Field 0
     category: Fields
     dtype: string
-    default: ${field0}
+    default: field0
     hide: part
 """
 


marcusmueller:
@skoslowski merge at your own discretion

grc/blocks/variable_struct.block.yml.py
```diff
@@ -17,21 +17,21 @@
         self.${{id}} = ${{id}} = struct({{
             % for i in range({0}):
             <%
-                field = getVar('field' + str(i))
+                field = context.get('field' + str(i))
```
mbr0wn: Maybe you want to do `value = context.get(...)` here too?

grc/blocks/variable_struct.block.yml.py
```diff
@@ -17,21 +17,21 @@
...
             %>
             % if len(str(field)) > 2:
-            ${{field}}: getVar('value' + str(i)),
+            ${{field}}: context.get('value' + str(i)),
```
mbr0wn: I see `${field}` in my entry box.hhhhhh

btashton:
@mbr0wn I'll take another stab at it tonight and fix these. Makes sense to me, I was relying on trying to make sense of the broken code and I think I missed exactly what was trying to be accomplished.

grc/blocks/variable_struct.block.yml.py
btashton: Rebuilding to test, but I pushed fixes to what I think is correct.  I am a little unclear what "var_value" is for so I am not sure why the field key is getting assigned to field.  Looks like there was a bug #2007 opened a while ago with someone else confused about what it is for.

mbr0wn:
So, looks like the compiles are failing now. Can you please take a look if you can fix those? Thanks!

btashton:
Build should pass now, but I am worried that you might still have that `${field}` issue you mentioned.  The yml template after built looks like
```
    var_make: |-
        self.${id} = ${id} = struct({
            % for i in range(20):
            <%
                field = context.get('field' + str(i))
                value = context.get('value' + str(i))
            %>
            % if len(str(field)) > 2:
            ${field}: ${value},
            % endif
            % endfor
        })
```

btashton:
@mbr0wn What I pushed should be working, I did add another minor change  changing the default field name for `Field 0` to be `field0` instead of `${field0}`.  It is not clear to me why this ever should have been `${field0}`

btashton:
@mbr0wn I think this should be good to go then.

mbr0wn:
This is an improvement, and a small diff, so I'll declare a assignment-timeout from @skoslowski and merge it manually (there's a typo in the commit message). However, a bunch of other things are broken with the struct, unrelated to this commit.

PR #2133, opened 2018-11-03
Replace dict.iteritems() with dict.items()

phansel:
dict.iteritems() is not implemented in python3; dict.items() in python3 has the same function as dict.iteritems() in python2. This preserves backward and forward compatibility at the cost of possible higher memory usage in python2.

### DIFF
--- a/gr-qtgui/grc/qtgui_check_box.block.yml
+++ b/gr-qtgui/grc/qtgui_check_box.block.yml
@@ -48,7 +48,7 @@ templates:
         %>
         ${win} = Qt.QCheckBox(${(label if (len(label) - 2 > 0) else repr(id))})
         self._${id}_choices = {True: ${true}, False: ${false}}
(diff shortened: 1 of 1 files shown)


marcusmueller:
Hi!

Thanks for this; the whole `iteritems` problem has slipped our attention!

I'll go ahead and slightly modify your commit message (add the component affected) and merge this manually. Also, this is far from the only place we used `iteritems`, so we'll have to fix this in a lot more places!

Generally, this is a single-line fix, and we can merge this without much paperwork. However, if it so happens that you're planning to contribute a larger piece of code, we'll need to get you into our contributor's list. If you want that, let us (especially @bhilburn ) know, and we'll get you set up in no time.

marcusmueller:
By the way, if you want to appear in the CHANGELOG as anything else but `phansel`, let me know.

phansel:
I don't have anything else to commit at the moment, but I will let you know. Thanks for pushing the fix!

PR #2135, opened 2018-11-03
grc: Remove deprecated use of imp module

btashton:
imp module is deprecated as of Python 3.4 but in Python3 we can
use the ModuleType to create a new module.

### DIFF
--- a/grc/core/FlowGraph.py
+++ b/grc/core/FlowGraph.py
@@ -18,7 +18,6 @@
 from __future__ import absolute_import, print_function
 
 import collections
-import imp
 import itertools
 import sys
 from operator import methodcaller, attrgetter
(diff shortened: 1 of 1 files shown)


btashton:
@skoslowski in this case its a little more complicated because `importlib.new_module` does not exist.  You have to use types as I did or create a spec.  Either way we need conditional logic on this one.

skoslowski:
`type(sys)` returns the module type on both pythons, therefore replacing that helper function you added. Try it.

btashton:
We can actually just use ModuleType from types in both py2 and py3.  Accessing by using type on another module is unnecessary.

skoslowski:
Agreed, this would be the preferred solution then.

In general, for GRC, whenever legacy python support is required `six` would be first choice. Then there is grc/core/utils/backports for functions/types beyond what six has.

PR #2136, opened 2018-11-04
gr-analog: fixed analog_random_source bug

777arc:
A bug was introduced in recently merged PR #2095, the ending parenthesis was put in the wrong spot.  I also added a little bit of documentation while I was at it, making sure people know this block isn't really a block but is a wrapper for Vector Source. 

If anyone has some time (@btashton perhaps), it might be worth checking the other files modified in #2095 to see if any other ending parenthesis were put in the wrong spot.


btashton:
@777arc Shoot, you are right, that line had a slightly different pattern to it than the rest of them.  I did just go through the old patch and confirm the closing parenthesis.   This looks good.

777arc:
@btashton np, gave me a chance to learn more about how that block worked

PR #2140, opened 2018-11-04
CMake: remove duplicately installed files

marcusmueller:
These subsequently appeared twice in install_manifest.txt, which lead to
warnings/errors on deinstallation.


marcusmueller:
@noc0lour you'll notice there's something fundamentally wrong about the codecov numbers here…

marcusmueller:
:heart_eyes:

Issue #2137, opened 2018-11-04
python: Do not use bare except, specify exception instead


btashton:
There are a 39 places in the code base that have bare exceptions which are bad practice and tend to hide bugs.  It also makes it hard to understand the expected behaviour.

We should at least remove all the ones that fit this pattern with no comments.
```
try:
    foo()
except:
    pass
```

```
$ pycodestyle-3 --exclude=./volk/* --select=E722 --statistics ./ 
./gr-fec/python/fec/LDPC/Generate_LDPC_matrix_functions.py:68:3: E722 do not use bare 'except'
./gr-filter/python/filter/qa_fft_filter.py:412:13: E722 do not use bare 'except'
./gr-uhd/python/uhd/__init__.py:121:21: E722 do not use bare 'except'
./gr-uhd/apps/uhd_app.py:113:9: E722 do not use bare 'except'
./gr-uhd/apps/uhd_app.py:354:13: E722 do not use bare 'except'
./gr-qtgui/apps/uhd_display.py:256:9: E722 do not use bare 'except'
./gr-qtgui/apps/uhd_display.py:265:9: E722 do not use bare 'except'
./gnuradio-runtime/examples/mp-sched/wfm_rcv_pll_to_wav.py:93:9: E722 do not use bare 'except'
./gnuradio-runtime/examples/mp-sched/wfm_rcv_pll_to_wav.py:106:9: E722 do not use bare 'except'
./gnuradio-runtime/python/gnuradio/eng_arg.py:37:5: E722 do not use bare 'except'
./gnuradio-runtime/python/gnuradio/eng_arg.py:50:5: E722 do not use bare 'except'
./gnuradio-runtime/python/gnuradio/eng_notation.py:74:5: E722 do not use bare 'except'
./gnuradio-runtime/python/gnuradio/eng_option.py:33:5: E722 do not use bare 'except'
./gnuradio-runtime/python/gnuradio/eng_option.py:40:5: E722 do not use bare 'except'
./gnuradio-runtime/python/gnuradio/ctrlport/monitor.py:43:9: E722 do not use bare 'except'
./gnuradio-runtime/python/gnuradio/ctrlport/monitor.py:60:9: E722 do not use bare 'except'
./gnuradio-runtime/python/gnuradio/gr/gr_threading_23.py:448:13: E722 do not use bare 'except'
./gnuradio-runtime/python/gnuradio/gr/gr_threading_23.py:462:13: E722 do not use bare 'except'
[...]

PR #2144, opened 2018-11-04
runtime: gr_threading clutch obsolete; remove it and its usage

marcusmueller:
We essentially monkeypatched Python's `threading` module inelegantly to
get around shortcomings in Python 2.3 and 2.4.

Since support for these versions is long gone: removal of this clutch
and its usage.

As a nice side effect, this removes `gr_threading.*.py` files from btashton's
list of catchall-excepting #2137.


btashton:
Awesome!  I had a commit for this in my WIP exception cleanup and I was thinking about breaking it out into its own.  I had still left the gr_threading.py because I was not sure if that interface was considered an API, but happy to see it gone.  Looks good to me.

PR #2147, opened 2018-11-04
cmake: add debugall buildtype

noc0lour:
This adds a new buildtype called DebugAll, unlike Debug it's fully turning off optimizaton and adding -Wall flags. 

Might be useful for other people as well

### DIFF
--- a/cmake/Modules/GrBuildTypes.cmake
+++ b/cmake/Modules/GrBuildTypes.cmake
@@ -37,7 +37,7 @@ set(__INCLUDED_GR_BUILD_TYPES_CMAKE TRUE)
 # Defines the list of acceptable cmake build types. When adding a new
 # build type below, make sure to add it to this list.
 list(APPEND AVAIL_BUILDTYPES
-  None Debug Release RelWithDebInfo MinSizeRel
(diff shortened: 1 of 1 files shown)


cmake/Modules/GrBuildTypes.cmake
```diff
@@ -94,6 +94,37 @@ if(NOT WIN32)
...
+    "-Wl,--warn-unresolved-symbols,--warn-once" CACHE STRING
+    "Flags used by the shared lib linker during full debug builds." FORCE)
+
+  MARK_AS_ADVANCED(
+    CMAKE_CXX_FLAGS_DEBUGALL
+    CMAKE_C_FLAGS_DEBUGALL
+    CMAKE_EXE_LINKER_FLAGS_DEBUGALL
+    CMAKE_SHARED_LINKER_FLAGS_DEBUGALL)
+endif(NOT WIN32)
+
```
michaelld: Very minor, but you don't need the extra 2x <cr> here.

noc0lour:
Ah, I'm going through parts of the BuildType.cmake file and we have that already  `NOOPTWITHASM` looks like it's the same thing?

noc0lour:
Maybe need to append something there

Issue #2148, opened 2018-11-05
grc: Page does not have a method term_proc


btashton:
grc/gui/Application.py:719 still refers to term_proc even though it is missing.
I'm actually not sure what happened to it, it was added here:
https://github.com/gnuradio/gnuradio/commit/5999bce89cdad39a6eb678b7e4b4063aef211c1d

Unfortunately this is one of those calls behind a bare escape so we did not see this failure....

btashton:
Looks like this was a casualty of a merge conflict
https://github.com/gnuradio/gnuradio/commit/41e1a272e6794feb4d58fed9807d409bd1930c01#diff-94292971eb421de05e01edb207f7540dL706

Do we want the functionality of the PR added back in, or should we just kill?
https://github.com/gnuradio/gnuradio/pull/1393

btashton:
For now restoring this with https://github.com/gnuradio/gnuradio/pull/2138/commits/345408c9858974dad4cddc2fc8ea19bb697acd49

PR #2150, opened 2018-11-05
grc: Clean up flowgraph process management

### LINKED ISSUE #2149: grc: resources are not cleaned up when terminating flowgraph
When running a flow graph and then terminating it two resource warnings are thrown.

The source of these object allocations can be determined with:
`python3.6 -Wd -X tracemalloc=5 -m gnuradio.grc`

```
/usr/lib64/python3.7/subprocess.py:852: ResourceWarning: subprocess 19201 is still running
  ResourceWarning, source=self)
Object allocated at (most recent call last):
  File "/home/bashton/gr-target/lib/python3.7/dist-packages/gnuradio/grc/main.py", lineno 88
    sys.exit(app.run())
  File "/usr/lib64/python3.7/site-packages/gi/overrides/Gio.py", lineno 44
    return Gio.Application.run(self, *args, **kwargs)
  File "/home/bashton/gr-target/lib/python3.7/dist-packages/gnuradio/grc/gui/Application.py", lineno 714
    callback=self.update_exec_stop
  File "/home/bashton/gr-target/lib/python3.7/dist-packages/gnuradio/grc/gui/Executor.py", lineno 46
    self.process = self.page.process = self._popen()
  File "/home/bashton/gr-target/lib/python3.7/dist-packages/gnuradio/grc/gui/Executor.py", lineno 74
    shell=False, universal_newlines=True
/usr/lib64/python3.7/site-packages/gi/overrides/Gio.py:44: ResourceWarning: unclosed file <_io.TextIOWrapper name=12 encoding='UTF-8'>
  return Gio.Application.run(self, *args, **kwargs)
Object allocated at (most recent call last):
  File "/usr/lib64/python3.7/site-packages/gi/overrides/Gio.py", lineno 44
    return Gio.Application.run(self, *args, **kwargs)
  File "/home/bashton/gr-target/lib/python3.7/dist-packages/gnuradio/grc/gui/Application.py", lineno 714
    callback=self.update_exec_stop
  File "/home/bashton/gr-target/lib/python3.7/dist-packages/gnuradio/grc/gui/Executor.py", lineno 46
    self.process = self.page.process = self._popen()
  File "/home/bashton/gr-target/lib/python3.7/dist-packages/gnuradio/grc/gui/Executor.py", lineno 74
    shell=False, universal_newlines=True
  File "/usr/lib64/python3.7/subprocess.py", lineno 756
    encoding=encoding, errors=errors)


```

btashton:
The flowgraph execution had a couple issues.
* The stdout pipe was being treated as bytes instead of strings so the decode needed to be dropped. 
* The kill command was not calling the correct interface #2148 
* The stdout pipe was not being closed  #2149
* The thread was not was not waiting until the process was fully terminated to close #2149


marcusmueller:
The decode / bytes issue is a result / longstanding issue due to py3/2 incompat. I haven't had time to test this yet; is this proven to work under both versions of Python?

grc/gui/Executor.py
```diff
@@ -83,8 +84,14 @@ def run(self):
         r = "\n"
         while r:
             GLib.idle_add(Messages.send_verbose_exec, r)
-            r = self.process.stdout.read(1024).decode('utf-8','ignore')
-        self.process.poll()
+            r = self.process.stdout.read(1)
+
+        # Properly close pipe before thread is terminated
+        self.process.stdout.close()
+        while self.process.poll() is None:
```
mbr0wn: This is a definitive improvement, ideally it would have a timeout value.

btashton: I mostly agree, but we are sending a kill signal not a term signal, so if we were to exit after a time-out the process would still be running.  I would be more inclined to add the logic to send terminate() and then after a timeout fall-back to a kill.   Python 3 wait take a timeout option, so we could use that instead of poll, but we would have to implement a not great workaround for Python 2.

btashton:
@mbr0wn Looks like something got merged into master to "fix" part of this but it is not right.

As I mention in this PR the result of the read will always be a string type because Popen is used with `universal_newlines=true`.  Also we should not be using isinstance with `str` instead use https://pythonhosted.org/six/#six.string_types.  I am rebasing with this change.

PR #2153, opened 2018-11-05
logger: remove duplicate semicolon after macro invocation

noc0lour:


michaelld:
Seems legit. Either this or remove the semicolon from the macro defining GR_LOG_DECLARE_LOGPTR. I prefer the latter, but whatever this works too.

noc0lour:
Whatever is more common with macros is fine with me.

michaelld:
I don't know about more common of less common, but my preference for C/C++ is to end virtually any line or clause with a semicolon. Extras are usually OK, even if they generate compiler warnings. If it were up to me, I'd remove the semicolon from the macro defining GR_LOG_DECLARE_LOGPTR ... of course this version might break OOT builds if any rely on that macro having the trailing semicolon. All IMHO ... each person will have his/her/their own opinion ... so go with whatever you think is more correct.

marcusmueller:
You can satisfy both requirements (no double semicolon and complete statement including terminator):

do{macrobody}while(0)

marcusmueller:
(Iirc, that is. A smart phone isn't a desirable platform to install cpp/g++ on)

PR #2163, opened 2018-11-14
Remove Qt4 support: Part 1

mbr0wn:


mbr0wn:
This is a bit of a hodgepodge of things. There's just so much that's broken w.r.t. Qt.

gr-qtgui/CMakeLists.txt
```diff
@@ -29,15 +29,13 @@ include(GrPython)
 set(DESIRED_QT_VERSION 5 CACHE STRING "The Qt version to use (4 or 5)")
```
michaelld: Do we still need this line & the next line? Guessing they can now be removed & the variable replaced by "5" everywhere.

noc0lour: I agree, removing that will cleanup the cmake considerably.

gr-qtgui/docs/qtgui.dox
```diff
@@ -129,7 +129,7 @@ The QT GUI blocks require the following dependencies.
 \li QtCore (version >= 4.4)
 \li QtGui (version >= 4.4)
 \li QtOpenGL (version >= 4.4)
-\li PyQt4 for Qt4 (version >= 4.4)
+\li PyQt5 for Qt5 (version >= 5.0)
```
noc0lour: What's the lowest version we are able to test with?

gr-uhd/apps/uhd_fft
```diff
@@ -54,7 +54,7 @@ except ImportError:
     from gnuradio.uhd.uhd_app import UHDApp
 
 
-class uhd_fft(gr.top_block, Qt.QWidget, UHDApp):
+class uhd_fft(UHDApp, gr.top_block, Qt.QWidget):
```
noc0lour: Why the resort?

mbr0wn: Something something Python3k. This defines an inheritance order (uhd_fft -> UHDApp -> top_block -> Widget), which means that UHDApp "inherits" from Widget. If it's the other way, then invoking Widget will call the ctor on UHDApp, which fails. The other way, it works.

mbr0wn:
Sure removing `DESIRED_QT_VERSION` is much nicer. Will do that.

gr-qtgui/docs/qtgui.dox
mbr0wn: I don't know -- and the current code has no Qt version requirements whatsoever. Hence, I went down to the bottom.

mbr0wn:
I resolved all of @noc0lour's complaints.

PR #2164, opened 2018-11-14
qtgui: Remove the command line plot tools

mbr0wn:
Note: In gr-utils, we have a version of these tools that don't have Qt4
dependencies (note: they have matplotlib dependencies, though).

None of these tools work with Qt5.


mbr0wn:
@marcusmueller I'm usually the guy that warns about removing stuff, but these are mostly duplicate.

noc0lour:
You have to remove the files from the appropriate `CMakeLists.txt` as well.

marcusmueller:
So, I discussed that in coredevs a while ago, sorry for the long silence.

So, sorry, I'm not 100% OK with this :(

So, we need to get rid of the matplotlib plot thingies. These are typically unusably slow and a pain to use.

We should be sticking to the applications that are nice and fast and use our own plot widgets. That would make sense, as a) higher usability value, b) illustration of how to use the widgets, c) actually our software. 

And we should be moving these to `gr-qtgui/apps`, as that's the module they'd "logically" belong too (also, that'd ensure their dependencies are met when they are installed).

noc0lour:
I'd rather have the applications (we don't have that many) all in examples/apps in the root folder instead of hiding them in `apps/` folders of our libraries. 
This would enable us to have sensible applications depending on a lot of our "sublibraries".

PR #2165, opened 2018-11-14
gr-blocks: use libvolk to speed up nlog10_ff block

mbr0wn:
This results in an order of magnitude improvement for the throughput.

Reviewed-By: Marcus Müller <marcus@hostalia.de>
Reviewed-By: Martin Braun <martin.braun@ettus.com>


mbr0wn:
BBot has two failures: qa_filterbank fails on F28/Qt5, and coverity can't compile because random reasons. I couldn't replicate any of those issues, though.

mbr0wn:
(Oh, and I tried replicating the F28/Qt5 bug on F28/Qt5).

PR #2166, opened 2018-11-15
grc: Add optional parameter to pad_* blocks description

### LINKED ISSUE #2098: grc hier block generation ignores optional parameter in pad sink/source
Even if you set the optional parameter of an pad sink/source to true the generated yml block code has no entry for the optional parameter in the input / output sections.

This means, the pad sink / source is always required.

dl1ksv:
Pad sink/sources blocks may be optional, so connections in the calling flowgraph may not be required.
But at the moment setting optional to yes in the pad property window has no effect. The pad block is alwas required.
This is due to an error in the yml description of pad* blocks.

This fixes #2098

### DIFF
--- a/grc/blocks/pad_sink.block.yml
+++ b/grc/blocks/pad_sink.block.yml
@@ -38,6 +38,8 @@ inputs:
     dtype: ${ type }
     vlen: ${ vlen }
     multiplicity: ${ num_streams }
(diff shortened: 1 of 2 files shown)


mbr0wn:
I put your PR description into the commit message and merged manually.

PR #2169, opened 2018-11-15
rescale constellation points only if max(abs(points))>1

hcab14:
This is needed for QAM constellations which are used, _e.g._, on HF and have max(abs(points))<=1.

If such QAM constellations are normalized to have average magnitude 1, the channel equalizer gets disturbed when switching from 8-PSK to QAM modulation and vice versa.

qam32

### DIFF
--- a/gr-digital/include/gnuradio/digital/constellation.h
+++ b/gr-digital/include/gnuradio/digital/constellation.h
@@ -66,7 +66,8 @@ namespace gr {
       constellation(std::vector<gr_complex> constell,
 		    std::vector<int> pre_diff_code,
 		    unsigned int rotational_symmetry,
-		    unsigned int dimensionality);
+		    unsigned int dimensionality,
+		    bool normalize_points=true);
       constellation();
       virtual ~constellation();
 
@@ -253,11 +254,13 @@ namespace gr {
        *                      coding) (order of list matches constell)
        * \param rotational_symmetry Number of rotations around unit circle that have the same representation.
        * \param dimensionality Number of dimensions to the constellation.
+       * \param normalize_points Normalize constellation points to mean(abs(points))=1 (default is true)
        */
       static sptr make(std::vector<gr_complex> constell,
 		       std::vector<int> pre_diff_code,
 		       unsigned int rotational_symmetry,
-		       unsigned int dimensionality);
+		       unsigned int dimensionality,
+		       bool normalize_points=true);
 
       unsigned int decision_maker(const gr_complex *sample);
       // void calc_metric(gr_complex *sample, float *metric, trellis_metric_type_t type);
@@ -268,7 +271,8 @@ namespace gr {
       constellation_calcdist(std::vector<gr_complex> constell,
 			     std::vector<int> pre_diff_code,
 			     unsigned int rotational_symmetry,
-			     unsigned int dimensionality);
+			     unsigned int dimensionality,
+			     bool nomalize_points=true);
     };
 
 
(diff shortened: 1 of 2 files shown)


michaelld:
I like the concept of normalizing constellations, in general -- even though the end result is non-standard compared with (literal) standardized constellation definitions.

That said, the primary issue here is that, I believe, none of the actual constellations use this particular constructor (they all use the 0 argument one, I believe) -- and, hence this change doesn't actually do anything immediately useful for the constellations provided and used by GNU Radio. This is my read of the constellation classes; I might be incorrect here, of course ... there might be some constellation(s) that do(es) use this particular constructor. I tests BPSK & QPSK, and they do not benefit from this change; hence I would guess none of the other constellations would either.

hcab14:
> That said, the primary issue here is that, I believe, none of the actual constellations use this particular constructor (they all use the 0 argument one, I believe) -- and, hence this change doesn't actually do anything immediately useful for the constellations provided and used by GNU Radio. 

This constructor is used in `constellation_calcdist`
https://github.com/gnuradio/gnuradio/blob/master/gr-digital/lib/constellation.cc#L421
and all constellation classes inheriting  `constellation_calcdist`: `constellation_sector` `constellation_rect` `constellation_psk`.

> This is my read of the constellation classes; I might be incorrect here, of course ... there might be some constellation(s) that do(es) use this particular constructor. I tests BPSK & QPSK, and they do not benefit from this change; hence I would guess none of the other constellations would either.

This change should not affect any of the constellations used in gnuradio/gr-digital but allows to define QAM constellations using `constellation_calcdist` which do not get normalized if max(abs(points)) == 1. This is needed for the case when a channel equalizer is trained with 8PSK symbols and the following data symbols are in a given QAM constellation which is the case, _e.g._, for MIL-STD-188-110 Appendix C modes.

As far as I understand the main reason for normalizing constellations is to ensure that lookup tables can be generated using a fixed area (`constellation::gen_soft_dec_lut`). Note that the proposed change does not violate this assumption.

marcusmueller:
So, I do like that constellation, too, but as Michael says: This is not what a user would expect when asking for a QAM.

We should definitely have these edge-removed, rescaled, "outer-point-circularized" constellations; we just shouldn't be confusing those with the "plain old boring QAM".

So, what I'm a bit confused about is why you assign 1.0 the property to decide whether scaling should be happening. The scaling of my constellation really should effect which constellation I want, right?

hcab14:
>So, what I'm a bit confused about is why you assign 1.0 the property to decide whether scaling should be happening. The scaling of my constellation really should effect which constellation I want, right?

- mean(abs(points)) for such a constellation is < 1, so it would be scaled which I would like to avoid
- For what you call "plain boring QAM" constellations it makes sense to scale them because otherwise the assumption used when generating soft decision lookup tables is violated.

Checking for max(abs(points)) == 1 is one way of having both behaviors. An alternative would be to add an additional parameter to the constructor of `constellation_calc_dist` which allows to turn off rescaling with a default value such that rescaling is done.

michaelld:
I like the idea of this scaling being available but optional. I'd say to add it in as an optional parameter that is by default -disabled-, just to keep behavior the same as it is right now -- while providing the option to enable it for anyone interested.

You are correct that some constructors would use this addition -- those that use this specific (non-default) constructor as part of their instantiation. But the major constellations (*PSK and *QAM) -- as currently written -- use the default constructor, and hence would not benefit from this change.

michaelld:
@noc0lour @marcusmueller @mbr0wn any thoughts on your end on this PR? I'm good with merging it if you are.

michaelld:
I think you mis-force-pushed some UDP fixes to this PR. Please remove them & then I'll get this PR merged. Thx!

hcab14:
> I think you mis-force-pushed some UDP fixes to this PR. Please remove them & then I'll get this PR merged. Thx!

Yes you are right. I have now removed the unrelated commits.
Thanks

michaelld:
Great; thanks!

PR #2175, opened 2018-11-16
gr-utils: Condense gr_plot commands and fix datatype arg

btashton:
Remove a lot of duplicate commands that just called the same base functions with different data types.  The data type argument was broken as well and was always using `numpy.complex64`
Some of the plotting code was completely broken missing the required pylab/matplotlib imports.

In addition to this, I would be in favour of consolidating `_psd`, `_fft`, `_const` into the same command.


btashton:
Should I change the pylab stuff that was being used to call matplotlib instead.  Does not seem like we need to have pylab as a requirement.

btashton:
@mbr0wn no, but if we want to get rid of the matplotlib stuff this will all go away and be part of a single plotting tool.

After playing around a little with the existing plotting tool in that MR, I just started from scratch.  Too much old code and a lot of the settings did not really make sense.   It was also really easy to get it to crash.

There is no real dependency on gnuradio (except for some type formatting), so for now I have it over at https://github.com/btashton/grplot  There is nothing really to see yet.   We can fold it back into gnuradio once it is a little more stable.

mbr0wn:
@btashton I'm thinking we can still merge this just so we start reducing the plot-tools mess on master branch and so people see where this is going. What do you think? If you want to go straight to your single-plot-tool solution (which yes, thank god, is something we've needed for a while) then you can close this PR. Or, we merge this as an intermediate step.

btashton:
Lets merge it.

gr-utils/python/utils/plot_data.py
```diff
@@ -28,6 +28,24 @@
 
 import numpy
 
+try:
```
mbr0wn: This needs `import argparse`

gr-utils/python/utils/plot_data.py
```diff
@@ -152,6 +172,23 @@ def step_backward(self):
...
+                    "int16", "uint8", "int8"),
+                help="Specify the data type [default=%(default)r]")
+        parser.add_argument("-B", "--block", type=int, default=1000,
+                help="Specify the block size [default=%(default)r]")
+        parser.add_argument("-s", "--start", type=int, default=0,
+                help="Specify where to start in the file [default=%(default)r]")
+        parser.add_argument("-R", "--sample-rate", type=float, default=1.0,
+                help="Set the sampler rate of the data [default=%(default)r]")
+        parser.add_argument("file", metavar="FILE", nargs='+',
+                help="Input file with samples")
```
mbr0wn: Function doesn't return

gr-utils/python/utils/plot_data.py
```diff
@@ -152,6 +172,23 @@ def step_backward(self):
...
+                choices=("complex64", "float32", "uint32", "int32", "uint16",
+                    "int16", "uint8", "int8"),
+                help="Specify the data type [default=%(default)r]")
+        parser.add_argument("-B", "--block", type=int, default=1000,
+                help="Specify the block size [default=%(default)r]")
+        parser.add_argument("-s", "--start", type=int, default=0,
+                help="Specify where to start in the file [default=%(default)r]")
+        parser.add_argument("-R", "--sample-rate", type=float, default=1.0,
+                help="Set the sampler rate of the data [default=%(default)r]")
+        parser.add_argument("file", metavar="FILE", nargs='+',
```
mbr0wn: Called 'file' here but 'files' in `gr_plot`

gr-utils/python/utils/plot_data.py
```diff
@@ -152,6 +172,23 @@ def step_backward(self):
                 hf.seek(-hf.tell(),1)
         self.update_plots()
 
+    @staticmethod
+    def setup_options():
+        description = "Takes a GNU Radio binary file and displays the samples versus time. You can set the block size to specify how many points to read in at a time and the start position in the file. By default, the system assumes a sample rate of 1, so in time, each sample is plotted versus the sample number. To set a true time axis, set the sample rate (-R or --sample-rate) to the sample rate used when capturing the samples."
+
+        parser = ArgumentParser(conflict_handler="resolve", description=description)
+        parser.add_argument("-d", "--data-type", default="complex64",
```
mbr0wn: Called `data_type` here but `datatype` in `gr_plot`

mbr0wn:
OK let's merge it -- but this code doesn't even run here. I tried fixing all the errors that came up, but eventually gave up.

btashton:
I can take a more detailed look tomorrow, but it was very broken to start with.

gr-utils/python/utils/plot_data.py
btashton: Fixed as files for gr_plot others only take a single file.

gr-utils/python/utils/plot_data.py
btashton: fixed

gr-utils/python/utils/plot_data.py
btashton: yep, was also missing some other modules under pylab

btashton:
@mbr0wn Fixed a few issues including the ones you pointed out.  Did some basic functional testing against:
 * `gr_plot`
 * `gr_plot_fft`
 * `gr_plot_psd`

mbr0wn:
No idea what buildbot is on about.

mbr0wn:
Merged manually. 745b2b2d34cff22468ea9a4a785e82d3398dc4e6

Added some info to commit message.

btashton:
@mbr0wn were there any specific ones that stood out? In my testing I ran into issues with seeking out of bounds and data that was not compatible with settings in fft functions.

PR #2177, opened 2018-11-16
uhd: added asynchronous messages handling to the sink

ptrkrysik:
UHD asynchronous messages are needed in order to obtain information
about issues happening on Tx side (Underflows, Sequence Errors,
Time Errors).

This change adds handling of these messages to the usrp sink by adding async_event_loop() function executed in a separate thread (_async_event_thread) where these messages are:
- received from the USRP,
- transformed into PMT messages,
- sent to the new "async_msgs" message output port.

On the top level the output PMT messages are (message_type, content) pairs:
- message_type is always "uhd_async_msg",
- content is a dictionary with key names taken from uhd::async_metadata_t.
- there is no 'has_time_spec' as in PMT it's not needed. There just won't be 'time_spec' in the PMT message   when 'has_time_spec' is false.
- there is no 'user_payload' - but it can be easily added in the future as a PMT blob.

Example message from the 'async_msgs' output is below:
(uhd_async_msg (channel . 0) (time_spec 4 . 0.0608192) (event_code burst_ack))

**For the reviewer: please check especially if:**
- the place and the way PMT keys for everything async msgs related are created are ok (they were added at the top of usrp_sink_impl.h). Also maybe it would be good if these keys were available in some public header?
- format of the messages is acceptable,
- the way metadata.event_code is transformed into 'event_code' message is ok,
- the way the thread is created and destructed is ok.

### DIFF
--- a/gr-uhd/grc/gen_uhd_usrp_blocks.py
+++ b/gr-uhd/grc/gen_uhd_usrp_blocks.py
@@ -113,15 +113,20 @@ inputs:
     id: command
     optional: true
     hide: ${'$'}{hide_cmd_port}
-% if sourk == 'sink':
--   domain: stream
-% else:
+% if sourk == 'source':
 
 outputs:
--   domain: stream
 % endif
+-   domain: stream
     dtype: ${'$'}{type.type}
     multiplicity: ${'$'}{nchan}
+% if sourk == 'sink':
+
+outputs:
+-   domain: message
+    id: async_msgs
+    optional: true
+% endif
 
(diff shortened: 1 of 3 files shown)


mbr0wn:
Merging manually. I changed the following:
- Removed the `boost::bind` (was too annoyed by it :smile: )
- Updated the copyright header

ptrkrysik:
Good to know that usage of modern C++ features is now encouraged in GNU Radio :). I didn't know about it.

PR #2180, opened 2018-11-16
digital: fix OFDM CA 'output_is_shifted' parameter ...

michaelld:
... use in GRC and documentation. Simple, but useful!

### DIFF
--- a/gr-digital/grc/digital_ofdm_carrier_allocator_cvc.block.yml
+++ b/gr-digital/grc/digital_ofdm_carrier_allocator_cvc.block.yml
@@ -26,6 +26,12 @@ parameters:
     label: Length tag key
     dtype: string
     default: '"packet_len"'
+-   id: output_is_shifted
+    label: Shift Output
+    dtype: enum
+    default: 'True'
+    options: ['False', 'True']
+    option_labels: ['No', 'Yes']
 
 inputs:
 -   domain: stream
@@ -38,7 +44,8 @@ outputs:
 
 templates:
     imports: from gnuradio import digital
-    make: digital.ofdm_carrier_allocator_cvc(${fft_len}, ${occupied_carriers}, ${pilot_carriers},
-        ${pilot_symbols}, ${sync_words}, ${len_tag_key})
+    make: digital.ofdm_carrier_allocator_cvc(
+        ${fft_len}, ${occupied_carriers}, ${pilot_carriers},
+        ${pilot_symbols}, ${sync_words}, ${len_tag_key}, ${output_is_shifted})
 
 file_format: 1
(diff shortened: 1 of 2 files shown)


gr-digital/grc/digital_ofdm_carrier_allocator_cvc.block.yml
```diff
@@ -26,6 +26,12 @@ parameters:
     label: Length tag key
     dtype: string
     default: '"packet_len"'
+-   id: output_is_shifted
+    label: Shift Output
+    dtype: enum
```
haakov: Can also use `bool` here if you'd like to use a switch button instead of the drop-down menu. 

(In that case lines 33-34 can be removed)

michaelld: Had thought as much but this construct was the first I found & so that's what went with.

I can go either way: enum or bool. Is there an advantage / disadvantage to either? If I recall, in "old" GRC XML, using a bool type one couldn't set it to an arbitrary variable value, while using an enum one could. Maybe the same holds true for "new" GRC YML? I'm not familiar with YML yet.

marcusmueller: That's an interesting point. You're right, with the bool input widget, you can't instead use a variable name, which is kind of a bummer.

So: 

* `bool` is the "sensible" type here, but
* `enum` gives us the ability to actually use a variable at this point.

Hm. Should we be extending the `bool` entry widget to somehow accept text/variable names somehow? And how would that look preferably?

michaelld: My vote about extending the `bool` entry widget is: YES!!! Or, rather, I'd like to see an option added to -any- type that allows for using a variable; not just `bool` type ... but, I'd guess that most other types allow for it already (as they did in "old" GRC XML) ... so, augmenting `bool` might be just what's needed.

haakov: @michaelld @marcusmueller I haven't really used the old GRC much, could you explain what you mean by "using a variable"?

michaelld: Using a single variable or parameter (for a hier block) to set multiple block's settings. For example, `verbose` or `packet_length_tag_string`. A variable is used in just the local GRC context, while a parameter is settable in the hier block from a GRC context using the hier block. I'm a big fan of hier blocks and parameters, especially `verbose`, `debug`, `log`, and the like ... all of which are (or can be) `bool` types.

haakov: Oh right :smiley: I don't think I've seen that used anywhere yet, but it does indeed sound useful!

PR #2181, opened 2018-11-16
cmake: replace 'CMAKE_COMPILER_IS_GNUCXX' with CMAKE_CXX_COMPILER_ID

michaelld:
... and expand from just GNU to both GNU and Clang, since they work comparably for the flags being tested.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -193,16 +193,17 @@ include(CMakeOverloads)
 ########################################################################
 include(GrMiscUtils) #compiler flag check
 
-if(CMAKE_COMPILER_IS_GNUCXX AND NOT WIN32)
-    #http://gcc.gnu.org/wiki/Visibility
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -193,16 +193,17 @@ include(CMakeOverloads)
...
 include(GrMiscUtils) #compiler flag check
 
-if(CMAKE_COMPILER_IS_GNUCXX AND NOT WIN32)
-    #http://gcc.gnu.org/wiki/Visibility
-    GR_ADD_CXX_COMPILER_FLAG_IF_AVAILABLE(-fvisibility=hidden HAVE_VISIBILITY_HIDDEN)
-endif()
-
-if(CMAKE_COMPILER_IS_GNUCXX)
+if(CMAKE_CXX_COMPILER_ID MATCHES "Clang" OR
+   CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
```
mbr0wn: Why not use `MATCHES` here? Is this always exactly "GNU"?

marcusmueller: `man cmake-variables` says: yes, always exactly `GNU`.

PR #2184, opened 2018-11-16
runtime: fix use of boost::math::gcd and boost::integer::gcd

michaelld:
+ the former was augmented by the latter in Boost 1.58.0;
+ they do the same computation;
+ the former is being deprecated as of Boost 1.66.0, and recommends moving to the latter;
+ GR supports Boost 1.54.0 and newer, so we need to be able to support both APIs for now.


mbr0wn:
Labeled 'Bug Fix' because it doesn't work with all Boost versions without it, and that's a bug.

PR #2185, opened 2018-11-16
grc: silence import warnings caused by hier blocks on namespace renewal

btashton:


marcusmueller:
@skoslowski I must admit I'm having a hard time visualizing the solution you're describing there: 

What would one have to implement to Make This Work<sup>TM</sup> using block flags?

btashton:
Sounds like everyone is in agreement that what we have going on here is not great right now, I think there are some changes to the import interface that would clean up both the generator code and this code. I should have some time this weekend.

skoslowski:
Sure, so the plan is to 
- have the hier block generator add a flag marking it as hier block
- let the top block generator figure out if the fg is using such a block
- append hier_block_lib_dir to sys.path (in the platform?) or drop import errors here...

btashton:
@skoslowski I'm kind of uncomfortable with playing with `sys.path` unless we __really__ need to be doing that.  PEP 302 added a lot of functionality that we could leverage to make things a lot cleaner.  All of these blocks could then be namespaced under `gnuradio.hb.myhierblock` we would then register a handler to look these up in the specified directory.  I get that that might be more than we want to be doing right now.

Here is what I propose as a short term solution, we add the flag in (I can write this if it has not been already, I did not see anything when I looked). If the flag exists we skip the import.  If the import fails we look to see if the comment was there and then warn the user that they might need to regenerate the block.  If the comment is not there and the import failed we log the error with more context on what the import was that failed.

skoslowski:
Its always nice when people cite PEPs... =) 

IMO, the obvious short term solution is to restore the 3.7 behavior. If you also want to have the top block generator stop relying on that comment, awesome. But that is a separate thing (and PR).

We can, of course, have a more general discussion on the grc-generated hier-blocks and how to make these imports succeed. Although we don't really care about the imported objects (at design time), I can see benefit in reporting failed imports (instead of other blocks failing due to missing names).

I have experience with custom Finders and Loader for Python, so you find me intrigued. There is no need to implement anything right anyway. Please sketch out your proposal first.

dl1ksv:
At this moment the easiest way to avoid the error messages is to add ~/.grc_gnuradio to PYTHONPATH

btashton:
@skoslowski I add the logic and comment on why we are not catching ImportError there.  I do not see the message on my test flow graph.

Issue #1634, opened 2018-02-23
Replace PyQwt with PyQtGraph


noc0lour:
PyQwt is effectively unmaintained since 2014 and thus has no support for Qt5/Qwt6 and Python3. 
Fortunately there is PyQtGraph which could be used to replace all occurences of PyQwt.

Occurences:  
 - [ ] gr-filter/python/filter/design/filter_design.py
 - [ ] gr-filter/python/filter/gui/idealbanditems.py
 - [ ] gr-filter/python/filter/gui/pyqt_filter_stacked.py
 - [ ] gr-filter/python/filter/gui/polezero_plot.py
 - [ ] gr-qtgui/apps/plot_form.py
 - [ ] gr-qtgui/apps/plot_time_form.py
 - [ ] gr-utils/python/utils/pyqt_filter.py
 - [ ] gr-utils/python/utils/pyqt_plot.py
 - [ ] gr-zeromq/examples/python/gui.py

ekigwana:
Any progress on this? It'd be nice to be able to use the filter design tool.

noc0lour:
I barely looked into it, any help is appreciated.

ekigwana:
I have started to look at this and having only PyQt5 I have noticed that a few other changes will be required. From PyQt5  [gotchas](http://pyqt.sourceforge.net/Docs/PyQt5/gotchas.html) I noticed QStrings are gone and PyQt5 is not compatible with PyQt5. I forked the next branch and I believe it is targeting PyQT5 and Python v2, is that correct? Is PyQt4 support also required?

noc0lour:
If possible I'd recommend to have development in py2/py3 compat mode. With GNU Radio 3.8 the goal is to support py3 as well.

ekigwana:
To keep things reasonable since QT4 is EOL and I don't even have a system with it, I am targeting the following:

1. Python v2/v3
2. PyQt5
3. pyqtgraphics

ekigwana:
I have been working on the filter design tool and have a question below.

New "Filter Taps" plot in progress unless it is not required
![filter taps_new](https://user-images.githubusercontent.com/13283456/41314665-fb023912-6e7c-11e8-9461-aa9855da7b77.png)
Old "Filter Taps" plot
![filter taps_old](https://user-images.githubusercontent.com/13283456/41314683-0637b4ce-6e7d-11e8-9405-ef00b0565d98.png)
New "Impulse Response" plot
[...]

PR #2187, opened 2018-11-16
WIP: QT: Convert the gr_plot_qt application to use pyqtgraph

btashton:
Right now this is using a mixture of QwtPlot and pyqtgraph.  The time plot is using pyqtgraph and the other plots have been left using QwtPlot.  I would like some feedback on the time plot before converting the other two.


btashton:
I think the main missing feature is rectangle zoom instead of panning.
http://www.pyqtgraph.org/documentation/graphicsItems/viewbox.html#pyqtgraph.ViewBox.setMouseMode

mbr0wn:
Can you post some screenshots here? And can anyone just run the current version?

btashton:
select
sinplot
white_background

Here are a few showing some of the functionality.  Something is a little funky with the symbol syle (circle, triangle, etc...)

btashton:
I thought the spectrogram was going to be a little harder, but it looks trivial via ImageItem.

btashton:
Here are some of the old version, looks like I'm missing the legend 
old_select
sin_old

On the new one we get some cool features with right click like exporting the plot as images or csv data.
new_rightclick

btashton:
Yeah, Python 3 won't work because it still has PyQwt code in it. Tomorrow I'll try and get the rest of it converted over.  As for the matplotlib bit I think that's for the other plotting utilities.  The with the pyqtgraph plot if you right click it you can export matplotlib code. I would say if we do that we should just put all that functionality into this tool so there is a single plotting tool and not the 15 we have now in mixed States of function

noc0lour:
If I see this correctly this PR is supersed by https://github.com/btashton/grplot and we might merge the utility back in-tree?

noc0lour:
Or keep it in a separate project but maybe keep in in the `GR` org?

btashton:
I was going to get back to this after the new year. But yeah I'm open to whatever, if we move it in tree I'll need to change the tests from pytest and do the cmake stuff.

noc0lour:
I'm open for any. Basically I'm mainly concerned on how to get people to use & embrace it.

mbr0wn:
Please, please can we merge your tools into GNU Radio and replace all the crap we have!

btashton:
I have not forgotten.  Just trying to wrap up an FPGA project that I was working on over the holidays that is taking longer than expected (shocking I know).  I probably could get a MR together this weekend.

noc0lour:
Fixes another part of #1634.
@btashton how are things?

PR #2189, opened 2018-11-17
grc: Correctly toggle the visibility of blocks

dl1ksv:
grc does not initialize the visibility of disabled blocks.
This  sometimes leads to stranges effects.

Toggling the visibility does not update the flowgraph. So this takes effect not until a change
to the flowgraph is made that forces an update of the flowgraph.


grc/gui/Application.py
```diff
@@ -721,7 +725,7 @@ def flow_graph_update(fg=flow_graph):
                     print("could not terminate process: %d" % page.process.pid)
 
         elif action == Actions.PAGE_CHANGE:  # pass and run the global actions
-            pass
+            flow_graph_update()
```
marcusmueller: that's the right thing to do, here, I guess, but then the comment the line above is obsolete :)

PR #2190, opened 2018-11-17
Alternative solution for pr 2185

dl1ksv:
This is an alternative solution for #2185 without using comments on import statements.

### DIFF
--- a/grc/core/FlowGraph.py
+++ b/grc/core/FlowGraph.py
@@ -20,6 +20,7 @@ from __future__ import absolute_import, print_function
 import collections
 import itertools
 import sys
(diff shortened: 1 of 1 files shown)


skoslowski:
thanks for your effort here, but this is not going in the right direction (IMO). We don't want to be doing a fs lookup here. Also, there is not separate interpreter for the FG evals. That means there is only one sys module and only one set of paths to lookup imports. There is no point in adding the same fixed path over and over. As I mentioned on @btashton s PR, if we really want these imports to succeed we can add that path, once, on GRC startup (in conf.py) and be done.

PR #2193, opened 2018-11-21
cmake: FindQwt: Allow pkg-config to set the library name

mbr0wn:
Currently, the library name is hard-coded to qwt-qt5 to force
disambiguation from the Qt4 version. However, when pkg-config is
invoked, other package names (e.g. qwt) are also permissable, since we
trust pkg-config to not point us to a Qt4 version of Qwt.

This will let @balister build on OE.


marcusmueller:
That sounds right! But shouldn't this imply that we use the pkgconfig-found results with highest priority?

noc0lour:
Are results in `NAMES` priority sorted?

mbr0wn:
Good question. I think they are.

mbr0wn:
Actually, the CMake docs say nothing about this. However, I'll still put them front, even if it's just to document the intention.

mbr0wn:
I changed the order to use the PC vars first, and tested on a system without the PC file to make sure the empty var doesn't screw things up.

michaelld:
CMake should search in the order provided. So listing the PC provided library name first is the correct way to go here.

PR #2195, opened 2018-11-22
fft: Fix GRC bindings

mbr0wn:
- 'hide' attribute for fftshift was supposed to default to on for real
  mode. It now really does.
- The actual FFT block doesn't have a vlen parameter, and it was causing
  havok in the GRC bindings. It's now removed.


cclamb:
This should be reopened, but with fft_logpwrfft_x.block.yml; fix seems to be the same.

cclamb:
(I just removed the offending line in fft_logpwrfft_x.block.yml and it fixed the issue)

PR #2196, opened 2018-11-22
qtgui: time_sink: Fix GRC bindings for channels 3 and up

mbr0wn:
- YML now auto-generated via CMake
- hide attributes match chan count
- Usable defaults


marcusmueller:
Hey, I'm very thankful that you fix this bug!

But tbh, @noc0lour just removed gengen from our C++ world, and I honestly don't think that YAML strings in a Python file are an overly great way forward; this breaks test-/parseability by tooling prior to build.

So, I'm really not all that happy with the specific approach of hiding one language in strings in another. I think that's an architectural approach that we should really avoid.

On the other hand, as the fixed issue illustrates, code duplication within YAML files is _really_ bad.

I'll go ahead and say: wouldn't it be, as integral over our codebase, much better if GRC understood a `generate({variable name}, {python generator code})` property for ports, parameters and options?

I think @haakov 's and @skoslowski 's input would be most valueable here.

marcusmueller:
Hey, I'm very thankful that you fix this bug!

But tbh, @noc0lour just removed gengen from our C++ world, and I honestly don't think that YAML strings in a Python file are an overly great way forward; this breaks test-/parseability by tooling prior to build.

So, I'm really not all that happy with the specific approach of hiding one language in strings in another. I think that's an architectural approach that we should really avoid.

On the other hand, as the fixed issue illustrates, code duplication within YAML files is _really_ bad.

I'll go ahead and say: wouldn't it be, as integral over our codebase, much better if GRC understood a `generate({variable name}, {python generator code})` property for ports, parameters and options?

I think @haakov 's and @skoslowski 's input would be most valueable here.

btashton:
There are some yaml features that I don't think we are using that might help a lot with the code duplication. I suspect that we could have some high level yaml file the defines useful anchors that the blocks then use and extend. 
https://blog.daemonl.com/2016/02/yaml.html

btashton:
Anstible might also be something to get some inspiration from. The playbooks are quite advanced and defined via yaml. 
https://docs.ansible.com/ansible/latest/user_guide/playbooks.html#working-with-playbooks

noc0lour:
Also some min/max feature which mimics the io_signature might be useful here. Where the minimum number of some port is given and the maximum number and then you can supply a list of names which will be given. I don't know how the user interface can be designed to have that flexibility. Maybe if you hover over the block a new empty port appears where you could connect a new connection (given you are in range of [min,max])?

skoslowski:
For me it feels like, the qtgui blocks have always used GRC markup to it's limits. Blocks wrappers of this complexity are rare, so the question is whether saving some lines of (generated) code is worth putting effort into let alone maintaining it.

That said, an ansible like solution sounds like it could work. Since in GRC each block wrapper ends up a subclass of Block, we could experiment with this or anyelse that helps these specific blocks by creating a special base QtGuiBlock in do the extra logic.

mbr0wn:
@marcusmueller Would an acceptable intermediate solution be to check in the script output and remove the YAML file generation with CMake?

FYI, the variable struct is also using this (that's where I copied the CMake from).

mbr0wn:
I would like to not have an architectural discussion in this thread for a PR that fixes a simple bug. How about this, does this work (keeping the Python script, but checking in the yml file)? We can have a separate discussion on the architectural philosophy in a github issue.

PR #2197, opened 2018-11-24
grc: Correctly building dummy block

### LINKED ISSUE #2160: grc:does not correcly  builds dummy blocks
If a flowgraph contains a block without a block description ( perhaps depreceated in 3.7 and now missing , or missing from other reasons ) grc should build a dummy block. https://github.com/gnuradio/gnuradio/blob/b1e0b54d02bb981eba734ea29174e6134950dde6/grc/core/FlowGraph.py#L371

But the dummy block will not be build:

    Loading: "/home/schroer/gnuradiocomponents/gr3.8tests/OptionalPadTest.grc"
    Block id "hier2Msg" not found
    Error: 'id'
>>> Failure
  File "/usr/local/gnuradio/lib/python3.6/dist-packages/gnuradio/grc/gui/MainWindow.py", line 261, in new_page
    file_path=file_path,
  File "/usr/local/gnuradio/lib/python3.6/dist-packages/gnuradio/grc/gui/Notebook.py", line 107, in __init__
    flow_graph.import_data(initial_state)
  File "/usr/local/gnuradio/lib/python3.6/dist-packages/gnuradio/grc/core/FlowGraph.py", line 377, in import_data
    block.import_data(**block_data)
  File "/usr/local/gnuradio/lib/python3.6/dist-packages/gnuradio/grc/core/blocks/block.py", line 344, in import_data
    self.params['id'].value = name
KeyError: 'id'

and in consequence, the flowgraph will not be build.
The fowgraph isn't   build even if the missing block is disabled.
So you have no chance to edit this  flowgraph and make it work again.

The only way to get this flowgraph building is to modify the grc file by hand.

dl1ksv:
Flowgraphs referencing unknown blocks fail to build and display as the dummy block that should be build instead of the missing block misses some parameters.

This fixes #2160

### DIFF
--- a/grc/core/FlowGraph.py
+++ b/grc/core/FlowGraph.py
@@ -371,7 +371,7 @@ class FlowGraph(Element):
             )
 
             if isinstance(block, blocks.DummyBlock):
-                print('Block id "%s" not found' % block_id)
+                block.add_error_message('Block id "{}" not found.'.format(block_id))
 
             block.import_data(**block_data)
 
--- a/grc/core/FlowGraph.py
+++ b/grc/core/FlowGraph.py
@@ -370,9 +370,6 @@ class FlowGraph(Element):
                 self.new_block(block_id='_dummy', missing_block_id=block_id, **block_data)
             )
 
-            if isinstance(block, blocks.DummyBlock):
-                block.add_error_message('Block id "{}" not found.'.format(block_id))
-
             block.import_data(**block_data)
 
         self.rewrite()  # evaluate stuff like nports before adding connections
@@ -417,6 +414,14 @@ class FlowGraph(Element):
                 'Connection between {}({}) and {}({}) could not be made.\n\t{}'.format(
                     src_blk_id, src_port_id, snk_blk_id, snk_port_id, e))
             had_connect_errors = True
+
+        for block in self.blocks:
+            if block.is_dummy_block :
+                block.rewrite()      # Make ports visible
+                # Flowgraph errors depending on disabled blocks are not displayed
+                # in ther error dialog box
+                # So put a messsage into the Property window of the dummy block
+                block.add_error_message('Block id "{}" not found.'.format(block.key))
 
         self.rewrite()  # global rewrite
         return had_connect_errors
--- a/grc/core/blocks/dummy.py
+++ b/grc/core/blocks/dummy.py
@@ -19,6 +19,7 @@ from __future__ import absolute_import
 
 from . import Block, register_build_in
 
+from ._build import _build_params, _build_ports
 
 @register_build_in
 class DummyBlock(Block):
@@ -28,9 +29,12 @@ class DummyBlock(Block):
     label = 'Missing Block'
     key = '_dummy'
 
-    def __init__(self, parent, missing_block_id, parameters, **_):
+    def __init__(self, parent, missing_block_id, parameters,**kwargs):
+
         self.key = missing_block_id
+        self.parameters_data = _build_params([kwargs],False, False,self.flags, self.key)
         super(DummyBlock, self).__init__(parent=parent)
+        del self.params[self.key] # Remove missing keys from parameter list
 
         param_factory = self.parent_platform.make_param
         for param_id in parameters:
@@ -42,7 +46,7 @@ class DummyBlock(Block):
     @property
     def enabled(self):
         return False
-
+        
     def add_missing_port(self, port_id, direction):
         port = self.parent_platform.make_port(
             parent=self, direction=direction, id=port_id, name='?', dtype='',
@@ -51,4 +55,6 @@ class DummyBlock(Block):
             self.sources.append(port)
         else:
             self.sinks.append(port)
+        self.rewrite()     # Make port visible
+        self.validate()    # Generate appropriate error messages     
         return port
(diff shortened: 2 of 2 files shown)


grc/core/FlowGraph.py
```diff
@@ -371,7 +371,7 @@ def import_data(self, data):
             )
 
             if isinstance(block, blocks.DummyBlock):
-                print('Block id "%s" not found' % block_id)
+                block.add_error_message('Block id "{}" not found.'.format(block_id))
```
skoslowski: IMO this error should be attached to fg, `self.add_error_message`

grc/core/blocks/dummy.py
```diff
@@ -19,6 +19,7 @@
 
 from . import Block, register_build_in
 
+from ._build import _build_params, _build_ports
```
skoslowski: lets not import protected functions from another module. If these are required, they'll have to be renamed

grc/core/blocks/dummy.py
```diff
@@ -28,9 +29,12 @@ class DummyBlock(Block):
     label = 'Missing Block'
     key = '_dummy'
 
-    def __init__(self, parent, missing_block_id, parameters, **_):
+    def __init__(self, parent, missing_block_id, parameters,**kwargs):
+
         self.key = missing_block_id
+        self.parameters_data = _build_params([kwargs],False, False,self.flags, self.key)
```
skoslowski: What are you trying to do here exactly? I can see the advantage of adding the the standard params with this call (known type, category, ...). What I don't get is why you need to pass in kwargs? What is this supposed to be doing? 
Since the code in line 40 is supposed to be doing just that, I believe, an empty list should be passed here

grc/core/blocks/dummy.py
```diff
@@ -42,7 +46,7 @@ def is_valid(self):
     @property
     def enabled(self):
         return False
-
+
```
skoslowski: ?!?

grc/core/blocks/dummy.py
skoslowski: ...or properly replace that code by constructing params_raw appropriately

grc/core/blocks/dummy.py
```diff
@@ -28,9 +29,12 @@ class DummyBlock(Block):
     label = 'Missing Block'
     key = '_dummy'
 
-    def __init__(self, parent, missing_block_id, parameters, **_):
+    def __init__(self, parent, missing_block_id, parameters,**kwargs):
+
         self.key = missing_block_id
+        self.parameters_data = _build_params([kwargs],False, False,self.flags, self.key)
         super(DummyBlock, self).__init__(parent=parent)
+        del self.params[self.key] # Remove missing keys from parameter list
```
skoslowski: I suspect this can be removed if _build_params is used correctly

grc/core/blocks/dummy.py
```diff
@@ -51,4 +55,6 @@ def add_missing_port(self, port_id, direction):
             self.sources.append(port)
         else:
             self.sinks.append(port)
+        self.rewrite()     # Make port visible
+        self.validate()    # Generate appropriate error messages
```
skoslowski: I am not sure if these should be placed here. Maybe better on the call-site in FlowGraph.import_data

grc/core/blocks/dummy.py
haakov: Might be worth noting that I did the same when fixing the virtual blocks. I can rename them :+1: 

https://github.com/gnuradio/gnuradio/blob/df4e7a85263eb523395ce6bef0b2e7769bd63465/grc/core/blocks/virtual.py#L23

grc/core/FlowGraph.py
dl1ksv: I changed this because I wanted to get an error message into grc either in the errors and warnings window or into the Properties window of the selected block, and not into the terminal window running grc.
But neither block.add_error_messages nor self.add_error_message delivers an error message.
I have no idea where this message goes to.

So I'd propose to this message completely.

grc/core/blocks/dummy.py
dl1ksv: As I'm new to python I indeed had a look how other blocks handle this. In this case it was @haakov 's virtual block.

Renaming the function is no problem.

grc/core/blocks/dummy.py
dl1ksv: @skoslowski You are right, kwargs is not needed. An empty list does the trick and I don't have to remove self.params[self.key]

But I wonder if this piece of code should be part of the block's __init__ function, as the generated parameters belong  to a core part of the block and each sublcassed block.

marcusmueller:
uh oh, @dl1ksv and @skoslowski: feels like there was much authoring and reviewing effort put into this, but then it kind of got lost. @skoslowski, what about the changes in the latest commit (which came post-review)? @dl1ksv is this commit addressing @skoslowski's comments?

dl1ksv:
@marcusmueller  Yes, the latest commit reflects the comments.

dl1ksv:
As I failed to sum up both commits I made a new pr #2278 that replaces this one.

mbr0wn:
Superseded by #2278.

Issue #2120, opened 2018-11-01
Fix compiler warnings about unsigned / signed comparisons


marcusmueller:
#2069, as desired, introduced a lot of warnings into the build process, essentially everywhere we should've been using `unsigned` vs `signed`.

Now that we identified these places, we need to fix' em.

Essentially, this is the process of:

* pick one compiler warning about signed/unsigned comparison
* track down which side of the comparison is right, i.e. should the data types compared be signed or unsigned
* modify the types not to eliminate the warning but to make sure the comparison is logically sound.

Fadila82:
Hello, is this done ? I tried to do it as my first issue, but the only warnings I got were from Volk. Should I do it in volk then ?

noc0lour:
@Fadila82 building with `-DCMAKE_BUILD_TYPE=Debug` might reveal those warnings

Fadila82:
Of course, it makes sense. Thank you.

noc0lour:
Sure, thank you for looking at this!

kaydeetwo:
Hi - is there still need for help here? I'd be keen to tackle it.

michaelld:
@kaydeetwo There is -always- a need for help fixing warnings! Each compiler suite provides similar yet different warnings, even between same-compiler but different versions (e.g., GCC7 versus GCC8). Whatever help you want to provide would be awesome! If you think you'll participate in GR development beyond just a few such warnings fixed here and there, then you'll need a "CLA" (license assignment) allowing you to commit code to the project; if so let us know & we'll get you in touch with the correct GR person to do so. Thx!

kaydeetwo:
@michaelld - thanks and sounds great!!

PR #2198, opened 2018-11-24
Fix compiler warnings about unsigned / signed comparisons

dl1ksv:
See #2120
Fixing compiler warnings, excluding volk. In submodule volk there exist more warnings.

### DIFF
--- a/gr-trellis/lib/interleaver.cc
+++ b/gr-trellis/lib/interleaver.cc
@@ -113,13 +113,13 @@ namespace gr {
       std::vector<int> tmp(d_K);
       unsigned char *bytes = reinterpret_cast<unsigned char*>(&tmp[0]);
 
-      for(unsigned int i = 0; i < d_K; i += 8) {
+      for(int i = 0; i < d_K; i += 8) {
               *(reinterpret_cast<uint64_t*>(bytes + i)) = xoroshiro128p_next(rng_state);
       }
       if(d_K % 8) {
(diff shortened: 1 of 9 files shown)


gr-trellis/lib/interleaver.cc
```diff
@@ -113,13 +113,13 @@ namespace gr {
       std::vector<int> tmp(d_K);
       unsigned char *bytes = reinterpret_cast<unsigned char*>(&tmp[0]);
 
-      for(unsigned int i = 0; i < d_K; i += 8) {
+      for(int i = 0; i < d_K; i += 8) {
```
marcusmueller: This single change is surprising, as it takes unsignedness away from something that just counts upwards; I'll do the following: I'll merge your change, but will then consistify `d_K` to be unsigned all throughout the trellis interleaver usage.

dl1ksv: d_k is declared as int and there are a lot of comparisons to int's.
For instance https://github.com/gnuradio/gnuradio/blob/b00ebcae4cf2047ba9bf8d7d2d16f9053fcd45b7/gr-trellis/lib/interleaver.cc#L87

So it makes sence to remove the 'unsigned' at this point.

On the other side d_k may not become negative, so d_k could be changed to unsigned int.
If you think that will increase code quality, I could change this, but this will have effects even to calling functions.

PR #2199, opened 2018-11-24
trellis interleaver: made table size unsigned int

marcusmueller:
This complements the excellent work of #2198.

This is part of the ongoing effort to avoid signedness confusion.

It completes df4e7a85263eb523395ce6bef0b2e7769bd63465 in that it at
least strives to handle sizes consistently in interleaver.cc.

Of course, the returned vectors still contain signed ints where there
should be unsigned ints, but the trellis core_algorithms.cc code is
convoluted enough to warrant carefulness when touching, so this part of
the API was kept.

### DIFF
--- a/gr-trellis/include/gnuradio/trellis/interleaver.h
+++ b/gr-trellis/include/gnuradio/trellis/interleaver.h
@@ -37,17 +37,17 @@ namespace gr {
     class TRELLIS_API interleaver
     {
     private:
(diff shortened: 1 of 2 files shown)


mbr0wn:
Although we use `unsigned int` a lot where we /should/ be using `size_t`.

mbr0wn:
No complaints, so let's merge.

PR #2200, opened 2018-11-26
Add UHD multi_usrp filter APIs to gr-uhd

dkozel:
This PR will be complete when it provides the ability to list, get, and set all filters supported and used by UHD.

- [x] APIs exposed in C++ USRP Source/Sink blocks
- [x] Add IFDEF protection for UHD versions < 3.9.0
- [x] APIs wrapped into Python

EDIT: Moving these to another issue/PR
- Write a Python example
- Functionality exposed (where useful) in GRC
- Write a GRC example (will be B2xx/E3xx specific)

I'm thinking of adding a text field to the GRC blocks which accepts a Python dictionary with `Filter Path: taps`. This provides a simple way of setting custom filters.

I'm testing the above with a B205mini.

@mbr0wn the SWIG wrapping is a bit clumsy at the moment, but I think it is about as good as it can be, second opinions much appreciated. Also any thoughts on the GRC UI. In a future PR I'd like to add a control port based API for getting and setting the filters but that's "future work".

### DIFF
--- a/gr-uhd/lib/CMakeLists.txt
+++ b/gr-uhd/lib/CMakeLists.txt
@@ -77,6 +77,10 @@ add_library(gnuradio-uhd SHARED ${gr_uhd_sources})
 target_link_libraries(gnuradio-uhd ${uhd_libs})
 GR_LIBRARY_FOO(gnuradio-uhd RUNTIME_COMPONENT "uhd_runtime" DEVEL_COMPONENT "uhd_devel")
 
+if(UHD_VERSION VERSION_GREATER_EQUAL "3.9.0")
+  add_compile_definitions(UHD_USRP_MULTI_FILTER_API)
+endif(UHD_VERSION VERSION_GREATER_EQUAL "3.9.0")
+
 ########################################################################
 # Build static libraries if enabled
 ########################################################################
--- a/gr-uhd/lib/usrp_block_impl.cc
+++ b/gr-uhd/lib/usrp_block_impl.cc
@@ -353,7 +353,7 @@ usrp_block_impl::get_gpio_attr(
 std::vector<std::string>
 usrp_block_impl::get_filter_names(const std::string &search_mask)
 {
-#if UHD_USRP_MULTI_FILTER_API
+#ifdef UHD_USRP_MULTI_FILTER_API
     return _dev->get_filter_names(search_mask);
 #else
   throw std::runtime_error("not implemented in this version");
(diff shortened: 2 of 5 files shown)


dkozel:
@mbr0wn I'd like to get your review on this. I think it represents a useful addition and is worth merging ahead of an example and the GRC wrapper.

gr-uhd/lib/usrp_block_impl.cc
```diff
@@ -389,6 +389,36 @@ usrp_block_impl::get_gpio_attr(
 #endif
 }
 
+std::vector<std::string>
+usrp_block_impl::get_filter_names(const std::string &search_mask)
+{
+#if UHD_USRP_MULTI_FILTER_API
```
mbr0wn: Shouldn't these be `ifdef`?

mbr0wn:
Did you want to merge this before you finish the Python part and the examples?

dkozel:
@mbr0wn Yes, I'd like to go ahead with merging this. GRC support and examples can come "later". The C++ and Python wrappers have been tested and function as expected.

mbr0wn:
Can you rebase and remove the don't merge tag? Hopefully we get more green checkmarks. I'll review then and hopefully merge.

gr-uhd/lib/CMakeLists.txt
```diff
@@ -77,6 +77,10 @@ add_library(gnuradio-uhd SHARED ${gr_uhd_sources})
 target_link_libraries(gnuradio-uhd ${uhd_libs})
 GR_LIBRARY_FOO(gnuradio-uhd RUNTIME_COMPONENT "uhd_runtime" DEVEL_COMPONENT "uhd_devel")
 
+if(UHD_VERSION VERSION_GREATER_EQUAL "3.9.0")
+  add_compile_definitions(UHD_USRP_MULTI_FILTER_API)
```
ncorgan: Is there a difference between this and the existing #define here? At first glance, they appear to refer to the same feature.

https://github.com/EttusResearch/uhd/blob/release_003_009_000/host/include/uhd/usrp/multi_usrp.hpp#L33

dkozel: ! I will take a look. I hadn't noticed those defines in UHD.

The PR has been stalled because `add_compile_definitions` isn't supported in our (current) minimum cmake version, but I think the current cmake PR actually bumps the min up to a version that does have it. If I can use this define then it's irrelevant.

noc0lour: Actually I'd prefer using `target_compile_definitions(gnuradio-uhd PUBLIC UHD_USRP_MULTI_FILTER_API)`  which will transitively add that to the exported targets. (If public headers in UHD or gnuradio-uhd contain ifdef's based on that define).

dkozel: Cool to know about that option. I think the UHD define will work, but am test building my change  now with various UHD versions.

noc0lour: Yes, probably the define in the UHD header is sufficient in this case (or even the right way).

mbr0wn:
Continuing here: https://github.com/gnuradio/gnuradio/pull/3052

PR #2203, opened 2018-11-27
WIP: Unsigned/signed warnings

Fadila82:
I was working on issue #2198 , but I've seen someone else has already merged a pull request for this issue.

But I was going to use size_t where it seemed to make sense. 

I'm still working on it but I wonder if I should go on or is it just a waste of time?

### DIFF
--- a/gnuradio-runtime/include/gnuradio/sync_block.h
+++ b/gnuradio-runtime/include/gnuradio/sync_block.h
@@ -43,13 +43,24 @@ namespace gr {
                gr::io_signature::sptr output_signature);
 
   public:
+
     /*!
      * \brief just like gr::block::general_work, only this arranges to
      * call consume_each for you
      *
      * The user must override work to define the signal processing code
      */
-    virtual int work(int noutput_items,
+    virtual int work(size_t noutput_items,
+                    gr_vector_const_void_star &input_items,
+                    gr_vector_void_star &output_items);
+
+    /*!
+     * \brief just like gr::block::general_work, only this arranges to
+     * call consume_each for you
+     *
+     * The user must override work to define the signal processing code
+     */
+    virtual int __GR_ATTR_DEPRECATED work(int noutput_items,
                      gr_vector_const_void_star &input_items,
                      gr_vector_void_star &output_items) = 0;
 
--- a/gnuradio-runtime/include/gnuradio/tagged_stream_block.h
+++ b/gnuradio-runtime/include/gnuradio/tagged_stream_block.h
@@ -134,7 +134,12 @@ namespace gr {
      * \param input_items See gr::block
      * \param output_items See gr::block
      */
-    virtual int work(size_t noutput_items,
+    int work(size_t noutput_items,
+                     gr_vector_int &ninput_items,
+                     gr_vector_const_void_star &input_items,
+                     gr_vector_void_star &output_items);
+
+    virtual int __GR_ATTR_DEPRECATED work(int noutput_items,
                      gr_vector_int &ninput_items,
                      gr_vector_const_void_star &input_items,
                      gr_vector_void_star &output_items) = 0;
--- a/gnuradio-runtime/include/gnuradio/tagged_stream_block.h
+++ b/gnuradio-runtime/include/gnuradio/tagged_stream_block.h
@@ -134,7 +134,7 @@ namespace gr {
      * \param input_items See gr::block
      * \param output_items See gr::block
      */
-    virtual int work(int noutput_items,
+    virtual int work(size_t noutput_items,
                      gr_vector_int &ninput_items,
                      gr_vector_const_void_star &input_items,
                      gr_vector_void_star &output_items) = 0;
--- a/gr-digital/lib/protocol_formatter_bb_impl.cc
+++ b/gr-digital/lib/protocol_formatter_bb_impl.cc
@@ -67,7 +67,7 @@ namespace gr {
     }
 
     int
-    protocol_formatter_bb_impl::work(int noutput_items,
+    protocol_formatter_bb_impl::work(size_t noutput_items,
                                      gr_vector_int &ninput_items,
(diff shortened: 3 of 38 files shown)


gnuradio-runtime/include/gnuradio/tagged_stream_block.h
```diff
@@ -134,7 +134,7 @@ namespace gr {
      * \param input_items See gr::block
      * \param output_items See gr::block
      */
-    virtual int work(int noutput_items,
+    virtual int work(size_t noutput_items,
```
michaelld: Any changes to "work" or "general_work" like this one have the potential to break a -lot- of OOT modules because the method prototype will be different between the initial declaration here and the one trying to be overloaded in the OOT. Need to be -very- careful here!

michaelld:
I love the idea of moving obvious variables from signed to unsigned! All the rest here is IMHO and I'm open for others thoughts on the matter:

We cannot change the API for "work" since it is so critical to how GR functions. If we want to change GR's "work" or "general_work" API, the right way to do this would be to create "work_unsigned" (or whatever alternative name makes sense), which provides the unsigned API (and is overloadable just like current "work" is), and which by default does type casting and calls the "signed" work method. In this manner, we can maintain current API compatibility that is used by -all- current OOT modules, while still providing a path forward with a correct unsigned API method. We can somehow mark the "old" (signed) version as deprecated to get OOT developers to start moving to the "new" (unsigned) API.

michaelld:
I was going for unique method names, but yes we should be able to use same-named methods with unique prototypes.

noc0lour:
And the old prototypes can be marked with `__GR_ATTR_DEPRECATED`

noc0lour:
Also, in 3.8 API changes should be allowed. (Of course while trying to ease the migration).

michaelld:
I love the deprecated attribute ... very nice!

michaelld:
Yes, we can break the API for 3.8 ... not sure we want to break the "work" API though ... it's so critical to how GR works!

michaelld:
I'm wondering what happens if we break the API by removing the "signed" API ... will the OOT build cleanly but just not work? That's what I'm trying to avoid by deprecating the current API for a while before either disabling it or removing it.

noc0lour:
I think we can do 2 things: Remove the signed completely for blocks inside of GR. and Just deprecate the signed API in the common blocks. (a.k.a sync, general, sync_decimator, ...)
This will allow OOTs to compile. But we remove the signed functions from blocks included in the tree.

michaelld:
Agreed; that should work.

Fadila82:
I am working on getting it done.

Fadila82:
I have started to replace the work(int noutput_items) to work(size_t noutput_items).

However, I'm not sure I'm doing it well. Can someone have a look ?

As I did there (leaving work(int) as a pure virtual method), I have to add an implementation of the work(int noutput_items) to each block in the tree, which does not seem a good idea.

The solution I see is to implement the work(int) method and call work(size_t) from there. But then there are no virtual method left, and thus no obligation to override the work method.

This leads me to think I got something wrong. Can someone explain to it what it is ?

michaelld:
I think the following will work, allowing gradual deprecation of the `int` API.

Drop back a level to `block_executor.cc:501` where `general_work` is called, and change it to use the `size_t` API instead of `int`. Then, in `block.*`, provide both `general_work(int)` and `general_work(size_t)` as virtual methods. By default, `general_work(size_t)` calls `general_work(int)`, which doesn't need to be changed (it should error out if not defined, and be overloaded to provide [to be deprecated] functionality as currently done). Both APIs need to be provided in the GR-provided major block types (either `general_work` or `work`; e.g., sync_block, tagged_stream_block, sync_decimator, sync_interpolator, block_gateway), with the `size_t` version calling the `int` one by default. Otherwise, all internal GR-`_impl` block should use the new `size_t` API. The primary "catch" that I can think of is that each inheriting major block type needs to correctly overload the `size_t` API so that it is what will be used -first- instead of the `int` version (which will be called if the `size_t` version isn't overloaded).

Fadila82:
Ok. Thank you for your clear explanation.

mbr0wn:
I would like to get @marcusmueller's input on this. Personally, I would suggest not touching the signature of `work()` and `general_work()`. Where do the warnings come from? Shouldn't we rather fix that?

The return value of this function is nominally also the number of produced items. However, for `general_work()`, we can also return negative values which have a special meaning. Who says the same can't be true on the input side? We might want to call blocks with `ninput_items == -1` which would have a special meaning.

gnuradio-runtime/include/gnuradio/sync_block.h
```diff
@@ -43,13 +43,24 @@ namespace gr {
...
 
   public:
+
+    /*!
+     * \brief just like gr::block::general_work, only this arranges to
+     * call consume_each for you
+     *
+     * The user must override work to define the signal processing code
+     */
+    virtual int work(size_t noutput_items,
```
mbr0wn: It's already inconsistent to change `work()` and not `general_work()`.

gr-digital/lib/protocol_formatter_bb_impl.cc
```diff
@@ -67,7 +67,7 @@ namespace gr {
     }
 
     int
-    protocol_formatter_bb_impl::work(int noutput_items,
+    protocol_formatter_bb_impl::work(size_t noutput_items,
```
mbr0wn: We have to change almost every block if we change this. In total, this seems more work than just doing `static_cast<>()` whenever we get a signed/unsigned warning.

marcusmueller:
So, I haven't read all this, and I'll definitely do, but just to not forget: I agree with @mbr0wn, changing `work`/`general_work` signature won't happen before 3.9. There's still certainly value in this, and I will be looking into the issue at hand.

mbr0wn:
Closing this, since we can't get an agreement on how to continue.

PR #2204, opened 2018-11-28
gr-zeromq: Remove VLA from last_endpoint()

btashton:
The zeromq code as a whole code repeats itself a lot, but changing that would mean creating some new base classes and doing quite a bit of restructuring that I am not very comfortable doing.   This should remove the variable length array that is not allowed by MSVC.

### DIFF
--- a/gr-zeromq/lib/base_impl.cc
+++ b/gr-zeromq/lib/base_impl.cc
@@ -57,8 +57,8 @@ namespace gr {
     std::string
     base_impl::last_endpoint()
     {
-      size_t addr_len = 256;
-      char addr[addr_len];
+      char addr[256];
+      size_t addr_len = sizeof(addr);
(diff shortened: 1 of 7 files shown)


btashton:
@gvanem This should resolve your issue, but I don't have a MSVC environment to test it on.

gvanem:
@btashton Your diff looks okay (untested though).

gr-zeromq/lib/base_impl.cc
```diff
@@ -57,8 +57,8 @@ namespace gr {
     std::string
     base_impl::last_endpoint()
     {
-      size_t addr_len = 256;
-      char addr[addr_len];
+      char addr[256];
```
mbr0wn: Or, you `constexpr` the first line and keep it as-is.

btashton:
A comment that was made here: https://github.com/gnuradio/gnuradio/commit/b76a95c52f812181605248f1f590a199e2e5d0e3#commitcomment-31490370

PR #2205, opened 2018-11-29
udp: wait a short time after creating starting the receiver

noc0lour:
Running unit tests the sender socket could be created before the receiver socket
was ready. Try to avoid that by adding a sleep in between the receiver socket
creation and the sender socket creation.

### DIFF
--- a/gr-blocks/python/blocks/qa_udp_source_sink.py
+++ b/gr-blocks/python/blocks/qa_udp_source_sink.py
@@ -23,6 +23,7 @@
 
 from gnuradio import gr, gr_unittest, blocks
 import os
+import time
 
 from threading import Timer
 
@@ -71,6 +72,7 @@ class test_udp_sink_source(gr_unittest.TestCase):
         self.tb_rcv.connect(udp_rcv, dst)
 
         self.tb_rcv.start()
+        time.sleep(0.1)
         self.tb_snd.run()
         udp_snd.disconnect()
         self.timeout = False
@@ -102,6 +104,7 @@ class test_udp_sink_source(gr_unittest.TestCase):
         self.tb_rcv.connect(udp_rcv, dst)
 
         self.tb_rcv.start()
+        time.sleep(0.1)
         self.tb_snd.run()
         udp_snd.disconnect()
         self.timeout = False


noc0lour:
Err: 
```
227/360 Test #227: qa_udp_source_sink .................................***Failed    5.68 sec
.F/data/src/pull/2205/master/gr-blocks/python/blocks/qa_udp_source_sink.py:118: DeprecationWarning: Please use assertTrue instead.
  self.assert_(self.timeout)  # source ignores EOF?
.
FAIL: test_002 (__main__.test_udp_sink_source)
  File "/data/src/pull/2205/master/gr-blocks/python/blocks/qa_udp_source_sink.py", line 85, in test_002
    self.assertEqual(expected_result, result_data)
AssertionError: Tuples differ: (0.0, 1.0, 2.0, 3.0, 4.0, 5.0, 6.0, 7.0, 8[543 chars]99.0) != ()
First tuple contains 100 additional elements.
First extra element 0:
0.0
Diff is 895 characters long. Set self.maxDiff to None to see it.
Ran 3 tests in 2.378s
FAILED (failures=1)
```
lol wat?

noc0lour:
Is it possible the recv thread now sleeps and misses all of the data instead of missing data randomly?

noc0lour:
Makes sense: a wait is needed after setting up the sockets, before starting the flowgraphs... Maybe a wait/block is needed in the constructors?

noc0lour:
Even with:
```
      while (not d_socket->is_open()){
        sleep(10);
      }
```
In the constructor I get 3 fails out of 1000 runs....

hcab14:
Hi @noc0lour,

I had the same problem when I implemented a gnuradio source block for KiwiSDRs (https://github.com/hcab14/gr-kiwisdr) which I first based on the udp_source block.

The problem is the following: `notify_one` assumes that on the other end someone is waiting for the notification. But as far as I know there are no guarantees when the `work` method is called by the scheduler, so it can and does happen that that `notify_one` is being called without `work` waiting on the mutex and in this case the data packet is lost.

noc0lour:
Ok, thanks for the pointer. I'll investigate how to proceed tomorrow

noc0lour:
Hm, running the udp test for 1000 times it still failed 5 times.

noc0lour:
Maybe we should just split the source/sink testing in separate tests and have a udp python socket for recv/send

hcab14:
It is unlikely but UDP packets are not guaranteed to be delivered. Do you have a way to check for dropped UDP packets? Do you also get errors when you run the tests not in parallel?

noc0lour:
tests are carried out serial only. Yes, we don't have that guarantee for UDP packets, but on localhost there shouldn't be 3 out of 1000 dropped packets...

noc0lour:
Let's see if the python only recv/send endpoint will have better results. (Which will be better from a test perspective anyway)

hcab14:
Ok. 

I am now testing with address sanitizer but get a crash triggered by `self.tb_rcv.stop()` in `test_003` which is called when the timeout is reached. I am not sure if this is a serious issue or a false positive.

```
==21465==AddressSanitizer CHECK failed: ../../../../src/libsanitizer/asan/asan_interceptors.cc:313 "((__interception::real___cxa_throw)) != (0)" (0x0, 0x0)
==21465==AddressSanitizer CHECK failed: ../../../../src/libsanitizer/asan/asan_interceptors.cc:313 "((__interception::real___cxa_throw)) != (0)" (0x0, 0x0)
    #0 0x7fe98c9c3122  (/usr/lib/gcc/x86_64-linux-gnu/8/libasan.so+0xf9122)
    #0 0x7fe98c9c3122  (/usr/lib/gcc/x86_64-linux-gnu/8/libasan.so+0xf9122)
    #1 0x7fe98c9df609 in __sanitizer::CheckFailed(char const*, int, char const*, unsigned long long, unsigned long long) (/usr/lib/gcc/x86_64-linux-gnu/8/libasan.so+0x115609)
    #1 0x7fe98c9df609 in __sanitizer::CheckFailed(char const*, int, char const*, unsigned long long, unsigned long long) (/usr/lib/gcc/x86_64-linux-gnu/8/libasan.so+0x115609)
    #2 0x7fe98c9018dc in __interceptor___cxa_throw (/usr/lib/gcc/x86_64-linux-gnu/8/libasan.so+0x378dc)
    #3 0x7fe98567fae5 in boost::this_thread::interruption_point() (/usr/lib/x86_64-linux-gnu/libboost_thread.so.1.65.1+0x11ae5)
    #2 0x7fe98c9018dc in __interceptor___cxa_throw (/usr/lib/gcc/x86_64-linux-gnu/8/libasan.so+0x378dc)
    #4 0x7fe986f09cd2 in boost::condition_variable::do_wait_until(boost::unique_lock<boost::mutex>&, timespec const&) /usr/include/boost/thread/pthread/condition_variable.hpp:124
    #5 0x7fe986f09cd2 in boost::condition_variable::timed_wait(boost::unique_lock<boost::mutex>&, boost::posix_time::ptime const&) /usr/include/boost/thread/pthread/condition_variable_fwd.hpp:142
    #3 0x7fe98567fae5 in boost::this_thread::interruption_point() (/usr/lib/x86_64-linux-gnu/libboost_thread.so.1.65.1+0x11ae5)
    #6 0x7fe986efbb2e in gr::tpb_thread_body::tpb_thread_body(boost::shared_ptr<gr::block>, boost::shared_ptr<boost::barrier>, int) /home/hcab14/gnuradio/gnuradio-runtime/lib/tpb_thread_body.cc:159
    #7 0x7fe986ecabd0 in gr::tpb_container::operator()() /home/hcab14/gnuradio/gnuradio-runtime/lib/scheduler_tpb.cc:47
    #4 0x7fe986f09cd2 in boost::condition_variable::do_wait_until(boost::unique_lock<boost::mutex>&, timespec const&) /usr/include/boost/thread/pthread/condition_variable.hpp:124
    #8 0x7fe986ecabd0 in gr::thread::thread_body_wrapper<gr::tpb_container>::operator()() /home/hcab14/gnuradio/gnuradio-runtime/include/gnuradio/thread/thread_body_wrapper.h:51
    #5 0x7fe986f09cd2 in boost::condition_variable::timed_wait(boost::unique_lock<boost::mutex>&, boost::posix_time::ptime const&) /usr/include/boost/thread/pthread/condition_variable_fwd.hpp:142
    #6 0x7fe986efbb2e in gr::tpb_thread_body::tpb_thread_body(boost::shared_ptr<gr::block>, boost::shared_ptr<boost::barrier>, int) /home/hcab14/gnuradio/gnuradio-runtime/lib/tpb_thread_body.cc:159
    #9 0x7fe986ecabd0 in boost::detail::function::void_function_obj_invoker0<gr::thread::thread_body_wrapper<gr::tpb_container>, void>::invoke(boost::detail::function::function_buffer&) /usr/include/boost/function/function_template.hpp:159
    #10 0x7fe986d78a4b in boost::function0<void>::operator()() const /usr/include/boost/function/function_template.hpp:759
    #7 0x7fe986ecabd0 in gr::tpb_container::operator()() /home/hcab14/gnuradio/gnuradio-runtime/lib/scheduler_tpb.cc:47
    #11 0x7fe986d78a4b in boost::detail::thread_data<boost::function0<void> >::run() /usr/include/boost/thread/detail/thread.hpp:116
    #12 0x7fe98567fbcc  (/usr/lib/x86_64-linux-gnu/libboost_thread.so.1.65.1+0x11bcc)
    #8 0x7fe986ecabd0 in gr::thread::thread_body_wrapper<gr::tpb_container>::operator()() /home/hcab14/gnuradio/gnuradio-runtime/include/gnuradio/thread/thread_body_wrapper.h:51
    #13 0x7fe98c2c16da in start_thread (/lib/x86_64-linux-gnu/libpthread.so.0+0x76da)
    #9 0x7fe986ecabd0 in boost::detail::function::void_function_obj_invoker0<gr::thread::thread_body_wrapper<gr::tpb_container>, void>::invoke(boost::detail::function::function_buffer&) /usr/include/boost/function/function_template.hpp:159
    #14 0x7fe98c5fa88e in __clone (/lib/x86_64-linux-gnu/libc.so.6+0x12188e)
    #10 0x7fe986d78a4b in boost::function0<void>::operator()() const /usr/include/boost/function/function_template.hpp:759

    #11 0x7fe986d78a4b in boost::detail::thread_data<boost::function0<void> >::run() /usr/include/boost/thread/detail/thread.hpp:116
```

noc0lour:
Oh, that doesn't look good. I didn't decipher what exactly happened. But ASAN is a bad sign. 
One thing that could've happened in the current test: the recv socket creates a udp socket on IPv4 and the send socket connects to the IPv6 loopback interface (or vice-versa).

hcab14:
The ASAN messages are not related to the `udp_{source,sink}` blocks and are not there when using clang-6 instead of gcc-7 or gcc-8.

PR #2211, opened 2018-12-04
blocks: fix the udp_source_sink test flakiness

### LINKED ISSUE #2210: QA: qa_udp_source_sink flakiness
On my local machine some I could traceback some errors due to the server socket binding to IPv6 and the client socket sending to the IPv4 endpoint (if using localhost). 

Probably there is more room for improvement if the server socket binds on both IP stacks.
In the meantime the unit tests could just focus on IPv4 succeeding.

This is split from #1261 for better traceability.

noc0lour:
Provide one unit test for the source and sink with the raw python
socket module.
Fix the tests on IPv4 only to avoid mismatched connections on IPv4 &
IPv6.

Fixes #2210.

### DIFF
--- a/gr-blocks/python/blocks/qa_udp_source_sink.py
+++ b/gr-blocks/python/blocks/qa_udp_source_sink.py
@@ -22,10 +22,22 @@
 
 
 from gnuradio import gr, gr_unittest, blocks
+import numpy
 import os
+import socket
 import time
 
-from threading import Timer
+from threading import Timer, Thread
+
+
+def recv_data(sock, result):
+    while True:
+        data = sock.recv(4*1000)
+        if len(data) == 0:
+            break
+        real_data = numpy.frombuffer(data, dtype=numpy.float32)
+        result.extend(list(real_data))
+
 
 class test_udp_sink_source(gr_unittest.TestCase):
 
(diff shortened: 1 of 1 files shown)


gr-blocks/python/blocks/qa_udp_source_sink.py
```diff
@@ -53,37 +65,66 @@ def test_001(self):
...
         n_data = 100
         src_data = [float(x) for x in range(n_data)]
         expected_result = tuple(src_data)
+
+        recvsock = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
+        recvsock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
+        recvsock.bind(('127.0.0.1', port))
+
+        result = []
+        t = Thread(target=recv_data, args=(recvsock, result))
```
mbr0wn: Pfff.... in any other language this would be quite a sin. Thank the GIL :)

gr-blocks/python/blocks/qa_udp_source_sink.py
```diff
@@ -53,37 +65,66 @@ def test_001(self):
...
+        t.join()
+        recvsock.close()
+
+        self.assertEqual(expected_result, tuple(result))
+
+    def test_source_001(self):
+        port = 65520
+
+        n_data = 100
+        src_data = [float(x) for x in range(n_data)]
```
mbr0wn: `numpy.arange` lets you go straight to float, but meh.

mbr0wn:
8 days, no complaints, means merge time.

PR #2213, opened 2018-12-05
udp: enable dualstack IPv6/IPv4 on udp_source

noc0lour:


noc0lour:
Question: Should the block be improved further before merging this?

gr-blocks/lib/udp_source_impl.cc
```diff
@@ -33,6 +33,8 @@
 #include <stdio.h>
 #include <string.h>
 
+#include <iostream>
```
noc0lour: That's unecessary after debugging is completed.

noc0lour:
Maybe I shouldn't remove the sleep after opening the socket.

noc0lour:
This kinda superseeds #2211 since the changes there are partially not required anymore.

noc0lour:
I'm currently looking into https://github.com/ghostop14/gr-grnet.git to pull it in-tree. The dual-stack changes need to be applied there as well, but the codebase is a bit newer.

gr-blocks/lib/udp_source_impl.cc
```diff
@@ -56,22 +56,18 @@ namespace gr {
                       io_signature::make(0, 0, 0),
                       io_signature::make(1, 1, itemsize)),
         d_itemsize(itemsize), d_payload_size(payload_size),
-        d_eof(eof), d_connected(false), d_residual(0), d_sent(0)
+        d_eof(eof), d_connected(false)
     {
       // Give us some more room to play.
-      d_rxbuf = new char[4*d_payload_size];
```
mbr0wn: If the bufs go, so should the comment.

gr-blocks/lib/udp_source_impl.cc
```diff
@@ -56,22 +56,18 @@ namespace gr {
...
     }
 
     udp_source_impl::~udp_source_impl()
     {
       if(d_connected)
-        disconnect();
-
-      delete [] d_rxbuf;
-      delete [] d_residbuf;
+        {
```
mbr0wn: GNU Indents? :nauseated_face:

gr-blocks/lib/udp_source_impl.cc
```diff
@@ -90,16 +86,26 @@ namespace gr {
...
-
-        d_socket->bind(d_endpoint);
 
+        boost::asio::ip::udp::resolver::iterator results(resolver.resolve(query)), end;
+        while (results != end){
+          d_endpoints.emplace_back(*results);
+          results++;
+        }
+        for (const auto& ep: d_endpoints){
+          auto sock = std::unique_ptr<boost::asio::ip::udp::socket>(new boost::asio::ip::udp::socket(d_io_service, ep.protocol()));
```
mbr0wn: /me shakes fist angrily at the C++ gods to forget `make_unique()` in C++11

gr-blocks/lib/udp_source_impl.h
```diff
@@ -38,27 +41,28 @@ namespace gr {
       int     d_payload_size; // maximum transmission unit (packet length)
       bool    d_eof;          // look for an EOF signal
       bool    d_connected;    // are we connected?
-      char   *d_rxbuf;        // get UDP buffer items
-      char   *d_residbuf;     // hold buffer between calls
-      ssize_t d_residual;     // hold information about number of bytes stored in residbuf
-      ssize_t d_sent;         // track how much of d_residbuf we've outputted
+      std::vector<std::unique_ptr<char[]>> d_rxbufs;
```
mbr0wn: Much, much better.

gr-blocks/lib/udp_source_impl.h
```diff
@@ -38,27 +41,28 @@ namespace gr {
...
 
       static const int BUF_SIZE_PAYLOADS; //!< The d_residbuf size in multiples of d_payload_size
 
       std::string d_host;
       unsigned short d_port;
 
-      boost::asio::ip::udp::socket *d_socket;
-      boost::asio::ip::udp::endpoint d_endpoint;
-      boost::asio::ip::udp::endpoint d_endpoint_rcvd;
+      std::vector<std::unique_ptr<boost::asio::ip::udp::socket> > d_sockets;
```
mbr0wn: Just a minor nit, but `> >` is no more necessary since C++11.

noc0lour:
@marcusmueller I do not want to bother you, but we have two new segfaulting tests on F29. tag_share & qa_skiphead. Segfaulting consistently on py2&py3. All other distros not affected.

noc0lour:
Maybe be due to rebuilt docker containers.

gr-blocks/lib/udp_source_impl.h
noc0lour: Oh, I'll remove it. I'm just used to pre-C++11 (cautions) coding.

gr-blocks/lib/udp_source_impl.h
noc0lour: Still a poor-man's buffer (:

gr-blocks/lib/udp_source_impl.cc
noc0lour: I don't know what happened there. I'll check again tomorrow.

gr-blocks/lib/udp_source_impl.cc
noc0lour: @es.

noc0lour:
@mbr0wn Thanks for the first round of review. I'll try to incorporate the stuff from the gr-net (maybe in gr-network) and add the appropriate changes there as well.

noc0lour:
@marcusmueller That's how you can open a socket on IPv4 & IPv6 without limiting yourself to a special usecase. The sockets will interleave their contents. But that's also true if you send data from multiple clients into the udp source.  (So you shouldn't do that anyway).  The usecase is: Create a open socket on `myhostname.mynet`. Which will open on the resolved IP adresses (regardless of number & IP stack). Also from working with other software: An empty value for host intuitively should mean: Open socket on any IP (0.0.0.0, ::) which currently isn't possible for both IPv4 & IPv6 as well.

The `localhost` test case covers both (IPv4 & IPv6), since your system picks one of those anyway. But I can make a test for both explicitly.

gr-blocks/lib/udp_source_impl.cc
```diff
@@ -90,16 +86,26 @@ namespace gr {
...
-
-        d_socket = new boost::asio::ip::udp::socket(d_io_service);
-        d_socket->open(d_endpoint.protocol());
-
-        boost::asio::socket_base::reuse_address roption(true);
-        d_socket->set_option(roption);
-
-        d_socket->bind(d_endpoint);
 
+        boost::asio::ip::udp::resolver::iterator results(resolver.resolve(query)), end;
```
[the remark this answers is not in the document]
noc0lour: It's possible this won't work due to overloads. But I haven't tested that. (In later boost versions this get's 10dB nicer).

devnulling:
@noc0lour I was talking with @ghostop14 yesterday about this issue, and he just pushed IPv6 support for gr-grnet https://github.com/ghostop14/gr-grnet/commits/master

Since we have an open grep that will soon bring in gr-grnet to the main tree https://github.com/gnuradio/greps/pull/22 , should we close this issue out?

ghostop14:
Hi folks, I just incorporated IPv6 support in gr-grnet.  TCP and UDP source/sinks.  I have another item around graceful shutdown for the TCP sink in server mode that I'm going to work on over the weekend but IPv6 support just got pushed if you need something now.

noc0lour:
I'll give gr-grnet a spin and see if it has same functionality w.r.t IPv6. Please, only close this PR once you have a PR replacing the blocks with gr-grnet.

noc0lour:
Quick check yields I might submit a PR for your UDP & TCP blocks to add tricks from this PR in there so we can have real dual-stack support there as well. This is very convenient for "regular" users as you don't have to care about if it is ipv4 or ipv6 you are using e.g. for listening to 0.0.0.0/0 or ::/0 or localhost.

mormj:
@noc0lour - can this be closed now that gr-grnet has moved in tree as gr-network and the networking blocks in blocks have been deprecated?

noc0lour:
I probably want to port patches from here to the new blocks because they also don't support dualstack iirc.

mbr0wn:
@noc0lour I'm still going to close this just so we have a better clue of what's going on w/ GNU Radio PRs. I still hope to see some patches move to gr-network. Thanks for your networking experience, here and elsewhere.

noc0lour:
:+1:

PR #2218, opened 2018-12-11
dtv: Enable DTV component.

drmpeg:


drmpeg:
Some YAML notes.

1) You can't use hide_something: ['', all]. It has to be hide_something: [none, all].

2) You can't use ${variable1 + variable2}. It has to be ${variable1} + ${variable2}.

mbr0wn:
Manually merged. Split the commit and fixed the commit messages.

PR #2226, opened 2018-12-14
cmake: Bump minimum version of CMake to 3.5.1

mbr0wn:
This will work on Ubuntu 16.04.


noc0lour:
I have a follow up PR which will set in effect all of the currently disabled cmake policies.

PR #2227, opened 2018-12-15
uhd: Fix GRC bindings (gain, freq...)

### LINKED ISSUE #2221: UHD: uhd sink not working

mbr0wn:
The YML file generated during CMake was simply not correct. Frequency,
gain, etc. weren't propagated properly to the underlying C++ object.

Fixes #2221.

### DIFF
--- a/gr-uhd/grc/CMakeLists.txt
+++ b/gr-uhd/grc/CMakeLists.txt
@@ -1,4 +1,4 @@
-# Copyright 2011 Free Software Foundation, Inc.
+# Copyright 2011,2018 Free Software Foundation, Inc.
 #
 # This file is part of GNU Radio
 #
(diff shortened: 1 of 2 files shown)


drmpeg:
I was able to transmit RF at the correct frequency, so big improvement. But still some issues.

1) Subdev Spec still doesn't work. Can't select other channel with A:B.

2) Wire Format still doesn't work. Can't select anything except Automatic.

3) Sync parameter doesn't "stick". Always returns to "Unknown PPS" when reloading flow graph.

4) Bandwidth value of 0 doesn't work as before. Error message:

WARNING] [AD936X] The requested bandwidth 0.000000 MHz is out of range (0.200000 - 56.000000 MHz).
The bandwidth has been forced to 0.200000 MHz.

mbr0wn:
Thanks for the feedback. I'll merge this -- it fixes a bunch of stuff -- and continue on the rest in a separate PR.

However, some of these things might be UHD issues, or no issues at all. For example, the bandwidth of 0 will always get coerced to a different value, if the device supports it (a 0 Hz bandwidth is not supported).

PR #2231, opened 2018-12-16
uhd: More fixes to GRC bindings (time/clock source, etc.)

### LINKED ISSUE #2221: UHD: uhd sink not working

mbr0wn:
- Subdev spec wasn't being applied at all
- time/clock source, neither
- Sync wouldn't stay the same value when it was set to "don't sync"
- Bandwidth is no longer set when the value is 0, allowing for a default
  behaviour
- OTW format wasn't being applied properly

Fixes #2221.


mbr0wn:
OK. I'll keep adding fixes, and merge this to enable people's productivity.

PR #2234, opened 2018-12-16
uhd: Check for mako at CMake time

### LINKED ISSUE #2154: CMake enables gr-uhd but compilation fails due to missing Python module mako
On a system without `python-mako` CMake enables `gr-uhd`, however compilation fails due to missing Python module `mako`. I can reproduce this error on Ubuntu 18.04 LTS and Arch Linux.

mbr0wn:
Building gr-uhd requires mako. GRC only requires it at runtime, so
there's check for this anywhere in the CMake build system.

Fixes #2154, but it's not a great fix.

### DIFF
--- a/gr-uhd/CMakeLists.txt
+++ b/gr-uhd/CMakeLists.txt
@@ -27,6 +27,11 @@ find_package(UHD "3.5.5")
 ########################################################################
 # Register component
 ########################################################################
+GR_PYTHON_CHECK_MODULE_RAW(
+    "mako >= 0.9.1"
+    "import mako; assert mako.__version__ >= '0.9.1'"
+    MAKO_FOUND
+)
 include(GrComponent)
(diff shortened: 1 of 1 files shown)


marcusmueller:
wait, this means gr-uhd won't build on machines that don't have mako, since that would be required to build the GRC bindings, right?

Isn't that kinda bad for Ettus customers with headless, embedded boxes, who potentially want to use gr-uhd on their target?

noc0lour:
I wouldn't say mako isn't that bad of a dependency (since it takes about 1.2MBytes of space only)' And we require python for building anyway ;)

noc0lour:
On a side-note: We should have all the grc subdirs dependent on GRC. Since you don't need the grc bindings if you are not going to use GRc?

mbr0wn:
I surely hope no one's building UHD on little embedded boxes, but all or our embedded boxes (aka USRPs) have Mako on them.

PR #2241, opened 2018-12-16
uhd: Fix even more GRC binding issues

### LINKED ISSUE #2221: UHD: uhd sink not working

mbr0wn:
- The callback for gain would always use normalized gain, which would
  cause errors since the non-normalized gain range would cause
  exceptions
- The "Show Command Port" menu was removed, it didn't work, but no other
  block has this.

Finally fixes #2221.


drmpeg:
Looks good. Run time setting tested with both absolute and normalized.

Issue #2242, opened 2018-12-16
qa: moving_average flakiness


mbr0wn:
```
205/360 Test #205: qa_moving_average ..................................***Failed    0.69 sec
...F
======================================================================
FAIL: test_04 (__main__.test_moving_average)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/cache/workdir/Ubuntu_16_04_64_remote_0/build_Ubuntu_16_04_64/src/gr-blocks/python/blocks/qa_moving_average.py", line 155, in test_04
    self.assertComplexTuplesAlmostEqual(expected_result, dst_data, 4)
  File "/cache/workdir/Ubuntu_16_04_64_remote_0/build_Ubuntu_16_04_64/src/gnuradio-runtime/python/gnuradio/gr_unittest.py", line 86, in assertComplexTuplesAlmostEqual
    self.assertComplexAlmostEqual (a[i], b[i], places, msg)
  File "/cache/workdir/Ubuntu_16_04_64_remote_0/build_Ubuntu_16_04_64/src/gnuradio-runtime/python/gnuradio/gr_unittest.py", line 54, in assertComplexAlmostEqual
    msg or '%r != %r within %r places' % (first, second, places))
AssertionError: (-11.104617556156482-11.9834404148583j) != (-11.104567527770996-11.983440399169922j) within 4 places
----------------------------------------------------------------------
Ran 4 tests in 0.170s
FAILED (failures=1)
```

Build: https://ci.gnuradio.org/buildbot/#/builders/49/builds/405/steps/7/logs/stdio

mbr0wn:
I have not seen this fail in a long time -- closing.

PR #2243, opened 2018-12-16
blocks: Give qa_moving_average.py a fixed random seed

mbr0wn:


mbr0wn:
https://github.com/gnuradio/gnuradio/issues/2242

mbr0wn:
Not sure if this is 'it'.

marcusmueller:
@mbr0wn It is, I think. As far as I can see, there's no chance of two concurrent Python threads accessing the random pool, and the seed is reset after every run

Even if this wasn't it, we should be having this. Meeeeerge.

PR #1955, opened 2018-08-17
Pull request for issue #1952 (adding bandwidth message port to qtgui freq / waterfall sink)

jjhein:
Added support to qtgui frequency and waterfall sink for a bandwidth message port so a user can change the display bandwidth programatically from another block.

PR #2244, opened 2018-12-17
Merge #1955 (jjhein's QTGui bandwidth message port)

### LINKED ISSUE #1952: qtgui freq-sink and waterfall-sink need a bandwidth message port
Per an email thread to the development alias, we need an additional message port for the frequency and waterfall sink blocks within the qtgui library.

marcusmueller:
This is the rebased version of #1955.

This fixes #1952.

### DIFF
--- a/gr-qtgui/lib/freq_sink_c_impl.cc
+++ b/gr-qtgui/lib/freq_sink_c_impl.cc
@@ -64,6 +64,7 @@ namespace gr {
   d_center_freq(fc), d_bandwidth(bw), d_name(name),
   d_nconnections(nconnections),
   d_port(pmt::mp("freq")),
+  d_port_bw(pmt::mp("bw")),
   d_parent(parent)
     {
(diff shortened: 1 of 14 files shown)


gr-qtgui/lib/freq_sink_c_impl.cc
```diff
@@ -74,6 +75,11 @@ namespace gr {
       d_argv = new char;
       d_argv[0] = '\0';
 
+      // setup bw input port
+      message_port_register_in(d_port_bw);
+      set_msg_handler(d_port_bw,
+                      boost::bind(&freq_sink_c_impl::handle_set_bw, this, _1));
```
mbr0wn: Ideally, we'd not use `boost::bind` (see also our GREP on coding standards). But it's OK for now.

marcusmueller: Agreed, but a rewrite didn't seem desirable at this point.

PR #2246, opened 2018-12-18
dtv: Fix optimized BCH encoder for N=8 codes.

drmpeg:


marcusmueller:
Hey @drmpeg, got any background on this? I see you've reimplemented Julius' sped up implementation, probably because it was wrong. But the title change suggests this was part of a bigger effort?

drmpeg:
The problem has to do with using to_ulong(). The variable "parity_bits" is always 192 bits long. For BCH codes with N=8, the number of parity bits is 128. parity_bits >> (num_parity_bits - 8) ends up being 72 bits, which doesn't fit into a ulong, and the you get an overflow_error exception.

For 32-bit systems, N=10 and M=12 codes also have this problem.

I tried using boost::dynamic_bitset, but that ended up being a nightmare.

Issue #2247, opened 2018-12-18
alloc-dealloc mismatch in tcp_server_sink_impl


dlrobertson:
# Summary

A mismatching smart pointer type in `tcp_server_sink_impl` is causing a alloc-dealloc
mismatch.

# Details

The `tcp_server_sink_impl` structure contains the `shared_ptr` `d_buf`. It is initialized
with the following:

```c++
boost::shared_ptr<uint8_t> d_buf;
```

Which results in the following:

```
==9826==ERROR: AddressSanitizer: alloc-dealloc-mismatch (operator new [] vs operator delete) on 0x7fc6481da800
    #0 0x7fc656af35e0 in operator delete(void*) /var/tmp/portage/sys-devel/gcc-8.2.0-r5/work/gcc-8.2.0/libsanitizer/asan/asan_new_delete.cc:135
    #1 0x7fc64d752534 in boost::detail::sp_counted_base::release() /usr/include/boost/smart_ptr/detail/sp_counted_base_std_atomic.hpp:110
    #2 0x7fc64da3f723 in boost::detail::shared_count::~shared_count() /usr/include/boost/smart_ptr/detail/shared_count.hpp:426
    #3 0x7fc64da3f723 in boost::shared_ptr<unsigned char>::~shared_ptr() /usr/include/boost/smart_ptr/shared_ptr.hpp:341
    #4 0x7fc64da3f723 in gr::blocks::tcp_server_sink_impl::~tcp_server_sink_impl() ../gr-blocks/lib/tcp_server_sink_impl.cc:113
    #5 0x7fc64da3fa90 in gr::blocks::tcp_server_sink_impl::~tcp_server_sink_impl() ../gr-blocks/lib/tcp_server_sink_impl.cc:129
    #6 0x7fc651a62b85 in boost::detail::sp_counted_base::release() /usr/include/boost/smart_ptr/detail/sp_counted_base_std_atomic.hpp:110
    #7 0x7fc651a62b85 in boost::detail::shared_count::~shared_count() /usr/include/boost/smart_ptr/detail/shared_count.hpp:426
    #8 0x7fc651a62b85 in boost::shared_ptr<gr::basic_block>::~shared_ptr() /usr/include/boost/smart_ptr/shared_ptr.hpp:341
    #9 0x7fc651a62b85 in gr::endpoint::~endpoint() ../gnuradio-runtime/include/gnuradio/flowgraph.h:37
[...]

PR #2248, opened 2018-12-18
Fix alloc-dealloc-mismatches

dlrobertson:
The `d_rotator` member of `pfb_decimator_ccf_impl` is allocated with `new[]`
but deallocated with `delete`. It should instead be deallocated with
`delete[]`.

The `d_buf` member of `tcp_server_sink_impl` is currently typed as a shared
pointer to a `uint8_t` but should be a shared pointer to a `uint8_t[]`.

Fixes: #2247 
Fixes: #2249


mbr0wn:
Merging manually.

PR #2250, opened 2018-12-19
Remove XML runner

mbr0wn:
What the hell is this even still doing here.


mbr0wn:
I thought this would speed up QA significantly, too, but it only went down from 230 -> 200 seconds on my machine. Very non-scientific test, too.

noc0lour:
I think it's not really used? But I might be mistaken.

PR #2251, opened 2018-12-19
qa: Use random.seed(0) everywhere import random is used

mbr0wn:
This affects all Python-based unit tests which use the Python random
module. If they do, this change adds random.seed(0) to every setUp()
call, so that all QA runs are reproducible.

### DIFF
--- a/gr-blocks/python/blocks/qa_keep_m_in_n.py
+++ b/gr-blocks/python/blocks/qa_keep_m_in_n.py
@@ -29,7 +29,7 @@ import random
 class test_keep_m_in_n(gr_unittest.TestCase):
 
     def setUp(self):
-        pass
+        random.seed(0)
 
     def tearDown(self):
         pass
--- a/gr-blocks/python/blocks/qa_pack_k_bits.py
+++ b/gr-blocks/python/blocks/qa_pack_k_bits.py
@@ -28,6 +28,7 @@ from gnuradio import gr, gr_unittest, blocks
 class test_pack(gr_unittest.TestCase):
 
     def setUp(self):
+        random.seed(0)
         self.tb = gr.top_block()
 
     def tearDown(self):
--- a/gr-blocks/python/blocks/qa_packed_to_unpacked.py
+++ b/gr-blocks/python/blocks/qa_packed_to_unpacked.py
@@ -28,6 +28,7 @@ import random
 class test_packing(gr_unittest.TestCase):
 
     def setUp(self):
+        random.seed(0)
         self.tb = gr.top_block ()
 
     def tearDown(self):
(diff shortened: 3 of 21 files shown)


noc0lour:
Out of curiosity: It's not really possible to hack this into `gr_unittest.TestCase` ?

michaelld:
I'm good with this change as-is as a temporary fix, but I'm with @noc0lour in wondering if this can't be taking up a level to be a common setting for all Python-based QA. I've done this "up a level" in some of my OOTs & it works fine, and the code is much more concise. Not sure if/how well it might work for GR-proper.

noc0lour:
Can't we set `random.seed(0)` in the "generic" setup method of gr_unittest and require derived classes to call `super().setUp` ?

marcusmueller:
I'm going to agree with @michaelld and @noc0lour, doing this generally in the framework is the preferable solution.

Let's however rather do this now, and not forget to do the other later. Gut feeling: We might be running into problems if we end up doing seed(0) in every setup without telling people. Someone will write consecutive test cases that don't expect the RNG state to be reset between single test cases.

@michaelld could you open an issue, maybe even with a piece of code (maybe even in a PR) to do it like you do in your OOTs? Better to have a good starting point than to discuss "we could do it like this or like that or that or ".

marcusmueller:
@mbr0wn @noc0lour are there issues that get closed by this?

noc0lour:
@marcusmueller could affect the flakiness of soft_decoder & socket_pdu. (But latter rather unlikely due relation with "wrong" message passing.)

michaelld:
@marcusmuller Here's what I do: Create an "intermediate" Python file in the OOT that imports from GR, and then change all of the OOT's QA to import the local intermediate instead of from GR & use my local intermediate class for inheritance rather than that provided by GR. I've never looked into what it might take to change GR's implementation ... so, while I -can- create an issue it'll take me a bit to work up some demo code that changes GR's implementation to do what we want. Maybe someone else has done this before & hence is better suited to do this than me?

mbr0wn:
@noc0lour @marcusmueller I tried putting it into gr_unittest.py, but it wouldn't "stick". I'm not sure why -- it's something to do with the module being in a different context maybe, although I was still surprised. Basically, in gr_unittest.py I'd import random, and run `random.seed(0)`, but leave the other QA files unchanged. Maybe the seed gets reset on import, then the `import random` in the QA files would overwrite my seeding. I didn't research the reasons though.

There's another reason to not do this: We don't know how exactly people are generating randomness. They could use numpy random, or other RNGs. By forcing the QA files to be explicit, we remind people all the time that this is how it goes.

@marcusmueller What @noc0lour said. I can't prove that it fixes any issues, but I have a hunch it'll fix the precision assertion errors we've been seeing recently.

marcusmueller:
@michaelld that's fine! I think @mbr0wn's arguments weigh heavily, and you truly aren't the only one who could do that.

Issue #2229, opened 2018-12-15
digital: OFDM Cyclic Prefixer block Length Tag Key value doesn't "stick"


drmpeg:
For DTV OFDM based flow graphs, the Length Tag Key is set to blank. When the flow graph is reloaded, the Length Tag Key always reverts to the default "frame_len".

drmpeg:
This can be solved by using empty quotes ("") instead of blank.

noc0lour:
Hm, so doesn't need a fix? Or should it do the same thing for blank as well?

drmpeg:
I guess blank would be a nice feature. I'm regenerating all of the DTV example flow graphs, so I can live with empty quotes (which is why I closed the issue).

noc0lour:
Ok, I see. I'm not too fit in the current yaml format, maybe @haakov might be able to tell if it would be possible to implement such feature (or at least think about it and open an issue for that).

haakov:
Interesting! Actually, GRC discards all blank parameters upon creating the `.grc` file. I've created a PR for it: https://github.com/gnuradio/gnuradio/pull/2254

PR #2254, opened 2018-12-19
grc: Make GRC save blank parameters

haakov:
Following the discussion from https://github.com/gnuradio/gnuradio/issues/2229. 

For the time being, GRC discards empty parameters when generating the `.grc` file. This PR alters this behaviour. I haven't tested if this creates any problems yet, but can't think of any off the top of my head. 

It will result in more cluttered `.grc` files though. @skoslowski what are your thoughts on this?

### DIFF
--- a/grc/core/platform.py
+++ b/grc/core/platform.py
@@ -348,9 +348,6 @@ class Platform(Element):
         try:
             for d in chain([data['options']], data['blocks']):
                 d['states']['coordinate'] = yaml.ListFlowing(d['states']['coordinate'])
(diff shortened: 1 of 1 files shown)


marcusmueller:
I feel this is a very positive change; this should probably some one or two sporadic issues in the variable editor.

drmpeg:
I tested this with the OFDM Cyclic Prefixer, and it works good.

mbr0wn:
No complaints, only positive feedback, and I'm personally not too worried about cluttering auto-generated files. Better in than out. Merging.

PR #2256, opened 2018-12-19
C++11: Convert boost::lexical_cast to std::to_string

marcusmueller:
This is a minor change towards slowly replacing Boost with newer std::
functionality.


noc0lour:
:champagne:

PR #2257, opened 2018-12-19
blocks: make file_source take (u)int64_t for lengths, offsets, not size_t

### LINKED ISSUE #2245: File Source block doesn't work with files over 4GB
Using size_t for the new offset and length features is no bueno.

marcusmueller:
This is meant to address #2245; @drmpeg rightfully complained that it's a bad
idea to use size_t as file length or offset specifier, since that's very
limited in usefulness on 32 bit systems. 

This requires testing by @drmpeg before merging. 

Closes #2245. (presumably. Don't merge if it doesn't allow large offsets)

### DIFF
--- a/gr-blocks/include/gnuradio/blocks/file_source.h
+++ b/gr-blocks/include/gnuradio/blocks/file_source.h
@@ -62,7 +62,7 @@ namespace gr {
        * \param len     produce only items [offset, offset+len)
        */
       static sptr make(size_t itemsize, const char *filename, bool repeat = false,
-                       size_t offset = 0, size_t len = 0);
+                       uint64_t offset = 0, uint64_t len = 0);
 
       /*!
        * \brief seek file to \p seek_point relative to \p whence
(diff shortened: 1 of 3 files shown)


mbr0wn:
Yikes, this'll require some careful testing.

marcusmueller:
As @mbr0wn says, this will require **extensive** testing, and it makes no sense at all to merge this without having tested it thoroughly on both 64 bit and 32 bit.

drmpeg:
Just FYI, this is not a 32-bit system problem. size_t is 32 bits even on 64-bit systems.

However, I will test the fix on both 64 and 32-bit systems.

drmpeg:
Shouldn't offset and length also be 64-bits in the .yml file?

drmpeg:
So far so good with a 64-bit system. However, when testing with a large file (9 GB), there's a fairly long delay (10 seconds) before the flow graph starts up. Disk light is on, so it's apparently accessing the entire file.

drmpeg:
I have a new scheme that works perfectly on a 32-bit system. You can preview it here:

https://github.com/drmpeg/gnuradio/tree/filesource-fix

Files are:

gr-blocks/include/gnuradio/blocks/file_source.h
gr-blocks/lib/CMakeLists.txt
gr-blocks/lib/file_source_impl.cc
gr-blocks/lib/file_source_impl.h

It uses fseeko and ftello along with the compiler flag FILE_OFFSET_BITS=64.

I can submit the pull request after this one is closed.

marcusmueller:
Closing this PR in favor of @drmpeg's branch – which I haven't looked through yet, but if this doesn't fix the issue on 32 bit, then it's not worth the merge, anyway.

PR #2258, opened 2018-12-19
GRC: enable displaying of variable IDs

### LINKED ISSUE #2235: Block ID's missing
Block ID's for Variable blocks are missing from 3.8. Seems like a useful feature to me.

New:

image

Old:

image

marcusmueller:
This was a regression from GRC 3.7: Variable IDs were no longer
displayed.

This restores that functionality for any block with the flag "variable".

This tags the Variable, Variable Config, Variable Struct, Parameter, and
Function Probe blocks with that particular flag.

Closes #2235.

### DIFF
--- a/gr-blocks/grc/blocks_tag_object.block.yml
+++ b/gr-blocks/grc/blocks_tag_object.block.yml
@@ -1,5 +1,6 @@
 id: variable_tag_object
 label: Tag Object
+flags: variable
 
 parameters:
 -   id: offset
--- a/gr-digital/grc/digital_constellation.block.yml
+++ b/gr-digital/grc/digital_constellation.block.yml
@@ -1,5 +1,6 @@
 id: variable_constellation
 label: Constellation Object
+flags: variable
 category: Modulators
 
 parameters:
--- a/gr-digital/grc/digital_constellation_rect.block.yml
+++ b/gr-digital/grc/digital_constellation_rect.block.yml
@@ -1,5 +1,6 @@
 id: variable_constellation_rect
 label: Constellation Rect. Object
+flags: variable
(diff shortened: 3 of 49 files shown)


marcusmueller:
Does that look OK, @drmpeg?

haakov:
There are some variable blocks in gr-qtgui, should probably display their IDs too :)

marcusmueller:
@haakov done

haakov:
Wait, doesn't this break compatibility with older .grc files?  Removing the ID prefix, I mean?

marcusmueller:
@haakov good point. Wish we had a "block alias" field that I could use to do this without breaking stuff, but you're right, we shouldn't be unnecessarily breaking .grc files. Let me rework that.

mbr0wn:
@marcusmueller Is this still WIP?

marcusmueller:
@mbr0wn oh! I have a version on my laptop that I simply forgot to push here.

nickoe:
@marcusmueller Any update?

noc0lour:
@marcusmueller push it (:

noc0lour:
Apparently the commit is still missing.

bastibl:
I think it's a good idea to introduce a flag to show the id of the block, but I'm not sure if we should tie it to variables. My understanding is that in some parts of the code a variable is something that has a `variable` flag (which actually just means that the id is shown). In other parts of the code a variable is something that has a `value` (and the semantics between the two are not differentiated). How about a `show_id` flag to show the id?

drmpeg:
This PR can be closed. Superseded by https://github.com/gnuradio/gnuradio/pull/2568

PR #2263, opened 2018-12-25
blocks: Fix file_source offset/length to properly access large files.

drmpeg:


drmpeg:
Thanx for pointing this out. It's because I removed the fd = ::open() since it's not required anymore.

gr-blocks/lib/file_source_impl.cc
```diff
@@ -105,45 +89,37 @@ namespace gr {
...
-                           size_t start_offset_items, size_t length_items)
+                           uint64_t start_offset_items, uint64_t length_items)
     {
       // obtain exclusive access for duration of this function
       gr::thread::scoped_lock lock(fp_mutex);
 
-      int fd;
-
-      // we use "open" to use to the O_LARGEFILE flag
-      if((fd = ::open(filename, O_RDONLY | OUR_O_LARGEFILE | OUR_O_BINARY)) < 0) {
```
noc0lour: Why did this go and `fopen` is used below instead of `fdopen`? Probably it needs some smarter checking for non-existence.

noc0lour: Well you actually pointed that out in the same moment.

drmpeg: The reason ::open was being used was to open with O_LARGEFILE. Now that FILE_OFFSET_BITS=64 is being used, the d_new_fp = fopen() is all that's required.

This fix is to just remove the fclose(d_new_fp).

noc0lour: Ah, right the throw is just after the fclose()

drmpeg:
Still failing the QA. Investigating...

drmpeg:
All QA tests passing here on my local builds (both x86 and arm). Did the build machine pick up the latest push?

noc0lour:
Looks like it did. I'll try to reproduce locally

gr-blocks/lib/CMakeLists.txt
```diff
@@ -199,6 +199,8 @@ list(APPEND gr_blocks_sources
     wavfile_source_impl.cc
 )
 
+set_source_files_properties(file_source_impl.cc PROPERTIES COMPILE_FLAGS -D_FILE_OFFSET_BITS=64)
```
noc0lour: I'm wondering if a `#define` in the implementation would be better?

drmpeg: No. It doesn't work on 32-bit architectures.

drmpeg: It works, but you have to put it before the #includes.

noc0lour: Maybe that's better since it will keep the `#define` and the corresponding code together instead of pulling it out into the build system. 
As long as it works I'm personally fine with both.

marcusmueller:
guys, I'm sorry to chime in this late, but:

we can't do this. `fseeko` is POSIX.1-2001. This means it's never going to be available on Windows. So, this code simply won't compile.

noc0lour:
there are `_fseeki64` and `_ftelli64`, so need to distinguish between MSVC and other compilers.

drmpeg:
I think _fseeki64 and _ftelli64 would work fine. Unfortunately, I don't have any way to build or test this. My most recent Windows compiler is MSVC 2003. Doh!

noc0lour:
I'm personally fine with the changes now. I pinged Travis in the GNU Radio slack if he is able to test that in a windows environment.

gnieboer:
Sure. I'll take a look.

drmpeg:
Cool. Let me know if you need any help. It's a pretty old branch, so I can provide a re-based branch if you like.

gnieboer:
No worries on the rebase, I did a quick patch onto a local branch.

First, duplicated the issue using a 9GB audio file (with different tone after ~4GB), where once the offset was >2GB, there was an fread error on the old code.

Applied patch, built fine, and had no error, and also confirmed the correct tone was playing at the correct offset.

BTW: Wouldn't bit be a bit cleaner to #define GR_FSEEKO at the top and avoid the multiple #defines in the code? Especially since the declarations are the same.

drmpeg:
I like it. I like it a lot. Squashed commit with #define changes pushed.

noc0lour:
Cool!

noc0lour:
I think this is all fine, looking through the changes again I'm wondering: is there a reason why the len/offset_type is `int64_t` and not `uint64_t` ? I'd guess it's better compatible with the underlying API?

drmpeg:
It's because the offset can be negative and it's being tested for in line 105 of gr-blocks/python/blocks/qa_file_source.py.

noc0lour:
Now that this works on Windows, @marcusmueller can you please merge, since you had objections before.

drmpeg:
Looking at the code again, it's the seek method that can have a negative value, not the offset. I'm going to give this one more change back to uint64_t.

Issue #2262, opened 2018-12-25
The GR DTV blocks SSE2 tests succeeds also when building on ARM


ast:
Tested on Raspberry Pi with clang 3.8.

ast:
Should it be flagged as a bug?

balister:
So the problem is the sse2 test succeeds when run with an arm compiler?

ast:
Same things happen in volk but then they filter it out because arch is not x86. I think default clang is multiarch and compiles it just fine... Not sure, have to test more.

balister:
Sounds like a clang bug :) Or you are running clang that is producing x86 output. Define CC to have the right arch flags. Something like this:
export CC="arm-oe-linux-gnueabi-gcc  -march=armv7-a -mfloat-abi=hard -mfpu=neon --sysroot=$SDKTARGETSYSROOT"
export CXX="arm-oe-linux-gnueabi-g++  -march=armv7-a -mfloat-abi=hard -mfpu=neon --sysroot=$SDKTARGETSYSROOT"
export CPP="arm-oe-linux-gnueabi-gcc -E  -march=armv7-a -mfloat-abi=hard -mfpu=neon --sysroot=$SDKTARGETSYSROOT"

drmpeg:
@balister Passing the correct flags doesn't help. Already tried that.

drmpeg:
BTW, I'm using clang-6.0 on arm.

balister:
This looks bad:
https://github.com/gnuradio/gnuradio/blob/master/gr-dtv/lib/CMakeLists.txt#L120
I'd replace that with a try compile to see if the compiler can build sse2 code. Using flags to check for the presence of simd is not great, use try_compile blocks.

drmpeg:
@balister Agreed, although that's been in there since 10/2015. I guess there's not too many clang users on arm.

drmpeg:
Pull request submitted. See https://github.com/gnuradio/gnuradio/pull/2266

ast:
There is a very similar bug in libvolk I submitted, not sure if you are also involved in that?
https://github.com/gnuradio/volk/issues/221

Thanks for the quick fix!

PR #2266, opened 2018-12-26
dtv: Make SSE2 check clang compatible.

drmpeg:


drmpeg:
It's a problem for ARM architecture. Read the referenced issue. https://github.com/gnuradio/gnuradio/issues/2262

michaelld:
My preference would be to check for the flag first & if found then try the actual compile. My reasoning is that some Clang will pass the former but not the latter because the compiler has flaws in how it does pre-processing; this issue is mostly with older clang, but I like the general two-pronged approach.

This said, I'm not sure if this approach would solve whatever the actual issue is. Hopefully @drmpeg will comment more specifically.

drmpeg:
Just to reiterate, the issue manifests itself on ARM architecture. If you pass the compiler flag -msse2 to clang on ARM, it doesn't give an error, it just gives a warning. SSE2_SUPPORTED gets set to true, and the build fails.

michaelld:
!@#$ Clang ... just has some issues here and there with compiler flags. Hence trying both the flag and compile.

mbr0wn:
Yeah I see your point (and I didn't realize this was specific to ARM). For me (and maybe also for @michaelld) there's just a hesitation here because `check_c_compiler_flag()` is what you're "supposed" to do, vs. check_compile which is considered hacky.

However, none of that matters if the former solution (aka the "clean solution") doesn't work. Non-working solutions are never better than working ones.
We should, however, add a comment in the CMake which explains why we're deviating from what is considered better CMake code. @michaelld, would that work for you?

Also, we should ask @balister if there's any issues with this solution.

mbr0wn:
@balister approves from the coaching box.

mbr0wn:
Merging manually.

mbr0wn:
I squashed the commits and added some verbage. 83c0b7c0651a1959ac118c67154fce50541755df

PR #2272, opened 2018-12-29
fixup! cmake: Only go into grc/ subdirs when ENABLE_GRC=ON

### LINKED ISSUE #2259: Commit 7daef13 completely breaks GRC
https://github.com/gnuradio/gnuradio/commit/7daef13332e7ae8fbe7bf90d1a80240c315c7e85 completely and utterly breaks GNU Radio Companion. Almost all blocks are missing.

image

mbr0wn:
The grc subdir was parsed last, making ENABLE_GRC appear false in all
components, thus disabling blocks.

Fixes #2259.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -474,6 +474,7 @@ add_definitions(${MPLIB_DEFINITIONS})
 ########################################################################
 add_subdirectory(docs)
 add_subdirectory(gnuradio-runtime)
+add_subdirectory(grc)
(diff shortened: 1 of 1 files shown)


mbr0wn:
I like it, since it says "there was a commit called XYZ and it's broken, here's the fix". But I can change it, I'm not married to this. I can't just remove the 'fixup', though, I'd rename the entire commit.

PR #2273, opened 2018-12-30
GRC: Fix bugs after merging C++ generation

noc0lour:


drmpeg:
Working good now.

haakov:
Ouch, I must've missed the changes to `top_block.py` when looking through the PR. Sorry about this!

I believe all the changes to that file can (and should) be reverted.

marcusmueller:
@haakov honestly, so much awesomeness in there, such things happen. Anyway, if we should be reverting all changes, that kind of superseedes @noc0lour's proposed changes. @noc0lour ?

noc0lour:
Hm, that can't be entirely true, since you need changes in there to generate something based on the `cpp_template` ?

noc0lour:
I'll first merge this so GRC on master will work again for generating flowgraphs and if there are changes wrongly included we can fix them afterwards, ok?

haakov:
Sure! That works :)

PR #2275, opened 2018-12-31
grc: fix gr-fec GRC definitions.

jdemel:
Several FECAPI blocks have issues to generate correct Python code. These problems should be fixed now.

Issues with the  option persist, though. Those must be handled in the encoder/decoder block definitions.

As a side note: basically all examples `.grc` files are still XML definitions which are auto converted. Will those be converted at a later date? Or should they be updated whenever one works on them?

I'd like to add another commit with a few fixes to those files which make them work with Py3k. Mainly, integer vs. float division problems.

### DIFF
--- a/gr-fec/grc/variable_cc_decoder_def_list.block.yml
+++ b/gr-fec/grc/variable_cc_decoder_def_list.block.yml
@@ -48,12 +48,12 @@ parameters:
     default: '-1'
 -   id: mode
     label: Streaming Behavior
-    dtype: enum
-    options: [fec.CC_STREAMING, fec.CC_TERMINATED, fec.CC_TAILBITING, fec.CC_TRUNCATED]
+    dtype: raw
+    options: ['fec.CC_STREAMING', 'fec.CC_TERMINATED', 'fec.CC_TAILBITING', 'fec.CC_TRUNCATED']
     option_labels: [Streaming, Terminated, Tailbiting, Truncated]
 -   id: padding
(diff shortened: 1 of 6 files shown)


gr-fec/grc/variable_cc_decoder_def_list.block.yml
```diff
@@ -48,12 +48,12 @@ parameters:
...
-    options: [fec.CC_STREAMING, fec.CC_TERMINATED, fec.CC_TAILBITING, fec.CC_TRUNCATED]
+    dtype: raw
+    options: ['fec.CC_STREAMING', 'fec.CC_TERMINATED', 'fec.CC_TAILBITING', 'fec.CC_TRUNCATED']
     option_labels: [Streaming, Terminated, Tailbiting, Truncated]
 -   id: padding
     label: Byte Padding
-    dtype: enum
+    dtype: bool
     default: 'False'
     options: ['False', 'True']
```
haakov: `options` and `option_labels` are ignored for `bool` types, and can be removed.

mbr0wn:
Manually merged (4813feed742142507b4e43a9f2b431bf8de5dd89) and amended with @haakov's comments.

Thanks @jdemel!

gr-fec/grc/variable_cc_decoder_def_list.block.yml
mbr0wn: Manually amended.

PR #2279, opened 2019-01-01
cmake: remove obsolete code

jdemel:
GR 3.8 requires CMake 3.5.1 but it contained a workaround for CMake < 3.1 in its root `CMakeList.txt` file which is now obsolete. Thus, this patch removes the now obsolete workaround.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -158,12 +158,6 @@ ELSE()
     message(warning "C standard could not be set because compiler is not GNU, Clang or MSVC.")
 ENDIF()
 
-# if cmake version is < 3.1, explicitly set C standard to use.
-IF(${CMAKE_VERSION} VERSION_LESS "3.1")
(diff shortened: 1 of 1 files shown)


mbr0wn:
Ah @jdemel please keep the line lengths to <= 72 characters in git commit messages.

jdemel:
@mbr0wn which tool do you use to write commit messages? For now I just use plain git on the command line. Therefore, I would insert line breaks based on gut feeling. I don't count characters.

noc0lour:
I'd bet on vim syntax highlighting. Maybe with some correct plugin/config.

mbr0wn:
@jdemel yeah don't use `-m` with `git commit`. I started writing much better commit messages once I used an editor for all my commit messages.

btashton:
I use a client side git hook to enforce a few style rules, then it does not matter what tool I use.
see Client-Side Hooks here https://git-scm.com/book/en/v2/Customizing-Git-An-Example-Git-Enforced-Policy

noc0lour:
@mbr0wn that's actually true, creating commit messages with `-m` creates very short, complex statements ;)

PR #2281, opened 2019-01-03
grc: Suppress annoying Gtk message on exit from command line.

drmpeg:


marcusmueller:
Improved commit msg, manually put onto master, and done! Thank you!

PR #2293, opened 2019-01-05
Fix build on NetBSD.

mboersch:
There was an #ifdef missing in lib/thread/thread.c and in gr-vocoder
the include directive for gsm.h contains a gsm/ prefix.
The cmake FindGSM.cmake looks for **/include/ and **/include/gsm
anyways.

### DIFF
--- a/gr-vocoder/lib/CMakeLists.txt
+++ b/gr-vocoder/lib/CMakeLists.txt
@@ -104,6 +104,9 @@ endif(LIBGSM_LIBRARIES)
 
 add_library(gnuradio-vocoder SHARED ${gr_vocoder_sources})
 target_link_libraries(gnuradio-vocoder ${vocoder_libs})
+if(LIBGSM_INCLUDE_DIR)
+    target_include_directories(gnuradio-vocoder PUBLIC ${LIBGSM_INCLUDE_DIR})
+endif(LIBGSM_INCLUDE_DIR)
 GR_LIBRARY_FOO(gnuradio-vocoder)
 
 if(ENABLE_STATIC_LIBS)
--- a/gr-vocoder/lib/gsm_fr_decode_ps_impl.h
+++ b/gr-vocoder/lib/gsm_fr_decode_ps_impl.h
@@ -26,7 +26,7 @@
 #include <gnuradio/vocoder/gsm_fr_decode_ps.h>
 
 extern "C"{
-#include "gsm/gsm.h"
+#include "gsm.h"
 }
 
 namespace gr {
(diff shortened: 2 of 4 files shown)


michaelld:
I'd want to verify that the updated #include's now work. Just because CMake searches for a header in various paths & finds it in one such path doesn't mean that the path ends up being correctly used. For example in this specific case, even though CMake is directed to find GSM, only the GSM_LIBRARY is actually used; GSM_INCLUDES is not currently used. That needs to be fixed too, ideally, for this change to work; note that this change might work because the GSM_INCLUDES are part of some usual system path that's already part of the include_directories ... & it might not too. I'll have to install GSM to verify.

mbr0wn:
Are you saying that the include paths now work on NetBSD but before they don't? If FindGSM.cmake adds both include paths, why remove the gsm prefix? No other dependency does this (i.e. not provide an include path prefix). Is libgsm packaged such on NetBSD that the files get plopped straight into `/usr/include`?

gr-vocoder/lib/gsm_fr_decode_ps_impl.h
```diff
@@ -26,7 +26,7 @@
 #include <gnuradio/vocoder/gsm_fr_decode_ps.h>
 
 extern "C"{
-#include "gsm/gsm.h"
+#include "gsm.h"
```
mbr0wn: Not yet convinced this non-standard includery is necessary.

marcusmueller: agreeing with @mboersch here.

gr-vocoder/lib/CMakeLists.txt
```diff
@@ -104,6 +104,9 @@ endif(LIBGSM_LIBRARIES)
 
 add_library(gnuradio-vocoder SHARED ${gr_vocoder_sources})
 target_link_libraries(gnuradio-vocoder ${vocoder_libs})
+if(LIBGSM_INCLUDE_DIR)
+    target_include_directories(gnuradio-vocoder PUBLIC ${LIBGSM_INCLUDE_DIR})
```
marcusmueller: @noc0lour that's not modern CMake, is it? How do I fix it?

noc0lour: the FindGSM might need some touch ups (:

noc0lour:
@mboersch Can you try `master` again? The CMake changes are in and  should fix the issue you see.

noc0lour:
Merged first commit in 8ecdef41e47641f820fff73e0ad11c0890544b71 Second commit is supersed. 
Include path is now used and thus the prefix is not needed anymore. Thanks!

PR #2294, opened 2019-01-06
includes: use quotes for local includes

noc0lour:


noc0lour:
Side note: This is a preliminary change for the cmake target since it removes the current directory from the system include path (and suddenly the compiler stops looking for system includes in the current workdirectory).

PR #2298, opened 2019-01-08
cmake: Remove unused FindUSB.cmake module

mbr0wn:


noc0lour:
Note: was used for funcubedongle.

PR #2301, opened 2019-01-11
dtv: Bug fixes for ATSC receiver.

drmpeg:


gr-dtv/lib/atsc/atsc_sync_impl.cc
```diff
@@ -35,8 +35,8 @@ namespace gr {
     static const double ADJUSTMENT_GAIN = 1.0e-5 / (10 * ATSC_DATA_SEGMENT_LENGTH);
     static const int	SYMBOL_INDEX_OFFSET = 3;
     static const int	MIN_SEG_LOCK_CORRELATION_VALUE = 5;
-    static const char   SSI_MIN = -16;
-    static const char   SSI_MAX =  15;
+    static const signed char   SSI_MIN = -16;
```
mbr0wn: Doesn't `char` default to `signed`?

mbr0wn: Nope, looked it up. Learned something again. This change is legit.

PR #2304, opened 2019-01-12
uhd: fix freq_hopping example - missing 'time'

mbr0wn:
Freq_hopping example wasn't working correctly and symptom of it was that
frequency changes were happening in random moments. The reason was that
time of execution of tune commands wasn't specified and so they were
executed as non-timed commands.

In case of of full tuning (dsp_tuning == false) more changes were
needed. 'tx_freq' tag can't be executed as a timed command at all, so I
changed it to 'tx_command' with 'freq' element and also added missing
'time'.

I found out about the issue but credit of fixing it by adding 'time'
element to 'tx_command' tag value goes to Sylvain Munaut, as this change
was suggested by him.

Note: This replaces https://github.com/gnuradio/gnuradio/pull/1759


mbr0wn:
@ptrkrysik FYI.

Issue #2285, opened 2019-01-04
Test all flow graphs, make sure they're converted to YAML


mbr0wn:
Let's make sure we've gone over all the flow graphs. Please help update this. If you can't edit the first comment (i.e., this one), post below and I or someone else will update this.

### gr-uhd

- [x] uhd_const_wave.grc
- [ ] uhd_fft.grc: Problems with function probe
- [ ] uhd_msg_tune.grc: Problems with GUI hints and the message passing
- [ ] uhd_normalized_gain.grc
- [ ] uhd_siggen_gui.grc
- [ ] uhd_two_tone_loopback.grc
- [ ] uhd_wbfm_receive.grc

### gr-fec

- [ ] ber_curve_gen.grc: **bus ports missing**
- [ ]  ber_curve_gen_ldpc.grc: **bus ports missing**
- [x] ber_test.grc
- [x] fecapi_async_decoders.grc: float division bug, FECAPI parallelism broken #2276
- [x] fecapi_async_encoders.grc: float division bug
- [x] fecapi_async_ldpc_decoders.grc: float division bug
- [ ] fecapi_async_ldpc_encoders.grc: float division bug, **encoder results differ** #2274 
- [x] fecapi_async_packed_decoders.grc: float division bug
- [x] fecapi_async_to_stream.grc: float division bug
- [x] fecapi_cc_decoders.grc: float division bug
- [x] fecapi_decoders.grc: float division bug
- [x] fecapi_encoders.grc: float division bug
- [x] fecapi_ldpc_decoders.grc: float division bug
- [x] fecapi_ldpc_encoders.grc: float division bug
- [x] fecapi_polar_async_packed_decoders.grc: float division bug
- [x] fecapi_polar_decoders.grc: float division bug
- [x] fecapi_polar_encoders.grc: float division bug
- [ ] fecapi_tagged_decoders.grc: error on open `TypeError: string indices must be integers`
- [x] fecapi_tagged_encoders.grc
- [x] fecapi_tagged_ldpc_decoders.grc: float division bug
- [x] fecapi_tagged_ldpc_encoders.grc: float division bug
- [ ] polar_ber_curve_gen.grc: **bus ports missing**
- [x] polar_code_example.grc: float division bug
- [ ] tpc_ber_curve_gen.grc: **bus ports missing**

### gr-analog

- [x] noise_power.grc

### gr-audio

- [x] cvsd_sweep.grc
- [x] dial_tone.grc

### gr-channels

[...]

PR #2305, opened 2019-01-12
Convert gr-analog flowgraphs to YAML format

jdemel:
This is part of an ongoing effort to convert all flowgraphs in GR to the
new YAML format. cf #2285.
This commit converts all flowgraphs in `gr-analog` to the new YAML
format.


mbr0wn:
Merging manually.

PR #2306, opened 2019-01-12
Convert gr-audio GRC examples to YAML format

jdemel:
This commit converts the examples in `gr-audio` to the new YAML format.
Compare #2285.


mbr0wn:
MM.

mbr0wn:
@jdemel Thanks a lot!

PR #2308, opened 2019-01-12
Suppress some warnings.

drmpeg:


gr-fec/lib/reed-solomon/exercise.c
```diff
@@ -70,7 +70,7 @@ int trials){
 
       /* Load block with random data and encode */
       for(i=0;i<NN-NROOTS;i++)
-	block[i] = random() & NN;
+	block[i] = rand() & NN;
```
mbr0wn: This fix is an improvement, but we should stop using `rand()` altogether. See the comments here: https://en.cppreference.com/w/cpp/numeric/random/rand

mbr0wn:
Merged manually.

gr-fec/lib/reed-solomon/exercise.c
drmpeg: Agree. But this is just test code executed only by "make test", not runtime.

mbr0wn: That's exactly where we've been having issues with `rand()`, though, due to unreproducible results. @marcusmueller even wrote his own RNG code for this purpose.

Like I said, this is fine though. Just a general comment that anyone writing C++ code should move away from these functions.

PR #2309, opened 2019-01-13
dtv: Whitespace, fix indentation.

drmpeg:


mbr0wn:
@drmpeg Thanks for the change, and thanks in general for keeping the gr-dtv section clean! Question: Did you use the .clang-format file from the root of the project to do the change?

Also, @noc0lour was going to apply the .clang-format file to the entire source tree once we're going into stabilization mode for 3.8. That might make this PR superfluous.

If you did use the .clang-format file, and since you're the one spending most of the time in this files, we can probably merge this right now, but I would hate to apply a second formatting/whitespace diff to the file later.

drmpeg:
The .clang-format file doesn't follow the GNU Radio coding guidelines. Are we tossing the old rules?

mbr0wn:
The `.clang-format` *is* the coding guidelines, see also https://github.com/gnuradio/greps/blob/master/grep-0001-coding-guidelines.md.

mbr0wn:
And yes, we tossed the old rules a while ago (see the GREP again).

drmpeg:
Okay. I was a fan of 2 space indentation, but I'm guessing there's no room for discussion. Closing this PR.

drmpeg:
I'm just going to add one more comment. I have an old guideline compatible .clang-format file that (at least on my files) provides a minimum of changes (just long lines and one function parameter per line). This would minimize a complete "history reset".

I think it makes for more beautimous files, but it's all personal preference.

mbr0wn:
If you have a small whitespace change diff, or, honestly, even a big one, I'll merge it if it makes the files nicer and you more productive.

PR #2311, opened 2019-01-13
codec2: only check length of encoded messages in unit test

noc0lour:
previously the encoder & decoder pair was tested for a known input to produce a
precomputed output. Which can change with improvements in the codec2 library.
A similar test for the decoder needs to be implemented

### DIFF
--- a/gr-vocoder/python/vocoder/qa_codec2_vocoder.py
+++ b/gr-vocoder/python/vocoder/qa_codec2_vocoder.py
@@ -32,21 +32,49 @@ class test_codec2_vocoder (gr_unittest.TestCase):
     def tearDown (self):
         self.tb = None
 
-    def test001_module_load (self):
-        data = 40*(100,200,300,400,500,600,700,800)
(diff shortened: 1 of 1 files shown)


noc0lour:
After a short correspondence with David Rowe, he suggested checking known sequences (since codec2 pretty much stabilized now). Maybe we just have to regenerate the known output for some modes. (Checking the length separately makes still sense, though)

gr-vocoder/python/vocoder/qa_codec2_vocoder.py
```diff
@@ -32,21 +32,49 @@ def setUp (self):
...
-        self.assertEqual(expected_data, actual_result)
-        self.tb.disconnect(src, enc, dec, snk)
+        result = snk.data()
+        self.assertEqual(expected_length, len(result))
 
 if __name__ == '__main__':
-    # Note: The Vocoder is stateful, which means this test will produce failure when removing the xml option.
-    # Perhaps this is not the best way to test such a vocoder.
-    gr_unittest.run(test_codec2_vocoder, "test_codec2_vocoder.xml")
+    gr_unittest.run(test_codec2_vocoder)
```
mbr0wn: Should we just do that on all QA files?

mbr0wn: (Not relevant for this PR)

Issue #2307, opened 2019-01-12
GRC: QtGui: Widgets missing


jdemel:
If you run a flowgraph from GRC, e.g. `qtgui_tags_viewing.grc`, The `Qt GUI Range` widget will not show up. Thus, you can't control its value during runtime. It is missing from the generated Python file as well.

This happens with  _3.8tech-preview-231-g2826fd09 (Python 3.6.7)_

haakov:
I'll put up a PR for it tomorrow :)

drmpeg:
@haakov probably already knows this, but it's a regression in a recent commit. It was working fine not long ago.

![image](https://user-images.githubusercontent.com/5397196/51082717-d9cb7680-16c1-11e9-98dc-318a8bce6cd7.png)

gr-digital/python/digital/utils/tagged_streams.py needs to be fixed also.

-        tag.value = pmt.from_long(length / vlen)
+        tag.value = pmt.from_long(length // vlen)

There's another divide in that file that may need changing.

mbr0wn:
Looks like this is fixed.

PR #2296, opened 2019-01-06
grc: Fix Python Module block

haakov:
Fixes the Python Module mentioned in https://github.com/gnuradio/gnuradio/issues/2039.

**Edit:** Just noticed some warnings from the file chooser dialog. Don't have time to fix these tonight though.
```
/usr/local/lib/python3.7/dist-packages/gnuradio/grc/gui/Dialogs.py:362: PyGTKDeprecationWarning: Using positional arguments with the GObject constructor has been deprecated. Please specify keyword(s) for "title, parent, action, buttons" or use a class specific constructor. See: https://wiki.gnome.org/PyGObject/InitializerDeprecations
  transient_for=parent
/usr/local/lib/python3.7/dist-packages/gnuradio/grc/gui/Dialogs.py:362: PyGTKDeprecationWarning: The "buttons" argument must be a Gtk.ButtonsType enum value. Please use the "add_buttons" method for adding buttons. See: https://wiki.gnome.org/PyGObject/InitializerDeprecations
  transient_for=parent
/usr/lib/python3.7/site-packages/gi/overrides/Gtk.py:571: PyGTKDeprecationWarning: The keyword(s) "parent" have been deprecated in favor of "transient_for" respectively. See: https://wiki.gnome.org/PyGObject/InitializerDeprecations
  self._init(*args, **new_kwargs)
```

PR #2312, opened 2019-01-14
grc: Fix a bug where all variable blocks' make code is ignored when generating flowgraph

haakov:
Fixes https://github.com/gnuradio/gnuradio/issues/2307, a bug introduced by yours truly while fixing the Python Module block: https://github.com/gnuradio/gnuradio/pull/2296.

Haven't had time to test it yet, but I believe it should be correct now.


mbr0wn:
Hm, for the record, I think I merged that PR and I did test it. However, I only tested the feature that was being patched...

haakov:
Absolutely, my bad! :)

PR #2321, opened 2019-01-15
digital: Fix Python 3 divide for tagged streams.

drmpeg:


drmpeg:
I just found the duplicate file, so this is the wrong way to fix this. Closing.

drmpeg:
After some investigation, I'm not sure what the correct thing to do is. This patch does fix the immediate problem of gr-qtgui/examples/qtgui_tags_viewing.grc not running, so reopening.

mbr0wn:
Hrm. I'm going to merge this -- it's an improvement. I'm also not 100% sure what the perfect way to do this is, though.

PR #2323, opened 2019-01-17
dtv: VL-SNR bug fix and 16/32APSK constellation enhancement.

drmpeg:


drmpeg:
The constellation change is interesting. It provides a huge 25 dB S/N improvement on  STMicroelectronics DVB-S2 decoders.

The VL-SNR change was pointed out to me by a researcher in Ireland.

PR #2329, opened 2019-01-20
 blocks: Convert GRC examples to YAML format

jdemel:
This commit converts the examples in `gr-blocks` to the new YAML format.
This conversion is tracked in #2285.

The assertion in `grc/blocks_multiply_matrix_xx.block.yml` to check if
the selected tag propagation policy and the given matrix are compatible
is disabled. This might be an issue with `dtype: enum` and its
resolution in Mako.

Change `Extra Dict.` dtype in `File Meta Sink` to `raw`. Also, put
`pmt.make_dict()` as argument. Otherwise these flowgraphs fail due to a
type error.

The example file `simple_copy_controller.py` seems to fail with Py3k. It
fails on a `postMessage` call. Though, this seems like another bug.

The flowgraphs in `examples/msg_passing/hier` are converted as well,
though hier block generations seems to be broken. In order to load
`test_msg_hier_topblock.grc` it is necessary to:
1. Open and generate `test_msg_hier.grc` manually
2. Update the generated `test_msg_hier.py.blocks.yml` in the system
`.grc_gnuradio` folder. The input and output IDs are not unique. All
inputs are named `out0` all outputs `in0`. Rename the second input and
output to end in `1` in order to make things work.

### DIFF
--- a/gr-blocks/examples/ctrlport/comparing_resamplers.grc
+++ b/gr-blocks/examples/ctrlport/comparing_resamplers.grc
@@ -1,390 +1,173 @@
-<?xml version='1.0' encoding='ASCII'?>
-<flow_graph>
-  <timestamp>Thu May 23 22:25:13 2013</timestamp>
-  <block>
-    <key>options</key>
-    <param>
-      <key>id</key>
-      <value>comparing_resamplers</value>
-    </param>
-    <param>
-      <key>_enabled</key>
-      <value>True</value>
-    </param>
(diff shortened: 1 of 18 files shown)


mbr0wn:
Hm, stuff is broken. However, this is an improvement and I would prefer to have the conversion to YAML and the bugfix on separate commits anyway, so this can go in. Thanks for doing all the conversion work!

PR #2332, opened 2019-01-21
CMake: MSVC fixes for 3.8

gnieboer:
Two minor fixes for 3.8 to build on windows.
First just sets a compiler version, will error out later if the variable is empty

Second deals with the 8191 character limitation for command lines on windows by doing in 3 commands what could be otherwise done in one.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -230,6 +230,8 @@ IF(MSVC)
     IF(MSVC_VERSION LESS "${MSVC_MIN_VERSION}")
         MESSAGE(FATAL_ERROR "MSVC Versions < minimum version ${MSVC_MIN_VERSION}")
     ENDIF()
+    SET(cmake_c_compiler_version ${MSVC_VERSION})
+    SET(cmake_cxx_compiler_version ${MSVC_VERSION})
 ELSE()
     execute_process(COMMAND ${CMAKE_C_COMPILER} --version
             OUTPUT_VARIABLE cmake_c_compiler_version)
--- a/grc/CMakeLists.txt
+++ b/grc/CMakeLists.txt
@@ -142,7 +142,19 @@ GR_PYTHON_INSTALL(
 )
 
 GR_PYTHON_INSTALL(
-    DIRECTORY core gui converter
+    DIRECTORY core
+    DESTINATION "${GR_PYTHON_DIR}/gnuradio/grc"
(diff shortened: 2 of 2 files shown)


grc/CMakeLists.txt
```diff
@@ -142,7 +142,19 @@ GR_PYTHON_INSTALL(
 )
 
 GR_PYTHON_INSTALL(
-    DIRECTORY core gui converter
+    DIRECTORY core
```
mbr0wn: This seems fine, but do you know why we need this change?

michaelld: I'll second MB's remarks: I'm fine with the change, but wonder what's going on that requires it?

This is a CMake file, so maybe some version of CMake running on Windows can't properly handle multiple DIRECTORY arguments? Or, maybe GR_PYTHON_INSTALL isn't working correct on Windows? The GR_PYTHON_INSTALL code is pretty generic CMake scripting; should work in general if it does on any OS (which I'm thinking it does on at least Linux and OSX, since all of the desired install files seem seem to be being installed).

CMakeLists.txt
```diff
@@ -230,6 +230,8 @@ IF(MSVC)
     IF(MSVC_VERSION LESS "${MSVC_MIN_VERSION}")
         MESSAGE(FATAL_ERROR "MSVC Versions < minimum version ${MSVC_MIN_VERSION}")
     ENDIF()
+    SET(cmake_c_compiler_version ${MSVC_VERSION})
+    SET(cmake_cxx_compiler_version ${MSVC_VERSION})
```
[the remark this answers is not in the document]
gnieboer: The cmake_c_compiler_version does nothing as far as I know in windows-land.  This assignment displays a value (purely cosmetic) but more importantly several lines later a string function is called (replace I think?) and if this variable isn't set, then cmake fails with "parameters needed: 4 provided: 3) error.

grc/CMakeLists.txt
gnieboer: The cmake portion of this command does work as expected... it results in a single line command to python.exe with each file as a parameter.  (confirmed in the MSVC project files)

What doesn't work is that that command ends up exceeding the windows 8191 character limit so it gets truncated when passed to python.   So therefore some files don't get copied and things go downhill from there.

michaelld: @gnieboer OK then this change makes a LOT of sense, and is a good idea regardless of OS used. Thanks for the explanation!

PR #2278, opened 2019-01-01
Correctly building dummy block

dl1ksv:
This is a rebased version of #2197 and replaces #2197 directly.
As I failed to rebase directly I made this new pr. This should be more readable than the old one.
It fixes #2160.

PR #2334, opened 2019-01-21
digital: Convert GRC examples to YAML format

jdemel:
This commit converts the examples in `gr-digital` to the new YAML format.
This conversion is tracked in #2285.

Only the flowgraphs in `examples` and `examples/demod` are updated. Some
flowgraphs can't be opened due to some 'id' error.

### DIFF
--- a/gr-digital/examples/burst_shaper.grc
+++ b/gr-digital/examples/burst_shaper.grc
@@ -1,880 +1,268 @@
-<?xml version='1.0' encoding='ASCII'?>
-<?grc format='1' created='3.7.8'?>
-<flow_graph>
-  <timestamp>Fri Apr 17 12:31:57 2015</timestamp>
-  <block>
-    <key>options</key>
-    <param>
-      <key>id</key>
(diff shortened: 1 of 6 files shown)


drmpeg:
If you apply the patch from https://github.com/gnuradio/gnuradio/pull/2278, you can open the flow graphs with the 'id' error.

jdemel:
Did you test this with the flowgraphs in `gr-digital/examples/demod` that I did not convert in this patch? I ask because it seems that the error is different depending on the flowgraph. Sometimes, it is a missing block, sometimes it is a missing port, sometimes it is a non-unique port. That's my observation here.

I'd like to make a PR such as this one only depend on the latest master. One results of this conversion effort is that a lot of things get tested. Probably parts that are somehow not covert by unittests.

Anyways, I'd like to see #2278 merged and then I'd test the missing flowgraphs. Maybe they need to be fixed.

drmpeg:
I only tested the patch with gr-dtv flow graphs that had missing blocks.

Issue #2283, opened 2019-01-03
symbol_sync_cc stuck when error too large (resulting d_avg_period goes negative)


warmonkey:
### Bug
When input amplitude is too large, symbol_sync_cc get stuck.

### Screenshot
![frame1](https://user-images.githubusercontent.com/7480424/50651077-3ee5e480-0fbd-11e9-9da4-a4d7f473944f.png)
![frame2](https://user-images.githubusercontent.com/7480424/50651078-3f7e7b00-0fbd-11e9-8b1b-6c42ef2c4d5b.png)
### Tracing
* in file symbol_sync_cc_impl.cc line 537:
error = -488.4;
advance_loop(error);
* in file clock_tracking_loop.cc line 128 function advance_loop(float error) :
before: d_avg_period=10.771521, d_beta=0.0090558, error=-488.4;
`d_avg_period = d_avg_period + d_beta * error;`  
**after: d_avg_period = -0.417**
**out of range [d_min_avg_period, d_max_avg_period]**
* back to file symbol_sync_cc_impl.cc line 540:
`d_clock->phase_wrap();`
* in file clock_tracking_loop.cc line 145 function phase_wrap()

```
void clock_tracking_loop::phase_wrap()
{
    float period = d_avg_period; // One could argue d_inst_period instead
    float limit = period/2.0f;

    while (d_phase > limit)
            d_phase -= period;

    while (d_phase <= -limit)  //stuck here
          d_phase += period;
}
```

**phase = -87.227, limit=-0.417/2=-0.20895, now we stuck in these while loops;**
_ps: i think period should be d_avg_period, because avg period is estimated symbol period. when loop bandwidth relatively larger ( 0.05~0.25 ), limiting d_inst_period can make tracking error larger, even loop unlock. i'll benchmark both later._
### Workaround
apply limit to d_avg_period immediately after d_avg_period changed
* in file clock_tracking_loop.cc line 127

```
// Integral arm of PI filter
d_avg_period = d_avg_period + d_beta * error;
+++period_limit();
```

### Ultimate solution
* check every input, state and output in range when calculate control loop

PR #2335, opened 2019-01-22
gr-digital: Prevent the clock tracking loop from generating negative …

awalls-cx18:
…estimates

Under extreme circumstances of the error signal input, the
clock tracking loop would allow the average clock period and
instantaneous clock period estimates to go negative, resulting in
an infinite loop when wrapping the clock phase.

Change clock period estimate limiting to happen when the estimates
are being formed, so that the negative, non-sensical values aren't
allowed to happen.

This addresses Issue #2283


mbr0wn:
Should this patch also go onto master?

PR #2337, opened 2019-01-23
gr-audio: fix windows audio when receiving > 1 output_multiple per call

gnieboer:
So the windows audio sink was garbled when played from WAV files, but worked with other sources.  The flaw was my overlooking the details of "set_output_multiple()", which clearly indicates more than one chunk of data may be sent.  So the work() function was blindly overflowing, much to my (now) chagrin.

Easy fix... fill up one buffer at a time, and if no buffer is available, block if possible until one is.

The patch only affects windows audio, and can also be merged w/ master for 3.8

### DIFF
--- a/gr-audio/lib/windows/windows_sink.cc
+++ b/gr-audio/lib/windows/windows_sink.cc
@@ -135,81 +135,87 @@ namespace gr {
     {
       const float *f0, *f1;
 
-      // Pick the first available wave header (buffer)
-      // If none available, then wait until the processing event if fired and check again
-      // Not all events free up a buffer, so it could take more than one loop to get one
-      // however, to avoid a lock, only wait 1 second for a freed up buffer then abort.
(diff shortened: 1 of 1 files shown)


gr-audio/lib/windows/windows_sink.cc
```diff
@@ -135,81 +135,87 @@ namespace gr {
...
-        }
-        break;
-      }
-      if (write_waveout
-        (chosen_header) < 0) {
-        perror("audio_windows_sink: write failed");
-      }
-
-      return noutput_items;
+	  int samples_sent = 0;
```
mbr0wn: No biggie, but we need to do something about the garbled indents.

gnieboer: Arg.  I thought I had VS reset to spaces indent.   I'll force push another one.

gnieboer:
OK, a couple pushes later I think the indents are under control.

PR #2340, opened 2019-01-28
dtv: Expand LDPC lut array size for worst case table generation (3.7).

drmpeg:


drmpeg:
Some LDPC tables would cause the lut generator to overwrite the first element of the array, causing a segfault during run-time. All tables tested for worst case.

PR #2341, opened 2019-01-28
dtv: Expand LDPC lut array size for worst case table generation.

drmpeg:


noc0lour:
@mbr0wn Thanks for merging. I was looking at that and mentally quit at the macro and subsequent usage of local variables (: Overall the stuff looks like an improvement :+1:

drmpeg:
I agree it's pretty ugly. The original macro was fairly small and wasn't too horrible. Then the Rohde and Schwarz Engineering Competition optimization doubled the lines of code (and it's difficult to decipher). And now the bug fix adds even more code. I'll put this on my list to refactor.

Original macro:

```
#define LDPC_BF(TABLE_NAME, ROWS) \
for (int row = 0; row < ROWS; row++) \
{ \
    for(int n = 0; n < 360; n++) \
    { \
        for (int col = 1; col <= TABLE_NAME[row][0]; col++) \
        { \
            ldpc_encode.p[index] = (TABLE_NAME[row][col] + (n * q)) % pbits; \
            ldpc_encode.d[index] = im; \
            index++; \
        } \
        im++; \
    } \
} 
```

noc0lour:
Maybe need to think about adding C++ niceness as @mbr0wn suggested (:

Issue #2039, opened 2018-09-24
Python blocks are broken


argilo:
Python Blocks don't work on the master branch.

To reproduce, create a new flow graph and drop a Python block into it. The following error will appear on the console:
```
/usr/lib/python3.6/subprocess.py:766: ResourceWarning: subprocess 10052 is still running
  ResourceWarning, source=self)
Traceback (most recent call last):
  File "/home/argilo/prefix_next/lib/python3.6/dist-packages/gnuradio/grc/gui/DrawingArea.py", line 127, in _handle_drag_data_received
    self._flow_graph.add_new_block(selection_data.get_text(), coords)
  File "/home/argilo/prefix_next/lib/python3.6/dist-packages/gnuradio/grc/gui/canvas/flowgraph.py", line 149, in add_new_block
    block.params['id'].set_value(id)
KeyError: 'id'
```

haakov:
Thanks! Same problem with Python Module.

haakov:
I'm going away for about a week so I won't be able to fix this, but if anyone wants to have a crack at it while I'm gone, the fix for this issue is more or less identical to this one: https://github.com/gnuradio/gnuradio/pull/2102

For the Python Block at least, not so sure about the Python Module.

miguelzetina:
Hi, can I take this issue?

dkozel:
@miguelzetina Please do! If you solve the issue and post a PR just include a link to this issue.

mbr0wn:
This is still an issue. @miguelzetina, did you still want to fix it

mbr0wn:
OK, Python modules now fine. Python blocks still not.

ptrkrysik:
I think I have a fix for broken Python blocks here #2342. However both for new Python modules and new Python blocks the default code is not passed to a file after opening it in an editor. I haven't figured out yet where it is lost.

ptrkrysik:
I looked at this again today for a moment. The problem is somewhere in creation of OrderedDict here:
https://github.com/gnuradio/gnuradio/blob/master/grc/core/blocks/block.py#L76

Somehow '_source_code' is not copied from parameters_data to the new collection.
I see this if I paste after construction of the OrderedDict (i.e. in line https://github.com/gnuradio/gnuradio/blob/master/grc/core/blocks/block.py#L80):
        try:
          print("Epy source:", self.params['_source_code'].get_value())
          self.params['_source_code'].set_value('Some content!')
        except:
          pass

[...]

PR #2342, opened 2019-01-28
GRC: fix embedded python block

ptrkrysik:
Fix for a broken Python blocks  #2039. After this fix python blocks work, but after opening a new python block in an editor the default Python code is not pasted into the file.

### DIFF
--- a/grc/core/blocks/embedded_python.py
+++ b/grc/core/blocks/embedded_python.py
@@ -81,7 +81,7 @@ class EPyBlock(Block):
     label = 'Python Block'
     documentation = {'': DOC}
 
-    parameters_data = [dict(
+    parameters = [dict(
         label='Code',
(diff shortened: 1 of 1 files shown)


grc/core/blocks/embedded_python.py
```diff
@@ -81,17 +81,19 @@ class EPyBlock(Block):
...
+    parameters = [dict(
         label='Code',
         id='_source_code',
         dtype='_multiline_python_external',
         value=DEFAULT_CODE,
         hide='part',
     )]
-    inputs_data = []
-    outputs_data = []
+    inputs = []
```
ptrkrysik: I have some doubt now if removing '_data' from inputs_data and outputs_data is correct thing to do.

skoslowski:
superseded by #2358

PR #2345, opened 2019-01-31
QtGui: fix 'Show Msg Port' slider

PE1NUT:
The slider for 'Show Msg Port' worked the wrong way around. As it actually triggers a 'hide' in the Yaml, I've changed it to 'Hide Msg Port', so the behaviour and default value actually matches what is shown by the slider.

### DIFF
--- a/gr-blocks/grc/blocks_copy.block.yml
+++ b/gr-blocks/grc/blocks_copy.block.yml
@@ -25,9 +25,7 @@ parameters:
 -   id: showports
     label: Show Msg Ports
     dtype: bool
-    default: 'True'
-    options: ['False', 'True']
-    option_labels: ['Yes', 'No']
+    default: 'False'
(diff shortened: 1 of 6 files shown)


haakov:
Great catch!

I'm afraid changing the parameter name to `hideports` will break compatibility with flowgraphs from before this change. We could for example do `hide: ${ not showports }` to avoid this

PE1NUT:
Thanks for your comment. I've fixed the pull request, it no longer changes the parameter 'showports'.

PE1NUT:
Note that, because the sense of 'false' and 'true' was previously inverted, we might want to include flipping that when converting a flowgraph from xml to yml?

noc0lour:
Hm, I'm wary of the side-effect this might have for the half-done flowgraph conversion.

PR #2355, opened 2019-02-06
grc: Fix missing default code in python block and python module

ptrkrysik:
Fix for the missing default code in python block and python module. For more look here: #2039.
Please review if the key should be changed to 'default' or stay as 'value' and the change should be done elsewhere.


skoslowski:
manually merged

PR #2360, opened 2019-02-13
gr-audio: fix windows when receiving > 1 output_multiple per work() call

mbr0wn:
Master branch version of #2337.


gnieboer:
No issues, code is the same between the branches.

PR #2361, opened 2019-02-15
gr_modtool: Fix for parameter wrap in cli inputs

aru31:
cli `input()` replaced by `click.prompt()`, and parameters are now asked to be given in new line, thus fixing text wrap.

### DIFF
--- a/gr-utils/python/modtool/cli/add.py
+++ b/gr-utils/python/modtool/cli/add.py
@@ -127,7 +127,7 @@ def get_copyrightholder(self):
 def get_arglist(self):
     """ Get the argument list of the block to be added """
     if self.info['arglist'] is not None:
-        self.info['arglist'] = cli_input('Enter valid argument list, including default arguments: ')
+        self.info['arglist'] = click.prompt(click.style('Enter valid argument list, including default arguments: \n', fg='cyan'), prompt_suffix='')
 
 def get_py_qa(self):
     """ Get a boolean value for addition of py_qa """
--- a/gr-utils/python/modtool/core/info.py
+++ b/gr-utils/python/modtool/core/info.py
@@ -119,7 +119,7 @@ class ModToolInfo(ModTool):
                     inc_dirs += line.replace('GNURADIO_RUNTIME_INCLUDE_DIRS:{}='.format(path_or_internal), '').strip().split(';')
         except IOError:
             pass
-        if (not inc_dirs or inc_dirs.isspace) and self._suggested_dirs is not None:
+        if not inc_dirs and self._suggested_dirs is not None:
             inc_dirs = [os.path.normpath(path) for path in self._suggested_dirs.split(':') if os.path.isdir(path)]
         return inc_dirs
 


aru31:
@swap-nil7: please review this PR.

noc0lour:
Trying this patch out I've noticed it creates a newline after the input message and thus prevents wrapping. If I try removing the newline statement it wraps after max_line_length + length of the message. Is there a way to have both, input on the same line and no wrapping?

swap-nil7:
Yes, the user input should not happen in a new line. I will try to figure out a smooth methodology.

aru31:
@noc0lour: I tried, but maybe `click.style()` appears to override a function here which causes wrapping because the input() function alone works perfectly fine but it's without styles. Working on it.

aru31:
@noc0lour: sorry it took me a long time to respond back with an update. This weird line wrap issue happens because bash originally doesn't support text colours and textwrap is fooled by the raw string length of those control codes.
So, I tried various packages for proper wrapping of ANSI colours (ansiwrap), but they don't work in the way they are expected to.
So what I propose is that for the CLI arguments that may have input as a large number of characters (that is no sequence completer required there too) can be asked to input in next line which would thus remove the wrapping issue as in this PR, while other CLI arguments remain the same.

aru31:
`gr_modtool info` was not working earlier because `.isspace()` function cannot be directly used for list, and is not required in this case because `gr_modtool newmod` itself doesn't allow to name a module as spaces only.
wrapping issue is fixed for long arguments as described in this [comment](https://github.com/gnuradio/gnuradio/pull/2361#issuecomment-470041512)

swap-nil7:
Yes. There is an issue of re-wrapping while using ANSI colors on bash terminal and taking input on new line for argument list is looking ok. 
We will have to import external libraries for preventing text re-wrapping otherwise.
`click.prompt` works nicely for taking input on new line.

gr-utils/python/modtool/cli/add.py
```diff
@@ -127,7 +127,7 @@ def get_copyrightholder(self):
 def get_arglist(self):
     """ Get the argument list of the block to be added """
     if self.info['arglist'] is not None:
-        self.info['arglist'] = cli_input('Enter valid argument list, including default arguments: ')
+        self.info['arglist'] = click.prompt(click.style('Enter valid argument list, including default arguments: ' + '\n', fg='cyan'), prompt_suffix='')
```
swap-nil7: Remove String concatenation for `\n`.

gr-utils/python/modtool/core/info.py
```diff
@@ -119,7 +119,7 @@ def _get_include_dirs(self, mod_info):
                     inc_dirs += line.replace('GNURADIO_RUNTIME_INCLUDE_DIRS:{}='.format(path_or_internal), '').strip().split(';')
         except IOError:
             pass
-        if (not inc_dirs or inc_dirs.isspace) and self._suggested_dirs is not None:
+        if (not inc_dirs) and self._suggested_dirs is not None:
```
swap-nil7: Remove `()` from `(not inc_dirs)`.

aru31:
@swap-nil7: changes have been done. :)

PR #2371, opened 2019-02-19
Modtool: Add copyright feature

swap-nil7:
The feature for adding `copyrightholder` should be there in the `3.8 release`.

### DIFF
--- a/gr-utils/python/modtool/cli/add.py
+++ b/gr-utils/python/modtool/cli/add.py
@@ -25,6 +25,7 @@ from __future__ import absolute_import
 from __future__ import unicode_literals
 
 import re
+import getpass
 
 import click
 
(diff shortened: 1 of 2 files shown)


gr-utils/python/modtool/cli/add.py
```diff
@@ -25,6 +25,7 @@
 from __future__ import unicode_literals
 
 import re
+import getpass
```
mbr0wn: Another silent dependency?

gr-utils/python/modtool/cli/add.py
```diff
@@ -109,7 +110,16 @@ def get_blockname(self):
 def get_copyrightholder(self):
     """ Get the copyrightholder of the block to be added """
     if self.info['copyrightholder'] is None:
-        self.info['copyrightholder'] = '<+YOU OR YOUR COMPANY+>'
+        user = getpass.getuser()
```
mbr0wn: I don't think this is worth the cost of another dependency.

swap-nil7: I am not very sure but I think `getpass` is a internal python library.

mbr0wn: Ah alrighty.

gr-utils/python/modtool/cli/add.py
mbr0wn: Turns out no.

Issue #2369, opened 2019-02-18
GRC: Save As is broken, Save and Save Copy are OK


nickoe:
It looks like the "Save As"  menu does not work. I use the mount to select the menu.

When I try to "Save As" an existing grc it will not save. The normal Save for a modified or new grc file works, and "Save Copy" works as well, even adding _copy to the basename.

3.8tech-preview-280-g3a533f78  with python3

aru31:
The problem with save_as was that `_handle_action()` was not called in case the grc file was already saved, which caused this problem. So, as a fix, I made a try_catch block for that particular action, thus solving the issue.

nickoe:
@aru31 your fix almost works, see https://github.com/gnuradio/gnuradio/pull/2373#issuecomment-465312494

nickoe:
@aru31 yay, your fix was merged. May I suggest you have add a "Fixes" line to your commit next time, then the issue the patch is to fix will be closed automagically by github.

If you don't know what I am talking about, please see:
https://help.github.com/en/articles/closing-issues-using-keywords

aru31:
Thank you @nickoe, will keep that in mind :)

noc0lour:
Adding the a `fixes #1234` as separate, last line is very encouraged. :)

PR #2373, opened 2019-02-19
grc: Fix save as for existing grc file

aru31:
fix for issue #2369


nickoe:
This _almost_ works. For a python2 build it seems good, but for a python3 build, you need to modify the file, for example move a block around to make it **save as**. At least it is an improvement :)

aru31:
@nickoe I am currently working with python3 build, the save_as functions work even if I do not modify the file.

nickoe:
@aru31 Hmm, ok yeah, I just retested it. I guess I forgot to `make install`. I would say, merge it :heavy_check_mark:

Issue #2374, opened 2019-02-19
GRC: Dark theme not working well, input box text is white


nickoe:
I use dark gtk theme and for me I can't see text in white input boxes as the text becomes white as well. I can see it in the inputboxes that are colored. Is there somewhere that text color is hardcoded?

![image](https://user-images.githubusercontent.com/352202/53048557-56a5f900-3495-11e9-940a-2a8eba6eddfd.png)

I selected the rest of the test in the "Extra Dict." field, such that you can see that there are text input.

3.8tech-preview-280-g3a533f78

nickoe:
On a related note; the tabs in GRC to choose the active canvas has black text which does not work well on dark tabs.

chuangzhu:
Still doesn't work on KDE Plasma (5.20.5) when GTK theme set to `Breeze` and color scheme set to `Breeze Dark`. GRC version v3.8.2.0.

![image](https://user-images.githubusercontent.com/31200881/104962544-9b38c700-5a13-11eb-935a-f21a7791b3e9.png)

Here's my `~/.config/gtk-3.0/settings.ini`:

```ini
[Settings]
gtk-application-prefer-dark-theme=true
gtk-theme-name=Breeze
```

It seems that an unintentional mistake in #3017 resulted in `gtk-application-prefer-dark-theme` field not affecting the color:

```python
    prefer_dark = config.get(
        'Settings', Constants.GTK_INI_PREFER_DARK_KEY, fallback=None)
    if prefer_dark in ('1', 'yes', 'true', 'on'):
        theme_name = config.get(
            'Settings', Constants.GTK_INI_THEME_NAME_KEY, fallback=None)
-       return is_dark_theme(theme_name)
+       return True
    theme = subprocess.check_output(
        ["gsettings", "get", "org.gnome.desktop.interface", "gtk-theme"]
    ).decode('utf-8').strip().replace("'", "")
    return is_dark_theme(theme)
```

PR #2375, opened 2019-02-20
GRC: Fix color for input box in parameter widget with dtype raw

aru31:
Fix for issue #2374 
Changed the input color from `#FFFFFF` (white)  to `#A9A9A9` (dark grey)

### DIFF
--- a/grc/gui/ParamWidgets.py
+++ b/grc/gui/ParamWidgets.py
@@ -41,7 +41,7 @@ style_provider.load_from_data(b"""
     #dtype_string          { background-color: #CC66CC; }
     #dtype_id              { background-color: #DDDDDD; }
     #dtype_stream_id       { background-color: #DDDDDD; }
-    #dtype_raw             { background-color: #FFFFFF; }
+    #dtype_raw             { background-color: #23272A; }
 
     #enum_custom           { background-color: #EEEEEE; }
 """)


skoslowski:
I'd prefer a solution were we use the default color of the current theme for this.

aru31:
@skoslowski color code now  changed to #23272A

skoslowski:
I was actually proposing not to alter the background-color at all for this dtype. That way the default is kept which then works for both light and dark themes. The current color, white, was meant represent that - before themes were supported, that is

aru31:
`

aru31:
I am sorry for accidentally closing the pull request

aru31:
@skoslowski just to be clear, you mean that the color should change according to the default theme, Right?

nickoe:
I think some care must be taken to make sure there is a good contrast between the background color and the text color.

I guess in my dark theme the default text is light grey or white as seen on my screenshot. I would assume that gtk had a way to handle this a bit more nicely, but I am not familiar with framework.

noc0lour:
I think what @skoslowski wants to outline: The color scheme should be fully defined by the theme and not be hardcoded at all.

aru31:
@noc0lour: would this be feasible, I mean there are a lot of gtk-themes currently and it would really be a tedious job in order to explicitly define colour scheme for all of them.

marcusmueller:
@aru31 the matter is not whether it's feasible – any change we do needs to be for the better. We won't just pick a random color that works better with your dark theme and not really bad with your light theme, if it breaks an equal amount of people's readability with other color schemes. That's not progress!

So, yeah, I've played around with this myself, as my dark colorscheme doesn't allow for easy readability. My intuition is that coloring the background to give type information was a sub-optimal UX decision to begin with – input boxes should have the recognizable default input box coloring. We could work with colored borders, or with a separate color button, or simply a textual / symbolic hint.

nickoe:
I guess the coloring is ok, but it should be constrained to contrasting colors for the background and the foreground (text), and not let the foreground be chosen by the theme.

aru31:
@marcusmueller thank you for the insight. but what I was trying to say has been clearly stated by @nickoe, that it would be a much better solution if the font text and the background colour are chosen by colour matching both of them and not let the colours be decided by theme.

noc0lour:
I don't know how the Gtk theming engine works, but probably we can get information about the type of theme (dark, light).

aru31:
@noc0lour, you are absolutely right that we can classify Gtk themes as light or dark, but there is no particular or to say official categorization for them, a lot of different software (like tweak tools) have their own themes and have categorized them as dark or light accordingly.
Taking your point in the account, what instead can be done is that we can make our own list of dark or light themes and change the background colour accordingly, and the list can be updated time to time. For all other themes that are initially not present in the list, the default background colour can be used.

PR #2380, opened 2019-02-26
dtv: Prevent out of bounds array access in DVB-T2 interleaver (3.7).

drmpeg:


drmpeg:
Another bug in the R&S optimization code. The affected array is twist256s[8]. It doesn't currently cause a segfault, but could in the future. I also changed the large (518400 bytes) stack allocation to the heap.

Issue #2372, opened 2019-02-19
Support for std::vector<int64_t> for ControlPort registration


gr-sp:
While attempting to export some statistics from our application, we were surprised by the fact that the RPC registration facilities in GNURadio don't seem to support exporting vectors of int64_t. We really needed uint64_t, but it seems this isn't supported due to Thrift type limitations (unsigned integers aren't supported).

I have attached a patch (for GNURadio 3.7.13.4) that attempts to add this support.
[rpc-int64-vector.patch.txt](https://github.com/gnuradio/gnuradio/files/2881752/rpc-int64-vector.patch.txt)

After applying this patch, I was able to perform something akin to the following:
```
add_rpc_variable(
   rpcbasic_sptr(new rpcbasic_register_get<MyObject, std::vector<int64_t> > (...))
);
```

dkozel:
@gr-sp Would you be able to open up a pull request with the contents of the patch? That will help the review and merge process.

PR #2381, opened 2019-02-27
ctrlport: Add int64_t vector to rpcbasic_inserter

gr-sp:
Add std::vector<int64_t> template specialization to rpcbasic_inserter.

#2372


marcusmueller:
Nice work! I've directly merged that because our code contribution guidelines that under 10 lines of actual code, no CLA needs to be signed – and it's clear that the core of this was just filling in a new type in existing code.

But: if you'd feel the desire to contribute more, please do reach out to @bhilburn! We'll get you set up with a CLA, so that we can accept even very complex code contributions from you :)

PR #2388, opened 2019-03-02
gr-analog: Fix typo in documentation

johschmitz:


johschmitz:
You are right about the double commit but I didn't know how to do this with
the github web based editor. Any hints on that? For such simple changes I
usually just want to do it on the website but it doesn't seem to let me
bundle changes in several files easily.

Derek Kozel <notifications@github.com> schrieb am So., 3. März 2019, 10:54:

dkozel:
Ah. I don't think there's any way to do this with the web editor alone. You would need to pull the branches down to a local computer and merge them there. Either you can do that or one of us can do that when merging.

johschmitz:
For larger changes I usually go through my local machine and create proper
patches. Would just be nice if there was a + button in the web editor to
add more files and then create the pull request with all of them. Maybe
github doesn't want to allow such sophisticated features.

>

PR #2392, opened 2019-03-08
utils: gr_modtool whitespace fixes

drmpeg:


gr-utils/python/modtool/templates/gr-newmod/CMakeLists.txt
```diff
@@ -141,9 +141,7 @@ add_subdirectory(apps)
 add_subdirectory(docs)
 add_subdirectory(swig)
 add_subdirectory(python)
-if(ENABLE_GRC)
-    add_subdirectory(grc)
-endif(ENABLE_GRC)
+add_subdirectory(grc)
```
marcusmueller: yeah, that should be fine. 

We should, however, probably be doing this based on a tiny bits of smarts. For example, conditionalize on the GNU Radio installation actually having GRC enabled. However, our modern CMake Stuff currently doesn't have a GRC target, sooo. Yeah. Don't conditionalize on a variable nobody set.

gr-utils/python/modtool/tools/util_functions.py
```diff
@@ -57,15 +57,32 @@ def remove_pattern_from_file(filename, pattern):
 def str_to_fancyc_comment(text):
     """ Return a string as a C formatted comment. """
     l_lines = text.splitlines()
-    outstr = "/* " + l_lines[0] + "\n"
+    if len(l_lines[0]) == 0:
```
marcusmueller: Hm, while the above is actually a whitespace fix (in the sense: removes superfluous/erroneous whitespace from source code), this fixes code generation. OK if I split this out?

gr-utils/python/modtool/tools/util_functions.py
```diff
@@ -57,15 +57,32 @@ def remove_pattern_from_file(filename, pattern):
...
+        outstr = "/*" + l_lines[0] + "\n"
```
marcusmueller: If `len(l_lines[0])==0` I see little sense in including it in `outstr`?

gr-utils/python/modtool/tools/util_functions.py
```diff
@@ -57,15 +57,32 @@ def remove_pattern_from_file(filename, pattern):
...
+    else:
+        outstr = "/* " + l_lines[0] + "\n"
     for line in l_lines[1:]:
-        outstr += " * " + line + "\n"
+        if len(line) == 0:
+            outstr += " *" + line + "\n"
```
marcusmueller: same

gr-utils/python/modtool/tools/util_functions.py
```diff
@@ -57,15 +57,32 @@ def remove_pattern_from_file(filename, pattern):
...
+            outstr += " * " + line + "\n"
     outstr += " */\n"
     return outstr
 
 def str_to_python_comment(text):
     """ Return a string as a Python formatted comment. """
-    return re.compile('^', re.MULTILINE).sub('# ', text)
+    l_lines = text.splitlines()
+    if len(l_lines[0]) == 0:
+        outstr = "#" + l_lines[0] + "\n"
```
marcusmueller: same

gr-utils/python/modtool/tools/util_functions.py
```diff
@@ -57,15 +57,32 @@ def remove_pattern_from_file(filename, pattern):
...
+    else:
+        outstr = "# " + l_lines[0] + "\n"
+    for line in l_lines[1:]:
+        if len(line) == 0:
+            outstr += "#" + line + "\n"
```
marcusmueller: same as above :)

gr-utils/python/modtool/templates/gr-newmod/swig/CMakeLists.txt
```diff
@@ -22,8 +22,8 @@
 # Check if there is C++ code at all
 ########################################################################
 if(NOT howto_sources)
-	MESSAGE(STATUS "No C++ sources... skipping swig/")
-	return()
+    MESSAGE(STATUS "No C++ sources... skipping swig/")
```
marcusmueller: ah, thanks. That was my crime against humanity ;)

PR #2397, opened 2019-03-09
modtool: Instead of installing test_modtool, run it as test when testing

marcusmueller:


swap-nil7:
@marcusmueller: I think we should comment out the section of `Pylint tests`. They will take unnecessarily 8-10 seconds.

mbr0wn:
@swap-nil7 I think 8-10 seconds of testing for static code errors is worth it.

noc0lour:
Do we need to test for more python modules in CMake?
```
  File "/cache/workdir/Ubuntu_18_04_64_0/build_Ubuntu_18_04_64/src/gr-utils/python/modtool/tests/test_modtool.py", line 32, in <module>
    from pylint.epylint import py_run
ImportError: No module named pylint.epylint
```

swap-nil7:
I think we should make the pylint test conditional since `pylint` module is an added dependency for this test.

noc0lour:
I'll update the CI docker containers to contain pylint then we rerun the tests here.

marcusmueller:
Superseeded by #2431

marcusmueller:
@noc0lour you can still add pylint if you want.

PR #2398, opened 2019-03-10
Fix color for input boxes in parameter widget according to dark-gtk themes

### LINKED ISSUE #2374: GRC: Dark theme not working well, input box text is white

aru31:
Fixes #2374

### DIFF
--- a/grc/gui/ParamWidgets.py
+++ b/grc/gui/ParamWidgets.py
@@ -17,34 +17,77 @@
 
 from __future__ import absolute_import
 import os
+import configparser
+import subprocess
 
 from gi.repository import Gtk, Gdk
 
+from . import Constants
 from . import Utils
 
 
+config = configparser.ConfigParser()
+config.read(os.path.expanduser('~/.config/gtk-3.0/settings.ini'))
+theme = config.get('Settings', 'gtk-application-prefer-dark-theme')
 style_provider = Gtk.CssProvider()
+dark_theme_config = b"""
+                         #dtype_complex         { background-color: #3399FF; }
+                         #dtype_real            { background-color: #FF8C69; }
+                         #dtype_float           { background-color: #FF8C69; }
+                         #dtype_int             { background-color: #00FF99; }
 
-style_provider.load_from_data(b"""
-    #dtype_complex         { background-color: #3399FF; }
-    #dtype_real            { background-color: #FF8C69; }
-    #dtype_float           { background-color: #FF8C69; }
-    #dtype_int             { background-color: #00FF99; }
+                         #dtype_complex_vector  { background-color: #3399AA; }
+                         #dtype_real_vector     { background-color: #CC8C69; }
+                         #dtype_float_vector    { background-color: #CC8C69; }
+                         #dtype_int_vector      { background-color: #00CC99; }
 
-    #dtype_complex_vector  { background-color: #3399AA; }
-    #dtype_real_vector     { background-color: #CC8C69; }
-    #dtype_float_vector    { background-color: #CC8C69; }
-    #dtype_int_vector      { background-color: #00CC99; }
+                         #dtype_bool            { background-color: #00FF99; }
+                         #dtype_hex             { background-color: #00FF99; }
+                         #dtype_string          { background-color: #CC66CC; }
+                         #dtype_id              { background-color: #DDDDDD; }
+                         #dtype_stream_id       { background-color: #DDDDDD; }
+                         #dtype_raw             { background-color: #23272A; }
 
-    #dtype_bool            { background-color: #00FF99; }
-    #dtype_hex             { background-color: #00FF99; }
-    #dtype_string          { background-color: #CC66CC; }
-    #dtype_id              { background-color: #DDDDDD; }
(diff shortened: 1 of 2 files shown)


aru31:
Refer PR #2375. As per this [comment](https://github.com/gnuradio/gnuradio/pull/2375#issuecomment-471177082), different input box colors can now be provided for dark-gtk themes.
I don't know much about color combinations, so if you could help me figuring out the colors best suited for dark themes. :)

nickoe:
This does not seem to work for me.

```
gsettings get org.gnome.desktop.interface gtk-theme
'Adwaita'
```

For gtk2
```
cat ~/.gtkrc-2.0 | grep gtk-theme-name
gtk-theme-name="Adwaita-dark"
```

and for gtk3
```
grep -r gtk-theme-name ~/.config/gtk-3.0/
/home/nickoe/.config/gtk-3.0/settings.ini:gtk-theme-name=Adwaita-dark
```

I am not sure if there is a theme config somewhere else, but both my gtk2 and gtk3 apps are dark when I run them.

aru31:
@nickoe: `gsettings get org.gnome.desktop.interface gtk-theme` gives us the current system theme [(I referred this link)](https://askubuntu.com/questions/765569/how-to-determine-which-theme-is-currently-enabled).
But different gtk version apps can have different themes for themselves, gtkrc-2.0, gtkrc-3.0 files have themes for gtk-2.0 and gtk-3.0 apps respectively.

nickoe:
Hmm, that seems a bit fragile. I tried to google a bit, and found this post which you may already have considered.

https://lzone.de/blog/Detecting%20a%20Dark%20Theme%20in%20GTK

It tries to detect what "polarity"  the contrast has by summing the color codes for normal text and background. Could this be used for the case of GRC?

aru31:
@nickoe, yes I think we can do this for GRC, will try definitely, but as it is mentioned in the link you shared in your comment, there is no official way of classifying a gtk theme as light or dark, so they implemented their own version of detecting dark gtk themes using color contrasts of text and background.
I will try this and get back to you after working on it. :)

nickoe:
@aru31 Thank you for picking the task. I will definitely help you test it :)

aru31:
@nickoe, I'm afraid we cannot use the fix you mentioned because this fix has to applied to gtk.h file of the system of the user, but we need something for the GRC itself.
Instead, your above [comment](https://github.com/gnuradio/gnuradio/pull/2398#issuecomment-471621149), gave me an idea that what we can do is to first check if `gtk-application-prefer-dark-theme=1` and `gtk-theme-name={some dark theme}`, if that's true, we can apply dark theme configuration for grc and if that's false, we can check for main system theme by `gsettings get org.gnome.desktop.interface gtk-theme`, and if this results in a dark theme, we can apply the same

aru31:
@nickoe, Please review my PR. :)

nickoe:
@aru31 Just a small update, I did try to test this, but it did not seem to work on my system, but I also had other issue with opening the about dialog, so I suspect I made PEBKAC appear here, so I shall retest later and check that I run grc correctly. I will probably not have time to do this very soon as I go on vacation for a week next week.

aru31:
Thanks, @nickoe.

grc/gui/ParamWidgets.py
```diff
@@ -17,34 +17,77 @@
...
+import subprocess
 
 from gi.repository import Gtk, Gdk
 
+from . import Constants
 from . import Utils
 
 
+config = configparser.ConfigParser()
+config.read(os.path.expanduser('~/.config/gtk-3.0/settings.ini'))
```
mbr0wn: This file might not exist.

grc/gui/ParamWidgets.py
```diff
@@ -17,34 +17,77 @@
...
+theme = config.get('Settings', 'gtk-application-prefer-dark-theme')
```
mbr0wn: ...and then this might throw.

grc/gui/ParamWidgets.py
```diff
@@ -17,34 +17,77 @@
...
+                        #dtype_bool            { background-color: #00FF99; }
+                        #dtype_hex             { background-color: #00FF99; }
+                        #dtype_string          { background-color: #CC66CC; }
+                        #dtype_id              { background-color: #DDDDDD; }
+                        #dtype_stream_id       { background-color: #DDDDDD; }
+                        #dtype_raw             { background-color: #FFFFFF; }
+
+                        #enum_custom           { background-color: #EEEEEE; }
+                    """
+if theme == '1' or theme == 'yes' or theme == 'true' or theme == 'on':
```
mbr0wn: This is a bit clunky.

mbr0wn:
Continued here: https://github.com/gnuradio/gnuradio/pull/3017

Issue #1994, opened 2018-08-27
Constant Source won't do char (byte) type


MattEttus:
Constant Source doesn't have a choice to output bytes.  Null source does.

MattEttus:
Added in 3.8.  Thanks!

PR #2400, opened 2019-03-13
Update of sig_source to support byte output (signed) #1994

eraltu:


mbr0wn:
However, the history needs cleaning up (squashed and rebased).

eraltu:
Thank you for reviewing. CLA process is initiated. I have written @trondeau an e-mail as stated in the wiki. Clean-up of history will follow shortly

noc0lour:
@eraltu looks like the info in the wiki is outdated. Please direct your mail to bhilburn at gnuradio.org

PR #2401, opened 2019-03-13
fixup! Modifying dummy block and error messages

dkozel:
I accidentally merged an older version of PR https://github.com/gnuradio/gnuradio/pull/2278 while fixing a typo. This fixup brings in the correct changes to dummy.py

I wanted to get a second set of eyes to prevent a second mistake.


dkozel:
@drmpeg took a quick look and I've double checked it. Going ahead with merging.

dkozel:
drmpeg took a look and I've double checked. Merged.

PR #2411, opened 2019-03-18
Reduce peak memory usage of Octave function read_cshort_binary

starsway:
Hi,

I was trying to read rather large complex<int16> recordings with **gr-utils/octave/read_cshort_binary.m** under _Ubuntu 18.04.2 LTS_ with a version of Octave 6.0.0 built from sources just today (18MAR2019) (_HG ID for this build is "1771bed38482"_), on a computer with:  _MemTotal: 32694812 kB_.

I kept getting memory overflows and Octave was quitting with the error message: 'octave exited with signal 6'.

I realized that, on the contrary, **gr-utils/octave/read_short_binary.m** _was_ able to read the same data which allowed me to play with the data, and I eventually managed to reduce the memory highest-water mark in the '...cshort...' function when compared to the peak usage induced by current code:

`cv = v(1:2:end)+v(2:2:end)*j;`

I'm copying a few logs below to give a more accurate idea of the amounts of data involved. With the change I'm suggesting, there is no memory overflow when calling ' **read_cshort_binary** ' and I'm able to load and work with the data.

I am submitting this pull request in case it could help anyone facing the same issue.

Thank you.

----

**Octave logs showing in detail how the suggested change manages to stay within the memory limits:**

```
>> addpath('~/sdr/gnuradio/gr-utils/octave/');
>> dat = read_short_binary('~/sdr/recordings/lbuc100/20190305/recorded-data-20190305-143055_samples');
>> dat_r=dat(1:2:end);
>> dat_i=dat(2:2:end)*j;
>> whos
Variables visible from the current scope:

variables in scope: top scope

   Attr Name            Size                     Bytes  Class
   ==== ====            ====                     =====  =====
        dat    1199833088x1                 9598664704  double
   c    dat_i   599916544x1                 9598664704  double
        dat_r   599916544x1                 4799332352  double

Total is 2399666176 elements using 23996661760 bytes

>> clear dat;
>> dat=dat_r+dat_i;
>> whos
Variables visible from the current scope:

variables in scope: top scope

   Attr Name           Size                     Bytes  Class
   ==== ====           ====                     =====  =====
   c    dat    599916544x1                 9598664704  double
   c    dat_i  599916544x1                 9598664704  double
        dat_r  599916544x1                 4799332352  double

Total is 1799749632 elements using 23996661760 bytes
```

----

**Commit message**

Attempts to read large amounts of data with 'read_cshort_binary' may
make Octave 6.0.0 overflow and exit with the message:
'octave exited with signal 6'.
The purpose of this change is to use more sparingly memory in order
to avoid overflowing in cases where the computer *does* have enough
memory to perform the desired operation.

### DIFF
--- a/gr-utils/octave/read_cshort_binary.m
+++ b/gr-utils/octave/read_cshort_binary.m
@@ -42,5 +42,8 @@ function cv = read_cshort_binary (filename, count)
   else
     v = fread (f, count, 'short');
     fclose (f);
-    cv = v(1:2:end)+v(2:2:end)*j;
+    v_r = v(1:2:end);
+    v_i = v(2:2:end)*j;
+    clear v;
+    cv = v_r+v_i;
   end


hcab14:
What happens if you do this: (no need to implicitly convert int16 to double)
`cv = complex(single(v(1:2:end)), single(v(2:2:end)));`

starsway:
> What happens if you do this: (no need to implicitly convert int16 to double)
> `cv = complex(single(v(1:2:end)), single(v(2:2:end)));`

Hi - thanks for your prompt reply and suggestion.
The change you have suggested works as well (meaning that data is loaded with no out-of-memory problem), and leads to an array of singles with halved size. I must admit that I have no idea whether the fact that we now have a single array can force some of the subsequent operations to single precision, or have any other consequence.

```
>> addpath('~/sdr/gnuradio/gr-utils/octave/')
>> dat = read_cshort_binary('~/sdr/recordings/lbuc100/20190305/recorded-data-20190305-143055_samples');
>> whos
Variables visible from the current scope:

variables in scope: top scope

   Attr Name           Size                     Bytes  Class
   ==== ====           ====                     =====  =====
   c    dat    599916544x1                 4799332352  single

Total is 599916544 elements using 4799332352 bytes
```

marcusmueller:
crazy idea: instead of introducing a ttemporary variable `v` just to clear it later, why not directly

```matlab
cv = fread (f, count, 'short');
cv = complex(cv(1:2:end), cv(2:2:end));
```

Regarding precision: `single` has enough bits to represent any 16 bit integer, so that shouldn't change anything on the input side. On the output side, it might depend – so, I'm not in favour of changing the data type we produce, as this function might be used in *old* matlab code in actual complex simulations, e.g. from scientific publications.


So, I'd recommend the above solution, if that's OK with you, @starsway – it should use as little memory as @hcab14's solution, with the added elegance of avoiding having to clean anything. (sadly, it looks a bit like the qbasic that was the first  code that I ever read.)

starsway:
Hi @marcusmueller ,

Thank you for having looked into this.
I have just tried the solution you suggested and unfortunately it does not work on the same file and the same computer as the ones I used to test my own patch and then @hcab14 suggestion (I described both file and computer in my first post).

Octave 6 'blows up' (meaning the window closes and the shell from which I launched './run-octave --gui' says 'octave exited with signal 6').

I agree with you that the patch I submitted is inelegant, to say the least ('ugly' is another word that comes to mind). But it seems that trying to perform two operations (extracting elements with odd indices on the one hand, and even indices on the other) on the huge vector read from the file at the same time -without altering in some way the precision- is too much to fit in the memory...

Kind Regards.

marcusmueller:
It's not *that* ugly! So, yeah, because @hcab14's change sadly changes the data type (and I don't want to risk doing that in a function that might be used all over the place in publication-grade simulations), I'll pick yours. Thank you!

starsway:
Thanks a lot.

PR #2412, opened 2019-03-18
GRC: Fix for GRC crash in case of flowgraph error

### LINKED ISSUE #2327: GRC: hier block with multiple message pad source ports crashes GRC
If you generate `/gr-blocks/examples/msg_passing/hier/test_msg_hier.grc` it does produce an invalid `.grc` file which prevents GRC from loading. GRC will complain about non-unique port IDs. If you look at the generated `.grc` file, this is indeed the case.

What makes that issue really annoying is the fact, that once this flowgraph is generated, GRC will crash because it can't load that particular flowgraph.

aru31:
Fixes #2327

### DIFF
--- a/grc/core/Messages.py
+++ b/grc/core/Messages.py
@@ -25,6 +25,10 @@ import sys
 MESSENGERS_LIST = list()
 _indent = ''
 
+# Global FlowGraph Error and the file that caused it
+global flowgraph_error
+global flowgraph_error_file
+
 
 def register_messenger(messenger):
     """
--- a/grc/core/platform.py
+++ b/grc/core/platform.py
@@ -177,6 +177,9 @@ class Platform(Element):
                 except Exception as error:
                     logger.exception('Error while loading %s', file_path)
                     logger.exception(error)
+                    Messages.flowgraph_error = error
+                    Messages.flowgraph_error_file = file_path
+                    continue
 
         for key, block in six.iteritems(self.blocks):
             category = self._block_categories.get(key, block.category)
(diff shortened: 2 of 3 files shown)


aru31:
As in this [comment](https://github.com/gnuradio/gnuradio/issues/2327#issuecomment-473921769), I figured out why `Messages.send()` didn't work, it was because the `Message.send(message)` function loaded the message in GRC before it even opened, so fix for this was to define a global variable that stored the error, and thus print it on the terminal after GRC loaded.

mbr0wn:
...and no one complained about them, either.

bastibl:
This seems to break GRC for me:
```
Error: module 'gnuradio.grc.core.Messages' has no attribute 'flowgraph_error'
>>> Failure
  File "/home/basti/usr/gnuradio/lib/python3.7/dist-packages/gnuradio/grc/gui/MainWindow.py", line 263, in new_page
    if str(Messages.flowgraph_error) is not None:
AttributeError: module 'gnuradio.grc.core.Messages' has no attribute 'flowgraph_error'
```

mbr0wn:
Yeah it breaks GRC. I pushed an emergency fixup. Sorry, I merged to many PRs yesterday I must have thought I tested it when I didn't.

mbr0wn:
@aru31 Can you see if this still fixes your problem after the fixup commit?

aru31:
@mbr0wn Sorry, for the bad pull request, I didn't initialize the global variables which caused GRC to break.

PR #2414, opened 2019-03-20
grc sometimes generates wrong yml description of  hier2 block

dl1ksv:
If  a hier2 block contains more than one ( input or output ) message port
grc generates a wrong yml description of this block.

If you restart grc , grc terminates with an error messages is no longer usable.

You can check with the example in

gr-blocks/examples/msg_passing/hier

This example is running in 3.7 without any problems.

The modification in platform.py avoids the termination of grc
if there is an error in the flowgraph to be loaded.

The modification fixes the bug in the generated yml code.

Another proposal to prevent the termination of grc can be found in #2412 . But this pr does not fix the wrong yml code.

### DIFF
--- a/grc/core/generator/hier_block.py
+++ b/grc/core/generator/hier_block.py
@@ -97,8 +97,6 @@ class HierBlockGenerator(TopBlockGenerator):
             data[direction] = []
             for port in get_hier_block_io(self._flow_graph, direction):
                 p = collections.OrderedDict()
-                if port.domain == Constants.GR_MESSAGE_DOMAIN:
-                    p['id'] = port.key
                 p['label'] = port.parent.params['label'].value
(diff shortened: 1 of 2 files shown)


aru31:
@dl1ksv, you are absolutely right that #2412 doesn't fix the bug in YAML, but I thought that #2412 could be more of a generalized solution for any flowgraph error that occurs in the GRC and so we still can get the error message printed on the GRC terminal itself without terminating it.

dl1ksv:
@aru31 That's why I referenced your PR

aru31:
Thanks. :)

noc0lour:
Cleaned up the commit message and merged in bcb67601eb166acf6a07fd9e1053aab8213d0861

PR #2418, opened 2019-03-24
gr_modtool: fix for .yml file make template

### LINKED ISSUE #2417: gr_modtool .yml file make template missing curly braces
The 3.8 gr_modtool creates a yaml make template that looks like this:

make: module_name.block_name($parameter1, $parameter2)

Which doesn't work. Should be:

make: module_name.block_name(${parameter1}, ${parameter2})

aru31:
Fixes #2417

### DIFF
--- a/gr-utils/python/modtool/tools/util_functions.py
+++ b/gr-utils/python/modtool/tools/util_functions.py
@@ -108,7 +108,7 @@ def strip_arg_types_grc(string):
         return ""
     else:
         string = strip_default_values(string)
-        return ", ".join(['$' + part.strip().split(' ')[-1] for part in string.split(',')])
+        return ", ".join(['${' + part.strip().split(' ')[-1] + '}' for part in string.split(',')])
 
 def get_modname():
     """ Grep the current module's name from gnuradio.project or CMakeLists.txt """


drmpeg:
Sorry, but you're fixing the wrong thing. The make template line is here:

https://github.com/gnuradio/gnuradio/blob/master/gr-utils/python/modtool/templates/templates.py#L551

aru31:
@drmpeg thank you for your review.
I have fixed the make template missing curly braces and also included missing curly brackets in make parameters of .yml files.

drmpeg:
I've tested both patches. Looks good.

BTW, It's correct to fix both places. When I said it was the wrong thing above, I only meant that in the context of issue #2417, where I was only concerned with the output of the gr_modtool "add" command.

However, the YAML generator (gr_modtool "makeyaml" command) probably needs a little more work to make it useful. But that's for another day.

aru31:
@drmpeg, I reverted the changes that I made in yaml generator.
Thanks :)

drmpeg:
If you want to fix the yaml generator in another PR, here are the issues.

1) Filename. Should be modulename_blockname.block.yml
2) id should be modulename_blockname
3) label should be blockname
4) category should be modulename with just the first letter capitalized.
5) make should have parameters in curly braces.

Basically, the yaml generator should create the same .yml file as the "add" function, but with the parameter and in/out definitions filled in.

aru31:
Sure, I will work on this. :)
Thanks

noc0lour:
Merged in fc58c1f7fcd21432500a24dc220b5f42778dff98 with slight fixes in commit message wording. Thanks!

PR #2420, opened 2019-03-24
modtool: Update Python templates

jdemel:
Update templates for Python blocks and Python QA code. Previously, these
templates created broken import statements. Also, a deprecated call to
`gr_unittest.run(...)` is updated to not include a filename argument
anymore. Finally, the python block signature generator was updated.


swap-nil7:
Yes, the import statements are currently broken.

PR #2424, opened 2019-03-28
grc: Increase logging level to INFO.

drmpeg:


noc0lour:
Why remove? Probably it would be better to increase the default log level?

drmpeg:
Yes. I've updated the commit.

PR #2425, opened 2019-03-30
blocks: Modify File Source block to work with hardware devices.

drmpeg:


noc0lour:
Can you elaborate? Best put the reasoning into the commit message :+1:

drmpeg:
The block uses fseek and ftell, which don't work with named pipes and hardware devices.

noc0lour:
Is there a way to tell that by looking at the file(descriptor)? Otherwise your code looks fine.

drmpeg:
Yes, automatic detection is much better.

noc0lour:
@drmpeg I agree. But probably a rewrite using boost::filesystem would be better anyway. Using only C-API there is no real portable way to distinguish files.

gr-blocks/lib/file_source_impl.cc
```diff
@@ -123,20 +129,40 @@ namespace gr {
...
-        if (start_offset_items) {
-          GR_LOG_WARN(d_logger, "file is too small for start offset");
-        }
-        else {
-          GR_LOG_WARN(d_logger, "file is too small");
+      if(fstat(fileno(d_new_fp), &st)) {
+        GR_LOG_ERROR(d_logger, boost::format("%s: %s") % filename % strerror(errno));
+        throw std::runtime_error("can't fstat file");
+      }
+      if(S_ISREG(st.st_mode)) {
```
noc0lour: Oh, this is for the automatic determination if it's seekable?

drmpeg: Yes, but I think it needs to be fixed for Windows (functions need leading underbars).

PR #2426, opened 2019-03-30
[WIP] uhd: Control of frequency and gain in both directions at the same moment

ptrkrysik:
Over a year ago mrjacobagilbert added "direction" key which can be used together with "freq" or "tune" keys of "tx_command" to control frequency in both directions with use of gr-uhd source or sink block and their command ports or tag processing capabilities: https://github.com/gnuradio/gnuradio/commit/cfb6e49e33d7f664268000e312367d091a60b9c3.

This is a great idea.

However underlying code was still not 100% ready for this, as it is impossible to control Rx and Tx freqs at the same time. This is because there is single vector for storing frequencies, shared by both Tx and Rx: https://github.com/gnuradio/gnuradio/blame/master/gr-uhd/lib/usrp_block_impl.cc#L648

Few months ago I prepared patches adding such capability for frequencies and gains, but I wasn't 100% happy with the code - especially the fact that I added repeating checks for direction pmt in each affected function that sets frequency.

Anyway it's very usable for me, so maybe others would want to make use of such capability as well. This is more work-in-progress pull request, so I'm open to any suggestions how to improve this commits before merging.

It probably would be also great to have ability to control other things in both directions i.e. why not (most importantly) gains, bw, rate and antenna? Maybe the code of uhd blocks should be restructured a bit to support this without bloating it too much the way I did.

If this pull request will be successful I would be happy to add the same improvement in maint-3.7 branch.

### DIFF
--- a/gr-uhd/include/gnuradio/uhd/usrp_block.h
+++ b/gr-uhd/include/gnuradio/uhd/usrp_block.h
@@ -159,7 +159,7 @@ namespace gr {
        * \param gain the gain in dB
        * \param chan the channel index 0 to N-1
        */
-      virtual void set_gain(double gain, size_t chan = 0) = 0;
+      virtual void set_gain(double gain, size_t chan = 0, pmt::pmt_t direction = pmt::PMT_NIL) = 0;
 
       /*!
        * Set the named gain on the dboard.
--- a/gr-uhd/lib/usrp_block_impl.cc
+++ b/gr-uhd/lib/usrp_block_impl.cc
@@ -640,15 +640,21 @@ void usrp_block_impl::_cmd_handler_looffset(const pmt::pmt_t &lo_offset, int cha
 
 void usrp_block_impl::_cmd_handler_gain(const pmt::pmt_t &gain_, int chan, const pmt::pmt_t &msg)
 {
+  //See if a direction was specified
+  pmt::pmt_t direction = pmt::dict_ref(
+      msg, cmd_direction_key(),
+      pmt::PMT_NIL // Anything except "TX" or "RX will default to the messaged block direction"
+  );
+
   double gain = pmt::to_double(gain_);
   if (chan == -1) {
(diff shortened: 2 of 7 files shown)


mbr0wn:
I'm going to have to close this; it's gotten too stale. I'm also not quite sure when you'd need that: Why can't you  just talk to the source for RX and the sink for TX?

ptrkrysik:
@mbr0wn suppose you have a system with Tx and Rx and you want to change frequency synchronously both in Tx and Rx channels.

Currently it's hard (without writing another GNU Radio block for USRPs): you can try to implement this by sending control PMT messages separately to UHD sink and source - but taking into account asynchronous nature of the messages that travel separately for Tx and Rx, it's unlikely it will work well (I don't know if it's even possible for this to work with current blocks).

This change is intended for controlling both Tx and Rx synchronously, mainly with use of stream tags. I'm currently using this functionality in a Radar application, where UHD sink receives stream tags with commands to do timed frequency changes. This way it implements arbitrary carrier frequency changes from one coherent processing interval to another.

This change also helps with controlling the frequency with use of message passing. I used it for example here for experimental implementation of frequency hopping for GSM with use of GNU Radio:
https://media.ccc.de/v/osmodevcon2019-115-frequency-hopping-for-sdr-based-gsm-mobile-station/playlist

Also consider the fact that currently you **can** control Rx or Tx frequency with use of UHD source (or sink) - but not both Rx and Tx frequencies at the same time. This pull request fixes that inconsistency and adds similar ability for gain.

The change was marked as WIP only because I didn't like code repetition and that maybe some better solution could be found that would enable to change with similar result everything else that can be controlled with use of UHD's timed commands.

Since no better solution was found yet I think it is be reasonable to accept this change. If you agree I can make another pull request with updated version (there was major code formatting update since this pull request was created that probably broken it). I already have the changes in my branch https://github.com/ptrkrysik/gnuradio/tree/bidir_ctrl_of_freq_and_gain .

PR #2431, opened 2019-03-31
modtool: fix modtool tests for missing pylint dependency

aru31:
gr-modtool tests depend on `pylint`, so earlier modtool tests failed in case of the missing dependency (pylint) which has been fixed in this PR. :)

### DIFF
--- a/gr-utils/python/modtool/core/base.py
+++ b/gr-utils/python/modtool/core/base.py
@@ -29,7 +29,22 @@ import re
 import glob
 import logging
 import itertools
-from types import SimpleNamespace
+
+try:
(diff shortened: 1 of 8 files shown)


marcusmueller:
Please don't do that!

We disable the unit test if we don't find pylint – this makes it looks like success if that test depency is missing; that's a lie!

Can we please extract the pylint-depending code to separate `test_xxx` methods and decorate them with [`@unittest.skipIf(someFlag)`](https://docs.python.org/2.7/library/unittest.html#skipping-tests-and-expected-failures)?

aru31:
@marcusmueller, working on it. :)

PR #2434, opened 2019-04-02
dtv: Fix DVB-S2X 64APSK constellation.

drmpeg:
Transcription error from specification (p and q reversed).


noc0lour:
Looking at that sad definition of constellation points it itches me that this might be suited for some `constexpr` loop to define those constants using the compiler. But looks good to me.

marcusmueller:
I'll agree with @noc0lour – we need to come up with "beautiful" `constexpr` magic for these tables, which we have a lot of all over the code base. However, this is clearly the right thing to do – merge.

Issue #2439, opened 2019-04-05
GRC: QSS Theme not working in 3.8


drmpeg:
Here's the error:
```
  File "/home/re/xfer/dvbs2_tx.py", line 252
    tb.setStyleSheetFromFile(/opt/gnuradio-3.7.12git/share/gnuradio/themes/dark.qss)
                             ^
SyntaxError: invalid syntax
```

noc0lour:
looks like it's not a string anymore?

drmpeg:
I used the file finder dialog. BTW, when the file finder dialog opens, you get a large PyGTK warning.

```
/opt/gnuradio-3.7.12git/lib/python3.5/dist-packages/gnuradio/grc/gui/ParamWidgets.py:327: PyGTKDeprecationWarning: Using positional arguments with the GObject constructor has been deprecated. Please specify keyword(s) for "title, parent, action, buttons" or use a class specific constructor. See: https://wiki.gnome.org/PyGObject/InitializerDeprecations
  transient_for=self._transient_for,
/opt/gnuradio-3.7.12git/lib/python3.5/dist-packages/gnuradio/grc/gui/ParamWidgets.py:327: PyGTKDeprecationWarning: The "buttons" argument must be a Gtk.ButtonsType enum value. Please use the "add_buttons" method for adding buttons. See: https://wiki.gnome.org/PyGObject/InitializerDeprecations
  transient_for=self._transient_for,
/usr/lib/python3/dist-packages/gi/overrides/Gtk.py:537: PyGTKDeprecationWarning: The keyword(s) "parent" have been deprecated in favor of "transient_for" respectively. See: https://wiki.gnome.org/PyGObject/InitializerDeprecations
  self._init(*args, **new_kwargs)
Gtk-Message: GtkDialog mapped without a transient parent. This is discouraged.
```

dl1ksv:
Possible fix for error and depreciation warnings in #2447

drmpeg:
Thanx for fixing the PyGTK warning. Very nice for those of us that start gnuradio-companion on the command line.

If you're interested in fixing another one, try right clicking in the console window of GRC. Here's the warning I get.
```
[...]

PR #2447, opened 2019-04-12
Add missing quotes in themes parameter of option block and fix some depreciation warnings

dl1ksv:
setStyleSheetFromFile requires a string as parameter.
While generating the python code the quotes are missing.
See #2439


drmpeg:
I've just tested this patch. Works perfectly.

marcusmueller:
Thanks!

PR #2451, opened 2019-04-18
socket_pdu_impl.cc: fix build with boost 1.70.0

### LINKED ISSUE #2446: Fails to build with boost 1.70: error: no member named 'get_io_service' in 'boost::asio::basic_socket_acceptor<boost::asio::ip::tcp, boost::asio::executor>'
See the downstream bug report here: https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=236601

ffontaine:
Fix #2446

Signed-off-by: Fabrice Fontaine <fontaine.fabrice@gmail.com>

### DIFF
--- a/gr-blocks/lib/socket_pdu_impl.cc
+++ b/gr-blocks/lib/socket_pdu_impl.cc
@@ -29,6 +29,12 @@
 #include <gnuradio/io_signature.h>
 #include <gnuradio/blocks/pdu.h>
 
+#if BOOST_VERSION >= 107000
+#define GET_IO_SERVICE(s) ((boost::asio::io_context&)(s)->get_executor().context())
+#else
+#define GET_IO_SERVICE(s) ((s)->get_io_service())
+#endif
+
 namespace gr {
   namespace blocks {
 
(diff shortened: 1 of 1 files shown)


michaelld:
I think this is a good solution to the Boost issue here. We need to test it to verify, which means installing Boost 1.70.0 ... which might take a bit so please have patience!

gr-blocks/lib/socket_pdu_impl.cc
```diff
@@ -29,6 +29,12 @@
 #include <gnuradio/io_signature.h>
 #include <gnuradio/blocks/pdu.h>
 
+#if BOOST_VERSION >= 107000
+#define GET_IO_SERVICE(s) ((boost::asio::io_context&)(s)->get_executor().context())
+#else
+#define GET_IO_SERVICE(s) ((s)->get_io_service())
+#endif
```
SoapGentoo: Recommended in https://github.com/boostorg/asio/commit/59066d80b26e1d5b83b60d127ee17948d9ae9702#commitcomment-33184421
The 1.70 cast is potentially unsafe

michaelld:
As noted already, the cast is potentially unsafe. We also have the issue that `boost::asio::io_service` is being deprecated starting as of starting in 1.66.0, replaced directly by `boost::asio::io_context`. So ... let's just get these all fixed up together shall we?! There's a better way to implement the various parts here that will build cleanly on the current Boost as well as prior ones, which does not involve the potentially unsafe cast. Thus, thank you for the prod to get this fixed; I'm closing this ticket & will submit another PR with the more comprehensive fix. I'll make sure this PR is noted for reference.

PR #2454, opened 2019-04-18
Add AGC support for USRP Source block.

### LINKED ISSUE #2442: gr-uhd AGC support
UHD currently has the ability to turn AGC on/off via this function: https://files.ettus.com/manual/classuhd_1_1usrp_1_1multi__usrp.html#abdab1f6c3775a9071b15c9805f866486

I am hoping that this functionality gets exposed in gr-uhd, so that I can toggle AGC on/off using GNU Radio.

dwd-pete:
Fixes #2442

### DIFF
--- a/gr-uhd/grc/gen_uhd_usrp_blocks.py
+++ b/gr-uhd/grc/gen_uhd_usrp_blocks.py
@@ -170,11 +170,26 @@ templates:
         % for n in range(max_nchan):
         ${'%'} if context.get('nchan')() > ${n}:
         self.${'$'}{id}.set_center_freq(${'$'}{${'center_freq' + str(n)}}, ${n})
+        % if sourk == 'source':
+        ${'%'} if context.get('rx_agc${n}')() == 'Enabled':
+        self.${'$'}{id}.set_rx_agc(True, ${n})
+        ${'%'} elif context.get('rx_agc${n}')() == 'Disabled':
+        self.${'$'}{id}.set_rx_agc(False, ${n})
+        ${'%'} endif
+        ${'%'} if context.get('rx_agc${n}')() != 'Enabled':
         ${'%'} if bool(eval(context.get('norm_gain' + '${n}')())):
         self.${'$'}{id}.set_normalized_gain(${'$'}{${'gain' + str(n)}}, ${n})
         ${'%'} else:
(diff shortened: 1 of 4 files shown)


dwd-pete:
Full disclosure: I don't have a USRP to test with, but the generated GRC blocks look sane as does the generated top_block.py for each possible set of gain options.

dkozel:
@dwd-pete Thanks for posting this up. I've quickly read over and everything makes logical sense and I can't think of edge cases you've missed. I do have a B2xx USRP so can do a functional test early next week if not sooner.

marcusmueller:
Hi!

I agree with @dkozel, this looks nice, @dwd-pete! Seeing this is not just a handful lines of code, and you're a first-time contributor: Do we already have a copyright assignment from you (or the organization you work for)? If not, @bhilburn can set you up with one.

dwd-pete:
I currently do not have a CLA in place, and totally forgot about this. That said, I have gotten the ball rolling between my organization and @bhilburn  to get the CLA taken care of.

dwd-pete:
CLA should now be in place.

gr-uhd/grc/gen_uhd_usrp_blocks.py
```diff
@@ -202,7 +217,14 @@
     -   set_samp_rate(${'$'}{samp_rate})
     % for n in range(max_nchan):
     -   set_center_freq(${'$'}{${'center_freq' + str(n)}}, ${n})
+    % if sourk == 'source':
+    -   ${'$'}{'set_rx_agc(True, ${n})' if context.get('rx_agc${n}')() == 'Enabled' else ''}
```
mbr0wn: I wonder if we can't get those True/False values directly from the rx_agc var, but this code is touchy and anything that works is fine.

gr-uhd/lib/usrp_source_impl.cc
```diff
@@ -159,6 +159,17 @@ namespace gr {
       return _dev->set_rx_gain(gain, name, chan);
     }
 
+    void
+    usrp_source_impl::set_rx_agc(const bool enable, size_t chan)
+    {
+#if UHD_VERSION >= 30900
```
mbr0wn: It's 30803, actually.

mbr0wn:
Merged manually.

PR #2459, opened 2019-04-21
modtool: Set C/C++ standard to 11 for OOT's.

drmpeg:
Allows OOT's to compile with gcc 5.4 (Ubuntu 16.04).


marcusmueller:
@michaelld 
It does, I think, because `std::string` has a different interface pre-C++11 and after (maybe only on GCC and clang, but those are the main compilers, anyway)

drmpeg:
The 3.8 Gnu Radio API requires C++11. rpcmanager.h uses unique_ptr and basic_block.h uses to_string().

PR #2465, opened 2019-04-26
cmake: Enable CMAKE_EXPORT_COMPILE_COMMANDS

johschmitz:
When CMAKE_EXPORT_COMPILE_COMMANDS is ON, a file compile_commands.json
will be created in the build directory that enables code completion
engines to find all include directories and provide a very comprehensive
auto completion experience without any manual IDE configuration.

A nice example for this is the VS Code extension cpptools that is
developed by Microsoft. Of course this is not limited to VS Code and
works with other engines such as YouCompleteMe as well.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -65,6 +65,9 @@ set(CLANG_MIN_VERSION "3.4.0")
 set(APPLECLANG_MIN_VERSION "500")
 set(MSVC_MIN_VERSION "1800")
 
+# Enable generation of compile_commands.json for code completion engines
+set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
+
 ########################################################################
 # Configure CMake policies
 ########################################################################
--- a/gr-utils/python/modtool/templates/gr-newmod/CMakeLists.txt
+++ b/gr-utils/python/modtool/templates/gr-newmod/CMakeLists.txt
@@ -49,6 +49,9 @@ set(VERSION_INFO_MAINT_VERSION git)
 
 cmake_policy(SET CMP0011 NEW)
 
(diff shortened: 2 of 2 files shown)


johschmitz:
See the relevant sections about compile_commands.json in
https://github.com/Microsoft/vscode-cpptools/blob/master/Documentation/Getting%20started%20with%20IntelliSense%20configuration.md
and
https://github.com/Valloric/YouCompleteMe#c-family-semantic-completion-engine-usage
for more information.

JayKickliter:
I always try to remember to run `cmake -DCMAKE_EXPORT_COMPILE_COMMANDS=ON ...`, but often forget. So even doesn't _need_ to be in `CMakeLists.txt`, I'd appreciate if it was. 

In addition to what you posted, `compile_commands.json` makes it very easy to run `cppcheck` and `clang-tidy`. I also use it with Emacs `irony-mode` for code completion and live type checking.

noc0lour:
In general this makes sense, but I don't know if we want to enable this for all builds. Maybe for the `Debug` build type?

johschmitz:
It just creates one file in the build folder that will not be installed. Are you sure that is problematic for the release build? Also I think people need to develop with the optimized release build enabled due to performance reasons sometimes as well.

For me it's more a question of if we want to enable this at all or not. I don't see the advantage in enabling it only for some build types.

michaelld:
So long as the generated file(s) aren't actually installed -- meaning that they are truly just for debugging purposes && not end-users who don't care about such things -- then I see no harm in this PR && I could see some benefit for developers too. We (a few GR devs) should try this out & verify what it does (creates this specific file) and doesn't do (install said file or do anything else).

PR #2467, opened 2019-04-29
Make phase parameter of analog sig source available in grc

dl1ksv:
The analog signal source implements an initial phase offset of the signal. But this offset is not available in grc.
This pr makes the phase parameter available in grc.


mormj:
These changes look good to me.  The only thing I see that might be helpful is calling out the units of the phase somewhere (label of the parameter).  It's not in the documentation in sig_source.h either, but it would be nice to explicitly say somewhere that it is in radians - probably there in the \param so it shows up in the documentation tab.

mbr0wn:
Merged manually.

Issue #2470, opened 2019-05-07
some errors in gr-digital


dl1ksv:
Loading example_corr_est.grc from the gr-digital examples in grc leads to a lot of errors.

ERROR:gnuradio.grc.core.FlowGraph:Failed to evaluate variable block hdr_const
Traceback (most recent call last):
  File "/usr/local/gnuradio/lib/python3.7/dist-packages/gnuradio/grc/core/FlowGraph.py", line 226, in renew_namespace
    value = eval(variable_block.value, namespace, variable_block.namespace)
  File "<string>", line 1, in <module>
  File "/usr/local/gnuradio/lib/python3.7/dist-packages/gnuradio/digital/digital_swig2.py", line 524, in make
    return _digital_swig2.constellation_calcdist_make(constell, pre_diff_code, rotational_symmetry, dimensionality, normalize_points)
TypeError: in method 'constellation_calcdist_make', argument 1 of type 'std::vector< gr_complex,std::allocator< gr_complex > >'
ERROR:gnuradio.grc.core.FlowGraph:Failed to evaluate variable block rrc_taps
Traceback (most recent call last):
  File "/usr/local/gnuradio/lib/python3.7/dist-packages/gnuradio/grc/core/FlowGraph.py", line 226, in renew_namespace
    value = eval(variable_block.value, namespace, variable_block.namespace)
  File "<string>", line 1, in <module>
  File "/usr/local/gnuradio/lib/python3.7/dist-packages/gnuradio/filter/filter_swig.py", line 253, in root_raised_cosine
    return _filter_swig.firdes_root_raised_cosine(gain, sampling_freq, symbol_rate, alpha, ntaps)
TypeError: in method 'firdes_root_raised_cosine', argument 1 of type 'double'
ERROR:gnuradio.grc.core.FlowGraph:Failed to evaluate variable block rx_psf_taps1
Traceback (most recent call last):
  File "/usr/local/gnuradio/lib/python3.7/dist-packages/gnuradio/grc/core/FlowGraph.py", line 226, in renew_namespace
    value = eval(variable_block.value, namespace, variable_block.namespace)
  File "<string>", line 1, in <module>
  File "/usr/local/gnuradio/lib/python3.7/dist-packages/gnuradio/filter/filter_swig.py", line 253, in root_raised_cosine
    return _filter_swig.firdes_root_raised_cosine(gain, sampling_freq, symbol_rate, alpha, ntaps)
[...]

PR #2471, opened 2019-05-07
Moving enum from digital.corr_est_cc to digital

dl1ksv:
The generated python code for corr_est_cc fails with

AttributeError: 'function' object has no attribute 'THRESHOLD_ABSOLUTE'

Moving the enum from corr_est_cc to the digital namespace fixes this problem.

See #2470 
With this patch at least the generated code runs.

### DIFF
--- a/gr-digital/grc/digital_corr_est_cc.block.yml
+++ b/gr-digital/grc/digital_corr_est_cc.block.yml
@@ -19,7 +19,7 @@ parameters:
 -   id: threshold_method
     label: Threshold Method
     dtype: enum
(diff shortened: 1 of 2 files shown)


mormj:
I'm guessing that this could also be fixed by putting the enum in the swig somehow to keep it within the class, but I see other enums at the top level of the digital namespace, so it's probably ok.  Perhaps since it's at the top level, the enums should be prepended with CORR_ since they are specific to this block

mbr0wn:
I'm going to resubmit this, since there is no CI results.

mbr0wn:
Superseded by  #2561.

PR #2475, opened 2019-05-11
runtime: prefs: parse user config even if sys conf dir isn't a directory

marcusmueller:
This comes from helping a user on the mailing list (hi there, Gary),
who's been having severe problems with the Windows audio source; told
him to configure GNU Radio to use portaudio instead. He did - but GR
ignored his config file. Turned out that the hardcoded path to the
system config dir (set at built) doesn't match how it was installed on
Windows, and that prefs don't even try the user confs if they can't find
that directory. Fixed that.

### DIFF
--- a/gnuradio-runtime/lib/prefs.cc
+++ b/gnuradio-runtime/lib/prefs.cc
@@ -67,16 +67,15 @@ namespace gr {
     std::vector<std::string> fnames;
 
     fs::path dir = prefsdir();
-    if(!fs::is_directory(dir))
-      return fnames;
-
-    fs::directory_iterator diritr(dir);
-    while(diritr != fs::directory_iterator()) {
-      fs::path p = *diritr++;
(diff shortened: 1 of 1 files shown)


gnuradio-runtime/lib/prefs.cc
```diff
@@ -67,16 +67,15 @@ namespace gr {
...
-      if(p.extension() == ".conf")
-        fnames.push_back(p.string());
+    if(fs::is_directory(dir)) {
+      fs::directory_iterator diritr(dir);
+      while(diritr != fs::directory_iterator()) {
+        fs::path p = *diritr++;
+        if(p.extension() == ".conf")
+          fnames.push_back(p.string());
+      }
+      std::sort(fnames.begin(), fnames.end());
```
mbr0wn: Both in the old and new code I'm not entirely sure what this sort does, but it's not an issue with this PR.

PR #2479, opened 2019-05-14
cmake: check for python module click as dependency of modtool

### LINKED ISSUE #2292: modtool requires click dependencies
The new modtool version requires click as a dependency. On Ubuntu 18.04, it was enough to run 
`apt install python3-click python3-click-plugins`

noc0lour:
Previously cmake did not check if click is installed as it is a runtime
requirement. If CMAKE_CROSSCOMPILING is set the module checking is
skipped.

Closes #2292

### DIFF
--- a/gr-utils/python/modtool/CMakeLists.txt
+++ b/gr-utils/python/modtool/CMakeLists.txt
@@ -19,6 +19,18 @@
 
 include(GrPython)
 
+GR_PYTHON_CHECK_MODULE_RAW(
+  "click"
+  "import click; import click_plugins"
+  CLICK_FOUND
+  )
+
+if(NOT CMAKE_CROSSCOMPILING)
+  if(NOT CLICK_FOUND)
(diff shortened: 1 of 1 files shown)


drmpeg:
Should it check for click-plugins too?

noc0lour:
good point

noc0lour:
I'll add click to all worker images to make our CI happy.

mbr0wn:
Can you re-run this? I noticed you updated buildbot, so this should work now.

noc0lour:
Thanks to @mbr0wn  for adding click dependencies to the CI worker images. After an hour or so I'll kick the builders here to rebuild the code since then the new images should be up in the docker cloud.

noc0lour:
It appears Fedora 29 does not contain python2-click-plugins and Testing Python2 on F29 makes little sense anyway, so the build job for python2 is removed from Fedora 29.

PR #2483, opened 2019-05-14
Selector Block replacement 2460

### LINKED ISSUE #2460: Implement blks2 Selector replacement
We've deprecated and finally removed `blks2_selector`, and we're not fixing its bugs anymore.

Thus, we need a replacement. Typically, a simple copy block that copies (and consumes) from either input. Shouldn't be too hard.

Also, needs a 3.7 backport.

mormj:
Adds c++ implementation of selector block based on copy block and a limited set of QA tests

fixes #2460

### DIFF
--- a/gr-blocks/lib/selector_impl.cc
+++ b/gr-blocks/lib/selector_impl.cc
@@ -1,0 +1,143 @@
+/* -*- c++ -*- */
+/*
+ * Copyright 2019 Free Software Foundation, Inc.
+ *
+ * This file is part of GNU Radio
+ *
+ * GNU Radio is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 3, or (at your option)
+ * any later version.
+ *
+ * GNU Radio is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with GNU Radio; see the file COPYING.  If not, write to
+ * the Free Software Foundation, Inc., 51 Franklin Street,
+ * Boston, MA 02110-1301, USA.
+ */
+
+#ifdef HAVE_CONFIG_H
+#include "config.h"
+#endif
+
+#include "selector_impl.h"
+#include <gnuradio/io_signature.h>
+#include <string.h>
+#include <stdexcept>
+
+namespace gr {
+namespace blocks {
+
+selector::sptr
+selector::make(size_t itemsize, unsigned int input_index, unsigned int output_index)
+{
+    return gnuradio::get_initial_sptr(
+        new selector_impl(itemsize, input_index, output_index));
+}
+
+selector_impl::selector_impl(size_t itemsize,
+                             unsigned int input_index,
+                             unsigned int output_index)
+    : block("selector",
+            io_signature::make(1, -1, itemsize),
+            io_signature::make(1, -1, itemsize)),
+      d_itemsize(itemsize),
+      d_enabled(true),
+      d_input_index(input_index),
+      d_output_index(output_index),
+      d_num_inputs(0),
+      d_num_outputs(0)
+{
+    message_port_register_in(pmt::mp("en"));
+    set_msg_handler(pmt::mp("en"), [this](pmt::pmt_t msg) { this->handle_enable(msg); });
+
+    // TODO: add message ports for input_index and output_index
+}
+
+selector_impl::~selector_impl() {}
+
+void selector_impl::set_input_index(unsigned int input_index)
+{
+    gr::thread::scoped_lock l(d_mutex);
+    if (input_index < d_num_inputs)
+        d_input_index = input_index;
+    else
+        throw std::out_of_range("input_index must be < ninputs");
+}
+
+void selector_impl::set_output_index(unsigned int output_index)
+{
+    gr::thread::scoped_lock l(d_mutex);
+    if (output_index < d_num_outputs)
+        d_output_index = output_index;
+    else
+        throw std::out_of_range("output_index must be < noutputs");
+}
+
+void selector_impl::handle_enable(pmt::pmt_t msg)
+{
+    if (pmt::is_bool(msg)) {
+        bool en = pmt::to_bool(msg);
+        gr::thread::scoped_lock l(d_mutex);
+        d_enabled = en;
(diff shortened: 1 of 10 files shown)


777arc:
set_input_index and set_output_index need to have their callbacks added to the yml file, so that they can be changed while the flowgraph is running (e.g. using QT GUI Chooser).  Other than that, seems good!

mormj:
@777arc thanks for the feedback.  Added the inputs to the yaml and squashed the commits to clean up the history

gr-blocks/lib/selector_impl.cc
```diff
@@ -0,0 +1,165 @@
+/* -*- c++ -*- */
+/*
+ * Copyright 2006,2009,2013 Free Software Foundation, Inc.
```
mbr0wn:
```suggestion
 * Copyright 2019 Free Software Foundation, Inc.
```

gr-blocks/python/blocks/qa_selector.py
```diff
@@ -0,0 +1,223 @@
...
+            self.tb.connect((op,jj),dst[jj])
+
+        
+        self.tb.run()
+
+        dst_data = dst[output_index].data()
+        self.assertEqual(expected_result, dst_data)
+
+
+    # These tests cannot be run as set_index can only be called after check_topology is called
```
mbr0wn: You could lock and unlock `self.tb` if you want  to run these

gr-blocks/python/blocks/qa_selector.py
```diff
@@ -0,0 +1,223 @@
...
+        
+        self.tb.run()
+
+        expected_result = [float(input_index)+1]*10*veclen
+        dst_data = list(dst[output_index].data())
+
+        self.assertEqual(expected_result, dst_data)
+
+if __name__ == '__main__':
+    gr_unittest.run(test_copy, "test_copy.xml")
```
mbr0wn:
```suggestion
    gr_unittest.run(test_copy)
```

gr-blocks/lib/selector_impl.cc
```diff
@@ -0,0 +1,165 @@
...
+      if(d_enabled) {
+
+        
+
+        memcpy(out[d_output_index], in[d_input_index], noutput_items*d_itemsize);
+
+        produce(d_output_index,noutput_items);
+        for (int i=0; i<noutputs; i++) {
+          if (i != d_output_index)
+            produce(i,0);
```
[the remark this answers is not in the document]
mormj: I'm not sure either - was going off the documentation:
  172      * WORK_CALLED_PRODUCE is used where not all outputs produce the
  173      * same number of items. general_work must call produce() for each
  174      * output to indicate the numer of items actually produced.

This made me think that produce(i,0) needed to be explicitly called, or at least just to be safe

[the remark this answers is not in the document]
mbr0wn: After looking at the code, I think it's not necessary. It just adds some locking load to update write pointers.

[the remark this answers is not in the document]
mbr0wn: However, this is unusual. I would advocate removing it, but add a comment here why there's only one produce() call on the active output.

gr-blocks/lib/selector_impl.cc
```diff
@@ -0,0 +1,141 @@
...
+                                gr_vector_int& ninput_items,
+                                gr_vector_const_void_star& input_items,
+                                gr_vector_void_star& output_items)
+{
+    const uint8_t** in = (const uint8_t**)&input_items[0];
+    uint8_t** out = (uint8_t**)&output_items[0];
+
+    gr::thread::scoped_lock l(d_mutex);
+    if (d_enabled) {
+        memcpy(out[d_output_index], in[d_input_index], noutput_items * d_itemsize);
```
mbr0wn: I would still like to see `std::copy` here, but I won't make a big stink if you don't.

mbr0wn: See also my previous comment: Similar to 2826fd0, I would strongly suggest moving to std::copy. It can be faster in some cases, but it's also more safe and generally more C++-y.

gr-blocks/lib/selector_impl.cc
```diff
@@ -0,0 +1,165 @@
...
+      else
+        throw std::out_of_range("input_index must be < ninputs");
+    }
+
+    void
+    selector_impl::handle_enable(pmt::pmt_t msg)
+    {
+      if(pmt::is_bool(msg)) {
+        bool en = pmt::to_bool(msg);
+        gr::thread::scoped_lock l(d_mutex);
```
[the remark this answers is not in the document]
mbr0wn: I still don't think we need the mutex. If we really care about synchronization issues, then we can use a `std::atomic_bool`, which is guaranteed lock-free, but I don't see a case where one thread calling `set_enable()` could make an output occur that is not what we want.

gr-blocks/lib/selector_impl.cc
```diff
@@ -0,0 +1,141 @@
...
+
+int selector_impl::general_work(int noutput_items,
+                                gr_vector_int& ninput_items,
+                                gr_vector_const_void_star& input_items,
+                                gr_vector_void_star& output_items)
+{
+    const uint8_t** in = (const uint8_t**)&input_items[0];
+    uint8_t** out = (uint8_t**)&output_items[0];
+
+    gr::thread::scoped_lock l(d_mutex);
```
mbr0wn: See my previous comment about the mutex. `work()` is in the hot loop, and this gives us some safety against... what exactly? whereas it adds locking load.

gr-blocks/lib/selector_impl.cc
mormj: Sorry I blew through the recommendation to switch to std::copy before - old habits die hard with memcpy.   Just pushed through the change

gr-blocks/lib/selector_impl.cc
mormj: Seems like @marcusmueller was fairly emphatic about the mutex - if it's safer to leave it out I'll delete those lines

mbr0wn: @marcusmueller Has the last word here. It's OK. It's safer to *not* leave it out, my argument is just that we don't get anything by adding that safety, and I don't think anyone can show my an app that works with it and not without it.

gr-blocks/lib/selector_impl.cc
mbr0wn: Thanks! And `std::copy` is still new. I haven't used it a lot myself.

mbr0wn:
@mormj This still needs a rebase.

mbr0wn:
Fantastic  work @mormj .

Issue #2055, opened 2018-09-30
Tag propagation doesn't work correctly when relative rate changes in runtime


ptrkrysik:
The 'set_relative_rate' makes misleading impression that GNU Radio blocks have support for runtime changes of processing block's relative rate (number of output samples / number of input samples).

However offset of tags is incorrectly computed from the moment when 'set_relative_rate' function changes the relative rate for the first time in the runtime. The reason for that is incorrect computation of output tag offset in 'propagate_tags' function:
https://github.com/gnuradio/gnuradio/blob/007923c64d2505b37dffb8e11cb1e3ecf4e489eb/gnuradio-runtime/lib/block_executor.cc#L140
https://github.com/gnuradio/gnuradio/blob/007923c64d2505b37dffb8e11cb1e3ecf4e489eb/gnuradio-runtime/lib/block_executor.cc#L149
https://github.com/gnuradio/gnuradio/blob/007923c64d2505b37dffb8e11cb1e3ecf4e489eb/gnuradio-runtime/lib/block_executor.cc#L183
https://github.com/gnuradio/gnuradio/blob/007923c64d2505b37dffb8e11cb1e3ecf4e489eb/gnuradio-runtime/lib/block_executor.cc#L191

Tag offset of the tag at the output is computed using this equation:
y = a*x
where:
* 'a' is relative rate
* 'x' is the offset of the tag at the input
* 'y' is the offset of the tag at the output

This assumes that from the beginning relative rate was equal to 'a', which isn't true after the relative rate is changed.

Why this is wrong? Look at the images below what happens on increasing and decreasing relative rate.
![image](https://cdn.rawgit.com/ptrkrysik/e5a9a0a8d4d8c984f3ece846884a75cd/raw/74d80a005fa4aadbf4d4e27bde520ebea7aad5e6/increasing_relative_rate.svg)
![image](https://cdn.rawgit.com/ptrkrysik/5316f2c2d75657673b05676b0f4164cb/raw/364b72cdfb31b0a7dce2ee7468070dc6b7ee7487/resampling_decreasing.svg)

To fix this the information when the relative rate was changed (a pair of input sample number 'n' and output samples number 'm') can be used:
y = a*(x-n)+m

I don't know yet how to pass the information about '(n,m)' pair to the 'set_relative_rate' fuction.
It would be perfect if it could get it from 'nitems_read(...)' and 'nitems_written(...)' functions.

[...]

PR #2484, opened 2019-05-15
Add support for changing relative rate during runtime and calculate correct offsets

noc0lour:
This works on solving #2055 but it's definitely not done yet.


mbr0wn:
Since this is both a draft and needs rebase, and hasn't had some activity for a while, I'll close it. I do hope to see changes like this at some point.

PR #2487, opened 2019-05-15
cmake: Cleanup SO versioning

noc0lour:
Use reasonable SO versioning as it is already done in downstream Debian


noc0lour:
afaik GR compatibilit is conveyed in the MINOR version number, since that guarantees ABI compability and the API_COMPAT guarantees API compability. I also basically copied this patch from debian downstream.

michaelld:
> afaik GR compatibilit is conveyed in the MINOR version number, since that guarantees ABI compability and the API_COMPAT guarantees API compability. I also basically copied this patch from debian downstream.

Maybe so; maybe not. I do wish "we" in GR / Volk-land had a well-defined concept of API and ABI compatibility versioning. I know of no "formal" document that states what you write; it's all knowledge we're just supposed to have; not very "professional" IMHO.

I have no objections to this change. I'll patch in MacPorts if/as required anyway to get hot-swapping between the current release & devel for a given version (3.7 or 3.8) correct anyway; just takes more testing on my end to verify hot-swapping works.

noc0lour:
Hm, @marcusmueller had some semantic versioning document to follow, but I can't find it anywhere.

noc0lour:
Also currently on Linux we create libraries with the naming:
```
libgnuradio-$module-X.Y.Z.patch.so.0.0.0
```
and in future those libraries will be
```
libgnuradio-$module.so.X.Y.Z
```

marcusmueller:
Wow, super happy about this discussion!

>Maybe so; maybe not. I do wish "we" in GR / Volk-land had a well-defined concept of API and ABI compatibility versioning. I know of no "formal" document that states what you write; it's all knowledge we're just supposed to have; not very "professional" IMHO.

@michaelld is absolutely right here. 

(avoiding ambiguity by not using the naming you're using above)

     {W}.{X}.{Y}.{Z}

# State of Versioning on 3.7

So, here's what I have been living on `maint-3.7`:

* W=3 is 3 for historical reasons. We'll probably only change that when we *really* change the way GNU Radio is used – for example, if we replace the semantic that you have a work function and inter-work executed message handlers. By our standards of today, the progression of 3.6 -> 3.7 would probably have been 3.6 -> 4.0, as it introduced so much new semantics.
* W.X=3.7 is the large-scale functionality version; API. It's basically "If you know how to use any 3.7 version, then 3.7.X isn't going to be a big surprise to you". I'm a bit careful about actually saying "API", because it's really not – there's enough code written for 3.7.1 that probably won't compile with 3.7.13; while fixing a few things, I'm pretty sure we broke a bit of API here and there within the 3.7 series. Clearly, maintaining 100% API compatibility over 10 years is unlikely¹.
* W.X.Y=3.7.y is the C++ ABI version. You don't have to recompile if you're updating from 3.7.13.4 to 3.7.13.5.
* W.X.Y.Z=3.7.y.z is the patch level and just used to "count" our releases.

¹ This is a description of *is*, not of *how we'll handle this for 3.8*. Thing is simple: through our current development model, the day we release 3.8 is the day that forward development happens towards 3.9 and only fixes end up in 3.8; I think we can actually guarantee a stable API for `maint-3.8`.

# State of versioning for `maint-3.8`

* W.X=3.8 is the public API version. We'll only accept patches that don't change the public API.²
* W.X.Y=3.8.y is the ABI version. This includes C/C++, Python bindings and the interfaces of GRC blocks. In other words, if you don't want to increase y, don't remove any Python or C++ parameters from functions, don't add any mandatory parameters, and don't change a block YAML in a way that breaks existing GRC flowgraphs' generation (unless it was broken prior, of course). 
* W.X.Y.Z=3.8.y.z patch level

²It's extremely hard to tell what "public" means, since we've –justifiably so– didn't care too much about that. Especially, I don't think we should be too strict with that when it comes to Pythonworld things. It will probably be a case-by-case issue when we fix individual blocks whether we allow API changes.

marcusmueller:
In other words, since for the linker, only ABI compatibility counts, it's appropriate to have

    libgnuradio-runtime.so.3.7.13

or 

    libgnuradio-runtime.so.3.8.0

marcusmueller:
@michaelld  I think this is so important that we should document it in-tree. What would you think: Readme or CHANGELOG, or separate file VERSIONING?

cmake/Modules/GrMiscUtils.cmake
```diff
@@ -72,6 +70,7 @@ endmacro(GR_ADD_CXX_COMPILER_FLAG_IF_AVAILABLE)
 function(GR_LIBRARY_FOO target)
     #set additional target properties
     set_target_properties(${target} PROPERTIES SOVERSION ${LIBVER})
+    set_target_properties(${target} PROPERTIES VERSION ${VERSION})
```
noc0lour: This will actually set versioning of our shared libraries like so:
```
╰─ ls -a ~/.usrlocal/lib/libgnuradio-analog.so*                                      ✔  15:43:04
/home/andrej/.usrlocal/lib/libgnuradio-analog.so
/home/andrej/.usrlocal/lib/libgnuradio-analog.so.3.8git
/home/andrej/.usrlocal/lib/libgnuradio-analog.so.3.8tech-preview-365-gd66bc948
```

```
lrwxrwxrwx 1 andrej users      27 May 16 16:15 /home/andrej/.usrlocal/lib/libgnuradio-analog.so -> libgnuradio-analog.so.3.8.0
lrwxrwxrwx 1 andrej users      29 May 16 16:15 /home/andrej/.usrlocal/lib/libgnuradio-analog.so.3.8.0 -> libgnuradio-analog.so.3.8.0.0
```

mbr0wn: FYI, in UHD, we set both these to the ABI version. I have no clue if that's better or worse, but it's what Maitland suggested.

mbr0wn:
@marcusmueller We might also want to rename the CMake variables.

michaelld:
@marcusmueller Thanks for the detailed explanation. I like the idea of a file called VERSIONING, so-as to make the intent of it relatively self explanatory.

michaelld:
It's clear (to me) that what we've been aiming for in the versioning is "general" compatibility, not explicit / guaranteed compatibility. I'd guess that code designed for 3.7.0 generally probably won't build with 3.7.5 let alone 3.7.14 (hence no explicit API or ABI compatibility), but it shouldn't be difficult to fix it to work since the base gnuradio-runtime won't have changed a lot, primarily bug fixes and slight API augmentations rather than major API changes (though there might be some of those, which are allowed by the change from 3.7.X to 3.7.(X+1)). Code designed for 3.7.0 will probably be a bit of work to get working for 3.8 --- but, not as much of a pain as it was moving from 3.6 to 3.7. I'm not even sure we really followed the versioning for 3.7.X.Y for when -any- API change for it's then current was put in place, which would then require bumping the version to 3.7.(X+1).0 ... but I don't care enough to go back and find out examples. I think if we add this VERSIONING, then we just need to make sure we're abiding by it -- in the future.

michaelld:
> @marcusmueller We might also want to rename the CMake variables.

I agree with @mbr0wn's thought here: just use different CMake variables.

noc0lour:
Well, we shouldn't care about mistakes made in the past but follow rules we specify in the future.

michaelld:
> Well, we shouldn't care about mistakes made in the past but follow rules we specify in the future.

I agree with your intent here, @noc0lour : look to the future & the greatness it holds! That said, we cannot forget the past; we should always care about the past and the successes and mistakes made in it; after all, it is what brought us here to now (one way or the other)!

michaelld:
To the best of our abilities we should abide by the "rules" and "guidelines" we've set for ourselves ... in this case, with any "GREP"s as well as if we add a "VERSIONING" file and/or anything else that defines how our organization works (handles commits, merge requests, Code of Conduct, whatever).

Where there are conflicts, we should discuss them to hopefully form group consensus on how to proceed; and if no consensus if formed, then someone of appropriate authority must make the executive decision & everyone else must honor the process that lead to the decision even if they don't agree with the decision.

These are one part of how a professional organization should work IMHO.

mbr0wn:
Looks like we all agree on what we want, now we just need to clean up the deets and the docs. In fact, we should merge this ASAP, even w/o the versioning, because that reduces patches elsewhere.

mbr0wn:
https://github.com/EttusResearch/uhd/blob/master/host/docs/uhd_semvar.dox

Permission to help yourselves :)

marcusmueller:
@mbr0wn this was only meant to be merged later on, after we've merged a versioning document (that would've been the right order; documenting what we're about to structurally do, and then doing it). Still, thanks for merging, as that means I'll have to quickly come up with a versioning document-

mbr0wn:
@marcusmueller FYI @noc0lour is already using this patch to package debs, so I thought it wise to speed it up.

PR #2488, opened 2019-05-16
Issue #1974:  Unnecessary delay in selective fading model

### LINKED ISSUE #1974: selective_fading_model has off-by-one error; introduces 1-sample delay
When the selective_fading_model block [calls set_history](https://github.com/gnuradio/gnuradio/blob/aa2bc60b2ba07133115dd8e88084b5900a5290be/gr-channels/lib/selective_fading_model_impl.cc#L66), I believe it's using a value that's one greater than it should be.

As I understand it, `set_history(1)` means "no history" (i.e. in `work()`, `input_items[0][0]` refers to the first sample). `set_history(2)` is used for "one sample of history" and so on.

When `ntaps==1`, the user is effectively asking for a FIR filter with only one tap -- i.e. a multiplication against a (slowly-changing, in the case of this model) constant. However, in this case, the fading model will use `set_history(2)` which requests one sample of history. As a result, the model introduces a delay.

For most users, I suppose this isn't an issue - the fading model generally is used for separating the transmit-side and receive-side logic of a given design. In my case, however, I'm using tags (which are ending up on the wrong sample due to not being delayed as well) and I'm allowing the receiver to access the transmitter's reference clock (so I need to add a `Delay: 1` block to that flow for my prototype to work).

mormj:
Backport from master branch
Reduce set_history from ntaps+1 to ntaps as an extra delay is introduced

Fixes #1974

### DIFF
--- a/gr-channels/lib/selective_fading_model_impl.cc
+++ b/gr-channels/lib/selective_fading_model_impl.cc
@@ -63,7 +63,7 @@ namespace gr {
 
         // set up tap history
         if(ntaps < 1){ throw std::runtime_error("ntaps must be >= 1"); }
-        set_history(1+ntaps);
+        set_history(ntaps);
         d_taps.resize(ntaps, gr_complex(0,0));
     }
 


mbr0wn:
Merged manually

PR #2495, opened 2019-05-21
filter: continue transition to pyqtgraph and pyqt5

mormj:
continue the work of @ekigwana to get a stable version of filter_design_tool on 3.8 branch

Still missing functionality and issues


777arc:
The IIR option doesn't seem to be doing anything for me, no errors, but no updates in the output either.

mormj:
@777arc Interesting - this was working at some point - no trick I know to get it working.  Will debug on my end

mormj:
@777arc - should be good to go now.  Looks like the text in the box was changed from IIR(numpy) to IIR(scipy) which it was keying off of.

mbr0wn:
This seems to a big improvement.

Issue #2493, opened 2019-05-20
Analog AM Demod Block Silently Fails with Bad Passband


bhilburn:
Per the thread "free(): invalid pointer for AM demod block at Gnuradio 3.7.13.5" on the list, if a passband of `0` is used (thus making the passband from "zero to zero"), this block will silently fail with a `free()` error. A parameter check should be added to this block to catch this instance.

dl1ksv:
Same problem exists in the master branch and should fixed.

rear1019:
Closing, fixed both on maint-3.7 (pull request #2773) and on master (commit 27dd99e49a7d).

PR #2496, opened 2019-05-21
Introduce some minimal passband parameter checks

dl1ksv:
Add some checks to give meaningfull error messages , if parameters are wrong.
See #2493


mbr0wn:
Merged manually.

PR #2503, opened 2019-05-23
zeromq: fixes to allow building using CPPZMQ 4.3.1 as well as prior

michaelld:
ZMQ deprecated some prototypes of "recv" and "send" in some version prior to 4.3.1. The "recv" one has been removed in 4.3.1 and thus generates an error since no valid prototype exists in the way we're trying to call "recv". This fix updates our usage of both calls to work with the "new" and "old" ways, depending on the ZMQ version. Move the ZMQ inclusion and a macro determine which version to use into a common header.

I'm not 100% sure this is the correct solution, but it works (for me on OSX) and is relatively straight forward to understand.

### DIFF
--- a/gr-zeromq/lib/pull_msg_source_impl.h
+++ b/gr-zeromq/lib/pull_msg_source_impl.h
@@ -1,6 +1,6 @@
 /* -*- c++ -*- */
 /*
- * Copyright 2013,2014 Free Software Foundation, Inc.
+ * Copyright 2013,2014,2019 Free Software Foundation, Inc.
  *
  * This file is part of GNU Radio.
  *
@@ -24,7 +24,7 @@
 #define INCLUDED_ZEROMQ_PULL_MSG_SOURCE_IMPL_H
 
 #include <gnuradio/zeromq/pull_msg_source.h>
-#include <zmq.hpp>
+#include "zmq_common_impl.h"
 
 namespace gr {
   namespace zeromq {
--- a/gr-zeromq/lib/zmq_common_impl.h
+++ b/gr-zeromq/lib/zmq_common_impl.h
@@ -1,6 +1,6 @@
 /* -*- c++ -*- */
 /*
- * Copyright 2013,2014,2019 Free Software Foundation, Inc.
+ * Copyright 2019 Free Software Foundation, Inc.
  *
  * This file is part of GNU Radio.
  *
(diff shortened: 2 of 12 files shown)


gr-zeromq/lib/zmq_common_impl.h
```diff
@@ -0,0 +1,40 @@
...
+#include <zmq.hpp>
+
+#if defined (ZMQ_VERSION) && defined (ZMQ_MAKE_VERSION) && ZMQ_VERSION >= ZMQ_MAKE_VERSION(4, 3, 1)
+#define USE_NEW_ZMQ_SEND_RECV 1
+#else
+#define USE_NEW_ZMQ_SEND_RECV 0
+#endif
+
+#if USE_NEW_ZMQ_SEND_RECV
+#warning "ZMQ using new >= 4.3.1 send/recv!"
```
mbr0wn: Do we really want a warning in both cases? A warning indicates something bad, but really, we're just saying we're using the APIs like we're supposed to.

mbr0wn: I would advocate removing this entire section.

michaelld: Oops! That's cruft leftover from when I was testing. I'll remove it!

michaelld:
One thing to check is when ZMQ introduced the new APIs & marked the ones we're using as deprecated. If long enough ago & our required ZMQ version is newer than that, then we could switch over to the new one entirely. But ... I'm guessing this will not be the case. Just noting for completion that I didn't check this yet.

michaelld:
@mbr0wn I removed the cruft #warning from the common header. All should be good to go now.

michaelld:
Looks from the build logs that my fix here isn't quite correct: whatever the version of ZMQ is on Debian 10 is being mis-detected as >= 4.3.1 to use the "new" API, but in reality is < 4.3.1. Grrr ... How do I find out which version of ZMQ is installed on the buildbots?

Looks like I need to get some VMs going to allow me to check at a minimum building on various alternative OSs.

michaelld:
Found it on the buildbot itself:
{{{
-- Checking for module 'libzmq'
--   Found libzmq, version 4.3.1
-- Found ZEROMQ: /usr/lib/x86_64-linux-gnu/libzmq.so
}}}
Hmmm .... that's ... unexpected!

michaelld:
So .. is cppzmq a dependency of GR or not? That's where I found the issue in the first place, because GR failed to build & this was the most obvious solution. I'm changing the title to "WIP" since I need to do some more investigation to figure out what's what here!

michaelld:
Yes, `cppzmq` is required for `gr-zeromq` ... without this PR, a bunch of the `_impl` depend on that header. Not sure that's checked for during configuration ...

michaelld:
I just pushed fixes that should fix the buildbot issues. Let's let them run & if they succeed then I think this PR will be good to go.

michaelld:
The change is that the version of _CPPZMQ_ is what matters, not the version of _ZMQ_ ... regardless of what the `zmq.hpp` header uses.

michaelld:
OK so all of the builds now succeed, and all of the zeromq tests pass for all of the buildbots. I also see successes on my various OSX build bots. So, I'm moving this from "WIP" back to a normal PR for review.

gr-zeromq/lib/pull_msg_source_impl.h
```diff
@@ -24,7 +24,7 @@
 #define INCLUDED_ZEROMQ_PULL_MSG_SOURCE_IMPL_H
 
 #include <gnuradio/zeromq/pull_msg_source.h>
-#include <zmq.hpp>
+#include "zmq_common_impl.h"
```
noc0lour: Is including an external `_impl` header the right thing?

michaelld: This is a local `_impl` header file, created by the PR ...

michaelld: It, in turn, does `#include <zmq.hpp>` ...

PR #2505, opened 2019-05-23
Added a VERSIONING description

marcusmueller:
For future reference, we'll need make sure that there's a clear
understanding of how GNU Radio release versioning takes place, and what
the criteria are that go into choosing how to increment a version
number.

This document strives to achieve that.

### DIFF
--- a/VERSIONING
+++ b/VERSIONING
@@ -1,0 +1,16 @@
+# Versioning
+
+For the 3.8 Release and going on, the following versioning scheme is to be used. For 3.7 and earlier, other schemes apply.
+
+GNU Radio uses a four-number versioning:
+
+```
+A.B.C.D
+```
+
+* A is the paradigm version. This has been 3 for more than a decade, meaning that
+> GNU Radio is a block-based flow graph framework that uses classes to represent blocks whose main feature is a `work()` function taking input samples from quasi-circular buffers and putting results in quasi-circular output buffers.
+It's encouraged that in the future, relevant modifications and extensions of that model will lead to changes to the paradigm version number.
+* A.B is the public API version. We'll only accept patches that don't change the public API on a `maint-A.B` branch. GNU Radio users  need not change their code to stay compatible with any future A.B.x.x version (barring bugs). Forward-going development should happen on the A.(B+1) branch, once A.B is released.
+* A.B.C is the ABI version. This includes C/C++, Python bindings and the interfaces of GRC blocks. Programmers need not recompile their code if A.B.C stays unchanged.
+* A.B.C.D is the patch level.
--- a/VERSIONING
+++ b/VERSIONING
@@ -1,6 +1,12 @@
-# Versioning
+# Versioning Core idea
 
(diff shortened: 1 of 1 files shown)


VERSIONING
```diff
@@ -0,0 +1,116 @@
...
+
+```
+A.B.C.D
+```
+
+* A is the paradigm version. This has been 3 for more than a decade, meaning that
+> GNU Radio is a block-based flow graph framework that uses classes to represent blocks whose main feature is a `work()` function taking input samples from quasi-circular buffers and putting results in quasi-circular output buffers.
+It's encouraged that in the future, relevant modifications and extensions of that model will lead to changes to the paradigm version number.
+* A.B is the public API version. We'll only accept patches that don't change the public API on a `maint-A.B` branch. GNU Radio users  need not change their code to stay compatible with any future A.B.x.x version (barring bugs). Forward-going development should happen on the A.(B+1) branch, once A.B is released.
+* A.B.C is the ABI version. This includes C/C++, Python bindings and the interfaces of GRC blocks. Programmers need not recompile their code if A.B.C stays unchanged.
```
mbr0wn: Python ABI? This means we have to trust SWIG to maintain ABI?

marcusmueller: we need to trust GCC to maintain ABI, too. So yeah. If you're using the same SWIG, the same ABI should result.

This in essence means that we can't be changing swigged methods. Typically no problem, since that is the same as keeping C++ ABI.

mbr0wn: It also means we're trusting SWIG not to do funky stuff with ABI. However, who cares about the SWIG ABI? No one will link against it, except for us.
If we remove this requirement, then we don't have to run abi-dumber against SWIG .so files with their crazy symbol names. Much less work for us. And I don't see the downside.

michaelld: I don't think even we link against the SWIG .so ABI files ... they are not designed to be linked against, and I think on OSX one actually can't link against them since they don't contain the correct internal linkage to allow doing so.

VERSIONING
```diff
@@ -0,0 +1,116 @@
...
+5.  ABI version Y (w.x.Y.z) MUST be incremented if changes break ABI
+    compatibility with the previous release.
+
+6.  API version X (w.X.y.z) MUST be incremented if changes break public
+    API compatibility with the previous release. It MAY include ABI
+    and patch level changes. It MAY be incremented if substantial new
+    functionality or improvements are introduced within private code.
+    ABI and PATCH version MUST be reset to 0 when API version
+    is incremented. An API breakage is defined as the case where
+    recompiling software against UHD without modifications may yield
```
michaelld: `UHD` or `GNU Radio`?

mbr0wn: lol

VERSIONING
mbr0wn: The text doesn't actually call out SWIG, it only says people won't have to change their Python. Since there's no more complaints, that means we can leave this as is, but I remain unconvinced that SWIG ABI is something we should even care about.

mbr0wn:
Merged manually, removed some oopsies.

Issue #2515, opened 2019-05-28
CMake using absolute path for installation of runtime_swig-release


flynn378:
During cross compiling the absolute path is saved as part of the file created by cmake:
 runtime_swigTargets-release.cmake
GNU Radio will build, install and execute with these settings, but OOT modules will not build.  
Based on the bug report:
https://cmake.org/Bug/print_bug_page.php?bug_id=14367
Which is marked as not a bug, but explains the issue very well. 
gnuradio-runtime.swig/CMakeLists.txt uses an absolute path for the DESTINATION ( or something that gets turned into an absolute path) in the lines:
install(
  TARGETS runtime_swig
  EXPORT runtime_swig-export
  DESTINATION ${GR_PYTHON_DIR}/gnuradio/gr
  ) 

The DESTINATION should point to a relative path, based on the above bug report.

There is an email chain on the mailing list with more information about the issue.

michaelld:
Should be fixed now : https://github.com/gnuradio/gnuradio/pull/2502

michaelld:
This PR was merged today. It changes the runtime_swig `DESTINATION` to `${GR_CMAKE_DIR}`, which by default is a relative path.

flynn378:
I will test this tomorrow, I believe the actual error is a relative path needed for the 
install(
TARGETS ...
${GR_PYTHON_DIR}/gnuradio/gr
lines 62-65
Not the cmake install directory which was changed in merge request #2502.  I think I tried that change in the days of testing.  I could not find anything which provides the relative Python Path, It has been added to CMake, I think 3.13 or 3.14.
If there is a variable which is the relative Python Path it would be useful.  
I have tested the following changes and they work,  I am not sure it is the correct solution.
Based on the information from:
https://stackoverflow.com/questions/1242904/finding-python-site-packages-directory-with-cmake

I modified GrPython.cmake by adding a new variable, PYTHON_SITE:
########################################################################
# Sets the python relative installation directory PYTHON_SITE
########################################################################
if(NOT DEFINED PYTHON_SITE)
execute_process(
  COMMAND "${PYTHON_EXECUTABLE}" -c "if True:
    from distutils import sysconfig as sc
    print(sc.get_python_lib(prefix='', plat_specific=True))"
  OUTPUT_VARIABLE PYTHON_SITE
  OUTPUT_STRIP_TRAILING_WHITESPACE)
endif()

and then changed the target install line 65 in 
[...]

PR #2517, opened 2019-05-29
Add GR_PYTHON_RELATIVE to fix issue #2515 with runtime_swigTargets-re…

flynn378:
…lease.cmake

This commit fixes issue #2515.  A new CMake variable, GR_PYTHON_RELATIVE
is needed to properly create the CMake files used by OOTs when cross
compiling or using sysroot based build systems

see  https://cmake.org/Bug/print_bug_page.php?bug_id=14367  for more
information

### DIFF
--- a/cmake/Modules/GrPython.cmake
+++ b/cmake/Modules/GrPython.cmake
@@ -111,6 +111,19 @@ endif()
 file(TO_CMAKE_PATH ${GR_PYTHON_DIR} GR_PYTHON_DIR)
 
 ########################################################################
+# Sets the python relative installation directory GR_PYTHON_RELATIVE
+########################################################################
+if(NOT DEFINED GR_PYTHON_RELATIVE)
+execute_process(COMMAND "${PYTHON_EXECUTABLE}" -c "
+from distutils import sysconfig as sc
+print(sc.get_python_lib(prefix='', plat_specific=True))
+"
+  OUTPUT_VARIABLE GR_PYTHON_RELATIVE  OUTPUT_STRIP_TRAILING_WHITESPACE
+)
+endif()
+
+
+########################################################################
 # Create an always-built target with a unique name
 # Usage: GR_UNIQUE_TARGET(<description> <dependencies list>)
 ########################################################################
--- a/gnuradio-runtime/swig/CMakeLists.txt
+++ b/gnuradio-runtime/swig/CMakeLists.txt
@@ -62,7 +62,7 @@ GR_SWIG_MAKE(runtime_swig runtime_swig.i)
 install(
   TARGETS runtime_swig
   EXPORT runtime_swig-export
-  DESTINATION ${GR_PYTHON_DIR}/gnuradio/gr
+  DESTINATION ${GR_PYTHON_RELATIVE}/gnuradio/gr
   )
 
 include(GrPython)
@@ -74,7 +74,7 @@ GR_PYTHON_INSTALL(FILES ${CMAKE_CURRENT_BINARY_DIR}/runtime_swig.py
 install(EXPORT runtime_swig-export
   FILE runtime_swigTargets.cmake
   NAMESPACE gnuradio::
-  DESTINATION ${GR_LIBRARY_DIR}/cmake/gnuradio
+  DESTINATION ${GR_CMAKE_DIR}
 )
 
 install(


flynn378:
After more testing this did not completely fix the issue.  I will work more on the issue

flynn378:
The error in my other testing was do to a .bbappend file being added to the build systems I was not aware of, and the worng revision of GNU Radio was getting built in Yocto.  This pull request solves #2515

cmake/Modules/GrPython.cmake
```diff
@@ -110,6 +110,18 @@ if os.name == 'nt':
 endif()
 file(TO_CMAKE_PATH ${GR_PYTHON_DIR} GR_PYTHON_DIR)
 
+########################################################################
+# Sets the python relative installation directory GR_PYTHON_RELATIVE
+########################################################################
+if(NOT DEFINED GR_PYTHON_RELATIVE)
+execute_process(
+  COMMAND "${PYTHON_EXECUTABLE}" -c "if True:
```
michaelld: why the "if True:"? that command is always True, of course ...so seems unnecessary!

gnuradio-runtime/swig/CMakeLists.txt
```diff
@@ -74,7 +74,7 @@ GR_PYTHON_INSTALL(FILES ${CMAKE_CURRENT_BINARY_DIR}/runtime_swig.py
 install(EXPORT runtime_swig-export
   FILE runtime_swigTargets.cmake
   NAMESPACE gnuradio::
-  DESTINATION ${GR_CMAKE_DIR}
+  DESTINATION ${GR_LIBRARY_DIR}/cmake/gnuradio
```
michaelld: why are you reverting this? it should -not- be GR_LIBRARY_DIR, since that's not where the other exports are installed!

cmake/Modules/GrPython.cmake
flynn378: I was attempting to perform basic error checking.  CMake scripting is not really in my area.  I will remove this and test on Monday.

gnuradio-runtime/swig/CMakeLists.txt
flynn378: I accidentally changed this back when I copied the working file I had for the PR when I opened the issue.  I will change this to the other commit and retest on Monday.

flynn378:
Based on the issue and guidance from CMake changing the install (EXPORT ..... sections is not the solution to the issue.  The section that needs to be changed for this issue is the install( TARGETS.......)   It is not the location of the installation that causes the issue, it is what CMake puts into the .cmake file to locate the file that is the issue.  I am not a CMake expert at all, and from what I can tell the install(EXPORT is where the files go, but the install (TARGETS .... effects the contents of the .cmake file.  I have tested this change with 3 different OOT modules and they all built with the changes.  I will complete operational testing of the changes on Monday.  I will not say this is the most correct way to fix the issue, just that currently it seems to solve the issue.

noc0lour:
@flynn378 thanks for putting effort into this. Ping me once you are done :+1:.

flynn378:
I have completed testing. This has been tested on a complete OpenEmbedded build and local build.  Basic testing with 3.8 OOT modules was performed without any issues appearing.

flynn378:
@noc0lour This should be ready for review

noc0lour:
I tested yesterday and in general the change doesn't change that much for me. Is this only an issue on cross-compilation? I think I'm missing some context. After rebasing & diffing your changeset to master the changeset is also quite small (fits easily in one commit)'.

But if that fixes the cross-build I'm fine if you rebase & squash your changes.

flynn378:
I have rebased and squashed the commits, I tested again and it is still working.

flynn378:
@noc0lour This only effects cross-compiling, although per the CMAKE manual it is the also the correct solution for in general use.   
see https://cmake.org/Bug/print_bug_page.php?bug_id=14367 for more information.
What is happening without the change is the path for the  runtime_swigTargets-release.cmake files which OOT modules use to link and build against are set to the full path they will be installed at build time.  So if  the file will be in /usr/lib/*.so that will be the path set in the file.  It drops the variable "${_IMPORT_PREFIX}/" from the front which prevents OOT modules from finding the correct package during cross-compile.  There is a an email chain from me which details the issue in more detail and adds context.

PR #2520, opened 2019-06-02
gr-video-sdl: Fixed typo in cmake (MSVC only)

gnieboer:
Fixes minor typo in cmake file that only impacts MSVC


marcusmueller:
merged! Thank you. 

Note for myself: this bug is not present in maint-3.7

PR #2521, opened 2019-06-02
CMake: Fixes error when both debug and release python libs installed

gnieboer:
If both debug/release versions are present, cmake will failed with an error about the "optimized" keyword, essentially we have to pick a version when using the INTERFACE_LINK_LIBRARIES property.

So this does a quick check for that case, and links to Debug if Debug, and Release if anything else.


michaelld:
Hmm ... let me see if I can find a way to install or fake both release and debug Python libraries & test this out. I have to believe that this scenario will impact very few end-users!

mbr0wn:
This seems harmless. Let's review and test other stuff...

PR #2523, opened 2019-06-02
gr-uhd: Replaced not with !

gnieboer:
Replacing 'not' with '!' for msvc compatibility


marcusmueller:
Hm, I don't *personally* like `not` (it's unnecessarily inconsistent with the majority of our code base), so I'm generally fine with replacing it with the original `!`, but: 

MSVC should the support the _heck_ of this standard C++ feature, afaik. Don't have a windows machine at hand – can you try whether the `/Za` compiler flag fixes the issue? (disables Microsoft-specific language extensions, also known as "tries to make MSVC behave like a real C++ compiler")

mbr0wn:
I also don't like `not`.

gnieboer:
Sure, I need to rebuild again anyways, so will give that a shot and see what happens tomorrow just for future reference.

PR #2525, opened 2019-06-02
cmake: fixed gsl definition add for windows

gnieboer:
Minor syntax fix related to GSL on windows


marcusmueller:
Does this need a backport to maint-3.7?

noc0lour:
@marcusmueller How does it look like in 3.7? If it's the same I bet you can just do the same.

Issue #2518, opened 2019-05-31
qtgui time sink complex message input broken under grc


ChromeBumper:
If I connect to a qtgui time sink in grc it shows the input port as "in0" but the cc code declares the message port as "in"
short term options are to either modify time_sink_c_impl.cc to accept "in0" as:
```
      // setup PDU handling input port
//      message_port_register_in(pmt::mp("in"));
//      set_msg_handler(pmt::mp("in"),
//                      boost::bind(&time_sink_c_impl::handle_pdus, this, _1));
      message_port_register_in(pmt::mp("in0"));
      set_msg_handler(pmt::mp("in0"),
                      boost::bind(&time_sink_c_impl::handle_pdus, this, _1));
```
or to modify the python code that grc generates (every time) 
```
#        self.msg_connect((self.dave_specsplit_0, 'AngleVsFreq'), (self.qtgui_time_sink_x_0_0, 'in0'))
```
change to:
```
        self.msg_connect((self.dave_specsplit_0, 'AngleVsFreq'), (self.qtgui_time_sink_x_0_0, 'in'))
```

dl1ksv:
in0 is the stream port and in is the message port, but the yml definition does not contain a message port definit ion as for instance the frequency sink does.

marcusmueller:
Fixed by #2543

PR #2526, opened 2019-06-05
Add message input port to qtgui_time_sink

dl1ksv:
The qtgui time sink can act as message sink. But as the definition of a message port is missing, the code generated by grc is wrong and fails to run. See #2518 
This patch adds a message port definition

### DIFF
--- a/gr-qtgui/grc/qtgui_time_sink_x.block.yml
+++ b/gr-qtgui/grc/qtgui_time_sink_x.block.yml
@@ -815,7 +815,12 @@ inputs:
     dtype: ${ type.t }
     multiplicity: ${ (0 if type.startswith('msg') else nconnections) }
     optional: true
-
+    hide: ${ (type.startswith('msg') ) }
+-   domain: message
+    id: in
+    optional: true
+    hide: ${ (not type.startswith('msg') ) } 
+    
 templates:
     imports: |-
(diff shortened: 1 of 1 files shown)


dl1ksv:
Please don't merge. This is a more general problem. It's also related to #2528 .
My proposed fix works, but it's a wrong solution. The ports will be distinguished by the port numbers.
in01 is a stream port an in0 is a message stream port. I'll make a new pr to fix #2518 and #2528 tomorrow.

dl1ksv:
See #2540 . This is a better and more complete fix

dl1ksv:
Meanwhile my favourite solution is #2543

bhilburn:
We need someone to pull this change, load up the qtgui time sink, and double check that the new message port works.

mormj:
Verified this PR that it fixes the issue, but following the discussion on the other referenced PR - it looks like this fix is superceded by #2543 

If folks are ok reverting to the 3.7 automatic naming, this one can probably be closed in favor of 2543.  The only issue with 2543 is that it will break a lot of flowgraphs, which #2540 is probably safer in that sense

mbr0wn:
Closing in favour of  #2543

mbr0wn:
New message port works fine.

mbr0wn:
...but some examples need updating.

PR #2533, opened 2019-06-05
modtool: add 3.8 version for info command

### LINKED ISSUE #2403: gr-modtool: Update gr_modtool info for the 3.8 release
We need to update the API version in the `gr_modtool info` for the `3.8 release`.

mormj:
Add version information detection to the part of the code being called by gr_modtool info

fixes #2403

### DIFF
--- a/gr-utils/python/modtool/core/info.py
+++ b/gr-utils/python/modtool/core/info.py
@@ -57,6 +57,8 @@ class ModToolInfo(ModTool):
                 os.path.isdir(os.path.join('include', 'gnuradio', mod_info['modname']))
                 ):
             self.info['version'] = '37'
+        if not os.path.isfile(os.path.join('cmake', 'Modules', 'FindCppUnit.cmake')):
+            self.info['version'] = '38'
         mod_info['version'] = self.info['version']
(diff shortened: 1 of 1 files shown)


gr-utils/python/modtool/core/info.py
```diff
@@ -57,6 +57,8 @@ def run(self):
                 os.path.isdir(os.path.join('include', 'gnuradio', mod_info['modname']))
                 ):
             self.info['version'] = '37'
+        if not os.path.isfile(os.path.join('cmake', 'Modules', 'FindCppUnit.cmake')):
```
mbr0wn: Not sure if this is the best way to do this.... but OK.

mormj: It was the code in place, but in a different place - went with the quick fix.  I can make this more robust if you think it is necessary - and ideally a version number would be dumped into the module directory when the modtool is run.  Would you support this as a method going forward, to get a more specific version as what version of the modtool created the module?

mbr0wn: That would be a good future change, but for now, this is fine.

PR #2535, opened 2019-06-06
blocks: re-implement selector block

### LINKED ISSUE #2460: Implement blks2 Selector replacement

mormj:
blks2_selector is deprecated - this is a backport for 3.7 from the master branch for 3.8

Code review done as part of #2483 

fixes #2460


gr-blocks/lib/selector_impl.cc
```diff
@@ -0,0 +1,151 @@
...
+            io_signature::make(1, -1, itemsize)),
+      d_itemsize(itemsize),
+      d_enabled(true),
+      d_input_index(input_index),
+      d_output_index(output_index),
+      d_num_inputs(0),
+      d_num_outputs(0)
+{
+    message_port_register_in(pmt::mp("en"));
+    set_msg_handler(pmt::mp("en"), [this](pmt::pmt_t msg) { this->handle_enable(msg); });
```
mbr0wn: This will fail to compile on 3.7, even if it's so much more beautiful.
```suggestion
    set_msg_handler(pmt::mp("en"), boost::bind(&selector_impl::handle_enable, this _1));
```

Issue #2528, opened 2019-06-05
qtgui_message_inputs example invalid message port


mormj:
On master branch, the qtgui_message_inputs.py when run from grc returns the following error:

-----
Executing: /usr/bin/python3 -u /share/gnuradio/grdev/src/gnuradio/gr-qtgui/examples/qtgui_message_inputs.py

Could not find port: in0 in:
in
system

Traceback (most recent call last):
  File "/share/gnuradio/grdev/src/gnuradio/gr-qtgui/examples/qtgui_message_inputs.py", line 593, in <module>
    main()
  File "/share/gnuradio/grdev/src/gnuradio/gr-qtgui/examples/qtgui_message_inputs.py", line 581, in main
    tb = top_block_cls()
  File "/share/gnuradio/grdev/src/gnuradio/gr-qtgui/examples/qtgui_message_inputs.py", line 515, in __init__
    self.msg_connect((self.blocks_tagged_stream_to_pdu_0, 'pdus'), (self.qtgui_const_sink_x_0, 'in0'))
  File "/share/gnuradio/grdev/lib/python3.6/dist-packages/gnuradio/gr/hier_block2.py", line 60, in wrapped
    func(self, src.to_basic_block(), srcport, dst.to_basic_block(), dstport)
  File "/share/gnuradio/grdev/lib/python3.6/dist-packages/gnuradio/gr/hier_block2.py", line 132, in msg_connect
    self.primitive_msg_connect(*args)
  File "/share/gnuradio/grdev/lib/python3.6/dist-packages/gnuradio/gr/runtime_swig.py", line 5473, in primitive_msg_connect
    return _runtime_swig.top_block_sptr_primitive_msg_connect(self, *args)
RuntimeError: invalid msg port in connect() or disconnect()

PR #2540, opened 2019-06-07
Adopt PDU input port naming to yml description

dl1ksv:
In #2518 and #2528 invalid port names are reported.
This patch adopts the message stream input port names to the yml description which handles the port definition by the multiplicity parameter.

With this fix the qtgui_message_inputs example from the example gr-qtgui/examples directory runs without errors

### DIFF
--- a/gr-qtgui/lib/const_sink_c_impl.cc
+++ b/gr-qtgui/lib/const_sink_c_impl.cc
@@ -68,8 +68,8 @@ namespace gr {
       d_index = 0;
 
       // setup PDU handling input port
-      message_port_register_in(pmt::mp("in"));
-      set_msg_handler(pmt::mp("in"),
+      message_port_register_in(pmt::mp("in0"));
+      set_msg_handler(pmt::mp("in0"),
                       boost::bind(&const_sink_c_impl::handle_pdus, this, _1));
 
       for(int i = 0; i < d_nconnections; i++) {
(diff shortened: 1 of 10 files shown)


michaelld:
Let me state up front that (1) it's important for examples to work! (2) I'm not intimately familiar with the new YML nor GRC in general. (3) I'm still playing catchup getting the new GR 3.8 prerelease working on my OSX boxes, mainly because I have to have GR 3.7 installed for my primary work.

If there is just a single message port and it has no better name, then we have traditionally named it "in" or "out". We only append the # if there are multiple message ports. We have done this even when the number is variable: when 1 port is selected, we use "in"/"out", when 2, "in[01]" and "out[01]" and so forth.

I note the traditional naming because -- unless this issue is limited just to qtgui, which I highly doubt -- then this issue will break other examples not in qtgui as well as OOT message port naming. This breakage is not what we desire ... and thus this PR's "fix" is, to me, not the way to go. IMHO we need to be fixing this naming inside GRC, rather than outside.

All of the above said: In GR 3.8 we're already breaking away from various GR 3.7 traditions ... maybe this is another one to be broken?

dl1ksv:
@michaelld You are right. There is a change in the port naming scheme. Take a simple add block with two inputs and one output. In 3.8 the input ports are named in00 and in01 and the output port out0. While in 3.7 they are named in0, in1 and out.  For stream ports this has no effects as the connections happen by port number. If the prefix is in or in0 has no relevance in this case. But message ports are connected by name and the name is coded in the block code.

So I aggree: Fixing it inside grc it would be the safer side. I see if I can find out.

dl1ksv:
@michaelld It's easy to revert the naming scheme completely, which I'd prefer. See #2543 . But another solution could be only revert for message ports.

mbr0wn:
Superseded by #2543

PR #2541, opened 2019-06-07
grc: introduce flag 'show_id' to show block id

bastibl:
This is supposed to address https://github.com/gnuradio/gnuradio/issues/2235 by introducing a flag `show_id` to show the block ID in GRC's main canvas. This is useful for variables and parameters but maybe other blocks too. So maybe a flag is better than a list or a regex on the block ID.


bastibl:
arg, there is already a PR for that :-/

Issue #2494, opened 2019-05-21
grc fails to build constellation object if constellation != variable constellation


dl1ksv:
grc fails with

Value "digital.constellation_calcdist(const_points, sym_map, rot_sym, dims) if (str(type) == "calcdist") else digital.constellation_type" cannot be evaluated:
module 'gnuradio.digital' has no attribute 'constellation_type'

This is due to https://github.com/gnuradio/gnuradio/blob/c7780bb2abaebd4a0a3ce0ada5d446d554b8adcc/gr-digital/grc/digital_constellation.block.yml#L42

as at this point type cannot be evaluated which probably is intended.

dl1ksv:
Just open a new flowgraph in grc and add a Constellation object from Modulators . Then you get immediately:

ERROR:gnuradio.grc.core.FlowGraph:Failed to evaluate variable block variable_constellation_0
Traceback (most recent call last):
  File "/usr/local/gnuradio/lib/python3.7/dist-packages/gnuradio/grc/core/FlowGraph.py", line 226, in renew_namespace
    value = eval(variable_block.value, namespace, variable_block.namespace)
  File "<string>", line 1, in <module>
AttributeError: module 'gnuradio.digital' has no attribute 'constellation_type'

dj0abr:
the problem still exists if the output language is switched to C++. While it work in Python.

PR #2543, opened 2019-06-07
Go back to 3.7 port naming scheme.

dl1ksv:
See discussion in #2540

### DIFF
--- a/grc/core/ports/port.py
+++ b/grc/core/ports/port.py
@@ -46,7 +46,7 @@ class Port(Element):
         self._dir = direction
         self.key = id
         if not label:
-            label = id if not id.isdigit() else {'sink': 'in', 'source': 'out'}[direction] + id
+            label = id if not id.isdigit() else {'sink': 'in', 'source': 'out'}[direction]
         self.name = self._base_name = label
 
         self.domain = domain


michaelld:
Hmmm ... does this change the default label to -always- be just "in" or "out", regardless of how many I or O message ports there are? I'm no expert in how message ports are named in GRC ...

dl1ksv:
As far as I understand : 
In the yaml description ports can be named by id or use an automatic generated name by grc. 
In 3.7 automatic generated stream input  ports are named in0, in1, in2, ... while in 3.8  they are named in00, in01, in02 ... . As long as there are no port named in10, in11, in12 I see no advantage in this naming scheme. But there are no problems to do connections as there are referenced by the number of the port.

Message ports are differently handled.The ports are addressed by name, which is hardcoded in the block code.( As far as I can see).   Here again yaml description can use id or generate the port name by grc.

In 3.7 the generated name (input ) port name is in while in 3.8 the name is in0. I could not detect any block that contained more than one automated named port.

Summary:
I don't no why there is a new port naming in 3.8 and I can't any advantages in the moment.

So before proposing further fixes I need a decision how (automatic) port names should be generated in the future:

message: in / out or in0 / out0 

streams: in0,in1,in2, ../out0,out1,out2,..    or in00, in01, in02.... /out00 , out01,out02,......

I'd prefer message in/out and streams in0,.. /out0 ..

which would be handled by this pr.

mbr0wn:
I'm in favour of this PR. Will merge if there are no complaints.

michaelld:
+1 from me

mormj:
Tested out this PR - will require many of the example grc files to be fixed up, but it works and appears to generally fix the problem.

mbr0wn:
Merging manually.

mbr0wn:
@dl1ksv Can you go through your issues and PRs and see which ones are still required after this change?

dl1ksv:
@mbr0wn All port related issues I know are fixed. #2494 and #2557 still remain.

PR #2550, opened 2019-06-12
cmake: Allow custom library name to be supplied for FindZeroMQ module

gnieboer:
Newer versions of libzmq use the boost naming convention on Windows (libzmq-v140-mt-1.x.x.dll).  However the FindZeroMQ cmake module doesn't natively find these variations. 

So this just adds a user variable to the cmake script so that the windows build scripts can manually generate the correct filename and pass it to cmake depending on debug/release builds.


gnieboer:
Oh, and this affects both master and maint-3.7

marcusmueller:
Merging this. Do you happen to have a 3.7 branch, too?

PR #2562, opened 2019-06-19
wavfile.cc bugfix

gmazilla:
'Wav File Source' block in gnuradio reads PCM 8bit wav files incorrectly. It adds random huge DC component. Caused by garbage in high byte of buf_16bit (gnuradio 3.7.13.4)


mbr0wn:
Many thanks! Merged manually.

PR #2563, opened 2019-06-19
cmake: check for python module click as dependency of modtool

### LINKED ISSUE #2292: modtool requires click dependencies

mbr0wn:
Previously cmake did not check if click is installed as it is a runtime
requirement. If CMAKE_CROSSCOMPILING is set the module checking is
skipped.

Closes #2292

Resubmitted from #2479. Wanted to re-run CI

### DIFF
--- a/gr-utils/python/modtool/CMakeLists.txt
+++ b/gr-utils/python/modtool/CMakeLists.txt
@@ -19,6 +19,18 @@
 
 include(GrPython)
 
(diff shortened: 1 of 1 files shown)


mbr0wn:
Closing until we have a BuildBot update, since this is a dupe of #2479 anyway.

PR #2565, opened 2019-06-20
ControlPort QA

bastibl:
I forgot to port the ControlPort QA test to the new API.
I also noticed that the tests throw an exception with Python3 as RPC calls for strings use Python2 string conversion. This patch would switch to Python3 conversion but breaks Python2 compatibility. Not sure if there is a better solution... On the other hand `RPC_get_string` is only used in unit tests, so maybe it's *good enough*?

### DIFF
--- a/gnuradio-runtime/include/gnuradio/pycallback_object.h
+++ b/gnuradio-runtime/include/gnuradio/pycallback_object.h
@@ -161,7 +161,11 @@ private:
 template <>
 std::string pycallback_object<std::string>::pyCast(PyObject* obj)
 {
+#if PY_MAJOR_VERSION >= 3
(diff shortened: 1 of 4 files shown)


mbr0wn:
I would say we take it, but we would need to disable those QA tests when using Python 2 (and maybe print a little warning, but not fail the tests).

mbr0wn:
...I guess it's not impossible to have Python2k compat, right? It's just that the current unit tests don't account for that?

bastibl:
I'm not sure if I got you right, but I added a compile switch for Py2/Py3. Let's see what buildbot thinks about it. The unit test for CtrlPort w/ Py3 seem to be disabled by the CI, maybe we can enable them again after this change.

PR #2566, opened 2019-06-21
filter: Add file taps loader block

mbr0wn:
This is a block that allows loading filter taps from a file (from the
filter design tool).

This is  #1753, but rebased. It doesn't fully work though. @m3x1m0m, do you want to take another stab at this?
@willcode you also had some comments. @haakov maybe you can help with var_value, it's not doing what it should.


haakov:
var_value and var_make are missing some curly brackets :)

mbr0wn:
Thanks, I'll try that! `var_make` seems to be working though so I'm not 100% convinced yet.

mormj:
Pulled this and ran - not able to run as I'd expect in GRC due to the variable_id issues - perhaps due to changes in #2568 

Basically, there is no id for this block to populate as the input to other blocks

image

image

Perhaps, it just needs a show_id flag as mentioned in #2568

gr-filter/grc/variable_file_filter_taps.block.yml
```diff
@@ -0,0 +1,26 @@
+id: variable_file_filter_taps
+label: Filter Taps Loader
+flags: [ python ]
```
mormj: flags: [ show_id, python ]

mormj:
Also, after adding the show_id flag, the File loader block returns None when using the complex taps file that was included in the examples/ dir

image

haakov:
Did you move `var_value` to a top-level `value` field?

gr-filter/grc/variable_file_filter_taps.block.yml
```diff
@@ -0,0 +1,26 @@
...
+parameters:
+-   id: file
+    label: File
+    dtype: file_open
+-   id: verbose
+    label: Verbose Output
+    dtype: bool
+    default: False
+
+templates:
```
mormj:
```
value: ${filter.file_taps_loader(file, verbose).get_taps()}

templates:
    imports: from gnuradio import filter
    var_make: |-
        self.${id} = ${id} = filter.file_taps_loader(${file}, ${verbose}).get_taps()
    var_value: |-
        filter.file_taps_loader(${file}, ${verbose}).get_taps()
```

mormj:
Thanks - that pushes it further.  Still issues with the example taps provided - they load as real taps.  Filename indicates complex taps but nowhere in the file is this stated.  Definitely being loaded as real taps:

image

Further issue is that I don't think the saving of taps from gr_filter_design is currently working

mormj:
Close in favor of #2689 (which is the continuation of this PR)

PR #2569, opened 2019-06-22
docs: removed sphinx

777arc:
If in the future we identify stuff that *was* in Sphinx and is now not documented, but needed, we will just figure out how to get it incorporated into doxygen.


777arc:
Since posting the message to https://www.gnuradio.org/doc/sphinx/ there have been two people who voiced that they think there should definitely be a Python API, but they both also mentioned that the current one was difficult to use and didn't always have the info they were looking for.  Yes, we could just add Python parsing to Doxygen, but I don't think just doing that will get us to a better point.  Because almost all of our Python functions/classes are SWIGed, we really just need a well-organized "listing" of these functions/classes, with links to the corresponding C++ function/class.  The question is how to create this type of listing automatically with Doxygen, or if it needs to just be created manually.  Either way it should probably live in tree.

777arc:
A couple months later and there still hasn't been anyone with concerns about sphinx manual being removed.  I did receive some emails but it's people complaining about stuff unrelated to the python docs altogether. 

So all we really need to do is make sure there is a clear reference for the functions/classes people actually use from within a python block or python flowgraph (as opposed to GR internal stuff, like block specific stuff).  E.g. functions related to stream tags, message passing, PMTs, and maybe a few runtime things.  

How to do it without needing Sphinx?  Well we can get doxygen to parse a small hand-picked selection of the python code, that way we only get what we need.  It will involve minor touching up of the inline docs.   Any thoughts?

michaelld:
@777arc can you rebase this PR to deal with the conflict in "docs/sphinx/source/index.rst"?

777arc:
The "Sphinx" manual that is produced is here, it's pretty repetitive and the python specific documentation is largely not filled out https://www.gnuradio.org/doc/sphinx-v3.7.13.4/

Yeah I can rebase if we decide to move forward with this.

So one idea is to just have a list of all these Python specific functions at the end of their corresponding Usage Manual page, on the wiki.  Most of these functions have no actual doc-strings, and the ones that do can be copied into that list as well. So for example, the relevent functions/classes from https://www.gnuradio.org/doc/sphinx-v3.7.13.4/#pmt would go at the end of https://wiki.gnuradio.org/index.php/Polymorphic_Types_(PMTs)

michaelld:
OK; thanks for the info. Yes, I see that the Doxygen docs are much better & nicer than the Sphinx ones. :+1: from me to merge this PR. That said, let's get feedback from at least 1 other, just to have a few more brain cells on it.

mbr0wn:
Resolved and merged manually.

PR #2572, opened 2019-06-23
dtv: Remove superfluous set_relative_rate() in DVB-T inner coder (3.7).

drmpeg:
Also remove assert that was failing for 7/8 rate on debug builds.


marcusmueller:
hm, seems like the right thing given this thing is `block`, but: doesn't that processing step actually *have* a fixed relative rate?

PR #2576, opened 2019-06-24
Fix CMake Error "VOLK required but not found."

Imperator26:
When manually building from source:
CMake Error at CMakeLists.txt:420 (message):
  VOLK required but not found.

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,7 +80,7 @@ Abbreviated instructions are duplicated below.
 
 2. Checkout the latest code:
     ```
-    $ git clone https://github.com/gnuradio/gnuradio.git
+    $ git clone --recursive https://github.com/gnuradio/gnuradio.git
     ```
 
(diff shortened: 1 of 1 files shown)


marcusmueller:
Nice one!

So, I'm going ahead and will merge this. One thing, though: If you can imagine contributing more to GNU Radio (I hope you'll really do!), then we'll need to set you up with a piece of paperwork that says "hey, my contributions to GNU Radio are free software, and to defend that, I'm assigning copyright to the Free Software Foundation". That's really not necessary for small changes like this one, but as soon as you, say, end up writing a Saturn V thrust control interface for GNU Radio that you want to upstream, we'll need that. @bhilburn can set you up with the whole thing, should you want it.

Imperator26:
@marcusmueller If I'll write some stuff in the future I won't mind doing that.

Issue #2536, opened 2019-06-06
grc: Help --> About box displays nothing and gives error


mormj:
Clicking on the Help-->About menu item in grc has no effect - no dialog pops up, and in the terminal used to launch gnuradio-companion, the following error is displayed:

Traceback (most recent call last):
  File "/share/gnuradio/grdev/lib/python3.6/dist-packages/gnuradio/grc/gui/Dialogs.py", line 274, in show_about
    ad.set_logo(Gtk.IconTheme().load_icon('gnuradio-grc', 64, 0))
GLib.Error: gtk-icon-theme-error-quark: Icon 'gnuradio-grc' not present in theme (null) (0)

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/share/gnuradio/grdev/lib/python3.6/dist-packages/gnuradio/grc/gui/Application.py", line 445, in _handle_action
    Dialogs.show_about(main, self.platform.config)
  File "/share/gnuradio/grdev/lib/python3.6/dist-packages/gnuradio/grc/gui/Dialogs.py", line 276, in show_about
    log.debug("Failed to set window logo")
NameError: name 'log' is not defined

haakov:
It works fine for me, but my master branch is a couple weeks old. (https://github.com/gnuradio/gnuradio/commit/73b074f121b0ab2ac38336916a60891c4d88d2cb is the latest commit)

bastibl:
Had the same problem. It seems to happen if the freedesktop stuff is not installed (in the build directory `bash ./grc/scripts/freedesktop/grc_setup_freedesktop install`) Anyway, I think the exception handler in Dialogs.py is broken since the logger is not created.

dl1ksv:
Got the same problems with the lastest commit. But the icons are in CMAKE_INSTALL_PREFIX/share/icons

But looking at 
Gtk.IconTheme().get_search_path()
this directory is not part of the search path.

As ~/.local/share/icons is part of the search path I copied the icons subdirectory to this position and it works.
So adding CMAKE_INSTALL_PREFIX/share/icons to the icon_search_path could solve this issue.
The name error has to be fixed separately.

dl1ksv:
An install script grc_setup_freedesktop is generated, but will not be executed during the installation process. So if CMAKE_INSTALL_PREFIX/share is not in the XDG_DATA_DIRS path  list the icons won't be found.
The easiest way to fix is to execute the generated script during installation. see #2580

[...]

PR #2580, opened 2019-06-26
Execute generated freedesktop install script

dl1ksv:
At the moment an install script for the freedesktop icons is generated but not executed, so the icons are missing. This sometimes leads to missing icons or even errors . ( See #2536 )

Executing the generated freedesktop installaition script will fix this problem.

### DIFF
--- a/grc/scripts/freedesktop/CMakeLists.txt
+++ b/grc/scripts/freedesktop/CMakeLists.txt
@@ -66,5 +66,9 @@ if(UNIX AND HAVE_XDG_UTILS)
     install(
         PROGRAMS ${CMAKE_CURRENT_BINARY_DIR}/grc_setup_freedesktop
         DESTINATION ${GR_PKG_LIBEXEC_DIR}
+        PERMISSIONS OWNER_READ OWNER_WRITE OWNER_EXECUTE GROUP_READ GROUP_EXECUTE WORLD_READ WORLD_EXECUTE
+    )
+    install (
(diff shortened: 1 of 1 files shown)


ZeroChaos-:
typically package managers (apt, dnf, etc) do these kinds of things in a post install routine.  executing it too early (say, during install) means the package is still staged in a temp folder and thus things often end up wrong.  You could certainly add this, but I'd suggest having some flag to turn it off so it doesn't break *everyone* installing a gnuradio package

marcusmueller:
So, @dl1ksv is right: these icons should be copied to the destination during the *install* step; @ZeroChaos- is also right, the update of the icon database is a *post-installation* task in most packagers.

I don't know a lot of distro policies, but as far as I can tell, it'd be discouraged to copy icons during post-installation. @ZeroChaos-, would that apply to Gentoo and XDG icons, too?

If that is the case, we need to:

* merge this PR,
* simultaneously change the freedesktop install script to make separate steps `install` and `update-db`
* have a CMake flag that optionally disables the `update-db` step
* communicate to distro packagers that they should use that flag and update themselves using `xdg-icon-resource forceupdate` in post-installation

ZeroChaos-:
@marcusmueller yes, in gentoo we do this step in post install, it's not possible to do it during install (permissions)
https://github.com/gentoo/gentoo/blob/master/net-wireless/gnuradio/gnuradio-9999.ebuild#L220

Nothing wrong with this idea, it would help source users.  We just need a way to turn it off so it doesn't break all packagers.

mbr0wn:
@ZeroChaos- I've opened https://github.com/gnuradio/gnuradio/pull/3031 to discuss the changes required to disable this.

PR #2581, opened 2019-06-26
Use gnuradio icon if available

dl1ksv:
The existence of a gnuradio ( theme ) icon is checked, but if found not set.
So try to load the icon and if found set it.


marcusmueller:
Thank you :)

PR #2582, opened 2019-06-27
uhd: docs: Removed spurious reference to io_type_t

mbr0wn:
io_type_t was removed a while ago, but there was a reference to it in
the documentation which is now removed.


777arc:
Ah thanks.  Does the "User Docs" page on USRP Source look ok?  https://wiki.gnuradio.org/index.php/USRP_Source

mbr0wn:
Yeah, that looks good.

Issue #1841, opened 2018-06-23
remove non-thread-safe `rand` in remaining code


marcusmueller:
Much as in the spirit of https://github.com/gnuradio/gnuradio/pull/1562, we need to *completely* eradicate `rand` from our code base wherever things  *could* be multithreaded. That's basically everywhere, even, and especially, in test cases.

marcusmueller:
This will entail replacing `rand`, `srand` and friends with `xoroshiro128p_next` in a lot of places.

marcusmueller:
Seems like it's done.

rear1019:
> Seems like it's done.

I beg to differ. See output of `git grep -E "\<s?rand\>[ \t]*\(" -- "*.cc" "*.c" "*.h"`:

- `gr-digital/lib/ofdm_mapper_bcv_impl.cc`
- `gr-digital/lib/qa_header_format.cc` (A fix can be backported from master if I remember correctly.)
- `gr-trellis/lib/interleaver.cc`
- three files in `gr-vocoder/lib/codec2`
- `cmake/msvc/config.h`

The last file provides a wrapper for `[s]random()` (defined by POSIX) which calls `[s]rand()`. `[s]random()` is used by QA code (see output of `git grep -E "\<s?random\>[ \t]*\(" -- "*.cc" "*.c" "*.h"`).

rear1019:
Please assign to me, I will look into the remaining usage of `[s]rand()` and `[s]random()`.

rear1019:
Pull request to remove `[s]random()` #2549. Concerning the usage of `[s]rand()`:

- `gr-digital/lib/ofdm_mapper_bcv_impl.cc`: Changes to code might be sort of an API change as user visible behavior changes. How should I proceed here?
- `gr-trellis/lib/interleaver.cc`: As above. This interleaver is used by gr-trellis only as far as I can tell.
- `gr-digital/lib/qa_header_format.cc`: Fix can be back-ported from master.
- `gr-vocoder/lib/codec2`: Files using `rand()` are not compiled or used in gr-vocoder. No changes are required here IMO.

marcusmueller:
Thank you for this!

The API change thing is hard. We officially can't do it on 3.7.x.y, because API is signified by the 7. We can't fix things and release them as 3.8, because that's a different thing already.

However, looking back at the history of the 3.7 releases, I'm afraid we won't significantly make matters less consistent than they are.

[...]

PR #2594, opened 2019-07-03
Replace rand() by gr::random() (maint-3.7 branch)

rear1019:
This is the last pull request to fix issue #1841. Changes in gr-digital and gr-trellis can be considered as API changes: User-visible behavior is modified. However, no code modifications in GNU Radio or OOT modules are required (as far as I can tell).


mbr0wn:
Needed rebasing after clang-format. Continuing on https://github.com/gnuradio/gnuradio/pull/3012

PR #2595, opened 2019-07-03
uhd: Add missing function calls and rearrange USRP YAML

haakov:
@michaelld has brought to my attention that `set_auto_dc_offset` and `set_auto_iq_balance` are missing from the USRP block. That is, the parameters already exist but the calls are missing. I've added them here.

I've also taken the liberty to reorder parts of the make template so that the order matches the XML definition in 3.7, although I don't know if that's really desirable. @mbr0wn some input on this?

### DIFF
--- a/gr-uhd/grc/gen_uhd_usrp_blocks.py
+++ b/gr-uhd/grc/gen_uhd_usrp_blocks.py
@@ -154,58 +154,30 @@ templates:
             ${'$'}{len_tag_name},
             ${'%'} endif
         )
-        % for m in range(max_mboards):
-        ${'%'} if context.get('num_mboards')() > ${m}:
-        ${'%'} if context.get('sd_spec${m}')():
-        self.${'$'}{id}.set_subdev_spec(${'$'}{${'sd_spec' + str(m)}}, ${m})
-        ${'%'} endif
-        ${'%'} if context.get('time_source${m}')():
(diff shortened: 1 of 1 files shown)


gr-uhd/grc/gen_uhd_usrp_blocks.py
```diff
@@ -154,65 +154,78 @@
...
+        self.${'$'}{id}.set_bandwidth(${'$'}{${'bw' + str(n)}}, ${n})
+        ${'%'} endif
+        
+        % if sourk == 'source':
+        ${'%'} if context.get('rx_agc${n}')() == 'Enabled':
+        self.${'$'}{id}.set_rx_agc(True, ${n})
+        ${'%'} elif context.get('rx_agc${n}')() == 'Disabled':
+        self.${'$'}{id}.set_rx_agc(False, ${n})
         ${'%'} endif
+        self.${'$'}{id}.set_auto_dc_offset(${'$'}{${'dc_offs_enb' + str(n)}}, ${n})
```
mbr0wn: This call shouldn't always be called. There are USRPs which don't have this feature, and will then throw. Same with the next line.

mbr0wn: I fixed this in https://github.com/gnuradio/gnuradio/pull/3021.

mbr0wn:
Continued in https://github.com/gnuradio/gnuradio/pull/3021.

PR #2598, opened 2019-07-08
grc: add busports back into 3.8

### LINKED ISSUE #2277: GRC: bus ports missing
In `gr-fec` many examples make use of bus ports. These ports seem to be broken with 3.8. At least all corresponding examples are broken. Instead of connected bus ports the aforementioned flowgraphs show lots of unconnected ports. This is an issue for `BER examples` because these flowgraphs tend to have *lots* of ports per block.

mormj:
Bus ports had not been added back in since the refactor of grc
Hopefully this fully enables busports though there are still some
issues with the gr-fec flowgraphs unrelated to the bus connections

Fixes #2277

### DIFF
--- a/gr-blocks/grc/blocks_null_sink.block.yml
+++ b/gr-blocks/grc/blocks_null_sink.block.yml
@@ -21,7 +21,7 @@ parameters:
     dtype: int
     default: '1'
     hide: part
--   id: bus_conns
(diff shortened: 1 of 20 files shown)


mormj:
Merged as part of rebased #2622.  Closing.

PR #2602, opened 2019-07-09
Check seperately for click-plugins missing

nickoe:
It was hard to determine from the cmake output what python module was
missing. Let the error output be more exact.


noc0lour:
Thanks, looks good.

nickoe:
Thank you too :)

PR #2605, opened 2019-07-10
digital: constellation object yml error

### LINKED ISSUE #2494: grc fails to build constellation object if constellation != variable constellation

mormj:
value parameter field has invalid reference to constellation_type

Cleaned up the yml, but probably needs some insight into the
yml generator

fixes #2494

### DIFF
--- a/gr-digital/grc/digital_constellation.block.yml
+++ b/gr-digital/grc/digital_constellation.block.yml
@@ -39,7 +39,8 @@ parameters:
     dtype: raw
     default: None
     hide: ${ ('part' if str(soft_dec_lut) == 'None' else 'none') }
-value: ${ digital.constellation_calcdist(const_points, sym_map, rot_sym, dims) if (str(type) == "calcdist") else digital.constellation_type}
+value: ${ digital.constellation_calcdist(const_points, sym_map, rot_sym, dims) if (str(type) == "calcdist") else getattr(digital,'constellation_'+str(type))()  }
+
 
 templates:
     imports: from gnuradio import digital


dl1ksv:
This fix leads to: 

ERROR:gnuradio.grc.core.FlowGraph:Failed to evaluate variable block constel
  File "/usr/local/gnuradio/lib/python3.7/dist-packages/gnuradio/grc/core/FlowGraph.py", line 226, in renew_namespace
  File "/usr/local/gnuradio/lib/python3.7/dist-packages/gnuradio/digital/digital_swig2.py", line 524, in make
    return _digital_swig2.constellation_calcdist_make(constell, pre_diff_code, rotational_symmetry, dimensionality, normalize_points)
TypeError: in method 'constellation_calcdist_make', argument 1 of type 'std::vector< gr_complex,std::allocator< gr_complex > >'
>>> Done
/usr/lib/python3/dist-packages/apport/report.py:13: DeprecationWarning: the imp module is deprecated in favour of importlib; see the module's documentation for alternative uses
  import fnmatch, glob, traceback, errno, sys, atexit, locale, imp
  File "/usr/local/gnuradio/lib/python3.7/dist-packages/gnuradio/grc/gui/VariableEditor.py", line 195, in set_value
    evaluated = str(block.params['value'].evaluate())
KeyError: 'value'

The variable block constel does not contain a value parameter.

A similar problem as in #2620

mormj:
I'm seeing this error pop up all over the place since some of the recent merges, but haven't traced it all the way back yet.  Will re-take a look at this PR now, though.

dl1ksv:
Meanwhile I found this issue on a lot of other blocks.

modulate_vector.block.yml
digital_constellation_rect.block.yml
variable_encoder_def_list.block.yml
variable_polar_encoder_systematic.block.yml

to name only a few.

Sometimes this error is hidden by an error preceding error.
Remove your patch and you won't see this error anymore.


Each block of type variable is affected that does not have an 

    -id: value
entry in the parameter section.

mormj:
I could be wrong, but I don't think the patch is the underlying issue.  The value field is necessary for other blocks to populate and verify generated object from within grc when referencing this object.  Prior to this patch, the value field was producing garbage and not returning anything.  Now that it returns an object ... maybe that triggers some other validation code??

dl1ksv:
Your are right, your patch is not the reason. But your patch makes the next issue visible. Without your patch an error Happens before renew_namespace is called .

mormj:
I'll need to re-verify this patch after the underlying issue is addressed - have you found an issue/PR in the queue for it, or are adding one?

mormj:
nevermind, i see it #2619

dl1ksv:
First I thougt that‘s an error in the yml description, but meanwhile I think that‘s an general grc problem. I‘m working on but at the moment I have no solution.

haakov:
So from what I can tell, the errors that appear are caused by a chain of inconsistencies in how variable blocks are handled. The first step of this chain is old XML files:

https://github.com/gnuradio/gnuradio/blob/76ecb19eb74e18a445011135217e486a08317a09/gr-digital/grc/digital_constellation.xml#L12-L32

One might expect that `var_value` and `var_make` will end up in the same part of the corresponding YAML file, since they are neighbours in the XML files. But they don't, :

https://github.com/gnuradio/gnuradio/blob/4d8d8d2923e52d6a6f97aaaa46fdda8a6581b02a/gr-digital/grc/digital_constellation.block.yml#L42-L57

`var_value` is renamed to `value` and placed as a top-level field in the YAML file.

If this is "wrong" or "right" is a matter of opinion, I suppose. `var_value` is after all not used for generating files like the rest of the `templates` fields. I'd love some discussion on this.

Moving on, GRC uses this `value` field to decide whether a block is a variable block or not:

https://github.com/gnuradio/gnuradio/blob/4d8d8d2923e52d6a6f97aaaa46fdda8a6581b02a/grc/core/blocks/block.py#L300-L302

However, there are parts of GRC that expect all variable blocks to have a _parameter_ called `value`. This is true for some blocks, like the standard GRC variable block:

https://github.com/gnuradio/gnuradio/blob/4d8d8d2923e52d6a6f97aaaa46fdda8a6581b02a/grc/blocks/variable.block.yml#L5-L10

Another example:

https://github.com/gnuradio/gnuradio/blob/4d8d8d2923e52d6a6f97aaaa46fdda8a6581b02a/gr-fec/grc/variable_ldpc_encoder_G.block.yml#L5-L28

haakov:
Needless to say, this causes problems when parts of GRC, like the Variable Editor, try to evaluate the value parameter of a block that doesn't have it. This is probably causing the var_value trouble in https://github.com/gnuradio/gnuradio/pull/2566 as well.

haakov:
Another related PR: https://github.com/gnuradio/gnuradio/pull/2620

dl1ksv:
@haakov To ungarble the different sources of the problems I limited my trials to 
https://github.com/gnuradio/gnuradio/blob/master/gr-digital/grc/variable_header_format_default.block.yml

Entering this block to a flowgraph leads to #2619 

This is why this block has no value parameter.

I found two possible solutions:

adding a value parameter to the block description see #2620 

But as there about 10 blocks that have this problem I tried to find a grc solution:

Here are my thoughts:
https://github.com/gnuradio/gnuradio/blob/0ab6a74caf362f2f95509e4303fc111d24a9cc1b/grc/gui/VariableEditor.py#L193-L196

This suggests to check the value of the block. But as not each variable block has a value parameter we should check the top level value field , each variable block owns.

This could be done by repacing this peace of code by:
`                 # Evaluate and show the value (if it is a variable)

                 if block.is_variable:
                        block.rewrite()
                        evaluated = str( eval(block.value,block.parent.namespace,block.namespace))
                        self.set_tooltip_text(evaluated)
`

The block.rewrite() is important, otherwise block.namespace is incorrect.

And in front of 
https://github.com/gnuradio/gnuradio/blob/0ab6a74caf362f2f95509e4303fc111d24a9cc1b/grc/core/FlowGraph.py#L226

`     variable_block.rewrite()`
should be added, too. Otherwise variable_block.namespace is incorrect if the block does not own a param value.

mormj:
Just to clarify - this PR is a fairly simple fix in yml to keep the constellation object from creating errors.

The conversation above corresponds to #2619.  I think the two are separable - it just happened that the value error issue discussed above emerged around the same time this PR was tested.

dl1ksv:
I checked this fix and it solves #2494 . Indepentenly how the value error problem will be solved, this fix should be applied asap.

dl1ksv:
My preferred solution is now  #2662

PR #2606, opened 2019-07-10
blocks: random_pdu mask parameter fails under python3

### LINKED ISSUE #2557: random pdu block fails if python 3 is used and byte mask > 0x7f
The random pdu block fails in gnuradio 3.8 with python 3.7 if byte_mask is set between 0x80 and 0xff .
See gr-digital/examples/packet/tx_stage0.py

  File "/usr/local/gnuradio/lib/python3.7/dist-packages/gnuradio/blocks/blocks_swig5.py", line 175, in make
    return _blocks_swig5.random_pdu_make(mintime, maxtime, byte_mask, length_modulo)
TypeError: in method 'random_pdu_make', argument 3 of type 'char'

If byte_mask <= 0x7f it runs without errors.

In gnuradio 3.7 and python 2.7 no errors happen if 0 <= byte_mask <= 0xff.

mormj:
The use of chr() in the grc file for the mask
creates an invalid input to the random_pdu make
function

1) Change the inputs to unsigned char since this is a mask
2) Remove the chr() from the grc yml

I'm not entirely sure what the purpose of the chr(mask) in
the make function was, but it created invalid input

Instead just pass the bitmask directly

fixes #2557

### DIFF
--- a/gr-blocks/grc/blocks_random_pdu.block.yml
+++ b/gr-blocks/grc/blocks_random_pdu.block.yml
@@ -34,6 +34,6 @@ templates:
     imports: |-
         from gnuradio import blocks
         import pmt
-    make: blocks.random_pdu(${minsize}, ${maxsize}, chr(${mask}), ${length_modulo})
+    make: blocks.random_pdu(${minsize}, ${maxsize}, ${mask}, ${length_modulo})
 
 file_format: 1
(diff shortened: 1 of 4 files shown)


dl1ksv:
Works for me ( with python 3.7 ). Not checked with python 2.7

PR #2615, opened 2019-07-12
 fec: grc code generation for encoder/decode objects

mormj:
add list() around nested map expressions in yml

Without the extra list expressions, grc passes map objects
into the BER curve gen hier blocks, instead of lists
of the enc/dec objects, which is what it is expecting

The examples still will not work without the busports code, which is included in this PR

### DIFF
--- a/gr-fec/grc/ldpc_decoder_def_list.block.yml
+++ b/gr-fec/grc/ldpc_decoder_def_list.block.yml
@@ -44,7 +44,7 @@ templates:
         % elif int(ndim)==1:
         self.${id} = ${id} = list(map( (lambda a: fec.ldpc_decoder.make(${file}, ${sigma}, ${max_iter})), range(0,${dim1})))
         % else:
-        self.${id} = ${id} = list(map( (lambda b: map( ( lambda a: fec.ldpc_decoder.make(${file}, ${sigma}, ${max_iter})), range(0,${dim2}) ) ), range(0,${dim1})))
+        self.${id} = ${id} = list(map( (lambda b: list( map( ( lambda a: fec.ldpc_decoder.make(${file}, ${sigma}, ${max_iter})), range(0,${dim2}) ) ) ), range(0,${dim1})))
         % endif
 
 documentation: |-
--- a/gr-fec/grc/variable_dummy_encoder_def_list.block.yml
+++ b/gr-fec/grc/variable_dummy_encoder_def_list.block.yml
@@ -33,9 +33,9 @@ templates:
         self.${id} = ${id} = list(map((lambda a:fec.dummy_encoder_make(${framebits})),\
         range(0,${dim1})))
         % else:
-        self.${id} = ${id} = list(map((lambda b:map((lambda a:\
+        self.${id} = ${id} = list(map((lambda b:list(map((lambda a: \
         fec.dummy_encoder_make(${framebits})),\
-        range(0,${dim2}))), range(0,${dim1})))
+        range(0,${dim2})))), range(0,${dim1})))
         % endif
 
 documentation: ""
--- a/gr-fec/examples/ber_curve_gen_ldpc.grc
+++ b/gr-fec/examples/ber_curve_gen_ldpc.grc
@@ -1,2026 +1,421 @@
-<?xml version='1.0' encoding='utf-8'?>
-<?grc format='1' created='3.7.9'?>
-<flow_graph>
-  <timestamp>Sun Jul 27 12:41:54 2014</timestamp>
-  <block>
-    <key>options</key>
-    <param>
-      <key>author</key>
-      <value></value>
-    </param>
-    <param>
-      <key>alias</key>
-      <value></value>
-    </param>
-    <param>
-      <key>window_size</key>
-      <value>2000,2000</value>
-    </param>
-    <param>
-      <key>category</key>
-      <value>Custom</value>
-    </param>
-    <param>
-      <key>comment</key>
-      <value></value>
-    </param>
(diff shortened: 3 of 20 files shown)


gr-fec/grc/ldpc_decoder_def_list.block.yml
```diff
@@ -44,7 +44,7 @@ templates:
         % elif int(ndim)==1:
         self.${id} = ${id} = list(map( (lambda a: fec.ldpc_decoder.make(${file}, ${sigma}, ${max_iter})), range(0,${dim1})))
         % else:
-        self.${id} = ${id} = list(map( (lambda b: map( ( lambda a: fec.ldpc_decoder.make(${file}, ${sigma}, ${max_iter})), range(0,${dim2}) ) ), range(0,${dim1})))
+        self.${id} = ${id} = list(map( (lambda b: list( map( ( lambda a: fec.ldpc_decoder.make(${file}, ${sigma}, ${max_iter})), range(0,${dim2}) ) ) ), range(0,${dim1})))
```
marcusmueller: this would be a nice place to use a list generator expression, but for now it's an improvement.

grc/core/FlowGraph.py
```diff
@@ -377,6 +378,7 @@ def import_data(self, data):
 
             block.import_data(**block_data)
 
+        self.rewrite()  # TODO: Figure out why this has to be called twice to populate bus ports correctly
```
marcusmueller: That's a relevant TODO. Can we make this a Github issue, @mormj?

gr-fec/grc/variable_dummy_encoder_def_list.block.yml
```diff
@@ -30,11 +30,12 @@ templates:
         % if int(ndim)==0:
         self.${id} = ${id} = fec.dummy_encoder_make(${framebits})\
         % elif int(ndim)==1:
-        self.${id} = ${id} = list(map((lambda a: fec.dummy_encoder_make(${framebits})),\
```
jdemel: I'd leave in this space. That'll result in properly formatted Python code.

gr-fec/grc/variable_dummy_encoder_def_list.block.yml
```diff
@@ -30,11 +30,12 @@ templates:
...
+        self.${id} = ${id} = list(map((lambda a:fec.dummy_encoder_make(${framebits})),\
         range(0,${dim1})))
         % else:
-        self.${id} = ${id} = list(map((lambda b: map((lambda\
-        a: fec.dummy_encoder_make(${framebits})), range(0,${dim2}))), range(0,${dim1})))\
+        self.${id} = ${id} = list(map((lambda b:list(map((lambda a:\
```
jdemel: I suggest a space in `lambda a: \` in order to have properly formatted.

grc/core/blocks/block.py
```diff
@@ -86,9 +86,28 @@ def __init__(self, parent):
...
+        self.current_bus_structure = {'source': '', 'sink': ''}
+
+    def get_bus_structure(self, direction):
+        if direction == 'source':
+            bus_structure = self.bus_structure_source
+        else:
+            bus_structure = self.bus_structure_sink
+
+        if not bus_structure:
+            return ''  # TODO: Don't like empty strings. should change this to None eventually
```
jdemel: Return `None` if this is the proper way forward. Do not leave this TODO here.

grc/core/blocks/block.py
```diff
@@ -86,9 +86,28 @@ def __init__(self, parent):
...
+
+        try:
+            clean_bus_structure = self.evaluate(bus_structure)
+            return clean_bus_structure
+        except:
+            return ''
```
jdemel: This should return `None` as well I assume.

grc/core/ports/port.py
```diff
@@ -39,14 +39,22 @@ class Port(Element):
...
+                 vlen='', multiplicity=1, optional=False, hide=False, bus_struct=None, **_):
         """Make a new port from nested data."""
         Element.__init__(self, parent)
 
         self._dir = direction
         self.key = id
         if not label:
             label = id if not id.isdigit() else {'sink': 'in', 'source': 'out'}[direction]
+        if dtype == 'bus':
+            #TODO - add some error checking in here
```
jdemel: Well, do so. What should be checked here?

mormj:
@jdemel - thank you for the review.  The busports part from grc got merged as part of another PR, so I will clean up those todos in a separate PR, and rebase this PR to resolve conflicts and update the gr-fec part.

jdemel:
Why do you remove `ber_curve_gen_ldpc`? Is this example broken beyond repair?

mormj:
I don't believe ber_curve_gen_ldpc was deleted in this PR.  My intention was just to trim down this one to the yml changes, but it looks like the .grc got converted to yml here.  _ldpc.grc was working for me with this PR.

jdemel:
Oh, I thought so because of the GitHub interface. It says ~1.5k lines deleted. But no new lines added for that file. Thus, I assumed it was deleted. Though, this is not the case but it is converted. That's great.
Did you also convert the other ber_curve_gen examples? It would be great to do so with this PR. Fits right in.

mormj:
Looks like ber_curve_gen.grc got updated and merged in with #2598 
Just pushed a commit with the other 2 (polar and tpc) updated

jdemel:
That's great. This should fix almost conclude the GRC examples conversion in `gr-fec`. At least the ones that depend on bus ports. Thanks!

Issue #2619, opened 2019-07-13
grc: adding default header variable produces a lot errors


dl1ksv:
Adding the default header variable object to a flowgraph produces the following errors:

ERROR:gnuradio.grc.core.FlowGraph:Failed to evaluate variable block variable_header_format_default_0
Traceback (most recent call last):
  File "/usr/local/gnuradio/lib/python3.7/dist-packages/gnuradio/grc/core/FlowGraph.py", line 226, in renew_namespace
    value = eval(variable_block.value, namespace, variable_block.namespace)
  File "<string>", line 1, in <module>
  File "/usr/local/gnuradio/lib/python3.7/dist-packages/gnuradio/digital/digital_swig2.py", line 1677, in make
    return _digital_swig2.header_format_default_make(access_code, threshold, bps)
ValueError: invalid null reference in method 'header_format_default_make', argument 1 of type 'std::string const &'
Traceback (most recent call last):
  File "/usr/local/gnuradio/lib/python3.7/dist-packages/gnuradio/grc/gui/VariableEditor.py", line 195, in set_value
    evaluated = str(block.params['value'].evaluate())
KeyError: 'value'
Traceback (most recent call last):
  File "/usr/local/gnuradio/lib/python3.7/dist-packages/gnuradio/grc/gui/VariableEditor.py", line 195, in set_value
    evaluated = str(block.params['value'].evaluate())
KeyError: 'value'

The KeyError is correct as this variable contains no parameter named value !

As far as I can see all other variable block have a parameter named value that is used in the value assignmend of the block.

**Question:** Must all variable blocks must have a parameter named value ?

Despite the error meassage the flowgraph can be generated and the generated code seems to be correct

haakov:
Is this issue fixed?

dl1ksv:
Yes, this issue is fixed with #2662 
But #2605 is still open and should be applied.

PR #2620, opened 2019-07-13
Adding value parameter to variable variable_header_format_default

dl1ksv:
Adding this block to a flowgraph produces error messages. 
See #2619 
This proposal comes accross this problem

### DIFF
--- a/gr-digital/grc/variable_header_format_default.block.yml
+++ b/gr-digital/grc/variable_header_format_default.block.yml
@@ -15,7 +15,12 @@ parameters:
     label: Payload Bits per Symbol
     dtype: int
     default: 1
-value: ${ digital.header_format_default(access_code, threshold, bps) }
+-   id: value
+    label: Resulting access code
+    dtype: string
+    default: 'digital.header_format_default(access_code, threshold, bps)'
+    hide: 'all'
+value: ${ value }
 
 templates:
     imports: from gnuradio import digital
--- a/gr-digital/grc/variable_header_format_default.block.yml
+++ b/gr-digital/grc/variable_header_format_default.block.yml
@@ -25,7 +25,7 @@ value: ${ value }
 templates:
     imports: from gnuradio import digital
     var_make: |-
-        % if int(eval(access_code))==0:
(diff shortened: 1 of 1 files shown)


gr-digital/grc/variable_header_format_default.block.yml
```diff
@@ -15,12 +15,17 @@ parameters:
...
+    dtype: string
+    default: 'digital.header_format_default(access_code, threshold, bps)'
+    hide: 'all'
+value: ${ value }
 
 templates:
     imports: from gnuradio import digital
     var_make: |-
-        % if int(eval(access_code))==0:
+        % if access_code == '0':
```
jdemel: What happens if one puts `0000`? Should this result in an all zero access code? Or should this also trigger the default access code? At least this does change the blocks behavior.

gr-digital/grc/variable_header_format_default.block.yml
```diff
@@ -32,7 +37,7 @@ cpp_templates:
     includes: ['#include <gnuradio/digital/header_format_default.h>']
     declarations: 'digital::header_format_default::sptr ${id};'
     var_make: |-
-        % if int(eval(access_code))==0:
+        % if access_code == '0' :
```
jdemel: See line 28 comment.

gr-digital/grc/variable_header_format_default.block.yml
```diff
@@ -15,12 +15,17 @@ parameters:
     label: Payload Bits per Symbol
     dtype: int
     default: 1
-value: ${ digital.header_format_default(access_code, threshold, bps) }
```
jdemel: I assume this is the actual change, that should fix this issue?

dl1ksv:
But as there exist about 10 block that own the same 
KeyError: 'value' problem, I think this problem should be fixed inside grc and not by modifying the yml description.
I'm searching for a solution, but I've not finished yet. You can find some thougths at the end of #2605

dl1ksv:
This issue should better be fixed inside grc. See #2662

PR #2622, opened 2019-07-13
rebased #2598: grc: add busports back into 3.8

### LINKED ISSUE #2277: GRC: bus ports missing

marcusmueller:
@mormj work, just rebased onto origin/master.
<hr>
Bus ports had not been added back in since the refactor of grc
Hopefully this fully enables busports though there are still some
issues with the gr-fec flowgraphs

Fixes #2277

### DIFF
--- a/gr-blocks/grc/blocks_null_sink.block.yml
+++ b/gr-blocks/grc/blocks_null_sink.block.yml
@@ -21,7 +21,7 @@ parameters:
     dtype: int
     default: '1'
     hide: part
--   id: bus_conns
+-   id: bus_structure_sink
     label: Bus Connections
     dtype: raw
     default: '[[0,],]'
--- a/gr-fec/grc/tpc_decoder_def_list.block.yml
+++ b/gr-fec/grc/tpc_decoder_def_list.block.yml
@@ -65,9 +65,11 @@ templates:
         % if int(ndim)==0:
         self.${id} = ${id} = fec.tpc_decoder_make(${row_poly}, ${col_poly}, ${krow}, ${kcol}, ${bval}, ${qval}, ${max_iter}, ${decoder_type})
         % elif int(ndim)==1:
-        self.${id} = ${id} = list(map( (lambda a: fec.tpc_decoder_make(${row_poly}, ${col_poly}, ${krow}, ${kcol}, ${bval}, ${qval}, ${max_iter}, ${decoder_type})), range(0,${dim1}) ))
+        self.${id} = ${id} = list(map( (lambda a: fec.tpc_decoder_make(${row_poly}, ${col_poly}, ${krow}, ${kcol}, ${bval}, ${qval}, \
+        ${max_iter}, ${decoder_type})), range(0,${dim1}) ))
         % else:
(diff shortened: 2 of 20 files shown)


marcusmueller:
@mormj, please check whether the ports.py is like you intended

grc/core/ports/port.py
```diff
@@ -39,14 +39,22 @@ class Port(Element):
...
-                 vlen='', multiplicity=1, optional=False, hide=False, **_):
+                 vlen='', multiplicity=1, optional=False, hide=False, bus_struct=None, **_):
         """Make a new port from nested data."""
         Element.__init__(self, parent)
 
         self._dir = direction
         self.key = id
         if not label:
             label = id if not id.isdigit() else {'sink': 'in', 'source': 'out'}[direction]
+        if dtype == 'bus':
```
mormj: This is the way I intend ports.py to be - the conflict was line 49 - the previous change of removing "+id" to the label was preserved

mormj:
@marcusmueller - yes ports.py has the merge as intended.

gr-blocks/grc/blocks_null_sink.block.yml
```diff
@@ -21,7 +21,7 @@ parameters:
     dtype: int
     default: '1'
     hide: part
--   id: bus_conns
+-   id: bus_structure_sink
```
mormj: explicitly calling out as bus_structure_sink, as that is what the underlying property is.  Also a good pattern for blocks that have both sink and source

gr-fec/grc/tpc_decoder_def_list.block.yml
```diff
@@ -65,9 +65,11 @@ templates:
         % if int(ndim)==0:
         self.${id} = ${id} = fec.tpc_decoder_make(${row_poly}, ${col_poly}, ${krow}, ${kcol}, ${bval}, ${qval}, ${max_iter}, ${decoder_type})
         % elif int(ndim)==1:
-        self.${id} = ${id} = list(map( (lambda a: fec.tpc_decoder_make(${row_poly}, ${col_poly}, ${krow}, ${kcol}, ${bval}, ${qval}, ${max_iter}, ${decoder_type})), range(0,${dim1}) ))
```
mormj: There are many such issues with the yml in gr-fec that cause the passed objects not to be understood.  
Mostly fixed by adding list(map(...))  and some line continuation issues such as lambda\a

grc/blocks/grc.tree.yml
```diff
@@ -4,6 +4,10 @@
   - pad_sink
   - virtual_source
   - virtual_sink
+  - bus_sink
```
mormj: Not sure if these lines need to be put back here but in adding back busports functionality, copied a bunch of code from 3.7

PR #2626, opened 2019-07-15
Be specific about the versions of Python that are usable

### LINKED ISSUE #2625: actually use the full minimum required versions for Python
* grc/CMakeLists requires a specific sub-version of Py2.7. We should be using that as Python version requirement.
* Check whether the same applies to Py3

marcusmueller:
i.e. add the third version depth as requirement.

Fixes #2625


grc/CMakeLists.txt
```diff
@@ -25,9 +25,9 @@ include(GrPython)
 message(STATUS "")
 
 GR_PYTHON_CHECK_MODULE_RAW(
-    "python2 >= 2.7.6 or python3 >= 3.4.0"
+    "python2 >= 2.7.6 or python3 >= 3.6.6"
```
mbr0wn:
```suggestion
    "python2 >= 2.7.6 or python3 >= 3.6.5"
```

grc/CMakeLists.txt
```diff
@@ -25,9 +25,9 @@ include(GrPython)
...
     "import sys; \
-     requirement = (3, 4, 0) if sys.version_info.major >= 3 else (2, 7, 6); \
+     requirement = (3, 6, 6) if sys.version_info.major >= 3 else (2, 7, 6); \
```
mbr0wn:
```suggestion
     requirement = (3, 6, 5) if sys.version_info.major >= 3 else (2, 7, 6); \
```

grc/CMakeLists.txt
mbr0wn:
```suggestion
    "python2 >= ${GR_PYTHON_MIN_VERSION} or python3 >= ${GR_PYTHON3_MIN_VERSION}"
```

Issue #2489, opened 2019-05-17
gr_modtool python tagged_stream block fails


mormj:
Issue raised by Sebastian.  

gr_modtool consistently fails when trying to create a new tagged stream block of type python.  This is with master (3.8) branch.  Also, it forces you to enter valid argument list, but that looks to be a separate issue.
```

josh@josh-p51:~/gr-tmp$ gr_modtool add -t tagged_stream -l python foo
GNU Radio module name identified: tmp
Language: Python
Block/code identifier: foo
Please specify the copyright holder: 
Enter valid argument list, including default arguments: 

Enter valid argument list, including default arguments: 

Enter valid argument list, including default arguments: 
0
Add Python QA code? [Y/n] n
Adding file 'python/foo.py'...
Traceback (most recent call last):
  File "/share/gnuradio/grdev/bin/gr_modtool", line 30, in <module>
    cli()
  File "/usr/local/lib/python3.6/dist-packages/click/core.py", line 764, in __call__
    return self.main(*args, **kwargs)
  File "/usr/local/lib/python3.6/dist-packages/click/core.py", line 717, in main
    rv = self.invoke(ctx)
  File "/usr/local/lib/python3.6/dist-packages/click/core.py", line 1137, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/local/lib/python3.6/dist-packages/click/core.py", line 956, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/local/lib/python3.6/dist-packages/click/core.py", line 555, in invoke
    return callback(*args, **kwargs)
  File "/share/gnuradio/grdev/lib/python3.6/dist-packages/gnuradio/modtool/cli/base.py", line 149, in wrapper
    return func(*args, **kwargs)
  File "/share/gnuradio/grdev/lib/python3.6/dist-packages/gnuradio/modtool/cli/add.py", line 81, in cli
    run(self)
  File "/share/gnuradio/grdev/lib/python3.6/dist-packages/gnuradio/modtool/cli/base.py", line 168, in run
    module.run()
  File "/share/gnuradio/grdev/lib/python3.6/dist-packages/gnuradio/modtool/core/add.py", line 144, in run
    self._run_python()
  File "/share/gnuradio/grdev/lib/python3.6/dist-packages/gnuradio/modtool/core/add.py", line 275, in _run_python
    self._write_tpl('block_python', self.info['pydir'], fname_py)
[...]

PR #2633, opened 2019-07-17
modtool: prevent tagged_stream under python

mormj:
Currently tagged stream blocks are not implemented
for python in modtool.  There is a nasty exception
that happens, so this is intended to validate
and throw the appropriate exception before
trying to generate the block

Addresses #2489

### DIFF
--- a/gr-utils/python/modtool/core/add.py
+++ b/gr-utils/python/modtool/core/add.py
@@ -67,6 +67,8 @@ class ModToolAdd(ModTool):
             raise ModToolException('Programming language not specified.')
         if self.info['lang'] not in self.language_candidates:
             raise ModToolException('Invalid programming language.')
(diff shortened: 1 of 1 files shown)


gr-utils/python/modtool/core/add.py
```diff
@@ -127,10 +129,10 @@ def _write_tpl(self, tpl, path, fname):
 
     def run(self):
         """ Go, go, go. """
-        # This portion will be covered by the CLI
-        if not self.cli:
```
swap-nil7: In my opinion, we should not validate the fields that are already verified through the CLI. Rather, we should raise similar exception in cli version of `add.py` after taking the input for language.

mbr0wn: Don't want to brush you off @swap-nil7, but we need to move forward on this PR and its harmless enough.

PR #2640, opened 2019-07-18
cmake: fix install dir of runtime_swig

michaelld:
subject says it all. simple and easy. Toby will have to figure out how to work around this change!

### DIFF
--- a/gnuradio-runtime/swig/CMakeLists.txt
+++ b/gnuradio-runtime/swig/CMakeLists.txt
@@ -62,7 +62,7 @@ GR_SWIG_MAKE(runtime_swig runtime_swig.i)
 install(
   TARGETS runtime_swig
   EXPORT runtime_swig-export
-  DESTINATION ${GR_PYTHON_RELATIVE}/gnuradio/gr
+  DESTINATION ${GR_PYTHON_DIR}/gnuradio/gr
   )
 
 include(GrPython)


michaelld:
Ref: https://github.com/gnuradio/gnuradio/commit/5ae09071f9503edb13eccb0aefc72856a35a801a

michaelld:
The underlying issue is that -all- of the other SWIG and Python scripts are installed somewhere inside `GR_PYTHON_DIR`, while just this one SWIG file is installed into `GR_PYTHON_RELATIVE` ... and for some systems, these are entirely different directories that will result in Python no finding this one SWIG file. All SWIG installs must be to the same base directory, regardless of whether is is relative or absolute ... CMake should properly handle either properly!

marcusmueller:
this is unrelated, but our ubuntu 18.04 py3/qt5 builder fails at 38 - qa_volk_32f_log2_32f. 

Um, @noc0lour?

noc0lour:
```
_avx2 completed in 0.089164 ms
offset 90473 in1: -inf in2: -127 tolerance was: 1e-05
volk_32f_log2_32f: fail on arch a_avx2_fma
offset 90473 in1: -inf in2: -127 tolerance was: 1e-05
volk_32f_log2_32f: fail on arch a_avx2
offset 90473 in1: -inf in2: -127 tolerance was: 1e-05
volk_32f_log2_32f: fail on arch a_sse4_1
offset 90473 in1: -inf in2: -127 tolerance was: 1e-05
volk_32f_log2_32f: fail on arch u_sse4_1
offset 90473 in1: -inf in2: -127 tolerance was: 1e-05
volk_32f_log2_32f: fail on arch u_avx2_fma
offset 90473 in1: -inf in2: -127 tolerance was: 1e-05
volk_32f_log2_32f: fail on arch u_avx2
Best aligned arch: a_avx2
Best unaligned arch: u_avx2
```
looks like a bug.

PR #2644, opened 2019-07-20
xml2yaml: convert more GRC examples to YAML format

jdemel:
This PR should make some progress on #2228 .
Although, this does reveal some problems.
- Some examples in `gr-fec` are still not fixed because of bus ports. 
- `gr-fec/examples/fecapi_tagged_decoders.grc` causes GRC to throw an error. 
- `gr-trellis` examples rely on a removed block `blks2_rate_error`.

The only examples that are not converted yet should be in `gr-fec` and `gr-trellis`. All other examples should be done now.

GRC did not open some files in `gr-digital` because it threw an error.

> Error: (ValueError("invalid literal for int() with base 10:
> 'digital.packet_utils.default_access_code'",), '% if
> int(eval(access_code))==0:\nself.${id} = ${id} =
> digital.header_format_default(digital.packet_utils.default_access_code,\\\n${threshold},
> ${bps})\n% else:\nself.${id} = ${id} =
> digital.header_format_default(${access_code},\\\n${threshold}, ${bps})\n% endif')

This could be fixed by manually editing the corresponding GRC files. All uses of `digital.packet_utils.default_access_code` are replaced with a string. This allows GRC to open these files. Still, GRC will hang in case `digital.packet_utils.default_access_code` is used again. This hints at a bug with the `default_access_code` code.

The conversion of the `gr-channel` example may have revealed a bug. The channel model block does
not work with a Rician channel model but only with a Rayleigh channel model.

### DIFF
--- a/gr-channels/examples/channel_tone_response.grc
+++ b/gr-channels/examples/channel_tone_response.grc
@@ -1,762 +1,418 @@
-<?xml version='1.0' encoding='ASCII'?>
-<flow_graph>
-  <timestamp>Fri Feb 15 15:13:55 2013</timestamp>
-  <block>
-    <key>import</key>
-    <param>
-      <key>id</key>
-      <value>import_0</value>
-    </param>
-    <param>
-      <key>_enabled</key>
-      <value>True</value>
-    </param>
-    <param>
-      <key>import</key>
-      <value>import math</value>
-    </param>
-    <param>
-      <key>_coordinate</key>
-      <value>(181, 10)</value>
-    </param>
-    <param>
-      <key>_rotation</key>
-      <value>0</value>
-    </param>
-  </block>
-  <block>
-    <key>variable</key>
-    <param>
-      <key>id</key>
-      <value>samp_rate</value>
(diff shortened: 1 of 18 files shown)


marcusmueller:
> Some examples in gr-fec are still not fixed because of bus ports.

can you elaborate on this? I thought we just fixed bus ports.

haakov:
PR for the gr-digital bug: https://github.com/gnuradio/gnuradio/pull/2620

jdemel:
> > Some examples in gr-fec are still not fixed because of bus ports.
> 
> can you elaborate on this? I thought we just fixed bus ports.

I didn't figure out how to use bus ports in 3.8. I couldn't find a button. I assume, I was just blind. The other issue is documented in #2285 .

haakov:
I believe the bus port buttons can be found at the bottom of the pop-up menu when you right-click a block

mbr0wn:
This PR is progress with or without the bus port GRCs, so it can go in.
Commits are cheap!

PR #2648, opened 2019-07-21
Add some documentation to squelch blocks

jmendeth:
Explain `ramp` parameter better, and add GRC documentation to the Power Squelch block.

### DIFF
--- a/gr-analog/grc/analog_pwr_squelch_xx.block.yml
+++ b/gr-analog/grc/analog_pwr_squelch_xx.block.yml
@@ -53,4 +53,17 @@ cpp_templates:
         True: true
         False: false
 
+documentation: |-
+    This will either pass the input unchanged or block it, depending on
+    whether its envelope is over a certain threshold. The envelope is calculated
+    by taking the squared magnitude of the signal and lowpassing it with a
+    single pole IIR filter (with the specified alpha).
+
+    The ramp parameter specifies the attack / release time, in samples.
+    A sinusodial ramp is used to progressively mute / unmute the input.
+    If ramp is set to 0, input is muted / unmuted without a ramp.
+
+    If the gate parameter is true no output will be produced when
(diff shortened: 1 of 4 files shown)


Notou:
While you wait for the PR to be reviewed, may I suggest you add your documentation to the block pages on the [wiki](https://wiki.gnuradio.org/index.php?search=squelch&title=Special%3ASearch&fulltext=1)?

777arc:
@Notou That's why Marcus assigned me to review it, to make sure the change also gets updated on the wiki.  Moving forward, any PRs that change the usage of blocks, or the block's documentation itself (unless its purely developer-centric) will have to be reflected in the wiki.  These kind of changes don't happen much so this is perfectly fine with me.  It's just a small price to pay in order to have our User Documentation live in the wiki and be really easy for people to read and edit.

777arc:
@marcusmueller I've updated the wiki, feel free to merge.

jmendeth:
I wasn't aware of the wiki, thanks @777arc for updating it :)

jmendeth:
Hmm, today I noticed a little inconsistency. The documentation says `gate` is false by default.
This was copied straight from the C++ docs, and it is true:

https://github.com/gnuradio/gnuradio/blob/c57004b61d8dac8958995b535a225b81d2b12a3d/gr-analog/include/gnuradio/analog/pwr_squelch_cc.h#L61-L62

But when you create the block in GRC, it sets `gate` to Yes by default.
This is odd, most people would expect `gate = false` behaviour by default, could this be changed?

777arc:
Yeah, weird.  I would say don't change the functionality, just update the docs.

jmendeth:
Done!

777arc:
I've updated it on the wiki as well, feel free to merge @marcusmueller

mbr0wn:
We let this get a bit stale -- merged and resolved manually.

Issue #2021, opened 2018-09-20
QTGui Flowgraphs: If no graphical sink present, don't display window


marcusmueller:
Since I can think of it, all GRC-generated Qt GUIs have displayed a window, regardless of whether there's actually something to display.

We should consider hiding the window in case of no GUI elements being used.

I'll look at this after GRCon

marcusmueller:
Haven't done that, so still not done

willcode:
This is the equivalent of automatically detecting whether to output a qtgui or console app. Should this be automatic instead of a manual selection? If manual, I'd argue that having the window pop up reminds people that they've made a qtgui app.

marcusmueller:
That's a fair assessment, also, not keeping the window around might be the opposite of what people want if they want to integrate their own QT elements

PR #2650, opened 2019-07-21
Do not show a window if there are no qtgui blocks.

krk:
Fixes https://github.com/gnuradio/gnuradio/issues/2021

### DIFF
--- a/grc/core/generator/top_block.py
+++ b/grc/core/generator/top_block.py
@@ -31,6 +31,7 @@ class TopBlockGenerator(object):
 
         self._flow_graph = FlowGraphProxy(flow_graph)
         self._generate_options = self._flow_graph.get_option('generate_options')
+        self._hide_qt_window = self._flow_graph.get_option('hide_qt_window')
 
         self._mode = TOP_BLOCK_FILE_MODE
         # Handle the case where the directory is read-only
@@ -119,7 +120,8 @@ class TopBlockGenerator(object):
             'monitors': monitors,
             'generate_options': self._generate_options,
             'version': platform.config.version,
-            'catch_exceptions': fg.get_option('catch_exceptions')
+            'catch_exceptions': fg.get_option('catch_exceptions'),
+            'hide': self._hide_qt_window
         }
         flow_graph_code = python_template.render(
             title=title,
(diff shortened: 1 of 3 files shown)


grc/core/generator/top_block.py
```diff
@@ -111,14 +112,17 @@ def _build_python_code_from_template(self):
             elif key == 'epy_module':
                 src = block.params['source_code'].get_value()
                 output.append((file_path, src))
+            elif not hasQtGui and "qtgui" in key:
```
marcusmueller: this is a bit redundant: the first condition `not hasQtGui` could be omitted.

Anyway, this is a bit *too* shallow for actually checking for whether a QtGui element exists in a flow graph. We're not forcing anyone to use "qtgui" in the names of their graphical sinks, and there's multiple out-of-tree visualization sinks that don't contain that string.

krk: That makes sense, I have only checked the in-tree blocks to see if they have something in common and that is how I arrived at name having "qtgui".

What about checking the python import statements to see if they contain `from PyQt5 import Qt`?

marcusmueller: Hm, that's a very hard question. Maybe we should be forcing people to explicitly declare their blocks as having QT windows, simply by having a new `flag` for the YAML files. That would make your life a lot easier here, wouldn't it?

mbr0wn:
I would love this behaviour. All of @marcusmueller 's points are valid too though.

nickoe:
@krk, do you intend to follow up on this pull request?

krk:
@nickoe Thank you for the ping. I have added a `hide_qt_window` option instead and rebased on the latest master.

grc/core/generator/top_block.py
sdh11: Qt blocks get added to the `top_grid_layout` in the flowgraph object in order to show up in the window. We can check the `count` of the layout to see if any blocks have been added to it and if not, hide the window. Unless someone is manually modifying the generated python, that should have all the included blocks.

sdh11: Having the explicit option to override and hide the window even if blocks exist would be nice too.

nickoe:
@krk what do we gain from having this hide option when you could just select the "No GUI" option?

mormj:
I agree with @nickoe here - the real power in this feature will be to not open a window when there are no blocks. If there is a manual step involved, you may as well just select the no gui option.

nickoe:
@krk This pull request is becoming stale. It looks to me like you have misunderstood wish in #2021. The idea is to have it it _automatically_ detect if ti should how a window or now depending on any gui blocks being present and active.

mbr0wn:
Looks like the consensus here is that this is a suboptimal solution. @krk thanks for your contribution, hopefully we can see more!

PR #2655, opened 2019-07-22
Use raw strings for paths substituted directly into python strings.

ryanvolz:
Feeding CMake strings directly into Python strings can cause errors when the CMake strings contain unintentional escape sequences that then get interpreted by Python. This bit me when building on Windows where native paths use backslashes and I was unlucky enough to have a "\U" interpreted as a missing unicode character.

This patch makes the Python strings where this happens into raw strings so that the escape sequences are not unintentionally interpreted.


mbr0wn:
This should go in before the release, @marcusmueller .

PR #2656, opened 2019-07-22
Don't use hard-coded path constants.

ryanvolz:
I'm using this patch to get Windows builds to function correctly for my in-development [conda-forge feedstock](https://github.com/ryanvolz/gnuradio-feedstock/commits/prerelease). Conda-build requires that the built package be relocatable so that it can install to the path of a conda environment. To facilitate this, it can use string replacement in package files to replace the build prefix path with the install prefix path of the environment. Unfortunately, it does not do this replacement for binary files on Windows, so the use of hard-coded paths gnuradio-runtime/lib/constants.cc breaks the prefix/sysconf/preferences paths on Windows using a conda package.

The only sensible strategy that I could think of to handle this situation was to ditch hard-coded paths and devise a runtime prefix determination scheme. I'm not sure if this is something you'd want to adopt in the mainline, but I offer it up just in case.

This replaces the path constants for the prefix, system configuration, and preferences directories with a runtime prefix lookup based on the gnuradio-runtime library location and relative paths from that.

Also, use cmake-style paths instead of native paths for substitution to avoid unescaped character problems. For relocatable prefix string replacement, conda always substitutes in forward-slash paths anyway, so it's just easier to use them from the start. As far as I can tell, Windows doesn't actually care that forward slashes are used (and indeed, they are already used in some parts of the GNU Radio code).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -309,12 +309,20 @@ endif(ENABLE_PERFORMANCE_COUNTERS)
 ########################################################################
 # Variables replaced when configuring the package config files
 ########################################################################
-file(TO_NATIVE_PATH "${CMAKE_INSTALL_PREFIX}"           prefix)
-file(TO_NATIVE_PATH "\${prefix}"                        exec_prefix)
-file(TO_NATIVE_PATH "\${exec_prefix}/${GR_LIBRARY_DIR}" libdir)
-file(TO_NATIVE_PATH "\${prefix}/${GR_INCLUDE_DIR}"      includedir)
-file(TO_NATIVE_PATH "${SYSCONFDIR}"                     SYSCONFDIR)
-file(TO_NATIVE_PATH "${GR_PREFSDIR}"                    GR_PREFSDIR)
+file(TO_CMAKE_PATH "${CMAKE_INSTALL_PREFIX}"           prefix)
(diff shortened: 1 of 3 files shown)


jdemel:
Just the other day I had an issue with a path in my `$HOME/.gnuradio.config.conf` file. It pointed GRC to look in the wrong folder for installed blocks. Would this PR improve on that too? Having multiple PyBOMBS prefixes and a user config file seems to be problematic at the moment. It would be great to improve this situation.

ryanvolz:
> Just the other day I had an issue with a path in my `$HOME/.gnuradio.config.conf` file. It pointed GRC to look in the wrong folder for installed blocks. Would this PR improve on that too? Having multiple PyBOMBS prefixes and a user config file seems to be problematic at the moment. It would be great to improve this situation.

I don't think it improves the current situation in that case, but perhaps it could help in coming up with an acceptable fix. A block path provided in `$HOME/.gnuradio/config.conf` file will always override (or add to?) the default block path as far as I understand it, so having that config file picked up when using multiple prefixes will always present a problem with absolute paths or mixed GR versions. Maybe paths relative to the prefix of the running GR could be used?

marcusmueller:
Since this compiles on none of our native platforms, I'll have to ask you to resubmit this PR once it doesn't break on a major Linux distro.

PR #2662, opened 2019-07-24
KeyError patch for variable blocks 

### LINKED ISSUE #2494: grc fails to build constellation object if constellation != variable constellation

dl1ksv:
Together with #2605 this fixes #2494 and KeyErrors in other variable blocks

### DIFF
--- a/grc/core/FlowGraph.py
+++ b/grc/core/FlowGraph.py
@@ -223,8 +223,11 @@ class FlowGraph(Element):
         # Load variables
         for variable_block in self.get_variables():
             try:
+                variable_block.rewrite()
                 value = eval(variable_block.value, namespace, variable_block.namespace)
                 namespace[variable_block.name] = value
+            except TypeError: #Type Errors may happen, but that desn't matter as they are displayed in the gui
+                pass
             except Exception:
                 log.exception('Failed to evaluate variable block {0}'.format(variable_block.name), exc_info=True)
                 pass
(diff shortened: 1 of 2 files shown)


mormj:
Pulled and tested this and re-ran a bunch of the digital example flowgraphs - appears to fix the underlying issue

grc/core/FlowGraph.py
```diff
@@ -223,8 +223,11 @@ def renew_namespace(self):
         # Load variables
         for variable_block in self.get_variables():
             try:
+                variable_block.rewrite()
```
mormj: What made the variable_block need the rewrite at this point?

dl1ksv: Otherwise variable_block.namespace is 'invalid' and the following eval fails.
After the rewrite the variable_block.namespace is valid and the eval works as expected.

block.namespace is defined as
`    def namespace(self):
        return {key: param.get_evaluated() for key, param in six.iteritems(self.params)}
`
Without the rewrite param.get_evaluated() returns None

dl1ksv:
#2605 should be applied, too

PR #2663, opened 2019-07-24
Using correct signal in freqdisplayform.cc

dl1ksv:
Running gr-filter/examples/filter_taps.grc
leads to
QObject::connect: No such signal FFTSizeMenu::whichTrigger(unsigned int)

This proposed change fixes this issue


marcusmueller:
merging this, since it fixes the acute issue. Underlying issue is still confusing: some `whichTrigger` are `int`, others are `unsigned int`, and the latter would be the correct choice, and consistency would be most desirable.

PR #2670, opened 2019-07-25
qtgui: restore enums/defaults for QT GUI Time Sink

mormj:
Somewhere along the line, the QT Gui Time sink lost enums in .grc going to yml.  Also some of the 3.8 color and marker defaults were not ideal, though probably need for discussion here - resets to 3.7 color/marker defaults.  The enums I think are necessary for usability.

- moves the line style parameters to the Config Tab
- fixes issue with green becoming the second color

image

### DIFF
--- a/gr-qtgui/grc/qtgui_time_sink_x.block.yml
+++ b/gr-qtgui/grc/qtgui_time_sink_x.block.yml
@@ -1,6 +1,5 @@
 id: qtgui_time_sink_x
 label: QT GUI Time Sink
-flags: [ python ]
 
 parameters:
 -   id: type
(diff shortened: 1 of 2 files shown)


haakov:
Looks like the removal of enums was deliberate: https://github.com/gnuradio/gnuradio/commit/f19899328b9bd96ce97b9c18e1a9966ae2e26023

mormj:
Interesting - just wanted to throw it out there - the lack of enums makes usability quite difficult in grc, so I'm curious if the removal was deliberate - @mbr0wn

mbr0wn:
I can't remember why they got removed The commit that @haakov mentions was supposed to fix multiple channels > 2, but the enums are unrelated. I think this is fine.

PR #2673, opened 2019-07-27
variable_header_default correctly check access_code

dl1ksv:
int(eval(access_code)) fails with 
ValueError: invalid literal for int() with base 10
if access_code does not simply consists of digits.
The main purpose of this check is to use digital.packet_utils.default_access_code as access_code  if access_code  == 0
which is the default setting.
This patch uses digital.packet_utils.default_access_code as default but sets the access code to $access_code if this code consits of digits.

This fixes the poblems described in #2644

### DIFF
--- a/gr-digital/grc/variable_header_format_default.block.yml
+++ b/gr-digital/grc/variable_header_format_default.block.yml
@@ -6,7 +6,7 @@ parameters:
 -   id: access_code
     label: Access Code
     dtype: string
-    default: '0'
+    default: digital.packet_utils.default_access_code
 -   id: threshold
     label: Threshold
     dtype: int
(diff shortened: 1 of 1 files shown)


gr-digital/grc/variable_header_format_default.block.yml
```diff
@@ -20,25 +20,25 @@ value: ${ digital.header_format_default(access_code, threshold, bps) }
...
     declarations: 'digital::header_format_default::sptr ${id};'
     var_make: |-
-        % if int(eval(access_code))==0:
-        this->${id} = ${id} = digital::header_format_default(digital.packet_utils.default_access_code,\
+        % if access_code.isdigit():
+        this->${id} = ${id} = digital::header_format_default(${access_code},\
         ${threshold}, ${bps});
         % else:
-        this->${id} = ${id} = digital::header_format_default(${access_code},\
+        this->${id} = ${id} = digital::header_format_default(digital.packet_utils.default_access_code,\
```
marcusmueller: This doesn't look like it'd produce valid C++. We'd have to `${..}` the python.

dl1ksv: @marcusmueller You are right, this does not produce valid c++ code, though this is the same code as before just in an other order.
At this level you can't evaluate digital.packet_utils.default_access_code as the digital package is not known here.

But I found out that there are some more issues to be solved in grc.
See #2688 
There happens a KeyError for variable blocks without a value param when trying to build the cpp code. This is similar to #2662 for python code.
I try to find a fix.

dl1ksv:
The python part is probably already fixed by #2844

Issue #2558, opened 2019-06-18
GRC: Options: Sensible defaults


marcusmueller:
@lscardoso gives a very nice intro tutorial; and he's having a lot of students using CorteXlab. 

He recommends, and I agree:

* Strongly encouraging people to fill in author, title

I'd also like to suggest that people should change the top block id so it's not `top_block` in 90% of cases.

Technically, I think a sane default would be:

* Prefill "author" by `$(whoami)`
* Setting Title to "Not titled yet"
* defaulting id to `default`, **but checking the `id` against `!= 'default'`**, refusing to generate if it's `default`. Upon saving the GRC flowgraph, automagically replace `default` with the file name.

marcusmueller:
This has been implemented, yaay!

Issue #2061, opened 2018-10-02
GRC parameter window's Apply button does not work properly


haakov:
The Apply button only works once for every time you open the parameter window.

Steps to reproduce:
- Add a block
- Open its parameter window
- Change a parameter and press Apply (this works)
- Change another parameter and press Apply again (this doesn't)

Update:
Actually, the Ok button does not work either once you've pressed Apply.

btashton:
Made some progress on this one today.  It looks like there are a couple things that are causing this behaviour.

haakov:
@btashton what did you find?

haakov:
Can't reproduce, closing

PR #2676, opened 2019-07-27
grc: Introduce sensible defaults

haakov:
Defaults as described in: https://github.com/gnuradio/gnuradio/issues/2558

**Bug warning:**
If you
1. create a new flow graph
2. save it without changing the ID (So that the ID is replaced by the file name)
3. open any block's parameter dialog
4. press "Cancel"

the Options block's ID will be reset to `default` again, and you will have to manually change the ID to be allowed to generate code from GRC.

I haven't had time to investigate why, but I believe this bug to be out of scope for this PR and perhaps related to https://github.com/gnuradio/gnuradio/issues/2061

### DIFF
--- a/grc/core/FlowGraph.py
+++ b/grc/core/FlowGraph.py
@@ -49,9 +49,9 @@ class FlowGraph(Element):
             the flow graph object
         """
         Element.__init__(self, parent)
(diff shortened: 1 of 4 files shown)


marcusmueller:
tested and merged

Issue #2646, opened 2019-07-20
CMake aborts when click isn't found, instead of disabling gr_modtool


marcusmueller:
## Issue Description ##

As noticed by Barry Dugan & @michaelld:

CMake aborts when click isn't found, instead of disabling gr_modtool. This happens even if nothing forces CMake to not simply disable gr-utils' modtool.

## Setup Details ##
current master, i.e. v3.8.0.0-rc2, without click installed. 

## Expected Behavior ##

Missing click is not a build-time dependency at all. It is a gr-utils test-time dependency, and a modtool run-time dependency, but if click is missing, CMake should simply run through.

## Actual Behavior ##

CMake aborts and you can't build GR if you don't have click installed.

## Steps to reproduce the problem ##

1. don't install click
2. cd build; cmake ..

PR #2682, opened 2019-07-29
cmake: Don't abort everything when modtool can't find click

haakov:
Fix for https://github.com/gnuradio/gnuradio/issues/2646: Modtool is simply disabled if `click` and `click-plugins` can't be found. With this fix, modtool can also be enabled and disabled with `-DENABLE_GR_MODTOOL=ON/OFF` like the top-level modules. Looks like it works to me, but I don't know if this is the best way to do it. @noc0lour thoughts?

### DIFF
--- a/gr-utils/CMakeLists.txt
+++ b/gr-utils/CMakeLists.txt
@@ -24,6 +24,18 @@ include(GrPython)
 
 GR_PYTHON_CHECK_MODULE("Mako >= ${GR_MAKO_MIN_VERSION}" mako "mako.__version__ >= '${GR_MAKO_MIN_VERSION}'" MAKO_FOUND)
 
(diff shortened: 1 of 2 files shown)


michaelld:
General question: I don't think modtool is required to build GR ... yes? If not, then why are we even checking for `click` here? Why not check when executing `modtool` itself & tell the user then of the deficiencies? I think this is consistent with other utilities installed by GR ... thinking a tool that requires SciPy, but since GR itself doesn't we don't check for it at configure or build, just runtime.

marcusmueller:
True, click shouldn't even be necessary for testing gr_modtool (because gr_modtool needs to work non-interactively even if click isn't found).

marcusmueller:
We should basically be removing the compile time checks alltogether, but for now, this is a clear improvement, so: merge

PR #2698, opened 2019-08-02
grc: prevent scrolling to top of canvas

### LINKED ISSUE #2660: GRC scrolls to top of canvas when regains focus
GRC keeps scrolling to the top of the canvas when clicking.  Mild annoyance

This occurs when clicking into the console or search tree then back into the canvas, the view is reset to the top left.  Particularly annoying when adding blocks from the search tree.

mormj:
GRC was scrolling to the top of the canvas when clicking into the console or search tree then back into the canvas. The view reset to the top left.

Commented out one line in _handle_mouse_button_press in DrawingArea.py -
makes the problem go away.  Need some additional testing to verify this won't cause any other negative side effects.  

Fixes #2660

### DIFF
--- a/grc/gui/DrawingArea.py
+++ b/grc/gui/DrawingArea.py
@@ -119,7 +119,12 @@ class DrawingArea(Gtk.DrawingArea):
         """
         Forward button click information to the flow graph.
         """
-        self.grab_focus()
+        # The following line causes the canvas to reset position to 0,0
+        #  when changing focus from the console or block search back to 
+        #  the canvas
+        #  Removing this line leaves the canvas alone when focus changes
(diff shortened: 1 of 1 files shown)


marcusmueller:
Since side-effects aren't clear, I'm putting this off to post-3.8.0.0, if that's OK with you – I wouldn't mind this being in 3.8.0.1 if it doesn't break anything else.

mormj:
I haven't observed any side effects yet, but that definitely makes sense to play it safe.

marcusmueller:
If you tested it, that's good enough then – it's a bug fix for a serious bug, and the worst it can break is about as bad as what if fixes.

PR #2703, opened 2019-08-05
Don't use hard-coded path constants.

ryanvolz:
Re-submission of #2656 with proper includes that should fix build on buildbot.

This replaces the path constants for the prefix, system configuration,
and preferences directories with a runtime prefix lookup based on the
gnuradio-runtime library location and relative paths from that.

Also use cmake-style paths instead of native paths for substitution to
avoid unescaped character problems. For relocatable prefix string
replacement, conda always substitutes in forward-slash paths anyway, so
it's just easier to use them from the start.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -309,12 +309,20 @@ endif(ENABLE_PERFORMANCE_COUNTERS)
 ########################################################################
 # Variables replaced when configuring the package config files
 ########################################################################
-file(TO_NATIVE_PATH "${CMAKE_INSTALL_PREFIX}"           prefix)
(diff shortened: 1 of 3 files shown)


marcusmueller:
Hey, sorry, this still doesn't build anywhere. Maybe we can help you set up a development system that allows you to quickly check buildability yourself?

ryanvolz:
Yeah, I'll have to get a clean environment set up with the right dependencies when I have a little more time. Based on these errors, though, I've figured out that this would require at least Boost 1.61. Since that's higher than what GR currently requires (1.53?), it may not be the right time for this change anyway.

PR #2716, opened 2019-08-09
Disable Python 2k

marcusmueller:
As announced, GNU Radio 3.9.x.x will no longer support Python2k.


marcusmueller:
The expected CI behaviour is that all Py2 builds fail

grc/CMakeLists.txt
```diff
@@ -25,9 +25,9 @@ include(GrPython)
 message(STATUS "")
 
 GR_PYTHON_CHECK_MODULE_RAW(
-    "python2 >= 2.7.6 or python3 >= 3.6.5"
+    "python3 >= 3.6.5"
```
mbr0wn: I guess we should be checking for GR_MIN_PYTHON here

cmake/Modules/GrSwig.cmake
```diff
@@ -118,12 +118,8 @@ macro(GR_SWIG_MAKE name)
     list(INSERT GR_SWIG_INCLUDE_DIRS 0 "$<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}>")
     list(INSERT GR_SWIG_INCLUDE_DIRS 0 "$<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}>")
 
-    if (PYTHON3)
-        set(py3 "-py3")
-    endif (PYTHON3)
-
     #setup the swig flags with flags and include directories
-    set(CMAKE_SWIG_FLAGS -fvirtual -modern -keyword -w511 -w314 -relativeimport ${py3} -module ${name} ${GR_SWIG_FLAGS})
+    set(CMAKE_SWIG_FLAGS -fvirtual -modern -keyword -w511 -w314 -relativeimport -module ${name} ${GR_SWIG_FLAGS})
```
mbr0wn: So this removes the `-py3` flag. Is that intentional?

grc/CMakeLists.txt
marcusmueller: true.

cmake/Modules/GrSwig.cmake
marcusmueller: huh. no, not actually, but: the testcases _work_ without the flag.

So. Maybe. yes?

grc/CMakeLists.txt
michaelld: technically `GR_PYTHON_MIN_VERSION` ...

grc/CMakeLists.txt
```diff
@@ -25,9 +25,9 @@ include(GrPython)
...
     "import sys; \
-     requirement = (3, 6, 5) if sys.version_info.major >= 3 else (2, 7, 6); \
+     requirement = (3, 6, 5); \
```
michaelld: Any way we can use the `GR_PYTHON_MIN_VERSION` here too? Better to automate this from a single variable than to have to manually tweak this every time a change is needed. IMHO.

michaelld:
I love that removing Py2 is actually pretty simple! A few more tweaks & I think this PR will be good to go!

cmake/Modules/GrSwig.cmake
michaelld: Hmmm ... SWIG3 docs on Py3 are < http://swig.org/Doc3.0/SWIGDocumentation.html#Python_python3support > ... doesn't say that `-py3` is a default, so I'm guessing if we don't include it then SWIG tries to generate code that is compliant with both Py2 and Py3. Adding in the flag adds Py3-only features ... so, thinking to just use the flag outright here is the right thing to do.

marcusmueller: no. no, removing that is not OK: `-py3` isn't documented via `--help`, or `man swig`, but it seems to [_do_](http://www.swig.org/Doc3.0/SWIGDocumentation.html#Python_python3support) things. So, adding this back.

grc/CMakeLists.txt
marcusmueller: Absolutely true!

cmake/Modules/GrPython.cmake
```diff
@@ -33,8 +33,7 @@ if (PYTHON_EXECUTABLE)
     find_package(PythonInterp ${GR_PYTHON_MIN_VERSION} REQUIRED)
 else (PYTHON_EXECUTABLE)
     message(STATUS "PYTHON_EXECUTABLE not set - using default python3")
-    message(STATUS "Use -DPYTHON_EXECUTABLE=/path/to/python2 to build for python2.")
-    find_package(PythonInterp ${GR_PYTHON3_MIN_VERSION} REQUIRED)
+    find_package(PythonInterp ${GR_PYTHON_MIN_VERSION} REQUIRED)
```
michaelld: OK so maybe this is a change for another PR ... but this "if" statement makes no sense now! It's doing a different status message then the same search. There has -got- to be a better way!

cmake/Modules/GrPython.cmake
```diff
@@ -33,8 +33,7 @@ if (PYTHON_EXECUTABLE)
...
 endif (PYTHON_EXECUTABLE)
 
 if (${PYTHON_VERSION_MAJOR} VERSION_EQUAL 3)
```
michaelld: Seems like this clause & the `PYTHON3` setting aren't needed any longer ... yes?

cmake/Modules/GrSwig.cmake
michaelld: `swig -python -help` shows me the `-py3` flag ...

cmake/Modules/GrPython.cmake
michaelld: looks like this would work ... haven't tested it yet though: `find_package(Python3 COMPONENTS Interpreter Development)`

marcusmueller: uuuuuh I like that!

cmake/Modules/GrPython.cmake
marcusmueller: yes!

cmake/Modules/GrPython.cmake
michaelld: Note that the variables returned by this command are `PYTHON3_*` ... so we'd want to also set `PYTHON_*` from them. Small price to pay ...

mbr0wn: I *have* tested for UHD, and for CMake 3.8, this won't work. It needs a newer version. I had to make a stupid compat layer in UHD to support both :angry:

michaelld: @mbr0wn ?meaning? ... not sure how this connects to the comment thread here ...

dkozel:
@marcusmueller With the current level of disabling/stripping out of Python 2 support we can merge this (once all in scope review comments are addressed) right? Removing six can be done in a second pass but we benefit immediately from enforcing a Python 3 only policy.

mbr0wn:
Leaving six in for a while will allow us to backport onto maint-3.8 more easily.

cmake/Modules/GrPython.cmake
mbr0wn: find_package(Python3) needs a version of CMake that we don't require.

mbr0wn: You're right on both accounts: This can be improved, and it can be done in another commit.

mbr0wn:
I fixed the comments and replaced this with https://github.com/gnuradio/gnuradio/pull/2997.

grc/CMakeLists.txt
mbr0wn: Fixed on the other PR.

grc/CMakeLists.txt
mbr0wn: Fixed on the other PR.

cmake/Modules/GrSwig.cmake
mbr0wn: Fixed on the other PR.

cmake/Modules/GrPython.cmake
mbr0wn: Fixed on the other PR.

PR #2728, opened 2019-08-12
Module name correction for bokehgui

Notou:
I finished porting [bokehgui](https://github.com/Inria-Maracas/gr-bokehgui/tree/master) to the 3.8 and python3.

The previous version imported the bokeh_layout module under an alias but I couldn't reproduce it, so I removed the alias and put the actual module name in its place.

I've tested it with python3 but not python2. I'm not sure if it would change something.


mormj:
@Notou -  Did the original fork for this PR get deleted?

Notou:
Apparently it did, I'm not sure why. I may have done something wrong, sorry. 
Should I do something about it?

Notou:
I've made a [new PR](https://github.com/gnuradio/gnuradio/pull/2809) for that, sorry for the inconvenience.

PR #2734, opened 2019-08-14
cmake: don't cache thrift found

bastibl:
If GNU Radio is compiled with Thrift, cmake fails in OOTs in all but the first run. AFAIS, Thirft libraries and paths are cached but not the target (not sure if this is possible at all). So cmake fails with

```
CMake Error at lib/CMakeLists.txt:36 (add_library):
  Target "gnuradio-foo" links to target "Thrift::thrift" but the target was
  not found.  Perhaps a find_package() call is missing for an IMPORTED
  target, or an ALIAS target is missing?
```

### DIFF
--- a/cmake/Modules/targetConfig.cmake.in
+++ b/cmake/Modules/targetConfig.cmake.in
@@ -21,6 +21,6 @@ include(CMakeFindDependencyMacro)
 
 set(target_deps "@TARGET_DEPENDENCIES@")
 foreach(dep IN LISTS target_deps)
-    find_dependency(${dep})
+    find_package(${dep})
 endforeach()
 include("${CMAKE_CURRENT_LIST_DIR}/@TARGET@Targets.cmake")


michaelld:
Fascinating! I really don't see how this change fixes the issue -- which I don't quite understand in the first place LOL! [not doubting you] I'm working on GR some today ... will try to test this issue & fix.

bastibl:
If GNU Radio is compiled with Thrift, gnuradio-runtime depends on the target Thrift::thrift
https://github.com/gnuradio/gnuradio/blob/master/gnuradio-runtime/lib/CMakeLists.txt#L177

This target is created by `find_package(thrift)` on the fly. This works in the first run. In subsequent runs, the target is not created since `Thrift_found` is cached. However, this means that only `Thrift_lirbraries` etc are valid but the target does not exist.

michaelld:
OK that explanation makes sense. Now that I have PRs #2738 and #2739 , I can easily work with this one. It seems like the way I replicate this is by building GR, then re-doing the CMake configuration without deleting the `CMakeCache.txt` file ... yes? Because then `THRIFT_FOUND` will be cached as `TRUE` and hence CMake should ignore the `FindThrift.cmake` script & hence the includes and libraries & such shouldn't be set. Does this sound correct?

bastibl:
For whatever reason, it only happens with OOTs. So use gr_modtool to create an OOT, add a C++ block (you don't have to fix anything), and run cmake twice.

michaelld:
> For whatever reason, it only happens with OOTs. So use gr_modtool to create an OOT, add a C++ block (you don't have to fix anything), and run cmake twice.

ACK. I need to create an OOT for testing for bug in "get_tags_in_range" anyway ... will test out this issue with it at the same time.

bastibl:
OK, I had another look at this. It's more complicated:

In-tree, we call `find_package(THRIFT)` in `runtime/lib` but also use `THRIFT_FOUND`, for example, in `runtime/python`. When we do not cache `THRIFT_FOUND`, it will not be available in `runtime/python` and the Python Thrift stuff is not installed.

Out-of-tree, we import the runtime component of GNU Radio, which depends on Thrift, if GNU Radio is compiled with Thrift support. `gnuradio-runtimeConfig.cmake` then iterates over all dependencies and searches them with `find_dependency`. (`find_dependency` is just a wrapper around `find_package` that is supposed to be used in `...Config.cmake` files. So we use it, as it is supposed to.) The problem is that `find_dependency` had a bug and did not call `find_package(foo)` if `FOO_FOUND` was already defined. This was not indented and is fixed in more recent versions of cmake (I think from 3.15):
https://gitlab.kitware.com/cmake/cmake/issues/17583
https://gitlab.kitware.com/cmake/cmake/merge_requests/3161
This means an OOT will call `find_dependency(THRIFT)` once and fail in subsequent runs, as the target is not defined (only `THRIFT_LIBRARIES` etc are available).

Since we require a minimum cmake version of 3.8, I think we have to workaround this. I don't know a good solution, but we could:
- move in-tree `find_package(THRIFT)` up in the hierarchy and not cache `THRIFT_FOUND`
- replace `find_dependency` with `find_package` in `gnuradio-<component>Config.cmake`
- change OOTs to call `find_package(THRIFT)` after `find_package(Gnuradio....)`

michaelld:
> In-tree, we call `find_package(THRIFT)` in `runtime/lib` but also use `THRIFT_FOUND`, for example, in `runtime/python`. When we do not cache `THRIFT_FOUND`, it will not be available in `runtime/python` and the Python Thrift stuff is not installed.

Right: This is why we decided to cache `THRIFT_FOUND` ... it's a hack to allow this variable to be used in multiple unrelated locations in the build. One option that would solve this particular issue would be to move Thrift finding up to the main `runtime` and disable the caching of `THRIFT_FOUND`. Not unreasonable since the value needs to be used in 2 separate directories that would both inherit from the main `runtime`.

michaelld:
I'm reinstalling the latest GR release +ctrlport_thrift ... will then be able to test out the rest of your comment about what happens when GR find happens in CMake.

michaelld:
OK: GR 3.7.13.5 +ctrlport_thrift is installed & I've verified that the `gnuradio-runtime` library links against Thrift.

Reviewing the CMake files installed by GR, I find no reference to Thrift beyond `FindThrift.cmake` via `grep -inr thrift` from that directory.

I have a GR OOT for testing this issue on, and I haven't found a way to replicate it.

I believe that, unlike `log4cpp` [1], `Thrift` usage is entirely embedded within the `GR-runtime` library. I believe that the GR thrift headers are not required to link against GR ... except maybe if +ctrlport_thrift is enabled in the OOT ... not even sure what that means nor how to do it!

I don't doubt that this is an issue: caching variables should be done cautiously, and `FOUND` variables should really never be cached because of how they are used within CMake. Hence, I'd welcome more explicit advice on how to test this issue.

[1] `log4cpp` is required for linking when logging using it is enabled, regardless of whether the OOT directly uses logging and/or log4cpp, purely because its header is included somewhere from the main GR headers ... something "we" should work on dealing with somehow to remove this dependency unless logging using log4cpp in the OOT is explicitly enabled.

bastibl:
I was talking about GR 3.8. GR 3.7.13.5 didn't have target configs AFAIS.

michaelld:
> I was talking about GR 3.8. GR 3.7.13.5 didn't have target configs AFAIS.

Ah OK. Let me get GR38 +ctrlport_thrift installed then ...

michaelld:
OK yeah I see what you mean ... so many issues with my GR38 install right now (via MacPorts) ... have to debug those first before I can deal with this PR.

michaelld:
Getting closer. Building GR right now with all of the fixes I know of for +ctrlport and +thrift and +pythion3 and whatever else I can come up with! Then I'll really be able to test out this PR. Whew!

michaelld:
Thanks for the new change. I was coming to the same conclusion ... other packages also exhibited strange behaviors like what Thrift was going. I'll check out your change & see if it helps in my testing ... hopefully it fixes things! I've got some other changes to the GR Cmake find top level script coming beyond this one ...

PR #2738, opened 2019-08-14
GR37 thrift fixes

michaelld:
Allows building GR37 +ctrlport +ctrlport::thrift using Thrift 0.9.2 through current (0.12.0). Builds and passes `make test` using Thrift 0.10.0, 0.11.0, and 0.12.0 on OSX.


michaelld:
Wishing there were a better way to handle the selection of either `std::shared_ptr` or `boost::shared_ptr`, but since c++11 is not guaranteed in GR37, using a macro is the best way I could figure out. [For GR38 where c++11 is required, we can use the c++11 syntax `using`.]

marcusmueller:
I quite like everything about this, even if it's very version-specific

michaelld:
> I quite like everything about this, even if it's very version-specific

I have a PR coming for GR38 ... hopefully tomorrow. Their code bases have diverged somewhat so I have to redo most of the patches to be GR-version specific.

michaelld:
Ping on this issue ... please let's deal with it before it gets stale!

noc0lour:
Closing in favor of rebased & reformatted #3869.

PR #2740, opened 2019-08-15
clang-tidy: Throw exceptions by value, catch by reference

### LINKED ISSUE #2708: Watch out for catch-by-value
```
gnuradio-runtime/lib/block_detail.cc:260:33: warning: catching polymorphic type ‘class std::runtime_error’ by value [-Wcatch-value=]
  260 |       catch (std::runtime_error e) {
```

The solution [is simple](http://ptgmedia.pearsoncmg.com/images/0321113586/items/sutter_item73.pdf): 

Throw by value, catch by reference:

```
  260 |       catch (std::runtime_error &e) {
```

The warning should be available on modern GCC; fixing it consistently all over tree is preferable.

AndersKaloer:
Exceptions should be thrown by value and caught by reference.
Closes #2708

### DIFF
--- a/gnuradio-runtime/lib/block_detail.cc
+++ b/gnuradio-runtime/lib/block_detail.cc
@@ -231,9 +231,8 @@ void block_detail::set_processor_affinity(const std::vector<int>& mask)
     if (threaded) {
         try {
             gr::thread::thread_bind_to_processor(thread, mask);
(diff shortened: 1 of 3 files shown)


marcusmueller:
Awesome, @AndersKaloer, thank you!

PR #2741, opened 2019-08-15
modtool: Improve GRC YAML template

AndersKaloer:
This pull request adds a couple of improvements to the GRC YAML template in gr-modtool:
- Adds a missing "outputs:" key
- Removes redundant documentation
- Removes a couple of trailing whitespaces

### DIFF
--- a/gr-utils/python/modtool/templates/templates.py
+++ b/gr-utils/python/modtool/templates/templates.py
@@ -550,11 +550,11 @@ templates:
   imports: import ${modname}
   make: ${modname}.${blockname}(${strip_arg_types_grc(arglist)})
 
-#  Make one 'parameters' list entry for every Parameter you want settable from the GUI.
-#     Sub-entries of dictionary:
+#  Make one 'parameters' list entry for every parameter you want settable from the GUI.
+#     Keys include:
 #     * id (makes the value accessible as \$keyname, e.g. in the make entry)
-#     * label
-#     * dtype 
+#     * label (label shown in the GUI)
+#     * dtype (e.g. int, float, complex, byte, short, xxx_vector, ...)
 parameters:
 - id: ...
   label: ...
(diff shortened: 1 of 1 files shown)


gr-utils/python/modtool/templates/templates.py
```diff
@@ -563,26 +563,25 @@ def test_001_t(self):
   label: ...
   dtype: ...
 
-#  Make one 'inputs' list entry per input. Sub-entries of dictionary:
+#  Make one 'inputs' and one 'outputs' list entry per input/output. Keys include:
```
mormj: I wouldn't club the input/output together in one description, because blocks can have just input or just output or none.  Perhaps this should read "Make one 'inputs' entry per input and one "outputs" entry per output"

gr-utils/python/modtool/templates/templates.py
```diff
@@ -563,26 +563,25 @@ def test_001_t(self):
...
-#      * optional (set to 1 for optional inputs) 
+outputs:
 - label: ...
   domain: ...
-  dtype: ... #!-- e.g. int, float, complex, byte, short, xxx_vector, ...--
+  dtype: ...
+  vlen: ...
+  optional: ...
 
 file_format: 1
```
mormj: A comment on file_format would be nice here.  It is a required field but there is no documentation anywhere as to why it is needed or what it means.  Perhaps just a "required field" comment would be helpful so people don't figure out the hard way.

skoslowski: The file format specifies the syntax version which makes handling future changes easier. With the XML files this always involved some guess working

gr-utils/python/modtool/templates/templates.py
AndersKaloer: This is a good point. The reason for merging is mainly to prevent stating the same things twice and to reduce the "overhead". I have changed the wording as you suggest.

gr-utils/python/modtool/templates/templates.py
AndersKaloer: I agree. Comment added.

dkozel:
Thanks @AndersKaloer! This will also go onto `maint-3.8` shortly.

PR #2742, opened 2019-08-17
Code improvements through clang-tidy.

marcusmueller:
Requesting a pull to gnuradio:master from marcusmueller:clang-tidy_foundation

Write a message for this pull request. The first block
of text is the title and the rest is the description.

   This is C++11: you can convert

```
   std::vector<complextype> vec; vec.push_back(complextype(foo, bar, baz));
```

   by

```
   std::vector<complextype> vec; vec.emplace_back(foo, bar, baz);
```

   which saves one unnecessary copy (and imho can be more concise for long types).

   This mostly happened in rpc code.

   The automated clang-tidy check failed miserably, so most of this was done
   by hand.

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -1,0 +1,9 @@
+---
+Checks:          '-*,misc-throw-by-value-catch-by-reference,'
+WarningsAsErrors: ''
+HeaderFilterRegex: ''
+AnalyzeTemporaryDtors: false
+FormatStyle:     file
+CheckOptions:
+...
+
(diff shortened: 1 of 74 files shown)


.clang-tidy
```diff
@@ -0,0 +1,10 @@
+---
+Checks:          '-*,misc-throw-by-value-catch-by-reference,misc-static-assert'
```
mbr0wn: Is the idea that we append this list as we get stricter and stricter?

marcusmueller:
Pushing more changes here for review.

.clang-tidy
marcusmueller: exactly; every check I do (and add as a commit here) would be added to the list.

PR #2744, opened 2019-08-17
Fix memset clearing non-trivial type gr_complex warnings in gcc-8

### LINKED ISSUE #2743: GCC>=8 warnings: memset clearing an object of non-trivial type gr_complex
Compiling with GCC >= 8 results in several warnings about memset being used for non-trivial type `gr_complex`. The warning is completely harmless due to the array access requirement to the underlying representation of `gr_complex` (`std::complex<float>`).

For instance, https://github.com/gnuradio/gnuradio/blob/master/gr-filter/lib/fir_filter.cc#L83 and https://github.com/gnuradio/gnuradio/blob/master/gr-blocks/lib/mute_impl.cc#L69 generate warnings.

The easiest way to prevent the warnings is probably to cast the pointer passed to `memset` to `void*`. Are there any better alternatives?

SWIG used to generate similar warnings, but these have been fixed in SWIG 4.0.

AndersKaloer:
Fixes #2743 by casting pointers passed to `memset` to `void*`.

### DIFF
--- a/gr-blocks/lib/mute_impl.cc
+++ b/gr-blocks/lib/mute_impl.cc
@@ -1,6 +1,6 @@
 /* -*- c++ -*- */
 /*
- * Copyright 2004,2010,2013,2018 Free Software Foundation, Inc.
+ * Copyright 2004,2010,2013,2018,2019 Free Software Foundation, Inc.
  *
  * This file is part of GNU Radio
  *
@@ -28,6 +28,7 @@
 #include "mute_impl.h"
 #include <gnuradio/io_signature.h>
 #include <string.h>
+#include <algorithm>
 
 namespace gr {
 namespace blocks {
@@ -66,7 +67,7 @@ int mute_impl<T>::work(int noutput_items,
     int size = noutput_items;
 
     if (d_mute) {
-        memset(optr, 0, noutput_items * sizeof(T));
+        std::fill_n(optr, noutput_items, 0);
     } else {
         while (size >= 8) {
             *optr++ = *iptr++;
--- a/gr-digital/lib/burst_shaper_impl.cc
+++ b/gr-digital/lib/burst_shaper_impl.cc
@@ -1,6 +1,6 @@
 /* -*- c++ -*- */
 /*
- * Copyright 2015,2018 Free Software Foundation, Inc.
+ * Copyright 2015,2018,2019 Free Software Foundation, Inc.
  *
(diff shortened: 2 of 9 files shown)


drmpeg:
An alternative method is to use std::fill_n. This is what I used in the DTV component.

https://github.com/gnuradio/gnuradio/commit/2826fd096e3e157d9f488d058ff85c5b68e06b3e

AndersKaloer:
Thanks for the comments. It seems that `fill_n` is the preferred solution, so I will update the code to use that as soon as possible. I agree that `fill_n` is a more clean solution, especially for generic types – the only reason that I casted initially was to minimize the changes.

michaelld:
I've been using the type casting for a very long time ... it's my default go-to instead of more modern techniques like `fill_n` ... I'm slowly updating my brain to use the latter ... just makes good sense, but old habits are hard to break!

AndersKaloer:
I have updated the code to use `fill_n´. Please verify that I haven't overlooked something during my own review.. Thanks.

michaelld:
Excellent! A quick read-through LGTM. I'll verify via compiling & "make test" too ... doesn't necessarily guarantee correctness, but it's about the best we can do with limited time!

michaelld:
-After- this PR, there are some 190 instances of `memset` or `bzero`, and of the former probably 170 are setting to 0. This in over some 80 files. All + or - a few ... not exact science here, but rather getting rough numbers for the sake of this query.

This PR handles those that GCC8 caught. There are some for which the code currently uses `(void*)` for the type, and plenty for which the type is a standard compiler type (e.g., `int`).

Do we want wholesale conversion to `fill_n` for consistency? Doing so won't be particularly challenging, just will take some time and persistence.

I'm not saying here that @AndersKaloer should do this ... just thinking out loud that when we accept a PR like this I'd like to see the GR codebase made consistent for all such function usage.

Wondering what others think ... ?

michaelld:
Passes `make test` just as well as before, so I think this PR is good to go. My meta-query remains from the prior post.

mbr0wn:
I guess if someone wants to do that... we have other fish to fry :) At least there's no more compiler warnings now.

michaelld:
> I guess if someone wants to do that... we have other fish to fry :) At least there's no more compiler warnings now.

Truth! Every step helps here. Fixing 190 of those ain't going to be quick or soon if I'm doing it. I've got too many other fish to fry!

PR #2746, opened 2019-08-19
gr-filter: Fix a typo in the YAML for Polyphase Channelizer

argilo:
When a Polyphase Channelizer is used in GRC, the generated Python code fails with the following error:
```
  File "./channelizer.py", line 83
    self.pfb_channelizer_ccf_0.set_channel_map([])\nself.pfb_channelizer_ccf_0.declare_sample_delay(0)
                                                                                                     ^
SyntaxError: unexpected character after line continuation character
```
This is due to a typo in the YAML file. I've replaced `\n` with an actual carriage return to fix it.

### DIFF
--- a/gr-filter/grc/filter_pfb_channelizer.block.yml
+++ b/gr-filter/grc/filter_pfb_channelizer.block.yml
@@ -53,7 +53,8 @@ templates:
             ${taps},
             ${osr},
             ${atten})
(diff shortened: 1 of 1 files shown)


argilo:
I noticed this problem in the 3.8.0.0 release, but it's still present on the master branch.

mbr0wn:
cp'd to maint-3.8.

PR #2747, opened 2019-08-19
Fix typo that creates build issues from GR dependencies

eblot:
I'm not sure whether it is intentional or not, but it seems a typo in `cmake/Modules/FindLOG4CPP.cmake` creates build error for applications that depends on GR CMake module to build, `e.g.` with [gr-osmosdr](https://osmocom.org/projects/osmosdr)

    CMake Error at /usr/local/lib/cmake/gnuradio/FindLOG4CPP.cmake:52 (add_library):
      add_library cannot create imported target "Log4Cpp::log4cpp" because
      another target with the same name already exists.
      /usr/local/Cellar/cmake/3.15.2/share/cmake/Modules/CMakeFindDependencyMacro.cmake:47 (find_package)
      /usr/local/lib/cmake/gnuradio/GnuradioConfig.cmake:25 (find_dependency)
      CMakeLists.txt:184 (find_package)

### DIFF
--- a/cmake/Modules/FindLOG4CPP.cmake
+++ b/cmake/Modules/FindLOG4CPP.cmake
@@ -48,7 +48,7 @@ else ()
 endif ()
 
 
-if (LOG4CPP_FOUND AND NOT TARGET log4cpp::log4cpp)
+if (LOG4CPP_FOUND AND NOT TARGET Log4cpp::log4cpp)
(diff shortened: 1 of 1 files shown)


eblot:
@michaelld I'm not sure whether it is another macOS specific issue. There is no reason why it would, but I did not check how it goes on Linux.

marcusmueller:
Thank you, @eblot and @michaelld!

If testing is successful: please merge at will :)

PR #2750, opened 2019-08-20
Block header parsing tool: GSoC 2019

aru31:
gr_blocktool: Also known as the block header parsing tool, this tool automatically parses any GNU Radio or OOT block header.
- Parse any GNU Radio or OOT header file with just the `file-path` as an input.
- Parse a complete header directory with `directory-path` as the input.
- Get the output in the form of a YAML or a JSON file.
- Add blocktool comments automatically in the header file from the implementation file. (Blocktool comments are the special comment syntax in the block header files which is read by the tool to get all the necessary information for the I/O signatures, message-ports which otherwise were not available because of the missing header implementation file).
- Blocktool API can also be called from modtool.
- Modtool `makeyaml` subcommand along with -b flag and a file-path as the input can be used to create YAML files for the GRC.
- YAML output is much better using the blocktool API.
- Blocktool can be also be used as an independent API which thus can be used to
parse a block header file during runtime.


aru31:
@skoslowski @primercuervo

primercuervo:
@marcusmueller we had our review rounds during development in the fork, but this seems all good to be review'd here. Please let us know what kind of support is needed to get this merged here

PR #2756, opened 2019-08-23
Patch FindQwt.cmake to work on ArchLinux

SpectreJan:
Supercedes issue 2545 partly. It adds the (plain) qwt to the find_library call but preserves the original search order. Tested on Manajaro 18.10rc3 and it is enough to build GNU Radio without adding additional CMAKE (cmdline) flags during the configure part. 

 have not looked at the other changes made in 2545 to FindQwt.cmake but for now it seems to solve build issues on Arch and adhering to the change requests made to the original commt in issue 2545.


marcusmueller:
cherry-picked manually.

PR #2768, opened 2019-08-28
runtime: fixup errors from clang-tidy changes

bastibl:
introduced in https://github.com/gnuradio/gnuradio/commit/c6cb3bbc35153a9635d69eddab0d252f7441b854

Since `d_rpc_vars` is of type `std::vector<boost::any>` the sptr was not constructed as before.


marcusmueller:
'doh. Yes, thanks.

Issue #2764, opened 2019-08-26
QPSK constellation object not scaled correctly


jdemel:
If you use a `Constellation Object` with QPSK modulation, the symbols are not scaled properly, impling `|s|^2 != 1`. 

In `gr-digital/lib/constellation.cc` the QPSK symbols are defined as 
`gr_complex(-GR_M_SQRT2, -GR_M_SQRT2)`. 
They should be defined as 
`gr_complex(-1.0 / GR_M_SQRT2, -1.0 / GR_M_SQRT2)`.

All other constellations seem to stick to the `|s|^2 == 1` convention except for QPSK.

marcusmueller:
Still the case:

```c++
constellation_qpsk::constellation_qpsk()
{
    d_constellation.resize(4);
    // Gray-coded
    d_constellation[0] = gr_complex(-GR_M_SQRT2, -GR_M_SQRT2);
    d_constellation[1] = gr_complex(GR_M_SQRT2, -GR_M_SQRT2);
    d_constellation[2] = gr_complex(-GR_M_SQRT2, GR_M_SQRT2);
    d_constellation[3] = gr_complex(GR_M_SQRT2, GR_M_SQRT2);

    /*
      d_constellation[0] = gr_complex(GR_M_SQRT2, GR_M_SQRT2);
      d_constellation[1] = gr_complex(-GR_M_SQRT2, GR_M_SQRT2);
      d_constellation[2] = gr_complex(GR_M_SQRT2, -GR_M_SQRT2);
      d_constellation[3] = gr_complex(GR_M_SQRT2, -GR_M_SQRT2);
    */

    d_pre_diff_code.resize(4);
    d_pre_diff_code[0] = 0x0;
    d_pre_diff_code[1] = 0x2;
    d_pre_diff_code[2] = 0x3;
    d_pre_diff_code[3] = 0x1;

    d_rotational_symmetry = 4;
    d_dimensionality = 1;
    // normalization_t norm = POWER_NORMALIZATION;
    // normalize(norm);
    calc_arity();
}
```

However, fixing this will require an API break, as we reduce things by 6 dB…

So, doesn't look like we could backport a fix to this for 3.10, and I don't know whether it's worth fixing on main, then. @mormj , @willcode ?

willcode:
Also dqpsk.

Commented out code in qpsk and dqpsk (added in #6748 but never uncommented) would POWER_NORMALIZATION. In the PR, the author said this broke other stuff (equalizers, QA).

PR #2771, opened 2019-08-28
Normalize the constellation points so that the mean square magnitude is 1 (instead of the non-squared magnitude).

eduardosm:
This allows to easily create signals with normalized power and makes it consistent with the 16-QAM constellation (which is also normalized for mean square magnitude = 1).

### DIFF
--- a/gr-digital/grc/digital_constellation.block.yml
+++ b/gr-digital/grc/digital_constellation.block.yml
@@ -29,6 +29,13 @@ parameters:
     dtype: int
     default: '1'
     hide: ${ ( 'none' if str(type) == "calcdist" else 'all') }
+-   id: normalization
+    label: Normalization Type
+    dtype: int
+    default: digital.constellation.AMPLITUDE_NORMALIZATION
+    options: [digital.constellation.NO_NORMALIZATION, digital.constellation.POWER_NORMALIZATION, digital.constellation.AMPLITUDE_NORMALIZATION]
+    option_labels: [None, Power, Amplitude]
+    hide: ${ ( 'none' if str(type) == "calcdist" else 'all') }
 -   id: precision
     label: Soft Decisions Precision
     dtype: int
@@ -39,7 +46,7 @@ parameters:
     dtype: raw
     default: None
     hide: ${ ('part' if str(soft_dec_lut) == 'None' else 'none') }
-value: ${ digital.constellation_calcdist(const_points, sym_map, rot_sym, dims) if (str(type) == "calcdist") else getattr(digital,'constellation_'+str(type))()  }
+value: ${ digital.constellation_calcdist(const_points, sym_map, rot_sym, dims, normalization) if (str(type) == "calcdist") else getattr(digital,'constellation_'+str(type))()  }
 
 
 templates:
--- a/gr-digital/include/gnuradio/digital/constellation.h
+++ b/gr-digital/include/gnuradio/digital/constellation.h
@@ -49,11 +49,17 @@ typedef std::shared_ptr<constellation> constellation_sptr;
 class DIGITAL_API constellation : public std::enable_shared_from_this<constellation>
 {
 public:
+    enum normalization_t {
+        NO_NORMALIZATION,
+        POWER_NORMALIZATION,
+        AMPLITUDE_NORMALIZATION,
+    };
+
     constellation(std::vector<gr_complex> constell,
                   std::vector<int> pre_diff_code,
                   unsigned int rotational_symmetry,
                   unsigned int dimensionality,
-                  bool normalize_points = true);
+                  normalization_t normalization = AMPLITUDE_NORMALIZATION);
     constellation();
     virtual ~constellation();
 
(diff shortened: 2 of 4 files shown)


marcusmueller:
Hi! Thanks for contributing this PR (I like it, especially since you've even thought of modifying the documenation).

This of course breaks the unit tests, so unless you also adapt these, we can't merge this. Also, it changes the way an existing block that people rely on working the way it does – that's a no-go, I'm afraid.

I can see your use case, though! Power normalization actually *is* the more common way of normalization than amplitude normalization. So, what about the following:

1. change the `normalize_points` parameter to not be a `bool`, but an `enum` that allows choosing between `NO_NORMALIZATION`, `POWER_NORMALIZATION`, `AMPLITUDE_NORMALIZATION`
2. use that parameter to pick the appropriate normalization method
3. (important) add a unit test that tests all three modes, including the new power normalization mode.

marcusmueller:
Oh, and while you can possibly fly under the radar with this "small change" (it could boil down to < 10 LOC) in terms of copyright: I think it would be wisest if you contacted bhilburn AT gnuradio PERIOD org (@bhilburn) and got a Contributor license agreement, which allows the FSF to defend the GPL code, and allows you to do whatever you want with your own code.

eduardosm:
>     1. change the `normalize_points` parameter to not be a `bool`, but an `enum` that allows choosing between `NO_NORMALIZATION`, `POWER_NORMALIZATION`, `AMPLITUDE_NORMALIZATION`
> 
>     2. use that parameter to pick the appropriate normalization method

I have pushed a new commit implementing it as suggested.

>     3. (important) add a unit test that tests all three modes, including the new power normalization mode.

Tests are now passing, because I set the default to `AMPLITUDE_NORMALIZATION`, but I am not sure how to extend the current tests to cover all the cases.

I have also modified the GRC block definition (`digital_constellation.block.yml`) to allow choosing the normalization type for 'Variable Constellation', but I haven't had a chance to test it yet.



> Oh, and while you can possibly fly under the radar with this "small change" (it could boil down to < 10 LOC) in terms of copyright: I think it would be wisest if you contacted bhilburn AT gnuradio PERIOD org (@bhilburn) and got a Contributor license agreement, which allows the FSF to defend the GPL code, and allows you to do whatever you want with your own code.

Thanks for the tip, I have sent an email to get more information.

jdemel:
This should fix #2764 . I have a QA test in `gr-gfdm` that relies on normalized amplitudes in the constellation object. It did work with 3.7 and broke in 3.8. I'd argue that 3.8 actually introduced a regression here. Thus, making normalized amplitudes the default would actually NOT break other things.

michaelld:
@eduardosm @bhilburn did we get your CLA finalized?

mormj:
@eduardosm  - we'd love to be able to include this PR - any update on the CLA process?

eduardosm:
I have been asked to get a disclaimer for my university. Due to the current situation here in Spain, I don't know when that will be possible.

Anyways, is the CLA that important for this PR? It is not like I have invented the wheel here...

michaelld:
> Anyways, is the CLA that important for this PR? It is not like I have invented the wheel here...

Yes the CLA is required, since the changes are more than ~10 lines of code. Even much more simple changes can require a CLA if there are enough changes ...

eduardosm:
> Yes the CLA is required, since the changes are more than ~10 lines of code. Even much more simple changes can require a CLA if there are enough changes ...

This will have to wait then. This CLA thing feels to me like a barrier to entry for new contributors...

mohmmadd:
> > Yes the CLA is required, since the changes are more than ~10 lines of code. Even much more simple changes can require a CLA if there are enough changes ...
> 
> This will have to wait then. This CLA thing feels to me like a barrier to entry for new contributors...

@eduardosm Are you have any good news?

mormj:
@eduardosm - We have removed some of the barrier for new contributors by using the DCO instead of CLA (https://github.com/gnuradio/gnuradio/blob/master/CONTRIBUTING.md)

All this means, is that we can merge your changes if you signoff your commit git commit --amend -s and force push

(looks like it needs a rebase as well)

eduardosm:
> @eduardosm - We have removed some of the barrier for new contributors by using the DCO instead of CLA (https://github.com/gnuradio/gnuradio/blob/master/CONTRIBUTING.md)
> 
> All this means, is that we can merge your changes if you signoff your commit git commit --amend -s and force push
> 
> (looks like it needs a rebase as well)

Thanks for the notification. I have rebased and added the signoff as suggested.


Build is failing with
```
  Python bindings for constellation.h are out of sync
  gr-digital/python/digital/bindings/CMakeLists.txt:97 (GR_PYBIND_MAKE_CHECK_HASH)
```

But I am not sure where those bindings are located. Aren't those generated automatically?

mormj:
> But I am not sure where those bindings are located. Aren't those generated automatically?

Unfortunately no - I can help with this if you would like because it would be nice to have this feature ready for the 3.9 release cutoff.  The python bindings are located in the python/bindings/*_python.cc files and follow standard pybind11 syntax.  There is a helper script bind_intree_file.py that might be able to help.

eduardosm:
I think I've got the bindings

marcusmueller:
@eduardosm Awesome!

PR #2776, opened 2019-08-30
Add an interface to set the receive buffer size,Optimizing UDP Receiving Capability

uniradar:
Add an interface to set the UDP receive buffer size,Optimizing UDP Receiving Capability

### DIFF
--- a/gr-blocks/grc/blocks_udp_source.block.yml
+++ b/gr-blocks/grc/blocks_udp_source.block.yml
@@ -31,6 +31,10 @@ parameters:
     label: Vec Length
     dtype: int
     default: '1'
+-   id: udpbufsize
+    label: udp receive buffer size
+    dtype: int
+    default: '1048576'
     hide: ${ 'part' if vlen == 1 else 'none' }
 
 outputs:
@@ -43,12 +47,12 @@ asserts:
 
 templates:
     imports: from gnuradio import blocks
-    make: blocks.udp_source(${type.size}*${vlen}, ${ipaddr}, ${port}, ${psize}, ${eof})
+    make: blocks.udp_source(${type.size}*${vlen}, ${ipaddr}, ${port}, ${psize}, ${eof},${udpbufsize})
 
 cpp_templates:
     includes: ['#include <gnuradio/blocks/udp_source.h>']
     declarations: 'blocks::udp_source::sptr ${id};'
-    make: 'this->${id} = blocks::udp_source::make(${type.size}*${vlen}, ${ipaddr}, ${port}, ${psize}, ${eof});'
+    make: 'this->${id} = blocks::udp_source::make(${type.size}*${vlen}, ${ipaddr}, ${port}, ${psize}, ${eof},${udpbufsize});'
     translations:
         gr.sizeof_: 'sizeof('
         'True': 'true'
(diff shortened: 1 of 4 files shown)


uniradar:
Tests on Ubuntu platform show that the UDP receiving rate can be increased from 90MB/s to 1GB/s by increasing the receiving buffer.

mormj:
Looks like quite a performance improvement, and more configurability of blocks is generally a good thing.  As far as the default setting - is this what the socket would default to if the recv_option was not set?  

Might be safer to default to 0, and then not set the option if the size is set to 0.  This way previously generated flowgraphs will work exactly the same as before.

gr-blocks/grc/blocks_udp_source.block.yml
```diff
@@ -43,12 +47,12 @@ asserts:
 
 templates:
     imports: from gnuradio import blocks
-    make: blocks.udp_source(${type.size}*${vlen}, ${ipaddr}, ${port}, ${psize}, ${eof})
+    make: blocks.udp_source(${type.size}*${vlen}, ${ipaddr}, ${port}, ${psize}, ${eof},${udpbufsize})
```
michaelld: Would you please add a space between the "," and "$", to make the new argument the same style as previous arguments? Ditto here and line 55.

mbr0wn:
@uniradar Do we have a CLA from you? Please remove the label if we do. If not, please reach out to bhilburn at gnuradio dot org for a CLA.

gr-blocks/lib/udp_source_impl.cc
```diff
@@ -40,22 +40,23 @@ const int udp_source_impl::BUF_SIZE_PAYLOADS =
     gr::prefs::singleton()->get_long("udp_blocks", "buf_size_payloads", 50);
 
 udp_source::sptr udp_source::make(
-    size_t itemsize, const std::string& ipaddr, int port, int payload_size, bool eof)
+    size_t itemsize, const std::string& ipaddr, int port, int payload_size, bool eof,int udp_recv_buf_size)
```
mbr0wn: Please use clang-format.

ghostop14:
I incorporated this into gr-grnet, so if we do roll in those networking blocks this will get included there.

ghostop14:
I'm working on the grnet code to PR this week, but I ran into an issue with the receivebuffersize setting talked about here.  It wasn't evident until I enabled grnet's UDP sequence numbers to track packet drops.  When I tried it with sequence numbers to track missed frames, the setting was actually causing lost data so I had to back it out.  I do have a quick fix though that'll be in the documentation for the UDP blocks.  Edit /etc/sysctl.conf and add the following lines then reboot (you do have to reboot rather than sysctl -w for these to get picked up):

net.core.rmem_default=26214400
net.core.rmem_max=104857600
net.core.wmem_default=65536
net.core.wmem_max=104857600

Missed frames should go away and throughput should go back up.

mormj:
@uniradar - any update on the CLA process?

mormj:
Recommending we close this one, since this functionality has been included into gr-network.  @ghostop14 - do you concur?

ghostop14:
I would agree.  Changing the settings for the kernel has gotten me the highest throughput and least dropped packets on 10G / 40G networks.

mormj:
Closing.  @uniradar - please open a new PR if there are changes you would like to introduce to gr-network.

PR #2779, opened 2019-08-30

mormj:
-- Looking for feedback on this 

Here is a contrived example that just queries the ref_locked sensor:
image
image

This would then generate the .py output with the code inserted after the specified section:
image

### DIFF
--- a/grc/blocks/grc.tree.yml
+++ b/grc/blocks/grc.tree.yml
@@ -12,6 +12,7 @@
   - epy_module
   - note
   - import
+  - snippet 
 - Variables:
   - variable
   - variable_struct
--- a/grc/blocks/options.block.yml
+++ b/grc/blocks/options.block.yml
@@ -123,6 +123,53 @@ parameters:
     default: '.:'
     hide: part
 
+-   id: snippet_editor
+    label: Editor
+    dtype: enum
+    category: Code Snippets
+    options: ['Dialog','External']
+    hide: ${'part' if output_language == 'python' else 'all'}
+
+-   id: snippet_top
+    label: Top
+    dtype: ${'_multiline_python_external' if snippet_editor == 'External' else '_multiline'}
+    category: Code Snippets
+    hide: ${'part' if output_language == 'python' else 'all'}
+-   id: snippet_imports
+    label: Imports
+    dtype: ${'_multiline_python_external' if snippet_editor == 'External' else '_multiline'}
+    category: Code Snippets
+    hide: ${'part' if output_language == 'python' else 'all'}
+-   id: snippet_variables
+    label: Variables
+    dtype: ${'_multiline_python_external' if snippet_editor == 'External' else '_multiline'}
+    category: Code Snippets
+    hide: ${'part' if output_language == 'python' else 'all'}
+-   id: snippet_blocks
+    label: Blocks
+    dtype: ${'_multiline_python_external' if snippet_editor == 'External' else '_multiline'}
+    category: Code Snippets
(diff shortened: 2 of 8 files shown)


marcusmueller:
I like the idea – but since we really can have but one code snippet per "section", wouldn't it be wiser to just have this as property of the "Options" block?

mormj:
I'm having a hard time visualizing how it would fit into the options block.  Are you imagining a second popup from there that defines your snippets for each section?

Also, in theory we could have more than one snippet per section, but currently like the other variables there is no concept of ordering them upon rendering

skoslowski:
This reminds me of Balints "any code" blocks.

I agree with @marcusmueller - if this would be added, it should suffice to have one per section and therefore put it into the options - a new tab containing multiline inputs for all these "extension points". 
 
In general, I always disliked the idea of introducing elements that enable users to generate invalid code. This would be one of those features: Can't exec the code (could have side effects) - only compile it to check for syntax errors - maybe analyse the ast for unknown identifiers. However, there are (advanced-level) use-cases where features like these are nice - even though it is no longer "pure" fllowgraph generation.

AFAIK the current way to achieve this, is to create a separate module and import the generated flowgraph class. Here you are free to extend/patch/customize the code however you see fit. As far as GRC is concerned it generates a valid flowgraph - any custom logic is in a separate file.
And you can customize what file to run per flowgraph - so you can have the extended code executed from the GUI. This solution may be suited for more involved customization. Your example adds a really short piece of code for which creating a separate module may feel a bit cumbersome.

@mormj, variables are in fact ordered upon rendering (how else could you reliably use another variable to build a new value?). blocks and connections shouldn't require a special order (think declarative) - with the qtgui being the usual exception.

marcusmueller:
exactly, that's why I thought this should be a singleton thing; I imagine the "advanced" tab in the Options to hold a couple of text fields.

mormj:
Cool - thanks for the feedback - really appreciate it!  I'll give it another pass in the options section.  

I know that the idea is really giving the users the rope to hang themselves (in terms of unchecked code being inserted), but sometimes there's things that just aren't there in GRC and you need a band-aid until it's properly supported.

mormj:
Here it is thrown onto the options block:

image

with a choice to use external editor if so desired:
image

My concerns with this approach (as opposed to the snippet block) are
1) If the number of insertion points grows, this gets over-populated
2) Cut and paste snippets between flowgraphs - with the snippet block you can copy your snippets to another flowgraph easily.

777arc:
Now that it's moved into the options block, I think the big thing missing is a clear indication that custom code is being injected into the Python.  With your first version, at least the presence of that extra block was a clear indicator.  But now it's pretty hidden away, if someone opens a flowgraph they aren't going to realize it's there unless they really go digging.   Perhaps we could add an extra line of text inside the options block (like in the canvas view) that has some bold red text indicating a snippet of python is being added.  What do you think?

mormj:
Good point @777arc 

Personally I prefer the separate snippets block for the purpose of visibility and also because of copy paste to other flowgraphs, but understand the desire to keep option-like things in the options block.

What if it were it's own block - all the snippets like shown in the options block, but there could only be one of them (like the options block).

mormj:
Closing in favor of #2814

liumaolin-931:
Hi，
I'm just beginning to learn gnuradio. Where can I find more complete C + + and python codes for those existing modules? If I want to modify the function and content of one of the modules, what should I do? such as“constellation Modulator”,Is there any good website to help me?thank you!

PR #2781, opened 2019-09-02
Use `dist-packages` on Linux only

eblot:
`dist-packages` directory is a Debian/Ubuntu/... specific path
On other posix systems the default `site-packages` dir should be used.

https://stackoverflow.com/questions/42339034/python-module-in-dist-packages-vs-site-packages

Ex. on macOS, there is no "dist-packages" directory, whatever the installed packages

    $ ls /usr/local/opt/python/Frameworks/Python.framework/Versions/3.7/lib/python3.7/site-* |wc  -l
         208
    $ ls /usr/local/opt/python/Frameworks/Python.framework/Versions/3.7/lib/python3.7/dist-* | wc -l
       No such file or directory
       0

Moreover, even GR hesitates about the installation path:

    $ ls /usr/local/opt/gnuradio/lib/python3.7/dist-packages/
    gnuradio pmt
    $ ls /usr/local/opt/gnuradio/lib/python3.7/site-packages/
    volk_modtool

This ends up with confusion, multiple `PYTHONPATH` to make GR/GRC run, etc.

I'm not sure this PR is the proper way to fix this issue, but `dist-packages` does not seem to be a valid path except for Debian-style installations.

### DIFF
--- a/cmake/Modules/GrPython.cmake
+++ b/cmake/Modules/GrPython.cmake
@@ -112,7 +112,10 @@ execute_process(COMMAND ${PYTHON_EXECUTABLE} -c "
 import os
 import sys
 if os.name == 'posix':
-    print(os.path.join('lib', 'python' + sys.version[:3], 'dist-packages'))
+    if sys.platform == 'linux':
+        print(os.path.join('lib', 'python' + sys.version[:3], 'dist-packages'))
+    else:
+        print(os.path.join('lib', 'python' + sys.version[:3], 'site-packages'))
 if os.name == 'nt':
     print(os.path.join('Lib', 'site-packages'))
 " OUTPUT_VARIABLE GR_PYTHON_DIR OUTPUT_STRIP_TRAILING_WHITESPACE


skoslowski:
How about `distutils.sysinfo.get_python_lib()`?

eblot:
@skoslowski There is no `sysconfig` in distutils on macOS (and iOS) at least, but there's indeed a standard `sysconfig` module.

However, this gives access to the Python interpreter paths, not application-specific paths:

#### macOS/Homebrew
    >>> from sysconfig import get_paths
    >>> from pprint import pprint
    >>> pprint(get_paths())
    {'data': '/usr/local/Cellar/python/3.7.4/Frameworks/Python.framework/Versions/3.7',
     'include': '/usr/local/Cellar/python/3.7.4/Frameworks/Python.framework/Versions/3.7/include/python3.7m',
     'platinclude': '/usr/local/Cellar/python/3.7.4/Frameworks/Python.framework/Versions/3.7/include/python3.7m',
     'platlib': '/usr/local/Cellar/python/3.7.4/Frameworks/Python.framework/Versions/3.7/lib/python3.7/site-packages',
     'platstdlib': '/usr/local/Cellar/python/3.7.4/Frameworks/Python.framework/Versions/3.7/lib/python3.7',
     'purelib': '/usr/local/Cellar/python/3.7.4/Frameworks/Python.framework/Versions/3.7/lib/python3.7/site-packages',
     'scripts': '/usr/local/Cellar/python/3.7.4/Frameworks/Python.framework/Versions/3.7/bin',
     'stdlib': '/usr/local/Cellar/python/3.7.4/Frameworks/Python.framework/Versions/3.7/lib/python3.7'}

#### linux/Mint

    >>> from sysconfig import get_paths
    >>> from pprint import pprint
    >>> pprint(get_paths())
    {'data': '/usr',
     'include': '/usr/include/python3.6m',
     'platinclude': '/usr/include/python3.6m',
     'platlib': '/usr/lib/python3.6/site-packages',
     'platstdlib': '/usr/lib/python3.6',
     'purelib': '/usr/lib/python3.6/site-packages',
     'scripts': '/usr/bin',
     'stdlib': '/usr/lib/python3.6'}

You may not want to install GR Python modules into the main Python interpreter directory, but into a sub GR directory, *e.g.* on macOS/Homebrew:
  * Python path: `/usr/local/Cellar/python/3.7.4/Frameworks/Python.framework/Versions/3.7/lib/python3.7/site-packages`
  * GR path: ` /usr/local/Cellar/gnuradio/3.8.0.0/lib/python3.7/site-packages`

The fun part is that `sysconfig` module on Linux/Mint (wild guess: same on Debian) does not seem to know about `dist-packages` directory, which nevertheless exists on the FS.

michaelld:
I'd have to do some searching, but I know this topic has been discussed at least once, probably more than once, in the past. We never figured out a "one size fits all" set of defaults.

GR's CMake provides an externally settable variable for this install location (`GR_PYTHON_DIR`), which though obviously not an internal OS-dependent default, provides a reasonable means for this setting.

I tend to believe that most package managers just use the external variable setting, since oftentimes they wish to install Python scripts in some sub-directory of `CMAKE_INSTALL_PREFIX`. This is certainly the case for MacPorts, and it looks from the above like not Brew either. Is OSX that non-standard LOL?

Given the diversity of Python instlall locations, I just set my shell's `PYTHONPATH` to include multiple variants (`site-packages` and `dist-packages`) for multiple top-level directories (`CMAKE_INSTALL_PREFIX/lib/pythonX.Y` along with `CMAKE_INSTALL_PREFIX/Library/Frameworks/Python.framework/Versions/X.Y/lib/pythonX.Y`).

All of the above said, I don't see this change as being harmful. I doubt it will change much from the end-user's perspective.

I'm curious what other Linux user think here; I think the OSX users have had plenty to say!

eblot:
> Given the diversity of Python instlall locations, I just set my shell's `PYTHONPATH` to include multiple variants (`site-packages` and `dist-packages`) for multiple top-level directories 
Sure, but:

 1- From my perspective, even if Debian is my second OS of choice, it's weird to enforce a Debian-only rule for all *nix environments, I guess than *BSD users are not aware of `dist-packages`?
 2- There is only an exception to this for Windows :-)
 3- Why is `volk` installed in one path, `gnuradio` in the other one?

noc0lour:
if at all, python module installation should default/fallback to `site-packages` for Unixoids, Introduction of `dist-packages` was a mistake imho.

marcusmueller:
the old form is worse, but the new one is still wrong, since really only debianoids use dist-packages.

eblot:
I'd suggest two options:
- always use `site-packages` on Unix, whatever GR runs on linux or not, and leave to the distribution package managers the decision to patch the code if they really want to use a custom-specific destination directory. This is what is done for other environments. 
  * At least, it seems it is better to follow Python rules than specific distribution rules, rather to force the most and regular platforms to use support non-standard paths as it actually is done for now.
- detect debianoid environments, using /etc/os-release and comparing ID and ID_LIKE key values  to -at  least- `debian` and `ubuntu`

ThomasHabets:
As someone who's tried to build gnuradio on non-linux OSs, and with lots of experience porting between OSs in general, yes please try to find a way to do this without an `if (linux) … else …`. It's technical debt making future porting harder.

Example problem from "if OS X, Y or Z": https://github.com/gnuradio/gnuradio/issues/2902

derekjchen:
Hi all,

I wrote a dvbs2 bbheader file sink. (for learning purpose). I did cmake, make, make install, ldconfig on it.

and I'm writing python tests for dvbs2 bbheader using the bbheader file sink block.
however, in the python test file, qa_bbheader_bb.py, it can’t resolve this:
import bbheader_file_sink_swig as bbheader_file_sink

2: Test command: /usr/bin/sh "/data/Frontier/gr-dvbs2/build/python/qa_bbheader_bb_test.sh"
2: Test timeout computed to be: 10000000
2: Traceback (most recent call last):
2:   File "/data/Frontier/gr-dvbs2/python/qa_bbheader_bb.py", line 27, in <module>
2:     import bbheader_file_sink_swig as bbheader_file_sink
2: ImportError: No module named bbheader_file_sink_swig
2/8 Test #2: qa_bbheader_bb ...................***Failed    0.25 sec
test 3
    Start 3: qa_bbscrambler_bb


and I can see bbheader_file_sink is under:
/user/local/include/bbheader_file_sink/bbheader_file_sink/swig/bbheader_file_sink_swig.i

should I edit dvbs2 OOT CMakeFile? and how?

Really appreciate any help.

Regards,
Derek

michaelld:
closed in favor of #2939

PR #2783, opened 2019-09-03
Fixed ambiguity of C++ nested template syntax vs. right-shift operator.

seeAppRun:
Added space between nested template brackets to fix C++ syntax error.

### DIFF
--- a/gnuradio-runtime/include/pmt/pmt.h
+++ b/gnuradio-runtime/include/pmt/pmt.h
@@ -542,9 +542,9 @@ PMT_API pmt_t init_f32vector(size_t k, const std::vector<float>& data);
 PMT_API pmt_t init_f64vector(size_t k, const double* data);
 PMT_API pmt_t init_f64vector(size_t k, const std::vector<double>& data);
 PMT_API pmt_t init_c32vector(size_t k, const std::complex<float>* data);
-PMT_API pmt_t init_c32vector(size_t k, const std::vector<std::complex<float>>& data);
+PMT_API pmt_t init_c32vector(size_t k, const std::vector<std::complex<float> >& data);
 PMT_API pmt_t init_c64vector(size_t k, const std::complex<double>* data);
-PMT_API pmt_t init_c64vector(size_t k, const std::vector<std::complex<double>>& data);
+PMT_API pmt_t init_c64vector(size_t k, const std::vector<std::complex<double> >& data);
 
 PMT_API uint8_t u8vector_ref(pmt_t v, size_t k);
 PMT_API int8_t s8vector_ref(pmt_t v, size_t k);
@@ -603,8 +603,8 @@ PMT_API const std::vector<uint64_t> u64vector_elements(pmt_t v);
 PMT_API const std::vector<int64_t> s64vector_elements(pmt_t v);
 PMT_API const std::vector<float> f32vector_elements(pmt_t v);
 PMT_API const std::vector<double> f64vector_elements(pmt_t v);
-PMT_API const std::vector<std::complex<float>> c32vector_elements(pmt_t v);
-PMT_API const std::vector<std::complex<double>> c64vector_elements(pmt_t v);
+PMT_API const std::vector<std::complex<float> > c32vector_elements(pmt_t v);
+PMT_API const std::vector<std::complex<double> > c64vector_elements(pmt_t v);
 
 // len is in elements
 PMT_API const std::vector<uint8_t> pmt_u8vector_elements(pmt_t v);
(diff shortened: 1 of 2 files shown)


michaelld:
@seeAppRun Did you do a thorough check through the GR codebase to make sure these are all of the ">>"s & related? If there are more, let's get them fixed up here, please & thank you!

seeAppRun:
> @seeAppRun Did you do a thorough check through the GR codebase to make sure these are all of the ">>"s & related? If there are more, let's get them fixed up here, please & thank you!

These were found when building from source. I can also grep through the code to see if there are other files with this issue.

michaelld:
> These were found when building from source. I can also grep through the code to see if there are other files with this issue.

Makes sense. Yes, please do the grep & see what you can get there too.

Unrelated: Do you know if you have an assigning document ("CLA") on file for GNU Radio with the Free Software Foundation (FSF)? These changes might be "small enough" to not require one, but if you're going to be a regular code contributor to GR, then you'll want one in place.

seeAppRun:
> > These were found when building from source. I can also grep through the code to see if there are other files with this issue.
> 
> Makes sense. Yes, please do the grep & see what you can get there too.
> 
> Unrelated: Do you know if you have an assigning document ("CLA") on file for GNU Radio with the Free Software Foundation (FSF)? These changes might be "small enough" to not require one, but if you're going to be a regular code contributor to GR, then you'll want one in place.

I do not (new contributor). Could you please point me to CLA process?

michaelld:
@bhilburn can you get the CLA going with @seeAppRun ? Thx!

bhilburn:
@michaelld - of course!

@seeAppRun - shoot me an e-mail, bhilburn@gnuradio.org

seeAppRun:
> > @seeAppRun Did you do a thorough check through the GR codebase to make sure these are all of the ">>"s & related? If there are more, let's get them fixed up here, please & thank you!

After looking at the code base, there are a lot of files that use the '>>' syntax for nested templates. Apparently this is newly allowed in C++11, but it is illegal syntax in previous versions of the C++ standard.  So should the code be updated to use the less restrictive style, or left as is? Updating the existing code is trivial using a regex,. 

For users of C++11, the change is unnecessary, but without it, users of earlier versions of C++ will get compiler errors when including the offending files. Suggestions?

michaelld:
hmm ... GR38 -requires- C++11 now, while GR37 doesn't. Both codebases are supposed to be C++11 compliant though ... so, honestly, it's a matter of coding style more than C++ compliance.

My coding style for these is to include the space between. We're trying to use `clang-format`, but I don't know if it has a default or setting in this regard. Let me query the GR core devs & see what our general preference is.

michaelld:
Ok so a quick review of GR master shows that the change from prior-to-C++11 `> >` (with the space) to `>>` was done intentionally here; https://github.com/gnuradio/gnuradio/commit/f7bbf2c1d8d780294f3e016aff239ca35eb6516e

Hence, for better or worse, I was wrong about thinking this change is desirable. _I_ would code using extra spaces, because I think it reads more clearly.

michaelld:
Thus, coming back to this PR: Why is this change required? Does the compiler generate warnings or errors when compiling C++ code that uses these headers? Are you building an OOT that doesn't require C++11 compliance, and including these headers? Thinking OOTs should have to require C++11 compliance if GR does ... but not sure.

seeAppRun:
> Thus, coming back to this PR: Why is this change required? Does the compiler generate warnings or errors when compiling C++ code that uses these headers? Are you building an OOT that doesn't require C++11 compliance, and including these headers? Thinking OOTs should have to require C++11 compliance if GR does ... but not sure.

Yes, I ran into this issues building OOT and including GR headers. I agree, if C++11 is a GR requirement, then I think the code is OK as is. Though I personally also prefer the '> >' style for clarity.

michaelld:
OK good to know. Let me -really- query the GR devs & see what they prefer. Just because the clang-tidy is doing no space between these doesn't mean it can't be changed if our preferences are otherwise.

michaelld:
FYI we did discuss this, and there is no clear consensus on the formatting. There was clear consensus that in GR38+ we should force C++11 during "find" ... such that the issue this PR addresses should no longer be an issue for OOT modules (though obviously -after- GR is found the OOT can always unset the c++ standard, but that's not our fault & we've done what we can do in this regard to notify the caller).

michaelld:
So ... longer story shorter: I have modifications coming to GnuradioConfig.cmake.in anyway, and I will add the c++11 standard check and setting to those changes. At that time, this PR will no longer be necessary, and I will note so in the PR discussion.

PR #2784, opened 2019-09-03
Uhd: Add cpp generation to gr-uhd

seeAppRun:
Updated gen_uhd_usrp_blocks.py to create .yml files which support c++ code generation.


mbr0wn:
Merged manually. Fixed commit messages.

PR #2790, opened 2019-09-05
grc: embedded python parameter values disappear

### LINKED ISSUE #2765: grc: embedded python parameter values disappear
When using an embedded python block in a flowgraph, editing the block causes the values for the parameters to be cleared out when the block is changed.

Steps to reproduce:
1) Create an embedded block in the flowgraph
2) Give the block properties

image

3) assign the properties to variables in the flowgraph:

image

4) Edit the block's python code and save

The values entered are wiped away:

image

cbdonohue:
grc: embedded python parameter values disappear
fixes #2765

### DIFF
--- a/grc/core/blocks/embedded_python.py
+++ b/grc/core/blocks/embedded_python.py
@@ -149,10 +149,10 @@ class EPyBlock(Block):
     def _update_params(self, params_in_src):
         param_factory = self.parent_platform.make_param
         params = {}
-        for param in list(self.params):
-            if hasattr(param, '__epy_param__'):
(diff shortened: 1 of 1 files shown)


mormj:
Tried a flowgraph that originally exposed this problem, and it has gone away.  Thanks @cbdonohue !!

marcusmueller:
awesome, Chris! Can you get into contact with @bhilburn about getting you a CLA? You're becoming a regular contributor, and for larger changes we need a CLA.

bhilburn:

PR #2795, opened 2019-09-09
grc: option toggle the show id on all blocks

### LINKED ISSUE #2780: grc: add ability to force blocks to show id
There are some blocks where the show_id field is not set, nor is it desired to be set, but I might want to see/change the block id in GRC. Could be for various reasons

1) clean up the automatic naming so that the generated .py is cleaner
2) reference the block by id from some other part of the flowgraph
3) the show_id didn't get set properly in the yml, and as a workaround I can change it, not wait until it is fixed.

This could be a context menu --> show block id

Thoughts?

mormj:
Addresses the issue of losing access to the block id on many/most blocks
though there are scenarios where you may need to see them in GRC

Added a menu item tied into the grc prefs that will show the block ids

Fixes #2780


dl1ksv:
Just a thought:
Sholud this be handled on a flowgraph basis or on a general basis? In the later case this should be part of grc.conf. 
The grc.conf could be edited by an text editor. 
But I think it could make sense to have a general configuration gui to handle the different configuration options.

mormj:
This PR handles it on a general basis as a grc.conf preference line:
image

and applies to all currently open flowgraphs (but nothing related to the option is saved in the flowgraph file).  It is solely a visual thing.

@dl1ksv - does this meet your needs in this way?

dl1ksv:
@mormj I think your proposal is the quickest way to solve the underlying issue.
I myself change the blocks yml file if needed.

But in the long run  this option  should be part of the grc.conf file and if I see all the possible configuration files and options I'm curious if we don't need a configuration editor.

mbr0wn:
Works as advertised.

PR #2797, opened 2019-09-10
grc/cppgen: C++ code generation fixes

seeAppRun:
Fixed a few bugs in processing C++ includes and type translation,

Fixed generation of parameter declarations in C++ class headers (Please confirm this is the intended use. This change now produces compilable code when using parameters).

Modified mako templates to add parameters to C++ flowgraph class constructor and instantiation


grc/blocks/options.block.yml
```diff
@@ -51,7 +51,7 @@ parameters:
 -   id: cmake_opt
     label: CMake options
     dtype: string
-    default: ''
+    default: '-DCMAKE_CXX_STANDARD=11'
```
seeAppRun: Add -std=c++11 to default CMake options as gnuradio requires C++ >= 11

grc/core/blocks/_build.py
```diff
@@ -64,7 +64,7 @@ def build(id, label='', category='', flags='', documentation='',
 
     cpp_templates = cpp_templates or {}
     cls.cpp_templates = MakoTemplates(
-        includes=cpp_templates.get('includes', ''),
+        includes=cpp_templates.get('includes', []),
```
seeAppRun: includes is a python list not a string

grc/core/blocks/block.py
```diff
@@ -407,7 +407,7 @@ def make_callback(callback):
 
         return [make_callback(c) for c in self.cpp_templates.render('callbacks')]
 
-    def decide_type(self):
+    def format_expr(self, py_type):
         """
         Evaluate the value of the variable block and decide its type.
```
seeAppRun: Modified this method to take the parameter type as an argument. This way the rvalue expression can be formatted appropriately. 

i.e. given the following flowgraph parameters:
- foo = 2
- bar = 3.5
- foobar = foo*bar 

Python generated code will have a variable foobar of type float initaliazed to 7.0. 
However c++ generated code will have the following:
std::string foobar = "foo*bar";

With the changes, c++ generated code now looks like this:
int foo = 2;
double bar = 3/5;
double foobar = foo*bar;

grc/core/blocks/block.py
```diff
@@ -417,46 +417,65 @@ def decide_type(self):
...
 
-                        except IndexError: # empty list
-                            return 'std::vector<std::string>'
+                    except IndexError: # empty dict
+                        return 'std::map<std::string, std::string>'
 
-                        else:
-                            return 'std::vector<' + list_type + '>'
+                    else:
+                        return 'std::map<' + key_type + ', ' + val_type +'>'
```
seeAppRun: Added support for std::map<> types

grc/core/generator/cpp_templates/flow_graph.cpp.mako
```diff
@@ -13,6 +13,10 @@ GNU Radio version: ${version}
 ********************/
 
 #include "${flow_graph.get_option('id')}.hpp"
+% if flow_graph.get_option('realtime_scheduling'):
+#include <gnuradio/realtime.h>
+% endif
+
```
seeAppRun: Fixed compiler error due to missing include

grc/core/generator/cpp_templates/flow_graph.cpp.mako
```diff
@@ -25,12 +29,20 @@ class_name = flow_graph.get_option('id') + ('_' if flow_graph.get_option('id') =
...
+  initializer_str = '\n: ' + initializer_str
 %>\
 
 % if generate_options == 'no_gui':
-${class_name}::${class_name} (${param_str}) {
+${class_name}::${class_name} (${param_str}) ${initializer_str} {
 % elif generate_options == 'qt_gui':
-${class_name}::${class_name} (${param_str}) : QWidget() {
+${class_name}::${class_name} (${param_str}) : ${initializer_str} { 
+
```
seeAppRun: Added initializer list to c++ constructor

grc/core/generator/cpp_templates/flow_graph.cpp.mako
```diff
@@ -140,7 +148,7 @@ int main (int argc, char **argv) {
     % endif
 
     % if generate_options == 'no_gui':
-    ${class_name}* top_block = new ${class_name}();
+    ${class_name}* top_block = new ${class_name}(${param_str_without_types});
```
seeAppRun: Pass parameters to c++ class constructor

grc/core/generator/cpp_top_block.py
```diff
@@ -263,15 +263,37 @@ def _variable_types(self):
...
+        try:
+          exec(prog, namespace)
+          var_types = namespace['get_decl_types']()
+        except Exception as excp:
+          print('Failed to get parameter lvalue types: %s' %(excp))
+        
+
+        # Format the rvalue of each variable expression
+        for var in variables:
+            var.format_expr(var_types[str(var.params['id'].value)])
```
seeAppRun: This modification determines the python type of variables by evaluating them collectively. The type is then passed to var.format_expr() to decide the C++ type. This change allows C++ code to be consistent with Python code in which variables can be to used in the initialization of other variables. 
See also changes to grc/core/blocks/_build.py.

grc/core/generator/cpp_top_block.py
```diff
@@ -263,15 +263,37 @@ def _variable_types(self):
...
 
     def _parameter_types(self):
         fg = self._flow_graph
         parameters = fg.get_parameters()
 
         for param in parameters:
-            type_translation = {'eng_float' : 'double', 'intx' : 'int', 'std' : 'std::string'};
+            type_translation = {'eng_float' : 'double', 'intx' : 'int', 'str' : 'std::string'};
```
seeAppRun: Fixed typo: 'std' should be 'str'
Python 'str' : C++ 'std::string'

haakov:
Awesome! I've been busy moving abroad for the last few months and haven't been able to give the C++ code generation any attention lately. I'll look at it later today.

seeAppRun:
I'm glad to get some feedback on the changes. This is a work-in-progress as
there are a lot of issues to fix with the C++ code generation. These
changes don't address everything but I think it is a good start.
I'm not an expert in Python or Yaml, so some of the changes may not be as
elegant as they could be. My goal has been to get the C++ generation to
produce compilable/runnable code with the same semantics as the python
flowgraphs.
I have also enabled  C++ generation in several other gnuradio blocks (e.g.
pull request #2898). I will push the other C++ changes (gr-analog,
gr-zeromq, etc.) as I get time. But most of those changes are dependent on
this pull request.

Thanks,
Terry

gr-blocks/grc/blocks_repeat.block.yml
```diff
@@ -42,7 +42,7 @@ templates:
 cpp_templates:
     includes: ['#include <gnuradio/blocks/repeat.h>']
     declarations: 'blocks::repeat::sptr ${id};'
-    make: 'this->${id} = blocks::repeat::make(${type.size}*${vlen}, ${interp});'
+    make: 'this->${id} = blocks::repeat::make(${type.size})*${vlen}, ${interp});'
```
haakov: Are you sure about this?

grc/core/generator/cpp_top_block.py
haakov: Great catch, thanks

gr-blocks/grc/blocks_repeat.block.yml
seeAppRun: I know this looks strange, but if you look below at:
```
translations:
        gr.sizeof_: 'sizeof('
```
The ${type.size} causes a replacement of  **gr.sizeof** with **sizeof(**, which is missing a closing parens.

seeAppRun:
Håkon,
I wanted to let you know I'm currently working on cleaning up and fixing
c++ generation of variables in block.py and cpp_top_block.py. In my initial
push, I did not fully understand how all the pieces worked. If you haven't
begun reviewing those files yet, I'd ask you wait a day or so and I'll
update those files.
The new changes will be much simpler and more correct. In short, I've
modified the code to evaluate variable blocks to determine the appropriate
C++ type, even when the is a compound type, such as a dictionary containing
lists of complex numbers.

Thanks,
Terry

grc/blocks/options.block.yml
seeAppRun: I wonder if DCMAKE_CXX_STANDARD=11 option should go in the CMakeLists.txt.mako file instead?

haakov:
No problem! I'll wait a couple days.

gr-blocks/grc/blocks_repeat.block.yml
seeAppRun: After looking at this more closely, I found that the translation is not needed at all. It is handled implicitly in cpp_top_block.py. So I removed it from this file because the unmodified translation generates non-compilable code (missing a closing parens) i.e. 

`sizeof(gr_complex` 

Note there are at least 40 other YAML files with this broken translation. I'll address those in a separate pull request since this one is already fairly involved.

grc/blocks/options.block.yml
mbr0wn: Yeah I think it should.

mbr0wn:
Looks like this needs some attention/history cleanup? @seeAppRun are you still working on this?

seeAppRun:
This work includes all the issues I've addressed thus far, so I don't have
anything new to add, unless review comments suggest something needs to
change.

What exactly is meant by history cleanup?

grc/core/generator/cpp_top_block.py
```diff
@@ -330,6 +382,11 @@ def make_port_sig(port):
             # Remove the virtual connection
             connections.remove(connection)
 
+        # Remove virtual sink connections too
+        virtual = [c for c in connections if isinstance(c.sink_block, blocks.VirtualSink)]
+        for connection in virtual:
+            connections.remove(connection)
+
```
haakov: I think we should consider changing this so that it matches the corresponding code in top_block.py:

 https://github.com/gnuradio/gnuradio/blob/590d73242d5755d3b98fc86e2f8f46ae4d9906a8/grc/core/generator/top_block.py#L252-L264

Both work, but it seems cleaner to implement them identically since they are responsible for completing the _exact_ same task

grc/core/generator/FlowGraphProxy.py
```diff
@@ -160,7 +160,7 @@ def links(self):
         Returns:
             a list of GNU Radio modules
         """
-        return [block.cpp_templates.render('link') for block in self.iter_enabled_blocks()]
+        return [block.cpp_templates.render('link') for block in self.iter_enabled_blocks() if not (block.is_virtual_sink() or block.is_virtual_source())]
```
haakov: What problem does this solve exactly?

I'm asking because if this change (and the other virtual block-related changes you've made in this PR) is necessary for the virtual blocks, then I'm wondering if it might also be necessary for other blocks such as the embedded Python blocks.

seeAppRun:
Unfortunately I'm on travel at the moment and don't have access to a system
to review this change now, but I recall it is necessary to prevent
attempting to generate connection code for virtual blocks, which will cause
the generation to fail.
I can address comments in greater detail in a few days.

gr-blocks/grc/blocks_repeat.block.yml
```diff
@@ -45,7 +45,5 @@ cpp_templates:
     make: 'this->${id} = blocks::repeat::make(${type.size}*${vlen}, ${interp});'
     callbacks:
     - set_interpolation(${interp})
-    translations:
```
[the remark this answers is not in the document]
seeAppRun: Yes, I think this was also fixed in PR #2935. It was fixed here before I found that the translation issue was widespread.

grc/core/generator/FlowGraphProxy.py
seeAppRun: Without this modification, generating C++ with Virtual blocks produces the following error:
`Generate Error: 'VirtualSink' object has no attribute 'cpp_templates'`

michaelld:
@seeAppRun is this PR finalized from your perspective? If so, let's get it reviewed! If not, please take the time to fix it up. Thx!

seeAppRun:
I don't currently have any other changes to add, so I guess it's ready for review. I know the commit history needs to be cleaned up as the some files have changed multiple times, but I'm not sure the best way to go about that.

mbr0wn:
Let's worry about git history last. I might be able to help. @haakov, once you give a green checkmark I'll work with @seeAppRun to get it merged.

seeAppRun:
@haakov I also had issues with the `qwt_symbol.h` but don't remember how I resolved it. Hopefully it wasn't a 'local' fix.
I'll address the conncection code in cpp_top_block.py and look at the Qt sink issue and follow up tomorrow.

seeAppRun:
@haakov I pushed an update to cpp_top_block.py to mirror the virtual connection handling code in top_block.py

**1.**
I have the same issue with `<qwt_global.h`>. I think that `include_directories` needs to be updated in `CMakeLists.txt.mako`. I tried adding `QWT_INCLUDE_DIRS` to the includes, and I added a find_package(...) for Qwt, but that didn't work either. Perhaps related to your findings?  FWIW, I've tried this on CentOS and Ubuntu. 

**2.**
I also saw the issue with single vs. double quotes. The same thing also applies when you use a string value in a _variable_ or _parameter_ block, though in these cases a `std::string` is declared.

mbr0wn:
Merged manually.

velichkov:
Three grc tests are failing as a result of 64ee8030682

```
         if 'cpp' in self.flags and not (self.is_virtual_source() or self.is_virtual_sink()):
 >           self.orig_cpp_templates = self.cpp_templates # The original template, in case we have to edit it when transpiling to C++
 E           AttributeError: 'DummyBlock' object has no attribute 'cpp_templates'
 
 ../../../grc/core/blocks/block.py:81: AttributeError
```
To me it's not clear what the purpose of this `DummyBlock` is and so I'm not sure whether a correct fix is to add a `cpp_templates` to it.

<details>
<summary>And here is the  full output of test_generator</summary>

```
154: ________________________________ test_generator ________________________________
154: 
154:     def test_generator():
154:         # c&p form compiler code.
154:         # todo: make this independent from installed GR
154:         grc_file = path.join(path.dirname(__file__), 'resources', 'test_compiler.grc')
154:         out_dir = tempfile.gettempdir()
154:     
154:         platform = Platform(
154:             name='GNU Radio Companion Compiler',
154:             prefs=None,
154:             version='0.0.0',
154:         )
154:         platform.build_library()
154:     
154: >       flow_graph = platform.make_flow_graph(grc_file)
154: 
154: ../../../grc/tests/test_generator.py:29: 
154: _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
154: ../../../grc/core/platform.py:410: in make_flow_graph
154:     fg.import_data(data)
154: ../../../grc/core/FlowGraph.py:410: in import_data
154:     self.new_block(block_id='_dummy', missing_block_id=block_id, **block_data)
154: ../../../grc/core/FlowGraph.py:300: in new_block
154:     block = self.parent_platform.make_block(self, block_id, **kwargs)
154: ../../../grc/core/platform.py:418: in make_block
154:     return cls(parent, **kwargs)
154: ../../../grc/core/blocks/dummy.py:24: in __init__
154:     super(DummyBlock, self).__init__(parent=parent)
154: _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
154: 
154: self = block[], parent = <grc.core.FlowGraph.FlowGraph object at 0x7fb7516d0a50>
154: 
154:     def __init__(self, parent):
154:         """Make a new block from nested data."""
154:         super(Block, self).__init__(parent)
154:         param_factory = self.parent_platform.make_param
154:         port_factory = self.parent_platform.make_port
154:     
154:         self.params = collections.OrderedDict(
154:             (data['id'], param_factory(parent=self, **data))
154:             for data in self.parameters_data
154:         )
154:         if self.key == 'options':
154:             self.params['id'].hide = 'part'
154:     
154:         self.sinks = [port_factory(parent=self, **params) for params in self.inputs_data]
154:         self.sources = [port_factory(parent=self, **params) for params in self.outputs_data]
154:     
154:         self.active_sources = []  # on rewrite
154:         self.active_sinks = []  # on rewrite
154:     
154:         self.states = {'state': True, 'bus_source': False, 'bus_sink': False, 'bus_structure': None}
154:     
154:         if 'cpp' in self.flags and not (self.is_virtual_source() or self.is_virtual_sink()):
154: >           self.orig_cpp_templates = self.cpp_templates # The original template, in case we have to edit it when transpiling to C++
154: E           AttributeError: 'DummyBlock' object has no attribute 'cpp_templates'
154: 
154: ../../../grc/core/blocks/block.py:81: AttributeError
154: ----------------------------- Captured stdout call -----------------------------
```
</details>

seeAppRun:
I'm not sure what the DummyBlock is for either, but looking at the class (`grc/core/blocks/dummy.py`) it appears to represent a missing block.  Perhaps the method `enabled()` could be used in the conditional to avoid trying to generate C++ for the block in question. 

This method explicitly returns False in class DummyBlock :
`    
def enabled(self):
        return False
`

seeAppRun:
@velichkov 
I did a quick test and it does seem that DummyBlock can represent a missing block.
I renamed a block YAML file to force a missing block and got the same error as above. 

However, adding `self.enabled` (note this is actually a property rather than a method) to the conditional at `grc/core/blocks/block.py:80` does eliminate the error.  No code is generated for the missing block. 
I don't know if this is the correct solution as I'm not sure of the expected behavior for missing blocks.

velichkov:
Hi @seeAppRun,

Thanks for looking into this. I can confirm that adding `and self.enabled` to `block.py:80` fixes this problem. The behavior with this fix when there is a missing block seems to be the same as in gnuradio 3.7 - e.g. I'm able to open the flowgraph and I'm seeing all the missing blocks marked in red and I'm not able to build or run the flowgraph as it is invalid. Without this fix I don't see the flowgraph at all.

In the grc tests some of the blocks are missing because some more paths have to be added to the `GRC_BLOCKS_PATH` variable in `grc/tests/CMakeLists.txt` and we will probably need to add a test for flowgraph with missing blocks.

ghostop14:
Hi folks, I'm running into a related issue with that DummyBlock notification with the latest master branch (refreshed 2 days ago).  When a flowgraph has a missing block I'm getting that same non-descript "Error: 'DummyBlock' object has no attribute 'cpp_templates'" message.  It doesn't indicate the issue is a missing block or what the requested block id is (could be helpful in identifying what's missing), and won't open the flowgraph.  Seems like an error message including the requested block id could be helpful and let the flowgraph open with the "missing block" blocks present rather than just not opening?

PR #2803, opened 2019-09-17
update cloned port's dtype

glennlive:
Fix a bug that left the type of some of multi-port (the cloned ports) unchanged when incrementing or decrementing block type.


glennlive:
This bug also applies to 3.8, so back-port requested.  Perhaps there is a better one-liner which can fix this bug?

PR #2787, opened 2019-09-04
Always include gnuradio/rpcregisterhelpers.h in basic_block.h

carlesfernandez:
If `GR_CTRLPORT` is not defined, then rpcbasic_sptr is not defined in basic_block.h, which breaks compilation of external applications including basic_block.h but not having `GR_CTRLPORT` defined.

This PR removes the `#ifdef GR_CTRLPORT` from around `#include <gnuradio/rpcregisterhelpers.h>`.

See #2785

PR #2805, opened 2019-09-18
Fix cmake ctrlport basic block rpcregisterhelpers

michaelld:
Replacement of https://github.com/gnuradio/gnuradio/pull/2787 to fix building issue, by not moving the header include & also adding a header include. This PR should be replicated for master.

### DIFF
--- a/gnuradio-runtime/include/gnuradio/basic_block.h
+++ b/gnuradio-runtime/include/gnuradio/basic_block.h
@@ -38,9 +38,7 @@
 #include <map>
 #include <string>
 
-#ifdef GR_CTRLPORT
 #include <gnuradio/rpcregisterhelpers.h>
-#endif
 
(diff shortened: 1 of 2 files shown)


michaelld:
Looks like the CI build failures are not related to this change.

michaelld:
@777arc Says verified that this change allows building when ENABLE_GR_CTRLPORT=FALSE , which fixes a critical bug in the current GR38 and master build system.

777arc:
Yep, I usually build GR with controlport disabled because it really reduces build time, and this prevents it from erroring out a quarter of the way into the build.

carlesfernandez:
I confirm that this PR works for building gnuradio and gnss-sdr against it either ENABLE_GR_CTRLPORT is ON or OFF.

PR #2806, opened 2019-09-19
runtime: Fix get_tags_in_range for (end - start) > delay.

ryanvolz:
Previously, stream tags would be lost (not propagated downstream) when a block has a tag delay set and it is called with a number of input samples that is less than the delay. Since this happens depending on the scheduler and the number of samples provided, tags might mostly work or seemingly randomly disappear. The problem occurs because a wrong correction is included for underflows in unsigned arithmetic that limits the tags propagated to a smaller sample window than desired. Correctly detecting the underflow and setting the value to 0 rather than abs_start or abs_end fixes the problem.

For a real example, let the `d_attr_delay=209`, `abs_start=0`, and `abs_end=200` with a tag at input sample `0`. Then tags from `min(abs_start, abs_start - d_attr_delay) == 0` to `min(abs_end, abs_end - d_attr_delay) == 200` are iterated through, including the tag at `0` (`item_time == 209`), but it is not propagated to the output yet since `item_time >= abs_end` (`209 >= 200`). On the next call to the block, with `abs_start=200` and `abs_end=400`, the tag at input sample `0` is ignored because `min(abs_start, abs_start - d_attr_delay) == 200` and the iterator skips it. The correct calculation for the lower bound of the tag iterator would result in `0` and find the tag since `(abs_start - d_attr_delay) == -9` -> `0` when constrained to unsigned values. With the proper lower bound (and upper bound corrected as well), the tag at `0` is found with an `item_time == 209` which falls between `abs_start` and `abs_end` and is propagated to output sample `209`.

### DIFF
--- a/gnuradio-runtime/lib/buffer.cc
+++ b/gnuradio-runtime/lib/buffer.cc
@@ -298,8 +298,6 @@ buffer_reader::buffer_reader(buffer_sptr buffer, unsigned int read_index, block_
       d_attr_delay(0)
 {
     s_buffer_reader_count++;
-
-    buffer->d_max_reader_delay = 0;
 }
 
 buffer_reader::~buffer_reader()
--- a/gnuradio-runtime/lib/buffer.cc
+++ b/gnuradio-runtime/lib/buffer.cc
@@ -340,11 +340,20 @@ void buffer_reader::get_tags_in_range(std::vector<tag_t>& v,
 {
     gr::thread::scoped_lock guard(*mutex());
 
+    uint64_t lower_bound = abs_start - d_attr_delay;
+    // check for underflow and if so saturate at 0
+    if (lower_bound > abs_start)
+      lower_bound = 0;
+    uint64_t upper_bound = abs_end - d_attr_delay;
+    // check for underflow and if so saturate at 0
+    if (upper_bound > abs_end)
+      upper_bound = 0;
+
     v.clear();
     std::multimap<uint64_t, tag_t>::iterator itr =
-        d_buffer->get_tags_lower_bound(std::min(abs_start, abs_start - d_attr_delay));
+        d_buffer->get_tags_lower_bound(lower_bound);
     std::multimap<uint64_t, tag_t>::iterator itr_end =
-        d_buffer->get_tags_upper_bound(std::min(abs_end, abs_end - d_attr_delay));
+        d_buffer->get_tags_upper_bound(upper_bound);
 
     uint64_t item_time;
     while (itr != itr_end) {


mormj:
This seems a very reasonable improvement.  Do you have a simple flowgraph that exhibited the original issue?

ryanvolz:
This seems to demonstrate it. The Tag Debug block shows the tags on input stream 0 but never shows the delayed tags on input stream 1.

```python
from gnuradio import blocks
from gnuradio import eng_notation
from gnuradio import filter
from gnuradio import gr
from gnuradio.eng_option import eng_option
from gnuradio.filter import firdes
from optparse import OptionParser
import pmt


class test_tag_delay(gr.top_block):

    def __init__(self):
        gr.top_block.__init__(self, "Test Tag Delay")

        # Blocks
        self.fir_filter_xxx_0 = filter.fir_filter_ccc(1, ([1]))
        self.fir_filter_xxx_0.declare_sample_delay(1024)
        self.blocks_throttle_0 = blocks.throttle(gr.sizeof_gr_complex*1, 1000,True)
        (self.blocks_throttle_0).set_min_output_buffer(512)
        (self.blocks_throttle_0).set_max_output_buffer(512)
        self.blocks_tags_strobe_0 = blocks.tags_strobe(gr.sizeof_gr_complex*1, pmt.intern("TEST"), 1000, pmt.intern("strobe"))
        self.blocks_tag_debug_0 = blocks.tag_debug(gr.sizeof_gr_complex*1, '', ""); self.blocks_tag_debug_0.set_display(True)



        # Connections
        self.connect((self.blocks_tags_strobe_0, 0), (self.blocks_throttle_0, 0))
        self.connect((self.blocks_throttle_0, 0), (self.blocks_tag_debug_0, 0))
        self.connect((self.blocks_throttle_0, 0), (self.fir_filter_xxx_0, 0))
        self.connect((self.fir_filter_xxx_0, 0), (self.blocks_tag_debug_0, 1))


def main(top_block_cls=test_tag_delay, options=None):

    tb = top_block_cls()
    tb.start()
    tb.wait()


if __name__ == '__main__':
    main()
```

mormj:
I'm getting the same output from the tag delay block before and after applying the commit.

```
python3 .debug/tag_delay.py 

set_min_output_buffer on block 2 to 512

Tag Debug: 
Input Stream: 00
  Offset: 1000  Source: tags_strobe0     Key: strobe   Value: TEST
Input Stream: 01

Tag Debug: 
Input Stream: 00
  Offset: 2000  Source: tags_strobe0     Key: strobe   Value: TEST
Input Stream: 01

Tag Debug: 
Input Stream: 00
  Offset: 3000  Source: tags_strobe0     Key: strobe   Value: TEST
Input Stream: 01

Tag Debug: 
Input Stream: 00
  Offset: 4000  Source: tags_strobe0     Key: strobe   Value: TEST
Input Stream: 01

Tag Debug: 
Input Stream: 00
  Offset: 5000  Source: tags_strobe0     Key: strobe   Value: TEST
Input Stream: 01

Tag Debug: 
Input Stream: 00
  Offset: 6000  Source: tags_strobe0     Key: strobe   Value: TEST
Input Stream: 01
```

ryanvolz:
Hmmm, this patch fixes my real-world example so perhaps the simple test script is bad or has revealed another bug. I'll do a little more investigation on why the test script doesn't work even with the patch.

gnuradio-runtime/lib/buffer.cc
```diff
@@ -340,11 +340,20 @@ void buffer_reader::get_tags_in_range(std::vector<tag_t>& v,
 {
     gr::thread::scoped_lock guard(*mutex());
 
+    uint64_t lower_bound = abs_start - d_attr_delay;
+    // check for underflow and if so saturate at 0
+    if (lower_bound > abs_start)
+      lower_bound = 0;
```
michaelld: My wondering is whether the `lower_bound` should be 0 for either case. Need to follow the code for a bit to see how the methods `get_tags_*_bound` use the argument.

ryanvolz:
I ran the above test script with gdb and it does seem to trigger a different bug. By the time that get_tags_in_range is called for the FIR filter block for samples spanning 1000 + 1024 = 2024 (first tag offset + sample delay, for me this happened with abs_start=1989 and abs_end=2192), the tag at offset 1000 had been removed from the buffer and the one at offset 2000 was at the front of the list. So it seems that there is more to be done in ensuring that the tag buffer plays nicely with large tag delays, but I'm not familiar with how the buffered is handled to know more at this point.

I'll see if I can come up with a script that only triggers the problem that this change is meant to solve.

ryanvolz:
Here's a simpler script the demonstrates a dropped tag:
```python
from gnuradio import blocks
from gnuradio import eng_notation
from gnuradio import filter
from gnuradio import gr
from gnuradio.eng_option import eng_option
from gnuradio.filter import firdes
from optparse import OptionParser
import pmt


class test_tag_delay(gr.top_block):
    def __init__(self):
        gr.top_block.__init__(self, "Test Tag Delay")

        # Blocks
        self.blocks_throttle_0 = blocks.throttle(gr.sizeof_gr_complex*1, 96,True)
        self.blocks_throttle_0.declare_sample_delay(97)
        self.blocks_throttle_0.set_min_noutput_items(96)
        self.blocks_throttle_0.set_max_noutput_items(96)
        self.blocks_tags_strobe_0 = blocks.tags_strobe(gr.sizeof_gr_complex*1, pmt.intern("TEST"), 94, pmt.intern("strobe"))
        self.blocks_tags_strobe_0.set_min_noutput_items(96)
        self.blocks_tags_strobe_0.set_max_noutput_items(96)
        self.blocks_tag_debug_0 = blocks.tag_debug(gr.sizeof_gr_complex*1, '', ""); self.blocks_tag_debug_0.set_display(True)

        # Connections
        self.connect((self.blocks_tags_strobe_0, 0), (self.blocks_throttle_0, 0))
        self.connect((self.blocks_tags_strobe_0, 0), (self.blocks_tag_debug_0, 0))
        self.connect((self.blocks_throttle_0, 0), (self.blocks_tag_debug_0, 1))


def main(top_block_cls=test_tag_delay, options=None):
    tb = top_block_cls()
    tb.start()
    tb.wait()


if __name__ == '__main__':
    main()
```
The Tag Debug block should show a tag on input stream 1 with offset 191, but instead the first delayed tag that makes it is at offset 285:
```
$ python test_tag_delay.py 

Tag Debug: 
Input Stream: 00
  Offset: 94  Source: tags_strobe0     Key: strobe   Value: TEST
Input Stream: 01

Tag Debug: 
Input Stream: 00
  Offset: 188  Source: tags_strobe0     Key: strobe   Value: TEST
Input Stream: 01

Tag Debug: 
Input Stream: 00
  Offset: 282  Source: tags_strobe0     Key: strobe   Value: TEST
Input Stream: 01
  Offset: 285  Source: tags_strobe0     Key: strobe   Value: TEST
```

If you change the sample delay from 97 to 99, get_tags_in_range will hang in an infinite loop:
```
$ python test_tag_delay.py 

Tag Debug: 
Input Stream: 00
  Offset: 94  Source: tags_strobe0     Key: strobe   Value: TEST
Input Stream: 01
^CTerminated
```
Both of these behaviors are fixed with the patch. The original test script in https://github.com/gnuradio/gnuradio/pull/2806#issuecomment-545102123 still shows another bug that needs to be tracked down.

ryanvolz:
I just added a second commit that fixes the original test script in https://github.com/gnuradio/gnuradio/pull/2806#issuecomment-545102123. The problem was that `d_max_reader_delay` was getting set to 0 by instantiation of another `buffer_reader` after it had been set to the correct value by `buffer_reader::declare_sample_delay`. I can't think of a purpose for setting `d_max_reader_delay` to 0 in `buffer_reader::buffer_reader`, so I deleted that line and it fixed the problem.

gnuradio-runtime/lib/buffer.cc
mbr0wn: You mean for the `else` case? Or for upper_bound?

PR #2808, opened 2019-09-23
add set_max_output_buffer to block gateway

solsenp:
Just a small change to let you call set_max_output_buffer from pure Python blocks


marcusmueller:
I like this, too! @solsenp, have we (or @bhilburn and you) talked about getting you a CLA? Your changeset is exactly 1 line of code below the threshold from which we need you to have a CLA.

If you'd want to contribute more in the future, getting a CLA would be very desirable.

bhilburn:

PR #2814, opened 2019-09-24

mormj:

One example is synchronizing multiple USRP objects - sometimes you want
different sync than what is offered in the multi-usrp object, so you can
put a bit of code in the snippet block to do the custom synchronization

Singleton block to contain all the snippets (moved out of the options block so that the snippets can be copy and pasted).

Split _multiline and _multiline_external dtypes from string in the
processing as code block and snippet are getting evaluated multiple
times.  If the code contains sleep or print, this wreaks havoc.

image

image

image

image

### DIFF
--- a/grc/blocks/grc.tree.yml
+++ b/grc/blocks/grc.tree.yml
@@ -10,6 +10,7 @@
   - bus_structure_source
   - epy_block
   - epy_module
+  - snippets_config
   - note
   - import
 - Variables:
--- a/grc/blocks/snippets_config.block.yml
+++ b/grc/blocks/snippets_config.block.yml
@@ -1,0 +1,57 @@
+id: snippets_config
+label: Python Snippets
+flags: [ python ]
+
+parameters:
+-   id: snippet_editor
+    label: Editor
+    dtype: enum
+    options: ['Dialog','External']
+    hide: 'part'
+-   id: snippet_top
+    label: Top
+    dtype: ${'_multiline_python_external' if snippet_editor == 'External' else '_multiline'}
+    hide: 'part'
+-   id: snippet_imports
+    label: Imports
+    dtype: ${'_multiline_python_external' if snippet_editor == 'External' else '_multiline'}
+    hide: 'part'
+-   id: snippet_variables
+    label: Variables
+    dtype: ${'_multiline_python_external' if snippet_editor == 'External' else '_multiline'}
+    hide: 'part'
+-   id: snippet_blocks
+    label: Blocks
+    dtype: ${'_multiline_python_external' if snippet_editor == 'External' else '_multiline'}
+    hide: 'part'
+-   id: snippet_connections
+    label: Connections
+    dtype: ${'_multiline_python_external' if snippet_editor == 'External' else '_multiline'}
+    hide: 'part'
+-   id: snippet_callbacks
+    label: Callbacks
+    dtype: ${'_multiline_python_external' if snippet_editor == 'External' else '_multiline'}
+    hide: 'part'
+-   id: snippet_main_beginning
+    label: Main (Beginning)
+    dtype: ${'_multiline_python_external' if snippet_editor == 'External' else '_multiline'}
+    hide: 'part'
+-   id: snippet_main_end
+    label: Main (End)
+    dtype: ${'_multiline_python_external' if snippet_editor == 'External' else '_multiline'}
+    hide: 'part'
+
+templates:
+    var_make: ${code}
+
+documentation: |-
+    Insert snippets of Python code directly into the flowgraph at the end of the specified section 
+
+    Indents will be handled upon insertion into the python flowgraph
+
+    Examples:
+    print("This is my flowgraph")
+
+    Will place the print statement in the generated .py file
+
+file_format: 1
--- a/grc/core/FlowGraph.py
+++ b/grc/core/FlowGraph.py
@@ -92,6 +92,16 @@ class FlowGraph(Element):
         parameters = [b for b in self.iter_enabled_blocks() if b.key == 'parameter']
         return parameters
 
+    def get_snippets_config(self):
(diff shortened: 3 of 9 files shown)


mormj:
@marcusmueller @skoslowski  - so what do we think about this one?

skoslowski:
Thanks for the ping, Marcus and I discussed this in person a while back and probably both thought the outer would comment here...
We both started out with +0 and discussed the simple vs. multiple block design or ours. However, we ended up questioning whether this is something that GRC really needs / should have a feature like this: 

As we understand, most of what is sensible to do can be done now by extending the generated class or replacing main() in a separate module (as well as have GRC run this instead of the generated code). E.g. if there really is a good use-case to inject some code between variables and blocks/connections we could do sth like
```python
class ...:
    def __init__(self):
        self.var = ....
       
        self.init_extra_vars()

        # blocks
        ...
        # connections
        ...

    def init_extra_vars(self):
        """override this to do some init between vars and blocks/connections"""
```

Maybe we're not seeing the strong use-cases you have in mind for this.

mormj:
Thanks @skoslowski - I'll try your suggestions tomorrow and see if I can do what I need to do without modifying GRC.  I fully appreciate the desire to not bring in unsupportable functionality. 

The main goal of this feature is to be able to add in those pieces of code _directly from GRC_ for which blocks cannot currently provide.  The main use case is doing USRP synchronization - all the injection points are probably not necessary for this particular case.  Was just looking to make the idea flexible.

mormj:
So here is the simple use case I have had in mind.  Say we have a 2 USRP setup and don't want to run the multi-usrp object for some reason, maybe to provide custom synchronization, or run each uhd block in its own thread:

image

I can create an embedded py module 'sync' which has the following code to do my synchronization:
```
def synchronize(usrp_list):
    print("Synchronization ... ")
    last_pps_time = usrp_list[0].get_time_last_pps()
    while uhd.time_spec_t.get_real_secs(last_pps_time) == uhd.time_spec_t.get_real_secs(usrp_list[0].get_time_last_pps()):
        time.sleep(0.1)

    for usrp in usrp_list:
        usrp.set_time_next_pps(uhd.time_spec_t(0.0))

    time.sleep(2)
    # Check the last pps time

    print("Checking 10 MHz lock")
    for usrp in usrp_list:
            ref_locked = usrp.get_mboard_sensor("ref_locked",0)
            print(ref_locked)

    print("Checking PPS synchronization")
    error_occurred = False
    for ii in range(2):
            last_pps_time = usrp_list[0].get_time_last_pps()

            while uhd.time_spec_t.get_real_secs(last_pps_time) == uhd.time_spec_t.get_real_secs(usrp_list[0].get_time_last_pps()):
                    time.sleep(0.1)

            for usrp in usrp_list:
                    last_pps = usrp.get_time_last_pps()
                    print("last_pps : %6.12f"%uhd.time_spec_t.get_real_secs(last_pps))

    # Wait for the next pps
    while uhd.time_spec_t.get_real_secs(last_pps_time) == uhd.time_spec_t.get_real_secs(usrp_list[0].get_time_last_pps()):
            time.sleep(0.1)

    last_pps = usrp.get_time_last_pps()

    cmd_time = last_pps + uhd.time_spec_t(2.0)

    for usrp in usrp_list:
            usrp.set_start_time(cmd_time)

```

So from my flowgraph, after the line
```
        self.connect((self.uhd_usrp_source_0, 0), (self.qtgui_time_sink_x_0, 0))
        self.connect((self.uhd_usrp_source_0_0, 0), (self.qtgui_time_sink_x_0, 1))

```

I want to insert the line 
```
sync.synchronize([self.uhd_usrp_source_0_0,self.uhd_usrp_source_0])
```

I know this can be easily done by editing the generated py after the fact, but the goal here is to be able to do everything from within GRC

I'm not seeing how this can be done currently without going back and editing the py file.

With the python snippets PR, I would just need to put the one line in the "connections" region (after connections are defined), and my synchronization would be contained in the grc and completely portable.

dkozel:
I regularily encounter times that I want to inject very small amounts of code into a GRC based application. This PR's functionality really nicely complements the existing Python Block, Python Module, and Import blocks.

skoslowski:
Okay, so studying your example, I observe that
- the ultimate goal here is to keep everything in the .grc-file
- you still have to leave GRC and go your editor to "implement" the snippet
- that snippet should be put in start() or main(), not __init__()

For comparison, here is what works today (following the approach above)
```python
import top_block as generated

def synchronize(usrp_list):
    ...

class my_top_block(generated.top_block):
    def start(self):
        synchronize([self.uhd_usrp_source_0_0,self.uhd_usrp_source_0])
        generated.top_block.start(self)  # only have 3.7 here =(

if __name__ == "__main__":
    generated.main(my_top_block)
```
I then set the `Run command` accordingly.

Now, this solution obviously violates the desire to contain everything in one file, however it makes the actual customization very clearly visible.

Thinking about this, an alternative design came to mind:
Since the snippet code is in an embedded module anyway, why generate a call directly.
It would work much like pytest plugins - if the module contains an object with a certain name a call to it would be generated at the appropriate moment
* a decorator ``@grc.on_start`` would be nice - however, don't like that runtime requirement
* e.g. with function ``modify_flowgraph(cls)`` that acts as a class decorator for e.g. ``top_block`` would allow one to do a whole bunch of stuff.

To me that feels a bit simpler overall, what do you think? 
@marcusmueller, thoughts?

dkozel:
@mormj, @marcusmueller, and I just talked at the hackfest and decided that going back to the general model from https://github.com/gnuradio/gnuradio/pull/2779 is an acceptable balance of simplicity and utility. The snippets will be wrapped into functions and the functions called from the desired area of code in the flowgraph's main function or inside specific top block functions.

mormj:
Closing in favor of #3169

PR #2819, opened 2019-09-27
qtgui: time sink config options for float message

### LINKED ISSUE #2796: qtgui: config options missing for float message input
maint-3.8 branch

The config options where you can set line colors, etc. are missing when the type is set to float message.  Complex message seems to work fine:

image

mormj:
In 3.8, the config options for float message input on qt gui time sink
disappeared.  This updates the .py file that generates the yml for this
block to always have 1 channel of config options for float message and 2
channels of config options for complex message.

The auto-generated .yml code is not the prettiest, but imo it is better than having
special case for line1, then special case for line2, etc. in the generation script

Fixes #2796


marcusmueller:
I agree! This is better all the way (and not necessarily pretty, but so be it).

PR #2820, opened 2019-09-27
.clang-tidy: more improvements

marcusmueller:
* use emplace now where possible


mbr0wn:
@marcusmueller I'm not super happy about this. I think the diff is too big in relation to its utility.

For example: Using emplace_back instead of push_back has its benefits. However, push_back actually takes rvalues, and when the underlying type is a POD type, then it doesn't matter if you're moving, doing rvalues or lvalues or whatnot. So you now have hundreds of lines changed for zero effect.

Similar with nullptr vs NULL. Sure, nullptr is better and more C++-y and so on. But in contexts like this:
```cpp
float *f = nullptr; // Or NULL
do_something(f);
```

It makes zero difference. The compiler doesn't transitively apply the nullptr-ness via the raw pointer.

`auto` vs not-auto is a subjective matter. I think there are cases where `auto` is a must (e.g. when there's make_shared on the RHS), but other cases are more subjective and I would like to give devs some freedom (TBH, I didn't read up on how exactly that rule gets executed in clang-tidy).

Overall, I would say changing 952 files for almost no benefit is not worth it. I would prefer we discuss a way how to use clang-tidy with GNU Radio and put it into our review process or something like that.

I will close this PR, because I think we need a bit of a discussion before we do something like this.

PR #2821, opened 2019-09-27
qtgui: usrp source gain callbacks not rendered

### LINKED ISSUE #2791: gr-uhd: usrp source gain callbacks not rendered to py (3.8)
In 3.7, creating a simple flowgraph with a QT GUI Range used as the gain setting on the USRP Source block, the following is rendered on the output py:

```
    def set_gain(self, gain):
        self.gain = gain
        self.uhd_usrp_source_0.set_gain(self.gain, 0)

        self.uhd_usrp_source_0.set_gain(self.gain, 1)

        self.uhd_usrp_source_0.set_gain(self.gain, 2)

        self.uhd_usrp_source_0.set_gain(self.gain, 3)
```

However, in 3.8, the callbacks for gain are not rendered, we get:
```
    def set_gain(self, gain):
        self.gain = gain
```

mormj:
In the 3.8 gen_uhd_usrp_blocks.py generated an invalid statement for the
set_gain callbacks, so they were never rendered to the output py.

Fixes #2791

### DIFF
--- a/gr-uhd/grc/gen_uhd_usrp_blocks.py
+++ b/gr-uhd/grc/gen_uhd_usrp_blocks.py
@@ -220,8 +220,10 @@ templates:
     % if sourk == 'source':
     -   ${'$'}{'set_rx_agc(True, ${n})' if context.get('rx_agc${n}')() == 'Enabled' else ''}
     -   ${'$'}{'set_rx_agc(False, ${n})' if context.get('rx_agc${n}')() == 'Disabled' else ''}
-    -   ${'$'}{'set_gain(${'$'}{${'gain' + str(n)}}, ${n})' if not bool(eval(context.get('norm_gain${n}')())) and context.get('rx_agc${n}')() != 'Enabled' else ''}
-    -   ${'$'}{'set_normalized_gain(${'$'}{${'gain' + str(n)}}, ${n})' if bool(eval(context.get('norm_gain${n}')())) and context.get('rx_agc${n}')() != 'Enabled' else ''}
+    -   |
+        ${'%'} if context.get('rx_agc${n}')() != 'Enabled':
+        self.${'$'}{id}.set_${'$'}{'normalized_' if bool(eval(context.get('norm_gain${n}')())) else ''}gain(${'$'}{${'gain' + str(n)}}, ${n})
+        ${'%'} endif
     % else:
     -   self.${'$'}{id}.set_${'$'}{'normalized_' if bool(eval(context.get('norm_gain${n}')())) else ''}gain(${'$'}{${'gain' + str(n)}}, ${n})
     % endif


jdemel:
I ran into the same issue.
I just tested your commit with 1 and multiple ports. The generated file looks correct now. The `set_gain` callbacks are included now.

gr-uhd/grc/gen_uhd_usrp_blocks.py
```diff
@@ -220,9 +220,11 @@
...
     -   ${'$'}{'set_rx_agc(True, ${n})' if context.get('rx_agc${n}')() == 'Enabled' else ''}
     -   ${'$'}{'set_rx_agc(False, ${n})' if context.get('rx_agc${n}')() == 'Disabled' else ''}
-    -   ${'$'}{'set_gain(${'$'}{${'gain' + str(n)}}, ${n})' if not bool(eval(context.get('norm_gain${n}')())) and context.get('rx_agc${n}')() != 'Enabled' else ''}
-    -   ${'$'}{'set_normalized_gain(${'$'}{${'gain' + str(n)}}, ${n})' if bool(eval(context.get('norm_gain${n}')())) and context.get('rx_agc${n}')() != 'Enabled' else ''}
-    % else:
+    -   |
+        ${'%'} if context.get('rx_agc${n}')() != 'Enabled':
+        self.${'$'}{id}.set_${'$'}{'normalized_' if bool(eval(context.get('norm_gain${n}')())) else ''}gain(${'$'}{${'gain' + str(n)}}, ${n})
+        ${'%'} endif
+   % else:
```
michaelld: There is a space removed before the `%` here ... please add it back!

ThomasHabets:
I applied the patch to my pybombs directory, and nothing changed. Am I missing something?

mormj:
If I create a simple flowgraph:

image

Before the patch, there are no usrp calls in the gain callbacks:
```

    def get_gain(self):
        return self.gain

    def set_gain(self, gain):
        self.gain = gain



```

After applying the patch and rebuilding, the set_gain callback makes calls to update the usrp object:
```
    def get_gain(self):
        return self.gain

    def set_gain(self, gain):
        self.gain = gain
        self.uhd_usrp_source_0.set_gain(self.gain, 0)
        self.uhd_usrp_source_0.set_gain(self.gain, 1)
```

Make sure after rebuilding, also to restart GRC or reload blocks because the changes are in the yaml.

gr-uhd/grc/gen_uhd_usrp_blocks.py
mormj: ah I see - that took me a minute :)  will fix

mormj: should be good now - nice catch!

ThomasHabets:
@mormj I take it your install is not built with pybombs? Do you know what step I need to do to rebuild after patching my `~/opt/pybombs/default/…` file with this pull request?

mormj:
my setup was initially built with pybombs, but for applying patches or updating to the latest code, I just go right to the src directories and recompile

For this case, I just pulled the patch manually in (this may not be the best way):
```
cd $PREFIX/src/gnuradio/
wget https://github.com/gnuradio/gnuradio/pull/2821/commits/cfb64aa377ae8b8a3cb5ebdba749d50efc0f325d.diff
```
just the url from the commit, suffixed with ".diff", then apply it
```
git apply cfb64aa377ae8b8a3cb5ebdba749d50efc0f325d.diff
cd build
make install -j4
```

gr-uhd/grc/gen_uhd_usrp_blocks.py
michaelld: yup; gone. Thx!

michaelld:
Let's give @mbr0wn 48 more hours, then we'll merge with or without his approval. This PR is UHD-related, after all ...

ThomasHabets:
@mormj thanks. Can confirm this pull request fixes it for me too.

michaelld:
Enough time has passed, and the reviews are positive. Merging.

devnulling:
This needs to be backported to 3.8.

PR #2824, opened 2019-09-28
DTV: LDPC encoder: refactoring

marcusmueller:
* use `uint16_t` instead of `int` for tables
  * half the storage: should increase cache locality significantly
  * prefer unsigned types for unsigned jobs (here: indexing)
* convert humongous macro to template function
  * type safety
  * found and eliminated unused variables
  * cleaner code
  * better compiler warnings

NB: This class came without ANY tests.


marcusmueller:
@drmpeg could you check whether my assessment that uint16_t must be enough to hold all values is alright?

marcusmueller:
@drmpeg thanks!

marcusmueller:
(nice to hear perf on 32bit ARM increased :) my guess is that on x86-typical cache sizes, the chance of this reducing memory bandwidth bottlenecking is slimmer)

Issue #1466, opened 2017-09-17
python: replace star imports in python modules


noc0lour:
In some submodules (e.g. analog) `__init__.py` has a lot of star imports to submodules which makes tracing the called function hard. Clashing function names in the submodule can also occur without noticing it.

Replacing star imports with named imports in `__init__.py` would make clean which functions are intended to be in the top-level package.

mbr0wn:
We should, in general, aim for a high PyLint score. It would flag star imports. Could we somehow integrate this into BuildBot?

noc0lour:
```
Raw metrics
-----------

+----------+-------+------+---------+-----------+
|type      |number |%     |previous |difference |
+==========+=======+======+=========+===========+
|code      |123787 |44.38 |NC       |NC         |
+----------+-------+------+---------+-----------+
|docstring |63092  |22.62 |NC       |NC         |
+----------+-------+------+---------+-----------+
|comment   |5975   |2.14  |NC       |NC         |
+----------+-------+------+---------+-----------+
|empty     |86058  |30.85 |NC       |NC         |
+----------+-------+------+---------+-----------+



Duplication
-----------

+-------------------------+------+---------+-----------+
|                         |now   |previous |difference |
+=========================+======+=========+===========+
|nb duplicated lines      |11326 |NC       |NC         |
+-------------------------+------+---------+-----------+
|percent duplicated lines |4.064 |NC       |NC         |
+-------------------------+------+---------+-----------+



Messages by category
--------------------

+-----------+-------+---------+-----------+
|type       |number |previous |difference |
+===========+=======+=========+===========+
|convention |24826  |NC       |NC         |
+-----------+-------+---------+-----------+
|refactor   |2262   |NC       |NC         |
+-----------+-------+---------+-----------+
|warning    |12576  |NC       |NC         |
+-----------+-------+---------+-----------+
|error      |45434  |NC       |NC         |
+-----------+-------+---------+-----------+



% errors / warnings by module
-----------------------------

+-------------------------------------------------+------+--------+---------+-----------+
|module                                           |error |warning |refactor |convention |
+=================================================+======+========+=========+===========+
|gnuradio.filter.filter_swig                      |7.98  |1.69    |5.17     |4.25       |
+-------------------------------------------------+------+--------+---------+-----------+
|gnuradio.dtv.dtv_swig                            |7.83  |1.69    |8.58     |4.03       |
+-------------------------------------------------+------+--------+---------+-----------+
[...]

PR #2829, opened 2019-10-01
Remove some star imports to replace them with explicit names

Antetokounpo:
This removes some star imports to make the code cleaner as talked about in: [#1466](https://github.com/gnuradio/gnuradio/issues/1466)


mormj:
This looks to be an improvement in making code cleaner and it also helps with the code parsing tools in vs code.  Thanks for the contribution

The * imports look to be more pervasive through the codebase - any reason channels was singled out?  There are more elsewhere in channels too so it would be nice to get at least all of them in the submodule.

mbr0wn:
Merged manually.

PR #2831, opened 2019-10-01
gr-uhd: fix Python apps

argilo:
I noticed a few bugs in gr-uhd's Python apps:

1. uhd_rx_cfile doesn't work at all in Python 3 because it uses the old `print` syntax in some places.
1. uhd_fft's "Scope" tab displays the channel names as `Re{Channel 0}` and `Im{Channel 0.5}` in Python 3 because of the change to the `/` operator.
1. Massive clipping occurs in uhd_rx_nogui's WFM receiver because it requests an audio gain of 20 (instead of the default of 1).

I've fixed these problems, and also cleaned up the whitespace in uhd_rx_nogui, which previously used a mixture of tabs and spaces.


marcusmueller:
Thanks!

marcusmueller:
Cherry-picked onto master, too.

PR #2832, opened 2019-10-01
Band pass filter validates low/high frequency and transition width

### LINKED ISSUE #2777: GRC38 Band pass filter does not validate parameter
When low/high frequency or transition width is not fill in (left blank), GRC does not detect the missing parameters and generates invalid Python code such as:

      self.band_pass_filter_0 = filter.fir_filter_ccf(
            1,
            firdes.band_pass(
                1,
                samp_rate,
                ,
                ,
                1000,
                firdes.WIN_HAMMING,
                6.76))

*i.e.* several comma separator in a row.

cbdonohue:
Added some minimal parameter validation for band pass filter
fixes #2777

### DIFF
--- a/gr-filter/grc/filter_band_pass_filter.block.yml
+++ b/gr-filter/grc/filter_band_pass_filter.block.yml
@@ -58,8 +58,9 @@ parameters:
     default: '6.76'
 
 asserts:
--   ${ low_cutoff_freq > 0 }
--   ${ high_cutoff_freq >= low_cutoff_freq }
+-   ${ low_cutoff_freq }
+-   ${ high_cutoff_freq }
+-   ${ low_cutoff_freq <= high_cutoff_freq }
 -   ${ width > 0 }
 
 inputs:
--- a/gr-filter/grc/filter_band_pass_filter.block.yml
+++ b/gr-filter/grc/filter_band_pass_filter.block.yml
@@ -39,12 +39,15 @@ parameters:
 -   id: low_cutoff_freq
     label: Low Cutoff Freq
     dtype: real
+    default: '0'
 -   id: high_cutoff_freq
     label: High Cutoff Freq
     dtype: real
+    default: '1'
 -   id: width
     label: Transition Width
     dtype: real
+    default: '1'
 -   id: win
     label: Window
     dtype: int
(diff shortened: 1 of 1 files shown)


drmpeg:
Sorry, but this doesn't work for a complex band pass filter with negative frequencies.

An example of band pass filter with negative frequencies usage is an SSB transmitter. The lower sideband (LSB) mode is generated with a negative frequency band pass filter.

image

cbdonohue:
Thanks for the feedback!
I just pushed another change to my branch that allows for complex band pass filters. cbdonohue/gnuradio@2ba39c14f026ff22362ae905143be1b361e60e3e

primercuervo:
@mormj based on your feedback, would you mind checking if this check is now valid?

drmpeg:
@primercuervo I guess you meant me instead of @mormj. I've tested the patch, and it mostly works. Unfortunately, the two asserts that are trying to make sure something is entered:
```
-   ${ low_cutoff_freq }
-   ${ high_cutoff_freq }
```
also disallow having the value 0, which is valid.

primercuervo:
@drmpeg you are absolutely right that I meant you! I'm sorry for the spam this might have caused. Thanks for checking that.

@cbdonohue mind having a look at the assert of 0?

cbdonohue:
> @drmpeg you are absolutely right that I meant you! I'm sorry for the spam this might have caused. Thanks for checking that.
> 
> @cbdonohue mind having a look at the assert of 0?

I'll check it out

cbdonohue:
The assert statements are eventually sent to the builtin **eval()** method. I think what is happening is that it returns **False** if it is evaluating **0**. 

The solution for now might be to remove:
```
-   ${ low_cutoff_freq }
-   ${ high_cutoff_freq }
```

 add default values to the **low_cutoff_freq** and **high_cutoff_freq**. 
```
-   id: low_cutoff_freq
    label: Low Cutoff Freq
    dtype: real
    default: '0'
-   id: high_cutoff_freq
    label: High Cutoff Freq
    dtype: real
    default: '1'
```

Another possible solution is to modify the template python code. We could check for empty values when populating the lower level filter function parameters.

mbr0wn:
Looks like all the comments have been addressed; merging manually (commit message needed cleanup).

PR #2833, opened 2019-10-03
volk: Replace star imports with explicit imports

jgkamat:
I tested this as much as I know how, but I'm not that familiar with gnuradio so I can't cover all the cases. Unit tests seem to be unaffected for me, but let me know if I'm missing any additional testing I can make.

Part of #1466

### DIFF
--- a/gnuradio-runtime/examples/volk_benchmark/volk_math.py
+++ b/gnuradio-runtime/examples/volk_benchmark/volk_math.py
@@ -5,7 +5,8 @@ from __future__ import unicode_literals
 from gnuradio import gr
 from gnuradio import blocks
 import argparse
-from volk_test_funcs import *
+from volk_test_funcs import (create_connection, new_table, replace_results,
+                             helper, timeit, format_results)
 
 try:
     from gnuradio import blocks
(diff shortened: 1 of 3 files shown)


gnuradio-runtime/examples/volk_benchmark/volk_math.py
```diff
@@ -5,7 +5,8 @@
 from gnuradio import gr
 from gnuradio import blocks
 import argparse
-from volk_test_funcs import *
+from volk_test_funcs import (create_connection, new_table, replace_results,
```
primercuervo: Thank you for tackling this! You seem to have made this file executable. Mind changing that back?

gnuradio-runtime/examples/volk_benchmark/volk_plot.py
```diff
@@ -4,7 +4,8 @@
 from __future__ import unicode_literals
 import sys, math
 import argparse
-from volk_test_funcs import *
+from volk_test_funcs import (create_connection,
```
primercuervo: It seems that the way you are adding the line breaks is different for each of your changes. It would be better if they are consistent along the whole PR. Mind formatting these changes consistently?

gnuradio-runtime/examples/volk_benchmark/volk_math.py
mbr0wn: The parens are also un-Pythonic. But that's not a major issue.

PR #2834, opened 2019-10-03
Save the new duplicate page to the cache

### LINKED ISSUE #2762: New duplicate flowgraph disappears if you escape out of dialog box
Create a flowgraph.  Make a duplicate with File->Duplicate.  On the new one, double click on the "Options Box".  Then hit escape to close the box.  The whole canvas is emptied.

Happens with GR 3.8 and python 2.7

cbdonohue:
If you close a dialog without make a change, it restores the flow graph from the cache, but a new page has no cache so the flow graph disappears.

Here is where it loads from the cache:
https://github.com/gnuradio/gnuradio/blob/55c55fe3e17d556484ed9139e8e941a0da00f9c4/grc/gui/Application.py#L559

This fix saves the new page to the cache so it can be loaded from the cache in the future.
fixes #2762

### DIFF
--- a/grc/gui/Application.py
+++ b/grc/gui/Application.py
@@ -693,6 +693,7 @@ class Application(Gtk.Application):
             # Import the old data and mark the current as not saved
             new_flow_graph.import_data(previous.export_data())
             flow_graph_update(new_flow_graph)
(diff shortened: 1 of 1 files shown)


marcusmueller:
Woohoo! Thanks. And welcome to the group of contributors!

PR #2835, opened 2019-10-03
runtime: add GR_PREFS_PATH env variable search

mormj:
When looking for the config file, allow the env variable
GR_PREFS_PATH to be searched so that multiple installations
of GR each with its own prefs can be used, similar to what
is currently done in GRC

There is much more work to be done in redesigning the prefs module
but just adding a env variable to point the user generated conf 
file to a different location seems like a low hanging fruit to allow better
usage with multiple installations on one machine.  This change can also
be backported to 3.8 perhaps

### DIFF
--- a/gnuradio-runtime/python/gnuradio/gr/qa_prefs.py
+++ b/gnuradio-runtime/python/gnuradio/gr/qa_prefs.py
@@ -1,0 +1,40 @@
+#!/usr/bin/env python
+#
+# Copyright 2019,2020 Free Software Foundation, Inc.
+#
+# This file is part of GNU Radio
+#
+# GNU Radio is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 3, or (at your option)
+# any later version.
+#
+# GNU Radio is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with GNU Radio; see the file COPYING.  If not, write to
+# the Free Software Foundation, Inc., 51 Franklin Street,
+# Boston, MA 02110-1301, USA.
+#
+
+from __future__ import print_function
+
+
+from gnuradio import gr, gr_unittest
+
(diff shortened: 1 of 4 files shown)


marcusmueller:
I really really like this, and wish we have had this for years – it **is** an API change in that it introduces an "environment sensibility" that wasn't there before, but I'd argue that this is OK, since it's not a breaking change for anyone not using a "reserved" environment variable name.

Might be of great interest to windows packagers, too, since env variable setting can easily be done at installation time.

It, however, breaks the `gr_prefs` unit test. What do we do about that.

mormj:
I'll take a look at the unit test and see what is breaking.

gnuradio-runtime/python/gnuradio/gr/qa_prefs.py
```diff
@@ -0,0 +1,47 @@
+#!/usr/bin/env python
+#
+# Copyright 2007,2010,2019 Free Software Foundation, Inc.
```
mbr0wn:
```suggestion
# Copyright 2019 Free Software Foundation, Inc.
```

gnuradio-runtime/python/gnuradio/gr/qa_prefs.py
```diff
@@ -0,0 +1,47 @@
...
+import imp
+
+class test_prefs (gr_unittest.TestCase):
+
+    def test_001(self):
+
+        p = gr.prefs()
+
+        # Read some options
+        self.assertTrue(p.has_option('grc','default_flow_graph'))
```
mbr0wn: PEP8 please
```suggestion
        self.assertTrue(p.has_option('grc', 'default_flow_graph'))
```

mbr0wn:
@marcusmueller I understand your comment as "it breaks API-ish, but that's OK on master, just not on maint-3.8"?

mbr0wn:
I don't think we can assume the existence of default config files during QA.

marcusmueller:
@mbr0wn  it's an "it's an API extension, so even OK for maint-3.8", to be honest. I could have worded that more explicitly.

(the thing is: when you add an exported function, you don't break API, and nobody's going to call your function if they don't know it; now, env var sensibility is something different, because you *can* accidentally break your program by having something set that formerly had no effect, and now has – but honestly, it's prefixed with `GR_PREFS`; I'm positive nobody uses such envvars randomly)

mbr0wn:
It passes!

Issue #2717, opened 2019-08-10
Upstream gr-iio


marcusmueller:
Working together with @tfcollins, we'll make the IIO interface part of mainline GNU Radio.

777arc:
Looking forward to it!

michaelld:
I like the idea, generally. Wondering if ADI be mostly in charge of keeping it up to date / bug fixes / etc? Or, us GR devs? We don't need more work ... so hoping Travis/ADI will take on that burden.

tfcollins:
@michaelld Yes, I will take care of maintenance.

michaelld:
> @michaelld Yes, I will take care of maintenance.

@tfcollins Excellent! I'm sold! Let's do this!

PR #1029, opened 2016-09-22
Add a mechanism to cancel blocking work() function

pcercuei:
Introduce a cancel_work() virtual function to the block class, which can
be overriden by child classes.

This function can be used to interrupt a blocking process in the work()
function.

It will be used in the gr-iio project.

Signed-off-by: Paul Cercueil <paul.cercueil@analog.com>

PR #797, opened 2016-04-21
Termination of run-to-completion flow graphs

bastibl:
This is supposed to address [Issue #797](http://gnuradio.org/redmine/issues/797), which is about shutting down run-to-completion flow graphs as used in simulations and unit tests.

The point is that the current code does not allow to stop blocks through the message port if it has at least on stream port. This policy is implemented in [block.cc](https://github.com/gnuradio/gnuradio/blob/d13de2faffc6e9c50d353e1c3567d688b326b805/gnuradio-runtime/lib/block.cc#L753). I don't know whether this was a deliberate decision, but it's at least unexpected that this is not supported.

I think it would make sense to allow at least blocks that have only message port inputs to be shut down via the message port. (Note that blocks with a stream output are only shut down if they received a *done* message and don't produce anymore).
Apart from making simulation scripts easier to terminate, this also allows for nicer units tests as shown in the last commit.

PR #2840, opened 2019-10-04
Integrate gr-iio

tfcollins:
This PR is related to grep 0017 (https://github.com/gnuradio/greps/blob/master/grep-0017-iio.md) and issue https://github.com/gnuradio/gnuradio/issues/2717. In short it adds the gr-iio OOT module in tree.

Changes include
- Integrated gr-iio in tree
- Make libad9361 dependency optional for gr-iio
- Add new DDS and attribute updater block
- Add documentation into headers
- Disable AD9361 filters when not using them
- Update examples to 3.8

Signed-off-by: Travis F. Collins <travis.collins@analog.com>


carlesfernandez:
Please check https://github.com/analogdevicesinc/gnuradio/pull/2

gr-iio/CMakeLists.txt
```diff
@@ -0,0 +1,80 @@
...
+
+set(GR_IIO_DEPENDENCIES
+  Boost_FOUND
+  ENABLE_GNURADIO_RUNTIME
+  ENABLE_GR_BLOCKS
+  ENABLE_GR_ANALOG
+  ENABLE_VOLK
+  IIO_FOUND)
+
+if (NOT LIBAD9361_FOUND STREQUAL "OFF")
```
michaelld: AD9361 should be an `option`, then do the `if` on the option result. The end result is roughly the same as what this code does, but the `option` code is more robust.

gr-iio/lib/CMakeLists.txt
```diff
@@ -0,0 +1,73 @@
...
+  device_source_impl.cc
+  device_sink_impl.cc
+  attr_source_impl.cc
+  attr_sink_impl.cc
+  attr_updater_impl.cc
+  dds_control.cc
+)
+
+target_link_libraries(gnuradio-iio PUBLIC
+  gnuradio-runtime
```
michaelld: Since this new `gr-iio` submodule depends on GR's `analog` and ` blocks` in its top-level CMakeLists.txt, shouldn't those libraries be included here for linking? I think this is what @carlesfernandez was meaning about https://github.com/analogdevicesinc/gnuradio/pull/2 ...

michaelld:
I didn't check whether `gnuradio::analog` is actually used, but it is listed as a component dependency. If it's not used, then please remove it from the dependency list; if it is used then my comment above holds.

tfcollins:
> 
> 
> I didn't check whether `gnuradio::analog` is actually used, but it is listed as a component dependency. If it's not used, then please remove it from the dependency list; if it is used then my comment above holds.

I believe gr-analog was a dependency on the older "math" blocks that existed in gr-iio. Removing now and rebuilding.

gr-iio/CMakeLists.txt
tfcollins: Fixed

gr-iio/lib/CMakeLists.txt
tfcollins: Integrated the PR

michaelld: thx!

gr-iio/CMakeLists.txt
michaelld: thx!

michaelld:
@tfcollins looks like you've resolved the issues and requested changes. Are you ready for us to -really- start testing this PR?

tfcollins:
Yes!

michaelld:
> Yes!

OK great! Give us a bit to do testing, as this is a significant addition to the GR codebase.

gr-iio/lib/device_source_impl.cc
```diff
@@ -0,0 +1,487 @@
...
+             * this block send itself a dummy message on its input
+             * system message port, so that the scheduler calls us
+             * again promptly. */
+#ifdef GR_VERSION_3_7_OR_LESS
+            pmt::pmt_t payload = pmt::from_long(0);
+#else
+            pmt::pmt_t payload = pmt::from_bool(false);
+#endif
+            pmt::pmt_t msg = pmt::cons(pmt::mp("done"), payload);
+            post(pmt::mp("system"), msg);
```
bastibl: Could you explain a bit about the interaction of the block's thread and the IO thread, and this hack, which seems to force the scheduler into busy-waiting. At the moment the scheduler polls with an interval of 100ms even if nothing changes from the perspective of the scheduler. This can lead to higher delays as mentioned in the comment, given the random scheduling of the OS. I think in the long run, the scheduler shouldn't poll at all. If there is new data available the block should be notified.
In this case, I'd expect the IO thread to notify the scheduler that there is new data available. This can be done through a message to the system port. Since this is thread-safe, this can also happen from the IO thread. Is there a reason why it's done like that?

tfcollins: The threading model here provides backpressure to hardware and also tries to get data into the block buffers as soon as space is made available. As your comment and the code comment suggested this can create throughput slowdowns and latency issues due to the scheduler's update rate. So this hack forces the timeout control to be within the block itself and not from the scheduler.

We could move the message post call (hack) to the refill_thread which would break the polling effect from the work thread. Ideally, this block would completely ignore the scheduler and simply be paced by backpressure and hardware rates.

bastibl: Sorry, but I still don't get it. To me it looks like:
- it blocks in `work` (which I think is bad)
- if it times out after blocking, it just schedules again, i.e., busy-waiting (which I think is bad)
- I don't get the relevance of the "timeout" message. What does it indicate to downstream blocks. I thought is indicates discontinuities (lost samples), but that doesn't seem to be the case. It looks like it just means that the output buffer couldn't be filled in 100ms. Why is this relevant?

Why doesn't the IO thread fill the buffer continuously (and not only on demand) and `work` copies it to the output_buffer? If work copied all samples and returned 0, the IO thread should also send  a message when new samples are available so that the block is scheduled again.

tfcollins: I'm not totally sure why blocking the work thread is bad. Does this stop the scheduler from servicing other blocks? However, I do agree that busy waiting is not good. The sections of code relating to the "fast_enough" variable and the "timeout" message where I believe from earlier versions of libiio which did not support timeouts.

> Why doesn't the IO thread fill the buffer continuously (and not only on demand) and `work` copies it to the output_buffer?
This can only happen when there is space to put data. So the work function has to backpressure the IO thread somehow, which is accomplished by signaling it when to run.

I rewrote the refill and work threads here which should prevent work blocking and remove busy waiting: https://gist.github.com/tfcollins/7c0f6fc389c7e3ae6c21a865f322335d
Does that seem more reasonable? We can also require libiio versions which have timeout features which has been available since 2016.

michaelld: Blocking in "work" happens in some blocks, like the throttle.

For a source block, it's totally OK to block since the block can't generate data anyway & thus there's nothing more that it can do.

For the sink block, blocking indefinitely can be problematic since the scheduler needs to spin every now and then to deal with incoming samples.

Thus: A timed block & followup check is the way to go regardless of source or sink [... note that I haven't looked at your code yet to see if this is what you're doing].

michaelld: _I wish there were a better way to handle the situation "I have no data to produce" in the scheduler, so that it doesn't spin calling "work" & getting nothing done time and again (which can take 100% of a CPU core; hence why it's bad)._

gr-iio/CMakeLists.txt
```diff
@@ -0,0 +1,80 @@
...
+
+set(GR_IIO_DEPENDENCIES
+  Boost_FOUND
+  ENABLE_GNURADIO_RUNTIME
+  ENABLE_GR_BLOCKS
+  ENABLE_VOLK
+  IIO_FOUND)
+
+option(LIBAD9361_ENABLE "Build with libad9361-iio" ON)
+if (LIBAD9361_ENABLE)
```
michaelld: Thinking the default here (when `LIBAD9361_ENABLE` is not set on the commandline or elsewhere, and hence defaults to `ON`) should be to use `libad9361-iio` if it is found & otherwise don't use it. As written, the default if `libad9361-iio` is -not- found will cause this component to be disabled.

tfcollins: Yes, you have to explicitly disable libad9361 for it to enable the module without it installed. To build without libad9361-iio by default would just require a if(LIBAD9361_FOUND) around the append. Is that the better default behavior?

gr-iio/lib/device_source_impl.cc
tfcollins: If blocking work doesn't matter I can remove the refill_thread completely as the libiio buffer API should be enough

gr-iio/CMakeLists.txt
michaelld: hmmm ... not quite. If `-DLIBAD9361_ENABLE=ON` is specified on the cmake commandline, then we always want to add `LIBAD9361_FOUND` to the `GR_IIO_DEPENDENCIES`. OTOH, if `LIBAD9361_ENABLE` is not specified one way or the other on the cmake commandline, then we want to add `LIBAD9361_FOUND` to the `GR_IIO_DEPENDENCIES` only if `LIBAD9361_FOUND` is TRUE.

tfcollins: With that specification I don't think using the cmake OPTION command is useful since it cannot distinquish between user input and a default option.

gr-iio/lib/device_source_impl.cc
bastibl: I still think blocking is bad :-)
- A source block might have filled nearly the whole output buffer and is called again with space for only one sample in the output buffer. Then the block sleeps for a minute until a lot of samples are available but it is only able to write one. (Even though downstream might have consumed all  samples meanwhile). -> suboptimal
- It's not that you have to choose between busy-waiting or blocking. Why not just do it properly and only schedule the block when there are samples available ready to be written to the output  buffer?
- For now, there is one thread per block. But I thought we want to go forward with the runtime and decouple threads from blocks. Then, there will be one thread that handles multiple blocks. In that case, we just cannot block in work. I don't think that we should add more blocks that block in work, otherwise we have to rewrite whole GNU Radio before we can move forward.

marcusmueller: @bastibl up to now, the dogma has been that it's OK to block in work, especially in a source. In fact, all hardware drivers I can think of do this under the hood; there's also not that many alternatives. 

gr-uhd just calls `recv()`; that is a blocking call. How else should IIO do this? There's no way of telling the scheduler "come back in 20 ms" or so.

bastibl: I'm not sure if this is a general comment (how to do things with GNU Radio) or if you think this block is fine.

> @bastibl up to now, the dogma has been that it's OK to block in work

Looks like I didn't get the memo. Blocking in work caused many problems in the past (incoming messages are not recognized, shutdown didn't work, you are tight to the TPB scheduler, .....). We removed `delete_head_blocking` to stop people from doing it.
Please see this threads
https://github.com/gnuradio/gnuradio/pull/1029 
https://github.com/gnuradio/gnuradio/pull/797
https://lists.gnu.org/archive/html/discuss-gnuradio/2013-06/msg00546.html
https://lists.gnu.org/archive/html/discuss-gnuradio/2013-06/msg00552.html

> There's no way of telling the scheduler "come back in 20 ms" or so.

This block uses a dedicated IO thread to refill the buffer. It could just as well schedule the block (done=false message to system port) when data is available... Again, see https://github.com/gnuradio/gnuradio/pull/1029.

Anyway, I give up...

tfcollins:
Squashed commits as per request @devnulling

marcusmueller:
I'd like to very politely ask you to undo the squashing, if possible, because as it is now, I'll have to re-review this whole 73-file PR from scratch, and that's kinda bad.

I think we can avoid a lot of the overhead, though, and get this out the door rather quickly; let's go through what I can still remember from @tfcollins', et al's discussion in `#coredevs`:

* we're currently working on a scheduler which will fail on blocks that block in work. Two ways forward: fix gr-iio as soon as we merge that, or avoiding blocking in work now. What of these options have been chosen?
  * Block now, fix later: have the remnants of the off-work threads been removed? Is it guaranteed that blocking calls *do* at some point finish?
  * don't block now: requires `return 0;` and pinging the `system` port as soon as the `iio_fill_buffer` is ready to be called again. Has that been implemented?

tfcollins:
Unsquashed. Please include me on the non-blocking work discussion. Do any other source blocks implement this?

nickoe:
Hmm, @michaelld , I think you have a requested change that github does not show as resolved. It may be me not being familiar with the github web interface, but are there outstanding changes needed or is it mostly testing?

michaelld:
@nickoe I've resolved the conversation. I don't know the state of this PR ... so I haven't done a formal review ... I know there were some issues in the past ...

michaelld:
@marcusmueller are we reviewing this now for merging?

adamhorden:
Please see:

https://github.com/analogdevicesinc/gnuradio/pull/4

I have updated this change set to work with GNU Radio version >= 3.9 :nerd_face:

mbr0wn:
Superseded by #4264

PR #2841, opened 2019-10-05
Deboostify random

marcusmueller:
C++11 offers all the random variates that we used from boost, but with a far less awkward interface.

In the course of this, a lot of unused includes were found.

Since this not only changes API in gr-channels, but might also have subtle consequences on the random sequences generated, this is **not** backportable to maint-3.8


marcusmueller:
Fully agreeing on the testing necessity. I haven't added any tests in this PR; the existing unit tests still work, but they are in all honesty pretty superficial.

PR #2842, opened 2019-10-07
modtool: fix non-empty argument list on cli

### LINKED ISSUE #2836: gr_modtool add `--argument-list' broken
The `--argument-list` parameter of `gr_modtool add` does not take effect, which I think is caused by 
https://github.com/gnuradio/gnuradio/blob/55c55fe3e17d556484ed9139e8e941a0da00f9c4/gr-utils/python/modtool/cli/add.py#L129-L130

This will always ask for prompt either set `argument-list` or not.

May be it should be
```
if not self.info['arglist']:
```
since the default value of `self.info['arglist']` is `""` instead of `None`:
https://github.com/gnuradio/gnuradio/blob/55c55fe3e17d556484ed9139e8e941a0da00f9c4/gr-utils/python/modtool/core/add.py#L54

mormj:
When the --argument list was used on gr_modtool cli, the logic for
determining whether it was there or not was off as the empty list was
returned as an empty string

includes unmerged #2826

Fixes #2836

### DIFF
--- a/gr-utils/python/modtool/cli/add.py
+++ b/gr-utils/python/modtool/cli/add.py
@@ -124,10 +124,17 @@ def get_copyrightholder(self):
         click.secho("For GNU Radio components the FSF is added as copyright holder",
                     fg='cyan')
 
+
 def get_arglist(self):
     """ Get the argument list of the block to be added """
-    if self.info['arglist'] is not None:
-        self.info['arglist'] = click.prompt(click.style('Enter valid argument list, including default arguments: \n', fg='cyan'), prompt_suffix='')
+    if self.info['arglist'] in [None,'']: 
+        self.info['arglist'] = click.prompt(click.style(
+            'Enter valid argument list, including default arguments: \n',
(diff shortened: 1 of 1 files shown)


gr-utils/python/modtool/cli/add.py
```diff
@@ -124,10 +124,17 @@ def get_copyrightholder(self):
         click.secho("For GNU Radio components the FSF is added as copyright holder",
                     fg='cyan')
 
+
 def get_arglist(self):
     """ Get the argument list of the block to be added """
-    if self.info['arglist'] is not None:
-        self.info['arglist'] = click.prompt(click.style('Enter valid argument list, including default arguments: \n', fg='cyan'), prompt_suffix='')
+    if self.info['arglist'] in [None,'']:
```
swap-nil7: `if not self.info['arglist'] or self.info['arglist'].isspace()` will take care of spaces as well.

mormj:
Closing in favor of #2846

PR #2844, opened 2019-10-08
grc: update namespace properly for nested variables

### LINKED ISSUE #2634: grc: address why rewrite needs to be called twice for busports load correctly
grc/core/FlowGraph.py:
```
        self.rewrite()  # TODO: Figure out why this has to be called twice to populate bus ports correctly
```

This was introduced with PR #2589

mormj:
Was getting lots of evaluation error prints when using gr-digital
examples because the nested variables were not updated until rewrite()
and evaluate were called over and over

Seems to fix the issue to update the flowgraph namespace variable after
each variable block is evaluated

I believe this also removes a previous issue (#2634) where rewrite
had to be called twice when data was being imported

Here is a situation that created the ugly printout:

image

image

would print out
```
[ERROR] Failed to evaluate variable block rxmod (FlowGraph.py:233)
  File "/share/gnuradio/grdev/src/gnuradio/grc/core/FlowGraph.py", line 227, in renew_namespace
AttributeError: 'NoneType' object has no attribute 'to_basic_block'
ERROR:grc.core.FlowGraph:Failed to evaluate variable block rxmod
  File "/share/gnuradio/grdev/src/gnuradio/grc/core/FlowGraph.py", line 227, in renew_namespace
AttributeError: 'NoneType' object has no attribute 'to_basic_block'
>>> Done
```

Because the variables needed for the parameter evaluation did not exist in the namespace at the time they were being evaluated.

This also created the situation that the ber_curve_gen flowgraphs would not load up properly

image

with the error:
```
[ERROR] Failed to evaluate variable block enc_ldpc (FlowGraph.py:233)
  File "/share/gnuradio/grdev/src/gnuradio/grc/core/FlowGraph.py", line 227, in renew_namespace
  File "/share/gnuradio/grdev/lib/python3.6/dist-packages/gnuradio/fec/fec_swig.py", line 1317, in ldpc_encoder_make
    return _fec_swig.ldpc_encoder_make(alist_file)
RuntimeError: Bad AList file name!
```

where the workaround was to call rewrite() twice in the import data.  This is no longer necessary with the self.namespace updated in between variable block loads

Fixes #2634

### DIFF
--- a/grc/core/FlowGraph.py
+++ b/grc/core/FlowGraph.py
@@ -226,6 +226,7 @@ class FlowGraph(Element):
                 variable_block.rewrite()
                 value = eval(variable_block.value, namespace, variable_block.namespace)
                 namespace[variable_block.name] = value
+                self.namespace.update(namespace) # rewrite on subsequent blocks depends on an updated self.namespace 
             except TypeError: #Type Errors may happen, but that desn't matter as they are displayed in the gui
                 pass
             except Exception:
@@ -381,8 +382,7 @@ class FlowGraph(Element):
 
             block.import_data(**block_data)
 
-        self.rewrite()  # TODO: Figure out why this has to be called twice to populate bus ports correctly
-        self.rewrite()  # evaluate stuff like nports before adding connections
+        self.rewrite() 
 
         # build the connections
         def verify_and_get_port(key, block, dir):


primercuervo:
Hi @mormj 

Thank you for the analysis. I was able to reproduce the first case by only placing the `modulate vector` in a fresh flowgraph, without needing to modify its internals, but somehow didn't see your second use-case (with the BER curve gen). Mind elaborating a bit more on a way to reproduce it? I tested it in v3.8 as well.

Also, now that the self.namespace is updated in every iteration, would that make the lines https://github.com/gnuradio/gnuradio/blob/bd56b7a0635298821a9caecb86959a6380d335bc/grc/core/FlowGraph.py#L236-L238 superfluous?

PR #2845, opened 2019-10-08
Change rounding method to `std::lround`

### LINKED ISSUE #2822: use right rounding method for floating point rounding
```
/home/marcus/src/gnuradio/gnuradio-runtime/lib/block_executor.cc:146:38: warning: casting (double + 0.5) to integer leads to incorrect rounding; consider using lround (#include <cmath>) instead [bugprone-incorrect-roundings]
                    new_tag.offset = ((double)new_tag.offset * rrate) + 0.5;

```

for large offsets, +0.5 can silently round to 0 ('doh).

might be more than related to #2055

mathiasvr:
Fixes #2822

### DIFF
--- a/gnuradio-runtime/lib/block_executor.cc
+++ b/gnuradio-runtime/lib/block_executor.cc
@@ -143,7 +143,7 @@ static bool propagate_tags(block::tag_propagation_policy_t policy,
             } else if (use_fp_rrate) {
                 for (t = rtags.begin(); t != rtags.end(); t++) {
                     tag_t new_tag = *t;
-                    new_tag.offset = ((double)new_tag.offset * rrate) + 0.5;
+                    new_tag.offset = std::llround((double)new_tag.offset * rrate);
                     for (int o = 0; o < d->noutputs(); o++)
                         out_buf[o]->add_item_tag(new_tag);
                 }
@@ -191,7 +191,7 @@ static bool propagate_tags(block::tag_propagation_policy_t policy,
                 } else if (use_fp_rrate) {
                     for (t = rtags.begin(); t != rtags.end(); t++) {
                         tag_t new_tag = *t;
-                        new_tag.offset = ((double)new_tag.offset * rrate) + 0.5;
+                        new_tag.offset = std::llround((double)new_tag.offset * rrate);
                         out_buf->add_item_tag(new_tag);
                     }
                 } else {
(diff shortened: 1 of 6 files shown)


mathiasvr:
@michaelld Thanks, from the issue it wasn't even clear to me that `0.5` was part of the rounding, so your comment cleared things up for me. I have updated PR accordingly.

I believe using the `lround` function instead of the `+0.5` trick is preferable since it reads more clearly the intent. Also, it is supposedly faster and more correct in some cases, see https://clang.llvm.org/extra/clang-tidy/checks/bugprone-incorrect-roundings.html

michaelld:
> @michaelld Thanks, from the issue it wasn't even clear to me that `0.5` was part of the rounding, so your comment cleared things up for me. I have updated PR accordingly.

OK. Thanks for that change. I can agree now that the code is roughly equivalent to what it was doing before.

> I believe using the `lround` function instead of the `+0.5` trick is preferable since it reads more clearly the intent. Also, it is supposedly faster and more correct in some cases, see https://clang.llvm.org/extra/clang-tidy/checks/bugprone-incorrect-roundings.html

Interesting. Makes sense, sort of. Makes me sad that my "old school" go to for float to integer conversion isn't guaranteed to be as fast or accurate as I thought it was! :(

michaelld:
I've tagged @marcusmueller for a 2nd review, just for the sake of 1 more set of eyes here.

ValZapod:
Mmm... What is the point of changing only one place?

willcode:
Using llround() for clarity is good, and this looks correct here. I don't think it improves accuracy or performance in this case (rrate/offset are positive doubles). Also in tagged_stream_block.cc.

michaelld:
I think with the change to line 195, I'm happy with this PR. I'd wish for a review from @marcusmueller  ... but if he's too busy I'm OK with moving this along as it's simple and makes sense.

mathiasvr:
I went ahead and fixed every line reported as `bugprone-incorrect-roundings` by `clang-tidy`.

michaelld:
@mathiasv do you have a CLA (commit agreement) with the GNU Project? If you don't or you don't know, then you don't & we need to get you one for this many changes.

mathiasvr:
@michaelld I see, how do I get one?

michaelld:
@bhilburn would you please get @mathiasvr started with the CLA process?

bhilburn:
@mathiasvr - Just send me an email, `bhilburn@gnuradio.org` =)

mathiasvr:
I signed the CLA a while ago, am I supposed to do something or did you already receive it?

bhilburn:
Hi Mathias - Nope, you are all set! All your contributions are clear for
merge.

michaelld:
All CI passes. Multiple positive reviews. CLA. -> merging

PR #2846, opened 2019-10-09
gr-modtool: Update the validation in Modtool Add

swap-nil7:
The validation of the arguments list and copyright holder should be updated to take care of the potential whitespaces. 
The check on the arguments list is currently incorrect.

### DIFF
--- a/gr-utils/python/modtool/cli/add.py
+++ b/gr-utils/python/modtool/cli/add.py
@@ -101,15 +101,15 @@ def get_lang(self):
 
 def get_blockname(self):
     """ Get the blockname"""
-    if self.info['blockname'] is None:
+    if not self.info['blockname'] or self.info['blockname'].isspace():
         while not self.info['blockname'] or self.info['blockname'].isspace():
             self.info['blockname'] = cli_input("Enter name of block/code (without module name prefix): ")
-    if not re.match('[a-zA-Z0-9_]+', self.info['blockname']):
+    if not re.match('^[a-zA-Z0-9_]+$', self.info['blockname']):
         raise ModToolException('Invalid block name.')
 
 def get_copyrightholder(self):
     """ Get the copyrightholder of the block to be added """
-    if self.info['copyrightholder'] is None:
+    if not self.info['copyrightholder'] or self.info['copyrightholder'].isspace():
         user = getpass.getuser()
         git_user = self.scm.get_gituser()
         if git_user:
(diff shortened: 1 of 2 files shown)


gr-utils/python/modtool/cli/add.py
```diff
@@ -101,15 +101,15 @@ def get_lang(self):
 
 def get_blockname(self):
     """ Get the blockname"""
-    if self.info['blockname'] is None:
+    if not self.info['blockname'] or self.info['blockname'].isspace():
         while not self.info['blockname'] or self.info['blockname'].isspace():
             self.info['blockname'] = cli_input("Enter name of block/code (without module name prefix): ")
     if not re.match('[a-zA-Z0-9_]+', self.info['blockname']):
```
mormj: Strangely, I'm allowed to create a block with a space in the name (separate issue than this PR), but probably could be prevented here

swap-nil7: @mormj: Yes, that's right. What do you think would be better, replacing spaces with an underscore or disallowing the users to use a space in the `blockname`?

mormj: Personally I think you could go either way - it's a very fringe case that someone would have to intentionally do with quotes.  Just disallowing is probably the safest.

primercuervo: I think avoiding spaces is the way to go. I really don't see a real use-case where I would like to manage spaces, needless to say that they are handled different for different OS. I suggest just disallowing spaces

gr-utils/python/modtool/cli/add.py
```diff
@@ -101,15 +101,15 @@ def get_lang(self):
...
         while not self.info['blockname'] or self.info['blockname'].isspace():
             self.info['blockname'] = cli_input("Enter name of block/code (without module name prefix): ")
-    if not re.match('[a-zA-Z0-9_]+', self.info['blockname']):
+    if not re.match('^[a-zA-Z0-9_]*$', self.info['blockname']):
         raise ModToolException('Invalid block name.')
 
 def get_copyrightholder(self):
     """ Get the copyrightholder of the block to be added """
-    if self.info['copyrightholder'] is None:
+    if not self.info['copyrightholder'] or self.info['copyrightholder'].isspace():
```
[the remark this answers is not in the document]
swap-nil7: Yes, currently if the copyrightholder is empty, then by default gr-blockname owner. We can make that mandatory if that is good to have. :)

Issue #2748, opened 2019-08-19
Rotator is broken in 3.8


argilo:
While porting one of my out-of-tree modules to GNU Radio 3.8, I noticed that the Rotator block is broken. The problem can be reproduced with the following flowgraph:

![Screenshot from 2019-08-19 08-45-39](https://user-images.githubusercontent.com/583749/63266470-fdf5c400-c25d-11e9-92db-054cd7a6898f.png)

This produces the following output:

![Screenshot from 2019-08-19 08-46-05](https://user-images.githubusercontent.com/583749/63266486-03530e80-c25e-11e9-8c25-442235bc6f82.png)

I haven't yet had time to investigate what's going on here.

drmpeg:
The rotator block works fine on ARM, so most likely it's a VOLK issue.

I haven't tested on X86 yet.

argilo:
That seems likely. I changed `#if 0` to `#if 1` here to prevent VOLK from being used, and the rotator went back to working correctly:

https://github.com/gnuradio/gnuradio/blob/39c922e8b28eb67515c134712a4bfb0bed3cd1f7/gr-blocks/lib/rotator_cc_impl.cc#L55-L70

argilo:
I suspect https://github.com/gnuradio/volk/commit/05b92b79c17b80a08c564a73f4386a573f60a610 introduced the bug; when I comment out the rotator FMA protokernels from that commit, the rotator starts working properly again.

argilo:
I tracked down the bug. A fix is here: https://github.com/gnuradio/volk/pull/278

I'm curious why this wasn't caught by tests, since qa_rotator.cc appears to test `rotateN`.

argilo:
When I run `build/gr-blocks/lib/blocks_qa_rotator.cc` on its own, it fails:
```
Running 2 test cases...
/home/argilo/prefix_next/src/gnuradio38/gr-blocks/lib/qa_rotator.cc(95): error: in "t2": check std::abs(expected - actual) <= 0.0001 has failed
/home/argilo/prefix_next/src/gnuradio38/gr-blocks/lib/qa_rotator.cc(95): error: in "t2": check std::abs(expected - actual) <= 0.0001 has failed
/home/argilo/prefix_next/src/gnuradio38/gr-blocks/lib/qa_rotator.cc(95): error: in "t2": check std::abs(expected - actual) <= 0.0001 has failed
...
[...]

PR #2853, opened 2019-10-19
Pull in volk rotator fixes

argilo:

https://github.com/gnuradio/volk/pull/278 was recently merged into volk to fix the broken AVX FMA rotator kernel. This PR pulls the fix into GNU Radio so the rotator block will work properly on recent Intel CPUs.


michaelld:
@argilo I can't figure out which GIT commit this corresponds to ... can you confirm that it is https://github.com/gnuradio/volk/commit/f04a46f18308771b5ebf46292bf0fe6918578b35 ?

michaelld:
Also: does this need to be applied to `maint-3.7` and `master` too?

argilo:
@michaelld Yes, this updates volk to https://github.com/gnuradio/volk/commit/f04a46f18308771b5ebf46292bf0fe6918578b35. The full list of differences is here: https://github.com/gnuradio/volk/compare/1299d72c396a88fd2679adfd7a919ac00d2cf678...f04a46f18308771b5ebf46292bf0fe6918578b35

This change should also be applied to the `master` branch, but not `maint-3.7` as it uses a version of volk that predates the broken rotator kernel.

michaelld:
@argilo OK thx & that makes sense to me.

michaelld:
Tagging @marcusmueller for another review, just for confirmation.

mbr0wn:
Merged manually (fixed commit message).

PR #2858, opened 2019-10-24
codec2: development (master) branch support

ra1nb0w:
add support for the development branch (master) of codec2 but still
maintaining compatibility with version 0.8.1.

The unique way that I know to enable or disable `extern "C"` is to
check codec2 version but the version bump was not synchronized with
the implementation therefore you should always use the latest commit.

This fix should also be merged into `maint-3.8`.

### DIFF
--- a/gr-vocoder/include/gnuradio/vocoder/freedv_api.h
+++ b/gr-vocoder/include/gnuradio/vocoder/freedv_api.h
@@ -25,11 +25,18 @@
 
 #include <gnuradio/vocoder/api.h>
 
+// version >=0.9.1 contains fixes that doesn't require "extern C"
+// between 0.8.1 and 0.9.1 the build fail
+#include <codec2/version.h>
+#if CODEC2_VERSION_MAJOR == 0 && CODEC2_VERSION_MINOR < 9
 extern "C" {
+#endif
 #include <codec2/codec2.h>
 #include <codec2/freedv_api.h>
 #include <codec2/modem_stats.h>
+#if CODEC2_VERSION_MAJOR == 0 && CODEC2_VERSION_MINOR < 9
 }
+#endif
 
 namespace gr {
 namespace vocoder {


ra1nb0w:
> Reviewing the codec2 code ( https://github.com/drowe67/codec2 ), I think what you're getting at is that the `extern "C"` was added to codec2 headers between version 0.8 and 0.9 ... hence we don't need to explicitly do the same here in this GR code. Yes?

yes

> If so, my question is whether the current code fails because of the change in codec2's use of `extern` ... in other words: does doing an `extern "C"` inside something that is already `extern "C"` produce an error during compiling?

yes

Probably I can ask codec2 main developers to bump minor version (like 0.9.1) to avoid discovery problem. Anyway, gnuradio 3.8 fail to build if you use version between 0.8.1 and the latest commit.

michaelld:
OK; thanks for the info. I was trying GR MacPorts GR-next updating to the latest -after- installing codec2 & it failed ... might be for this reason. I'll try this patch & see if it works.

michaelld:
Yes, this patch fixes the issue with the codec2 update.

ra1nb0w:
I am waiting upstream if they bump version.

ra1nb0w:
Upstream updated version to 0.9.1 thefore I updated the comment.
They should release an official version soon.

marcusmueller:
hey, @ra1nb0w, awesome to have you here!

Since this is a short addition, we don't need a CLA for this, but @bhilburn can get in touch with you if you want to contribute more to GNU Radio – for changes > 10 lines, we need a CLA.

michaelld:
@ra1nb0w Please do followup with BenH about the CLA (code license assignment: roughly: you're agreeing that any code changes you submit for GR are copyright the owner and not you). I expect you'll be contributing to GR in the future well beyond this small tweak!

bhilburn:

ra1nb0w:
thank you!

mormj:
This appears to break builds on Ubuntu 18 with codec2 0.7 installed (distro default).  Are we requiring codec >= 0.8 at this point?

```
In file included from /share/gnuradio/grdev/src/gnuradio/gr-vocoder/lib/../include/gnuradio/vocoder/freedv_rx_ss.h:28:0,
                 from /share/gnuradio/grdev/src/gnuradio/gr-vocoder/lib/freedv_rx_ss_impl.h:26,
                 from /share/gnuradio/grdev/src/gnuradio/gr-vocoder/lib/freedv_rx_ss_impl.cc:27:
/share/gnuradio/grdev/src/gnuradio/gr-vocoder/lib/../include/gnuradio/vocoder/freedv_api.h:30:10: fatal error: codec2/version.h: No such file or directory
 #include <codec2/version.h>
          ^~~~~~~~~~~~~~~~~~
```

ra1nb0w:
@mormj can you test the patch https://github.com/gnuradio/gnuradio/pull/2886

mormj:
tested - that fixes it.  Thanks!

PR #2865, opened 2019-10-29
cmake: Remove find package for pythonlibs in gr-qtgui

btashton:
There was an unneeded find_package(PythonLibs 2) in the gr-qtgui
CMakeLists.txt that was occasionally causing issues building the
demo when python3 was targeted.  We should just be using the
GrPython module anyway that was already included.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -357,7 +357,6 @@ include(GrBoost)
 # Enable python component
 ########################################################################
 include(GrPython)
-find_package(PythonLibs ${GR_PYTHON_MIN_VERSION})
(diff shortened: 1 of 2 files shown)


michaelld:
Request: There's another one of these here https://github.com/gnuradio/gnuradio/blob/master/CMakeLists.txt#L360 ... which I don't think is required. Can you test / verify & if redundant then add it's removal to this PR as well? Thx!

michaelld:
I've tagged @marcusmueller to see what he thinks.

btashton:
@michaelld sorry about that, I had miss-clicked that re-review link and there was no way to undo it.

michaelld:
> @michaelld sorry about that, I had miss-clicked that re-review link and there was no way to undo it.

NP!

PR #2866, opened 2019-10-29
gr-zeromq: Fix warnings depricated send since 4.3.1 extends PR #2503

btashton:
Wrap some missed send calls with helper define and make sure
zmq_common_impl.h is being used everywhere instead of zmq.hpp

Most were already pulling this in via the base_impl.h


marcusmueller:
yes! Manually merging this!

Issue #2782, opened 2019-09-03
Flowgraph generation with python 3 still imports python2 libraries


epeters13:
Generating a flow graph that uses "Variable Config" and "XMLRPC Server" in GNU radio 3.8.0.0 (Python 3.7.4) still loads ConfigParser and SimpleXMLRPCServer, which is replaced by configparser and xmlrpc.server in python 3.

output:
```
Generating: '/home/epeters/src/adfags-modem/src/GRC/zmq_file_single.py'

Generating: '/home/epeters/src/adfags-modem/src/GRC/zmq_file_single.py'

Executing: /usr/bin/python3 -u /home/epeters/src/adfags-modem/src/GRC/zmq_file_single.py

Traceback (most recent call last):
  File "/home/epeters/src/adfags-modem/src/GRC/zmq_file_single.py", line 21, in <module>
    import ConfigParser
ModuleNotFoundError: No module named 'ConfigParser'

>>> Done (return code 1)
```

after modifying the "import ConfigParser" to "import configparser as ConfigParser" in the output file SimpleXMLRPCServer also throws errors:

```
Traceback (most recent call last):
  File "radioController.py", line 168, in <module>
    from zmq_file_single import zmq_file_single as GRC
  File "/home/epeters/src/adfags-modem/src/GRC/zmq_file_single.py", line 22, in <module>
    import SimpleXMLRPCServer
ModuleNotFoundError: No module named 'SimpleXMLRPCServer'
```

mugpahug:
In addition I found the following issues:
`def __init__(self, cfg=default.ini):` has to be ` def __init__(self, cfg='default.ini'):`

all `config.read('cfg')` have to be `config.read(cfg)`

Running the following command line against the python script generated by GNU radio mitigates XMLRPC and ConfigParser errors, as well as a the above mentioned issues with variable naming

`sed -i -e "s/'cfg'/cfg/g" -e "s/import ConfigParser/import configparser as ConfigParser/g" -e "s/import SimpleXMLRPCServer/import xmlrpc.server as SimpleXMLRPCServer/g" -E -e "s/cfg=(.*)\)/cfg='\1'\)/g" top_block.py
`

I tested this successfully on my python files generated using GNU radio 3.8.0.0

bidimensional:
[...]

PR #2867, opened 2019-10-29
grc: fix importing configparser in variable config block

karel:
This porblem has been mentioned in #2782. `ConfigParser` was replaced by `configparser` in Python 3 and therefore "Variable Config" block doesn't work. Renaming the import fixes it.


marcusmueller:
Ok, this is the right fix for master, as that's Py3 only. What do we do about 3.8.x.x, where it can be either Py2 _or_ Py3?

PR #2869, opened 2019-10-30
runtime: Purge snprintf

btashton:
We are writting cpp for these operations we should be using std:string
and boost:format.  This also cleans up some compiler warnings about
possible string truncation.

### DIFF
--- a/gnuradio-runtime/lib/local_sighandler.cc
+++ b/gnuradio-runtime/lib/local_sighandler.cc
@@ -28,6 +28,7 @@
 #include <stdio.h>
 #include <string.h>
 #include <stdexcept>
+#include <boost/format.hpp>
 
 namespace gr {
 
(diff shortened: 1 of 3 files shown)


gnuradio-runtime/lib/local_sighandler.cc
```diff
@@ -170,11 +171,11 @@ std::string signal::name() const
 #if defined(HAVE_SNPRINTF)
```
marcusmueller: shouldn't we then also be getting rid of these `#ifdefs`?

marcusmueller:
@michaelld proposal: for "getting the PR-queue moving again" reasons, I'll merge this with the `#ifdef HAVE_SNPRINTF` removed.  Then, we get rid of the evils; we don't even have a HAVE_SNPRINTF check afaik anymore, btw.

marcusmueller:
manually merged; #2875 opened to track the "remove all evil"

michaelld:
:+1:

PR #2870, opened 2019-10-30
uhd: Bump minimum version to 3.9.7

mbr0wn:


marcusmueller:
@michaelld Exactly that; I think the idea is `min(current UHD LTS; UHD version in current Ubuntu LTS)`.

michaelld:
@marcusmueller :+1: makes sense & works for me; thx for replying!

PR #2890, opened 2019-11-07
Modernize by using range loop, and also constify the name

ThomasHabets:


ThomasHabets:
whoa, no this ain't right. Sorry for the noise.

PR #2897, opened 2019-11-11
runtime: fix gr_unittest floatAlmostEqual

### LINKED ISSUE #2883: GNURadio unit test package apparently has a bug in assertFloatTuplesAlmostEqual
GNURadio version 3.8.0.0, Python 3.6.8

The unit test function assertFloatTuplesAlmostEqual is only checking the first element in the two tuples being compared. If other elements are unequal, the assert still passes. Example below; it does three tests using the Python unittest routines. Then it does a couple of tests using the gr_unittest routines.  All the tests pass, even though (in the gr_unittest ones), elements beyond the first elements of each tuple are significantly unequal.
```
import unittest
from gnuradio import gr, gr_unittest

class TestStringMethods(unittest.TestCase):

    def test_upper(self):
        self.assertEqual('foo'.upper(), 'FOO')



    def test_isupper(self):
        self.assertTrue('FOO'.isupper())
        self.assertFalse('Foo'.isupper())

    def test_split(self):
        s = 'hello world'
        self.assertEqual(s.split(), ['hello', 'world'])
        # check that s.split fails when the separator is not a string
        with self.assertRaises(TypeError):
            s.split(2)

class TestFloatMethods(gr_unittest.TestCase):

#    Above we see that the Python unit test package works,
# now let's try the gnuradio QA pack

# THE 2 TESTS BELOW PASS, EVEN THOUGH THE TUPLES ARE GROSSLY UNEQUAL

    def test_float(self):
        self.assertFloatTuplesAlmostEqual([1.0,2.0,4.0],[1.0,300.0,7],5)

    def test_float2(self):
        a = (2.0, 3.0)
        b = (2.0, 4.0) # definitely not equal to tuple above
        self.assertFloatTuplesAlmostEqual(a,b,6)

if __name__ == '__main__':
    unittest.main()
```
---------------------------------------------
Results below:
```
odroid@odroid:~/projects/grc_module/gr-howto/python$ python3 testTheTest.py
Ran 5 tests in 0.001s

OK
odroid@odroid:~/projects/grc_module/gr-howto/python$ 
```

primercuervo:
The call inside this assert was using a generator inside the standard all(),
which was always returning `False`, and silently passing the test, instead of
actually asserting in fail. Changing to list comprehension for assertion raising

Fixes #2883

### DIFF
--- a/gnuradio-runtime/python/gnuradio/gr_unittest.py
+++ b/gnuradio-runtime/python/gnuradio/gr_unittest.py
@@ -82,10 +82,10 @@ class TestCase(unittest.TestCase):
         places.0
         """
         self.assertEqual(len(a), len(b))
-        return all((
+        return all([
             self.assertComplexAlmostEqual(x, y, places, msg)
             for (x, y) in zip(a, b)
-        ))
+        ])
 
 
     def assertComplexTuplesAlmostEqual2(self, a, b,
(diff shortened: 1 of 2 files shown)


mbr0wn:
Thanks for backporting!

PR #2906, opened 2019-11-14
Modtool: Update - Add option to convert blacklisted files

### LINKED ISSUE #2905: `gr_modtool update` fails silently, deleting xml without warning nor output
I know this is an experimental feature, but adding here for tracking:

when updating an OOT module to GNU Radio 3.8 following https://wiki.gnuradio.org/index.php/GNU_Radio_3.8_OOT_Module_Porting_Guide, the command seems to run correctly showing a "satisfactory" output as
```
Converted module_block.xml to module_block.block.yml
Updated the CMakeLists.txt
```
even when the XML isn't converted because the block has a **blacklisted name** as per the Converter limitations (see: https://github.com/gnuradio/gnuradio/blob/master/grc/converter/main.py#L32)

The most cumbersome is, however, that it **deletes** the source .xml.

primercuervo:
Also avoid the misleading behavior that implies conversion
when it isn't happening while also deleting the xml with
a blacklisted name

Fixes #2905

### DIFF
--- a/grc/converter/main.py
+++ b/grc/converter/main.py
@@ -94,10 +94,13 @@ class Converter(object):
             with open(self.cache_file, 'w', encoding='utf-8') as cache_file:
                 json.dump(self.cache, cache_file)
 
-    def load_block_xml(self, xml_file):
+    def load_block_xml(self, xml_file, force=False):
         """Load block description from xml file"""
-        if any(part in xml_file for part in excludes):
-            return
+        if any(part in xml_file for part in excludes) and not force:
+            logger.warn('Skipping {} because name is blacklisted!'.format(xml_file))
+            return False
+        elif any(part in xml_file for part in excludes) and force:
+            logger.warn('Forcing conversion of blackisted file: {}'.format(xml_file))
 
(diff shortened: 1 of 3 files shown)


grc/converter/main.py
```diff
@@ -94,10 +94,13 @@ def run(self, force=False):
...
-    def load_block_xml(self, xml_file):
+    def load_block_xml(self, xml_file, force=False):
         """Load block description from xml file"""
-        if any(part in xml_file for part in excludes):
-            return
+        if any(part in xml_file for part in excludes) and not force:
+            logger.warn('Skipping {} because name is blacklisted!'.format(xml_file))
+            return False
+        elif any(part in xml_file for part in excludes) and force:
+            logger.warn('Forcing conversion of blackisted file: {}'.format(xml_file))
```
mbr0wn:
```suggestion
            logger.warn('Forcing conversion of blacklisted file: {}'.format(xml_file))
```

mbr0wn: Fixed this myself.

mbr0wn:
Merged manually.

PR #2916, opened 2019-11-20
Replace many `const` variables with `constexpr`

ThomasHabets:
constexpr is like const but (for variables) guarantees evaluation at
compile time (as opposed to runtime).

Likely this change will do nothing on its own (though it could, since
it gives the compiler more information). But it still has benefits.

It allows programmer to know that initialization is not expensive (it
was done at compile time), and reduces risk of a refactoring
regressing the compiletimeness.

Runtime initialization can be nonobvious in larger codebases. E.g.:
```c++
  struct S {
    static int foo();
  };
  const int bar = S::foo();  // Called and initialized at *runtime*.
  int S::foo() { return 10; }
```
With constexpr:
```c++
  struct S {
    static constexpr int foo();
  };
  constexpr int bar = S::foo();   // Error: used before definition.
  constexpr int S::foo() { return 10; }
```
Initializing at runtime is not just startup costs, but also can save
memory since it'll end up in a R/O section of a binary and therefore
doesn't need to be swapped out, but can be shared (in the mmap() sense
of the word).


gnuradio-runtime/include/gnuradio/xoroshiro128p.h
```diff
@@ -66,7 +66,7 @@ static inline uint64_t xoroshiro128p_next(uint64_t* state)
   parallel computations. */
 static inline void xoroshiro128p_jump(uint64_t* state)
 {
-    static const uint64_t JUMP[] = { 0xbeac0467eba5facb, 0xd86b048b86aa9922 };
+    static constexpr uint64_t JUMP[] = { 0xbeac0467eba5facb, 0xd86b048b86aa9922 };
```
marcusmueller: No, that won't fly: this is C, not C++!

ThomasHabets:
Reverted that change. Should be good now.

PR #2917, opened 2019-11-21
gr-blocks: VOLKize add_const_ff implementation

### LINKED ISSUE #2825: add_const_ff: not VOLKized
I was surprised to find that instead of an appropriate VOLK this is what `add_const_ff` does:

```
    while (size >= 8) {
        *optr++ = *iptr++ + d_k;
        *optr++ = *iptr++ + d_k;
        *optr++ = *iptr++ + d_k;
        *optr++ = *iptr++ + d_k;
        *optr++ = *iptr++ + d_k;
        *optr++ = *iptr++ + d_k;
        *optr++ = *iptr++ + d_k;
        *optr++ = *iptr++ + d_k;
        size -= 8;
    }

    while (size-- > 0) {
        *optr++ = *iptr++ + d_k;
    }
```

I mean, if we don't have a VOLK kernel for this yet, the second while loop does certainly constitute a good generic kernel. I have my doubts that this manual 8-vectorization actually helps any modern compiler be smart about auto-SIMDization; it's actually more likely to be harmful.

grant-cox:
closes #2825

### DIFF
--- a/gr-blocks/include/gnuradio/blocks/add_const_cc.h
+++ b/gr-blocks/include/gnuradio/blocks/add_const_cc.h
@@ -25,6 +25,7 @@
 
 #include <gnuradio/blocks/api.h>
 #include <gnuradio/sync_block.h>
+#include <volk/volk_alloc.hh>
 
 namespace gr {
 namespace blocks {
--- a/gr-blocks/lib/add_const_cc_impl.cc
+++ b/gr-blocks/lib/add_const_cc_impl.cc
@@ -57,10 +57,10 @@ int add_const_cc_impl::work(int noutput_items,
     const gr_complex* iptr = (const gr_complex*)input_items[0];
     gr_complex* optr = (gr_complex*)output_items[0];
 
-    unsigned long input_size = static_cast<unsigned long> (noutput_items);
+    unsigned long input_size = static_cast<unsigned long>(noutput_items);
 
     gr::thread::scoped_lock guard(d_setlock);
-    if(d_k_copy.size() < input_size) {
+    if (d_k_copy.size() < input_size) {
         d_k_copy.resize(input_size, d_k);
     }
 
--- a/gr-blocks/lib/add_const_cc_impl.cc
+++ b/gr-blocks/lib/add_const_cc_impl.cc
@@ -43,6 +43,13 @@ add_const_cc_impl::add_const_cc_impl(gr_complex k)
 {
 }
 
+void add_const_cc_impl::set_k(gr_complex k)
+{
+    gr::thread::scoped_lock guard(d_setlock);
+    d_k = k;
+    std::fill(d_k_copy.begin(), d_k_copy.end(), d_k);
+}
+
 int add_const_cc_impl::work(int noutput_items,
                             gr_vector_const_void_star& input_items,
                             gr_vector_void_star& output_items)
@@ -50,23 +57,14 @@ int add_const_cc_impl::work(int noutput_items,
     const gr_complex* iptr = (const gr_complex*)input_items[0];
     gr_complex* optr = (gr_complex*)output_items[0];
 
-    int size = noutput_items;
+    unsigned long input_size = static_cast<unsigned long> (noutput_items);
 
-    while (size >= 8) {
-        *optr++ = *iptr++ + d_k;
-        *optr++ = *iptr++ + d_k;
-        *optr++ = *iptr++ + d_k;
-        *optr++ = *iptr++ + d_k;
-        *optr++ = *iptr++ + d_k;
-        *optr++ = *iptr++ + d_k;
-        *optr++ = *iptr++ + d_k;
-        *optr++ = *iptr++ + d_k;
-        size -= 8;
+    gr::thread::scoped_lock guard(d_setlock);
+    if(d_k_copy.size() < input_size) {
+        d_k_copy.resize(input_size, d_k);
     }
 
-    while (size-- > 0) {
-        *optr++ = *iptr++ + d_k;
-    }
+    volk_32fc_x2_add_32fc(optr, iptr, d_k_copy.data(), noutput_items);
 
     return noutput_items;
 }
--- a/gr-blocks/lib/add_const_ff_impl.cc
+++ b/gr-blocks/lib/add_const_ff_impl.cc
@@ -26,7 +26,6 @@
 
 #include "add_const_ff_impl.h"
 #include <gnuradio/io_signature.h>
-#include <volk/volk.h>
 
 namespace gr {
 namespace blocks {
@@ -44,6 +43,12 @@ add_const_ff_impl::add_const_ff_impl(float k)
 {
 }
 
+void add_const_ff_impl::set_k(float k)
+{
+    d_k = k;
+    std::fill(d_k_copy.begin(), d_k_copy.end(), d_k);
+}
+
 int add_const_ff_impl::work(int noutput_items,
                             gr_vector_const_void_star& input_items,
                             gr_vector_void_star& output_items)
@@ -51,18 +56,14 @@ int add_const_ff_impl::work(int noutput_items,
     const float* iptr = (const float*)input_items[0];
     float* optr = (float*)output_items[0];
 
-    int size = noutput_items;
+    unsigned long input_size = (unsigned long)noutput_items;
 
-    // build column vec. d_k * ones(size, 1) for volk addition
-    unsigned int alignment = volk_get_alignment();
-    float * k_copy_vec = (float*)volk_malloc(sizeof(float)*size, alignment);
-    for(int ii = 0; ii < size; ii++) {
-        k_copy_vec[ii] = d_k;
+    if(d_k_copy.size() < input_size) {
+        d_k_copy.resize(noutput_items, d_k);
     }
 
-    volk_32f_x2_add_32f(optr, iptr, k_copy_vec, size);
+    volk_32f_x2_add_32f(optr, iptr, &d_k_copy[0], noutput_items);
 
-    volk_free(k_copy_vec);
     return noutput_items;
 }
 
--- a/gr-blocks/lib/add_const_ff_impl.cc
+++ b/gr-blocks/lib/add_const_ff_impl.cc
@@ -57,10 +57,10 @@ int add_const_ff_impl::work(int noutput_items,
     const float* iptr = (const float*)input_items[0];
     float* optr = (float*)output_items[0];
 
-    unsigned long input_size = static_cast<unsigned long> (noutput_items);
+    unsigned long input_size = static_cast<unsigned long>(noutput_items);
 
     gr::thread::scoped_lock guard(d_setlock);
-    if(d_k_copy.size() < input_size) {
+    if (d_k_copy.size() < input_size) {
         d_k_copy.resize(input_size, d_k);
     }
 
--- a/gr-blocks/lib/add_const_ff_impl.cc
+++ b/gr-blocks/lib/add_const_ff_impl.cc
@@ -26,6 +26,7 @@
 
 #include "add_const_ff_impl.h"
 #include <gnuradio/io_signature.h>
+#include <volk/volk.h>
 
 namespace gr {
 namespace blocks {
@@ -52,22 +53,16 @@ int add_const_ff_impl::work(int noutput_items,
 
     int size = noutput_items;
 
-    while (size >= 8) {
-        *optr++ = *iptr++ + d_k;
-        *optr++ = *iptr++ + d_k;
-        *optr++ = *iptr++ + d_k;
-        *optr++ = *iptr++ + d_k;
-        *optr++ = *iptr++ + d_k;
-        *optr++ = *iptr++ + d_k;
-        *optr++ = *iptr++ + d_k;
-        *optr++ = *iptr++ + d_k;
-        size -= 8;
+    // build column vec. d_k * ones(size, 1) for volk addition
+    unsigned int alignment = volk_get_alignment();
+    float * k_copy_vec = (float*)volk_malloc(sizeof(float)*size, alignment);
+    for(int ii = 0; ii < size; ii++) {
+        k_copy_vec[ii] = d_k;
     }
 
-    while (size-- > 0) {
-        *optr++ = *iptr++ + d_k;
-    }
+    volk_32f_x2_add_32f(optr, iptr, k_copy_vec, size);
 
+    volk_free(k_copy_vec);
     return noutput_items;
 }
 
--- a/gr-blocks/lib/add_const_ff_impl.cc
+++ b/gr-blocks/lib/add_const_ff_impl.cc
@@ -45,6 +45,7 @@ add_const_ff_impl::add_const_ff_impl(float k)
 
 void add_const_ff_impl::set_k(float k)
 {
+    gr::thread::scoped_lock gaurd(d_setlock);
     d_k = k;
     std::fill(d_k_copy.begin(), d_k_copy.end(), d_k);
 }
@@ -56,13 +57,14 @@ int add_const_ff_impl::work(int noutput_items,
     const float* iptr = (const float*)input_items[0];
     float* optr = (float*)output_items[0];
 
-    unsigned long input_size = (unsigned long)noutput_items;
+    unsigned long input_size = static_cast<unsigned long> (noutput_items);
 
+    gr::thread::scoped_lock gaurd(d_setlock);
     if(d_k_copy.size() < input_size) {
-        d_k_copy.resize(noutput_items, d_k);
+        d_k_copy.resize(input_size, d_k);
     }
 
-    volk_32f_x2_add_32f(optr, iptr, &d_k_copy[0], noutput_items);
+    volk_32f_x2_add_32f(optr, iptr, d_k_copy.data(), noutput_items);
 
     return noutput_items;
 }
--- a/gr-blocks/lib/add_const_ff_impl.cc
+++ b/gr-blocks/lib/add_const_ff_impl.cc
@@ -45,7 +45,7 @@ add_const_ff_impl::add_const_ff_impl(float k)
 
 void add_const_ff_impl::set_k(float k)
 {
-    gr::thread::scoped_lock gaurd(d_setlock);
+    gr::thread::scoped_lock guard(d_setlock);
     d_k = k;
     std::fill(d_k_copy.begin(), d_k_copy.end(), d_k);
 }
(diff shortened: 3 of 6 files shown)


gr-blocks/lib/add_const_ff_impl.cc
```diff
@@ -52,22 +53,16 @@ int add_const_ff_impl::work(int noutput_items,
...
-        *optr++ = *iptr++ + d_k;
-        *optr++ = *iptr++ + d_k;
-        *optr++ = *iptr++ + d_k;
-        *optr++ = *iptr++ + d_k;
-        *optr++ = *iptr++ + d_k;
-        *optr++ = *iptr++ + d_k;
-        *optr++ = *iptr++ + d_k;
-        *optr++ = *iptr++ + d_k;
-        size -= 8;
+    // build column vec. d_k * ones(size, 1) for volk addition
```
michaelld: This whole section of code should be moved to the `set` method, which will have to be moved from the header `gr-blocks/lib/add_const_ff_impl.h` to this file. And then it will have to be called at instantiation as well. At instantiation declaration, set the copy_vec to NULL. Then in `set` check whether the copy_vec is NULL & if not then `volk_free` it. Same check in the destructor. Not trivial, but it'll clean up this code and make it faster and safer!

grant-cox:
@michaelld thank you for the feedback. I will look into it and ask if there are any issues.

As a related issue to this (@marcusmueller may want to see this), there are at least two more places in the block implementations where the same non-volk loop is used, i.e. in [add_const_ii_impl.cc](https://github.com/gnuradio/gnuradio/blob/b6580394aea850f5176469d854696e4f55e11918/gr-blocks/lib/add_const_ii_impl.cc#L55-L69) and also in [add_const_bb_impl.cc](https://github.com/gnuradio/gnuradio/blob/b6580394aea850f5176469d854696e4f55e11918/gr-blocks/lib/add_const_bb_impl.cc#L57-L71), maybe others. Should I go ahead and look into these here or does refactoring multiple implementations warrant a new issue? I'm happy to do it, let me know what you think.

michaelld:
@grant-cox since these are basically the same code, please go ahead and do them all. It might be wise to finish up this one first & push the changes, then we can review & finalize. Then the changes to the others will be straight forward / known.

grant-cox:
> @grant-cox since these are basically the same code, please go ahead and do them all. It might be wise to finish up this one first & push the changes, then we can review & finalize. Then the changes to the others will be straight forward / known.

@michaelld Here is what I've done so far, but I have a question about the initialization of the copy vector.
1. added d_k_copy_vec as a private member 
`class BLOCKS_API add_const_ff_impl : public add_const_ff`
`{`
`private:`
    `float d_k;`
    `float * d_k_copy_vec;`
2. added getter for `d_k_copy_vec` and moved `set_k` to the .cc file
`float * k_copy_vec() const { return d_k_copy_vec; }`
 `void set_k(float k);`
3. set the pointer to NULL in the constructor
4. in `set_k()`, `volk_free(d_k_copy_vec)` if not null

I am confused at step `4`. Forgive me if it's a basic question, this is my first time with C++ in this codebase. In `set_k()`, I don't have access to the size (number of input items) that will be present in `work()`. If I understand correctly, I should initialize `d_k_copy_vec` in `set_k()`, correct? I don't see how to do it aside from making a vector of modest size and resizing it in `work()` (worse than what I was already doing). Thanks for any help!

hcab14:
It would be good to use `volk::vector<float>` (see https://github.com/gnuradio/volk/blob/master/include/volk/volk_alloc.hh) instead of a raw pointer, but for this the latest master branch of volk needs to be used.

michaelld:
@grant-cox Ah ... you make a good point there about not knowing the "work" number of items size in `set_k`! Here's an alternative algorithm, which will still involve a `set_k`-like method:

* instead of `d_k_copy_vec` being a pointer, you could follow @hcab14 's recommendation and use `volk::vector<float>` ... we'll get a release out soon-ish and this will be a feature of it! This usage will then keep both the data as well as the size of the vector. By default (at instantiation), this vector will be of `size()` 0. 

--> If you want to use 2 variables for now that's fine: `size_t d_vec_size`, and `float* d_k_copy_vec`. They need to be cleared at instantiation.

* in `work`, check `d_k_copy_vec.size() >= noutput_items` (or `d_vec_size >= noutput_items`), and if so, then use the copy vector as is: it contains more than enough items to do the addition.

* if not, then either `resize` the `d_k_copy_vec` to the correct size using the `d_k` as the default setting (which might not work, since `volk::vector` doesn't provide a `resize` ... but we can add one if necessary); or if doing the 2-variable method, you'll need to do something like:
```
if (d_vec_size < noutput_items) {
  unsigned int alignment = volk_get_alignment();
  float* new_k_copy_vec = (float*)volk_malloc(sizeof(float)*size, alignment);
  // copy existing copy vector from current to new
  memcpy(new_k_copy_vec, d_k_copy_vec, d_vec_size*sizeof (float));
  // set new entries to the constant
  for (size_t nn = d_vec_size; nn < noutput_items; ++nn) {
    new_k_copy_vec[nn] = d_k;
  }
  // if the current copy vector exists, free it
  if (!d_k_copy_vec) {
    volk_free(d_k_copy_vec);
  }
  // set the new copy vector
  d_k_copy_vec = new_k_copy_vec;
}
[volk stuff]
```

* on destruction, if using the 2 variable method:
```
  if (!d_k_copy_vec) {
    volk_free(d_k_copy_vec);
  }
```

grant-cox:
@michaelld I think that adding a `resize()` method to `volk::vector<float>` is the best long term solution, given that this is an improvement issue and this method will be copied across multiple fundamental `add_const_XX` blocks. I do see in https://github.com/gnuradio/volk/blob/master/include/volk/volk_alloc.hh that `volk::vector` type aliases to `std::vector`, which has a [resize method](http://www.cplusplus.com/reference/vector/vector/resize). Is that not usable here? If not, should I postpone this and file an issue in the volk repository?

hcab14:
> @michaelld I think that adding a `resize()` method to `volk::vector<float>` is the best long term solution, given that this is an improvement issue and this method will be copied across multiple fundamental `add_const_XX` blocks. I do see in https://github.com/gnuradio/volk/blob/master/include/volk/volk_alloc.hh that `volk::vector` type aliases to `std::vector`, which has a [resize method](http://www.cplusplus.com/reference/vector/vector/resize). Is that not usable here? If not, should I postpone this and file an issue in the volk repository?

`volk::vector` is an alias for `std::vector` with a custom allocator (+deallocator). So all methods of `std::vector` are available for `volk::vector` as well.

grant-cox:
> > @michaelld I think that adding a `resize()` method to `volk::vector<float>` is the best long term solution, given that this is an improvement issue and this method will be copied across multiple fundamental `add_const_XX` blocks. I do see in https://github.com/gnuradio/volk/blob/master/include/volk/volk_alloc.hh that `volk::vector` type aliases to `std::vector`, which has a [resize method](http://www.cplusplus.com/reference/vector/vector/resize). Is that not usable here? If not, should I postpone this and file an issue in the volk repository?
> 
> `volk::vector` is an alias for `std::vector` with a custom allocator (+deallocator). So all methods of `std::vector` are available for `volk::vector` as well.

@hcab14 thanks for the confirmation. I'll use @michaelld 's method with `resize` and come back here later.

grant-cox:
@michaelld @hcab14 what is estimated release date for volk to have `volk::vector<float>`? I could use `std::vector` in the meantime, but I'd prefer to use native volk. Let me know what you think.

michaelld:
@grant-cox there is no set time yet ... but "sooner rather than later" is our hope ... like in December still.

grant-cox:
> @grant-cox there is no set time yet ... but "sooner rather than later" is our hope ... like in December still.

@michaelld Understood. Would you prefer that I implement this with `float *` for now, then open another issue to VOLKize all of those blocks with `volk::vector` when the volk release comes?

michaelld:
@grant-cox hmmm ... if you don't mind using the current volk GIT master & `volk::vector` then that will give me/us some incentive to get the Volk 2.0.1 release dealt with ... sooner rather than later so that this PR doesn't linger or get stale. Thx!

grant-cox:
> @grant-cox hmmm ... if you don't mind using the current volk GIT master & `volk::vector` then that will give me/us some incentive to get the Volk 2.0.1 release dealt with ... sooner rather than later so that this PR doesn't linger or get stale. Thx!

@michaelld I updated the changes using the current master branch of volk and `volk::vector` as we discussed. If you approve of these, I'll go ahead and implement it for the other add_const blocks that still have the while loops. Thanks!

gr-blocks/lib/add_const_ff_impl.cc
```diff
@@ -43,30 +43,26 @@ add_const_ff_impl::add_const_ff_impl(float k)
 {
 }
 
+void add_const_ff_impl::set_k(float k)
+{
```
michaelld: Put the following here: `gr::thread::scoped_lock guard(d_setlock);`

gr-blocks/lib/add_const_ff_impl.cc
```diff
@@ -43,30 +43,26 @@ add_const_ff_impl::add_const_ff_impl(float k)
...
+
 int add_const_ff_impl::work(int noutput_items,
                             gr_vector_const_void_star& input_items,
                             gr_vector_void_star& output_items)
 {
     const float* iptr = (const float*)input_items[0];
     float* optr = (float*)output_items[0];
 
-    int size = noutput_items;
+    unsigned long input_size = (unsigned long)noutput_items;
```
michaelld: Put the following here: `gr::thread::scoped_lock guard(d_setlock);`

gr-blocks/lib/add_const_ff_impl.cc
```diff
@@ -43,30 +43,26 @@ add_const_ff_impl::add_const_ff_impl(float k)
...
-        *optr++ = *iptr++ + d_k;
-        size -= 8;
+    if(d_k_copy.size() < input_size) {
+        d_k_copy.resize(noutput_items, d_k);
     }
 
-    while (size-- > 0) {
-        *optr++ = *iptr++ + d_k;
-    }
+    volk_32f_x2_add_32f(optr, iptr, &d_k_copy[0], noutput_items);
```
michaelld: Since we're using c++11, let's use `d_k_copy.data()` instead of `&d_k_copy[0]`

gr-blocks/lib/add_const_ff_impl.cc
```diff
@@ -43,30 +43,26 @@ add_const_ff_impl::add_const_ff_impl(float k)
...
-        *optr++ = *iptr++ + d_k;
-        *optr++ = *iptr++ + d_k;
-        *optr++ = *iptr++ + d_k;
-        *optr++ = *iptr++ + d_k;
-        *optr++ = *iptr++ + d_k;
-        *optr++ = *iptr++ + d_k;
-        *optr++ = *iptr++ + d_k;
-        size -= 8;
+    if(d_k_copy.size() < input_size) {
+        d_k_copy.resize(noutput_items, d_k);
```
michaelld: use `input_size` instead of `noutput_items` to match the `resize` API first argument is unsigned.

gr-blocks/lib/add_const_ff_impl.h
```diff
@@ -32,14 +32,20 @@ class BLOCKS_API add_const_ff_impl : public add_const_ff
...
     add_const_ff_impl(float k);
 
     void setup_rpc();
 
     float k() const { return d_k; }
-    void set_k(float k) { d_k = k; }
+
+    void set_k(float k);
+
+    volk::vector<float> k_copy() const { return d_k_copy; }
```
michaelld: Do we really need to return the vector? I can't think of a case where this is useful, but I'm open to suggestions. If there are cases, then I'd suggest using `const volk::vector<float>&` as the return type.

gr-blocks/lib/add_const_ff_impl.h
```diff
@@ -32,14 +32,20 @@ class BLOCKS_API add_const_ff_impl : public add_const_ff
...
+
+    size_t k_copy_size() const { return d_k_copy.size(); }
```
michaelld: Do we really need to return the vector size? I can't think of a case where this is useful, but I'm open to suggestions.

gr-blocks/lib/add_const_ff_impl.cc
michaelld: Reason: We don't want to be setting and using either class variable simultaneously (no "both/and" here), so we use a scoped lock to make sure it's "either/or" ...

gr-blocks/lib/add_const_ff_impl.cc
michaelld: prefer the C++ `static_cast < unsigned long > (noutput_items);` over `(unsigned long)noutput_items;` ... they both do the same thing, of course , so this is style preference ...

grant-cox:
> Thanks for your work! A few more tweaks & it'll be ready IMHO.

@michaelld I believe that addresses all of your requested changes. If those work for you, let me know if you'd like for me to refactor the other instances with volk. Thanks.

grant-cox:
> LGTM on current reading. Let me verify functionality locally & I'll report back and assuming positive then you can tackle the other `add_const` blocks. Thx!

@michaelld would you like for me to continue with the other blocks? Wanted to check in. Thanks!

michaelld:
> > LGTM on current reading. Let me verify functionality locally & I'll report back and assuming positive then you can tackle the other `add_const` blocks. Thx!
> 
> @michaelld would you like for me to continue with the other blocks? Wanted to check in. Thanks!

Sorry for the delay! I just tested this; it builds cleanly, which is expected and always good!

Yes, I'd say go ahead with the other implementations that can be volk-ified. The code is really simple now, and even without QA I'm sure it's OK.

Related: Would you be willing to create QA for these blocks? There does not seem to be any right now! It would be simple, like what `gr-blocks/python/blocks/qa_conjugate.py` does: just verify that adding some random cost to a short vector works. NP if not; just thought it worth asking since they are related!

michaelld:
@grant-cox Thanks for pointing out that QA ... I thought I did both 'find' and 'grep' looking for `add_const_ff` ... but ... well ... obviously I didn't see that! No, please just leave the QA alone ... sorry about the mis-direct!

That QA works just fine with your changes. I have a slightly older version (I think) that prints out during 'resize' and it shows that it's working as desired!

Thus: yes, Please do go ahead with the Volk-ification of the other `add_const` blocks as best works / makes sense. Thx!

grant-cox:
@michaelld as I'm looking through these, I saw there aren't any volk kernels for adding _two_ (x2)`short *` arrays togther (there is a [x5 adder](http://libvolk.org/doxygen/volk_16i_x5_add_quad_16i_x4.html) but I don't think that is what I need). I assume the x5 adder optimizes the 32-bit floating point units in the CPU by adding two 16-bit shorts in parallel...is there an x2 adder that I'm not seeing in the doxygen?

I could:
1. Use the input array, my `d_k_copy` vector, and 3 additional all-zero vectors to pass as the other inputs to the x5 adder, but this could cause problems with floating point errors (and it ain't pretty). 
2. Write a `volk_16i_x2_add_16i` kernel (don't know how hard this is).
3. Don't volkize `add_const_ss`.

I hope I'm overlooking something in the doxygen that makes this trivial. Maybe @hcab14 can chime in here as well?

michaelld:
For now: (4) Don't volkize add_const_ss. Just do the ones that are easy. We can revisit 'ss' in some other PR(s).

grant-cox:
> For now: (4) Don't volkize add_const_ss or `cc` or those that we don't already provide an `x2` volk function for. Just do the ones that are easy. We can revisit those others in some other PR(s).

@michaelld all right, I've volkized add_const_ff and add_const_cc. These built with volk master branch and passed qa_add_mult_div_sub. Let me know if changes are needed, and thank you for your help.

grant-cox:
> Excellent! LGTM. The CI will fail because Volk doesn't provide `volk_vector.hh` yet ... we need to get the next Volk release out the door ... then this can be merged!

@michaelld , I saw the recent volk release on the mailing list. Do you know when the new version will be checked out in gnuradio? If there are new gnuradio CI tests needed for the new volk features, I'd be happy to help. Thanks!

gr-blocks/include/gnuradio/blocks/add_const_cc.h
```diff
@@ -25,6 +25,7 @@
 
 #include <gnuradio/blocks/api.h>
 #include <gnuradio/sync_block.h>
+#include <volk/volk_alloc.hh>
```
mbr0wn: Why are they here if they're used in the impl? Was this already missing?

gr-blocks/lib/add_const_cc_impl.cc
```diff
@@ -43,30 +43,28 @@ add_const_cc_impl::add_const_cc_impl(gr_complex k)
...
+
 int add_const_cc_impl::work(int noutput_items,
                             gr_vector_const_void_star& input_items,
                             gr_vector_void_star& output_items)
 {
     const gr_complex* iptr = (const gr_complex*)input_items[0];
     gr_complex* optr = (gr_complex*)output_items[0];
 
-    int size = noutput_items;
+    unsigned long input_size = static_cast<unsigned long> (noutput_items);
```
mbr0wn: This looks like it doesn't agree with our clang-format.

gr-blocks/include/gnuradio/blocks/add_const_cc.h
grant-cox: @mbr0wn I put that in the public-facing header since it wouldn't build without it and it would be easily visible from the public facing header that the class was using VOLK. I will gladly move it to `add_const_cc_impl.h` and `add_const_ff_impl.h` if you'd like.

gr-blocks/lib/add_const_cc_impl.cc
grant-cox: @mbr0wn sorry, I didn't realize that. I hadn't seen in the guides to use git-clang-format. I'll push another commit to fix these unless you say otherwise. Thanks.

gr-blocks/include/gnuradio/blocks/add_const_cc.h
mbr0wn: Hm, I'm OK with that justification.

gr-blocks/lib/add_const_cc_impl.cc
mbr0wn: That would be great. It avoids us from having to run clang-format outside of other changes, which doesn't help with cherry-pickability and  bisecting.

grant-cox: @mbr0wn okay, these are fixed in the most recent commit. Thanks!

mbr0wn:
Merged manually (had to fix history).

PR #2918, opened 2019-11-21
cmake: add boost 1.71 as acceptable version

mormj:
Tested Boost 1.71 against the master branch.  All tests pass.

Also, Boost_ADDITIONAL_VERSIONS is an input to FindBoost, so should come before calling it.

### DIFF
--- a/cmake/Modules/GrBoost.cmake
+++ b/cmake/Modules/GrBoost.cmake
@@ -69,18 +69,20 @@ if(MSVC)
     endif(BOOST_ALL_DYN_LINK)
 endif(MSVC)
 
-find_package(Boost ${GR_BOOST_MIN_VERSION} COMPONENTS ${BOOST_REQUIRED_COMPONENTS})
-
 # This does not allow us to disable specific versions. It is used
 # internally by cmake to know the formation newer versions. As newer
 # Boost version beyond what is shown here are produced, we must extend
 # this list. To disable Boost versions, see below.
+
 set(Boost_ADDITIONAL_VERSIONS
     "1.53" "1.54.0" "1.54"
     "1.55.0" "1.55" "1.56.0" "1.56" "1.57.0" "1.57" "1.58.0" "1.58" "1.59.0" "1.59"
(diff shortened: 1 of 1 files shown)


michaelld:
Related items to note, for posterity as much as anything I guess:

* `Boost_ADDITIONAL_VERSIONS` is used by the CMake-provided `FindBoost.cmake` script only; it is not used by the Boost-provided CMake scripts.

* The Boost-provided CMake scripts are installed by default as of Boost 1.70.0.

* Boost 1.70.0 CMake scripts are broken. If they are found and used without workarounds, they will with very high probability cause the build to fail (not just GR, but pretty much any project using this Boost version). There are workarounds to make this Boost version work correctly (see, e.g., UHD's `UHDBoost.cmake` script).

* The Boost 1.70.0 API & ABI work just fine with GR, as I think we've patched GR to work with then (and, 1.71.0 as well).

* The purpose of `Boost_ADDITIONAL_VERSIONS` is for when:

(a) the CMake-provided `FindBoost.cmake` script is used -- regardless of whether the Boost-provided CMake scripts are installed, but generally the case will be that the Boost-provided CMake scripts are not installed or not found for some reason;

(b) the version of Boost is more recent than the `FindBoost.cmake` script is programmed for. This setting basically tells the `FindBoost.cmake` to warn, but go ahead and use the found newer-but-unknown Boost version with its best guess as to dependencies internal to Boost.

* The current CMake release (3.15.5) knows about Boost through 1.71.0. The CMake folks do a pretty good job of keeping up with known Boost versions in the `FindBoost.cmake` script. That said, there are likely OSs that use older CMake ... where if Boost is installed from source rather than a package manager then the `Boost_ADDITIONAL_VERSIONS` will be useful.

* As _most_ Boost versions are compatible with GR, and Boost is a bit of a pain to build and install successfully -- it's _so_ much easier to just install Boost using the OS package manager -- I really don't think most GR users will have a custom Boost install; they will just use whatever the OS package manager provides. If any reasonably modern CMake is installed from source this variable won't do anything as CMake will know about all Boost versions.

* Hence, from all of the above, I really doubt this variable will have meaning much longer, if it does right now at all for the vast majority of GR users. I'd guess that we can remove this variable right now -- but just for safety we should keep it around for another year.

PR #2923, opened 2019-11-23
revise install instructions

duggabe:
revise install instructions to add detail and add links to Wiki and PyBOMBS

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,9 @@ as many questions have already been asked and answered. Please also
 the mailing list and post your new questions there.
 
 
-## How to Build GNU Radio
+## How to Install GNU Radio
+
+### Prebuilt Binaries
 
 The recommended way to install GNU Radio on most platforms is using available binary package distributions. 
 
@@ -55,7 +57,7 @@ To run GNU Radio Companion, enter:
 
 Complete PyBOMBS instructions are in the [PyBOMBS README](https://github.com/gnuradio/pybombs#pybombs).
 
-### Build from Source
+### From Source
 
 Complete instructions for building Gnuradio from source code are detailed in 
 [Installing GR From Source](https://wiki.gnuradio.org/index.php/InstallingGR#From_Source). 
--- a/README.md
+++ b/README.md
@@ -28,13 +28,33 @@ the mailing list and post your new questions there.
 
 ## How to Build GNU Radio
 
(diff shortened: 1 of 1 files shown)


777arc:
FYI people can see this new version here- https://github.com/duggabe/gnuradio/tree/install-doc

README.md
```diff
@@ -28,102 +28,37 @@ the mailing list and post your new questions there.
 
 ## How to Build GNU Radio
```
michaelld: I'd recommend changing this header to "How to Obtain GNU Radio" or "How to Install GNU Radio" ... "build" to me implies compiling, which the immediate next paragraph is not about. I like the new paragraph, so let's fix this header.

README.md
```diff
@@ -28,102 +28,37 @@ the mailing list and post your new questions there.
...
+
+<b>Wait</b>. The terminal will show the progress.
+
+To run GNU Radio Companion, enter:
+
+        pybombs run gnuradio-companion
+
+Complete PyBOMBS instructions are in the [PyBOMBS README](https://github.com/gnuradio/pybombs#pybombs).
+
+### Build from Source
```
michaelld: Recommend removing "Build" ... just "From Source" here ... or even just "Source" to match "PyBOMBS" ...

README.md
duggabe: "How to Install GNU Radio" sounds good. There are actually three sub-topics here, so to be consistent, I suggest putting "Package Distributions" as a sub-header right under "How to Install GNU Radio". So then we would have that, "PyBOMBS", and "From Source"  (or do you like "Source Code" better?).

michaelld: "Install" sounds good. Sub-topics also work nicely there. "From Source" makes sense. "PyBOMBS" makes sense too. Maybe "Prebuilt Binaries" for the other ... some we are distributing (Windows) ... yes? Some are for package managers (definitely). We might include links for others that are also pre-built. How does that sound to you?

duggabe: Rather than complicating the page here, I recommend putting that info in the Wiki: https://wiki.gnuradio.org/index.php/InstallingGR#From_Binaries
let's keep it simple here. My theme is "document in ONE place". :)
I will make the updates tonight or in the morning and do another push.

michaelld: I like that theme. Thanks for working on this! I look forward to the next iteration of this PR!

777arc:
CLA has been executed, squashing/merging now

PR #2934, opened 2019-11-28
gr-zeromq: Added C++ templates for zeromq blocks

seeAppRun:


gr-zeromq/grc/zeromq_pub_msg_sink.block.yml
```diff
@@ -20,4 +21,10 @@ templates:
     imports: from gnuradio import zeromq
     make: zeromq.pub_msg_sink(${address}, ${timeout})
     
+cpp_templates:
+    includes: [ '#include <gnuradio/zeromq/pub_msg_sink.h>' ]
+    declarations: gr::zeromq::pub_msg_sink::sptr ${id};
+    make: this->${id} = gr::zeromq::pub_msg_sink::make(const_cast<char *>(${address}${'.c_str())' if str(address)[0] not in '"\'' else ')'}, ${timeout});
```
seeAppRun: An explanation of the make recipe:
The conditional checks whether the address is a string literal or a `std::string`. If it is a `std::string`, the variable name is used followed by `.c_str()`, otherwise the literal is used. This is because the underlying C++ constructor expects a `char *`, not a `std::string`.

This will produce correct code whether the address is a string literal in the zeromq block, or if it references a flowgraph variable or parameter.

haakov:
Looks good, great work. My laptop hasn't finished building yet, but I'll test it once it does.

mbr0wn:
Does this work without your PR on cppgen fixes?

gr-zeromq/grc/zeromq_pub_msg_sink.block.yml
mbr0wn: Thanks for the explanation.

mbr0wn:
Since this is purely additive I'll clean up the history and merge. We'll need the GRC example fixes too, but that doesn't make this changeset useless.
@haakov Many thanks for the review too!

mbr0wn:
Rebased and merged manually.

Issue #2928, opened 2019-11-26
vocoder: FreeDV text message off by one error


devnulling:
## Issue Description ##
There is a off by one error within the FreeDV vocoder modulator block for the text string.

Example flowgraph:
Note the `Text Message` is set to be: `GNU Radio1234`

![freedv](https://user-images.githubusercontent.com/11846824/69672025-1559ff80-104c-11ea-8016-dd316d9a77d3.png)

When the modulated wav file is played back through FreeDV, only `GNU Radio123` is demodulated. 

![freedv_rx](https://user-images.githubusercontent.com/11846824/69672071-3589be80-104c-11ea-8721-e92c67310fa3.png)

## Setup Details ##
- GR master https://github.com/gnuradio/gnuradio/commit/6414c7c642ab79aa2e96d2e3b394856470cc1444
- FreeDV 1.4 

- Modulated wav file: 
[encoded_gnuradio1234.zip](https://github.com/gnuradio/gnuradio/files/3893984/encoded_gnuradio1234.zip)

PR #2936, opened 2019-11-29
vocoder: Append carriage return instead of replace last char

devnulling:
This pr appends a carriage return char to the end of the text string instead of looping and replacing the last char.

Fixes: #2928

### DIFF
--- a/gr-vocoder/lib/freedv_tx_ss_impl.cc
+++ b/gr-vocoder/lib/freedv_tx_ss_impl.cc
@@ -42,7 +42,6 @@ char get_next_tx_char(void* callback_state)
 
     if (*pstate->ptx_str == 0) {
         pstate->ptx_str = pstate->tx_str;
-        c = 0x0d; // FreeDV uses Carriage Return termination
     }
 
     return c;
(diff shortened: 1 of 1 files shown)


gr-vocoder/lib/freedv_tx_ss_impl.cc
```diff
@@ -82,7 +81,7 @@ freedv_tx_ss_impl::freedv_tx_ss_impl(int mode,
     if ((d_freedv = freedv_open(mode)) == NULL)
         throw std::runtime_error("freedv_tx_ss_impl: freedv_open failed");
 #endif
-    snprintf(d_cb_state.tx_str, 79, "%s", d_msg_text.c_str());
+    snprintf(d_cb_state.tx_str, 80, "%s\r", d_msg_text.c_str());
```
marcusmueller: please investigate max length of freedv payload

devnulling: so this fix actually overflows and doesn't catch the carriage return. I'll find another solution and resubmit.

PR #2939, opened 2019-11-30
cmake: Use standard way of getting python prefix

### LINKED ISSUE #2006: dist-packages is Debian specific and not a general "unix thing"

smunaut:
Fixes #2006

This is exactly the same logic as is used in volk. Also the same logic
used for the GR_PYTHON_RELATIVE just below ... so if it works for those
on all OSes/config, no reasons it won't work here.

Signed-off-by: Sylvain Munaut <tnt@246tNt.com>

### DIFF
--- a/cmake/Modules/GrPython.cmake
+++ b/cmake/Modules/GrPython.cmake
@@ -109,12 +109,8 @@ endmacro(GR_PYTHON_CHECK_MODULE)
 ########################################################################
 if(NOT DEFINED GR_PYTHON_DIR)
 execute_process(COMMAND ${PYTHON_EXECUTABLE} -c "
-import os
-import sys
-if os.name == 'posix':
-    print(os.path.join('lib', 'python' + sys.version[:3], 'dist-packages'))
-if os.name == 'nt':
-    print(os.path.join('Lib', 'site-packages'))
+from distutils import sysconfig
+print(sysconfig.get_python_lib(plat_specific=True, prefix=''))
 " OUTPUT_VARIABLE GR_PYTHON_DIR OUTPUT_STRIP_TRAILING_WHITESPACE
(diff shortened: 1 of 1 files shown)


smunaut:
Build failures are completely unrelated to the patch ... something is just broken in master ATM ...

michaelld:
Related to #2781

nickoe:
I have tested this on archlinux by cherry-picking to maint-3.8 using a default of python 3.8, and this works great! :)

nickoe:
This also works for python2, but I need to turn off the enable internal volk build option:
```
cmake -DCMAKE_INSTALL_PREFIX=/tmp/gnuradio -DENABLE_GR_QTGUI=ON -DPYTHON_EXECUTABLE=$(which python2) -DENABLE_INTERNAL_VOLK=OFF ..
```

michaelld:
@nickoe interesting ... what does turning off internal Volk do that fixes some issue with Python?

nickoe:
@michaelld no, _not_ turning enable internal volk off just makes cmake configure step fail with:
```
CMake Error at volk/CMakeLists.txt:93 (message):
  Mako templates required to build VOLK
```

But I suspect this to be a completely unrelated issue than the site-packages vs dist-packages issue. So just ignore that, I just wanted to confirm that I like this patch. I would love to see it merged ASAP.

nickoe:
@michaelld  Yes, the mako issue was simply because I didn't have the mako module for python2 installed.

michaelld:
> @michaelld Yes, the mako issue was simply because I didn't have the mako module for python2 installed.

:+1:

smunaut:
So ... everyone agrees ? Can someone presse the big green "merge" button ?

PR #2946, opened 2019-12-04
digital: add filter response truncation to generic mod

### LINKED ISSUE #2920: gr3.7/3.8 | Constellation modulator | data gets shifted
Data is shifted (rotated) after going through a `constellation modulator`.

Input data: 0b**00001111**
output data: 0b**00111100**

image

Could this be an (un)expected effect of the RRC filter?

```python
generic_mod(gr.hier_block2):
(...)
# pulse shaping filter
        nfilts = 32
        ntaps = nfilts * 11 * int(self._samples_per_symbol)    # make nfilts filters of ntaps each
        self.rrc_taps = filter.firdes.root_raised_cosine(
            nfilts,          # gain
            nfilts,          # sampling rate based on 32 filters in resampler
            1.0,             # symbol rate
            self._excess_bw, # excess bandwidth (roll-off factor)
            ntaps)
        self.rrc_filter = filter.pfb_arb_resampler_ccf(self._samples_per_symbol,
                                                       self.rrc_taps)
(...)
```

mormj:
The generic mod implementation is a convenience hier block to modulate
bits to symbols and apply an RRC filter.  One downside is the output is
delayed by the length of the RRC filter (which is specified inside the
block).  This adds an option to truncate the output according to the
length of the filter response such that the start of output is aligned to the first symbol.

For instance, with the following flowgraph:

image

the output bursts are not aligned:

image

But when the truncation flag is set, the pulses are now aligned because the filter response was sent through a skip head:

image

And we get aligned bursts of 00001111 as we would expect
image

Fixes #2920

### DIFF
--- a/gr-digital/python/digital/generic_mod_demod.py
+++ b/gr-digital/python/digital/generic_mod_demod.py
@@ -41,6 +41,7 @@ _def_samples_per_symbol = 2
 _def_excess_bw = 0.35
 _def_verbose = False
 _def_log = False
+_def_truncate = False
 
 # Frequency correction
 _def_freq_bw = 2*math.pi/100.0
@@ -93,6 +94,7 @@ class generic_mod(gr.hier_block2):
         excess_bw: Root-raised cosine filter excess bandwidth (float)
         verbose: Print information about modulator? (boolean)
         log: Log modulation data to files? (boolean)
+        truncate: Truncate the modulated output to account for the RRC filter response (boolean)
     """
 
     def __init__(self, constellation,
@@ -101,7 +103,8 @@ class generic_mod(gr.hier_block2):
                  pre_diff_code=True,
                  excess_bw=_def_excess_bw,
                  verbose=_def_verbose,
-                 log=_def_log):
+                 log=_def_log,
+                 truncate=_def_truncate):
 
         gr.hier_block2.__init__(self, "generic_mod",
                                 gr.io_signature(1, 1, gr.sizeof_char),       # Input signature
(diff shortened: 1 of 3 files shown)


gr-digital/python/digital/generic_mod_demod.py
```diff
@@ -143,13 +146,21 @@ def __init__(self, constellation,
         self.rrc_filter = filter.pfb_arb_resampler_ccf(self._samples_per_symbol,
                                                        self.rrc_taps)
 
+        # Remove the filter transient at the beginning of the transmission
+        if truncate:
+            self.skiphead = blocks.skiphead(gr.sizeof_gr_complex*1, 2*11*int(self._samples_per_symbol)-1)
```
mbr0wn: Where does 2*11 come from?

mbr0wn: ...can you please add a comment.

mormj: that is an incorrect calculation of the delay caused by the pfb resampler.  need to redo the math, as it is not coming out what i expect it to be for various sample per symbol settings

michaelld:
Is this PR finished? If so, let's review it & deal with it!

mormj:
Yes, I'm good with this one - let's get some reviews

haakov:
Does this interfere with any example flowgraphs?

mormj:
@haakov - the default behavior (no truncation) is maintained.  Would impact some of the digital correlation estimation flowgraphs, but only if the user changes the truncation selection.

michaelld:
Changes leave things as current but with the option to use the new code. CI passes. 2 approvals. Merging.

PR #2955, opened 2019-12-11
Update FindLOG4CPP to include current installation location for Ubuntu

lospheris:
Adds /usr/include as a directory for Log4cpp. It appears to be the default installation location for ubuntu.


michaelld:
FYI `/usr/include` is already in the search path for the header `log4cpp/Category.hh` ...

lospheris:
My computer for this is at work, but you can see on the Ubuntu Packages page that the headers are installed to /usr/include

https://packages.ubuntu.com/focal/amd64/liblog4cpp5-dev/filelist

I compiled GNURadio earlier today and it refused to compile until I added the directory in the cmake file.

PR #2958, opened 2019-12-14
Add a LICENSE.txt file with GPL-3.0 license 

sam-caldwell:
This will make the LICENSE block in the main github screen turn green.

### DIFF
--- a/LICENSE.txt
+++ b/LICENSE.txt
@@ -1,0 +1,674 @@
+                    GNU GENERAL PUBLIC LICENSE
+                       Version 3, 29 June 2007
+
+ Copyright (C) 2007 Free Software Foundation, Inc. <http://fsf.org/>
+ Everyone is permitted to copy and distribute verbatim copies
+ of this license document, but changing it is not allowed.
+
+                            Preamble
+
+  The GNU General Public License is a free, copyleft license for
+software and other kinds of works.
+
+  The licenses for most software and other practical works are designed
+to take away your freedom to share and change the works.  By contrast,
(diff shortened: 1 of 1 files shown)


devnulling:
Hi @sam-caldwell 

We will need a CLA in place in order to merge this commit. Please email Ben Hilburn at bhilburn@gnuradio.org to get the CLA process started.

marcusmueller:
@bhilburn  I think we can merge this without any ado (esp. without CLA)

777arc:
Seems like we should just be renaming COPYING.txt to LICENSE.txt, not creating a second file with the same content.  Want to go ahead and do that yourself @marcusmueller ?

bhilburn:
So, GNU Radio follows the GNU guidelines, which require the GPL to be in a
file called "COPYING". See here:

https://www.gnu.org/licenses/gpl-howto.en.html

I think I'm confused about what "turning the block green" does? The
`COPYING` name is obviously supported by Github's license auto-detect since
it shows up as GPL-3.0 on our project homepage. Is there something further
we would get by also having a `LICENSE` file?

777arc:
By "turning the block green" we mean what gets rendered at the top of the README when you visit https://github.com/gnuradio/gnuradio .   And I just realized it's not because our license file has the wrong name, it's because we are currently using https://img.shields.io/aur/license/yaourt.svg but should be using https://img.shields.io/github/license/gnuradio/gnuradio so I'll close this PR and create a new one with the minor change to the README.  Thanks @sam-caldwell for noticing it is broken and trying to fix it!

PR #2963, opened 2019-12-18
gr-blocks: SimpleXMLRPCServer was moved in Python3

jkbecker:
The **XML RPC Server** block will not work in a python3-based GNU Radio installation, because of this:

> **Note** The SimpleXMLRPCServer module has been merged into xmlrpc.server in Python 3.
> (Source: https://docs.python.org/2/library/simplexmlrpcserver.html)

This PR fixes the issue and works correctly in Python 3.

NB: This should probably also make its way into maint-3.8, but I'm new to the party... please let me know if I should resubmit in a different way to achieve this.

### DIFF
--- a/gr-blocks/grc/xmlrpc_server.block.yml
+++ b/gr-blocks/grc/xmlrpc_server.block.yml
@@ -14,10 +14,10 @@ parameters:
 
 templates:
     imports: |-
-        import SimpleXMLRPCServer
+        from xmlrpc.server import SimpleXMLRPCServer
         import threading
     make: |-
-        SimpleXMLRPCServer.SimpleXMLRPCServer((${addr}, ${port}), allow_none=True)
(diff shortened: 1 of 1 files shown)


mbr0wn:
@jkbecker maint-3.8 needs to support both Python2 and Python3, so that's an issue. However, in general, you can just flag your PR as "backport to 3.8" (see labels) and we'll handle it.

jkbecker:
I'll be traveling until mid January, won't be able to look into it before then. If anyone else wants to pick this up for py2 compatibility feel free to, otherwise I'll give it a shot later this month.

mbr0wn:
@jkbecker thanks. A backport to maint-3.8 would be good but that doesn't invalidate this PR. Reminder for us: https://github.com/gnuradio/gnuradio/issues/3011

PR #2964, opened 2019-12-19
Fix Log4Cpp cmake dependency

japm48:
This bug breaks `gr-iio` build in Arch Linux.

This was partially fixed in 23bf13685faabe40fed4314534ae71936cdbeece.
Apparently, proper capitalization is hard :)

### DIFF
--- a/cmake/Modules/FindLOG4CPP.cmake
+++ b/cmake/Modules/FindLOG4CPP.cmake
@@ -48,7 +48,7 @@ else ()
 endif ()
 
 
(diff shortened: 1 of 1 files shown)


nickoe:
Maybe @eblot should verify this as well?

japm48:
More details: I used branch `upgrade-3.8` from https://github.com/analogdevicesinc/gr-iio

I believe this happens whenever `FindLOG4CPP.cmake` is executed twice, so this probably breaks other OOT modules.

michaelld:
Given how trivial this change is, and how important it is somehow for CMake, I'm going to merge without further comment.

PR #2967, opened 2019-12-20
grc: loosen the validation of block in function probe

### LINKED ISSUE #2966: Can't omit Block ID when using Function Probe
Documentation for the Function Probe block in GRC states: "An empty Block ID references the flow graph itself." However, the `dtype` of the `block_id` is now set to `id` as of this PR: https://github.com/gnuradio/gnuradio/pull/2511/files#diff-09bef8b2a2b7093fa2be676d03a7cf24L8-R8

So now, when the Block ID field is left blank, the following validation error occurs:

> Param - Block ID(block_id):
> 	ID "" must begin with a letter and may contain letters, numbers, and underscores.

Is there a way to allow for empty `block_id` while still performing `id` validation when non-empty?

mormj:
The block_id parameter for the block being probed was being validated as
a block_id.  But there is a use case where this can be blank to call a
function on the flowgraph object itself.

Changed the block_id dtype to "name" just like function_name.  Could
have just changed it to raw and done no validation, but that doesn't get
displayed very nicely in the grc block

Fixes #2966

### DIFF
--- a/grc/core/params/dtypes.py
+++ b/grc/core/params/dtypes.py
@@ -70,10 +70,12 @@ def validate_block_id(param):
 
 @validates('name')
 def validate_name(param):
-    # Name of a function that will be generated literally not as a string
+    # Name of a function or other block that will be generated literally not as a string
     value = param.value
+
+    # Allow blank to pass validation
     # Can python use this as a variable?
-    if not re.match(r'^[a-z|A-Z]\w*$', value):
+    if not re.match(r'^([a-z|A-Z]\w*)?$', value):
         raise ValidateError('ID "{}" must begin with a letter and may contain letters, numbers, '
                             'and underscores.'.format(value))
     return value
--- a/grc/blocks/variable_function_probe.block.yml
+++ b/grc/blocks/variable_function_probe.block.yml
@@ -5,7 +5,7 @@ flags: [ show_id, python ]
 parameters:
 -   id: block_id
     label: Block ID
-    dtype: id
+    dtype: name
     default: my_block_0
 -   id: function_name
     label: Function Name


smunaut:
Doesn't that get rid of validation for a lot of other things ?
Couldn't we just have an 'optional' attribute or something to that effect ?

mormj:
That particular dtype is only used in one place - for giving the name of another function such as in the function probe block.  It's just a string, but there was an issue in the changes for 3.8 where some things that used to be strings got extra '' added when the fg was generated.  So this was a workaround to have a string type that gets generated without ''.  So I think no other validation will be broken other than it will now let you put a blank function name in the function probe :\

I like the idea of the optional attribute so that any parameter can be given the "it's ok if left blank" - will give that a shot.

grc/core/params/dtypes.py
```diff
@@ -70,10 +70,12 @@ def validate_block_id(param):
 
 @validates('name')
 def validate_name(param):
-    # Name of a function that will be generated literally not as a string
+    # Name of a function or other block that will be generated literally not as a string
```
mbr0wn: These changes should go into another commit.

grc/core/params/dtypes.py
```diff
@@ -70,10 +70,12 @@ def validate_block_id(param):
...
     value = param.value
+
+    # Allow blank to pass validation
     # Can python use this as a variable?
-    if not re.match(r'^[a-z|A-Z]\w*$', value):
+    if not re.match(r'^([a-z|A-Z]\w*)?$', value):
```
mbr0wn: Since you're touching the regex, it's broken: Since we're in Python3, it will match way too broadly. This needs the ASCII flag.

mormj: will redo this with the additonal attribute as mentioned above as a separate pr rather than opening up the regex for this dtype.  closing this one for now

the other option is just to set the dtype to raw and lose parameter validation which isn't really thorough in this case anyway.

PR #2968, opened 2019-12-21
volk: Add VOLK signify public key

jdemel:
This is to verify VOLK releases with a known key.


mbr0wn:
Do we even need this in GNU Radio?

mbr0wn:
@michaelld can merge this if it's OK.

marcusmueller:
I'll go ahead and merge this, @mbr0wn: I've discussed with @michaelld how we can have a chain of trust for our purposes.

PR #2969, opened 2019-12-21
digital/map_bb: Fix buffer overflow

ThomasHabets:
This also breaks out 0x100 to have one source of truth.

### DIFF
--- a/gr-digital/lib/map_bb_impl.cc
+++ b/gr-digital/lib/map_bb_impl.cc
@@ -61,6 +61,7 @@ void map_bb_impl::set_map(const std::vector<int>& map)
 std::vector<int> map_bb_impl::map() const
 {
     std::vector<int> m(s_map_size);
+    gr::thread::scoped_lock guard(d_mutex);
     for (unsigned i = 0; i < s_map_size; i++) {
         m[i] = d_map[i];
     }
--- a/gr-digital/lib/map_bb_impl.cc
+++ b/gr-digital/lib/map_bb_impl.cc
@@ -49,19 +49,21 @@ void map_bb_impl::set_map(const std::vector<int>& map)
 {
     gr::thread::scoped_lock guard(d_mutex);
 
-    for (int i = 0; i < 0x100; i++)
+    for (int i = 0; i < s_map_size; i++) {
         d_map[i] = i;
+    }
 
-    unsigned int size = std::min((size_t)0x100, map.size());
+    const unsigned int size = std::min(s_map_size, map.size());
     for (unsigned int i = 0; i < size; i++)
         d_map[i] = map[i];
 }
 
 std::vector<int> map_bb_impl::map() const
 {
-    std::vector<int> m;
-    for (unsigned i = 0; i < 0x100; i++)
+    std::vector<int> m(s_map_size);
+    for (unsigned i = 0; i < s_map_size; i++) {
         m[i] = d_map[i];
+    }
     return m;
 }
 
--- a/gr-digital/lib/map_bb_impl.h
+++ b/gr-digital/lib/map_bb_impl.h
@@ -32,7 +32,8 @@ namespace digital {
 class map_bb_impl : public map_bb
 {
 private:
-    unsigned char d_map[0x100];
+    static constexpr size_t s_map_size = 0x100;
+    unsigned char d_map[s_map_size];
     gr::thread::mutex d_mutex;
 
 public:
--- a/gr-digital/lib/map_bb_impl.h
+++ b/gr-digital/lib/map_bb_impl.h
@@ -34,7 +34,7 @@ class map_bb_impl : public map_bb
 private:
     static constexpr size_t s_map_size = 0x100;
     unsigned char d_map[s_map_size];
-    gr::thread::mutex d_mutex;
+    mutable gr::thread::mutex d_mutex;
 
 public:
     map_bb_impl(const std::vector<int>& map);


ThomasHabets:
This block maps bytes, so I'm guessing this is to detect if a user intends to only map a few values, but more values than that go through?

I'm actually a bit confused on why `map()` returns a vector of int instead of unsigned char.

The current code defaults to identity mapping if set_map is given something < 256 in length. Is that not the intention?

michaelld:
Oh ... OK ... you're totally correct there ... my bad! I though this was a general mapping block `*int_*`) ... but yes I see now it's `_bb` ... sorry about that!

michaelld:
So ignore my query about `std::vector` ... that's really not useful, is it!!

michaelld:
let me re-review based on this new understanding (my bad again!) ..

gr-digital/lib/map_bb_impl.cc
```diff
@@ -49,18 +49,19 @@ void map_bb_impl::set_map(const std::vector<int>& map)
...
+    const unsigned int size = std::min(d_map_size, map.size());
     for (unsigned int i = 0; i < size; i++)
         d_map[i] = map[i];
 }
 
 std::vector<int> map_bb_impl::map() const
 {
-    std::vector<int> m;
-    for (unsigned i = 0; i < 0x100; i++)
+    gr::thread::scoped_lock guard(d_mutex);
```
michaelld: Technically I don't think this guard is required, since it's not changing anything internal to the class. That said, I prefer using it in places like this to guarantee that the output result is valid, since it is technically possible that one thread could be doing `::map` while another is doing ``::set_map` ... and this protects against that possibility (not matter how remote it is).

gr-digital/lib/map_bb_impl.cc
```diff
@@ -49,18 +49,19 @@ void map_bb_impl::set_map(const std::vector<int>& map)
 {
     gr::thread::scoped_lock guard(d_mutex);
 
-    for (int i = 0; i < 0x100; i++)
+    for (int i = 0; i < d_map_size; i++)
```
michaelld: I have no idea why this is set this way. I guess it's just to make sure there's -some- value set for every entry? I'd say to leave this alone, since it's worked this way all along ...

michaelld:
This is all good. Let's get another set of eyes on it.

gr-digital/lib/map_bb_impl.h
```diff
@@ -32,7 +32,8 @@ namespace digital {
 class map_bb_impl : public map_bb
 {
 private:
-    unsigned char d_map[0x100];
+    static constexpr size_t d_map_size = 0x100;
```
mbr0wn: I often get linker errors with static constexpr class members, but I can't quite remember when and how.

mbr0wn:
```suggestion
    static constexpr size_t s_map_size = 0x100;
```

gr-digital/lib/map_bb_impl.cc
```diff
@@ -49,18 +49,19 @@ void map_bb_impl::set_map(const std::vector<int>& map)
...
     for (unsigned int i = 0; i < size; i++)
         d_map[i] = map[i];
 }
 
 std::vector<int> map_bb_impl::map() const
 {
-    std::vector<int> m;
-    for (unsigned i = 0; i < 0x100; i++)
+    gr::thread::scoped_lock guard(d_mutex);
+    std::vector<int> m(d_map_size);
```
mbr0wn: Good catch thanks.

gr-digital/lib/map_bb_impl.h
ThomasHabets: Googling around it looks like you've maybe encountered [this thing that's been fixed in C++17 (according to one of the answers)](https://stackoverflow.com/questions/8452952/c-linker-error-with-class-static-constexpr).

Shouldn't affect this PR.

ThomasHabets: I'll push an updated commit with all instances fixed, shortly.

ThomasHabets:
Sorry for exposing brief dirty state. It's because `git rebase` doesn't take `--no-gpg-sign`.

mbr0wn:
You also don't have to sign your commits. Signatures may get lost anyway when we merge/rebase.

ThomasHabets:
There, all cleaned up.

No, the problem was that I was working ssh'd into a machine, and git was absolutely refusing to not *try* to sign, but it put up a prompt on X… on my desktop at work where I'm not there to click.

Anyway, should be two clean commits here now.

gr-digital/lib/map_bb_impl.cc
```diff
@@ -49,18 +49,19 @@ void map_bb_impl::set_map(const std::vector<int>& map)
...
         d_map[i] = map[i];
 }
 
 std::vector<int> map_bb_impl::map() const
 {
-    std::vector<int> m;
-    for (unsigned i = 0; i < 0x100; i++)
+    std::vector<int> m(s_map_size);
+    gr::thread::scoped_lock guard(d_mutex);
+    for (unsigned i = 0; i < s_map_size; i++)
```
marcusmueller: please add  {} while you're at it.

gr-digital/lib/map_bb_impl.h
```diff
@@ -32,8 +32,9 @@ namespace digital {
 class map_bb_impl : public map_bb
 {
 private:
-    unsigned char d_map[0x100];
-    gr::thread::mutex d_mutex;
+    static constexpr size_t s_map_size = 0x100;
+    unsigned char d_map[s_map_size];
+    mutable gr::thread::mutex d_mutex;
```
marcusmueller: uh, good catch. I'm usually a bit careful when it comes to `mutable`, because especially applied to mutexes that's one great way to hide the fact that something actually might result in a deadlock, but looks appropriate here.

ThomasHabets:
Braces added

willcode:
Was backported to maint-3.8 as part of https://github.com/gnuradio/gnuradio/pull/3248

PR #2972, opened 2019-12-21
Remove newlines from exception messages

ThomasHabets:


ThomasHabets:
This cleanup is for consistency and what one normally does. It's easy to accidentally turn a print with a newline into an exception and forget to remove the newline. And since the codebase had a mix of both I thought it best to switch all to the usual type. It's always easier to add a newline than to strip it.

Code tends to already append a newline. E.g. [this](https://github.com/gnuradio/gnuradio/blob/master/gnuradio-runtime/lib/prefs.cc#L116).


So here's an example output.

## Before

```
  File "[…]/default2.py", line 128, in <module>
    main()
  File "[…]/default2.py", line 106, in main
    tb = top_block_cls()
  File "[…]/default2.py", line 77, in __init__
    self.blocks_file_meta_source_0 = blocks.file_meta_source('/uaoeu', True, False, '')
  File "[…]/opt/gnuradio/lib/python3.7/dist-packages/gnuradio/blocks/blocks_swig0.py", line 1339, in make
    return _blocks_swig0.file_meta_source_make(*args, **kwargs)
RuntimeError: file_meta_source: can't open file


>>> Done (return code 1)
```

(note the *two* blank lines after RuntimeError)

## After

```
  File "[…]/default2.py", line 128, in <module>
    main()
  File "[…]/default2.py", line 106, in main
    tb = top_block_cls()
  File "[…]/default2.py", line 77, in __init__
    self.blocks_file_meta_source_0 = blocks.file_meta_source('', True, False, '')
  File "[…]/opt/gnuradio/lib/python3.7/dist-packages/gnuradio/blocks/blocks_swig0.py", line 1339, in make
    return _blocks_swig0.file_meta_source_make(*args, **kwargs)
RuntimeError: file_meta_source: can't open file

>>> Done (return code 1)
```

michaelld:
Thanks for the debug printout info. I have no issues with this change. Let's get 1 more set of eyes on it ...

PR #2974, opened 2019-12-23
Switch from boost pointers to std C++11 pointers

ThomasHabets:
Most of this code is automated code changes:

```
SUB="s/dummy/dummy/"
for i in shared_ptr make_shared dynamic_pointer_cast weak_ptr enable_shared_from_this get_deleter; do
    SUB="$SUB;s/boost::$i/std::$i/g"
done
SUB="$SUB;s^#include <boost/shared_ptr.hpp>^#include <memory>^g"
SUB="$SUB;s^namespace boost^namespace std^g"

find . \( -name "*.cc" -o -name "*.h" -o -name "*.i" -o -name "*.cxx" -o -name "*.py" \) -print0 | xargs -0 sed -i "$SUB"
```

Only one manual change. In `./gr-fec/lib/fec_mtrx_impl.cc`, add
`#include <algorithm>`.

### DIFF
--- a/README.hacking
+++ b/README.hacking
@@ -105,14 +105,14 @@ Each significant class shall be contained in its own file.
 
 * Storage management
 
-Strongly consider using the boost smart pointer templates, scoped_ptr
-and shared_ptr.  scoped_ptr should be used for locals that contain
+Strongly consider using the std smart pointer templates, unique_ptr
+and shared_ptr.  unique_ptr should be used for locals that contain
 pointers to objects that we need to delete when we exit the current
 scope.  shared_ptr implements transparent reference counting and is a
 major win.  You never have to worry about calling delete.  The right
 thing happens.
 
-See http://www.boost.org/libs/smart_ptr/smart_ptr.htm
+See https://en.cppreference.com/book/intro/smart_pointers
 
 
 * Unit tests
--- a/docs/doxygen/other/shared_ptr_docstub.h
+++ b/docs/doxygen/other/shared_ptr_docstub.h
@@ -1,4 +1,4 @@
-namespace boost {
+namespace std {
 /*!
  * \brief shared_ptr documentation stub
  *
@@ -20,4 +20,4 @@ public:
 }; // shared_ptr
 
 
-} // namespace boost
+} // namespace std
--- a/gnuradio-runtime/include/gnuradio/basic_block.h
+++ b/gnuradio-runtime/include/gnuradio/basic_block.h
@@ -54,7 +54,7 @@ namespace gr {
  * processing functions.
  */
 class GR_RUNTIME_API basic_block : public msg_accepter,
-                                   public boost::enable_shared_from_this<basic_block>
+                                   public std::enable_shared_from_this<basic_block>
 {
     typedef boost::function<void(pmt::pmt_t)> msg_handler_t;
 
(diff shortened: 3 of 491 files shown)


ThomasHabets:
Maybe this pull request should only be merged if also 3.9 (and even this PR?) moves away what it can from boost.

Obviously pointer changes is a breaking API change. Depends on how much of other boost changes are also API breaking, and how wide the breakage is.

ThomasHabets:
Actually because of the wide spread use of `_sptr` typedefs in gnuradio, is this even an API-breaking change? (as opposed to ABI-breaking)

ThomasHabets:
(1) I don't think any dependency can be removed from this, unfortunately. Smart pointers are template magic in headers.

I'm not a Boost expert, but of the remaining listed components I'd say this:

* date_time — I've not looked into what this contains.
* program_options — No alternative in std C++ AFAIK. Sure, there's `getopt`, but that's not new so I assume there's good reason it's not already being used.
* filesystem — [Not until C++17](https://en.cppreference.com/w/cpp/filesystem)
* regex — Likely [yes](https://en.cppreference.com/w/cpp/regex/basic_regex)
* system — I've not looked into what this contains.
* thread — I don't see why not. Maybe some small workarounds until C++20 in case we use [jthread](https://www.modernescpp.com/index.php/a-new-thread-with-c-20-std-jthread) or equivalent in boost.

(2) tnt & noc0lour on IRC made two lists based on (presumably) grepping:
* http://dpaste.com/3M1W4YK
* https://pastebin.com/U97T0ekz

ThomasHabets:
Actually, I just built and did `make test` with the component `system` removed. Maybe it's not needed anymore?

But this PR doesn't change that. Current master branch also builds fine on my system without `system` listed.

Maybe I did it wrong though. I'm also not a cmake expert. Could be that other versions of boost does require it to be in the list, but the one I have installed (1.67) does not.

michaelld:
Thanks for continuing to poke at this! GR is pretty old-school Boost usage; I strongly doubt you'll find Boost usage more modern than 1.65 ... excepting where the old usage was actually removed & we require more the new API (whatever that might be). Most of older Boost has been integrated into C++11 or more recent ... which means we can probably swap `boost::` for `std::` in a lot more places. Just guessing.

michaelld:
As for the Boost usage list ... thinking we _can't_ really get rid of `program_options` ... we can make `filesystem` like what we did in Volk < https://github.com/gnuradio/volk/pull/255 >, which mostly means not getting rid of it ... but I believe all of the rest are in C++11, just a namespace swap like what you did in this PR (certainly -most-, but I think -all-).

mormj:
Love this!!! Nice work

michaelld:
CI errors: Need to fixup Thrift / CTRLPORT to work with this change!

michaelld:
I'd be happy to fixup CTRLPORT after merging this. There's a PR for this change already.

michaelld:
@marcusmueller can you review & merge this? I'll fixup the issues later.

mbr0wn:
Another comment on big changes like this: One downside is our potential inability to easily backport fixes from master to maint-3.8.

michaelld:
> Another comment on big changes like this: One downside is our potential inability to easily backport fixes from master to maint-3.8.

@mbr0wn My preference would be for this to be just master. It's a HUGE change, and we can't know for certain that it doesn't totally bork the ABI especially for OOTs. Hence, IMHO, it's a 3.9 only change. I'm open to being convinced otherwise, and obviously the buck doesn't stop with me :)

michaelld:
> Boost smart pointers and std smart pointers are not compatible -- we need to make sure this doesn't break any SWIG, nor any OOT compatibility.

Hence why a GR39 change only ... we can break things there if it makes sense ... we really can't afford to break anything more in GR38 ... we need to be spending time fixing GR38 to work better overall!

michaelld:
@ThomasHabets would you please rebase to fix conflicts?

ThomasHabets:
@michaelld Rebase uploaded.

mbr0wn:
@ThomasHabets and I have communicated outside of Github; we'll close this PR for now and come up with a plan to do this safely and effectively. For anyone else reading this: We, the GNU Radio devs, want all the C++ modernizing we can get, so this is not a rejection. We just don't want to have massive open PRs.

PR #2976, opened 2019-12-23
Move locks from boost to C++11

ThomasHabets:
This one includes a bit more manual fixups, to clarify which type of lock is actually being used.

* `lock_guard`: Simple RAII lock
* `scoped_lock`: RAII lock that can lock multiple locks (C++17, and not needed in codebase)
* `unique_lock`: RAII lock, plus ability to unlock. Needed for `condition_variable`.

I kept the `gr::thread::` typedefs, since at least with C++11 it's a bit verbose to type without the template type inference that's newer than C++11.

This PR has only been minimally tested.

Note that this also fixes a bug, where a lock is not actually acquired: https://github.com/gnuradio/gnuradio/blob/master/gr-audio/lib/portaudio/portaudio_source.cc#L109

### DIFF
--- a/README.hacking
+++ b/README.hacking
@@ -41,6 +41,21 @@ Boost 1.35 and later is fairly common in modern distributions.
 If it isn't available for your system, please refer to
 README.building-boost for instructions
 
+As features from Boost are incorporated into standard C++ GNU Radio
+will reduce dependence on Boost.
+
+Status:
+  * program_options — Will remain
+  * shared_ptr      — PR in flight
+  * shared_mutex    — C++17
+  * filesystem      — C++17
+  * thread          — Unknown
+  * date_time       — Unknown
+  * regex           — Unknown
+  * system          — Unknown
+  * chrono          — Unknown
+  * atomic          — Unknown
+
 * C++ and Python
 
 GNU Radio is a hybrid system.  Some parts of the system are built in C++ and
--- a/gnuradio-runtime/lib/vmcircbuf.h
+++ b/gnuradio-runtime/lib/vmcircbuf.h
@@ -27,7 +27,7 @@
 #include <gnuradio/thread/thread.h>
 #include <vector>
 
-extern gr::thread::mutex s_vm_mutex;
+extern std::mutex s_vm_mutex;
 
 namespace gr {
 
--- a/gr-audio/lib/portaudio/portaudio_source.h
+++ b/gr-audio/lib/portaudio/portaudio_source.h
@@ -27,6 +27,8 @@
 #include <gnuradio/buffer.h>
 #include <gnuradio/thread/thread.h>
 #include <portaudio.h>
+#include <mutex>
+#include <condition_variable>
 #include <stdexcept>
 #include <string>
 
@@ -58,8 +60,8 @@ class portaudio_source : public source
     gr::buffer_sptr d_writer; // buffer used between work and callback
     gr::buffer_reader_sptr d_reader;
 
-    gr::thread::mutex d_ringbuffer_mutex;
-    gr::thread::condition_variable d_ringbuffer_cond;
+    std::mutex d_ringbuffer_mutex;
+    std::condition_variable d_ringbuffer_cond;
     bool d_ringbuffer_ready;
 
     // random stats
--- a/gr-fft/lib/ctrlport_probe_psd_impl.cc
+++ b/gr-fft/lib/ctrlport_probe_psd_impl.cc
@@ -62,8 +62,8 @@ void ctrlport_probe_psd_impl::forecast(int noutput_items,
 }
 
 //    boost::shared_mutex mutex_buffer;
-//    mutable boost::mutex mutex_notify;
-//    boost::condition_variable condition_buffer_ready;
+//    mutable std::mutex mutex_notify;
+//    std::condition_variable condition_buffer_ready;
 std::vector<gr_complex> ctrlport_probe_psd_impl::get()
 {
     mutex_buffer.lock();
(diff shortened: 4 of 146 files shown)


ThomasHabets:
(I expect a little bit of merge conflict when https://github.com/gnuradio/gnuradio/pull/2974 is merged, but I'll take care of it)

README.hacking
```diff
@@ -41,6 +41,21 @@ Boost 1.35 and later is fairly common in modern distributions.
 If it isn't available for your system, please refer to
 README.building-boost for instructions
 
+As features from Boost are incorporated into standard C++ GNU Radio
+will reduce dependence on Boost.
```
mbr0wn: Yes please.

mbr0wn:
I like this direction. However, we need some input from @marcusmueller if this is OK.
With hundreds of lines changed across modules, it's easy to miss a subtle bug and almost impossible to find those with code reviews. We'll need to first make sure that all tests pass (they currently don't).

mbr0wn:
Because of the risk/impact quotient (it's a bit high), this will take a bit before we can merge. I suggest  we focus on one C++ improvment PR at a time.

marcusmueller:
So, first of all: Yeah, using `std::` instead of `gr::thread::foobar` wherever possible was on our to-do list anyway, so thanks!

Then, yes, this kind of is a major API change ->  no backporting to 3.8 (which doesn't hurt me).

gnuradio-runtime/lib/vmcircbuf.h
```diff
@@ -27,7 +27,7 @@
 #include <gnuradio/thread/thread.h>
 #include <vector>
 
-extern gr::thread::mutex s_vm_mutex;
+extern std::mutex s_vm_mutex;
```
marcusmueller: Had forgotten about that surprising `extern`; nothing to do with your work here, but let me / us investigate the design here in #2996.

gr-audio/lib/portaudio/portaudio_source.h
```diff
@@ -27,6 +27,8 @@
 #include <gnuradio/buffer.h>
 #include <gnuradio/thread/thread.h>
 #include <portaudio.h>
+#include <mutex>
```
marcusmueller: Nice, IWYU!

gr-fft/lib/ctrlport_probe_psd_impl.cc
```diff
@@ -62,16 +62,16 @@ void ctrlport_probe_psd_impl::forecast(int noutput_items,
 }
 
 //    boost::shared_mutex mutex_buffer;
```
marcusmueller: Can you clean up this dead/commented code?

gr-qtgui/lib/SpectrumGUIClass.cc
```diff
@@ -227,7 +227,7 @@ void SpectrumGUIClass::updateWindow(const bool updateDisplayFlag,
                                     const gr::high_res_timer_type timestamp,
                                     const bool lastOfMultipleFFTUpdateFlag)
 {
-    // gr::thread::scoped_lock lock(d_mutex);
```
marcusmueller: This feels like a red flag...

marcusmueller: but has nothing to do with this change set.

marcusmueller:
So, really, rebase this / resolve the conflicts. From my side, this code looks fine.

marcusmueller:
(and I'd like to run exactly the same version as you to understand the unit test failures, so I'll have to leave it up to you to do the rebase)

ThomasHabets:
Curious. I don't know what's up with the tests. Pushed a rebase now, but it's late so I'll check more tomorrow.

mbr0wn:
The test failure looks like a deadlock.

ThomasHabets:
Yeah. This is due to Boost thread cancellation, which won't be in standard C++ until C++20, and even then it's [not exception-based](https://en.cppreference.com/w/cpp/thread). :-(

I can fix the test by changing `tpb_thread_body.cc` to:

```c++
 case block_executor::BLKD_OUT: // Wait for output buffer space.                                                                                       
        {
            gr::thread::unique_lock guard(d->d_tpb.mutex);
            while (!d->d_tpb.output_changed) {
                d->d_tpb.output_cond.wait_for(guard, std::chrono::milliseconds(250));
                boost::this_thread::interruption_point();
            }
        } break;
```

But that's just this place. Also needlessly polls every 250ms.

There aren't many places where threads are interrupted:

```
$ grep -r 'interrupt()' .
./gr-blocks/lib/socket_pdu_impl.cc:        d_thread.interrupt();
./gr-blocks/lib/message_strobe_impl.cc:    d_thread->interrupt();
./gr-blocks/lib/stream_pdu_base.cc:        d_thread.interrupt();
./gr-blocks/lib/message_strobe_random_impl.cc:    d_thread->interrupt();
./gnuradio-runtime/lib/thread/thread_group.cc:        (*it)->interrupt();
./gnuradio-runtime/lib/logger.cc:        instance.watch_thread->interrupt();
./gnuradio-runtime/lib/controlport/thrift/thrift-codebase-shutdown-patch.diff:+    serverTransport_->interrupt();
./gnuradio-runtime/lib/controlport/thrift/thrift-codebase-shutdown-patch.diff:-    serverTransport_->interrupt();
```

but every single condition_variable (though not mere mutex lock) is a [trigger point](https://www.boost.org/doc/libs/1_67_0/doc/html/thread/thread_management.html#thread.thread_management.tutorial.interruption.predefined_interruption_points).

What do you think is best? Drop the whole thing, or scale back leaving boost mutexes when they're used for condition variables?

mbr0wn:
We had the same issue in UHD, and we first removed all the thread interruptions. In most cases, they were a lazy way of getting out of threads anyway. I would suggest doing the same -- and do it separately.

mbr0wn:
@ThomasHabets and I have communicated outside of Github; we'll close this PR for now and come up with a plan to do this safely and effectively. For anyone else reading this: We, the GNU Radio devs, want all the C++ modernizing we can get, so this is not a rejection. We just don't want to have massive open PRs.

PR #2977, opened 2019-12-23
gr-audio/portaudio_source: Fix lock acquisition

ThomasHabets:
This is also part of https://github.com/gnuradio/gnuradio/pull/2976, but maybe cleaner to merge this PR on its own, as it's easier to review.


mbr0wn:
Thanks for splitting this out. This is trivial enough so we can merge even if you don't have a CLA.

PR #2893, opened 2019-11-08
Use C++11 for range loops, and some more const

ThomasHabets:

PR #2978, opened 2019-12-24
gr-filter/fft_filter: Use smart instead of raw pointers

ThomasHabets:
This reduces amount of code, and risk of dangling pointers and memory leaks.

I thought I'd do this one to get opinions. It takes a while (especially with custom deleter), so don't expect me to do all of it at least at this point. :-)

### DIFF
--- a/gr-filter/include/gnuradio/filter/fft_filter.h
+++ b/gr-filter/include/gnuradio/filter/fft_filter.h
@@ -26,6 +26,7 @@
 #include <gnuradio/fft/fft.h>
 #include <gnuradio/filter/api.h>
 #include <gnuradio/gr_complex.h>
+#include <volk/volk_alloc.hh>
 #include <vector>
 
 namespace gr {
@@ -70,13 +71,13 @@ private:
     int d_ntaps;
     int d_nsamples;
     int d_fftsize; // fftsize = ntaps + nsamples - 1
-    int d_decimation;
-    fft::fft_real_fwd* d_fwdfft; // forward "plan"
-    fft::fft_real_rev* d_invfft; // inverse "plan"
+    const int d_decimation;
+    std::unique_ptr<fft::fft_real_fwd> d_fwdfft; // forward "plan"
+    std::unique_ptr<fft::fft_real_rev> d_invfft; // inverse "plan"
     int d_nthreads;              // number of FFTW threads to use
     std::vector<float> d_tail;   // state carried between blocks for overlap-add
     std::vector<float> d_taps;   // stores time domain taps
-    gr_complex* d_xformed_taps;  // Fourier xformed taps
+    volk::vector<gr_complex> d_xformed_taps;  // Fourier xformed taps
 
     void compute_sizes(int ntaps);
     int tailsize() const { return d_ntaps - 1; }
@@ -94,8 +95,6 @@ public:
      * \param nthreads   The number of threads for the FFT to use (int)
      */
     fft_filter_fff(int decimation, const std::vector<float>& taps, int nthreads = 1);
-
-    ~fft_filter_fff();
 
     /*!
      * \brief Set new taps for the filter.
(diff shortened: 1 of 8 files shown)


hcab14:
Looks good. Just to let you know that there is a recent addition to volk
called volk::vector which helps to simplify the fft code.

ThomasHabets:
Thanks for the pointer. Looks like I'll have to wait for gnuradio to pick up https://github.com/gnuradio/volk/commit/4cc47ea71b4c7d3f9433156f5f61b6c7e11b4f53 though.

ThomasHabets:
Ack. I'm holding off on this until https://github.com/gnuradio/gnuradio/pull/2981 is merged.

marcusmueller:
I mean, this is seriously an improvement. However, I'm still not convinced the fftw plan has to be handled via pointer at all – it's not like anything else would be using it. 

@ThomasHabets I'm certain you're more intimate with the details of the FFT handling by now, but what's the advantage of having a unique pointer instead of holding the raw object?

mbr0wn:
Going pointer -> smart pointer is something I would consider refactoring, and going pointer -> no pointer is something I would consider a code change. If we want the latter, the former is no longer necessary. However, the refactoring by itself makes a lot of sense.

gr-filter/lib/fft_filter.cc
```diff
@@ -24,6 +24,7 @@
 #include "config.h"
 #endif
 
+#include <boost/smart_ptr/make_unique.hpp>
```
mbr0wn: Meh. I wonder if we should go straight to C++14, or create our own `std::make_unique` for C++11 as a stop-gap.

mbr0wn:
Personally, I think I prefer the pointers though.

mbr0wn:
But please cut the commit subject to 72 characters.

ThomasHabets:
@marcusmueller looks to me like the plans are not copyable (because in their default the member pointers will be copied by value). It's actually another cleanup I've had in mind. It becomes easier with (again) C++11 and move semantics to fix these.

Currently `fft::fft_real_fwd` and friends appear to be a bit of a minefield, since they don't prohibit copy assignment or copy constructor at compile time, but they leave dangling pointers if called at runtime. (unless I'm missing something)

So yeah, implementing move semantics on the ffts, disabling copy assignment and copy constructor, and de-pointering `fft_filter_*` seems like a good next step.

De-pointering requires some further rework of `fft_filter_*` though: The fft size is not known at construction time, so would need to be broken out into its own function so that it can be provided on the constructor initializer list. Not hard. it's a direct function of the tap count.

Any preferences? In the mean time I'll sort out the volk vector thing. Now that the other pull request is in it seems like that one can go in this commit, right?

mbr0wn:
I think it's not worth de-pointerizing this. Your commit seems to take care of @marcusmueller's question and I like this PR.

mbr0wn:
I also don't think the PR failure is your fault.

mbr0wn:
@ThomasHabets do we have a  CLA from you?

ThomasHabets:
@mbr0wn yup, CLA sorted during https://github.com/gnuradio/gnuradio/pull/2893. Google corp CLA.

ThomasHabets:
There. Now using `volk::vector`. Should be ready to go.

gr-filter/include/gnuradio/filter/fft_filter.h
```diff
@@ -174,13 +173,13 @@ class FILTER_API fft_filter_ccc
     int d_ntaps;
     int d_nsamples;
     int d_fftsize; // fftsize = ntaps + nsamples - 1
-    int d_decimation;
-    fft::fft_complex* d_fwdfft;     // forward "plan"
-    fft::fft_complex* d_invfft;     // inverse "plan"
+    const int d_decimation;
```
marcusmueller: We coulda/shoulda `unsigned` this, but for now, come on, let's get this merged.

marcusmueller:
Merging as soon as CI passes :)

marcusmueller:
Thomas, I have yet to express my gratitude: Thanks for all these contributions!

gr-filter/include/gnuradio/filter/fft_filter.h
```diff
@@ -26,6 +26,7 @@
 #include <gnuradio/fft/fft.h>
 #include <gnuradio/filter/api.h>
 #include <gnuradio/gr_complex.h>
+#include <volk/volk_alloc.hh>
```
marcusmueller: How did CI not catch this? There's an `h` to much...

marcusmueller:
@noc0lour  just a heads up: buildbot somehow missed the above typo, which definitely should have killed the build, somehow. I don't understand how that happens?

gr-filter/include/gnuradio/filter/fft_filter.h
ThomasHabets: Is it? It's not just the indicator saying it's a C++-only header?

ThomasHabets: Ah, I see you reverted that commit now.

PR #2981, opened 2019-12-25
VOLK: update submodule pointer to v2.1.0

marcusmueller:
This is a pseudo-PR to let CI run with the updated VOLK.


michaelld:
The new Volk submodule is indeed included in the CI, and all of the CI passes, so IMHO all is good to go here.

michaelld:
Adding a couple other GR devs to re-verify that all is good. Please go ahead and merge once verified.

PR #2983, opened 2019-12-27
Fix parameter hiding on symbol sync block YAML

igorauad:
Fix the options that are used in order to hide the "Filterbank Arms" and "PFB MF
Taps" parameters of the symbol sync block (digital_symbol_sync_xx), depending on
the chosen interpolating resampler.

Before this, if *Interpolating Resampler* type was set for instance to *Polyphase Filterbank, MF* on the Symbol Sync block, that is, parameter `resamp_type` set to `digital.IR_PFB_MF`, gnuradio companion would display:

symbol_sync_xx_err

I saw that the original line came from auto-generated YAML on baf7eaf8f29d5a490f2580917362cf5b3db47281, so I assumed it was a mistake.


mbr0wn:
Merged manually (needed to fix commit message, too many chars/line).

mbr0wn:
Thanks for the fix!

PR #2984, opened 2019-12-28
Add msg port to control rotator's phase increment

igorauad:
Often receivers estimate the frequency offset on a block that is downstream with
respect to the frequency correction NCO. In such architectures, typically the
frequency offset estimator feeds the estimation back to the NCO.

This patch adds a message port to the rotator block that makes this architecture
feasible using the rotator as NCO. With that, a downstream block can estimate
frequency offset and send a message to the rotator in order to update its
rotating frequency (i.e. its phase increment).

The input message port receives a PMT pair containing both the new desired phase
increment as well as the absolute index of the rotator's output sample where
this new phase increment should start to take effect.

The rationale of scheduling the phase increment update to a specific absolute
index is as follows. Usually the receiver will rely on a frame structure and
will estimate the frequency offset based on pilots along the frame. If the true
frequency offset changes in the middle of the frame, the frame-averaged
frequency offset estimation likely becomes corrupted. It is often better to
update the frequency correction right at the start of a new frame. This way, the
true frequency offset is preserved in the course of a frame and only changes on
the boundary between consecutive frames.

For flexibility when using this feature, it is responsibility of the downstream
block (more generally the message sender) to schedule the update properly, based
on absolute index. Thus, for example, if there is a decimator in between the
rotator and the message sender, the sender will need to take the sample rate
conversion factor into account when computing the absolute index of the update.

### DIFF
--- a/gr-blocks/include/gnuradio/blocks/rotator_cc.h
+++ b/gr-blocks/include/gnuradio/blocks/rotator_cc.h
@@ -20,6 +20,23 @@ namespace blocks {
 /*!
  * \brief Complex rotator
  * \ingroup math_operators_blk
+ *
+ * \details
+ *
+ * Rotates an input complex sequence using a complex exponential in the form of
+ * exp(1j * phase_inc * n), where "phase_inc" is a chosen phase increment in
+ * radians and "n" is the sample index.
+ *
+ * Message Ports:
+ *
+ * - cmd (input):
+ *    Receives a PMT dictionary with a command message to set a new phase
+ *    increment on the rotator at a specified sample offset. The new increment
+ *    must be provided as a PMT double on a key named "inc". The target sample
+ *    offset on which to update the phase increment must be given as a PMT
+ *    uint64 (with the absolute output item number) on a key named
+ *    "offset". Unlike the "inc" key, the "offset" key is optional. When not
+ *    provided, the rotator updates its phase increment immediately.
  */
 class BLOCKS_API rotator_cc : virtual public sync_block
 {
@@ -28,10 +45,13 @@ public:
     typedef std::shared_ptr<rotator_cc> sptr;
 
     /*!
-     * \brief Make an complex rotator block
+     * \brief Make a complex rotator block
      * \param phase_inc rotational velocity
+     * \param tag_inc_updates Tag the sample where a phase increment update is
+     *                        applied following the reception of a control
+     *                        message received via the input message port.
      */
-    static sptr make(double phase_inc = 0.0);
+    static sptr make(double phase_inc = 0.0, bool tag_inc_updates = false);
 
     virtual void set_phase_inc(double phase_inc) = 0;
 };
--- a/gr-blocks/lib/rotator_cc_impl.cc
+++ b/gr-blocks/lib/rotator_cc_impl.cc
@@ -20,17 +20,23 @@
 namespace gr {
 namespace blocks {
 
-rotator_cc::sptr rotator_cc::make(double phase_inc)
+rotator_cc::sptr rotator_cc::make(double phase_inc, bool tag_inc_updates)
 {
-    return gnuradio::make_block_sptr<rotator_cc_impl>(phase_inc);
+    return gnuradio::make_block_sptr<rotator_cc_impl>(phase_inc, tag_inc_updates);
 }
 
-rotator_cc_impl::rotator_cc_impl(double phase_inc)
+rotator_cc_impl::rotator_cc_impl(double phase_inc, bool tag_inc_updates)
     : sync_block("rotator_cc",
                  io_signature::make(1, 1, sizeof(gr_complex)),
-                 io_signature::make(1, 1, sizeof(gr_complex)))
+                 io_signature::make(1, 1, sizeof(gr_complex))),
+      d_tag_inc_updates(tag_inc_updates),
+      d_inc_update_queue(cmp_phase_inc_update_offset)
 {
     set_phase_inc(phase_inc);
+
+    const pmt::pmt_t port_id = pmt::mp("cmd");
+    message_port_register_in(port_id);
+    set_msg_handler(port_id, [this](pmt::pmt_t msg) { this->handle_cmd_msg(msg); });
 }
 
 rotator_cc_impl::~rotator_cc_impl() {}
@@ -40,10 +46,43 @@ void rotator_cc_impl::set_phase_inc(double phase_inc)
     d_r.set_phase_incr(exp(gr_complex(0, phase_inc)));
 }
 
+void rotator_cc_impl::handle_cmd_msg(pmt::pmt_t msg)
+{
+    gr::thread::scoped_lock l(d_mutex);
+
+    static const pmt::pmt_t inc_key = pmt::intern("inc");
+    static const pmt::pmt_t offset_key = pmt::intern("offset");
+
+    if (!pmt::is_dict(msg)) {
+        throw std::runtime_error("rotator_cc: Rotator command message "
+                                 "must be a PMT dictionary");
+    }
+
+    bool handled = false;
+
+    if (pmt::dict_has_key(msg, inc_key)) {
+        /* Prepare to update the phase increment on a specific absolute sample
+         * offset. If the offset is not defined in the command message,
+         * configure the phase increment update to occur immediately. */
+        phase_inc_update_t update{};
+        update.phase_inc = pmt::to_double(pmt::dict_ref(msg, inc_key, pmt::PMT_NIL));
+        update.offset = pmt::dict_has_key(msg, offset_key)
+                            ? pmt::to_uint64(pmt::dict_ref(msg, offset_key, pmt::PMT_NIL))
+                            : nitems_written(0);
+        d_inc_update_queue.push(update);
+        handled = true;
+    }
+
+    if (!handled) {
+        throw std::runtime_error("rotator_cc: Unsupported command message");
+    }
+}
+
 int rotator_cc_impl::work(int noutput_items,
                           gr_vector_const_void_star& input_items,
                           gr_vector_void_star& output_items)
 {
+    gr::thread::scoped_lock l(d_mutex);
     const gr_complex* in = (const gr_complex*)input_items[0];
     gr_complex* out = (gr_complex*)output_items[0];
 
@@ -51,7 +90,50 @@ int rotator_cc_impl::work(int noutput_items,
       for (int i=0; i<noutput_items; i++)
       	out[i] = d_r.rotate(in[i]);
 #else
-    d_r.rotateN(out, in, noutput_items);
+
+    const uint64_t n_written = nitems_written(0);
+
+    if (d_inc_update_queue.empty()) {
+        d_r.rotateN(out, in, noutput_items);
+    } else {
+        /* If there are phase increment updates scheduled for now, handle the
+         * rotation in steps and update the phase increment in between. */
+        int nprocessed_items = 0;
+        while (!d_inc_update_queue.empty()) {
+            auto next_update = d_inc_update_queue.top();
+
+            if (next_update.offset < (n_written + nprocessed_items)) {
+                d_inc_update_queue.pop();
+                continue; // we didn't process this update on time - drop it
+            }
+
+            if (next_update.offset >= (n_written + noutput_items)) {
+                break; // the update is for a future batch of samples
+            }
+
+            // The update is scheduled for this batch of samples. Apply it now.
+            d_inc_update_queue.pop();
+
+            // Process all samples until the scheduled phase increment update
+            int items_before_update = next_update.offset - n_written - nprocessed_items;
+            d_r.rotateN(out, in, items_before_update);
+            nprocessed_items += items_before_update;
+
+            set_phase_inc(next_update.phase_inc);
+
+            if (d_tag_inc_updates) {
+                add_item_tag(0,
+                             next_update.offset,
+                             pmt::string_to_symbol("rot_phase_inc"),
+                             pmt::from_float(next_update.phase_inc));
+            }
+        }
+
+        // Rotate the remaining samples
+        d_r.rotateN(out + nprocessed_items,
+                    in + nprocessed_items,
+                    (noutput_items - nprocessed_items));
+    }
 #endif
 
     return noutput_items;
--- a/gr-blocks/lib/rotator_cc_impl.cc
+++ b/gr-blocks/lib/rotator_cc_impl.cc
@@ -85,12 +85,6 @@ int rotator_cc_impl::work(int noutput_items,
     gr::thread::scoped_lock l(d_mutex);
     const gr_complex* in = (const gr_complex*)input_items[0];
     gr_complex* out = (gr_complex*)output_items[0];
-
-#if 0
-      for (int i=0; i<noutput_items; i++)
-      	out[i] = d_r.rotate(in[i]);
-#else
-
     const uint64_t n_written = nitems_written(0);
 
     if (d_inc_update_queue.empty()) {
@@ -134,7 +128,6 @@ int rotator_cc_impl::work(int noutput_items,
                     in + nprocessed_items,
                     (noutput_items - nprocessed_items));
     }
-#endif
 
     return noutput_items;
 }
--- a/gr-blocks/lib/rotator_cc_impl.cc
+++ b/gr-blocks/lib/rotator_cc_impl.cc
@@ -86,6 +86,7 @@ int rotator_cc_impl::work(int noutput_items,
     const gr_complex* in = (const gr_complex*)input_items[0];
     gr_complex* out = (gr_complex*)output_items[0];
     const uint64_t n_written = nitems_written(0);
+    static const pmt::pmt_t phase_inc_tag_key = pmt::intern("rot_phase_inc");
 
     if (d_inc_update_queue.empty()) {
         d_r.rotateN(out, in, noutput_items);
@@ -118,7 +119,7 @@ int rotator_cc_impl::work(int noutput_items,
             if (d_tag_inc_updates) {
                 add_item_tag(0,
                              next_update.offset,
-                             pmt::string_to_symbol("rot_phase_inc"),
+                             phase_inc_tag_key,
                              pmt::from_float(next_update.phase_inc));
             }
         }
--- a/gr-blocks/python/blocks/qa_rotator_cc.py
+++ b/gr-blocks/python/blocks/qa_rotator_cc.py
@@ -1,0 +1,322 @@
+#!/usr/bin/env python
+#
+# Copyright 2021 Free Software Foundation, Inc.
+#
+# This file is part of GNU Radio
+#
+# SPDX-License-Identifier: GPL-3.0-or-later
+#
+#
+
+from gnuradio import gr, gr_unittest
+from gnuradio import blocks
+import numpy as np
+from numpy.random import uniform
+import pmt
+
+
+class qa_rotator_cc(gr_unittest.TestCase):
+    def _setUp(self, n_samples=100, tag_inc_updates=True):
+        """Base fixture: set up flowgraph and parameters"""
+        self.n_samples = n_samples  # number of IQ samples to generate
+        self.f_in = uniform(high=0.5)  # input frequency
+        self.f_shift = uniform(high=0.5) - \
+            self.f_in  # rotator's starting frequency
+
+        # Input IQ samples
+        in_angles = 2 * np.pi * np.arange(self.n_samples) * self.f_in
+        in_samples = np.exp(1j * in_angles)
+
+        # Rotator's starting phase increment
+        phase_inc = 2 * np.pi * self.f_shift
+
+        # Flowgraph:
+        #
+        #                             /-> Vector Sink
+        # Vector Source -> Rotator --/
+        #                            \
+        #                             \-> Tag Debug
+        #
+        self.tb = gr.top_block()
+        self.source = blocks.vector_source_c(in_samples)
+        self.rotator_cc = blocks.rotator_cc(phase_inc, tag_inc_updates)
+        self.sink = blocks.vector_sink_c()
+        self.tag_sink = blocks.tag_debug(gr.sizeof_gr_complex, "rot_phase_inc",
+                                         "rot_phase_inc")
+        self.tag_sink.set_save_all(True)
+        self.tb.connect(self.source, self.rotator_cc, self.sink)
+        self.tb.connect(self.rotator_cc, self.tag_sink)
+
+    def setUp(self):
+        self._setUp()
+
+    def tearDown(self):
+        self.tb = None
+
+    def _post_phase_inc_cmd(self, new_phase_inc, offset=None):
+        """Post phase increment update command to the rotator block"""
+        cmd = pmt.make_dict()
+        cmd = pmt.dict_add(cmd, pmt.intern("inc"),
+                           pmt.from_double(new_phase_inc))
+        if (offset is not None):
+            cmd = pmt.dict_add(cmd, pmt.intern("offset"),
+                               pmt.from_uint64(offset))
+        self.rotator_cc.insert_tail(pmt.to_pmt("cmd"), cmd)
+
+    def _assert_tags(self, expected_values, expected_offsets):
+        """Check the tags received by the tag debug block"""
+        tags = self.tag_sink.current_tags()
+        expected_tags = list(zip(expected_values, expected_offsets))
+        self.assertEqual(len(tags), len(expected_tags))
+
+        for idx, (val, offset) in enumerate(expected_tags):
+            self.assertAlmostEqual(pmt.to_double(tags[idx].value),
+                                   val,
+                                   places=5)
+            self.assertEqual(tags[idx].offset, offset)
+
+    def test_freq_shift(self):
+        """Complex sinusoid frequency shift"""
+        f_out = self.f_in + self.f_shift  # expected output frequency
+
+        # Expected IQ samples
+        expected_angles = 2 * np.pi * np.arange(self.n_samples) * f_out
+        expected_samples = np.exp(1j * expected_angles)
+
+        self.tb.run()
+        self.assertComplexTuplesAlmostEqual(self.sink.data(),
+                                            expected_samples,
+                                            places=4)
+
+    def _test_scheduled_phase_inc_update(self):
+        """Update the phase increment at a chosen offset via command message
+
+        Returns:
+            Tuple with the new phase increment, the phase increment update
+            sample offset, and a list with the expected rotated IQ samples.
+
+        """
+        new_f_shift = uniform(high=0.5) - self.f_in  # rotator's new frequency
+        new_phase_inc = float(2 * np.pi * new_f_shift)  # new phase increment
+        offset = int(self.n_samples / 2)  # when to update the phase increment
+        f_out_1 = self.f_in + self.f_shift  # output frequency before update
+        f_out_2 = self.f_in + new_f_shift  # output frequency after update
+
+        # Post the phase increment command message to the rotator block
+        self._post_phase_inc_cmd(new_phase_inc, offset)
+
+        # Samples before and after the phase increment update
+        n_before = offset
+        n_after = self.n_samples - offset
+
+        # Expected IQ samples
+        #
+        # Note: when the phase increment is updated, the assumption is that the
+        # update occurs at the beginning of the sample period. Hence, the new
+        # phase rotation pace will only be observed in the following sample. In
+        # other words, the new phase increment can only be observed after some
+        # time goes by such that the rotator's phase accumulates.
+        #
+        # For example, suppose there are ten samples, and the update occurs on
+        # offset 5 (i.e., the sixth sample). Suppose also that the initial
+        # phase increment is 0.1, and the new increment is 0.2. In this case,
+        # we have the following sequence of angles:
+        #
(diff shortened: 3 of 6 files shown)


mbr0wn:
@igorauad Do we have a CLA from you? I'm flagging this just in case we don't. If we do, many apologies in advance. I've been away from GR/github a bit so I'm not quite up to date.

What would definitely spice up this PR is a GRC-based example.

gr-blocks/include/gnuradio/blocks/rotator_cc.h
```diff
@@ -32,6 +32,18 @@ namespace blocks {
...
  * \brief Complex rotator
  * \ingroup math_operators_blk
+ *
+ * \details
+ *
+ * Message Ports:
+ *
+ * phase_inc (input):
+ *
+ *    Receives a PMT pair with a new phase increment to be set on the rotator at
```
mbr0wn: Suggestion: Make this a dict. Then the index can be optional. That could actually be useful for applications that don't have an actual control loop.

gr-blocks/lib/rotator_cc_impl.cc
```diff
@@ -40,9 +40,17 @@ rotator_cc::sptr rotator_cc::make(double phase_inc)
...
                  io_signature::make(1, 1, sizeof(gr_complex)),
-                 io_signature::make(1, 1, sizeof(gr_complex)))
+                 io_signature::make(1, 1, sizeof(gr_complex))),
+    d_idx_next_inc_update(0),
+    d_next_phase_inc(0.0),
+    d_inc_update_pending(false)
 {
     set_phase_inc(phase_inc);
+
+    message_port_register_in(pmt::mp("phase_inc"));
```
mbr0wn: We should make it a habit to not hard code the port name down here. Can you make a const further up?

gr-blocks/lib/rotator_cc_impl.cc
```diff
@@ -40,9 +40,17 @@ rotator_cc::sptr rotator_cc::make(double phase_inc)
...
+    set_msg_handler(
+        pmt::mp("phase_inc"),
```
mbr0wn: ...and then reuse it here.

gr-blocks/lib/rotator_cc_impl.cc
```diff
@@ -52,6 +60,19 @@ void rotator_cc_impl::set_phase_inc(double phase_inc)
     d_r.set_phase_incr(exp(gr_complex(0, phase_inc)));
 }
 
+void rotator_cc_impl::handle_phase_inc_msg(pmt::pmt_t msg)
+{
+    if (pmt::is_pair(msg)) {
```
mbr0wn: What if not? Please log a warning or other debug message.

gr-blocks/lib/rotator_cc_impl.cc
```diff
@@ -52,6 +60,19 @@ void rotator_cc_impl::set_phase_inc(double phase_inc)
...
+        pmt::pmt_t offset    = pmt::car(msg);
+        pmt::pmt_t phase_inc = pmt::cdr(msg);
+        if (pmt::is_uint64(offset) && pmt::is_real(phase_inc)) {
+            d_idx_next_inc_update = pmt::to_uint64(offset);
+            d_next_phase_inc      = pmt::to_double(phase_inc);
+            d_inc_update_pending  = true;
```
mbr0wn: We'll need a queue of outstanding phase inc changes. Here, sending multiple messages will overwrite the earlier ones.

igorauad:
Thanks for reviewing this PR.

> @igorauad Do we have a CLA from you? I'm flagging this just in case we don't. If we do, many apologies in advance. I've been away from GR/github a bit so I'm not quite up to date.

I don't have a CLA yet. I will send a message to Ben and will let you know.

> What would definitely spice up this PR is a GRC-based example.

I agree.  I wanted to propose the idea first, but I can definitely try to provide at least some unit tests, if not a GRC-based example. I'm working on an OOT module that would benefit from this and I thought this is really something for the in-tree rotator module, as it is such a common use case.

Also, something else that could be nice for this PR is tagging when the update is effectively handled by the rotator. I think this could be an optional behavior and would default to disabled. But it can be quite useful for some use cases. The main scenario I can think of is when the chain is something like:

rotator (as de-rotator) -> symbol sync / decimation -> frequency recovery

The frequency recovery block can schedule the update by taking the sample rate conversion (of the decimator) into account easily. In contrast, it may be more difficult (or at least more time consuming) to take the filter delay into account when scheduling the update. A solution is for the rotator to tag when it updates its increment and then for the frequency recovery block to check whether the update tag came where it expected. If it doesn't come where expected, the frequency recovery block can rely on the tag offset in order to infer the delay of the decimation block. 

This is something that we implemented in this [Blockstream Satellite - gr-blocksat module](https://github.com/Blockstream/gr-blocksat/blob/master/lib/frame_synchronizer_cc_impl.cc#L430). This is more specifically the consumer of the tag, which in turn schedules a frequency update on an upstream block that sits behind a polyphase clock sync (decimator) block.

Regarding your other comments, will work on them soon.

Thanks

mbr0wn:
I like the idea of optional tagging. Thanks for submitting a PR for discussion!

gr-blocks/include/gnuradio/blocks/rotator_cc.h
igorauad: Done

gr-blocks/lib/rotator_cc_impl.cc
igorauad: Agreed. Updated.

gr-blocks/lib/rotator_cc_impl.cc
igorauad: Do you prefer a print/warning message or really a `throw std::runtime_error`? I've put the latter, but let me know what you think.

gr-blocks/lib/rotator_cc_impl.cc
igorauad: This is still missing. Will complete on my next spare time slot :)

igorauad:
Also:
- I've added the mechanism for tagging the sample where the update starts to take effect.
- I've introduced some QA tests covering both the update that specifies sample offset as well as the update that does not.

gr-blocks/lib/rotator_cc_impl.cc
```diff
@@ -63,7 +102,38 @@ int rotator_cc_impl::work(int noutput_items,
...
-    d_r.rotateN(out, in, noutput_items);
+    int items_before_update, items_after_update;
+    /* If there is a phase increment update scheduled, handle rotation in two
+     * steps and update the phase increment in between. */
+    const uint64_t n_written = nitems_written(0);
+    if (d_inc_update_pending &&
+        d_idx_next_inc_update > n_written &&
+        d_idx_next_inc_update <= (n_written + noutput_items)) {
+
+        items_before_update = d_idx_next_inc_update - n_written - 1;
```
igorauad: I've pushed this fix also. The `-1` is needed here in order for the new phase increment to start taking effect on sample index `d_idx_next_inc_update`.  I realised this after introducing the QA tests.

gr-blocks/python/blocks/qa_rotator_cc.py
```diff
@@ -0,0 +1,199 @@
...
+
+        # Expected IQ samples
+        expected_angles  = 2 * np.pi * np.arange(self.n_samples) * f_out
+        expected_samples = np.exp(1j*expected_angles)
+
+        self.tb.connect(self.source, self.rotator_cc, self.sink)
+        self.tb.run()
+
+        for (x,y) in zip(self.sink.data(), expected_samples):
+            self.assertComplexAlmostEqual(x, y, places=5)
```
igorauad: I am aware that there is `assertComplexTuplesAlmostEqual`. However, I couldn't get it to work. It would always pass even when it shouldn't. I couldn't understand why.

gr-blocks/python/blocks/qa_rotator_cc.py
```diff
@@ -0,0 +1,199 @@
...
+        # Message to be sent to the rotator in order to update its increment
+        ctrl_msg = pmt.make_dict()
+        ctrl_msg = pmt.dict_add(ctrl_msg,
+                                pmt.intern("inc"),
+                                pmt.from_double(new_phase_inc))
+        ctrl_msg = pmt.dict_add(ctrl_msg,
+                                pmt.intern("offset"),
+                                pmt.from_uint64(offset))
+
+        rot_ctrl = blocks.message_strobe(ctrl_msg, msg_period_ms)
```
igorauad: This isn't very reliable and I accept suggestions. I'm using message period as 0 so that I try to get the message sent before processing the 100 samples that are tested. However, this does not work every time. Sometimes the samples are processed first. Hints?

gr-blocks/lib/rotator_cc_impl.cc
```diff
@@ -32,17 +32,28 @@
...
+    d_next_phase_inc(0.0),
+    d_inc_update_pending(false)
 {
     set_phase_inc(phase_inc);
+
+    const pmt::pmt_t port_id = pmt::mp("phase_inc");
+    message_port_register_in(port_id);
+    set_msg_handler(
+        port_id,
+        boost::bind(&rotator_cc_impl::handle_phase_inc_msg, this, _1));
```
mbr0wn:
```suggestion
        [this](pmt::pmt_t& msg){ this->handle_phase_inc_msg(msg); });
);
```

Please no boost::bind. And don't trust my lambda, I just winged it.

gr-blocks/lib/rotator_cc_impl.cc
mbr0wn: I would have just logged the error and skipped the message, but I have no objective reason to say so.

gr-blocks/python/blocks/qa_rotator_cc.py
mbr0wn: Meh. The macro is supposed to save you time developing, and it's clearly not, so let's move on.

gr-blocks/lib/rotator_cc_impl.cc
igorauad: Fixed.

gr-blocks/lib/rotator_cc_impl.cc
igorauad: I've just had a chance to add this. Used an `std::queue`. Let me know if OK.

gr-blocks/python/blocks/qa_rotator_cc.py
igorauad: @mbr0wn aside from this, I believe this MR is ready. If you have suggestions on how to solve the above testing issue, let me know.

To clarify, the goal here is to test a message sent to the rotator asking for an update of its's phase  increment. To do so, the message has to be sent first, before the rotator actually processes the samples. In this particular test the message asks for an update on sample offset 50, so `blocks.message_strobe` should convey the message before `blocks.rotator_cc` processes the first 50 samples, otherwise the phase increment will not be updated. This does not seem to work reliably using `blocks.message_strobe`, so not always the test will pass.

gr-blocks/lib/rotator_cc_impl.cc
```diff
@@ -40,6 +48,33 @@ void rotator_cc_impl::set_phase_inc(double phase_inc)
...
+        const double phase_inc =
+            pmt::to_double(pmt::dict_ref(msg, inc_key, pmt::PMT_NIL));
+
+        /* Absolute sample offset when update is to be applied (optional) */
+        const uint64_t update_offset =
+            pmt::dict_has_key(msg, offset_key)
+                ? pmt::to_uint64(pmt::dict_ref(msg, offset_key, pmt::PMT_NIL))
+                : nitems_written(0) + 1; // update right away when not provided
+
+        d_inc_update_queue.push(std::make_pair(update_offset, phase_inc));
```
mbr0wn: Almost! You  need to hold a mutex while accessing the queue, because the message handling and the work function are going to happen in separate threads. Since you're not already using it, using `d_setlock` is probably fine.

gr-blocks/lib/rotator_cc_impl.cc
```diff
@@ -51,7 +86,53 @@ int rotator_cc_impl::work(int noutput_items,
       for (int i=0; i<noutput_items; i++)
       	out[i] = d_r.rotate(in[i]);
 #else
-    d_r.rotateN(out, in, noutput_items);
+
+    const uint64_t n_written = nitems_written(0);
+
+    /* Find out if there is a phase increment update pending */
+    while (!d_inc_update_queue.empty() && !d_inc_update_pending) {
+        auto next_update = d_inc_update_queue.front();
```
mbr0wn: You also need to hold the mutex for this line and the next.

gr-blocks/python/blocks/qa_rotator_cc.py
mbr0wn: If you call `to_basic_block()._post()` to pass in the message instead of using a message strobe, you might be able to force the message into the queue before you start the FG.

mbr0wn: Quick grep found this: https://github.com/gnuradio/gnuradio/blob/64ee803068269fe3b9563fd415c420fb513a1c9b/gr-digital/python/digital/qa_header_payload_demux.py#L332-L336

This might work for you also.

gr-blocks/lib/rotator_cc_impl.cc
```diff
@@ -40,6 +48,33 @@ void rotator_cc_impl::set_phase_inc(double phase_inc)
     d_r.set_phase_incr(exp(gr_complex(0, phase_inc)));
 }
 
+void rotator_cc_impl::handle_phase_inc_msg(pmt::pmt_t msg)
+{
+    const pmt::pmt_t inc_key = pmt::intern("inc");
```
michaelld: these could be globally defined

gr-blocks/lib/rotator_cc_impl.cc
michaelld: Here's the code, for both here and in `::work`:
```
gr::thread::scoped_lock guard(d_setlock);
```
Using the overall block lock like this is the way to go.

gr-blocks/lib/rotator_cc_impl.cc
michaelld: for example `static const`, not just `const`

mbr0wn:
@igorauad this is still a great change, but the mutex is really important. It would be great if you can finish the work you did here so we can merge this. If you're not available, I might close it just so we have a better overview of our PR queue, but of course you can reopen anytime you have changes. I'll give it a few days before doing that though.

igorauad:
@mbr0wn thanks. I haven't abandoned this change. However, I've been busy lately and haven't had a chance to complete the work. I'm thinking I might be able to work on it next week. This change is a requirement for an OOT module I'm developing, so I will definitely want to complete the work.

Please feel free to close the PR temporarily if that works best for your organization. I will reopen it as soon as I find a chance. That's all fine.

mbr0wn:
> Please feel free to close the PR temporarily if that works best for your organization. I will reopen it as soon as I find a chance. That's all fine.

Many thanks. I'll close it for the time being --  your code is still very much appreciated!

igorauad:
Hi @mbr0wn @michaelld ,

I would like to reopen this PR after a long time. I finally had a chance to get back to the project I was working on last year when I created this PR. So I've worked a bit more on it and ended up changing the implementation considerably. The essence is still the same, though. It will let other blocks control the phase increment adopted by the rotator block. 

I've rebased the changes using the new pybind11 structure, added the changes you had suggested last time, and a couple of extra unit test cases.

Let me know if you are available to take another look.

Thanks!

mbr0wn:
Just rebase and reopen. No process required! Or you can open a new PR. Up to you.

igorauad:
Thanks, @mbr0wn. The reopen button was disabled, but I figured out the problem after checking [this](https://gist.github.com/robertpainsi/2c42c15f1ce6dab03a0675348edd4e2c). So I've just reopened it.

gr-blocks/python/blocks/qa_rotator_cc.py
igorauad: Thanks. I used `insert_tail`, which is what would `_post` would call AFAIU.

gr-blocks/lib/rotator_cc_impl.cc
igorauad: Add the `scoped_lock` to `handle_ctrl_msg()` and `work()` as suggested.

gr-blocks/lib/rotator_cc_impl.cc
igorauad: Done :+1:

gr-blocks/lib/rotator_cc_impl.cc
```diff
@@ -40,18 +46,91 @@ void rotator_cc_impl::set_phase_inc(double phase_inc)
...
+     * in the control message, update the phase increment right away. */
+    phase_inc_update_t update{};
+    update.phase_inc = pmt::to_double(pmt::dict_ref(msg, inc_key, pmt::PMT_NIL));
+    update.offset = pmt::dict_has_key(msg, offset_key)
+                        ? pmt::to_uint64(pmt::dict_ref(msg, offset_key, pmt::PMT_NIL))
+                        : nitems_written(0);
+
+    d_inc_update_queue.push(update);
+}
+
```
igorauad, line 77: In the previous implementation, the message port was named `phase_inc` and this method was called `handle_phase_inc_msg`. I've decided to make it more generic to facilitate future expansion. So I named the port `ctrl` and the message handler `handle_ctrl_msg`.

Not sure if this is common practice in the GR codebase, but the idea would be that we could send multiple commands to the rotator. This implementation adds a single command (to change the phase increment), but it would be easy to add other commands (e.g., to set the phase) using the same port and message handler function.

Now, two questions:
- Should we stick to this or should I revert back to a dedicated `phase_inc` port?
- If we use the suggested generic `ctrl` port, should I add a `command` field on the PMT message and call it `set_phase_inc`? The rationale is that, if other commands are added in the future, it would be easier to coordinate them with `command` field on the input `msg`.

willcode:
I think we should move toward having a single command/control message port on blocks, instead of adding new ports for each function. That way, if something wanted to command a frequency change and set the phase at the same time, it could. This also means that all keys would be optional. One key might still be required if another one is to make sense.

gr-blocks/python/blocks/qa_rotator_cc.py
```diff
@@ -0,0 +1,322 @@
...
+        # Now, schedule an update that is out-of-range due to being in the past
+        self._post_phase_inc_ctrl_msg(new_phase_inc, offset=0)
+
+        # And run the flowgraph to completion
+        self.tb.wait()
+
+        # The increment update initially scheduled for the future should have
+        # been applied when processing the second half of samples. In contrast,
+        # the update scheduled for the past should have been discarded.
+        self._assert_tags([new_phase_inc], [n_half_samples])
```
igorauad, line 318: I should note that this test may be slightly unreliable, given that it has a bit of racing. I couldn't think of a better way to implement it, though.

Essentially, I wanted to call the `_assert_tags` on L303 after the first `work` call on the rotator block. Also, I wanted to post a control message on L310 after processing half of the samples. So I added `while` loops, which are not perfect, but worked for me.

Let me know what you think.

igorauad:
> I think we should move toward having a single command/control message port on blocks, instead of adding new ports for each function. That way, if something wanted to command a frequency change and set the phase at the same time, it could. This also means that all keys would be optional. One key might still be required if another one is to make sense.

Cool, yes, that's what I had in mind. So, in this case, if `command == set_phase_inc`, the `inc` key is required. Otherwise, the `inc` is not mandatory. For instance, if there was a `set_phase` command, the required key would probably be named `phase`.

willcode:
Since there is already a `freq` port, that should be changed to be a `cmd` port now. The port would allow things like (making this up, may not all make sense):

`{'freq': 100000}` or `{'freq': 100000, 'phase': 0.1, 'offset': 10}` or `{'phase_inc': 0.1, 'offset': 10}`

with some rules about which combinations make sense.

igorauad:
> Since there is already a `freq` port, that should be changed to be a `cmd` port now. The port would allow things like (making this up, may not all make sense):
> 
> `{'freq': 100000}` or `{'freq': 100000, 'phase': 0.1, 'offset': 10}` or `{'phase_inc': 0.1, 'offset': 10}`
> 
> with some rules about which combinations make sense.
 
I see and I agree. We don't need a `cmd` key on the dictionary as long as we can infer the command based on the keys that are present. For example, if `phase_inc` is present, we should update the phase increment. If something else (e.g., `phase`) was present (say the message is `{'phase_inc': 0.1, 'offset': 10, 'phase': 0.2}`, we would handle both the phase and the phase increment. So I changed the implementation slightly in 9dc9dfe2fb1149d3fd979ab238fe4b42785c66ec.

willcode:
Yes, something like that. Then we can (later) add handling for freq to the same message handler, maybe keeping it also in the old message handler for compatibility.

We use 'cmd' in the signal source, and 'command' in the uhd blocks. Maybe one of those would be better than ctrl. This hasn't been well standardized. Someone else have an opinion on that?

willcode:
What is the purpose of having a message queue? From a technical perspective, it seems fine, but from a control loop perspective? Is there the thought that the desired phase shift will be known for future samples? Or is the thought that this will be used as a way to set up future configuration? I can see a potential use for setting up precise frequency hopping using something like this.

igorauad:
> What is the purpose of having a message queue? From a technical perspective, it seems fine, but from a control loop perspective? Is there the thought that the desired phase shift will be known for future samples? Or is the thought that this will be used as a way to set up future configuration? I can see a potential use for setting up precise frequency hopping using something like this.

I'm trying to address frequency correction on frame-by-frame processing schemes used in so many PHY implementations (e.g., DVB-S2). Usually, there is a frame structure with some pilot symbols at the beginning of the frame and then a pilot-aided carrier phase recovery loop that restarts on every frame. Whenever a new frame starts, and there are some pilots, the phase recovery loop will quickly snap back to the estimated carrier phase (more easily estimated with the pilots). Then, once the phase recovery loop starts to process the normal (non-pilot) data symbols, it will proceed with decision-directed carrier phase recovery (relying on decisions instead of know pilots), which is far more challenging under low SNR.

In addition to the carrier phase recovery, there is usually a separate frequency recovery block, which is meant to correct larger frequency errors (the so-called coarse freq. recovery). Furthermore, this coarse freq. recovery block usually comes first in the pipeline. In this context, the key point is that the coarse frequency correction block can disturb the carrier phase recovery block downstream in the pipeline. Any change in the frequency correction can easily disturb the phase recovery loop, especially if the frequency correction happens when the phase recovery loop is in its decision-directed processing operation (processing the data symbols instead of the pilots).

To overcome all this, it's quite important to update the coarse frequency correction only at the beginning of frames.  WIth frame-by-frame processing, the carrier phase recovery loop typically starts over from scratch on every frame. Hence, if the coarse frequency correction happens at the beginning of a frame, there is usually plenty of time for the phase recovery loop to adapt. As a result, the two blocks can coexist nicely.

So the queue is meant to provide deterministic scheduling of phase increment updates. The block sending the control message will say exactly when the update needs to be applied. Presumably, the sender block has the notion of frames already, so it knows exactly when a frame starts and schedules the update on that sample. 

There are some extra comments on the commit message (see b3f4232e12c21468783cf714423195f643cdb899).

willcode:
If you don't hear from anyone else, let's go with "cmd". The work "command" takes up too much space in GRC.

igorauad:
> If you don't hear from anyone else, let's go with "cmd". The work "command" takes up too much space in GRC.

Sounds good, @willcode. I've just renamed.

willcode:
I'll give it a try soon. By the way, the GRC and .h still have "ctrl".

igorauad:
> I'll give it a try soon. By the way, the GRC and .h still have "ctrl".

Ah, right! Fixed. Thanks

willcode:
Built/ran tests, and it went smoothly. The QA looks very complete, though I haven't spent the time to QA the QA.

One more naming thing ... once the tag flows downstream, `new_inc` could mean anything. I don't know what
a good name would be. Something like `phase_inc_update`? Something shorter but that still gives a clue what
the tag pertains to?

Other than that, I think we should stop torturing you and merge this :-)

igorauad:
> Built/ran tests, and it went smoothly. The QA looks very complete, though I haven't spent the time to QA the QA.
> 
> One more naming thing ... once the tag flows downstream, `new_inc` could mean anything. I don't know what
> a good name would be. Something like `phase_inc_update`? Something shorter but that still gives a clue what
> the tag pertains to? Just `rotator_update`?
> 
> Other than that, I think we should stop torturing you and merge this :-)

Thanks, @willcode No problem with code review rounds at all! 😄

I like the idea of a more descriptive tag. I think the tag should have a `rotator_` or `rot_` prefix so that downstream blocks can know where it comes from. Or, at least, it is easier to get a clue regarding which block is producing the tag if looking at a GUI plot. So, perhaps one of the following:

1. `rot_phase_inc`/`rotator_phase_inc`
2. `rot_inc_update`/`rotator_inc_update`
3. `rot_phase_inc_update`/`rotator_phase_inc_update`

The shortest one is `rot_phase_inc`. What do you think?

willcode:
rot_phase_inc sounds good.

igorauad:
> rot_phase_inc sounds good.

Ok 👍 Updated, squashed and rebased.

willcode (approved):

igorauad:
@willcode I've just pushed again to update the commit message with a `blocks: ` prefix.

gr-blocks/include/gnuradio/blocks/rotator_cc.h
```diff
@@ -20,6 +20,23 @@ namespace blocks {
 /*!
  * \brief Complex rotator
  * \ingroup math_operators_blk
+ *
```
mormj, line 23: :+1: for adding documentation to the block

gr-blocks/lib/rotator_cc_impl.cc
```diff
@@ -40,18 +46,94 @@ void rotator_cc_impl::set_phase_inc(double phase_inc)
...
 int rotator_cc_impl::work(int noutput_items,
                           gr_vector_const_void_star& input_items,
                           gr_vector_void_star& output_items)
 {
+    gr::thread::scoped_lock l(d_mutex);
     const gr_complex* in = (const gr_complex*)input_items[0];
     gr_complex* out = (gr_complex*)output_items[0];
 
 #if 0
```
mormj, line 89: This #if 0 code can go

gr-blocks/lib/rotator_cc_impl.cc
```diff
@@ -40,18 +46,94 @@ void rotator_cc_impl::set_phase_inc(double phase_inc)
...
+            int items_before_update = next_update.offset - n_written - nprocessed_items;
+            d_r.rotateN(out, in, items_before_update);
+            nprocessed_items += items_before_update;
+
+            set_phase_inc(next_update.phase_inc);
+
+            if (d_tag_inc_updates) {
+                add_item_tag(0,
+                             next_update.offset,
+                             pmt::string_to_symbol("rot_phase_inc"),
```
mormj, line 127: The tag key can be stored as a static class variable so the pmt for it doesn't have to be regenerated each time

mormj (approved):
Some minor things, but overall looks good.

mormj:
I think it's also worth considering (beyond the scope of this PR) adding this functionality more broadly for many types of blocks - how do we consistently change a parameter value at a specific sample number, and indicate that the change has occurred.  This could apply to almost any block.

gr-blocks/lib/rotator_cc_impl.cc
igorauad: Done (5f1052a5a3fd97e5c0a6b7e435aafcde1275ec99)

gr-blocks/lib/rotator_cc_impl.cc
igorauad: Agreed! Done in 11c3a1da1662bf13327782be721ccbcb4054db25

igorauad:
> I think it's also worth considering (beyond the scope of this PR) adding this functionality more broadly for many types of blocks - how do we consistently change a parameter value at a specific sample number, and indicate that the change has occurred. This could apply to almost any block.

That would be nice indeed. Also related is a general-purpose (and optionally registered) message port capable of calling public methods of the class with arguments provided via a PMT dictionary. The combination of such a message port with what you are suggesting (the ability to schedule changes at a specific sample offset) can enable lots of use cases. In the rotator case, if both features were available at a base class level, I could use the general-purpose port to call the pre-existing `set_phase_inc` method on a scheduled sample offset.

I'll see if I can understand how that could possibly be introduced at a base class level. As you said, it's something for a future investigation beyond the scope of this PR.

bistromath:
@mormj, what about your newsched work here?

https://github.com/gnuradio/newsched/pull/71

Seems like it would pretty seamlessly bridge that gap, although it might be out of the scope of this discussion.

bistromath:
Is this a candidate for its own GREP, or at least its own issue tag? A Qt-style signals/slots pattern for GRC would enable all sorts of fun applications.

mormj:
Certainly would be a great candidate for a GREP!

willcode:
There would be a lot to consider to do it right. Worth a read: https://github.com/pothosware/PothosCore/blob/master/include/Pothos/Managed/Class.hpp

Is there anything else about this PR that needs attention?

mormj:
I think this looks good as a solution for this block, so going ahead and merging.

PR #2992, opened 2020-01-02
runtime: fix python math script generating wrong sine table

ArtemPisarenko:
Fixes commit 48d535e9d9d95524b2c8ed35b44c7eda541aea68

### DIFF
--- a/gnuradio-runtime/lib/math/gen_sine_table.py
+++ b/gnuradio-runtime/lib/math/gen_sine_table.py
@@ -20,7 +20,6 @@
 # Boston, MA 02110-1301, USA.
 #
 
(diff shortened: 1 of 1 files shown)


ArtemPisarenko:
Have no idea why CI build failed. The file I changed isn't used anywhere in build/test process (and this is a reason why mistake being fixed is still alive).

ArtemPisarenko:
> How on earth did you find this? Who uses this function?
The one who tried to analyze complete fxpt algorithm to experiment with :)

Even better solution would be to generate sine_table.h during build. It's pretty straightforward with CMake. I could prepare pull request, if anybody agrees it makes sense.

PR #2994, opened 2020-01-02
More backport 3.8

mbr0wn:
Backporting a few things...

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,104 +26,41 @@ as many questions have already been asked and answered. Please also
 the mailing list and post your new questions there.
 
 
-## How to Build GNU Radio
+## How to Install GNU Radio
+
+### Prebuilt Binaries
+
+The recommended way to install GNU Radio on most platforms is using available binary package distributions. 
+
+The following command is for Debian, Ubuntu, and derivatives. It will install Release 3.7 with Python2. 
+
+    sudo apt install gnuradio
+
+For other operating systems, see [Installing from Binaries](https://wiki.gnuradio.org/index.php/InstallingGR#From_Binaries)
 
 ### PyBOMBS
-PyBOMBS (Python Build Overlay Managed Bundle System) is the recommended 
-method for building and installing GNU Radio. Please see 
-https://github.com/gnuradio/pybombs for detailed instructions. Abbreviated 
-instructions are duplicated below.
 
-1. Install PyBOMBS:
-    ```
-    $ [sudo] pip install PyBOMBS
-    ```
-    or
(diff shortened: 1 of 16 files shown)


grc/core/blocks/embedded_python.py
```diff
@@ -149,10 +149,10 @@ def rewrite(self):
     def _update_params(self, params_in_src):
         param_factory = self.parent_platform.make_param
         params = {}
-        for param in list(self.params):
-            if hasattr(param, '__epy_param__'):
-                params[param.key] = param
-                del self.params[param.key]
+        for key, value in self.params.copy().items():
```
mbr0wn: An open Q was, does this work in Python2.

marcusmueller: oh, that's the reason I was fighting with oldish containers to even get to try that.

marcusmueller: I forgot what the reason was and gave up. Now I have new motivation. Also, we need internal unit test for GUI stuff, honestly, at some point.

README.md
```diff
@@ -26,104 +26,41 @@ as many questions have already been asked and answered. Please also
...
 
 
-## How to Build GNU Radio
+## How to Install GNU Radio
+
+### Prebuilt Binaries
+
+The recommended way to install GNU Radio on most platforms is using available binary package distributions. 
+
+The following command is for Debian, Ubuntu, and derivatives. It will install Release 3.7 with Python2.
```
marcusmueller: Since maint-3.8 isn't as much of a moving target, this needs a 

> (as of DEBIAN_VERSION, UBUNTU_VERSION)

We're obviously expecting things to start moving.

Also, since this is maint-3.8, this should explicitly link to Josh's PPA! That's the recommended way these days.

grc/core/blocks/embedded_python.py
marcusmueller: (by the way, yes, we should be testing things with Py2. Why are you esp. nervous about this line?)

mbr0wn: I can't remember...

README.md
mbr0wn: If people only need the native packaged version, is that a true statement? In any case, I would prefer we refer to the wiki here, and only have build-from-source-related instructions checked into git.

grc/core/blocks/embedded_python.py
mbr0wn: This works in Py2k.

PR #2997, opened 2020-01-03
Disable Python 2k

mbr0wn:
- Update SWIG settings
- Update CMake
- Remove Python2 references

Note: This does not touch a lot of Python files, and many Python files
will still work with Python 2 after this commit. However, we won't allow
that in our CMake, and the support will be gone too.

This supersedes https://github.com/gnuradio/gnuradio/pull/2716.

Note GR devs: I pushed this branch to the main repo, so we can force-push-update it if necessary.


mbr0wn:
This was already reviewed before, and the CI passes. If no one objects, I'll merge it before Monday.

Issue #2926, opened 2019-11-25
Additive scrambler reset by count


michael-roe:
Additive scrambler seems to work OK with a tagged stream. But if a count rather than a tag is used to reset it, it looks like the sequence gets out of step.

Here's a dump of the output with mask=011,seed=0x1c9,length=8,count=8,bits_per_byte=1,reset_tag_key=""

contents = 
0000: 01 00 00 01 00 00 01 01 01 01 00 00 01 00 00 01 
0010: 01 01 00 00 01 00 00 01 01 01 00 00 01 00 00 01 
0020: 01 01 00 00 01 00 00 01 01 01 00 00 01 00 00 01 
0030: 01 01 00 00 01 00 00 01 01 01 00 00 01 00 00 01 
0040: 01 01 00 00 01 00 00 01 01 01 00 00 01 00 00 01 

The first 8 values are OK, and then (after the first reset) it gets out of step.

(The particular LFSR above is the same as the one used in Yaesu System Fusion. In real use, frames would be longer than 8 bits but I'm using 8 bit frames to show the bug more easily),

michael-roe:
P.S. This was  with gnuradio built from the master branch.

rear1019:
There is indeed an error in ``additive_scrambler_bb_impl.cc``: Every time ``work()`` is called the very first reset is delayed by one byte (the reset is byte based). I might make a pull request with a fix by end of the week.

PR #2999, opened 2020-01-03
Fix for #2926: Additive scrambler reset by count

rear1019:
See commit messages. Consider to ignore whitespace changes (``git diff -w``) when viewing the diff.


willcode:
Was backported to maint-3.8 as part of https://github.com/gnuradio/gnuradio/pull/3248

PR #3000, opened 2020-01-03
gr-fft: Modernize and simplify code

ThomasHabets:


gr-fft/include/gnuradio/fft/fft.h
```diff
@@ -29,25 +29,20 @@
...
 #include <gnuradio/gr_complex.h>
+#include <volk/volk_alloc.hh>
 #include <boost/thread.hpp>
 
 namespace gr {
 namespace fft {
 
 
 /*! \brief Helper function for allocating complex* buffers
+ * TODO: Remove once with the single user of this stops using it.
```
mbr0wn:
```suggestion
 * TODO: Remove once when the single user of this stops using it.
```

gr-fft/include/gnuradio/fft/fft.h
```diff
@@ -29,25 +29,20 @@
...
-/*! \brief Helper function for allocating float* buffers
- */
-FFT_API float* malloc_float(int size);
-
-/*! \brief Helper function for allocating double* buffers
- */
-FFT_API double* malloc_double(int size);
-
 /*! \brief Helper function for freeing fft buffers
+ * TODO: Remove once with the single user of this stops using it.
```
mbr0wn:
```suggestion
 * TODO: Remove once when the single user of this stops using it.
```

gr-fft/lib/fft.cc
```diff
@@ -167,37 +156,27 @@ static void export_wisdom()
 // ----------------------------------------------------------------
 
 fft_complex::fft_complex(int fft_size, bool forward, int nthreads)
+    :d_fft_size(fft_size),
```
mbr0wn: I don't think this is what clang-format says.

gr-fft/include/gnuradio/fft/fft.h
```diff
@@ -115,26 +113,29 @@ class FFT_API fft_complex
...
 
     /*
      * These return pointers to buffers owned by fft_impl_fft_real_fwd
      * into which input and output take place. It's done this way in
      * order to ensure optimal alignment for SIMD instructions.
      */
-    float* get_inbuf() const { return d_inbuf; }
-    gr_complex* get_outbuf() const { return d_outbuf; }
+    float* get_inbuf() const { return const_cast<float*>(d_inbuf.data()); }
+    gr_complex* get_outbuf() const { return const_cast<gr_complex*>(d_outbuf.data()); }
```
mbr0wn: Does this also need to be writable? If not, can't we just return a const pointer and drop the const cast?

gr-fft/include/gnuradio/fft/fft.h
```diff
@@ -115,26 +113,29 @@ class FFT_API fft_complex
...
     virtual ~fft_real_fwd();
 
     /*
      * These return pointers to buffers owned by fft_impl_fft_real_fwd
      * into which input and output take place. It's done this way in
      * order to ensure optimal alignment for SIMD instructions.
      */
-    float* get_inbuf() const { return d_inbuf; }
-    gr_complex* get_outbuf() const { return d_outbuf; }
+    float* get_inbuf() const { return const_cast<float*>(d_inbuf.data()); }
```
mbr0wn: const_cast is almost always cheating. In this case, I think that we should simply drop the const qualifier on the method (we're changing API already anyway). Rationale: If we want to return a writable pointer, then this function is not truly const, even if it's only assisting in mutation.

gr-fft/include/gnuradio/fft/fft.h
```diff
@@ -115,26 +113,29 @@ class FFT_API fft_complex
...
+    gr_complex* get_outbuf() const { return const_cast<gr_complex*>(d_outbuf.data()); }
 
-    int inbuf_length() const { return d_fft_size; }
-    int outbuf_length() const { return d_fft_size / 2 + 1; }
+    int inbuf_length() const { return d_inbuf.size(); }
+    int outbuf_length() const { return d_outbuf.size(); }
 
     /*!
      *  Set the number of threads to use for caclulation.
```
mbr0wn:
```suggestion
     *  Set the number of threads to use for calculation.
```

(while you're touching this...)

gr-fft/include/gnuradio/fft/fft.h
ThomasHabets: I put this in the commit message, but should have put it in the pull request:

```
 For reviewers:
    Note the const_cast<>. I'm not sure what you prefer, but we need to
    return a writable pointer.
```

I understood `get_inbuf()` to be the way things are written into here. But I'll give it a try to see if it breaks anything.

ThomasHabets: Yeah the pointers need to be writable (both), so dropping the const.

ThomasHabets:
Updated.

mbr0wn:
Will let this sit for others to weigh in, but it looks good.

Issue #2950, opened 2019-12-06
GRC does not connect() blocks with no ports even though GR supports this


kpreid:
Steps to reproduce:

1. Define a hierarchical block which has no input or output ports, but which has blocks that do some detectable work inside it.
2. Define a top block, in GNU Radio Companion, which contains this hierarchical block.
3. Run the generated top block; observe that nothing happens.

The generated code instantiates the hier_block (e.g. `self.hier_block_no_ports_0 = hier_block_no_ports()`) but does not connect it (e.g. `self.connect(self.hier_block_no_ports_0)`, using the 1-argument form of the `connect` method), so the blocks within the hier block do not do anything. If I manually add the `connect` call then the self-contained graph within the hier block does execute.

Thus, GR supports this case but GRC does not. I propose that GRC be modified so that, **if a block in a top/hier block has no ports, then a 1-argument `connect` call will be generated for it.**

Compatibility considerations: As far as I know, this should not affect the behavior of existing correct programs; however, if for some reason a block is declared to GRC to have no ports but actually does have ports, this will cause a new runtime error (whereas the block would previously have been instantiated but otherwise ignored).

dkozel:
Thanks for filing the issue. This seems like a clear and hopefully reasonably straightforward improvement to GRC!

haakov:
I can take care of this. Note to self: Don't `connect()` blocks that don't have a make template

haakov:
@kpreid I've added a PR: https://github.com/gnuradio/gnuradio/pull/3007

Can you please confirm that it fixes the bug?

kpreid:
@haakov I attempted, but I'm currently having trouble building GR from source. Sorry. The code looks reasonable to me.

PR #3007, opened 2020-01-04
grc: Call connect() for hier blocks that don't have ports

haakov:
Fixes https://github.com/gnuradio/gnuradio/issues/2950.

### DIFF
--- a/grc/core/generator/cpp_hier_block.py
+++ b/grc/core/generator/cpp_hier_block.py
@@ -124,6 +124,9 @@ class CppHierBlockGenerator(CppTopBlockGenerator):
             )
         else:
             t['make'] = '{cls}()'.format(cls=block_id)
+        # Self-connect if there aren't any ports
+        if not data['inputs'] and not data['outputs']:
+            t['make'] += '\nthis->connect(this->${id});'
 
         # Callback data
         t['callbacks'] = [
--- a/grc/core/generator/hier_block.py
+++ b/grc/core/generator/hier_block.py
@@ -117,6 +117,10 @@ class HierBlockGenerator(TopBlockGenerator):
             )
         else:
             t['make'] = '{cls}()'.format(cls=block_id)
+        # Self-connect if there aren't any ports
+        if not data['inputs'] and not data['outputs']:
+            t['make'] += '\nself.connect(self.${id})'
+
         # Callback data
         t['callbacks'] = [
             'set_{key}(${{ {key} }})'.format(key=param_block.name) for param_block in parameters


skoslowski:
Maybe these kind of connect calls should be added as a template to the stream domain spec.

haakov:
I thought about that. But wouldn't that cause problems for blocks that only have message ports?

marcusmueller:
I see the message-port-only problem; but what about e.g. RFNoC blocks?

haakov:
I'm afraid I haven't used RFNoC blocks yet. Could you explain how they relate to this?

bastibl:
There are also GRC Blocks that are just there to inject code in the flowgraph template (e.g. CtrlPort Monitor). They might also have code in `make`. Not sure if it would work with this logic.
https://github.com/gnuradio/gnuradio/blob/master/gr-blocks/grc/blocks_ctrlport_performance.block.yml

nickoe:
@haakov Ping

haakov:
@nickoe thanks for the reminder!

Maybe it is easier to just add the `connect()` call in the hier block's make template.
Like this: https://github.com/haakov/gnuradio/commit/3718527ceb54b8e4be2bbe43875b84eb8288a060 (C++ part not tested yet)

If there are other blocks that need to be connected, the call can be added in their YAML files too, but I can't think of any right now.

haakov:
Updated to only affect hier blocks.

haakov:
oop, just realized this won't work for hier blocks with exclusively optional ports. Never mind

mbr0wn:
> oop, just realized this won't work for hier blocks with exclusively optional ports. Never mind

Looks like this will be more work then?

haakov:
Actually, no. I'd like to withdraw my previous comment :)

Turns out that connecting a hier block with exclusively optional ports is not allowed anyways:
```
  File "/usr/local/lib/python3.9/site-packages/gnuradio/gr/hier_block2.py", line 100, in connect
    self.primitive_connect(*args)
ValueError: Block Sink(1) must not have any input or output ports
```
This PR is ready for review.

grc/core/generator/cpp_hier_block.py
```diff
@@ -124,6 +124,9 @@ def var_or_value(name):
             )
         else:
             t['make'] = '{cls}()'.format(cls=block_id)
+        # Self-connect if there aren't any ports
+        if {len(data['inputs']), len(data['outputs'])} == {0}:
```
skoslowski, line 128: IMO this does not convey your intend very well. May I suggest (for readability) to use  
`if not data["inputs"] and not data["outputs"]:`

skoslowski (changes requested):
I can confirm, that this adds the desired single-arg connect() call. I used a hier_block with a message strobe and message debug, which I put in a otherwise empty top_block. Only with this patch applied output is generated.

grc/core/generator/cpp_hier_block.py
haakov: Agreed! I'll fix it

mbr0wn: When you do, can you please rebase onto the latest master?

Issue #2952, opened 2019-12-09
Too much space around gui widgets


Phil67ago:
This due to a lot of QT QLayout objects being used frequently.
Each adds a 9px margin.
A 'QT Gui Range' has one QLayout around Slider and Edit box.
Then one QLayout around Label and above.
So 2 Range widgets can not get closer than (9+9)+6+(9+9)=42px (6px is Vertical spacing).

Patch against maint-3.8:

[gnuradio-layout-graphs-widget_maint-3.8.patch.txt](https://github.com/gnuradio/gnuradio/files/3938692/gnuradio-layout-graphs-widget_maint-3.8.patch.txt)
[gnuradio-layout-range-widget_maint-3.8.patch.txt](https://github.com/gnuradio/gnuradio/files/3938693/gnuradio-layout-range-widget_maint-3.8.patch.txt)

dkozel:
Thanks @Phil67ago for the report and submitting patches! We'll take a look and see how this effects the layout.

PR #3008, opened 2020-01-04
QT: Add margin around QT widgets

devnulling:
These are the patches from #2952, from @Phil67ago. 

Fixes: https://github.com/gnuradio/gnuradio/issues/2952


devnulling:
Screen cap of this patch in action.

Left side has this patch, right side is current master. 

qtfixup

Issue #2907, opened 2019-11-15
Request: Add parameter for filename in Tagged File Sink. 


Phil67ago:
Hi!
Would it be possible to add File Name in 'Tagged File Sink'

File Name = Pluto-SDR %Y-%M-%d %H:%M:%S - 
Result: Pluto-SDR 2019-51-15 15:51:56 - 17_0_262.34583333.dat

Untested patch:
[gnuradio-3.7.13.5-tagged_file_sink_name.patch.txt](https://github.com/gnuradio/gnuradio/files/3851988/gnuradio-3.7.13.5-tagged_file_sink_name.patch.txt)

marcusmueller:
Wow, thanks! This is way more than a request as you claim in your title, but already a patch. Nice!

Problem: This is a patch against the current 3.7-series source code- which kind of is in legacy/maintenance mode. We generally prefer to add features to the current master branch, before backporting them to maint-3.8 (if feasible) and (if someone champions to test) even to maint-3.7. In fact, this _can't_ be merged into the 3.7 series, as it uses `auto`, which is a C++11 feature, and we can't require C++11 on maint-3.7.

As for this specific patch: I spot at least one error that would inhibit compilation (dot in parameter list of make where there should be a comma).

Is there a realistic chance you could fix that and try to make a pull request against master or maint-3.8 out of this, so that we can put it through our automated testing? Ideally, you'd even invent a short python test for it and add it to the tests we run.

Phil67ago:
Fixing patch for master is no problem.
But I have hard time to cope with PyBOMBS.

1. I got first run to work on Fedora 30:

sudo dnf install python3-pip
sudo pip3 install pybomb			
/usr/local/bin/pybombs recipes add gr-recipes git+https://github.com/gnuradio/gr-recipes.git
mkdir ~/prefix
/usr/local/bin/pybombs prefix init ~/prefix/default/
sed -i 's!gitbranch: maint-3.8!gitbranch: master!g; s!^\(.*config_opt: .*\) \"$!\1 -DENABLE_CTRLPORT_THRIFT=OFF\"!g' ~/.pybombs/recipes/gr-recipes/gnuradio.lwr
sed -i 's!python2.6!python3.7!g;' ~/prefix/default/setup_env.sh
/usr/local/bin/pybombs install gnuradio 
/usr/local/bin/pybombs run gnuradio-companion <-- Ok, got 3.9.0.0 running!

2. Apply patch

Open another Terminal window:
[...]

PR #3010, opened 2020-01-05
Blocks: Add filename param to Tagged Filesink

devnulling:
PR up for testing and discussion.

Fixes: https://github.com/gnuradio/gnuradio/issues/2907

### DIFF
--- a/gr-blocks/include/gnuradio/blocks/tagged_file_sink.h
+++ b/gr-blocks/include/gnuradio/blocks/tagged_file_sink.h
@@ -57,8 +57,14 @@ public:
      * \param itemsize The item size of the input data stream.
      * \param samp_rate The sample rate used to determine the time
      *                  difference between bursts
+     * \param filename Ex: 'FileNameKeyword %%Y-%%m-%%d %%H:%%M:%%S.dat' 
+     *                 result as:
+     *                 'FileNameKeyword 2019-12-05 15:51:56.dat'.
+     *                 If empty: 'file17_0_262.34583333.dat' 
+     *                 (file, unique_id, no, timeval)
+     *                 See also, boost time Format Flags...
      */
-    static sptr make(size_t itemsize, double samp_rate);
+    static sptr make(size_t itemsize, double samp_rate, const char *filename );
 };
 
 } /* namespace blocks */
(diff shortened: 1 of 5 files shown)


gr-blocks/include/gnuradio/blocks/tagged_file_sink.h
```diff
@@ -57,8 +57,14 @@ class BLOCKS_API tagged_file_sink : virtual public sync_block
      * \param itemsize The item size of the input data stream.
      * \param samp_rate The sample rate used to determine the time
      *                  difference between bursts
+     * \param filename Ex: 'FileNameKeyword %%Y-%%m-%%d %%H:%%M:%%S.dat' 
+     *                 result as:
+     *                 'FileNameKeyword 2019-12-05 15:51:56.dat'.
```
mormj: Can we make the filenames without spaces?

mormj:
I love the idea here - the behavior of the current tagged stream file sink block - default filenames in the current location is a bit problematic.  When running grc the files get dumped to whatever directory you launched grc in.

With this implementation I have a few issues:
1) it won't run - there is a swig exception when I compile and run
2) filenames end up with spaces which I don't prefer
3) directory is still default (ambiguous)

Perhaps we should add an output directory, or allow that to be part of the pathname

but overall I think this idea will be a big improvement on the tagged stream file sink

gr-blocks/include/gnuradio/blocks/tagged_file_sink.h
dkozel: That's just an example. The default value is an empty string, it isn't immediately clear what name that will result in. The default not having spaces would be a good thing.

devnulling: I'll fixup the example, I didn't like this space either.

devnulling:
This currently doesn't work as expect, I ran out of time the other day debugging it, I'll dig into it more and push up the required changes.

mbr0wn:
This PR needs a rebase. @mormj  also had some good issues. I'll close due to lack of activity. Please reopen when it's back on track. Thanks!

PR #3012, opened 2020-01-05
Replace rand() by gr::random() (maint-3.7 branch)

mbr0wn:
Continuation of https://github.com/gnuradio/gnuradio/pull/2594/commits.

To quote @rear1019:

This is the last pull request to fix issue #1841. Changes in gr-digital and gr-trellis can be considered as API changes: User-visible behavior is modified. However, no code modifications in GNU Radio or OOT modules are required (as far as I can tell).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,22 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 Older Logs can be found in `docs/RELEASE-NOTES-*`.
 
+## UNRELESEASED
+
+### Changed
+#### Project Scope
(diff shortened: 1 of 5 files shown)


marcusmueller:
@mbr0wn It's still important to note the API change this includes, so a  release note warning would be necessary, I'll write a CHANGELOG entry shortly.

However, this breaks compilation on Ubuntu 14.04 (ugh, I know) right now. I'll look into it, but the error just makes no sense, looks like someone messed up their standard libraries.

marcusmueller:
ah, rear1019 added cstdint, which is C++11, but since we already have a workaround for that, employed said workaround, and --fixup'ed the commit.

marcusmueller:
This is ready for merging. I'll rebase it locally and force push it here, merge it and delete this branch

PR #3016, opened 2020-01-06
apply: Fix heap corruption bug in async_decoder

marcusmueller:
When using async_decoder with a CC decoder with k = 7 and no input byte packing,
the frame size of the underlying FEC decoder is set incorrectly to a too large
value. This causes out of bounds accesses which corrupt the heap.


mbr0wn:
CI passed, syncing master to maint-3.8.

PR #3020, opened 2020-01-06
digital: pfb clock sync: Remove deprecated set_taps() API call

mbr0wn:
This API call was deprecated a long time ago and contained a warning
that it will be removed, so we're now being true to our promise.
This also fixes a compiler error in SWIG *if* you have old headers
installed that SWIG will find (instead of the ones in the source tree).

Note: I saw this compiler bug and it drove me crazy. SWIG shouldn't be using installed headers :anger:

### DIFF
--- a/gr-digital/include/gnuradio/digital/pfb_clock_sync_ccf.h
+++ b/gr-digital/include/gnuradio/digital/pfb_clock_sync_ccf.h
@@ -193,18 +193,6 @@ public:
     virtual void update_taps(const std::vector<float>& taps) = 0;
 
     /*!
-     * Used to set the taps of the filters in the filterbank and
-     * differential filterbank.
-     *
-     * WARNING: this should not be used externally and will be moved
-     * to a private function in the next API.
-     */
-    virtual void
-    set_taps(const std::vector<float>& taps,
-             std::vector<std::vector<float>>& ourtaps,
-             std::vector<std::unique_ptr<gr::filter::kernel::fir_filter_ccf>>& ourfilter) = 0;
-
-    /*!
      * Returns all of the taps of the matched filter
      */
     virtual std::vector<std::vector<float>> taps() const = 0;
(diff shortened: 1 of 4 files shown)


rear1019:
I encountered this error yesterday as well. The root cause seems to be that SWIG does not support ``std::uniq_ptr`` [1][2]. I basically made the same changes as in this pull request to make the code compile, but it still does not compile (neither with your pull request):
```
/home/*/dev/gnuradio/build/gr-digital/swig/CMakeFiles/digital_swig2.dir/digital_swig2PYTHON_wrap.cxx: In function 'PyObject* _wrap_pfb_clock_sync_ccf_set_taps(PyObject*, PyObject*, PyObject*)':
/home/*/dev/gnuradio/build/gr-digital/swig/CMakeFiles/digital_swig2.dir/digital_swig2PYTHON_wrap.cxx:12669:15: error: 'class gr::digital::pfb_clock_sync_ccf' has no member named 'set_taps'
12669 |       (arg1)->set_taps((std::vector< float,std::allocator< float > > const &)*arg2,*arg3,*arg4);
      |
```
(repeated three more times at different locations)

It seems like SWIG still tries to wrap ``set_taps()`` and I have no idea why. This is probably relevant:
> This also fixes a compiler error in SWIG if you have old headers
installed that SWIG will find (instead of the ones in the source tree).

My system has SWIG 4 and a copy of GNU Radio is installed globally (package from distribution’s repo). 

[1] http://swig.org/Doc4.0/SWIGDocumentation.html#CPlusPlus11_general_purpose_smart_pointers
[2] http://swig.org/Doc3.0/SWIGDocumentation.html#CPlusPlus11_general_purpose_smart_pointers

mbr0wn:
@rear1019 Yeah if you have older headers in your system, SWIG will use those instead of the ones from the source tree. That's a separate issue. Took me quite a bit to figure that out -- I ran two bisects over a lot of commits before it dawned on me.

PR #3022, opened 2020-01-06
logging: (msg) to avoid logged statement taken apart by operator precedence

marcusmueller:
Oldie but a goldie:
Macro containing

     logger << "something " << msg << "\n";

will happily break code in hard-to-understand ways, since `<<` has
higher precedence than binary, comparison and logical operators.

I wouldn't have opened this PR if this hadn't cost me 10 precious minutes
before....

PR #3023, opened 2020-01-06
logger () (#3022 for maint-3.8)

marcusmueller:


michaelld:
Since this PR is the same as the recently merged #3022 just for maint-3.8 instead of master, and it's super-important, I'm going to go ahead and merge it.

PR #3026, opened 2020-01-06
gr-fec: Add QA test for buffer overflow in CC decoder

daniestevez:
Adds a QA test that tries to detect the kind of buffer overflows in the CC decoder that were fixed in #2965.

The decoder is run with different frame lengths, and several frames are tried to be decoded for each frame length. The test is considered successful if it doesn't segfault or abort.

### DIFF
--- a/gr-fec/python/fec/qa_fecapi_cc_buffer_overflow.py
+++ b/gr-fec/python/fec/qa_fecapi_cc_buffer_overflow.py
@@ -1,0 +1,58 @@
+#!/usr/bin/env python
+#
+# Copyright 2020 Free Software Foundation, Inc.
+#
+# This file is part of GNU Radio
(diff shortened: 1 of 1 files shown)


daniestevez:
As per the comments in #2965, I have added a simple test that reproduces the buffer overflow bug (at least in my system). Different frame sizes are tried to try to produce a segfault or abort (one of the problematic frame sizes is 1020).

I can confirm that the test fails when the fix in #2965 is not applied (the abort it produces is shown as failure in `make test`) and succeeds when the fix is applied.

I have limited the test coverage regarding frame sizes to prevent the test from taking too long.

So far this is very basic and only tests the decoder configuration with which I found problems, but it leaves the door open to trying other settings or decoders.

PR #3031, opened 2020-01-06
cmake: grc: Provide ENABLE_POSTINSTALL option

mbr0wn:
This option allows disabling post-install tasks. Currently, there is one
such task: The execution of `grc_setup_freedesktop install`.

For developers building from source, this is very convenient to do
during installation. However, for package maintainers, this is not
desired and should be executed separately. Maintainers of packages can
thus provide ENABLE_POSTINSTALL=OFF to avoid permission errors.

This continues bc0592b6, which defaulted to always running the script.

See the discussion in https://github.com/gnuradio/gnuradio/pull/2580/files.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -479,6 +479,12 @@ find_package(MPLIB)
 add_definitions(${MPLIB_DEFINITIONS})
 
 ########################################################################
+# Post-Install tasks are tasks that are usually not executed during
+# install, but for source builds, that's actually more convenient.
+########################################################################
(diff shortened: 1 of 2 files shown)


marcusmueller:
Generally in favour of this. If I was to write raw Makefiles, though, I'd handle it in the following way:

1. add a phony target `postinstall`
2. if given the `ENABLE_POSTINSTALL` option, add it as last step to  `install`
3. I could still manually run `make install`. 

I think the mental overhead for implementing that in CMake isn't worth it, tbh. So, merging this.

PR #3034, opened 2020-01-07
Backport to 3.8 (Jan 6 2019)

mbr0wn:
logging: (msg) to avoid logged statement taken apart by operator prec…edence

Oldie but a goldie:
Macro containing

     logger << "something " << msg << "\n";

will happily break code in hard-to-understand ways, since `<<` has
higher precedence than binary, comparison and logical operators.


michaelld:
This is the same as https://github.com/gnuradio/gnuradio/pull/3023 , which I just merged, so closing this one since it's no longer relevant.

PR #3037, opened 2020-01-07
pmt: Fix RuntimeError in pmt_to_python

### LINKED ISSUE #2961: Python unable to parse cons-wrapped-cons
Assume an existing dictionary of meta-data, **meta**, and a vector of complex samples, **out**.

If one tries to pass these bundled up in to a cons, and **meta** contains a cons itself, the receiver (if python) will be unable to parse the embedded cons.

**Somewhat-pseudo C code**:
```C++
meta = dict_add(meta, string_to_symbol('key'), cons(12.34, 56.78));
message_port_pub("out", cons(meta, out));
```
**Somewhat-pseudo Python code**
```python
bits = socket.recv()
pdu = pmt.deserialize_str(bits)
meta = pmt.to_python(pmt.car(pdu)) # Crashes
```

The stack of the crash hints the following:
>  RuntimeError: pmt_length: wrong_type : (12.34 . 56.78)

This appears to be coming from the call 
```python
for i in range(pmt.length(items))
```
of `pmt_to_dict`

**edit:**
Or just see pastebin link [here](https://pastebin.com/65m8JfmK) showing some python that'll crash it.

MileB:
Python3 now has a RuntimeError that's thrown during the pmt to_python function.
Instead of throwing this error, SWIG has been updated to throw a TypeError.
This allows us to keep the same behavior whereby we iterate over PMT types until the proper conversion is found

Resolves #2961

### DIFF
--- a/gnuradio-runtime/include/pmt/pmt.h
+++ b/gnuradio-runtime/include/pmt/pmt.h
@@ -89,7 +89,7 @@ public:
     exception(const std::string& msg, pmt_t obj);
 };
 
-class PMT_API wrong_type : public exception
+class PMT_API wrong_type : public std::invalid_argument
 {
 public:
     wrong_type(const std::string& msg, pmt_t obj);
--- a/gnuradio-runtime/lib/pmt/pmt.cc
+++ b/gnuradio-runtime/lib/pmt/pmt.cc
@@ -37,7 +37,7 @@ exception::exception(const std::string& msg, pmt_t obj)
 }
 
 wrong_type::wrong_type(const std::string& msg, pmt_t obj)
-    : exception(msg + ": wrong_type ", obj)
+    : invalid_argument(msg + ": wrong_type " + write_string(obj))
 {
 }
 
--- a/gnuradio-runtime/swig/gr_extras.i
+++ b/gnuradio-runtime/swig/gr_extras.i
@@ -19,6 +19,9 @@
     try {
         $action
     }
+    catch(std::invalid_argument &e) {
+        SWIG_exception(SWIG_TypeError, e.what());
+    }
     catch(std::exception &e) {
         SWIG_exception(SWIG_RuntimeError, e.what());
     }
--- a/gr-blocks/python/blocks/qa_hier_block2.py
+++ b/gr-blocks/python/blocks/qa_hier_block2.py
@@ -74,7 +74,7 @@ class test_hier_block2(gr_unittest.TestCase):
         nop1 = blocks.nop(gr.sizeof_int)
         nop2 = blocks.nop(gr.sizeof_int)
         hblock.connect(nop1, hblock)
-        self.assertRaises(RuntimeError,
+        self.assertRaises(TypeError,
             lambda: hblock.connect(nop2, hblock))
 
     def test_006_connect_invalid_src_port_neg(self):
@@ -82,7 +82,7 @@ class test_hier_block2(gr_unittest.TestCase):
                                 gr.io_signature(1,1,gr.sizeof_int),
                                 gr.io_signature(1,1,gr.sizeof_int))
         nop1 = blocks.nop(gr.sizeof_int)
-        self.assertRaises(RuntimeError,
+        self.assertRaises(TypeError,
             lambda: hblock.connect((hblock, -1), nop1))
 
     def test_005_connect_invalid_src_port_exceeds(self):
@@ -90,7 +90,7 @@ class test_hier_block2(gr_unittest.TestCase):
                                 gr.io_signature(1,1,gr.sizeof_int),
                                 gr.io_signature(1,1,gr.sizeof_int))
         nop1 = blocks.nop(gr.sizeof_int)
-        self.assertRaises(RuntimeError,
+        self.assertRaises(TypeError,
             lambda: hblock.connect((hblock, 1), nop1))
 
     def test_007_connect_invalid_dst_port_neg(self):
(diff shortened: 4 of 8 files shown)


mbr0wn:
@MileB we can merge this without a CLA (it's small enough), but if you want to submit more code, please consider submitting a CLA. Get in touch with ben ät gnuradio döt org.

MileB:
Looks like I need to update my email settings to notify me of things, sorry for the lack of replies on this.

I'll spin up my VM and try out the different versions of Python to vet your questions, @michaelld

MileB:
Okay, it looks like my original understanding of the problem was incorrect.

This problem **does** impact Python2. My test environments from the initial issue tracker were a little dirty. I've been porting a project from gnuradio 3.7 to gnuradio 3.8, including migrating from Python2 to Python3.

It looks like my notes on Python2 not having this issue were from testing on GNURadio 3.7. After putting together a clean environment of Py2, gnuradio on maint-3.8, the issue **is** present, having  the following from [commit 2f66412](https://github.com/gnuradio/gnuradio/commit/2f66412afaca4a42319c666958c89827edd8f2f3#diff-65170e7a507659f3b8d7cfe72f6dead9L127)

The point of this patch is to bubble up something that's caught as a `TypeError` or a `ValueError` so this branch flows properly. Right now, anything that bubbles up as a `std::exception` is going to crash this segment.

This patch works on both Python2 and Python3.

So, circling back to your original questions, @michaelld 

> Thanks for sleuthing through the details to come up with this change!
> 
> Some wonderings:
> 
> (1) Is this change valid for -any- Py3, or do you know if it starts at some specific Py3 version (e.g., Py35)?

It works on both Python2.7 and Python3.6.9. If you want me test a specific version I can, but I have no reason to suspect that any version in between them would cause issues.

> (2) Do we need to keep the `exception` version around too if just Py3, or do the wrong type errors all get moved to `invalid_argument` when using Py3?

If something bubbles up as `SWIG_RuntimeError` it'll crash [here](https://github.com/gnuradio/gnuradio/blob/master/gnuradio-runtime/python/pmt/pmt_to_python.py#L127) 

So it depends on what constitutes an error. For the `to_python` method, I feel that `wrong_type` should bubble up to something that'll let it roll along. I'm sure there's still a valid case for RunTime errors somewhere.

> (3) This change should work with GR39 since it will be Py3 only. Is there a reasonable way to handle mixed Py2/Py3 in GR38?

Apologies again for my error in my initial notes, this works fine on Python2 and Python3.

I've attached a simple python script [here](https://pastebin.com/65m8JfmK) that'll crash without this patch applied.

michaelld:
Testing on my setup (macOS 10.14, GR 3.7 latest release):
```
% python2.7 ~/Desktop/test_gr_pmt_crash.py
{'key1': (1, 2)}
some data
```
Is this expected?

michaelld:
Wondering if this is a GR38+ issue ...

MileB:
> Wondering if this is a GR38+ issue ...

Yes, the commit [here](https://github.com/gnuradio/gnuradio/commit/2f66412afaca4a42319c666958c89827edd8f2f3#diff-65170e7a507659f3b8d7cfe72f6dead9L127) is what introduced this new behavior, and this is on gnuradio 3.8

michaelld:
> @MileB we can merge this without a CLA (it's small enough), but if you want to submit more code, please consider submitting a CLA. Get in touch with ben ät gnuradio döt org.

@MileB did you start the CLA process yet? It does take some time ... as Martin noted, you don't it for this merge since it's small enough.

michaelld:
Tagging a few other GR devs to put some extra eyes on this, just to be safe!

MileB:
> > @MileB we can merge this without a CLA (it's small enough), but if you want to submit more code, please consider submitting a CLA. Get in touch with ben ät gnuradio döt org.
> 
> @MileB did you start the CLA process yet? It does take some time ... as Martin noted, you don't it for this merge since it's small enough.

I have not. I just happened to put together this patch out of necessity

mbr0wn:
@MileB Thanks for the extremely detailed report. I think this change is great. Any idea why this fails on all the CI builds? Does make test work for you?

MileB:
> @MileB Thanks for the extremely detailed report. I think this change is great. Any idea why this fails on all the CI builds? Does make test work for you?

Running `make test`, and `ctest --rerun-failed -V` are showing the same tests failing on my side. 
I noticed a few in particular (like `test_099` of `qa_ofdm_serializer_vcc.py`) are looking specifically for a `RuntimeError` to propagate up from a call... I could see this change being a likely reason for that to fail.

What direction makes most sense to you, the reviewers? Either...
 - I can spend time making these tests pass with the new error behavior
 - We could just accept `RuntimeError` as a valid reason to try new casts [here](../blob/master/gnuradio-runtime/python/pmt/pmt_to_python.py#L127), effectively reverting a portion of [this](https://github.com/gnuradio/gnuradio/commit/2f66412afaca4a42319c666958c89827edd8f2f3#diff-65170e7a507659f3b8d7cfe72f6dead9L127) commit
- Some third option?

mbr0wn:
@MileB sorry for the late response. Can you please update the failing tests. I think accepting RuntimeError here is a recipe for trouble, so let's just accept the API change. However, if we want to backport this, we'll have to add RuntimeError as a possible exception type in the except clause you mentioned.

michaelld:
I'm in agreement with @mbr0wn here: please fixup tests so that whatever passed before this changes still passes after this change, and hopefully more pass than before. No reversions in passing test because of this change!

MileB:
Wow. I really dorked up that rebase.

Please hold ...

MileB:
Soon as you're available, @mbr0wn or @michaelld, the tests are updated to be passing.

These were all impacted because the new behavior bubbles up std::invalid_argument as TypeError instead of a RuntimeError.

PR #3038, opened 2020-01-07
Improve gnuradio-runtime fixed-point math

ArtemPisarenko:
Currently it fails tests. Seems like sine/cosine output precision was degraded. Although I cannot see noticeable difference when examining signal via GUI Sink blocks. Maybe tests require too strict tolerance... Hoping for feedback to solve this problem...


ArtemPisarenko:
Hmm... Those tests I expected to fail are succeeded actually. (They failed in my local environment on low-end notebook which found to be misconfigured because of my attempts to optimize gnuradio build workflow causing great pain). It's just qa_agc which failed. But that's completely different story! Huh... I'll investigate that...

ArtemPisarenko:
Rebased to latest master and corrected 5e10c23 commit message where I did wrong statement

ThomasHabets:
I'll need to stare at the maths here a bit. But I'll get to it.

Note that it currently has a `CMakeLists.txt` conflict, so if you could rebase that'd be great.

ThomasHabets (approved):
I can't say I'm following the math, or know the "why" of before this change, but the results seem tested and correct.

One question though: Why [1,-1), why not [1,-1]?

ArtemPisarenko:
> One question though: Why [1,-1), why not [1,-1]?

That's my interpretation of how float-type value range maps to signed integer range in this math. Minimum is -2^N, but maximum is +2^N-1. See corresponding limit macros in stdint.h. I understand conversion operation this math does is equivalent to `integer = float * INT32_MAX`, so when `float` = `-1.0` then `integer` will not be `INT32_MIN`, but `INT32_MIN+1`. But maybe floating-point arithmetic has similar assymetry, where `-1.0*-1 != 1.0`. Not sure.

ArtemPisarenko:
I'm trying to rebase but encountered problem with building due to new pybind11 introduced in recent changes. I haven't dealt with neither old swig, nor new pybind11.
```
  Python bindings for fxpt.h are out of sync
  gnuradio-runtime/python/gnuradio/gr/bindings/CMakeLists.txt:85 (GR_PYBIND_MAKE_CHECK_HASH)
```
I didn't found any info about this except [porting guide](https://wiki.gnuradio.org/index.php/GNU_Radio_3.9_OOT_Module_Porting_Guide#Pybind11_Python_Bindings). But it doesn't explain how to regenerate (just update?) bindings for internal modules (i.e. non-OOT).

ThomasHabets:
You need to change the hash of any include file from `include/` directories (but not `lib` directories). You hash the file you changed using md5, and edit the corresponding python binding file to have this new value.

Here's the script I use. It takes as parameters the list of header files you changed, and it'll change the hash in the "binding file".

```
#!/usr/bin/bash

set -e

while [[ ! "$1" = "" ]]; do
  FILE="$1"
  BINDING="$(echo "$FILE" | sed -r 's,gr-([a-z-]+)/include/gnuradio/([a-z]+),gr-\1/python/\1/bindings,;s/[.]h$/_python.cc/')"
  MD5="$(md5sum "$FILE" | awk '{print $1}')"
  echo $FILE $MD5 $BINDING
  sed -i -r "s/(BINDTOOL_HEADER_FILE_HASH)\([a-z0-9]+\)/\1($MD5)/" "$BINDING" || true
  shift
done
```

ThomasHabets:
Aside from the closed/open range question that I have no idea about, looks good. In any case it looks like this is strictly an improvement over stepping *over* the -1 to 1 range.

ArtemPisarenko:
@ThomasHabets , thank you. I wasn't sure that it's enough to update hash only. So I tried to do it properly by temporal change of BINDTOOL_GEN_AUTOMATIC(0) to (1) in fxpt_python.cc which triggers regenerating of binding. It was a great pain again to get working build environment for updated master because of pybind. Even after I realized that regenerating has a problem of egg and chicken (i.e. build fails when it's not being run from already installed gnuradio), it was still hard to satisfy all required dependencies. But now I succeeded.

ThomasHabets (approved):

mormj:
This all seems to make sense to me, but tagging @marcusmueller - since this affects quite a bit of the code

mbr0wn:
Sorry to close, but this has had a lack of activity recently and needs a rebase. A DCO signoff would be good too. Please reopen when it's working.

PR #2778, opened 2019-08-30
Reintroduce default setting in qtgui_chooser

dl1ksv:
In gnuradio 3.7 it was possible to select an option as default option.
In 3.8 /3.9 the default option is always option 0.

This pr enables to choose a default option different from option 0

PR #3042, opened 2020-01-08
Chooser default value change in example flowgraphs

### LINKED ISSUE #3027: Multiply Matrix example flowgraph throws exception
Looks to be QT related, not necessarily related to the matrix multiply example itself.

```
Executing: /usr/bin/python3.6 -u /share/gnuradio/grdev/src/gnuradio/gr-blocks/examples/matrix_multiply_demo.py

  File "/share/gnuradio/grdev/src/gnuradio/gr-blocks/examples/matrix_multiply_demo.py", line 264, in <module>
    main()
  File "/share/gnuradio/grdev/src/gnuradio/gr-blocks/examples/matrix_multiply_demo.py", line 242, in main
    tb = top_block_cls()
  File "/share/gnuradio/grdev/src/gnuradio/gr-blocks/examples/matrix_multiply_demo.py", line 94, in __init__
    self._new_A_callback(self.new_A)
  File "/share/gnuradio/grdev/src/gnuradio/gr-blocks/examples/matrix_multiply_demo.py", line 93, in <lambda>
    self._new_A_callback = lambda i: Qt.QMetaObject.invokeMethod(self._new_A_combo_box, "setCurrentIndex", Qt.Q_ARG("int", self._new_A_options.index(i)))
ValueError: tuple.index(x): x not in tuple
```

mormj:
After the way the default value was configured for QT GUI Chooser, this did not propagate correctly in flowgraphs that use it.  

The new "default" value was set to whatever open 0 was set to, then option 0 was set to "0"

This fixes the example flowgraphs only, not the underlying problem that made the new behavior not update automatically

Fixes #3027

### DIFF
--- a/gr-blocks/examples/matrix_multiplexer.grc
+++ b/gr-blocks/examples/matrix_multiplexer.grc
@@ -1,6 +1,7 @@
 options:
   parameters:
     author: Martin Braun
+    catch_exceptions: 'True'
     category: Custom
     cmake_opt: ''
     comment: ''
@@ -25,6 +26,9 @@ options:
     title: Matrix Multiply/Multiplex Demo
     window_size: 1280, 1024
   states:
+    bus_sink: false
+    bus_source: false
+    bus_structure: null
     coordinate: [8, 12.0]
     rotation: 0
     state: enabled
@@ -43,6 +47,7 @@ blocks:
     label4: ''
     labels: '[]'
     num_opts: '3'
+    option0: ((1, 0), (0, 1))
     option1: ((0, 1), (1, 0))
     option2: ((.5, .5), (.5, .5))
     option3: '3'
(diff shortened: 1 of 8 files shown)


michaelld:
@mormj To get these changes, did you basically have the latest GRC "compile" them & then push those changes here? Thinking these weren't done by hand ... mostly just curious. I'm OK with them either way.

mormj:
The process that got me here was - 
- opened up the Matrix Multiply example and it hit an exception 
- figured out it was the change in default value #2778 that caused the change in behavior
- tested all the flowgraphs that use the chooser, and found 3 from that set that were adversely affected by the change in default value

I don't think more grc example flowgraphs need to be updated - unless we want to do a flush so they all have stored the current state variables that are stored in the grc rather than take the default values when they are first opened.  This would require a lot of manual testing (which we should probably do ahead of any release date)

michaelld:
@mormj does this PR need a backport to maint-3.8 (if not done already)?

mormj:
Yes - added into #3391

PR #3044, opened 2020-01-08
blocks: support saving of all tags on tag debug

igorauad:
For testing, it is often useful to check all tags ever received, rather than
solely the tags from the last call to the Tag Debug's work function. This patch
adds this option and a public method for activating it.

This is more appropriate for controlled test environments, where the vector of
tags is known not to grown indefinitely. Hence, the option is not exposed as a
block parameter.

### DIFF
--- a/gr-blocks/lib/tag_debug_impl.cc
+++ b/gr-blocks/lib/tag_debug_impl.cc
@@ -35,7 +35,8 @@ tag_debug_impl::tag_debug_impl(size_t sizeof_stream_item,
                  io_signature::make(1, -1, sizeof_stream_item),
                  io_signature::make(0, 0, 0)),
       d_name(name),
-      d_display(true)
+      d_display(true),
+      d_save_all(false)
 {
     set_key_filter(key_filter);
 }
@@ -45,7 +46,12 @@ tag_debug_impl::~tag_debug_impl() {}
 std::vector<tag_t> tag_debug_impl::current_tags()
 {
     gr::thread::scoped_lock l(d_mutex);
-    return d_tags;
+
+    if (d_save_all) {
+        return d_tags;
+    } else {
+        return d_work_tags;
+    }
 }
 
 int tag_debug_impl::num_tags()
@@ -56,6 +62,17 @@ int tag_debug_impl::num_tags()
 }
 
 void tag_debug_impl::set_display(bool d) { d_display = d; }
+
+void tag_debug_impl::set_save_all(bool s)
+{
+    gr::thread::scoped_lock l(d_mutex);
+
+    if (d_save_all && !s) {
+        d_tags.clear();
+    }
+
+    d_save_all = s;
+}
 
 void tag_debug_impl::set_key_filter(const std::string& key_filter)
 {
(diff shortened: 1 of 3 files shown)


gr-blocks/lib/tag_debug_impl.cc
```diff
@@ -57,7 +58,11 @@ tag_debug_impl::~tag_debug_impl() {}
 std::vector<tag_t> tag_debug_impl::current_tags()
 {
     gr::thread::scoped_lock l(d_mutex);
-    return d_tags;
+
+    if (d_save_all)
```
mbr0wn: Please don't skimp on the {}.

gr-blocks/lib/tag_debug_impl.cc
```diff
@@ -69,6 +74,13 @@ int tag_debug_impl::num_tags()
 
 void tag_debug_impl::set_display(bool d) { d_display = d; }
 
+void tag_debug_impl::set_save_all(bool s) {
+    if (d_save_all && !s)
```
mbr0wn:
```suggestion
    if (d_save_all && !s) {
```

gr-blocks/lib/tag_debug_impl.cc
igorauad: Updated

gr-blocks/lib/tag_debug_impl.cc
igorauad: Updated

igorauad:
> Very useful. Please make sure your code matches our clang-format settings, I think it might not. Also, we need your CLA :) Thanks for the submission!

OK @mbr0wn I've ran `clang-format` and pushed the updates.

Regarding the CLA, I will ping you once it is arranged.

igorauad:
@mbr0wn I have finally got the CLA assigned, which I just sent to Ben over email. So I will rebase my pending MRs and remove the CLA-needed label from them.

gr-blocks/lib/tag_debug_impl.cc
```diff
@@ -57,6 +63,15 @@ int tag_debug_impl::num_tags()
 
 void tag_debug_impl::set_display(bool d) { d_display = d; }
 
+void tag_debug_impl::set_save_all(bool s)
+{
+    if (d_save_all && !s) {
+        d_tags.clear();
```
mbr0wn: One more thing: Can you hold `d_setlock` while you access `d_tags`. Otherwise, this is not thread-safe.

gr-blocks/lib/tag_debug_impl.cc
```diff
@@ -109,6 +123,11 @@ int tag_debug_impl::work(int noutput_items,
                 sout << tag.value << std::endl;
             }
         }
+
+        if (d_save_all) {
+            for (const auto& tag : d_work_tags)
+                d_tags.push_back(tag);
```
mbr0wn: Same here.

mbr0wn:
@igorauad Thanks for the update. Other than the thread-safety, this is good and I think it's a great addition to the code.

gr-blocks/lib/tag_debug_impl.cc
igorauad: Thanks, @mbr0wn . Tag Debug already had mutex `d_mutex`, which is locked both in `work()` and `current_tags()`. It was missing on `set_save_all()` and I've just added it.

gr-blocks/lib/tag_debug_impl.cc
igorauad: I believe the `gr::thread::scoped_lock l(d_mutex);` in the beginning of the `work` function suffices.

igorauad:
> @igorauad Thanks for the update. Other than the thread-safety, this is good and I think it's a great addition to the code.

Thanks, @mbr0wn . Updated and rebased.

michaelld:
@mbr0wn please give this one last read-through & if you approve then please merge

PR #3046, opened 2020-01-08
Rotator-based Freq Shift Convenience Wrapper

ghostop14:
This yml adds a convenience-wrapper block to provide volk-accelerated frequency shifting based on the rotator block that doesn't require the user to "import math" and code the "2*Pi*freq/samp_rate" as the phase increment.  I don't think newbies know a rotator with a phase increment results in a frequency shift (I didn't for the longest time) and may be relying on less CPU-efficient means such as a signal source/multiply or a frequency xlating fir filter (the latter one-sample-at-a-time rotates as it decimates rather than leveraging volk to do a block).

### DIFF
--- a/gr-blocks/grc/blocks_freqshift_cc.block.yml
+++ b/gr-blocks/grc/blocks_freqshift_cc.block.yml
@@ -1,0 +1,34 @@
+id: blocks_freqshift_cc
+label: Frequency Shift
+category: '[Core]/Math Operators'
+
+parameters:
+-   id: sample_rate
+    label: Sample Rate
+    dtype: real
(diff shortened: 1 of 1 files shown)


ghostop14:
CLA has been submitted.

gr-blocks/grc/blocks_freqshift_cc.block.yml
```diff
@@ -0,0 +1,36 @@
+# auto-generated by grc.converter
```
mbr0wn:
```suggestion
```

ghostop14:
Updated.

PR #3047, opened 2020-01-08
Added FFT-based Low Pass Filter Convenience Wrapper

ghostop14:
Based on analyses such as Tom R's filter_timing.py and some advanced users preferring FFT-based LPF's over FIRs in some implementations, this convenience wrapper provides a drag-and-drop LPF option mirroring the FIR-based LPF rather than having to select an FFT Filter block and build the taps manually.  This block also provides significant performance improvements over the FIR-based filter for any meaningful tap lengths.


drmpeg:
Why is the commit for Rotator-based Freq Shift Convenience Wrapper included here?

ghostop14:
Hi drmpeg, guidance on the Slack coredevs channel was these minor items could be put in as a PR rather than a more complex grep.  Happy to shuffle it wherever the team agrees is the best place.

drmpeg:
Yes, but it's also included in PR https://github.com/gnuradio/gnuradio/pull/3046

ghostop14:
Hmmm, they were definitely separate PRs from different local isolated branches.  Can always close/redo the FFT one if it looks like something got banged up.

drmpeg:
I'll guess you based your fft_lpf branch on the freqshift branch instead of master.

ghostop14:
They should be separate PR's.  3047 should not include it, just 3046.  I'll close 3047 and re-submit just what was intended to be in 3047.

ghostop14:
I was thinking that might've been what happened when I went in to create the new branch. :)

PR #3050, opened 2020-01-08
Filter: Added FFT Low Pass Filter Convenience Wrapper

ghostop14:
Based on analyses such as Tom R's filter_timing.py and some advanced users preferring FFT-based LPF's over FIRs in some implementations, this convenience wrapper provides a drag-and-drop LPF option mirroring the FIR-based LPF rather than having to select an FFT Filter block and build the taps manually. This block also provides significant performance improvements over the FIR-based filter for any meaningful tap lengths.

(this is a close/resubmit of 3047 that seemed to include 3046 as well)


ghostop14:
CLA has been submitted.

PR #3051, opened 2020-01-08
runtime: Add terminate handling

mbr0wn:
```
This consists of the following changes:

1. Add a top_block parameter to control exception handling

This restores past behavior where the scheduler catches exceptions
raised in block threads, allowing flowgraphs to continue running after
the failure of an individual block. It also adds optional new behavior,
selected by setting catch_exceptions=False to the top block, which causes
exceptions to not be caught. In this mode of operation, a std::terminate
handler can be installed to print a stack trace before the flowgraph exits.

2. Add terminate_handler to top_block_impl

This adds a terminate_handler function that prints information
about any uncaught exceptions and also attempts to print a backtrace
for said exception after which it calls std::abort. The backtrace is
printed via libunwind, which is a new optional dependency. If libunwind
is not found/installed, then the terminate handler will still print
information about the exception and then exit.

Co-Authored-By: Scott Torborg <storborg@gmail.com>
```

This is a rebased version of https://github.com/gnuradio/gnuradio/pull/1894. Used @noc0lour 's rebase, went from there, but I had to squash or it would have been really awkward to rebase.

### DIFF
--- a/cmake/Modules/Findlibunwind.cmake
+++ b/cmake/Modules/Findlibunwind.cmake
@@ -1,0 +1,29 @@
+# - Try to find libunwind
+# Once done, this will define
+#
+#  libunwind_FOUND - system has libunwind
+#  libunwind_INCLUDE_DIRS - the libunwind include directories
+#  libunwind_LIBRARIES - link these to use libunwind
+
+include(LibFindMacros)
+
+# Use pkg-config to get hints about paths
+libfind_pkg_check_modules(libunwind_PKGCONF libunwind)
+
+# Include dir
+find_path(libunwind_INCLUDE_DIR
+  NAMES libunwind.h
+  PATHS ${libunwind_PKGCONF_INCLUDE_DIRS}
+)
+
+# Finally the library itself
+find_library(libunwind_LIBRARY
+  NAMES unwind
+  PATHS ${libunwind_PKGCONF_LIBRARY_DIRS}
+)
+
+# Set the include dir variables and the libraries and let libfind_process do the rest.
+# NOTE: Singular variables for this library, plural for libraries this this lib depends on.
+set(libunwind_PROCESS_INCLUDES ${libunwind_INCLUDE_DIR})
+set(libunwind_PROCESS_LIBS libunwind_LIBRARY)
+libfind_process(libunwind)
--- a/gnuradio-runtime/lib/terminate_handler.cc
+++ b/gnuradio-runtime/lib/terminate_handler.cc
@@ -1,0 +1,147 @@
+/* -*- c++ -*- */
+/*
+ * Copyright 2018 Free Software Foundation, Inc.
+ *
+ * This file is part of GNU Radio
+ *
+ * SPDX-License-Identifier: GPL-3.0-or-later
+ */
+
+#include <pmt/pmt.h>
+#include <boost/thread.hpp>
+#include <exception>
+#include <iostream>
+#include <regex>
+#include <stdexcept>
+
+#include <cxxabi.h>
+
+#ifdef HAVE_LIBUNWIND
+#define UNW_LOCAL_ONLY
+#include <libunwind.h>
+#include <cstdio>
+#include <cstdlib>
+#endif
+
+namespace gr {
+
+#ifdef HAVE_LIBUNWIND
+void print_backtrace()
+{
+    unw_cursor_t cursor;
+    unw_context_t context;
+
+    // Initialize cursor to current frame for local unwinding.
+    unw_getcontext(&context);
+    unw_init_local(&cursor, &context);
+
+    // Unwind frames one by one, going up the frame stack.
+    while (unw_step(&cursor) > 0) {
+        unw_word_t offset, pc;
+        unw_get_reg(&cursor, UNW_REG_IP, &pc);
+        if (pc == 0) {
+            break;
+        }
+        std::fprintf(stderr, "0x%lx:", pc);
+
+        char sym[256];
+        if (unw_get_proc_name(&cursor, sym, sizeof(sym), &offset) == 0) {
+            char* nameptr = sym;
+            int status;
+            char* demangled = abi::__cxa_demangle(sym, nullptr, nullptr, &status);
+            if (status == 0) {
+                nameptr = demangled;
+            }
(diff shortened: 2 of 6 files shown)


noc0lour:
Oh, woah, nice!

cmake/Modules/Findlibunwind.cmake
```diff
@@ -0,0 +1,29 @@
+# - Try to find libunwind
+# Once done, this will define
+#
+#  libunwind_FOUND - system has libunwind
+#  libunwind_INCLUDE_DIRS - the libunwind include directories
+#  libunwind_LIBRARIES - link these to use libunwind
+
+include(LibFindMacros)
```
michaelld: It looks like the source for this macro is here: https://github.com/Tronic/cmake-modules/blob/master/LibFindMacros.cmake . Is the license compatible? If so, does it make sense to update the file in GR to be closer to the current version?

michaelld: Why don't we use this module more? It's really cool!

cmake/Modules/Findlibunwind.cmake
```diff
@@ -0,0 +1,29 @@
...
+
+# Finally the library itself
+find_library(libunwind_LIBRARY
+  NAMES unwind
+  PATHS ${libunwind_PKGCONF_LIBRARY_DIRS}
+)
+
+# Set the include dir variables and the libraries and let libfind_process do the rest.
+# NOTE: Singular variables for this library, plural for libraries this this lib depends on.
+set(libunwind_PROCESS_INCLUDES libunwind_INCLUDE_DIR)
```
michaelld: Isn't this and the next line setting the variable to the string, not to the value of the variable found in the string? I know CMake allows this in some circumstances, but I don't think this is one of them ...

cmake/Modules/Findlibunwind.cmake
mbr0wn: Can you please check if the license is compatible? I'm not even sure what that means in the CMake context. Do we need to copy this into the source tree?

cmake/Modules/Findlibunwind.cmake
mbr0wn: Yeah I think you're right. Can you provide a suggestion (using the GH suggestion button)?

gnuradio-runtime/lib/terminate_handler.cc
```diff
@@ -0,0 +1,160 @@
...
+    unw_init_local(&cursor, &context);
+
+    // Unwind frames one by one, going up the frame stack.
+    while (unw_step(&cursor) > 0) {
+        unw_word_t offset, pc;
+        unw_get_reg(&cursor, UNW_REG_IP, &pc);
+        if (pc == 0) {
+            break;
+        }
+        std::fprintf(stderr, "0x%lx:", pc);
```
mbr0wn, line 58: This will also require some logging updates when @marcusmueller is done.

noc0lour:
This partially resolves the failing qa_uncaught_exception in  #3871 and #3530 as bonus

cmake/Modules/Findlibunwind.cmake
noc0lour: The files says it is public domain, so should be fine.

cmake/Modules/Findlibunwind.cmake
noc0lour, line 27:
```suggestion
set(libunwind_PROCESS_INCLUDES ${libunwind_INCLUDE_DIR})
```

mormj (approved):
Seems to be working well.
Tested with and without libunwind-dev installed - both compile and flowgraphs run.
Put a runtime exception in the work function of a block, and with catch_exceptions turned on, get a printout like:
```
gr::log :ERROR: thread_body_wrapper - ERROR thread[thread-per-block[2]: <block fft(4)>]: blah
```
Also, verified that it resolves the qa_uncaught_exceptions test case

KyyB:
does the changes above resolve the payload test fails:

	 19 - qa_uncaught_exception (Failed)
	197 - qa_header_payload_demux (Failed)

I have been getting this error when running make test. Any ideas on this?

marcusmueller (approved):

marcusmueller:
CI seems to be stuck on the repo_push_runner trying to create the source archive, the successes of the builders seem to be stale

gnuradio-runtime/lib/terminate_handler.cc
```diff
@@ -0,0 +1,147 @@
...
+ */
+
+#include <pmt/pmt.h>
+#include <boost/thread.hpp>
+#include <exception>
+#include <iostream>
+#include <regex>
+#include <stdexcept>
+
+#include <cxxabi.h>
```
gvanem, line 17: Where are MSVC/clang-cl supposed to pick up this file from?
Surely it should be:
```diff
--- a/terminate_handler.cc 2020-12-19 16:41:48
+++ b/terminate_handler.cc 2020-12-19 17:18:19
@@ -14,7 +14,9 @@
 #include <regex>
 #include <stdexcept>

+#ifdef __GNUC__
 #include <cxxabi.h>
+#endif
```

gnuradio-runtime/lib/terminate_handler.h
```diff
@@ -0,0 +1,19 @@
...
+ * SPDX-License-Identifier: GPL-3.0-or-later
+ */
+
+#ifndef INCLUDED_GR_TERMINATE_HANDLER_H
+#define INCLUDED_GR_TERMINATE_HANDLER_H
+
+#include <gnuradio/api.h>
+
+namespace gr {
+void install_terminate_handler() GR_RUNTIME_API;
```
gvanem, line 16: Again some backward GCC-stuff. This doesn't create a valid `dllexport` directive. This does:
```diff
--- a/terminate_handler.h 2020-12-19 16:41:48
+++ b/terminate_handler.h 2020-12-19 17:19:38
@@ -13,7 +13,7 @@
 #include <gnuradio/api.h>

 namespace gr {
-void install_terminate_handler() GR_RUNTIME_API;
+void GR_RUNTIME_API install_terminate_handler();
 }

 #endif /* INCLUDED_TERMINATE_HANDLER_H */
```

PR #3055, opened 2020-01-09
gr-digital/costas_loop: use std::norm for |z|^2

japm48:
Use `std::norm(z)` instead of `(z*std::conj(z)).real()`. This should be faster (except for compiler optimizations) as the imaginary part is not computed and it is also easier to read.
The only downside is that the poorly named `std::norm` can lead to confusion.

This is based on fdb80ea8d32497a358b8787f47f1c70bcf18bb21.


mbr0wn:
@japm48 do you have a CLA with us? We can merge this without a CLA, it's small enough, but maybe you have future contributions which require a CLA so it's good to get that rolling.

japm48:
I still don't have the CLA signed, but I will do it, as I plan to do nontrivial contributions. If you prefer to wait to accept the PR until I've finished, then it's ok.

PR #3056, opened 2020-01-10
VOLK: update submodule pointer to v2.1.0

argilo:

The Rotator block is currently broken in GNU Radio 3.8. It was fixed in https://github.com/gnuradio/volk/pull/278. Now that VOLK 2.1.0 has been released, I propose to update the maint-3.8 branch to VOLK 2.1.0 to pick up the fix.

/cc @michaelld @devnulling


argilo:
@mbr0wn Could you elaborate on what the rules are for the VOLK dependency? What is the path forward here?

michaelld:
> We can't just update the VOLK version on maint-3.8. If the rotator is broken, we'll need to come up with a workaround to fix it.

@mbr0wn We have a test for that: `blocks_qa_rotator` ... yes? CI looks good so far, which I'm taking to mean that this GR's usage of the Volk API is compatible with the new Volk API. Or ... am I missing something?

michaelld:
@argilo We're allowing GR GIT master to be any Volk & it might not even build cleanly -- we'll fix it if not! For `maint-3.8` and `maint-3.7`, we require that the GR codebase build cleanly with the provided Volk submodule. I think those are our only conditions.

argilo:
I'm closing the off since the problem was resolved in https://github.com/gnuradio/gnuradio/pull/3062 instead.

PR #3062, opened 2020-01-11
gr-blocks: work around broken rotator in VOLK 2.0.0

argilo:

Replaces https://github.com/gnuradio/gnuradio/pull/3056.

Since GNU Radio 3.8 appears to be stuck with VOLK 2.0.0's broken rotator implementation, I propose to work around the problem by not using VOLK in the rotator block. It seems to me that it's better to have a slow rotator block than one that doesn't work at all.

This workaround is not needed on the master branch, since it already uses VOLK 2.1.0, which has a functioning rotator kernel.

@mbr0wn @michaelld @devnulling

### DIFF
--- a/gr-blocks/lib/rotator_cc_impl.cc
+++ b/gr-blocks/lib/rotator_cc_impl.cc
@@ -59,7 +59,7 @@ int rotator_cc_impl::work(int noutput_items,
     const gr_complex* in = (const gr_complex*)input_items[0];
     gr_complex* out = (gr_complex*)output_items[0];
 
-#if 0
+#if 1
       for (int i=0; i<noutput_items; i++)
       	out[i] = d_r.rotate(in[i]);
 #else


argilo:
> (2) GR Leads will be discussing if/how to update the Volk submodule for maint-3.8 (#3056) this coming week.

I'm happy to hear it. I'll hold off until that's decided.

mbr0wn:
@argilo Yeah please don't spend time until we've decided.

gr-blocks/lib/rotator_cc_impl.cc
```diff
@@ -59,7 +59,7 @@ int rotator_cc_impl::work(int noutput_items,
     const gr_complex* in = (const gr_complex*)input_items[0];
     gr_complex* out = (gr_complex*)output_items[0];
 
-#if 0
+#if 1
```
mbr0wn: What would be good here is if we could tell the VOLK version. Sadly, VOLK doesn't have a version.h file. We could get it from the GNU Radio cmake. Then, we could if based on if the VOLK version has the correct rotator.

mbr0wn:
The verdict: maint-3.8 will have to work with VOLK 2.0.0 (but could also work with newer VOLK versions). To guarantee that, we merge this.

Future VOLK versions will have a VOLK_VERSION macro so we can be smarter at compile time.

smunaut:
!?!?

How is there not a VOLK 2.0.1 that fixes that bug ...

Gnuradio is not the only thing that use Volk, do you expect everyone to realize most people will be using a broken volk because it's distributed by default and implement the work around ?

argilo:
I agree that VOLK should have been patched. This workaround comes with a serious performance penalty and affects all users of GNU Radio 3.8, even if they're using VOLK 2.1 (like Debian testing already is).

smunaut:
Yeah. First off I don't see exactly why it can't be bumped to 2.1.0 in GR ... when there are such arbitrary decisions, it'd be nice to have the rationale explained or linked to ...

But even then :
 - Fix the bug in a maintenance branch of 2.0.0
 - Also add a #define of VOLK_VERSION 0x020001 in it. ( Also, in parallel, add the VOLK_VERSION define in volk master so it's included for all new releases. )
 - Tag and relase as 2.0.1
 - Update the GR maint-3.8 volk submodule to it
 - In rotator, check if VOLK_VERSION is both defined and greater or equal to 0x020001 to enable / disable VOLK usage in rotator block and/or the optimized FIR filter.

This way, new GR 3.8 buils will always work. If you have a broken installed system volk, the work around is kept in place. If you have a newed fixed volk or no volk installed (hence using the sub module) you get more performance.

I don't see any downside here and I don't see why this obvious solution hasn't been implemented in the last 6 months.

PR #3065, opened 2020-01-11
audio: VOLKize alsa_sink 

### LINKED ISSUE #3025: Alsa sink: use VOLK to do the float->audio sample format conversion
the work_s32, work_s16,… functions do things in unaccelerated code, but should be using the appropriate VOLK functions instead – this might not be a large issue, since audio rates are by definition low, but it reduces code duplication and has potential for speedup, especially on embedded platforms.

grant-cox:
Will close #3025. This is a placeholder for future commits, I wanted to get @marcusmueller 's opinion on the implementation. I've worked on the `work_s32` function so far.

Use `volk_32f_s32f_convert_32i` to speed up converting each `float` channel to `int32_t`. Use `volk::vector` to hold each `int32_t` channel buffer until output buffer is built. I'll remove those long comments after I hear back.

### DIFF
--- a/gr-audio/lib/alsa/alsa_sink.cc
+++ b/gr-audio/lib/alsa/alsa_sink.cc
@@ -369,7 +369,16 @@ int alsa_sink::work_s32(int noutput_items,
     unsigned int nchan = input_items.size();
     const float** in = (const float**)&input_items[0];
     sample_t* buf = (sample_t*)d_buffer;
-    int bi;
+
+    /*
+     * vector of vectors to keep each channel after conversion
+     * to int32_t. These are then placed in the buffer in the
+     * standard WAV file format:
+     * [chan0samp0, chan1samp0, ..., chanNsamp0, chan0samp1, ...]
+     * */
+    volk::vector<volk::vector<sample_t>> buf_channels(
+        nchan, volk::vector<sample_t>(d_period_size));
+
     int n;
 
     unsigned int sizeof_frame = nchan * sizeof(sample_t);
@@ -377,16 +386,24 @@ int alsa_sink::work_s32(int noutput_items,
 
     for (n = 0; n < noutput_items; n += d_period_size) {
         // process one period of data
-        bi = 0;
-        for (unsigned int i = 0; i < d_period_size; i++) {
-            for (unsigned int chan = 0; chan < nchan; chan++) {
-                buf[bi++] = (sample_t)(in[chan][i] * scale_factor);
-            }
-        }
+        for (unsigned int chan = 0; chan < nchan; chan++)
+            volk_32f_s32f_convert_32i(
+                buf_channels[chan].data(), in[chan], scale_factor, d_period_size);
 
         // update src pointers
         for (unsigned int chan = 0; chan < nchan; chan++)
             in[chan] += d_period_size;
+
+        /*
+         * build buf in the standard WAV file format:
+         * [chan0samp0, chan1samp0, ..., chanNsamp0, chan0samp1, ...]
+         * */
+        unsigned int sample = 0;
+        while (sample < d_period_size * nchan)
+            for (unsigned int channel = 0; channel < nchan; channel++) {
+                buf[sample] = buf_channels[channel][sample];
+                sample++;
+            }
 
         if (!write_buffer(buf, d_period_size, sizeof_frame))
             return -1; // No fixing this problem.  Say we're done.
--- a/gr-audio/lib/alsa/alsa_sink.cc
+++ b/gr-audio/lib/alsa/alsa_sink.cc
@@ -264,9 +264,15 @@ bool alsa_sink::check_topology(int ninputs, int noutputs)
     if (err < 0)
         bail("snd_pcm_sw_params", err);
 
-    d_buffer_size_bytes = d_period_size * nchan * snd_pcm_format_size(d_format, 1);
+    unsigned int output_buffer_sample_size = snd_pcm_format_size(d_format, 1);
+
+    d_buffer_size_bytes = d_period_size * nchan * output_buffer_sample_size;
 
     d_buffer = new char[d_buffer_size_bytes];
+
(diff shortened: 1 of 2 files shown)


gr-audio/lib/alsa/alsa_sink.cc
```diff
@@ -369,25 +369,42 @@ int alsa_sink::work_s32(int noutput_items,
...
         // update src pointers
         for (unsigned int chan = 0; chan < nchan; chan++)
             in[chan] += d_period_size;
 
+        /*
+         * build buf in the standard WAV file format:
+         * [chan0samp0, chan1samp0, ..., chanNsamp0, chan0samp1, ...]
+         * */
+        unsigned int sample = 0;
+        while (sample < d_period_size * nchan)
```
grant-cox: Is there a better way to do this output loop? From what I know of compiler optimization, I feel like this loop will be difficult for the compiler to optimize.

gr-audio/lib/alsa/alsa_sink.cc
```diff
@@ -369,25 +369,42 @@ int alsa_sink::work_s32(int noutput_items,
...
 
     for (n = 0; n < noutput_items; n += d_period_size) {
         // process one period of data
-        bi = 0;
-        for (unsigned int i = 0; i < d_period_size; i++) {
-            for (unsigned int chan = 0; chan < nchan; chan++) {
-                buf[bi++] = (sample_t)(in[chan][i] * scale_factor);
-            }
-        }
+        for (unsigned int chan = 0; chan < nchan; chan++)
```
mbr0wn:
```suggestion
        for (unsigned int chan = 0; chan < nchan; chan++) {
```

gr-audio/lib/alsa/alsa_sink.cc
```diff
@@ -369,25 +369,42 @@ int alsa_sink::work_s32(int noutput_items,
...
     sample_t* buf = (sample_t*)d_buffer;
-    int bi;
+
+    /*
+     * vector of vectors to keep each channel after conversion
+     * to int32_t. These are then placed in the buffer in the
+     * standard WAV file format:
+     * [chan0samp0, chan1samp0, ..., chanNsamp0, chan0samp1, ...]
+     * */
+    volk::vector<volk::vector<sample_t>> buf_channels(
```
mbr0wn: This should probably be allocated outside of work()

grant-cox:
@mbr0wn thanks for the feedback. @marcusmueller mentioned VOLKizing the type casting in alsa_sink to unify the code base and pass some complexity to volk. Marcus mentioned in the issue that his won't be a huge speedup given the low rate of audio data but it could improve performance in embedded systems.

I only looked at one of the work functions here -- if you all think this method is a good idea, I'll move on with the rest and update the branch. Thanks!

gr-audio/lib/alsa/alsa_sink.cc
grant-cox: I'll put that in the constructor / initialization. Thanks

gr-audio/lib/alsa/alsa_sink.cc
marcusmueller: I think this will be just fine, actually. The point is: that's a very common pattern, and it will end up being mainly memory-bandwidth limited, and this minimizes random memory access.

gr-audio/lib/alsa/alsa_sink.cc
```diff
@@ -369,25 +369,42 @@ int alsa_sink::work_s32(int noutput_items,
...
     for (n = 0; n < noutput_items; n += d_period_size) {
         // process one period of data
-        bi = 0;
-        for (unsigned int i = 0; i < d_period_size; i++) {
-            for (unsigned int chan = 0; chan < nchan; chan++) {
-                buf[bi++] = (sample_t)(in[chan][i] * scale_factor);
-            }
-        }
+        for (unsigned int chan = 0; chan < nchan; chan++)
+            volk_32f_s32f_convert_32i(
```
marcusmueller: You should include `volk.h` in this .cc: "Include what you use" is a very desirable pattern.

gr-audio/lib/alsa/alsa_sink.h
```diff
@@ -29,6 +29,7 @@
 
 #include <gnuradio/audio/sink.h>
 #include <alsa/asoundlib.h>
+#include <volk/volk_alloc.hh>
```
marcusmueller: Move this to the .cc; you don't need any of the things declared in that header in the public header for the ALSA sink, but you need it in the implementation :)

gr-audio/lib/alsa/alsa_sink.cc
grant-cox: Sounds good. Thanks!

gr-audio/lib/alsa/alsa_sink.h
grant-cox: Sounds good. Thanks!

gr-audio/lib/alsa/alsa_sink.cc
grant-cox: @mbr0wn I was looking into moving this to the constructor but I'm not sure about how to do it given that the output of the work functions are either `int32_t` or `int16_t`. I don't see anywhere in the constructor that knows the output type at initialization (it's determined in `check_topology()`). Do you have any ideas for this? Hope I'm missing something.

mbr0wn: Nothing wrong with allocating in check_topology(). I think that gets called more than once potentially, so you'll need to check if it's allocated already there.

grant-cox:
Do not merge this at the moment. If you see an issue, feel free to comment. I am working on quantization errors in the output.

grant-cox:
Unfortunately, my Master’s courses have picked up pace and I have not been able to look at the Alsa sink between class and my work schedule. I would like to look at it another time, but I think I should close my pull request for now to reduce the load on reviewers. Thanks for all the help that was given.

PR #3067, opened 2020-01-12
uhd: Move AGC-related calls to fix TypeError

haakov:
Fixes https://github.com/gnuradio/gnuradio/issues/3039

I believe I messed up the first attempt because the AGC calls don't exist in 3.7.

Someone with access to a USRP should probably test this more thoroughly to make sure it's right this time :)


drmpeg:
Looks good here. Tested on B210 with both sink and source blocks.

michaelld:
I tagged @mbr0wn to get his review, him being a UHD guru and all :)

PR #3068, opened 2020-01-13
Key filter zmq (v2)

### LINKED ISSUE #2236: gr-zeromq  subscriber block does not support filtering by key
The gr-zeromq subscriber block does not support subscribing to a specific key in a multi-part message. 
This is useful if different messages are transmitted to the same port.

talih0:
Fixes #2236

### DIFF
--- a/gr-zeromq/grc/zeromq_pub_sink.block.yml
+++ b/gr-zeromq/grc/zeromq_pub_sink.block.yml
@@ -35,6 +35,10 @@ parameters:
     dtype: int
     default: '-1'
     hide: ${ ('part' if hwm == -1 else 'none') }
+-   id: key
+    label: Filter Key
+    dtype: string
+    default: ''
 
 inputs:
 -   domain: stream
@@ -44,7 +48,7 @@ inputs:
 templates:
     imports: from gnuradio import zeromq
     make: zeromq.pub_sink(${type.itemsize}, ${vlen}, ${address}, ${timeout}, ${pass_tags},
-        ${hwm})
+        ${hwm}, ${key})
         
 cpp_templates:
     includes: [ '#include <gnuradio/zeromq/pub_sink.h>' ]
--- a/gr-zeromq/include/gnuradio/zeromq/pub_sink.h
+++ b/gr-zeromq/include/gnuradio/zeromq/pub_sink.h
@@ -22,11 +22,17 @@ namespace zeromq {
  * \ingroup zeromq
  *
  * \details
- * This block acts a a streaming sink for a GNU Radio flowgraph
- * and writes its contents to a ZMQ PUB socket.  A PUB socket may
+ * This block acts as a streaming sink for a GNU Radio flowgraph
+ * and writes its contents to a ZMQ PUB socket. A PUB socket may
  * have subscribers and will pass all incoming stream data to each
- * subscriber.  Subscribers can be either another gr-zeromq source
- * block or a non-GNU Radio ZMQ socket.
+ * subscriber with a matching key. If the publisher's key is set to
+ * "GNURadio", the following example subscriber keys will match: "G",
+ * "GN", .., "GNURadio". In other words, the subscriber must contain
+ * the first set of characters from the publisher's key. If the subscriber
+ * sets an empty key "", it will accept all input messages from the
+ * publisher (including the key itself if one is set). Subscribers
+ * can either be another gr-zeromq source block or a non-GNU Radio
+ * ZMQ socket.
  */
 class ZEROMQ_API pub_sink : virtual public gr::sync_block
 {
@@ -42,13 +48,15 @@ public:
      * \param timeout  Receive timeout in milliseconds, default is 100ms, 1us increments.
      * \param pass_tags Whether sink will serialize and pass tags over the link.
      * \param hwm High Watermark to configure the socket to (-1 => zmq's default)
+     * \param key Prepend a key/topic to the start of each message (default is none)
      */
     static sptr make(size_t itemsize,
                      size_t vlen,
                      char* address,
                      int timeout = 100,
                      bool pass_tags = false,
-                     int hwm = -1);
+                     int hwm = -1,
+                     const std::string& key = "");
 
     /*!
      * \brief Return a std::string of ZMQ_LAST_ENDPOINT from the underlying ZMQ socket.
--- a/gr-zeromq/lib/base_impl.cc
+++ b/gr-zeromq/lib/base_impl.cc
@@ -19,8 +19,13 @@
 namespace gr {
 namespace zeromq {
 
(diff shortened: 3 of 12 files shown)


mbr0wn:
Can you please fix the formatting:
```
The following files have formatting issues:
* ./gr-zeromq/lib/base_impl.cc
* ./gr-zeromq/lib/sub_source_impl.cc
* ./gr-zeromq/lib/pub_sink_impl.cc
* ./gr-zeromq/lib/base_impl.h
##[error]Docker run failed with exit code 1
```

All  you need to do is run clang-format. The PR formatting fix will tell you if it worked.

talih0:
> Can you please fix the formatting:
> 
> ```
> The following files have formatting issues:
> * ./gr-zeromq/lib/base_impl.cc
> * ./gr-zeromq/lib/sub_source_impl.cc
> * ./gr-zeromq/lib/pub_sink_impl.cc
> * ./gr-zeromq/lib/base_impl.h
> ##[error]Docker run failed with exit code 1
> ```
> 
> All you need to do is run clang-format. The PR formatting fix will tell you if it worked.

Sorry about that. I added a commit to fix this. 
Let me know if you need the changes rebased into each separate commit.

talih0:
@devnulling would you be able to help review this?

gr-zeromq/python/zeromq/qa_zeromq_sub.py
```diff
@@ -61,6 +61,7 @@ def test_001 (self):
     def test_002 (self):
         vlen = 10
 
+
```
michaelld: remove this extraneous CR/LF

gr-zeromq/include/gnuradio/zeromq/pub_sink.h
```diff
@@ -54,13 +58,15 @@ class ZEROMQ_API pub_sink : virtual public gr::sync_block
...
+     * \param key Prepend a key/topic to the start of each message (default is none)
      */
     static sptr make(size_t itemsize,
                      size_t vlen,
                      char* address,
                      int timeout = 100,
                      bool pass_tags = false,
-                     int hwm = -1);
+                     int hwm = -1,
+                     std::string key = "");
```
michaelld: should all of the `std::string` arguments be `const`? A reference (`std::string&`)? From a C++ perspective, I'd use both to improve the code. But, I tend to like 'const' in arguments and I like passing by reference where possible since it's (1) guaranteed value (unlike a pointer); and (2) faster than copying (a class or struct directly).

talih0: @michaelld Thanks for reviewing, I'll make the changes and resend

talih0:
ping

michaelld:
Yes, looks like we do have a CLA since a big PR of yours was merged a while ago ... so nothing more needed there.

michaelld:
I've tagged a few more GR developers to please review this PR so we can get it dealt with. Apologies for the delay; somehow this just dropped out of my queue without me knowing it!

talih0:
> I've tagged a few more GR developers to please review this PR so we can get it dealt with.

sounds good, thanks

gr-zeromq/grc/zeromq_pub_sink.block.yml
```diff
@@ -35,6 +35,11 @@ parameters:
     dtype: int
     default: '-1'
     hide: ${ ('part' if hwm == -1 else 'none') }
+-   id: key
+    label: Filter Key
+    dtype: string
+    default: ''
+    hide: ${ ('part' if key == '' else 'none') }
```
devnulling: @talih0 I'm trying to test this but the `key` field is not displaying when dragging in any of the ZMQ blocks. It would probably be best to just remove the logic to hide the field and display it all the time.

Ignore the color theme, but below is what I'm seeing when building your branch with this PR. 

zmq2
zmq1

talih0: @devnulling Thanks for looking over the patch.
Strange that the key option doesn't appear. Looks fine on mine. 

I'll remove the auto hide option just in case and redo the PR. 

Thanks
screenshot_pub_sink

devnulling: @talih0 It would be great if you could also squash all these commits into a single commit. 

Something like below would be sufficient:

"gr-zeromq: Add optional key filtering"

talih0: Thanks @devnulling, I made the change you suggested and squashed the commits

gr-zeromq/include/gnuradio/zeromq/pub_sink.h
```diff
@@ -25,7 +25,11 @@ namespace zeromq {
  * This block acts a a streaming sink for a GNU Radio flowgraph
  * and writes its contents to a ZMQ PUB socket.  A PUB socket may
  * have subscribers and will pass all incoming stream data to each
- * subscriber.  Subscribers can be either another gr-zeromq source
+ * subscriber with a matching key. If the publisher's key is set to
+ * "GNURadio", the following example subscriber keys will match: "G",
+ * "GN", .., "GNURadio", and "". In other words, the subscriber must
```
michaelld: I don't think "" will work here. As noted here & in the code, the key size has to be > 0, so all the rest might work but I don't think "" will. Yes? If so, please correct the comment.

gr-zeromq/lib/base_impl.cc
```diff
@@ -192,6 +210,23 @@ bool base_source_impl::load_message(bool wait)
     d_socket->recv(&d_msg);
 #endif
 
+    /* Throw away key and get the first message. Avoid blocking if a multi-part
+     * message is not sent */
+    if (d_key.size() > 0 && !more) {
+        int64_t is_multipart;
+        d_socket->getsockopt(ZMQ_RCVMORE, &is_multipart, &more_len);
+
```
michaelld: guessing there is an extra whitespace here

gr-zeromq/include/gnuradio/zeromq/pub_sink.h
talih0: Yes, I'll modify the docs to make it clearer.

The empty string is a bit of special case: If a publisher inserts a key, but the subscriber sets the key to an empty string, the key will end up being output as data on the subscriber.
Sometimes this will give garbage data, but it's more likely that the exception in gr-zeromq/lib/base_impl.cc:245 will be thrown because key.size() will not be a multiple of sizeof(output_data_type).

gr-zeromq/lib/base_impl.cc
talih0: thanks, I'll remove the extra line

talih0:
ping

michaelld:
Reads well. Commit format is good. Passes all CI, including the new tests it provides. Let's please get at least 1 more review.

talih0:
thanks!

PR #3069, opened 2020-01-13
Changed numpy to scipy since fftpack is in scipy

shanejohnpaul:
fftpack is present in scipy not numpy

### DIFF
--- a/gr-digital/examples/example_timing.py
+++ b/gr-digital/examples/example_timing.py
@@ -39,7 +39,7 @@ except ImportError:
     print("Error: could not from matplotlib import pyplot (http://matplotlib.sourceforge.net/)")
     sys.exit(1)
 
-from numpy.fft import fftpack
+from scipy import fftpack
 
 class example_timing(gr.top_block):
     def __init__(self, N, sps, rolloff, ntaps, bw, noise,


mbr0wn:
Did you get an error with the original code? `from numpy.fft import fftpack` works fine for me, and unlike scipy, numpy is an actual dependency of GNU Radio.

shanejohnpaul:
Yeah. It said that it cannot import fftpack from numpy.fft. But with the scipy version, everything works fine. Does gnuradio require a specific numpy version?

velichkov:
@mbr0wn,

It turns out the numpy's `fftpack` was replaced with `pocketfft` in https://github.com/numpy/numpy/pull/11888, https://github.com/numpy/numpy/commit/062dc8f831e695c2c179482da35affcc6045b4f3 so no `fftpack` in versions >  v1.17.0rc1 

The `scipy` is actually used in few places although there are no checks in cmake's files.

https://github.com/gnuradio/gnuradio/blob/96a5e82123ddc8a3ea6ced7bb903cafacd5113fb/gr-filter/python/filter/design/filter_design.py#L53
https://github.com/gnuradio/gnuradio/blob/96a5e82123ddc8a3ea6ced7bb903cafacd5113fb/gr-filter/python/filter/design/fir_design.py#L23
https://github.com/gnuradio/gnuradio/blob/96a5e82123ddc8a3ea6ced7bb903cafacd5113fb/gr-trellis/python/trellis/fsm_utils.py#L33

mbr0wn:
Yeah scipy is one of those optional dependencies. Scripts using scipy are supposed to check for its existence and print a useful error on ImportError.

velichkov:
The same problem in `gr-filter/python/filter/design/filter_design.py` was fixed in https://github.com/gnuradio/gnuradio/pull/2755 by importing `pocketfft` in case importing of `fftpack` fails.

The `fftpack` is being used in:
- gr-utils/python/utils/plot_fft_base.py
- gr-dtv/examples/atsc_ctrlport_monitor.py
- gr-digital/examples/example_timing.py

devnulling:
@shanejohnpaul Would you want to update this PR to behave the same as what the solution was in https://github.com/gnuradio/gnuradio/pull/2755/files ? 

Also we will need a CLA from you, not specifically for this minor change, but it will streamline PRs that may have more LOC in the future. Please email bhilburn at gnuradio.org to get the CLA started.

shanejohnpaul:
@devnulling Yes, it should behave the same. The `fftpack` is in `scipy` and not `numpy.fft`, and that is all I have changed.

I'm new to PRs, so I have no idea about CLAs.

drmpeg:
It should also be noted that the polar code QA tests fail during "make test" without scipy.

marcusmueller:
I'm really no fan of depending on `scipy` just because numpy's naming changed...

michaelld:
I made an issue #3235  for this topic. Agreed we don't want to promote SciPy to being a configure requirement. There are better ways ...

mbr0wn:
@shanejohnpaul based on our discussions in #3280 and #3290, it turns out this is the wrong fix. We need to call straight into numpy.fft. I'll close this PR, but still many thanks for submitting

PR #3070, opened 2020-01-13
Maint 3.8

gaspar1987:
image

hello i want to give you the solution problem solve it.but can't make it work. give the pictures that scares him.


michaelld:
@gaspar1987 Please open an Issue instead of a Pull Request with your query. Or, better yet, query the GR discussion list < https://lists.gnu.org/mailman/listinfo/discuss-gnuradio >!

Issue #3005, opened 2020-01-04
Make Volk a dependency / remove as submodule 


devnulling:
Place holder for 3.9 Project

- Make Volk a dependency 
- Remove as submodule

marcusmueller:
closed by #3072

PR #3072, opened 2020-01-13
Remove VOLK as a submodule

mbr0wn:
This removes the volk/ submodule pointer and updates the CMake to detect
VOLK like any other dependency. The VOLK_MIN_VERSION CMake variable is
added (and set to 2.1.0).

The GR_VOLK_LIB variable is replaced with Volk::volk everywhere.
The VOLK_INSTALL_LIBRARY_DIR and VOLK_INSTALL_INCLUDE_DIR variables
weren't used and were removed.

CMake will now fail if VOLK cannot be detected. Therefore, ENABLE_VOLK
was also removed as a variable; all in-tree components may assume the
existence of VOLK.

Note: As of now, this will fail BBot, because it won't find any VOLK. This is for review.


devnulling:
Fixes: https://github.com/gnuradio/gnuradio/issues/3005

drmpeg:
Looks like there's a typo in VolkConfigVersion.cmake. Here's my CMake dump (for just the volk portion). Note that my install directory (/opt/gnuradio-3.7.12git) doesn't have anything to do with 3.7, it's just reusing old environment variables.
```
-- Configuring VOLK support...
CMake Warning (dev) in /opt/gnuradio-3.7.12git/lib/cmake/volk/VolkConfigVersion.cmake:
  A logical block opening on the line

    /opt/gnuradio-3.7.12git/lib/cmake/volk/VolkConfigVersion.cmake:29 (if)

  closes on the line

    /opt/gnuradio-3.7.12git/lib/cmake/volk/VolkConfigVersion.cmake:32 (endif)

  with mis-matching arguments.
  CMakeLists.txt:393 (find_package)
This warning is for project developers.  Use -Wno-dev to suppress it.

-- Found VOLK: Volk::volk  
--   Found VOLK:
--   * Version: 
--   * Libraries: Volk::volk
--   * Includes: /opt/gnuradio-3.7.12git/include
```
Otherwise, it builds and runs fine.

mbr0wn:
@drmpeg Yup, can confirm. That's a VOLK issue, though, so we'll need to fix that elsewhere.

mbr0wn:
@drmpeg Also, thanks for testing :)

noc0lour:
@mbr0wn will update the bb workers to include the correct version of VOLK

ghostop14:
Testing close permissions

noc0lour:
@mbr0wn would it be enough to install the latest volk version from the package manager for now? If a newer version is needed in the future we can work on that.

michaelld:
> Looks like there's a typo in VolkConfigVersion.cmake. Here's my CMake dump (for just the volk portion). Note that my install directory (/opt/gnuradio-3.7.12git) doesn't have anything to do with 3.7, it's just reusing old environment variables.

This issue was fixed a long time ago ... https://github.com/gnuradio/volk/commit/21d86728946b915bfe2f177dfa97fb2c94a5957a

michaelld:
Where are we in dealing with this PR?

drmpeg:
@michaelld Except the fix is incorrect. It should be MAINT_VERSION, not MINOR_VERSION.

michaelld:
> @michaelld Except the fix is incorrect. It should be MAINT_VERSION, not MINOR_VERSION.

oops ... LOL my bad! I'll fix it shortly!

michaelld:
Fixup PR here: https://github.com/gnuradio/volk/pull/337

noc0lour:
Also for now I will include volk v2.1.0 in the docker images & build it from source, as only debian 10 has volk, but only in version 1.4, which is partially my fault anyway...

mbr0wn:
@noc0lour We need to make sure we keep the minimum VOLK versions in the automated tests, so we don't miss a GNU Radio bug that gets hidden with a later VOLK version. I have a script here that can get the correct VOLK tarball for a GR git checkout, maybe we can use that.

noc0lour:
@mbr0wn I would really like to avoid rebuilding VOLK for each compilation as we want to treat it as an external dependency for master and thus it should be a binary compiled once per docker image. How about we create different docker images for master & 3.8 which have their dependencies fixed to the versions we want to have?

noc0lour:
Basically I can do a layered cake which for now just includes the different volk versions on top of the normal docker images.

mbr0wn:
Sure we can layer cake. We'll need 3 versions (1.4 for maint-3.7, 2.0 for maint-3.8, and 2.1.0 for master), and some logic to switch.

noc0lour:
@mbr0wn I just force-pushed the same commit to rerun the pipeline, somehow the original build to retrigger was missing in buildbot.

noc0lour:
Ah, the push runner shouldn't be run at all, but that's an issue for me in gnuradio-buildbot.

marcusmueller:
@mbr0wn are clear for takeoff on this one? Can we remove the "Don't Merge" tag and merge?

mbr0wn:
Just rebased. Let's check the CI!

drmpeg:
Unfortunately, there's a couple of issues.

1) In CMakeLists.txt, VOLK_MIN_VERSION isn't really a minimum version. That is, cmake fails if the VOLK version is 2.2.0 or 2.2.1.

2) The CMakeLists.txt file(s) in gr-network need to be updated.

mbr0wn:
Hm, I was relying on BuildBot to catch any issues. Guess we'll need to fixup...

PR #3073, opened 2020-01-14
Filter: Improve Performance in xlating FIR Filter

ghostop14:
Re-ordered work call to let rotate take advantage of volk.  Rather than rotating individually with each for, the rotate happens after with a rotateN to use the volk version for the data.  Removed decimation() function stack push/pop with each loop iteration.

### DIFF
--- a/gr-filter/lib/freq_xlating_fir_filter_impl.cc
+++ b/gr-filter/lib/freq_xlating_fir_filter_impl.cc
@@ -57,7 +57,8 @@ freq_xlating_fir_filter_impl<IN_T, OUT_T, TAP_T>::freq_xlating_fir_filter_impl(
       d_proto_taps(taps),
       d_center_freq(center_freq),
       d_sampling_freq(sampling_freq),
-      d_updated(false)
+      d_updated(false),
+      d_decim(decimation)
 {
     std::vector<gr_complex> dummy_taps;
     d_composite_fir =
@@ -175,9 +176,15 @@ int freq_xlating_fir_filter_impl<IN_T, OUT_T, TAP_T>::work(
 
     unsigned j = 0;
     for (int i = 0; i < noutput_items; i++) {
-        out[i] = d_r.rotate(d_composite_fir->filter(&in[j]));
-        j += this->decimation();
+        out[i] = d_composite_fir->filter(&in[j]);
+        j += d_decim;
     }
+
+    // re-use of the same buffer as the input and output is safe for many volk functions
+    // and faster than creating local temporary memory in the work function and doing an
+    // extra copy. So out is used below as both the in and out params to the rotate
+    // function.
+    d_r.rotateN(out, out, noutput_items);
 
     return noutput_items;
 }
--- a/gr-filter/lib/freq_xlating_fir_filter_impl.h
+++ b/gr-filter/lib/freq_xlating_fir_filter_impl.h
@@ -42,6 +42,7 @@ protected:
     double d_center_freq;
     double d_sampling_freq;
     bool d_updated;
+    const int d_decim;
 
     virtual void build_composite_fir();
 


gr-filter/lib/freq_xlating_fir_filter_impl.cc
```diff
@@ -174,10 +175,22 @@ int freq_xlating_fir_filter_impl<IN_T, OUT_T, TAP_T>::work(
     }
 
     unsigned j = 0;
+    /*
```
mbr0wn: Please remove the old code.

gr-filter/lib/freq_xlating_fir_filter_impl.cc
```diff
@@ -174,10 +175,22 @@ int freq_xlating_fir_filter_impl<IN_T, OUT_T, TAP_T>::work(
...
     for (int i = 0; i < noutput_items; i++) {
         out[i] = d_r.rotate(d_composite_fir->filter(&in[j]));
         j += this->decimation();
     }
+    */
+
+    // Moved rotate out to support using volk
+    for (int i = 0; i < noutput_items; i++) {
+        out[i] = d_composite_fir->filter(&in[j]);
+        // eliminated stack push/pop for decimation() and used recommended j + d_decim rather than +=
```
mbr0wn: This comment is better suited in the commit message; please move it there.

gr-filter/lib/freq_xlating_fir_filter_impl.h
```diff
@@ -42,6 +42,7 @@ class FILTER_API freq_xlating_fir_filter_impl
     double d_center_freq;
     double d_sampling_freq;
     bool d_updated;
+    int d_decim;
```
mbr0wn:
```suggestion
    const int d_decim;
```

ghostop14:
Thanks mbr0wn, I'll make the changes later.  For the += versus +, it was an enhancement recommendation I came across years ago working on lfast.  I have been running timing tests with each incremental change (I've seen cases where changes actually make things worse instead of better so I like to verify and quantify them first).  I have a python script based on filter_timing.py I use as a framework for incremental testing.  The += change was bundled into the d_decim change, in total a 3.3% improvement just from that change.

ghostop14:
CLA has been submitted.

ghostop14:
The "Fixed formatting issues" commit removed those comments.  Are they still showing up?

argilo:
Just a heads up that the rotator kernel in VOLK's 2.0.0 release (which is still used by GNU Radio 3.8) is broken. More details can be found in #2748. This may not may or may not matter by the time this PR is merged, but I'm mentioning it just in case.

ghostop14:
I actually opened an issue on that this morning too.  Sounds like it may be an already-known issue then.  Was seeing weird cycling noise in the volk version.

ghostop14:
FYI I was testing against GR master and still saw the same volk effect.

argilo:
@ghostop14 Was your VOLK submodule up to date? The master branch picked up the VOLK rotator fix back in October: abba02bd39b6f6fabb63755b4909d1d3f7d8ba4b

argilo:
if you're talking about https://github.com/gnuradio/volk/issues/329, that looks like a different issue. The bug I encountered wasn't just some additional noise; a large number of output samples were completely incorrect.

rear1019:
> Thanks mbr0wn, I'll make the changes later. For the += versus +, it was an enhancement recommendation I came across years ago working on lfast. I have been running timing tests with each incremental change (I've seen cases where changes actually make things worse instead of better so I like to verify and quantify them first). I have a python script based on filter_timing.py I use as a framework for incremental testing. The += change was bundled into the d_decim change, in total a 3.3% improvement just from that change.

GCC 9.2 generates ths same assembly code regardless whether ``+=`` or ``+`` is used, with -O2 and -O3. The generated code only changes if ``d_decim`` is replaced by ``this->decimate()`` (3 instructions instead of 1 to update ``j``); it does not matter whether ``+=`` or ``+`` is used.

ValZapod:
@rear1019 You are forgetting the time gcc intermediate GIMPLE format spends on it which is obviously faster for +=. But seriously it reminded me of https://stackoverflow.com/questions/24848359/which-is-faster-while1-or-while2

gr-filter/lib/freq_xlating_fir_filter_impl.cc
```diff
@@ -175,10 +176,12 @@ int freq_xlating_fir_filter_impl<IN_T, OUT_T, TAP_T>::work(
 
     unsigned j = 0;
     for (int i = 0; i < noutput_items; i++) {
-        out[i] = d_r.rotate(d_composite_fir->filter(&in[j]));
-        j += this->decimation();
+        out[i] = d_composite_fir->filter(&in[j]);
+        j = j + d_decim;
```
michaelld: I would prefer using `j +=` here. WIth optimization the resulting code should be the same. Using `+=` is just cleaner IMHO.

gr-filter/lib/freq_xlating_fir_filter_impl.cc
```diff
@@ -175,10 +176,12 @@ int freq_xlating_fir_filter_impl<IN_T, OUT_T, TAP_T>::work(
 
     unsigned j = 0;
     for (int i = 0; i < noutput_items; i++) {
-        out[i] = d_r.rotate(d_composite_fir->filter(&in[j]));
-        j += this->decimation();
+        out[i] = d_composite_fir->filter(&in[j]);
```
michaelld: While using `out` as a temporary buffer will work, I'd prefer to use an actual temporary buffer instead, just to be memory-safe. So, using a `std::vector` of the correct type and size here for the output and then as the input for line 183. I know this change isn't technically necessary since the rotator -should- be able to handle in-place computations ... but, IMHO better safe than sorry! If anyone disagrees, please talk me out of it...

ghostop14: With how frequently work gets called, newing and deleting memory for a temp buffer can be expensive.  You can do it with a class-level buffer that gets allocated once and destroyed in stop/destructor and only re-allocated if the necessary size is greater than the currently allocated size.  But there's a number of volk kernels where it is safe to use the same space in in and out.  In the interest of performance if it works okay and is safe this way, i wouldn't complicate the class by doing additional temp buffer handling.  Just my personal preference.

michaelld: Ok. Then let me change my request: add a comment here to the effect of what you just wrote, about how `out` is being used then reused here and why, how it interfaces with Volk, etc...

ghostop14:
Thanks to mbr0wn's talk, I think I have the commits merged in this PR.  I did give in to the "+=" comments and make that change back.  I've encountered it as a performance improvement but it's been so long I can't point back to my reference.

gr-filter/lib/freq_xlating_fir_filter_impl.cc
ghostop14: Happy to add comments.  Were you thinking in association with the "out[i] = d_r.rotate(d_composite_fir->filter(&in[j]))" line or the "d_r.rotateN(out, out, noutput_items)" line?  I think it would make more sense on the 2nd since the first really wouldn't require any temp space since out isn't on the right side of the = anywhere?

michaelld: Since the code is so short, I'd say before the line `unsigned j = 0;` ... anyone reading the comment can quickly the following lines to understand what's going on then. And, doesn't interrupt code flow. Thx!

ghostop14:
@michaelld Added a comment in there summarizing our discussion.  Let me know if you'd like to see more.

michaelld:
Adding in a few other GR devs for at least 1 more review.

mbr0wn:
CLA submitted.

argilo:
A quick update on my previous comment: the final decision was to keep the maint-3.8 branch on VOLK 2.0.0: https://github.com/gnuradio/gnuradio/pull/3062#issuecomment-575289283 That means that this PR should only be merged to master and not backported to the maint-3.8 branch, since `rotateN` would use the broken VOLK rotator there.

PR #3075, opened 2020-01-14
blocktool: minor convenience changes and resolve some parsing errors

mormj:
My setup here is Ubuntu 19.10 with gcc-9 which may come into play as to why these changes were needed

1) pygccxml was throwing "invalid constraint in asm" related to boost
to resolve this had to add the define BOOST_ATOMIC_DETAIL_EXTRA_BACKEND_GENERIC
without this you get a bunch of:
```
/usr/include/boost/atomic/detail/extra_ops_gcc_x86.hpp:56:50: error: invalid output constraint '=@ccc' in asm
            : [storage] "+m" (storage), [result] "=@ccc" (res)
                                                 ^
/usr/include/boost/atomic/detail/extra_ops_gcc_x86.hpp:80:50: error: invalid output constraint '=@ccc' in asm
            : [storage] "+m" (storage), [result] "=@ccc" (res)
```
also had to add -std=c++11, which without that you get
```
/usr/include/c++/9/bits/stl_function.h:418:6: error: use of undeclared identifier '__builtin_is_constant_evaluated'
        if (__builtin_is_constant_evaluated())
```
2) include paths were assuming /usr/include which is not always the case
- added extra cli parameters for include paths
3) made gr_blocktool a top level program that gets installed

### DIFF
--- a/gr-utils/python/blocktool/cli.py
+++ b/gr-utils/python/blocktool/cli.py
@@ -66,6 +66,8 @@ def run_blocktool(module):
               help='blocktool helper comments will be added in the header file')
 @click.option('-o', '--output', is_flag=True,
               help='If given, a file with desired output format will be generated')
+@click.option('-I', '--include_paths', default=None,
+              help='Comma separated list of include paths for header files')
 def cli(**kwargs):
     """
     Block header parsing tool.
--- a/gr-utils/python/blocktool/core/base.py
+++ b/gr-utils/python/blocktool/core/base.py
@@ -39,7 +39,7 @@ class BlockTool(ABC):
 
     def __init__(self, modname=None, filename=None, targetdir=None,
                  target_file=None, module=None, impldir=None, impl_file=None,
-                 yaml=False, json=False, **kwargs):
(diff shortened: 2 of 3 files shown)


aru31:
Hi @mormj, thank you for the blocktool patches. 
Just one comment:
`self.included_paths = None` is required to be defined in the `__init__` function otherwise `AttributeError` will be is thrown if not used.
Also, I didn't make the gr_blocktool script as it was decided at the time of my GSoC coding period.

mormj:
Thanks @aru31 and great work on the blocktool.  I'm going to be pounding on it pretty hard trying to use it for replacing the swig bindings with pybind11.  

I'll remove the gr_blocktool commit (though I think this would be useful as a top level script) and make the changes you noted.  Thanks!

michaelld:
Adding in c++11 is wise, as is the BOOST definition. Boost is such a picky beast ... hopefully that definition works for both current Boost as well as back to whatever minimum is GR requires ... it's about impossible to test all of these, of course!

PR #3076, opened 2020-01-14
Digital: Costas Loop Performance Enhancements

ghostop14:
Pulled in-tree the performance enhancements from gr-lfast's Costas loops.  Performance tricks incorporated here in code include removing all function stack push/pops in the for loop, doing division operations as multiplication and with pre-computed constants such as 1/(2*pi), and leveraging Fused Multiply/Add (FMA) CPU capabilities when available.  (FMA executes a*b+c as a single CPU operation rather than 2 ops).  

Straight C++ code is about twice as fast for 2nd order, 50% faster for 4th order.   Overall flowgraph tests show 33% performance improvement in flowgraphs with tag and optional outputs on the Costas block.  Block can now do almost 40 MSPS throughput on 2nd order loop on newer CPU's or run on lower-end hardware/SBCs with narrower bandwidth.


gr-digital/lib/costas_loop_cc_impl.cc
```diff
@@ -31,9 +31,28 @@
...
+#endif
+
+#if defined(FP_FAST_FMA)
+#define __FMA__ 1
+#endif
+
 namespace gr {
 namespace digital {
 
+struct ComplexStruct {
```
bistromath: Is there a reason to use this instead of gr_complex?

gr-digital/lib/costas_loop_cc_impl.cc
```diff
@@ -204,6 +224,189 @@ int costas_loop_cc_impl::work(int noutput_items,
     return noutput_items;
 }
 
+int costas_loop_cc_impl::work2(int noutput_items,
```
bistromath: I really like the performance increase. However, for a number of reasons -- cross-compiling, runtime platform detection, ease of debugging, etc. -- we usually put platform-specific optimizations into Volk and call out to them from the block's work function. Is it feasible to export the fast FMA (and generic) core of this block to a Volk kernel?

ghostop14:
I'd leave that to the volk experts if it can be done there.  The sequential nature of the calculations don't lend themselves to SIMD since they have to be done in order and can't be done in parallel.

gr-digital/lib/costas_loop_cc_impl.cc
ghostop14: Reaching for every ounce of performance, each real() and imag() call in the loop with gr_complex was causing function jumps.  the ComplexStruct overlays a struct over the same memory to access the data without the jumps.  In general I wouldn't worry too much about it, but when you wrap it up in a work for loop, the gain adds up.

rear1019: I can not reproduce this with current master, compiling with GCC 9.2 (-O2). I see no functions calls, i.e. ``complex::real()/imag()`` is implemented (as expected) efficiently.

CMakeLists.txt
```diff
@@ -232,6 +232,26 @@ if(WIN32)
     endif(CMAKE_SIZEOF_VOID_P EQUAL 8)
 endif(WIN32)
 
+# Setup FMA Capabilities Flag
+IF(UNIX)
+    MESSAGE(STATUS "Checking for FMA support...")
+    IF(NOT APPLE)
```
michaelld: why not swap the ordering here and do `IF(APPLE)` ... really just a small tweak, but IMHO makes it a little more readable.

CMakeLists.txt
```diff
@@ -232,6 +232,26 @@ if(WIN32)
...
+        EXEC_PROGRAM(cat ARGS "/proc/cpuinfo" OUTPUT_VARIABLE CPUINFO)
```
michaelld: Do all non-Apple UNIX support `/proc/cpuinfo`? Or most of them? Or just Linux-y ones? I'm guessing this isn't supported by all non-Apple UNIX ... so, maybe `ELSEIF(LINUX)` this and then in the final `ELSE()` clause set `CPUINFO` to empty & print a warning.

CMakeLists.txt
```diff
@@ -232,6 +232,26 @@ if(WIN32)
...
+    ENDIF(NOT APPLE)
+
+    STRING(REGEX REPLACE "^.*(fma).*$" "fma" FMA_THERE ${CPUINFO})
+    STRING(COMPARE EQUAL "fma" "${FMA_THERE}" FMA_TRUE)
+    IF (FMA_TRUE)
+		MESSAGE(STATUS "Found FMA support.  Setting fma compile flag.")
+		set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -mfma")
+    ELSE (FMA_TRUE)
+		MESSAGE(STATUS "FMA support not found.  Not setting fma flag.")
+	ENDIF (FMA_TRUE)
```
michaelld: Fix line spacing

gr-digital/lib/costas_loop_cc_impl.cc
```diff
@@ -31,9 +31,28 @@
 #include <gnuradio/sincos.h>
 #include <boost/format.hpp>
 
+#define CL_TWO_PI 6.28318530717958647692
```
michaelld: Guessing at least some of these are defined in <math.h> or <cmath> or whatever. Please use those instead of redefining here.

gr-digital/lib/costas_loop_cc_impl.cc
michaelld: I'd guess that with `-O2` optimization enabled using this or `std::complex` will result in roughly the same code and code speed, since the calls to `real()` and `imag()` should be optimized out by the compiler. My preference would be to use `gr_complex` and related class calls here.

gr-digital/lib/costas_loop_cc_impl.cc
```diff
@@ -204,6 +224,189 @@ int costas_loop_cc_impl::work(int noutput_items,
...
+    return noutput_items;
+}
+
+int costas_loop_cc_impl::work(int noutput_items,
+                              gr_vector_const_void_star& input_items,
+                              gr_vector_void_star& output_items)
+{
+    return workDefault(noutput_items, input_items, output_items);
+
+    switch (d_order) {
```
michaelld: Because of the `return` just above, none of this code will be executed ... is that the intent? If so, please remove the extra code. If not, please remove the `return` and/or fixup this section to make sense for the code intent.

CMakeLists.txt
michaelld: Or just set `CPUINFO` to "" up front, and in the final `ELSE()` print a warning about this OS not being supported for FMA & please consider adding in support!

ghostop14:
Thanks for the review michaelld, that one return at the top in particular I was flipping on/off for timing tests.  I'll go through all of the comments and address them over the weekend and push updates.  Appreciate all the feedback!

For the CMakelists changes if anyone has some other systems just to confirm the cpuinfo question, I've tested on Ubuntu, straight debian, and Raspbian and it works on those but some of the other variants I don't know about.  I googled and looks like OpenSUSE has it as well but if anyone has a less popular linux to check it may be helpful.

gr-digital/lib/costas_loop_cc_impl.cc
ghostop14: I'm going to go back to the pure cpp test timing code I have in gr-lfast and confirm the real() imag() item.  When I originally wrote it a couple years back I was splitting hairs on performance and timing a line change at a time to squeeze out every last drop of performance and it made a difference.  Would you know off the top of your head if -O2 or -O3 is the default in the GNU Radio build?  I think the value in some of those types of explicit changes comes down to trust (I can be a control freak about it) and whether "the compiler should" versus every compiler "always will" optimize out.  But I'll revalidate it line-by-line just to confirm there's a difference or not.  I think it'll be a good exercise to redo.

CMakeLists.txt
```diff
@@ -232,6 +232,26 @@ if(WIN32)
...
+        EXEC_PROGRAM(cat ARGS "/proc/cpuinfo" OUTPUT_VARIABLE CPUINFO)
+    ELSE(NOT APPLE)
+        EXEC_PROGRAM("/usr/sbin/sysctl -n machdep.cpu.features" OUTPUT_VARIABLE CPUINFO)
+    ENDIF(NOT APPLE)
+
+    STRING(REGEX REPLACE "^.*(fma).*$" "fma" FMA_THERE ${CPUINFO})
+    STRING(COMPARE EQUAL "fma" "${FMA_THERE}" FMA_TRUE)
+    IF (FMA_TRUE)
+		MESSAGE(STATUS "Found FMA support.  Setting fma compile flag.")
+		set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -mfma")
```
rear1019: Don’t do this! This allows the compiler to issue FMA instructions. The resulting code will not run on a machine without support for FMA. Think of a Linux distribution using a fancy build machine and some user installing the packing on a machine without FMA.

gr-digital/lib/costas_loop_cc_impl.cc
ghostop14: I found the flag setting in the CMakeCache file.  Looks like the general release build is -O3, and with debug is -O2.

CMakeLists.txt
ghostop14: rear1019, the if process there is a detection system to determine if FMA is available on the system or not, so if it doesn't support it, it won't be set.  I have an Udoo board for instance that doesn't support FMA so the modules built on there don't issue FMA instructions and everything runs fine.  Do you have an instance where it could false-positive detect FMA that's not covered in the detection logic?

gr-digital/lib/costas_loop_cc_impl.cc
ghostop14: michaelld: I double-checked.  Those particular defines are not in math.h.  The only Pi options there are:
# define M_PI		3.14159265358979323846	/* pi */
# define M_PI_2		1.57079632679489661923	/* pi/2 */
# define M_PI_4		0.78539816339744830962	/* pi/4 */
# define M_1_PI		0.31830988618379067154	/* 1/pi */
# define M_2_PI		0.63661977236758134308	/* 2/pi */
# define M_2_SQRTPI	1.12837916709551257390	/* 2/sqrt(pi) */

Not sure why 2*Pi wouldn't be a standard define.  You'd think that would be pretty common.

bistromath:
@ghostop14, what happens when a package manager puts together a .deb? Which compile flag will it have? What machines will it run on?

This is one of the arguments for doing platform-specific work in Volk, where it is selected at runtime, not at compile time.

gr-digital/lib/costas_loop_cc_impl.cc
bistromath: #include <gnuradio/math.h>

GR_M_PI = <pi>
GR_M_TWOPI = (2*pi)

michaelld: @ghostop14 OK ... well, worth checking! At least those that @bistromath notes can be fixed!

ghostop14:
I think I've addressed everyone's feedback in the latest commit.  The IF(APPLE) section and indention has been cleaned up, I added a safety check for /proc/cpuinfo if in the unlikely event it's not there on a linux distro, removed the wrong return, and got rid of the struct relying on -O3 optimization to get the same result with gr_complex.

Squashed the commit.  See if I missed any of the other suggestions?

bistromath:
I don't think you've appropriately addressed the concern @rear1019 and I raised. This will break .deb builds for users with machines without FMA. There is a difference between _compile-time_ and _run-time_ support for instructions. If you build on the same machine you run on, this will work fine. If you don't, this will result in illegal instruction errors for .deb users on machines without FMA, if the build machine has FMA support.

There are two solutions. One is disabling FMA support by default, making it contingent on a CMake flag which specifically enables it when requested. The other is moving this into Volk (if it is indeed possible to do so), where automatic runtime platform selection is a thing. The second approach is much preferred.

Sorry to be the stick in the mud here, as I'd like to see a faster Costas loop in GR. But it has to work for everybody.

ghostop14:
@bistromath good point on the package managers.  Is there a way to detect that in the build process?  Maybe we default to off for those and allow folks building from source/pybombs to detect on their system?  I don't know CMake well enough to suggest how to do that.

ghostop14:
@bistromath definitely not a stick in the mud man!  I like the discussion actually.  Keep me honest.  It's worked great as an OOT, but as you pointed out they're not in a package manager.  Giving it some thought, maybe the build flag is the best approach with the default being none.  There's other good optimization in the routine even without FMA, it's just not as good, but still worthwhile.  Let me give that a shot.  What do you and @rear1019 think about that approach?

gr-digital/lib/costas_loop_cc_impl.cc
mbr0wn: Echo this: I also prefer gr_complex. I would be very surprised that a local hack will beat the STL performance, especially with other optimization levels.

CMakeLists.txt
mbr0wn: It detects on the build system, not the system where the binaries get executed (what if the Debian build server supports FMA, then all Debians would need it).

ghostop14:
I like the option to have FMA off by default and have a compile flag we can set to then enable the detection process.  I think that would solve the package build issue and let those who really need the performance enhancement enable it.  We'd probably need to put a note somewhere in the build docs that it's available.  I like that in that for those who really want increased performance might pick up other benefits from FMA compilation.  Even without FMA there will be a moderate performance increase with the code.  In testing it was about 7% on my i7-7700HQ laptop.  Thoughts?

velichkov:
@ghostop14,

> Thoughts?

Usually I build with `-march=native -ftree-vectorize` in `CFLAGS` and `CXXFLAGS`. This way there is no need of any additional cmake checks and all available instructions (FMA/SIMD/...) for the current CPU gets used by gcc/clang, but of course the resulting binaries are not portable. So instead of fma specific cmake option you could probably add an option for a `-march=native` build.

```
cmake -D CMAKE_CXX_FLAGS="-march=native -ftree-vectorize" -D CMAKE_C_FLAGS="-march=native -ftree-vectorize" -D CMAKE_BUILD_TYPE=RelWithDebInfo ..
```

rear1019:
In my opinion there are the following things to consider:

* I think we all agree that the detection of FMA (``-mfma``) should not be enabled implicitly, only if requested explicitly. Enabling this flag globally benefits all compiled code.
* Manual usage of ``__builtin_fma()`` (or ``fma()`` available in C++/C99) is not required when compiling with ``-mfma``: the compiler uses FMA automagically where appropriate. I checked the code for ``control_loop.cc`` and ``costas_loop_cc_impl.cc``, using GCC 9.2 -O2, compiling current HEAD (commit 63e490434).
* Rather problematic is the mixing of ``double`` and ``float`` values in the ``phase_detector_*`` functions. When compiling with ``-mfma``, appending ``f`` to the ``double`` constants makes the compiler produce pretty and partially branchless code. Even without FMA, reducing the number of conversions will (should?) help improve performance. However, I do not know whether the usage of ``double`` constants is intentional or not (for increased precision).
* As far as I unsterstand comments by @ghostop14, most of the speed gain is due to the usage of FMA. The other change in this pull requesting ist the inlining of ``advance_loop()``, ``phase_wrap()``, ``frequency_limit()``. These functions are defined in ``blocks::control_loop``, which is a base class for ``costas_loop`` and other blocks. We could check whether making these functions ``inline`` (move them to header) will help the compiler optimize. This will benefit all blocks using ``blocks::control_loop`` as a base class.
We also could write specific ``work()`` functions as done in the pull request – if need be to get the last few % of performance.  This basically inlines the call of ``phase_detector_*()`` instead of using a function pointer.

As you might read between the lines, I am rather against merging this pull request (in its current form). But we definitely can use some ideas from it.

If a user really need the performance, it is way faster and probably way easier to compile and install the OOT module instead of having to deal with GNU Radio compilation.

ghostop14:
@rear1019, I really like those bullets.  Let me go back and incorporate those into the code.  It wouldn't be hard to just strip out the FMA-specific if statements I've added too and let the compiler do the work.  I haven't tested it yet, but in terms of optimizations, would you suspect if you had an x=a*b+c or an x=c+a*b that the compiler would still be smart enough to apply FMA on its own?
  
@velichkov, maybe bundling those into a readable command-line setting and handling some of it in the CMake like a -DENABLE_NATIVE might be something I could add instead to pick them up.

I know I'm being persistent and I really like and appreciate all the discussion and perspectives (keep it coming!).  A few thoughts on my end driving all this, 1. The Costas enhancements have been picked up in some other FOSS projects to actually bring LRIT decoding in real-time down to Raspberry Pi levels, so the optimizations can let some processing happen on lower-end hardware, which I think opens up some great possibilities.  2. Thinking strategically as bandwidth keeps increasing, squeezing more performance out of what we already have without relying so much on next-gen CPU's sounds appealing (even 10% across the board can help, but some I think we can do even better).  and 3. Pitching 3.9 as the "fastest GNU Radio yet" by the time it's ready for release has a nice ring to it too.

ghostop14:
I think I captured all of the bullets in the latest push.  

- CMakeLists.txt now has an off-by-default     ENABLE_NATIVE that if enabled, use a number of additional optimizations that velichkov recommended, including FMA support.  
- Tested with the -mfma and taking the builtin_fmaf forces out and I did get the same results so relying on the compiler seems fine.  So I took the ifdef wrappings out.
- I didn't see any mixing of floats/doubles in the phase_detector_* but it shouldn't be an issue.  CPP reference says that calculations are done with max precision and only rounded once.
- The control_loop block has been enhanced to move the calls used in the generic Costas loop to inline functions, and optimized 2nd and 4th order loops are still there added to eliminate function jumps.  

New to this push, optimizations were added to eliminate unnecessary if checks per loop iteration if optional outputs are not connected.  The changes for the inline control loop also required a minor ifdef check in fll_band_edge_cc.impl to go with the changes.  M_TWOPI was defined in 2 different locations throughout GR (maybe worth moving it to gr_math.h), made them match and ifdef'd the redefinition out.

gr-digital/lib/costas_loop_cc_impl.h
```diff
@@ -41,21 +41,21 @@ class costas_loop_cc_impl : public costas_loop_cc
      *  \param sample complex sample
      *  \return the phase error
      */
-    float phase_detector_8(gr_complex sample) const; // for 8PSK
+    inline float phase_detector_8(gr_complex sample) const; // for 8PSK
```
velichkov: I don't think this `inline` here has the desired effect. The `pahse_detector_*` functions are used only in the `workDefault` through the function pointer `d_phase_detector` so they are not inlined.

CMakeLists.txt
```diff
@@ -448,6 +448,19 @@ message(STATUS "  Override with -DENABLE_INTERNAL_VOLK=ON/OFF")
 # Handle logging
 find_package(LOG4CPP REQUIRED)
 
+########################################################################
+# Setup Native Capabilities Flag
+########################################################################
+IF(UNIX)
+    IF (ENABLE_NATIVE)
+		MESSAGE(STATUS "Found GNU Radio optimization flag.  Setting optimization flags.")
```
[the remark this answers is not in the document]
ghostop14: I like the option approach.  I've never used cmake-gui before but I just checked it out.  Looks interesting.  I'll add this line in as well.

gr-digital/lib/costas_loop_cc_impl.cc
```diff
@@ -193,6 +206,72 @@ int costas_loop_cc_impl::work(int noutput_items,
...
+        n_r = cosf(-d_phase);
+
+        o_r = (iptr[i].real() * n_r) - (iptr[i].imag() * n_i);
+        o_i = (iptr[i].real() * n_i) + (iptr[i].imag() * n_r);
+
+        optr[i].real(o_r);
+        optr[i].imag(o_i);
+
+        d_error = o_r * o_i;
+        d_error = 0.5 * (std::abs(d_error + 1) - std::abs(d_error - 1));
```
rear1019: This looks like ``branchless_clip()``. No need to inline manually, the compiler does. People reading the code will wonder what this line makes (at least to me it is not obvious).

gr-digital/lib/costas_loop_cc_impl.h
rear1019: I did not check the compiled code, but I believe that as well.

ghostop14:
@rear1019, it is.  The flattening of the routines generate a good bit of the performance updates.  I know the compiler "should", but in gr-lfast comparing original code versus the optimized code, here's the performance differences:

(test-lfast output timed directly in cpp):
Testing 2nd order Costas Loop with 8,192 samples...
Original Code Run Time:      0.000517 s  (15,852,481.000000 sps)
LFAST Code Run Time:      0.000209 s  (39,189,532.000000 sps)
Speedup:        147.21% faster

Testing 4th order Costas Loop with 8,192 samples...
Original Code Run Time:      0.000387 s  (21,141,586.000000 sps)
LFAST Code Run Time:      0.000247 s  (33,178,402.000000 sps)
Speedup:         56.93% faster

If we rely on the compiler for the optimizations, then the original code should produce the same results but it's not even close.  Even looking at branchless_clip which is defined as inline in gnuradio/math.h.  My concern is that if we roll back to the routines all of the demonstrable performance improvements will go away with it.  

Given those results, and I don't know the answer so this is a question for the group (maybe there's something else we're missing)... "Even with -O3 compilation of the code into the same executable (so same compilation parameters), why is the flattened version so much better performing than the original in the generated code when the expectation around optimizations and inlines is that they should be relatively equivalent?"

gr-blocks/include/gnuradio/blocks/control_loop.h
```diff
@@ -99,7 +107,13 @@ class BLOCKS_API control_loop
      * method in case another way is desired as this is fairly
      * heavy-handed.
      */
-    void phase_wrap();
+    void phase_wrap()
+    {
+        while (d_phase > M_TWOPI)
```
michaelld: I know that this is the traditional way to do phase wrap. I wonder if it would make any difference to wrap each phrase in an `if` ... for example:
```
if (d_phase) {
    while (d_phase > M_TWOPI)
    d_phase -= M_TWOPI;
} else {
    while (d_phase < -M_TWOPI)
         d_phase += M_TWOPI;
}
```
since they are totally separate: if the first `while` happens, then the 2nd won't & vice versa. It's a small issue, really. Just wondering.

gr-digital/lib/costas_loop_cc_impl.cc
```diff
@@ -31,9 +31,12 @@
 #include <gnuradio/sincos.h>
 #include <boost/format.hpp>
 
+// Used to convert division to multiplication for speed
+#define CL_ONE_OVER_2PI 0.15915494309189533577
+#define CL_MINUS_TWO_PI -6.28318530717958647692
```
michaelld: Should be this `- GR_M_TWOPI`?

gr-digital/lib/costas_loop_cc_impl.cc
```diff
@@ -59,28 +62,29 @@ costas_loop_cc_impl::costas_loop_cc_impl(float loop_bw, unsigned int order, bool
...
+    case 8:
+        if (use_snr) {
+            return &costas_loop_cc_impl::phase_detector_snr_8;
+        }
+        return &costas_loop_cc_impl::phase_detector_8;
     }
-    return &costas_loop_cc_impl::phase_detector_8;
-  }
-  throw std::invalid_argument("order must be 2, 4, or 8");
+    throw std::invalid_argument("order must be 2, 4, or 8");
```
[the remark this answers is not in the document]
michaelld: If these are already in the master branch then they need to be removed here before we can merge to master. Please make it so!

gr-digital/lib/costas_loop_cc_impl.cc
```diff
@@ -175,6 +179,15 @@ int costas_loop_cc_impl::work(int noutput_items,
                       nitems_read(0) + noutput_items,
                       pmt::intern("phase_est"));
 
+    // Get this out of the for loop if not used:
+    bool hasAdditionalOutputs = false;
+    if (freq_optr != NULL)
```
michaelld: why directly compare with `NULL`? Why not `if (freq_optr)`? The compiler output code should be identical, and it is, at least in my opinion and experience, easier to read. either way works though for this purpose.

gr-digital/lib/costas_loop_cc_impl.cc
```diff
@@ -175,6 +179,15 @@ int costas_loop_cc_impl::work(int noutput_items,
                       nitems_read(0) + noutput_items,
                       pmt::intern("phase_est"));
 
+    // Get this out of the for loop if not used:
+    bool hasAdditionalOutputs = false;
```
michaelld: if the number of streams don't change during runtime, then this value can be computed elsewhere & doesn't need to be done every single call to work. I think we (ab)use `::check_topology` for this purpose.

michaelld: I'm also not convinced of the benefit of this variable, later in the work function. If none of the streams are in use then it's 1 `if` & otherwise 4. Since we know the number of streams before any of this happens, we should be able to make optimal use of that knowledge here. Worth considering IMHO.

gr-digital/lib/costas_loop_cc_impl.cc
```diff
@@ -193,6 +206,72 @@ int costas_loop_cc_impl::work(int noutput_items,
...
+        optr[i].real(o_r);
+        optr[i].imag(o_i);
+
+        d_error = o_r * o_i;
+        d_error = 0.5 * (std::abs(d_error + 1) - std::abs(d_error - 1));
+
+        d_freq = d_beta * d_error + d_freq;
+        d_phase = d_phase + d_alpha * d_error + d_freq;
+
+        // frequency_limit();
```
michaelld: yes ... why not use the `frequency_limit` function as done in the `control_loop` class? it is usable here in any capacity, or is that a totally separate object / class?

gr-digital/lib/fll_band_edge_cc_impl.cc
```diff
@@ -32,7 +32,9 @@
 namespace gr {
 namespace digital {
 
-#define M_TWOPI (2 * M_PI)
+#ifndef M_TWOPI
```
michaelld: again: use `<gnuradio/math.h>` and then `GR_M_TWOPI` ...

gr-digital/lib/costas_loop_cc_impl.cc
```diff
@@ -204,6 +283,94 @@ int costas_loop_cc_impl::work(int noutput_items,
...
+
+    return noutput_items;
+}
+
+int costas_loop_cc_impl::work(int noutput_items,
+                              gr_vector_const_void_star& input_items,
+                              gr_vector_void_star& output_items)
+{
+    switch (d_order) {
+    case 2:
```
michaelld: There's a lot of repetition in these 3 `work` functions. Maybe it makes more sense branch off the common parts & `switch` around the different parts? Just a single `work` function then?

gr-blocks/include/gnuradio/blocks/control_loop.h
```diff
@@ -24,6 +24,10 @@
 #define GR_BLOCKS_CONTROL_LOOP
 
 #include <gnuradio/blocks/api.h>
+#include <gnuradio/math.h>
+#ifndef M_TWOPI
```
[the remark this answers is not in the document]
ghostop14: That's an easy one to roll in.  I'll adjust that tonight.

gr-digital/lib/costas_loop_cc_impl.cc
ghostop14: @michaelld, the throw() was original from master.  If I remove it, is there a new standard being used in master to raise the error state?  I can incorporate it here.  I did update all the !=NULL references for readability so when I push a new update here it'll be more readable.

gr-digital/lib/costas_loop_cc_impl.cc
ghostop14: @michaelld I did look at the API doc and the source code for check_topology and it isn't clear on the meaning of valid combination.  It would be great to check the actual connected outputs in the constructor and set it once then use it in work.  But I couldn't see an easy way to do that?

michaelld: basically `check_topology` will be provided with the hoped-for # of input and output connections & returns whether that topology meets the blocks needs. For this block, it has better be 1 input & 1-3 outputs. So long as those provided meet the criteria of the block, return true; otherwise return false. AND, since the topology won't change once set, in that method you can setup whatever else you want to handle the possibly variable output topology. I'm pretty sure some of the audio sources / sinks do this. `grep` the source for `check_topology` to get some examples ...

gr-digital/lib/costas_loop_cc_impl.cc
```diff
@@ -31,9 +31,12 @@
 #include <gnuradio/sincos.h>
 #include <boost/format.hpp>
 
+// Used to convert division to multiplication for speed
+#define CL_ONE_OVER_2PI 0.15915494309189533577
```
michaelld: You might consider moving this constant into gr/math.h ... seems like something that might be used elsewhere too!

ghostop14:
Take a look at the latest pushed version.  Spent last night going back through line-by-line and timing.  Changed the unrolled back to inlines, added to the math.h defines, and hunted down the 3 biggest culprit performance lines:
nco_out = gr_expj(-d_phase);
optr[i] = iptr[i] * nco_out;
d_error = (*this.*d_phase_detector)(optr[i]);

So improving those was the only thing that really needed updating after the inlines were done.  I like this version better.  It's cleaner.  Testing results on my system:


Costas Loop Analysis | Measured Throughput | Percent Improvement
-- | -- | --
Original Code: | 15,211,409.00 |  
FMA Just for Costas: | 23,059,684.00 | 51.59%
Adding Inline Functions: | 23,902,168.00 | 57.13%
Replacing Fn Ptr w switch statement | 28,504,874.00 | 87.39%
Flattening gr_expj and cc mult | 42,703,628.00 | 180.73%
Flattened gr_expj but gr_complex mult | 33,316,926.00 | 119.03%

Question on gr_expj, it should just be a sin/cos but when I look at sincos.h it's wrapped with all kinds of ifdefs.  I think the performance would be better if these were just inline defs but since there's all the wrappings I'm guessing there's a good reason for it.  The current call versus just doing it in the loop is a huge difference so I think we could get significant boosts to everything calling gr_expj if it were simply an an inline.

gr-digital/lib/costas_loop_cc_impl.cc
noc0lour: I'd suggest adding more constants to `<gnuradio/math.h>` as we see fit (since they might have some use in the future.

noc0lour: Nvm, this all shows as outdated in GitHub, I shouldn't dig up corpses :)

gnuradio-runtime/include/gnuradio/math.h
```diff
@@ -43,8 +43,10 @@
 #define GR_M_LOG2E 1.4426950408889634074          /* log_2 e */
 #define GR_M_PI 3.14159265358979323846            /* pi */
 #define GR_M_PI_4 0.78539816339744830961566084582 /* pi/4 */
-#define GR_M_TWOPI (2 * GR_M_PI)                  /* 2*pi */
+#define GR_M_TWOPI 6.28318530717958647692         /* 2*pi */
 #define GR_M_SQRT2 1.41421356237309504880         /* sqrt(2) */
+#define GR_M_ONE_OVER_2PI 0.15915494309189533577  /* 1 / (2*pi) */
+#define GR_M_MINUS_TWO_PI -6.28318530717958647692 /* - 2*pi */
```
rear1019: No need for this. Just use ``-GR_M_TWOPI``.

gr-digital/lib/costas_loop_cc_impl.cc
```diff
@@ -167,15 +77,24 @@ int costas_loop_cc_impl::work(int noutput_items,
...
-
     std::vector<tag_t> tags;
     get_tags_in_range(tags,
                       0,
                       nitems_read(0),
                       nitems_read(0) + noutput_items,
                       pmt::intern("phase_est"));
 
+    // Get this out of the for loop if not used:
+    bool hasAdditionalOutputs = false;
```
rear1019: ``hasAdditionalOutputs => has_additional_outputs``. CamelCase is not used in GNU Radio’s code as far as I know.

gr-digital/lib/costas_loop_cc_impl.h
```diff
@@ -35,28 +35,59 @@ class costas_loop_cc_impl : public costas_loop_cc
...
+   where Q is larger than I. The error is then calculated
+   proportionally to these squashed constellations by the const
+   K = sqrt(2)-1.
+
+   The signal magnitude must be > 1 or K will incorrectly bias
+   the error value.
+
+   Ref: Z. Huang, Z. Yi, M. Zhang, K. Wang, "8PSK demodulation for
+   new generation DVB-S2", IEEE Proc. Int. Conf. Communications,
+   Circuits and Systems, Vol. 2, pp. 1447 - 1450, 2004.
```
rear1019: Fix indention.

gr-digital/lib/costas_loop_cc_impl.h
```diff
@@ -65,7 +96,18 @@ class costas_loop_cc_impl : public costas_loop_cc
      *  \param sample complex sample
      *  \return the phase error
      */
-    float phase_detector_snr_8(gr_complex sample) const; // for 8PSK
+    float phase_detector_snr_8(gr_complex sample) const // for 8PSK
+    {
+        const float K = (sqrtf(2.0) - 1);
+        const float snr = (sample * std::conj(sample)).real() / d_noise;
```
rear1019: Why the change from ``std::norm()`` to ``x * std::conj(x)``? (Also in ``snr_4`` and ``snr_2``)

rear1019:
> Question on gr_expj, it should just be a sin/cos but when I look at sincos.h it's wrapped with all kinds of ifdefs. I think the performance would be better if these were just inline defs but since there's all the wrappings I'm guessing there's a good reason for it. The current call versus just doing it in the loop is a huge difference so I think we could get significant boosts to everything calling gr_expj if it were simply an an inline.

Yes, the code in ``sincos.cc`` should probably be moved to ``sincos.h``. Anything using ``sincos()`` will benefit, including ``gr_expj()``.

ghostop14:
Okay, I think this is a good cut.  Least amount of Costas-specific changes, performance enhancements pushed to core GR areas to benefit everything.  I did inline sincos.h which helped the gr_expj performance.  So recapping here: 
1. ENABLE_NATIVE was added to CMake, option off by default.
2. sincos was inlined in sincos.h and sincos.cc removed from the appropriate CMake.
3. Added some constants added to math.h
4. inlined functions in costas loop to header
5. Used switch instead of function pointer (much faster)
6. Used fast complex multiply to get around all the range checking in complex.h for this function.

However, git rebasing is still giving me fits to get what I think the file edit list here for the PR right.  Moving the functions to the header seems to keep causing the conflict message.  I fix it, and it still says conflict.  And every time I try to rebase it keeps intermingling other master updates.  Unfortunately after spending a couple hours here, I think I'm at the point where I don't know how to tell git to fix it, so any help would be appreciated.

noc0lour:
@ghostop14 I rebased locally and pushed to https://github.com/noc0lour/gnuradio/tree/ghostop14-costas

You can pull & force-push that branch to your "costas" branch if you want. Make a local copy of your work first, though.

ghostop14:
Okay I think I got the repo back under control.  Files shown for the PR should only include the appropriate updates now.

CMakeLists.txt
```diff
@@ -448,6 +448,20 @@ message(STATUS "  Override with -DENABLE_INTERNAL_VOLK=ON/OFF")
...
+########################################################################
+option(ENABLE_NATIVE "Enable native build optimizations" OFF)
+IF(UNIX)
+    IF (ENABLE_NATIVE)
+		MESSAGE(STATUS "Found GNU Radio native optimization flag.  Setting native CPU optimization flags.")
+		set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -march=native -ftree-vectorize")
+		set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -march=native -ftree-vectorize")
+    ELSE (ENABLE_NATIVE)
+		MESSAGE(STATUS "Not using additional GNU Radio native architecture optimizations.")
+	ENDIF (ENABLE_NATIVE)
```
michaelld: looks like you used a tab rather than spaces ... please fix!

gnuradio-runtime/include/gnuradio/math.h
michaelld: +1 from me too for `-GR_M_TWOPI`

gr-digital/lib/costas_loop_cc_impl.cc
michaelld: +1 here from me too

ghostop14:
@michaelld, I tell ya spaces-not-tabs, snake_case not camelCase..... :)  Should be all squared up.

mbr0wn:
@ghostop14 the CX_LIMITED_RANGE PR is in, can you give this a scrub?

gr-digital/lib/costas_loop_cc_impl.cc
mbr0wn: :camel:

gr-digital/lib/costas_loop_cc_impl.h
```diff
@@ -35,28 +35,74 @@ class costas_loop_cc_impl : public costas_loop_cc
     const unsigned int d_order;
     float d_error;
     float d_noise;
+    bool d_use_snr;
+
+    void fast_cc_multiply(gr_complex& out, const gr_complex cc1, const gr_complex cc2)
```
mbr0wn: This is useful, should go into math.h or some global scope.

ghostop14:
@mbr0wn , fixed the case on the hasAdditionalOutputs to has_additional_outputs, and moved fast_cc_multiply() to math.h.  I also retested with the latest push that included CX_LIMITED_RANGE in the cmake.  Looks good.

Not sure what the noted conflicts are. The CMakeLists.txt is in sync with the latest gnuradio/master (I manually synced it up this morning then added in the changes so it should be good), and sincos.cc is removed since the functions were moved inline into sincos.h.  Let me know how this looks now and if it needs any additional updates.

CMakeLists.txt
```diff
@@ -448,6 +436,28 @@ message(STATUS "  Override with -DENABLE_INTERNAL_VOLK=ON/OFF")
 # Handle logging
 find_package(LOG4CPP REQUIRED)
 
+########################################################################
```
michaelld: Guessing this clause is the conflict in this file.

michaelld:
The conflict with sincos.cc is that it was modified and is now being deleted.

ghostop14:
Okay @michaelld , @mbr0wn... thanks to @devnulling the merge conflicts are all taken care of.  There's a ton of good wide-ranging performance improvements in this PR so I'm excited about this one.  Give it another look.  This, combined with the FFT RRC wrapper and some minor flowgraph changes I have here locally and GR performed well enough across the board for me to get ATSC working in real-time.  Once these 2 PR's are in, I'll put one together for the updates that got that working too.

CMakeLists.txt
devnulling: This should be resolved with the lastest push

ghostop14:
Hi folks, just checking in if there's anything outstanding on this one.  I've taken the same PLL enhancements from here into the other PLL blocks so as soon as this gets merged and the fast_cc_multiply is available I'll finish those updates.  The other blocks didn't have the virtual function pointer so they'll only see about a 50% improvement, but all the inline work in here will carry forward to benefit all of them as well.

gnuradio-runtime/include/gnuradio/nco.h
```diff
@@ -49,10 +49,10 @@ class nco
         phase += phase_inc;
```
michaelld: Would you please combine `step(void)` and `step(int)` into just `step (int n = 1)` ... silly to have 2 functions doing the same thing for all practical purposes. The extra `*n` will cost an extra CPU cycle or 2, but I strongly doubt it will make or break anybody's flowgraph functionality!

gnuradio-runtime/include/gnuradio/sincos.h
```diff
@@ -12,12 +12,48 @@
...
 
 #include <gnuradio/api.h>
+#include <math.h>
 
 namespace gr {
 
-// compute sine and cosine at the same time
-GR_RUNTIME_API void sincos(double x, double* sin, double* cos);
-GR_RUNTIME_API void sincosf(float x, float* sin, float* cos);
+#if defined(HAVE_SINCOS)
```
michaelld: Does it make sense to change the API on master by simplifying `sincos*` to just `template < T > sincos (const T x, T* sinx, T* cosx)` or better yet `template < T > sincos (const T x, T& sinx, T& cosx)` ... I mean, we're using C++ here which can handle functions with different arguments!

gr-digital/lib/costas_loop_cc_impl.cc
```diff
@@ -171,22 +89,46 @@ int costas_loop_cc_impl::work(int noutput_items,
...
-        optr[i] = iptr[i] * nco_out;
-
-        d_error = (*this.*d_phase_detector)(optr[i]);
+        const gr_complex nco_out = gr_expj(-d_phase);
+
+        gr::fast_cc_multiply(optr[i], iptr[i], nco_out);
+
+        // EXPENSIVE LINE with function pointer, switch was about 20% faster in testing.
+        // Left in for logic justification/reference. d_error = phase_detector_2(optr[i]);
+        switch (d_order) {
```
michaelld: A moving target here, of sorts! Before you had different `work` implementation, one of which was called based on the instantiation settings. Now it's a `switch` and `if` to determine which is called, inside `work` ... which is really inefficient!

Suggestion: Since the settings can't be changed after instantiation, create a class variable that's a pointer to the method to call based on the instantiation arguments. Then just call it here. Much tighter code & will be faster too!

ghostop14:
@michaelld I like the nco change.  I agree that shouldn't really impact things too much.  Made that change and recompiled to make sure it didn't break anything.  All looks well so I pushed an update with it.  I also reran everything through clang just to be safe.  I don't think it made any major updates but if you see something that catches your eye let me know.

For the sincos to template, I think it would be a bit more intrusive to go through everything that uses sincos for template references.  Maybe a separate PR to templatize it across all the gr code?

For the function pointer, that's what it had before. I agree it's more elegant, but when I tested and went line-by-line through the loop and measured changes in each, using the function pointer took about a pretty big 20% reduction in speed.  I'm not sure why, maybe something about dereferencing at runtime or something.  But it was definitely noticeably worse in performance with the pointer.  The switch was the trade-off between the function pointer and separate functions.  Let me know what you think.

michaelld:
> @michaelld I like the nco change. I agree that shouldn't really impact things too much. Made that change and recompiled to make sure it didn't break anything. All looks well so I pushed an update with it. I also reran everything through clang just to be safe. I don't think it made any major updates but if you see something that catches your eye let me know.

:+1:

> For the sincos to template, I think it would be a bit more intrusive to go through everything that uses sincos for template references. Maybe a separate PR to templatize it across all the gr code?

:+1: Yes that makes sense. Let's not do it in this PR.

> For the function pointer, that's what it had before. I agree it's more elegant, but when I tested and went line-by-line through the loop and measured changes in each, using the function pointer took about a pretty big 20% reduction in speed. I'm not sure why, maybe something about dereferencing at runtime or something. But it was definitely noticeably worse in performance with the pointer. The switch was the trade-off between the function pointer and separate functions. Let me know what you think.

I think the function pointer you started with is different than I'm suggesting ... but since I can't see the code any longer I honestly don't know!

What I'm suggesting now is to add to the constructor the whole `switch` and `if` to set the function pointer for the `phase_detector*` to use. Then in `work` instead of the `switch` and `if` do `(*func_ptr)(optr[i]);` ... I find it difficult to believe that this will be 20% slower than the `switch` and `if` code, since it's just a pointer to function dereference rather than the `switch` (a code jump) and `if` (another code jump) ... but, if you try this & come back that this new way is that much slower than I'll be impressed & go with the current.

gnuradio-runtime/include/gnuradio/math.h
michaelld: ping on this change

gnuradio-runtime/include/gnuradio/sincos.h
michaelld: for another PR

gr-digital/lib/costas_loop_cc_impl.h
michaelld: ping on this change

gr-digital/lib/costas_loop_cc_impl.h
michaelld: ping on this query

ghostop14:
What you're describing was what was there before.  What's in maint-3.8 is pretty close.  There's a float (costas_loop_cc_impl::*d_phase_detector)(gr_complex sample) const; that gets set in the constructor then called in the work loop.  There was definitely an improvement.  This was the throughput test results with one incremental change at a time:

Costas Loop Analysis | Measured Throughput | Percent Improvement | Incremental Improvement
-- | -- | -- | --
Original Code: | 15,211,409.00 |   |  
FMA: | 23,059,684.00 | 51.59% | 51.59%
Adding Inline Functions: | 23,902,168.00 | 57.13% | 3.65%
Replacing Fn Ptr w switch statement | 28,504,874.00 | 87.39% | 19.26%
inlining gr_expj, but still std gr_complex mult | 33,316,926.00 | 119.03% | 16.88%
inlining gr_expj and fast_cc_multiply cc mult | 42,703,628.00 | 180.73% | 49.81%

I'm with you in thinking the pointer should be the same (it's "just" a pointer) and was just as surprised when I found changing that line made such a big difference.  Not sure what code the compiler is generating but it's less efficient.

gr-digital/lib/costas_loop_cc_impl.h
rear1019: If you are wondering which changes Micheal is referring to (they seem to be buried by now):
https://github.com/gnuradio/gnuradio/pull/3076#discussion_r370948883

gr-digital/lib/costas_loop_cc_impl.h
rear1019: If you are wondering which changes Micheal is referring to (they seem to be buried by now):
https://github.com/gnuradio/gnuradio/pull/3076#discussion_r370948883

ghostop14:
@rear1019 good catches.  I think when I fixed the master/merge issue the inlines and the std::norm got lost (that was a mess to straighten out).  They're fixed.  The inlines are definitely faster, and they should definitely have been there in control_loop.  I also changed the tests you caught doing conversions to 1.0f.  Since complex is floats I don't think we're losing any precision by doing the zero test as doubles in these cases, and I'd agree I'd rather avoid the unnecessary type conversions.  Give this push a once-over.

ghostop14:
@rear1019 , @michaelld checking in if there's any additional inputs from the last push.

gnuradio-runtime/include/gnuradio/math.h
marcusmueller: I'm going ahead and will kind of veto the addition of constant `#define`s: these are worse than global scope variables, and any compiler in any optimizing step will just precompute `(2*GR_M_PI)` [at compile time](https://gcc.godbolt.org/z/Z8vrXz), even at `-O0`).

marcusmueller: ... which probably means I will just clean up this PR  a bit :)

ghostop14: Sounds good @marcusmueller.  Let me know if you need me to do anything.

devnulling:
Going to close this PR as #3203 includes the commit for this PR and supersedes it.

ThomasHabets:
@ghostop14 hi. Mind sharing exactly how you ran the benchmarks? Looks like you have some setup that could help me look for more things.

ghostop14:
Hi @ThomasHabets, sure.  A good bit of the testing for this was actually in gr-lfast (https://github.com/ghostop14/gr-lfast).  This PR was an adaptation of the enhancements first implemented there.  In that OOT, there's a lib/test_lfast.cc where you'll see how I wrapped the work functions and timed them (new versus original).  I'd actually take iterations up to 1000 to average a bit more.  There's tools like valgrind/callgrind and Intel's vtune that can help you find hotspots, but I had better luck measuring runtime and throughput and adjusting lines a little at a time (even commenting them out just to get an idea of their impact on the results then putting them back).  

There's a few wins you can look for in your OOT module right off the bat.  If you have any complex multiply ops, the standard std::complex has a very thorough (and slow) cc*cc multiplication operator to verify there's no NaN or INF's.  Most don't apply to signals use cases and just slow down what should be a quick calculation (GR master now has a fast_cc_multiply function you can use if you don't want to or add or your compiler doesn't support the cx-limited-range flag).  And if you don't need to build to deploy across different hardware (say via apt or something), you can native-build on the target hardware and pick up performance with a native build flag.  Of course this all applies to C++ modules, not python.  Try cmake with -D CMAKE_CXX_FLAGS="-fcx-limited-range -march=native -ftree-vectorize".  I'd measure before and after to see if it helps at all.  Make sure any functions in loops are inlined to avoid the stack push/pops, and I'd make sure anything that requires performance is obviously in C++, not python (cx-limited-range and native builds won't impact already-built numpy for example).  Then look at for loops and see if you can use a buffer to store intermediate calculations and see if you can pull some things out of the loop to leverage volk kernels.  If you have any cascading calculations like the PLL loops it's tough, but for others if you get creative you can sometimes refactor the loops to leverage volk kernels.  Hope that helps!

ThomasHabets:
@ghostop14 thanks. I see you broke out the relevant code to benchmark it. Sounds like a good way. The way I tested does bring in (for right and wrong) some other GNU Radio code.

```c++
#include<gnuradio/blocks/null_sink.h>
#include<gnuradio/blocks/vector_source.h>
#include<gnuradio/digital/costas_loop_cc.h>
#include<gnuradio/top_block.h>
#include<vector>

int main() {
  const auto iterations = 100;
  const auto size = 10000000;

  std::vector<gr_complex> input(size);
  std::vector<gr_complex> output(size);
  const std::vector<const void*> t = {input.data()};
  const std::vector<void*> out = {output.data()};

  for (int i = 0; i < iterations; i++) {
    auto src = gr::blocks::vector_source_c::make(input, false);
    auto costas = gr::digital::costas_loop_cc::make(3.14/100, 8, true);
    auto sink = gr::blocks::null_sink::make(8);
    auto tb = gr::make_top_block("costas");
    tb->connect(src, 0, costas,0);
    tb->connect(costas,0, sink,0);
    tb->run();
  }
}
```

I'll play around with the C++ flags, and will look into the other things you mentioned.

Thanks. Really appreciate it.

ghostop14:
Sure thing!  You may want to move everything except the run outside the for loop to narrow down what you're timing to give you a more accurate measure of throughput during a steady-state run too.

PR #3077, opened 2020-01-14
Blocks: Changed decimation() and interpolation() methods to inline

ghostop14:
Some blocks use the function call to decimation() or interpolation() in a for loop resulting in stack push/pops to get the static result every iteration.  Moving to inline should eliminate the jump and improve performance when used in loops.


rear1019:
A function defined inside a class definition is implicitely ``inline``. See [1] in second paragraph.

[1] https://en.cppreference.com/w/cpp/language/inline

ghostop14:
rear1019: I never knew that was implicit.  Decades and I've always been explicit about them.  Good to know.

PR #3086, opened 2020-01-17
gr-qtgui: always true check

ValZapod:
index is unsigned int


balister:
Looks good to me, thanks!

michaelld:
CI passes; multiple approvals; merging.

PR #3087, opened 2020-01-17
alsa_sink: fix always true check

ValZapod:
It is int snd_pcm_nonblock(snd_pcm_t *pcm, int nonblock)


ValZapod:
Also see https://github.com/alsa-project/alsa-lib/blob/master/src/pcm/pcm.c#L784

PR #3088, opened 2020-01-17
gr-dtv: atsc_equalizer_impl: bool not int

ValZapod:
The 'true' and 'false' values are implicitly cast to the integer type.
See atsc_equalizer_impl.cc

### DIFF
--- a/gr-dtv/lib/atsc/atsc_equalizer_impl.h
+++ b/gr-dtv/lib/atsc/atsc_equalizer_impl.h
@@ -56,7 +56,7 @@ private:
     unsigned short d_flags;
     short d_segno;
 
-    int d_buff_not_filled;
+    bool d_buff_not_filled;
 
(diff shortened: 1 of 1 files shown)


ValZapod:
Note: I do not know C++ vs C and bool is part of this. This variable is used ONLY here 
https://github.com/gnuradio/gnuradio/blob/51f01c0ecbd098b4346d8b5a84ec3f04f77120eb/gr-dtv/lib/atsc/atsc_equalizer_impl.cc#L65-L79
and here:
https://github.com/gnuradio/gnuradio/blob/51f01c0ecbd098b4346d8b5a84ec3f04f77120eb/gr-dtv/lib/atsc/atsc_equalizer_impl.cc#L124-L142

mbr0wn:
Skipping lack of CLA (under the 10 line radar).

PR #3089, opened 2020-01-17
gr-digital: the variable is being rewritten

ValZapod:
Ready for comments, as maybe it is a mistake somehow (the first commit)

### DIFF
--- a/gr-digital/lib/correlate_access_code_bb_impl.cc
+++ b/gr-digital/lib/correlate_access_code_bb_impl.cc
@@ -97,15 +97,11 @@ int correlate_access_code_bb_impl::work(int noutput_items,
         out[i] = t;
 
         // compute hamming distance between desired access code and current data
-        unsigned long long wrong_bits = 0;
-        unsigned int nwrong = d_threshold + 1;
-        int new_flag = 0;
-
-        wrong_bits = (d_data_reg ^ d_access_code) & d_mask;
-        nwrong = gr::blocks::count_bits64(wrong_bits);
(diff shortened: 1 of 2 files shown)


gr-digital/lib/correlate_access_code_bb_impl.cc
```diff
@@ -98,7 +98,7 @@ int correlate_access_code_bb_impl::work(int noutput_items,
 
         // compute hamming distance between desired access code and current data
         unsigned long long wrong_bits = 0;
-        unsigned int nwrong = d_threshold + 1;
+        unsigned int nwrong;
```
michaelld: Since this is C++, how about setting each variable as it is being used, instead of beforehand? That would simplify the code as well as fix little issues like this one you found (which is valid, BTW).

michaelld: Meaning, do something like:
```
unsigned long long wrong_bits = (d_data_reg ^ d_access_code) & d_mask;
unsigned int nwrong = gr::blocks::count_bits64(wrong_bits);
...
bool new_flag = (nwrong <= d_threshold);
...
```
and change `new_flag` to a bool type, since that's what it actually is (and, again, this is C++ ..)

ValZapod:
@michaelld Think of it, NULL check here is actually just dead code/will never be executed... Will fix it tomorrow, maybe FSF will sign back my application then))

michaelld:
I've requested a few more eyes just to make sure I didn't miss anything here. Changes are small and straight forward ... should be quick to review!

michaelld:
CI passes; 2 positive reviews; easy to parse and justify changes ==> Merging

PR #3095, opened 2020-01-18
examples: fixup filters cmake and polyphase channelizer demo

devnulling:
- Fixes up cmake for `gr-filters/examples` to install a few files that were missing

- Renames `channelizer_demo.grc` to be `polyphase_channelizer_demo.grc`

Minor tweaks to `polyphase_channelizer_demo.grc`:
- Sets QT layout for Freq sinks
- Adds `Noise Source` and `Throttle Block`
- Adds Labels to each of the Freq sinks
- Disables Autoscaling on the Ch2 Freq sink block 

polyphase_channelizer

polyphase_channelizer_running

### DIFF
--- a/gr-filter/examples/CMakeLists.txt
+++ b/gr-filter/examples/CMakeLists.txt
@@ -21,6 +21,7 @@ include(GrPython)
 
 # Base stuff
 GR_PYTHON_INSTALL(PROGRAMS
+    benchmark_filters.py
     channelize.py
     chirp_channelize.py
     decimate.py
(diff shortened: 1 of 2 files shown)


mormj:
Filter taps loader example doesn't resolve the relative path to the example taps file though - do we have a way in grc to put relative paths to files in a variable?

devnulling:
@mormj I dont know of an easy/clean way to use a relative path within GRC. It'd be useful to have `gnuradio-config-info` values available as a var. 

It's super hacky, but I have a work around for this specific flowgraph:

Add import block with:
`import subprocess`

For the file name use:

`subprocess.getoutput("gnuradio-config-info --prefix") + "/share/gnuradio/examples/filter/filter_taps_example_complex_bandpass_taps"`

michaelld:
All CI passes, which is always nice since this PR shouldn't change that status

@devnulling anything else you want changed in this PR? If not, let's merge it.

mormj:
@devnulling - I don't think the workaround is unreasonable - whatever it takes to have the example flowgraphs work out of the box.  Maybe that can be a separate PR though as there are probably more examples of that

devnulling:
I will submit another PR to fix the filter_taps_example_complex_bandpass_taps path.

PR #3097, opened 2020-01-18
grc: Fix eng_float default on argparse template

igorauad:
The `import` of `eng_notation` was removed from `flow_graph.py.mako` in 678a7c2ad, and so float parameter blocks were leading to flowgraph compilation error.

Since `eng_notation` is still ultimately imported in the generated python script, we can put the call to `num_to_str` in the generated script, instead of calling it during script generation.

The second commit of this PR simply removes a definition that seems to be unused and which also contains a call to `eng_notation.num_to_str`.

### DIFF
--- a/grc/core/generator/flow_graph.py.mako
+++ b/grc/core/generator/flow_graph.py.mako
@@ -303,7 +303,7 @@ def argument_parser():
 
         default = param.templates.render('make')
         if type_ == 'eng_float':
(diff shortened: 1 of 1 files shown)


devnulling:
Looks like this fixes https://github.com/gnuradio/gnuradio/issues/3107

igorauad:
Yes, @devnulling . That is the error!

PR #3102, opened 2020-01-20
Fix minor issues found by coverity

ThomasHabets:
Feel free to add me to gr-devs so that I can add labels


ThomasHabets:
Two of the commits already had it. Now the third does too (in commit message)

gr-blocks/lib/tagged_file_sink_impl.cc
```diff
@@ -63,12 +63,8 @@ tagged_file_sink_impl::tagged_file_sink_impl(size_t itemsize, double samp_rate)
                  io_signature::make(1, 1, itemsize),
                  io_signature::make(0, 0, 0)),
       d_itemsize(itemsize),
-      d_n(0),
```
michaelld: I don't understand why these need to be moved from the constructor into the class itself ... I thought it "normal" to set all class variables in the constructor ... what am I missing?

gr-blocks/lib/tagged_file_sink_impl.cc
```diff
@@ -63,12 +63,8 @@ tagged_file_sink_impl::tagged_file_sink_impl(size_t itemsize, double samp_rate)
...
       d_sample_rate(samp_rate)
 {
-    d_state = NOT_IN_BURST;
```
michaelld: I can see why these would get tagged as an issue ... why not move these into the constructor instantiations above instead?

gr-blocks/lib/tagged_file_sink_impl.cc
michaelld: Also: Why are some variables left in the constructor instantiation (d_itemsize, d_sample_rate) while others are not? Is it that those left behind are set from constructor arguments while the others are set in the class declaration? Hmmm ... not sure it really matters ... does it?

michaelld: Hmmm ... so my preference is to keep all settings in the constructor instantiation ... either above or here. Not setting in the class declaration itself. I don't see why Coverity doesn't like these ...

gr-digital/include/gnuradio/digital/constellation.h
```diff
@@ -206,7 +206,7 @@ class DIGITAL_API constellation : public boost::enable_shared_from_this<constell
     unsigned int d_arity;
     //! The factor by which the user given constellation points were
     //! scaled by to achieve an average amplitude of 1.
-    float d_scalefactor;
+    float d_scalefactor = 1.0;
```
michaelld: why this variable and not the others such as `d_dimensionality` or `d_re_min` and so forth? I see a pattern in the prior changes ... I don't see it here!

gr-blocks/lib/tagged_file_sink_impl.cc
ThomasHabets: Doesn't need to be moved. If not consting (which we aren't, here) then it's the same.

But since we have two constructors, where one forgot to set it, it seems better to me to set it in one place, than remember it in every constructor.

Coverity didn't like that in one constructor it was not set *at all*.

ThomasHabets: Wait, this was actually a reply applicable to another change.

So you're saying if there's only one constructor you prefer set by constructor?

gr-blocks/lib/tagged_file_sink_impl.cc
ThomasHabets: If you prefer. I suppose this belongs in https://github.com/ThomasHabets/greps/blob/grep0018/grep-0018-modernize-c%2B%2B.md

gr-digital/include/gnuradio/digital/constellation.h
ThomasHabets: I only took the one that was left out and not initialized, but yes better to take them all where they are common or have an obvious default. (e.g. scale 1.0)

Pushing new version of commit momentarily.

gr-blocks/lib/tagged_file_sink_impl.cc
michaelld: Hmmm ... not sure these are "modern" C++ ... thinking this is just reasonable C++ programming, not specific to C++11 or newer even.

ThomasHabets: Declaration initialization is C++11.

gr-blocks/lib/tagged_file_sink_impl.cc
michaelld: Ah ... I didn't realize there were multiple constructors. OK yeah so if there are multiple, then I can see setting it in the class itself so-as to avoid the annoying duplication issue. When I was learning C++ a -long- time ago, I was taught to not set variables directly int the class itself, but rather somehow in the constructors. What I've always done is create an "init" function that's called by each constructor sometime during construction. It then sets variables as appropriate given the state of construction. In this case, it seems like all of the variables have defaults that are not other needed during construction, so why not set them in the class itself? I'm OK with this new way, but let's be consistent with -any- such variables in this class, please!

michaelld: > So you're saying if there's only one constructor you prefer set by constructor?

Yes that'd be my preference for a single constructor.

gr-blocks/lib/tagged_file_sink_impl.cc
ThomasHabets: My preference is for default values at declare time (that is, in header files). Two other alternatives (in the presence of multiple constructors at least) is duplicate code (our current state), or let constructors call constructors (also new in C++11).

michaelld: > Declaration initialization is C++11.

Ah OK. that explains why I don't use them yet! I'm -so- behind on my C++ syntax!

gr-digital/include/gnuradio/digital/constellation.h
michaelld: Much better thx!

gr-blocks/lib/tagged_file_sink_impl.cc
michaelld: OK so what if there's a Declaration initialization as well as a constructor initialization of the same variable ... is that "bad" coding style? Which setting "wins"? Guessing the constructor initialization ... if the compiler allows it ... hmmm ....

michaelld: OK , so yes, I'd say to add this to your GREP ... now that I know/understand what it's all about :)

gr-blocks/lib/tagged_file_sink_impl.cc
michaelld: OK since we're using C++11, we can use either constructor initialization or Declaration initialization. Both have benefits and drawbacks. I think we need to discuss this in more detail in your GREP to decide on which to use when and where.

ThomasHabets:
I'll make this PR be old-style for now, and will add declaration initializer to GREP18

gr-blocks/lib/tagged_file_sink_impl.cc
rear1019: > OK so what if there's a Declaration initialization as well as a constructor initialization of the same variable ... is that "bad" coding style? Which setting "wins"? Guessing the constructor initialization ... if the compiler allows it ... hmmm ....

The compiler allows it and the constructor wins, see [1]. There, the term _default member initializer_ is used instead of _declaration initialization_. You better not look up what other kinds of initializations exist in C++ ;)

[1] https://en.cppreference.com/w/cpp/language/data_members#Member_initialization

ThomasHabets:
There, now it should be uncontroversial. :-)

gr-blocks/lib/tagged_file_sink_impl.cc
michaelld: oh lordy!

PR #3103, opened 2020-01-20
Fix CC Encoder for K > 8 and improve documentation

### LINKED ISSUE #2344: CC Encoder is limited to K <= 8
cc_encoder doesn't work if K > 8, because it uses a 256 element lookup table to compute the parity.

a) Ideally, I'd like it to support values of K up to 32. Some protocols commonly used in Amateur Radio, such as WSPR, use K=32. (I believe that K values that large can't be decoded with the Viterbi decoder, and a different algorithm is needed), At the very least, I'd like encode to be supported for K up to 32, even if decode isn't.

b) If the limitation to K <= 8 is part of the spec, then the documentation should state this and the implementation should check that K is in range.

I consider the current behavior - silently generating the wrong answer, rather than giving the right answer or an error message - is a bug,

rear1019:
The first commit should fix #2344. The second improves the documenation about the way the generator polynomials are represented. Someone should carefully proofread this. Confusion over the polynomials seems to be a known issue [1][2][3].

[1] https://lists.gnu.org/archive/html/discuss-gnuradio/2015-09/msg00001.html
[2] https://web.archive.org/web/20160407002340/https://gnuradio.org/redmine/issues/841
[3] https://lists.nongnu.org/archive/html/discuss-gnuradio/2017-03/msg00250.html

### DIFF
--- a/gr-fec/include/gnuradio/fec/cc_encoder.h
+++ b/gr-fec/include/gnuradio/fec/cc_encoder.h
@@ -71,23 +71,22 @@ namespace code {
  * samples into the encoder, and the output stream is
  * continually encoded.
  *
- * \li 'CC_TERMINATED': is a mode designed for packet-based
- * systems. This mode adds rate*(k-1) bits to the output as a
- * way to help flush the decoder.
+ * \li 'CC_TERMINATED': is a mode designed for packet-based systems. This mode
+ * flushes the encoder with K-1 bits which adds rate*(K-1) bits to the output.
+ * This improves the protection of the last bits of a block and helps the
+ * decoder.
  *
- * \li 'CC_TAILBITING': is another packet-based method. Instead of
- * adding bits onto the end of the packet, this mode will
- * continue the code between the payloads of packets by
- * pre-initializing the state of the new packet based on the
- * state of the last packet for (k-1) bits.
+ * \li 'CC_TAILBITING': is another packet-based method. Instead of adding bits
+ * onto the end of a packet (as with 'CC_TERMINATED'), this mode will
+ * pre-initialize the state of the encoder with a packet’s last (k-1) bits.
  *
  * \li 'CC_TRUNCATED': a truncated code always resets the registers
  * to the \p start_state between frames.
  *
- * A common convolutional encoder uses K=7, Rate=1/2,
- * Polynomials=[109, 79]. This is the Voyager code from NASA:
- * \li   109: b(1101101) --> 1 + x   + x^3 + x^4 + x^6
- * \li   79:  b(1001111) --> 1 + x^3 + x^4 + x^5 + x^6
+ * A common convolutional encoder uses K=7, Rate=1/2, and the polynomials
+ * \li  1 + x^2 + x^3 + x^5 + x^6
+ * \li  1 + x   + x^2 + x^3 + x^6
+ * This is the Voyager code from NASA.
  *
  * Another encoder class is provided with gr-fec called the
  * gr::fec::code::ccsds_encoder, which implements the above code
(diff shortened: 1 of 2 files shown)


gr-fec/lib/cc_encoder_impl.cc
```diff
@@ -71,6 +71,10 @@ cc_encoder_impl::cc_encoder_impl(int frame_size,
             "cc_encoder: Number of polynomials must be the same as the value of rate");
     }
 
+    if (k > 32) {
```
michaelld: Since we're here with this specific argument check, can you also add checks for other variables where it makes sense? For example: k > 0 (yes?) ... rate > 0 (yes?), start_state has to be in (0,(2^k)-1) (yes?), frame_size > 0 ... any others?

gr-fec/include/gnuradio/fec/cc_encoder.h
```diff
@@ -103,7 +103,7 @@ class FEC_API cc_encoder : virtual public generic_encoder
      * \param frame_size Number of bits per frame. If using in the
      *        tagged stream style, this is the maximum allowable
      *        number of bits per frame.
-     * \param k Constraint length (K) of the encoder.
+     * \param k Constraint length (K) of the encoder. Maximum possible value is 32.
```
michaelld: Why this limitation? Can you add why to the comment, if it's brief?

rear1019: On second thoughts this is wrong, it must be 31: Constraint length K = 32 implies a polynomial with 32 coefficients. But the sign bit (a negative value) is used to determine whether to invert the output.

gr-fec/lib/cc_encoder_impl.cc
rear1019: Yes, more checks can be added. I will do so this week.

rear1019:
Just pushed added checks, more fixes to documentation and rebased against current master.

gr-fec/include/gnuradio/fec/cc_encoder.h
```diff
@@ -71,23 +71,22 @@ namespace code {
...
+ * pre-initialize the state of the encoder with a packet’s last (k-1) bits.
  *
  * \li 'CC_TRUNCATED': a truncated code always resets the registers
  * to the \p start_state between frames.
  *
- * A common convolutional encoder uses K=7, Rate=1/2,
- * Polynomials=[109, 79]. This is the Voyager code from NASA:
- * \li   109: b(1101101) --> 1 + x   + x^3 + x^4 + x^6
- * \li   79:  b(1001111) --> 1 + x^3 + x^4 + x^5 + x^6
+ * A common convolutional encoder uses K=7, Rate=1/2, and the polynomials
```
mormj: It was nice having in the documentation how the decimal maps to the binary maps to the polynomial

michaelld:
Especially given the documentation change, I'd like a few more eyes just to triple-confirm correctness. I've tagged some more GR devs for this.

777arc:
I have copied the doc improvements into the [wiki page for this block](https://wiki.gnuradio.org/index.php/CC_Encoder_Definition).  Recall that the "block docs" on the wiki are intended to be GR's future "User Documentation", meant to show a user how to use a block, while the Doxygen docs are more for developers, and there will be a lot of overlap.  But this an example of why it's not a huge issue; it only took me a couple minutes to copy the docs over to the wiki, block's docs tend to not change very often, at least through PRs, maybe once every couple months.

michaelld:
CI passes. 2 approvals. Plenty of time for reviews. Merging.

argilo:
@michaelld The code in this pull request was not correctly formatted, which is now causing the format check to fail on https://github.com/gnuradio/gnuradio/pull/3157.

michaelld:
Well drat; sorry about that! I can't see the CI any longer, but I thought it passed the new Clang-format CI as well as the rest. I'll check more thoroughly next time & thanks for the heads up!

PR #3105, opened 2020-01-22
examples: fixup filter_tap_loader with subprocess path trick

devnulling:
Follow up PR for: https://github.com/gnuradio/gnuradio/pull/3095#issuecomment-576033781

This adds an import block with:

`import subprocess`

And sets the file path to be:

`subprocess.getoutput("gnuradio-config-info --prefix") + "/share/gnuradio/examples/filter/filter_taps_example_complex_bandpass_taps"`


devnulling:
BB failures seem unrelated to this change

PR #3113, opened 2020-01-23
grc: Display documentation for blocks without category key

velichkov:
Check that the category list is not empty before accessing its first
element. Some OOT modules does not have category key and documentation
tab is left empty because of this missing check.
```
  File "/usr/local/lib64/python3.7/site-packages/gnuradio/grc/gui/PropsDialog.py", line 210, in update_gui
    self._update_docs_page()
  File "/usr/local/lib64/python3.7/site-packages/gnuradio/grc/gui/PropsDialog.py", line 220, in _update_docs_page
    if self._block.category[0] == "Core":
IndexError: list index out of range
```

### DIFF
--- a/grc/gui/PropsDialog.py
+++ b/grc/gui/PropsDialog.py
@@ -217,7 +217,7 @@ class PropsDialog(Gtk.Dialog):
         pos = buf.get_end_iter()
 
         # Add link to wiki page for this block, at the top, as long as it's not an OOT block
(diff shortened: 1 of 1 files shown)


velichkov:
Follow up on #3003. 

The `qa_header_payload_demux` test has failed but the failure is unrelated to this change.

Can you add me to the gr-devs group?

devnulling:
@velichkov I've added you to gr-devs

PR #3114, opened 2020-01-23
Improve complex math speed by disabling NaN/Inf checks

ghostop14:
(This PR is to start the conversation on this performance enhancement)

During other testing, it was identified that basic cc*cc math was
significantly slower than doing the multiplication inline.  Research
uncovered that without the CX_LIMITED_RANGE flag set, complex math
has significant overhead in NaN and INF checking.  One researcher
quantified it as 6-times slower.  This was observed in perfromance
tuning other code as well.  I'm recommending that the
-fcx-limited-range CXX_FLAG be set to improve overall GR performance.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -449,6 +449,14 @@ message(STATUS "  Override with -DENABLE_INTERNAL_VOLK=ON/OFF")
 find_package(LOG4CPP REQUIRED)
 
 ########################################################################
+# Disable complex math NaN/INFO range checking for performance
+########################################################################
+check_cxx_compiler_flag(-fcx-limited-range HAVE_CX_LIMITED_RANGE)
+if(HAVE_CX_LIMITED_RANGE)
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fcx-limited-range")
+endif(HAVE_CX_LIMITED_RANGE)
+
+########################################################################
 # Distribute the README file
 ########################################################################
 install(


devnulling:
@ghostop14 can you please rebase this? It looks like it unintentionally includes this commit: https://github.com/gnuradio/gnuradio/commit/9fd179cc49c38c50a145a93a2f98cc74786ad660

CMakeLists.txt
```diff
@@ -448,11 +448,16 @@ message(STATUS "  Override with -DENABLE_INTERNAL_VOLK=ON/OFF")
 # Handle logging
 find_package(LOG4CPP REQUIRED)
 
+########################################################################
+# Disable complex math range checking
+########################################################################
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fcx-limited-range")
```
mbr0wn: How does this fare with clang or MSVC?

CMakeLists.txt
```diff
@@ -448,11 +448,16 @@ message(STATUS "  Override with -DENABLE_INTERNAL_VOLK=ON/OFF")
...
+
 ########################################################################
 # Distribute the README file
 ########################################################################
 install(
-    FILES README.md CONTRIBUTING.md CHANGELOG.md
+    FILES README.md README.hacking CHANGELOG.md
```
mbr0wn: As @devnulling is saying, this needs to go.

ghostop14:
Should be good now.

Adding some reference material.  This link https://medium.com/@smcallis_71148/complex-arithmetic-is-complicated-873ec0c69fc5 and the optimize options ref https://gcc.gnu.org/onlinedocs/gcc/Optimize-Options.html  provide a good overview.  I'm suspecting if it's in the spec MSVC supports it too (makes sense since it impacts complex.h) but I'd like to see some diverse testing too just to confirm all is well.  The performance boost is worth the effort I think.

CMakeLists.txt
michaelld: Is this just a C++ flag, or does it also work with C? Guessing both ... We do have some C code in GR after all ... so maybe use `add_definitions` or `add_compile_options` instead? I'll test as-is on macOS using Clang.

michaelld: looking like the answer is "no" as to whether Clang supports this flag ... still testing ...

michaelld: OK so at least AppleClang doesn't support this flag. Guessing it's more clang-generic. Have to limit it to GCC!

ghostop14:
@michaelld, think this might be a good wrapper?

IF(UNIX)
     IF(CMAKE_COMPILER_IS_GNUCC)
          set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fcx-limited-range")
     ENDIF(CMAKE_COMPILER_IS_GNUCC)
ENDIF(UNIX)

mbr0wn:
Looks a bit like MSVC settings are less granular: https://docs.microsoft.com/en-us/cpp/build/reference/fp-specify-floating-point-behavior?view=vs-2019

michaelld:
> @michaelld, think this might be a good wrapper?
> 
> IF(UNIX)
> IF(CMAKE_COMPILER_IS_GNUCC)
> set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fcx-limited-range")
> ENDIF(CMAKE_COMPILER_IS_GNUCC)
> ENDIF(UNIX)

@ghostop14 Here's what I'd advise, because this is how we do it elsewhere in the top-level `CMakeLists.txt` file:
```
...
ENDIF()
```

michaelld:
FYI looks like the clang/llvm folks are actively looking into the MSVC version using `#pragma float_control` < https://reviews.llvm.org/D72841 >

ghostop14:
Added the if wrapper.

ghostop14:
What do you think about functions like an inline cc_fast_multiply(gr_complex cc1, gr_complex cc2) function in math.h as an additional option besides the compiler directive?  If there's routines it can be used in we can be explicit about it, which would cover systems the flag doesn't work on.  The flag can be a fall-back for code that hasn't been reviewed.  Just a thought.

michaelld:
Is it the complex multiply or complex divide that's the major slowdown (or both)? I was thinking it's the latter, since the former cannot result in NaN or Inf ... maybe FP overflow or underrun, but I think those are covered by the CPU ... anyway ... I can see a use-case for specifying the math when other speed-ups aren't available ... thinking those can be separated into different PRs.

michaelld:
What happens on GCC before this flag was added? Will it error out? How old GCC does this go back to?

michaelld:
Instead of just setting the flag for GCC, a more robust way would be to check to see if it works; if so, use it; if not, don't. There are CMake functions for doing this; guessing we do some checks like this in GR already somewhere ...

ghostop14:
I saw the performance hit on multiplies.  Short of the divide block I don't think I've seen complex divide in any common modules but I'm sure it's in there somewhere else.  I did test the divide block with the flag set and it never crashed.

Any suggestions on where/how to check if the flag works in cmake?

michaelld:
@ghostop14 https://cmake.org/cmake/help/latest/module/CheckCXXCompilerFlag.html

michaelld:
Nothing like that in GR or Volk ... not yet! Had thought so, but a quick `grep` shows otherwise. It's a pretty straight forward function to call.

CMakeLists.txt
```diff
@@ -448,6 +448,13 @@ message(STATUS "  Override with -DENABLE_INTERNAL_VOLK=ON/OFF")
 # Handle logging
 find_package(LOG4CPP REQUIRED)
 
+########################################################################
+# Disable complex math range checking
+########################################################################
+IF(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fcx-limited-range")
+ENDIF()
+
```
mbr0wn:
```cmake
check_cxx_compiler_flag("-fcx-limited-range" HAVE_CX_LIMITED_RANGE)                                                                          if(HAVE_CX_LIMITED_RANGE)
    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fcx-limited-range")
endif()
```

is what we want, I think. What if clang adds the flag in some compiler version, we want CMake to figure that out for us.

mbr0wn: Also (I almost feel bad pointing that out), we have a CMake coding style too. CMake commands are lower-case, and our custom macros are uppercase.

CMakeLists.txt
mbr0wn: Let's say, we limit it to compilers that support it. I wouldn't trust all versions of gcc, either.

michaelld: Ditto for Clang & I'd guess MSVC too. Some older Clang will pass this test but then fail during the build (from experience). We have minimum compiler versions already set (and I assume checked for ...) ... might be time to review those settings on master & bump them as necessary to avoid this issue. I'd prefer less complicated code here if possible!

ghostop14:
@mbrown, don't feel bad.  Just getting the flow of the PR process so keep pointing it out.  I'll catch on... eventually!  Just incorporated those changes.  I like doing the check that way.  It's clean.

mbr0wn:
I do have one questions though: It seems like no one really knows what the switch exactly does. I spent some hours yesterday going through glibc and gcc source code, and I'm not a lot wiser. We treat it like it's disabling NaN and Inf checks, but the name of switch and its docs refer to disabling range extension. Before we walk away from this, two obvious questions for me are: a) how do we disable NaN/Inf checks for real FP, and when does the disabling of range reduction hurt?

ghostop14:
From what I read in the gcc documentation, it says:
-fcx-limited-range
    When enabled, this option states that a range reduction step is not needed when performing complex division. Also, there is no checking whether the result of a complex multiplication or division is NaN + I*NaN, with an attempt to rescue the situation in that case. The default is -fno-cx-limited-range, but is enabled by -ffast-math. 

So it only applies to complex division and multiplication.  I don't think given our data, we would run into the general use case of Nan + I*NaN for multiplication (I could be wrong)?  I think if turn it on globally, we should disable it (set it to the default -fno-cx-limited-range) for compiling divide_impl.cc only just to be safe.

ghostop14:
@mbr0wn, I had a thought this morning and I just checked.  I'm not a volk guy by any means, but looking through the volk code, it doesn't look like volk does any NaN/INF checking on its complex math currently either.  The generic 32fc_x2 multiply kernel does rely on the standard std::complex multiplier operator so it would only pick it up there.  But the kernels don't have any in the multiply or divide that I saw.  So I think we wouldn't be introducing anything new by enabling this.  We would only pick up that benefit when standard C++ is doing the complex math.  Thoughts?

mbr0wn:
@ghostop14 Yeah, hence my comment regarding regular FP math. This is for complex math only. And as for my comment, I'm curious about the range reduction step being skipped, not the Inf/NaN checks being skipped, since they seem to be an afterthought (if you just follow the documentation, and also, the switch is called `-fcx-limited-range`, and not `-fcx-skip-inf-nan-check`).

mbr0wn:
And to be clear, I still think this PR is OK, I just don't think we're "finished" with FP optimization switches, and I don't think we 100% understand this change either.

ghostop14:
I'm with ya.  If I could see the source code it would be easier to confirm.  I can't tell if the code generation would be part of the gcc source code (where the Annex G of the spec makes an example suggestion of how the compiler could generate the code), or if there's true implementation code in stdlib.  Got down a rabbit hole trying to hunt it down myself.

ghostop14:
I did find another reference here: https://en.cppreference.com/w/c/preprocessor/impl

#pragma STDC CX_LIMITED_RANGE (enabled by -fcx-limited-range)
Informs the compiler that multiplication, division, and absolute value of complex numbers may use simplified mathematical formulas (x+iy)×(u+iv) = (xu-yv)+i(yu+xv), (x+iy)/(u+iv) = [(xu+yv)+i(yu-xv)]/(u2+v2), and |x+iy| = √x2+y2, despite the possibility of intermediate overflow. In other words, the programmer guarantees that the range of the values that will be passed to those function is limited. The default value is OFF.

mbr0wn:
Ah! That is a very useful reference. BTW, glibc uses pretty much verbatim the code from Annex G. I spent some time parsing through glibc and gcc code (NOT fun), but I hadn't found where exactly the range reduction happens. Now I have the right context, though.

mbr0wn:
BTW, I didn't realize this was a standard pragma. The pragma might understood by MSVC and Clang, icc definitely understands it.

mbr0wn:
MSVC doesn't support it: https://docs.microsoft.com/en-us/cpp/preprocessor/pragma-directives-and-the-pragma-keyword?view=vs-2019

michaelld:
We'll leave the pragma as a future effort. Right now, let's just stick with the compiler flag. It should work for GCC and not for any other compiler yet, though it looks like at least Clang should support it in some not-to-distant-future version.

mbr0wn:
Before we merge, I'd like to improve the git commit message, since it's such a big change. @ghostop14 are you OK if I edit-while-merge?

ghostop14:
@mbr0wn sure thing.

mbr0wn:
Merged manually, and updated commit message.

Issue #3099, opened 2020-01-19
GRC 'ModuleNotFoundError'


duggabe:
After loading GR maint-3.8 from source, I tried to execute gnuradio-companion and got a window titled 'ModuleNotFoundError'. See https://wiki.gnuradio.org/index.php/File:ModNotFound.png
After discussions on #gnuradio and experimentation, I found the following solution:

put the following in your <code>.bashrc</code> or <code>.bash_aliases</code> file and restart your terminal.<br>
For Debian / Ubuntu, and derivatives, and other 32-bit OS use:
<pre>
export PYTHONPATH=/usr/local/lib/python3/dist-packages:/usr/local/lib/python3/site-packages
</pre>
Other 64-bit systems may require:
<pre>
export PYTHONPATH=/usr/local/lib64/python3/site-packages
</pre>

The 'ModuleNotFoundError' window needs to be updated:
1) remove reference to "library path environment variable" (it doesn't seem to be needed).
2) add the information given above for PYTHONPATH.

michaelld:
I'd advise you, in any such change, to do:
```
export PYTHONPATH=[STUFF]:$PYTHONPATH
```
just in case the `PYTHONPATH` is set externally somewhere / somehow. It's also just good practice in general to prepend, not overwrite, environment variables ... not -always-, just generally.

duggabe:
yes, I will fix that right away in https://wiki.gnuradio.org/index.php/InstallingGR#To_install_system_wide .
If you tell me where to find the code for that error message, I'll look at changing it myself.

velichkov:
https://github.com/gnuradio/gnuradio/blob/63e490434a19b757baaa0920055cde94b9362d18/grc/scripts/gnuradio-companion#L24-L34
and
https://github.com/gnuradio/gnuradio/blob/63e490434a19b757baaa0920055cde94b9362d18/grc/scripts/grcc#L23-L33

rear1019:
> The 'ModuleNotFoundError' window needs to be updated:
> 
>     1. remove reference to "library path environment variable" (it doesn't seem to be needed).

It is required if a non-standard ``CMAKE_INSTALL_PATH`` is used, i.e. anything other than ``/usr`` or ``/usr/local`` (depending on the distribution?) => Do not remove the reference to ``LD_LIBRARY_PATH``.

duggabe:
@rear1019 OK. can you help me with what it should say?

duggabe:
[...]

PR #3119, opened 2020-01-25
add a reference to https://wiki.gnuradio.org/index.php/ModuleNotFound…

duggabe:
…Error in the ModuleNotFoundError message produced by GRC and grcc
to resolve issue 3099.

### DIFF
--- a/grc/scripts/gnuradio-companion
+++ b/grc/scripts/gnuradio-companion
@@ -24,13 +24,14 @@ import warnings
 GR_IMPORT_ERROR_MESSAGE = """\
 Cannot import gnuradio.
 
-Is the model path environment variable set correctly?
+Is the Python path environment variable set correctly?
     All OS: PYTHONPATH
 
 Is the library path environment variable set correctly?
     Linux: LD_LIBRARY_PATH
     Windows: PATH
-    MacOSX: DYLD_LIBRARY_PATH
+
+See https://wiki.gnuradio.org/index.php/ModuleNotFoundError
 """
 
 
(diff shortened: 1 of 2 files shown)


grc/scripts/gnuradio-companion
```diff
@@ -24,13 +24,14 @@ import warnings
 GR_IMPORT_ERROR_MESSAGE = """\
 Cannot import gnuradio.
 
-Is the model path environment variable set correctly?
+Is the Python path environment variable set correctly?
     All OS: PYTHONPATH
 
 Is the library path environment variable set correctly?
-    Linux: LD_LIBRARY_PATH
+    Linux and MacOSX: LD_LIBRARY_PATH
```
michaelld: delete "MacOSX" ... it doesn't use LD_LIBRARY_PATH ... and setting the DYLD_LIBRARY_PATH is dangerous unless you -really- know what you're doing!

michaelld: same here as in "grcc" below

michaelld: Per our discussion last night, please add something like: “On MacOSX / macOS the library search path is set primarily by the environment variable `DYLD_LIBRARY_PATH`. Because of the way the OSX dynamic library loader works, this variable works differently than on Linux. Its primary purpose is in finding libraries in a local path, such as when testing software before installing it or inside an application. We strongly recommend against setting this variable globally.”

duggabe: added https://wiki.gnuradio.org/index.php/ModuleNotFoundError#MacOSX_.2F_macOS

duggabe:
revised grc/scripts/gnuradio-companion
revised grc/scripts/grcc
tested both on 3.9.0.0-git

devnulling:
Fixes: https://github.com/gnuradio/gnuradio/issues/3099

michaelld:
CI doesn't matter here. Small and relevant changes. 2 approvals. Merging.

PR #3122, opened 2020-01-25
gr-dtv: add missing std::nothrow

ValZapod:
std::nothrow

### DIFF
--- a/gr-dtv/lib/dvbt/dvbt_reference_signals_impl.cc
+++ b/gr-dtv/lib/dvbt/dvbt_reference_signals_impl.cc
@@ -149,7 +149,7 @@ dvbt_pilot_gen::dvbt_pilot_gen(const dvbt_configure& c)
     d_sampling_freq_correction = 0.0;
 
     // allocate PRBS buffer
-    d_wk = new char[d_Kmax - d_Kmin + 1];
+    d_wk = new (std::nothrow) char[d_Kmax - d_Kmin + 1];
     if (d_wk == NULL) {
         std::cerr << "Reference Signals, cannot allocate memory for d_wk." << std::endl;
         throw std::bad_alloc();


drmpeg:
I would rather have it changed to:

d_wk = new (std::nothrow) char[d_Kmax - d_Kmin + 1];

Which is the convention in the rest of the module.

ValZapod:
Here you are))

ValZapod:
Is this commit wrong then? https://github.com/gnuradio/gnuradio/commit/26886d80d0ade51709c4c0389fd54b813ade9ec6

drmpeg:
Thanks for changing it. As for your question about https://github.com/gnuradio/gnuradio/commit/26886d80d0ade51709c4c0389fd54b813ade9ec6, here's the story. Using std::nothrow was probably a function of keeping the old error messages when moving the gr-dvbt OOT into the GNU Radio mainline.

I'm the maintainer (and author of most of the commits) of gr-dtv, so I tried to be consistent and use std::nothrow in the rest of gr-dtv (although it's not 100%).

There's no use of std::nothrow in gr-digital, so I'd say your https://github.com/gnuradio/gnuradio/commit/26886d80d0ade51709c4c0389fd54b813ade9ec6 commit is fine.

michaelld:
I'll admit that I hadn't checked for `nothrow` in `gr-digital` for that change; we got lucky! If the main author prefers one way or the other, then the code should reflect that preference and we should try to honor it as best makes sense (IMHO). Hence, this commit makes sense in the context of `gr-dtv`s use of `new` (at least the vast majority of them).

ValZapod:
Yeah. Please look into https://github.com/gnuradio/gnuradio/issues/3125

mormj:
2 approvals and @drmpeg 's suggested change ... merging

mbr0wn:
@drmpeg can you please use the review button to approve/request changes? Given the high number of PRs we're recently having, it really helps when we don't have to read all the comments to see if people are approving PRs.

PR #3123, opened 2020-01-25
Improve Performance of PFB Arbitrary Resampler

ghostop14:
The code updates take advantage of the fact that the filter and
differential filter in the work loop can be computed independent
of each other.  A second thread is used to process the filter while
the differential filter is being calculated to cut the processing
time significantly.  Valgrind measurements with and without the change
showed showed a call cost reduction from 41% to 14%.  A significant
increase in throughput.

### DIFF
--- a/gr-filter/lib/pfb_arb_resampler.cc
+++ b/gr-filter/lib/pfb_arb_resampler.cc
@@ -38,6 +38,7 @@ namespace kernel {
 pfb_arb_resampler_ccf::pfb_arb_resampler_ccf(float rate,
                                              const std::vector<float>& taps,
                                              unsigned int filter_size)
+    : filter_thread(NULL)
 {
     d_acc = 0; // start accumulator at 0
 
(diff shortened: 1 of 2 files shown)


ghostop14:
Hold on this, use case testing is producing some choppy audio I want to work out.

gr-filter/lib/pfb_arb_resampler.cc
```diff
@@ -197,24 +200,34 @@ float pfb_arb_resampler_ccf::phase_offset(float freq, float fs)
...
-    ;
-    gr_complex o0, o1;
+    gr_complex o1;
 
     while (i_in < n_to_read) {
         // start j by wrapping around mod the number of channels
         while (j < d_int_rate) {
             // Take the current filter and derivative filter output
-            o0 = d_filters[j]->filter(&input[i_in]);
+            filter_thread = new boost::thread(
```
rear1019: I strongly doubt this will increase performance. You create a thread for every single sample to be processed. Thread creation alone will increase latency (however fast it is). Also, this requires free CPU time to be available the very moment the thread is created. If this is not the case, latency will increase even further.

ghostop14: I agree, I think that's what I ran into.  I have another way I've done it without creating the threads every call that should alleviate that.  I'm going to try to see if the performance stays.

ghostop14:
I'm going to close this for now.  It's still causing jitter even though throughput is better.  This is going to take some more research.

PR #3126, opened 2020-01-26
gr-digital: fix undefined behaviour

ValZapod:
A variable is used several times between two sequence points


mormj:
Thanks - much more clear what the code will do - I can't remember all the operator precedence

michaelld:
Simple change to provide code clarify. CI passes. 2 approvals. Merging.

Issue #3106, opened 2020-01-22
UHD: WBFM example needs to be fixed up


devnulling:
SyntaxError: unexpected EOF while parsing
ERROR:gnuradio.grc.core.FlowGraph:Failed to evaluate parameter block audio_output
Traceback (most recent call last):
  File "/home/user/sdr/installs/v3.15.0.0_master/lib/python3/dist-packages/gnuradio/grc/core/FlowGraph.py", line 216, in renew_namespace
    value = eval(parameter_block.params['value'].to_code(), namespace)
  File "<string>", line 0
    
    ^
SyntaxError: unexpected EOF while parsing

Loading: "/home/user/sdr/installs/v3.15.0.0_master/share/gnuradio/examples/uhd/uhd_wbfm_receive.grc"
ERROR:gnuradio.grc.core.FlowGraph:Failed to evaluate parameter block audio_output
Traceback (most recent call last):
  File "/home/user/sdr/installs/v3.15.0.0_master/lib/python3/dist-packages/gnuradio/grc/core/FlowGraph.py", line 216, in renew_namespace
    value = eval(parameter_block.params['value'].to_code(), namespace)
  File "<string>", line 0
    
    ^
SyntaxError: unexpected EOF while parsing

velichkov:
The `Value` in address and audio_output blocks needs to be a string, similar fixes are needed for some of the other uhd examples as well.
```diff
diff --git a/gr-uhd/examples/grc/uhd_wbfm_receive.grc b/gr-uhd/examples/grc/uhd_wbfm_receive.grc
index 20979d36b4..34a1d21780 100644
--- a/gr-uhd/examples/grc/uhd_wbfm_receive.grc
+++ b/gr-uhd/examples/grc/uhd_wbfm_receive.grc
@@ -137,7 +137,7 @@ blocks:
     label: IP Address
     short_id: a
     type: ''
-    value: addr=192.168.10.2
+    value: '"addr=192.168.10.2"'
   states:
     bus_sink: false
     bus_source: false
@@ -171,7 +171,7 @@ blocks:
     label: Audio Output Device
     short_id: O
     type: ''
-    value: ''
+    value: '""'
   states:
     bus_sink: false
     bus_source: false
```

devnulling:
@velichkov Yeah i had identified the string issue before making this post and just put it up as a reminder to myself. However, after fixing the string issue, this parameter block issue came up https://github.com/gnuradio/gnuradio/issues/3107

[...]

Issue #3107, opened 2020-01-22
Parameter block conversion error with float


devnulling:
Using a parameter block with a float type, and with a `0` value will throw an error and not generate a flowgraph

```
AttributeError: 'Undefined' object has no attribute 'num_to_str'
```

devnulling:
Fixed with #3097

PR #3131, opened 2020-01-27
examples: fixup UHD WBFM example

devnulling:
Fixes: https://github.com/gnuradio/gnuradio/issues/3106

Includes a bunch of misc changes to this flow graph.

- Removes most param blocks to variable blocks to avoid https://github.com/gnuradio/gnuradio/issues/3107
- Changes the structure of how `tun_freq` is handled. Instead of `freq/1e6` and adding an offset from another slider (`tun_fine`), the specific freq is provided. 
- Sets the `tun_freq` value to step in `100e3` steps (so the `tun_fine` is not needed)
- Sets `tun_gain` to use a `Normalized` gain value
- Changes audio multiplier (volume) to range from `0.0` to `1.0` in `0.1` steps instead of `0.0` to `10.0`
- Adds QT Chooser for Antenna selector 

uhd_wbfm

uhd_wbfm1

### DIFF
--- a/gr-uhd/examples/grc/uhd_wbfm_receive.grc
+++ b/gr-uhd/examples/grc/uhd_wbfm_receive.grc
@@ -1,6 +1,7 @@
 options:
   parameters:
     author: Example
+    catch_exceptions: 'True'
     category: Custom
     cmake_opt: ''
     comment: ''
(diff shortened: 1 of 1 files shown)


ValZapod:
Maybe change to start from 65.8 MHz? https://en.wikipedia.org/wiki/FM_broadcasting#Broadcast_bands

michaelld:
Wondering if if the `'"type=b200"'` is truly desirable ... but then the prior version had `addr=192.168.10.2` ... so maybe this is better?

michaelld:
It looks like what you did, mostly, is update the script to a more recent GRC, tweaked the USRP to be used, and fixed some of the settings ... such that quite a few of the changes are just from updating the script to a more recent GRC ... yes? I'm good with this either way; just trying to understand some of the changes. Thx!

dkozel:
I think the USRP device argument should be left blank for widest compatibility? We can merge as is or with that removed, whatever you prefer.

devnulling:
I'll push another update to this to address the issues raised.

devnulling:
I've update the example to have a blank default device arg param string.

Also updated the flowgraph to have 3 different tun_freq QT ranges, that should be enabled or displayed based on the location of the user, with notes for each as show. The example as pushed has the CCIR band enabled and others disabled as shown in the second screen cap below. First one is just for reference to the notes that are displayed when the blocks are enabled.

uhdwbfm2

uhdwbfm1

marcusmueller:
This is super nice! Thanks, @devnulling !

PR #3135, opened 2020-01-28
Changed all str.format() of modtool directory to python f'strings

### LINKED ISSUE #2383: Python String Formatting: Switch to f-strings
We should change the string formatting from `%s` or `str.format` to `f-strings` in the `Release 3.9`. It makes the code looks much cleaner.

ritvikjain99:
Python f-strings make code look cleaner and more readable. Not only this, it's faster because it's evaluated at run-time and then formatted using `__format__` unlike string formatting . Also there are some AST related advantages which make it preferable.
Fixes #2383


sbmueller:
Hi @ritvikjain99 , thanks for the work. However, it's not clear to me what this changes should solve. Can you explain a bit? Also extending the commit message with the explanation would be great.

If I'm assuming correctly, this is rather an enhancement than maintenance commit, so it probably is better suited for `master` branch.

ritvikjain99:
@sbmueller sorry for lack of detailing. I have updated the description. Thanks for pointing it out!

gr-utils/python/modtool/cli/add.py
```diff
@@ -36,7 +36,7 @@
 
 @click.command('add')
 @click.option('-t', '--block-type', type=click.Choice(ModToolAdd.block_types),
-              help="One of {}.".format(', '.join(ModToolAdd.block_types)))
+              help=f"One of {', '.join(ModToolAdd.block_types)}."
```
sbmueller: Here is a `)` missing.

marcusmueller: Oh! Syntax errors; that indicates `gr_modtool` has not  been actually exercised.

marcusmueller:
Note that this is a very welcome change! Not only do I prefer the clarity of f-strings, I find them less error-prone.

However, I won't merge this into `maint-3.8`; that's impossible, since that still allows for python2.

So, please re-submit for master, and make sure the tests pass!

PR #3136, opened 2020-01-28
Properly read AList in LDPC_matrix

kc1212:
read_alist_file is not an LDPC_matrix method.


michaelld:
This change is trivial and a no-brainer. Merging without further ado.

PR #3139, opened 2020-01-29
Add FFT-based Root Raised Cosine Convenience Wrapper

ghostop14:
This new convenience wrapper takes advantage of the performance
gains of using an FFT-Based rather than a FIR-based filter for
Root Raised Cosine filters.  The RRC taps are generated just like
the FIR-Based RRC but passed to an FFT filter instead.


michaelld:
Tagging a couple other GR devs for more eyes.

PR #3143, opened 2020-01-30
Harmonize QT Waterfall Sink argument name with freq_sink_c

### LINKED ISSUE #3142: First argument of QT frequency and waterfall sink is inconsistent
Compare
https://github.com/gnuradio/gnuradio/blob/master/gr-qtgui/include/gnuradio/qtgui/waterfall_sink_c.h#L101
and
https://github.com/gnuradio/gnuradio/blob/master/gr-qtgui/include/gnuradio/qtgui/freq_sink_c.h#L96

waterfall_sink_c has the first argument as `size`, while freq_sink_c has it as `fftsize`. This results in the python function keywords being different.

Looking at the implementation (https://github.com/gnuradio/gnuradio/blob/master/gr-qtgui/lib/waterfall_sink_c_impl.cc#L30) it looks like `fftsize` is the correct name.

Sec42:
Fixes  #3142

### DIFF
--- a/gr-qtgui/include/gnuradio/qtgui/waterfall_sink_c.h
+++ b/gr-qtgui/include/gnuradio/qtgui/waterfall_sink_c.h
@@ -85,7 +85,7 @@ public:
     /*!
      * \brief Build a complex waterfall sink.
      *
-     * \param size size of the FFT to compute and display. If using
+     * \param fftsize size of the FFT to compute and display. If using
      *        the PDU message port to plot samples, the length of
      *        each PDU must be a multiple of the FFT size.
      * \param wintype type of window to apply (see gr::fft::window::win_type)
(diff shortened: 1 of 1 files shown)


gr-qtgui/include/gnuradio/qtgui/waterfall_sink_c.h
```diff
@@ -98,7 +98,7 @@ class QTGUI_API waterfall_sink_c : virtual public sync_block
      *        the PDU message port is being used.
      * \param parent a QWidget parent object, if any
      */
-    static sptr make(int size,
+    static sptr make(int fftsize,
```
dkozel: Changing the parameter name to `fftsize` is correct/consistent with the purpose and implementation. The doxygen documentation on line 88 needs to be updated as well to match.

The GRC YAML already uses FFT Size.
https://github.com/gnuradio/gnuradio/blob/master/gr-qtgui/grc/qtgui_waterfall_sink_x.block.yml#L21

Sec42:
also fixed the doxygen comment and changed the commit message as requested. Hope it's ok now.

dkozel:
Thanks a lot @Sec42 for the fix! I hope you'll keep an eye out for other fixes.

PR #3149, opened 2020-01-31
gr-blocks: use monotonic clock for throttle

argilo:
To ensure that the Throttle block operates normally even if the system clock changes, I've switched it to use `std::chrono::steady_clock`, which is a monotonic clock.

Prior to this change, setting the system clock back would stop the throttle, and setting it ahead would allow a large burst of samples to pass through the throttle.

This change also rids the Throttle block of Boost.

### DIFF
--- a/gr-blocks/lib/throttle_impl.cc
+++ b/gr-blocks/lib/throttle_impl.cc
@@ -14,9 +14,9 @@
 
 #include "throttle_impl.h"
 #include <gnuradio/io_signature.h>
-#include <boost/thread/thread.hpp>
 #include <cstring>
 #include <limits>
+#include <thread>
 
 pmt::pmt_t throttle_rx_rate_pmt(pmt::intern("rx_rate"));
 
@@ -43,7 +43,7 @@ throttle_impl::~throttle_impl() {}
 
 bool throttle_impl::start()
 {
-    d_start = boost::get_system_time();
+    d_start = std::chrono::steady_clock::now();
     d_total_samples = 0;
     return block::start();
 }
@@ -51,13 +51,13 @@ bool throttle_impl::start()
 void throttle_impl::set_sample_rate(double rate)
 {
     // changing the sample rate performs a reset of state params
-    d_start = boost::get_system_time();
+    d_start = std::chrono::steady_clock::now();
     d_total_samples = 0;
-    d_samps_per_tick = rate / boost::posix_time::time_duration::ticks_per_second();
-    d_samps_per_us = rate / 1e6;
+    d_sample_rate = rate;
+    d_sample_period = std::chrono::duration<double>(1 / rate);
 }
 
-double throttle_impl::sample_rate() const { return d_samps_per_us * 1e6; }
+double throttle_impl::sample_rate() const { return d_sample_rate; }
 
 int throttle_impl::work(int noutput_items,
                         gr_vector_const_void_star& input_items,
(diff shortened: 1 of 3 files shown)


gr-blocks/lib/throttle_impl.cc
```diff
@@ -76,20 +76,18 @@ int throttle_impl::work(int noutput_items,
         }
     }
 
-    // calculate the expected number of samples to have passed through
-    boost::system_time now = boost::get_system_time();
-    boost::int64_t ticks = (now - d_start).ticks();
-    uint64_t expected_samps = uint64_t(d_samps_per_tick * ticks);
+    auto now = std::chrono::steady_clock::now();
+    auto expected_time = d_start + d_sample_period * d_total_samples;
```
mbr0wn: This looks wrong, can you please check? The expected time should depend on noutput_items, and `d_total_samples` looks like it's always zero.

mbr0wn:
Is it feasible to run the qa code and time the execution time of a flow graph approximately?

marcusmueller:
@mbr0wn it might, but that feels more fragile than I like my unit tests to be, considering we do parallel tests, I think

gr-blocks/lib/throttle_impl.cc
argilo: `d_total_samples` is an accumulator, and `noutput_items` is added to it on line 97.

While working on this, I did ponder whether the accumulator should be incremented *before* the sleep rather than after. It might make more sense to use the *last* sample as the time reference rather than the first.

argilo: I hesitated to change that since I didn't want to change the behaviour of the throttle, but I'd be happy to change the order if it makes sense to others.

mbr0wn:
I was thinking you start a flow graph with a bunch of samples, let it complete, and make sure it didn't finish too soon. Would have to be pretty coarse.

gr-blocks/lib/throttle_impl.cc
mbr0wn: I agree: Right now, the first call to work is never throttled.
Changing the behaviour doesn't seem to bad. First, this is for 3.9 so we're allowed to change, and second, I can't see how this would break someone's app. Finally, it makes a QA code very easy: You request a rate of say 4Hz, pipe one sample through the block, and make sure the QA takes at least 1/4 second.

mbr0wn: Oh and yeah forgot my original comment. I missed that accumulator line.

argilo: OK, I've added a second commit which defers the sleep until after copying samples and accumulating.

argilo:
I added a couple tests as well.

michaelld:
I'm tagging a few more GR devs for a few more eyes here.

gr-blocks/python/blocks/qa_throttle.py
```diff
@@ -19,9 +22,47 @@ def setUp(self):
...
+        thr = blocks.throttle(gr.sizeof_gr_complex, 10)
+        dst = blocks.vector_sink_c()
+        self.tb.connect(src, thr, dst)
+
+        start_time = time.perf_counter()
+        self.tb.run()
+        end_time = time.perf_counter()
+
+        total_time = end_time - start_time
+        self.assertGreater(total_time, 0.3)
```
mbr0wn: This is perfect, thanks.

marcusmueller:
> I was thinking you start a flow graph with a bunch of samples, let it complete, and make sure it didn't finish too soon. Would have to be pretty coarse.

That sounds like a good idea, imho.

argilo:
@marcusmueller I've already added that to this pull request. See qa_throttle.py.

gr-blocks/python/blocks/qa_throttle.py
marcusmueller: absolutely!

marcusmueller: absolutely!

marcusmueller:
@argilo as usual, awesome work! Thank you so much!

PR #3151, opened 2020-01-31
uhd: Control of frequency and gain in both directions at the same moment

ptrkrysik:
It implements ability to control frequency and gain in both directions (Tx and Rx) with use of a single UHD source or sink - in the **same moment**. It is possible to do that in separate moments with current UHD blocks but this way it is not possible to change both Rx and Tx settings for the same burst. This pull request improves this situation.

This is next incarnation of recently closed pull request #2426.
Look at my (a bit long) explanation in the last comment there in case of doubt that such functionality have any practical application.

### DIFF
--- a/gr-uhd/include/gnuradio/uhd/usrp_block.h
+++ b/gr-uhd/include/gnuradio/uhd/usrp_block.h
@@ -144,7 +144,8 @@ public:
      * \param gain the gain in dB
      * \param chan the channel index 0 to N-1
      */
-    virtual void set_gain(double gain, size_t chan = 0) = 0;
+    virtual void
+    set_gain(double gain, size_t chan = 0, pmt::pmt_t direction = pmt::PMT_NIL) = 0;
 
     /*!
      * Set the named gain on the dboard.
--- a/gr-uhd/lib/usrp_block_impl.cc
+++ b/gr-uhd/lib/usrp_block_impl.cc
@@ -638,15 +638,23 @@ void usrp_block_impl::_cmd_handler_gain(const pmt::pmt_t& gain_,
                                         int chan,
                                         const pmt::pmt_t& msg)
 {
+    // See if a direction was specified
+    pmt::pmt_t direction =
+        pmt::dict_ref(msg,
+                      cmd_direction_key(),
+                      pmt::PMT_NIL // Anything except "TX" or "RX will default to the
+                                   // messaged block direction"
+        );
+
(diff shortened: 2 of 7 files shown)


segdy:
Hi @ptrkrysik , I am struggling with your same problem (tuning frequency on TX+RX at the same time) for weeks now. The current implementation (with two separate USRP Source and USRP Sink) is super painful and I still did not get it working under all conditions (specifically with DSP retuning on RX and TX in parallel).

Your patch seems like an amazing option. Would be happy to see it implemented.

segdy:
> I do, however, see the need for this feature. I'm wondering if there's a different way to achieve this. Maybe we can have a block that takes a message and sends it to multiple UHD blocks? That might still cause commands to be executed at separate times, though.

Yes feature is absolutely useful and, in fact, just missing right now.

I disagree that there is a better way to achieve this: messages are asynchronous with non-guaranteed processing times and require hacks like "future" time stamps.

This solution allows to tune TX+RX sample-accuracy on the samples that are **currently being transmitted** by just attaching a tag with tx_command for USRP Sink and setting the time to "(nitems_read(0)+offset)/samp_rate". It's definitely something I would expect to work out of the box.

segdy:
I am afraid there is also a bug related to gain settings. I just compared it to a vanilla version and version with the patch applied.
I have a simple test flowgraph with USRP Source+Sink. I send a simple Sinusoid in and watch the response in loopback configuration (QT Frequency Sink). I enable averaging and "Max Hold"
I have sliders for setting both RX and TX gain.
What I see is the following:

- As default values I set TX Gain=25 and RX Gain=20 (I have quite some attenuation in between).
- Then I change RX Gain to 19.9 dB (not a lot!). The signal drops by 10dB
- I change back RX Gain to 20 dB but nothing happens (except increase by 0.1dB)
- Then I **decrease** the TX Gain also by 0.1dB. Now the signal **increases** again by 10dB to the original value.

I have to investigate this more but I assume it's related to the added direction functionality for the gain. Probably it changes both gains at the same time now.

segdy:
Suggesting to close this in favor of https://github.com/gnuradio/gnuradio/pull/3642.

mbr0wn:
> Suggesting to close this in favor of #3642.

Yep, will do. Thanks for taking this over.

mbr0wn:
@ptrkrysik sorry this took so long to merge. I think we'll have it soon. Thanks for this contribution!

PR #3152, opened 2020-01-31
gr-digital: Fix typo in block tree for 'Correlate Access Code - Tag'

kerel-fs:
Previously the block wasn't part of any category and thus it was
only possible to find it by searching for it using some
filter string, not by clicking trough the categories.

All credits for finding and describing this bug to Johannes Wegener
aka @hpfmn.


michaelld:
Trivial fix; no CLA needed. Merging.

PR #3154, opened 2020-02-01
Fix ATSC Flowgraphs and Allow Realtime Viewing

ghostop14:
The ATSC RX flowgraph has been fixed to allow for proper decoding.
The original flowgraph had rate mismatches between the RX filter
and the FPLL block that prevented proper decoding. This version
also has a number of performance enhancements that ultimately allow
the stream to be watched in real-time.  These changes capitalize
on new PRs for GR master branch.  The ATSC RX filter was broken
into its 2 respective parts (an RRC block and a resampler).  The
RRC block now uses the faster FFT RRC convenience wrapper, and the
PFB arb resampler was crushing performance.  So the sampling rate
was adjusted to match the ATSC symbol rate.  This removed the need
for the arb resampler in favor of a simpler rational resampler.
Processor core affinity based on a GRCon17 presentation was also
applied to separate out blocks and prevent thread migration.
Some cleanup was also done in the FPLL receiver block.  Note that
in using this in real-time, performance enhancements in PR 3076
that has not been merged into master yet are also important.  The
FPLL loop has some speedups due to the inlining of the NCO and
sincos functions, and takes advantage of the already-merged
cx_limited_range performance improvement in the FPLL loop too.
An additional change to support fast CC multiply on systems that
do not support the cx_limited_range compiler flag can be incorporated
once 3076 is merged.  This block also is better annotated for
others to understand some of the design choices.  Ultimately with
the networking block in grnet, smplayer can be pointed at the
flowgraph and TV watched in real-time.  See the flowgraph for
additional notes.

### DIFF
--- a/gr-dtv/lib/atsc/atsc_fpll_impl.cc
+++ b/gr-dtv/lib/atsc/atsc_fpll_impl.cc
@@ -41,32 +41,39 @@ int atsc_fpll_impl::work(int noutput_items,
                          gr_vector_const_void_star& input_items,
                          gr_vector_void_star& output_items)
 {
+    constexpr float alpha = 0.01;
+    constexpr float beta = alpha * alpha / 4.0;
+
     const gr_complex* in = (const gr_complex*)input_items[0];
     float* out = (float*)output_items[0];
+    float a_cos, a_sin;
+    float x;
+    gr_complex result, filtered;
 
     for (int k = 0; k < noutput_items; k++) {
-        float a_cos, a_sin;
-
         d_nco.step();                 // increment phase
         d_nco.sincos(&a_sin, &a_cos); // compute cos and sin
 
         // Mix out carrier and output I-only signal
-        gr_complex result = in[k] * gr_complex(a_sin, a_cos);
+
+        // PR Merge Note: Once the Costas Optimization PR #3076 merges, this
+        // line below helps with performance when cx_limited_range is not available
+        // such as on Macs and Windows.  Once the merge happens I'll push an update.
+        // gr::fast_cc_multiply(result, in[k], gr_complex(a_sin, a_cos));
+        result = in[k] * gr_complex(a_sin, a_cos);
+
         out[k] = result.real();
 
         // Update phase/freq error
-        gr_complex filtered = d_afc.filter(result);
-        float x = gr::fast_atan2f(filtered.imag(), filtered.real());
+        filtered = d_afc.filter(result);
+        x = gr::fast_atan2f(filtered.imag(), filtered.real());
 
         // avoid slamming filter with big transitions
-        static const float limit = GR_M_PI / 2.0;
-        if (x > limit)
-            x = limit;
-        else if (x < -limit)
-            x = -limit;
+        if (x > M_PI_2)
+            x = M_PI_2;
+        else if (x < -M_PI_2)
+            x = -M_PI_2;
 
-        static const float alpha = 0.01;
-        static const float beta = alpha * alpha / 4.0;
         d_nco.adjust_phase(alpha * x);
         d_nco.adjust_freq(beta * x);
     }
--- a/gr-dtv/examples/uhd_atsc_capture.grc
+++ b/gr-dtv/examples/uhd_atsc_capture.grc
@@ -1,6 +1,7 @@
 options:
   parameters:
     author: ''
+    catch_exceptions: 'True'
     category: Custom
     cmake_opt: ''
     comment: ''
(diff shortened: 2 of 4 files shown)


ghostop14:
Just a few more notes for anyone testing.  Grab the PR updates for the Costas loop and build with -D ENABLE_NATIVE.  Real-time worked on 7th gen i7-7700HQ laptop CPU with 4 cores (8 virtual pipelines), so any modern CPU at that level or better should do the trick.  Also a note on ATSC, it's horizontally polarized.  I used a standard ANT-900 connected to a USRP B200 with the antenna length set for 557 MHz and just set the antenna horizontally.  You'll also want to install smplayer (in apt on Ubuntu).  VLC didn't seem to handle errors in the data very gracefully but smplayer did.  There's notes in the flowgraph about turning on some of the performance settings in smplayer to allow for errors in the stream.

drmpeg:
I would prefer that the transmit flow graph not be changed. Suggesting transmitting on the actual TV band isn't recommended. 429 MHz was selected to be at least within an amateur radio band and it's cable channel 58. Also, the transmit file advatsc.ts (which is a known good Transport Stream) is referenced in the README.atsc.

ghostop14:
Hi @drmpeg, good points. I put that flowgraph back the way it was and just added a note directly within the flowgraph on where to find the file for folks like me that don't always RTFM.

ghostop14:
@drmpeg I'll make the tweaks and retest here too tonight.  Question, the FFT RRC and resampler basically split apart that ATSC RX Filter that's just a hierarchical block.  But the hier block is still in there gr-dtv.  To avoid confusion, is it worth adding the oversampling rate to the hierarchical block's parameters and incorporating these up-front changes?  Leave the ATSC RX Filter block as-is? Remove it?  I like having the blocks broken out like this in the flowgraph since if someone wanted to RFNoC the first few blocks it's pretty obvious what the flowgraph is doing.  Thoughts?

drmpeg:
I'm in favor of removing all the hierarchical block stuff. atsc_rx_filter.py, atsc_rx.py and file_atsc_rx.grc. But I think a separate pull request would be best.

ghostop14:
@drmpeg , just pushed flowgraph updates.  I actually made one more really big performance improvement in the rx flowgraph.  As soon as you mentioned the sym_rate/2 and filtering bandwidth, it gave me an idea.  I adjusted the sampling rate to match the PLL input rate but left the filter width the same.  That let me pull the resampler out altogether, and that one was the top CPU consuming block.  Try the updated ones out.  If you use the grnet block and smplayer for local viewing, watching in real-time is working really well.

drmpeg:
Excellent! This is the best I've ever seen it working.

gr-dtv/lib/atsc/atsc_fpll_impl.cc
```diff
@@ -41,32 +41,39 @@ int atsc_fpll_impl::work(int noutput_items,
                          gr_vector_const_void_star& input_items,
                          gr_vector_void_star& output_items)
 {
+    static const float alpha = 0.01;
```
michaelld: Since we're using C++11, why not make these `constexpr` & not `static`, and allow the compiler to directly substitute them like macros? I think that's roughly how this new `const` variant works ...

ghostop14: I looked it up, I like it!  Done.  Just pushed.   I like constexpr.  I'll have to keep an eye out for other places we can use it.  If we merge in 3076 we'll get a little boost in the FPLL too.  There's an inline addition for the nco.sincos() in that PR that'll help with FPLL performance here too.

michaelld: Yeah I'm a convert to `constexpr`; it's a great addition to C++! Simple little things like this & what you're doing reduce compiler challenges & make it much easier for the compiler to produce faster code.

michaelld:
I've tagged a couple other GR devs just for the sake of a few more eyes mostly on the YML.

drmpeg:
It may be useful to increase the number of taps on the FFT filter. Here's the difference between 18 (int((2*8 + 1)*sps) in the flow graph) and 50.

image

image

ghostop14:
It did definitely lower the noise floor.  And with the FFT RRC I'm not seeing any noticeable difference in CPU load with the extra taps.  It still seems hit or miss if I'm syncing on Ch28-1 or 28-2 from the stream but I think that's a separate PR to figure out.  This one's working nicely.

PR #3157, opened 2020-02-01
gr-analog: clean up examples

argilo:
I've converted the fm_tx and fm_rx examples to YML, and also made a few
corrections so that the flow graphs work properly:

* increased the output sample rate from 200k to 400k to fix distortion
* added missing deemphasis to fm_rx
* removed superfluous division by 8.0 from quadrature demod in fm_rx
* increased output volume in fm_rx from 0.2 to 1.0

After these changes, a dial tone is heard when running fm_tx followed
by fm_rx.

I also lined up the blocks in the noise_power flow graph.

### DIFF
--- a/gr-analog/examples/fm_rx.grc
+++ b/gr-analog/examples/fm_rx.grc
@@ -1,912 +1,758 @@
-<?xml version='1.0' encoding='ASCII'?>
-<flow_graph>
-  <timestamp>Thu Nov 21 20:41:12 2013</timestamp>
-  <block>
-    <key>options</key>
-    <param>
-      <key>id</key>
-      <value>fm_rx</value>
-    </param>
-    <param>
-      <key>_enabled</key>
-      <value>True</value>
-    </param>
-    <param>
-      <key>title</key>
-      <value></value>
-    </param>
-    <param>
-      <key>author</key>
-      <value></value>
-    </param>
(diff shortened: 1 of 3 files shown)


argilo:
Here's the updated flowgraphs:

fm_tx:
fm_tx

fm_rx:
fm_rx

noise_power:
noise_power

argilo:
I noticed that deemphasis was missing from the receiver, so I added that in.

drmpeg:
I believe that commercial FM broadcast used 2112 Hz for the Preemphasis High Corner Frequency. Maybe that's a better example?

argilo:
If you create a new WBFM block it's set to -1.0 by default, which means to use the default value of `0.925 * fs/2.0`. I set it to that, since the previous version of the flowgraph had `0.925 * tx_rate/2.0` anyway. Let me know if you think a non-default value would be better.

argilo:
Rebased to pick up aa1b75b83e84a221dbe4b4e4e45eb5df6bb3ec50 so that the formatting check passes.

argilo:
@devnulling I already updated the flowgraph to use the default configuration of the WBFM Transmit block, so i'm happy with it as-is.

devnulling:
@argilo great thanks!

PR #3159, opened 2020-02-02
cmake: Increase optimization level to 3 for gcc and clang

mbr0wn:
```
For gcc, increasing the optimization level adds the following
optimizations over the existing ones:

-fgcse-after-reload
-finline-functions
-fipa-cp-clone
-floop-interchange
-floop-unroll-and-jam
-fpeel-loops
-fpredictive-commoning
-fsplit-paths
-ftree-loop-distribute-patterns
-ftree-loop-distribution
-ftree-loop-vectorize
-ftree-partial-pre
-ftree-slp-vectorize
-funswitch-loops
-fvect-cost-model
-fversion-loops-for-strides

In particular, this includes loop optimizations and tree vectorization.
Some of these options can increase code size and most likely will
increase compilation times, but performance is preferable and code size
can be minimized by switching to MinSizeRel as the build target type.

In clang, the difference to -O2 isn't quite as big, but we keep the same
optimization level for symmetry.
```

What prompted me to do this was @velichkov's comment to enable tree vectorization. I thought wait, aren't we doing this already? Then I realized we default to the usual -O2. Hence this PR.

Note: I didn't profile the changes, but -O3 is still strictly standard-compliant (unlike -Ofast) and I have used it in the past. It can increase binary size and compile times, and good luck throwing gdb against it. However, we need all the performance we can get. GNU Radio is full of loops etc. that have optimization potential.


CMakeLists.txt
```diff
@@ -82,13 +82,7 @@ endif()
 ########################################################################
 # Compiler version setup
 ########################################################################
-# Append -O2 optimization flag for Debug builds (Not on MSVC since conflicts with RTC1 flag)
```
mbr0wn: This just got moved down for consistency.

CMakeLists.txt
```diff
@@ -97,6 +91,13 @@ IF(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
     ELSE()
         MESSAGE(WARNING "\nCannot determine the version of the compiler selected to build GNU Radio (GCC : ${CMAKE_CXX_COMPILER}). This build may or not work. We highly recommend using GCC version ${GCC_MIN_VERSION} or more recent.")
     ENDIF()
+    # Release flags: -O3 enables tree vectorization
+    set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -O2")
```
mbr0wn: I'm not convinced we should do debug builds at -O2. But that's a different change.

rear1019: Definitely not.

CMakeLists.txt
```diff
@@ -97,6 +91,13 @@ IF(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
...
+    set(CMAKE_CXX_FLAGS_RELEASE "-O3 -DNDEBUG")
```
rear1019: This is the default in CMake if I remember correctly, i.e. this line can/should be removed.

Is there any reason why ``${CMAKE_BUILD_TYPE}`` is *not* reused as above?

CMakeLists.txt
```diff
@@ -97,6 +91,13 @@ IF(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
...
+    set(CMAKE_CXX_FLAGS_RELWITHDEBINFO "-O3 -g -DNDEBUG")
```
rear1019: ``-O2 -g -DNDEBUG`` is the default for ``RELWITHDEBINFO`` in CMake if I remember correctly. ``-DNDEBUG`` is definitely a default which I find very, very questionable – I have been bitten by this several times…

CMakeLists.txt
```diff
@@ -97,6 +91,13 @@ IF(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
...
+    set(CMAKE_C_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -O2")
```
rear1019: This should be ``${CMAKE_C_FLAGS_DEBUG}``?

CMakeLists.txt
```diff
@@ -124,6 +125,14 @@ ELSEIF(CMAKE_CXX_COMPILER_ID MATCHES "Clang")
     ELSE()
         MESSAGE(WARNING "\nCannot determine the version of the compiler selected to build GNU Radio (${APPLE_STR}Clang : ${CMAKE_CXX_COMPILER}). This build may or not work. We highly recommend using Apple Clang version ${APPLECLANG_MIN_VERSION} or more recent, or Clang version ${CLANG_MIN_VERSION} or more recent.")
     ENDIF()
+    # Same flags as gcc. Note that -Og has a different meaning for Clang, but
+    # may be changed to be more optimized for debuggability in the future.
+    set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -O2")
+    set(CMAKE_CXX_FLAGS_RELEASE "-O3 -DNDEBUG")
+    set(CMAKE_CXX_FLAGS_RELWITHDEBINFO "-O3 -g -DNDEBUG")
+    set(CMAKE_C_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -O2")
```
rear1019: ``${CMAKE_C_FLAGS_DEBUG}``?

CMakeLists.txt
```diff
@@ -124,6 +125,14 @@ ELSEIF(CMAKE_CXX_COMPILER_ID MATCHES "Clang")
     ELSE()
         MESSAGE(WARNING "\nCannot determine the version of the compiler selected to build GNU Radio (${APPLE_STR}Clang : ${CMAKE_CXX_COMPILER}). This build may or not work. We highly recommend using Apple Clang version ${APPLECLANG_MIN_VERSION} or more recent, or Clang version ${CLANG_MIN_VERSION} or more recent.")
     ENDIF()
+    # Same flags as gcc. Note that -Og has a different meaning for Clang, but
```
rear1019: Ignoring this issue I do not think there is any reason to differentiate beween GCC and clang.

CMakeLists.txt
velichkov: I was a bit surprised to hear that `-O3` is the default for `Release` and it turns out that [it is](https://gitlab.kitware.com/cmake/cmake/blob/v3.16.3/Modules/Compiler/GNU.cmake#L56-59) but some distributions like Fedora [patch it and and change it to `-O2`](https://src.fedoraproject.org/rpms/cmake/blob/master/f/cmake-fedora-flag_release.patch)

CMakeLists.txt
velichkov: Yes `-O2 -g -DNDEBUG` is the default even on distributions like Fedora. The `-DNDEBUG` is mainly used to remove all `assert` checks from the result binaries and in my opinion it is desired flag for release builds.

CMakeLists.txt
velichkov: Good catch!

CMakeLists.txt
mbr0wn: Yup! Fedora user here.

CMakeLists.txt
mbr0wn: The reason for having this twice is that we have separate branches for gcc and clang flags already. MSVC has different flags, so these won't work. I preferred doing this over having a separate "if gcc or clang" branch, but I could also go that route.

CMakeLists.txt
mbr0wn: So, if we're all in agreement, I'll add a separate change to change this to `-Og`. But like I said, that's not what we have right now, one change at  a time, yadda yadda.

CMakeLists.txt
mbr0wn: To be clear, this line is the entire reason I'm doing this. My Fedora defaults to `-O2` and I want those juicy optimizations.

CMakeLists.txt
mbr0wn: Are you two saying you would prefer to set this to `-O2`? I would be OK with that, although I'm not sure who uses this build setting. Do distro maintainers use this to generate debug info? In that case, we'd be shipping O2-level binaries to all Ubuntu/Debian users.

marcusmueller:
aaaaah you realize that this is a change that our distro maintainers will probably happily patch away if we don't argue and document this extremely well?

CMakeLists.txt
michaelld: Clang is fairly compatible with GCC, but the reverse isn't generally true. Clang dev intentionally try to keep compatibility with GCC, while also adding in their own tweaks and options. This, while in general we do use the same settings for both Clang and GCC, we could decide to use different ones some day.

CMakeLists.txt
michaelld: I'd prefer the `RELWITHDEBINFO` settings here and below to be `-O2` ... should make debugging reasonable since those optimizations mostly follow the code ... while `-O3` does a lot of stuff that make the resulting binary objects not follow the code so closely.

rear1019:
* Whatever we do, the environment variables ``CXXFLAGS`` and ``CFLAFS`` should (must?) be respected and have precedence. This way users and distributions can mess around with the flags without having to touch ``CMakeLists.txt``.
To be honest I don’t know how, when and under which circumstances CMake respects the environment variables. According to [1][2] they are respected, but I don’t quite get under which circumstances *not*.
* CMake’s defaults [3] are quite reasonable apart from the usage ``-O3`` (which is not generally accepted) and ``-DNDEBUG`` (this is my personal preference – I don’t want ``assert()`` to be disabled in a build type called ``relwithDEBINFO``):
```
string(APPEND CMAKE_${lang}_FLAGS_INIT " ")
string(APPEND CMAKE_${lang}_FLAGS_DEBUG_INIT " -g")
string(APPEND CMAKE_${lang}_FLAGS_MINSIZEREL_INIT " -Os -DNDEBUG")
string(APPEND CMAKE_${lang}_FLAGS_RELEASE_INIT " -O3 -DNDEBUG")
string(APPEND CMAKE_${lang}_FLAGS_RELWITHDEBINFO_INIT " -O2 -g -DNDEBUG")
```
* Whether ``-O3`` is of any use – that’s a contentious issue and only benchmarking will tell. The Gentoo wiki discourages the usage, and Fedora as well, as noted above. I suspect that the Fedora maintainers will ask for benchmarks to prove that ``-O3`` is beneficial.

[1] https://cmake.org/cmake/help/v3.16/envvar/CFLAGS.html
[2] https://cmake.org/cmake/help/v3.16/envvar/CXXFLAGS.html
[3] ``/usr/share/cmake-3.16/Modules/Compiler/GNU.cmake``, applies to clang as well as far as I can tell.

PS: Just opened the link (provided above by velichkov) regarding ``-O3`` in Federa
> If you can present benchmarks that show a significant speedup for this particular code, this could be revisited on a case-by-case basis. Adding to and overriding or filtering parts of these flags is permitted if there’s a good reason to do so; the rationale for doing so must be documented in the specfile.

Well, Fedora maintainers do ask for benchmarks to enable ``-O3`` :wink:.

velichkov:
Hi @rear1019,

> Whatever we do, the environment variables CXXFLAGS and CFLAFS should (must?) be respected and have precedence. 

The `CFLAGS` and `CXXFLAGS` are respected by `cmake` but they **do not** have precedence over `CMAKE_CXX_FLAGS_<BUILD_TYPE>` as cmake puts the env variable before the `CMAKE_...` ones.

For example if you put `-O2` in the env variables and `-O3` in the cmake variables you are going to have two `-O` options on the command line and  `gcc` will compile with `-O3` (the last one). That can be very annoying but it's just how cmake works.

If you don't want any of the default options you could build with `-DCMAKE_BUILD_TYPE=None` as in Debian and specify all options in the env variables.

CMakeLists.txt
michaelld: ping on this change request, here and for Clang too

mbr0wn:
So, there's a couple of good things in here. @rear1019 I'm OK with removing `-DNDEBUG` for RelWithDebInfo.

As for -O3, the Gentoo wiki does not recommend it as a default. If or if not -O3 works with GNU Radio specifically is something we have control over.
We could also go `-O2` and then add a bunch of optimization flags, such as `ftree-vectorize` (which was what prompted me to do this in the first place. However, I looked over all the optimization flags that get added by -O3, and they all seemed sensible. GNU Radio has a lot of work functions that are fairly short and contain a lot of loops, so my gut says let's try them. BUT! Never trust your gut on optimization.

mbr0wn:
Another comment: No matter how this PR goes, we need it because we are currently getting different defaults for different distros. Optimization levels are important, and we need to be in control of them. It's no bueno if Fedora and Ubuntu already have different defaults.

CMakeLists.txt
mbr0wn: [x]

mbr0wn:
Given the various comments on -O3, I am leaning towards changing the default to -O2. This doesn't invalidate the commit, because we didn't actually have a default level, and I would like at least everyone to use the same optimization level. We'll need to do some more experiments with `-f*` flags, though it bugs me that people who build from source get potentially better performance, but I also don't see a practical way around that.

ghostop14:
So @mbr0wn, question if we explicitly set the default to -O2 but folks want -O3 when building from source, does that mean they need to edit the CMake to change it?  If we add it as a cmake command line flag and it appends to the existing and gcc gets both -O2 and -O3 which would it pick?  I think my 2 cents would be rather than downgrading the distros that default to -O3, just let it be and maybe put a note in the documentation pages somewhere on how to build with the optimizations?  We now have the ENABLE_NATIVE which defaults to off too so it may be worth some performance optimization discussion in the docs if it's not already there.

ghostop14:
@mbr0wn just had another thought looking something else up.  gcc says that going from -O2 to -O3 enables these flags: "-finline-functions -funswitch-loops -fpredictive-commoning -fgcse-after-reload
-ftree-vectorize -fipa-cp-clone".  Looking for a way around the Fedora issue, we could always check the compiler supports them and enable them in the make files and not mess with the optimization level setting.  Would have the same outcome as -O3 just without explicitly doing -O3 to keep the Fedora folks happy.

japm48:
@mbr0wn For debug builds I would suggest `-ggdb` (or `-ggdb3` if you want macro debug info, the default level for just `-ggdb` is 2). It is also compatible with clang.

More info:
https://gcc.gnu.org/legacy-ml/gcc-help/2009-02/msg00130.html
https://clang.llvm.org/docs/UsersManual.html#controlling-debugger-tuning

CMakeLists.txt
```diff
@@ -97,6 +91,18 @@ IF(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
...
+    # Release flags: -O3 enables tree vectorization
+    set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -Og")
+    set(CMAKE_CXX_FLAGS_RELEASE "-O3 -DNDEBUG")
+<<<<<<< HEAD
+    set(CMAKE_C_FLAGS_DEBUG "${CMAKE_C_FLAGS_DEBUG} -O2")
+    set(CMAKE_CXX_FLAGS_RELWITHDEBINFO "-O2 -g -DNDEBUG")
+=======
+    set(CMAKE_CXX_FLAGS_RELWITHDEBINFO "-O3 -g -DNDEBUG")
+    set(CMAKE_C_FLAGS_DEBUG "${CMAKE_C_FLAGS_DEBUG} -Og")
+>>>>>>> 0688b9902... cmake: Change debug optimization level to -Og for gcc/clang
```
marcusmueller, line 103: OOPS!

mbr0wn:
I don't think I'll be able to resolve the comments. Maybe let's leave as-is for now.

PR #3168, opened 2020-02-04
cmake: Make cmake_c*_standard unconditionally set to 11

mbr0wn:
The previous CMake would only set the C/C++ standard to 11 for gcc,
clang, and MSVC. This has two issues:
1) It prevents non-common compilers to properly function. We might hit
   the else() clause for a compiler, which then goes to C++98, which
   then doesn't work.
2) The compiler check is not necessary for setting the C/C++ standard
   and originates from an era where we had to set different compiler
   flags per compiler. Now, we're trying to beat CMake at its own game
   by assuming that it can't handle CMAKE_CXX_STANDARD properly.

This removes the compiler type check, and fully gives control to CMake
to set the C/C++ standard.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -132,28 +132,10 @@ ENDIF()
 # set after CppUnit check below). Use the variable CMAKE_CXX_STANDARD
 # since it will actually be used for this purposes starting in CMake 3.1.
 
-SET(CMAKE_C_EXTENSIONS OFF)
-SET(CMAKE_CXX_EXTENSIONS OFF)
-
-IF(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
-    SET(CMAKE_CXX_STANDARD 11)
-ELSEIF(CMAKE_CXX_COMPILER_ID MATCHES "Clang")
-    SET(CMAKE_CXX_STANDARD 11)
-ELSEIF(CMAKE_CXX_COMPILER_ID STREQUAL "MSVC")
-    SET(CMAKE_CXX_STANDARD 11)
-ELSE()
-    message(WARNING "C++ standard could not be set because compiler is not GNU, Clang or MSVC.")
-ENDIF()
-
-IF(CMAKE_C_COMPILER_ID STREQUAL "GNU")
-    SET(CMAKE_C_STANDARD 11)
-ELSEIF(CMAKE_C_COMPILER_ID MATCHES "Clang")
-    SET(CMAKE_C_STANDARD 11)
-ELSEIF(CMAKE_C_COMPILER_ID STREQUAL "MSVC")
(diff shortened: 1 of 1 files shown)


marcusmueller:
I like standardizing changes. What has changed since we merged `next`, where we still needed that magic on some platforms?

CMakeLists.txt
```diff
@@ -132,28 +132,10 @@ ENDIF()
...
-    SET(CMAKE_C_STANDARD 11)
-ELSEIF(CMAKE_C_COMPILER_ID STREQUAL "MSVC")
-    SET(CMAKE_C_STANDARD 11)
-ELSE()
-    message(WARNING "C standard could not be set because compiler is not GNU, Clang or MSVC.")
-ENDIF()
+set(CMAKE_C_EXTENSIONS OFF)
+set(CMAKE_CXX_EXTENSIONS OFF)
+set(CMAKE_CXX_STANDARD 11)
+set(CMAKE_C_STANDARD 11)
```
michaelld: Does ordering here matter? C first then C++? Whatevers ...

CMakeLists.txt
```diff
@@ -132,28 +132,10 @@ ENDIF()
...
-    SET(CMAKE_C_STANDARD 11)
-ELSEIF(CMAKE_C_COMPILER_ID MATCHES "Clang")
-    SET(CMAKE_C_STANDARD 11)
-ELSEIF(CMAKE_C_COMPILER_ID STREQUAL "MSVC")
-    SET(CMAKE_C_STANDARD 11)
-ELSE()
-    message(WARNING "C standard could not be set because compiler is not GNU, Clang or MSVC.")
-ENDIF()
+set(CMAKE_C_EXTENSIONS OFF)
+set(CMAKE_CXX_EXTENSIONS OFF)
```
michaelld: How about also adding in `set(CMAKE_CXX_STANDARD_REQUIRED TRUE)` and `set(CMAKE_C_STANDARD_REQUIRED TRUE)`, just to make it clear that C11/C++11 is what we're using. I don't think we want to let C++14 or newer in just yet.

mbr0wn: I would almost say that's another commit, then. But good idea, I didn't even know those settings existed.

marcusmueller:
Since this is master, I think I've been overthinking this: if any platform doesn't work with CMake setting the C++ standard to C++11 out of the box, it really just means that the platform won't be compat with GR 3.9.

mbr0wn:
> I like standardizing changes. What has changed since we merged next, where we still needed that magic on some platforms?

To close the loop on this, we used to have compiler-specific flags set instead using the CMake setting, which was introduced in some 3.x version of CMake.

PR #3169, opened 2020-02-04

mormj:

One example is synchronizing multiple USRP objects - sometimes you want
different sync than what is offered in the multi-usrp object, so you can
put a bit of code in the snippet block to do the custom synchronization

-----------------------------------
This supercedes #2779 and #2814 - hopefully the issues identified on those PRs have been sufficiently resolved, as the details of last iteration were worked out with @dkozel and @marcusmueller at the ESA hackfest.

- Went back to a single snippet block.  This way they can be easily arranged and copied/pasted
- Priority was added to the block so multiple snippets added to the same section can be sorted by priority
- Reduce the possible placements in the rendered flowgraph to inside the main function (can add more later if requested)
- Cleaned up the rendered flowgraph to put the snippet code within their own functions

image

image

### DIFF
--- a/gr-blocks/examples/CMakeLists.txt
+++ b/gr-blocks/examples/CMakeLists.txt
@@ -9,6 +9,7 @@ install(
   FILES
   matrix_multiplexer.grc
   peak_detector2.grc
+  py_snippets_demo.grc
   selector.grc
   test_stream_mux_tags.grc
   vector_source_with_tags.grc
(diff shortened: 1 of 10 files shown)


devnulling:
@mormj Can you please add `py_snippets_demo.grc` to the install list for cmake here: https://github.com/gnuradio/gnuradio/blob/master/gr-blocks/examples/CMakeLists.txt

mormj:
@devnulling thanks for staying on top of the examples - too easy to forget.  Should be good to go now.

michaelld:
@devnulling please go ahead and merge if this meets your change request

PR #3170, opened 2020-02-04
uhd: Add support for UHD 4.0 feature: RFNoC

mbr0wn:
This merges the code required to run UHD 4.0 RFNoC features within GNU
Radio. For UHD 4.0, this replaces the requirement to install gr-ettus
(the OOT module for experimental features produced by Ettus Research
/ NI).

The features added by this commit will only be compiled and included in
gr-uhd if the underlying UHD version is at least 4.0.

This commit includes the following components:

- Block controllers for radio, DDC/DUC, FFT block (C++ and GRC)
- Blocks for streamers (ingress/egress)
- GRC bindings for GRC-based image builder
- GRC domains for RFNoC, ctrl_port, timekeeper bus, and radio
  connections

### DIFF
--- a/gr-uhd/CMakeLists.txt
+++ b/gr-uhd/CMakeLists.txt
@@ -11,6 +11,11 @@
 include(GrBoost)
 
 find_package(UHD "3.9.7")
+if ("${UHD_VERSION}" VERSION_GREATER_EQUAL "4")
+    set(ENABLE_UHD_RFNOC TRUE)
+else()
+    set(ENABLE_UHD_RFNOC FALSE)
+endif()
 
 ########################################################################
 # Register component
@@ -24,14 +29,14 @@ GR_REGISTER_COMPONENT("gr-uhd" ENABLE_GR_UHD
     ENABLE_GR_BLOCKS
     ENABLE_GR_ANALOG
 )
-message(STATUS "  UHD Version: ${UHD_VERSION}")
-
 SET(GR_PKG_UHD_EXAMPLES_DIR ${GR_PKG_DATA_DIR}/examples/uhd)
 
 ########################################################################
 # Begin conditional configuration
 ########################################################################
 if(ENABLE_GR_UHD)
+message(STATUS "  UHD Version: ${UHD_VERSION}")
+message(STATUS "  UHD 4.0 RFNoC enabled: ${ENABLE_UHD_RFNOC}")
 
 ########################################################################
 # Add subdirectories
--- a/gr-uhd/grc/uhd_rfnoc_graph.block.yml
+++ b/gr-uhd/grc/uhd_rfnoc_graph.block.yml
@@ -1,0 +1,127 @@
+id: uhd_rfnoc_graph
+label: RFNoC Graph (Device)
+flags: [ show_id, python ]
+
+templates:
+  imports: |-
+      from gnuradio import uhd
+  var_make: |
+      self.rfnoc_graph = ${id} = uhd.rfnoc_graph(uhd.device_addr(",".join((${dev_addr}, ${dev_args}))))
+
+value: ${ 'RFNoC Graph' }
+
+parameters:
+- id: dev_addr
+  label: Device Address
+  dtype: string
+  default: ""
+  hide: ${ 'none' if dev_addr else 'part' }
+- id: dev_args
+  label: Device Args
+  dtype: string
+  default: ""
+  hide: ${ 'none' if dev_args else 'part' }
+- id: num_mboards
+  label: Num Mboards
+  dtype: int
+  default: 1
+  options: [1, 2, 3, 4, 5, 6, 7, 8]
+  hide: part
+- id: clock_source_0
+  label: 'Mb0: Clock Source'
+  dtype: string
+  options: ['', internal, external, mimo, gpsdo]
+  option_labels: [Default, Internal, External, MIMO Cable, O/B GPSDO]
+  hide: ${ 'all' if not (num_mboards > 0) else ( 'none' if clock_source_0 else 'part')}
+- id: time_source_0
+  label: 'Mb0: Time Source'
+  dtype: string
+  options: ['', external, mimo, gpsdo]
+  option_labels: [Default, External, MIMO Cable, O/B GPSDO]
+  hide: ${ 'all' if not (num_mboards > 0) else ('none' if time_source_0 else 'part')}
+- id: clock_source_1
+  label: 'Mb1: Clock Source'
+  dtype: string
+  options: ['', internal, external, mimo, gpsdo]
+  option_labels: [Default, Internal, External, MIMO Cable, O/B GPSDO]
+  hide: ${ 'all' if not (num_mboards > 1) else ( 'none' if clock_source_1 else 'part')}
+- id: time_source_1
+  label: 'Mb1: Time Source'
+  dtype: string
+  options: ['', external, mimo, gpsdo]
+  option_labels: [Default, External, MIMO Cable, O/B GPSDO]
+  hide: ${ 'all' if not (num_mboards > 1) else ('none' if time_source_1 else 'part')}
+- id: clock_source_2
+  label: 'Mb2: Clock Source'
+  dtype: string
+  options: ['', internal, external, mimo, gpsdo]
+  option_labels: [Default, Internal, External, MIMO Cable, O/B GPSDO]
+  hide: ${ 'all' if not (num_mboards > 2) else ( 'none' if clock_source_2 else 'part')}
+- id: time_source_2
+  label: 'Mb2: Time Source'
+  dtype: string
+  options: ['', external, mimo, gpsdo]
+  option_labels: [Default, External, MIMO Cable, O/B GPSDO]
+  hide: ${ 'all' if not (num_mboards > 2) else ('none' if time_source_2 else 'part')}
+- id: clock_source_3
+  label: 'Mb3: Clock Source'
+  dtype: string
+  options: ['', internal, external, mimo, gpsdo]
+  option_labels: [Default, Internal, External, MIMO Cable, O/B GPSDO]
+  hide: ${ 'all' if not (num_mboards > 3) else ( 'none' if clock_source_3 else 'part')}
+- id: time_source_3
+  label: 'Mb3: Time Source'
+  dtype: string
+  options: ['', external, mimo, gpsdo]
+  option_labels: [Default, External, MIMO Cable, O/B GPSDO]
+  hide: ${ 'all' if not (num_mboards > 3) else ('none' if time_source_3 else 'part')}
+- id: clock_source_4
+  label: 'Mb4: Clock Source'
+  dtype: string
+  options: ['', internal, external, mimo, gpsdo]
+  option_labels: [Default, Internal, External, MIMO Cable, O/B GPSDO]
+  hide: ${ 'all' if not (num_mboards > 4) else ( 'none' if clock_source_4 else 'part')}
+- id: time_source_4
+  label: 'Mb4: Time Source'
+  dtype: string
+  options: ['', external, mimo, gpsdo]
+  option_labels: [Default, External, MIMO Cable, O/B GPSDO]
+  hide: ${ 'all' if not (num_mboards > 4) else ('none' if time_source_4 else 'part')}
+- id: clock_source_5
+  label: 'Mb5: Clock Source'
+  dtype: string
+  options: ['', internal, external, mimo, gpsdo]
+  option_labels: [Default, Internal, External, MIMO Cable, O/B GPSDO]
+  hide: ${ 'all' if not (num_mboards > 5) else ( 'none' if clock_source_5 else 'part')}
+- id: time_source_5
+  label: 'Mb5: Time Source'
+  dtype: string
+  options: ['', external, mimo, gpsdo]
+  option_labels: [Default, External, MIMO Cable, O/B GPSDO]
+  hide: ${ 'all' if not (num_mboards > 5) else ('none' if time_source_5 else 'part')}
+- id: clock_source_6
+  label: 'Mb6: Clock Source'
+  dtype: string
+  options: ['', internal, external, mimo, gpsdo]
+  option_labels: [Default, Internal, External, MIMO Cable, O/B GPSDO]
+  hide: ${ 'all' if not (num_mboards > 6) else ( 'none' if clock_source_6 else 'part')}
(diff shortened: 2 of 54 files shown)


mbr0wn:
@atrnati check this out.

mormj:
Do we have a way of getting UHD 4.0 into the CI?

gr-uhd/CMakeLists.txt
```diff
@@ -37,6 +42,7 @@ GR_REGISTER_COMPONENT("gr-uhd" ENABLE_GR_UHD
     ENABLE_GR_ANALOG
 )
 message(STATUS "  UHD Version: ${UHD_VERSION}")
```
michaelld: I know this isn't necessarily part of your changes, but I'd ask you add `if(UHD_FOUND)` around these message lines. If UHD isn't found, then we don't want to be printing these!

gr-uhd/CMakeLists.txt
```diff
@@ -37,6 +42,7 @@ GR_REGISTER_COMPONENT("gr-uhd" ENABLE_GR_UHD
...
+message(STATUS "  UHD 4.0 RFNoC enabled: ${ENABLE_UHD_RFNOC}")
 
 SET(GR_PKG_UHD_EXAMPLES_DIR ${GR_PKG_DATA_DIR}/examples/uhd)
```
michaelld: Likewise, please move this variable setting into the `if(ENABLE_GR_UHD)` below, since it's not needed unless UHD is enabled.

gr-uhd/lib/CMakeLists.txt
```diff
@@ -20,13 +20,29 @@
 ########################################################################
 # Setup library
 ########################################################################
-add_library(gnuradio-uhd
+set(LIB_GR_UHD_SOURCES
     usrp_block_impl.cc
     usrp_source_impl.cc
```
michaelld: since you're reordering headers into alphabetical order, would you please also reorder these?

gr-uhd/swig/uhd_swig.i
```diff
@@ -44,6 +44,12 @@
 ////////////////////////////////////////////////////////////////////////
 %ignore gr::uhd::usrp_sink::get_device;
 %ignore gr::uhd::usrp_source::get_device;
+// Also ignore some other UHD types
```
michaelld: These and the other similar RFNoC additions need to be surrounded by `#ifdef ENABLE_UHD_RFNOC` or the equivalent. Otherwise they will be incorporated into the SWIG C++ and then become accessible in Python regardless of whether they are in the gr-uhd library.

michaelld: These are easy to add the `#ifdef` around, since they are in the top-level scope.

gr-uhd/swig/uhd_swig.i
```diff
@@ -52,6 +58,15 @@
 #include <gnuradio/uhd/usrp_source.h>
 #include <gnuradio/uhd/usrp_sink.h>
 #include <gnuradio/uhd/amsg_source.h>
+#include <gnuradio/uhd/rfnoc_block.h>
```
michaelld: The `#ifdef` around these might need in its own `%{` scope ... have to experiment to see if `#ifdef` works inside it or not.

gr-uhd/swig/uhd_swig.i
```diff
@@ -132,21 +147,42 @@
 %include <gnuradio/uhd/usrp_source.h>
 %include <gnuradio/uhd/usrp_sink.h>
 %include <gnuradio/uhd/amsg_source.h>
+%include <gnuradio/uhd/rfnoc_graph.h>
```
michaelld: `#ifdef` around these RFNoC headers

gr-uhd/swig/uhd_swig.i
```diff
@@ -132,21 +147,42 @@
...
+%include <gnuradio/uhd/rfnoc_block_generic.h>
+%include <gnuradio/uhd/rfnoc_ddc.h>
+%include <gnuradio/uhd/rfnoc_duc.h>
+%include <gnuradio/uhd/rfnoc_rx_radio.h>
+%include <gnuradio/uhd/rfnoc_tx_radio.h>
 
 GR_SWIG_BLOCK_MAGIC2(uhd, usrp_source)
 GR_SWIG_BLOCK_MAGIC2(uhd, usrp_sink)
 GR_SWIG_BLOCK_MAGIC2(uhd, amsg_source)
+GR_SWIG_BLOCK_MAGIC2(uhd, rfnoc_graph)
```
michaelld: and around these ...

gr-uhd/lib/rfnoc_graph_impl.cpp
```diff
@@ -0,0 +1,193 @@
...
+        auto streamer = _graph->create_tx_streamer(num_ports, args);
+        const std::string streamer_id =
+            std::dynamic_pointer_cast<::uhd::rfnoc::node_t>(streamer)->get_unique_id();
+        _tx_streamers.insert({ streamer_id, streamer });
+        return streamer;
+    }
+
+    void commit()
+    {
+        if (!_commit_called.exchange(true)) {
```
atrnati: Interesting use of `exchange()`--do you expect `commit()` to be called from multiple threads and are trying to protect against multiple commits?

atrnati: If so, is that the only call that can get called from multiple threads? Do the rest of the functions that mutate state also need to be concerned?

gr-uhd/lib/rfnoc_graph_impl.cpp
```diff
@@ -0,0 +1,193 @@
...
+                             const int device_select,
+                             const int block_select)
+    {
+        std::string block_hint = block_name;
+        if (device_select >= 0) {
+            block_hint = str(boost::format("%d/%s") % device_select % block_hint);
+        }
+        if (block_select >= 0) {
+            block_hint = str(boost::format("%s#%d") % block_name % block_select);
+        }
```
atrnati: Is there ever a case where both `device_select` and `block_select` are `>= 0`? Because if so, it doesn't seem like this would generate the right `device/name#block` string combination in that case.

mbr0wn:
@mormj Maybe after we release UHD 4.0 :smile:

gr-uhd/lib/rfnoc_graph_impl.cpp
mbr0wn: Ouch.

gr-uhd/grc/uhd_rfnoc_graph.block.yml
```diff
@@ -0,0 +1,127 @@
+id: uhd_rfnoc_graph
+label: RFNoC Graph (Device)
+flags: [ show_id, python ]
```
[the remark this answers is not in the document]
mbr0wn: The default flags are fine for the other blocks, and the default flags imply 'python'. For these, we need to set non-default flags, and thus need to provide 'python' ourselves.

gr-uhd/include/gnuradio/uhd/rfnoc_rx_radio.h
```diff
@@ -0,0 +1,67 @@
...
+                     const int device_select,
+                     const int instance);
+
+    virtual double set_rate(const double rate) = 0;
+    virtual void set_antenna(const std::string& antenna, const size_t chan) = 0;
+    virtual double set_frequency(const double frequency, const size_t chan) = 0;
+    virtual void set_tune_args(const ::uhd::device_addr_t& args, const size_t chan) = 0;
+    virtual double set_gain(const double gain, const size_t chan) = 0;
+    virtual double
+    set_gain(const double gain, const std::string& name, const size_t chan) = 0;
```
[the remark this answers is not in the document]
mbr0wn: The clang-format config file for GNU Radio is different than it is for UHD... I'm sometimes surprised myself, but the CI seems to think it's fine so I'll leave it.

gr-uhd/lib/rfnoc_graph_impl.cpp
mbr0wn: commit() will be called from all the streamers (at least, we might add more commits in the future). The block controllers are fine, they have mutexes either in UHD, or they are uniquely accessing a block controller. rfnoc_graph is the only one that gets accessed from many threads, and I think I have all the mutexes in place.

gr-uhd/lib/rfnoc_graph_impl.cpp
mbr0wn: :medal_sports: for spotting that just by reading the code.

gr-uhd/lib/rfnoc_graph_impl.cpp
atrnati: Okay, works for me!

gr-uhd/include/gnuradio/uhd/rfnoc_duc.h
```diff
@@ -0,0 +1,59 @@
...
+#define INCLUDED_GR_UHD_RFNOC_DUC_H
+
+#include <gnuradio/uhd/api.h>
+#include <gnuradio/uhd/rfnoc_block.h>
+#include <gnuradio/uhd/rfnoc_graph.h>
+
+namespace gr {
+namespace uhd {
+
+/*! RFNoC Digital Unconverter Block
```
marcusmueller: Typo, please fix

gr-uhd/lib/rfnoc_graph_impl.cpp
```diff
@@ -0,0 +1,214 @@
+/* -*- c++ -*- */
```
marcusmueller: Filename wrong, please rename!

gr-uhd/lib/rfnoc_graph_impl.cpp
```diff
@@ -0,0 +1,214 @@
...
+#include "config.h"
+#endif
+
+#include "gr_uhd_common.h"
+#include <gnuradio/uhd/rfnoc_block.h>
+#include <gnuradio/uhd/rfnoc_graph.h>
+#include <uhd/rfnoc/mb_controller.hpp>
+#include <uhd/rfnoc_graph.hpp>
+#include <unordered_map>
+#include <boost/make_shared.hpp>
```
marcusmueller: C++11 has `std::make_shared`, please don't introduce boost dependencies here.

gr-uhd/lib/rfnoc_rx_streamer_impl.cc
```diff
@@ -0,0 +1,192 @@
...
+    return true;
+}
+
+int rfnoc_rx_streamer_impl::work(int noutput_items,
+                                 gr_vector_const_void_star& input_items,
+                                 gr_vector_void_star& output_items)
+{
+    const size_t max_num_items_to_rx = noutput_items * d_vlen;
+    const size_t num_items_recvd =
+        d_streamer->recv(output_items, max_num_items_to_rx, d_metadata, d_timeout);
```
marcusmueller: no. I'm not letting Travis jump through hoops to ensure `gr-iio` not blocking in `work` explaining that gr-uhd does that for legacy reasons, since it was added prior to the self-call timeout and the opportunity to wake up the block executor after a timeout, with Ettus then re-introducing more of this.

gr-uhd/lib/rfnoc_rx_streamer_impl.cc
```diff
@@ -0,0 +1,192 @@
...
+    d_start_time_set = true;
+    d_start_time = time;
+}
+
+/******************************************************************************
+ * Helpers
+ *****************************************************************************/
+void rfnoc_rx_streamer_impl::flush()
+{
+    constexpr size_t nbytes = 4096;
```
marcusmueller: Unexplained magic constants in code?

gr-uhd/lib/rfnoc_rx_streamer_impl.cc
```diff
@@ -0,0 +1,192 @@
...
+    const size_t nchan = d_streamer->get_num_channels();
+    std::vector<std::vector<uint8_t>> buffs(nchan, std::vector<uint8_t>(nbytes));
+
+    gr_vector_void_star outputs;
```
marcusmueller: could have called the `std::vector` ctor that accepts a length, and avoided all the pushback's, but that's just a matter of preference in a low-frequency function like this :)

gr-uhd/lib/rfnoc_rx_streamer_impl.h
```diff
@@ -0,0 +1,73 @@
...
+                           const bool issue_stream_cmd_on_start);
+    ~rfnoc_rx_streamer_impl();
+
+    std::string get_unique_id() const { return d_unique_id; }
+
+    /***** API ***************************************************************/
+    void set_start_time(const ::uhd::time_spec_t& time);
+
+    /***** GNU Radio API *****************************************************/
+    bool check_topology(int, int) override;
```
marcusmueller: nice usage of `override`!

gr-uhd/include/gnuradio/uhd/rfnoc_graph.h
```diff
@@ -0,0 +1,135 @@
...
+    // \param source Clock source (e.g., "internal")
+    // \param mb_index Motherboard index, starting at 0
+    virtual void set_clock_source(const std::string& source, const size_t mb_index) = 0;
+
+    //! Return a reference to the block with a given block ID
+    //
+    // This allows to retrieve a shared pointer to a block controller with a
+    // limit. It keeps track of the references given out, and will throw an
+    // exception if called more than max_ref_count times. This is to help ensure
+    // that GNU Radio blocks don't share block controllers unintentionally.
```
marcusmueller: if that's the motivation, it'd be a sign of bad design: wouldn't you use a `unique_ptr` if you got that problem?

gr-uhd/include/gnuradio/uhd/rfnoc_graph.h
```diff
@@ -0,0 +1,135 @@
...
+
+/*! GNU Radio-specific wrapper for uhd::rfnoc::rfnoc_graph
+ *
+ * This wraps uhd::rfnoc::rfnoc_graph for simpler insertion into GNU Radio flow
+ * graphs. All API calls match those on said class.
+ */
+class GR_UHD_API rfnoc_graph
+{
+public:
+    using sptr = boost::shared_ptr<rfnoc_graph>;
```
marcusmueller: why no `std::` instead? We do not likes them ugly boosts, stupid hobbitssesss..

PR #3171, opened 2020-02-04
Blocks: Add DC Spike Removal and IQ Swap Native Blocks

ghostop14:
This block incorporates the OOT modules in correctiq that provide
3 different techniques to remove the DC spike inherent in IQ sampling.
The first technique mirrors SDR GUI receivers with an IIR filter
approach, the second provides a time-limited IIR approach where after
a user-configurable number of seconds locks into a basic offset
correction to eliminate the effect of the filter on the signal while
maintaining the correction, and the last is manual I and Q
configurable offsets.  The Swap IQ block is a drop-in block to help
correct for inverted spectrums and just swaps I<->Q.

### DIFF
--- a/gr-blocks/grc/blocks_correctiq.block.yml
+++ b/gr-blocks/grc/blocks_correctiq.block.yml
@@ -1,0 +1,20 @@
+id: blocks_correctiq
+label: Remove DC Spike
+category: '[Core]/IQ Correction'
+
+templates:
+  imports: from gnuradio import blocks
+  make: blocks.correctiq(${})
+
+inputs:
+-   domain: stream
+    dtype: complex
+
+outputs:
+-   domain: stream
+    dtype: complex
+
+documentation: |-
+    This block to removes that center frequency IQ DC spike with an IIR filter.
+
+file_format: 1
--- a/gr-blocks/include/gnuradio/blocks/correctiq.h
+++ b/gr-blocks/include/gnuradio/blocks/correctiq.h
@@ -1,0 +1,44 @@
+/* -*- c++ -*- */
+/*
+ * Copyright 2020 Free Software Foundation, Inc.
+ *
+ * This file is part of GNU Radio
+ *
+ * SPDX-License-Identifier: GPL-3.0-or-later
+ *
+ */
+
+#ifndef INCLUDED_CORRECTIQ_CORRECTIQ_H
+#define INCLUDED_CORRECTIQ_CORRECTIQ_H
+
+#include <gnuradio/blocks/api.h>
+#include <gnuradio/sync_block.h>
+
+namespace gr {
+namespace blocks {
+
+/*!
+ * \brief This block to removes that center frequency IQ DC spike with an IIR filter.
+ * \ingroup iq_correction
+ *
+ */
+class BLOCKS_API correctiq : virtual public gr::sync_block
+{
+public:
+    typedef boost::shared_ptr<correctiq> sptr;
+
+    /*!
+     * \brief Return a shared_ptr to a new instance of correctiq::correctiq.
+     *
+     * To avoid accidental use of raw pointers, correctiq::correctiq's
+     * constructor is in a private implementation
+     * class. correctiq::correctiq::make is the public interface for
+     * creating new instances.
+     */
+    static sptr make();
+};
+
+} // namespace blocks
+} // namespace gr
+
+#endif /* INCLUDED_CORRECTIQ_CORRECTIQ_H */
--- a/gr-blocks/include/gnuradio/blocks/correctiq_auto.h
+++ b/gr-blocks/include/gnuradio/blocks/correctiq_auto.h
@@ -1,0 +1,55 @@
+/* -*- c++ -*- */
+/*
+ * Copyright 2020 Free Software Foundation, Inc.
+ *
+ * This file is part of GNU Radio
+ *
+ * SPDX-License-Identifier: GPL-3.0-or-later
+ *
+ */
+
+#ifndef INCLUDED_CORRECTIQ_CORRECTIQ_AUTO_H
+#define INCLUDED_CORRECTIQ_CORRECTIQ_AUTO_H
+
+#include <gnuradio/blocks/api.h>
+#include <gnuradio/sync_block.h>
+
+namespace gr {
+namespace blocks {
+
+/*!
+ * \brief This block to removes that center frequency IQ DC spike with a slight variation.
+ * \p It automatically calculates the offset then switches to straight DC offset mode to
+ * \p prevent any possible IIR filtering after it's been tuned.  However, if frequency
+ * \p or upstream gain is changed, it must retune, so frequency and upstream gain are
+ * \p all taken as parameters and monitored for changes.
+ * \ingroup iq_correction
+ *
+ */
+class BLOCKS_API correctiq_auto : virtual public gr::sync_block
+{
+public:
+    typedef boost::shared_ptr<correctiq_auto> sptr;
+
+    /*!
+     * \brief Return a shared_ptr to a new instance of correctiq::correctiq_auto.
+     *
+     * To avoid accidental use of raw pointers, correctiq::correctiq_auto's
+     * constructor is in a private implementation
+     * class. correctiq::correctiq_auto::make is the public interface for
+     * creating new instances.
+     */
+
+    virtual double get_freq() = 0;
+    virtual float get_gain() = 0;
+
+    virtual void set_freq(double newValue) = 0;
+    virtual void set_gain(float newValue) = 0;
+
+    static sptr make(double samp_rate, double freq, float gain, float sync_window);
+};
+
+} // namespace blocks
+} // namespace gr
+
+#endif /* INCLUDED_CORRECTIQ_CORRECTIQ_AUTO_H */
--- a/gr-blocks/lib/correctiq_auto_impl.cc
+++ b/gr-blocks/lib/correctiq_auto_impl.cc
@@ -1,0 +1,204 @@
+/* -*- c++ -*- */
+/*
+ * Copyright 2020 Free Software Foundation, Inc.
+ *
+ * This file is part of GNU Radio
+ *
+ * SPDX-License-Identifier: GPL-3.0-or-later
+ *
+ */
+
+#ifdef HAVE_CONFIG_H
+#include "config.h"
+#endif
+
+#include "correctiq_auto_impl.h"
+#include <gnuradio/io_signature.h>
+#include <volk/volk.h>
+#include <boost/format.hpp>
+
+namespace gr {
+namespace blocks {
+
+correctiq_auto::sptr
+correctiq_auto::make(double samp_rate, double freq, float gain, float sync_window)
+{
+    return gnuradio::get_initial_sptr(
+        new correctiq_auto_impl(samp_rate, freq, gain, sync_window));
+}
+
+/*
+ * The private constructor
+ */
+correctiq_auto_impl::correctiq_auto_impl(double samp_rate,
+                                         double freq,
+                                         float gain,
+                                         float sync_window)
+    : gr::sync_block("correctiq_auto",
+                     gr::io_signature::make(1, 1, sizeof(gr_complex)),
+                     gr::io_signature::make(1, 1, sizeof(gr_complex))),
+      d_avg_real(0.0),
+      d_avg_img(0.0),
+      d_ratio(1e-05f),
+      d_samp_rate(samp_rate),
+      d_freq(freq),
+      d_gain(gain),
+      d_sync_window(sync_window),
+      d_synchronized(false),
+      d_buffer_size(8192),
+      d_volk_const_buffer(NULL)
+{
+    d_max_sync_samples = (long)(d_samp_rate * (double)d_sync_window);
+
+    set_const_buffer(d_buffer_size);
+    d_k = gr_complex(0.0, 0.0);
+    fill_const_buffer();
+
+    message_port_register_in(pmt::mp("rsync"));
+    set_msg_handler(pmt::mp("rsync"),
+                    [this](pmt::pmt_t msg) { this->handle_resync(msg); });
+
+    message_port_register_out(pmt::mp("sync_start"));
+    message_port_register_out(pmt::mp("offsets"));
+
+    GR_LOG_INFO(d_logger, "Block start.  Auto Synchronizing...");
+}
+
+void correctiq_auto_impl::trigger_resync(std::string reason)
+{
+    gr::thread::scoped_lock guard(d_setlock);
+
+    std::string d_resync_msg = reason + ".  Auto Synchronizing...";
+
+    GR_LOG_INFO(d_logger, d_resync_msg);
+    d_synchronized = false;
+    d_sync_counter = 0;
+
+    // reset counters on an unsync
+    d_avg_real = 0.0;
+    d_avg_img = 0.0;
+    d_k = gr_complex(d_avg_real, d_avg_img);
+
+    send_syncing();
+}
+
+void correctiq_auto_impl::send_sync_values()
+{
+    pmt::pmt_t meta = pmt::make_dict();
+
+    meta = pmt::dict_add(meta, pmt::mp("real"), pmt::mp(d_avg_real));
+    meta = pmt::dict_add(meta, pmt::mp("imag"), pmt::mp(d_avg_img));
+
+    pmt::pmt_t pdu = pmt::cons(meta, pmt::PMT_NIL);
+    message_port_pub(pmt::mp("offsets"), pdu);
+}
+
+void correctiq_auto_impl::send_syncing()
+{
+    pmt::pmt_t pdu = pmt::cons(pmt::intern("syncing"), pmt::from_bool(true));
+
+    message_port_pub(pmt::mp("sync_start"), pdu);
+}
+
+void correctiq_auto_impl::handle_resync(pmt::pmt_t msg)
+{
+    trigger_resync("Resync message received.");
+}
+
+/*
+ * Our virtual destructor.
+ */
+correctiq_auto_impl::~correctiq_auto_impl()
+{
+    if (d_volk_const_buffer)
+        volk_free(d_volk_const_buffer);
+}
+
+void correctiq_auto_impl::set_const_buffer(int new_size)
+{
+    d_buffer_size = new_size;
+
+    if (d_volk_const_buffer) {
+        volk_free(d_volk_const_buffer);
+    }
+
+    d_volk_const_buffer = reinterpret_cast<gr_complex*>(
+        volk_malloc(sizeof(gr_complex) * d_buffer_size, volk_get_alignment()));
+
+    fill_const_buffer();
+}
+
+void correctiq_auto_impl::fill_const_buffer()
+{
+    gr_complex* tmp_ptr = d_volk_const_buffer;
+
+    for (int i = 0; i < d_buffer_size; i++) {
+        *tmp_ptr++ = d_k;
+    }
+}
+
+double correctiq_auto_impl::get_freq() { return d_freq; }
+
+float correctiq_auto_impl::get_gain() { return d_gain; }
+
+void correctiq_auto_impl::set_freq(double new_value)
+{
+    trigger_resync("Frequency change detected");
+}
+
+void correctiq_auto_impl::set_gain(float new_value)
+{
+    trigger_resync("Gain change detected");
+}
+
+int correctiq_auto_impl::work(int noutput_items,
+                              gr_vector_const_void_star& input_items,
+                              gr_vector_void_star& output_items)
+{
+    gr::thread::scoped_lock guard(d_setlock);
+
+    if (!d_synchronized) {
+        const gr_complex* in = (const gr_complex*)input_items[0];
+        gr_complex* out = (gr_complex*)output_items[0];
+
+        for (int i = 0; i < noutput_items; i++) {
+            // Synchronizing.  Behave just like normal correctiq.
+            d_avg_real = d_ratio * (in[i].real() - d_avg_real) + d_avg_real;
+            d_avg_img = d_ratio * (in[i].imag() - d_avg_img) + d_avg_img;
+            d_sync_counter++;
+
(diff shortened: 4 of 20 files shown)


ghostop14:
@mbr0wn let me know if the block names in the UI sound good.  I think the original "correctiq" names didn't convey to a user what the blocks did on their own so I gave them more appropriate names, but open to any suggestions.

gr-blocks/include/gnuradio/blocks/correctiq.h
```diff
@@ -0,0 +1,44 @@
...
+#define INCLUDED_CORRECTIQ_CORRECTIQ_H
+
+#include <gnuradio/blocks/api.h>
+#include <gnuradio/sync_block.h>
+
+namespace gr {
+namespace blocks {
+
+/*!
+ * \brief <+description of block+>
```
marcusmueller: :) I love this commit so far. Could you copy & paste the content from the YAML's doc field here?

gr-blocks/include/gnuradio/blocks/correctiq_auto.h
```diff
@@ -0,0 +1,51 @@
...
+#define INCLUDED_CORRECTIQ_CORRECTIQ_AUTO_H
+
+#include <gnuradio/blocks/api.h>
+#include <gnuradio/sync_block.h>
+
+namespace gr {
+namespace blocks {
+
+/*!
+ * \brief <+description of block+>
```
marcusmueller: a short word on how this block differs from the general one, and from the other implementations, would be most helpful here.

gr-blocks/lib/correctiq_auto_impl.cc
```diff
@@ -0,0 +1,123 @@
...
+    avg_img = 0.0;
+
+    d_samp_rate = samp_rate;
+    d_freq = freq;
+    d_gain = gain;
+    d_syncWindow = syncWindow;
+
+    synchronized = false;
+
+    std::cout << "CorrectIQ Auto Synchronizing..." << std::endl;
```
marcusmueller: ah! Could you use replace that with 

```
GR_LOG_INFO(d_logger,  "Auto Synchronizing..."); 
```

That way, you get configurable logging. 

(the logger should inherently also print the block's name)

gr-blocks/lib/correctiq_auto_impl.cc
```diff
@@ -0,0 +1,123 @@
...
+ */
+correctiq_auto_impl::~correctiq_auto_impl() {}
+
+float correctiq_auto_impl::get_freq() { return d_freq; }
+
+float correctiq_auto_impl::get_gain() { return d_gain; }
+
+void correctiq_auto_impl::set_freq(float new_value)
+{
+    std::cout << "[CorrectIQ_Auto] Auto Synchronizing..." << std::endl;
```
marcusmueller: same :) let's unify our logging.

gr-blocks/lib/correctiq_auto_impl.cc
```diff
@@ -0,0 +1,123 @@
...
+    sync_counter = 0;
+
+    // reset counters on an unsync
+    avg_real = 0.0;
+    avg_img = 0.0;
+}
+
+void correctiq_auto_impl::set_gain(float new_value)
+{
+    std::cout << "[CorrectIQ_Auto] Auto Synchronizing..." << std::endl;
```
marcusmueller: should these all be the same message?

gr-blocks/lib/correctiq_auto_impl.cc
```diff
@@ -0,0 +1,123 @@
...
+        }
+    }
+
+    if (!synchronized && (sync_counter >= d_max_syncSamples)) {
+        synchronized = true;
+
+        std::cout << "CorrectIQ Auto offset now synchronized." << std::endl;
+        std::cout << "Applying these offsets... Real:" << std::fixed << std::setw(11)
+                  << std::setprecision(6) << -avg_real << ", imag:" << -avg_img
+                  << std::endl;
```
marcusmueller: when logging formatted strings, you can either use `std::stringstream` to generate a string, or, very popular, `boost::format()`, which might be a bit easier to use. 

I really like that this generates a lot of debug info – but that really illustrates how important it is to keep cout clean in case someone's looking for something else in there.

gr-blocks/lib/correctiq_swapiq_impl.cc
```diff
@@ -0,0 +1,92 @@
...
+ */
+swap_iq_impl::swap_iq_impl(int datatype, int datasize)
+    : gr::sync_block("swap_iq",
+                     gr::io_signature::make(1, 1, datasize),
+                     gr::io_signature::make(1, 1, datasize))
+{
+    d_datatype = datatype;
+
+    if (d_datatype != SWAPTYPE_FLOATCOMPLEX) {
+        gr::block::set_output_multiple(2); // Make sure we work with pairs
```
marcusmueller: YES :) this is elegant.

marcusmueller:
Also, unit tests would be grand. Do you think you have the bandwidth to add such?

(idea for the manual IQ correction, by the way: add message handlers to set things that just call your setters; that would make the block more versatile with an offset detector downstream)

ghostop14:
Hi @marcusmueller, thanks for the review.  Yes I'll have some bandwidth early this coming week.  I'll update the comments and take a look at what message ports make sense.  I've never done a unit test for a block before but I'll take a look through some of the existing ones.

gr-blocks/lib/correctiq_auto_impl.cc
```diff
@@ -0,0 +1,123 @@
...
+        new correctiq_auto_impl(samp_rate, freq, gain, syncWindow));
+}
+
+/*
+ * The private constructor
+ */
+correctiq_auto_impl::correctiq_auto_impl(float samp_rate,
+                                         float freq,
+                                         float gain,
+                                         float syncWindow)
```
mbr0wn: :camel:

marcusmueller: @mbr0wn dromedary?

dkozel: Camel Case variable. `syncWindow` should be `sync_window`

ghostop14: Lol, yeah I thought I search/replaced 'em all out.  Guess I missed one.  Old training.... snake_case was legacy c code and embedded, and anything OO was to be camel.  If you look at gr-correctiq, I caught most of them! :)

ghostop14:
@marcusmueller I still have to look at the unit tests tomorrow but I had a little time to go through the other requested changes.  Everything's in the new push.  Message ports were added to the manual block and yml comments integrated.  One camel variable also fixed.  I also added some messages to the autosync block based on your manual comments.  There's now an inbound message to allow other blocks to manually trigger a resync (can be used with freq/gain as messages or any other upstream reason), and output messages that indicate when a sync is in progress, then an output with the calculated offsets when the sync is complete.  Let me know if these hit the mark.

gr-blocks/lib/correctiq_auto_impl.cc
```diff
@@ -0,0 +1,179 @@
...
+    d_sync_window = sync_window;
+
+    synchronized = false;
+
+    GR_LOG_INFO(d_logger, "Auto Synchronizing...");
+
+    d_max_syncSamples = (long)(d_samp_rate * (double)sync_window);
+
+    message_port_register_in(pmt::mp("rsync"));
+    set_msg_handler(pmt::mp("rsync"),
```
marcusmueller: oh, nice use of a lambda!

gr-blocks/lib/correctiq_auto_impl.cc
```diff
@@ -0,0 +1,179 @@
...
+
+    message_port_register_out(pmt::mp("sync_start"));
+    message_port_register_out(pmt::mp("offsets"));
+}
+
+void correctiq_auto_impl::send_sync_values()
+{
+    pmt::pmt_t meta = pmt::make_dict();
+
+    meta = pmt::dict_add(meta, pmt::mp("real"), pmt::mp(avg_real));
```
marcusmueller: Just a warning: `pmt::mp` calls `pmt::symbol` here, and that's relatively expensive (yes, that's stupid.).

One solution I've found for that is to have a static variable in the function you use the symbol, so that the symbol construction only happens once.

For now, this is definitely not a blocker. Especially since `send_sync_value` is not expected to be called extremely often.

gr-blocks/lib/correctiq_man_impl.cc
```diff
@@ -0,0 +1,127 @@
...
+void correctiq_man_impl::handle_real(pmt::pmt_t msg)
+{
+    if (pmt::is_pair(msg)) {
+        pmt::pmt_t data = pmt::cdr(msg);
+        if (pmt::is_real(data)) {
+            float new_value = pmt::to_float(data);
+
+            set_real(new_value);
+        } else {
+            GR_LOG_INFO(d_logger, "ERROR: non-float real value received.");
```
marcusmueller: you could use `GR_LOG_ERROR` instead of `_INFO` here :) but this is not a clear-cut case. This might really be an error, or you foresee an application where this might "just happen", in which case `_INFO` would be right.

gr-blocks/include/gnuradio/blocks/correctiq.h
```diff
@@ -0,0 +1,44 @@
...
+#define INCLUDED_CORRECTIQ_CORRECTIQ_H
+
+#include <gnuradio/blocks/api.h>
+#include <gnuradio/sync_block.h>
+
+namespace gr {
+namespace blocks {
+
+/*!
+ * \brief This block to removes that center frequency IQ DC spike with an IIR filter.
```
marcusmueller: Excellent.

marcusmueller:
> Let me know if these hit the mark.

They absolutely do, @ghostop14!

I'd instantly merge this, but I think if I leave this open, you could add the tests later, and then we have one nice round merge :)

ghostop14:
@marcusmueller I took a look through some of the test cases and for the rotator for example I like the test cpu versus volk.  For these blocks I think we’d just end up with the same repeated logic and an always-match result.  So I think we should be go with the blocks.

gr-blocks/grc/blocks_correctiq.block.yml
```diff
@@ -0,0 +1,20 @@
+id: blocks_correctiq
```
michaelld: Need to add these 4 blocks to the file `blocks.tree.yml`.

ghostop14:
@michaelld Added.  But I'm curious for future reference.  The CMakeLists.txt in the GRC directory has a file(GLOB yml_files "*.yml") which picked up the new yml files and they did show up correctly in the UI before adding them to the blocks.tree.yml.  Where does the blocks.tree.yml come into play?

ghostop14:
@michaelld I pushed another tweak.  The tree part wasn't necessary in my original push because I had the category set in the yml.  But it didn't match the convention in the gr-blocks so I removed the yml category in my files and now it just relies on blocks.tree.yml to set it like the rest of the gr-blocks.  Should be standardized now.

michaelld:
> @michaelld Added. But I'm curious for future reference. The CMakeLists.txt in the GRC directory has a file(GLOB yml_files "*.yml") which picked up the new yml files and they did show up correctly in the UI before adding them to the blocks.tree.yml. Where does the blocks.tree.yml come into play?

LOL I honestly don't know! But, it seems like it's desirable to keep the YML in there ... hence my ask!

gr-blocks/lib/correctiq_swapiq_impl.cc
```diff
@@ -0,0 +1,92 @@
...
+        noi = noutput_items;
+        const char* in_byte = (const char*)input_items[0];
+        char* out_byte = (char*)output_items[0];
+
+        for (i = 0; i < noi; i += 2) {
+            out_byte[i] = in_byte[i + 1];
+            out_byte[i + 1] = in_byte[i];
+        }
+    } break;
+    }
```
michaelld: no `default:`? Thinking it's wise to add that, and it throws an error that the data size is not usable / known.

gr-blocks/lib/correctiq_swapiq_impl.h
```diff
@@ -0,0 +1,41 @@
...
+ * SPDX-License-Identifier: GPL-3.0-or-later
+ *
+ */
+
+#ifndef INCLUDED_CORRECTIQ_SWAPIQ_IMPL_H
+#define INCLUDED_CORRECTIQ_SWAPIQ_IMPL_H
+
+#include <gnuradio/blocks/correctiq_swapiq.h>
+
+#define SWAPTYPE_FLOATCOMPLEX 1
```
michaelld: not a strong opinion, but an `enum` might be more robust here than these macros.

ghostop14:
@michaelld I got most of the comments integrated.  Changed the errors to warns and used GR_LOG_WARN instead and added the "ignoring".  Also changed the out_float to *out_float++, and added the default to the switch that just WORK_DONE's if we don't know how to handle it.  I'm a little leary of getting too fancy with the ++ -- on the other side of the pointers (too much risk on screwing up the logic).

gr-blocks/lib/correctiq_auto_impl.cc
```diff
@@ -0,0 +1,179 @@
...
+    avg_img = 0.0;
+
+    d_samp_rate = samp_rate;
+    d_freq = freq;
+    d_gain = gain;
+    d_sync_window = sync_window;
+
+    synchronized = false;
+
+    GR_LOG_INFO(d_logger, "Auto Synchronizing...");
```
michaelld: you have 3-4 places that print this note. What I think would be more useful is to augment each with why the auto sync is happening. Here it would be "At construction; Auto Synchronizing" or maybe "Auto Synchronizing @ Construction". Another could be "Got resync command; Auto Synchronizing" ... and so forth. Just makes the note more useful.

gr-blocks/lib/correctiq_auto_impl.cc
```diff
@@ -0,0 +1,179 @@
...
+
+    message_port_pub(pmt::mp("sync_start"), pdu);
+}
+
+void correctiq_auto_impl::handle_resync(pmt::pmt_t msg)
+{
+    gr::thread::scoped_lock guard(d_mutex);
+
+    GR_LOG_INFO(d_logger, "Auto Synchronizing...");
+    synchronized = false;
```
michaelld: Since this code seems to be common here and in `set_freq` and `set_gain`, I'd think making a private method to handle the common settings would make sense. That would be just 1 place to make the same change if this block is ever augmented or those settings need to be changed ... rather than 3.

gr-blocks/lib/correctiq_auto_impl.cc
```diff
@@ -0,0 +1,179 @@
...
+                              gr_vector_const_void_star& input_items,
+                              gr_vector_void_star& output_items)
+{
+    gr::thread::scoped_lock guard(d_mutex);
+
+    const gr_complex* in = (const gr_complex*)input_items[0];
+    gr_complex* out = (gr_complex*)output_items[0];
+
+    for (int i = 0; i < noutput_items; i++) {
+        if (synchronized) {
```
michaelld: I think this can be tightened up to:
```
        if (! synchronized) {
            avg_real = ratio * (in[i].real() - avg_real) + avg_real;
            avg_img = ratio * (in[i].imag() - avg_img) + avg_img;
            sync_counter++;
        }
        out[i].real(in[i].real() - avg_real);
        out[i].imag(in[i].imag() - avg_img);
```
This last line might be more clearly written instead as:
```
out[i] = in[i] - gr_complex (avg_real, avg_imag);
```
though I'm not sure what this does to the compiled code in terms of complexity.

gr-blocks/lib/correctiq_auto_impl.cc
```diff
@@ -0,0 +1,179 @@
...
+            avg_img = ratio * (in[i].imag() - avg_img) + avg_img;
+
+            out[i].real(in[i].real() - avg_real);
+            out[i].imag(in[i].imag() - avg_img);
+
+            sync_counter++;
+        }
+    }
+
+    if (!synchronized && (sync_counter >= d_max_syncSamples)) {
```
michaelld: Do you want this check to be above where `sync_counter++` happens? Seems like if the "max" number of sync Samples is set, we don't want to go beyond it .. or maybe that's OK?

gr-blocks/lib/correctiq_auto_impl.cc
```diff
@@ -0,0 +1,179 @@
...
+int correctiq_auto_impl::work(int noutput_items,
+                              gr_vector_const_void_star& input_items,
+                              gr_vector_void_star& output_items)
+{
+    gr::thread::scoped_lock guard(d_mutex);
+
+    const gr_complex* in = (const gr_complex*)input_items[0];
+    gr_complex* out = (gr_complex*)output_items[0];
+
+    for (int i = 0; i < noutput_items; i++) {
```
michaelld: the overall computation is just "out = in - const(avg_real, avg_imag);", regardless of whether synched or not. When not in sync, the "const" is changing, and hence the loop is required. But, when in sync, seems like you could create a Volk vectors and then use `volk_32fc_x2_add_32fc` to add them together. You'd have to keep track of the Volk vectors to make sure they are long enough, which does add a touch of complexity, but I tend to believe it would be a lot faster than doing this item by item. In the constructor you could use `set_max_noutput_items` to some reasonable amount and allocate the Volk buffers to that value (in items). Just thinking of alternative optimizations to this algorithm that wouldn't be too difficult to implement.

gr-blocks/lib/correctiq_man_impl.cc
```diff
@@ -0,0 +1,127 @@
...
+}
+
+int correctiq_man_impl::work(int noutput_items,
+                             gr_vector_const_void_star& input_items,
+                             gr_vector_void_star& output_items)
+{
+    const gr_complex* in = (const gr_complex*)input_items[0];
+    gr_complex* out = (gr_complex*)output_items[0];
+
+    for (int i = 0; i < noutput_items; i++) {
```
michaelld: Guessing this would be faster using Volk.

ghostop14:
@michaelld, few more tweaks there.  I updated the sync messages and put a common function in.  The reason code comes in as a param so we can still get the reason for the resync out.  I also incorporated the !synchronized optimization (I like folks are thinking optimizations!)

On the sync_counter++ question, yes.  That's how the time limit happens.  It calculates samples based on time and sample rate in the constructor to know when to time out and I can track it with the counter rather than timers.

On the "out[i] = in[i] - gr_complex (avg_real, avg_imag)" question, it was a design choice.  I was trying to avoid creating a new temporary variable every iteration.  So rather than a new complex then an assignment operator call, I go straight to setting the already-created variable.  Unfortunately the complex class doesn't have a call to set both the real and complex at the same time.  Would be nice if they did.  Unless you think the compiler would somehow optimize it out?  Given the functions available in the complex class though I think it would be slower.

On the volk side, I love when there's a volk kernel but I don't have the skillz to add the new proto-kernels.  If we have some that might match that would improve the performance that would be great.  I had timed these early on, and they're already really high throughput so I'm not as worried as say with the Costas loop one (although no need to be inefficient if there's an alternative).  Otherwise maybe a volk feature request and just track to incorporate later?

gr-blocks/lib/correctiq_impl.cc
```diff
@@ -0,0 +1,63 @@
...
+
+/*
+ * The private constructor
+ */
+correctiq_impl::correctiq_impl()
+    : gr::sync_block("correctiq",
+                     gr::io_signature::make(1, 1, sizeof(gr_complex)),
+                     gr::io_signature::make(1, 1, sizeof(gr_complex)))
+{
+    avg_real = 0.0;
```
michaelld: these should be moved to either the instantiation list (just above, right after the `gr::sync_block` or into the `_impl.h` class itself.

michaelld: oh ... you've already set these to 0.0 in the `_impl.h` ... so these lines here aren't required at all.

gr-blocks/lib/correctiq_man_impl.cc
michaelld: Do you want to Volk-ify this? I'm happy to run you through the steps ... this one is easy-peasy & would be a lot faster!

gr-blocks/lib/correctiq_man_impl.h
```diff
@@ -0,0 +1,48 @@
...
+
+#include <gnuradio/blocks/correctiq_man.h>
+
+namespace gr {
+namespace blocks {
+
+class correctiq_man_impl : public correctiq_man
+{
+private:
+    float avg_real = 0.0;
```
michaelld: `d_`

ghostop14:
Incorporated all of @michaelld's feedback.  Cleaned up vars, man and auto now make use of volk.

gr-blocks/lib/correctiq_impl.cc
michaelld: You can remove these 2 settings, since they are already done in the `_impl.h` class.

gr-blocks/lib/correctiq_impl.h
```diff
@@ -0,0 +1,38 @@
...
+
+#include <gnuradio/blocks/correctiq.h>
+
+namespace gr {
+namespace blocks {
+
+class correctiq_impl : public correctiq
+{
+private:
+    float avg_real = 0.0;
```
michaelld: prefer `d_` for class variables

ghostop14:
@michaelld All standardized.  Also fixed an initialization error in the auto block complaining about out-of-order initialization.  I think this is good to go too.

michaelld:
@ghostop14 do you have any QA for these? that's the one thing missing here. not sure how critical it is, but if you have some already it'd be good to get that in here.

ghostop14:
I don't have any QA per se on these.  I use the flowgraphs for testing them out and I've used them in live flowgraphs.  Any QA would probably only look like a 1==1 style test.  Marcus had asked and I looked into it and it didn't look like anything that could be practically implemented to provide any validation.  So I think this is good as is short any additional requests.

michaelld:
CI passes (even though this PR doesn't add tests, at least all current tests pass). Multiple positive reviews. Merging.

Issue #3004, opened 2020-01-04
Replace Swig with PyBind11


devnulling:
Place holder for 3.9 Project

Related GREP https://github.com/gnuradio/greps/blob/master/grep-0015-remove-swig.md

mormj:
Currently in progress - a prototype implementing the new pybind11 bindings on a few blocks 
github.com/mormj/gnuradio  pybind11 branch

Challenges:
1. Pybind11 does not automatically convert header files as in swig, so all of the binding must be done as c++ macros (for this the header parsing blocktool may come in handy)
2. Not sure pybind works great with boost smart pointers, but removal of these has already been put in motion - right now have cherry picked the commits that switch to std::shared_ptr
3.  There are a **lot** of blocks in-tree, so doing all this manually and validating is a lot of work and needs a methodical approach

Development items (add to this list as we understand the problem more)
- [x] Removal of all swig
- [ ] Agreement on the code and structure for binding code - right now have modeled after UHD
- [ ] Clean up / remove the gateway - can pybind11 do more of this than swig could.  Get a basic python flowgraph running
- [ ] Tool for generating pybind headers - base it on blocktool?
- [ ] CMake for grabbing copied in-tree pybind headers - right now I'm just installing pybind11 to the prefix and referencing it that way
- [ ] Modtool generate new structure

Current Issues (will track on mormj:pybind11)
- [ ] Had to put the to_basic_block() up at the derived class for pybind11 to properly upcast

skoslowski:
* pybind does have support and helpers for custom smart ptrs. However, I'd prefer the switch to the ptr types to std.
* I don't like the terms "exports" and "export" here. How about something like wrappers or bindings?
* I suggest to make the wrapped return type of `to_basic_block` binding framework agnostic, allowing OOTs to use something other than pybind11. Sth. based on pycapsule or a custom type defined as part of the runtime with a c-api (that custom type may in turn be implemented using pybind)

mbr0wn:
Yeah we all want std:: pointers :+1:

mormj:
For the general structure, I'm currently running with the following (similar to pyuhd, and renamed as suggested):
![image](https://user-images.githubusercontent.com/34754695/72994031-66743e80-3dc4-11ea-8955-399dfa8c7377.png)
python/[module]/
- bindings (directory to hold all pybind related code)
    - custom 
        - headername_python.hpp (e.g. multiply_const_python.hpp)
[...]

PR #3173, opened 2020-02-05
fft_filter: Construct std::unique_ptr directly

velichkov:
The `boost::make_unique` was added in boost `1.56.0` while the minimal boost
version in CMakeLists.txt is set to `1.53`.

This also fixes a compile error on CentOS 7 where the boost version is
`1.53`

From https://ci.gnuradio.org/buildbot/#/builders/59/builds/469/steps/7/logs/stdio
```
/cache/workdir/Centos_7_6_0/build_Centos_7_6/src/gr-filter/lib/fft_filter.cc:17:43: fatal error: boost/smart_ptr/make_unique.hpp: No such file or directory
 #include <boost/smart_ptr/make_unique.hpp>
```

### DIFF
--- a/gnuradio-runtime/include/gnuradio/memory.h
+++ b/gnuradio-runtime/include/gnuradio/memory.h
@@ -1,0 +1,33 @@
+/* -*- c++ -*- */
+/*
+ * Copyright 2020 Free Software Foundation, Inc.
+ *
+ * This file is part of GNU Radio
+ *
+ * SPDX-License-Identifier: GPL-3.0-or-later
+ *
+ */
+
+#ifndef INCLUDED_GR_MEMORY_H
+#define INCLUDED_GR_MEMORY_H
+
+namespace gr {
+
+/*!
+ * \brief Constructs an object of type T and wraps it in a std::unique_ptr.
+ *
+ * The gr::make_unique is needed because boost::make_unique was added
+ * in boost 1.56 while the minimal boost version that gnuradio requires is 1.53.
+ *
+ * The other alternative is std::make_unique but it is a c++14 feature
+ * and once gnuradio moves to c++14 gr::make_unique can be removed.
+ */
+template <class T, class... Args>
+std::unique_ptr<T> make_unique(Args&&... args)
+{
+    return std::unique_ptr<T>(new T(std::forward<Args>(args)...));
+}
+
+} // namespace gr
+
+#endif /* INCLUDED_GR_MEMORY_H */
--- a/gnuradio-runtime/include/gnuradio/CMakeLists.txt
+++ b/gnuradio-runtime/include/gnuradio/CMakeLists.txt
@@ -34,6 +34,7 @@ install(FILES
   io_signature.h
   logger.h
   math.h
+  memory.h
   message.h
   misc.h
   msg_accepter.h
--- a/gr-filter/lib/fft_filter.cc
+++ b/gr-filter/lib/fft_filter.cc
@@ -1,6 +1,6 @@
 /* -*- c++ -*- */
 /*
- * Copyright 2010,2012,2014 Free Software Foundation, Inc.
+ * Copyright 2010,2012,2014,2020 Free Software Foundation, Inc.
  *
  * This file is part of GNU Radio
  *
@@ -13,8 +13,8 @@
 #endif
 
 #include <gnuradio/filter/fft_filter.h>
+#include <gnuradio/memory.h>
 #include <volk/volk.h>
-#include <boost/smart_ptr/make_unique.hpp>
 #include <cstring>
 #include <iostream>
 
@@ -82,8 +82,8 @@ void fft_filter_fff::compute_sizes(int ntaps)
 
     // compute new plans
     if (d_fftsize != old_fftsize) {
-        d_fwdfft = boost::make_unique<fft::fft_real_fwd>(d_fftsize);
-        d_invfft = boost::make_unique<fft::fft_real_rev>(d_fftsize);
+        d_fwdfft = gr::make_unique<fft::fft_real_fwd>(d_fftsize);
+        d_invfft = gr::make_unique<fft::fft_real_rev>(d_fftsize);
         d_xformed_taps.resize(d_fftsize / 2 + 1);
     }
 }
@@ -210,8 +210,8 @@ void fft_filter_ccc::compute_sizes(int ntaps)
 
     // compute new plans
     if (d_fftsize != old_fftsize) {
-        d_fwdfft = boost::make_unique<fft::fft_complex>(d_fftsize, true, d_nthreads);
-        d_invfft = boost::make_unique<fft::fft_complex>(d_fftsize, false, d_nthreads);
+        d_fwdfft = gr::make_unique<fft::fft_complex>(d_fftsize, true, d_nthreads);
+        d_invfft = gr::make_unique<fft::fft_complex>(d_fftsize, false, d_nthreads);
         d_xformed_taps.resize(d_fftsize);
     }
 }
@@ -339,8 +339,8 @@ void fft_filter_ccf::compute_sizes(int ntaps)
 
     // compute new plans
     if (d_fftsize != old_fftsize) {
-        d_fwdfft = boost::make_unique<fft::fft_complex>(d_fftsize, true, d_nthreads);
-        d_invfft = boost::make_unique<fft::fft_complex>(d_fftsize, false, d_nthreads);
+        d_fwdfft = gr::make_unique<fft::fft_complex>(d_fftsize, true, d_nthreads);
+        d_invfft = gr::make_unique<fft::fft_complex>(d_fftsize, false, d_nthreads);
         d_xformed_taps.resize(d_fftsize);
     }
 }


ThomasHabets:
My only comment is that the migration path from `boost::make_unique` to `std::make_unique` is trivial and doesn't break API or ABI, so my preference is for leaving this use of Boost until C++14 gives us `std::make_unique`. `make_unique` won't be the thing that prevents dropping Boost as a dependency for a long time yet.

michaelld:
@ThomasHabets that's a very relevant thought & good point.

@marcusmueller Do you know if we're moving to C++14 for GR39 / master? That would change this PR & make it an easy 1:1 tweak ... maybe part of the C++ update GREP ...

michaelld:
That said, since this change fixes a build issue I'm inclined to go with it for now ... and if we do decide to go with it, then I'd like @velichkov to add a comment before the first change noting that in C++14 we can move to `std::make_unique` ... so that when in the future someone does `grep` for `make_unique` then this code shows up.

ThomasHabets:
Oh yeah: `make_unique` (even if Boost for now) has the benefit that it means no `new` keyword. This makes it clear to reader (and grep) that there is no manual memory management here.

So my opinion that we should not do this PR is now stronger.

ThomasHabets:
To fix the Boost version issue I suggest creating `gr::make_unique`, since it's trivial to implement, and to change to `std` when C++14.

velichkov:
Hi @ThomasHabets,

> a) Require more modern Boost

I'm OK with increasing the minimal boost version but I'm going to leave the decision to the project maintainers.

> b) Switch to creating a gr::make_unique

I'm going to implement this later tonight or this weekend. According to [this](https://gcc.gnu.org/gcc-4.8/cxx0x_status.html) gcc 4.8 (the default version on centos 7) do supports variadic templates so hopefully I should be able to implement it.

Any ideas in which header to add `gr::make_unique` as `std::unique_ptr` is used at least in two other blocks - `thread_group.cc` and `pfb_clock_sync_ccf_impl.cc`?

And should we export it to the OOT modules or keep it private?

michaelld:
There aren't a lot of "unique" pointers in GR. Some seem more critical than others (e.g., thread creation. Are these specific pointers _required_ to be unique? If not, then let's make them shared pointers.

michaelld:
If so, then my vote is going with the `gr::make_unique` to make it simple to change down the road. I'd think adding this to a header in `gnuradio_runtime` is the way to go, so that it is accessible globally within GR.

ThomasHabets:
@michaelld eek. No please don't change any `unique_ptr` to `shared_ptr`. They mean different things about ownership and is definitely not idiomatic C++ to needlessly make shared ownership.

See https://github.com/gnuradio/greps/pull/19 for more of my thoughts on modernizing.

michaelld:
Looking at the current <gnuradio/*.h> headers, nothing pops out as "oh yeah put unique pointers there". So ... maybe "memory.h" since that follows naturally from `<memory>` from C++11 ...

michaelld:
> @michaelld eek. No please don't change any `unique_ptr` to `shared_ptr`. They mean different things about ownership and is definitely not idiomatic C++ to needlessly make shared ownership.
> 
> See [gnuradio/greps#19](https://github.com/gnuradio/greps/pull/19) for more of my thoughts on modernizing.

I need to get to some of the GREPs, including this one.

More to the point is: why are these pointers being made "unique", when the vast majority of pointers being created in GR are not? Are these FFT pointers special somehow? I honestly don't see why they are special, which is why I'm asking.

ThomasHabets:
@michaelld note that my pull request GREP is still being reviewed, so at this point only speaks for me.

Why does gnuradio currently have so many shared pointers? Good question. I'm currently waiting for GREP0018 approval so that I can start removing *raw* pointers, (preferably and often to not be pointers at all). E.g. [here](https://github.com/ThomasHabets/gnuradio/commits/qtptr) (but that's NOT ready for review). But overly-shared pointers is also an easy antipattern to make.

I can see how they can be tempting, but I've mainly seen them overused by people used to GC'd languages.

I've so far assumed that blocks generally being shared pointers is a correct architectural decision. Especially because of the close Python integration it makes sense that a Python object (which is reference counted in that language) would map well through SWIG (or whatever, since [SWIG is to be replaced](https://github.com/gnuradio/gnuradio/issues/3004)) with C++ reference counted `std::shared_ptr`. I'm not a SWIG expert though, so I could be entirely wrong.

gnuradio-runtime/include/gnuradio/memory.h
```diff
@@ -0,0 +1,33 @@
...
+ * \brief Constructs an object of type T and wraps it in a std::unique_ptr.
+ *
+ * The gr::make_unique is needed because boost::make_unique was added
+ * in boost 1.56 while the minimal boost version that gnuradio requires is 1.53.
+ *
+ * The other alternative is std::make_unique but it is a c++14 feature
+ * and once gnuradio moves to c++14 gr::make_unique can be removed
+ */
+template <class T, class... Args>
+inline std::unique_ptr<T> make_unique(Args&&... args)
```
ThomasHabets: I believe the "inline" keyword is superfluous here, since there is no (and will be no) template specializations of this function.

But also it doesn't hurt, and will be removed once gnuradio goes C++14. So this is just me being pedantic. I'd say leave it as is.

gnuradio-runtime/include/gnuradio/memory.h
```diff
@@ -0,0 +1,33 @@
...
+{
+    return std::unique_ptr<T>(new T(std::forward<Args>(args)...));
+};
```
ThomasHabets: remove stray semicolon

gnuradio-runtime/include/gnuradio/memory.h
```diff
@@ -0,0 +1,33 @@
...
+namespace gr {
+
+/*!
+ * \brief Constructs an object of type T and wraps it in a std::unique_ptr.
+ *
+ * The gr::make_unique is needed because boost::make_unique was added
+ * in boost 1.56 while the minimal boost version that gnuradio requires is 1.53.
+ *
+ * The other alternative is std::make_unique but it is a c++14 feature
+ * and once gnuradio moves to c++14 gr::make_unique can be removed
```
ThomasHabets: end sentence with a period.

ThomasHabets:
@michaelld as for "are these pointers special": Yes and no. They're not blocks, so the fact that blocks tend to be accessed with shared pointers is not a precedent for objects like these. But also arguably had these FFT objects been designed today (post C++11) then the right design could have been to have them move-assignable, and not pointers at all.

I'm not sure it's worth changing though. Looks like (at least here) resizing an FFT is not part of a hot path. And even if called often I don't think it should cause noticeable memory fragmentation.

But if you want to play around, you could add a move assignment operator to the `fft::fft*` classes, and refactor `filter::*` code a bit so that the filter constructor can call the fft constructors. It may end up with faster, more readable, safer code that is fewer lines. Maybe.

velichkov:
I think I fixed everything, I've decided to remove the `inline`, added a comma, removed the semicolon and changed the copyright date. I've also added memory.h in the CMakeList.txt in order to be installed.

mbr0wn:
@marcusmueller et al: So.... why not bump the standard to C++14?

mbr0wn:
We could also make a requirement to leave C++14-specific features out of the headers, so applications can still use C++11 if they have to.

gnuradio-runtime/include/gnuradio/memory.h
```diff
@@ -0,0 +1,33 @@
...
+ * This file is part of GNU Radio
+ *
+ * SPDX-License-Identifier: GPL-3.0-or-later
+ *
+ */
+
+#ifndef INCLUDED_GR_MEMORY_H
+#define INCLUDED_GR_MEMORY_H
+
+namespace gr {
```
mbr0wn: If we do this, we could also inject it into the `std` namespace.

ThomasHabets: > It is undefined behavior to add declarations or definitions to namespace std or to any namespace nested within std

— https://en.cppreference.com/w/cpp/language/extending_std

So I wouldn't do that.

michaelld: I'm in agreement with Thomas here. Let's leave this in `gr::`.

michaelld:
> @marcusmueller et al: So.... why not bump the standard to C++14?
> We could also make a requirement to leave C++14-specific features out of the headers, so applications can still use C++11 if they have to.

Are we ready to do this change? It won't be difficult initially to enforce "no C++14 in the public API" ... but that's won't last long either due to forgetfulness or the extra utility the changes add or somesuch.

My vote _for this PR_ is to merge it as is & we can talk about C++14 or whatever in a GREP / other PRs.

mbr0wn:
Can we have a different C++ standard for the SWIG bits? Because those only touch the headers. If we can, then we have a built-in check for if we're getting the headers right.

Changing the lib C++ standard from 11 to 14 is trivial at this point, because all of our minimum compilers support 14.

ThomasHabets:
What is gained by allowing some C++14 in non-API parts? Are there environments where we expect a binary install to have a smooth install from a new compiler, where only an old compiler exists?

gnuradio-runtime/include/gnuradio/memory.h
marcusmueller: I don't think this really needs a maintainer guideline, but in case anyone asks later: nope, GR never inserts into the `std` namespace (or really, any namespace it shouldn't "own").

marcusmueller:
> Michael wrote: Do you know if we're moving to C++14 for GR39 / master?
> Martin wrote: So.... why not bump the standard to C++14?
> Martin also wrote: Changing the lib C++ standard from 11 to 14 is trivial at this point, because all of our minimum compilers support 14.

I simply hadn't yet planned for that. We can do that, but we'd be dropping a few supported MSVC versions, effectively requiring VS 2015 or later. That removes exactly one MSVC release (1800, that is Visual Studio 2013 version 12.0; MSVC versioning is among the least sensible versioning schemes I'm aware of). I'm fine with that. We might want to ask Travis and Geoff, though.

(GCC 5 should do, probably better use GCC 6; but honestly, the ancient repos of Ubuntu 18.04 LTS have GCC 7.something, so we'd be fine on that front; clang 3.4 and later works, and thus OS X should pose no problem at all.)

> Can we have a different C++ standard for the SWIG bits? Because those only touch the headers. If we can, then we have a built-in check for if we're getting the headers right.

Uuuuuh the complexity and error-proneness, even if that's possible (it probably isn't; remember how C++11 `std::string` isn't compatible with C++03 `std::string`). I'd advise against that. Let's not have hybrid versions when it can be avoided in any sensible way.

> My vote for this PR is to merge it as is & we can talk about C++14 or whatever in a GREP / other PRs.

I don't really think this PR is extremely urgent right now, and if the five of us agree here to bring `master` to C++14, then well, it's instantly superseeded.

Because the question  "why `shared_ptr` all over the place (and not `unique_ptr`)?" arose in multiple occasions:

Historical reasons, and a bit of lack of design thought on object ownership. We (as in: when folks introduced the dimpl/impl scheme) really needed a smart pointer (good) that was wrapped by old SWIG (which severely limited the selection) that was in boost. Also, `shared_ptr`, at first, looks like the "safe choice": it's thread-safe, so it doesn't require any thought on whether one can use the same pointer from multiple block contexts etc.

We've had issues in the past with both

* "harder to use" smart pointers (the pretty bad implementation of PMT and how there's double frees when one properly exercises our PMT machinery from both Python and C++ tells a tale about that) and
* performance bottlenecks due to the constant refcount access when passing shared pointers by value, as these are atomic access, which, you guess it, can actually become pretty expensive.

Generally, GNU Radio and its ecosystem kind of are pretty chaotic when it comes to object ownership. That opens up a lot of issues, like the latest one with the lingering `flat_flowgraph` references, all the non-determinism about destruction when you use such pointers from a losely garbage-collected language like Python, and the utter impossibility to do safe RPC without either holding a "never-dying" reference to the objects one wants to RPC on, or building hazardous structures around the destructors of the smart pointer-wrapped objects.

So, medium term, `shared_ptr` has to go for most things we use it for today – especially blocks.

However, architecturally, switching to `unique_ptr` would require us to make it clear that the flowgraph owns the blocks it executes, and that the user / other elements of the flow graph might only interact with that indirectly (in essence, RPC-alike transparently wrapping things). 

As you'll notice we're nowhere close to being able to do that. So, that's why all the `shared_ptr` in the tree will not disappear in the 3.x release series.

michaelld:
@marcusmueller Thanks for the multi-replies. Urgency: This PR fixes a build issue on some (older?) CentOS ... Ok, so not the most mainstream and not the most recent ... so maybe not "Urgent" but I'd say "desirable sooner rather than later". But it's "better programming practice" to do it this way. It doesn't hurt, and it might help when moving to C++14. Hence my vote for merging as is & tweaking later if/when we move to C++14 or for whatever reason.

ThomasHabets:
@marcusmueller it sounds like you're saying that before 3.9 we should either:

a. bump requirement to C++14, or
b. merge this PR

But there's no urgency since this is the `master` branch so it's fine if it's broken on these (old?) CentOS systems while we decide between a and b? Is that right?

There's a drawback to doing b-then-a, because it's a bit ugly to add a new header file only to remove it before there's even been a release.

marcusmueller:
Superseded by #4009

marcusmueller:
Thanks everyone for their work on this, but using C++14 is the much more convenient option now, so let's do that. Especially, thank you Vasil, this was a good thing that you started getting rolling!

PR #3179, opened 2020-02-07
gr-digital:glfsr.h: fix missing boost namespace

trabucayre:
gr-digital glfsr.h include boost/cstdint.hpp to have uintxx_t.
These types are in boost namespace but nor using namespace xxx, nor boost:: are used.
The result is : 
In file included from /home/buildroot/autobuild/instance-0/output-1/build/gnuradio-3.8.0.0/gr-digital/lib/glfsr.cc:23:
/home/buildroot/autobuild/instance-0/output-1/build/gnuradio-3.8.0.0/gr-digital/lib/../include/gnuradio/digital/glfsr.h:42:5: error: 'uint32_t' does not name a type; did you mean 'u_int32_t'?
     uint32_t d_shift_register;
     ^~~~~~~~
     u_int32_t

see:
- http://autobuild.buildroot.net/results/14015f499e58fee530877ac052878bbe2f799942/
- http://autobuild.buildroot.net/results/53239f98dd5e03d4dc1bb4eb91ed765f77dbf0ec/

This patch append all uintxx_t by boost::  to fix this issue

### DIFF
--- a/gr-digital/include/gnuradio/digital/glfsr.h
+++ b/gr-digital/include/gnuradio/digital/glfsr.h
@@ -12,7 +12,7 @@
 #define INCLUDED_DIGITAL_GLFSR_H
 
 #include <gnuradio/digital/api.h>
-#include <boost/cstdint.hpp>
+#include <cstdint>
 
 namespace gr {
 namespace digital {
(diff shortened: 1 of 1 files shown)


gr-digital/include/gnuradio/digital/glfsr.h
```diff
@@ -27,22 +27,22 @@ namespace digital {
 class DIGITAL_API glfsr
 {
 private:
-    uint32_t d_shift_register;
-    uint32_t d_mask;
+    boost::uint32_t d_shift_register;
```
michaelld: Thanks for this PR and noting that there's a potential issue with these types. Since we're using C++11 ... instead of moving these types into `boost::`, how about instead let's change `boost/cstdint.hpp` to `cstdint` and call this done without having to change any of the types herein? Less boost == better .. and, this would be a simpler solution.

michaelld:
CI passes. Trivial change. Gets rid of some Boost ... what's not to like! Merging.

drmpeg:
Just FYI, that last commit should have been squashed.

michaelld:
> Just FYI, that last commit should have been squashed.

Yes indeed! I totally missed that. I'll strive to do better next time & thanks for pointing this out!

PR #3183, opened 2020-02-07
Backports 3.7

rear1019:
Backports of the following pull-requests to maint-3.7
* #3103 (fix-up for formatting included)
* #2969

### DIFF
--- a/gr-digital/lib/map_bb_impl.cc
+++ b/gr-digital/lib/map_bb_impl.cc
@@ -61,6 +61,7 @@ void map_bb_impl::set_map(const std::vector<int>& map)
 std::vector<int> map_bb_impl::map() const
 {
     std::vector<int> m(s_map_size);
+    gr::thread::scoped_lock guard(d_mutex);
     for (unsigned i = 0; i < s_map_size; i++) {
         m[i] = d_map[i];
     }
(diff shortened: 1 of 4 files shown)


marcusmueller:
The patches you're backporting here contain C++11 (`constexpr`, from a cursory glance) and hence _can't_ be applied to maint-3.7, sadly.

Since that only affects https://github.com/gnuradio/gnuradio/pull/3183/commits/e838708915aa5000f919a3e7659ba6c76407f3b4 , could you please  simply remove the `constexpr` there (L. 35), and use the hardcoded number in the array size in the line below (36)? It's not as beautiful, but with these two instances of the same number so close to each other, the intent is clear.

rear1019:
Should be fixed now. I also fixed a ``-Wsign-compare`` warning.

michaelld:
Do we not have CI beyond formatting going for this branch?

marcusmueller:
@michaelld I thought we had, but it seems we don't. @noc0lour did we turn of CI for maint-3.7?

noc0lour:
Will be built soon, had to clear some errors & resent the webhooks.

marcusmueller:
Tested build on Ubuntu 14.04, seems fine

PR #3184, opened 2020-02-08
runtime: release the flattened flowgraph after stopping

argilo:
GNU Radio uses shared pointers to keep track of blocks, so that they are automatically destroyed when they are no longer referenced. While investigating an issue in Gqrx, I noticed a bug in this system: in certain circumstances, a block that is added to a flowgraph continues to be referenced even after it is removed from the flowgraph.

This happens because `top_block_impl` keeps a flattened copy of the flowgraph:

https://github.com/gnuradio/gnuradio/blob/c3333eb0d7238c82442d1d901df03d5199557213/gnuradio-runtime/lib/top_block_impl.h#L67

The flattened copy is calculated in `top_block_impl::start` (and updated in `top_block_impl::restart` if the flowgraph is updated while it is running), but it is not discarded in `top_block_impl::stop`. As a result, `top_block_impl` keeps blocks alive even after they are removed from a flowgraph, since the flattened copy still references them and continues to do so until the next time the flowgraph is started.

This is particularly problematic for hardware blocks, since a hardware device is typically kept open until its block is destroyed. In Gqrx, it is currently impossible to reconfigure an RTL-SDR device because `top_block_impl` holds onto a reference to it indefinitely.

To solve the problem, I've changed `top_block_impl::stop` to release the reference to the flattened flowgraph, so that its blocks can be destroyed if they are no longer referenced. I've added a test which verifies that blocks are no longer referenced after they are removed from a stopped flowgraph. (Without the change to `top_block_impl.cc` this test fails.)

This change has one small side effect: `top_block_impl::edge_list`, `top_block_impl::msg_edge_list` and `top_block_impl::dump` all calculate their return values based on the cached flow graph. Currently they will continue to return information about a flow graph even after it is stopped (and if it is subsequently reconfigured, they will continue to return information about its previous state). After this change, they will only return information about a flow graph while it is running. Since the previous behaviour seems like a bug to me, it probably wouldn't be risky to backport this change to GNU Radio 3.8.

### DIFF
--- a/gnuradio-runtime/lib/top_block_impl.cc
+++ b/gnuradio-runtime/lib/top_block_impl.cc
@@ -122,6 +122,8 @@ void top_block_impl::stop()
     if (d_scheduler)
         d_scheduler->stop();
 
+    d_ffg.reset();
+
     d_state = IDLE;
 }
 
--- a/gr-blocks/lib/qa_gr_top_block.cc
+++ b/gr-blocks/lib/qa_gr_top_block.cc
@@ -282,3 +282,33 @@ BOOST_AUTO_TEST_CASE(t11_set_block_affinity)
     // least one thread core exists to use.
     BOOST_CHECK_EQUAL(set[0], ret[0]);
 }
+
+BOOST_AUTO_TEST_CASE(t12_release_shared_pointers)
+{
+    if (VERBOSE)
+        std::cout << "qa_top_block::t12()\n";
+
+    gr::top_block_sptr tb = gr::make_top_block("top");
+
+    gr::block_sptr src = gr::blocks::null_source::make(sizeof(int));
+    gr::block_sptr head = gr::blocks::head::make(sizeof(int), 100000);
+    gr::block_sptr dst = gr::blocks::null_sink::make(sizeof(int));
+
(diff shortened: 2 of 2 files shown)


bastibl:
I think the change might be fine but two quick comments:
- Since blocks are smart pointers, it seems strange that the code relies on the fact that it is destroyed at a particular point. 
- Hardware should not be claimed/released in constructor/destructor but in the block's start()/stop() functions, which are called in the context of the thread that will run the block.

argilo:
@bastibl It may well be that gr-osmosdr should be redesigned so that it doesn't keep hardware devices open after `stop()`. But I still think that GNU Radio's top block shouldn't hold onto references unnecessarily.

marcusmueller:
I'd agree with "GR shouldn't be holding references to things it won't use". I also agree with @bastibl: as usual, having had a formally written statement about what block authors should and shouldn't do would have been gold ... 10 years ago. So, we need to clean this up now :(

marcusmueller:
Regarding merging to 3.8: Let's track this in an issue. I'd like to verify this doesn't break anything on master for a week, and then it's a smooth cherry-pick to 3.8.

PR #3186, opened 2020-02-08
gr-fec: fix compiler warning

argilo:
The change to cc_encoder_impl.h fixes the following warning:
```
/home/argilo/prefix_39/src/gnuradio/gr-fec/lib/cc_encoder_impl.cc: In constructor ‘gr::fec::code::cc_encoder_impl::cc_encoder_impl(int, int, int, std::vector<int>, int, cc_mode_t, bool)’:
/home/argilo/prefix_39/src/gnuradio/gr-fec/lib/cc_encoder_impl.cc:71:23: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]
     if (d_start_state >= (1u << (d_k - 1))) {
         ~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~
```
I also explicitly wrote out `unsigned int` in cc_encoder_impl.cc for clarity.

### DIFF
--- a/gr-fec/lib/cc_encoder_impl.h
+++ b/gr-fec/lib/cc_encoder_impl.h
@@ -34,7 +34,7 @@ private:
     unsigned int d_rate;
     unsigned int d_k;
     std::vector<int> d_polys;
-    int d_start_state;
+    unsigned int d_start_state;
     cc_mode_t d_mode;
     int d_padding;
     int d_output_size;
--- a/gr-fec/lib/cc_encoder_impl.cc
+++ b/gr-fec/lib/cc_encoder_impl.cc
@@ -168,7 +168,7 @@ void cc_encoder_impl::generic_work(void* in_buffer, void* out_buffer)
     const unsigned char* in = (const unsigned char*)in_buffer;
     unsigned char* out = (unsigned char*)out_buffer;
 
-    unsigned my_state = d_start_state;
+    unsigned int my_state = d_start_state;
 
     if (d_mode == CC_TAILBITING) {
         for (unsigned int i = 0; i < d_k - 1; ++i) {


gr-fec/lib/cc_encoder_impl.h
```diff
@@ -34,7 +34,7 @@ class FEC_API cc_encoder_impl : public cc_encoder
     unsigned int d_rate;
     unsigned int d_k;
     std::vector<int> d_polys;
-    int d_start_state;
+    unsigned int d_start_state;
```
rear1019: Note that ``start_state`` is passed as ``signed int`` (``signed`` added for clarity :wink:). For consistency that should be changed as well – but this is an API change. I haven’t changed the type of ``d_start_state`` in #3103 as I wanted to backport the fixes therein to maint-3.[78].

There are also other parameters whose type should be changed to ``unsigned`` .

argilo: Are you proposing that I make those API changes in this pull request?

marcusmueller: Yeah, so hard call to make: on one hand, let's not change API arbitrarily, but really, the constructor has no business taking a signed int for something that's a bitfield.

So, considering that, for `master`, the saner thing might actually be to change the API to take an `unsigned int`. 

The fact that the variable type was changed to `signed` (instead of `signed int`) slipped my attention; that'd be the only place all over GNU Radio where we do that. I'm not changing that now, but it doesn't really strike me as aesthetically fitting ;)

michaelld: GR has -so- many API changes needed to go from `singed` to `unsigned` ... I'd prefer to not do that here but rather as a whole shebang change. Since we have a check for the max value of `d_start_state` already, I don't see a need to do more than what's in this PR right now. It's a simple and needed fix as is IMHO.

michaelld:
I wonder if any compiler(s) will complain when setting the `d_start_state` variable since it is `unsigned` while the variable being set to is `signed` ...

argilo:
I didn't get a warning about that.

michaelld:
> I didn't get a warning about that.

Yeah I'm thinking we won't get warnings because the setting is part of the instantiation list, rather than in some `if` or `for` (etc) statement doing comparisons.

Since this PR doesn't change any API and it fixes a valid issue, I'm ready to merge ... but I've tagged a few other GR devs just for the sake of it.

michaelld:
CI passes. 2 approvals. Simple tweaks that do not change API. Merging.

rear1019:
> I wonder if any compiler(s) will complain when setting the `d_start_state` variable since it is `unsigned` while the variable being set to is `signed` ...

``-Wsign-conversion``. Warning: Here be dragons…

michaelld:
> > I wonder if any compiler(s) will complain when setting the `d_start_state` variable since it is `unsigned` while the variable being set to is `signed` ...
> 
> `-Wsign-conversion`. Warning: Here be dragons…

OK so that's not a surprise. I'd guess there are a -ton- of issues in GR with that warning enabled. We need an API overhaul to deal with the use of `int` in the API while using `unsigned` internally.

PR #3187, opened 2020-02-09
cmake: declare compatibility with 3.8

argilo:
Many 3.8-compatible out-of-tree modules refuse to build with GNU Radio's master branch because they request GNU Radio 3.8, while the master branch calls itself 3.9 and declares itself incompatible with 3.8. With a modification to `GnuradioConfigVersion.cmake` GNU Radio can declare that it is still compatible with 3.8.

Perhaps the check should be more nuanced, but this might be reasonable as a quick fix. The extra check could be reverted if any major compatibility-breaking changes appear.


marcusmueller:
Hey ;) so, I saw the discussion you had on osmocom-sdr Friday, and forgot to answer, sorry.

So, how is upstream supposed to know what a breaking change is, for all the modules in the world?
We can't.

Thus, I disagree with @smunaut. I'll hook into that discussion on the osmo mailing list (For context:  [link to thread](http://lists.osmocom.org/pipermail/osmocom-sdr/2020-February/002050.html)) and we'll sort this out, together.

However, I really can't add a "we're probably compatible" flag, somewhere hidden in a generated CMake file, I'm afraid :(

marcusmueller:
Hoping to sort this out in #3188

PR #3190, opened 2020-02-09
Revert "uhd: Add support for UHD 4.0 feature: RFNoC"

marcusmueller:
Reverts gnuradio/gnuradio#3170. 

This sadly is necessary due to the blocking nature of `work` in the impl. See discussion on gr-iio upstreaming why we do no longer accept blocks that block in work without doing work.

In short: by returning 0, you can be sure that you'll be called again in 250 ms (or was it 100? I forget.) With a simple async that sends a "not done" to the system port of itself, a block can have a (nearly) arbitrary short blocking period if desired, without blocking  the executor loop – which is the technical reason we can't switch to a multiple-blocks-per-thread scheduler (which is stupid, because it eats a lot of resources to not respect data locality, as shown e.g. at GRCon San Diego by simpleXecutive, and by Bloessl'19).


smunaut:
Where is there blocking without work ? AFAICT during normal operation you'll always get some actual work done in work, but maybe I'm missing something here, the removal diff is kind of hard to read.

marcusmueller:
@smunaut there's the blocking `recv` call, which should be done with a 0 timeout, a `return 0` if nothing was ready, and a `std::async` sending a PMT to the `system` port.

smunaut:
I'm definitely unconvinced.

Although I recognize there has been abuse and "return 0 + wakeup" has a place and can take care of a bunch of cases, the case for high speed streaming IO interface is not one of them.

I see the need and benefit for multi-block per thread support but I think that blocks should be able to exclude themselves from that merging and declare themselves to be "IO threads". Without that you'll end up with situations where the block has no other choice but to ... spawn its own thread, which is just dumb to force them to do that ...

(1) HW buffers are typically small, you want to minimize _any_ latency between your vast and generous host memory buffers and the actual hardware. And I sure wouldn't want the scheduler to suddenly decide to place some computation block between two of my calls to dequeue/enqueue hardware buffers. It might not "block" but it can still take a few milli-seconds of doing math and at the current high speeds SDRs, that's a _lot_ of data, possibly enough to over/underflow small hw buffers.

(1b) Everything that's not absolute pure math can block for some time ... (even if indirectly, by for instance requiring a mutex and that mutex can be held by another thread that can block on something ...).
So unless you want to be insanely strict (we're talking basically no call whatsoever to any external API that doesn't explicitly guarantee no blocking or locking or any kind), you really don't want to merge that with your SDR hardware IO threads anyway.

(2) In normal flow mode, you're not expected to block (either at all or for very long) before you actually get data. Even if by some chance there is no data just ready now, the time it would take to setup a notification via another thread that some data has become available will end up taking longer than a small grace wait time in recv()

(3) You don't always get the choice of API you're dealing with.  Having a timeout in the read()/write() call is already pretty nice, you don't even get that all the time. Having the option to poll if data is there at all, also not always there. Being able to call those API from different threads ? Yeah, not always possible ... so you're only option is what ? Spawn a thread for IO and setup a FIFO between it and the GR block ... wait, isn't that what GR is supposed to do ?

And I'm very familiar with _having_ to do that ... fosphor spawns its own processing thread with a FIFO exactly because it doesn't fit in the GR model already and that sucks (it needs _large_ chunks of data at once and GR doesn't allow a block to control the min size of its input buffers). And now you want to force half (and I'm being generous) the SDR HW IO blocks to do that too ?

bastibl:
> (1) HW buffers are typically small, you want to minimize any latency between your vast and generous host memory buffers and the actual hardware. And I sure wouldn't want the scheduler to suddenly decide to place some computation block between two of my calls to dequeue/enqueue hardware buffers. It might not "block" but it can still take a few milli-seconds of doing math and at the current high speeds SDRs, that's a lot of data, possibly enough to over/underflow small hw buffers.

Maybe I don't get the point, but the "scheduler" is in this case the Linux scheduler. GNU Radio has no control about when it schedules which thread. So it might well schedule math blocks even though there are samples available in the SDR. AFAIS, there's not much we can do atm and that's completely independent from the blocking/non-blocking discussion. It's not that the thread would remain active/at the front of the runqueue or cannot be interrupted, just because it's blocking. What you suggest sounds more like DPDK (dedicating a core to IO).

smunaut:
I'm not talking about the linux scheduler here.

The point marcus was making is that in between calls to work() of one block, they want in the future to add calls to work of _another_ block in the _same_ OS thread. The whole "multiple block per thread" concept.

And as you point out the OS is already unpredictable enough and there is really no need for the gnuradio scheduler itself to make things worse for IO.

bastibl:
But then what are you advocating for:

1. We stick to one thread per block forever, block in work, and hope for the best. This is exactly the approach that leads to the random schedules that might mess up things. (as described in the previous post). To me, it sounds like you think blocking is fine, but you don't want the random schedules...
2. Have a dedicated IO thread (that blocks or polls continuously) for _high-performance IO_. We already have that as an advanced feature with DPDK in UHD. But I guess that's not what most users are going to use.
3. We have a thread that can handle multiple blocks. This is what could actually guarantee a reasonable schedule. If we don't have a separate thread for the SDR, we could at least guarantee something like: get all samples from SDR -> do some math -> get all samples from SDR -> do some math... I guess it won't get any better than that. And the external trigger is only needed, if there are no samples, which might only happen during startup.

smunaut:
I'm advocating that blocks should be able to declare themselves as having (short) blocking calls inside their work function and that those should remain in the "1 thread per block model".
All other blocks that don't declare themselves that way can't block at all and the GR scheduler is free to merge several blocks in a single worker thread.

Look at th GR-IIO ... Forcing them to _not_ block (so you can in the future save one thread) meant they now spawn their own thread ... today.   So in an attempt to have N-1 thread tomorrow, we will instead have N+1 thread today and N tomorrow ...

There is a VAST amount of SDRs between the slow few 100ksps and the X3xx with > 100 Msps that don't need DPDK style stuff and can perfectly handle their own IOs "nicely enough" as long as they just have a single normal OS thread to themselves (which GR provides currently) because the OS scheduling isn't all that bad.

bastibl:
> There is a VAST amount of SDRs between the slow few 100ksps and the X3xx with > 100 Msps that don't need DPDK style stuff and can perfectly handle their own IOs "nicely enough" as long as they just have a single normal OS thread to themselves (which GR provides currently) because the OS scheduling isn't all that bad.

I agree, that's why I don't think DPDK/a dedicated IO thread is what we should optimize for.

> Look at th GR-IIO ...

I did. Did you? The initial PR had (1) a dedicated IO thread with complicated locking (2) blocking in work (3) a strange form of busy-waiting by immediately scheduling again. 
All of these are anti-patterns. Some might not be avoidable, but there is no need to combine all of them in one block. The current version only blocks, is much simpler, and I don't see anyone complaining:

https://github.com/gnuradio/gnuradio/pull/2840/commits/4f0f3a5354ccc9a5dcfde3fa4ff94cbfb7be8188

smunaut:
There is a vast difference between DPDK style and a simple dedicated IO OS thread. The former is overkill for most SDRs, the second is not.

For gr-iio, fine, it was mostly based on what I overheard and looks like it had way more wrong than blocking. But that doesn't change my original point.

In any case, I'm done, I said all I wanted and it's clear I'm not changing anyone's mind and I'm just banging my head against the wall.

PR #3191, opened 2020-02-10
dtv: enhance ATSC RX example

devnulling:
- Renamed `uhd_rx_atsc.grc` to `uhd_atsc_rx.grc`
- Add QT GUI Chooser populated with NA ATSC channels 2-36, plus CATV 58 (429 MHz)
	- The Frequency values are the lower edge (center_freq-3e6)
	- 3e6 is added to the `channel` value within `center_freq`
	- Leaves `center_freq` as a Entry widget in case someone wants to hard set a freq
- Switch `gain` to use Normalized values in UHD Source
- Switch QT GUI Frequency sink to have 2 inputs => RX Spectrum and post RRC Filter
- Update FFT RRC Filter to use 50 taps instead of 18
	- See @drmpeg's post here: https://github.com/gnuradio/gnuradio/pull/3154#issuecomment-582143764
	- I've observed improved performance with 50 taps too
- Updated the QT grid layout for the widgets

atsc_fixup

### DIFF
--- a/gr-dtv/examples/CMakeLists.txt
+++ b/gr-dtv/examples/CMakeLists.txt
@@ -42,7 +42,7 @@ install(
     README.dvbt2
     uhd_atsc_capture.grc
     uhd_atsc_tx.grc
-    uhd_rx_atsc.grc
+    uhd_atsc_rx.grc
     vv001-cr35.grc
     vv003-cr23.grc
     vv004-8kfft.grc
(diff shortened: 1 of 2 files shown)


michaelld:
@devnulling do you want to add the extra bands per @drmpeg 's comment? I think this PR is good either way.

devnulling:
I'd like to push up a change before merge to include those extra channels.

devnulling:
I've updated this example to include the extra CATV channels suggested by @drmpeg for amateur use.

One other minor update was changing the flowgraphs `option`'s block `id` to be consistent with the file name and TX flowgraph (uhd_rx_atsc => uhd_atsc_rx)

PR #3193, opened 2020-02-11
gr-dtv: Use new FFT allocation scheme for DVB-T2 blocks.

drmpeg:


gr-dtv/lib/dvbt2/dvbt2_p1insertion_cc_impl.cc
```diff
@@ -43,9 +43,25 @@ dvbt2_p1insertion_cc_impl::dvbt2_p1insertion_cc_impl(dvbt2_extended_carrier_t ca
...
+      imag_negative_threshold(-vclip),
+      real_positive_threshold_count(0),
+      real_negative_threshold_count(0),
+      imag_positive_threshold_count(0),
+      imag_negative_threshold_count(0),
+      p1_fft(1024, false, 1)
 {
     int s1, s2, index = 0;
+    int p1_fft_size = 1024;
+    int fft_size, N_P2, guard_interval;
```
michaelld: I had to double take the difference between `fftsize` and `fft_size` here and in the `switch` below. They are clearly different... but would you please make them even more different to help clarify the code a little? Maybe `fftsize_type` on the input?

drmpeg: I changed the local variable to symbol_size.

PR #3198, opened 2020-02-12
gr-dtv: Add QA test for DVB and DVB-T2 blocks.

drmpeg:
comment deleted.


drmpeg:
It's actually the smallest BBC test vector. I thought it would be cool to use an external reference. For the pilot generator block, it's only one iteration of work().

Since it doesn't compress very much, it would increase the clone size from 54 to 57 MiB. Is that really a big deal? This is 2020. The current Linux repository is 2.7 GiB and the largest file is 14,151,728 bytes.

mbr0wn:
We need more of this kind of testing. I think the 3 MiB don't make a huge difference, but it would be nice (in particular for gr-dtv) if we had another way of testing against binary data, something like an official repository of test vectors. So let's say we do come up with this other way of storing data, then we would also not need to check in this file.

This discussion -- on another separate repository of binary data -- comes up at irregular intervals since I joined the project. Like @drmpeg says, it's 2020, we should be able to come up with something. Let me tag this "Don't Merge" for now, maybe there's something we can come with fairly quickly.

mbr0wn:
@drmpeg please poke us if we don't do anything here within the next 2 weeks. This kind of test is super valuable.

drmpeg:
I'm certainly open to an alternative solution. Just for reference, these test vectors come from here:

ftp://ftp.kw.bbc.co.uk/t2refs/streams/VV009-4KFFT_CSP.zip

I've converted the text files VV009-4KFFT_TP01_CSP.txt and VV009-4KFFT_TP19_CSP.txt in that .zip file to binary (vv009-4kfft.ts and the large vv009-4kfft.cfile respectively).

devnulling:
@noc0lour is this something we can host on gnuradio.org, or perhaps spin up another subdomain for? perhaps files.gnuradio.org ?

noc0lour:
if the license permits we can host these test-vectors.

drmpeg:
Here's the copyright statement on each file. Looks like distributing binaries isn't allowed without permission.
```
% DVB-T2 verification & validation working group
% Modulator reference stream file
%
% Created by the Common Simulation Platform (https://corben.us.es)
%
% Notification to Recipient
%
% This data is copyright (c) Digital Video Broadcasting Project (DVB) 2010
% Distribution of this data is permitted but only in unmodified form,
% unless prior written permission has been obtained from DVB.
% At the date of creation of this data, DVB believe this data to be accurate 
% (according to the DVBT2 standard) and that it does not infringe third party rights,
% however DVB do not provide any warranties or guarantees in this respect.
% Any other warranties are excluded to the fullest extent permitted in law.
% The manner in which this data is used is entirely at the election of the user.
%
% DVB Project website: www.dvb.org
```

noc0lour:
Hm, so we might be able to host the originals and convert them to binary during the test? That would also be fine I think. We can have these tests run only once a week or something.

drmpeg:
This is getting too messy for me. Would everyone be happy with a 388,096 byte complex file? I can do this with a home grown test vector.

The current largest file in the source tree is 358,544 bytes.

michaelld:
> This is getting too messy for me. Would everyone be happy with a 388,096 byte complex file? I can do this with a home grown test vector.
> 
> The current largest file in the source tree is 358,544 bytes.

I'd be fine with this "small" size file. I'd love to see GR host such test / example binary files, per the discussion above!

mormj:
I think 388k would be ok in the source tree - long term it would be very nice to have more comprehensive testing with binary files.  Could see this done throughout the gr modules.  In which case we wouldn't be able to put all the bins in git.

Perhaps something like git lfs to tie the binary files into git, but have them hosted on gnuradio.org?
https://git-lfs.github.com/

noc0lour:
I personally wouldn't want to tie us to GitHubs LFS. If we want to do this we need to run our own Git$Foo. Hosting binaries on GR infra is absolutely no problem, but of course having test files in tree is the way to go. I think the files @drmpeg proposes are reasonably small to pull them in.

drmpeg:
The text version of the files can be fairly large. The largest is 222,155,705 bytes. It's because the format looks like this:
```
% Test point 19 (complex data)
%
% Generated on 2010-09-21 19:17
%
# frame 1 
# block 1 of 105
-7.144345e-01 +0.000000e+00
+6.053111e-01 +1.042586e+00
+1.471985e+00 +5.937135e-01
+8.744205e-02 +1.208740e+00
+2.175685e-01 +2.385738e-01
+1.226565e+00 -4.776997e-01
+2.290761e-01 +3.327809e-01
+3.202761e-03 -1.307125e+00
+4.692896e-01 -1.163361e+00
+3.541141e-02 +6.420036e-01
+5.393088e-01 -3.283025e-01
+1.761048e-01 +4.941161e-02
...
```
Let me make sure I'm on the same page here. Any hosting of test files would require a wget or curl during "make test". Yes? In other words, the dtv portion of "make test" could fail if you can't download a test file. Then there's embedded systems that don't have Internet access.

It just seems super ugly to me.

noc0lour:
@drmpeg Either wget/curl or if we stay in python we can use built-in requests or similar to retrieve the files. In case we add a lot more of binary files it would make sense to host them outside the tree and then these tests have to be made optional. Due to constraints on network/firewalls etc on the user-side.

For now I would suggest taking the file in-tree and then we'll see.

mbr0wn:
We need to take the file storage discussion offline. Eventually, I would love to see the bigger test vector to come back, and a separate test target for GNU Radio which runs against binary test files (`make test_signals` or something like that). They don't only have to be official test vectors, we could also have OTA recordings (with noise and all) to make sure our equalizers and channel coding work.

Let's assume that we accumulate a fair amount of test data, let's say it's on the order of 1 GiB for the sake of discussion. In that case, it would probably be important to save bandwidth, right (@noc0lour is probably the best person to ask this)? What's a good way to store this kind of thing -- via bittorrent? Or maybe we have a manifest file for our binaries, and we do something like UHD does for FPGA images for people to download them?

mbr0wn:
@drmpeg Sorry you got caught in this discussion. Your test vectors are very welcome, and if it were only the 3 MiB file, it wouldn't be an issue, but I think it's good if we can figure out a way to store -- potentially -- Gigabytes of test signals for better testing.

PR #3200, opened 2020-02-12
gr-newmod: Update MANIFEST with `gr_supported_version`

sbmueller:
Backport of #3189 

Cgran was extended to display supported GNU Radio versions of OOTs. To
leverage this feature, a `gr_supported_version` key needs to be provided
by the OOT's MANIFEST. This commit extends gr-newmod's MAIFEST by this
key.


devnulling:
BB failures are unrelated.

PR #3202, opened 2020-02-13
gr-blocks: Add Msg to Var and Var to Msg Conversion Blocks

ghostop14:
These 3 new blocks provide the missing glue to move between messages
and variables in the same flowgraph.  There are 3 variants here:
1. Monitor a variable and produce a user-specified message (pair)
when the variable changes.  Useful bridging standard GUI controls
to message-based blocks.
2. When an inbound message (pair) is received, update a specified
variable.
3. When an inbound message (dict) is received, extract a single
dictionary entry and produce a message pair (useful if you have a
multi-value dictionary but you just want to pull off a single
attribute such as frequency or gain without modifying the upstream
block. This can be paired with (2) if necessary to move from a
dictionary item to a variable.

### DIFF
--- a/gr-blocks/python/blocks/__init__.py
+++ b/gr-blocks/python/blocks/__init__.py
@@ -24,6 +24,9 @@ except ImportError:
     from .blocks_swig import *
 
 from .stream_to_vector_decimator import *
+from .msg_meta_to_pair import meta_to_pair
+from .msg_pair_to_var import msg_pair_to_var
+from .var_to_msg import var_to_msg_pair
 
 #alias old add_vXX and multiply_vXX
 add_vcc = add_cc
--- a/gr-blocks/python/blocks/msg_meta_to_pair.py
+++ b/gr-blocks/python/blocks/msg_meta_to_pair.py
@@ -1,0 +1,67 @@
+#!/usr/bin/env python
+#
+# Copyright 2020 Free Software Foundation, Inc.
+#
+# This file is part of GNU Radio
+#
+# SPDX-License-Identifier: GPL-3.0-or-later
+#
+#
+
+from gnuradio import gr
+import pmt
+
+class meta_to_pair(gr.sync_block):
+    """
+    This block converts a metadata dictionary item to a pmt pair that is
+    compatible with other blocks expecting a pair in.  You can specify
+    which item in the incoming metadata to output as a pair and what
+    the pair name is.
+    """
+    def __init__(self, incomingKeyName, outgoingPairName):
+        gr.sync_block.__init__(self, name="meta_to_pair", in_sig=None, out_sig=None)
+
+        self.incomingKeyName = str(incomingKeyName)
+        self.outgoingPairName = str(outgoingPairName)
+
+        self.message_port_register_in(pmt.intern("inmeta"))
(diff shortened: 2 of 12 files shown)


gr-blocks/python/blocks/msg_meta_to_pair.py
```diff
@@ -0,0 +1,67 @@
+#!/usr/bin/env python
+#
+# Copyright 2012 Free Software Foundation, Inc.
```
devnulling: Please change to 2020

gr-blocks/python/blocks/msg_pair_to_var.py
```diff
@@ -0,0 +1,36 @@
+#!/usr/bin/env python
+#
+# Copyright 2012 Free Software Foundation, Inc.
```
devnulling: 2020

gr-blocks/python/blocks/var_to_msg.py
```diff
@@ -0,0 +1,38 @@
+#!/usr/bin/env python
+#
+# Copyright 2012 Free Software Foundation, Inc.
```
devnulling: 2020

ghostop14:
Just pushed the updates.

gr-blocks/python/blocks/__init__.py
```diff
@@ -24,6 +24,9 @@
     from .blocks_swig import *
 
 from .stream_to_vector_decimator import *
+from .msg_meta_to_pair import MetaToPair
```
michaelld: thinking the internal CamelCase should be moved to lower_case ...

michaelld:
wondering what @marcusmueller thinks of CamelCase in Python names ...

ghostop14:
@michaelld: it was only 3 small classes, so in the interest of time I changed them to expedite.  

I know snake is the GR standard (and I'll adapt) but love the camel! QtGui for example uses capitalizations, with classes properly capitalized.  Since Qt is out of our control, it does create a mixed standard already for any Qt gui-based controls.

michaelld:
> @michaelld: it was only 3 small classes, so in the interest of time I changed them to expedite.
> 
> I know snake is the GR standard (and I'll adapt) but love the camel! QtGui for example uses capitalizations, with classes properly capitalized. Since Qt is out of our control, it does create a mixed standard already for any Qt gui-based controls.

@ghostop14 yes I totally get the mixed standard. I prefer like "the GR powers that be" to make such decisions. The buck stops with @marcusmueller , regardless of how much he's delegated out to us GR team devs!

michaelld:
@devnulling did you test out this PR locally? there's no real CI here for it right now! It reads well to me, but I haven't actually tested it.

michaelld:
If you did test it & it works to your specs, please merge

devnulling:
I've used these blocks when they existed in @ghostop14 's OOT. I just pinged ghostop14 about adding an example to demo these.

michaelld:
@ghostop14 are you done with this beyond any further review requests?

ghostop14:
@michaelld yes I think everything's incorporated.  Should be good to go.

michaelld:
CI passes (even if this PR doesn't provide new tests, at least current ones pass). Positive reviews. Tested locally with success. Merging.

PR #3208, opened 2020-02-15
gr-blocks: Added scalar to Complex to/from ishort and ichar

ghostop14:
The existing block combinations of complex-to-ichar/ichar-to-complex
and complex-to-ishort/ishort-to-complex was not exposing a scale
factor to the UI (it was hard-coded at 1.0) which prevented the blocks
from being used with voltage-based inputs and outputs, for example
a -127 to 127 byte input would be mapped directly to the same value
as a float.  The same was true in reverse and there were notes in
the code about FIX clipping.  Adding the scale factor provides a
mechanism for the user to appropriately fix the clipping and scale
the conversion correctly.  Additional documentation was added to
the block yml to provide guidance to users on how to appropriately
select the scale factor for their use case, and the default value
was set to 1.0 for backward compatibility.


ghostop14:
@michaelld I noticed one other item while I was in the code, that the volk call to volk_8i_s32f_convert_32f() uses an int8_t data type inside.  I'm wondering the impact of that if someone wanted to scale complex to unsigned char output since int8_t is signed.  One thought I had was a check to see if the scale factor is > 127.0 and cast the input as unsigned in the volk call versus signed if it's <= 127?  Thoughts?

ghostop14:
I think I'm good with the latest.  I just rewrote the yml doc to talk through signed.

gr-blocks/lib/complex_to_interleaved_short_impl.h
```diff
@@ -19,10 +19,11 @@ namespace blocks {
 class BLOCKS_API complex_to_interleaved_short_impl : public complex_to_interleaved_short
 {
 private:
+    float d_scalar;
     bool d_vector;
 
 public:
-    complex_to_interleaved_short_impl(bool vector);
+    complex_to_interleaved_short_impl(float scale_factor, bool vector);
```
mormj: Since the scale_factor currently is hardcoded to 1.0 - should the new parameter to these blocks be placed at the end with the default value so that the API is unchanged?  That way we can backport to 3.8 and also not force anyone to modify their flowgraphs on master.

gr-blocks/grc/blocks_complex_to_interleaved_char.block.yml
```diff
@@ -33,4 +37,16 @@ cpp_templates:
         'True': 'true'
         'False': 'false'
 
+documentation: |-
```
michaelld: So nice to get some documentation in GRC for these blocks!

gr-blocks/include/gnuradio/blocks/complex_to_interleaved_char.h
```diff
@@ -39,7 +39,7 @@ class BLOCKS_API complex_to_interleaved_char : virtual public sync_interpolator
     /*!
      * Build a complex to interleaved chars block.
      */
-    static sptr make(bool vector = false);
+    static sptr make(float scale_factor = 1.0, bool vector = false);
```
michaelld: I'm with @mormj here: let's put the new parameter -after- the current one, so that the prior API will still work; just augmenting it.

gr-blocks/include/gnuradio/blocks/complex_to_interleaved_short.h
```diff
@@ -39,7 +39,7 @@ class BLOCKS_API complex_to_interleaved_short : virtual public sync_interpolator
     /*!
      * Build a complex to interleaved shorts block.
      */
-    static sptr make(bool vector = false);
+    static sptr make(float scale_factor = 1.0, bool vector = false);
```
michaelld: Since `scale_factor` is a potentially runtime-settable parameter, please add in the methods for doing so.

gr-blocks/lib/complex_to_interleaved_char_impl.cc
```diff
@@ -37,13 +41,11 @@ int complex_to_interleaved_char_impl::work(int noutput_items,
...
-    const gr_complex* in = (const gr_complex*)input_items[0];
+    const float* in = (const float*)input_items[0];
     char* out = (char*)output_items[0];
 
-    int npairs = (d_vector ? noutput_items : noutput_items / 2);
-    for (int i = 0; i < npairs; i++) {
-        *out++ = (char)lrintf(in[i].real()); // FIXME saturate?
-        *out++ = (char)lrintf(in[i].imag());
+    for (int i = 0; i < noutput_items; i++) {
+        *out++ = (char)lrintf(in[i] * d_scalar);
```
michaelld: (1) Since we're using C++11, let's move to include `cmath` and then use `std::rint`.

(2) What happens if the abs output of `rint` is > SCHAR_MAX? Thinking we want to define this behavior, and guessing what we want is to clip. If you feel otherwise, please go with that here & elsewhere, and add whatever you decide to the documentation too.

(3) instead of a direct cast I think we want a `static_cast` here and elsewhere for the conversion.

gr-blocks/lib/complex_to_interleaved_char_impl.cc
```diff
@@ -37,13 +41,11 @@ int complex_to_interleaved_char_impl::work(int noutput_items,
...
 {
-    const gr_complex* in = (const gr_complex*)input_items[0];
+    const float* in = (const float*)input_items[0];
     char* out = (char*)output_items[0];
 
-    int npairs = (d_vector ? noutput_items : noutput_items / 2);
-    for (int i = 0; i < npairs; i++) {
-        *out++ = (char)lrintf(in[i].real()); // FIXME saturate?
-        *out++ = (char)lrintf(in[i].imag());
+    for (int i = 0; i < noutput_items; i++) {
```
michaelld: Could potentially move to using Volk's `volk_32f_s32f_convert_8i` here: `void volk_32f_s32f_convert_8i(int8_t* outputVector, const float* inputVector, const float scalar, unsigned int num_points)` ... seems to match up well with this use case! There are versions for `16i` and `32i` as well.

gr-blocks/lib/interleaved_char_to_complex_impl.cc
```diff
@@ -19,16 +19,20 @@
...
 
-interleaved_char_to_complex_impl::interleaved_char_to_complex_impl(bool vector_input)
+interleaved_char_to_complex_impl::interleaved_char_to_complex_impl(float scale_factor,
+                                                                   bool vector_input)
     : sync_decimator("interleaved_char_to_complex",
                      gr::io_signature::make(1, 1, (vector_input ? 2 : 1) * sizeof(char)),
                      gr::io_signature::make(1, 1, sizeof(gr_complex)),
-                     vector_input ? 1 : 2)
+                     vector_input ? 1 : 2),
+      d_scalar(scale_factor)
```
michaelld: whitespace here should match that above ... or that above should match here ...

gr-blocks/lib/interleaved_short_to_complex_impl.cc
```diff
@@ -43,9 +46,10 @@ int interleaved_short_to_complex_impl::work(int noutput_items,
                                             gr_vector_void_star& output_items)
 {
     const short* in = (const short*)input_items[0];
-    gr_complex* out = (gr_complex*)output_items[0];
+    float* out = (float*)output_items[0];
 
-    interleaved_short_array_to_complex(in, out, 2 * noutput_items);
+    // This calculates in[] * 1.0 / d_scalar
+    volk_16i_s32f_convert_32f_u(out, in, d_scalar, 2 * noutput_items);
```
michaelld: Nice use of Volk! I'm not sure you have to specify `_u` for the unaligned here. Can you do benchmarks on these, to see if there's any difference between adding `_u` and not?

gr-blocks/lib/complex_to_interleaved_char_impl.cc
michaelld: .. or ... just use Volk as noted above & all of this goes away ... poof!

michaelld:
> @michaelld I noticed one other item while I was in the code, that the volk call to volk_8i_s32f_convert_32f() uses an int8_t data type inside. I'm wondering the impact of that if someone wanted to scale complex to unsigned char output since int8_t is signed. One thought I had was a check to see if the scale factor is > 127.0 and cast the input as unsigned in the volk call versus signed if it's <= 127? Thoughts?

@ghostop14 All of the Volk "float to int" converters use signed internally and externally ... or a quick look at the code makes me think that's the case. Let's make sure the GR side matches up and doesn't use unsigned ... I don't think any of these blocks or changes do ... but maybe I'm overlooking something here?

ghostop14:
@michaelld all great feedback.  I think I have everything incorporated.  set_scalar callback added, moved the 2 non-volk versions to volk, removed the _u in the other volk calls.  This was a legacy item in the original code I just hadn't touched.  But agreed, not sure why it was there.  Added alignment sets in all the constructors.  Maybe originally that was why.  New variables moved to end of constructors for backward compatibility, and alignment taken care of with clang-format.

ghostop14:
@michaeld interesting legacy bug find in troubleshooting a qa-test failure.  The interleaved->complex versions supported a bool vector parameter but conversions this way never used it in the work function to adjust the number of data points to account for vectors.  So the current/original blocks would've produced bad data with the vector param set (I guess nobody uses it).  Anyway I fixed it but just maybe 2nd set of eyes check my d_vector test going into the volk calls in the interleaved_<type>_to_complex_impl.cc work functions.

gr-blocks/lib/complex_to_interleaved_char_impl.cc
```diff
@@ -37,14 +42,11 @@ int complex_to_interleaved_char_impl::work(int noutput_items,
...
-    int npairs = (d_vector ? noutput_items : noutput_items / 2);
-    for (int i = 0; i < npairs; i++) {
-        *out++ = (char)lrintf(in[i].real()); // FIXME saturate?
-        *out++ = (char)lrintf(in[i].imag());
-    }
+    const float* in = (const float*)input_items[0];
+    int8_t* out = (int8_t*)output_items[0];
+
+    volk_32f_s32f_convert_8i(
+        out, in, d_scalar, d_vector ? noutput_items * 2 : noutput_items);
```
michaelld: Thinking through this: If a vector of length 2x sizeof(char), noutput_items will be in # of vectors, so it needs 2x to get the correct number of total elements to manipulate. If a vector of length 1x, then the data will in in I/Q pairs, so you'll need 2x noutput_items for that option. In other words, either way you need 2x noutput_items ... so, please fix that & add something like what I wrote here as a comment for future reference. Here and for `short` too.

michaelld: Or maybe I'm wrong? been a long day ...

gr-blocks/lib/interleaved_char_to_complex_impl.cc
```diff
@@ -38,10 +43,12 @@ int interleaved_char_to_complex_impl::work(int noutput_items,
...
                                            gr_vector_void_star& output_items)
 {
-    const int8_t* in = (const int8_t*)input_items[0];
     float* out = (float*)output_items[0];
+    const int8_t* in = (const int8_t*)input_items[0];
 
-    volk_8i_s32f_convert_32f_u(out, in, 1.0, 2 * noutput_items);
+    // This calculates in[] * 1.0 / d_scalar
+    volk_8i_s32f_convert_32f(
+        out, in, d_scalar, d_vector ? 4 * noutput_items : 2 * noutput_items);
```
michaelld: should be `2 * noutput_items` either way again, yes?

gr-blocks/lib/interleaved_short_to_complex_impl.cc
```diff
@@ -43,13 +48,15 @@ int interleaved_short_to_complex_impl::work(int noutput_items,
                                             gr_vector_void_star& output_items)
 {
     const short* in = (const short*)input_items[0];
-    gr_complex* out = (gr_complex*)output_items[0];
+    float* out = (float*)output_items[0];
 
-    interleaved_short_array_to_complex(in, out, 2 * noutput_items);
+    // This calculates in[] * 1.0 / d_scalar
+    volk_16i_s32f_convert_32f(
+        out, in, d_scalar, d_vector ? 4 * noutput_items : 2 * noutput_items);
```
michaelld: `2 * noutput_items` here and in the `for` loop below ... yes?

gr-blocks/lib/interleaved_char_to_complex_impl.cc
michaelld: also: do we want to add in a `swap` like the `short` version has? I honestly don't know why this block doesn't have the option.

ghostop14:
I think I have the math right.... for char/short->complex, 2 chars/shorts go into a single complex with a block decim=2.  So num input samples interleaved is 2*noutput_items.  If we have a vector of length 2, then it's twice that or 2*(2*noutput_items).  So I think it's correct. (tests validate it works, but just wanted a 2nd set of eyes).

I wonder if the other block had a swap option since there was no native swap block (you'd have to do it with a few blocks or a conj).  Now that we have one, I don't know we need to add another loop here.  Thoughts?

michaelld:
> I think I have the math right.... for char/short->complex, 2 chars/shorts go into a single complex with a block decim=2. So num input samples interleaved is 2_noutput_items. If we have a vector of length 2, then it's twice that or 2_(2*noutput_items). So I think it's correct. (tests validate it works, but just wanted a 2nd set of eyes).
> 
> I wonder if the other block had a swap option since there was no native swap block (you'd have to do it with a few blocks or a conj). Now that we have one, I don't know we need to add another loop here. Thoughts?

OK ... let me think about this again ... quickly too!

gr-blocks/lib/complex_to_interleaved_char_impl.cc
michaelld: Ping on my thinking here. I think I'm correct for the "complex to *" blocks. I'll comment on the "* to complex" blocks in another comment.

gr-blocks/lib/interleaved_char_to_complex_impl.cc
michaelld: Here's one way to look at this question: suppose we want 1 output item. If vector input, this requires 1 input item (a vector of length 2), for a total of 2 `char` inputs; if not vector, this requires 2 input items, for a total of 2 `char` inputs. Either way, we require a total of 2 `char` inputs to generate this 1 output item. So the number of items for `volk_8i_s32f_convert_32f` to process is 2 for this specific situation. Hence we always require `2 * noutput_items` input `char` to generate `noutput_items` outputs. Yes?

michaelld: _thou must count to 2. 2 shall be the number of the counting and the number of the counting shall be 2. 3 shalt thou not count, neither shalt thou count 1, excepting that thou then proceedeth to 2. 4 is right out._

gr-blocks/lib/complex_to_interleaved_char_impl.cc
michaelld: Same logic here as I just noted for the "X to complex" block, just in reverse: 1 input item will generate either 1 vector item output or 2 non-vector items output. Either way, for Volk purposes, we're processing 2 items. Hence, we always need to process `2 * noutput_items` items in Volk.

ghostop14: Logic on these as an interpolator goes that noutput_items with an interp of 2 will be 2x the number of complex's in, but will match1-to-1 the number of floats in.  So it should be a direct map (noutput_items).  If it's a vector in and there's twice as much in then just 2*noutput_items.

michaelld: OK. This is correct. I was messing up `noutput_items` values. On to the other way.

michaelld:
Waiting for CI ... just to make sure nothing is broken one way or the other

michaelld:
All CI passes. 2 maintainer approvals. Merging.

PR #3212, opened 2020-02-17
gr-filter: Add ichar and ishort decimator

ghostop14:
Current resamplers do not support ichar as an input type.  However,
when processing high-speed wide bandwidth streaming ichar 
data from external sources (such as the ATA), decimation in 
native ichar would be faster than converting
to another data type then decimating.  This block provides the ability
to decimate the ichar stream directly.

### DIFF
--- a/gr-filter/grc/filter.tree.yml
+++ b/gr-filter/grc/filter.tree.yml
@@ -26,6 +26,7 @@
   - pfb_arb_resampler_xxx
   - rational_resampler_xxx
   - rational_resampler_base_xxx
+  - ival_decimator
 - Channelizers:
   - freq_xlating_fft_filter_ccc
   - freq_xlating_fir_filter_xxx
--- a/gr-filter/grc/filter_ival_decimator.block.yml
+++ b/gr-filter/grc/filter_ival_decimator.block.yml
@@ -1,0 +1,34 @@
+id: ival_decimator
+label: Interleaved Stream Decimator
+
+parameters:
+-   id: datatype
+    label: Input Type
+    dtype: enum
+    options: [byte, short]
+    option_attributes:
+        datasize: [gr.sizeof_char, gr.sizeof_short]
+    hide: part
+-   id: decimation
+    label: Decimation
+    dtype: int
+    default: '1'
+
+inputs:
+-   domain: stream
+    dtype: ${ type }
+
+outputs:
+-   domain: stream
+    dtype: ${ type }
+    
+templates:
+    imports: from gnuradio import filter
+    make: filter.ival_decimator(${decimation}, ${datatype.datasize})
+
+documentation: |-
+    This block will directly decimate an incoming stream made up of the specified complex samples.  
+    One example would be if you have a source streaming 8-bit complex at high speeds and you want to 
+    decimate directly from a high-speed source, or before writing to a file or network sink.
+    
+file_format: 1
--- a/gr-filter/include/gnuradio/filter/ival_decimator.h
+++ b/gr-filter/include/gnuradio/filter/ival_decimator.h
@@ -1,0 +1,40 @@
+/* -*- c++ -*- */
+/*
+ * Copyright 2020 Free Software Foundation, Inc.
+ *
+ * This file is part of GNU Radio
+ *
+ * SPDX-License-Identifier: GPL-3.0-or-later
+ *
+ */
+
+#ifndef INCLUDED_IVAL_DECIMATOR_H
+#define INCLUDED_IVAL_DECIMATOR_H
+
+#include <gnuradio/filter/api.h>
+#include <gnuradio/sync_decimator.h>
+
+namespace gr {
+namespace filter {
+
+/*!
+ * \brief Filter-Delay Combination Block.
+ * \ingroup resamplers_blk
+ *
+ * \details
+ * This block provides interleaved char and short decimation without
+ * first having to convert to a type supported by a resampler.
+ *
+ */
+class FILTER_API ival_decimator : virtual public gr::sync_decimator
+{
+public:
+    typedef boost::shared_ptr<ival_decimator> sptr;
+
+    static sptr make(int decimation, int data_size);
+};
+
+} // namespace filter
+} // namespace gr
+
+#endif /* INCLUDED_IVAL_DECIMATOR_H */
--- a/gr-filter/lib/ival_decimator_impl.cc
+++ b/gr-filter/lib/ival_decimator_impl.cc
@@ -1,0 +1,93 @@
+/* -*- c++ -*- */
+/*
+ * Copyright 2020 Free Software Foundation, Inc.
+ *
+ * This file is part of GNU Radio
+ *
+ * SPDX-License-Identifier: GPL-3.0-or-later
+ *
+ */
+
+#ifdef HAVE_CONFIG_H
+#include "config.h"
+#endif
+
+#include "ival_decimator_impl.h"
+#include <gnuradio/io_signature.h>
+
+namespace gr {
+namespace filter {
+
+
+ival_decimator::sptr ival_decimator::make(int decimation, int data_size)
+{
+    return gnuradio::get_initial_sptr(new ival_decimator_impl(decimation, data_size));
+}
+
+/*
+ * The private constructor
+ */
+ival_decimator_impl::ival_decimator_impl(int decimation, int data_size)
+    : gr::sync_decimator("ival_decimator",
+                         gr::io_signature::make(1, 1, data_size),
+                         gr::io_signature::make(1, 1, data_size),
+                         decimation),
+      d_data_size(data_size)
+{
+    if ((data_size != 1) && (data_size != 2))
+        throw std::invalid_argument("data size must be in 1 (char) or 2 (short)");
+
+    d_doubledecimation = 2 * decimation;
+
+    // Make sure we work with pairs of bytes (I and Q as byte)
+    gr::block::set_output_multiple(2);
+}
+
+/*
+ * Our virtual destructor.
+ */
+ival_decimator_impl::~ival_decimator_impl() {}
+
+int ival_decimator_impl::work(int noutput_items,
+                              gr_vector_const_void_star& input_items,
+                              gr_vector_void_star& output_items)
+{
+    long i = 0;
+    int octr = 0;
+
+    long nin = noutput_items * decimation();
+
+    switch (d_data_size) {
+    case 2: {
+        const int16_t* in = (const int16_t*)input_items[0];
+        int16_t* out = (int16_t*)output_items[0];
+
+        while (i < nin) {
+            *out++ = in[i];
+            *out++ = in[i + 1];
+
+            i += d_doubledecimation;
+            octr += 2;
+        }
+    } break;
+
+    case 1: {
+        const char* in = (const char*)input_items[0];
+        char* out = (char*)output_items[0];
+
+        while (i < nin) {
+            *out++ = in[i];
+            *out++ = in[i + 1];
+
+            i += d_doubledecimation;
+            octr += 2;
+        }
+    } break;
+    }
+
+    // Tell runtime system how many output items we produced.
+    return octr;
+}
+
+} /* namespace filter */
+} /* namespace gr */
--- a/gr-filter/examples/test_ichar_decim.grc
+++ b/gr-filter/examples/test_ichar_decim.grc
@@ -1,0 +1,299 @@
+options:
+  parameters:
+    author: ''
+    catch_exceptions: 'True'
+    category: '[GRC Hier Blocks]'
+    cmake_opt: ''
+    comment: ''
+    copyright: ''
+    description: ''
+    gen_cmake: 'On'
+    gen_linking: dynamic
+    generate_options: qt_gui
+    hier_block_src_path: '.:'
+    id: test_ichar_decim
+    max_nouts: '0'
+    output_language: python
+    placement: (0,0)
+    qt_qss_theme: ''
+    realtime_scheduling: ''
+    run: 'True'
+    run_command: '{python} -u {filename}'
+    run_options: prompt
+    sizing_mode: fixed
+    thread_safe_setters: ''
+    title: IChar Decimator
+  states:
+    bus_sink: false
+    bus_source: false
+    bus_structure: null
+    coordinate: [8, 8]
+    rotation: 0
+    state: enabled
+
+blocks:
+- name: decim
+  id: variable
+  parameters:
+    comment: ''
+    value: '5'
+  states:
+    bus_sink: false
+    bus_source: false
+    bus_structure: null
+    coordinate: [381, 20]
+    rotation: 0
+    state: true
+- name: samp_rate
+  id: variable
+  parameters:
+    comment: ''
+    value: '500000'
+  states:
+    bus_sink: false
+    bus_source: false
+    bus_structure: null
+    coordinate: [184, 12]
+    rotation: 0
+    state: enabled
+- name: analog_sig_source_x_0
+  id: analog_sig_source_x
+  parameters:
+    affinity: ''
+    alias: ''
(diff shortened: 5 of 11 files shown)


mormj:
@ghostop14 - if this block is doing straight decimation with no filtering as it appears to be doing, shouldn't we have that functionality for all input types?  Seem like it would be inconsistent in the block set to have a signal processing function specific to one data type.  So - could this be just a decimator block that takes in size_t for the sizeof_input_items?

Also, could you just use keep 1 in N?

gr-filter/include/gnuradio/filter/ichar_decimator.h
```diff
@@ -0,0 +1,41 @@
...
+ * This block provides interleaved char decimation without
+ * first having to convert to a type supported by a resampler.
+ *
+ */
+class FILTER_API ichar_decimator : virtual public gr::sync_decimator
+{
+public:
+    typedef boost::shared_ptr<ichar_decimator> sptr;
+
+
```
michaelld: whitespace

gr-filter/lib/ichar_decimator_impl.cc
```diff
@@ -0,0 +1,71 @@
...
+ichar_decimator_impl::ichar_decimator_impl(int decimation)
+    : gr::sync_decimator("ichar_decimator",
+                         gr::io_signature::make(1, 1, sizeof(char)),
+                         gr::io_signature::make(1, 1, sizeof(char)),
+                         decimation)
+{
+    d_doubledecimation = 2 * decimation;
+
+    // Make sure we work with pairs of bytes (I and Q as byte)
+    gr::block::set_output_multiple(2);
```
michaelld: should this be `d_doubledecimation` instead, to make sure in `work` that there are an integer number of decimations? Otherwise I believe that there could be glitches where more or fewer than `decimation` complex<char> are dropped.

michaelld:
I'm wondering the same thing as @mormj  ... this seems to be a "keep 1 in N" block ... pre-filtering required ... so what makes this block special? I do like the GRC scripts & interface!

ghostop14:
Hi @michaelld and @mormj, fair questions.  A little history may help.  But first, @michaelld in answer to the set_output_items, I think 2 is still the correct answer.  If you use doubledecimation as the multiplier, increased decimation would increase output multiples.  Because the input and output counters are tracked separately and the actual number of outputs is returned from work rather than returning noutput_items, no samples should be missed (any unused should be sent in on the next work call).

For the justification for the block, a little history: The complex/ichar scaling (#3208), this block, and a good bit of refactoring in grnet that I'll have time to focus on a PR for the first week in March all came out of work done at the ATA hackathon last year.  The metric was 104.8576 MCS/s, 2 feeds: H and V polarization for a total of about 3.354 Gbps of streaming ichar data from the beamformer.  So every drop of performance counted.  Folks doing 10 MCSPS wouldn't notice it but at those speeds it mattered to not drop packets.  #3204 allowed for scaling the voltage-based ichar inputs, and this block saved the memcpy overhead of having to go stream-to-vector:keep 1-in-n:vector-to-stream which necessitated 2 extra memcpy's, 2 extra blocks/threads, and a tag loop.  So this block does it at the highest performance possible in a single loop to keep up with the stream.  I thought about enhancing the rational resampler which doesn't currently support ichar or byte, but the fact that it also takes filter taps felt like it wasn't the right spot for this functionality either and that's what drove this to be a new/unique block.  Right now the gr-ata flowgraphs require gr-grnet, but with these 2 PR's then the integration of the net blocks in March, gr-ata would be native GR 3.9.

Thoughts?

michaelld:
> But first, @michaelld in answer to the set_output_items, I think 2 is still the correct answer. If you use doubledecimation as the multiplier, increased decimation would increase output multiples. Because the input and output counters are tracked separately and the actual number of outputs is returned from work rather than returning noutput_items, no samples should be missed (any unused should be sent in on the next work call).

After reviewing the GR classes `sync_decimator` and `sync_block`, I agree with you here. I hadn't put all of the pieces together. Well done!

michaelld:
@ghostop14 so the backstory is: this block -is- keep 1 in N, but is separate for performance reasons. hmmm ... no tags here ... should tags be considered? maybe an option that the user can decline? You could consider rewriting `keep_one_in_n_impl` to be more efficient ... it's a nice academic implementation, which could be a lot faster!

michaelld:
Also wondering if there should be `ishort` & `iint` versions ...

ghostop14:
I think it would be pretty easy to add ishort and iint versions.  Could be nice coming out of a USRP block.  Maybe I rename it IVal Decimator and give it drop-options for ishort, iint, and ichar.   On the tags, I always hate the extra get_tags and additional for loop when performance is key, but I think we could do that with a block configuration option and just if-check it in the work() if you think folks would really want tag capabilities too?

ghostop14:
@michaelld, renamed it to ival decimator and added support for ishorts in the same way as a drop-list option.  I initially added iint but it didn't look like we had other blocks that supported Complex->int / int->complex in the same way, just ichar and ishort, so I removed it.

I don't know the answer to this so this is a question.... looking at Keep-1-in-n it mentions it only has to manually deal with tags since it sets a TPP_DONT policy.  Since we're not doing that here (and I think this is the case), tags are automatically taken care of and we don't need to manually do it in the work function since we're not setting that policy?  Is that accurate?

michaelld:
> @michaelld, renamed it to ival decimator and added support for ishorts in the same way as a drop-list option. I initially added iint but it didn't look like we had other blocks that supported Complex->int / int->complex in the same way, just ichar and ishort, so I removed it.

Sounds good.

> I don't know the answer to this so this is a question.... looking at Keep-1-in-n it mentions it only has to manually deal with tags since it sets a TPP_DONT policy. Since we're not doing that here (and I think this is the case), tags are automatically taken care of and we don't need to manually do it in the work function since we're not setting that policy? Is that accurate?

Hmm ... I'll need to look at the core code that handles tags. I'm guessing there's a reason why the K1inN does tags the way it does ... just a guess though until I have hard facts from a code review.

michaelld:
> On the tags, I always hate the extra get_tags and additional for loop when performance is key, but I think we could do that with a block configuration option and just if-check it in the work() if you think folks would really want tag capabilities too?

I do think folks expect and also want tags to propagate properly through most blocks. Hence, I could see this as an option: propagate tags properly or don't at all. Default would be don't.

ghostop14:
Sounds good.  Let me know what you think from the core code if they propagate automatically all things being equal.  If its not handled automatically I’ll add it.

gr-filter/lib/ival_decimator_impl.cc
```diff
@@ -0,0 +1,98 @@
...
+ival_decimator_impl::ival_decimator_impl(int decimation, int data_type, int data_size)
+    : gr::sync_decimator("ichar_decimator",
+                         gr::io_signature::make(1, 1, data_size),
+                         gr::io_signature::make(1, 1, data_size),
+                         decimation),
+      d_data_type(data_type)
+{
+    d_doubledecimation = 2 * decimation;
+
+    if ((d_data_type < 1) || d_data_type > 2)
```
michaelld: need a check for `data_size` too

michaelld: can you just use `data_type` for both it and `data_size` ... since they happen to overlap exactly?

gr-filter/include/gnuradio/filter/ival_decimator.h
```diff
@@ -0,0 +1,41 @@
...
+ * This block provides interleaved char decimation without
+ * first having to convert to a type supported by a resampler.
+ *
+ */
+class FILTER_API ival_decimator : virtual public gr::sync_decimator
+{
+public:
+    typedef boost::shared_ptr<ival_decimator> sptr;
+
+
```
michaelld: Extract space

gr-filter/grc/filter.tree.yml
```diff
@@ -26,6 +26,7 @@
   - pfb_arb_resampler_xxx
   - rational_resampler_xxx
   - rational_resampler_base_xxx
+  - ival_decim
```
michaelld: Is this supposed to be the block name or the file name? I'm still not sure ...

ghostop14:
@michaeld: Fixed the header and added size checks.  I don't think I want to just rely on size for data type checks.  If in the future we decide that int / float makes sense we can run into scenarios where data sizes are the same for multiple data types.  For the tree.yml, it's a good question.  I made them all match just to be safe.

gr-filter/lib/ival_decimator_impl.cc
```diff
@@ -0,0 +1,110 @@
...
+    return gnuradio::get_initial_sptr(
+        new ival_decimator_impl(decimation, data_type, data_size));
+}
+
+/*
+ * The private constructor
+ */
+ival_decimator_impl::ival_decimator_impl(int decimation, int data_type, int data_size)
+    : gr::sync_decimator("ival_decimator",
+                         gr::io_signature::make(1, 1, data_size),
```
michaelld: Since `data_size` and `data_type` -are- directly related, let's get rid of 1 of them. either one will work alone; using both is redundant and prone to user error.

gr-filter/include/gnuradio/filter/ival_decimator.h
```diff
@@ -0,0 +1,40 @@
...
+namespace gr {
+namespace filter {
+
+/*!
+ * \brief Filter-Delay Combination Block.
+ * \ingroup resamplers_blk
+ *
+ * \details
+ * This block provides interleaved char decimation without
```
michaelld: please correct this comment, meaning "`char` or `short`" ... or something that represents the new block data types

michaelld:
Asking for at least 1 more review of this PR, since it's about ready to go.

ghostop14:
Comment fixed.

gr-filter/lib/ival_decimator_impl.cc
mormj: Why wouldn't we just use data_size so it opens the block up to all data types.  Not a big fan of the new #defines.  If we go with data_type, then we should probably be able to reuse the gr enums that are exposed through swig.  With just data_size (like other blocks), could the if statement be removed and void *'s be std::copy'd two elements at a time?

mormj:
> 
> 
> Also wondering if there should be `ishort` & `iint` versions ...

I'm still hoping there's a way to do it without new #defines and so it isn't an if or switch for every type.

Also - can the same performance improvements be gained from non-interleaved types, so that this just becomes a vanilla decimator block?  gr_complex, float, etc.  Or will these just have the net effect of keep 1 in N

gr-filter/grc/filter_ival_decimator.block.yml
```diff
@@ -0,0 +1,34 @@
+id: ival_decimator
+label: IValue Stream Decimator
```
mormj: If this is strictly for i values - perhaps the label should read "Interleaved Stream Decimator" to spell it out

gr-filter/grc/filter_ival_decimator.block.yml
```diff
@@ -0,0 +1,34 @@
...
+
+parameters:
+-   id: datatype
+    label: Input Type
+    dtype: enum
+    options: [byte, short]
+    option_attributes:
+        datatype: ['1', '2']
```
mormj: As noted in the other comments, I'm not a fan of the hardcoded mappings here - would be nice to use existing enums, or just use the sizeof

ghostop14:
Okay folks, a couple updates just pushed and some perspectives too on design choices I made.  So I changed the #defines to constexpr's.  I may be old-school but in this case I think #defines were fine.  But I get the impression gr is trying to clean them out of the code, so they're constexpr's now.  I changed the name to "Interleaved Stream Operator" as requested as well.  I'm concerned it's long on a flowgraph which is why I had kept it short (I hate blocks that take up a lot of horizontal space just for the name.

In terms of the size versus type, trust me, this is easier to scale later.  And because data size isn't unique it can cause problems and unintended redo's later.  I personally prefer the strongly typed switch case blocks in work over memcpy's.  Same scenario as moving from #defines to constexpr's.  Take advantage of the strong types.  It's less error-prone than the memcpy's.  Just my design preferences, but I'm happy to consider better ways.  

Just FYI, I'm going to be short on time to work on these till the beginning of March after today so I'd like to try to wrap them up so they don't sit if possible.

michaelld:
Thanks for fixing the comment. I'm fine with the PR as is. Unless someone is completely against the way the `data_type` argument is set, I'd like to get this merged & then we can revisit it down the road if/as necessary. The performance will be a -lot- better than K-1-i-N, though not as robust overall -- a typical trade-off we make in programming!

gr-filter/lib/ival_decimator_impl.cc
michaelld: Using a `data_type` allows the `switch` and then using normal C/C++ types to do the work, which will be a -lot- more efficient than using `memcpy` or whatever. That said, the reverse could be done too: use the correct `int` type for the size provided, then do the style of copying as done already. Since this block isn't -interpreting- the data, just moving the value around, this would work.

michaelld: NOTE: of course this wouldn't work for 3 byte items, but we don't use those in GR anyway. No sure how to handle 16 byte items, since there is no `int128_t` yet TTBOMK ... but everything up to 8 byte items would be doable.

michaelld:
I'd guess that many blocks could use a `data_type` concept to help speed them up! We're not going to cover that here in this PR today :)

ghostop14:
After several side conversations, I did remove the data_type in the interest of integration, so have another look.  Just for the record I still stand by the other approach as the better design for future expansion.  In my experience, choosing a non-unique identifier will come back to bite you down the road when you want to add to it.  But in the interest of getting it merged, I changed it to rely on size only.

gr-filter/lib/ival_decimator_impl.cc
```diff
@@ -0,0 +1,90 @@
...
+ */
+ival_decimator_impl::ival_decimator_impl(int decimation, int data_size)
+    : gr::sync_decimator("ival_decimator",
+                         gr::io_signature::make(1, 1, data_size),
+                         gr::io_signature::make(1, 1, data_size),
+                         decimation),
+      d_data_size(data_size)
+{
+    d_doubledecimation = 2 * decimation;
+
```
michaelld: please add a check for `data_size` to be either `sizeof(char)` or `sizeof(short)` ... or throw an error here during instantiation.

gr-filter/grc/filter_ival_decimator.block.yml
```diff
@@ -0,0 +1,34 @@
+id: ival_decimator
+label: Interleaved Stream Decimator
+
+parameters:
+-   id: datatype
+    label: Input Type
+    dtype: enum
+    options: [byte, short]
+    option_attributes:
+        datasize: [gr.sizeof_char, gr.sizeof_short]
```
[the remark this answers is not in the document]
mormj: This makes me feel much better with enums

ghostop14:
I just pushed the update to address the data validation.  Thanks for all the reviews!  I know it's time-consuming.

gr-filter/lib/ival_decimator_impl.cc
```diff
@@ -0,0 +1,93 @@
...
+ */
+ival_decimator_impl::ival_decimator_impl(int decimation, int data_size)
+    : gr::sync_decimator("ival_decimator",
+                         gr::io_signature::make(1, 1, data_size),
+                         gr::io_signature::make(1, 1, data_size),
+                         decimation),
+      d_data_size(data_size)
+{
+    if ((data_size != 1) && (data_size != 2))
+        throw std::invalid_argument("data size must be in 1 (char) or 2 (short)");
```
michaelld: yes!

michaelld:
CI passes. 2 maintainer approvals. Merging.

Issue #3188, opened 2020-02-09
Sanity for GnuradioConfigVersion.cmake


marcusmueller:
That is the file CMake executes when someone does a `find_package(Gnuradio "VERSION.VERSION")`. 

#3187 highlighted that we're not doing a good job there: Of course, an OOT module can't know whether 3.9-prerelease versions are compatible with what 3.9 will be, nor can it know whether they're compatible with 3.8.

So, a sane environment would allow the OOT author to allow "hey, I want at least GNU Radio 3.8.1.2", because they need a specific API (3.8) and they need a specific bug fix to be already present. But if there's 3.9.0.0 on the system, CMake shouldn't fail at that point. 

The solution I'd propose would be to change GnuradioConfigVersion.cmake.in to:

* set `PACKAGE_VERSION_COMPATIBLE` on
```REQUESTED MAJOR==GR MAJOR && GR VERSION >= REQUESTED VERSION```
* set `PACKAGE_VERSION_EXACT` on 
```GR VERSION == REQUESTED VERSION```
(all digits specified, not just the first three)

smunaut:
Note that the changes I was proposing (for which argilo made the PR), were _not_ meant to make it in release 3.9.

My PoV is that until 3.9 is released and is "guaranteed API" released, version checks make little sense at all since it might break at any time depending on what gets merged to master.

And so the cmake file in the "git version" of GR should be tolerant and declare itself 3.8 compatible during the dev phase, with the assumption that user of an unstable branch know what they're doing.

When 3.9 is _actually_ released and tagged (or the release candidate is), then version check should go to "strict" again and at that point the OOT peole need to go, check their OOT against 3.9 API and accept it as a supported version.

smunaut:
BTW that's actually similar to what most project do. When working on the "next version" they usually declare themselves as like 3.8.9999  and not 3.9 for that exact reason ... then only switch the actual git version to 3.9 at release time.

marcusmueller:
* You're right, no sense in using version checks against something that's not a version
* So, if you're consumer of an unreleased version of anything, you'll have to live with not being able to rely on a version check
* Therefore, it's not upstream's responsibility to try and guess what you think should be covered under compatibility. Current `master` is not compatible with 3.8, full stop. Otherwise, we wouldn't be developing it on master, but `maint-3.8`, which we dedicatedly decided *against*, for very good reason.

So, how to implement your recommendation:

[...]

PR #3218, opened 2020-02-19
cmake: declare compatibility if newer than the requested version

argilo:
Fixes https://github.com/gnuradio/gnuradio/issues/3188.

The GNU Radio 3.8 OOT Module Porting Guide recommends finding
GNU Radio using `find_package(Gnuradio "3.8" REQUIRED)`. GNU Radio's
version check is currently strict, which prevents OOT modules from
building against the master branch, which carries a version number of
3.9.

To resolve this problem, I've modified the check to indicate that the
GNU Radio version is compatible with the requested version so long as
it is greater than or equal to the requested version.

If an OOT module knows that it is incompatible with GNU Radio 3.9,
it can use `find_package(Gnuradio "3.8" EXACT REQUIRED)` to prevent
building with versions other than 3.8.

The version check was copied from the CMake wiki:
https://gitlab.kitware.com/cmake/community/-/wikis/doc/tutorials/How-to-create-a-ProjectConfig.cmake-file#the-foobarfoobarconfigversioncmakein-file


marcusmueller:
hm, we had arguments for what was there before at the Feb hackfest, I THINK, but honestly, if this has a clear case for making the ecosystem work...

PR #3224, opened 2020-02-20
digital: restructure equalizers into linear/dfe + adaptive algorithm

mormj:
This follows the FOSDEM talk:
https://fosdem.org/2020/schedule/event/fsr_hannel_equalization_using_gnu_radio/

The main changes are:
1) Separate the equalization structure from the adaptive algorithm
-- e.g. the previous lms_dd_equalizer becomes linear_equalizer with an argument of an adaptive_algorithm_lms object
2) Add training based equalization
-- In this way we can specify a preamble and that used to more quickly adapt the taps
-- After the data-aided training, the equalizer can be switched into decision directed mode

Some things from the out-of-tree implementation have not been added in yet, namely RLS because it requires (currently) a libarmadillo dependency.  Refactoring this to do the matrix math with existing libraries will be a fairly easy change, but left it off for now.


gr-digital/lib/meas_evm_cc_impl.cc
```diff
@@ -0,0 +1,78 @@
...
+      d_cons(cons),
+      d_meas_type(meas_type)
+{
+
+    d_cons_points = d_cons->points();
+
+    double sum = 0.0;
+    int N = d_cons_points.size();
+    for (int i = 0; i < N; i++) {
+        sum += pow(real(d_cons_points[i]), 2.0) + pow(imag(d_cons_points[i]), 2.0);
```
marcusmueller: I was about to write "um, `pow(x, 2.0)` sounds like a very expensive way of writing `x*x`: Modern GCC [doesn't seem to care](https://gcc.godbolt.org/z/Ranv-p). The only difference is the order of additions?! nice.

tbh, this is again not really asking for improvement, but `std::norm` would keep the calculation float for as long as it's not detrimental, so should be faster. Also, VOLK, maybe?

marcusmueller: But really, this is a one-time thing in the constructor; couldn't care less.

gr-digital/lib/meas_evm_cc_impl.cc
```diff
@@ -0,0 +1,78 @@
...
+    }
+    sum /= (double)N;
+    d_cons_mag_sq = sum;
+    d_cons_mag = sqrt(sum);
+}
+
+double meas_evm_cc_impl::mag(gr_complex x)
```
marcusmueller: `std::abs` does exactly this, so this feels kind of duplicate, unless we really need the squares and square root to happen on doubles instead of floats.

marcusmueller: generally, are we using this?

gr-digital/lib/meas_evm_cc_impl.cc
```diff
@@ -0,0 +1,78 @@
...
+{
+    const gr_complex* in = (const gr_complex*)input_items[0];
+    float* out = (float*)output_items[0];
+
+    // Compare incoming symbols to the constellation decision points
+    for (int s = 0; s < noutput_items; s++) {
+        gr_complex u_n = in[s];
+        gr_complex decision;
+        d_cons->map_to_points(d_cons->decision_maker(&u_n), &decision);
+        double err_mag = std::abs(decision - u_n);
```
marcusmueller: why the conversion from single to double precision here?

gr-digital/python/digital/qa_decision_feedback_equalizer.py
```diff
@@ -0,0 +1,157 @@
+#!/usr/bin/env python
```
marcusmueller: I love unit tests!

gr-digital/python/digital/qa_decision_feedback_equalizer.py
```diff
@@ -0,0 +1,157 @@
...
+                
+        return out
+
+    def map_symbols_to_constellation(self, symbols, cons):
+        l = list(map(lambda x: cons.points()[x], symbols))
+        return l
+
+
+    def setUp(self):
+        random.seed(987654)
```
marcusmueller: This likely won't ever become a problem here, but beware: unit tests are executed in parallel (though at this point in different python interpreters), and thus the state of the global `random` might change unexpectedly! (this is more relevant for python blocks, though.) 

A way around this could be `rng = gr.random(987654);` here and and instead of getrandbits `rng.ran_int() & 0xFF` or so below :) (MEMO TO SELF: might want to switch `ran_int` from std:: to gr::xoroshiro at some point for speed reasons)

gr-digital/swig/equalizers.i
```diff
@@ -0,0 +1,20 @@
+/* -*- c++ -*- */
```
marcusmueller: Awesome! thanks for doing this; it tends to take me hours to figure swig out.

mormj:
Thanks for the thorough feedback @marcusmueller - much appreciated.  I'll get to work on cleaning this up

gr-digital/lib/meas_evm_cc_impl.cc
mormj: most likely sloppiness

gr-digital/lib/decision_feedback_equalizer_impl.cc
```diff
@@ -0,0 +1,269 @@
...
+    }
+
+    unsigned long int nread = nitems_read(0);
+    vector<tag_t> tags;
+    vector<unsigned int> training_start_samples;
+    get_tags_in_window(
+        tags, 0, 0, noutput_items * decimation(), pmt::intern(d_training_start_tag));
+    for (unsigned i = 0; i < tags.size(); i++) {
+        training_start_samples.push_back(tags[i].offset - nread);
+    }
```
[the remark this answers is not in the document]
mormj: true!  old habits die hard - still getting used to c++11 niceties :)

gr-digital/include/gnuradio/digital/decision_feedback_equalizer.h
```diff
@@ -0,0 +1,85 @@
...
+         int sps,
+         adaptive_algorithm_sptr alg,
+         bool adapt_after_training = true,
+         std::vector<gr_complex> training_sequence = std::vector<gr_complex>(),
+         const std::string& training_start_tag = "");
+
+    virtual void set_taps(const std::vector<gr_complex>& taps) = 0;
+    virtual std::vector<gr_complex> taps() const = 0;
+
+    // Public "work" function - allow signal processing to be called externally
```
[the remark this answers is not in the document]
mormj: Fair question.  There a couple of reasons for structuring the block this way:
1. The signal processing can be called outside a gnu radio flowgraph.  i could simply instantiate an equalizer object in python or c++ and call the equalize method directly.  state would be maintained, but doesn't necessarily have to be performed in a streaming context.  Some blocks have done this with the "kernel" namespaces.  
2. Makes it easier to also add pdu or tagged stream versions wrapping the same signal processing

mormj:
@jdemel  - thanks for the review.  I will incorporate the changes shortly.

gr-digital/include/gnuradio/digital/decision_feedback_equalizer.h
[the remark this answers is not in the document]
jdemel: OK. That's a good reason. Though, I'd recommend to go with the kernel structure. This makes it actually usable outside GR. Otherwise, you'd need all the block infrastructure if just want to use the kernel. That would be a clean way to separate functionality from the interface.

gr-digital/include/gnuradio/digital/adaptive_algorithm_nlms.h
```diff
@@ -0,0 +1,78 @@
...
+    void update_taps(gr_complex* taps,
+                     const gr_complex* in,
+                     const gr_complex error,
+                     const gr_complex decision,
+                     unsigned int num_taps)
+    {
+
+        gr_complex dp;
+        volk_32fc_x2_conjugate_dot_prod_32fc(&dp, in, in, num_taps);
+        float magsq = abs(dp);
```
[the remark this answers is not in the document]
mormj: You are correct abs is not necessary here - since the dot product does the squaring and x*x' is a positive number, I think we can actually get away with real() here.

gr-digital/include/gnuradio/digital/adaptive_algorithm_cma.h
```diff
@@ -0,0 +1,92 @@
...
+
+    float d_step_size;
+    float d_modulus;
+
+protected:
+    adaptive_algorithm_cma(constellation_sptr cons, float step_size, float modulus)
+        : adaptive_algorithm(adaptive_algorithm_t::CMA, cons)
+    {
+        this->d_step_size = step_size;
+        this->d_modulus = modulus;
```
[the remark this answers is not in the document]
mormj: These adaptive algorithms are doing so little, it seemed easier just to leave everything in the header and limit the additional files and cmake.  Moved the members to the initializer list

gr-digital/include/gnuradio/digital/adaptive_algorithm_cma.h
```diff
@@ -0,0 +1,92 @@
...
+        conj_vector = (gr_complex*)volk_malloc(sizeof(gr_complex) * num_taps, alignment);
+
+        gr_complex err_x_mu = -d_step_size * error;
+
+        volk_32fc_conjugate_32fc(conj_vector, in, num_taps);
+        volk_32fc_s32fc_multiply_32fc(prod_vector, conj_vector, err_x_mu, num_taps);
+        volk_32fc_x2_add_32fc(taps, taps, prod_vector, num_taps);
+
+        volk_free(prod_vector);
+        volk_free(conj_vector);
```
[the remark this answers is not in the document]
mormj: thanks for the suggestion re: volk 2.1 - switching over to that now!

gr-digital/include/gnuradio/digital/adaptive_algorithm_lms.h
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    void update_tap(gr_complex& tap,
+                    const gr_complex& u_n,
+                    const gr_complex err,
+                    const gr_complex decision)
+    {
+
+        tap = conj(conj(tap) + d_step_size * u_n * conj(err));
+    }
```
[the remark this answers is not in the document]
mormj: This was a copy over from the previous equalizer code - agreed the return value is cleaner.

gr-digital/include/gnuradio/digital/decision_feedback_equalizer.h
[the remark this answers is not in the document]
mormj: I see the benefit of the kernel namespace outside of GR - for this implementation, the equalize() function still relies on tags which are GR specific so not sure if the benefits of the kernel namespace would be fully realized.  Having the equalize() function broken off does bring some benefits - I'd like to leave it as is if that is ok.

gr-digital/include/gnuradio/digital/adaptive_algorithm_nlms.h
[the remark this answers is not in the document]
rear1019: Yes, you’re right. I did saw that the required dot product with a real result is performed. But then my brain failed and only saw the temptingly wrong looking ``std::abs()``…

gr-digital/lib/linear_equalizer_impl.cc
```diff
@@ -0,0 +1,232 @@
...
+    }
+
+    return nout;
+}
+
+int linear_equalizer_impl::work(int noutput_items,
+                                gr_vector_const_void_star& input_items,
+                                gr_vector_void_star& output_items)
+{
+    const gr_complex* in = (const gr_complex*)input_items[0];
```
ThomasHabets, line 191: Consider using C++ casts.

```c++
auto in = static_cast<const gr_complex*>(input_items[0]);
```

gr-digital/lib/linear_equalizer_impl.cc
```diff
@@ -0,0 +1,232 @@
...
+        d_updated = false;
+        return 0; // history requirements may have changed.
+    }
+
+    unsigned long int nread = nitems_read(0);
+    vector<tag_t> tags;
+    vector<unsigned int> training_start_samples;
+    get_tags_in_window(
+        tags, 0, 0, noutput_items * decimation(), pmt::intern(d_training_start_tag));
+    for (auto& tag : tags) {
```
ThomasHabets, line 217: `const auto& tag : tags`

gr-digital/include/gnuradio/digital/adaptive_algorithm_cma.h
```diff
@@ -0,0 +1,87 @@
...
+        volk_32fc_conjugate_32fc(conj_vector.data(), in, num_taps);
+        volk_32fc_s32fc_multiply_32fc(
+            prod_vector.data(), conj_vector.data(), err_x_mu, num_taps);
+        volk_32fc_x2_add_32fc(taps, taps, prod_vector.data(), num_taps);
+    }
+
+    gr_complex update_tap(const gr_complex tap,
+                          const gr_complex& u_n,
+                          const gr_complex err,
+                          const gr_complex decision)
```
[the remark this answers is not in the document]
mormj: this is a virtual function used by all algorithms, so even though the cma flavor doesn't use decision it is used in the other ones

Issue #3217, opened 2020-02-19
ZMQ Message Source blocks fail to receive data from non-GRC sources


duggabe:
In trying to use the ZMQ PULL Message Source block to communicate with a separate, non-GRC Python process, the PULL Message Source block fails to receive any data. A series of tests have been performed to validate the condition. Similar tests were performed using REQ/REP and PUB/SUB sockets. They also failed. Note that only interactions between GRC and non-GRC sources fail.

platform: Raspberry Pi 3B+
OS: 2020-02-13-raspbian-buster-full
GR: 3.8.1.1-rc1
GRC: 3.8.1.0
Current libzmq version is 4.3.1
Current  pyzmq version is 18.1.1

The following files are used in the tests below:
msg_test.py -    https://pastebin.com/wabF6Tsz
msg_test2.py -  https://pastebin.com/e3TE7dcb
zmqPULL.py -   https://pastebin.com/p93GGbtu
zmqPUSH.py -  https://pastebin.com/hdUkhysV

Test case 1: GRC from GRC
- terminal A:
-   execute 'python3 -u msg_test.py'
- terminal B:
-   execute 'python3 -u msg_test2.py'
- terminal A displays "TEST" once per second.
- stop both programs by clicking the X in the GUI windows

Test case 2: Python from Python
- terminal A:
-   execute 'python3 zmqPULL.py'
- terminal B:
-   execute 'python3 zmqPUSH.py'
- terminal A displays the date and time once per second.
- stop both programs with Control-C

Test case 3: GRC from Python
- terminal A:
-   execute 'python3 -u msg_test.py'
- terminal B:
-   execute 'python3 zmqPUSH.py'
- terminal A *should* display the date and time once per second, but it FAILS: there is no output.
- stop terminal A by clicking the X in the GUI window
- stop terminal B with Control-C

talih0:
@duggabe 
The pull source block connects by default (https://github.com/gnuradio/gnuradio/blob/master/gr-zeromq/lib/pull_msg_source_impl.cc#L48)

Your python script is also connecting on the push side. 
If you change your python script to bind on the push side, it will work.

talih0:
@duggabe 
Also, If the zmq pull source is setup to read gr_complex, it is expecting to receive sizeof(gr_complex)*num_symbols bytes. 
This is likely not to be satisfied if you send a general string, and gnuradio will throw an exception.

velichkov:
Hi @duggabe,

[...]

PR #3230, opened 2020-02-22
zmq: Handle the PMT exceptions

velichkov:
Handle the PMT exceptions when deserializing the received ZMQ messages,
log an error message and continue the execution.

Previously the unhandled exceptions were aborting the process: #3217

terminate called after throwing an instance of 'pmt::exception'
  what():  pmt::deserialize: malformed input stream, tag value = : 50
Aborted (core dumped)

### DIFF
--- a/gr-zeromq/lib/pull_msg_source_impl.cc
+++ b/gr-zeromq/lib/pull_msg_source_impl.cc
@@ -91,8 +91,12 @@ void pull_msg_source_impl::readloop()
 
             std::string buf(static_cast<char*>(msg.data()), msg.size());
             std::stringbuf sb(buf);
-            pmt::pmt_t m = pmt::deserialize(sb);
-            message_port_pub(d_port, m);
+            try {
+                pmt::pmt_t m = pmt::deserialize(sb);
+                message_port_pub(d_port, m);
+            } catch (pmt::exception& e) {
+                GR_LOG_ERROR(d_logger, std::string("Invalid PMT message: ") + e.what());
+            }
 
         } else {
             boost::this_thread::sleep(boost::posix_time::microseconds(100));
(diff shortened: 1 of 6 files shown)


gr-zeromq/lib/pull_msg_source_impl.cc
```diff
@@ -91,8 +91,12 @@ void pull_msg_source_impl::readloop()
 
             std::string buf(static_cast<char*>(msg.data()), msg.size());
             std::stringbuf sb(buf);
-            pmt::pmt_t m = pmt::deserialize(sb);
-            message_port_pub(d_port, m);
+            try {
+                pmt::pmt_t m = pmt::deserialize(sb);
```
mormj: I like this change, but should the try be isolated to the pmt::deserialize - i.e. will message_port_pub throw a pmt exception?

velichkov: The `message_port_pub` does not throw `pmt::exception` but I think it's better to publish/post only if the PMT message was deserialized correctly and that's why I put `message_port_pub` inside the try block.

I can rewrite it as
```cpp
            pmt::pmt_t m;
            try {
                m = pmt::deserialize(sb);
            } catch (pmt::exception& e) {
                GR_LOG_ERROR(d_logger, std::string("Invalid PMT message: ") + e.what());
                continue;
            }
            message_port_pub(d_port, m);
```
but I found this less readable. Let me know if you like more this variant or if you have another idea and I'm going to change it.

mormj: True - I agree with you there.  The original change is better

mormj:
Non-controversial change and new QA!  Merging

PR #3232, opened 2020-02-22
fix(filter_design): Use public fft interface of numpy

schneider42:
At least on my machine numpy changed from `pocketfft` to `_pocketfft`. The public interface works just as well though.


schneider42:
Superseded by https://github.com/gnuradio/gnuradio/pull/3291

Issue #3220, opened 2020-02-20
Missing deconstructor in ZMQ_PUSH_Sink


Dur0k:
Due to a missing deconstructror in gnuradio/gr-zeromq/lib/push_sink_impl.cc the used port is not released after "deleting" block object with pythons del function.

After adding 
```
push_sink_impl::~push_sink_impl()
{
}
```
to  gnuradio/gr-zeromq/lib/push_sink_impl.cc and
```
~push_sink_impl();
```
to gnuradio/gr-zeromq/lib/push_sink_impl.h the port is correctly released and can be bound again.

This occurred while trying to disconnect, delete and afterwards create a new ZMQ_PUSH_Sink on the same port. The flowgraph was started externally in python via class object of the flowgraph.

talih0:
The compiler adds a default destructor automatically if the user does not create one.

You should probably instead test setting linger time to 0 as is done in the push_msg_sink code:

https://github.com/gnuradio/gnuradio/blob/master/gr-zeromq/lib/push_msg_sink_impl.cc#L44

michaelld:
> The compiler adds a default destructor automatically if the user does not create one.

This statement isn't _necessarily_ true or correct. In some recent C++ standards, if a destructor isn't supplied the compiler might or not create a virtual default. This case will generally result in a compiler warning during build, but might result in a build failure. If the code does compile, then it stands a decent chance of runtime failure. In my testing, such issues happen more with Clang and GCC; no idea why.

In this specific case, adding in the virtual destructor is the wise thing to do, if nothing else to avoid such situations in the future (e.g., moving to C++14 or C++17 or C++20).

michaelld:
@Dur0k do you feel comfortable creating a PR for this issue? If you do, please check _all_ of the ZMQ sink/source blocks in gr-zeromq & fix any that need this tweak.

marcusmueller:
there's a virtual destructor in `gr::zeromq::base_impl`, so that must get called. Which at least shuts down the context, closes the socket.

However, usually dtor bubbling happens at the end of a destructor call. 

So, will need to reproduce to make sure I'm understanding which dtors really get called here.

PR #3234, opened 2020-02-26
gr-zeromq: Add missing deconstructors

Dur0k:
Closes: #3220 

Some zmq sink and source blocks are missing deconstructors, which results in ports not being freed after deleting block object. This PR is created after issue #3220.

### DIFF
--- a/gr-zeromq/lib/pub_sink_impl.h
+++ b/gr-zeromq/lib/pub_sink_impl.h
@@ -28,6 +28,8 @@ public:
                   bool pass_tags,
                   int hwm);
 
+    ~pub_sink_impl();
+
     int work(int noutput_items,
              gr_vector_const_void_star& input_items,
              gr_vector_void_star& output_items) override;
--- a/gr-zeromq/lib/pub_sink_impl.cc
+++ b/gr-zeromq/lib/pub_sink_impl.cc
@@ -36,6 +36,8 @@ pub_sink_impl::pub_sink_impl(
     /* All is delegated */
 }
 
+pub_sink_impl::~pub_sink_impl() {}
+
 int pub_sink_impl::work(int noutput_items,
                         gr_vector_const_void_star& input_items,
                         gr_vector_void_star& output_items)
--- a/gr-zeromq/lib/pull_source_impl.cc
+++ b/gr-zeromq/lib/pull_source_impl.cc
@@ -36,6 +36,8 @@ pull_source_impl::pull_source_impl(
     /* All is delegated */
 }
 
+pull_source_impl::~pull_source_impl() {}
+
 int pull_source_impl::work(int noutput_items,
                            gr_vector_const_void_star& input_items,
                            gr_vector_void_star& output_items)
--- a/gr-zeromq/lib/pull_source_impl.h
+++ b/gr-zeromq/lib/pull_source_impl.h
@@ -28,6 +28,8 @@ public:
                      bool pass_tags,
                      int hwm);
 
+    ~pull_source_impl();
+
     int work(int noutput_items,
              gr_vector_const_void_star& input_items,
              gr_vector_void_star& output_items) override;
--- a/gr-zeromq/lib/push_sink_impl.cc
+++ b/gr-zeromq/lib/push_sink_impl.cc
@@ -36,6 +36,8 @@ push_sink_impl::push_sink_impl(
     /* All is delegated */
 }
 
+push_sink_impl::~push_sink_impl() {}
+
 int push_sink_impl::work(int noutput_items,
                          gr_vector_const_void_star& input_items,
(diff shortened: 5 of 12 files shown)


gr-zeromq/lib/pub_sink_impl.h
```diff
@@ -28,6 +28,8 @@ class pub_sink_impl : public pub_sink, public base_sink_impl
                   bool pass_tags,
                   int hwm);
 
+    ~pub_sink_impl();
```
michaelld: Since these methods will be empty, will putting them entirely here work for the issue at hand? If so, let's do that ... the code is simpler and cleaner.

michaelld:
I think if the destructors are combined, then we don't need a license assignment (CLA) for this.

@Dur0k do you have hopes of contributing more to GR code-wise? If so, you'll need the CLA, which does take a little time to deal with. Let us know here & we'll get the ball rolling for you.

gr-zeromq/lib/pub_sink_impl.h
hcab14: Shouldn't this destructor be virtual? Otherwise it will override the destructors of the base classes.

Dur0k:
Hi, sorry for the late reply. Yes this is my first PR. If I run into more issues and am able to fix them, then I would definetly like to contribute more.
After some testing I found out that without the destructor in .cc files everything succesfully compiles, but I could no longer import the zeromq library from gnuradio in python:  
```
  File "/home/durok/prefix/gr39/lib/python3/dist-packages/gnuradio/zeromq/zeromq_swig.py", line 14, in swig_import_helper
    return importlib.import_module(mname)
  File "/usr/lib/python3.6/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 994, in _gcd_import
  File "<frozen importlib._bootstrap>", line 971, in _find_and_load
  File "<frozen importlib._bootstrap>", line 955, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 658, in _load_unlocked
  File "<frozen importlib._bootstrap>", line 571, in module_from_spec
  File "<frozen importlib._bootstrap_external>", line 922, in create_module
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
ImportError: /home/durok/prefix/gr39/lib/libgnuradio-zeromq.so.3.8.1: undefined symbol: _ZTTN2gr6zeromq14push_sink_implE

  File "<stdin>", line 1, in <module>
  File "/home/durok/prefix/gr39/lib/python3/dist-packages/gnuradio/zeromq/__init__.py", line 35, in <module>
    from .zeromq_swig import *
  File "/home/durok/prefix/gr39/lib/python3/dist-packages/gnuradio/zeromq/zeromq_swig.py", line 17, in <module>
    _zeromq_swig = swig_import_helper()
  File "/home/durok/prefix/gr39/lib/python3/dist-packages/gnuradio/zeromq/zeromq_swig.py", line 16, in swig_import_helper
    return importlib.import_module('_zeromq_swig')
  File "/usr/lib/python3.6/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
ModuleNotFoundError: No module named '_zeromq_swig'
```
How can I try to debug this issue?

gr-zeromq/lib/pub_sink_impl.h
Dur0k: My understanding of OOP, classes and destructors is quite limited, so I just oriented myself on other zeromq blocks with existing destructors.

ThomasHabets:
Isn't it just the base class that needs a virtual destructor?

Just a one-line `virtual ~base_sink_impl() {}` ?

ThomasHabets:
And yeah they should be virtual. Without the word "virtual" these added destructors are no-ops, no?

michaelld:
> Isn't it just the base class that needs a virtual destructor?
> 
> Just a one-line `virtual ~base_sink_impl() {}` (and for source)?

That might work. Worth a try! The base class sink -must- be virtual! All of the inheriting class' destructors might need to be declared and virtual ... that's worth testing one way or the other! I'd guess that some compiles will bork if these aren't added ... and, honestly, it won't hurt to have them ... most modern compilers will optimize them out.

ThomasHabets:
Do you mean depend on LTO (probably won't work for dynamically loaded), or inlining the destructor in the header file?

michaelld:
Just adding in the virtual destructor in the header. Doesn't need to be inline specifically. Just there.

mbr0wn:
Since there are unresolved issues, and also lack of activity, I'm closing this PR. Please reopen when you think they've been resolved. We are very happy about new contributors!

Issue #3120, opened 2020-01-25
clang-format option `SortIncludes` potentially breaks code


sbmueller:
The `SortIncludes` option in the .clang-format file potentially breaks code. Includes of each group (e.g. `^[a-Z]*>$`) are sorted alphabetically. This is dangerous since the include order should be determined by functionality, not by name.

Example of what this could break: Qt uses the keyword `slots`. However, Python also uses this name for other purposes [1]. This makes it necessary to include Python before Qt, or else the `slots` statements in Python will be treated as Qt keywords, which leads to build failure.
I reckon only by coincidence, this has not yet caused errors in gr-qtgui.

One solution could be to disable the `SortIncludes` option, which is highly controversial [2].

[1] https://stackoverflow.com/questions/23068700/embedding-python3-in-qt-5
[2] https://news.ycombinator.com/item?id=21580829

ValZapod:
@sbmueller It could be even worse. https://travisdowns.github.io/blog/2019/11/19/toupper.html

sbmueller:
To add some more thought to this: Disabling this option won't result in different `clang-format` outputs since include order will remain untouched. All formatted source files should remain unmodified when re-running `clang-format`.

sbmueller:
Related #1664

sbmueller:
Won't be changed, discussion can be found in #3238 .

PR #3238, opened 2020-03-02
clang-format: disable `SortIncludes` option

sbmueller:
The `SortIncludes` option in clang-format potentially breaks code.
Includes of each group (e.g. ^[a-Z]*>$) are sorted alphabetically.  This
is dangerous since the include order should be determined by
functionality, not by name.
Example of what this could break: Qt uses the keyword slots. However,
Python also uses this name for other purposes. This makes it necessary
to include Python before Qt, or else the slots statements in Python will
be treated as Qt keywords, which leads to build failure.
Solves #3120 .

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -87,7 +87,7 @@ PenaltyExcessCharacter: 1000000
 PenaltyReturnTypeOnItsOwnLine: 60
 PointerAlignment: Left
 ReflowComments:  true
-SortIncludes:    true
+SortIncludes:    false
 SpaceAfterCStyleCast: false
 SpaceAfterTemplateKeyword: true
 SpaceBeforeAssignmentOperators: true


mbr0wn:
I would like to not merge this. Rationale:

1. In general, touching `.clang-format` should be considered very, very carefully. This particular PR would not break anything, because we're removing a rule, not changing or adding one, but changes to said file should be reserved for extreme cases. But let's consider this file sacrosanct unless it's really, really breaking stuff.

2. clang-format doesn't just sort alphabetically, it sorts according to specific rules. We could even amend the sorting rules to put Python before Qt, but see 1.

3. Could also be 2a: clang-format doesn't sort all the includes alphabetically, and you can modify its behaviour in-source. For example, you can put comments or newlines between includes to avoid reordering. That means if you know that includes have to be in a certain order, then you can still do that without touching `.clang-format`.

4. Almost always when include-sorting "breaks" something it's a bug elsewhere. The most common case is that you didn't IWYU (and then you would IWYU, not change the include order). In your issue #3120 you quote an actual bug in libc, and you could consider the Python/Qt example a "bug" in the design of C++ (which is why C++23 or whatever might actually fix the include/import system of C++).

5. And very importantly, this won't actually fix anything by itself. Let's say we currently had a include-order bug in GNU Radio, say like in your Python/Qt ordering. Then, on top of this PR, you would have to manually change the order. You might even add a comment to clarify why this order is important... in which case you wouldn't need this PR (see 3).

sbmueller:
Hello @mbr0wn thanks for your feedback. Let me respond to your arguments:

>In general, touching .clang-format should be considered very, very carefully. This particular PR would not break anything, because we're removing a rule, not changing or adding one, but changes to said file should be reserved for extreme cases. But let's consider this file sacrosanct unless it's really, really breaking stuff.

I guess this is subjective, but as you mentioned, this change does not introduce any formatting changes. Good to know how you want to treat `.clang-format`.

> clang-format doesn't just sort alphabetically, it sorts according to specific rules. We could even amend the sorting rules to put Python before Qt, but see 1.

Agreed, but this is also not what I stated. Sorting within one category is alphabetical. This is also mentioned in the clang documentation [1]: "The value belonging to the first matching regular expression is assigned and #includes are sorted first according to increasing category number and then alphabetically within each category." In our current implementation, we only treat a few include statements that match certain regexes a certain way. All remaining includes are sorted alphabetically. Please correct me if I'm wrong here.
The issue with Python before Qt is, I could include Python in a header file, that has a name `sender.h`. In this case, Python would be included after Qt wherever I include `sender.h` and Qt (since it's masked). Again, this underlines my argument **include order should not be dependent on file names**, since names are completely arbitrary, while include order can have severe consequences (again, see #3120 ).

> Could also be 2a: clang-format doesn't sort all the includes alphabetically, and you can modify its behaviour in-source. For example, you can put comments or newlines between includes to avoid reordering. That means if you know that includes have to be in a certain order, then you can still do that without touching .clang-format.

This is very helpful, thank you.

> Almost always when include-sorting "breaks" something it's a bug elsewhere. The most common case is that you didn't IWYU (and then you would IWYU, not change the include order). In your issue #3120 you quote an actual bug in libc, and you could consider the Python/Qt example a "bug" in the design of C++ (which is why C++23 or whatever might actually fix the include/import system of C++).

Not sure what the argument here is. Even if there are bugs elsewhere, we need to deal with them somehow. Saying "let's wait until they're fixed" seems not sensible to me. A sensible workaround seems a better approach, in particular if a bugfix is probably many years away.

> And very importantly, this won't actually fix anything by itself. Let's say we currently had a include-order bug in GNU Radio, say like in your Python/Qt ordering. Then, on top of this PR, you would have to manually change the order. You might even add a comment to clarify why this order is important... in which case you wouldn't need this PR (see 3).

Agreed, this change does not fix anything now. However, it can prevent issues in the future! As I stated, this **potentially** is an issue. Also, OOTs could make use of this `.clang-format` file and issues can arise there (exactly this happened in gr-inspector). But since I wasn't aware of 3., I see how this can be viewed as obsolete.

[1] https://clang.llvm.org/docs/ClangFormatStyleOptions.html

To conclude, I agree with you on 1. and 3., which is enough to close this in case no one else objects. Thanks again!

mbr0wn:
> Not sure what the argument here is. Even if there are bugs elsewhere, we need to deal with them somehow. Saying "let's wait until they're fixed" seems not sensible to me. A sensible workaround seems a better approach, in particular if a bugfix is probably many years away.

To clarify: My suggestion for a workaround is to.... reorder the includes! All you need to do is either tell clang-format to not touch the includes, or add newlines, or comments.

When we first applied clang-format, we saw approx. a million compilation failures[1] because include files were written in a way that they would not IWYU, or assume prior includes. We actually took the time to fix those include files, and thus, I think the reordering was super useful to find bugs (we could have also found those with other linter tools, but here we are). In general, SortIncludes has some advantages: It makes the include section more predictable, and filters dupes.

I'm not super-adamant about this, but because the actual fix (when such bugs occur) is the same regardless of whether or not we touch clang-format, I stay by my case that this PR is not necessary.
After all, there's no reason to assume that humans automatically order includes correctly, either. You have to be deliberately reordering includes, and you can still do that with `SortIncludes`.

[1] Rounded to the nearest million.

sbmueller:
Using `SortIncludes` to enforce good include behaviour is a very good point. I'm going to retract this PR.

PR #3240, opened 2020-03-02
gr-blocks: Add Phase Shift Block with Msg Capabilities

ghostop14:
This new block provides a native ability to phase shift signals
based on radians for solutions such as DoA and array-based projects.
The block supports both variable and message based updates such that
other blocks can provide the appropriate calculations for shifting
and only send message-based updates when necessary.

### DIFF
--- a/gr-blocks/grc/blocks_phase_shift.block.yml
+++ b/gr-blocks/grc/blocks_phase_shift.block.yml
@@ -1,0 +1,40 @@
+id: blocks_phase_shift
+label: Phase Shift
+
+parameters:
+-   id: shift
+    label: Phase Shift
+    dtype: float
+    default: '0.0'
+-   id: is_radians
+    label: Units
+    dtype: enum
+    options: ['True', 'False']
+    option_labels: ['Radians', 'Degrees']
+
+inputs:
+-   domain: stream
+    dtype: complex
+-   domain: message
+    id: shift
+    optional: true
+
+outputs:
+-   domain: stream
+    dtype: complex
+
+templates:
+    imports: from gnuradio import blocks
+    make: blocks.phase_shift(${shift}, ${is_radians})
+    callbacks:
+    - set_shift(${shift})
+
+documentation: "This block will phase shift the input signal by the specified phase\
+    \ by multiplying the input times a shift value: \n\
+    \ gr_complex(cos(d_shift_in_radians),sin(d_shift_in_radians))\n\n\
+    \ Notes:\n\
+    \ If degrees are provided, the block automatically handles the conversion.\n\
+    \ This block functions like a multiply const, but with the const limited to \
+    \ abs() == 1 to provide a constant phase shift." 
+
+file_format: 1
--- a/gr-blocks/include/gnuradio/blocks/phase_shift.h
+++ b/gr-blocks/include/gnuradio/blocks/phase_shift.h
@@ -1,0 +1,46 @@
+/* -*- c++ -*- */
+/*
+ * Copyright 2020 Free Software Foundation, Inc.
+ *
+ * This file is part of GNU Radio
+ *
+ * SPDX-License-Identifier: GPL-3.0-or-later
+ *
+ */
+
+#ifndef INCLUDED_PHASESHIFT_H
+#define INCLUDED_PHASESHIFT_H
+
+#include <gnuradio/blocks/api.h>
+#include <gnuradio/sync_block.h>
+
+namespace gr {
+namespace blocks {
+
+/*!
+ * \brief This block will shift the incoming signal by the specified amount.
+ * Shift can be specified in either radians or degress which is configurable
+ * in the constructor.
+ *
+ * This block functions like a multiply const, but with the const limited to
+ * abs() == 1 to provide a constant phase shift.
+ * \ingroup misc_blk
+ *
+ */
+class BLOCKS_API phase_shift : virtual public gr::sync_block
+{
+public:
+    typedef boost::shared_ptr<phase_shift> sptr;
+
+    /*!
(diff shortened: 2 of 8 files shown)


mormj:
@ghostop14 - this may be unwarranted concern, but it seems like a decently large amount of code just to wrap the multiply const parameter.  I'm sure there are more of this type of convenience block around existing blocks, in which case should they be just handled as hier blocks around the existing block?  Just wondering if we did a bunch of this type of thing if we are unnecessarily adding to the compilation time.  

By the way I love this block and have always gotten frustrated adding the appropriate conversions to do this with stock blocks.

ghostop14:
@mormj, thanks for all the feedback on the PR's!  On your comment here, I thought about implementing this as just a wrapper or something, but I have some antenna array apps where I need the message port to feed the shift as it's calculated by another block and updated as a message, and the multiply_xxx blocks don't have that capability.  So I think this version gives us the most flexibility to cover those types of use cases.  What do you think?

gr-blocks/lib/phase_shift_impl.cc
```diff
@@ -0,0 +1,92 @@
...
+
+/*
+ * The private constructor
+ */
+phase_shift_impl::phase_shift_impl(float shift_in_radians)
+    : gr::sync_block("phase_shift",
+                     gr::io_signature::make(1, 1, sizeof(gr_complex)),
+                     gr::io_signature::make(1, 1, sizeof(gr_complex)))
+{
+    d_shift_in_radians = shift_in_radians;
```
michaelld: I'd prefer a call to `set_shift` here. Reduce code complexity!

gr-blocks/include/gnuradio/blocks/phase_shift.h
```diff
@@ -0,0 +1,46 @@
...
+ * \ingroup misc_blk
+ *
+ */
+class BLOCKS_API phase_shift : virtual public gr::sync_block
+{
+public:
+    typedef boost::shared_ptr<phase_shift> sptr;
+
+    /*!
+     * \brief Return a shared_ptr to a new instance of mesa::phase_shift.
```
michaelld: `mesa`?

gr-blocks/include/gnuradio/blocks/phase_shift.h
```diff
@@ -0,0 +1,46 @@
...
+#define INCLUDED_PHASESHIFT_H
+
+#include <gnuradio/blocks/api.h>
+#include <gnuradio/sync_block.h>
+
+namespace gr {
+namespace blocks {
+
+/*!
+ * \brief This block will shift the incoming signal by the specified radians.
```
michaelld: can you note here and above in the GRC that this block is like a multiply const, but with the const limited to abs() == 1 to provide a constant phase shift (and thus the block name)? I think that would be useful for some users. ..

mormj:
> @mormj, thanks for all the feedback on the PR's! On your comment here, I thought about implementing this as just a wrapper or something, but I have some antenna array apps where I need the message port to feed the shift as it's calculated by another block and updated as a message, and the multiply_xxx blocks don't have that capability. So I think this version gives us the most flexibility to cover those types of use cases. What do you think?

Yes - you are right there does not seem to be a cleaner way to wrap at the current moment and maintain full performance.  Perhaps it would be helpful to also have the phase shift specified in degrees for those of us that don't think in radians?

ghostop14:
@mormj I like that idea.  Maybe a drop-list to select and just default to radians,  Same on the msg, default to radians and have an option to support degrees.  Let me work that in.

@michaelld I'll get those edits incorporated too.

ghostop14:
@mormj: Drop-list for degrees or radians.  I like it!

@michaelld: Edits all addressed.  In answer to your question about mesa, gr-mesa is my OOT where this block has been before upstreaming here.  Missed that clean-up reference.

gr-blocks/grc/blocks_phase_shift.block.yml
```diff
@@ -0,0 +1,40 @@
...
+    make: blocks.phase_shift(${shift}, ${is_radians})
+    callbacks:
+    - set_shift(${shift})
+
+documentation: "This block will phase shift the input signal by the specified phase\
+    \ by multiplying the input times a shift value: \n\
+    \ gr_complex(cos(d_shift_in_radians),sin(d_shift_in_radians))\n\n\
+    \ Notes:\n\
+    \ If degrees are provided, the block automatically handles the conversion.\n\
+    \ This block functions like a multiply const, but with the const limited to \
```
michaelld: "is like" ... yes?

gr-blocks/include/gnuradio/blocks/phase_shift.h
```diff
@@ -0,0 +1,46 @@
...
+
+namespace gr {
+namespace blocks {
+
+/*!
+ * \brief This block will shift the incoming signal by the specified amount.
+ * Shift can be specified in either radians or degress which is configurable
+ * in the constructor.
+ *
+ * This block functions like a multiply const, but with the const limited to
```
michaelld: "is like"

gr-blocks/grc/blocks_phase_shift.block.yml
michaelld: Oh ... sorry ... ignore this ... not enough coffee ...

gr-blocks/include/gnuradio/blocks/phase_shift.h
michaelld: ditto ... ignore this

mormj:
Two approvals - new block, requests addressed.  Merging.  thanks @ghostop14

Issue #3100, opened 2020-01-19
Filter design tool not opening


devnulling:
Testing with latest master , ubuntu 18.04. When trying to open, nothing happens and get this error in the terminal:

```
/usr/lib/python3.6/subprocess.py:616: ResourceWarning: unclosed file <_io.BufferedReader name=11>
  _cleanup()
/usr/lib/python3.6/subprocess.py:786: ResourceWarning: subprocess 22460 is still running
  ResourceWarning, source=self)
```

Semi related, but with older filter tool design: https://github.com/gnuradio/gnuradio/issues/1518

cbdonohue:
I was able to open the filter design tool with Ubuntu 19.10 and master commit  4846be906ad2d4b9d60dfae07f3c4dc6dd95fa95

However, pyqtgraph 0.11.0.dev0 didn't work, and I had to install pyqtgraph 0.11.0.dev0+g6a2791e
`pip3 install git+https://github.com/pyqtgraph/pyqtgraph@develop`

wbarnha:
Tried the same solution for 3.8 on Ubuntu 18.04 with pyqtgraph, no luck running it directly in GNU Radio. I tried running filter_design.py as indicated from #1518 with no luck. The issue is that in `filter_design.py`, (see line 40), the block of code is
```
try:
    from numpy.fft import fftpack as fft_detail
except ImportError:
```
Changing it to 
```
try:
    from scipy import fft as fft_detail
except ImportError:
```
gets filter_design.py working when the program is run from command line.

wbarnha:
I had to setup a new development environment and it looks like that specific version of PyQt is also needed to get the Filter Tool GUI working again as well as changing the modules imported by Python.

nickoe:
I think this is addressed via this PR on maint-3.8 https://github.com/gnuradio/gnuradio/pull/3259 and https://github.com/gnuradio/gnuradio/pull/3291 on master.

nickoe:
With https://github.com/gnuradio/gnuradio/pull/3291 (master) and https://github.com/gnuradio/gnuradio/pull/3259 (maint-3.8) merged, which essentially included the proposal to use `numpy.fft` directly https://github.com/gnuradio/gnuradio/pull/3291/commits/d3cffa3580a5377c48c313d628c298c81cee7573#diff-aa15db7a96c16ff668fd14aa585dfb90R29 I think we can close this one.

Issue #3235, opened 2020-02-27
Replace NumPy FFT with GR interface


michaelld:
As noted in https://github.com/gnuradio/gnuradio/pull/3069 and maybe elsewhere, NumPy changed their FFT interface from 'fftpack' to 'pocketfft' starting in 1.17.1. GR Leads discussion today on this topic resulted in a decision to create a GR interface to NumPy FFT and then change all internal references to be `from gnuradio import fft` or the like; which would then internally do something like from https://github.com/gnuradio/gnuradio/pull/2755 . We don't want to make SciPy a configure-time requirement.

nickoe:
With https://github.com/gnuradio/gnuradio/pull/3291 (master) and https://github.com/gnuradio/gnuradio/pull/3259 (maint-3.8) merged, which essentially included the proposal to use `numpy.fft` directly https://github.com/gnuradio/gnuradio/pull/3291/commits/d3cffa3580a5377c48c313d628c298c81cee7573#diff-aa15db7a96c16ff668fd14aa585dfb90R29 I think we can close this one.

PR #3241, opened 2020-03-03
filter_design: Changed numpy.fft.fftpack to numpy.fft

wbarnha:
I originally encountered this issue in #3100 and came up with this solution, but apparently this was already addressed for 3.9 in #3069 and #3232. I'd just like to fix this issue for 3.8.1+.

Apparently I was unaware or forgot about the contents of my folder, `Volk`, and this is part of the push. Should I resubmit?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,109 @@ Versioning](http://semver.org/spec/v2.0.0.html), starting with version 3.7.12.0.
 
 Older Logs can be found in `docs/RELEASE-NOTES-*`.
 
+## [3.8.1.0-rc1] - 2020-02-16
+
+### Changed
+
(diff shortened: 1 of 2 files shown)


mormj:
Hi @wbarnha - thank you for the commits.  Yes, please remove the volk update from the commit, this will cause undesired changes.  But I believe the current train of thought here is to handle the fft changes consistently across the code base by making a top level gr.fft per #3235

wbarnha:
Thanks for letting me know. I saw that post as well and gr.fft absolutely makes sense so that deprecation doesn't cause an issue like this. I have made some additional modifications that fixes the IIR issue with deprecation of scipy.arange() and more. However, I think this may need a much more in-depth discussion in a separate pull request.

PR #3242, opened 2020-03-03
ctrlport: fix examples

japm48:
`GNURadioControlPortClient` changed its constructor interface in cb044a4c4b2c6657876f996617a24a44d1ac3b5d.

This updates the examples accordingly. I could not test the USRP examples as I don't have the hardware.

### DIFF
--- a/gr-blocks/examples/ctrlport/simple_copy_controller.py
+++ b/gr-blocks/examples/ctrlport/simple_copy_controller.py
@@ -13,8 +13,7 @@ if(len(args) < 4):
 hostname = args[1]
 portnum = int(args[2])
 msg = args[3].lower()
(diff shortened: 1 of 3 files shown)


bastibl:
Looks good to me. I tested the copy and the USRP source example with the same setup and workarounds described in https://github.com/gnuradio/gnuradio/pull/2739#issuecomment-593910061 (Thrift 13).

michaelld:
I'm thinking this is the right thing to do. Let me double check my current patches to make sure ... Sorry for the delays ... such busy-ness!

bastibl:
@michaelld Just for reference: this were my previous changes, which are in line with this patch. I forgot to change these example. https://github.com/gnuradio/gnuradio/pull/2565/files

PR #3246, opened 2020-03-04
gr-network: Add grnet networking blocks to GNU Radio

ghostop14:
This PR is to create a new gr-network structure that includes
the grnet TCP and UDP source/sink blocks for inclusion into GNU Radio.
This includes new modes (TCP source/sink can act as either a TCP
server/listener or a client), and UDP supports multiple new header
options for dropped packet detection and integration with external
sources such as the Allen Telescope Array.  The UDP blocks have a
number of important performance and tuning comments in the block
documentation to help ensure success with the blocks, and examples
for each scenario are included.  This PR also deprecates the old
TCP/UDP blocks but keeps them in place.  The new blocks and new
functionality resulted in not a 1:1 drop-in replacement for the
old blocks so the old blocks were moved to the deprecated UI
group to avoid breaking flowgraphs going to GR 3.9.


ghostop14:
I know this one is going to have a good bit of discussion as there's a lot to review.  @mbr0wn this was my first shot at creating a new GR module so I copied some of the others, but another set of eyes to make sure I have the CMakes, etc. set up correct would be great. 

@noc0lour: I know you have a lot of interest in the IPv6 and performance side here.  This should integrate it so any testing would be great.  

For any testing on the UDP blocks, read the block documentation in the yml first.  There's stack-tuning advice to make sure packets don't get dropped before doing anything in "production".  I'm expecting a lot of discussion on these so all comments/suggestions are welcome till we get it dialed in for a successful PR.

mbr0wn:
I think this is great and we should expedite this. I would like to merge https://github.com/gnuradio/greps/pull/25 first, but that doesn't seem to be an issue.

ghostop14:
Some folks not familiar with the blocks were asking about the differences/enhancements between these and the original ones.  Here's a quick feature list:
1. The only TCP block is a sink that listens for inbound connections.  This inherently limits the ability to transmit data to another 3rd party application listening for data in TCP server mode.  A source block is included here as well.
2. The TCP sink block supports both TCP client and server modes.
3. All blocks (TCP and UDP) support IPv6.
4. UDP blocks now include a variety of header options: None, a 64-bit sequence number that can be used to track dropped packets, CHDR, and the Allen Telescope Array header format for GR-native ATA integration.
5. UDP blocks paired with headers can now notify if any data is lost in transit.
6. UDP blocks now have the option to source 0's (no signal) to allow the flowgraph to run if no incoming data is available.
7. UDP blocks now include a buffering mechanism to ensure data is not lost between different timing domains (network packets and GNU Radio scheduler).
8. Block documentation has been added to help guide users through how to properly configure addresses for IPv6 or dual-stack operations, and tuning / testing before using UDP blocks in a production environment.
9. TCP sink has enhanced work logic to remain running and continue to listen for reconnections if a client disconnects.

There's probably a number of other items as the code has been refactored several times, but those are the highlights.

PR #3247, opened 2020-03-04
fix(filter_design): Reduce scipy dependency and change scipy.arange to numpy.arange

wbarnha:
Sorry for creating another pull request, but I did find a bug here and I wanted to keep this separate from #3241 since I found a potential bug with IIR Bessel filters. For whatever reason, band pass and notch filters are not generated correctly. This issue is consistent in #3241 after changing only `scipy.arange()` to `np.arange()`, since a notification kept appearing and keeping me from using the IIR tool due to the deprectation of `scipy.arange()`.

Additionally, I changed the label from `IIR(scipy)` to `IIR` in `pyqt_filter_stacked.py`, since scipy is being abandoned in the future. I also set up `filter_design.py` to call `poly1d` from `numpy` rather than `scipy`, since there's a consensus that GNU Radio needs to abandon its dependency on scipy. I know that `gr.fft` is going to be used in the future, but I think numpy is a fair substitute for now.

If you want me to clean up my commit history, let me know.

### DIFF
--- a/gr-filter/python/filter/design/filter_design.py
+++ b/gr-filter/python/filter/design/filter_design.py
@@ -43,9 +43,14 @@ except ImportError:
     raise SystemExit('Please install NumPy to run this script (https://www.np.org/)')
     
 try:
-    from scipy import poly1d, signal
+    from scipy import signal
 except ImportError:
     raise SystemExit('Please install SciPy to run this script (https://www.scipy.org)')
+
+try:
+    from numpy import poly1d
+except ImportError:
+    raise SystemExit('Please install NumPy to run this script (https://www.np.org)')
 
 try:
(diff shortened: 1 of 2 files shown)


dkozel:
Hi @wbarnha Thanks for the PR.

I've taken a look over and I thing the general change over to using numpy is a good one. Can you please redo the branch so this PR only has the important commits. I think you only need the final commit for all the numpy.arange changes, though maybe it would be nice to have the `poly1p` change in a separate commit.

It looks like you had some trouble with the Volk submodule. The trick is to either run `git submodule update` so you're using the version of volk that matches the GNU Radio codebase at that time, or just make sure not to add the volk submodule change when you are doing `git add` to select the files to save.

Issue #3256, opened 2020-03-11
Python Module ID Missing 3.8.0.0


maxwell-maurice:
Hello, I am running 

OS: Manjaro Linux x86_64 
Kernel: 5.4.23-1-MANJARO 
GNU-Companion: 3.8.0.0 (python 3.8.1)

and it looks like the ID is missing from the python module block. Running 3.7.11
<img width="1117" alt="Screenshot_20200311_184619" src="https://user-images.githubusercontent.com/11984955/76476735-a02af400-63c8-11ea-9124-20c766675760.png">
 in Ubuntu I see the ID and can work from there. Any way to find out the ID of the block? 
<img width="1300" alt="Screenshot_20200311_175048" src="https://user-images.githubusercontent.com/11984955/76474317-df554700-63c0-11ea-94e9-f5f00195fa1f.png">

dkozel:
Hi @maxwell-maurice thanks for the report. Looks like it was removed in 3.8.0.0. The default value has not changed and you could see it by looking at the file name of the generated `.py` file after running or generating the flowgraph.

@mormj and I just has a quick thought last night about the ID of the `Python Block` block and making it user specifiable. It looks like this is actually a regression so probably pretty easy to fix.

mormj:
At least on master the block ids are exposed - I'm not seeing them on maint-3.8.  Also on master there is a feature to show all block ids which can be useful - this should be ported back to maint-3.8 as well.  I'll put together a few pull requests this afternoon to address

PR #3257, opened 2020-03-12
grc: backport show_id related commits to maint-3.8

mormj:
Closes: #3256 

This addresses the situation that blocks explicitly need a show_id flag in 3.8.  Several blocks whose block ids need to be given to other variables or blocks need this populated, and had been updated on master. 

This also backport the feature to 3.8 where a toggle is given to show ids for all blocks.


mormj:
This should address #3256

PR #3258, opened 2020-03-12
gr-utils: '#'replaced by '%' for commenting in codes.

### LINKED ISSUE #3195: Comments break matlab code in gr_utils/octave
The comments are made using "#" instead of "%" which causes the code to be nonfunctional in matlab. Octave can use either symbol so using "%" will only expand functionality and not remove it.

ex:
"# Copyright 2008 Free Software Foundation, Inc."
should be replaced with 
"% Copyright 2008 Free Software Foundation, Inc."

alekhgupta1441:
Fixes #3195 
In gr-utils/octave, replaced every '#' by '%' in matlab as Octave supports either of them for commenting while Matlab supports only '%'.This expands usability of code.

### DIFF
--- a/gr-utils/octave/cic_comp_taps.m
+++ b/gr-utils/octave/cic_comp_taps.m
@@ -1,17 +1,17 @@
-#
-# Copyright 2008 Free Software Foundation, Inc.
-#
-# This file is part of GNU Radio
-#
-# SPDX-License-Identifier: GPL-3.0-or-later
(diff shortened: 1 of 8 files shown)


dkozel:
Also looks good to me. Bypassing @mormj due to the limited number of changes, simplicity of changes, and because the changes are localized to files that aren't likely being worked on by anyone else.

PR #3259, opened 2020-03-13
fix(filter_design): Change numpy.fft.fftpack to numpy.fft and increase dependency on Numpy instead of Scipy

wbarnha:
Sorry for making a mess with my previous pull requests. Hopefully you will find this to be much more organized. For convenience, here are my previous notes:

Referring back to #3100, it seems that the only way for the Filter Design Tool to run correctly run is to update `numpy.fft.fftpack` to `numpy.fft` and run `pip3 install git+https://github.com/pyqtgraph/pyqtgraph@develop`. It seems that only the development build of `pyqtgraph` works with the filter tool. Should the development version be added as a dependency?

I changed the label from `IIR(scipy)` to `IIR` in `pyqt_filter_stacked.py`, since scipy is being abandoned in the future. I also set up `filter_design.py` to call `poly1d` from `numpy` rather than `scipy`, since there's a consensus that GNU Radio needs to abandon its dependency on scipy. I know that `gr.fft` is going to be used in the future, but I think numpy is a fair substitute for now. 

I also found a potential bug with IIR Bessel filters. For whatever reason, band pass and notch filters are not generated correctly. This issue is consistent in #3241 after changing only `scipy.arange()` to `np.arange()`, since a notification kept appearing and keeping me from using the IIR tool due to the deprectation of `scipy.arange()`.

Also, when installing from source, `python3-scipy` and `python3-numpy` don't seem to get this working, but when I ran `pip3 install numpy scipy`, it worked fine.

### DIFF
--- a/gr-filter/python/filter/design/filter_design.py
+++ b/gr-filter/python/filter/design/filter_design.py
@@ -1009,10 +1009,10 @@ class gr_plot_filter(QtGui.QMainWindow):
 
         # Set Data.
         if(type(self.taps[0]) == scipy.complex128):
-            self.rcurve.setData(scipy.arange(ntaps), self.taps.real)
-            self.icurve.setData(scipy.arange(ntaps), self.taps.imag)
+            self.rcurve.setData(np.arange(ntaps), self.taps.real)
+            self.icurve.setData(np.arange(ntaps), self.taps.imag)
         else:
-            self.rcurve.setData(scipy.arange(ntaps), self.taps)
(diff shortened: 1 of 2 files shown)


mbr0wn:
Argh I didn't review the git commit split & messages. This is a good change, see our other discussions on this topic, but please stick to the git commit message format we use elsewhere.

mbr0wn:
...and, we need this on master.

PR #3261, opened 2020-03-13
grc: workaround to allow epy blocks/modules to load

### LINKED ISSUE #3260: grc: embedded python objects fail to load because of missing cpp_templates
`
AttributeError: 'EPyBlock' object has no attribute 'cpp_templates'
`
This is with latest master edaa88fbc889a3d09fd374c3eb49cd46174b049f

Reproduce by opening a blank flowgraph in grc, and adding "Python Block" or "Python Module"

Going to see if there is an easy fix, but wanted to go ahead and post the issue

mormj:
The underlying issue is that the built in blocks all end up getting the
'cpp' flag, which is not valid for the embedded python blocks

To fix this some work needs to be done in the "Flags" class, as all the
flags added to classes that get accumulated in the build_ins list get
added to all the classes

See register_build_in decorator

Fixes #3260 for now, though a more thorough solution is needed going forward

### DIFF
--- a/grc/core/blocks/block.py
+++ b/grc/core/blocks/block.py
@@ -77,8 +77,11 @@ class Block(Element):
 
         self.states = {'state': True, 'bus_source': False, 'bus_sink': False, 'bus_structure': None}
 
-        if 'cpp' in self.flags and self.enabled and not (self.is_virtual_source() or self.is_virtual_sink()):
-            self.orig_cpp_templates = self.cpp_templates # The original template, in case we have to edit it when transpiling to C++         
+        if Flags.HAS_CPP in self.flags and self.enabled and not (self.is_virtual_source() or self.is_virtual_sink()):
+            # This is a workaround to allow embedded python blocks/modules to load as there is 
+            # currently 'cpp' in the flags by default caused by the other built-in blocks
+            if hasattr(self,'cpp_templates'):
+                self.orig_cpp_templates = self.cpp_templates # The original template, in case we have to edit it when transpiling to C++         
 
         self.current_bus_structure = {'source': None, 'sink': None}
 


mormj:
@seeAppRun - can you please take a look here?

japm48:
I had the same error while loading a `.grc` file that included components that were not installed.
This patch allowed opening that file (previously it crashed before opening).

mormj:
Adding a couple of reviewers to get this one merged quickly, as is prevents all epy blocks on the current master

michaelld:
@dkozel @marcusmueller can one of you review & deal with this please ... it's a simple change & should be a quick review!

nickoe:
@mormj I am not familiar with that code, but I have a comment.

I think that it is worth getting this in even if it is a workaround, but do you have a specific idea on what it would take to resolve this better. I mean making sure that what you state in your comment is not the case?

```
# This is a workaround to allow embedded python blocks/modules to load as there is 
# currently 'cpp' in the flags by default caused by the other built-in blocks
```

dkozel:
I don't see what the workaround is, changing from `cpp` to `Flags.HAS_CPP`? 

It definitely works, but can you add something to the workaround comment that explains what the workaround is doing? Right now reading the code it isn't clear at all why an Embedded Python Block would have 'cpp' but not pass Flags.HAS_CPP. Is it even a workaround or is this a strict improvement and the bug is that the Embedded Python Block has `cpp` added, but thankfully `Flags.HAS_CPP` doesn't pick that up or accounts for having the cpp_templates directly?

mormj:
This is the workaround - just to ignore the flag if it can't do anything with it
```
            if hasattr(self,'cpp_templates'):      
```
The root problem is that for the grc built in blocks, such as epy, the flags accumulate as subsequent built in blocks are loaded because Flags is a class variable.  It would require changing the structure a bit to make flags track the instance not the class I think.

The change to Flags.cpp was just to remove a hardcoded string comparison when it already was an enum elsewhere.

michaelld:
CI passes. 2 approvals. Critical fix. Merging.

PR #3265, opened 2020-03-14
Change assert to static_assert, where possible

ThomasHabets:


gr-audio/lib/jack/jack_source.cc
```diff
@@ -126,7 +126,7 @@ jack_source::jack_source(int sampling_rate,
 
     set_output_multiple(d_jack_buffer_size);
 
-    assert(sizeof(float) == sizeof(sample_t));
+    static_assert(sizeof(float) == sizeof(sample_t), "sample_t must be float");
```
rear1019: Considering the error message, a check for ``std::is_same<float, sample_t>::value`` is more appropriate.

ThomasHabets:
Fixed.

gr-dtv/lib/atsc/atsc_trellis_encoder_impl.cc
```diff
@@ -17,7 +17,7 @@
 namespace gr {
 namespace dtv {
 
-static const int DIBITS_PER_BYTE = 4;
+static constexpr int DIBITS_PER_BYTE = 4;
```
michaelld: is this change required for the rest of the changes to work?

michaelld: If not, I'd prefer to keep it separate ... in a different commit at least, if not just not in this PR.

ThomasHabets: My interpretation of the C++ standard is that because `static_assert` [takes a bool constexpr](https://en.cppreference.com/w/cpp/language/static_assert) it needs to be constexpr.

GCC seems to allow just const, but I think it's safer to not assume all compilers behave the same.

My interpretation may be wrong, though.

Anyway, I've split this into its own commit now.

michaelld: Thx!

michaelld:
Tagging a few more GR developers, just for a few more eyes. I think if any one of them approves this as-is they should go ahead & merge.

Issue #2873, opened 2019-11-01
missing 'options' key in bokeh gui option hider (GRC fails on 3.8.0.0 py3.8 qt5)


barjac:
```
[baz@jackodesktop ~]$ gnuradio-companion
Traceback (most recent call last):
  File "/usr/bin/gnuradio-companion", line 102, in <module>
    run_main()
  File "/usr/bin/gnuradio-companion", line 95, in run_main
    exit(main())
  File "/usr/lib/python3.8/site-packages/gnuradio/grc/main.py", line 83, in main
    platform.build_library()
  File "/usr/lib/python3.8/site-packages/gnuradio/grc/core/platform.py", line 197, in build_library
    utils.hide_bokeh_gui_options_if_not_installed(self.blocks['options'])
  File "/usr/lib64/python3.8/collections/__init__.py", line 891, in __getitem__
    return self.__missing__(key)            # support subclasses that define __missing__
  File "/usr/lib64/python3.8/collections/__init__.py", line 883, in __missing__
    raise KeyError(key)
KeyError: 'options'
```

marcusmueller:
Seems to be the same issue as in #2678

marcusmueller:
@barjac can you try to move (or delete) your `/home/baz/.gnuradio` folder and try again?

marcusmueller:
@skoslowski is it supposed that there's _always_ an `options` block?

barjac:
`[baz@jackodesktop ~]$ rm -rf ~/.gnuradio && gnuradio-companion
Traceback (most recent call last):
  File "/usr/lib/python3.8/site-packages/gnuradio/grc/gui/Application.py", line 102, in do_activate
    self.main_window = MainWindow(self, self.platform)
  File "/usr/lib/python3.8/site-packages/gnuradio/grc/gui/MainWindow.py", line 84, in __init__
    generate_modes = platform.get_generate_options()
  File "/usr/lib/python3.8/site-packages/gnuradio/grc/core/platform.py", line 379, in get_generate_options
    for param in self.block_classes['options'].parameters_data:
  File "/usr/lib64/python3.8/collections/__init__.py", line 891, in __getitem__
    return self.__missing__(key)            # support subclasses that define __missing__
  File "/usr/lib64/python3.8/collections/__init__.py", line 883, in __missing__
    raise KeyError(key)
KeyError: 'options'
^CTraceback (most recent call last):
  File "/usr/bin/gnuradio-companion", line 102, in <module>
[...]

Issue #2812, opened 2019-09-24
gr_modtool newmod path error on GR 3.8 pybombs install


ghostop14:
After installing GNU Radio 3.8 into a prefix via pybombs, the path created in the .gnuradio/config.conf is missing a "templates" portion of the path for 3.8.  The result is that the command "gr_modtool newmod" throws a "ModToolException: Could not find gr-newmod source dir." error.

I found the source of the issue.... the config.conf path generated during pybombs install looks like this:
newmod_path = <install prefix>/share/gnuradio/modtool/gr-newmod
but it should look like this:
newmod_path = <install prefix>/share/gnuradio/modtool/**templates**/gr-newmod

Updating config.conf manually by adding the missing /templates part fixes the issue.

primercuervo:
hi @ghostop14,

where is the absolute path of the `gnuradio/config.conf` file that you are mentioning? If it is `/home/user/.gnuradio/config.conf`, there is a high chance that this file was there _before_ your gnuradio 3.8 installation via PyBOMBS.

The logic behind my reasoning follows this [documentation link](https://wiki.gnuradio.org/index.php/Configuration_Files). The configuration file for the modtool newmod is located in `{prefix}/etx/gnuradio/conf.d/` and it's called `modtool.conf`, where the path should be correct (or at least it appeared to be so the last time I checked).

Please have a look at the contents of your prefix in order to confirm/deny the error. Keep in mind that a configuration file in `~/.gnuradio/config.conf` will override the settings of the prefixes for the user in that given machine.

ghostop14:
This was the one system I "upgraded".  Although I blew away the prefix and started from scratch.  I don't think I deleted $HOME/.gnuradio configs.  Maybe a solution could be for pybombs to check if it exists and ask the user to delete it and create a new one, back it up, and/or keep it?

marcusmueller:
this is a 3.7->3.8 upgrade problem, hence closing it. Sorry!

PR #3266, opened 2020-03-16
modtool: use prefix path for modtool instead of prefs

mormj:
Putting this up here for discussion - though I don't see the downsides to this change just yet.

When a prefix gets installed, the [modtool] section in prefs hardcodes
the path to the templates in the newly installed prefix as a global path
for all modtools.  This creates major problems when bouncing between
prefixes, especially when newmod changes between releases

Why not just use the prefix as the root of where to look for the
templates

This potentially contributed to issues such as #2812 and #2873

### DIFF
--- a/gr-utils/python/modtool/cli/newmod.py
+++ b/gr-utils/python/modtool/cli/newmod.py
@@ -21,7 +21,6 @@ from gnuradio import gr
 from ..core import ModToolNewModule
 from .base import common_params, run, cli_input, ModToolException
 
-
 @click.command('newmod', short_help=ModToolNewModule.description)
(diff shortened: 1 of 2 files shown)


michaelld:
Looks like it should be Py27 and Py3 compatible, too ... so, could be backported back to GR37!

mormj:
Could this technically be considered an api change though?

michaelld:
> Could this technically be considered an api change though?

hmmm ... I don't think so. All this change does is fix where to find the default source location for modtool  (or something like that) ... yes?

mormj:
two approvals ... merging.  Will do backport PRs for this as well

PR #3272, opened 2020-03-21
docs: updated usage manual export from wiki

777arc:
(these changes were automatically generated using the export script)

### DIFF
--- a/docs/usage-manual/(exported from wiki) Block Thread Affinity and Priority.txt
+++ b/docs/usage-manual/(exported from wiki) Block Thread Affinity and Priority.txt
@@ -3,17 +3,17 @@
     <ns>0</ns>
     <id>3489</id>
     <revision>
-      <id>4226</id>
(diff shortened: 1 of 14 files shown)


mbr0wn:
Do these files get pulled into something somewhere?

777arc:
Maybe every 6 months, or when there are big wiki improvements of these pages.  These pages live on the wiki, so this export is meant purely as a way of having an offline copy of the usage manual for people without internet.  Not 100% sure what you mean by the 2nd question.

mbr0wn:
I meant, do these txt files get consumed by Doxygen or something else. I think you've answered that -- it's just an offline copy.

777arc:
Ah, yeah these exported files are just wikimedia's format which is pretty readable and looks kind of like markup.  If you only look at the lines that changed, it doesn't look very readable, but that's just because most of the changes were in the tags.

PR #3276, opened 2020-03-22
Remove almost all manual memory management from gr-blocks

ThomasHabets:
The remaining one is `tcp_server_sink`, which is a bit trickier, so leaving it for a future PR.

### DIFF
--- a/gr-blocks/lib/udp_source_impl.cc
+++ b/gr-blocks/lib/udp_source_impl.cc
@@ -19,6 +19,7 @@
 #include <errno.h>
 #include <stdio.h>
 #include <string.h>
+#include <boost/make_unique.hpp>
 #include <stdexcept>
 
 namespace gr {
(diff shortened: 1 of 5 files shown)


gr-blocks/lib/udp_source_impl.cc
```diff
@@ -105,7 +101,7 @@ void udp_source_impl::disconnect()
     d_udp_thread.join();
 
     d_socket->close();
-    delete d_socket;
+    d_socket = nullptr;
```
mbr0wn: Don't you usually do `d_socket.reset()`?

ThomasHabets: [Looks to me like `reset()` mandates an argument until C++17](https://en.cppreference.com/w/cpp/memory/unique_ptr/reset), so could also be `d_socket.reset(nullptr);`.

I have no preference. Whatever you prefer.

rear1019: ``reset()`` has a default argument, both in pre and post C++17. The things which differ in pre/post C++17 only apply to array types (``T[]``), not to the “scalar” case here.

As far is the freeing of the memory is concerned: I prefer calling ``reset()`` to state the intent more clearly. Apart from that: LGTM.

mbr0wn: I have no strong opinion, but @rear1019's sounds good.

ThomasHabets: Fixed. Here and in udp sink.

PR #3280, opened 2020-03-27
fft: add shim implementation for python fft library

### LINKED ISSUE #3235: Replace NumPy FFT with GR interface

mormj:
Wrap the underlying available fft library inside of gnuradio.fft so that
the search is not done in multiple places.  The search here will fail
raise an ImportError on import of gnuradio.fft if no appropriate
underlying implementation is found

Fixes #3235

### DIFF
--- a/gr-fft/python/fft/__init__.py
+++ b/gr-fft/python/fft/__init__.py
@@ -22,3 +22,22 @@ except ImportError:
     dirname, filename = os.path.split(os.path.abspath(__file__))
     __path__.append(os.path.join(dirname, "..", "..", "swig"))
     from .fft_swig import *
+
+# Shim the appropriate python fft library here
+fft_detail = None
+try:
+    from numpy.fft import fftpack as fft_detail
+except ImportError:
+    # Numpy changed fft implementation in version 1.17
+    # from fftpack to pocketfft
+    try:
+        from numpy.fft import pocketfft as fft_detail
+    except ImportError:
+        try:
+            from numpy.fft import fft as fft_detail
+        except ImportError:
+            try:
+                from scipy import fft as fft_detail
+            except ImportError:
+                raise ImportError("Unable to load FFT library from NumPy or SciPy")
+        
--- a/gr-filter/python/filter/design/filter_design.py
+++ b/gr-filter/python/filter/design/filter_design.py
@@ -19,24 +19,13 @@ import warnings
 from optparse import OptionParser
 
 from gnuradio import filter
+from gnuradio.fft import fft_detail
 
 try:
     import numpy as np
 except ImportError:
     raise SystemExit('Please install NumPy to run this script (https://www.np.org/)')
 
-try:
-    from numpy.fft import fftpack as fft_detail
-except ImportError:
-
-    print('Could not import fftpack, trying pocketfft')
-    # Numpy changed fft implementation in version 1.17
-    # from fftpack to pocketfft
-    try:
-        from numpy.fft import pocketfft as fft_detail
-    except ImportError:
-        raise SystemExit('Could not import fft implementation of numpy')
-    
 try:
     from scipy import poly1d, signal
 except ImportError:
@@ -51,7 +40,6 @@ try:
     import pyqtgraph as pg
 except ImportError:
     raise SystemExit('Please install pyqtgraph to run this script (http://www.pyqtgraph.org)')
-
 
 try:
     from gnuradio.filter.pyqt_filter_stacked import Ui_MainWindow
--- a/gr-utils/python/utils/plot_fft_base.py
+++ b/gr-utils/python/utils/plot_fft_base.py
@@ -12,8 +12,7 @@ from __future__ import print_function
 from __future__ import division
 from __future__ import unicode_literals
 
-import numpy
-from numpy.fft import fftpack
+from gnuradio.fft import fft_detail
 
 try:
     from pylab import Button, connect, draw, figure, figtext, get_current_fig_manager, show, rcParams, ceil
@@ -88,7 +87,7 @@ class plot_fft_base(object):
 
     def dofft(self, iq):
         N = len(iq)
-        iq_fft = numpy.fft.fftshift(fftpack.fft(iq))       # fft and shift axis
+        iq_fft = numpy.fft.fftshift(fft_detail.fft(iq))       # fft and shift axis
         iq_fft = 20*numpy.log10(abs((iq_fft+1e-15) / N)) # convert to decibels, adjust power
         # adding 1e-15 (-300 dB) to protect against value errors if an item in iq_fft is 0
         return iq_fft


gr-filter/python/filter/design/filter_design.py
```diff
@@ -19,29 +19,21 @@
...
-    except ImportError:
-        raise SystemExit('Could not import fft implementation of numpy')
-    
 try:
     from scipy import poly1d, signal
 except ImportError:
     raise SystemExit('Please install SciPy to run this script (https://www.scipy.org)')
 
+if not fft_detail:
+    raise SystemExit('Unable to load fft library (numpy, scipy)')
```
nickoe: Wouldn't it make sense to reword to something like the following?

> Unable to load fft library from either numpy or scipy.

gr-utils/python/utils/plot_fft_base.py
```diff
@@ -88,7 +87,7 @@ def get_data(self):
 
     def dofft(self, iq):
         N = len(iq)
-        iq_fft = numpy.fft.fftshift(fftpack.fft(iq))       # fft and shift axis
+        iq_fft = numpy.fft.fftshift(fft_detail.fft(iq))       # fft and shift axis
```
nickoe: What about numpy.fft?

nickoe:
@mormj  I am a little bit confused with this issue, I probably missed some conversation. My main concern is that there are quite a bit of issues and merge requests trying to address this partially. It looks like this one also addresses https://github.com/gnuradio/gnuradio/pull/3232. Maybe we can close that merge request by @schneider42 with this?

mbr0wn:
@nickoe I think you're right. The many PRs related to this are one of the reasons we're doing this.

gr-utils/python/utils/plot_fft_base.py
mbr0wn: Well, this could be numpy.fft depending on your numpy version.

gr-fft/python/fft/__init__.py
```diff
@@ -22,3 +22,28 @@
...
+        from numpy.fft import pocketfft as fft_pylib
+    except ImportError:
+        try:
+            from numpy.fft import fft as fft_pylib
+        except ImportError:
+            try:
+                from scipy import fft as fft_pylib
+            except ImportError:
+                pass
+                # out of options, fail silently
```
mbr0wn: Should we fail silently? It seems we're asking our tools to rely on this. Plus, numpy is a dependency of GNU Radio and we'd fail elsewhere if it weren't there.

gr-filter/python/filter/design/filter_design.py
mbr0wn: If we move the failure to the `__init__` file (see above) then we never have to do this in tools.

schneider42:
Can someone explain why it is necessary to import an implementation detail from numpy and not simply `numpy.fft`?

gr-utils/python/utils/plot_fft_base.py
mormj: > I guess the comment message should be adjusted as well, it does not fail silently this time.

good point!

mormj:
> Can someone explain why it is necessary to import an implementation detail from numpy and not simply `numpy.fft`?

That is an excellent point - I was guessing it was due to historical numpy api changes? But I'm not really sure.

nickoe:
> Can someone explain why it is necessary to import an implementation detail from numpy and not simply `numpy.fft`?

I am not too familiar with the module or api, but related to https://numpy.org/doc/1.17/release.html

> NumPy’s FFT implementation was changed from fftpack to pocketfft, resulting in faster, more accurate transforms and better handling of datasets of prime length. See below under Improvements.

So in short. Something chaged in numpy. I don't know if numpy.fft.fft(...) would be the same.

nickoe:
Am I misunderstanding anything? `numpy.fft.fft(...)` do seem to work for me. Note the numpy version strings in the console output below.

On an old docker image I had around:
```
root@6274fa874761:/# ipython
Python 2.7.6 (default, Nov 13 2018, 12:45:42) 
Type "copyright", "credits" or "license" for more information.

IPython 1.2.1 -- An enhanced Interactive Python.
?         -> Introduction and overview of IPython's features.
%quickref -> Quick reference.
help      -> Python's own help system.
object?   -> Details about 'object', use 'object??' for extra details.

In [1]: import numpy as np

In [2]: np.__version__
Out[2]: '1.11.0'

In [3]: signal = np.array([-2, 8, 6, 4, 1, 0, 3, 5], dtype=float)

In [4]: fourier = np.fft.fft(signal)

In [5]: fourier
Out[5]: 
array([ 25.00000000+0.j        ,   3.36396103-7.94974747j,
       -10.00000000+1.j        ,  -9.36396103-1.94974747j,
        -9.00000000+0.j        ,  -9.36396103+1.94974747j,
       -10.00000000-1.j        ,   3.36396103+7.94974747j])
``` 

On my desktop:
```
$ ipython3 
Python 3.8.2 (default, Feb 26 2020, 22:21:03) 
Type 'copyright', 'credits' or 'license' for more information
IPython 7.13.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import numpy as np                                                                           

In [2]: np.__version__                                                                               
Out[2]: '1.18.0'

In [3]: signal = np.array([-2, 8, 6, 4, 1, 0, 3, 5], dtype=float)                                    

In [4]: fourier = np.fft.fft(signal)                                                                 

In [5]: fourier                                                                                      
Out[5]: 
array([ 25.        +0.j        ,   3.36396103-7.94974747j,
       -10.        +1.j        ,  -9.36396103-1.94974747j,
        -9.        +0.j        ,  -9.36396103+1.94974747j,
       -10.        -1.j        ,   3.36396103+7.94974747j])

```

mormj:
@nickoe I'm starting to wonder the same thing - and whether we need the shim at all.  Does anyone know why the underlying reference to fftpack or pocketfft was used in the first place?  Doesn't numpy.fft already wrap the underlying implementation?  In which case, we just use numpy.fft.  @mbr0wn @marcusmueller ?

mbr0wn:
I was assuming that `numpy.fft` doesn't work across all numpy versions that we support, but I didn't double-check that. If `numpy.fft` works, then there's really no point in the shim.

mbr0wn:
Hm, we don't even have a minimum numpy version, and it doesn't get checked in the top-level CMakeLists.txt.

We can declare numpy existential for GNU Radio and set the minimum to 1.13 (that works with 18.04). Then numpy.fft is just fine.

mbr0wn:
If we merge https://github.com/gnuradio/gnuradio/pull/3290, then we will always have numpy.fft available if ENABLE_PYTHON is ON.

nickoe:
> We can declare numpy existential for GNU Radio and set the minimum to 1.13 (that works with 18.04). Then numpy.fft is just fine.

I even think we can go as far as using 1.11 as the minumum if we just consider the fft issue at hand. See my earlier comment (quick link: https://github.com/gnuradio/gnuradio/pull/3280#issuecomment-605444067)

mormj:
closing in favor of #3291

Issue #3145, opened 2020-01-30
Ctrl+0 should reset zoom to 100%


primercuervo:
Now that we just were mindblown with the realization that you can, indeed, zoom in and out a flowgraph using `ctrl`+mouse wheel, it would be useful to have a hotkey to reset the zoom to 100%.

dkozel:
`ctrl`+`+` and `ctrl`+`-` should zoom in and out.

skoslowski:
Haha, that is hilarious :smile:

dkozel:
The mindblowing has continued. First it was about half of the hackfest participants, then it was more folks over dinner at FOSDEM. One pulled out a laptop on the spot to try it. ;) Great to have people smile.

jaskiratsingh2000:
Hey @primercuervo @skoslowski @dkozel I am quite beginner and super interested to begin my contribution towards this repo and this issue seems interesting to me. Can you please let me know that where exactly do I have to make changes or add functionalities?
Any examples corresponding to this would be highly appreciable. Thanks!

jaskiratsingh2000:
Also, is this fully opened yet @mbr0wn ?

PR #3285, opened 2020-03-28
grc: added shortcut to reset zoom

jdak907:
Fixes https://github.com/gnuradio/gnuradio/issues/3145


dkozel:
I worked with @jdak907 on this. Code is good and functions correctly. We'll add the zoom in/out soon in a separate PR.

nickoe:
Maybe add it to `show_keyboard_shortcuts` in `grc/gui/Dialogs.py` as well?

japm48:
Wow, I just started working on this. I should have informed earlier :smiley:.

By the way, just to save you some searching with the zoom in/out patch: to enable both `<Ctrl>` +  regular plus/minus and  `<Ctrl>` keypad plus/minus, you can use this:
```
ZOOM_IN = actions.register(...,
    keypresses=["<Ctrl>plus", "<Ctrl>KP_Add"],
)
ZOOM_OUT = actions.register(...,
    keypresses=["<Ctrl>minus", "<Ctrl>KP_Subtract"],
)```

japm48:
Also, you may want to change `["<Ctrl>0"]` to `["<Ctrl>0","<Ctrl>KP_0"]` in this PR to also allow using the keypad.

For the record, I found the key names here: https://gitlab.gnome.org/GNOME/gtk/-/blob/3.24.14/gdk/keyname-table.h

jdak907:
@japm48 and @nickoe thanks for the input, I have amended my commit.

jdak907:
@japm48 I've got it working with the 'add' and 'subtract' KP. You have to do ctrl+shift+= to trigger the 'add' keypress. Assuming that all keyboard layouts are the same with a +/= button: I think it would be less confusing for a human if: <ctrl>EQUALS and KP_EQUALS leaving the dialogues.py key shortcut as "<ctrl>++". Or would you recommend just leaving it as 'Add' and change the shortcut to "<ctrl>+<shift>+"?

japm48:
Oh, I see.

I generally use a US keyboard, but I can confirm that, unfortunately, the +/= button is not ubiquitous and not present even in common layouts. E.g. [Spanish](https://es.m.wikipedia.org/wiki/Archivo:KB_Spanish.svg) and [German](https://en.wikipedia.org/wiki/German_keyboard_layout#/media/File:KB_Germany.svg), also Latin Spanish, French, etc. They have a `+` key that is input with no modifier key. The minus/underscore key is identical (except it's located elsewhere).

I just tried with Chrome and Firefox and with US layout they happen to ignore the shift modifier, and the same for ES layout (well, except Firefox behaves different with `Ctrl+Shift+-`, but shouldn't matter).

So I would suggest trying this (not tested):
`keypresses=["<Ctrl>plus", "<Ctrl><Shift>plus", "<Ctrl>KP_Add"]`

If that doesn't work:
`keypresses=["<Ctrl>plus", "<Ctrl>equal", "<Ctrl>KP_Add"]`

For `ZOOM_OUT`, this should still be correct: `keypresses=["<Ctrl>minus", "<Ctrl>KP_Subtract"]`

mormj:
@jdak907 - were you able to get the CLA in place?

jdak907:
I thought I had submitted all of the information. I had recently moved from
Alaska to Colorado and forgot about it! This was my first CLA so I am not
sure what the entire process looks like.
Jeff

mormj:
@jdak907 - Good news is that we are using the DCO instead of CLA (https://github.com/gnuradio/gnuradio/blob/master/CONTRIBUTING.md)

All this means, is that we can merge your changes if you signoff your commit `git commit --amend -s` and force push

(looks like it needs a rebase as well)

mbr0wn:
@jdak907 We are really, really interested in this change. However, I'll close it since it's waiting on your update (DCO + rebase). Please reopen when that's done. We're trying to get the PR queue pared down a bit, and not distract attention from other PRs.

Issue #3282, opened 2020-03-27
examples: test_msg_hier.grc doesn't load


mormj:
I think a similar issue has been fixed in other flowgraphs, but documenting here:

![image](https://user-images.githubusercontent.com/34754695/77783281-f2a61a80-702e-11ea-939e-63fbe32d95ed.png)

```
Loading: "/share/gnuradio/grdev/src/gnuradio/gr-blocks/examples/msg_passing/hier/test_msg_hier_topblock.grc"
>>> Loading: /share/gnuradio/grdev/src/gnuradio/gr-blocks/examples/msg_passing/hier/test_msg_hier.grc
>>> Error: Connection between blocks_message_strobe_0(strobe) and pad_sink_0(in0) could not be made.
	sink key 'in0' not in sink block keys
>>> Load Error: /share/gnuradio/grdev/src/gnuradio/gr-blocks/examples/msg_passing/hier/test_msg_hier.grc: Flowgraph invalid
```

nickoe:
What branch is this on?

mormj:
> What branch is this on?

master

nickoe:
It looks like this on maint-3.8 (a65d1d1c236f95e4a106db6431ff4d51021dc29c)

```
>>> Error: Connection between blocks_message_strobe_0(strobe) and pad_sink_0(in0) could not be made.
	sink key 'in0' not in sink block keys
Traceback (most recent call last):
  File "/tmp/gnuradio/lib/python3.8/site-packages/gnuradio/grc/core/FlowGraph.py", line 418, in import_data
    sink_port = verify_and_get_port(snk_port_id, sink_block, 'sink')
  File "/tmp/gnuradio/lib/python3.8/site-packages/gnuradio/grc/core/FlowGraph.py", line 399, in verify_and_get_port
    raise LookupError('%s key %r not in %s block keys' % (dir, key, dir))
LookupError: sink key 'in0' not in sink block keys
```

nickoe:
Adding the connections and diffing what is saved gives:
```
...
@@ -118,16 +136,18 @@ blocks:
     type: message
     vlen: '1'
   states:
+    bus_sink: false
+    bus_source: false
+    bus_structure: null
     coordinate: [120, 148.0]
     rotation: 0
     state: enabled
 
 connections:
-- [blocks_message_strobe_0, strobe, pad_sink_0, in0]
-- [blocks_message_strobe_0_1, strobe, pad_sink_0, in0]
[...]

PR #3286, opened 2020-03-28
Fixup some grc example files in; blocks, docs and digital

nickoe:
Fixes https://github.com/gnuradio/gnuradio/issues/3282 and partly fixes unresolved notes in https://github.com/gnuradio/gnuradio/issues/2285

There are some grc blocks that can not open correctly because of the connections between the blocks does not seem to have an number suffix if there is only one port.

See for example `test_msg_hier.grc`.

If the "Num Streams" are set to 1 the single input port is called "in", if there are multiple they are suffixed with a number, as in "in0", "in1"...

### DIFF
--- a/gr-blocks/examples/msg_passing/hier/test_msg_hier.grc
+++ b/gr-blocks/examples/msg_passing/hier/test_msg_hier.grc
@@ -24,6 +24,9 @@ options:
     title: ''
     window_size: 1280, 1024
   states:
+    bus_sink: false
+    bus_source: false
+    bus_structure: null
     coordinate: [10, 10]
     rotation: 0
     state: enabled
@@ -40,6 +43,9 @@ blocks:
     msg: pmt.cons(pmt.PMT_NIL, pmt.make_u8vector(16,0x77))
     period: '200'
   states:
+    bus_sink: false
+    bus_source: false
+    bus_structure: null
     coordinate: [362, 81]
     rotation: 0
     state: enabled
@@ -54,6 +60,9 @@ blocks:
     msg: pmt.intern("OUTPUT2")
     period: '100'
   states:
+    bus_sink: false
+    bus_source: false
+    bus_structure: null
     coordinate: [400, 156]
     rotation: 0
     state: enabled
@@ -69,6 +78,9 @@ blocks:
     type: message
     vlen: '1'
   states:
+    bus_sink: false
+    bus_source: false
+    bus_structure: null
     coordinate: [618, 87]
     rotation: 0
     state: enabled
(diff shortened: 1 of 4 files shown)


gr-blocks/examples/msg_passing/hier/test_msg_hier.grc
```diff
@@ -118,16 +136,18 @@ blocks:
...
+    bus_sink: false
+    bus_source: false
+    bus_structure: null
     coordinate: [120, 148.0]
     rotation: 0
     state: enabled
 
 connections:
-- [blocks_message_strobe_0, strobe, pad_sink_0, in0]
-- [blocks_message_strobe_0_1, strobe, pad_sink_0, in0]
```
nickoe: Hmm, did `blocks_message_strobe_0_1` really go to `pad_sink_0` as well? Any idea why this may make sense?

nickoe: Ok, I checked with the version in maint-3.7 and it should now be correct. There were indeed some "funny" connections.

nickoe:
I just updated it to to have that connection in `test_msg_hier.grc` that I was unsure about. I verified in maint-3.7 that that is indeed what it is supposed to look like.

gr-blocks/examples/msg_passing/hier/test_msg_hier.grc
```diff
@@ -24,6 +24,9 @@ options:
     title: ''
     window_size: 1280, 1024
   states:
+    bus_sink: false
```
michaelld: It looks like these 3 entries here and elsewhere come from using a recent GRC, and that these are defaults ... yes? If so, what I'd greatly prefer is to do these updates in 1 commit, then the msg port id updates in another commit. They are different, so let's keep them separate.

michaelld:
Are these already on master somehow? Is this a backport?

gr-blocks/examples/msg_passing/hier/test_msg_hier.grc
nickoe: I used gnuradio-companion from `maint-3.8`, and this is for the `maint-3.8` branch so I don't see why they shall not be included here. I don't think it makes sense to split them out in a separate commit in this case.

nickoe:
@michaelld In the other pull request for `master` https://github.com/gnuradio/gnuradio/pull/3289 the docs change is not there, but the blocks and digital flowgraphs still needs to be fixed as far as I can tell. Not really a backport as much as it is just something that needs to be fixed.

PR #3288, opened 2020-03-29
Delete __init__.py

SuperGops7:
Removed an empty file off the codebase.


nickoe:
This should not be removed. Please read section 6.4 on https://docs.python.org/3/tutorial/modules.html

PR #3289, opened 2020-03-29
Fixup some grc example files in; blocks and digital

### LINKED ISSUE #3282: examples: test_msg_hier.grc doesn't load

nickoe:
Fixes #3282 and partly fixes unresolved notes in #2285

Backport of https://github.com/gnuradio/gnuradio/pull/3286 maint-3.8 fix onto master.

The docs change is not needed on master as it seems that has been fixed and moved to gr-analog.

### DIFF
--- a/gr-blocks/examples/msg_passing/hier/test_msg_hier.grc
+++ b/gr-blocks/examples/msg_passing/hier/test_msg_hier.grc
@@ -24,6 +24,9 @@ options:
     title: ''
     window_size: 1280, 1024
   states:
+    bus_sink: false
+    bus_source: false
+    bus_structure: null
     coordinate: [10, 10]
     rotation: 0
     state: enabled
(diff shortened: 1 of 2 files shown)


nickoe:
I could technically split it, but I don't think it makes sense as it is just "one action" done in gnuradio-companion, that is; add the connections and save. I don't think we gain anything history wise in the git log by splitting it more than this.

michaelld:
Yeah let's just leave it be as is ...

PR #3290, opened 2020-03-29
cmake: Make numpy a first-class dependency for Python

mbr0wn:
Yeah, I hate dependencies. But let's not kid ourselves, numpy was always a dependency of GNU Radio. Might as well treat it as such. Should then obviate #3280.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -50,6 +50,7 @@ set(GR_SWIG_MIN_VERSION "3.0.8")
 set(GR_CMAKE_MIN_VERSION "3.5.1")
 set(GR_MAKO_MIN_VERSION "0.4.2")
 set(GR_PYTHON_MIN_VERSION "3.6.5")
+set(GR_NUMPY_MIN_VERSION "1.13.0")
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -50,6 +50,7 @@ set(GR_SWIG_MIN_VERSION "3.0.8")
 set(GR_CMAKE_MIN_VERSION "3.5.1")
 set(GR_MAKO_MIN_VERSION "0.4.2")
 set(GR_PYTHON_MIN_VERSION "3.6.5")
+set(GR_NUMPY_MIN_VERSION "1.13.0")
```
mbr0wn: Note: This is what 18.04 has.

mormj: Seems like for master this is a good floor

nickoe:
@morm @mbr0wn  is this something we need for maint-3.8?

michaelld:
> @morm @mbr0wn is this something we need for maint-3.8?

Yes, if not done already. Would be useful for maint-3.7 too if that's easily doable ...

PR #3295, opened 2020-04-01
digital: test_corr_est example 3.8 - missing parenthesis

mormj:
Previous version missed something in the transition to yaml, namely
parenthesis around the lengths list in the stream mux.

Now the example runs out of the box


michaelld:
This is a backport. All CI passes. Tested on Py2. Multiple approvals. Merging.

PR #3024, opened 2020-01-06
Logging instead of stderr

marcusmueller:

PR #3297, opened 2020-04-02
WAV file sink append file option

### LINKED ISSUE #1569: WAV file sink append file option
Hi, it is possible to add the append file option to wav file sink block?. It would be very useful feature.

 Regards.

japm48:
Fixes #1569.

This was a bit more complicated than I expected.

The test `.wav` was corrupt, which causes some problems; see first commit for details.

I have also refactored a bit of the sink and source, but there is still room for improvement.
I didn't split the commit, as it would make things more difficult, hopefully that's is good enough.


gr-blocks/lib/wavfile.cc
```diff
@@ -74,15 +69,29 @@ bool wavheader_parse(FILE* fp,
...
+
     fresult = fread(str_buf, 1, 4, fp);
     if (fresult != 4 || strncmp(str_buf, "RIFF", 4) || feof(fp)) {
         return false;
     }
 
     fresult = fread(&file_size, 1, 4, fp);
+    file_size = wav_to_host(file_size);
+    if (fresult != 4 || file_size != real_file_size - 8L) {
+        fprintf(stderr,
```
mormj, line 88: Would be nice to get the fprintfs replaced with GR_LOG_ERROR(...), but that may be better suited as a separate PR across the entire codebase.

mormj:
I'm trying to manually test out the wavfile sink in GRC, and want to understand what I should expect to see.  

image

If I make a simple flowgraph - setting append to false with the test wavfile (76 bytes) outputs a 52 bit wavfile.  

Setting append to True with no output file already having been created, outputs a 0 byte output file

Setting append to True with the first 52 byte output file adds 8 bytes on each run of the flowgraph

Should append with no file existing create a new file first?

gr-blocks/lib/wavfile.cc
japm48: Thanks, I didn't know that. I will use a separate PR after this one as you suggest.

japm48:
> Setting append to True with no output file already having been created, outputs a 0 byte output file
> ...
> Should append with no file existing create a new file first?

Well, the intended behaviour was to fail if the file did not exist (mode `r+b`), it is a bug that it creates one (I will fix it: `open` should not use `O_CREAT`; I will add a test case for this.).

If you think that "create if doesn't exist" is more reasonable, please tell me so.

> Setting append to True with the first 52 byte output file adds 8 bytes on each run of the flowgraph

That is correct: the data size in `test_16bit_1chunk.wav` is only 8 bytes (4 samples in mono 16-bit PCM), the rest is just header chunks.

japm48:
Error on non-existing file is implemented now.

I'll wait for #3024 and rebase on top of it using proper logging.

gr-blocks/lib/wavfile.cc
dkozel: @marcusmueller was just taking a look at reviving https://github.com/gnuradio/gnuradio/pull/3024

japm48:
I just rebased, as promised, on top of #3339.

I had never done a non-trivial conflict resolution before, so it was tougher than expected but at least, I learned a couple of things...

japm48:
I messed up the commits and accidentally squashed them. I will try to fix this.

mormj:
I think we should go ahead and merge as is - it's been rebased and fixed up a lot and these changes would be great to have in there.

japm48:
@michaelld Ok I will try that. I realized I still have a couple of `printf`s scattered.

japm48:
Should be done now.

michaelld:
All CI passes. Commit messages look good & commits make sense. Multiple positive reviews. Merging.

drmpeg:
I'm not seeing the append option. Looks like the update to blocks_wavfile_sink.block.yml is missing?

Also, the use of ftell() is not 32-bit compatible with large files. If you're interested in the recommended 32-bit techniques, see https://github.com/gnuradio/gnuradio/blob/master/gr-blocks/lib/file_source_impl.cc and https://github.com/gnuradio/gnuradio/blob/master/gr-blocks/lib/CMakeLists.txt#L188.

japm48:
> I'm not seeing the append option. Looks like the update to blocks_wavfile_sink.block.yml is missing?

Oops, I forgot to add it after changing the commits. Will fix.

> Also, the use of ftell() is not 32-bit compatible with large files. If you're interested in the recommended 32-bit techniques, see https://github.com/gnuradio/gnuradio/blob/master/gr-blocks/lib/file_source_impl.cc and https://github.com/gnuradio/gnuradio/blob/master/gr-blocks/lib/CMakeLists.txt#L188.

Thanks, I knew that (I put a [TODO comment](https://github.com/gnuradio/gnuradio/pull/3297/files#diff-e820041fbdedc33b2c7c047c091d3d7eR165)). I just wanted to minimize the changes from the original version.
There are other uses in file_meta_source/sink. Would you suggest me to extract `GR_FSEEK` and `GR_FTELL` to a separate header?

drmpeg:
Ahh, I didn't see that comment. But yes, a separate header would be the way to go. Do so at your leisure though. Probably not many folks writing giant WAV files on 32-bit systems.

PR #3292, opened 2020-03-30
fec: make SciPy an optional dependency

nickoe:
This is a backport of https://github.com/gnuradio/gnuradio/pull/1639 that was made to fix https://github.com/gnuradio/gnuradio/issues/1545 which specifically caught my attention as it was still open and I just tried it on maint-3.8.

PR #3298, opened 2020-04-02
fec: make SciPy an optional dependency

nickoe:
This patch fixes issues like #1545 and #1539.
Basically if a module in fec.polar requires SciPy and fails to do ,
it overwrites the missing function with an approximate version

This is already merged on maint-3.8 in https://github.com/gnuradio/gnuradio/pull/3292


michaelld:
CI passes. 2 approvals. Forward-port of useful code. Merging.

PR #3299, opened 2020-04-02
replace boost endian check with cmake

japm48:
This uses cmake's TEST_BIG_ENDIAN instead of Boost.

Note that current use of `boost/detail/endian.hpp` is deprecated:
```The use of BOOST_*_ENDIAN and BOOST_BYTE_ORDER is deprecated. Please include <boost/predef/other/endian.h> and use BOOST_ENDIAN_*_BYTE instead```

Also, I am not sure if GNU Radio works on any of the remaining big-endian platforms.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -168,6 +168,12 @@ include(CMakeOverloads)
 ########################################################################
 include(GrMiscUtils) #compiler flag check
 
+include(TestBigEndian)
+TEST_BIG_ENDIAN(GR_IS_BIG_ENDIAN)
+if(GR_IS_BIG_ENDIAN)
+    add_definitions(-DGR_IS_BIG_ENDIAN)
+endif(GR_IS_BIG_ENDIAN)
+
 if(CMAKE_CXX_COMPILER_ID MATCHES "Clang" OR
    CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
     if(NOT WIN32)
--- a/gr-audio/lib/osx/osx_common.h
+++ b/gr-audio/lib/osx/osx_common.h
@@ -48,8 +48,7 @@ namespace osx {
         GR_LOG_WARN(d_logger, boost::format("  %s:%d") % __FILE__ % __LINE__);        \
     }
 
-#include <boost/detail/endian.hpp> //BOOST_BIG_ENDIAN
-#ifdef BOOST_BIG_ENDIAN
+#ifdef GR_IS_BIG_ENDIAN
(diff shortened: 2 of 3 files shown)


mormj:
This makes sense to me and seems to be a positive logical change, but I am no expert in CMake

michaelld:
Good catch! I found that deprecation a bit ago but forgot to do anything about it. Removing Boost is always desirable if we can do it! For this, "TestBigEndian" was added in CMake 3.0.2, and GR requires 3.8 ... so that's not an issue. There are still folks who build on PPC, so we do need this. Let me do a quick code review to verify.

michaelld:
My primary concern is for cross-compiling ...

michaelld:
@balister do you know if CMake's `TestBigEndian` works for cross-compiling? That's kinda important here ... if so, then IMHO this PR is good to go. If not, then there's another solution.

japm48:
I am trying to test with PowerPC64 (big-endian variant) on QEMU, but the setup is more complicated than expected.
I will report back when finished.

japm48:
Ok I managed to test it: it works in MIPS; I could not manage to install.

michaelld:
@japm48 Cool! seems like it's mostly working for you .. yes? is this a direct build, or a cross-build?

japm48:
@michaelld Only cross-build. I think direct build would take too long without real hardware (I would have to build the toolchain inside QEMU).

japm48:
I've had a look at how it is done in CMake:
https://github.com/Kitware/CMake/blob/master/Modules/TestBigEndian.cmake
https://github.com/Kitware/CMake/blob/master/Modules/TestEndianess.c.in

It compiles a test executable that uses 2-byte words and examines the binary output.
The endianness is guessed from the (absence of) byte swapping in the result.
So it makes sense that it should work in both cross-compile and native.

michaelld:
> I've had a look at how it is done in CMake:
> https://github.com/Kitware/CMake/blob/master/Modules/TestBigEndian.cmake
> https://github.com/Kitware/CMake/blob/master/Modules/TestEndianess.c.in
> 
> It compiles a test executable that uses 2-byte words and examines the binary output.
> The endianness is guessed from the (absence of) byte swapping in the result.
> So it should most likely work in both cross-compile and native.

Cool ... thanks for that info & review!

michaelld:
CI passes. Multiple approvals. Less Boost is better ... so, we'll see if this works for XC once somebody gets there! Merging.

michaelld:
Can you do a backport to maint-3.8 too?

PR #3303, opened 2020-04-02
grc: show file path as tooltip in tab label

### LINKED ISSUE #3294: GRC: Hover over text for open flowgraphs
It would be nice if hovering the mouse over the tab for open flowgraphs displayed the full path of the file.  Often I have multiple flowgraphs open with the same name, and lose track of which is which.

japm48:
Fixes #3294.

### DIFF
--- a/grc/gui/MainWindow.py
+++ b/grc/gui/MainWindow.py
@@ -361,6 +361,10 @@ class MainWindow(Gtk.ApplicationWindow):
                 foreground='black' if page.saved else 'red', ro=' (ro)' if page.get_read_only() else '',
                 title=Utils.encode(file_name or NEW_FLOGRAPH_TITLE),
             ))
+            fpath = page.file_path
+            if not fpath:
+                fpath = '(unsaved)'
+            page.set_tooltip(fpath)
         # show/hide notebook tabs
         self.notebook.set_show_tabs(len(self.get_pages()) > 1)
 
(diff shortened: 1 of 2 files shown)


grc/gui/MainWindow.py
```diff
@@ -361,6 +361,10 @@ def update(self):
                 foreground='black' if page.saved else 'red', ro=' (ro)' if page.get_read_only() else '',
                 title=Utils.encode(file_name or NEW_FLOGRAPH_TITLE),
             ))
+            fpath = page.file_path
```
nickoe, line 364: Can't we use `page.file_path` directly here? Do we really need the `fpath`variable?

I am not against it, just wondering.

mormj: in general i'm not a fan of unnecessary intermediate variables, but it looks like in this case it might be desired to not modify page.file_path?

japm48: @nickoe I don't want to modify `file_path`, as I it is used elsewhere (`file_path == ''` means it is a created flowgraph that was not saved).
Of course you can do this: `page.set_tooltip(page.file_path if page.file_path else '(unsaved)')`
But I think it's too cumbersome.

nickoe: Ah yeah, right. I agree.

PR #3306, opened 2020-04-03
digital: restructure equalizers and add dfe

mormj:
This commit restructures the linear equalizer to have a separate
specifiable adaptive algorithm.  Generally this works the same as the
previous LMS and CMA decision directed equalizers, but also adds the
ability to equalize using training sequences as well.

Also, a Decision Feedback Equalizer structure is added

This is a rebase of #3224 that got stale

This follows the FOSDEM talk:
https://fosdem.org/2020/schedule/event/fsr_hannel_equalization_using_gnu_radio/

The main changes are:

    Separate the equalization structure from the adaptive algorithm
    -- e.g. the previous lms_dd_equalizer becomes linear_equalizer with an argument of an adaptive_algorithm_lms object
    Add training based equalization
    -- In this way we can specify a preamble and that used to more quickly adapt the taps
    -- After the data-aided training, the equalizer can be switched into decision directed mode

Some things from the out-of-tree implementation have not been added in yet, namely RLS because it requires (currently) a libarmadillo dependency. Refactoring this to do the matrix math with existing libraries will be a fairly easy change, but left it off for now.


mormj:
@ThomasHabets - thanks for the guidance on modern c++ - it's great to get the codebase up to par and your recommendations are much appreciated.  Will clean this up shortly.

gr-digital/lib/linear_equalizer_impl.cc
```diff
@@ -0,0 +1,232 @@
...
+namespace digital {
+
+linear_equalizer::sptr linear_equalizer::make(unsigned num_taps,
+                                              unsigned sps,
+                                              adaptive_algorithm_sptr alg,
+                                              bool adapt_after_training,
+                                              std::vector<gr_complex> training_sequence,
+                                              const std::string& training_start_tag)
+{
+    return gnuradio::get_initial_sptr(new linear_equalizer_impl(
```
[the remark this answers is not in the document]
mormj: I'd prefer to keep this consistent with the rest of the blocks in the codebase for now and make the switch over the make_shared a different PR for all the blocks

gr-digital/lib/decision_feedback_equalizer_impl.cc
```diff
@@ -0,0 +1,267 @@
...
+    set_history(d_num_taps_fwd);
+
+    alg->initialize_taps(d_new_taps);
+    // NOTE: the filter kernel reverses the taps internally
+    filter::kernel::fir_filter_ccc::set_taps(d_new_taps);
+}
+
+void decision_feedback_equalizer_impl::update_decision_history(gr_complex decision)
+{
+    std::rotate(d_decision_history.begin(),
```
ThomasHabets, line 100: This is called a lot, in a loop from `equalize()`, and in turn in from `work()`. `d_decision_history` is a vector. That's a lot of data copying, isn't it?

Did you consider a `std::queue` for this instead? Just `d_decision_history.pop(); d_decision_history.push(…);`?

Maybe I'm missing something.

gr-digital/lib/decision_feedback_equalizer_impl.cc
```diff
@@ -0,0 +1,267 @@
...
+        gr::thread::scoped_lock guard(d_mutex);
+        d_taps = d_new_taps;
+        set_history(d_taps.size());
+        d_updated = false;
+        return 0; // history requirements may have changed.
+    }
+
+    unsigned long int nread = nitems_read(0);
+    vector<tag_t> tags;
+    vector<unsigned int> training_start_samples;
```
ThomasHabets, line 134: Probably want to do `training_start_samples.reserve(tags.size())` before looping over the tags. If you think it may affect performance in a good way (because zeroing the values it may make it worse) maybe go so far as to create it at the correct size to start with, and set the values by index instead of `push_back`.

gr-digital/lib/decision_feedback_equalizer_impl.cc
```diff
@@ -0,0 +1,267 @@
...
+
+    unsigned short* state = nullptr;
+    gr_complex* taps = nullptr;
+
+    if (outlen > 1)
+        taps = static_cast<gr_complex*>(output_items[1]);
+    if (outlen > 2)
+        state = static_cast<unsigned short*>(output_items[2]);
+
+    if (d_updated) {
```
ThomasHabets, line 124: `d_updated` is here read without holding a lock, thus is technically a data race and therefore undefined behaviour, I believe.

Maybe switch type to `std::atomic_flag`?

gr-digital/lib/decision_feedback_equalizer_impl.cc
```diff
@@ -0,0 +1,267 @@
...
+
+    const gr_complex* samples;
+    if (history_included) {
+        samples = input_samples;
+    } else {
+        vector<gr_complex> in_prepended_history(num_inputs + d_num_taps_fwd - 1);
+        std::copy(input_samples,
+                  input_samples + num_inputs,
+                  in_prepended_history.begin() + d_num_taps_fwd - 1);
+        samples = in_prepended_history.data();
```
ThomasHabets, line 188: dangling pointer. When the vector is destroyed the line below `samples` points to unallocated data.

gr-digital/lib/decision_feedback_equalizer_impl.cc
```diff
@@ -0,0 +1,267 @@
...
+    unsigned int tag_index = 0;
+
+    int j = 0;
+
+    for (unsigned i = 0; i < nout; i++) {
+        std::copy(&samples[j + d_sps - 1],
+                  &samples[j + d_sps - 1] + d_num_taps_fwd,
+                  d_filt_buf.data() + d_num_taps_rev);
+        std::copy(
+            d_decision_history.begin(), d_decision_history.end(), d_filt_buf.data());
```
ThomasHabets, line 200: why `.data()` and not `.begin()`?

(also elsewhere, e.g. three lines up)

gr-digital/lib/decision_feedback_equalizer_impl.cc
```diff
@@ -0,0 +1,267 @@
...
+        vector<gr_complex> in_prepended_history(num_inputs + d_num_taps_fwd - 1);
+        std::copy(input_samples,
+                  input_samples + num_inputs,
+                  in_prepended_history.begin() + d_num_taps_fwd - 1);
+        samples = in_prepended_history.data();
+    }
+
+    unsigned int tag_index = 0;
+
+    int j = 0;
```
ThomasHabets, line 193: maybe a comment what `j` is?

gr-digital/lib/linear_equalizer_impl.cc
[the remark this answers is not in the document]
ThomasHabets: 👍

gr-digital/lib/linear_equalizer_impl.cc
```diff
@@ -0,0 +1,231 @@
...
+    if (history_included) {
+        samples = input_samples;
+    } else {
+        // When process() is called outside the scheduler, allow zeros to be added
+        // as history
+        vector<gr_complex> in_prepended_history(num_inputs + d_num_taps - 1);
+        std::copy(input_samples,
+                  input_samples + num_inputs,
+                  in_prepended_history.begin() + d_num_taps - 1);
+        samples = in_prepended_history.data();
```
ThomasHabets, line 120: same. dangling pointer.

gr-digital/lib/linear_equalizer_impl.cc
```diff
@@ -0,0 +1,231 @@
...
+    }
+
+    unsigned int tag_index = 0;
+
+    unsigned int j = 0;
+    size_t k, l = d_taps.size();
```
ThomasHabets, line 126: Move `k` to the for loop where it's used.

gr-digital/lib/linear_equalizer_impl.cc
```diff
@@ -0,0 +1,231 @@
...
+
+    unsigned short* state = nullptr;
+    gr_complex* taps = nullptr;
+
+    if (outlen > 1)
+        taps = static_cast<gr_complex*>(output_items[1]);
+    if (outlen > 2)
+        state = static_cast<unsigned short*>(output_items[2]);
+
+    if (d_updated) {
```
ThomasHabets, line 203: Again: undefined behaviour.

In practice probably fine, but many hairs have been pulled in the past from "surely my arch will do the right thing here?".

gr-digital/lib/meas_evm_cc_impl.cc
```diff
@@ -0,0 +1,72 @@
...
+meas_evm_cc_impl::meas_evm_cc_impl(constellation_sptr cons, evm_measurement_t meas_type)
+    : gr::sync_block("meas_evm_cc",
+                     gr::io_signature::make(1, 1, sizeof(gr_complex)),
+                     gr::io_signature::make(1, 1, sizeof(float))),
+      d_cons(cons),
+      d_cons_points(d_cons->points()),
+      d_meas_type(meas_type)
+{
+    // Calculate the average power of the constellation
+    float sum = 0.0;
```
ThomasHabets, line 35: Why not:

```c++
const float d_cons_mag_sq = std::accumulate(d_cons_points.begin(), d_cons_points.end(), 0.0, [](float s, gr::complex e)->float{
  return s+std::norm(e);
}) / d_cons_points.size();
d_cons_mag = sqrt(d_cons_mag_sq);
```

gr-digital/include/gnuradio/digital/adaptive_algorithm.h
```diff
@@ -0,0 +1,90 @@
...
+#include <math.h>
+#include <algorithm>
+#include <vector>
+
+namespace gr {
+namespace digital {
+enum class adaptive_algorithm_t { LMS, NLMS, CMA };
+
+class adaptive_algorithm;
+typedef std::shared_ptr<adaptive_algorithm> adaptive_algorithm_sptr;
```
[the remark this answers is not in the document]
mormj: For this class I copied the way it was done in constellation.h where the base class had a unique *_sptr perhaps because it wouldn't be instantiated directly?

gr-digital/include/gnuradio/digital/adaptive_algorithm_cma.h
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+public:
+    static sptr make(constellation_sptr cons, float step_size, float modulus)
+    {
+        return adaptive_algorithm_cma::sptr(
+            new adaptive_algorithm_cma(cons, step_size, modulus));
+    }
+
+    gr_complex error(const gr_complex& out) const
```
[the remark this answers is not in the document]
mormj: Ah yes that is a good point.

gr-digital/lib/meas_evm_cc_impl.cc
ThomasHabets: Oops, drop the `const float`, of course.

gr-digital/lib/decision_feedback_equalizer_impl.cc
mormj: d_decision_history gets copied into the filter buffer (along with the incoming samples) which needs direct element access.  I thought queues didn't have iterators so there was no way to copy a queue to an array.   I can see what I've implemented there is a queue, but don't see an easy way around it.

ThomasHabets: Doh, of course. Just switch to `std::deque`, which is the default container class for `std::queue` anyway, but has iterators. Just use `push_back` instead of `push` in my example.

gr-digital/include/gnuradio/digital/adaptive_algorithm.h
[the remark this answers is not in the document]
ThomasHabets: I don't know. Seems like a style issue more than anything, and not one I can answer.

gr-digital/lib/decision_feedback_equalizer_impl.cc
mormj: Not sure about std::atomic_flag - looks like it might complicate the code, but I will protect d_updated

ThomasHabets: Yeah atomics can be subtle, so usually it's better to use a mutex.

gr-digital/lib/decision_feedback_equalizer_impl.cc
mormj: Would you recommend a copy into a std::vector here?

ThomasHabets: I'm not sure what `samples` is supposed to be, here. How long, and containing what data. My DSP algorithm skills are holding me back on understanding the intention.

E.g. `samples` if history is not included, appears to always get a history of `d_num_taps_fwd - 1` zeroes, and I don't know what that means, including if that means input samples will be dropped.

But in terms of keeping that behaviour, without the dangling pointer, maybe:

```c++
std::unique_ptr<std::vector<gr_complex>>  in_prepended_history;
if (history_samples) {
  samples = input_samples;
} else {
      in_prepended_history = boost::make_unique<std::vector<gr_complex>>(num_inputs + d_num_taps_fwd - 1);
        std::copy(input_samples,
                  input_samples + num_inputs,
                  in_prepended_history->begin() + d_num_taps_fwd - 1);
        samples = in_prepended_history->data();
}
```

But of course the copying could be skipped by creating a simple wrapper that acts like a `vector` but gives zero for anything below `N`. Then it'd be:

```c++
wrap_thing samples(input_samples);
if (!history_included) {
  samples.set_N(d_num_taps_fwd);
}
```

But maybe I missed something.

mormj: Still getting used to smart pointers - but I like this suggestion.  

The in_prepended_history is for the case that equalize() is not being called from the work() function, and instead is a buffer of samples that doesn't relate to the previous buffer from a continuous stream.  But the filter routines work by doing dot products on the samples prior to the first new sample (hence the history).  In the case of no work function, there is no history, and I'm not maintaining it here.  So for the filter to not 1) skip samples or 2) run over invalid data, I decided to just prepad with zeros.

mormj:
@ThomasHabets - a sincere thank you for all the feedback on this PR.  I learned quite a bit and was alerted to some potentially dangerous code.

gr-digital/lib/decision_feedback_equalizer_impl.cc
```diff
@@ -0,0 +1,267 @@
...
+                         sps),
+      filter::kernel::fir_filter_ccc(
+          sps,
+          vector<gr_complex>(num_taps_forward + num_taps_feedback, gr_complex(0, 0))),
+      d_num_taps_fwd(num_taps_forward),
+      d_num_taps_rev(num_taps_feedback),
+      d_sps(sps),
+      d_alg(alg),
+      d_adapt_after_training(adapt_after_training),
+      d_training_sequence(training_sequence),
```
ThomasHabets, line 74: `d_training_sequence(std::move(training_sequence))`

gr-digital/lib/decision_feedback_equalizer_impl.cc
```diff
@@ -0,0 +1,267 @@
...
+                                  std::vector<gr_complex> training_sequence,
+                                  const std::string& training_start_tag)
+{
+    return gnuradio::get_initial_sptr(
+        new decision_feedback_equalizer_impl(num_taps_forward,
+                                             num_taps_feedback,
+                                             sps,
+                                             alg,
+                                             adapt_after_training,
+                                             training_sequence,
```
ThomasHabets, line 41: `std::move(training_sequence)`

Removes one copy.
(when a function takes an object that will *always* be copied, then passing it by value (like is already done) is better than by reference.

A smart compiler *may* already see that it can move, here, but adding `std::move` forces it.

You can make the same case for `training_start_tag` being by value instead of constref.

The total reduction of copies if passed by value over constref when callee always makes a copy is when caller calls with an rvalue. Then there will be zero copies in total.

gr-digital/lib/decision_feedback_equalizer_impl.cc
```diff
@@ -0,0 +1,267 @@
...
+      d_decision_history(num_taps_feedback),
+      d_new_taps(num_taps_forward + num_taps_feedback),
+      d_updated(false),
+      d_training_sample(0),
+      d_filt_buf(num_taps_forward + num_taps_feedback)
+{
+    // not reversed at this point
+    d_new_taps[0] = gr_complex(1.0, 0.0); // help things along with initialized taps
+
+    if (training_start_tag == "" || training_sequence.empty()) {
```
ThomasHabets, line 86: if you change to move out of `training_start_tag`, then of course use `d_training_start_tag` here.

gr-digital/lib/decision_feedback_equalizer_impl.cc
```diff
@@ -0,0 +1,267 @@
...
+            d_updated = false;
+            return 0; // history requirements may have changed.
+        }
+    }
+
+    auto in = static_cast<const gr_complex*>(input_items[0]);
+    auto out = static_cast<gr_complex*>(output_items[0]);
+
+    int outlen = output_items.size();
+    unsigned short* state = nullptr;
```
ThomasHabets, line 123: I'd move `state`, `taps`, and `outlen` declaration and setting below the tag work (line 140), to make it clear to the reader they're not used in that block of code.

gr-digital/lib/linear_equalizer_impl.cc
```diff
@@ -0,0 +1,233 @@
...
+
+    if (outlen > 1)
+        taps = static_cast<gr_complex*>(output_items[1]);
+    if (outlen > 2)
+        state = static_cast<unsigned short*>(output_items[2]);
+
+    unsigned long int nread = nitems_read(0);
+    vector<tag_t> tags;
+    get_tags_in_window(
+        tags, 0, 0, noutput_items * decimation(), pmt::intern(d_training_start_tag));
```
ThomasHabets, line 215: `d_training_start_tag` is const, and `pmt::intern` seem like nontrivial work (caching interned strings). Better to intern the string in the constructor?

gr-digital/lib/meas_evm_cc_impl.cc
```diff
@@ -0,0 +1,72 @@
...
+meas_evm_cc_impl::meas_evm_cc_impl(constellation_sptr cons, evm_measurement_t meas_type)
+    : gr::sync_block("meas_evm_cc",
+                     gr::io_signature::make(1, 1, sizeof(gr_complex)),
+                     gr::io_signature::make(1, 1, sizeof(float))),
+      d_cons(cons),
+      d_cons_points(d_cons->points()),
+      d_meas_type(meas_type)
+{
+    // Calculate the average power of the constellation
+    float d_cons_mag_sq =
```
ThomasHabets, line 36: remove the `float`, so that member variable is actually set. That was my mistake in my previous comment. sorry.

gr-digital/lib/meas_evm_cc_impl.h
```diff
@@ -0,0 +1,41 @@
...
+#include <gnuradio/digital/meas_evm_cc.h>
+
+namespace gr {
+namespace digital {
+
+class meas_evm_cc_impl : public meas_evm_cc
+{
+private:
+    constellation_sptr d_cons;
+    std::vector<gr_complex> d_cons_points;
```
ThomasHabets, line 23: Is `d_cons_points` never changed? Const, maybe?

gr-digital/lib/decision_feedback_equalizer_impl.cc
mormj: I like the suggestion but think I'm going to avoid this for now since it's not a lot of data, and I don't see this pattern (yet) elsewhere in the codebase.

dkozel:
The code changes looks good to me. The block documentation on the wiki needs to be updated (marking deprecated and preferably an example of migrating to the newer blocks) and new pages made for the new blocks.

https://wiki.gnuradio.org/index.php/Kurtotic_Equalizer
https://wiki.gnuradio.org/index.php/LMS_DD_Equalizer
https://wiki.gnuradio.org/index.php/CMA_Equalizer

mormj:
Thanks @dkozel - wiki pages are in place now for the equalizers, and the old blocks marked as deprecated.  Hopefully we can get this PR merged before it goes stale again.

gr-digital/include/gnuradio/digital/decision_feedback_equalizer.h
```diff
@@ -0,0 +1,103 @@
...
+     * each sample of the output
+     * \param state           Optional output state of the equalizer for debug {IDLE,
+     * TRAINING, DD}
+     */
+    virtual int equalize(
+        const gr_complex* input_samples,
+        gr_complex* output_symbols,
+        unsigned int num_inputs,
+        unsigned int max_num_outputs,
+        std::vector<unsigned int> training_start_samples = std::vector<unsigned int>(0),
```
ThomasHabets, line 94: `(0)` means a vector of size 0, so why not use empty parentheses as elsewhere?

gr-digital/include/gnuradio/digital/linear_equalizer.h
```diff
@@ -0,0 +1,104 @@
...
+     * each sample of the output
+     * \param state           Optional output state of the equalizer for debug {IDLE,
+     * TRAINING, DD}
+     */
+    virtual int equalize(
+        const gr_complex* input_samples,
+        gr_complex* output_symbols,
+        unsigned int num_inputs,
+        unsigned int max_num_outputs,
+        std::vector<unsigned int> training_start_samples = std::vector<unsigned int>(0),
```
ThomasHabets, line 95: same, why not `()`

gr-digital/lib/decision_feedback_equalizer_impl.cc
```diff
@@ -0,0 +1,266 @@
...
+                                           gr_vector_const_void_star& input_items,
+                                           gr_vector_void_star& output_items)
+{
+    {
+        gr::thread::scoped_lock guard(d_mutex);
+        if (d_updated) {
+            d_taps = d_new_taps;
+            set_history(d_taps.size());
+            d_updated = false;
+            return 0; // history requirements may have changed.
```
ThomasHabets, line 115: hm… so after putting this near the top, why does it matter if history requirements have changed?

PR #3310, opened 2020-04-05
Remove warnings in gr-blocks and gr-network.

drmpeg:


gr-network/lib/tcp_sink_impl.cc
```diff
@@ -219,9 +219,6 @@ bool tcp_sink_impl::stop()
 
 void tcp_sink_impl::check_for_disconnect()
 {
-    char buff[1];
```
mbr0wn, line 222: Is this correct? Is the `receive()` call maybe necessary to update `ec`?

Also, when is one supposed to call this method, and do we need it? It never gets called, and has no documented contract.

drmpeg: Thanks for catching this. Also, the receive() call blocks, so probably not good in general. Seems like the function should be deleted, but maybe @ghostop14 can comment.

michaelld: Looks like the call to `receive()` is meant to check whether the socket is still open (at all) which will then set the error code `ec`: https://www.boost.org/doc/libs/1_72_0/doc/html/boost_asio/reference/basic_stream_socket/receive/overload3.html . So ... seems like removing this code is a bad idea ...

ghostop14:
Hey folks, yeah the receive has to stay to detect if the socket has been closed or not.

gr-network/lib/tcp_sink_impl.cc
drmpeg: I've put the code back in. But the question still remains as to whether the entire check_for_disconnect() method should be deleted.

drmpeg:
@ghostop14 What is check_for_disconnect() for if it never gets called?

ghostop14:
Ah, yeah.  Looks like I moved that functionality directly to the work function and checked the ec return values there rather than calling the function.  You can remove check_for_disconnect().

Issue #3309, opened 2020-04-04
gr-filter: fir_filter kernels are not thread-safe


jacobgilbert-snl:
Because the FIR filter kernels store output data temporarily in member variables (see: https://github.com/gnuradio/gnuradio/blob/master/gr-filter/lib/fir_filter.cc#L130) use of this is not safe within multithreaded blocks. Simple fix is to use local variables.

This issue was observed with this OOT module which uses omp for multithreading: https://github.com/sandialabs/gr-fhss_utils/blob/master/lib/tagged_burst_to_pdu_impl.cc

jacobgilbert-snl:
Possibly redundant to https://github.com/gnuradio/gnuradio/pull/3062

mrjacobagilbert:
So not being an expert in volk, I'm not sure our solution of using local variables is the best option as they won't necessarily be VOLK aligned.

mrjacobagilbert:
Not really a GR problem, will address in OOT.

PR #3315, opened 2020-04-06
use local variables in fir_filter

mrjacobagilbert:
Addresses https://github.com/gnuradio/gnuradio/issues/3309 at the expense of not using output aligned output variables to the dot product VOLK kernel, though this does not appear to be necessary given how VOLK returns data: https://github.com/gnuradio/volk/blob/master/kernels/volk/volk_32fc_32f_dot_prod_32fc.h#L176

This enables the use of omp multithreading (e.g.: https://github.com/sandialabs/gr-fhss_utils/blob/master/lib/tagged_burst_to_pdu_impl.cc).


mrjacobagilbert:
after discussing w/ @awalls-cx18 on slack, there is not a good way to make fir_filter re-entrant without performance penalties so we will address this within the OOT directly.

PR #3317, opened 2020-04-07
allow GRC to use module specific variables in assert statements

mrjacobagilbert:
Currently GRC will not evaluate asserts that contain objects from imported modules, because the assert check is performed without using the parent flowgraph namespace and the objects do not exist there. This is not a significant in-tree issue, but it is an issue for several OOTs and I believe one check that was commented out [here](https://github.com/gnuradio/gnuradio/blob/master/gr-blocks/grc/blocks_multiply_matrix_xx.block.yml#L40):

image

This fix explicitly uses the parent flowgraph namespace and allows for checks using module objects.

Credit: Peter Knee

### DIFF
--- a/grc/core/blocks/block.py
+++ b/grc/core/blocks/block.py
@@ -130,6 +130,19 @@ class Block(Element):
         self.active_sources = [p for p in self.sources if not p.hidden]
         self.active_sinks = [p for p in self.sinks if not p.hidden]
 
+        # namespaces may have changed, update them
+        self.block_namespace = {}
+        try:
+            exec(self.templates.render('imports'), self.block_namespace)
+        except ImportError:
+            # We do not have a good way right now to determine if an import is for a
+            # hier block, these imports will fail as they are not in the search path
+            # this is ok behavior, unfortunately we could be hiding other import bugs
+            pass
+        except Exception:
+            log.exception('Failed to evaluate import expression "{0}"'.format(expr), exc_info=True)
+            pass
+
     def update_bus_logic(self):
         ###############################
         ## Bus Logic
@@ -581,7 +594,9 @@ class Block(Element):
     ##############################################
     @property
     def namespace(self):
-        return {key: param.get_evaluated() for key, param in six.iteritems(self.params)}
+        # update block namespace
+        self.block_namespace.update({key:param.get_evaluated() for key, param in six.iteritems(self.params)})
+        return self.block_namespace
 
     @property
     def namespace_templates(self):


mrjacobagilbert:
Also, let me know if I should re-enable that check here as part of the PR if commenting it out was simply because it did not evaluate.

skoslowski:
Wouldn't it make more sense to have just the imports of that block
available?

mrjacobagilbert:
Seems like that would work too -- that said, I'm not quite sure how to do that.

skoslowski:
To implement this, one would
- need to render the imports (`self.templates.render('imports')`)
- execute them (eval) and save the resulting dict in the instance
- use that as dict as base (...copy().update()) for the blocks namespace property

mrjacobagilbert:
Incorporating what Sebastian suggested, please don't merge this yet.

skoslowski:
Also note, global and local namespaces are switch here. Flowgraph-wide names could override block-locals...

mrjacobagilbert:
@skoslowski I think the changes i just pushed are more inline with what you are recommending

This will cause the in tree assert to evaluate (although you have to first also import gr in the multiply by matrix yaml file so that gr.TPP_ONE_TO_ONE exists) however there is another issue lurking here where enum parameters don't actually evaluate in assert statements, and remain whatever the default value is, so you won't actually be able to see the assert fail on this block... it's unrelated and i'll open a separate issue for it; no fix yet but looking into it.

mrjacobagilbert:
this can be merged from my perspective.

skoslowski:
I think this is ready to be merged too. Great work!

One caveat that I can see, is that the block namespace is only populated once on __init__. In practice, this will suffice in nearly all cases. However, technically, the imports are a template and can therefore change. So technically, this code would have to be executed in rewrite(). What do you think?

mormj:
removing the don't merge tag based on @skoslowski comments

mrjacobagilbert:
@skoslowski getting back to this - i agree with this. the operation is very quick and moving it somewhere that is called anytime blocks change is best. this will ensure that the appropriate imports for blocks that are used will be available.

skoslowski:
Thanks @mrjacobagilbert

drmpeg:
This commit creates a spectacular bug for me. Here's the flow graph.

http://www.w6rz.net/amfft.grc

Steps to reproduce.

1) Load the flow graph. gnuradio-companion amfft.grc

image

2) Disable the QT Time Sink (with one input), Multiply and Signal Source blocks.

image

3) Generate the flow graph.

4) Exit GRC.

5) Load the flow graph. gnuradio-companion amfft.grc

Many connections are lost.

image

skoslowski:
Thanks. My bad, fix pushed

jacobgilbert-snl:
Whoa...very odd. Thanks for getting this fixed. This is probably an issue on maint-3.8.

skoslowski:
@mrjacobagilbert, don't worry, I should've caught this in the review. :shrug: 
maint-3.8 is now fixed, too. Thanks for the reminder

PR #3318, opened 2020-04-07
Block gateway add alias affinity

mrjacobagilbert:
Both `set_block_alias()` and `set_processor_affinity()` exist as GUI settable fields for all blocks, but for python blocks these are not exposed causing runtime exceptions if used the way GRC does (can only be used via `blockname.to_basic_block().set_block_alias()`). This PR adds both to block_gateway objects, and allows python blocks to use `self.alias()` to return their alias.

### DIFF
--- a/gnuradio-runtime/include/gnuradio/block_gateway.h
+++ b/gnuradio-runtime/include/gnuradio/block_gateway.h
@@ -116,6 +116,18 @@ public:
 
     void block__set_history(unsigned history) { return gr::block::set_history(history); }
 
+    void block__set_block_alias(std::string alias)
+    {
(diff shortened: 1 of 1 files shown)


michaelld:
> Looks ok to me, but don't we want to prefix the commit title with "`runtime: `"?

Correct! @mrjacobagilbert please update the commit message to start with `runtime:` ... then we'll merge this.

michaelld:
All CI passes. multiple approvals. Useful and desirable feature. Merging.

michaelld:
Do we want a backport to maint-3.8? I'm thinking so ...

mormj:
Yes, I can do the cherry-pick PR right now

PR #3320, opened 2020-04-08
pmt: speed up serialization

mrjacobagilbert:
This PR substantially improves serialization speed for PMTs, which was done primarily to facilitate movement of large PDUs over ZMQ.

U8 PDU speedup: ~1.8x
F32 PDU speedup: ~2.1x
C32 PDU Speedup: ~2.2x

Needs review, especially for non-x86-linux platforms as this has not been extensively tested on anything but...

### DIFF
--- a/gnuradio-runtime/lib/pmt/pmt_serialize.cc
+++ b/gnuradio-runtime/lib/pmt/pmt_serialize.cc
@@ -14,6 +14,7 @@
 
 #include "pmt/pmt_serial_tags.h"
 #include "pmt_int.h"
+#include <endian.h>
 #include <pmt/pmt.h>
 #include <limits>
 #include <vector>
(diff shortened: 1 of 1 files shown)


drmpeg:
I have an armv7-a (32-bit) platform that I can test on. Is "make test" sufficient, or do you have some test flow graphs?

mrjacobagilbert:
I have some python i use for benchmarking and accuracy outside of whatever QA is included for PMT serialization (which is reasonable: https://github.com/gnuradio/gnuradio/blob/master/gnuradio-runtime/python/pmt/qa_pmt.py). Standby...

mrjacobagilbert:
Accuracy: https://pastebin.com/8kg0Hn6u
Speed: https://pastebin.com/XDSTaZW2

jacobgilbert-snl:
Thanks for running that; consistent to what i see on x86-64 linux.

mormj:
@balister - do you want to weigh in on this PR for embedded platforms - any concern?

gnuradio-runtime/lib/pmt/pmt_serialize.cc
```diff
@@ -26,25 +27,23 @@ static pmt_t parse_pair(std::streambuf& sb);
...
     return sb.sputc((i >> 0) & 0xff) != std::streambuf::traits_type::eof();
 }
 
 // always writes big-endian
-static bool serialize_untagged_u16(unsigned int i, std::streambuf& sb)
+static bool serialize_untagged_u16(uint16_t i, std::streambuf& sb)
 {
-    sb.sputc((i >> 8) & 0xff);
-    return sb.sputc((i >> 0) & 0xff) != std::streambuf::traits_type::eof();
+    i = htobe16(i);
```
marcusmueller, line 38: In the sense of "include what you use", `#include <endian.h>` (I wonder what transitively includes that) would be right!

gnuradio-runtime/lib/pmt/pmt_serialize.cc
```diff
@@ -15,6 +15,7 @@
 #include "pmt/pmt_serial_tags.h"
 #include "pmt_int.h"
 #include <pmt/pmt.h>
+#include <iostream>
```
marcusmueller, line 18: slight confusion: what's that for? Do you mean to include `streambuf`? (which would be very appropriate)

marcusmueller:
> Unfortunately, qa_pmt always fails on 32-bit systems because it's using from_long to test for 64-bit numbers. But that's a side issue.

Aargh, that needs to be fixed. Actually, `from_long` is an architectural bug. A serialization format whose datatype depends on the host...

mrjacobagilbert:
iostream must be a holdover. streambuf seems used via transitive inclusion throughout the core of GR so i wasnt going to put it here (by all means tell me otherwise and i will).

I will add endian.h; that is a good idea and i don't know how it got pulled in...maybe from boost? Either way it should be in here.

marcusmueller:
yeah, ICWYU is a good pattern; streambuf: you're right, if anywhere, the header to this file must include that. I would have included it in the implementation myself, too, but I've far too weak an opinion on that to enforce this arguably superfluous include.

marcusmueller:
pinging @SpectreJan, the maintainer of standalone libpmt for external usage.

japm48:
Sorry for the bad news, but I believe `endian.h` does not exist in Windows.

michaelld:
The `endian.h` issue is fixed by #3369 .

PR #3321, opened 2020-04-08
grc: add hotkeys to clear/save console, and to toggle autoscroll

mrjacobagilbert:
Adds the following GRC hotkeys:
<ctrl><shift>C = clear console
<ctrl><shift>P = save the console to file
<ctrl><shift>L = toggle console scroll lock

Debatable if these are the best bindings, but specifically clearing the console is very helpful to have a hotkey for. Copying console to clipboard would be awesome too but that was not immediately apparent and I do not know much about how to implement that....

### DIFF
--- a/grc/gui/Actions.py
+++ b/grc/gui/Actions.py
@@ -480,7 +480,8 @@ TOGGLE_BLOCKS_WINDOW = actions.register("win.toggle_blocks_window",
 TOGGLE_SCROLL_LOCK = actions.register("win.console.scroll_lock",
     label='Console Scroll _Lock',
     tooltip='Toggle scroll lock for the console window',
-    preference_name='scroll_lock'
+    preference_name='scroll_lock',
+    keypresses=["Scroll_Lock"],
 )
 ABOUT_WINDOW_DISPLAY = actions.register("app.about",
     label='_About',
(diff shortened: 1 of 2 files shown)


mormj:
I love the idea of more keyboard shortcuts.  Is there a basis for the chosen keys - i.e. are they used in other common applications?  L makes sense to me, but not sure about C and P

mrjacobagilbert:
Not a very good basis, but:
ctrl-shift-C: **C**lear **C**onsole
ctrl-shift-L: Scroll **L**ock << Was going to use 'A' for **A**utoscroll but it is named 'Console Scroll Lock'.
ctrl-shift-P: currently ctrl-p saves output to file (screenshot) so this was augmenting that.

I'm certainly open to other suggestions.

dkozel:
Hi @mrjacobagilbert Can you please add the shortcuts to the help dialog?
https://github.com/gnuradio/gnuradio/blob/master/grc/gui/Dialogs.py#L298

Standard Gnome terminal uses Ctrl-L for clear console.
What about Ctrl-Shift-P for print console? I'd prefer not to overload/extend the printscreen shortcut.
For scroll lock, how about `Scroll_Lock`?

mrjacobagilbert:
@dkozel ctrl-shift-p is used in this PR to save console to file, not printscreen. I was just referencing that ctrl-p is currently used to save a screenshot of the flowgraph to file, and this is similar.

CTRL-L makes sense for clear console, so does scroll lock.

I'll get the Dialogs.py updated and make the changes to bindings.

mrjacobagilbert:
Ok, updated bindings and the help dialog for hotkeys. I did not add scroll lock as that is really something that 'works' now.

PR #3331, opened 2020-04-10
runtime: Make random seed default to current time, as documented

ThomasHabets:
This is what gnuradio currently documents seeding will do, but it's
not what C++ defaults to.

See C++11 26.5.3.2, paragraph 3. Default for mersenne twister is
constexpr 5489.

Or easier, the bottom of
https://en.cppreference.com/w/cpp/numeric/random/mersenne_twister_engine

### DIFF
--- a/gr-analog/lib/fastnoise_source_impl.cc
+++ b/gr-analog/lib/fastnoise_source_impl.cc
@@ -59,7 +59,8 @@ fastnoise_source_impl<T>::fastnoise_source_impl(noise_type_t type,
                  io_signature::make(0, 0, 0),
                  io_signature::make(1, 1, sizeof(T))),
       d_type(type),
-      d_ampl(ampl)
+      d_ampl(ampl),
+      d_rng(seed)
 {
     d_samples.resize(samples);
     xoroshiro128p_seed(d_state, (uint64_t)seed);
@@ -76,7 +77,8 @@ fastnoise_source_impl<gr_complex>::fastnoise_source_impl(noise_type_t type,
                  io_signature::make(0, 0, 0),
                  io_signature::make(1, 1, sizeof(gr_complex))),
       d_type(type),
-      d_ampl(ampl / sqrtf(2.0f))
+      d_ampl(ampl / sqrtf(2.0f)),
+      d_rng(seed)
 {
     d_samples.resize(samples);
     xoroshiro128p_seed(d_state, (uint64_t)seed);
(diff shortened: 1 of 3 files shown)


ThomasHabets:
Possibly want to backport this to 3.8?

ThomasHabets:
Hmm, actually let me improve this a bit, to use more time resolution than second boundary.

ThomasHabets:
Wow, improving that exposed another bug.

gr-analog/lib/fastnoise_source_impl.cc
```diff
@@ -76,7 +77,8 @@ fastnoise_source_impl<gr_complex>::fastnoise_source_impl(noise_type_t type,
                  io_signature::make(0, 0, 0),
                  io_signature::make(1, 1, sizeof(gr_complex))),
       d_type(type),
-      d_ampl(ampl / sqrtf(2.0f))
+      d_ampl(ampl / sqrtf(2.0f)),
```
marcusmueller, line 80: really doesn't matter at the rate at which this will be called, but as something I've learned: even GCC 9.3 with `-O3` will implement that as a division by a constant  ([godbolt](https://gcc.godbolt.org/z/pFzmYY)); `ampl * sqrtf(0.5f)` would be more efficient.

marcusmueller:
> Wow, improving that exposed another bug.

I like bugs when you find them: What is that bug?

gr-analog/lib/fastnoise_source_impl.cc
ThomasHabets: `sqrt(2.0)` to me is a known formula, thus makes it easier to understand where that magic number came from. But you prefer multiplying by `sqrt(0.5)`)

ThomasHabets:
> I like bugs when you find them: What is that bug?

First bug was that the current time was not actually used as the seed. So effective seed would be constant (as per standard). First I changed to use `time(nullptr)` (so only second resolution). For the unit tests that run in under a second, that's effectively constant for any given run. Changing to nanosecond resolution makes every generator get a unique seed no matter how tight the loop is. That exposed that the seed wasn't forwarded from the `xor…` generator to the `std::mt…` generator, so the latter would now use the ns-resolution time as seed, which for the unit test now would make it nondeterministic.

The unit test passes in a seed, that wasn't passed along.

marcusmueller:
Ah, nice!

gr-analog/lib/fastnoise_source_impl.cc
marcusmueller: yeah, you're right, as said, this is really not the place to replace a clear code by a slightly more performant one :)

PR #3335, opened 2020-04-10
runtime: Remove most manual memory management

ThomasHabets:
The remaining ones:
* `pmt_pool.cc`, which is a memory allocator so that makes sense
* the tricky and aptly named `sptr_magic.cc`.

### DIFF
--- a/gnuradio-runtime/lib/thread/thread_group.cc
+++ b/gnuradio-runtime/lib/thread/thread_group.cc
@@ -13,46 +13,36 @@
  */
 
 #include <gnuradio/thread/thread_group.h>
+#include <boost/make_unique.hpp>
 
 namespace gr {
 namespace thread {
 
 thread_group::thread_group() {}
 
-thread_group::~thread_group()
-{
-    // We shouldn't have to scoped_lock here, since referencing this
-    // object from another thread while we're deleting it in the
-    // current thread is going to lead to undefined behavior any
-    // way.
-    for (std::list<boost::thread*>::iterator it = m_threads.begin();
-         it != m_threads.end();
-         ++it) {
-        delete (*it);
-    }
-}
+thread_group::~thread_group() {}
 
 boost::thread* thread_group::create_thread(const boost::function0<void>& threadfunc)
 {
     // No scoped_lock required here since the only "shared data" that's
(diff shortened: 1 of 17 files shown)


gnuradio-runtime/lib/thread/thread_group.cc
```diff
@@ -72,21 +64,16 @@ void thread_group::remove_thread(boost::thread* thrd)
 void thread_group::join_all()
 {
     boost::shared_lock<boost::shared_mutex> guard(m_mutex);
-    for (std::list<boost::thread*>::iterator it = m_threads.begin();
-         it != m_threads.end();
-         ++it) {
-        (*it)->join();
+    for (auto& thrd : m_threads) {
```
marcusmueller, line 67: This is a memory-management-unrelated change, but honestly, with the change in data structure, keeping the old for-loop would've been stupid, :+1:

gnuradio-runtime/lib/vmcircbuf.cc
```diff
@@ -183,21 +183,15 @@ test_a_bunch(vmcircbuf_factory* factory, int n, int size, int* start_ptr, bool v
...
-        init_buffer(c[i], counter[i], size);
+        init_buffer(*c[i], counter[i], size);
         cum_size += size;
     }
 
     for (int i = 0; i < n; i++) {
         std::string msg =
             str(boost::format("test_a_bunch_%dx%s[%d]") % n % memsize(size) % i);
-        ok &= check_mapping(c[i], counter[i], size, msg.c_str(), verbose);
+        ok &= check_mapping(*c[i], counter[i], size, msg.c_str(), verbose);
```
marcusmueller, line 193: this will conflict with #3339, but oh well, we'll have to solve that then.

marcusmueller: resolved the conflict there.

gnuradio-runtime/lib/thread/thread_group.cc
ThomasHabets: Yeah the loop needed a minor change because of the type change anyway, otherwise I wouldn't have touched it here.

PR #3337, opened 2020-04-10
boost: fix deprecated use of math/common_factor.hpp

### LINKED ISSUE #2712: gr-digital: symbol_sync_ff_impl uses deprecated Boost header
```
In file included from /usr/include/boost/config/header_deprecated.hpp:18,
                 from /usr/include/boost/math/common_factor_ct.hpp:14,
                 from /usr/include/boost/math/common_factor.hpp:13,
                 from /buildbot/gnuradio/gr-digital/lib/symbol_sync_ff_impl.cc:28:
/usr/include/boost/math/common_factor_ct.hpp:16:1: note: #pragma message: This header is deprecated. Use <boost/integer/common_factor_ct.hpp> instead.
   16 | BOOST_HEADER_DEPRECATED("<boost/integer/common_factor_ct.hpp>");
      | ^~~~~~~~~~~~~~~~~~~~~~~
/usr/include/boost/math/common_factor_rt.hpp:13:1: note: #pragma message: This header is deprecated. Use <boost/integer/common_factor_rt.hpp> instead.
   13 | BOOST_HEADER_DEPRECATED("<boost/integer/common_factor_rt.hpp>");
      | ^~~~~~~~~~~~~~~~~~~~~~~
/usr/include/boost/math/common_factor.hpp:16:1: note: #pragma message: This header is deprecated. Use <boost/integer/common_factor.hpp> instead.
   16 | BOOST_HEADER_DEPRECATED("<boost/integer/common_factor.hpp>");
      | ^~~~~~~~~~~~~~~~~~~~~~~
```

Boost deprecations are the wurst, since Boost seldom cleanly documents _since_ when a new function is available from a certain file, so you end up in git-bisect, and then Ubuntu comes along and happily mixes versions of boost sublibraries. Often, only `#ifdefs` on Boost versions are a viable solution :/

A future-bound `#ifdef __cplusplus >= 201703L` using `std::lcm` might be cool when fixing this.

japm48:
Another little spring clean patch.

Remove deprecation warning and prefer using `std::{lcm,gcd}` to Boost.
Fixes #2712.

I'm not sure if `gnuradio-runtime` is the best place to put the new header.

### DIFF
--- a/gnuradio-runtime/include/gnuradio/CMakeLists.txt
+++ b/gnuradio-runtime/include/gnuradio/CMakeLists.txt
@@ -31,6 +31,7 @@ install(FILES
   gr_complex.h
   hier_block2.h
   high_res_timer.h
+  integer_math.h
   io_signature.h
   logger.h
   math.h
--- a/gnuradio-runtime/include/gnuradio/integer_math.h
+++ b/gnuradio-runtime/include/gnuradio/integer_math.h
@@ -1,0 +1,35 @@
+/* -*- c++ -*- */
+/*
+ * Copyright 2020 Free Software Foundation, Inc.
+ *
+ * This file is part of GNU Radio
+ *
+ * SPDX-License-Identifier: GPL-3.0-or-later
+ *
+ */
+#ifndef INCLUDED_GR_INTEGER_MATH_H
+#define INCLUDED_GR_INTEGER_MATH_H
+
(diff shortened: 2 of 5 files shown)


marcusmueller:
I like this!

I'd agree, runtime feels like a bit of a overly prominent place. But we don't have a better place. So: *shrug*.

PR #3341, opened 2020-04-11
CMake/SWIG: Remove SWIG flag superfluous since 3.0.12, suppress warning

marcusmueller:
SWIG complains about "Deprecated command line option: -modern. This
option is now always on.".

As traditional with SWIG, policy changes don't appear in any
documentation; with a bit of git blaming, I found the change to have
been introduced after SWIG 3.0.12.

### DIFF
--- a/cmake/Modules/GrSwig.cmake
+++ b/cmake/Modules/GrSwig.cmake
@@ -107,7 +107,11 @@ macro(GR_SWIG_MAKE name)
     list(INSERT GR_SWIG_INCLUDE_DIRS 0 "$<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}>")
 
     #setup the swig flags with flags and include directories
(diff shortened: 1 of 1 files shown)


drmpeg:
Tested with SWIG 4.0.1. Looks good!

japm48:
Thank goodness, that was super annoying.

drmpeg:
I almost submitted this same PR. I used this to determine the correct version cutoff.

http://www.swig.org/Release/CHANGES

marcusmueller:
@drmpeg  I wonder what was wrong with me when I couldn't find that document. Thanks! bookmarked for future reference :)

marcusmueller:
@japm48 @mbr0wn reminder that "warnings become useless due to clutter" is definitely an _issue_ and there should be nothing stopping us from noting such (and much smaller) issues in the bug tracker here. Especially in times of GSoC coming up, we've got lot of students hungry for small problems to get their fingers warm.

japm48:
I created #3349 after reading your last message.

PR #3342, opened 2020-04-11
pmt: speed up serialization - backport to maint 3.8

mrjacobagilbert:
This PR substantially improves serialization speed for PMTs, which was done primarily to facilitate movement of large PDUs over ZMQ.

U8 PDU speedup: ~1.8x
F32 PDU speedup: ~2.1x
C32 PDU Speedup: ~2.2x

Backport to maint-3.8 of #3320 and #3369

### DIFF
--- a/gnuradio-runtime/lib/pmt/pmt_serialize.cc
+++ b/gnuradio-runtime/lib/pmt/pmt_serialize.cc
@@ -26,6 +26,7 @@
 #include "pmt/pmt_serial_tags.h"
 #include "pmt_int.h"
 #include <pmt/pmt.h>
+#include <boost/endian/conversion.hpp>
 #include <limits>
 #include <vector>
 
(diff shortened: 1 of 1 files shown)


marcusmueller:
Not to appear mean, but: this takes a low priority, because 3.8 should mainly see bugfixes, and avoid breakage very much. I'd put this a bit on the backburner until usage on master shows we've not created some fancy cross 32/64 bit corner cases etc. 

@mrjacobagilbert or do you have an acute use case where having this on maint-3.8 was very very advantageous to you?

jacobgilbert-snl:
We do require this, but this PR is mostly to offer it if others would benefit since it doesn't break the API. That said, we have managed these changes well back into 3.7 and this is already in our private fork, so it's perfectly fine by me if you choose not to merge it; i appreciate the hesitancy.

dkozel:
Possibly issue of backwards-compatibility with the previous behavior for fc32.

Previous
```
pmt.serialize_str(pmt.init_f32vector(2,[1, 2]))
b'\n\x08\x00\x00\x00\x02\x01\x00?\xf0\x00\x00\x00\x00\x00\x00@\x00\x00\x00\x00\x00\x00\x00'
```
vs post change.
```
b'\n\x08\x00\x00\x00\x02\x01\x00?\x80\x00\x00@\x00\x00\x00'
```

jacobgilbert-snl:
yep, so this change correctly uses 4 bytes for 32-bit floats and 8 bytes for 32-bit complex values, however that is not the previous behavior. i'll force the 64 bit methods and get some absolute QA together...

marcusmueller:
Hey, as @jdemel noticed, there's breakage for existing applications; which is OK for master, but not on 3.8 :(

mrjacobagilbert:
Its a relatively simple fix in gr-uhd; I can either pull #3555 into this backport, or we can just forget about this for 3.8. I'm good w/ either way.

PR #3346, opened 2020-04-11
VOLK: update submodule pointer to v2.2.1

marcusmueller:
While we've not yet removed the VOLK submodule alltogether, at least make it
point to the current VOLK release. That was the whole point of having more
frequent VOLK releases.


marcusmueller:
Autonomously merging – the intention of keeping the master submodule counter up-to-date, at the very least within the same VOLK major, was agreed upon.

drmpeg:
Unfortunately, there's a couple of issues.

1) In CMakeLists.txt, VOLK_MIN_VERSION isn't really a minimum version. That is, cmake fails if the VOLK version is 2.2.0 or 2.2.1.

2) The CMakeLists.txt file(s) in gr-network need to be updated.

PR #3354, opened 2020-04-12
CMake: warn about VOLK versions not setting version, but try with them

marcusmueller:
I'm not proud of this workaround, but this should keep the GNU Radio
builds afloat, until we can fix the [issue](https://github.com/gnuradio/volk/issues/377) upstream.


marcusmueller:
Circumventing principle of four eyes: This should re-enable CI

PR #3356, opened 2020-04-13
peak_detector2: Fix wrong average in an a edge case

### LINKED ISSUE #3345: peak_detector2: The average might not be correct
I think I found a problem using the Peak Detector2 block.
The debug output, which should output the moving average of the input signal, is outputting something slightly different.
An input sample making `d_found` switch from `false` to `true` might be counted twice in the average, resulting in a wrong average.

rear1019:
This fixes #3345.

### DIFF
--- a/gr-blocks/lib/peak_detector2_fb_impl.cc
+++ b/gr-blocks/lib/peak_detector2_fb_impl.cc
@@ -90,7 +90,7 @@ int peak_detector2_fb_impl::work(int noutput_items,
                 d_found = true;
                 d_peak_val = std::numeric_limits<float>::min();
                 set_output_multiple(d_look_ahead);
-                return i;
+                return i + 1;
             }
         }
         return noutput_items;
--- a/gr-blocks/lib/peak_detector2_fb_impl.cc
+++ b/gr-blocks/lib/peak_detector2_fb_impl.cc
@@ -88,10 +88,17 @@ int peak_detector2_fb_impl::work(int noutput_items,
                 sigout[i] = d_avg;
             if (iptr[i] > d_avg * (1.0f + d_threshold_factor_rise)) {
                 d_found = true;
-                d_peak_val = iptr[i];
-                d_peak_ind = i;
                 set_output_multiple(d_look_ahead);
-                return i + 1;
+                /*
+                 * Consume any samples filtered so far except the last one (which
+                 * caused the threshold to be exceeded). As per documentation
+                 * in peak_detector2_fb.h this sample is part of the window where
+                 * the peak is to be searched. The next time work() is called
+                 * execution of the code jumps to the else-if clause below. There,
+                 * the first sample will be the last sample processed here (because
+                 * not consumed here).
+                 */
+                return i;
             }
         }
         return noutput_items;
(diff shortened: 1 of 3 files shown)


marcusmueller:
thanks! will merge that as soon as CI works; do you, or @Callyan, have an approach on what one would have to do to test this? I'd really would've liked to catch such mistakes in testing!

Callyan:
I think the fix suggested by @rear1019 might not take into account the peak if it is the first sample in input. I tried to fix this and added a test case for the debug output here : https://github.com/Callyan/gnuradio/commit/55b2d2864b2f7b5fd1aade4a13dc7c938c7c89ce

marcusmueller:
@Callyan that is awesome! I don't thing @rear1019 will be angry if you opened a "competitive" Pull Request. Is that something you could do?

Callyan:
OK that's done, here it is : #3358

rear1019:
@marcusmueller Not angry at all. The comment by @Callyan made me look into the documentation of `peak_detector2` contained in the public API header [1]:

> An average of the input signal
> is calculated (through a single-pole autoregressive
> filter) and when the value of the input signal goes
> over threshold_factor_rise*average, we assume we are 
> in the neighborhood of a peak. The block will then
> find the position of the maximum within a window of
> look_ahead samples starting at the point where the
> threshold was crossed upwards.

As far as I understand it, the first sample which causes the threshold to be exceeded is part of the window where the peak is searched. This means that this sample _must not_ be consumed: If the sample causing the exceeding of the threshold is the maximum value in the window (the peak), there is no way to correctly mark the peak in the output if this sample is consumed [2]. This is exactly the issue mentioned by Callyan. This also implies that both my and Callyan’s fixes are wrong [3]. I have updated this pull request with the (hopefully) correct fix.

[1] https://github.com/gnuradio/gnuradio/blob/master/gr-blocks/include/gnuradio/blocks/peak_detector2_fb.h#L38
[2] Unless we state that the peak output is delayed by one sample.
[3] See Callyan’s test changed in commit 97f0b69 to actually test the described case. The fix by Callyan also introduces another issue: In the ` if (d_found == false)` clause `d_peak_ind` may be set to a value `>= d_look_ahead`. This may lead to a out-of-bound memory write in the next `work()` call. Example: In the updated test case the first `work()` call sets `d_peak_ind = 3` and `d_peak_val`. In the next `work()` call no sample exceeding `d_peak_val` is found. This means that `d_peak_ind = 3` is used to write into `optr[d_peak_ind]` (`d_peak_index` must not be greater than `d_look_ahead`).

willcode (commented):
The PR does not match the title. The returned number of consumed items was already correct and is not changed by this PR.

This PR makes a couple of variable local and adds a test case (both good). Was it supposed to do something else?

rear1019:
I have fixed the title of the PR which is supposed to fix the error described in #3345. The wrong title is the result of the incorrect initial fix (first commit of this PR). See my previous comment for details.

willcode (approved):
Logic looks good to me.

mbr0wn (approved):

mbr0wn:
I shoulda squashed these.... meh.

drmpeg:
The qa_peak_detector2 fails after these commits. ```make test``` on CI really needs to be fixed.

willcode:
Backported as part of https://github.com/gnuradio/gnuradio/pull/4142

willcode:
Backport to 3.8 would be useful. The QA doesn't merge well, so putting this off.

PR #3357, opened 2020-04-13
file_sink: Flush file on stop()

### LINKED ISSUE #2590: GR 3.7/3.8 | top_block.wait() does not flush/close open file
I believe that `wait()` should not return until all files have been flushed/closed. Do you guys agree?

Follows an example:

image

```python
# my_app.py

from foo2 import foo2

f = foo2()
f.start()
f.wait()
# foo2.dat is an empty file at this point
del f    # flush occurs now
# foo2.dat has now the expected content
```

rear1019:
Make sure that data passed to the file sink is written to the disk when
the flowgraph is stopped. Before this change this only happened
implicitly due to a fclose() in the base class destructor. This in turn
means that data was written delayed after stopping the flowgraph [1].

This fixes #2590.

[1] https://github.com/gnuradio/gnuradio/issues/2590

### DIFF
--- a/gr-blocks/lib/file_sink_impl.cc
+++ b/gr-blocks/lib/file_sink_impl.cc
@@ -67,5 +67,12 @@ int file_sink_impl::work(int noutput_items,
     return nwritten;
 }
 
+bool file_sink_impl::stop()
+{
+    do_update();
+    fflush(d_fp);
+    return true;
+}
+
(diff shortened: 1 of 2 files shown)


marcusmueller:
Low risk doing this – and it seems to be the right thing to do. Merging.

PR #3358, opened 2020-04-13
peak_detector2: bug fix and add a test case

### LINKED ISSUE #3345: peak_detector2: The average might not be correct

Callyan:
Fixes #3345 and adds a test case for the debug output.

### DIFF
--- a/gr-blocks/lib/peak_detector2_fb_impl.cc
+++ b/gr-blocks/lib/peak_detector2_fb_impl.cc
@@ -88,9 +88,10 @@ int peak_detector2_fb_impl::work(int noutput_items,
                 sigout[i] = d_avg;
             if (iptr[i] > d_avg * (1.0f + d_threshold_factor_rise)) {
                 d_found = true;
-                d_peak_val = std::numeric_limits<float>::min();
+                d_peak_val = iptr[i];
+                d_peak_ind = i;
                 set_output_multiple(d_look_ahead);
-                return i;
+                return i + 1;
             }
         }
         return noutput_items;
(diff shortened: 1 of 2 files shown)


marcusmueller:
hey, @Callyan, this looks very nice, and correct. Um, now, I really want to merge this, but: do you happen to have a CLA for GNU Radio already? (or maybe your organisation, if any?) If not, please hit up @bhilburn under bhilburn (at) gnuradio (dot) org and say "I need a CLA"

rear1019:
With all due respect, but this fix is flawed just as mine :smile:. See [1].

[1] https://github.com/gnuradio/gnuradio/pull/3356#issuecomment-612977923

Callyan:
So the mail about the CLA has been sent, guess I have to wait now.

@rear1019 I guess you won this "comptetition" between pull requests 😂. Also, there might be the same issue with Peak Detector (when the work function doesn't return the whole batch of items), but changing the way Peak Detector works would make it very close to Peak Detector2 so I don't know what to do here.

mormj:
@Callyan - were you able to get the CLA in place?

Callyan:
After having left this aside for some time, I am now working on getting it. I hope I will have it soon.

Callyan:
I have now my CLA

mormj (approved):

willcode (commented):
I think the fix is not necessary, and is incorrect. The index from the threshold detection pass is saved off, and could possibly be used as the peak index into the input to the _next_ work() call. The referenced issue was that a sample could be counted twice in the average. This could be fixed by putting the d_avg calculation at the bottom of the threshold detection loop so it is not hit before the return.

rear1019:
> I think the fix is not necessary, and is incorrect.

The fix here is incorrect indeed; see [1]

[1] https://github.com/gnuradio/gnuradio/pull/3356#issuecomment-612977923

mbr0wn:
Since we have two people in agreement that this is incorrect, and no further activity, closing now.

PR #3364, opened 2020-04-14
gnuradio-companion: add Meta modifier key for macOS

ra1nb0w:
implements support for the Meta key on macOS.
Others platforms continue to use Ctrl.

if you like, I also have the patch for maint-3.8

### DIFF
--- a/grc/gui/Actions.py
+++ b/grc/gui/Actions.py
@@ -13,6 +13,7 @@ import logging
 
 from gi.repository import Gtk, Gdk, Gio, GLib, GObject
 
+from . import Utils
 
 log = logging.getLogger(__name__)
 
@@ -125,7 +126,10 @@ class Action(Gio.SimpleAction):
         self.label = label
         self.tooltip = tooltip
         self.icon_name = icon_name
-        self.keypresses = keypresses
+        if keypresses:
+            self.keypresses = [kp.replace("<Ctrl>", Utils.get_modifier_key(True)) for kp in keypresses]
+        else:
+            self.keypresses = None
         self.prefix = prefix
         self.preference_name = preference_name
         self.default = default
--- a/grc/gui/Dialogs.py
+++ b/grc/gui/Dialogs.py
@@ -287,9 +287,10 @@ def show_help(parent):
         <u>Remove connection</u>: select the connection and press delete, or 
        drag the connection.
         \n\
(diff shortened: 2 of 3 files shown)


ra1nb0w:
@michaelld what do you think?

japm48:
For readability, I would suggest leaving `<Ctrl>` as it is in the shortcut table and use a `replace` as "post-processing" (i.e. [right here](https://github.com/gnuradio/gnuradio/blob/8e777117e10a70a57e138ddbe2144732a3469cfc/grc/gui/Actions.py#L128)).

I mean, something like: 
```self.keypresses = keypresses.replace("<Ctrl>", Utils.get_modifier_key(True))```

Also append: `markup = markup.replace(...)`
Just after markup assignment.

**Comment by ra1nb0w** [comment]
Can be optimized (I did in that way the first time) setting the `if platform == "darwin":` per file seen that `Actions.py` use `<Ctrl>` and `Dialogs.py` use `Ctrl`

**Comment by ra1nb0w** [comment]
@japm48 good idea unfortunately
```
File "/opt/local/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/site-packages/gnuradio/grc/gui/Actions.py", line 141, in __init__
    self.keypresses = self.keypresses.replace("<Ctrl>", Utils.get_modifier_key(True))
AttributeError: 'NoneType' object has no attribute 'replace'
```

**Comment by japm48** [comment]
Oh, OK; what about this?
```python
if keypresses:
    self.keypresses = [kp.replace("<Ctrl>", Utils.get_modifier_key(True)) for kp in keypresses]
else:
    self.keypresses = None
```

ra1nb0w:
Done. thank you @japm48

michaelld:
Tagging a couple other GR developers to look at this, just in case.

michaelld:
Once this is done, I'd love to see a backport to maint-3.8

ra1nb0w:
> Once this is done, I'd love to see a backport to maint-3.8

sure :)

michaelld:
All CI passes. Changes just for OSX, and the primary maintainer (me) approves. Timeout for other reviews. Merging.

ra1nb0w:
thank you! the backport https://github.com/gnuradio/gnuradio/pull/3390

PR #3365, opened 2020-04-14
blocks: move the `new` from the block `::make()` to core

ThomasHabets:
This PR is a preliminary one to deprecate the direct use of `get_initial_pointer`.

The reasons to change this centralizing the place where `new` is called are:
1. Ease of readability
2. For future magic-making it's easier if less logic is distributed (e.g. if `sptr_magic` need is removed, then it's easy to just change to `std::make_shared`)
3. Fewer steps for all blocks to perform by unifying creating the object and creating the shared pointer.

If you agree then I can extend this to all call sites.


mbr0wn:
This will also require modtool changes.

ThomasHabets:
Added a commit with modtool changes.

Curious. Modtool generates output that doesn't conform to style guide. Is it assumed it'll be `clang-format`ted?

mbr0wn:
> Curious. Modtool generates output that doesn't conform to style guide. Is it assumed it'll be `clang-format`ted?

No -- that's a bug.

ThomasHabets:
I'll fix the generated output in a separate PR.

michaelld:
@ThomasHabets So is the intent here to do the change and show how it will be implemented via the null_source ... and, then once this is in place, you'll do another PR that changes all of the uses of `new` to `::make` as per the null_source? I'm just trying to understand why just the null_source is changed, and nothing else ... thx!

ThomasHabets:
@michaelld that's right. That's what I meant with "If you agree then I can extend this to all call sites".

gnuradio-runtime/include/gnuradio/sptr_magic.h
```diff
@@ -41,6 +45,17 @@ std::shared_ptr<T> get_initial_sptr(T* p)
     return std::dynamic_pointer_cast<T, gr::basic_block>(
         detail::sptr_magic::fetch_initial_sptr(p));
 }
+
+/*
+ * \brief GNURadio version of std::make_shared<> that also provides magic. For
```
mbr0wn, line 50:
```suggestion
 * \brief GNU Radio version of std::make_shared<> that also provides magic. For
```

ThomasHabets:
@mbr0wn So it should be safe to do the rest of the changes now?

michaelld:
> @mbr0wn So it should be safe to do the rest of the changes now?

My vote is: yes!

michaelld:
@marcusmueller should be the decider here if possible, since this will start the process of changing how `make` happens with blocks ... which is, like, pretty serious stuff!

ThomasHabets:
By the power of `sed` I've changed all callsites.

Wasn't that much work after all.

mormj:
would like to get signoff from @marcusmueller on this one

ThomasHabets:
@marcusmueller ping?

ThomasHabets:
@marcusmueller ping again?

ThomasHabets:
@marcusmueller weekly ping :-)

marcusmueller:
hey Thomas, sorry, somehow this was constantly slipping. 

So, let's merge in pybind11 and then this!

ThomasHabets:
@marcusmueller looks like pybind11 was merged? Or are we still waiting for https://github.com/gnuradio/gnuradio/pull/3611 ?

Or were you waiting for me to resolve the conflicts? (I will, shortly)

mormj:
I don't think this should be held up by #3611

ThomasHabets:
Rebased.

ThomasHabets:
@marcusmueller so this should be all set for merging?

ThomasHabets:
ping?

Issue #3366, opened 2020-04-14
PMT: Serialization using endian.h does not work on windows


mrjacobagilbert:
The move to speed up serialization using endian.h in #3320 appears to be broken on windows as endian.h does not exist.

hcab14:
Maybe one of the following can help?
https://github.com/mikepb/endian.h
https://github.com/majn/tgl/blob/master/tl-parser/portable_endian.h
https://github.com/search?q=endian.h

I have no idea of how correct they are though.

mrjacobagilbert:
im not much help here with anything windows unfortunately. i can submit a PR to revert that change, or modify that change so it just fixes the f32/c32 using f64/c64.

citizenrich:
I’m also seeing this error on a build from master on Mac. Not using macports though. Do you see this on macports @michaelld ?

mrjacobagilbert:
this will be moved to boost::endian soon; you should be able to revert without too much pain.

`git revert c1c56d2d149fa1663e0d41b7f50ad60f67258ce9`

sorry about that.

mrjacobagilbert:
Solved in PR #3369 - backport updated also.

michaelld:
Yes this is an issue for OSX too: https://github.com/gnuradio/gnuradio/issues/3401

michaelld:
This is why we need -any- functional CI/QA for OSX and Windows for PRs. It doesn't have to be amazing; it just has to be sufficient.

michaelld:
Closed by #3369 .

PR #3369, opened 2020-04-15
pmt: serialization change from htobeXX() to boost::endian

mrjacobagilbert:
this preserves the behavior in a more portable way since endian.h has a variety of portability issues

Addresses #3366 and #3401 and is related to #3320, the backport has been edited to reflect boost::endian #3342

### DIFF
--- a/gnuradio-runtime/lib/pmt/pmt_serialize.cc
+++ b/gnuradio-runtime/lib/pmt/pmt_serialize.cc
@@ -14,8 +14,8 @@
 
 #include "pmt/pmt_serial_tags.h"
 #include "pmt_int.h"
-#include <endian.h>
 #include <pmt/pmt.h>
+#include <boost/endian/conversion.hpp>
 #include <limits>
 #include <vector>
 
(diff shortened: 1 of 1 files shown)


gnuradio-runtime/include/pmt/pmt.h
```diff
@@ -14,6 +14,7 @@
 #include <pmt/api.h>
 #include <stdint.h>
 #include <boost/any.hpp>
+#include <boost/endian/conversion.hpp>
```
mbr0wn, line 17: Why is this include here?

jacobgilbert-snl: moved to pmt_serialization.cc

michaelld:
@mormj this is a critical bug fix for Windows and OSX. If this LGTY, then go ahead and merge it ASAP ... thx!

jwmelto:
boost/endian/conversion.hpp was introduced in Boost 1.58. gnuradio still allows Boost 1.53 as an acceptable version. This doesn't compile with the older Boost.

PR #3372, opened 2020-04-16
grc: import block namespaces

mrjacobagilbert:
backport for #3317 which fixes a bug in assert handling


mormj:
Added the Don't Merge because that label currently exists on #3317

mrjacobagilbert:
@mormj this backport has been updated to be reflective of the latest feedback from @skoslowski which should be good. i think the `don't merge` label can be removed.

PR #3373, opened 2020-04-17
grc: start process on flowgraph folder

japm48:
Allow the use of predictable relative paths.

This also includes changes on cmake compilation.

### DIFF
--- a/grc/gui/Executor.py
+++ b/grc/gui/Executor.py
@@ -17,6 +17,7 @@ from distutils.spawn import find_executable
 from gi.repository import GLib
 
 from ..core import Messages
+from . import Utils
 
 
 class ExecFlowGraphThread(threading.Thread):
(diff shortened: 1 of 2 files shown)


mormj:
Love the idea of predictable relative paths - tagging a few more devs just to get eyes on this since it potentially changes how existing flowgraphs run

grc/gui/Executor.py
```diff
@@ -83,11 +87,14 @@ def _cpp_popen(self):
 
         xterm_executable = find_executable(self.xterm_executable)
 
-        run_command_args = ['cmake .. &&', 'make && ', xterm_executable, '-e', run_command]
-        Messages.send_start_exec(' '.join(run_command_args))
+        # See: https://docs.python.org/3.8/library/os.html#os.cpu_count
+        nproc = len(os.sched_getaffinity(0))
```
michaelld, line 91: not surprisingly: `os.sched_getaffinity` is not defined on OSX ... Python's build system checks for whether `sched_setaffinity` exists and works on the target OS, and if not then the `os` module is built without support for it!

You can always use something like `min(int(os.cpu_count()/2)-1,1)` ...

japm48: Thanks, also broken on Windows...
Why half cpu_count, though?
Note that this only affects compilation, not flowgraph runtime.

michaelld: For some reason, `cpu_count` returns the number of hyperthreads, not cores, at least on my OSX laptop. We really don't want to use that many compile threads! 1 per actual core max, and preferable 1 smaller than that, provides a reasonable trade-off between compile time and CPU saturation.

japm48: Ok. I could not find a portable way to find the physical cores, so I basically used your solution.

michaelld:
All CI passes. Multiple positive reviews. Merging.

PR #3375, opened 2020-04-17
grc: use correct labels in log window context menu

japm48:
Simple GUI fix.
Previously the context menu showed `gtk-clear` and `gtk-save`.


mormj:
Merging - also would be good to have backport to 3.8 on this one.  I can create a backport PR

PR #3377, opened 2020-04-17
utils: Make modtool generate C++ style compliant code

ThomasHabets:
This doesn't entirely make it perfect, since some formatting depends
on the length of the name of the block, and the length of its
parameters.

In the general case it can't be made perfect without reimplementing
`clang-format`, but even then the generated code is not actually valid
C++ (the size of the data is not defined at `gr_modtool add` time, so
uncompilable placeholders are put in place).

So this may be as close as we'll get generating the code.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -421,6 +421,12 @@ install(
     DESTINATION ${GR_PKG_DOC_DIR}
 )
 
+install(
+    FILES .clang-format
+    RENAME clang-format.conf
+    DESTINATION ${GR_PKG_DATA_DIR}
+)
+
 ########################################################################
 # The following dependency libraries are needed by all gr modules:
 ########################################################################
--- a/gr-utils/modtool/core/add.py
+++ b/gr-utils/modtool/core/add.py
@@ -15,12 +15,27 @@ from __future__ import unicode_literals
 import os
 import re
 import logging
+import subprocess
 
+from gnuradio import gr
 from ..tools import render_template, append_re_line_sequence, CMakeFileEditor
 from ..templates import Templates
 from .base import ModTool, ModToolException
 
 logger = logging.getLogger(__name__)
+
+
+def clang_format(s):
+    try:
+      p = subprocess.Popen(["clang-format"], stdin=subprocess.PIPE, stdout=subprocess.PIPE)
+      out, err = p.communicate(s.encode('utf-8'))
+      if p.returncode != 0:
+        print("Failed to run clang-format: %s", err)
+        return s
+      return out.decode('utf-8')
+    except (RuntimeError, FileNotFoundError) as e:
+      print("Failed to run clang-format: %s", e)
+      return s
 
 
 class ModToolAdd(ModTool):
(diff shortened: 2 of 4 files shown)


japm48:
@ThomasHabets Would it make sense to automatically call `clang-format` on the generated file if it exists? (or by specifying a command flag, etc.).

ThomasHabets:
@japm48 normally I'd say yes, but like I said the generated code is not valid C++. It needs some manual fixups with stream item sizes, etc.

So unless that changes, which would make it easier for a developer to accidentally forget to change these placeholder values, I don't think we can use clang-format.

japm48:
@ThomasHabets 
Oh right, that's true. Does something depend on those placeholders? If not, you could change them to valid C++ identifiers but it still shouldn't compile (e.g. something like `GR_CHANGEME__xy`, that is supposed to be changed).

Also, I'm not sure if this has been discussed before, but would it make sense for `modtool` to fill in these placeholders with user-provided values?

ThomasHabets:
That change is slightly more involved. The shelling out to `clang-format` is easy enough, but then the `.clang-format` would need to be in the example directory. And how would it get there? Currently it's not installed anywhere, so it's less obvious how to get there.

I've made an attempt as a separate commit in this PR now. It works, but may not be right.

ThomasHabets:
Curious. The tests are not failing for me locally.

japm48:
@ThomasHabets
Right, you can always distribute the `.clang-format` file, but I guess this is out of the scope of this PR.

japm48:
> Curious. The tests are not failing for me locally.

Apparently, `clang-format` is not installed in the CI server. You should probably also catch `FileNotFoundError` in your try/except.
See: https://ci.gnuradio.org/buildbot/#/builders/61/builds/1747/steps/8/logs/stdio
```
ERROR: test_add (__main__.TestModToolCore)
Tests for the API function add
  File "/cache/workdir/Debian_10_64_0/build_Debian_10_64/src/gr-utils/modtool/tests/test_modtool.py", line 67, in setUp
    ModToolAdd(**args).run()
  File "/cache/workdir/Debian_10_64_0/build_Debian_10_64/src/gr-utils/modtool/core/add.py", line 149, in run
    self._run_lib()
  File "/cache/workdir/Debian_10_64_0/build_Debian_10_64/src/gr-utils/modtool/core/add.py", line 218, in _run_lib
    self._write_tpl('block_impl_h',   'lib', self.info['blockname'] + '_impl.h')
  File "/cache/workdir/Debian_10_64_0/build_Debian_10_64/src/gr-utils/modtool/core/add.py", line 133, in _write_tpl
    f.write(formatter(render_template(tpl, **self.info)))
  File "/cache/workdir/Debian_10_64_0/build_Debian_10_64/src/gr-utils/modtool/core/add.py", line 30, in clang_format
    p = subprocess.Popen(["clang-format"], stdin=subprocess.PIPE, stdout=subprocess.PIPE)
  File "/usr/lib/python3.7/subprocess.py", line 775, in __init__
    restore_signals, start_new_session)
  File "/usr/lib/python3.7/subprocess.py", line 1522, in _execute_child
    raise child_exception_type(errno_num, err_msg, err_filename)
FileNotFoundError: [Errno 2] No such file or directory: 'clang-format': 'clang-format'
```

ThomasHabets:
Yeah, I realized that must be it just as I was about to fall asleep yesterday. Until then I'd assumed that be caught by RuntimeError. Thanks!

ThomasHabets:
Fixed now, BTW

ThomasHabets:
@mbr0wn please take a look

ThomasHabets:
This was left way too long, and the part that cleans up the output has to be re done from scratch. I'll send a separate PR just for the clang-formatting.

Issue #3302, opened 2020-04-02
GRC: make middle mouse click on tab close flowgraphs


mormj:
Just like what happens when you click on a browser tab with the middle mouse button (wheel)

sidkapoor97:
I believe this will be edited in the grc/gui/actions.py [1] under Old Actions 
also references #3145
[1] https://github.com/gnuradio/gnuradio/blob/master/grc/gui/Actions.py

mormj:
@sidkapoor97 - are you interested in addressing this feature?

sidkapoor97:
Yes I wanted to @mormj I tried going through related issues and PRs

sidkapoor97:
@mormj I tried making changes to gui/Actions.py and gui/Applications.py, I needed help with the 
`keypresses=["Middle Mouse button"]`

`FLOW_GRAPH_CLOSE_MIDDLE_MOUSE = actions.register("app.flowgraph.close",
    label='_Close using the middle mouse',
    tooltip='Close the current flow graph using Middle Mouse',
    icon_name='window-close',
    keypresses=[" "],
)`

These were the following links i referred [https://gitlab.gnome.org/GNOME/gtk/-/blob/3.24.14/gdk/keyname-table.h](url)
[https://developer.gnome.org/pygtk/stable/class-gdkevent.html#constructor-gdkevent](url) - under `gtk.gdk.BUTTON_PRESS` "button"

mormj:
@sidkapoor97 - I actually haven't looked into the implementation here, was just throwing it out as a feature enhancement.  Does your code as shown above work when added to GRC?

sidkapoor97:
@mormj I am not clear on how to test that feature, I tried editing the script and installing gnu radio from source but encountered some errors.  any help on testing would be really appreciated, once I test the feature i will open up a PR

dkozel:
@sidkapoor97 You should first solve all the build errors without making any changes to the script. When GNU Radio works normally, then make your changes. If there are errors with your changes then you will know where the error must be.

To test this feature, middle click on the tab of a flowgraph and it should close. If the flowgraph is unsaved then it should ask to save first, like if you close an unsaved flowgraph already without this feature.

PR #3379, opened 2020-04-17
#3302 make middle mouse close flowgraphs

sidkapoor97:
References #3302 
I have added the changes, the only issue is right now the graph closes just on left clicking the label_button. I could not figure out the keypress for middle mouse. I used [https://gitlab.gnome.org/GNOME/gtk/-/blob/3.24.14/gdk/keyname-table.h](url) as reference, any help would be really appreciated. Thanks for the advice @dkozel  i have tested the change.

### DIFF
--- a/grc/gui/Actions.py
+++ b/grc/gui/Actions.py
@@ -265,7 +265,7 @@ FLOW_GRAPH_CLOSE_MIDDLE_MOUSE = actions.register("app.flowgraph.close_middle_mou
     label='_Close using Middle Mouse Button',
     tooltip='Close the current flow graph using Middle Mouse',
     icon_name='window-close-middle-mouse',
-    keypresses=["Pointer_Button3"],
+    keypresses=["Pointer_Right"],
 )
 APPLICATION_INITIALIZE = actions.register("app.initialize")
 APPLICATION_QUIT = actions.register("app.quit",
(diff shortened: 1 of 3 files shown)


sidkapoor97:
Also for some reason the Block tree panel isn't visible, I couldn't figure out why so

mormj:
Thank you for tacking this feature - as you mentioned it looks like there are still some issues:

Normal GRC, the tabs look like this:
image

But after the PR, the tabs look like:
image

So we have also lost the filename, and it looks like a button to close it.  I have no idea how to implement it, but I was hoping this feature would look more like browser tabs where the middle mouse click closes the tab.  

The block tree panel shows up fine for me.  Perhaps that has just been dragged all the way to the right - I've gotten stuck in that state before, but you should be able to grab it and drag it back.

sidkapoor97:
@mormj so I changed the label to a button hence the outline. I have removed the outline, I forgot that to do before, the filename is because of the `set_markup()` function not running for the `label_button` as I couldn't figure it how. Any help would be appreciated

sidkapoor97:
@mormj I have added the markup and now it's working fine, the only issue is middle mouse click where in any help would be appreciated

mormj:
@sidkapoor97 - I don't have the answers as to how to make the middle mouse click work.  Please let us know if you were able to figure it out.

mormj:
@sidkapoor97 - I'm going to close this PR for now, please open it back up if you are able to get this feature working.  Thanks!

PR #3380, opened 2020-04-18
SWIG: Split dtv_swig.i and trellis_swig*.i files for less compiler memory usage.

drmpeg:


michaelld:
Tagging a few GR devs for another set of eyes ...

PR #3381, opened 2020-04-18
gr-filter: Remove warning in fft_filter.cc logging.

drmpeg:


mormj:
@drmpeg - not too familiar with the logging, but I see some places in this file it uses d_logger which is initialized in the constructor, and then here it configures a new logger.  Can d_logger just be used in these places and then:

```c++
        //gr::logger_ptr logger, debug_logger;
        //gr::configure_default_loggers(logger, debug_logger, "fft_filter_ccc");
        std::ostringstream msg;
        msg << "fft_filter_ccc: ntaps = " << d_ntaps << " fftsize = " << d_fftsize
            << " nsamples = " << d_nsamples;
        GR_LOG_ERROR(d_logger, msg.str());
```

drmpeg:
This is really @marcusmueller code. What I did was just copy the syntax from the same code in fft_filter_ccf::compute_sizes to resolve the warning.

https://github.com/gnuradio/gnuradio/blob/master/gr-filter/lib/fft_filter.cc#L346

gr-filter/lib/fft_filter.cc
```diff
@@ -210,6 +210,7 @@ void fft_filter_ccc::compute_sizes(int ntaps)
 
     if (VERBOSE) {
         gr::logger_ptr logger, debug_logger;
```
michaelld, line 212: I think this will work: you remove this line and the one you've just added, then change the `GR_LOG_ERROR` to use `d_logger` ... can you give that a go & see if it does indeed work? If so, let's use that here & anywhere else in this file .. correcting what Marcus did, since I think `d_logger` is available there too. Thx!

drmpeg: Fixed and tested. I also changed the log from ERROR to ALERT, which seems more appropriate.

michaelld:
I don't see a point in backporting, since these changes aren't really bug fixes; just good to have them in place. If someone else sees differently then please do ahead with the backports.

drmpeg:
@michaelld 3.8 is using std::cerr for these messages, so a backport doesn't apply.

michaelld:
> @michaelld 3.8 is using std::cerr for these messages, so a backport doesn't apply.

Ug! Glad this is fixed here then! logging in this manner is so much nicer & more reliable too in terms of not having debug printouts be overlapping.

PR #3382, opened 2020-04-18
grc: do not assume gsettings exists

japm48:
is_dark_theme() should not throw an exception gsettings is not found (can happen in Windows).


mbr0wn:
FYI, this bugfix does not need to go onto maint-3.8.

Issue #3349, opened 2020-04-11
remove remaining compile warnings in 3.9


japm48:
I thought it makes sense to gather this all in a global "tracking" issue (I know labels exist, but still...).

Here is the list of what I have ([build log](https://gist.github.com/japm48/4a77604329f6b8eb398cec000e8a5b99); I used ninja for an easier to parse output).

--------------

- [ ] cmake: `find_package_handle_standard_args` (PkgConfig)

```
  The package name passed to `find_package_handle_standard_args` (XYZ) does
  not match the name of the calling package (Xyz).  This can lead to
  problems in calling code that expects `find_package` result variables
  (e.g., `_FOUND`) to follow a certain pattern.
  [...]
```

Appears multiple times.
I don't know enough CMake to make a judgment here. I guess most people use `-Wno-dev` anyways. Apparently casing should match.

--------------

- [ ] cmake: `CheckIncludeFileCXX`

```
  Policy CMP0075 is not set: Include file check macros honor
  CMAKE_REQUIRED_LIBRARIES.  Run "cmake --help-policy CMP0075" for policy
  details.  Use the cmake_policy command to set the policy and suppress this
  warning.

  CMAKE_REQUIRED_LIBRARIES is set to:

    -lpthread

  For compatibility with CMake 3.11 and below this check is ignoring it.
Call Stack (most recent call first):
  gnuradio-runtime/lib/CMakeLists.txt:216 (check_include_file_cxx)
```

Ditto.
Documentation: https://cmake.org/cmake/help/v3.17/policy/CMP0075.html

--------------
- [x] tuntap: `strncpy` (PR: #3384)

```
gnuradio/gr-blocks/lib/tuntap_pdu_impl.cc:151:12: warning: ‘char* strncpy(char*, const char*, size_t)’ specified bound 16 equals destination size [-Wstringop-truncation]
  151 |     strncpy(ifr.ifr_name, dev, IFNAMSIZ);
      |     ~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

`strncpy` does not guarantee null termination in this case.
Trivial fix: replace with
```c++
strncpy(ifr.ifr_name, dev, IFNAMSIZ-1);
ifr.ifr_name[IFNAMSIZ-1] = '\0';
```
[...]

PR #3384, opened 2020-04-18
gr-blocks: Fixed strncpy warning in tuntap_pdu_impl.cc

alekhgupta1441:
References #3349 
Trivial fix for strncpy.

### DIFF
--- a/gr-blocks/lib/tuntap_pdu_impl.cc
+++ b/gr-blocks/lib/tuntap_pdu_impl.cc
@@ -49,7 +49,7 @@ tuntap_pdu_impl::tuntap_pdu_impl(std::string dev, int MTU, bool istunflag)
     // make the tuntap
     char dev_cstr[1024];
     memset(dev_cstr, 0x00, 1024);
-    strncpy(dev_cstr, dev.c_str(), std::min(sizeof(dev_cstr), dev.size()));
+    strncpy(dev_cstr, dev.c_str(), std::min(sizeof(dev_cstr) - 1, dev.size()));
 
     bool istun = d_istunflag;
     if (istun) {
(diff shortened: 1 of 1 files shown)


japm48:
I just realized that because of the `memset(...)`  line above, this can be rewritten as simply:
```c++
strncpy(ifr.ifr_name, dev, IFNAMSIZ-1);
```

Also, there is a similar problematic line (not detected by the compiler):

```c++
strncpy(dev_cstr, dev.c_str(), std::min(sizeof(dev_cstr), dev.size()));
```

This should become:

```c++
strncpy(dev_cstr, dev.c_str(), std::min(sizeof(dev_cstr) - 1, dev.size()));
```

But I don't know if this should be a separate PR.

alekhgupta1441:
@japm48 Thanks for seeing this. Changes suggested are done.

michaelld:
My meta-question is: why use a buffer of 1024 length in the first place? `IFNAMSIZ` is 16, and that limit is imposed both inside `tun_alloc` and in `struct ifreq`, so 1024 is silly and unnecessary. My proposal: (1) expand or truncate `d_dev` to `IFNAMSIZ` bytes, and set the last one to 0. Then just copy that whole string any time its needed. Nice and easy, and no need to worry about `strncpy` .. just use `IFNAMSIZ` for any copy and the copied string will always be NULL terminated.

japm48:
@michaelld That's true. I didn't realize that those were the same string!

alekhgupta1441:
@mormj @michaelld Sir, I will surely try to bring a PR soon with above mentioned approach.

mormj:
It does indeed make the warning go away, so merging this fix, and save the larger changes to a different PR.

PR #3386, opened 2020-04-18
gr-digital/python : Updated soft_dec_lut_gen.py(Backport for maint-3.7)

alekhgupta1441:
Backport of PR #3351 for maint-3.7.


mormj (approved):

willcode (approved):
prec is an integer, and linspace works the same with int or float, so this should give the same answer. Why is 2.0 being used in the first place? Is there a use case where prec is non-integer?

argilo (approved):
Like @willcode I'm confused why this isn't just `2**prec`, but since this is a backport it's best to leave it as-is.

PR #3387, opened 2020-04-18
modtool: Added feature to prevent creation of multiple blocks with same name (Backport for maint-3.8)

alekhgupta1441:
Backport of PR #3340 for maint-3.8.

### DIFF
--- a/gr-utils/python/modtool/cli/add.py
+++ b/gr-utils/python/modtool/cli/add.py
@@ -24,6 +24,7 @@ from __future__ import print_function
 from __future__ import absolute_import
 from __future__ import unicode_literals
 
+import os
 import re
 import getpass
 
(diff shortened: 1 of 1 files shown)


swap-nil7:
To be honest, I am shocked how we forgot to add this check! Thanks a lot for bringing it out. Clean and nice work. :)
I just want to point out two things:
- Please use imperative style present tense in the commit message.
- Secondly, we'll need a similar check for the API version in the `core/add.py` file, so you can either add the feature in both the files or just move it to `core/add.py`. The issue with it being only in `core/add.py` (although I would prefer it) is that the user won't be notified as soon as the blockname is entered through the CLI. So, you can update the `core/add.py` file as well.

alekhgupta1441:
@swap-nil7 Thanks for suggesting changes. I have pushed the changes.

swap-nil7:
@alekhgupta1441 : Surely, the CLI will block on the call to `core/add.py` if we are invoking it using the CLI, but if someone invokes it directly using the API, the CLI check will be bypassed.

swap-nil7:
We'll have to fix this in the `master` as well then.

mormj:
Tested and works well - seems to be general improvement and feedback taken into account. Merging.  
I'll also make a forward port PR with these changes for master for consistency

mormj (approved):

Issue #965, opened 2016-09-16
PMT Inconsistencies


mbr0wn:
I'd say these are bed enough they can be considered a bug. Here it goes:

```
P = pmt.cons(pmt.to_pmt('foo'), pmt.to_pmt(1))
if pmt.is_dict(P): # (1)
    length = pmt.length(pmt.dict_items(P)) # Fails
    keys = pmt.dict_keys(P) # Obviously fails
```

It's already annoying that (1) returns true. This makes it impossible to iterate over dicts, because such code doesn't work on pairs.

marcusmueller:
This has been fixed in #3389

PR #3389, opened 2020-04-18
PMT: Provide inherent differentiation between PMT dictionaries and pairs

mrjacobagilbert:
This modification attempts to resolve the persistent [bug identified by @mbr0wn back in 2016](https://github.com/gnuradio/gnuradio/issues/965) in which PMT objects have no built in tests between dicts and pairs which can be frustrating and causes clumbsy workarounds when working with PMT data structures such as PDUs. This fix is not necessarily ideal, but is relatively low risk and should retain all existing functionality and be API compatible.

This adds a new class `pmt_dict` that is essentially identical to `pmt_pair`, however the built in base class `is_dict()` returns true. The public `is_dict()` function has been updated to call the underlying objects `is_dict()` function instead of checking if it is a pair. Thus, dicts are still a special case of pair and tests accordingly, but all pairs are not dicts.

Previously `acons()` would create a new `pmt_pair` of the provided key and value, and then return a new `pmt_pair` of the K-V PMT pair and the provided existing PMT dict. Now it returns a new `pmt_dict` object of the KV pair and the existing dictonary. Thus, objects returned from `acons()` will be `pmt_dict` and not `pmt_pair` objects.

The car and cdr data elements borrowed from the `pmt_pair` class have not been changed for `pmt_dict` to preserve the PMT API, however it may be a good idea to have a subsequent modification that makes these less similar to pairs. Because these inherent pair constructs still exist and the `pmt_dict` can be upcast to the `pmt_pair` the majority of existing alist functionality remains intact for both pairs and dicts.

This could be extended by adding another class `pmt_pdu` which will be similar and allow for more consistent handling of PDU type objects which consist of a dictionary (now easily verifiable!) and either a PMT_NIL or uniform vector object, however a more comprehensive solution would probably be to create an intermediate class derived from `pmt_base` which pair-type objects like `pmt_pair`, `pmt_dict`, and `pmt_pdu` could themselves be derived from.

Specific changes to testing include:
- non-dictionary pairs (and by extension PDUs) return `false` for `is_dict()`
- `dict_keys()` and `dict_values()` will throw `wrong_type` exceptions on all PDUs (previously PDUs that had dictionaries other than PMT_NIL would return a result)
- `dict_items()` will throw `wrong_type` exceptions on all PDUs (previously PDUs with NULL vectors would pass).

There are still a couple things that should probably throw `wrong_type` exceptions, but this is more complicated as far as i can tell:
- `dict_delete()` and `dict_update()` both return values when provided a PDU which has a dictionary that is not NULL, and have a PMT_NIL for the uniform vector (cdr) component... its debatable if this is actually even a PDU since it is not a `(dict . uniform_vector)` pair...
- various other  cases which have to do with the use of PMT_NIL for an empty dictionary not being of special type `pdu_dict`

All these changes i would consider to be fixes for unintended behavior.

### DIFF
--- a/gnuradio-runtime/include/pmt/pmt.h
+++ b/gnuradio-runtime/include/pmt/pmt.h
@@ -642,8 +642,12 @@ c64vector_writable_elements(pmt_t v, size_t& len); //< len is in elements
  * ------------------------------------------------------------------------
  */
 
-//! Return true if \p obj is a dictionary (warning: also returns true for a pair)
+//! Return true if \p obj is a dictionary
 PMT_API bool is_dict(const pmt_t& obj);
+
+//! Return a newly allocated dict whose car is a key-value pair \p x and whose cdr is a
+//! dict \p y.
+PMT_API pmt_t dcons(const pmt_t& x, const pmt_t& y);
 
 //! Make an empty dictionary
 PMT_API pmt_t make_dict();
@@ -713,6 +717,14 @@ PMT_API std::shared_ptr<gr::messages::msg_accepter> msg_accepter_ref(const pmt_t
  *			  General functions
  * ------------------------------------------------------------------------
  */
+
+/*!
+ * Returns true if the object is a PDU meaning:
+ *   the object is a pair
+ *   the car is a dictionary type object (including an empty dict)
+ *   the cdr is a uniform vector of any type
+ */
+PMT_API bool is_pdu(const pmt_t& obj);
 
 //! Return true if x and y are the same object; otherwise return false.
 PMT_API bool eq(const pmt_t& x, const pmt_t& y);
@@ -801,7 +813,7 @@ PMT_API pmt_t reverse_x(pmt_t list);
 /*!
  * \brief (acons x y a) == (cons (cons x y) a)
  */
-inline static pmt_t acons(pmt_t x, pmt_t y, pmt_t a) { return cons(cons(x, y), a); }
+inline static pmt_t acons(pmt_t x, pmt_t y, pmt_t a) { return dcons(cons(x, y), a); }
 
 /*!
  * \brief locates \p nth element of \n list where the car is the 'zeroth' element.
--- a/gnuradio-runtime/lib/pmt/pmt.cc
+++ b/gnuradio-runtime/lib/pmt/pmt.cc
@@ -82,6 +82,8 @@ static pmt_any* _any(pmt_t x) { return dynamic_cast<pmt_any*>(x.get()); }
 //                           Globals
 ////////////////////////////////////////////////////////////////////////////
 
+pmt_null::pmt_null() {}
+
 pmt_t get_PMT_NIL()
 {
     static pmt_t _NIL = pmt_t(new pmt_null());
@@ -314,7 +316,6 @@ std::complex<double> to_complex(pmt_t x)
 //                              Pairs
 ////////////////////////////////////////////////////////////////////////////
 
-pmt_null::pmt_null() {}
 pmt_pair::pmt_pair(const pmt_t& car, const pmt_t& cdr) : d_car(car), d_cdr(cdr) {}
 
 bool is_null(const pmt_t& x) { return x == PMT_NIL; }
@@ -658,9 +659,22 @@ void* uniform_vector_writable_elements(pmt_t vector, size_t& len)
  * Chris Okasaki, 1998, section 3.3.
  */
 
-bool is_dict(const pmt_t& obj) { return is_null(obj) || is_pair(obj); }
+pmt_dict::pmt_dict(const pmt_t& car, const pmt_t& cdr) : pmt_pair::pmt_pair(car, cdr) {}
+
+bool is_dict(const pmt_t& obj) { return is_null(obj) || obj->is_dict(); }
 
 pmt_t make_dict() { return PMT_NIL; }
+
+pmt_t dcons(const pmt_t& x, const pmt_t& y)
+{
+    // require arguments to be a PMT pair and PMT dictionary respectively
+    if (!is_pair(x))
+        throw wrong_type("pmt_dcons: not a pair", x);
+    if (!is_dict(y))
+        throw wrong_type("pmt_dcons: not a dict", y);
+
+    return pmt_t(new pmt_dict(x, y));
+}
 
 pmt_t dict_add(const pmt_t& dict, const pmt_t& key, const pmt_t& value)
 {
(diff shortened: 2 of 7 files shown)


mrjacobagilbert:
Also worth noting: this may not be the most desirable way to do this, and that's fine. But I would like to see this addressed somehow and wanted to have at least one suggestion out there that seemed reasonable.

gnuradio-runtime/lib/pmt/pmt_serialize.cc
```diff
@@ -279,7 +279,10 @@ bool serialize(pmt_t obj, std::streambuf& sb)
     }
 
     if (is_pair(obj)) {
-        ok = serialize_untagged_u8(PST_PAIR, sb);
+        if (is_dict(obj))
```
mbr0wn, line 282:
```suggestion
        if (is_dict(obj)) {
```

gnuradio-runtime/include/pmt/pmt.h
```diff
@@ -642,9 +642,13 @@ c64vector_writable_elements(pmt_t v, size_t& len); //< len is in elements
  * ------------------------------------------------------------------------
  */
 
-//! Return true if \p obj is a dictionary (warning: also returns true for a pair)
+//! Return true if \p obj is a dictionary
 PMT_API bool is_dict(const pmt_t& obj);
 
+//! Return a newly allocated dict whose car is a key-value pair \p x and whose cdr is a
+//! dict \p y.
```
mbr0wn, line 649: This should fail when `y` is not a dict, and `x` is not a key/value pair. Please document the exception or error case if they're not.

gnuradio-runtime/lib/pmt/pmt_int.h
```diff
@@ -128,6 +128,15 @@ class pmt_pair : public pmt_base
     void set_cdr(pmt_t cdr) { d_cdr = cdr; }
 };
 
+class pmt_dict : public pmt_pair
+{
+public:
+    pmt_dict(const pmt_t& car, const pmt_t& cdr);
+    //~pmt_dict(){};
```
mbr0wn, line 135: If we don't need, it remove it please.

jacobgilbert-snl: the rest of the classes in this file have commented destructors, should they all be removed? I wasn't sure why so i put this one in.

gnuradio-runtime/lib/pmt/pmt_serialize.cc
jacobgilbert-snl: Not sure what you are getting at here; most of the code here does not use brackets for one-line if statements but i can certainly add here if its desired.

mrjacobagilbert:
I pushed a fix for a mistake in which dict_delete returned a simple pair instead of a dict type pair. I also went through and verified which functions return dicts and verified they do:

make_dict() << returns PMT_NIL 'dict'
dcons()
dict_add() << uses acons() which uses dcons() internally
dict_update() << uses dict_add()
dict_delete() << returns PMT_NIL, the cdr() of a dict which is also a dict or PMT_NIL, or now a dcons() constructed PMT dict


There are also some PMT functions that _could_ be expected to return dicts under certain situations:

assq(), assv(), and assoc() << can return dicts, but only if they are part of the PMT data structure, and are never modified in these functions.
map() << this makes a list of pairs of the car() of the input which cannot be a dict, so no change
reverse() << this has been updated to use dcons() if the input is a dictionary
reverse_x() << see reverse
list_add() << won't ever return a dict currently. should it?? its an odd function

mrjacobagilbert:
now with extra QA to catch issues with the above tests.

mrjacobagilbert:
I have one other change i'd like to make that is somewhat related, and that is adding a convenience check `is_pdu()` - is it appropriate to include that here?

This would simply verify that the object is a pair, and if so that the car is_dict() and the cdr is_uniform_vector() with corresponding QA.

gnuradio-runtime/include/pmt/pmt.h
mrjacobagilbert: good call and this has been fixed

mrjacobagilbert:
ok i added it; i can remove if anyone objects! I think its helpful and relevant, and cleans up PMT input sanitation for PDU type blocks. QA included.

gnuradio-runtime/lib/pmt/pmt_serialize.cc
japm48: What about this?
```c++
ok = serialize_untagged_u8(is_dict(obj) ? PST_DICT : PST_PAIR, sb);
```

mrjacobagilbert: thats cleaner, thanks; updated.

gnuradio-runtime/lib/pmt/pmt_int.h
mbr0wn: consistency is nice too so...meh don't care.

gnuradio-runtime/lib/pmt/pmt.cc
```diff
@@ -904,6 +923,7 @@ size_t length(const pmt_t& x)
     if (x->is_null())
         return 0;
 
+    // also returns corrrect result for dictionaries
```
mbr0wn, line 926:
```suggestion
    // also returns correct result for dictionaries
```

gnuradio-runtime/lib/pmt/pmt_serialize.cc
```diff
@@ -744,7 +746,11 @@ pmt_t parse_pair(std::streambuf& sb)
     while (1) {
         expr = deserialize(sb); // read the car
 
-        nptr = cons(expr, PMT_NIL); // build new cell
+        if (type == PST_DICT)
```
mbr0wn, line 749:
```suggestion
        if (type == PST_DICT) {
```

yeah, we have places without braces, but we really don't like them and @marcusmueller has been trying to fight them with clang-tidy, so please don't introduce new ones.

mrjacobagilbert: good to know! i'll be sure to keep this in mind (and learn how to use clang-tidy at some point...); thanks

gnuradio-runtime/lib/pmt/pmt.cc
mrjacobagilbert: good catch

jacobgilbert-snl:
@marcusmueller @michaelld - there were quite a few updates from the initial PR but just FYI i do not plan to change anything else in this PR other than address specific reviewer feedback

mormj:
three approvals.  merging.

PR #3395, opened 2020-04-21
Update param.py

ewxl:
If value has a length, print this length in the tooltip when hovering. Useful for taps and other lists when assigned in variables, so you dont have to use the trick to encapsulate it with len while editing in grc.

### DIFF
--- a/grc/gui/canvas/param.py
+++ b/grc/gui/canvas/param.py
@@ -67,7 +67,10 @@ class Param(CoreParam):
         errors = self.get_error_messages()
         tooltip_lines = ['Key: ' + self.key, 'Type: ' + self.dtype]
         if self.is_valid():
-            value = str(self.get_evaluated())
+            value = self.get_evaluated()
+            if hasattr(value,"__len__"):
+                tooltip_lines.append('Length: {}'.format(len(value)))
+            value = str(value)
             if len(value) > 100:
                 value = '{}...{}'.format(value[:50], value[-50:])
             tooltip_lines.append('Value: ' + value)


grc/gui/canvas/param.py
```diff
@@ -67,7 +67,10 @@ def format_tooltip_text(self):
         errors = self.get_error_messages()
         tooltip_lines = ['Key: ' + self.key, 'Type: ' + self.dtype]
         if self.is_valid():
-            value = str(self.get_evaluated())
+            value = self.get_evaluated()
+            if hasattr(value,"__len__"):
+                tooltip_lines.append('Length: {}'.format(len(value)))
+            vaule = str(value)
```
mormj, line 73: This appears to be a typo

grc/gui/canvas/param.py
```diff
@@ -67,7 +67,10 @@ def format_tooltip_text(self):
         errors = self.get_error_messages()
         tooltip_lines = ['Key: ' + self.key, 'Type: ' + self.dtype]
         if self.is_valid():
-            value = str(self.get_evaluated())
+            value = self.get_evaluated()
+            if hasattr(value,"__len__"):
+                tooltip_lines.append('Length: {}'.format(len(value)))
```
mormj, line 72: Should the len be added at the beginning of the tooltip instead?

grc/gui/canvas/param.py
ewxl: Yes, well that was embarrassing. XD This is my first pull request...

Will comply with your suggestions.

grc/gui/canvas/param.py
ewxl: Perhaps, this is personal preference?  I like it after Type:

mormj:
@ewxl - if you get a chance, can you squash your commits and force push back to your branch?  If not whoever merges needs to make sure these are squashed first.

ewxl:
Ok, I think I got it squashed right and forced push.

michaelld:
All CI passes. Multiple approvals. Simple yet useful change. Merging.

PR #3400, opened 2020-04-21
filter: Fix firdes RRC filter gain for alpha == 1

bgottula:
`firdes.root_raised_cosine()` does not normalize the coefficients correctly when `alpha == 1` due to a bug in special-case handling for that specific value of `alpha`. Fixed in this PR along with new unit test.

To reproduce (annotated):
```
In [1]: from gnuradio.filter import firdes

# setting alpha to 0.99 (any value less than 1.0 should give similar results)
In [2]: h = firdes.root_raised_cosine(1, 4, 1, 0.99, 128)                                                                           

# as expected, the sum of all coefficients is roughly 1.0
In [3]: np.sum(h)                                                                                                                   
Out[3]: 1.0000000041691237

# setting alpha to 1.0
In [4]: h = firdes.root_raised_cosine(1, 4, 1, 1.0, 128)                                                                            

# uhhhh...that's not right
In [5]: np.sum(h)                                                                           
Out[5]: 2.0001506736152805
```

### DIFF
--- a/gr-filter/lib/firdes.cc
+++ b/gr-filter/lib/firdes.cc
@@ -571,6 +571,7 @@ vector<float> firdes::root_raised_cosine(
         } else {
             if (alpha == 1) {
                 taps[i] = -1;
+                scale += taps[i];
                 continue;
             }
             x3 = (1 - alpha) * x1;
--- a/gr-filter/python/filter/qa_firdes.py
+++ b/gr-filter/python/filter/qa_firdes.py
@@ -163,6 +163,13 @@ class test_firdes(gr_unittest.TestCase):
         new_taps = filter.firdes.root_raised_cosine(1, 4, 1, 0.35, 11)
         self.assertFloatTuplesAlmostEqual(known_taps, new_taps, 5)
 
+    def test_root_raised_cosine_gain(self):
+        """Confirm DC gain is as expected"""
+        taps = filter.firdes.root_raised_cosine(1, 4, 1, 0.35, 11)
+        self.assertAlmostEqual(sum(taps), 1.0)
+        taps = filter.firdes.root_raised_cosine(1, 4, 1, 1.0, 11)
+        self.assertAlmostEqual(sum(taps), 1.0)
+
     def test_gaussian(self):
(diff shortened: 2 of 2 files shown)


mormj:
@bgottula - Nice catch!  Thank you for the contribution.  This seems like a logical and necessary fix, and thank you for updating the QA tests as well.  This is below 10 line threshold for simple changes without requiring a CLA in place, so for this contribution it is not necessary. But if in the future you are interested in contributing larger code changes, please email "bhilburn AT gnuradio.org" for a form to start the copyright process. Thanks again!

gr-filter/lib/firdes.cc
```diff
@@ -571,6 +571,7 @@ vector<float> firdes::root_raised_cosine(
         } else {
             if (alpha == 1) {
                 taps[i] = -1;
+                scale += taps[i];
```
michaelld, line 574: This line is already below outside the `if` statement ... as well as setting `taps[i]` ... I think the duplication is problematic. I know it was here before this PR ... but, can you please verify the correct algorithm & fix this up?

bgottula: There is a `continue` on the next line, so the duplicate lines you refer to are not executed when the condition of this if statement is true.

I'm not familiar with whatever manipulations were performed to the canonical root raised cosine formula to produce the form used in the `else` part of this. I started attempting to understand the approach in this function by manipulation of the standard formula I'm familiar with but did not get far in the time I have available presently.

michaelld: Ah ... yes, yes there is a `continue` ... somehow missed that ... my bad! Thanks for pointing it out to me. Let me re-review with that new info in mind!

michaelld:
All CI passes. Multiple approvals. No CLA needed. Bug fix. Merging.

michaelld:
@bgottula can you do a backport to maint-3.8 please? and, if need be, maint-3.7 as well? these are small things, so it should be obvious whether the bug exists there or not. And, please do work on the CLA ... these are small commits so we don't need it, but for much more we'll need one.

bgottula:
Can do. I will work on the CLA.

PR #3403, opened 2020-04-24
filter: Fix firdes RRC filter gain for alpha == 1 (backport for maint-3.8)

bgottula:
* Fix gain of firdes.root_raised_cosine() when alpha is set to 1.0. Was
  2x the expected gain for alpha==1.0 previously.
* Adds corresponding unit test.


michaelld:
All CI passes. Multiple approvals. Simple bug fix backport. No CLA needed. Merging.

PR #3406, opened 2020-04-24
gr-fft: added tukey and gaussian windows

mrjacobagilbert:
Adds Tukey window generation.

Sanity check:
```
>>> from gnuradio import fft
>>> from matplotlib import pyplot as plt
>>> plt.plot(fft.window.hann(17), '-o');
[<matplotlib.lines.Line2D object at 0x7f0a64b49748>]
>>> plt.plot(fft.window.tukey(33, .5), '-o'); plt.show()
[<matplotlib.lines.Line2D object at 0x7f0a64b49828>]
>>> plt.show()
```
image

### DIFF
--- a/gr-fft/include/gnuradio/fft/window.h
+++ b/gr-fft/include/gnuradio/fft/window.h
@@ -211,9 +211,20 @@ public:
     static std::vector<float> nuttal_cfd(int ntaps);
 
     /*!
-     * \brief Build a flat top window.
+     * \brief Build a flat top window per the SRS specification
      *
-     * See: http://en.wikipedia.org/wiki/Window_function#Flat_top_window
+     * See:
+     * <pre>
+     *     Stanford Research Systems, "Model SR785 Dynamic Signal
+     *     Analyzer: Operating Manual and Programming Reference,"
+     *     2017, pp 2-13
+     * </pre>
+     *
+     * Note: there are many flat top windows, and this implementation is different from
+     * SciPY and Matlab which use the coefficients from D’Antona et al. "Digital Signal
+     * Processing for Measurement Systems" with the following cosine coefficients: <pre>
+     *     [0.21557895, 0.41663158, 0.277263158, 0.083578947, 0.006947368]
+     * </pre>
      *
      * \param ntaps Number of coefficients in the window.
      */
--- a/gr-fft/include/gnuradio/fft/window.h
+++ b/gr-fft/include/gnuradio/fft/window.h
@@ -288,6 +288,16 @@ public:
     static std::vector<float> riemann(int ntaps);
 
     /*!
+     * \brief Build a Tukey window.
+     *
+     * \param ntaps Number of coefficients in the window.
+     * \param alpha Shaping parameter for the Tukey window, an
+     *        alpha of zero is equivalent to a rectangular
+     *        window, an alpha of 1 is equivalent to Hann.
+     */
+    static std::vector<float> tukey(int ntaps, float alpha);
+
+    /*!
      * \brief Build a window using gr::fft::win_type to index the
      * type of window desired.
      *
--- a/gr-fft/include/gnuradio/fft/window.h
+++ b/gr-fft/include/gnuradio/fft/window.h
@@ -300,6 +300,10 @@ public:
 
     /*!
      * \brief Build a Tukey window.
+     * <pre>
+     * Bloomfield, P. Fourier Analysis of Time Series: An Introduction. New York:
+     * Wiley-Interscience, 2000, pp 69 (eqn 6.9)
+     * </pre>
      *
      * \param ntaps Number of coefficients in the window.
      * \param alpha Shaping parameter for the Tukey window, an
@@ -307,6 +311,17 @@ public:
      *        window, an alpha of 1 is equivalent to Hann.
      */
     static std::vector<float> tukey(int ntaps, float alpha);
+
+    /*!
+     * \brief Build a Gaussian window using the equation
+     * <pre>
+     * exp(-(1/2) * (n/sigma)^2)
+     * </pre>
+     *
+     * \param ntaps Number of coefficients in the window.
+     * \param sigma Standard deviation of gaussian distribution.
+     */
+    static std::vector<float> gaussian(int ntaps, float sigma);
 
     /*!
      * \brief Build a window using gr::fft::win_type to index the
--- a/gr-fft/lib/window.cc
+++ b/gr-fft/lib/window.cc
@@ -339,6 +339,22 @@ std::vector<float> window::tukey(int ntaps, float a)
     return taps;
 }
 
+std::vector<float> window::gaussian(int ntaps, float sigma)
+{
+    if (sigma <= 0)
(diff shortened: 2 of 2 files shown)


michaelld:
A note for those reviewing, at least IMHO: the code being added matches up well with the other window code: no error checking on 'ntaps', no 'const' anywhere, that sort of stuff. Given the rest of the programming in the main code, I think adding in more of a similar style is fine. I'd love to see this code tweaked with ntaps being 'unsigned' but also checking for it > 0; I'd love to see some 'const' and the like ... tweaks that allow the compiler to better optimize the code. None of this code is timing-critical, but it's nice to be consistent & once the rest of GR code is doing it, this code should too. These changes are -not- for this PR!

mrjacobagilbert:
@michaelld so thats a great suggestion and i almost did that before i realized that many filter types are not enumerated in either the fft window.h or firdes.h enums. I also was not sure about reusing the beta parameter. Is a more complete enum list desired? If the 'beta' can be re-used as a generic 'param1' field all except the coswindow()s can be added.

Along the same lines, does the firdes enum need to be updated accordingly as well? its a bit confusing since firdes.window() calls fft.window.build() under the hood but it seems like it still has its own copy of the window type enum.

mrjacobagilbert:
The following are enumerated:

```
hamming(int ntaps);
hann(int ntaps);
blackman(int ntaps);
rectangular(int ntaps);
kaiser(int ntaps, double beta);
blackman_harris(int ntaps, int atten = 92);
bartlett(int ntaps);
flattop(int ntaps);
```

The following are not and probably should be:
```
hanning(int ntaps); (alias to Hann)
nuttall(int ntaps);
blackman_nuttall(int ntaps);
welch(int ntaps);
parzen(int ntaps);
exponential(int ntaps, double d);
riemann(int ntaps);
tukey(int ntaps, float alpha);
```

The following probably should be deprecated:
```
nuttal(int ntaps);
blackman_nuttal(int ntaps);
blackmanharris(int ntaps, int atten = 92);
```

As a part of a larger improvement i agree w/ your comments about compiler optimization, also:

- the Hann and Hamming windows can be computed with the generic coswin() functions
- adding of a gaussian window
- probably some cleanup between firdes.h and window.h

michaelld:
Thanks for looking further into this. After your prodding, I looked further info the code & I do have to agree with you here ... so ... let's do this PR separately, and "as is". I'll redo my review.

michaelld:
I'd love to see windowing dealt with as you note & I've commented. That's for another PR.

mrjacobagilbert:
sounds good. i'll open an issue 'for a rainy day' so this doesnt get lost...

jacobgilbert-snl:
I force pushed a second small commit to this PR in which i fixed the flat top filter coefficients, which were calculated strangely and did not agree with matlab or scipy.

https://github.com/scipy/scipy/blob/master/scipy/signal/windows/windows.py#L606

Accumulated error magnitude on this filter was ~5 orders of magnitude greater than others (which are in the 1e-8 roundoff-error).

I can remove this but if you want but it did not seem worth another separate PR...

michaelld:
> I force pushed a second small commit to this PR in which i fixed the flat top filter coefficients, which were calculated strangely and did not agree with matlab or scipy.
> 
> https://github.com/scipy/scipy/blob/master/scipy/signal/windows/windows.py#L606
> 
> Accumulated 'error' (disagreement between GR and scipy) magnitude on this filter was ~5 orders of magnitude greater than other filters (which are in the 1e-8 roundoff-error).
> 
> I can remove this but if you want but it did not seem worth another separate PR...

Hmmmm ... I have mixed feelings about adding on this commit. It's really not what the PR is initially about, and I'd also like a little more justification beyond "it's what SciPy does". Can you work through the maths to determine how far off GR's or SciPy's version is from ideal? GR's code has been around for a very long time, and was written by someone who I trust would know how to create filters correctly. That said, we all make mistakes here and there ... and maybe that's the case here. I love the idea of fixing bugs in the code, and if this truly does so then that's wonderful! Let's justify the change more precisely, please!

mrjacobagilbert:
Yeah ok let me remove that and ill address it separately in a little more detail and I'll include references in the code. Fwiw, scipy and Matlab agree... Though entirely possible they simply copied one another.

mrjacobagilbert:
Ok, just a couple more details on the flat top coefficients (i know this isnt particularly related to this PR...): 

There are quite a few 'flat top' windows, of which GR appears to implement one from the Stanford Research SR785 spectrum analyzer user manual. Matlab and Scipy use one from [D’Antona, Gabriele, and A. Ferrero. Digital Signal Processing for Measurement Systems. New York: Springer Media, 2006, pp. 70–72]. Octave uses a yet another, as does National Instruments.

The code references wikipedia, which references the same source Matlab and Scipy use which is why i thought this was a bug but its not really, just a documentation thing. I would be happy to either:

1) update the documentation to reference the actual window
2) update the taps to match the source

I would be inclined to do the former so it does not change the code. Performance is nominal; the D’Antona implementation is slightly flatter, the Stanford implementation has slightly better rejection:
image

```python
from matplotlib import pyplot as plt
import numpy as np
from gnuradio import fft

N=256
scale = 4.63867
gr = [1.0/scale, 1.93/scale, 1.29/scale, 0.388/scale, 0.028/scale]
ml = [0.21557895, 0.41663158, 0.277263158, 0.083578947, 0.006947368]

gr_taps = np.array(fft.window.coswindow(N, gr[0], gr[1], gr[2], gr[3], gr[4]))
ml_taps = np.array(fft.window.coswindow(N, ml[0], ml[1], ml[2], ml[3], ml[4]))

#plt.plot(gr_taps - ml_taps)
#plt.show()

gr_f = 20*np.log10(np.fft.fftshift(np.fft.fft(gr_taps)))
ml_f = 20*np.log10(np.fft.fftshift(np.fft.fft(ml_taps)))

plt.plot(gr_f - ml_f, 'kx')
plt.plot(gr_f, 'go-')
plt.plot(ml_f, 'b-')
plt.show()
```

michaelld:
> Ok, just a couple more details on the flat top coefficients (i know this isnt particularly related to this PR...):

I am so not surprised that there are multiple "flattop" implementations. If/when someone gets to updating this code, I'd love to have them in there as options ... leave the current one as is in name and methods, and add the others as options. Thanks for the deep dive & clarifications!

Let's just document why this specific flattop window. Can be here or in another PR; either is good so long as it gets done! Thx!

michaelld:
@mormj are you around to review?

mrjacobagilbert:
Ok @michaelld since this is open im going to throw a gaussian window in here as well as this should be present. I also added a source for the Tukey window in the docs (to hopefully avoid the flat top discussion above at some point in the future...).

Gaussian sanity check:
```python
>>> from matplotlib import pyplot as plt
>>> import numpy as np
>>> from gnuradio import fft
>>> from scipy import signal
>>> gr = np.array(fft.window.gaussian(63, 7.9))
>>> sci = np.array(signal.gaussian(63, 7.9))
>>> print(np.average(abs(gr-sci)))
6.562446273021937e-09
>>> plt.plot(sci, 'bo-'); plt.plot(gr, 'rx'); plt.show()
```

image

PR #3412, opened 2020-04-26
Fixed UI Issue Where Mouse Clicks Did Not Change Frequency

ghostop14:
Clicking on the upper or lower parts of numbers is supposed to
change the frequency when not in read-only mode.  A bug in
mousePressEvent was preventing the updates from being detected.

### DIFF
--- a/gr-qtgui/python/qtgui/digitalnumbercontrol.py
+++ b/gr-qtgui/python/qtgui/digitalnumbercontrol.py
@@ -128,6 +128,8 @@ class DigitalNumberControl(QFrame):
         self.offset = event.pos()
 
         if self.read_only:
+            if self.debug_click:
+                gr.log.info("click received but read-only.  Not changing frequency.")
             return
 
         fm = QFontMetrics(self.numberFont)
@@ -195,10 +197,12 @@ class DigitalNumberControl(QFrame):
                     else:
                         cur_freq -= increment
 
-                    self.setFrequency(cur_freq)
+                    # Cannot call setFrequency to emit.  Change must happen now for
+                    # paint event when clicked.
+                    self.setFrequencyNow(cur_freq)
+                    
                     if self.click_callback is not None:
                         self.click_callback(self.getFrequency())
-
                 break
 
         if (not found_number) and (not clicked_thousands):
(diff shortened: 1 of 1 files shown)


gr-qtgui/python/qtgui/digitalnumbercontrol.py
```diff
@@ -286,6 +304,7 @@ def paintEvent(self, event):
         else:
             textstr = str(self.getFrequency())
         rect = QtCore.QRect(0, 0, size.width()-4, size.height())
+
```
mormj, line 307: looks like extra whitespace was added

gr-qtgui/python/qtgui/digitalnumbercontrol.py
```diff
@@ -195,7 +197,12 @@ def mousePressEvent(self, event):
                     else:
                         cur_freq -= increment
 
-                    self.setFrequency(cur_freq)
+                    # Cannot call setFrequency to emit.  Change must happen now for
+                    # paint event when clicked.
+                    if (cur_freq >= self.min_freq) and (cur_freq <= self.max_freq):
```
mormj, line 202: I see this block used below verbatim - is it also used elsewhere?  perhaps, wrap into a method

ghostop14: Yeah, probably a good idea.  Just pushed update.  I cleaned up the whitespace a little too but left some in for readability.  The control should work like SDR packages where you click on the upper/lower part of the number to change the frequency, so this fixes that functionality.

mormj:
For some reason the numbers are incrementing and decrementing fine for me **before** the patch is applied. (it works after the patch as well).  How can I make the issue manifest?

ghostop14:
@mormj, had a thought this morning.  The numbers had worked when I originally built it, and along the way I had to do some Qt updates for another app.  Then I noticed the numbers suddenly weren't working.  I wonder if it's an artifact of a newer version of Qt somewhere.  There may have been a subtle change (or bug) in Qt in the way and when it can process those signals or something.  I'm on Ubuntu 18.04.3 LTS and I had upgraded PyQtChart (pip3 install --upgrade PyQtChart) along the way to 5.14.0 with all the appropriate lib upgrades.

mormj:
I'm not sure what the combination of things I have going on here - I'm on Ubuntu 19.04, and it worked with and without the patch.  Do you think the patch will make the widget more robust across variations of QT?

ghostop14:
I would say yes.  It consistently works now and the only related change I can think of was the Qt piece.

PR #3417, opened 2020-04-28
gr-dtv: Align DVB scrambler lookup table for 64-bit access.

drmpeg:


drmpeg:
gcc 9.3 has exposed this issue. It causes an alignment fault on 32-bit armv7.

PR #3418, opened 2020-04-28
grc: Deduplicate block paths

mbr0wn:
GRC reads paths from various sources, including the GRC_BLOCKS_PATH
environment variable. It can happen that paths appear twice in the list,
causing warnings about findings blocks twice.

This deduplicates the paths in the config object.

### DIFF
--- a/grc/core/Config.py
+++ b/grc/core/Config.py
@@ -9,6 +9,7 @@ from __future__ import absolute_import
 
 import os
 from os.path import expanduser, normpath, expandvars, exists
+from collections import OrderedDict
 
 from . import Constants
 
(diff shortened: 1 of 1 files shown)


drmpeg:
You need to deduplicate that import statement.

mbr0wn:
> You need to deduplicate that import statement.

This makes we wonder what on earth I was testing and it worked :thinking:

michaelld:
All CI passes. Multiple approvals. Simple change that works nicely! Merging.

michaelld:
@mbr0wn can you do a backport to maint-3.8? It would work on maint-3.7 if we wanted that, but I don't see the need ....

PR #3420, opened 2020-04-28
uhd: Change import order for sip and Qt modules

mbr0wn:
This moves the `import sip` statements in uhd_fft and uhd_siggen_gui
below the import for Qt, without which import errors can occur on
platforms such as Fedora 31.

### DIFF
--- a/gr-uhd/apps/uhd_fft
+++ b/gr-uhd/apps/uhd_fft
@@ -26,11 +26,10 @@ from __future__ import print_function
 from __future__ import division
 import ctypes
 import sys
-import sip
 import threading
 import time
-from distutils.version import StrictVersion
 from PyQt5 import Qt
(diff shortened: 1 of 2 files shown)


michaelld:
Just saw what @atrnati stated, which matches my request ... great minds think alike?!

mbr0wn:
> What would you think about a comment indicating that the `sip` import needs to come after the Qt modules for certain platforms?

Sure.

> since it violates the alphabetical ordering

We don't sort alphabetically, or enforce that. There is a standard import order (not sure if it's PEP8 or from elsewhere), where it's system libraries first, then 3rd-party, then ours. But also, you should group imports from the same source. PyLint will complain about any of these.

But that's unrelated to your change request, which I will apply.

PR #3421, opened 2020-04-28
runtime: attributes: Remove check for gcc 3

mbr0wn:
gcc 4 has been a requirement for many years now, so we can remove the
check for gcc 3.


michaelld:
Relevant CI passes. Multiple approvals. Simple change to remove cruft. Merging.

PR #3429, opened 2020-04-30
grc: fix importing configparser in variable config block

### LINKED ISSUE #2782: Flowgraph generation with python 3 still imports python2 libraries

argilo:
Partially fixes #2782.

This backports #2867 to GNU Radio 3.8 so that the Variable Config block works in both Python 2 and Python 3.

### DIFF
--- a/grc/blocks/variable_config.block.yml
+++ b/grc/blocks/variable_config.block.yml
@@ -34,8 +34,12 @@ parameters:
 value: ${ value }
 
 templates:
-    imports: import ConfigParser
-    var_make: 'self._${id}_config = ConfigParser.ConfigParser()
+    imports: |-
+        try:
+            import configparser
+        except ImportError:
+            import ConfigParser as configparser
+    var_make: 'self._${id}_config = configparser.ConfigParser()
 
         self._${id}_config.read(${config_file})
 
@@ -46,7 +50,7 @@ templates:
         self.${id} = ${id}'
     callbacks:
     - self.set_${id}(${value})
-    - "self._${id}_config = ConfigParser.ConfigParser()\nself._${id}_config.read(${config_file})\n\
+    - "self._${id}_config = configparser.ConfigParser()\nself._${id}_config.read(${config_file})\n\
         if not self._${id}_config.has_section(${section}):\n\tself._${id}_config.add_section(${section})\n\
         self._${id}_config.set(${section}, ${option}, str(${writeback}))\nself._${id}_config.write(open(${config_file},\
         \ 'w'))"


argilo:
I'm not sure if it's a bug in GRC or something I've done wrong here, but after this change I get one copy of the imports per block:

https://github.com/capturethesignal/cts-tools/blob/c4457643c6b10fe8bf474c8e05a7000687c43ade/cts-cli-38/rx_to_fifo.py#L26-L37

argilo:
Ah, that's happening because import deduplication is done line-by-line, and only for lines beginning with `import` or `from`:

https://github.com/gnuradio/gnuradio/blob/5808314010d533690c7c20c81d7062396e0110f5/grc/core/generator/top_block.py#L161-L174

It seems like that code should also consider which multi-line import blocks have been seen before. It's not a big problem though, since Python doesn't mind duplicate imports.

michaelld:
@argilo do you want to try to also fix the deduplication? I think that's fair game for this PR since the changes are causing it. NP if not; just throwing it out as an option. I know Python is OK with multiple same-name imports ... but it's always nice to have clean code!

argilo:
> do you want to try to also fix the deduplication?

Yes, I was of the mind to do that soon. But I think it would be best to put that in a separate pull request since the issue is unrelated to the broken Variable Config block.

Issue #3411, opened 2020-04-26
Error setting frequency with a message


duggabe:
While testing https://github.com/gnuradio/gnuradio/blob/master/gr-qtgui/examples/test_qtgui_msg.grc I found that changing frequency either by entering a value in the Message Edit Box or by double-clicking a frequency on either Sink trace caused the frequency to change to twice the value entered. For example: entering '2500' in the Message Edit Box displays a signal at 5000 Hz. It does the same for double-clicking a value on the x-axis.

I moved the input of the Message Debug block to obtain the message sent. It shows:
<pre>
******* MESSAGE DEBUG PRINT ********
(freq . 2500)
************************************
</pre>
Values with decimal points behaved the same way.
Tested with Raspian Buster with 3.8.1.0; and Ubuntu 20.04 with 3.9.0.0-git

mrjacobagilbert:
This is because that particular flowgraph is sending the set_freq message to both the signal source and the qt sinks which treat this information differently. The signal source now generates a 2.5kHz tone instead of 1kHz _and_ the qt sinks are being told to use 2.5 kHz as the new center frequency. 

One or the other should be used, not both. The blocks are all doing the correct thing its just a mistake in the flowgraph (which should be fixed).

duggabe:
@mrjacobagilbert I confirm that removing those links fixes the problem.

I am also working on https://github.com/gnuradio/gnuradio/issues/1971 and will be generating a PR. Should I leave this issue open and put a note that the PR will close this?

mrjacobagilbert:
I would leave it open for now. I believe the intention was to demonstrate the 'Simple Spectrum Analyzer' functionality where you use the QT GUI to tune a USRP. It's just that the signal source doesnt quite provide the exact functionality. Give this a shot:

![image](https://user-images.githubusercontent.com/7121282/80388562-45116b80-8867-11ea-9ddd-5deb867ffa06.png)

Issue #1971, opened 2018-08-22
Need GRC examples that show usage of every gr-qtgui instrumentation block


marcusmueller:
We have only very specific examples. I think especially beginnner's would greatly benefit from examples that show them how to solve specific problems that involve instrumentation.

marcusmueller:
These flowgraphs should end up in gr-qtgui/examples, but also yield screenshots and thus edits for our block documentation in the wiki:

https://wiki.gnuradio.org/index.php?title=Category:Block_Docs&pagefrom=Q#mw-pages

duggabe:
There are 18 QT GUI items. I have completed adding flowgraphs and reviewing all of them.

duggabe:
1) The example grc for https://wiki.gnuradio.org/index.php/QT_GUI_Bercurve_Sink is in https://github.com/gnuradio/gnuradio/blob/master/gr-fec/examples/ber_curve_gen.grc
Should that be copied to gr-qtgui/examples or do you want to use a `git symbolic-ref`?

2) If I have several flowgraphs with more than one QT GUI in it, is that sufficient for the examples, or does it need to be one-to-one?

marcusmueller:
@duggabe Re:

1. I think the bercurve thing is so FEC-specific that'd be OK to stay in gr-fec/examples; maybe: add a text file "QT GUI Bercurve sink.txt" that just contains "see the gr-fec examples for an illustration of the QT GUI BER curve generation"
2. I love these examples, I don't think isolation makes the examples better: they're still easy enough to understand, but are still "realistic" to illustrate things in relation.

duggabe:
It turns out that the "git symbolic-ref" is a text file as well. I will test both and see what looks better.

duggabe:
There are 10 Instrumentation QT GUIs. As far as QT GUI Widgets, there are 8 in Rel 3.8 and 24 in Master!
Since the wording of the issue asks for "every gr-qtgui instrumentation block", is having those examples sufficient to satisfy this issue?

duggabe:
Each QT GUI block is listed with a flowgraph example.

* https://wiki.gnuradio.org/index.php/QT_GUI_Bercurve_Sink
**  gr-qtgui/examples/Bercurve_Sink.txt => gr-fec/examples/ber_curve_gen.grc
* https://wiki.gnuradio.org/index.php/QT_GUI_Constellation_Sink
**  gr-qtgui/examples/qtgui_message_inputs.grc
[...]

PR #3430, opened 2020-04-30
add and revise flowgraphs for qt_gui examples

duggabe:
resolves https://github.com/gnuradio/gnuradio/issues/1971
resolves https://github.com/gnuradio/gnuradio/issues/3411

### DIFF
--- a/gr-digital/examples/mpsk_stage6.grc
+++ b/gr-digital/examples/mpsk_stage6.grc
@@ -1,0 +1,954 @@
+options:
+  parameters:
+    author: ''
+    category: Custom
+    cmake_opt: ''
+    comment: ''
+    copyright: ''
+    description: ''
+    gen_cmake: 'On'
+    gen_linking: dynamic
+    generate_options: qt_gui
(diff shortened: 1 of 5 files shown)


mrjacobagilbert:
@duggabe thank you for doing this, i think its great to have a complete set of examples. I looked at the FG from #3411 I had commented on, and I would suggest:

1) put the throttle between the source and freq_xlating_fir block and get rid of the null sinks

2) put a note or comment in the FG about what the freq_xlating_fir block is emulating (e.g.: a hardware device such as a USRP that is being tuned by a message). you and I understand that, but I don't necessarily think it is obvious to the users these examples are intended for

mrjacobagilbert:
@duggabe message updates look good. not sure if you saw it or not, but the block comments on by default was addressed in #3432

777arc:
Looks good to me, the reference to an example in a different module is something we have never done before, but seems better than nothing

777arc:
Thanks for updating the example flowgraphs!  Merging

PR #3431, opened 2020-04-30
GRC: Make Actions with preferences explicitly default to something

marcusmueller:
This fixes unexpected behaviour in absence of a prior preference.


mbr0wn:
Merged manually. Conflicts.

PR #3432, opened 2020-04-30
GRC: default to showing of block comments

marcusmueller:
This is a reduced backport of #3431 to 3.8.
It only touches the "show block comments" setting to default to True


mormj:
Applied PR and verified it works.  Merging

PR #3439, opened 2020-05-02
qtgui: remove qt warnings

japm48:
Related bug: #3349.

This replaces obsolete API calls (most of the newer alternatives are present on Qt 5.0).


japm48:
@dkozel No problem. I guess we're all pretty busy right now.

PR #3441, opened 2020-05-02
msg_handler: Switch from boost::function to std::function

jdemel:
This commit is a first stab at moving from `boost::function` to `std::function`.
For now, it does only update gr-blocks. Also, this requires more testing.

If others can confirm that this change works, I'll continue to update all modules.

@marcusmueller recommends to merge this into pybind to consolidate OOT API breaking changes.

This is a work in progress at the moment!

### DIFF
--- a/gnuradio-runtime/lib/block.cc
+++ b/gnuradio-runtime/lib/block.cc
@@ -49,8 +49,7 @@ block::block(const std::string& name,
 {
     global_block_registry.register_primitive(alias(), this);
     message_port_register_in(d_system_port);
-    set_msg_handler(d_system_port,
-                    std::bind(&block::system_handler, this, std::placeholders::_1));
+    set_msg_handler(d_system_port, [this](pmt::pmt_t msg) { this->system_handler(msg); });
 }
 
 block::~block() { global_block_registry.unregister_primitive(symbol_name()); }
--- a/gnuradio-runtime/lib/block.cc
+++ b/gnuradio-runtime/lib/block.cc
@@ -49,7 +49,8 @@ block::block(const std::string& name,
 {
     global_block_registry.register_primitive(alias(), this);
     message_port_register_in(d_system_port);
-    set_msg_handler(d_system_port, boost::bind(&block::system_handler, this, _1));
+    set_msg_handler(d_system_port,
+                    std::bind(&block::system_handler, this, std::placeholders::_1));
 }
 
 block::~block() { global_block_registry.unregister_primitive(symbol_name()); }
--- a/gnuradio-runtime/include/gnuradio/basic_block.h
+++ b/gnuradio-runtime/include/gnuradio/basic_block.h
@@ -18,9 +18,9 @@
 #include <gnuradio/runtime_types.h>
 #include <gnuradio/sptr_magic.h>
 #include <gnuradio/thread/thread.h>
-#include <boost/function.hpp>
 #include <boost/thread/condition_variable.hpp>
 #include <deque>
+#include <functional>
 #include <map>
 #include <string>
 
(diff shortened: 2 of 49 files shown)


mormj:
Applied the patch on pybind branch, and everything seems to run as I'd expect it to.  There are still a few failing qa tests on this branch that need to be cleaned up, but unrelated to this change.

jdemel:
In this case, I'll go through all the other modules and update them as well.

gnuradio-runtime/lib/block.cc
```diff
@@ -49,7 +49,8 @@ block::block(const std::string& name,
 {
     global_block_registry.register_primitive(alias(), this);
     message_port_register_in(d_system_port);
-    set_msg_handler(d_system_port, boost::bind(&block::system_handler, this, _1));
+    set_msg_handler(d_system_port,
+                    std::bind(&block::system_handler, this, std::placeholders::_1));
```
ThomasHabets, line 53: Is there a style preference between `std::bind` and using lambdas?

```
set_msg_handler(d_system_port, [this]{ system_handler(); });
```

mormj: I think the lambdas are much cleaner - I was given feedback to use that form when implementing this block:
https://github.com/gnuradio/gnuradio/blob/bfeef69cc9bdec8e721e07b444392708098cfb90/gr-blocks/lib/selector_impl.cc#L44

jdemel: Looking at the lambda version, it looks way cleaner and better understandable. There's just one catch. There a a lot of places where this needs to be updated.

ThomasHabets: I did some checking on godbolt and benchmarking locally. Lambda seems to be about 3.5x faster than `bind` to call, too.

```c++
#include<functional>
void foo(std::function<void(int)> f)
{
  f(0);
}
```

```c++
#include<iostream>
#include<vector>
#include<functional>

void foo(std::function<void(int)> f);

class Foo {
 public:
  void f(int a){};
  void bar() {
    foo(std::bind(&Foo::f, this, std::placeholders::_1));
  }
  void baz() {
    foo([this](int a){
          this->f(a);
        });
  }
};

int main() {
  Foo f;
  for (int i = 0; i < 1000000000; i++) {
    f.bar();
    //f.baz();
  }
}
```

jdemel: So each call to the bound function is faster with lambdas? That's a very good reason to move to lambdas.

ThomasHabets: Yeah. Ugh, I'm slow to wake up today, so my benchmark code doesn't actually isolate the call itself.

This should be a replacement for the second file:

```c++

class Foo {
 public:
  std::function<void(int)> cb;
  void f(int a){};
};

int main() {
  Foo f;

#if 1
  f.cb = std::bind(&Foo::f, &f, std::placeholders::_1);
#else
  f.cb = [&f](int a){
          f.f(a);
        };
#endif
  for (int i = 0; i < 1000000000; i++) {
    foo(f.cb);
  }
}
```

Still 3.5x cheaper with lambda, yes.

jdemel: Using lambdas has the additional advantage that we can use it with both boost::function and std::function

jdemel:
So the most important change is in `gnuradio-runtime/include/gnuradio/basic_block.h`. Here, we switch from `boost::function` to `std::function`. In case `set_msg_handler` was already called with a lambda, everything continues to work. In case `boost::bind` was used to register message handlers, the code needs to be updated to use lambdas.

The switch from `boost::bind` to lambdas makes up the bulk of this PR.

It is now complete and ready to be reviewed/merged.

mormj:
merging over to the pybind branch

Issue #2875, opened 2019-11-01
Remove `snprintf` usage


marcusmueller:
List of files that detect and/or us snprintf:
{{{
./gnuradio-runtime/ConfigChecks.cmake
./gnuradio-runtime/lib/vmcircbuf_mmap_shm_open.cc
./gnuradio-runtime/lib/vmcircbuf_mmap_tmpfile.cc
./gnuradio-runtime/lib/local_sighandler.cc
./gr-vocoder/lib/freedv_tx_ss_impl.cc
}}}
How about let's fix them all, to totally purge this evil from our midst!

michaelld:
oh yes let's please !

btashton:
@marcusmueller I'll throw together the PR. I think vocoder is the only one still using it, the rest just need the macros removed. 

The vocoder one if I remember is more complicated because there is a struct with a allocated buffer passed around and I was not sure how that was being used which is why I did not include it in the first patch.

argilo:
Re-opening, since #3817 and #3834 haven't been merged yet.

PR #3443, opened 2020-05-02
gr-vocoder: Removed 'snprintf' usage

alekhgupta1441:
Fixes: #2875 Removed 'snprintf' usage from gr-vocoder

### DIFF
--- a/gr-vocoder/lib/freedv_tx_ss_impl.cc
+++ b/gr-vocoder/lib/freedv_tx_ss_impl.cc
@@ -70,7 +70,8 @@ freedv_tx_ss_impl::freedv_tx_ss_impl(int mode,
     if ((d_freedv = freedv_open(mode)) == NULL)
         throw std::runtime_error("freedv_tx_ss_impl: freedv_open failed");
 #endif
-    snprintf(d_cb_state.tx_str, 79, "%s", d_msg_text.c_str());
+    std::string txstr = str(boost::format("%s") % d_msg_text.c_str());
+    strcpy(d_cb_state.tx_str, txstr.c_str()); 
     d_cb_state.ptx_str = d_cb_state.tx_str;
     freedv_set_callback_txt(d_freedv, NULL, get_next_tx_char, (void*)&d_cb_state);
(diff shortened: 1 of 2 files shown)


drmpeg:
Your commit is failing the CheckPR Formatting step because you have a space at the end of the line:
```
    strcpy(d_cb_state.tx_str, txstr.c_str());
```
The utility that's being run by Check PR Formatting is clang-format.

Also, you misspelled gr-vocoder in the commit message.

alekhgupta1441:
@drmpeg Thanks for telling my mistake. I was still wondering on this.
I will soon make changes but I am unable to do this for a while as my system has got some hardware damage recently. I will do this as early as possible.

gr-vocoder/lib/freedv_tx_ss_impl.cc
```diff
@@ -70,7 +70,8 @@ freedv_tx_ss_impl::freedv_tx_ss_impl(int mode,
     if ((d_freedv = freedv_open(mode)) == NULL)
         throw std::runtime_error("freedv_tx_ss_impl: freedv_open failed");
 #endif
-    snprintf(d_cb_state.tx_str, 79, "%s", d_msg_text.c_str());
+    std::string txstr = str(boost::format("%s") % d_msg_text.c_str());
```
mbr0wn, line 73: One of these is not like the other: The left hand has the 79-char limit, the right hand does not, and `txstr` serves no purpose, nor does `boost::format()`.

ThomasHabets: Why not replace with `auto txstr = d_msg_text.substr(0,79)`?

But also, what's so bad about `snprintf()`? Seems like a safe way to cap a string and always have it be null-terminated.

ThomasHabets: In terms of "code smell", `strcpy` is much more pungent than `snprintf`.

mbr0wn:
Agree with @ThomasHabets. This does not seem like an improvement. Also, we don't like to use Boost if we don't have to.

PR #3450, opened 2020-05-04
qtgui_chooser.block.yml fails to build

dl1ksv:
After updating to latest git version of gnuradio and  Ubuntu 20.04
trying to build a chooser block fails with Generate Error: (NameError("'i' is not defined")

This is correct
https://github.com/gnuradio/gnuradio/blob/aa8f15d6408b0a2b70b6a012c53b8cf0bd955b9c/gr-qtgui/grc/qtgui_chooser.block.yml#L132

i is not defined at this point. No idea why it worked before.

Creating the labels list the same way as the options list fixes this issue

### DIFF
--- a/gr-qtgui/grc/qtgui_chooser.block.yml
+++ b/gr-qtgui/grc/qtgui_chooser.block.yml
@@ -126,11 +126,7 @@ templates:
         % if int(num_opts):
         self._${id}_labels = (\
         % for lbl in all_labels:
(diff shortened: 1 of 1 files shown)


mormj:
Tested and see the issue on 19.10 as well.  Not sure why it would be OS dependent, maybe the mako lib is different??  Anyhow, applied the patch and fixes the rendering.

PR #3455, opened 2020-05-05
blocks: improvements to the message_debug block

mrjacobagilbert:
aggregates the data to be printed into a stringstream so that it is printed at the same time reducing debug information becoming muddled between competing cout statements. this change also provides an argument to this block to disable the printing of PDU uniform vectors which are often not needed for debug purposes

I also moved the constructor and destructor to be more consistent with other blocks

### DIFF
--- a/gr-blocks/grc/blocks_message_debug.block.yml
+++ b/gr-blocks/grc/blocks_message_debug.block.yml
@@ -2,6 +2,14 @@ id: blocks_message_debug
 label: Message Debug
 flags: [ python, cpp ]
 
+parameters:
+-   id: en_uvec
+    label: PDU Vectors
+    dtype: enum
+    default: 'True'
+    options: ['True', 'False']
+    option_labels: ['On', 'Off']
+
 inputs:
 -   domain: message
     id: print
@@ -15,7 +23,9 @@ inputs:
 
 templates:
     imports: from gnuradio import blocks
-    make: blocks.message_debug()
+    make: blocks.message_debug(${en_uvec})
+    callbacks:
+    - set_vector_print(${en_uvec})
 
 cpp_templates:
     includes: ['#include <gnuradio/blocks/message_debug.h>']
--- a/gr-blocks/lib/message_debug_impl.cc
+++ b/gr-blocks/lib/message_debug_impl.cc
@@ -20,16 +20,36 @@
 namespace gr {
 namespace blocks {
 
-message_debug::sptr message_debug::make()
+message_debug::sptr message_debug::make(bool en_uvec)
 {
-    return gnuradio::make_block_sptr<message_debug_impl>();
+    return gnuradio::make_block_sptr<message_debug_impl>(en_uvec);
 }
+
+message_debug_impl::message_debug_impl(bool en_uvec)
+    : block("message_debug", io_signature::make(0, 0, 0), io_signature::make(0, 0, 0)),
+      d_en_uvec(en_uvec)
+{
+    message_port_register_in(pmt::mp("print"));
+    set_msg_handler(pmt::mp("print"), [this](pmt::pmt_t msg) { this->print(msg); });
+
+    message_port_register_in(pmt::mp("store"));
+    set_msg_handler(pmt::mp("store"), [this](pmt::pmt_t msg) { this->store(msg); });
+
+    message_port_register_in(pmt::mp("print_pdu"));
+    set_msg_handler(pmt::mp("print_pdu"),
+                    [this](pmt::pmt_t msg) { this->print_pdu(msg); });
+}
+
+message_debug_impl::~message_debug_impl() {}
 
 void message_debug_impl::print(pmt::pmt_t msg)
 {
-    std::cout << "******* MESSAGE DEBUG PRINT ********\n";
-    pmt::print(msg);
-    std::cout << "************************************\n";
+    std::stringstream sout;
+
+    sout << "******* MESSAGE DEBUG PRINT ********" << std::endl
+         << pmt::write_string(msg) << std::endl
+         << "************************************" << std::endl;
+    std::cout << sout.str();
 }
 
 void message_debug_impl::store(pmt::pmt_t msg)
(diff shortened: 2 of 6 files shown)


gr-blocks/grc/blocks_message_debug.block.yml
```diff
@@ -2,6 +2,15 @@ id: blocks_message_debug
 label: Message Debug
 flags: [ python, cpp ]
 
+parameters:
+-   id: en_uvec
+    label: Print PDU Vectors?
```
mormj, line 7: probably don't need the question mark in the label - i don't see it in any other gr blocks bool parameters

mrjacobagilbert: good call. it's been changed.

gr-blocks/lib/message_debug_impl.cc
```diff
@@ -20,16 +20,36 @@
...
-    return gnuradio::get_initial_sptr(new message_debug_impl());
+    return gnuradio::get_initial_sptr(new message_debug_impl(en_uvec));
 }
 
+message_debug_impl::message_debug_impl(bool en_uvec)
+    : block("message_debug", io_signature::make(0, 0, 0), io_signature::make(0, 0, 0)),
+      d_en_uvec(en_uvec)
+{
+    message_port_register_in(pmt::mp("print"));
+    set_msg_handler(pmt::mp("print"), boost::bind(&message_debug_impl::print, this, _1));
```
ThomasHabets, line 33: Can these `boost::bind` be lambdas instead?

See https://github.com/gnuradio/gnuradio/pull/3441

gr-blocks/lib/message_debug_impl.cc
```diff
@@ -20,16 +20,36 @@
...
+message_debug_impl::~message_debug_impl() {}
+
 void message_debug_impl::print(pmt::pmt_t msg)
 {
-    std::cout << "******* MESSAGE DEBUG PRINT ********\n";
-    pmt::print(msg);
-    std::cout << "************************************\n";
+    std::stringstream sout;
+
+    sout << "******* MESSAGE DEBUG PRINT ********" << std::endl;
```
ThomasHabets, line 49: what about using just one semicolon?

```
sout << "…"
   << "…";
```

gr-blocks/lib/message_debug_impl.cc
```diff
@@ -20,16 +20,36 @@
...
+    sout << pmt::write_string(msg) << std::endl;
+    sout << "************************************" << std::endl;
+    std::cout << sout.str();
```
ThomasHabets, line 52: What's the intermediate `stringstream` supposed to accomplish? Writing to stdout is not atomic either way.

gr-blocks/lib/message_debug_impl.cc
```diff
@@ -70,25 +105,8 @@ pmt::pmt_t message_debug_impl::get_message(int i)
     if ((size_t)i >= d_messages.size()) {
         throw std::runtime_error("message_debug: index for message out of bounds.");
     }
-
     return d_messages[i];
 }
 
-message_debug_impl::message_debug_impl()
```
ThomasHabets, line 77: wait, this is just moved around. What are the "improvements" in this commit?

Can you split this commit into actual changes with a separate commit for moving things around?

gr-blocks/lib/message_debug_impl.cc
jacobgilbert-snl: It’s not atomic but it’s much much better in terms of getting associated printouts together  without mixing them. Intermediate stringstreams are used in other in tree blocks like the tag debug.

You have lots of good comments, I’ll get to the rest soon thanks!

gr-blocks/lib/message_debug_impl.cc
mrjacobagilbert: That's pretty neat; i hadn't seen that. I have updated this block to use lambdas.

gr-blocks/lib/message_debug_impl.cc
mrjacobagilbert: Good idea. Done.

gr-blocks/lib/message_debug_impl.cc
mrjacobagilbert: I split the move into a second commit (and i can remove it entirely if it is not desired).

mrjacobagilbert:
@ThomasHabets Thank you for the review, you should find your feedback incorporated. This approach isn't perfect but its a significant improvement. These changes have lived within a gr-pdu_utils block for some time (message debug++) and we have found them very useful.

@mormj I made several changes you may want to look over again...

gr-blocks/lib/message_debug_impl.cc
```diff
@@ -40,25 +60,42 @@ void message_debug_impl::store(pmt::pmt_t msg)
...
+    if (pmt::is_uniform_vector(vector)) {
+        size_t len = pmt::blob_length(vector);
+        if (d_en_uvec) {
+            sout << "pdu length = " << len << " bytes" << std::endl
+                 << "pdu vector contents = " << std::endl;
+            size_t offset(0);
+            const uint8_t* d =
+                (const uint8_t*)pmt::uniform_vector_elements(vector, offset);
+            for (size_t i = 0; i < len; i += 16) {
+                sout << boost::format("%04x: ") % (unsigned int)i;
```
ThomasHabets, line 84: As we want to remove dependence on boost, how about:

```
sout << std::hex << static_cast<unsigned int>(i) << ": ";
```

(and similar below)

gr-blocks/lib/message_debug_impl.cc
ThomasHabets: This doesn't prevent clashing output. You're saying though that in practice it improves it?

gr-blocks/lib/message_debug_impl.h
```diff
@@ -22,6 +22,8 @@ namespace blocks {
 class message_debug_impl : public message_debug
 {
 private:
+    bool d_en_uvec;
```
ThomasHabets, line 25: maybe `const` this?

gr-blocks/lib/message_debug_impl.h
```diff
@@ -63,11 +65,12 @@ class message_debug_impl : public message_debug
     std::vector<pmt::pmt_t> d_messages;
 
 public:
-    message_debug_impl();
+    message_debug_impl(bool en_uvec);
     ~message_debug_impl();
 
     int num_messages();
     pmt::pmt_t get_message(int i);
+    void print_pdu_vector(bool en) { d_en_uvec = en; };
```
ThomasHabets, line 73: Oh, that's a setter function? Is there a precedent in the code base for naming a setter function like this?

ThomasHabets:
In my opinion anything that has plural in "improvements" should be split into separate commits.

In this case it's "printing the vector is optional" and "make fewer syscalls for writing output" being two completely orthogonal improvements, thus should not be in the same commit.

But that's my opinion.

mrjacobagilbert:
I think in general I agree, but not sure its worth making separate commits for all the independent changes: stringstream, boost::bind --> lambda, removal of boost::format, optional PDU vector printing, consistency with constructor/destructor at the top. I'll go back and do that if required.

As to your comment above: yes, in practice we have found using stringbuffers reduces stdout collisions.

Updated to use better setter name and get rid of boost.

mrjacobagilbert:
@mormj rebased

mormj:
Appears concerns have been addressed.  Simple feature addition.  Merging

Issue #3437, opened 2020-05-01
USRP Source Unusable: len_tag_name error with USRP source


cclamb:
Everybody, just did a pybombs update this morning, and now I can't run anything. I get this error on len_tag_name when I try to generate a flowgraph:

>>> Generate Error: /home/user/grc/test_usrp_src_fail.grc: (NameError("'len_tag_name' is not defined"), 'uhd.usrp_source(\n...

I have eliminated the remainder of the trace to keep this short. Can supply if needed. Attached example flowgraph containing a stock USRP source connected to a null sink.

Was able to fix by removing the line:

             ),                                                                                                                                                                                                                                                                           
             % if len_tag_name:                                                                                                                                                                                                                                                           
             ${len_tag_name},                                                                                                                                                                                                                                                             
             % endif  

from uhd_usrp_source.block.yml

and replacing it with ')', but I'm not sure what the consequences of this change are. Thanks! I've attached a simple example showing the issue.

[test_usrp_src_fail.zip](https://github.com/gnuradio/gnuradio/files/4565838/test_usrp_src_fail.zip)

cclamb:
...also able to fix by adding this as a parameter to uhd_usrp_source.bloc.yml:

 -   id: len_tag_name                                                                                                                                                                                                                                                                     
     label: TSB tag name                                                                                                                                                                                                                                                                  
     dtype: string                                                                                                                                                                                                                                                                        
     hide: ${ 'none' if len(str(len_tag_name)) else 'part'}    

Currently at (git log --oneline):

6ad350b06 (HEAD -> maint-3.8, origin/maint-3.8) grc: use correct labels in log window context menu
17b5308a4 grc: Enable strict_undefined to ease Mako template debugging
3b86c6aa3 zeromq: Add bind argument for message based zeromq blocks.
a263ad80b filter: Fix firdes RRC filter gain for alpha == 1
...

cclamb:
So the above will fix the GRC issue where I can't gen a flowgraph, but then the resulting graph won't run; tried to tweak some of the template settings, but had no luck. Just went with the first possible fix.

drmpeg:
I've verified this bug. It's caused by commit https://github.com/gnuradio/gnuradio/commit/17b5308a406f5d2e6694d0913e82f30abd2844ca on maint-3.8 and commit https://github.com/gnuradio/gnuradio/commit/202cecbde26463bb5e2a0da1549a965a16d3b4d3 on master.

[...]

PR #3464, opened 2020-05-08
uhd: Properly ignore len_tag_name in USRP source (3.8)

haakov:
Fixes https://github.com/gnuradio/gnuradio/issues/3437.
I'll make a PR for master too if this one gets approved.


dkozel:
Properly checked this this time and it does work. Filtering for Sink only makes sense. Thanks!

PR #3465, opened 2020-05-08
Extending tagged_file_sink file naming options.

aiicore:
- Introducing target directory configurable option.
- Introducing two naming styles: legacy and non legacy.

Legacy:
 - Works as before.

Non legacy:
 - Introduces configurable filename prefix.
 - Naming convention has chaned to:
  ` <prefix>_%Y-%m-%d_%H_%M_%S.%f.dat`

### DIFF
--- a/gr-blocks/lib/tagged_file_sink_impl.cc
+++ b/gr-blocks/lib/tagged_file_sink_impl.cc
@@ -18,6 +18,8 @@
 #include <fcntl.h>
 #include <sys/stat.h>
 #include <sys/types.h>
+#include <boost/date_time/posix_time/posix_time.hpp>
+#include <boost/filesystem.hpp>
 #include <iostream>
 #include <stdexcept>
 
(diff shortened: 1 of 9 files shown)


aiicore:
Introduces feature from #2907,  inspired by #3010.

gr-blocks/lib/tagged_file_sink_impl.cc
```diff
@@ -142,19 +156,43 @@ int tagged_file_sink_impl::work(int noutput_items,
...
+                        filename.setf(std::ios::fixed, std::ios::floatfield);
+                        filename.precision(8);
+                        filename << "file" << unique_id() << "_" << d_n << "_"
+                                 << d_timeval;
+                    } else {
+                        const boost::posix_time::ptime now =
+                            boost::posix_time::microsec_clock::local_time();
+                        static std::locale time_date(
+                            filename.getloc(),
+                            new boost::posix_time::time_facet("%Y-%m-%d_%H_%M_%S.%f"));
```
aiicore, line 169: This precision will be sufficient, or should I add d_n or d_timeval also?

aiicore:
@dkozel, did few improvements: Filename prefix is now required in non-legacy mode, also is related with state of legacy switch (GUI). Class declaration have more default values, so any old api that would call tagged_file_sink's make, should work as before.

mormj:
@aiicore - were you able to get a CLA in place?

aiicore:
@mormj not yet, the process still is ongoing.

aiicore:
Can't get CLA.

PR #3467, opened 2020-05-10
Remove dependency on Python module `six`

### LINKED ISSUE #3446: Remove dependency on Python module `six`
As code on `master` is Python3 only depndency on `six` should be removed and affected code needs to be adapted to use Python3 constructs.

aiicore:
Resolves #3446 

### DIFF
--- a/gr-digital/python/digital/qa_packet_format.py
+++ b/gr-digital/python/digital/qa_packet_format.py
@@ -9,7 +9,6 @@
 #
 
 import time, struct
-import six
 
 import pmt
 from gnuradio import gr, gr_unittest, digital, blocks
@@ -63,7 +62,7 @@ class test_packet_format_fb(gr_unittest.TestCase):
         rx_access_code = header[0:len(access_code)]
 
         length = len(send_str)
-        rx_length = struct.unpack_from(b"!H", six.b(header), len(access_code))[0]
+        rx_length = struct.unpack_from(b"!H", header.encode("latin-1"), len(access_code))[0]
 
         self.assertEqual(access_code, rx_access_code)
         self.assertEqual(length, rx_length)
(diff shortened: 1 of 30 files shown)


gr-digital/python/digital/qa_packet_format.py
```diff
@@ -63,7 +62,7 @@ def test_packet_format_async_default(self):
         rx_access_code = header[0:len(access_code)]
 
         length = len(send_str)
-        rx_length = struct.unpack_from(b"!H", six.b(header), len(access_code))[0]
+        rx_length = struct.unpack_from(b"!H", header.encode("latin-1"), len(access_code))[0]
```
mbr0wn, line 65: I don't think latin-1 is a good choice. Given how this string is constructed, use `ascii` instead.

grc/core/utils/extract_docs.py
```diff
@@ -277,7 +275,7 @@ def send(code, cmd, args):
 elif __name__ == '__main__':
     def callback(key, docs):
         print(key)
-        for match, doc in six.iteritems(docs):
+        for match, doc in docs.itmes():
```
mbr0wn, line 278: This is a typo. Did you test all of these changes?

gr-digital/python/digital/qa_packet_format.py
aiicore: According to six documentation, should be latin=1, should I change it to ascii instead?
https://six.readthedocs.io/#six.b

grc/core/utils/extract_docs.py
aiicore: Sorry, will check again everything for typos, I've checked most of the changes.

gr-digital/python/digital/qa_packet_format.py
aiicore: Ok, now I get your point about "ascii". Header is constructed from pmt.u8vector, so it should have values of 0-255, then encode("ascii") would be additional "assert" for proper u8vector values.

aiicore: Wel ... 

```
rx_length = struct.unpack_from(b"!H", header.encode("ascii"), len(access_code))[0]
UnicodeEncodeError: 'ascii' codec can't encode characters in position 0-5: ordinal not in range(128)
```

From python docs:

> If the code point is 128 or greater, the Unicode string can’t be represented in this encoding. (Python raises a UnicodeEncodeError exception in this case.)

Changing back to latin-1:

> Latin-1, also known as ISO-8859-1, is a similar encoding. Unicode code points 0–255 are identical to the Latin-1 values, so converting to this encoding simply requires converting code points to byte values; if a code point larger than 255 is encountered, the string can’t be encoded into Latin-1.

mbr0wn: Fair play. I'm still not happy about this encoding/decoding, but that's beside the point of this PR, which is about removing six. We should probably directly use `struct`.

drmpeg:
Superseded by https://github.com/gnuradio/gnuradio/pull/3799, https://github.com/gnuradio/gnuradio/pull/3808, https://github.com/gnuradio/gnuradio/pull/3809, https://github.com/gnuradio/gnuradio/pull/3811 and https://github.com/gnuradio/gnuradio/pull/3835. Closing.

PR #3472, opened 2020-05-11
pybind: add hash check to binding file creation

mormj:
Since bindings aren't automatically generated each time the corresponding public header file changes, this PR flags each binding file with an md5 hash of the header file.  The workflow still needs some thought, and this would need to be extended to all in-tree modules and OOT/modtool

Right now a sequence of operations on in-tree code could look like:
1) change the public header file
2) build and cmake will flag a fatal error
3) regenerate the bindings (bind_gr_module script)
4) copy the hash into the CMakeLists.txt for the module to clear the build error

the hash is just 
```
md5sum /path/to/include/gnuradio/module/blockname.h
```

This is kind of ugly, so I'm looking for suggestions on better ways to tie this into the overall workflow.


mormj:
The first commit put all the hashing information in the CMakeLists.txt file.  Problem with this is that when files are to be regenerated, CMakeLists.txt (the file currently being processed) needs to be modified on the fly.  Switched over instead to putting the info in each pybind cc file:
```
/***********************************************************************************/
/* This file is automatically generated using bindtool and can be manually edited  */
/* The following lines can be configured to regenerate this file during cmake      */
/* If manual edits are made, the following tags should be modified accordingly.    */
/* BINDTOOL_HEADER_FILE(abs_blk.h)                                        */
/* BINDTOOL_HEADER_FILE_HASH(56dfda2d14a2aa7945b7b44cd10aa8a0)                     */
/***********************************************************************************/
```
`BINDTOOL_GEN_AUTOMATIC` - flags whether this file can be built automatically.  Many files in tree were first generated with bindtool, but then manually edited because the code parsing and generation tools can only do so much

`BINDTOOL_USE_PYGCCXML` - flags whether pygccxml is needed as the parser inside bindtool (dependency), because currently without it the simple parser can handle the make function and that's it

`BINDTOOL_HEADER_FILE` - the name of the header file that this binding file binds

`BINDTOOL_HEADER_FILE_HASH` - md5 hash of the header file to detect changes

mormj:
@mbr0wn - I'd like your feedback on this approach before propagating through the rest of the code.  Thanks!

PR #3479, opened 2020-05-13
gr-uhd: Properly ignore len_tag_name in USRP source.

drmpeg:


mbr0wn:
Somehow this was fixed on maint-3.8 but not master :thinking:

drmpeg:
The original issue was reported for 3.8, so @haakov fixed that first. https://github.com/gnuradio/gnuradio/issues/3437

Issue #2852, opened 2019-10-19
"transcendental function" block has bad name; shouldn't exist


marcusmueller:
… it includes sqrt, which is an analytic function, not transcendental at all.

The real problem, though, is that it's a wrapper for a few functions from a specific library (cmath). Since the number of functions wrapped is limited to begin with (13, to be exact), and nearly all of these exist as VOLK kernels somehow, this block should actually rather be split – who's going to look for "transcendental function" when they want a sine, or an exponential function?

The functions wrapped aren't even named in the doxygen, and that is misleading, to top that off, because the cmath reference suggests that all `cmath` functions are wrapped.

marcusmueller:
A viable solution would be to kill the `#defines` and finally getting around writing templated blocks that take `float (*func)(float)`, or more generally `typename T`, `T (*func)(T)`, and maybe `T (*func)(std::complex<T>)`, `std::complex<T> (*func)(T)`, as well as the VOLK-typical `void (*func)(T1*, T2*, unsigned int)`.

aiicore:
I can take this issue, would be nice to get familiar with blocks and refresh rusty C++ knowledge. Transcendental block should be split into individual ~13 blocks under Math Operations category, preferably using VOLK kernels + documentation, right?

marcusmueller:
this was, looking back, a feature request, and as such is still valid.

aiicore:
@marcusmueller if CLA's are no longer a requirement, could someone drop `Needs CLA` tag from #3480 and maybe all of my other PR's? Then I'll reopen them.

willcode:
Right, CLAs are not required. If you reopen the PRs, we'll assume that you accept the DCO. 3480 already has a signoff. Just make sure the others have a git signoff.

aiicore:
@marcusmueller, Transcendental block should be cleared up (defines -> templates) and renamed to Library C function wrapper, or should I continue with splitting it to 13 independent blocks like in the old PR?

Issue #1230, opened 2017-03-10
Move source blocks from gr-analog to gr-blocks


jdemel:
The blocks:
const_source_x
fastnoise_source_x
noise_source_x
random_source_x
random_uniform_source_x
sig_source_x

should be moved to gr-blocks.

Reason: They are essential to most signal processing flowgraphs. And they are not strictly 'analog' blocks.
Furthermore, the random sources should be renamed. Their names should reflect the distribution.

jmcorgan:
If anything, I'd be in favor of breaking gr-blocks up into smaller groups (we already have to split it into 10 separate SWIG builds for memory reasons).

That aside, renaming these to have more descriptive/intuitive names would be useful--what is your suggestion?

marcusmueller:
Personally, I'd like to have **one** `make` factory for different types of noise, not one make per impl of the gengen templates for each type, and then branching in the `work`. The question is how to make that visible in doxygen; but I'm sure we can come up with something.

So, my idea would actually be having a single public class ("noise_source"), with the following factory function

    foo_sptr make( enum {UNIFORM, GAUSSIAN, LAPLACIAN, WHATEVS} dist,
         enum {FLOAT, INT, GR_COMPLEX, BYTE} type,
         float paramA, float paramB,
         size_t precompute_buffer_size=0)

For most of these things, we, by now, have heavily specialized impls – see @stwunsch's boot::random improvements, so I think it'd be a pretty good idea to get rid of the gengen.

Obviously, we'd need message handlers for run-time parameter adjustments (can't know the specific methods if you can't know which kind of source you're handling) – but these would then, at least, be coherent.

jdemel:
Given that gr-blocks is too big. We might introduce gr-source. Move all source blocks, that provide some general functionality into this module.

Naming:
constant_source
fast_complex_gaussian_source (or rayleigh)
complex_gaussian_source
fast_uniform_random_source
uniform_random_source
signal_source (or a better name)

marcusmueller:
I like the "put functionally similar things into the same module" approach. Naming would be gr-siggen to avoid confusion with the general "this block is a source" concept.

[...]

PR #3480, opened 2020-05-13
Splitting Transcendental block into separate ones.

aiicore:
- Removed blocks under Math Operations:
   - blocks_transcendental

- New blocks under Math Operations:
   - blocks_math_functions_cos
   - blocks_math_functions_cosh
   - blocks_math_functions_exp
   - blocks_math_functions_log
   - blocks_math_functions_sin
   - blocks_math_functions_sin
   - blocks_math_functions_sqrt
   - blocks_math_functions_tan
   - blocks_math_functions_tanh
   - blocks_math_functions_acos
   - blocks_math_functions_acosh
   - blocks_math_functions_asin
   - blocks_math_functions_asinh
   - blocks_math_functions_atan
   - blocks_math_functions_atanh
- Blocks are configured for std::complex<float> and float operations.


aiicore:
Resolves #2852

Introducing VOLK functions as a parameter to templates was not possible (correct me if I'm wrong) due to the fact that they are not recognized as types at compilation time.

Added all related functions from cmath/complex, that use only one argument.

Few questions:
1. Should I also introduce double and std::complex<double> here? Those types were defined, but were not used.
2. Should I introduce `std::complex<T> (*func)(T)` and `T (*func)(std::complex<T>)` if possible?

Still tests need to be written.

Please let me know if I'm going in the right direction.

mrjacobagilbert:
Seems like a great idea but gr-blocks doesn't need more...blocks. Maybe a good time to create gr-math per #1230 and [GREP #26](https://github.com/gnuradio/greps/pull/26)?

aiicore:
@mrjacobagilbert Yeah, I wasn't happy with putting everything into gr-blocks, but this is the easiest part to change. I'm more curious if those changes are 'valid' in matter of general approach to the topic. I'm just starting with gnuradio dev.

I will spend some time for writing tests, then will adjust blocs positions according to suggestions in this PR.

Thanks for the hints!

devnulling:
@aiicore Please reach out to bhilburn at gnuradio dot org for a CLA.

aiicore:
@devnulling already in progress

aiicore:
Can't get CLA.

PR #3481, opened 2020-05-13
gr-uhd: Fix Parameter block errors in GRC examples (backport 3.8).

drmpeg:


mormj:
ci errors not related to this commit.  Example files previously broken, so low enough risk, going ahead and merging.

PR #3490, opened 2020-05-18
gr-blocks/peak_detector: Remove UB on infinities

ThomasHabets:
This code fixes two issues:

1. The code used `::min()` in one place, and "minus infinity" in
   another, for the same purpose.
2. It used minus infinity on integer types, which is undefined
   behavior.

(1) seems to have been accidentally left out of
9d9ea63c45b5f314eb344a69340ef49e8edafdfa.

(2)
Section 4.9:
> The behavior is undefined if the truncated value cannot be
> represented in the destination type.

And trunc() on infinity is still infinity, which can't be represented
in integer types.


ThomasHabets:
Note that there are three values in play here:

1. `min()`. Was initial value, now not used. Is -lots for ints, but almost zero for float.
2. -infiniity, Was looped value, now used for initial value (for floats) too.
3. `lowest()`. Lowest non-infinite value. This PR uses it for ints.

Now, arguably `lowest()` can just be used. Or `min()`. It would make for simpler code to not use infinity.

And `peak_detector2_fb` uses `min()`.

Maybe better if they all just use `lowest()`?

PR #3496, opened 2020-05-21
qtgui: Add set_fft_window_normalized() API call to freq_sink

mbr0wn:
This adds an API call set_fft_window_normalized() to the QT GUI
frequency sink. It will modify the window to have normalized power,
regardless of the window shape. This will allow changing window types
without signal peaks varying in amplitude due to window power
differences. Note that this does not affect scalloping loss, or
processing loss from the window. In other words, this normalizes the
energy content to remain constant during window, the amplitude of a peak
can still vary when switching windows.

### DIFF
--- a/gr-qtgui/lib/freq_sink_c_impl.cc
+++ b/gr-qtgui/lib/freq_sink_c_impl.cc
@@ -20,6 +20,9 @@
 #include <qwt_symbol.h>
 #include <volk/volk.h>
 
+#include <boost/range/adaptor/transformed.hpp>
+#include <boost/range/numeric.hpp>
+
 #include <string.h>
 #include <algorithm>
 
@@ -207,6 +210,12 @@ void freq_sink_c_impl::set_fft_window(const filter::firdes::win_type win)
 }
 
 filter::firdes::win_type freq_sink_c_impl::fft_window() { return d_wintype; }
+
+void freq_sink_c_impl::set_fft_window_normalized(const bool enable)
+{
+    d_window_normalize = enable;
+    buildwindow();
+}
 
 void freq_sink_c_impl::set_frequency_range(const double centerfreq,
                                            const double bandwidth)
@@ -410,6 +419,22 @@ void freq_sink_c_impl::buildwindow()
     if (d_wintype != filter::firdes::WIN_NONE) {
         d_window = filter::firdes::window(d_wintype, d_fftsize, 6.76);
     }
+
+    if (!d_window_normalize) {
+        return;
+    }
+    // Calculate power of the window as 1/N * sum of squares
+    const float power =
+        boost::accumulate(
+            d_window | boost::adaptors::transformed([](const float x) { return x * x; }),
+            0.0) /
+        d_window.size();
+    const double norm_factor = std::sqrt(power);
+    GR_LOG_INFO(d_logger, boost::format("norm factor: %1%") % norm_factor);
+    GR_LOG_INFO(d_logger, boost::format("win before: %1%") % d_window.at(10));
+    // Now normalize such that the power of d_window is 1
+    std::for_each(d_window.begin(), d_window.end(), [&](float& x) { x /= norm_factor; });
+    GR_LOG_INFO(d_logger, boost::format("win after: %1%") % d_window.at(10));
 }
 
 bool freq_sink_c_impl::fftresize()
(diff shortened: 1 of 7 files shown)


mbr0wn:
As I submitted, I realize the commit message sucks. I'll update it when CI is done.

mbr0wn:
Old code, Blackman-harris Window:
image
Old code, Boxcar Window:
image


Note the Blackman-harris window is approx. 8 dB below the boxcar window (the input signal is a tone with unit amplitude/power).

New code, Blackman-harris Window:
image

New code, Boxcar:
image

In the new code, the Bh window is only approx. 3 dB below the rectangular window. This is explained by the intentional property of the Bh window to have energy in the sidelobes.

Default behaviour == old behaviour.

gr-qtgui/lib/freq_sink_c_impl.cc
```diff
@@ -410,6 +419,22 @@ void freq_sink_c_impl::buildwindow()
...
     }
+
+    if (!d_window_normalize) {
+        return;
+    }
+    // Calculate power of the window as 1/N * sum of squares
+    const float power =
+        boost::accumulate(
+            d_window | boost::adaptors::transformed([](const float x) { return x * x; }),
+            0.0) /
```
rear1019, line 430: Can be replaced by `std::inner_product()` (or `std::accumulate()` + lambda).

gr-qtgui/lib/freq_sink_c_impl.cc
```diff
@@ -410,6 +419,22 @@ void freq_sink_c_impl::buildwindow()
...
+        d_window.size();
+    const double norm_factor = std::sqrt(power);
+    GR_LOG_INFO(d_logger, boost::format("norm factor: %1%") % norm_factor);
+    GR_LOG_INFO(d_logger, boost::format("win before: %1%") % d_window.at(10));
+    // Now normalize such that the power of d_window is 1
+    std::for_each(d_window.begin(), d_window.end(), [&](float& x) { x /= norm_factor; });
+    GR_LOG_INFO(d_logger, boost::format("win after: %1%") % d_window.at(10));
```
rear1019, line 437: I assume the logging has been added for debugging purposes? The value 10 looks rather arbitrary.

gr-qtgui/lib/freq_sink_c_impl.cc
```diff
@@ -410,6 +419,22 @@ void freq_sink_c_impl::buildwindow()
...
+    if (!d_window_normalize) {
+        return;
+    }
+    // Calculate power of the window as 1/N * sum of squares
+    const float power =
+        boost::accumulate(
+            d_window | boost::adaptors::transformed([](const float x) { return x * x; }),
+            0.0) /
+        d_window.size();
+    const double norm_factor = std::sqrt(power);
```
ThomasHabets, line 432: Is it worth using `1 / sqrt(power)` instead, to save on the divisions?

mbr0wn:
Wow I can't believe I wrote this twice.... in different ways...

gr-qtgui/lib/freq_sink_c_impl.cc
mbr0wn: Thanks, I used that.

gr-qtgui/lib/freq_sink_c_impl.cc
mbr0wn: Yeah this was junk. Thanks for pointing out.

gr-qtgui/lib/freq_sink_c_impl.cc
mbr0wn: Possibly. This is not high-perf code though. I would have used your suggestion had I seen it earlier -- note how I brainfarted this whole operation.

mbr0wn:
OK, here's what happened. This week, I wanted to keep working on this feature. However, I couldn't find a local branch with this feature anywhere. I had forgotten that I'd already submitted the PR. So I started doing this again from scratch.

Note that my newer attempt is actually much nicer, and split into multiple PRs. I will resurrect this when https://github.com/gnuradio/gnuradio/pull/3696 has been merged.

@rear1019 @ThomasHabets thanks for your comments. Despite my brainfart, they were appreciated and the new code fixes the issues. Except for the FP div/mult thing, but I don't think it's a big deal.

PR #3497, opened 2020-05-21
qtgui: Apply unit of frequency sink to mouseover

mbr0wn:
The mouse hover feature of the frequency sink now respects the unit that
was set in the API.

Example: Here, I changed the unit to dBFS (default remains dB).
image


mormj:
should this go on maint-3.8 as well?

mbr0wn:
Sure why not.

PR #3486, opened 2020-05-16
gr-uhd: Add set_start_time for grc in 3.7

segdy:
Currently, `set_start_time` cannot be set in GRC.
When this is needed, the python file has to be changed manually which is overwritten when saving the next time.

This patch adds set_start_time to USRP Source/Sink.
It only shows up when it is set. Normally, it is set to `Default` which results in identical behavior than now. It can be set to an arbitrary float value indicating a time to start in seconds.

PR #3498, opened 2020-05-22
gr-uhd: Add set_start_time for grc in 3.7 (Update)

segdy:
Fixing: https://github.com/gnuradio/gnuradio/pull/3486

Currently, `set_start_time` cannot be set in GRC.
When this is needed, the python file has to be changed manually which is overwritten when saving the next time.

This patch adds `set_start_time` to USRP Source/Sink.
It only shows up when it is set. Normally, it is set to `Default` which results in identical behavior than now. It can be set to an arbitrary float value indicating a time to start in seconds.


mbr0wn (approved):

segdy:
Great! JFYI, a CLA is in place meanwhile

PR #3501, opened 2020-05-22
grc: iterate all selected blocks when changing port type (fixes #3500)

### LINKED ISSUE #3500: gnuradio-companion: Change type of multiple blocks does not work
Affected Version: Gnuradio 3.8.1.0 (of Ubuntu 20.04)

= Steps to reproduce
1. Create flow graph with multiple blocks of same type (e.g. complex)
2. Select at least 2 blocks
3. Use cursor down or up to change type (e.g. to float)

= Actual behaviour
Only one block changes type

= Expected behaviour
All selected blocks change type simultaneously

skoslowski:
previously the first hit terminated further processing

### DIFF
--- a/grc/gui/canvas/flowgraph.py
+++ b/grc/gui/canvas/flowgraph.py
@@ -311,7 +311,7 @@ class FlowGraph(CoreFlowgraph, Drawable):
         Returns:
             true for change
         """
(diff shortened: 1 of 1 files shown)


mormj:
Would be nice to have this on maint-3.8 as well

PR #3529, opened 2020-06-01
cppgen: Update to modern CMake

haakov:
I'd appreciate some testing and feedback on this.

A few comments:

- I was originally planning on using the `link` field in the blocks' `cpp_templates` to figure out what components to include in the `find_package` call, but this isn't as straightforward as one might think, since some of the components depend on others and `find_package` doesn't follow these dependencies. This could work, but one would need to hard code the dependencies somewhere. In this PR, the find_package call enables **all available components** instead (all components that GR has been built with).
- I put the file/append call in the top-level `CMakeLists.txt` because `grc/CMakeLists.txt` doesn't know what components are enabled. This seems a little untidy, so I wouldn't mind any comments on other ways to let GRC know what components are enabled.
- grc.conf ends up looking something like this: 
```
# This file contains system wide configuration data for GNU Radio.
# You may override any setting on a per-user basis by editing
# ~/.gnuradio/config.conf

[grc]
global_blocks_path = /usr/local/share/gnuradio/grc/blocks
local_blocks_path =
default_flow_graph =
xterm_executable = /usr/bin/xterm
canvas_font_size = 8
canvas_default_size = 1280, 1024
enabled_components = testing-support;python-support;post-install;gnuradio-runtime;gr-ctrlport;gnuradio-companion;gr-blocks;gr-fec;gr-fft;gr-filter;gr-analog;gr-digital;gr-dtv;gr-audio;* alsa;* oss;* jack;* portaudio;gr-channels;gr-qtgui;gr-trellis;gr-utils;gr_modtool;gr_blocktool;gr-vocoder;* codec2;* freedv;* gsm
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -518,6 +518,7 @@ install(
   DESTINATION ${CMAKE_MODULES_DIR}/gnuradio
 )
 
+file(APPEND ${CMAKE_CURRENT_BINARY_DIR}/grc/grc.conf "enabled_components = ${_gr_enabled_components}")
 
 ########################################################################
 # Print summary
--- a/grc/core/Config.py
+++ b/grc/core/Config.py
@@ -25,6 +25,7 @@ class Config(object):
         self._gr_prefs = prefs if prefs else DummyPrefs()
         self.version = version
         self.version_parts = version_parts or version[1:].split('-', 1)[0].split('.')[:3]
+        self.enabled_components = self._gr_prefs.get_string('grc', 'enabled_components', '')
         if name:
             self.name = name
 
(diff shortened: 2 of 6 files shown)


mormj (approved):
Did some basic testing - no luck with the QT widgets compiling and running, but for non QT stuff seems to work fine.  Big improvement over the current c++ flowgraphs that don't compile at all right now without these cmake changes.

My recommendation is to go ahead and merge this as is, also backport to 3.8 (assuming c++ templates don't work there either)

marcusmueller:
Agreed :)

Getting it out there hopefully will foster community engagement and hence fixing.

marcusmueller (approved):

Issue #3534, opened 2020-06-02
In-tree examples that fail to compile


mormj:
The following examples cause an exception when being compiled by grcc and need to be investigated

- [ ] gr-trellis/examples/grc/pccc.grc
- [ ] gr-trellis/examples/grc/sccc.grc 
- [ ] gr-trellis/examples/grc/viterbi_equalization.grc 
- [ ] gr-trellis/examples/grc/turbo_equalization.grc 
- [ ] gr-trellis/examples/grc/tcm.grc
- [ ] gr-trellis/examples/grc/interference_cancellation.grc
- [x] gr-fec/examples/tpc_ber_curve_gen.grc 
- [ ] gr-fec/examples/fecapi_async_ldpc_encoders.grc 
- [ ] gr-fec/examples/fecapi_tagged_decoders.grc

mormj:
The trellis examples are also covered under #2529

mcdiarmid:
Perhaps related as I have been going through examples to figure out how to use the trellis blocks - it appears that some of the trellis blocks either have no python binding, or have a naming mis-match between the flowgraph block and python binding library.  For example:

```bash
/usr/bin/python3 -u /home/mcdiarmid/Documents/GNURadio Examples/viterbi_equalization.py

Traceback (most recent call last):
  File "/home/mcdiarmid/Documents/GNURadio Examples/viterbi_equalization.py", line 300, in <module>
    main()
  File "/home/mcdiarmid/Documents/GNURadio Examples/viterbi_equalization.py", line 276, in main
    tb = top_block_cls()
  File "/home/mcdiarmid/Documents/GNURadio Examples/viterbi_equalization.py", line 98, in __init__
    self.trellis_viterbi_combined_xx_0 = trellis.viterbi_combined_fb(trellis.fsm(fsm), block//bpsym, -1, -1, tot_mod[0], tot_mod[1], digital.TRELLIS_EUCLIDEAN)
AttributeError: module 'gnuradio.trellis' has no attribute 'viterbi_combined_fb'
```

```python
>>> from gnuradio import trellis
>>> dir(trellis)
[...]

Issue #3510, opened 2020-05-24
GRC: Lack of CI testing makes it too easy to introduce bugs


argilo:
As far as I know, the aren't yet any CI tests for GRC. As a result, serious bugs can slip through. A couple recent examples:

https://github.com/gnuradio/gnuradio/pull/3317#issuecomment-629635942
https://github.com/gnuradio/gnuradio/pull/3442#issuecomment-632901964

A couple ideas for some simple tests that might have a good chance of catching bugs:

1. Run `grcc` to compile a basic flow graph, then execute it to verify that it produces the expected output.
2. Iterate over all the `.grc` files in `/*/examples/` folders and compile them with `grcc`, then verify that the output is syntactically valid Python.

mormj:
This is an excellent point and a huge weakness in our CI.  There does appear to be some grc related tests that will run when calling ctest, but are not run in the buildbot containers (missing pytest?)

https://github.com/gnuradio/gnuradio/tree/master/grc/tests

But I love the idea of running through all the example flowgraphs in the CI

cclamb:
PLEASE to this - I have had two critical, breaking bugs in GRC in the past two months., it's making it very tough to convince folks to use these tools in critical applications.

argilo:
One more suggestion (which would have caught #3527):

* Run `grcc` to compile a basic **hierarchical** flow graph, then execute it to verify that it produces the expected output.

marcusmueller:
solved in #4208

PR #3535, opened 2020-06-02
grc: add test case to iterate over all examples in the source tree

mormj:
The test looks for all the .grc files in all the examples/ directories in the source tree, and runs grcc.  

Immediately found 15 examples that don't load, see #3534 

I believe the buildbot containers need to be updated with pytest to be able to run this and the other grc tests that were already there.

Addresses #3510

### DIFF
--- a/grc/tests/test_examples.py
+++ b/grc/tests/test_examples.py
@@ -1,0 +1,53 @@
+# Copyright 2020 Free Software Foundation, Inc.
+#
+# This file is part of GNU Radio
+#
+# SPDX-License-Identifier: GPL-3.0-or-later
+#
+
+import pytest
+
+from argparse import Namespace
+from os import path
+import tempfile
+from pathlib import Path
+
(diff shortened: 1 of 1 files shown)


mormj:
I don't think this would run in the CI environment because it relies on things already installed to the prefix.  Needs more thought

noc0lour:
I would be keen to just write a separate `cmake` function and add tests for the grc files in the examples. For starters a successful compile with `grcc` should be all we want.

haakov:
I thought it would be really neat to have each file as a separate test. I believe this Python gist also does not rely on the prefix:
https://gist.github.com/haakov/ad0bedfa68229dc7ea1479861850b72e
(to be placed in `grc/tests`)

mormj:
@haakov - i really like that idea - if you have this working are you able to open it as a PR?

mormj:
Closing in favor of #4208

Issue #3521, opened 2020-05-28
Runtime: Setting block_alias has a severe impact on latency for message blocks


jacobgilbert-snl:
Setting the block alias on a message block causes a severe impact on latency. This affects at least maint-3.7, maint-3.8, and master.

This was identified by @jdemel and is easily reproduced if you have a mechanism to validate the time a message is published. The following flowgraph using blocks from gr-timing_utils illustrates this. One set_pdu block has an alias set, the other does not.

![image](https://user-images.githubusercontent.com/36242998/83203154-0bea4680-a106-11ea-9db8-714151e5e9b1.png)

Output after letting it run for a little time:
```
message_counter1 :INFO: Message counter noalias got 4060 messages
message_counter0 :INFO: Message counter alias got 4051 messages
gr::log :INFO: time_delta0 - WALL_CLOCK_TIME_DEBUG (time_delta_noalias): Mean = 0.111014 ms, Var = 0.019549 ms
gr::log :INFO: time_delta1 - WALL_CLOCK_TIME_DEBUG (time_delta_alias): Mean = 125.483784 ms, Var = 5214.583800 ms
```

The distribution of 'latency' values (several are shown below, in ms) is specific and appears to be periodic.

```
time_delta_alias . 241.579
time_delta_alias . 231.052
time_delta_alias . 220.788
time_delta_alias . 210.429
time_delta_alias . 199.37
time_delta_alias . 188.272
time_delta_alias . 177.174
time_delta_alias . 166.546
time_delta_alias . 155.161
time_delta_alias . 144.187
time_delta_alias . 133.57
time_delta_alias . 122.527
time_delta_alias . 111.448
time_delta_alias . 100.6
time_delta_alias . 89.5689
time_delta_alias . 78.373
time_delta_alias . 68.049
time_delta_alias . 56.89
time_delta_alias . 46.1071
time_delta_alias . 34.9669
time_delta_alias . 23.958
time_delta_alias . 12.7881
time_delta_alias . 1.35708
time_delta_alias . 240.599
time_delta_alias . 230.209
time_delta_alias . 219.339
time_delta_alias . 208.341
time_delta_alias . 196.912
time_delta_alias . 185.87
time_delta_alias . 174.934
time_delta_alias . 163.942
time_delta_alias . 152.676
time_delta_alias . 141.518
time_delta_alias . 130.302
time_delta_alias . 119.346
time_delta_alias . 108.357
time_delta_alias . 97.4538
[...]

PR #3536, opened 2020-06-02
runtime: fix latency issue caused by setting block alias on message blocks

mrjacobagilbert:
This addresses #3521 by ~~creating a new class object that contains an identifier that does not fail block lookup when messages are published to a block.~~ conducting early notification of message blocks by their `symbol_name` instead of their `alias`.

~~This is a crutch though,  a more desirable solution would be to unify the various ways blocks are addressed, mapped, and/or registered with runtime fixtures.~~

### DIFF
--- a/gnuradio-runtime/lib/basic_block.cc
+++ b/gnuradio-runtime/lib/basic_block.cc
@@ -188,7 +188,7 @@ void basic_block::insert_tail(pmt::pmt_t which_port, pmt::pmt_t msg)
     msg_queue_ready[which_port]->notify_one();
 
     // wake up thread if BLKD_IN or BLKD_OUT
-    global_block_registry.notify_blk(alias());
+    global_block_registry.notify_blk(d_symbol_name);
 }
 
 pmt::pmt_t basic_block::delete_head_nowait(pmt::pmt_t which_port)
--- a/gnuradio-runtime/lib/block.cc
+++ b/gnuradio-runtime/lib/block.cc
@@ -47,7 +47,7 @@ block::block(const std::string& name,
       d_pmt_done(pmt::intern("done")),
       d_system_port(pmt::intern("system"))
 {
-    global_block_registry.register_primitive(alias(), this);
+    global_block_registry.register_primitive(d_symbol_name, this);
     message_port_register_in(d_system_port);
     set_msg_handler(d_system_port, [this](pmt::pmt_t msg) { this->system_handler(msg); });
 }
@@ -598,7 +598,7 @@ void block::system_handler(pmt::pmt_t msg)
     pmt::pmt_t op = pmt::car(msg);
     if (pmt::eqv(op, d_pmt_done)) {
         d_finished = pmt::to_long(pmt::cdr(msg));
-        global_block_registry.notify_blk(alias());
+        global_block_registry.notify_blk(d_symbol_name);
     } else {
         std::cout << "WARNING: bad message op on system port!\n";
         pmt::print(msg);
--- a/gnuradio-runtime/lib/block_registry.cc
+++ b/gnuradio-runtime/lib/block_registry.cc
@@ -120,10 +120,14 @@ void block_registry::notify_blk(std::string blk)
     gr::thread::scoped_lock guard(d_mutex);
 
     if (primitive_map.find(blk) == primitive_map.end()) {
-        return;
+        throw std::runtime_error("block notify failed: block not found!");
     }
-    if (primitive_map[blk]->detail().get())
+    if (primitive_map[blk]->detail().get()) {
         primitive_map[blk]->detail()->d_tpb.notify_msg();
+    } else {
+        // not having block detail is not necessarily a problem; this will happen when
+        // publishing a message to a block that exists but has not yet been started
+    }
 }
 
 } /* namespace gr */


gnuradio-runtime/lib/basic_block.cc
```diff
@@ -35,6 +35,7 @@ basic_block::basic_block(const std::string& name,
       d_unique_id(s_next_id++),
       d_symbolic_id(global_block_registry.block_register(this)),
       d_symbol_name(global_block_registry.register_symbolic_name(this)),
+      d_registry_name(name + std::to_string(d_symbolic_id)),
```
jdemel, line 38: Would it be a solution to use `d_symbolic_id` directly?
There's also `d_unique_id` which might be another alternative.

jacobgilbert-snl: `d_symbolic_id` is not unique, and `d_unique_id` is not the correct type but it looks like we can use `d_symbol_name`. This significantly simplifies this PR so I'll get it updated as soon as i am able.

jdemel: `d_symbol_name` is the default value of `alias()`. This seems like the right way to go.

gnuradio-runtime/lib/block.cc
```diff
@@ -47,12 +47,12 @@ block::block(const std::string& name,
       d_pmt_done(pmt::intern("done")),
       d_system_port(pmt::intern("system"))
 {
-    global_block_registry.register_primitive(alias(), this);
+    global_block_registry.register_primitive(registry_name(), this);
     message_port_register_in(d_system_port);
     set_msg_handler(d_system_port, boost::bind(&block::system_handler, this, _1));
 }
 
-block::~block() { global_block_registry.unregister_primitive(symbol_name()); }
```
jacobgilbert-snl, line 55: @jdemel something i did not understand about the existing codebase is why `alias()` is registered but `symbol_name()` is unregistered. Do you understand this? I think these changes in the PR are no longer necessary but wanted to understand this a little better.

gnuradio-runtime/lib/basic_block.cc
```diff
@@ -188,7 +188,7 @@ void basic_block::insert_tail(pmt::pmt_t which_port, pmt::pmt_t msg)
     msg_queue_ready[which_port]->notify_one();
 
     // wake up thread if BLKD_IN or BLKD_OUT
-    global_block_registry.notify_blk(alias());
+    global_block_registry.notify_blk(d_symbol_name);
```
jdemel, line 191: I like this change. I hope it'll improve the situation.
For some reason the block interaction with `global_block_registry` is based on `alias()`. see:
https://github.com/gnuradio/gnuradio/blob/5a2983534fa1faac49528aa41c6cacad806a70d8/gnuradio-runtime/lib/block.cc#L601
https://github.com/gnuradio/gnuradio/blob/5a2983534fa1faac49528aa41c6cacad806a70d8/gnuradio-runtime/lib/block_registry.cc#L50
I went through `gnuradio-runtime`. Though, `alias()` is used in many places. Mostly with logging and RPC. As fas as I can tell, these are the only occurences that might interact with your PR.

I assume @bastibl  and @marcusmueller are more familiar with the details.

jacobgilbert-snl: OK, i agree these need to change. If blocks are registered according to their `symbol_name` instead of `alias`, the call you mention in block_registry.cc becomes unnecessary entirely as it is already removed from the map here:

https://github.com/gnuradio/gnuradio/blob/5a2983534fa1faac49528aa41c6cacad806a70d8/gnuradio-runtime/lib/block_registry.cc#L48

Though this does give me some hesitation that I am missing something...

jacobgilbert-snl: On second thought, i believe the block_registry alias() call is still required and this area is unrelated to the modification of how blocks are notified. the ref_map gets additions here

https://github.com/gnuradio/gnuradio/blob/master/gnuradio-runtime/lib/block_registry.cc#L70

and here

https://github.com/gnuradio/gnuradio/blob/master/gnuradio-runtime/lib/block_registry.cc#L89

both of which should be cleaned up in `block_registry::block_unregister()` I believe.

PR has been updated accordingly.

jacobgilbert-snl:
I agree about letting it simmer... and will hold off on backports for now (this affects 3.7 and 3.8 also).

This is one of those changes for which complete regression testing is very difficult, especially since I am not particularly familiar with the block registry. QA passes on gnuradio and a dozen or so OOTs I have verified, I've used this within a couple dozen applications as a quick-test, but it would be great if others familiar with this part of runtime could review.

bastibl:
I think that's the correct way to fix the problem.

My understanding of what happens is this: When a block posts an async message to a downstream block, it puts the message in the corresponding input queue and wakes up the block through the block registry. This mechanism does not work when an alias is set. It means the block is not notified and only woken up through the [periodic poll every 250ms](https://github.com/gnuradio/gnuradio/blob/master/gnuradio-runtime/lib/tpb_thread_body.cc#L146-L149). At this point, the block recognizes all messages that piled up and processes them in one iteration of the main loop.

I think the wakeup does not work, since the `block` is registered with the block registry in the constructor (`register_primitive`). At this point, there cannot be any alias set. And unlike the blocks symbolic name, this identifier is also not updated if an alias is set later. To make it clear that we only use the `symbol_name`, I think we should update [the block](https://github.com/gnuradio/gnuradio/blob/master/gnuradio-runtime/lib/block.cc#L50) accordingly.

Also, is there any case where we might want to wakeup a block that doesn't exists? If not, I'd suggest to add a warning or exception to the [block registry](https://github.com/gnuradio/gnuradio/blob/master/gnuradio-runtime/lib/block_registry.cc#L122-L124).

mrjacobagilbert:
Thoughts on this?

I have been testing with this and have not had any issues so far. I agree w/ @bastibl about a warning though if an exception is deemed more appropriate let me know.

mrjacobagilbert:
@bastibil as you can see from the failed make test, there apparently is a reason to allow posting to blocks that do not exist. 

https://github.com/gnuradio/gnuradio/blob/master/gr-analog/python/analog/qa_sig_source.py#L222

Should this even be a warning to stdout? I'll leave the failed commit in there for now so this doesnt get merged yet.

mrjacobagilbert:
Reposting this due to mistake in previous comment. 

@bastibl as you can see from the failed make test, there apparently is a reason to allow posting to blocks that do not exist. 

https://github.com/gnuradio/gnuradio/blob/master/gr-analog/python/analog/qa_sig_source.py#L222

Should this even be a warning to stdout? I'll leave the failed commit in there for now so this doesnt get merged yet.

bastibl:
My understanding is that the test fails, because the current patch raises an exception if the block is not part of a running flowgraph, i.e., its `detail` is not yet set. Posting to a block before the flowgraph is started (as it is done in the test of the signal source) makes sense and is nice to have. I also used it for benchmarking, for example.
I didn't test it, but my suggestion would be to leave the exception if the block is not found but remove the exception when the block has no detail.

jacobgilbert-snl:
@bastibl Thanks; this makes sense to me and I've updated the PR accordingly. The failing tests now pass as expected. Because this is normal and acceptable behavior there is no user warning if there is no block detail.

This is complete from my perspective.

mrjacobagilbert:
@bastibl Thanks; this makes sense to me and I've updated the PR accordingly. The failing tests now pass as expected. Because this is normal and acceptable behavior there is no user warning if there is no block detail.

This is complete from my perspective.

bastibl (approved):
Looks good to me. Thanks.

mormj:
Appropriate approvals and passing tests.  Merging!

jacobgilbert-snl:
@mormj @bastibl thoughts on back porting this fix to 3.8?

PR #3537, opened 2020-06-02
examples: fix a few that would not compile with grcc

mormj:
In fec and network there were a few example files that had various issues


ghostop14:
LGTM.  Network changes tested successfully.

mormj:
Thanks @ghostop14 - since it is just example files impacted, going ahead and merging without additional review.

PR #3538, opened 2020-06-03
gr-dtv: Decrease send_frame_size in all examples with a UHD sink block.

drmpeg:
UHD no longer supports large send_frame_size values after v3.13.0.3-rc1.
Large values will cause "Got a ctrl packet with unknown SID" errors.


mormj:
Yep, this one has bit me a few times.  Merging.

PR #3545, opened 2020-06-05
qa: Run qtgui tests with xvfb-run when needed

velichkov:
If running on Linux, the `DISPLAY` environment variable is not set and xvfb-run is present then run the Qt tests with xvfb-run. This would allows running the Qt test in the gnuradio's CI.

### DIFF
--- a/gr-qtgui/python/qtgui/CMakeLists.txt
+++ b/gr-qtgui/python/qtgui/CMakeLists.txt
@@ -48,10 +48,17 @@ if(ENABLE_TESTING)
     ${CMAKE_BINARY_DIR}/gr-qtgui/swig
     )
 
+  if(LINUX AND NOT DEFINED ENV{DISPLAY})
+    find_program(XVFB_EXECUTABLE xvfb-run)
+    if(XVFB_EXECUTABLE)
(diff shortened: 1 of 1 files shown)


noc0lour:
Oh, that looks interesting, I'll check it out.

marcusmueller:
I like this change, Andrej, what's your stance on this?

Also, throwing this out there: http://xpra.org/trac/wiki/Xdummy

noc0lour:
I've played around with some virtual X a few years back and it didn't work, I'll test this when I have some time. Otherwise if someone else wants to test this and confirms it's working well we can merge this and I enable qtgui tests.

noc0lour:
Ping me in 14 days again.

gr-qtgui/python/qtgui/CMakeLists.txt
```diff
@@ -48,10 +48,17 @@ if(ENABLE_TESTING)
     ${CMAKE_BINARY_DIR}/gr-qtgui/swig
     )
 
+  if(LINUX AND NOT DEFINED ENV{DISPLAY})
```
noc0lour, line 51: We definitely don't want to detect this at compile/config time but everytime we run a test with qtgui.

noc0lour:
Although for starters this should be fine.

PR #3547, opened 2020-06-06
cmake: use TIMESTAMP function to generate build_date and save to cache

noc0lour:
Previously a a custom python command was used to generate the build date
but this can be done with native CMake functions for all supported CMake
versions.

Additionally it's compatible with reproducible builds since CMake 3.8 as
it honors SOURCE_DATE_EPOCH set in the environment varibales.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -532,12 +532,16 @@ CONFIGURE_FILE(
   ${CMAKE_CURRENT_BINARY_DIR}/gnuradio-runtime/include/gnuradio/config.h
 )
 
-#Re-generate the constants file, now that we actually know which components will be enabled.
-configure_file(
-    ${CMAKE_CURRENT_SOURCE_DIR}/gnuradio-runtime/lib/constants.cc.in
-    ${CMAKE_CURRENT_BINARY_DIR}/gnuradio-runtime/lib/constants.cc
-    ESCAPE_QUOTES
-@ONLY)
+########################################################################
+# Handle the generated constants
+########################################################################
+
+# Use TIMESTAMP to be compatible with reproducible builds
+# and put in in the cache so configure_file sees it
+string(TIMESTAMP BUILD_DATE "%a, %d %b %Y %H:%M:%S")
+set(BUILD_DATE ${BUILD_DATE} CACHE INTERNAL "Build date")
+message(STATUS "Loading build date ${BUILD_DATE} into constants...")
+message(STATUS "Loading version ${VERSION} into constants...")
 
 # Install config.h in include/gnuradio
 install(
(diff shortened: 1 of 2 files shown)


skoslowski:
Nice. However, putting it in cache makes this the "time of **initial** configure". That is different from what you had before.

jdemel:
I guess you created a duplicate in #3548 . 

I didn't follow the discussion on Matrix closely but I thought you wanted to put a commit timestamp in the build.

noc0lour:
@skoslowski if you show me the way without `CACHE` I would be very thankful :) I spend about an hour to figure out `CACHE` is required to make it work at all.

Currently `BUILD_DATE` in `configure_file` is empty.

noc0lour:
@jdemel Nope, second pull request is a backport to 3.8

skoslowski:
@noc0lour sure, I'll you the way =)

So, this file seems to be generated twice. first in runime/lib where you made the change and then again later at the bottom of the main list file. By putting the value in cache you extend its scope beyond the declaring subfolder (runime/lib) - i.e. making it a global. This also extends its lifetime: a new timestamp in a subsequent configure will be overriden by the value in cache.

Solution:
* only generate the file once at the time when you have gathered all the info
* if you really want to keep the logic in the runtime/lib, put it in a function and call that from the main list file

noc0lour:
@skoslowski 

Oh wow, yeah now everything makes sense, I literally did not check if it gets generated a second time and clearly this is the main issue here. 

Previously instead of add_subdirectory we had this namespace-clobbering `include_subdirectory` function which included all variables in the parent scope and then it obviously worked.

I need to rework then to generate this file only once.
Thanks!

noc0lour:
@skoslowski good catch, forgot about changing the line to remove CACHE.

noc0lour:
Can you change that within the github review thing? And then I just accept the change?

CMakeLists.txt
```diff
@@ -532,7 +532,24 @@ CONFIGURE_FILE(
...
+# and put in in the cache so configure_file sees it
+string(TIMESTAMP BUILD_DATE "%a, %d %b %Y %H:%M:%S")
+set(BUILD_DATE ${BUILD_DATE} CACHE INTERNAL "Build date")
+message(STATUS "Loading build date ${BUILD_DATE} into constants...")
+message(STATUS "Loading version ${VERSION} into constants...")
+
+#double escape for windows backslash path separators
+string(REPLACE "\\" "\\\\" prefix "${prefix}")
+string(REPLACE "\\" "\\\\" SYSCONFDIR "${SYSCONFDIR}")
+string(REPLACE "\\" "\\\\" GR_PREFSDIR "${GR_PREFSDIR}")
```
skoslowski, line 549: Before, these values were only modified with the scope of gnuradio-runtime/lib and therefore not visible here. I don't have (nor want to setup) a windows box to verify that. Not saying it isn't useful, just noting this is probably changing the result.

noc0lour: Uh, oh, maybe they have some use in `gnuradio-runtime/lib` I'll check that. Thanks.

marcusmueller:
This would be a good thing to have in 3.9, I guess. @noc0lour do you think it's ready for merge?

mbr0wn:
Merged manually.

PR #3548, opened 2020-06-06
cmake: use TIMESTAMP function to generate build_date and save to cache

noc0lour:
Previously a a custom python command was used to generate the build date
but this can be done with native CMake functions for all supported CMake
versions.

Additionally it's compatible with reproducible builds since CMake 3.8 as
it honors SOURCE_DATE_EPOCH set in the environment varibales.


noc0lour:
Will reopen once fixed properly on master.

PR #3552, opened 2020-06-07
blocks/moving_average: use vector functions

ThomasHabets:
This improves performance of moving average for me by about 4-5x.

My test is [here](https://github.com/ThomasHabets/radiostuff/blob/master/amateur/listen_70cm.grc), which processes 10MHz to find the strongest signal.

Without this PR I see `moving_average` in `top` taking about 89.4-94%
CPU. With this patch it's ~20.4-23.7%. Since without this patch it's
that high, I don't know that it's not even better.

Test measured on a Libremv2 with Intel Core i7-6500U @ 2.5GHz.

The memory access pattern is probably worse with this patch, but at
least on my hardware on my workload this seems to be dwarfed by the
gain of using volk. It can be hard to reason about memory access patterns, so benchmarks overrule theory.

The test only actually uses float averaging.

More benchmarking may be required.

Possible improvements:
* Add template specialization for `uint16` and `uint32`?
* Refactor for non-volk fallback to use only one loop, with
  (presumably) better memory access pattern

### DIFF
--- a/gr-blocks/lib/moving_average_impl.cc
+++ b/gr-blocks/lib/moving_average_impl.cc
@@ -15,9 +15,90 @@
 
 #include "moving_average_impl.h"
 #include <gnuradio/io_signature.h>
+#include <volk/volk.h>
+#include <numeric>
 
 namespace gr {
 namespace blocks {
+
+namespace {
+template <typename T>
+inline void volk_add(T* out, const T* add, unsigned int num)
+{
+    for (unsigned int elem = 0; elem < num; elem++) {
+        out[elem] += add[elem];
+    }
+}
+
+template <>
+inline void volk_add<float>(float* out, const float* add, unsigned int num)
+{
+    volk_32f_x2_add_32f(out, out, add, num);
+}
+
+template <>
+inline void volk_add<gr_complex>(gr_complex* out, const gr_complex* add, unsigned int num)
+{
+    volk_32fc_x2_add_32fc(out, out, add, num);
+}
+
+template <typename T>
+inline void volk_sub(T* out, const T* sub, unsigned int num)
+{
+    for (unsigned int elem = 0; elem < num; elem++) {
+        out[elem] -= sub[elem];
+    }
+}
+
+template <>
+inline void volk_sub<float>(float* out, const float* sub, unsigned int num)
+{
+    volk_32f_x2_subtract_32f(out, out, sub, num);
+}
+
+// For gr_complex, do some checking that the types are as we expect before
+// allowing the specialization.
+// That long `enable_if` boils down to the word `void`.
+template <>
+inline typename std::enable_if<std::is_same<gr_complex::value_type, float>::value>::type
+volk_sub<gr_complex>(gr_complex* out, const gr_complex* sub, unsigned int num)
+{
+    // std::complex is required to be implemented by adjacent and non-padded
+    // imaginary and real values, so it's safe to treat as such.
+    auto fout = reinterpret_cast<gr_complex::value_type*>(out);
+    auto fsub = reinterpret_cast<const gr_complex::value_type*>(sub);
+
+    // But just to be sure, let's double-check that.
+    static_assert(sizeof(gr_complex) == 2 * sizeof(gr_complex::value_type),
+                  "Can't happen: sizeof(gr_complex) != 2*sizeof(gr_complex::value_type)");
+
+    volk_32f_x2_subtract_32f(fout, fout, fsub, 2 * num);
+}
+
+template <typename T>
+inline void volk_mul(T* out, const T* in, const T* scale, unsigned int num)
+{
+    for (unsigned int elem = 0; elem < num; elem++) {
+        out[elem] = in[elem] * scale[elem];
+    }
+}
+
+template <>
+inline void
+volk_mul<float>(float* out, const float* in, const float* mul, unsigned int num)
+{
+    volk_32f_x2_multiply_32f(out, in, mul, num);
+}
+
+template <>
+inline void volk_mul<gr_complex>(gr_complex* out,
+                                 const gr_complex* in,
+                                 const gr_complex* mul,
+                                 unsigned int num)
+{
+    volk_32fc_x2_multiply_32fc(out, in, mul, num);
+}
+} // namespace
 
 template <class T>
 typename moving_average<T>::sptr
@@ -44,11 +125,11 @@ moving_average_impl<T>::moving_average_impl(int length,
       d_updated(false)
 {
     this->set_history(length);
-    // we don't have C++11's <array>, so initialize the stored vector instead
     // we store this vector so that work() doesn't spend its time allocating and freeing
     // vector storage
     if (d_vlen > 1) {
-        d_sum = std::vector<T>(d_vlen);
+        d_sum = volk::vector<T>(d_vlen);
+        d_scales = volk::vector<T>(d_vlen, d_scale);
     }
 }
 
@@ -87,21 +168,18 @@ int moving_average_impl<T>::work(int noutput_items,
     if (d_updated) {
         d_length = d_new_length;
         d_scale = d_new_scale;
+        std::fill(std::begin(d_scales), std::end(d_scales), d_scale);
         this->set_history(d_length);
         d_updated = false;
         return 0; // history requirements might have changed
     }
 
-    const T* in = (const T*)input_items[0];
-    T* out = (T*)output_items[0];
+    const T* in = static_cast<const T*>(input_items[0]);
+    T* out = static_cast<T*>(output_items[0]);
 
-    unsigned int num_iter =
-        (unsigned int)((noutput_items > d_max_iter) ? d_max_iter : noutput_items);
+    const unsigned int num_iter = std::min(noutput_items, d_max_iter);
     if (d_vlen == 1) {
-        T sum = in[0];
-        for (int i = 1; i < d_length - 1; i++) {
-            sum += in[i];
-        }
+        T sum = std::accumulate(&in[0], &in[d_length - 1], T{});
 
         for (unsigned int i = 0; i < num_iter; i++) {
             sum += in[i + d_length - 1];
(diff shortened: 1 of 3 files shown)


ThomasHabets:
More benchmarks. This test shows more like 1.5-3x, but still a big improvement:

```
$ for V in 3.8.1.0 master volk-moving; do echo $V; grgr $V c++ -O6 -ggdb3 -I$HOME/opt/gnuradio-$V/include -L$HOME/opt/gnuradio-$V/lib -o moving-speed moving-speed.cc -lgnuradio-blocks -lboost_system -llog4cpp && grgr $V  ./moving-speed;done
3.8.1.0
vlen   128 (   524288) : 0.065478
vlen  1024 (    65536) : 0.562154
vlen  4096 (    16384) : 2.348729
master
vlen   128 (   524288) : 0.065552
vlen  1024 (    65536) : 0.556610
vlen  4096 (    16384) : 2.288686
volk-moving
vlen   128 (   524288) : 0.024458
vlen  1024 (    65536) : 0.365433
vlen  4096 (    16384) : 1.649784
```

```c++
#include <gnuradio/blocks/moving_average.h>

#include <sys/time.h>
#include <sys/resource.h>

#include <vector>
#include <iostream>

int main() {
  const auto vlens = std::vector<int>{128, 1024, 4096};
  const auto max_vlen = *std::max_element(std::begin(vlens), std::end(vlens));
  const auto len = 1024;
  const auto in_size = 16384;
  const auto work_calls = 100;
  const auto max_iter = 4096;

  const auto in = std::vector<float>(in_size * max_vlen, 1.618);
  std::vector<float> out(in_size * max_vlen);
  std::vector<const void*> inv{in.data()};
  std::vector<void*> outv{out.data()};

  std::vector<gr::blocks::moving_average<float>::sptr> blocks;
  blocks.reserve(vlens.size());
  for (const auto& vlen : vlens) {
    blocks.push_back(gr::blocks::moving_average<float>::make(len, 1.0, max_iter, vlen));
  }

  for (int i = 0 ; i < vlens.size(); i++) {
    const auto& vlen = vlens[i];
    auto& block = blocks[i];
    const auto out_size = in_size * max_vlen/vlen;

    struct rusage before;
    struct rusage after;
    if (getrusage(RUSAGE_SELF, &before)) {
      perror("getrusage");
      return 1;
    }

    // Call benchmarked code.                                                                                                                   for (int i = 0; i < work_calls; i++) {
      block->work(out_size, inv, outv);
    }

    if (getrusage(RUSAGE_SELF, &after)) {
      perror("getrusage");
      return 1;
    }
    after.ru_utime.tv_sec -= before.ru_utime.tv_sec;
    if (after.ru_utime.tv_usec < before.ru_utime.tv_usec) {
      after.ru_utime.tv_sec--;
      after.ru_utime.tv_usec += 1000000;
    }
    after.ru_utime.tv_usec -= before.ru_utime.tv_usec;
    std::cout << "vlen "
              << std::setfill(' ') << std::setw(5) << vlen
              << " (" << std::setw(9) << out_size << ") "
              << ": "
              << after.ru_utime.tv_sec << "."
              << std::setfill('0') << std::setw(6) << after.ru_utime.tv_usec
              << std::endl;
  }
}
```

This time on 6 core (2 HT/core) Intel Xeon W-2135 @ 3.7GHz.

jacobgilbert-snl:
I would be curious to see how the performance before/after compares to using one of the conventional FIR kernels to implement a moving average.

gr-blocks/lib/moving_average_impl.cc
```diff
@@ -44,11 +46,11 @@ moving_average_impl<T>::moving_average_impl(int length,
       d_updated(false)
 {
     this->set_history(length);
-    // we don't have C++11's <array>, so initialize the stored vector instead
     // we store this vector so that work() doesn't spend its time allocating and freeing
     // vector storage
     if (d_vlen > 1) {
-        d_sum = std::vector<T>(d_vlen);
+        d_sum = volk::vector<T>(d_vlen);
+        d_scales = volk::vector<T>(d_vlen, d_scale);
```
rear1019, line 53: Is this really required? `volk_32f_s32f_normalize()` or `volk_32f_s32f_multiply_32f()` can be used instead of `volk_32f_x2_multiply_32f()`, even for the `gr_complex` case (apply type cast and double the vector length).

[1] https://www.libvolk.org/doxygen/volk_32fc_s32fc_multiply_32fc.html
[2] https://www.libvolk.org/doxygen/volk_32f_s32f_normalize.html

gr-blocks/lib/moving_average_impl.cc
```diff
@@ -87,21 +150,18 @@ int moving_average_impl<T>::work(int noutput_items,
...
 
-    unsigned int num_iter =
-        (unsigned int)((noutput_items > d_max_iter) ? d_max_iter : noutput_items);
+    const unsigned int num_iter = std::min(noutput_items, d_max_iter);
     if (d_vlen == 1) {
-        T sum = in[0];
-        for (int i = 1; i < d_length - 1; i++) {
-            sum += in[i];
-        }
+        T sum = std::accumulate(&in[0], &in[d_length - 1], T{});
```
rear1019, line 164: `volk_32f_accumulator_s32f()`

[1] https://www.libvolk.org/doxygen/volk_32f_accumulator_s32f.html

gr-blocks/lib/moving_average_impl.cc
ThomasHabets: For `gr_complex` that assumes the scaling factor is not complex, no?

ThomasHabets:
With your suggestions, three runs:

```
$ for i in $(seq 3); do for V in volk-moving; do echo $V; grgr $V c++ -O6 -I$HOME/opt/gnuradio-$V/include -L$HOME/opt/gnuradio-$V/lib -o moving-speed moving-speed.cc -lgnuradio-blocks -lboost_system -llog4cpp && grgr $V  ./moving-speed;done;done
volk-moving
vlen   128 (   524288) : 0.027964
vlen  1024 (    65536) : 0.408475
vlen  4096 (    16384) : 1.737717
volk-moving
vlen   128 (   524288) : 0.028238
vlen  1024 (    65536) : 0.410601
vlen  4096 (    16384) : 1.748107
volk-moving
vlen   128 (   524288) : 0.027350
vlen  1024 (    65536) : 0.398560
vlen  4096 (    16384) : 1.724082
```

Re-running as-is:

```
$ for i in $(seq 3); do for V in volk-moving; do echo $V; grgr $V c++ -O6 -I$HOME/opt/gnuradio-$V/include -L$HOME/opt/gnuradio-$V/lib -o moving-speed moving-speed.cc -lgnuradio-blocks -lboost_system -llog4cpp && grgr $V  ./moving-speed;done;done
volk-moving
vlen   128 (   524288) : 0.024035
vlen  1024 (    65536) : 0.360620
vlen  4096 (    16384) : 1.634118
volk-moving
vlen   128 (   524288) : 0.024813
vlen  1024 (    65536) : 0.369788
vlen  4096 (    16384) : 1.606274
volk-moving
vlen   128 (   524288) : 0.025274
vlen  1024 (    65536) : 0.367620
vlen  4096 (    16384) : 1.623408
```

So weirdly it seems slower with your suggestions.

The code with your suggestions is [here](https://github.com/ThomasHabets/gnuradio/commit/6b926ba63927a68f69efb62eabe2ebb4f7b22368).

ThomasHabets:
I'll benchmark the equivalent flowgraph using FIR filters when I have some more time.

ThomasHabets:
Actually of all the changes I made I'm not sure anymore what the best combination is.

My [other branch](https://github.com/ThomasHabets/gnuradio/blob/volk-moving-second/gr-blocks/lib/moving_average_impl.cc) with all the changes seems fastest, but I I don't yet know if backing out some changes would make it even faster.

I am aware that using volk functions like this on maybe unaligned data could be one problem.

I'm also not entirely convinced about the test coverage. E.g. doesn't look like `scale` is actually tested.

ThomasHabets:
Is there a FIR block that takes vectors, though?

gr-blocks/lib/moving_average_impl.cc
jdemel: That's correct. A moving average with a complex scaling factor would be odd though. I'm fine with it either way.

gr-blocks/lib/moving_average_impl.cc
```diff
@@ -79,6 +81,67 @@ void moving_average_impl<T>::set_scale(T scale)
...
+    }
+}
+
+template <>
+inline void volk_sub<float>(float* out, const float* sub, unsigned int num)
+{
+    volk_32f_x2_subtract_32f(out, out, sub, num);
+}
+
+// TODO: Add volk_sub specialization for gr_complex?
```
jdemel, line 118: That would be useful to add this todo. I suggest:
```cpp
template <>
inline void volk_sub<gr_complex>(gr_complex* out, const gr_complex* sub, unsigned int num)
{
    volk_32f_x2_subtract_32f(out, out, sub, 2 * num);
}
```

ThomasHabets: Doh. Of course. Uploading a new version of the commit as soon as I've run the tests.

ThomasHabets: Added som paranoia compile time tests.

gr-blocks/lib/moving_average_impl.cc
ThomasHabets: `std::accumulate` I think vectorizes pretty well already, according to my experimentation.

Sure, volk can do more at runtime to choose the best kernel. For now though I'd prefer to leave this for a future PR.

rear1019: `std::accumulate` does vectorize when applied to integer types, but not when using float types – unless you use `-ffast-math` (according to my experimentation).

mormj:
Three approvals, mighty performance improvement! Merging.

PR #3553, opened 2020-06-08
blocks: Fix XMLRPC Client block templates

argilo:
The XMLRPC Client block is currently broken because it tries to import 
the Python 2 XMLRPC module, and because the "addr" and "callback" 
parameters are inserted as quoted strings. To fix these problems, I've 
switched to the Python 3 version of the XMLRPC module, and changed the 
data types of the problematic parameters to "raw".

I tested this with the following flowgraphs:

Server:

Screenshot from 2020-06-07 21-53-46

Client:

Screenshot from 2020-06-07 21-53-56

I'll open a separate pull request to fix this in GNU Radio 3.8, which needs to support both Python 2 and Python 3.

### DIFF
--- a/gr-blocks/grc/xmlrpc_client.block.yml
+++ b/gr-blocks/grc/xmlrpc_client.block.yml
@@ -5,7 +5,7 @@ flags: [ python ]
 parameters:
 -   id: addr
     label: Address
-    dtype: string
+    dtype: raw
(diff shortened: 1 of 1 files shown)


argilo:
The 3.8 backport is in #3554.

mormj:
Looks good to me!

Issue #3011, opened 2020-01-05
Backport SimpleXMLRPCServer fix to 3.8


mbr0wn:
See discussion on https://github.com/gnuradio/gnuradio/pull/2963

maitbot:
Fixed in ed0f6d7dc630ed388beb97cb4e9f887717f55179
This Issue should be closed.

willcode:
Fix mentioned looks like it addresses the issue mentioned. Closing.

PR #3554, opened 2020-06-08
blocks: Fix XMLRPC Client and Server block templates

argilo:
The XMLRPC Client block is currently broken because it tries to import 
the Python 2 XMLRPC module, and because the "addr" and "callback" 
parameters are inserted as quoted strings. To fix these problems, I've 
switched to the Python 3 version of the XMLRPC module with a fallback to 
the Python 2 version, and changed the data types of the problematic 
parameters to "raw".

The XMLRPC Server block is currently broken because it tries to import 
the Python 2 XMLRPC module. To fix this problem, I've switched to the 
Python 3 version of the XMLRPC module with a fallback to the Python 2 
version.

I tested this with the following flowgraphs:

Server:

Screenshot from 2020-06-07 21-53-46

Client:

Screenshot from 2020-06-07 21-53-56

### DIFF
--- a/gr-blocks/grc/xmlrpc_client.block.yml
+++ b/gr-blocks/grc/xmlrpc_client.block.yml
@@ -5,7 +5,7 @@ flags: [ python ]
 parameters:
 -   id: addr
     label: Address
-    dtype: string
+    dtype: raw
     default: localhost
 -   id: port
     label: Port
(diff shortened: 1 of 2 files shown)


argilo:
This should fix #2782.

argilo:
This also fixes #3011.

PR #3555, opened 2020-06-08
gr-uhd: Fix assumption that pmt pair passes is_dict

jdemel:
Previously, `pmt::is_dict` would return `true` for pairs and vice versa.
This assumption was built into gr-uhd command handling. With this
commit, a new check `if(pmt::is_pair(...))` is introduced to convert
pairs to dicts for further processing.

This behavior is [documented here](https://www.gnuradio.org/doc/doxygen/namespacepmt.html#a2b5a9b0e21ac7a90078ce5033182104e) but apparently, it does not hold anymore.

The new behavior looks like this:
```cpp
pmt::is_pair(pmt::make_dict()); //return false.
pmt::is_dict(pmt::cons(pmt::intern("key), pmt::intern("val))); //return false.
```
When did that change?

### DIFF
--- a/gr-uhd/lib/usrp_block_impl.cc
+++ b/gr-uhd/lib/usrp_block_impl.cc
@@ -487,22 +487,19 @@ void usrp_block_impl::msg_handler_command(pmt::pmt_t msg)
     }
     // End of legacy backward compat code.
 
+    // Apparently, pmt::is_dict does not return true for pairs anymore. Yeah!
+    if (pmt::is_pair(msg)) {
+        GR_LOG_DEBUG(d_logger,
+                     boost::format("Command message is pair, converting to dict: %s") %
+                         msg);
+        msg = pmt::dict_add(pmt::make_dict(), pmt::car(msg), pmt::cdr(msg));
+    }
+
     // Turn pair into dict
     if (!pmt::is_dict(msg)) {
         GR_LOG_ERROR(d_logger,
                      boost::format("Command message is neither dict nor pair: %s") % msg);
         return;
-    }
-
-    // OK, here comes the horrible part. Pairs pass is_dict(), but they're not dicts. Such
-    // dicks.
-    try {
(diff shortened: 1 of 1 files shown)


jacobgilbert-snl:
This changed recently in #3389. I did update the documentation on master but did not fix uhd.

jdemel:
I think I need to update my initial description. While
```cpp
pmt::is_dict(pmt::cons(pmt::intern("key), pmt::intern("val))); //return false
```
still:
```cpp
pmt::is_pair(pmt::make_dict()); //return true
```
because a dict is a subclass of `pmt::pair` now. I guess I need to update to:
```cpp
    if (pmt::is_dict(msg)) {
        GR_LOG_DEBUG(d_logger, boost::format("Command message is dict: %s") % msg);
    }
    else if (pmt::is_pair(msg)) {
        GR_LOG_DEBUG(d_logger, boost::format("Command message is pair, converting to dict: %s") % msg);
        msg = pmt::dict_add(pmt::make_dict(), pmt::car(msg), pmt::cdr(msg));
    }
```
Otherwise, gr-uhd is still broken for dicts.

marcusmueller:
Please also write a "unit testy"-example to test whether this works

mrjacobagilbert:
@jdemel you are correct about dicts/pairs. i looked into making dicts _not_ pass `is_pair()` but that would have had major consequence.

The simplest fix to retain existing behavior would be:
```c++
    // Turn pair into dict
    if (!(pmt::is_dict(msg) || pmt::is_pair(msg))) {
        GR_LOG_ERROR(d_logger,
                     boost::format("Command message is neither dict nor pair: %s") % msg);
        return;
    }
```

jdemel:
I'd love to see this PR merged. Is there anything blocking?

mrjacobagilbert:
I'd also like to see this merged; the changes look good to me and work as expected.

PR #3562, opened 2020-06-10
Replace SWIG with pybind11

mormj:
This is the collection of changes to replace SWIG with pybind11.  There will be many implications to the workflow and the API, and OOTs will have to be changed to be compatible with the pybind11 branch.

Things that are *known* to not be working at this point:
- ctrlport
- modtool disable
- qa tests: uncaught_exception, header_payload_demux, qa_uhd

Information on usage and workflow here:
https://github.com/gnuradio/gnuradio/blob/pybind/docs/PYBIND11.md
and detailed information about OOT porting is/will be here:
https://wiki.gnuradio.org/index.php/GNU_Radio_3.9_OOT_Module_Porting_Guide


mormj:
So the buildbots don't have pybind11 :(  Will have to figure out how to get that included since we have not vendorized at this point.

noc0lour:
@mormj I kicked the PR runner and now they build (minus the forseen test failures)

drmpeg:
Commit https://github.com/gnuradio/gnuradio/pull/3562/commits/3c302ef4c1655ff27c3f66a924d897a6fdd07060 fixes the issue of all bindings in a component being rebuilt when a module changes.

However, I believe the fix goes too far. When a docstring in a .h file is changed, the binding is not rebuilt.

marcusmueller:
@drmpeg good point, will have to check that. I've merged nevertheless, because otherwise this gets stale and Josh ends up in an endless rebase loop...

mormj:
@marcusmueller thank you for merging and adding issue #3578 has been added.
@drmpeg - this is a good catch, still I think the current state is better than the constant rebuilding of all modules, but I will look into how to get the docstring scraping run appropriately on individual files being changed.  Probably some better cmake magic than is currently implemented

noc0lour:
Yeah, if you set up the depndencies correctly (not sure this is even possible) it should be able to run this automatically.

PR #3563, opened 2020-06-11
uhd: Add PC clock resync command

jdemel:
USRPs may be synchronized to a host clock. Over time USRP and host clock
deviate. This is an issue for timed commands. With this commit it is
possible to resync both devices.

The new command key is `pc_clock_resync`. The command does not expect
any argument.

It may be possible that USRP time is not strictly monotonic anymore. If
that is a requirement, other options should be explored.

e.g. connect a 'Message Strobe' block to a USRP command port and send a
message with `pmt.dict_add(pmt.make_dict(), pmt.intern('pc_clock_resync'), pmt.PMT_T)`
periodically. It is sufficient to send such a message every couple of
seconds.

### DIFF
--- a/gr-uhd/lib/usrp_block_impl.cc
+++ b/gr-uhd/lib/usrp_block_impl.cc
@@ -96,6 +96,11 @@ const pmt::pmt_t gr::uhd::cmd_tag_key()
     static const pmt::pmt_t val = pmt::mp("tag");
     return val;
 }
+const pmt::pmt_t gr::uhd::cmd_pc_clock_resync_key()
+{
+    static const pmt::pmt_t val = pmt::mp("pc_clock_resync");
+    return val;
+}
 
 const pmt::pmt_t gr::uhd::ant_direction_rx()
 {
@@ -141,6 +146,7 @@ usrp_block_impl::usrp_block_impl(const ::uhd::device_addr_t& device_addr,
     REGISTER_CMD_HANDLER(cmd_rate_key(), _cmd_handler_rate);
     REGISTER_CMD_HANDLER(cmd_bandwidth_key(), _cmd_handler_bw);
     REGISTER_CMD_HANDLER(cmd_antenna_key(), _cmd_handler_antenna);
+    REGISTER_CMD_HANDLER(cmd_pc_clock_resync_key(), _cmd_handler_pc_clock_resync);
 }
 
 usrp_block_impl::~usrp_block_impl()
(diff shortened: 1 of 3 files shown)


mbr0wn:
@jdemel 

```
/cache/workdir/Debian_10_64_0/build_Debian_10_64/src/gr-uhd/lib/usrp_block_impl.cc: In member function ‘void gr::uhd::usrp_block_impl::_cmd_handler_pc_clock_resync(const pmt_t&, int, const pmt_t&)’:
/cache/workdir/Debian_10_64_0/build_Debian_10_64/src/gr-uhd/lib/usrp_block_impl.cc:684:14: error: ‘std::chrono’ has not been declared
         std::chrono::duration_cast<std::chrono::nanoseconds>(
              ^~~~~~
/cache/workdir/Debian_10_64_0/build_Debian_10_64/src/gr-uhd/lib/usrp_block_impl.cc:684:41: error: ‘std::chrono’ has not been declared
         std::chrono::duration_cast<std::chrono::nanoseconds>(
                                         ^~~~~~
/cache/workdir/Debian_10_64_0/build_Debian_10_64/src/gr-uhd/lib/usrp_block_impl.cc:685:18: error: ‘std::chrono’ has not been declared
             std::chrono::high_resolution_clock::now().time_since_epoch())
                  ^~~~~~
make[2]: *** [gr-uhd/lib/CMakeFiles/gnuradio-uhd.dir/build.make:76: gr-uhd/lib/CMakeFiles/gnuradio-uhd.dir/usrp_block_impl.cc.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:7679: gr-uhd/lib/CMakeFiles/gnuradio-uhd.dir/all] Error 2
```

dkozel:
This method is good enough to sync the USRP to the host time with a few ms resolution, but doesn't make any use of an external PPS if available. I think that's OK, but a check to see if the time source is configured to external and warning that this doesn't use the PPS would be helpful in clarifying it to users.

dkozel (approved):
This method should reliably set the USRP time to the host time with just a few ms of delay and jitter. It is potentially confusing to a user if they have an external PPS connected and expect that PPS to be used automatically. I think it's ok not to try to detect and use the PPS since that'd be a slightly different thing than keeping sync with the PC clock.

I think it would be actively helpful to have a check if the time source is set to External and throw out a warning to the user that the PPS is unused.

jdemel:
@mbr0wn I wonder why `std::chrono` can't be found. Was that a temporary hiccup or does this need more investigation?

@dkozel The command name is `pc_clock_resync`. The initial value in GRC is `Sync: PC Clock`. I thought, a PPS or GPS sync would be considered different. How should this be made more obvious?

gr-uhd/lib/usrp_block_impl.cc
```diff
@@ -670,6 +676,18 @@ void usrp_block_impl::_cmd_handler_rate(const pmt::pmt_t& rate_, int, const pmt:
     set_samp_rate(rate);
 }
 
+void usrp_block_impl::_cmd_handler_pc_clock_resync(const pmt::pmt_t&,
+                                                   int,
+                                                   const pmt::pmt_t&)
+{
+    const uint64_t ticks =
```
jdemel, line 683: Should we add a warning here? smth like:
```cpp
if(clock_source() != "internal"){
    GR_LOG(WARN, "Clock source mismatch: Syncing USRP to PC clock but USRP is configured to an external source!");
}

dkozel: Yes. This is what I was imagining. Though it'd be Time Source.

> Time source mismatch: Setting USRP Time to PC clock but USRP is configured to use an external PPS. The PPS will be ignored.

Something like this?

nickoe:
@jdemel Are you going to follow up on this?

jdemel:
I'll work on this.
@mbr0wn what about your error? Does it still appear or was it a hiccup?

mbr0wn:
Merged manually (resolved the GPIO conflict).

PR #3565, opened 2020-06-12
qtgui: remove optional flag on stream ports

### LINKED ISSUE #3036: QT gui time sink doesn't validate missing input connections
The QT GUI time sink doesn't complain that nothing is connected to it:

image

which means that missing connections throw exceptions at runtime, but should be prevented in GRC:

image

The above flowgraph throws a topology error at runtime.

mormj:
The qtgui blocks are all labeled as optional on their input ports, which is only appropriate for message blocks.  This PR puts a conditional flag on the optional parameter for the input ports

Fixes #3036


marcusmueller:
Does this need 3.8 backportin'?

mormj:
Yes - I'll throw that in later today

PR #3489, opened 2020-05-17
boost: qualify placeholders with their full namespace.

Ponce:
This is needed with boost >= 1.73.0.

PR #3566, opened 2020-06-12
boost: qualify placeholders with their full namespace. (Backport to 3.8)

noc0lour:
This is needed with boost >= 1.73.0.

This is just a backport of #3489

### DIFF
--- a/docs/usage-manual/(exported from wiki) Message Passing.txt
+++ b/docs/usage-manual/(exported from wiki) Message Passing.txt
@@ -94,7 +94,7 @@ must then bind this port to the message handler. For this, we use
 Boost's 'bind' function:
 
   set_msg_handler(pmt::pmt_t port_id,
-    boost::bind(&block_class::message_handler_function, this, _1));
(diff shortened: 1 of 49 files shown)


noc0lour:
failing Ubuntu 18.04 test is unrelated, message passing \o/

jdemel:
I suggest to move to lambdas instead of fully qualified placeholders. @ThomasHabets benchmarked them and the result was basically: use lambdas.

noc0lour:
Is there a separate path for master we can backport? I really don't mind which solution is chosen, I just want to compile it on my machine ;)

jdemel:
You could use #3441 and just change those 2 lines in `basic_block` such that you use `boost::function` instead of `std::function` again.
Unfortunately, the development history of this PR was, I want `std::function` and then I fixed all the issues that arose. Instead of, let's just move to lambdas and then just do the trivial switch to `std::function`.

noc0lour:
Will see that I can pull out the lambda commit and apply it on 3.8.

mbr0wn:
Inactive, needs rebase, closing.

PR #3572, opened 2020-06-16
blocks: fix tag propagation in the selector block

### LINKED ISSUE #3569: runtime/blocks (?): tag propagation anomalies when selectively dropping data
While debugging a simple block, we noticed some unusual tag propagation behavior when dropping large amounts of data. This can be demonstrated using the in-tree selector block as follows:

image

Initially, the tags are correct:

image

However, by selecting and deselecting the `Enabled` check box a couple times, its clear the tag propagation has some errors:

image

This example was put together with only in-tree blocks, but using the `tag_uhd_offset_f` block from gr-timing_utils, its possible to see that old tags remain in the output buffer with newer data (runtime does not account for the updated tag placement in the output buffer now that some data has been dropped?).

We attributed this to a runtime issue, as it appears that runtime appears to have sufficient information to propagate tags correctly (consuming X samples, producing 0 samples), though this is a bit of a special case of `general_work()` and propagation can also be fixed in the block directly via manual tag propagation in the selector block.

Regardless, tags should propagate correctly through this block. Thoughts on the correct approach?

This affects 3.7 (after the new selector was added), 3.8, and master. Flowgraph is here: [tagpropagation.zip](https://github.com/gnuradio/gnuradio/files/4774544/tagpropagation.zip)

mrjacobagilbert:
previously tags were propagated automatically which after dropping data on an output caused stale tags to be propagated with data; dropped data is now tracked for each output and tag offsets are updated accordingly.

Fixes #3569

### DIFF
--- a/gr-blocks/lib/selector_impl.cc
+++ b/gr-blocks/lib/selector_impl.cc
@@ -35,6 +35,7 @@ selector_impl::selector_impl(size_t itemsize,
             io_signature::make(1, -1, itemsize)),
       d_itemsize(itemsize),
       d_enabled(true),
+      d_dropped_samples(0),
       d_input_index(input_index),
       d_output_index(output_index),
       d_num_inputs(0),
@@ -42,6 +43,10 @@ selector_impl::selector_impl(size_t itemsize,
 {
     message_port_register_in(pmt::mp("en"));
     set_msg_handler(pmt::mp("en"), [this](pmt::pmt_t msg) { this->handle_enable(msg); });
+
+    // tag propagation is handled manually: tags from the selected input stream propagate
+    // to the selected output stream only
+    set_tag_propagation_policy(TPP_CUSTOM);
 
     message_port_register_in(pmt::mp("iindex"));
     set_msg_handler(pmt::mp("iindex"),
@@ -134,6 +139,8 @@ bool selector_impl::check_topology(int ninputs, int noutputs)
     if ((int)d_input_index < ninputs && (int)d_output_index < noutputs) {
         d_num_inputs = (unsigned int)ninputs;
         d_num_outputs = (unsigned int)noutputs;
+        d_dropped_samples.clear();
+        d_dropped_samples.resize(d_num_outputs, 0);
         return true;
     } else {
         GR_LOG_WARN(d_logger,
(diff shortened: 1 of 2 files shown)


mrjacobagilbert:
Question: Should this block use `TPP_CUSTOM` instead of `TPP_DONT`? Seems like identical behavior with different meaning.

mormj:
Yes, I would think a `TPP_CUSTOM` would make more sense and just pass tags from the selected input to the selected output.  Tracking the dropped samples seems overly-complicated.

mrjacobagilbert:
I must have misunderstood what TPP_CUSTOM does. I agree tracking dropped samples per output is complicated but it seems like even if using TPP_CUSTOM, this is still required to ensure correct sample alignment.

mormj:
Oooh - looks like DONT and CUSTOM are handled the same in the scheduler, and only used as a flag for the block to behave differently.  Functionally equivalent but semantically, I'm not sure if CUSTOM is more correct?  

I'm not following what the expected tag behavior should be as it relates to the dropped samples.  Wouldn't the expected behavior just be to copy tags from the selected input to the selected output within the given window (since it is essentially a sync block between selected inputs) - in which case why would dropped samples need to be tracked?

mrjacobagilbert:
Issue #3569 has more detail but let’s say you disable the selector and 100 items are consumed but none produced. If there is a tag on the 101st item, it should be on the first produced item, but without updating the offset the tag will end up on the 101st output item (201st input item). 

Similar problematic behavior holds true for selecting different signal processing paths. 

This also happens to fix counterintuitive behavior where tags from deselected inputs propagate. 

The two tag propagation policies seem to differ only in that TPP_DONT seems to be a code level indication for situations where input tags are not expected to propagate, and TPP_CUSTOM for situations like this. I’ll switch it.

marcusmueller:
This looks pretty good! Would it be possible to add a test case that would fail exactly when tag offsets aren't handled correctly?

mrjacobagilbert:
I’ll put something together for that.

One weakness of this PR is that if the block topology change during runtime tag propagation will break. I see the handlers to accomplish this but I am not really sure how that would work in practice. Thoughts?

gr-blocks/lib/selector_impl.cc
```diff
@@ -152,10 +159,32 @@ int selector_impl::general_work(int noutput_items,
 
     gr::thread::scoped_lock l(d_mutex);
     if (d_enabled) {
+
+        // propagate tags
+        std::vector<tag_t> tags;
+        get_tags_in_window(tags, 0, 0, noutput_items);
+        for (auto tag : tags) {
+            tag.offset -= d_dropped_samples[d_output_index];
```
rear1019, line 167: If I am not mistaken, `tag.offset = nitems_written(d_output_index)` could be used instead of tracking the number of dropped samples.

mrjacobagilbert:
Resolved in #3712

PR #3574, opened 2020-06-17
Move fft functions and blocks to templates

noc0lour:
To reduce code duplication and prepare for some template action in `qtgui` gr-fft is moved to templates. 

This is still WIP, I haven't written complete specialization for `fft_vfc` & `fft_vcf` blocks for forward and inverse real FFT.


noc0lour:
Also some boost `_1` compatibility commits are still in this PR to compile them on my machine.

gnuradio-runtime/lib/controlport/rpcserver_aggregator.cc
```diff
@@ -10,6 +10,8 @@
 
 #include <gnuradio/rpcserver_aggregator.h>
 #include <gnuradio/rpcserver_booter_base.h>
+
+#include <algorithm>
```
noc0lour, line 14: Oops, this should be in a separate commit.
Also, why isn't this already in the code? is nobody building rpcserver_aggregator.cc ?

noc0lour:
Changes API in two places:
 - `fft_complex` becomes `fft_complex_fwd` and `fft_complex_rev`
 - `polyphase_filterbank` only used `fft_complex_rev` since this is what all of polyhpase_filterbanks was used in-tree. Don't know about usage in the wild.

gnuradio-runtime/lib/controlport/rpcserver_aggregator.cc
marcusmueller: *crickets*

dkozel:
I think the overall change to templating the FFT is great, less code duplication and readability is fine. In most cases I can think of only forward or reverse is needed, not the ability to swap between them at runtime, so any external users of the API will probably be able to adapt quite easily.

The Polyphase Filterbank seems like it'd be pretty easy to adapt to still have support for forward and reverse FFTs. That'd be nice to have and there do seem to be usecases for it at the application level, just we don't have any in GR Core.

dkozel (changes requested):
I think the Polyphase Filterbank should still support inverse FFT usage. Looking at the code that seems pretty straightforward to retain, though probably means having two FFT objects and using the flag to select which is inited and used.

mormj:
Closing and rebase continued (now merged) in #3903.

PR #3579, opened 2020-06-19
cmake: Remove FindSWIG from modules and don't check for it

noc0lour:
Remove the FindSWIG module and don't check for SWIG in the
main CMakeLists


noc0lour:
Don't be sorry, it's alright :)

noc0lour:
Now I have to exclude/fix the tests :-P

PR #3582, opened 2020-06-19
Updated signal source block to accept amplitude - message command

alekhgupta1441:
This PR helps to make signal source accept message pairs which can help in dynamically changing **amplitude** in addition to presently available for only **frequency**.

### DIFF
--- a/gr-analog/grc/analog_sig_source_x.block.yml
+++ b/gr-analog/grc/analog_sig_source_x.block.yml
@@ -41,7 +41,7 @@ parameters:
 
 inputs:
 -   domain: message
-    id: freq
+    id: cmd
     optional: true
 
 outputs:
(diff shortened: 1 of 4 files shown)


dkozel:
@marcusmueller How do you feel about the API change here caused by renaming the message port from `freq` to `cmd`? The actual behavior is preserved (numerical only messages are interpreted as frequency tunes). We could keep the `freq` command port and map it and `cmd` to the same handler for 3.9 and remove the `freq` in 3.10?

marcusmueller:
@dkozel I'd prefer if we also kept the `freq` port around, yes. Let me add that to this PR real quick.

marcusmueller:
superseded by #3588. Alekh, please review that PR!

PR #3585, opened 2020-06-20
Modtool: remove remaining SWIG template, improve C++ template overall

marcusmueller:


marcusmueller:
> Instead of specifically asking the users to provide `descriptive_test_name`, should we just write `default_test`?
> 
> This looks good as well though, users will rename the function according because of this.

That was my hope!

marcusmueller:
> Finally, no existence of Py2k.

It's such a relief, isn't it?

swap-nil7:
Yeah, it is. :)

PR #3588, opened 2020-06-20
signal source: accept commands on cmd port, work with dicts

marcusmueller:
also:
extend message functionality to include setting of phase, amplitude and offset.

Adds an example flowgraph.

Supersedes #3582

### DIFF
--- a/gr-analog/lib/sig_source_impl.cc
+++ b/gr-analog/lib/sig_source_impl.cc
@@ -53,9 +53,12 @@ sig_source_impl<T>::sig_source_impl(double sampling_freq,
 {
     this->set_frequency(frequency);
     this->set_phase(phase);
+    this->message_port_register_in(pmt::mp("cmd"));
+    this->set_msg_handler(pmt::mp("cmd"),
+                          [this](pmt::pmt_t msg) { this->set_cmd_msg(msg); });
     this->message_port_register_in(pmt::mp("freq"));
     this->set_msg_handler(pmt::mp("freq"),
-                          [this](pmt::pmt_t msg) { this->set_frequency_msg(msg); });
+                          [this](pmt::pmt_t msg) { this->set_freq_msg(msg); });
 }
 
 template <class T>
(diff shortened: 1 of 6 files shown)


gr-analog/lib/sig_source_impl.cc
```diff
@@ -64,32 +67,80 @@ sig_source_impl<T>::~sig_source_impl()
...
+            if (pmt::is_number(val)) {
+                set_amplitude(pmt::to_double(val));
+            } else {
+                GR_LOG_WARN(this->d_logger, "amplitude value needs to be a number")
+            }
+        } else if (key == phase_key) {
+            if (pmt::is_number(val)) {
+                set_phase(pmt::to_double(val));
+            } else {
+                GR_LOG_WARN(this->d_logger, "phaseitude value needs to be a number")
```
dkozel, line 128: @marcusmueller Some copypasta with: phaseitude and offsetitude

gr-analog/lib/sig_source_impl.cc
```diff
@@ -64,32 +67,80 @@ sig_source_impl<T>::~sig_source_impl()
...
+{
+    static auto freq_key = pmt::intern("freq");
+    static auto ampl_key = pmt::intern("ampl");
+    static auto phase_key = pmt::intern("phase");
+    static auto offset_key = pmt::intern("offset");
     // Accepts either a number that is assumed to be the new
     // frequency or a key:value pair message where the key must be
     // "freq" and the value is the new frequency.
 
     if (pmt::is_number(msg)) {
```
dkozel, line 89: Since you have the nice deprecation function above, how about moving this legacy behavior into the `set_freq_msg` function so code that updates also is ensured to move to the key-pair format?

gr-analog/python/analog/qa_sig_source.py
```diff
@@ -198,7 +198,7 @@ def test_freq_msg(self):
         self.assertAlmostEqual(src.frequency(), 1.0)
 
         frequency = 3.0
-        src._post(pmt.to_pmt('freq'), pmt.from_double(frequency))
+        src._post(pmt.to_pmt('cmd'), pmt.from_double(frequency))
```
dkozel, line 201: We can keep this test and add another it to test setting frequency, phase, amplitude, and offset using key/value pairs.

alekhgupta1441:
@marcusmueller Thanks for such quick upgarde to the PR. To me it looks quite extensive and far sighted changes, and should be ready to go, after some minor corrections mentioned by @dkozel .

gr-analog/lib/sig_source_impl.cc
marcusmueller: oh thanks for your diligencitude!

gr-analog/lib/sig_source_impl.cc
```diff
@@ -64,32 +67,77 @@ sig_source_impl<T>::~sig_source_impl()
...
+    if (pmt::is_dict(msg)) {
+        list_of_items = pmt::dict_items(msg);
     } else if (pmt::is_pair(msg)) {
-        pmt::pmt_t key = pmt::car(msg);
-        pmt::pmt_t val = pmt::cdr(msg);
-        if (pmt::eq(key, pmt::intern("freq"))) {
+        list_of_items = pmt::list1(msg);
+    } else {
+        GR_LOG_WARN(this->d_logger,
+                    "malformed message: is neither plain number, nor dict or pair");
```
dkozel, line 99: Plain numbers are only accepted by the calling set_freq_msg function. Should this message include plain numbers?

gr-analog/python/analog/qa_sig_source.py
```diff
@@ -203,6 +209,30 @@ def test_freq_msg(self):
 
         self.assertAlmostEqual(src.frequency(), frequency)
 
+    def test_cmd_msg(self):  # deprecated but still tested
```
dkozel, line 212: Copy paste - This test is not deprecated!

gr-analog/lib/sig_source_impl.cc
marcusmueller: no, it shouldn't!

marcusmueller:
merged myself after OK from Derek via chat

drmpeg:
@marcusmueller @dkozel This commit has a bug. The following in printed when GRC is started on the command line.
```
WARNING:gnuradio.grc.core.platform:analog_sig_source_x.block.yml            block.inputs[1]: error: Value type 'int' for 'hide' not in valid types
INFO:gnuradio.grc.core.platform:YAML schema check failed for: /opt/gnuradio-3.9.0git/share/gnuradio/grc/blocks/analog_sig_source_x.block.yml
```

Here's the location of the problem. https://github.com/gnuradio/gnuradio/blob/master/gr-analog/grc/analog_sig_source_x.block.yml#L50

dkozel:
Thanks @drmpeg 
This also prompts the idea of adding a YAML check to the CI, though that should have been noticible while testing.

PR #3589, opened 2020-06-20
blocks: remove log2_const

marcusmueller:
Was not standards-compliant (assert(0) in constexpr)

Only used in packed_to_unpacked to get the log2(number of bits(type)) in
packed_to_unpacked<type>.

Was not wrapped to python nor public API.

To little surprise, a static method compiles faster, and gets
compile-time-calculated by a modern compiler, too.


ThomasHabets:
It's not guaranteed to be inline, though. With C++14 we can slap a `constexpr` on the function and guarantee it.

Alt this should be constexprable in C++11:

```c++
constexpr int log2(int i)
{
  static_assert(!(i & (i-1), "blah");
  return i == 1 ? 0 :
      i == 2 ? 1 :
     […];
}
```

ThomasHabets:
Actually that won't work. This will, though:

```c++

template<int i>
constexpr int log2() noexcept
{
  static_assert(!(i & (i-1)) && i <= 1024, "blah");
  return i == 1 ? 0 :
      i == 2 ? 1 :
      i == 4 ? 2 :
      i == 8 ? 3 :
      i == 16 ? 4 :
      i == 32 ? 5 :
      i == 64 ? 6 :
      i == 128 ? 7 :
      i == 256 ? 8 :
      i == 512 ? 9 :
      1024;
}
constexpr int foo = log2<1024>();
```

marcusmueller:
@ThomasHabets hah! Like all of these, but instead of the ternary expression chain, we could just template metaprogram it:

specialize 

     constexpr unsigned int log2<1>(){return 0;}


recurse down 

    constexpr uint log2<N>(){
         static_assert(!(N&(N-1)));
         return log2<N/2>;
    }

ThomasHabets:
Nice.

mbr0wn:
@marcusmueller were you going to change this to use your template foo? Or are you done here?

gr-blocks/lib/packed_to_unpacked_impl.cc
```diff
@@ -62,19 +62,33 @@ void packed_to_unpacked_impl<T>::forecast(int noutput_items,
     }
 }
 
+template <class T>
+unsigned int packed_to_unpacked_impl<T>::log2_l_type()
+{
+    unsigned int val = sizeof(T);
```
mbr0wn, line 68: I would have preferred 
```suggestion
    unsigned int val = sizeof(T) * 8;
```
instead of the + 3 below, but I'll classify that as a matter of taste.

PR #3590, opened 2020-06-20
blocks: use posix_fadvise to tell OS about sequential access and prefetch

marcusmueller:


gr-blocks/lib/file_source_impl.cc
```diff
@@ -188,7 +188,34 @@ void file_source_impl::open(const char* filename,
...
     // Rewind to start offset
     if (d_seekable) {
-        if (GR_FSEEK(d_new_fp, start_offset_items * d_itemsize, SEEK_SET) == -1) {
+        auto start_offset = start_offset_items * d_itemsize;
+#ifdef _POSIX_C_SOURCE
+#if _POSIX_C_SOURCE >= 200112L
+        // If supported, tell the OS that we'll be accessing the file sequentially
+        // and that it would be a good idea to start prefetching it
+        auto fd = fileno(d_new_fp);
+        static const std::map<int, const std::string> fadv_errstrings = {
```
daniestevez, line 197: Just wondering: is there a reason for using this ad-hoc map instead of `strerror_r()`?

mbr0wn:
Does this have a measurable performance effectd?

gr-blocks/lib/file_source_impl.cc
marcusmueller: not really, no.

marcusmueller:
> Does this have a measurable performance effectd?

I mainly did this to avoid some strange effects I saw on Reiner's NTP mounts when repeating small files. I haven't actually gotten around benchmarking it on these machines since then, though.

I have to admit that benchmarking on my SSD array doesn't show a measurable advantage. But then again, as soon as your file is larger than a couple of megabytes, prefetching would happen anyway, and on smaller files: well, sooo many confounding factors.

mormj (approved):

PR #3596, opened 2020-06-22
grc: do not run stale flowgraph

### LINKED ISSUE #3580: GRC: when rendering fails, old flowgraph is executed
When a flowgraph is created that creates some issue in the mako rendering (because of a bug usually), text is displayed in the console stating the error, but the previously generated python file will be executed after the failed rendering.

This can create confusion as to the stale flowgraph that is being executed, since it won't match the GRC canvas at that point.

Issue can be simulated by raising an exception inside this try block
https://github.com/gnuradio/gnuradio/blob/75286c00a3ff8175a05a9a6d2a1482b05ff8d95b/grc/gui/Application.py#L712

mormj:
When something bad happens in the rendering of a flowgraph, usually due to a mako generation failure, the previously rendered flowgraph with the same name is run.  This can be particularly confusing and mask the underlying error.

Fixes #3580

### DIFF
--- a/grc/gui/Application.py
+++ b/grc/gui/Application.py
@@ -702,6 +702,7 @@ class Application(Gtk.Application):
         # Gen/Exec/Stop
         ##################################################
         elif action == Actions.FLOW_GRAPH_GEN:
+            self.generator = None
             if not page.process:
                 if not page.saved or not page.file_path:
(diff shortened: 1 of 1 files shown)


mbr0wn:
Qualifies for backport, I'd say.

PR #3597, opened 2020-06-22
uhd: Add default value to uhd.time_spec call

jdemel:
With pybind11 `uhd.time_spec()` doesn't work anymore because default
values are not defined.


mormj:
Also added a pull request #3602 so that the default values are added in the bindings themselves.  Still not a bad idea to explicitly set them in the call to time_spec_t

PR #3600, opened 2020-06-23
gr-blocks: Transition the WAV sink and source blocks to libsndfile.

drmpeg:
Compressed input and output with FLAC and Ogg Vorbis now supported.


drmpeg:
CI will fail until the libsndfile dependency is added to the build machine.

mormj:
Used the dialtone example to test:

image


and then looked at the generated files:

image

image

Which look accurate plus some compression noise

image

drmpeg:
@mormj Is there a reason that the ```set_append``` method was left out of the binding? I think it should be added back in.

mormj:
I can't find any reason why it wouldn't be.  Just ran

```
python3 /share/gnuradio/grdev/src/gnuradio/gr-utils/bindtool/scripts/bind_intree_file.py --output_dir /share/tmp/ --prefix /share/gnuradio/grdev --src /share/gnuradio/grdev/src/gnuradio --module blocks --filename /share/gnuradio/grdev/src/gnuradio/gr-blocks/include/gnuradio/blocks/wavfile_sink.h 
```

and it made an entry for set_append in the generated file here.

drmpeg:
I meant why was it missing from your original commit. Anyway, I've put it back in just now.

mormj:
Ah ok - not sure why that wasn't in the bindings for the original block.  The bindings for the in-tree blocks were done iteratively and I believe there were some updates to wavefile_sink along the way.  But still not sure why that one callback function got skipped.  Thank you for putting it back.

mormj:
@drmpeg - did some more testing to verify that the libsnd based blocks are compatible with files produced from the legacy wavfile_sink and vice versa with the wavfile_source - and unsurprisingly we're in good shape there.

A couple of things we should consider before merging:
1) The blocks should be placed under gr-audio instead of gr-blocks.  Not sure why the original wavfile blocks were under blocks, but they are clearly for audio (or using an audio format)
2) Since it goes beyond just wav files now, the blocks should also be named more generally, such as audio_file_source/ audio_file_sink.  

Since the compatibility works in both directions, we can just put in a python shim to have blocks/wavfile_{source/sink} and remove the old blocks (and deprecate the shim I guess).

Again, this is a huge improvement over the existing functionality and thank you for being the first to go through the effort of pybind11 binding a new block.

drmpeg:
I'm still waiting for libsndfile to be added to the build machine.

drmpeg:
After looking at the structure of gr-audio, I believe the current pull request is the way to go. Please merge now that libsndfile is available on the build machine.

gr-blocks/include/gnuradio/blocks/wavfile_sink.h
```diff
@@ -35,12 +36,15 @@ class BLOCKS_API wavfile_sink : virtual public sync_block
...
+     * \param format Output format (WAV, FLAC, Ogg Vorbis, RF64)
+     * \param subformat Bits per sample
+     * \param append Append to existing file
      */
     static sptr make(const char* filename,
                      int n_channels,
                      unsigned int sample_rate,
-                     int bits_per_sample = 16,
+                     wavfile_format_t format,
+                     wavfile_subformat_t subformat,
```
argilo, line 47: Just an FYI -- I noticed this breaking API change when attempting to build Gqrx against 3.9.

https://github.com/csete/gqrx/blob/57334b493aac5a3c17aee909dba9c4349e02bee9/src/applications/gqrx/receiver.cpp#L1000

Gqrx can easily work around this with an `#ifdef`, as it already does in various places to maintain compatibility with multiple GNU Radio versions simultaneously. I just wanted to double check that the breaking change is intentional here. (It could presumably be avoided by overloading `make` but I'm not convinced it's worth doing that.)

argilo: On second thought, the change to `std::shared_ptr` for blocks is already a much bigger API change in 3.9, so I don't think maintaining compatibility here is of any value.

drmpeg: Yes, it's intentional. The WAV file sink and source in 3.9 have been upgraded to handle more file formats including FLAC and Ogg Vorbis (selected by wavfile_format_t) . Also, more WAV file bit depths (and float) are supported (selected by wavfile_subformat_t)..

It does require a new dependency, libsndfile1-dev. This should be resolved automatically in OOT's that include gnuradio-blocks.

PR #3603, opened 2020-06-24
grc: do not run stale flowgraph

mormj:
Backport of #3596 to 3.8


mbr0wn:
@mormj sometimes I'm just a dumbo.

PR #3609, opened 2020-06-26
uhd: Fix missing bindings

jdemel:
uhd python bindings were missing a couple of functions.

1. `uhd.ALL_MBOARDS` was not defined.
2. `uhd.streams_args` was incomplete.
Instead of import magic, we define a pybind constructor which supports
all cases. This should ease the maintainance burden as well.

Be aware: some notes that `this file was generated with bindtool` were
removed because this files contain manual edits. The foundations were
probably generated with bindtool once.


gr-uhd/python/uhd/__init__.py
```diff
@@ -46,26 +43,6 @@ def __init__(self, *args, **kwargs):
             for key, val in list(kwargs.items()): setattr(self, key, val)
     setattr(uhd_python, 'tune_request_t', tune_request_t)
 
-    #make the streamer args take **kwargs on init
-    class stream_args_t(uhd_python.stream_args_t):
```
mormj, line 50: Glad to be rid of this wrapper

mormj:
Tagging a few more devs to approve, but this is a pretty critical one, so should be merged soon

mormj:
@drmpeg - thanks for reviewing.  Merging.

PR #3611, opened 2020-06-26
uhd: wrap get_usrp_info to compatible return type

### LINKED ISSUE #3608: uhd: pybind cannot convert get_usrp_info to python type
If one creates a uhd source (or sink) and calls `get_usrp_info` on it, the following error is thrown:
```python
>>> d = uhd.usrp_source(device_args, stream_args)
>>> info = d.get_usrp_info()
TypeError: Unable to convert function return value to a Python type! The signature was
	(self: gnuradio.uhd.uhd_python.usrp_source, chan: int = 0) -> uhd::dict<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > >

Did you forget to `#include <pybind11/stl.h>`? Or <pybind11/complex.h>,
<pybind11/functional.h>, <pybind11/chrono.h>, etc. Some automatic
conversions are optional and require extra headers to be included
when compiling your pybind11 module.
```

mormj:
Copied the wrapping mechanism into a lambda from pyuhd for these functions

Fixes #3608

### DIFF
--- a/gr-uhd/lib/usrp_block_impl.cc
+++ b/gr-uhd/lib/usrp_block_impl.cc
@@ -741,3 +741,9 @@ void usrp_block_impl::_cmd_handler_dspfreq(const pmt::pmt_t& dspfreq,
 
     _update_curr_tune_req(new_tune_request, chan);
 }
+
+::uhd::dict<std::string, std::string> usrp_block_impl::get_usrp_info(size_t chan)
+{
+    chan = _stream_args.channels[chan];
+    return _dev->get_usrp_rx_info(chan);
+}
--- a/gr-uhd/python/uhd/bindings/usrp_sink_python.cc
+++ b/gr-uhd/python/uhd/bindings/usrp_sink_python.cc
@@ -57,12 +57,6 @@ void bind_usrp_sink(py::module& m)
              D(usrp_sink, set_start_time))
 
 
-        .def("get_usrp_info",
-             &usrp_sink::get_usrp_info,
-             py::arg("chan") = 0,
-             D(usrp_sink, get_usrp_info))
-
-
         .def("get_lo_names",
              &usrp_sink::get_lo_names,
              py::arg("chan") = 0,
--- a/gr-uhd/include/gnuradio/uhd/usrp_block.h
+++ b/gr-uhd/include/gnuradio/uhd/usrp_block.h
@@ -572,6 +572,15 @@ public:
      * @return the filter object
      */
     virtual ::uhd::filter_info_base::sptr get_filter(const std::string& path) = 0;
+
+    /*!
+     * Returns identifying information about this USRP's configuration.
+     * Returns motherboard ID, name, and serial.
+     * Returns daughterboard TX ID, subdev name and spec, serial, and antenna.
+     * \param chan channel index 0 to N-1
+     * \return TX info
+     */
+    virtual ::uhd::dict<std::string, std::string> get_usrp_info(size_t chan = 0) = 0;
 };
 
 } /* namespace uhd */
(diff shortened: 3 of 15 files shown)


gr-uhd/python/uhd/bindings/usrp_sink_python.cc
```diff
@@ -57,10 +57,13 @@ void bind_usrp_sink(py::module& m)
...
-             &usrp_sink::get_usrp_info,
-             py::arg("chan") = 0,
-             D(usrp_sink, get_usrp_info))
+        .def(
+            "get_usrp_info",
+            [](usrp_sink& self, const size_t chan = 0) {
+                return static_cast<std::map<std::string, std::string>>(
+                    self.get_usrp_info(chan));
+            },
+            py::arg("chan") = 0)
```
jdemel, line 66: Except for `usrp_sink` this is exactly the same code as in `usrp_source`.
Can you move this to `usrp_block`?

mormj: Good point - no reason to have duplicate code.  Second commit is added to address this.

mormj:
Struggling to get CI to pass on this one (it builds on my machine lol).  Perhaps there is a different version of pybind11 between the buildbot and mine.  more investigation.

mormj:
Unfortunately I haven't been able to get this to compile on the docker used in CI - perhaps it is the difference in gcc.  Will have to investigate how to do this with gcc7 or whatever other tools are in the ci containers

gr-uhd/lib/usrp_block_impl.cc
```diff
@@ -741,3 +741,9 @@ void usrp_block_impl::_cmd_handler_dspfreq(const pmt::pmt_t& dspfreq,
 
     _update_curr_tune_req(new_tune_request, chan);
 }
+
+::uhd::dict<std::string, std::string> usrp_block_impl::get_usrp_info(size_t chan)
+{
+    chan = _stream_args.channels[chan];
+    return _dev->get_usrp_rx_info(chan);
```
ncorgan, line 748: The `multi_usrp` class has both `get_usrp_tx_info` and `get_usrp_rx_info` functions, with some different information between each. The separate implementations in `usrp_source` and `usrp_sink` were intentional for this reason. Consolidating these into `usrp_block` removes getting TX-specific info.

jdemel: That's a very good reason to keep them separate. But also, it's a very good reason to update these function names.
I suggest to add `get_usrp_rx_info` to `usrp_source` and `get_usrp_tx_info` to `usrp_sink`. Further, I suggest to keep `get_usrp_info` in the base class, deprecate it and log a warning everytime it is called. Alternatively, we can remove it. But I doubt this approach receives approval without deprecation for at least one release cycle.
Anyways, I'd consider it a bug that the function does not fully describe what it does and needs to be fixed.

ncorgan: The `get_usrp_*x_info` solution, with deprecation of the old function name, sounds like a good plan.

mbr0wn: Every other function in these blocks has the same name for the sink and source (like set gain, set freq) and is declared in usrp_block, and then defined in the sink or source. I don't see a necessity to change that here.

mbr0wn:
I'm closing this in favour of https://github.com/gnuradio/gnuradio/pull/3691, which compiles and doesn't have the drawbacks mentioned here. @jdemel, @ncorgan, @mormj I'd be very happy if you could take another look there.

PR #3613, opened 2020-06-28
gr-qtgui: Replace PyLong_AsLong with PyLong_AsLongLong for 32-bit sys…

drmpeg:
…tems.


mormj:
Tested local on 64 bit and widgets work the same.  Thanks for the fix.

PR #3615, opened 2020-06-29
blocks: Use templates more efficiently in add_blk

jdemel:
Compare recent changes in the moving_average block with those templatized
`volk_add` functions instead of whole work functions being templatized.

It should now be easier to add further templates for other block types.
A volkized version for gr_complex was added.

### DIFF
--- a/gr-blocks/lib/add_blk_impl.cc
+++ b/gr-blocks/lib/add_blk_impl.cc
@@ -20,6 +20,29 @@
 namespace gr {
 namespace blocks {
 
+namespace {
+template <typename T>
+inline void volk_add(T* out, const T* add, unsigned int num)
+{
+    for (unsigned int elem = 0; elem < num; elem++) {
+        out[elem] += add[elem];
+    }
+}
+
+template <>
+inline void volk_add<float>(float* out, const float* add, unsigned int num)
+{
+    volk_32f_x2_add_32f(out, out, add, num);
+}
+
+template <>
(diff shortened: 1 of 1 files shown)


marcusmueller (approved):

marcusmueller:
Nice; invalidates my #3899, but that's fine with me. The chanest test case fails for me, too. It's wonky, for some reason.

jdemel:
CI says this PR is fine. What should we do with the test that fails for you? Ignore it as an outlier?

gr-blocks/lib/add_blk_impl.cc
```diff
@@ -34,25 +57,19 @@ add_blk_impl<float>::add_blk_impl(size_t vlen)
...
-    const int alignment_multiple = volk_get_alignment() / sizeof(float);
-    set_alignment(std::max(1, alignment_multiple));
+    set_alignment(std::max(1, int(volk_get_alignment() / sizeof(float))));
 }
 
 template <>
-int add_blk_impl<float>::work(int noutput_items,
-                              gr_vector_const_void_star& input_items,
-                              gr_vector_void_star& output_items)
+add_blk_impl<gr_complex>::add_blk_impl(size_t vlen)
```
rear1019, line 64: Just wondering: Is this specialization of the constructor (and the one for `float`) required? Was it ever required? (As per C++ rules – my knowledge of templates is limited…)

rear1019 (approved):

mbr0wn:
Three approvals and I can't repro the CI issue... so it's going in.

PR #3617, opened 2020-07-01
uhd: remove import from uhd

mormj:
There is no need to have the dependence on uhd python bindings as the
types that are used in gr-uhd are wrapped separately

This especially causes issues when using the PPA packages, since uhd doesn't by default package their python bindings, so this unnecessary line will cause an exception


mormj:
Two reviews - thanks!  Merging.

PR #3620, opened 2020-07-02
modtool: Run all generated C++ code through clang-format

ThomasHabets:
Forked off of https://github.com/gnuradio/gnuradio/pull/3377 since that was left unreviewed to the point where it could not be salvaged.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -410,6 +410,12 @@ install(
     DESTINATION ${GR_PKG_DOC_DIR}
 )
 
+install(
+    FILES .clang-format
+    RENAME clang-format.conf
+    DESTINATION ${GR_PKG_DATA_DIR}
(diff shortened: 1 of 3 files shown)


drmpeg:
This should be optional. I find the current .clang-format rules to be hideous and would much rather not use them in my OOT.

ThomasHabets:
> This should be optional. I find the current .clang-format rules to be hideous and would much rather not use them in my OOT.

It is inevitable that the generated code will have *some* style (the only code that doesn't have a style is a 0 byte file). Currently it's a mess with *no* style. It's not like this PR forces you to keep the same style in your OOT module.

ThomasHabets:
@drmpeg fixed!

mormj:
Approvals.  Merging.

PR #3621, opened 2020-07-04
Add a metainfo/appstream file for GRC, so that distro appstores find us

marcusmueller:
This might warrant a backport to maint-3.8, too.

### DIFF
--- a/grc/scripts/freedesktop/CMakeLists.txt
+++ b/grc/scripts/freedesktop/CMakeLists.txt
@@ -27,6 +27,9 @@ install(FILES gnuradio-grc.desktop DESTINATION share/applications)
 # Install mime
 install(FILES gnuradio-grc.xml DESTINATION share/mime/packages)
 
+# Install appstream / metainfo file
+install(FILES org.gnuradio.grc.metainfo.xml DESTINATION share/metainfo)
+
 # Install icons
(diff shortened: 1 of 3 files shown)


grc/scripts/freedesktop/CMakeLists.txt
```diff
@@ -55,3 +58,9 @@ if(UNIX AND HAVE_XDG_UTILS)
     )
     endif(ENABLE_POSTINSTALL)
 endif(UNIX AND HAVE_XDG_UTILS)
+
+find_program(APPSTREAMCLI appstreamcli)
+
+if(UNIX AND APPSTREAMCLI)
+  GR_ADD_TEST(metainfo_test "${APPSTREAMCLI}" validate "${CMAKE_CURRENT_SOURCE_DIR}/org.gnuradio.grc.metainfo.xml")
```
mbr0wn, line 65: Increase indent by 2

grc/scripts/freedesktop/org.gnuradio.grc.metainfo.xml
```diff
@@ -0,0 +1,60 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
```
mbr0wn, line 2: Is this comment something to be concerned about?

marcusmueller: Yes. That's why I added it.

PR #3632, opened 2020-07-09
GPIO handler (via stream tags/messages), works with timed commands

segdy:
gr-uhd does not provide any ability to manipulate GPIO (except by directly calling member functions).

This pull request fixes that and adds a message handler for gpio that can be used via strema tags or messages and works in combination with timed commands etc.

It adds a new command `gpio` having a dict as parameter. It contains the usual fields `bank`, `attr`, `value` and `mask`


segdy:
Note: The checks say `bb/Ubuntu_14_04_64_py2_qt4_maint-3.7` failed. It is because a test for `qa_udp_source_sink` fails. I don't think this can be related to my pull request.

mbr0wn:
Merged manually (had to squash commits).

PR #3634, opened 2020-07-10
Add start time capability to UHD USRP blocks in grc 3.8

segdy:
Currently, set_start_time cannot be set in GRC.
When this is needed, the python file has to be changed manually which is overwritten when saving the next time.

This patch adds set_start_time to USRP Source/Sink.
It only shows up when it is set. Normally, it is set to Default which results in identical behavior than now. It can be set to an arbitrary float value indicating a time to start in seconds.

Same as https://github.com/gnuradio/gnuradio/pull/3498 but for 3.8


mbr0wn (approved):
This is a no-brainer.

mbr0wn:
Merged manually (had to resolve conflicts).

PR #3636, opened 2020-07-10
 analog: signal source block accepts cmd messages (Backport 3.8)

alekhgupta1441:
Backport of #3588 for signal source block.


dkozel:
Just checked and this has the bug (`hide: 1`) from the master PR fixed. Thanks for making this!

alekhgupta1441:
Yes @dkozel  It had another bug in qa_sig_source.py [here](https://github.com/gnuradio/gnuradio/pull/3636/files#diff-4059371da6a9b97b1a15375b63faa94dR143).

Issue #3593, opened 2020-06-21
WBFM_Receive_PLL block distorted audio


duggabe:
The sound has a lot of distortion but there are no audio underruns. The traces on the Time Sink have a lot of periodic high-frequency "fuzz" (it may be the 19khz pilot carrier).

Using the code from https://github.com/gnuradio/gnuradio/blob/master/gr-analog/python/analog/wfm_rcv_fmdet.py with several modifications, I built a FM stereo receiver using only basic blocks. See https://wiki.gnuradio.org/index.php/File:USRP_FM_stereo_fg.png

My original intent was just to use the Python code to construct the block, but then discovered several places where the numbers were wrong or where a block would not accept the parameters used.

duggabe:
revised https://wiki.gnuradio.org/index.php/File:USRP_FM_stereo_fg.png per Ron Economos:
1. Gain of Quadrature Demod
2. PLL Ref Out values.
3. Addition of Complex to Imag block.
4. Different stereo matrix.
5. Import math to get Pi constant.
6. Offset tuning on USRP source. Gets rid of noise when audio transitions to silence.

It works now.

PR #3640, opened 2020-07-15
correct fm stereo heir block, add example grc

duggabe:
resolves https://github.com/gnuradio/gnuradio/issues/3593

### DIFF
--- a/gr-analog/examples/USRP_FM_stereo.grc
+++ b/gr-analog/examples/USRP_FM_stereo.grc
@@ -157,7 +157,7 @@ blocks:
     bus_sink: false
     bus_source: false
     bus_structure: null
-    coordinate: [488, 200.0]
+    coordinate: [488, 168.0]
     rotation: 0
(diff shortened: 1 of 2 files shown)


duggabe:
1. My work was based on http://www.w6rz.net/USRP_FM_stereofft.grc  Were the Left and Right channels reversed in that flowgraph? Working through the math, it looks like they would be. The two Multiply Const blocks seem to have the signs reversed.
2. Does a complex band pass filter with a low cutoff freq of -80k and a high cutoff freq of +80k translate into a complex low pass filter with a cutoff freq of 80k or 160k?
3. I didn't know I could (or needed to) set the file permissions in the master branch. That issue has never come up before.

777arc:
@drmpeg Thanks for the tips!  Should be pretty easy to address, let me know if you need help Barry.

mormj:
LGTM.  Merging.

PR #3642, opened 2020-07-15
uhd: Control of frequency and gain in both directions at the same moment

segdy:
This is a new revision of https://github.com/gnuradio/gnuradio/pull/3151.

After talking with Piotr, while he would like to see this patch integrated, he does not have the resources right now.

This commit fixes some issue with the original pull request.

From the original pull request:

It implements ability to control frequency and gain in **both** directions (Tx and Rx) with use of a single UHD source or sink - in the same moment. It is possible to do that in separate moments with current UHD blocks but this way it is not possible to change both Rx and Tx settings for the same burst. This pull request improves this situation.

This is next incarnation of recently closed pull request #2426.
Look at my (a bit long) explanation in the last comment there in case of doubt that such functionality have any practical application.

PR #3643, opened 2020-07-15
uhd: Control of frequency and gain in both directions at the same moment (maint-3.8 port)

segdy:
This is the the port for `maint-3.8` of https://github.com/gnuradio/gnuradio/pull/3642


mbr0wn:
This change is a bit too big for 3.8 at this point. We are open to discussions if someone really needs it, but I'll close this PR.

PR #3646, opened 2020-07-16
Get rid of unnecessary 'from __future__ import'

### LINKED ISSUE #3584: get rid of `from __future__ import stuff`
`unicode_literals` was necessary pre-Python 2.6; `absolute_import` pre-2.5.

It's especially not necessary on Py3.

Hence in the sence of cleaning up the source tree, deletion of all `from __future__ import something` from the python files is desirable. This should ideally not happen manually, but with a script, or command line: we need to do that on `master`, but it might also be desirable to offer tooling to OOT developers.

Reference for the `__future__` module: https://docs.python.org/3.6/library/__future__.html#module-__future__

(this issue should be reserved for beginners, I think. It's a good way to learn building GNU Radio, running tests, making a Pull Request.)

sashko:
All of the removed `from __future__ import` were needed in older
versions of Python (mostly 2.5.x and below) but later became mandatory
in most versions of Python 3 hence are not necessary anymore.

More specifically, according to __future__.py[1]:
- unicode_literals is part of Python since versions 2.6.0 and 3.0.0;
- print_function is part of Python since versions 2.6.0 and 3.0.0;
- absolute_import is part of Python since versions 2.5.0 and 3.0.0;
- division is part of Python since versions 2.2.0 and 3.0.0;

Get rid of those unnecessary imports to slightly clean up the codebase.

Fixes #3584.

[1] https://github.com/python/cpython/blob/master/Lib/__future__.py


mormj:
@sashko - thank you for the contribution - do you have a CLA in place? If not, please email "bhilburn AT gnuradio.org" for a form to start the copyright process.  This one may fly under the level of trivial fix since it is just removing code, but I see you have more contributions in place as well, which is great!

sashko:
> @sashko - thank you for the contribution - do you have a CLA in place? If not, please email "bhilburn AT gnuradio.org" for a form to start the copyright process. This one may fly under the level of trivial fix since it is just removing code, but I see you have more contributions in place as well, which is great!

I have emailed him a week ago, but no reply yet. Is he on vacation?

sashko:
@bhilburn

mbr0wn:
I think we can  grant an exception here since the CLA has been triggered by @sashko, but, more importantly, it only removes LOCs using a `sed`-like command.

mbr0wn:
@sashko Thanks for removing 800+ LOCs :)

Merged manually (minor rephrasing of the git commit message).

Issue #3362, opened 2020-04-13
Convert remaining printf debugging to GNU Radio logging


marcusmueller:
After #3339, we still have a few places in-tree where things are just printed out to stdout. 

Seek and destroy!

japm48:
For reference: https://gist.github.com/japm48/2f89966a03a4bcfb1f050e536340a297/raw/c804a1f5758952251a96a3572370e7b4f15ecd7a/search.txt

Although many of those lines are part of comments that are should probably be removed anyways.

BTW @marcusmueller, what are your thoughts on replacing `boost::format` with something like `fmt` ([link](https://github.com/fmtlib/fmt))? I know it would require yet another dependency and that it requires C++14 (so it should be done far in the future), but it's small enough and pretty handy.

Also is there a policy on `throw`ing vs logging + `exit(-1)`?

dkozel:
Marcus gets the final call, but we already require boost in other parts of GNU Radio that aren't likely to be easy to remove, adding another dependency and having to refactor seems like probably not worth it. We will likely have C++14 on master very soon though.

drmpeg:
It should be noted that the file lib/atsc/atsc_viterbi_gen.cc (with all the printf's) isn't really part of GNU Radio. I eliminated it's usage with this commit.

https://github.com/gnuradio/gnuradio/commit/5af80cc3146a908f3d3fd4fb201f8eebe03eff1e

It's only in the tree for legacy purposes.

adhorden:
I would like to get started on this and I feel this is a great first issue for me to start contributing to GNU Radio 🤓 .

I have had a good grep through the code and found many of the places that need refactoring so I will get started with these in a branch. I will follow the same convention as the original pull request.

85ayush:
I would love to work on this issue, I am new to open source contribution, can you please guide me?

PR #3647, opened 2020-07-16
runtime: prefs: use GR logging for debug messages

sashko:
Part of #3362.

### DIFF
--- a/gnuradio-runtime/lib/prefs.cc
+++ b/gnuradio-runtime/lib/prefs.cc
@@ -13,6 +13,7 @@
 #endif
 
 #include <gnuradio/constants.h>
+#include <gnuradio/logger.h>
 #include <gnuradio/prefs.h>
 #include <gnuradio/sys_paths.h>
 
@@ -72,6 +73,10 @@ std::vector<std::string> prefs::_sys_prefs_filenames()
 
 void prefs::_read_files(const std::vector<std::string>& filenames)
 {
+    gr::logger_ptr logger, debug_logger;
+    gr::configure_default_loggers(logger, debug_logger, "prefs");
+    std::ostringstream msg;
+
     for (const auto& fname : filenames) {
         std::ifstream infile(fname.c_str());
         if (infile.good()) {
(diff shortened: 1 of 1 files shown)


drmpeg:
It's not really useful to modify this file. It was never executed and it's no longer being built. It's just there for historical purposes. It's been replaced with a pre-built file so that gr-dtv can be cross compiled. See:

https://github.com/gnuradio/gnuradio/commit/5af80cc3146a908f3d3fd4fb201f8eebe03eff1e

sashko:
> It's not really useful to modify this file. It was never executed (during runtime) and it's no longer being built. It's just there for historical purposes. It's been replaced with a pre-built file so that gr-dtv can be cross compiled. See:
> 
> [5af80cc](https://github.com/gnuradio/gnuradio/commit/5af80cc3146a908f3d3fd4fb201f8eebe03eff1e)

Understood. Let's try with something else then...

gnuradio-runtime/lib/prefs.cc
```diff
@@ -72,6 +73,10 @@ std::vector<std::string> prefs::_sys_prefs_filenames()
 
 void prefs::_read_files(const std::vector<std::string>& filenames)
 {
+    gr::logger_ptr logger, debug_logger;
+    gr::configure_default_loggers(logger, debug_logger, "prefs");
+    std::ostringstream msg;
```
willcode, line 78: msg never gets logged.

gnuradio-runtime/lib/prefs.cc
```diff
@@ -98,14 +103,14 @@ void prefs::_read_files(const std::vector<std::string>& filenames)
...
+                msg << "WARNING: Config file '" << fname
+                    << "' failed to parse:" << std::endl;
+                msg << e.what() << std::endl;
+                msg << "Skipping it" << std::endl;
             }
         } else { // infile.good();
-            std::cerr << "WARNING: Config file '" << fname
-                      << "' could not be opened for reading." << std::endl;
+            msg << "WARNING: Config file '" << fname
+                << "' could not be opened for reading." << std::endl;
```
marcusmueller, line 113: This is in the right spirit, but `msg` will have to end up in the logger, too!

marcusmueller (changes requested):

mbr0wn:
@sashko We would still like to merge this. It's a small enough change, and a good one. However, the comments need to be resolved (i.e., `msg` needs to be logged), and can you please sign-off your commit (`git commit -s`). We'll close this if it doesn't get resolved, to reduce clutter in our PR queue.

mbr0wn:
Closing due to lack of action (we need to de-clutter the PR queue). Please reopen if you can update the PR. Thanks for your contribution!

PR #3649, opened 2020-07-17
Fix bug with non seekable file source

esqt:
Reaching the end of a non seekable file (such as named pipes) currently throws and exception.
This patch checks the EOF flag if the read operation doesn't return the expected number of bytes.

### DIFF
--- a/gr-blocks/lib/file_source_impl.cc
+++ b/gr-blocks/lib/file_source_impl.cc
@@ -265,7 +265,8 @@ int file_source_impl::work(int noutput_items,
             // EOF reached for non seekable file, return
             if (!d_seekable && feof((FILE*)d_fp))
                 return nitems_read == 0 ? WORK_DONE : nitems_read;
-            // For seekable files, the bounds of the file are known, so unexpected nitems is an error
+            // For seekable files, the bounds of the file are known, so unexpected nitems
+            // is an error
             throw std::runtime_error("fread error");
         }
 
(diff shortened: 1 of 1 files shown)


gr-blocks/lib/file_source_impl.cc
```diff
@@ -260,13 +260,19 @@ int file_source_impl::work(int noutput_items,
 
         uint64_t nitems_to_read = std::min(size, d_items_remaining);
 
-        // Since the bounds of the file are known, unexpected nitems is an error
-        if (nitems_to_read != fread(o, d_itemsize, nitems_to_read, (FILE*)d_fp))
+        size_t nitems_read = fread(o, d_itemsize, nitems_to_read, (FILE*)d_fp);
+        if (nitems_to_read != nitems_read) {
+            // EOF reached for non seekable file, return
+            if (!d_seekable && feof((FILE*)d_fp))
```
marcusmueller, line 266: while this is nice and compact, we typically use { } around conditional blocks; small issue.

nickoe:
@esqt Are you going to followup on the feedback? Also, any progress on the CLA?

jacobagilbert:
@esqt if the CLA is an issue, its no longer required. See #3992 but all you need to do is read the DCO and sign your commits to acknowledge it. I'm actually not sure if GR is accepting new CLAs or not; the DCO is a much simpler process.

mbr0wn:
@esqt we would really like this fix. I'll still be closing it for now, due to lack of activity. We'll require at least a rebase, a DCO signoff, and some responses to the review comments. Thanks for making such useful contributions!

willcode:
Submitted new PR https://github.com/gnuradio/gnuradio/pull/4094 based on this one, with a couple changes.

PR #3650, opened 2020-07-17
PROPOSAL: grc: Make it possible to copy error message to buffer

sashko:
Hey folks.

So, I have this proposal: error messages displayed in the `Errors and Warnings` dialog are quite long sometimes. As I user I would like to be able to copy it to buffer, so I can web search it or post somewhere, e.g. forum, GitHub issue, IRC, etc.

I have made a quick proof of concept where the error message is copied to the buffer on right mouse click. The UI mechanism can be reworked, e.g. we can have a menu with the 'Copy' item showing up or whatever.

Please let me know what you think of this idea. Thank you.

### DIFF
--- a/grc/gui/Dialogs.py
+++ b/grc/gui/Dialogs.py
@@ -10,7 +10,7 @@ import sys
 import textwrap
 from distutils.spawn import find_executable
 
-from gi.repository import Gtk, GLib
+from gi.repository import Gtk, GLib, Gdk
 
 from . import Utils, Actions, Constants
 from ..core import Messages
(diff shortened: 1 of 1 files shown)


sashko:
My CLA is finally approved.

mormj:
@sashko - I tried this out and it does work in such a way to allow copy and paste of the errors - so that is a great feature to have.  The right click does not seem obvious as a mechanism to get the data out, and it would be nice to have all the information (i.e. block ids, aspect, and message) formatted to a text string.

It also would be nice as you mentioned above to have the feature more integrated - e.g. a copy button, or context menu option.  Love the idea of being able to copy to clipboard this information, though.

mbr0wn:
I love the idea too. @mormj would be OK merging this and using your comment as a feature suggestion for further work?

mbr0wn (approved):

mormj:
That works for me!

Issue #3651, opened 2020-07-17
NBFM_Transmit incompatible constructor arguments


duggabe:
NBFM_test.grc on https://pastebin.com/86T2yakk
With Ubuntu 18.04 and GRC 3.8.1.0 all works as expected.
With Ubuntu 20.04 and GRC 3.9.0.0-git I get the following error:
```
Executing: /usr/bin/python3 -u /home/barry/GRdev/NBFM_test.py

Traceback (most recent call last):
  File "/home/barry/GRdev/NBFM_test.py", line 237, in <module>
    main()
  File "/home/barry/GRdev/NBFM_test.py", line 213, in main
    tb = top_block_cls()
  File "/home/barry/GRdev/NBFM_test.py", line 159, in __init__
    self.analog_nbfm_tx_0 = analog.nbfm_tx(
  File "/usr/local/lib/python3/dist-packages/gnuradio/analog/nbfm_tx.py", line 64, in __init__
    self.interpolator = filter.interp_fir_filter_fff (interp_factor, interp_taps)
TypeError: __init__(): incompatible constructor arguments. The following argument types are supported:
    1. gnuradio.filter.filter_python.interp_fir_filter_fff(interpolation: int, taps: List[float])

Invoked with: 4.0, [0.017187240899841555, -0.0021026244256288347, -0.002144490189440985,
 ...
-0.0022832044682693613, -0.002144490189440985, -0.0021026244256288347, 0.017187240899841555]

>>> Done (return code 1)
```
At line 55 of https://github.com/gnuradio/gnuradio/blob/master/gr-analog/python/analog/nbfm_tx.py
<code>interp_factor = quad_rate / audio_rate</code>
produced '4.0' instead of '4' even though both factors were integers.

PR #3652, opened 2020-07-17
cast integer divide to type int

duggabe:
resolves https://github.com/gnuradio/gnuradio/issues/3651

### DIFF
--- a/gr-analog/python/analog/nbfm_tx.py
+++ b/gr-analog/python/analog/nbfm_tx.py
@@ -52,7 +52,7 @@ class nbfm_tx(gr.hier_block2):
         do_interp = audio_rate != quad_rate
 
         if do_interp:
-            interp_factor = quad_rate / audio_rate
+            interp_factor = int(quad_rate / audio_rate)           # force integer
             interp_taps = filter.optfir.low_pass(interp_factor,   # gain
                                                  quad_rate,       # Fs
                                                  4500,            # passband cutoff


gr-analog/python/analog/nbfm_tx.py
```diff
@@ -52,7 +52,7 @@ def __init__(self, audio_rate, quad_rate, tau=75e-6, max_dev=5e3, fh=-1.0):
         do_interp = audio_rate != quad_rate
 
         if do_interp:
-            interp_factor = quad_rate / audio_rate
+            interp_factor = int(quad_rate / audio_rate)           # force integer
```
daniestevez, line 55: I'm not sure this is the correct solution to the problem, because if the quotient `quad_rate / audio_rate` is not an integer, then the block will appear to be working, but its output will be wrong because the interpolation factor will be wrong.

Still, something needs to be done, because in its present state this doesn't work at all, since `interp_factor` is always a `float` in Python3, so the interpolating FIR filter will complain.

Probably the problem dates back to GNU Radio 3.7 and Python 2.7, where the division of integers had the same semantics as the solution suggested by @duggabe here. But I guess it had the same problem I had mentioned above.

So I think we can take this as an opportunity for a better solution. Two options come to mind:
1. Check if the quotient is an integer and throw an error otherwise
2. Use a Polyphase arbitrary resampler to allow a non-integer `interp_factor`

I would prefer 1., because it is much simpler and doesn't change significantly the current code.

duggabe: If you look at lines 48 and 49, that check is already done, so it's just a matter of giving a warning or quitting.

I just checked with GRC 3.9.0.0-git trying to set a non-integral value and got
`Assertion "(quad_rate)%(audio_rate) == 0" failed.` 
so the user can not even execute the flowgraph with a bad value.

So all that's left is to do the type casting.

daniestevez: My bad, sorry. Somehow I didn't see those lines before. Then I think your solution is correct.

mormj:
The CI failure looks unrelated, but re-running that job just to get the green light before merging.

mormj:
@duggabe - Before merging, would you mind updating your commit message to identify the module and the block for future scraping of the commit log.  In this case it could look like:
`analog: cast integer divide to type int for nbfm_tx`
Thanks!

PR #3654, opened 2020-07-19
controlport: include <algorithm> when using std::for_each()

daniestevez:
I'm building master in my Gentoo system and I need to `#include <algorithm>` whenever `std::for_each()` is used, since otherwise it won't build. I believe that this is the correct way to fix this problem, but since I'm not so savvy about C++, please correct me if I'm not right.


daniestevez:
Sorry for the force-push. I've just done it to make `clang-lint` happy.

mormj:
LGTM :+1:

PR #3655, opened 2020-07-19
gr-qtgui: use boost placeholders explicitly in eye sink

daniestevez:
I'm building master in my Gentoo system. I've found that if [Boost placeholders](https://www.boost.org/doc/libs/1_69_0/libs/hof/doc/html/include/boost/hof/placeholders.html) are used, I need to `#include <boost/bind/placeholders.hpp>` and use the placeholders from namespace `boost::placeholders`. Otherwise this won't build.

I'm using Boost 1.73 and gcc 10.1.0.

It think this is the correct way to fix this problem, but as I'm not C++-savvy, please correct me if I'm wrong.

### DIFF
--- a/gr-qtgui/lib/eye_sink_c_impl.cc
+++ b/gr-qtgui/lib/eye_sink_c_impl.cc
@@ -24,6 +24,9 @@
 #include <string.h>
 #include <algorithm>
 
+#include <boost/bind/placeholders.hpp>
+using namespace boost::placeholders;
+
 namespace gr {
 namespace qtgui {
 
--- a/gr-qtgui/lib/eye_sink_f_impl.cc
+++ b/gr-qtgui/lib/eye_sink_f_impl.cc
@@ -24,6 +24,9 @@
 
 #include <string.h>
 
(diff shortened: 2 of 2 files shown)


mormj:
That will make the issue go away for eye sink, but I think the direction we are moving is replacing all the boost binding with lambda functions:

See #3566 and #3441 

It looks like there were some modules that didn't make it into #3441 due to order of merging stuff in from the pybind branch, so all the places where there is boost::bind should be changed to lambda functions

mormj:
Spoke too soon there - looks like it is just eye sink that has not been converted to lambda function.

daniestevez:
Good suggestion. This is probably left to someone that understands `boost::bind` and C++ lambda functions better than me.

Shall we leave this PR open as a reminder, or rather open up an issue to change the eye sink to use lambda functions and close the PR?

mormj:
We can just leave this PR open - I can do the PR for eye sink shortly and then we'll close this one.  Thanks @daniestevez!

mormj:
Closing in favor as #3660

PR #3656, opened 2020-07-19
grc: Add dir_select parameter type

Yamakaja:
This commit adds a new parameter type to GNURadio Companion that allows
you to select directories using a gui. This functionality is
very similar to that of the previously available "file_open" and
"file_save" parameter types, and mainly differs by calling GTKs file
chooser with FileChooserAction.SELECT_FOLDER.

Example:
image

### DIFF
--- a/grc/gui/ParamWidgets.py
+++ b/grc/gui/ParamWidgets.py
@@ -334,3 +334,41 @@ class FileParam(EntryParam):
             self._editing_callback()
             self._apply_change()
         file_dialog.destroy()  # destroy the dialog
+
+class DirectoryParam(FileParam):
+    """Provide an entry box for a directory and a button to browse for it."""
+
+    def _handle_clicked(self, widget=None):
+        """
+        Open the directory selector, when the button is clicked.
+        On success, update the entry.
+        """
+        dirname = self.param.get_evaluated() if self.param.is_valid() else ''
+        
+        if not os.path.isdir(dirname): # Check if directory exists, if not fall back to workdir
+            dirname = os.getcwd()
+
+        if self.param.dtype == "dir_select": # Setup directory selection dialog, and fail for unexpected dtype
+            dir_dialog = Gtk.FileChooserDialog(
+                title = 'Select a Directory...', action = Gtk.FileChooserAction.SELECT_FOLDER,
+                transient_for = self._transient_for
+            )
+        else:
+            raise ValueError("Can't open directory chooser dialog for type " + repr(self.param.dtype))
+
+        # Set dialog properties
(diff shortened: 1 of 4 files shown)


grc/gui/ParamWidgets.py
```diff
@@ -334,3 +334,47 @@ def _handle_clicked(self, widget=None):
...
         file_dialog.destroy()  # destroy the dialog
+
+class DirectoryParam(EntryParam):
+    """Provide an entry box for a directory and a button to browse for it."""
+
+    def __init__(self, *args, **kwargs):
+        EntryParam.__init__(self, *args, **kwargs)
+        self._open_button = Gtk.Button(label='...')
+        self._open_button.connect('clicked', self._handle_clicked)
+        self.pack_start(self._open_button, False, False, 0)
```
skoslowski, line 345: How do you feel about deriving from FileParam to avoid duplication the `__init__` here

grc/gui/ParamWidgets.py
```diff
@@ -334,3 +334,47 @@ def _handle_clicked(self, widget=None):
...
+
+    def _handle_clicked(self, widget=None):
+        """
+        Open the directory selector, when the button is clicked.
+        On success, update the entry.
+        """
+        dirname = self.param.is_valid() and self.param.get_evaluated() or '' # Obtain current value with fallback
+        
+        if not os.path.exists(dirname): # Check if directory exists, if not fall back to workdir
```
skoslowski, line 354: may I suggest to use `os.path.isdir`. I am not quite sure if that "sees" through symlinks. So that should be tested.

grc/gui/ParamWidgets.py
```diff
@@ -334,3 +334,47 @@ def _handle_clicked(self, widget=None):
...
+        self._open_button = Gtk.Button(label='...')
+        self._open_button.connect('clicked', self._handle_clicked)
+        self.pack_start(self._open_button, False, False, 0)
+
+    def _handle_clicked(self, widget=None):
+        """
+        Open the directory selector, when the button is clicked.
+        On success, update the entry.
+        """
+        dirname = self.param.is_valid() and self.param.get_evaluated() or '' # Obtain current value with fallback
```
skoslowski, line 352: please replace this with a condition expression (aka inline if)

grc/gui/ParamWidgets.py
Yamakaja: According to the documentation [os.path.isdir](https://docs.python.org/3.8/library/os.path.html#os.path.isdir) symlinks are handled. Thanks for the suggestion!

skoslowski:
Thanks again. Since this is not used in any block in GR itself, consider posting a link to an example here once you make use of it

Issue #3637, opened 2020-07-11
Upstreaming some blocks from gr-satellites


daniestevez:
@bistromath has [suggested over on Twitter](https://twitter.com/bistromath/status/1281941914011435011) that it might be good to upstream some of the blocks used in [gr-satellites](https://github.com/daniestevez/gr-satellites). I think he's right. I try to do things with upstream blocks as much as possible, but there are some blocks I've needed to write and which might be useful in a more general context. I'm opening this issue mainly as a self-reminder, and also for additional discussion with the community about the best way to proceed.

This is a rough list of the blocks that off the top of my head I think it could be interesting to upstream.

### HDLC framer/deframer

Upstream already has their own blocks. The difference between the upstream and gr-satellites blocks is that the gr-satellites blocks are done in Python, the framer allows to specify the length of the preamble and postamble (upstream uses only 1 repetition of the `0x7e` flag), the deframer allows to skip the CRC-16 check (useful for debugging), and the framer outputs PDUs instead of bytes (might be interesting for burst transmissions).

I think that C++ is the way to go, for performance, so rather than upstreaming the gr-satellites blocks, it would be good to incorporate the additional options into the upstream blocks.

The point I'm not sure how to go about is the difference between bytes output and PDU output for the framer. Maybe it's best to support both options?

### NRZI

HDLC often goes with NRZI, which is some kind of differential encoding plus an inversion. This can be achieved with upstream blocks (Differential encoder and Not), but it is nice to have a more compact solution, as gr-satellite does (one NRZI encoder block, one NRZI decoder block) These would be easy to upstream. They're now written in C++ and have some Python QA code.

### Reed-Solomon

The "most general" Reed-Solomon block in gr-satellites is the "Reed Solomon Decoder (libfec)" block. This is basically a wrapper for the `init_rs_char()` and `decode_rs_char()` functions in Phil Karn's libfec. I'm not sure how this would fit into FECAPI. Maybe it's best to do FECAPI blocks for Reed-Solomon (which I think don't exist), since I'm quite happy using FECAPI for Viterbi decoding.

### Matrix interleaving/deinterleaving

The prototypical algorithm where data is written onto a matrix by columns and read by rows. I don't think it's possible to do this easily with upstream blocks. gr-satellites has a fairly general block for matrix deinterleaving (it allows to define the matrix size and also to crop the output). However, it is only implemented for floats. This should be implemented as a template to allow all the data types.

### Generating PDUs by finding a syncword

[...]

PR #3657, opened 2020-07-19
gr-fec: build and export (en|de)code_rs_8() and (en|de)code_rs_ccsds()

daniestevez:
These functions implement the CCSDS Reed-Solomon code in the conventional
and dual basis respectively. The code for them was already included in
gr-fec, but they weren't built.

While these functions are currently not used in GNU Radio, they might
be used by other OOT modules, as discussed in #3637. The behaviour of
the functions (en|de)code_rs_8() could actually be achieved with the
currently available (en|de)code_rs_char(), by passing the correct
parameters to init_rs_char(), but since the CCSDS code is so popular,
having the specific functions is handy. On the other hand, it is not
possible to obtain (en|de)code_rs_ccsds() by using (en|de)code_rs_char(),
since there is an additional change of basis involved (see decode_rs_ccsds.c).

Some reorganization of the code has been needed. The different encode and
decode functions use some kind of templating using the C preprocessor. The
implmementation uses several symbols that should be defined by the preprocessor
before including the implementation. The implementation can be included several
times to instance the different functions.

The code in (en|de)code_rs.c has been moved to (en|de)code_rs.h, since now it
is included and compiled twice instead of being compiled directly. The files
char.c and ccsds.c include the appropriate preprocessor definitions from char.h
and fixed.h respectively and these are what is built.


drmpeg:
I've tested this for backwards compatibility with gr-dtv. No issues found.

daniestevez:
While doing some related work I [have learned](https://github.com/conda-forge/staged-recipes/pull/12274#issuecomment-671983018) that MSVC requires a workaround to build the Reed-Solomon decoder. Inadvertently I had broken this workaround, so I've pushed a new commit to this PR to fix it.

PR #3658, opened 2020-07-19
gr-qtgui: Repair qtgui examples for pybind11.

drmpeg:

PR #3660, opened 2020-07-20
qtgui: eye_sink - add lambda functions for message handling and other pybind11 related fixes

mormj:
1) convert boost::bind to lambda functions
2) ~~add missing method on python bindings~~
3) ~~make function needs extra None~~

see #3655


drmpeg:
Just FYI, I made the exact same changes as your second commit in my PR https://github.com/gnuradio/gnuradio/pull/3658.

mormj:
Ah - I should have caught that.  I'll pull back the second commit on this PR and get yours merged.

PR #3668, opened 2020-07-24
uhd: Change Unknown PPS behavior

jdemel:
Instead of setting USRP time to 0 with `time_unknown_pps`, set it to current
time. This does not help with fine sync but it helps to stay in the same
ball park with all clocks.
One might accidently send timed commands with PC clock time. A USRP
would try to wait 50 years at this point to actually execute that
command or to send that burst if the time given in case of "Unknown
PPS" is 0. This may cause issues during shutdown because it may force a user
to send a kill signal to the flowgraph. In turn that may require a USRP
reboot.

### DIFF
--- a/gr-uhd/grc/gen_uhd_usrp_blocks.py
+++ b/gr-uhd/grc/gen_uhd_usrp_blocks.py
@@ -300,7 +300,7 @@ cpp_templates:
       ${'%'} endif
       this->${'$'}{id}->set_samp_rate(${'$'}{samp_rate});
       ${'%'} if sync == 'sync':
-      this->${'$'}{id}->set_time_unknown_pps(::uhd::time_spec_t());
+      this->${'$'}{id}->set_time_unknown_pps(::uhd::time_spec_t(time(NULL)));
       ${'%'} elif sync == 'pc_clock':
       this->${'$'}{id}->set_time_now(::uhd::time_spec_t(time(NULL)), ::uhd::usrp::multi_usrp::ALL_MBOARDS);
       ${'%'} else:
--- a/gr-uhd/grc/gen_uhd_usrp_blocks.py
+++ b/gr-uhd/grc/gen_uhd_usrp_blocks.py
@@ -169,7 +169,7 @@ templates:
         self.${'$'}{id}.set_samp_rate(${'$'}{samp_rate})
 
         ${'%'} if sync == 'sync':
-        self.${'$'}{id}.set_time_unknown_pps(uhd.time_spec(0))
+        self.${'$'}{id}.set_time_unknown_pps(uhd.time_spec(time.time()))
         ${'%'} elif sync == 'pc_clock':
         self.${'$'}{id}.set_time_now(uhd.time_spec(time.time()), uhd.ALL_MBOARDS)
         ${'%'} else:


dkozel:
This is a significant behavioral change to the behavior of the setting. I have a fair number of flowgraphs that assume that the time is set to 0 on the PPS edge that this will break. I'm not against the change, but @marcusmueller this should be highlighted in the changelog/release announcement.

mbr0wn:
> This is a significant behavioral change to the behavior of the setting. I have a fair number of flowgraphs that assume that the time is set to 0 on the PPS edge that this will break. I'm not against the change, but @marcusmueller this should be highlighted in the changelog/release announcement.

In that case, maybe we simply should make it easier to sync the PPS to the PC clock. Which @jdemel already has a PR for.

jdemel:
This would be #3563 then.
This PR here helps to get all chains in sync. The other PR helps to get a PC host clock and a USRP roughly in sync. But you'll notice a constant offset between receiver chains. I haven't tested this for transmitter chains.
I assume we can't just use a PPS signal in every case because some USRPs lack a PPS signal. Correct?

mbr0wn:
> I assume we can't just use a PPS signal in every case because some USRPs lack a PPS signal. Correct?

All the USRPs have a PPS input, except for USRP1. That said, I'm not sure I understand the question. What's true is that if you don't use a common PPS (e.g. from an OctoClock), then you can't get all your devices to the same time.

mbr0wn:
> One might accidently send timed commands with PC clock time.

I guess one might always accidentally send timed commands that are far in the future...?

I'm trying to get some opinions from practitioners here. Looks like Johannes pro, Derek con -- right now I would say le'ts keep the status quo then.

mbr0wn:
Actually, after thinking about this again: The desired behaviour (desired by @jdemel, that is) is exactly the "sync to PC clock" behaviour, which we already support. However, we could decide to change the sync-to-pc-clock behaviour to use set_time_unknown_pps() instead of set_time_now(). For some USRPs and UHD versions, that does make a difference -- like making sure the radios on the same device are actually in sync.

If devices are not being sourced by a shared PPS, then using set_time_unknown_pps() is anyway not useful for multiple devices.

@jdemel what do you think? Shall we change the `pc_clock` behaviour?

jdemel:
Yes, I want "sync to PC clock" with aligned streams. (I saw a constatnt ~10us offset between streams).
I know, that most USRPs can sync to an external PPS signal. But do they also have their own internal signal? Apparently N310s do with their GPSDO even if there's no GPS signal available. In that case both daughterboards are synced with unknown PPS but have an offset with "sync to PC clock". 
Changing "sync to PC clock" behavior to `set_time_unknown_pps` will probably result in coarser sync between host and USRP. Would that be acceptable?

mbr0wn:
All USRPs have an internal PPS, but they are not aligned.

>  Changing "sync to PC clock" behavior to `set_time_unknown_pps` will probably result in coarser sync between host and USRP. Would that be acceptable?

I see your problem; if we use the internal PPS, it's going to be all over the place within a second. If we use 'set_time_now()' then it's going to be all over the place within your network and software jitter, which is going to be less than a second.
We might actually be able to speed that up if we parallelize the set-time-now calls (that would have to happen within UHD).

> Yes, I want "sync to PC clock" with aligned streams. (I saw a constatnt ~10us offset between streams).

Is this between radios on the same device? If so, which device? Or is it between radios from different devices? Because then 10µs is actually pretty good, given that you have no common time base.

mbr0wn:
FYI, I'm still strongly in favour of not merging this PR. The simple reason is: You're choosing between a default sync option, and a "sync to PC clock" option. If you are not choosing the latter, then the argument that 

> One might accidently send timed commands with PC clock time.

Becomes invalid. After all, you deliberately chose to not use the PC clock sync option.

I would like to understand more of your use case though.

jdemel:
> > Yes, I want "sync to PC clock" with aligned streams. (I saw a constatnt ~10us offset between streams).
> 
> Is this between radios on the same device? If so, which device? Or is it between radios from different devices? Because then 10µs is actually pretty good, given that you have no common time base.

This behavior can be observed between 2 daughterboards on the same device. e.g. choose one N310, set "sync to PC clock" and receive on channels 0 and 3 (assuming those are on 2 different daughterboards internally). I observed the same behavior with an X310.
I guess, it would be highly favorable to fix this issue in UHD instead of this PR here. Would this be an option?

My use case
I set "sync to PC clock" for one device to be able to send timed commands. 
I resync time between USRP and host every couple of seconds to ensure they don't deviate over time.
Now, I can use a lazy approach to send timed commands by just using a host timestamp. Even values like 100us in the future will be transmitted (i.e. no late packet drop) because the delay over the network affected the "set_time" command just like the burst, I'd like to transmit.

A different approach would be to monitor the RX stream and pass this timing information to the TX stream to calculate the TX timestamp. With that approach, I'd periodically emit a message with the current RX time and use this info in my TX stream. If I emit a new message every 100us, I basically just introduced another 100us delay. If I decrease the interval between 2 messages, I put way more work on the messaging system.

TX and RX timestamps use different formats (different pmt types) by the way.

mbr0wn:
> This behavior can be observed between 2 daughterboards on the same device. e.g. choose one N310, set "sync to PC clock" and receive on channels 0 and 3 (assuming those are on 2 different daughterboards internally). I observed the same behavior with an X310.

OK after some head-scratchin', I think I have all the puzzle pieces:
- None of this applies for UHD 4.0 (hitherto unreleased, but still). There is only one timekeeper for all radios in UHD 4.
- In UHD 3, X310 actually syncs the clocks on init, but it seems like N310 does not.
- There's a set_time_sync() API call, but it's not public on UHD 3. It's only on radio_ctrl_impl.

That means that X310 has all the pieces to do what you want, but doesn't expose the API. N310 doesn't have the sync_now API enabled.

Merging this PR won't solve your problem, because, as discussed, PPS signals don't bear any relation to one another, so we can safely close it. It would help for a single N310, but then you could also modify the sync-to-PC-clock option to not disturb people who assume clocks start at 0.

Side note: You could also read back the time from the device, and remember the offset to TOD in your software. If the N310 happens to bring up all clocks in sync, then you'd be fine (but I haven't verified that).

Something like this (pseudocode):
```
time_offset = time.time() - usrp->get_time_now();
// stuff happens
usrp->set_center_freq(new_freq, time.time() - time_offset);
```

jdemel:
Thanks for discussing this PR. I agree this PR is a workaround and I look forward to use UHD4. Is there a timeline for a UHD4 release?

PR #3674, opened 2020-07-29
cmake: Add -undefined dynamic_lookup linker flag back for macOS

### LINKED ISSUE #3279: Add -undefined dynamic_lookup linker flag back for macOS
Hi

Homebrew maintainer here.
We are trying to update gnuradio to version 3.8.0.0.
https://github.com/Homebrew/homebrew-core/pull/49733

It looks like the python `-undefined dynamic_lookup` linker flag has been removed in the last release. See https://github.com/gnuradio/gnuradio/commit/ab2fb35677e38a384df3f9503d1f45f64bbc0374#diff-45856d0ec64e3d28b38f044024648892L163

Could this be added back?

SeekingMeaning:
Closes #3279

### DIFF
--- a/cmake/Modules/GrPython.cmake
+++ b/cmake/Modules/GrPython.cmake
@@ -37,9 +37,14 @@ set(PYTHON_EXECUTABLE ${PYTHON_EXECUTABLE} CACHE FILEPATH "python interpreter")
 set(QA_PYTHON_EXECUTABLE ${QA_PYTHON_EXECUTABLE} CACHE FILEPATH "python interpreter for QA tests")
 
 add_library(Python::Python INTERFACE IMPORTED)
+if(APPLE)
+    set_target_properties(Python::Python PROPERTIES
+      INTERFACE_INCLUDE_DIRECTORIES "${PYTHON_INCLUDE_DIRS}"
+      INTERFACE_LINK_LIBRARIES "-undefined dynamic_lookup"
+      )
 # Need to handle special cases where both debug and release
 # libraries are available (in form of debug;A;optimized;B) in PYTHON_LIBRARIES
-if(PYTHON_LIBRARY_DEBUG AND PYTHON_LIBRARY_RELEASE)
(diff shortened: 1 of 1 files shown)


cmake/Modules/GrPython.cmake
```diff
@@ -37,9 +37,14 @@ set(PYTHON_EXECUTABLE ${PYTHON_EXECUTABLE} CACHE FILEPATH "python interpreter")
...
 add_library(Python::Python INTERFACE IMPORTED)
+if(APPLE)
+    set_target_properties(Python::Python PROPERTIES
+      INTERFACE_INCLUDE_DIRECTORIES "${PYTHON_INCLUDE_DIRS}"
+      INTERFACE_LINK_LIBRARIES "-undefined dynamic_lookup"
+      )
 # Need to handle special cases where both debug and release
 # libraries are available (in form of debug;A;optimized;B) in PYTHON_LIBRARIES
-if(PYTHON_LIBRARY_DEBUG AND PYTHON_LIBRARY_RELEASE)
+elseif(PYTHON_LIBRARY_DEBUG AND PYTHON_LIBRARY_RELEASE)
```
mbr0wn, line 47: Is it not possible to be both on Mac, and also have PYTHON_LIBRARY_DEBUG?

michaelld: yes this is possible on macOS ... but the linkage depends on how Python ABI is provided. If in a framework then we link against the master framework & one can select debug or not via special environment variables. If as libraries, then we link against the desired library ... same as other Linux & others. One can always link against the object file, of course, but with frameworks there's a more robust alternative.

michaelld: This change is, really, about as good as it gets for macOS with respect to allowing future Python linkage to work with current ABI, because of how the macOS DYLD works. We can fixup the debug + release case later if desired.

michaelld:
Approval from macOS dev. All CI passes. Desirable reversion change. Merging.

PR #3682, opened 2020-08-05
Fix to d_data_reg in 'correlate access code tag stream' for bb and ff

mcquiggi:
Currently the variable d_data_reg is used to compare a received bit stream with a known access code.  This variable is not re-initialized between received packets. 

When the state machine finishes a packet and enters the 'enter_search' state, a poorly-chosen cyclical access code (simplest example: '11111111') will cause false detection of a new, correct access code before the requisite number of valid bits have been read.  

In my example of '11111111' (8 bits), after a valid packet when the state machine starts a new search, the first 1 of the next packet will cause a false detection of a new packet As only one bit will have been read rather than the requisite 8 bits.  

It is better to re-initialize d_data_reg when starting a new search in order to prevent this false detection.

### DIFF
--- a/gr-digital/lib/correlate_access_code_bb_ts_impl.cc
+++ b/gr-digital/lib/correlate_access_code_bb_ts_impl.cc
@@ -90,6 +90,8 @@ unsigned long long correlate_access_code_bb_ts_impl::access_code() const
 inline void correlate_access_code_bb_ts_impl::enter_search()
 {
     d_state = STATE_SYNC_SEARCH;
+    d_data_reg = 0;                 // K. McQuiggin - zero between access code searches to eliminate false detects due to poorly-chosen
+                                    // access codes.  2020.07.25
 }
 
 inline void correlate_access_code_bb_ts_impl::enter_have_sync()
--- a/gr-digital/lib/correlate_access_code_bb_ts_impl.cc
+++ b/gr-digital/lib/correlate_access_code_bb_ts_impl.cc
@@ -90,8 +90,9 @@ unsigned long long correlate_access_code_bb_ts_impl::access_code() const
 inline void correlate_access_code_bb_ts_impl::enter_search()
 {
     d_state = STATE_SYNC_SEARCH;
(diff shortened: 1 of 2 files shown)


gr-digital/lib/correlate_access_code_bb_ts_impl.cc
```diff
@@ -90,6 +90,8 @@ unsigned long long correlate_access_code_bb_ts_impl::access_code() const
 inline void correlate_access_code_bb_ts_impl::enter_search()
 {
     d_state = STATE_SYNC_SEARCH;
+    d_data_reg = 0; // K. McQuiggin - zero between access code searches to eliminate false
```
ThomasHabets, line 93: The code speaks for itself. Doesn't need signatures.

Resetting data reg also doesn't seem like it should need a comment at all.

ThomasHabets:
Seems to me that a better fix may be to make sure we've processed at least the number of bits while in `STATE_SYNC_SEARCH`, after which the masking will take care of the problem automatically.

What do you think?

mcquiggi:
Hi Thomas:

Okay.  It used to be a “best practise” for those changing code to leave a record of who/when it was changed, with a short description of the change.  Remove comments as you wish in accordance with project standards.

Thanks for reviewing the changes!

Kevin

mcquiggi:
That would work but its another thing to keep track of, why not just zero the accumulator at start of search?  Simpler and less overhead.

ThomasHabets:
The downside of not keeping track of symbols count is if the access code has leading zeroes, you may get more false positives.

ThomasHabets:
For keeping track of changes nowadays that's delegated to version control. `git blame` is a better tool than source comments to find when and why something was changed.

ThomasHabets:
Alternative PR: https://github.com/gnuradio/gnuradio/pull/3692

willcode:
Go with #3692

mormj:
closing in favor of #3692

PR #3694, opened 2020-08-13
fft: window: Allow normalizing windows

mbr0wn:
```
commit f960615febe49a5cc68d3359c090a8c8b0068de1 (HEAD -> norm-win, mb/norm-win)
Author: Martin Braun <martin.braun@ettus.com>
Date:   Thu Aug 13 17:06:36 2020 +0200

    fft: window: Allow normalizing windows
    
    In some applications, it is useful to generate windows that have unit
    power. The boxcar window (rectangle) is always of unit power, but the
    other windows are not, leading to apple-to-oranges comparisons, e.g.,
    when switching between window types in a live spectrum estimation
    application.

commit 56d4ad1d69eafe299880b4aad407f5717c5933d1
Author: Martin Braun <martin.braun@ettus.com>
Date:   Thu Aug 13 15:08:08 2020 +0200

    fft: window: Provide default value for beta param on window::build
    
    The API call fft::window::build() takes a third parameter, beta, that
    only applies for the Kaiser window. For other windows, it is necessary
    to provide a dummy value to call this function.
    
    The declared-deprecated version firdes::window() on the other hand does
    not require specifying beta when not needed. Instead, we provide
    a default value of 6.76 which will still generate a valid (and generally
    useful) Kaiser window, but means we don't have to specify beta for other
    windows.
```

### DIFF
--- a/gr-fft/include/gnuradio/fft/window.h
+++ b/gr-fft/include/gnuradio/fft/window.h
@@ -332,7 +332,7 @@ public:
      * \param ntaps Number of coefficients in the window.
      * \param beta Used only for building Kaiser windows.
      */
-    static std::vector<float> build(win_type type, int ntaps, double beta);
+    static std::vector<float> build(win_type type, int ntaps, double beta = 6.76);
 };
 
 } /* namespace fft */
--- a/gr-fft/include/gnuradio/fft/window.h
+++ b/gr-fft/include/gnuradio/fft/window.h
@@ -331,8 +331,10 @@ public:
      * \param type a gr::fft::win_type index for the type of window.
      * \param ntaps Number of coefficients in the window.
      * \param beta Used only for building Kaiser windows.
+     * \param normalize If true, return a window with unit power
      */
-    static std::vector<float> build(win_type type, int ntaps, double beta = 6.76);
+    static std::vector<float>
+    build(win_type type, int ntaps, double beta = 6.76, const bool normalize = false);
 };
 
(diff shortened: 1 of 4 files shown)


gr-fft/include/gnuradio/fft/window.h
```diff
@@ -330,8 +330,10 @@ class FFT_API window
      * \param type a gr::fft::win_type index for the type of window.
      * \param ntaps Number of coefficients in the window.
      * \param beta Used only for building Kaiser windows.
+     * \param normalize If true, return a window with unit power
      */
-    static std::vector<float> build(win_type type, int ntaps, double beta);
+    static std::vector<float>
+    build(win_type type, int ntaps, double beta = 6.76, const bool normalize = false);
```
mormj, line 336: Does this become awkward for a window that doesn't require a beta value, but user wants it normalized?  Should there be an overloaded build that has normalize as the third parameter.

mbr0wn: That's a good suggestion. But note that in Python, you're already golden because of the combo of PyBind11 and kwargs. This is legit:
```python
win = fft.window.build(fft.win_type.WIN_BLACKMAN_hARRIS, 21, normalize=True)
```

mbr0wn: I would say, you're maybe right,  but we can add a second `build()` function if someone wants it in a different PR.

PR #3695, opened 2020-08-13
fft: window: Add WIN_NONE

mbr0wn:
fft::window is slated to replace firdes::window, we add the missing
WIN_NONE for consistency between the two.


mormj:
Simple enum addition, going ahead and merging because #3696 depends on it

PR #3696, opened 2020-08-13
qtgui: Replace firdes::window with fft::window

mbr0wn:
The former has been declared deprecated, hence the shift.

### DIFF
--- a/gr-qtgui/lib/waterfall_sink_c_impl.cc
+++ b/gr-qtgui/lib/waterfall_sink_c_impl.cc
@@ -52,7 +52,7 @@ waterfall_sink_c_impl::waterfall_sink_c_impl(int fftsize,
       d_fftsize(fftsize),
       d_fft_shift(fftsize),
       d_fftavg(1.0),
-      d_wintype((filter::firdes::win_type)(wintype)),
+      d_wintype((fft::window::win_type)(wintype)),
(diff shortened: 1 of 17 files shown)


mbr0wn:
Note: This is 100% search-and-replace.

gr-qtgui/lib/waterfall_sink_c_impl.cc
```diff
@@ -333,8 +333,8 @@ void waterfall_sink_c_impl::windowreset()
 void waterfall_sink_c_impl::buildwindow()
 {
     d_window.clear();
-    if (d_wintype != filter::firdes::WIN_NONE) {
-        d_window = filter::firdes::window(d_wintype, d_fftsize, 6.76);
+    if (d_wintype != fft::window::WIN_NONE) {
```
mormj, line 336: shouldn't need the 6.76 magic number hardcoded here since it is the default

mbr0wn: Hah yeah! That's why I submitted 3adb522b0ddf6736605beea797016353a1d4a2b8 then forgot about it.

mbr0wn: Fixed that in all 4 places.

PR #3700, opened 2020-08-14
uhd: uhd_fft: Clean up Python, fix PyLint warnings

mbr0wn:
This in preparation for future modifications of uhd_fft (and easier
review thereof). This is mostly:
- Whitespace cleanup
- Removal of dead code
- Breaking of long lines

PyLint is much happier and can provide more useful warnings. No
functional changes.


mbr0wn:
FYI I noticed that the GUI stalls when an overrun occurs, which I'm sure it didn't always. However, this is already an issue on master, so I'd rather merge this first and continue fixing on the version without the million linter warnings.

PR #3705, opened 2020-08-17
qtgui: freq_sink: Allow normalization of window functions

mbr0wn:
When switching between window functions in a live FFT GUI session, the
difference in power of the window functions is notable. This allows
asserting the 'normalize' argument of fft::window::build() by setting a
bit in the window type during construction,, or by calling a new API
call: set_fft_window_normalized().

Note: The choice to use an unused bit to set the normalization comes
from an attempt to keep the API backward compatible. Adding a new
argument to the make call would require reordering the make arguments.

This feature allows changing window types without signal power varying
due to window power differences.

Note that this does not affect scalloping loss, or processing loss from
the window. For example, the peak value of a sine tone will still vary
for different windows, because different windows distribute the power
differently, and most window functions move power to the sidelobes.

Supersedes https://github.com/gnuradio/gnuradio/pull/3496

### DIFF
--- a/gr-qtgui/grc/qtgui_freq_sink_x.block.yml
+++ b/gr-qtgui/grc/qtgui_freq_sink_x.block.yml
@@ -26,8 +26,8 @@ parameters:
 -   id: freqhalf
     label: Spectrum Width
     dtype: enum
-    default: 'True'
-    options: ['True', 'False']
+    default: True
+    options: [True, False]
     option_labels: [Full, Half]
     hide: ${ ('part' if type == "float" or type == "msg_float" else 'all') }
 -   id: wintype
@@ -39,6 +39,13 @@ parameters:
     option_labels: [Blackman-harris, Hamming, Hann, Blackman, Rectangular, Kaiser,
         Flat-top]
     hide: part
+-   id: norm_window
+    label: Normalize Window Power
+    dtype: enum
+    default: 'False'
+    options: ['True', 'False']
+    option_labels: [Yes, No]
+    hide: ${ ('none' if norm_window == 'True' else 'part') }
 -   id: fc
     label: Center Frequency (Hz)
     dtype: real
@@ -417,6 +424,7 @@ templates:
         self.${id}.set_fft_average(${average})
         self.${id}.enable_axis_labels(${axislabels})
         self.${id}.enable_control_panel(${ctrlpanel})
+        self.${id}.set_fft_window_normalized(${norm_window})
 
         % if legend == "False":
         self.${id}.disable_legend()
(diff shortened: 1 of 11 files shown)


mbr0wn:
@rear1019 @ThomasHabets You requested changes in https://github.com/gnuradio/gnuradio/pull/3496, so if you want to double-check then here's the new version.

gr-qtgui/lib/freq_sink_c_impl.cc
```diff
@@ -20,6 +20,9 @@
 #include <qwt_symbol.h>
 #include <volk/volk.h>
 
+#include <boost/range/adaptor/transformed.hpp>
+#include <boost/range/numeric.hpp>
```
rear1019, line 24: These includes are not required as far as I can tell.

gr-qtgui/grc/qtgui_freq_sink_x.block.yml
```diff
@@ -39,6 +39,13 @@ parameters:
     option_labels: [Blackman-harris, Hamming, Hann, Blackman, Rectangular, Kaiser,
         Flat-top]
     hide: part
+-   id: norm_window
+    label: Normalize Window Power
+    dtype: enum
+    default: 'False'
+    options: ['True', 'False']
```
rear1019, line 46: The change above changes the string values (`'True'`) to type values (`True`). Is this required here as well?

gr-qtgui/lib/freq_sink_f_impl.cc
```diff
@@ -52,6 +55,7 @@ freq_sink_f_impl::freq_sink_f_impl(int fftsize,
       d_fft_shift(fftsize),
       d_fftavg(1.0),
       d_wintype((fft::window::win_type)(wintype)),
```
rear1019, line 57: A masking operation is missing (see change in `freq_sink_c_impl.cc`)

gr-qtgui/lib/freq_sink_f_impl.cc
```diff
@@ -20,6 +20,9 @@
 #include <qwt_symbol.h>
 #include <volk/volk.h>
 
+#include <boost/range/adaptor/transformed.hpp>
+#include <boost/range/numeric.hpp>
```
rear1019, line 24: These includes are not required as far as I can tell.

gr-qtgui/include/gnuradio/qtgui/freq_sink_c.h
```diff
@@ -83,7 +83,11 @@ class QTGUI_API freq_sink_c : virtual public sync_block
      * \param fftsize size of the FFT to compute and display. If using
      *        the PDU message port to plot samples, the length of
      *        each PDU must be a multiple of the FFT size.
-     * \param wintype type of window to apply (see gr::fft::window::win_type)
+     * \param wintype type of window to apply (see gr::fft::window::win_type).
+     *        By setting bit 16 to one, this block will normalize the window
```
rear1019, line 87: I wonder whether this hack is of any use. You can call `set_fft_window_normalized()` after calling the constructor (this is also what the template for GRC does).

PR #3707, opened 2020-08-18
network: build fixes

ncorgan:
* Add missing Windows resource file
* Include in component list in GnuradioConfig.cmake


mormj:
Follows method of other in-tree modules, 2 approvals, merging.

PR #3709, opened 2020-08-18
boost: increase minimum version to 1.58 for boost/endian

mormj:
PR #3369 incorporated boost/endian, which does not appear until Boost 1.58
Either more conditional includes need to be added or we bump up the min boost version (losing support for Centos7)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -45,7 +45,7 @@ SET(VERSION_PATCH 0-git)
 include(GrVersion) #setup version info
 
 # Minimum dependency versions for central dependencies:
-set(GR_BOOST_MIN_VERSION "1.53")
+set(GR_BOOST_MIN_VERSION "1.58")
 set(GR_CMAKE_MIN_VERSION "3.5.1")
 set(GR_MAKO_MIN_VERSION "0.4.2")
 set(GR_PYTHON_MIN_VERSION "3.6.5")


noc0lour:
In EPEL there is Boost 1.69, isn't that sufficient? also this is for master, right? How about only supporting Centos 8 for master?

noc0lour:
CentOs 8 seems to ship Boost 1.66

noc0lour (approved):

mormj:
2 approvals.  Merging

michaelld:
A little late for the party, but this is fine with me / for OSX users. Xcode doesn't install Boost anyway TTBOMK, so one has to use MP or Brew or Fink or install from source ... so the Boost version really doesn't matter very much.

I will note that this change means CentOS 7 users will now have to install Boost from source, since the 'yum' version is exactly 1.53.0 ... but, they need to anyway for any reasonably recent version of UHD ... so, really not a biggie here ;)

michaelld:
Might I ask for 1 "change of MO" here: If someone is tagged to give a review, give them 24 -- or maybe even better: 48 -- hours to reply! Unless the change is an emergency, in which case by all means get it pushed through. I saw this last night (US/Eastern) but had other obligations the whole evening. I'm clearly getting to it this morning in a reasonably timely manner, and I don't think it's an emergency ... so ... a little more timing flexility would be greatly appreciated. Either that or don't tag those "slow" people ;)

mormj:
Sorry I'm a little trigger happy :) Will do in the future.

noc0lour:
> A little late for the party, but this is fine with me / for OSX users. Xcode doesn't install Boost anyway TTBOMK, so one has to use MP or Brew or Fink or install from source ... so the Boost version really doesn't matter very much.
> 
> I will note that this change means CentOS 7 users will now have to install Boost from source, since the 'yum' version is exactly 1.53.0 ...

CentOS 7 users have the possibility to install from EPEL which ships Boost 1.69, no idea if that works or not for those users?

michaelld:
> CentOS 7 users have the possibility to install from EPEL which ships Boost 1.69, no idea if that works or not for those users?

Oh ... nice ... what is EPEL & how do I use it? I'm a total CentOS N00B ... trying to get UHD / Volk / GR going for a customer's query about UHD networking issues that seem very CentOS-specific. They are using CentOS 7 & can't upgrade to 8, but if I can find a way for them to install UHD 3.15 then that's an option.

mbr0wn:
I would also have considered this PR something that requires maintainer approval.

noc0lour:
> > CentOS 7 users have the possibility to install from EPEL which ships Boost 1.69, no idea if that works or not for those users?
> 
> Oh ... nice ... what is EPEL & how do I use it? I'm a total CentOS N00B ... trying to get UHD / Volk / GR going for a customer's query about UHD networking issues that seem very CentOS-specific. They are using CentOS 7 & can't upgrade to 8, but if I can find a way for them to install UHD 3.15 then that's an option.

EPEL is something like the unofficial? package archive with recent software compiled with toolchains from RHEL/CentOS so you can have recent software compiled with old compilers/base systems https://fedoraproject.org/wiki/EPEL

michaelld:
> EPEL is something like the unofficial? package archive with recent software compiled with toolchains from RHEL/CentOS so you can have recent software compiled with old compilers/base systems https://fedoraproject.org/wiki/EPEL

Gotcha. Thx very much! That should save me some headaches! TBH I don't really want to be using CentOS ... but, it's a job of mine at the moment & so I'm doing what I can to facilitate the customers' requests & queries.

PR #3710, opened 2020-08-19
grc: prevent search keystrokes from modifying flowgraph 

### LINKED ISSUE #2610: grc: block search applies hotkeys to flowgraph
If a block is selected on the flowgraph, then typing a search in the block tree that includes the letters {'e','d','b', etc} applies those hotkeys to the flowgraph

E.g. if I search for 'digital' entering the 'd' will disable the block that is selected, then 'igital' in the search bar

### LINKED ISSUE #3507: GRC: Finding Blocks broken (search creates a new flow graph)
Affected version: 3.8.1.0 (Ubuntu 20.04)

# Steps to reproduce:

1. Start gnuradio-companion
2. Create or open flow graph with some blocks
3. Select a block
4. Type "Ctrl + F" for finding blocks. The block is deselected. Type field opens. Type "s" "e" "l" to find selector block. This works as intended.
5. Select a block in the flow graph. The type field is still open and you can continue searching (you can even click in the search field to continue typing). Continue searching by typing "e". -> Works
6. Type "c" => A new flow graph is created

# Expected behaviour:
Typing "c" will just continue search for selector block

# Actual behaviour:
A new flow graph is created. There seems to be a general clash with gnuradio hotkeys. When a block is selected "B" (bypass), "D" (disable), "E" (enable), ... seem to have higher priority than the search field, too.

argilo:
Fixes #2610.
Fixes #3507.

A confluence of bugs introduced while updating GRC to Python 3 results in keystrokes being sent to the drawing area while typing in the block search box.

First, the drawing area needs to be placed in a Viewport before being added into a ScrolledWindow. Fixing this allows the drawing area to grab focus as intended when it is clicked.

Second, the drawing area needs to unselect all blocks when it loses focus so that the relevant keystroke actions can be disabled. This functionality was commented out during the transition to Python 3, and I've restored it here.

The partial workaround implemented in #2574 is no longer needed, so I removed it.

### DIFF
--- a/grc/gui/Application.py
+++ b/grc/gui/Application.py
@@ -749,7 +749,6 @@ class Application(Gtk.Application):
             main.update_pages()
 
         elif action == Actions.FIND_BLOCKS:
-            flow_graph.unselect()
             main.update_panel_visibility(main.BLOCKS, True)
             main.btwin.search_entry.show()
             main.btwin.search_entry.grab_focus()
--- a/grc/gui/DrawingArea.py
+++ b/grc/gui/DrawingArea.py
@@ -75,9 +75,9 @@ class DrawingArea(Gtk.DrawingArea):
 
         self.connect('leave-notify-event', _handle_notify_event, False)
         self.connect('enter-notify-event', _handle_notify_event, True)
-        # todo: fix
-#        self.set_flags(Gtk.CAN_FOCUS)  # self.set_can_focus(True)
-#        self.connect('focus-out-event', self._handle_focus_lost_event)
+
+        self.set_can_focus(True)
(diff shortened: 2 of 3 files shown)


mormj:
Big thumbs up here!! Tested and it definitely fixes this nasty problem that has driven me nuts!  Thank you for the fix @argilo 

Also will fix #2610 (which is the same as #3507)

argilo:
This should definitely be backported to 3.8 as well. I can open a pull request for that tonight.

argilo:
I noticed one case that still remains: if you select blocks by dragging the mouse to make a rectangular selection, then the selection is not cleared when moving to the search box. I'll take a stab at fixing that in a follow-up PR.

PR #3712, opened 2020-08-19
blocks: align output tags in selector block

### LINKED ISSUE #3569: runtime/blocks (?): tag propagation anomalies when selectively dropping data

mormj:
Aligns the output tags using nitems_written as recommended by @rear1019 in #3572 to address the situation where tags on the input are not aligned with the output due to dropped samples

@mrjacobagilbert @rear1019 - please take a look, I believe this can supercede #3572 with less bookkeeping

Fixes #3569

### DIFF
--- a/gr-blocks/lib/selector_impl.cc
+++ b/gr-blocks/lib/selector_impl.cc
@@ -48,6 +48,8 @@ selector_impl::selector_impl(size_t itemsize,
     message_port_register_in(pmt::mp("oindex"));
     set_msg_handler(pmt::mp("oindex"),
                     [this](pmt::pmt_t msg) { this->handle_msg_output_index(msg); });
+
+    set_tag_propagation_policy(TPP_CUSTOM);
 }
 
 selector_impl::~selector_impl() {}
@@ -149,8 +151,20 @@ int selector_impl::general_work(int noutput_items,
     const uint8_t** in = (const uint8_t**)&input_items[0];
     uint8_t** out = (uint8_t**)&output_items[0];
 
+
     gr::thread::scoped_lock l(d_mutex);
     if (d_enabled) {
+        auto nread = nitems_read(d_input_index);
+        auto nwritten = nitems_written(d_output_index);
+
+        std::vector<tag_t> tags;
+        get_tags_in_window(tags, d_input_index, 0, noutput_items);
+
+        for (auto tag : tags) {
+            tag.offset -= (nread - nwritten);
+            add_item_tag(d_output_index, tag);
+        }
+
         std::copy(in[d_input_index],
                   in[d_input_index] + noutput_items * d_itemsize,
                   out[d_output_index]);


gr-blocks/lib/selector_impl.cc
```diff
@@ -149,8 +151,20 @@ int selector_impl::general_work(int noutput_items,
...
     uint8_t** out = (uint8_t**)&output_items[0];
 
+
     gr::thread::scoped_lock l(d_mutex);
     if (d_enabled) {
+        auto nread = nitems_read(d_input_index);
+        auto nwritten = nitems_written(d_output_index);
+
+        std::vector<tag_t> tags;
+        get_tags_in_window(tags, 0, 0, noutput_items);
```
rear1019, line 161: This selects tags from the first input. Why not from input `d_input_index`?

gr-blocks/lib/selector_impl.cc
```diff
@@ -149,8 +151,20 @@ int selector_impl::general_work(int noutput_items,
...
+
+        for (auto tag : tags) {
+            tag.offset -= (nread - nwritten);
```
rear1019, line 164: This should be correct, however, I can not test right now.

Me suggestion from #3572 to simply use `tag.offset = nitems_written(d_output_index)` is wrong. The tag offset must be modified to `offset at input − number of dropped samples = offset at input − (number items read − number items written)` (as used in this code). Or put another way to `number items written + relative offset at input = number items written + (offset at input − number items read)`.

rear1019 (approved):

mrjacobagilbert:
@mormj I took a look over this and agree this is a better solution than #3569. i tested it using the attached FG and it looks like it behaves properly.

[tagpropagation.zip](https://github.com/gnuradio/gnuradio/files/5137817/tagpropagation.zip)

mormj:
Thanks @mrjacobagilbert and @rear1019 - merging based on review

PR #3713, opened 2020-08-19
uhd: Add power reference APIs

mbr0wn:
UHD 4.0 is introducing a power reference level API, which will now
become available in GNU Radio. Its availability is conditionally
compiled into GNU Radio based on whether or not the feature is available
in the UHD version GNU Radio was compiled against.

USRP source and sink blocks receive the following new APIs:
- has_power_reference()
- set_power_reference()
- get_power_reference()

The command interface now understands a new `power_dbm` dictionary key.
The GRC bindings can now switch between absolute gain, normalized gain,
and absolute power settings.

### DIFF
--- a/gr-uhd/include/gnuradio/uhd/usrp_block.h
+++ b/gr-uhd/include/gnuradio/uhd/usrp_block.h
@@ -20,6 +20,7 @@ namespace uhd {
 
 GR_UHD_API const pmt::pmt_t cmd_chan_key();
 GR_UHD_API const pmt::pmt_t cmd_gain_key();
+GR_UHD_API const pmt::pmt_t cmd_power_key();
 GR_UHD_API const pmt::pmt_t cmd_freq_key();
 GR_UHD_API const pmt::pmt_t cmd_lo_offset_key();
 GR_UHD_API const pmt::pmt_t cmd_tune_key();
@@ -228,6 +229,65 @@ public:
      */
     virtual ::uhd::gain_range_t get_gain_range(const std::string& name,
                                                size_t chan = 0) = 0;
+
+    /*! Query if this device is capable of absolute power levels
+     *
+     * If true, the set_power_reference() and get_power_reference() APIs can be
+     * used as well.
+     * Note that if the underlying UHD version doesn't support power APIs, a
+     * warning will be printed, and the return value is always false.
+     *
+     * \param chan the channel index 0 to N-1
+     * \returns true if there is a power reference API available for this channel
+     */
+    virtual bool has_power_reference(size_t chan = 0) = 0;
+
+    /*! Set the absolute power reference level for this channel
+     *
+     * Note that this API is available for certain devices only, and only if
+     * calibration data is available. Refer to the UHD manual for greater
+     * detail: https://files.ettus.com/manual/page_power.html
+     *
+     * In a nutshell, using the power reference will configure the device such
+     * that a full-scale signal (0 dBFS) corresponds to a signal at the
+     * antenna connector of \p power_dbm.
+     * After calling this function, the device will attempt to keep the power
+     * level constant after retuning, which means the gain level may be changed
+     * after a re-tune.
+     *
+     * The device may coerce the available power level (for example, if the
+     * requested power level is not achievable by the device). The coerced
+     * value may be read by calling get_power_reference().
+     *
+     * \param power_dbm The power reference level in dBm
+     * \param chan the channel index 0 to N-1
+     * \throws std::runtime_error if the underlying UHD version does not support
+     *         the power API.
+     */
+    virtual void set_power_reference(double power_dbm, size_t chan = 0) = 0;
+
+    /*! Return the absolute power reference level for this channel
+     *
+     * Note that this API is only available for certain devices, and assuming
+     * the existence of calibration data. Refer to the UHD manual for greater
+     * detail: https://files.ettus.com/manual/page_compat.html
+     *
+     * See also set_power_reference().
+     *
+     * \param chan the channel index 0 to N-1
+     * \throws std::runtime_error if the underlying UHD version does not support
+     *         the power API.
+     */
+    virtual double get_power_reference(size_t chan = 0) = 0;
+
+    /*! Return the available power range
+     *
+     * \param chan the channel index 0 to N-1
+     * \return the power range in dBm
+     * \throws std::runtime_error if the underlying UHD version does not support
+     *         the power API.
+     */
+    virtual ::uhd::meta_range_t get_power_range(size_t chan = 0) = 0;
 
     /*!
      * Set the antenna to use for a given channel.
--- a/gr-uhd/lib/usrp_sink_impl.cc
+++ b/gr-uhd/lib/usrp_sink_impl.cc
@@ -187,6 +187,66 @@ std::vector<std::string> usrp_sink_impl::get_gain_names(size_t chan)
     return _dev->get_tx_gain_range(name, chan);
 }
 
+bool usrp_sink_impl::has_power_reference(size_t chan)
+{
+#ifdef UHD_USRP_MULTI_USRP_POWER_LEVEL
+    if (chan >= _stream_args.channels.size()) {
+        throw std::out_of_range("Invalid channel: " + std::to_string(chan));
+    }
+    const size_t dev_chan = _stream_args.channels[chan];
+    return _dev->has_tx_power_reference(dev_chan);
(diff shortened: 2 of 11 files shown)


gr-uhd/lib/usrp_sink_impl.cc
```diff
@@ -187,6 +187,46 @@ ::uhd::gain_range_t usrp_sink_impl::get_gain_range(const std::string& name, size
...
+#endif
+}
+
+void usrp_sink_impl::set_power_reference(const double power_dbm, const size_t chan)
+{
+#ifdef UHD_USRP_MULTI_USRP_POWER_LEVEL
+    const size_t dev_chan = _stream_args.channels[chan];
+    _dev->set_tx_power_reference(power_dbm, dev_chan);
+#else
+    throw std::runtime_error("not implemented in this version");
```
michaelld, line 206: After some thought -- me trying to decide whether the `throw` is the way to go here, or a warning & still let the user's code execute -- I have settled on the former (the `throw` as you use) ... but with a more verbose message, which can be printed out before the `throw` or as part of the `throw` as you see fit: something to the effect of "The code being executed is trying to use the UHD power reference functionality, which was introduced in UHD 4.0. The actual UHD version being used is 3.15 or earlier, and does not provide this functionality. Please either update the installed UHD to version 4.0 or later, or find a way for the code being executed to conditionally use the power reference."

mbr0wn: @michaelld I tried to make these functions look like the rest of this file. There's a lot of other APIs that have this exact format, which I followed for these.

That's not to say I don't agree with your review comments, but I have follow-up questions:
- Are you comments taking the rest of the file into consideration? E.g., LO controls
- Your message is longer than typical exception messages. Should that go to the logger, or into the exception?
- On the "still let the user execute", are you suggesting to just do nothing, or something else?

mbr0wn: > Your message is longer than typical exception messages. Should that go to the logger, or into the exception?

You've actually answered this....

mbr0wn: OK I opted for a slightly shorter message (which I prefer), but took your suggestion. what do you think/

mbr0wn: ?

michaelld: Yes that works for me. Please fixup in the source as well. Thx!

mbr0wn: > Please fixup in the source as well.

What do you mean by that?

gr-uhd/lib/usrp_source_impl.cc
```diff
@@ -203,6 +203,46 @@ ::uhd::gain_range_t usrp_source_impl::get_gain_range(const std::string& name, si
...
+#endif
+}
+
+void usrp_source_impl::set_power_reference(const double power_dbm, const size_t chan)
+{
+#ifdef UHD_USRP_MULTI_USRP_POWER_LEVEL
+    const size_t dev_chan = _stream_args.channels[chan];
+    _dev->set_rx_power_reference(power_dbm, dev_chan);
+#else
+    throw std::runtime_error("not implemented in this version");
```
michaelld, line 222: I mean here ... this file, for the "source" ...

mbr0wn: Oh! :D

I was thinking "what's he saying, I've clearly changed the source code.... yup, coming right up.

gr-uhd/lib/usrp_sink_impl.cc
```diff
@@ -187,6 +187,54 @@ ::uhd::gain_range_t usrp_sink_impl::get_gain_range(const std::string& name, size
     return _dev->get_tx_gain_range(name, chan);
 }
 
+bool usrp_sink_impl::has_power_reference(const size_t chan)
+{
+#ifdef UHD_USRP_MULTI_USRP_POWER_LEVEL
```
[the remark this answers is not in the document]
mbr0wn: What @marcusmueller said. This is also how all the other dependent-on-UHD-version features are pulled in.

gr-uhd/lib/usrp_sink_impl.cc
```diff
@@ -187,6 +187,54 @@ ::uhd::gain_range_t usrp_sink_impl::get_gain_range(const std::string& name, size
     return _dev->get_tx_gain_range(name, chan);
 }
 
+bool usrp_sink_impl::has_power_reference(const size_t chan)
```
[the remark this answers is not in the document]
mbr0wn: Sure, if you prefer. I've moved to const-ing everything unless required (like Rust) so this ^^^ is just the new muscle memory typing.

gr-uhd/lib/usrp_sink_impl.cc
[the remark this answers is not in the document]
michaelld: Gotcha. Works for me. Thx!

gr-uhd/include/gnuradio/uhd/usrp_block.h
```diff
@@ -229,6 +230,56 @@ class GR_UHD_API usrp_block : public gr::sync_block
     virtual ::uhd::gain_range_t get_gain_range(const std::string& name,
                                                size_t chan = 0) = 0;
 
+    /*! Query if this device is capable of absolute power levels
+     *
+     * If true, the set_power_reference() and get_power_reference() APIs can be
+     * used as well.
```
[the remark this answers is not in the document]
mbr0wn: I added Doxygen's `\throw` tag in all functions.

gr-uhd/include/gnuradio/uhd/usrp_block.h
```diff
@@ -229,6 +230,65 @@ class GR_UHD_API usrp_block : public gr::sync_block
...
+    virtual bool has_power_reference(size_t chan = 0) = 0;
+
+    /*! Set the absolute power reference level for this channel
+     *
+     * Note that this API is available for certain devices only, and only if
+     * calibration data is available. Refer to the UHD manual for greater
+     * detail: https://files.ettus.com/manual/page_power.html
+     *
+     * In a nutshell, using the power reference will configure the device such
+     * that a a full-scale signal (0 dBFS) corresponds to a signal at the
```
michaelld, line 252: "that a a" -> "that a"

gr-uhd/lib/usrp_source_impl.cc
```diff
@@ -203,6 +203,65 @@ ::uhd::gain_range_t usrp_source_impl::get_gain_range(const std::string& name, si
     return _dev->get_rx_gain_range(name, chan);
 }
 
+bool usrp_source_impl::has_power_reference(size_t chan)
+{
+#ifdef UHD_USRP_MULTI_USRP_POWER_LEVEL
+    if (chan >= _stream_args.channels.size()) {
```
michaelld, line 209: Please also use these "chan" checks in "usrp_sink_impl.cc"

mbr0wn: Oh yikes. Thanks for the patient review. That was clearly done in haste.

drmpeg:
Is the hiding of ```Gain Type``` correct? It seems to only show up with ```Absolute (dB)```. ```Normalized``` and ```Absolute Power (dBm)``` are hidden,

drmpeg:
This commit also breaks flow graphs that had ```Gain Type``` set to ```Normalized```. The ```Gain Type``` is set to ```Absolute (dB)``` instead. For example:

gr-uhd/examples/grc/uhd_normalized_gain.grc

mbr0wn:
@drmpeg Thanks for pointing out.

mbr0wn:
@drmpeg There's nothing I can do about the breakage. The reason: Before, there was a Boolean (normalized: true/false). Now we have three options, so the bool won't work, so I had to change it -- and so it's a breaking change. If the person writing that originally (most likely me) had had the foresight to use strings or enums instead of bools, we wouldn't have the issue. But here we are.

drmpeg:
It only affect four examples, so we could just regenerate those.
```
gr-uhd/examples/grc/uhd_msg_tune.grc
gr-uhd/examples/grc/uhd_normalized_gain.grc
gr-uhd/examples/grc/uhd_wbfm_receive.grc
gr-dtv/examples/uhd_atsc_rx.grc
```

How about the hiding behavior?

PR #3714, opened 2020-08-19
grc: prevent search keystrokes from modifying flowgraph

argilo:
This backports #3710 to GNU Radio 3.8. I've tested and this works as in 3.9.


mormj:
Tested this one as well and works the same as on master :+1: 
Did reproduce the issue mentioned in the original PR with selecting multiple blocks, but this is a huge improvement

argilo:
I've added in the follow-up commit from #3715 as well. I tested again in 3.8, and everything appears to work as in 3.9.

PR #3727, opened 2020-08-24
grc: try-except around "gsettings" for when those settings do not exist

michaelld:
subject says it all. Here's the original PR commit and my comments from debugging this issue.


michaelld:
Turns out this is a backport of 7826795923375fef264070e9af4e21f5932737f6 from GR master ... the exact same change, created by 2 different people ... nice!

michaelld:
Given that this change has already been merged into master, and it is small, and critical for GRC functionality, I'm going to just merge this without further review.

PR #3730, opened 2020-08-25
Remove hardcoded GR_PYTHON_DIR

### LINKED ISSUE #3729: GR_PYTHON_DIR can not be set from cmake command line
# Reproducing
```
gr_modtool newmod testoot && pushd gr-testoot && mkdir build && cd build && temp_inst=`mktemp -d` && cmake ../ -DCMAKE_INSTALL_PREFIX=$temp_inst -DGR_PYTHON_DIR=$temp_inst/special-python/dist-packages && make install ; popd
```

```
Install the project...
-- Install configuration: "Release"
-- Installing: /tmp/tmp.NuaXqMHk7S/lib/cmake/testoot/testootConfig.cmake
-- Installing: /tmp/tmp.NuaXqMHk7S/include/testoot/api.h
-- Installing: /tmp/tmp.NuaXqMHk7S/lib/python3/dist-packages/testoot/__init__.py
-- Installing: /tmp/tmp.NuaXqMHk7S/lib/python3/dist-packages/testoot/__init__.pyc
-- Installing: /tmp/tmp.NuaXqMHk7S/lib/python3/dist-packages/testoot/__init__.pyo
```

It ignores the override and uses hard coded value from https://github.com/gnuradio/gnuradio/blob/f871040e8129ccf7b0392aee3744d9867ffa0a1a/cmake/Modules/GnuradioConfig.cmake.in#L38

# Approach
#2939 closed #2006 sets the default and so the GnuradioConfig shouldn't need to set it anymore.

Solving this may also solve #2856

elms:
GrPython.cmake now sets the default correctly thanks to #2939

Closes #3729

### DIFF
--- a/cmake/Modules/GnuradioConfig.cmake.in
+++ b/cmake/Modules/GnuradioConfig.cmake.in
@@ -35,7 +35,6 @@ find_dependency(Volk)
 set(ENABLE_PYTHON @ENABLE_PYTHON@ CACHE BOOL "Enable Python & pybind11 Bindings")
 if(${ENABLE_PYTHON})
   set(PYTHON_EXECUTABLE @PYTHON_EXECUTABLE@)
-  set(GR_PYTHON_DIR @GR_PYTHON_DIR@)
   include(GrPython)
 endif()
 


noc0lour:
This line is only partially related to the specified issue. We need to find a middleground between automatically setting the right GR_PYTHON_DIR and having the option to override it on the commandline.

elms:
@noc0lour Can you explain what you mean by a middle ground?

It seems like the desired behavior is to default to the automatic _right_ version, but allow for users to override if they know what they are doing or want to.

This change will by default not change the behavior and use the same version as base GNU Radio, but allow overriding `GR_PYTHON_DIR` when it is required (which is not possible currently).

I suppose the other option is to complain loudly when someone tries to override it and say that it's being ignored.

noc0lour:
Ah, I see that PYTHON_EXECUTABLE is also passed, I guess this should be enough to have everything else behave sane.

noc0lour (approved):
I think this is a correct change as the correct Python interpreter is set with `PYTHON_EXECUTABLE` and the python dir should be set accoording to install prefix etc and not be inherited from the base installation.

mormj:
Merging according to @noc0lour comment above

noc0lour:
@mormj I intended to test this behaviour with some OOT/script but never got to that point

PR #3731, opened 2020-08-25
blocks: prevent losing rx_time precision in gr_read_file_metadata

daniestevez:
The utility gr_read_file_metadata uses parse_file_metadata.parse_header()
from gr-blocks to parse and print headers in a metadata file. Currently,
the rx_time field is printed with us precision. However, the rx_time is
stored as the integer seconds of the UNIX timestamp in an uint64_t plus
a double storing the fraction of a second, so the rx_time has more
precision.

A precision on the order of 1ns is necessary for many ranging applications,
since 1us is approximately 300m of range.

This modifies parse_file_metadata.parse_header() to print the full
fractional part as stored in the double. Aditionally, it adds
"rx_time_secs" and "rx_time_fracs" fields to the dict() return value of
parse_header(), in case they are ever needed. The "rx_time" field in this
dict() is a double, and as such it is unable to store a UNIX timestamp with
1ns precision.

**Note:** This is just a proposal of how to improve the current behaviour. Others might have better suggestions. In my particular use case I just need `gr_read_file_metadata` to print the full precision of `rx_time` as stored in the headers, since the output of `gr_read_file_metadata` is then parsed by another tool.

### DIFF
--- a/gr-blocks/python/blocks/parse_file_metadata.py
+++ b/gr-blocks/python/blocks/parse_file_metadata.py
@@ -10,6 +10,7 @@
 
 
 import sys
+import decimal
 from gnuradio import gr, blocks
 import pmt
 
@@ -75,12 +76,20 @@ def parse_header(p, VERBOSE=False):
         r = pmt.dict_ref(p, pmt.string_to_symbol("rx_time"), dump)
         secs = pmt.tuple_ref(r, 0)
         fracs = pmt.tuple_ref(r, 1)
-        secs = float(pmt.to_uint64(secs))
+        secs = pmt.to_uint64(secs)
         fracs = pmt.to_double(fracs)
-        t = secs + fracs
+        t = float(secs) + fracs
+        info["rx_time_secs"] = secs
+        info["rx_time_fracs"] = fracs
         info["rx_time"] = t
         if(VERBOSE):
-            print("Seconds: {0:.6f}".format(t))
+                # We need are going to print with ~1e-16 resolution,
+                # which is the precision we can expect in secs.
+                # The default value of precision for decimal
+                # is 28. The value of secs is not expected to be larger
+                # than 1e12, so this precision is enough.
+                s = decimal.Decimal(secs) + decimal.Decimal(fracs)
+                print("Seconds: {0:.16f}".format(s))
     else:
         sys.stderr.write("Could not find key 'time': invalid or corrupt data file.\n")
         sys.exit(1)


mormj (approved):

gr-blocks/python/blocks/parse_file_metadata.py
```diff
@@ -75,12 +75,14 @@ def parse_header(p, VERBOSE=False):
         r = pmt.dict_ref(p, pmt.string_to_symbol("rx_time"), dump)
         secs = pmt.tuple_ref(r, 0)
         fracs = pmt.tuple_ref(r, 1)
-        secs = float(pmt.to_uint64(secs))
+        secs = pmt.to_uint64(secs)
         fracs = pmt.to_double(fracs)
-        t = secs + fracs
+        t = float(secs) + fracs
+        info["rx_secs"] = secs
```
argilo, line 81: Maybe `rx_time_secs`?

gr-blocks/python/blocks/parse_file_metadata.py
```diff
@@ -75,12 +75,14 @@ def parse_header(p, VERBOSE=False):
...
+        info["rx_time_fracs"] = fracs
         info["rx_time"] = t
         if(VERBOSE):
-            print("Seconds: {0:.6f}".format(t))
+            print("Seconds: {0}.{1}".format(secs, str(fracs)[2:]))
```
argilo, line 85: This will no longer print a fixed number of decimal places. Is that intended?

I don't have a strong preference.

gr-blocks/python/blocks/parse_file_metadata.py
daniestevez: Thanks! That seems a better choice of a name. I'll push a commit changing this.

gr-blocks/python/blocks/parse_file_metadata.py
daniestevez: Yes. This is deliberate. I preferred to print out all the resolution that is stored in the double rather than printing a fixed number of decimal places, since different users might have different needs (though I guess that at least 10 decimal places would be a sensible choice).

However, I don't like the way I did this in Python. Maybe there's a better way.

argilo: > Maybe there's a better way.

My suggestion would be to use the `decimal` module:

```
>>> import decimal
>>> decimal.getcontext().prec = 20
>>> secs = 1602513164
>>> fracs = 0.15
>>> print(decimal.Decimal(secs) + decimal.Decimal(fracs))
1602513164.1500000000
```
If you want more than 10 decimal places, you can increase `prec`.

daniestevez: Thanks! That seems a good suggestion. I've just pushed a commit with my take at using `decimal` for this. In fact I want more than 20 decimal places resolution, since I want to have 1e-16 resolution (which is basically what we can expect from `fracs`) no matter what the value of `secs` is. Apparently this works out to 36 decimal places.

gr-blocks/python/blocks/parse_file_metadata.py
```diff
@@ -75,12 +76,21 @@ def parse_header(p, VERBOSE=False):
...
+        info["rx_time_fracs"] = fracs
         info["rx_time"] = t
         if(VERBOSE):
-            print("Seconds: {0:.6f}".format(t))
+            with decimal.localcontext() as ctx:
+                # we need 36 decimal places in order not to lose precision
+                # since an uint64 can contain 20 decimal places and the
+                # resolution we should expect from the fracs duoble
+                # is the machine epsilon, which is ~1e-16
+                ctx.prec = 36
```
argilo, line 91: 26 would probably be enough here, since Unix timestamps won't have more than 10 digits until the year 2286. :smile:

36 will result in a lot of useless trailing digits:
```
>>> print(decimal.Decimal(1600000000) + decimal.Decimal(0.123456789012345))
1600000000.12345678901234499735828365
```

argilo: The default precision is 28, which is enough, so you could also stick with that and limit the number of decimal places in output.

gr-blocks/python/blocks/parse_file_metadata.py
```diff
@@ -75,12 +76,21 @@ def parse_header(p, VERBOSE=False):
...
+                s = decimal.Decimal(secs) + decimal.Decimal(fracs)
+                print("Seconds: {0}.{1}".format(s))
```
argilo, line 93: This needs to be updated. You could update the code like so:
```suggestion
                print("Seconds: {0:.16f}".format(s))
```

gr-blocks/python/blocks/parse_file_metadata.py
daniestevez: You've convinced me that using 36 decimal places was an overkill, so I've changed this to use the default precision, and limited the output to 1e-16 resolution.

gr-blocks/python/blocks/parse_file_metadata.py
daniestevez: This is done in 0dfb8c794c01f481556afe952b7904e96f8d4e39

daniestevez:
At some point before merging it will be appropriate to squash all the commits. Just tell me and I will do that.

argilo:
The code change looks good now, and produces the expected output.

Before:
```
$ gr_read_file_metadata meta.dat 
HEADER 0
Version Number: 0
Sample Rate: 32000.00 sps
Seconds: 1234567890.123457
Item size: 4
Data Type: float (5)
Complex? False
Header Length: 150 bytes
Extra Length:  1
Extra Header?  True
Size of Data: 554492 bytes
              138623.0 items

Extra Header:

```
After:
```
$ gr_read_file_metadata meta.dat 
HEADER 0
Version Number: 0
Sample Rate: 32000.00 sps
Seconds: 1234567890.1234567890123456
Item size: 4
Data Type: float (5)
Complex? False
Header Length: 150 bytes
Extra Length:  1
Extra Header?  True
Size of Data: 554492 bytes
              138623.0 items

Extra Header:

```

argilo (approved):

mormj:
@daniestevez  - LGTM and since @argilo approves, can you please squash, and then I can go ahead and merge this one.  Thanks!

daniestevez:
@mormj thanks! I've just squashed this.

rear1019:
I know I am late to the game, but this is possibly worth considering: `float.hex()` and `float.fromhex()` [1] can be used to avoid precision issues. If human readable output is not a requirement. 

[1] https://docs.python.org/3/library/stdtypes.html#additional-methods-on-float

daniestevez:
Here I would say that we want human readable output. If someone only needs machine readability, then calling `parse_file_metadata.parse_header()` directly and using the data it returns would be better. In the end this is what I've ended up doing in the application that motivated this PR (because I have too many files to managed them manually).

PR #3732, opened 2020-08-26
uhd: uhd_fft: Add support for --power argument

mbr0wn:
This enables absolute power support for uhd_fft. For example:

    uhd_fft --power -20 --args [...]

will run uhd_fft with all the usual settings, but will set the gain such
that the reference power level is -20 dBm. The y-axis will be labeled
differently ("RX Power Level (dBm)") and the gain slider will be
replaced by a power reference level slider.

Reading the actual received power from the display requires some DSP
knowledge. First, the choice of window will move energy into sidelobes,
and of course, scalloping loss plays a role. By choosing a boxcar window,
and moving the signal close to the center of a bin, it is possible to
measure the power of a single tone, assuming the USRP has been
previously calibrated.

The y-axis scaling is achieved by throwing in a multiplier that moves
a full-scale signal to the corresponding power value, and using the
recently introduced features to normalize windows.

### DIFF
--- a/gr-uhd/apps/uhd_fft
+++ b/gr-uhd/apps/uhd_fft
@@ -30,6 +30,7 @@ import ctypes
 import sys
 import threading
 import time
+import math
 from PyQt5 import Qt
 import sip # Needs to be imported after PyQt5, could fail otherwise
 from gnuradio import eng_notation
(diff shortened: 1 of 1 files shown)


mbr0wn:
Screenshot:
image

Note the y-label, and the slider at the bottom.

mbr0wn:
Merger: Please merge after #3733.

PR #3733, opened 2020-08-26
uhd: Update uhd_siggen_gui to allow setting absolute power

mbr0wn:
Note: This requires a USRP with power API enabled.

Summary of changes:
- uhd_app now provides the --power argument and applies it as
  a reference power level. It clashes with --gain. UHDApp now has two
  'modes' for setting the power: Via gain (the previous, standard way)
  or by setting a power. UHDApp calls all the right APIs for each use
  case.
- uhd_siggen_base is now aware of these settings and will make sure the
  output power is correct by combining the requested power and the
  requested amplitude (meaning, it sets a ref power that is at the
  correct level above the requested power level).
- The GUI can now set either power or gain, depending on command line
  arguments. When changing the amplitude in power mode, the ref power is
  also changed so the output power (in dBm) stays the same when the
  amplitude changes.

### DIFF
--- a/gr-uhd/apps/uhd_app.py
+++ b/gr-uhd/apps/uhd_app.py
@@ -28,7 +28,13 @@ LONG_TPL = """{prefix}  Motherboard: {mb_id} ({mb_serial})
 {prefix}  Antenna: {ant}
 """
 
+# PyLint can't reliably detect C++ exports in modules, so let's disable that
+# pylint: disable=no-member
+
 class UHDApp:
+    GAIN_TYPE_GAIN = 'dB'
+    GAIN_TYPE_POWER = 'power_dbm'
+
     " Base class for simple UHD-based applications "
     def __init__(self, prefix=None, args=None):
         self.prefix = prefix
@@ -37,7 +43,7 @@ class UHDApp:
         if self.args.sync == 'auto' and len(self.args.channels) > 1:
             self.args.sync = 'pps'
         self.antenna = None
-        self.gain_range = None
+        self.gain_range = None # Can also be power range
         self.samp_rate = None
         self.has_lo_sensor = None
         self.async_msgq = None
@@ -55,6 +61,7 @@ class UHDApp:
         self.lo_export = None
         self.usrp = None
         self.lo_source_channel = None
+        self.gain_type = self.GAIN_TYPE_GAIN
 
     def vprint(self, *args):
         """
@@ -179,8 +186,20 @@ class UHDApp:
                 ))
         self.antenna = self.usrp.get_antenna(0)
         # Set receive daughterboard gain:
-        self.set_gain(args.gain)
-        self.gain_range = self.usrp.get_gain_range(0)
+        if args.power:
+            if args.gain is not None:
+                print("[ERROR] Providing both --gain and --power is invalid!")
+                sys.exit(1)
+            self.gain_type = self.GAIN_TYPE_POWER
+            if not self.usrp.has_power_reference(0):
+                print("[ERROR] Device does not have power reference capabilities!")
+                sys.exit(1)
+            self.set_power_reference(args.power)
+            self.gain_range = self.usrp.get_power_range(0)
+        else:
+            self.set_gain(args.gain)
+            self.gain_range = self.usrp.get_gain_range(0)
+            self.gain_type = self.GAIN_TYPE_GAIN
         # Set frequency (tune request takes lo_offset):
         if hasattr(args, 'lo_offset') and args.lo_offset is not None:
             treq = uhd.tune_request(args.freq, args.lo_offset)
@@ -268,10 +287,19 @@ class UHDApp:
                         chan=chan, g=self.usrp.get_gain(i)
                     ))
         else:
-            self.vprint("Setting gain to {g} dB.".format(g=gain))
+            self.vprint("Setting gain to {g:.2f} dB.".format(g=gain))
             for chan in range(len(self.channels)):
                 self.usrp.set_gain(gain, chan)
         self.gain = self.usrp.get_gain(0)
+
+    def set_power_reference(self, power_dbm):
+        """
+        Safe power-ref-setter.
+        """
+        assert power_dbm is not None
+        self.vprint("Setting ref power to {g:.2f} dBm.".format(g=power_dbm))
+        self.usrp.set_power_reference(power_dbm)
+        self.gain = self.usrp.get_power_reference(0)
 
     def set_freq(self, freq, skip_sync=False):
         """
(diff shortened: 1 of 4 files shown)


mbr0wn:
Note to merge: This needs to be merged before #3732. My bad, I submitted in wrong order.

mbr0wn:
Screenshot of the tool using `--power`:
image

The only real difference is the gain slider is now a power slider.

gr-uhd/apps/uhd_app.py
```diff
@@ -364,12 +392,11 @@ def cslist(string):
...
-        group.add_argument("--gain-type", choices=('db', 'normalized'), default='db',
-                           help="Gain Type (applies to -g)")
-        group.add_argument("-p", "--power-ref", type=eng_arg.eng_float, default=None,
-                           help="Reference power level (in dBm). "
+        group.add_argument("-p", "--power", type=eng_arg.eng_float, default=None,
+                           help="(Reference) power level (in dBm). "
                                 "Not supported by all devices (see UHD manual). "
-                                "Will fail if not supported. Precludes --gain. ")
+                                "Will fail if not supported. Precludes --gain. "
+                                "Behaviour may differ between applications.")
```
michaelld, line 399: Since `--power` is related to `--amplitude`, I think this relationship should be noted ... here and/or with that argument.

gr-uhd/apps/uhd_siggen_base.py
```diff
@@ -55,6 +57,12 @@ class USRPSiggen(gr.top_block, pubsub, UHDApp):
     def __init__(self, args):
         gr.top_block.__init__(self)
         pubsub.__init__(self)
+        # If the power argument is given, we need to turn that into a power
+        # *reference* level. This is a bit of a hack because we're assuming
+        # knowledge of UHDApp (i.e. we're leaking abstractions). But it's simple
+        # and harmless enough.
+        if args.power:
+            args.power -= 20 * math.log10(args.amplitude)
```
michaelld, line 65: should there be a check here for `args.amplitude > 0.0`?

mbr0wn: Yeah good idea. It should be elsewhere, I guess, since a negative amplitude makes no sense anyway.

gr-uhd/apps/uhd_app.py
mbr0wn: `uhd_app.py` shouldn't have `--amplitude` anyway, that's specific to siggen. I'll move it out.

mbr0wn: To clarify: Changing `--amplitude` does not change the output power (UHD will shift gains around accordingly). I'll add a comment.

gr-uhd/apps/uhd_siggen_base.py
mbr0wn: I've put it right before this.

mbr0wn:
New exception:
```
ValueError: Invalid value for amplitude: 12.0. Must be in [0.0, 1.0]
```

Also throws for neg values.

`--amplitude` now only shows up for siggen:
```
Siggen Arguments:
  -m AMPL, --amplitude AMPL
                        Set output amplitude to AMPL (0.0-1.0). Note that if
                        --power is given, UHD will attempt to match the output
                        power regardless of the amplitude.
  -x WAVEFORM_FREQ, --waveform-freq WAVEFORM_FREQ
                        Set baseband waveform frequency to FREQ
  -y WAVEFORM2_FREQ, --waveform2-freq WAVEFORM2_FREQ
                        Set 2nd waveform frequency to FREQ
  --sine                Generate a carrier modulated by a complex sine wave
  --const               Generate a constant carrier
  --offset OFFSET       Set waveform phase offset to OFFSET
  --gaussian            Generate Gaussian random output
  --uniform             Generate Uniform random output
  --2tone               Generate Two Tone signal for IMD testing
  --sweep               Generate a swept sine wave

GUI Arguments:
  -q, --show-freq-sink  Show QT Frequency Widget
```

gr-uhd/apps/uhd_app.py
```diff
@@ -179,8 +186,20 @@ def setup_usrp(self, ctor, args, cpu_format='fc32'):
                 ))
         self.antenna = self.usrp.get_antenna(0)
         # Set receive daughterboard gain:
-        self.set_gain(args.gain)
-        self.gain_range = self.usrp.get_gain_range(0)
+        if args.power:
+            if args.gain is not None:
+                print("[ERROR] Cannot provide both --gain and --power!")
```
michaelld, line 191: Prefer "Cannot parse" or something like that. The user can _provide_ both, but the app here can't use them both ...

gr-uhd/apps/uhd_siggen_base.py
```diff
@@ -234,8 +246,48 @@ def set_amplitude(self, amplitude):
...
+        self.update_gain_range()
         return True
 
+    def get_gain_or_power(self):
+        """
+        Depending on gain type, return either a power level or the current gain
+        """
+        if self.gain_type == self.GAIN_TYPE_GAIN:
+            return self.usrp.get_gain(self.channels[0])
+        # else:
```
michaelld, line 258: this line can be removed

gr-uhd/apps/uhd_siggen_base.py
```diff
@@ -55,6 +57,16 @@ class USRPSiggen(gr.top_block, pubsub, UHDApp):
     def __init__(self, args):
         gr.top_block.__init__(self)
         pubsub.__init__(self)
+        if not 0.0 <= args.amplitude <= 1.0:
```
michaelld, line 60: technically: 0.0 < args.ampl ... not <= ... log(0) -> -inf ... which we want to avoid here ... this the "Must be" line would be "(0.0, 1.0]" to be math notationally correct ... yes?

gr-uhd/apps/uhd_siggen_gui
```diff
@@ -190,7 +191,8 @@ class uhd_siggen_gui(Qt.QWidget):
         self._freq2_offset_win.setEnabled(self._sg[uhd_siggen.TYPE_KEY] in self._freq2_enable_on)
         self.top_grid_layout.addWidget(self._freq2_offset_win, 4, 3, 1, 2)
         ### Amplitude
-        self._amplitude_range = Range(0, 1, .001, .7, 200)
+        self._amplitude_range = \
+            Range(0, 1, .001, self._sg[uhd_siggen.AMPLITUDE_KEY], 200)
```
michaelld, line 195: bottom of AMPL range shoould be not "0" ... but the minimum delta .. so 0.001 ... to avoid the log(0) issue ... yes?

gr-uhd/apps/uhd_siggen_base.py
mbr0wn: I often leave this in just as a reading aid, but I'm also happy to remove it.

gr-uhd/apps/uhd_siggen_base.py
mbr0wn: No, 0.0 is a valid amplitude for practical reasons (want to temporarily turn off the signal to see what spurs are coming from the LOs or other clocks, for example). [0.0, 1.0] is correct.
But you're right, when in power mode, we need a way of handling that. I think I'll follow your suggestions until someone has a better idea -- but only conditionally on power mode.

gr-uhd/apps/uhd_siggen_gui
mbr0wn: See above.

gr-uhd/apps/uhd_siggen_base.py
mbr0wn: OK implemented and tested that.

michaelld (approved):
LGTM ... good changes!

PR #3736, opened 2020-08-26
Add `override` for common virtual function overrides

ThomasHabets:
Mostly done with:

```
find -name "*_impl.h" | xargs sed -i -r '/(void forecast|int work|int general_work|bool check_topology)\(/{:back /\)/b nxt;N;b back;:nxt s/\)$|\)(;)/) override\1/g}'
```

Then I removed an incorrect `work` that this found.

It may seem like a silly change, but it did find some confusing code I deleted.


mormj (approved):

marcusmueller:
Ha! Sorry for letting this ripen for so long. Generally, nice work! (would have done it through a `clang-tidy` pass, though :))

Let me go through this real quickly.

marcusmueller:
huh, this introduces failing unit tests, which is surprising, but probably a positive side effect...

ThomasHabets:
True, but clang-tidy would not have found the incorrect `work` declaration. :-)

marcusmueller:
@ThomasHabets I'm still throwing in the whole-tree clang-tidy rework. My compiler otherwise becomes *very* verbose about inconsistent `override` usage...

marcusmueller:
Man, I'm not doing this nice and tidy today...

marcusmueller (approved):

marcusmueller:
@ThomasHabets as soon as these checks pass, could I ask you to review this?

ThomasHabets:
@marcusmueller looks good

PR #3745, opened 2020-08-30
zeromq: Remove manual memory management

ThomasHabets:
I believe this fixes a memory leak, as the thread objects were never deleted.

### DIFF
--- a/gr-zeromq/lib/base_impl.cc
+++ b/gr-zeromq/lib/base_impl.cc
@@ -25,7 +25,12 @@ base_impl::base_impl(int type,
                      int timeout,
                      bool pass_tags,
                      const std::string& key)
-    : d_vsize(itemsize * vlen), d_timeout(timeout), d_pass_tags(pass_tags), d_key(key)
+    : d_context(1),
+      d_socket(d_context, type),
+      d_vsize(itemsize * vlen),
+      d_timeout(timeout),
+      d_pass_tags(pass_tags),
+      d_key(key)
 {
     /* "Fix" timeout value (ms for new API, us for old API) */
     int major, minor, patch;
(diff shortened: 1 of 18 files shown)


mbr0wn (approved):

gr-zeromq/lib/base_impl.cc
```diff
@@ -25,7 +25,12 @@ base_impl::base_impl(int type,
                      int timeout,
                      bool pass_tags,
                      const std::string& key)
-    : d_vsize(itemsize * vlen), d_timeout(timeout), d_pass_tags(pass_tags), d_key(key)
+    : d_context(1),
+      d_socket(d_context, type),
```
talih0, line 29: This dependent initialization is probably not a good idea. It is prone to fail if the order of the objects is changed in the header file. 
See https://isocpp.org/wiki/faq/ctors#ctor-initializer-order

ThomasHabets: I think the alternative is worse.
Before this PR both the construction and destruction must be in the correct order. And breaking one does not create a warning with any compiler.

With this PR in place changing either location (declaration or this list) will trigger a warning.

[Discussion in C++ core guideline C.47](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#discussion-define-and-initialize-member-variables-in-the-order-of-member-declaration) uses dependent member initialization in its example so does not appear to have a negative opinion on it.

mbr0wn: If the order were to be changed, gcc would throw a warning. C++ initializes attributes in declaration order, and that's something that can be relied upon. So as long as there's no warnings about initializing in the wrong order, we can be sure this is working.

marcusmueller (approved):

PR #3746, opened 2020-08-30
dtv: Remove manual memory management

ThomasHabets:
All removed except some tricky stuff in `dvb_ldpc_bb_impl`.

### DIFF
--- a/gr-dtv/lib/atsc/atsc_deinterleaver_impl.cc
+++ b/gr-dtv/lib/atsc/atsc_deinterleaver_impl.cc
@@ -33,7 +33,7 @@ atsc_deinterleaver_impl::atsc_deinterleaver_impl()
     m_fifo.reserve(s_interleavers);
 
     for (int i = 0; i < s_interleavers; i++)
-        m_fifo.emplace_back((52 - 1 - i) * 4);
+        m_fifo.emplace_back((s_interleavers - 1 - i) * 4);
 
     sync();
 }
@@ -64,8 +64,8 @@ int atsc_deinterleaver_impl::work(int noutput_items,
         if (in[i].pli.first_regular_seg_p())
             sync();
 
-        // remap OUTPUT pipeline info to reflect 52 data segment end-to-end delay
-        plinfo::delay(out[i].pli, in[i].pli, 52);
+        // remap OUTPUT pipeline info to reflect all data segment end-to-end delay
+        plinfo::delay(out[i].pli, in[i].pli, s_interleavers);
 
         // now do the actual deinterleaving
         for (unsigned int j = 0; j < sizeof(in[i].data); j++) {
--- a/gr-dtv/lib/atsc/atsc_deinterleaver_impl.cc
+++ b/gr-dtv/lib/atsc/atsc_deinterleaver_impl.cc
@@ -30,25 +30,22 @@ atsc_deinterleaver_impl::atsc_deinterleaver_impl()
                      io_signature::make(1, 1, sizeof(atsc_mpeg_packet_rs_encoded))),
       alignment_fifo(156)
 {
-    m_fifo.resize(52);
+    m_fifo.reserve(s_interleavers);
 
-    for (int i = 0; i < 52; i++)
-        m_fifo[52 - 1 - i] = new interleaver_fifo<unsigned char>(i * 4);
+    for (int i = 0; i < s_interleavers; i++)
+        m_fifo.emplace_back((52 - 1 - i) * 4);
 
     sync();
 }
 
-atsc_deinterleaver_impl::~atsc_deinterleaver_impl()
-{
-    for (int i = 0; i < 52; i++)
-        delete m_fifo[i];
-}
+atsc_deinterleaver_impl::~atsc_deinterleaver_impl() {}
 
 void atsc_deinterleaver_impl::reset()
 {
     sync();
-    for (int i = 0; i < 52; i++)
-        m_fifo[i]->reset();
+
+    for (auto& i : m_fifo)
+        i.reset();
 }
 
 int atsc_deinterleaver_impl::work(int noutput_items,
--- a/gr-dtv/lib/atsc/interleaver_fifo.h
+++ b/gr-dtv/lib/atsc/interleaver_fifo.h
@@ -11,7 +11,8 @@
 #ifndef _INTERLEAVER_FIFO_H_
 #define _INTERLEAVER_FIFO_H_
 
-#include <string.h>
+#include <algorithm>
+#include <vector>
 
 /*!
  * \brief template class for interleaver fifo
@@ -22,7 +23,6 @@ class interleaver_fifo
 {
 public:
     interleaver_fifo(unsigned int size);
-    ~interleaver_fifo();
 
     //! reset interleaver (flushes contents and resets commutator)
     void reset();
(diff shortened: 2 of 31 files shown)


gr-dtv/lib/atsc/atsc_deinterleaver_impl.cc
```diff
@@ -30,25 +30,22 @@ atsc_deinterleaver_impl::atsc_deinterleaver_impl()
                      io_signature::make(1, 1, sizeof(atsc_mpeg_packet_rs_encoded))),
       alignment_fifo(156)
 {
-    m_fifo.resize(52);
+    m_fifo.reserve(s_interleavers);
 
-    for (int i = 0; i < 52; i++)
-        m_fifo[52 - 1 - i] = new interleaver_fifo<unsigned char>(i * 4);
+    for (int i = 0; i < s_interleavers; i++)
+        m_fifo.emplace_back((52 - 1 - i) * 4);
```
jdemel, line 36: Is `s_interleaver` a constant? aka `s_interleaver = 52;`? In that case, I'd suggest to use the constant consistently, I assume it would be correct to replace `52` with `s_interleaver` as well. Also, the other literals might be good candidates for replacement too. Well, only `4`.

Found it: `static constexpr int s_interleavers = 52;`. Since this is a constant, would it be better to spell it `S_INTERLEAVERS`?

gr-dtv/lib/atsc/interleaver_fifo.h
```diff
@@ -22,52 +23,40 @@ class interleaver_fifo
 {
 public:
     interleaver_fifo(unsigned int size);
```
jdemel, line 25: What is the difference between this FIFO and the one in `atsc_interleaver_fifo.h`? Can they be merged into one?

gr-dtv/lib/dvbt/dvbt_map_impl.h
```diff
@@ -20,20 +20,20 @@ class dvbt_map_impl : public dvbt_map
...
-    unsigned char d_alpha;
+    const unsigned char d_alpha;
     // Gain for the complex values
-    float d_gain;
+    const float d_gain;
 
-    gr_complex* d_constellation_points;
+    std::vector<gr_complex> d_constellation_points;
 
     void make_constellation_points(int size, int step, int alpha);
```
jdemel, line 38: What's the difference betwenn this function and the one in `dvbt_demap`? Would it be possible to have one constellation object for both?

gr-dtv/lib/dvbt/dvbt_viterbi_decoder_impl.h
```diff
@@ -101,13 +103,13 @@ class dvbt_viterbi_decoder_impl : public dvbt_viterbi_decoder
     int mettab[2][256];
```
jdemel, line 103: I know this has fixed size, but  for consistency, wouldn't it be better to use a `std::vector` here as well?

gr-dtv/lib/dvbt2/dvbt2_paprtr_cc_impl.cc
```diff
@@ -808,22 +736,26 @@ int dvbt2_paprtr_cc_impl::work(int noutput_items,
...
-                            ctemp, ones_freq, result, papr_fft_size);
-                        volk_32f_x2_subtract_32f(
-                            (float*)c, (float*)c, (float*)ctemp, papr_fft_size * 2);
-                        memcpy(r, rNew, sizeof(gr_complex) * N_TR);
+                            ctemp.data(), ones_freq.data(), result, papr_fft_size);
+                        volk_32f_x2_subtract_32f((float*)c.data(),
+                                                 (float*)c.data(),
+                                                 (float*)ctemp.data(),
+                                                 papr_fft_size * 2);
+                        std::copy(std::begin(rNew), std::end(rNew), std::begin(r));
```
jdemel, line 755: So `std::copy` translates to a `memcpy`? Or does it incur more overhead? I've seen some `memcpy` in the other files, that might want to be replaced as well.

gr-dtv/lib/dvbt2/dvbt2_paprtr_cc_impl.h
```diff
@@ -94,6 +94,11 @@ class dvbt2_paprtr_cc_impl : public dvbt2_paprtr_cc
                          float vclip,
                          int iterations,
                          unsigned int vlength);
+
+    // Disable copy because raw pointers.
+    dvbt2_paprtr_cc_impl(const dvbt2_paprtr_cc_impl&) = delete;
```
jdemel, line 99: Which members are raw pointers? Do you refer to `p2_papr_map_*k`?

jdemel (commented):
I have a few comments. Besides this looks good. It's amazing how modern c++ enables one to do more with fewer lines of code.

willcode (approved):
Great improvement. If it works and performance is similar or better, we could come back and make more tweaks later.

gr-dtv/lib/atsc/atsc_deinterleaver_impl.cc
ThomasHabets: Static member variables [should be prefixed with `s_`](https://github.com/gnuradio/greps/blob/master/grep-0001-coding-guidelines.md#code-formatting-guidelines).

I could make it a global variable in `gr::dtv` and capitalize it, but there's no existing namespace outside the class for just atsc.

gr-dtv/lib/atsc/interleaver_fifo.h
ThomasHabets: I don't know. I'd say that would make for a separate PR, though.

gr-dtv/lib/dvbt/dvbt_map_impl.h
ThomasHabets: Same here. Looks like there more simplifying work to do in `dtv` that isn't about removing manual memory management.

gr-dtv/lib/dvbt/dvbt_viterbi_decoder_impl.h
ThomasHabets: Maybe `std::array<>` would be even better.

I'd say so, for another PR. :-)

gr-dtv/lib/dvbt2/dvbt2_paprtr_cc_impl.cc
ThomasHabets: `std::copy` is like `memcpy` but actually knows what it's copying. `memcpy` has to be told how many bytes, which makes it easy to get something wrong, as types change and when copy-pasting.

Usually template implementations actually compile down to faster than simple functions (because the compiler has type access . Though `memcpy` is special in most compilers.

tl;dr: they'll be just as fast, and with today's compilers may even compile down to the exact same instructions. That's my default unless benchmarks prove otherwise.

gr-dtv/lib/dvbt2/dvbt2_paprtr_cc_impl.h
ThomasHabets: `p2_papr_map` and `tr_papr_map`.

But hmm… actually those point to const static lifetime objects, so are safe to copy. Removing these lines.

ThomasHabets:
Running tests before pushing revised version.

ThomasHabets:
Updated.

jdemel (approved):
LGTM.

PR #3750, opened 2020-09-03
gr-blocks: More graceful error handling of missing sndfile dependency.

drmpeg:


mormj (changes requested):
It seems a bit draconian to enforce libsndfile1 for all of gr-blocks, since it is only used for the wavfile blocks.  Can SNDFILE_FOUND be made to conditionally compile and install the wavfile related files? So that blocks will compile without libsndfile1, and just not have wavfile blocks?

drmpeg:
Okay, but how do you exclude the wavfile blocks from the block tree?

mormj (approved):
Tested in a docker with Ubuntu 20.04 with and without libsndfile1.  Thanks for the quick turnaround and making the dependency less of a dependency!

dkozel (approved):

gr-blocks/lib/CMakeLists.txt
```diff
@@ -168,15 +168,29 @@ add_library(gnuradio-blocks
...
+set(BLOCKS_LIBS
     gnuradio-runtime
     Volk::volk
-    sndfile::sndfile
-)
+    )
+
+if (SNDFILE_FOUND)
+    set(BLOCKS_LIBS ${BLOCKS_LIBS}
+        sndfile
```
ryanvolz, line 189: Using the library name `sndfile` instead of the CMake target `sndfile::sndfile` seems to have come up because of issues with gr-osmosdr according to https://lists.gnu.org/archive/html/discuss-gnuradio/2020-09/msg00055.html. This fix feels like a workaround to another issue, either the target not being properly defined in CMake for OOT modules or that sndfile should be linked privately. As someone who has mucked around with packaging GR for conda, I much prefer a CMake error for a missing target to a linker error, or even worse, silently linking to a library that's not explicitly listed as a dependency (which it seems is what would happen currently with gr-osmosdr).

drmpeg: It's over my head. @noc0lour Any comment?

drmpeg: I was able to figure it out by using the FFT component as a guide. The primary issue was not adding SNDFILE to GR_LIBRARY_FOO in gr-blocks/lib/CMakeLists.txt.

drmpeg: Update pushed.

jacobgilbert-snl:
I built and verified that this works, and fixes an issue (#3772) i had with an OOT that requires the `blocks` component from GR.

dkozel:
Merging before a check on macOS since this is affecting a variety of users and the change seems unlikely to be platform dependent.

michaelld:
This PR shouldn't be an issue on macOS any more than it is on any other OS where SNDFILE is not found. The Cmake looks good to me & resolved a serious build issue. :+1:

PR #3754, opened 2020-09-04
blocks: New block 'Stream Demux'

### LINKED ISSUE #1995: Stream demux block
There is a very useful stream mux block, but there is no corresponding stream demux block.  It would be very useful for simple fixed-sized packets.

dave-pi:
Stream demuxing block to demultiplex one stream into N output streams.

Demuxes a stream producing N outputs streams that contains n_0 items in
the first stream, n_1 items in the second, etc. and repeats. Number of
items of each output stream is specified using the 'lengths' parameter
like so [n_0, n_1, ..., n_N-1].

Example:
lengths = [2, 3, 4]
input stream: [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, ...]
output_streams: [0, 1, 9, 10, ...]
                [2, 3, 4, 11, ...]
                [5, 6, 7, 8, ...]

Closes #1995


gr-blocks/lib/stream_demux_impl.cc
```diff
@@ -0,0 +1,120 @@
...
+    const char* in = (const char*)input_items[0];
+    int input_index = 0;                             // Items read
+    gr_vector_int output_index(d_lengths.size(), 0); // Items written
+    std::vector<gr::tag_t> stream_t;
+
+    while (input_index < ninput_items[0]) {
+        if (noutput_items <= output_index[d_stream]) {
+            break;
+        }
+        int space_left_in_buffers =
```
ThomasHabets, line 82: sprinkle some `const` on these temp variables for clarity.

ThomasHabets (changes requested):

gr-blocks/lib/stream_demux_impl.cc
```diff
@@ -0,0 +1,120 @@
...
+
+int stream_demux_impl::general_work(int noutput_items,
+                                    gr_vector_int& ninput_items,
+                                    gr_vector_const_void_star& input_items,
+                                    gr_vector_void_star& output_items)
+{
+    char* out;
+    const char* in = (const char*)input_items[0];
+    int input_index = 0;                             // Items read
+    gr_vector_int output_index(d_lengths.size(), 0); // Items written
```
ThomasHabets, line 75: `,0` is superflous.

ThomasHabets (changes requested):

gr-blocks/lib/stream_demux_impl.cc
```diff
@@ -0,0 +1,120 @@
...
+            throw std::invalid_argument("At least one size must be non-zero.");
+        }
+    }
+    d_residual = d_lengths[d_stream];
+    set_tag_propagation_policy(TPP_DONT);
+}
+
+void stream_demux_impl::forecast(int noutput_items, gr_vector_int& ninput_items_required)
+{
+    unsigned ninputs = ninput_items_required.size();
```
ThomasHabets, line 61: `std::fill(std::begin(ninput_items_required), std::end(ninput_items_required), 1)`

ThomasHabets (changes requested):

gr-blocks/lib/stream_demux_impl.cc
```diff
@@ -0,0 +1,120 @@
...
+stream_demux_impl::stream_demux_impl(size_t itemsize, const std::vector<int>& lengths)
+    : gr::block("stream_demux",
+                gr::io_signature::make(1, 1, itemsize),
+                gr::io_signature::make(1, -1, itemsize)),
+      d_itemsize(itemsize),
+      d_stream(0),
+      d_residual(0),
+      d_lengths(lengths)
+{
+    while (d_lengths[d_stream] == 0) {
```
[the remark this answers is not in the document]
dave-pi: We can have zero length output streams, just not all of them. We can't filter them out since we need to keep track which ones to write to.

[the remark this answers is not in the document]
ThomasHabets: Doh, of course.

gr-blocks/lib/stream_demux_impl.cc
```diff
@@ -0,0 +1,114 @@
...
+            );
+        const int items_to_copy = std::min(space_left_in_buffers, d_residual);
+        const char* in =
+            reinterpret_cast<const char*>(input_items[0]) + input_index * d_itemsize;
+        char* out = reinterpret_cast<char*>(output_items[d_stream]);
+        memcpy(&out[output_index[d_stream] * d_itemsize], in, items_to_copy * d_itemsize);
+
+        std::vector<gr::tag_t> stream_t;
+        get_tags_in_window(stream_t, 0, input_index, input_index + items_to_copy);
+        for (auto t : stream_t) {
```
marcusmueller, line 85: You don't want to work with a copy of the `tag_t`s (that comes with reference increase/decrease penalty), you'd want to simply work with a reference, thus

```
for (auto &x : stream_t)
```

the rest of the code stays exactly the same.

marcusmueller (commented):
Just a performance recommendation (i.e. low-hanging fruit).

Can be merged without (and will be latest by 2020-09-29), but optimally would be fixing the `for` loop

gr-blocks/lib/stream_demux_impl.cc
dave-pi: Good point. So, just to check my understanding, `get_tags_in_window()` already makes a copy (via copy assignment [here](https://github.com/gnuradio/gnuradio/blob/master/gnuradio-runtime/lib/buffer.cc#L342)) so `auto t : stream t` would be doing another redundant copy (via copy ctor). The reference counting you are refering to is from `pmt_t` objects in `tag_t`, which are `shared_ptr`?

mormj:
Appears comments have been resolved - and the indication by Marcus that it can be merged.  Merging!

dave-pi:
Awesome! I'll write some notes on the wiki.

mormj:
Good point - I don't see the Stream Mux block in the list of blocks either.  It would be great to have a wiki page for both the Stream Mux and Stream Demux

dave-pi:
I've added a page for the Stream Demux block. Both are present in the blocks list.
There were some review comments that could just as easily apply to the Stream Mux, so I'll do some work on that.

PR #3758, opened 2020-09-09
Resolved compiling errors on macOS

MKesenheimer:
I resolved two compiler errors on macOS.
- size() of a vector was missing
- GR_LOG_ERROR was used in a static function

### DIFF
--- a/gr-audio/lib/osx/osx_sink.cc
+++ b/gr-audio/lib/osx/osx_sink.cc
@@ -933,9 +933,7 @@ OSStatus osx_sink::au_output_callback(void* in_ref_con,
                     << "number of available items changing "
                     << "unexpectedly (should never happen): was " << in_number_frames
                     << " now " << t_n_output_items;
-#if _OSX_AU_DEBUG_RENDER_
-                GR_LOG_ERROR(d_logger, msg.str());
-#endif
+                GR_LOG_ERROR(This->d_logger, msg.str());
                 err = kAudioUnitErr_TooManyFramesToProcess;
             }
         }
--- a/gr-audio/lib/osx/osx_sink.cc
+++ b/gr-audio/lib/osx/osx_sink.cc
@@ -115,7 +115,7 @@ void osx_sink::setup()
             err_str += ": ";
             for (UInt32 nn = 0; nn < all_names.size(); ++nn) {
                 err_str += all_names[nn];
-                if (nn != all_names - 1) {
+                if (nn != all_names.size() - 1) {
                     err_str += ", ";
                 }
             }
@@ -933,7 +933,9 @@ OSStatus osx_sink::au_output_callback(void* in_ref_con,
                     << "number of available items changing "
                     << "unexpectedly (should never happen): was " << in_number_frames
                     << " now " << t_n_output_items;
+#if _OSX_AU_DEBUG_RENDER_
                 GR_LOG_ERROR(d_logger, msg.str());
+#endif
                 err = kAudioUnitErr_TooManyFramesToProcess;
             }
         }


gr-audio/lib/osx/osx_sink.cc
```diff
@@ -115,7 +115,7 @@ void osx_sink::setup()
             err_str += ": ";
             for (UInt32 nn = 0; nn < all_names.size(); ++nn) {
                 err_str += all_names[nn];
-                if (nn != all_names - 1) {
+                if (nn != all_names.size() - 1) { // size() was missing
```
mbr0wn, line 118: Please remove the comment. Yeah it's wrong -- no need to track our mistakes in the code, it's enough to track them in the git log.

gr-audio/lib/osx/osx_sink.cc
```diff
@@ -933,7 +933,9 @@ OSStatus osx_sink::au_output_callback(void* in_ref_con,
                     << "number of available items changing "
                     << "unexpectedly (should never happen): was " << in_number_frames
                     << " now " << t_n_output_items;
-                GR_LOG_ERROR(d_logger, msg.str());
+                // commented due to error "invalid use of member 'd_logger' in static
+                // member function"
+                // GR_LOG_ERROR(d_logger, msg.str());
```
mbr0wn, line 938: Instead of the comment, can you encapsulate these lines into 
```cpp
#if _OSX_AU_DEBUG_RENDER_
// ...
#endif
```

Now, frankly, that doesn't fix the fact that we're calling `d_logger` in a static function, but your goal was to make it compile (and that'll do that) while also maintaining consistency. Maybe someone else will fix the issue that arises if _OSX_AU_DEBUG_RENDER_ is actually asserted (maybe you? :innocent: ) But for now, that would do.

mbr0wn (changes requested):
Can you please change as mentioned, and also squash the commits, and prefix them with `audio: osx: `.

MKesenheimer:
Ok, thanks for the tips. I did what you requested and squashed the commits.

mbr0wn (approved):

mbr0wn:
@michaelld had you seen these compiler errors before?

michaelld:
@mormj let me update to the latest master commit and see what happens. Both changes look legit on first read.

ryanvolz:
I'm working on a branch of my conda recipe to compile 3.9 master, and this is the patch I came up with to solve compilation errors on macOS: https://github.com/ryanvolz/gnuradio-feedstock/blob/3.9pre/recipe/0001-gr-audio-osx-Fix-compilation.patch. I was going to get around to submitting it, but since this issue exists already I figured I'd just comment. All you need to access the logger objects is to use the already-defined This instance.

marcusmueller (approved):

michaelld:
I'm still fighting with my MacPorts' install ... I intermix too much dev work with production / release ports! I expect to be able to test this tweak out by the end of the week.

In the meantime: @MKesenheimer can you test out the tweaks that ryanvolz came up with? I'd much prefer to keep logging in place using `This->` if that works rather than #if'def'ing it out. Thx!

ryanvolz:
I put my patch in a PR over at #3787.

MKesenheimer:
> I'm still fighting with my MacPorts' install ... I intermix too much dev work with production / release ports! I expect to be able to test this tweak out by the end of the week.
> 
> In the meantime: @MKesenheimer can you test out the tweaks that ryanvolz came up with? I'd much prefer to keep logging in place using `This->` if that works rather than #if'def'ing it out. Thx!

Sure. I tried the tweaks that ryanvolz came up with and they seem to work. Compilation succeeds when using logging with `This->`.
Should I push the new changes or are you using ryanvolz' branch?

michaelld:
Sorry for the delay here folks! I didn't see @ryanvolz 's PR until just now looking through this PR. Given that those changes are ready to go -- and this PR requires a "squash" -- I'm going to close this one out and go with that PR. Thanks for all of your work here @MKesenheimer !

PR #3764, opened 2020-09-14
uhd: Fix gain type hiding, fix uhd_normalized_gain.grc

mbr0wn:
Gain type hiding was inverted: It was supposed to hide gain type for
default (absolute/dB), and not hide for normalized and absoulute/dBm.
Fixed by this commit.

Also fixes an issue in uhd_normalized_gain.grc, where the gain type was
set to absolute/dB.

This was broken by one of the recent power API commits.

The example now looks like this:
image
Note that the gain type is not hidden (it says "Normalized"). The example was tested and works.

### DIFF
--- a/gr-uhd/examples/grc/uhd_normalized_gain.grc
+++ b/gr-uhd/examples/grc/uhd_normalized_gain.grc
@@ -1,6 +1,7 @@
 options:
   parameters:
     author: ''
+    catch_exceptions: 'True'
     category: Custom
     cmake_opt: ''
     comment: ''
@@ -22,12 +23,11 @@ options:
     sizing_mode: fixed
     thread_safe_setters: ''
     title: Example to showcase absolute vs. relative gain settings
-    window_size: 1280, 1024
   states:
     bus_sink: false
     bus_source: false
     bus_structure: null
-    coordinate: [8, 12.0]
+    coordinate: [11, 9]
     rotation: 0
     state: enabled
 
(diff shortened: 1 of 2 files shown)


mbr0wn:
This issue was reported by @drmpeg.

gr-uhd/examples/grc/uhd_normalized_gain.grc
```diff
@@ -1,6 +1,7 @@
 options:
   parameters:
     author: ''
+    catch_exceptions: 'True'
```
michaelld, line 4: for our edification, why are there so many changes in this GRC script?

michaelld (approved):
The logic LGTM ... sorry I didn't catch it the first time around!

Since the -required- changes are just in just file "gr-uhd/grc/gen_uhd_usrp_blocks.py", wondering why so many changes to the file "gr-uhd/examples/grc/uhd_normalized_gain.grc"?

gr-uhd/examples/grc/uhd_normalized_gain.grc
drmpeg: He regenerated the flow graph to just change gain type back to normalized. But it's an old flow graph, so you get all the changes made to GNU Radio in addition.

michaelld: that's my guess too ... OK so if that's the case I'd prefer these to be 2 separate commits ... since these are not required for the others -- the main purpose of this PR -- to work ... assuming my note here correct that is ... if they are required to be together, please note as such and I'll be OK with that. THx!

PR #3767, opened 2020-09-15
grc: Remove extra logger in grc/core/platform

sdh11:
This is an update for PR #2424. The extra debug output from GRC was
being generated by another logger instance in grc/core/platform that
did not respect the command line arguments.

This removes the extra logger and sets the logger level in main to
whatever is requested from the command line. See grc/main.py:43 for more
detail.

### DIFF
--- a/grc/main.py
+++ b/grc/main.py
@@ -2,7 +2,7 @@
 # This file is part of GNU Radio
 #
 # SPDX-License-Identifier: GPL-2.0-or-later
-# 
+#
 
 import argparse, logging, sys
 
@@ -40,7 +40,15 @@ def main():
     # Enable logging
     # Note: All other modules need to use the 'grc.<module>' convention
     log = logging.getLogger('grc')
-    log.setLevel(logging.INFO)
+    # NOTE: This sets the log level to what was requested for the logger on the
+    # command line, but this may not be the correct approach if multiple handlers
+    # are intended to be used. The logger level shown here indicates all the log
+    # messages that are captured and the handler levels indicate messages each
+    # handler will output. A better approach may be resetting this to logging.DEBUG
+    # to catch everything and making sure the handlers have the correct levels set.
+    # This would be useful for a future GUI logging window that can filter messages
+    # independently of the console output. In this case, this should be DEBUG.
+    log.setLevel(LOG_LEVELS[args.log])
 
     # Console formatting
     console = logging.StreamHandler()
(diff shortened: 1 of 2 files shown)


drmpeg (approved):
Tests fine here. How do you set the logging level on the command line?

sdh11:
`gnuradio-companion --log=debug` should dump all of the debug output.

grc/main.py
```diff
@@ -40,7 +40,15 @@ def main():
...
     log = logging.getLogger('grc')
-    log.setLevel(logging.INFO)
+    # NOTE: This sets the log level to what was requested for the logger on the
+    # command line, but this may not be the correct approach if multiple handlers
+    # are intended to be used. The logger level shown here indicates all the log
+    # messages that are captured and the handler levels indicate messages each
+    # handler will output. A better approach may be resetting this to logging.DEBUG
+    # to catch everything and making sure the handlers have the correct levels set.
+    # This would be useful for a future GUI logging window that can filter messages
+    # independently of the console output. In this case, this should be DEBUG.
```
mormj, line 50: Do you think bringing grc logging into the grc preferences file would be feasible.  It seems that GRC has a lot of functionality for configuring the loggers for each module, which is currently hardcoded, and the command line is very limiting.  The prefs file might provide a more flexible place to specify all the granularity.

sdh11: That should be possible. Are you thinking that each module could have different logging levels?

mormj: I was just going off the comment `Note: All other modules need to use the 'grc.<module>' convention` which seems to imply that was the original design.  Seems like it could be overkill, in which case a single grc logging level would be fine.  But you and @skoslowski should have a better idea what is appropriate grc logging granularity

sdh11: Ah ok. That `grc.<module>` convention needs to be enforced in each of the modules because it makes each of them a child logger of the parent instance in main (`grc`). Because of how Python imports work the `logging.getLogger(__name__)` statement already handles this. I think I left that comment just for reference.

Adding individual logging levels per module is probably overkill, but I have been experimenting with additional levels to increase verbosity (DEBUG, DEBUG1, DEBUG2, ...). That might be something interesting.

mbr0wn:
There's some good ideas in the conversation, but I don't see why we can't merge this and continue the discussion elsewhere. If this PR doesn't get any attention, it'll anyway bitrot so I'm going to hit that merge button.

Issue #3398, opened 2020-04-21
GRC: handling of certain values within variables


mrjacobagilbert:
From a blank flowgraph, changing a variable to `pmt.init_u8vector(1,[203])` will cause the following:

```
UnicodeEncodeError: 'utf-8' codec can't encode character '\udccb' in position 30: surrogates not allowed
Traceback (most recent call last):
  File "/opt/gnuradio/lib/python3/dist-packages/gnuradio/grc/gui/ParamWidgets.py", line 124, in _apply_change
    self._update_gui()
  File "/opt/gnuradio/lib/python3/dist-packages/gnuradio/grc/gui/ParamWidgets.py", line 94, in _update_gui
    self.set_tooltip_text(self.param.format_tooltip_text())
  File "/opt/gnuradio/lib/python3/dist-packages/gnuradio/grc/gui/ParamWidgets.py", line 156, in set_tooltip_text
    self._input.set_tooltip_text(text)
UnicodeEncodeError: 'utf-8' codec can't encode character '\udccb' in position 30: surrogates not allowed
```

Also, PMT is not a default import, and it probably should be...

mrjacobagilbert:
This is on both master and maint-3.8.

ThomasHabets:
This seems to me like a bug in Python code. With Python3 it should use type `bytes` instead of `str`, right?

PR #3770, opened 2020-09-19
grc: ensure that strings are valid utf8 when evaluating parameters

daniestevez:
Some objects, notably PMTs can produce non-UTF-8 strings when their `__str__()` method is called. For instance
```
>>> str(pmt.init_u8vector(1, [203]))
'#[\udccb]'
```

These kinds of strings eventually give a `UnicodeEncodeError`, for instance when calling `print()` or displaying them with GTK. Python3 strings are supposed to be valid UTF-8.

When these objects are used inside block parameters in GRC, it will produce several tracebacks and hang up. This commit fixes the problem by sanitizing the strings to convert them to valid UTF-8. The sanitization uses the `'backslashreplace' `error response from Python.

Issue #3398 is caused by this problem.

---

I also may add that issue #3398 is important rather than a curiosity. It is common to define test packet as u8vector PMTs (in a message strobe, for instance). If these packets contain non-ASCII characters, then GRC will hang.

As a final remark, the underlying problem is that the `__str()__` method of a PMT can return a string which is not valid UTF-8. This is quite dangerous in Python 3. It might be good to take a look at this underlying problem.

### DIFF
--- a/grc/gui/VariableEditor.py
+++ b/grc/gui/VariableEditor.py
@@ -183,6 +183,9 @@ class VariableEditor(Gtk.VBox):
                     # Evaluate the params
                     for key in block.params :
                         evaluated = str(block.params[key].evaluate())
+                        # ensure that evaluated is a UTF-8 string
+                        # PMTs can produce non-UTF-8 strings
+                        evaluated = evaluated.encode('utf-8', 'backslashreplace').decode('utf-8')
                         self.set_tooltip_text(evaluated)
 
                     # Evaluate the block value
--- a/grc/gui/canvas/param.py
+++ b/grc/gui/canvas/param.py
@@ -73,6 +73,9 @@ class Param(CoreParam):
             if hasattr(value,"__len__"):
                 tooltip_lines.append('Length: {}'.format(len(value)))
             value = str(value)
+            # ensure that value is a UTF-8 string
+            # PMTs can produce non-UTF-8 strings
+            value = value.encode('utf-8', 'backslashreplace').decode('utf-8')
             if len(value) > 100:
                 value = '{}...{}'.format(value[:50], value[-50:])
             tooltip_lines.append('Value: ' + value)
@@ -139,6 +142,9 @@ class Param(CoreParam):
         else:
             # Other types
             dt_str = str(e)
+            # ensure that value is a UTF-8 string
+            # PMTs can produce non-UTF-8 strings
+            dt_str = dt_str.encode('utf-8', 'backslashreplace').decode('utf-8')
 
         # Done
         return _truncate(dt_str, truncate)


skoslowski:
There is a helper function in
https://github.com/gnuradio/gnuradio/blob/c78198d72a5ae16c596c7bb9aa7768e43d6f1f94/grc/gui/Utils.py#L97
that could/should be (adapted and) used.

daniestevez:
Thanks! I didn't know of that `encode()` helper function. I see that in Python2 it does a similar trick as I used (note however the order of `.encode` and `.decode` is not the same), while for Python3 is a passthrough. It also ends up calling `GLib.markup_escape_text()`, which probably we don't want here.

I'm new to the grc source, so I don't know why the `encode()` helper function is needed in the first place or what is the type of its argument (I guess this is quite important in Python 3, since you can only do `.encode()` on `bytes` and `.decode()` on `str`; Python 2 is more lax in this respect).

I'm not sure if it would be better to make a smarter function that handles both use cases or add another helper function (the three lines I've added do the same trick, so it is perfectly reasonable to do it in a function, and maybe something slightly different is needed for Python2, in case we want to backport this to GR 3.8).

Suggestions are welcome to adapt the PR.

mrjacobagilbert:
@daniestevez I agree this is not a trivial or niche problem; thank you for taking a look at fixing this.

The latest code you have still fails for me when I have, for example, an `import pmt` block and a variable block with the definition `pmt.init_u8vector(1,[166])`:

```
/usr/lib/python3/dist-packages/apport/report.py:13: DeprecationWarning: the imp module is deprecated in favour of importlib; see the module's documentation for alternative uses
  import fnmatch, glob, traceback, errno, sys, atexit, locale, imp
  File "/opt/gnuradio-master/lib/python3/dist-packages/gnuradio/grc/gui/Application.py", line 550, in _handle_action
    flow_graph_update()
  File "/opt/gnuradio-master/lib/python3/dist-packages/gnuradio/grc/gui/Application.py", line 121, in flow_graph_update
    fg.update()
  File "/opt/gnuradio-master/lib/python3/dist-packages/gnuradio/grc/gui/canvas/flowgraph.py", line 193, in update
    self.create_labels()
  File "/opt/gnuradio-master/lib/python3/dist-packages/gnuradio/grc/gui/canvas/flowgraph.py", line 483, in create_labels
    element.create_labels(cr)
  File "/opt/gnuradio-master/lib/python3/dist-packages/gnuradio/grc/gui/canvas/block.py", line 161, in create_labels
    markups = [param.format_block_surface_markup()
  File "/opt/gnuradio-master/lib/python3/dist-packages/gnuradio/grc/gui/canvas/block.py", line 161, in <listcomp>
    markups = [param.format_block_surface_markup()
  File "/opt/gnuradio-master/lib/python3/dist-packages/gnuradio/grc/gui/canvas/param.py", line 157, in format_block_surface_markup
    label=Utils.encode(self.name), value=Utils.encode(self.pretty_print().replace('\n', ' '))
  File "/opt/gnuradio-master/lib/python3/dist-packages/gnuradio/grc/gui/canvas/param.py", line 142, in pretty_print
    dt_str = str(e)
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xa6 in position 2: invalid start byte
  File "/opt/gnuradio-master/lib/python3/dist-packages/gnuradio/grc/gui/VariableEditor.py", line 185, in set_value
    evaluated = str(block.params[key].evaluate())
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xa6 in position 2: invalid start byte
  File "/opt/gnuradio-master/lib/python3/dist-packages/gnuradio/grc/gui/VariableEditor.py", line 185, in set_value
    evaluated = str(block.params[key].evaluate())
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xa6 in position 2: invalid start byte
  File "/opt/gnuradio-master/lib/python3/dist-packages/gnuradio/grc/gui/VariableEditor.py", line 185, in set_value
    evaluated = str(block.params[key].evaluate())
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xa6 in position 2: invalid start byte
```

And then this repeats for all of the blocks:

```
  File "/opt/gnuradio-master/lib/python3/dist-packages/gnuradio/grc/gui/DrawingArea.py", line 204, in draw
    self._flow_graph.draw(cr)
  File "/opt/gnuradio-master/lib/python3/dist-packages/gnuradio/grc/gui/canvas/flowgraph.py", line 513, in draw
    draw_element(cr)
  File "/opt/gnuradio-master/lib/python3/dist-packages/gnuradio/grc/gui/canvas/block.py", line 254, in draw
    port.draw(cr)
  File "/opt/gnuradio-master/lib/python3/dist-packages/gnuradio/grc/gui/canvas/port.py", line 131, in draw
    cr.rectangle(*self._area)
```

daniestevez:
Stupid me was working with 3.8 and then ported the changes over to 3.9 and figured out it should work. Of course it doesn't.

The problem seems to boil down to a difference between SWIG PMTs and pybind11 PMTs. Say you have
```
a = pmt.init_u8vector(1,[166])
```
as in your example. Then with SWIG calling `str(a)` will give us a `str` object that nevertheless doesn't contain valid UTF-8, while with pybind11 calling `str(a)` will give give us a `UnicodeDecodeError`.

Thus, the change I proposed would work as a workaround for 3.8, but for 3.9 I'm not sure what to do. I can't find a way to get a string representation for the PMT object (even one that would need to be fixed afterwards). Maybe this problem needs to be fixed in the pybind11 bindings.

One workaround for 3.9 would be to catch the `UnicodeDecodeError` and replace the string by some message saying that the object can't be displayed. But I would rather try to show the object replacing non-ASCII data by escape sequences, as my workaround for 3.8 does.

daniestevez:
Looking into this further, for me the problem is that `pmt::write()` in
https://github.com/gnuradio/gnuradio/blob/c78198d72a5ae16c596c7bb9aa7768e43d6f1f94/gnuradio-runtime/lib/pmt/pmt_io.cc#L41
will write non-ASCII characters.This is the underlying function that gets called by the `__str__` pybind11 method.

My question is, do we really want `pmt::write()` to generate non-ASCII characters? From the documentation it's not so clear what `pmt::write()` should do or what is it used for. I guess it's only for display purposes, since it lives in `pmt_io.cc`, while there are seriaization function in `pmt_serialize.cc`. There's also `pmt::read()`, which presumably should be able to parse PMTs written by `pmt::write()`, but `pmt::read()` is currently not implemented.

From the code I've seen, my proposal would be to modify `pmt_u8vector::string_ref()` in
https://github.com/gnuradio/gnuradio/blob/c78198d72a5ae16c596c7bb9aa7768e43d6f1f94/gnuradio-runtime/lib/pmt/pmt_unv.cc#L139
to produce escaped strings for non-ASCII characters. However I lack the larger picture, so something else might be more appropriate.

daniestevez:
Maybe this should have the "Don't merge" tag, since it seems that it's better to fix the underlying problem in the pmt library.

mormj:
@daniestevez - I don't think it is an unnecessary fix, but I just added an issue as you guys described above to track the underlying problem: #3778

daniestevez:
@mormj thanks for adding the issue. I think for 3.9 this needs to be solved at the pmt or pybind11 level, not in grc.

The patch in this PR should work well for 3.8, though. Do we want to fix this in the main3.8 branch?

mormj:
@daniestevez - let's close this for 3.9, and open up a separate one for 3.8 if you think that case needs to be addressed.

mormj:
I think I've addressed the specific string conversion issue in #3916 - it seems pybind11 chokes on non-uft8 strings, and swig let them through.  But the uniform vectors should not be producing those invalid strings, and hopefully that PR fixes this.

Still not a bad thing to scrub them in GRC, so I think this PR is valid and should be backported to 3.8 as well.

mbr0wn:
I don't think the comment in the PR is still valid (I couldn't reproduce the problem any more) but seems like there is no reason not to include this, just for extra protection.

mbr0wn:
Merged manually.

PR #3771, opened 2020-09-19
modtool: CMake template should allow building OOT modules without python if desired

mrjacobagilbert:
The gr_modtool utility should respect the `-DENABLE_PYTHON=OFF` setting to allow modules to be built without python support. This also does not install any GRC files even though it is possible to use GRC to generate flowgraphs without runtime python; however, this disabling python _and_ using GRC for C++ generation is a very advanced feature so the included note should be sufficient to inform anyone using in this way.

It should be possible to automatically install blocks that have `cpp_template`s defined at some point in the future should that become a necessary feature; for now even C++ only OOT's is a fairly niche need.

### DIFF
--- a/gr-utils/modtool/templates/gr-newmod/CMakeLists.txt
+++ b/gr-utils/modtool/templates/gr-newmod/CMakeLists.txt
@@ -152,8 +152,14 @@ add_subdirectory(include/howto)
 add_subdirectory(lib)
 add_subdirectory(apps)
 add_subdirectory(docs)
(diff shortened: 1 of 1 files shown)


mormj (approved):

mormj (approved):
I think this is a good idea to set up a more usable modtool template.  One **very** small thing - can you make Python have proper caps?  

Tested this and it works, keeping the usual functionality if ENABLE_PYTHON is not set (defaults to true)

PR #3776, opened 2020-09-24
gr-blocks: Fix wavfile sink and source block performance issues.

drmpeg:
Calling libsndfile sf_read/write_float() for every sample created
too much overhead. sf_read/write_float() now called every 1024 samples.

### DIFF
--- a/gr-blocks/lib/wavfile_sink_impl.cc
+++ b/gr-blocks/lib/wavfile_sink_impl.cc
@@ -49,8 +49,9 @@ wavfile_sink_impl::wavfile_sink_impl(const char* filename,
 {
     int bits_per_sample;
 
-    if (n_channels > 24) {
-        throw std::runtime_error("Number of channels greater than 24 not supported.");
+    if (n_channels > s_max_channels) {
+        throw std::runtime_error("Number of channels greater than " +
+                                 std::to_string(s_max_channels) + " not supported.");
     }
 
     d_h.sample_rate = sample_rate;
@@ -85,6 +86,9 @@ wavfile_sink_impl::wavfile_sink_impl(const char* filename,
     }
     set_bits_per_sample_unlocked(bits_per_sample);
     d_h.bytes_per_sample = d_bytes_per_sample_new;
+
+    set_max_noutput_items(s_items_size);
+    d_buffer.resize(s_items_size * d_h.nchans);
 
     if (!open(filename)) {
         throw std::runtime_error("Can't open WAV file.");
@@ -243,7 +247,6 @@ int wavfile_sink_impl::work(int noutput_items,
     int n_in_chans = input_items.size();
     int nwritten;
     int errnum;
-    float sample[24];
 
     gr::thread::scoped_lock guard(d_mutex); // hold mutex for duration of this block
     do_update();                            // update: d_fp is read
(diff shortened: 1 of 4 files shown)


drmpeg:
@ThomasHabets Are you happy with the memory management (std::vector and resize)? I'm not exactly a C++ expert.

gr-blocks/lib/wavfile_sink_impl.h
```diff
@@ -15,6 +15,9 @@
 #include <gnuradio/blocks/wavfile_sink.h>
 #include <sndfile.h> // for SNDFILE
 
+#define ITEMS_SIZE 1024
```
ThomasHabets, line 18: Make these member variable `constexpr`.

gr-blocks/lib/wavfile_source_impl.h
```diff
@@ -15,6 +15,9 @@
 #include <gnuradio/blocks/wavfile_source.h>
 #include <sndfile.h> // for SNDFILE
 
+#define ITEMS_SIZE 1024
```
ThomasHabets, line 18: If these are by definition the same value as in sink, maybe find a single place for to specify them?

Or are these abviously forever the same value always for wav files per spec?

gr-blocks/lib/wavfile_sink_impl.cc
```diff
@@ -86,6 +86,9 @@ wavfile_sink_impl::wavfile_sink_impl(const char* filename,
     set_bits_per_sample_unlocked(bits_per_sample);
     d_h.bytes_per_sample = d_bytes_per_sample_new;
 
+    set_max_noutput_items(ITEMS_SIZE);
+    sample.resize(ITEMS_SIZE * d_h.nchans);
```
ThomasHabets, line 90: slight optimization is to init `sample` in initializer list as `d_sample(n_channels)`.

Won't affect runtime performance, only construction. But in my opinion makes it clearer for the reader to see the size.

gr-blocks/lib/wavfile_sink_impl.h
```diff
@@ -29,6 +32,7 @@ class wavfile_sink_impl : public wavfile_sink
     float d_min_sample_val;
     float d_normalize_shift;
     float d_normalize_fac;
+    std::vector<float> sample;
```
ThomasHabets, line 35: Member variables should have prefix `d_`, so `d_sample`.

This is also now more of a buffer than "a sample", so `d_buf`, or `d_scratch` or something maybe?

gr-blocks/lib/wavfile_sink_impl.cc
```diff
@@ -49,7 +49,7 @@ wavfile_sink_impl::wavfile_sink_impl(const char* filename,
 {
     int bits_per_sample;
 
-    if (n_channels > 24) {
+    if (n_channels > MAX_CHANNELS) {
         throw std::runtime_error("Number of channels greater than 24 not supported.");
```
ThomasHabets, line 53: If you're moving out the contants then this error message should use it too. To avoid adding boost stuff I'd use ` + std::to_string(s_max_channels) +`

ThomasHabets (changes requested):
Yeah memory management looks good to me. I don't know if there's anything VOLK that can be used in the inner loop (likely not), so `std::vector` like this would be the way.

Some other comments though.

gr-blocks/lib/wavfile_sink_impl.h
drmpeg: Do you prefer ```constexpr``` or ```static constexpr```? I'm seeing both in the tree.

gr-blocks/lib/wavfile_sink_impl.cc
drmpeg: Did you mean ```d_sample(ITEMS_SIZE * n_channels)```?

gr-blocks/lib/wavfile_source_impl.h
drmpeg: Definitely could be different. In fact, I'm going to change the sink to 8192 to replicate the previous behavior.

gr-blocks/lib/wavfile_sink_impl.h
ThomasHabets: `static constexpr` is probably best. (static members should have `s_` prefix).

I believe the only difference here is that non-static would possibly waste a byte of memory.

gr-blocks/lib/wavfile_sink_impl.cc
ThomasHabets: Yes, sorry.

drmpeg: I'm going to stick with ```resize()```, just to keep it the same in both files (n_channels is not available in the source block).

ThomasHabets (approved):

ThomasHabets:
Looks good.

mormj (approved):

PR #3780, opened 2020-09-25
cmake: Remove Policy CMP0075 CMake warning.

drmpeg:


drmpeg:
Fixes second issue in https://github.com/gnuradio/gnuradio/issues/3349

drmpeg:
Fix taken from here:
https://github.com/gnuradio/gnuradio/blob/master/gnuradio-runtime/ConfigChecks.cmake#L92-L100

noc0lour (approved):
I'm not :100: sure what this is doing in CMake, but seems to be the right thing.

PR #3783, opened 2020-09-29
modtool: leave out bindings for python only modules

### LINKED ISSUE #3782: make fails when module is Python only
On the latest master build, make fails with the following error when trying to build Python-only OOT modules

[ 42%] Built target pygen_python_6b7f4
make[2]: *** No rule to make target '../python/bindings/gnuradio-dragon', needed by 'python/bindings/extracted_docstrings.json'.  Stop.
make[1]: *** [CMakeFiles/Makefile2:350: python/bindings/CMakeFiles/extracted_docstrings.dir/all] Error 2
make: *** [Makefile:141: all] Error 2

This error persists until a C++ block is added to the module. It seems like cmake doesn't have a way to handle not having any pybind11 bindings.

mormj:
Python-only modules were not skipping out on the binding compilation causing build errors.  This copies the breakout code previously from the swig directory

Fixes #3782

### DIFF
--- a/gr-utils/modtool/templates/gr-newmod/python/bindings/CMakeLists.txt
+++ b/gr-utils/modtool/templates/gr-newmod/python/bindings/CMakeLists.txt
@@ -5,6 +5,17 @@
 # SPDX-License-Identifier: GPL-3.0-or-later
 #
 
+########################################################################
+# Check if there is C++ code at all
+########################################################################
+if(NOT howto_sources)
+    MESSAGE(STATUS "No C++ sources... skipping python bindings")
+    return()
+endif(NOT howto_sources)
+
+########################################################################
+# Check for pygccxml
+########################################################################
 GR_PYTHON_CHECK_MODULE_RAW(
     "pygccxml"
     "import pygccxml"
--- a/gr-utils/modtool/templates/gr-newmod/python/__init__.py
+++ b/gr-utils/modtool/templates/gr-newmod/python/__init__.py
@@ -14,11 +14,10 @@ import os
 
 # import pybind11 generated symbols into the howto namespace
 try:
+    # this might fail if the module is python-only
     from .howto_python import *
 except ImportError:
-    dirname, filename = os.path.split(os.path.abspath(__file__))
-    __path__.append(os.path.join(dirname, "bindings"))
-    from .howto_python import *
+    pass
 
 # import any pure python here
 #


drmpeg (changes requested):
As a test case, I'm using gr-paint which has a Python module.

In the installed file <install_dir>/lib/python3/dist-packages/paint/__init__.py, the following code is inserted.
```
# import pybind11 generated symbols into the paint namespace
try:
    from .paint_python import *
except ImportError:
    dirname, filename = os.path.split(os.path.abspath(__file__))
    __path__.append(os.path.join(dirname, "bindings"))
    from .paint_python import *
```
But that code doesn't exist, so the following error happens:
```
Executing: /usr/bin/python3.6 -u /home/re/xfer/image.py

  File "/opt/gnuradio-3.9.0git/lib/python3/dist-packages/paint/__init__.py", line 17, in <module>
    from .paint_python import *
ModuleNotFoundError: No module named 'paint.paint_python'

  File "/home/re/xfer/image.py", line 33, in <module>
    import paint
  File "/opt/gnuradio-3.9.0git/lib/python3/dist-packages/paint/__init__.py", line 21, in <module>
    from .paint_python import *
ModuleNotFoundError: No module named 'paint.paint_python'
```

gr-utils/modtool/templates/gr-newmod/python/bindings/CMakeLists.txt
```diff
@@ -5,6 +5,17 @@
 # SPDX-License-Identifier: GPL-3.0-or-later
 #
 
+########################################################################
+# Check if there is C++ code at all
+########################################################################
+if(NOT tmp_sources)
+    MESSAGE(STATUS "No C++ sources... skipping swig/")
```
drmpeg, line 12: Swig?

mormj: oops.  Bad copy/paste.  fixed now.

mormj:
So the python `__init__` in 3.8 has the following comment:

```
# import swig generated symbols into the howto namespace
try:
    # this might fail if the module is python-only
    from .howto_swig import *
except ImportError:
    pass
```

Reduced the import from the python bindings to that, added the comment - this could probably be eliminated in the case there are no c++ blocks with a little work, but this should match 3.8 behavior

drmpeg:
The fix to \_\_init\_\_.py works, but if I add a C++ module after, the pybind11 modules don't get compiled because the ```No C++ sources... skipping python bindings"``` path gets taken in ```gr-paint/python/bindings/CMakeLists.txt```.

mormj:
Ok, that was an even dumber copy paste (copied the template after howto had already been replaced in my local dir) :/  Should be working now.

drmpeg (approved):
Looks good. There is a small downside. If you have a linking error, you get the old ```AttributeError: module 'paint' has no attribute 'paint_bc'``` versus a more informative ```ImportError: /opt/gnuradio-3.9.0git/lib/libgnuradio-paint.so.1.0.0git: undefined symbol: _ZN2gr3fft11fft_complexC1Eibi```.

mormj:
Yeah, I think that issue existed in the swig bindings as well because of the `pass`.  Going to go ahead and merge this and add an issue that that masking should be resolved

PR #3339, opened 2020-04-10
replace stderr logging by calls to GR's logging facilties

marcusmueller:
Supersedes #3024.

This is a squashed attempt to make #3024 viable.

It incorporates gvanh's work, which sadly wasn't error-free and bisectable.

Atop of that, rework from my side, where inappropriate loggers were chosen.

*Attention:* This is not the end to logging conversion; this really focuses on stderr logging; printf debugging is still at large, but since output to stderr almost certainly is log-relevant, and I tried to keep the changeset manageable, let's do this first and overhaul the rest of logging later.

PR #3784, opened 2020-09-29
Fix Windows logging

japm48:
PR #3339 introduced some compilation errors in Windows. This should fix them.

/cc @marcusmueller

### DIFF
--- a/gnuradio-runtime/lib/vmcircbuf_createfilemapping.cc
+++ b/gnuradio-runtime/lib/vmcircbuf_createfilemapping.cc
@@ -46,7 +46,7 @@ static void werror(char* where, DWORD last_error)
                   buf,
                   sizeof(buf) / sizeof(TCHAR), // buffer size
                   NULL);
-    GR_LOG_ERROR(logger, boost::format("%s: Error %d: %s" % where % last_error % buf));
+    GR_LOG_ERROR(logger, boost::format("%s: Error %d: %s") % where % last_error % buf);
     return;
 }
 #endif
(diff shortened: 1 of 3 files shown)


mormj:
@japm48 - would you mind separating out the changes in runtime and audio into two separate commits, each prepended with the module in the commit message (e.g. `runtime: ...` and `audio: ...`).

japm48:
@mormj No problem, done!

ryanvolz (approved):
I used this PR in my [last conda-forge build](https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=219342&view=results) and it went well, so this looks good to me!

japm48:
> I used this PR in my [last conda-forge build](https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=219342&view=results) and it went well, so this looks good to me!

Nice work!
Can I ask you how did you get GTK (and GRC) working on Windows? Several months ago I managed to do that using msys2, but my build still needs to be "deSWINGified". And I also tried conan but didn't get very far.

ryanvolz:
> Can I ask you how did you get GTK (and GRC) working on Windows? Several months ago I managed to do that using msys2, but my build still needs to be "deSWINGified". And I also tried conan but didn't get very far.

Conda-forge has a good community, so it wasn't too bad. A lot of GTK's dependencies were already in conda-forge when I started looking at it, and then I was able to take a recipe that someone else had gotten 99% of the way there and push it over the finish line. It helped immensely that GTK added support for building with meson a little while ago. Because of that, I didn't have to go the msys2 route and could use the Visual Studio environment that conda-forge defaults to. With GTK available, building gnuradio with GRC was a piece of cake.

japm48:
Thanks @ryanvolz!

I must recognize that misread conda as conan (need some sleep), even when I had tried both.

I also see you have patched your way into getting something that works. I hope I can publish my half-working version when/if I have time (possibly in a few weeks), as there may (or may not) be something useful for your version.

In my case, I managed to tear all the necessary runtime requirements into a separate self-standing zip file, mainly by taking advantage of the package system (pacman, the same as in Arch Linux).

Ok, I'll stop hijacking this PR thread.

PR #3791, opened 2020-10-01
vocoder: cmake: Use pybind macros to manage bindings.

ryanvolz:
This helps prevent current and future headaches by making sure improvements to the pybind11 CMake don't also need to be added specifically for gr-vocoder.

(I noticed that a change similar to #3788 would be needed for gr-vocoder as well, but this fixes it instead by not duplicating the CMake code that needs to be fixed.)

### DIFF
--- a/gr-vocoder/python/vocoder/bindings/CMakeLists.txt
+++ b/gr-vocoder/python/vocoder/bindings/CMakeLists.txt
@@ -2,7 +2,7 @@
 # Python Bindings
 ########################################################################
 
-pybind11_add_module(vocoder_python
+list(APPEND vocoder_python_files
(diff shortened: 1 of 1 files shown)


mormj (approved):

mormj:
Thanks for doing this - vocoder has so many conditional dependencies I kicked the can down the road on using the same macros as the other modules.  Much cleaner though!

mormj:
This actually did pass CI - there was just a disconnect at the time with github API:

image

drmpeg (approved):
Built locally and tested with FreeDV and codec2 examples. Nice cleanup. LGTM.

PR #3795, opened 2020-10-01
network: lib: Fixes to get network to compile with MSVC

ryanvolz:
This includes using functions from the standard library to sleep instead of `usleep`, and a weird error where MSVC barfed on the variable name `s_host`.


ryanvolz:
I don't know how performance-critical the `usleep` calls are, so there might be ramifications from replacing them. It would be good to have someone who knows better comment on that, or suggest a better cross-platform alternative. This was just the first thing that I found after searching, and it compiled.

mbr0wn (approved):

PR #3800, opened 2020-10-03
grc: update the "Get Involved" dialog

argilo:
The Help -> Get Involved dialog in GRC is out of date: the "Get Involved" link is broken, and it still references Slack. I've refreshed it here.

Before:

Screenshot from 2020-10-03 14-42-34

After:

Screenshot from 2020-10-03 14-41-59

### DIFF
--- a/grc/gui/Dialogs.py
+++ b/grc/gui/Dialogs.py
@@ -338,11 +338,11 @@ def show_keyboard_shortcuts(parent):
 def show_get_involved(parent):
     """Get Involved Instructions"""
     markup = textwrap.dedent("""\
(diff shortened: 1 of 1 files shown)


mbr0wn (approved):

mormj:
Looks good - also, would be nice to have this one backported to 3.8 as well.

argilo:
@mormj This issue doesn't exist in 3.8 because the dialog was newly added in 3.9.

Or did you mean that the dialog itself should be added to 3.8?

mormj:
Nope - my bad.  3.9 only it is then!  (I had forgotten this was entirely new)

PR #3806, opened 2020-10-05
Don't install some examples if python is disabled

doronbehar:


doronbehar:
On NixOS, our ecosystem allows packages to be overrided, via Nix' expressions. We'd like to package gnuradio with examples, but we don't want examples that need python installed if the build is disabling python. Without this patch (or more precisely, a version of it suitable for v3.8.2.0) disabling python doesn't allow the build to not reference python.

mormj (approved):
Thank you for the fix - indeed this is the only one where the examples lives outside if(python) for some reason

drmpeg (commented):
This is fine. But a couple of notes:

1) Examples are still being installed for gr-fec.

2) Did the author really want this on maint-3.8?

doronbehar:
>     2. Did the author really want this on maint-3.8 instead?

It seems to me the PR targets `master`:

ZENIX-selection-2020-10-12T17:07:56+03:00


>     1. Examples are still being installed for gr-fec.

It seems to me that gr-fec's examples are not python files, on current master:

https://github.com/gnuradio/gnuradio/blob/4b402404e152ede29cfaa5c0648bdfb5d327038b/gr-fec/CMakeLists.txt#L45-L47

And: https://github.com/gnuradio/gnuradio/tree/4b402404e152ede29cfaa5c0648bdfb5d327038b/gr-fec/examples

Hence they don't reference python at all, at least in the sense of [what NixOS considers "references"](https://nixos.org/guides/nix-pills/automatic-runtime-dependencies.html#idm140737320201584).

drmpeg:
.grc files can't run without Python.

In your original comment, you said "Without this patch (or more precisely, a version of it suitable for v3.8.2.0) disabling python doesn't allow the build to not reference python."

Are you packaging GNU Radio 3.8.2.0 (branch maint-3.8) or GNU Radio 3.9.0.0-git (master branch)?

doronbehar:
> Are you packaging GNU Radio 3.8.2.0

Yes, the latest 3.8.2.0 tag.

> .grc files can't run without Python.

I see, and indeed iterating all occourances of `add_subdirectory(examples` shows that only fec's cmakelists.txt didn't implement this understanding. Pushing another commit for that now.

drmpeg:
Thanks for the additional commit. As for versions and branches, here's some information.

1) The master branch (where this PR will be merged into) is a development branch and is unofficially called GNU Radio 3.9. It is not compatible with GNU Radio 3.8 (that is, out of tree modules for GNU Radio 3.8 will not build against the current head of this branch). Right now, I'm not aware of any plans for a release, so it could be many months before your change becomes part of any release or tag.

2) The maint-3.8 branch is where GNU Radio 3.8 is maintained. The latest tag/release on that branch is v3.8.2.0. Right now, there are 8 new commits after the v3.8.2.0 tag. At some point, there will probably be a v3.8.3.0 tag/release and most likely it will happen before GNU Radio 3.9 is released. So if you want to see your change in a tag/release, it would be better if it was submitted to the maint-3.8 branch.

However, this PR certainly makes sense for master, so don't change it. But you may want to consider an additional PR (with the same changes) for the maint-3.8 branch. In other words, a "backport".

Making a PR for the maint-3.8 branch takes a couple of additional steps. After you clone your repo:

git checkout maint-3.8

Then make a branch, make your changes, commit, and push as usual.

Then on Github, you have to change the destination branch (at the top of the pull request page) to maint-3.8.

Hope this makes sense.

doronbehar:
Thanks I am familiar with the git workflow, thanks :+1:, will do it shortly.

PR #3808, opened 2020-10-06
runtime: remove six

### LINKED ISSUE #3446: Remove dependency on Python module `six`

argilo:
Partially resolves #3446.

This change removes the last usage of six from the runtime component.


mormj (approved):
Awesome!  Seems like a straightforward fix

willcode (approved):

PR #3810, opened 2020-10-06
audio: fix lcm calculation

argilo:
While working on #3809, I noticed that `gnuradio.digital.packet_utils.make_packet` is broken because `gnuradio.gru.lcm` returns a float instead of an integer. I've fixed that here, and `make_packet` now executes.

I also checked where `lcm` is used, and found that test_resampler.py was also broken; the LCM calculations were broken in https://github.com/gnuradio/gnuradio/commit/9e625c4821f4c63421b3d3747c0c4f358fef6c5f#diff-da129366a03c0bb69563f36127facf1d. I've restored them to their previous form here, except using Python 3's integer division operator `//`. After this change, test_resampler.py works again.

### DIFF
--- a/gnuradio-runtime/python/gnuradio/gru/mathmisc.py
+++ b/gnuradio-runtime/python/gnuradio/gru/mathmisc.py
@@ -15,7 +15,7 @@ def gcd(a,b):
     return a
 
 def lcm(a,b):
-    return a * b / gcd(a, b)
+    return a * b // gcd(a, b)
 
 def log2(x):
     return math.log(x) / math.log(2)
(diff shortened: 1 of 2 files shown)


mormj:
@argilo - any way we can get test_resampler integrated into the CI with some minimal test case that will hit lcm?

argilo:
That might not be easy since test_resampler.py uses the sound card. It seems intended as an example.

There are a couple other things that use `lcm`, so it might be practical to test them, or just test `lcm` directly.

argilo:
Some other changes we could make to mathmisc.py next: `log2` is unused, so it could probably go. Python has had `math.log2` since 3.3. And `gcd` is used only by rational_resampler.py; Python has had `math.gcd` since 3.5 so we could switch to that. (GNU Radio now requires Python 3.6.)

mormj (approved):

willcode (approved):

PR #3818, opened 2020-10-07
gr-blocks: Add missing methods to Multiply Const block pybind11 binding.

drmpeg:


drmpeg:
Fixes bug reported on mailing list.

https://lists.gnu.org/archive/html/discuss-gnuradio/2020-10/msg00028.html

mormj (approved):

mormj:
Thanks for the fix @drmpeg.  Must have missed this one - the templated blocks had a lot of trouble with the automated tools, so they had to be populated manually, this being one that slipped through the cracks apparently.

ccera-astro:
The add_const_vxx appears to have slipped through, as well :(

PR #3823, opened 2020-10-08
cmake: Set VERSION_PATH to be only "git", fixing RC_MAINT_VERSION.

ryanvolz:
The version parsing in [`GrVersion.cmake`](https://github.com/gnuradio/gnuradio/blob/master/cmake/Modules/GrVersion.cmake#L51-L79) expects exactly the string "git" as either the `MINOR_VERSION` or `MAINT_VERSION` in order to set the dependent variables correctly. Without this fix, Windows builds will fail because `RC_MAINT_VERSION` contains an invalid string.


marcusmueller:
Sounds sane to me, would like @maitbot's input on this, to be honest. If this is OK to do without breaking debian, let's do it for 3.9

mbr0wn (approved):

PR #3824, opened 2020-10-08
grc: restore pre-3.8 block id behavior

mrjacobagilbert:
Prior to 3.8 block IDs were shown in the properties dialogue for every block as these are important for function of some GR features. #2795 brought this back as an option for 3.8+, however it forces the `hide` value to effectively be `none` instead of `part` (not visible on the block canvas but visible in the properties dialogue) for all blocks. This changes this back to the pre-3.8 behavior where the ID value is not shown on the canvas but is shown in the properties dialogue.

I realize this was a deliberate change by @mormj but most block IDs add little value on the canvas and the exceptions to this are explicit in their yaml files (e.g.: variables). I also realize that the change to the PropsDialog.py does not actually change anything, but it matches the intention.

### DIFF
--- a/grc/gui/PropsDialog.py
+++ b/grc/gui/PropsDialog.py
@@ -171,7 +171,7 @@ class PropsDialog(Gtk.Dialog):
 
                 for param in self._block.params.values():
                     if force_show_id and param.dtype == 'id':
-                        param.hide = 'none'
+                        param.hide = 'part'
                     # todo: why do we even rebuild instead of really hiding params?
                     if param.category != category or param.hide == 'all':
                         continue
--- a/grc/gui/canvas/block.py
+++ b/grc/gui/canvas/block.py
@@ -157,8 +157,8 @@ class Block(CoreBlock, Drawable):
 
         # update the params layout
         if not self.is_dummy_block:
-            markups = [param.format_block_surface_markup()
-                       for param in self.params.values() if (param.hide not in ('all', 'part') or (param.dtype == 'id' and force_show_id))]
+            markups = [param.format_block_surface_markup() for param in self.params.values()
+                       if (param.hide not in ('all', 'part'))]
         else:
             markups = ['<span font_desc="{font}"><b>key: </b>{key}</span>'.format(font=PARAM_FONT, key=self.key)]
 


mormj:
I'm not opposed to this at all - visually it is better to still have access to all the block ids, but for them not to pollute the canvas

The only reason for this option is when blocks need to be used in probes or python snippets, or for some reason the show_id attribute was forgotten.  

Tested it and it looks good.

mormj:
I'm not opposed to this at all - visually it is better to still have access to all the block ids, but for them not to pollute the canvas

The only reason for this option is when blocks need to be used in probes or python snippets, or for some reason the show_id attribute was forgotten.  

Tested it and it looks good.

mormj (approved):

grc/gui/PropsDialog.py
```diff
@@ -171,7 +171,7 @@ def update_gui(self, widget=None, force=False):
 
                 for param in self._block.params.values():
                     if force_show_id and param.dtype == 'id':
-                        param.hide = 'none'
+                        param.hide = 'part'
```
skoslowski, line 174: This (permanently) modifies the param object (as does the current version). I find this highly troublesome.

grc/gui/canvas/block.py
```diff
@@ -158,7 +158,7 @@ def create_labels(self, cr=None):
         # update the params layout
         if not self.is_dummy_block:
             markups = [param.format_block_surface_markup()
-                for param in self.params.values() if (param.hide not in ('all', 'part') or (param.dtype == 'id' and force_show_id))]
+                for param in self.params.values() if (param.hide not in ('all', 'part'))]
```
skoslowski, line 161: I like this change, as it remove logic in the presentation layer which could have easily been handled via `param.hide` in the first place. Note, that you'll probably also what to remove `force_show_id = ...` a few lines above, as it is no longer needed.

Nitpick: if this line is touched it deserves to get proper indentation...

skoslowski (changes requested):
I like what you're try to do here. It the "how" that bothers me a little...

grc/gui/PropsDialog.py
mrjacobagilbert: I had thought this initially as well, but I could not find any actual issue; selecting the toggle overrides the PropsDialog ID behavior to part/none and deselecting uses the yaml behavior. My change to 'part' here actually does nothing as none/part have the same behavior for the PropsDialog.

grc/gui/canvas/block.py
mrjacobagilbert: Removed and I did what I think is proper indentation (basically every python beautifier disagreed with this and each other though...if youd like a change can you be more specific?).

mrjacobagilbert:
@mormj Rebased.

@skoslowsk have you looked over this since your initial feedback? I don't think think that modifying the object is an issue here because it appears to be reloaded when `Show All Block IDs` is toggled.

PR #3825, opened 2020-10-09
Backport 3.8: uhd: Update possible master_clock_rate

mormj:
Backport of #3667 from @jdemel - any reason for this to not go on maint-3.8?

A USRP block offers the option to set a `clock_rate` which in turn sets
`master_clock_rate` in UHD. There were some common options
(153.6MHz, 125MHz, 122.88MHz) missing for N310s. Now these options are added.

Instead of a separate call to `set_clock_rate`, `master_clock_rate` is
now set as part of the device address string. This behavior makes it
more consistent with UHD examples. Flowgraphs that set MCR as part of
their `device_args` option, are currently overruled if clock rate is set
too. This behavior may easily be reversed.

According to the UHD manual, X300s don't support
`set_master_clock_rate`, thus in this case it is strictly necessary to
set MCR as part of the device address string.

### DIFF
--- a/gr-uhd/grc/gen_uhd_usrp_blocks.py
+++ b/gr-uhd/grc/gen_uhd_usrp_blocks.py
@@ -71,8 +71,8 @@ parameters:
     label: Clock Rate (Hz)
     dtype: real
     default: 0e0
-    options: [0e0, 200e6, 184.32e6, 120e6, 30.72e6]
-    option_labels: [Default, 200 MHz, 184.32 MHz, 120 MHz, 30.72 MHz]
+    options: [0e0, 200e6, 184.32e6, 153.6e6, 125.0e6, 122.88e6, 120e6, 30.72e6]
+    option_labels: [Default, 200 MHz, 184.32 MHz, 153.6 MHz, 125 MHz, 122.88 MHz, 120 MHz, 30.72 MHz]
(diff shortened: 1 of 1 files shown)


drmpeg (commented):
The diff looks different than the original PR. In the original PR, there was an additional deletion starting at line 147 of the old file.

https://github.com/gnuradio/gnuradio/commit/f25a97122d74ce8217efcdec807b4c338627ec06#diff-6883ea3a476840439f9d3669e4aee06fL147

jdemel (approved):
LGTM. @drmpeg comment on line 147ff I assume these lines are 138ff here.

drmpeg:
@jdemel Umm, no. It would be line 207-209 in the 3.8 file. This is where the old method of setting the clock rate should be deleted.
```
        ${'%'} if clock_rate():
        self.${'$'}{id}.set_clock_rate(${'$'}{clock_rate}, uhd.ALL_MBOARDS)
        ${'%'} endif
```

mormj:
Should be fixed now to match the cherry-pick (not sure why git cherry-pick didn't go clean there).  @mbr0wn - can you verify that it is ok to add the additional 2 clock rates to the master clock rate list?

mbr0wn (approved):

PR #3828, opened 2020-10-10
PMT: replace custom hash implementation with std::hash and increase hash table size

mrjacobagilbert:
Some of libpmt's features and behaviors have been a bit of a sore spot over the years; the hashtable being one of them. While replacing PMT with another serialization library is on the roadmap, we are stuck with them for now, so we might as well make them work for us.

Two annoying things I have noticed are the custom hash algorithm, and the small hash table size; I view both of these changes as suggestions for discussion. While this PR addresses both, they are separate changes and can be split up if necessary.

---

The hashing algorithm has been updated to use `std::hash<std::string>` which is very slightly faster (~3% on my ubuntu system), but more importantly not a homebrew undocumented algorithm. I did verify uniformity of the output of both algorithms:

image

image

The hash table size is a more convoluted and https://github.com/gnuradio/gnuradio/pull/1448 has some good information about this. The increased memory use seems negligible given GR's memory footprint, and improved my test code (`pmt::eq()` comparison against 15k unique random strings) by a factor of ~8.5x.

### DIFF
--- a/gnuradio-runtime/lib/pmt/pmt.cc
+++ b/gnuradio-runtime/lib/pmt/pmt.cc
@@ -18,6 +18,7 @@
 #include <pmt/pmt_pool.h>
 #include <stdio.h>
 #include <string.h>
+#include <functional>
 #include <vector>
 
 namespace pmt {
@@ -150,27 +151,11 @@ static std::vector<pmt_t>* get_symbol_hash_table()
 pmt_symbol::pmt_symbol(const std::string& name) : d_name(name) {}
 
 
-static unsigned int hash_string(const std::string& s)
-{
-    unsigned int h = 0;
-    unsigned int g = 0;
-
-    for (std::string::const_iterator p = s.begin(); p != s.end(); ++p) {
-        h = (h << 4) + (*p & 0xff);
-        g = h & 0xf0000000;
-        if (g) {
-            h = h ^ (g >> 24);
-            h = h ^ g;
-        }
-    }
-    return h;
-}
-
 bool is_symbol(const pmt_t& obj) { return obj->is_symbol(); }
 
 pmt_t string_to_symbol(const std::string& name)
 {
-    unsigned hash = hash_string(name) % get_symbol_hash_table_size();
+    unsigned hash = std::hash<std::string>()(name) % get_symbol_hash_table_size();
 
     // Does a symbol with this name already exist?
     for (pmt_t sym = (*get_symbol_hash_table())[hash]; sym; sym = _symbol(sym)->next()) {
--- a/gnuradio-runtime/lib/pmt/pmt.cc
+++ b/gnuradio-runtime/lib/pmt/pmt.cc
@@ -138,7 +138,7 @@ bool to_bool(pmt_t val)
 
 static const unsigned int get_symbol_hash_table_size()
 {
-    static const unsigned int SYMBOL_HASH_TABLE_SIZE = 701;
+    static const unsigned int SYMBOL_HASH_TABLE_SIZE = 8192;
     return SYMBOL_HASH_TABLE_SIZE;
 }
 


argilo (approved):

gnuradio-runtime/lib/pmt/pmt.cc
```diff
@@ -137,7 +137,7 @@ bool to_bool(pmt_t val)
 
 static const unsigned int get_symbol_hash_table_size()
 {
-    static const unsigned int SYMBOL_HASH_TABLE_SIZE = 701;
```
mormj, line 140: does anyone know what the reasoning for 701 originally was.  seems like an oddly specific table size

argilo: There's speculation on that question here: https://github.com/gnuradio/gnuradio/pull/1448#issuecomment-326732363

jacobgilbert-snl: I am currently re-running the distribution tests with random words, and the word list here https://github.com/dwyl/english-words and will update with results.

jacobgilbert-snl: Results are aligned with @marcusmueller observations, and std::hash still does a much better job:

image

466k words were from the list above, the rest are random strings.

mrjacobagilbert: (Removed my other account's comments in favor of one consolidated comment)

I am re-ran the distribution tests with the word list here https://github.com/dwyl/english-words and a few random strings. The `hash_string()` results are well aligned with what @marcusmueller observed previously for the custom GR hash and a 2^n table size, but std::hash returns a very reasonable distribution:

image

It's probably more useful to view the hash value distribution directly (histogram of above):

image

Overall this looks as one would expect.

willcode (approved):
Since there is no standard algorithm for computing hashes, is there a way to make sure the distribution is good (even if different) on different platforms?

I think 701 was just a prime number of about the right size. Prime size is needed to help out where a sub-optimal hash algorithm has favorite multiples.

argilo:
> is there a way to make sure the distribution is good (even if different) on different platforms

The C++ specification does have a basic quality requirement:

> For two different parameters `k1` and `k2` that are not equal, the probability that `std::hash<Key>()(k1) == std::hash<Key>()(k2)` should be very small, approaching `1.0/std::numeric_limits<std::size_t>::max()`.

mrjacobagilbert:
@willcode Unfortunately the algorithm selection is not strictly defined. GCC uses (or used) MurmurHashUnaligned2 for example (https://stackoverflow.com/questions/19411742/what-is-the-default-hash-function-used-in-c-stdunordered-map). The requirement  @argilo mentions is the only mechanism to ensure hash quality I could find. Due to the modulo, table size should ideally be of size 2^n, though for any table size much less than the hash size the distribution should be acceptable.

I feel this is an acceptable trade to get rid of an undocumented algorithm (for which the table size appears to be an unexpected and essential part of).

gnuradio-runtime/lib/pmt/pmt.cc
```diff
@@ -150,27 +150,11 @@ static std::vector<pmt_t>* get_symbol_hash_table()
...
-    }
-    return h;
-}
-
 bool is_symbol(const pmt_t& obj) { return obj->is_symbol(); }
 
 pmt_t string_to_symbol(const std::string& name)
 {
-    unsigned hash = hash_string(name) % get_symbol_hash_table_size();
+    unsigned hash = std::hash<std::string>()(name) % get_symbol_hash_table_size();
```
marcusmueller, line 157: Include what you use: according to all I know, `std::hash` is in `<functional>`, but we're including that nowhere; this still works (and `g++ -M -c -I/path/to/includes pmt.cc` tells me that, yes, somehow <functional> is recursively included), but we should really do `#include <functional>` at the top of this file.

marcusmueller (commented):
Great work, @mrjacobagilbert. Thanks! We should have done this eons ago. I'll go ahead, add the necessary include, and send this patch on its way.

marcusmueller (approved):

gnuradio-runtime/lib/pmt/pmt.cc
jacobgilbert-snl: I had this originally but reading http://www.cplusplus.com/reference/functional/hash/ made me think that `std::hash<std::string>` lives in `<string>`. So I removed it and it still worked. I could be wrong.

PR #3832, opened 2020-10-11
vocoder: refactor get_next_tx_char callback

argilo:
Fixes #2928.
Partially resolves #2875.
Replaces #3443.

The get_next_tx_char callback in the FreeDV modulator block is one of 
the last places where snprintf is used. I've eliminated it by using C++ 
string methods instead. This also removes the 79-character limit.

This change also fixes a bug where the last character of the message was 
not transmitted.

Lastly, I refactored the callback to be a static member function, 
eliminating the need for a separate structure to hold the callback 
state.

PR #3833, opened 2020-10-11
vocoder: add text message output to FreeDV demodulator

### LINKED ISSUE #2929: vocoder: FreeDV add text message demod as message output
The FreeDV modulator supports a `Text Message` encoding, but the FreeDV demodulator does not decode it/output it.

argilo:
Fixes #2929.

The FreeDV protocol includes a low bit rate text stream which is used to 
transmit information such as call sign and location. GNU Radio's FreeDV 
demodulator block sets up a callback to receive this text data, but the 
callback currently does nothing.

Here I've added an optional message output to the block. It collects 
characters until a carriage return is encountered, then outputs a 
string. I've updated the sample flow graph to print these messages using 
the Message Debug block.

### DIFF
--- a/gr-vocoder/examples/grfreedv.grc
+++ b/gr-vocoder/examples/grfreedv.grc
@@ -1,6 +1,7 @@
 options:
   parameters:
     author: A. Maitland Bottoms
+    catch_exceptions: 'True'
     category: Custom
     cmake_opt: ''
     comment: ''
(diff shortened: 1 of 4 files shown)


argilo:
Note: The FreeDV blocks will not be built unless codec2 is available; on Ubuntu it can be installed with `sudo apt install libcodec2-dev`.

argilo:
Here's the updated sample flowgraph:


freedv-sample-new

argilo:
Output from the Message Debug block looks like so:
```
GNU Radio
GNU Radio
```

devnulling:
@argilo this looks great

Fixes: https://github.com/gnuradio/gnuradio/issues/2929

It should also resolve: https://github.com/gnuradio/gnuradio/issues/2928

devnulling (approved):

argilo:
> It should also resolve: #2928

That one is fixed in #3832. I'll add it to the description.

willcode (approved):

Issue #3759, opened 2020-09-09
gr_modtool bind fails if additional includes -- like qtwidgets -- are used


dl1ksv:
gr_modtool bind fails if I try to (re)bind an oot module that include some widgets of qt with a message like

fatal error: 'QWidget' file
      not found
#include \<QWidget>

The same problem arises if I try to build the oot module inside the cmake chain.
For this problem a proposed a pr #3629 ,though still not accepted yet.

I wonder how to pass these extra_includes list to gr_modtool.

Before I think about a fix I'd like to get some advice, where to provide this info.

dl1ksv:
Fixed with #3837

PR #3837, opened 2020-10-12
Adding optional parameter addl_includes to gr_modtool bind

dl1ksv:
gr_modtool bind fails if the module references external packages like qt.
In such a case you see something like:

INFO Parsing source file "/home/schroer/gnuradiocomponents/gr-display/include/display/display_text_msg.h" ... 
/home/schroer/gnuradiocomponents/gr-display/include/display/display_text_msg.h:33:10: fatal error: 
      'qapplication.h' file not found
#include <qapplication.h>
         ^~~~~~~~~~~~~~~~
1 error generated.
Error occurred while running CASTXML:  status:1

So I propose to add an optional parameter --addl_includes to gr_modtool bind to tell where to find the required includes.
This addresses #3759

### DIFF
--- a/gr-utils/modtool/core/bind.py
+++ b/gr-utils/modtool/core/bind.py
@@ -37,6 +37,7 @@ class ModToolGenBindings(ModTool):
     def __init__(self, blockname=None, **kwargs):
         ModTool.__init__(self, blockname, **kwargs)
         self.info['pattern'] = blockname
+        self.info['addl_includes'] = kwargs['addl_includes']
 
     def validate(self):
         """ Validates the arguments """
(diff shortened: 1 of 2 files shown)


mormj (approved):

mormj:
Tested this out with 2 random includes in different directories, and it seems to work well

gr-utils/modtool/core/bind.py
```diff
@@ -54,9 +55,10 @@ def run(self):
                 "Bindtool required to generate bindings  ... Aborting")
             return
 
+   
         with warnings.catch_warnings():
             warnings.filterwarnings("ignore", category=DeprecationWarning)
             file_to_process = os.path.join(self.dir, self.info['includedir'], self.info['blockname'] + '.h')
             bg = BindingGenerator(prefix=gr.prefix(), namespace=[
-                                  'gr', self.info['modname']], prefix_include_root=self.info['modname'], output_dir=os.path.join(self.dir, 'python'))
+                                  'gr', self.info['modname']], prefix_include_root=self.info['modname'], output_dir=os.path.join(self.dir, 'python'),addl_includes=self.info['addl_includes'])
```
swap-nil7, line 63: Please leave a space after `,`

swap-nil7 (approved):

gr-utils/modtool/core/bind.py
dl1ksv: Done, Thanks

Issue #2172, opened 2018-11-15
python: Start enforcing some PEP8 rules


btashton:
I would really like to see us start to follow some PEP8 rules on the code ideally through CI.
I have been using https://github.com/PyCQA/pycodestyle with a specific whitelist.  I think if we don't start adding this to the CI process the work that people have done to bring the style inline with the recommendations will get lost again.   To be clear I am not suggesting turning it all on and going crazy with whitespace, line length etc.

Kenneth Reitz wrote a really nice guide here about PEP8 https://pep8.org/

Here is what I would advocate turning on right away:

- Disallow wildcard imports (this has caused issues with things like numpy overwriting sum)
- Disallow bare exceptions (this has been hiding dead broken code)
- 4 spaces (no tabs, 3 spaces or anything else)

I would like to turn on trailing whitespace, but the code has a ton of that right now.

with pycodestyle we can include a config file in the source tree with the rules we want.

btashton:
With that context (I did not expect people to be open to a mass reformat) I would be happy to spearhead that on the python side.  I have never been super excited about auto formatters without a close look at the result, its hard for them to know what looks "good".   If we can define a list of PEP rules that we want (likely with some modifications around line length etc...) I'm happy to also write the config file.  `pycodestyle` is really good at _style_ checking, I could see adding pylint in for more static analysis (things like hey did you want to call __init__ on the inherited class)

PR #3841, opened 2020-10-12
grc: fix whitespace issues

argilo:
GRC's Python code has a lot of trailing whitespace, which makes it annoying to work with in a text editor that automatically removes trailing whitespace. I've corrected that here, and also corrected some other whitespace issues identified by Pylint.


mormj (approved):
These all look reasonable - we should probably add something to enforce python formatting like we do with clang-format

skoslowski:
I am all in favor of this. In fact, I'd suggest we take it one step further and just [go black](https://black.readthedocs.io/en/stable/) and then enforce that for future PRs.

mormj:
There is some discussion of PEP8 enforcement in #2172 - black looks like an interesting tool.  We can probably go ahead and merge this one.

willcode (approved):

skoslowski (approved):

PR #3843, opened 2020-10-12
runtime backport-3.8: fix latency issue caused by setting block alias on message blocks

mrjacobagilbert:
This is a direct backport of #3536; resolution of the 3.8 backport will in my mind address #3521. While this does affect 3.7 I am not sure it is a big enough issue to justify a 3.7-backport.

This changes how the blocks are identified when they need to be notified that they have messages waiting. It also clarifies the name the block is registered with explicitly, and adds exceptions for looking up blocks that do not exist.

### DIFF
--- a/gnuradio-runtime/lib/block_registry.cc
+++ b/gnuradio-runtime/lib/block_registry.cc
@@ -132,10 +132,21 @@ void block_registry::notify_blk(std::string blk)
     gr::thread::scoped_lock guard(d_mutex);
 
     if (primitive_map.find(blk) == primitive_map.end()) {
+        /* In later GNU Radio versions, we throw here:
+         *
+         * throw std::runtime_error("block notify failed: block not found!");
+         *
+         * But since this is an API change that we'd like to avoid on a maintenance
+         * branch, we're not doing that on maint-3.8.
+         */
         return;
     }
-    if (primitive_map[blk]->detail().get())
+    if (primitive_map[blk]->detail().get()) {
         primitive_map[blk]->detail()->d_tpb.notify_msg();
+    } else {
+        // not having block detail is not necessarily a problem; this will happen when
+        // publishing a message to a block that exists but has not yet been started
+    }
 }
 
 } /* namespace gr */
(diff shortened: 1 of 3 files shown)


gnuradio-runtime/lib/block_registry.cc
```diff
@@ -132,10 +132,14 @@ void block_registry::notify_blk(std::string blk)
     gr::thread::scoped_lock guard(d_mutex);
 
     if (primitive_map.find(blk) == primitive_map.end()) {
-        return;
+        throw std::runtime_error("block notify failed: block not found!");
```
jdemel, line 135: We might need to discuss if we want a behavioral change in 3.8.

jdemel (approved):
LGTM

bastibl (approved):

mrjacobagilbert:
@bastibl do you have an opinion on the suggestion by @jdemel to remove the exception for the backport?

bastibl:
Can you explain why you think that an exception is not a good idea here?

My impression was that this notification stuff is internal to the runtime. So if the runtime tries to notify a block that it didn't create, that sounds like something is fundamentally broken and an exception would be OK. Or am I missing a normal/acceptable case where this could happen?

bastibl:
I guess that's a question for @marcusmueller

marcusmueller:
Tough one!

I'd agree with @bastibl, there's something fundamentally broken when this exception would be raised, and hence, whether or not we're compatible doesn't matter much.

Then again, it's a not-inherently-100%-necessary API change, and I can't tell who's depended on that behaviour in the past. It does, sadly, not sound to absurd that someone, especially during flow graph reconfiguration, would change some parameters and then decide to trigger changes on blocks without wanting to check whether they do exist.

So, after quite some deliberation, I'll opt for: Johannes is right, we probably shouldn't be raising an exception here, where we just did nothing before. It's a very stable maintenance branch, after all.

marcusmueller:
I've fixed that, now waiting for CI to complete.

marcusmueller (approved):

Issue #1814, opened 2018-06-08
gr-uhd: tx/rx tune commands can be dropped if same frequency


jacobagilbert:
Issue #1422 caused am edge case where a tx tune followed by an RX tune to the same frequency (or vice versa) will result in the second command being dropped since the tuned freq is not different. 

Fix coming soon.

jacobagilbert:
With #3844 merged I'm going to close this. Still needs backporting to 3.9 and 3.8 but the issue is resolved.

PR #3844, opened 2020-10-13
 gr-uhd: fix directional tune commands being dropped when TX and RX are the same frequency

jacobagilbert:
This is a [long overdue] fix for #1814 which adds a class member bool which will force a tune to occur if the 'direction' key is provided, and the value of that key is a symbol.

This PR also includes a commit fixing some misspellings in gr-uhd.

This should be backported all the way back to 3.7 in my opinion, but first I would like to see if this is acceptable.

### DIFF
--- a/gr-uhd/apps/uhd_app.py
+++ b/gr-uhd/apps/uhd_app.py
@@ -325,7 +325,7 @@ class UHDApp:
                 treq = uhd.tune_request(
                     target_freq=freq,
                     rf_freq=freq,
-                    rf_freq_policy=uhd.tune_reqest.POLICY_MANUAL,
+                    rf_freq_policy=uhd.tune_request.POLICY_MANUAL,
                     dsp_freq=tune_resp.actual_dsp_freq,
                     dsp_freq_policy=uhd.tune_request.POLICY_MANUAL)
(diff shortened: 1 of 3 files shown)


gr-uhd/apps/uhd_app.py
```diff
@@ -325,7 +325,7 @@ def set_freq(self, freq, skip_sync=False):
                 treq = uhd.tune_request(
                     target_freq=freq,
                     rf_freq=freq,
-                    rf_freq_policy=uhd.tune_reqest.POLICY_MANUAL,
```
mormj, line 328: how did this ever work?

mrjacobagilbert: I think this is just an example of some code nobody ever uses. The variable `tune_reqest` does not appear within the UHD code base, and it was similarly misspelled in a couple other places (local variables that were consistently misspelled thus not causing any real problems) that this PR fixes.

mrjacobagilbert:
I fixed the clang formatting failure, and added a reset to the force tune variable after processing of the command message.

mbr0wn (approved):
@jacobagilbert Sorry for the long overdue review. This is absolutely right.

PR #3848, opened 2020-10-14
cmake: manual override to ignore pybind hash checking

mormj:
This permits an OOT to override the hash checking

putting a 0 for BINDTOOL_HEADER_FILE_HASH will ignore the hash checking.
This would:
1) Not be the default behavior
2) only for OOTs
3) Use at your own risk

/***********************************************************************************/
/* This file is automatically generated using bindtool and can be manually edited  */
/* The following lines can be configured to regenerate this file during cmake      */
/* If manual edits are made, the following tags should be modified accordingly.    */
/* BINDTOOL_HEADER_FILE(foo.h)                                                     */
/* BINDTOOL_HEADER_FILE_HASH(0)                                                    */
/***********************************************************************************/


drmpeg (approved):
Tested and works good.

PR #3851, opened 2020-10-16
modtool: update of cli block name checking

mormj:
This updates the logic of the modtool checking for duplicate block names to be consistent with what is on maint-3.8 (#3387 - which was a more recent PR and got updated with feedback). 

There is no functional change here, just a cleanup of the code by @alekhgupta1441 to be consistent with the 3.8 branch.

### DIFF
--- a/gr-utils/modtool/cli/add.py
+++ b/gr-utils/modtool/cli/add.py
@@ -44,6 +44,7 @@ def cli(**kwargs):
     kwargs['cli'] = True
     self = ModToolAdd(**kwargs)
     click.secho("GNU Radio module name identified: " + self.info['modname'], fg='green')
+    get_blockname(self)
     get_blocktype(self)
     get_lang(self)
     info_lang = {'cpp': 'C++', 'python': 'Python'}[self.info['lang']]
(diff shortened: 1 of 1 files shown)


gr-utils/modtool/cli/add.py
```diff
@@ -53,7 +53,6 @@ def cli(**kwargs):
         raise ModToolException('Missing or skipping relevant subdir.')
     get_blockname(self)
```
dl1ksv, line 54: I propose to move the name checking before requesting the block type. Otherwise you have to enter some parameters before the check is made and gr_modtool stopped

dl1ksv (approved):

swap-nil7:
This looks good! Thanks, @mormj :)

swap-nil7 (approved):

gr-utils/modtool/cli/add.py
mormj: That is a very good suggestion.  i'll have to take a look at where that should go.  For this PR, was just trying to get master consistent with 3.8 - perhaps that should be another PR.

dl1ksv: It is quite easy. Just move 
line 54 get_blockname(self)
between line 46 click.secho("GNU Radio module name identified: " + self.info['modname'], fg='green') 
and line 47 get_blocktype(self) 

That's it. I tested it this weekend.

mormj: thanks!  yeah that wasn't so bad lol

mormj:
2 approvals.  Merging.

PR #3854, opened 2020-10-16
backport-38: audio: lib: windows: Add <cctype> header for std::isdigit.

ryanvolz:
Previously the header must have been pulled in by one of the Boost includes, because with a newer Boost version I had build errors. This makes the include explicit so std::isdigit is guaranteed to be declared.


noc0lour (approved):

noc0lour:
Maybe at some point I can convince enough people that include-what-you-use is a good idea

PR #3692, opened 2020-08-12
digital/correlate_access_code: Prevent false triggers

ThomasHabets:
Leftover data from last time CAC triggered, or just leading zeroes
being ignored.

This commit prevents these by assuring that we've processed at least
as many bits as the code is, before allowing a trigger.

This PR is an alternative to https://github.com/gnuradio/gnuradio/pull/3682

PR #3859, opened 2020-10-17
fix header_format_default access code search method

lawchange:
1.only search bits from input
2.reset d_data_reg and d_data_reg_len if access_code matched
3.fix overflow?

### DIFF
--- a/gr-digital/include/gnuradio/digital/header_format_default.h
+++ b/gr-digital/include/gnuradio/digital/header_format_default.h
@@ -178,6 +178,7 @@ protected:
     uint16_t d_bps; //!< bits/sec of payload modulation
 
     unsigned long long d_data_reg; //!< used to look for access_code
+    size_t d_data_reg_len; //!< Length of the saved access_code
     unsigned long long d_mask;     /*!< masks access_code bits (top N bits are set where
                                      N is the number of bits in the access code) */
     unsigned int d_threshold;      //!< how many bits may be wrong in sync vector
--- a/gr-digital/lib/header_format_default.cc
+++ b/gr-digital/lib/header_format_default.cc
@@ -33,6 +33,7 @@ header_format_default::header_format_default(const std::string& access_code,
     : header_format_base(),
       d_bps(bps),
       d_data_reg(0),
+      d_data_reg_len(0),
       d_mask(0),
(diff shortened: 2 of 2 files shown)


mormj:
@lawchange - thank you for the contribution.  This looks very similar to #3692 - can you verify whether that pull request achieves the same goal?

gr-digital/include/gnuradio/digital/header_format_default.h
```diff
@@ -178,6 +178,7 @@ class DIGITAL_API header_format_default : public header_format_base
     uint16_t d_bps; //!< bits/sec of payload modulation
 
     unsigned long long d_data_reg; //!< used to look for access_code
+    size_t d_data_reg_len; //!< Length of the saved access_code
```
willcode, line 181: //!< Number of valid bits in data register

gr-digital/lib/header_format_default.cc
```diff
@@ -112,23 +113,27 @@ bool header_format_default::parse(int nbits_in,
             while (nbits_processed < nbits_in) {
                 // shift in new data
                 d_data_reg = (d_data_reg << 1) | ((input[nbits_processed++]) & 0x1);
+                if (d_data_reg_len < d_access_code_len){
+                    ++d_data_reg_len;
+                }
+                if (d_data_reg_len == d_access_code_len){
```
willcode, line 119: Use else?

willcode (changes requested):
Looks good after a couple of very minor changes.

willcode:
@mormj This is a different piece of code than #3692 isn't it?

The check failed on a pybind header mismatch.

mormj:
Yes, you're right it is - made too cursory a glance.  To fix the pybind headers, since nothing was changed in the API, you can just update the hashes in the _python.cc

```
md5sum gr-digital/include/gnuradio/digital/header_format_default.h
```
replace the hash in gr-digital/python/bindings/header_format_default_python.cc

mbr0wn:
Due to lack of activity, I'll close this. Please reopen if that was an error on my part, we're certainly very happy about all contributions. If you do, please make sure formatting and DCO (sign-off) are in order. Thanks!

PR #3861, opened 2020-10-18
fix(change) header_format_parser_xxx ret value type

lawchange:
hdr_fmt_parser require more data is not an error.


willcode (changes requested):
The idea looks good. Explicit function return types would be good, even if we all know an enum is an int.
The build failed due to a pybind header mismatch.

marcusmueller:
In need of a rebase (and squash, and pybinding)!

mbr0wn:
This needs a  lot of work, and hasn't had a lot of activity in a while, so I'll close the PR. We do value the contribution, but we're also trying to keep the open PR list relevant, so please excuse the closing. You can reopen when it has DCO signoff, and no conflicts.

Issue #1132, opened 2016-12-06
Finite LINGER for Zeromq blocks


jaredd:
I have a dynamic distributed application where multiple flowgraphs communicate using zeromq blocks.  It is often the case that from test to test the set of flowgraphs operating changes, requiring complete teardown of one flowgraph and construction of new one in its place.  I cannot reliably produce a minimal example of this behavior, but almost surely with some number of reconfigurations (under 10) a flowgraph will fail to exit.  The prevailing theory is that ZMQ is hanging during destruction due to unserved messages in its queue.  The theory was somewhat validated by placing a finite LINGER sockopt in the zeromq base_impl constructor.  After this change flowgraphs reliably cleaned up.  LINGER seems like a reasonable option to make available to users.

jmcorgan:
Can you do a pull request with your proposed changes?

jareddatdatac:
Just a small update to this issue, when the GUI experienced a freeze when trying to teardown the flowgraph I attached to the process with gdb and got the following traceback which suggests that it is zmq waiting on unsent messages:
#0  0x00007f235424bfdd in poll () at ../sysdeps/unix/syscall-template.S:81
#1  0x00007f22f9630a6a in ?? () from /usr/lib/x86_64-linux-gnu/libzmq.so.3
#2  0x00007f22f961d2f7 in ?? () from /usr/lib/x86_64-linux-gnu/libzmq.so.3
#3  0x00007f22f961463c in ?? () from /usr/lib/x86_64-linux-gnu/libzmq.so.3
#4  0x00007f22f9645388 in zmq_ctx_term () from /usr/lib/x86_64-linux-gnu/libzmq.so.3
#5  0x00007f22f987054e in close (this=0x37adbc0) at /usr/include/zmq.hpp:288
#6  ~context_t (this=0x37adbc0, __in_chrg=<optimized out>) at /usr/include/zmq.hpp:281
#7  gr::zeromq::base_impl::~base_impl (this=0x334e5f8, __vtt_parm=<optimized out>, 
    __in_chrg=<optimized out>) at /usr/local/src/gnuradio/gr-zeromq/lib/base_impl.cc:54
#8  0x00007f22f987cdb9 in ~base_sink_impl (
    __vtt_parm=0x7f22f9a95778 <VTT for gr::zeromq::push_sink_impl+24>, this=<optimized out>, 
[...]

PR #3866, opened 2020-10-19
Add ZMQ_Linger of 1000ms to ZMQ socket creation to prevent infinite block in top_block.stop()

MattMills:
Re: #1132 the default value of ZMQ_Linger was incorrect in the ZMQ docs, and is currently -1, which causes sockets with pending messages to block indefinitely during tear down, causing flow graphs to lock up indefinitely during top_block.stop() if outbound data is pending on the ZMQ socket.

It's challenging to reliably reproduce this condition so I don't have code for that, but in my code that uses ZMQ pub/sub sinks/sources to pass data around, it regularly locks up at least once about every 8 hours in top_block.stop()

It might make more sense for ZMQ_Linger to be configurable, but that is beyond my C++ abilities, and I think a ZMQ_LINGER of 0 is sensible since it really only effects data during tear down.

### DIFF
--- a/gr-zeromq/lib/base_impl.cc
+++ b/gr-zeromq/lib/base_impl.cc
@@ -16,6 +16,10 @@
 #include "tag_headers.h"
 #include <gnuradio/io_signature.h>
 
+namespace {
+constexpr int LINGER_DEFAULT = 1000; // 1 second.
+}
+
 namespace gr {
 namespace zeromq {
 
@@ -71,6 +75,9 @@ base_sink_impl::base_sink_impl(int type,
         d_socket.setsockopt(ZMQ_HWM, &tmp, sizeof(tmp));
 #endif
     }
+
+    /* Set ZMQ_LINGER so socket won't infinitely block during teardown */
+    d_socket.setsockopt(ZMQ_LINGER, &LINGER_DEFAULT, sizeof(LINGER_DEFAULT));
 
     /* Bind */
     d_socket.bind(address);
@@ -142,6 +149,9 @@ base_source_impl::base_source_impl(int type,
         d_socket.setsockopt(ZMQ_HWM, &tmp, sizeof(tmp));
 #endif
     }
+
+    /* Set ZMQ_LINGER so socket won't infinitely block during teardown */
+    d_socket.setsockopt(ZMQ_LINGER, &LINGER_DEFAULT, sizeof(LINGER_DEFAULT));
 
     /* Connect */
     d_socket.connect(address);


willcode (changes requested):
Looks, good. Can we (1) make the default value a compromise like 1000ms so that short running programs will tend to work as people expect, and (2) put in an issue to make it configurable.

MattMills:
@willcode What do you think about the follow up issue being for exposing setsockopt and getsockopt so people can just use those to set whatever options they need/want rather than just being able to configure ZMQ_LINGER?

MattMills:
I'm not sure I understand why those tests are failing, I don't currently have a gnuradio build environment, if anyone can help with that I'd appreciate it, if not, I'll try to get my own environment building that code and see if I can figure out how to run the tests some time in the next weeks.

willcode:
@MattMills I'd lean toward having as little ZMQ specific stuff visible as possible, unless it's required. Are there other sockopts you think people will need? If set/getsockopt were exposed, how would they be used, i.e., how would you represent them in GRC?

willcode (approved):

MattMills:
@willcode I haven't a clue how that'd be represented in GRC, I do use ZMQ_RCVTIMEO and ZMQ_SNDTIMEO (and ZMQ_LINGER) on the python side of my code to tune timeouts, they aren't strictly necessary but it'd be a nice to have.

willcode (changes requested):
Missed this - get rid of "&" in sizeof. QA tests fail. After removing "&", tests pass.

MattMills:
@willcode Thanks for your help.

willcode (approved):

gr-zeromq/lib/base_impl.cc
```diff
@@ -72,6 +72,11 @@ base_sink_impl::base_sink_impl(int type,
 #endif
     }
 
+    // Set ZMQ_LINGER to 0 so socket won't infinitely block during teardown
+    // Re: Github #1132
+    const int linger_default = 1000;
```
ThomasHabets, line 77: The comment says 0, but the value says 1000?

gr-zeromq/lib/base_impl.cc
```diff
@@ -143,6 +148,11 @@ base_source_impl::base_source_impl(int type,
...
+    d_socket.setsockopt(ZMQ_LINGER, &linger_default, sizeof(linger_default));
```
ThomasHabets, line 154: Put the 1000 in one place and reuse, to not repeat.

ThomasHabets (changes requested):

gr-zeromq/lib/base_impl.cc
MattMills: Do you have a suggestion on the "right" place to initialize the const so that it is in the correct scope? That was what I was unsure of, so I figured having it on the line above worked well enough.

gr-zeromq/lib/base_impl.cc
MattMills: Fixed the comments by removing the value.

gr-zeromq/lib/base_impl.cc
ThomasHabets: Probably safest to put it in this same file, in an anonymous namespace as a global variable. Do use `constexpr`, though.

```
namespace {
constexpr int LINGER_DEFAULT = 1000; // 1 second.
}
```

Nonmember const should be capitalized per styleguide.

gr-zeromq/lib/base_impl.cc
```diff
@@ -72,6 +76,10 @@ base_sink_impl::base_sink_impl(int type,
 #endif
     }
 
+    // Set ZMQ_LINGER so socket won't infinitely block during teardown
+    // Re: Github #1132
```
dkozel, line 80: I'd prefer the GitHub issue reference to be in the commit message rather than the code. Otherwise this seems like a great improvement, thanks!

dkozel (approved):

ThomasHabets (approved):
What dkozel said, but otherwise LGTM

ThomasHabets (approved):

mormj:
3 approvals.  CI passes.  Merging

PR #3869, opened 2020-10-19
RPC: fixes to allow using Thrift 0.9 through 0.12

noc0lour:
Same PR as #2738 but rebased & reformatted on maint-3.7 to
resolve conflicts


noc0lour:
@michaelld rebased and fixed formatting, but apparently `VERSION_GREATER_EQUAL` is not available in cmake <3.7 
I'll close the previous PR.... in case you are interested to add some fix here, feel free.

michaelld:
> @michaelld rebased and fixed formatting, but apparently `VERSION_GREATER_EQUAL` is not available in cmake <3.7
> I'll close the previous PR.... in case you are interested to add some fix here, feel free.

I think we can use `NOT ... VERSION_LESS` and the like for the opposite .. that should work on older CMake as well as current.

noc0lour:
Ah, I didn't check which version checks existed before. Will adapt.

marcusmueller (approved):

Issue #3586, opened 2020-06-20
GrPybind.cmake style/formatting/...


skoslowski:
Some thought on GrPybind.cmake 
* (lots of trailing whitespace, indentation errors)
* these marcos should really be functions
* as this is gonna be used in OOTs, a proper API using `cmake_parse_arguments` would ensure future extensibility and improve readability

marcusmueller:
should be dealt with using our current cmake-format infrastructure in af78fad

PR #3870, opened 2020-10-20
digital: ofdm_equalizer_simpledfe: add output_exact_syms option

japm48:
This should solve a limitation in [`ofdm_equalizer_simpledfe`](https://www.gnuradio.org/doc/doxygen/classgr_1_1digital_1_1ofdm__equalizer__simpledfe.html):
> Note that the equalized symbols are exact points on the constellation. This means soft information of the modulation symbols is lost after the equalization, which is suboptimal for channel codes that use soft decision.

Hopefully, this simple fix has no downsides... (it seemed *too* obvious to me)

Here is a grc flowgraph based upon the rx_ofdm example: https://gist.github.com/japm48/4ce310709702bb84d12c980cc0ed2330

I can add this to the examples folder, or maybe create a testbench (but I'm not really sure how properly do that in this case).

Here is the relevant result:

screenshot

### DIFF
--- a/gr-digital/python/digital/bindings/ofdm_equalizer_simpledfe_python.cc
+++ b/gr-digital/python/digital/bindings/ofdm_equalizer_simpledfe_python.cc
@@ -14,7 +14,7 @@
 /* BINDTOOL_GEN_AUTOMATIC(0)                                                       */
 /* BINDTOOL_USE_PYGCCXML(0)                                                        */
 /* BINDTOOL_HEADER_FILE(ofdm_equalizer_simpledfe.h) */
-/* BINDTOOL_HEADER_FILE_HASH(8091e37dbba4afc0b7a2fd9441996144)                     */
+/* BINDTOOL_HEADER_FILE_HASH(e484906ff9740a46f83b999ec69f511e)                     */
 /***********************************************************************************/
 
 #include <pybind11/complex.h>
@@ -47,6 +47,7 @@ void bind_ofdm_equalizer_simpledfe(py::module& m)
              py::arg("symbols_skipped") = 0,
              py::arg("alpha") = 0.10000000000000001,
              py::arg("input_is_shifted") = true,
+             py::arg("enable_soft_output") = false,
              D(ofdm_equalizer_simpledfe, make))
 
 
--- a/gr-digital/python/digital/qa_ofdm_frame_equalizer_vcvc.py
+++ b/gr-digital/python/digital/qa_ofdm_frame_equalizer_vcvc.py
@@ -167,6 +167,7 @@ class qa_ofdm_frame_equalizer_vcvc (gr_unittest.TestCase):
         # Check data
         self.assertComplexTuplesAlmostEqual(rx_expected, sink.data()[0], places=4)
 
+
     def test_002_static (self):
         """
         - Add a simple channel
@@ -229,6 +230,9 @@ class qa_ofdm_frame_equalizer_vcvc (gr_unittest.TestCase):
         expected_dict = {
                 'ofdm_sync_chan_taps': channel[-fft_len:]
         }
+        # Won't be exactly the same when compiled with -fcx-limited-range
+        self.assertTrue(numpy.allclose(tag_dict['ofdm_sync_chan_taps'], expected_dict['ofdm_sync_chan_taps']))
+        expected_dict['ofdm_sync_chan_taps'] = tag_dict['ofdm_sync_chan_taps']
         self.assertEqual(tag_dict, expected_dict)
 
     def test_002_static_wo_tags (self):
@@ -330,6 +334,9 @@ class qa_ofdm_frame_equalizer_vcvc (gr_unittest.TestCase):
         equalizer = digital.ofdm_equalizer_simpledfe(
             fft_len, cnst.base(), occupied_carriers, pilot_carriers, pilot_symbols, 0, 0.01
         )
+        equalizer_soft = digital.ofdm_equalizer_simpledfe(
+            fft_len, cnst.base(), occupied_carriers, pilot_carriers, pilot_symbols, 0, 0.01, enable_soft_output=True
+        )
         channel = [
             0, 0,  1,  1, 0,  1,  1, 0,
             0, 0,  1,  1, 0,  1,  1, 0, # These coefficients will be rotated slightly...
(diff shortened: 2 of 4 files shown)


willcode (changes requested):
The change looks safe, and the default will be the current behavior.

Is there better terminology (e.g., hard/soft, sliced) that should be used instead of exact/not?

mormj:
Maybe "forced constellation alignment"?  Hard/soft could work as well.

noc0lour:
On terminology I think you could also use hard/soft decision. So the switch would be "output_soft_decision". Although we then would need to also have a soft-decision demodulator which then can take this soft decision input and convert it to soft-decision bits.

willcode:
Hunted around for terminology a little. A soft decision implies that a decision has been made but with some goodness metric, such as log likelyhood. Here, it's a question of slicing (making a decision) or not. How about something along the lines of slice_enable or slice_disable.

noc0lour:
@japm48 for the testbench you basically can look into the python/ folder in a `qa_*.py` file and basically copy the setup and e.g. generate a symbolstream which you know the outcome of and check for it after the operation.

noc0lour:
@willcode makes sense and probably a good idea to implement a soft-decision demapper which outputs LLRs or similar as input for FEC decoders as a follow up.

japm48:
@noc0lour
> for the testbench you basically can look into the python/ folder in a `qa_*.py` file and basically copy the setup and e.g. generate a symbolstream which you know the outcome of and check for it after the operation.

Ok, I will try that this weekend.

> makes sense and probably a good idea to implement a soft-decision demapper which outputs LLRs or similar as input for FEC decoders as a follow up.

That was what I had in mind. Note that LLR demappers are surprisingly tricky for anything more complicated than QPSK (see this [nice paper](https://arxiv.org/pdf/1907.01512.pdf)), and even if we pick only the easy approximate method, we need to measure the channel noise σ^2 (possibly from the equalizer? or we can estimate that from the noisy symbols themselves).

Something could be reused from @drmpeg's https://github.com/drmpeg/gr-dvbs2rx.

noc0lour:
I've been told there already exists some soft demapper in GNU Radio but it might not be doing what we think it is doing, but this is kinda irrelevant for this PR, just to keep in mind for the future :) 

For the tests: testing for a known sequence in/out will only ensure the math in future works the same it does right now, it does not really logically prove this block works. Basically just giving us a safeguard that some random change in the codebase does not break current state.

noc0lour:
Also the builders are reporting this: 
```
CMake Error at cmake/Modules/GrPybind.cmake:82 (message):
  Python bindings for ofdm_equalizer_simpledfe.h are out of sync
  gr-digital/python/digital/bindings/CMakeLists.txt:96 (GR_PYBIND_MAKE_CHECK_HASH)
```

mormj:
The pybind bindings need to be updated since the public header has changed.
This can probably be done most easily by manually editing ofdm_equalizer_simpledfe_python.cc to add the additional argument and default value, then updating the md5sum in that file (`md5sum include/gnuradio/digital/ofdm_equalizer_simpledfe.h`)

japm48:
@mormj Oops, probably forgot to update them.
BTW, I wrote [this](https://gist.github.com/japm48/1a728b589dfbd311ca13ba0126f4f50c) bash function so I can just call `gr_update_hash ofdm_equalizer_simpledfe`.

noc0lour:
@japm48 I was about to write a similar function, maybe you are interested to contribute it to our `tools/` directory?

japm48:
@noc0lour I can do that if you want, but it's a bit too hackish.
I think a better option could be to add a custom target in cmake that updates it on demand and get a more helpful error message (e.g. `Python bindings for ... are out of sync. You may execute "make update_hash" to solve this`).

mormj:
If you're in the pybind cmake and up for making improvements there, there is also https://github.com/gnuradio/gnuradio/issues/3586 .  Any ideas you have to make the binding workflow smoother are much appreciated as well.

willcode:
@japm48 any plans to update terminology, e.g., "Disable Slicing", disable_slicing?

japm48:
@willcode Sorry, I was busy.
I'll probably go with "enable_soft_output" if no one objects to that. "Slicing" doesn't sound evocative enough to me, but possibly it's just me (not a native English speaker).

gr-digital/python/digital/qa_ofdm_frame_equalizer_vcvc.py
```diff
@@ -167,6 +167,8 @@ def test_001c_carrier_offset_cp (self):
         # Check data
         self.assertComplexTuplesAlmostEqual(rx_expected, sink.data()[0], places=4)
 
+
+    @gr_unittest.unittest.skip("FIXME: test fails")
```
noc0lour, line 171: what's up wtih this? or is the test just broken in general?

noc0lour (commented):
Aside from the test weirdness looks good. Lmk if you need some help with that.

gr-digital/python/digital/qa_ofdm_frame_equalizer_vcvc.py
willcode: Since this is the non-sliced output, it's hard to test. How do we usually test the analog output of a complicated block?

japm48: @noc0lour It was already broken, at least on my machine.

gr-digital/python/digital/bindings/ofdm_equalizer_simpledfe_python.cc
```diff
@@ -14,7 +14,7 @@
 /* BINDTOOL_GEN_AUTOMATIC(0)                                                       */
 /* BINDTOOL_USE_PYGCCXML(0)                                                        */
 /* BINDTOOL_HEADER_FILE(ofdm_equalizer_simpledfe.h) */
-/* BINDTOOL_HEADER_FILE_HASH(8091e37dbba4afc0b7a2fd9441996144)                     */
+/* BINDTOOL_HEADER_FILE_HASH(c0a20cbce550a3d46250a09524c9a229)                     */
```
willcode, line 17: e484906ff9740a46f83b999ec69f511e

japm48: Thanks, I always forget that.

gr-digital/python/digital/qa_ofdm_frame_equalizer_vcvc.py
willcode: This test works without the patch. Something needs updating ... hunting.

japm48: I get this error using current master: https://gist.github.com/japm48/0d01120518c4311138284c421a927194
I'll try to dig in and see what's happening.

drmpeg: I know what that is. I see it on ARM.  It's caused by the compiler flag ```-fcx-limited-range```.

https://github.com/gnuradio/gnuradio/pull/3114

willcode: I think this is just a path issue. The test fails for me if there is an existing install in my path ... it picks up the wrong gr-digital libraries. Give it a try with no env set. Unfortunately that means making a parallel installation of Volk somewhere. First, just try enabling the test and see if it runs under buildbot.

noc0lour: Ye, the test python stuff is messed up w.r.t to existing libraries and for the tests to run successfully you have to install GNU Radio because the python library is only assembled then. (IIRC)

noc0lour: > I get this error using current master: https://gist.github.com/japm48/0d01120518c4311138284c421a927194
> I'll try to dig in and see what's happening.

Shouldn't be almostEqual and using the last (5) digits be enough precision for that?

japm48: @drmpeg Thanks! I can confirm that `-fcx-limited-range` was the culprit (I recompiled after removing the flag). Volk is unrelated here (I'm using the system package in Arch). In case it's needed: gcc version 10.2.0.

@willcode No duplicate library problem here, I already learned the hard way that I need to avoid that.

@noc0lour I'm guessing it wasn't done that because assertAlmostEqual doesn't work on dicts, so I'll use a simple workaround.

willcode (approved):

mormj (approved):

mormj:
CI passes. Two reviews.  Low risk, backwards compatible with default value.  Merging.

PR #3874, opened 2020-10-20
runtime/build: Install swig even without python

doronbehar:
On NixOS, we noticed that if python-support is not enabled, many applications that depend on gnuradio-runtime fail to build. With this patch, swig files are always installed, regardless of whether python-support is enabled.


doronbehar:
3.7 backport: https://github.com/gnuradio/gnuradio/pull/3875

mrjacobagilbert:
Can you be more specific on the failures you are seeing?

doronbehar:
Sure, for (3.7) with gr-osmosdr (with the last version of gr-osmosdr that supports 3.7):

```
gr-osmosdr> /build/gr-osmosdr/swig/osmosdr_swig.i:8: Error: Unable to find 'gnuradio.i'
gr-osmosdr> make[2]: *** [swig/CMakeFiles/osmosdr_swig_swig_2d0df.dir/build.make:117: swig/osmosdr_swig_swig_2d0df] Error 1
gr-osmosdr> make[2]: *** Deleting file 'swig/osmosdr_swig_swig_2d0df'
gr-osmosdr> make[1]: *** [CMakeFiles/Makefile2:358: swig/CMakeFiles/osmosdr_swig_swig_2d0df.dir/all] Error 2
gr-osmosdr> make[1]: *** Waiting for unfinished jobs....
```

mrjacobagilbert:
It seems to me like the OOT should not be building swig related stuff at all if it is being built with `-DENABLE_PYTHON=OFF`

https://github.com/osmocom/gr-osmosdr/blob/gr3.7/CMakeLists.txt#L240

I wonder why its doing swig stuff at all...

mormj:
There should be no reason for swig if python is not enabled.  Seems like this should be fixed on the OOT side

doronbehar:
Well if python and swig are available during build of gr-osmosdr, it assumes (righteously I suppose) that GNUradio was built with swig files as well. And, if I don't add python and swig to the build of gr-osmosdr, and I build it with a gnuradio without python support, it builds fine, as expected. Hence it means we should not enable python support for gr-osmosdr when it's built with a gnuradio without python support as well. I'll close this and the other PR as they are not justified indeed. Thanks for the feedback.

PR #3878, opened 2020-10-22
gr-video-sdl: Fix YUV formats.

drmpeg:


gr-video-sdl/grc/video_sdl_sink.block.yml
```diff
@@ -32,9 +32,9 @@ parameters:
     default: '0'
 -   id: num_channels
     label: Channels
-    dtype: enum
+    dtype: int
```
mormj, line 35: This change is sensible, I'm assuming to be able to programmatically change the option in grc, but it would also be nice to expose the enums through the API and use that instead of 1/2/3

drmpeg: You can't use an enum for "multiplicity". It just doesn't work.

mormj: gotcha.

mormj (approved):

drmpeg:
Test flow graph and YUV files. Warning, large download - 662,396,510 bytes.

www.w6rz.net/sdltest.tar.gz

mormj:
Thanks for providing the test files.  Before the patch, the flowgraph is completely broken.  After the patch it works great!

mormj:
Low enough impact - merging

PR #3629, opened 2020-07-07
Add extra includes to build oot modules that use extra packages like qt5

dl1ksv:
Some oot modules like gr-display use additional packages like for instance qt5.  
To create the bindings it is required to know the directories of the corresponding include files.

This patch provides a parameter for these directories to be added.

PR #3879, opened 2020-10-22
Make additional includes and defines available in the bind process

dl1ksv:
At the moment neither gr_modtool bind nor the GR_PYBIND_MAKE_OOT macro consider additional ( required ) includes like QWidgets or consider defines like ENABLE_PYTHON.

So bind may fail or generate wrong bind code.

This pr fixes this issue and includes #3629


dl1ksv:
The test_case_2 branch in my [gr_display ](https://github.com/dl1ksv/gr-display) may serve as test example

mormj (approved):
Tested with and without using the -D switch.  Works as expected

mormj:
Do you think there is any need to specify defines with values (#if vs #ifdef)?

dl1ksv:
The -D switch behaves like in gcc -D. The values are passed to the passer. I tried some combinations and it seems to work. So we don't have to differ on this level.

PR #3838, opened 2020-10-12
gr-qtgui: Remove unused set_title() method in QT GUI Eye Sink.

drmpeg:

PR #3880, opened 2020-10-22
Update qa_qtgui.py

dl1ksv:
Test fails with
  File "/home/schroer/gnuradiocomponents/gnuradio-volker/gr-qtgui/python/qtgui/qa_qtgui.py", line 78, in test13
    self.qtsnk = qtgui.eye_sink_c(1024, 1, "Test", 1, None)
TypeError: __init__(): incompatible constructor arguments. The following argument types are supported:
    1. gnuradio.qtgui.qtgui_python.eye_sink_c(size: int, samp_rate: float, nconnections: int = 1, parent: gnuradio.qtgui.qtgui_python.QWidget = None)

Invoked with: 1024, 1, 'Test', 1, None

eye_sink_f/c has no string parameter

And eye_sink_c and eye_sink_f should be tested

### DIFF
--- a/gr-qtgui/python/qtgui/qa_qtgui.py
+++ b/gr-qtgui/python/qtgui/qa_qtgui.py
@@ -72,10 +72,10 @@ class test_qtgui(gr_unittest.TestCase):
         self.qtsnk = qtgui.histogram_sink_f(1024, 100, -1, 1, "Test", 1, None)
 
     def test13(self):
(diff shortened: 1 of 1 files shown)


drmpeg (approved):
This was caused by https://github.com/gnuradio/gnuradio/pull/3838 The build machine/CI doesn't test qtgui, so it slipped through.

mormj (approved):

PR #3881, opened 2020-10-22
pybind: add implicit conversion from int to enum

mormj:
With swig, there was an automatic conversion from int to enum, so ints could be passed through the python bindings.  This reenables that functionality with pybind11 ~~- if this looks acceptable, I'll do this for all the modules and it as an automatic step in binding enums.~~

UPDATE: propagated for all enums used in block parameters
Also, added to the bindtool so that `modtool bind` will add the implicit conversions

### DIFF
--- a/gr-analog/python/analog/bindings/cpm_python.cc
+++ b/gr-analog/python/analog/bindings/cpm_python.cc
@@ -51,4 +51,6 @@ void bind_cpm(py::module& m)
                          py::arg("L"),
                          py::arg("beta") = 0.3,
                          D(cpm, phase_response));
+
+    py::implicitly_convertible<int, gr::analog::cpm::cpm_type>();
 }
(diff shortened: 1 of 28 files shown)


drmpeg (approved):
I'm not exactly sure how to test this. Would this solve the problem from https://github.com/gnuradio/gnuradio/pull/3878 where ```multiplicity``` can't be an enum?

However, it doesn't seem to break anything with dtv flow graphs, so I'm going to approve it.

mormj:
It should be testable by doing something by instantiating a block where the API is `dtv.block(...,enum,...)`, now should work with `dtv.block(...,int,...)`.  The second instantiation should fail before this patch.

I don't think this would work the other way around that would solve the multiplicity issue.  Only would allow the block API to have an enum, then from GRC pass it an int - the int could be used for multiplicity, but then is there much use in having the enum in the API at that point?

mormj:
also, i can squash since there isn't much in each of these commits

drmpeg:
I get it now. I was able to test it with this stand-alone Python code for dtv. For example, changing ```dtv.SHOWLEVELS_OFF``` to 0 triggers the error (which this patch fixes).

http://www.w6rz.net/dvbt2-b210.py

The use case might be a little obscure, but still good stuff.

PR #532, opened 2015-05-11
remove and/or/xor keywords.

LazyDodo:

PR #3885, opened 2020-10-24
utils: Added print-all to gnuradio-config-info

dkozel:
Signed-off-by: Derek Kozel <derek@bitstovolts.com>

A small usability addition to print most of the config info easily. This doesn't change the existing behaviors or formatting. I think the formatting could be improved, but am not sure if anyone is using the raw strings and we'd break their scripts. Possibly something to introduce early in the 3.10 development to give lots of advance warning.

### DIFF
--- a/gnuradio-runtime/apps/gnuradio-config-info.cc
+++ b/gnuradio-runtime/apps/gnuradio-config-info.cc
@@ -29,7 +29,9 @@ int main(int argc, char** argv)
         (format("Program options: %1% [options]") % argv[0]).str());
     po::variables_map vm;
 
+    // clang-format off
     desc.add_options()("help,h", "print help message")(
+        "print-all", "print all information")(
         "prefix", "print GNU Radio installation prefix")(
         "sysconfdir", "print GNU Radio system configuration directory")(
         "prefsdir", "print GNU Radio preferences directory")(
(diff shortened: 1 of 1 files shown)


drmpeg:
I'm pretty sure the use of ```or``` instead of ```||``` is not compatible with all compilers.

marcusmueller:
> I'm pretty sure the use of or instead of || is not compatible with all compilers.

used to be problematic with MSVC, iirc, but our currently required versions are recent enough if I'm groking all this correctly (btw, if you want this in C, `#include <iso646.h>`.)

marcusmueller:
The question is whether `or` is in compliance with our style. And since clang-format doesn't complain about it, I'll shrug...

marcusmueller (approved):

drmpeg:
Yeah, I just remember PR's like this one https://github.com/gnuradio/gnuradio/pull/532

If you look through the code, I believe they've all been removed. I guess I'll shrug too.

marcusmueller:
@dkozel Knowing you a bit, I guess we'll have about three shrugs among us then, and since I rebased this by hand anyways... search and replaced the `or` for a `||`, as well as unifying the use of `vm.count()`. Just waiting for CI to go through with this, merging then.

dkozel:
Shrug. I find `or` clearer in a fresh codebase. No complaints about `||` here though!

PR #3887, opened 2020-10-26
gr-uhd/lib/usrp_block_impl.cc: fix compile error using boost>=1.73

vowstar:
When using boost>=1.73 and compile with uhd, it always have errors.
This patch fix the compile error when using uhd and gcc10: error: ‘_1’ was not declared.

Bug: https://bugs.gentoo.org/751259
Log(xz format): https://bugs.gentoo.org/attachment.cgi?id=668564

I've test and it seems works under my environment.

### DIFF
--- a/gr-uhd/lib/usrp_block_impl.cc
+++ b/gr-uhd/lib/usrp_block_impl.cc
@@ -133,9 +133,11 @@ usrp_block_impl::usrp_block_impl(const ::uhd::device_addr_t& device_addr,
                     [this](pmt::pmt_t msg) { this->msg_handler_command(msg); });
 
 // cuz we lazy:
-#define REGISTER_CMD_HANDLER(key, _handler) \
-    register_msg_cmd_handler(key,           \
-                             boost::bind(&usrp_block_impl::_handler, this, _1, _2, _3))
+#define REGISTER_CMD_HANDLER(key, _handler)                                   \
+    register_msg_cmd_handler(                                                 \
+        key, [this](const pmt::pmt_t& var, int chan, const pmt::pmt_t& msg) { \
+            this->_handler(var, chan, msg);                                   \
+        })
     // Register default command handlers:
     REGISTER_CMD_HANDLER(cmd_freq_key(), _cmd_handler_freq);
     REGISTER_CMD_HANDLER(cmd_gain_key(), _cmd_handler_gain);
@@ -251,11 +253,13 @@ bool usrp_block_impl::_check_mboard_sensors_locked()
         } else if (_dev->get_clock_source(mboard_index) == "mimo") {
             sensor_name = "mimo_locked";
         }
-        if (not _wait_for_locked_sensor(
-                get_mboard_sensor_names(mboard_index),
-                sensor_name,
-                boost::bind(
-                    &usrp_block_impl::get_mboard_sensor, this, _1, mboard_index))) {
+        if (not _wait_for_locked_sensor(get_mboard_sensor_names(mboard_index),
+                                        sensor_name,
+                                        [this, mboard_index](const std::string& name) {
+                                            return static_cast<::uhd::sensor_value_t>(
+                                                this->get_mboard_sensor(name,
+                                                                        mboard_index));
+                                        })) {
             GR_LOG_WARN(
                 d_logger,
                 boost::format(
--- a/gr-uhd/lib/usrp_source_impl.cc
+++ b/gr-uhd/lib/usrp_source_impl.cc
@@ -46,7 +46,9 @@ usrp_source_impl::usrp_source_impl(const ::uhd::device_addr_t& device_addr,
     _samp_rate = this->get_samp_rate();
     _samps_per_packet = 1;
     register_msg_cmd_handler(cmd_tag_key(),
-                             boost::bind(&usrp_source_impl::_cmd_handler_tag, this, _1));
+                             [this](const pmt::pmt_t& tag, const int, const pmt::pmt_t&) {
+                                 this->_cmd_handler_tag(tag);
+                             });
 }
 
 usrp_source_impl::~usrp_source_impl() {}
--- a/gr-uhd/lib/usrp_block_impl.h
+++ b/gr-uhd/lib/usrp_block_impl.h
@@ -14,8 +14,8 @@
 #include <gnuradio/uhd/usrp_block.h>
 #include <pmt/pmt.h>
 #include <uhd/usrp/multi_usrp.hpp>
-#include <boost/bind.hpp>
 #include <boost/dynamic_bitset.hpp>
+#include <functional>
 
 
 namespace gr {
@@ -31,9 +31,8 @@ static const std::string ALL_LOS;
 class usrp_block_impl : virtual public usrp_block
 {
 public:
-    typedef boost::function<::uhd::sensor_value_t(const std::string&)> get_sensor_fn_t;
-    typedef boost::function<void(const pmt::pmt_t&, int, const pmt::pmt_t&)>
-        cmd_handler_t;
+    typedef std::function<::uhd::sensor_value_t(const std::string&)> get_sensor_fn_t;
+    typedef std::function<void(const pmt::pmt_t&, int, const pmt::pmt_t&)> cmd_handler_t;
 
     static const double LOCK_TIMEOUT;
 


drmpeg:
I think you want ```using namespace boost::placeholders;``` instead of ```using namespace std::placeholders;```.

vowstar:
> I think you want `using namespace boost::placeholders;` instead of `using namespace std::placeholders;`.

Thanks, I'll change it to `using namespace boost::placeholders;`

vowstar:
The patch i posted at https://751259.bugs.gentoo.org/attachment.cgi?id=668573 works (for old version gnuradio-3.8.2.0)

mormj:
do you also need this on usrp_source_impl.cc?  there is a bind with `_1` there as well

vowstar:
> do you also need this on usrp_source_impl.cc? there is a bind with `_1` there as well

Done

marcusmueller:
@jdemel did this a while ago on a larger scale, IIRC

jdemel:
Yeah. We discussed this issue in #3441. The gist of it: Use C++ `lambda`s instead of `boost::bind`. 
1. `lambda`s are faster.
2. `lambda`s work with `boost` and `std`. We want to move to `std` anyways.

vowstar:
@jdemel Thanks for review, I've changed related code to lambda.
And then found a bug:  in files usrp_block_impl.cc and usrp_block_impl.h , usrp_block_impl::register_msg_cmd_handler uses cmd_handler_t as handler type, but it have 3 parameters. So fixed the parameter mismatch bug and it could passed the CI test.

gr-uhd/lib/usrp_block_impl.cc
```diff
@@ -251,11 +253,13 @@ bool usrp_block_impl::_check_mboard_sensors_locked()
...
-        if (not _wait_for_locked_sensor(
-                get_mboard_sensor_names(mboard_index),
-                sensor_name,
-                boost::bind(
-                    &usrp_block_impl::get_mboard_sensor, this, _1, mboard_index))) {
+        if (not _wait_for_locked_sensor(get_mboard_sensor_names(mboard_index),
+                                        sensor_name,
+                                        [this, mboard_index](const std::string& name) {
+                                            return static_cast<::uhd::sensor_value_t>(
+                                                this->get_mboard_sensor(name,
```
vowstar, line 260: NOTE: here uses static_cast<::uhd::sensor_value_t> to fix the convert problem of lambda function with reference of local variable

drmpeg:
At this point, you really need to squash all these commits.

git rebase --interactive HEAD~13

vowstar:
@drmpeg Thank you for help, done

noc0lour:
> At this point, you really need to squash all these commits.
> 
> git rebase --interactive HEAD~13

and/or if you want to rebase onto the latest master to resolve merge conflicts yourself: `git rebase --interactive origin/master` (in case origin points to github.com/gnuradio/gnuradio)

noc0lour (approved):
:+1:  had to check a bit for the convert problem, but seems legit. Apparently an equivalent method is using trailing return type but that might not be available in whatever C++ standard is used for master.

vowstar:
>  had to check a bit for the convert problem, but seems legit. Apparently an equivalent method is using trailing return type but that might not be available in whatever C++ standard is used for master.

Thank you

vowstar:
@mormj @noc0lour The code have been rebased to origin/master. 
I've send mail to bhilburn<at>gnuradio.org and ask for a copy of CLA, are there other steps I need to complete?
Thank you

vowstar:
Excuse me, I've send mail to bhilburn < at > gnuradio.org for a copy of CLA, but seems no response.  
Is it because I sent the wrong email address or the wrong way?

vowstar:
Are there any news? @noc0lour @marcusmueller @jdemel

mohmmadd:
> Excuse me, I've send mail to bhilburn < at > gnuradio.org for a copy of CLA, but seems no response.
> Is it because I sent the wrong email address or the wrong way?

Maybe you should try: bhilburn< at >gmail.com

mormj:
@vowstar - we can now merge code if you sign your commits (agreeing to the DCO).  Use git commit -s --amend and force push to pass the DCO check

https://github.com/gnuradio/gnuradio/blob/master/CONTRIBUTING.md
(this was very recently updated to allow DCO)

jdemel (approved):
LGTM

vowstar:
> @vowstar - we can now merge code if you sign your commits (agreeing to the DCO). Use git commit -s --amend and force push to pass the DCO check
> 
> https://github.com/gnuradio/gnuradio/blob/master/CONTRIBUTING.md
> (this was very recently updated to allow DCO)

I've signed and forced re-push.

gr-uhd/lib/usrp_source_impl.h
```diff
@@ -113,7 +113,7 @@ class usrp_source_impl : public usrp_source, public usrp_block_impl
     //! Like set_center_freq(), but uses _curr_freq and _curr_lo_offset
     ::uhd::tune_result_t _set_center_freq_from_internals(size_t chan,
                                                          pmt::pmt_t direction) override;
-    void _cmd_handler_tag(const pmt::pmt_t& tag);
+    void _cmd_handler_tag(const pmt::pmt_t& tag, int chan, const pmt::pmt_t& msg);
```
mbr0wn, line 116:
```suggestion
    void _cmd_handler_tag(const pmt::pmt_t& tag);
```

gr-uhd/lib/usrp_source_impl.cc
```diff
@@ -449,7 +451,12 @@ void usrp_source_impl::set_stream_args(const ::uhd::stream_args_t& stream_args)
     }
 }
 
-void usrp_source_impl::_cmd_handler_tag(const pmt::pmt_t& tag) { _tag_now = true; }
+void usrp_source_impl::_cmd_handler_tag(const pmt::pmt_t& tag,
+                                        int chan,
+                                        const pmt::pmt_t& msg)
+{
```
mbr0wn, line 457:
```suggestion
void usrp_source_impl::_cmd_handler_tag(const pmt::pmt_t&)
{
```

gr-uhd/lib/usrp_source_impl.cc
```diff
@@ -45,8 +45,10 @@ usrp_source_impl::usrp_source_impl(const ::uhd::device_addr_t& device_addr,
 
     _samp_rate = this->get_samp_rate();
     _samps_per_packet = 1;
-    register_msg_cmd_handler(cmd_tag_key(),
-                             boost::bind(&usrp_source_impl::_cmd_handler_tag, this, _1));
+    register_msg_cmd_handler(
+        cmd_tag_key(), [this](const pmt::pmt_t& tag, int chan, const pmt::pmt_t& msg) {
+            this->_cmd_handler_tag(tag, chan, msg);
```
mbr0wn, line 50:
```suggestion
        cmd_tag_key(), [this](const pmt::pmt_t& tag, const int, const pmt::pmt_t&) {
            this->_cmd_handler_tag(tag, chan, msg);
```

mbr0wn (changes requested):
Many many thanks for using the lambdas. I have two comments:
- It is not necessary to also change the signature of _cmd_handler_tag()
- The code will have unused-arg warnings as-is

Please change, and it's good otherwise. Thanks again.

gr-uhd/lib/usrp_source_impl.h
vowstar: Done, the latest push keep signature of ``_cmd_handler_tag()`` unchanged

gr-uhd/lib/usrp_source_impl.cc
vowstar: Done, the latest push keep signature of ``_cmd_handler_tag()`` unchanged.

gr-uhd/lib/usrp_source_impl.cc
vowstar: Done, changed to ``cmd_tag_key(), [this](const pmt::pmt_t& tag, const int, const pmt::pmt_t&)`` and let ``this->_cmd_handler_tag(tag);`` unused ``chan, msg``.

mbr0wn (approved):

mbr0wn:
I think this can be backported to 3.8 too.

PR #3895, opened 2020-10-28
docs: use MathJax for HTML documentation

### LINKED ISSUE #2707: Doxygen: consider moving away from latex towards MathJax
doxygen does support that, and it reduces the cumulative dependency size of a full build significantly if we don't need latex for the dozen or so formulas we have in doxygen, and it does in fact look nicer.

Close if consideration yields negative result.

marcusmueller:
Rebase of #3456, minor reversal of accidental changes

----
Fixes #2707.

A full LaTeX installation is no longer required for generating the HTML
docs. Also adds some nice-to-have functionalities accessible via
right-click context menu.

Details:
 - SVG backend selected by default.
 - MathJax recommends not (ab)using their CDN, so
   this adds a local dependency.
 - Some spurious doxygen warnings were also removed.


marcusmueller (commented):
As with the base PR, I approve this. For some reasons, I just couldn't push over japm48's branch, so I made a new PR.

Issue #1591, opened 2018-02-06
cleanup gru Python module


marcusmueller:
I'm generally not a big fan of "misc" namespaces, and I'm pretty sure that there's a few things in there that no-one uses.

Notice that formally, `gnuradio.gru` is external API, but honestly... There's so little in there that was ever meant to be API that I don't think we'd break anything external if we deleted that whole module today.

Obviously, low prio.

bravech:
Hi, I'm looking to get started with contributing to open source (hopefully joining GNURadio for GSoC 2020!). Is this change still something that's needed? If not, could you please point me in the direction of an issue to get started with.

gateship1:
Hello, I am looking for a first issue. I noticed this is still under "To do" in Release 3.9. Can i take a crack at this one?

dkozel:
@gateship1 Yes, please do. I'm not clear whether there are only some unused functions that can be deleted, the entire module should be deprecated, or the functions should be moved into the calling functions. You'll have to see what's still in there as this issue is over two years old.

argilo:
Re-opening, since gru isn't removed yet.

marcusmueller:
With #3896, we have but left in actual use:

- [ ] `daemonize` in the dialtone example. Do we really need to use that? It feels like it was some Python <2.5-era hack
- [ ] `freqz`/`gnuplot_freqz`: used only in commented out code in `fm_emph`, and honestly, what about `scipy.signal`? 
- [ ] `hexshort`: unused
- [ ] `hexint`: used in crc32. `hexint = lambda mask: (mask - 2**32) if (mask >= 2**31) else mask`.
- [ ] `msgq_runner`: single-point usage in `uhd_app.py`. @mbr0wn is moving it to that app itself sensible?

marcusmueller:
Thanks, @argilo! Awesome!

PR #3896, opened 2020-10-29
Remove gcd, lcm, and log2 from gru module

argilo:
This continues the work on #1591.

`math.gcd` is available since Python 3.5, and we can use this to replace the single usage of `gru.gcd` in GNU Radio.

`gru.lcm` is not needed, since LCM can be replaced with GCD using the formula LCM(a, b) = a * b / GCD(a, b). I've done that in the few places where `gru.lcm` was used in GNU Radio.

`math.log2` is available since Python 3.3, and `gru.log2` is not used anywhere in GNU Radio. (The few places that need base-2 logarithms use `numpy.log2`.)


argilo:
I tested the modified scripts and they all appear to work as before.

The `uhd_rx_nogui` currently has another bug that prevents it from working at all; I'll fix that in a separate PR since it's unrelated to the removal of gru.

marcusmueller (approved):

PR #3899, opened 2020-10-30
add_blk: only memcpy for exactly 1 input, complex specialization

### LINKED ISSUE #986: Avoid unnecessary memcpy() in some basic math operation blocks
A few basic math operation blocks that support a variable number of inputs (add_ff, multiply_ff, multiply_cc, etc.), always perform a usually unnecessary memcpy() to support the extreme case of only 1 input stream.

These blocks can get a performance boost for the very common case of 2 input streams by rewriting the loop to avoid the memcpy(), if the number of input items is greater than 1.

A revised add_ff work() body might look like this:
```
if (input_items.size() == 1) {
    memcpy(out, input_items[0], noi_sizeof(float));
} else {
      volk_32f_x2_add_32f(out, (const float *)input_items[0], (const float_)input_items[1], noi);
      for(size_t i = 2; i < input_items.size(); i++)
        volk_32f_x2_add_32f(out, out, (const float*)input_items[i], noi);
}
return noutput_items;
````

### PR DESCRIPTION
**Pull request by marcusmueller**
fixes #986

This is ... kinda obvious.

### DIFF
--- a/gr-digital/python/digital/qa_ofdm_chanest_vcvc.py
+++ b/gr-digital/python/digital/qa_ofdm_chanest_vcvc.py
@@ -301,10 +301,11 @@ class qa_ofdm_chanest_vcvc (gr_unittest.TestCase):
 
         def run_flow_graph(sync_sym1, sync_sym2, data_sym):
             top_block = gr.top_block()
-            carr_offset = random.randint(-max_offset / 2, max_offset / 2) * 2
+            carr_offset = random.randint(-max_offset // 2, max_offset // 2) * 2
             tx_data = shift_tuple(sync_sym1, carr_offset) + \
                 shift_tuple(sync_sym2, carr_offset) + \
-                shift_tuple(data_sym, carr_offset)
+                shift_tuple(data_sym, carr_offset) + \
+                tuple((0,) * fft_len)
             channel = [
                 rand_range(
                     min_chan_ampl,
--- a/gr-blocks/lib/add_blk_impl.cc
+++ b/gr-blocks/lib/add_blk_impl.cc
@@ -27,6 +27,7 @@ typename add_blk<T>::sptr add_blk<T>::make(size_t vlen)
 }
 
 
+// Float implementation
 template <>
 add_blk_impl<float>::add_blk_impl(size_t vlen)
     : sync_block("add_ff",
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Comment by marcusmueller** [comment]
Um, why does `qa_ofdm_chanest_vcvc` now consistently fail?!

**Comment by marcusmueller** [comment]
```
  File "/home/marcus/src/gnuradio/gr-digital/python/digital/qa_ofdm_chanest_vcvc.py", line 263, in test_999_all_at_once
    (carr_offset, rx_sym) = run_flow_graph(sync_sym, ref_sym, data_sym)
  File "/home/marcus/src/gnuradio/gr-digital/python/digital/qa_ofdm_chanest_vcvc.py", line 253, in run_flow_graph
    if shifted_carrier_mask[i] and channel_est[i]:
TypeError: 'NoneType' object is not subscriptable

```

feels super unrelated to this patch, but it only fails with it.

**Comment by marcusmueller** [comment]
testing reveals that in fact the `add_cc` never emits data in this test case. So, that's a bug in my PR :(

**Comment by marcusmueller** [comment]
@mbr0wn  shot in the blue, maybe you've got the more alert eyes: I basically just took your approach from #986; where did I go wrong?

**Comment by marcusmueller** [comment]
Huh! If I send twice the data through the test case, it does in fact work. Why?

**Review by noc0lour — CHANGES_REQUESTED** [review]
I'd suggest using C++ style casts instead of C-style casts. Thanks for commenting the code :+1:

**Comment by marcusmueller** [comment]
> I'd suggest using C++ style casts instead of C-style casts. 

Good call! Done.

> Thanks for commenting the code +1

Needed to do that myself, honestly, because `git add -i` is sometimes being a bit difficult w.r.t. splitting hunks, and I really don't like `git add -e`.

**Comment by marcusmueller** [comment]
Took me a while to realize that this might be a bit of a bug in the scheduler, but I'm not sure about that (and if I'm not sure...). 

Point is: the `ofdm_chanest` test case relied on feeding a whole of 3 consecutive 32 sample vectors through add into a the channel estimator, which has its own multiplicity constraints. With the optimized add block advising a preference for 8-complexes alignment, the scheduler couldn't reconciliate these requirements, and thus, the flow graph was done before the estimator even ran. 

Takeaway:

* unit test unrealistic, since no channel will ever only be used for 96 samples in total
* reconsider `add_blk` having preferred alignment. I don't think the gains here are worth it.

**Comment by marcusmueller** [comment]
Really requesting @mbr0wn's review (or anyone else who's used the OFDM infrastructure personally), since I fiddled with its unit test to add flushing zeros.

File: gr-digital/python/digital/qa_ofdm_chanest_vcvc.py [hunk 1]
```diff
@@ -301,10 +301,11 @@ def test_999_all_at_once(self):
 
         def run_flow_graph(sync_sym1, sync_sym2, data_sym):
             top_block = gr.top_block()
-            carr_offset = random.randint(-max_offset / 2, max_offset / 2) * 2
+            carr_offset = random.randint(-max_offset // 2, max_offset // 2) * 2
             tx_data = shift_tuple(sync_sym1, carr_offset) + \
                 shift_tuple(sync_sym2, carr_offset) + \
-                shift_tuple(data_sym, carr_offset)
+                shift_tuple(data_sym, carr_offset) + \
+                tuple((0,) * fft_len)
```
mbr0wn, line 308: I'm a bit surprised by this. The first input vector to ofdm_chanest_vcvc should be the first sync symbol, in frequency domain. If it's all zero, then this should fail. Are we still losing data?

marcusmueller: That seems to be the case, on some runs, sadly, for me. Haven't rebuilt this branch lately, though.

marcusmueller:
This can be redone in a better way atop of #3615, so closing this

PR #3901, opened 2020-10-30
Fix examples that use wavfile_sink

argilo:
The recent libsndfile transition in #3600 changed the API of the `wavfile_sink` block, so the two examples that use it need to be updated.


drmpeg (approved):
LGTM. Thanks for catching this.

PR #3902, opened 2020-10-30
gr-blocks: Add example flowgraphs for Stream Mux/Demux blocks.

dave-pi:
Add a couple of example flowgraphs for these blocks. Usage is demonstrated in the wiki page of each block.


mormj:
Getting a missing block error in the stream_demux_demo.

image

If I replace it with the stream demux block that is in tree, everything works great:

image

Is your example reference a block from an OOT?

dave-pi:
> Getting a missing block error in the stream_demux_demo.
> 
> image
> 
> If I replace it with the stream demux block that is in tree, everything works great:
> 
> image
> 
> Is your example reference a block from an OOT?

oh shoot, yes. I'll replace it...

dave-pi:
I've fixed the flowgraph, if you could test that would be great. I'm having trouble with grc 3.9 segfaulting so haven't been able to test.

mormj (approved):
Tested and the example look good. Thanks!

PR #3903, opened 2020-10-30
Move FFT functions and blocks to templates

mormj:
This is a rebase and continuation of https://github.com/gnuradio/gnuradio/pull/3574

A couple of things I had to do differently/additionally:
1) The vfc blocks needed to keep a fixed float to complex in/out to match current behavior, regardless of `forward` flag.  Not sure why there is not a c2r block for reverse float fft, but I stuck with current behavior which mean changing some of the template specializations
2) Wrapped the fwd and rev blocks as fft_vcc with the parameter to keep compatibility at least in the python API (and not have to change all the grc bindings and qa tests)

TODO:
- add shift for the vfc blocks
- clean up the commits and squash


gr-fft/lib/fft_v_fftw.cc
```diff
@@ -0,0 +1,217 @@
+/* -*- c++ -*- */
+/*
+ * Copyright 2004,2007,2008,2010,2012 Free Software Foundation, Inc.
```
marcusmueller, line 3: Copyright year might deserve a bit of updating :)

marcusmueller (approved):

PR #3912, opened 2020-11-02
zeromq: Backport to 3.8: Add ZMQ_LINGER value to prevent infinite block

MattMills:
Backport of #3866 to gr 3.8

### DIFF
--- a/gr-zeromq/lib/base_impl.cc
+++ b/gr-zeromq/lib/base_impl.cc
@@ -28,6 +28,10 @@
 #include "tag_headers.h"
 #include <gnuradio/io_signature.h>
 
+namespace {
+constexpr int LINGER_DEFAULT = 1000; // 1 second.
+}
+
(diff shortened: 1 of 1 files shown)


willcode:
This seems low risk. Does anyone have an existing application that depends on one ZMQ endpoint hanging indefinitely until the other end empties a queue?

MattMills:
@willcode Should I just assume this is closed as wontfix?

dkozel:
Hi @MattMills sorry for the long delay. This wasn't followed up on after willcode's question. I don't think anyone knows of an application that relies on this behavior, and I think this is pretty clearly a bug so changing the behavior on maint is justified.

This is a small change and it's been merged on master for a while now without any new bug reports, so merging now.

MattMills:
Thanks!

willcode:
Already backported to maint-3.8, removing tag.

Issue #3894, opened 2020-10-28
Save recent opened flowgraphs before running a flowgraph


noc0lour:
Just a little annoyance if you have to kill GRC while it's executing a flowgraph the recent opened flowgraphs is reset to the last time gnuradio-companion was closed properly. From a UI point of view it would be nice to have it restored to the last possible moment and saving the recently opened graphs before running something would seem reasonable from a UX perspective.

PR #3913, opened 2020-11-02
grc: Save config before execution and update recent file sub-menu.

dave-pi:
Saves the grc prefs before executing a flowgraph to avoid losing them if
anything goes wrong and grc needs to be forcibly closed.

Also fixes recent files sub-menu refresh. When a file is opened, opened
from recent, or saved as, the recent files sub-menu is now updated.

Fixes: #3894


dave-pi:
There is still one use case that does not update the recent files menu, which is files opened directly from the command-line.

dl1ksv:
Works for me as expected

dl1ksv (approved):
Works for me as expected.

mbr0wn (approved):

PR #3915, opened 2020-11-03
Change "top_block" to "default" to find options_block in GRC create hier

jaredd:
The default name in the flowgraph options_block seems to have changed from "top_block" to "default". This PR changes the flowgraph block search term while creating a hierarchical block from the GRC UI. Prior to this change, hierarchical block creation would fail, failing to find the "top_block" block in the flow_graph.get_block method. 

This change enables creation of hierarchical blocks in the GRC interface again. There appears to be some need to refactor / rewrite a bit of the code that performs the creation (e.g. the rest of the code opts for flow_graph.options_block vs. flow_graph.get_block(...) ).

```bash
$ gnuradio-config-info --version
3.8.2.0
$ python3 --version
Python 3.8.6
$ 
```

### DIFF
--- a/grc/gui/Application.py
+++ b/grc/gui/Application.py
@@ -324,7 +324,7 @@ class Application(Gtk.Application):
             flow_graph.move_selected(coords)
 
             # Set flow graph to heir block type
-            top_block  = flow_graph.get_block("top_block")
+            top_block  = flow_graph.get_block("default")
             top_block.params['generate_options'].set_value('hb')
 
             # this needs to be a unique name


mbr0wn:
@jaredd should this also go on newer branches?

jaredd:
Well the bug is still in master, albeit on a different line:

https://github.com/gnuradio/gnuradio/blob/f788f65c7b5b52183a3d97fd8d3b17f9bd21c733/grc/gui/Application.py#L318

Try creating a hier block within GRC, see if it works. I don't have a 3.9 install.  So I would recommend the fix on all branches that are in use.

mbr0wn:
I just tried creating a hier block on master, and it works without this patch. Then I tried it with 3.8, but I still can't repro the issue.

mbr0wn:
Hm I'm going to close this. I just don't understand what this solves, or what the original issue is. Maybe we can find someone more GRC-savvy to chime in in the chat.

PR #3916, opened 2020-11-04
pmt: use to_string for char string_refs

### LINKED ISSUE #3398: GRC: handling of certain values within variables

mormj:
The current method of converting pmt char values to strings results in non-unicode characters - looks like it was using the pointer value in the string reference.  This just uses std::to_string on the char values

Fixes #3398

### DIFF
--- a/gnuradio-runtime/lib/pmt/pmt_unv.cc
+++ b/gnuradio-runtime/lib/pmt/pmt_unv.cc
@@ -140,7 +140,7 @@ uint8_t* u8vector_writable_elements(pmt_t vector, size_t& len)
 const std::string pmt_u8vector::string_ref(size_t k) const
 {
     const auto& ch = ref(k);
-    return std::string(&ch, &ch + 1);
+    return std::to_string(ch);
 }
 
 } /* namespace pmt */
@@ -263,7 +263,7 @@ int8_t* s8vector_writable_elements(pmt_t vector, size_t& len)
 const std::string pmt_s8vector::string_ref(size_t k) const
 {
     const auto& ch = ref(k);
-    return std::string(&ch, &ch + 1);
+    return std::to_string(ch);
 }
 
 } /* namespace pmt */


ThomasHabets:
But this won't return the same type of output, will it? E.g. it'll output `97` instead of `a`.

https://godbolt.org/z/bT44fh

mormj:
That is correct - this does change the functionality.  But if it is being used for uint8_t uniform vectors, returning ascii characters doesn't make much sense.  Seems like the stringification should return the decimal (or hex) values.  I'd be curious if anyone is using these string_ref method to show ascii strings.

ThomasHabets:
hmm… I'm looking at the commit history. Trying to understand the intention. This used to be:

```
return boost::lexical_cast<std::string, uint8_t>(ref(k));
```

I'm trying to get godbolt to include boost for some testing, but it's not cooperating. IIRC lexical_cast outputs `a`.

mormj:
So maybe I need to leave string_ref alone, and make another pmt function that does this and wraps into the `__str__` method?  The problem in #3398 is that the non-utf8 strings choke in the pybind11, whereas they printed garbage but didn't exception out with the swig bindings.

ThomasHabets:
So the place this breaks is in the QT code for writing to widgets, right? I at least want to make sure there's a way to print bytes to the terminal, since that's what I use the most. If that's broken unicode, then that's fine and terminals tend to not mind a lot.

I'm not too familiar with pybind11, but maybe yes.

But this is an annoying inconsistency: Treating uint8 as characters, but unit16 as ints. int8/char is special here. It can mean numbers (like uintXX_t), bytes (current implementation), or characters (and then we'd have to pick a charset to translate high bytes from).

mormj:
It is the inconsistency that bothers me about this.  My thought is that all the write_string methods for pmt uniform vectors should return a string of space separated numbers (like uintXX_t currently does), and characters or bytes should be handled by a different underlying method, and not get mapped to the python `__str__` binding.  Perhaps a `write_bytes` function which is valid for some pmts.

ThomasHabets:
Agreed on the problem. And we inherited it from C, really.

Your solution sounds reasonable. I'm not sure about which places this will change will show up. It runs the risk of breaking things even if it's the right thing to do. `write_bytes`, or just `bytes`.

But then again, maybe that's a reasonable risk to do before 3.9? I'd have to poke around a bit to see what specific observable behaviour this would break, and how it could be worked around.

ThomasHabets:
I can't think of a problem this PR could cause, and it doesn't seem to break message debug at least.

But I dunno. I don't feel confident that I know where this could cause problems, if anywhere. LGTM FWIW, but with that caveat.

mrjacobagilbert (approved):
I think this is a reasonable fix. In my mind there is nothing ascii-specific about u8vectors, and it addresses a major problem in GRC. It does change the runtime behavior of blocks that print PMTs, but since 3.9 is still not released that should be acceptable. Processing code should be unaffected, this call is really only accessed when moving data out of GR or presenting something to a user.

Previously printing `pmt.init_u8vector(4,[0,49,50,244])` from a debug block would look like:
```
#[ 1 2 �]
```
and now it looks like:
```
#[0 49 50 244]
```
Similar updates to GRC displays fixing the crash. LGTM.

Issue #3877, opened 2020-10-21
Frequency Xlating FIR Filter block is not phase-continuous


daniestevez:
The Frequency Xltating FIR Filter block is not phase-continuous when the frequency is updated. The prototypical use case where this causes problems is when it's used to correct the the Doppler shift of a PSK signal.

The cause is that [`freq_xlating_fir_filter_impl::build_composite_fir()`](https://github.com/gnuradio/gnuradio/blob/912d48596569812ddcd05539bc7086617971b65e/gr-filter/lib/freq_xlating_fir_filter_impl.cc#L65`) doesn't take into account the current phase of the rotator `d_r` when rebuilding the taps of the FIR filter after a frequency change. I think that it can be fixed by multiplyin the taps by a suitable complex exponential that depends on the phase of `d_r` (I didn't think of the details, but I think it's easy to write the formula).

The [`rotator`](https://github.com/gnuradio/gnuradio/blob/912d48596569812ddcd05539bc7086617971b65e/gr-blocks/include/gnuradio/blocks/rotator.h#L22) doesn't have a method to get its current phase. We could do `d_r.rotate(gr_complex(1.0, 0.0))`, and then `d_r.set_phase()` with what we get (since calling `rotate()` increases the phase, which we don't want to do here). However, it seems much cleaner to add a `get_phase()` method to the rotator.

I'll probably send a PR with this in a few days/weeks, unless someone says here that they want to do it themselves.

willcode:
@daniestevez Is this possible? Even though we know the phase of the rotator, the cumulative phase shift due to the spun taps is unknown. So, we don't know what the effect of respinning the taps would be and can't correct for it. Am I missing something?

willcode:
Since the passband changes when the taps are respun, there's not even a guarantee that the same signals are being processed by the filter.

marcusmueller:
This sounds like a good idea to get right. However, why modify the phase of the filter if we could as well modify phase of the rotator?

willcode:
@marcusmueller That would be the way to do it ... but the change in phase during retune depends on the phase response of the filter. Even if the phase response is linear, it is unknown. I wasn't able to work out the math to make this work. Could just be my math :-)

daniestevez:
Hi! Yes, the idea to change the phase of the filter was sloppy on my part. It's much better to change the phase of the rotator. I bet we can derive a formula for this. I'll try to work out the math or some example.

[...]

PR #3917, opened 2020-11-04
Phase continuity for freq_xlating_fir_filter

willcode:
Retuning the freq_xlating_fir_filter result in arbitrary phase jumps, which can affect phase-sensitive demodulators.

https://github.com/gnuradio/gnuradio/issues/3877

This PR
- adds rotator::phase() to allow access to the internal state of a rotator object
- adjusts the rotator used by the xlating filter to keep phase constant during retune

The algorithm here works for FIR filters with a constant phase delay over the passband, and having an odd number of taps.

It is unlikely that anyone depends on the current behavior (arbitrary phase jump).

Tested using https://gist.github.com/willcode/f32e905bc268eeebd0cf4cb95bb4b4f1.

### DIFF
--- a/gr-blocks/include/gnuradio/blocks/rotator.h
+++ b/gr-blocks/include/gnuradio/blocks/rotator.h
@@ -29,6 +29,7 @@ private:
 public:
     rotator() : d_phase(1), d_phase_incr(1), d_counter(0) {}
 
+    gr_complex phase() { return d_phase; }
     void set_phase(gr_complex phase) { d_phase = phase / std::abs(phase); }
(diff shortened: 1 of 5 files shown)


willcode:
Of course the QA code depends on the absolute phase.

mormj (approved):
LGTM

argilo (approved):
This does appear to get rid of phase discontinuities when I test with a symmetric FIR filter.

mormj (approved):

mormj:
@willcode - is this one good to go now?

willcode:
@mormj yes good to go.

PR #3924, opened 2020-11-08
gr-filter: Remove unused parameter in polyphase_filterbank.

drmpeg:


willcode (approved):
Variable doesn't appear anywhere else in the codebase, and it doesn't make any sense here anyway.

mormj:
I think leaving it was an incomplete attempt at satisfying: https://github.com/gnuradio/gnuradio/pull/3574#issuecomment-716691299

Is there any scenario where the polyphase filterbank would need a forward fft (right now the implementation is fixed to ifft)

willcode:
polyphase_filterbank seems like a fairly special-purpose class used to factor some logic out of GR pfb filters. It always sets up fir and fft taps, always creates a fft_rev. Would this ever be used for something other than despinning? I'm not a polyphase expert, but I think the direction of the spin is fixed by the partitioning in set_taps.

mormj (approved):

PR #3925, opened 2020-11-08
RPC: fixes to allow using Thrift 0.9 through 0.12

marcusmueller:
supersedes #2739.

The reason to do it this way instead of going @michaelld's route (sorry!) is that it removes the need to export the Thrift version that was used at build time. Since I guess that thrift will still be a source of version-specific joy in the future, having code that we can reuse directly on master to that allows compile-time checking of a numerical Thrift version will probably come in handy too, and this can be more or less directly cherry-picked onto master.


marcusmueller:
The brunt of making this work has still be done by @michaelld's hands – I only had to adapt the used type/define.

michaelld (approved):
Nice! LTGM!

michaelld:
I might have started the work, but this is a huge improvement over what I did. Thanks @marcusmueller !

PR #3926, opened 2020-11-09
modtool: Change handled exception to ModuleNotFoundError

### LINKED ISSUE #3884: Pybind: Module loading failures get masked in the init
If there is some linker error, or another issue with the pybind in an OOT, this gets masked because of 

```
except ImportError:
    pass
```
in `__init__.py`

There are some situations where the ImportErrors should be caught and ignored, such as a python only module (which is why the pass is there).  But in other situations, the underlying error should be displayed

mormj:
Change the pass on exception to pass only on ModuleNotFoundError, not any ImportError

This should make the behavior as follows:
1) Python only module should import without error (by passing at ModuleNotFoundError)
2) C++ module with missing symbol should be caught and displayed

Tested 2) by creating a block with a dummy function in the header with no implementation.  Then call that implementation from another function.  Compiles ok, but will cause undefined symbol error at import.

Fixes #3884

### DIFF
--- a/gr-utils/modtool/templates/gr-newmod/python/__init__.py
+++ b/gr-utils/modtool/templates/gr-newmod/python/__init__.py
@@ -16,7 +16,7 @@ import os
 try:
     # this might fail if the module is python-only
     from .howto_python import *
-except ImportError:
+except ModuleNotFoundError:
     pass
 
 # import any pure python here


drmpeg (approved):
Tested with gr-paint OOT. Error message before properly linking with FFT component looks good.
```
  File "/home/re/xfer/paint_tx.py", line 40, in <module>
    import paint
  File "/opt/gnuradio-3.9.0git/lib/python3/dist-packages/paint/__init__.py", line 18, in <module>
    from .paint_python import *
ImportError: /opt/gnuradio-3.9.0git/lib/libgnuradio-paint.so.1.0.0git: undefined symbol: _ZN2gr3fft3fftISt7complexIfELb0EEC1Eii
```

Issue #3921, opened 2020-11-07
`wfm_rcv_pll` constructor has `audio_decimation` parameter, but doesn't use it


marcusmueller:
... and the docstring is inappropriate too, therefore.

Either fix by removing argument, or implement decimation.

Probably just replacing the `1` in 

```
        self.fft_filter_xxx_3 = filter.fft_filter_fff(1, stereo_carrier_filter_coeffs_0, 1)
```

with `audio_decimation` + writing a test case.

marcusmueller:
Also, that flow graph could have used some more descriptive (and shorter) block naming!

dave-pi:
`demod_rate` is also unused, effectively forcing the input sample rate to 3840000.

Issue #2799, opened 2019-09-12
Unexpected behavior of grcc with --output for hier blocks


bgottula:
The behavior of the `--output` option of `grcc` is inconsistent with the documentation when `grcc` is run on a GRC hier block. Specifically, even when a path is supplied to `--output`, `grcc` ignores the path provided and proceeds to write the generated .py and .yml files to HIER_BLOCK_LIB_DIR (~/.grc_gnuradio by default) anyway.

Usage info for `--output` does not mention any exceptions:
```
  -o DIR, --output DIR  Output directory for compiled program [default=.]
```

The source code is also misleading. A `file_path` is provided to the constructor of `HierBlockGenerator`, with [doctring](https://github.com/gnuradio/gnuradio/blob/master/grc/core/generator/hier_block.py#L22)
```
file_path: where to write the py file (the yml goes into HIER_BLOCK_LIB_DIR)
```
But the source code of the constructor actually clobbers `file_path` with `hier_block_lib_dir` on [this line](https://github.com/gnuradio/gnuradio/blob/master/grc/core/generator/hier_block.py#L32):
```
self.file_path = os.path.join(hier_block_lib_dir, self._flow_graph.get_option('id') + '.py')
```

Personally I'd prefer to be able to set the output path of GRC for both the .py file and the .yml file with command line arguments as an alternative to using the HIER_BLOCK_LIB_DIR environment variable. But even if this is not possible the documentation should be updated to clarify the fact that `--output` is ignored for hier blocks.

jaredd:
Yes, I just ran into this myself on the maint-3.8 branch. It appears this is still an issue on master. One commit that appears to be part of the issue is f25f0c573a519babd3bc6099230177cdd022ecdd which removes the `file_path` parameter in [HierBlockGenerator](https://github.com/gnuradio/gnuradio/commit/f25f0c573a519babd3bc6099230177cdd022ecdd#diff-7656238266b046810186088ee826bacbbc2ffbbb7571ef3e3f49bb21744dfef2L12-L22). That commit seems to have to do with embedded python files.

For maint-3.8 the issue is different, but in a similar vein: the file_path passed to the HierBlockGenerator is ignored in lieu of the `platform.config.hier_block_lib_dir` variable which can be set, confusingly, by the GRC_HIER_PATH env var (perhaps this changed from when the OP looked at this, the docstring is also wrong). And as mentioned in the OP, yml file is _always_ put in that `hier_block_lib_dir`.

[...]

PR #3927, opened 2020-11-10
gr-analog: Fix wfm_rcv_pll arguments

dave-pi:
Fixes: #3921 

There seems to be an additional problem which is the modulation seems off compared to the 3.8 block implementation. If I run a tone through a pair if WBFM tx and rcv_pll it comes out very dirty with weird harmonics. An identical flowgraph in 3.8 doesn't do that. I'll see what I can find out.

### DIFF
--- a/gr-analog/python/analog/wfm_rcv_pll.py
+++ b/gr-analog/python/analog/wfm_rcv_pll.py
@@ -42,57 +42,56 @@ class wfm_rcv_pll(gr.hier_block2):
         ##################################################
         self.demod_rate = demod_rate
         self.deemph_tau = deemph_tau
-        self.stereo_carrier_filter_coeffs_0 = stereo_carrier_filter_coeffs_0 = firdes.band_pass(-2.0, demod_rate, 37600, 38400, 400, fft.window.WIN_HAMMING, 6.76)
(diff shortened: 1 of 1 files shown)


duggabe:
might this be related to https://github.com/gnuradio/gnuradio/issues/3593 ?

dave-pi:
> might this be related to #3593 ?

I'm not sure what the problem is yet. This PR has enough changes to fix the issue which it addresses, so I'll open another one if I find something wrt to this difference in behaviour between 3.8 and 3.9.

ValZapod:
I saw you comments on mailing list @duggabe, https://lists.gnu.org/archive/html/discuss-gnuradio/2020-06/msg00117.html and etc, just wanted to say that there is a nice implementation of IMHO perfect stereo FM receiver/transmitter [here](https://github.com/bastibl/gr-rds/pull/17), and because it also can extract digital info from analog stream (RDS), I beleive it should be even better than other alternatives that you recently implemented in #3593. It is insane compexity and has wonderful example of triggering this error #2799.

mbr0wn:
@dave-pi Can you see if this is still relevant? If it is, it requires a rebase. Thanks!

dave-pi:
The arguments were fixed by #3991 so all that's left is renaming the blocks from auto-generated to sensible names.
I've uploaded a change with that. I tested by demodulating before and after the changes and diffing the resulting .wav files to make sure they were identical. Input file used for testing is attached.
[fm_l_384k.zip](https://github.com/gnuradio/gnuradio/files/5838625/fm_l_384k.zip)

mbr0wn (approved):

PR #3928, opened 2020-11-10
Remove warnings with clang compiler build

drmpeg:


drmpeg:
The full move assignment warning:
```
/home/re/xfer/gnuradio/gr-filter/lib/../include/gnuradio/filter/fir_filter_with_buffer.h:56:33: warning: explicitly defaulted move assignment operator is implicitly deleted [-Wdefaulted-function-deleted]
    fir_filter_with_buffer_fff& operator=(fir_filter_with_buffer_fff&&) = default;
                                ^
/home/re/xfer/gnuradio/gr-filter/lib/../include/gnuradio/filter/fir_filter_with_buffer.h:31:15: note: move assignment operator of 'fir_filter_with_buffer_fff' is implicitly deleted because field 'd_align' is of const-qualified type 'const int'
    const int d_align;
              ^
```

ThomasHabets:
On the `const` parts: We have to choose between move cons/assign or const.

Seems that these things sometimes get put into `vectors`, so we should probably go with allowing move.

ThomasHabets (approved):

mbr0wn (approved):

PR #3929, opened 2020-11-11
Resize the windows properly

### LINKED ISSUE #3865: QT GUI Sink display anomaly
When a QT GUI Sink is opened, the Frequency Display tab is selected, but the plot only occupies about 1/2 of the window width. As soon as the window is resized or another tab is selected, going back to the Frequency Display displays properly.

See ~/gnuradio/gr-qtgui/examples/gui_sink.grc

dl1ksv:
There a some problems in resizing the windows.
Fixes #3865

### DIFF
--- a/gr-qtgui/include/gnuradio/qtgui/displayform.h
+++ b/gr-qtgui/include/gnuradio/qtgui/displayform.h
@@ -41,7 +41,6 @@ public:
     void enableMenu(bool en = true);
 
 public slots:
-    void resizeEvent(QResizeEvent* e) override;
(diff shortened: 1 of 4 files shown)


mormj (approved):
Tested on ubuntu 20 and indeed does fix the issue

drmpeg (approved):
Works good here on Ubuntu 18.04/arm7. I'm going to approve this, but I'm not a big fan of commenting out code (in displayform.*). Just delete it. That's what version control is for.

dl1ksv:
You are right. Code removed

Issue #3816, opened 2020-10-07
GRC: Hidden block ports are always shown


mrjacobagilbert:
In both 3.8 and 3.9, GRC has stopped respecting the 'hidden' field of ports and they are always shown. This can clutter flowgraphs considerably and can make some blocks harder to use as a common reason to hide ports is to keep advanced or complicated features out of the way.

dave-pi:
The problem seems to be related to the bus toggle logic. `update_bus_logic()` is called as part of the block initialization but, when disabling a bus, it doesn't check if it's been previously enabled. This causes it to restore `port.hidden` with a `None` which ultimately is converted to `False`. So it never hides the message ports.
I think there's an additional problem in `form_bus_structure()` which doesn't filter message ports (I'm assuming they are not to be part of buses).
I'll upload a PR with some proposed changes that seem to fix this. I'm not 100% certain on the bus logic functionality, the wiki seems to be outdated (the flowgraph screenshot shows some blocks like 'Bus Source Structure' which I couldn't find).

PR #3934, opened 2020-11-12
grc: Fix port show/hide problem due to bus logic update

dave-pi:
update_bus_logic() is called when disabling a bus, it doesn't check if it's been previously enabled. This causes it to restore port.hidden with a None which ultimately is converted to False. So it never hides the ports.

Also, filter message ports from buses.

Fixes: #3816


drmpeg (approved):
Works good here. Tested with Qt GUI blocks and selector block.

PR #3935, opened 2020-11-12
Update parameter background colours realtime

### LINKED ISSUE #3487: Parameter background colours in Param Editor are not updated in realtime
Changing the type of a value in the grc parameter editor does not change the background colour of the value entry . You have to close and reopen the editor to get the correct background colour in the value entry.

dl1ksv:
If you change the type of a variable in the Properties Parameter editor the background colour of the parameter value is not updated.
You have to reload the editor for this parameter block to see the background colour according to the parameter type.

Fixes #3487 

This happens because the value box is evaluated before the type box.
Changing the order of the entries to type before value in the parameter.block.yml file fixes this issue.

### DIFF
--- a/grc/blocks/parameter.block.yml
+++ b/grc/blocks/parameter.block.yml
@@ -7,10 +7,6 @@ parameters:
     label: Label
     dtype: string
     hide: ${ ('none' if label else 'part') }
--   id: value
-    label: Value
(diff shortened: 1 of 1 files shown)


mormj (approved):
Tested and works.  LGTM

PR #3936, opened 2020-11-12
move msgq_runner out of runtime into gr-uhd/apps, its single consumer

marcusmueller:
This removes the last file in the gru module, thus addressing the last issue in #1591


argilo:
```
$ uhd_rx_nogui 
  File "/home/argilo/prefix_39/bin/uhd_rx_nogui", line 59, in <module>
    import msgq_runner
ModuleNotFoundError: No module named 'msgq_runner'
```

argilo:
I was able to fix that by removing the `import msgq_runner` lines and changing `msgq_runner.msgq_runner` to `uhd.msgq_runner`.

argilo:
@mormj I think the import cleanup needs to be applied to `uhd_app.py` as well.

argilo (approved):

mormj (approved):

Issue #2719, opened 2019-08-10
Identify desirable dependency version increases


marcusmueller:
GR 3.9 can rely on whatever is state of the art in 2020.

Main Candidates, from the top of my head

* [x] GCC/Clang/MSVC
  * is a C++ version increment to C++14 possible? How far are we from C++17?
* [ ] Boost
* [ ] CMake
* [x] SWIG

jdemel:
I assume those dependency versions depend on distros we'd like to support. Same goes for MacOS and Windows (probably Win10 only?) I'd require those distros to still receive support by whomever works on them. e.g. Ubuntu 14.04 is not a target.
So we should name those distros. I assume Fedora/RedHat/CentOS but which versions?
Ubuntu should be on the list. I suggest 18.04 and 20.04. Other versions are non-LTS or too old. 16.04 is debatable at this point. 

C++17 might be usable with GCC9 in Ubuntu 20.04 but has incomplete support on Ubuntu 18.04 with GCC 7.

jdemel:
We move from SWIG to PyBIND11. Thus, we'll drop the SWIG dependency entirely and depend on PyBIND11 instead. I suggest at least PyBIND11 2.2 because there they changed how to add a module.

marcusmueller:
re: thrift: #3920

PR #3937, opened 2020-11-12
Bump C++ standard to C++14

marcusmueller:
This ties in with #2719. Doing it now ensures that maint-3.9 can use C++14.

At the very beginning of the 3.10 dev cycle, we should check whether we can go for C++17, so that we don't slow down modernization (like we, sadly, probably done with Thomas' work on current master due to being late with switching to the next C++ standard).


drmpeg:
There is no C14 standard.

PR #3938, opened 2020-11-13
gnuradio-runtime: Fix qa_fast_atan2f test code.

drmpeg:


drmpeg:
Test t1 has been broken since day one. The inner loop has never executed.

mormj (approved):
LGTM!

mormj:
Makes sense that the test never ran properly.  CI passes.  Merging.

PR #3943, opened 2020-11-15
RPC: fixes to allow using Thrift 0.13

maitbot:
Adds ability for maint-3.8 to build with newer Thrift versions
by checking availability of thrift include files.
(Also spelling fix: Thift -> Thrift)


noc0lour:
I think the CI has no installed thrift? Thus the CI green flag is a false positive. Need to see if we can integrate a functional Thrift installation into the image finally ...

mbr0wn:
We already have this patch on master... so it can go here too.

Issue #2512, opened 2019-05-28
qtgui: uhd_display and usrp_display_qtgui PyQt4 dependencies


mormj:
Like several other components, these are still running on PyQt4 and should be upgraded or removed.

marcusmueller:
Done!

PR #3945, opened 2020-11-15
more qt4 qt5 updates

maitbot:
Fixes: qtgui: uhd_display and usrp_display_qtgui PyQt4 dependencies #2512

This is one of my patches for Debian packaging of the maint-3.8 branch.


marcusmueller:
Looks good to me; we'll need to update FindQwt.cmake, too, to reject Qwt < 6.1.

Do we still know why we don't allow QWT >= 6.2.0?

marcusmueller:
Oh wait! I thought this was against master! The one against master has nearly the same number (4<->5 at the end)

Can't bump deps on a released branch; we'll need to be slightly finer in granularity.

PR #3949, opened 2020-11-16
grc: fix gnome-terminal deprecation warning

mormj:
When running a "No GUI" flowgraph from GRC, it gives the following deprecation warning:
```
    Option “-e” is deprecated and might be removed in a later version of gnome-terminal.
    Use “-- ” to terminate the options and put the command line to execute after it.
```
This follows the suggested fix to update the deprecation

### DIFF
--- a/grc/gui/Executor.py
+++ b/grc/gui/Executor.py
@@ -56,7 +56,10 @@ class ExecFlowGraphThread(threading.Thread):
         # When in no gui mode on linux, use a graphical terminal (looks nice)
         xterm_executable = find_executable(self.xterm_executable)
         if generator.generate_options == 'no_gui' and xterm_executable:
-            run_command_args = [xterm_executable, '-e', run_command]
+            if ('gnome-terminal' in xterm_executable):
+                run_command_args = [xterm_executable, '--'] + run_command_args
+            else:
+                run_command_args = [xterm_executable, '-e', run_command]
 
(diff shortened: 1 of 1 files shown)


drmpeg:
Why is that modtool commit from https://github.com/gnuradio/gnuradio/pull/3926 mixed in?

mormj:
dirty workspace.  fixing momentarily

willcode:
If `xterm_exectutable` were actually `xterm` ... `xterm: bad command line option "--"`

mormj:
Guess I’ll have to put a bit more logic in there

grc/gui/Executor.py
```diff
@@ -56,7 +56,10 @@ def _popen(self):
         # When in no gui mode on linux, use a graphical terminal (looks nice)
         xterm_executable = find_executable(self.xterm_executable)
         if generator.generate_options == 'no_gui' and xterm_executable:
-            run_command_args = [xterm_executable, '-e', run_command]
+            if ('gnome-terminal' in xterm_executable):
+                run_command_args = [xterm_executable, '--'] + run_command_args
```
willcode, line 60: Do you need run_command somewhere on this line?

mormj: run_command is the non-splitted version of run_command_args.  When calling gnome-terminal with `--` it needs the splitted version due to where the quotation marks get put

willcode (approved):

Issue #3798, opened 2020-10-03
NBFM Receive block causes audio underruns


duggabe:
The NBFM Receive block seems to not work properly in version 3.9.0.0-git as of 26 Aug 2020.  When I receive WBFM stereo, there is no problem. But using the same basic flowgraph, tuning to the 144MHz band, and substituting a NBFM Receive block, or an FM Demod block, I get LOTS of audio under-runs and loud noise.  It seems to be 3.9 specific because in 3.9, test2 **FAILS** with lots of aU's; test4 **PASSES** with no aU's and good signals, even though they are ostensibly equivalent.

My setup is:
Ubuntu 20.04.1
UHD 4.0 built from source
GR/GRC 3.9.0.0-git built from source 26 Aug 2020
Lenovo ThinkCentre AMD® Ryzen 7 PRO 3700 Processor
USRP B200-mini

Tests I have done:
1) extend tuning range of WBFM stereo; add squelch - **PASS** in 3.9, almost no aU's
2) replace WBFM Receive PLL block with NBFM Receive block - **FAIL** in 3.9, lots of aU's
3) replace NBFM Receive block block with FM Demod block - **FAIL** in 3.9, lots of aU's
4) replace FM Demod block with a Quadrature Demod and FM Deemphasis - **PASS** in 3.8 + 3.9, almost no aU's
5) in test 2, replace FFT Low Pass Filter with an FFT Filter - **FAIL** in 3.9 but **PASS** in 3.8

Notes that the FFT Low Pass Filter is new in 3.9

I revised the filter parameters for NBFM and revised the order of the blocks in test4 to yield the following:
The NBFM Receive block (in test2) is equivalent to the Quadrature Demod, FM Deemphasis, and FFT Filter in test4. See https://github.com/gnuradio/gnuradio/blob/master/gr-analog/python/analog/nbfm_rx.py

USRP_NFM_test2.grc https://pastebin.com/R0wavmAy
USRP_NFM_test3.grc https://pastebin.com/imRjyWEc
USRP_NFM_test4.grc https://pastebin.com/tiFikkp3
USRP_NFM_test5.grc https://pastebin.com/iwGbSSJV

duggabe:
To summarize the issue above:
test2 FAILS in 3.9
test3 FAILS in 3.9
test4 PASSES in 3.9 and 3.8
test5 FAILS in 3.9 and PASSES in 3.8

My conclusion: 
[...]

PR #3955, opened 2020-11-21
audio: raise nperiods for alsa source/sink

duggabe:
raise nperiods to 32
resolves https://github.com/gnuradio/gnuradio/issues/3798

### DIFF
--- a/gr-audio/lib/alsa/alsa_sink.cc
+++ b/gr-audio/lib/alsa/alsa_sink.cc
@@ -55,7 +55,7 @@ static double default_period_time()
 
 static int default_nperiods()
 {
-    return std::max(2L, prefs::singleton()->get_long("audio_alsa", "nperiods", 4));
+    return std::max(2L, prefs::singleton()->get_long("audio_alsa", "nperiods", 32));
 }
 
 // ----------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


argilo:
I think this is a good change. Audio underruns caused by the ALSA sink's very short default buffer size have been a common source of frustration for me. The only downside is that the default audio latency will increase by 140 milliseconds, but anyone who wants low latency can override this in their `~/.gnuradio/config.conf` file.

If this change is made, we should probably also update these defaults:

https://github.com/gnuradio/gnuradio/blob/4675ce0579ba99b9ebe2fce9ac37e5420f325ee6/gr-audio/lib/alsa/alsa_sink.cc#L56-L59

https://github.com/gnuradio/gnuradio/blob/4675ce0579ba99b9ebe2fce9ac37e5420f325ee6/gr-audio/lib/alsa/alsa_source.cc#L55-L58

duggabe:
I changed the source and sink files, but now I'm curious. How does the value in gr-audio-alsa.conf get used, when these are hard-coded in the .cc files?

argilo:
> How does the value in gr-audio-alsa.conf get used, when these are hard-coded in the .cc files?

`prefs::singleton()->get_long("audio_alsa", "nperiods", 32)` reads from the config file, and falls back to 32 if `nperiods` is not present in the `audio_alsa` section of the config file.

duggabe:
Thank you. Is this ready to go now?

noc0lour (approved):

PR #3958, opened 2020-11-23
Qt compass

duggabe:
qtgui_compass.block.yml contains parameters in the "documentation" section which are not used in the implementation.


duggabe:
done in error.

PR #3959, opened 2020-11-23
gr-qtgui: remove erroneous parameters in compass.block.yml

duggabe:
there are parameters in the 'documentation' section of the yml file which do not match the implementation


ghostop14:
LGTM.  Thanks!

PR #3965, opened 2020-11-24
Do not disable boost, if only the unit_test_framework is missing

dl1ksv:
Boost will be disabled even all required components besides unit_test_framework are available.
It would suffice to  disable testing.


mormj (approved):
Seems like a reasonable combo of options

noc0lour (approved):
Really good change, `GR_REGISTER_COMPONENT` first checks if it is user-set and if yes, Boost utf will be a hard requirement otherwise testing will not be enabled. Thanks!

PR #3966, opened 2020-11-24
fir_filter: remove unused param

willcode:
The 'decimation' parameter in the fir_filter constructor is unused. A number of other classes appear to think this parameter does something, and it is set in various places to '1', 'sps', 'decim'. Tests pass with the parameter removed.


mormj (approved):
Seems like a sensible thing to remove if it is unused in the fir_filter class itself.  Was it ever used?

willcode:
@mormj The fir_filters were created on 2012-05-01 and decimation was passed through to the gr_sync_decimator. The following day, the code was refactored and the parameter has been ignored ever since.

PR #3968, opened 2020-11-27
gr-digital: improved performance and capability to 63-bit registers. …

ewxl:
gr-digital: improved performance and capability to 64-bit registers (up to polynomials of degree 63). Now uses __builtin_parityl (a compiler built-in function?) to check for parity to determine the next bit.

I would also like to change the way masks are entered in GRC, it is quite easy to convert from exponent list to integer register, as can be seen in the qa_scramblers.py: make_mask-function.

I do have concerns, I'm not sure this will work on a raspberry pie, can I get help to test this?

### DIFF
--- a/gr-digital/include/gnuradio/digital/lfsr.h
+++ b/gr-digital/include/gnuradio/digital/lfsr.h
@@ -37,7 +37,7 @@ namespace digital {
  *
  * \details
  * Generates a maximal length pseudo-random sequence of length
- * 2^degree-1
+ * 2^degree-1, if supplied with a primitive polynomial.
  *
  * Constructor: digital::lfsr(int mask, int seed, int reg_len);
  *
@@ -47,9 +47,9 @@ namespace digital {
  *             order bit.
  *
  *             Some common masks might be:
- *              x^4 + x^3 + x^0 = 0x19
- *              x^5 + x^3 + x^0 = 0x29
- *              x^6 + x^5 + x^0 = 0x61
+ *              x^4 + x^3 + x^0 = 0x19, K=3
+ *              x^5 + x^3 + x^0 = 0x29, K=4
+ *              x^6 + x^5 + x^0 = 0x61, K=5
  *
  * \param seed - the initialization vector placed into the
  *             register during initialization. Low order bit
@@ -90,32 +90,26 @@ namespace digital {
 class lfsr
 {
 private:
-    uint32_t d_shift_register;
-    uint32_t d_mask;
-    uint32_t d_seed;
-    uint32_t d_shift_register_length; // less than 32
-
-    static uint32_t popCount(uint32_t x)
-    {
-        uint32_t r = x - ((x >> 1) & 033333333333) - ((x >> 2) & 011111111111);
-        return ((r + (r >> 3)) & 030707070707) % 63;
-    }
+    uint64_t d_shift_register;
+    uint64_t d_mask;
+    uint64_t d_seed;
+    uint64_t d_shift_register_length; // less than 32
 
 public:
-    lfsr(uint32_t mask, uint32_t seed, uint32_t reg_len)
+    lfsr(uint64_t mask, uint64_t seed, uint64_t reg_len)
         : d_shift_register(seed),
           d_mask(mask),
           d_seed(seed),
           d_shift_register_length(reg_len)
     {
-        if (reg_len > 31)
-            throw std::invalid_argument("reg_len must be <= 31");
+        if (reg_len > 63)
+            throw std::invalid_argument("reg_len must be <= 63");
     }
 
     unsigned char next_bit()
     {
         unsigned char output = d_shift_register & 1;
-        unsigned char newbit = popCount(d_shift_register & d_mask) % 2;
+        unsigned char newbit = __builtin_parityl(d_shift_register & d_mask);
         d_shift_register =
             ((d_shift_register >> 1) | (newbit << d_shift_register_length));
         return output;
@@ -124,7 +118,7 @@ public:
     unsigned char next_bit_scramble(unsigned char input)
     {
         unsigned char output = d_shift_register & 1;
-        unsigned char newbit = (popCount(d_shift_register & d_mask) % 2) ^ (input & 1);
+        uint64_t newbit = __builtin_parityl(d_shift_register & d_mask) ^ (input & 1);
         d_shift_register =
             ((d_shift_register >> 1) | (newbit << d_shift_register_length));
         return output;
@@ -132,8 +126,8 @@ public:
 
     unsigned char next_bit_descramble(unsigned char input)
     {
-        unsigned char output = (popCount(d_shift_register & d_mask) % 2) ^ (input & 1);
-        unsigned char newbit = input & 1;
+        unsigned char output = __builtin_parityl(d_shift_register & d_mask) ^ (input & 1);
+        uint64_t newbit = input & 1;
         d_shift_register =
             ((d_shift_register >> 1) | (newbit << d_shift_register_length));
         return output;
@@ -155,7 +149,7 @@ public:
         }
     }
 
-    int mask() const { return d_mask; }
+    uint64_t mask() const { return d_mask; }
 };
 
 } /* namespace digital */
--- a/gr-digital/python/digital/qa_scrambler.py
+++ b/gr-digital/python/digital/qa_scrambler.py
@@ -20,8 +20,8 @@
 # Boston, MA 02110-1301, USA.
 #
 
-
 from gnuradio import gr, gr_unittest, digital, blocks
+import numpy as np
 import pmt
 
 # See gr-digital/lib/additive_scrambler_bb_impl.cc for reference.
@@ -35,6 +35,10 @@ def additive_scramble_lfsr(mask, seed, reglen, bpb, data):
         out.append(d ^ scramble_word)
     return tuple(out)
 
+def make_mask(*exp):
+    from functools import reduce
+    return reduce(int.__xor__,map(lambda x:2**x,exp)),max(exp)-1
+
 class test_scrambler(gr_unittest.TestCase):
 
     def setUp(self):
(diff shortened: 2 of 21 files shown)


willcode:
I don't have GR set up on a RPi, but here's the code generated by a test of `__builtin_parity()` and `__builtin_parityl()` on a RPi3. They both call `__paritysi2()` which looks no less efficient than calling `popCount()`.

```
000104ec <xl>:
   104ec:       e92d4010        push    {r4, lr}
   104f0:       eb000003        bl      10504 <__paritysi2>
   104f4:       e8bd8010        pop     {r4, pc}

000104f8 <x>:
   104f8:       e92d4010        push    {r4, lr}
   104fc:       eb000000        bl      10504 <__paritysi2>
   10500:       e8bd8010        pop     {r4, pc}

00010504 <__paritysi2>:
   10504:       e0200820        eor     r0, r0, r0, lsr #16
   10508:       e59f3014        ldr     r3, [pc, #20]   ; 10524 <__paritysi2+0x20>
   1050c:       e0200420        eor     r0, r0, r0, lsr #8
   10510:       e0200220        eor     r0, r0, r0, lsr #4
   10514:       e200000f        and     r0, r0, #15
   10518:       e1a00053        asr     r0, r3, r0
   1051c:       e2000001        and     r0, r0, #1
   10520:       e12fff1e        bx      lr
   10524:       00006996        .word   0x00006996
```
On x86_64, the `setnp` opcode is used.

willcode:
@ewxl what performance gain did you see?

ewxl:
Thank you for the comment on RPi, that means it should not be affected of this change.

Here are some stats from my system:

Test-code:

    def test_scrambler_descrambler_big():
        mask,k = make_mask(31,26,15,7,0) 
        src_data = np.random.randint(0,2,512,dtype=np.int8)
        src = blocks.vector_source_b(src_data, True)
        head = blocks.head(1,100000000)
        scrambler = digital.scrambler_bb(mask, 1, k)
        descrambler = digital.descrambler_bb(mask, 0b111, k)
        dst = blocks.null_sink(1)
        tb = gr.top_block()
        tb.connect(src, scrambler, descrambler, head, dst)
        %time tb.run()


Original:
CPU times: user 1.86 s, sys: 59.8 ms, total: 1.92 s
Wall time: 985 ms

My version:
CPU times: user 1.2 s, sys: 50.8 ms, total: 1.25 s
Wall time: 606 ms

So an improvement of around 50%, plus of course that my version can do higher polynomials.

I also had an idea to remake the blocks to eat bytes of data instead of only 1 bit /byte. Should be more memory efficient and possibly also more speedy. Maybe an approach where both methods are available, but that will be for a later commit.

ThomasHabets:
Built in functions surely aren't portable, right? Can this use volk popcount instead? https://github.com/gnuradio/volk/blob/master/kernels/volk/volk_64u_popcnt.h

ThomasHabets:
If volk popcount isn't fast enough, maybe it's missing using this builtin where available?

ewxl:
Thomas, those were very nice references.

I will test if volk speeds this method up using popcnt from there instead, I guess most systems have MMX and this might do the trick instead of using __builtins_parity.

A response to the popcnt, what the scrambler-functions need is a parity check function, not the number of set bits. The x86 CPU has had an instruction for this quite some time. https://en.wikipedia.org/wiki/Parity_flag (the reference correctly describes that it only does this on the LSB, but the __builtins_parity solves this somehow.

You may be right that this should be an improvement to volk, to check for parity. Should I try and improve this method that way instead?

When I read about builtins, they are compiler-specific, so I guess it's a GCC-thing, and Gnuradio states to support GCC which has it. About portability, I don't know enough to say how good/bad this gets.

ewxl:
If the __builtins_ are not to be used, there appears to be a compiler-flag for this, -fno-builtin

This is where you can read about builtins in gcc:
https://gcc.gnu.org/onlinedocs/gcc/Other-Builtins.html

ewxl:
Yeah on my systems the volk-version of popcnt is just the same performance as __builtins_parity, so I'll change for that instead, thanks for the input! Changing my commit soon.

ThomasHabets:
I can't find documentation on builtins, but my default assumption would be that clang has what GCC has. But I believe some build gnuradio with MSVC.

So volk is a safer bet, I'd say. But maybe someone more knowledgable can correct me.

drmpeg (commented):
Is maint-3.8 the correct target for this? Seems like sort of an API change or at least a new feature.

ewxl:
> Is maint-3.8 the correct target for this? Seems like sort of an API change or at least a new feature.

I'm not sure, I am a junior contributor to the project so please advise me if this is the wrong place.

I see the change as improvement, it will not break anything built on the old usage of the blocks.

gr-digital/python/digital/qa_scrambler.py
```diff
@@ -44,24 +48,64 @@ def tearDown(self):
...
+        self.assertEqual(tuple(np.convolve(dst_tap.data(),[1,0,0,0,0,1,0,1])%2)[7:-10],
+                tuple(src_data[:-10])) # manual descrambling test
+
+    def test_scrambler_descrambler_big(self):
+        mask,k = make_mask(51,6,0) #p(x) = x^51+x^6+1
+        src_data = np.random.randint(0,2,1000,dtype=np.int8)
         src = blocks.vector_source_b(src_data, False)
-        scrambler = digital.scrambler_bb(0x8a, 0x7F, 7)     # CCSDS 7-bit scrambler
-        descrambler = digital.descrambler_bb(0x8a, 0x7F, 7)
+        scrambler = digital.scrambler_bb(mask, 1, k)     # p(x) = x^7 + x^2 + 1
```
jdemel, line 68: I suggest to leave the old QA code in place and add more tests instead. That'll make it easier to merge this PR.

gr-digital/include/gnuradio/digital/lfsr.h
```diff
@@ -90,32 +90,26 @@ namespace digital {
...
-            throw std::invalid_argument("reg_len must be <= 31");
+        if (reg_len > 63)
+            throw std::invalid_argument("reg_len must be <= 63");
     }
 
     unsigned char next_bit()
     {
         unsigned char output = d_shift_register & 1;
-        unsigned char newbit = popCount(d_shift_register & d_mask) % 2;
+        unsigned char newbit = __builtin_parityl(d_shift_register & d_mask);
```
jdemel, line 112: We have people who compile GR with MSVC. So, this is probably not the way to go. I saw in the discussions that you switched to VOLK for this. That'd be the way to go I assume.

jdemel (changes requested):
I like this change. This should make it easier to use scrambling too.
I'd recommend to change the target branch to `master`. It is an API change after all. e.g. gqrx might be linked against it and that'll cause issues. gqrx is only an example.
As per `__builtin*` discussion, we'd need to have this code to be more portable as well.

Is this your first PR? If so, we might need to worry about a CLA.

gr-digital/include/gnuradio/digital/lfsr.h
ewxl: Yes, and first I thought I got good results with VOLK popcnt but I cannot reproduce it, I am sorry to say I think I made a mistake and did performance-tests still on the __builtin-code and not VOLK. I only get a minor improvement with VOLK from HEAD. Perhaps like ThomasHabets suggested, this function should be added to VOLK, and use the __builtin_parity when available? If I was better at MSVC, perhaps this could be resolved, because what __builtin does is use a CPU-feature for this calculation. In VOLK, it could be added as inline assemlby if the arch is right?

gr-digital/python/digital/qa_scrambler.py
ewxl: I will comply, but I will add;

The QA has room for even more improvement, it did not correctly detect a bug that was fixed with this commit: https://github.com/gnuradio/gnuradio/commit/fddb0d4ebe36516ee7caa3ebf25cf411b62db6c0#diff-930c609d625aa95d48a093a3e8e06f2ea0e211ba5fb863683bf2342d95574f66

I believe the QA was not updated this time, so it was not improved to detect this error.

I'm pretty sure the QA right now does not actually use the scrambler it intends to. It does complete without error, because it undoes what it does, it does not check that what it does, is correct.

ewxl:
> I like this change. This should make it easier to use scrambling too.
> I'd recommend to change the target branch to `master`. It is an API change after all. e.g. gqrx might be linked against it and that'll cause issues. gqrx is only an example.
> As per `__builtin*` discussion, we'd need to have this code to be more portable as well.
> 
> Is this your first PR? If so, we might need to worry about a CLA.

It will be my second, if you count the very small first one I did for the GRC. Can I read the CLA somewhere?

ewxl:
I have thoughts aswell in this matter, because it is not consistent with other modules like gr-fec and gr-trellis. If I remember correctly, those modules use the reciprocal polynomials compared to gr-digital. I'm not the one to say which one is correct to use, but I personally prefer the way gr-digital does it.

ewxl:
> Is maint-3.8 the correct target for this? Seems like sort of an API change or at least a new feature.

Would the best solution be to fork master instead and create a new PR?

drmpeg:
> Would the best solution be to fork master instead and create a new PR?

Yes, just close this one and submit a new PR.

gr-digital/include/gnuradio/digital/lfsr.h
ThomasHabets: I see these options:

1. Add this to volk either as inline asm or just using `__builtin_xxx` where available, and wait for gnuradio to have a minimum volk version that includes this.
2. Have CMake in gnuradio detect existence of this builtin, and fall back to volk popcount where not available.
3. Use `#ifdef` to check for volk support, then (2), in that order. And also implement in volk.

(2) and (3) can of course also have an MSVC option, but I also don't know what that would be.

Seems the decision could depend on if we think 3.9 can depend on a newer volk, and on how critical performance is in this code.

ewxl: I just spent some time to get into VOLK and when I try to use __builtin_parity from a kernel I implemented there, there is overhead that more than doubles the time than calling __builtin_parity directly, from a little test-program. I _must_ be doing something wrong, but I don't know what.

I like your option 2 the best, there is a simple #ifdef __GNUC__ that I can use for this. Someone with knowledge in MSVC may help to complement on that. Might not look pretty inside a block like that but right know that's my best shot.

ThomasHabets: Does you volk solution end up in a header file as static inline? Otherwise yeah the call would eat up most or all of the gain.

ewxl: This is what I came up with, and when I follow the call to volk in gdb, there are alot of calls before I end up where I want to. Should I try something else?

 67 static inline void volk_64u_parity_generic(int32_t* res, uint64_t value)
 68 {
 69     *res = __builtin_parityl(value);
 70 }

jdemel: VOLK is usually more oriented towards longer vectors. So the overhead may outweigh the benefit here.
Though, benchmarks are notoriously fragile. e.g. did you run a couple of iterations before you started to measure? 
VOLK registers the correct function during runtime. So this might include overhead at first.

ThomasHabets: Is there a way to skip the function registration benchmarking thing, and just "know" that `__builtin_parityl` is the fastest?

As for `#ifdef GNUC`, I would say that it's an *antipattern* to check for which environment you're in. You should check for *capability*. CMake surely has a way of checking, and then give you `ifdef HAVE_BUILTIN_PARITYL`?

PR #3972, opened 2020-12-01
gr-digital: improved performance and capability of scramblers to 64-bit

ewxl:
registers. Use __builtin_parity or volk popcnt. qa was also enhanced to detect errors.

See this thread for reference: https://github.com/gnuradio/gnuradio/pull/3968

### DIFF
--- a/gr-digital/CMakeLists.txt
+++ b/gr-digital/CMakeLists.txt
@@ -30,6 +30,11 @@ SET(GR_PKG_DIGITAL_EXAMPLES_DIR ${GR_PKG_DATA_DIR}/examples/digital)
 ########################################################################
 if(ENABLE_GR_DIGITAL)
 
+# Set HAVE_BUILTIN_PARITYL for use in scrambler if compiler is GNU
+if(CMAKE_CXX_COMPILER_ID MATCHES "GNU|Clang")
+    add_definitions(-DHAVE_BUILTIN_PARITYL)
+endif()
+
 ########################################################################
 # Add subdirectories
 ########################################################################
--- a/gr-digital/include/gnuradio/digital/lfsr.h
+++ b/gr-digital/include/gnuradio/digital/lfsr.h
@@ -12,6 +12,10 @@
 #define INCLUDED_DIGITAL_LFSR_H
 
 #include <gnuradio/digital/api.h>
+#ifndef HAVE_BUILTIN_PARITYL
+#include <volk/volk.h>
+#endif
+#include <stdint.h>
 #include <cstdint>
 #include <stdexcept>
 
@@ -25,7 +29,7 @@ namespace digital {
  *
  * \details
  * Generates a maximal length pseudo-random sequence of length
- * 2^degree-1
+ * 2^degree-1, if supplied with a primitive polynomial.
  *
  * Constructor: digital::lfsr(int mask, int seed, int reg_len);
  *
@@ -35,9 +39,9 @@ namespace digital {
  *             order bit.
  *
  *             Some common masks might be:
- *              x^4 + x^3 + x^0 = 0x19
- *              x^5 + x^3 + x^0 = 0x29
- *              x^6 + x^5 + x^0 = 0x61
+ *              x^4 + x^3 + x^0 = 0x19, K=3
+ *              x^5 + x^3 + x^0 = 0x29, K=4
+ *              x^6 + x^5 + x^0 = 0x61, K=5
  *
  * \param seed - the initialization vector placed into the
  *             register during initialization. Low order bit
@@ -78,32 +82,33 @@ namespace digital {
 class lfsr
 {
 private:
-    uint32_t d_shift_register;
-    uint32_t d_mask;
-    uint32_t d_seed;
-    uint32_t d_shift_register_length; // less than 32
-
-    static uint32_t popCount(uint32_t x)
-    {
-        uint32_t r = x - ((x >> 1) & 033333333333) - ((x >> 2) & 011111111111);
-        return ((r + (r >> 3)) & 030707070707) % 63;
-    }
+    uint64_t d_shift_register;
+    uint64_t d_mask;
+    uint64_t d_seed;
+    uint8_t d_shift_register_length; // less than 64
 
 public:
-    lfsr(uint32_t mask, uint32_t seed, uint32_t reg_len)
+    lfsr(uint64_t mask, uint64_t seed, uint8_t reg_len)
         : d_shift_register(seed),
           d_mask(mask),
           d_seed(seed),
           d_shift_register_length(reg_len)
     {
-        if (reg_len > 31)
-            throw std::invalid_argument("reg_len must be <= 31");
+        if (reg_len > 63)
+            throw std::invalid_argument("reg_len must be <= 63");
     }
 
     unsigned char next_bit()
     {
         unsigned char output = d_shift_register & 1;
-        unsigned char newbit = popCount(d_shift_register & d_mask) % 2;
+        uint64_t newbit;
+#ifdef HAVE_BUILTIN_PARITYL
+        newbit = __builtin_parityl(d_shift_register & d_mask);
+#else
+        volk_64u_popcnt(&newbit, d_shift_register & d_mask);
+        newbit %= 2;
+#endif
+
         d_shift_register =
             ((d_shift_register >> 1) | (newbit << d_shift_register_length));
         return output;
@@ -112,7 +117,13 @@ public:
     unsigned char next_bit_scramble(unsigned char input)
     {
         unsigned char output = d_shift_register & 1;
-        unsigned char newbit = (popCount(d_shift_register & d_mask) % 2) ^ (input & 1);
+        uint64_t newbit;
+#ifdef HAVE_BUILTIN_PARITYL
+        newbit = __builtin_parityl(d_shift_register & d_mask) ^ (input & 1);
+#else
+        volk_64u_popcnt(&newbit, d_shift_register & d_mask);
+        newbit = (newbit ^ input) & 1;
+#endif
         d_shift_register =
             ((d_shift_register >> 1) | (newbit << d_shift_register_length));
         return output;
@@ -120,8 +131,15 @@ public:
 
     unsigned char next_bit_descramble(unsigned char input)
     {
-        unsigned char output = (popCount(d_shift_register & d_mask) % 2) ^ (input & 1);
-        unsigned char newbit = input & 1;
+        unsigned char output;
+#ifdef HAVE_BUILTIN_PARITYL
+        output = __builtin_parityl(d_shift_register & d_mask) ^ (input & 1);
+#else
+        uint64_t _tmp;
+        volk_64u_popcnt(&_tmp, d_shift_register & d_mask);
+        output = (_tmp ^ input) & 1;
+#endif
+        uint64_t newbit = input & 1;
         d_shift_register =
             ((d_shift_register >> 1) | (newbit << d_shift_register_length));
         return output;
@@ -143,7 +161,7 @@ public:
         }
     }
 
-    int mask() const { return d_mask; }
+    uint64_t mask() const { return d_mask; }
 };
 
 } /* namespace digital */
--- a/gr-digital/lib/additive_scrambler_bb_impl.cc
+++ b/gr-digital/lib/additive_scrambler_bb_impl.cc
@@ -18,22 +18,22 @@
 namespace gr {
 namespace digital {
 
-additive_scrambler_bb::sptr additive_scrambler_bb::make(int mask,
-                                                        int seed,
-                                                        int len,
-                                                        int count,
-                                                        int bits_per_byte,
+additive_scrambler_bb::sptr additive_scrambler_bb::make(uint64_t mask,
+                                                        uint64_t seed,
+                                                        uint8_t len,
+                                                        int64_t count,
+                                                        uint8_t bits_per_byte,
                                                         const std::string& reset_tag_key)
 {
     return gnuradio::make_block_sptr<additive_scrambler_bb_impl>(
         mask, seed, len, count, bits_per_byte, reset_tag_key);
 }
 
-additive_scrambler_bb_impl::additive_scrambler_bb_impl(int mask,
-                                                       int seed,
-                                                       int len,
-                                                       int count,
-                                                       int bits_per_byte,
+additive_scrambler_bb_impl::additive_scrambler_bb_impl(uint64_t mask,
+                                                       uint64_t seed,
+                                                       uint8_t len,
+                                                       int64_t count,
+                                                       uint8_t bits_per_byte,
                                                        const std::string& reset_tag_key)
     : sync_block("additive_scrambler_bb",
                  io_signature::make(1, 1, sizeof(unsigned char)),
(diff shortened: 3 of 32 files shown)


ewxl:
I get a test-fail for qa_header_payload_demux.py that I don't believe I'm responsible for, not sure how to proceed with that. I also encounter a weird failure that I can't reproduce when using the gr_unittest, assertEqual on two lists. It works when the lists are identical, as intended, but when my code was faulty and the lists were not equal, it get's stuck in an infinite loop. When I try this to make it fail, it seems to work as intended and points to where the difference was in the list, but not when my code gave different lists. I guess that error might fall under a new Issue instead.

gr-digital/lib/glfsr_source_f_impl.cc
```diff
@@ -21,31 +21,30 @@ namespace gr {
...
+                                         uint64_t mask,
+                                         uint64_t seed)
     : sync_block("glfsr_source_f",
                  io_signature::make(0, 0, 0),
                  io_signature::make(1, 1, sizeof(float))),
       d_glfsr(mask ? mask : glfsr::glfsr_mask(degree), seed),
       d_repeat(repeat),
       d_index(0),
-      d_length((((uint32_t)1) << degree) - 1)
+      d_length((1L << degree) - 1)
```
rear1019, line 39: Use `1ULL` or better (?) `static_cast<uint64_t>()`: `long` has only 32 bit on x64 Windows (as far as I remember) and to avoid dealing with signed integers. Also consider using `d_length{…}` instead of `d_length(…)` to prevent narrowing conversion [1]

[1] https://en.cppreference.com/w/cpp/language/list_initialization#Narrowing_conversions

gr-digital/lib/glfsr_source_b_impl.cc
```diff
@@ -20,31 +19,30 @@ namespace gr {
...
+                                         uint64_t mask,
+                                         uint64_t seed)
     : sync_block("glfsr_source_b",
                  io_signature::make(0, 0, 0),
                  io_signature::make(1, 1, sizeof(unsigned char))),
       d_glfsr(mask ? mask : glfsr::glfsr_mask(degree), seed),
       d_repeat(repeat),
       d_index(0),
-      d_length((((uint32_t)1) << degree) - 1)
+      d_length((1L << degree) - 1)
```
rear1019, line 37: See note below.

gr-digital/lib/glfsr.cc
```diff
@@ -52,9 +52,9 @@ static uint32_t s_polynomial_masks[] = {
 
 glfsr::~glfsr() {}
 
-uint32_t glfsr::glfsr_mask(unsigned int degree)
+uint64_t glfsr::glfsr_mask(unsigned int degree)
 {
-    if (degree < 1 || degree > 32)
+    if (degree < 1 || degree > 64)
         throw std::runtime_error(
             "glfsr::glfsr_mask(): degree must be between 1 and 32 inclusive");
```
rear1019, line 59: Fix error message (32 -> 64).

gr-digital/include/gnuradio/digital/lfsr.h
```diff
@@ -78,32 +81,33 @@ namespace digital {
...
+            throw std::invalid_argument("reg_len must be <= 63");
     }
 
     unsigned char next_bit()
     {
         unsigned char output = d_shift_register & 1;
-        unsigned char newbit = popCount(d_shift_register & d_mask) % 2;
+        uint64_t newbit;
+#ifdef HAVE_BUILTIN_PARITYL
+        newbit = __builtin_parityl(d_shift_register & d_mask);
```
rear1019, line 105: From [1]
> I don't have GR set up on a RPi, but here's the code generated by a test of __builtin_parity() and __builtin_parityl() on a RPi3. They both call __paritysi2() which looks no less efficient than calling popCount().

`__builtin_parity()` (or `__builtin_popcount()`) will not issue special instructions unless they are enabled: `-mpopcnt` for x86/x64 (you might get away with one of the `-msse4X` options, but that depends on AMD vs. Intel as far as I remember).

From [2]
> Built in functions surely aren't portable, right?

The resulting binary will not be portable if special instructions are issued (no runtime detection), but the source is (at least when using GCC/Clang).

[1] https://github.com/gnuradio/gnuradio/pull/3968#issuecomment-734927631
[2] https://github.com/gnuradio/gnuradio/pull/3968#issuecomment-735467759

gr-digital/lib/additive_scrambler_bb_impl.cc
```diff
@@ -56,25 +56,26 @@ additive_scrambler_bb_impl::additive_scrambler_bb_impl(int mask,
...
-int additive_scrambler_bb_impl::_get_next_reset_index(int noutput_items,
-                                                      int last_reset_index /* = -1 */)
+int64_t
+additive_scrambler_bb_impl::_get_next_reset_index(int64_t noutput_items,
+                                                  int64_t last_reset_index /* = -1 */)
 {
-    int reset_index =
+    int64_t reset_index =
         noutput_items; // This is a value that gets never reached in the for loop
     if (d_count == -1) {
```
rear1019, line 73: Can’t we use `d_count = 0` as a special value and consequently `uint64_t` instead of `int64_t` for `d_count`? Resetting the scrambler after 0 Bytes doesn't make sense.

rear1019 (changes requested):

drmpeg:
> I get a test-fail for qa_header_payload_demux.py that I don't believe I'm responsible for, not sure how to proceed with that.

qa_header_payload_demux (and qa_uncaught_exception) is known to fail for the master branch. It's been deleted from the CI tests.

gr-digital/lib/glfsr_source_f_impl.cc
ewxl: Wilco, good spot

gr-digital/include/gnuradio/digital/lfsr.h
ewxl: [1]: I'm looking for the use of the PF flag, and maybe it's the setnp instruction in x64? I seem to get this with __builtin_parityl without having to enable anything when compiling GR. I get a speedup of 50% on my system. Are you sure PF and setnp depend on SSE or others? From what I understand this asm-instruction was included in the original x86 set.

[2]: Do you know of a better way to solve this? When I tried the VOLK-way, the overhead gave me the same performance as previous.

gr-digital/lib/additive_scrambler_bb_impl.cc
ewxl: You are possibly correct, I haven't looked in to this. This block had a bug that was fix due to this weird usage of the d_count, so I'm keen to correct this aswell, there is no point in waiting, I guess.

ewxl: The only thing I can think of is that d_count=0 will disable the tags-section. If that is desired or planned, I don't know. But I don't see any issue with using int64_t either though.

gr-digital/include/gnuradio/digital/lfsr.h
rear1019: [1]: To be honest, this is the first time I came across `setnp` (seems to be indeed part of the original i386 instruction set) and the Parity Flag (PF). `setnp` alone ist not sufficient as it processes a single byte only. This is the code produced by GCC with `__builtin_parity(val)` if `popcnt` is not available: (`RCX` contains `val`, result is in `RAX`)
```
0x55555555518e  <+   30>        48 89 ce              mov    %rcx,%rsi
0x555555555191  <+   33>        89 cf                 mov    %ecx,%edi
0x555555555193  <+   35>        48 c1 ee 20           shr    $0x20,%rsi
0x555555555197  <+   39>        31 f7                 xor    %esi,%edi
0x555555555199  <+   41>        89 f8                 mov    %edi,%eax
0x55555555519b  <+   43>        c1 ef 10              shr    $0x10,%edi
0x55555555519e  <+   46>        31 f8                 xor    %edi,%eax
0x5555555551a0  <+   48>        30 e0                 xor    %ah,%al
0x5555555551a2  <+   50>        0f 9b c0              setnp  %al
```
[2] Short of advising users to compile with `-march=native`, no. Using `volk` here is pointless because only a single value is processed.

gr-digital/lib/additive_scrambler_bb_impl.cc
ewxl: Will resolve this conversation, since there is no problem here. I do see a potential fix for this in the future, so it's not that it is a bad idea, but not just part of this commit.

willcode (commented):
This looks like a good change to make. Haven't tried it out yet.

Do we target any 32-bit architectures where forcing 64-bit ops would affect performance in any way we care about?

Is there a way to do more definitive performance tests? In the wall-time tests you did previously, was volk_profile run, and what kernel was it using for popcnt?

If there are performance gains, are we better off adding a parity kernel to volk?

gr-digital/python/digital/utils/lfsr.py
```diff
@@ -0,0 +1,22 @@
...
+#
+# Copyright 2020 Free Software Foundation, Inc.
+# 
+# This file is part of GNU Radio
+# 
+# SPDX-License-Identifier: GPL-3.0-or-later
+#
+# 
+
+def lfsr_args(seed,*exp):
```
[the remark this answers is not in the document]
ewxl: volk_profile was run, and I compiled and test that both variants work, that is, both volk_popcnt and builtin. Volk_profile was run, and it says:
volk_64u_popcnt a_sse4_2 generic
volk_64u_popcnt generic generic
volk_64u_popcnt a_sse4_2 generic

Not sure why i get 3 of them.

The volk_popcnt is pretty much the same as the original.

I tried implementing the builtin_parityl-call as a kernel i volk but what happens in volk seems to eat up any performance gain, sadly. 

About 32-bit, the builtin_parityl seems to solve this issue for us, but someone with access to a system that can test this would answer any doubts.

As for lfrs_args, I agree to the signature-change, however my suggestion is to leave it as for now, that lfrs_args is just there now for the QA. Let's do a makeover regarding all gr-digital blocks in another commit?

daniestevez:
> Do we target any 32-bit architectures where forcing 64-bit ops would affect performance in any way we care about?

I think that a fair amount of people have simple GNU Radio flowgraphs running on 32-bit ARMs, like the older Rasberry Pi's, etc. Maybe it would be a good idea to template this and instance the 32-bit or  the 64-bit version depending on the required length.

I'm curious about what would be the use case for a 64-bit scrambler? All the real-world systems and experiments I can think of use much less than 32 bits.

ewxl:
> > Do we target any 32-bit architectures where forcing 64-bit ops would affect performance in any way we care about?
> 
> I think that a fair amount of people have simple GNU Radio flowgraphs running on 32-bit ARMs, like the older Rasberry Pi's, etc. Maybe it would be a good idea to template this and instance the 32-bit or the 64-bit version depending on the required length.

I'm sure there is, but we believe there will be no performance impact on the 32-bit arm (this was discussed in some resolved thread), the builtin_parity function is implemented by the compiler and provides a solution to the architecture. But if any test shows that this doesn't work, of course it should be looked in to.

> I'm curious about what would be the use case for a 64-bit scrambler? All the real-world systems and experiments I can think of use much less than 32 bits.

I'm not sure how to respond to this, the main reason was of course the performance-gain in time, but since it was simple to expand to 64-bit register, why not? The future will probably give you examples and I consider GnuRadio to be the place to be able to experiment with such polynomials.

gr-digital/lib/additive_scrambler_bb_impl.h
```diff
@@ -21,29 +21,29 @@ class additive_scrambler_bb_impl : public additive_scrambler_bb
...
+                               uint8_t bits_per_byte = 1,
                                const std::string& reset_tag_key = "");
     ~additive_scrambler_bb_impl() override;
 
-    int mask() const override;
-    int seed() const override;
-    int len() const override;
-    int count() const override;
-    int bits_per_byte() override { return d_bits_per_byte; };
+    uint64_t mask() const;
```
marcusmueller, line 42: is there a reason we're removing the `override` here and elsewhere?

gr-digital/python/digital/qa_lfsr.py
```diff
@@ -25,14 +25,27 @@ def tearDown(self):
...
         l = digital.lfsr(1, 1, reglen)
-
         result_data = []
         for i in range(4 * (reglen + 1)):
             result_data.append(l.next_bit())
 
         expected_result = 4 * ([1, ] + reglen * [0, ])
         self.assertFloatTuplesAlmostEqual(expected_result, result_data, 5)
 
+    def test_lfsr_002(self):
```
marcusmueller, line 35: nice!

gr-digital/python/digital/qa_scrambler.py
```diff
@@ -34,6 +34,83 @@ def setUp(self):
...
+        scrambler = digital.additive_scrambler_bb(*_a)
+        dst = blocks.vector_sink_b()
+        self.tb.connect(src, scrambler, dst)
+        self.tb.run()
+        reg = np.zeros(52,np.int8)
+        reg[::-1][(51,3,0),] = 1
+        res = (np.convolve(dst.data(),reg)%2)[52:-52]
+        self.assertEqual(len(res),sum(res)) # when convolved with mask, 
+                                            # after sync, only 1's whould be returned.
+
```
marcusmueller, line 113: Very nice unit testing, indeed.

gr-digital/lib/additive_scrambler_bb_impl.h
ewxl: No, this is probably a typo, I didn't think of it when writing, but it should be there I think! Will update.

gr-digital/CMakeLists.txt
```diff
@@ -30,6 +30,11 @@ SET(GR_PKG_DIGITAL_EXAMPLES_DIR ${GR_PKG_DATA_DIR}/examples/digital)
 ########################################################################
 if(ENABLE_GR_DIGITAL)
 
+# Set HAVE_BUILTIN_PARITYL for use in scrambler if compiler is GNU
```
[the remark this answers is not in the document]
ewxl: Certainly, I hope I just did what you requested, I'm not super experienced with git yet.

gr-digital/CMakeLists.txt
```diff
@@ -30,6 +30,11 @@ SET(GR_PKG_DIGITAL_EXAMPLES_DIR ${GR_PKG_DATA_DIR}/examples/digital)
...
+if(CMAKE_CXX_COMPILER_ID MATCHES "GNU|Clang")
+    add_definitions(-DHAVE_BUILTIN_PARITYL)
```
mbr0wn, line 35: Is this true for all gcc/clang versions down to our minimum?

mbr0wn (commented):
I think this looks very good and we should get this merged. We're not as strict with our Python coding style as we are with C++, but that's mostly because we don't have that automated. It would be helpful if you don't mix the space-after-comma with no-space-after-comma. In general, we like to follow PEP8 (or default Pylint settings).

Has anyone checked 32-bit performance yet?

gr-digital/CMakeLists.txt
ewxl: I don't know this, but __builtin_parity seem to have been around forever.

mbr0wn: Hm yeah, definitely much longer than 8.3.0.

mbr0wn:
@ewxl I saw you force-pushed, but there's still conflicts. Also, I notice you put the commits onto your own master branch. That's not a dealbreaker, but I recommend creating a branch like "scrambler-update" or something like that.

mbr0wn:
Also, some hashes fail now. See the CMake logs from our CI.

ewxl:
> Also, some hashes fail now. See the CMake logs from our CI.



> @ewxl I saw you force-pushed, but there's still conflicts. Also, I notice you put the commits onto your own master branch. That's not a dealbreaker, but I recommend creating a branch like "scrambler-update" or something like that.

Yes, I will do this next time, this is a learning experience for me with git.

I think I succeeded with a rebase now, or am I still missing something?

mbr0wn:
> > Also, some hashes fail now. See the CMake logs from our CI.
> 
> > @ewxl I saw you force-pushed, but there's still conflicts. Also, I notice you put the commits onto your own master branch. That's not a dealbreaker, but I recommend creating a branch like "scrambler-update" or something like that.
> 
> Yes, I will do this next time, this is a learning experience for me with git.
> 
> I think I succeeded with a rebase now, or am I still missing something?

I think it's looking good now.

Issue #3971, opened 2020-11-30
File Source Under Windows cannot handle >4GB


MountainLogic:
Two apparent File Source issue.  The fundamental is a 4.1 GB limit to file source under windows 3.8.1.0 (Python 2.7.10).  I've not seen this limit under MacOS.  Is this a feature or a bug?

I've also seen an odd parsing of under scores in window file names on windows that I'm not sure is real or just an odd MyFlow.py generation artifact.

I'm having trouble opening a file with the "File Source" block I recorded with gqrx that are larger than 4 GB.  I have File Source set to vector=1, offset-0, length=0.  I am able to use this file on my Mac GRC 3.8, but not windows GRC 3.8.  I'm running GRC 3.8.0 on windows from the pre-built bins (also got same results after upgrading to the current 'official' 3.8.1 bins with the needed bat file fix).    I found one apparent bug with underscores, but even with working around this (changing underscores to dashes in file names) I still get a fstat failure.  

I'm getting the following in the console output:

```
Generating: 'C:\\Users\\FooBar\\Google Drive\\Weather\\SDR\\GR\\wsr88d_001.py'

Executing: C:\Program Files\GNURadio-3.8\gr-python27\python.exe -u C:\Users\FooBar\Google Drive\Weather\SDR\GR\wsr88d_001.py

gr::log :ERROR: file_source0 - C:\Users\FooBar\wip\gqrx_20201116_162754_2743499985_19999999_fc.raw: Unknown error
Traceback (most recent call last):
  File "C:\Users\FooBar\Google Drive\Weather\SDR\GR\wsr88d_001.py", line 230, in <module>
    main()
  File "C:\Users\FooBar\Google Drive\Weather\SDR\GR\wsr88d_001.py", line 208, in main
    tb = top_block_cls()
  File "C:\Users\FooBar\Google Drive\Weather\SDR\GR\wsr88d_001.py", line 129, in __init__
    self.blocks_file_source_0 = blocks.file_source(gr.sizeof_gr_complex*1, "C:\Users\FooBar\wip\gqrx_20201116_162754_2743499985_19999999_fc.raw", True, 0, 0)
  File "C:\Program Files\GNURadio-3.8\lib\site-packages\gnuradio\blocks\blocks_swig0.py", line 1100, in make
    return _blocks_swig0.file_source_make(itemsize, filename, repeat, offset, len)
RuntimeError: can't fstat file

>>> Done (return code 1)
 
```

[...]

PR #3975, opened 2020-12-02
gr-blocks: Fix file source block large file handling for Windows.

drmpeg:


ThomasHabets (commented):
I don't think I have anything to say about this. Yes, [`_fstatd64` exists](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/fstat-fstat32-fstat64-fstati64-fstat32i64-fstat64i32?view=msvc-160), and it seems that this brings `fstat` in line with the other calls here.  But that's about all I can say not knowing Windows.

drmpeg:
This has been tested. See https://github.com/gnuradio/gnuradio/issues/3971 Also approved elsewhere https://github.com/gnuradio/gnuradio/pull/3976

ThomasHabets:
I didn't mean to imply it's not been tested, just that I was added as a reviewer but am not competent in windows so I can't review aside from "sure, I don't see why not". In any case I don't have merging powers.

ryanvolz (approved):
I can confirm that this builds. I haven't personally verified that it fixes the large file problem but @MountainLogic says it does and that's good enough for me.

mbr0wn:
I didn't test, but trust @ryanvolz and @mountainlogic's testing.

Issue #3542, opened 2020-06-04
cmake: Check for PyQtGraph as a runtime dependency of gr_filter_design


noc0lour:
As we do this for other Python runtime dependencies: Check if pyqtgraph is available and fail otherwise.

mbr0wn:
We do this: https://github.com/gnuradio/gnuradio/blob/21a409fb7526609bd234c1e6c7f5bd2ddf0c6d74/gr-filter/python/filter/design/filter_design.py#L45-L48

PR #3987, opened 2020-12-08
Build filterdesign tools ,only if required components are available

dl1ksv:
gr_filter_design is build without warnings or error messages even if required python libs are missing.
In this using this tool you get an error messages.
I propose to check during cmake and build these if the libs are available. As the filter components  in gr-filter can be build without the design tools I suggest to show the filterdesign tools as component in cmake.


mbr0wn:
An argument can be made that since these are all runtime-dependencies, we could add those checks to the tool itself. That helps when building GNU Radio for other platforms without needing all the dependencies.

dl1ksv:
Well, I see. In this case a user could install the missing python libs himself. But I personally would prefer to get a hint when running cmake.

dl1ksv:
Perhaps it would suffice to give a warning during cmake.

mbr0wn:
Superseded by #3542

PR #3989, opened 2020-12-10
Some improvements on gr_filter_design

### LINKED ISSUE #3951: Python3 and 'gr-filter/python/filter/design/filter_design.py'
Running this tool shows it's not completely Python3 compatible. When using it, I got an exception:

```

  File "f:\gv\dx-radio\gnuradio\gv-build\lib\site-packages\gnuradio\filter\filter_design.py", line 777, in design
    self.design_iir()
  File "f:\gv\dx-radio\gnuradio\gv-build\lib\site-packages\gnuradio\filter\filter_design.py", line 905, in design_iir
    except StandardError as e:
NameError: name 'StandardError' is not defined
```

Caused by lines [905](https://github.com/gnuradio/gnuradio/blob/master/gr-filter/python/filter/design/filter_design.py#L905) and [917](https://github.com/gnuradio/gnuradio/blob/master/gr-filter/python/filter/design/filter_design.py#L917).

According to [this](https://portingguide.readthedocs.io/en/latest/exceptions.html#the-removed-standarderror), simply replacing `StandardError` with `Exception` should work for Python3.

dl1ksv:
Small
* Some input windows (End of Pass Band etc.) are to small to read the input values. So the size of these windows was modified.
* The file pyqt_filter_stacked.py now will be generated by puic5 during the cmake/build process.
* StandardError which is not available in Python 3 is replaced by  Exception. Fixes #3951


willcode:
This cleans up the display a bit for me (can see the numbers now). I still don't see the QLabels for the filter specs on my display, but seeing the numbers is great.

mormj:
on my setup:

before:
image

after:
image

Not sure if it is my screen resolution, but I am not seeing the issue before applying the patch.

Though it is a nicer workflow, one concern I have about generating the pyqt_filter_stacked.py on the fly is now pyuic is a dependency, whereas previously one did not have to have it unless modifying the GUI here.  Also, I'm wondering if slight variations in versions of pyuic could create differences in the generated python that we won't know about until people are compiling on their own systems.  We don't test for any of this in CI.

willcode:
grfilter
Here's what i see, and it doesn't improve when the windows is stretched. It was the same after applying the patch, but I'll go back and make sure.

dl1ksv:
Here is my patched version
patched
It's taken on a notebook with 17'' display and 1920x1080 resolution.

dl1ksv:
Meanwhile I tested on another laptop with 15'' display and 1920x1080 resolution. Unpatched I get the same picture as @mormj .
The numbers of the band values are completely to be seen.
But after the patch I get the following picture.

Patched

Have a look a the band values. Unpatched they are left aligned, after the patch the are right aligned and the text 'Magnitude Response' is no longer cut.
@mormj As in your patched version the numbers are still left aligned, something went wrong with your patch. Please check, if  the file pyqt_filter_stacked.py is really regenerated.
I agree, that the use of pyuic5 has to be discussed. But building gr-qtgui we do the same: we use uic to generate spectrumdisplayform.ui.h from spectrumdisplayform.ui

Another important influence on the gui has the default font and the fontsize as you can see on @willcode 's example above.

willcode:
I'm admittedly recovering from some bad High DPI settings. What I get now is more along the lines of what you are showing. I'll test with the PR when I get a chance.

willcode:
Did some more tests, trying to factor out any impact of High DPI. There are some improvements with the patch, mainly in the area of the check boxes. How well the numbers fit in the input boxes depends on the exact QT_FONT_DPI setting. With the actual setting of my screen, no High DPI scaling, there wasn't quite enough space for '50000', for instance. There was enough space without the patch. We must be trying to use fixed-size widgets instead of grabbing font metrics at runtime.

mormj:
Looks like I definitely did not properly apply the patch before.  This time, I actually did not have pyuic on my machine, and it threw an error in the cmake, continued to build, then hit an error in the build.  

```
[  6%] Generating pyqt_filter_stacked.py
/bin/sh: 1: PYUIC_PROGRAM-NOTFOUND: not found
make[2]: *** [gr-filter/python/filter/gui/CMakeFiles/pygen_gr_filter_python_filter_gui_a1eeb.dir/build.make:133: gr-filter/python/filter/gui/pyqt_filter_stacked.py] Error 127
make[1]: *** [CMakeFiles/Makefile2:6225: gr-filter/python/filter/gui/CMakeFiles/pygen_gr_filter_python_filter_gui_a1eeb.dir/all] Error 2
```
Still not a fan of requiring pyuic to build gr though.

Here is my after now - it looks good to me.
image

dl1ksv:
That looks as expected.
@mormj Your doubt on using pyuic: We do the same on building gr-qtgui with uic. But there is one difference. The use of uic is embedded in a cmake macro provided by cmake. This should ensure that it works on windows, too.
But as I have now windows box, I cannot test my solution on windows. But leaving as is, is no option.

What about comparing the modification time of pyqt_filter_stacked.ui and pyqt_filter_stacked.py and ask the user to run pyuic, if the former is newer?

mormj:
The difference I noticed is that I didn't have pyuic installed, and with the patch, it failed to build.  But for some reason it is still able to build gr-qtgui.  I haven't looked into that cmake - is it a situation where if pyuic is not installed it will use the generated py that already lives in tree?

dl1ksv:
In the source tree pyqt_filter_stacked.py no longer exists. It will be compiled from pyqt_filter_stacked.ui by pyuic5.
The same is on gr-qtgui and spectrumdisplayform. spectrumdisplayform.ui exists in the source tree and spectrumdisplayform.ui.h and ui_spectrumdisplayform.h only in the build tree, generated by uic.

If building pyqt_filter_stacked.py fails and you install it on an existing gnuradio installation the old one will be used.

pyuic5 does not belong to the qt environment. I thought it would belong to pyqt but it belongs to pyqt5-dev-tools or something like that , depending on your distro.

As far as I can see the only use of pyuic5 is to generate this gui. A normal user won't change the gui This is a developer's task.
So I'd tend to provide pyqt_filter_stacked.py in the source tree and only give a warning if pyqt_filter_stacked.ui was modified.
So you don't have to care about the package containing puyic5.

dl1ksv:
Even without applying my patch, the gui looks now well. I don't know why.

PR #3991, opened 2020-12-11
gr-analog: Fixed orientation of L and R audio in wfm_rcv_pll. Changed gain of multiply_const_ff blocks from 5.5 to 1.0.

efardin:
This change resolves two issues:

1. L and R audio were reversed. Negation should occur in the R audio path - refer to [this](https://cdn.rohde-schwarz.com/pws/dl_downloads/dl_application/application_notes/7bm105/7BM105_0E.pdf) R&S app note for details.
2. Gain of multiply_const_ff block should be 1. In this case, the demodulated audio can be multiplied by the max_deviation parameter of the quadrature demodulator (75 kHz) to give the L and R deviation values.

### DIFF
--- a/gr-analog/python/analog/wfm_rcv_pll.py
+++ b/gr-analog/python/analog/wfm_rcv_pll.py
@@ -17,7 +17,7 @@ from gnuradio.filter import firdes
 from gnuradio import analog
 
 class wfm_rcv_pll(gr.hier_block2):
-    def __init__(self, demod_rate, audio_decimation):
+    def __init__(self, demod_rate, audio_decimation, deemph_tau):
         """
         Hierarchical block for demodulating a broadcast FM signal.
 
@@ -27,6 +27,7 @@ class wfm_rcv_pll(gr.hier_block2):
         Args:
             demod_rate: input sample rate of complex baseband input. (float)
             audio_decimation: how much to decimate demod_rate to get to audio. (integer)
+            deemph_tau: deemphasis ime constant in seconds (75us in US, 50us in EUR). (float)
         """
         gr.hier_block2.__init__(self, "wfm_rcv_pll",
                                 gr.io_signature(1, 1, gr.sizeof_gr_complex), # Input signature
@@ -40,11 +41,13 @@ class wfm_rcv_pll(gr.hier_block2):
         # Variables
         ##################################################
         self.demod_rate = demod_rate
+        self.deemph_tau = deemph_tau
         self.stereo_carrier_filter_coeffs_0 = stereo_carrier_filter_coeffs_0 = firdes.band_pass(-2.0, demod_rate, 37600, 38400, 400, fft.window.WIN_HAMMING, 6.76)
         self.stereo_carrier_filter_coeffs = stereo_carrier_filter_coeffs = firdes.complex_band_pass(1.0, demod_rate, 18980, 19020, 1500, fft.window.WIN_HAMMING, 6.76)
         self.deviation = deviation = 75000
         self.audio_filter = audio_filter = firdes.low_pass(1, demod_rate, 15000,1500, fft.window.WIN_HAMMING, 6.76)
         self.audio_decim = audio_decim = audio_decimation
+        self.audio_rate = audio_rate = demod_rate / audio_decim
         self.samp_delay = samp_delay = (len(stereo_carrier_filter_coeffs) - 1) // 2 + (len(stereo_carrier_filter_coeffs_0) - 1) // 2
 
         ##################################################
@@ -65,8 +68,8 @@ class wfm_rcv_pll(gr.hier_block2):
         self.blocks_add_xx_0 = blocks.add_vff(1)
         self.analog_quadrature_demod_cf_0 = analog.quadrature_demod_cf(demod_rate/(2*math.pi*deviation))
         self.analog_pll_refout_cc_0 = analog.pll_refout_cc(0.001, 2*math.pi * 19200 / demod_rate, 2*math.pi * 18800 / demod_rate)
-        self.analog_fm_deemph_0_0 = analog.fm_deemph(fs=demod_rate//audio_decim, tau=75e-6)
-        self.analog_fm_deemph_0 = analog.fm_deemph(fs=demod_rate//audio_decim, tau=75e-6)
+        self.analog_fm_deemph_0_0 = analog.fm_deemph(fs=audio_rate, tau=deemph_tau)
+        self.analog_fm_deemph_0 = analog.fm_deemph(fs=audio_rate, tau=deemph_tau)
         self.blocks_delay_0 = blocks.delay(gr.sizeof_float*1, samp_delay)
 
         ##################################################
--- a/gr-analog/python/analog/wfm_rcv_pll.py
+++ b/gr-analog/python/analog/wfm_rcv_pll.py
@@ -26,7 +26,7 @@ class wfm_rcv_pll(gr.hier_block2):
 
         Args:
             demod_rate: input sample rate of complex baseband input. (float)
-            audio_decimation: how much to decimate demod_rate to get to audio. (integer) FIXME: Not actually implemented!
+            audio_decimation: how much to decimate demod_rate to get to audio. (integer)
         """
         gr.hier_block2.__init__(self, "wfm_rcv_pll",
                                 gr.io_signature(1, 1, gr.sizeof_gr_complex), # Input signature
@@ -39,14 +39,13 @@ class wfm_rcv_pll(gr.hier_block2):
         ##################################################
         # Variables
         ##################################################
-        self.samp_rate = samp_rate = 3840000
-        self.rf_decim = rf_decim = 10
-        self.demod_rate = demod_rate = (int)(samp_rate/rf_decim)
-        self.stereo_carrier_filter_coeffs_0 = stereo_carrier_filter_coeffs_0 = firdes.band_pass(1.0, demod_rate, 37600, 38400, 400, fft.window.WIN_HAMMING, 6.76)
+        self.demod_rate = demod_rate
+        self.stereo_carrier_filter_coeffs_0 = stereo_carrier_filter_coeffs_0 = firdes.band_pass(-2.0, demod_rate, 37600, 38400, 400, fft.window.WIN_HAMMING, 6.76)
         self.stereo_carrier_filter_coeffs = stereo_carrier_filter_coeffs = firdes.complex_band_pass(1.0, demod_rate, 18980, 19020, 1500, fft.window.WIN_HAMMING, 6.76)
         self.deviation = deviation = 75000
         self.audio_filter = audio_filter = firdes.low_pass(1, demod_rate, 15000,1500, fft.window.WIN_HAMMING, 6.76)
-        self.audio_decim = audio_decim = (int)(demod_rate/48000)
+        self.audio_decim = audio_decim = audio_decimation
+        self.samp_delay = samp_delay = (len(stereo_carrier_filter_coeffs) - 1) // 2 + (len(stereo_carrier_filter_coeffs_0) - 1) // 2
 
         ##################################################
         # Blocks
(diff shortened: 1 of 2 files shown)


drmpeg:
The channel orientation has previously been tested to be correct. Can you provide a way to verify that it is not?

efardin:
Attached IQ data file recorded from signal generator, set for 1 kHz tone on L channel. The result can be checked with a WAV file sink and an audio editor (Audacity, for example). Don't make the same mistake I made when testing with headphones - they can go either way around ;-)

[fm_l_384k.zip](https://github.com/gnuradio/gnuradio/files/5682586/fm_l_384k.zip)

drmpeg:
Well, it looks like the channels really are reversed. I was led astray by testing with the stereo transmitter from gr-rds, which is also reversed.

How did you arrive at the value of math.pi for the filter gain? Is that a calculated or theoretical value, or an empirical value? If it's empirical, then I would suggest just using a number instead of math.pi. math.pi implies some deeper meaning, which may not exist. Testing here shows that a value of 2.9 may be a little better for stereo separation.

efardin:
It's empirical. It could be argued that because the pilot is scaled by a factor of 10 in the [multiplex signal](https://en.wikipedia.org/wiki/FM_broadcasting#Stereo_FM), rescaling by a factor of 10 in power or sqrt(10) in voltage is required for demodulation. I don't think this is a good argument though.

willcode:
Since the multiply_const is now +/- 1.0, could they be eliminated. They feed into add blocks, and one of them could just turn into a subtract.

dave-pi:
I think the demodulation should work regardless of sending sin() or cos(). I've just tested all of the scenarios:
|     | left=-1<br>imag | left=-1<br>real | right=-1<br>imag | right=-1<br>real |
|-----|-----------------|-----------------|------------------|------------------|
| sin |        v        |        x        |         x        |         v        |
| cos |        x        |        x        |         v        |         v        |

So i think we should switch the coefficients but take the real instead of the imag (last column) since it's the only one that works with sin() or cos(). It also works for anything in between (as long as, in the sender, the initial phase offset of the 38k carrier is double that of the 19k carrier)

duggabe (approved):
I agree with these changes.

efardin:
The real multiplication to downconvert the DSBSC signal centered at 38 kHz will overlap the L-R (S) component with the L+R (M) component from DC to 15 kHz. 
gr_s_path
gr_spectrum

This was avoided in the wfm_rcv_fmdet demodulator by using complex multiplication of filtered signals. There are a couple of problems with the wfm_rcv_fmdet implementation I believe:

1. The gain of the DSBSC filter should be 2, not 10. This makes sense because there is half the power in each sideband of the L-R component. High gain would cause audio distortion.
2. The complex_to_real block should be complex_to_imag.
3. (Note) The gain of 10 in the carrier filter has no effect because the output of the PLL is always in the range [-1,+1]. It does recover the pilot signal back to its full level at the PLL input.

Both wfm_rcv_pll and updated wfm_rcv_fmdet implementations are attached as flowgraphs for testing.

[wfm_rcv.zip](https://github.com/gnuradio/gnuradio/files/5699366/wfm_rcv.zip)

Noted above comments wrt multiply_const and real / imag, will push changes.

drmpeg:
@efardin Be aware that wfm_rcv_fmdet.py is used anywhere.

dave-pi:
I would have thought the multiply would shift the spectrum to be centered
around ±38k. So the mono would be from 38 to 38±15 and from -38 to -38±15.
And the stereo would be at DC.
I'm afk and half asleep so could be not making sense though.

efardin:
Understood that [wfm_rcv_fmdet.py](https://github.com/gnuradio/gnuradio/blob/master/gr-analog/python/analog/wfm_rcv_fmdet.py) isn't used in wfm_rcv_pll.py. I was running some comparisons between the two different FM stereo demodulators while testing.

drmpeg:
@efardin Okay, just making sure that you don't waste time fixing it.

And when I say it's not used anywhere, I mean it can't be instantiated in gnuradio-companion, since there's no .yml binding.

efardin:
@dave-pi Yes, you're absolutely right, the mono gets shifted. 

I think the reason for the unexpected amplitude shift (and different behaviour with sin / cos) is the delay of the filters in the 19 kHz / 38 kHz pilot tone path, which causes phase offset. A delay block can be added to restore the phase alignment. I've attempted to show this below:

gr_delay

mormj:
@drmpeg  - does this one look good to go now?

drmpeg:
No, he's still working on it.

dave-pi:
I've done a bit more investigation in the way of compensating for the stereo path:

- I think the receiver 38k sub-carrier needs to be multiplied by 2 before mixing with the input signal, to account for the double-side band in the tx.
- Putting a delay of 308+1156 ((N-1)/2 with N the number of taps) in the l+r path to account for the FIR and FFT filters phase shifts.

With these changes I get a well-balanced l-r and l+r and also unit gain. I'm attaching a flowgraph with the tests I'm making.
[USRP_FM_stereofft3.tar.gz](https://github.com/gnuradio/gnuradio/files/5712646/USRP_FM_stereofft3.tar.gz). I'm using 3840 fft size in the frequency sinks, just to avoid scalloping loss.

efardin:
I've updated wfm_rcv_pll based on comments received. L / R test IQ signals are attached below. They were generated with a R&S SFE broadcast tester and recorded with a USRP N210.
[test_signals.zip](https://github.com/gnuradio/gnuradio/files/5714335/test_signals.zip)

drmpeg:
This is looking very good. Since it's been worked on quite a bit, it would be nice to squash the commits into one (with a nice commit message).

```git rebase --interactive HEAD~4```

File: gr-analog/python/analog/wfm_rcv_pll.py [hunk 1]
```diff
@@ -39,14 +39,13 @@ def __init__(self, demod_rate, audio_decimation):
...
-        self.demod_rate = demod_rate = (int)(samp_rate/rf_decim)
-        self.stereo_carrier_filter_coeffs_0 = stereo_carrier_filter_coeffs_0 = firdes.band_pass(1.0, demod_rate, 37600, 38400, 400, firdes.WIN_HAMMING, 6.76)
+        self.demod_rate = demod_rate
+        self.stereo_carrier_filter_coeffs_0 = stereo_carrier_filter_coeffs_0 = firdes.band_pass(-2.0, demod_rate, 37600, 38400, 400, firdes.WIN_HAMMING, 6.76)
         self.stereo_carrier_filter_coeffs = stereo_carrier_filter_coeffs = firdes.complex_band_pass(1.0, demod_rate, 18980, 19020, 1500, firdes.WIN_HAMMING, 6.76)
         self.deviation = deviation = 75000
         self.audio_filter = audio_filter = firdes.low_pass(1, demod_rate, 15000,1500, firdes.WIN_HAMMING, 6.76)
-        self.audio_decim = audio_decim = (int)(demod_rate/48000)
+        self.audio_decim = audio_decim = audio_decimation
+        self.samp_delay = samp_delay = (int)((len(stereo_carrier_filter_coeffs)-1)/2 + (len(stereo_carrier_filter_coeffs_0)-1)/2)
```
dave-pi, line 48: Python doesn't have c-style cast. Just use the // operator (floor division):
`(len(stereo_carrier_filter_coeffs) - 1) // 2 + (len(stereo_carrier_filter_coeffs_0) - 1) // 2`

efardin: Thanks @dave-pi, I've updated and pushed changes. Also squashed multiple commits into one.

drmpeg:
Unfortunately, the file wfm_rcv_pll.py has been changed by someone else and committed to master. This means you have to rebase your branch. Here are the steps:

1) git checkout master
2) git fetch upstream
3) git merge upstream/master
4) git checkout development
5) git rebase master
6) The merge is too complex to be done automatically, so you have to edit wfm_rcv_pll.py by hand.

image

You have to change the occurrences of ```firdes.WIN_HAMMING``` to ```fft.window.WIN_HAMMING``` in the bottom half of the diff. Then delete the top half and the extra lines with ```<<<<<<< HEAD```, ```=======``` and ```>>>>>>> Fixed orientation of L and R audio. Removed hard coded parameters.```.

7) git add wfm_rcv_pll.py
8) git rebase --continue

Your commit message still has some of the old comments.

9) git commit --amend

dave-pi:
@efardin I was wondering if this is actually the correct way to obtaint the 38k. While what we have works for your test signals, I suspect it will break down if there is an initial phase offset !=0 in the 19k and 38k tones used in the TX (this is easy to demonstrate with a TX flowgraph). I think the way to obtain the 38k is not by multiplying the 19k which will screw up the phase relationship, but to use a PLL with a /2 in the feedback path (essentialy a frequency doubler). This will ensure the initial phase offset is the same in the 19k and 38k, as it is done in the TX. I was trying it out but I'm running into some problems with the flowgraph freezing. Our PLL_ref_out block doesn't support dividers in the feedback path so I was trying a 'keep 1 in N' block (with N=2) to decimate by 2 but that will freeze up, because the multiplier that brings the L-R to baseband is now seeing 2 different sample rates...(384k from the quad_demod and 192k from the 38k)
This could well be a figure of my imagination but if you could convice your tester to output something with an initial phase offset would be great to test. Any thoughts appreciated...

efardin:
@drmpeg  I've pushed changes after merge + rebase. Thanks for pointing me in the right direction here.
@dave-pi Agreed, there is leakage from the L to R channel and vice versa if there is a phase offset on the 19 kHz pilot (I can set an offset on the generator). However, I think it's safe to assume that the pilot phase is well controlled in FM broadcast transmitters.

gr-analog/python/analog/wfm_rcv_pll.py
```diff
@@ -61,15 +60,14 @@ def __init__(self, demod_rate, audio_decimation):
...
-        self.blocks_multiply_const_vxx_0 = blocks.multiply_const_ff(-5.5)
         self.blocks_complex_to_imag_0 = blocks.complex_to_imag(1)
-        self.blocks_add_xx_0_0 = blocks.add_vff(1)
+        self.blocks_sub_xx_0 = blocks.sub_ff(1)
         self.blocks_add_xx_0 = blocks.add_vff(1)
         self.analog_quadrature_demod_cf_0 = analog.quadrature_demod_cf(demod_rate/(2*math.pi*deviation))
         self.analog_pll_refout_cc_0 = analog.pll_refout_cc(0.001, 2*math.pi * 19200 / demod_rate, 2*math.pi * 18800 / demod_rate)
-        self.analog_fm_deemph_0_0 = analog.fm_deemph(fs=48000, tau=75e-6)
-        self.analog_fm_deemph_0 = analog.fm_deemph(fs=48000, tau=75e-6)
+        self.analog_fm_deemph_0_0 = analog.fm_deemph(fs=demod_rate//audio_decim, tau=75e-6)
```
dave-pi, line 68: fs is a float so no need to convert to integer. It should be whatever comes out of demo_rate/audio_decim so that the fm_deemph filter response is correct. Also maybe create an audio_rate variable right after demod_rate instead of doing the divides here in the argument list.

efardin: I've added the audio_rate variable, and also added deemph_tau as a parameter, since the de-emphasis time constant will be different between US and EUR locales.

efardin:
@drmpeg this should be good to go now. Let me know if it needs any more work...

drmpeg (approved):
Looks good.

duggabe:
will this be back-ported to maint-3.9 and maint-3.8 branches?

mbr0wn:
Well look at this PR. Massive (and interesting) discussion, and a bunch of approvals. I think this *should* get backported.

mbr0wn:
Merged manually -- the commit messages weren't formatted according to our coding conventions, but I didn't want to delay this any further by requiring more intervention.

mbr0wn:
If someone wants a maint-3.8 backport, they'll need to provide a PR. Cherry-picking these PRs has conflicts.

PR #3993, opened 2020-12-12
gr-digital: Update docs for MPSK SNR Estimator

ThomasHabets:
I'm submitting my question about what this means and how it should be used as a documentation change. I don't know if this is correct, so the main thing to review here is if what I added is correct or not.


duggabe (approved):
Looks fine to me.

777arc:
These changes have now been reflected in the wiki block doc as well, https://wiki.gnuradio.org/index.php/MPSK_SNR_Estimator_Probe

Thanks!

777arc (approved):

mbr0wn:
Also backported to 3.7 and 3.8.

PR #3994, opened 2020-12-14
grc: indent multiline connections in py flowgraph generation

miek:
This fixes flowgraph generation when using connection templates with
multiple lines. Previously it would only indent the first line and
subsequent lines would have no indentation, causing a python error.

### DIFF
--- a/grc/core/generator/flow_graph.py.mako
+++ b/grc/core/generator/flow_graph.py.mako
@@ -217,7 +217,7 @@ gr.io_signature.makev(${len(io_sigs)}, ${len(io_sigs)}, [${', '.join(size_strs)}
         # Connections
         ${'##################################################'}
         % for connection in connections:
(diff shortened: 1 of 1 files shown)


mormj:
@miek - can you please give an example that would cause this issue before the PR?

miek:
Sure, here's the domain definition I'm using that was breaking before this patch: https://github.com/greatscottgadgets/amalthea/blob/master/amalthea/gnuradio/wishbone.domain.yml
Here's another that was working fine without the patch as it's only single-line: https://github.com/greatscottgadgets/amalthea/blob/master/amalthea/gnuradio/hybridsdr.domain.yml

mormj (approved):
This one looks fine.  Doesn't seem to break any of the grc files that I tried.

mbr0wn (approved):

PR #3995, opened 2020-12-16
digital: deprecate clock recovery blocks covered by symbol_sync

mormj:
The pfb, msk and mm clock recovery blocks are all covered by the
symbol_sync block.  

See https://www.youtube.com/watch?v=uMEfx_l5Oxk


willcode (commented):
msk_timing_recovery was not in the grc tree so didn't get moved, and didn't get the updated `ingroup` in its header. Does this matter?

mormj:
That one seems to be showing up in deprecated for me:
image

willcode:
It didn't show up in the tree before the PR unless you searched for it. The ultimate in deprecated - unlikely anyone even found it! Good now.

willcode (approved):

marcusmueller:
Yep, deprecation is the right thing to reduce confusion and to help focus development!

PR #3996, opened 2020-12-16
Remove things marked as deprecated to be removed in 3.8

mormj:
The win_type in firdes and all its references

### DIFF
--- a/docs/usage-manual/(exported from wiki) GNURadioCompanion.txt
+++ b/docs/usage-manual/(exported from wiki) GNURadioCompanion.txt
@@ -185,7 +185,7 @@ Note: Most blocks with a taps parameter automatically import the firdes module. 
 Enter the following into the taps parameter of a filter block:
 
 <pre>
-firdes.low_pass(1.0, samp_rate, 1000, 100, firdes.WIN_HAMMING)
+firdes.low_pass(1.0, samp_rate, 1000, 100, fft.window.WIN_HAMMING)
 </pre>
 === Use Taps from a File ===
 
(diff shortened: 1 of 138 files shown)


willcode (commented):
- The user may wonder why they have to get window type from fft, where fft is not being used for any other (visible) purpose. This isn't related to this PR, as the decision was apparently made in the past. It makes sense to put the enum in one place, though, I'm just not sure that fft is the obvious place to look for it.
- Running `grep -rl firdes.WIN .` on the code base turns up a lot of in-tree places where Python and YML (aka, untested by CI) code would need to be changed.

mormj:
> It makes sense to put the enum in one place, though, I'm just not sure that fft is the obvious place to look for it.

Excellent point, but you're right that that decision was made to put it there.  I'm guessing part of the decision was that filter module depends on fft, and fft code needs the window?

>Running grep -rl firdes.WIN . on the code base turns up a lot of in-tree places where Python and YML (aka, untested by CI) code would need to be changed.

That is a lot of references, and I suspect this PR will need a lot of manual testing to verify all these changes that fall outside of CI.

willcode (approved):
Looks good, built, no related tests failed. Still remaining to be converted:
```
./gr-filter/examples/interpolate.py
./gr-filter/examples/gr_filtdes_live_upd.py
./gr-filter/examples/reconstruction.py
./gr-filter/examples/channelize.py
./gr-filter/examples/chirp_channelize.py
./gr-filter/examples/decimate.py
./gr-analog/examples/fm_demod.py
./gr-analog/examples/fmtest.py
./gr-qtgui/examples/pyqt_example_f.py
./gr-qtgui/examples/pyqt_freq_f.py
./gr-qtgui/examples/pyqt_freq_c.py
./gr-qtgui/examples/pyqt_example_c.py
./gr-qtgui/examples/pyqt_waterfall_f.py
./gr-qtgui/examples/pyqt_waterfall_c.py
```
and we'll have to see what the "beta testers" find!

mormj:
Not sure how I missed those.

There is one commit in here I'm going to separate out because in current master fft.window.WIN_* became fft.win_type.WIN_* due to the coding in the pybind files.  This definitely needs to go in before maint-3.9 is split because it represents an unnecessary API change from 3.8 to 3.9

marcusmueller (approved):
Looks fine; this is good progress to good naming and sane typing :) !

PR #3998, opened 2020-12-17
FFT window type enums

mormj:
In 3.8 the window enums were placed under fft.window, but in the pybind transition, these got wrapped as fft.win_type.WIN_*

This is an unnecessary departure from the 3.8 API, and this restores the 3.8 API.

This is a subset of #3996 as this is the more critical part of that PR


mbr0wn (approved):

marcusmueller (approved):

marcusmueller:
Clearly the right thing, definitely the right time.

PR #4000, opened 2020-12-18
digital: ofdm: Fix whitespace in OFDM chanest block

mbr0wn:
Both the C++ and Python files have really weird whitespace issues,
making them hard to read. This makes the whitespace usage for consistent
with usual formatting, and fixes some Doxygen issues.

### DIFF
--- a/gr-digital/include/gnuradio/digital/ofdm_chanest_vcvc.h
+++ b/gr-digital/include/gnuradio/digital/ofdm_chanest_vcvc.h
@@ -17,10 +17,9 @@
 namespace gr {
 namespace digital {
 
-/*!
- * \brief Estimate channel and coarse frequency offset for OFDM from preambles
+/*! Estimate channel and coarse frequency offset for OFDM from preambles
  * \ingroup ofdm_blk
- * \ingroup syncronizers_blk
+ * \ingroup synchronizers_blk
  *
  * Input: OFDM symbols (in frequency domain). The first one (or two) symbols are expected
  *        to be synchronisation symbols, which are used to estimate the coarse freq offset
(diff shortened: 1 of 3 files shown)


drmpeg:
This is just a general comment. I noticed you changed ```/usr/bin/env python``` to ```/usr/bin/env python3```. There are many (in the 400 range) instances of this in the code base. Should they all be changed?

marcusmueller:
thanks! Sadly, now you'll have to recompute the hashes in the bindings... tell me if you want us to do that for you.

marcusmueller:
> This is just a general comment. I noticed you changed `/usr/bin/env python` to `/usr/bin/env python3`. There are many (in the 400 range) instances of this in the code base. Should they all be changed?

@drmpeg  good question; technically, doesn't matter; all python goes through GrPython.cmake and that regex-replaces '^#!.*$' with whatever was detected or set for `cmake` explicitly.

marcusmueller:
@drmpeg now, after checking,

>  all python *should* go through GrPython and get regex'ed

mbr0wn:
> This is just a general comment. I noticed you changed `/usr/bin/env python` to `/usr/bin/env python3`. There are many (in the 400 range) instances of this in the code base. Should they all be changed?

That was just a reflex tbh. If we want to change them, not here. I think we might want to do that, because that allows running qa_*.py files without them being re-she-banged. But not in here.

mbr0wn:
> thanks! Sadly, now you'll have to recompute the hashes in the bindings... tell me if you want us to do that for you.

Nah I got it. Didn't run the compiler, since I didn't change any code... tsk tsk.

marcusmueller (approved):

PR #4002, opened 2020-12-18
Clean up doxygen for latest commits.

drmpeg:


mormj (approved):
Thank you for catching this!

PR #4003, opened 2020-12-18
Bump versions

marcusmueller:


balister (commented):
Can we set gcc to 9.3.0? This is what is in The Yocto Project Dunfell release that will be supported for another year or so. I'm aware this has a bug that breaks fftw, but I hope to be able to find a patch from gcc that fixes the fftw issue one day. All the other bumps in this set look great.

CMakeLists.txt
```diff
@@ -52,16 +52,16 @@ SET(VERSION_PATCH 0-git)
 include(GrVersion) #setup version info
 
 # Minimum dependency versions for central dependencies:
-set(GR_BOOST_MIN_VERSION "1.58")
-set(GR_CMAKE_MIN_VERSION "3.8.0")
-set(GR_MAKO_MIN_VERSION "0.4.2")
-set(GR_PYTHON_MIN_VERSION "3.6.5")
+set(GR_BOOST_MIN_VERSION "1.67")
```
mbr0wn, line 55:
```suggestion
set(GR_BOOST_MIN_VERSION "1.65")
```
I do like Boost bumps, but these two versions don't make such a difference, and we get 18.04 for free.

mbr0wn (commented):
I would generally suggest an Ubuntu 18.04-compatible set of deps. That means CMake 3.10.2, gcc 7.4, Python 3.6.7.

Now if you know of a change in any of those deps that we really need, that's a different story. But I think our dependencies are in OK shape. I think we might still all be traumatized from the 3.7 deps, which we carried for so long. But 18.04-compatible deps are still OK minimums. Of course, I would *prefer* a newer compiler and Python (the rest, I really don't care about), but I think we can make do just fine with those 18.04 minimums without any sacrifice.

CMakeLists.txt
```diff
@@ -52,16 +52,16 @@ SET(VERSION_PATCH 0-git)
...
+set(GR_PYTHON_MIN_VERSION "3.7.3")
 set(GR_NUMPY_MIN_VERSION "1.13.0")
-set(GCC_MIN_VERSION "5.0.0")
-set(CLANG_MIN_VERSION "3.4.0")
+set(GCC_MIN_VERSION "8.3.0")
+set(CLANG_MIN_VERSION "7.0.1")
 set(APPLECLANG_MIN_VERSION "500")
 set(MSVC_MIN_VERSION "1900")
-set(VOLK_MIN_VERSION "2.1.0")
+set(VOLK_MIN_VERSION "2.3.0")
```
jdemel, line 64: Could we use VOLK 2.4.1 here?

marcusmueller: If we keep it down to 2.3.0, we can build GNU Radio on Ubuntu 20.10 without building VOLK from source, which would be a benefit. 

For GR3.10 development, we'll directly be upgrading to requiring 2.4.1.

CMakeLists.txt
marcusmueller: you can `apt install gcc-8` to get GCC 8.4on Ubuntu 18.04, so that shouldn't be a blocker.

about Boost: yeah, 1.65 doesn't really hurt.

Generally: I don't consider Ubuntu 18.04 a viable target; it makes all future maintenance harder, for little added benefit: People who want to use new GNU Radio in 2021 can certainly use the *current* LTS ubuntu.

That being said, of course if we can keep retrocompatibility at little added cost, we could go for it – but there's real overhead in CI, Pybombs, and *forwards* compatibility; my no. 1 reason for wanting Py3.7, for example, is `async`; something we're not using at all currently (and it's unlikely a lot of things using maint-3.9 will magically start doing that), but if an OOT wants to be both GR3.9 and 3.10/master compatible, it cannot use that feature that's really useful to GUI and Web things.

So, I don't agree on the "we really don't need anything in 20.04 that 18.04 doesn't have"; it's just that it was not obvious at this point. You're right, I might be a little paranoid from GR3.7 experience, but then again, the whole hybrid Python dependency in GR3.8 was and is a nightmare, too.  Please don't forget that our maint-* branches do have a longish lifetime after all; if I'm asked to pick whether I want to be able to build something two years longer into the future or with a system from 2 years longer ago, I'm picking the future.

marcusmueller:
Don't merge in current master, but directly into `maint-3.9` if done.

marcusmueller:
So, after discussion with #development, we'll go for (reason)

* GCC 8.3 (Buster)
* Python 3.6.7 (Ubuntu 18.04)
* Boost 1.65 (Ubuntu 18.04)
* VOLK 2.4.1 (need >= 2.3.0, 2.4.1 brings "CPU features", which is a bug avoidance pinnacle)
* Mako 1.0.7 (Buster, 18.04)

for GR 3.9.

For master after branching off, I'd propose

* GCC 10.1
* Python 3.9.0
* Boost 1.71
* VOLK 2.4.1 or significantly later

jdemel:
I suggested to go with VOLK 2.4.1. 
2.3.0 is available in Ubuntu 20.10 which, unlike 20.04 (with VOLK 2.2.1), will run out of support in about 6 months.
The next Debian release will include 2.4.1. This seems like a good choice to target.
We released VOLK 2.4 with cpu features to detect cpu extensions that we rely on. Previously there were bugs in our cpu feature detection code that affected AVX(2) etc. I'd expect this fix to become more important over time.

PR #4004, opened 2020-12-18
QTGUI: Toggle switch was reversed. 

### LINKED ISSUE #3939: QT GUI Toggle Switch is backwards

willcode:

Fixes #3939.


duggabe (approved):
LGTM  Thanks!

mormj:
@willcode - can you please edit the commit message to include the module and the nature of the fix?  the change looks good.

mormj (approved):

PR #4010, opened 2020-12-19
Bump dependency versions to ease 3.10 development

marcusmueller:


ThomasHabets:
Any features missing from the version in current Debian Stable (Buster)? I realize Bullseye may be out by the time 3.10 is out, but it would make it easier for developers (specifically me, at least) if the requirements fit Debian Stable.

mormj (approved):

marcusmueller:
Thomas, I must admit that I was expecting you to be most in favor of pushing ahead on compiler versioning. So, we don't have to merge this PR *today* in this shape – we just need to agree on a "common development platform" for everyone to be able to use the feature they want.

That being said, I really wanted full C++17 to be available to avoid doing what I did last cycle, namely raising the C++ standard at the end, when it's least useful.

Now, technicaly, GCC 7 gives us almost full C++17 + stdlib (exceptions: tiny bits of std::filesystem, parallelism TS, C11 as the ref C standard, but we're using that internally anyways).

How about this: we first heave master onto the same dependency versions as maint-3.9, and then figure out when to progress to bullseye, giving everyone chance to use the same platforms for now, but making clear that we might be upgrading dep versions before releasing 3.10.

CMakeLists.txt
```diff
@@ -125,7 +125,7 @@ ENDIF()
 
 set(CMAKE_C_EXTENSIONS OFF)
 set(CMAKE_CXX_EXTENSIONS OFF)
-set(CMAKE_CXX_STANDARD 14)
+set(CMAKE_CXX_STANDARD 17)
 set(CMAKE_C_STANDARD 11)
```
jdemel, line 129: Would it make sense to bump this to C17 as well? [C17 is a bugfix release](https://stackoverflow.com/questions/47529854/what-is-c17-and-what-changes-have-been-made-to-the-language). For new development, I'd argue it'd make sense to rely these bugfixes.

marcusmueller: I only checked whether the proposed compilers offered C++17 support, not C17; a cursory glance at Wikipedia tells me that, yes, even with maint-3.9 compilers we can do that, probably. I really don't expect it to *change* anything, though.

maitbot: CMake has no concept of C17.

ThomasHabets:
Yes, going C++17 from the beginning sounds great. If there's any compromises that we have to make to fit in current Debian Stable I agree that we should go with the higher version.

I would just prefer that an arbitrary version check of libraries should maybe not be chosen higher than that *unless* they have features we want to use.

E.g. for 3.9 we took a while to say "C++14", and it caused some wasted time with `make_unique`, and I think saying C++17 now is great. But unless we see a hurdle with specific library versions, it seems needlessly strict to pick specific versions that are not yet in Debian Stable before there's a reason.

Or more generally: dependency versions I think should be chosen based on what we need, not on what some people have.

We can say "3.10 will at least not require anything newer than X" as a committment, but setting that as a minimum ahead of time seems premature.

But maybe they can be set high, and I'll patch the versions down in all my branches.

CMakeLists.txt
nickoe: Re. C17, https://gitlab.kitware.com/cmake/cmake/-/issues/17755  I am not sure why a fix has not been made in cmake. (I only cross checked the doc)

nickoe:
@ThomasHabets  So what specific min versions do you propose? It does not even seem that libvolk 2.1.0 (currently required) is available in debian stable (10).

ThomasHabets:
Volk is easy to install in the same directory as gnuradio. Python dependencies and such tend to have bigger problems, and a world of transitive dependencies.

Like I said, in my opinion it should be done by need. Do we have a need, or expect to have one, for version X?

jdemel:
This is kind of a chicken and egg problem. If I know I develop a feature etc. for GR 3.10 and I can rely on newer dependency versions, I'm gonna use them. If I need to start a discussion if I really need a newer version because I want to use feature X, I might work around it because this discussion seems to be difficult and time consuming.
I suggest to find a middle ground here. Something between let's set the minimum requirements to "bleeding edge" and only bump if strictly necessary.

nickoe:
What about: If we want c++17 support, then just bump the compilers to the appropriate versions. As as volk is maintained by the gnuradio project mostly and "easy" to install, then bump that as well and leave the others as is.

ThomasHabets:
That's what I would propose. But if any of the others are unreasonably old, then take a look at bumping those too, under the assumption that surely there's *something* that is sufficiently improved to warrant it.

mbr0wn (approved):
I think this is a tad aggressive, but who knows when 3.10 will be released.

marcusmueller:
To update the status: I've brought master to the same dependencies as 3.9. The bump to C++17 does, effectively, only require slightly more modern MSVC; so all these are more of a future "shopping list" right now.

PR #4014, opened 2020-12-20
WindowsAudio - fix error-reporting

gvanem:
Basically replace `strerror(errno)` with `GetLastError()` and update the string-format.

### DIFF
--- a/gr-audio/lib/windows/windows_sink.cc
+++ b/gr-audio/lib/windows/windows_sink.cc
@@ -91,7 +91,7 @@ windows_sink::windows_sink(int sampling_freq,
     d_wave_write_event = CreateEvent(NULL, FALSE, FALSE, NULL);
     if (open_waveout_device() < 0) {
         GR_LOG_ERROR(d_logger,
-                     boost::format("open_waveout_device() failed: %s") % strerror(errno));
+                     boost::format("open_waveout_device() failed: %d") % GetLastError());
         throw std::runtime_error("audio_windows_sink:open_waveout_device() failed");
     } else {
         GR_LOG_INFO(d_debug_logger, "Opened windows waveout device");
@@ -184,8 +184,8 @@ int windows_sink::work(int noutput_items,
                                      d_buffers[i]->dwFlags);
                 }
                 GR_LOG_ERROR(d_logger,
-                             boost::format("no audio buffers available: %s") %
-                                 strerror(errno));
+                             boost::format("no audio buffers available: %d") %
+                                 GetLastError());
                 return -1;
             }
         }
@@ -214,7 +214,7 @@ int windows_sink::work(int noutput_items,
             break;
         }
         if (write_waveout(chosen_header) < 0) {
-            GR_LOG_ERROR(d_logger, boost::format("write failed: %s") % strerror(errno));
+            GR_LOG_ERROR(d_logger, boost::format("write failed: %d") % GetLastError());
         }
         samples_sent += samples_tosend;
     }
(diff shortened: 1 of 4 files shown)


gr-audio/lib/windows/windows_sink.cc
```diff
@@ -91,7 +91,7 @@ windows_sink::windows_sink(int sampling_freq,
     d_wave_write_event = CreateEvent(NULL, FALSE, FALSE, NULL);
     if (open_waveout_device() < 0) {
         GR_LOG_ERROR(d_logger,
-                     boost::format("open_waveout_device() failed: %s") % strerror(errno));
+                     boost::format("open_waveout_device() failed: %d") % GetLastError());
```
jdemel, line 94: I assume this could be updated to [std::strerror](https://en.cppreference.com/w/cpp/string/byte/strerror). Besides, `GetLastError` seems to be a non-standard function. We should avoid such functions wherever possible. Or put its usage in another function and make it depend on preprocessor macros.
I agree with your suggestion in your other PR to remove a "GNU-ism". Similarly, I suggest to avoid any "MSVC-isms".

gvanem: You did note this file is for Windos only?

jdemel: Yes. I'm aware of this. Still, I'd insist on sticking with C++ and not use custom compiler functions. This is a maintenance burden for others to look over it. It'll make it more difficult to refactor, e.g. if it were possible to merge different audio implementations for different OSes in the future. Or to consolidate those.

marcusmueller: I'm not as strict here, since this is really platform-specific, and the platform here simply doesn't do `std::strerror` as far as I understand. @gvanem, did I get that correctly?

gvanem: @marcusmueller I'm just assuming `std::strerror (GetLastError())` prints the wrong text. But I'll leave this for someone knowing STL better than I do to decide.

marcusmueller: @gvanem not an experienced Win32 API coder myself, but yes, there's no reason to assume that good ole' POSIX/C89 strerror "understands" windows sound system error codes.


Other question: I've tried to figure out what even might be in GetLastError, since https://docs.microsoft.com/en-us/windows/win32/api/mmeapi/nf-mmeapi-waveoutopen doesn't mention the called waveOutOpen function would set a global error flag; instead, it would just return an error code.

So, honestly, kind of confused here: what's the thing this is solving?

gvanem: Oh I see. So we should do what PortAudio does:
https://github.com/PortAudio/portaudio/blob/master/src/hostapi/wmme/pa_win_wmme.c#L313

marcusmueller: If we want proper error reporting on windows, that seems indeed to be the right way, @gvanem. But: seeing that portaudio does that better than our Windows source... hm; maybe the proper way would just be deprecating or even removing the windows audio sink/source, since we have the well-working portaudio infrastructure. I don't want to do that without consulting actual audio-on-windows users, though.

So here goes: Gisle, how would you feel if we just removed these sources altogether and forced windows users to go through portaudio?

gvanem: @marcusmueller 👍

marcusmueller:
OK… this is a bit hard: every commit is now there four times on your branch, it seems? I honestly don't know what happened there?

mbr0wn:
@gvanem OK I've rebased this and moved the PR to #4261. But please make sure to use regular git workflows. That means:
- Separate your commits appropriately
- Sign off your commits
- Make sure the git commit message subject line is explicit, and starts with the subsystem (`audio:` or something like that)

We won't have time to fix everyone's conflicts and rebase for them. I think this change is good, and I want to clear the queue, but usually, that means I'll just close these PRs.

PR #4016, opened 2020-12-21
filter: update decimation after gcd calculation

mormj:
When interp/decim was calculated in python, the d_decimation value could be set in the constructor.  Now it needs to be set after the calculation in the constructor.


mrjacobagilbert (approved):
LGTM, important bug fix. Tested against a number of failure cases per discussion on matrix chat. There are no public setters for decimation or interpolation so setting this in the constructor is sufficient.

drmpeg (approved):
Tested and works good.

duggabe (approved):
I have tested it and it works for me.

mrjacobagilbert:
While I think this fix should be merged now as-is, for what it's worth it looks like the QA for this block could use some work both to fix some apparent existing issues and make sure something like this is caught.

mbr0wn (approved):

PR #4017, opened 2020-12-21
Backport: update rational resampler decimation value after GCD

mormj:


mbr0wn (approved):

drmpeg:
This PR was mistakenly committed to master instead of maint-3.9. It's a crucial bug fix, and really needs to be in maint-3.9 before release of GNU Radio 3.9.

mormj:
Should be covered now in #4053.  Agreed - this is a critical fix that needs to be in the release

Issue #4015, opened 2020-12-21
Broken PFB Channelizer flowgraph


mormj:
![image](https://user-images.githubusercontent.com/34754695/102778298-53654800-4360-11eb-9e60-fff5516b70f1.png)

Output from one of the channelizer channels in 3.9
![image](https://user-images.githubusercontent.com/34754695/102778360-6ed05300-4360-11eb-87f6-f6340d705793.png)

Output in 3.8 looks fine
![image](https://user-images.githubusercontent.com/34754695/102778404-81e32300-4360-11eb-8b78-1657c67dd632.png)

drmpeg:
If you use custom taps (low-pass filter), it works okay.

willcode:
Replacing the Hierarchical Polyphase Channelizer block with the Polyphase Channelizer block works. Why do we have both of these?

mormj:
Not sure the reasoning there.

Taking the lazy approach in the background of trying different commits:
a56c3a7f4908133dc6ceb93ffd3b2f9e1dda8e0e - hosed
1d60888071 — good

mormj:
572428babb6db488cc17bfe2ada63ffc23a50d63 appears to be the culprit

willcode:
That's what caused the problem addressed in https://github.com/gnuradio/gnuradio/pull/4021 too.

PR #4021, opened 2020-12-22
pfb.py: fix variable in static function

willcode:
When create_taps() was moved to a static function, a _self reference
tagged along. Use function parameter as intended.

Note: this was found while looking at Issue #4015, but it does not fix the issue.

### DIFF
--- a/gr-filter/python/filter/pfb.py
+++ b/gr-filter/python/filter/pfb.py
@@ -66,7 +66,7 @@ class channelizer_ccf(gr.hier_block2):
         ripple = 0.1
         while True:
             try:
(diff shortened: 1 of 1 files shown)


mormj (approved):
LGTM!

drmpeg (approved):
That line is never executed. You can't set taps = None in the block (because the .yml dtype is real_vector).

willcode:
The Polyphase Channelizer block (not the hier one) takes `[]` but not `None`. This line is executed in this case, and bombs.

drmpeg:
Doh! One of these days, I'll learn Python.

mbr0wn (approved):

PR #4022, opened 2020-12-22
Discussion/Don't Merge: Add new in tree namespace (gr-message or gr-pdu)

jacobagilbert:
The gr-message in tree module is intended to house message (including PDU) based blocks, and enable significant expasion of GNU Radio's message API without cluttering the catch-all . This commit builds and passes QA, but there is still more work to do before its up to the standard for an in-tree block repository (e.g.: examples, documentation, doxygen build, etc).

See [GREP 0020 proposal here](https://github.com/gnuradio/greps/pull/29).

The intended order of operations here is to:
1) create new in tree `message` module (does not break API, can be done now/3.10)
2) pull in new message blocks per GREP (does not break API, can be done now/3.10)
3) move existing message blocks (breaks API, could wait for 4.0)

Questions:
1) I put this together quickly to get the conversation started but before I go too much further, is this even a good idea? Obviously I think so as I have found a ton of use for enhancing the message API but its something that requires input.
2) Are the proposed blocks reasonable to bring in-tree? I think so... i have a few more actually but will take a little time to sort though them and make some tweaks to names and other API breaking tweaks that have been deferred.
3) Is `gr-message` a good name? `gr-msg`? I think a single module for PDU-->PDU, Stream/PDU conversion, and general message tools makes sense for now, though i could see separating PDU specific blocks if using PDUs for data plane processing sees traction.


mrjacobagilbert:
One reservation I have is that the name `gr-message` feels somewhat lacking for a module to house what is mostly PDU processing blocks, a specific subset of messaging blocks. Another option for this would be to create the namespace `gr-pdu` specifically for blocks that operate on PDU objects, PMT pairs of dict/uvec type objects. 

Of the 34 blocks the GREP proposes moving into gr-message now, 26 of them are specifically PDU processing blocks, and I think this is an area where there will be and should be growth; I have more PDU based processing blocks to upstream and there are a number of missing blocks.

I think the existing and 5 new non-PDU message blocks discussed for inclusion in gr-message, which are mostly debug and developer-tool type blocks, could remain in blocks (which will be substantially cleaned out after the rest of the GREP is addressed). I don't see non-PDU based message blocks growing significantly, and if it does maybe `gr-messaging` could be added at that point.

In many ways, this is preferable to what this PR currently proposes; either way this PR should not be merged its just to move discussion forward.

jacobagilbert:
OK - I'm going to close this as it has served the purpose of getting some feedback:

1. Yes, this is acceptable
2. Block list to bring in seems reasonable
3. Some don't cares, some felt gr-pdu is a little better. I think gr-pdu is better for the reasons above, so I'll go with that.

PR #4027, opened 2020-12-23
pfb.py: fix variable in static function

willcode:
When create_taps() was moved to a static function, a _self reference
tagged along. Use function parameter as intended.


mbr0wn (approved):

mbr0wn:
Also forward-ported to master.

PR #4028, opened 2020-12-23
qtgui: toggle switch was reversed

### LINKED ISSUE #3939: QT GUI Toggle Switch is backwards

willcode:
Fixes #3939.


mbr0wn:
@willcode Shouldn't this go onto master?

mbr0wn:
Confirmed bug and fix.

Issue #4034, opened 2020-12-31
qtgui: Number sink average not working?


jaredd:
The qtgui number sink has an Average setting that ostensibly sets the alpha for a single-pole iir filter to average inputs before displaying them. Testing this setting against an external single-pole iir average shows a stark difference in behavior. 

Setup: See the flowgraph snapshot below. Gaussian source with amplitude 1 into a square function, feed two number sinks directly, one number sink through a single-pole iir filter block. One directly connected number sink has its Average parameter set to the same value as the alpha parameter in the single-pole iir. All other number sinks have Average parameter set to 0.

![qtgui_number_sink_Moment](https://user-images.githubusercontent.com/3237083/103422894-fdaf4d80-4b60-11eb-8583-b9ab3b8bf20e.jpg)

Output: The External average displays the expected value while other number sinks show very different numbers that fluctuate as if no averaging were being applied. That behavior was only expected on the No average sink.

Expected: Both the External average and Average sinks show similar numbers approximately equal to 1 with little fluctuation. The No average sink should show a fluctuating value.

The short video below shows the relative fluctuation of the number sink with internal averaging and no averaging vs external averaging.

https://user-images.githubusercontent.com/3237083/103422990-7a422c00-4b61-11eb-8349-9946b0726b73.mp4

A cursory review of the work function seems like a reasonable implementation so I'm kind of scratching my head.

```bash
$ uname -a
Linux system76-pc 5.8.0-7630-generic #32~1605108806~20.10~7e52b13-Ubuntu SMP Wed Nov 11 19:10:30 UTC  x86_64 GNU/Linux

$ gnuradio-config-info --print-all
/home/innerdev/pybombs
/home/innerdev/pybombs/etc
/home/innerdev/pybombs/etc/gnuradio/conf.d
/home/innerdev/.gnuradio

testing-support;python-support;volk;gnuradio-runtime;gr-ctrlport;gnuradio-companion;gr-blocks;gr-fec;gr-fft;gr-filter;gr-analog;gr-digital;gr-dtv;gr-audio;* alsa;* oss;gr-channels;gr-qtgui;gr-trellis;gr-uhd;gr-utils;gr_modtool;gr-vocoder;gr-wavelet;gr-zeromq
3.8.2.0
clang version 11.0.0-++20200414100631+5c1d1a62e37-1~exp1~20200414201304.3176  
Target: x86_64-pc-linux-gnu 
Thread model: posix 
InstalledDir: /usr/bin
[...]

PR #4038, opened 2021-01-03
gr-qtgui number sink: Honor block setting of 'avg' in GUI

willcode:
GUI was overriding block parameter (setting to 0) at init.
- Use block parameter for 'avg'
- Add callback for 'avg' to yml

It is still possible to set the avg from the GUI middle click. The menu
value is not updated when avg is set from a block parameter (there is no
slot in the widget to do this). If the user creates a GRC file with a
widget (e.g., entry) to set avg, and also sets it via the menu, the
values shown will not match. However, the value can be successfully set
either way.

Fixes https://github.com/gnuradio/gnuradio/issues/4034
Should be backported to 3.9 and 3.8.

### DIFF
--- a/gr-qtgui/grc/qtgui_number_sink.block.yml
+++ b/gr-qtgui/grc/qtgui_number_sink.block.yml
@@ -241,6 +241,7 @@ templates:
         import sip
     callbacks:
     - set_update_time(${update_time})
+    - set_average(${avg})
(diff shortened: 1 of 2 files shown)


mbr0wn:
Confirmed patch works as intended.

mbr0wn (approved):

PR #4040, opened 2021-01-04
Fix default parameters in Stream_Mux block

adriengit:
fix default parameter 'Length' in Stream Mux block. Cannot connect in the flowgraph otherwise. Stream mux block example is OK as it uses '[3, 2]'.

### DIFF
--- a/gr-blocks/grc/blocks_stream_demux.block.yml
+++ b/gr-blocks/grc/blocks_stream_demux.block.yml
@@ -14,7 +14,7 @@ parameters:
 -   id: lengths
     label: Lengths
     dtype: int_vector
-    default: 1, 1
+    default: (1, 1)
 -   id: num_outputs
     label: Num Outputs
     dtype: int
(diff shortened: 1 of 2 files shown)


mbr0wn (changes requested):
Please do the following:
- Sign off your commit (`git commit -s` or amend your commit message manually)
- Make a single commit on top of master

Otherwise, this is good. Many thanks!

dave-pi:
I think Stream Demux also has the same problem.

adriengit:
Default parameters in Stream Demux are OK

dave-pi:
> Default parameters in Stream Demux are OK

No they're not:

https://github.com/gnuradio/gnuradio/blob/0f1ec604f758e28d788ad71283236928d2d82300/gr-blocks/grc/blocks_stream_demux.block.yml#L17

adriengit:
> > Default parameters in Stream Demux are OK
> 
> No they're not:
> 
> https://github.com/gnuradio/gnuradio/blob/0f1ec604f758e28d788ad71283236928d2d82300/gr-blocks/grc/blocks_stream_demux.block.yml#L17

Change is also done in Demux Block. In 3.8 branch, this block does exist, I was refering to Streams to Stream Block, quite similar.

mbr0wn (approved):

PR #4041, opened 2021-01-04
[Foward Port] qtgui: toggle switch was reverse

### LINKED ISSUE #3939: QT GUI Toggle Switch is backwards

mbr0wn:
Fixes #3939.

Note from @mbr0wn: I forward-ported this to master from #4028.


mbr0wn:
My bad. Missed a PR there.

PR #4049, opened 2021-01-06
Add GH Actions Make Test

mbr0wn:
This is a GH-A based version of our unit tests. No ccache, and dual-core machines, thus takes much longer than BuildBot. But less maintenance overhead. Yes, some amount of vendor lock-in. But less maintenance for now, and we can more easily have different configs per branch.

Requires Docker images to be on Docker Hub. I've put them here: https://github.com/gnuradio/gnuradio-docker/tree/master/ci. Those Dockerfiles could also go into the `.github` dir, but I think they might be useful for other stuff, too.


mbr0wn:
...Debian is still missing. But it's quite easy to add containers.

.github/workflows/make-test.yml
```diff
@@ -0,0 +1,110 @@
+name: 'Make Test'
+on:
+  push:
+    branches:
+      - maint-3.9
```
mbr0wn, line 5: The containers and the deps therein are specific to 3.9, so this is intentional. The same would work, right now, for master. But that should be a separate PR.

.github/workflows/make-test.yml
```diff
@@ -0,0 +1,110 @@
...
+    container:
+      image: 'gnuradio/ci-centos-8.3-3.9:1.0'
+      volumes:
+        - build_data:/build
+      options: --cpus 2
+    steps:
+    - uses: actions/checkout@v2
+      name: Checkout Project
+    - name: CMake
+      run: 'cd /build && cmake ${GITHUB_WORKSPACE} -DENABLE_DOXYGEN=OFF -DBoost_unit_test_framework_FOUND=ON'
```
mbr0wn, line 88: On some distros, CMake thinks that Boost unit testing is not installed. Unsure why, but this is a workaround.

.github/workflows/make-test.yml
```diff
@@ -0,0 +1,110 @@
...
+      options: --cpus 2
+    steps:
+    - uses: actions/checkout@v2
+      name: Checkout Project
+    - name: CMake
+      run: 'cd /build && cmake ${GITHUB_WORKSPACE}'
+    - name: Make
+      run: 'cd /build && make -j2'
+    - name: Make Test
+      run: 'cd /build && ctest --output-on-failure -E "qa_uhd|qa_uncaught_exception|qa_header_payload_demux|qa_agc|qa_cpp_py_binding|qa_cpp_py_binding_set|qa_ctrlport_probes|qa_file_taps_loader|qa_qtgui"'
```
mbr0wn, line 38: The number of tests that need disabling is absurd, but that's a separate matter. Also, I'm not sure if it's necessary to repeat the tests for every distro we're testing, but I'm learning GH-A as I go along.

PR #4054, opened 2021-01-08
Qt shutdown cleanup

smunaut:
So originally I introduced the aboutToQuit() stuff in the hope to cleanly shutdown the flow graph, including the Qt widgets blocks before the Qt stuff is torn down.
    
Turns out that in aboutToQuit some stuff is already invalid, especially in OpenGL and so causes SegFault on exit.
    
Instead this patch modify the logic so that we stop the flow graph before Qt even begin disabling things.
    
The two shutdown path that exits are :
     - The sig_handler in case we get TERM or INT signals
     - The closeEvent() of the main window
    
This makes sure that both path call tb.stop() tb.wait() and also any snippet 'main_after_stop'.

### DIFF
--- a/grc/core/generator/flow_graph.py.mako
+++ b/grc/core/generator/flow_graph.py.mako
@@ -44,6 +44,34 @@ ${imp}
 % endfor
 
 ########################################################
+##Prepare snippets
+########################################################
+% for snip in flow_graph.get_snippets_dict():
+
+${indent(snip['def'])}
+% for line in snip['lines']:
+    ${indent(line)}
+% endfor
+% endfor
+\
+<%
+snippet_sections = ['main_after_init', 'main_after_start', 'main_after_stop']
+snippets = {}
+for section in snippet_sections:
+    snippets[section] = flow_graph.get_snippets_dict(section)
+%>
+\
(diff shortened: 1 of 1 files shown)


mormj:
@smunaut - based on the conversation on chat, it seems this fix is a blocker to getting gr-fosphor working with gr 3.9?

smunaut:
Yeah, it segfaults without it, at least on Intel GPU.

Ideally it would need testing on AMD and NVidia drivers too. I'm working on testing with AMD drivers now.

But the general code flow here seems cleaner since this is calling tb.stop() before any Qt teardown at all. And that should always be safe (since calling tb.stop() is allowed at any time ...).

smunaut:
Update:
Tested on AMD GPU (by me) and on NVidia GPU (by `willcode4[m|gr]` on IRC) and fixes the segfault in both cases.

mbr0wn (approved):

mbr0wn:
@smunaut  are you using the snippets feature at all?

smunaut:
@mbr0wn No I'm not using the snippets features personally. But I did check it still worked after my changes to make sure I didn't break it with the shuffling around.

PR #3892, opened 2020-10-27
filter: remove needless empty destructors and problematic const

ThomasHabets:
You can't move from a const object, so unfortunately it's not possible
to keep constness on these.

This should fix #3757

PR #4055, opened 2021-01-08
filter: Fix polyphase_filterbank for MSVC, w/ default move constructor.

ryanvolz:
Because of the deleted copy constructor (previously implicit, now made explicit in this commit to match fir_filter.h), polyphase_filterbank.cc fails to compile with MSVC trying to use the fft_filter_ccf copy constructor. Explicitly declaring a default move constructor fixes this by allowing it to use the move constuctor instead, which is the behavior we want. It's a mystery why MSVC requires this and other compilers don't, but it works.

See #3892 for similar discussion and change for other filter classes. (This one wasn't caught there because changes from other PRs came in parallel.)

This was needed to get MSVC compilation to work for my [3.9.0.0-rc2 conda package](https://github.com/conda-forge/gnuradio-feedstock/runs/1669068304).

It will also need a backport to 3.9 that I can push when approved.

### DIFF
--- a/gr-filter/include/gnuradio/filter/fft_filter.h
+++ b/gr-filter/include/gnuradio/filter/fft_filter.h
@@ -87,6 +87,15 @@ public:
      */
     fft_filter_fff(int decimation, const std::vector<float>& taps, int nthreads = 1);
 
+    // Disallow copy.
+    //
+    // This prevents accidentally doing needless copies, not just of fft_filter_xxx,
+    // but every block that contains one.
+    fft_filter_fff(const fft_filter_fff&) = delete;
+    fft_filter_fff& operator=(const fft_filter_fff&) = delete;
+    fft_filter_fff(fft_filter_fff&&) = default;
+    fft_filter_fff& operator=(fft_filter_fff&&) = default;
+
     /*!
      * \brief Set new taps for the filter.
      *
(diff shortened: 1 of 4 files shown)


ThomasHabets (approved):

mbr0wn (approved):

guruofquality:
Just for the sake of interest I ran into the same issue, I only needed to delete copy construct and assign on polyphase_filterbank to get the build to fix

```
class FILTER_API polyphase_filterbank
{
protected:
    unsigned int d_nfilts;
    std::vector<kernel::fir_filter_ccf> d_fir_filters;
    std::vector<kernel::fft_filter_ccf> d_fft_filters;
    std::vector<std::vector<float>> d_taps;
    unsigned int d_taps_per_filter;
    fft::fft_complex_rev* d_fft;

public:
    // Disallow copy.
    polyphase_filterbank(const polyphase_filterbank&) = delete;
    polyphase_filterbank& operator=(const polyphase_filterbank&) = delete;
 
 ```

ryanvolz:
That's good to know. @ThomasHabets and I have been stumbling around in the dark trying to figure out why MSVC behaves differently than GCC and Clang in this instance and the others like it from #3892. Do you have a good explanation, or did you also just try things until it worked? I really know nothing, I'm just a poor sap trying to get it to compile.

guruofquality:
All of my build issue showed errors referencing a deleted copy function for kernel::fir_filter_ccf and kernel::fft_filter_ccf. And in polyphase_filterbank the vector of kernel::fir_filter_ccf and kernel::fft_filter_ccf are also therefore uncopyable. I think in MSVC the polyphase_filterbank gets a default copy assign and copy constructor even though it technically cant be copied. And then the error doesnt show up until you compile something inheriting or using the class.

Now a good explanation? I'm also a poor sap :-) I think either MSVC doesn't look deep enough to inherit this "uncopyability" in its default copy functions, or gets messed up by the std::vector indirection. Or maybe, they MS just read the standard differently, and its GCC that took a shortcut. Or this is undefined behaviour and both are correct and the best action is for our source code to be explicit.

ThomasHabets:
@guruofquality yeah that's what I'm thinking, but I don't think I managed to create a minimal example in godbolt to be clear on what's needed.

PR #4057, opened 2021-01-08
cmake: python: Fix linking when runtime py interpreter != build time.

ryanvolz:
I've been using this for a little while in my conda packages, and it's about time some version of it got upstream. In light of possibly requiring CMake 3.16 for 3.10 development, this could probably be simplified. (By requiring CMake 3.16+, it would be possible to use the `Python::Python` and `Python::Module` targets defined by the builtin CMake FindPython module that was added for 3.12 and drop that part of GrPython.cmake.) But I'd rather trigger a discussion before going off and changing things, and at least maint-3.9 would require something closer to this. I can send that PR once this has had a few eyes on it.

So the description of the fix:

This fixes a segmentation fault with the Pybind11 bindings when trying to load any Python module on OSX with a different interpreter than was used at build time (relevant to at least conda and homebrew). This fix was already in place for the prior Swig bindings, but the Pybind11 changes do not incorporate it since they link with ${PYTHON_LIBRARIES} instead of the (fixed) Python::Python target.

For clearer separation and to adopt the approach used by the new-ish (3.12) builtin CMake FindPython module, this commit creates a separate Python::Module target to be used when creating a Python extension module (which includes "-undefined" and "dynamic_lookup" for macOS instead of linking directly to the Python library). The Python::Python target that does link to the Python library is still necessary for use as a PRIVATE target (e.g. for runtime and qtgui).

For future reference and a general description of best practices for linking Python extension modules, see https://blog.tim-smith.us/2015/09/python-extension-modules-os-x.

### DIFF
--- a/cmake/Modules/GrPython.cmake
+++ b/cmake/Modules/GrPython.cmake
@@ -37,14 +37,9 @@ set(PYTHON_EXECUTABLE ${PYTHON_EXECUTABLE} CACHE FILEPATH "python interpreter")
 set(QA_PYTHON_EXECUTABLE ${QA_PYTHON_EXECUTABLE} CACHE FILEPATH "python interpreter for QA tests")
 
 add_library(Python::Python INTERFACE IMPORTED)
-if(APPLE)
-    set_target_properties(Python::Python PROPERTIES
-      INTERFACE_INCLUDE_DIRECTORIES "${PYTHON_INCLUDE_DIRS}"
-      INTERFACE_LINK_LIBRARIES "-undefined dynamic_lookup"
-      )
 # Need to handle special cases where both debug and release
 # libraries are available (in form of debug;A;optimized;B) in PYTHON_LIBRARIES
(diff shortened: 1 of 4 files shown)


ryanvolz:
See #3279 and #3674 for previous issue/PR that fixed this for 3.8, but the switch to pybind11 necessitates additional changes as with this PR for 3.9 and master.

cmake/Modules/GrPython.cmake
```diff
@@ -56,6 +51,35 @@ else()
       )
 endif()
 
+# separarate target when linking to make an extension module, which should not
```
mbr0wn, line 54:
```suggestion
# separate target when linking to make an extension module, which should not
```

mbr0wn (approved):
This all looks reasonable, but I'm just not knowledgeable enough to be certain. That said, I'd be inclined to merge this if it works. I can give it another spin later today.

mbr0wn:
After some cursory review and testing, I've come to the conclusion that the risk of bit rot on this PR is higher than the risk of merging without additional review.

PR #4058, opened 2021-01-09
cmake: Update VERSION_PATCH so that correct versioning is enabled.

drmpeg:
i.e. version - commit number from last tag - last commit hash.


mbr0wn (approved):
Yeah this is what we should be doing. Especially given how much backporting is still happening on 3.8.

PR #4060, opened 2021-01-09
gr-digital: Fixup example_corr_est

nickoe:
In example_corr_est switch to dynamic threshold instead of absolute.
This make the example do something at start up instead of just
flatlining and it seems like the example becomes more interesting.

The tab control were missing a proper GUI hint and was just plced in
the bottom corner.

Add GUI hint to it and adjust the amount of rows the other time sink
consumes to let the sliders not consume that much screen real estate.

Slightly align blocks to look nicer.

---

This fix is very likely applicable on maint-3.9 and master as well. I have not verified it though.

### DIFF
--- a/gr-digital/examples/packet/example_corr_est.grc
+++ b/gr-digital/examples/packet/example_corr_est.grc
@@ -24,6 +24,9 @@ options:
     title: ''
     window_size: ''
   states:
+    bus_sink: false
+    bus_source: false
+    bus_structure: null
     coordinate: [8, 11]
     rotation: 0
     state: enabled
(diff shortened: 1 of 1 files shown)


mbr0wn:
3.9 and above has some issue with the constellation object. It doesn't seem related to this PR, but means it can't be tested there :angry:

nickoe:
@mbr0wn  Now it is a couple of days ago, but I think I tested this on 3.9 and master as well, but keep in mind this is purely patching the grc file. I think this should just be merged on maint-3.8, maint-3.9 and master ASAP if there are no comments about the actual changes.

I can try to recheck when I get home.

mbr0wn:
Yeah this can be merged.

mbr0wn (approved):

PR #4062, opened 2021-01-09
gr-blocks: Check fseek() for failure

ZackerySpytz:
Some fseek() calls were not being checked for failure.


mbr0wn (approved):
Yup, that's how you use fseek(). Thanks!

PR #4069, opened 2021-01-12
Backport changes from master to maint-3.8 (Jan 2021)

mbr0wn:


mbr0wn:
Re DCO fail: There's a backported commit from master from pre-DCO, and it's not mine, so I can't/won't sign it. See no reason why it's not OK though.

PR #4087, opened 2021-01-14
gr-uhd: adding a bool to force tune request

mbr0wn:
This is used if a 'direction' key is provided with the tune command.
This fixes an issue when the 'TX' and 'RX' directions are tuned
explicitly to the same frequency; prior to this fix the second tune
request is dropped.


willcode:
Already backported to maint-3.8, removing label.

PR #4089, opened 2021-01-14
cmake: Fixup BUILD_DATE commit.

drmpeg:


mbr0wn:
I wonder if I messed up the original merge; I thought the code would look something like this.

PR #3485, opened 2020-05-16
gr-uhd: Added message handler for (fully) manual tuning

segdy:
The current message handler of USRP Source/Sink is very sparse. The message handler for "tune" lacks the ability of the original uhd::tune_request_t class. It also sets the policy to "AUTO".
Of particular importance is the "args" part which encodes options for integer-N tuning. Current implementation does not allow for integer-N tuning.

While I would love to fix the "tune" command, it would break current code that relies on "tune" being a pair of dsp_freq and rf_freq.
Hence I added the command "mtune" (="manual tune").
It implements all fields of uhd::tune_request_t completely as a dict (args, dsp_freq, dsp_freq_policy, rf_freq, rf_freq_policy, target_freq).

It would be great if this could be back-integrated into 2.7 as well.

PR #4091, opened 2021-01-14
gr-uhd: Add command uhd handler for mtune ("manual tune")

mbr0wn:
This command implements uhd::tune_request_t completely (and not just a
subset). Also includes updates to the manual.

Supersedes https://github.com/gnuradio/gnuradio/pull/3485

### DIFF
--- a/gr-uhd/docs/uhd.dox
+++ b/gr-uhd/docs/uhd.dox
@@ -92,6 +92,7 @@ Command name | Value Type   | Description
 `freq`       | double       | Sets the Tx or Rx frequency. Defaults to all channels. If specified without `lo_offset`, it will set the LO offset to zero.
 `lo_offset`  | double       | Sets an LO offset. Defaults to all channels. Note this does not affect the effective center frequency.
 `tune`       | tune_request | Like freq, but sets a full tune request (i.e. center frequency and DSP offset). Defaults to all channels.
+`mtune`      | tune_request_t | Like tune, but supports a full manual tune request as uhd::tune_request_t. Defaults to all channels.
 `lo_freq`    | double       | For fully manual tuning: Set the LO frequency (RF frequency). Conflicts with `freq`, `lo_offset`, and `tune`.
 `dsp_freq`   | double       | For fully manual tuning: Set the DSP frequency (CORDIC frequency). Conflicts with `freq`, `lo_offset`, and `tune`.
 `direction`  | string       | Used for timed transceiver tuning to ensure tuning order is maintained. Values other than 'TX' or 'RX' will be ignored.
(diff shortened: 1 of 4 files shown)


segdy:
I just found a small bug that I pushed into `manual_tune_request` but it's not showing up here (yet?): `target_freq` did not update `target_freq` but `dsp_freq`:

```
$ git diff f9931a1eec4d349726cabe4ced63cb51f13061b4
diff --git a/gr-uhd/lib/usrp_block_impl.cc b/gr-uhd/lib/usrp_block_impl.cc
index 7970881aa..278d593fe 100644
--- a/gr-uhd/lib/usrp_block_impl.cc
+++ b/gr-uhd/lib/usrp_block_impl.cc
@@ -700,7 +700,7 @@ void usrp_block_impl::_cmd_handler_mtune(const pmt::pmt_t& tune,
             pmt::to_double(pmt::dict_ref(tune, pmt::mp("rf_freq"), 0));
     }
     if (pmt::dict_has_key(tune, pmt::mp("target_freq"))) {
-        new_tune_request.dsp_freq =
+        new_tune_request.target_freq =
             pmt::to_double(pmt::dict_ref(tune, pmt::mp("target_freq"), 0));
     }
     if (pmt::dict_has_key(tune, pmt::mp("dsp_freq_policy"))) {

```

mbr0wn:
There was also a big bug (the handler was implemented twice). @segdy I've added you to gr-devs; if you accept, I can make you a reviewer for this PR.

gr-uhd/docs/uhd.dox
```diff
@@ -92,6 +92,7 @@ Command name | Value Type   | Description
 `freq`       | double       | Sets the Tx or Rx frequency. Defaults to all channels. If specified without `lo_offset`, it will set the LO offset to zero.
 `lo_offset`  | double       | Sets an LO offset. Defaults to all channels. Note this does not affect the effective center frequency.
 `tune`       | tune_request | Like freq, but sets a full tune request (i.e. center frequency and DSP offset). Defaults to all channels.
+`mtune`      | tune_request | Like tune, but supports a full manual tune request as uhd::tune_request_t. Defaults to all channels.
```
mbr0wn, line 95: Whoops! I badly merged this. Should be tune_request_t.

gr-uhd/docs/uhd.dox
```diff
@@ -104,8 +105,30 @@ Command name | Value Type   | Description
 Special types:
 
 - tune_request: Like a uhd::tune_request_t, but always uses POLICY_AUTO. This is a pair, composed of (target_frequency, lo_offset)
+- tune_request_t: Exact copt of uhd::tune_request_t, allowing full control. See details below
```
mbr0wn, line 108:
```suggestion
- tune_request_t: Exact copy of uhd::tune_request_t, allowing full control. See details below
```

segdy:
@mbr0wn Thanks, I did, that'd be great.

mbr0wn:
@segdy let me know if you think this is OK and click "approve"; we can merge it then. CI seems happy.

marcusmueller (approved):
Checking against `pmt::mp("string")` every time used to come with a *significant* overhead, because that hashes the string (every time), checks for existence of that hash in a hash table (which used to be self-written), adds it to the table if not present, retrieves the `pmt_t` smartpointer stored there, and compares the address (and that all to avoid comparing two strings shorter than 8 characters...)

I think @jacobagilbert would have a more up-to-date view on the computational overhead.

Note that this is a performance bug only, but it's bad enough a practice to manifast a low-prio bug; I think this is good functionality and we should be having it nevertheless, but if we decide to merge it this way, we should open an issue to fix it eventually.

jacobagilbert:
@marcusmueller @mbr0wn this is merged now which I also would have advocated for as-is. Runtime interning of pmt strings is pretty easy to avoid, and it is relatively expensive especially if the hash table is shallow (though that has recently been improved). That said in this case it's probably not a factor at all given the expected frequency of doing this. It becomes an issue if you are doing it on every tag or something like that. 

gr-uhd has pretty good internal usage of getters for PMT constant strings. At some point it would be good to use that, no real reason not to.

PR #4093, opened 2021-01-14
uhd: Control of frequency and gain in both directions at the same moment

mbr0wn:
It implements the ability to control frequency and gain in both
directions (Tx and Rx) with use of a single UHD source or sink - in the
same moment. It is possible to do that in separate moments with current
UHD blocks but this way it is not possible to change both Rx and Tx
settings for the same burst. Now, a single message can do that.

Note that there is a bit of an oddity here: The GNU Radio USRP blocks
are either Rx or Tx (source or sink). However, they all encapsulate an
underlying multi_usrp object. That means that under the hood, a USRP
source or a USRP sink potentially has access to the "other" direction.
This is made use in this patch.

To use this feature, add a direction key to the message dictionary going
to the block.

*note:* This is a new PR for https://github.com/gnuradio/gnuradio/pull/3642. @segdy and @ptrkrysik wrote this, not me. I'm just doing the git wrangling.

### DIFF
--- a/gr-uhd/include/gnuradio/uhd/usrp_block.h
+++ b/gr-uhd/include/gnuradio/uhd/usrp_block.h
@@ -147,8 +147,11 @@ public:
      *
      * \param gain the gain in dB
      * \param chan the channel index 0 to N-1
+     * \param direction TX or RX. This is mostly used by the internal message
+     *        handling.
      */
-    virtual void set_gain(double gain, size_t chan = 0) = 0;
+    virtual void
+    set_gain(double gain, size_t chan = 0, pmt::pmt_t direction = pmt::PMT_NIL) = 0;
 
     /*!
      * Set the named gain on the dboard.
--- a/gr-uhd/lib/usrp_block_impl.cc
+++ b/gr-uhd/lib/usrp_block_impl.cc
@@ -137,8 +137,10 @@ usrp_block_impl::usrp_block_impl(const ::uhd::device_addr_t& device_addr,
       _stream_now(_nchan == 1 and ts_tag_name.empty()),
       _start_time_set(false),
       _force_tune(false),
-      _curr_tune_req(stream_args.channels.size(), ::uhd::tune_request_t()),
-      _chans_to_tune(stream_args.channels.size())
+      _curr_tx_tune_req(stream_args.channels.size(), ::uhd::tune_request_t()),
+      _curr_rx_tune_req(stream_args.channels.size(), ::uhd::tune_request_t()),
+      _tx_chans_to_tune(stream_args.channels.size()),
+      _rx_chans_to_tune(stream_args.channels.size())
 {
     _dev = ::uhd::usrp::multi_usrp::make(device_addr);
 
@@ -292,11 +294,22 @@ bool usrp_block_impl::_check_mboard_sensors_locked()
     return clocks_locked;
 }
 
-void usrp_block_impl::_set_center_freq_from_internals_allchans(pmt::pmt_t direction)
+void usrp_block_impl::_set_center_freq_from_internals_allchans()
 {
-    while (_chans_to_tune.any()) {
+    unsigned int chan;
+    while (_rx_chans_to_tune.any()) {
         // This resets() bits, so this loop should not run indefinitely
-        _set_center_freq_from_internals(_chans_to_tune.find_first(), direction);
+        chan = _rx_chans_to_tune.find_first();
+        _set_center_freq_from_internals(chan, ant_direction_rx());
+        _rx_chans_to_tune.reset(chan);
+    }
+
+    while (_tx_chans_to_tune.any()) {
+        // This resets() bits, so this loop should not run indefinitely
+        chan = _tx_chans_to_tune.find_first();
+        _set_center_freq_from_internals(_tx_chans_to_tune.find_first(),
+                                        ant_direction_tx());
+        _tx_chans_to_tune.reset(chan);
     }
 }
 
@@ -592,8 +605,9 @@ void usrp_block_impl::msg_handler_command(pmt::pmt_t msg)
             break;
         }
     }
+
     /// 5) Check if we need to re-tune
-    _set_center_freq_from_internals_allchans(direction);
+    _set_center_freq_from_internals_allchans();
     _force_tune = false;
 }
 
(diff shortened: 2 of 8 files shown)


marcusmueller:
Build breaks :(

```
/cache/workdir/Ubuntu_20_10_64_0/build_Ubuntu_20_10_64/src/gr-uhd/python/uhd/bindings/usrp_block_python.cc: In function ‘void bind_usrp_block(pybind11::module&)’:
/cache/workdir/Ubuntu_20_10_64_0/build_Ubuntu_20_10_64/src/gr-uhd/python/uhd/bindings/usrp_block_python.cc:93:67: error: no matches converting function ‘set_gain’ to type ‘void (class gr::uhd::usrp_block::*)(double, size_t)’ {aka ‘void (class gr::uhd::usrp_block::*)(double, long unsigned int)’}
   93 |              (void (usrp_block::*)(double, size_t)) & usrp_block::set_gain,
      |                                                                   ^~~~~~~~
In file included from /cache/workdir/Ubuntu_20_10_64_0/build_Ubuntu_20_10_64/src/gr-uhd/python/uhd/bindings/usrp_block_python.cc:16:
/cache/workdir/Ubuntu_20_10_64_0/build_Ubuntu_20_10_64/src/gr-uhd/python/uhd/bindings/../../../include/gnuradio/uhd/usrp_block.h:162:18: note: candidates are: ‘virtual void gr::uhd::usrp_block::set_gain(double, const string&, size_t)’
  162 |     virtual void set_gain(double gain, const std::string& name, size_t chan = 0) = 0;
      |                  ^~~~~~~~
/cache/workdir/Ubuntu_20_10_64_0/build_Ubuntu_20_10_64/src/gr-uhd/python/uhd/bindings/../../../include/gnuradio/uhd/usrp_block.h:153:5: note:                 ‘virtual void gr::uhd::usrp_block::set_gain(double, size_t, pmt::pmt_t)’
  153 |     set_gain(double gain, size_t chan = 0, pmt::pmt_t direction = pmt::PMT_NIL) = 0;
      |     ^~~~~~~~
make[2]: *** [gr-uhd/python/uhd/bindings/CMakeFiles/uhd_python.dir/build.make:89: gr-uhd/python/uhd/bindings/CMakeFiles/uhd_python.dir/usrp_block_python.cc.o] Error 1
make[2]: Leaving directory '/cache/workdir/Ubuntu_20_10_64_0/build_Ubuntu_20_10_64/build'
make[1]: *** [CMakeFiles/Makefile2:7683: gr-uhd/python/uhd/bindings/CMakeFiles/uhd_python.dir/all] Error 2
```

jacobagilbert:
I get the same error @marcusmueller does in the binding compilation.

mbr0wn:
@jacobagilbert I didn't write this; I reviewed it and it seems OK but if you have comments on the bigger picture, that would be good to know. Maybe you can chat with @segdy about it.

mbr0wn:
Build failures resolved.

jacobagilbert:
After looking at this, I think the difference is a product of how tuning is handled differently from other simple commands (like set_freq) which call the public setters. I think consistency is fine.

@segdy can you explain how one could use this to set both direction gains at the same time? I have tested that this works to set gain in either direction publishing the message to either block, but can you describe how this can be used to set both TX and RX gains or frequency with a single message?

segdy:
@mbr0wn @marcusmueller @jacobagilbert First of all, thank you very much for pushing this forward and fixing those bugs.
When I used/modified @ptrkrysik 's branch, I used gnuradio 3.8 (not master).
I have checked out `segdy-bidir_ctrl_of_freq_and_gain` and it builds nicely for me.
I do not have access to USRP hardware right now but I will test this tomorrow and check the patch. Maybe there need to be modifcations for #4091.

@jacobagilbert I think the important point of this patch is that it works for **tags** (not just messages). Tags can only be supplied to USRP Sink. When someone uses messages, the patch would not strictly be requires but still works.

This patch is not only useful for implementing a phase coherent transmitter+receiver, it is even strictly **required**. There is no hope to get sample accurate timing with messages (except it happens super slowly like hundres of ms or s range). There is no other way (using gr-uhd) to make a phase-coherent TX/RX.

So, long story short, this is an example in C++ of tuning RX/TX precicely, resulting in phase coherent operation:

```
tag_t tune_tag_tx;
tune_tag_tx.offset = n;
tune_tag_tx.key = pmt::string_to_symbol("tx_command");
tune_tag_tx.value = pmt::dict_add(tune_tag_tx.value, pmt::mp("chan"), pmt::from_long(0));

// NOTE: This requires my other patch for mtune which is already merged (#4091)
pmt::pmt_t tune_request_tx = pmt::make_dict();
tune_request_tx = pmt::dict_add(tune_request_tx, pmt::mp("rf_freq_policy"), pmt::mp("A"));
tune_request_tx = pmt::dict_add(tune_request_tx, pmt::mp("args"), pmt::mp("mode_n=integer,int_n_step=1000e3"));
tune_request_tx = pmt::dict_add(tune_request_tx, pmt::mp("dsp_freq_policy"), pmt::mp("A"));

// For the RX tune request we use double the frequency (as an example) 
pmt::pmt_t tune_request_rx = pmt::dict_add(tune_request_tx, pmt::mp("target_freq"), pmt::from_double(2*freq));
tune_request_tx = pmt::dict_add(tune_request_tx, pmt::mp("target_freq"), pmt::from_double(freq));

tag_t tune_tag_rx = tune_tag_tx;
tune_tag_rx.offset += nrxofs;

tune_tag_tx.value = pmt::dict_add(tune_tag_tx.value, pmt::mp("time"), time_to_pmt_time((n)/_samp_rate));
tune_tag_rx.value = pmt::dict_add(tune_tag_rx.value, pmt::mp("time"), time_to_pmt_time((n+nrxofs)/_samp_rate));

tune_tag_tx.value = pmt::dict_add(tune_tag_tx.value, pmt::mp("mtune"), tune_request_tx);
tune_tag_rx.value = pmt::dict_add(tune_tag_rx.value, pmt::mp("mtune"), tune_request_rx);

tune_tag_tx.value = pmt::dict_add(tune_tag_tx.value, pmt::mp("direction"), pmt::mp("TX"));
tune_tag_rx.value = pmt::dict_add(tune_tag_rx.value, pmt::mp("direction"), pmt::mp("RX"));

this->add_item_tag(0, tune_tag_tx);
this->add_item_tag(0, tune_tag_rx);
```

Note this example code uses `nrxofs` to slightly offset the message of the RX tuning. However, I am sure this was working with `nrxofs=0`. I can confirm this as soon as I get a hold of the USRP. This patch would be required regardless of `nrxofs`.

segdy:
As another remark: The main application is controlling **RX via TX tags**. If people don't like this interface, I could imagine still implementing this patch as-is under the hood but exposing the API only in `usrp_sink_impl.cc`. Then we would leave the handling for `usrp_source_impl.cc` as is. For  `usrp_sink_impl.cc`, `direction` always defaults to `TX`, except an explicit `direction=RX` is given.

I personally love the interface as it is and find it useful, that's just an alternative suggestion.

jacobagilbert:
@segdy makes perfect sense. While reviewing I have tested that this allows gain and frequency control in either direction via message to either source or sink block, just not tags. I will give that an once-over when I have a USRP available.

mbr0wn:
Fixed formatting.

jacobagilbert (approved):
I have confirmed both message and tag based command handling is correctly executed based on the 'direction' key. For instantaneous command execution, a 'time' key is required but this is inherent to how UHD processes commands under the hood.

Looks good to me, thanks @segdy

PR #4095, opened 2021-01-15
gr-fft: Log Power FFT Python fix

willcode:
Block would fail as soon as it was placed in GRC.
A couple of imports/renames were needed.


mbr0wn:
Should this go onto 3.9?

drmpeg:
Yes. Block doesn't work at all without this patch.

drmpeg (approved):
Tested and works good.

PR #3721, opened 2020-08-22
3.8: Use lambdas instead of boost::bind

marcusmueller:
This is mostly a backport of master functionality.

Fixes #3720

Discussion there.

PR #4103, opened 2021-01-16
msg_handler: Use lambdas in gr-uhd

mbr0wn:
Supersedes https://github.com/gnuradio/gnuradio/pull/3721

### DIFF
--- a/gr-uhd/lib/usrp_block_impl.cc
+++ b/gr-uhd/lib/usrp_block_impl.cc
@@ -21,7 +21,6 @@
  */
 
 #include "usrp_block_impl.h"
-#include <boost/make_shared.hpp>
 
 using namespace gr::uhd;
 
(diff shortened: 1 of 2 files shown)


gr-uhd/lib/usrp_block_impl.cc
```diff
@@ -163,6 +162,7 @@ usrp_block_impl::usrp_block_impl(const ::uhd::device_addr_t& device_addr,
     REGISTER_CMD_HANDLER(cmd_bandwidth_key(), _cmd_handler_bw);
     REGISTER_CMD_HANDLER(cmd_antenna_key(), _cmd_handler_antenna);
     REGISTER_CMD_HANDLER(cmd_gpio_key(), _cmd_handler_gpio);
+#undef REGISTER_CMD_HANDLER
```
marcusmueller, line 165: Feels like a separate commit ;)

marcusmueller (approved):
Aside from the slightly spurious (but very right `#undef`) very fine and good to go

gr-uhd/lib/usrp_block_impl.cc
mbr0wn: Well, it's your commit :)

PR #4104, opened 2021-01-16
Remove MSVC shim header cruft, reduce CMake overhead

marcusmueller:
closes #1680, closes #1674

PR #4105, opened 2021-01-16
MSVC/CMake Cruft: 39-backport of #4104

marcusmueller:


gr-blocks/lib/udp_sink_impl.cc
```diff
@@ -101,13 +101,16 @@ int udp_sink_impl::work(int noutput_items,
                         gr_vector_void_star& output_items)
 {
     const char* in = (const char*)input_items[0];
-    ssize_t r = 0, bytes_sent = 0, bytes_to_send = 0;
-    ssize_t total_size = noutput_items * d_itemsize;
+    size_t r = 0;
+    std::ptrdiff_t bytes_sent = 0, bytes_to_send = 0;
+    size_t total_size = noutput_items * d_itemsize;
```
mbr0wn, line 106:
```suggestion
    const size_t total_size = noutput_items * d_itemsize;
```
...but I'm not a stickler for it.

gr-blocks/lib/ConfigChecks.cmake
```diff
@@ -1,8 +1,8 @@
-# Copyright 2012 Free Software Foundation, Inc.
+#Copyright 2012 Free Software Foundation, Inc.
 #
-# This file is part of GNU Radio
+#This file is part of GNU Radio
 #
-# SPDX-License-Identifier: GPL-3.0-or-later
+#SPDX - License - Identifier : GPL - 3.0 - or -later
```
mbr0wn, line 5: Were these changes intentional?

marcusmueller: No, that was clang-format it seems which I let run on the CMake file to increase readability of the C++ code snippets

mbr0wn: I'm not a fan of these changes, but if we want them, maybe they should go onto their own commit. The comments look weird now, and I don't think the snippets required increased readability.

mbr0wn (approved):
I don't think I would have touched the cmake files with clang-format, and I would really prefer you revert the headers. Otherwise, this is a good change.

marcusmueller:
Not quite sure what to do about this?

marcusmueller:
Closing, let's first work out the master aspect of this.

PR #4107, opened 2021-01-17
audio: reduce Windows audio priority to LOW, this prefers Portaudio on win

marcusmueller:


mbr0wn:
How wide-spread is PortAudio on Windows? Is it really an acceptable alternative? This PR is still fine, but I'm asking in general since you discussed moving away from native Windows audio entirely.

marcusmueller:
Portaudio is fairly easy to get / build on Windows – not quite sure what you mean with with "widespread", since on windows, theres no real packaging of development libraries done by microsoft.

Portaudio is meant to allow projects to avoid exactly what we do: have an OS-specific driver for sound systems without making any use of special features, so windows is a very "native" platform for it.

ryanvolz:
I can't speak to how frequently portaudio is used on Windows in general, but I think using it with current GR Windows packaging efforts should be relatively easy (both vcpkg [Geof's binaries] and conda have portaudio available).

Does the Windows sink also need to be changed to low priority?

gnieboer:
I agree with the downgrade to below portaudio... if it's available, use that over (both) the "native" source and sink.  Part of the reason I say that is that those (native) blocks were one of my first projects when working on windows compatibility (2016), and it is not an area that we spend a lot of time maintaining... I got it to basically work and moved on.  Portaudio, on the other hand, is actively maintained by people with experience in the multimedia subsystems, so it seem likely to be better code in the long run.

But keep the native sinks as a backup seems prudent.  I don't know if @guruofquality packages portaudio in his distributions.

guruofquality:
> But keep the native sinks as a backup seems prudent. I don't know if @guruofquality packages portaudio in his distributions.

I do package portaudio FWIW, thanks for keeping me in the loop

mbr0wn:
Looks like the Windows experts agree to deprioritize. Let's merge.

PR #4092, opened 2021-01-14
qtgui:Added Time Raster Sink Control Axis Configuration

willcode:
This update allows a time raster sink's axis minimum, maximum,
and label to now be specified in the block configuration options.
This allows the block to be formatted to match a waterfall plot
in terms of having frequency on the lower axis.

Based on PR from ghostop14 <ghostop14@gmail.com>

Signed-off-by: Jeff Long <willcode4@gmail.com>

PR #4110, opened 2021-01-17
qtgui:Added Time Raster Sink Control Axis Configuration (backport 3.9)

willcode:
Backport https://github.com/gnuradio/gnuradio/pull/4092

### DIFF
--- a/gr-qtgui/grc/qtgui_time_raster_x.block.yml
+++ b/gr-qtgui/grc/qtgui_time_raster_x.block.yml
@@ -31,6 +31,36 @@ parameters:
     label: Num. Cols
     dtype: int
     default: 256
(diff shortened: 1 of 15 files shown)


jacobagilbert (approved):
Looks good to me. Builds/tests/works on ubuntu 20.04 same as #4092.

There are a number of test failures but they seem related to CI issues, not this PR.

mormj (approved):

mbr0wn:
I'm unsure if this this should get backported (it's a feature rather than a bugfix). That said, 3.9 is probably going to be around for a while.

mormj:
I would vote for backporting this early in the 3.9 release

PR #4118, opened 2021-01-18
Pybind: Adding header hash hotfixer

marcusmueller:
DO NOT USE THIS TOOL IF ADDING OR REMOVING ANY FUNCTIONALITY FROM A
HEADER.

This tool allows to quickly update the hash in a binding file
(*/python/*/binding/something.cc).

WARNING: This script updates header hashes without checking whether the
underlying functionality changed. If you added or removed any python-bound
function in the header to this binding file THIS IS NOT THE TOOL YOU SHOULD BE
USING. Use bindtool instead to generate correct binding code.

### DIFF
--- a/gr-utils/bindtool/scripts/binding-hash-hotfixer.zsh
+++ b/gr-utils/bindtool/scripts/binding-hash-hotfixer.zsh
@@ -1,0 +1,52 @@
+#!/usr/bin/zsh
+verbose=false
+if [[ "$1" == "-v" || "$1" == "--verbose" ]]; then
+    verbose=true
+    shift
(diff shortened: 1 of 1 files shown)


willcode:
Should there be a `Are you sure (y/N)` and a `-y` to force?

marcusmueller:
@willcode I think whoever got that far basically would be resistant to these anyway, so nah :) I already feel stupid for putting this many warnings into people's way.

mormj (approved):
Tested and it works!  Thanks for putting this one in-tree - certainly beats the manual method.

PR #4120, opened 2021-01-19
gr-qtgui: Fix deprecated use of QPixmap::grabWidget()

ZackerySpytz:
Use QWidget::grab() instead (which was introduced in Qt 5).


mbr0wn:
Reference: https://doc.qt.io/qt-5/qpixmap-obsolete.html

CI failures unrelated.

PR #4121, opened 2021-01-19
gnuradio-runtime: missing <algorithm> header for std:sort

gnieboer:
MSVC is looking for <algorithm> header to have sort be part of the std namespace

### DIFF
--- a/gnuradio-runtime/include/gnuradio/tag_checker.h
+++ b/gnuradio-runtime/include/gnuradio/tag_checker.h
@@ -12,6 +12,7 @@
 #define INCLUDED_GR_RUNTIME_TAG_CHECKER_H
 
 #include <gnuradio/tags.h>
+#include <algorithm>
 #include <vector>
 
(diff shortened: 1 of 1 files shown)


mbr0wn:
@gnieboer These are all good changes, thanks. After adding includes, you need to re-run clang-format (include order is not arbitrary). You can integrate clang-format into your editor, this isn't something you should spend brain-cycles on. When you fix the formatting, the CI builds will re-run.
Sorry for the nitpicks -- but we have to keep our codebase clean, even with the little stuff.

mbr0wn (approved):

gnieboer:
clang'ed

mbr0wn:
This is failing because the Python bindings weren't updated. However, I was also unsure where I had see this before -- turns out, we fixed this on master, and didn't backport. I included it in https://github.com/gnuradio/gnuradio/pull/4142, which will supersede this.

PR #4127, opened 2021-01-19
runtime: lib: Don't use hard-coded absolute path constants.

ryanvolz:
This replaces the path constants for the prefix, system configuration, and preferences directories with a runtime prefix lookup based on the gnuradio-runtime library location and relative paths from that. Boost >= 1.61 is required, but this is now compatible with the minimum version required by GNU Radio overall.

I'm still using `boost::filesystem` here since that's currently what is used elsewhere, but with C++17 there is no reason that this and elsewhere couldn't be `std::filesystem` with the `<filesystem>` header.

Also use cmake-style paths instead of native paths for substitution to avoid unescaped character problems. Windows is able to use the forward-slash paths just fine, and this also has external benefits (e.g. conda will do path replacement during installation and it substitutes forward-slash paths, so this avoids mixed-slash paths).

I also ran clang-format on `constants.cc.in` and that makes the diff bigger since (being a .in file) it evidently had not been clang-formatted prior to this.

Side-note: I've been using a version of this patch for the entirety of GNU Radio's life as a conda package. Since conda installs to a different prefix than the one it is compiled for, something like this has always been necessary. With the bump in Boost version and others running into similar issues (#4123), it was clear to me that another attempt at upstreaming was necessary.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -279,12 +279,20 @@ endif(ENABLE_PERFORMANCE_COUNTERS)
 ########################################################################
 # Variables replaced when configuring the package config files
 ########################################################################
-file(TO_NATIVE_PATH "${CMAKE_INSTALL_PREFIX}"           prefix)
-file(TO_NATIVE_PATH "\${prefix}"                        exec_prefix)
-file(TO_NATIVE_PATH "\${exec_prefix}/${GR_LIBRARY_DIR}" libdir)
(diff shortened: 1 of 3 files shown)


mbr0wn:
I'd like to separate the clang-format (https://github.com/gnuradio/gnuradio/pull/4129). Our CI should be checking that file -- we missed it because we used a *.cc glob.

ryanvolz:
I rebased on top of @mbr0wn's `.cc.in` clang-format branch.

PR #4128, opened 2021-01-19
digital: change ofdm_chanest qa test to use head

### LINKED ISSUE #4102: Fix qa_ofdm_chanest_vcvc
That unit test likes to fail in a lot of places, and curiously, it started happening to me after checking in a change to `add_block`. Barring strange quantum stuff, there's something fishy going on with FPU state, maybe?

mormj:
Fixes #4102

### DIFF
--- a/gr-digital/python/digital/qa_ofdm_chanest_vcvc.py
+++ b/gr-digital/python/digital/qa_ofdm_chanest_vcvc.py
@@ -303,14 +303,12 @@ class qa_ofdm_chanest_vcvc (gr_unittest.TestCase):
                 for x in range(fft_len)]
             src = blocks.vector_source_c(tx_data, False, fft_len)
             chan = blocks.multiply_const_vcc(channel)
-            noise = analog.noise_source_c(analog.GR_GAUSSIAN, wgn_amplitude)
+            noise = blocks.vector_source_c(numpy.random.normal(0,wgn_amplitude,(len(tx_data),)), False, fft_len)
             add = blocks.add_cc(fft_len)
             chanest = digital.ofdm_chanest_vcvc(sync_sym1, sync_sym2, 1)
             sink = blocks.vector_sink_c(fft_len)
             top_block.connect(src, chan, (add, 0), chanest, sink)
-            top_block.connect(
-                noise,
-                blocks.stream_to_vector(gr.sizeof_gr_complex, fft_len), (add, 1))
+            top_block.connect(noise, (add, 1))
             top_block.run()
             channel_est = None
             carr_offset_hat = 0


drmpeg (approved):
Works good here.

gr-digital/python/digital/qa_ofdm_chanest_vcvc.py
```diff
@@ -301,13 +301,14 @@ def run_flow_graph(sync_sym1, sync_sym2, data_sym):
...
-            src = blocks.vector_source_c(tx_data, False, fft_len)
+            src = blocks.vector_source_c(tx_data, True, fft_len)
             chan = blocks.multiply_const_vcc(channel)
             noise = analog.noise_source_c(analog.GR_GAUSSIAN, wgn_amplitude)
             add = blocks.add_cc(fft_len)
             chanest = digital.ofdm_chanest_vcvc(sync_sym1, sync_sym2, 1)
             sink = blocks.vector_sink_c(fft_len)
-            top_block.connect(src, chan, (add, 0), chanest, sink)
+            head = blocks.head(gr.sizeof_gr_complex * fft_len, 1)
+            top_block.connect(src, chan, (add, 0), chanest, head, sink)
```
jdemel, line 311: So you want to add a `head` block to make sure the flowgraph finishes. The issue would be that the `noise` block, in parallel to `src` does not have a "generate a finite amount of samples" option. The scheduler might hang on that. Thus, the qa test doesn't pass.
tbh, I'd prefer a solution that doesn't require another block that just forces the flowgraph to stop. Would it be possible to use a `fast_noise_source` instead of the current noise source? The `fast_noise_source` block should be a GRC only block that basically wraps `blocks.vector_source_c(np.random.normal(...), [True/False], fft_len)`.

mormj: It also makes the vector source repeat to keep pushing samples, and using the head instead of the vector source to signal done acts to make sure there are enough samples at the end of flowgraph before it is killed.

jdemel: Yeah. And I suggest to take a different approach here because it would be good to have a test with exact input `x` to produce exact output `y`. If that's not the case, this might be a good indicator that something is broken. I know that `np.random....` or some other random function interferes with that.

mormj: Gotcha - the general flow works if I change the noise_source to a vector source with np.random.normal vector as its values.
I think the underlying issue is the single value items for the noise source upstream from the add block requires more samples to be pushed through the pipeline before the vector source with 1 vector says it is done.

mbr0wn (approved):
Change is fine; but I need to learn more about why this fixes anything.

mbr0wn:
@mormj Does this need backporting? Tests don't fail on 3.9 so I'm not sure.

drmpeg:
Doesn't need backporting. This PR is to fix the failure associated with commit https://github.com/gnuradio/gnuradio/commit/c8b16186cda5b195e8df6db7e3aa8715127478b7 (which wasn't backported to 3.9).

PR #4136, opened 2021-01-22
gr-digital: fixes name error in chunks_to_symbols cpp_templates logic

seeAppRun:
Fixes a name error in the cpp_templates logic that results in invalid C++ code being generated.


mbr0wn:
I had to look this up: float_vector is float, and real_vector is double. We never use double types as block outputs, so here we go.

mbr0wn:
Superseded by https://github.com/gnuradio/gnuradio/pull/4143

PR #4138, opened 2021-01-23
MSVC 2019 fixes

guruofquality:
I have built gnuradio under MSVC 2019 with python3.9 and I now have GRC up and running. The following fixes were required to both compile and run. Please let me know if I need to adjust any of the commits. The patches are split into multiple functional commits so they can be easily cherry picked. I am interested in building patch free in the future.

Screen shot here: gnuradio-companion launcher working as well as launching generated qtgui apps from GRC itself. Note that gr-osmosdr is still missing, and does need some help to get python bindings ported to 3.9

grc_msvc2019

And further, because I was unable to simply pip install pygtk for GRC, I created a installable wheel file for Python3.9 AMD64, should that be of interest to anyone: https://downloads.myriadrf.org/binaries/python39_amd64/PothosSDRPyGTK-2021.1.21-cp39-cp39-win_amd64.whl

# Changes in the pull request

## runtime: replace use of std::unary_function

std::unary_function is deprecated of as c++11 and removed in c++17
https://en.cppreference.com/w/cpp/utility/functional/unary_function
The solution was to replace unary_function with std::function.

## qtgui: replace use of `or` macro with `||` 

This macro is not defined without including ciso646
https://www.cplusplus.com/reference/ciso646/
gnuradio otherwise does not use these macros

## runtime: replace /dev/stdin in sine table generator

/dev/stdin does not exist and was replaced by directing stdout to the
file rather than piping stdout to the stdin of an install command

## python: add the dll runtime path in __init__.py

python3.8 and up needs to have the dll search path set explicitly before importing.
More details can be found here: https://docs.python.org/3/whatsnew/3.8.html#bpo-36085-whatsnew

Without the fix, python libraries will fail to import regardless of how PATH is configured.
The solution searches for the runtime path for gnuradio dlls and calls add_dll_directory() before imports.
As a precedent, a similar behaviour is implemented in the gi/__init__.py module used with pygtk.

### DIFF
--- a/gnuradio-runtime/include/gnuradio/rpcserver_aggregator.h
+++ b/gnuradio-runtime/include/gnuradio/rpcserver_aggregator.h
@@ -13,6 +13,7 @@
 
 #include <gnuradio/rpcmanager_base.h>
 #include <gnuradio/rpcserver_base.h>
+#include <functional>
 #include <string>
 #include <vector>
 
@@ -42,7 +43,7 @@ public:
 
 private:
     template <class T, typename Tcallback>
-    struct registerConfigureCallback_f : public std::unary_function<T, void> {
+    struct registerConfigureCallback_f : public std::function<void(T)> {
         registerConfigureCallback_f(const std::string& _id, const Tcallback _callback)
             : id(_id), callback(_callback)
         {
@@ -55,7 +56,7 @@ private:
     };
 
     template <class T, typename Tcallback>
-    struct unregisterConfigureCallback_f : public std::unary_function<T, void> {
+    struct unregisterConfigureCallback_f : public std::function<void(T)> {
         unregisterConfigureCallback_f(const std::string& _id) : id(_id) { ; }
 
         void operator()(T& x) { x->i()->unregisterConfigureCallback(id); }
@@ -63,7 +64,7 @@ private:
     };
 
     template <class T, typename Tcallback>
-    struct registerQueryCallback_f : public std::unary_function<T, void> {
+    struct registerQueryCallback_f : public std::function<void(T)> {
         registerQueryCallback_f(const std::string& _id, const Tcallback _callback)
             : id(_id), callback(_callback)
         {
@@ -76,7 +77,7 @@ private:
     };
 
     template <class T, typename Tcallback>
-    struct unregisterQueryCallback_f : public std::unary_function<T, void> {
+    struct unregisterQueryCallback_f : public std::function<void(T)> {
         unregisterQueryCallback_f(const std::string& _id) : id(_id) { ; }
 
         void operator()(T& x) { x->i()->unregisterQueryCallback(id); }
@@ -85,7 +86,7 @@ private:
 
 
     template <class T, typename Tcallback>
-    struct registerHandlerCallback_f : public std::unary_function<T, void> {
+    struct registerHandlerCallback_f : public std::function<void(T)> {
         registerHandlerCallback_f(const std::string& _id, const Tcallback _callback)
             : id(_id), callback(_callback)
         {
@@ -98,7 +99,7 @@ private:
     };
 
     template <class T, typename Tcallback>
-    struct unregisterHandlerCallback_f : public std::unary_function<T, void> {
+    struct unregisterHandlerCallback_f : public std::function<void(T)> {
         unregisterHandlerCallback_f(const std::string& _id) : id(_id) { ; }
 
         void operator()(T& x) { x->i()->unregisterHandlerCallback(id); }
--- a/gnuradio-runtime/lib/CMakeLists.txt
+++ b/gnuradio-runtime/lib/CMakeLists.txt
@@ -43,10 +43,11 @@ add_subdirectory(pmt)
 #Generate sine table header file for Math part of library
 set(MATH_SINE_TABLE_HEADER ${CMAKE_CURRENT_BINARY_DIR}/math/generated/sine_table.h)
 set(MATH_SINE_TABLE_GENERATOR ${CMAKE_CURRENT_SOURCE_DIR}/math/gen_sine_table.py)
+execute_process(COMMAND ${CMAKE_COMMAND} -E make_directory "${CMAKE_CURRENT_BINARY_DIR}/math/generated")
 add_custom_command(
   DEPENDS ${MATH_SINE_TABLE_GENERATOR}
   OUTPUT ${MATH_SINE_TABLE_HEADER}
-  COMMAND ${PYTHON_EXECUTABLE} ${MATH_SINE_TABLE_GENERATOR} | install -D /dev/stdin ${MATH_SINE_TABLE_HEADER}
+  COMMAND ${PYTHON_EXECUTABLE} ${MATH_SINE_TABLE_GENERATOR} > ${MATH_SINE_TABLE_HEADER}
 )
 
 add_library(gnuradio-runtime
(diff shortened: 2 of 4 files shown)


gnieboer:
For the python 3.8+ path issue, another option is to add a gnuradio.pth file that adds a fixed path to the search path.  Either approach I think is fine, the search / add_dll_directory() is appropriate for more flexible installations where the location of the DLL's is variable but on PATH, but the .pth option is more robust in a case where all the files will be in a known directory, as it avoids dependencies on the common PATH variable that could get confused by multiple installations.

For gr-osmosdr there is a forked repo out there that has enough fixes in place to function with 3.9, BTW, if you want to just get it going.

guruofquality:
> For the python 3.8+ path issue, another option is to add a gnuradio.pth file that adds a fixed path to the search path. Either approach I think is fine, the search / add_dll_directory() is appropriate for more flexible installations where the location of the DLL's is variable but on PATH, but the .pth option is more robust in a case where all the files will be in a known directory, as it avoids dependencies on the common PATH variable that could get confused by multiple installations.

So just to clarify, ideally my use case does not modify or use environment variables like PATH or PYTHONPATH. I don't change filesystem layout either, just transport it to different root paths based on the path selected by the user in the installer.

## Python module search path

So for the python search path, the installer writes en entry in `SOFTWARE\Python\PythonCore\3.X\PythonPath` registry key to get the project (gnuradio modules and many others) into the search path for the python executable. So I think that would obviate the need for the .pth file. But im not 100% clear on .pth, what else it does and how it works with the dll search path, so I might be missing something else I could be doing here.

## DLL search PATH

So actually setting the PATH environment variable is an option on the installer although if you don't everything is fine. The launchers all start in the "install_root\bin" which is how they can find the necessary runtime DLLs from the package. And then changes like the one in question allow python apps (or grc generated apps) to work pretty seamlessly without setting the PATH either.

## How gnuradio can find the bin PATH

So I agree that even using add_dll_directory, this change could be more robust. But to clarify I'm not searching the PATH, I'm just looking up from site-packages a few directories to find `bin`, so it should be a search relative within the directory structure of the installation. I did this because the python site-packages install path can be arbitrary with `-DGR_PYTHON_DIR` and I wanted to support those situations where the directory structure may have different depths for python modules.

If I wanted to be more robust, I would have cmake's `file(RELATIVE_PATH` calculate the relative path between GR_PYTHON_DIR and bin and shove it into the __init__.py. However this would change the __init__.py into a configure_file() situation, and that would have been too invasive of a change in my opinion to get it upstreamed. :-)

> For gr-osmosdr there is a forked repo out there that has enough fixes in place to function with 3.9, BTW, if you want to just get it going.

Thanks for the info. I patched it enough just to get the c++ side of things building for GQRX, but that fork might be nice. Do you have a URL for that fork?

gnieboer:
Ah, I see, you are searching the python dir descendents for bin, not PATH, that makes sense.  PyQt5 has some code in it that searches all PATH for Qt DLL's and adds the first one it finds.  The only thing I see here is that it can't go "up and over" to find the files if the python file is lower in the tree than the DLLs, right?

My gnuradio.pth is single line of text:
import os; bindir = os.path.normpath(os.path.join(__file__,'..\\..\\..\\..\\bin')); os.add_dll_directory(bindir); 

In my installer that's now the relative path to the consolidated DLL/EXE directory, and adding that file to lib/site-packages at this point seems to be enough.   I don't need to do much error checking since python version etc are known embedded quantities.  

The repo I got the gr-osmosdr code from is this fork: https://github.com/argilo/gr-osmosdr  (gr3.9 branch).  It may have taken a bit of fiddling (line endings to LF manually at a minimum depending on your git-global settings).

guruofquality:
> Ah, I see, you are searching the python dir descendents for bin, not PATH, that makes sense. PyQt5 has some code in it that searches all PATH for Qt DLL's and adds the first one it finds.

Funny enough pygtk does exactly the same thing. So at least for the pygtk wheel I'm providing, I know where the runtime is located, I can patch its `gi/__init__.py` to do that more robustly -- on the next build.

> The only thing I see here is that it can't go "up and over" to find the files if the python file is lower in the tree than the DLLs, right?

I think its robust if your configuration and install maintains the file system hierarchy, where all of the files live in the same root containing bin/. 

However the fix would break down in cases where the python libs are transplanted after install in a non-relative way out of the relative root --OR-- if the GR_PYTHON_DIR was perhaps an absolute path. I think in this case, the search would just fail to find bin and continue, or I guess find some other bin where the python libs were transplanted too.

> My gnuradio.pth is single line of text:
import os; bindir = os.path.normpath(os.path.join(file,'..\..\..\..\bin')); os.add_dll_directory(bindir);
In my installer that's now the relative path to the consolidated DLL/EXE directory, and adding that file to lib/site-packages at this point seems to be enough. I don't need to do much error checking since python version etc are known embedded quantities.

What python version are you testing with? I like this if it works, but I tried making this gnuradio.pth in both the gnuradio site-packages and also in a default python site-packages (with an absolute path). And I couldn't get it work work.

As far as I can tell .pth files don't allow python code, they just take a list of search directories separated by newline, and must be in the default search path for the site (whatever site.getsitepackages() returns).

## A few proposed solutions

I do want to be able to fix these other issues and close out the pull. And I don't want to disrupt anything you are trying to maintain. I guess it would be kind of cool if the default build sort of solved this, but I'm cool either way:

1) I can drop the changes entirely and just prepend `import os; bindir = os.path.normpath(os.path.join(file,'..\..\..\..\bin')); os.add_dll_directory(bindir);` manually onto the start of `__init__.py` post install. I could still build patch free, and that would be fairly simple, just as simple as creating a .pth, if I can figure that out :-)

2) Actually build the known relative path between GR_PYTHON_DIR and bin into `__init__.py` (at cmake configure time). `__init__.py` would always have to resolve this path relative to `__file__` and add it if it exists. This would be pretty robust, but wouldnt work non-relativistic install transplants.

3) Dump a file (just like the .pth) next to `__init__.py` that actually tells gnuradio's `__init__.py` where to add the DLL path. `__init__.py` would look for this file and use its path if present. The cmake build could provide this file by default with the relative path known at configure time. And someone transplanting things in a weird way could always replace/override the file if thats needed.

3.a) Same thing but have cmake generate the gnuradio.pth to install if it works like that

4) Use an environment variable like GR_ROOT. If `__init__.py` saw that GR_ROOT was set and GR_ROOT/bin was found, it would use that as the dll search path. I like that a little less, but its easy enough to set env vars from a launcher.

Anyway, sorry for the thought dump, but let me know if any of that floats your boat.

## CI says bb/Fedora_33_master is failing

./gr-blocks/python/blocks/qa_cpp_py_binding_test.sh is failing. I don't think this is me. It seems to be failing at the branch point. I did spin up a fedora 33 vm to check this :-(

mbr0wn:
Yeah I think we can ignore those CI failures. The GH-Actions based F33 is working.

mbr0wn:
I don't know much about the paths and Windows, but the rest is fine. If this thread goes on, I'll merge those other commits onto master before we conclude the rest of the discussion. @guruofquality thanks a lot, and the commits are set up perfectly.

ryanvolz (approved):
These changes look good and should be compatible with what I'm doing with conda.

(I had to look up the add_dll_directory stuff. As it turns out, conda-forge's Python has decided to patch out that change and that's why I haven't run into that issue!)

mbr0wn:
Superseded by #4159.

Issue #3409, opened 2020-04-24
Filter/FFT Windowing Cleanup


mrjacobagilbert:
Per the discussion in #3406 there are several improvements and cleanup that should happen within GR's windowing code:

1. Windowing has been formally moved from filter/firdes into fft/window however much of the code still uses filter::firdes. This seems like it needs some attention, at a minimum GR should not have two window type enums as it does now.
2. Many filter types are not included in the window type enum and are not exposed to the filter::firdes::window() or fft::window::build() functions. 
3. The window() and build() top-level window generators take an optional 'beta' parameter currently used for Kaiser windows. This should be changed to a generic optional parameter that can be used for any window that needs an argument (kaiser, exponential, tukey, gaussian, etc)
4. Add a guassian window
5. Tap lengths should be size_t or some sort of unsigned value
6. Windowing code needs const and the like for more effective compiler optimization consistent with the rest of GR.

mrjacobagilbert:
7. Explore including additional 'flat top' windows (e.g.: the Matlab implementation from  [D’Antona, Gabriele, and A. Ferrero. Digital Signal Processing for Measurement Systems. New York: Springer Media, 2006, pp. 70–72]). There are also several other appealing versions in [Heinzel, G.; Rüdiger, A.; Schilling, R. (2002). Spectrum and spectral density estimation by the Discrete Fourier transform (DFT), including a comprehensive list of window functions and some new flat-top windows (Technical report). Max Planck Institute (MPI) für Gravitationsphysik / Laser Interferometry & Gravitational Wave Astronomy].

PR #4139, opened 2021-01-23
gr-fft: Expand window enum, fix a few other minor things

jacobagilbert:
This addresses most of #3409 to allow use of all of the various windows that are included in GR now. Additionally bindings for the gaussian and Tukey windows, recent additions, have been added. Use of `beta` which is specific to the Kaiser filter has been replaced by `param` as a general parameter for the new kernels that also require a parameter for construction.

I think the C++ modernization aspect of the issue "Windowing code needs const and the like for more effective compiler optimization consistent with the rest of GR." would benefit from a look over by @ThomasHabets, please share any thoughts.

### DIFF
--- a/gr-fft/lib/window.cc
+++ b/gr-fft/lib/window.cc
@@ -50,7 +50,7 @@ double freq(int ntaps) { return 2.0 * GR_M_PI / ntaps; }
 
 double rate(int ntaps) { return 1.0 / (ntaps >> 1); }
 
-double window::max_attenuation(win_type type, double beta)
+double window::max_attenuation(win_type type, double param)
 {
     switch (type) {
     case (WIN_HAMMING):
@@ -66,7 +66,7 @@ double window::max_attenuation(win_type type, double beta)
         return 21;
         break;
     case (WIN_KAISER):
-        return (beta / 0.1102 + 8.7);
+        return (param / 0.1102 + 8.7);
         break;
     case (WIN_BLACKMAN_hARRIS):
         return 92;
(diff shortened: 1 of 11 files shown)


gr-fft/lib/window.cc
```diff
@@ -50,33 +50,51 @@ double freq(int ntaps) { return 2.0 * GR_M_PI / ntaps; }
...
+    case WIN_RIEMANN:
+        return 39;
+    case WIN_GAUSSIAN:
+        // value not meaningful for gaussian windows, but return something reasonable
+        return 100;
+    case WIN_TUKEY:
+        // low end is a rectangular window, attenuation exponentialy approaches Hann
+        // piecewise linear estimate, will be slightly greater than expected attenuation
+        if (param < 0.7)
+            return (param * 5 + 21);
```
ThomasHabets, line 93: Do these magic values have a semantic meaning? If so, it would be nice to have them be named constants.

gr-fft/lib/window.cc
```diff
@@ -50,33 +50,51 @@ double freq(int ntaps) { return 2.0 * GR_M_PI / ntaps; }
...
+        else if (param < 0.9)
+            return ((param - 0.6) * 20 + 24);
+        else
```
ThomasHabets, line 96: Needless "else".

ThomasHabets (approved):

gr-fft/lib/window.cc
jacobagilbert: All of that is a empirically derived piecewise linear approximation (orange) due to how the attenuation changes with alpha (blue). The value is not critical, but underestimating it can result in firdes returning a filter with too few coefficients for the desired attenuation. I'll clarify that.

image

ThomasHabets: Yup, worth a comment at least. 👍

gr-filter/python/filter/bindings/firdes_python.cc
```diff
@@ -49,7 +49,7 @@ void bind_firdes(py::module& m)
                     py::arg("cutoff_freq"),
                     py::arg("transition_width"),
                     py::arg("window") = ::gr::fft::window::win_type::WIN_HAMMING,
-                    py::arg("beta") = 6.7599999999999998,
+                    py::arg("param") = 6.7599999999999998,
```
mormj, line 52: Should the truncated values (which were automatically added at some point) be changed here to the intended 6.76?

mormj (approved):

gr-filter/python/filter/bindings/firdes_python.cc
jacobagilbert: I thought about doing that....but figured autogenerated code might get autogenerated again and it's pretty clear where the value came from.

PR #4140, opened 2021-01-23
Prevent Windows CRLF conversion for files subject to pybind hash comparisons

gnieboer:
In GNURadio 3.9, pybind11 compares hash values of headers and source files against knowns values.  Conversion of values to CRLF on checkout break those checks so keep LF values when files are subject to said checks.

I only saw code doing hashing for C source and header files.  Are there any other file types where this is used I didn't see?
Also, does these hash checks serve additional purposes after build that we need to be aware of?

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -7,3 +7,11 @@
 *.m4	-crlf
 *.ac	-crlf
 *.scm	-crlf
+#
+# In GNURadio 3.9, pybind11 compares hash values of headers and source
+# files against knowns values.  Conversion of values to CRLF on checkout
+# break those checks so keep LF values when files are subject to said checks
+#
+*.h    text eol=lf
+*.c    text eol=lf
+*.cc   text eol=lf


mbr0wn:
> I only saw code doing hashing for C source and header files. Are there any other file types where this is used I didn't see?
No, that's it. This is for matching the Pybind11 bindings with the original C++ source code.
> Also, does these hash checks serve additional purposes after build that we need to be aware of?
Not that I know of.

mormj (approved):

mbr0wn:
For people's reference: https://www.git-scm.com/docs/gitattributes

- `text` attribute: Setting the text attribute on a path enables end-of-line normalization and marks the path as a text file. End-of-line conversion takes place without guessing the content type.
- `eol` attribute: Set to string value "lf": This setting forces Git to normalize line endings to LF on checkin and prevents conversion to CRLF when the file is checked out.

[copy & pasted from reference above]

PR #4142, opened 2021-01-24
3.9 Backports: File Source, Qt Chooser, and more

mbr0wn:

PR #4143, opened 2021-01-24
digital: Fix output type in GRC bindings of chunks_to_symbols

mbr0wn:
Type was flagged as 'real_vector', which would mean vector<double>. We
want vector<float>, thus this fix.
This would cause an error when creating CPP code from GRC.

Signed-off-by: Terry May <terrydmay@gmail.com>

Same as https://github.com/gnuradio/gnuradio/pull/4136, but with fixed history.


mbr0wn:
FYI both @seeAppRun and I think this is OK. If CI passes, I think we can merge (if someone else sees this).

willcode:
Backported to maint-3.9 as part of https://github.com/gnuradio/gnuradio/pull/4142.

PR #4149, opened 2021-01-25
portaudio: fix unintentionally synchronous std::async

marcusmueller:
Note this classic (which I didn't do in ALSA back in the day, 'doh):

If you don't assign your future coming from std::async to something, its
destructor is instantly called, leading to the call effectively being
blocking.

### DIFF
--- a/gr-audio/lib/portaudio/portaudio_sink.cc
+++ b/gr-audio/lib/portaudio/portaudio_sink.cc
@@ -114,7 +114,7 @@ int portaudio_sink_callback(const void* inputBuffer,
     }
 
     else { // underrun
-        std::async(&portaudio_underrun_notification, self->d_logger);
+        auto future_local = std::async(&portaudio_underrun_notification, self->d_logger);
         // FIXME we should transfer what we've got and pad the rest
         memset(outputBuffer, 0, nreqd_samples * sizeof(sample_t));
 


gr-audio/lib/portaudio/portaudio_sink.cc
```diff
@@ -114,7 +114,7 @@ int portaudio_sink_callback(const void* inputBuffer,
     }
 
     else { // underrun
-        std::async(&portaudio_underrun_notification, self->d_logger);
+        auto future_local = std::async(&portaudio_underrun_notification, self->d_logger);
```
mbr0wn, line 117: Did you verify that this actually defers the async? Quote:

    If both the std::launch::async and std::launch::deferred flags are set in policy, it is up to the implementation whether to perform asynchronous execution or lazy evaluation. 

When I started playing with std::async, I noticed that most of the compilers went "oh well, in that case, let's defer". https://stackoverflow.com/questions/46102206/how-does-a-c-compiler-choose-between-deferred-and-async-execution-for-stdasy

In other words, I'm unsure if you meant to do this:
```suggestion
        auto future_local = std::async(std::launch::async, &portaudio_underrun_notification, self->d_logger);
```

Now maybe you did that intentionally, since 'deferred' is still different from 'do it now'.

mbr0wn: Well it's still an improvement so let's merge.

willcode:
3.8 does not use async at all. Removing Backport-3.8 tag.

willcode:
Already in maint-3.9, commit 4164b2c528.

PR #4152, opened 2021-01-25
Clean up usage of assert where an exception would've been wiser, sort out includes on the way

marcusmueller:


gr-blocks/lib/packed_to_unpacked_impl.cc
```diff
@@ -38,8 +38,12 @@ packed_to_unpacked_impl<T>::packed_to_unpacked_impl(unsigned int bits_per_chunk,
       d_endianness(endianness),
       d_index(0)
 {
-    assert(bits_per_chunk <= this->d_bits_per_type);
-    assert(bits_per_chunk > 0);
+    if (bits_per_chunk > d_bits_per_type) {
+        GR_LOG_ERROR(this->d_logger,
+                     boost::format("Requested to get %d out of a %d bit chunk") %
+                         bits_per_chunk % d_bits_per_type);
+        throw std::domain_error("can't have more bits in chunk than in output type");
```
mbr0wn, line 45: Is this the correct exception type? Never seen this used before.

gr-audio/lib/alsa/alsa_sink.cc
```diff
@@ -282,8 +283,6 @@ int alsa_sink::work(int noutput_items,
                     gr_vector_const_void_star& input_items,
                     gr_vector_void_star& output_items)
 {
-    assert((noutput_items % d_period_size) == 0);
```
mbr0wn, line 285: I think this change is OK, but in general, I'm OK with people asserting things that "should be like that". Who knows in which context code gets run in the future.

gr-audio/lib/oss/oss_source.cc
```diff
@@ -162,7 +161,8 @@ int oss_source::work(int noutput_items,
             break;
 
         default:
-            assert(0);
+            GR_LOG_ERROR(d_logger, "not mono nor stereo");
```
mbr0wn, line 164: If someone actually sees this message, they'll be like "huh?"

gr-filter/lib/filter_delay_fc_impl.cc
```diff
@@ -77,7 +78,8 @@ int filter_delay_fc_impl::work(int noutput_items,
         break;
 
     default:
-        assert(0);
+        GR_LOG_ERROR(d_logger, "not 1 or 2 input streams");
```
mbr0wn, line 81: Maybe we need a macro for log-error-and-throw.

mbr0wn (commented):
Looks like one of your commits has a too-long subject line.

marcusmueller:
Closing this PR; needs rework (imho, good ideas in here), no time.

will make issue to remind me.

PR #4171, opened 2021-01-30
Adding direction to mtune in gr-uhd

segdy:
This is an additional fix for https://github.com/gnuradio/gnuradio/pull/4093 and https://github.com/gnuradio/gnuradio/pull/4091.

The `mtune` command does not take the `direction` parameter into account. I think I have pushed this change previously into my old pull request but I think it has not found its way to the new one.

So this is a trivial one and all it does is it reads out `direction` for `mtune` command and supplies it to `_update_curr_tune_req`.

Thanks everyone for https://github.com/gnuradio/gnuradio/pull/4093 and https://github.com/gnuradio/gnuradio/pull/4091 and thanks for merging!

### DIFF
--- a/gr-uhd/lib/usrp_block_impl.cc
+++ b/gr-uhd/lib/usrp_block_impl.cc
@@ -847,6 +847,14 @@ void usrp_block_impl::_cmd_handler_mtune(const pmt::pmt_t& tune,
                                          int chan,
                                          const pmt::pmt_t& msg)
 {
+    // See if a direction was specified
+    pmt::pmt_t direction =
+        pmt::dict_ref(msg,
+                      cmd_direction_key(),
+                      pmt::PMT_NIL // Anything except "TX" or "RX will default to the
+                                   // messaged block direction"
+        );
+
     ::uhd::tune_request_t new_tune_request;
     if (pmt::dict_has_key(tune, pmt::mp("dsp_freq"))) {
         new_tune_request.dsp_freq =
@@ -887,7 +895,7 @@ void usrp_block_impl::_cmd_handler_mtune(const pmt::pmt_t& tune,
             pmt::symbol_to_string(pmt::dict_ref(tune, pmt::mp("args"), pmt::mp(""))));
     }
 
-    _update_curr_tune_req(new_tune_request, chan);
+    _update_curr_tune_req(new_tune_request, chan, direction);
 }
 
 void usrp_block_impl::_cmd_handler_bw(const pmt::pmt_t& bw,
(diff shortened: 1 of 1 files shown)


gr-uhd/lib/usrp_block_impl.cc
```diff
@@ -847,6 +847,14 @@ void usrp_block_impl::_cmd_handler_mtune(const pmt::pmt_t& tune,
                                          int chan,
                                          const pmt::pmt_t& msg)
 {
+    // See if a direction was specified
+    pmt::pmt_t direction =
+        pmt::dict_ref(msg,
+                      cmd_direction_key(),
+                      pmt::PMT_NIL // Anything except "TX" or "RX will default to the
+                                   // messaged block direction"
```
mbr0wn, line 855:
```suggestion
                      pmt::PMT_NIL // Anything except "TX" or "RX" will default to the
                                   // messaged block direction
```

mbr0wn (changes requested):
Can you please:
- Fix the minor typo in the comment
- Change your git subject line to something like `uhd: Add direction to mtune`
    - Note how the line is prefixed by the name of the in-tree module
    - Even better would be if you add some text to the message itself, explaining what you can now do that you couldn't before.
  
Thanks!

segdy:
> Can you please:
> 
>     * Fix the minor typo in the comment
> 
>     * Change your git subject line to something like `uhd: Add direction to mtune`
>       
>       * Note how the line is prefixed by the name of the in-tree module
>       * Even better would be if you add some text to the message itself, explaining what you can now do that you couldn't before.

Fixed and re-checked in. Hope this was OK.

mbr0wn:
@segdy Almost. I'm merging this by hand because there were two things still wrong:
- The subject line exceeded the 72 character limit, and
- The idea was that you squash the changes into a single commit (not add another one).

segdy:
> @segdy Almost. I'm merging this by hand because there were two things still wrong:

Thank you, will keep these in mind for next time.
To make sure:

>     * The subject line exceeded the 72 character limit, and

Multiple lines are permitted though, each line 72 chars max?

>     * The idea was that you squash the changes into a single commit (not add another one).

Ok.
How would I have done this the correct way? `git commit --amend` ?
Or `git rebase`? This one is discouraged for commits that I pushed already (https://www.atlassian.com/git/tutorials/rewriting-history)

Also: the changes in `master` (like this pull request) will be re-integrated into maint-3.9 and appear in the next released 3.9 version, is this correct? (but not in maint-3.8 or earlier).

mbr0wn:
Rewriting history on branches is OK if they're not merged yet. So, rebase -i is my preferred tool.

There's no limit to how many lines you can have -- it's like documentation. More is often better!

PR #4175, opened 2021-02-01
cppgen: Use correct required version in CMakeLists

haakov:
Looks like config.version changed at some point.

Fixes following error:
```
CMake Error at CMakeLists.txt:13 (find_package):
  find_package called with invalid argument "v3.10"
```

caused by
```
find_package(Gnuradio "v3.10" COMPONENTS
```


mbr0wn (approved):
Looks like this change was made elsewhere, but forgotten here.

willcode:
Backported to maint-3.9 as part of https://github.com/gnuradio/gnuradio/pull/4187.

Issue #3519, opened 2020-05-28
filter: Out of bound memory access in the fft_filter.cc filter() functions 


jacobgilbert-snl:
Illegal memory access in filter, similar to #3393. It is possible for out of bound memory access within the the fft_filter kernel; in the example below output memory addresses beyond the specified number of items are written to:

```c++
#include <iostream>
#include <vector>
#include <gnuradio/filter/fft_filter.h>

int main()
{
  int ntaps = 5;
  int nsamps = 2000;
  int decimation = 32;

  // generate taps
  std::vector<float> taps(ntaps,1);

  // generate samples
  std::vector<float> samples(nsamps,.5);

  // generate filter
  gr::filter::kernel::fft_filter_fff filt(32,taps,1);

  // output vector - make large enough
  std::vector<float> output((nsamps/decimation)+1, 0.0);

   // filter operates one past length specified
  filt.filter(10,samples.data(),output.data());
  if (output[10] != 0.0) {
    std::cout << "An invalid memory access occurred" << std::endl;
  }
  
  return 0;
}
```

marcusmueller:
I hoped #7093 would be the only one of this quality; @argilo  would you take a look?

Repro repo with Makefile for your comfort: https://gist.github.com/marcusmueller/4a16f0faddaf9a52c6fb6bc65e52f0be

```c++
#include <gnuradio/filter/fft_filter.h>
#include <iostream>
#include <vector>

int main() {
  unsigned int ntaps = 5;
  unsigned int nsamps = 2000;
  unsigned int decimation = 32;

  // generate taps
  std::vector<float> taps(ntaps, 1);

  // generate samples
  std::vector<float> samples(nsamps, .5);

  // generate filter
  gr::filter::kernel::fft_filter_fff filt(32, taps, 1);

  // output vector - make large enough
  std::vector<float> output((nsamps / decimation) + 1, 0.0);

  // filter operates one past length specified
  filt.filter(10, samples.data(), output.data());
  if (output[10] != 0.0) {
    std::cerr << "An invalid memory access occurred. Expected 0.0f, got "
              << output[10] << "\n";
  }

  return 0;
}
```

output 

```text
An invalid memory access occurred. Expected 0.0f, got 2.5
```

argilo:
From the documentation, it looks like some bounds checks are intentionally omitted:

[...]

Issue #3493, opened 2020-05-20
filter: firdes functions can return denormal or near denormal numbers


mrjacobagilbert:
This can result in substantial performance impact.

Consider the following:
```python
>>> filter.firdes.gaussian(1,16,3,21)
(3.2424925425704807e-40, 2.1742435751068437e-32, ... , 2.1742435751068437e-32, 3.2424925425704807e-40, 0.0)
>>> filter.firdes.gaussian(1,16,2,21)
(1.3652488831168542e-18, 4.107945469791589e-15, ... , 4.107945469791589e-15, 1.3652488831168542e-18, 1.863411258255796e-22)

```
The first results in a filter with two denormal filter coefficients, the impact of which is extreme. In the following flowgraph on my particular system:

![image](https://user-images.githubusercontent.com/7121282/82458571-0ccf0880-9a74-11ea-90df-6e5318db0eb7.png)

Using normal taps results in a throughput of ~32 MS/s
Using the first set of taps with two denormal values results in a throughput of ~ 8.8 MS/s
Using a set of taps which are all denormal results in a throughput of ~2.2 MS/s

This is a contrived example of a real world issue I faced on two different hardware platforms due to a single denormal tap returned by firdes cutting throughput in half.

Even near-denormal taps can cause issues as they can produce denormal output which is also a performance problem.

GR cannot save users completely from themselves, but it should probably not return denormal values from tap generators. Other recommendations from chat.gnuradio are to use compiler flags to set the FTZ and DAZ flags.

mrjacobagilbert:
Note: using the `-ffast-math` complier flag (as suggested by @noc0lour) does resolve this, however it seems like this does more than just enable the DAZ and FTZ flags.

noc0lour:
For avoiding this issue in real time systems we could add a build option to set the correct flags (not `fastmath`). Maybe coercing taps from (almost) denormal to 0 in filter design function is a good idea.

marcusmueller:
should consider using a better implementation of a Gaussian shape in firdes. For now, this persists. (it's really not hard to write a better Gaussian generator, we can calculate an acceptable cutoff by the area under the curve we're losing when setting small coefficients to zero)

PR #4176, opened 2021-02-01
Add new in-tree module gr-pdu and move existing pdu functionality out of gr-blocks

jacobagilbert:
The `gr-pdu` in tree module is intended to house PDU processing blocks, and enable expansion of GNU Radio's PDU API without cluttering the catch-all `gr-blocks`. This commit builds and passes QA though there may be some minor doxygen kinks to work out.

Specifically, this PR moves the following items from `gr-blocks` to `gr-pdu`:
  - pdu (noblock)
  - pdu_filter block
  - pdu_remove block
  - pdu_set block
  - pdu_to_tagged_stream block
  - random_pdu block
  - socket_pdu block
  - stream_pdu_base (noblock)
  - tagged_stream_to_pdu block
  - tcp_connection (noblock)
  - tuntap_pdu block

And adds the following blocks from `gr-pdu_utils` to `gr-pdu`:
  - pdu_frequency_modulator_fc (formerly pdu_gmsk_fc, but this takes an arbitrary filter as an input and was poorly named originally)
  - pdu_split

This commit also removes the `gr::blocks::pdu` namespace and incorporates this directly into `gr::pdu`.

The intention is to get this merged so that it is not waiting and becoming stale while I go through and add the rest of the blocks that are in GREP-0020 from the `gr-pdu_utils` OOT. **This PR breaks the `gr-blocks` API,** but this seems like a reasonable way to go about this.

This is a follow-up PR to #4022, and is now intended to be mergeable whenever 3.10 is stable. There is some more discussion of the motivation there.

The GREP this begins addressing is [here](https://github.com/gnuradio/greps/pull/30/files).


willcode:
I did not try to build, but the PR looks sane to me. Wondering if the PDU frequency modulator should go in gr-analog even though it works with PDUs. Eventually it could be combined with the streaming version. If we're going to have a number of PDU processing blocks in the future, we wouldn't want them all piled in gr-pdu any more than we'd want all our other blocks piled in gr-stream.

jacobagilbert:
@willcode That is a reasonable suggestion. There are a few areas where the streaming API does dangerous memory accesses (see #3493 and #3519) that are only safe (and I don't think guaranteed to be) due to the streaming circular buffers. That said, it sounds like you are advocating that these blocks be added directly to the appropriate module and combined some point in the future.

Taking a quick look over my GREP suggestion:

```
  +add_wall_clock_time
  +align
  +binary_tools
  +complex_to_mag2
  +filter
  +fir_filter                      ---> gr-filter
  +gmsk_fc                         ---> gr-analog (as pdu_frequency_modulator_fc)
  +length_filter
  +logger
  +pack_unpack
  +pdu_lambda
  +pdu_to_bursts
  +pdu_to_tagged_stream
  +pfb_resamp                      ---> gr-filter
  +preamble
  +quadrature_demod_cf             ---> gr-analog
  +random_pdu
  +remove_meta
  +set_meta
  +socket_pdu
  +split
  +tagged_stream_to_pdu
  +tag_message_trigger
  +tags_to_pdu
  +take_skip_to_pdu
  +time_delta
  +tuntap_pdu
  +whole_packet_clock_recovery     ---> gr-digital
```

jacobagilbert:
@willcode I removed the PDU FM block for now so this is mergable. Thoughts on the suggestions above?

mormj:
agree with the suggestions here 

things that are pdu versions of signal processing blocks should be either part of the same block, or in the same module - things that are pdu utilities should be in gr-pdu

willcode:
Ideally, work() and the message handler would both call the same processing code. Control messages would be common. The PDU version would be passing PMTs instead of using the vmcircbuf, so history has to be supported by individual blocks. Or better yet by a base class. I'm still thinking through the use cases, and wonder how Barry will explain this to the rest of the world :-)

mbr0wn:
I think we should get the GREP done before we proceed here. I think this is all going in the right direction, but I'd like to get some high-level agreement before doing the line-by-line review.

jacobagilbert:
Makes sense to me @mbr0wn - I updated a couple things over there per feedback from Josh.

mbr0wn:
@jacobagilbert quick note on expectation: Major changes like adding a new in-tree module should linger a while in review just so we get it right. It's not a sign of us ignoring this PR!

mbr0wn:
Can you please rebase.

jacobagilbert:
@mbr0wn yes I will do that shortly. And I totally understand the need for thorough review on the concept and implementation.

jacobagilbert:
@mbr0wn rebased, including pulling over the changes in ce88621d8e3e45cb808227d417f1e9e839a15453 which git did not catch. I don't think there were any other changes to what was moved from gr::blocks.

mormj:
Looks like there might be some other shimming necessary to not break API completely - tried to run `strobe.grc` from my 3.9 prefix:
```
/usr/bin/python3 -u /share/gnuradio/gr39/src/gnuradio/gr-blocks/examples/msg_passing/strobe.py
  File "/share/gnuradio/gr39/src/gnuradio/gr-blocks/examples/msg_passing/strobe.py", line 79, in <module>
    main()
  File "/share/gnuradio/gr39/src/gnuradio/gr-blocks/examples/msg_passing/strobe.py", line 57, in main
    tb = top_block_cls()
  File "/share/gnuradio/gr39/src/gnuradio/gr-blocks/examples/msg_passing/strobe.py", line 33, in __init__
    self.blocks_tagged_stream_to_pdu_0 = blocks.tagged_stream_to_pdu(blocks.byte_t, 'packet_len')
AttributeError: module 'gnuradio.blocks' has no attribute 'byte_t'
```

jacobagilbert:
@mormj good catch. that was moved as were a few other noblocks. It does not look like tcp_connection or stream_pdu_base are used outside the specific blocks that were moved. 

@mbr0wn they were previously in gr::blocks::pdu::x` but exposed to python in `gr.blocks.x`. I moved them to `gr::pdu::x` and exposed them as `gr.pdu.x`. Your suggestion does make sense though. It might be nice to put the message port names as `pdu.ports.<PORTNAME>` also.

I'll push a fix expanding the shim shortly.

jacobagilbert:
Updated, this issue affected two blocks which are widely used in our in-tree examples, the following were affected and should work now from 3.9 or this PR:

```
gr-blocks/examples/msg_passing/strobe.grc
gr-digital/examples/burst_shaper.grc
gr-digital/examples/packet/packet_tx.grc
gr-digital/examples/packet/formatter_crc.grc
gr-digital/examples/packet/formatter_ofdm.grc
gr-digital/examples/packet/simple_bpsk_tx.grc
gr-digital/examples/packet/tx_stage4.grc
gr-digital/examples/packet/tx_stage6a.grc
gr-digital/examples/packet/packet_rx.grc
gr-digital/examples/packet/tx_stage6.grc
gr-digital/examples/packet/tx_stage5.grc
gr-fec/examples/fecapi_async_decoders.grc
gr-fec/examples/fecapi_async_encoders.grc
gr-fec/examples/fecapi_async_ldpc_decoders.grc
gr-fec/examples/fecapi_async_ldpc_encoders.grc
gr-fec/examples/fecapi_async_packed_decoders.grc
gr-fec/examples/fecapi_async_to_stream.grc
gr-fec/examples/fecapi_polar_async_packed_decoders.grc
gr-fec/examples/fecapi_polar_encoders.grc
gr-qtgui/examples/qtgui_message_inputs.grc
```

mormj:
Tested out the set of grc files from the maint-3.9 tree, but rendered and run with gr-pdu enabled code on top of master:

4 failures with _both_ 3.9 and 3.10 gnuradio (all running the maint-3.9 grc files) - likely not caused by this PR:
```
gr-digital/examples/packet/formatter_ofdm.grc
gr-digital/examples/packet/tx_stage6.grc
gr-digital/examples/packet/tx_stage5.grc
gr-digital/examples/packet/tx_stage6a.grc
```

`gr-fec/examples/fecapi_async_ldpc_encoders.grc` has different behavior between maint3.9 and master.  May or may not be related to this PR - will need more investigation.  Basically the alist version signal doesn't line up with the H version

jacobagilbert:
I did not actually test all of those, but i did run a few and grcc the rest. I just went back and can confirm what you saw. 

The specific error for those flowgraphs is somewhat interesting and is documented in #4282.

jacobagilbert:
@mbr0wn I pushed a commit that moves the PDU vector_types into their own namespace/submodule. I liked how that came out so I also moved the `gr::pdu::PMTCONSTSTR__xx()` getters to `gr::pdu::ports::xx()`.

docs/usage-manual/(exported from wiki) Message Passing.txt
```diff
@@ -168,7 +168,7 @@ can call a block's gr::basic_block::_post method directly and pass it a
 message. So any block with an input message port can receive messages
 from the outside in this way.
 
-The following example uses a gr::blocks::pdu_to_tagged_stream block
```
mormj, line 171: @777arc @duggabe - I'm not too familiar with the "exported from wiki" procedure.  Is this file something that should be manually edited until we update the export?

gr-blocks/python/blocks/pdu_compatibility.py
```diff
@@ -0,0 +1,60 @@
+#
+# Copyright 2021 Free Software Foundation, Inc.
+#s
```
mormj, line 3: stray s?

mormj (approved):

docs/usage-manual/(exported from wiki) Message Passing.txt
jacobagilbert: grep caught this but I left it as its an export. i think i will have a non-trivial amount of wiki updating to do when this is merged.

gr-blocks/python/blocks/pdu_compatibility.py
jacobagilbert: thanks, fixed.

gr-blocks/grc/blocks.tree.yml
```diff
@@ -114,9 +107,6 @@
 - Modulators:
   - blocks_vco_f
   - blocks_vco_c
-- Networking Tools:
-  - blocks_tuntap_pdu
-  - blocks_socket_pdu
```
jdemel, line 119: I agree that these PDU blocks should go into `gr-pdu` for now. Though, I'd expect a `gr-network` or `gr-ip` module that groups all the network stuff together at some point.

gr-blocks/python/blocks/pdu_compatibility.py
```diff
@@ -0,0 +1,60 @@
+#
+# Copyright 2021 Free Software Foundation, Inc.
```
jdemel, line 2: I assume this should state "Jacob Gilbert" because we use DCO now?

gr-blocks/python/blocks/pdu_compatibility.py
```diff
@@ -0,0 +1,60 @@
...
+from gnuradio import pdu
+
+byte_t = pdu.types.byte_t
+float_t = pdu.types.float_t
+complex_t = pdu.types.complex_t
+vector_type = pdu.types.vector_type
+
+class pdu_filter(pdu.pdu_filter):
+    def __init__(self, *args, **kwargs):
+        print("WARNING: `pdu_filter` has moved to gr-pdu and will be removed from gr-blocks soon. Please update to use pdu.pdu_filter()")
```
jdemel, line 24: I assume this should go to `gr.log.warn`.

gr-pdu/CMakeLists.txt
```diff
@@ -0,0 +1,57 @@
+# Copyright 2021 Free Software Foundation, Inc.
```
jdemel, line 1: "Jacob Gilbert" instead of "FSF"?

gr-pdu/docs/CMakeLists.txt
```diff
@@ -0,0 +1,11 @@
+# Copyright 2021 Free Software Foundation, Inc.
```
jdemel, line 1: "Jacob Gilbert" instead of "FSF"?

gr-pdu/lib/CMakeLists.txt
```diff
@@ -0,0 +1,59 @@
...
+
+    configure_file(
+        ${CMAKE_CURRENT_SOURCE_DIR}/gnuradio-pdu.rc.in
+        ${CMAKE_CURRENT_BINARY_DIR}/gnuradio-pdu.rc
+    @ONLY)
+
+    target_sources(gnuradio-pdu PRIVATE
+        ${CMAKE_CURRENT_BINARY_DIR}/gnuradio-pdu.rc
+    )
+endif(MSVC)
```
jdemel, line 38: Is this something new? I haven't seen it in any OOT so far. Or is it specific to this module?

gr-pdu/lib/pdu_split_impl.cc
```diff
@@ -0,0 +1,60 @@
...
+    if (!(pmt::is_pdu(pdu))) {
+        GR_LOG_WARN(d_logger, "Message received is not a PDU, dropping");
+        return;
+    }
+
+    pmt::pmt_t meta = pmt::car(pdu);
+    pmt::pmt_t data = pmt::cdr(pdu);
+
+    if ((!pmt::equal(meta, pmt::PMT_NIL)) | d_pass_empty_data) {
+        message_port_pub(ports::dict(), pmt::car(pdu));
```
jdemel, line 51: Since there exists a `meta` variable, wouldn't it be better to use that one here?

gr-pdu/lib/pdu_split_impl.cc
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    if (pmt::length(data) | d_pass_empty_data) {
+        message_port_pub(ports::data(), pmt::cdr(pdu));
```
jdemel, line 55: see meta

gr-pdu/lib/pdu_split_impl.h
```diff
@@ -0,0 +1,40 @@
...
+
+#include <gnuradio/pdu/pdu_split.h>
+
+namespace gr {
+namespace pdu {
+
+class pdu_split_impl : public pdu_split
+{
+private:
+    bool d_pass_empty_data;
```
jdemel, line 22: Should it be possible to change that value at runtime? I suggest to make it `const`. as well as the ctor and make signatures. That would convey the intent.

gr-pdu/python/pdu/bindings/pdu_python.cc
```diff
@@ -0,0 +1,84 @@
...
+ *
+ */
+
+/***********************************************************************************/
+/* This file is automatically generated using bindtool and can be manually edited  */
+/* The following lines can be configured to regenerate this file during cmake      */
+/* If manual edits are made, the following tags should be modified accordingly.    */
+/* BINDTOOL_GEN_AUTOMATIC(0)                                                       */
+/* BINDTOOL_USE_PYGCCXML(0)                                                        */
+/* BINDTOOL_HEADER_FILE(pdu.h)                                        */
```
jdemel, line 16: a few more spaces.

gr-pdu/python/pdu/bindings/pdu_split_python.cc
```diff
@@ -0,0 +1,45 @@
...
+ *
+ */
+
+/***********************************************************************************/
+/* This file is automatically generated using bindtool and can be manually edited  */
+/* The following lines can be configured to regenerate this file during cmake      */
+/* If manual edits are made, the following tags should be modified accordingly.    */
+/* BINDTOOL_GEN_AUTOMATIC(0)                                                       */
+/* BINDTOOL_USE_PYGCCXML(0)                                                        */
+/* BINDTOOL_HEADER_FILE(pdu_split.h)                                        */
```
jdemel, line 16: a few more spaces

gr-pdu/python/pdu/qa_pdu_split.py
```diff
@@ -0,0 +1,92 @@
...
+
+        self.tb.msg_connect((split, 'dict'), (d1, 'store'))
+        self.tb.msg_connect((split, 'data'), (d2, 'store'))
+
+        in_meta1 = pmt.dict_add(pmt.make_dict(), pmt.intern('num'), pmt.from_long(4))
+        in_meta2 = pmt.dict_add(pmt.make_dict(), pmt.intern('n'), pmt.from_long(99))
+        in_pdu = pmt.cons(in_meta1, pmt.init_u8vector(6, range(6)))
+
+        self.tb.start()
+        time.sleep(.001)
```
jdemel, line 38: Are these `time.sleep` lines here to make sure that some output is produced? @mbr0wn Added a spin wait function that would make sense here, I assume.

jdemel (commented):
It's great to see that `gr-blocks` gets reorganized. I have some suggestions for improvements. Of course, overall the PR looks good.

jacobagilbert:
Thanks for the in depth review @jdemel - I'll get to these asap. A couple general comments:

1. Regarding copyright: I did this on my own and am perfectly happy to leave the templated stuff as (C) FSF unless this is a problem. I do have a CLA that I assume did not evaporate its just not required anymore (?).

2. Does git-clang with --style=file do that space thing on python binding cc files to anyone else?

gr-pdu/lib/CMakeLists.txt
jacobagilbert: I based some of the templated code in this module off gr-trellis and this came from: https://github.com/gnuradio/gnuradio/blob/master/gr-trellis/lib/CMakeLists.txt#L45

I am not a CMake guru so let me know if it should be pulled.

gr-pdu/lib/pdu_set_impl.cc
```diff
@@ -29,10 +29,9 @@ pdu_set_impl::pdu_set_impl(pmt::pmt_t k, pmt::pmt_t v)
       d_k(k),
       d_v(v)
 {
-    message_port_register_out(pdu::pdu_port_id());
-    message_port_register_in(pdu::pdu_port_id());
-    set_msg_handler(pdu::pdu_port_id(),
-                    [this](pmt::pmt_t msg) { this->handle_msg(msg); });
+    message_port_register_out(ports::pdus());
+    message_port_register_in(ports::pdus());
+    set_msg_handler(ports::pdus(), [this](pmt::pmt_t msg) { this->handle_msg(msg); });
```
marcusmueller, line 34: Correct me, but if we change this lambda's, and the msg handler's, signature to `const pmt::pmt_t& msg`, we'd be saving the schmart pointer a ref increase (and decrease), which might be nice, seeing that these `shared_ptr`s imply refcount change atomicity.

Note that this is a general architectural problem with PMTs, not limited to this block, at all, but, knowing that you build high-rate applications out of this, this might be a good place to start that practice.

Definitely not a blocker here, but `perf` / kernelshark can be pretty harsh on how often we change refcounts.

gr-blocks/python/blocks/CMakeLists.txt
```diff
@@ -16,6 +16,7 @@ GR_PYTHON_INSTALL(
     msg_pair_to_var.py
     msg_meta_to_pair.py
     var_to_msg.py
+    pdu_compatibility.py # REMOVE IN 3.11
```
marcusmueller, line 19: You're effectively deprecating something, and announcing it for later removal. Since this can be hard to track, please also append to CHANGELOG.md.

marcusmueller (changes requested):

gr-blocks/python/blocks/pdu_compatibility.py
jacobagilbert: Agree

gr-pdu/lib/pdu_split_impl.cc
jacobagilbert: Agree, fixed.

gr-pdu/lib/pdu_split_impl.cc
jacobagilbert: Agree, fixed.

gr-pdu/python/pdu/bindings/pdu_python.cc
jacobagilbert: Fixed.

gr-pdu/python/pdu/bindings/pdu_split_python.cc
jacobagilbert: Fixed.

gr-pdu/python/pdu/qa_pdu_split.py
jacobagilbert: @jdemel yes, this is just to give a small amount of time to ensure the message handler is called before the flowgraph is stopped. Can you point me to an example of what @mbr0wn did?

gr-pdu/lib/pdu_set_impl.cc
jacobagilbert: I think this makes sense in general but should be a separate issue/PR to cover everything in-tree IMO.

gr-blocks/python/blocks/CMakeLists.txt
jacobagilbert: Agree, fixed.

jacobagilbert:
I pushed an update that addresses a number of issues. Comments/questions outstanding:

1. Copyright - @mormj are you OK if I keep personal copyright out of this and assign to FSF where i can?

2. MSVC CMake stuff from gr-tellis - does this need to be removed?

3. QA Sleep statement fix - better way to do this? Discussion above.

4. PDU network blocks in gr-pdu or gr-network - I'm not sure on this.... modules provide organization and build-selectivity. From a code organization it seems like it could go either way and from a build-selectivity perspective if I don't want PDU stuff, i will -DENABLE_GR_PDU=OFF and I would prefer if none of the PDU stuff was built (including network...). 

I 'm going to be a wimp on #4 and defer to our new maintainer: @mormj please tell me what to do!

gr-pdu/python/pdu/qa_pdu_split.py
jdemel: This was merged a few days ago with #4275

jacobagilbert: Perfect, thank you. I rebased so that is available and switched to use that, GREAT addition for message based QA @mbr0wn!

willcode:
@jacobagilbert I don't think we're assigning to FSF any more. DCO has you keep the copyright.

jacobagilbert:
> @jacobagilbert I don't think we're assigning to FSF any more. DCO has you keep the copyright.

Ok thanks I'll get that updated later on today.

willcode:
Don't know if we have a default copyright owner for people that don't want to hold a copyright. If you sign off on the DCO, I think that makes it you anyway. Maybe we need a default, like "GNU Radio Project".

jacobagilbert:
I'm not opposed to adding me as copyright holder it just seemed cleaner not to keep the template stuff under a single project copyright. No big deal either way.

gr-pdu/lib/CMakeLists.txt
jdemel: This is very MSVC specific. I found [this page](https://docs.microsoft.com/en-us/cpp/windows/resource-files-visual-studio?view=msvc-160). Someone with MSVC experience would be great. As far as I know, we don't have these files in other modules. Thus, I suggest to remove it here as well. It clutters this CMake file and makes the whole process more difficult to understand.

jacobagilbert: All other modules have this, but this was a good catch for windows (which I have not tried)  because I did not include the referenced gnuradio-pdu.rc.in file. I agree someone with MSVC would be awesome to look this over but for now I think doing what other modules have is the right thing.

mormj:
I tend to agree with @marcusmueller and @jdemel on the networking blocks - primarily they are networking in function, and happen to output a PDU, and are not strictly PDU utilities.  So let's put those in the gr-network module.

willcode:
@jacobagilbert Turns out if you have an existing CLA then you can still assign to FSF.

jacobagilbert:
@willcode thanks for the info

@mormj @jdemel @marcusmueller I'll get this refactored so the two PDU network blocks are in gr-network. Thanks all

gr-blocks/grc/blocks.tree.yml
jacobagilbert: These two blocks are now in gr-network.

jacobagilbert:
@mormj @willcode I refactored this a bit based on suggestions above. Change summary:
- moved `socket_pdu` and `tuntap_pdu` to `gr-network` along with corresponding QA.
- Moved pdu.h (and pdu.cc) into runtime
  - general purpose PDU functions are in the `gr::pdu::` namespace
  - types are in the `gr::types::` namespace
  - common port name getters are in the `gr::ports::` namespace
- added gr-network to `gnuradio/__init__.py`

Full diff: https://pastebin.com/ZV4smKK0

I think all of the comments were addressed. It will be relatively easy to adjust the namespace changes if the ones above don't work.

willcode:
I like it!

One minor thing - can ports:: be called portname(s):: instead? That way, it doesn't imply that it's some kind of port object.

jacobagilbert:
Sure. Good idea.

I'll also add in PDU vector_types int_t and short_t per our discussion on matrix.

willcode:
What is the purpose of all the bpdu, cpdu, etc., names anyway? I know they're just interned strings. Ports probably wouldn't be named for their data type, except where that indicates their function. Looks like gr-pdu_utils uses this to indicate pdu type to the user. This is not checked by `connect()`, right? A future thing would be to do type checking on PDU port connections so that the names are not important.

jacobagilbert:
@willcode the idea would be that a port named `pdu` should accept any PDU type, or at least multiple PDU types. b/f/cpdu indicate specific types only are accepted. As you note, this is a user indication only as any PDU (any PMT object actually) will be accepted. for example, the `pdu_split` block is going to operate on any PDU but `pdu_frequency_modulator` is going to require a float_t PDU (or maybe a byte_t PDU in the future) and is going to produce a complex_t PDU.

I do not see how with runtime can reasonably check/verify this our current libpmt, but a secondary objective of getting PDUs more integrated into GR is to get more thought behind using messages for data-plane operations and the necessary type checking when pmtf or whatever comes next for PMT is implemented.

The pdu_utils module is a bit all over the place on this, but I am using the upstream process (which will obviously require updates to current use of pdu_utils blocks) as an opportunity to standardize some of this type of thing.

jacobagilbert:
Also, when updating this code, there are some type-size-assumptions going on in these functions:
- [itemsize](https://github.com/gnuradio/gnuradio/blob/master/gr-blocks/lib/pdu.cc#L27)
- [make_pdu_vector](https://github.com/gnuradio/gnuradio/blob/master/gr-blocks/lib/pdu.cc#L55)

For the current libpmt I don't think we can get too much better. The latest commit I pushed adding short and int might be too awful to handle when it comes to this honestly as they bake in the assumption that int ≡ s32vector and short ≡ s16vector.

mbr0wn:
> I'm not opposed to adding me as copyright holder it just seemed cleaner to keep the boilerplate stuff under a single project copyright. No big deal either way.

@willcode @jacobagilbert : Just an FYI: Legally speaking, anyone with a CLA can still assign the copyright to FSF. Some people might not like having their name on things so this is a totally valid option. Of course, this requires having a CLA filed previously.

willcode:
@jacobagilbert In the process of integrating pdu and streaming blocks, we should figure out how to give pdu ports functional names, specify types (which may be richer than those for streams) and integrate into GRC.

gr-pdu/examples/CMakeLists.txt
```diff
@@ -0,0 +1,12 @@
+# Copyright 2021 Jacob Gilbert
+#
+# This file is part of GNU Radio
+#
+# SPDX-License-Identifier: GPL-3.0-or-later
+#
+
+install(
+  FILES
+  DESTINATION ${GR_PKG_DATA_DIR}/examples/pdu
```
mormj, line 10: It would be nice to have some example .grc files for gr-pdu in here.  Shouldn't hold up the merge but would be a good part of the overall package.

jacobagilbert: I agree. I'll build some with the next addition of blocks.

mormj:
Looks like the feedback has been addressed - going ahead and merging this!  Thanks @jacobagilbert for bringing your excellent work in-tree.

PR #4180, opened 2021-02-03
grc: Deduplicate block paths

mbr0wn:
GRC reads paths from various sources, including the GRC_BLOCKS_PATH
environment variable. It can happen that paths appear twice in the list,
causing warnings about findings blocks twice.

This deduplicates the paths in the config object.


skoslowski (approved):

willcode:
Already backported to maint-3.8, removing label.

PR #4186, opened 2021-02-05
gr-uhd: ALL_LOS binding

karel:
This is a patch to prevent the following issue, when using GNU Radio (master) to configure USRP LO sharing.
```
  File "/home/user/n321_rx.py", line 125, in __init__
    self.uhd_usrp_source_0.set_lo_source('external', uhd.ALL_LOS, 0)
AttributeError: module 'gnuradio.uhd' has no attribute 'ALL_LOS'
```


mbr0wn (approved):

willcode:
Backported to maint-3.9 as part of https://github.com/gnuradio/gnuradio/pull/4187.

PR #4188, opened 2021-02-07
If GSL not found exclude modules requiring GSL from python bindings

dl1ksv:
If gnuradio was build without GSL test  105 -117 fail with 
    from gnuradio import gr, gr_unittest, blocks, fec
  File "/home/schroer/gnuradiocomponents/gnuradio-volker/gr-fec/python/fec/__init__.py", line 21, in <module>
    from .fec_python import *
ImportError: /home/schroer/gnuradiocomponents/gnuradio-volker/build/gr-fec/python/fec/bindings/fec_python.cpython-39-x86_64-linux-gnu.so: undefined symbol: _Z18bind_ldpc_G_matrixRN8pybind117module_E

Some modules depend on GSL_FOUND. So they should be excluded in the python bindings, if GSL was not found.
Other way to check than running the tests:
Run in python 
from gnuradio import fec

### DIFF
--- a/gr-fec/python/fec/bindings/CMakeLists.txt
+++ b/gr-fec/python/fec/bindings/CMakeLists.txt
@@ -70,5 +70,7 @@ GR_PYBIND_MAKE_CHECK_HASH(fec
    "${fec_python_files}") 
 
 target_link_libraries(fec_python PUBLIC gnuradio-runtime gnuradio-blocks)
-
+if(GSL_FOUND)
+    target_compile_definitions(fec_python PUBLIC -DHAVE_GSL)
+endif(GSL_FOUND)
 install(TARGETS fec_python DESTINATION ${GR_PYTHON_DIR}/gnuradio/fec COMPONENT pythonapi)
--- a/gr-fec/python/fec/bindings/python_bindings.cc
+++ b/gr-fec/python/fec/bindings/python_bindings.cc
@@ -33,18 +33,20 @@ void bind_dummy_decoder(py::module&);
 void bind_dummy_encoder(py::module&);
 void bind_encode_ccsds_27_bb(py::module&);
 void bind_encoder(py::module&);
-void bind_fec_mtrx(py::module&);
 void bind_generic_decoder(py::module&);
 void bind_generic_encoder(py::module&);
 void bind_gf2mat(py::module&);
 void bind_gf2vec(py::module&);
+#ifdef HAVE_GSL
+void bind_fec_mtrx(py::module&);
 void bind_ldpc_G_matrix(py::module&);
 void bind_ldpc_H_matrix(py::module&);
 void bind_ldpc_bit_flip_decoder(py::module&);
-void bind_ldpc_decoder(py::module&);
-void bind_ldpc_encoder(py::module&);
 void bind_ldpc_gen_mtrx_encoder(py::module&);
 void bind_ldpc_par_mtrx_encoder(py::module&);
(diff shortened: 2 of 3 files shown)


gr-fec/python/fec/bindings/CMakeLists.txt
```diff
@@ -51,6 +51,7 @@ list(APPEND fec_python_files
     )
 
 if(GSL_FOUND)
+add_definitions(-DENABLE_GSL)
```
mbr0wn, line 54: This is an odd place to put it, especially with that indent. Also, the CMake doc says this has been superseded.

I suggest using `target_compile_definitions()` further down. Also, please rename it to `HAVE_GSL` for consistency. The `ENABLE_*` are used as user input options. Checks for existence in the code use `HAVE_`.
```cmake
target_compile_definitions(fec_python PUBLIC -DHAVE_GSL)
```

gr-fec/python/fec/bindings/python_bindings.cc
```diff
@@ -33,18 +33,20 @@ void bind_dummy_decoder(py::module&);
 void bind_dummy_encoder(py::module&);
 void bind_encode_ccsds_27_bb(py::module&);
 void bind_encoder(py::module&);
-void bind_fec_mtrx(py::module&);
 void bind_generic_decoder(py::module&);
 void bind_generic_encoder(py::module&);
 void bind_gf2mat(py::module&);
 void bind_gf2vec(py::module&);
+#ifdef ENABLE_GSL
```
mbr0wn, line 40:
```suggestion
#ifdef HAVE_GSL
```

gr-fec/python/fec/bindings/python_bindings.cc
```diff
@@ -104,16 +106,18 @@ PYBIND11_MODULE(fec_python, m)
     bind_dummy_decoder(m);
     bind_dummy_encoder(m);
     bind_encode_ccsds_27_bb(m);
-    bind_fec_mtrx(m);
     // bind_gf2mat(m);
     // bind_gf2vec(m);
+#ifdef ENABLE_GSL
```
mbr0wn, line 111:
```suggestion
#ifdef HAVE_GSL
```

mbr0wn (changes requested):
Very useful change, thanks. Can you please address the comments, and also change your commit message to something like:
```
fec: Exclude modules from Python bindings if not GSL is present

This removes the same blocks from the Python bindings that are already not
present in the C++ library if GSL is not present.
```
thanks!

gr-fec/python/fec/bindings/python_bindings.cc
dl1ksv: Thanks for your comments. I apply your proposals

mbr0wn (approved):

mbr0wn (changes requested):
See the CMake error:
```
CMake Error at gr-fec/python/fec/bindings/CMakeLists.txt:61 (target_compile_definitions):
  Cannot specify compile definitions for target "fec_python" which is not
  built by this project.
```

I think the `target_compile_definitions()` needs to go further down (it'll need another if-clause).

dl1ksv:
target_compile_definitions has to go behind target_link_libraries.
Without installed GSL yo don't get the error above.

mbr0wn (approved):
Excellent! CMake is passing.

mormj (approved):
LGTM!

PR #4190, opened 2021-02-07
github: Add Centos 7 to GH-A, separate Doxygen

mbr0wn:
- This adds another worker for compiling on Centos7. So far, only Python
  3 works there.
- The Doxygen build step is farmed into its own worker.


mbr0wn:
Merged manually.

PR #4195, opened 2021-02-08
modtool: add default .gitignore

mormj:
Every time I create a new OOT, have to remember to not check in the build dir and create a .gitignore

This adds a .gitignore to remove that step from the process

### DIFF
--- a/gr-utils/modtool/templates/gr-newmod/.gitignore
+++ b/gr-utils/modtool/templates/gr-newmod/.gitignore
@@ -1,0 +1,5 @@
+*~
+*.pyc
+*.pyo
+build*/
+examples/grc/*.py


mbr0wn (approved):

gr-utils/modtool/templates/gr-newmod/.gitignore
```diff
@@ -0,0 +1,5 @@
+*~
+*.pyc
+*.pyo
+build*/
+example/*.py
```
mbr0wn, line 5:
```suggestion
```
This line assumes that there will be no Python-based examples. Maybe we can have an entry like
```suggestion
examples/grc/*.py
```
instead?

Also, it's example*s*.

mbr0wn (changes requested):

gr-utils/modtool/templates/gr-newmod/.gitignore
mormj: Yes - that makes sense!

mbr0wn (approved):

PR #4198, opened 2021-02-09
filter: cmake: Warn if scipy or pyqtgraph are unavailable

mbr0wn:
Both are dependencies for the filter design tool, but are otherwise not
build dependencies. We thus warn during CMake.


mbr0wn:
Supersedes https://github.com/gnuradio/gnuradio/pull/3987.

dl1ksv (approved):
Works perfect

PR #4199, opened 2021-02-09
filter: remove unnecessary set_history(1)

mormj:
since d_history is set to 1 in gr::block, these statements are
unnecessary


mbr0wn (approved):

willcode (approved):
All but the last one (pfb_synthesizer_ccf) are in constructors, where this definitely duplicates the initializer in the block class. The last one, I believe is also safe, even though it's in set_taps.

willcode:
Backported to maint-3.9 in https://github.com/gnuradio/gnuradio/pull/4205

PR #4200, opened 2021-02-09
fec: If GSL is missing, remove those yml files that require GSL

dl1ksv:
Some fec require GSL and are not build, if GSL is missing.

In this case, the corresponding yml files should be removed from grc, too. 


mbr0wn (approved):

willcode:
Backported to maint-3.9 in https://github.com/gnuradio/gnuradio/pull/4205

Issue #1740, opened 2018-04-19
Adding automatic unit testing for .grc grcc compilation


marcusmueller:
grcc compilability could be a "collatoral" unit test that we could have for every single .grc added to the tree. 

* CMake needs to automagically find all the .grc
* … and to add them to a list of files to be pushed through `grcc` during a unit test
* … and to check the resulting's Python syntax

This would allow us to detect f'ups caused by GRC changes in non-GRC code.

PR #4208, opened 2021-02-11
grc: Test all in-tree examples

haakov:
See https://github.com/gnuradio/gnuradio/pull/3535
Addresses https://github.com/gnuradio/gnuradio/issues/3510

This way, each example (162 in total) is registered as a separate test.

#### Run without PYTHONPATH and LD_LIBRARY_PATH set:
~~70~~ 115  passed, 7 warnings
~~76~~ 47 failed, most are either gr-fec or gr-trellis

#### Run with PYTHONPATH and LD_LIBRARY_PATH set:
~~110~~ 155 passed, 7 warnings
~~36~~ 7 failed, 6 from gr-trellis and tpc_ber_curve_gen.grc (seems to be a problem with bus ports)
Not quite sure why I'm getting the port errors, the flowgraphs seem fine when opening them in GRC.

### DIFF
--- a/grc/tests/test_examples.py
+++ b/grc/tests/test_examples.py
@@ -1,0 +1,67 @@
+import os
+
+from os import path
+
+import pytest
+
+from grc.core.platform import Platform
+
+try:
+    os.mkdir(path.join(path.dirname(__file__), 'resources/tests'))
+except FileExistsError:
+    pass
+
+# Gather blocks
+BLOCK_PATHS = []
+ROOT = path.join(path.dirname(__file__), '../..')
+BLOCK_PATHS = [path.join(path.dirname(__file__), '../../grc/blocks'), '../../build/gr-uhd/grc']
+for file_dir in os.scandir(ROOT):
+    # If it is a module
+    if path.isdir(file_dir) and file_dir.name.startswith("gr-"):
+        BLOCK_PATHS.append(path.join(file_dir, "grc"))
+
+def gather_examples():
+    global ROOT
+    example_paths = []
+    for file_dir in os.scandir(ROOT):
+        # If it is a module
+        if path.isdir(file_dir) and file_dir.name.startswith("gr-"):
+            try:
+                for pos_ex in os.scandir(path.join(file_dir, "examples")):
+                    if path.isfile(pos_ex) and pos_ex.name.endswith(".grc"):
+                        example_paths.append(pos_ex)
+
+                # Some modules have their .grc files in a subdirectory called "grc"
+                for pos_ex in os.scandir(path.join(file_dir, "examples/grc")):
+                    if path.isfile(pos_ex) and pos_ex.name.endswith(".grc"):
+                        example_paths.append(pos_ex)
+            except FileNotFoundError:
+                continue
+    return example_paths
+
+def print_proper(element):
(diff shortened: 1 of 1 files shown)


mormj (approved):
This is much more elegant than my original draft - probably another PR to tie it in to make test if we want to - too many in-tree failures right now to start enforcing it in QA anyway.  But looks great to me!

grc/tests/test_examples.py
```diff
@@ -0,0 +1,70 @@
+# Copyright 2016 Free Software Foundation, Inc.
```
mbr0wn, line 1:
```suggestion
# Copyright 2021 Haakov Industries
```
(or whatever)

grc/tests/test_examples.py
```diff
@@ -0,0 +1,70 @@
...
+#
+# This file is part of GNU Radio
+#
+# SPDX-License-Identifier: GPL-3.0-or-later
+#
+
+import pytest
+import os
```
mbr0wn, line 9:
```suggestion
import os
import pytest
```

I'm sure Pylint prefers this.

mbr0wn (changes requested):
Can you run Pylint against it and make sure the most obvious warnings are resolved?

haakov:
Remaining pylint warnings:
```
************* Module grc.tests.test_examples
grc/tests/test_examples.py:64:0: C0301: Line too long (132/100) (line-too-long)
grc/tests/test_examples.py:1:0: C0114: Missing module docstring (missing-module-docstring)
grc/tests/test_examples.py:26:8: W0621: Redefining name 'file_dir' from outer scope (line 18) (redefined-outer-name)
grc/tests/test_examples.py:23:0: C0116: Missing function or method docstring (missing-function-docstring)
grc/tests/test_examples.py:24:4: W0603: Using the global statement (global-statement)
grc/tests/test_examples.py:42:0: C0116: Missing function or method docstring (missing-function-docstring)
grc/tests/test_examples.py:48:0: C0116: Missing function or method docstring (missing-function-docstring)
grc/tests/test_examples.py:49:4: W0603: Using the global statement (global-statement)
```

haakov:
Updated numbers in PR text/comment

mbr0wn:
As you can see, both F33 and CentOS 8.3 run this test, and it fails. So how do we proceed? I really like this test -- but if it causes 100% test failure, it's not something we can merge in.

mormj:
@haakov  - I'm going to close this for now until we can address the persistent failures on Fedora and CentOS.  But, adding it to the project tracker for the next release, as I think this is an important goal to achieve in the release cycle.  Just trying to keep the PR queue as clean as possible.  This is a good implementation so closing it definitely does not mean it will go away.  If you (or anyone else) come up with fixes, please open it back up.  Thanks!

mormj:
Also addresses https://github.com/gnuradio/gnuradio/issues/1740

mormj:
@haakov - I really want this test in, to be run manually at first until all the examples are fixed.  Just pushed a commit to disable it from running during `make test`, but can be run from `build/` with `python3 -m pytest test_examples.py`.  If you are ok with this, we can merge it and enable the test when the examples are fixed.

haakov:
Sounds good :+1:

PR #4215, opened 2021-02-12
zeromq: Fix warnings with recv()

mbr0wn:
The recv() call on a ZMQ socket produces a warning if the return value
is not stored. We follow the advice and check the return value, just in
case.

Signed-off-by: Martin Braun <martin.braun@ettus.com>

### DIFF
--- a/gr-zeromq/lib/rep_sink_impl.cc
+++ b/gr-zeromq/lib/rep_sink_impl.cc
@@ -59,10 +59,15 @@ int rep_sink_impl::work(int noutput_items,
         /* Get and parse the request */
         zmq::message_t request;
 #if USE_NEW_CPPZMQ_SEND_RECV
-        d_socket.recv(request);
+        bool ok = bool(d_socket.recv(request));
 #else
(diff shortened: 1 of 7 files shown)


gr-zeromq/lib/rep_sink_impl.cc
```diff
@@ -59,10 +59,15 @@ int rep_sink_impl::work(int noutput_items,
         /* Get and parse the request */
         zmq::message_t request;
 #if USE_NEW_CPPZMQ_SEND_RECV
-        d_socket.recv(request);
+        bool ok = bool(d_socket.recv(request));
```
mbr0wn, line 62: const

willcode:
Casting to `(void)` is easier than making up error messages you don't really want.

mbr0wn:
> Casting to `(void)` is easier than making up error messages you don't really want.

Yeah but zeromq really wants us to read the result. Besides, there is a tiny chance we run into this issue and then at least we have proper error handling.

Issue #4191, opened 2021-02-08
Message Strobe/Debug Error?


bmagistro:
Was looking to use the `message strobe` block to send a periodic message for testing.  While setting up and validating things, had the output connected to a `message debug` block, specifically the `print_pdu` input.   When trying to run, it throws the following error: `thread[thread-per-block[0]: <block message_debug(5)>]: pmt_car: wrong_type : TEST`.

Attached is a minimal flow graph that can replicate this issue.
[strobe.grc.txt](https://github.com/gnuradio/gnuradio/files/5940659/strobe.grc.txt)

GR 3.8.1 via pybombs
Ubuntu 18.04

willcode:
Message PMT should be of the form `pmt.to_pmt(['meta','vector'])`.

willcode:
Here's one that actually works: `pmt.cons(pmt.PMT_NIL, pmt.init_u8vector(3,[0x01,0x02,0x03]))`.

bmagistro:
Thanks for the insight, while I can manually transform a string to a vector easy enough, the defaults for the blocks seem to say a string should work.  Your comments also make me wonder if the defaults for the `message strobe` and `message strobe random` blocks are wrong then?

Will keep looking at it later.

dkozel:
The Message Strobe default value correctly displays if you connect it to the Message Debug print port.

Some nice improvements to the Message Debug blog were added to GNU Radio 3.9 in October.
https://github.com/gnuradio/gnuradio/commit/9248b29768c7810f029a1188726b2ad05141f382

This includes a check on the `print_pdu` port which discards messages that aren't pairs (or are null). It does discard silently. I think logging something would be useful. The older behavior was a bit obscure, but did print a technically useful message.

I think that there is no error in the latest code. Possibly back-porting the discard behavior and adding a print statement "Non-PDU message received" or similar would be a useful enhancement. What do you two think?

jacobagilbert:
I think adding a "non PDU message received on port print_pdu" or something is reasonable. This should also be updated to use the [3.9+ is_pdu() check](https://github.com/gnuradio/gnuradio/blob/master/gnuradio-runtime/lib/pmt/pmt.cc#L821) that was added. The current check is more restrictive than what is in this block as:

```
 A PDU (protocol data unit) in GNU Radio has a special PMT type, it is a pair of a dictionary (on CAR) and a uniform vector type.
```

https://wiki.gnuradio.org/index.php/Message_Passing#PDUs

[...]

PR #4220, opened 2021-02-13
gr-blocks: Update message debug error handling on Print PDU port

jacobagilbert:
Currently non PDU messages get silently dropped. This can mask connection issues; given this is a debug block a WARNing messages seems appropriate. 

Suggestion from #4191

Not sure this should be backported to 3.9 or not; it does slightly change the conditions under which data is dropped - let me know.


dkozel (approved):

dkozel:
I think backporting it makes sense. Changing the drop criteria to !is_pdu is a more correct behavior and I don't think substantially changes what gets dropped? The message is useful and a better behavior overall.

mbr0wn (approved):

PR #4222, opened 2021-02-13
gr_filter_sign: bessel bandpass calculation always fails

dl1ksv:
bessel bandpass calculation always fails as a wrong str to float
conversion is used. Using the same conversion as in lowpass calculation
fixes this problem.

StandardError is not valid in python 3. It was replaced by Exception.

### DIFF
--- a/gr-filter/python/filter/design/filter_design.py
+++ b/gr-filter/python/filter/design/filter_design.py
@@ -6,7 +6,6 @@
 #
 #
 
-
 import sys
 import os
 import re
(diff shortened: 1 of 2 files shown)


gr-filter/python/filter/design/fir_design.py
```diff
@@ -27,10 +27,14 @@ def design_win_lpf(fs, gain, wintype, mainwin):
...
                                             atten, wintype)
-        except RuntimeError as e:
+        except RuntimeError  as e:
             reply = QtGui.QMessageBox.information(mainwin, "Runtime Error",
                                                   e.args[0], QtGui.QMessageBox.Ok)
             return ([], [], ret)
+        except IndexError  as e:
+            reply = QtGui.QMessageBox.information(mainwin, "Runtime Error",
+                                                  e.args[0], QtGui.QMessageBox.Ok)
+            return ([], [], ret)
```
mbr0wn, line 37:
```suggestion
        except (RuntimeError, IndexError) as e:
            reply = QtGui.QMessageBox.information(mainwin, "Runtime Error",
                                                  e.args[0], QtGui.QMessageBox.Ok)
            return ([], [], ret)
```
Alternatively, have different error messages in the message box.

gr-filter/python/filter/design/filter_design.py
```diff
@@ -2166,7 +2165,7 @@ def action_open_dialog(self):
             self.gui.iirfilterTypeComboBox.setCurrentIndex(iirft[params["filttype"]])
             self.gui.iirfilterBandComboBox.setCurrentIndex(bandpos[params["bandtype"]])
             if params["filttype"] == "bessel":
-                critfreq = map(float, params["critfreq"][1:-1].split(','))
+                critfreq = list(map(float, params["critfreq"][1:-1].split(',')))
```
mbr0wn, line 2168: My preference would be:
```suggestion
                critfreq = [float(x) for x in params["critfreq"][1:-1].split(',')]
```

...which I've seen quoted as being more Pythonic, and I think it is too. But this is personal preference, so keep whichever version you prefer.

mbr0wn (changes requested):
Thanks, this is a great change. I have one nit about the last catch clause.

gr-filter/python/filter/design/filter_design.py
dl1ksv: looks more elegant.

dl1ksv:
Thanks for your comments .  Both changes pushed

mbr0wn (approved):

PR #4224, opened 2021-02-13
gr-digital: qa_header_payload_demux fix

willcode:
The HeaderToMessageBlock adapter assumed it would receive a complete
header in work(). Also, forecast() is not currently working for Python
blocks, so a local buffer was required in the adapter.

Two asserts needed list().

### DIFF
--- a/.github/workflows/make-test.yml
+++ b/.github/workflows/make-test.yml
@@ -49,7 +49,7 @@ jobs:
     - name: Make
       run: 'cd /build && make -j2'
     - name: Make Test
-      run: 'cd /build && ctest --output-on-failure -E "qa_uhd|qa_uncaught_exception|qa_header_payload_demux|qa_agc|qa_cpp_py_binding|qa_cpp_py_binding_set|qa_ctrlport_probes|qa_file_taps_loader|qa_qtgui"'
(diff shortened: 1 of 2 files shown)


mbr0wn:
Will this fix the QA issues associated with it? If so, can you please open .github/workflows/make-test.yml and remove the test. To run the CI, you will need to add this branch name to the pull-branches at the top, too. Then we can see if gets executed and passes.

willcode:
Tests passed. I removed the additional branch from make-test.yml. Not sure it did anything anyway. Seems to test fine on master.

drmpeg (approved):
Works good on 32-bit ARM.

drmpeg:
Why are qa_uhd and qa_uncaught_exception still disabled? I thought those were fixed.

Also why are qa_agc and qa_file_taps_loader disabled? When were those a problem?

willcode:
@drmpeg all of the disabled tests pass on my system, so no idea why they're disabled.

mbr0wn:
FYI, qa_uhd, qa_uncaught_exception and qa_file_taps_loader should be enabled. The rest still fail.

willcode:
Already in maint-3.9, commit aa6ced4795bd65d68b8778a1fb8a18d87b16b043.

PR #4225, opened 2021-02-14
gr-blocks: combine the print and print_pdu port of the message_debug block

jacobagilbert:
Having two print ports on this block has confused a number of people, most recently issue #4191, and combining them is straightforward and should make this block easier to use. This PR also updates some out of date documentation in the public header file.

Potentially breaks existing usage of this block so no backport.

### DIFF
--- a/gr-blocks/include/gnuradio/blocks/message_debug.h
+++ b/gr-blocks/include/gnuradio/blocks/message_debug.h
@@ -27,14 +27,13 @@ namespace blocks {
  * The message debug block is used to capture and print or store
  * messages as they are received. Any block that generates a
  * message may connect that message port to one or more of the
- * three message input ports of this debug block. The message
+ * two message input ports of this debug block. The message
  * ports are:
  *
  * \li print: prints the message directly to standard out.
  * \li store: stores the message in an internal vector. May be
  *            access using the get_message function.
- * \li print_pdu: specifically designed to handle formatted PDUs
- *                (see pdu.h).
+ * \li print_pdu: DEPRECATED! use print() for all printing
  */
 class BLOCKS_API message_debug : virtual public block
 {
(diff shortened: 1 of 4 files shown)


dkozel:
I'm inclined to say this is useful but should be run through a deprecation period like is being done with the Signal Source block's message port. (On which note we should remove that port, it was deprecated in 3.8 so definitely ok to remove in 3.10.)

https://github.com/gnuradio/gnuradio/blob/ec3d116546aa9710b324b1713cea058d80c906a8/gr-analog/lib/sig_source_impl.cc#L70-L81

@mbr0wn What do you think about the correctness of adding a deprecation warning to this port in maint-3.9? Is it poor form to deprecate something on a maint branch after the initial release?

mbr0wn:
> Is it poor form to deprecate something on a maint branch after the initial release?

Yeah I would say so. Let's leave it on maint, and deprecate on master.

jacobagilbert:
Ok makes sense. I will slightly rework this.

jacobagilbert:
Now print_pdu is marked for deprecation instead of removed. Warnings are printed and all core functionality lives in print() so whenever the deprecation window is up the print_pdu() function and handler can be cleanly removed.

mbr0wn (approved):

willcode:
Good, this confused me while I was trying to answer the ref'd question!

The yml file needs updating, too.

Minor: is there any reason to say "VERBOSE" in the print?

jacobagilbert:
@willcode I was not sure about the yml file. I originally had removed the `print pdu` port in there but was not sure if it needed to stay due to @mbr0wn's deprecation-before-removal guidance. Getting rid of it will break existing GRC sooner (which might be a good thing....) but in an obvious way while retaining API compatibility; happy to get rid of it, let me know.

And no, 'VERBOSE' is not necessary here IMO, however I intentionally left the printed syntax exactly as it was since that might break someone stdout parsing tools. It looks like I added this in the diff due to the code block that was moved.

mormj (approved):

PR #4226, opened 2021-02-14
digital: include control_loop class in child class bindings

jledet:
costas_loop_cc, fll_band_edge_cc, and constellation_receiver_cb all
inherit from gr::blocks::control_loop. Add the class to the bindings so
the control loop methods are available in Python.

Signed-off-by: Jeppe Ledet-Pedersen <jlp@satlab.com>


mbr0wn (approved):

mormj (approved):
Nice catch!

willcode:
Already in maint-3.9, commit 4f610dce209338e5fd7af38b692c7306a57ade9b

PR #4228, opened 2021-02-15
github: Add -Werror for CI builds Ubuntu 20.04, Debian 10, Centos 8.3

mbr0wn:
Also add -k to make so output is as useful as possible.


mbr0wn:
Merged manually.

PR #4230, opened 2021-02-15
github: ci: Enable qa_uhd test

mbr0wn:
This test was failing, but was fixed. Now re-enabling it in the CI
pipeline.


willcode:
uhd not diabled in make-test.yml, so no need for backport.

PR #4234, opened 2021-02-15
pybind: remove forecast from python bindings

mormj:
There is no scenario that a block or app should be calling forecast over
the python interface.  This is only called by the scheduler, and for
python blocks would pass through the gateway interface.

There may be more we want to remove from the pybind interface at the block level as well


willcode:
This (and other bindings) should be removed for the reason above, but it's not as much of an emergency as I thought it was yesterday. It seems you can override `forecast()` even with the binding in play. I need to learn more about how pybind and python subclassing (e.g., the gateway) interact.

mbr0wn (approved):

Issue #3920, opened 2020-11-07
Raise required Thrift version (on master) to 0.13.0


marcusmueller:
Problem: ctrlport-thrift doesn't build if we don't use the thrift smart pointer type inside our rpcmanager framework.

On maint-3.8, the solution looks like #2739; code is different for Thrift up to including 0.10 from newer version.

On master, we can avoid all the required version detection by bumping dependency version requirements to 0.13.0, which is available: (2020-11-06)

- fedora 31 (close to EOL): 0.10.0 
- fedora 32: 0.13.0
- CentOS8/EPEL: 0.13.0
- Ubuntu 18.04: it's getting colder, someone is writing thrift 0.9.1 on the  walls of their cave
- Ubuntu 20.04: 0.13.0
- Debian testing: 0.13.0
- Debian unstable: 0.13.0
- Opensuse tumbleweed: 0.13.0
- (Gentoo: special)

So, we'd be not supporting stock Thrift on Ubuntu 18.04, which means the ctrlport-thrift component won't work there. I'm OK with that.

marcusmueller:
@ryanvolz @ZeroChaos- Will this pose a problem?

ryanvolz:
> @ryanvolz @ZeroChaos- Will this pose a problem?

This will be just fine with conda-forge, which has 0.13 and where I haven't compiled with thrift in a long time because of version incompatibilities.

maitbot:
See

RPC fixes to allow using Thrift-0.13 #3953 

This adds building with 0.13.0 on master, as well as updates to logging for rpcserver_thrift class.
Hopefully will work with Thrift 0.14.0 as well...

ZeroChaos-:
apologies I totally missed this, but yeah it's fine :-)

PR #4236, opened 2021-02-15
cmake: Thrift min version is set at top level, raise min Thrift to 0.13

mbr0wn:
- Modifies top-level CMake to set the min Thrift version
- Modifies FindThrift.cmake to use a version argument

Closes https://github.com/gnuradio/gnuradio/issues/3920


mormj (approved):

willcode:
Dependency note: Thrift 0.13 dependency starts at GR v3.10. Ubuntu 18.04 has 0.9.1. Pybombs recipe currently pull 0.10.0.

willcode:
Backported to 3.9, but with thrift version left at 0.9.2. Required for later PR (see above).

PR #4240, opened 2021-02-15
github: ci: Re-enable qa_uncaught_exception and qa_file_taps_loader

mbr0wn:
These were fixed and now pass.


marcusmueller:
Should these be 3.9 backported? Maybe even 3.8?

PR #4243, opened 2021-02-15
dtv/atsc: replace structs with streams and tags

mormj:
ATSC implementation is old, and I don't know the entire history of it. But it seems to be the only module that uses structs to pass data between blocks.  This causes the grc blocks to be very special purpose and only have byte vectors as the interface (which many are floats).

The other reasoning behind this is also trying to keep packet data in contiguous memory so that hardware accelerators can be used more easily on larger buffer sizes

### DIFF
--- a/gr-dtv/include/gnuradio/dtv/atsc_plinfo.h
+++ b/gr-dtv/include/gnuradio/dtv/atsc_plinfo.h
@@ -1,0 +1,119 @@
+/* -*- c++ -*- */
+/*
+ * Copyright 2001,2006,2014 Free Software Foundation, Inc.
+ *
+ * This file is part of GNU Radio
+ *
+ * SPDX-License-Identifier: GPL-3.0-or-later
+ *
+ */
+
+#ifndef DTV_INCLUDED_ATSC_PLINFO_H
+#define DTV_INCLUDED_ATSC_PLINFO_H
+
+#include <gnuradio/dtv/atsc_consts.h>
+#include <boost/endian/conversion.hpp>
+#include <cassert>
+#include <cstring>
+
+#include <gnuradio/dtv/api.h>
+
+namespace gr {
+namespace dtv {
+
+/*!
+ * \brief pipeline info that flows with data
+ *
+ * Not all modules need all the info
+ */
+class DTV_API plinfo
+{
+public:
+    plinfo();
+    plinfo(uint16_t flags, int16_t segno);
+
+    /**
+     * @brief Resets the flags and segno fields
+     *
+     */
+    void reset();
+
+    /**
+     * @brief Load the flags and segno from an endian safe value that came over the tags
+     *
+     * @param tag_value
+     */
+    void from_tag_value(uint32_t tag_value);
+
+    /**
+     * @brief Return an endian safe value containing the flags and segno
+     *
+     * @return uint32_t
+     */
+    uint32_t get_tag_value() const;
+
+    // accessors
+    bool field_sync1_p() const;
+    bool field_sync2_p() const;
+    bool field_sync_p() const;
+
+    bool regular_seg_p() const;
+
+    bool in_field1_p() const;
+    bool in_field2_p() const;
+
+    bool first_regular_seg_p() const;
+
+    bool transport_error_p() const;
+
+    unsigned int segno() const;
+    unsigned int flags() const;
+
+    // setters
+
+    void set_field_sync1();
+
+    void set_field_sync2();
+
+    void set_regular_seg(bool field2, int segno);
+
+    void set_transport_error(bool error);
+
+    /*!
+     * Set \p OUT such that it reflects a \p NSEGS_OF_DELAY
+     * pipeline delay from \p IN.
+     */
+    static void delay(plinfo& out, const plinfo& in, int nsegs_of_delay);
+
+    // these three are mutually exclusive
+    //     This is a regular data segment.
+    static constexpr int fl_regular_seg = 0x0001;
+    //	 This is a field sync segment, for 1st half of a field.
+    static constexpr int fl_field_sync1 = 0x0002;
+    //	 This is a field sync segment, for 2nd half of a field.
+    static constexpr int fl_field_sync2 = 0x0004;
+
+    // This bit is on ONLY when fl_regular_seg is set AND when this is
+    // the first regular data segment AFTER a field sync segment.  This
+    // segment causes various processing modules to reset.
+    static constexpr int fl_first_regular_seg = 0x0008;
+
+    // which field are we in?
+    static constexpr int fl_field2 = 0x0010; // else field 1
+
+    // This bit is set when Reed-Solomon decoding detects an error that it
+    // can't correct.  Note that other error detection (e.g. Viterbi) do not
+    // set it, since Reed-Solomon will correct many of those.  This bit is
+    // then copied into the final Transport Stream packet so that MPEG
+    // software can see that the 188-byte data segment has been corrupted.
+    static constexpr int fl_transport_error = 0x0020;
+
+private:
+    uint16_t d_flags = 0; // bitmask
+    int16_t d_segno = 0;  // segment number [-1,311] -1 is the field sync segment
+};
+
+} /* namespace dtv */
+} /* namespace gr */
+
+#endif /* _ATSC_PLINFO_H_ */
--- a/gr-dtv/lib/atsc/atsc_deinterleaver_impl.cc
+++ b/gr-dtv/lib/atsc/atsc_deinterleaver_impl.cc
@@ -25,9 +25,10 @@ atsc_deinterleaver::sptr atsc_deinterleaver::make()
 }
 
 atsc_deinterleaver_impl::atsc_deinterleaver_impl()
-    : gr::sync_block("atsc_deinterleaver",
-                     io_signature::make(1, 1, sizeof(atsc_mpeg_packet_rs_encoded)),
-                     io_signature::make(1, 1, sizeof(atsc_mpeg_packet_rs_encoded))),
+    : gr::sync_block(
+          "atsc_deinterleaver",
+          io_signature::make(1, 1, ATSC_MPEG_RS_ENCODED_LENGTH * sizeof(uint8_t)),
+          io_signature::make(1, 1, ATSC_MPEG_RS_ENCODED_LENGTH * sizeof(uint8_t))),
       alignment_fifo(156)
 {
     m_fifo.reserve(s_interleavers);
@@ -36,6 +37,8 @@ atsc_deinterleaver_impl::atsc_deinterleaver_impl()
         m_fifo.emplace_back((s_interleavers - 1 - i) * 4);
 
     sync();
+
+    set_tag_propagation_policy(TPP_CUSTOM);
 }
 
 atsc_deinterleaver_impl::~atsc_deinterleaver_impl() {}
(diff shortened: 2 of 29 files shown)


gr-dtv/include/gnuradio/dtv/atsc_plinfo.h
```diff
@@ -0,0 +1,151 @@
...
+ */
+class plinfo
+{
+public:
+    plinfo() : _flags(0), _segno(0) {}
+    // plinfo(uint16_t flags_, int16_t segno_) : _flags(flags_), _segno(segno_) {}
+
+    // accessors
+
+    bool field_sync1_p() const { return (_flags & fl_field_sync1) != 0; }
```
ThomasHabets, line 34: Since this class is now part of the public API (and even ABI) it could make sense to clean it up a bit.

E.g.:
* Add `noexcept` where appropriate
* change state machine (?) to `enum class`
* use member initialization
* remove `operator==/!=` (they're default, no?
* Make member variables private?
* Put implementation in a .cc in `lib/`

What do you think?

gr-dtv/lib/atsc/atsc_equalizer_impl.cc
```diff
@@ -148,20 +163,39 @@ int atsc_equalizer_impl::general_work(int noutput_items,
...
             filterN(data_mem, data_mem2, ATSC_DATA_SEGMENT_LENGTH);
 
-            memcpy(out[output_produced].data,
+            memcpy(&out[output_produced * ATSC_DATA_SEGMENT_LENGTH],
                    data_mem2,
                    ATSC_DATA_SEGMENT_LENGTH * sizeof(float));
 
-            out[output_produced].pli._flags = d_flags;
-            out[output_produced].pli._segno = d_segno;
+            // out[output_produced].pli._flags = d_flags;
```
ThomasHabets, line 170: commented out code?

gr-dtv/lib/atsc/atsc_equalizer_impl.cc
```diff
@@ -148,20 +163,39 @@ int atsc_equalizer_impl::general_work(int noutput_items,
...
+            // out[output_produced].pli._segno = d_segno;
+            plinfo pli_out;
+            pli_out._flags = d_flags;
```
ThomasHabets, line 173: What about endianness if using multiple hosts, or merely replaying a capture file?

ThomasHabets (changes requested):
It is weird that ATSC sends serialized structs, yes. But maybe it's worth breaking out the data portions of the tag into a just-the-data struct? (endian safe)

gr-dtv/lib/atsc/atsc_fs_checker_impl.cc
```diff
@@ -59,17 +59,20 @@ int atsc_fs_checker_impl::general_work(int noutput_items,
                                        gr_vector_const_void_star& input_items,
                                        gr_vector_void_star& output_items)
 {
-    const atsc_soft_data_segment* in = (const atsc_soft_data_segment*)input_items[0];
-    atsc_soft_data_segment* out = (atsc_soft_data_segment*)output_items[0];
+    const float* in = reinterpret_cast<const float*>(input_items[0]);
```
ThomasHabets, line 62: Is `static_cast<>` not enough?

gr-dtv/lib/atsc/atsc_sync_impl.cc
```diff
@@ -82,15 +82,15 @@ int atsc_sync_impl::general_work(int noutput_items,
                                  gr_vector_const_void_star& input_items,
                                  gr_vector_void_star& output_items)
 {
-    const float* in = (const float*)input_items[0];
-    atsc_soft_data_segment* soft_data_segment_out =
-        (atsc_soft_data_segment*)output_items[0];
+    const float* in = reinterpret_cast<const float*>(input_items[0]);
```
ThomasHabets, line 85: same

gr-dtv/lib/atsc/atsc_sync_impl.cc
```diff
@@ -180,8 +180,13 @@ int atsc_sync_impl::general_work(int noutput_items,
             d_data_mem[d_symbol_index] = interp_sample;
 
             if (d_symbol_index >= (ATSC_DATA_SEGMENT_LENGTH - 1)) {
-                for (int i = 0; i < ATSC_DATA_SEGMENT_LENGTH; i++)
-                    soft_data_segment_out[d_output_produced].data[i] = d_data_mem[i];
+                // for (int i = 0; i < ATSC_DATA_SEGMENT_LENGTH; i++)
```
ThomasHabets, line 183: Should this commented-out code remain?

gr-dtv/include/gnuradio/dtv/atsc_plinfo.h
```diff
@@ -0,0 +1,157 @@
...
+namespace gr {
+namespace dtv {
+
+/*!
+ * \brief pipeline info that flows with data
+ *
+ * Not all modules need all the info
+ */
+
+struct plinfo_data {
```
ThomasHabets, line 27: so this is a new one, but it's not used by `plinfo`? I'm not sure I follow.

ThomasHabets (changes requested):

gr-dtv/include/gnuradio/dtv/atsc_plinfo.h
mormj: The idea here was to break out the data portion of the class into a struct.  It is only this portion that gets sent along on the tags now.

gr-dtv/lib/atsc/atsc_equalizer_impl.cc
mormj: I'm not sure I follow - is this in reference to a struct being sent over std::any?

gr-dtv/include/gnuradio/dtv/atsc_plinfo.h
mormj: Agreed - I have more work to clean up this class

gr-dtv/lib/atsc/atsc_equalizer_impl.cc
ThomasHabets: No. The binary representation of the tag value is not portable across little/big endian. So saving a stream on a big endian machine it won't be correctly readable on a little endian machine unless they're encoded somehow. Either `htonl()` and friends, or some other serialization.

As I see it now the values are just used in the data stream as whatever the native endianness is.

willcode:
If atsc_plinfo.h is in include/gnuradio/dtv, it (and the functions in it) become part of the API. Is that the intent? This file could be split into public/impl or just moved to lib if not.

mormj:
Yes, the intent is that atsc_plinfo.h public, since a user could break off from the receive chain and insert, say, hardware accelerated receive blocks that are in an OOT that would need access to the struct in  atsc_plinfo.h to decode tags

gr-dtv/include/gnuradio/dtv/atsc_plinfo.h
mormj: Ok, I've done the things here except for the `noexcept` - would like to see this applied consistently through the codebase and it is currently mainly used for `noexcept(false)`

gr-dtv/lib/atsc/atsc_equalizer_impl.cc
mormj: Sending a big endian packed U32 with the two values over the tags now.

gr-dtv/lib/atsc/atsc_deinterleaver_impl.cc
```diff
@@ -52,24 +55,37 @@ int atsc_deinterleaver_impl::work(int noutput_items,
...
+    auto tag_pmt = pmt::intern("plinfo");
     for (int i = 0; i < noutput_items; i++) {
-        assert(in[i].pli.regular_seg_p());
-        plinfo::sanity_check(in[i].pli);
+        plinfo pli_in;
+        get_tags_in_window(tags, 0, i, i + 1, tag_pmt);
+        if (tags.size() > 0) {
+            pli_in.from_tag_value(pmt::to_uint64(tags[0].value));
+        } else {
+            throw std::runtime_error("No plinfo on tag");
```
ThomasHabets, line 69: On "no plinfo tag on sample", you mean?

gr-dtv/lib/atsc/atsc_plinfo.cc
```diff
@@ -0,0 +1,109 @@
...
+
+void plinfo::reset()
+{
+    _flags = 0;
+    _segno = 0;
+}
+
+void plinfo::from_tag_value(uint32_t tag_value)
+{
+    auto native_val = boost::endian::big_to_native(tag_value);
```
ThomasHabets, line 27: Why not `htonl()`, to avoid dependency on boost?

gr-dtv/include/gnuradio/dtv/atsc_plinfo.h
```diff
@@ -0,0 +1,117 @@
...
+
+    // This bit is set when Reed-Solomon decoding detects an error that it
+    // can't correct.  Note that other error detection (e.g. Viterbi) do not
+    // set it, since Reed-Solomon will correct many of those.  This bit is
+    // then copied into the final Transport Stream packet so that MPEG
+    // software can see that the 188-byte data segment has been corrupted.
+    static constexpr int fl_transport_error = 0x0020;
+
+private:
+    uint16_t _flags = 0; // bitmask
```
ThomasHabets, line 110: With it being moved to public headers, is this a good time to make it conform to naming standard with `d_flags`?

ThomasHabets (approved):

gr-dtv/lib/atsc/atsc_plinfo.cc
mormj: We already have the dependency on boost::endian via pmt, which is already cross platform.  htonl is very reasonable but require the inclusion of different headers depending on Linux/Windows/Mac, and hasn't yet been used in the codebase.

gr-dtv/include/gnuradio/dtv/atsc_plinfo.h
mormj: good point.

gr-dtv/include/gnuradio/dtv/atsc_plinfo.h
```diff
@@ -0,0 +1,117 @@
...
+     * @param tag_value
+     */
+    void from_tag_value(uint32_t tag_value);
+
+    /**
+     * @brief Return an endian safe value containing the flags and segno
+     *
+     * @return uint32_t
+     */
+    uint32_t get_tag_value();
```
ThomasHabets, line 51: `const`

ThomasHabets (approved):

ThomasHabets (approved):

drmpeg (approved):
Tested live over the air. Looks good! Maybe takes slightly more CPU cycles, but nothing to worry about.

mormj:
Doesn't make a functional difference, but added the `DTV_API` tag to the plinfo class for consistency (cross-platform ramifications?)

gr-dtv/include/gnuradio/dtv/atsc_plinfo.h
mbr0wn: There's something with `noexcept` and new C++ standards. Something changed -- @ThomasHabets would know best.

ThomasHabets: `noexcept(false)` is the default. I'm not sure if you're thinking of the now deprecated `throw()` spec. Except when writing templates that can instantiatie to `noexcept` or not depending on template args, there's no need for parentheses.

`throw()` was always a mess, unfortunately. `noexcept` is fine though. `throw()` ended up being only used for empty throw specs.

In my testing `noexcept` reduces code size. I've not benchmarked but it can enable other optimizations beyond reduced instruction cache use.

But the main reason I suggested it is that it's a nice annotation to the reader that this function cannot even in principle throw. I'm pretty sure it's not part of the ABI, but for public APIs better safe than sorry so should only be added when no possible future implementation could throw, in my opinion.

Of course in this case it'll in practice be inlined anyway, and the compiler will not generate exception handling code on entry or exit.

mormj: I definitely trust @ThomasHabets on this one.  Pushed a commit that puts `noexcept` _everywhere_ in the plinfo class

mormj: On second thought, I don't think we're ready to move in the direction of putting `noexcept` in lots of places, and probably need a consistent strategy for exceptions and exception handling.  Backing out the last commit to keep this file consistent with the rest of the codebase

ThomasHabets: Thinking first and rolling out a consistent use based on clear criteria sounds best, I agree.

mbr0wn:
CentOS failure is unrelated.

drmpeg:
I'm afraid this version came out pretty glitchy. Somethings not right somewhere, but it's going to take a while for me to track down the offending block(s).

mormj:
Darn.  Is it glitchy in terms of decoded video or in some other way?
Nothing other than a squash and the DTV_API attribute changed.

drmpeg:
Yes, glitchy decoded video. Looks like the ATSC Receiver SYNC block is having problems. I'm uploading a test file right now, but I only have 12 Mbps upstream on my cable Internet. I'll let you know when it's ready.

I'm not sure why I didn't catch this. When I tested before, I did see some glitches at one point, and then they went away. Maybe I got fooled with the way I was switching between versions. I had two builds and was just using ```sudo make install`` to switch between the two. Today, I have one build, and I'm switching commits and rebuilding (and installing).

mormj:
Thanks for digging in - Looks like I've got some debugging to do.  If I can't figure out we'll obviously have to back out this commit.

drmpeg:
Doh! I just realized I can't compare block outputs between the two versions without taking into account the new structures. So it may not be the ATSC Receiver SYNC block.

drmpeg:
Yup, ATSC Receiver SYNC block is okay when the header and padding on the old receiver is discarded.

drmpeg:
Okay, I'm pretty sure it's the ATSC Viterbi Decoder block. It fails pretty far into the test bitstream after 690,552 output bytes, which is exactly 3,336 207-byte packets.

Then it recovers after byte 693,035, which is 2,484 bytes or 12 207-byte packets later.

Then it fails again after 1,428,300 bytes or 6,900 207-byte packets (and recovers after 2484 bytes).

1,614,600 or 7,800 packets
3,174,552 or 15,336 packets
4,192,992 or 20,256 packets
4,523,364 or 21,852 packets
5,412,636 or 26,148 packets
8,793,360 or 42,480 packets

The first failing block had all zeroes in it, so it seems that 12 207-byte packets don't get written out after some corner case.

drmpeg:
The test bitstream is ready. It's 5 minutes, so it's pretty ginormous. 28,451,236,144 bytes.

http://www.w6rz.net/atscsator.cfile

New test flow graph:

image

Old test flow graph:

image

```cmp -l atscvitold.bin atscvitnew.bin | more```

drmpeg:
Although the ATSC decoder is deterministic, the bug isn't. You'll see different error offsets than what I posted above and different offsets from run to run.

mormj:
@drmpeg - thank you for setting up the debug so well.  I can't seem to find any arithmetic errors in the inputs to the viterbi decoder or copying the outputs just yet - most of the code is the same apart from the plinfo and copying into the various structs.  

One thing I do notice is that the plinfo coming into the viterbi decoder is off by one sample which could cause issues - not sure whether this would cause non-deterministic issues

Now, I'm wondering though whether it would be better to pass the plinfo as a separate stream of data rather than tags - since it is always one per sample.

mormj:
Nope, it is not the plinfo - eliminated that variable - strictly something happening in the viterbi decoder :/

PR #4244, opened 2021-02-16
gnuradio-runtime: configurable buffer size

karel:
Currently block output buffers are limited to 32768 bytes in size, although smaller buffers can be configured using `set_max_output_buffer` functionality. This patch allows increasing the 32768 byte limitation from the configuration file. Increasing the buffer size is useful in rare cases and this patch greatly simplifies doing so. Or at least it simplifies experimenting with different buffer sizes.

https://github.com/gnuradio/gnuradio/blob/63cd1cd0570b387d5b0ccdfbbf92a093a6494fc1/gnuradio-runtime/lib/flat_flowgraph.cc#L28

This patch was originally proposed in [discuss-gnuradio](https://lists.gnu.org/archive/html/discuss-gnuradio/2016-02/msg00404.html) by @marcusmueller. I found it useful.

### DIFF
--- a/gnuradio-runtime/gnuradio-runtime.conf.in
+++ b/gnuradio-runtime/gnuradio-runtime.conf.in
@@ -9,6 +9,8 @@ verbose = False
 # the queue by popping messages from the front.
 max_messages = 8192
 
+# Block output buffer size in bytes.
+#buffer_size = 32768
 
 [LOG]
 # Levels can be (case insensitive):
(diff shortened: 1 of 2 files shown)


willcode (approved):
Looks good.

mbr0wn (approved):

marcusmueller (approved):

willcode:
In the past, the allocation was calculated in `flat_flowgraph.cc`, so that's where the setting was read. The logic, but not the setting, moved to `block.cc` when the buffer work was done. So, this looks right.

PR #4248, opened 2021-02-17
Add gr-soapy in tree

willcode:
gr-soapy ported to 3.9+ and brought in-tree

This is to start a discussion on how to bring SoapySDR support in-tree.

SoapySDR itself is used by other projects, so it does not belong in-tree. The gr-soapy OOT is from
https://gitlab.com/librespacefoundation/gr-soapy. According to the authors, TX has not been tested
extensively.

Questions:
- Is this the architecture we want, gr-soapy in-tree and SoapySDR as an external dependency?
- Do gr-soapy and SoapySDR support everything needed for the variety targeted SDRs?
- Is this code in good shape?

Note that to run this code, SoapySDR must already be installed. It is not configured as a submodule.


noc0lour:
I think this module would be a good extension to functionality we already have in tree. So far no device specific features are implemented in the module. For a very generic SDR use case this module will probably be enough for most users and being in tree missing functionality can be easily added.

willcode:
Working on this without a PR to avoid triggering CI on every update.
https://github.com/willcode/gnuradio/tree/feature/master/gr-soapy

willcode:
@acinonyx Are there direct users of the C++ interface, or is all use via GRC Python wrappers? I'm starting to work on factoring out sink/source into a base class, and might need to change some function names in the process.

PR #4257, opened 2021-02-17
runtime: top_block: Add precondition to Python version of wait()

mbr0wn:


marcusmueller (approved):
It is a bit of a hack (why not wait to call `wait()` in the test?), but it seems to be sensible functionality in the right place, and it doesn't actually cost anything.

marcusmueller:
(this, btw, would be the kind of change that'd make sense to have in the CHANGELOG)

mbr0wn:
lulz-- I must have fat-fingered this PR. I wasn't going to put it up yet. But glad you still approve :)

mbr0wn:
> It is a bit of a hack (why not wait to call `wait()` in the test?), but it seems to be sensible functionality in the right place, and it doesn't actually cost anything.

BTW, I totally agree this is a hack. But in principle, I think it is the right idea to give the flowgraph some more info about what's going on externally.

mbr0wn:
Well I'll be damned -- this didn't actually fix the underlying issue. Huh?

willcode:
Is there a way to put this in gr_unittest.py? Or are you thinking that it might be of wider use?

A wrapper would be nice, just because wait() has a fairly well known meaning. Could it be wrapped or patched in just for qa?

mormj:
Yeah - I'm on the fence about this one because it creates a diversion between the python and c++ versions of `wait`

Wrap wait with this functionality as `wait_until`?

mbr0wn:
I'm actually having second thoughts about this entire approach. How about we close this and think about it. It didn't even fix the thing I was trying to fix (qa_agc).

PR #4258, opened 2021-02-18
uhd: Fix filter API

mbr0wn:
9cdfe5141a exposed a bug in gr-uhd: The filter API was incorrectly
implemented. However, the #ifdef to check for the existence of the
filter API also had a typo, and therefore, the filter API was never
compiled in.

This is an API change, b/c the existing gr-uhd code was actually never
compatible to UHD's filter API. This tells us no one's actually been
using it, so the API change is low-risk.

The change is to adapt gr-uhd's filter API to that in UHD.


mbr0wn:
OK this is wrong. The filter API changed between UHD 3 and 4.

willcode:
@mbr0wn Based on above comments, is this good for a backport to 3.9?

PR #4259, opened 2021-02-18
gr-uhd: Generate tags when frequency changed from message port.

### LINKED ISSUE #1357: UHD: USRP source block does not generate stream tags when RX frequency changed via its message port.
Generally, if the USRP source block's centre frequency is altered then the output stream is tagged with the current time, frequency, and rate. However, if the centre frequency is altered via messages received on its command port then these  tags are not generated. This seems like a bug.

The root cause is that command messages use the SET_CENTER_FREQ_FROM_INTERNALS macro to set the RX frequency instead of the set_center_freq() function. The macro does not set the _tag_now flag, so no tags are generated.

### LINKED ISSUE #4204: gr-uhd: Tags not sent on "tune" message
When tuning the USRP using the message port interface, the rx_freq, rx_time, rx_rate tags are not sent to the downstream blocks.  They are sent properly when calling usrp_source_impl::set_samp_rate(), usrp_source_impl::set_center_freq().

seldad8:
Call set_center_freq() rather _dev->set_rx_freq() so that _tag_now is set on the change.

The end result is that the tags are generated as needed when changes are directed from the message port. set_center_freq() performs the same functionality as the original code with the addition of properly setting the _tag_now flag.

Signed-off-by: Steve Lunsford <seldad8@gmail.com>

fixes #1357 
fixes #4204.

### DIFF
--- a/gr-uhd/lib/usrp_source_impl.cc
+++ b/gr-uhd/lib/usrp_source_impl.cc
@@ -118,7 +118,7 @@ usrp_source_impl::_set_center_freq_from_internals(size_t chan, pmt::pmt_t direct
         return _dev->set_tx_freq(_curr_tune_req[chan], _stream_args.channels[chan]);
     } else {
         return set_center_freq(_curr_tune_req[chan], chan);
-   }
+    }
 }
 
 double usrp_source_impl::get_center_freq(size_t chan)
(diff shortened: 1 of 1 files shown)


mbr0wn (approved):

willcode:
@mbr0wn Does this close the two issue above, and does the get ported to master/maint-3.9?

seldad8:
I haven't looked at the 3.9 code yet (I will) but I presume it will need to be ported there as well.

PR #4262, opened 2021-02-19
gr-blocks: remove indexing in file source cpp template

karel:
`"${file[1:-1]}"` results in the file name becoming `NotImplemented`. This seems to be the only place where such indexing is being tried. Otherwise, `${file}` is used everywhere (file sink etc). This patch makes the code across all such blocks consistent and fixes the `NotImplemented` issue for file source block (when generating C++ output).

However, the original issue, which I believe was being targeted by indexing and double quotes, is still not solved. When simply using `${file}`, the file name is passed to C++ code in single quotes, which results in an error. A solution to that is to enclose the file name in double quotes in the respective GRC block, for example.

### DIFF
--- a/gr-blocks/grc/blocks_file_source.block.yml
+++ b/gr-blocks/grc/blocks_file_source.block.yml
@@ -60,7 +60,7 @@ templates:
 cpp_templates:
     includes: ['#include <gnuradio/blocks/file_source.h>']
     declarations: 'blocks::file_source::sptr ${id};'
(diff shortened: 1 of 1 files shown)


mormj:
@karel - so it seems this allows the c++ generation to be usable, but only if the input is passed with double quotes?
Even if that is the case, seems like this is an important fix.  Any ideas how to make the input from GRC input more robust on cpp generation either in the yml or in the parameter evaluation handlers in GRC?

haakov (approved):
Great! Good to have consistency. I can make a new PR with a suggestion for improving the input robustness

mormj (approved):

PR #4263, opened 2021-02-19
grc: remove gnuradio prefix from linked libraries

karel:
This allows out-of-tree modules to be linked, as those will not link with the prefix. In-tree modules link still. Or are there some cases where the in-tree modules will not link with this approach?

### DIFF
--- a/grc/core/generator/cpp_templates/CMakeLists.txt.mako
+++ b/grc/core/generator/cpp_templates/CMakeLists.txt.mako
@@ -55,7 +55,7 @@ target_link_libraries(${class_name}
     gnuradio::gnuradio-blocks
     % for link in links:
     % if link:
-    ${link.replace("gnuradio-", "gnuradio::gnuradio-")}
+    ${link}
     % endif
     % endfor
 


haakov:
This change gives me the following error when compiling a `Null source ----> Null sink` flowgraph

```
-- Configuring done
-- Generating done
-- Build files have been written to: /home/hkon/grc/test_cpp_linking/build
Scanning dependencies of target test_cpp_linking
[ 50%] Building CXX object CMakeFiles/test_cpp_linking.dir/test_cpp_linking.cpp.o
[100%] Linking CXX executable test_cpp_linking
/usr/bin/ld: CMakeFiles/test_cpp_linking.dir/test_cpp_linking.cpp.o: undefined reference to symbol '_ZN2gr9top_block5startEi'
/usr/bin/ld: /usr/local/lib/libgnuradio-runtime.so.3.10.0git: error adding symbols: DSO missing from command line
collect2: error: ld returned 1 exit status
make[2]: *** [CMakeFiles/test_cpp_linking.dir/build.make:103: test_cpp_linking] Error 1
make[1]: *** [CMakeFiles/Makefile2:95: CMakeFiles/test_cpp_linking.dir/all] Error 2
make: *** [Makefile:103: all] Error 2
```

Maybe the gnuradio:: prefix should be moved to the YAML files instead?

karel:
Oddly `Null source ----> Null sink` flowgraph compiles/runs on my system. But I like the idea of moving this prefix to the YAML files. This would allow to replicate current situation for in-tree modules (i.e., nothing should break) and allow using out-of-tree modules as well.

mormj:
@karel  - what do you recommend we do here?  Close this PR and open one that moves prefix into the YAML?

karel:
@mormj, closing this PR and moving the prefix to YAML files seems to be the best course of action. I'll do that.

karel:
After discussing the PR with @haakov, we found that the initial null sink/source test was somewhat misleading (null sink/source, i.e. gr-blocks, do not rely on this patch). It seems that this patch is actually suitable.

haakov (approved):

PR #4265, opened 2021-02-19
Search bar

moesay:
1- Allowing the search entry to be visible by default because users usually use it and nobody navigates through the blocks to get a specific one.

2- Auto clearing the previous search text when `search_entry` gains focus because the users usually delete the previous search text to search for something else. This makes use of the implemented interactive searching mechanism so the user 100% won't make a typo in the block name because he can usually and on-spot edit it and see the searching results changing.


willcode:
I like this idea. Will give it a try.

You can get rid of the corresponding `search_entry.show()` if it will never be hidden.

willcode:
A couple of suggestions. If the search field is there by default, we could also get rid of the (x) in the corner that closes the search box. As you say, it's used enough that I don't think people would mind it being there all the time.

The behavior of deleting text when you click back in the box is probably not what a user would expect. Instead, the (x) in the corner could delete the text, allowing the use the option to do a new search or edit the last one.

moesay:
> A couple of suggestions. If the search field is there by default, we could also get rid of the (x) in the corner that closes the search box. As you say, it's used enough that I don't think people would mind it being there all the time.
> 
> The behavior of deleting text when you click back in the box is probably not what a user would expect. Instead, the (x) in the corner could delete the text, allowing the user the option to do a new search or edit the last one.

The search entry was hidden by default and you have to click on the search icon to show it and if you'd like, you can hide it again by clicking on the X icon. Now everything is the same, you can hide and reshow it if you wanna, but the default is it being shown.

Deleting the previous search text is kind of automating the deletion because at least, 99% of the time you will delete it. For example, if you are searching for a signal source you would simply search with "source" (or at least i do so). Then if i wanna add a noise source, the "source" search result is still there. i can add a noise source in no time. Deleting the text makes no sense if the user searches with the full block name each time and i don't think this is the case.

mormj:
I like the idea as well here and thank you for the contribution - seems like a reasonable effort to improve the UX for the average user.  Need to get more feedback on this before we change the default behavior of GRC.

mormj:
@moesay - I tend to agree with @willcode here that leaving the search bar by default is a good design, but auto-clearing it could cause some confusion and frustration.

jacobagilbert:
I like the idea of having the search bar there by default, but I am not sure auto-delete is the correct behavior. Maybe auto-select-all?

ei8fdb:
Hi there!

I'd like to give some feedback on this PR.

# Make block search field visible by default

Yup 100% agree with this suggestion. It is possible that a new user would navigate through the blocks initially to see what's there, however being able to see the search field visually as opposed to just keyboard shortcut (which is hidden in the view menu) should also be supported.

# Auto-clear the search box when the user moves focus from it

I don't agree with this change. Automatically clearing the search box when it looses focus is not a standard interaction pattern.

> Deleting the previous search text is kind of automating the deletion because at least, 99% of the time you will delete it. 

You can't work out what that "99% of the time" is. This is only perceivable to the user themselves.

 A user will not consider "clicking somewhere else on the screen" as an intentional action. (Random clicking on a screen is a very common user behaviour...especially when they are thinking, trying to work something out, or just distracted).

I agree with @willcode's [suggestion above](https://github.com/gnuradio/gnuradio/pull/4265#issuecomment-782208965) on how to deal with clearing the search term:

> Instead, the (x) in the corner could delete the text, allowing the user the option to do a new search or edit the last one.

This is how it currently works. Here's a quick screenshot:
Peek 2021-03-08 18-25

## To sum up! 
1. Yes to making search field visible by default
2. No to auto-clearing the search field search terms when a user takes focus off the search field

Thoughts, criticism welcome.

mormj:
@moesay - seems the consensus is the first change is positive, but the second one not recommended - are you able to update the PR to remove the second change?

duggabe:
1) I think the 'show by default' is good. We could keep the hide choice if anyone really cared.
2) I think that clicking the 'x' on the right side should clear the entry, but **not** hide the box.

moesay:
Now I'm convinced that auto-clearing the search entry is not as good as I thought. I'm also convinced that better behavior would be to select the text, not clear it. And the 'X' icon clears the text. I will work on that. Thanks for the reviews.

mormj (changes requested):
@moesay - thank you for updating the PR.  I like the always present search bar.  The auto-select behaves a bit forcefully, though.  If I click in there, it autoselects everything, but if I click again, it keeps autoselecting - the only way to not select is then use the arrow keys.  The behavior that might be a tad nicer is if when you click, it autoselects, but when you click again the cursor goes where you've clicked, unselected.

Always autoselecting seems a bit draconian - can you please either update to unselect, or remove the autoselect part of this?

willcode:
Double-click is fine for select all, single click for position cursor. Too bad there are many standards for this (broswer URL, text, Windows file rename ...). The main use for select all is to delete, and the x can do that.

moesay:
I've updated the PR

1. The Text selection/clearing feature is dropped. I'm convinced that it's not necessary.
2. The search button now toggles the search entry. Its shown by default but you can hide it.
3. Fixing a bug with the old behavior of the search button when the block tree panel is hidden and the button is pressed. If the block tree has been hidden using the menu option `View > Show Block Panel` And shown using the search button, The menu option is never updated.

mormj:
Thank you so much. Would you mind squashing your commits and this should be good to merge.

mormj:
also, to keep the commit history consistent, can you please prefix the commit message with the module, e.g.
```
grc: make search visible by default
```

moesay:
done, I think. any comments?

grc/gui/Actions.py
```diff
@@ -646,10 +646,10 @@ def connect(action, handler=None):
                                  label='Reload _Blocks',
                                  tooltip='Reload Blocks',
                                  icon_name='view-refresh')
-FIND_BLOCKS = actions.register(
-    "win.find_blocks",
-    label='_Find Blocks',
-    tooltip='Search for a block by name (and key)',
+TOGGLE_SEARCH = actions.register(
+    "win.toggle_search",
+    label='_Toggle Search',
```
mormj, line 651: In the menu, this is not particularly clear.  Maybe "Show Block Search" and have it be a checkbox - would like some feedback from others on this

mormj:
@moesay - I'm going to close this for now since it has gone quiet.  I like the search bar being there by default, but the verbage in the menu item to show/hide it is inconsistent with other options.  If you would like to solicit more feedback on the ideas for this and other things that affect GRC, please coordinate on chat.gnuradio.org in the GRC Development or Development rooms.  

Just closing this for now, you can always re-open when you have more changes to submit.  Thanks!

PR #4266, opened 2021-02-19
qtgui: Remove some manual memory management

ThomasHabets:
There's more to do in qtgui, but this PR will be big enough already.

### DIFF
--- a/gr-qtgui/lib/eye_sink_c_impl.h
+++ b/gr-qtgui/lib/eye_sink_c_impl.h
@@ -15,6 +15,8 @@
 #include <gnuradio/high_res_timer.h>
 #include <gnuradio/qtgui/eyedisplayform.h>
 
+#include <volk/volk_alloc.hh>
+
 namespace gr {
 namespace qtgui {
 
@@ -30,14 +32,19 @@ private:
     const pmt::pmt_t d_tag_key;
 
     int d_index, d_start, d_end;
-    std::vector<gr_complex*> d_cbuffers;
-    std::vector<double*> d_buffers;
+    std::vector<volk::vector<gr_complex>> d_cbuffers;
+    std::vector<volk::vector<double>> d_buffers;
     std::vector<std::vector<gr::tag_t>> d_tags;
 
-    int d_argc;
-    char* d_argv;
+    // Required now for Qt; argc must be greater than 0 and argv
+    // must have at least one valid character. Must be valid through
+    // life of the qApplication:
+    // http://harmattan-dev.nokia.com/docs/library/html/qt4/qapplication.html
+    char d_zero = 0;
+    int d_argc = 1;
+    char* d_argv = &d_zero;
     QWidget* d_parent;
-    EyeDisplayForm* d_main_gui;
+    EyeDisplayForm* d_main_gui = nullptr;
 
     gr::high_res_timer_type d_update_time;
     gr::high_res_timer_type d_last_time;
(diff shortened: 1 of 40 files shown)


ThomasHabets:
The tests pass, I had another read-through, and I've tested the QT GUI by playing around. So seems like it's good and ready to be reviewed.

mbr0wn (approved):

mormj (approved):
@ThomasHabets - thanks as always for the modernization.  The changes look good and pass tests, and are now consistent with all the other modules you fixed up!

gr-qtgui/lib/eye_sink_c_impl.h
```diff
@@ -70,6 +77,12 @@ class QTGUI_API eye_sink_c_impl : public eye_sink_c
                     QWidget* parent = NULL);
     ~eye_sink_c_impl() override;
 
+    // Disallow copy/move because of the raw pointers.
+    eye_sink_c_impl(const eye_sink_c_impl&) = delete;
+    eye_sink_c_impl& operator=(const eye_sink_c_impl&) = delete;
+    eye_sink_c_impl(eye_sink_c_impl&&) = delete;
+    eye_sink_c_impl& operator=(eye_sink_c_impl&&) = delete;
```
jdemel, line 84: The raw pointer you're referring to is the `QWidget* parent`, correct? Are there any others?

gr-qtgui/lib/sink_c_impl.cc
```diff
@@ -187,41 +163,41 @@ void sink_c_impl::set_frequency_range(const double centerfreq, const double band
...
+  d_main_gui.setConstellationAxis(xmin, xmax, ymin, ymax);
 }
 
 void
 sink_c_impl::set_constellation_pen_size(int size)
 {
-  d_main_gui->setConstellationPenSize(size);
+  d_main_gui.setConstellationPenSize(size);
 }
 */
```
jdemel, line 195: Would it make sense to remove this section? For now, it's dead code.

jdemel (commented):
LGTM. I have 2 questions but besides, I'd approve this PR.

Regarding the raw pointer business. Would it make sense to think about how to remove it?

gr-qtgui/lib/eye_sink_c_impl.h
ThomasHabets: Also `d_argv` and `d_main_gui`.

It could be that it's actually safe to copy `d_main_gui` and `d_argv`, as they are not destroyed locally (I assume QT owns destroying them?). `d_argv` may have to be solved some other way, though.

gr-qtgui/lib/sink_c_impl.cc
ThomasHabets: You mean because they're commented out in the public header file?

I would prefer in a separate commit, but either works for me.

jdemel: A separate commit would be reasonable. I just stumbled over. Obviously, this could is prone to become stale. I'd say, either we use it or remove it.

ThomasHabets: Added a commit deleting this commented out code.

mormj:
@ThomasHabets - it appears that all the comments were addressed and it is ready to merge - would you mind signing off that last commit so that it passes the DCO check?

ThomasHabets:
Sorry, not used to that workflow, and missed the check. Done now.

Issue #3371, opened 2020-04-15
GUI layout conflict


duggabe:
With GRC 3.8.1.0 if a flowgraph has two GUI Range blocks, where the first one has no GUI hint and the second one has `gui_hint: 0,0,3,1` then the result is that the first one is displayed but does not operate. In the example given, the `Volume` does not work, but the `Frequency` does. No error messages are generated.
OS Ubuntu 19.10
also OS Raspbian Buster 2020-02-13
GR 3.8.1.0
GUI_issue.grc: https://pastebin.com/07djpQxQ
GUI_issue.py: https://pastebin.com/CSwmpftB

duggabe:
Another example of the problem is in https://github.com/gnuradio/gnuradio/blob/master/gr-qtgui/examples/qtgui_tags_viewing.grc
The Delay range GUI can be seen behind the Time Sink.

ChristopheSeguinot:
I investigated this issue. Finally found it's not related to  #4249
The problem arises from using some widget without GUI Hint, and some with GUI Hint. Widgets are inserted in a QGridLayout. Examining flowgaph generated .py file we for example get

`self.top_grid_layout = Qt.QGridLayout()
self.top_layout.addLayout(self.top_grid_layout)
...  
self.top_grid_layout.addWidget(self._volume_win)
...   
self.top_grid_layout.addWidget(self._freq_win, 0, 0, 1, 1)
`

In case some widget without GUI Hint is added first, it use (row,col) =0,0 . 
In the present code it is superposed to the widget using 0,0

I found that this can be caused by any widget, not only QTGUI Range. 
Possible solutions: 

1. all widgets must have a GUI Hint , OR all must have no GUI Hint : rejected, this is not convenient
2. all widgets without GUI Hint should be added after those having a GUI Hint (tested by modifying generated python : it works

Looking into grc/core/utils/expr_utils.py and /usr/lib/python3/dist-packages/gnuradio/gr/top_block.py I found that blocks are ordered by name, and by GUI HInt. 
We could modify this so that widgets without GUI Hint are processed after all other. 

However, it seem's that this odering is not applied before generating the flowgraph.

Questions:
is it possible to apply ordering before generating the flowgraph ?
is it possible to modifiy ordering method _get_block_sort_text whithout breaking GNURadio ? (or do we need to define  new sorting method ? )

Could someone give me some advice to fix this issue?

duggabe:
Is this ready to close now?

[...]

PR #4270, opened 2021-02-20
grc: order blocks with GUI Hint first

ChristopheSeguinot:

This PR add ordering of block in a flowgraph before the .py code is generated.
This prevent PyQT5 from superposing blocks as explained in issue #3371, by simpling drawing first those blocks having a GUI Hint.

### DIFF
--- a/grc/core/generator/top_block.py
+++ b/grc/core/generator/top_block.py
@@ -196,11 +196,12 @@ class TopBlockGenerator(object):
         blocks = expr_utils.sort_objects(blocks, operator.attrgetter('name'), _get_block_sort_text)
 
         # Ordering blocks : blocks with GUI Hint must be processed first to avoid PyQT5 superposing blocks
-        for block in blocks:
+        def without_gui_hint(block):
             hint = block.params.get('gui_hint')
-            block.without_gui_hint = not hint # (hint == None) or (not hint.get_value())
+            Messages.send('>>> name =: {} without_gui_hint={} hint={}\n'.format(block.name, hint,type(hint)))
+            return hint is None or not hint.get_value()
 
-        blocks = sorted(blocks, key=lambda block: block.without_gui_hint)
+        blocks.sort(key=without_gui_hint)
 
         blocks_make = []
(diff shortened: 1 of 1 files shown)


ChristopheSeguinot:
Please **wait before merging .** 
I discovered one of my flowgraph which don't run after this PR. 
The error is that a widget is added on a layout which is not yet declared (this is related to block ordering used in my PR). 
I will investigate this and try to add a .grc file illustrating the problem (mine is too large , multiple variables sliders QT Frequency sink... ).

ChristopheSeguinot:
I'm sorry to confirm this PR is not fixing the issue and I will have to restart my homework!. 

For simple flowgraph without QT GUI Tab, it works. 
As soon as a QT GUI TAB widget without GUI HINT is used  , and some other widgets inside tabs (let's call them a second level widget) don't have GUI HINT too: it fails:

For example this structure will fail 
* main GUI _layout 
* variable 
* qt time sink 
* tab widget (without GUI_HINT)
  * qt freq sink (without GUI_HINT)
  * qt range
* tab widget2 (without GUI_HINT)
  * qt freq sink2 (without GUI_HINT)
  * qt range2
* variable_2

Of course, most users  avoid to have such bad flowgraph. However, suppose someone construct a flowgraph without any  GUI_HINT, and then after he decides to implement GUI_HINT one by one; he will certainly fall in this situation.

The trouble is that with the current PR the tab widget can be processed after the widget it contains ! I think we should also address  possible to have nested tab widget inside a tab widget.

The solution (if we really want to fix the issue in any possible flowgraph) would be to work level by level and branch by branch (tab widget and tabwidget 2 are 2 branches) and apply the new sorting individually at each branch (this looks like a recursive process). 

At present, I have to look into the sort_objects method to see if it can be use and extended for that. 
`blocks = expr_utils.sort_objects(blocks, operator.attrgetter('name'), _get_block_sort_text)`
May be we could add the GUI_HINT to define a new dependancy between blocks

mormj:
Closing to keep work on `master` PR #4271

PR #4271, opened 2021-02-20
grc: order blocks with GUI Hint first

ChristopheSeguinot:
Signed-off-by: Christophe Seguinot <christophe.seguinot@univ-lille.fr>

This PR add ordering of block in a flowgraph before the .py code is generated.
This prevent PyQT5 from superposing blocks as explained in issue #3371, by simply drawing first those blocks having a GUI Hint.

PR #4272, opened 2021-02-20
grc: order blocks with GUI Hint first

ChristopheSeguinot:

This PR add ordering of block in a flowgraph before the .py code is generated. 
This prevent PyQT5 from superposing blocks as explained in issue #3371, by simpling drawing first those blocks having a GUI Hint.

### DIFF
--- a/grc/core/generator/top_block.py
+++ b/grc/core/generator/top_block.py
@@ -196,6 +196,18 @@ class TopBlockGenerator(object):
         ]
 
         blocks = expr_utils.sort_objects(blocks, operator.attrgetter('name'), _get_block_sort_text)
+
+        # Ordering blocks : blocks with GUI Hint must be processed first to avoid PyQT5 superposing blocks
+        for block in blocks:
+            try:
+                block.without_gui_hint = (block.params['gui_hint'].get_value()=='')
+            except KeyError:
+                # No gui hint
+                block.without_gui_hint = True
+                pass
+
+        blocks = sorted(blocks, key=lambda block: block.without_gui_hint)
+
         blocks_make = []
         for block in blocks:
             make = block.templates.render('make')
--- a/grc/core/generator/top_block.py
+++ b/grc/core/generator/top_block.py
@@ -199,12 +199,8 @@ class TopBlockGenerator(object):
 
         # Ordering blocks : blocks with GUI Hint must be processed first to avoid PyQT5 superposing blocks
         for block in blocks:
-            try:
-                block.without_gui_hint = (block.params['gui_hint'].get_value()=='')
-            except KeyError:
-                # No gui hint
-                block.without_gui_hint = True
-                pass
+            hint = block.params.get('gui_hint')
+            block.without_gui_hint = (hint == None) or (not hint.get_value())
 
         blocks = sorted(blocks, key=lambda block: block.without_gui_hint)
 


willcode:
This looks reasonable. You could get rid of the need exception handling with something like:
```
hint = block.params.get('gui_hint')
block.without_gui_hint = not hint
```

ChristopheSeguinot:
Thank I was looking for such an answer. 
 Your proposal don't work since hint is None or is empty  but I will get rid of the excepetion with: 
 
`            hint = block.params.get('gui_hint')
            block.without_gui_hint = (hint == None) or (not hint.get_value())
`
 
I will soon modify the PR accordingly

willcode:
None and empty both resolve to False.

ChristopheSeguinot:
Tested with 2 widgets, volume and freq.  Only freq
      has a non empty GUIhint.
    In that case "not hint" is false for both widget, I get: 
    
                hint = block.params.get('gui_hint')
                  block.without_gui_hint = not hint # (hint == None) or
      (not hint.get_value())
                  
                  Messages.send('>>> name =: {}
      without_gui_hint={} hint={}\n'.format(block.name, 
    
    >>> name =: volume without_gui_hint=False
      hint=<class 'gnuradio.grc.gui.canvas.param.Param'>
      >>> name =: freq without_gui_hint=False
      hint=<class 'gnuradio.grc.gui.canvas.param.Param'>
    
    
    
    So I think I need first test if hint is none; then i can
      hint.get_value() to test hint. 
    
    
    
    Christophe Seguinot
      Enseignant-Chercheur IRCICA
      03 20 43 64 09
      Pages Personnelles: http://christophe-seguinot.fr/pro/
    On 21/02/2021 15:57, Jeff Long wrote:
    
    
      
      None and empty both resolve to False.
      —
        You are receiving this because you authored the thread.
        Reply to this email directly, view it on GitHub, or unsubscribe.
      [
        {
        "@context": "http://schema.org",
        "@type": "EmailMessage",
        "potentialAction": {
        "@type": "ViewAction",
        "target":
        "https://github.com/gnuradio/gnuradio/pull/4272#issuecomment-782870951",
        "url":
        "https://github.com/gnuradio/gnuradio/pull/4272#issuecomment-782870951",
        "name": "View Pull Request"
        },
        "description": "View this Pull Request on GitHub",
        "publisher": {
        "@type": "Organization",
        "name": "GitHub",
        "url": "https://github.com"
        }
        }
        ]

willcode:
Right, I missed the `get_value()`.

ChristopheSeguinot:
New commit remove the exception.

grc/core/generator/top_block.py
```diff
@@ -196,6 +196,14 @@ def _get_block_sort_text(block):
         ]
 
         blocks = expr_utils.sort_objects(blocks, operator.attrgetter('name'), _get_block_sort_text)
+
+        # Ordering blocks : blocks with GUI Hint must be processed first to avoid PyQT5 superposing blocks
+        for block in blocks:
+            hint = block.params.get('gui_hint')
+            block.without_gui_hint = (hint == None) or (not hint.get_value())
+
+        blocks = sorted(blocks, key=lambda block: block.without_gui_hint)
```
skoslowski, line 205: This code is a little cumbersome, actually. May I suggest to something like this instead:
```python
def without_gui_hint(block):
    hint = block.params.get('gui_hint')
    return hint is None or not hint.get_value()

blocks.sort(key=without_gui_hint)
``` 
Now, if only we could use the walrus operator, then I'd be a one-liner =)

willcode:
And that also avoids adding the field `without_gui_hint` to a block object defined elsewhere.

ChristopheSeguinot:
Great Sebastian, I will use your code.

willcode:
Just noticed block sort right before the diff. Is the sort by name there to make generated code more readable, and does it still do anything if the new sort is added?

ChristopheSeguinot:
Effectively, we must investigate more before making PR. 

- `blocks = expr_utils.sort_objects(blocks, operator.attrgetter('name'), _get_block_sort_text)` is used to sort blocks so that dependent variables are processed first
- ` blocks.sort(key=without_gui_hint)` is added to sort so that blocks with GUI Hint are processed first

Is this conflicting ? 

- Taking into account the goal of sort_objects(), I would say yes. 
- However, looking deeper in the code, I can't find any situation in which sort_objects() is usefull ,since parameter are declared before every widgets (as in the sample code extract below)

So I conclude that whether : 

- I don't understand sort_objects() usefullness or 
- sort_objects() is not longer needed here. 

Who could lead us into the right direction? 

`
        # Variables
        self.zzvariable_qtgui_chooser_0 = zzvariable_qtgui_chooser_0 = 0
        self.volume = volume = 0.50
        self.variable_qtgui_chooser_0_0 = variable_qtgui_chooser_0_0 = 0
        self.samp_rate = samp_rate = 48000
        self.freq = freq = 600

        # Blocks
        self.blocks_multiply_const_vxx_0 = blocks.multiply_const_ff(volume)
        self.audio_sink_0 = audio.sink(48000, '', True)
        self.analog_sig_source_x_0 = analog.sig_source_f(samp_rate, analog.GR_COS_WAVE, freq, 0.5, 0, 0)

        self._freq_range = Range(volume, 10000, 50, 600, 200)
`

willcode:
In the mean time, this is one of those small PRs where it's not worth tracking 3 PRs and keeping them sync'd. The whole discussion ended up on this 3.9 version, unfortunately, but the master version would be the one to keep open until there's a good solution. Once this is accepted and seems to be working on master, then let's get it back to 3.8/3.9.

mormj:
Closing to keep work on `master` PR #4271

PR #4275, opened 2021-02-22
digital: qa: Add preconditions to wait for qa_packet_headerparser_b

mbr0wn:
This is an attempt to avoid QA failures due to messages not being
delivered before wait() concludes.

### DIFF
--- a/gr-digital/python/digital/qa_packet_headerparser_b.py
+++ b/gr-digital/python/digital/qa_packet_headerparser_b.py
@@ -7,8 +7,6 @@
 #
 #
 
-
-import time
 import random
 
 from gnuradio import gr, gr_unittest, blocks, digital
@@ -48,7 +46,7 @@ class qa_packet_headerparser_b (gr_unittest.TestCase):
         self.tb.connect(src, parser)
         self.tb.msg_connect(parser, "header_data", sink, "store")
         self.tb.start()
-        time.sleep(1)
+        self.waitFor(lambda: sink.num_messages() == 3)
         self.tb.stop()
         self.tb.wait()
         self.assertEqual(sink.num_messages(), 3)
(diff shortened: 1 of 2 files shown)


gr-digital/python/digital/qa_packet_headerparser_b.py
```diff
@@ -16,6 +16,16 @@
 import pmt
 
 
+def _wait_for_msgs(precondition):
+    """
+    Simple loop to check for a condition
+    """
+    stop_time = time.time() + 2.0
```
jdemel, line 23: Maybe put the default time `2.0` in a function parameter.

gr-digital/python/digital/qa_packet_headerparser_b.py
```diff
@@ -16,6 +16,16 @@
 import pmt
 
 
+def _wait_for_msgs(precondition):
```
jdemel, line 19: Would it make sense to add this function as a method to `gr_unittest`?

jdemel (commented):
I have 2 suggestions for broader application. I assume these would make msg QA tests better for everyone. And easier to use.

mormj:
@mbr0wn - does the suggestion to move to gr_unittest seem reasonable?

mbr0wn:
Yeah, I think it is.

gr-digital/python/digital/qa_packet_headerparser_b.py
mbr0wn: OK done.

gr-digital/python/digital/qa_packet_headerparser_b.py
mbr0wn: It is in gr_unittest

mormj (approved):

mormj:
Feedback has been addressed.  Merging.

Issue #4109, opened 2021-01-17
GRC doesn´t show variable names only values


reald:
At the moment gnuradio-companion shows only values for block attributes in a flow graph but not the variables or expressions behind. This makes it hard to understand and explain more complex flowgraphs because you have to look inside every single block if other variables are used here. So instead of 

**Signal source**
- Sample Rate: 32k
- Frequency: 1k

it should be possible to show block properties on the gnuradio canvas with their expressions behind like:

**Signal source**
- Sample Rate: samp_rate
- Frequency: freq

Additionally it should be possible to show expression and the current value like:

**Signal source**
- Sample Rate: samp_rate (32k)
- Frequency: freq (1k)

The new behaviour should be selectable in the menu.

dkozel:
I can see the reasoning behind this and think it would be a nice option to have. When longer values are supplied it won't be very readable, but that's already true of values like vectors or long strings.

jasonuher:
> I can see the reasoning behind this and think it would be a nice option to have. When longer values are supplied it won't be very readable, but that's already true of values like vectors or long strings.

I used the same truncate function on the variable names as was used on the values, so the behavior should be the same.

ChristopheSeguinot:
It seems that this small issue is closely related to this feature request (from the code change point of view)
Suppose I've a block with Sample Rate defined by variable samp_rate fixed at 1000 Hz. 

- When the block is enabled, it actually shows as "Sample Rate : 1000 Hz"
- When a block is disabled, it's displayed in red as  "Sample Rate : samp_rate"

1. The text should be the same (with/wo string "samp_rate") in both case
2. I think there is no reason to show this in red for disabled blocks.

mormj:
Feature merged as part of #4279.  Closing

PR #4279, opened 2021-02-24
grc: optionally show variable names in graph #4109

jasonuher:
1. Add option to show variable name only
2. Add option to show variable name and value
3. default to traditional behavior

Signed-off-by: Jason Uher <jason.uher@jhuapl.edu>


jasonuher:
I think this solves #4109

By default the behavior matches the traditional behavior (showing just the *value* of the parameter).  You can optionally enable showing the name, value, or both.  If you disable showing both the name and value it reverts back to the traditional formatting (ie, just showing the value).

"Normal" Mode
default_(classic)

SHowing variable name only (note truncation of long names)
name_only

Showing both name and value (ditto on truncation)
name_and_value

A look at the menu switches:
menu_items

mormj:
@jasonuher - I absolutely love this feature and think it improves the usability of GRC.  Can't even tell you how many times I'm clicking on block properties to see which variable is set for the value, and then to see it goes several variables deep.

mormj (approved):

grc/gui/canvas/param.py
```diff
@@ -156,6 +160,20 @@ def format_block_surface_markup(self):
         Returns:
             a pango markup string
         """
+
+        if (str(self.value) == str(self._evaluated)): #This is not a parameter defined by a variable
```
mbr0wn, line 164:
```suggestion
        if str(self.value) == str(self._evaluated): # This is not a parameter defined by a variable
```

grc/gui/canvas/param.py
```diff
@@ -156,6 +160,20 @@ def format_block_surface_markup(self):
...
+            var_value = Utils.encode( self.pretty_print().replace('\n', ' ') )
+        else: # this is defined by a variable, show either the variable name or the value depending on options
+            var_value = None
+
+            if( Actions.TOGGLE_SHOW_VARIABLE_PARAMS_NAME.get_active() ):
+                var_value = "<i>"+Utils.encode( self.truncate(self.value) ) + "</i>"
+
+                if( Actions.TOGGLE_SHOW_VARIABLE_PARAMS_VALUE.get_active() ):
```
mbr0wn, line 172:
```suggestion
                if Actions.TOGGLE_SHOW_VARIABLE_PARAMS_VALUE.get_active():
```

mbr0wn (commented):
Please keep the Python formatting as-is. We don't (yet) to automatic Python formatting checks, but we don't want to make our life unnecessarily difficult for once we do.

mbr0wn:
The feature is super nice!

jasonuher:
@mbr0wn I think I'm an old man who's behind the times on auto-linters (I'm still not sure I'm using clang correctly :p)  Is there something for python that I can use to check it matches?

mbr0wn:
Pylint is what I use. And also don't forget to sign off commits :)

Or squash them.

grc/gui/Application.py
```diff
@@ -490,6 +492,14 @@ def flow_graph_update(fg=flow_graph):
...
             for page in main.get_pages():
                 flow_graph_update(page.flow_graph)
+        elif action == Actions.TOGGLE_SHOW_VARIABLE_PARAMS_NAME:
+            action.set_active(not action.get_active())
+            action.save_to_preferences()
+            flow_graph_update()
+        elif action == Actions.TOGGLE_SHOW_VARIABLE_PARAMS_VALUE:
+            action.set_active(not action.get_active())
+            action.save_to_preferences()
+            flow_graph_update()
```
skoslowski, line 502: What happens if both are `False`? Maybe ticking one off, enables the other? That way 3 states are possible.

grc/gui/canvas/param.py
```diff
@@ -156,6 +160,21 @@ def format_block_surface_markup(self):
         Returns:
             a pango markup string
         """
+
+        if str(self.value) == str(self._evaluated): # This is not a parameter defined by a variable
```
skoslowski, line 164: I believe `value` is already a `str`
Also, please update comment to wording change mentioned above, e.g. a value of "1+1" will result in false even though it not defined by a variable

grc/gui/canvas/param.py
```diff
@@ -156,6 +160,21 @@ def format_block_surface_markup(self):
...
+            var_value = Utils.encode(self.pretty_print().replace('\n', ' '))
```
skoslowski, line 165: `var_value` has some other historical meaning in grc code. May I kindly suggest to use maybe `dispaly_value` instead

grc/gui/canvas/param.py
```diff
@@ -156,6 +160,21 @@ def format_block_surface_markup(self):
...
+
+            if Actions.TOGGLE_SHOW_VARIABLE_PARAMS_NAME.get_active():
+                var_value = "<i>"+Utils.encode( self.truncate(self.value) ) + "</i>"
+
+                if Actions.TOGGLE_SHOW_VARIABLE_PARAMS_VALUE.get_active():
+                    var_value = var_value + "=" + Utils.encode(
+                        self.pretty_print().replace('\n', ' '))
+            else:
+                var_value = Utils.encode( self.pretty_print().replace('\n', ' ') )
+
```
skoslowski, line 177: The two actions are independently toggled in the GUI, but here one depends on the other. From a users perspective that may be unexpected. 
With the suggestion above (allow only 3 states) the logic here could be flattened to reflect the 3 possible outcomes. It may be useful to sort-out the logic beforehand by defining `show_value`, `show_expression` based on the Actions and the "is literal" check, and then have a flat `if`, `elif` here for output formatting

skoslowski (changes requested):
I really :heart: this feature. Very useful! Thank you for putting this together.
I have a found few issues / nits in the patch. Lastly, I am not 100% sure about the "=" if both, expression and value, are displayed. But I don't have a better alternative ATM

grc/gui/canvas/param.py
jasonuher: You're right, this is probably not the best test to achieve the goal stated in the comment.  This also applies to things that are variables in other classes (ie, the wfm in the signal source block or the constellations in the demod block)

Do you think there's a cleaner way to determine if the string the user inputs is an original input vs something that's evaluated?

grc/gui/canvas/param.py
jasonuher: this in excellent idea

jasonuher:
> Lastly, I am not 100% sure about the "=" if both, expression and value, are displayed. But I don't have a better alternative ATM

I agree, but I wasn't sure what the right path forward there was.  The parens (like shown in the original issue) probably aren't terrible but I wasn't sure how to make it clear which was the 'input' and which was the evaluated field that would actually be executed (ie the 'value').  I did experiment with adding a second line, but it made the boxes insanely crowded when there was more than one or two parameters

grc/gui/canvas/param.py
```diff
@@ -10,6 +10,7 @@
 from .drawable import Drawable
 from .. import ParamWidgets, Utils, Constants
 from ...core.params import Param as CoreParam
+from .. import Actions
```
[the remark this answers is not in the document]
jasonuher: I thought about parens - opinion?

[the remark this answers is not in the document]
jasonuher: done

grc/gui/Application.py
jasonuher: right now it reverts to the default behavior, which I think is the right thing to do.  I don't have a good concept for how it would work with the 'full coverage' of options (like a checkbox for "show both", "show vals", and "show params" that all automatically check/uncheck eachother?)

jasonuher:
Okay I think I did the rebase right and installed the pep8 linter so everything should be good to go?

The only real 'outstanding' item is whether this is the correct way to covey the setting to the user (ie 'what happens when both settings are off').  I don't there's a perfect answer here, but for now if both settings are off it it just reverts to the original behavior of calling 'self.pretty_print'

mbr0wn:
> The only real 'outstanding' item is whether this is the correct way to covey the setting to the user (ie 'what happens when both settings are off'). I don't there's a perfect answer here, but for now if both settings are off it it just reverts to the original behavior of calling 'self.pretty_print'

I think that's the right call -- minimally invasive.

grc/gui/canvas/param.py
```diff
@@ -86,33 +86,36 @@ def format_tooltip_text(self):
             tooltip_lines.extend(' * ' + msg for msg in errors)
         return '\n'.join(tooltip_lines)
 
+
+
+    ##################################################
+    # Truncate helper method
+    ##################################################
+    def truncate(self, string, style=0):
```
mbr0wn, line 94: You could have made this static, but that's a teeny tiny nit.

grc/gui/canvas/param.py
```diff
@@ -156,6 +159,29 @@ def format_block_surface_markup(self):
...
+        if (not is_evaluated or show_value or not show_expr):
+            display_value += Utils.encode(
+                self.pretty_print().replace('\n', ' '))
+
+        # Include the expression that was evaluated to get the value
+        if (is_evaluated and show_expr):
+            expr_string = "<i>" + \
+                Utils.encode(self.truncate(self.value)) + "</i>"
+
+            if(display_value != ""):  # We are already displaying the value
```
mbr0wn, line 180:
```suggestion
            if display_value != "":  # We are already displaying the value
```
Or even:
```suggestion
            if display_value:  # We are already displaying the value
```

grc/gui/canvas/param.py
```diff
@@ -156,6 +159,29 @@ def format_block_surface_markup(self):
...
+
+        display_value = ""
+
+        # Include the value defined by the user (after evaluation)
+        if (not is_evaluated or show_value or not show_expr):
+            display_value += Utils.encode(
+                self.pretty_print().replace('\n', ' '))
+
+        # Include the expression that was evaluated to get the value
+        if (is_evaluated and show_expr):
```
mbr0wn, line 176:
```suggestion
        if is_evaluated and show_expr:
```

grc/gui/canvas/param.py
```diff
@@ -156,6 +159,29 @@ def format_block_surface_markup(self):
...
+
+        # TODO: is this the correct way to do this?
+        is_evaluated = self.value != str(self.get_evaluated())
+        show_value = Actions.TOGGLE_SHOW_PARAMETER_EVALUATION.get_active()
+        show_expr = Actions.TOGGLE_SHOW_PARAMETER_EXPRESSION.get_active()
+
+        display_value = ""
+
+        # Include the value defined by the user (after evaluation)
+        if (not is_evaluated or show_value or not show_expr):
```
mbr0wn, line 171:
```suggestion
        if not is_evaluated or show_value or not show_expr:
```

grc/gui/canvas/param.py
```diff
@@ -156,6 +159,29 @@ def format_block_surface_markup(self):
         Returns:
             a pango markup string
         """
+
+        # TODO: is this the correct way to do this?
```
mbr0wn, line 163: I think we should resolve this question before merging?

mbr0wn (approved):
I would be OK merging in the current state. I flagged some minor nits which aren't a showstopper.

grc/gui/canvas/param.py
jasonuher: I think this actually a question for @skoslowski  since he seemed to have some idea of how to determine 'literalness'

grc/gui/canvas/param.py
jasonuher: I didn't think it could be since it uses "self.name".  I think if you want this to be static you would have to pass the name (or, alternatively, max_len) in from the caller right?

grc/gui/canvas/param.py
skoslowski: Well, to find out if there no variable dependencies one would have try to eval the expression in an otherwise namespace, to detect a literal `ast.literal_parse()` could be used. But I think neither is really worth it and not really appropriate here. I suggest to put in a comment and keep the check as is.

grc/gui/canvas/param.py
```diff
@@ -156,6 +159,29 @@ def format_block_surface_markup(self):
...
+            display_value += Utils.encode(
+                self.pretty_print().replace('\n', ' '))
+
+        # Include the expression that was evaluated to get the value
+        if (is_evaluated and show_expr):
+            expr_string = "<i>" + \
+                Utils.encode(self.truncate(self.value)) + "</i>"
+
+            if(display_value != ""):  # We are already displaying the value
+                display_value = expr_string+"="+display_value
```
skoslowski, line 181:
```suggestion
                display_value = expr_string + "=" + display_value
```

willcode:
@jasonuher Would you be interested in backporting this to 3.8? I think the only differences have to do with code formatting. I'll get to it eventually if you can not.

PR #4285, opened 2021-02-28
filter: Remove manual memory management for polyphase filterbank

ThomasHabets:
Signed-off-by: Thomas Habets <thomas@habets.se>


mormj (approved):
Changes here make sense.  Thank!

willcode (approved):

Issue #2214, opened 2018-12-05
QA: qa_pdu flakiness


noc0lour:
From https://ci.gnuradio.org/buildbot/#/builders/56/builds/337/steps/5/logs/stdio
```
198/360 Test #198: qa_pdu .............................................***Failed    0.59 sec
E..
======================================================================
ERROR: test_000 (__main__.test_pdu)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/data/src/pull/2211/master/gr-blocks/python/blocks/qa_pdu.py", line 77, in test_000
    result_msg = dbg.get_message(0)
  File "/cache/workdir/Fedora_29_0/build_Fedora_29/build/gr-blocks/python/blocks/../../swig/blocks_swig0.py", line 7074, in get_message
    return _blocks_swig0.message_debug_sptr_get_message(self, i)
RuntimeError: message_debug: index for message out of bounds.
----------------------------------------------------------------------
Ran 3 tests in 0.012s
FAILED (errors=1)
```

nickoe:
Maybe this can be closed? It looks like the qa_pdu test passes on all the buildbot tests consistently. Or are there still cases where this test fails?

noc0lour:
Hmmm, If we didn't do something about it it's probably not yet fixed.

noc0lour:
It's pretty rare to spot

mbr0wn:
Have not seen this in a very long time. Closing.

PR #4287, opened 2021-03-01
cmake: Fix Boost unit test framework component availability variable

acinonyx:
This fix enables back 'testing-support' component which has been
disabled by a regression introduced on a3061b8f.

Signed-off-by: Vasilis Tsiligiannis <acinonyx@openwrt.gr>

### DIFF
--- a/.github/workflows/make-test.yml
+++ b/.github/workflows/make-test.yml
@@ -75,7 +75,7 @@ jobs:
     - name: CMake
       env:
         CXXFLAGS: ${{ matrix.cxxflags }}
-      run: 'cd /build && cmake ${GITHUB_WORKSPACE} -DENABLE_DOXYGEN=OFF -DBoost_unit_test_framework_FOUND=ON'
+      run: 'cd /build && cmake ${GITHUB_WORKSPACE} -DENABLE_DOXYGEN=OFF'
     - name: Make
       run: 'cd /build && make -j2 -k'
     - name: Make Test
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-# Copyright 2011-2020 Free Software Foundation, Inc.
+# Copyright 2011-2021 Free Software Foundation, Inc.
 #
 # This file is part of GNU Radio
 #
@@ -47,7 +47,7 @@ include(GrComponent)
 ########################################################################
 include(GrBoost)
 GR_REGISTER_COMPONENT("testing-support" ENABLE_TESTING
-         Boost_unit_test_framework_FOUND )
+         Boost_UNIT_TEST_FRAMEWORK_FOUND )
 
 # Set the version information here
 SET(VERSION_MAJOR 3)


drmpeg (approved):
This is a very good fix. This issue only affects older versions of CMake, and this patch solves it.

mormj:
Currently our CI tests have the old variable name baked in: 
https://github.com/gnuradio/gnuradio/blob/031c7d863f38534809a2d8badf74886cbc9fc944/.github/workflows/make-test.yml#L78

I'm guessing that this variable (since it is named _FOUND) is intended to be the output of a find function.  Curious whether removing the variable from the github actions will still pass CI.

drmpeg:
> Currently our CI tests have the old variable name baked in:
> 
> https://github.com/gnuradio/gnuradio/blob/031c7d863f38534809a2d8badf74886cbc9fc944/.github/workflows/make-test.yml#L78
> 
> I'm guessing that this variable (since it is named _FOUND) is intended to be the output of a find function. Curious whether removing the variable from the github actions will still pass CI.

That hard coding is to get around this bug.

mormj:
Then let's remove it from the github action so that CI can run cleanly against this fix.

acinonyx:
> Currently our CI tests have the old variable name baked in:
> 
> https://github.com/gnuradio/gnuradio/blob/031c7d863f38534809a2d8badf74886cbc9fc944/.github/workflows/make-test.yml#L78
> 
> I'm guessing that this variable (since it is named _FOUND) is intended to be the output of a find function. Curious whether removing the variable from the github actions will still pass CI.

Hehe, nice catch! I totally missed that. This basically almost defeated the purpose of having the 'cmake' test in the first place. I pushed an update commit which removes this variable from 'make-test.yml'.

acinonyx:
It appears that `Fedora 33` tests failed but there is no apparent relation to this PR. I am unable to reproduce the failure locally, in the same container. Is this test case known to be unstable?

mormj:
Haven't seen that error but it seems unrelated to this patch
@jacobagilbert - have you ever seen qa_pdu fail like that?

acinonyx:
https://github.com/gnuradio/gnuradio/issues/2214

jacobagilbert:
@mormj I have seen something similar with my own QA before, essentially what is happening is the flowgraph is ending before the message_debug block can store the message so its not available. A time.sleep() statement can make this better but I wouldn't worry about this specific failure. I wish there was a nice way to click 'run again' on CI QA.

mormj:
>  I wish there was a nice way to click 'run again' on CI QA.

Best 2 out 3 ;)

acinonyx:
I force-pushed the commit with a new Git SHA to trigger the CI. Now all tests are passing \o/

mbr0wn (approved):

mbr0wn:
Thanks. The CMake hack in our github actions foo was always a workaround... very glad we can remove it.

PR #4290, opened 2021-03-01
qtgui: Remove manual memory management from FrequencyDisplayPlot

ThomasHabets:
qtgui is tricker to fix than the other sections, as it juggles pointers around with less clear ownership. That plus the fact that QT does some pointer ownership magic with lots of foot-guns.

So I'm doing the low hanging fruit that aren't juggled buffers and QT objects first.

### DIFF
--- a/gr-qtgui/include/gnuradio/qtgui/FrequencyDisplayPlot.h
+++ b/gr-qtgui/include/gnuradio/qtgui/FrequencyDisplayPlot.h
@@ -132,7 +132,7 @@ private:
     void _resetXAxisPoints();
     void _autoScale(double bottom, double top);
 
-    std::vector<double*> d_ydata;
+    std::vector<std::vector<double>> d_ydata;
 
     QwtPlotCurve* d_min_fft_plot_curve;
     QwtPlotCurve* d_max_fft_plot_curve;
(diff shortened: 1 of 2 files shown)


mormj (approved):
LGTM!

willcode (approved):
Slightly different behavior during resize (existing data not zeroed). Would that have any noticeable effect? Can revisit if it shows.

ThomasHabets:
@willcode for these changes I try to maintain existing zeroing behavior. `new[]` does not initialize the data.

There is a slight difference for these commits, but they always aim to "err" on the side of over-initializing. E.g. `delete[]` followed by `new[]` initializes nothing, whereas `.resize()` does initialize any *increase* elements to zero.

But I have now double-checked. More info [here](https://stackoverflow.com/questions/7546620/operator-new-initializes-memory-to-zero). I also made a small test:

```c++
#include<iostream>
#include<algorithm>

int main() {
  double *p = nullptr;
  for (int i = 0; i < 100; i++) {
    p = new double[100];
    for (int j = 0; j < 100; j++) {
      if (p[j] != 0.0) {
        std::cout << "not inited\n";
        return 1;
      }
    }
    std::fill(p, p+100, 99.999);
    delete[] p;
  }
}
```

Or is this not what you meant; that I missed a spot?

willcode:
Ah, then it was previously starting off with totally uninitialized data. Ok, thanks!

Issue #4177, opened 2021-02-02
stream_mode_t API (from stream_cmd_t) not exposed in gnuradio uhd module


than97:
In transitioning from Python 2.7/ GNURadio 3.7.9 to Python 3/ GNURadio 3.10, the stream_mode_t API (from stream_cmd_t) is no longer exposed in gnuradio uhd module.  It does not appear as though the gnuradio uhd module fully exposes the public API for the stream command. This results in breaks between 3.7 and 3.10. The desired outcome is that the gnuradio uhd block correctly exploses all the public API's, as it previously did, using swig.

Command Input: 
from gnuradio import uhd
uhd.stream_cmd.__dict__

Python 2.7/ GNURadio 3.7.9 Output: 
dict_proxy({'STREAM_MODE_NUM_SAMPS_AND_MORE': 109, '__module__': 'gnuradio.uhd.uhd_swig', 'stream_mode': <property object at 0x7f287b5b2a70>, 'thisown': <property object at 0x7f287b5b2a10>, 'STREAM_MODE_START_CONTINUOUS': 97, 'STREAM_MODE_STOP_CONTINUOUS': 111, '__repr__': <function _swig_repr at 0x7f287afcd2d0>, 'num_samps': <property object at 0x7f287b5b2290>, 'STREAM_MODE_NUM_SAMPS_AND_DONE': 100, '__dict__': <attribute '__dict__' of 'stream_cmd_t' objects>, 'time_spec': <property object at 0x7f287b5b2710>, 'stream_now': <property object at 0x7f287b5b2470>, '__weakref__': <attribute '__weakref__' of 'stream_cmd_t' objects>, '__doc__': 'Proxy of C++ uhd::stream_cmd_t class.', '__init__': <function __init__ at 0x7f287aaaf050>, '__swig_destroy__': <built-in function delete_stream_cmd_t>})

Python 3/ GNURadio 3.10: 
mappingproxy({'__init__': <instancemethod __init__ at 0x7f396f4c90d0>, '__doc__': None, '__module__': 'gnuradio.uhd.uhd_python', 'num_samps': <property object at 0x7f396f70d1d0>, 'time_spec': <property object at 0x7f396f70d2c0>, 'stream_now': <property object at 0x7f396f70d3b0>})

dkozel:
[...]

PR #4293, opened 2021-03-01
GR-UHD: Support for stream_mode_t 

vmw:
Add support for stream_mode_t, partially fixing #4177.

This may be invoked within python using something along the lines of;

```
cmd = uhd.stream_cmd_t(uhd.stream_mode_t.STREAM_MODE_NUM_SAMPS_AND_DONE)
cmd.num_samps = frame[1]
cmd.stream_now = False
cmd.time_spec = uhd.time_spec(frame[0])
csrc.issue_stream_cmd(cmd)
```

This behaviour is slightly different than what was in place with previous (swig) versions of gnuradio. When using pybind, the result we replace old instances of  "uhd.stream_cmd_t" with "uhd.stream_mode_t".

This has been tested on hardware.

Still pending a fix for "tune_request_t::policy_t".

### DIFF
--- a/gr-uhd/python/uhd/bindings/uhd_types_python.cc
+++ b/gr-uhd/python/uhd/bindings/uhd_types_python.cc
@@ -95,9 +95,22 @@ void bind_uhd_types(py::module& m)
     py::class_<stream_cmd_t>(m, "stream_cmd_t")
         .def(py::init<stream_cmd_t::stream_mode_t>())
         // Properties
+        .def_readwrite("stream_mode", &stream_cmd_t::stream_mode)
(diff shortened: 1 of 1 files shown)


mormj:
@vmw - Thank you for providing the PR - in order to pass the C++ Formatting check, you will need to run `clang-format` on any modified files.  You can do this indiscriminately by running:
```
find ./ -iname *.h -o -iname *.cc | xargs clang-format -i
```
Then git add/commit/force push the changes

vmw:
Thanks for the advice; I've force pushed, and looks like everything is passing now.

mbr0wn (approved):

mormj (approved):

Issue #3028, opened 2020-01-06
Cleanup gr::impl namespace


marcusmueller:
gr::impl only contains realtime stuff, and using `impl` instead of `realtime` really seems a bad choice..

PR #4298, opened 2021-03-02
runtime: clean up realtime impl namespace

mormj:
Renames the gr::impl namespace to gr::realtime and moves impl stuff inside lib/ 
Removes the impl stuff from the python bindings
Some other minor cleanup

Addresses #3028


drmpeg:
I think that grc/tests/resources/top_block.py file is spurious. It gets created when you do ```make test```.

mormj:
good catch!  need to get that into the .gitignore

gnuradio-runtime/include/gnuradio/realtime.h
```diff
@@ -12,10 +12,16 @@
 #define INCLUDED_GR_REALTIME_H
 
 #include <gnuradio/api.h>
-#include <gnuradio/realtime_impl.h>
 
 namespace gr {
 
+typedef enum { RT_OK = 0, RT_NOT_IMPLEMENTED, RT_NO_PRIVS, RT_OTHER_ERROR } rt_status_t;
+
+enum rt_sched_policy {
```
ThomasHabets, line 20: Maybe switch to `enum class` while you're at it, for better type safety?

ThomasHabets (changes requested):

gnuradio-runtime/include/gnuradio/realtime.h
mbr0wn: I also prefer enum classes, but that's an API break that we might be able to avoid.

ThomasHabets (approved):

mbr0wn (approved):

Issue #4239, opened 2021-02-15
Signal Source block pair message handling broken


bgottula:
When sending a `pair` message to the the Signal Source block's `cmd` port, I get the following exception (on maint-3.8):

```
thread[thread-per-block[0]: <block sig_source(5)>]: pmt_car: wrong_type : offset
```

The message was constructed as follows (using a Message Strobe block in GRC), which should be allowed if I understand the source code and documentation correctly:

```
pmt.cons(pmt.intern("offset"), pmt.from_double(25.0))
```

Digging into the source, I found the following lines which seem problematic:
```
    if (pmt::is_dict(msg)) {
        list_of_items = pmt::dict_items(msg);
    } else if (pmt::is_pair(msg)) {
        list_of_items = pmt::list1(msg);
    } else {
        GR_LOG_WARN(this->d_logger, "malformed message: is not dict nor pair");
        return;
    }
```
Ref: https://github.com/gnuradio/gnuradio/blob/master/gr-analog/lib/sig_source_impl.cc#L93-L96

The docstring for `pmt::is_dict()` says "Return true if \p obj is a dictionary (warning: also returns true for a pair)", and similarly the docstring for `pmt::is_pair()` says "Return true if \p obj is a pair, else false (warning: also returns true for a dict)". This means the first condition in the `if` statement will evaluate to `true` if `msg` is a pair _or_ a dict, and I don't think the `else if` part is reachable under any circumstances, which is clearly not what the author of this if statement was expecting.

I'm new to message passing in GNU Radio so I'm not sure what the right course of action is to fix this. Is there a reliable way to distinguish between pairs and dicts? Or is that even the right thing to be doing here?

willcode:
In this case,
```
>>> x=pmt.cons(pmt.intern("offset"), pmt.from_double(25.0))
>>> x.is_pair()
True
>>> x.is_dict()
False
```
In the QA code for this block, here is the message format:
```
        src._post(
            pmt.to_pmt('freq'),
            pmt.to_pmt({
                "freq": frequency,
                "ampl": amplitude,
                "offset": offset
            }))
```
where not all of the params need to be present. See if that helps.

willcode:
@bgottula Have you had a chance to try this again using the `to_pmt` method above?

jacobagilbert:
@willcode The `pdu_dict` class that allows differentiation between pairs and dicts as you mention was not added until 3.9. 

[...]

PR #4313, opened 2021-03-03
pmt: stop treating all pairs like they are dicts (backport to maint-3.8)

willcode:
create a new derived class for pmt dicts so they can be distinguished without complicated try/catch logic, updated QA with a few additional checks. also added an is_pdu() method which returns true if the pmt is a pair of a dict-type PMT and a uniform-vector type, otherwise false

(cherry picked from commit 941c6c895a5724c142cfb0a32220885062373d57)

Backport https://github.com/gnuradio/gnuradio/pull/3389

### DIFF
--- a/gnuradio-runtime/include/pmt/pmt.h
+++ b/gnuradio-runtime/include/pmt/pmt.h
@@ -654,8 +654,12 @@ c64vector_writable_elements(pmt_t v, size_t& len); //< len is in elements
  * ------------------------------------------------------------------------
  */
 
-//! Return true if \p obj is a dictionary (warning: also returns true for a pair)
+//! Return true if \p obj is a dictionary
 PMT_API bool is_dict(const pmt_t& obj);
+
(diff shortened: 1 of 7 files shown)


willcode:
Addresses https://github.com/gnuradio/gnuradio/issues/4239

@jacobagilbert Running this backport through CI to see what happens. It looks API backward compatible since the changes are internal or static. Do you know whether the change in behavior will affect users in any bad way?

jacobagilbert:
This was not backported when the original PR went in because this definitely does change existing behavior. The New derived class is not really risky, and we could go through in-tree use of `is_dict()` who knows what it could do to OOTs expecting previous behavior.

I really like this change and can help go through in tree use if you want. But there is risk to OOTs. Thoughts?

willcode:
Interesting that most of the OOTs where I've found `is_pair()` really mean `is_pdu()`. Haven't found any use of `is_pair()` that would be very happy to run across a dict.

PR #4314, opened 2021-03-03
cmake: remove custom FindGSL module

mormj:
The in-tree FindGSL.cmake module differs from the standard cmake FindGSL
in the target names that are output.  (gsl::gsl vs GSL::gsl).  This
makes linking an OOT to something that links to gsl cause issues and
requires copying in FindGSL.cmake into the OOT.  Since CMake 3.8
(current project minimum) installs FindGSL, use that instead

There are probably other cmake modules we could remove but I haven't done a thorough inventory.


marcusmueller (approved):

noc0lour (approved):
Hell yeah!

Issue #1956, opened 2018-08-17
Deprecate generic_mod, work over narrowband examples


marcusmueller:
Basically, everywhere that generic_mod / `type_1_mods` is used should be deprecated, since these no longer contain many useful constellations.

marcusmueller:
outstanding issue, but might not have high priority since decruftification of the 3.x code base won't entail removing parts of the ecosystem

PR #4318, opened 2021-03-04
Python Implementation of QAM modulation

Joseph-Cannaday:
The python script for QAM(qam.py) had no method of QAM modulation, and only the relevant methods for QAM constellation. This edit implements a qam_mod class to allow for functioning QAM modulation.

### DIFF
--- a/gr-digital/python/digital/qam.py
+++ b/gr-digital/python/digital/qam.py
@@ -4,7 +4,7 @@
 # This file is part of GNU Radio
 #
 # SPDX-License-Identifier: GPL-3.0-or-later
-#
+# 
 #
 
 """
@@ -253,5 +253,37 @@ def large_ampls_to_corners_mapping(side, points, width):
             sector_values.append(index)
     return sector_values
 
+# /////////////////////////////////////////////////////////////////////////////
+#                           QAM modulation
+# /////////////////////////////////////////////////////////////////////////////
+
+
+class qam_mod(generic_mod):
+    """
+    Hierarchical block for RRC-filtered QAM modulation.
+
+    The input is a byte stream (unsigned char), treated as a series of packed
+    symbols. Symbols are grouped from MSB to LSB.
+
+    The output is the complex modulated signal at baseband, with a given number
+    of samples per symbol.
+
+    If "Samples/Symbol" is 2, and "Number of Constellation Points" is 4, a
+    single byte contains four symbols, and will produce eight samples.
+
+    Args:
+        constellation_points: Number of constellation points (must be a power of two) (integer).
+        mod_code: Whether to use a gray_code (digital.mod_codes.GRAY_CODE) or not (digital.mod_codes.NO_CODE).
+        differential: Whether to use differential encoding (boolean).
+    """
+    # See generic_mod for additional arguments
+    __doc__ += shared_mod_args
+    def __init__(self, constellation_points = _def_constellation_points,
+                      differential = _def_differential,
+                      mod_code = _def_mod_code,
+                 *args, **kwargs):
+        constellation = qam_constellation(constellation_points, differential, mod_code)
+        super(qam_mod, self).__init__(constellation, differential, *args, **kwargs)
 
 modulation_utils.add_type_1_constellation('qam', qam_constellation)
+modulation_utils.add_type_1_constellation('qam', qam_mod)


gr-digital/python/digital/qam.py
```diff
@@ -253,5 +257,33 @@ def large_ampls_to_corners_mapping(side, points, width):
             sector_values.append(index)
     return sector_values
 
+class qam_mod(generic_mod):
```
Joseph-Cannaday, line 260: added qam_mod class for QAM modulation

gr-digital/python/digital/qam.py
```diff
@@ -3,13 +3,28 @@
...
 #
+# GNU Radio is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with GNU Radio; see the file COPYING.  If not, write to
+# the Free Software Foundation, Inc., 51 Franklin Street,
+# Boston, MA 02110-1301, USA.
```
mbr0wn, line 19: Please keep the SPDX

gr-digital/python/digital/qam.py
```diff
@@ -3,13 +3,28 @@
...
 #
 
 """
 QAM modulation and demodulation.
 """
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import unicode_literals
```
mbr0wn, line 27:
```suggestion
```

GNU Radio 3.9/10 is Python3-only, so this is not necessary.

gr-digital/python/digital/qam.py
```diff
@@ -19,7 +34,7 @@
 from .utils.gray_code import gray_code
 from .utils import mod_codes
 from . import modulation_utils
-from . import digital_python as digital
+from . import digital_swig as digital
```
mbr0wn, line 37: This is a 3.8 thing

Joseph-Cannaday: Ok, I was working off the 3.8 build code, so that's likely why

mbr0wn:
@Joseph-Cannaday Please squash all commits, and sign it (see DCO check). Please use a "proper" commit subject line (like: `digital: Add Python implementation of QAM modulation`. Please fix the failing QA tests.

gr-digital/python/digital/qam.py
mbr0wn: ...I mean, instead of the verbose header. Please keep the header like on the original file.

gr-digital/python/digital/qam.py
```diff
@@ -29,16 +42,13 @@
...
 def is_power_of_four(x):
     v = log(x) / log(4)
     return int(v) == v
 
-
 def get_bit(x, n):
     """ Get the n'th bit of integer x (from little end)."""
-    return (x & (0x01 << n)) >> n
-
+    return (x&(0x01 << n)) >> n
```
mbr0wn, line 51: I think Pylint prefers the original version.

mbr0wn: In fact, can you please remove all whitespace changes from this diff? They make our Pylint scores worse, and should not be mixed up with a new class.

mbr0wn (changes requested):

gr-digital/python/digital/qam.py
Joseph-Cannaday: I'm not really sure what you are referring to here. There are 2 blank lines between defs in the current code in the master branch

Joseph-Cannaday: Nevermind, I believe I've fixed them all now.

mormj:
@Joseph-Cannaday - thank you for the contribution.  In the past we have tried to move away from these "type_1" modulations and toward Constellation objects.  Depending on what you are trying to do the functionality may or may not exist in this main mechanism.  Is there a particular reason you are using `generic_mod`? 

See: 
https://github.com/gnuradio/gnuradio/issues/1956
https://github.com/gnuradio/gnuradio/commit/9fa00984d8e5d04bf911f8d6d1ca818ca8e9fd1f
as to why most of this type of code was removed in the past

Joseph-Cannaday:
> @Joseph-Cannaday - thank you for the contribution.  In the past we have tried to move away from these "type_1" modulations and toward Constellation objects.  Depending on what you are trying to do the functionality may or may not exist in this main mechanism.  Is there a particular reason you are using `generic_mod`? 
> 
@mormj This is mainly because I was working with some older gr documentation from a professor for a project. I noticed that QAM modulation just didn't work when I used the corresponding blocks, yet the blocks themselves hadn't been deprecated(at least in 3.8). Essentially the main reason is for visualization purposes/ high level simulation, but if y'all are moving away from that kind of thing that's fine too.

mormj:
Take a look at the constellation_soft_decoder example under gr-digital/examples/demod.  This uses a constellation object to define the constellation, and can be swapped to QAM, or you can load a custom QAM using the Variable Constellation selection.  

image

mormj:
Closing this PR for now so as to not move back in the direction of the type 1 modulations.  @Joseph-Cannaday - again thank you for the contribution here and please open up any issues or PRs in the future as you see fit.

Issue #4251, opened 2021-02-17
GRC hiding block ids for blocks that should never hide them


mormj:
With the show block ids option checked, blocks that have the show_id flag should always show the blockid on the canvas.

## Steps to reproduce
1) Select "Show All Block IDs"
2) Double click on a variable
3) Press OK on the properties dialog
![image](https://user-images.githubusercontent.com/34754695/108255868-cc97d600-712a-11eb-836e-1c35ab7871df.png)
Variable is shown with no id - just a value.

possibly related to: https://github.com/gnuradio/gnuradio/pull/4080 ?  <-- (edit) - not related to this per @jacobagilbert

PR #4327, opened 2021-03-06
Revert " grc: restore pre-3.8 block id behavior"

mormj:
This reverts commit 1fbf9ba89c643f478024d6af797e0a0dbbdf5ba3.

It resolves https://github.com/gnuradio/gnuradio/issues/4251 but also reverts behavior back to the uglier 3.8 behavior where the canvas is littered with block IDs when that option is checked.

### DIFF
--- a/grc/gui/PropsDialog.py
+++ b/grc/gui/PropsDialog.py
@@ -175,7 +175,7 @@ class PropsDialog(Gtk.Dialog):
 
                 for param in self._block.params.values():
                     if force_show_id and param.dtype == 'id':
-                        param.hide = 'part'
+                        param.hide = 'none'
                     # todo: why do we even rebuild instead of really hiding params?
                     if param.category != category or param.hide == 'all':
                         continue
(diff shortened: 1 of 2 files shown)


jacobagilbert:
Hmmm. Obviously some blocks should never have their IDs hidden but I'm not sure reverting this is the best way to address that. I'll take a look when I'm not on mobile.

jacobagilbert:
I reverted #4080 which did not resolve this, so I don't think its related to that. Rolling back the commit does fix this but the result is really kind of awful.  I think this is [what @skoslowski meant in this comment](https://github.com/gnuradio/gnuradio/pull/3824#pullrequestreview-505896880) and I did not sufficiently address it.

I'd like to try to actually fix this. In the mean time Reloading blocks will reset.

mormj:
I agree that your design to have partial hiding of the ids is the better solution - hopefully we can get to that point without the current side effect.

Issue #4317, opened 2021-03-04
Low pass, High pass convenience blocks not worked properly with QT GUI Range 


patel999jay:
I was testing GNURadio Filter Introduction Example and found out that ` GRC convenience blocks`  for the filter tap generation don't have working callbacks as @bistromath  suggested in my post in the discussion forum. 

**Initial Testing:**

![image](https://user-images.githubusercontent.com/5512610/109914319-1550a280-7c86-11eb-90fb-32906f3a31ed.png)

@patchvonbraun : suggested me that it should work with my existing setup as  from the YAML it looks like they should work, but unfortunately didn't work(refer below screenshot).

![image](https://user-images.githubusercontent.com/5512610/109914425-53e65d00-7c86-11eb-8577-00e12e02ad18.png)

But when i bypass those `convenience blocks` and directly replace `lp_taps` in the low-pass filter with `firdes.low_pass(1, samp_rate, cutoff_high, transition, firdes.WIN_HANN)`  - The actual filter generation callbacks work . I confirmed @bistromath suggestions is correct that

> for instance, changing the parameters with a QT GUI Range slider will work just fine. It's just the convenience blocks which don't work.

![image](https://user-images.githubusercontent.com/5512610/109914829-28b03d80-7c87-11eb-9334-7d0c7f85dbd1.png)

[filter_taps_modified.pdf](https://github.com/gnuradio/gnuradio/files/6081062/filter_taps_modified.pdf)

**Output QT GUI:**

![image](https://user-images.githubusercontent.com/5512610/109914904-47aecf80-7c87-11eb-9486-012ca478f475.png)

**Issue may be related to pybind11 - GRC 3.9**

ChristopheSeguinot:
The issue is present in GR>=3.8.2 (probably 3.8.0 too)
Looking at callbacks_all (generated inside top_block.pyfor the present flowgraph), I discovered that no callback is related to "xx filter taps blocks" (the RRC filter taps not present in the original flowgraph is also buggy)

in Yaml /gr-filters/grc/variable_low_pass_filter_taps.block.yml I don't see any callback !

I've just discovered what GNURadio callbacks are, and don't understand the difference between template and CPP_/usr/share/gnuradio/grc/blocks/variable_low_pass_filter_taps.block.yml so I Feel not able to fix this issue myself...

Regards, Christophe

willcode:
Callbacks added, so this should be resolved.

PR #4329, opened 2021-03-08
gr_filter : fix variable filter taps blocks

ChristopheSeguinot:

This partly fix an issue reported in [discuss gnuradio](https://lists.gnu.org/archive/html/discuss-gnuradio/2021-03/msg00008.html) and in #4317
It appears that the following variable taps blocks do not had callbacks : 
- /gnuradio/gr-filter/grc/variable_band_pass_filter_taps.block.yml
- /gnuradio/gr-filter/grc/variable_band_reject_filter_taps.block.yml
- /gnuradio/gr-filter/grc/variable_high_pass_filter_taps.block.yml
- /gnuradio/gr-filter/grc/variable_low_pass_filter_taps.block.yml
- /gnuradio/gr-filter/grc/variable_rrc_filter_taps.block.yml 

The fix has been tested under GR 3.8.2 and 3.9 using flowgraph 
[filter_taps_test.grc.txt](https://github.com/gnuradio/gnuradio/files/6100586/filter_taps_test.grc.txt). 

However, as I don't know what are cpp_templates callbacks , when they are used/needed ... I did not add any cpp_template callbacks. May be this need more work ?

All this can be backported to maint-3.9 and 3.8 excepted in one case described below

**Attention when backporting to 3.8** 
	${type} variable is used in band_pass and band_reject filters in master branch but
	${type} was introduced for band_pass in maint-3.8
	${type} was introduced for band_reject in maint-3.9

So when backporting to 3.8 replace  `${type}` by `band_reject` FOR band_reject filter NOT FOR band_pass
maint-3.8 call back for variable_band_reject_filter_taps.block.yml
   ` - self.set_${id}(firdes.band_reject(${gain}, ${samp_rate}, ${low_cutoff_freq}, ${high_cutoff_freq}, ${width}, ${win}, ${beta}))`
maint-3.9 and master  call back for variable_band_reject_filter_taps.block.yml
  `  - self.set_${id}(firdes.${type}(${gain}, ${samp_rate}, ${low_cutoff_freq}, ${high_cutoff_freq}, ${width}, ${win}, ${beta}))`

### DIFF
--- a/gr-filter/grc/variable_band_pass_filter_taps.block.yml
+++ b/gr-filter/grc/variable_band_pass_filter_taps.block.yml
@@ -44,6 +44,8 @@ templates:
     var_make: |-
         self.${id} = ${id} = firdes.${type}(${gain}, ${samp_rate}, ${low_cutoff_freq}, \
         ${high_cutoff_freq}, ${width}, ${win}, ${beta})
+    callbacks:
+    - self.set_${id}(firdes.${type}(${gain}, ${samp_rate}, ${low_cutoff_freq}, ${high_cutoff_freq}, ${width}, ${win}, ${beta}))
 
 cpp_templates:
     includes: ['#include <gnuradio/filter/firdes.h>']
--- a/gr-filter/grc/variable_band_reject_filter_taps.block.yml
+++ b/gr-filter/grc/variable_band_reject_filter_taps.block.yml
@@ -44,6 +44,8 @@ templates:
     var_make: |-
         self.${id} = ${id} = firdes.${type}(${gain}, ${samp_rate}, ${low_cutoff_freq},\
         ${high_cutoff_freq}, ${width}, ${win}, ${beta})
+    callbacks:
+    - self.set_${id}(firdes.${type}(${gain}, ${samp_rate}, ${low_cutoff_freq}, ${high_cutoff_freq}, ${width}, ${win}, ${beta}))
 
 cpp_templates:
     includes: ['#include <gnuradio/filter/firdes.h>']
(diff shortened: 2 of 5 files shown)


mormj:
> However, as I don't know what are cpp_templates callbacks , when they are used/needed ... I did not add any cpp_template callbacks. May be this need more work ?

There is more work to get the cpp_templates in place, but that can be handled separately on an as-needed basis.  The python interface is the primary use case for GRC, and getting the cpp_templates working and tested is a more advanced feature - they are currently not consistently implemented through the codebase.

I think we can proceed here without the cpp templates, and those can be updated later by someone who is so inclined to do so.

ChristopheSeguinot:
Does a comment inside thesecfiles should inform future developper about the "need CC_templates callbacks"" ?

mormj:
> Does a comment inside thesecfiles should inform future developper about the "need CC_templates callbacks"" ?

@haakov - what is your feeling about this?  

I had assumed that additions of cpp templates would be added in batches as needed and not necessarily put the burden on every developer to add that functionality unless it is a simple, parallel change to the python.

willcode:
Is there some way we can mark blocks in the block list and/or in the flowgraph layout to say which ones are supported by which languages?

mormj:
that's a good idea - the flag is already there in the yml files.  Seems feasible to bring that out to the UI somehow.

mormj (approved):
Tested a couple of these by modifying the filter_taps example.  LGTM!

duggabe (approved):
LGTM :+1:

PR #4332, opened 2021-03-08
dtv/atsc: fix arithmetic error / replace magic num

mormj:
There was an extra factor of 12 in the output copy of viterbi decoder
when the noutput_items was > 1 (which doesn't happen all the time)
leading to a glitch in the output video.

Also, replaced the hardcoded 832 with the appropriate enum

Addresses the glitch introduced and discussed in #4243


drmpeg:
Isn't the "magic number" of 832 equal to ATSC_DATA_SEGMENT_LENGTH, not ATSC_MPEG_RS_ENCODED_LENGTH?

mormj:
doh!  you are correct.

drmpeg (approved):
Working good now.

PR #4333, opened 2021-03-08
uhd: Fix generation of FE correction functions

mbr0wn:
The DC offset and IQ balance functions were incorrectly implemented in
the GRC blocks: It was possible to force-enable, but not force-disable
them. Also, the underlying Python/C++ API allowed changing the actual
correction value, but the GRC did not.

Now, there are four options for IQ and DC:
- Default: Whatever UHD does by default
- Auto: Enable auto-correction
- Disabled: No correction
- Manual: Allows setting the correction value separately


mbr0wn:
what it looks like:

image

drmpeg:
Are the correction values runtime settable? That would be a lot more useful.

mbr0wn:
No, but that could be a useful addition. I would make that a separate change though (enhancement rather than bugfix).

mormj (approved):

PR #4334, opened 2021-03-08
gui : fix block can move outside top and left boundary

### LINKED ISSUE #4178: Can't access the blocks which has accidentally moved beyond left or top boundary
The new interface has the ability to create flow graphs of unlimited size by adaptive canvas expanding. However, the canvas can't expand to the left or up, and there is no "overshoot protection" for blocks being moved beyond the left or top boundary. This leads to losing access to the block groups which have been moved too far because the screen can not scroll to them being visible. I want to delete and re-create them, but I can't select them from GUI. Is there any way to edit the GRC file correctly without breaking it? Also, boundary protection needs to be added.
image

ChristopheSeguinot:

This fixes #4178 
Tested on GR 3.8.2 
If ever a block have been saved with negative coordinate, coordinates are sanitized when opening the flowgraph
This can be backported to maint-3.8 and maint 3.9 
As compared to Derek proposal, the coordinate sanitizing is made inside `scale(coor, reverse=False):` method which is only called in https://github.com/gnuradio/gnuradio/blob/master/grc/gui/canvas/block.py#L61

### DIFF
--- a/grc/gui/canvas/flowgraph.py
+++ b/grc/gui/canvas/flowgraph.py
@@ -362,7 +362,7 @@ class FlowGraph(CoreFlowgraph, Drawable):
         delta_coordinate = (max(delta_coordinate[0],-min_x), max(delta_coordinate[1], -min_y))
 
         # Move selected blocks     
-        for selected_block in self.selected_blocks():
+        for selected_block in blocks:
             selected_block.move(delta_coordinate)
             self.element_moved = True
 
--- a/grc/gui/canvas/flowgraph.py
+++ b/grc/gui/canvas/flowgraph.py
@@ -347,6 +347,21 @@ class FlowGraph(CoreFlowgraph, Drawable):
         Args:
             delta_coordinate: the change in coordinates
         """
+
+        # Determine selected blocks top left coordinate
+        blocks = list(self.selected_blocks())
+        if not any(blocks):
+            return False
+
+        min_x, min_y  = self.selected_block.coordinate
+        for selected_block in blocks:
+            x, y = selected_block.coordinate
+            min_x, min_y = min(min_x, x), min(min_y, y)
+
+        # Sanitize delta_coordinate so that blocks don't move to negative coordinate
+        delta_coordinate = (max(delta_coordinate[0],-min_x), max(delta_coordinate[1], -min_y))
+
+        # Move selected blocks     
         for selected_block in self.selected_blocks():
             selected_block.move(delta_coordinate)
             self.element_moved = True
--- a/grc/gui/canvas/flowgraph.py
+++ b/grc/gui/canvas/flowgraph.py
@@ -351,7 +351,7 @@ class FlowGraph(CoreFlowgraph, Drawable):
         # Determine selected blocks top left coordinate
         blocks = list(self.selected_blocks())
         if not blocks:
-            return False
+            return
 
         min_x, min_y  = self.selected_block.coordinate
         for selected_block in blocks:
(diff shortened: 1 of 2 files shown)


skoslowski:
`Utils.scale()` is used in other places too... I'd recommend to put this in the setter for Block.coordinate (it's usually not a good idea to hide fixes like that away).

Another option to solve this would be to allow negative coordinates during a move, but on mouse release, shift all blocks back in non-negative values and adjust the viewport accordingly

ChristopheSeguinot:
Sorry not having seen Utils.scale was used eldwhere (I tested it but confused scale() and scale_scalar())

>> Another option to solve this would be to allow negative coordinates during a move, but on mouse release, shift all blocks back in non-negative values and adjust the viewport accordingly
Yes, the ideal would be to; on mouse release sanitize coordinates so that all seleted/moved blocks keep their relative position (all shifted of the same amount).  
Sanitizing individually each block as I did, when several blocks are selected/moved outside windows, they are all put on  left or top and superposing each other which is not convenient.

mormj:
@ChristopheSeguinot - do you plan to update this PR based on the feedback from @skoslowski?  I tested out the changes and they work as advertised, but it seems the current implementation could have other side effects based on the guidance re: `Utils.scale()`. 
Another thing I noticed is that when dragging a group of blocks, when one block hits the edge the other blocks keep moving and bunch up.

ChristopheSeguinot:
This PR must be updated, but I need some time to work on this and I'm quite busy at present.
 
> Another thing I noticed is that when dragging a group of blocks, when one block hits the edge the other blocks keep moving and bunch up.
Yes this is annoying and must be avoided.

ChristopheSeguinot:
I finally implement this fix inside the move_selected(self, delta_coordinate) method. 
This method is continuously fired when a block or a selection of blocks is moved (the final coordinate is not set on mouse release, but on the last mouse_pressed move ). 
The block selection is no longer allowed to go above top and left of the window. 
I saw that the "options blocks" could sometimes go left of 0 (-1 or -2 pixel probably). I Only observe this when moving the options blocks alone without understanding why. 

The fix was tested on 3.8.2,  whilst the PR applies to master. 
This can and should probably be backported to 3.8 and 3.9

mormj (approved):
Tested this out (on top of master) and it works - block are prevented from going off-canvas and don't bunch up when they hit the edge.  Thanks for the fix!

grc/gui/canvas/flowgraph.py
```diff
@@ -347,6 +347,21 @@ def move_selected(self, delta_coordinate):
...
+        if not any(blocks):
+            return False
+
+        min_x, min_y  = self.selected_block.coordinate
+        for selected_block in blocks:
+            x, y = selected_block.coordinate
+            min_x, min_y = min(min_x, x), min(min_y, y)
+
+        # Sanitize delta_coordinate so that blocks don't move to negative coordinate
+        delta_coordinate = (max(delta_coordinate[0],-min_x), max(delta_coordinate[1], -min_y))
```
skoslowski, line 362:
```suggestion
        delta_coordinate = max(delta_coordinate[0],-min_x), max(delta_coordinate[1], -min_y)
```

skoslowski (changes requested):
Looking over the code changes, this should solve the issue. I only found some minor code-style nits.

dkozel:
@ChristopheSeguinot Can you make the small style tweaks that Sebastian has requested? I'd like to merge this. I can do them as a fixup if that's easiest.

ChristopheSeguinot:
It's done. 
Sorry for the delay, I'm not so familiar to Github and didn't realize it just need some click to accept Sebastien requests. 
Thanks Sebastien for cleaning my code.

skoslowski (approved):

devnulling (approved):

PR #4335, opened 2021-03-09
Fix for #3678

### LINKED ISSUE #3678: MSK symbol synchronizers are broken
Created re: the [discussion](https://lists.gnu.org/archive/html/discuss-gnuradio/2020-06/msg00173.html) on the list.

The older MSK timing recovery block works while the newer symbol synchronizer doesn't because the MSK timing recovery block deviates from the paper to conjugate the squared symbol **half** a sample delayed, rather than a full sample. This is equivalent to changing the nonlinearity in `timing_estimator_impl.cc` to:

`u = (d_input[0] * d_input[0] * conj(d_input[1] * d_input[1])) - (d_input[1] * d_input[1] * conj(d_input[2] * d_input[2]))`

However, this results in sampling at exactly the wrong time -- the above needs to be modified to delay by an additional half symbol period to lock onto the correct symbol phase:

`u = (d_input[1] * d_input[1] * conj(d_input[2] * d_input[2])) - (d_input[2] * d_input[2] * conj(d_input[3] * d_input[3]))`

The older timing recovery block's behavior appears to be a happy accident due to an order of operations bug: `d_dly_conj_2` is set after `d_dly_conj_1`, and thus isn't actually delayed, resulting in the nonlinearity using a delay of half a symbol period rather than a full symbol period.

I can't account for two things: why a working nonlinearity deviates from the one in the D'Andrea and Mengali paper, and why the timing estimator needs to be delayed an additional half a symbol period in order to arrive at the correct sampling instant.

I believe the latter can be answered by looking at the original paper -- there's no resampler in the output, and so the paper doesn't have to deal with answering the question of what phase to output. Also regarding the latter, I **think** that eq. 9.4.4 (from Synch. Techn. for Dig. Recv. p. 498) implies a one-half-symbol offset in the error detector since it asks for the previous and next half-symbol periods, not the current and 2x periods.

As for the former, I just don't know. I'm trying to work that out now.

bistromath:
This fixes #3678 (MSK timing error detector) for the "MSK Gen TED". The fix can be cherry-picked onto maint-3.8 and maint-3.9 as it is not an API change.

It does not change the "GMSK TED", which should be deprecated as it is redundant and does not work for its intended application, confusing users.

### DIFF
--- a/gr-digital/lib/timing_error_detector.cc
+++ b/gr-digital/lib/timing_error_detector.cc
@@ -311,8 +311,8 @@ float ted_generalized_msk::compute_error_cf()
 {
     gr_complex u;
 
-    u = (d_input[0] * d_input[0] * conj(d_input[2] * d_input[2])) -
-        (d_input[1] * d_input[1] * conj(d_input[3] * d_input[3]));
+    u = (d_input[1] * d_input[1] * conj(d_input[5] * d_input[5])) -
+        (d_input[3] * d_input[3] * conj(d_input[7] * d_input[7]));
 
     return gr::branchless_clip(u.real(), 3.0f);
 }
@@ -321,8 +321,8 @@ float ted_generalized_msk::compute_error_ff()
 {
     float u;
 
-    u = (d_input[0].real() * d_input[0].real() * d_input[2].real() * d_input[2].real()) -
-        (d_input[1].real() * d_input[1].real() * d_input[3].real() * d_input[3].real());
+    u = (d_input[1].real() * d_input[1].real() * d_input[5].real() * d_input[5].real()) -
+        (d_input[3].real() * d_input[3].real() * d_input[7].real() * d_input[7].real());
 
     return gr::branchless_clip(u, 3.0f);
 }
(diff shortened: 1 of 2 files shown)


willcode:
What makes sure that there are at least 8 samples? Not a new thing ... I just traced it back to ntaps, and I think the assumption is that the user won't provide a ridiculously small filter?

bistromath:
Internal SPS and history are set by the instantiation of ted_generalized_msk in timing_error_detector.h:
`ted_generalized_msk()
        : timing_error_detector(
              TED_DANDREA_AND_MENGALI_GEN_MSK, 4, 8, false, false, constellation_sptr())
`

willcode (approved):
Looks safe. And I vaguely understand why it's better :-)

bistromath:
DCO fixed with amended push.

mbr0wn:
@daniestevez or @awalls-cx18, do you want to take a look?

daniestevez:
Thanks for the mention, Martin.

I'm not so familiar D'Andrea and Mengali's paper. The fix that Andy has proposed to the current TED looks good in the simulations, and I've checked that Nick has implemented exactly that in this PR. So I can say this looks good to me and can be merged.

Even though I'm convinced that this works, I would like to understand "why". When I have time, I'll try to go through D'Andrea and Mengali's paper and read the comments in the issue carefully. I'll probably learn a lot! But please don't wait for me to merge this. I'm not certain when I'll be able to do this.

mormj (approved):

PR #4340, opened 2021-03-10
github: Use autogenerated containers instead of manually tagged ones

mbr0wn:


mormj (approved):

mbr0wn:
Ugh those core dumps in Debian look ugly.

mbr0wn:
OK I backed out the Debian container. We can fix that separately.

PR #4356, opened 2021-03-11
gr_filter_design fails to load previously saved csv files

dl1ksv:
For  FIR filters:
loading a saved csv file fails with

  File "/usr/local/gnuradio/lib64/python3.9/site-packages/gnuradio/filter/filter_design.py", line 2052, in action_open_dialog
    self.gui.filterDesignTypeComboBox.setCurrentIndex(int(params["wintype"]))
ValueError: invalid literal for int() with base 10: 'win_type.WIN_HAMMING'

  File "/usr/local/gnuradio/lib64/python3.9/site-packages/gnuradio/filter/filter_design.py", line 2052, in action_open_dialog
    self.gui.filterDesignTypeComboBox.setCurrentIndex(int(params["wintype"]))
ValueError: invalid literal for int() with base 10: 'win_type.WIN_HANN'

depending on the used window.

Saving and loading IIR filters work.
This pr fixes this issue.

### DIFF
--- a/gr-filter/python/filter/design/filter_design.py
+++ b/gr-filter/python/filter/design/filter_design.py
@@ -2047,9 +2047,10 @@ class gr_plot_filter(QtGui.QMainWindow):
             self.gui.filterGainEdit.setText(str(params["gain"]))
 
             # Set up GUI parameters for each filter type.
+            checkbox_entry=int(eval('fft.window.'+params["wintype"]))
             if(params["filttype"] == "lpf"):
                 self.gui.filterTypeComboBox.setCurrentIndex(0)
-                self.gui.filterDesignTypeComboBox.setCurrentIndex(int(params["wintype"]))
+                self.gui.filterDesignTypeComboBox.setCurrentIndex(checkbox_entry)
 
                 self.gui.endofLpfPassBandEdit.setText(str(params["pbend"]))
                 self.gui.startofLpfStopBandEdit.setText(str(params["sbstart"]))
@@ -2058,7 +2059,7 @@ class gr_plot_filter(QtGui.QMainWindow):
                     self.gui.lpfPassBandRippleEdit.setText(str(params["ripple"]))
             elif(params["filttype"] == "bpf"):
                 self.gui.filterTypeComboBox.setCurrentIndex(1)
-                self.gui.filterDesignTypeComboBox.setCurrentIndex(int(params["wintype"]))
+                self.gui.filterDesignTypeComboBox.setCurrentIndex(checkbox_entry)
 
                 self.gui.startofBpfPassBandEdit.setText(str(params["pbstart"]))
                 self.gui.endofBpfPassBandEdit.setText(str(params["pbend"]))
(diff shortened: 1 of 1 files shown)


mormj:
@dl1ksv - how can I reproduce this issue? - If I save and load in the filter_design gui, it seems to work fine.

dl1ksv:
@mormj :
gr_filter_design
Select Hann Window
Push Design
File save
Restart gr_filter_design
load the saved filter.

The gui shows no errors, but the wrong window type and in your teminal window you get

  File "/usr/local/gnuradio/lib64/python3.9/site-packages/gnuradio/filter/filter_design.py", line 2052, in action_open_dialog
    self.gui.filterDesignTypeComboBox.setCurrentIndex(int(params["wintype"]))
ValueError: invalid literal for int() with base 10: 'win_type.WIN_HANN'

mormj (approved):

mormj (changes requested):
So it looks like the change is to the saved file:
```
josh@josh-m15:/share/gnuradio/gr310/src/gnuradio/build$ diff hann_orig hann
4c4
< wintype,win_type.WIN_HANN
---
> wintype,1
```
I'm wondering if there is a way to make the file load recognize the enum rather than saving to an int

mormj:
In 3.9 and beyond the win types enums moved away from firdes and into fft.window

dl1ksv:
Yes, saving the wintype as int  corresponding to the enum is the result of this patch.
I did not find an elegant way to treat the string win_type.WIN_HANN as enum type.
The only way I found was something like this
wintype=win_type.WIN_HANN
int(eval("fft.window."+wintype))

dl1ksv:
@mormj I now derive the index from the stored wintype string.
You can build str(enum) which is used when saving to csv or int(enum), but I found nothing like enum(string), so I  used eval.
It works , but I'd prefer to store the int value of the enum instead the string.

mormj:
I'm getting another error here when saving the file
```
  File "/share/gnuradio/gr310/lib/python3/dist-packages/gnuradio/filter/filter_design.py", line 1981, in action_save_dialog
    for k in list(self.params.keys()):
AttributeError: 'list' object has no attribute 'keys'
```

I think saving the enum value is fine - but being able to load previously saved files that have the enum string would be good as well.

dl1ksv:
This has nothing to do with my pr. I get this error  on an unpatched system if I try to save to a file before a filter was calculated.
This means starting gr_filter_design and saving to a file at once I get this error. Pressing Design before you can save the filter without errors.

So did you calculate a filter before saving ? If not, I try to handle this case in my pr.

As I handle the wintype string from the saved csv file, I can read previously saved files.

mormj:
> So did you calculate a filter before saving ?

Nope - I did not.  Guess that is a separate issue I should have checked on the unpatched version.  Will give it another go doing things the right way

mormj (approved):
Tested and fix seems reasonable

mormj:
@willcode - not sure if this is an issue on 3.8 because of the change in window type enum.

mormj:
QA test failures unrelated.  Merging

PR #4367, opened 2021-03-12
Packaging: Add debian/fedora packaging in-tree

mormj:
The goal of this PR is to be able to trigger automated packaging builds based on Mait's Debian packaging.  It is unlikely that we would want packaging files to be generated every pull request or push onto master.  Ideally this will be expanded to build .deb files as build artifacts as well that can be downloaded from github.

Process:
- ~~Push from master to pkg-master will trigger the gh-action (do the same for pkg-maint-3.x)~~
- Actions set up to build on demand
- Debian source package is generated and uploaded to launchpad
- Fedora srpms uploaded to COPR

Notes:
- Right now all of modules are rolled up into a single gnuradio.deb
- With launchpad it is not necessary to reduce the number of .debs, but if we host them here it could become unwieldy
- Need to change the launchpad target and credentials to the gnuradio launchpad


willcode:
TODO: add libiio, libad9361, libsoapysdr (and Fedora counterparts) when these get merged.

.packaging/debian/control
```diff
@@ -0,0 +1,383 @@
+Source: gnuradio
```
mbr0wn, line 1: I know there was some discussion on chat about keeping these files separate from the main repo, but we'll never be able to keep this file in sync with the master (or whatever) branch if we do that. So I'm glad this is in-tree!

.packaging/debian/control
```diff
@@ -0,0 +1,383 @@
...
+	            python3-gi-cairo,
+               python3-lxml,
+               python3-mako,
+               python3-numpy,
+               python3-opengl,
+	            python3-pyqt5 [!hurd-i386],
+               python3-scipy,
+	            python3-yaml,
+               python3-zmq [!hurd-i386],
+               python3-six,
```
mbr0wn, line 61: No longer a dependency of master branch

.packaging/debian/control
```diff
@@ -0,0 +1,383 @@
...
+               libboost-date-time-dev,
+               libboost-dev,
+               libboost-filesystem-dev,
+               libboost-program-options-dev,
+               libboost-regex-dev,
+               libboost-system-dev,
+               libboost-test-dev,
+               libboost-thread-dev,
+               libcodec2-dev (>=0.8),
+               libcppunit-dev (>=1.9.14),
```
mbr0wn, line 22: No longer a dependency

.packaging/debian/docs/dial_tone.1
```diff
@@ -0,0 +1,15 @@
+.TH DIAL_TONE "1" "May 2012" "DIAL_TONE 3.6.0" "User Commands"
```
mbr0wn, line 1: This (and a bunch of others) are also in `docs/man`. Maybe we can do something clever to not duplicate those?

.packaging/fedora/gnuradio.spec
```diff
@@ -0,0 +1,873 @@
...
+BuildRequires:	python3-cairo
+BuildRequires:	python3-click-plugins
+BuildRequires:	python3-gobject
+BuildRequires:	python3-numpy
+BuildRequires:	python3-pyyaml
+BuildRequires:	python3-lxml
+BuildRequires:	python3-mako
+BuildRequires:	python3-qt5-devel
+BuildRequires:	python3-scipy
+BuildRequires:	python3-six
```
mbr0wn, line 73:
```suggestion
```

.packaging/fedora/gnuradio.spec
```diff
@@ -0,0 +1,873 @@
...
+BuildRequires:	xmlto
+BuildRequires:	zeromq-devel
+BuildRequires:	python3-gobject
+BuildRequires:	pybind11-devel
+BuildRequires:	volk-devel
+BuildRequires:	libsndfile-devel
+
+Requires:	python3-%{name} = %{version}-%{release}
+Requires:	python3-numpy
+Requires:	python3-cheetah
```
mbr0wn, line 90:
```suggestion
```

.packaging/fedora/gnuradio.spec
```diff
@@ -0,0 +1,873 @@
...
+Requires:	python3-thrift
+%if ! 0%{?rhel}
+Requires:	python3-pyopengl
+%endif
+Requires:	python3-pyyaml
+Requires:	python3-gobject
+Requires:	python3-six
```
mbr0wn, line 97:
```suggestion
```

.github/workflows/pkg-debian.yml
```diff
@@ -0,0 +1,63 @@
...
+    name: ${{ matrix.distro }}
+    container:
+      image: ${{ matrix.containerid }}
+      volumes:
+        - build_data:/build
+      options: --cpus 2
+    steps:
+    - uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
```
mbr0wn, line 34:
```suggestion
        fetch-depth: 0
        ref: ${{ github.event.inputs.build_branch }}
```

.github/workflows/pkg-debian.yml
```diff
@@ -0,0 +1,63 @@
+name: 'Package-Debian'
+on:
+  push:
+    branches:
+      - pkg-master
```
mbr0wn, line 5:
```suggestion
  workflow_dispatch:
    build_branch:
      description: 'Branch to build from'
      required: true
      default: 'pkg-master'
```

mbr0wn (commented):
I think my YAML suggestions make it such that you can build on demand, rather than on push. If you want that.

mbr0wn:
There's a trigger that occurs when a release is published: https://docs.github.com/en/actions/reference/events-that-trigger-workflows#release

Might come in handy.

.packaging/debian/docs/dial_tone.1
mormj: Yes, seems like there is a set of man files already in tree I was not aware of.

mormj:
Looks like you can also do scheduled events with GH actions

mbr0wn:
> Looks like you can also do scheduled events with GH actions

Yep -- my dream of having automated weekly builds for master branch binaries is getting more complate WUAHAHAHA

mormj:
There may be some things that we can clean up in the packaging defs, but I've tested this on my fork (has to be pushed over to master to be able to build off the master branch via the `workflow_dispatch`)

The only other major thing I know needs to be changed is the dput.cf which specifies which ppa to push to - currently points to mormj, but need to get the gpg set up with launchpad to push directly there.

mbr0wn:
> There may be some things that we can clean up in the packaging defs, but I've tested this on my fork (has to be pushed over to master to be able to build off the master branch via the `workflow_dispatch`)
> 
> The only other major thing I know needs to be changed is the dput.cf which specifies which ppa to push to - currently points to mormj, but need to get the gpg set up with launchpad to push directly there.

Do we want to have a discussion on package naming? I.e., do we want separate PPAs for 3.8 and 3.9, or do we want to have package names that reflect the version (like what Maitland does)?

mormj:
For the "build off of branch" packages I had assumed we would have separate PPAs because it would be easier to manage.
For the "build off of tarball" packages I think that is a valid discussion, and the current approach of separate PPAs has created issues.  It would probably be good to have a single `gnuradio-releases` PPA for official releases with properly named packages

mormj:
It's still probably not perfect, but I'm going to go ahead and merge so that the buttons show up to build packages

PR #4368, opened 2021-03-12
 Replace boost::filesystem with std::filesystem (updated)

mormj:
Original Author: @ZackerySpytz - https://github.com/gnuradio/gnuradio/pull/4280

Applied comments from the github conversation

### DIFF
--- a/cmake/Modules/GnuradioConfig.cmake.in
+++ b/cmake/Modules/GnuradioConfig.cmake.in
@@ -16,7 +16,6 @@ find_dependency(MPLIB)
 set(BOOST_REQUIRED_COMPONENTS
     date_time
     program_options
-    filesystem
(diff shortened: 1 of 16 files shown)


mormj:
@ryanvolz - I had to put back the .string() to get it to compile on Centos/Debian.  Will this cause an issue for Windows, or does there need to be some conditional logic?

ryanvolz:
> @ryanvolz - I had to put back the .string() to get it to compile on Centos/Debian. Will this cause an issue for Windows, or does there need to be some conditional logic?

Hmm. It should work fine with the .string(), but I'm curious why it doesn't compile because it seems indicative that the function is not returning a std::string as expected. It's probably not worth holding up just to figure that out though.

PR #4381, opened 2021-03-13
Sync C++ vs C style include macros with master branch

willcode:
Differences in include files and ordering make backporting harder.


willcode:
@marcusmueller @ThomasHabets if you could check to make sure I didn't do anything silly. This is supposed to have no effect at all, other than to make backporting easier in the future.

ThomasHabets (approved):

Issue #4324, opened 2021-03-04
Cannot automate gr_modtool add with cli params


blsmit5728:
We Ran into an issue trying to automate a 3.7->3,8 conversion, so we were going to just "rebuild" the OOT Module with automated gr_modtool and copy-ing in the code. But we found that we couldn't get past the argument-list parameter no matter what we did.

the tool thinks that ["arglist"] is not None so if you provide a list you'll always be hit with the prompt.

reproduce:
`gr_modtool add -t general --copyright me -l cpp --argument-list "test" --add-python-qa --add-cpp-qa mycoolblockname`

To Fix, removed "not" from the if statment.

[Line Error](https://github.com/gnuradio/gnuradio/blob/08f188579761aa5300457ad2083405833174b223/gr-utils/modtool/cli/add.py#L122)

PR #4384, opened 2021-03-14
gr-utils: modtool/cli/add: check for arglist inverted

willcode:
Bug reported and located by Brandon Smith.

`self.info['arglist']` defaults to `""` so it can never be `None`. And the sense of the check is inverted.

Closes https://github.com/gnuradio/gnuradio/issues/4324

### DIFF
--- a/gr-utils/modtool/cli/add.py
+++ b/gr-utils/modtool/cli/add.py
@@ -119,7 +119,7 @@ def get_copyrightholder(self):
 
 def get_arglist(self):
     """ Get the argument list of the block to be added """
-    if self.info['arglist'] is not None:
+    if not self.info['arglist']:
         self.info['arglist'] = click.prompt(click.style(
             'Enter valid argument list, including default arguments: \n',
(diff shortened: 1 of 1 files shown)


mormj (approved):
Tested this out and it is definitely an improvement.  The only thing that doesn't work is an empty argument list: `--argument-list ""` will still prompt for args, but `--argument-list " "` will not

willcode:
More work needed then. The cli part will have to default to `None` and convert that to `''` for the core part, since it's used verbatim there.

dl1ksv:
We should set the default for the argument-list to None in the cli  part and only request "a valid argument list" if the argument-list is None. So you don't have to convert it to ''
If  the argument-list contains a string , use this string. If this string is empty, you'll get an empty argument list.

dl1ksv:
As I'm not very familiar with the git gui I pt an additional pr

mormj:
Closing in favor of #4401

PR #4386, opened 2021-03-14
gr-blocks: Fix Remove DC Spike AutoSync block.

drmpeg:

### DIFF
--- a/gr-blocks/lib/correctiq_auto_impl.cc
+++ b/gr-blocks/lib/correctiq_auto_impl.cc
@@ -123,11 +123,13 @@ float correctiq_auto_impl::get_gain() const { return d_gain; }
 
 void correctiq_auto_impl::set_freq(double new_value)
 {
+    d_freq = new_value;
     trigger_resync("Frequency change detected");
 }
 
 void correctiq_auto_impl::set_gain(float new_value)
 {
+    d_gain = new_value;
     trigger_resync("Gain change detected");
 }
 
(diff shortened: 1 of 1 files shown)


gr-blocks/lib/correctiq_auto_impl.cc
```diff
@@ -123,11 +123,13 @@ float correctiq_auto_impl::get_gain() const { return d_gain; }
 
 void correctiq_auto_impl::set_freq(double new_value)
 {
+    d_freq = new_value;
```
mbr0wn, line 126: What's the purpose of these?

drmpeg: I don't really know. You'll have to ask @ghostop14. I guess it's some convenience placeholder.

When I was repairing the block, it seemed like setting the variables had been omitted, so I put those lines in.

mbr0wn:
This PR is in a state where we can merge it. I'll leave it open for another day or so in case @ghostop14 or someone else wants to chime in.

gr-blocks/lib/correctiq_auto_impl.cc
ghostop14: Hi folks, I just looked at the current implementation and I think as long as the synchronized flag gets set to false, it can be removed.  Originally I probably checked that the value of freq there actually changed to save an update, but doesn't look like I'm doing that anymore.  So looks safe to pull it.  Same looks true for d_gain.

drmpeg: But don't you want the values returned by get_freq() and get_gain to be what you set them to with set_freq() and set_gain()?

ghostop14: True.  So maybe just leave it as-is.

PR #4389, opened 2021-03-15
uhd: Fix Qt apps (XInitThreads call order, terminate handling)

mbr0wn:
uhd_fft and uhd_siggen_gui are standalone apps that, a long time ago,
were spun off from GRC apps. In the last years, there were some fixes to
Qt apps that haven't yet made it into these two UHD apps:

- 0cc238fb: Better terminate handling
- 40f5e6cd: XInitThreads() as early as possible. This was accidentally
  moved downwards as part of refactoring in 77871a9b.


mormj (approved):
Changes look good and the apps run fine.

PR #4391, opened 2021-03-15
uhd: Replace multi_usrp::recv_async_msg() with tx_stream version (3.8 version)

mbr0wn:
The multi_usrp behaviour is deprecated starting in UHD 4.0.
Note: For UHD 3.5.5, we use the old API because the new one did not yet
exist.

This commit also removes some usages of Boost from usrp_sink_impl that
did not have headers to go along with them.

### DIFF
--- a/gr-uhd/lib/usrp_sink_impl.cc
+++ b/gr-uhd/lib/usrp_sink_impl.cc
@@ -23,8 +23,11 @@
 #include "gr_uhd_common.h"
 #include "usrp_sink_impl.h"
 #include <gnuradio/io_signature.h>
+#include <boost/thread/thread.hpp> // for Boost thread interruption disabling
+#include <chrono>
 #include <climits>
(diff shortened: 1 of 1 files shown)


willcode:
@mbr0wn did you mean to add a boost include (threading)?

willcode:
CI died on clang-format after a couple hours. Built OK on debian-10 and ubuntu-18.04 CI locally, so merging.

mbr0wn:
> OK, required for sleep, and we are using boost threads in 3.8. Or rather in the older UHD that is supported by 3.8?

It was required for the thread interruption disabling (see also my comment at the top of the file). To be honest, I mixed IWYU and Boost cleanup with the rest of the code which isn't great, but since this is 3.8 branch I thought that was OK.

mbr0wn:
To be clear, this was not required for older UHD. UHD should handle its own includes.

PR #4392, opened 2021-03-15
dtv/atsc: create separate plinfo stream ports

mormj:
Rather than adding/searching for tags, just put plinfo on its own
streaming port.  This should be more efficient than putting tags on
every item and easier to inspect/debug.

So now, the receive chain looks like this:
image

### DIFF
--- a/gr-dtv/lib/atsc/atsc_deinterleaver_impl.cc
+++ b/gr-dtv/lib/atsc/atsc_deinterleaver_impl.cc
@@ -27,8 +27,10 @@ atsc_deinterleaver::sptr atsc_deinterleaver::make()
 atsc_deinterleaver_impl::atsc_deinterleaver_impl()
     : gr::sync_block(
           "atsc_deinterleaver",
-          io_signature::make(1, 1, ATSC_MPEG_RS_ENCODED_LENGTH * sizeof(uint8_t)),
-          io_signature::make(1, 1, ATSC_MPEG_RS_ENCODED_LENGTH * sizeof(uint8_t))),
+          io_signature::make2(
+              2, 2, ATSC_MPEG_RS_ENCODED_LENGTH * sizeof(uint8_t), sizeof(plinfo)),
+          io_signature::make2(
+              2, 2, ATSC_MPEG_RS_ENCODED_LENGTH * sizeof(uint8_t), sizeof(plinfo))),
       alignment_fifo(156)
 {
     m_fifo.reserve(s_interleavers);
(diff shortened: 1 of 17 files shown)


gr-dtv/lib/atsc/atsc_deinterleaver_impl.cc
```diff
@@ -27,8 +27,10 @@ atsc_deinterleaver::sptr atsc_deinterleaver::make()
 atsc_deinterleaver_impl::atsc_deinterleaver_impl()
     : gr::sync_block(
           "atsc_deinterleaver",
-          io_signature::make(1, 1, ATSC_MPEG_RS_ENCODED_LENGTH * sizeof(uint8_t)),
-          io_signature::make(1, 1, ATSC_MPEG_RS_ENCODED_LENGTH * sizeof(uint8_t))),
+          io_signature::make2(
+              2, 2, ATSC_MPEG_RS_ENCODED_LENGTH * sizeof(uint8_t), sizeof(plinfo)),
```
ThomasHabets, line 31: Not only is the code with this PR not endian safe, but now also padding/alignment/size unsafe for `plinfo`.

But that may be fine, since the file sink blocks just treat their input as bytes anyway, so this was never endian safe.

Is it a goal that stream dumps (file sink/source), and what's sent over zeromq should be portable? Thinking about it more now I think the answer must be "no", because performance. Or rather, to leave it up to the user to put in `htonl` blocks where appropriate in their graphs.

That said, now the `get_tag_value` and `from_tag_value` functions can be deleted in this PR, right?

ThomasHabets (changes requested):

gr-dtv/lib/atsc/atsc_deinterleaver_impl.cc
mormj: My main goal for these blocks has been to create hardware accelerated versions of some of them in the newsched (gr 4.0) framework.  The original structs interfered with copying larger contiguous buffers into device memory without intermediate buffers.  And then tags for the plinfo made it harder to debug (we probably need some better file based debug mechanisms for tags) and streams seemed more efficient as well.  
Endian safety doesn't seem like it has been a concern throughout the codebase - wouldn't the original struct based streams have the same issues?  Probably leaving it to the user with `htonl` blocks makes sense in the general case.
But you are right, those 2 functions can be deleted now.

ThomasHabets: Yeah making gnuradio endian safe is out of scope here. It just isn't, now, like I said.

ThomasHabets (approved):

drmpeg (approved):
Looks good here. Performance is definitely better than the previous version.

Issue #4396, opened 2021-03-15
QT GUI: Time sink segfault


jacobagilbert:
The master branch QT GUI Time sink segfaults in complex PDU mode when when the number of elements in the PDU does not match the Number of Points set in the block config.

The Number of Points entry box is hidden when in either of the PDU modes are configured as the input type, and they should not be used to set behavior which should be controlled by the number of elements (and is for float type PDUs it appears).

Flowgraph: 

![image](https://user-images.githubusercontent.com/7121282/111231522-0daacb00-85af-11eb-8d3f-5a7d8c3cc8e4.png)

[pdu_test.grc.txt](https://github.com/gnuradio/gnuradio/files/6144952/pdu_test.grc.txt)

jacobagilbert:
I can confirm that after reverting 4b7006db (and 371ea713) the issue is gone, so its somewhere in there. 

I don't think this is severe enough to warrant rolling these back, @ThomasHabets if you have a few minutes to take a look at this I'd appreciate it. I'll post here if I make any progress.

ThomasHabets:
Yeah, my bad. PR coming in a few minutes.

ThomasHabets:
PR #4399 in flight.

Thanks for the minimal test case to reproduce!

PR #4399, opened 2021-03-16
qtgui: Fix segfaulting overflow

ThomasHabets:
Problem introduced in 4b7006db76b570e4d916e263301333d2f4d2a2df.

It appears to be the only instance of this mistake in that commit.

Fixes: #4396 
Signed-off-by: Thomas Habets <habets@google.com>

### DIFF
--- a/gr-qtgui/lib/time_sink_c_impl.cc
+++ b/gr-qtgui/lib/time_sink_c_impl.cc
@@ -56,10 +56,6 @@ time_sink_c_impl::time_sink_c_impl(int size,
     if (nconnections > 12)
         throw std::runtime_error("time_sink_c only supports up to 12 inputs");
 
-    // setup PDU handling input port
-    message_port_register_in(pmt::mp("in"));
-    set_msg_handler(pmt::mp("in"), [this](pmt::pmt_t msg) { this->handle_pdus(msg); });
-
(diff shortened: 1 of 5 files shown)


mbr0wn (approved):

mormj (approved):

jacobagilbert (approved):
I can confirm this fixes #4396 however the same issue is present in the QTGUI Waterfall sink in complex PDU mode as well. I tested the other QT sinks that accept complex PDUs (frequency sink, constellation sink, and eye sink) and they all operate fine fine. 

Moving the message port is not necessary but it works either way.

Approved, suggest working a fix

This can be merged as-is or a fix for the waterfall sink can be applied also.

ThomasHabets:
Thanks. I found the problem there too. Stand by.

jacobagilbert (approved):
Thanks for the super fast fix @ThomasHabets  - working great for me on my application flowgraph and in the test FG i put together on the issue.

Fixes #4402.

willcode:
@mormj @jacobagilbert Most of the manual memory management massacre missed maint-3.9. Check me on that, but I don't think this applies to maint-3.9.

jacobagilbert:
@willcode yes, 3.9 is both functional and does not have 4b7006d so I am removing the backport label.

PR #4401, opened 2021-03-16
gr_modtool add: Correctly handle arglist

### LINKED ISSUE #4324: Cannot automate gr_modtool add with cli params

dl1ksv:
See discussion in #4384 

gr_modtool add : no argument-list specified : ask user 
gr_modtool add --argument-list 'Some string' : use this string as arglist
    special case : empty string -> empty arglist

Fixes #4324

### DIFF
--- a/gr-utils/modtool/cli/add.py
+++ b/gr-utils/modtool/cli/add.py
@@ -27,7 +27,7 @@ from .base import common_params, block_name, run, cli_input, ModToolException
               help="File containing the license header for every source code file.")
 @click.option('--copyright',
               help="Name of the copyright holder (you or your company) MUST be a quoted string.")
-@click.option('--argument-list', default="",
+@click.option('--argument-list', default=None,
               help="The argument list for the constructor and make functions.")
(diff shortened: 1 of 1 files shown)


mormj (approved):

mormj:
Tested and it works as described.  Merging in favor of #4384

PR #4410, opened 2021-03-17
CI: Test and build conda packages for Linux, macOS, and Windows

ryanvolz:
This PR leverages the `conda-smithy` Python package to create and maintain CI infrastructure for building conda packages on Linux, macOS, and Windows (and potentially other architectures apart from amd64 down the line): https://conda-forge.org/docs/user/ci-skeleton.html. It only just recently became possible to do this with Github Actions, so I just had to play around with it and see what it could do for GNU Radio.

The way this works is as follows. There is a configuration file `conda-forge.yml` that must reside in the repository root. There is a conda recipe for building GNU Radio that can live anywhere but currently has gone in `.packaging/conda_recipe`. With those, you run `conda-smithy rerender` and it creates the specificed CI infrastructure including a bunch of files in various hidden directories. Whenever there is an update to conda-smithy to change/upgrade the CI scripts, or there is a change in the configuration file or conda recipe, you run `conda-smithy rerender` again and commit the changes to the CI files. As long as you're happy with what conda-smithy and the conda-forge maintainers do with the CI, it works well and is relatively painless to maintain. Normally this infrastructure is used in "feedstock" repositories that only contain a conda recipe to build packages to upload to Anaconda.org, but at least some effort has been made to accommodate this CI use case.

I see two benefits:

1. GNU Radio gets some macOS and Windows CI with relatively little effort.
2. It is possible to build and distribute conda packages that are *not* tied to a release. This could be a good way for people to test new things on master without having to build GNU Radio themselves.

As far as package creation is concerned, the CI scripts are designed to upload packages to [Anaconda.org](https://anaconda.org/) when run from a commit on a specified branch. Right now that branch is master. You can also configure the channel and label that it uploads to for Anaconda.org, currently "gnuradio" and "main". This would allow someone to install these packages using `conda install -c gnuradio gnuradio-dev`. As of now these uploads will not actually happen, because the "gnuradio" channel does not exist and there is no "BINSTAR_TOKEN" environment variable set through Github that would provide the credentials to upload to such a channel. Those are not hard to set up should that be desired. The drawback of uploading to Anaconda.org is that there is a 3 GB limit on storage, and I imagine that would fill quickly if a package was built to every commit to master. Other strategies could be explored to distribute built packages, but I'll leave that discussion for the future.

Right now I have the build passing for all three platforms, but only after excluding extra tests on macOS and Windows. I figured that was a good place to start so that it could test to see if future PRs make the build *more* broken, but there appears to be some work to do to get all tests passing. It's possible to see the tests that fail and how they fail by inspecting the build logs. I have it set up to run tests in two passes. First the non-skipped tests, where a failure results in a CI failure. Then the skipped tests, where a failure still lets the CI pass. The skipped test names are in the recipe directory in the `build.sh` and `bld.bat` scripts.

I'd love feedback and discussion on this, so please chime in!


ryanvolz:
I see that some of the jobs are currently queued, presumably because we've reached a job limit through these and other PRs/commits. It is possible to limit the conda build matrix to a single Python version to limit the number of jobs, and I can try that out if desired.

mormj:
First of all this is great, thanks for all the work you have put into this conda packaging - having Windows CI would be amazing - even though the long term desire would probably be doing some sort of direct build and having native Windows packaging (but there seem like a lot of hurdles / work to get there) - this gets us there for the foreseeable future.  And the conda installation methods have gotten a lot of traction, so having validation of it is valuable on its own.

For CI I think it would be sufficient to test a single version of python for the Windows and macOS - do you see benefit for doing the linux conda builds as part of CI compared to the native linux builds in the rest of CI?

Not a big fan of the conda-forge.yml at the top level - is there a way to bury it under `.packaging` and copy it into $GITHUB_WORKSPACE as an action?

For packaging purposes, it would be nice to be able to push latest master/maint3.9 to conda builds, and this can be done with `workflow_dispatch` actions so it doesn't happen on every push/pull request.  With that said we are striving to do more regular maint and master releases (3 mo / 6 mo respectively) - so there is less urgency to have something off of `HEAD`.

Also having this in CI in-tree and visible on every PR might lead others to dig in and help with the broken tests since they show up in the logs.

ryanvolz:
> For CI I think it would be sufficient to test a single version of python for the Windows and macOS - do you see benefit for doing the linux conda builds as part of CI compared to the native linux builds in the rest of CI?

The benefit is not as strong as with Windows and macOS, but it would help ensure that the recipe stays up to date when it comes release time and ensure that it builds successfully with the particular versions of packages available through conda-forge, which tend to be relatively recent. That's perhaps a bit of a difference with the existing CI builds.

> Not a big fan of the conda-forge.yml at the top level - is there a way to bury it under `.packaging` and copy it into $GITHUB_WORKSPACE as an action?

I'm not a huge fan of that either, and this is a good idea if I can get it to work. The problem is in customizing the CI script output from conda-smithy to do that copying. Any changes we make get overwritten by a re-render, so it would need to modify the CI script every time after being rendered. That adds an extra step, and maybe I can wrap that in a custom "re-render" script that also lives in .packaging. That would give us a wedge to make other changes if desired.

An alternative is to work with conda-smithy to get it to accept a `conda-forge.yml` that is not at the top level, which might be easier overall. Are the other hidden directories it creates at the top level also annoying? (Aside from .github workflow file, which obviously needs to be there.) I find them less so, and the fewer changes I'd have to get merged to conda-smithy the better.

Also, I think I can get rid of the .circleci directory through a .gitignore addition since it's only useful to kill builds if the repo is connected to CircleCI infrastructure in the first place.

I'll give some thought about what to do here and try something out.

> For packaging purposes, it would be nice to be able to push latest master/maint3.9 to conda builds, and this can be done with `workflow_dispatch` actions so it doesn't happen on every push/pull request. With that said we are striving to do more regular maint and master releases (3 mo / 6 mo respectively) - so there is less urgency to have something off of `HEAD`.

Building packages through `workflow_dispatch` would be another thing requiring modification of the re-rendered output, unless it's possible to trigger a workflow in that way where the workflow yaml itself does not need any modification for `workflow_dispatch`. Maybe that could be a separate yaml file that can refer to the generated one? I don't know yet what all of the possibilities here are with Github Actions. I do like the "on-demand" package building idea though, which would make it more compatible with uploads to Anaconda.org and not running out of space there.

Making built packages available through Github artifacts is another possibility I can see. I think that's something I can get added to conda-smithy (since I've already done something similar for Azure).

That brings up another point that I didn't note anywhere yet. This is not actually restricted to using Github Actions CI if it's desirable to work with other providers. Conda-smithy can render CI scripts for Azure, Appveyor, Circle, Travis, and Drone, although not all support all platforms. Azure could do the exact same thing with Linux, macOS, and Windows builds. I just note this so that people are aware of the possibilities should we want to diversify.

> Also having this in CI in-tree and visible on every PR might lead others to dig in and help with the broken tests since they show up in the logs.

I could also see something like this living out-of-tree and being updated/run semi-autonomously. The drawback there is it removes visibility in PRs and with the CI check mark. That could be something it moves to in the future if direct Windows and macOS CI are implemented.

ryanvolz:
> An alternative is to work with conda-smithy to get it to accept a `conda-forge.yml` that is not at the top level, which might be easier overall.

So far I have managed to patch conda-smithy to be able to use a specified configuration file (currently `.conda-forge.yml`, so it is now hidden). This required a fix to the build scripts which has been merged upstream, but the actual change to conda-smithy is still pending review. At least now it is possible to use my patched version to get CI scripts that work again.

---

I've also turned up a few new build bugs for Windows that have slipped in over the past few weeks since making this PR. Patches are currently in this PR, but I will split them out and submit them separately soon.

ryanvolz:
> Also, I think I can get rid of the .circleci directory through a .gitignore addition since it's only useful to kill builds if the repo is connected to CircleCI infrastructure in the first place.

It turns out, I cannot get rid of the `.circleci` directory with a `.gitignore` addition, because conda-smithy (through GitPython) ignores the ignore file when staging the changes after re-rendering.

> > For packaging purposes, it would be nice to be able to push latest master/maint3.9 to conda builds, and this can be done with `workflow_dispatch` actions so it doesn't happen on every push/pull request. With that said we are striving to do more regular maint and master releases (3 mo / 6 mo respectively) - so there is less urgency to have something off of `HEAD`.
> 
> Building packages through `workflow_dispatch` would be another thing requiring modification of the re-rendered output, unless it's possible to trigger a workflow in that way where the workflow yaml itself does not need any modification for `workflow_dispatch`. Maybe that could be a separate yaml file that can refer to the generated one? I don't know yet what all of the possibilities here are with Github Actions.

I've looked into this now, and the only way to do `workflow_dispatch` is to include it in the Github Action yaml, which means modifying the output of conda-smithy. This wouldn't necessarily be hard (I imagine a Python script that runs conda-smithy, reads the generated `conda-build.yml`, then writes it back out with some modifications), but it does put a layer between this CI and conda-smithy that could potentially be more sensitive to breakage. Combined with other potential tweaks to the output (like removing the `.circleci`), maybe this starts to make sense.

So for binary package outputs from this CI, I see 3 options:

1. Proceed with the current capabilities, including not uploading any binary packages anywhere or uploading them whenever the CI is run from a particular branch.
2. Modify the Github Action output from conda-smithy to handle binary packages in a more customized way, which could include uploading them to anaconda.org whenever the CI is triggered by a `workflow_dispatch` event.
3. Have the resulting binary packages stored as Github Action artifacts whenever the CI is run, which would require (likely accepted) changes to conda-smithy to add that capability. Installing binary packages made available in this way would be a bit cumbersome and require manual effort that could reasonably only be expected from someone with more than a passing familiarity with `conda`.

Another option that doesn't make use of this CI:

4. When pre-release conda packages are desired, create a git tag with a pre-release version identifier and then rely on the separate conda-forge infrastructure to build and upload those packages. Currently I (or a mythical other maintainer) would have to manually trigger the conda-forge build, but I'm sure it would be possible to build some tooling around that to make it easier (i.e. trigger an appropriate PR to the conda-forge gnuradio-feedstock through a Github Action here).

---

If there is no strong desire to wrap conda-smithy's re-render step and modify it in some way, then I would be inclined to recommend a combination of options (3) and (4): conda packages available for every CI run through Github Actions artifacts, and utilizing development/pre-release tags to get specific dev versions out in the regular conda-forge channel.

Since following this discussion is probably getting unwieldy, I'm going to go edit the PR description with a checklist of things that need to be settled before this is ready to go.

ryanvolz:
Noting here that `qa_add_system_time` is pretty flaky, sometimes failing on Windows or macOS.

NeuroForLunch:
@ryanvolz thanks for working on this feature. I know a lot of people use Conda but probably less than the people who don't so maybe in the future we can figure out a way to create non-conda packages as well.

ryanvolz:
> @ryanvolz thanks for working on this feature. I know a lot of people use Conda but probably less than the people who don't so maybe in the future we can figure out a way to create non-conda packages as well.

More regular testing of other (non-conda) build methods on macOS and Windows would certainly be nice, I agree! If the majority of Apple users are getting GNU Radio from MacPorts, for example, it behooves us to make sure that all of the details of that method are working properly. That said, there are plenty of common issues that any build method for a particular platform will identify, so having these conda builds going in CI will do a lot to keep other macOS and Windows build methods working properly. Hopefully the improvements this will bring in reliability will attract more developers who use macOS and Windows who will then contribute some more focused CI scripts!

As of now, conda is convenient for a few reasons:

1. All of the dependencies are provided as binary packages, and those are supported by a wider community that is completely external to GNU Radio.
2. There's a single recipe for all platforms (though with a different build script for Windows) that makes scaling to maintaining multiple platforms a little easier.
3. Aside from the recipe, the actual CI infrastructure is maintained elsewhere ([conda-smithy](https://github.com/conda-forge/conda-smithy)), which further lowers the burden for GNU Radio maintainers.
4. People will be able to download the packages that the CI built to test things in a self-contained environment that are not feasible to test in an automated manner.

ryanvolz:
Alright folks, I think this is getting close to being ready to merge. Uploading the packages as Github Artifacts is still something that you need a [PR branch of conda-smithy](https://github.com/conda-forge/conda-smithy/pull/1479) to use, but I expect that to be merged at some point. The only other high-level outstanding question has to do with whether we need a way to re-render the conda CI with conda-smithy that doesn't involve someone just doing it locally and sending a PR. I could imagine a triggered Github Action that does that, but I'm open to suggestions. That's also something that we could decide on and add later, if we just want to get this merged.

On a low level, I've rebased this to test #4724, and a new fix is needed with #4798 to get Windows building for the IIO component. If we're agreed on the high level items, I'll need to rebase again and get the build fix merged in order to demonstrate green check marks on all platforms, and then this will actually be ready to merge. So please, comment away!

ryanvolz:
Ok folks, support for saving build artifacts on GitHub Actions has finally landed in `conda-smithy`, so that last domino has fallen. Given the lack of complaints, I think this is ready to merge.

Now I'll be able to stop rebasing to catch Windows build errors, and make the PR authors fix them!

mormj:
Most of the conda piece of this looks like magic to me, but it seems to fit nicely into the hidden folders and such and produce checks and artifacts for the windows/mac which is huge!  Thank you for taking this on.  One question - what is the automerge.yml bit - does that have ramifications for merging things into *this* repo?

ryanvolz:
> Most of the conda piece of this looks like magic to me, but it seems to fit nicely into the hidden folders and such and produce checks and artifacts for the windows/mac which is huge! Thank you for taking this on. One question - what is the automerge.yml bit - does that have ramifications for merging things into _this_ repo?

Ah, good question that I needed to do some research on. I'll have to see if there's a way to not generate that, since we don't want it here. That said, it only will do anything if either of the following is true: the PR has an "automerge" tag, or the automatic merge is enabled through `.conda-forge.yml` (it is not) and "[bot-automerge]" is in the PR title and the creator of the PR is "regro-cf-autotick-bot". None of the latter conditions would ever be true here, so the only thing to consider is the "automerge" tag. Don't allow an "automerge" tag and then nothing could happen, and otherwise, what happens probably depends on the permissions of the `GITHUB_TOKEN` secret which I'm not entirely sure about.

So, I'll see if I can remove the `automerge.yml` action entirely. If that's not possible, then it should have no effect as long as no "automerge" tag is ever applied.

ryanvolz:
> > Most of the conda piece of this looks like magic to me, but it seems to fit nicely into the hidden folders and such and produce checks and artifacts for the windows/mac which is huge! Thank you for taking this on. One question - what is the automerge.yml bit - does that have ramifications for merging things into _this_ repo?
> 
> Ah, good question that I needed to do some research on. I'll have to see if there's a way to not generate that, since we don't want it here. That said, it only will do anything if either of the following is true: the PR has an "automerge" tag, or the automatic merge is enabled through `.conda-forge.yml` (it is not) and "[bot-automerge]" is in the PR title and the creator of the PR is "regro-cf-autotick-bot". None of the latter conditions would ever be true here, so the only thing to consider is the "automerge" tag. Don't allow an "automerge" tag and then nothing could happen, and otherwise, what happens probably depends on the permissions of the `GITHUB_TOKEN` secret which I'm not entirely sure about.
> 
> So, I'll see if I can remove the `automerge.yml` action entirely. If that's not possible, then it should have no effect as long as no "automerge" tag is ever applied.

There's no way (right now) to disable including this action through conda-smithy. All told, that leaves `.github/workflows/automerge.yml`, `.github/workflows/webservices.yml`, and `.circleci/config.yml` as files created by conda-smithy that do nothing and would be better if not included. So for now, I've settled on deleting them manually after re-rendering the CI, and adding those instructions in the README in the recipe directory and in the `.conda-forge.yml` file. Since I'm likely to be the only person in the near future who will do this rendering step, it's good enough for now. If I someday implement a triggered Github action that can do the rendering on demand, we can easily include those cleanups there as well.

PR #4413, opened 2021-03-19
build: remove clang warnings

drmpeg:


drmpeg:
@ThomasHabets The warning(s) for polyphase_filterbank.h was:
```
In file included from /home/ubuntu/xfer/gnuradio/gr-filter/lib/pfb_decimator_ccf_impl.cc:15:
In file included from /home/ubuntu/xfer/gnuradio/gr-filter/lib/pfb_decimator_ccf_impl.h:16:
/home/ubuntu/xfer/gnuradio/gr-filter/lib/../include/gnuradio/filter/polyphase_filterbank.h:111:5: warning: explicitly defaulted move constructor is implicitly deleted [-Wdefaulted-function-deleted]
    polyphase_filterbank(polyphase_filterbank&&) = default;
    ^
/home/ubuntu/xfer/gnuradio/gr-filter/lib/../include/gnuradio/filter/polyphase_filterbank.h:99:26: note: move constructor of 'polyphase_filterbank' is implicitly deleted because field 'd_fft' has a deleted move constructor
    fft::fft_complex_rev d_fft;
                         ^
/home/ubuntu/xfer/gnuradio/gr-fft/lib/../include/gnuradio/fft/fft.h:83:5: note: 'fft' has been explicitly marked deleted here
    fft(const fft&) = delete;
    ^
1 warning generated.
```

ThomasHabets:
Declaring a default move constructor implies that there's a definition, and could risk a link error for users of this public header, right?

The current state is similarly deceptive to the code reader though, but is at least a compile time error. Maybe just delete the line?

mbr0wn:
> Declaring a default move constructor implies that there's a definition, and could risk a link error for users of this public header, right?
> 
> The current state is similarly deceptive to the code reader though, but is at least a compile time error. Maybe just delete the line?

Agree with deleting.

ThomasHabets:
Oh, while I don't disagree, I forgot to say that if `d_fft` is ever changed to be move constructible, then if we fix this warning we *won't* automatically make the polyphase filterbank movable. As-is it would be. So that's the only downside.

ThomasHabets (approved):

marcusmueller (approved):

PR #4414, opened 2021-03-19
CMake: Look for MPIR only when no GMP is found, be verbose

marcusmueller:
This seems to be a common confusion, so let's alleviate the confusion by
letting users know what CMake does and why failing to find GMP is not a
problem when MPIR is present.

Since we preferred GMP over MPIR anyways, only check for MPIR if GMP
detection fails.


marcusmueller:
Proposing backport to both 3.8 and 3.9: the affected CMake logic across these is identical, and the compile-time confusion is the same for all branches.

marcusmueller:
Let's convert this to QUIET `find_package` and only be verbose about what actually matters to the user.

marcusmueller:
> Let's convert this to QUIET `find_package` and only be verbose about what actually matters to the user.

Doesn't work, can't silence the failing detection (greeeeeat), so the amount of verbosity necessary to explain the situation stays the same.

mormj (approved):
LGTM

cmake/Modules/FindMPLIB.cmake
```diff
@@ -1,7 +1,10 @@
+message(STATUS "Trying to find GMP or MPIR.")
+message(STATUS "Failure to find either is not a problem. Only one of them is required.")
```
jacobagilbert, line 2: What % of users are going to be using GMP vs MPIR? Probably a significant portion; to manage verbosity this could be moved into the else(GMP_FOUND) conditional. The "other two new messages above/below are possibly not necessary.

marcusmueller: Done that.

jacobagilbert (approved):
Looks good!

PR #4415, opened 2021-03-19
uhd: Add option to sync timestamp to PC Clock or GPS time on next 1PPS

azl5361:
Currently, the only options for setting the timestamp in GRC are `unknown_pps` and `pc_clock`:

- `pc_clock` sets the timestamp to the system time, but does not align these timestamps to the 1PPS input (it instead uses `set_time_now()`). 

- `unknown_pps` aligns the timestamps to the 1PPS input, but sets the timestamp to 0, rather than some form of UTC epoch time. 

This pull request seeks to add options in GRC that will set the timestamp to a value that reflects UTC epoch time, and is also aligned to the 1PPS. Two new options are considered-  `gps_time` and `pc_clock_next_pps`:

- `gps_time` obtains the integer portion of GPS time, adds one second, and sets this timestamp on the next 1PPS (similar to lines 227 - 232 of [query_gpsdo_sensors.cpp](https://github.com/EttusResearch/uhd/blob/master/host/utils/query_gpsdo_sensors.cpp))

- `pc_clock_next_pps` obtains the integer portion of the system time, adds one second, and sets this timestamp on the next 1PPS (useful in scenarios where GPS is not available)

### DIFF
--- a/gr-uhd/grc/gen_uhd_usrp_blocks.py
+++ b/gr-uhd/grc/gen_uhd_usrp_blocks.py
@@ -52,8 +52,8 @@ parameters:
 -   id: sync
     label: Sync
     dtype: enum
-    options: [sync, pc_clock, none]
-    option_labels: [Unknown PPS, PC Clock, No Sync]
+    options: [sync, pc_clock, pc_clock_next_pps, gps_time, none]
+    option_labels: [Unknown PPS, PC Clock, PC Clock on Next PPS, GPS Time on Next PPS, No Sync]
     hide: ${'$'}{ 'none' if sync else 'part'}
 -   id: start_time
     label: Start Time (seconds)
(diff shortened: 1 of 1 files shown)


gr-uhd/grc/gen_uhd_usrp_blocks.py
```diff
@@ -177,6 +177,12 @@
         self.${'$'}{id}.set_time_unknown_pps(uhd.time_spec(0))
         ${'%'} elif sync == 'pc_clock':
         self.${'$'}{id}.set_time_now(uhd.time_spec(time.time()), uhd.ALL_MBOARDS)
+        ${'%'} elif sync == 'pc_clock_next_pps':
+        self.${'$'}{id}.set_time_next_pps(uhd.time_spec(int(time.time()) + 1.0))
+        time.sleep(1)
+        ${'%'} elif sync == 'gps_time':
+        self.${'$'}{id}.set_time_next_pps(uhd.time_spec(self.${'$'}{id}.get_mboard_sensor("gps_time").to_int() + 1.0))
```
mbr0wn, line 184: I'm not entirely sure that this will actually operate directly after a PPS edge, but you're right it's what we do in query_gpsdo_sensors. Please add a comment here explaining why this is valid.

gr-uhd/grc/gen_uhd_usrp_blocks.py
```diff
@@ -177,6 +177,12 @@
         self.${'$'}{id}.set_time_unknown_pps(uhd.time_spec(0))
         ${'%'} elif sync == 'pc_clock':
         self.${'$'}{id}.set_time_now(uhd.time_spec(time.time()), uhd.ALL_MBOARDS)
+        ${'%'} elif sync == 'pc_clock_next_pps':
+        self.${'$'}{id}.set_time_next_pps(uhd.time_spec(int(time.time()) + 1.0))
```
mbr0wn, line 181: This does not sync to a PPS edge, so I would suggest leaving this mode out. If the next PPS comes immediately after this arrives at the device, the sync will be almost 1s off. In other words, this does the same as `pc_clock` except much worse.

mbr0wn (changes requested):
- A change like this must go onto master first. So please rebase. A single commit will suffice, so please squash also.
- I would like to discuss the utility of the pc-clock-to-next PPS thing. I don't think it's useful.

gr-uhd/grc/gen_uhd_usrp_blocks.py
azl5361: I had envisioned this mode being useful when the user wants the timestamps' integer values to be synchronized to the 1PPS input (i.e. 1PPS input triggers on timestamps 1616430761.0 rather than 1616430761.29557...). This might also be useful when setting/synchronizing the timestamps across different USRP devices that are connected to the same 1PPS input. While this can be accomplished using the `unknown_pps` mode, this mode also initializes the timestamps to 0. The proposed `pc_clock_next_pps` mode initializes the timestamps using the system's UTC epoch time, which gives the user a better sense of absolute time of each timestamp.

To address your point of sync being nearly 1s off (depending on when the timestamp is set), I have changed the code so that the time is set immediately after the 1PPS input is received. This is done using the process given in the [Device Synchronization](https://files.ettus.com/manual/page_sync.html) section of the USRP manual. Here, `get_time_last_pps()` is polled every 50 ms until a change is detected, then `set_time_next_pps()` is called.

gr-uhd/grc/gen_uhd_usrp_blocks.py
azl5361: Comments added in requested section of code

mormj:
@mbr0wn - what is your recommendation with these changes at this point?

mbr0wn (approved):

gr-uhd/grc/gen_uhd_usrp_blocks.py
mbr0wn: OK let's go with this.

PR #4417, opened 2021-03-19
Remove boost::bind where trivial

marcusmueller:
That being everywhere where boost::asio isn't involved.

The boost::asio error placeholders for some reason don't play nicely with lambdas, I can't really figure out what's going wrong there, so let's leave it at this for now. This leaves boost::bind in place where boost::asio is used.

### DIFF
--- a/gnuradio-runtime/include/gnuradio/thrift_application_base.h
+++ b/gnuradio-runtime/include/gnuradio/thrift_application_base.h
@@ -202,8 +202,8 @@ void thrift_application_base<TserverBase, TserverClass>::start_application()
             "thrift", "init_attempts", d_default_max_init_attempts));
 
     if (!p_impl->d_application_initialized) {
-        p_impl->d_start_thrift_thread.reset(
-            (new gr::thread::thread([app = d_application] { app->start_thrift(); })));
+        p_impl->d_start_thrift_thread = std::make_shared<gr::thread::thread>(
+            [app = d_application] { app->start_thrift(); });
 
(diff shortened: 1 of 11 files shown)


gnuradio-runtime/include/gnuradio/thrift_application_base.h
```diff
@@ -202,8 +202,8 @@ void thrift_application_base<TserverBase, TserverClass>::start_application()
             "thrift", "init_attempts", d_default_max_init_attempts));
 
     if (!p_impl->d_application_initialized) {
-        p_impl->d_start_thrift_thread.reset((new gr::thread::thread(
-            boost::bind(&thrift_application_base::start_thrift, d_application))));
+        p_impl->d_start_thrift_thread.reset(
```
ThomasHabets, line 205: Slightly neater to change this to use `make_unique` now that we have C++14, to make it clearer from grep and reading that this is not manual memory management.

ThomasHabets (approved):

gnuradio-runtime/include/gnuradio/thrift_application_base.h
marcusmueller: oh, nice!  will do.

marcusmueller: it's a shared pointer, but `make_shared` is still a thing, so.

ThomasHabets (approved):

Issue #4348, opened 2021-03-11
qa_packet_headerparser_b: Flaky


mbr0wn:
I thought this would have been fixed by 02b909bbcfc27ffea582b06dd2f95e478e0cdee2, but that just changed the error:

```
214/244 Test #214: qa_packet_headerparser_b .....................***Failed    5.87 sec
..F.
======================================================================
FAIL: test_003_ofdm (__main__.qa_packet_headerparser_b)
Header 1: 193 bytes
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/__w/gnuradio/gnuradio/gr-digital/python/digital/qa_packet_headerparser_b.py", line 121, in test_003_ofdm
    self.waitFor(lambda: sink.num_messages() == 2)
  File "/__w/gnuradio/gnuradio/gnuradio-runtime/python/gnuradio/gr_unittest.py", line 147, in waitFor
    self.fail(fail_msg)
AssertionError: Timeout exceeded during call to waitFor()!

----------------------------------------------------------------------
Ran 4 tests in 5.621s
```

Latest failure: https://github.com/gnuradio/gnuradio/pull/4340/checks?check_run_id=2081291093

willcode:
In test 001, are there 3 headers in the data and one random tag? Then the test for num_messages might be 3 or you might not test it until it's 4. Then it would time out.

willcode:
Verified that it is too few messages coming through. In the cases seen, it was the last message that was dropped, not one in the middle. The happened randomly in different test cases. Additional logging in waitFor() shows that it is working properly.

willcode:
With additional input data, and the qa adjusted to ignore extra messages, the error does not happen. I do not think this is a problem with either the qa or the block. Best _guess_ is that this is a race between the DONE and the packet messages. If the DONE gets processed first, the block shuts down and further messages are ignored.

willcode:
Just got a qa_pdu failure on maint-3.9, one CI box, can't replicate. Same effect, I think.

willcode:
Possible that this happens more often with fewer cpus. CI uses 2?

PR #4418, opened 2021-03-19
Make message_debug::store/num_messages thread-safe

marcusmueller:
Only needs a mutex. This was found while trying to figure out the headerparser mystery #4348.

On the way, use the native `size_t` for messgae index handling.

Backport: 3.7-3.9, only the mutexing in 4464766fa .

### DIFF
--- a/gr-blocks/include/gnuradio/blocks/message_debug.h
+++ b/gr-blocks/include/gnuradio/blocks/message_debug.h
@@ -52,7 +52,7 @@ public:
     /*!
      * \brief Reports the number of messages received by this block.
      */
-    virtual int num_messages() = 0;
+    virtual size_t num_messages() = 0;
 
     /*!
      * \brief Get a message (as a PMT) from the message vector at index \p i.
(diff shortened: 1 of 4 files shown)


willcode:
The num_messages read should be atomic already.

marcusmueller:
I'll say "it's extremely unlikely that a 64 bit read is not atomic", but atomic is not the same as should be atomic!

willcode:
There is an actual race with `size()` and `resize()` in a vector if the capacity is exceeded and the vector is reallocated. So, good idea to lock anyway and it's cheap.

rear1019:
> The num_messages read should be atomic already.

It’s not. At least the `libstdc++` implementation of `std::vector` doesn’t store the size directly; it stores `ptr` and `ptr + size` instead (i.e. the `begin()/end()` iterators). Even if the size was stored directly: beware of races during a reallocation.

mbr0wn (approved):
zomfg I so hope this kills that QA failure.

willcode:
That race is really low probability. We can hope.

PR #4422, opened 2021-03-20
CMake: Look for MPIR only when no GMP is found, be verbose (backport to maint-3.8)

willcode:
This seems to be a common confusion, so let's alleviate the confusion by
letting users know what CMake does and why failing to find GMP is not a
problem when MPIR is present.

Since we preferred GMP over MPIR anyways, only check for MPIR if GMP
detection fails.

(cherry picked from commit 43bfd85e10d8c83e2518ca2dc424d21c4b319f36)

Backport https://github.com/gnuradio/gnuradio/pull/4414


jacobagilbert (approved):
Changes look good to me.

PR #4428, opened 2021-03-21
Make xoroshiro128p actually conformant C

marcusmueller:
Used to include cstdint, which of course is C++


drmpeg:
This commit is also in https://github.com/gnuradio/gnuradio/pull/4430

mbr0wn (approved):

marcusmueller:
@drmpeg that's right! That #4430 depends on this, but will be rebased soon as this is merged.

PR #4430, opened 2021-03-21
Make gr::random API consistently 64bit, use XOROSHIRO128+

### LINKED ISSUE #1559: gr::random(unsigned int seed) <-> fastnoise_source(long seed)
This is separate from #1542, and it's a 64bit-specific issue:

Whereas the gr::random() constructor only takes an int (which has sizeof==4 on x86_64), the fastnoise_source takes a long (which has sizeof==64 on x86_64), and directly passes that to the gr::random c'tor in its initializer list.

On 32 bit x86, sizeof int == sizeof long, so no problem there. But this makes the call inconsistent across different machines (and will make two different seeds be, unexpectedly, the same seed on x86_64).

Although I'd guess that people using fastnoise source will have little practical problem with that, it's still a bug (fixable simply by adjusting either c'tor signature, and I'm leaning towards changing both to uint64_t) that requires an API change.

marcusmueller:
fixes #1559.

Seeding being inconsistent between gr::random and things like fastnoise
source is an outstanding issue (#1559).

Fix that by fixing the API, and pivoting to our built-in XOROSHIRO128+,
which is also substantially faster than MT19937.

For that purpose, introduce, and python-bind, a wrapper class that can be
used with STL distribution shapers.

Add a unit test for the (P)RNG.


marcusmueller:
Note that arguably, changing the sequence of random numbers generated with a seed is a behavioural break and can't be done on a release, so I don't think backporting is possible.

mbr0wn:
Consistently failes `qa_random` on all Dockers:

```======================================================================
ERROR: test_005_xoroshiro128p_seed_stability (__main__.test_random)
  File "/__w/gnuradio/gnuradio/gnuradio-runtime/python/gnuradio/gr/qa_random.py", line 75, in test_005_xoroshiro128p_seed_stability
    self.assertEqual(3520422898491873512, rng())
TypeError: 'gnuradio.gr.gr_python.xoroshiro128p_prng' object is not callable
```

marcusmueller:
wut.

gr-analog/lib/fastnoise_source_impl.cc
```diff
@@ -117,25 +117,25 @@ void fastnoise_source_impl<gr_complex>::set_amplitude(float ampl)
 template <class T>
 void fastnoise_source_impl<T>::generate()
 {
-    int noutput_items = d_samples.size();
+    size_t noutput_items = d_samples.size();
```
mbr0wn, line 120:
```suggestion
    const size_t noutput_items = d_samples.size();
```

no biggie

gr-analog/lib/fastnoise_source_impl.cc
```diff
@@ -24,24 +24,24 @@ namespace analog {
...
+fastnoise_source<T>::make(noise_type_t type, float ampl, uint64_t seed, size_t samples)
 {
     return gnuradio::make_block_sptr<fastnoise_source_impl<T>>(type, ampl, seed, samples);
 }
 
 template <>
 void fastnoise_source_impl<gr_complex>::generate()
 {
-    int noutput_items = d_samples.size();
+    size_t noutput_items = d_samples.size();
```
mbr0wn, line 35:
```suggestion
    const size_t noutput_items = d_samples.size();
```

mbr0wn (approved):
Did you ever benchmark xoroshiro vs. mt19937?

marcusmueller:
> Did you ever benchmark xoroshiro vs. mt19937?

Randomness-quality-wise:

yup, did! Working [upstream](https://lists.phy.duke.edu/pipermail/dieharder-devel/2020-October/000107.html) with dieharder, after I had to justify using XOROSHIRO128+ in a GS/s Gaussian PRNG for testing ca 10⁴–10⁵ long codes.

Long story short: it's fine, it uses way less internal state, but generates randomness nearly as white as MT19937.

Throughput-wise:

Theoretically, the much smaller internal state should make a significant difference for in-loop generated PRN. In reality, the computational load is dominated by the std:: distribution shapers, which makes the expected returns on this rather small.

I do know / [show](https://github.com/gnuradio/gnuradio/pull/1562#discussion_r164283113) that an averaging based Gaussian-shaper based on taking 32 uniform XOROSHIRO128+ integers is faster but better *around the expectation* than MT19937 + `std::normal_distribution`. However, that's *not better* in general, as it definitely can't have tails; after much digital hardware dev by my student for above decoder testing hardware, I believe the only way out here is to actually know how many bits of "hi-fi" Gaussianness you actually need, and using an appropriately designed Box-Muller transform, if you actually need those tails.

For the "easier" distributions, you should see the difference. Can hand in some actual before/after benchmarks, of course! 

A periphery motivation was that there's an optimized XOROSHIRO128+ that does multiple PRNGs in parallell, really making use of the compactness of the algorithm. Armed with that, later rework could include things like not generating a single value (especially not a single bit) from a call to `xoroshiro128p_next`, and then discarding all caches, but to cache as much as was "cheap" to produce at once and then hand it out in subsequent calls; `gr::random()` does a good job of encapsulating the PRNG opaquely already.

mbr0wn:
Can you please squash those suggestions so we have DCO coverage.

mormj:
manually rebased with https://github.com/gnuradio/gnuradio/pull/4496

PR #4439, opened 2021-03-23
filter designer: restore compatibility with 3.8

willcode:
Previously, an integer value was used to represent the wintype param
when saving to a CSV file. A preceeding fix to enable CSV loading
changed this to a string value specific to the internal Python enum.
This commit changes it back to an int.

### DIFF
--- a/gr-filter/python/filter/design/filter_design.py
+++ b/gr-filter/python/filter/design/filter_design.py
@@ -1978,8 +1978,13 @@ class gr_plot_filter(QtGui.QMainWindow):
         else:
             csvhandle.writerow(["restype","fir"])
 
+        enum_params = ('wintype',)
         for k in list(self.params.keys()):
-            csvhandle.writerow([k, self.params[k]])
+            # For enum fields, save the value, not the name.
+            if k in enum_params:
+                csvhandle.writerow([k, self.params[k].value])
+            else:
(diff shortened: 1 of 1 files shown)


mormj (approved):
I had previously given incorrect feedback on @dl1ksv PR that prevented backwards compatibility.  This should solve that issue

dl1ksv:
Sorry, I'm a bit confused now. I just checked  out gr maint-3.8 . There the wintype is saved as int in the csv file and properly reread on open.

My first proposal to fix this issue was to save wintype as int not as string.
This was not accepted. Of course this breaks compatibility between 3.8 and actual master. I was not aware of that at that time.

The proposed patch now saves the wintype as integer in are more complicated way.
Why not to revoke my pr and go back to my initial proposal and change 
https://github.com/gnuradio/gnuradio/blob/3e4e7feffb29f54364b370594ee953d1c2d53327/gr-filter/python/filter/design/filter_design.py#L805 to
wintype = int(self.filterWindows[winstr])

which has the same effekt

willcode:
Simpler works for me.

mormj:
@dl1ksv - that sounds like a good plan.  Reverting #4356 and we'll go back to your original proposal.

willcode:
Closing in favor of @dl1ksv original solution.

PR #4444, opened 2021-03-23
gr_filter_design: save wintype as int instead of str

dl1ksv:
@mormj  The conversion to has to be done. Otherwise we can't load the saved file. See discussion in #4439


mormj (approved):
This does indeed fix the original issue and prevents the issues caused by my earlier recommendation.  Sorry for the churn on this one.

willcode (approved):

willcode:
Checked compatibility with 3.8 - good.

PR #4448, opened 2021-03-23
Fastnoise_source: when picking indices from random pool, avoid modulo operation when possible

marcusmueller:
During an ambulatory encounter with a simulation flowgraph, I benchmarked and found `fastnoise_source` to be surprisingly stuck (ca 15% of time of a 20s execution, which included pool generation!) in the `div` instruction, plus surroundings that also contributed significantly to runtime.

When the pool is power-of-2-sized, index generation can be done using a
simple bitmask. So I did that.

On my private machine, the stats are:

Prior this patch: Standardnormal random variables

* `fastnoise_source`, pool size 2²⁰: 93.5 MS/s
* `noise_source`: 9.7 MS/s (just for comparison)

After the patch:

* `fastnoise_source`, pool size 2²⁰: 251 MS/s
* `fastnoise_source`, pool size 2²⁰-1: 91 MS/s

Since there's usually no reason to use anything but power-of-two sizes for the pool of random numbers that follows the selected distribution, the <2% performance regression is IMHO tolerable.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,12 @@ Older Logs can be found in `docs/RELEASE-NOTES-*`.
   - requires MSVC 1914 (Microsoft VS 2017 15.7)
 - Windows build: removed unnecessary MSVC-specific system include overrides
 
+#### gr-analog
+
+- `fastnoise_source`: Use `uint64_t` seed API, use `size_t` for vector length/indices
+- `fastnoise_source`: Use a simple bitmask if the random pool length is a power
+  of 2 to determine indices, instead of `%`, which consumed considerable CPU
+
 ### Added
 
 - New in-tree module gr-pdu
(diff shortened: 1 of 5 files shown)


marcusmueller:
note that the unsigned / fixed length commits are in there out of necessity and can be rebased away after #4430

mormj (approved):

drmpeg (approved):
Looks good. I really like the removal of ```long``` (good for 32-bit small systems). I'm not a fan of the extra verbosity ```gr::log :INFO: fastnoise_source0 - Generating 65536 complex values. This might take a while.``` though.

Perhaps this message could be limited to really large pools?

marcusmueller:
Limiting this to pools > 2¹⁸ might be a good idea, indeed.

drmpeg:
> Limiting this to pools > 2¹⁸ might be a good idea, indeed.

I would go higher. A 2^22 pool only takes 1 second on my armv7 system.

marcusmueller:
Went higher. 2²³ seemed to be a sensible lower bound for spamming the log. We might want to squash all this before merging.

PR #4449, opened 2021-03-24
blocks: tcp_server_sink QA: sequence compare failed because list!=tuple

marcusmueller:
Although the contained values are the same, this unit test failed on me.

Using Python's unittest assertSequenceEqual solves that, it seems.

### DIFF
--- a/gr-blocks/python/blocks/qa_tcp_server_sink.py
+++ b/gr-blocks/python/blocks/qa_tcp_server_sink.py
@@ -15,7 +15,7 @@ import socket
 from time import sleep
 
 from threading import Timer
-from multiprocessing import Process
+from multiprocessing import Process, Queue
 
 
 class test_tcp_sink(gr_unittest.TestCase):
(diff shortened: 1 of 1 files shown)


mormj:
Where was this one failing previously?

marcusmueller:
Interestingly, the test doesn't fail, although the assertion fails, so you only see it when running with `-VV`

``` ctest -VV -R tcp                                                                                                                                                                                                          
UpdateCTestConfiguration  from :/home/marcus/src/gnuradio/build/DartConfiguration.tcl
UpdateCTestConfiguration  from :/home/marcus/src/gnuradio/build/DartConfiguration.tcl
Test project /home/marcus/src/gnuradio/build
Constructing a list of tests
Done constructing a list of tests
Updating test list for fixtures
Added 0 tests to meet fixture requirements
Checking test dependency graph...
Checking test dependency graph end
test 94
    Start 94: qa_tcp_server_sink

94: Test command: /usr/bin/sh "/home/marcus/src/gnuradio/build/gr-blocks/python/blocks/qa_tcp_server_sink_test.sh"
94: Test timeout computed to be: 10000000
94: Process Process-1:
94: Traceback (most recent call last):
94:   File "/usr/lib64/python3.9/multiprocessing/process.py", line 315, in _bootstrap
94:     self.run()
94:   File "/usr/lib64/python3.9/multiprocessing/process.py", line 108, in run
94:     self._target(*self._args, **self._kwargs)
94:   File "/home/marcus/src/gnuradio/gr-blocks/python/blocks/qa_tcp_server_sink.py", line 50, in _tcp_client
94:     self.assertEqual(self.data, dst.data())
94:   File "/usr/lib64/python3.9/unittest/case.py", line 831, in assertEqual
94:     assertion_func(first, second, msg=msg)
94:   File "/usr/lib64/python3.9/unittest/case.py", line 824, in _baseAssertEqual
94:     raise self.failureException(msg)
94: AssertionError: (0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15) != [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15]
94: /usr/lib64/python3.9/multiprocessing/process.py:331: ResourceWarning: unclosed <socket.socket fd=7, family=AddressFamily.AF_INET, type=SocketKind.SOCK_STREAM, proto=0, laddr=('127.0.0.1', 55492), raddr=('127.0.0.1', 65510)>
94:   traceback.print_exc()
94: ResourceWarning: Enable tracemalloc to get the object allocation traceback
94: .
94: ----------------------------------------------------------------------
94: Ran 1 test in 0.011s
94:
94: OK
1/1 Test #94: qa_tcp_server_sink ...............   Passed    0.33 sec

The following tests passed:
	qa_tcp_server_sink

100% tests passed, 0 tests failed out of 1

Total Test time (real) =   0.38 sec
```

mormj (approved):

mormj:
Interesting.  I'm supposing this could also be fixed with removing the `tuple()` from:
```
        self.data = tuple([x for x in range(n_data)])
```
but your change is probably more robust.

marcusmueller:
let's actually do both (making a tuple out of a list out of a generator expression is inelegant, anyway.)

marcusmueller:
Oh I just realized why the unit test wasn't failing: that exception-raising function is called in a subprocess 🤨

marcusmueller:
That shifts the focus; need to make that unit test actually fail when it fails.

mormj:
Since the tcp_server_sink block is deprecated, I figure we can port this over to gr-network eventually because that module is very lightly qa'd and these test should apply there

marcusmueller:
so, just closing this would be fine?

mormj:
Makes sense for this one - but gr-network needs the same mechanism.  Adding this closed PR to the project list so we don't lose track of it.

PR #4450, opened 2021-03-24
filter: fix out of bound access in qa_fir_filter_with_buffer, add length check

marcusmueller:
Did a build with `export CXXFLAGS=-D_GLIBCXX_ASSERTIONS`, was screamed at by libcxx. Fun times!

### DIFF
--- a/gr-filter/lib/fir_filter_with_buffer.cc
+++ b/gr-filter/lib/fir_filter_with_buffer.cc
@@ -18,6 +18,7 @@
 #include <algorithm>
 #include <cstdio>
 #include <cstring>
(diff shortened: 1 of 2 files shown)


marcusmueller:
Argh, fir_filter_with_buffers needs to be empty-constructable for pfb_synth, but then does a bunch of uninitialized memory access. Will need to check whether, instead of `throw std::runtime_error("need at least 1 tap")` a simple NOP (maybe a INFO log) would do.

mormj:
Closing this draft for now, obviously re-open it if you want, just trying to prune the PR queue a bit

PR #4457, opened 2021-03-24
gr-audio: Check CreateEvent() for failure

ZackerySpytz:
CreateEvent() should be checked for failure.  An access violation
could occur if it is not.


mormj (approved):
Looks reasonable to me

marcusmueller (approved):

PR #4459, opened 2021-03-24
pmt: tag_to_pmt use from_python

### LINKED ISSUE #1072: tag_to_pmt newtag.key should be a from_python call
See [tag_utils.py:26](https://github.com/gnuradio/gnuradio/blob/master/gnuradio-runtime/python/gnuradio/gr/tag_utils.py#L26) which I think should read 

`newtag.key = pmt.from_python(tag.key)`

otherwise newtag.key will not be a pmt.  Probably just a cut and paste typo.

mormj:
Fixes #1072

### DIFF
--- a/gnuradio-runtime/python/gnuradio/gr/tag_utils.py
+++ b/gnuradio-runtime/python/gnuradio/gr/tag_utils.py
@@ -19,15 +19,6 @@ def tag_to_python(tag):
     newtag.srcid = pmt.to_python(tag.srcid)
     return newtag
 
-def tag_to_pmt(tag):
-    """ Convert a Python-readable object to a stream tag """
-    newtag = gr.tag_t()
-    newtag.offset = tag.offset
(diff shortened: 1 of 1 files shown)


marcusmueller (approved):

willcode:
Shoud this be `pmt.to_pmt()` instead of `pmt.from_python()`? Can't find the latter anywhere, unless it just got added.

mormj:
It appears you are right - from_python is not a thing.  Was taking the seemingly obvious fix to try and kill an issue.  I'll change it in a bit (my workspace is currently occupied)

willcode:
Apparently `tag_utils.tag_to_pmt()` is not used anywhere. Now that I look at it, it doesn't even do what it says it does. Consider removing the function?

mormj:
That is fair.  Also guessing no one is using it if it blatantly doesn't work

willcode (approved):

Issue #1541, opened 2017-12-19
Memory Allocation Bug


silidan:
gnuradio fails to allocate memory of size 1048576 KB or larger.

gr::vmcircbuf_sysv_shm: shmget (1): Invalid argument
gr::vmcircbuf_sysv_shm: shmget (1): Invalid argument
gr::vmcircbuf_sysv_shm: shmget (1): Invalid argument
gr::buffer::allocate_buffer: failed to allocate buffer of size 1048576 KB
gr::vmcircbuf_sysv_shm: shmget (1): Invalid argument
gr::vmcircbuf_sysv_shm: shmget (1): Invalid argument
gr::vmcircbuf_sysv_shm: shmget (1): Invalid argument
gr::buffer::allocate_buffer: failed to allocate buffer of size 1048576 KB
terminate called after throwing an instance of 'std::bad_alloc'
  what():  std::bad_alloc

>>> Done

Ubuntu 16.06 64bit, 4.4.0-103-lowlatency,
kernel.shmall = 18446744073692774399
kernel.shmmax = 18446744073692774399
kernel.shmmni = 4096

[memory_allocation_bug.grc.log](https://github.com/gnuradio/gnuradio/files/1572795/memory_allocation_bug.grc.log)

![memory_allocation_bug_flowgraph](https://user-images.githubusercontent.com/4339095/34172407-3fd451de-e4f2-11e7-94fc-a1490267f57f.png)
[memory_allocation_bug.log](https://github.com/gnuradio/gnuradio/files/1572783/memory_allocation_bug.log)

thasti:
To add some information to this:

- The flowgraph tries to create a vector with 1024*1024*64 complex values, which will cause a gr::buffer allocation for 2^30 elements. 
- gr::buffer implements this as a gr::vmcircbuf, with the size argument passed as an int (which is still okay so far). 
- gr::vmcircbuf_sysv_shm will try to allocate a shared memory buffer (shmget) with int(2 * size + 2 * pagesize) as an argument. This becomes greater than 2^31, overflows to negative number, which is in turn converted to size_t by being passed as an argument to shmget
- the cast to size_t causes the argument to become a huge positve number (close to 2^64 on 64 bit systems) and therefore being larger than either SHMMAX or the available system memory.

One can see multiple problems:
- gr::buffer should restrict the size of allocated buffers (maybe)
[...]

PR #4460, opened 2021-03-25
runtime: buffer factories use size_t

mormj:
Addresses the issue where large buffer size requests get converted back
and forth to signed/unsigned and can cause an overflow and the buffer
allocation to fail

Addresses #1541


willcode:
vmcircbuf d_size field and constructor still have int.

marcusmueller (approved):
Sensible change!

willcode:
API change, but backward compatible. We can backport to 3.9 ... is anyone using GB circbufs?

PR #4461, opened 2021-03-25
runtime: remove pybind for internal things

mormj:
Because much of the pybind code was generated automatically, things that shouldn't be and make no sense to be exposed in the python API are.  This removes many of the things that were added between 3.8 and 3.9 from the runtime API


mormj:
Looks like QA is timing out - might have gotten a bit too aggressive - will have to investigate

mormj:
Rolled back the changes to block_python.cc as these are needed by the gateway blocks apparently.

willcode (approved):

PR #4462, opened 2021-03-25
runtime: size_t for vmcircbuf constructor

mormj:


mbr0wn (approved):

willcode:
Even though this is an API change, it should be backward compatible. So we can backport it to 3.9 if needed. Are there many people allocated GB circbufs?

PR #4465, opened 2021-03-27
gr-iio: Handle some build cases

dl1ksv:
Building gr-iio fails, if libiio is not located in /usr/local.
You can come across this issue, if you point cmake on the commandline to
LIBIIO_INCLUDE_DIR and LIBIIO_LIBRARY.
Then the makefiles will be generated, but compiling fails as iio.h is not found, as the corresponding  include directory is not added.

If the ad9361 component is missing gr-iio builds, but the test fails with:

```
  File "/home/schroer/gnuradiocomponents/gnuradio-volker/gr-iio/python/iio/qa_iio.py", line 12, in <module>
    from gnuradio import gr, gr_unittest, iio, blocks
  File "/home/schroer/gnuradiocomponents/gnuradio-volker/gr-iio/python/iio/__init__.py", line 35, in <module>
    from .iio_python import *
ImportError: /home/schroer/gnuradiocomponents/gnuradio-volker/buildgriio/gr-iio/python/iio/bindings/iio_python.cpython-39-x86_64-linux-gnu.so: undefined symbol: _ZN2gr3iio10pluto_sink4makeERKNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEEymmmbdPKcSB_ff
```

This pr adds the missing include directory and excludes the components from building that depend on the ad9361 lib.


gr-iio/CMakeLists.txt
```diff
@@ -10,7 +10,7 @@
 ########################################################################
 include(GrBoost)
 
-find_package(AD9361 QUIET)
+find_package(AD9361)
```
mbr0wn, line 13: Do we need both this change and the one below?

dl1ksv: Well, not necessarily. But nowhere else in the CMakeFiles find_package is used with the quiet option.
I personally prefer more verbosity when developing.

As gr-iio will be build independently whether AD9361 is found or not, the user should get a hint about the consequences.

dl1ksv: Just detected

find_package(Qt5Widgets QUIET)

dl1ksv:
This pr is included in #4471

PR #4467, opened 2021-03-27
gr_filter_design: conditionally enable save

dl1ksv:
File->save fails with key error, if no filter was calculated.
So disable 'save' up to a filter was calculated.

### DIFF
--- a/gr-filter/python/filter/design/filter_design.py
+++ b/gr-filter/python/filter/design/filter_design.py
@@ -1988,6 +1988,16 @@ class gr_plot_filter(QtGui.QMainWindow):
         else:
             csvhandle.writerow(["taps",] + list(self.taps))
         handle.close()
+        self.gui.action_save.setEnabled(False)
+        # Iterate through all plots and delete the curves
+        for window in self.plots.values():
+            window.drop_plotdata()
+        # Clear filter coeffs
+        self.gui.filterCoeff.setText("")
+        self.gui.mfilterCoeff.setText("")
+        # Clear poles and zeros plot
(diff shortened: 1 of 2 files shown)


willcode:
It should really also be disabled any time you change anything, and before "design". Otherwise you get old data.

mormj:
> It should really also be disabled any time you change anything, and before "design". Otherwise you get old data.

If this is relatively simple to fix, that would be another good check to add here.  Otherwise this change alone makes some improvement.

mormj (approved):

dl1ksv:
Of course, this is doable, but not quite easy. And it needs a deeper discussion. What does it mean 'outdated'.
Example:

You start gr_filter_design
switch to IIR
design           now you have a valid calculated filter

go back to FIR. Is the IIR filter now outdated ?
go back to IIR  . Is the filter now again valid ?

To be consequent, you had to clean up each 'result window' like Magnitude response, Pole plot etc.  at once, if you just change one parameter entry. Is this intended ?

willcode:
Would it work to make Save a one-shot, e.g., enable Save after Design, then disable it after Save.

dl1ksv:
Disableing 'Save' after saving is no big deal. I'll look if I can clear the results of 'Design' after saving in addition. I think, this would be perfect.

dl1ksv:
Now 'Save' will be disabled after saving of a designed filter has been applied and the plotted are deleted, to make clear that you have to design a new filter before saving again.

willcode (approved):
Tried it out, looks good.

PR #4468, opened 2021-03-28
Bessel lp/hp filter have only one critical parameter

dl1ksv:
Bessel Lp / Hp filters have only one critcal parameter. If you modify the critical point 2 in a Bessel Lp filter,you get the same filter coefficients.

So remove the critical point 2 entry in the gui in case of calclating a bessel Lp / Hp filter 


mormj:
I know it's unrelated to this, but I haven't seen that qa_filterbank failure.  
change here looks good!

mormj (approved):

PR #4470, opened 2021-03-29
Fix python errors reported by lgtm.com

mormj:


gr-utils/blocktool/core/parseheader_generic.py
```diff
@@ -75,10 +75,9 @@ def initialize(self):
 
         if type(self.target_file) == list:
             self.module = self.target_file
-            for dirs in self.target_file:
-                if not os.path.basename(self.module).startswith(Constants.GR):
```
swap-nil7, line 79: Shouldn't `self.module` be changed to `dirs` instead? I am not aware about basename for a list.

swap-nil7 (changes requested):

gr-utils/blocktool/core/parseheader_generic.py
swap-nil7: And if that is the case, please update `dirs` to `dir` instead.

mormj: I would need to step through to remember what was happening here.  You may be correct - this is just replicating the existing behavior

PR #4471, opened 2021-03-29
gr-iio: Modernize CMake

ryanvolz:
I believe this should fix the complaints addressed by #4465 (iio headers installed to a local prefix not found) while also modernizing the CMake files to make them easier to maintain going forward.

I'm flexible on renaming the `Find*.cmake` modules, but I figured that these names fit better in the more modern style and they are what the upstream projects use to call themselves. If they ever decide to start shipping their own CMake Config files, these names should allow for better interoperability.


adamhorden:
"I'm flexible on renaming the `Find*.cmake` modules, but I figured that these names fit better in the more modern style and they are what the upstream projects use to call themselves."

When I wrote these I was not sure what naming convention to use. I prefer the naming convention in this pull request 🤓 .

**Adam Horden**

PR #4472, opened 2021-03-29
gr-uhd: fix command tuning

jacobagilbert:
Command tuning has seen several updates recently and incomplete regression testing. One issue introduced is that when tuning without a direction key will result in the tune applied in the TX direction instead of the block direction (intended). This fixes that by adding member getters for block direction as well as a number of minor updates to how this logic is handled internally.

Here is a flowgraph to test tuning: [usrp_cmd_tuning.grc.txt](https://github.com/gnuradio/gnuradio/files/6224941/usrp_cmd_tuning.grc.txt)

### DIFF
--- a/gr-uhd/lib/usrp_block_impl.h
+++ b/gr-uhd/lib/usrp_block_impl.h
@@ -227,6 +227,12 @@ protected:
 
     //! Stores the individual command handlers
     ::uhd::dict<pmt::pmt_t, cmd_handler_t> _msg_cmd_handlers;
+
+    //! Will check a command for a direction key, if it does not exist this will use
+    // the default value for the block via _direction() defined below
+    const pmt::pmt_t get_cmd_or_default_direction(const pmt::pmt_t& cmd) const;
+    //! Block direction overloaded by block impl to return "RX"/"TX" for source/sink
+    virtual const pmt::pmt_t _direction() const = 0;
 };
 
 } /* namespace uhd */
--- a/gr-uhd/lib/usrp_source_impl.h
+++ b/gr-uhd/lib/usrp_source_impl.h
@@ -146,6 +146,8 @@ private:
     double _samp_rate;
 
     std::recursive_mutex d_mutex;
+
+    const pmt::pmt_t _direction() const override { return ant_direction_rx(); };
 };
 
 } /* namespace uhd */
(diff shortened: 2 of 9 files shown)


willcode:
This looks like it will have the right effect. Having `get_direction(cmd)` return an direction when there isn't one in the cmd ... maybe a slight renaming of things (`get_direction_or_default_based_on_block_type()`) :-)  ... or a comment line everywhere it's used will save some confusion down the road.

For the 3.9 backport, we can just skip the `ant_` renames, even though the new names make more sense.

jacobagilbert:
I'm going to split this into two commits for backport reasons. I'll update the function name too.

gr-uhd/lib/usrp_block_impl.h
```diff
@@ -227,6 +227,12 @@ class usrp_block_impl : virtual public usrp_block
 
     //! Stores the individual command handlers
     ::uhd::dict<pmt::pmt_t, cmd_handler_t> _msg_cmd_handlers;
+
+    //! Will check a command for a direction key, if it does not exist this will use
+    // the default value for the block via _direction() defined below
+    const pmt::pmt_t get_cmd_or_default_direction(pmt::pmt_t cmd);
+    //! Block direction overloaded by block impl to return "RX"/"TX" for source/sink
+    virtual const pmt::pmt_t _direction() = 0;
```
mbr0wn, line 235: Did you mean
```suggestion
    pmt::pmt_t get_cmd_or_default_direction(pmt::pmt_t cmd) const;
    //! Block direction overloaded by block impl to return "RX"/"TX" for source/sink
    virtual pmt::pmt_t _direction() const = 0;
```
?

If not, then please make this a change request (those methods shoulds be `const`)

gr-uhd/lib/usrp_source_impl.h
```diff
@@ -146,6 +146,9 @@ class usrp_source_impl : public usrp_source, public usrp_block_impl
     double _samp_rate;
 
     std::recursive_mutex d_mutex;
+
+    // returns "RX" PMT string for the source
+    const pmt::pmt_t _direction() override;
```
mbr0wn, line 151: You don't have the comment in the usrp_sink -- I think it's also not necessary here.

mbr0wn (approved):
Please make the new methods `const`. Otherwise, this is fine. Many thanks!

gr-uhd/lib/usrp_source_impl.h
jacobagilbert: Agree

PR #4476, opened 2021-04-01
cmake: Fix codec2 and MPIR find modules for Windows (conda).

ryanvolz:
Tweak the CMake find modules for codec2 and MPIR so that they work on Windows (at least when using conda packages). Previously it was necessary to specify paths to these libraries/headers with the cmake command so that they could be used.

### DIFF
--- a/cmake/Modules/FindCodec2.cmake
+++ b/cmake/Modules/FindCodec2.cmake
@@ -17,17 +17,16 @@ INCLUDE(FindPackageHandleStandardArgs)
 pkg_check_modules(LIBCODEC2_PKG QUIET codec2)
 
 find_path(LIBCODEC2_INCLUDE_DIR NAMES codec2.h
+  HINTS ${LIBCODEC2_PKG_INCLUDE_DIRS}
   PATHS
-  ${LIBCODEC2_PKG_INCLUDE_DIRS}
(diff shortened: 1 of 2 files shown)


cmake/Modules/FindCodec2.cmake
```diff
@@ -17,17 +17,16 @@ INCLUDE(FindPackageHandleStandardArgs)
 pkg_check_modules(LIBCODEC2_PKG QUIET codec2)
 
 find_path(LIBCODEC2_INCLUDE_DIR NAMES codec2.h
+  HINTS ${LIBCODEC2_PKG_INCLUDE_DIRS}
```
noc0lour, line 20: I guees this might be a good change in other cmake modules as well. Only hardcoded paths should go into `PATHS`.

noc0lour (approved):
Looks good.

cmake/Modules/FindCodec2.cmake
ryanvolz: Yes, although practically it probably doesn't make too much of a difference. For anyone curious, the practical difference between HINTS and PATHS is that CMake will search standard system paths before PATHS but after HINTS.

mormj (approved):

PR #4477, opened 2021-04-01
cmake: Move swath of add_definitions to gnuradio-runtime target.

ryanvolz:
I think every target affected by the top-level add_definitions links to gnuradio-runtime, so functionally this should be the same. Besides being a more modern CMake style, this has the benefit of propagating the public definitions to any consumers of the gnuradio-runtime library, which includes OOT modules that imported their targets through CMake. For example, OOT modules would previously have to know to define "NOMINMAX" on MSVC, but now that happens transitively. This should make it a little more friendly to build OOT modules, on Windows at least.

This also takes the opportunity to rearrange `gnuradio-runtime/lib/CMakeLists.txt` to what I find is a more logical order, but I can back that out if the diff is deemed too messy.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -167,9 +167,6 @@ include(GrMiscUtils) #compiler flag check
 
 include(TestBigEndian)
 TEST_BIG_ENDIAN(GR_IS_BIG_ENDIAN)
-if(GR_IS_BIG_ENDIAN)
-    add_definitions(-DGR_IS_BIG_ENDIAN)
-endif(GR_IS_BIG_ENDIAN)
 
 if(CMAKE_CXX_COMPILER_ID MATCHES "Clang" OR
    CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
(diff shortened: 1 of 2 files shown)


noc0lour (approved):
I think this is a nice change. 

We need to be aware of us pushing a lot of more definitions on our down-stream libraries (OOT) which they might not want to have.

I agree with the changes to performance counters & ctrlport, very nice so OOTs don't have do define them again. Although this might mean for bigger distributions which enable everything all OOTs will autobuild these features always.

For Windows I'm trusting your judgement of improved user experience for building OOTs.

mormj (approved):

willcode:
@ryanvolz @noc0lour can you give me some guidance on the risk of porting this change (and similar ones in the PR queue) to 3.8? I know little about Windows dev, and not enough about cmake. Since 3.9 cmake stuff is close to master, it seems reasonable to backport that far.

ryanvolz:
I can start with the data point that these couple recent PRs by me have already been deployed for 3.9 and 3.8 in the conda-forge packages, and I have seen no ill effects for subsequently building and using `gr-osmosdr`, `gr-soapy`, `gr-satellites`, and `gqrx`. Indeed, they have allowed me to simplify those recipes by removing some of the tweaks that these changes cover.

External projects should only see a few new flags inherited through the `gnuradio-runtime` target at build time, and in my experience they either will have already needed to set these flags manually to get it to build anyway or they won't have an effect for that particular application. From my perspective, I think it's safe to backport to at least 3.8, but it would be good for someone else to weigh in and verify my thinking.

noc0lour:
Even though these changes may not intefere with API/build stability of 3.8 my personal opinion would be to maybe backport only to 3.9 to provide more convenience for OOTs and people packaging OOTs. For 3.8 packagers already have stable build scripts for GNU Radio and environment and have figured out the tweaks to get a results they like.

Issue #4475, opened 2021-04-01
Fresh install, AttributeError: type object 'audio' has no attribute 'sink'


PenelopeFudd:
I'm a new GnuRadio user, and I'm having a problem:

  *  Fedora 33 system
  *  [Installed Gnuradio according to the Gnuradio wiki](https://wiki.gnuradio.org/index.php/InstallingGR#Fedora_COPR_Installation)
  *  Tried to hook a signal source to an audio sink and run it
  *  Got the error "AttributeError: type object 'audio' has no attribute 'sink'"

Here are the commands I used to install it:
```
    sudo dnf copr enable gnuradio/gnuradio-master  
    sudo dnf install gnuradio  # gnuradio-3.8.2.0-2.fc33.x86_64  python3-gnuradio-3.8.2.0-2.fc33.x86_64
    gnuradio-companion  
```
 
Here's how I triggered the error:

* I dragged a Signal Source and an Audio Sink onto the field, and hooked them together
* It gave a flowgraph error (red arrow) because the signal source was set to 'Complex', while the audio sink was set to 'Integer'.
* When I set the signal source to Integer, the error went away (black arrow), and the Run>Generate and Run>Execute menu options enabled themselves.
* I generated and ran it.

Doing so gave the following error, both in and out of the GUI:
```
    Executing: /usr/bin/python3 -u /home/user/work/gr/audio.py   
      
    Warning: failed to XInitThreads()   
    QSocketNotifier: Can only be used with threads started with QThread   
    Traceback (most recent call last):   
      File "/home/user/work/gr/audio.py", line 135, in <module>   
        main()   
      File "/home/user/work/gr/audio.py", line 111, in main   
        tb = top_block_cls()   
      File "/home/user/work/gr/audio.py", line 77, in __init__   
        self.audio_sink_0 = audio.sink(samp_rate, '', True)   
    AttributeError: type object 'audio' has no attribute 'sink'   
      
    >>> Done (return code 1)  
```

Here is the python program it generated:
```
    #!/usr/bin/env python3  
    # -*- coding: utf-8 -*-  
      
    #  
    # SPDX-License-Identifier: GPL-3.0  
    #  
    # GNU Radio Python Flow Graph  
    # Title: Not titled yet  
    # GNU Radio version: 3.8.2.0  
      
    from distutils.version import StrictVersion  
      
    if __name__ == '__main__':  
        import ctypes  
        import sys  
        if sys.platform.startswith('linux'):  
            try:  
[...]

PR #4480, opened 2021-04-05
Blacklist audio

dl1ksv:
The id audio should be blacklisted , too
This fixes the attribute error in #4475 

### DIFF
--- a/grc/core/params/dtypes.py
+++ b/grc/core/params/dtypes.py
@@ -13,7 +13,7 @@ from .. import Constants
 
 
 # Blacklist certain ids, its not complete, but should help
-ID_BLACKLIST = ['self', 'options', 'gr', 'math', 'firdes', 'default'] + dir(builtins)
+ID_BLACKLIST = ['self', 'options', 'gr', 'math', 'firdes', 'default', 'audio'] + dir(builtins)
 try:
     from gnuradio import gr
     ID_BLACKLIST.extend(attr for attr in dir(gr.top_block()) if not attr.startswith('_'))


grc/core/params/dtypes.py
```diff
@@ -13,7 +13,7 @@
 
 
 # Blacklist certain ids, its not complete, but should help
-ID_BLACKLIST = ['self', 'options', 'gr', 'math', 'firdes', 'default'] + dir(builtins)
+ID_BLACKLIST = ['self', 'options', 'gr', 'math', 'firdes', 'default','audio'] + dir(builtins)
```
mormj, line 16: super minor, but would you mind adding the space before `audio` to be consistent?

willcode:
Wonder if a better solution would be to make the top block name something like `block_{id}`, so if the user made the flowgraph `audio`, the top block would be `block_audio`. Or something similar. There are an awful lot of ids that would cause a problem otherwise - any Python or GR package name, any variable name used in an embedded block or snipped, ...

haakov (approved):

haakov:
@willcode I like that idea! We should probably call it `top_block_audio` or `tb_audio` to show that it is not one of the actual blocks

mormj:
> @willcode I like that idea! We should probably call it `top_block_audio` or `tb_audio` to show that it is not one of the actual block

I like tb_audio for conciseness

skoslowski:
I always liked that the block ID end up being their names in Python, makes consuming the resulting code very transparent.
Same goes for referring to other blocks within the FG. 

I am opposed to prefixing. To me this feels like a work-around, not a solution.
GRC runs all the imports of the FG anyway. Why not blacklist all names defined at that point?

willcode:
GRC could detect many of the potential collisions that way. It's not great, from a user perspective, to say "you can't use that name" until they pick a good one. Is there something we could do with a Python decorator that would allow use of any name, but internally munge it?

skoslowski:
Yes, it would detect all collisions that way - for Python Flowgraphs, that
is.

To avoid the "until a good one is picked" scenario GRC could suggest a
name, following the same scheme it has for picking the initial id.
Maybe I am wrong here, but users who don't care about the generated code
don't even have to touch block ids for the most part.
Variables, however are another matter, but again, do we really want to
rewrite the users' expressions adding prefixes to variables.

dl1ksv:
Statically blacklisting and prefixing is a kind of workaround and not a real  solution, as you can always construct cases where they don't work.  I clean solution  could be to  do all the imports in grc while generating the flowgraph and building a dynamic blacklist depending on these imports.

dl1ksv:
An alternative solution that reflects the imports is #4522

mormj:
Closing in favor of #4522

PR #4483, opened 2021-04-05
grc: Check connection types, whether they match

dl1ksv:
At the moment grc treats a connection int -> float as valid as the itemsizes match. See #4475 
This patch checks the dtype of a port. Connections are only valid if the dtype's match


willcode (approved):

mormj (approved):

mormj:
@dl1ksv - would you mind updating the commit message to start with `grc: ...`

dl1ksv:
@mormj Done

Issue #2763, opened 2019-08-26
grc: global_blocks_path prevents switching gr versions


mormj:
When going back and forth between GNU Radio versions (e.g. 3.7 and 3.8), GRC sometimes gets stuck looking for blocks for the other gr version.  GR 3.7 crashes when it gets stuck looking for the 3.8 blocks because of the format change.  

Here is the error
`
Traceback (most recent call last):
  File "/share/gnuradio/gr37/bin/gnuradio-companion", line 99, in <module>
    run_main()
  File "/share/gnuradio/gr37/bin/gnuradio-companion", line 92, in run_main
    exit(main())
  File "/share/gnuradio/gr37/lib/python2.7/dist-packages/gnuradio/grc/main.py", line 51, in main
    install_prefix=gr.prefix()
  File "/share/gnuradio/gr37/lib/python2.7/dist-packages/gnuradio/grc/gui/Platform.py", line 38, in __init__
    _Platform.__init__(self, *args, **kwargs)
  File "/share/gnuradio/gr37/lib/python2.7/dist-packages/gnuradio/grc/core/Platform.py", line 75, in __init__
    self.build_block_library()
  File "/share/gnuradio/gr37/lib/python2.7/dist-packages/gnuradio/grc/core/Platform.py", line 175, in build_block_library
    hide_bokeh_gui_options_if_not_installed(self.blocks['options'])
  File "/share/gnuradio/gr37/lib/python2.7/dist-packages/gnuradio/grc/core/utils/odict.py", line 35, in __getitem__
    return self._data[key]
KeyError: 'options'
`

Workaround:
- in ~/.gnuradio/grc.conf
- comment out global_blocks_path ...

Not sure exactly what the correct solution is, but I think it would be preferable instead of global options in one .conf file, for the conf file to be local to or tied to the pybombs installation.

schneider42:
Same happens for me the other way round:

(System installed) GRC 3.7 seem to have added `global_blocks_path` to `~/.gnuradio/config.conf` which makes GRC 3.8 crash on start:
```
Traceback (most recent call last):
  File "/home/schneider/gr-38/bin/gnuradio-companion", line 102, in <module>
    run_main()
  File "/home/schneider/gr-38/bin/gnuradio-companion", line 95, in run_main
    exit(main())
  File "/home/schneider/gr-38/lib/python3/dist-packages/gnuradio/grc/main.py", line 83, in main
    platform.build_library()
[...]

PR #4502, opened 2021-04-06
grc: Remove global_blocks_path preference and use prefix path

ryanvolz:
Addresses #2763.

Instead of storing the default (built-in) blocks path in the preferences, store the path relative to the prefix (`GRC_BLOCKS_DIR`) in a `constants.yml` file and evaluate the full blocks path at runtime by joining with the installation prefix. This should make it easier to switch between different versions by making it so the user preference for `global_blocks_path` doesn't interfere. Users still have `local_blocks_path` preference and the `GRC_BLOCKS_PATH` environment variable for pointing to additional blocks.

This PR also includes a CMake tweak to `GR_PYTHON_INSTALL` that is no longer necessary for this PR to work but adds what looks to be expected functionality for `FILES_MATCHING REGEX`. It also includes a commit to make the cause of the `KeyError: 'options'` error more clear (even though the removal of `global_blocks_path` should do a lot to prevent that error from happening).

I tested running `gnuradio-companion` on both Linux and Windows and the blocks load fine, avoiding a bad or empty `global_blocks_path` preference as intended. It would be great for someone more familiar with this part of the code to double-check to make sure I haven't broken anything unintentionally.

### DIFF
--- a/cmake/Modules/GrPython.cmake
+++ b/cmake/Modules/GrPython.cmake
@@ -168,6 +168,7 @@ function(GR_PYTHON_INSTALL)
     install(
       FILES ${GR_PYTHON_INSTALL_FILES}
       DESTINATION ${GR_PYTHON_INSTALL_DESTINATION}
+      ${GR_PYTHON_INSTALL_UNPARSED_ARGUMENTS}
     )
 
         #create a list of all generated files
@@ -228,6 +229,7 @@ function(GR_PYTHON_INSTALL)
     install(
       DIRECTORY ${GR_PYTHON_INSTALL_DIRECTORY}
       DESTINATION ${GR_PYTHON_INSTALL_DESTINATION}
+      ${GR_PYTHON_INSTALL_UNPARSED_ARGUMENTS}
     )
 
 
@@ -334,6 +336,7 @@ function(GR_PYTHON_INSTALL)
 
             install(PROGRAMS ${pyexefile} RENAME ${pyfile_name}
                 DESTINATION ${GR_PYTHON_INSTALL_DESTINATION}
+                ${GR_PYTHON_INSTALL_UNPARSED_ARGUMENTS}
             )
         endforeach(pyfile)
 
--- a/grc/CMakeLists.txt
+++ b/grc/CMakeLists.txt
@@ -80,14 +80,8 @@ GR_REGISTER_COMPONENT("gnuradio-companion" ENABLE_GRC
 if(ENABLE_GRC)
 
 ########################################################################
-# Create and install the grc and grc-docs conf file
+# Create and install the grc and grc-docs conf files and constants YAML
 ########################################################################
-file(TO_NATIVE_PATH ${CMAKE_INSTALL_PREFIX}/${GRC_BLOCKS_DIR} blocksdir)
-if(CMAKE_INSTALL_PREFIX STREQUAL "/usr")
-    # linux binary installs: append blocks dir with prefix /usr/local
-    set(blocksdir ${blocksdir}:/usr/local/${GRC_BLOCKS_DIR})
-endif(CMAKE_INSTALL_PREFIX STREQUAL "/usr")
-
 if(UNIX)
     find_program(GRC_XTERM_EXE
         NAMES xterminal-emulator gnome-terminal konsole xterm
@@ -119,6 +113,13 @@ configure_file(
 install(
     FILES ${CMAKE_CURRENT_BINARY_DIR}/00-grc-docs.conf
     DESTINATION ${GR_PREFSDIR}
+)
+
+configure_file(core/constants.yml.in core/constants.yml @ONLY)
+
+install(
+    FILES "${CMAKE_CURRENT_BINARY_DIR}/core/constants.yml"
+    DESTINATION "${GR_PYTHON_DIR}/gnuradio/grc/core"
 )
 
 ########################################################################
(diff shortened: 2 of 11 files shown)


willcode:
I like the idea. Have not tried it out yet. What's the /usr -> /usr/local bit about?

ryanvolz:
> I like the idea. Have not tried it out yet. What's the /usr -> /usr/local bit about?

It's replacing this from `grc/CMakeLists.txt`:
```
if(CMAKE_INSTALL_PREFIX STREQUAL "/usr")
    # linux binary installs: append blocks dir with prefix /usr/local
    set(blocksdir ${blocksdir}:/usr/local/${GRC_BLOCKS_DIR})
endif(CMAKE_INSTALL_PREFIX STREQUAL "/usr")
```
I don't know if anybody still relies on that, but I put it in just in case.

ryanvolz:
I see we have some GRC test failures. I think it's probably because the test is running from the `grc` Python package in the source directory which no longer has `core/Constants.py` (it is generated in the build directory and then installed)? Any thoughts on how to handle that? Obviously I don't want to generate the `Constants.py` right in the source directory because then it won't be cleaned up and is liable to be checked in to git or cause confusion.

skoslowski:
I like the idea, but I am strongly opposed to sacrificing the ability to start GRC from the source tree in the process.
A way out would be to not rename `Constants.py` - it still is valid Python code after all.
If `@GRC_BLOCKS_DIR@` shows up at runtime it could/should be replaced with `grc.core.__file__.replace("__init__.py", "blocks")`. This way anyone can start tinking with GRC code without the need to setup a full build env.

ryanvolz:
I was finally able to wrap my head around how the GRC pytest tests work, with the various paths at play, and getting it to work out of the build dir is messy. So I agree that we should preserve the ability to start GRC from the source tree, not only for users to tinker easily but also for ease of keeping the tests working.

This has given me an idea of a new approach to try, so I'll update this PR soon.

ryanvolz:
Ok, new approach: don't configure `grc/core/Constants.py` and instead have it load `GRC_BLOCKS_DIR` from a YAML file that gets configured at build time and installed along with the Python module. When the configuration YAML is not present, such as when running from the source dir, it will fall back to the path to the blocks in the source dir.

ryanvolz:
Now that this has sat for a while, I don't love loading `GRC_BLOCKS_DIR` from yet another file. This approach made more sense when that wasn't necessary, but that makes it almost impossible to debug GRC from the source tree. Adding a file that provides this information seems like duplicating what the configuration system was meant to solve, it's just that it currently does a poor job of it.

On that note, I'll bring this discussion over from #4504 since it's relevant. I said:

> It has always bugged me that the preferences will be loaded from global and local sources, but when saving the local preferences it writes out all of the values regardless of where they were loaded from. This leads to the situation where you get the global preferences for a clean install, but as soon as the preferences are saved you will always have all of the preferences loaded from that local configuration file. So if you then install to another prefix, or if you upgrade and some of the defaults change, you don't see any of the corresponding preference changes even if you never intended to override those particular values. This is felt most painfully with the global_blocks_path preference that makes it painful to switch between different installs or versions, e.g. #2763. (Coincidentally I took another tactic in #4502 rather than touching the preferences code, but that doesn't mean this couldn't also be better.) Do you have any thoughts on how significant a re-structure it would be to keep the global preferences separate from the locally-set preferences so that setting a preference and saving only writes the new value for that preference locally?

@ThomasHabets responded:
> Hmm… that shouldn't be too hard. Just tag the settings with where they came from. I think I want to put it in a separate PR, though.

I would much prefer a fix as described in my quote. Basically, almost nobody would ever have a need to set `global_blocks_path` in their config, and if the value was pulled from the default configuration that is installed in `$PREFIX/etc/gnuradio/conf.d` then things would be fine. The trouble comes from the fact that GR only thinks in terms of a monolithic configuration; then when that configuration is saved, the value for `global_blocks_path` gets written to the local configuration file and overrides the installed default forever. I think the first step to solving this lies in keeping track of which settings the user has supplied and which came from the default configuration and not mixing the two. Complications with versioned configs and storing to `XDG_CONFIG_HOME` are not necessary to just fixing that major sticking point, and while those would be nice I think we'd get pretty far with a simpler change.

mormj:
> while those would be nice I think we'd get pretty far with a simpler change.

@ryanvolz - are you recommending this PR as the simpler change, or something else?

ryanvolz:
> > while those would be nice I think we'd get pretty far with a simpler change.
> 
> @ryanvolz - are you recommending this PR as the simpler change, or something else?

No, I meant simpler than going to a full versioned config file setup and implementing other desired features like `XDG_CONFIG_HOME`.

I'm happy to re-work this PR a bit if someone sees an implementation that would be acceptable, but I think another direction modifying the config file hierarchy would be better. I'm just not likely to be the person to implement that better PR since it gets into touching more of the C++ code than I'm comfortable with. Thus I leave this PR as an option if the pain of #2763 is judged to be enough to require a less-than-ideal fix.

willcode:
@ryanvolz We haven't revisited this one in a while - is it something we still want to look at?

ryanvolz:
I don't love where this ended up, with installing an extra yaml file to hold a single compile-time constant. It still feels like that job is better handled by an improved config system that doesn't mix installation defaults and values explicitly set by the user. Long term I think that's what needs to happen.

But maybe there is another way to implement this that is closer to my original proposal (and what is still being used in the conda package) while still allowing GRC to be run from the source tree. I'll give it a deeper look soon.

willcode:
Closing this PR for now since it's pretty old.

Issue #4505, opened 2021-04-07
Default embeded python code block doesn't work when parameters are used to specify input/output sizes


tclarke:
Example:
```
def __init__(self, vector_size=1):
    gr.basic_block.__init__(self, name="foo", in_sig=[(np.float32, vector_size)], out_sig=[])
    self.vector_size=vector_size
```
This will use the default vector_size=1 when GRC calculates input sizes. The self.vector_size=vector_size needs to happen before the base class __init__ is called.

dave-pi:
Just to clear what happens because it can be confusing. When you are designing the flowgraph, the only thing that matters is the value of the length in the epy block construtor. This can come from a literal hardcoded value, a parameter, an instance attribute, a local variable, whatever you want as long as the value passed in `in_sig` is correct when the constructor is executed (which happens when you press OK in the dialog box or connect arrows). These are all valid:
```
def __init__(self, example_param=1.0, foo=9999):  # only default arguments here
        """arguments to this function show up as parameters in GRC"""
        local_foo = 9999
        self.foo = 9999
        gr.basic_block.__init__(
            self,
            name='Embedded Python Block',   # will show up in GRC
            in_sig=[(np.complex64, 1024)],
            out_sig=[np.complex64]
        )
```
```
def __init__(self, example_param=1.0, foo=1024):  # only default arguments here
        """arguments to this function show up as parameters in GRC"""
        local_foo = 9999
        self.foo = 9999
        gr.basic_block.__init__(
            self,
            name='Embedded Python Block',   # will show up in GRC
            in_sig=[(np.complex64, foo)],
            out_sig=[np.complex64]
        )
```
```
def __init__(self, example_param=1.0, foo=9999):  # only default arguments here
        """arguments to this function show up as parameters in GRC"""
        local_foo = 1024
        self.foo = 9999
        gr.basic_block.__init__(
            self,
            name='Embedded Python Block',   # will show up in GRC
            in_sig=[(np.complex64, local_foo)],
            out_sig=[np.complex64]
        )
```
```
def __init__(self, example_param=1.0, foo=1024):  # only default arguments here
        """arguments to this function show up as parameters in GRC"""
        local_foo = 9999
        self.foo = 1024
        gr.basic_block.__init__(
            self,
            name='Embedded Python Block',   # will show up in GRC
            in_sig=[(np.complex64, self.foo)],
            out_sig=[np.complex64]
        )
```
```
[...]

PR #4506, opened 2021-04-07
Fixed embedded python block and documentation to ensure in/out sizes …

tclarke:
…using parameters are calculated correctly.

This fixes issue #4505


dkozel (commented):
I'm not sure this is necessary or desirable. The discussion in the issue is worth carrying on to the root cause before merging.

mormj:
Closing based on feedback here and in #4196

PR #4513, opened 2021-04-09
analog: PLL Block: better block naming, better Parameter labels

marcusmueller:
Also, minimally refactored pll_freqdet_cf_impl.h, which is a header and has no business including private method implementations.


mormj (approved):
LGTM!

Issue #4514, opened 2021-04-09
Source IO / Sink IO Compatibility


cclamb:
Hey everybody, it seems like the commit 7628a8f5a351faff88baa6de88f634b10aff3dfd in maint-38 is causing some issues. Specifically, whenever I try to attach a complex source to a sink on a USRP block, I end up getting an incompatible port error. Specifically, 'Source IO type "fc32" does not match sink IO type "complex"'. I've attached a flowgraph to demonstrate this - this installation is a fresh pybombs install from maint-38 with UHD-40 (though this failure also occurs at UHD-3.15 LTS).
[test_grc.zip](https://github.com/gnuradio/gnuradio/files/6288802/test_grc.zip)

willcode:
@mormj @mbr0wn @dl1ksv GRC does not know that "alias" types are the same as "core" types, so "fc32" is not the same as 'complex". Should we try to fix this or revert in all branches?

willcode:
Did not actually try it, but something like this could work:
```
diff --git a/grc/core/Connection.py b/grc/core/Connection.py
index 6fd78a41f..ddf402613 100644
--- a/grc/core/Connection.py
+++ b/grc/core/Connection.py
@@ -8,6 +8,7 @@ SPDX-License-Identifier: GPL-2.0-or-later
 
 
 from .base import Element
+from .Constants import ALIAS_OF
 from .utils.descriptors import lazy_property
 
 
@@ -92,7 +93,7 @@ class Connection(Element):
 
         source_dtype = self.source_port.dtype
         sink_dtype = self.sink_port.dtype
-        if source_dtype != sink_dtype:
+        if source_dtype != sink_dtype and source_dtype != ALIAS_OF(sink_dtype):
             self.add_error_message('Source IO type "{}" does not match sink IO type "{}".'.format(source_dtype, sink_dtype))
 
         source_size = self.source_port.item_size
diff --git a/grc/core/Constants.py b/grc/core/Constants.py
index b5b7c270c..65b4ac2ab 100644
--- a/grc/core/Constants.py
+++ b/grc/core/Constants.py
@@ -115,5 +115,21 @@ ALIAS_TYPES = {
     'bits':    (1, GRC_COLOR_PURPLE_A100),
 }
 
+ALIAS_OF = {
+    'complex': 'fc32',
+    'float': 'f32',
+    'int': 's32',
+    'short': 's16',
+    'byte': 's8',
+    'bits': 'bit',
+
[...]

PR #4516, opened 2021-04-10
grc: match data type aliases as well as types

willcode:
A recent fix to validate connections by type, rather than by sizeof(type), exposed the fact that GRC did not
handle type aliases.

Fixes https://github.com/gnuradio/gnuradio/issues/4514


mormj (approved):

mormj:
LGTM.  Thanks for catching this one.

PR #4522, opened 2021-04-11
grc: update id blackist with imports

dl1ksv:
This is an alternative solution for #4480 
It uses the import statements to update the blacklist for id's.

That is: If you don't

from gnuradio import audio

you can use audio as id. Otherwise it is blacklisted

### DIFF
--- a/grc/core/params/dtypes.py
+++ b/grc/core/params/dtypes.py
@@ -13,7 +13,7 @@ from .. import Constants
 
 
 # Blacklist certain ids, its not complete, but should help
-ID_BLACKLIST = ['self', 'options', 'gr', 'math', 'firdes', 'default'] + dir(builtins)
+ID_BLACKLIST = ['self', 'default'] + dir(builtins)
 try:
     from gnuradio import gr
     ID_BLACKLIST.extend(attr for attr in dir(gr.top_block()) if not attr.startswith('_'))
@@ -32,19 +32,18 @@ def validates(*dtypes):
         return func
     return decorator
 
-
 class ValidateError(Exception):
     """Raised by validate functions"""
 
 
 @validates('id')
-def validate_block_id(param):
+def validate_block_id(param,black_listed_ids):
     value = param.value
     # Can python use this as a variable?
     if not re.match(r'^[a-z|A-Z]\w*$', value):
         raise ValidateError('ID "{}" must begin with a letter and may contain letters, numbers, '
                             'and underscores.'.format(value))
-    if value in ID_BLACKLIST:
+    if value in (black_listed_ids + ID_BLACKLIST):
         raise ValidateError('ID "{}" is blacklisted.'.format(value))
     block_names = [block.name for block in param.parent_flowgraph.iter_enabled_blocks()]
     # Id should only appear once, or zero times if block is disabled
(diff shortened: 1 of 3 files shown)


willcode:
I think we need a deterministic solution, even something as simple as requiring the id to start with a certain prefix. Better yet, a solution that allows the user to use whatever id they desire.

dl1ksv:
@willcode : I disagree. This solution is deterministic and works even if new modules are added. And if you ave to look for bugs the generated python code contains the id's the user has given his modules. If you modify the id's you can't overview the side effects. And you cannot allow the user to use id's whatever they desire as the may conflict with python rules.

willcode:
OK, it's better than using the blacklist in general. It could, however, surprise a user when an import gets added by a new block and an id is now invalid. Not a huge deal. Can this be done without using exec()?

dl1ksv:
I know no other way. This part of code is taken from Flowgraph.py

willcode:
Something like the following. This also gets rid of any performance issue with actually importing packages.
```
import ast

imports = """
import sys
import os, time
from gnuradio import blocks
from gnuradio import audio
from gnuradio import window, fft as some_other_name
"""

tree = ast.parse(imports)
for expr in tree.body:
    if type(expr) in (ast.Import, ast.ImportFrom):
        for name in expr.names:
            print(name.asname or name.name)
```

grc/core/params/dtypes.py
```diff
@@ -32,19 +32,18 @@ def decorator(func):
...
     return decorator
 
-
 class ValidateError(Exception):
     """Raised by validate functions"""
 
 
 @validates('id')
-def validate_block_id(param):
+def validate_block_id(param,bl):
```
skoslowski, line 40: little nit: `bt` is not a very (self-)descriptive name, why not use sth. like `black_listed_ids`.

grc/core/params/param.py
```diff
@@ -154,9 +154,24 @@ def validate(self):
...
+                try:
+                    exec(expr, bl)
+                except ImportError:
+                    # We do not have a good way right now to determine if an import is for a
+                    # hier block, these imports will fail as they are not in the search path
+                    # this is ok behavior, unfortunately we could be hiding other import bugs
+                    pass
+                except Exception:
+                    log.exception('Failed to evaluate import expression "{0}"'.format(expr), exc_info=True)
+                    pass
```
skoslowski, line 171: This re-creates the black list for every block - even though it does not change. My suggestion here is to have the FlowGraph class build that list as it creates the namespace (and runs the imports anyway). Then have getter, `get_imported_names()` which you can use below

skoslowski (changes requested):
This is great! Thanks @dl1ksv for the implementation.

grc/core/params/param.py
dl1ksv: Many thanks for your hints. This makes the code smoother

mormj (approved):

willcode:
Would it be useful to add `keyword.kwlist` to the list?

```
>>> keyword.kwlist
['and', 'as', 'assert', 'break', 'class', 'continue', 'def', 'del', 'elif', 'else', 'except', 'exec', 'finally', 'for', 'from', 'global', 'if', 'import', 'in', 'is', 'lambda', 'not', 'or', 'pass', 'print', 'raise', 'return', 'try', 'while', 'with', 'yield']
```

PR #4526, opened 2021-04-13
Windows build fixes (non-portable ssize_t, undefined HAVE_IO_H in gr-network)

ryanvolz:
These were turned up in #4410 by the in-progress conda package CI, and have been build tested there (see [here](https://github.com/gnuradio/gnuradio/runs/2327712180)).


gr-qtgui/lib/timeRasterGlobalData.cc
```diff
@@ -158,9 +159,9 @@ double TimeRasterData::value(double x, double y) const
 
     double _y = floor(top - y);
     double _loc = _y * (d_cols) + x + d_resid;
-    auto location = static_cast<ssize_t>(_loc);
+    auto location = static_cast<std::ptrdiff_t>(_loc);
```
ThomasHabets, line 162: Ugh, `ptrdiff`. It's signed, but it's not actually a pointer diff we're using it for.

How about just:

```c++
    const auto location = static_cast<size_t>(_loc);
    if (location < d_data.size()) {
        return d_data[location];
    }
    return 0.0;
``` 

(getting rid of the needless `returnValue` variable, since multiple `return` statements is not in fact an antipattern)

ThomasHabets (changes requested):
I'm not qualified to say anything about the cmake parts.

gr-qtgui/lib/timeRasterGlobalData.cc
ryanvolz: Great! I'll take whatever you say because I was just guessing here.

ThomasHabets (approved):
Thumbs up for the C++ parts.
No comment on the CMake part.

gr-qtgui/lib/timeRasterGlobalData.cc
ryanvolz: Confirming that this change has successfully built on the conda Windows CI PR.

willcode:
Backport to 3.9: taking only the cmake changes. The other change does not apply cleanly, and does not look like it's relevant since 3.9 uses int where master had been changed to ssize_t.

ryanvolz:
That makes sense. These each stemmed from separate recent changes on master and I doubt the first was backported to maint-3.9.

PR #4531, opened 2021-04-15
Fix variable not defined should show red

### LINKED ISSUE #4519: Unused variable not defined in flowgraph cause GNURadio crashing
Hi, 
I unintentionally created a flowgraph ([test_variable_not_defined.grc.txt](https://github.com/gnuradio/gnuradio/files/6289949/test_variable_not_defined.grc.txt) ) in which variable `test1` was defined as `2*ovesampling1`, but `oversampling1` was undefined (after I deleted `ovesampling1` variable block )

- In the flowgraph, `test1` block is not red, so no one can't suspect the error. In fact, as `test1` is not used, this can be considered as a correct flowgraph. 
- Generating the flowgraph succeed. 
- while running the flowgraph crash GNURadio because of line `self.test1 = test1 = 2*oversampling1`

So one may conclude that: 

- whether this flowgraph should be considered as erroneous, because it contains undefined variable , and `test1` block should be highlighted in red, and python code not generated
- whether we consider this flowgraph as correct (I think this solution is more satisfying), in that case unused variable(s) should not appears when rendering the template

Sorry for rising an issue coming from this "stupid" flowgraph, but users could as me obtain such flowgraph by partially deleting unused blocks coming from a larger flowgraph. 

Regards, Christophe 

P.S. I will try to investigate this and propose a PR
variable_not_defined

ChristopheSeguinot:
This fix #4519 
With this, in case of error in variable block:

- block title turn red
- error is clearly reported in GRC 

I'm not satisfied by the way the block title is flagged in red. This should probably be done in some block.rewrite(), but this block is not rewritten when an error is detected. 

Thank to Volker for guiding me to this solution.

### DIFF
--- a/grc/gui/VariableEditor.py
+++ b/grc/gui/VariableEditor.py
@@ -182,11 +182,16 @@ class VariableEditor(Gtk.VBox):
                 if block.is_variable:
                     # Evaluate the params
                     for key in block.params:
-                        evaluated = str(block.params[key].evaluate())
-                        # ensure that evaluated is a UTF-8 string
-                        # Old PMTs could produce non-UTF-8 strings
-                        evaluated = evaluated.encode('utf-8', 'backslashreplace').decode('utf-8')
-                        self.set_tooltip_text(evaluated)
+                        try: 
+                            evaluated = str(block.params[key].evaluate())
+                            # ensure that evaluated is a UTF-8 string
+                            # Old PMTs could produce non-UTF-8 strings
+                            evaluated = evaluated.encode('utf-8', 'backslashreplace').decode('utf-8')
+                            self.set_tooltip_text(evaluated)
+                        except Exception as error:
+                            block.params[key].add_error_message(str(error))
+                            self.set_tooltip_text(str(error))
+                            pass
 
                     # Evaluate the block value
                     try:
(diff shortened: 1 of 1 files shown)


drmpeg:
Branch is based on another branch. You have to get back to master when you create a new branch.

ChristopheSeguinot:
Sorry for the mistake!

ChristopheSeguinot:
Do not merge. 
I didn't seen that the variable block id is missing after editing the block and applying change! 
I must find the exact way to "rewrite" variable block.

dl1ksv:
Just did a fresh install  ( fedora 33 ) of your pr and it works when loading the flowgraph. Opening the erroneous variable is ok, but changing to the Documentation tab and changing back the the General tab, the error message is missing.
But the label 'Value' keeps red.

dl1ksv:
Same on Ubuntu 20.04 The 'Documentation' and 'Generated Code' tab delete the error message. After closing and reopening the variable block, the error messages appear again.

dl1ksv:
I just tested the latest commit on fedora 33. Opening the variable editor gives me 
first
Then clicking the Advanced tab and back to the General tab  shows
second
The variable is no longer marked as faulty.
Clicking OK marks the block as valid.

dl1ksv:
@ChristopheSeguinot What does happen if youclick the Advanced tab and then the General tab again ?

I checked this morning on fedora 34 and the effects are as before.

ChristopheSeguinot:
I tested #4629 (under GR 3.9.1 whitout the `black_listed_ids` parameter ). 
It solves this issue in the right way so the present PR can be abandonned

PR #4535, opened 2021-04-16
pkg: add manpages back to 3.9

mormj:
Since man pages in tree was added in tree - desirable to keep 3.9
representative of what is currently on this branch for in-tree packaging

This is required to build debian packages on maint-3.9 branch

Signed-off-by: Josh Morman <jmorman@perspectalabs.com>


dave-pi:
This is adding man pages with old 'hardcoded' versions to the source tree. Is it specific to 3.9? What is the plan going forward for packaging? Sorry, I've been out of the discussion but if there's anything I can chip in, since I've done the original man page bring-in (7397abea21af8424cae0ab91ddcef2b85e55e293).

mormj:
@dave-pi - so this is specific to 3.9.  In master it uses the automatically generated man pages, which was a great addition.  I see that mait has patched that PR into his official releases of 3.9, but I'm avoid patches in these in-tree packages because it is intended to reflect what is currently at the tip of the branch.  I guess the more accurate thing to do in that regard would be to not put these man pages in at all.

dave-pi:
I see. I don't fully understand how the packaging is done (I see there's a pkg-debian.sh presumably run as part of a github job, but don't understand where it's fetching the man pages from). If you can point me to some resources to understand how this stuff works woud be great. Also, If it's useful to port the dynamic man pages thingy to 3.9 I can have a crack at it. Let me know.

mormj:
This scheme uses the github job to kick off the packaging using the pkg-debian.sh script.  In the case here it uses the manpages buried under ./packaging (in this PR).  On master it uses the dynamic man page generation which is very nice.  Just wanted to keep the in-tree packaging here consistent with what is on the tip.  Backporting that feature could make since, since at least on Debian, it is patched in anyhow.  Would leave that up to @willcode

willcode (approved):

PR #4541, opened 2021-04-18
cmake: Move swath of add_definitions to gnuradio-runtime target. (backport to maint-3.9)

willcode:
Besides being a more modern CMake style, this has the benefit of
propagating the public definitions to any consumers of the
gnuradio-runtime library, which includes OOT modules. For example, OOT
modules would previously have to know to define "NOMINMAX" on MSVC, but
now that happens transitively.

Signed-off-by: Ryan Volz <ryan.volz@gmail.com>
Patch pulled directly from ryanvolz tree.

Backport https://github.com/gnuradio/gnuradio/pull/4477

### DIFF
--- a/gnuradio-runtime/lib/CMakeLists.txt
+++ b/gnuradio-runtime/lib/CMakeLists.txt
@@ -122,6 +122,42 @@ target_sources(gnuradio-runtime PRIVATE
   ${MATH_SINE_TABLE_HEADER}
   )
 
+target_link_libraries(gnuradio-runtime PUBLIC
+  gnuradio-pmt
+  Volk::volk
+  Boost::program_options
+  Boost::filesystem
+  Boost::system
+  Boost::regex
+  Boost::thread
+  Log4Cpp::log4cpp
+  MPLib::mplib
+  ${libunwind_LIBRARIES}
+  # INTERFACE/PRIVATE split so users of the library can choose how to link to Python
+  # (importantly, extension modules can avoid linking against Python and resolve
+  #  their own Python symbols at runtime through the Python interpreter's linking)
+  INTERFACE
+    Python::Module
+  PRIVATE
+    Python::Python
+  )
+
+target_include_directories(gnuradio-runtime
+  PUBLIC
+    ${PYTHON_NUMPY_INCLUDE_DIR}
+    ${pybind11_INCLUDE_DIR}
+    $<INSTALL_INTERFACE:include>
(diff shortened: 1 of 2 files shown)


willcode:
@ryanvolz need help with the 3.9 backport.

gnuradio-runtime/lib/CMakeLists.txt
```diff
@@ -122,6 +122,47 @@ target_sources(gnuradio-runtime PRIVATE
...
+  # (importantly, extension modules can avoid linking against Python and resolve
+  #  their own Python symbols at runtime through the Python interpreter's linking)
+  INTERFACE
+    Python::Module
+  PRIVATE
+    Python::Python
+  )
+
+# Address linker issues with std::filesystem on Centos 8 and Debian
+target_link_libraries(gnuradio-runtime PUBLIC $<$<AND:$<CXX_COMPILER_ID:GNU>,$<VERSION_LESS:$<CXX_COMPILER_VERSION>,9.0>>:stdc++fs>)
```
ryanvolz, line 145: This should be dropped for 3.9.

gnuradio-runtime/lib/CMakeLists.txt
```diff
@@ -122,6 +122,47 @@ target_sources(gnuradio-runtime PRIVATE
...
+    $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/../include>
+    $<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}/../include>
+  PRIVATE
+    ${CMAKE_CURRENT_SOURCE_DIR}
+    )
+
+target_compile_definitions(gnuradio-runtime PUBLIC ${MPLIB_DEFINITIONS})
+
+# constants.cc includes boost::dll headers, force them to use std::filesystem
+target_compile_definitions(gnuradio-runtime PRIVATE BOOST_DLL_USE_STD_FS)
```
ryanvolz, line 161: And this should be dropped for 3.9.

ryanvolz (approved):
My branch is here: https://github.com/ryanvolz/gnuradio/tree/cmake_target_definitions_3.9. I took this straight from the patch that have for conda-forge vs. 3.9.1.0, and it doesn't look like there were any changes to maint-3.9 since then that would mess it up. I think my comments note the only differences, so you can either remove those bits or grab it from my branch, whichever is easier.

PR #4542, opened 2021-04-18
filter: filter_delay_fc documentation used to be unclear about delay

marcusmueller:
This clarifies that the block does not, indeed, construct a
delay-compensating all-pass, but just takes half the length of the
passed taps vector and uses that as delay.

This bit a student of mine; seeing she's rather capable on her own, this
indicated a need for improved documentation.


marcusmueller:
The Wiki needs updating, too.

mormj (approved):

duggabe (commented):
I will put these changes into the Filter_Delay wiki page.

mormj (approved):

PR #4543, opened 2021-04-18
GRC: Make deprecated blocks appear in orange on Canvas

marcusmueller:
To make it clear that a block is deprecated, blocks appear with an orange
border on the canvas, and, if enabled, with a light warm yellow background.

deprecated

I hope this reduces the effort needed to spot problems in flow graphs.

For this to work, I gave the core block (not the GUI block) a property
"deprecated". For now, this just searches (case-normalized) categories for the
string "deprecated", but in case we'd want to make that a tag (like
"throttling/hardware"), separating logic from rendering seemed wise.

### DIFF
--- a/grc/core/blocks/block.py
+++ b/grc/core/blocks/block.py
@@ -1,5 +1,6 @@
 """
 Copyright 2008-2020 Free Software Foundation, Inc.
+Copyright 2021 GNU Radio contributors
 This file is part of GNU Radio
 
 SPDX-License-Identifier: GPL-2.0-or-later
@@ -37,7 +38,7 @@ class Block(Element):
 
     key = ''
     label = ''
-    category = ''
+    category = []
     vtype = '' # This is only used for variables when we want C++ output
     flags = Flags('')
     documentation = {'': ''}
@@ -74,6 +75,7 @@ class Block(Element):
 
         self.states = {'state': True, 'bus_source': False, 'bus_sink': False, 'bus_structure': None}
         self.block_namespace = {}
+        self.deprecated = self.is_deprecated()
 
         if Flags.HAS_CPP in self.flags and self.enabled and not (self.is_virtual_source() or self.is_virtual_sink()):
             # This is a workaround to allow embedded python blocks/modules to load as there is
@@ -434,8 +436,8 @@ class Block(Element):
                 if _vtype == None:
                     _vtype = type(evaluated)
             except ValueError or SyntaxError as excp:
-                    if _vtype == None:
-                        print(excp)
+                if _vtype == None:
+                    print(excp)
 
             if _vtype in [int, float, bool, list, dict, str, complex]:
                 if _vtype == (int or long):
@@ -511,8 +513,8 @@ class Block(Element):
                 val_str += self.get_cpp_value(element) + ', '
 
             if len(val_str) > 1:
-              # truncate to trim superfluous ', ' from the end
-              val_str = val_str[0:-2]
+                # truncate to trim superfluous ', ' from the end
+                val_str = val_str[0:-2]
 
             return val_str + '}'
 
@@ -523,8 +525,8 @@ class Block(Element):
                 val_str += '{' + self.get_cpp_value(key) + ', ' + self.get_cpp_value(pyval[key]) + '}, '
 
             if len(val_str) > 1:
-              # truncate to trim superfluous ', ' from the end
-              val_str = val_str[0:-2]
+                # truncate to trim superfluous ', ' from the end
+                val_str = val_str[0:-2]
 
             return val_str + '}'
 
@@ -538,6 +540,26 @@ class Block(Element):
 
     def is_virtual_source(self):
         return self.key == 'virtual_source'
+
+    def is_deprecated(self):
+        """
+        Check whether the block is deprecated.
+
+        For now, we just check the category name for presence of "deprecated".
+
+        As it might be desirable in the future to have such "tags" be stored
+        explicitly, we're taking the detour of introducing a property.
+        """
+        if not self.category:
+            return False
+
+        try:
+            return any("deprecated".casefold() in cat.casefold()
+                       for cat in self.category)
+        except Exception as exception:
+            print(exception.message)
+        return False
+
 
     # Block bypassing
     def get_bypassed(self):
--- a/grc/gui/canvas/block.py
+++ b/grc/gui/canvas/block.py
@@ -1,5 +1,6 @@
 """
 Copyright 2007, 2008, 2009 Free Software Foundation, Inc.
+Copyright 2020-2021 GNU Radio Contributors
 This file is part of GNU Radio
 
 SPDX-License-Identifier: GPL-2.0-or-later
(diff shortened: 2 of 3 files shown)


marcusmueller:
Since the users in most dire need of this feature would probably be the ones using released GNU Radio versions, it seems desirable to backport to both 3.8 and 3.9.

grc/gui/canvas/block.py
```diff
@@ -95,17 +96,39 @@ def rotation(self, rot):
...
-            colors.BLOCK_DISABLED_COLOR
-        )
+        def get_bg():
+            """
+            Get the background color for this block
+
+            Explicit is better than a chain of if/else expressions,
+            so this was extracted into a nested function.
+            """
+            c = colors
```
skoslowski, line 106: I wonder if it is really worth saving 5 characters?

grc/core/blocks/block.py
```diff
@@ -539,6 +540,30 @@ def is_virtual_sink(self):
...
+        For now, we just check the category name for presence of "deprecated".
+
+        As it might be desirable in the future to have such "tags" be stored
+        explicitly, we're taking the detour of introducing a property.
+        """
+        if not self.category:
+            return False
+
+        try:
+            if type(self.category) is list:
```
skoslowski, line 557: Using `isinstance()` is usually preferred. However, when is self.category not a list?https://github.com/gnuradio/gnuradio/blob/39a173fe78a127e57019600e52173e02e492827d/grc/core/blocks/_build.py#L29

skoslowski (changes requested):
Great idea, Marcus!

grc/core/blocks/block.py
```diff
@@ -539,6 +540,30 @@ def is_virtual_sink(self):
     def is_virtual_source(self):
         return self.key == 'virtual_source'
 
+    @property
+    def deprecated(self):
```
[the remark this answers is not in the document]
marcusmueller: Yeah, I was torn for that exact reason. We can convert that into an actual getter later on, if there's multiple sources of truth for that e.g. if we should come around to implementing something like block servers. Will change this to be plain.

grc/core/blocks/block.py
marcusmueller: The core.Block ctor sets it to empty string; that might be the only time it's not a list. I guess the right thing to do is change the ctor to initialize it as empty list and don't even check for type here?

grc/core/blocks/block.py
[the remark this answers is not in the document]
marcusmueller: do we know all categories a block is in  in _build.py? Can parsing another block_tree lead to reorganization later on?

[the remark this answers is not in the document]
skoslowski: good point, categories may be overwritten after the block class was created. Once, directly after, see https://github.com/gnuradio/gnuradio/blob/0b3dc1b157b709fcd64a80f13ec604bda3f86e66/grc/core/platform.py#L169

one could make sure block_tree will get loaded first and pass the final category directly to the class builder. I doubt that worth the effort. `__init__` then.

It occurs to me, that you might want to support deprecation as a block flag. pull that into a class attribute `is_deprecated`, an have the library loader set that when it overwrites the category.

skoslowski:
Another thought here: Maybe deprecated block could be styled differently in the block library panel. Darker/ligher text color (theme..) may be a nice alternative to moving the into separate categories?

grc/core/blocks/block.py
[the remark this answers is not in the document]
ei8fdb: > It occurs to me, that you might want to support deprecation as a block flag. pull that into a class attribute is_deprecated, an have the library loader set that when it overwrites the category.

Creating a deprecated block flag would be helpful to UI design as it could be leveraged in design. Instead of deprecation being communicated via block colour/block stroke/something arbitary, it could be communicated by that flag.

image
In this instance the block is deprecated, but also by-passed.

This deprecated flag would be set, until the the block was no longer deprecated, right?

ei8fdb:
> Another thought here: Maybe deprecated block could be styled differently in the block library panel. Darker/ligher text color (theme..) may be a nice alternative to moving the into separate categories?

Using colour to try and communicate this is really hard - What colour do we choose? Will *all* users know the colour means deprecated? What about those who don't know what the colour means? How would they find out? What if you have a visual colour deficiency and you can't see the colour we choose?

Your mention of a deprecated flag is a better place to start.

grc/core/blocks/block.py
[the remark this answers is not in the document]
skoslowski: ...not talking about UI here. This discussion is about how to efficiently handle this in the backend ("grc/core"). 
How this information is used in the GUI part of GRC is another matter....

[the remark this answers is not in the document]
ei8fdb: Right, but that flag would be available to be used in the GUI, right?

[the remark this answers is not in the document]
skoslowski: Of course. 

Note, with the term "flag" I mean attribute of block descriptions (YAML) as will as a corresponding class attribute in the backend block classes. User visible flags (on blocks) would be a different concept.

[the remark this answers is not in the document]
marcusmueller: > It occurs to me, that you might want to support deprecation as a block flag. pull that into a class attribute `is_deprecated`, an have the library loader set that when it overwrites the category.

Exactly (I think I've even hinted at that :) )! Generally, I'd rather have a consistent convention on flags (like deprecated, rate_limiting, …) than have to infer it from category names.

> `__init__` then.

Ah, that's great. You saying that implies at ctor time the categories are fixed!

grc/gui/canvas/block.py
marcusmueller: fixed

marcusmueller:
> Another thought here: Maybe deprecated block could be styled differently in the block library panel. Darker/ligher text color (theme..) may be a nice alternative to moving the into separate categories?

Sounds like a good idea to do that. Personally, "block category" and "our knowledge of the problems with the block" are different ... categories.

The question is how to mark that. Color is one way, but maybe a triangular warning sign would be better, something like

```
[Packet Tools]
    Packet Encoder ⚠Deprecated
```

in the block list.

marcusmueller:
following discussion on matrix / GRC Dev: Not targetting this at maint-3.8 maint-3.9 for now, until we've made a more final decision at how to implement this functionality.

haakov:
I like the yellow triangle. I also agree that the YAML flags would be a reasonable place to store the deprecation status

ei8fdb:
> I like the yellow triangle. I also agree that the YAML flags would be a reasonable place to store the deprecation status

Before a decision is made on which icon is used, I want to do some research with users.

marcusmueller:
> > I like the yellow triangle. I also agree that the YAML flags would be a reasonable place to store the deprecation status
> 
> Before a decision is made on which icon is used, I want to do some research with users. Give me some time to do that.

Sure, but that was meant to be a separate thing: how to mark an entry in the list of available blocks as deprecated.

That covers different aspects of communication of depreciation: the user will only interact with that list on creation of a new flow graph, not when importing an existing one. So, we need to do both.

As discussed, none of this ends up in a release anytime soon, so the question is whether you'd want me to still like to integrate the current approach into GRC on the master branch, so that you can start experimenting within GRC with a way of highlighting deprecated blocks.

grc/core/blocks/block.py
```diff
@@ -539,6 +541,26 @@ def is_virtual_sink(self):
     def is_virtual_source(self):
         return self.key == 'virtual_source'
 
+    def is_deprecated(self):
```
skoslowski, line 544: Given the limited use of `self`, this could/should be a module-level function. I would also mark this protected (leading _) - no code outside of this module should use this.

grc/core/blocks/block.py
```diff
@@ -539,6 +541,26 @@ def is_virtual_sink(self):
...
+        explicitly, we're taking the detour of introducing a property.
+        """
+        if not self.category:
+            return False
+
+        try:
+            return any("deprecated".casefold() in cat.casefold()
+                       for cat in self.category)
+        except Exception as exception:
+            print(exception.message)
```
skoslowski, line 560: This feels wrong, but, yes, there is no logger defined for this module. If you wanted to go the extra mile you could add one and report this through it. There is logging configuration...

skoslowski (changes requested):
Hi Marcus, this code looks much better and cleaner. Thanks. 
I have two more minor things, that I believe would further improve things

willcode:
How will the type change in `category` (from str to list) be made backward compatible? What about another field that can contain additional information:

```
deprecated: 3.10
```

skoslowski:
@willcode, there is actually no type change at all: In the block descriptions (YAML) `category` has to be a `str`. Internally this was and still is parsed into a list. All that's being changed here is to fix the type of a placeholder value in the base class of blocks. This value is always shadowed by the actual block classes (that are created from block descriptors data). Sadly `typing` wasn't around when this was written (Python 2.6) which makes this code a little hard to reason with.

I like the idea of additional information. Such a field would have to be added to the schema check, older versions of GRC would ignore it. The question is how do you propose to make use that field?

willcode:
Oh, good. I only read the diff, glad it's already a list.

The depricate version could be added to the documentation. I don't think it's worth taking any action based on the version.

It's a backward compatible change to backport addition of a valid field in the schema. What happens if a newer yaml makes it back (e.g., in an OOT) to a version of GRC that doesn't have that in the schema? That could be an argument for keeping it in the category list and making it like `category: [Core, Deprecated-3.10]`.

skoslowski:
I designed the YAML to be extended at some point. the schema checker will to only emit a warning on extra fields.

grc/core/blocks/block.py
marcusmueller: yep!

grc/core/blocks/block.py
marcusmueller: I'll opt to not do this at this time, though, and do that separately, and then consistently for the whole module. Let me add a `#fixme`, however.

grc/core/blocks/block.py
marcusmueller: problem: if we add that flag, we're suddenly back in a situation where it makes sense that this would be a method.

skoslowski: This function does one thing - parse the categories. You can always write `or ... in self.flags` at the call-side

marcusmueller: If whenever I need the function I need to write `or some other criterion that's always the same`, then that should be part of the function, no? Principle of least surprise!

mormj (approved):

mormj:
Let's go ahead and merge this so it gets visibility on master - any UX issues can be worked out prior to backporting.

Thanks @marcusmueller

PR #4551, opened 2021-04-25
grc: fix dark theme on ParamWidgets

### LINKED ISSUE #2374: GRC: Dark theme not working well, input box text is white

chuangzhu:
Fix `gtk-application-prefer-dark-theme` not affecting the color displayed.

For example, when you have a `~/.config/gtk-3.0/settings.ini` like this, the background color of the input box will still be white.

```
gtk-application-prefer-dark-theme=true
gtk-theme-name=Breeze
```

This also fixes #2374.

### DIFF
--- a/grc/gui/ParamWidgets.py
+++ b/grc/gui/ParamWidgets.py
@@ -29,9 +29,9 @@ def have_dark_theme():
     config.read(os.path.expanduser(Constants.GTK_SETTINGS_INI_PATH))
     prefer_dark = config.get(
         'Settings', Constants.GTK_INI_PREFER_DARK_KEY, fallback=None)
-    if prefer_dark in ('1', 'yes', 'true', 'on'):
-        theme_name = config.get(
-            'Settings', Constants.GTK_INI_THEME_NAME_KEY, fallback=None)
+    theme_name = config.get(
+        'Settings', Constants.GTK_INI_THEME_NAME_KEY, fallback=None)
+    if prefer_dark in ('1', 'yes', 'true', 'on') or is_dark_theme(theme_name):
         return True
     try:
         theme = subprocess.check_output(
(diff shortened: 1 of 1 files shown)


mormj:
@chuangzhu - thank you for the contribution.  GNU Radio uses the DCO as described here: https://github.com/gnuradio/gnuradio/blob/master/CONTRIBUTING.md which basically requires that you sign your commit using `git commit -s`

This can be done easily with `git commit -s --amend` on this branch and then force pushing.

Thanks!

mormj:
The logic makes sense here in that if you have the gtk setting `gtk-application-prefer-dark-theme=true` then it should default to a dark them, so that
- If a user has selected a dark theme, this method will return True
- If a user has config option "prefer dark theme" then this will always return true

I don't have a good setup here with dark themes and haven't reproduced the original issue - going to get an additional set of eyes on the logic here.

grc/gui/ParamWidgets.py
```diff
@@ -32,7 +32,7 @@ def is_dark_theme(theme_name):
     if prefer_dark in ('1', 'yes', 'true', 'on'):
         theme_name = config.get(
```
mormj, line 33: If this if statement always returns true, then finding the theme name should not be necessary - this doesn't seem right as this theme name is not used anywhere else in the code.

mormj (approved):
This logic makes sense now

willcode (approved):
This works.

If we want to, we can replace the subprocess part, then move that check earlier to make better logic, using the programmatic interface:
```
settings = gtk.Settings.get_default()
theme = settings.get_property('gtk-theme-name')
```
There ought to be a way to check the "prefer" value to, but I didn't find it right away.

PR #4553, opened 2021-04-25
gr-iio: Set correctly dependent libs of gnuradio-iio

dl1ksv:
The gnuradio-iio lib depends on libiio and conditionally on libad9361.
Correct setting is required if you reuse the gr-iio component for instance in gr-osmosdr.


ryanvolz (approved):

dl1ksv:
The set_params function of pluto_source/sink has no parameter buffer_size. So it should be removed from the python callback

dl1ksv:
The first commit meanwhile  is part of the feture/gr-iio branch and the second issue is sovled by #4585.

Issue #3616, opened 2020-06-30
gr_modtool newmod gives vague error messages 


ewhite42:
When I attempted to install an out of tree module using the command: **gr_modtool newmod gr-newname** (not the actual name, just an example), the exception "ModToolException: Invalid module name" was thrown. It was not clear to me until asking a member of the GR community that this error was a result of the fact that I had included "gr-" at the beginning of the name. Just thought it might be useful for the program to print a more informative error, specifying why the module name is invalid. Thanks so much for considering this, take care!

fgerlits:
fixed in https://github.com/gnuradio/gnuradio/pull/4555

PR #4555, opened 2021-04-29
Improve the 'invalid name' error message in modtool

fgerlits:
... and fix some validations which checked a prefix instead of the whole name.  https://github.com/gnuradio/gnuradio/issues/3616

### DIFF
--- a/gr-utils/modtool/tests/test_modtool.py
+++ b/gr-utils/modtool/tests/test_modtool.py
@@ -239,12 +239,20 @@ class TestModToolCore(unittest.TestCase):
         test_dict['directory'] = module_dir
         # Missing 2 arguments blockname, new_name
         self.assertRaises(ModToolException, ModToolRename(**test_dict).run)
+
         test_dict['blockname'] = 'square_ff'
         # Missing argument new_name
         self.assertRaises(ModToolException, ModToolRename(**test_dict).run)
+
         test_dict['new_name'] = '//#'
         # Invalid new block name!
+        expected_message = "Invalid new block name '//#': names can only contain letters, numbers and underscores"
+        self.assertRaisesRegex(ModToolException, expected_message, ModToolRename(**test_dict).run)
+
+        test_dict['new_name'] = 'non-alpha-chars-like-hyphen-are-not-allowed'
+        # Invalid new block name!
         self.assertRaises(ModToolException, ModToolRename(**test_dict).run)
+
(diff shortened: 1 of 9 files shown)


mormj:
@fgerlits - thank you for the contribution!  Love the consolidation of common code to one place and cleaning up the regex match.  I will give this a test shortly.

Would you mind signing your commit since we adhere to the DCO for licensing?  Just a `git commit -s --amend` and then a force push should do the trick

mormj:
Getting an error with the import of validate name:
```
josh@josh-m15:/share/gnuradio/gr310/src/gnuradio/build$ gr_modtool newmod tmp
ERROR:root:cannot import name 'validate_name' from 'gnuradio.modtool.cli.base' (/share/gnuradio/gr310/lib/python3/dist-packages/gnuradio/modtool/cli/base.py)
```

mormj:
Looks like the relative import for validate_name needs to go somewhere else so that it grabs it from core/base.py instead of trying to grab it from cli/base.py

fgerlits:
> Looks like the relative import for validate_name needs to go somewhere else so that it grabs it from core/base.py instead of trying to grab it from cli/base.py

Yes, you're right.  I've fixed it in https://github.com/gnuradio/gnuradio/commit/ece93c6397ae85d2408b00713eb84f9182be43e6.

mormj (approved):
This looks good to me!  Would like to get @swap-nil7 's eyes on this as well.

gr-utils/modtool/tests/test_modtool.py
```diff
@@ -239,12 +239,19 @@ def test_rename(self):
...
         # Missing 2 arguments blockname, new_name
         self.assertRaises(ModToolException, ModToolRename(**test_dict).run)
+
         test_dict['blockname'] = 'square_ff'
         # Missing argument new_name
         self.assertRaises(ModToolException, ModToolRename(**test_dict).run)
+
         test_dict['new_name'] = '//#'
         # Invalid new block name!
+        self.assertRaisesRegex(ModToolException, "Invalid new block name '//#': names can only contain letters, numbers and underscores", ModToolRename(**test_dict).run)
```
swap-nil7, line 249: Please reduce the length of this line. :)

swap-nil7 (approved):
Looks good. Thanks for your contribution.

mormj (approved):

PR #4563, opened 2021-05-02
GRC: Supersede Constants, color.py with Theming engine

marcusmueller:
This PR adds a ThemeEngine, which applies dict-specified styles to GRC elements.

For now, I only reworked the Block canvas Element, as is probably the most sensible demonstrator for this.

The idea is simple: instead of distributing styling information (fonts, sizes, colors, paddings) throughout the source code, configuration files, and things that are called Constants but get modified at runtime, we just put everything in one  dictionary. We allow the user to override parts of that through their own JSON.

The dictionary structure is hierarchical and simple:

```
---+Block
   |
   ---+Defaults
   |  |
   |  --- property: value
   |  --- property2: value
   |
   ---+Selected state
   |  |
   |  --- property2: different value
   |
   ---+Disabled state
   |  |
   |  --- property2: another different value
   |
   ---+Disabled selected state
      |
      --- property2: yet another different value
---+Port
   |
   ---+Defaults
   |  |
   |  --- property: value
   |  --- property2: value
   |
   ---+Selected state
   |  |
   |  --- property2: different value
   |
   ---+Disabled state
      |
      --- property2: another different value
  ………
```

The trick is that on lookup, the most specific property is selected. This is a bit like CSS, but less convoluted, and doesn't depend on a third-party parser. (the Gtk-integrated parser can't be used, as it doesn't support arbitray object and property types)

In order to not completely break performance, lookup is done through a cache, which gets invalidated on updating the ThemeEngine; that feature allows for live experimentation with themes.

The default theme (which strives to be identical to its hard-coded predecessor) is found in grc/gui/defaulttheme.py.

This PR does *not* already convert all of the canvas to the theming engine, on purpose, to keep the changeset reviewable.
However, on positive review, and after squashing this, porting Flowgraph and the ports should be quick.

Next steps would be:

* Conversion of all Elements to use ThemeEngine instead of hard-coded visual properties
* Adding a minimal RPC / HTTP server (optionally enablable through prefs) so that HTML + JS can live-fiddle with the theme
* Working with the UX experts on what properties to add (maybe something like an "icon" property for blocks? Theming things that aren't themed yet, e.g. the block list?)

### DIFF
--- a/grc/gui/Config.py
+++ b/grc/gui/Config.py
@@ -109,6 +109,15 @@ class Config(CoreConfig):
         self._gr_prefs.set_string("qtgui", "qss", value)
         self._gr_prefs.save()
 
+    @property
+    def theme_file(self):
+        return self._gr_prefs.get_string('grc', 'theme_file', '')
+
+    @theme_file.setter
+    def theme_file(self, value):
+        return self._gr_prefs.set_string('grc', 'theme_file', value)
+        self._gr_prefs.save()
+
     ##### Originally from Preferences.py #####
     def main_window_size(self, size=None):
         if size is None:
--- a/grc/gui/Constants.py
+++ b/grc/gui/Constants.py
@@ -6,7 +6,9 @@ SPDX-License-Identifier: GPL-2.0-or-later
 
 """
 
-
+import gi
+gi.require_version('Gdk', '3.0')
+gi.require_version('Gtk', '3.0')
 from gi.repository import Gtk, Gdk
 
 from ..core.Constants import *
--- a/grc/gui/DrawingArea.py
+++ b/grc/gui/DrawingArea.py
@@ -9,9 +9,9 @@ SPDX-License-Identifier: GPL-2.0-or-later
 
 from gi.repository import Gtk, Gdk
 
-from .canvas.colors import FLOWGRAPH_BACKGROUND_COLOR
 from . import Constants
 from . import Actions
+from .Theme import Theme
 
 
 class DrawingArea(Gtk.DrawingArea):
@@ -188,7 +188,8 @@ class DrawingArea(Gtk.DrawingArea):
         width = widget.get_allocated_width()
         height = widget.get_allocated_height()
 
-        cr.set_source_rgba(*FLOWGRAPH_BACKGROUND_COLOR)
+        cr.set_source_rgba(
+            *Theme.get_property("background-color", "Flowgraph"))
         cr.rectangle(0, 0, width, height)
         cr.fill()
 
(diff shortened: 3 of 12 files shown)


marcusmueller:
A bit on future-proofness: I intentionally did come up with a slightly abstract dict model of canvas elements, so that I can add the same support to future developments of GRC. It should also be rather straightforward to export the internal dict as CSS, should we find ourselves in the situation that we directly want to export graphical flow graphs as HTML/SVG and style them accordingly.

marcusmueller:
Ah nice, the CI failures are actually something I missed by running pytest from the grc/tests directory manually.

skoslowski:
I like the idea. Its gonna take a few days before I have take an in depth look. Initial thought: have two build-in theme - dark/light and lets not use for human-writable files.

marcusmueller:
> I like the idea. Its gonna take a few days before I have take an in depth look. 

That perfectly aligns with my plans! Got to fix that unit test, which requires a bit of rework to get red of the circular inclusion without in the end converting everything at once (and thus making this hard to review).

> Initial thoughts: have two build-in themes - dark/light

Good idea. First step was to recreate the current (light) one, but a dark one would be a good demonstrator for the ability to load themes at will / through configuration.

> lets not use JSON for human-writable files

Well I don't honestly care about the format (I like writing JSON still way better than the too-many-ways-to-do-the-same YAML 1.2 superset, but I don't care nearly enough; if we specify it to be YAML, I can still write JSON and call it YAML :D ). The idea was that JSON was

* common,
* needs no external library, and 
* it would be easy to use JS to generate it from settings on an HTML page.

If you have a replacement that fits the first two criteria that you'd prefer, let's stick with that!
Anything that can be represented as a nested python dict works. There's one big argument for YAML: way it's implemented, you can override the style for a single element type as long as that type has a name. We could allow developers to style their GRC blocks, should we want to do that (might make sense especially if we add things like a unicode or SVG icon later on). GRC block defs are already yaml, we can add a `style` AST element and parse that as overriding that particular block style.

At first I thought plain python would be good enough, but meh, then people start integrating logic and we're back to the point where logic and looks get mixed, also we'd have to be extra careful about people starting to exchange code when they mean to exchange themes.

mormj:
Still seeing `grc_tests` failing on Fedora and CentOS - was seeing similar failure in #4208 - just reopened that one to re-run CI to remember what the failure was there.

mormj:
Going to close this PR for now since I don't see this big of a change going into the upcoming release, but adding it to the TODO for 3.11

Issue #4562, opened 2021-05-02
QT GUI Digital Number Control / inaccuracy


cbxtom:
Hi team,
when I use "QT GUI Digital Number Control" to display a frequency I see some "special effects".
I feed the control by sending numbers from Gpredict to it.

Below 128MHz, it works fine but above I get wrong display data:

ok:
Input 128.001.000Hz -> Display = 128.001.000

not ok:
Input 138.001.000Hz -> Display = 138.000.992
Input 137.999.000Hz -> Display = 137.999.008

I simulate that input by using Gpredict via tcp port 4532 but found debug messages ok.

Does that depend on the QT GUI Digital Number Control and the way it converts numerical data types, since (128 = 2⁷)?

![grafik](https://user-images.githubusercontent.com/83543567/116816243-6505fa00-ab61-11eb-8d87-9d712ae15e08.png)

willcode:
This is due to use of `pmt.from_float()` or similar. Verified that, with two chained controls, the second one displays wrong values. Could be just in the control, or could also be in whatever is sending messages to the control

willcode:
@cbxtom I've put in a fix for some floating point precision problems in the Python GUI widgets. I'm still not quite sure this fixes your problem, since I do not know how you have gpredict wired up to send data to the number control. Can you describe that in any more detail?

cbxtom:
Here I explain what I did and I add a screendump of my flowchart:
I use a program called gpredict to calculate a frequency drift for satellites flying by to compensate the doppler shift. You can setup the frequency also manually and there I found that misinterpretation of the values coming from gpredict.

![grafik](https://user-images.githubusercontent.com/83543567/117349752-0b2c6980-aeac-11eb-883c-54671b513f7d.png)

cbxtom:
In the left bottom edge one can see that the incoming values seem to be correct:
![grafik](https://user-images.githubusercontent.com/83543567/117350407-cfde6a80-aeac-11eb-85b2-417afd9c8ee1.png)

cbxtom:
Does that help?
If a solution is available I offer to test it immediatly.

willcode:
That helps. So, while you pointed out something that needed to be looked at in GNU Radio, the problem you are seeing is actually this code in `gr-gpredict-doppler/python/doppler.py`:

```
  def sendFreq(self,freq):
    p = pmt.from_float(freq)
[...]

PR #4567, opened 2021-05-04
qtgui: Use pmt.from_double() instead of pmt.from_float()

willcode:
Messages should be constructed using double values instead of float
to avoid loss of precision.

Related to (but does not fix) https://github.com/gnuradio/gnuradio/issues/4562


willcode:
The togglebutton and toggleswitch are broken for other reasons, comparing literals instead of strings, e.g., `real == bool`).

marcusmueller (approved):

PR #4568, opened 2021-05-04
soapy: added missing MSVC resource file referenced by CMake

ncorgan:


willcode:
Thanks @ncorgan, will merge after the CI check.

willcode:
@ncorgan need a signoff for DCO.

PR #4569, opened 2021-05-04
blocks: use VOLK popcnt implementations for count_bits

ncorgan:


willcode:
@ncorgan need a signoff for DCO.

gr-blocks/lib/count_bits.cc
```diff
@@ -45,16 +41,18 @@ unsigned int count_bits16(unsigned int x)
...
+    return res;
 }
 
 unsigned int count_bits64(unsigned long long x)
 {
-    return count_bits32((x >> 32) & 0xffffffff) + count_bits32(x & 0xffffffff);
+    uint64_t res_as_u64 = 0;
+    volk_64u_popcnt(&res_as_u64, static_cast<uint64_t>(x));
+
+    return static_cast<unsigned long long>(res_as_u64);
```
marcusmueller, line 55: hm, but you're return an `unsigned int`, not an `unsigned long long`; and frankly, that should suffice for the near future :)

marcusmueller (changes requested):

gr-blocks/lib/count_bits.cc
mbr0wn: While we're touching this, can you please change this to uint64_t?

mbr0wn: Sorry I meant the function argument, not the return value.

ncorgan: What, where did that odd mistake come from...I'll make these changes when I get home from work.

ncorgan: I'd argue the return type is also wrong. Should `count_bits64` be returning a 32-bit number?

ncorgan: Derp, ignore my uint64_t comment, I was thinking byteswap for some reason.

mbr0wn (approved):

PR #4570, opened 2021-05-04
pkg: add soapy support for in-tree packaging

mormj:


.packaging/fedora/gnuradio.spec
```diff
@@ -76,6 +76,7 @@ BuildRequires:	qwt-qt5-devel
 BuildRequires:	SDL-devel
 BuildRequires:	thrift
 BuildRequires:	uhd-devel
+BuildRequires:	SoapySDR-devel-devel
```
willcode, line 79: Should this just be SoapySDR-devel?

willcode (approved):
Looks right. The debian stuff is used by Ubuntu also? Debian 10 has soapysdr0.6, which is too old.

mormj:
> Looks right. The debian stuff is used by Ubuntu also? Debian 10 has soapysdr0.6, which is too old.

Yeah Ubuntu will use the debian packages that are either in main Ubuntu repos or our ppa - so i just backported soapy 0.7 into our ppa.  Folks on Debian 10 will have to get the updated package somehow.

Issue #4571, opened 2021-05-04
soapy: API completeness


ncorgan:
With the intent of matching the granularity of gr-uhd:

- [x] [Some functions](https://github.com/gnuradio/gnuradio/blob/95cc649e12fb25820c0cdd5dceb1977387122fa3/gr-soapy/lib/block_impl.h#L130) protected in block_impl instead of public for block, should be exposed 
- [x] Remaining 0.7.2 [functions](https://github.com/pothosware/SoapySDR/blob/cc4ff057d672099c48db03ed64684add3173b393/include/SoapySDR/Device.hpp), I assume unneeded for original users
- [x] 0.8.0 functions, using [Soapy's API defines](https://github.com/pothosware/SoapySDR/blob/master/include/SoapySDR/Version.h#L41), as [gr-uhd does](https://github.com/gnuradio/gnuradio/blob/master/gr-uhd/lib/usrp_sink_impl.cc#L179)
- [ ] Messages for new functionality

willcode:
Yes, I was thinking of adding more functions, but did not expose anything that was not exposed in the original code, for the first round. Many users don't need anything more than the current API, so this gets the feature out and tested a bit. The getters should be an easy one.

One thing I want to put some thought into is how interactive things should be. There could even be a fully interactive QT Soapy SDR block, like the Soapy blocks in Pothos, but not sure yet.

ncorgan:
I've been knee-deep in Soapy recently, so go ahead and assign this to me. I can get PRs up for some of this over the next few days.

willcode:
Cool - we can coordinate, no need for a specific assignment. Thanks!

We also need to dup the Custom Soapy Source/Sink logic out of the original yml files to support any 'ole device, command line specification of devices, etc.

willcode:
@ncorgan anything left for this issue?

ncorgan:
The checkboxes in the issue description are still valid. The most recent branch added everything needed for the 0.7.2 features currently exposed by the block, but I'd like to add the remaining 0.7.2 functionality appropriate for a block. I see this as two more branches:
 * The remaining functions needing the new custom pybind11, as there's a minor refactor there.
 * The rest of it, which is just thin wrappers around the `SoapySDR::Device` functions.

willcode:
@ncorgan what do you have in mind for additional messages?

ncorgan:
gr-uhd has messages for most setters, so I have a WIP branch with similar.

[...]

PR #4573, opened 2021-05-04
soapy: made majority of protected functions public

ncorgan:
* Partially addresses https://github.com/gnuradio/gnuradio/issues/4571


ncorgan:
For some reason, I'm running into issues running bindtool on my machines, so as of this comment, the PR only has the C++ change.

willcode:
`gr-utils/bindtool/scripts/binding-hash-hotfixer.zsh` will help with the hashes

Shoud the gain setters be changed to double to match the getter and everything else?

PR #4592, opened 2021-05-05
pkg: add soapy support for in-tree packaging (backport to maint-3.9)

willcode:
Backport https://github.com/gnuradio/gnuradio/pull/4570


willcode:
Merge after backporting gr-soapy.

PR #4596, opened 2021-05-05
soapy: added Python support for Soapy types needed for full API support

ncorgan:
* Fixed gr-soapy Python import
* Partially addresses https://github.com/gnuradio/gnuradio/issues/4571

### DIFF
--- a/gnuradio-runtime/python/gnuradio/__init__.py
+++ b/gnuradio-runtime/python/gnuradio/__init__.py
@@ -59,3 +59,4 @@ if path.endswith(path_ending):
     __path__.append(os.path.join(build_path, 'gr-pdu', 'python'))
     __path__.append(os.path.join(build_path, 'gr-network', 'python'))
     __path__.append(os.path.join(build_path, 'gr-zeromq', 'python'))
+    __path__.append(os.path.join(build_path, 'gr-soapy', 'python'))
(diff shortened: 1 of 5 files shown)


willcode:
For arginfo bool, int, float ... can the correct type be returned instead of having the caller do a value_to_x? The type is known in the arginfo. Is the type fixed (per key) or should it be set based on the value setter?

ncorgan:
Sounds good, I'll make that update.

willcode:
Ubuntu 20.04 ship with 0.7.2, so that's what we have on CI. Centos and Debian ship with older versions, so we're using 0.8.0 from source. Did you include some 0.8.0 functionality?

ncorgan:
Looks like some of Soapy's string conversion helper functions were post-0.7.2. Simple enough to work around.

willcode:
* Where possible, we're trying to de-boostify the code. Is there an easy alternative to boost::lexical_cast()?
* Can you check that float is only used for Python? Should string_to_setting<float> should be a double? Others?

ncorgan:
Ah, good catch on the float, that's a typo. Should be double.

And in theory, a stringstream could work. I'll see how the unit tests hold up.

willcode:
Build, test, and a local run (did not try out args/ranges though) worked. Let me know when you're happy with it.

ncorgan:
What's up there works for me. The unit tests work for me to confirm the fields work as expected.

willcode:
And the maint-3.9 test worked, so all good.

PR #4599, opened 2021-05-06
soapy: completed API related to block's current functionality

ncorgan:
* Consistency in validate_channel() calls

### DIFF
--- a/gr-soapy/include/gnuradio/soapy/CMakeLists.txt
+++ b/gr-soapy/include/gnuradio/soapy/CMakeLists.txt
@@ -13,6 +13,7 @@
 install(FILES
     api.h
     block.h
+    soapy_types.h
     source.h
     sink.h
     DESTINATION ${GR_INCLUDE_DIR}/gnuradio/soapy
--- a/gr-soapy/include/gnuradio/soapy/block.h
+++ b/gr-soapy/include/gnuradio/soapy/block.h
@@ -11,6 +11,7 @@
 
 #include <gnuradio/block.h>
 #include <gnuradio/soapy/api.h>
+#include <gnuradio/soapy/soapy_types.h>
 #include <cstdint>
 #include <string>
 #include <vector>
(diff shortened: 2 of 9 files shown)


willcode:
Looks good. I'll take a closer look tomorrow, try it out, and merge.

Do you have any idea about testing the whole interface? Does the dummy driver get built in, and can it be used?

ncorgan:
The only testing to really be done would be making sure the right Soapy function is called for each function. Anything beyond that would turn into testing the underlying driver itself.

willcode:
The null driver would allow for a couple of tests. It has no channels, so it would throw on most calls.

willcode:
The SoapySDR types are in the SoapySDR namespace, where we've wrapped everything else into the `gr::soapy` namespace. Something seems a little asymmetric here. Ideas? The original, minimal wrapping was just enough to make things work. Now that it's full-featured, we could potentially have provided more direct access to the SoapySDR functions.

I think I still like the full wrapping, to make everything have a GNU Radio feel. Should we just `using` the types into `gr::soapy`?

ncorgan:
The enum would become `gr::soapy::arginfo_t::Type`. Not ideal. How about `argtype_t`? I've already been using `argtype` as a parameter name.

willcode:
I like that. It seems like a pain to have to rename/rewrap everything, but it keeps the API more consistent across gnuradio.

Minor thing: is there a way to make `SOAPY_SDR_ARG_INFO_BOOL` into just `ARGINFO_BOOL`, or really `gr::soapy::ARGINFO_BOOL`?

ncorgan:
With this, we have `gr::soapy::BOOL`. If we want `gr::soapy::ARGINFO_BOOL`, we'd need a new enum, which means casting for each access.

willcode:
Got it, thanks. I was looking in `Types.h` instead of `Types.hpp`.

I think this is all good.

PR #4603, opened 2021-05-09
grc: Make exception for epy in blacklist id

solomonbstoner:
I refer to the mailing list thread on May 8 2021 regarding "Embedded
Python Block Tutorial Param ID Error".

The bug is caused by #4522 as names of embedded python blocks
such as `epy_block_0` are blacklisted because of "blocks" in the
argument `black_listed_ids`. This pull request resolves it.

### DIFF
--- a/grc/core/params/dtypes.py
+++ b/grc/core/params/dtypes.py
@@ -43,7 +43,7 @@ def validate_block_id(param,black_listed_ids):
     if not re.match(r'^[a-z|A-Z]\w*$', value):
         raise ValidateError('ID "{}" must begin with a letter and may contain letters, numbers, '
                             'and underscores.'.format(value))
-    if value in (black_listed_ids + ID_BLACKLIST):
+    if value in (black_listed_ids + ID_BLACKLIST) and not re.search('^epy', value):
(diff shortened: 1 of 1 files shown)


skoslowski:
Hi Salomon, thanks for reporting and providing a fix. I actually had a different solution in mind: Use that blacklist only for variable-like blocks and/or changing the codegen for epy block to use a different name for the imported module (there actually is an import alias, we could simply remove).

solomonbstoner:
Hi Sebastian. Its Solomon. :P

I think I could do the codegen one. Right off the bat, I could have `epy_block` be changed to `epy_blk`. That shouldnt trigger the blacklist. If youre agreeable, I would like to make that change.

solomonbstoner:
Created a new pull request #4604 .

Removing the alias removed the bug but I cannot do so as it is required for the rest of the flowchart python file. I also couldnt change the epy block to a different name because `self.name` is used in the flowchart python generation. Under current code, the import alias must be the same as `self.name`, otherwise the flowchart will not run. #4604 is the only way out of all these methods I tried that does not break the other python scripts.

PR #4604, opened 2021-05-09
[v7] grc: Make exception for epy in blacklist id

solomonbstoner:
I refer to the mailing list thread on May 8 2021 regarding "Embedded
Python Block Tutorial Param ID Error".

The bug is caused by #4522 as names of embedded python blocks
such as `epy_block_0` are blacklisted because of "blocks" in the
argument `black_listed_ids`. It affects both embedded python blocks and
modules. This commit fixes the bug by granting exemption for these two
types of blocks from id validation.

### DIFF
--- a/grc/core/blocks/embedded_python.py
+++ b/grc/core/blocks/embedded_python.py
@@ -70,6 +70,7 @@ class EPyBlock(Block):
 
     key = 'epy_block'
     label = 'Python Block'
+    exempt_from_id_validation = True  # Exempt epy block from blacklist id validation
     documentation = {'': DOC}
 
     parameters_data = build_params(
@@ -201,6 +202,7 @@ class EPyBlock(Block):
 class EPyModule(Block):
     key = 'epy_module'
     label = 'Python Module'
+    exempt_from_id_validation = True  # Exempt epy module from blacklist id validation
     documentation = {'': dedent("""
         This block lets you embed a python module in your flowgraph.
 
--- a/grc/core/params/dtypes.py
+++ b/grc/core/params/dtypes.py
@@ -40,10 +40,13 @@ class ValidateError(Exception):
 def validate_block_id(param,black_listed_ids):
     value = param.value
     # Can python use this as a variable?
+
     if not re.match(r'^[a-z|A-Z]\w*$', value):
         raise ValidateError('ID "{}" must begin with a letter and may contain letters, numbers, '
                             'and underscores.'.format(value))
-    if value in (black_listed_ids + ID_BLACKLIST):
+    if value in (black_listed_ids + ID_BLACKLIST) and \
+        not getattr(param.parent_block, 'exempt_from_id_validation', False):
+        # Grant blacklist exemption to epy blocks and modules
         raise ValidateError('ID "{}" is blacklisted.'.format(value))
     block_names = [block.name for block in param.parent_flowgraph.iter_enabled_blocks()]
     # Id should only appear once, or zero times if block is disabled


solomonbstoner:
Removing the alias removed the bug but I cannot do so as it is required for the rest of the flowchart python file. I also couldnt change the epy block to a different name because self.name is used in the flowchart python generation. Under current code, the import alias must be the same as self.name, otherwise the flowchart will not run. #4604 is the only way out of all these methods I tried that does not break the other python scripts.

solomonbstoner:
Wait a minute. I think I found a better way to do it

solomonbstoner:
I see! I missed out on the `mod=self.module_name` part thats why! Thanks Sebastian! Amended as suggested

solomonbstoner:
With commit 9fc22a3, those using epy modules with python snippets will have to amend their scripts.

In the documentation, it is currently stated that:
```
This block lets you embed a python module in your flowgraph.

        Code you put in this module is accessible in other blocks using the ID of this
        block. Example:

        If you put

            a = 2

            def double(arg):
                return 2 * arg

        in a Python Module Block with the ID 'stuff' you can use code like

            stuff.a  # evals to 2
            stuff.double(3)  # evals to 6

        to set parameters of other blocks in your flowgraph.
```

However, with commit 9fc22a3, the code would have to be changed from `<NAME_OF_EPY_MODULE_ID>` to `<NAME_OF_FLOWGRAPH>_<NAME_OF_EPY_MODULE_ID>`. So if my flowgraph is called `example_flowgraph.grc`, the code in the documentation would need to be changed to the following to work.
```
            example_flowgraph_stuff.a  # evals to 2
            example_flowgraph_stuff.double(3)  # evals to 6
```

skoslowski:
Ah I hadn't considered that. @solomonbstoner, your solution is better then. 

Ideally we would have the epy block classes configure their id parameter object to opt-out of checking against imported names. This entails a bit more changes, though. Special-casing epy inside dtype is okay, too.

solomonbstoner:
@skoslowski is this code ok?

grc/core/blocks/embedded_python.py
```diff
@@ -88,6 +88,7 @@ def __init__(self, flow_graph, **kwargs):
         self.module_name = self.name
         self._epy_source_hash = -1
         self._epy_reload_error = None
+        self.exempt_from_id_validation = True  # Exempt epy block from blacklist id validation
```
skoslowski, line 91: This could/should be a class-level variable, maybe above line 75?

grc/core/blocks/embedded_python.py
```diff
@@ -235,6 +236,7 @@ def double(arg):
     def __init__(self, flow_graph, **kwargs):
         super(EPyModule, self).__init__(flow_graph, **kwargs)
         self.module_name = self.name
+        self.exempt_from_id_validation = True  # Exempt epy module from blacklist id validation
```
skoslowski, line 239: same as above

grc/core/params/dtypes.py
```diff
@@ -40,6 +40,11 @@ class ValidateError(Exception):
 def validate_block_id(param,black_listed_ids):
     value = param.value
     # Can python use this as a variable?
+
+    is_exempt_from_id_validation = hasattr(param.parent_block, 'exempt_from_id_validation')
```
skoslowski, line 44:
```suggestion
    is_exempt_from_id_validation = getattr(param.parent_block, 'exempt_from_id_validation', False)
```

grc/core/params/dtypes.py
```diff
@@ -40,6 +40,11 @@ class ValidateError(Exception):
...
+    # Grant exemption to epy blocks and modules
+    if is_exempt_from_id_validation:
+        return value
```
skoslowski, line 47: Hm, not sure about this one. The other checks are still useful.
I think it would be better to use this flag only for disabling the blacklist check

skoslowski (changes requested):
I like this. Good work. I have suggest a few minor tweak, that I hope you like.

grc/core/blocks/embedded_python.py
solomonbstoner: Ok. Will amend

grc/core/blocks/embedded_python.py
solomonbstoner: Ok. Will amend

grc/core/params/dtypes.py
solomonbstoner: Goodness me. How did I not catch that? Thanks! Will make the amendment

grc/core/params/dtypes.py
solomonbstoner: What do you mean? Would you like me to have the `getattr` check _and_ also check the block key like in v3?

solomonbstoner: Am I correct that you mean something like the code below by "the other checks"?
```python
is_exempt_from_id_validation = getattr(param.parent_block, 'exempt_from_id_validation', False)
if is_exempt_from_id_validation and re.match(r'^epy_(block|module)$', param.parent_block.key):
    return value
```
I cant appreciate how having that additional block key check would be beneficial. Wouldn't having this additional check  complicate maintenance? Because the bug will come back should the block key or name change for whatever reason. I feel a single `getattr` check would be better. Makes it easier to add or remove blocks to the exemption too should the need ever arise. What do you think?

solomonbstoner: Ok I gave it even more thought. I think I get what you mean now by "disabling the blacklist check"! Let me make a v7. Im so sorry Im commenting so much! I just cant help but try to think of better ways to make this patch

solomonbstoner:
Now with v7 (commit 3ac4614), the epy blocks and modules wont trigger the blacklist warning, thus solving the bug mentioned. Unlike patch v6 and before, now the blocks/modules will still be subject to the other id requirements like all other blocks/modules which are: 
1. the naming convention
2. the need to be unique
3. the check that they must exist

As an example, this is a screenshot of the epy module that meets requirements 1 to 3, and the blacklist bug is fixed.

Screenshot from 2021-05-16 00-11-57

This is the screenshot of an epy module that violates requirement 2.

Screenshot from 2021-05-16 00-12-03

grc/core/params/dtypes.py
skoslowski: Yes, that is what I meant. I should have made that clearer in my comment... Sorry about that

skoslowski (approved):
Thanks again for reporting and fixing this issue!

marcusmueller:
Let's merge this live on the call!

marcusmueller:
One minor: please squash the commits, and we'll merge right away!

PR #4608, opened 2021-05-11
soapy: added remaining functions that need custom pybind functions

ncorgan:
* Refactored out pybind functions into own utility functions

### DIFF
--- a/gr-soapy/include/gnuradio/soapy/block.h
+++ b/gr-soapy/include/gnuradio/soapy/block.h
@@ -354,6 +354,103 @@ public:
      * \return the name of the clock source
      */
     virtual std::string get_clock_source() const = 0;
+
+    /*!
+     * List the available global readback sensors.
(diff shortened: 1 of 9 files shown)


ncorgan:
With this, the remaining functions should just be straight "public header -> impl -> python".

ncorgan:
If we're keeping these as strings, I'm not sure we'd have parsing at all. The ArgInfo field is itself a string with no getter/setter, and ArgInfo is not an input to anything (maybe that was too much fanciness to begin with...), and if we're not returning bool, etc, the return value would just be the return value from SoapySDR. So if we're removing differences between C++ and Python, we'd either:
 * Make everything a string and remove parsing
 * Move fanciness up to C++ and keep Python the thin wrapper

willcode:
Let's go with the way you're doing it here. Later, we can add C++ `_as_bool()` or similar if needed.

PR #4609, opened 2021-05-11
grc: fix Wrong order in the generated .py caused by uncorrect function without_gui_hint(block)

### LINKED ISSUE #4557: Wrong order in the generated .py causes error "object has no attribute 'tabs_layout_2'"
I am trying to run the example at https://github.com/bastibl/gr-rds/blob/3a80121f0858adf4ebd2f5f9f36ff1e4bb18b7a4/examples/rds_rx.grc (maint-3.9 branch)

I get the following error:

```
Executing: /usr/bin/python3.8 -u /home/kaushik/gr-rds/examples/rds_rx_kauzik.py

  File "/home/kaushik/gr-rds/examples/rds_rx_kauzik.py", line 625, in <module>
    main()
  File "/home/kaushik/gr-rds/examples/rds_rx_kauzik.py", line 603, in main
    tb = top_block_cls()
  File "/home/kaushik/gr-rds/examples/rds_rx_kauzik.py", line 131, in __init__
    self.tabs_layout_2.addWidget(self._qtgui_waterfall_sink_x_0_win)
  File "/usr/local/lib/python3/dist-packages/gnuradio/gr/hier_block2.py", line 82, in __getattr__
AttributeError: 'gnuradio.gr.gr_python.top_block_pb' object has no attribute 'tabs_layout_2'

>>> Done (return code -11)
```
Problem seems to be the placement of the following section of code:
```
        self.tabs = Qt.QTabWidget()
        self.tabs_widget_0 = Qt.QWidget()
        self.tabs_layout_0 = Qt.QBoxLayout(Qt.QBoxLayout.TopToBottom, self.tabs_widget_0)
        self.tabs_grid_layout_0 = Qt.QGridLayout()
        self.tabs_layout_0.addLayout(self.tabs_grid_layout_0)
        self.tabs.addTab(self.tabs_widget_0, 'BB')
        self.tabs_widget_1 = Qt.QWidget()
        self.tabs_layout_1 = Qt.QBoxLayout(Qt.QBoxLayout.TopToBottom, self.tabs_widget_1)
        # And so on...
```
The member variables tabs\_layout\__[0..(n-1)]_ are being defined after the first reference. The user is forced to edit the .py file manually and run from terminal. 

GRC should place the such definitions before first reference.

GnuRadio maint-3.9. Built and installed from commit 44cd1949a027101f805233092e54b36978c55269.

ChristopheSeguinot:

This fix #4557 
I tested it under GR 3.9.1 using different flowgraph combining widgets with/without GUI_hint positioned Inside /outside tabs. 

Note: I =was not able to reproduce errror  in  #4557 first message ( https://github.com/bastibl/gr-rds/blob/3a80121f0858adf4ebd2f5f9f36ff1e4bb18b7a4/examples/rds_rx.grc (maint-3.9 branch) 
Thank to @kaushiksv for proposing this fix

### DIFF
--- a/grc/core/generator/top_block.py
+++ b/grc/core/generator/top_block.py
@@ -196,14 +196,6 @@ class TopBlockGenerator(object):
         ]
 
         blocks = expr_utils.sort_objects(blocks, operator.attrgetter('name'), _get_block_sort_text)
-
-        # Ordering blocks : blocks with GUI Hint must be processed first to avoid PyQT5 superposing blocks
-        def without_gui_hint(block):
-            hint = block.params.get('gui_hint')
-            return hint is None or not hint.get_value()
-
-        blocks.sort(key=without_gui_hint)
-
         blocks_make = []
         for block in blocks:
             make = block.templates.render('make')


kaushiksv:
Actually I was also thinking that this kind-of defeats the purpose of f45c82436c2a19fa1e5ff15177fbc4cc591017cb at least partially. I'm no python/gr expert, however I can dig this after after a couple of days (assuming there is no hurry).

ChristopheSeguinot:
Yes I agree with Jeff that the name of the function is not corresponding to what it does. I will investigate this.

willcode:
Hints are being used for two different things:
- `(0,0)` sets placement in the top level grid
- `tab@0` puts a widget in a tab, not really a hint

The sort order needs to take into account that the tab widget needs to exist before other widgets can be placed in it. Hints containing `@` should be equivalent to `not hint` since they are not put in the top level grid.

willcode:
This should work:
```return hint is None or not ',' in hint.get_value()```
No such param -> no hint
Does not contain a comma (empty, tab spec, ...) -> no hint
Contains a comma -> hint

willcode:
@ChristopheSeguinot I pushed this change onto your branch - see what you think, and squash the commits if you want to keep it.

ChristopheSeguinot:
Jeff I was out for some time. 
Your commit do the right without_hint function, however it does not work at least in the present case 
[test_tabs2.grc.txt](https://github.com/gnuradio/gnuradio/files/6484807/test_tabs2.grc.txt)
In order to really test some fix, we must have several widgets: 
- with and without GUI_hint (the value after the tab@0 containing at least one `,`)
- inside and outside tabs
- declared in different order  : the fact that the no gui_hint widget appear before or after the one with a gui_hint in the .grc file matters
that make 8 widgets for testing all cases (my test_tabs2.grc only have 6). 

I think my first fix whilst not corresponding to a without_hint function, fix the issue. IT seems that widgets in tabs are not exactly processed in the same way as those outside. This is probably related to the fact that the gui_hint parameter is never empty for widgets placed in tabs. 

I need further investigation to correctly fix this.

ChristopheSeguinot:
I finally found that the initial commit #4271 does 2 things: 

1. reorder block relatively to GUI_hint (this was my first step top_block.py modifications)
2. put widgets which where outside of tabs and without gui_hint value (such as 0,0...) in a different layout then those with gui_hint  value (modification in param.py)
Note that the second step was already coded for widget found inside tabs

In the present issue, we found a case where reordering block breaks GR. This is caused by the fact that GR order blocks () 
`blocks = expr_utils.sort_objects(blocks, operator.attrgetter('name'), _get_block_sort_text)`
so that widget depending on some parameter are make after the parameter is generated. The ordering against `without_gui_hint` breaks the original ordering.

So I finally conclude that the ordering using ` _get_block_sort_text` is enough to fix issues #4557 and  #3371 (keeping the param.py code modification) . 
The new code has been tested under GR 3.9.1 using test_tabs.grc

willcode:
@ChristopheSeguinot https://github.com/gnuradio/gnuradio/pull/4638 seems to work, and preserves ordering for blocks that have no gui hint param at all. Could you and @dl1ksv come to an agreement on which of these PRs should be used?

mormj (approved):

PR #4614, opened 2021-05-14
soapy: full SoapySDR 0.8 support (with 0.7.2 compatibility), minus messages

ncorgan:

### DIFF
--- a/gr-soapy/include/gnuradio/soapy/block.h
+++ b/gr-soapy/include/gnuradio/soapy/block.h
@@ -23,6 +23,29 @@ class SOAPY_API block : virtual public gr::block
 {
 public:
     /*!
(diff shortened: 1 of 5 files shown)


willcode:
Are registers in soapy required to be on `unsigned` boundaries?

Looks pretty straightforward, will read through in more detail.

ncorgan:
Yup, these interfaces are straight out of SoapySDR::Device.

willcode:
OK, merging.

Some of the types are a bit odd, like using `unsigned` and `std::string` to carry data. Since this is how it's done on SoapySDR, seems better to leave it alone.

Thanks for getting all this in here! It will be interesting to see how people start using it.

Issue #4619, opened 2021-05-14
pmt::pmt_*vector_elements() functions not exported


ncorgan:
Adding a call to one of these functions to the QA test results in the following:
```
[100%] Building CXX object gnuradio-runtime/lib/CMakeFiles/pmt_unv.dir/pmt/qa_pmt_unv.cc.o
[100%] Linking CXX executable pmt_unv
/usr/bin/ld: CMakeFiles/pmt_unv.dir/pmt/qa_pmt_unv.cc.o: in function `test_u8vector::test_method()':
qa_pmt_unv.cc:(.text+0x1052c): undefined reference to `pmt::pmt_u8vector_elements(std::shared_ptr<pmt::pmt_base>)'
collect2: error: ld returned 1 exit status
make[2]: *** [gnuradio-runtime/lib/CMakeFiles/pmt_unv.dir/build.make:96: gnuradio-runtime/lib/pmt_unv] Error 1
make[1]: *** [CMakeFiles/Makefile2:4567: gnuradio-runtime/lib/CMakeFiles/pmt_unv.dir/all] Error 2
make: *** [Makefile:141: all] Error 2
```

Confirming with `nm`:
```
nm -C pmt/libgnuradio-pmt.so | grep ' T .*vector_elem'
00000000000317a0 T pmt::s8vector_elements(std::shared_ptr<pmt::pmt_base>)
000000000002cbf0 T pmt::s8vector_elements(std::shared_ptr<pmt::pmt_base>, unsigned long&)
00000000000315e0 T pmt::u8vector_elements(std::shared_ptr<pmt::pmt_base>)
000000000002c720 T pmt::u8vector_elements(std::shared_ptr<pmt::pmt_base>, unsigned long&)
0000000000031220 T pmt::c32vector_elements(std::shared_ptr<pmt::pmt_base>)
0000000000030600 T pmt::c32vector_elements(std::shared_ptr<pmt::pmt_base>, unsigned long&)
0000000000031400 T pmt::c64vector_elements(std::shared_ptr<pmt::pmt_base>)
0000000000030f40 T pmt::c64vector_elements(std::shared_ptr<pmt::pmt_base>, unsigned long&)
0000000000032390 T pmt::f32vector_elements(std::shared_ptr<pmt::pmt_base>)
000000000002f3c0 T pmt::f32vector_elements(std::shared_ptr<pmt::pmt_base>, unsigned long&)
[...]

PR #4623, opened 2021-05-16
pmt: remove extra, mis-named pmt::pmt_ functions

willcode:

Fixes https://github.com/gnuradio/gnuradio/issues/4619


mormj (approved):

marcusmueller:
Isn't this a non-backportable ABI break, removing symbols?

jacobagilbert:
I don't think this should be backported, it appears to be removing cruft, not really fixing an issue.

willcode:
It's an API break (the ABI is already broken), but apparently anyone using the API would get a link error. Not super important to take it out, so I'll let it sit here for now.

willcode:
Even though this looks like an API change, I'm going to call it a bug and backport so no one tries to use these functions.

PR #4629, opened 2021-05-17
grc: add validation for variables of type raw

### LINKED ISSUE #4519: Unused variable not defined in flowgraph cause GNURadio crashing

dl1ksv:
At the moment grc does not validate variables, as they are of type raw.
So they are always treated as valid, which leads to some problems, see: #4519.

A approach to fix #4519 is #4531.
This approach improves the situation, but has still some problems.

param.py provides a validator function for dtype raw, but this is not used, as dtypes.py provides no validator function for dtype raw.

This fix introduces a validator function in dtypes.py which calls the corresponding function in param.py.

In addition this makes it possible, to make the code in VariableEditor.py smoother.

The error messages are kept on switching the different tabs in VariableEditor.

### DIFF
--- a/grc/core/base.py
+++ b/grc/core/base.py
@@ -23,8 +23,6 @@ class Element(object):
         Validate this element and call validate on all children.
         Call this base method before adding error messages in the subclass.
         """
-        del self._error_messages[:]
-
         for child in self.children():
             child.validate()
 
@@ -77,6 +75,7 @@ class Element(object):
         Rewrite this element and call rewrite on all children.
         Call this base method before rewriting the element.
         """
+        del self._error_messages[:]
         for child in self.children():
             child.rewrite()
 
--- a/grc/core/ports/port.py
+++ b/grc/core/ports/port.py
@@ -92,6 +92,7 @@ class Port(Element):
         return not self.dtype
 
     def validate(self):
+        del self._error_messages[:]
         Element.validate(self)
         platform = self.parent_platform
 
--- a/grc/gui/VariableEditor.py
+++ b/grc/gui/VariableEditor.py
@@ -180,21 +180,7 @@ class VariableEditor(Gtk.VBox):
             else:
                 # Evaluate and show the value (if it is a variable)
                 if block.is_variable:
-                    # Evaluate the params
-                    for key in block.params:
-                        evaluated = str(block.params[key].evaluate())
-                        # ensure that evaluated is a UTF-8 string
-                        # Old PMTs could produce non-UTF-8 strings
-                        evaluated = evaluated.encode('utf-8', 'backslashreplace').decode('utf-8')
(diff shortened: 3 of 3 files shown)


mormj (approved):
Tested this out and it seems to work beautifully, even propagated the errors from the embedded python modules - not too familiar with the pitfalls in this part of the GRC code, so looking for more reviews on this.

ChristopheSeguinot:
I tested this under GR 3.9.1 (without the black_listed_ids parameter ).
It fixes issues #4519 and #4531 in an elegant way. Moving from advanced tab and general tab , the error is always shown in red and correctly documented. 
(I did not tested for embedded python block)

dl1ksv:
Just a question: 
Why does the variable block only have the fixed type raw ?
I would be easy to extend to int, float and complex .....

skoslowski:
Looking at the diff, I can see how this would fixe problems with validating raw string parameters. The cost, however, is re-running every single raw param eval. This is a pretty steep cost considering that evals always run just before check in the rewrite stage. So, if there is an error, it should happen during `Param.rewrite()`. Why doesn't it show up after validation?

A quick glance over the code shows a call `Element.validate()` in `Param.validate()`. This clears previous errors on the node, which is why after `validate()` all seems well. 
In order to fix this, 
* either `rewrite()` must do the clearing of errors before running and `validate()` must not, or
* the error state must be passed some other way between the two stages.

dl1ksv:
The latest commit fixes this issue without an additional eval.

willcode:
In the original PR description, is "This fix introduces a validator function in dtypes.py ..." still correct?

dl1ksv:
You are right, there is no longer a validator function for the raw type. As before raw types are always treated as valid.

PR #4632, opened 2021-05-17
digital: Remove crc32 utility and most of packet_utils

mbr0wn:
Changes:
```
15543919f (Martin Braun, 6 minutes ago)
   digital: Gut packet_utils.py

   This module is a deprecated module and is used nowhere in GNU Radio, with
   the exception of a single variable (default_access_code). We keep that
   variable and remove all the rest.

   Because this module was the only consumer of crc.py, that was also removed.

   Signed-off-by: Martin Braun <martin.braun@ettus.com>

991725253 (Martin Braun, 2 hours ago)
   digital: Remove crc32 utility

   This patch removes a crc32 utility that used to be used in our CRC32 
   blocks, but no longer is. Right now, all this does is expose a generic 
   CRC32 function to the user (both via Python and C++), but there are enough
   other libraries that let the user do that. Therefore, we'll remove this.

```


mormj:
Looks like `__init__.py` is still referencing `crc.py`:

```
2021-05-17T16:06:32.8604558Z Scanning dependencies of target pygen_gr_digital_python_digital_471c2
2021-05-17T16:06:32.8646413Z make[2]: *** No rule to make target '/__w/gnuradio/gnuradio/gr-digital/python/digital/crc.py', needed by 'gr-digital/python/digital/__init__.pyc'.
2021-05-17T16:06:32.8649106Z make[2]: Target 'gr-digital/python/digital/CMakeFiles/pygen_gr_digital_python_digital_471c2.dir/build' not remade because of errors.
2021-05-17T16:06:32.8650644Z make[1]: *** [CMakeFiles/Makefile2:6887: gr-digital/python/digital/CMakeFiles/pygen_gr_digital_python_digital_471c2.dir/all] Error 2
```

marcusmueller (approved):
Burninate all the cruft. Good change – I think I was too hesitant touching packet_utils.py back when we got rid of most of the python something_mod* blocks.

Issue #4615, opened 2021-05-14
Interleave / Deinterleave GRC block: inconsistent naming of parameters


marcusmueller:
These blocks use "Vec Length" instead of "vector length" AND this is not documented on the [wiki page](https://wiki.gnuradio.org/index.php/Deinterleave)s. It's really just a multiplier for the itemsize.

To fix this, 

- [x] Fix the GRC .block.yml files to use the canonical "vector size" / "vector length" (look that up)
- [x] Modify the wiki pages for the deinterleave / interleave blocks, explaining what the effect actually is, maybe make a figure / drawing to understand what the interleaver does when you set vector size >1.

solomonbstoner:
I notice sometimes "vector size" is used to refer to "vector length". Shall I change em to "vector length" or is "vector size" ok? Is the goal of this issue just to standardize "vec" to "vector" in the documentation?

![Screenshot from 2021-05-18 23-22-35](https://user-images.githubusercontent.com/24548334/118679106-2db96d80-b830-11eb-80cc-609cae1e48d4.png)

solomonbstoner:
We're left with part 2 of this issue:

> Modify the wiki pages for the deinterleave / interleave blocks, explaining what the effect actually is, maybe make a figure / drawing to understand what the interleaver does when you set vector size >1.

PR #4633, opened 2021-05-18
gr-blocks: Standardize Vector parameter name

solomonbstoner:
Addresses point 1 of Issue #4615. Replaces all label instances of "Vec
Length" to "Vector Length". Also replaces "Vec Rate" to "Vector Rate"
for standardization.

There is no occurrence of "vec size", verified with the command:
`find . -name "*.yml" -exec grep -Hin "vec size" {} ';'` .

All instances of "vec length" were found and replaced with the 
commands:
```
find . -name "*.yml" -exec grep -Hin "vec length" {} ';'
find . -name "*.yml" -exec sed -i 's/Vec Length/Vector Length/g' {} ';'
```

### DIFF
--- a/gr-blocks/grc/blocks_abs_xx.block.yml
+++ b/gr-blocks/grc/blocks_abs_xx.block.yml
@@ -4,7 +4,7 @@ flags: [ python, cpp ]
 
 parameters:
 -   id: vlen
-    label: Vec Length
+    label: Vector Length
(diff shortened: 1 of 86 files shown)


dkozel (approved):
Makes user facing labels clearer, yes!

mormj (approved):

Issue #4634, opened 2021-05-18
Interleave block: Unable to modify Vector Length


solomonbstoner:
I was just playing around with vectors to learn something about it. I tried to change the interleaver block from Vector Length of 3 to Vector Length of 1. I was not allowed to directly change the Vector Lengths for all the blocks from 3 to 1. Doing so gave me the error below for the bottom vector source:

> Source IO size "12" does not match sink IO size "4".

![Screenshot from 2021-05-18 22-37-52](https://user-images.githubusercontent.com/24548334/118671097-c3053380-b829-11eb-89b0-4f6a375e56c8.png)

Instead, to make the change, I had to go the long route. So, back to beginning, everything is fine with Vector Length of 3.

![Screenshot from 2021-05-18 22-34-57](https://user-images.githubusercontent.com/24548334/118670749-7883b700-b829-11eb-8f10-a00b975f72d8.png)

I change Vector Length of 3 to 1 in the Interleave block only. The others are not changed yet. Only the upper input and output ports have changed from vector length of 3 to 1, shown in the red arrow indicating an error. The bottom input port somehow still has a vector length of 3.

![Screenshot from 2021-05-18 22-35-11](https://user-images.githubusercontent.com/24548334/118671446-0cee1980-b82a-11eb-83ce-a1a8a064d5a0.png)

I change the Vector Length only for the blocks with the errors. Now there are _no more errors_ . Note that the vector source at the bottom still have a Vector Length of 3. All other blocks have their Vector Length changed from 3 to 1. I can generate the flowgraph and run, but _nothing gets shown_. The QT Time sink does not show.

![Screenshot from 2021-05-18 22-35-28](https://user-images.githubusercontent.com/24548334/118671589-327b2300-b82a-11eb-8e7a-41a3236e4b03.png)

If I were to manaully change the Vector Length of the bottom vector source myself, an error appears. I cannot make that change even though I had specified all the blocks to have Vector Length of 1.

![Screenshot from 2021-05-18 22-37-52](https://user-images.githubusercontent.com/24548334/118671915-71a97400-b82a-11eb-8ac3-446fa0424ebb.png)

The only way I could was to double click on the Interleave block to enter its properties, then click "Cancel". Clicking on "Ok" does not work. Only then does the error go away and the program runs properly (ie QT Time sink appears).

solomonbstoner:
Imma try to fix this issue myself

willcode:
[...]

PR #4637, opened 2021-05-19
grc: Fix vector length bug

solomonbstoner:
This commit resolves Issue #4634 by forcing the flowgraph to reimport
the data it exports.

### DIFF
--- a/grc/core/blocks/block.py
+++ b/grc/core/blocks/block.py
@@ -213,6 +213,7 @@ class Block(Element):
         for port in ports:
             if hasattr(port, 'master_port'):  # Not a master port and no left-over clones
                 port.dtype = port.master_port.dtype
+                port.vlen = port.master_port.vlen
                 continue
             nports = port.multiplicity
             for clone in port.clones[nports-1:]:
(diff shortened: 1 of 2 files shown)


solomonbstoner:
Apparently I could bypass Issue #4634 by pressing `Apply` then `Cancel`. And everything would be peachy. And the difference between the `Ok` button and the `Cancel` button is the import of the current state. So I made the import happen regardless of what button was pressed, and voila. The issue is gone.

solomonbstoner:
As I suspected, something has modified `vlen` just for only the first input and output block. Im one step closer

Screenshot from 2021-05-20 23-21-53

solomonbstoner:
@skoslowski With #4637, upon pressing `Ok` or `Apply`, the vector length of all the ports will be updated. Fixes #4634. For your review and consideration, please

willcode (approved):

mormj (approved):

dkozel:
Nice find! This seems very clean and minimal. Did Sebastian see this change in chat and approve it?

skoslowski (approved):
This looks good to me. Thanks!
(and sorry for the delay)

PR #4638, opened 2021-05-19
grc: fixing build order for blocks with gui hints

### LINKED ISSUE #4631: Gnuradio 3.8 AttributeError: 'top_block_sptr' object has no attribute 'tabid_grid_layout_0'
In this simple flowgraph:
image
image
image

I'm getting the following error:
  File "/home/aaa/Desktop/Prova.py", line 196, in <module>
    main()
  File "/home/xxx/Desktop/Prova.py", line 172, in main
    tb = top_block_cls()
  File "/home/aaa/Desktop/Prova.py", line 128, in __init__
    self.Tab_grid_layout_0.addWidget(self._qtgui_time_sink_x_0_win, 1, 1, 1, 1)
  File "/home/aaa/sdr/lib/python3/dist-packages/gnuradio/gr/hier_block2.py", line 93, in __getattr__
AttributeError: 'top_block_sptr' object has no attribute 'Tab_grid_layout_0'

>>> Done (return code 1)

Gnuradio Companion Version: v3.8.2.0-118-g8b2012ab (Python 3.6.9)

dl1ksv:
In some cases the build order of blocks with gui hints is wrong,
which leads to errors.

See #4631.
This pr fixes #4631

### DIFF
--- a/grc/core/generator/top_block.py
+++ b/grc/core/generator/top_block.py
@@ -198,10 +198,10 @@ class TopBlockGenerator(object):
         blocks = expr_utils.sort_objects(blocks, operator.attrgetter('name'), _get_block_sort_text)
 
         # Ordering blocks : blocks with GUI Hint must be processed first to avoid PyQT5 superposing blocks
+        # The block must have a GUI HINT, but the hint may be empty
         def without_gui_hint(block):
-            hint = block.params.get('gui_hint')
-            return hint is None or not hint.get_value()
-
+            return block.params.get('gui_hint') is None
+ 
         blocks.sort(key=without_gui_hint)
 
         blocks_make = []


willcode:
Does this replace https://github.com/gnuradio/gnuradio/pull/4609?

dl1ksv:
Sorry, I was not aware of #4609 , and yes this pr is a replacement.
The tab wigdet must be generated before other widgets that will be inserted into this tab.
As Christophe noticed the first sort generates a correct order of the blocks.
The second sort sorts the blocks with a gui_hint before the blocks without a gui_hint but the second sort garbles the sort order inside the group of block with gui_hint.

But in the past there must have been some problems with PyQT5.
So let's keep the second sort but modify it to keep the sort order inside the blocks with / without gui_hints.

That's what my pr does.

willcode:
The current code was addressing a problem where if a widget had no hint, it could be placed at (0,0), then if a widget with hint (0,0) was processed later, the two would be superimimposed. Complicating this, widgets that a placed in a tab use the same parameter to specify a parent tab and position in that tab. The logic has to do construction in the right order, considering both of these uses. Do you think this PR does that?

dl1ksv:
Yes !

willcode:
This works in a few tests I tried with tabs and ordering.

mormj:
Yes, this seems to fix the issue

mormj (approved):

dl1ksv:
@ChristopheSeguinot  you commented in commit 9a25928 
        # Ordering blocks : blocks with GUI Hint must be processed first to avoid PyQT5 superposing blocks
Do you have an example for such a conflict ?

I think 
https://github.com/gnuradio/gnuradio/blob/b029dc427ea633409069198823390abd912b18e7/grc/core/generator/top_block.py#L198
sorts the blocks in a correct dependency, but mixes  blocks with and without gui hint.

https://github.com/gnuradio/gnuradio/blob/b029dc427ea633409069198823390abd912b18e7/grc/core/generator/top_block.py#L205
sorts blocks with gui hint in front of blocks without gui hints but changes the order of blocks with gui hint so the dependencies may be violated which leads to the bug.

So if the blocks with and without gui hint can be mixed we could remove the second sort, otherwise we must modify the sort in the proposed way,

willcode:
@dl1ksv @ChristopheSeguinot any conclusions and this PR and https://github.com/gnuradio/gnuradio/pull/4609?

dl1ksv:
Up to now I have not found any example, where all blocks with gui_hint must be sorted before blocks without gui_hint.

As this sorting was introduced by @ChristopheSeguinot and now he proposed to remove this sorting, I suggest to follow his proposal. So close this pr in favour of #4609

Issue #4640, opened 2021-05-20
UHD tuner_request_t::policy_t has no Python binding


Codey-McCodeface:
I am unable to set the `rf_freq_policy` or `dsp_freq_policy` of a uhd tune request from Python.

Adding the binding for the enum did solve this problem. The changes below worked for me:

```diff
diff --git a/gr-uhd/python/uhd/bindings/uhd_types_python.cc b/gr-uhd/python/uhd/bindings/uhd_types_python.cc
index c593ce480..f05df299c 100644
--- a/gr-uhd/python/uhd/bindings/uhd_types_python.cc
+++ b/gr-uhd/python/uhd/bindings/uhd_types_python.cc
@@ -141,7 +141,8 @@ void bind_uhd_types(py::module& m)
 
     // ::uhd::tune_request_t
     using tune_request_t = ::uhd::tune_request_t;
-    py::class_<tune_request_t>(m, "tune_request_t")
+    py::class_<tune_request_t> tune_request(m, "tune_request_t");
+    tune_request
         .def(py::init<double, double>())
         .def(py::init<double>(), py::arg("target_freq") = 0.0)
         .def_readwrite("target_freq", &tune_request_t::target_freq)
@@ -151,6 +152,13 @@ void bind_uhd_types(py::module& m)
         .def_readwrite("dsp_freq", &tune_request_t::dsp_freq)
         .def_readwrite("args", &tune_request_t::args);
 
+    // ::uhd::tune_request_t::policy_t
+    using policy_t = ::uhd::tune_request_t::policy_t;
+    py::enum_<policy_t>(tune_request, "policy_t")
+        .value("POLICY_NONE", policy_t::POLICY_NONE)
+        .value("POLICY_AUTO", policy_t::POLICY_AUTO)
+        .value("POLICY_MANUAL", policy_t::POLICY_MANUAL)
+        .export_values();
 
     // ::uhd::sensor_value_t
     using sensor_value_t = uhd::sensor_value_t;
```

mormj:
@Codey-McCodeface - this appears to the be correct solution to the issue you are seeing - would you mind submitting this as a pull request and reference this issue?

willcode:
Cross ref did not trigger auto-close. Manu-closing.

PR #4641, opened 2021-05-21
gr-uhd: Adding missing Python binding for uhd::tune_request_t::policy_t.

Codey-McCodeface:
Fix for issue: #4640

### DIFF
--- a/gr-uhd/python/uhd/bindings/uhd_types_python.cc
+++ b/gr-uhd/python/uhd/bindings/uhd_types_python.cc
@@ -141,7 +141,8 @@ void bind_uhd_types(py::module& m)
 
     // ::uhd::tune_request_t
     using tune_request_t = ::uhd::tune_request_t;
-    py::class_<tune_request_t>(m, "tune_request_t")
+    py::class_<tune_request_t> tune_request(m, "tune_request_t");
+    tune_request
         .def(py::init<double, double>())
         .def(py::init<double>(), py::arg("target_freq") = 0.0)
         .def_readwrite("target_freq", &tune_request_t::target_freq)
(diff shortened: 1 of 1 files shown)


mormj:
@Codey-McCodeface - Thank you for the contribution.  Two simple things to get it to run through the CI tests:

1) Please sign off your commit.  GNU Radio adheres to the DCO (https://wiki.gnuradio.org/index.php/Development#What.27s_this_Developer_Certificate_of_Origin_.28DCO.29.3F) -- this can be done simply by `git commit --amend -s`, then force push the amended commit
2) The clang-format check is failing - the lazy approach here is just run it on the whole codebase: `find ./ -iname *.h -o -iname *.cc | xargs clang-format -i` from the top level source directory

Please let me know if you run into any issues.

Codey-McCodeface:
@mormj Thanks. I think it is all set.

gr-uhd/python/uhd/bindings/uhd_types_python.cc
```diff
@@ -141,8 +141,8 @@ void bind_uhd_types(py::module& m)
 
     // ::uhd::tune_request_t
     using tune_request_t = ::uhd::tune_request_t;
-    py::class_<tune_request_t>(m, "tune_request_t")
-        .def(py::init<double, double>())
+    py::class_<tune_request_t> tune_request(m, "tune_request_t");
+    tune_request.def(py::init<double, double>())
```
willcode, line 145: What is this change for?

mormj: usually the class definition gets broken up that way to shove an enum in with the class, but nothing is being added here.  the change is not harmful, but also not necessary

Codey-McCodeface: I needed to reference the tune_request python class to put the policy_t enum within it. I was trying to be consistent with the uhd hierarchy.

I now see that other uhd enums are just defined at the module level. Would this be preferred over nesting it within tune_request?

mormj: Yes, what you did makes sense - that is indeed the proper way to get the enum nested under the class name in the bindings.

mormj (approved):

Issue #2366, opened 2019-02-18
CRC16


michael-roe:
(Feature request)

gnuradio has a standard block for CRC32, but not for CRC16. CRC16 with the CCITT polynomial (x**16+x**12+x**5+1) is used in lots of radio protocols (e.g. NATO STANAG 4202, Yaesu System Fusion) so I think it should be included in GNU radios standard library of blocks.

(Ok, I could write a custom block to do this),

noc0lour:
Maybe the different CRC implementations could be added as generic FEC decoder/encoder

noc0lour:
Moving away from the CRC tagged stream block we currently have

marcusmueller:
> (Ok, I could write a custom block to do this),

Code contributions more than welcome!

I generally agree with @noc0lour's sentiment that this would be nice to have in a more general way, but adding a generally usable CRC16 would be a good thing, sooner than later.

daniestevez:
Hi! I just stumbled upon this issue. I also have a generic CRC block on the roadmap for gr-satellites, since I keep adding ad-hoc CRC implementations. I can't see to find the time to get to it, but my idea is to have something where the CRC parameters can be specified in the manner of [pycrc](https://pypi.org/project/pycrc/).

For gr-satellites the main use case is checking a CRC at the end of a PDU, but adding a CRC at the end of a PDU is also an important use case.

daniestevez:
After [looking at the current CRC32 block](https://github.com/gnuradio/gnuradio/blob/4d306e1f98eb60cf62ca8f0aa59e3939e7826e84/gr-digital/lib/crc32_async_bb_impl.h#L23), I think that an easy way to get a generic CRC block would be to design an interface around `boost::crc_optimal`. Though this is not very good if we want to get away from using boost as a dependency.

PR #4649, opened 2021-05-23
Crc16 addded. Issue #2366

cmrincon:
It resolves the issue #2366. It's a modified copy of the crc32_async_bb block.

There is a basic test for the block attached.
[crc16_async_first_unit_test.grc.gz](https://github.com/gnuradio/gnuradio/files/6529107/crc16_async_first_unit_test.grc.gz)


gr-digital/lib/crc16_async_bb_impl.h
```diff
@@ -0,0 +1,48 @@
...
+#include <boost/crc.hpp>
+
+namespace gr {
+namespace digital {
+
+class crc16_async_bb_impl : public crc16_async_bb
+{
+private:
+    
+    boost::crc_optimal<16, 0x1021, 0xFFFF, 0, false, false> d_crc_ccitt_impl;
```
jwmelto, line 24: I don't know anything about this stuff, but [this link](http://stackoverflow.com/questions/17196743/ddg#17196947) suggests that either you should start with 0x0000 or this isn't really CCITT?

gr-digital/python/digital/bindings/crc16_async_bb_python.cc
```diff
@@ -0,0 +1,48 @@
...
+ * Copyright 2021 Cesar Martinez.
+ *
+ * This file is part of GNU Radio
+ *
+ * SPDX-License-Identifier: GPL-3.0-or-later
+ * 
+ */
+
+/***********************************************************************************/
+/* This file is automatically generated using bindtool and can be manually edited  */
```
jwmelto, line 12: if this file is generated, should it be controlled?

gr-digital/python/digital/bindings/docstrings/crc16_async_bb_pydoc_template.h
```diff
@@ -0,0 +1,25 @@
...
+ * This file is part of GNU Radio
+ *
+ * SPDX-License-Identifier: GPL-3.0-or-later
+ * 
+ */
+#include "pydoc_macros.h"
+#define D(...) DOC(gr, digital, __VA_ARGS__)
+/*
+  This file contains placeholders for docstrings for the Python bindings.
+  Do not edit! These were automatically extracted during the binding process
```
jwmelto, line 14: if generated, should these be controlled?

jwmelto (commented):
**CAVEAT!** This is my first GNU Radio PR review. I have no history with the code base, so some of my comments may be inappropriate for this project. I have no approval authority or any other reason why I should be listened to, except for my decades of experience writing and reviewing C++ code.

It's possible that the CRC16 code is little more than the CRC32 code with the polynomial parameters changed, but if that's the case, it argues for an abstraction of the common bits and only the unique parts changed.

YMMV

gr-digital/include/gnuradio/digital/crc16_async_bb.h
```diff
@@ -0,0 +1,58 @@
...
+ * calculated on the PDU and appended to it. The output is then 2
+ * bytes longer than the input.
+ *
+ * This block implements the CRC16 using the Boost crc_optimal
+ * class for 16-bit CRCs with the standard generator 0x1021.
+ */
+class DIGITAL_API crc16_async_bb : virtual public block
+{
+public:
+    typedef std::shared_ptr<crc16_async_bb> sptr;
```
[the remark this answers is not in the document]
daniestevez: I think the `typedef` is standard in most (or all) of the GNU Radio blocks. I don't know what approach is better, since I'm not so familiar with C++11.

gr-digital/lib/crc16_async_bb_impl.cc
```diff
@@ -0,0 +1,105 @@
...
+#endif
+
+#include "crc16_async_bb_impl.h"
+#include <gnuradio/io_signature.h>
+#include <volk/volk.h>
+
+namespace gr {
+namespace digital {
+
+crc16_async_bb::sptr crc16_async_bb::make(bool check)
```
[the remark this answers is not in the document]
daniestevez: Again, this is how most GNU Radio blocks look like (@cmrincon just copied the skeleton from a similar CRC block).

@jwmelto, I think your comments about code style are quite valuable, but they apply to all the code base and the code generation tools (like `gr_modtool`). Maybe it's good to open an issue about this to come up with good suggestions?

gr-digital/lib/crc16_async_bb_impl.h
```diff
@@ -0,0 +1,48 @@
...
+public:
+    crc16_async_bb_impl(bool check);
+    ~crc16_async_bb_impl() override;
+
+    int general_work(int noutput_items,
+                     gr_vector_int& ninput_items,
+                     gr_vector_const_void_star& input_items,
+                     gr_vector_void_star& output_items) override;
+
+    uint64_t d_npass;
```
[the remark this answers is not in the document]
daniestevez: Agreed. I think these should be marked `private:`

gr-digital/lib/crc16_async_bb_impl.cc
```diff
@@ -0,0 +1,105 @@
...
+    return gnuradio::make_block_sptr<crc16_async_bb_impl>(check);
+}
+
+crc16_async_bb_impl::crc16_async_bb_impl(bool check)
+    : block("crc16_async_bb", io_signature::make(0, 0, 0), io_signature::make(0, 0, 0)),
+      d_npass(0),
+      d_nfail(0)
+{
+    d_in_port = pmt::mp("in");
+    d_out_port = pmt::mp("out");
```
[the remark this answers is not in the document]
daniestevez: I think the code base is scattered of examples like this, and also many examples where `pmt::mp("out")` is repeated several times throughout the block implementation instead of being stored in a class member as here.

Perhaps an idea for improvement is to use [`msgport_names`](https://github.com/gnuradio/gnuradio/blob/4d306e1f98eb60cf62ca8f0aa59e3939e7826e84/gnuradio-runtime/include/gnuradio/pdu.h#L19), which contains standard PDU port names. However `"in"`, and `"out"` are not included there, so we'd like to add them first. Maybe it's worth discussing how to improve this in a separate issue?

For this PR I think this is OK, since I guess it's copied over from the other CRC block.

gr-digital/lib/crc16_async_bb_impl.cc
```diff
@@ -0,0 +1,105 @@
...
+    message_port_register_in(d_in_port);
+    message_port_register_out(d_out_port);
+
+    if (check)
+        set_msg_handler(d_in_port, [this](pmt::pmt_t msg) { this->check(msg); });
+    else
+        set_msg_handler(d_in_port, [this](pmt::pmt_t msg) { this->calc(msg); });
+}
+
+crc16_async_bb_impl::~crc16_async_bb_impl() {}
```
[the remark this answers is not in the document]
daniestevez: This is very common in the code base and code generated by `gr_modtool`. Probably a good opportunity to improve the whole code base.

gr-digital/python/digital/bindings/crc16_async_bb_python.cc
daniestevez: Yes. This is how the pybind11 bindings are handled. Probably others can give good reasons why these are in the VCS.

gr-digital/python/digital/bindings/docstrings/crc16_async_bb_pydoc_template.h
daniestevez: Yes. This is how the pybind11 bindings are handled. Probably others can give good reasons why these are in the VCS.

gr-digital/lib/crc16_async_bb_impl.cc
```diff
@@ -0,0 +1,105 @@
...
+
+void crc16_async_bb_impl::calc(pmt::pmt_t msg)
+{
+    // extract input pdu
+    pmt::pmt_t meta(pmt::car(msg));
+    pmt::pmt_t bytes(pmt::cdr(msg));
+
+    unsigned int crc;
```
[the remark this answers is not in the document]
daniestevez: Technically, this should hold at least 16 bits, in order to hold a CRC-16. For me it would be a good idea to use `uint16_t`, since it even helps convey that this is a CRC-16. However, how about instead of declaring `crc` here, doing
```
auto crc = d_crc_ccitt_impl();
```
below and letting the return type of `d_crc_ccitt_impl()` handle this?

gr-digital/lib/crc16_async_bb_impl.cc
```diff
@@ -0,0 +1,105 @@
...
+    const uint8_t* bytes_in = pmt::u8vector_elements(bytes, pkt_len);
+    std::vector<uint8_t> bytes_out(2 + pkt_len);
+
+    d_crc_ccitt_impl.reset();
+    d_crc_ccitt_impl.process_bytes(bytes_in, pkt_len);
+    crc = d_crc_ccitt_impl();
+    memcpy((void*)bytes_out.data(), (const void*)bytes_in, pkt_len);
+    memcpy((void*)(bytes_out.data() + pkt_len),
+           &crc,
+           2); // FIXME big-endian/little-endian, this might be wrong
```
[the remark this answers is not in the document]
daniestevez: Agreed, and I've just checked that this is also a [problem in the original CRC-32 block](https://github.com/gnuradio/gnuradio/blob/4d306e1f98eb60cf62ca8f0aa59e3939e7826e84/gr-digital/lib/crc32_async_bb_impl.cc#L61), so we should also fix it there.

I don't like that this depends on endianness, but there isn't an obvious standard choice either. Maybe we can rely on network order (big-endian) and use `ntohs()` here?

gr-digital/include/gnuradio/digital/crc16_async_bb.h
```diff
@@ -0,0 +1,58 @@
...
+ *
+ * Processes packets (as async PDU messages) for CRC16. The \p
+ * check parameter determines if the block acts to check and strip
+ * the CRC or to calculate and append the CRC16.
+ *
+ * The input PDU is expected to be a message of packet bytes.
+ *
+ * When using check mode, if the CRC passes, the output is a
+ * payload of the message with the CRC stripped, so the output
+ * will be 4 bytes smaller than the input.
```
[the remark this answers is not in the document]
daniestevez: Perhaps a copy & paste problem, since this comes from adapting a CRC-32 block?

gr-digital/include/gnuradio/digital/crc16_async_bb.h
```diff
@@ -0,0 +1,58 @@
...
+ */
+class DIGITAL_API crc16_async_bb : virtual public block
+{
+public:
+    typedef std::shared_ptr<crc16_async_bb> sptr;
+
+    /*!
+     * \param check Set to true if you want to check CRC, false to create CRC.
+     */
+    static sptr make(bool check = false);
```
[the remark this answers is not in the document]
daniestevez: I think the use case is that you can instance a block that either checks the CRC or adds the CRC. This is what seems more common in all applications that come to mind, since you are either transmitting things (so you need to put a CRC) or receiving things (so you need to check the CRC). For a transceiver you would have two separate data paths and CRC blocks for the received packets and transmitted packets.

gr-digital/lib/crc16_async_bb_impl.cc
```diff
@@ -0,0 +1,105 @@
...
+    unsigned int crc;
+    size_t pkt_len(0);
+    const uint8_t* bytes_in = pmt::u8vector_elements(bytes, pkt_len);
+
+    d_crc_ccitt_impl.reset();
+    d_crc_ccitt_impl.process_bytes(bytes_in, pkt_len - 2);
+    crc = d_crc_ccitt_impl();
+    if (crc != *(unsigned int*)(bytes_in + pkt_len - 2)) { // Drop package
+        d_nfail++;
+        return;
```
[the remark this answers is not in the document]
daniestevez: This is probably a very personal matter of style. For me it is completely OK to provide an early return in case some pre-condition does not hold (like in this case, that the CRC is not valid). A (questionable) advantage of the early return is that it avoids an extra level of indentation for the rest of the code.

gr-digital/lib/crc16_async_bb_impl.h
daniestevez: In my [favourite online CRC calculator](http://www.sunshine2k.de/coding/javascript/crc/crc_js.html), you can see that there are two flavours of CRC-16-CCITT. What is called there CCITT_ZERO, which starts with `0x0000`, and CCITT_FALSE, which starts with `0xffff`.

To make things more complicated, the CRC-16-CCITT used by AX.25 is also different. It has a final `0xffff` XOR value and reflects the input and output.

It seems that the intention here was to implement CCITT_FALSE, which I think it's more popular than CCITT_ZERO. Looking at #2366, the idea was to implement the CRC used by STANAG 4202, and Yaesu System Fusion, but I don't know what flavour these use.

gr-digital/lib/crc16_async_bb_impl.h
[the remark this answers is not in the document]
daniestevez: On second thought, I think that these variables are only used to count how many packets pass and fail the CRC check. These variables probably come from the original CRC-32 block, but I wonder if some other block or code uses them.

daniestevez (changes requested):
@cmrincon, thanks for your contribution. I think that most of our comments are about issues that are also present in the original CRC-32 block or about improving the code style in the whole code base.

I'd perhaps like to see the feedback of @marcusmueller and/or @noc0lour (who were involved in the discussion in #2366). To me this block seems quite useful, and I think its current implementation is OK. However, I also think that this is essentially a duplication of the CRC-32 block, where only the definition of the CRC via `boost::crc_optimal` and the length of the CRC change. Perhaps this is an opportunity to try to modify instead the CRC-32 block to let the user supply the CRC parameters to `boost::crc_optimal`?

Besides this, please take care to add a signed-off line as described [here](https://github.com/gnuradio/gnuradio/blob/master/CONTRIBUTING.md#dco-signed) and to format the code using `clang-format`.

daniestevez:
@jwmelto thanks for all your comments. Your review is very much appreciated. As I have mentioned in my replies, I think there is an opportunity to make these kinds of code style improvements in the whole GNU Radio code base. Could you maybe open up a new Issue to describe what improvements you think should be done so that people can discuss and start applying them consistently to all the code base?

gr-digital/lib/crc16_async_bb_impl.h
daniestevez, line 24: Here you can use `boost::crc_ccitt_type` instead of `boost::crc_optimal<16, 0x1021, 0xFFFF, 0, false, false>`, by the way.

gr-digital/lib/crc16_async_bb_impl.cc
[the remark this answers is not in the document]
jwmelto: with all of the "this is like the majority of the codebase" comments, I suspected that might be the case. However, unlike some of the stylistic areas where the choice is purely subjective, cyclomatic complexity is something that can be objectively measured. 

On the whole, lower complexity is better; less complex code is easier to read and easier to test. Your point about the extra layer of indentation is the most common excuse, but it's masking code that is too complex and should be decomposed more. Pascal's [famous quote](https://quoteinvestigator.com/2012/04/28/shorter-letter/) applies here; writing good code is hard, but the effort required should rarely be justification for choosing an objectively less correct approach.

I'm nobody; the project is free to code as it sees fit. I offered my comments in the spirit of education. Accept them or reject them as you see fit.

jwmelto:
> @jwmelto thanks for all your comments. Your review is very much appreciated. As I have mentioned in my replies, I think there is an opportunity to make these kinds of code style improvements in the whole GNU Radio code base. Could you maybe open up a new Issue to describe what improvements you think should be done so that people can discuss and start applying them consistently to all the code base?

I won't open an issue; I don't have time to take on another open source project; I only commented here because I helped @cmrincon with his **git** issues in IRC; since it was a small PR, I took the time to comment.

Evolution of a code base must be gradual. When something new is learned (or the language changes), no one can afford to refactor everything to use the new feature or style. I agree with you that the project would benefit from a thrust towards more modern software practices, but it will have to be something that arises organically from those who are actively developing/maintaining.

gr-digital/python/digital/bindings/crc16_async_bb_python.cc
mormj: This is an unfortunate consequence of our workflow for generating python bindings - since the bindings are scraped from the public headers for each block using a tool called `pygccxml` - 1. this tool is very slow, and 2. for some blocks, the python bindings might not want to expose everything, or do some things differently. 

Perhaps an improvement here would be removing the auto-generated comment for the in-tree bindings.

mormj:
@jwmelto - these are very informative comments and thank you for taking the time to do a thorough review.  As discussed above with @daniestevez - some should be applied directly to this PR and some should be taken up separately throughout the codebase.

@cmrincon - Thank you for the contribution here - would you mind signing off the DCO (`git commit --amend --no-edit --signoff`), and doing the `clang-format` (`find ./ -iname *.h -o -iname *.cc | xargs clang-format -i`)

Also, per @daniestevez 's comments, the similarity with the CRC32 block could be addressed in several ways - perhaps by templating?  But in any case, I think this block is valuable as-is (with some minor tweaks)

daniestevez:
I also think that this block is valuable as-is.

Looking ahead, I think that it would be good to have a block that gives the user access to all the parameters of [`boost::crc_basic()`](https://www.boost.org/doc/libs/1_66_0/libs/crc/crc.html#crc_basic) and then perhaps some specific implementations for the more popular codes using [`boost::crc_optimal()`](https://www.boost.org/doc/libs/1_66_0/libs/crc/crc.html#crc_optimal) and some form of templating to avoid repeated code.

cmrincon:
> @cmrincon - Thank you for the contribution here - would you mind signing off the DCO (`git commit --amend --no-edit --signoff`), and doing the `clang-format` (`find ./ -iname *.h -o -iname *.cc | xargs clang-format -i`)

Done. I'm not sure how can i check if i did it properly. Can you check it? Thanks you

mormj:
Doesn't appear to have fixed it - you need to have each of your commits signed - probably the easiest here is to squash your commits to 1 since they are for a single feature.  One way to do this is:
```
git rebase -i
```
Then change your first commit to `edit`, then the rest to `f`
git will then want you to edit that first commit do this with `git commit -s --amend`
then `git rebase --continue`

mormj:
@cmrincon - thanks for the DCO and rebase.  That appears to have worked.

Looking at the above feedback, it would be good to get a bit more of it incorporated, especially the cleanup around duplicate code.  The `boost::crc_ccitt_type` is a good suggestion as well.

cmrincon:
Thanks you all for your suggestions. As @mormj suggested i will cleanup the duplicate code and use boost::crc_ccitt_type as well, however, i think that most of the other suggestions should be discussed in another issue as they affects to other tools and blocks.

gr-digital/include/gnuradio/digital/crc16_async_bb.h
```diff
@@ -5,7 +5,11 @@
  * This file is part of GNU Radio
  *
  * SPDX-License-Identifier: GPL-3.0-or-later
+<<<<<<< HEAD
```
daniestevez, line 8: Notice these leftover merge markers.

gr-digital/lib/crc16_async_bb_impl.cc
```diff
@@ -5,7 +5,11 @@
  * This file is part of GNU Radio
  *
  * SPDX-License-Identifier: GPL-3.0-or-later
+<<<<<<< HEAD
```
daniestevez, line 8: Merge markers.

gr-digital/lib/crc16_async_bb_impl.h
```diff
@@ -20,11 +20,22 @@ namespace digital {
 class crc16_async_bb_impl : public crc16_async_bb
 {
 private:
+<<<<<<< Updated upstream
```
daniestevez, line 23: Merge markers.

gr-digital/lib/crc16_async_bb_impl.h
```diff
@@ -20,11 +20,22 @@ namespace digital {
...
+    boost::crc_ccitt_type d_crc_ccitt_impl;
+=======
+<<<<<<< HEAD
+=======
+    
+>>>>>>> 868761d3565a1ba6092f258ed1554fae63e81945
     boost::crc_optimal<16, 0x1021, 0xFFFF, 0, false, false> d_crc_ccitt_impl;
+>>>>>>> Stashed changes
```
daniestevez, line 31: Merge markers.

gr-digital/python/digital/bindings/crc16_async_bb_python.cc
```diff
@@ -5,7 +5,11 @@
  * This file is part of GNU Radio
  *
  * SPDX-License-Identifier: GPL-3.0-or-later
+<<<<<<< HEAD
```
daniestevez, line 8: Merge markers.

gr-digital/python/digital/bindings/crc16_async_bb_python.cc
```diff
@@ -15,7 +19,11 @@
 /* BINDTOOL_GEN_AUTOMATIC(0)                                                       */
 /* BINDTOOL_USE_PYGCCXML(0)                                                        */
 /* BINDTOOL_HEADER_FILE(crc16_async_bb.h)                                        */
+<<<<<<< HEAD
```
daniestevez, line 22: Merge markers

daniestevez (commented):
Thanks for this. I've noticed you have some left over merge markers (things like `<<<<<<< HEAD`, etc.). I've tried to spot all of them with individual comments.

gr-digital/python/digital/bindings/crc16_async_bb_python.cc
cmrincon: Thanks you @daniestevez. Im not used to use git and things like this are a headhache for me.

daniestevez (approved):
Thanks! At this moment the PR looks good to me. I suggest squashing all the commits before or while merging.

cmrincon:
> Thanks! At this moment the PR looks good to me. I suggest squashing all the commits before or while merging.

I can do a reset + commit. Is there any other way this can be done easily?

daniestevez:
Basically, from your branch, `git rebase -i master`, choose to squash all the commits (except the first one). Then `git push --force` when the rebase is done.

mormj:
@cmrincon  - thank you for diligently going through the process to get this PR reviewed and cleaned up.  Merging.

PR #4657, opened 2021-05-24
runtime: Explicitly convert path to string to fix MSVC build

ryanvolz:
Noticed and and tested over at #4410.


mormj (approved):

willcode:
Note for backport: need `boost::filesystem::canonical(path).string()`

willcode:
Not needed for 3.9. Already uses `string()` in corresponding code.

PR #4661, opened 2021-05-24
python: replaced 'import *' with explicit imports

MarkBauer5:
Signed-off-by: Mark Bauer <markb5@illinois.edu>


willcode:
@MarkBauer5 thanks for taking a look at this. There is a lot python code that could use cleanup, PEP8ing, etc., and it's good to have someone work on it.

Before we start enforcing rules like "don't import all", could you take a look at the various `from x import *` and see which ones could unnecessarily pollute the namespace in a way that could cause confusion. While it's a good idea not to `import *` in general, it's not always a bad thing if a module designer does a relative import like

`from .my_python_helpers import *`

Or it could be! Depends on the situation.

There could also be other ways to clean some of this up. For instance, the place where only `np` is used could clearly import numpy itself. And `gr-fec/python/fec/polar/helper_functions.py` should place `def main()` inside the `if __name__` so `main()` doesn't get exported.

willcode:
@MarkBauer5 you'll end up having an easier time with development if you do work on branches off master. Then you can rebase your feature/fix branch instead of merging gnuradio/master into yours. I noticed your PR was off MarkBauer5/master, and that usually makes merging harder.

gr-fec/python/fec/polar/channel_construction.py
```diff
@@ -15,7 +15,7 @@
 from .channel_construction_bec import calculate_bec_channel_capacities
 from .channel_construction_bec import design_snr_to_bec_eta
 from .channel_construction_bec import bhattacharyya_bounds
-from .helper_functions import *
+from .helper_functions import np
```
daniestevez, line 18: I suggest the very common
```
import numpy as np
```
rather than importing `np` from `helper_functions` (which contains an `import numpy as np`).

This remark also applies to `gr-fec/python/fec/polar/testbed.py`

willcode:
@MarkBauer5 if your new PR supercedes this one, you can close.

PR #4667, opened 2021-05-25
Use Soapy instead of UHD sink/source blocks for flow graph testing.

drmpeg:


drmpeg:
Partially resolves https://github.com/gnuradio/gnuradio/issues/3534

PR #4670, opened 2021-05-26
gr-iio: Add ad9081 source/sink

Yamakaja:
This commit adds support for the AD9081 MxFE transceiver.

This device is special in that it is a direct sampling RF transceiver,
and completely based on NCOs, which are cascaded into coarse and fine
DDC / DUC. The datapath is dependent on the device
configuration and cannot be changed at runtime.

For more information see the [AD9081 MxFE Linux Driver](https://wiki.analog.com/resources/tools-software/linux-drivers/iio-mxfe/ad9081), [HDL reference design](https://wiki.analog.com/resources/eval/user-guides/ad9081_fmca_ebz/ad9081_fmca_ebz_hdl) and [User Guide](https://www.analog.com/media/en/technical-documentation/user-guides/ad9081-ad9082-ug-1578.pdf)


Yamakaja:
V2:
* Changed "DDC" to "DUC" in `gr-iio/grc/iio_ad9081_sink.block.yml`

adamhorden:
@Yamakaja Is this ready for review 🤓? I was about to merge it in into `feature/gr-iio` when I noticed the draft status been applied to the PR.

**Adam Horden**

willcode:
Some thoughts that apply more broadly to gr-iio, but got triggered by this PR:

Both UHD and Soapy define a top-level, generic device interface. Modular drivers then implement that interface. I think gr-iio would benefit from such a model. The way iio drivers are implemented for Linux might also be a good model, though I've just started to look at it.

There are advantages to both GR and to the driver developer (ADI here). Separating the development of drivers and framework means that new drivers (like this one) can be released independently from the GR maintenance cycle. GR does not have the resources or knowledge to carry all of the logic for all the 10's of ADI devices that might be used as a SDR, or the 100's of devices that might be useful in some way in GR. This should be implemented in separate modules of some sort, potentially under the GR umbrella.

What we don't want is for gr-iio to be a repository of top level blocks having nothing in common other than they use gr-iio. So, what is it that makes the drivers in gr-iio unique such that they are grouped together in a module?

For example, IIO provides discovery, transport independence, parameter control, etc. How can these features be made available to GR? Are the non-ADC/DAC products of use in GR, and how would their functions be exposed?

PR #4671, opened 2021-05-26
GRC: Cache time- and version-sensitive

marcusmueller:
This resolves issues where the GRC cache is invalid, because the system or individual YAML files changed.

Complete Cache will be discarded if current cache file

* has no version info
* version older than version in cacheversion.in (i.e. after an update)

Cached entry will be reloaded from yaml if

* missing
* cached-at property smaller than modification time of yaml file (half
second tolerance, due to inelegant floating point modification time
handling in python)

Also, a bit of debug logging.


skoslowski:
I like the idea, but this implementation breaks the ability to run grc from the source tree. Aside from that grc has already version info:

https://github.com/gnuradio/gnuradio/blob/44152d6c3fb19ab56ef6f839f63fbfb26650b46a/grc/core/Config.py#L24

I suggest to use that.

willcode:
Could parse from the result of:
```
from gnuradio import gr
gr.version()
```

willcode:
What happens when an older version that doesn't know about cache versions reads the cache with the additional dict layer?

marcusmueller:
> I like the idea, but this implementation breaks the ability to run grc from the source tree. Aside from that grc has already version info:
> 
> https://github.com/gnuradio/gnuradio/blob/44152d6c3fb19ab56ef6f839f63fbfb26650b46a/grc/core/Config.py#L24
> 
> I suggest to use that.

d'oh! yeah I'll use that.

cmake/Modules/GrPython.cmake
```diff
@@ -168,6 +168,7 @@ function(GR_PYTHON_INSTALL)
     install(
       FILES ${GR_PYTHON_INSTALL_FILES}
       DESTINATION ${GR_PYTHON_INSTALL_DESTINATION}
+      ${GR_PYTHON_INSTALL_UNPARSED_ARGUMENTS}
```
willcode, line 171: Do the cmake changes need to be in the same PR?

willcode:
When an older version opens a new cache file, it adds file entries at top level and leaves the entire "new" structure intact. So the file ends up being a new and old cache file all in one.

Since "version" and "cached-at" are not likely yml file names, I'd suggest doing away with the "cache-content" structure and putting them, and the files, all at the top leve.

That leaves the question of how to poison the cache for older versions. Without poisoning, behavior is the same as today going from new to old, and old to new behaves as intended. So maybe the answer is to backport this to 3.9 and say that the cache version detection only works for 3.9 and newer.

cmake/Modules/GrPython.cmake
marcusmueller: not really, but they were required in the previous iteration of it. Can extract it to a separate one.

marcusmueller:
@willcode yeah, I figured I can either nuke the cache the moment I find an unversioned cache, or live with the old-cache-lingers problem.

willcode:
I suppose a one-time name change for the cache file is a possibility (not one per version).

cmake/Modules/GrPython.cmake
willcode: That would be good. I patched my system for a test, expecting a python-only, quick install. Got a cmake run.

marcusmueller: @willcode #4717

willcode (approved):

willcode:
Tested bouncing between master, 3.9, and 3.8 (with minor mod). Works.

marcusmueller:
@willcode  how urgent do you think this is? I added @haakov out of a gut feeling to the list of reviewers, but if you want to have that in the upcoming releases, I don't know whether you really want to wait.

willcode:
I'll put this into next week's releases if it goes into master. Mentioned to @mormj already.

PR #4672, opened 2021-05-27
Output a message when a missing block is created.

xloem:
This PR is onto the maint-3.9 branch, since I have that cloned locally.  Not certain what's normal here.

Since the `new_block` function absorbs errors, missing blocks cause errors that hide the underlying issue and what the wrong key is.  I ran into this when I had failed to rename a block everywhere.

I'm a little worried that this message I added can't be caught by code like an exception would be, but I saw another message was output to handle an issue elsewhere in the file, so copied the pattern.

### DIFF
--- a/grc/gui/canvas/flowgraph.py
+++ b/grc/gui/canvas/flowgraph.py
@@ -159,6 +159,9 @@ class FlowGraph(CoreFlowgraph, Drawable):
         )
         # get the new block
         block = self.new_block(key)
+        if block is None:
+            Messages.send('>>> No block found with key "' + key + '".\n')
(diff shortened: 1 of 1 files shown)


willcode:
@xloem PRs are typically against the "master" branch.

Can you give an examle of where this applies? The GRC GUI shows errors in red if a block is missing and refuses to generate code.

xloem:
I'm actually not sure why this problem is triggering for me, I'm still narrowing down the issue =S  Maybe I can make a PR to master with something more appropriate if I figure it out.  Thanks for your reply!

PR #4674, opened 2021-05-27
Propagate to the user, errors making new grc blocks.

xloem:
Otherwise, when the None return value is used later, the user sees a misleading error.

The misleading error is thrown here: https://github.com/gnuradio/gnuradio/blob/master/grc/gui/canvas/flowgraph.py#L162

This the same issue as #4672 .

The problem was that I had duplicate option ids specified for an enum parameter.  I wanted to handle signed and unsigned int64s differently, but was using gnuradio's types as id names, and gnuradio doesn't differentiate between them.

### DIFF
--- a/grc/core/FlowGraph.py
+++ b/grc/core/FlowGraph.py
@@ -308,11 +308,8 @@ class FlowGraph(Element):
         """
         if block_id == 'options':
             return self.options_block
-        try:
-            block = self.parent_platform.make_block(self, block_id, **kwargs)
-            self.blocks.append(block)
-        except KeyError:
-            block = None
+        block = self.parent_platform.make_block(self, block_id, **kwargs)
+        self.blocks.append(block)
         return block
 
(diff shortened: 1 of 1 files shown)


mormj:
@xloem - this appears to be failing the `grc_tests.py` in CI.  Haven't looked into exactly why, but this will need to be addressed - not sure if this is causing or exposing grcc compilation errors

xloem:
I looked into the test failures.  It looks to me like they are because grc tests use blocks from gr-blocks (example: https://github.com/gnuradio/gnuradio/blob/master/grc/tests/resources/test_compiler.grc#L85 ), and the gr-blocks block path is not included when the test is run.  I'm assuming the error is thrown instead of absorbed because of my change, since it is the same kind of error I was dealing with.

xloem:
It looks like my change broke existing code that assumes failure is held by returning None.  This PR is incorrect.

Issue #4675, opened 2021-05-27
No Data Output QT GUI Vector Sink


xloem:
I'm not getting any actual display data with the QT GUI Vector Sink.  Autoscale is working to adjust the axes, but there's no line on the graph.  ~~Is there a working example that uses this block?~~

Contrariwise, QT GUI Frequency Sink is working fine for me, and the source code seems very similar.

EDIT: I found the example and it draws a line.  Hopefully this is just a configuration issue for me.

xloem:
Looks like my data is just bad.  Sorry for noise!

PR #4676, opened 2021-05-27
Fix qt gui vector margin for non-frequency data

xloem:
This is why I was running into #4675 .  My data was on the range 0.0-0.01 and was invisible with a total hardcoded margin of 20.

It looks like the margin of 10 on either side might have been unreviewed code copied from the gui frequency sink.  This change updates the margin to be 1% of the range of the data.


willcode (approved):
Looks right. The previous code must have been from something measured in dBs.

PR #4683, opened 2021-05-28
grc: Be more informative when missing "options" block.

ryanvolz:
While we mull over #4502 or a better fix that has yet to appear, I thought it would be good to pull out this commit and submit it separately.

This raises an RuntimeError describing that GRC could not find the built-in blocks rather than raising a KeyError on "options", which left users with little to go on. This error has come up in various situations involving an incorrect blocks path (due to mixing GR versions causing the configuration files to point to the wrong installation) or simply the blocks not being installed.

### DIFF
--- a/grc/core/platform.py
+++ b/grc/core/platform.py
@@ -179,7 +179,20 @@ class Platform(Element):
 
         self._docstring_extractor.finish()
         # self._docstring_extractor.wait()
-        utils.hide_bokeh_gui_options_if_not_installed(self.blocks['options'])
+        if 'options' not in self.blocks:
+            # we didn't find one of the built-in blocks ("options")
+            # which probably means the GRC blocks path is bad
+            errstr = (
+                "Failed to find built-in GRC blocks (specifically, the "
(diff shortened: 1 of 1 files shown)


mormj (approved):
LGTM!

grc/core/platform.py
```diff
@@ -179,7 +179,20 @@ def build_library(self, path=None):
...
+            # we didn't find one of the built-in blocks ("options")
+            # which probably means the GRC blocks path is bad
+            errstr = (
+                "Failed to find built-in GRC blocks (specifically, the "
+                "'options' block). Ensure your GRC block paths are correct "
+                "and at least one points to your prefix installation:"
+            )
+            errstr = "\n".join([errstr] + (path or self.config.block_paths))
+            raise RuntimeError(errstr)
+        else:
```
skoslowski, line 192: nit: don't really need this...

skoslowski (approved):
I always wanted to convert the options block into a built-in block class (or keep the yaml as package data) but never got around to it.
This is nice addition.

PR #4684, opened 2021-05-28
gr-iio: Add packetized transfer support using stream tags

Yamakaja:
This PR adds stream tagging support to the device_source and device_sink
blocks, and thus also deriviates like the fmcomms2 and pluto.

The important bits are:

* The source block now tags buffer boundaries
* When tagged_input is enabled on the device sink, it enforces that all input packet sizes match buffer_size / (1 + interpolation). This isn't very flexible, but can be understood as a safeguard against unexpected behavior.

This PR is independent of https://github.com/gnuradio/gnuradio/pull/4670, but depending on which PR is merged first, the remaining one will need an update.

(Sorry for the yaml reformatting ;) )

### DIFF
--- a/gr-iio/include/gnuradio/iio/device_sink.h
+++ b/gr-iio/include/gnuradio/iio/device_sink.h
@@ -14,6 +14,8 @@
 #include <gnuradio/iio/api.h>
 #include <gnuradio/sync_block.h>
 
+#include <string>
+
 #define DEFAULT_BUFFER_SIZE 0x8000
 
 extern "C" {
@@ -71,6 +73,18 @@ public:
                           unsigned int buffer_size = DEFAULT_BUFFER_SIZE,
                           unsigned int interpolation = 0,
                           bool cyclic = false);
+
+    /*!
+     * The key of the tag that indicates packet length.
+     * When not empty, samlpes are expected as "packets" and
+     * must be tagged as such, i.e. the first sample of a packet needs to be
+     * tagged with the corresponding length of that packet.
+     * Note, packet size MUST be equal to buffer_size / (1+interpolation),
+     * otherwise a runtime_error will be thrown. This is a deliberate design
+     * choice, because all other options would result in potentially unexpected
+     * behavior.
+     */
+    virtual void set_len_tag_key(const std::string& len_tag_key) = 0;
 };
 
 } // namespace iio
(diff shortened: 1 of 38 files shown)


gr-iio/include/gnuradio/iio/device_sink.h
```diff
@@ -61,7 +71,9 @@ class IIO_API device_sink : virtual public gr::sync_block
                      const std::vector<std::string>& params,
                      unsigned int buffer_size = DEFAULT_BUFFER_SIZE,
                      unsigned int interpolation = 0,
-                     bool cyclic = false);
+                     bool cyclic = false,
```
willcode, line 74: Wherever possible, use setters and minimize the number of arguments in the constructor. Here, for example, if tagged input is to be used, the developer needs to look up "cyclic" even if the default value is desired. Some setters can only be used right after initialization, so they don't get GRC callbacks. Maybe they log warnings if used while the flowgraph is running. That only applies to "cyclic" here, maybe others that already use setters.

The alternative is to use a param list of some sort, so that missing params are ok. Multiple constructor signatures work in some cases, but they can be confusing.

gr-iio/include/gnuradio/iio/device_sink.h
```diff
@@ -61,7 +71,9 @@ class IIO_API device_sink : virtual public gr::sync_block
...
+                     bool tagged_input = false,
```
willcode, line 75: Can tagged_input be implied by a non-blank len_tag_key? This would save a param/setter.

Yamakaja: I thought about that, but it would prevent us from providing a default tag name. Though i suppose that's not too much of an issue.

gr-iio/include/gnuradio/iio/device_sink.h
Yamakaja: So basically a one-time use setter? Wouldn't that be equally applicable to all those other parameters then?

willcode: Yes. Having to specify interpolation because you want to use cyclic is equally confusing, and buffer_size more so.

Yamakaja:
V2:

* Moved length tag initialization out of constructor / make

Issue #2672, opened 2019-07-27
gr_modtool rm deletes too much


ast:
I had two modules `stream_to_vector_decimator` and `stream_to_vector_decimator2`

`gr_modtool rm stream_to_vector_decimator`

removed both of them.

solomonbstoner:
I am able to reproduce this behavior with similarly named blocks. I will try to debug this

solomonbstoner:
Just to illustrate one of the cause(s) of the problem, let's look at the `rm` command. The `stream_to_vector_decimator2` and `stream_to_vector_decimator` are both deleted because the function `_run_subdir` finds all the files in the path of interest, and offers to delete files where the user entered name appears. So entering  "stream_to_vector_decimator" would match both "stream_to_vector_decimator" and "stream_to_vector_decimator2". That's why files from both blocks were offered to be deleted. 

```python
    def _run_subdir(self, path, globs, makefile_vars, cmakeedit_func=None):
        """ Delete all files that match a certain pattern in path.
        path - The directory in which this will take place
        globs - A tuple of standard UNIX globs of files to delete (e.g. *.yml)
        makefile_vars - A tuple with a list of CMakeLists.txt-variables which
                        may contain references to the globbed files
        cmakeedit_func - If the CMakeLists.txt needs special editing, use this
        """
        if self.cli:
            from ..cli import cli_input
        # 1. Create a filtered list
        files = []
        for g in globs:
            files = files + sorted(glob.glob(f"{path}/{g}"))
        files_filt = []
        logger.info(f"Searching for matching files in {path}/:")
        for f in files:
```
Now, `gr_modtool` will ask you to confirm if you want to delete the files, so a little careful attention could avoid unnecessary deletion. But I suppose we could still improve on the matching nonetheless

solomonbstoner:
@mormj assign this issue to me ill try to work something out

PR #4686, opened 2021-05-29
modtool: Fix `TypeError`

### LINKED ISSUE #4281: gr_modtool rm does not remove ref in python_bindings.cc
Hi,
It seems that gr_modtool rm does not remove ref of the deleted block in python_bindings.cc.
73 FM4PN

### LINKED ISSUE #4682: gr_modtool: TypeError when using regex
When running `gr_modtool bind` on a block named `decoder`, the following `TypeError` error is thrown. This error is not thrown when the command is run as `gr_modtool bind decoder`.

```
$ gr_modtool bind
GNU Radio module name identified: rds
Which blocks do you want to parse? (Regex): decoder
```
```
 File "/usr/local/lib/python3/dist-packages/gnuradio/modtool/cli/bind.py", line 46, in cli
    run(self)
  File "/usr/local/lib/python3/dist-packages/gnuradio/modtool/cli/base.py", line 152, in run
    module.run()
  File "/usr/local/lib/python3/dist-packages/gnuradio/modtool/core/bind.py", line 62, in run
    file_to_process = os.path.join(self.dir, self.info['includedir'], self.info['blockname'] + '.h')
TypeError: unsupported operand type(s) for +: 'NoneType' and 'str'
```

Im working on this on my setup. Could someone confirm if this is an issue?

solomonbstoner:
Fixes #4682 and fixes #4281, a bug triggered in the `rm` and `bind` commands when the
block name (or regex) is entered as stdin rather than an argument.

### DIFF
--- a/gr-utils/modtool/core/bind.py
+++ b/gr-utils/modtool/core/bind.py
@@ -22,7 +22,7 @@ except ImportError:
 else:
     have_bindtool = True
 
-from ..tools import ParserCCBlock, CMakeFileEditor, ask_yes_no
+from ..tools import ParserCCBlock, CMakeFileEditor, ask_yes_no, get_block_names
 from .base import ModTool, ModToolException
 
 from gnuradio import gr
@@ -58,9 +58,18 @@ class ModToolGenBindings(ModTool):
 
         with warnings.catch_warnings():
             warnings.filterwarnings("ignore", category=DeprecationWarning)
-            file_to_process = os.path.join(self.dir, self.info['includedir'], self.info['blockname'] + '.h')
-
+            blocknames_to_process = []
+            if self.info['blockname']:
+                # A complete block name was given
+                blocknames_to_process.append(self.info['blockname'])
+            elif self.info['pattern']:
+                # A regex resembling one or several blocks were given
+                blocknames_to_process = get_block_names(self.info['pattern'], self.info['modname'])
+            else:
+                raise ModToolException("No block name or regex was specified!")
             bg = BindingGenerator(prefix=gr.prefix(), namespace=[
                                   'gr', self.info['modname']], prefix_include_root=self.info['modname'], output_dir=os.path.join(self.dir, 'python'),
                                    define_symbols=self.info['define_symbols'], addl_includes=self.info['addl_includes'])
-            bg.gen_file_binding(file_to_process)
+            files_to_process = [os.path.join(self.dir, self.info['includedir'], f'{blockname}.h') for blockname in blocknames_to_process]
+            for file_to_process in files_to_process:
+                bg.gen_file_binding(file_to_process)
--- a/gr-utils/modtool/core/rm.py
+++ b/gr-utils/modtool/core/rm.py
@@ -15,7 +15,7 @@ import sys
 import glob
 import logging
 
-from ..tools import remove_pattern_from_file, CMakeFileEditor, CPPFileEditor
+from ..tools import remove_pattern_from_file, CMakeFileEditor, CPPFileEditor, get_block_names
 from .base import ModTool, ModToolException
 
 logger = logging.getLogger(__name__)
@@ -88,11 +88,6 @@ class ModToolRemove(ModTool):
             ed.remove_double_newlines()
 
         # Go, go, go!
-        if not self.skip_subdirs['lib']:
-            self._run_subdir('lib', ('*.cc', '*.h'), ('add_library', 'list'),
-                             cmakeedit_func=_remove_cc_test_case)
-        if not self.skip_subdirs['include']:
-            incl_files_deleted = self._run_subdir(self.info['includedir'], ('*.h',), ('install',))
         if not self.skip_subdirs['python']:
             py_files_deleted = self._run_subdir('python', ('*.py',), ('GR_PYTHON_INSTALL',),
                                                 cmakeedit_func=_remove_py_test_case)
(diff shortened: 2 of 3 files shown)


solomonbstoner:
Just to explain the change in order of deleting/amending the block files in `rm.py`:

Prior to this PR, the `include/module_name/CMakeLists.txt` file for the blocks' header files are amended prior to the deletion of the entries in the python binding files (which triggers the bug). This PR reverses the order of the process as my `get_block_names` function depends on said CMakeLists.txt file to match a regex pattern to the given blocks. Thus, the entries are deleted first. This fixes issues #4682 and #4281.

solomonbstoner:
@mormj for your review, please :)

mormj:
This PR seems to `rm` the blocks more consistently now, since previously blocks only c++ blocks that started with the pattern were removed.  Now python blocks seem to as well.

I'm wonder though about the behavior as it relates to: https://github.com/gnuradio/gnuradio/issues/2672

This doesn't seem to be desirable behavior if someone has similarly named blocks in their OOT.

solomonbstoner:
Actually, this PR fixes the problem of deleting c++ blocks while not removing the bindings. It doesnt affect how Python blocks are deleted. It is not related to #2672 either as that issue stems from a different function. Ill post more information in that issue

mormj:
Yes, this seems to work now, and clean up the things that were left behind in the bindings.  LGTM!

mormj (approved):

solomonbstoner:
Update: This force-push removed the thrown error for when there are no blocknames returned by `get_block_names`. 
`get_block_names` in this PR is for c++ files so there will not be a match for python files, so it should not throw an error on purpose

solomonbstoner:
> This PR seems to `rm` the blocks more consistently now, since previously blocks only c++ blocks that started with the pattern were removed. Now python blocks seem to as well.
> 
> I'm wonder though about the behavior as it relates to: #2672
> 
> This doesn't seem to be desirable behavior if someone has similarly named blocks in their OOT.

Issue #2672 addressed by PR #4709

solomonbstoner:
@swap-nil7 could you review, please?

gr-utils/modtool/core/bind.py
```diff
@@ -58,9 +58,18 @@ def run(self):
...
+            elif self.info['pattern']:
+                # A regex resembling one or several blocks were given
+                blocknames_to_process = get_block_names(self.info['pattern'], self.info['modname'])
+            else:
+                raise ModToolException("No block name or regex was specified!")
             bg = BindingGenerator(prefix=gr.prefix(), namespace=[
                                   'gr', self.info['modname']], prefix_include_root=self.info['modname'], output_dir=os.path.join(self.dir, 'python'),
                                    define_symbols=self.info['define_symbols'], addl_includes=self.info['addl_includes'])
-            bg.gen_file_binding(file_to_process)
+            files_to_process = list(map(lambda blockname:os.path.join(self.dir, self.info['includedir'], blockname + '.h'), blocknames_to_process))
```
swap-nil7, line 73: Use list comprehension instead, `[os.path.join(self.dir, self.info['includedir'], f'{blockname}.h') for blockname in blocknames_to_process]`
or `[f"{self.dir}/{self.info['includedir']}/{blockname}.h" for for blockname  in blocknames_to_process]`

gr-utils/modtool/core/rm.py
```diff
@@ -105,13 +100,28 @@ def _remove_py_test_case(filename=None, ed=None):
...
-            ed.remove_value('// BINDING_FUNCTION_CALLS(', '// ) END BINDING_FUNCTION_CALLS', 
-                'bind_' + self.info['blockname'] + '(m);')
-            ed.write()
+            blocknames_to_delete = []
+            if self.info['blockname']:
+                # A complete block name was given
+                blocknames_to_delete.append(self.info['blockname'])
+            elif self.info['pattern']:
+                # A regex resembling one or several blocks were given
+                blocknames_to_delete = get_block_names(self.info['pattern'], self.info['modname'])
```
swap-nil7, line 109: Are we deleting all blocks matching a regex, this is too unsafe.

swap-nil7 (changes requested):

gr-utils/modtool/core/rm.py
solomonbstoner: this was the original feature of gr_modtool. Any argument given as stdin is taken as a regex and deletes all those blocks. I merely kept it in

gr-utils/modtool/core/bind.py
solomonbstoner: Will make the change

gr-utils/modtool/core/rm.py
solomonbstoner: Im not sure if this is the actual intention of gr_modtool, but it seems to me that when a block name is given as an argument (e.g. `gr_modtool rm my_block`), it means the user wants that specific block to be acted on. On the other hand, when a regex is given as a stdin (afterall the prompt is for a regex), then the user wants a few blocks matching the regex to be acted on. So I didnt want to change the behavior of the modtool as it is.

However, I noticed that modtool does not behave exactly as its supposed to, so Ive made a fix in PR #4709. The issue and solution are described in greater detail there. Could I trouble you to review that too, please?

solomonbstoner:
@swap-nil7 switched to list comprehension as requested

gr-utils/modtool/core/rm.py
swap-nil7: Yeah, got your point, I think let's give the prompt for actual blockname, we can show the probable candidates if user enters incorrectly. Deletion based on regex can basically remove everything due to a single mistake, or do we notify before for confirmation, I forgot the behaviour?

solomonbstoner: Yep. There is a confirmation before it actually deletes "Really delete include/mod_name/block_name.h? [Y/n/a/q]: "

swap-nil7: Cool, then

swap-nil7 (approved):

PR #4687, opened 2021-05-29
digital: Add NRZI blocks

daniestevez:
```
This adds NRZI encoder and decoder blocks. The blocks are similar
to a Differential encoder / decoder with a modulus of 2, but there
is an extra inversion. Having dedicated NRZI blocks saves the user
from having to add a Not block to handle the inversion. NRZI is
used in AX.25 and other protocols as a differential coding method.

The NRZI blocks have been taken and adapted from the gr-satellites
OOT as described in #3637.

Signed-off-by: Daniel Estévez <daniel@destevez.net>
```

**Note:** I have kept my name in the copyright statements in the header of the new files. I've browsed recent commits to see if there was a consistent approach to this after the exit from the FSF, but I didn't arrive to any conclusion. If required, I can change the copyright statements.

### DIFF
--- a/gr-digital/grc/digital_diff_decoder_bb.block.yml
+++ b/gr-digital/grc/digital_diff_decoder_bb.block.yml
@@ -3,9 +3,19 @@ label: Differential Decoder
 flags: [ python, cpp ]
 
 parameters:
+-   id: coding
+    label: Coding
+    dtype: enum
+    default: digital.DIFF_DIFFERENTIAL
+    options: [digital.DIFF_DIFFERENTIAL, digital.DIFF_NRZI]
+    option_labels: [Differential, NRZI]
+    option_attributes:
+        hide_modulus: [none, all]
+        force_modulus: [-1, 2]
 -   id: modulus
     label: Modulus
     dtype: int
+    hide: ${ coding.hide_modulus }
 
 inputs:
 -   domain: stream
@@ -17,13 +27,15 @@ outputs:
 
 templates:
     imports: from gnuradio import digital
-    make: digital.diff_decoder_bb(${modulus})
+    make: digital.diff_decoder_bb(${modulus if coding.force_modulus == '-1' else coding.force_modulus}, ${coding})
 
 cpp_templates:
     includes: ['#include <gnuradio/digital/diff_decoder_bb.h>']
     declarations: 'digital::diff_decoder_bb::sptr ${id};'
     make: |-
-        this->${id} = digital::diff_decoder_bb::make(${modulus});
+        this->${id} = digital::diff_decoder_bb::make(${modulus if coding.force_modulus == '-1' else coding.force_modulus}, ${coding});
     link: ['gnuradio-digital']
+    translations:
+        digital.DIFF: 'digital::DIFF'
 
 file_format: 1
--- a/gr-digital/grc/digital_diff_encoder_bb.block.yml
+++ b/gr-digital/grc/digital_diff_encoder_bb.block.yml
@@ -3,9 +3,19 @@ label: Differential Encoder
 flags: [ python, cpp ]
 
 parameters:
+-   id: coding
+    label: Coding
+    dtype: enum
+    default: digital.DIFF_DIFFERENTIAL
+    options: [digital.DIFF_DIFFERENTIAL, digital.DIFF_NRZI]
+    option_labels: [Differential, NRZI]
+    option_attributes:
+        hide_modulus: [none, all]
+        force_modulus: [-1, 2]
 -   id: modulus
     label: Modulus
     dtype: int
+    hide: ${ coding.hide_modulus }
 
 inputs:
 -   domain: stream
@@ -17,13 +27,15 @@ outputs:
 
 templates:
     imports: from gnuradio import digital
-    make: digital.diff_encoder_bb(${modulus})
+    make: digital.diff_encoder_bb(${modulus if coding.force_modulus == '-1' else coding.force_modulus}, ${coding})
 
 cpp_templates:
     includes: ['#include <gnuradio/digital/diff_encoder_bb.h>']
     declarations: 'digital::diff_encoder_bb::sptr ${id};'
     make: |-
-        this->${id} = digital::diff_encoder_bb::make(${modulus});
+        this->${id} = digital::diff_encoder_bb::make(${modulus if coding.force_modulus == '-1' else coding.force_modulus}, ${coding});
     link: ['gnuradio-digital']
+    translations:
+        digital.DIFF: 'digital::DIFF'
 
 file_format: 1
--- a/gr-digital/include/gnuradio/digital/CMakeLists.txt
+++ b/gr-digital/include/gnuradio/digital/CMakeLists.txt
@@ -37,6 +37,7 @@ install(FILES
     crc32_async_bb.h
     decision_feedback_equalizer.h
     descrambler_bb.h
+    diff_coding_type.h
     diff_decoder_bb.h
     diff_encoder_bb.h
     diff_phasor_cc.h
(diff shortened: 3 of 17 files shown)


daniestevez:
It seems the pipeline is failing on some unrelated flaky tests:

```
FAIL: test_fft_filter_ccc_001 (__main__.test_freq_xlating_filter)
  File "/__w/gnuradio/gnuradio/gr-filter/python/filter/qa_freq_xlating_fft_filter.py", line 127, in test_fft_filter_ccc_001
    self.assert_fft_ok(expected_data, result_data)
  File "/__w/gnuradio/gnuradio/gr-filter/python/filter/qa_freq_xlating_fft_filter.py", line 76, in assert_fft_ok
    abs_eps=1e-9, rel_eps=1.5e-3)
  File "/__w/gnuradio/gnuradio/gnuradio-runtime/python/gnuradio/gr_unittest.py", line 89, in assertComplexTuplesAlmostEqual2
    for (x, y) in zip(a, b)
  File "/__w/gnuradio/gnuradio/gnuradio-runtime/python/gnuradio/gr_unittest.py", line 89, in <listcomp>
    for (x, y) in zip(a, b)
  File "/__w/gnuradio/gnuradio/gnuradio-runtime/python/gnuradio/gr_unittest.py", line 56, in assertComplexAlmostEqual2
    ref, x, abs(ref-x) / abs(ref), rel_eps
AssertionError: (-1.5216092136337213e-05+1.6304929656726734e-05j) != (-1.5195104424492456e-05+1.633969441172667e-05j) rel_error = 0.001820855984780177 rel_limit = 0.0015
```

Maybe the tolerance should be increased in these tests?

willcode:
Should NRZI be added to the existing diff_encoder and diff_decoder as an option? Then we can add Manchester without adding another block?

daniestevez:
Hi @willcode, in fact that's something that I was wondering as I was adding these blocks, since they are very similar to the existing `diff_*` and all these blocks are almost completely boilerplate code (the important calculation is just one line).

However, I have mixed feelings about this. The problem is that NRZI coding only makes sense with a modulus of 2, so the best way to combine NRZI with the differential blocks is not clear to me. We might just `raise` an error if the user requests NRZI and a modulus different than 2. Then the question is how to handle this in GRC, but we might have different GRC blocks which use the same C++ block behind the scenes.

Another point (perhaps not so important) is about performance. The `% d_modulus` part in the current differential blocks makes sense for a general modulus, but for the case of modulus 2 it might be written more efficiently (something like `out[i] = in[i] ^ last_out`, for instance). In fact, since modulus 2 is such a common case of the differential blocks, I wonder if it's worth to include this as a special case by checking `if (modulus == 2)` at the start of the work function and then going to a `for` loop that contains the best operation for that case. We could do the same for NRZI coding and have three different `for` loops.

However, if we're talking about performance (specially for the modulus of 2), then perhaps the best solution is to convert these loops into optimized volk kernels.

Regarding Manchester, I think that's a rather different case, since in the encoder you are upsampling by a factor of 2 (say you transform `0` into `10` and `1` into `01`), so the block wouldn't be a `sync_block`. In Manchester decoding you are decimating by 2 but you also need to be aligned correctly, so it makes sense for a Manchester decoder block to search for the alignment and also do the decoding itself (I have something like that as a Python block in gr-satellites, in fact).

willcode:
In GRC, the modulus field can be hidden and set to 2 if certain encodings are selected. Even if there have to be different `for` loop in the impl, it seems better than having a proliferation of blocks. Plus, it gives the user the opportunity to try different encodings without changing the graph.

If Manchester (other clocked encodings?) were added, the the blocks would be interp/decim blocks. Good point that self-sync would be a good feature, so chunked input/output would not be useful.

daniestevez:
I'll leave a few days in case anyone wants to comment on the best approach, and then I'll move the NRZI code to a separate `for` loop in the `work()` function of the differential blocks.

willcode:
That sounds good.

mbr0wn:
> **Note:** I have kept my name in the copyright statements in the header of the new files. I've browsed recent commits to see if there was a consistent approach to this after the exit from the FSF, but I didn't arrive to any conclusion. If required, I can change the copyright statements.

This is the correct thing to do. You won't find lots of files like that, because the move-from-CLA was pretty recent.

daniestevez:
I have now included the NRZI as an option of the differential encoder/decoder blocks. Additionally, I have written specific code for the `modulus = 2` case that is more efficient than using `%`.

I was considering merging the NRZI QA code I had from gr-satellites with the QA differential encoder in GNU Radio. However, I've decided to keep it separate, since the approach is quite different (the NRZI QA uses Numpy to replicate the functionality of the blocks and check against the C++ implementation). While evaluating this, I've seen that the differential encoder QA has a lot of unnecessary repetition, so I've fixed that in a different commit in this PR. Is that OK or should I send it on a different PR?

Finally, this PR is missing GRC blocks for the NRZI options. I'd like some feedback on the best way to proceed. Do we want different GRC blocks for NRZI (even if the same C++ block is used behind the scences) or do we want to re-use the differential encoder/decoder GRC blocks? If the latter, how to present the parameters to the user, since NRZI implies modulus = 2? Maybe a drop down to select "mode" between "differential" and "NRZI" and only show the "modulus" field when "differential" is selected?

willcode:
This looks pretty good.

Suggest renaming qa_nrzi to qa_diff_encoder_nrzi so it will be easier to find. Merge the two in the future?

Your choice: if we were to grow the encoder further later, it would be better to have an enum rather than a bool. I don't have an example that would require it though. Eventually, something where you can "explore" encodings when you're analyzing a signal would be cool. Maybe that's a new block that takes over for this one in the future, like Clock Recovery did for a bunch of individual blocks.

I like reusing the current GRC blocks, adding an option that defaults to the current behavior, and having a `hide` for modulus if NRZI.

Thanks!

daniestevez:
I've just pushed some new commits (when this is ready to merge I'll squash everything).

I've added the GRC blocks as agreed. The existing differential encoder / decoder blocks are re-used and now there is a drop-down option to select differential coding or NRZI coding. When NRZI is selected, the modulus field is hidden.

In the end I've also implemented @willcode's suggestion of using an enum to indicate the coding (this is include in a separate commit).

I think that this PR is ready for review. I've tested that the Python and C++ code generation in GRC work correctly, since the GRC blocks are a bit tricky.

willcode (approved):
Minor - is the `DIFF_` prefix necessary on the constants?

I see the answer to my question in #4754. Fixing our tooling would be better than writing code to get around it if that's possible. @skoslowski do you know what should be done here? Should the translation be something like
`'digital\.': 'digital::'`
instead of
`digital.: 'digital::'`

Let's hold off on this PR to see if there's a quick way to fix the code generator if that's what is needed.

daniestevez:
> Minor - is the DIFF_ prefix necessary on the constants?

I used the Symbol Sync block as a reference when implementing this, and saw that it used `TED_` to define its enum constants. I think that a good reason to do so is that these constants end up both in the `gr::digital` C++ namespace and in the `digital` Python module, so it's better to use a prefix to avoid polluting the namespace. However, I haven't given this any thought about whether there is another way to do it that doesn't place the constants directly in those namespaces (I always get a bit confused about the differences with enums in C and C++).

Regarding #4754, I already tried `\.` as you suggest, and it didn't seem to work. I didn't fiddle much to see if there was another way to escape the `.`.

daniestevez:
Thanks! I didn't have a strong opinion on holding this up or not. I guess that when we find a good way to solve #4754 we will need to search and fix all the affected blocks, including the diff encoder/decoder.

daniestevez:
Sorry, I forgot to squash this yesterday. I've now squashed all the commits, except the clean-up of the current QA test, which I've kept separately, since it can be treated as a totally independent thing.

mormj (approved):

mormj:
> Changing constants gives us days of fun at the help desk, even in API releases smile so it probably won't come up. @mormj were you thinking of any general fixup of constant/enum names for 3.10?

I wouldn't change the constant/enum names unless there is a strong argument to do so - you're right it creates all kinds of unforeseen problems even though it is the "right" thing to do.  3.10 is not such a departure from 3.9 that we shouldn't continue to maximize compatibility.  4.0 we should do all that the "right way"

PR #4706, opened 2021-06-03
v3.8.3.1-rc1

willcode:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,52 @@ starting with version 3.7.12.0.
 
 Older Logs can be found in `docs/RELEASE-NOTES-*`.
 
+## [3.8.3.1] - 2021-06-02
+
+This is a PATCH level revision. The API is compatible with C++ code written for previous v3.8 releases. ABI (shared library signature) is intended to be compatible, so code linked against v3.8.3.0 should not require recompilation.
+
+### Changed
+
+#### Build system
+
+- Improved messages related to dependencies MPIR and GMP. Only one of these packages is required, and the previous error messages were confusing.
+
+#### GRC
+
+- Parameter expressions and/or values can be displayed in blocks on the flowgraph. Previously, only values were displayed. Look for the "Show Parameter ..." toggles under the View menu.
+- Vector length is now correctly applied to all input ports.
+- Validation has been improved. Raw types are validated. Port connections are checked by type rather than by item size.
+- Variable names that cause conflicts in Python code (e.g., package names) are rejected.
+- Bus logic fixes.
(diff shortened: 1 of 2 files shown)


velichkov:
Hi Jeff,

> The volk submodule has not been checked out in the github-generated tar and zip files. Use
git submodule update --init

Would it be possible to release a tarball that contains the `volk` sub-module? A few weeks ago I've requested upgrade of gnuradio to 3.8.3-0 in Fedora 33 and one of the problems that prevent/delay the update is that volk is not included in the tarball. 

See https://bugzilla.redhat.com/show_bug.cgi?id=1956356#c3 and https://bugzilla.redhat.com/show_bug.cgi?id=1956356#c6

willcode:
Yes, I'll definitely post v3.8.3.1 with a `-with-volk` variant, as I did (retroactively) for v3.8.3.0. Will that work with your release procedures?

velichkov:
> Will that work with your release procedures? 

I hope so. 

@yarda  Do you have any other requirments?

> This will be out next week.

OK.

> Do you need this for the rc too?

No.

yarda:
> @yarda Do you have any other requirments?
> 
Thanks, at the moment I am not aware of anything else.

willcode:
@velichkov @yarda the volk submodule has been set to v2.0.0 for the last couple of years, so that's what you've been getting in the tar files. It is reported that the latest versions of volk work with gnuradio 3.8, are required for some architectures (armv7). I ran the QA tests with gnuradio 3.8 and volk 2.4.1 with no problems, using
`-DENABLE_VOLK=ON -DENABLE_INTERNAL_VOLK=OFF`
as part of the gnuradio cmake.

I don't know of any reason you need to use an upgraded/external volk. There are probably performance improvements and bugfixes.

willcode:
Johannes Demel advises that the submodule was left at v2.0.0 on purpose, at the time at least.

velichkov:
Hi Jeff,

I agree that it will be good `volk` to be upgraded to its latest stable version as well. `volk` has some problems on `s390x` and `ppc64le` architectures that are still not resolved but the very same problems exists in the version that's currently included in Fedora 33.

willcode:
I'll make a gnuradio-with-volk-2.5.0.tar.gz and gnuradio-with-volk-2.0.0.tar.gz so it's obvious. Then people can take their pick. I don't know how to decide for them. My own quick tests show that volk 2.5.0 should be fine, at least for gnuradio in-tree stuff. If I forget - remind me!

It could be useful for packagers to do the volk checkout themselves in the future, since they have to do this already for newer versions of gnuradio. But I don't mind adding a couple files to the releases if it helps people.

willcode:
Done.

PR #4709, opened 2021-06-04
modtool: Fix #2672 for `rm`, `disable`, `rename` and `makeyml`

### LINKED ISSUE #2672: gr_modtool rm deletes too much

solomonbstoner:
Fixes #2672 

When a block name is entered as an argument to a modtool command, (e.g. `gr_modtool rm my_block`), the user expects _only_ that specific block (ie `my_block`) to be acted upon. Issue #2672 is caused when similarly named block names are also acted upon (e.g. `my_block_1` or `temp_my_block`). This issue affects the four commands stated in the title. PR fixes said issue.

Edit:
PR also fixes two other seemingly unreported issues:
1. A minor AttributeError affecting the `disable` command OOT version 38 due to what I believe is a spelling error: `self._info` should be `self.info`. 
2. Failure of the `rename` command to rename the binding function in `python_bindings.cc`.

### DIFF
--- a/gr-utils/modtool/core/disable.py
+++ b/gr-utils/modtool/core/disable.py
@@ -67,7 +67,7 @@ class ModToolDisable(ModTool):
                 ed.write()
                 self.scm.mark_file_updated(self._file['qalib'])
             elif self.info['version'] == '38':
-                fname_qa_cc = f'qa_{self._info["blockname"]}.cc'
+                fname_qa_cc = f'qa_{self.info["blockname"]}.cc'
                 cmake.comment_out_lines(fname_qa_cc)
             elif self.info['version'] == '36':
                 cmake.comment_out_lines('add_executable.*'+fname)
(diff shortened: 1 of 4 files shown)


solomonbstoner:
To illustrate the changes my PR does, let me show an example using `gr_modtool rename decoder` in the `gr-rds` (chose arbitrarily) OOT module.

I introduced two other modules using `gr_modtool add` to replicate a similarly named block.
1. `adecoder` 
2. `decoder2` 

Running `gr_modtool rename decoder` before my PR would give us the following changes. Notice that all three blocks with "decoder" in it are modified.

before_my_patch

After my PR, only the block specifically named `decoder` is modified. The other two are not.

solomonbstoner:
@mormj for your review please :)

mormj:
Haven't gotten a chance yet to thoroughly test it but the code looks good!

mormj (approved):
Tested and this does indeed fix the original issue

Issue #2688, opened 2019-07-30
grc: fails to build simple c++ based flowgraph


dl1ksv:
Trying to build a simple flowgraph in c++ [test_cpp](https://github.com/gnuradio/gnuradio/blob/master/grc/tests/resources/test_cpp.grc)

fails with:
```
 Executing: cmake .. && make &&  /usr/bin/xterm -e /home/schroer/gnuradiocomponents/gnuradio/grc/tests/resources/horn/build/horn

-- Configuring done
-- Generating done
-- Build files have been written to: /home/schroer/gnuradiocomponents/gnuradio/grc/tests/resources/horn/build
Scanning dependencies of target horn_autogen
[ 25%] Automatic MOC for target horn
[ 25%] Built target horn_autogen
Scanning dependencies of target horn
[ 50%] Building CXX object CMakeFiles/horn.dir/horn.cpp.o
In file included from /home/schroer/gnuradiocomponents/gnuradio/grc/tests/resources/horn/horn.cpp:8:
/home/schroer/gnuradiocomponents/gnuradio/grc/tests/resources/horn/horn.hpp:13:10: fatal error: gnuradio/top_block.h: Datei oder Verzeichnis nicht gefunden
 #include <gnuradio/top_block.h>
          ^~~~~~~~~~~~~~~~~~~~~~
compilation terminated.
make[2]: *** [CMakeFiles/horn.dir/build.make:63: CMakeFiles/horn.dir/horn.cpp.o] Fehler 1
make[1]: *** [CMakeFiles/Makefile2:73: CMakeFiles/horn.dir/all] Fehler 2
make: *** [Makefile:84: all] Fehler 2

Done (return code 2)
```

haakov:
Yeah the CMakeLists Mako template needs to be updated to modern CMake. I've begun working on it, but I'm not quite finished yet.

haakov:
Note to self: Give that flowgraph another ID :)

dkozel:
@haakov I just tested this on the head of `maint-3.8` and hit the same error. Would you be able to work on this?

dkozel:
Tested on `master`, same error.

haakov:
Yep, I'll look into it.

kc2rxo:
Just adding that I am running into this as well. Installed on a clean Ubuntu 20.04 using apt. Version from GRC is 3.8.1.0 (Python 3.8.5).

[...]

PR #4427, opened 2021-03-21
qtgui: Fix wintype in Qt Waterfall Sink

haakov:

PR #4710, opened 2021-06-04
grc: generating a simple cpp qt-gui flowgraph fails

dl1ksv:
Generating cpp code for the simple flowgraph described in #2688 fails as
an include statement and a required link lib is missing.

This pr fixes this issue. More complex flowgraphs will still fail.

### DIFF
--- a/gr-analog/grc/analog_agc2_xx.block.yml
+++ b/gr-analog/grc/analog_agc2_xx.block.yml
@@ -63,6 +63,6 @@ cpp_templates:
     - set_reference(${reference})
     - set_gain(${gain})
     - set_max_gain(${max_gain})
-    link: ['gnuradio-analog']
+    link: ['gnuradio::gnuradio-analog']
 
 file_format: 1
--- a/gr-analog/grc/analog_agc3_xx.block.yml
+++ b/gr-analog/grc/analog_agc3_xx.block.yml
@@ -67,6 +67,6 @@ cpp_templates:
     - set_reference(${reference})
     - set_gain(${gain})
     - set_max_gain(${max_gain})
-    link: ['gnuradio-analog']
+    link: ['gnuradio::gnuradio-analog']
 
 file_format: 1
--- a/gr-analog/grc/analog_agc_xx.block.yml
+++ b/gr-analog/grc/analog_agc_xx.block.yml
@@ -57,6 +57,6 @@ cpp_templates:
     - set_reference(${reference})
     - set_gain(${gain})
     - set_max_gain(${max_gain})
-    link: ['gnuradio-analog']
+    link: ['gnuradio::gnuradio-analog']
 
 file_format: 1
--- a/gr-analog/grc/analog_const_source_x.block.yml
+++ b/gr-analog/grc/analog_const_source_x.block.yml
@@ -32,6 +32,6 @@ cpp_templates:
     make: 'this->${id} = analog::sig_source_${type.fcn}::make(0, analog::GR_CONST_WAVE, 0, 0, ${const});'
     callbacks:
     - set_offset(${const})
-    link: ['gnuradio-analog']
+    link: ['gnuradio::gnuradio-analog']
 
 file_format: 1
--- a/gr-analog/grc/analog_cpfsk_bc.block.yml
+++ b/gr-analog/grc/analog_cpfsk_bc.block.yml
@@ -34,7 +34,7 @@ cpp_templates:
     make: 'this->${id} = analog::cpfsk_bc::make(${k}, ${amplitude}, ${samples_per_symbol});'
     callbacks:
     - set_amplitude(${amplitude})
-    link: ['gnuradio-analog']
+    link: ['gnuradio::gnuradio-analog']
 
 
 file_format: 1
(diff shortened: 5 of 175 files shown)


willcode:
Wow, painful but simple! How do we test that all those work and they don't have additional link deps?

dl1ksv:
We do not have to care about link dependencies. They are handled by cmake. For example the audio sink depends on ALSA and OSS, but there is no entry in the corresponding yml file.
These dependencies are handled in gnuradio-audioConfig.cmake. 

I can't test if  you can valid cpp code for all referenced blocks, but at the moment no block can be used in cpp as linking fails for all of them.

dl1ksv:
Next step: make gui_hint work!

haakov:
I'm not quite sure about these changes. The `test_cpp` flowgraph in https://github.com/gnuradio/gnuradio/issues/2688 works for me on latest master after applying https://github.com/gnuradio/gnuradio/pull/4427

dl1ksv:
@haakov I'm using a fresh build of the master branch on fedora 34. Sorry, I was not aware of your #4427 .
You are right , it's sufficient to add Qt5::Widgets to the target link libraries but including gnuradio::gnuradio-qtgui work ,too, as it has Qt5::Widgets  as dependency.

But if I replace in test_cpp the null sink by a waterfall sink ( your second commit already applied) , building fails with:

Scanning dependencies of target horn1
[ 50%] Building CXX object CMakeFiles/horn1.dir/horn1.cpp.o
[ 75%] Linking CXX executable horn1
/usr/bin/ld: cannot find -lgnuradio-qtgui
collect2: Fehler: ld gab 1 als Ende-Status zurück

Or just another example:

Replace  the null source by an analog signal source and try to build:
You get:

[100%] Linking CXX executable horn1
/usr/bin/ld: cannot find -lgnuradio-analog
collect2: Fehler: ld gab 1 als Ende-Status zurück
gmake[2]: *** [CMakeFiles/horn1.dir/build.make:136: horn1] Fehler 1
gmake[1]: *** [CMakeFiles/Makefile2:96: CMakeFiles/horn1.dir/all] Fehler 2
gmake: *** [Makefile:103: all] Fehler 2

That is why my second commit is necessary.

haakov:
Interesting! I'm travelling these days, but it'd be great if someone could try it out on an Ubuntu platform.

Are you compiling with LD_LIBRARY_PATH correctly set?

Relevant PR: https://github.com/gnuradio/gnuradio/pull/4263

dl1ksv:
Yes,  LD_LIBRARY_PATH is correctly set, otherwise grc would not start. 
And in one of your comments to #4263 you propose the prefix in the yaml file.
That's just what my second commit does.
And in
https://github.com/gnuradio/gnuradio/blob/d84df80edaf1e8491a1b4fc208b407b7e5c5198c/grc/core/generator/cpp_templates/CMakeLists.txt.mako#L55

The lib is explicitly prefixed. So I think this is the right way to handle the links. And all the examples I tested work with this change.

dl1ksv:
@haakov Meanwhile I have setup ubuntu 20.04 and testet against. There is no difference between ubuntu and fedora 33/34

To avoid the problems with the qtgui I set up a block without gui, just an analog source connected to a null sink. See attached flowgraph

[analog_source.grc.gz](https://github.com/gnuradio/gnuradio/files/6671398/analog_source.grc.gz)

If the link entry looks like 
link: ['gnuradio-analog']
in analog_sig_source_x.block.yml

linking fails with 

[ 50%] Building CXX object CMakeFiles/signal_source.dir/signal_source.cpp.o
[100%] Linking CXX executable signal_source
/usr/bin/ld: cannot find -lgnuradio-analog
collect2: Fehler: ld gab 1 als Ende-Status zurück
gmake[2]: *** [CMakeFiles/signal_source.dir/build.make:118: signal_source] Fehler 1
gmake[1]: *** [CMakeFiles/Makefile2:95: CMakeFiles/signal_source.dir/all] Fehler 2
gmake: *** [Makefile:103: all] Fehler 2

but setting it to

link: ['gnuradio::gnuradio-analog']
the build succeeds without any error and you can run the executable.

dl1ksv:
@haakov Meanwhile I have setup ubuntu 20.04 and testet against. There is no difference between ubuntu and fedora 33/34

To avoid the problems with the qtgui I set up a block without gui, just an analog source connected to a null sink. See attached flowgraph

[analog_source.grc.gz](https://github.com/gnuradio/gnuradio/files/6671398/analog_source.grc.gz)

If the link entry looks like 
link: ['gnuradio-analog']
in analog_sig_source_x.block.yml

linking fails with 

[ 50%] Building CXX object CMakeFiles/signal_source.dir/signal_source.cpp.o
[100%] Linking CXX executable signal_source
/usr/bin/ld: cannot find -lgnuradio-analog
collect2: Fehler: ld gab 1 als Ende-Status zurück
gmake[2]: *** [CMakeFiles/signal_source.dir/build.make:118: signal_source] Fehler 1
gmake[1]: *** [CMakeFiles/Makefile2:95: CMakeFiles/signal_source.dir/all] Fehler 2
gmake: *** [Makefile:103: all] Fehler 2

but setting it to

link: ['gnuradio::gnuradio-analog']
the build succeeds without any error and you can run the executable.

haakov:
Okay great! That convinces me that the `gnuradio::` prefix should be added.

PR #4711, opened 2021-06-04
gr-iio: Add packetized transfer support using stream tags

Yamakaja:
This PR adds stream tagging support to the device_source and device_sink
blocks, and thus also deriviates like the fmcomms2 and pluto.

The important bits are:

* The source block now tags buffer boundaries
* When tagged_input is enabled on the device sink, it enforces that all input packet sizes match buffer_size / (1 + interpolation). 

This isn't very flexible, but can be understood as a safeguard against unexpected behavior.

(Continuation of #4684)

### DIFF
--- a/gr-iio/grc/iio_device_sink.block.yml
+++ b/gr-iio/grc/iio_device_sink.block.yml
@@ -43,6 +43,11 @@ parameters:
     dtype: raw
     default: []
 
+-   id: len_tag_key
+    label: Packet Length Tag
+    dtype: string
+    hide: ${( 'part' )}
+
 inputs:
 -   domain: stream
     dtype: short
@@ -52,7 +57,9 @@ asserts:
 - ${ len(channels) > 0 }
 
 templates:
-    imports: from gnuradio import iio
-    make: iio.device_sink(${uri}, ${device}, ${channels}, ${device_phy}, ${params}, ${buffer_size}, ${interpolation} - 1, ${cyclic})
+  imports: from gnuradio import iio
+  make: |
+    iio.device_sink(${uri}, ${device}, ${channels}, ${device_phy}, ${params}, ${buffer_size}, ${interpolation} - 1, ${cyclic})
+    self.${id}.set_len_tag_key(${len_tag_key})
 
 file_format: 1
--- a/gr-iio/include/gnuradio/iio/device_sink.h
+++ b/gr-iio/include/gnuradio/iio/device_sink.h
@@ -14,6 +14,8 @@
 #include <gnuradio/iio/api.h>
 #include <gnuradio/sync_block.h>
 
+#include <string>
+
 #define DEFAULT_BUFFER_SIZE 0x8000
 
 extern "C" {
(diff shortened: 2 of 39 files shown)


gr-iio/grc/iio_device_sink.block.yml
```diff
@@ -3,56 +3,71 @@ label: IIO Device Sink
 flags: [ python ]
 
 parameters:
--   id: uri
+  - id: uri
```
tfcollins, line 6: Standard gr seems to be "-   " not "   -"

gr-iio/include/gnuradio/iio/device_sink.h
```diff
@@ -71,6 +73,18 @@ class IIO_API device_sink : virtual public gr::sync_block
                           unsigned int buffer_size = DEFAULT_BUFFER_SIZE,
                           unsigned int interpolation = 0,
                           bool cyclic = false);
+
+    /*!
+     * The key of the tag that indicates packet length.
+     * When not empty, samlpes are expected as "packets" and
```
tfcollins, line 79: samlpes -> samples

gr-iio/lib/device_sink_impl.cc
```diff
@@ -165,12 +167,38 @@ void device_sink_impl::channel_write(const struct iio_channel* chn,
...
                            gr_vector_void_star& output_items)
 {
     int ret;
 
+    if (d_len_tag_key != pmt::PMT_NIL) {
+        for (size_t i = 0; i < input_items.size(); i++) {
+            auto items_read = nitems_read(i);
+            get_tags_in_range(d_tags, i, items_read, items_read + 1, d_len_tag_key);
+
+            if (d_tags.size() != 1)
```
tfcollins, line 189: When you have small IIO buffers I think this may break since the gr buffers could be larger. Meaning you could get multiple tagged buffers passed.

tfcollins (changes requested):
1. Can you remove the extra formatting as it breaks away from the standard gr (from what I can tell) and if formatting changes are required they should be in a separate commit.
2. With the restriction on packet length on TX, shouldn't the properties be named something like packetized_tag_key? Just to remove any confusion around actually expecting a length in the packet tag itself to be tunable.

Yamakaja:
V2:
* Addresses review comments *except* for the `packetized_tag_key` part, where i'm going to look for a little more input ;)
* Rebased on master

Yamakaja:
V3:
* Fixes device & pluto sink grc ymls

gr-iio/lib/device_source_impl.cc
```diff
@@ -290,12 +299,20 @@ int device_source_impl::work(int noutput_items,
             return 0;
 
         byte_offset = 0;
+
+        // Tag start of new packet
+        if (d_len_tag_key != pmt::PMT_NIL)
```
mormj, line 304: Can you please add braces around (even small) if and for loops

mormj (changes requested):

Yamakaja:
V4:
* Added some braces

mormj (approved):

mormj:
Looks like @tfcollins feedback was addressed - and thank you for adding the braces :)  Merging

PR #4717, opened 2021-06-04
cmake: Pass extra arguments to GR_PYTHON_INSTALL to install command.

marcusmueller:
GRC already does a few GR_PYTHON_INSTALLs with extra FILES_MATCHING
REGEX arguments, so this will actually pass these through.

This was necessary in an earlier version of #4671, but isn't anymore and has thus been moved to a separate PR.


ryanvolz (approved):
The change is originally mine, so obviously I approve 😆.

PR #4722, opened 2021-06-05
Remove Python 3 f-strings from previous backport

willcode:
Forgot we are still supporting Python 2 on 3.8!


marcusmueller (approved):
Clearly a good/necessary change

PR #4729, opened 2021-06-05
cmake: lowercase pkg-config module requires lowercase pybind11_INCLUDE_DIR

marcusmueller:
Stumbled across this while working on precompiled headers. Probably has not
been a problem since Pybind11 tends towards "boring" install locations.


willcode:
or should it be PYBIND11 in ./build/CMakeCache.txt and ./gnuradio-runtime/lib/CMakeLists.txt?

marcusmueller:
needs to be lower case, pkg-config package is called pybind11, not PYBIND11

willcode:
I don't really understand how all that works, but all the other packages seem to be upper case, and
```
$ pkg-config pybind11 --version
1.7.3
$ pkg-config PYBIND11 --version
1.7.3
```

mormj (approved):

ryanvolz:
FYI, this would be obsoleted with #4617.

mormj:
Going ahead and merging this one since it is a clear bug ... would like to figure out the pybind11 mandate on non-python builds before merging #4617

PR #4734, opened 2021-06-07
Replaced star imports with explicit imports

MarkBauer5:
Replaced 'import *' with explicit imports, and ran pep8 formatters on changed files.

Signed-off-by: Mark Bauer <markb5@illinois.edu>


gr-fec/python/fec/LDPC/Generate_LDPC_matrix.py
```diff
@@ -35,7 +40,7 @@
 p = 3     # column weight
 q = 5     # row weight
 
-parity_check_matrix = LDPC_matrix(n_p_q = [n,p,q])
+parity_check_matrix = LDPC_matrix(n_p_q=[n, p, q])
```
mormj, line 43: I'm surprised the pep8 formatter would have removed the spaces around the `=`

MarkBauer5: Upon closer review, the pep8 formatter I used made several questionable edits - I'll manually amend them to a style more consistent with correct pep8.

mormj: It is is also good practice on a PR when there is a lot of re-formatting to separate the functional parts of the commit (e.g. replacing the * imports) from the reformatting into separate commits

mbr0wn: I would suggest we apply some code formatting automation the same way we do to our C++ code. I've seen a lot of people use [black(https://pypi.org/project/black/), but there are other tools too (for C++, there's really only clang-format). But that's a discussion for elsewhere.

MarkBauer5: Thank you for the suggestion. I've ran the Black formatting tool on the committed code and put the results into a separate commit as recommended by @mormj .

PR #4743, opened 2021-06-08
soapy: less surprising bool cast

### LINKED ISSUE #4742: string == "0.0" isn't sufficient for testing zeroness of string
Soapy Python bindings do this:

https://github.com/gnuradio/gnuradio/blob/f29be5398081e78ed0edeaf712cc608e770f2523/gr-soapy/python/soapy/bindings/soapy_common.cc#L57-L66

this will be hilarious if someone tries 0e0, -0, or +0 and gets something different than if they use 0.0.

Let's not have that much hilarity in our lives :D

marcusmueller:
Soapy has its own way of casting strings to bools, and it special-cases
"0" and "0.0" and misses out on the opportunity to be consistent with
many other ways of specifying 0 as number.

I'm trying to upstream the same change to SoapSDR, in order to keep
things consistent. But this is worth having in GNU Radio, either way,
since GNU Radio has its own user expectations, which might even differ a
bit from these of SoapySDR users.

Fixes #4742

References https://github.com/pothosware/SoapySDR/pull/313

### DIFF
--- a/gr-soapy/python/soapy/bindings/soapy_common.cc
+++ b/gr-soapy/python/soapy/bindings/soapy_common.cc
@@ -11,6 +11,8 @@
 
 #include <SoapySDR/Types.hpp>
 #include <SoapySDR/Version.h>
+#include <stdexcept>
+#include <string>
 
 // SoapySDR doesn't have an API define for SettingToString, so we need
 // to check the version. 0.8 is the first tagged version to have this
@@ -49,22 +51,20 @@ static inline T string_to_setting(const std::string& str)
 }
 
 // Copied from SoapySDR 0.8
+//! convert empty and "false" strings to false, integers to their truthness
 template <>
 inline bool string_to_setting<bool>(const std::string& str)
 {
-    if (str == SOAPY_SDR_TRUE)
+    if (str.empty() || str == SOAPY_SDR_FALSE) {
+        return false;
+    }
+    if (str == SOAPY_SDR_TRUE) {
         return true;
-    if (str == SOAPY_SDR_FALSE)
-        return false;
-
-    // zeros and empty strings are false
-    if (str == "0")
-        return false;
-    if (str == "0.0")
-        return false;
-    if (str == "")
-        return false;
-
+    }
+    try {
+        return static_cast<bool>(std::stod(str));
+    } catch (std::invalid_argument& e) {
+    }
     // other values are true
     return true;
 }


gr-soapy/python/soapy/bindings/soapy_common.cc
```diff
@@ -49,22 +51,24 @@ static inline T string_to_setting(const std::string& str)
...
-        return false;
-    if (str == "")
-        return false;
-
+    }
+    if (str == SOAPY_SDR_TRUE) {
+        return true;
+    }
+    try {
+        return std::stoi(str);
```
willcode, line 65: std::stoi("0.1") returns 0.

marcusmueller: oh yes ... let's just check for stod, as every integer should cleanly cast to double, anyway.

willcode: Right.

In the qa on the SoapySDR PR,
`check_equal(SoapySDR::StringToSetting<bool>("0.2e12"), false);`
should be true
`check_equal(SoapySDR::StringToSetting<bool>("0.2"), true);`
should have failed the way it was coded

marcusmueller: yeah... shouldn't have written that code tired... fixed the unit test and both the conversion in this and the PR over at SoapySDR

gr-soapy/python/soapy/bindings/soapy_common.cc
```diff
@@ -49,22 +51,20 @@ static inline T string_to_setting(const std::string& str)
...
-    if (str == SOAPY_SDR_FALSE)
-        return false;
-
-    // zeros and empty strings are false
-    if (str == "0")
-        return false;
-    if (str == "0.0")
-        return false;
-    if (str == "")
+    if (str.empty() || str == SOAPY_SDR_FALSE) {
```
mbr0wn, line 58: What about lower-case version of `str`?

gr-soapy/python/soapy/bindings/soapy_common.cc
```diff
@@ -49,22 +51,20 @@ static inline T string_to_setting(const std::string& str)
...
         return false;
-
+    }
+    if (str == SOAPY_SDR_TRUE) {
+        return true;
+    }
+    try {
+        return std::stod(str);
```
mbr0wn, line 65: Even though it comes out the same, can you please:
```suggestion
        return static_cast<bool>(std::stod(str));
```
MSVC warns about this otherwise, but I also think it tells a better story when reading the code.

gr-soapy/python/soapy/bindings/soapy_common.cc
marcusmueller: was considering that, too, but I don't think we need to be more flexible there – `False` might be tempting.

If I were to change that behaviour however, I'd rather do it upstream in SoapySDR and then do the same in GR than the other way around

gr-soapy/python/soapy/bindings/soapy_common.cc
marcusmueller: very true

willcode (approved):

ncorgan:
This looks reasonable to me, and I don't see a reason @guruofquality would object to this in Soapy itself, but please make sure to keep these PRs in sync.

PR #4159, opened 2021-01-27
MSVC Fixes (Rebased)

mbr0wn:
Rebased https://github.com/gnuradio/gnuradio/pull/4138

If the tests pass this time, it can go in.

PR #4744, opened 2021-06-08
blocks: Windows defines required (maint-3.9)

ryanvolz:
Similar to #4539, we now need to make sure that `HAVE_WINDOWS_H` is defined for the `gnuradio-blocks` target. See discussion in #4159, patch taken from what @willcode posted there and fixed to use `gnuradio-blocks` instead of `gnuradio-network`.

### DIFF
--- a/gr-blocks/lib/CMakeLists.txt
+++ b/gr-blocks/lib/CMakeLists.txt
@@ -10,6 +10,7 @@
 ########################################################################
 include(GrMiscUtils)
 GR_CHECK_HDR_N_DEF(io.h HAVE_IO_H)
+CHECK_INCLUDE_FILE_CXX(windows.h HAVE_WINDOWS_H)
 
 ########################################################################
 # Setup library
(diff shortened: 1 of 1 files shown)


ryanvolz:
I'm waiting to see if this successfully builds with the `conda-forge` package before clicking the 'Ready' button.

willcode:
Any reason we end up doing this for network and gnuradio-runtime but not others.

Is this important enough to merge before this week's releases? The scope is limited, so I don't see much risk.

ryanvolz:
It's only used in the PDU code, which is why it's in blocks for 3.9 and network for master. This should definitely get into the release, assuming I verify it works, to save headaches for other Windows builders.

ryanvolz:
The Windows build looks good with this (after I fixed it because CMake complained about multiple `target_link_libraries` if they all don't use `PUBLIC`/`PRIVATE`). Ready to go.

PR #4745, opened 2021-06-08
cmake: Ensure that RC_*_VERSION are numeric.

ryanvolz:
Non-numeric values (such as "0-rc1") cause build errors.

Signed-off-by: Ryan Volz <ryan.volz@gmail.com>


willcode:
This won't bother us for a couple months, so no backport at the moment. Where did this show up? It doesn't cause any problems in Linux builds.

ryanvolz:
These variables are only used for Windows RC files, so it's our good friend Windows striking again! You're right that there's no rush here, unless you wanted to do another release candidate.

mormj (approved):

willcode (approved):

PR #4760, opened 2021-06-11
runtime: tags can be compared with < instead of offset_compare

marcusmueller:


gnuradio-runtime/include/gnuradio/tags.h
```diff
@@ -33,15 +33,15 @@ struct GR_RUNTIME_API tag_t {
...
-     * first in time
-     */
-    static inline bool offset_compare(const tag_t& x, const tag_t& y)
+    //! Comparison function to test which tag, \p x or \p y, came first in time
+    friend inline bool operator<(const tag_t& x, const tag_t& y)
     {
         return x.offset < y.offset;
     }
+    //! Comparison function to test which tag, \p x or \p y, came first in time
+    static inline bool offset_compare(const tag_t& x, const tag_t& y) { return x < y; }
```
mormj, line 42: Since this file is in the public API, should the implementations be moved into `tag.cc`?

marcusmueller: ah wait, can't be inline then. So, probably not.

mormj: doh - yep merging now.

PR #4762, opened 2021-06-12
testing: rm dependencies from disabled components in blocks/runtime tests

marcusmueller:
This included shifting of the PDU test from runtime to gr-pdu, and
making the tests that actually require blocks conditional on whether
there's going to be gr-blocks. Also, don't use gr-analog just because
you need data.

### DIFF
--- a/gnuradio-runtime/python/gnuradio/gr/CMakeLists.txt
+++ b/gnuradio-runtime/python/gnuradio/gr/CMakeLists.txt
@@ -1,4 +1,5 @@
 # Copyright 2012 Free Software Foundation, Inc.
+# Copyright 2021 Marcus Müller
 #
 # This file is part of GNU Radio
 #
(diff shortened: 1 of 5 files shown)


marcusmueller:
(this is nice, because now I can build a gnuradio-runtime-only system again and just test my tinkering in runtime)

gnuradio-runtime/python/gnuradio/gr/CMakeLists.txt
```diff
@@ -1,4 +1,5 @@
 # Copyright 2012 Free Software Foundation, Inc.
+# Copyright 2021 Free Software Foundation, Inc.
```
mormj, line 2: can this just be 2012, 2021 - or if separate line, copyright Marcus

mormj (approved):

mormj:
This is excellent - love the idea of being able to not build and rebuild all the modules.

mormj (approved):

Issue #4763, opened 2021-06-12
Remove <iostream> remnants


marcusmueller:
To my surprise we still have quite a few places in-tree that include iostream – many of these don't use the stuff anymore.

The others partially should use logging, some are totally OK using direct output.

marcusmueller:
aside from places where it's OK (gr::prefs, gnuradio-config-info, pmt::print, flowgraph dotfile dump, terminate handler…) still present in the C++ flowgraph template, Thrift, fec_mtrx_impl, polar_common (and other gr-fec places).

Since this leads to interrupted logs due to orthogonal buffering, this is a bug and I'm now tagging with the affected modules.

Find the critical occurrences grepping for `std::cout` or `std::cerr`.

PR #4764, opened 2021-06-12
global: remove iostream from all files not use cout|cerr|cin|clog

marcusmueller:
Adresses #4763 partially

### DIFF
--- a/gnuradio-runtime/lib/flowgraph.cc
+++ b/gnuradio-runtime/lib/flowgraph.cc
@@ -17,9 +17,12 @@
 #include <sstream>
 #include <stdexcept>
 
+// TODO: Replace with GNU Radio logging
+#include <iostream>
+
 namespace gr {
 
-#define FLOWGRAPH_DEBUG 0
+constexpr bool FLOWGRAPH_DEBUG = false;
 
 edge::~edge() {}
 
--- a/gnuradio-runtime/include/gnuradio/flowgraph.h
+++ b/gnuradio-runtime/include/gnuradio/flowgraph.h
@@ -14,7 +14,6 @@
 #include <gnuradio/api.h>
 #include <gnuradio/basic_block.h>
 #include <gnuradio/io_signature.h>
-#include <iostream>
 
 namespace gr {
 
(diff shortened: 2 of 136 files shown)


marcusmueller:
this might be worth squashing

mormj (approved):

gnuradio-runtime/lib/flowgraph.cc
```diff
@@ -73,9 +77,9 @@ void flowgraph::validate()
         std::vector<int> used_ports;
         int ninputs, noutputs;
 
-        if (FLOWGRAPH_DEBUG)
-            std::cout << "Validating block: " << (*p) << std::endl;
-
+#if FLOWGRAPH_DEBUG
```
ThomasHabets, line 80: I would not make this change just to remove an include. `#if` has the drawback that it makes it more complex to do static analysis, breaks indentation, and risks breaking code that happens to not be active.

I would avoid preprocessor stuff unless it definitely can't compile otherwise. For C++17 we can use `if constexpr`.

Also elsewhere.

ThomasHabets (changes requested):

gnuradio-runtime/lib/flowgraph.cc
marcusmueller: I'm fine with that line of thinking. It's not a header, so the transitive include isn't hurting anyone. Let me revert that file

marcusmueller: (along with `hier_block2_detail.cc`)

reverted, @ThomasHabets.

ThomasHabets (approved):

willcode:
Backport note: doesn't apply cleanly to maint-3.9 due to diverging code. I'd still like to get some of these changes in to keep from divergin further over trivial stuff, when time permits.

Recreate the main change with
```
sed -i '/^#include.*iostream.*$/d' $(ag -L `\b(cout|cerr|cin|clog)\b` $(ag -l 'include.*<iostream>' .))
```
followed by pybind11 hash fixups.

marcusmueller:
> followed by pybind11 hash fixups.

    gr-utils/bindtool/scripts/binding-hash-hotfixer.zsh gnuradio-runtime/python/gnuradio/gr/bindings/*.cc gr-*/python/*/bindings/*.cc

PR #4778, opened 2021-06-16
global: remove unused boost includes

marcusmueller:
used command:

    sed -i '/^#include.*boost.*$/d' $(ag -L '(boost::|BOOST)' $(ag --cpp -l '#include <boost'))

With this changeset, channels and wavelet don't use boost internally anymore; a boost-free GNU Radio: it *is* possible.

### DIFF
--- a/gnuradio-runtime/include/gnuradio/buffer.h
+++ b/gnuradio-runtime/include/gnuradio/buffer.h
@@ -16,7 +16,6 @@
 #include <gnuradio/runtime_types.h>
 #include <gnuradio/tags.h>
 #include <gnuradio/thread/thread.h>
-#include <boost/weak_ptr.hpp>
 #include <map>
 #include <memory>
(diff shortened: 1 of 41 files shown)


mormj:
I think @marcusmueller comment on #4779 would apply here re: backporting - this constitutes an API change because of things that might be transitively included via these headers in OOTs

mormj:
I especially love watching boost::format go away!

mormj (approved):

marcusmueller:
> I especially love watching boost::format go away!

ha! For now, it just moves into the files that actually use it, but that's still a nice thing :)

willcode:
Not sure if transitive includes are a feature or a bug, but I'll hold off backporting boost removal. The same thing could be said of <iostream> but that's much more in the "bug" category if OOTs assume it's included.

PR #4779, opened 2021-06-17
Remove #include <boost/format.hpp> from logger.h

marcusmueller:
depends on #4778 

This is a bigger cleanup, as it requires including <boost/format.hpp> wherever it was actually used, but will be necessary sooner or later when we replace the logging subsystem.

### DIFF
--- a/gnuradio-runtime/include/gnuradio/logger.h
+++ b/gnuradio-runtime/include/gnuradio/logger.h
@@ -28,7 +28,6 @@ typedef int mode_t;
 // buffer.cc, then only include it there
 #include <gnuradio/api.h>
 #include <log4cpp/Category.hh>
-#include <boost/format.hpp>
 #include <boost/thread.hpp>
 #include <cassert>
 #include <memory>
--- a/gnuradio-runtime/lib/buffer.cc
+++ b/gnuradio-runtime/lib/buffer.cc
@@ -15,6 +15,7 @@
 #include <gnuradio/buffer.h>
 #include <gnuradio/integer_math.h>
 #include <gnuradio/math.h>
+#include <boost/format.hpp>
 #include <algorithm>
 #include <cassert>
 #include <stdexcept>
--- a/gnuradio-runtime/lib/controlport/thrift/rpcserver_booter_thrift.cc
+++ b/gnuradio-runtime/lib/controlport/thrift/rpcserver_booter_thrift.cc
@@ -10,6 +10,7 @@
 
 #include <gnuradio/rpcserver_booter_thrift.h>
 #include <gnuradio/rpcserver_thrift.h>
+#include <boost/format.hpp>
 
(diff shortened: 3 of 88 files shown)


marcusmueller:
Squashing: will do.

whether or not to backport: probably not, because seeing how many files in-tree assumed boost/format to be transitively included, this will break a lot of OOTs.

mbr0wn (changes requested):
I think we should not merge this. Tom and I discussed adding boost/format.hpp to the logger years ago because frankly, the currently logger only really works with boost::format.

We all realize this is the "wrong" way to do things, but as long as we have this logging subsystem, I suggest we skip these "cleanup" changes, as the risk/value ratio doesn't warrant it. When we address the logging subsystem itself, then we can make sure that we properly remove these headers.

marcusmueller:
On the risk assessment, I don't quite agree: we could simply keep the `#include` in the logger header, and still add it where we use it in-tree everywhere else: that mitigate all risk for the ecosystem.

This changeset was really meant to facilitate working on the logging subsystem, and on removing boost: By including things where we use them, it's actually possible to change central things without breaking usage. So, the "include boost::format where we use it" change would be desirable, no matter what, because I don't think anyone would want to rewrite all in-tree logging in one go.

So, that's the motivation: making it less of a drag to start working on logging. This was a drag to do once, but anyone would basically want to do exactly that, removing the include from logger.h, seeing where things break, then including boost/format in these files, then change the logger in a backwards-compatible way, then go through all files that include boost/format and replace "legacy" boost format with plain "give it to spdlog to format" and finally remove the dependency from the modules that don't need it anymore.

But I do agree, we don't have to merge this now; it can be a bit of a `next`-branch thing, if you will, to be merged as soon as anyone actually dedicates time to logging. The list of changed files itself is a bit of a useful tool in itself!

mbr0wn:
Can we close this for #4981 ?

mbr0wn:
@mormj I misunderstood the purpose of this PR. It's a subset of the spdlog work, so it can be merged first (thus reducing the diff of the spdlog PR).

mbr0wn:
I'm still unsure if we should remove boost/format.hpp from logger.h right now. Wouldn't that also require all OOTs to include boost/format.hpp if they use it the loggers and haven't included? I mean, they should, of course. But this might still break stuff.

mormj:
> I'm still unsure if we should remove boost/format.hpp from logger.h right now. Wouldn't that also require all OOTs to include boost/format.hpp if they use it the loggers and haven't included? I mean, they should, of course. But this might still break stuff.

It's something that we should be prepared to break - but if the move to spdlog is in the very near term, then we shouldn't make 2 breaks - let's do it all at once.

mbr0wn:
OK, thanks for the guidance. I would still suggest we do the IWYU work first, so it doesn't pollute the spdlog PR. I'll move that to a separate PR then.

PR #4784, opened 2021-06-18
grc: Update flowgraph vars controlled by vars

### LINKED ISSUE #4781: Flowgraph doesn't update properly if samp_rate is controlled by another variable
If the samp_rate variable is a function of another variable, the flow graph is not updated properly when the variable is changed (and OK is clicked). Here's the example flow graph. samp_rate is set to (symbol_rate * 2). When symbol_rate is changed (in this case from 32k to 48k), Sample Rate in all the blocks is not updated.

I've bisected this behavior to commit https://github.com/gnuradio/gnuradio/commit/c4f46bfa6d790c57018f9c6fca264bb9e01a2854

image

solomonbstoner:
Fixes #4781. When a variable is controlled by another variable,
updating the latter does not update the former in the flowgraph when
`Ok` is pressed. Caused by commit c4f46bf. This commit patches the bug.

### DIFF
--- a/grc/gui/Application.py
+++ b/grc/gui/Application.py
@@ -557,7 +557,7 @@ class Application(Gtk.Application):
                         ### Following line forces a complete update of io ports
                         flow_graph_update()
                         page.saved = False
-                    else:  # restore the current state
+                    if response in (Gtk.ResponseType.REJECT, Gtk.ResponseType.ACCEPT):
                         n = page.state_cache.get_current_state()
                         flow_graph.import_data(n)
                         flow_graph_update()


solomonbstoner:
Commit 8806a36 introduces an error with the use of `Cancel`. Following said commit, the entered before `Cancel` is pressed will remain in the gui

solomonbstoner:
Fixed it. Apparently the `Cancel` button invokes the `Gtk.ResponseType.REJECT` in `Application.py` instead of `Gtk.ResponseType.CANCEL`.

I do not think this PR will reintroduce the bug that commit c4f46bf solved since `flow_graph_update` does not get called until the `Ok` button is pressed. Ready for review.

drmpeg (approved):
Tested and works good here.

PR #4785, opened 2021-06-18
runtime/block: send bools to the done port, accept bools

### LINKED ISSUE #4775: `block::system_handler` expects PMT long, gets PMT bool
Investigation #4770  I found something confusing:

`block.cc`:
```c++
void block::system_handler(pmt::pmt_t msg)
{
…
        d_finished = pmt::to_long(pmt::cdr(msg));
…}
```

notice the `to_long` in there: that fails on PMTs that `is_bool()`

in `notify_msg_neighbors`, however:

```c++
blk->post(d_system_port, pmt::cons(d_pmt_done, pmt::mp(true)));
```

Is it possible we currently can't shut down message-based flow graphs or am I misreading this?

marcusmueller:
but continue accepting numbers for compatibility

fixes #4775

### DIFF
--- a/gnuradio-runtime/lib/block.cc
+++ b/gnuradio-runtime/lib/block.cc
@@ -17,6 +17,7 @@
 #include <gnuradio/block_registry.h>
 #include <gnuradio/buffer.h>
 #include <gnuradio/prefs.h>
+#include <pmt/pmt.h>
 #include <iostream>
 #include <stdexcept>
 
@@ -597,7 +598,15 @@ void block::system_handler(pmt::pmt_t msg)
     // std::cout << "system_handler " << msg << "\n";
     pmt::pmt_t op = pmt::car(msg);
     if (pmt::eqv(op, d_pmt_done)) {
-        d_finished = pmt::to_long(pmt::cdr(msg));
+        auto value = pmt::cdr(msg);
+        if (pmt::is_bool(value)) {
+            d_finished = pmt::to_bool(value);
+        } else if (pmt::is_number(value)) {
+            d_finished = pmt::to_long(pmt::cdr(msg));
+        } else {
+            std::cout << "WARNING: bad message value on system port!\n";
+            return;
+        }
         global_block_registry.notify_blk(d_symbol_name);
     } else {
         std::cout << "WARNING: bad message op on system port!\n";
@@ -634,7 +643,7 @@ void block::notify_msg_neighbors()
 
             currlist = pmt::cdr(currlist);
             basic_block_sptr blk = global_block_registry.block_lookup(block);
-            blk->post(d_system_port, pmt::cons(d_pmt_done, pmt::mp(true)));
+            blk->post(d_system_port, pmt::cons(d_pmt_done, pmt::from_bool(true)));
         }
     }
 }
--- a/gnuradio-runtime/lib/block.cc
+++ b/gnuradio-runtime/lib/block.cc
@@ -8,6 +8,7 @@
  *
  */
 
+#include <cxxabi.h>
 #ifdef HAVE_CONFIG_H
 #include "config.h"
 #endif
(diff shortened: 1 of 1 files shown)


jacobagilbert (commented):
This needs to be fixed, but is there any reason to handle something other than a PMT bool? I would consider that [after this change] a non-bool cdr() sent to this port to be a bug.

marcusmueller:
@jacobagilbert  thing is that sending a pmt::cons(pmt::mp("done"), pmt::mp(false)) to the system port is a way to "unblock" a block that's been blocked by a previous iteration where it produced zero items. 

Hence, that's a pattern that SDR driver OOTs might have used. So, it's API. Let me show what I've got in mind with another commit here.

jacobagilbert:
I get why one would want to send `pmt::PMT_F` but i think what you are saying is that we need to accept `pmt::mp(0)` for now for compatibility also.

Sorry for 20 questions...but what is the reason to switch to accepting dicts() on this port?

marcusmueller:
> Sorry for 20 questions...

Um, never be sorry for asking question!

> but what is the reason to switch to accepting dicts() on this port?

Consistency – other message control ports accept dicts. (the question maybe is whether we need a "system" port that accept "command dicts or pairs" which really can only contain one key)

gnuradio-runtime/lib/block.cc
```diff
@@ -594,15 +596,54 @@ void block::reset_perf_counters()
...
+                }
+            }
+        } else {
+            std::cout << "WARNING: bad message op on system port: " << op << "\n";
+        }
+    } else if (pmt::is_dict(msg)) {
+        // dict
+        auto value = pmt::dict_ref(msg, d_pmt_done, pmt::PMT_NIL);
+
+        if (value != pmt::PMT_NIL) {
```
jacobagilbert, line 627: maybe simplify:
```c++
if (pmt::is_bool(value)) {
  d_finished = pmt::to_bool(value);
} else {
  std::cout << "WARNING: bad message op in dict on system port!\n";
  return;
}
```

gnuradio-runtime/lib/block.cc
```diff
@@ -594,15 +596,54 @@ void block::reset_perf_counters()
...
+            try {
+                d_finished = pmt::to_bool(value);
+            } catch (pmt::wrong_type& e) {
+                std::cout << "WARNING: got a " << d_pmt_done
+                          << " key with non-bool value: " << e.what() << "\n";
+                return;
+            }
+        } else {
+            std::cout << "WARNING: bad message op in dict on system port!\n";
```
jacobagilbert, line 636: is `d_pmt_done` the only thing this port can handle?

gnuradio-runtime/lib/block.cc
```diff
@@ -634,7 +675,7 @@ void block::notify_msg_neighbors()
 
             currlist = pmt::cdr(currlist);
             basic_block_sptr blk = global_block_registry.block_lookup(block);
-            blk->post(d_system_port, pmt::cons(d_pmt_done, pmt::mp(true)));
+            blk->post(d_system_port, pmt::cons(d_pmt_done, pmt::PMT_T));
```
jacobagilbert, line 678: Probably should change to
```c++
blk->post(d_system_port, pmt::dict_add(pmt.make_dict(), d_pmt_done, pmt::PMT_T));
```
if thats the recommendation

jacobagilbert (changes requested):

gnuradio-runtime/lib/block.cc
marcusmueller: yep, at this point that's true – but I'm sure the intention was to control more through that port. Honestly, don't know whether that's wise.

gnuradio-runtime/lib/block.cc
marcusmueller: I like simple.

mormj:
This seems to stall out on the message passing QA tests

mormj:
@marcusmueller - I like the added flexibility that this gives the system port, but going to close this one for now since it breaks the message passing qa tests.

PR #4789, opened 2021-06-19
gr-digital: Simple header payload demux example

solomonbstoner:
This commit adds a simple example combining the header/payload block
with the formatter_crc example. Its purpose is to show a simpler, more
beginner friendly example for the header/payload block compared to the
current ofdm hier block used as an example.

### DIFF
--- a/gr-digital/examples/packet/formatter_crc_header_payload_demux.grc
+++ b/gr-digital/examples/packet/formatter_crc_header_payload_demux.grc
@@ -1,0 +1,684 @@
+options:
+  parameters:
+    author: Solomon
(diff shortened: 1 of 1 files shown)


mormj (approved):
This is fantastic - examples are always welcome additions, especially when they more simply show a particular functionality.  Thanks for the contribution!

willcode:
@solomonbstoner The `master` version of the grc uses blocks from the new `pdu` namespace, so this would have to be rebuilt under maint-3.9 in order to be backported. If you end up wanting to do that, it can go in a separate backport PR.

PR #4792, opened 2021-06-20
blocks: zero-output, zero-copy Head mode

marcusmueller:
If you don't want to test a stream with an exact number of input, but
just want your flow graph to terminate after a given number of items,
having a head block without an output enables you to do that without
copying the data from Head's in- to its output.


mormj (approved):

dkozel:
Cool feature @marcusmueller!

PR #4793, opened 2021-06-20
blocks: Throttle consume-only mode

marcusmueller:
When throttling a sample flow, it's not necessary to produce items at
the desired average rate - consuming them at a limited rate suffices to
slow down the processing of the whole flow graph just as well, and
avoids a copy.

### DIFF
--- a/gr-blocks/grc/blocks_throttle.block.yml
+++ b/gr-blocks/grc/blocks_throttle.block.yml
@@ -35,6 +35,7 @@ outputs:
 -   domain: stream
     dtype: ${ type }
     vlen: ${ vlen }
+    optional: true
 
 asserts:
 - ${ vlen > 0 }
--- a/gr-blocks/include/gnuradio/blocks/throttle.h
+++ b/gr-blocks/include/gnuradio/blocks/throttle.h
@@ -1,6 +1,7 @@
 /* -*- c++ -*- */
 /*
  * Copyright 2005-2011,2013 Free Software Foundation, Inc.
+ * Copyright 2021 Marcus Müller
  *
  * This file is part of GNU Radio
  *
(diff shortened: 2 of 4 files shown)


jasonuher (changes requested):
I couldn't get it to work in GRC without this:
```
diff --git a/gr-blocks/grc/blocks_throttle.block.yml b/gr-blocks/grc/blocks_throttle.block.yml
index ae31a9be2..894091156 100644
--- a/gr-blocks/grc/blocks_throttle.block.yml
+++ b/gr-blocks/grc/blocks_throttle.block.yml
@@ -35,6 +35,7 @@ outputs:
 -   domain: stream
     dtype: ${ type }
     vlen: ${ vlen }
+    optional: true
 
 asserts:
 - ${ vlen > 0 }

```

marcusmueller:
@jasonuher Ah, right. That's necessary for GRC usage! Added it.

dkozel:
I think this is a reasonable feature, though understanding how it works relies on an understanding of how GNU Radio handles data movement into/out of buffers. Do you have a specific usecase which motivated you to use a throttle in parallel with the main data flow? In GRC you can use bypass to easily remove/add a Throttle in series with dataflow. When setting up a flowgraph in Python/C++ I guess it is simpler to be able to conditionally hang a throttle off an output without affecting the rest of the connections.

jacobagilbert:
@dkozel IMO this is a good feature. Using a throttle in parallel can save a memcpy at a minimum, and there are other minor niche/in-the-weeds benefits of allowing throttling between two blocks while still allowing them share the same output buffer.

mormj:
Going ahead and merging this - but @duggabe note that the wiki documentation will need to be updated to reflect this usage.

PR #4794, opened 2021-06-20
Fix BlockTreeWindow Tooltips

NeuroForLunch:
No more double tooltips & a nicer presentation. "undocumented" also shows up now when there is no documentation. Before it was impossible for the code to ever get to that line.

### DIFF
--- a/grc/gui/BlockTreeWindow.py
+++ b/grc/gui/BlockTreeWindow.py
@@ -30,19 +30,17 @@ NAME_INDEX, KEY_INDEX, DOC_INDEX = range(3)
 
 def _format_doc(doc):
     docs = []
-    if doc.get(''):
-        docs += doc.get('').splitlines() + ['']
     for block_name, docstring in six.iteritems(doc):
-        docs.append('--- {0} ---'.format(block_name))
+        docs.append(format(block_name))
         docs += docstring.splitlines()
         docs.append('')
     out = ''
     for n, line in enumerate(docs[:-1]):
-        if n:
+        if n>1:
             out += '\n'
         out += Utils.encode(line)
-        if n > 10 or len(out) > 500:
-            out += '\n...'
+        if n > 40 or len(out) > 1000:
+            out += '\n >>>>> continued...'
             break
     return out or 'undocumented'
 


grc/gui/BlockTreeWindow.py
```diff
@@ -30,19 +30,19 @@
 
 def _format_doc(doc):
     docs = []
-    if doc.get(''):
-        docs += doc.get('').splitlines() + ['']
+    #if doc.get(''):
```
mormj, line 33: please remove commented code

grc/gui/BlockTreeWindow.py
```diff
@@ -30,19 +30,19 @@
...
         docs.append('')
     out = ''
     for n, line in enumerate(docs[:-1]):
-        if n:
+        if n>1:
             out += '\n'
         out += Utils.encode(line)
-        if n > 10 or len(out) > 500:
-            out += '\n...'
+        if n > 40 or len(out) > 2000:
```
mormj, line 44: what is the reasoning of these hardcoded ints - can they be replaced by meaningfully named constants?

mormj:
@NeuroForLunch - this PR is targeted to GR 3.8 - is there a specific reason to target this toward 3.8 (has it been fixed in later GR releases), or is this just where you have been working and the fix also should be forward-ported to `master` and `maint-3.9`

grc/gui/BlockTreeWindow.py
NeuroForLunch: I just followed the convention of whoever wrote the original. They can be replaced by meaningfully named constants but I will leave that up to others.

mormj:
@NeuroForLunch - I got a chance to cherry-pick this onto master and see the effect.  It seems that some blocks have 2 levels of documentation.  Prior to this PR, blocks that don't have the docs in the yml file will display the doxygen derived docstring.  Some blocks have the same thing for both levels of documentation.  Eliminating the second one will cause a block to show "undocumented", which is undesirable behavior. 

@777arc - any thoughts on how this feature should be reworked?

777arc:
Well, users aren't going to care whether we stuck the docs in the yaml or docstrings.  For every block we have a link to the wiki page so I don't see the reason to ever use the term undocumented, if the wiki page is a stub we'll make it obvious

mormj:
@NeuroForLunch - I'm going to go ahead and close this because as it is it's 1) against 3.8 and 2) removes valuable documentation from both the tooltips and the properties page

If you come up with a solution that can consolidate the 2 versions of docs for a block when they are the same (as shown on the tooltips now), that would be worth submitting against the master branch.

NeuroForLunch:
@mormj, thanks for finding out where the duplicates were coming from.

PR #4807, opened 2021-06-21
runtime: dereference symlink in gr::prefix() (backport to maint-3.9)

willcode:
Signed-off-by: Ed Beroset <beroset@ieee.org>
(adapted from 0f851bd06 for maint-3.9 using boost::filesystem)

Backport https://github.com/gnuradio/gnuradio/pull/4748

### DIFF
--- a/gnuradio-runtime/lib/constants.cc.in
+++ b/gnuradio-runtime/lib/constants.cc.in
@@ -28,9 +28,9 @@ const std::string prefix()
 
     boost::filesystem::path prefix_rel_lib = "@prefix_relative_to_lib@";
     boost::filesystem::path gr_runtime_lib_path = boost::dll::this_line_location();
-    // Normalize before decomposing path so result is reliable
-    boost::filesystem::path prefix_path =
-        gr_runtime_lib_path.lexically_normal().parent_path() / prefix_rel_lib;
+    // Canonize before decomposing path so result is reliable and without symlinks
+    boost::filesystem::path canonical_lib_path = boost::filesystem::canonical(gr_runtime_lib_path);
+    boost::filesystem::path prefix_path = canonical_lib_path.parent_path() / prefix_rel_lib;
     return prefix_path.lexically_normal().string();
 }
 


willcode:
@beroset @ryanvolz can you suggest a modified versions that works with boost? The default second arg for `boost::filesystem::canonical()` is `current_path()`, which seems to be causing an error.

I started looking at the following link. You might already know the answer.
https://stackoverflow.com/questions/48229510/how-to-replace-the-base-parameter-removed-from-boostfilesystemweakly-relativ

ryanvolz:
I wasn't aware of this difference with the standard library, but it looks like `weakly_canonical` would also do the job and doesn't have the second argument to gum things up. The path going to it should exist because it's coming directly from the library location, so I think there shouldn't be a difference with `canonical`.

willcode:
Giving `weakly_canonical()` a try. That sounds a lot like `partially_authoritative()`, but what ever works :-)

willcode:
@ryanvolz same with weakly_canonical()

willcode:
Executing a program from a deleted directory causes this error. Here's the result of executing just the `prefix()` code from a deleted directory in C++:
```
$ ../test
terminate called after throwing an instance of 'boost::filesystem::filesystem_error'
  what():  boost::filesystem::current_path: No such file or directory
Aborted (core dumped)
```
The QA code for `modutil` uses `shutil.rmtree()`, so I'm guessing that there's a race with either the `run()` in the individual tests or with the beginning of the subsequent test. The error doesn't happen with `master` which has identical `modtool` code, though.

Another possibility is that `master` includes a mutex for prefs, which is used in this code.

@marcusmueller maybe you have an idea what's happening here?

ryanvolz:
If `boost::dll::this_line_location()` always returns an absolute path (I haven't found anything that confirms this, but maybe?), then we could probably get around this by passing any absolute path (say "/") as the `base` argument to `canonical` to avoid the call to `current_path`. I'm guessing that's how it's working on master, because there is no `base` argument and it never calls `current_path` because it already finds that the path is absolute.

willcode:
Hmm, it doesn't work the same either. This could be some difference between `boost` and `std`, or something else. Afraid I'm going to have to give up on this one for now. Too many mysteries.
```
[39 gnuradio]$ gnuradio-config-info --prefix
/home/user/gr39/.
```
```[ma grmaster]$ gnuradio-config-info --prefix
/home/user/grmaster/
```

ryanvolz:
> Hmm, it doesn't work the same either. This could be some difference between `boost` and `std`, or something else. Afraid I'm going to have to give up on this one for now. Too many mysteries.
> 
> ```
> [39 gnuradio]$ gnuradio-config-info --prefix
> /home/user/gr39/.
> ```
> 
> ```
> [ma grmaster]$ gnuradio-config-info --prefix
> /home/user/grmaster/
> ```

The dot at the end? That's not new for the boost version, AFAIK, and it represents the same thing.

willcode:
@ryanvolz Commenting out this line https://github.com/gnuradio/gnuradio/blob/078d40ddeab957fcc91060131a4b3a1dfe9df676/gr-utils/modtool/tests/test_modtool.py#L84 makes the test pass. So, current_path in boost requires a dir to still exists, where in std it doesn't? I'm hesitant to mess with the test too much because someone could write a script that uses modtool and does essentially the same thing. Hope this gives you a clue to work with.

willcode:
New version of PR (#4889) merged.

PR #4808, opened 2021-06-23
grc: introduce gui_hint for cpp code generation

dl1ksv:
At the moment some qtgui windows do support only very simple gui hints, as
the parse_gui_hint in param.py generates only python code independently of the output language.

This is fixed here, so the gui_hint() can be used in cpp code generation.

To test more complex gui hints the qtgui_tab_widget was extended to cpp.

After that compiling the generated cpp code for some examples led to

```
/usr/include/python3.9/object.h:206:23: Fehler: expected unqualified-id before »;« token
 206 |     PyType_Slot *slots; /* terminated by slot==0. */
```
This is a known error and can be fixed by

```
#ifdef ENABLE_PYTHON
#pragma push_macro("slots")
#undef slots
#include "Python.h"
#pragma pop_macro("slots")
#endif
```

instead of simply

```
#ifdef ENABLE_PYTHON
#include "Python.h"
#endif
```

which was applied to the corresponding gr-qtgui header files.

### DIFF
--- a/grc/core/generator/cpp_top_block.py
+++ b/grc/core/generator/cpp_top_block.py
@@ -222,9 +222,9 @@ class CppTopBlockGenerator(TopBlockGenerator):
         fg = self._flow_graph
         parameters = fg.get_parameters()
 
-        # List of blocks not including variables and imports and parameters and disabled
+        # List of blocks not including variables and parameters and disabled
         def _get_block_sort_text(block):
-            code = block.cpp_templates.render('make').replace(block.name, ' ')
+            code = block.cpp_templates.render('declarations')
             try:
                 code += block.params['gui_hint'].get_value()  # Newer gui markup w/ qtgui
             except:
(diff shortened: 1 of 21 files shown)


dl1ksv:
Here is just an example for testing
[htest_cpp.grc.gz](https://github.com/gnuradio/gnuradio/files/6703086/htest_cpp.grc.gz)

mormj (approved):
I'm not hugely familiar with the cpp generation code, but this all seems to make sense.  Also, tested and got the example running.  Thanks for providing this example.  Will leave this up through the end of the week in case anyone else wants to review, but the impact seems localized to cpp generation.

grc/core/generator/cpp_top_block.py
```diff
@@ -222,9 +222,9 @@ def _blocks(self):
         fg = self._flow_graph
         parameters = fg.get_parameters()
 
-        # List of blocks not including variables and imports and parameters and disabled
+        # List of blocks not including variables and parameters and disabled
         def _get_block_sort_text(block):
-            code = block.cpp_templates.render('make').replace(block.name, ' ')
+            code = block.cpp_templates.render('declarations')
```
haakov, line 227: Are we sure this does not break something else?

dl1ksv: This change only affects the order of generating the cpp code. Without this change sometimes the order how cpp code is generated was wrong.

PR #4813, opened 2021-06-24
add homebrew installation on readme

NeuroForLunch:
I added all the necessary dependencies on homebrew and now it installs perfectly with version 3.9.2 and includes gr_soapy and the prerequisites of soapysdr and soapyrtlsdr.

Signed-off-by: NeuroForLunch <neuro.for.lunch@gmail.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,21 +8,21 @@
 [![Docs](https://img.shields.io/badge/docs-doxygen-orange.svg)](https://www.gnuradio.org/doc/doxygen/)
 [![Packaging status](https://repology.org/badge/tiny-repos/gnuradio.svg)](https://repology.org/project/gnuradio/badges)
 
-GNU Radio is a free & open-source software development toolkit that 
-provides signal processing blocks to implement software radios. It can 
-be used with readily-available, low-cost external RF hardware to create 
-software-defined radios, or without hardware in a simulation-like 
-environment. It is widely used in hobbyist, academic, and commercial 
-environments to support both wireless communications research and real-world 
+GNU Radio is a free & open-source software development toolkit that
+provides signal processing blocks to implement software radios. It can
+be used with readily-available, low-cost external RF hardware to create
+software-defined radios, or without hardware in a simulation-like
+environment. It is widely used in hobbyist, academic, and commercial
+environments to support both wireless communications research and real-world
 radio systems.
 
-Please visit the GNU Radio website at https://www.gnuradio.org/ and the 
-wiki at https://wiki.gnuradio.org/. Bugs and feature requests are 
-tracked on GitHub's [Issue Tracker](https://github.com/gnuradio/gnuradio/issues). 
-If you have questions about GNU Radio, please search the **discuss-gnuradio** 
-mailing list [archive](https://lists.gnu.org/archive/html/discuss-gnuradio/), 
-as many questions have already been asked and answered. Please also 
-[subscribe](https://lists.gnu.org/mailman/listinfo/discuss-gnuradio) to 
+Please visit the GNU Radio website at https://www.gnuradio.org/ and the
+wiki at https://wiki.gnuradio.org/. Bugs and feature requests are
+tracked on GitHub's [Issue Tracker](https://github.com/gnuradio/gnuradio/issues).
+If you have questions about GNU Radio, please search the **discuss-gnuradio**
+mailing list [archive](https://lists.gnu.org/archive/html/discuss-gnuradio/),
+as many questions have already been asked and answered. Please also
+[subscribe](https://lists.gnu.org/mailman/listinfo/discuss-gnuradio) to
 the mailing list and post your new questions there.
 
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -8,34 +8,38 @@
...
 ### Prebuilt Binaries
 
-The recommended way to install GNU Radio on most platforms is using available binary package distributions. 
+The recommended way to install GNU Radio on most platforms is using available binary package distributions.
 
 The following command is for Debian, Ubuntu, and derivatives. Consult your distribution information to obtain the version of GNU Radio which is included.
 
     sudo apt install gnuradio
 
+For MacOS, install the latest release version of GNUradio with [homebrew](https://brew.sh)
```
marcusmueller, line 39: Wrong spelling/capitalization of "GNU Radio"

marcusmueller:
Either we introduce the preferred installation method for all platforms (is that really homebrew on OS X? I honestly don't know!), or we try to keep the Readme as minimal as possible and point to the wiki; honestly, not too sure adding homebrew here is a great idea. Either the reader is already familiar with homebrew, in which case they probably tried what's recommended, or they should really be reading the wiki.

NeuroForLunch:
I tried installing both 3.8 and 3.9 without homebrew and gave up after over a hundred hours. "Read the wiki" is no excuse for making the installation incredibly difficult and time consuming. Even pybombs doesn't work correctly on MacOS or Docker using images that other people have on github. The homebrew installation provides pre-compiled binaries and sets up the user for instant usability. Compiling v3.8 and having it fail after waiting for an hour because of a Doxygen error is not comprehensible to users who are wanting to use the software and failing to install it because of a code error in the documentation. 

I have yet to see a single pybombs Dockerfile that works (github will show ✅) so you might consider removing that section unless pybombs are fixed and work _reproducibly._  https://github.com/gnuradio/pybombs/issues shows plenty of unresolved issues on multiple OS's.

Here is how popular homebrew is vs the other installers:
- https://www.slant.co/versus/1588/1674/~macports_vs_homebrew
- https://www.scivision.dev/homebrew-macports-fink/

README.md
NeuroForLunch: fixed it

777arc (commented):
I'm not a macOS user but it sounds like we should be really thankful for @NeuroForLunch taking the time to get a working homebrew recipe. I know we hate to have preferred install methods but if homebrew on macOS is what works then I'm fine with calling that the preferred method, although we should get some of our macOS-user-coredevs to double check before adding it to such a prominent spot in the README.  And either way, the macOS part of the wiki definitely needs to get updated because some fraction of people will be using that as a guide https://wiki.gnuradio.org/index.php/MacInstall

As far as it going into the README, there's no reason it absolutely has to, we were trying to make the README concise and the only reason it mentions the "sudo apt install gnuradio" is because 90-some-% of the folks reading it are on Ubuntu and derivatives because it's the recommended OS for people new to GR, we are very clear with the "For other operating systems and versions", it's easy enough to see and click, and those OS-specific install pages are actively maintained.  The best reason to NOT include it in the README is because it will lead people to ask why we aren't mentioning the Windows version, or the CentOS/RHEL version, etc.

NeuroForLunch:
@777arc I think there should be concise installation instructions for the major operating systems in the readme and then extended installation instructions for non-standard installs in the wiki. https://wiki.gnuradio.org/index.php/MacInstall shows MacPorts as the recommended install but that install might be broken and it has only 11 installs vs 980 for homebrew so it doesn't seem like most people are following the recommendation. Also, to recommend a new OS just to install a program and expect people to actually follow through with that is a big leap. It's easier to pass up the program and just find something else than install a new OS. If anyone has the stats on install numbers per OS and if they used PyBombs or not, it would be great to see because that would make it a lot easier to decide which installs to put on the readme.

777arc:
> Also, to recommend a new OS just to install a program and expect people to actually follow through with that is a big leap.

Let me give an example to provide perspective.  So for the class I teach that involves GNU Radio, I actually have my students install Ubuntu 20 in a VM (assuming they don't have a debian-based distro installed natively) and it's not because you can't install GNU Radio on macOS or Windows, it's because I don't want them wasting their own time and my time when they are unable to build and install gr-xyz, i.e. some random out of tree module.  And it also helps to have 1 set of instructions/commands for doing stuff.  But my point is, there's reasons that might not be obviously apparent why we recommend just going with something like Ubuntu 20 if you want to dive deep into GNU Radio.  It sucks that we have to do that, but we can't ensure all out of tree modules are packages or buildable in every OS.  And GNU Radio is all about the out of tree modules if you haven't realized already.  This is all my personal opinion, others can feel free to chime in.

NeuroForLunch:
@777arc, thank you for your perspective. It makes perfect sense in a class scenario but if there are ~30 individuals a day installing GNU Radio via homebrew, then according to the 90% figure there's ~300 individuals installing it per day on linux. I don't know what the ratio of individuals to students are but I would assume most are not students. I would just like to make this software more accessible to people because currently it is very difficult to install and giving up "because I don't want them wasting their own time and my time when they are unable to build and install gr-xyz, i.e. some random out of tree module." I also think the modules should be added to homebrew, etc, so they can just be installed via the same install route and guaranteed to work.

777arc:
Regardless of updating the README, someone should probably overhaul this heavily trafficked page https://wiki.gnuradio.org/index.php/MacInstall I would but I don't have a macOS machine so it's out of my comfort zone

mormj:
I agree with @marcusmueller and @777arc - let's keep the specific methods updated on the wiki - and that would be a great place do document the excellent work toward a working homebrew recipe.  Closing this PR in favor of the wiki

marcusmueller:
Hey, @NeuroForLunch, by the way, I'm really really really super thankful for anyone making it easier to install GNU Radio, and I'm happy you've done that work for homebrew!

Issue #4174, opened 2021-01-31
GRC freezes when searching for components (OsX, BigSur)


mkierczak:
I have compiled from sources using Homebrew (OsX, BigSur):
`brew install --build-from-source gnuradio`. 

In general, it works, I can sometimes run a simple workflow, but GRC behaves erratic and freezes when I search for components, specifically, when I try to delete current search string. Sometimes "reload blocks" helps, sometimes not. Always with the following error message:
```
2021-01-31 20:36:54.225 Python[53213:865823] unlockFocus called too many times. Called on <GdkQuartzView: 0x7fcb4246e060>. 
```

sam-k:
Same issue here. GRC is unusable at this point on macOS Big Sur; nothing is responsive.

jackdally:
I am also running into the same issue. It freezes the GUI completely.

NeuroForLunch:
I found the answer in the source code - the GUI is limited to some maximum pixel size so if you are attempting to do full screen on a macbook it will freeze. On a macbook retina 15" the GUI has to be no more than 3/4 of the screen (maybe even smaller).

dkozel:
@NeuroForLunch Where is this limitation in the code. Clearly this would be excellent to fix.

NeuroForLunch:
I am unsure of the exact line of code that was causing the issue but I have narrowed it down to VariableEditor.py, Constants.py, Config.py, flowgraph.py, and block.py. Some of these set a default value but possibly limit the maximum size by only allowing some multiple which is still less than the resolution on a 15" mac which is 2880x1600.

NeuroForLunch:
Here is a video example of the GUI behavior. You'll see that at a certain point of shrinking the window it just clicks in and starts working normally. 

https://www.youtube.com/watch?v=eVMLWy6u_hM (HD is still processing so check back tomorrow)

NeuroForLunch:
The fix is in: #4814

ryanvolz:
Has this been fixed in some other way, since #4814 is closed, or does it still occur? It would be great to fix this before the next release so I don't have to end up carrying patches in conda.

michaelld:
Thinking NO, and also that your fix from #4814 should do the trick. Have you made a PR for that yet?

ryanvolz:
[...]

PR #4814, opened 2021-06-24
grc-mac-freeze-fix

NeuroForLunch:
No more freezing on high dpi displays up to 2.0x DPI

### DIFF
--- a/grc/gui/DrawingArea.py
+++ b/grc/gui/DrawingArea.py
@@ -154,7 +154,7 @@ class DrawingArea(Gtk.DrawingArea):
 
     def _update_size(self):
         w, h = self._flow_graph.get_extents()[2:]
-        self.set_size_request(w * self.zoom_factor + 100, h * self.zoom_factor + 100)
+        self.set_size_request(w * 2 * self.zoom_factor + 100, h * 2 * self.zoom_factor + 100)
 
     def _auto_scroll(self, event):
         x, y = event.x, event.y


mormj:
@NeuroForLunch - thank you for the contribution.  Would you mind doing a couple of minor things:
1) In the commit message, prepend with the module that is affected - in this case `grc: high dpi freeze fix for max` or something to that effect
2) Provide a comment in the code about why that `2 * ` is there?  I haven't tested the effect on mac (or non-mac) nor do i understand why this would be mac specific

NeuroForLunch:
When the horizontal scroll bar disappears, GRC freezes. Not sure why but this is the problem on macs. It might not have been freezing on high dpi displays on linux without this.

advorkovyy:
This fix does not work for me. Applied the change, rebuilt from sources - no luck. Still freezes in full-screen and works well with smaller window size.

NeuroForLunch:
@advorkovyy please include a screenshot at fullscreen size when its frozen

advorkovyy:
Screenshot 2021-07-12 at 22 56 31

advorkovyy:
with a little bit less width it does not freeze.

Do not mind the grc version - I've applied your fix directly in the code, uninstalled, removed build directory, built again, installed.

```
➜ git status
HEAD detached at v3.9.2.0
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
	modified:   grc/gui/DrawingArea.py

no changes added to commit (use "git add" and/or "git commit -a")
```

NeuroForLunch:
Okay so you would have to change the number to multiply by to 2.26 instead of 2 (you have 226 dpi). I guess I need to check what the highest DPI setting for a screen is. The freeze occurs when the distance from the rightmost node to the vertical list is too far making the horizontal scroll bar disappear so the multiplier just adds in extra space on the canvas to keep the scroll bar there.

advorkovyy:
Yes, it works now with 2.26! I will test it later today with external display

dkozel:
I think this function is the correct one to read the active value for the device to application pixel scaling.
https://developer.gnome.org/gdk3/stable/GdkMonitor.html#gdk-monitor-get-scale-factor

mormj:
@NeuroForLunch - does it seem reasonable to replace the hardcode `2` or `2.5` with a dynamically read value from the gdk API?

NeuroForLunch:
> @NeuroForLunch - does it seem reasonable to replace the hardcode `2` or `2.5` with a dynamically read value from the gdk API?

Seems like that would be best

ryanvolz:
I just tested this out with the user who reported https://github.com/ryanvolz/radioconda/issues/3, and it fixes #5085 which then seems to be the same issue as #4174. Has anyone done further work to make this into a proper fix with the suggested call to the gdk API?

ryanvolz:
I think adding the `scale_factor` in `grc/gui/DrawingArea.py` like this:

```
    def _update_size(self):
        w, h = self._flow_graph.get_extents()[2:]
        scale_factor = self.get_scale_factor()
        self.set_size_request(w * scale_factor * self.zoom_factor + 100, h * scale_factor * self.zoom_factor + 100)
```
will do the trick? Basically taking @NeuroForLunch's PR here and substituting `self.get_scale_factor()` for `2`. I can put in a PR to that effect. `self` is a `DrawingArea` which is a `Gtk.Widget`: https://lazka.github.io/pgi-docs/Gtk-3.0/classes/Widget.html#Gtk.Widget.get_scale_factor

Anyone with more GRC dev experience: is that the only and correct place to insert a DPI scale factor?

michaelld:
@ryanvolz did you make a PR with this latest fix? I think what you note does the trick in a nice generic fashion!

dkozel:
@ryanvolz Did you do functional tests against a macOS device with a high dpi display? I guessed on that get_scale_factor function, great if I was correct. Or @michaelld did you test with that patch?

ryanvolz:
I thought I'd gotten some positive results on that patch with a user who saw #5085, but that turned out to be fools gold. In that testing, we saw that `self.get_scale_factor()` was actually returning 1, but I don't know if that's because the issue that he was seeing is actually different and so it might still return 2 or whatever for someone experiencing this crash.

Basically, it would be good to get a report from some other user that my proposed patch actually does something to fix something, and then I'd be happy to do the PR. Because from what I know for sure right now, it's not useful.

dkozel:
Since this seems plausibly like a high resolution screen issue here's the Qt documentation page that discusses a few things to try/change.
https://doc.qt.io/qt-5/highdpi.html

geraman21:
@ryanvolz @dkozel 
I have tested the fix with self.get_scale_factor() on 16' MBP 2019 and it worked ! I can resize GRC window with no issues now. Thank you guys!

ryanvolz:
Alright, I made a new PR: #5111.

PR #4823, opened 2021-06-27
iio: Remove cstdio/iostream and log using GR_LOG

Yamakaja:
This commit cleans up unused headers (cstdio) and replaces all instances
of std::cout and std::cerr with the corresponding GR_LOG_* invocation.

Additionally, a helper is added to gr/lib/logger.cc to facilitate the
allocation of loggers which are automatically configured according to the
gnuradio configuration settings.

(I had some issues getting the logger to work in static contexts (As such no access to d_logger), so i created a new helper that does the logger configuration for me. I'm still not sure if that should be something global, but it is certainly not very iio specific.)

### DIFF
--- a/gnuradio-runtime/include/gnuradio/logger.h
+++ b/gnuradio-runtime/include/gnuradio/logger.h
@@ -61,12 +61,18 @@ typedef log4cpp::Category* logger_ptr;
 
 #define GR_LOG_ASSIGN_LOGPTR(logger, name) logger = gr::logger_get_logger(name)
 
+#define GR_LOG_ASSIGN_CONFIGURED_LOGPTR(logger, name) \
+    logger = gr::logger_get_configured_logger(name)
+
 #define GR_CONFIG_LOGGER(config) gr::logger_config::load_config(config)
 
 #define GR_CONFIG_AND_WATCH_LOGGER(config, period) \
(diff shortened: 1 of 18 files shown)


mormj:
@Yamakaja - thank you for the fixes here and additional logger methods - one minor thing that would help here is separate out the changes to runtime from the changes to gr-iio in separate commits to keep the commit history more specific.

Yamakaja:
@mormj Ah good point, i agree that the gr part should be in a separate commit!

Yamakaja:
V2:
* Split original PR into two commits
* Replaces duplicated code in gr-fec with new helper (See gr-fec [exhibit a](https://github.com/gnuradio/gnuradio/blob/master/gr-fec/lib/generic_encoder.cc#L27-L46), [exhibit b](https://github.com/gnuradio/gnuradio/blob/master/gr-fec/lib/generic_decoder.cc#L27-L46) vs [logger.cc](https://github.com/gnuradio/gnuradio/blob/master/gnuradio-runtime/lib/logger.cc#L328-L368))

Yamakaja:
V3:
* Adds `GR_LOG_GET_CONFIGURED_LOGGER` and `GR_LOG_ASSIGN_CONFIGURED_LOGPTR` macros.

mormj (approved):

PR #4830, opened 2021-07-01
grc: only rewrite a block if the block is enabled.

### LINKED ISSUE #4788: Flowgraph doesn't update disabled blocks properly if samp_rate is controlled by another variable

#### Expected behavior

#### Actual behavior

dl1ksv:
This fixes #4788. If the disabled block will be reenabled the block is rewriten.

### DIFF
--- a/grc/core/blocks/block.py
+++ b/grc/core/blocks/block.py
@@ -106,7 +106,8 @@ class Block(Element):
         """
         Add and remove ports to adjust for the nports.
         """
-        Element.rewrite(self)
+        if self.enabled:
+            Element.rewrite(self)
 
         def rekey(ports):
             """Renumber non-message/message ports"""


drmpeg:
Pretty ugly side effect with a disabled QT GUI block.

image

dl1ksv:
If you prefer another representation, why don't you the disabled block to bypassed instead?

drmpeg:
I would prefer that disabled blocks look just like enabled blocks. I'm not sure that can be attained, but this PR is definitely a step backwards.

drmpeg:
I would prefer that disabled blocks look and act  just like enabled blocks. I'm not sure that can be attained, but this PR is definitely a step backwards.

dl1ksv:
Well, to answer my question by myself: You can't bypass a bock that has only an input but no output.

PR #4835, opened 2021-07-05
soapy: added message support for SoapySDR 0.8 API

ncorgan:
* Moved string <-> setting conversion to common header
* Closes https://github.com/gnuradio/gnuradio/issues/4571

### DIFF
--- a/gr-soapy/lib/block_impl.cc
+++ b/gr-soapy/lib/block_impl.cc
@@ -11,6 +11,7 @@
 #endif
 
 #include "block_impl.h"
+#include "setting_string_conversion.h"
 #include <gnuradio/io_signature.h>
 #include <SoapySDR/Formats.h>
 #include <SoapySDR/Version.hpp>
@@ -106,13 +107,43 @@ static void check_abi(void)
     }
 }
 
-// TODO: messages for full 0.8 support
 const pmt::pmt_t CMD_CHAN_KEY = pmt::mp("chan");
 const pmt::pmt_t CMD_FREQ_KEY = pmt::mp("freq");
 const pmt::pmt_t CMD_GAIN_KEY = pmt::mp("gain");
 const pmt::pmt_t CMD_ANTENNA_KEY = pmt::mp("antenna");
 const pmt::pmt_t CMD_RATE_KEY = pmt::mp("rate");
 const pmt::pmt_t CMD_BW_KEY = pmt::mp("bandwidth");
+const pmt::pmt_t CMD_GAINMODE_KEY = pmt::mp("gain_mode");
+const pmt::pmt_t CMD_FREQCORR_KEY = pmt::mp("freq_corr");
+const pmt::pmt_t CMD_DCOFFSETMODE_KEY = pmt::mp("dc_offset_mode");
+const pmt::pmt_t CMD_DCOFFSET_KEY = pmt::mp("dc_offset");
+const pmt::pmt_t CMD_IQBAL_KEY = pmt::mp("iq_bal");
+const pmt::pmt_t CMD_IQBALMODE_KEY = pmt::mp("iq_bal_mode");
+const pmt::pmt_t CMD_MCR_KEY = pmt::mp("master_clock_rate");
+const pmt::pmt_t CMD_REFCLOCKRATE_KEY = pmt::mp("ref_clock_rate");
+const pmt::pmt_t CMD_CLOCKSRC_KEY = pmt::mp("clock_src");
+const pmt::pmt_t CMD_TIMESRC_KEY = pmt::mp("time_src");
+const pmt::pmt_t CMD_HWTIME_KEY = pmt::mp("hw_time");
+const pmt::pmt_t CMD_REG_KEY = pmt::mp("reg");
+const pmt::pmt_t CMD_REGS_KEY = pmt::mp("regs");
+const pmt::pmt_t CMD_SETTING_KEY = pmt::mp("setting");
+const pmt::pmt_t CMD_GPIO_KEY = pmt::mp("gpio");
+const pmt::pmt_t CMD_GPIODIR_KEY = pmt::mp("gpio_dir");
+const pmt::pmt_t CMD_I2C_KEY = pmt::mp("i2c");
+const pmt::pmt_t CMD_UART_KEY = pmt::mp("uart");
+
+const pmt::pmt_t CMD_NAME_KEY = pmt::mp("name");
+const pmt::pmt_t CMD_KEY_KEY = pmt::mp("key");
+const pmt::pmt_t CMD_VALUE_KEY = pmt::mp("value");
+const pmt::pmt_t CMD_ADDR_KEY = pmt::mp("addr");
+const pmt::pmt_t CMD_MASK_KEY = pmt::mp("mask");
+const pmt::pmt_t CMD_TIME_KEY = pmt::mp("time");
+const pmt::pmt_t CMD_BANK_KEY = pmt::mp("bank");
+const pmt::pmt_t CMD_DATA_KEY = pmt::mp("data");
+
+// Common default values for pmt::dict_ref
+const pmt::pmt_t PMT_EMPTYSTR = pmt::mp("");
+const pmt::pmt_t PMT_ZERO = pmt::mp(0);
 
 /*
  * The private constructor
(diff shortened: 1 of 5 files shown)


gr-soapy/lib/block_impl.cc
```diff
@@ -113,6 +114,33 @@ const pmt::pmt_t CMD_GAIN_KEY = pmt::mp("gain");
 const pmt::pmt_t CMD_ANTENNA_KEY = pmt::mp("antenna");
 const pmt::pmt_t CMD_RATE_KEY = pmt::mp("rate");
 const pmt::pmt_t CMD_BW_KEY = pmt::mp("bandwidth");
+const pmt::pmt_t CMD_GAINMODE_KEY = pmt::mp("gain_mode");
+const pmt::pmt_t CMD_FREQCORR_KEY = pmt::mp("freq_corr");
+const pmt::pmt_t CMD_DCOFFSETMODE_KEY = pmt::mp("dc_offset_mode");
+const pmt::pmt_t CMD_DCOFFSET_KEY = pmt::mp("dc_offset");
+const pmt::pmt_t CMD_IQBAL_KEY = pmt::mp("iq_bal");
+const pmt::pmt_t CMD_IQBALMODE_KEY = pmt::mp("iq_bal_mode");
+const pmt::pmt_t CMD_MCR_KEY = pmt::mp("mcr");
```
willcode, line 123: Can this be master_clock_rate? Same for any mcr below.

gr-soapy/lib/block_impl.cc
```diff
@@ -1193,18 +1298,292 @@ void block_impl::cmd_handler_bw(pmt::pmt_t val, size_t channel)
...
+void block_impl::cmd_handler_time_source(pmt::pmt_t val, size_t)
+{
+    if (!val->is_symbol()) {
+        GR_LOG_ERROR(d_logger, "soapy: time source must be string");
+        return;
+    }
+    set_time_source(pmt::symbol_to_string(val));
+}
+
+void block_impl::cmd_handler_hardware_time(pmt::pmt_t val, size_t)
```
willcode, line 1397: Why a dict here vs uint64_t?

gr-soapy/python/soapy/bindings/CMakeLists.txt
```diff
@@ -11,6 +11,9 @@ include(GrPybind)
 # Python Bindings
 ########################################################################
 
+# For setting string conversions
```
willcode, line 14: Not sure, but it seems like this is referencing an include file in a lib (private) dir as part of the public API?

willcode (commented):
There is a lot of new API here, so let's make sure a couple other people look at it too. GR messages have been pretty ad-hoc in the past and we should make sure they're intuitive (as possible) to the user.

gr-soapy/lib/block_impl.cc
```diff
@@ -1157,25 +1244,43 @@ std::string block_impl::read_uart(const std::string& which, long timeout_us) con
...
+    if (val->is_dict()) {
+        if (!pmt::dict_has_key(val, CMD_GAIN_KEY)) {
+            GR_LOG_ERROR(d_logger, "soapy: gain dict must contain key \"gain\"");
+            return;
+        }
+
+        // Accept no name, falls back to default argument ""
+        const auto name =
+            pmt::symbol_to_string(pmt::dict_ref(val, CMD_NAME_KEY, pmt::mp("")));
+        const auto gain = pmt::to_double(pmt::dict_ref(val, CMD_GAIN_KEY, 0));
```
marcusmueller, line 1272: uhhh I think this casts `0` to a shmart pointer? do we really intend to do that?

marcusmueller: `dict_ref(const pmt_t&, const pmt_t&, const pmt_t&)`: I wouldn't know how to convert 0 to a `const pmt_t==const std::shared_ptr<pmt_base>`

gr-soapy/lib/block_impl.cc
```diff
@@ -1157,25 +1244,43 @@ std::string block_impl::read_uart(const std::string& which, long timeout_us) con
...
+
+    if (val->is_dict()) {
+        if (!pmt::dict_has_key(val, CMD_GAIN_KEY)) {
+            GR_LOG_ERROR(d_logger, "soapy: gain dict must contain key \"gain\"");
+            return;
+        }
+
+        // Accept no name, falls back to default argument ""
+        const auto name =
+            pmt::symbol_to_string(pmt::dict_ref(val, CMD_NAME_KEY, pmt::mp("")));
```
marcusmueller, line 1271: we're not expecting this to be a high-rate call, right? Because this should call `pmt::mp("")` every time, and that in turn hashes the empty string, does a hash lookup in the map of all known symbols, finds the previously register PMT symbol... for that then to be thrown away.

gr-soapy/lib/block_impl.cc
```diff
@@ -1193,18 +1298,292 @@ void block_impl::cmd_handler_bw(pmt::pmt_t val, size_t channel)
...
+        !pmt::dict_has_key(val, CMD_VALUE_KEY)) {
+        GR_LOG_ERROR(d_logger,
+                     "soapy: multi-register write dict must contain keys \"name\", "
+                     "\"addr\", \"value\"");
+        return;
+    }
+
+    const auto name =
+        pmt::symbol_to_string(pmt::dict_ref(val, CMD_NAME_KEY, pmt::mp("")));
+    const auto addr = pmt::to_uint64(pmt::dict_ref(val, CMD_ADDR_KEY, 0));
```
marcusmueller, line 1457: see discussion above

gr-soapy/python/soapy/bindings/soapy_common.cc
```diff
@@ -7,97 +7,29 @@
  *
  */
 
+#include "setting_string_conversion.h"
 #include "soapy_common.h"
 
 #include <SoapySDR/Types.hpp>
 #include <SoapySDR/Version.h>
 #include <stdexcept>
```
marcusmueller, line 15: Not used any more, right? Can be removed.

gr-soapy/python/soapy/bindings/CMakeLists.txt
ncorgan: This points to the string <-> setting conversion code we went back and forth on, which this commit moves to the lib directory to be used there too.

gr-soapy/lib/block_impl.cc
ncorgan: https://github.com/gnuradio/gnuradio/blob/master/gr-soapy/include/gnuradio/soapy/block.h#L466

At the bottom, the handler calls into this function, which optionally takes in a name. I need a dict to take both the name and time.

gr-soapy/lib/setting_string_conversion.h
```diff
@@ -0,0 +1,91 @@
+/*
+ * Copyright 2021 Free Software Foundation, Inc.
```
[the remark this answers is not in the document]
ncorgan: This was copy+paste from some other file, not really a conscious decision on my part. For consistency, I'll switch it over.

gr-soapy/python/soapy/bindings/soapy_common.cc
ncorgan: This is needed because the functions were moved out of this file. They're still called into below.

gr-soapy/lib/block_impl.cc
ncorgan: Yay implicit casts. That should be fixed over in gr-uhd as well, and who knows where else:

https://github.com/gnuradio/gnuradio/blob/master/gr-uhd/lib/usrp_block_impl.cc#L731

gr-soapy/lib/block_impl.cc
ncorgan: I'll make a single instance up top with the keys and use that throughout.

marcusmueller: hey, that was an honest question about the rate. We shan't care if that happens 50 times a second!

You could do that static local or memer or if it's easier to read, split things to (attention, code that I'm not sure is prettier ahead)

```
auto& tmp = pmt::dict_ref(val, key, pmt::PMT_NIL);
std::string name = (tmp == PMT_NIL) ? "" : pmt::symbol_to_string(tmp);
```

or  so.

gr-soapy/lib/block_impl.cc
marcusmueller: thanks!

gr-soapy/python/soapy/bindings/soapy_common.cc
marcusmueller: I mean the `stdexcept`; the code that needed that was moved to setting_string_conversion.h, but that doesn't include it (it does, indirectly through Soapy, though).

ncorgan: Fixed

gr-soapy/lib/block_impl.cc
```diff
@@ -106,13 +107,43 @@ static void check_abi(void)
...
 const pmt::pmt_t CMD_ANTENNA_KEY = pmt::mp("antenna");
 const pmt::pmt_t CMD_RATE_KEY = pmt::mp("rate");
 const pmt::pmt_t CMD_BW_KEY = pmt::mp("bandwidth");
+const pmt::pmt_t CMD_GAINMODE_KEY = pmt::mp("gain_mode");
+const pmt::pmt_t CMD_FREQCORR_KEY = pmt::mp("freq_corr");
+const pmt::pmt_t CMD_DCOFFSETMODE_KEY = pmt::mp("dc_offset_mode");
+const pmt::pmt_t CMD_DCOFFSET_KEY = pmt::mp("dc_offset");
+const pmt::pmt_t CMD_IQBAL_KEY = pmt::mp("iq_bal");
+const pmt::pmt_t CMD_IQBALMODE_KEY = pmt::mp("iq_bal_mode");
+const pmt::pmt_t CMD_MCR_KEY = pmt::mp("master_clock_rate");
```
willcode, line 122: `CMD_MASTERCLOCKRATE_KEY`?

marcusmueller (approved):

marcusmueller:
By the way: I love this, thanks!

willcode (approved):
Let's leave this up for a couple more days to see if anyone else has comments. Thanks.

mormj (approved):

mormj:
Been here for a week - 3 reviews.  Merging.

PR #4837, opened 2021-07-05
Corrected spacings and updated String.split() usage

jfmadeira:
Greetings, 

I've been using custom LDPC matrices and noticed the helper python scripts needed some tweaking to output the preprocessed matrix. The scripts were tested with python3 and the resulting preprocessed alist file works correctly with the LDPC encoder and decoder blocks.

### DIFF
--- a/gr-fec/python/fec/LDPC/Generate_LDPC_matrix_functions.py
+++ b/gr-fec/python/fec/LDPC/Generate_LDPC_matrix_functions.py
@@ -27,12 +27,12 @@ def read_alist_file(filename):
 
   myfile    = open(filename,'r')
   data      = myfile.readlines()
-  size      = string.split(data[0])
+  size      = data[0].split()
   numCols   = int(size[0])
   numRows   = int(size[1])
   H = zeros((numRows,numCols))
   for lineNumber in arange(4,4+numCols):
-    indices = string.split(data[lineNumber])
+    indices = data[lineNumber].split()
     for index in indices:
       H[int(index)-1,lineNumber-4] = 1
       # The subsequent lines in the file list the indices for where
@@ -70,12 +70,12 @@ def write_alist_file(filename, H, verbose=0):
     rowWeight = nonzeros.shape[1]
     if rowWeight > maxRowWeight:
       maxRowWeight = rowWeight
-      tempstring1 = tempstring1 + repr(rowWeight) + ' '
+    tempstring1 = tempstring1 + repr(rowWeight) + ' '
     for tempArray in nonzeros:
       for index in tempArray:
         tempstring2 = tempstring2 + repr(index+1) + ' '
-        tempstring2 = tempstring2 + '\n'
-        tempstring1 = tempstring1 + '\n'
+    tempstring2 = tempstring2 + '\n'
+  tempstring1 = tempstring1 + '\n'
 
   tempstring3 = ''
   tempstring4 = ''
@@ -85,12 +85,12 @@ def write_alist_file(filename, H, verbose=0):
     colWeight = nonzeros.shape[1]
     if colWeight > maxColWeight:
       maxColWeight = colWeight
-      tempstring3 = tempstring3 + repr(colWeight) + ' '
+    tempstring3 = tempstring3 + repr(colWeight) + ' '
     for tempArray in nonzeros:
       for index in tempArray:
         tempstring4 = tempstring4 + repr(index+1) + ' '
-        tempstring4 = tempstring4 + '\n'
-        tempstring3 = tempstring3 + '\n'
+    tempstring4 = tempstring4 + '\n'
+  tempstring3 = tempstring3 + '\n'
 
   tempstring = repr(maxColWeight) + ' ' + repr(maxRowWeight) + '\n'
   # write out max column and row weights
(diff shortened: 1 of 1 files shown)


willcode:
What was the effect of the bad indentation? Was the generated matrix wrong?

jfmadeira:
Some of the functions were incorrectly indented to be inside the verbose if's, which made it not calculate the matrix properly unless verbose was active. There were also several lines incorrectly indented in the write_alist_file function that caused the alist file to be incorrectly filled (i.e. putting \n instead of spaces between the number of ones per column).

gr-fec/python/fec/LDPC/Generate_LDPC_matrix_functions.py
```diff
@@ -27,12 +27,12 @@ def read_alist_file(filename):
 
   myfile    = open(filename,'r')
   data      = myfile.readlines()
-  size      = string.split(data[0])
+  size      = data[0].split()
   numCols   = int(size[0])
   numRows   = int(size[1])
```
jdemel, line 32: I'd say this change is already an improvement. While you're at it, i suggest to restructure this as:
```python
with open(filename, 'r') as myfile:
    numCols, numRows = parse_alist_header(data[0])
```
This should avoid errors in case smth went wrong with the filename. Also, it makes the code more verbose.

Same goes for the write alist file function that tries to do error handling.

gr-fec/python/fec/LDPC/Generate_LDPC_matrix_functions.py
```diff
@@ -27,12 +27,12 @@ def read_alist_file(filename):
...
   H = zeros((numRows,numCols))
   for lineNumber in arange(4,4+numCols):
```
jdemel, line 34: Several things that would make it easier to understand.
Instead of
```python
from numpy import *
arange(4, 4 + numCols)
```
i suggest
```python
import numpy as np
np.arange(4, 4 + numCols)
```
Star imports almost always discouraged and I just stumbled over the `arange` function. I'd always expect `np.arange`. 
Also, the constant `4` needs some explaining.

gr-fec/python/fec/LDPC/Generate_LDPC_matrix_functions.py
```diff
@@ -584,7 +584,7 @@ def get_full_rank_H_matrix(H, verbose=False):
 
   return newH
 
-def get_best_matrix(H, numIterations=100, verbose=False):
+def get_best_matrix(H, numIterations=100, verbose=0):
```
jdemel, line 587: I think `verbose=False` should be the preferred type here. It just makes it more obvious that this is a simple switch.

jdemel (changes requested):
Thanks for cleaning this up. I added a few comments.
I suggest to run a Python auto formatter over this file.

gr-fec/python/fec/LDPC/Generate_LDPC_matrix_functions.py
jfmadeira: The rest of code has different levels of verbose (0, 1 and 2), hence why I changed this one as it was the only place that used True or False. Should the two verbose levels be merged then?

gr-fec/python/fec/LDPC/Generate_LDPC_matrix_functions.py
jfmadeira: I have added a commit with this restructuring

gr-fec/python/fec/LDPC/Generate_LDPC_matrix_functions.py
jfmadeira: The star import has been replaced. The constant 4 is because the lines with 1s start in the fifth line of the file.
I added a comment explaining this for now, as the 3 lines in-between are not needed for establishing the sparse matrix.

jfmadeira:
I ran an auto formatter, but was unsure if you wanted to keep the 2 space indentation so I left it in.

mormj:
@jfmadeira - Thank you for the code improvements here.  A couple of things to get this ready to be merged:
1) Please sign your commits so that it passes the DCO check - this signifies that you agree to the terms of the license
2) Please squash your commits where appropriate (the formatting commit can be separate, but commits that are related to the same change should be squashed), and edit each commit message to be prefixed with the module that is affected (in this case, `fec: short description of changes`)

All of this can be accomplished with `git rebase -i HEAD~3`
then change `pick` to `edit` for signing and `s` or `f` for squashing

gr-fec/python/fec/LDPC/Generate_LDPC_matrix_functions.py
jdemel: Thanks!

gr-fec/python/fec/LDPC/Generate_LDPC_matrix_functions.py
jdemel: I wasn't aware of the 3 verbosity levels. In this case, `int` is the appropriate type.

jfmadeira:
@mormj - Thank you for the instructions. Would it be alright if I left the commits unsquashed? I am having some trouble reordering and squashing the first and third commits, while keeping the auto formatter for last. I end up with a single commit after resolving all the rebase conflicts.

mormj:
The 3 commits here should be fine - thanks for updating!

PR #4838, opened 2021-07-05
GRC: Core features of the View-Only Mode

oscekh:
This PR contains the core parts of the GRC View-Only mode, which disables evaluation of untrusted code in flowgraphs (such as parameter expressions). This will later be accompanied by GUI elements, settings, and additional functionality such as persistent trust, which will be added in a second PR.

The `export_data`-method in [grc/core/blocks/block.py](https://github.com/gnuradio/gnuradio/compare/master...oscekh:feature/grc-view-only-mode?expand=1#diff-1a51545ee6aecb3f40600b52e6f880f922cd838bc74717b1f5c50c206bc54109) contains the changes exporting evaluated values to the file. Raw and complex parameters are exported using the `pretty_print` function which I've moved from gui to core, where as non-raw parameters are handled by yaml directly. All template expressions are exported as strings as there may be expressions of non-primitive types, such as `calcdist` in constellation blocks. I have not found a certain way of finding the type of these expressions which is why they are all stored as strings, even primitive types.

The disabled evaluation of expressions is mainly handled by modifying the parameter's and block's `evaluate`-methods. Execution of import-blocks is disabled in the flowgraph's and block's `rewrite`-methods.

There is also a [test case](https://github.com/gnuradio/gnuradio/commit/4cb244d2bb17626161c0a636c7b485d162c2e0a7#diff-dc01de441178b16c9bb34a245d82837676565888bdd93e478d332a14e4d10291) which checks for untrusted evaluation in the GRC. This is partly to check that my changes work but also to act as a safety measure for future changes that might break the View-Only mode unintentionally.

For further information on the View-Only mode itself and my GSoC work I refer to [my mail on the discuss mailing list](https://lists.gnu.org/archive/html/discuss-gnuradio/2021-05/msg00122.html).

Happy to hear your feedback!


oscekh:
The first force push is for me forgetting to sign off the commits, and the second should fix the failing grc-tests. The failing tests ran as untrusted, but I have now changed the setups so they are added to the trust list. `test_compiler` seems to fail locally for me both with my commits and without any of my changes though.

skoslowski (changes requested):
I really like how this is progressing. Good job so far. I have some minor/codestyle comments.
The test definition could be simplified IMO. To test this more thoroughly, I'd define the flowgraph directly ( I'll did up some example for you) and then we'll also need cases to check missing,incomplete,... cache values.

grc/core/params/param.py
```diff
@@ -8,6 +8,7 @@
 import ast
 import collections
 import textwrap
+import numbers
```
marcusmueller, line 11: I'm not quite sure of the GRC code style guidelines, but if in doubt, sort includes alphabetically?

grc/core/params/param.py
```diff
@@ -176,6 +190,59 @@ def is_float(self, num):
...
+         the string representation
+        """
+
+        from ...gui import Utils
+        ##################################################
+        # Simple conditions
+        ##################################################
+        value = self.get_value()
+        if not self.is_valid():
+            return self.truncate(value)
```
marcusmueller, line 208: Isn't truncate a member of the `gui` param, but not of the `core` param`?

marcusmueller (changes requested):

haakov (commented):
Great work!

I tried creating a simple flowgraph (Null source --> Null sink), saving it and opening it again. I'm getting an error from both blocks: `Can not cast evaluated value '1' to type <class 'int'>`. Seems like it's originating from the `vlen` param, but not from the `multiplicity` param. Is this caused by the `vlen` param's `hide` expression?

grc/core/FlowGraph.py
```diff
@@ -48,6 +48,8 @@ def __init__(self, parent):
 
         self.grc_file_path = ''
 
+        self.is_trusted = False
```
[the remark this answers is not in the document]
oscekh: Yup, I agree. Changed from `is_trusted` to `view_only` now

grc/core/params/param.py
oscekh: Good point, fixed now!

grc/core/params/param.py
oscekh: Good spot! I moved that over to `core/param` now :+1:

oscekh:
Thanks for all the feedback! I've updated the code accordingly and gotten more experienced with git in the process (as evidenced by the history ;) )

I noticed that the flowgraph's evaluate method was not completely taken care of. It does not evaluate (if in view only) but currently just returns None. Perhaps one could store values on a flowgraph level as well but unlike params/blocks evaluate-methods these are evald depending on the inputted namespaces. So I'm not completely sure what would be correct. What do you all think?

> I tried creating a simple flowgraph (Null source --> Null sink), saving it and opening it again. I'm getting an error from both blocks: `Can not cast evaluated value '1' to type <class 'int'>`. Seems like it's originating from the `vlen` param, but not from the `multiplicity` param. Is this caused by the `vlen` param's `hide` expression?

Right, this is due to always using the string representation as the expressions may be of non-primitive type. e.g. using the string '1' when it expects an int. As you noticed that is quite confusing, as one thinks something is "wrong".

I have changed this now so it checks the type of the expression, using the string representation if not a yaml-accepted type as with the param expressions. Thus it should not display errors for primitive types anymore, making it less confusing. For example using a null source + sink now shows up without errors. @haakov 

On this note, I would like some input on how value vs string representation is chosen ([in block's export_data](https://github.com/gnuradio/gnuradio/blob/0529fb45c185a21d9ea51e7d1a07bff25391c7ee/grc/core/blocks/block.py#L664)). Currently there are two helper-functions to keep the comprehensions readable, but it also seems somewhat silly as they are just basic conditionals. Do you think this is good, or do you have any other suggestions?

Finally, I will extend the tests to include more tests on the eval-blocking (with less ideal flowgraphs) and the caching of values.

warnes:
Feature request: include an "Add Trusted" button to the Trust Manager modal dialog which brings up a file browser that allows selection of one or more files to add to the trusted list.

haakov:
@oscekh Great, I see the Null Sink -> Null Source flowgraph works now. I think the two helper functions are fine. Maybe add a comment for each of them since the function names contain several abbrievations.

What is the plan with this PR? Merge it before the GUI PR, or merge this and the GUI PR when both are ready?

oscekh:
> Feature request: include an "Add Trusted" button to the Trust Manager modal dialog which brings up a file browser that allows selection of one or more files to add to the trusted list.

@warnes 
Good idea! I have added that feature now (locally on my repo). It will be part of the GUI PR which will be up in a few weeks.



> What is the plan with this PR? Merge it before the GUI PR, or merge this and the GUI PR when both are ready?

@haakov 
I think the plan is to merge when both are ready, core might need tweaking along with the gui work and it isn't very useful without the GUI components also I think.

warnes:
@oscekh Great!

mormj:
Can be closed as it is included in #5004

PR #4839, opened 2021-07-06
gr-newmod: Set OOT module VERSION_PATCH to 0 instead of git.

drmpeg:
This avoids creating multiple copies of the .so file
(each with a different git tag) during development.
The ldconfig command doesn't necessarily set the symlink
properly and the old .so can get loaded after an update.
This can be very confusing for developers and this patch
suppresses that behavior by always creating the .so with
the same name.


willcode (approved):
Sounds like a good idea. OOT devs might not even be using git for their code.

drmpeg:
Just FYI, if the OOT is not a git repository, the .so is named libgnuradio-modulename.so.v1.0-compat-xxx-xunknown

dkozel:
I agree this is a good change to reduce the foot shooting behavior that the current setup causes.

mormj (approved):

PR #4840, opened 2021-07-06
Upstreaming PDU blocks

jacobagilbert:
This contains four new general purpose blocks from gr-pdu_utils:

- **pdu_to_stream**: templated block that converts PDUs to streams suitable for burst transmission from UHD devices (contains `tx_sob` and `tx_eob` tags as well as optional time metadata).
- **take_skip_to_pdu**: templated block that converts a stream to PDUs based on configured take/skip values.
- **tags_to_pdu**: templated block that converts a stream to PDUs based on stream tags indicating the start and end of the PDU.
- **pdu_lambda**: a stateless python block that runs a user specified lambda function on data or metadata.

This is the second in a series of planned upstream commits to gr-pdu.


willcode:
Are you planning a backward compatibility layer for this, or will it be 3.10+?

jacobagilbert:
@willcode gr-pdu is a new module for 3.10+

this functionality is available in reasonably well maintained OOTs for 3.7/3.8/3.9

gnuradio-runtime/include/gnuradio/pdu.h
```diff
@@ -16,15 +16,26 @@
 #include <pmt/pmt.h>
 
 namespace gr {
+namespace metadata_keys {
```
marcusmueller, line 19: Like the cleanliness of this.

mormj (approved):

PR #4851, opened 2021-07-09
grc: update disabled blocks if they depend on others

### LINKED ISSUE #4788: Flowgraph doesn't update disabled blocks properly if samp_rate is controlled by another variable

dl1ksv:
If we want disabled blocks to be updated properly,
we have to consider all blocks when updating the fg not only the enabled.

This does the patch of Flowgraph.py and fixes #4788.

But a problem remains.
If you use the example in #4788 and disable the symbol_rate block,
only the samp_rate block 'turns to red'.
The other blocks don't change.
Only if you open and close arbitrary block, the other blocks are updated.

This is fixed by the patch of Application.py.

### DIFF
--- a/grc/core/FlowGraph.py
+++ b/grc/core/FlowGraph.py
@@ -216,7 +216,7 @@ class FlowGraph(Element):
         return elements
 
     def children(self):
-        return itertools.chain(self.iter_enabled_blocks(), self.connections)
+        return itertools.chain(self.blocks, self.connections)
 
     def rewrite(self):
         """
@@ -227,6 +227,7 @@ class FlowGraph(Element):
 
     def renew_namespace(self):
         namespace = {}
+        self.namespace.clear()
         # Load imports
         for expr in self.imports():
             try:


grc/core/FlowGraph.py
```diff
@@ -216,7 +222,7 @@ def get_elements(self):
         return elements
 
     def children(self):
-        return itertools.chain(self.iter_enabled_blocks(), self.connections)
+        return itertools.chain(self.iter_all_blocks(), self.connections)
```
skoslowski, line 225: Why not pass `self.blocks` directly?

grc/gui/Application.py
```diff
@@ -117,6 +117,7 @@ def _handle_action(self, action, *args):
         flow_graph = page.flow_graph if page else None
 
         def flow_graph_update(fg=flow_graph):
+            fg.update()
```
skoslowski, line 120: This doubles the effort for model updates! I think its worth investigating the actual bug and enforce the proper namespace update within a single rewrite call

grc/core/FlowGraph.py
dl1ksv: Yes, you are that works and is much easier

grc/gui/Application.py
dl1ksv: Well I looked some days for a proper solution and the only one I found was to update twice.
Due to your hint now the bug in renew_namespace.

drmpeg (approved):
Tested and works perfectly here. Very nice!

PR #4860, opened 2021-07-10
Matrix Interleaver block (backport to maint-3.8)

jaredd:
This PR is with respect to the matrix interleaver block discussed in [Issue 3637](https://github.com/gnuradio/gnuradio/issues/3637#issuecomment-841628354).

This matrix interleaver is a python hierarchical block that uses the base deinterleaver and interleaver blocks internally, so it's quasi-c++. This is also 3.8-specific, again perhaps easily ported to 3.9 due to being hierarchical.

The PR also includes an example grc and simple test case.

Backport https://github.com/gnuradio/gnuradio/pull/4861

### DIFF
--- a/gr-blocks/examples/CMakeLists.txt
+++ b/gr-blocks/examples/CMakeLists.txt
@@ -19,6 +19,7 @@
 
 install(
   FILES
+  matrix_interleaver.grc
   matrix_multiplexer.grc
   peak_detector2.grc
   py_snippets_demo.grc
--- a/gr-blocks/examples/matrix_interleaver.grc
+++ b/gr-blocks/examples/matrix_interleaver.grc
@@ -1,0 +1,341 @@
+options:
+  parameters:
+    author: Jared Dulmage
+    category: '[GRC Hier Blocks]'
+    cmake_opt: ''
+    comment: ''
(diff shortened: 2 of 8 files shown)


willcode:
@jaredd thanks for the contribution. Let's figure out how to get this on the master branch first, then backport to maintenance branches. If we do it the other way around, we risk having features that don't get ported to newer versions of software.

What makes this 3.8 specific?

jaredd:
> What makes this 3.8 specific?

Well, first I don't have a 3.9 install to test on. In fact, I can't recall the big change from 3.8 to 3.9, maybe it was swig->pybind, in which case it's true that perhaps this is compatible. I can try this on master and see what happens.

willcode:
That would be great. I have a feeling that it will work. Then we can introduce it to master and mark it for backport to maint-3.9 and maint-3.8.

willcode:
Same comment @mormj made on the other PR - could you change the commit message to be
```
blocks: add matrix interleaver block
```
or similar?

willcode:
@jaredd This is failing for Ubuntu 18.04 (Python 2).

willcode:
Needs rebase - this PR picked up some unrelated commits.

PR #4861, opened 2021-07-11
Matrix Interleaver block

jaredd:
This is the GR 3.9 equivalent of PR #4860.

The matrix interleaver is a hierarchical block that pairs a base
deinterleave block that effectively writes samples into "rows" (1 block of
samples per output) and a base interleave block that outputs samples
iteratively from each input like  "columns".

Signed-off-by: Jared Dulmage <jared.dulmage@caliola.com>

Add grc file and update blocks.tree.yml.

Fix module for matrix interleaver example

Signed-off-by: Jared Dulmage <jared.dulmage@caliola.com>

### DIFF
--- a/gr-blocks/python/blocks/matrix_interleaver.py
+++ b/gr-blocks/python/blocks/matrix_interleaver.py
@@ -1,0 +1,67 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+#
+# Copyright 2020 Caliola Engineering, LLC.
+#
+# This is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 3, or (at your option)
+# any later version.
+#
+# This software is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this software; see the file COPYING.  If not, write to
+# the Free Software Foundation, Inc., 51 Franklin Street,
+# Boston, MA 02110-1301, USA.
+#
+
+from gnuradio import gr, blocks
+
+class matrix_interleaver(gr.hier_block2):
+    def __init__(self, itemsize, rows=1, cols=1, deint=False):
+        gr.hier_block2.__init__(
+            self, "Matrix Interleaver",
+                gr.io_signature(1, 1, itemsize),
+                gr.io_signature(1, 1, itemsize),
+        )
+
+        self.itemsize = itemsize
+        self.set_rowsandcols(rows, cols, deint)
+
+    def set_rowsandcols(self, rows, cols, deint):
+        self.disconnect_all()
+
+        self.passthrough = None
+        self.interleaver = None
+        self.deinterleaver = None
+
+        ##################################################
+        # Parameters
+        ##################################################
+        self.rows = rows
+        self.cols = cols
+        self.deint = deint
+
+        ##################################################
+        # Blocks
+        ##################################################
+        # short circuit for unitary rows / columns
+        if rows == 1 or cols == 1:
+          self.passthrough = blocks.copy(self.itemsize)
+          self.connect((self, 0), (self.passthrough, 0), (self, 0))
+          return
+
+        self.deinterleaver = blocks.deinterleave(self.itemsize, 1 if deint else cols)
+        self.interleaver = blocks.interleave(self.itemsize, cols if deint else 1)
+
+        ##################################################
+        # Connections
+        ##################################################
+        self.connect((self, 0), (self.deinterleaver, 0))
+        for n in range(rows):
+            self.connect((self.deinterleaver, n), (self.interleaver, n))
+        self.connect((self.interleaver, 0), (self, 0))
--- a/gr-blocks/python/blocks/matrix_interleaver.py
+++ b/gr-blocks/python/blocks/matrix_interleaver.py
@@ -1,27 +1,23 @@
 #!/usr/bin/env python
 # -*- coding: utf-8 -*-
 #
-# Copyright 2020 Caliola Engineering, LLC.
+# Copyright 2021 Caliola Engineering, LLC.
 #
-# This is free software; you can redistribute it and/or modify
-# it under the terms of the GNU General Public License as published by
-# the Free Software Foundation; either version 3, or (at your option)
-# any later version.
+# This file is part of GNU Radio
 #
-# This software is distributed in the hope that it will be useful,
-# but WITHOUT ANY WARRANTY; without even the implied warranty of
-# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-# GNU General Public License for more details.
-#
-# You should have received a copy of the GNU General Public License
-# along with this software; see the file COPYING.  If not, write to
-# the Free Software Foundation, Inc., 51 Franklin Street,
-# Boston, MA 02110-1301, USA.
+# SPDX-License-Identifier: GPL-3.0-or-later
 #
 
 from gnuradio import gr, blocks
 
 class matrix_interleaver(gr.hier_block2):
+    """
+    Block interleaver writes inputs into conceptual rows of the matrix
+    and reads outputs by conceptual columns of the matrix.
+    In deinterleaver mode it writes inputs into conceptual columns
+    and reads outputs by conceptual rows.
+    """
+
     def __init__(self, itemsize, rows=1, cols=1, deint=False):
         gr.hier_block2.__init__(
             self, "Matrix Interleaver",
--- a/gr-blocks/python/blocks/qa_matrix_interleaver.py
+++ b/gr-blocks/python/blocks/qa_matrix_interleaver.py
@@ -1,0 +1,73 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+#
+# Copyright 2020 Caliola Engineering, LLC.
+#
+# This is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 3, or (at your option)
+# any later version.
+#
+# This software is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this software; see the file COPYING.  If not, write to
+# the Free Software Foundation, Inc., 51 Franklin Street,
+# Boston, MA 02110-1301, USA.
+#
+
+from gnuradio import gr, gr_unittest
+from gnuradio import blocks
+from matrix_interleaver import matrix_interleaver
+
+class qa_matrix_interleaver(gr_unittest.TestCase):
+
+    def setUp(self):
+        self.tb = gr.top_block()
+
+    def tearDown(self):
+        self.tb = None
+
+    def test_interleave(self):
+        tb = self.tb
+
+        # set up fg
+        cols, rows = 4, 10
+        vec = sum((cols * [x,] for x in range(rows)), [])
+        expected = cols * list(range(rows))
+
+        src = blocks.vector_source_f(vec, False)
+        itlv = matrix_interleaver(gr.sizeof_float, rows=rows, cols=cols)
+        snk = blocks.vector_sink_f()
+
+        tb.connect(src, itlv, snk)
+        tb.run()
+        result = snk.data()
+
+        # check data
+        self.assertFloatTuplesAlmostEqual(expected, result)
+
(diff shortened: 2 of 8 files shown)


mormj:
@jaredd - thank you for the contribution here.  Would you mind updating your commit message to prefix with the module that is impacted, e.g. `blocks: add matrix interleaver block`

@daniestevez - does this block fit what was described in #3637 ?

daniestevez:
@mormj thanks for tagging me in this PR.

First thing that I should be said is that I think that this PR is quite useful. The matrix interleaver here serves in more than 90% of the cases. And I like that it includes an example flowgraph and unit test.

I'm not certain if the approach followed here is the best, however. Interleaver codewords usually have hundreds or thousands of elements, so the number of columns and rows of the matrix is usually a few tens or hundreds. For this implementation, this means that the number of ports connecting the deinterleave and interleave blocks can be quite large. I think (but I'm not an expert on the internals of the runtime) that this is not so good because each connection has its own circular buffer.

I think that a more efficient implementation could be done in C++, use `set_output_multiple(rows * columns)`, and basically copy the data to the output in the appropriate order by striding through the input samples. The best thing about this PR is that this C++ implementation can be done at a later stage, and only the hierarchical Python flowgraph will need to be replaced. The rest of the PR (test, example, GRC block) could still be re-used.

Comparing with #3637, the main difference with the gr-satellites interleaver (which I should have submitted to upstream in a PR long ago, so my bad) is that the gr-satellites interleaver works on PDUs instead of a stream. I don't know which of the two options is better. Usually matrix interleaving is applied to frames of a comms system, and these are often handled in PDUs. But it's also true that a tagged stream or ordinary stream can be used to handle back-to-back frames rather than PDUs, so I don't have a clear preference. Working with PDUs has more sense for how the block is used with gr-satellites, but I'm curious to know what people think about an upstream block that should be usable in many contexts. Besides the PDU vs. stream difference the gr-satellites interleaver also has an option to throw away some part of the matrix/frame. This is only useful in rare cases, and so specific that it would make sense to do it in a separate block which is only in charge of removing some elements from the frame.

Something more to consider: in #3637 @jdemel mentioned that he had in gr-symbolmapping an interleaver which could use a general permutation, given to the block as a list / numpy array. A matrix interleaver could easily be constructed as [he indicated](https://github.com/gnuradio/gnuradio/issues/3637#issuecomment-662491711). The only downside I see to this implementation is that it's slightly less efficient than a less generic one if used for matrix interleaving. It is also a bit cumbersome to build the matrix interleaver with the numpy expression that Johannes gave, but this could be hidden in a GRC YAML block.

My suggestion is, regardless of the other similar blocks that Johannes and I have lying around, to review and eventually merge this PR, since it seems quite useful as it is.

gr-blocks/python/blocks/matrix_interleaver.py
```diff
@@ -0,0 +1,67 @@
...
+# You should have received a copy of the GNU General Public License
+# along with this software; see the file COPYING.  If not, write to
+# the Free Software Foundation, Inc., 51 Franklin Street,
+# Boston, MA 02110-1301, USA.
+#
+
+from gnuradio import gr, blocks
+
+class matrix_interleaver(gr.hier_block2):
+    def __init__(self, itemsize, rows=1, cols=1, deint=False):
```
daniestevez, line 25: A couple of suggestions that you can consider here:

* I always get confused about the meaning of rows and columns. Should data be written by rows and read by columns or the other way around? I think that there are different conventions in the literature, so I suggest to document clearly the approach we follow here (specially because it's not so obvious to me by looking at the code below). I've seen this is explained in the GRC block documentation, but probably it's worth to add the same as a comment in the Python code.

* I think that deinterleaving can be seen as the same process as interleaving but exchanging the number of rows and number of columns (i.e., matrix transpose). So we could do without the `deint` parameter. I think this is more a matter of taste, though, and I can see people arguing that they prefer to have a deinterleave option than to have to think about this property.

gr-blocks/python/blocks/matrix_interleaver.py
```diff
@@ -0,0 +1,67 @@
...
+        )
+
+        self.itemsize = itemsize
+        self.set_rowsandcols(rows, cols, deint)
+
+    def set_rowsandcols(self, rows, cols, deint):
+        self.disconnect_all()
+
+        self.passthrough = None
+        self.interleaver = None
```
daniestevez, line 39: Are this and the line below necessary? `self.interleaver` and `self.deinterleaver` are always re-assigned below.

gr-blocks/python/blocks/matrix_interleaver.py
```diff
@@ -0,0 +1,67 @@
...
+            self, "Matrix Interleaver",
+                gr.io_signature(1, 1, itemsize),
+                gr.io_signature(1, 1, itemsize),
+        )
+
+        self.itemsize = itemsize
+        self.set_rowsandcols(rows, cols, deint)
+
+    def set_rowsandcols(self, rows, cols, deint):
+        self.disconnect_all()
```
daniestevez, line 36: I'm asking because I honestly don't know the answer: Is it safe to call `self.disconnect_all()` while the flowgraph is running? I usually pause the flowgraph before changing block connections.

gr-blocks/python/blocks/matrix_interleaver.py
jaredd: > First thing that I should be said is that I think that this PR is quite useful. The matrix interleaver here serves in more than 90% of the cases. And I like that it includes an example flowgraph and unit test.

I really appreciate this review @daniestevez and the thoughtfulness of the recommendations. 

> 
> I'm not certain if the approach followed here is the best, however. Interleaver codewords usually have hundreds or thousands of elements, so the number of columns and rows of the matrix is usually a few tens or hundreds. For this implementation, this means that the number of ports connecting the deinterleave and interleave blocks can be quite large. 

This block was developed in the context of long streams of contiguous PHY frames you designate "back-to-back". The typical interleaver in my flowgraph has 8 rows & block sizes up to 312 kbits. Most of my testing has been with block sizes under 20 kbits which still means GR is handling around 3 k outputs (thus buffers) for the deinterleave sub-block in matrix deinterleaver mode. I haven't looked at the performance impact of block size scale at all so that may be an issue.

I also haven't looked at the (de)interleave to know if those implementations are particularly optimized, but this matrix (de)interleaver block would automatically benefit from any improvements to those.

> I think that a more efficient implementation could be done in C++, use `set_output_multiple(rows * columns)`, and basically copy the data to the output in the appropriate order by striding through the input samples. The best thing about this PR is that this C++ implementation can be done at a later stage, and only the hierarchical Python flowgraph will need to be replaced. The rest of the PR (test, example, GRC block) could still be re-used.

Agree. I think this block could easily work alongside an arbitrary interleaver in the interim until tests could definitively declare a uniformly superior solution.

gr-blocks/python/blocks/matrix_interleaver.py
jaredd: As this block allows changes after default, there's no guarantee that the resulting interleaver is passthrough or not. So I start all the blocks as None to avoid lingering blocks. I probably could assign them within the conditionals but this seemed more straightforward. There also may have been an issue with assigning a new block prior to clearing the old one. I feel like maybe a naming conflict? I'd have to try it again to check.

gr-blocks/python/blocks/matrix_interleaver.py
jaredd: This block was used in the context of a GRC that had some default settings but could be updated (via variables) prior to a run. The interleave and deinterleave blocks did not allow that at all (no exposed setters if I recall), so this was my solution. The way this block works if the number of connections between the two blocks changes, you'd have to disconnect anyways. 

The purpose was not to allow updates at runtime, but prior to runtime yet after the original flowgraph construction. This may have been specifically to work around GRC behavior more so than GR. I don't think callbacks in GRC (e.g. changing the interleaver block size) prompt a rebuild of the flowgraph, they just call into blocks on the existing flowgraph.

I have never called this callback during runtime, seems like it would be pretty cavalier in this case. Perhaps a disclaimer in the documentation for this use case.

gr-blocks/python/blocks/matrix_interleaver.py
jaredd: I don't know why my comment above showed up under this code comment when it was intended to respond to [the comment above](https://github.com/gnuradio/gnuradio/pull/4861#issuecomment-880981384), too much in the interface for me to parse apparently. 

To respond to this code comment, definitely I can add documentation to the python code. I like the deinterleave flag so that it reduces the cognitive load on the developer. Much easier to simply copy the interleaver block and select the flag, than swap arguments I believe. Maybe some GRC-fu to change the appearance of the block when the deint flag is asserted?

gr-blocks/python/blocks/matrix_interleaver.py
jdemel: Currently, the API should throw an exception or ignore this call during flowgraph execution. If you want to change anything internally before the first call to `start`, I suggest to instantiate a new block.

jdemel:
So this is the current interleaver that I have. This is just the core. I separated the interleaver implementation and the GR block.

https://github.com/ant-uni-bremen/gr-symbolmapping/blob/cc8f4ed8b013f2664364016129fb373d5ae3cd74/include/symbolmapping/bit_interleaver.h#L39-L45

I benchmarked it. Obviously the exact performance depends on the data type and your CPU. Roughly for less than 10k elements, it takes less than 4us on a TRX3970X. In case of packed bits, you'd want to unpack -> interleave -> pack. This operation incurs quite a penalty. My benchmarks indicate to cross the 4us bound at ~2500 elements. Performance for "random" interleaving (e.g. `np.random.permutation(1024)`) is mostly an upperbound. If you plug in anything like a matrix interleaver sequence, it'd take a lower latency penalty. (and thus throughput.)

The only part that might be an issue is that the interleaver always requires a full block. This might not be the case for all interleavers under all circumstances. Thus
```cpp
set_output_multiple(interleaver_length);
```
is required. This setting implies that you can't change the interleaver length after the flowgraph started.

I suggest to go with @daniestevez suggestion to merge this interleaver and exchange the internals if they become available. This implies that we agree on an interface. I wonder if `vlen` is necessary for the GRC block.

gr-blocks/python/blocks/matrix_interleaver.py
daniestevez: Regarding GRC-fu, not that I'm aware of, besides what you're already getting by using `hide: none` in the deint option in the YAML file. Another option could be to have two different blocks GRC for interleave and deinterleave, like we have scramble and descrambler, and make both use the same Python block. I slightly prefer a single block with the boolean flag, as you've made.

gr-blocks/python/blocks/matrix_interleaver.py
daniestevez: Since this callback shouldn't be called at runtime, is it a good idea to declare it in the GRC YAML file? I haven't tested it, but I suspect that if someone puts a QT GUI variable in GRC for the number of rows, then changes the QT GUI value during runtime, the callback will get called. I think it's better not to let GRC think that the parameters can be changed via callback.

daniestevez:
> I wonder if vlen is necessary for the GRC block.

I think it's nice to have a `vlen` for the use case in which you have elements
```
a_0 b_0 a_1 b_1 a_2 b_2 a_3 b_3
```
and want to interleave them as
```
a_0 b_0 a_2 b_2 a_1 b_1 a_3 b_3
```
Where could this problem show up? Perhaps interleaving soft symbols from a MIMO system (where `a` and `b` would represent to antennas).

gr-blocks/python/blocks/qa_matrix_interleaver.py
```diff
@@ -0,0 +1,73 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+#
+# Copyright 2020 Caliola Engineering, LLC.
+#
+# This is free software; you can redistribute it and/or modify
```
mormj, line 6: This (and other places) can be replaced by the more terse spdx header

gr-blocks/python/blocks/qa_matrix_interleaver.py
```diff
@@ -0,0 +1,73 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+#
+# Copyright 2020 Caliola Engineering, LLC.
```
mormj, line 4: 2021?

gr-blocks/python/blocks/matrix_interleaver.py
jaredd: I have a multimode transmitter with a set of interleaver options. The interleaver is buried in a hierarchy where I can set a mode with only a single top-level parameter and it propagates down to this block.

Without this capability I would have to have multiple versions of my flow graph with different defaults. This is a huge timesaver for interactive exploration of parameters between runs. I wonder if there's a way to get GRC to disable changes after starting the fg?  It would be a shame to have to workaround the reconfiguration limitations again.

gr-blocks/python/blocks/qa_matrix_interleaver.py
jaredd: Is it? Already? SMH.

mormj:
Going ahead and merging this - as mentioned in the reviews, the internals can be swapped later, but it is a useful block

PR #4863, opened 2021-07-11
gr-blocks: (un)pack-k-bits tag propagation Bugfix #1111

### LINKED ISSUE #1111: Pack k bits tag propagation error
Pack k bits uses the default mechanism for propagating tags which uses the relative rate to determine the output index based on the input index.  It seems, however, that the output index is determined as the round(input_index * relative rate) which means if I have an input with a tag on 5 through a pack 2 bits block, my tag comes out on index round(5 * 1/2) = 3 rather than index 2 (where it's the 1st bit of the output item).  Really, the output tags should be associated with the word containing the bit the tag was on at the input.

jasonuher:
fixes #1111

Signed-off-by: Jason Uher <jason.uher@jhuapl.edu>

### DIFF
--- a/gr-blocks/lib/pack_k_bits_bb_impl.cc
+++ b/gr-blocks/lib/pack_k_bits_bb_impl.cc
@@ -31,6 +31,8 @@ pack_k_bits_bb_impl::pack_k_bits_bb_impl(unsigned k)
                      k),
       d_pack(k)
 {
+    d_k = k;
+    set_tag_propagation_policy(TPP_CUSTOM);
 }
 
 int pack_k_bits_bb_impl::work(int noutput_items,
@@ -40,7 +42,21 @@ int pack_k_bits_bb_impl::work(int noutput_items,
     const unsigned char* in = (const unsigned char*)input_items[0];
     unsigned char* out = (unsigned char*)output_items[0];
 
+    std::vector<tag_t> wintags; // Temp variable to store tags for prop
+
+    // GR_LOG_DEBUG(d_logger, boost::format("Packing Outputs"));
     d_pack.pack(out, in, noutput_items);
+
+    // Propagate tags
+    get_tags_in_range(wintags, 0, nitems_read(0), nitems_read(0) + (noutput_items * d_k));
+
+    // GR_LOG_DEBUG(d_logger, boost::format("Propagating tags"));
+    std::vector<tag_t>::iterator t;
+    for (t = wintags.begin(); t != wintags.end(); t++) {
+        tag_t new_tag = *t;
+        new_tag.offset = std::floor((double)new_tag.offset / d_k);
+        add_item_tag(0, new_tag);
+    }
 
     return noutput_items;
 }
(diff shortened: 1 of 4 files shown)


mormj:
@jasonuher - thank you for the bug fix for the pack_k_bits block.  Can you please run clang-format to get the Check C++ Formatting test to pass?  Thanks!

jasonuher:
@mormj no problem.  I thought I did run it, but it seems I failed.  I'll figure it out and re-squash.

jasonuher:
qa_filterbank failed but only on ubuntu 20.04?  I didn't touch anything in there and all tests pass fine on my machine (ubuntu 20.04, pybombs, gr-master).

mormj:
`qa_filterbank` is one of those finicky qa tests that shows up failing every now and again.  Since it's not related to this PR, we can ignore it.  

Is this still WIP?

jasonuher:
> `qa_filterbank` is one of those finicky qa tests that shows up failing every now and again. Since it's not related to this PR, we can ignore it.
> 
> Is this still WIP?

Nope, should be good then!

gr-blocks/lib/pack_k_bits_bb_impl.cc
```diff
@@ -40,8 +42,22 @@ int pack_k_bits_bb_impl::work(int noutput_items,
...
+    // GR_LOG_DEBUG(d_logger, boost::format("Packing Outputs"));
     d_pack.pack(out, in, noutput_items);
 
+    // Propagate tags
+    get_tags_in_range(wintags, 0, nitems_read(0), nitems_read(0) + (noutput_items * d_k));
+
+    // GR_LOG_DEBUG(d_logger, boost::format("Propagating tags"));
+    std::vector<tag_t>::iterator t;
+    for (t = wintags.begin(); t != wintags.end(); t++) {
+        tag_t new_tag = *t;
```
marcusmueller, line 56: No need to copy the tags – `get_tags_in_range` fills your vector with tags you own!

gr-blocks/python/blocks/qa_pack_k_bits.py
```diff
@@ -54,6 +53,45 @@ def test_003(self):
         self.tb.run()
         self.assertEqual(list(expected_results), list(snk.data()))
 
+    def test_004(self):
```
marcusmueller, line 56: :+1:

marcusmueller (approved):

gr-blocks/lib/pack_k_bits_bb_impl.cc
jasonuher: Are you saying a better solution would be to just edit the offset of `t` directly and call `add_item_tag(0,t)`?

I largely copied this from the block execution code (ie, where the 'automatic' tag propagation happens).  My thought was the 'copy' to `new_tag` was to allow editing the offset.  I assumed it had something to do with directly editing the target object of the iterator but I'm also not really sure...

marcusmueller: jep, I'd say that would be the more performant method. HOWEVER, I did accept this because this is a common pattern in the codebase and ecosystem, and it's really not that expensive, and honestly, this patch is good and should be merged. We can improve the perf separately.

PR #4864, opened 2021-07-12
global: remove unused boost includes (backport to maint-3.9)

willcode:
Most changes to public include files were omitted for the backport.

However, `<boost/utility.hpp>` was removed from 'gnuradio-runtime/include/gnuradio/thread/thread_group.h` as this file is very unlikely to be used by OOTs, and is not propagated by any other public includes in-tree.

Removed emoji from one commit message :-)

F33 required a couple additional includes.

Backport https://github.com/gnuradio/gnuradio/pull/4778

### DIFF
--- a/gnuradio-runtime/include/gnuradio/high_res_timer.h
+++ b/gnuradio-runtime/include/gnuradio/high_res_timer.h
@@ -12,6 +12,7 @@
 #define INCLUDED_GNURADIO_HIGH_RES_TIMER_H
 
 #include <gnuradio/api.h>
+#include <boost/date_time/posix_time/posix_time.hpp>
 
 ////////////////////////////////////////////////////////////////////////
 // Use architecture defines to determine the implementation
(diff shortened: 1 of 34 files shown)


willcode:
@marcusmueller I think this maintains the API while including most of the boost include cleanups. Does removing the cmake GrBoost have some kind of ripple in deps, though? We might have to leave that dangling dep in there if so.

mormj (approved):

marcusmueller:
> @marcusmueller I think this maintains the API while including most of the boost include cleanups. Does removing the cmake GrBoost have some kind of ripple in deps, though? We might have to leave that dangling dep in there if so.

Oh, great catch! I'm actually not sure, but seeing that I especially don't know how linking works on non-gcc/clang builders, let's err on the side of caution.

willcode:
@ryanvolz Any thoughts on whether removing the transitive library dep on boost from the cmake files will cause linking problems anywhere? I'm not too worried about the stray person who linked against gr-channels and then used boost in their own code without linking, but anything more likely?

ryanvolz:
I don't anticipate any problems with the CMake changes. The case you described is the only potential issue I see, but I think it is exceedingly unlikely.

PR #4869, opened 2021-07-12
Update CODE_OF_CONDUCT.md

777arc:
official version lives in gr-governance.  this should probably get backported at least to 3.8 and 3.9


777arc:
Oops didn't realize the gr-governance version never got updated, ignore this, i'll redo

Issue #4868, opened 2021-07-12
Constant Source value 'short' not in allowed values v3.8.3.1


lmr12:
I recently installed v3.8.3.1 via pybombs, and now when I try to run flowgraphs which use the Constant Source block with the Output Type "short" the properties tab show that "value 'short' no in allowed values". This was not an issue in v3.8.2.0 and I am wondering if this was an intentional change. I did not see any mention in the change-log.
![Screenshot 2021-07-09 101243](https://user-images.githubusercontent.com/87322830/125313722-c29d8c00-e2f2-11eb-9bae-ef8ed3f7ebb2.png)
![Screenshot 2021-07-09 101150](https://user-images.githubusercontent.com/87322830/125313726-c3362280-e2f2-11eb-9de5-ac7b4d355f2f.png)
![Screenshot 2021-07-09 100954](https://user-images.githubusercontent.com/87322830/125313730-c3362280-e2f2-11eb-90ab-07f3430e497f.png)

willcode:
@ChristopheSeguinot @dl1ksv something related to type checking and/or ordering, I think.

Also affects master and maint-3.9. `byte` does not work either. Other blocks allow `short` and `byte`.

In `gr-analog/grc/analog_const_source_x.block.yml`:
```
parameters:
-   id: type
    label: Output Type
    dtype: enum
    options: [complex, float, int, short, byte]
    option_attributes:
        const_type: [complex, real, int, short, gr.sizeof_char]
        fcn: [c, f, i, s, b]
    hide: part
-   id: const
    label: Constant
    dtype: ${ type.const_type }
    default: '0'

outputs:
-   domain: stream
    dtype: ${ type }
```
Adding a print to `EvaluatedEnum(Evaluated)` show
```
Value 'gr.sizeof_char' not in allowed values {'_multiline_python_external', 'complex_vector', 'enum', 'id', 'real', 'int_vector', 'gui_hint', 'int', 'float_vector', 'file_open', 'stream_id', 'string', 'import', 'hex', 'dir_select', 'complex', 'real_vector', 'file_save', 'raw', 'float', '_multiline', 'name', 'bool'}
```
which is clearly not type checking the right enum.

willcode:
This gets it part way fixed:
```
$ git diff
diff --git a/gr-analog/grc/analog_const_source_x.block.yml b/gr-analog/grc/analog_const_source_x.block.yml
index 929ed6496..3fdb883d7 100644
[...]

PR #4876, opened 2021-07-13
grc: allow short and byte as valid types in an enum

willcode:
Add short and byte to valid param type names and param value handler. Also replace stray bad type in const_source yml.

Fixes https://github.com/gnuradio/gnuradio/issues/4868


mormj (approved):

ThomasHabets:
This doesn't seem to have fixed `const_source`. I can't use it in rc4 with bytes.

Issue #4761, opened 2021-06-12
OOT module import: Inadequate error handling 


errikosmes:
I recently had a problem where a mistake I did in one c++ block that I developed caused every other c++ block in my module not to be imported without any warning (I just got a python error from my main script saying that the submodule I tried to import didn't exist). 

The mistake that I did wasn't strictly an error in c++ terms (I declared a forecast function without defining it), so it didn't get caught by the compiler. However it made the swig generated python submodule to fail completely. I managed to find the issue with the help of the community over at Element (thank you very much guys!), but the error was really hard to find. Installing a module that makes every other disappear was a pretty weird issue, at least for me.

Having had a think about why I never saw any import error messages that could have helped find the issue I stumbled upon this try-except bit on the `__init__.py` file generated by gr_modtool for my module:

```
# import swig generated symbols into the mymodule namespace
try:
    # this might fail if the module is python-only
    from .mymodule_swig import *
except ImportError:
    pass  
```

In my opinion that's kind of a dangerous use of a try-expept statement because in my case it hid away the error that led my to the solution (I had to import my module manually in a python shell to see it). I think it would be better to at least have it like that: 
```
# import swig generated symbols into the mymodule namespace
try:
    # this might fail if the module is python-only
    from .mymodule_swig import *
except ImportError as error:
    print(error)  
```
I get that this will print an error even when there's not a problem if the user has only defined python blocks, so we do need to find a better solution from what I'm proposing. But in the end the first version is just not right for me.

velichkov:
Hi @errikosmes,

Welcome to gnuradio and many thanks for the detailed report! In the `master` and `maint-3.9` branches this problem is solved by changing `ImportError` to `ModuleNotFoundError`. 

@willcode, I think that #3926 needs to be cherry-picked in `maint-3.8`.

willcode:
ModuleNotFoundError is not in Python2, so we would need a different change. Suggestions?

errikosmes:
Hi @velichkov,

Good to hear that the issue is getting handled in the newer versions. I'm still on 3.8.3.0 because I'm using a LimeSDR and their blocks are still 3.7 or 3.8 only.

I tested it and I can confirm that `ModuleNotFoundError` would had raised the error that led me to the solution.

For Python2 @willcode I have no idea unfortunately.

velichkov:
> ModuleNotFoundError is not in Python2, so we would need a different change.

OK, I see.

> Suggestions?

[...]

PR #4881, opened 2021-07-13
modtool: Change handled exception to ModuleNotFoundError (backport to maint-3.8)

willcode:
Backport 08cd88d but maintain compatibility with Python2.

Backport https://github.com/gnuradio/gnuradio/pull/3926
Fixes https://github.com/gnuradio/gnuradio/issues/4761 for v3.8

### DIFF
--- a/gr-utils/python/modtool/templates/gr-newmod/python/__init__.py
+++ b/gr-utils/python/modtool/templates/gr-newmod/python/__init__.py
@@ -24,11 +24,18 @@ description here (python/__init__.py).
 '''
 from __future__ import unicode_literals
 
+# GNU Radio 3.8 is compatible with both Python 2 and 3,
+# which raise different exceptions if a module is not found.
+try:
+    module_not_found_error = ModuleNotFoundError
+except NameError:
+    module_not_found_error = ImportError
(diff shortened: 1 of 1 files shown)


willcode:
It seems like there should be some more elegant way to do this, but I couldn't think of one.

velichkov (approved):
To avoid code duplication you may try something like

```python
try:
    ModuleNotFoundError
except NameError:
    ModuleNotFoundError = ImportError
```
This should define `ModuleNotFoundError` as `ImportError` when not available (for python < 3.6) and later always catch only `ModuleNotFoundError`

I don't have gnuradio 3.8 with python2 and can't easily test this.

Also I'm OK with your change as is.

willcode:
It turns out that ImportError will catch the exception on both Python2 and Python3 since ImportError is a base class of ModuleNotFoundError. Even though that's less specific, I might just go back to ImportError for 3.8 to avoid any clever code.

willcode:
@velichkov I took that basic idea, but was a little worried about aliasing the actual exception name.

velichkov (approved):
LGTM

Issue #4241, opened 2021-02-15
Usage of std::to_string introduces locale-dependency where it shouldn't


marcusmueller:
I just realized: while merging the commit that got rid of the last couple dozen std::cout / printf logging calls, I had to fix a lot of stuff "on the fly".

While airborne, I used std::to_string on numeric values in quite a few places.

That was a terrible idea: to_string is localized, and you don't want your comma-separated sample values logged anywhere to suddenly contain commas as decimal separators (and a variety of other ugly things that can go wrong).

This is a bug :(

mbr0wn:
If we want to make it translatable, isn't a locale'd version of to_string() the right thing?

mbr0wn:
FYI, I've used to_string() too.

marcusmueller:
> If we want to make it translatable, isn't a locale'd version of to_string() the right thing?

Uff, good question. I was assuming GRC, and our CLI tools should be localized, but logging? Would you want to have a mailserver that logs in German, so that you can't search the internet for error messages?

marcusmueller:
This is a significant problem in gr-soapy: here ranges and other floats are converted using the localied to_string, which can lead to broken address strings

But I think it's solved in other places (i.e., the rest is pretty safe integer conversions).

marcusmueller:
tagging as medium, because this also affects parameters passed to libiio which passes them to the kernel which expects specific decimal separators

PR #4886, opened 2021-07-14
gr-iio: pluto source /sink fails to set paramter if german locale is used

dl1ksv:
pluto source / sink fails to set hardwaregain with

gr::log :WARN: iio::device::set_params - Unable to write attribute in_voltage0_hardwaregain: -22
gr::log :WARN: iio::device::set_params - Unable to write attribute out_voltage0_hardwaregain: -22

gain values are treated as double. The conversion std::to_string() uses the locale setting.
So in the US std::to_string(10.) leads to "10.0" while in Germany the result is "10,0" which the pluto handles as error.
Forcing LC_NUMERIC to "C" fixes this issue.

### DIFF
--- a/gr-iio/lib/device_source_impl.cc
+++ b/gr-iio/lib/device_source_impl.cc
@@ -92,7 +92,8 @@ void device_source_impl::set_params(struct iio_device* phy,
             ret = iio_device_debug_attr_write(phy, attr, val.c_str());
         if (ret < 0) {
             GR_LOG_WARN(logger,
-                        boost::format("Unable to write attribute %s: %d") % key % ret);
+                        boost::format("Unable to write attribute %s: %d %s") % key % ret %
+                            val);
         }
     }
 }
--- a/gr-iio/lib/fmcomms2_sink_impl.cc
+++ b/gr-iio/lib/fmcomms2_sink_impl.cc
@@ -185,9 +185,11 @@ void fmcomms2_sink_impl::set_attenuation(size_t chan, double attenuation)
         throw std::runtime_error("Channel out of range for this device");
     }
     std::vector<std::string> params;
-
-    params.push_back("out_voltage" + std::to_string(chan) +
-                     "_hardwaregain=" + std::to_string(-attenuation));
+    std::string att_value = std::to_string(-attenuation);
+    std::string::size_type idx = att_value.find(',');
+    if (idx != std::string::npos) // found , as decimal separator, so change to .
+        att_value.replace(idx, 1, ".");
+    params.push_back("out_voltage" + std::to_string(chan) + "_hardwaregain=" + att_value);
 
 
     device_source_impl::set_params(this->phy, params);
--- a/gr-iio/lib/fmcomms2_source_impl.cc
+++ b/gr-iio/lib/fmcomms2_source_impl.cc
@@ -220,10 +220,12 @@ void fmcomms2_source_impl::set_gain(size_t chan, double gain_value)
     std::vector<std::string> params;
 
     if (d_gain_mode[chan].compare("manual") == 0) {
-        // params.push_back("in_voltage" + std::to_string(chan) +
-        //                  "_gain_control_mode=" + d_gain_mode[chan]);
+        std::string gain_string = std::to_string(gain_value);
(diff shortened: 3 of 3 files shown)


tfcollins:
Thanks for the fix. Can this be moved to the device sink+source blocks so all IIO based blocks would get the update?

dl1ksv:
Setting LC_NUMERIC in device_source/sink_impl instead of pluto_source/sink works, too.
So I changed this pr accordingly.

willcode:
This would be the only use of locale in GR. Is there some more general way we can fix this?

marcusmueller:
> This would be the only use of locale in GR. Is there some more general way we can fix this?

see also #4241: I did the same mistake; `std::to_string` is not the right tool if you want to do anything but generate localized string representations.

Since we're not quite where we have `std::format`, I'd propose we use [`to_chars`](https://en.cppreference.com/w/cpp/utility/to_chars) instead. But that's C++17 only! So, that'd be an avoidable backport obstacle.

Problem with `setlocale` is that it sets the locale – for the whole process. That affects all stdlib/libc IO functions. Up to that point, logging messages would be formatted in German, after in English. And if someone uses GNU Radio in a program that rightfully uses system locale to display strings...

I think setting the locale might be a bit extreme!

dl1ksv:
My patch does not change the locale settings itself, it only forces the the decimal separator to be **.** !
This affects only users like me who use a locale that uses , as separator. So why not set LC_NUMERIC  explicitly set, if you want to be the separator . !
It's true, after changing the separator, all calls to std::to_string() use **.**  as separator.
If this is a problem, we could define a modified to_string like

get LC_NUMERIC
set LC_NUMERIC to .
to_string
restore LC_NUMERIC

marcusmueller:
> My patch does not change the locale settings itself, it only forces the the decimal separator to be . !

Yep, understood that :)

```
get LC_NUMERIC
set LC_NUMERIC to .
to_string
restore LC_NUMERIC
```

Sadly, setting a locale is a relatively intense (read: takes some time) process, and it's not threadsafe – which means this can go wrong and then be very hard to debug: 

> Because setlocale modifies global state which affects execution of locale-dependent functions, it is undefined behavior to call it from one thread, while another thread is executing any of the following functions: std::fprintf, std::isprint, std::iswdigit, std::localeconv, std::tolower, std::fscanf, std::ispunct, std::iswgraph, std::mblen, std::toupper, std::isalnum, std::isspace, std::iswlower, std::mbstowcs, std::towlower, std::isalpha, std::isupper, std::iswprint, std::mbtowc, std::towupper, std::isblank, std::iswalnum, std::iswpunct, std::setlocale, std::wcscoll, std::iscntrl, std::iswalpha, std::iswspace, std::strcoll, std::wcstod, std::isdigit, std::iswblank, std::iswupper, std::strerror, std::wcstombs, std::isgraph, std::iswcntrl, std::iswxdigit, std::strtod, std::wcsxfrm, std::islower, std::iswctype, std::isxdigit. 

<sup>[source](https://en.cppreference.com/w/cpp/locale/setlocale#Notes)</sup>

dl1ksv:
Not threadsafe could be a problem.
So , what's the conclusion to fix this issue?
The easiest way for me is to 
export LC_NUMERIC="C"
But in general which locale does require gnuradio , to run properly? Should we set this locale somewhere ?

marcusmueller:
Ah I think the answer is that GNU Radio must run with *any* locale, and the parts that
display and print localized strings *should* respect that locale, whereas the parts that
interface with other software (like iio) should not.

The conclusion hence is that gr-iio mustn't use `std::to_string` to convert things; it
would have to rely on `boost::format`, or `std::stringstream` (which can have it's own
independent locale) or `std::to_chars` for conversion. All of them come with their own
downsides :(

dl1ksv:
Following the discussion above, I removed the setlocale call and replaced std::to_string by boost::lexical_cast for double values.

dl1ksv:
This works for the rx gain as it's step size is one but only partially for the tx attenuation is it's step size is 0.25

dl1ksv:
Removed boost::lexical_cast . Now check if the converted string of a real number contains a , and change   it to a dot.

mormj (approved):

PR #4888, opened 2021-07-14
grc: Allow func prob args to be changed dynamically

### LINKED ISSUE #4942: grc: Inability to change func probe's func arg values during runtime
Function Probe allows us to call a function at a repeated interval from top block with an option of supplying an argument.
Sometimes, we would want to change the value of said argument or arguments through QT GUI (e.g. through a range object). Take the flow graph below for example. I have two range objects for two arguments to a function called by the function probe block. 

Screenshot from 2021-07-21 07-29-56

Screenshot from 2021-07-21 07-30-01

In the epy, I print out the values Im given. I would expect the arguments `test` and `test2` to be the same value I set in their corresponding range objects.

```python
    def testing123(self, test, test2):
        p_dict = pmt.make_dict()
        p_dict = pmt.dict_add(p_dict, pmt.intern("test"), pmt.from_long(test))
        p_dict = pmt.dict_add(p_dict, pmt.intern("test2"), pmt.from_long(test2))
        self.message_port_pub(pmt.intern("output"), p_dict)
```

However, based on current configuration, it does not change. It remains at its default value at 50. 

Screenshot from 2021-07-21 07-29-12

solomonbstoner:
When the func prob block is used to run a function defined in a epy
block to accept any number of arguments, it is impossible to dynamically
change the value of the argument during runtime despite a callback
function setting its value. This is because it is missing a 'self.' prefix.

An example GRC is [in this link](https://pastebin.com/raw/fu9KFtQi). Without the patch, one cannot change the value of the arguments, `Test` and `Test2` during runtime. This is made possible with this patch.

Fixes #4942 

### DIFF
--- a/grc/blocks/variable_function_probe.block.yml
+++ b/grc/blocks/variable_function_probe.block.yml
@@ -35,7 +35,13 @@ templates:
         def _${id}_probe():
           while True:
             <% obj = 'self' + ('.' + block_id if block_id else '') %>
-            val = ${obj}.${function_name}(${function_args})
+            <%
+            if function_args:
+                function_args_impr = ','.join(list('self.' + function_arg.strip() for function_arg in function_args.split(',')))
+            else:
+                function_args_impr = ''
+            %>
+            val = ${obj}.${function_name}(${function_args_impr})
             try:
               try:
                 self.doc.add_next_tick_callback(functools.partial(self.set_${id},val))


solomonbstoner:
Fixes #4942

mormj (approved):

grc/blocks/variable_function_probe.block.yml
```diff
@@ -35,7 +35,13 @@ templates:
         def _${id}_probe():
           while True:
             <% obj = 'self' + ('.' + block_id if block_id else '') %>
-            val = ${obj}.${function_name}(${function_args})
+            <%
+            if function_args:
+                function_args_impr = ','.join(list('self.' + function_arg.strip() for function_arg in function_args.split(',')))
```
mormj, line 40: Is it the case that any function args used to a method that would be used in the function probe would be `self. ...`?  What if the argument is a number or enum?

solomonbstoner: Regardless of the type, wouldn't the argument still come in the form of a variable block? A variable block still includes the `self.` e.g.
```
        # Variables
        self.test = test = 5
        self.samp_rate = samp_rate = 192000
```
Are there cases where arguments come without a variable block?

mormj: It wouldn't have to - what if my probe block was going to call the block method `blk.get_status(3)` - where 3 the value of some enum that returns what type of status to get.  Admittedly not the best design, but something someone could previously do using the function probe block

mormj: Yes, that is a much clearer way to phrase it :)

solomonbstoner: Yeah I think my PR would break arguments that are literal constants. Lemme work on it and ill mark it for review when Im done. Thanks for raising it josh

solomonbstoner:
Unable to think of a way to fix #4942 without breaking literal constants. Doesnt seem like a pressing issue either. Closing.

PR #4889, opened 2021-07-14
runtime: dereference symlink in gr::prefix() (backport to maint-3.9) (V2)

ryanvolz:
Closes #4807 (superseded).


ryanvolz:
@willcode The CI tests should be happy with this, but it could use some local verification when you have the time.

willcode:
Is the `"/"` the only difference? I'll build here if it passes. Thought I might have tried this, but I tried enough things that I can't remember now.

ryanvolz:
> Is the `"/"` the only difference? I'll build here if it passes. Thought I might have tried this, but I tried enough things that I can't remember now.

Effectively, since I expect `gr_runtime_lib_path` to always be absolute. I already ran this once with the CI on my fork, and it passed, so I guess this is indeed on that you didn't try.

willcode:
Cool, that would be nice and simple.

willcode:
Works here.

Issue #4561, opened 2021-05-02
Polyphase Arbitrary Resampler high CPU usage when throttled


daniestevez:
I have noticed that in certain circumstances the Polyphase Arbitrary Resampler block will have a high CPU usage (basically 100% of a CPU core) even when the data flow is limited by a throttle block.

The following flowgraph illustrates what I mean. There are throttle blocks that limit the rate to 2ksps, so the CPU usage should be very low.  If both throttles or only the throttle before the resampler are enabled, then the CPU usage is low as expected. However, if only the throttle after the resampler is enabled, the thread for the Polyphase Arbitrary Resampler will use up 100% of a CPU core. Moreover, the CPU usage of the Noise Source will also be high.

![test_arb_resampler](https://user-images.githubusercontent.com/15093841/116814388-09d00980-ab59-11eb-813a-71f39ae3188e.png)
[test_arb_resampler.grc.zip](https://github.com/gnuradio/gnuradio/files/6411424/test_arb_resampler.grc.zip)

I haven't looked into the cause of this problem yet, but my impression is that the resampler `work()` function is called needlessly too often.

I'm using GNU Radio v3.9.1.0 for this test.

willcode:
Sounds like it's handling one sample at a time when rates are low, especially when interpolating.

daniestevez:
Even though very inefficient, that would be fine from the point of view of latency.

I feel that there is some busy waiting involved. Something like most of the calls to `work()` producing no samples. Even if I set the throttle rate to something crazy low like 10 sps I still get high CPU usage on the arbitrary resampler and noise source threads.

willcode:
If no taps are specified, `history()` is 0, so when the downstream is starved, `forecast()`
`ninput_items_required[i] = noutput_items / relative_rate() + history() - 1;`
could return 0. Or even -1?

willcode:
Resampling rate of 1.0 gives low CPU, 1.0001 gives 100%.

daniestevez:
I think that if you don't specify the taps, then a sinc prototype filter is created for the taps, but I can't see where in the code this happens.

In any case, the same happens if you specify taps (I originally found this problem when using the resampler with an RRC filter). For instance, in my example it is possible to set the taps to `32*3*[1]` to have a rectangular window of 3 samples. This also gives high CPU usage.

daniestevez:
[...]

PR #4893, opened 2021-07-15
filter: use output multiple for pfb arbitrary resampler

willcode:
When interpolating, the resampler was taking up a full cpu even
at very low rates. This is due to polling when no input items are
available.

For interpolation, this change sets the output multiple to the greater
of the relative rate and the number of pfb filter arms. This is fairly
arbitrary, but results in usable output multiples and good performance.

The output multiple can only be set effectively in the block constructor.
This means that it will not change if the relative rate is set
dynamically. For widely varying rates, this could be a problem. For
rates that vary around 1.0, if the rate is initially set to 0.9 and
then changed to 1.1, performance will be as before (bad) when the rate
goes above 1.0. For this reason, and in hopes that the user will start
with 1.0 in such a case, the output multiple is set for 1.0.

Fixes https://github.com/gnuradio/gnuradio/issues/4561

### DIFF
--- a/gr-filter/lib/pfb_arb_resampler_ccc_impl.cc
+++ b/gr-filter/lib/pfb_arb_resampler_ccc_impl.cc
@@ -37,6 +37,10 @@ pfb_arb_resampler_ccc_impl::pfb_arb_resampler_ccc_impl(
 
     set_history(d_resamp.taps_per_filter());
     set_relative_rate(rate);
+    if (rate >= 1.0f) {
+        unsigned output_multiple = std::max<int>(rate, filter_size);
+        set_output_multiple(output_multiple);
+    }
 }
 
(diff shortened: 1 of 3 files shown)


mormj (approved):

mormj:
The logic here makes sense - I haven't tried it but also can't come up with any scenarios where this would cause any negative consequences.

mormj:
Let's go ahead and merge this on `master`

willcode:
@mormj Were you thinking this should not be backported?

willcode:
Possible truncation of up to 31 output samples? Maybe not even that. It's a filter, so it has history. I think it's safe, and we might want to look for other places (interpolating blocks?) that the scheduler doesn't deal with very well.

Issue #4898, opened 2021-07-16
Proposal: Deprecate tag_checker


marcusmueller:
runtime has a class tag_checker, which sorts a tag vector upon construction, and then can be used to look up a tag with a given index. 

This functionality is only used in one block internally (`chunks_to_symbols`).

This functionality also isn't very useful:

1. Tags are already sorted when they come out of `get_tags_in_range`
2. There's many less awkward ways of extracting the elements from a sorted vector that have the same key; the containers literally have methods for that, and you can use the iterator from the last call to these to start your next search.

Now, this is public API, so, deprecate it in 3.9, remove in 3.10, or do the same +0.1.

mormj:
If the method doesn't really do anything when using current GR, I would think going ahead and removing it in 3.10 - deprecating in the next 3.9 release (even though they could coincide) would make sense.

dkozel:
@marcusmueller Is tag_checker removed on master now? If so this issue can be closed.

marcusmueller:
Yep, it is, @dkozel ! Closing this issue.

PR #4901, opened 2021-07-16
digital/chunks to symbols: be less inefficient

marcusmueller:
The block formerly took the tags from `get_tags_in_range` (which are sorted),
gave them to `tag_checker`, which sorted them, and then went through every
input sample, checking its index against the next tag.

Removed the tag_checker; that saves us a sorting of a sorted vector per work call.

The other optimization is to not dispatch any tag encountered to the own
message handler; instead, the handler is called directly.

Found #4898 while doing this.

### DIFF
--- a/gr-digital/include/gnuradio/digital/chunks_to_symbols.h
+++ b/gr-digital/include/gnuradio/digital/chunks_to_symbols.h
@@ -51,9 +51,9 @@ public:
      * \param symbol_table: list that maps chunks to symbols.
      * \param D: dimension of table.
      */
-    static sptr make(const std::vector<OUT_T>& symbol_table, const int D = 1);
+    static sptr make(const std::vector<OUT_T>& symbol_table, const unsigned int D = 1);
 
-    virtual int D() const = 0;
+    virtual unsigned int D() const = 0;
     virtual std::vector<OUT_T> symbol_table() const = 0;
     virtual void set_symbol_table(const std::vector<OUT_T>& symbol_table) = 0;
 };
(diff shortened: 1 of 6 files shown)


marcusmueller:
hand-wavy Benchmarking:

benchmark_chunks_to_symbols

Vector source affinity `[1]`, C_t_S affinity `[0]`

State  |  Rate of indexes per second | Speedup over baseline
-------|--------------------------------------|--
Before| 9.28·10⁷ | 100%
After | 43.2·107 | 466%

marcusmueller:
@willcode um, you added the backport tag, but a backport will require you to undo the public header changes (which boil down to using const and unsigned where applicable); I didn't think you'd want to backport this, sorry :(

willcode:
Changing int to unsigned where the value already had to be unsigned should be compatible, right?

The const changes are private.

willcode:
@marcusmueller Given that the only user of D() is probably Python, which doesn't care ... if you think someone will use -Wsign-conversion -Werror, then we can keep the API the same and cast internally. Or, if you really don't want this backported at all, that's also fine - let me know.

marcusmueller:
> Changing int to unsigned where the value already had to be unsigned should be compatible, right?

It does change the function signature in a way that the mangling changes, so it's at least an ABI break. 

If anyone inherits from `chunk_to_symbol`, changing of signedness, at it means the function signatures aren't compatible anymore, would also be breach of API.

I don't mind this being backported, but I do think the signed->unsigned thing warrants me patching the commit.

willcode:
I always thought of the mangled signature as ABI. Will have to read up on opinions. Practically, it's likely harmless here, but I agree we should try to do the "right thing".

marcusmueller:
@willcode #4905 :)

marcusmueller:
> I always thought of the mangled signature as ABI.

And I'd agree. 

But the thing is: in C++, you can't override a `unsigned int foobar(int)`  method with a `int foobar(unsigned int)` function, or use them in the same `std::function` type and so on; separate types!

mormj (approved):

PR #4909, opened 2021-07-17
gr-digital: Fix constellation normalization by average power

### LINKED ISSUE #4892: Constellation power normalization and incorrect soft decision LUT for M-QAM with M > 4
Hello all,

I have a question about the constellation power normalization in [constellation.cc](https://github.com/gnuradio/gnuradio/blob/master/gr-digital/lib/constellation.cc). If using QPSK with power normalization (lines 53-63), the resulting constellation will contain (+-sqrt(2) +-sqrt(2)*j). Doesn't this constellation have an average power of 4?

Secondly, I believe the soft decision LUT generation in this file is incorrect for constellations greater QPSK, because it assumes the constellation is normalized by the max amplitude, and not the average amplitude (line 246), which in the case of QPSK are the same.

dave-pi:
Correct power normalization calculation.
Add a qa test for amplitude and power normalization.

Partially fixes #4892

### DIFF
--- a/gr-digital/lib/constellation.cc
+++ b/gr-digital/lib/constellation.cc
@@ -45,25 +45,26 @@ constellation::constellation(std::vector<gr_complex> constell,
       d_lut_scale(0)
 {
     unsigned int constsize = d_constellation.size();
-    float summed_mag = 0;
     switch (normalization) {
-    case NO_NORMALIZATION:
+    case NO_NORMALIZATION: {
         break;
-
-    case POWER_NORMALIZATION:
+    }
+    case POWER_NORMALIZATION: {
         // Scale constellation points so that average power is 1.
+        float summed_power = 0;
         for (unsigned int i = 0; i < constsize; i++) {
             gr_complex c = d_constellation[i];
-            summed_mag += std::norm(c);
+            summed_power += std::norm(c);
         }
-        d_scalefactor = constsize / sqrt(summed_mag);
+        d_scalefactor = sqrt(constsize / summed_power);
         for (unsigned int i = 0; i < constsize; i++) {
             d_constellation[i] = d_constellation[i] * d_scalefactor;
         }
         break;
-
-    case AMPLITUDE_NORMALIZATION:
+    }
+    case AMPLITUDE_NORMALIZATION: {
         // Scale constellation points so that average magnitude is 1.
+        float summed_mag = 0;
         for (unsigned int i = 0; i < constsize; i++) {
             gr_complex c = d_constellation[i];
             summed_mag += std::abs(c);
@@ -73,7 +74,7 @@ constellation::constellation(std::vector<gr_complex> constell,
             d_constellation[i] = d_constellation[i] * d_scalefactor;
         }
         break;
-
+    }
     default:
         throw std::runtime_error("Invalid constellation normalization type.");
     }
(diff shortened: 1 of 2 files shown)


willcode:
@dave-pi maint-3.8 has a little different code. Could you see if it's wrong too in a different way?

dave-pi:
> @dave-pi maint-3.8 has a little different code. Could you see if it's wrong too in a different way?
> 
> Looking again ... 3.8 does not have `case POWER_NORMALIZATION` so probably not applicable.

Not applicable. Only change was to replace:
```c
sqrt(c.real() * c.real() + c.imag() * c.imag());
```
with:
```c
std::abs
```

dave-pi:
qa test not right, please do not merge...

dave-pi:
> qa test not right, please do not merge...

Think it's OK now.

jdemel (approved):
LGTM.
This code is certainly an improvement.

Beyond this PR:
It'd be necessary to clean up some more code in `constellation.cc` in another PR. Maybe this would be a "good first issue" thing. e.g. `summed_mag` is a terrible variable name because it is actively misleading in case of power normalization. `summed_power` would be correct here. Also, it shouldn't be initialized, nor declared, until it is used.

dave-pi:
> LGTM.
> This code is certainly an improvement.
> 
> Beyond this PR:
> It'd be necessary to clean up some more code in `constellation.cc` in another PR. Maybe this would be a "good first issue" thing. e.g. `summed_mag` is a terrible variable name because it is actively misleading in case of power normalization. `summed_power` would be correct here. Also, it shouldn't be initialized, nor declared, until it is used.

Latest commit incorporates these changes.

gr-digital/lib/constellation.cc
```diff
@@ -131,7 +132,7 @@ float constellation::get_distance(unsigned int index, const gr_complex* sample)
 
 unsigned int constellation::get_closest_point(const gr_complex* sample)
 {
-    unsigned int min_index = 0;
+    unsigned int min_index;
     float min_euclid_dist;
     float euclid_dist;
```
jdemel, line 137: Now that you touched this function, I had a look at it.

https://github.com/gnuradio/gnuradio/blob/7a42de461b4d312fb736abbeeaacd1388a592704/gr-digital/lib/constellation.cc#L133-L149

I suggest to combine declaration and initialization. at least for `min_index` and `min_euclid_dist`. These should be the next 2 lines (L139-140) anyways.

In case of `euclid_dist`, I'm tempted to suggest to move its declaration to line L142. It is more precise to have its declaration there because the variable is only used in the for loop. I doubt this has any negative performance implications.

mormj:
@dave-pi - do you still plan to clean up the `get_closest_point` method?

dave-pi:
> @dave-pi - do you still plan to clean up the `get_closest_point` method?

Sorry thought I had done that already. Yes, will take care of it.

dave-pi:
> > @dave-pi - do you still plan to clean up the `get_closest_point` method?
> 
> Sorry thought I had done that already. Yes, will take care of it.

Done. Did some quick experiments with 3.9 vs master to test functionality.

mormj (approved):

gr-digital/python/digital/qa_constellation.py
```diff
@@ -177,6 +178,26 @@ def setUp(self):
...
+
+            points = np.array(constel.points())
+            avg_power = np.sum(abs(points)**2) / len(points)
+            self.assertAlmostEqual(avg_power, 1.0, 6)
+            constel = digital.constellation_rect(constel_points, code, rot_sym,
+                                                 side, side, width, width,
+                                                 constellation.AMPLITUDE_NORMALIZATION)
+            points = np.array(constel.points())
+            avg_amp = np.sum(abs(points)) / len(points)
+            self.assertAlmostEqual(avg_amp, 1.0, 6)
```
mbr0wn, line 199: I prefer
```suggestion
            self.assertAlmostEqual(avg_amp, 1.0, places=6)
```
...because that method has so many ways of defining "almost". But not a hold-up. This is just my preference.

mbr0wn (approved):

mbr0wn (approved):

PR #4910, opened 2021-07-17
runtime: deprecate tag_checker (removal in 3.10)

marcusmueller:
`tag_checker` used to be meant for cases where you've gotten an unsorted
tag vector (mostly, from `get_tags_in_range`), which you then had to go
through in parallel to your samples, to check which tag applies at what
sample.

For that it sorts the tags; the tags coming from the `get_tags*` functions
are sorted already

The checking pattern (which `chunks_to_symbols` is the last consumer of)
is inefficient: instead of taking the index of the first unprocessed tag
and processing all samples up to that index, a check is performed on
every sample, which includes calls and multiple indirections.

So, since very likely nobody uses this, and because it's a design
anti-pattern, deprecating this.

Together with the actual removal, addresses #4898.

Backport https://github.com/gnuradio/gnuradio/pull/4917 included.

### DIFF
--- a/.github/workflows/make-test.yml
+++ b/.github/workflows/make-test.yml
@@ -53,16 +53,16 @@ jobs:
         include:
           - distro: 'Ubuntu 20.04'
             containerid: 'gnuradio/ci:ubuntu-20.04-3.9'
-            cxxflags: -Werror
+            cxxflags: -Werror -Wno-error=deprecated-declarations
           - distro: 'Fedora 33'
             containerid: 'gnuradio/ci:fedora-33-3.9'
-            cxxflags: ''
+            cxxflags: -Werror -Wno-error=deprecated-declarations
           - distro: 'CentOS 8.3'
             containerid: 'gnuradio/ci:centos-8.3-3.9'
-            cxxflags: -Werror
+            cxxflags: -Werror -Wno-error=deprecated-declarations
           - distro: 'Debian 10'
             containerid: 'gnuradio/ci-debian-10-3.9:1.0'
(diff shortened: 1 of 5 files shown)


willcode:
If this is to be deprecated in 3.10, then you want it backported to 3.9, right? Or at least if the compilers don't Werr on it.

marcusmueller:
@willcode  this is already against maint-3.9; on master, we remove.

willcode:
Should have looked up top. Is there a way to turn of Werror on only certain warnings?

marcusmueller:
yep, I did the same to keep things building on master, where I think Fed33 fails due to deprecation warnings in ZMQ. Wait, let me figure that out and include it here

willcode:
Also `.github/workflows/pkg-debian.yml` and `.github/workflows/pkg-fedora.yml`?

marcusmueller:
> Also `.github/workflows/pkg-debian.yml` and `.github/workflows/pkg-fedora.yml`?

Addressed, @willcode :) !

willcode:
3.9.2.0 has been release, and I'm not organized to have a running changelog ... so you can rm that line and I'll put it in the changelog manually. Or just move it up above everything else and bootstrap me being organized :-)

marcusmueller:
> 3.9.2.0 has been released, 

aaaah! Oops!

> and I'm not organized enough to have a running changelog ... so you can rm that line

I will!

> and I'll put it in the changelog manually. Or just move it up above everything else and bootstrap me being organized :-)

Nah, let's not disturb your circles!

PR #4911, opened 2021-07-17
digital: constellation_receiver_cb rework

marcusmueller:
Removes `tag_checker`

Improves performance

### DIFF
--- a/gr-digital/lib/constellation_receiver_cb_impl.cc
+++ b/gr-digital/lib/constellation_receiver_cb_impl.cc
@@ -8,6 +8,8 @@
  *
  */
 
+#include <cstddef>
+#include <cstdint>
 #ifdef HAVE_CONFIG_H
 #include "config.h"
 #endif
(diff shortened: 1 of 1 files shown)


mormj (approved):

ThomasHabets (approved):

jdemel (approved):
LGTM.

Is this a deprecated block?

marcusmueller:
@jdemel no, not that I'd be aware of. Why?

jdemel:
Because I thought the modulator is deprecated. Also, this block performs a few distinct things. demodulation with hard decision and computes error margins. I'd argue these are several processing steps. Beginners would have a hard time figuring out how such a block works. Since this block computes symbol-wise results, it is not the best solution to in terms of performance for people who know exactly what they do. It is somewhere in between.

marcusmueller:
> Because I thought the modulator is deprecated. 

You've got me confused there! *Which* modulator?

> Also, this block performs a few distinct things. demodulation with hard decision and computes error margins. I'd argue these are several processing steps. 

I must admit I don't really see how – this is decision feedback phase sync. You need to directly feed back the error margins: 

        sym_value = d_constellation->decision_maker_pe(&sample, &phase_error);
        phase_error_tracking(phase_error); // corrects phase and frequency offsets

Here, the phase error occurs "naturally" inside the demod. It's a Costas loop! Of sorts at least, because Mr. Costas made that for PSKs only, and this works for any constellation.

> Since this block computes symbol-wise results, it is not the best solution to in terms of performance for people who know exactly what they do. It is somewhere in between.

Well, but it *is* a Costas loop. So, the error performance should be the same as for our costas loop block (for PSKs). A fair question would be "why is there two Costas Loops", but I argued to myself that the BPSK/QPSK Costas loop was such a special case that the optimizable implementation we have in `costas_loop_cc` deserved its own block; this indeed is a bit of an unusual use case, but I see why I'd want to use it – especially to calculate mean phase noise variance on QAM-style modulations, which is a nice metric in nonlinear channels.

PR #4912, opened 2021-07-17
Remove tag_checker

marcusmueller:
Deprecated on 3.9 in #4910

Depends on #4901 , #4911 to remove the last two consumers of it.

> `tag_checker` used to be meant for cases where you've gotten an unsorted tag
> vector (mostly, from `get_tags_in_range`), which you then had to go through in
> parallel to your samples, to check which tag applies at what sample.
> 
> For that it sorts the tags; the tags coming from the `get_tags*` functions are
> sorted already
> 
> The checking pattern (which `chunks_to_symbols`/a constellation receiver are
> the last consumers of) is inefficient: instead of taking the index of the first
> unprocessed tag and processing all samples up to that index, a check is
> performed on every sample, which includes calls and multiple indirections.
> 
> So, since very likely nobody uses this, and because it's a design anti-pattern,
> deprecating this on 3.9 and removing it with 3.10.


marcusmueller:
Draft until the consumers are removed; then this will need rebasing

willcode:
Did you want the constellation_receiver changes in the rm tag_checker PR?

marcusmueller:
@willcode  it *is* a dependency, so "yes", but I hope to rebase it away, hence the "needs rebase" tag, and the draft status :)

marcusmueller:
@willcode with these commits being in upstream, this was boiled down to a single commit through a rebase now :) Thanks for the patience :)

PR #4920, opened 2021-07-18
Disable metainfo_test until we've decided how to proceed (backport to maint-3.9)

marcusmueller:
see issue https://github.com/gnuradio/gnuradio/issues/4918 ; it's not really our fault this test fails, but it's our problem.


willcode (commented):
How about using `appstream-util validate-relax` instead of `appstream validate`. Then the test can stay enabled.

PR #4921, opened 2021-07-18
GRC: Test: Relax the checking of the metainfo file

### LINKED ISSUE #4918: Metainfo test fails because of appstreamcli upstream bug

marcusmueller:
This fixes a CI failure I introduced when appstreamcli's default checking started to reject mailto: contacts

Closes #4918


willcode (approved):
On Fedora, appstream-util comes with libappstream-glib. On Ubuntu, it is in the appstream-util deb. CI passes and we don't explicity include appstream, so this should be safe. Guessing it is a dep on one of the gtk-related packages.

marcusmueller:
awesome. Still adding libappstream-glib to the Fedora container. No sense in having a check that never runs.

mormj (approved):

PR #4934, opened 2021-07-19
digital: Changed soft decision LUT to scale off maximum point in constellation

### LINKED ISSUE #4892: Constellation power normalization and incorrect soft decision LUT for M-QAM with M > 4

jfmadeira:
The constellation can have points with coordinates larger than 1, even if normalized for 64-QAM and above.
Therefore, the LUT should be calculated for the largest coordinate.

Partially fixes #4892

### DIFF
--- a/gr-digital/lib/constellation.cc
+++ b/gr-digital/lib/constellation.cc
@@ -243,14 +243,14 @@ void constellation::gen_soft_dec_lut(int precision, float npwr)
     d_soft_dec_lut.clear();
     d_lut_scale = powf(2.0f, static_cast<float>(precision));
 
-    // We know we've normalized the constellation, so the min/max
-    // dimensions in either direction are scaled to +/-1.
-    float maxd = 1.0f;
-    float step = (2.0f * maxd) / (d_lut_scale - 1);
-    float y = -maxd;
-    while (y < maxd + step) {
-        float x = -maxd;
-        while (x < maxd + step) {
+    // We assume the constellation is symmetrical and has no offset
+    // Search for the largest coordinate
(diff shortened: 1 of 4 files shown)


jfmadeira:
Do not merge yet, still needs better review. 
After changing the max values of `constellation::gen_soft_dec_lut`, I realized there were rounding errors in the integration loop that could cause the loop to run one more time than expected. I decreased the threshold to `step / 2`, to stop this from happening, while still ensuring the last point will exist. 

In qa_constellation_soft_decoder_cf.py all functions expect complex constellations, which requires that the Es for real constellations be adjusted to compensate. This might be better fixed by separating the testing of real and complex constellations, or adding support for real constellations in the soft_dec_lut_gen.py. This is not a problem in the cpp code however.

mormj:
@jfmadeira - do you have more to do on this PR, or is it ready to have others review it?  Currently it is marked as draft.  Thanks!

mormj:
@dave-pi - would you mind taking a look at this one since it relates to your #4909

PR #4935, opened 2021-07-19
CI: Update the Fedora Container to F34

marcusmueller:
this has the main advantage of bringing UHD4, so we can test more of gr-uhd.

It's uncertain whether we'd want to backport this right now – F33 is still supported and maybe it's more important to test maint-3.9 with the current-at-release system than with UHD4.
It might thus make sense to hold off until the next 3.9.X release.


willcode:
Agree we should keep F33 for now. I build F34 myself and will notice if something terrible happens (but with UHD4). F35 will be out in Oct.

mbr0wn:
Why can't we have both?

marcusmueller:
I thought I had answered, but doesn't look that way:

> Why can't we have both?

@mbr0wn  I thought we only hate finite CPU resources. I'd rather have builds that finish sooner than multiple highly correlated builds; but that was just a wild guess on my side!

mbr0wn:
> @mbr0wn I thought we only hate finite CPU resources. I'd rather have builds that finish sooner than multiple highly correlated builds; but that was just a wild guess on my side!

Builds run in parallel; I don't think this will delay anything during builds.

mbr0wn:
> Are we switching back from the `/ci:` to the `/ci-` images? I thought that was only for Debian 10 which has a mysterious failure and didn't get updated.

I thought so too.
> 
> More Fedora people should be on 34 now, so good to move 3.9 testing up.

:+1:

marcusmueller:
the colon seems to get automagically converted to `-` if there's no ambiguity arising from an attached tag name, so it's not something you can rely on. That's why I removed it, after it took me a while to figure out.

willcode:
These are two completely different sets of images, build from different Dockerfiles, aren't they?

marcusmueller:
> These are two completely different sets of images, build from different Dockerfiles, aren't they?

I don't think so, but might be confused. That's what the gnuradio dockerhub repo entails:

Untitled

marcusmueller:
You might be right, the ci:something might be tags of the ci container. https://hub.docker.com/repository/docker/gnuradio/ci

willcode:
Yes, they are tags.

marcusmueller:
which means I need to push the things I pushed to ci-something as ci:something,  fix the yml here (and PR #4981).

.github/workflows/make-test.yml
```diff
@@ -54,8 +54,8 @@ jobs:
           - distro: 'Ubuntu 20.04'
             containerid: 'gnuradio/ci:ubuntu-20.04-3.9'
             cxxflags: -Werror -Wno-error=invalid-pch
-          - distro: 'Fedora 33'
-            containerid: 'gnuradio/ci:fedora-33-3.9'
+          - distro: 'Fedora 34'
+            containerid: 'gnuradio/ci-fedora-34-3.9:1.0'
```
mbr0wn, line 58:
```suggestion
          - distro: 'Fedora 33'
            containerid: 'gnuradio/ci:fedora-33-3.9'
          - distro: 'Fedora 34'
            containerid: 'gnuradio/ci:fedora-34-3.9'            
```

.github/workflows/pkg-fedora.yml
```diff
@@ -26,8 +26,8 @@ jobs:
         # descriptive name, and one key 'containerid' with the name of the
         # container (i.e., what you want to docker-pull)
         include:
-          - distro: 'Fedora 33'
-            containerid: 'gnuradio/ci:fedora-33-3.9'
+          - distro: 'Fedora 34'
+            containerid: 'gnuradio/ci-fedora-34-3.9:1.0'
```
mbr0wn, line 30:
```suggestion
          - distro: 'Fedora 33'
            containerid: 'gnuradio/ci:fedora-33-3.9'
          - distro: 'Fedora 34'
            containerid: 'gnuradio/ci:fedora-34-3.9' 
```

mbr0wn:
We had different ways of doing Docker builds. Now we use autobuilds, and they all have the same tag. I added some suggestions in the review.

marcusmueller:
@mbr0wn  how do these autobuilds work? What do I need to do to trigger one?

willcode:
Dockerhub will find the change in Github and redo the build automatically.

mbr0wn:
I thought containers get autobuilt when you push to master? @noc0lour set it up.

marcusmueller:
> I thought containers get autobuilt when you push to master? @noc0lour set it up.

Not anymore:
https://github.com/gnuradio/gnuradio/issues/5051

Opting to not to shave the fix-autobuilds yak right now. So, also, makes no difference whether we use ci:something or ci-anything as all is transient.

.github/workflows/pkg-fedora.yml
marcusmueller: These tags do not exist, so sadly not possible :(

willcode:
You mean back to build-container-on-the-fly?

noc0lour:
> > I thought containers get autobuilt when you push to master? @noc0lour set it up.
> 
> Not anymore:
> #5051
> 
> Opting to not to shave the fix-autobuilds yak right now. So, also, makes no difference whether we use ci:something or ci-anything as all is transient.

It's easier to trace if you have one repo with multiple tags vs a lot of new image repositories. Just my two cents.

.github/workflows/pkg-fedora.yml
mbr0wn: They do now.

.github/workflows/make-test.yml
mbr0wn: Booyah.

mbr0wn (approved):

mbr0wn:
Meh weird github auto-close. Will replace this.

PR #4950, opened 2021-07-23
Backport equalizers to maint-3.8

jaredd:
This PR essentially cherry-picked 3c1750c481a578bd831fce933502fde88444eb54 and fixed up the pointer types to conform to the 3.8 implementation. [GRC example from the wiki](https://wiki.gnuradio.org/index.php/Linear_Equalizer) works.


jaredd:
It appears that the checks are failing due to an outdated volk. I'm not sure how to fix that within the context of the automated tests. It seems volk v2.1.0 is sufficient.

jaredd:
It appears that tag v2.1.0 up to 353d2a167f3f71e1068840e8c9c8ad12eb27de63 will support Python2. Not sure how to enforce that using submodules. Could I check out, say, the volk commit prior to 353d2a167f3f71e1068840e8c9c8ad12eb27de63 and then commit that to this PR?

jaredd:
Ok, I think I finally got the right volk to support both Python2 and Python3 gnuradio/volk@e9d72f0d9581231671f7921b48c70451fe2f64a4.

willcode:
Overall, I'm ok with backporting this, as long as it all works as intended.

VOLK 2.0 should be 1299d72c, which is v2.0.0. Changing it to anything else would require a bigger discussion.

Backporting into deprecation ... I don't know the right way to handle this since it's still there in 3.9.

jaredd:
> Overall, I'm ok with backporting this, as long as it all works as intended.
> 
> VOLK 2.0 should be 1299d72c, which is v2.0.0. Changing it to anything else would require a bigger discussion.
> 
> Backporting into deprecation ... I don't know the right way to handle this since it's still there in 3.9.
I don't know what "backporting into deprecation" means, but the equalizer implementations definitely require a volk newer than v2.0.0. I found v2.1.0 worked, but may not be necessary. 

I don't know the implications of changing the volk version. I could really use the equalizers on a project that's 3.8-based. I am only using Python3 so it seems to work-for-me ™. I imagine a number of people would be in my situation, though I clearly am out of my depth trying to get both Python2 and Python3 to work.

willcode:
We're not encouraging new features on 3.8. If a new feature definitely works without any risk of messing up something else, then it's possible. Changing the version of a dependency (volk) is not something to do without some thought.

"backporting into deprecation": the PR marks this "new" feature for 3.8 as deprecated (via the GRC categories) because it is going to be removed 2 versions in the future (3.10).

jaredd:
> We're not encouraging new features on 3.8. If a new feature definitely works without any risk of messing up something else, then it's possible. Changing the version of a dependency (volk) is not something to do without some thought.
> 
> "backporting into deprecation": the PR marks this "new" feature for 3.8 as deprecated (via the GRC categories) because it is going to be removed 2 versions in the future (3.10).

Hmm, that's pretty funny. The equalizers are going away in 3.10 or re-implemented?

willcode:
For examle `gr-digital/grc/digital_cma_equalizer_cc.block.yml`. It looks like you picked up the future deprecations of digital_cma_equalizer_cc, digital_kurtotic_equalizer_cc, and digital_lms_dd_equalizer_cc. Should be easy enough to iron that one out. I think equalizers were combined the same way clock recovery was.

Will things work with VOLK 2.0.0?

jaredd:
> Will things work with VOLK 2.0.0?

No I think it has to be at least gnuradio/volk@caebda4e1f28182f46b2cc3f60044c203a15ef74 where [gnuradio/volk@include/volk_alloc.h](https://github.com/gnuradio/volk/blob/caebda4e1f28182f46b2cc3f60044c203a15ef74/include/volk/volk_alloc.h) was introduced. Even at that point some changes to includes need to be made.

v2.1.0 would work as the next tag.

jaredd:
Perhaps there's a way with cmake to conditionally enable these in the event someone wants to update volk themselves? Can condition on Volk_VERSION >= 2.1.0?

mormj:
Given the complications with VOLK, I would not think backporting in-tree to 3.8 would be preferred - this was based off and OOT: https://github.com/perspectalabs/gr-equalizers - would it be sufficient to fork and update the OOT and work with that in your project?

jaredd:
> Given the complications with VOLK, I would not think backporting in-tree to 3.8 would be preferred - this was based off and OOT: https://github.com/perspectalabs/gr-equalizers - would it be sufficient to fork and update the OOT and work with that in your project?

I suppose so, I'll take a look.

willcode:
Closing - if there is a solution that doesn't require a newer VOLK, please reopen.

Issue #4944, opened 2021-07-21
Some Python blocks may be able to be moved into C++


ncorgan:
There are a bunch of blocks (mostly hierarchical) developed in Python, I assume out of ease of development. This stops them from being used by C++ clients (or other hierarchical blocks). In theory, some of the could be reimplemented on the C++ layer. With GNU Radio's Python bindings (and maybe an interim compatibility `.py` file if necessary), these shouldn't break any Python clients, and the QA tests should still work.

Some of them use Python-specific stuff (socket stuff comes to mind) but certainly not all. These days, y'all seem to be more open to change than in the past, so I thought I'd bring it up.

mormj:
Can you start a checklist here of blocks that would be good candidates?

ncorgan:
I've gone through the first few modules and listed the technically viable blocks. I'll continue tomorrow.

ncorgan:
@mormj Done

solomonbstoner:
@ncorgan is the suggestion to create hier blocks in c++ to replace these hier blocks currently in python?

willcode:
Not totally related to this issue, but throwing it out there ...

When a block is used in a hier block, there will be only one downstream block. In-place operations are possible in many cases then, if blocks are written to support this. Separating the work "kernel" of a block from the `work()` of a block would allow multiple functions to be called from a hier block without the need to more threads, buffers, latency. This would not require (any? much?) scheduler work, just some changes to hier and participating blocks, and it would all be backward compatible.

ncorgan:
@solomonbstoner Correct. It should be possible to do this without breaking Python clients.

@willcode Ah, I wasn't aware of any non-block changes that would be needed. I assumed it would be a 1-to-1 conversion to some extent. That's obviously something to take into account for the necessary effort. FWIW, this doesn't apply to all Python conversions proposed above.

willcode:
@ncorgan I don't think there's any non-block work required for your idea (need to read all the blocks to know for sure). My comment was additional idea.

willcode:
@ncorgan Yah, there is a bunch of util stuff in python that would take some effort to port, and python versions would have to remain for backward compatibility. Could start with the easy/most valuable ones.

drmpeg:
[...]

PR #4951, opened 2021-07-24
grc: Fix cpp hier param yaml template

solomonbstoner:
The following issues are fixed for the cpp hier block parameter yaml
template generation.
1. The key of the `data` dictionary containing the parameters should be
`parameters` instead of `param`.
2. Fix the yaml template for the parameter to generate `${param}`
instead of `$param` in the yaml file.

Changes made as part of the effort for #4944 .

### DIFF
--- a/grc/core/generator/cpp_hier_block.py
+++ b/grc/core/generator/cpp_hier_block.py
@@ -91,7 +91,7 @@ class CppHierBlockGenerator(CppTopBlockGenerator):
             p['dtype'] = param_block.params['value'].dtype
             p['default'] = param_block.params['value'].get_value()
             p['hide'] = param_block.params['hide'].get_value()
-            data['param'].append(p)
+            data['parameters'].append(p)
 
         # Ports
         for direction in ('inputs', 'outputs'):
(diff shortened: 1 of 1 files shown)


dl1ksv (changes requested):
With your fix I can generate cpp code, but the code generates 
typedef boost::shared_ptr<detectMarkSpace> detectMarkSpace_sptr;

Shouldn't we generate std::shared_ptr instead, at keast for the master branch ?

haakov (approved):
@dl1ksv The boost -> std change should probably be placed in a different PR so this one can be backported.

Changes look good to me

dl1ksv:
I agree

dl1ksv (approved):

solomonbstoner:
Thanks for reviewing, guys. Yeah, I was going with solving one problem at a time in each PR. Feels less messy to me

solomonbstoner:
Just realized that my commit was dated 1970 due to some clock error on my computer. Recommited the patch to put the date back to 2021. No change to the code that was approved

Screenshot from 2021-08-04 10-10-41

PR #4965, opened 2021-07-29
pybind11: remove debug message

japm48:
Remove the annoying 'output docstrings to {path}' message.
Hopefully, this is no longer needed, otherwise, I'll try to make a 'verbose' compile-time option.


mormj (approved):
Yes these can definitely be removed - was helpful long ago.

PR #4969, opened 2021-07-31
gr-digital: replace clk recovery MM in hier blks

### LINKED ISSUE #4962: gr-digital: Replace clock recovery MM with Symbol Sync?
Since the Clock Recovery MM block is deprecated, should the [gfsk](https://github.com/gnuradio/gnuradio/blob/master/gr-digital/python/digital/gfsk.py) and [gmfsk](https://github.com/gnuradio/gnuradio/blob/master/gr-digital/python/digital/gmsk.py) demod hier blocks have their clock recovery MM replaced with the Symbol Sync block?

If so, shall the hier block parameters be changed to accept the parameters used by Symbol Sync block? (e.g. omega gain and mu gain to be replaced with loop bandwidth and damping factor as input) Or shall they remain unchanged (for backwards compatibility), meaning the hier block converts the parameters internally as shown in [slide 32](https://www.gnuradio.org/grcon/grcon17/presentations/symbol_clock_recovery_and_improved_symbol_synchronization_blocks/Andy-Walls-Samples-to-Digital-Symbols.pdf)?

solomonbstoner:
Fixes #4962. However, I have no idea what kind of test would prove that its performance is the same as the Clock Recovery MM block.

As Clock Recovery MM is deprecated, this commit replaces them
in the hier blocks, gmfsk demod and gfsk demod, with Andy Wall's Symbol
Sync block. Backward compatibility is maintained by keeping the
parameters unchanged, and converting the two gain values to the
corresponding loop bandwidth value that Symbol Sync requires.

**Note:** This means the `mu` parameter is no longer needed. Should I leave it in or take it out? I left it in for backward compatibility, but any value entered in the block will have no impact on the demod block.

My derivation:
math_for_PI_loop_filter

### DIFF
--- a/gr-digital/grc/digital_gfsk_demod.block.yml
+++ b/gr-digital/grc/digital_gfsk_demod.block.yml
@@ -16,7 +16,7 @@ parameters:
     dtype: real
     default: '0.175'
 -   id: mu
-    label: Mu
+    label: Mu (Unused)
     dtype: real
     default: '0.5'
 -   id: omega_relative_limit
--- a/gr-digital/grc/digital_gmsk_demod.block.yml
+++ b/gr-digital/grc/digital_gmsk_demod.block.yml
@@ -12,7 +12,7 @@ parameters:
     dtype: real
     default: '0.175'
 -   id: mu
-    label: Mu
+    label: Mu (Unused)
     dtype: real
     default: '0.5'
 -   id: omega_relative_limit
--- a/gr-digital/python/digital/gfsk.py
+++ b/gr-digital/python/digital/gfsk.py
@@ -12,6 +12,7 @@
 
 
 # See gnuradio-examples/python/digital for examples
+from math import log as ln
 
 import numpy
 
@@ -190,7 +191,7 @@ class gfsk_demod(gr.hier_block2):
 
         Args:
             gain_mu: controls rate of mu adjustment (float)
-            mu: fractional delay [0.0, 1.0] (float)
+            mu: unused but unremoved for backward compatibility (unused)
             omega_relative_limit: sets max variation in omega (float, typically 0.000200 (200 ppm))
             freq_error: bit rate error as a fraction
             float:
@@ -202,7 +203,6 @@ class gfsk_demod(gr.hier_block2):
 
         self._samples_per_symbol = samples_per_symbol
         self._gain_mu = gain_mu
-        self._mu = mu
         self._omega_relative_limit = omega_relative_limit
         self._freq_error = freq_error
         self._differential = False
@@ -217,15 +217,27 @@ class gfsk_demod(gr.hier_block2):
 
         self._gain_omega = .25 * self._gain_mu * self._gain_mu        # critically damped
 
+        self._damping = 1.0
+        self._loop_bw = -ln((self._gain_mu + self._gain_omega)/(-2.0) + 1)        # critically damped
+        self._max_dev = self._omega_relative_limit * self._samples_per_symbol
+
         # Demodulate FM
         #sensitivity = (pi / 2) / samples_per_symbol
         self.fmdemod = analog.quadrature_demod_cf(1.0 / sensitivity)
 
         # the clock recovery block tracks the symbol clock and resamples as needed.
         # the output of the block is a stream of soft symbols (float)
-        self.clock_recovery = digital.clock_recovery_mm_ff(self._omega, self._gain_omega,
-                                                           self._mu, self._gain_mu,
-                                                           self._omega_relative_limit)
+        self.clock_recovery = self.digital_symbol_sync_xx_0 = digital.symbol_sync_ff(digital.TED_MUELLER_AND_MULLER,
+                            self._omega,
+                            self._loop_bw,
+                            self._damping,
+                            1.0,  # Expected TED gain
+                            self._max_dev,
+                            1,  # Output sps
+                            digital.constellation_bpsk().base(),
+                            digital.IR_MMSE_8TAP,
+                            128,
+                            [])
 
         # slice the floats at 0, outputting 1 bit (the LSB of the output byte) per sample
         self.slicer = digital.binary_slicer_fb()
@@ -248,10 +260,10 @@ class gfsk_demod(gr.hier_block2):
 
     def _print_verbage(self):
         print("bits per symbol = %d" % self.bits_per_symbol())
-        print("M&M clock recovery omega = %f" % self._omega)
-        print("M&M clock recovery gain mu = %f" % self._gain_mu)
-        print("M&M clock recovery mu = %f" % self._mu)
-        print("M&M clock recovery omega rel. limit = %f" % self._omega_relative_limit)
+        print("Symbol Sync M&M omega = %f" % self._omega)
+        print("Symbol Sync M&M gain mu = %f" % self._gain_mu)
+        print("M&M clock recovery mu (Unused) = %f" % self._mu)
+        print("Symbol Sync M&M omega rel. limit = %f" % self._omega_relative_limit)
         print("frequency error = %f" % self._freq_error)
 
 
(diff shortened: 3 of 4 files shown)


willcode:
We can't delete mu from 3.9, and having it there with no functionality will confuse people. Sorry - I didn't know that was going to happen. Have to think about that. @marcusmueller? Anyone else who knows clock recovery well?

solomonbstoner:
In slide 32 of [Andy Wall's presentation on his Symbol Sync block](https://www.gnuradio.org/grcon/grcon17/presentations/symbol_clock_recovery_and_improved_symbol_synchronization_blocks/Andy-Walls-Samples-to-Digital-Symbols.pdf), there is a map from the Clock Recovery MM block for real values to his Symbol Sync block. The part marked as "Math" to convert `gain mu` and `gain omega` to `omega_n_T` and `zeta` is done by my derivation shown above. Everything's been mapped over except `mu`.

dave-pi:
`mu` in the MM TED was an initial estimate of the fractional interval that was passed to the block:
https://github.com/gnuradio/gnuradio/blob/master/gr-digital/lib/clock_recovery_mm_cc_impl.cc#L31
and then used in the interpolation:
https://github.com/gnuradio/gnuradio/blob/master/gr-digital/lib/clock_recovery_mm_cc_impl.cc#L108

Don't think there is currently support for initializing that. The equivalent variable (`d_phase_wrapped`) is initialized to 0:
https://github.com/gnuradio/gnuradio/blob/master/gr-digital/lib/interpolating_resampler.cc#L28
and then used in the interpolation, which apparently lives here now:
https://github.com/gnuradio/gnuradio/blob/master/gr-digital/lib/symbol_sync_cc_impl.cc#L422

I suppose you could change the Symbol Sync block to allow for that argument (only for the case of a MM TED) that would then be passed to the resampler at construction time.

solomonbstoner:
If I go that route, I'd have to change both the Symbol Sync and interpolating resampler blocks to accept `mu` as the `d_phase_wrapped`. I could do it, but is it worth the change? Afterall, according to Tomaž Šolc's blog, `mu` doesnt seem to matter that much.

> `Mu` is the detected phase shift in samples between the receiver and transmitter. The initial value given as the block parameter **doesn't matter** in practice, since it's usually impossible to guess the phase in advance. Internally, phase is tracked between 0 to omega (corresponding to 0 and 2π). 

I could create a deprecated version of one with `mu` for 3.9 (but still using Clock Recovery MM) and have the new one without `mu` for Symbol Sync.

dave-pi:
I agree it's not useful. Maybe marginally for a developer to test the
algorithm. What I think could be considered is setting the default not to 0
but to 0.5, which is the expected value (assuming the fractional delay is
uniformly distributed). This would presumably make it converge faster.
If there is debug that outputs the error (output of the TED) maybe this
could be verified.

solomonbstoner:
Regardless of whether its default value is 0 or 0.5, it still begs the question, what should we do with the `mu` parameter of the hier blocks?

dkozel:
I don't actually know what the behavior is when flowgraph is opened in a GRC version where a parameter has been *removed* from a block. It'd be a good feature to display an warning that there was an unused value during import of the flowgraph.

awalls-cx18:
On August 2, 2021 8:16:42 AM EDT, Derek Kozel ***@***.***> wrote:
***@***.*** would you be able to give any feedback here?
>

I'll check the code tomorrow, but my recollection is that setting an initial value for mu in any practical scenario is pointless.

It's like setting the initial state of a windsock on a windy day.  The transient influence of the initial condition is so short a duration, it's a don't care when it comes to convergence of the tracking.

This nothing to agonize over.  That's why I ditched initialization of it in the symbol sync blocks.

awalls-cx18:
On August 2, 2021 8:16:42 AM EDT, Derek Kozel ***@***.***> wrote:
***@***.*** would you be able to give any feedback here?
>

Some other notes:

IIRC the slicer used in the complex MM block is just plain wrong.

Also, both versions of the MM block can errantly drop samples if operated at over 8 samples/symbol.

If anyone is concerned about preserving the behavior of the MM blocks, don't be.  Unless errant behavior is the way you like to engineer communications receivers.

awalls-cx18:
On August 2, 2021 8:16:42 AM EDT, Derek Kozel ***@***.***> wrote:
***@***.*** would you be able to give any feedback here?
>

Regarding mu's expected value, it's actually not 0.5 or knowable in general, because mu's value is a cyclic alias.  The interval -0.5 to 0.5 is just as legitimate as the interval 0.0 to 1.0.

Using the windsock analogy again, it's direction is in the interval 0 to 360 or -180 to 180.  So what's the expected value of its direction?  The answer is that it depends on the environment, not the interval of values it can take on.  Once you know the azimuth reference direction and the prevailing winds in the area, then you can talk about the expected value of the windsock's direction.

In the absence of information about the input signal, 0.0 is just as good an initial value as 0.5 or 0.999999.

solomonbstoner:
Thanks! By the way @awalls-cx18, i really enjoyed your presentation on Symbol Sync.

Since `mu` really has no purpose in Symbol Sync, and I found out that there is *no* error generated in GRC when a parameter is removed, I shall just remove `mu` from the blocks. I'll push the new update right after I post this comment. If this arrangement is agreeable with yall, I will edit the wiki to reflect the changes

willcode:
Based on Andy's response, leave mu in the 3.9 version for API compatibility but don't wire it to anything. We can note that it was found to be irrelevent in the release notes, add `(unused)` in the GRC yaml, etc., but any flowgraphs/programs setting it will still work.

marcusmueller:
> Based on Andy's response, leave mu in the 3.9 version for API compatibility but don't wire it to anything. We can note that it was found to be irrelevent in the release notes, add `(unused)` in the GRC yaml, etc., but any flowgraphs/programs setting it will still work.

We can go one step further:

Add a second make` function without the µ parameter, and mark the one with it as [`[[deprecated]]`](https://en.cppreference.com/w/cpp/language/attributes/deprecated), so that at least usage in C++ systems triggers a compile time warning.

willcode:
@marcusmueller The portion addressed in this PR is a Python-only hier block. The entire C++ block that uses mu is already deprecated.

solomonbstoner:
@willcode modified as instructed. `mu` will be accepted as an argument, but is not used anywhere in the block. 
I also realized that the gfsk blocks' documentation is not showing in the documentation tab, so I shifted it so that it will.

willcode:
The prints and the parser add_option lines related to mu need to be there for backward compatibility. I have no idea if any prints are parsed, or if any program specifies mu via an option, but they'd have to stay. We can deprecate those params in 3.9. BTW, when are we saying that clock_recovery_mm_x is going to be removed? It's marked "deprecated in 3.9" in the wiki, so guessing removal in 3.11?

solomonbstoner:
Ok. Re-added the `mu` logging and parsing, but specified that they are "(Unused)". Im not so sure about when deprecated blocks are set to be removed. At least with this PR these hier blocks wont depend on the deprecated clock recovery MM anymore :)

willcode:
Great, and we can figure out how to mark mu as deprectated later.

Now, we really do need to figure out how to test this. I don't see any qa or examples in-tree. A couple OOTs I usually build use the gfsk_mod, but I can't find anything right off that uses the demod functions. Do you have an example that uses these, or do you know of OOTs that do?

solomonbstoner:
I tested that my modification does not break the demodulation for both GMSK and GFSK using default values. The bit stream before and after the mod + demod agree with each other.

Screenshot from 2021-08-06 18-29-32
Screenshot from 2021-08-06 18-29-40

If you wanna try it out: https://pastebin.com/ZpgdEgNU

However, I am not sure how to test it under fading/phase error/timing error/freq offset conditions. The only thing I can think of is: @awalls-cx18 could check the code to make sure I have modified the blocks in the same way he intends Symbol Sync to replace Clock Recovery MM

willcode:
I added a Channel Model and tried it out with various settings. It seems to run ok, but I don't know how to quantify that.

mormj (approved):
This looks good to me, and the confirmation of the parameter re-mappings in the conversation gives confidence that this replacement is the correct decision -saving extra work when these deprecated blocks are finally removed.

willcode (approved):

PR #4940, opened 2021-07-20
gr-digital: Remove packed to unpacked gfsk mod hier block

solomonbstoner:
Fixes #4939

PR #4970, opened 2021-07-31
gr-digital: Make unpack optional for gmsk mod hier block

solomonbstoner:
Allow the unpack option to be disabled for custom, non-byte-aligned bit
streams in gmsk mod, just like how it was done for gfsk in #4940.


mormj (approved):

dkozel:
Please update the wiki documentation with this new behavior. Thanks!
https://wiki.gnuradio.org/index.php/GMSK_Mod

solomonbstoner:
Wiki edited as requested

solomonbstoner:
Local clock error caused my commit to be set to 1970. Recommitted to set the year to 2021. The patch is the same as the version approved

PR #4972, opened 2021-07-31
grc cpp generation: Establish the possibility to add find packages

dl1ksv:
Flowgraphs may use oot modules. When generating cpp code an oot module
may require a package that is not provided by gnuradio.
So I propose to add an additional entry parameters in the cpp template
to provide a list of package names and add the corresponding find package calls to CMakeLists.txt

### DIFF
--- a/grc/core/blocks/_build.py
+++ b/grc/core/blocks/_build.py
@@ -57,6 +57,7 @@ def build(id, label='', category='', flags='', documentation='',
         callbacks=cpp_templates.get('callbacks', []),
         var_make=cpp_templates.get('var_make', ''),
         link=cpp_templates.get('link', []),
+        packages=cpp_templates.get('packages', []),
         translations=cpp_templates.get('translations', []),
         declarations=cpp_templates.get('declarations', ''),
     )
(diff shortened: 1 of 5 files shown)


haakov:
I haven't had time to test this yet, but it looks good at first glance.

mormj:
@dl1ksv - can you provide an example of a good use case to test this out with?

dl1ksv:
@mormj sorry for the delay, but I had problems with find_package, which did not find my oot package any more.
I'm unsure if there was a modification to cmake, but find_package only finds my package funcube if the file funcubeConfig.cmake can be found in a subdirectory named funcube.

To test you have to checkout and build the  [cpp_code_gen_test](https://github.com/dl1ksv/gr-funcube.git) branch.

Then you can use this simple example
[oot_cpp.grc.gz](https://github.com/gnuradio/gnuradio/files/7083197/oot_cpp.grc.gz)

To build this example you don't need a funcube dongle, but if you run this example without a dongle attached, you'll get a runtime-error.

mormj (approved):

Issue #4134, opened 2021-01-22
Move from log4cpp (implement GREP 0013)


marcusmueller:
https://github.com/gnuradio/greps/blob/master/grep-0013-remove-log4cpp.md specifies we want to move away from log4cpp.

Hottest replacement candidate seems to be spdlog (which Marcus, Martin and others have worked with before, so not a blind guess). That depends on fmtlib, which in itself could replase `boost::format` in a lot of places, and as its contained in `std::format` (nearly completely) in C++20, one day become a zero-overhead-dependency.

dkozel:
Putting a comment here to link log4cpp work/issues.

PR to remove log4cpp:
https://github.com/gnuradio/gnuradio/pull/4981

Issues:
https://github.com/gnuradio/gnuradio/issues/4681

PR #4981, opened 2021-08-04
Good riddance log4cpp

marcusmueller:
Been letting this linger long enough. With the prior work in #4779, this really wasn't that painful. (should we decide to merge #4779, the review load of this boils down to four files of C++ and a bit of CMake).

Still need to fix places where iomanip needs to be installed (these are separate bugs, honestly). Probably, using a compile-time parsed format string for the default logging pattern causes more trouble than it's worth, so might revert that back to a `constexpr std::string`. 

Closes https://github.com/gnuradio/gnuradio/issues/4134

**Edit**: since the discussion chain here is getting long, using :+1: to mark things I've read / addressed for myself. Not meaning anything in particular.


mormj:
Very excited about this one!  couple of questions:
1) should we stick with the logging macros `GR_LOG_DEBUG`, etc. so that logging can be completely compiled out?
2) since spdlog has fmt based string formatting built in, e.g. `warn("variable set to {}",var)` - is the plan here to migrate away from boost format entirely in the logging

marcusmueller:
Currently blocked by spdlog not being installed on the build hosts containers; new docker images are being uploaded currently.

marcusmueller:
I'm out of my depth. @noc0lour you don't happen to have an idea why I'm getting https://github.com/gnuradio/gnuradio/pull/4981/checks?check_run_id=3334922175#step:2:16

    invalid reference format


Seems like docker is unhappy about the `containerid`s

- `gnuradio/ci:ubuntu-20.04-3.9:0.5`,
- `gnuradio/ci:fedora-33-3.9:1.1` and
- `gnuradio/ci:centos-8.3-3.9:1.2`,

but

- `gnuradio/ci-debian-10-3.9:1.1` works, and so does
- `image: 'gnuradio/ci-ubuntu-20.04-3.9:0.5'` in the doxygen job

marcusmueller:
>     1. should we stick with the logging macros `GR_LOG_DEBUG`, etc. so that logging can be completely compiled out?

Long term: I'd like to deprecate them in their current form, and use format strings that only get evaluated if the current logging level is enabled.

short term: yes, for usability/not breaking the ecosystem reasons, this is absolutely necessary. 

>     2. since spdlog has fmt based string formatting built in, e.g. `warn("variable set to {}",var)` - is the plan here to migrate away from boost format entirely in the logging

yes :) I don't have that yet, but think of replacing 

    GR_LOG_WARN(d_logger, boost::format("Could not put another bottle to %d on the wall") % n_bottles);

by 

    d_logger->warnf("Could not put another bottle to {:d} on the wall", n_bottles);

Note that amongst that plan is replacing the `std::shared_ptr<gr::logger>` that is `block::d_logger` with a unique ptr or simply a plain object: the loggers are thread-safe and not actually shared across blocks. Lifetime is reasonably identical to that of the owning block.

But in the (maint-3.9/master) current state of affairs, `d_logger` *is* a shared pointer, and I did not want to break all the things at once.

gnuradio-runtime/lib/logger.cc
```diff
@@ -1,416 +1,157 @@
...
 #include <gnuradio/prefs.h>
 
-#include <log4cpp/FileAppender.hh>
-#include <log4cpp/OstreamAppender.hh>
-#include <log4cpp/PatternLayout.hh>
-#include <log4cpp/PropertyConfigurator.hh>
-#include <log4cpp/RollingFileAppender.hh>
-#include <boost/thread.hpp>
+#include <boost/format.hpp>
+// #include <boost/thread.hpp>
```
mbr0wn, line 21:
```suggestion
```

gnuradio-runtime/lib/logger.cc
```diff
@@ -1,416 +1,157 @@
...
+#include <spdlog/spdlog.h>
 
-#include <algorithm>
-#include <filesystem>
-#include <iostream>
 #include <memory>
-#include <stdexcept>
 
 namespace gr {
+using std::make_shared;
```
mbr0wn, line 32:
```suggestion
```
...since you're actually not using :)

gnuradio-runtime/lib/logger.cc
```diff
@@ -1,416 +1,157 @@
...
-std::vector<std::string> logger_get_logger_names(void)
+void logger::set_level(const std::string& level)
 {
-    std::vector<std::string> names;
-    std::vector<log4cpp::Category*>* loggers = log4cpp::Category::getCurrentCategories();
-    std::vector<log4cpp::Category*>::iterator logger = loggers->begin();
-
-    for (; logger != loggers->end(); logger++) {
-        names.push_back((*logger)->getName());
+    auto l = spdlog::level::from_str(level);
```
mbr0wn, line 119: Can you please not do the single-character vars? Here, `l` is level, below, it's `default_logger` (or something like that).

gnuradio-runtime/lib/logger.cc
```diff
@@ -1,416 +1,157 @@
...
+    d_logger->set_pattern(fmt::format(logging::default_pattern, name));
 }
 
-void gr_logger_reset_config(void) { GR_RESET_CONFIGURATION(); }
-
-// Remaining capability provided by gr::logger class in gnuradio/logger.h
-
-namespace gr {
-
 bool configure_default_loggers(gr::logger_ptr& l,
```
mbr0wn, line 141: Not your code, I know, but can you please improve these arg names.

mbr0wn (commented):
First review-round.
Question: Do we now have access to fmtlib in GNU Radio then?

Also, were you able to do some benchmarks? I'm not even sure how exactly to benchmark this, but it's important to know.

mbr0wn:
This is pretty cool. And also, pretty pretty.

mbr0wn:
>     2. since spdlog has fmt based string formatting built in, e.g. `warn("variable set to {}",var)` - is the plan here to migrate away from boost format entirely in the logging

Ah Josh asked the same as I did.

mormj:
>  Question: Do we now have access to fmtlib in GNU Radio then?

Since spdlog requires fmtlib it is implicitly a requirement now.  Some distros (Ubuntu) package spdlog with fmtlib, others (Fedora) require it but package it externally.  These details will need to be worked out - but the end result should be that we can directly use fmtlib (and get rid of boost::format)

gr-fec/lib/generic_decoder.cc
```diff
@@ -8,21 +8,23 @@
  *
  */
 
+#include <memory>
```
mbr0wn, line 11: dupe

mbr0wn (changes requested):
Defo needs some squashes. I think we should do all the big spdlog-related changes in one commit. Also, rebase. All already in work, I know :)

marcusmueller:
@mbr0wn for tracking what we want to rebase this on:

1. #5011 
2. #5045

I think we'll want to modify #4935 to include all CI changes herein, too.

gr-fec/lib/generic_decoder.cc
```diff
@@ -8,21 +8,23 @@
...
 namespace gr {
 namespace fec {
 
 generic_decoder::generic_decoder(std::string name)
 {
     d_name = name;
     my_id = base_unique_id++;
-
-    GR_LOG_ASSIGN_CONFIGURED_LOGPTR(d_logger, alias());
+    d_logger = std::make_shared<gr::logger>(alias());
```
mbr0wn, line 27: The macro is silly, but a `gr::logger::make()` factory might be nice.

marcusmueller: ah to get a shared ptr right away? Didn't do that, because I wanted to move away from these, in the long term: a logger is pretty much a privately held object. Here, it makes little sense to restructure the whole thing just to make this happen right away, but plain objects or unique ptrs make more sense – the logging itself is thread-safe.

marcusmueller: hm this definitely sounds like something I should document as design rationale; else, I might as well add a factory, you're right, that would've been nicer.

gnuradio-runtime/lib/qa_logger.cc
```diff
@@ -13,27 +13,25 @@
...
  */
 
+#include <memory>
 #ifdef HAVE_CONFIG_H
 #include <config.h>
 #endif
 
 #include <gnuradio/logger.h>
 #include <boost/test/unit_test.hpp>
+#include <memory>
```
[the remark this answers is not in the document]
marcusmueller: removed

willcode:
When you get closer, could you move `.github/workflows/make-test.yml` to another PR?

marcusmueller:
> When you get closer, could you move `.github/workflows/make-test.yml` to another PR?

That's actually the plan, but the problem #5011 shows I'm sadly not there yet, @willcode :(

marcusmueller:
ah darn, I'll need to fix the strncmp warning disabling on master (and it has already been backported). and honestly, looking at it, the fix is kinda bad... instead of silencing the warning, I could probably addressed the issue.

willcode:
@marcusmueller We know how to unbackport :-) Whatever the best fix is for the long run.

mbr0wn:
The CI update commit needs to go elsewhere (#5077).

marcusmueller:
kicked out the CI and strncmp stuff, rebased, let's see whether this builds. Not feeling to great, will probably not be able to bring this into shape today.

marcusmueller:
squashed.

mbr0wn:
Cool! I still think we need to squash the cmake and the main "logging" commits.

mormj (approved):
I think this looks great - thanks for taking this one on, @marcusmueller.   Just a note to self that https://wiki.gnuradio.org/index.php/Logging will need to be updated prior to the release so the updated documentation gets scraped into the 3.10 release.  We probably will want to point to that document in a blog post or email blast to make clear to everyone how to use the new logging.  This all looks great though!

gnuradio-runtime/python/gnuradio/gr/bindings/logger_python.cc
```diff
@@ -14,152 +15,127 @@
...
 namespace py = pybind11;
 
 #include <gnuradio/logger.h>
+#include <logger_pydoc.h>
 
 void bind_logger(py::module& m)
 {
-    using logger = gr::logger;
-    using logger_config = gr::logger_config;
+    py::enum_<spdlog::level::level_enum>(m, "level_enum")
```
mormj, line 33: This is a very small nit, but the terminology "_enum" isn't used anywhere else.  Perhaps this should be "log_levels", so that in python it shows up as
mylogger.set_level(gr.log_levels.info)

mormj (changes requested):
Did some testing with OOT modules - the python interfaces to the logging work fine, both the default block loggers and a custom logger.  The only thing I found to be not working was in the conf file:
```
[LOG]
# Levels can be (case insensitive):
#       DEBUG, INFO, WARN, TRACE, ERROR, ALERT, CRIT, FATAL, EMERG
log_level = debug
debug_level = emerg

# These file names can either be 'stdout' to output to standard output
# or 'stderr' to output to standard error. Any other string will
# create a file in which to output the logging information. An empty
# string or no value here will ignore this level of configuration
# completely.
log_file = stdout
debug_file = stderr
```
The log and debug level settings work fine and discriminate accordingly.  If I put filenames here - everything goes to whatever `log_file` is set to and `debug_file` is never used  

The last part of the logging section there in gnuradio-runtime.conf.in can probably go as well
```
# Used for advanced configuration of the logger
#log_config = @CMAKE_INSTALL_PREFIX@/etc/gnuradio/gr_log_default.xml
```

mormj:
Let's go ahead and merge this so people can give it some airtime.  I think the minor things can be cleaned up in a separate PR.

PR #4982, opened 2021-08-05
runtime: logging performance improvement

willcode:
Based on email discussion with Boris Marjanovic.

Logging is being overhauled on master branch, so this is being introduced directly to maint-3.9.

The logging macros always pass the "msg" param to the logger (currently
log4cpp). In the macros, "msg" can be an expression and it will be
evaluated, even if the log message would not be printed (based on log
level).

This commit adds checks to each macro to check whether the message will be
logged. Since this check is also done internally to log4cpp, some
duplication results, but it should be far less effort than evaluating
an arbitrary "msg".

### DIFF
--- a/gnuradio-runtime/include/gnuradio/logger.h
+++ b/gnuradio-runtime/include/gnuradio/logger.h
@@ -156,136 +156,158 @@ typedef log4cpp::Category* logger_ptr;
 #define GR_RESET_CONFIGURATION() gr::logger_config::reset_config();
 
 /* Logger name referenced macros */
-#define GR_DEBUG(name, msg)                                           \
-    {                                                                 \
-        gr::logger_ptr logger = gr::logger_get_logger(name);          \
-        *logger << log4cpp::Priority::DEBUG << (msg) << log4cpp::eol; \
-    }
-
-#define GR_INFO(name, msg)                                           \
-    {                                                                \
-        gr::logger_ptr logger = gr::logger_get_logger(name);         \
-        *logger << log4cpp::Priority::INFO << (msg) << log4cpp::eol; \
-    }
-
-#define GR_NOTICE(name, msg)                                 \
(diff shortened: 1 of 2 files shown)


gnuradio-runtime/include/gnuradio/logger.h
```diff
@@ -156,136 +156,158 @@ typedef log4cpp::Category* logger_ptr;
...
+                *logger << log4cpp::Priority::ERROR << (msg) << log4cpp::eol; \
+        }                                                                     \
+    }
+
+#define GR_LOG_ASSERT(logger, cond, msg)                                      \
+    {                                                                         \
+        if (!(cond)) {                                                        \
+            if (logger->isPriorityEnabled(log4cpp::Priority::EMERG))          \
+                *logger << log4cpp::Priority::EMERG << (msg) << log4cpp::eol; \
+            assert(0);                                                        \
```
bmarjanovic, line 309: A possible improvement would be to add message to assert so the logged message is shown on command line as well. That would require additional macro. Something like 
#ifdef NDEBUG
# define assertm(exp, msg)  ((void)(0))
#else
# define assertm(exp, msg)                      \
    if (!exp) {                                              \
         std::cerr << "Assertion: " << msg;   \
         assert(exp);                                      \
     }                                  
#endif

bmarjanovic (approved):
The changes look good. The new macros have been working correctly for me.
A possible (separate from this) improvement to show assert message on command line (cerr). That would require additional macro - something like 
#ifdef NDEBUG
# define assertm(exp, msg)  ((void)(0))
#else
# define assertm(exp, msg)                \
  if (!exp) {                                          \
    std::cerr << "Assertion: " << msg;  \
    assert(exp);                                     \
  }                                  
#endif

mormj (approved):
LGTM

PR #4990, opened 2021-08-07
grc/cmake: Improve pyyaml check (forward port from 3.9)

mormj:
The check checked for one version and reported another.


willcode (approved):
Interestingly, 3.11 came out in 2014 and 3.10 in 2011, so we're not asking for cutting edge deps. But at least 3.9 and master are in sync on this!

Issue #4857, opened 2021-07-09
Clean up pyqtwidget(), sip wrappers


willcode:
@Yamakaja pointed out a possible small memory leak where various qt widgets return an opaque handle as a python object
```
#ifdef ENABLE_PYTHON
PyObject* number_sink_impl::pyqwidget()
{
    PyObject* w = PyLong_FromVoidPtr((void*)d_main_gui);
    PyObject* retarg = Py_BuildValue("N", w);
    return retarg;
}
#else
void* number_sink_impl::pyqwidget() { return NULL; }
#endif
```
which is then wrapped by PyBind11
```
        .def(
            "pyqwidget",
            [](std::shared_ptr<number_sink> p) {
                return PyLong_AsLongLong(p->pyqwidget());
            },
            D(number_sink, pyqwidget))
```
but does not deallocate the object returned by `pyqwidget()`. This brings up the question of why we are going through all this trouble to return a `long long`. The end user of this value is SIP, for instance in block template code
```
        ${win} = sip.wrapinstance(self.${id}.pyqwidget(), Qt.QWidget)
```
where SIP is wrapping the custom widgets for Python.

Can this be simplified?

Yamakaja:
So this is a bit of an awkward situation because we're passing a pointer from C++ to python using pybind11, and then passing that back to SIP, another python <-> native wrapper, though i'm not sure what you'd go about simplifying this. Could we maybe return the wrapper directly?

The obvious first step here would be to replace that leftover python stuff with either a `void *pyqwidget();` or `uintptr pyqwidget();`, where the latter has the added benefit of not requiring a pybind trampoline.

PR #4994, opened 2021-08-09
qtgui: Replace pyqwidget() with qwidget()

Yamakaja:
Previously, two versions of the `qtwidget` functions existed,
`qwidget()` and `pyqwidget()`, with the only difference being that
`qwidget()` returned a pointer to the `QWidget` object managed by the
corresponding block, while `pyqwidget()` returned that same pointer, but as
an integer (Or `PyLong` in this case).

While `qwidget()` is used by C++ code accessing the widgets,
`pyqwidget()` is only used for the python interface. This makes these
two methods redundant, thus this commit entirely removes `pyqwidget()`,
and modifies the `qwidget()` python wrapper to behave like
`pyqwidget()`. Note that we can be fairly confident that this change
will not effect potential users of `qwidget()`, because any invocation
on the objects previously returned by `qwidget()` would cause a
segmentation fault.

This commit also fixes a memory leak:

Internally, the `pyqwidget()` functions were returning a PyLong `PyObject *`,
which was then upwrapped in a pybind trampoline without decrementing
the reference count of that python object.

This PR partially addresses issue #4857.

### DIFF
--- a/docs/usage-manual/(exported from wiki) QT GUI.txt
+++ b/docs/usage-manual/(exported from wiki) QT GUI.txt
@@ -154,7 +154,7 @@ import sys, sip
                              samp_rate,      #bw
                              "QT GUI Plot")  #name
  
-     self.snk_win = sip.wrapinstance(self.snk.pyqwidget(), Qt.QWidget)
+     self.snk_win = sip.wrapinstance(self.snk.qwidget(), Qt.QWidget)
      self.snk_win.show()
  
  def main():
--- a/gnuradio-runtime/python/gnuradio/ctrlport/GrDataPlotter.py
+++ b/gnuradio-runtime/python/gnuradio/ctrlport/GrDataPlotter.py
@@ -75,7 +75,7 @@ class GrDataPlotParent(gr.top_block, Qt.QWidget):
             else:
                 self.connect(self.src[n], (self.snk,n))
 
-        self.py_window = sip.wrapinstance(self.snk.pyqwidget(), Qt.QWidget)
+        self.py_window = sip.wrapinstance(self.snk.qwidget(), Qt.QWidget)
 
         self.layout.addWidget(self.py_window)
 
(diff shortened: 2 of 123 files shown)


willcode:
Backport note: even though this looks like an API change, we cast back to a pointer in pybind11, and the only consumer appears to be `sip.wrapinstance()`, which takes a pointer, not a PyObject. If anyone thinks of a reason why the PyObject returned by the API would have been used directly in C++ API, let me know. Note that if `ENABLE_PYTHON` is `NO` then then `pyqwidget()` returned `void*` anyway, so we had an API that changed based on a config variable.

willcode:
@Yamakaja Does this build with `ENABLE_PYTHON=NO`?

Yamakaja:
> we cast back to a pointer in pybind11

I'm not sure i follow, where do we cast back to a pointer?

Yamakaja:
So yes, i just tried. Building with
```
    -DENABLE_PYTHON=NO \
    -DENABLE_GRC=NO \
    -DENABLE_GR_QTGUI=ON \
```
leads to a refreshingly fast and successful build

Yamakaja:
Updated commit description.

Note: Should compile, but the code hasn't been properly tested yet!

willcode:
Looks right, let's see what Josh has to say since he's the PyBind11 guru.

mormj:
These changes are all perfectly sensible.  I'm guessing the legacy of things was that swig required the additional interface, and now that we have pybind, it can just wrap the c++ interface without issue.  Definitely on board with letting pybind do the heavy lifting and not maintaining two interfaces, and if it works ...

mormj (approved):

mormj:
LGTM - thanks @Yamakaja

willcode (approved):

drmpeg (approved):
Tested on 32-bit armv7. Works good.

Yamakaja:
Rebased onto master

willcode:
@Sec42 reports on the osmocom_sdr mailing list that `pyqwidget()` was being used by gr-osmosdr's `osmocom_fft`, and provides a patch to that app. We were thinking that this was only used in-tree, but apparently not?

smunaut:
How would one add a Qt block to a layout in an app without using it ?

Maybe `osmocom_fft` is doing it wrong and there is another way, but AFAIK this is how GRC does it so we used that as example code on how to do it in our apps.

This means any external app using Qt widget would need updating, I'd call that a breaking API change ...

Issue #4998, opened 2021-08-09
grc: sc16 ports on uhd/soapy can not be connected to anything else due to type mismatch


mchlwntz:
In the past we could use a file source with format "short" and vector length of 2 to send complex int16 to the USRP. Now this produces an error that the source IO type doesn't match the sink IO type. It seems like this was introduced by this commit, which checks the types in addition to the sizes: https://github.com/gnuradio/gnuradio/commit/d44ee73deeaab86a7cc5820ef62688f291331b88#diff-07eb034ef7119f3142143792b6a91429ea645e3dd21cd48251d5332abf7486a2. Maybe there needs to be a special case for sc16 and the other hardware formats.

willcode:
The type checking got a little stricter. Previously anything with the same number of bytes per item would match. Unfortunately, there isn't a short-to-ishort block. Maybe there should be one.

Making a vector of 2 shorts match an ishort is a possibility.

You can still feed complex values to the USRP Sink and have it send 16-bit (or smaller) values over the wire, if you're mainly looking for a more efficient OTW format.

willcode:
There's a missing alias in `grc/core/Constants.py`. I'll put in a fix.

dl1ksv:
@willcode I have only limited means to test #4999, but as far as I could test: it works

willcode:
Worked for C2I -> I2C, but not for USRP -> I2C
`ValueError: itemsize mismatch: usrp_source0:0 using 4, interleaved_short_to_complex0:0 using 2`

ryanvolz:
Hey, it's one of those times when you go to file a bug and find that some else recently did it!

This also kinda/sorta applies to the interleaved char blocks; I have an OOT block than uses an "sc8" type which previously worked with the output of the complex->ichar block, but now won't. I don't know if any other blocks out there use "sc8", but it would sure be nice if similar changes from #5001 could also be applied to the interleaved char blocks.

willcode:
It's one of those times where you were wondering if the other type option needed the same kind of change and ...

OK, I'll check it out.

willcode:
@mchlwntz Does the fix we're talking about work for you? The only downside of the I2C/C2I and I2B/B2I adapter blocks is that they do an unnecessary memcpy (to be addressed some day).

mchlwntz:
[...]

PR #4999, opened 2021-08-11
grc: introduce sc16 format to interleaved_short blocks

dl1ksv:
At the moment there exist no block to convert comlex to sc16 and vice versa,
but only from complex to ishort.
But you can interpret an ishort as sc16. So adding sc16 as possible input/output type
should solve this issue.

Fixes: #4998

### DIFF
--- a/gr-blocks/grc/blocks_complex_to_interleaved_short.block.yml
+++ b/gr-blocks/grc/blocks_complex_to_interleaved_short.block.yml
@@ -15,6 +15,11 @@ parameters:
     option_labels: ['No', 'Yes']
     option_attributes:
         vlen: [1, 2]
+-   id: output_type
+    label: Output type
+    dtype: enum
+    default: 'short'
+    options: [short, sc16]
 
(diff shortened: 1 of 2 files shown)


willcode (approved):
Tested, works.

willcode (commented):
It worked for C2I -> I2C, but it did not work for a USRP. USRP sc16 (and I assume Soapy sc16) is a 4-byte type and C2I/I2C use 2-byte types. Changing the port type in GRC is not sufficient.

dl1ksv:
Setting input to sc16 of i2c gives me an input itemsize of 4 as requested.

willcode:
@dl1ksv Take a look at #5001. I think this works. The type selector and vector selector are a bit redundant as far as the generated code is concerned, but are separate things in GRC. I didn't come up with a good way to combine them. Maybe a 3-way selection of short, short[2], ishort?

willcode:
Superceded by #5001

PR #5001, opened 2021-08-11
gr-blocks: add sc16 type to ishort->complex and complex->ishort

willcode:
Based on PR #4999 Volker Schroer

Addresses #4998, which notes that a vector of short[2] previously match sc16. Since stricter typechecking was added, this is no longer the cases. This PR allows the I2C/C2I blocks can now be used as adapters.

### DIFF
--- a/gr-blocks/grc/blocks_complex_to_interleaved_short.block.yml
+++ b/gr-blocks/grc/blocks_complex_to_interleaved_short.block.yml
@@ -7,6 +7,11 @@ parameters:
     label: Scale Factor
     dtype: float
     default: '1.0'
+-   id: output_type
+    label: Output type
+    dtype: enum
+    default: 'short'
+    options: [short, sc16]
 -   id: vector_output
     label: Vector Output
     dtype: enum
@@ -15,6 +20,7 @@ parameters:
     option_labels: ['No', 'Yes']
     option_attributes:
         vlen: [1, 2]
+    hide: ${ 'all' if output_type == 'sc16' else 'none' }
 
 inputs:
 -   domain: stream
@@ -22,19 +28,19 @@ inputs:
 
 outputs:
 -   domain: stream
-    dtype: short
-    vlen: ${ vector_output.vlen }
+    dtype: ${output_type}
+    vlen: ${ 1 if output_type == 'sc16' else vector_output.vlen }
 
 templates:
     imports: from gnuradio import blocks
-    make: blocks.complex_to_interleaved_short(${vector_output},${scale_factor})
+    make: blocks.complex_to_interleaved_short(${output_type == 'sc16' or vector_output},${scale_factor})
     callbacks:
     - set_scale_factor(${scale_factor})
 
 cpp_templates:
     includes: ['#include <gnuradio/blocks/complex_to_interleaved_short.h>']
     declarations: 'blocks::complex_to_interleaved_short::sptr ${id};'
-    make: 'this->${id} = blocks::complex_to_interleaved_short::make(${vector_output});'
+    make: 'this->${id} = blocks::complex_to_interleaved_short::make(${output_type == "sc16" or vector_output});'
     callbacks:
     - set_scale_factor(${scale_factor})
     translations:
--- a/gr-blocks/grc/blocks_complex_to_interleaved_char.block.yml
+++ b/gr-blocks/grc/blocks_complex_to_interleaved_char.block.yml
@@ -7,6 +7,11 @@ parameters:
     label: Scale Factor
     dtype: float
     default: '1.0'
+-   id: output_type
+    label: Output type
+    dtype: enum
+    default: 'byte'
+    options: [byte, sc8]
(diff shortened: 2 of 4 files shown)


willcode:
The C++ template code has not been tested yet.

gr-blocks/grc/blocks_complex_to_interleaved_short.block.yml
```diff
@@ -22,19 +27,19 @@ inputs:
...
-    make: blocks.complex_to_interleaved_short(${vector_output},${scale_factor})
+    make: blocks.complex_to_interleaved_short(${output_type == 'sc16' or vector_output},${scale_factor})
     callbacks:
     - set_scale_factor(${scale_factor})
 
 cpp_templates:
     includes: ['#include <gnuradio/blocks/complex_to_interleaved_short.h>']
     declarations: 'blocks::complex_to_interleaved_short::sptr ${id};'
-    make: 'this->${id} = blocks::complex_to_interleaved_short::make(${vector_output});'
+    make: 'this->${id} = blocks::complex_to_interleaved_short::make(${output_type == "sc16" || vector_output});'
```
dl1ksv, line 42: This leads to an syntax error.

Replace || by or and the cpp code will be generated.

dl1ksv:
I tried to test with UHD: USRP sink. Flowgraph generation works. Python executable generation works , too.But as I have no usrp I can't test, if the generated  code works correctly.
Linking the cpp code fails, as the link entry in uhd_usrp_sink.block.yml is wrong.
It should be gnuradio::gnuradio-uhd. After this change the cpp code can be linked.

willcode:
Tested USRP Source -> I2C -> C2I -> USRP Sink in Python and C++. Works.

I see the gnuradio-uhd vs gnuradio::gnuradio-uhd, but it did link and run without a change. So, we'll leave that to another PR.

willcode:
@ryanvolz Could you test out the sc8 part of this? I copied the sc16 changes with s/sc16/sc8 and s/short/byte pretty much.

willcode:
Need opinions on a better way to select the type/vectorness. Chosing sc8/sc16 and vector does not make sense, so that case is silently ignored and a non-vector sc8/sc16 is assumed.

The underlying C++ blocks just take a bool the doubles the type size, so they don't care about sc8/16 vs byte[2]/float[2].

GRC type checking improvements were really needed. Previously, a byte[4] could be connected to a sc16 with no warning/error. The additional type checking causes some backward compatibility glitches like this one. In my opinion, the change is worth it, since there are just these few corner cases. Of course, we have no idea how many people were counting on the lack of type checking.

ryanvolz:
Both the sc16 and sc8 work for me!

You can hide the "Vector Input/Output" option and set it to False when one of the "sc" types is selected, no?

willcode:
Thanks for testing.

That's the option I tried first. The "hide" part worked fine, but I don't know how to make an enum set the value of another param. That was before it was apparent that some logic was needed in the templates, so sure, a very minor tweak would make that work.

mormj:
@willcode - I'm hesitant on these types of changes.  I understand that the underlying issue prevents convenient connections such as from a file source vector output to a USRP sink, but it would be nice if these conversions could be handled in a central way in GRC.  Since gnuradio only does connection type checking based on size, this is strictly a GRC issue.  

Honestly, i'd rather roll back #4483 than peppering the grc files with if statements

willcode:
Agree - I left it as draft for discussion. If GRC understood that `short[2] == sc16`, `char[2] == ichar`, etc., that would be much better.

ryanvolz:
Since @willcode threatened to close this draft PR in the project talk, and I have an interest in seeing #4998 fixed...

> Agree - I left it as draft for discussion. If GRC understood that `short[2] == sc16`, `char[2] == ichar`, etc., that would be much better.

How about adding `short == sc16` and `byte == sc8` as additional aliases in `grc/core/Constants.py`? The intent would be to get the dtype check to pass, and then the size check would ensure that it only works when you connect a vector of 2 shorts to an sc16 port and vice-versa (ditto for byte and sc8).

willcode:
@ryanvolz Did you try that out? We could do that as a stopgap if it works.

ryanvolz:
Yes, I just tried this:
```
--- a/grc/core/Constants.py
+++ b/grc/core/Constants.py
@@ -121,14 +121,18 @@ ALIAS_OF = {
     'float': 'f32',
     'int': 's32',
     'short': 's16',
+    'short': 'sc16',
     'byte': 's8',
+    'byte': 'sc8',
     'bits': 'bit',
 
     'fc32': 'complex',
     'f32': 'float',
     's32': 'int',
     's16': 'short',
+    'sc16': 'short',
     's8': 'byte',
+    'sc8': 'byte',
     'bit': 'bits',
 }
```
and it works. I can make a quick PR if that's the direction we want to go.

willcode:
Let's give that a try. It will help until we find a "right" solution. I'll close this one.

PR #5004, opened 2021-08-14
GRC: View-Only Mode

oscekh:
This PR adds the view-only mode to the GRC, including the core-features from https://github.com/gnuradio/gnuradio/pull/4838 as well as the GUI parts of the mode.

This feature saves expression's evaluated values along with the expressions themselves in the grc-file. If the view-only mode is enabled (toggled in `File>Enable View-Only Mode`) untrusted flow graphs are opened using the saved expressions, and with evaluation disabled, to prevent arbitrary code execution attacks. 

When the user performs some non-view only action (defined in [`gui/Application.py`](https://github.com/gnuradio/gnuradio/commit/55ea8837694cae15e5682f8a3a62c64a6f2c6ba9#diff-a720b612a9bb32b2def305f3e5ffa79d7030b8b9474eab3dd71d23d2b4832dbeR124)) or if opening a flow-graph missing the saved evaluated values the trust prompt is displayed to the user. In the trust prompt the user may choose to trust the flow graph (once or indefinitely), or cancel the prompt.

In addition to the prompt there is also a trust manager dialog where the user can add or remove indefinitely trusted flow graphs, this acts as an interface to the user's config where the paths of the flow graphs are stored. This also supports trusting directories, meaning that any flow graphs inside that directory are also trusted.

In the prompt there is an expression overview which gathers the expressions that would be evaluated, including the expressions originating from the flow graph (parameters) and from the block definitions. The overview should ideally filter out any expression that is obviously benign. Though, currently it only filters out literals using `ast.literal_eval`.

I'd love to get some feedback on:
- Language: are the prompt messages clear?
- UX: are there parts of the feature that are confusing, annoying? Would be great to get some opinions from people less involved with the feature as well, to find issues I wouldn't spot.
- Code
- anything else

For further information on the View-Only mode itself and my GSoC work I refer to [my mail on the discuss mailing list.](https://lists.gnu.org/archive/html/discuss-gnuradio/2021-05/msg00122.html)


willcode:
Thanks for the PR. Giving this a quick try and documented my experience to get a discussion rolling.

The expression viewer turned out pretty good. I like that you can sort by ORIGIN so you can ignore all the block definitions if you want. Overall, the changes appear to work. Nothing blew up and the PR appears to accomplish its goal.

I would have expected the trust dialog to appear right when a file is opened. Instead, it came up when I opened a block and typed the first character into a param box. This would be when the eval is hit, since the flowgraph had cached evaluated data already. This isn't wrong, and maybe it's even good since you can see the overall flowgraph before having to decide whether you trust it. Just unexpected from a UX point of view. A flowgraph that included an embedded Python block triggered the trust dialog when opened.

The appearance of variables in the expression overview makes it harder to find things that might actually be a problem. Are there some things that can be assumed to be safe? Python makes this hard, since access to a field in an object can involve running user code, but is a variable name a safe expression? Or could they at least be sorted by ORIGIN in a way that you see all the complex expressions group together?

If something is an id, is it safe? For example, the list shows "samp_rate" as an expression and the Name field is "samp_rate.Id".

Some strings should always be safe. I don't know how to determine which ones are.

I set a directory as trusted and opened a flowgraph from that directory. I got the trust dialog. It was only after I changed something (even moving a block a little) and saved that the block was trusted on the next open. So, I think blocks are not trusted unless they contain evaluated expressions. It should be possible to trust a block that does not yet contain evaluated expressions. Then, it's a separate question whether we want to save evaluations in this case. I'd argue that a trusted flowgraph that doesn't contain evaluations should stay that way.

Various error messages are showing up in the console, some to do with the GUI changes. I can't say for sure that these problems with the PR, so I'll test more later.

oscekh:
Thanks for the quick response!

I agree that more should/could be filtered out of the overview. It is on my todo-list and I think filtering out the strings/identifiers shouldn't be too difficult.

Regarding when to open the prompt I don't know which is preferable, I thought that it might get annoying to have it pop up immediately, but I agree that it is sort of unexpected that it doesn't.

I think the reason you got the prompt even after trusting the dir was that I had forgotten to make the trust manager update the internal trust-list (attribute of platform), it just updates the config. So the trustlist wasn't updated until the config was reloaded (next open). I've fixed that now, so the manager updates both the config as well as the platform's trust-list.

As for the one test failing on ubuntu (qa_filterbank) I'm stumped, maybe someone else knows what might be the issue?

willcode:
qa_filterbank has spurious failures. Not related.

grc/compiler.py
```diff
@@ -56,6 +58,9 @@ def main(args=None):
         os.path.exists(grc_file) or exit('Error: missing ' + grc_file)
         Messages.send('\n')
 
+        if args.trusted:
+            platform.trusted_flowgraphs.add(str(grc_file))
```
skoslowski, line 62: I suspect this fails for flowgraphs with dependent hier_block in them. These can get loaded and generated as well, but they won't be trusted by default. I would change this to simply trust all flowgraphs.

grc/core/FlowGraph.py
```diff
@@ -286,10 +295,13 @@ def evaluate(self, expr, namespace=None, local_namespace=None):
         # Evaluate
         if not expr:
             raise Exception('Cannot evaluate empty statement.')
-        if namespace is not None:
-            return eval(expr, namespace, local_namespace)
+        if not self.view_only:
```
skoslowski, line 298: Early exit?
```suggestion
        if self.view_only:
            return
```

grc/core/FlowGraph.py
```diff
@@ -462,10 +474,12 @@ def verify_and_get_port(key, block, dir):
...
-            Messages.send_error_load(
-                'Connection between {}({}) and {}({}) could not be made.\n\t{}'.format(
-                    src_blk_id, src_port_id, snk_blk_id, snk_port_id, e))
-            had_connect_errors = True
+            # hide connection errors caused by missing view-only values
+            if self.view_only and not self.trigger_trust_prompt:
+                Messages.send_error_load(
+                    'Connection between {}({}) and {}({}) could not be made.\n\t{}'.format(
+                        src_blk_id, src_port_id, snk_blk_id, snk_port_id, e))
+                had_connect_errors = True
```
skoslowski, line 482: Is this a left-over from the "best-effort" loading experiments?

grc/core/blocks/block.py
```diff
@@ -133,16 +137,17 @@ def rekey(ports):
...
-            imports = self.templates.render('imports')
-            exec(imports, self.block_namespace)
-        except ImportError:
-            # We do not have a good way right now to determine if an import is for a
-            # hier block, these imports will fail as they are not in the search path
-            # this is ok behavior, unfortunately we could be hiding other import bugs
-            pass
-        except Exception:
-            self.add_error_message(f'Failed to evaluate import expression {imports!r}')
+        if not self.parent_flowgraph.view_only:
```
skoslowski, line 140: Switch to early return?

grc/core/blocks/block.py
```diff
@@ -667,6 +703,11 @@ def get_hash():
             for key, value in parameters.items():
                 try:
                     self.params[key].set_value(value)
+                    if evaluated:
+                        try:
+                            self.params[key]._saved_evaluated = evaluated[key]
+                        except KeyError:
+                            self.params[key]._using_saved = False
```
skoslowski, line 710: This attributes of `Param` should not have a leading `_` if they're accessed from outside the class

grc/core/platform.py
```diff
@@ -343,9 +361,11 @@ def parse_flow_graph(self, filename):
 
         return data
 
-    def save_flow_graph(self, filename, flow_graph):
+    def save_flow_graph(self, filename, flow_graph, data=None):
```
skoslowski, line 364: That `data` argument is not needed if I recall our last talk correctly

grc/core/platform.py
```diff
@@ -84,6 +100,8 @@ def load_and_generate_flow_graph(self, file_path, out_path=None, hier_only=False
         try:
             flow_graph = self.make_flow_graph()
             flow_graph.grc_file_path = file_path
+            if self.is_trusted(file_path):
+                flow_graph.view_only = False
```
skoslowski, line 104: The `viewl_only` attribute seems to be set twice. Here and in `self.make_flow_graph()`. I suspect this one can be removed

grc/gui/Constants.py
```diff
@@ -104,6 +104,15 @@
 GTK_INI_PREFER_DARK_KEY = 'gtk-application-prefer-dark-theme'
 GTK_INI_THEME_NAME_KEY = 'gtk-theme-name'
 
+TRUST_PROMPT_ACTION_MESSAGE = """\
+This action cannot be performed while in View-Only Mode. \
+To perform the action you will need to trust this flow graph first.\
+\n\nOnly trust the flow graph if you are sure its expressions don\'t contain malicious code!"""
+
+TRUST_PROMPT_MISSING_VALUES = """\
+This flow graph file is missing the values needed to open it in view only mode. \
```
skoslowski, line 113:
```suggestion
This flow graph file is missing the values needed to open it in view only-mode. \
```

grc/gui/Dialogs.py
```diff
@@ -452,3 +454,288 @@ def choose_editor(parent, config):
...
+        self.expander.add(self.overview_box)
+        self.expander.connect('activate', self.show_overview)
+
+        self.vbox.pack_start(message_label, False, True, 10)
+        self.vbox.pack_start(wiki_link_label, False, True, 10)
+        self.vbox.pack_start(self.expander, False, True, 10)
+
+        self.show_all()
+
+    def is_literal(self, value):
```
skoslowski, line 557: This can be a free function outside of the class

grc/gui/Dialogs.py
```diff
@@ -452,3 +454,288 @@ def choose_editor(parent, config):
...
+        blocks = []
+
+        for block in self.flowgraph.blocks:
+            imps = block.templates.render('imports')
+            if imps:
+                imports += [(imps, block.name)]
+
+            if block._eval_cache:
+                for expr in block._eval_cache.keys():
+                    if expr and not self.is_literal(expr):
```
skoslowski, line 581:
```suggestion
                    if expr and not (expr.isidentifier() or self.is_literal(expr)):
```

grc/gui/Dialogs.py
```diff
@@ -452,3 +454,288 @@ def choose_editor(parent, config):
...
+                        blocks += [(expr, block.name)]
+
+            for _, param in block.params.items():
```
skoslowski, line 584:
```suggestion
            for param in block.params.values():
```

grc/gui/Dialogs.py
```diff
@@ -452,3 +454,288 @@ def choose_editor(parent, config):
...
+                value = param.to_code()
+                if value and not self.is_literal(value):
```
skoslowski, line 586:
```suggestion
                if value and not (value.isidentifier() or self.is_literal(value)):
```

grc/gui/MainWindow.py
```diff
@@ -246,12 +246,29 @@ def new_page(self, file_path='', show=False):
             if file_path: Messages.send_start_load(file_path)
             flow_graph = self._platform.make_flow_graph()
             flow_graph.grc_file_path = file_path
-            #print flow_graph
+
+            if not file_path or self._platform.is_trusted(file_path):
+                flow_graph.view_only = False
```
skoslowski, line 251: Shouldn't this be already handled in `make_flow_graph()`?

grc/main.py
```diff
@@ -28,8 +28,27 @@
...
+    if event == 'exec':
+        stack = inspect.stack()
+        path = stack[1].filename
+        #print(*stack, sep="\n")
+
+        #if "/grc/core" in path:
+        if "/grc/" in path:
+            print("IN GRC")
+            print(*stack, sep="\n")
+
```
skoslowski, line 47: Maybe remove this, now that is in the tests?

skoslowski (changes requested):
Hi @oscekh, great job on getting a lot of the last-minute stuff into this PR!
This is my code-only review with a lot of (minor) suggestions.
I'll give separate feedback on how the runtime behavior.

grc/core/FlowGraph.py
oscekh: Yeah, good find!

grc/gui/MainWindow.py
oscekh: MainWindow doesn't pass a file path to `make_flow_graph` which means `is_trusted` will fail in `make_flow_graph` but succeed in MainWindow (if trusted). So `make_flow_graph` has the check for the times a path is provided directly, but in the case of MainWindow it is checked outside (as you wouldn't want to pass the path since it would parse+import as well).

```python
def make_flow_graph(self, from_filename=None):
    fg = self.FlowGraph(parent=self)
    if not self.view_only_mode or self.is_trusted(from_filename):
        fg.view_only = False
    if from_filename:
        data = self.parse_flow_graph(from_filename)
        fg.grc_file_path = from_filename
        fg.import_data(data)
    return fg
```
Though, there may be a more elegant solution. I'll try to think of one!

grc/main.py
oscekh: yeah, somehow missed that it was still there

777arc:
To enable this mode, do you have to check the option under File and then restart GRC?  I think it didn't start working for me until I did that.  If that's the intended behavior there should probably be a popup box that warns the user they have to restart GRC when the box goes from unchecked to checked.  Overall nothing appears broken and the feature seems to work great once I restarted GRC!  You are setting a high bar for future GSoC students =)

grc/gui/Dialogs.py
```diff
@@ -452,3 +454,288 @@ def choose_editor(parent, config):
...
+        self.add_buttons(*buttons)
+        self.dialog_height = 300
+        self.set_size_request(Constants.MIN_DIALOG_WIDTH, self.dialog_height)
+        self.set_border_width(10)
+        self.counter = 0
+
+        # show first so Gtk sets dialog width and wraps text accordingly
+        self.show_all()
+
+        message_label = Gtk.Label(message)
```
haakov, line 512: I'm getting this warning:
```
/usr/local/lib/python3.9/site-packages/gnuradio/grc/gui/Dialogs.py:512: PyGTKDeprecationWarning: Using positional arguments with the GObject constructor has been deprecated. Please specify keyword(s) for "label" or use a class specific constructor. See: https://wiki.gnome.org/PyGObject/InitializerDeprecations
  message_label = Gtk.Label(message)
```

haakov:
This is _really_ awesome! 

I've got a tiny question about the trust dialog. Would it make more sense to call the Cancel button "Don't trust" instead of "Cancel"? I feel like if you're opening a file and the trust dialog opens up at once, you might get the idea that the Cancel button will cancel the entire "Open file" operation altogether?

grc/gui/Dialogs.py
oscekh: Hmm, did not see that, thanks for pointing it out! I've made it a keyword arg so it should be fixed now.

oscekh:
> To enable this mode, do you have to check the option under File and then restart GRC? I think it didn't start working for me until I did that. If that's the intended behavior there should probably be a popup box that warns the user they have to restart GRC when the box goes from unchecked to checked. 

This should be fixed now!

> I've got a tiny question about the trust dialog. Would it make more sense to call the Cancel button "Don't trust" instead of "Cancel"? I feel like if you're opening a file and the trust dialog opens up at once, you might get the idea that the Cancel button will cancel the entire "Open file" operation altogether?

Yeah, I could see how it would be more clear. Though, it typically works as a cancel, cancelling the operation that triggered it. Such as when one tries to add a block in view only and press cancel in the prompt, the block will not be placed. And when opening a flow graph missing the values needed for view only, it will cancel the opening unless trusted. But still, naming it "Don't Trust" might be more clear.

What are other people's opinion on this?

mormj:
@oscekh - thank you so much for cleaning up and submitting here.  This appears to be a very productive GSoC project with a valuable feature, and your contributions are very much appreciated.  Here are my observations upon trying out the feature:
1) In the manage trust dialog - the "cancel" button should probably be "close" since it doesn't revert the state
2) When opening a flowgraph that is not in the trusted directory, there doesn't seem to be a way to open it up in "view only" without trusting the flowgraph.  This might have been beyond the scope of the project, but it would be nice if there was an option to show the flowgraph in some read only mode without evaluating the expressions - can't be rendered, but shown in some hobbled form.  I think some wiki documentation would go a long way towards the understanding of how the feature is expected to operate.
3) the text: "This flowgraph is missing the values needed to open it in View-Only mode" is not entirely clear. It seems the first 3 sentences are in reverse order and might make more sense as:  "this flowgraph contains python expressions that are evaluated in GRC and cannot be opened in View Only mode.  To open this flowgraph, ... Only trust the flowgraph ... "
4) As @willcode pointed out, a lot of the expressions that show up in the list seem innocuous - some are just strings, so it would be nice to pare down this list if possible.  Perhaps in the documentation it should explain why things could be malicious.

Overall a great feature to have in GRC, so thank you for the hard work this summer.

oscekh:
Thank you for the review, and apologies for the delayed response!

>     1. In the manage trust dialog - the "cancel" button should probably be "close" since it doesn't revert the state
>

I agree. Good point!

>     2. When opening a flowgraph that is not in the trusted directory, there doesn't seem to be a way to open it up in "view only" without trusting the flowgraph.  This might have been beyond the scope of the project, but it would be nice if there was an option to show the flowgraph in some read only mode without evaluating the expressions - can't be rendered, but shown in some hobbled form.  I think some wiki documentation would go a long way towards the understanding of how the feature is expected to operate.

Do you mean when opening a flow graph which misses the 'view only values' in the grc-file? Yeah, that was something we discussed during the project, but ultimately decided that just rendering it naively would too often lead to confusing flowgraphs which do not really represent the actual flowgraph. There may be a nice middle ground where incomplete flow-graphs may be displayed depending on what it's missing. For example lack of param values may show param expressions on surface level instead, whereas if it misses more 'serious' expressions such as dynamic number of inputs it may not be rendered. Though, this is beyond the scope for this project.

Any non-broken grc-file (saved after introduction of view-only mode) should be able to open in view only though, in the current state.

>     3. the text: "This flowgraph is missing the values needed to open it in View-Only mode" is not entirely clear. It seems the first 3 sentences are in reverse order and might make more sense as:  "this flowgraph contains python expressions that are evaluated in GRC and cannot be opened in View Only mode.  To open this flowgraph, ... Only trust the flowgraph ... "

Yeah that is an interesting point, I think I did it that way to clearly distinguish it from the main trust prompt message (activated when adding a block for example).

>     4. As @willcode pointed out, a lot of the expressions that show up in the list seem innocuous - some are just strings, so it would be nice to pare down this list if possible.  Perhaps in the documentation it should explain why things could be malicious.

Yes, I added so it filters out literals in the push after will's comment, and since then I've also added filtering of identifiers (though, I haven't pushed it yet). This seems to filter out pretty much all the innocuous expressions.

mormj:
Closing this PR temporarily, but adding to the TODO for 3.11 release

Issue #5006, opened 2021-08-15
xmlrpc_client block doesn't accept address/callback parameters 


muaddib1984:
xmlrpc client block has variables for address and callback that were changed to dtype raw [here](https://github.com/gnuradio/gnuradio/pull/3554#issue-430591489) which worked in 3.8, now in 3.9 the raw dtype doesn't seem to be evalutated correctly. 
![xmlrpc_block_toplevel](https://user-images.githubusercontent.com/70996692/129451426-dcd1bda7-a68f-4a26-83e4-32b1b0137465.png)
![xmlrpc_client_variables](https://user-images.githubusercontent.com/70996692/129451436-a924a277-7749-4770-905f-65622804cd0f.png)

if the variables are quoted as strings in the block, the flowgraph can be generated, but the parameters as strings breaks the xmlrpc behavior in the generated flowgraph shown here:
`self.xmlrpc_client_0 = ServerProxy('http://"localhost":8080')`
`...`
`self.xmlrpc_client_0."set_freq"(self.freq)

the desired output would be

`self.xmlrpc_client_0 = ServerProxy('http://localhost:8001')`
`...`
`self.xmlrpc_client_0.set_freq(self.freq)`

i'm submitting a pull request for what i think is the best temporary solution, but it requires that the callback and local variable share the same name.

muaddib1984:
Yes, we fixed this. Sorry for the dangling open issue.<end transmission>On Jan 29, 2024, at 07:47, Marcus Müller ***@***.***> wrote:﻿
https://gist.github.com/marcusmueller/f2b343e1e23e8fc903708c75d2df766d
repro5006.svg (view on web)
can't reproduce, seems to be fixed through #5008

marcusmueller:
no worries! Was so happy to a) see it fixed and b) knowing it was your first PR :)

PR #5007, opened 2021-08-15
adding xmlrpc examples and fix for maint-3.9

muaddib1984:
addressing my own issue #5006 
Attempting to fix 2 grc parameter issues with the xmlrpc_client block. Specifically, the address and callback raw formats aren't accepted by the block. The values pass the declaration test by the block if they are quoted as strings, but after that they don't translate properly to the generated python flowgraph. Similar to issues @argilo [fixed](https://github.com/gnuradio/gnuradio/pull/3554#issue-430591489) in maint-3.8. I'm aware that this changes the address **back** to  string from raw, but it seems to work fine now that we're completely in python3. A new condition for my fix is that the callback parameter and the variable both need to share the same name and 'set_' is hardcoded into the block's yaml.

I also added an xmlrpc example directory with  GRC flowgraphs for server and client.


willcode:
@muaddib1984 Thanks for working on this. Could you submit this PR against the "master" branch. We do development work there, and the backport to the maintenance branches as appropriate. I think you'll have to resubmit to do this.

The `address` is definitely broken currently. Something seems weird about the way `callback` and `variable` are used. Maybe that needs some thought in general.

muaddib1984:
hey, no problem will do. we can open the discussion about the block there too.

muaddib1984:
i made a new pull request against master, #5008 let me know if that's correct.

PR #5008, opened 2021-08-15
fixing xmlrpc block params

muaddib1984:
comments from original pull request #5007, re-pull against master:
addressing my own issue #5006 
Attempting to fix 2 grc parameter issues with the xmlrpc_client block. Specifically, the address and callback raw formats aren't accepted by the block. The values pass the declaration test by the block if they are quoted as strings, but after that they don't translate properly to the generated python flowgraph. Similar to issues @argilo [fixed](https://github.com/gnuradio/gnuradio/pull/3554#issue-430591489) in maint-3.8. I'm aware that this changes the address **back** to  string from raw, but it seems to work fine now that we're completely in python3. A new condition for my fix is that the callback parameter and the variable both need to share the same name and 'set_' is hardcoded into the block's yaml.

I also added an xmlrpc example directory with  GRC flowgraphs for server and client.


willcode:
Thanks for working on this and moving it to "master".

The `addr` fix looks correct and necessary. The rest could use some discussion.

muaddib1984:
I agree. Where is the best place to have that discussion? I’m a long time user, but a brand new contributor.

<end transmission>

willcode:
@muaddib1984 This is the right place for the discussion. It may be slow this time of year.

muaddib1984:
sounds good. i would really like to know how why the block doesn't accept 'set_<varname>' in the callback field of the block. i can tell that it doesn't like an undeclared variable. 
Thanks for putting this in the queue

gr-blocks/grc/xmlrpc_client.block.yml
```diff
@@ -14,18 +14,18 @@ parameters:
...
     label: Variable
     dtype: raw
 
 templates:
     imports: from xmlrpc.client import ServerProxy
-    make: ServerProxy('http://${addr}:${port}')
+    make: ServerProxy('http://'+${addr}+':${port}')
     callbacks:
-    - ${callback}(${variable})
+    - set_${(callback)}(${variable})
```
haakov, line 26: The parentheses here don't do anything, right?

gr-blocks/grc/xmlrpc_client.block.yml
```diff
@@ -14,18 +14,18 @@ parameters:
...
 
 documentation: |-
-    This block will create an XMLRPC client. The client will execute the callback on the server when the variable is changed. The callback should be a the name of a function registered on the server. The variable should be an expression containing a the name of a variable in flow graph.
+    This block will create an XMLRPC client. The client will execute the callback on the server when the variable is changed. The callback and variable should be the name of a function registered on the server. Note that this executes a 'set' command on the server function and the 'set_<variable name>' syntax is already handled in the block.
```
haakov, line 29: Well, the variable does not have to be the name of a server-side function, it only needs to correspond to a client-side variable.

haakov (changes requested):
Great work! I'll come back with a few notes shortly

haakov:
1. Personally, I don't think ZMQ is needed in the examples. I suggest to remove the ZMQ blocks, move the QT sinks to the server-side flowgraph, and only keep the QT Slider in the client-side flowgraph.

gr-blocks/grc/xmlrpc_client.block.yml
muaddib1984: That’s correct. I left them there in case there was a reason I didn’t know about.

gr-blocks/grc/xmlrpc_client.block.yml
```diff
@@ -14,18 +14,18 @@ parameters:
 -   id: callback
     label: Callback
     dtype: raw
```
haakov, line 16: The Callback's `dtype` can be a string

gr-blocks/grc/xmlrpc_client.block.yml
muaddib1984: So this is the secondary problem. 
you have to use a declared variable. If you try do use a variable name that doesn’t exist in the local flowgraph, it fails at init saying that <Remote server variable name> is an undeclared variable.

gr-blocks/grc/xmlrpc_client.block.yml
haakov: Or it can continue being a raw, but that would typically require quotes

muaddib1984: If you make a callback‘s dtype a string it will put it to the  generated python code with quotes around The actual function call which throws an error (rightfully so).

muaddib1984: I thought raw would indiscriminately pass along whatever text you gave it. Could this possibly be a problem with the Python code generation for raw type? Odd that it used to work fine.

muaddib1984: `-   id: callback`
    `label: Callback`
    `dtype: raw`
    `default: '0'` 
if you put quotes around the variable it returns this 
`self.xmlrpc_client_0."set_rmt_freq"(self.freq)`
 ` ^`
`SyntaxError: invalid syntax`

muaddib1984: -   id: callback
    label: Callback
    dtype: string
    default: '0'
    results in the same error
    `self.xmlrpc_client_0.set_'rmt_freq'(self.freq)`
`                               ^`
`SyntaxError: invalid syntax`

muaddib1984 (commented):
changing the examples to remove zmq is fine,but i think it showcases a completely headless solution really well. if you want to move everything but the one slider back into the server flowgraph i can do that when we figure out the callback variable formatting issue.

gr-blocks/grc/xmlrpc_client.block.yml
haakov: Yes, the quotes can be removed by taking the substring: `str(callback)[1:-1]`, this is done several in other templates. It's a little ugly, and I'll post a PR later today which enables us to do it in a more robust and elegant way.

haakov: But you're saying that the raw type is being treated differently on master than on the older branches, yes?

haakov: Yes, caused by https://github.com/gnuradio/gnuradio/pull/4629

muaddib1984:
@haakov i see the 'no quotes' convenience function. I'll integrate that into the xmlrpc_client block. Once i've confirmed it's good to go, should i submit a separate pull request with the new block code and simpler examples?

willcode:
@muaddib1984 Keep the current PR so we don't lose the history. Push (or force push) onto your current (tmp) branch.

muaddib1984:
The latest commit is dependent on @haakov  's 'no_quotes' function being merged into master #5016 
This commit addresses the following issues:
-fixed xmlrpc server documentation to use python3 syntax in example client code
-simplified examples to exclude zmq and keep everything except the client in the server flowgraph
both from  [here](https://github.com/gnuradio/gnuradio/pull/5008#issuecomment-902994901) 
-Added comments in example flowgraphs to articulate client server flowgraph relationship
-using the no_quotes function and changing the callback type to string makes the block behave as advertised. I made the decision on string as it's more direct to just input the name of the server variable than to make the user infer that it needs quotes. 
-hardcoded the 'set_' portion of the XMLRPC call into the client code. (see below)

I think get/set is a new feature topic, but one i'd like to help out with. The get method has different requirements from the standpoint of the flowgraph and makes things a little more complicated since the client graph is dependent on the server being up and accessible. I could see a Get/Set  'option' section in the client block. I could also see 2 client blocks,  xmlrpc_client_get and xmlrpc_client_set. If there's a good use case, adding introspection to the server could give the client_get block the ability to query the server for available functions in the server-side flowgraph and then put them into a dropdown or list in the client block for the user to choose from. That would add a lot of convenience when keeping function names between client and server organized and accurate. Just my 2 cents.

mormj:
@willcode - will this need a backport to 3.9?

willcode:
Yes, if this goes in master, I'll backport it.

haakov:
I've got one concern remaining about this PR, and that is the hardcoded "set_" prefix. Is it necessary? If so, I think it should be mentioned in the block's documentation field.

willcode:
I'm also thinking that hardcoding the `set_` is kind of limiting. But this block is already really limited. What is the use case for a XMLRPC client block? That may help resolve what this should look like.

Each control gets connected to it's own client, and each client has its own `ServerProxy` (http connection), which isn't terribly effieient. The XMLRPC server side, on the other hand, works as a single receiver that can fan out calls to a whole flowgraph. Whatever we do with the single-purpose client here, there's probably a need for a client that can mux changes from multiple controls.

muaddib1984:
OK, I like this discussion. I think I hardcoded the set_ parameter when I was getting my initial hack working. If it’s all right with you guys, I will remove The hard coding and make the default parameter set_  Which is what it’s been forever. This will accomplish the “bug fix “portion of the pull request.

Following that I would like to start implementing these pieces from the bigger discussion to streamline and add features to the client portion. I’m sure you guys have seen my similar comments about where it goes from here. I’m a big fan of working through adding the get_  parameter, Which currently will work (once I remove that hard coding :] ) but really integrating it in to the architecture properly. 
I’ve used it before in a very limited way so I think I have a good handle on the functionality that would be desirable from a user. I think Jeff’s idea about making the client be one block is really good as well. 

Should I go ahead and make this one change so we can close this bug fix pull request and then continue the discussion further?


Thx


<end transmission>

willcode:
That sounds good. It should work so that the `volume` slider on the client side can be attached to a `set_gain()` function on the server side. No reason to expect that the names are the same. Another example would be a periodic message on the client side that causes an arbitrary function to be executed on the server side. The server code would not necessarily be emitted by GRC, though people seem to think that GRC is the only way to use GNU Radio now. So a pushbutton on the client side might just execute `stop()` on the server side.

muaddib1984:
@haakov @willcode sorry for the delay.  
- I Just pushed a new commit where I removed the hardcoding for set_, changed the default value for the callback parameter to 'set_' and also changed the documentation on the server side per @haakov 's comment [here](https://github.com/gnuradio/gnuradio/pull/5008#issuecomment-902994901) to align with python 3 syntax for an example client.

- Please note, this pull request is dependent on #5016

haakov:
Okay great! I'll test it out later today. In the meantime, you might want to tidy up your commits? You can look at the commit log for inspiration

muaddib1984:
> Oh, sorry! Welcome smile
> 
> There's four things you'd want to do:
> 
>     * Right now, you've got 4 commits in this pull request. Some of them can be squashed together, and I suppose two commits would make sense, one for fixing the block YAML and one for adding the examples. The commit messages are usually prefixed with the affected in-tree module, so you'd want to rename them slightly. Something like "blocks: Add XMLRPC examples" and "blocks: Fix XMLRPC block YAML" perhaps.
> 
i think i got this right. 
>     * For this PR to be accepted, you will have to sign-off your commits using `git commit -s`. For reference: https://github.com/gnuradio/gnuradio/blob/master/CONTRIBUTING.md#dco-signed
> 
should be ok now
>     * The example `.py` files can be removed if they are just the code that is created when compiling/generating the `.grc` files.
ok, no more .py files
> 
>     * Your `siggen_controller_xmlrpc_client.grc` example contains a block that's missing on my end, `myxmlrpc_client`
oops sorry, its removed.

PR #5011, opened 2021-08-16
network: tuntap knows what it's doing when it uses strncpy, suppress warning

marcusmueller:
with -Werror, this became a problem (stopped builds under Debian 11).

Since we're setting the last byte to zero in the caller of this function, the truncation happens, if at all, there, not here.

### DIFF
--- a/gr-network/lib/tuntap_pdu_impl.cc
+++ b/gr-network/lib/tuntap_pdu_impl.cc
@@ -1,6 +1,7 @@
 /* -*- c++ -*- */
 /*
  * Copyright 2013 Free Software Foundation, Inc.
+ * Copyright 2021 Marcus Müller
  *
  * This file is part of GNU Radio
  *
@@ -52,13 +53,22 @@ tuntap_pdu_impl::tuntap_pdu_impl(std::string dev, int MTU, bool istunflag)
     strncpy(dev_cstr, dev.c_str(), IFNAMSIZ);
     dev_cstr[IFNAMSIZ - 1] = '\0';
 
-
     bool istun = d_istunflag;
+
+#if defined(__GNUG__) && !defined(__clang__)
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wstringop-truncation"
+#endif
+
     if (istun) {
         d_fd = tun_alloc(dev_cstr, (IFF_TUN | IFF_NO_PI));
     } else {
         d_fd = tun_alloc(dev_cstr, (IFF_TAP | IFF_NO_PI));
     }
+
+#if defined(__GNUG__) && !defined(__clang__)
+#pragma GCC diagnostic pop
+#endif
 
     if (d_fd <= 0)
         throw std::runtime_error(
@@ -117,9 +127,10 @@ int tuntap_pdu_impl::tun_alloc(char* dev, int flags)
      * the kernel will try to allocate the "next" device of the
      * specified type
      */
-    if (*dev)
+    if (*dev) {
+        // copy at most IFNAMSIZ - 1: If everything went well, the last byte is 0 anyway
         strncpy(ifr.ifr_name, dev, IFNAMSIZ - 1);
-
+    }
     /* try to create the device */
     if ((err = ioctl(fd, TUNSETIFF, (void*)&ifr)) < 0) {
         close(fd);


marcusmueller:
Waiiiit, these pragmas need to go around the actual call to strncpy, it seems

ghostop14:
I think it looks okay since the constructor explicitly does a memset and assigns the last byte to 0x00 after copy.  6 of one half dozen of another but maybe in the constructor do the memcpy as:
    strncpy(dev_cstr, dev.c_str(), IFNAMSIZ - 1);

which would be a couple less instructions not needing this.
    dev_cstr[IFNAMSIZ - 1] = '\0';

drmpeg (approved):
Seems to only be a problem with gcc 10.2.1 (which is some Debian special version of gcc). I'm not seeing this with gcc 10.3.0 on Ubuntu.

I'm fine with this fix.

mbr0wn (approved):

gr-network/lib/tuntap_pdu_impl.cc
```diff
@@ -54,11 +55,14 @@ tuntap_pdu_impl::tuntap_pdu_impl(std::string dev, int MTU, bool istunflag)
 
 
     bool istun = d_istunflag;
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wstringop-truncation"
```
mbr0wn, line 59: How about:
```suggestion
#if if defined(__GNUG__) && !defined(__clang__)
#    pragma GCC diagnostic push
#    pragma GCC diagnostic ignored "-Wstringop-truncation"
#endif
```

gr-network/lib/tuntap_pdu_impl.cc
```diff
@@ -54,11 +55,14 @@ tuntap_pdu_impl::tuntap_pdu_impl(std::string dev, int MTU, bool istunflag)
...
     if (istun) {
         d_fd = tun_alloc(dev_cstr, (IFF_TUN | IFF_NO_PI));
     } else {
         d_fd = tun_alloc(dev_cstr, (IFF_TAP | IFF_NO_PI));
     }
+#pragma GCC diagnostic pop
```
mbr0wn, line 65:
```suggestion
#if defined(__GNUG__) && !defined(__clang__)
#    pragma GCC diagnostic pop
#endif
```

marcusmueller:
In all honesty, I did not test, but unknown pragmas are but a warning under msvc, afair

On September 1, 2021 11:25:55 AM GMT+02:00, Martin Braun ***@***.***> wrote:
***@***.*** commented on this pull request.
>
>
>
>> @@ -54,11 +55,14 @@ tuntap_pdu_impl::tuntap_pdu_impl(std::string dev, int MTU, bool istunflag)
> 
> 
>     bool istun = d_istunflag;
>+#pragma GCC diagnostic push
>
>Do these `pragma`s work fine on Windows?
>
>-- 
>You are receiving this because you authored the thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/gnuradio/gnuradio/pull/5011#pullrequestreview-743691556
-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

mbr0wn:
> In all honesty, I did not test, but unknown pragmas are but a warning under msvc, afair 

Warnings are bad. The suggestions I made are copied from UHD, but I also checked other places where we GCC pragmas are used, and these `#ifdefs` seem common.

marcusmueller:
Sorry, I think my last mail came across differently than I intended: your suggestions are welcome and good, I just wanted to point out I hadn't tested it, assuming the breakage would be limited to warnings. I'll work in your changes later today, hopefully

On September 1, 2021 3:00:19 PM GMT+02:00, Martin Braun ***@***.***> wrote:
>> In all honesty, I did not test, but unknown pragmas are but a warning under msvc, afair 
>
>Warnings are bad. The suggestions I made are copied from UHD, but I also checked other places where we GCC pragmas are used, and these `#ifdefs` seem common.
>
>-- 
>You are receiving this because you authored the thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/gnuradio/gnuradio/pull/5011#issuecomment-910259866
-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

mormj:
Holding off merging this one until the warnings situation is sorted out and validation that it builds fine in Windows (since it adds #ifs with GCC checks)

drmpeg:
On clang 12.0, I get a warning.
```
/home/ubuntu/xfer/gnuradio/gr-network/lib/tuntap_pdu_impl.cc:59:32: warning: unknown warning group '-Wstringop-truncation', ignored [-Wunknown-warning-option]
#pragma GCC diagnostic ignored "-Wstringop-truncation"
```
I'd say @mbr0wn suggestion to add ```#if defined(__GNUG__) && !defined(__clang__)``` would be nice.

marcusmueller:
added @mbr0wn's excellent pragma-guard

gr-network/lib/tuntap_pdu_impl.cc
```diff
@@ -54,11 +55,14 @@ tuntap_pdu_impl::tuntap_pdu_impl(std::string dev, int MTU, bool istunflag)
 
 
     bool istun = d_istunflag;
+#pragma GCC diagnostic push
```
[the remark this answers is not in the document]
mbr0wn: Huh, I would have though our clang-format wants those indents.

mbr0wn (approved):

willcode:
If we find it necessary to do stuff like this more broadly, some compiler-independent macros would be useful (later).

marcusmueller:
for some reason, this doesn't seem to fix https://github.com/gnuradio/gnuradio/pull/4981/checks?check_run_id=3536756557#step:5:860

drmpeg:
@marcusmueller I have a Debian 11 system (on armv7). I'll take a look at this.

PR #5028, opened 2021-08-23
runtime: Accelerator Device Support (aka ngsched)

dsorber:
The following changes are from the Custom Buffer/Accelerator Device Support project (aka ngsched). The purpose of this project was to add support for accelerator devices directly to GR by creating a "custom buffer" interface and added the necessary hooks for the scheduler to handle data movement, all while maintaining backwards compatibility for existing GR blocks (both in tree and OOT modules).

Given the extensive (though backwards compatible) nature of these changes, an overview and detailed design documentation is available in this wiki page: https://github.com/gnuradio/gnuradio-ngsched/wiki

NOTE: This project has been developed in the [ngsched](https://github.com/gnuradio/gnuradio-ngsched) repo but since ngsched was not created as a fork it is not possible to open a PR back to the main GR repo. Therefore the necessary commits have been cherrypicked into a branch in my personal fork for purposes of creating this PR.


jsallay:
I like a lot of the ideas here.  Have you done any benchmarking?  I'd like to see if there is any impact on the data throughput when just running on the host.  Also, I didn't see it, but does this support having the tag stream on the device?  If data is offloaded then we would need to be able to get/set tags on the device too.

dsorber:
> I like a lot of the ideas here. Have you done any benchmarking? I'd like to see if there is any impact on the data throughput when just running on the host. Also, I didn't see it, but does this support having the tag stream on the device? If data is offloaded then we would need to be able to get/set tags on the device too.

@jsallay - Yes, I have done benchmarking but forgot to include it. I just put together [this wiki page](https://github.com/gnuradio/gnuradio-ngsched/wiki/Benchmark-Data) with benchmark data.  That benchmark data was collected for a presentation two months ago. There have been a few interface-related changes since then but nothing that should really affect performance.  That said, I will be generating fresh benchmark data for my presentation at GRCon and I will update that wiki page when the updated data is available.

These changes do not support having the tag stream on the device directly.  I'm not sure how that would work given the pmt data in the tag object (I realize you are working on the pmt overhaul but I haven't had time to follow that very closely).  I am very open to discussion about how things could/should work.

mormj:
@dsorber - thank you for including the wiki documenation.  That tremendously helps when trying to understand these changes.
Also, a good introduction will take place at GRCon21: https://events.gnuradio.org/event/8/contributions/37/
This should be a good time to kick off the discussion of the mechanics of the accelerator support
Definitely going to let this one stew for a while, but would like to get this into the 3.10 release as an "experimental" feature that is only enabled with a CMake switch (if that is feasible)

PR #5031, opened 2021-08-28
grc: Generate c++ vector from list compre in var

### LINKED ISSUE #4955: grc: cpp generation fails to generate vector variable
Define in grc a variable block with value
[float(x) for x in range(4)]

Python code will be generated but cpp code generation fails with:

  File "/usr/local/gnuradio/lib64/python3.9/site-packages/gnuradio/grc/gui/Application.py", line 720, in _handle_action
    generator.write()
  File "/usr/local/gnuradio/lib64/python3.9/site-packages/gnuradio/grc/core/generator/cpp_top_block.py", line 66, in write
    self._variable_types()
  File "/usr/local/gnuradio/lib64/python3.9/site-packages/gnuradio/grc/core/generator/cpp_top_block.py", line 301, in _variable_types
    var.format_expr(var_types[str(var.params['id'].value)])
  File "/usr/local/gnuradio/lib64/python3.9/site-packages/gnuradio/grc/core/blocks/block.py", line 479, in format_expr
    self.vtype = get_type(value, py_type)
  File "/usr/local/gnuradio/lib64/python3.9/site-packages/gnuradio/grc/core/blocks/block.py", line 460, in get_type
    return 'std::vector<' + get_type(str(evaluated[0]), type(evaluated[0])) + '>'
TypeError: 'NoneType' object is not subscriptable

That is why in in block.py
 ##
       def format_expr(self, py_type):
        """
        Evaluate the value of the variable block and decide its type.
        Returns:
            None
        """
        value = self.params['value'].value
        self.cpp_templates = copy.copy(self.orig_cpp_templates)

        # Determine the lvalue type
        def get_type(element, _vtype):
            evaluated = None
            try:
                evaluated = ast.literal_eval(element)`

evaluated = ast.literal_eval(element) fails for 
element = [float(x) for x in range(4)]

with ValueError: malformed node or string: <ast.ListComp object at 0x7fd8a0e92eb0>

solomonbstoner:
Fixes #4955 .
Issue 4955 states that when a list comprehension is used in a variable
block, c++ code cannot be generated from the flowgraph. The cause of the
problem is `ast.literal_eval` which only considers a small number of
Python syntaxes to be valid. `eval` correctly evaluates a list
comprehension as valid python syntax.

### DIFF
--- a/grc/CMakeLists.txt
+++ b/grc/CMakeLists.txt
@@ -56,6 +56,12 @@ GR_PYTHON_CHECK_MODULE_RAW(
     NUMPY_FOUND
 )
 
+GR_PYTHON_CHECK_MODULE_RAW(
+    "asteval"
+    "import asteval"
+    ASTEVAL_FOUND
+)
+
 ########################################################################
 # Register component
 ########################################################################
@@ -69,6 +75,7 @@ if(NOT CMAKE_CROSSCOMPILING)
         CAIRO_GI_FOUND
         PANGOCAIRO_GI_FOUND
         NUMPY_FOUND
+        ASTEVAL_FOUND
     )
 endif(NOT CMAKE_CROSSCOMPILING)
 
--- a/grc/core/blocks/block.py
+++ b/grc/core/blocks/block.py
@@ -14,7 +14,7 @@ import copy
 
 import re
 
-import ast
+from asteval import Interpreter
 
 from ._templates import MakoTemplates
 from ._flags import Flags
@@ -53,6 +53,8 @@ class Block(Element):
 
     extra_data = {}
     loaded_from = '(unknown)'
+
+    aeval = Interpreter()
 
     def __init__(self, parent):
         """Make a new block from nested data."""
@@ -433,7 +435,7 @@ class Block(Element):
         def get_type(element, _vtype):
             evaluated = None
             try:
-                evaluated = ast.literal_eval(element)
+                evaluated = self.aeval(element)
                 if _vtype == None:
                     _vtype = type(evaluated)
             except ValueError or SyntaxError as excp:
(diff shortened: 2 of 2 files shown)


willcode:
`literal_eval` is safe, where `eval` is not. What are the implications? This would be another hook for read-only mode, too.

solomonbstoner:
Indeed. But `literal_eval` simply doesnt accept a number of valid Python syntaxes. So if we wanna allow c++ to be generated from list comprehensions in variables (described in #4955), only way I can think of is to use `eval`.

> Safely evaluate an expression node or a string containing a Python literal or container display. The string or node provided may only consist of the following Python literal structures: strings, bytes, numbers, tuples, lists, dicts, sets, booleans, and None.
> 
> This can be used for safely evaluating strings containing Python values from untrusted sources without the need to parse the values oneself. It is not capable of evaluating arbitrarily complex expressions, for example involving operators or indexing.

solomonbstoner:
I just found [`asteval`](https://newville.github.io/asteval/motivation.html#how-safe-is-asteval), which appears to be the compromise between security and usability that we need.

willcode:
Not a bad idea, and could be useful for GR R/O mode in general. This is a new external dep, though.

mormj:
Need to consider the implications of adding a new dependency here - but I like that there is an existing solution for doing safe evals.  Is this something that we could use more broadly

mormj:
Let's hold off on this one for further discussion since it does introduce a new dependency which I'm not sure about

solomonbstoner:
Got it. Ill find out how to solve that issue without adding a new dependency.

mormj:
Closing this PR for now - @solomonbstoner - please reopen this or open a new one if you come up with a different approach.  Thanks!

Issue #4956, opened 2021-07-26
Build failure with Qwt 6.2


ryanvolz:
The recent Qwt 6.2.0 release has removed `qwt_compat.h` (see item #10 [here](https://sourceforge.net/p/qwt/git/ci/develop/tree/CHANGES-6.2)), resulting in build failures with all current branches. For example, [here's a conda build of 3.9.2](https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=356344&view=logs&jobId=d0d954b5-f111-5dc4-4d76-03b6c9d0cf7e&j=d0d954b5-f111-5dc4-4d76-03b6c9d0cf7e&t=841356e0-85bb-57d8-dbbc-852e683d1642). This is probably not an immediate problem given that the new Qwt version can be avoided for a little while, but it's certainly something to be aware of and to start planning for now.

wkazubski:
Causing build problems on openSUSE Factory and Tumbleweed

atsampson:
There's also the problem with Qwt 6.2 that `QwtRasterData::setInterval` has been removed - subclasses like GNU Radio's `TimeRasterData` now need to implement `interval` themselves.

bmuzika:
I've got a PR up that addresses this issue (#5034). Review would be greatly appreciated.

wkazubski:
I backported your commit to GR 3.8.3.1. The most important change was to make changes in gr-qtgui/lib/timeRasterGlobalData.cc and gr-qtgui/lib/waterfallGlobalData.cc conditional (only for Qwt-6.2), to mainain compatibility with older Qwt 6.1.x

wkazubski:
I did some tests and everything is working for me

marcoferrara10:
Hello everyone,
I've encountered the same issue on compiling gnuradio  and I've compiled successfully the git version with the head at the commit bcba1acbee1e5b6ded326b4962c1eb5dcf040e28 by applying the patch proposed by @wkazubski.

@wkazubski  thank you so much.

willcode:
I put in a draft PR https://github.com/gnuradio/gnuradio/pull/5282 against maint-3.8 using 0003-enables-use-of-qwt-6_2.patch.txt to see what CI thinks. We would need to get versions for master and maint-3.9 before we can merge it though.

willcode:
#5282 passed CI on 3.8, so trying #5283 against 3.9.

bmuzika:
[...]

PR #5034, opened 2021-08-30
qt-gui: enables use of Qwt 6.2

bmuzika:
Adds the typedefs and includes needed to build with Qwt 6.2.

Fixes: #4956

Signed-off-by: Bill Muzika <bill.muzika@outlook.com>


cmake/Modules/FindQwt.cmake
```diff
@@ -51,13 +51,13 @@ if(QWT_INCLUDE_DIRS)
...
   string(COMPARE LESS ${QWT_VERSION} "5.2.0" QWT_WRONG_VERSION)
-  string(COMPARE GREATER ${QWT_VERSION} "6.2.0" QWT_WRONG_VERSION)
+  string(COMPARE GREATER ${QWT_VERSION} "6.3.0" QWT_WRONG_VERSION)
 
   message(STATUS "QWT Version: ${QWT_VERSION}")
   if(NOT QWT_WRONG_VERSION)
     set(QWT_FOUND TRUE)
   else(NOT QWT_WRONG_VERSION)
-    message(STATUS "QWT Version must be >= 5.2 and <= 6.2.0, Found ${QWT_VERSION}")
+    message(STATUS "QWT Version must be >= 5.2 and <= 6.3.0, Found ${QWT_VERSION}")
```
MehdiChinoune, line 60: `string(COMPARE GREATER_EQUAL ${QWT_VERSION} "6.3.0" QWT_WRONG_VERSION)`
`message(STATUS "QWT Version must be >= 5.2 and < 6.3.0, Found ${QWT_VERSION}")`

gr-qtgui/lib/timeRasterGlobalData.cc
```diff
@@ -46,6 +46,13 @@ TimeRasterData::TimeRasterData(const double rows, const double cols)
 
 TimeRasterData::~TimeRasterData() {}
 
+void TimeRasterData::setInterval(Qt::Axis axis, const QwtInterval& interval)
+{
+    d_intervals[axis] = interval;
+}
+
+QwtInterval TimeRasterData::interval(Qt::Axis a) const { return d_intervals[a]; }
+
```
MehdiChinoune, line 55:
```
#ifdef QWT_VERSION >= 0x060200
...
#endif
```

gr-qtgui/lib/waterfallGlobalData.cc
```diff
@@ -49,6 +49,12 @@ void WaterfallData::reset()
 
     _numLinesToUpdate = -1;
 }
+void WaterfallData::setInterval(Qt::Axis axis, const QwtInterval& interval)
+{
+    d_intervals[axis] = interval;
+}
+
+QwtInterval WaterfallData::interval(Qt::Axis a) const { return d_intervals[a]; }
```
MehdiChinoune, line 57:
```
#ifdef QWT_VERSION >= 0x060200
...
#endif
```

gr-qtgui/lib/timeRasterGlobalData.cc
```diff
@@ -46,6 +46,13 @@ TimeRasterData::TimeRasterData(const double rows, const double cols)
 
 TimeRasterData::~TimeRasterData() {}
 
+void TimeRasterData::setInterval(Qt::Axis axis, const QwtInterval& interval)
+{
+    d_intervals[axis] = interval;
+}
+
+QwtInterval TimeRasterData::interval(Qt::Axis a) const { return d_intervals[a]; }
```
MehdiChinoune, line 54:
```suggestion
#if QWT_VERSION >= 0x060200
void TimeRasterData::setInterval(Qt::Axis axis, const QwtInterval& interval)
{
    d_intervals[axis] = interval;
}

QwtInterval TimeRasterData::interval(Qt::Axis a) const { return d_intervals[a]; }
#endif
```

gr-qtgui/lib/waterfallGlobalData.cc
MehdiChinoune, line 57:
```suggestion
#if QWT_VERSION >= 0x060200
void WaterfallData::setInterval(Qt::Axis axis, const QwtInterval& interval)
{
    d_intervals[axis] = interval;
}

QwtInterval WaterfallData::interval(Qt::Axis a) const { return d_intervals[a]; }
#endif
```

cmake/Modules/FindQwt.cmake
```diff
@@ -51,13 +51,13 @@ if(QWT_INCLUDE_DIRS)
   set(QWT_VERSION "No Version")
   string(REGEX MATCH "[0-9]+.[0-9]+.[0-9]+" QWT_VERSION ${QWT_STRING_VERSION})
   string(COMPARE LESS ${QWT_VERSION} "5.2.0" QWT_WRONG_VERSION)
-  string(COMPARE GREATER ${QWT_VERSION} "6.2.0" QWT_WRONG_VERSION)
+  string(COMPARE GREATER ${QWT_VERSION} "6.3.0" QWT_WRONG_VERSION)
```
MehdiChinoune, line 54:
```suggestion
  string(COMPARE GREATER_EQUAL ${QWT_VERSION} "6.3.0" QWT_WRONG_VERSION)
```

cmake/Modules/FindQwt.cmake
MehdiChinoune, line 60:
```suggestion
    message(STATUS "QWT Version must be >= 5.2 and < 6.3.0, Found ${QWT_VERSION}")
```

gr-qtgui/lib/timeRasterGlobalData.cc
bmuzika: Thanks for those code suggestions. I'll make those changes and test them locally first with the CI docker images and update the PR.

cmake/Modules/FindQwt.cmake
```diff
@@ -51,13 +51,13 @@ if(QWT_INCLUDE_DIRS)
   set(QWT_VERSION "No Version")
   string(REGEX MATCH "[0-9]+.[0-9]+.[0-9]+" QWT_VERSION ${QWT_STRING_VERSION})
   string(COMPARE LESS ${QWT_VERSION} "5.2.0" QWT_WRONG_VERSION)
-  string(COMPARE GREATER ${QWT_VERSION} "6.2.0" QWT_WRONG_VERSION)
```
mormj, line 54: Why wouldn't this be valid if we are now allowing QWT 6.2 - seems the check would have been previously too permissive, but is now correct

cmake/Modules/FindQwt.cmake
```diff
@@ -51,13 +51,13 @@ if(QWT_INCLUDE_DIRS)
...
+  string(COMPARE GREATER_EQUAL ${QWT_VERSION} "6.3.0" QWT_WRONG_VERSION)
 
   message(STATUS "QWT Version: ${QWT_VERSION}")
   if(NOT QWT_WRONG_VERSION)
     set(QWT_FOUND TRUE)
   else(NOT QWT_WRONG_VERSION)
-    message(STATUS "QWT Version must be >= 5.2 and <= 6.2.0, Found ${QWT_VERSION}")
```
mormj, line 60: same here

mormj (changes requested):
@bmuzika - thank you for the changes here.  General question - Is there a way to pull some of the things into a separate header file, or ways we can prevent having all of the qt code littered with `#if`s
Or - can the ways of doing things differently because qwt_compat is gone - e.g. using QwtInterval - can those also be used for older Qwt versions, or did they just show up in 6.2.  Just looking for ways of keeping the code a bit neater if possible

bmuzika:
@mormj Thanks for the feedback. I wanted to start off by making the simplest possible changes that could work and get feedback.

My gut feeling is that phasing out the Qwt5 compatibility include would be the better approach given that the CI builds all target Qwt6, so I'll try to replace that include for Qwt6 builds older than 6.2 as well (Unless I get feedback that it would be better to have a gnuradio qwt_compat.h file instead)

As a procedural question, is it preferred that I rebase my changes and keep them in one commit or use new ones as I respond to these reviews?

cmake/Modules/FindQwt.cmake
bmuzika: After thinking about this some more, it should be left as it is in `master`. The check before was letting in a version that was unreleased (and incompatible). I've reverted the change.

cmake/Modules/FindQwt.cmake
bmuzika: I've reverted this change.

bmuzika:
@mormj - I've made the requested changes and replaced the compatibility header with the required headers. The new use of the QWT_VERSION define is because the API for the colorMap related functions changed in 6.2 but I kept the same behavior as in Qwt versions < 6.2.

willcode:
As discussed on matrix, it's possible that Qwt >= 6.0 can be assumed.
 - Almost definitely for master
 - Probably for 3.9
 - Maybe even for 3.8
Need input from people who know Qwt.

bmuzika:
@willcode - I've prepared a new branch [1] that changes handling of Qwt version checks. I've split it into two commits for ease of backporting (and deciding exactly which versions to support). It does stack on this PR, but that could change if needed.

[1] - https://github.com/bmuzika/gnuradio/tree/features/tighten-req-qwt

MehdiChinoune:
Unbelievable!
More than a month and a simple patch not merged!

mormj (approved):
still need to do testing, but the changes seem reasonable.

bmuzika:
I saw some precompiled header errors on CentOS. Do I need to rebase onto the latest `master`?

willcode:
@bmuzika give rebasing a try - not sure what all those pch errors are coming from.

bmuzika:
I've rebased my changes. On a local build with the Centos docker image, I got the same pch warnings as on the last run

willcode:
@marcusmueller any idea about the pch error here?

bmuzika:
pch errors "gone" (just suppressed by the flag in the new centos pipeline). All that is left is that failed test that appeared in several recent PRs. Any suggestions where to look for fixing that?

willcode:
Don't worry about that one - it's a CI problem not related to the PRs.

bmuzika:
I think I found the change that broke the waterfall plots. I'll do some more testing on more than just Fedora 34

bmuzika:
This branch [1] has changes that should allow my original patch to work. I think the problem was defining `setInterval` in Qwt 6.1. In that version, that method was already defined by Qwt.

[1] https://github.com/bmuzika/gnuradio/tree/bugfix/fix-qwt-6.1-operation

PR #5035, opened 2021-08-30
grc: Raise error for unfilled parameters

### LINKED ISSUE #5032: Fractional Resampler default Phase Shift parameter
In GRC version v3.10.0.0git-429-g655e232a, if I leave the 'Phase Shift' parameter blank in a Fractional Resampler, I get the following error:
<pre>
  File "/home/barry/GRdev/speech_proc_4.py", line 183  
    self.mmse_resampler_xx_0 = filter.mmse_resampler_cc(, 48000/samp_rate)  
                                                        ^  
SyntaxError: invalid syntax
</pre>
Also, GRC shows the Phase Shift as '0' even before opening the Parameters.

A test flowgraph is in https://pastebin.com/4cWJNhjL
Note line 111

solomonbstoner:
Fixes #5032 .
Issue 5032 results from an unfilled parameter erroneously indicated as
filled with a default value of 0. This PR makes grc indicate an error
when there is an unfilled parameter by returning None.

### DIFF
--- a/grc/core/params/param.py
+++ b/grc/core/params/param.py
@@ -214,7 +214,7 @@ class Param(Element):
                 except Exception as e:
                     raise Exception('Value "{}" cannot be evaluated:\n{}'.format(expr, e))
             else:
-                value = 0
+                value = None   # No parameter value provided
             if dtype == 'hex':
                 value = hex(value)
             elif dtype == 'bool':


solomonbstoner:
**Before the PR:**
The empty parameter "Phase Shift" is falsely indicated to be filled with value 0. See #5032.
Screenshot from 2021-08-30 20-30-23

**After the PR:**
The empty parameter "Phase Shift" is highlighted in red in error to indicate that its been left empty.
Screenshot from 2021-08-30 20-29-24
Screenshot from 2021-08-30 20-29-20

Error thrown is:
> Param - Phase Shift(phase_shift):
>	Expression None is invalid for type 'real'.

mormj (approved):
This is a very sensible fix, and agreed having no error until flowgraph rendering is frustrating, and incorrect behavior

dkozel:
Some (many?) blocks do not currently come with default values for parameters that are optional. I think this change would break any flowgraph that has one of those optional values left blank.

For values which are required asserts should be added to the individual block YAML files. This will correctly mark blocks as in an Error state. I forget if an assert failure will highlight the failing parameter though, that'd be an excellent enhancement.

solomonbstoner:
Tested against two examples of optional parameters:
1. Packet Header Parser block's "Formatter Object" parameter (which is a raw dtype)
Prior to this PR, when left empty, the "Formatter Object" parameter is marked as '0', but this PR would have it marked as 'None' instead of throwing an error because the `evaluate` function is called by `rewrite` in `base.py` instead of `validate_scalar` in `dtypes.py`. 
2. Taps of filters (which are vectors dtype)
Vector dtypes are not affected by this PR. They were not marked by '0's when left empty, so will not be erroneously marked as false when left empty by this PR. 

Let me see its effects on the USRP and synchronization blocks.

solomonbstoner:
This is how it looks like for the synchronization and USRP blocks.

**Before PR:**
The scalar parameters which are left blank are marked with value '0' (as described in #5032 ). 
Screenshot from 2021-08-31 23-12-52

**After PR:**
The scalar parameters which are left blank are highlighted in red. 
Screenshot from 2021-08-31 23-10-19
I think the optional parameters you refer to are those like device arguments. Since those are strings, they arent affected by the PR or even issue #5032 . They can be left blank like before. 
Screenshot from 2021-08-31 23-15-07

mormj:
Love that for numerical fields this creates the visual distinction between a default value (from the .yml) and an empty value.  I'm on board with this.

mormj:
@solomonbstoner 's  testing on this appears to be thorough.  Let's hold off on a backport and see if it creates any issues for folks on master.  But it LGTM!

willcode:
Skipping backport to 3.8 since GRC in 3.8 might be used to open older flowgraphs. If someone wants to do some tests, we can revisit.

PR #5042, opened 2021-09-03
qtgui: Replace pyqwidget() with qwidget() (backport to maint-3.9) 

Yamakaja:
Previously, two versions of the `qtwidget` functions existed,
`qwidget()` and `pyqwidget()`, with the only difference being that
`qwidget()` returned a pointer to the `QWidget` object managed by the
corresponding block, while `pyqwidget()` returned that same pointer, but as
an integer (Or `PyLong` in this case).

While `qwidget()` is used by C++ code accessing the widgets,
`pyqwidget()` is only used for the python interface. This makes these
two methods redundant, thus this commit entirely removes `pyqwidget()`,
and modifies the `qwidget()` python wrapper to behave like
`pyqwidget()`. Note that we can be fairly confident that this change
will not effect potential users of `qwidget()`, because any invocation
on the objects previously returned by `qwidget()` would cause a
segmentation fault.

This commit also fixes a memory leak:

Internally, the `pyqwidget()` functions were returning a PyLong `PyObject *`,
which was then upwrapped in a pybind trampoline without decrementing
the reference count of that python object.

Backport of #4994

### DIFF
--- a/docs/usage-manual/(exported from wiki) QT GUI.txt
+++ b/docs/usage-manual/(exported from wiki) QT GUI.txt
@@ -154,7 +154,7 @@ import sys, sip
                              samp_rate,      #bw
                              "QT GUI Plot")  #name
  
-     self.snk_win = sip.wrapinstance(self.snk.pyqwidget(), Qt.QWidget)
+     self.snk_win = sip.wrapinstance(self.snk.qwidget(), Qt.QWidget)
      self.snk_win.show()
  
  def main():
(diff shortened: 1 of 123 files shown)


willcode:
Thanks for the backport. Any differences between this and the one for master?

Yamakaja:
I'd hope so - the merge applied (almost) cleanly ;)

willcode:
Diff'd the diffs. There were a couple of `.` -> '->`, but otherwise the same. Any idea what those changes were from?

Yamakaja:
See https://github.com/gnuradio/gnuradio/commit/4b7006db76b570e4d916e263301333d2f4d2a2df

willcode (approved):
Same as master, with minor changes due to the "remove manual memory management" changes on master.

PR #5045, opened 2021-09-07
spdlog pre-work: IWYU (boost::format, iomanip, iostream, sstream)

mbr0wn:
This is the "safe and boring" pre-work for the spdlog PR. It does not include the removal of boost/format.hpp from logger.h so we can merge this without breaking any OOT.

### DIFF
--- a/gnuradio-runtime/lib/buffer.cc
+++ b/gnuradio-runtime/lib/buffer.cc
@@ -15,6 +15,7 @@
 #include <gnuradio/buffer.h>
 #include <gnuradio/integer_math.h>
 #include <gnuradio/math.h>
+#include <boost/format.hpp>
 #include <algorithm>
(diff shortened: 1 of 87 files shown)


mbr0wn:
@mormj I would suggest merging this if CI is happy. Both @marcusmueller and I are :+1: with these commits.

mormj (approved):

marcusmueller:
Should we be squashing this? It's not really as if these individual commits will help a lot with bisecting later on..

mbr0wn:
Matter of taste, I guess. I'm OK leaving the commits. When I do a git log on a directory or file, I'm always happy when there's fewer all-repo commits to parse. I'm also OK squashing. But right now it's not squashed... so how about we tick this off in our heads and work on the next problem.

PR #5052, opened 2021-09-08
gr-network: Fixup tuntap suppress warning on Debian 11.

drmpeg:


mbr0wn (approved):
Much better. Puts the pragmas next to the actual code that's causing the warning.

PR #5075, opened 2021-09-20
ci: Only run clang-format and Doxygen conditionally

mbr0wn:
clang-format is only necessary when C++ files changed. Doxygen is only
necessary when headers, .dox, or any file in the docs/doxygen folder
changed.


mbr0wn:
This shaves about a minute off of CI runs. Not a big deal, but I'm always so impatient :)

mbr0wn:
PS: CI run passed, I cancelled after the relevant section.

mbr0wn:
...not working yet. I'll reopen when ready for prime time.

PR #5086, opened 2021-09-21
channels: Fix bad initialization of RNG seeds.

MattEttus:
---
name: channels: Fix bad initialization of RNG seeds.
about: 
title: ''
labels: ''
assignees: ''

---

# Pull Request Details

## Description
Previosuly the block attempted to initialize RNGs in the constructor initializers, but this does not work.  Moved the initialization to the body of the constructor.

## Checklist

### DIFF
--- a/gr-channels/lib/flat_fader_impl.cc
+++ b/gr-channels/lib/flat_fader_impl.cc
@@ -35,11 +35,12 @@ flat_fader_impl::flat_fader_impl(uint32_t N, float fDTs, bool LOS, float K, uint
       scale_los(sqrtf(d_K) / sqrtf(d_K + 1)),
       scale_nlos(1 / sqrtf(d_K + 1))
 {
+    rng_1 = std::mt19937(seed);
+    rng_2 = std::mt19937(seed + 1);
+
     d_theta = dist_1(rng_1);
     d_theta_los = dist_1(rng_1);
 
-    rng_1 = std::mt19937(seed);
-    rng_2 = std::mt19937(seed + 1);
 
     // generate initial phase values
     for (int i = 0; i < d_N + 1; i++) {
--- a/gr-channels/lib/flat_fader_impl.cc
+++ b/gr-channels/lib/flat_fader_impl.cc
@@ -15,15 +15,12 @@ namespace gr {
 namespace channels {
 
 flat_fader_impl::flat_fader_impl(uint32_t N, float fDTs, bool LOS, float K, uint32_t seed)
-    : rng_1(seed),
-      dist_1(-GR_M_PI, GR_M_PI),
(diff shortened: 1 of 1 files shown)


gr-channels/lib/flat_fader_impl.cc
```diff
@@ -38,6 +35,12 @@ flat_fader_impl::flat_fader_impl(uint32_t N, float fDTs, bool LOS, float K, uint
       scale_los(sqrtf(d_K) / sqrtf(d_K + 1)),
       scale_nlos(1 / sqrtf(d_K + 1))
 {
+    d_theta = dist_1(rng_1);
```
willcode, line 38: rng_1 is used here but is set below. Is that intentional?

MattEttus: yes, my mistake, I have pushed a fix to reverse it.

willcode (approved):

willcode:
Backport note: 3.8 uses boost instead of std, and the code is a little different.

marcusmueller:
Debian 11 failure is due to a volk bug, not a blocker here

marcusmueller:
> Backport note: 3.8 uses boost instead of std, and the code is a little different.

Ha! We've got xoroshiro128p.h so that we depend on neither boost nor std::mt19937 (both being slow) (and originally, because there were many places where people used `rand` / `srand` to generate long sequences (ouch) of uniform (double-ouch) random variables in multi-threaded GNU Radio (pain) from back in the STS days, and that had to be replaced with something).

I'm not saying we need to replace `<random>`/`<boost/random.h>` mt19937 with `"gnuradio/xorohsiro128p.h"`, as that leaves the problem of how to generate anything but uniform 64 bit ints as an exercise for the reader. Just a heads up that we do have a faster random engine if this ever becomes a bottleneck.

marcusmueller:
oops: 

> I'm not saying we need to replace `<random>`/`<boost/random.h>` mt19937 with `"gnuradio/xorohsiro128p.h"`, as that leaves the problem of how to generate anything but uniform 64 bit ints as an exercise for the reader. Just a heads up that we do have a faster random engine if this ever becomes a bottleneck.

I forgot I've implemented `gr::xoroshiro128p_prng`, which works exactly the same as `std::mt19937`, but again, not a strong recommendation to use that: it's more performant, that's all.

mormj (approved):
LGTM

willcode:
The v3.8 code has the same problem. Code was restructured since then, so cherry picking won't apply. Anyone want to take a hack at it?

willcode:
Skipping backport to 3.8. If someone wants to do the backport and test, we can revisit.

PR #5088, opened 2021-09-22
block comment font size incorrect, ignoring config file

### LINKED ISSUE #5033: Comments within the GRC canvas are always rendered in the coded-in default font size
Enable block comments.   Install a comment.

Regardless of the overall font-size setting in the config.conf file ([grc] section), the comments are rendered in the coded-in default, which is 8-point.  For those of us on high-DPI displays they're barely readable, even if you don't have ~60yo eyes, and if you're trying to do something like a tutorial, the audience may well not be able to read the comments, which rather defaults the purpose.

cbdonohue:
# Pull Request Details

## Description

Font size was being imported incorrectly resulting in the default font size always being used for comments.
Comment font size should change when canvas_font_size changes in grc.conf

## Related Issue

Fixes #5033

gui/grc/canvas

## Testing Done

Visual inspection if comment font size changes when canvas_font_size changes in grc.conf
before
after

## Checklist


ei8fdb (approved):
Yup, makes sense. A good PR to improve accessibility.

While not relevant to this PR specifically, I don't think there is currently an easy way for a user to modify font size. PR #4563 would help this.

mormj:
@cbdonohue - haven't trace through the grc code - so is `..Constants` different than `Constants.` - I'm guessing the latter gets updated with the conf file?

PR #5092, opened 2021-09-23
modify header_payload_demux

wsx000:
---
name: Pull Request Template
title: ''
labels: ''
assignees: ''

---

# Pull Request Details

## Description
i modify two files: gr-digital/lib/header_payload_demux_impl.cc and gr-digital/lib/header_payload_demux_impl.h in order to solve the problem of dropped frames. 
i use the example in gr-digital/examples/ofdm/rx_ofdm.grc to test my usrp equipment, i found a problem that even in an ideal channel simulation environment, frame dropping will occur randomly when i tried to modify it to transmit a picture , i can‘t uunderstood this phenomenon for a long time ， so i find a bug may lead to this issue in gr-digital/lib/header_payload_demux_impl.cc/h . 
When the original code consumes in general_work(), it ignores that if the data of the length of the current frame is consumed, the trigger signal from the Schmidl & Cox OFDM synch. block of the next frame of data may be overwritten, resulting in the failure of extracting the next frame of data ( Drop frame). So I modified the code so that when the code consumes, it also considers the position of the trigger signal of the next frame of data to prevent it from being eaten by the current frame. 

this will have a influence on block named Header/Payload Demux.

## Testing Done
I modified the original routine. An ofdm symbol uses 128 carriers, so there are 1920 constellation points in a burst of data. At first I found the phenomenon of frame dropping. I printed the position of the trigger signal of the previous few frames of data as following :
```
143
2061
3981
......
```
i found When the two trigger signals differ by 1918 data(such as 2061-143=1918), the frame drop will occur next. obviously the normal interval is 1920. so i guess there may be a situation in the source code that the current frame ate the trigger signal of the next frame.so i mainly modified the code in `find_trigger_signal` in `gr-digital/lib/header_payload_demux_impl.cc` as following:

```
......
    int trigger_nums = 0;
    if (max_rel_offset < skip_items) {
        return rel_offset;
    }
    if (in_trigger) {
        for (int i = skip_items; i < max_rel_offset; i++) {
        for (int i = skip_items; i < max_rel_offset; i++) {
            if (in_trigger[i]) {
                trigger_nums++;
                // record location of the first trigger
                if(trigger_nums == 1) {
                    rel_offset = i;
                }
                // If there is a second trigger signal,record offset of between the first and second trigger
                else if (trigger_nums == 2) {
                    d_next_trigger_offset = i - rel_offset;
                    break;
                }
            }
        }
    }
......
```
and modified the consume method in `general_work()` as following:

```
            // Calculate whether the current frame consumes the trigger signal of the next frame data
            // If yes, then make corrections
            int consume_compensation = 0;
            int consume_nums = 0;
            // Calculate total consumption
            consume_nums = (d_curr_payload_len + d_header_len) * (d_items_per_symbol + d_gi) + 
                            d_header_padding_total_items + 
                            d_curr_payload_offset - 
                            items_padding ;
            // make a decision
            if(d_next_trigger_offset <= consume_nums) {
                consume_compensation = consume_nums - d_next_trigger_offset + 1;
            }
            else {
                consume_compensation = 0;
            }
            // add the corrections
            const int items_to_consume =
                d_curr_payload_len * (d_items_per_symbol + d_gi) - items_padding - consume_compensation;
```
i add a parameter named consume_compensation to ensure that the trigger of the next frame of data will not be eaten.
Then I recompiled and installed the modified code, and successfully solved the problem of dropped frames

## Checklist

- [ ] I [have signed my commits before making this PR](https://github.com/gnuradio/gnuradio/blob/master/CONTRIBUTING.md#dco-signed)

### DIFF
--- a/gr-digital/lib/header_payload_demux_impl.cc
+++ b/gr-digital/lib/header_payload_demux_impl.cc
@@ -95,6 +95,7 @@ header_payload_demux_impl::header_payload_demux_impl(
             io_signature::make2(1, 2, itemsize, sizeof(char)),
             io_signature::make(
                 2, 2, (output_symbols ? itemsize * items_per_symbol : itemsize))),
+      d_next_trigger_offset(0),
       d_header_len(header_len),
       d_header_padding_symbols(header_padding / items_per_symbol),
       d_header_padding_items(header_padding % items_per_symbol),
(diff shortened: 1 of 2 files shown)


gr-digital/lib/header_payload_demux_impl.cc
```diff
@@ -361,14 +381,24 @@ int header_payload_demux_impl::find_trigger_signal(int skip_items,
                                                    const unsigned char* in_trigger)
 {
     int rel_offset = max_rel_offset;
+    int trigger_nums = 0;
     if (max_rel_offset < skip_items) {
         return rel_offset;
     }
     if (in_trigger) {
+        for (int i = skip_items; i < max_rel_offset; i++) {
```
willcode, line 389: Is this a duplicate line?

wsx000: Oh! yes, I didn't check it carefully,i'm sorry

PR #5105, opened 2021-09-28
global: Replace stdio logging with logger

Yamakaja:
This commit replaces many uses of std::c{out,err} and printf with the
appropriate GR_LOG_* directives.

I'm opening this PR as a draft, because in many places this wasn't as straight-forward as replacing a std::cout with a GR_LOG. There are compile time log conditions, multi-line outputs, and more weird things.


drmpeg:
Please revert the changes to ```gr-dtv/lib/dvbt2/dvbt2_p1insertion_cc_impl.cc```

mormj:
@drmpeg - can you please elaborate on why those should remain as printfs?

Yamakaja:
> @mormj
> 
> 1. They are being controlled by a block parameter.
> 2. I absolutely, positively do not want the messages prefixed with any other text.
> 
> You can change them to `std::cout`, but do not change the functionality of the block or the format of the message.

Are you parsing those elsewhere or something like that?

drmpeg:
@Yamakaja

Since the print out enable is controlled, any prefix text from ```gr::log``` is just spam. That is, since you've manually enabled the print out, knowing the block name source of the message is superfluous. Also, the resulting message with ```gr::log``` is more than 80 characters.

In addition, the print out is continuous with a typical period of 250 milliseconds, but could be as fast as 5 milliseconds. Adding superfluous text seems unwise at higher rates.

It's meant to be a tuning aid for the PAPR block. The PAPR block is very CPU intensive, so you want to tune for a clip level and number of iterations that's actually possible with your CPU. Once you've characterized your PAPR parameters, you shut this function off and never use it again.

What would be useful is a ```gr::log``` option to not print the prefix. Something like ```GR_LOG_NOPREFIX``` or ```GR_LOG_SILENT```.

Yamakaja:
* Removed changes from gr-dtv
* Fixed formatting issues

Yamakaja:
* Fixed header hashes

willcode:
Leaving the Backport-3.9 label on for now, in case someone wants to work on it.

willcode:
In backport to 3.9, I removed mods to gr-blocks/lib/qa_block_tags.cc due to differences in logging code. The rest worked out pretty easily.

PR #5109, opened 2021-09-29
runtime: fix pybind of get_tags_in_window

### LINKED ISSUE #5005: Calling `get_tags_in_window` from Python is broken in 3.9
It doesn't add the offset for the number of previously read samples, so in effect it acts exactly like `get_tags_in_range`. This only happens when the function is called on a Python block, not with any of the C blocks.

I can produce a failing test with this behavior by modifying the [`qa_block_gateway.py`](https://github.com/gnuradio/gnuradio/blob/1f1733eb4489b48fb73509e7806df19e1c738092/gr-blocks/python/blocks/qa_block_gateway.py#L125) script so that the `tag_source` skips outputting a tag in the first work call:
```
        # put code here to fill the output items...
        if self.nitems_written(0) == 0:
            # skip tagging in the first work block
            return num_output_items
```
Resulting in:
```
$ python qa_block_gateway.py 
........F
FAIL: test_tags_win (__main__.test_block_gateway)
  File "/home/rvolz/repos/gnuradio/gr-blocks/python/blocks/qa_block_gateway.py", line 287, in test_tags_win
    self.assertEqual(sink.key, "example_key")
AssertionError: None != 'example_key'

Ran 9 tests in 0.355s

FAILED (failures=1)
```

I haven't tracked down a cause for this, but presumably it happened with the SWIG->pybind11 transition since I see it with 3.9.2.0 but not 3.8.3.1. The easy workaround is to just use `get_tags_in_range` instead and add the offset manually.

mormj:
Fix issue where get_tags_in_window was bound to and consequently had the
same behavior as get_tags_in_range

Fixes #5005

## Description

get_tags_in_window in the python bindings was bound to the wrong method

## Related Issue

Fixes #5005 

All python blocks that use get_tags_in_window

## Checklist

### DIFF
--- a/gnuradio-runtime/python/gnuradio/gr/bindings/block_gateway_python.cc
+++ b/gnuradio-runtime/python/gnuradio/gr/bindings/block_gateway_python.cc
@@ -67,7 +67,7 @@ void bind_block_gateway(py::module& m)
         .def(
             "get_tags_in_window",
             (std::vector<gr::tag_t>(block_gateway::*)(unsigned int, uint64_t, uint64_t)) &
-                block_gateway::_get_tags_in_range,
+                block_gateway::_get_tags_in_window,
             py::arg("which_input"),
             py::arg("rel_start"),
             py::arg("rel_end"))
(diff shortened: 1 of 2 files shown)


ryanvolz (approved):
Ah, glad it was pretty simple. Looks good!

PR #5111, opened 2021-09-29
grc: Include scale factor when computing drawing area size

### LINKED ISSUE #4174: GRC freezes when searching for components (OsX, BigSur)

ryanvolz:
Fixes #4174.

Re-mix of #4814.

### DIFF
--- a/grc/gui/DrawingArea.py
+++ b/grc/gui/DrawingArea.py
@@ -154,7 +154,11 @@ class DrawingArea(Gtk.DrawingArea):
 
     def _update_size(self):
         w, h = self._flow_graph.get_extents()[2:]
-        self.set_size_request(w * self.zoom_factor + 100, h * self.zoom_factor + 100)
+        scale_factor = self.get_scale_factor()
+        self.set_size_request(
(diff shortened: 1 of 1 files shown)


michaelld (approved):
LGTM ; thx!

michaelld:
I can't yet prove that this PR fixes anything, but comments from user(s) on the noted PR and issue note that it does address the issue for them. Hence, I'm all for getting this fix in place.

mormj:
I have no way of verifying it either - but let's go ahead and merge it on master since it should be low-risk anyhow.

PR #5120, opened 2021-09-30
grc: Fix category and module tooltips

argilo:
Category and module tooltips have been broken since GNU Radio 3.8. The `_format_cat_tooltip` function expects a tuple representing the module & category hierarchy, but a single string is passed in. As a result, everyhing is treated as a category, and only the last letter of the category or module name is displayed. I've fixed the problem by passing in the full tuple instead.

Before:
tooltips-before-1 tooltips-before-2

After:
tooltips-after-1 tooltips-after-2

Signed-off-by: Clayton Smith <argilo@gmail.com>

## Checklist


mormj (approved):
works!  thanks for the fix

PR #5123, opened 2021-09-30
qtgui: remove vestigal parameter in fft cotr

mormj:

## Description

When the fft object is instantiated in qt gui freq sink (float), there is a `true` parameter that is interpreted as nthreads, but is really left over when the object was instantiated with bool as the forward rather than templated

This shouldn't affect any functionality, but is just a minor cleanup of the instantiation of this one object to remove confusion.

## Testing done
Verified that qt gui freq sink still works


ThomasHabets (approved):
Good catch. Added in ed95e8d45c.
(un)luckily no actual difference.

PR #5127, opened 2021-10-01
grc: Add aliases for "sc16==short" and "sc8==byte"

### LINKED ISSUE #4998: grc: sc16 ports on uhd/soapy can not be connected to anything else due to type mismatch

ryanvolz:
When type checking was added for GRC connections, the valid connections between a block with the "sc16" type (e.g. UHD blocks) and a block with the "short" type and a vector length of 2 (e.g. ishort->complex) now fail. Adding these aliases fixes that as a stop-gap until a better solution can be found.

Fixes #4998. See further discussion on this subject in #5001.

### DIFF
--- a/grc/core/Constants.py
+++ b/grc/core/Constants.py
@@ -121,14 +121,18 @@ ALIAS_OF = {
     'float': 'f32',
     'int': 's32',
     'short': 's16',
+    'short': 'sc16',
     'byte': 's8',
+    'byte': 'sc8',
     'bits': 'bit',
 
     'fc32': 'complex',
     'f32': 'float',
     's32': 'int',
     's16': 'short',
+    'sc16': 'short',
     's8': 'byte',
+    'sc8': 'byte',
     'bit': 'bits',
 }
 


dkozel (approved):
I agree with the sentiment in the draft PR. Restoring previously valid behavior that is also functionally correct is worth doing.

Personally, I'd also very much like to see shorts removed from public view, I don't know why it'd be the right answer ever? Making the type names much more explicit and specifically size based makes sense to me from a clarity perspective.

mormj (approved):

mormj:
Agreed - let's merge this to restore previously working behavior.  

We *could* standardize the set of allowable names for 3.10 (definitely for 4.0) - and align with SigMF.  Would require effort on OOT maintainers but would be a 1:1 search and replace.  This would especially make sense if gr-sigmf is getting upstreamed.

willcode:
@ryanvolz I just noticed that this is a dictionary, so one one direction this won't work as expected for short and byte.

PR #5129, opened 2021-10-02
examples: Repair all examples affected by commmit 91e9ddf8e1c37.

drmpeg:
Replace all occurrences of empty minoutbuf and maxoutbuf with 0.


dkozel:
Link to https://github.com/gnuradio/gnuradio/commit/91e9ddf for convenience.

> grc: Raise error for unfilled parameters
> 
> Issue 5032 results from an unfilled parameter erroneously indicated as
> filled with a default value of 0. This PR makes grc indicate an error
> when there is an unfilled parameter by returning None.

Good that things are becoming more explicit!

mormj (approved):

PR #5147, opened 2021-10-06
iio: simplify the wrappers leading up to pluto

mormj:
## Description
Currently, the pluto objects in gr-iio hold fmcomms2_fc32 hier block which wraps an fmcomms2 block with type conversions (but consuming extra threads).

The changes here do the following:
1) Template the fmcomms2_{src/sink} objects so they can be either short or complex
2) Get rid of the hier blocks surrounding fmcomms2
2) Get rid of the pluto objects, since they are just a very thin wrapper
4) make the pluto grc just instantiate the fmcomms2 object which now outputs complex

gr-iio

## Testing Done
pluto_source currently is working as it did before this PR

I will continue along these lines if this looks like a good approach to simplify the code.  There is probably more we want to do with IIO, but this overcomes some of my main gripes (replicated code, unnecessary hier blocks)

Left to do:
- [x] propagate for fmcomms2_sink / pluto_sink
- [x] sc16 input/output option
- [ ] ~~propagate for fmcomms5~~ -- holding off on fmcomms5 for now


mormj:
I think the templating needs to be gr_complex and std::complex<int16_t>

templating to just int16_t but assuming interleaved is confusing - it combines channels

When templated as short, should just have 2 short channels, one for i, one for q - the raw fmcomms interface

templating as std::complex<int16_t> should interleave them

PR #5153, opened 2021-10-08
gr-runtime: hier_block2.py __getattr__ avoid recursive loop in exception

### LINKED ISSUE #5072: hier_block2 __getattr__ calls itself recursively if it doesn't have the _impl attribute
So uh... I was trying to debug something and I found this bug. At least I think it's a bug.

In hier_block2 __getattr__ has a check for if not hasattr(self, "_impl") which raises a runtime error, but the runtime error calls self.__class__.__name__ which calls __getattr__ which checks for if not has attr(self, "_impl") which raises a runtime error, but the runtime error calls...

Well you get the point.

MattMills:
gnuradio-runtime: hier_block2.py `__getattr__` recursive loop bug fixes #5072

Simple test code to reproduce issue:
```
from gnuradio import gr
  
class test_block(gr.top_block):
    def __init__(self):
        #gr.top_block.__init__(self, "test_block")
        print("%s" % self.name)

tb = test_block()
```

### DIFF
--- a/gnuradio-runtime/python/gnuradio/gr/hier_block2.py
+++ b/gnuradio-runtime/python/gnuradio/gr/hier_block2.py
@@ -75,11 +75,16 @@ class hier_block2(object):
         """
         Pass-through member requests to the C++ object.
         """
-        if not hasattr(self, "_impl"):
+        
+        try:
+            object.__getattribute__(self, "_impl")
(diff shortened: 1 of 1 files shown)


willcode:
This was not compatible with Python2, so I had to revert the 3.8 backport.

PR #5186, opened 2021-10-11
grc: Look up type aliases as a set instead of a single value

ryanvolz:
After #5127, we inadvertently had duplicate dictionary keys, meaning the last entry is the only one that actually existed and some desired aliases were missing. This changes the ALIAS_OF dictionary to ALIASES_OF where the values are now a set of aliases instead of a single string value. The one use of the ALIAS_OF dictionary was changed to operate on the returned set.

### DIFF
--- a/grc/core/Connection.py
+++ b/grc/core/Connection.py
@@ -8,7 +8,7 @@ SPDX-License-Identifier: GPL-2.0-or-later
 
 
 from .base import Element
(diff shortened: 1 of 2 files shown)


willcode:
I like the `set()` better so we don't add invalid data to the dict. Any reason for prefering `set()` to `{}` as the default?

You have this one or do you want me to do it?

PR #5187, opened 2021-10-12
grc: Dependent variables sometimes fail to evaluate

### LINKED ISSUE #4788: Flowgraph doesn't update disabled blocks properly if samp_rate is controlled by another variable

### LINKED ISSUE #5100: gr-digital/examples/ofdm/tx_ofdm.grc example still broken (on master branch)
Hi all,

Today I was playing with OFDM blocks and looking at the tx_ofdm flowgraph, and it looks like there is still something broken with it.  If I just open it and run it as-is, it seems to work, but the moment I try to add a block, modify a block or change any parameter whatsoever, I fall what seems to be the same bug as reported in #4937 .
See below:
image
image

### LINKED ISSUE #5108: Error when opening flowgraph with 'Modulate vector' block
See [https://www.mail-archive.com/discuss-gnuradio@gnu.org/msg73840.html](url)

I am trying  to make my own version of packet_rx.grc.  I copied the
flowgraph and began modifying it.  I found that at some point  I started
getting this error when I open the flowgraph:

ERROR:gnuradio.grc.core.FlowGraph:Failed to evaluate variable block
modulated_sync_word
  File
"/home/user/git/toolbox/conda/miniforge3/envs/gnuradio/lib/python3.8/site-packages/gnuradio/grc/core/FlowGraph.py",
line 268, in renew_namespace
AttributeError: 'NoneType' object has no attribute 'to_basic_block'
ERROR:gnuradio.grc.core.FlowGraph:Failed to evaluate variable block
modulated_sync_word
  File
"/home/user/git/toolbox/conda/miniforge3/envs/gnuradio/lib/python3.8/site-packages/gnuradio/grc/core/FlowGraph.py",
line 268, in renew_namespace
AttributeError: 'NoneType' object has no attribute 'to_basic_block'

I traced the problem back to whether or not the preamble variables are
enabled in the flowgraph.  However, it doesn't matter whether these
variables are actually used anywhere.  They simply need to exist otherwise
this error occurs.

Please see the attached grc file.  It has the minimum number of blocks to
reproduce the error.  If the 3 disabled blocks are instead enabled then the
error will not occur.

[modulate_vector_error_example.zip](https://github.com/gnuradio/gnuradio/files/7251602/modulate_vector_error_example.zip)

In the linked thread @dl1ksv already suggested a solution:
```
It's an error in grc. The block 'Modulate vector' is not proper
evaluated. Just open and close the block 'Modulate vector' in the
property editor then you get the error message :


Value "digital.modulate_vector_bc(mod.to_basic_block(), data, taps)"
cannot be evaluated:
'NoneType' object has no attribute 'to_basic_block'


I think there is an error in the digital_modulate_vector.block.yml
definition.
Probably line 16 in this file

value: ${ digital.modulate_vector_bc(mod.to_basic_block(), data, taps) }

should be removed.
```

dl1ksv:
Sometimes dependent variables fail to evaluate or are not updated if the parent variable is disabled or deleted.

# Pull Request Details

In renew_namespace of FlowGraph.py the namespace must be deleted first, otherwise the namespace may contain
entries of deleted blocks.
Loading the variable blocks calls 
variable_block.rewrite() which requires the actual namespace.
So before loading the variable blocks , the namespace must be updated

## Related Issue

Fixes #4788
Fixes #5100
Fixes #5108

## Testing Done

Tested with the examples in the mentioned issues 
## Checklist

### DIFF
--- a/grc/core/FlowGraph.py
+++ b/grc/core/FlowGraph.py
@@ -227,6 +227,10 @@ class FlowGraph(Element):
 
     def renew_namespace(self):
         namespace = {}
+        # Before renewing the namespace, clear it
+        # to get rid of entries of blocks that
+        # are no longer valid ( deleted, disabled, ...)
+        self.namespace.clear()
         # Load imports
         for expr in self.imports():
             try:
@@ -262,6 +266,9 @@ class FlowGraph(Element):
                 pass
         namespace.update(np)  # Merge param namespace
 
+        # We need the updated namespace to evaluate the variable blocks
+        # otherwise sometimes variable_block rewrite / eval fails
+        self.namespace.update(namespace)
         # Load variables
         for variable_block in self.get_variables():
             try:
@@ -275,9 +282,7 @@ class FlowGraph(Element):
                 log.exception('Failed to evaluate variable block {0}'.format(variable_block.name), exc_info=True)
                 pass
 
-        self.namespace.clear()
         self._eval_cache.clear()
-        self.namespace.update(namespace)
 
     def evaluate(self, expr, namespace=None, local_namespace=None):
         """
--- a/grc/gui/Application.py
+++ b/grc/gui/Application.py
@@ -118,7 +118,6 @@ class Application(Gtk.Application):
 
         def flow_graph_update(fg=flow_graph):
             main.vars.update_gui(fg.blocks)
-            fg.namespace.clear()
             fg.update()
 
         ##################################################


drmpeg:
@willcode This PR should be backported. It probably should have been in 3.9.4.0

willcode:
@drmpeg On the list and I'll put it on maint-x soon ... it didn't get merged into master in time to get into the release, so Dec unless it's a complete blocker for a lot of people.

drmpeg:
Okay. I believe it only affects some specific flow graphs, so 3.9.4.0 should be okay.

PR #5188, opened 2021-10-12
grc: update test_examples qa test

mormj:

# Pull Request Details
The test_examples pytest relies on a proper path to the yaml of all blocks in the build tree.  For uhd this was off by a level

## Testing Done
test_examples.py now passes the uhd tests

Partially addresses https://github.com/gnuradio/gnuradio/issues/3534

- [x] update path for UHD blocks
- [x] reenable the test in cmake
- [x] add digital tests (manually adding the subdirs, nothing too smart
- [x] blacklist the trellis examples because they never got updated since 3.7

Right now I'm seeing 2 issues
1) Bus connections are hosed
2) Flowgraphs with Hier blocks - there is an order.  Perhaps we need to ship the hier blocks in-tree rather than leave it to the user to generate them


willcode:
@mormj Good for 3.9?

mormj:
I have not been able to get this to run. Closing for now.

drmpeg:
Maybe merge ```grc/tests/test_examples.py``` only? Then at least be can be run manually.

mormj:
good point - let's do that

drmpeg:
Looks good. I don't think Fedora likes ```apt install -qy libcodec2-dev libgsm1-dev``` though.

I'm seeing some new failures with ```packet_utils```. I'll try to bisect those.

drmpeg:
I've bisected the new ```packet_utils``` errors to commit https://github.com/gnuradio/gnuradio/commit/3cecf9268b15bb0e9e2fedaeb2528bd3038beee6

Unfortunately, that's a big commit and doesn't narrow things down much.

drmpeg:
Found it. Putting back ```from . import packet_utils``` into ```gr-digital/python/digital/__init__.py``` fixes it.

mormj:
Ooh - yeah that pep8 formatting was a bit aggressive and had some negative consequences that were hard to catch in the normal CI.  Good catch!

PR #5192, opened 2021-10-13
grc: Removed unused code in gnuradio/platform.py 

furiousluck:
grc: Removed unused code in gnuradio/platform.py


furiousluck:
@mormj please check if it is correct

PR #5208, opened 2021-10-15
pmt: Use VOLK allocator for PMT vectors

### LINKED ISSUE #3393: filter: Out of bound memory access in the fir_filter.cc filter() functions
It is possible for out of bound memory access in the filter() kernels due to the way volk alignment is ensured.

This is an issue that was originally identified by a developer who does not work for me anymore, but here's an attempt to describe it:

In `gr-filter/lib/fir_filter.cc ` there are multiple filter() functions that do a single inner product using a volk kernel. They are used by exposed API functions filterN() and filterNDec(). This inner product accesses memory at lower address spaces than the provided pointer. The taps at these locations are zero in the aligned taps used, so it is typically not a problem but unexpected data can be returned in the beginning of the output vector. We have also observed similar behavior at the end of the output vector. We have compensated by zero padding our data but this creates some obnoxious logic and i think a fix belongs in-tree.

Below is a code excerpt that has been commented to describe whats happening.

```c++
template <>
float fir_filter<float, float, float>::filter(const float input[])
{   
      // d_align is the bytes alignment (usually 32)
      // d_naligned is the <type> alignment (32/4 = 8 for float)

      // this is the dangerous pointer logic
      const float* ar = (float*)((size_t)input & ~(d_align - 1));
      // if input was 0x1234 and d_align was 32, mask is ~0x1f so ar will now be 0x1220
      unsigned al = input - ar;
      // in this pointer example, al would be 5 (or 20 bytes == 0x14 bytes)
      volk_32f_x2_dot_prod_32f_a(d_output, ar, d_aligned_taps[al], d_ntaps + al);
      return *d_output;
      
      // the volk kernel has this signature:
      //   static inline void volk_32f_x2_dot_prod_32f_generic(float* result,
      //                                                       const float* input,
      //                                                       const float* taps,
      //                                                       unsigned int num_points)
      //
      // the kernel will access both *input and *taps from 0 to num_points
      
      
      // input is given the pointer ar
      // taps is d_aligned_taps[i] which is zero padded to the correct start location for the data alignment
      // d_aligned_taps[i] will have (for 32-bit alignment and 9 taps):
      // [0000 000t tttt tttt] or [tttt tttt t000 0000] and any shift inbetween, where t is a float tap and 0 is a float 0.0
      //
      // at the ar* will be data that has up to 7 floats before the data
      // xxxx xxxd dddd dddd... or dddd dddd d...   where d is a data float and x is NOT OUR DATA
      //
      // This is not necessarily a mathematical problem because x*0 = 0, but
      // the c++ standard doesn't guarantee this to be true for things like NAN
      // https://stackoverflow.com/questions/30242196/is-floating-point-multiplication-by-zero-guaranteed-to-produce-zero
      //
      // Even more critical is this could cause a segfault because again, ITS NOT OUR MEMORY
}
```

Here is an example application that will return unexpected results (NAN's in this case) on at least some compilers, its difficult to know if this will be an issue everywhere but it definitely can be.:

example.cpp:
```c++
#include <gnuradio/filter/fir_filter.h>
#include <iostream>
#include <vector>

#define NSAMPLES 10

using namespace gr;

int main (int argc, char **argv) {
    std::cout << "running nan test:" << std::endl;

    uint32_t* some_data = (uint32_t*)malloc((NSAMPLES) * sizeof(uint32_t));
    for (size_t i=0; i<NSAMPLES; i++) some_data[i] = 0xffffffff;

    float* in = (float*)malloc((NSAMPLES) * sizeof(float));
    for (size_t i=0; i<NSAMPLES; i++) in[i] = 1.0;

    std::vector<float> out(NSAMPLES,0);
    std::vector<float> taps = {1, 1, 1, 1, 1};
    filter::kernel::fir_filter_fff *test_filter = new filter::kernel::fir_filter_fff(1,taps);
    test_filter->filterNdec(&out[0], in, NSAMPLES, 1);

    std::cout << "output: ["; for (size_t i=0; i<NSAMPLES; i++) std::cout << out[i] << " "; std::cout << "]" << std::endl;
    delete(test_filter);
}
```
build with the following CMakeLists.txt file (assuming you are using pybombs):
```
cmake_minimum_required(VERSION 3.8)
include_directories( $ENV{PYBOMBS_PREFIX}/include )
add_executable(example example.cpp)
target_link_libraries(example gnuradio-filter)
```

output is
```
jacob@ubuntu:~/fir_filter_issue/build$ ./example 
running nan test:
output: [-nan -nan -nan -nan 5 5 4 3 -nan -nan ]
```

schneider42:
This provides deterministic and VOLK compatible alignment of PMT
vectors.

Two reasons why to do this:
 - Some VOLK floating point kernels provide slightly different results if the alignment of the input vectors changes. The differences are small and can usually be neglected, but they make debugging and testing a bit harder.
 - This provides a workaround for #3393 when using `filterNDec()` or `filterN()` together with PDUs. It's obviously not a full solution for that issue.

I've only tested this with `pmt_c32vector` but I see no reason why it shouldn't work with the other vectors as well.

@jacobagilbert and I consider this worth a discussion. It obviously adds a VOLK dependency to `pmt` and I'm not sure what exactly the future of that part of GNURadio is.

Edit: Previously I stated that this only fixes #3393 in combination with PDUs in certain cases.

### DIFF
--- a/gnuradio-runtime/lib/pmt/CMakeLists.txt
+++ b/gnuradio-runtime/lib/pmt/CMakeLists.txt
@@ -17,6 +17,7 @@ target_link_libraries(gnuradio-pmt
   Boost::boost
   Boost::thread
   Log4Cpp::log4cpp
+  Volk::volk
 )
 
 target_include_directories(gnuradio-pmt
--- a/gnuradio-runtime/lib/pmt/pmt_unv_int.h
+++ b/gnuradio-runtime/lib/pmt/pmt_unv_int.h
@@ -13,6 +13,8 @@
 
 #include "pmt_int.h"
 
+#include <volk/volk_alloc.hh>
+
 #include <cstdint>
 #include <vector>
 
@@ -22,7 +24,7 @@ namespace pmt {
 ////////////////////////////////////////////////////////////////////////////
 class PMT_API pmt_u8vector : public pmt_uniform_vector
 {
-    std::vector<uint8_t> d_v;
+    volk::vector<uint8_t> d_v;
 
 public:
     pmt_u8vector(size_t k, uint8_t fill);
@@ -47,7 +49,7 @@ public:
 
 class pmt_s8vector : public pmt_uniform_vector
 {
-    std::vector<int8_t> d_v;
+    volk::vector<int8_t> d_v;
 
 public:
     pmt_s8vector(size_t k, int8_t fill);
@@ -72,7 +74,7 @@ public:
 
 class pmt_u16vector : public pmt_uniform_vector
 {
-    std::vector<uint16_t> d_v;
+    volk::vector<uint16_t> d_v;
 
 public:
     pmt_u16vector(size_t k, uint16_t fill);
@@ -97,7 +99,7 @@ public:
 
 class pmt_s16vector : public pmt_uniform_vector
 {
-    std::vector<int16_t> d_v;
+    volk::vector<int16_t> d_v;
 
 public:
     pmt_s16vector(size_t k, int16_t fill);
@@ -122,7 +124,7 @@ public:
 
 class pmt_u32vector : public pmt_uniform_vector
 {
-    std::vector<uint32_t> d_v;
+    volk::vector<uint32_t> d_v;
 
 public:
     pmt_u32vector(size_t k, uint32_t fill);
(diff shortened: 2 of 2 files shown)


schneider42:
I need to get QA tests to compile/link

PR #5220, opened 2021-10-20
grc: enable debug logging

### LINKED ISSUE #4628: grc: enable debug log
grc has the comman line option --log, which can be set to debug. Doing so enables the debug log in main.py but not in the started Application.py.

For debugging the output of Application.py would be more important.

How can I enable this log on the command line?

mpentler:
Partially fixes #4628 in grc only. The call to getLogger was missing the
grc. prefix, so I have added this in using an f-string to each call to
evaluate the module name.

Signed-off-by: Mark Pentler <xxx>

## Description
As above. It worked in the grc main.py, but nowhere else.

## Related Issue
Partially fixes #4628 but for grc only - there are other similar calls in gr-utils which may need similar work?

Just the logging output to console from grc (in the sense that it now works)

## Testing Done
Enabled logging and confirmed each individual source file was successfully outputting statements to the console. I really don't know much about unit testing etc and not sure how to do anything like that, so this was just manual testing.

## Checklist

### DIFF
--- a/grc/converter/main.py
+++ b/grc/converter/main.py
@@ -13,7 +13,7 @@ import os
 from . import block_tree, block
 
 path = os.path
-logger = logging.getLogger(__name__)
+logger = logging.getLogger(f"grc.{__name__}")
 
 excludes = [
     'qtgui_',
--- a/grc/core/FlowGraph.py
+++ b/grc/core/FlowGraph.py
@@ -19,7 +19,7 @@ from .base import Element
 from .utils import expr_utils
 from .utils.backports import shlex
 
-log = logging.getLogger(__name__)
+log = logging.getLogger(f"grc.{__name__}")
 
 
 class FlowGraph(Element):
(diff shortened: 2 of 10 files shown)


sdh11:
@mpentler: Thanks for creating the PR! I thought about this some more, and while these changes solve the issue, I think there is a better solution. The issue is the module's `__name__` prefix changes based on how grc is launched. 'grc' is the prefix if launching directly, but it is 'gnuradio.grc' if launched with the `gnuradio_companion` command. 

We can set the logger name dynamically based on the prefix seen by `main()`. That only requires changes to main.py rather than every file calling logging and doesn't result in prefixes like 'grc.gnuradio.grc.gui.Application'. 

I've pushed changes here: https://github.com/sdh11/gnuradio/tree/4628-debug-logging

mpentler:
Just to check did @sdh11 end up fixing this in the end? Pruning my branches…

mormj:
Not that I am aware of.  The fix suggested above I don't believe was made into another PR, so it only lives in the comments of this PR

sdh11:
Opened a new PR with the fixes.

PR #5227, opened 2021-10-21
gr-modtool: Don't copy source permissions

doronbehar:
# Pull Request Details

This is needed for systems such as NixOS, where the build tree isn't writable and the files copied should be. 

## Description

See the difference between `shutil.copy` and `shutil.copy2` (the default `copy_function`).

https://docs.python.org/3/library/shutil.html#shutil.copytree

## Related Issue

https://github.com/NixOS/nixpkgs/issues/142428

None probably.

## Testing Done

Tested that `gr_modtool newmod` works with this patch on NixOS.

## Checklist

### DIFF
--- a/gr-utils/modtool/core/newmod.py
+++ b/gr-utils/modtool/core/newmod.py
@@ -11,6 +11,7 @@
 
 import shutil
 import os
+import stat
 import re
 import logging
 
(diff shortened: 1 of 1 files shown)


doronbehar:
> I understand the design aspect, but that means NixOS is taking away file system features that work well and that we rely on, for us to then reimplement and maintain something less general within our own code base. Wouldn't the right solution be to tell `nix-daemon` to exclude the gr-newmod directory?

Nix will never do that. What other file system features does GR rely on? Even on a non-NixOS system, what is the reason you'd want to keep the permissions of the source directory? They have no meaning. The default permissions of the user creating the destination directory should be used.

marcusmueller:
> Nix will never do that. What other file system features does GR rely on? 

I don't think a maintainer would want to start worrying about the file permissions set during the installation process being changed afterwards. So, the FS feature would be "persistent permissions".

willcode:
Looked at this closer - no monkey-patching a Python library.

doronbehar:
> So, how do we proceed from here? I'm not the maintainer, but I don't think they'd want to merge something that is nixos-specific and Stackoverflow describes it as "hotfix, not recommended in the general case," which also monkeypatches shutil with unforeseeable side effect bugs in the future, but I do want to help GNU Radio install and run well on NixOS.
> 
> Isn't there a more general solution from the NixOS world, where we can keep the permissions as installed somehow and reconstruct on copying? Like a map for setting `umask` before copying each file, based on what the permissions / attributes used to be before nix-daemon replaced them?

Unfortunately, there is no way we can affect the way Nix makes the files readonly. Worse case we'll maintain this patch on our own, or, we can copy the code of `shutil.copytree` while removing conditionals and other logic not needed.

mormj:
Going to go ahead and close this one as we are pushing closer to 3.10 release and not ready to make this type of change that would have a lot of impact across the board.  I do appreciate you bringing this forward and educating us on the file system feature philosophy.  It should perhaps be revisited again after 3.10 release.  Thanks!

jpsimas:
This same issue also happens when using Guix System.

PR #5234, opened 2021-10-21
Soapy: Make sure target uses expected C++ standard

marcusmueller:
This is a guess, but it seems SoapySDR was exporting its C++ standard (C++11).

This led to problems when using C++17 features in gr-soapy

# Pull Request Details
As discussed in #5218 : SoapySDR seems to export its C++ standard (currently: C++11) to using libraries. 
This ensures gr-soapy can use C++14.

None as of now, but future-proofing things

## Testing Done

None, hoping CI runs through

## Checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -135,6 +135,8 @@ set(CMAKE_CXX_EXTENSIONS OFF)
 set(CMAKE_CXX_STANDARD 14)
 set(CMAKE_C_STANDARD 11)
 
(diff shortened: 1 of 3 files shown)


willcode:
It's a shame there's not a better way. If we add a dep to some other module, we'll have to remember the same trick.

marcusmueller:
@willcode tried and failed: even PUBLIC defining that gnuradio-runtime needs C++17 as compile_feature gets overridden. We'll very much have to remember to do that next time. :(

Issue #3006, opened 2020-01-04
Make GR OOT top-level GR components equivalent


devnulling:
Place holder for 3.9 project

- Not in GR proper; still OOT
- the min-in-tree route requires a robust OOT setup
- installing OOT Python into “${GR_PYTHON_DIR}/gnuradio/OOT”

For examples on how to find or use:
- Python: from gnuradio import blocks, fosphor, iqbal
- CMake: find_package(gnuradio COMPONENTS blocks fosphor iqbal)
- Have to update modtool
- Has other benefits for where Python files are installed, library names, etc.

mormj:
The bulk of these changes were implemented in 3.10 release

PR #5265, opened 2021-10-26
modtool: move OOTs equal to components for 3.10

mormj:

# Pull Request Details

## Description
It is desired to make OOT modules equivalent to in-tree modules, both from a packaging perspective, and semantically by ensuring that GR OOT modules are associated with gnuradio in naming convention, package import, etc.

This PR updates modtool for 3.10 to do the following
1. `include/gnuradio/howto`
2. `python/gnuradio/howto`
3. Rename the cmake target gnuradio-howto

The usage of OOTs then will slightly change to be
1. from gnuradio import myoot
2. #include <gnuradio/myoot/myblock.h>

When including an OOT in another OOT, right now this should look like:
`find_package(gnuradio-myoot)`
but going forward it would be desirable to have
`find_package(gnuradio COMPONENTS runtime blocks myoot)`
-- just not sure how to do this yet

## Related Issue
Partially addresses #3006 

Modtool - so all OOTs past and present (back to 3.7).  When creating a new module it will be 3.10 modtool compatible (can't use an old prefix to update the blocks), and older OOTs created with 3.9, 3.8, 3.7 should function the same and be updated with this modtool accordingly

## Testing Done
So far, have only created new OOTs

## Checklist


777arc:
> `find_package(gnuradio COMPONENTS runtime blocks myoot)`
> -- just not sure how to do this yet

Any progress with this one?

mormj:
Nope I have no idea how to do the components part. I think gnuradio-oot will suffice unless someone adept at cmake can figure it out.

mormj:
This should be at a point now where we can have gnuradio installation and oot installation in different prefixes, as long as LD_LIBRARY_PATH and PYTHONPATH include both

Here is the testing that I have done:
1) install gnuradio in /usr/
2) install oot in /usr/local
3) python import from both gnuradio and oot
4) c++ include headers and link to both gnuradio and oot using find_package(gnuradio-myoot)

So I think this should be ready to go at this point pending review

777arc:
When trying this out I got the error `Unknown CMake command "configure_package_config_file"`, I've got cmake version 3.16.3, does the new scheme require a newer version?

mormj:
Hmm - not sure what the dependency was there.  I have cmake 3.16.3 as well.  Must be something in my cmake module path - but I added the include and it doesn't cause any issues here.  Thanks for catching that.

mormj:
Added some changes found from people on chat who had tried it out.  Going ahead and merging this to expedite the RC.

PR #5270, opened 2021-10-27
Convert to modern logging

marcusmueller:
# Pull Request Details
## Description

Since #4981 has been implemented, we'll need to convert to more modern logging for two reasons:

1. to get rid of the high-cost boost::format that's still in tree
2. to get rid of logging macros (which make debugging harder and don't optimize away better than logging templated functions, unless you call anything expensive with side effects in the arguments to logging, which you basically never have to do – since formatting can be done on the fly only if enabled)
3. having references for "how to log" in-tree

## Related Issue

depends on #4981; merged / in 3.10.0.0

Whole tree.

## Testing Done

Fixed all unit tests.

There's been a unit test change in `qa_keep_m_in_n`, since that checks the verbatim string from an exception, and the strings weren't very good in the original code. Not sure whether we consider exception `what()` string API, though.

## Checklist

- [x] I have updated [the documentation](https://wiki.gnuradio.org/index.php/Main_Page#Documentation) where necessary: https://wiki.gnuradio.org/index.php?title=GNU_Radio_3.10_OOT_Module_Porting_Guide#Logging

### DIFF
--- a/gnuradio-runtime/include/gnuradio/rpcserver_thrift.h
+++ b/gnuradio-runtime/include/gnuradio/rpcserver_thrift.h
@@ -16,10 +16,8 @@
 #include <gnuradio/logger.h>
 #include <gnuradio/rpcpmtconverters_thrift.h>
 #include <gnuradio/rpcserver_base.h>
-#include <boost/format.hpp>
 #include <boost/thread/mutex.hpp>
 #include <map>
-#include <sstream>
 #include <string>
 
 #define S(x) #x
--- a/gnuradio-runtime/lib/buffer.cc
+++ b/gnuradio-runtime/lib/buffer.cc
@@ -20,7 +20,6 @@
 #include <gnuradio/buffer_type.h>
 #include <gnuradio/integer_math.h>
 #include <gnuradio/math.h>
-#include <boost/format.hpp>
 #include <algorithm>
 #include <cassert>
 #include <stdexcept>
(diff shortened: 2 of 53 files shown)


marcusmueller:
Note that this PR does *not* try to be comprehensive (in completely eradicating `boost::format`). It's aim is really to remove boost::format where it's high cost, e.g. is the last thing making a module depend on boost, or (in a few cases, e.g. message debug) where it's an especially good example for how you can write better logging.

marcusmueller:
seems like the conda recipe is lacking fmt; sadly, can't convince conda-smithy to work on my machine for some reason...

marcusmueller:
@ryanvolz had to add fmt explicitily to the conda recipe, otherwise conda wouldn't build, but not at CMake time, but at compile time, as it couldn't find fmt/core.h where included. This might not be a good thing.

marcusmueller:
aaaah I just remembered the "do we use the spdlog-internal or external fmt" discussion. Hm. I was expecting spdlog in the INTERNAL configuration to have include path set so that the internal `fmt/core.h` would be found, but I guess that would be wrong.

Realizing this: since spdlog has to include this stuff itself, best we can do is probably to simply remove the `#include <fmt/..` altogether and rely on transitive inclusion of the right thing through spdlog.

ryanvolz (approved):

mormj (approved):
LGTM!  Just going to coordinate with @willcode on when to merge this as it will make backporting to 3.8/3.9 slightly more challenging.

Issue #2665, opened 2019-07-24
`check_topology` and `forecast` should be non-state-modifying


marcusmueller:
i.e. they need to get the `const` (as in `void block::forecast() const`) qualifier, so that people can't use them to abuse them and assume that the last call to `check_topology` defines the number of in- and output streams that will be used. This should be queryable at `start()` and `work()` time, maybe?

Same goes for `forecast`.

Motivation: Make it clearer where block state is supposed to be modifiable and where not. This is essentially crucial if we want to allow for scheduling to detach the block executing thread / worker from the managing thread/worker.

marcusmueller:
There might simply be more methods of `block` that should be considered here:

- [ ] `check_topology`
- [ ] `forecast`
- [ ] `nitems_read` / `nitems_written`
- [ ] `get_tags_in_range`

michaelld:
Love the idea (again) ... wondering how much breakage we'll have to fix by tweaking methods that currently not `const` but should be...? Guessing there will be some breakage ... especially If we change `check_topology`: We'll break some GR IT (in-tree) and OOT blocks that use it to set the actual topology ... I think `gr-audio` does this, as do some others. We need to find a better way to setup initial topology!

marcusmueller:
- `check_topology` seems to simply need to get documented that it's being used for actual modifications, not checking
- `forecast` should be `const` but there's a solid chance that breaks the ecosystem
- `nitems_read,written` should be `const` (and is not a significant API difference)
- `get_tags_in_range` needs more thought

PR #5274, opened 2021-10-28
runtime: make forecast const

mormj:
# Pull Request Details
---
## Description
This make forecast const so that it cannot be used to improperly to modify state
Does not address `check_topology` as this is used already in-tree to modify state, so will require more thought

## Related Issue
#2665

All general blocks in-tree and in every OOT

This makes me think it shouldn't be considered until after 3.10 because it will cause breaking changes in OOTs with general blocks.  Certainly if this is introduced it should be along with other useful API breaking changes, otherwise it will just be an annoyance

## Testing Done
Just CI
- [ ] Need to test against some OOTs

## Checklist

### DIFF
--- a/gnuradio-runtime/include/gnuradio/block.h
+++ b/gnuradio-runtime/include/gnuradio/block.h
@@ -151,7 +151,7 @@ public:
      * number of data items required on each input stream.  The
      * estimate doesn't have to be exact, but should be close.
      */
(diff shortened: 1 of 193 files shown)


mormj:
Closing - but added to Release 3.11 Project - if we don't break 3.10 too bad - we might be able to lump more breaking changes in 3.11

PR #5278, opened 2021-10-28
Bump Major Dependency Versions for 3.10

### LINKED ISSUE #2719: Identify desirable dependency version increases

mormj:
# Pull Request Details
---
## Description
For the 3.10 release we will support the latest LTS releases from major distributions.  The major bottleneck here in terms of dependencies is Ubuntu 3.10.

## Related Issue
Fixes #2719

top level CMake

## Testing Done
None

## Checklist

--> Needs to be updated before merging

### DIFF
--- a/.github/workflows/make-test.yml
+++ b/.github/workflows/make-test.yml
@@ -66,13 +66,13 @@ jobs:
             cxxflags: -Werror -Wno-error=invalid-pch -Wno-error=deprecated-declarations
             ctest_args: '-E "qa_cpp_py_binding|qa_cpp_py_binding_set|qa_ctrlport_probes|qa_qtgui"'
             ldpath: /usr/local/lib64/
-          - distro: 'CentOS 8.3'
-            containerid: 'gnuradio/ci:centos-8.3-3.9'
+          - distro: 'CentOS 8.4'
+            containerid: 'gnuradio/ci:centos-8.4-3.10'
             cxxflags: ''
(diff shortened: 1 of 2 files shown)


marcusmueller (approved):
These seem to be very reasonable minimum versions.

mormj:
Debian 10 fails - but this should fall out of support for 3.10 as Debian 11 is released.
CentOS 8 fails because of Boost - but also is going EOL 12/31/2021 so we probably don't need to track it.

marcusmueller:
> CentOS 8 fails because of Boost - but also is going EOL 12/31/2021 so we probably don't need to track it.

If EPEL is enabled, `yum install boost169-devel` should fix that, right?

marcusmueller:
Propose to replace 

```yaml
- distro: 'CentOS 8.3'
  containerid: 'gnuradio/ci:centos-8.3-3.9'
```

with 

```yaml
- distro: 'CentOS 8.4'
  containerid: 'gnuradio/ci:centos-8.4-3.10'
```

in .github/workflows/make-test.yml if we merge https://github.com/gnuradio/gnuradio-docker/pull/23 .

mormj:
There is still an issue with Centos 8.4, but will clean that up later.  We can ignore CentOS 8.4 for now

PR #5279, opened 2021-10-28
modtool,cmake: make module dir for tests 

### LINKED ISSUE #3999: python: QA code module import order is confusing
In my opinion, the expected behavior of the `import MODNAME` statement in QA code should be to first attempt to import the module from the `./python` directory if it exists (e.g.: running `make test`) and then attempt to import the installed version on a system. This is mostly for development purposes, currently when you build and test a module if you have a module already installed it will run your QA against the installed version.

Rather than adding to the import logic in QA, what about having the build process create the entire python module in the build/python directory and that path be added to the beginning of the python path for the scripts that run the QA so it is imported first. 

Thoughts? I've tested this and it works, and seems a little more convenient for running QA at build time.

### LINKED ISSUE #4825: OOT QA tests fail unless module installed
This issue has been discussed in a [thread on the mailing list](https://lists.gnu.org/archive/html/discuss-gnuradio/2021-06/msg00106.html).

To summarize, when creating an OOT module, the QA tests don't work correctly until we run `sudo make install`. This seems backwards because we should be able to run QA tests before we decide to install the module.

I have set up a [repo that illustrates the problem](https://github.com/antonott/oot_tests_failing). I originally discovered the issue when trying to work through the [OOT tutorial](https://wiki.gnuradio.org/index.php/OutOfTreeModules), but in the repo I instead use the module [gr-iqlevels](https://github.com/drmpeg/gr-iqlevels), to reduce the likelihood that the issue is just due to a mistake by me.

As Vasil points out in the mailing list, the issue seems to come from the migration from swig to pybind. See his analysis [here](https://lists.gnu.org/archive/html/discuss-gnuradio/2021-06/msg00119.html).

I am running Ubuntu 20.04.2 LTS, and installed GNU Radio 3.9.2.0 via PPA. I have the 2.5.0 version of pybind11-dev from the PPA.

ckoehler:
# Pull Request Details

## Description

Updated:

This change creates a module directory inside the build tree (`build/test_modules/modulename`). It consists of the build artifacts from `build/python/bindings` and `python/*`. `GrTest` then adds `build/test_modules` to the PYTHONPATH.

Old description:
This change makes CTest look for locally-built modules (i.e. during development, in the build directory) before installed ones.

Also adjust test module name in the template, though that's not ideal. I _think_ we could remove the `_python` suffix from the module in this line, but not quite sure about that: https://github.com/gnuradio/gnuradio/blob/master/cmake/Modules/GrPybind.cmake#L9

If we can, existing modules should just work with the GrTest.cmake (and GrPybind.cmake) changes (i.e. a new GR version).

## Related Issue
I think it takes care of these issues:

Fixes #3999 
Fixes #4825

One line change in `GrTest.cmake` to prepend the bindings dir to PYTHONPATH for tests; `CMakeLists.txt` in `python/` changed to copy the module directory into a separate directory that can be imported.

## Testing Done

Successfully ran `make test` with these changes in place and saw passing tests.

## Checklist

### DIFF
--- a/cmake/Modules/GrTest.cmake
+++ b/cmake/Modules/GrTest.cmake
@@ -61,6 +61,10 @@ function(GR_ADD_TEST test_name)
     # Keep the original path conversion for pypath - the above commented line breaks CI tests
     file(TO_NATIVE_PATH "${GR_TEST_PYTHON_DIRS}" pypath) #ok to use on dir list?
 
+    # add test module directory to PYTHONPATH to allow CTest to find QA test modules.
+    # We add it to the beginning of the list to use locally-built modules before installed ones.
+    list(INSERT pypath 0 "${CMAKE_BINARY_DIR}/test_modules")
+
     set(environs "VOLK_GENERIC=1" "GR_DONT_LOAD_PREFS=1" "srcdir=${srcdir}"
         "GR_CONF_CONTROLPORT_ON=False")
     list(APPEND environs ${GR_TEST_ENVIRONS})
--- a/gr-utils/modtool/templates/gr-newmod/python/CMakeLists.txt
+++ b/gr-utils/modtool/templates/gr-newmod/python/CMakeLists.txt
@@ -31,3 +31,14 @@ GR_PYTHON_INSTALL(
 include(GrTest)
 
 set(GR_TEST_TARGET_DEPS gnuradio-howto)
+
+# Create a module directory that tests can import. It includes everything
+# from `python/` and the built bindings shared lib.
+add_custom_target(
+  copy_module_for_tests ALL
+  COMMAND ${CMAKE_COMMAND} -E copy_directory ${CMAKE_CURRENT_SOURCE_DIR}
+          ${CMAKE_BINARY_DIR}/test_modules/howto/
+  COMMAND
+    ${CMAKE_COMMAND} -E copy_directory ${CMAKE_CURRENT_BINARY_DIR}/bindings/
+    ${CMAKE_BINARY_DIR}/test_modules/howto/
+  DEPENDS howto_python)


ckoehler:
> The try/catch could be used to find the installed python package, but i thing the expected behavior would be to fail if the built package fails import.

Agreed, but not sure what that'd look like. Assuming all the paths are correctly set up, catching a `ModuleNotFoundError` may do the trick, since that'd mean there is no test to run from `build`, so it falls back on the installed one. Does that sound right?

> One other way to do this would be to create an entire module directory within the `<build/dir>/python` directory and have the bindings and python files copied into that folder as well during build. The python path generated for the QA scripts is then `<build/dir>/python` and the QA imports, regardless of whether it is a python or c++ block, can be the same as standard python imports for the module.

If I understand correctly, that'd allow the name to stay the same (without the `_python` suffix on the module), right? I'm not sure how to make CMake generate the directory structure for a Python module though...

ckoehler:
I played around a bit and think I know what needs to be done:

1. `python/__init__.py` needs to be copied to the `build/python/bindings/` dir (and properly import `modulename_python` instead of `modulename_swig`)
2. `build/python/bindings/` needs to be copied somewhere else as (e.g.) `build/something/modulename`
3. `build/something` needs to be added to PYTHONPATH

I think then existing QA scripts should just work when they `import modulename`. Thoughts?

ckoehler:
Alright, I just force pushed another commit that does what I said and updated the description. Thoughts?

ckoehler:
Changed `PREPEND` to `INSERT` in `CMakeLists.txt`, since `PREPEND` is relatively new and not available in Debian 10's CMake version.

ckoehler:
Hm, we could actually achieve this without touching GrTest by setting `GR_TEST_PYTHON_DIRS` in CMakeLists.txt:

`set(GR_TEST_PYTHON_DIRS ${CMAKE_BINARY_DIR}/test_modules)`

Thoughts?

ckoehler:
Sorry for so many changes. I moved the CMake specific stuff up one level and changed it to copy everything from `python/` into the module dir. Otherwise it would die if there were Python sources, since they would be missing.

samwhiting-snl:
This latest method seems like the right way to do this. A complete python package directory in the `build` directory is what we should be using to run test code since that is the same structure that will exist after install.

ckoehler:
Rebased on latest master. Can we get this merged? Anything holding it up? Any more feedback?

samwhiting-snl:
I'm in favor of merging this, for whatever that's worth. Building the entire python package directory in `build` makes sense since that's how it will exist after install. 

A note for people maintaining OOT modules: to fix QA after this merge you would need to change the QA import statements from the `try` statement mess to just `import <modulename>` as well as changing the `python/Cmakelists.txt` file by adding the test module `copy_module_for_test` custom target.

ckoehler:
What else is needed before merging this? We've been porting our modules to 3.9 with this change in place and it's worked well.

mormj:
I'd like to get a review or two on this before merging.  It is usually helpful to bring attention to the PR on chat.gnuradio.org - #development room

mormj (approved):
This does indeed allow make test to pass for OOTs - which is great!

mormj:
@ckoehler - going ahead and merging this - can you please update the wiki to reflect the new change.  Again this is a great change for OOT modules and thank you for submitting upstream.

ckoehler:
> @ckoehler - going ahead and merging this - can you please update the wiki to reflect the new change. Again this is a great change for OOT modules and thank you for submitting upstream.

I'd be happy to when I get to work later. Could you point me to the best place for that? I am unfamiliar with the wiki setup. Maybe the porting guide for 3.9 (https://wiki.gnuradio.org/index.php/GNU_Radio_3.9_OOT_Module_Porting_Guide)? Or a new page?

777arc:
Yeah that's fine with me, we link to the porting guide on our main page so folks will see it

ckoehler:
Sounds good! I requested an account on the wiki.

777arc:
> Sounds good! I requested an account on the wiki.

Approved!

ckoehler:
Thanks, I documented it [here](https://wiki.gnuradio.org/index.php/GNU_Radio_3.9_OOT_Module_Porting_Guide#CMakeLists.txt_changes_to_fix_OOT_module_testing). Feel free to move it or let me know if clarifications are needed.

mormj:
I think there might be an issue with the cmake code (in python/) when the oot module only has python blocks

samwhiting-snl:
Good catch. We should probably wrap the `<modulename>_python` portion in an if statement:

```cmake
 35 # Create a module directory that tests can import. It includes everything
 36 # from `python/` and the built bindings shared lib.
 37 add_custom_target(
 38   copy_module_for_tests ALL
 39   COMMAND ${CMAKE_COMMAND} -E copy_directory ${CMAKE_CURRENT_SOURCE_DIR}
 40           ${CMAKE_BINARY_DIR}/test_modules/pyonly/)
 41 if(pyonly_sources)
 42 add_custom_target(
 43   copy_bindings_for_tests ALL
 44   COMMAND
 45     ${CMAKE_COMMAND} -E copy_directory ${CMAKE_CURRENT_BINARY_DIR}/bindings/
 46     ${CMAKE_BINARY_DIR}/test_modules/pyonly/
 47   DEPENDS pyonly_python)
 48 endif(pyonly_sources)

```

ckoehler:
I'd prefer `if (TARGET pyonly_python)`

samwhiting-snl:
How about we just move that custom target (for bindings) into the `bindings/Cmakelists.txt` file since that is already returning early if there is no c++ source?

ckoehler:
Ah I got you. Yes. Split up the target (name it different things), add the python one to ALL, make the bindings one depend on that _python target.

Issue #5206, opened 2021-10-14
modtool: support tab completions


mormj:
### Feature Description

would be very nice to have tab completions with modtool, for example

gr_modtool <tab>
would print out 
newmod add ...

### Feature Urgency

low (just an idea)

marcusmueller:
b0912d31f1a2cfea93a5588898cae0c652cd89d4

really just need better application to everything!

AsriFox:
This seems to be relatively easy, because Click has this feature built-in ([docs](https://click.palletsprojects.com/en/latest/shell-completion/)). It works for me in bash.
Trying to figure out how to call `_GR_MODTOOL_COMPLETE=bash_source gr_modtool` from CMake.

PR #5286, opened 2021-10-31
modtool: add basic ZSH completion support

marcusmueller:
# Pull Request Details

This adds but rudimentary support for zsh completion to for `gr_modtool`, so that you can go `gr_modtool`<kbd>tab</kbd> and are shown a list of the commands, through which you can tab-cycle, or if you started to type one, can tab-complete it.

completion

This should / could of course do much more, but my guess is that we'll never get something that knows the arguments of the subcommands if we don't start somewhere.

If someone wants to pick up, the [oh-my-zsh vagrant plugin](https://github.com/ohmyzsh/ohmyzsh/blob/00ccb449907d322c6bfa53f17e96cbf48651fee3/plugins/vagrant/_vagrant) could possibly be used as example.

## Description

Just write a compsys script, drop it in fpath, start shell, and tab away.

## Related Issue

Is a stab at #5206 ; certainly doesn't close it, as it neither addresses bash (which is probably much, much more common) nor does it do sensible things like know arguments to subcommands. 

'modtool

## Testing Done

Wrote, installed, tabbed.

## Checklist

  - not quite sure if necessary.
  - hard to test interactive shell completion automatically.

### DIFF
--- a/gr-utils/modtool/CMakeLists.txt
+++ b/gr-utils/modtool/CMakeLists.txt
@@ -28,6 +28,10 @@ install(
     DESTINATION ${GR_PREFSDIR}
 )
 
+install(FILES ${CMAKE_CURRENT_SOURCE_DIR}/support/_gr_modtool
+  DESTINATION ${GR_DATA_DIR}/zsh/site-functions
+)
+
 GR_PYTHON_INSTALL(
     PROGRAMS    gr_modtool
     DESTINATION ${GR_RUNTIME_DIR}
--- a/gr-utils/modtool/support/_gr_modtool
+++ b/gr-utils/modtool/support/_gr_modtool
@@ -1,0 +1,25 @@
+#compdef gr_modtool
+#autoload
+
+# SPDX-License-Identifier: LGPL-3.0-or-later
+# Copyright 2021 Marcus Müller
+# This module offers tab-completion for ZSH
+# It's rudimentary.
+
+local -a subcmds
+subcmds=(
+    'add:Adds a block to the out-of-tree module.'
+    'bind:Generate Python bindings for GR block'
+    'disable:Disable selected block in module.'
+    'info:Return information about a given module'
+    'makeyaml:Generate YAML files for GRC block bindings.'
+    'newmod:Create new empty module, use add to add blocks.'
+    'rename:Rename a block inside a module.'
+    'rm:Remove a block from a module.'
+    'update:Update the grc bindings for a block'
+)
+_describe 'command' subcmds
+if (( CURRENT == 2 )); then
+    _describe -t commands 'subcommands' subcmds
+fi
+return 0


willcode:
Does this only work for zsh, or does it also work with bash?

marcusmueller:
@willcode I have no idea how bash completions work, but I don't think they're compatible. So, only zsh.

gr-utils/modtool/CMakeLists.txt
```diff
@@ -28,6 +28,10 @@ install(
     DESTINATION ${GR_PREFSDIR}
 )
 
+install(FILES ${CMAKE_CURRENT_SOURCE_DIR}/support/_gr_modtool
+  DESTINATION ${GR_DATA_DIR}/zsh/site-functions
```
marcusmueller, line 32: Before I do something stupid:

@maitbot is /usr/share/zsh/site-functions for ZSH completion scripts in line with debian packaging rules?

ckoehler:
I tried it on Ubuntu 20.04 and manually put the completion file into `/usr/share/zsh/site-functions` (I had to create `site-functions`), and it didn't work. Putting it instead in `/usr/local/share/zsh/site-functions` (which all existed already) worked well.

For what it's worth, here's the $FPATH after a fresh zsh install without custom config:

```sh
3784553b3693# echo $FPATH
/usr/local/share/zsh/site-functions:/usr/share/zsh/vendor-functions:/usr/share/zsh/vendor-completions:/usr/share/zsh/functions/Calendar:/usr/share/zsh/functions/Chpwd:/usr/share/zsh/functions/Completion:/usr/share/zsh/functions/Completion/AIX:/usr/share/zsh/functions/Completion/BSD:/usr/share/zsh/functions/Completion/Base:/usr/share/zsh/functions/Completion/Cygwin:/usr/share/zsh/functions/Completion/Darwin:/usr/share/zsh/functions/Completion/Debian:/usr/share/zsh/functions/Completion/Linux:/usr/share/zsh/functions/Completion/Mandriva:/usr/share/zsh/functions/Completion/Redhat:/usr/share/zsh/functions/Completion/Solaris:/usr/share/zsh/functions/Completion/Unix:/usr/share/zsh/functions/Completion/X:/usr/share/zsh/functions/Completion/Zsh:/usr/share/zsh/functions/Completion/openSUSE:/usr/share/zsh/functions/Exceptions:/usr/share/zsh/functions/MIME:/usr/share/zsh/functions/Math:/usr/share/zsh/functions/Misc:/usr/share/zsh/functions/Newuser:/usr/share/zsh/functions/Prompts:/usr/share/zsh/functions/TCP:/usr/share/zsh/functions/VCS_Info:/usr/share/zsh/functions/VCS_Info/Backends:/usr/share/zsh/functions/Zftp:/usr/share/zsh/functions/Zle
```

ckoehler (approved):

marcusmueller:
@ckoehler  interestingly, 

    comm -12 \
    =(echo "/usr/local/share/zsh/site-functions:/usr/share/zsh/vendor-functions:/usr/share/zsh/vendor-completions:/usr/share/zsh/functions/Calendar:/usr/share/zsh/functions/Chpwd:/usr/share/zsh/functions/Completion:/usr/share/zsh/functions/Completion/AIX:/usr/share/zsh/functions/Completion/BSD:/usr/share/zsh/functions/Completion/Base:/usr/share/zsh/functions/Completion/Cygwin:/usr/share/zsh/functions/Completion/Darwin:/usr/share/zsh/functions/Completion/Debian:/usr/share/zsh/functions/Completion/Linux:/usr/share/zsh/functions/Completion/Mandriva:/usr/share/zsh/functions/Completion/Redhat:/usr/share/zsh/functions/Completion/Solaris:/usr/share/zsh/functions/Completion/Unix:/usr/share/zsh/functions/Completion/X:/usr/share/zsh/functions/Completion/Zsh:/usr/share/zsh/functions/Completion/openSUSE:/usr/share/zsh/functions/Exceptions:/usr/share/zsh/functions/MIME:/usr/share/zsh/functions/Math:/usr/share/zsh/functions/Misc:/usr/share/zsh/functions/Newuser:/usr/share/zsh/functions/Prompts:/usr/share/zsh/functions/TCP:/usr/share/zsh/functions/VCS_Info:/usr/share/zsh/functions/VCS_Info/Backends:/usr/share/zsh/functions/Zftp:/usr/share/zsh/functions/Zle" | sed 's/:/\n/g' |sort) \
    =(echo $FPATH|sed 's/:/\n/g'|sort)

tells me the only common path would be `/usr/local/share/zsh/site-functions`, and we really shouldn't install in `/usr/local` if `/usr/` is the `CMAKE_INSTALL_PREFIX`, right? Sigh... more distro differences making life so much better...

marcusmueller:
Related: https://stackoverflow.com/questions/69927565/different-zsh-fpaths-under-different-linux-distros-how-to-find-the-right-at-cma

ckoehler:
Interested to hear what comes out of the SO post. One option that I've seen before is to install the completions in some custom directory and print out a caveat on install with directions on how to add that path to your $FPATH. It requires one more step, but is pretty simple on the install side.

Per https://unix.stackexchange.com/a/607810, we could follow systemd's example and install into `/usr/share/zsh/site-functions`, though they do make it configurable at build time.

Or, the hammer: install it both into `/usr/share/zsh/site-functions` (for Fedora & co) and `/usr/share/zsh/vendor-completions` (for Debian & co).

gr-utils/modtool/support/_gr_modtool
```diff
@@ -0,0 +1,25 @@
...
+    'newmod:Create new empty module, use add to add blocks.'
+    'rename:Rename a block inside a module.'
+    'rm:Remove a block from a module.'
+    'update:Update the grc bindings for a block'
+)
+_describe 'command' subcmds
+if (( CURRENT == 2 )); then
+    _describe -t commands 'subcommands' subcmds
+fi
+return 0
```
okapia, line 25: Don't unconditionally return zero. That will be taken to mean that the function has done something and there is no need to carry on and try something else like retrying with fuzzy matching, approximations, case-insensitivity or whatever the user has configured.

Usually you can just rely on the fact that a shell function passes on the return status from whatever the last command was, e.g. `_describe`. If the last thing was an if then else, or case statement, that still works correctly, passing on the status from the last command in whichever branch was taken.

But if you look at examples, it is common to accumulate a return status in a `$ret` variable.

gr-utils/modtool/support/_gr_modtool
```diff
@@ -0,0 +1,25 @@
...
+#autoload
+
+# SPDX-License-Identifier: LGPL-3.0-or-later
+# Copyright 2021 Marcus Müller
+# This module offers tab-completion for ZSH
+# It's rudimentary.
+
+local -a subcmds
+subcmds=(
+    'add:Adds a block to the out-of-tree module.'
```
okapia, line 11: Zsh convention is to prefer imperative mood form for descriptions: "add a block to the out-of-tree module". You have that for all the other descriptions ("disable" rather than "disables"). I tend to think that's good practice elsewhere - documentation and git commit messages - which you might want to consider if you copied this from documentation. It is shorter and more natural in the absence of an explicit subject for a sentence.

We also don't typically capitalize the first word or add a full-stop.

Of course, zsh conventions are really not something you're in any way obliged to follow given that this is going to the gnuradio project. We also use only two spaces for indentation in completions.

gr-utils/modtool/support/_gr_modtool
```diff
@@ -0,0 +1,25 @@
...
+    'makeyaml:Generate YAML files for GRC block bindings.'
+    'newmod:Create new empty module, use add to add blocks.'
+    'rename:Rename a block inside a module.'
+    'rm:Remove a block from a module.'
+    'update:Update the grc bindings for a block'
+)
+_describe 'command' subcmds
+if (( CURRENT == 2 )); then
+    _describe -t commands 'subcommands' subcmds
+fi
```
okapia, line 24: When `$CURRENT` is 2, i.e. the first argument to the command, this completes the same set of words twice. Once with "command" as the heading and once with "subcommands". Is that definitely the intention?

I suspect that line 21 isn't meant to be there. If that is correct. I would add an `else` condition and call `_default` for when `CURRENT != 2`. Whenever a completion is in some way not finished, it is better to call `_default` than to do nothing because user's can be unhappy if filename completion doesn't work.

gr-utils/modtool/support/_gr_modtool
```diff
@@ -0,0 +1,25 @@
+#compdef gr_modtool
+#autoload
+
```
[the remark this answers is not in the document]
marcusmueller: Note that I find this awesome - thanks for dropping by! This is super helpful.

* `#autoload`: gotcha, will be removed.
* `_gnuradio` instead of `_gr_modtool`: I must admit I have very thin understanding what you mean with `$service` here. I followed https://github.com/zsh-users/zsh-completions/blob/master/zsh-completions-howto.org#telling-zsh-which-function-to-use-for-completing-a-command ; what I understand now is that I'd start the file with `#compdef _gnuradio`, and have a `compdef _gnuradio gr_modtool gnuradio-config-info gr_plot_iq` (and so on) at the end.<br> But how does my script then know whether I'm completing for `gr_modtool` or `gnuradio-config-info`?

gr-utils/modtool/support/_gr_modtool
```diff
@@ -0,0 +1,25 @@
...
+# SPDX-License-Identifier: LGPL-3.0-or-later
```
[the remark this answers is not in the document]
marcusmueller: To give a little bit of insight: we have a subproject (VOLK) which is in the process of being re-licensed as LGPL. I was hoping to adapt this script for that later on, so I chose the license that both projects could use.

gr-utils/modtool/support/_gr_modtool
marcusmueller: Agreed, this was just me being lazy (I regexed this out of `gr_modtool --help`). These *should* be imperatives, for consistency with other tooling. I can fix this here (and upstream, too)

gr-utils/modtool/support/_gr_modtool
marcusmueller: hm, so

```shell
if (( CURRENT == 2 )); then
     _describe -t commands 'subcommands' subcmds
else
    _default
```

?

I didn't really know about `_default` (and it's only used in one file in my /usr/share/zsh/site-functions), so I'm not clear on what it does. Am I missing some docs here?

gr-utils/modtool/support/_gr_modtool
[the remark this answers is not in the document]
okapia: You start the file with `#compdef gr_modtool gnuradio-config-info gr_plot_iq`. No need to list `_gnuradio` in there, it's only with the `compdef` command that you'd say that. That would be used as the filename, though. Then in the function body you can switch on `$service`, typically:

```
    case $service in
      gr_modtool)
        ...
      ;;
      gr_plot_iq) ... ;;
    esac
```

gr-utils/modtool/support/_gr_modtool
okapia: Yes.

In the zsh documentation it is sort-of covered under "special contexts". These contexts mostly cover completion in places that aren't just normal command arguments like assignments, math, array subscripts etc. One of them is named `-default-` and handled by `_default`. In the absence of a proper completion it is what gets called. It mostly just does file completion. It's good to call it where nothing better has been implemented. Try below `/usr/share/zsh/5.8/functions/Completion/Unix` to see more examples. (Debian/Ubuntu may have moved that, perhaps taking out the version number directory).

I'll send in a zsh patch to document it better.

mormj:
@marcusmueller - is this one good to merge?  I'm not a zsh user so I'm not set up to test it.

mormj:
Closing due to lack of activity

PR #5287, opened 2021-11-01
qtgui range: setRange() requires int

### LINKED ISSUE #5341: Type error in QT GUI Range Widget
### What happened?

Using the QT GUI Range widget in either Counter, Slider or Counter+Slider mode causes a Type Error, regardless of Widget Type.

### System Information

OS: Fedora Linux 35 (Workstation Edition)
GR Installation Method: native distro packages - 3.9.2.0 (Python 3.10.0)

### GNU Radio Version

3.9 (maint-3.9)

### Specific Version

9.3.2.0

Use a QT GUI Range widget in either Counter, Slider or Counter+Slider mode.

### Relevant log output

```shell
  File "/home/ruurdjan/rangetest.py", line 199, in <module>
    main()
  File "/home/ruurdjan/rangetest.py", line 177, in main
    tb = top_block_cls()
  File "/home/ruurdjan/rangetest.py", line 87, in __init__
    self._freq_win = RangeWidget(self._freq_range, self.set_freq, 'Frequency', "counter_slider", int, QtCore.Qt.Horizontal)
  File "/usr/lib64/python3.10/site-packages/gnuradio/qtgui/range.py", line 121, in __init__
    self.d_widget = self.CounterSlider(self, self.range, self.notifyChanged, rangeType, orientation)
  File "/usr/lib64/python3.10/site-packages/gnuradio/qtgui/range.py", line 262, in __init__
    self.slider = RangeWidget.Slider(parent, ranges, self.sliderChanged, rangeType, orientation)
  File "/usr/lib64/python3.10/site-packages/gnuradio/qtgui/range.py", line 161, in __init__
    self.setRange(0, ranges.nsteps - 1)
TypeError: setRange(self, int, int): argument 2 has unexpected type 'float'
```

willcode:
Author: Chris Vine <vine35792468@gmail.com>

# Pull Request Details

## Description
Python 3.10 sees setRange( non-int-value ) as an error. Reported by Chris Vine.

## Related Issue
Fixes #5341 

qtgui/range

## Checklist

### DIFF
--- a/gr-qtgui/python/qtgui/range.py.cmakein
+++ b/gr-qtgui/python/qtgui/range.py.cmakein
@@ -131,7 +131,7 @@ class RangeWidget(QtWidgets.QWidget):
             self.rangeType = rangeType
 
             # Setup the dial
-            self.setRange(0, ranges.nsteps-1)
+            self.setRange(0, int(ranges.nsteps-1))
             self.setSingleStep(1)
             self.setNotchesVisible(True)
             self.range = ranges
@@ -158,7 +158,7 @@ class RangeWidget(QtWidgets.QWidget):
 
             # Setup the slider
             #self.setFocusPolicy(QtCore.Qt.NoFocus)
-            self.setRange(0, ranges.nsteps - 1)
+            self.setRange(0, int(ranges.nsteps - 1))
             self.setTickPosition(2)
             self.setSingleStep(1)
             self.range = ranges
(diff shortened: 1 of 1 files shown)


willcode:
There may be other places that need the same kind of thing for Python 3.10. I didn't look into why this is a problem, so it would be good for someone to get familiar with the root cause, whether this is the right solution, etc.

mormj (approved):
It doesn't seem like a bad solution, though it could be other places that we are not catching.  Maybe we should merge this to make the immediate problem go away but note in an issue that it could be other places.

willcode:
@mormj I'm wondering whether only the `slider.setValue(new)` should get a `int(round(new))`. Doing a `self.setValue(int(new))` will round down the actual value.

willcode:
Back to draft ... this should be isolated to setting the widget state, not the value itself.

willcode:
Alright, I read up on the Qt controls and it looks like this is fine.

PR #5289, opened 2021-11-01
runtime: remove unused test.cc/h

### LINKED ISSUE #5273: runtime: test.h/test.cc are not used anywhere
### What happened?

gnuradio-runtime/lib/test.{cc,h} are compiled, but not called or run anywhere.  Probably can be removed unless just compiling them was the test.

### System Information

N/A

### GNU Radio Version

3.10-git (master)

### Specific Version

0.0.0.0

Look at the codebase

### Relevant log output

_No response_

mormj:
## Description
these files appear to be unused - not public headers or used anywhere else in the codebase

## Related Issue
Fixes #5273


willcode (approved):

drmpeg:
I think ```gnuradio/gnuradio-runtime/lib/test_types.h``` could be deleted also.

PR #5293, opened 2021-11-02
runtime: mark buffer methods as override

mormj:
# Pull Request Details
found with 
```
run-clang-tidy -header-filter='.*' -checks='-*,misc-throw-by-value-catch-by-reference,misc-static-assert,readability-container-size-empty,modernize-use-override,modernize-make-unique,modernize-deprecated-headers,misc-unused-using-decls,misc-redundant-expression' -fix
```
runtime, buffers - the whole codebase

## Testing Done
CI should be sufficient

## Checklist

### DIFF
--- a/gnuradio-runtime/include/gnuradio/buffer_single_mapped.h
+++ b/gnuradio-runtime/include/gnuradio/buffer_single_mapped.h
@@ -33,7 +33,7 @@ public:
     gr::logger_ptr d_logger;
     gr::logger_ptr d_debug_logger;
 
-    virtual ~buffer_single_mapped();
+    ~buffer_single_mapped() override;
 
     /*!
      * \brief Return the block that owns this buffer.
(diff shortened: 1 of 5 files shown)


marcusmueller (approved):
Nice; clean!

gnuradio-runtime/include/gnuradio/buffer_single_mapped.h
```diff
@@ -33,7 +33,7 @@ class GR_RUNTIME_API buffer_single_mapped : public buffer
     gr::logger_ptr d_logger;
     gr::logger_ptr d_debug_logger;
 
-    virtual ~buffer_single_mapped();
```
dsorber, line 36: After reading up on this detail, it looks like `virtual` is propagated implicitly from the base class and it therefore not strictly needed here.  I guess that's a detail I'm just not used to yet.

dsorber (approved):
Looks good. I'll have to be more diligent about using `override`.

PR #5300, opened 2021-11-03
modtool: Use c++17 in template to match top level CMakeLists.txt.

drmpeg:


mormj (approved):
This is a good change regardless of std::any (but necessitated by that change).

drmpeg:
Should I change the commit message? Something like "gr_modtool template should match top-level CMakeLists.txt".

mormj:
Up to you - would probably be more accurate since std::any might get deferred.

drmpeg:
I think reverting std::any is a good idea. It's clearly more pain to users than usefulness in the tree.

mormj:
My bad on that one - premature merge before realizing all the ramifications

drmpeg:
No worries. It's pretty difficult to be prescient. Maybe some sort of quick gr_modtool test in the CI? Perhaps just newmod, add, bind. I'm not sure that's even possible.

PR #5310, opened 2021-11-04
cmake: handle optional components in `find_package`

### LINKED ISSUE #5309: Optional components with find_package not supported
### What happened?

We have OOT modules that optionally depend on UHD: if it's built into GR, great, if not, no big deal. We'd like to include those with CMake using `find_package(Gnuradio "3.9" REQUIRED COMPONENTS runtime pmt <...> OPTIONAL_COMPONENTS uhd)`.

However, this breaks if GR isn't actually built with UHD, because of [this line](https://github.com/gnuradio/gnuradio/blob/master/cmake/Modules/GnuradioConfig.cmake.in#L100). Optional components are included in `Gnuradio_FIND_COMPONENTS`, but the `gnuradio-uhd*.cmake` files won't exist, so CMake fails.

The solution could be to include the file only if:

1. The to-be-included .cmake file exists, OR
2. The component is required, using `Gnuradio_FIND_REQUIRED_<comp>`

That way, if the file exists, include it either way since it was requested one way or another. If the component is required, also include it whether or not the file exists, because we'd like it to fail explicitly if it doesn't exist.

Thoughts?

### System Information

OS: Ubuntu 20.04

### GNU Radio Version

3.9 (maint-3.9)

### Specific Version

0.0.0.0

Optionally include a component that is not built into GR like this (here it's `uhd`):

`find_package(Gnuradio "3.9" REQUIRED COMPONENTS runtime pmt OPTIONAL_COMPONENTS uhd)`

### Relevant log output

```shell
CMake Error at /opt/gnuradio/lib/cmake/gnuradio/GnuradioConfig.cmake:100 (include):
  include could not find load file:

    /opt/gnuradio/lib/cmake/gnuradio/gnuradio-uhdConfig.cmake
  CMakeLists.txt:76 (find_package)
```

ckoehler:
# Pull Request Details

## Description

See related issue. This change allows OPTIONAL_COMPONENTS in a `find_package` call: It'll include the component if the CMake config exists (i.e. it's been enabled during compilation), otherwise it won't.

## Related Issue

Fixes #5309 

CMake config

## Testing Done

Tested OPTIONAL_COMPONENTS with this change and they are correctly included when available, otherwise they are silently passed over.

## Checklist

### DIFF
--- a/cmake/Modules/GnuradioConfig.cmake.in
+++ b/cmake/Modules/GnuradioConfig.cmake.in
@@ -97,7 +97,11 @@ set(GR_COMPONENTS
   )
 
 foreach(target ${GR_COMPONENTS})
-  if (${target} IN_LIST Gnuradio_FIND_COMPONENTS)
-    include("${CMAKE_CURRENT_LIST_DIR}/gnuradio-${target}Config.cmake")
+  set(GR_COMPONENT_INCLUDE_FILE
+      "${CMAKE_CURRENT_LIST_DIR}/gnuradio-${target}Config.cmake")
+  if(${target} IN_LIST Gnuradio_FIND_COMPONENTS
+     AND (EXISTS ${GR_COMPONENT_INCLUDE_FILE}
+          OR "${Gnuradio_FIND_REQUIRED_${target}}"))
+    include(${GR_COMPONENT_INCLUDE_FILE})
   endif()
 endforeach(target)


ryanvolz (approved):
I haven't tested it, but the code looks good and if @ckoehler says it works then I believe him.

PR #5311, opened 2021-11-04
Revert "Replace boost::any with std::any"

mormj:
This reverts commit ccd28dfbb1504fdc29db267acd8aa3354fe10cd2.

# Pull Request Details
Decision made to not include std::any with the 3.10 release as it forces OOTs to compile with c++17 via the changes to the public headers.  

## Description
We will defer the change to std::any as it prevents OOTs between 3.9 and 3.10 from being compatible without changing to c++17 - and there is not necessarily a good reason to enforce this

all OOTs

## Testing Done
Created a new OOT, left it at C++14 and compiled.

## Checklist

### DIFF
--- a/gnuradio-runtime/include/gnuradio/thread/thread_group.h
+++ b/gnuradio-runtime/include/gnuradio/thread/thread_group.h
@@ -17,6 +17,7 @@
 
 #include <gnuradio/api.h>
 #include <gnuradio/thread/thread.h>
+#include <boost/any.hpp>
 #include <boost/core/noncopyable.hpp>
 #include <boost/function.hpp>
 #include <boost/thread/shared_mutex.hpp>
--- a/gnuradio-runtime/include/pmt/pmt.h
+++ b/gnuradio-runtime/include/pmt/pmt.h
@@ -12,8 +12,8 @@
 #define INCLUDED_PMT_H
 
 #include <pmt/api.h>
+#include <boost/any.hpp>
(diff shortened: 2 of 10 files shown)


willcode (approved):

drmpeg:
You may want to also revert the ```GR_PYGCCXML_MIN_VERSION``` portion of https://github.com/gnuradio/gnuradio/commit/0833e0a21abcf7894b3d6dc7c8c27023f91069de

willcode:
Would lowering it down to 1.8.0 to match 3.9 make sense?

drmpeg:
I think 1.9.1 for Ubuntu 20.04 and later would be good.

dl1ksv:
After updating the master branch I came across the problem that my oot's did not compile.
No big deal, but I had to change the compiler settings for each of my oot's. 
And in the future if there are other changes I have again to fix my settings.
So my proposal:
Why not  to use a separate cmake module that contains all compiler settings?

This module could be called in the top level CMakeLists.txt of gnuradio, in the CMakeLists.txt of modtoolt/templates/gr-newmod/CMakeLists.txt and of course in the oot's.

We had to maintain these settings only at one place and in case of a change we would not forget some files.

drmpeg:
Yes, folks do forget to update ```modtoolt/templates/gr-newmod/CMakeLists.txt```. See https://github.com/gnuradio/gnuradio/pull/5300

PR #5312, opened 2021-11-04
Replace boost::any with std::any

### LINKED ISSUE #4780: replace boost::any with std::any
Can do that with C++17, but it changes API in pmt.h

mormj:
This is a modernization possible through C++17

Fixes #4780

This is the commit with only the std::any changes for merging in the 3.11 release


mormj:
It's empty but once the revert is done, should have the single commit

mormj:
Closing - but added to Release 3.11 Project - if we don't break 3.10 too bad - we might be able to lump more breaking changes in 3.11

PR #5308, opened 2021-11-04
Enforce some PEP8 formatting before build

mormj:
# Pull Request Details
Uses a github action to enforce PEP8 with some exceptions
```
--max-line-length=120 --ignore E265,E266,E402,E501,E704,E712,E713,E714,E711,E722,E741,W504,W605
```

## Description
We have a ton of PEP8 violations in the python code throughout the tree.  This fixes most of them, enough to pass a pycodestyle check with the above exceptions

The code was run through autopep8 with 
```
autopep8 -r -i ./gr-fec --select E101,E11,E121,E122,E123,E124,E125,E126,E127,E128,E129,E131,E133,E20,E211,E22,E224,E225,E226,E227,E228,E231,E241,E242,E251,E252,E26,E265,E266,E27,E301,E302,E303,E304,E305,E306,E401,E402,E501,E502,E701,E70,E711,E712,E713,E714,E721,E722,E731,W291,W292,W293,W391,W503,W601,W602,W603,W604,W690
```
and can be checked locally with 
```
find . -iname *.py ! -name *.yml.py  | xargs pycodestyle --max-line-length=120 --ignore E265,E266,E402,E501,E704,E712,E713,E714,E711,E722,E741,W504,W605
```

## Related Issue
Fixes #2172

## Which blocks/areas does this affect?
all the python code in the codebase

## Testing Done
CI passes
Ran some grc examples
Ran test_examples `/usr/bin/python3 -B -m pytest /share/gnuradio/gr310/src/gnuradio/grc/tests -m "examples"`
-- Same grc files fail as usual

## Checklist

- [x] I have read the [CONTRIBUTING document](https://github.com/gnuradio/gnuradio/blob/master/CONTRIBUTING.md).
- [x] I have squashed my commits to have one significant change per commit. 
- [x] I [have signed my commits before making this PR](https://github.com/gnuradio/gnuradio/blob/master/CONTRIBUTING.md#dco-signed)
- [x] My code follows the code style of this project. See [GREP1.md](https://github.com/gnuradio/greps/blob/master/grep-0001-coding-guidelines.md).
- [ ] (NEEDS UPDATE to WIKI) I have updated [the documentation](https://wiki.gnuradio.org/index.php/Main_Page#Documentation) where necessary.
- [ ] I have added tests to cover my changes, and all previous tests pass.

Also, need some top level scripts to format the whole tree and run pycodestyle

PR #5313, opened 2021-11-04
Enforce some PEP8 formatting before build (backport to maint-3.9)

mormj:
Uses a github action to enforce PEP8 with some exceptions
Reformat all the python code to get it to pass this check

Backport of #5308  (would let it sit on master a bit before merging)


gnuradio-runtime/python/gnuradio/gr/CMakeLists.txt
```diff
@@ -27,7 +27,6 @@ if(ENABLE_TESTING)
   set(GR_TEST_TARGET_DEPS "")
   set(GR_TEST_LIBRARY_DIRS "")
   set(GR_TEST_PYTHON_DIRS
-    ${CMAKE_BINARY_DIR}/gruel/src/python
```
marcusmueller, line 30: loool. yeah, that shouldn't be there. ooops...

marcusmueller:
Can we also get that check / corrector as script somewhere in-tree? That way, we could tell contributors that, hey, please run this and that on your changes.

mormj:
> Can we also get that check / corrector as script somewhere in-tree? That way, we could tell contributors that, hey, please run this and that on your changes.

Good point - just added a script in the tools/ directory that does both the modification and check

grc/main.py
```diff
@@ -4,12 +4,14 @@
...
 #
 
-import argparse, logging, sys
+from gi.repository import Gtk
+import argparse
+import logging
+import sys
 
 import gi
 gi.require_version('Gtk', '3.0')
```
willcode, line 13: `gi.require_version()` should be before `from gi.repository ...`

willcode (changes requested):
Over-pepped some ordering by putting imports before some library load instruction. Might not be noticed if we're calling for the system default, but could fail on some systems.

PR #5314, opened 2021-11-10
Addition of a throttle flag to the network blocks

### LINKED ISSUE #5094: UDP source doesn't carry throttling flag
at least in 3.9.

The result is that flowgraphs fed from a UDP source will warn the user about lack of rate limiting.

sepiatone:
Signed-off-by: Aditya Thomas <sepia_tone@protonmail.com>

# Pull Request Details

## Description
A throttle flag is added to the network blocks - TCP source and TCP sink. This prevents a warning about the lack of rate limiting.

## Related Issue
Fixes #5094 

network

## Testing Done
Tested on Ubuntu 20.04

## Checklist

### DIFF
--- a/gr-network/grc/network_tcp_sink.block.yml
+++ b/gr-network/grc/network_tcp_sink.block.yml
@@ -1,6 +1,7 @@
 id: network_tcp_sink
 label: TCP Sink
 category: '[Core]/Networking Tools'
+flags: [ throttle ]
 
 parameters:
 -   id: type
--- a/gr-network/grc/network_tcp_source.block.yml
+++ b/gr-network/grc/network_tcp_source.block.yml
@@ -1,6 +1,7 @@
 id: network_tcp_source
 label: TCP Source
 category: '[Core]/Networking Tools'
+flags: [ throttle ]
 
 parameters:
 -   id: type


willcode:
UDP sink does not provide any backpressure, so it can not act as a throttle. If there is no throttle in the flowgraph, packets will flow at the max rate possible.

The others may or may not act as a throttle, depending on the overall design. However, we have no way of knowing what's on the other end of the connection, so it seems reasonable to mark these 3 blocks as `throttle`.

marcusmueller:
ah, true, I missed the complexity involving the UHD sink!

marcusmueller:
re: conditional flags: uff, tough one. Don't think they get evaluated (like, say, make arguments or the number of outputs a block can have)... hm. Even if they aren't, we probably can just put a variable name in there – it will just be ignored until we figure out how to evaluate flags. We might need to make sure it fits with the safe GRC model!

I know whom to ask: @oscekh knows their way around evaluated properties in GRC! 

@oscekh: situation is that we want to have a flag in a YAML file that depends on a setting in the block properties, something like 

```yaml
flags = [ ] if $padwithzeros else [ "throttle" ]
```

Would that work?

oscekh:
Yeah, the flags don't get evaluated currently. I think evaluating the flags, as you said, sounds like a good approach.

Putting an if/else like that would lead to errors when parsing the YAML, but you could write it like the other evaluated expressions:
```python
flags: ${[] if padwithzeros else [ "throttle" ]}
```
which would mean that the whole thing is parsed as a string that is sent to the [Flags class](https://github.com/gnuradio/gnuradio/blob/master/grc/core/blocks/_flags.py). There I guess you could implement some hack to make it behave sort of correctly (i.e. ignore the if/else part) in the meantime. Something along the lines of:

```python
class Flags(object):

    THROTTLE = 'throttle'
    # ...

    def __init__(self, flags=None):
        if flags is None:
            flags = set()
        if isinstance(flags, str):
            if flags.startswith('${') and flags.endswith('}'):      # +
                flags = flags[2:-1].strip().split('if')             # +
                flags = flags[0].replace('[', '').replace(']', '')  # +

            flags = (f.strip() for f in flags.replace(',', '').split())
        self.data = set(flags)
```
That's all I can think of right now, though I'm not that well versed in the template stuff. :)

marcusmueller:
@oscekh thanks, awesome!

> Putting an if/else like that would lead to errors when parsing the YAML, but you could write it like the other evaluated expressions:
> 
> ```python
> flags: ${[] if padwithzeros else [ "throttle" ]}
> ```

@sepiatone this might be our workaround for the UDP source!

    flags: ${ [] if src_zeros else [ "throttle" ] }

could you make that change (along with removing the flag from the UDP sink, sorry :( ! ), please?

sepiatone:
@marcusmueller I've modified the PR based on the suggestions - (a) removed the throttle flag from UDP sink and (b) modified the UDP source block yaml for eventual conditional evaluation of the 'src_zeros' property.

willcode:
Editorial note - the codebase is standalone without github, so please don't ref PRs in comments even though that seems like it makes docs more complete.

sepiatone:
> 
> 
> Editorial note - the codebase is standalone without github, so please don't ref PRs in comments even though that seems like it makes docs more complete.

@willcode I will make the change.

willcode:
I'm leaning toward changing the TCP blocks and leaving the UDP ones alone for now. The special purpose parser to implement templating for one field seems like a bit too much.

sepiatone:
@marcusmueller Perhaps as @willcode suggested, we could leave out the flag evaluation for now and just update the TCP sink and source blocks for now.

willcode (approved):

PR #5318, opened 2021-11-12
gr-zeromq linger backport from master

Teque5:
## backport linger from master; do not hang indefinitely on flowgraph exit
    
The following squashed commit was approved for public release by The Aerospace
Corporation on 2021-11-01. It is covered software release request #SW19-0024.
Commits made by the Communication Software Implementation Department.

## Description
This alllows ZMQ blocks to exit gracefully upon flowgraph close. We had a slightly different implementation in use for quite a while, but I saw that the `master` branch [already has this fix in place for GR3.9+](https://github.com/gnuradio/gnuradio/blob/e1ce5154560b9e850df4a3c9ba256c601661de31/gr-zeromq/lib/base_impl.cc#L79), so I just backported that implementation.

We have this change used in production, but I wasn't able to quickly add tests for this commit since it's not so easy to build GR3.7 on ubuntu20.04 even with pybombs 😕.

`gr-zeromq`

## Testing Done

### DIFF
--- a/gr-zeromq/lib/base_impl.cc
+++ b/gr-zeromq/lib/base_impl.cc
@@ -28,6 +28,10 @@
 #include "tag_headers.h"
 #include <gnuradio/io_signature.h>
 
+namespace {
+constexpr int LINGER_DEFAULT = 1000; // 1 second.
+}
+
 namespace gr {
 namespace zeromq {
 
@@ -73,6 +77,9 @@ base_sink_impl::base_sink_impl(int type,
         d_socket->setsockopt(ZMQ_HWM, &tmp, sizeof(tmp));
 #endif
     }
+
+    /* Set ZMQ_LINGER so socket won't infinitely block during teardown */
+    d_socket.setsockopt(ZMQ_LINGER, &LINGER_DEFAULT, sizeof(LINGER_DEFAULT));
 
     /* Bind */
     d_socket->bind(address);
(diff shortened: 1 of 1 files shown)


willcode:
That one must have fallen through the cracks as we were starting up 3.9. I can just backport the master version, which also adds LINGER to the source, or you can do it here (if it doesn't require another release request).

Here's the PR that merged on master: https://github.com/gnuradio/gnuradio/pull/3866

Teque5:
Okay I updated the PR to look nearly identical to #3866 now.

willcode:
If you can do a signoff (git amend -s) and another force push, then we can let this run through testing.

willcode:
Ah, this did get into 3.9. Strike my comment about it falling through the cracks. OK, we'll see what people think about putting something in 3.7.

On the DCO/signoff, it gets signed off by you, not the company.

willcode:
@Teque5 You started a great internal discussion on what we should do with maint-3.7. Since we are not testing, tagging, releasing, or even trying out changes to maint-3.7, we're not going to accept any more changes for it. If you are looking to have users pull from a repo, you could fork maint-3.7 and carry your changes on top of that. If you are relying on a distro, they have probably stopped updating 3.7 anyway. Thanks for contributing, and I hope we see more stuff from you on current branches.

We can leave this PR and #5319 up for a bit in case there is further discussion.

willcode:
We are no longer accepting PRs for maint-3.7.

PR #5319, opened 2021-11-12
extremely small output buffer sizes

Teque5:
## allow setting block output sizes from python; allow very small output buffers; fix resampler tag propagation

The following squashed commit was approved for public release by The Aerospace
Corporation on 2021-11-11. It is covered software release request #SW19-0024.
Commits made by the Digital Communications Implementation Department.

## Description
The patch fixes the following issues:
* set `max_output_block_size` and `output_multiple` from python
* allows set maximum output block size smaller than 4kB (for low latency applications)
* fixes tag propagation in the fractional resampler

Similar to #5318 we have this change used in production for quite a long time, but I wasn't able to quickly add tests for this commit since it's not so easy to build GR3.7 on ubuntu20.04 even with pybombs 😕. Ideally there should be some tests especially for the very small output buffer sizes.

I intend on submitting a separate PR to port these chanages forward to `master`.

* `gnuradio-runtime`
* `gr-filter`

## Testing Done

### DIFF
--- a/gnuradio-runtime/include/gnuradio/block_gateway.h
+++ b/gnuradio-runtime/include/gnuradio/block_gateway.h
@@ -150,6 +150,18 @@ public:
         return gr::block::set_min_output_buffer(size);
     }
 
+    long block__max_output_buffer(size_t i) {
+         return gr::block::max_output_buffer(i);
+    }
+
+    void block__set_max_output_buffer(long max_output_buffer) {
+        gr::block::set_max_output_buffer(max_output_buffer);
+    }
+
+    void block__set_max_output_buffer(int port, long max_output_buffer) {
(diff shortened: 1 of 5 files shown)


willcode:
We're not supporting or testing 3.7 now, so not sure whether we can take a commit for it.

@mormj @marcusmueller What do you think?

willcode:
If you submit this for master, there doesn't seem to be a fractional resampler block any more.

willcode:
Check whether you need the block mods on 3.9 and master. Wrapping is done with Pybind11 now, and most functions should be available in Python.

willcode:
We are no longer accepting PRs for maint-3.7.

PR #5322, opened 2021-11-12
CI: Add workflow to update conda CI by commenting in PR with "/rerender"

ryanvolz:
# Pull Request Details

## Description

This provides a Github Action for running conda-smithy to "re-render" (update) the conda CI scripts. To trigger the action, one must start a PR comment with "/rerender". If there are any changes, the action will commit them directly to the PR branch (I think this will work as long as the PR submitter has checked "Allow edits and access to secrets by maintainers" when creating the PR). Overall, this should make it easier for the GNU Radio maintainers to keep the conda CI updated.

I'm open to other ways of triggering this, but I think this is natural way to allow contributors to update the conda CI through an existing PR after discovering it to be necessary (e.g. the CI is failing and a re-render would fix it).

## Related Issue

None

CI

## Testing Done

The action ran successfully on my fork here: https://github.com/ryanvolz/gnuradio/pull/3

## Checklist

### DIFF
--- a/.conda-forge.yml
+++ b/.conda-forge.yml
@@ -13,6 +13,9 @@
 #
 # which will automatically commit any changes.
 # ***** RUN THE ABOVE TO CHECK FOR UPDATES WHENEVER EDITS ARE MADE HERE *****
+#
+# You can also re-render from a pull request by starting a comment with
+# "/rerender", which will trigger a Github workflow to perform the above steps.
 
 # The CI will build conda packages for commits and pull requests, and it will
 # upload the packages to Anaconda Cloud on commits to the branch specified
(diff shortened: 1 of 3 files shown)


mormj:
@ryanvolz - what are the conditions that require retriggering the conda smithy step?  Is it when the conda builds start to fail?

ryanvolz:
> @ryanvolz - what are the conditions that require retriggering the conda smithy step? Is it when the conda builds start to fail?

That's one reason, and that would likely come up if an update is needed because of some change in the Github Action workers. It could also be proactive, as in adding a dependency to the conda recipe and re-rendering to ensure that the build uses the conda-forge "pinned" version, if applicable. It could also be that perhaps nothing is broken, but it's still building against an old version of, say, boost, and you want to update the CI to use the newer "pinned" version.

It's common practice to re-render proactively whenever any sort of update is made to the recipe, just to catch anything you might not be thinking of. Since not everyone is set up to do that, something like this would make that easier.

mormj (approved):

PR #5335, opened 2021-11-18
modtool,cmake: fix python-only QA, clean up build directory

### LINKED ISSUE #5334: New python-only modules failing on build step
### What happened?

PR #5279 fixed one aspect of QA tests, but in the process introduced a bug that prevents a new python-only module from building.

See https://github.com/gnuradio/gnuradio/pull/5279#issuecomment-971941003 and below.

### System Information

Ubuntu 20.04
Source Build

### GNU Radio Version

3.9 (maint-3.9)

### Specific Version

3.9.4.0

- Make a new OOT module using `gr_modtool`
- add a python block
- `make` will fail with "No rule to make target" for the pybind "_python" target

### Relevant log output

```shell
make[2]: *** No rule to make target '../python/pyonly_python', needed by 'python/CMakeFiles/copy_module_for_tests'.  Stop.
make[1]: *** [CMakeFiles/Makefile2:321: python/CMakeFiles/copy_module_for_tests.dir/all] Error 2
make: *** [Makefile:141: all] Error 2
```

samwhiting-snl:
Signed-off-by: shwhiti <shwhiti@sandia.gov>

# Pull Request Details
This pull request allows a python-only module to build without error. All new OOT modules built with gr_modtool currently
will fail in the `make` step.

The `python/CmakeLists.txt` file is now only in charge of copying python source files into `build/test-module/module-name/` where previously it copied in anything in the `python` source directory as well as the pybind `.so` file from the `/build/python/bindings/` directory.

The `bindings/CmakeLists.txt` file is now in charge of copying the `build/python/bindings/` `.so` file into the test module directory. This is only done when there is `c++` code to be added to the module.

These changes make the python package in `build` match more closely to what is actually installed by not copying over things like header files.

## Related Issue
Fixes #5334
related to #3999 and #4825

CMake generated by gr_modtool

## Testing Done
Made a new OOT module with gr_modtool, added a single python block, it built and passed test. Added a c++ block, and the module still builds and passes test.

## Checklist

Will request access to update wiki page for porting modules to 3.9...

### DIFF
--- a/gr-utils/modtool/templates/gr-newmod/python/CMakeLists.txt
+++ b/gr-utils/modtool/templates/gr-newmod/python/CMakeLists.txt
@@ -32,13 +32,10 @@ include(GrTest)
 
 set(GR_TEST_TARGET_DEPS gnuradio-howto)
 
-# Create a module directory that tests can import. It includes everything
-# from `python/` and the built bindings shared lib.
+# Create a package directory that tests can import. It includes everything
+# from `python/`.
 add_custom_target(
   copy_module_for_tests ALL
   COMMAND ${CMAKE_COMMAND} -E copy_directory ${CMAKE_CURRENT_SOURCE_DIR}
           ${CMAKE_BINARY_DIR}/test_modules/howto/
-  COMMAND
-    ${CMAKE_COMMAND} -E copy_directory ${CMAKE_CURRENT_BINARY_DIR}/bindings/
-    ${CMAKE_BINARY_DIR}/test_modules/howto/
-  DEPENDS howto_python)
+)
--- a/gr-utils/modtool/templates/gr-newmod/python/bindings/CMakeLists.txt
+++ b/gr-utils/modtool/templates/gr-newmod/python/bindings/CMakeLists.txt
@@ -36,4 +36,12 @@ GR_PYBIND_MAKE_OOT(howto
    gr::howto
    "${howto_python_files}")
 
+# copy in bindings .so file for use in QA test module
+add_custom_target(
+  copy_bindings_for_tests ALL
+  COMMAND
+    ${CMAKE_COMMAND} -E copy "${CMAKE_CURRENT_BINARY_DIR}/*.so"
+    ${CMAKE_BINARY_DIR}/test_modules/howto/
+  DEPENDS howto_python)
+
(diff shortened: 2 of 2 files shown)


gr-utils/modtool/templates/gr-newmod/python/CMakeLists.txt
```diff
@@ -32,13 +32,10 @@ include(GrTest)
...
 set(GR_TEST_TARGET_DEPS gnuradio-howto)
 
-# Create a module directory that tests can import. It includes everything
-# from `python/` and the built bindings shared lib.
+# Copy in all python source to a build directory python package for QA
+file(MAKE_DIRECTORY ${CMAKE_BINARY_DIR}/test_modules/howto)
 add_custom_target(
   copy_module_for_tests ALL
-  COMMAND ${CMAKE_COMMAND} -E copy_directory ${CMAKE_CURRENT_SOURCE_DIR}
+  COMMAND ${CMAKE_COMMAND} -E copy "${CMAKE_CURRENT_SOURCE_DIR}/*.py"
```
ckoehler, line 39: Would anyone ever have a Python source file pull in some non-Python data, say a `coefficients.txt` file in the same directory? I don't know if that ever happens. If it did, just copying `.py` files over would break that. So it may be safer to just grab everything, I don't think that has side-effects.

ckoehler:
Just one comment, otherwise it looks good to me. I tested this with a python-only module/block and the tests ran successfully. I also tested it with one of our existing modules that contains bindings, and it also worked.

gr-utils/modtool/templates/gr-newmod/python/CMakeLists.txt
samwhiting-snl: I did this thinking that custom or extra files would be handled with custom Cmake, and to leave that to the OOT developer. I'd be fine copying over everything though as well just to simplify things. Anyone else have an opinion on this?

samwhiting-snl: Talked it over a bit and decided copying over everything isn't dangerous and could be useful as you suggested. Just force pushed an update.

mormj (approved):

PR #5340, opened 2021-11-20
Docs: updated usage manual export

777arc:
Couple years worth of updates to the usage manual.  These files were exported using the script (which I also had to tweak a bit to run).


willcode:
Which version is this for? I noticed that `gr::pdu::` was changed to `gr::blocks::` in places. In any case, we'll need `master` and `maint-3.9` versions of this.

777arc:
> Which version is this for? I noticed that `gr::pdu::` was changed to `gr::blocks::` in places. In any case, we'll need `master` and `maint-3.9` versions of this.

Master.  Moving forward we should just do exports more often, not backport any of them.  That way as long as the docs get updated quick enough, exporting it to just master should keep things aligned close enough.

mormj (approved):

willcode:
Backport: use script to generate for v3.9

docs/usage-manual/(exported from wiki) Message Passing.txt
```diff
@@ -225,20 +221,21 @@ However, there are some very good reasons to stick to this format:
...
 == Code Examples ==
 
+Note, in addition to the C++ or Python code below, if adding message passing to a block, you need to edit the block's YAML file as well. Add the corresponding input or output entries (ensuring the domain is set to "message"). You do not need to specify a dtype. You should also ensure that the label value corresponds to the registered port name. See [[YAML_GRC#Inputs_and_Outputs_.28optional.29 | here]] for more details.
+
 === C++ ===
 
 The following is snippets of code from blocks currently in GNU Radio
 that take advantage of message passing. We will be using
-gr::blocks::message_debug and gr::pdu::tagged_stream_to_pdu below
+gr::blocks::message_debug and gr::blocks::tagged_stream_to_pdu below
```
willcode, line 230: `tagged_stream_to_pdu` is in `gr::pdu` in master.

docs/usage-manual/(exported from wiki) Message Passing.txt
```diff
@@ -168,7 +164,7 @@ can call a block's gr::basic_block::_post method directly and pass it a
 message. So any block with an input message port can receive messages
 from the outside in this way.
 
-The following example uses a gr::pdu::pdu_to_tagged_stream block
+The following example uses a gr::blocks::pdu_to_tagged_stream block
```
willcode, line 167: `tagged_stream_to_pdu` is in `gr::pdu` in master.

willcode:
To backport this, would I backport the python script then run it? Or do I need to backport the whole thing and then run the script in the future?

777arc:
> To backport this, would I backport the python script then run it? Or do I need to backport the whole thing and then run the script in the future?

Backporting the python script definitely wont do anything, the content is stored in the wiki, you would have to go through and show an old version of the wiki.  But if you look at the changes made, almost all of them have nothing to do with the version of GR.  We should probably just focus on what to do moving forward, not worry about backporting this, I mean most folks will access it with the live wikipedia, not these exported files anyway.

777arc:
In fact I'll go through that page right now and just convert gr::blocks::pdu_to_tagged_stream block to simply pdu_to_tagged_stream because the point of reading this usage manual is to learn how things function, if someone wants to look at the API for pdu_to_tagged_stream they can find it plenty of other ways.

PR #5345, opened 2021-11-23
runtime: make IN_BLKD kick timer configurable

mormj:
# Pull Request Details
## Description
The fixed setting in the scheduler of 250ms when the input is blocked to retry running the thread is arbitrary and in some blocks with blocking implementations, this timeout can be troublesome and give some weird interaction with the scheduler and the block getting samples from hardware.  For these scenarios, make the value configurable so the user can set what this timeout should be.

Mainly blocks that poll for samples to be ready from some other source.  If the block returns immediately hoping the scheduler will call the work function again, that 250ms timeout could cause problems.

## Testing Done
Ran flowgraphs with UHD blocks

## Checklist

### DIFF
--- a/gnuradio-runtime/gnuradio-runtime.conf.in
+++ b/gnuradio-runtime/gnuradio-runtime.conf.in
@@ -12,6 +12,10 @@ max_messages = 8192
 # Block output buffer size in bytes.
 #buffer_size = 32768
 
+# The time to wait before continuing the block thread when input
+# is blocked
+input_blocked_retry_time_ms = 250
+
 [LOG]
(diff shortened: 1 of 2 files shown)


willcode:
Seems like a reasonable idea given the current code, so I'd be for adding this. There ought to be a way to signal a block that new data has been added to the buffer it's waiting on and it should wake up and check it out.

mormj:
True - yes this would impact all blocks/threads so specifying it somehow on a per-block basis would provide more targeted application.

mormj:
Closing as this should probably be done differently

PR #5347, opened 2021-11-24
cmake: Maintain compiler settings at one point.

dl1ksv:

---

# Pull Request Details

At the moment compiler settings for cmake are maintained at several points,

- the main CMakeLists.txt
- the template file ./gr-utils/modtool/templates/gr-newmod/CMakeLists.txt
- each oot module.

So changing some settings may be cumbersome and error-prone.

## Description
Put the minimum dependency versions for central dependencies in one cmake file and the compiler requirements in an another one, so these settings can be maintained at one place and reused.

## Testing Done
Built gnuradio from scratch and some of my oot's

## Checklist

### DIFF
--- a/cmake/Modules/GrCompilerSettings.cmake
+++ b/cmake/Modules/GrCompilerSettings.cmake
@@ -1,0 +1,98 @@
+# Copyright 2011-2021 Free Software Foundation, Inc.
+#
+# This file is part of GNU Radio
+#
+# SPDX-License-Identifier: GPL-3.0-or-later
+#
+
+if(DEFINED __INCLUDED_GR_COMPILER_SETTINGS_CMAKE)
+    return()
+endif()
+set(__INCLUDED_GR_COMPILER_SETTING_CMAKE TRUE)
+
+########################################################################
+# Compiler version setup
+########################################################################
+# Append -O2 optimization flag for Debug builds (Not on MSVC since conflicts with RTC1 flag)
+IF (NOT MSVC)
+    SET(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -O2")
+    SET(CMAKE_C_FLAGS_DEBUG "${CMAKE_C_FLAGS_DEBUG} -O2")
+ENDIF()
+
+# Check compiler version against our minimum
+IF(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
+    IF(DEFINED CMAKE_CXX_COMPILER_VERSION)
+        IF(${CMAKE_CXX_COMPILER_VERSION} VERSION_LESS ${GCC_MIN_VERSION})
+            MESSAGE(WARNING "\nThe compiler selected to build GNU Radio (GCC version ${CMAKE_CXX_COMPILER_VERSION} : ${CMAKE_CXX_COMPILER}) is older than that officially supported (${GCC_MIN_VERSION} minimum). This build may or not work. We highly recommend using a more recent GCC version.")
+        ENDIF()
+    ELSE()
+        MESSAGE(WARNING "\nCannot determine the version of the compiler selected to build GNU Radio (GCC : ${CMAKE_CXX_COMPILER}). This build may or not work. We highly recommend using GCC version ${GCC_MIN_VERSION} or more recent.")
+    ENDIF()
+ELSEIF(CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+    EXECUTE_PROCESS(COMMAND
+        ${CMAKE_CXX_COMPILER} -v
+        RESULT_VARIABLE _res ERROR_VARIABLE _err
+        ERROR_STRIP_TRAILING_WHITESPACE)
+    IF(${_res} STREQUAL "0")
+        # output is in error stream
+        STRING(REGEX MATCH "^Apple.*" IS_APPLE ${_err})
+        IF("${IS_APPLE}" STREQUAL "")
+            SET(MIN_VERSION ${CLANG_MIN_VERSION})
+            SET(APPLE_STR "")
+            # retrieve the compiler's version from it
+            STRING(REGEX MATCH "clang version [0-9.]+" CLANG_OTHER_VERSION ${_err})
+            STRING(REGEX MATCH "[0-9.]+" CLANG_VERSION ${CLANG_OTHER_VERSION})
+        ELSE()
+            SET(MIN_VERSION ${APPLECLANG_MIN_VERSION})
+            SET(APPLE_STR "Apple ")
+            # retrieve the compiler's version from it
+            STRING(REGEX MATCH "(clang-[0-9.]+)" CLANG_APPLE_VERSION ${_err})
+            STRING(REGEX MATCH "[0-9.]+" CLANG_VERSION ${CLANG_APPLE_VERSION})
+        ENDIF()
+        IF(${CLANG_VERSION} VERSION_LESS "${MIN_VERSION}")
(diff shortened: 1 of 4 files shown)


willcode:
Does this require existing OOTs to change. Assuming yes, since there's a change to gr-newmod.

mormj:
Aren't some things are enforced in the GR codebase but don't necessarily have to be propagated to the OOT, or the OOT can have different requirements.  Like right now, OOTs can use C++14

dl1ksv:
@willcode This pr does not require changes to existing oot's. The change in gr-newmod  only take effect if you are setting up a new oot module.
 In this case the compiler settings are not hard coded but taken vi a GrCompilerSettings from gnuradio and in this way synced with gnuradio.

@mormj Of course you can use different settings. Just copy GrCompilerSettings.cmake to your oot's cmake/Modules dir and modify the corresponding setting. 

THe idea behind this pr:
I got a problem when #5312 was active. Some of my oot's did not build any longer as the compiler settings used in gnuradio where different from my settings in my oot.

After finding te differences and modifying my settings everything worked again. So why trying to keep the settings in sync by maintaining them at one place ?

willcode:
I see ... existing OOTs will continue to work against the new GR since they already have all the info they need. None of the installed cmake files will change?

However, a new OOT won't build against existing versions of GR, since the new cmake files will not exist.

dl1ksv:
Yes and no. You only get a 'new'  OOT if you use the modified version of gr_modtool newmod.
As gr_modtool is part of gnuradio, a new version of gr_modtool implies that the cmake modules have been installed, so the files exist.
But of course, if you setup a new OOT with the modified gr_modtool and transfer this OOT to an old version of gnuradio, the module can't be build as the cmake files are missing.

mormj:
> But of course, if you setup a new OOT with the modified gr_modtool and transfer this OOT to an old version of gnuradio, the module can't be build as the cmake files are missing.

And the new OOT structure prevents new OOTs from being used with older modtool anyways, so this shouldn't be an issue

cmake/Modules/GrMinReq.cmake
```diff
@@ -0,0 +1,26 @@
...
+set(GR_PYTHON_MIN_VERSION "3.6.5")    ## Version in Ubuntu 18.04LTS
+set(GR_NUMPY_MIN_VERSION "1.13.3")    ## Version in Ubuntu 18.04LTS
+set(GR_PYGCCXML_MIN_VERSION "2.0.0")  ## Version to support c++17 (in pip)
+set(GCC_MIN_VERSION "8.3.0")          ## debian buster
+set(CLANG_MIN_VERSION "11.0.0")       ## debian bullseye, Fedora 33
+set(APPLECLANG_MIN_VERSION "1100")    ## same as clang 11.0.0, in Xcode11
+set(MSVC_MIN_VERSION "1914")          ## VS2017 15.7, for full-ish C++17 support
+set(VOLK_MIN_VERSION "2.4.1")         ## first version with CPU features
+set(PYBIND11_MIN_VERSION "2.4")       # pybind11 sets versions like 2.4.dev4, which compares < 2.4.3
+set(GR_THRIFT_MIN_VERSION "0.13")
```
jdemel, line 25: I recommend to stick with the versions that are set in the top-level `CMakeLists.txt`.
```
set(GR_BOOST_MIN_VERSION "1.69")      ## Version in CentOS 8 (EPEL)
set(GR_CMAKE_MIN_VERSION "3.16.3")    ## Version in Ubuntu 20.04LTS
set(GR_MAKO_MIN_VERSION "1.1.0")      ## Version in Ubuntu 20.04LTS
set(GR_NUMPY_MIN_VERSION "1.17.4")    ## Version in Ubuntu 20.04LTS
set(GCC_MIN_VERSION "9.3.0")          ## Version in Ubuntu 20.04LTS
```
Especially, GCC 8.3 will be annoying because it requires special treatment for `std::filesystem` while GCC 9+ does not. 
I'd argue that the next GR release doesn't need to support legacy systems that will become obsolete while GR 3.10 is still the current version.

For MSVC, I suggest to bump the minimum version to `1931` just to get rid of some headache with incompatibilities. Or at least `1920`, which is the first VS2019 release.

gr-utils/modtool/templates/gr-newmod/CMakeLists.txt
```diff
@@ -28,6 +28,8 @@ set(CMAKE_BUILD_TYPE ${CMAKE_BUILD_TYPE} CACHE STRING "")
 
 # Make sure our local CMake Modules path comes first
 list(INSERT CMAKE_MODULE_PATH 0 ${CMAKE_SOURCE_DIR}/cmake/Modules)
+# Find gnuradio to get access to the cmake modules
+find_package(Gnuradio "3.9" REQUIRED)
```
jdemel, line 32: This should be 3.10, or am I missing smth?

cmake/Modules/GrCompilerSettings.cmake
```diff
@@ -0,0 +1,98 @@
...
+ENDIF()
+
+# Configure C++ standard if not externally specified (will actually be
+# set after CppUnit check below). Use the variable CMAKE_CXX_STANDARD
+# since it will actually be used for this purposes starting in CMake 3.1.
+
+set(CMAKE_C_EXTENSIONS OFF)
+set(CMAKE_CXX_EXTENSIONS OFF)
+set(CMAKE_CXX_STANDARD 17)
+set(CMAKE_C_STANDARD 11)
```
jdemel, line 69: I suggest to bump this to `C17` because it allows everyone to rely on "bugfixes". [Compiler support](https://en.wikipedia.org/wiki/C17_(C_standard_revision)#Compiler_support) seems to be very good. We can just bump the MSVC minimum to `1928`.

dl1ksv:
@jdemel Thans for your review. I should have noticed, that the minimum version numbers had been updated.

Of course you can set different requirements for your OOT module. As the cmake/Modules directory of the oot module is
handled before the gnuradio/cmake/Modules directory, you can put your version of GrMinReq.cmake or GrCompilerSettings.cmake in the modules cmake/Modules directory.

jdemel:
A few comments.

I just saw, you signed your commit with some "@users.noreply.github.com" email address. It'd be very useful if you commit and sign your commits with an email address.

It seems like C17 was to ambitious. Most of our CI fails because of that. I guess we need to revert to C11 for now.

CentOS 8.4 fails with some outdated Boost version. In that case, I assume there are workarounds to install a newer version.

mormj:
> CentOS 8.4 fails with some outdated Boost version. In that case, I assume there are workarounds to install a newer version.

This one I still don't understand - it is set up in the CI docker to use boost 1.69 through EPEL, but then fails the check.  I need to debug - but also not terribly concerned about Centos8 as it is EOL

dl1ksv:
@jdemel 3470424+dl1ksv@users.noreply.github.com is a valid email address provided by github for signing and it is used to keep  my personal email address private.  This is the recommended way by github.

I wonder why C17 fails on Fedora 34 but works on 35. Both have the same compiler version 11.2.1

dl1ksv:
I'll go back to c 11.

mormj:
also, I'm going to merge https://github.com/gnuradio/gnuradio/pull/5383 when CI passes, so please downgrade the minimum boost to 1.65 as well.  Then this should be good to go and a great addition!  Thanks!

dl1ksv:
Switch back to boost 1.65 and c 11

mormj:
Sorry I ended up not pulling Boost back because Centos8 is a pain generally.  Will squash these and merge this.  Really love the cmake simplification.

PR #5371, opened 2021-12-05
gnuradio: replace deprecated distutils

dl1ksv:
---

# Pull Request Details
With the upcoming python 3.10 ( for example in fedora 35 ) the package distutils is marked as depreciated and will be removed in 3.12.

## Description
Python provides replacements for the distutils functions.
Only distutils.sysconfig.get_python_lib() has no direct counterpart,.

grc, grc generated flowgraphs

## Testing Done
Building gnuradio from scratch on fedora 33 and 35 

## Checklist

### DIFF
--- a/cmake/Modules/GrPython.cmake
+++ b/cmake/Modules/GrPython.cmake
@@ -127,7 +127,7 @@ macro(GR_PYTHON_CHECK_MODULE desc mod cmd have)
     GR_PYTHON_CHECK_MODULE_RAW(
         "${desc}" "
 #########################################
-from distutils.version import LooseVersion
+from packaging.version import Version as LooseVersion
 try:
     import ${mod}
     assert ${cmd}
@@ -145,19 +145,22 @@ endmacro(GR_PYTHON_CHECK_MODULE)
 if(NOT DEFINED GR_PYTHON_DIR)
 execute_process(
     COMMAND ${PYTHON_EXECUTABLE} -c "import os
+import sysconfig
 import site
-from distutils.sysconfig import get_python_lib
 
+install_dir = None
 prefix = '${CMAKE_INSTALL_PREFIX}'
-
-#ask distutils where to install the python module
-install_dir = get_python_lib(plat_specific=True, prefix=prefix)
 
 #use sites when the prefix is already recognized
 try:
   paths = [p for p in site.getsitepackages() if p.startswith(prefix)]
   if len(paths) == 1: install_dir = paths[0]
 except AttributeError: pass
+
+if not install_dir:
+    #find where to install the python module
+    install_dir = sysconfig.get_path('platlib')
+    prefix = sysconfig.get_config_var('prefix')
 
 #strip the prefix to return a relative path
 print(os.path.relpath(install_dir, prefix))"
--- a/.packaging/conda_recipe/meta.yaml
+++ b/.packaging/conda_recipe/meta.yaml
@@ -64,6 +64,7 @@ requirements:
     - mako
     - mpir  # [win]
     - numpy
+    - packaging
     - pip  # [win]
     - pybind11
     - pybind11-abi
@@ -107,6 +108,7 @@ requirements:
     - mako
     - menuinst  # [win]
     - numpy
+    - packaging
     - {{ pin_compatible('portaudio') }}
     - pygobject
     - pyqt
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -229,6 +229,20 @@ add_custom_target(uninstall
 # Enable python component
 ########################################################################
 include(GrPython)
+GR_PYTHON_CHECK_MODULE_RAW(
+    "packaging"
+    "import packaging"
+    PACKAGING_FOUND
+)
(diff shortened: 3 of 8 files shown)


dl1ksv:
As I could not test on Windows and OSX could someone provide the output of

os.name
and

from distutils.sysconfig import get_python_lib
get_python_lib()

for those systems, so that I can check my replacement.

willcode:
Is `packaging` a new dependency? It gets brought in with python3-devel in Fedora, not sure about others. For runtime only, it may not be there.

dl1ksv:
packaging is the recommended replacement for distutils.version  ( Python PEP 632 ) and cmake works in checks for Ubuntu 20.0, Fedora 34/35 and Debian 11 . So I think packaging is part of those distros.

mormj:
I fired up a fresh Ubuntu 20.04 docker and it doesn't come in with python3-dev.  I had to install python3-packaging.  It might come in from some other common thing that we install

willcode:
With an ubuntu:20.04 docker images, `apt install gnuradio` did not bring in `packaging`, so I think it will have to be an explicit dep. It does not even come in with gnuradio-dev.

mormj:
`python3-packaging` brings it in as well as pip3 install packaging

mormj:
It would be nice to get this into 3.10 since it introduces a new dependency that will be hard to address later.  Do you think you could put in the cmake check?

cmake/Modules/GrPython.cmake
```diff
@@ -146,12 +146,22 @@ if(NOT DEFINED GR_PYTHON_DIR)
...
+if prefix is None:
+    prefix = os.path.normpath(sys.exec_prefix)
+# To do: How to handle on OSX or Windows
+if os.name == 'posix':
+    libdir = getattr(sys, 'platlibdir', 'lib')
+else :
+    libdir = 'lib'
+
+libpython = os.path.join(prefix, libdir,'python' + sysconfig.get_python_version())
+install_dir = os.path.join(libpython, 'site-packages')
```
jdemel, line 164: So this is the tricky part that needs to be replaced. `distutils`' `get_python_lib` returns `dist-packages` under some circumstances. e.g.
```python
>>> from distutils import sysconfig
>>> sysconfig.get_python_lib()
'/usr/lib/python3/dist-packages' 
>>> sysconfig.get_python_lib(plat_specific=True, prefix='/foo/bar')
'/foo/bar/lib/python3.8/site-packages'
```
Now, this is hard-coded to `site-packages`. I assume there should be at least an option for package maintainers to change that behavior.
Since this whole thing is kinda broken, I would appreciate more options anyways. e.g. an install with the default prefix `/usr/local` won't work "out-of-the-box" because the Python packages end up in `/usr/local/lib/python3.8/site-packages`, which is not in the `PYTHONPATH`.
```python
>>> import sys
>>> sys.path
['', '/usr/bin', '/usr/lib/python38.zip', '/usr/lib/python3.8', '/usr/lib/python3.8/lib-dynload', '/usr/local/lib/python3.8/dist-packages', '/usr/lib/python3/dist-packages']
```

mormj: Seems like rather than hardcoding site-packages or using distutils, sysconfig should be able to be used here

dl1ksv: I found a solution completely based on sysconfig and it should work for different os.

dl1ksv:
sysconfig.get_path('platlib') results in /usr/local/lib/python3.9/site-packages/ instead of /usr/local/lib/python3.9/dist-packages/ on debian.
I have to figure out how to get the correct result on debian.

dl1ksv:
Do the conda package checks fail as python3_packaging is missing ?

jdemel:
```
L1146ff: -- Python checking for packaging - not found
-- 
-- ***************************** WARNING!!! *******
-- The python package packaging is missing
-- So some checks will fail
-- Please install python3-packaging
-- or run pip3 install packaging
-- ************************************************
...
L1170: --   Dependency PACKAGING_FOUND = FALSE
```
To answer your question: Yes. CMake can't find packaging, CMake disables all Python related components and finally the Conda package tests fails because `import gnuradio.analog` fails. The amount of output is huge. It's good that the build system shares that much information though.

If you happen to find a fix for this whole `sysconfig.get_path('platlib')` mess on Debian, please share it. I tried to solve this once for an OOT with `distutils` and finally gave up.

dl1ksv:
But how do I get the conda build system to install python3_packaging ?
@jdemel My latest fix works on debian , too. I tested it on an a pi with debian bullseye and the debian test above works , too.
The trick is to use site.getsitepackages() first to find the dist_packages directory instead of the site_packages.

ryanvolz:
> But how do I get the conda build system to install python3_packaging ?

Put `packaging` in the list of dependencies in https://github.com/gnuradio/gnuradio/blob/master/.packaging/conda_recipe/meta.yaml under both `host` (needed at build time) and `run` (needed when GR is installed).

mormj:
This all seems to work without changing the installation path behavior introduced by @schneider42.  Will merge unless anyone has any other concerns

PR #5372, opened 2021-12-06
network: fix segfaults when TCP & UDP blocks are restarted

### LINKED ISSUE #5112: network: UDP & TCP blocks segfault when stopped & restarted
While attempting to update Gqrx to use the new gr-network UDP Sink, I noticed that a segfault occurs if the flowgraph is stopped and restarted. (Gqrx does this in many places, for instance when the user changes the demodulator type or presses the Start/Stop button.)

The segfault occurs because the block deallocates all its resources in the `stop()` method, treating it like a destructor:

https://github.com/gnuradio/gnuradio/blob/65366505b3d04422cb9ec93414e1c8e474f5aa08/gr-network/lib/udp_sink_impl.cc#L164-L196

The UDP Source and TCP Sink blocks have the same problem.

To demonstrate the problem using GRC, create a flow graph containing any of these blocks. Manually edit the code, replacing `tb.start()` with `tb.start(); tb.stop(); tb.start()`. Run the modified Python code, and a segmentation fault will occur.

I think the code in these blocks' `stop()` methods should be moved into their destructors.

@ghostop14

argilo:
# Pull Request Details

## Description
The TCP and UDP blocks segfault if `start()` is called after `stop()`, because `stop()` frees resources that are not re-allocated by `start()`. To fix this, I've moved resource allocation for these blocks from the constructor to `start()`.

## Related Issue
Fixes #5112.
Also fixes the Gqrx crash noted in https://github.com/gqrx-sdr/gqrx/pull/982.

* TCP Sink
* UDP Sink
* UDP Source

## Testing Done
I've added regression tests. All of them fail with segfaults before the changes, and pass afterwards.

In addition, I've verified that UDP output in Gqrx works with this change applied.

## Checklist

### DIFF
--- a/gr-network/lib/udp_sink_impl.cc
+++ b/gr-network/lib/udp_sink_impl.cc
@@ -45,6 +45,8 @@ udp_sink_impl::udp_sink_impl(size_t itemsize,
     : gr::sync_block("udp_sink",
                      gr::io_signature::make(1, 1, itemsize * veclen),
                      gr::io_signature::make(0, 0, 0)),
+      d_host(host),
+      d_port(port),
       d_itemsize(itemsize),
       d_veclen(veclen),
       d_header_type(header_type),
@@ -61,10 +63,6 @@ udp_sink_impl::udp_sink_impl(size_t itemsize,
     // reassembled. Now for local nets that support jumbo frames, the max payload
     // size is 8972 (9000-the UDP 28-byte header) Same rules apply with
     // fragmentation.
-
-    d_port = port;
-
-    d_header_size = 0;
 
     switch (d_header_type) {
     case HEADERTYPE_SEQNUM:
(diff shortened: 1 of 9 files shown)


gr-network/lib/udp_sink_impl.cc
```diff
@@ -62,10 +64,6 @@ udp_sink_impl::udp_sink_impl(size_t itemsize,
     // size is 8972 (9000-the UDP 28-byte header) Same rules apply with
     // fragmentation.
 
-    d_port = port;
-
-    d_header_size = 0;
-
```
argilo, line 68: Removed these, since the initializer list takes care of them.

gr-network/lib/udp_sink_impl.cc
```diff
@@ -94,13 +92,21 @@ udp_sink_impl::udp_sink_impl(size_t itemsize,
             "least 8 bytes once header/trailer adjustments are made.");
     }
 
-    d_seq_num = 0;
-
```
argilo, line 98: This is already in the initializer list.

gr-network/lib/udp_sink_impl.h
```diff
@@ -24,6 +24,7 @@ namespace network {
 class NETWORK_API udp_sink_impl : public udp_sink
 {
 protected:
+    const std::string d_host;
```
ThomasHabets, line 27: I have come around to thinking that const member variables is more trouble than they're worth, so am leaning towards no-consting this. Other opinions?

ThomasHabets (commented):
Nice. These modules still have some needless manual memory management I meant to get to, but I'd need to make sure that boost doesn't do anything magic behind the scenes.

Other than my question about the const, this looks great.

gr-network/lib/udp_sink_impl.h
argilo: I think I just copy-pasted that from the constructor. The TCP sink doesn't have it. Removed.

PR #5383, opened 2021-12-09
ci: Remove CentOS 8 from list of containers

mormj:
~Boost 1.69, though available in CentOS 8 EPEL installs differently and is a bit unwieldy to use.  Instead, let's just keep with the same boost version that is in 3.9~

Removing CentOS 8 from the list of containers
It is too problematic to track and not worth changing our min versions over

### DIFF
--- a/.github/workflows/make-test.yml
+++ b/.github/workflows/make-test.yml
@@ -74,11 +74,11 @@ jobs:
             cxxflags: -Werror -Wno-error=invalid-pch -Wno-error=deprecated-declarations
             ctest_args: '-E "qa_cpp_py_binding|qa_cpp_py_binding_set|qa_ctrlport_probes|qa_qtgui"'
             ldpath: /usr/local/lib64/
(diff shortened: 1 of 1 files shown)


jdemel:
According to [this CentOS announcement](https://www.centos.org/centos-linux-eol/), CentOS 8 is EOL after 2021-12-31, i.e. in two weeks. I suggest to drop support for CentOS8 in GR 3.10 because of that.

However, [CentOS Stream 8](https://www.centos.org/centos-stream/#centos-stream-8) is still alive. EOL is set to 2024-05-31. Then again, we'd need to check which libboost ships with that release by default.

mormj:
> However, [CentOS Stream 8](https://www.centos.org/centos-stream/#centos-stream-8) is still alive. EOL is set to 2024-05-31. Then again, we'd need to check which libboost ships with that release by default.

Looks like CentOS Stream 8 is still on 1.66 unfortunately

PR #5384, opened 2021-12-09
Squelch_eob typo in comments propagates to online documentation.

taclane:
gnuradio documentation automatically generated for [pwr_squelch_ff](https://www.gnuradio.org/doc/doxygen/classgr_1_1analog_1_1pwr__squelch__ff.html) and [pwr_squelch_cc](https://www.gnuradio.org/doc/doxygen/classgr_1_1analog_1_1pwr__squelch__cc.html) states:

> The block will emit a tag with the key pmt::intern("squelch_sob") with the value of pmt::PMT_NIL on the first item it passes, and with the key pmt::intern("squelch:eob") on the last item it passes.

Based on squelch_base_ff and squelch_base_cc, it should be `pmt::intern("squelch_eob")`, not `pmt::intern("squelch:eob")`.

https://github.com/gnuradio/gnuradio/blob/1a0be2e6b54496a8136a64d86e372ab219c6559b/gr-analog/lib/squelch_base_ff_impl.cc#L28
https://github.com/gnuradio/gnuradio/blob/1a0be2e6b54496a8136a64d86e372ab219c6559b/gr-analog/lib/squelch_base_cc_impl.cc#L27

### DIFF
--- a/gr-analog/include/gnuradio/analog/pwr_squelch_cc.h
+++ b/gr-analog/include/gnuradio/analog/pwr_squelch_cc.h
@@ -44,7 +44,7 @@ public:
      *
      * The block will emit a tag with the key pmt::intern("squelch_sob")
      * with the value of pmt::PMT_NIL on the first item it passes, and with
-     * the key pmt::intern("squelch:eob") on the last item it passes.
+     * the key pmt::intern("squelch_eob") on the last item it passes.
      */
     static sptr make(double db, double alpha = 0.0001, int ramp = 0, bool gate = false);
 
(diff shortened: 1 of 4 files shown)


mormj (approved):

mormj:
Because this updates the public header files, the corresponding hashes need to be updated in the python binding files.

Run md5sum on each of these files and paste the result into {blockname}_python.cc and it will build here

mormj (changes requested):
Needs update to python bindings

mormj (approved):
LGTM

777arc:
FYI I've updated the wiki docs to reflect this change

PR #5385, opened 2021-12-10
runtime: configurable blkd_input timer

mormj:
# Pull Request Details
## Description
The fixed setting in the scheduler of 250ms when the input is blocked to retry running the thread is arbitrary and in some blocks with blocking implementations, this timeout can be troublesome and give some weird interaction with the scheduler and the block getting samples from hardware.  For these scenarios, make the value configurable so the user can set what this timeout should be.

Mainly blocks that poll for samples to be ready from some other source.  If the block returns immediately hoping the scheduler will call the work function again, that 250ms timeout could cause problems.

## Checklist

### DIFF
--- a/gnuradio-runtime/include/gnuradio/block.h
+++ b/gnuradio-runtime/include/gnuradio/block.h
@@ -518,6 +518,30 @@ public:
     void set_min_output_buffer(int port, long min_output_buffer);
 
     /*!
+     * \brief DEPRECATED Configure the timer set when input is blocked \p port.
+     *
+     * \details
+     * This is an advanced/experimental feature and might be removed in a future
+     * version. Calling this can affect some fundamental assumptions about the
+     * system behavior and
+     * performance.
+     *
+     * In the TPB scheduler, when a block has no work to do because there
+     * is no data at it inputs, it sets a timer and tries again after a
+     * period of time.  The default is 250 ms, but this can be configured
+     * differently per block when necessary
+     *
+     * \param timer_value_ms the timer value in milliseconds
+     */
+    void set_blkd_input_timer_value(unsigned int timer_value_ms);
+
+    /*!
+     * \brief DEPRECATED Returns timer value set when input is blocked
+     */
+    unsigned int blkd_input_timer_value();
+
+
+    /*!
      * \brief Allocate the block_detail and necessary output buffers for this
      * block.
      */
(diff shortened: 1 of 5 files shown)


bastibl:
Relying on this timeout is, in my opinion, an anti-pattern and I wouldn't add an API to help exploit it. At the moment it's just some detail of the runtime that would ideally be removed. Adding an API makes it a feature of the runtime and even hard to get rid of.

gnuradio-runtime/include/gnuradio/block.h
```diff
@@ -517,6 +517,29 @@ class GR_RUNTIME_API block : public basic_block
...
+     * \brief Configure the timer set when input is blocked \p port.
+     *
+     * \details
+     * This is an advanced feature. Calling this can affect some
+     * fundamental assumptions about the system behavior and
+     * performance.
+     *
+     * In the TPB scheduler, when a block has no work to do because there
+     * is no data at it inputs, it sets a timer and tries again after a
+     * period of time.  The default is 100 ms, but this can be configured
```
willcode, line 530: 250ms?

willcode (approved):

gnuradio-runtime/include/gnuradio/block.h
mormj: whoops - yep you're right.

willcode:
Agree with @bastibl that this should be done differently in the long term. I don't have any problem with adding this technically, but maybe we should mark it experimental or deprecated or something. Or there could be "Normal", "Prefer Low Latency" and other options that just happen to be backed by specific timeouts in the current implementation. I could see a goodness-based API living on.

mormj:
> Relying on this timeout is, in my opinion, an anti-pattern and I wouldn't add an API to help exploit it. At the moment it's just some detail of the runtime that would ideally be removed. Adding an API makes it a feature of the runtime and even hard to get rid of.

Long term I'd definitely agree with you, but I don't see that happening for quite a while and once that is fixed will be part of the major version bump and API will change drastically.  Ideally this wouldn't be necessary but without a complete gutting of the scheduler there is a fixed constant there that should at least be tunable in the short term?  Or it is just an ugly patch.

bastibl (approved):
Maybe I lack imagination, but I only know of blocks that were badly designed that wanted this feature. So it feels a bit like an API that helps you to shoot yourself in the foot :-) But probably it's helpful in some cases.

mormj:
It is definitely for badly design blocks lol!!

mormj:
Went ahead and deprecated it so we can yank it whenever a better solution is implemented

willcode:
Dropped the Backport-3.9 label since this is an experimental sort of thing.

PR #5386, opened 2021-12-10
3.9 ptest support

balister:
These patches are cherry picked from master onto maint-3.9. Make test passed on my x86 and the sausage machine ptest output from a build looked reasonable. See:

https://github.com/balister/gnnuradio-ptest/tree/hardknott-qemu-next


willcode:
@balister I saw not CI errors, so merged. But CI had not run and in reality all targets failed, so reverted.

PR #5391, opened 2021-12-13
s2x constell energy normalization

maswx:
---
name: s2x constellation energy normalization
about: DVB-S2X.
title: 's2x constellation energy normalization'
labels: 'DVB-S2X'
assignees: ''

---

### DIFF
--- a/gr-dtv/lib/dvbs2/dvbs2_modulator_bc_impl.cc
+++ b/gr-dtv/lib/dvbs2/dvbs2_modulator_bc_impl.cc
@@ -92,6 +92,10 @@ dvbs2_modulator_bc_impl::dvbs2_modulator_bc_impl(dvb_framesize_t framesize,
             r2 = 0;
             break;
         }
+        r0 = sqrt(8.0 / (2.0 * (r1 * r1) + 4.0 * (r2 * r2) + 2.0 * (r3 * r3)));
+        r1 *= r0;
+        r2 *= r0;
+        r3 *= r0;
         m_8psk[0] = gr_complex((r1 * cos(GR_M_PI)), (r1 * sin(GR_M_PI)));
         m_8psk[1] = gr_complex((r2 * cos(GR_M_PI * 1.352)), (r2 * sin(GR_M_PI * 1.352)));
         m_8psk[2] =
@@ -262,6 +266,9 @@ dvbs2_modulator_bc_impl::dvbs2_modulator_bc_impl(dvb_framesize_t framesize,
                 r1 = 0;
                 break;
             }
+            r0 = sqrt(16.0 / (8.0 * (r1 * r1) + 8.0 * (r2 * r2)));
+            r1 *= r0;
+            r2 *= r0;
             m_16apsk[0] =
                 gr_complex((r1 * cos(GR_M_PI / 8.0)), (r1 * sin(GR_M_PI / 8.0)));
             m_16apsk[1] =
@@ -409,6 +416,10 @@ dvbs2_modulator_bc_impl::dvbs2_modulator_bc_impl(dvb_framesize_t framesize,
                 break;
             }
         }
+        r0 = sqrt(32.0 / (4.0 * (r1 * r1) + 12.0 * (r2 * r2) + 16.0 * (r3 * r3)));
+        r1 *= r0;
+        r2 *= r0;
+        r3 *= r0;
         m_32apsk[0] =
             gr_complex((r3 * cos(11 * GR_M_PI / 16.0)), (r3 * sin(11 * GR_M_PI / 16.0)));
         m_32apsk[1] =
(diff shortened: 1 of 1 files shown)


marcusmueller:
Hi @maswx! Thanks for contributing this!

`clang-format` complains about the file that you've changed doesn't adhere to our "standardized" code formatting. Luckily, that's easy to fix: 

```shell
clang-format -i dvbs2_modulator_bc_impl.cc 
git commit --fixup=HEAD
git rebase -i --autosquash --autostash origin/master
git commit --amend -s
```

The last line actually fixes another small problem: 
Sadly, your commit is lacking the sign-off line. Using `-s` when `git commit` fixes that, and `--amend` just modifies the last commit instead of the current one.

See https://github.com/gnuradio/gnuradio/blob/master/DCO.txt for what you do when you add the "signed-off by:" line to your commits. 

Again, thank you! It's great to have you around :)

drmpeg:
@maswx Can you provide a reference document for the equations that you're using?

maswx:
Hi, @drmpeg , According to [DVBS2 standard](https://www.etsi.org/deliver/etsi_en/302300_302399/302307/01.03.01_60/en_302307v010301p.pdf). P27, E=1 (E= Unit Average symbol Energy) corresponding to [R1]2+ 3[R2]2 = 4, and R2=1, E0(equivalent to r0 in the code) could be calculated. The S2 part of the code already supports constellation energy normalization correctly, but all constellations of S2X do not. According to the current constellation calculation method, only the radius of the outermost circle is ensured to be 1, but its energy is not 1.   I just reviewed the code and found that S2X-8APSKL still does not support constellation energy normalization. I will modify it and upload it later。。。

drmpeg:
@maswx I understand what you're trying to do, but constellation normalization is not explicitly called out in the DVB-S2X specification and the equations that you're using do not appear in any published DVB document.

Have you tested these normalized constellations with a real demodulator and do they provide better performance?

To be honest, unless you can make a better case, I'm not going to accept this pull request.

maswx:
@drmpeg I am a DVB-S2X receiver chip designer. I have confirmed that the signals sent by Rohde & Schwarz’s BTC/satellites are energy-normalized signals instead of ensuring that the outermost circle is 1. This is obvious. At the beginning of the satellite signal design, it is necessary to ensure that the signal power is 1. In addition, there is a description in the [DVB-S2X standard ](https://www.etsi.org/deliver/etsi_en/302300_302399/30230702/01.03.01_60/en_30230702v010301p.pdf) P30 5.4.3 or 5.4.4 "See ETSI EN 302 307-1 [3], clause 5.4.4." e.t.

drmpeg:
Okay, that's interesting. Let me do some testing with the STiD135 demodulator chip over the next day or two.

maswx:
What is certain is that this kind of energy normalization modification will not improve the Gaussian performance in STiD135 demodulator。Because the AGC of the demodulator will automatically normalize the energy（no outermost=1）. This modification only affects the accuracy of adding Gaussian noise at the transmitter. The normalized energy of the constellation IQ data generated by the original code is less than 1, which causes the additive white Gaussian noise generated under the same noise variance to be smaller than the preset EsNo. If you want to experiment with the difference, it is recommended to add noise to the generated IQ data to the threshold before transmitting it to the STiD135 demodulator for comparison.

drmpeg:
With earlier ST DVB-S2 demodulator chips, going from R2 = 1 (or R3 = 1) to E = 1 made a huge difference. From almost not working to perfect.

drmpeg:
I've tested the constellation average power and it looks like the 128APSK equation is incorrect. I'm getting 1.766 instead of 1.0.

I also tested 8+8APSK with the STiD135, and there is a very noticeable improvement in minimum SNR from 8.4 dB to 6.0 dB (which is very close the quoted AWGN performance in the DVB-S2X specification Table 20a).

In addition, I checked the average power of constellations like 256APSK 20/30 that are specified with absolute values and they are indeed power normalized to 1.0.

So it's clear that this is required for optimum performance. Apologies for my initial doubts.

After you fix 128APSK, please squash your commits and use a single and better commit message. I'll suggest:

```gr-dtv: Add energy normalization for DVB-S2X constellations.```

drmpeg:
Quick squash tutorial.

image

How it looks in nano.

image

Squash commit by changing ```pick``` to ```s```.

image

After saving the squash, edit the commit message. Also comment out the previous commit messages.

image

maswx:
Sorry。when the 128APSK energy is normalized, the outermost radius should be 1.3206 instead of 1.7
This is a bug, I will fix this problem later and commit it.

maswx:
Sorry。when the 128APSK energy is normalized, the outermost radius should be 1.3206 instead of 1.7
This is a bug, I will fix this issue later and commit it.

drmpeg:
Thanks, 128APSK is good now.

If you would, one more squash would make this pull request perfect. You should have done ```git rebase -i HEAD~4``` previously, so there's still one extra commit. I should have warned you about that. A ```git rebase -i HEAD~2``` this time will finish things up.

maswx:
OK, this commit  seems to be fine. 
I rarely use git. 
SVN is used in daily work. 
sorry for the inconvenience!

drmpeg (approved):
Thank you for this awesome contribution. The DVB-S2X specification is about as clear as mud on this and I would have never figured it out on my own.

PR #5394, opened 2021-12-13
findlog4cpp: extend PATH hints to include default x86_64 lib path

andrepuschmann:
# Pull Request Details

## Description
The patch adds `/usr/lib/x86_64-linux-gnu` as a default search path when looking for the log4cpp lib.

## Related Issue
There is no related issue it seems but it does fail on Ubuntu 20.04 when installing the dependency and then trying to build.

## Checklist


marcusmueller:
Hm... this is pretty much Ubuntu 20.*-specific, but it solves a problem. 

I'm a bit surprised that relying on `pkg-config` isn't enough here.

marcusmueller (approved):

PR #459, opened 2015-04-05
Qtgui/freq control panel

trondeau:

PR #5400, opened 2021-12-16
qtgui: add bounds checks to calculation of d_fft_size_combo index

argilo:
## Description
The QT GUI Frequency Sink block segfaults if a large (e.g. 8192) or small (e.g. 17) FFT size is specified, and the Control Panel is subsequently opened. This happens because there are no bounds checks on the index computed here:
https://github.com/gnuradio/gnuradio/blob/19864fc6ede818b169259607bb85bd5931013e02/gr-qtgui/lib/freqcontrolpanel.cc#L243-L247

It looks like this bug has been around since the control panel was added in https://github.com/gnuradio/gnuradio/pull/459, so GNU Radio 3.7.7+ would be affected.

* QT GUI Frequency Sink

## Testing Done
I used this flow graph to test:
Screenshot from 2021-12-16 13-11-17
After starting the flow graph, middle click on the Frequency Sink and click "Control Panel". Prior to this change, a segfault will occur. Same if the FFT size is changed to 17.

## Checklist

### DIFF
--- a/gr-qtgui/lib/freqcontrolpanel.cc
+++ b/gr-qtgui/lib/freqcontrolpanel.cc
@@ -243,6 +243,8 @@ void FreqControlPanel::setFFTAverage(float val)
 void FreqControlPanel::toggleFFTSize(int val)
 {
     int index = static_cast<int>(round(logf(static_cast<float>(val)) / logf(2.0f))) - 5;
+    index = std::max(index, 0);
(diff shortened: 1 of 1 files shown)


drmpeg (approved):
Works good. Tested on Ubuntu 21.10.

mormj (approved):

willcode:
Can someone put in a feature request PR for cleaning this up, maybe adding more pulldown values? Better ideas?

Issue #5404, opened 2021-12-17
QT GUI Blocks FFT Size Selection


mormj:
### Feature Description

Frequency Sink and QT GUI Sink have some undesirable behavior with FFT size selection:
1) FFT sizes are limited 17 - 16383
2) FFT size dropdown is limiting

### Feature Urgency

medium (would be nice to have in the near future)

Issue #5399, opened 2021-12-16
Segfault with QT GUI Frequency Sink


drmpeg:
### What happened?

A bug happened!

### System Information

OS: Ubuntu 20.04 and 21.10
GR Installation Method: Source

### GNU Radio Version

3.10-git (master)

### Specific Version

3.10.0.0-rc2

### Steps to Reproduce the Problem

Place a QT GUI Frequency Sink is a flow graph. Set the FFT size to 8192 and enable the Control Panel in the Config tab. Run the flow graph

### Relevant log output

```shell
Generating: '/home/re/xfer/vv010.py'

Executing: /usr/bin/python3 -u /home/re/xfer/vv010.py


>>> Done (return code -11)

dmesg log

[Thu Dec 16 02:13:55 2021] python3[389019]: segfault at fffffffc03268140 ip 00007f22fe04ea9a sp 00007ffd07d1afe0 error 5 in libgnuradio-qtgui.so.3.10.0.0-rc2[7f22fdfe0000+115000]
[Thu Dec 16 02:13:55 2021] Code: 01 f3 48 0f 2a c2 0f 2f c1 0f 86 ed da f9 ff 8b 10 f3 44 0f 2c e9 83 fa 01 77 55 48 63 50 08 4d 63 ed be 01 00 00 00 49 01 d5 <4a> 8b 7c e8 10 e8 1c 61 f9 ff 48 89 ef e8 04 92 09 00 44 89 e6 48
```

PR #5413, opened 2021-12-20
qtgui: additional fft sizes

mormj:
# Pull Request Details
Additional FFT sizes for the Frequency Sinks

## Description

Rather than < 16384, make the range on the FFT frequency sink <=16384 but keep the size of the fft dropdown in the control panel the same

## Related Issue

#5404

QT GUI Frequency Blocks

## Checklist

### DIFF
--- a/gr-qtgui/include/gnuradio/qtgui/form_menus.h
+++ b/gr-qtgui/include/gnuradio/qtgui/form_menus.h
@@ -16,6 +16,7 @@
 #include <QtGui/QtGui>
 #include <stdexcept>
 #include <vector>
+#include <math.h>
 
 #if QT_VERSION >= 0x050000
 #include <QtWidgets/QtWidgets>
@@ -606,7 +607,7 @@ public:
             d_act[t]->setActionGroup(d_grp);
         }
 
-        QIntValidator* valid = new QIntValidator(32, 4096, this);
+        QIntValidator* valid = new QIntValidator(32, 16384, this);
         ((OtherAction*)d_act[d_act.size() - 1])->setValidator(valid);
 
         connect(d_act[0], SIGNAL(triggered()), this, SLOT(get05()));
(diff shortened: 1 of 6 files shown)


drmpeg:
Does not fix the freeze with me on Ubuntu 20.04. Also freezes with 16384 now.

mormj:
I think the buffer size might be coming into play here. It only gets set larger if the number of fft points is initialized at the larger value. Perhaps why 4096 was the dynamic limit in the freq control panel.

dl1ksv:
If I start with a fftsize of 8192 , the flowgraph runs and I can switch the fftsize without freezing.
But if I start with fftsize <= 4096 the flowgrap freezes if I change to 8192.

So I share the assumption of @mormj the this issue depends on buffer sizes.

So how to proceed ?
A quick fix could be , to switch back to the old code, but restrict the fftsize to 4096.
This would avoid the crash. But on the other side QT GUI sink supports a fftsize up to 32768 even for the frequency tab.

Bye the way, I think the signal/slot mechanism should be reworked, as the code is hard to maintain.

mormj:
@dl1ksv - I think for now we should just revert it back to 4096 until the whole thing can be reworked.  The GUI Sink widget allows selectable fft size up to 32k, but I haven't dug in how they do it.

drmpeg:
@dl1ksv Just a note. The segfault that I reported in https://github.com/gnuradio/gnuradio/issues/5399 has been fixed by another previous commit. https://github.com/gnuradio/gnuradio/commit/db626c1b23aa47a6a4c1138b405fbe5809d803b5

dl1ksv:
@drmpeg Thank you for this hint. I wasn't aware of this. But If you use this patch , the user gets no warning that his selected fftsize will not be applied but degraded to 4096.

I'd prefer changes in
https://github.com/gnuradio/gnuradio/blob/28b563c77233ea900b6c5f938d7aeb6fd07e38f1/gr-qtgui/lib/freq_sink_c_impl.cc#L144-L148

and 
https://github.com/gnuradio/gnuradio/blob/28b563c77233ea900b6c5f938d7aeb6fd07e38f1/gr-qtgui/lib/freq_sink_f_impl.cc#L144-L148

and set the bound to 8192 instead of 16384.
Then you get at runtime:

  File "/home/schroer/gnuradiocomponents/bugs/crash.py", line 189, in <module>
    main()
  File "/home/schroer/gnuradiocomponents/bugs/crash.py", line 167, in main
    tb = top_block_cls()
  File "/home/schroer/gnuradiocomponents/bugs/crash.py", line 82, in __init__
    self.qtgui_freq_sink_x_0 = qtgui.freq_sink_c(
RuntimeError: freq_sink: FFT size must be > 16 and < 8192.

>>> Done (return code 1)

mormj:
Currently, it works fine to set the fft size to 8192 as long as you don't have the frequency control panel open

willcode:
Probably graphic are not synced. Ran into this with the number display widget yesterday.

drmpeg:
I think we could go with Josh's commit here, but add an INFO message that says something like "FFT size clamped to 4096 when Control Panel is enabled".

Also, I think a check of the FFT size should be added to the Qt GUI Sink. If it's > 32768, you get this error on exit:
```
double free or corruption (out)

>>> Done (return code -6)
```
It's just a little code, so could be tagged on to this PR.

mormj:
Closing in favor of #5442

Issue #5411, opened 2021-12-19
v3.8 gr-vocoder does not build on (my) Fedora 35


willcode:
### What happened?

Building maint-3.8 on Fedora 35 fails at ` Swig source vocoder_swig.i`

### System Information

OS: Fedora 35
GR Installation Method: Source (PyBOMBS)

### GNU Radio Version

3.8 (maint-3.8)

### Specific Version

maint-3.8 (at v3.8.5.0-rc1)

### Steps to Reproduce the Problem

Build with default settings. Disabling gr-vocoder results in a good build.

### Relevant log output

```shell
[ 97%] Swig source vocoder_swig.i
/usr/include/codec2/freedv_api.h:148: Error: Syntax error in input(1).
make[2]: *** [gr-vocoder/swig/CMakeFiles/vocoder_swig_swig_compilation.dir/build.make:77: gr-vocoder/swig/CMakeFiles/vocoder_swig.dir/vocoder_swigPYTHON.stamp] Error 1
make[2]: *** Deleting file 'gr-vocoder/swig/CMakeFiles/vocoder_swig.dir/vocoder_swigPYTHON.stamp'
make[1]: *** [CMakeFiles/Makefile2:10181: gr-vocoder/swig/CMakeFiles/vocoder_swig_swig_compilation.dir/all] Error 2
make: *** [Makefile:146: all] Error 2
```
```

**Comment by willcode** [comment]
I don't see any related changes in maint-3.8, all the way back to the previous release. CI builds are good. I'm pretty sure I got a good build off a recent version. Mystery.

**Comment by willcode** [comment]
F35 upgraded codec2 from 1.0.0 to 1.0.1 recently. Possibly related.

**Comment by willcode** [comment]
According to Maitland, there is something that needs fixing with codec2 v1.0.1, and Debian patches it.

**Comment by marcusmueller** [comment]
Can confirm that fedora35 codec2 is unpatched; but line 148 in that file is:
https://github.com/drowe67/codec2/blob/v1.0.1/src/freedv_api.h#L148:

```c
typedef void (*freedv_callback_rx_sym)(void *, _Complex float, float);
```

Don't think that's a syntax error (it *is* C in its maximum beauty, though).

So, being on a train with bad internet, can't run a F35 container build locally. I'll work under the "SWIG being weird *again*" hypothesis.

So, foo.h:

```c
#include <complex.h>
typedef void (*freedv_callback_rx_sym)(void *, _Complex float, float);
```
[...]

PR #5416, opened 2021-12-26
Vocoder: SWIG needs a workaround for dealing with C99 keyword _Complex

marcusmueller:
`_Complex` (actual C99 keyword) is not supported by SWIG, `complex` is.

It is not unlikely this will break later on when SWIG starts behaving
standards-compliant.

This became an error when codec2 1.0.1 started exposing API with
_Complex.

Signed-off-by: Marcus Müller <marcus@hostalia.de-

# Pull Request Details

## Related Issue

#5411  

Vocoder SWIG  build only on maint-3.8, sadly might lead to incompatibilities in the future

## Testing Done

None, since no build machine at hand

## Checklist

### DIFF
--- a/gr-vocoder/swig/vocoder_swig.i
+++ b/gr-vocoder/swig/vocoder_swig.i
@@ -86,6 +86,49 @@ GR_SWIG_BLOCK_MAGIC2(vocoder, codec2_encode_sp);
 #endif
 
 #ifdef LIBCODEC2_HAS_FREEDV_API
+// == SWIG does not like standards ==
+// SWIG (as of SWIG 4.0.2) emulates C99 complex number support: C99 has a keyword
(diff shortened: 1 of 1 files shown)


marcusmueller:
seemingly doesn't fix it.

marcusmueller:
This now makes it build within my F35 container without patching codec2 itself.

willcode (approved):
Builds and passes QA on F35. This is pretty wild, but since the fix does not seem to propagate further than the swig file, whatever works!

marcusmueller:
> Builds and passes QA on F35. This is pretty wild, but since the fix does not seem to propagate further than the swig file, whatever works!

"pretty wild" is a pretty good description of this :) Good news is that D Rowe has already a fix for the versioning info in the pipeline, so that this will really only affect a single patchlevel release of codec2 ever

willcode:
Merge after v3.8.5.0.

Issue #5415, opened 2021-12-26
grc: make it clearer why the project cannot be executed


d33tah:
### Feature Description

Thanks for creating gnuradio! I recently became interested in signal processing and while I felt that I became acquainted with the basics, I decided to give Gnuradio Companion a try once again. Previously I found the UX overwhelming and this time I decided to spend more time trying to understand it. In doing that, I understood one of the problems I might have had. Basically, I tried to create a "hello, world" project. I started GRC, added random source, audio sink, wired them together, then tried to run the prototype... and found the "execute" button grayed out.

My first assumption was that the program was for some reason broken on my system and started Googling, but it's not easy to Google for UX problems. There weren't many hits for "gnuradio companion execute grayed out" or "disabled". It took some experimenting with the GUI to find view->flowgraph errors and then some more to understand the cryptic "ID "default" is blacklisted" error message (which was there because there's no project name set by default). I think that it's an easy problem to solve, here's my suggestion:

Instead of graying "execute" out with no feedback, either add a hint on the button and menu item (something saying "problems found - check view->flowgraph errors for suggestions") or straight up enable the user to press execute and then display the actual list of errors, with a red caption explaining why execution failed.

What do you think of it? I think that it's not necessarily difficult to implement and would simplify things for newcomers. I for one would have appreciate it.

### Feature Urgency

high (impacts broad portion of ecosystem)

### More Information

I marked this issue as high urgency because I believe that lack of this feature significantly increases barrier to entry for new users.

777arc:
Does the default ID have to be blacklisted?  It wasn't always, and there is something to say for the convenience of being able to quickly open a new flowgraph and test something without setting the ID.  What are we trying to protect by blacklisting it?  Is it just to help enforce good practice?  Also what if we let the ID be empty and then if it's empty we just use the name of the grc file?

mormj:
This has been addressed with #5421 - the flowgraph no longer starts in an "errored" state, and the id is set to whatever the flowgraph is saved as if it is left as "default".

PR #5421, opened 2021-12-27
grc: change default flowgraph id

mormj:
# Pull Request Details
This is a stopgap fix to help with the UX by reverting to the old behavior where the flowgraph is titled "top_block" (but doesn't really matter) - but it is not immediately in a state of error.  

There is likely a better UX solution for naming and saving flowgraphs, but hopefully this removes some of the frustration to a newly created flowgraph. 

Definitely open to ideas for doing something different here, but want to get *something* into 3.10.

## Related Issue
#5415

grc

## Testing Done
Opened new flowgraph and ran without having to change the name or the block being red.

## Checklist

### DIFF
--- a/grc/core/Constants.py
+++ b/grc/core/Constants.py
@@ -20,6 +20,7 @@ DATA_DIR = os.path.dirname(__file__)
 BLOCK_DTD = os.path.join(DATA_DIR, 'block.dtd')
 DEFAULT_FLOW_GRAPH = os.path.join(DATA_DIR, 'default_flow_graph.grc')
 DEFAULT_HIER_BLOCK_LIB_DIR = os.path.expanduser('~/.grc_gnuradio')
+DEFAULT_FLOW_GRAPH_ID = 'default'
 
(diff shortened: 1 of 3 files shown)


marcusmueller (approved):

dl1ksv (commented):
Of course, this patch is OK, but leads back to a behaviour we had about two years ago.
Not forcing to change the id bears the risk of overriding an existing generated flowgraph.
To avoid this perhaps we should check , if topblock.py  already exists.

mormj:
@marcusmueller - i like the suggestion.  Not sure the best way to do that in the grc code though.  In this case I would keep the default fg name as "default".  When saving the flowgraph as {filename}.grc, if it is still equal to "default" in the id, then swap the id with {filename}.  

Detecting a conflict of the default python file could get annoying

mormj:
@marcusmueller - I think by just removing 'default' from the blacklist - it forces you to save the file as something else when you first run, and then swaps the default id with the sanitized filename.  This should take away the major barrier of "start in exception state" without nagging about replacing the generated python file.  I would think the majority of grc users don't care about the generated python file, so it is not worth changing existing behavior that overwrites the file as it is.

PR #5422, opened 2021-12-27
Modernize C++

marcusmueller:
# Pull Request Details

## Description

Three things:

* using range-based for loops, especially for iterating through tags (this is a pattern I'd like to establish)
* using `string == "value"` instead of `string.compare("value")` (guess this is a C-ism)
* consistent `override`
* assert.h -> cassert
* ruthlessness and surprise

Seemed to be a good time to do this, what with kicking of a "fresh" master

all over the tree - motivated by the string comparisons in gr-iio.

## Testing Done

Testsuite passes. No new tests written.

## Checklist


willcode:
Would be good to backport to 3.9. Much of the patch applies as is, some tweaking will be needed.

gnuradio-runtime/lib/math/qa_fxpt_vco.cc
```diff
@@ -38,15 +38,15 @@ BOOST_AUTO_TEST_CASE(t0)
         input[i] = sin(double(i));
     }
 
-    for (int i = 0; i < SIN_COS_BLOCK_SIZE; i++) {
+    for (float i : input) {
```
ThomasHabets, line 41: Pretty poor name for the float, now.

Maybe just `in`?

gnuradio-runtime/lib/top_block_impl.cc
```diff
@@ -50,10 +50,9 @@ make_scheduler(flat_flowgraph_sptr ffg, int max_noutput_items, bool catch_except
         if (!v)
             factory = scheduler_table[0].f; // use default
         else {
-            for (size_t i = 0; i < sizeof(scheduler_table) / sizeof(scheduler_table[0]);
-                 i++) {
-                if (strcmp(v, scheduler_table[i].name) == 0) {
-                    factory = scheduler_table[i].f;
+            for (auto& i : scheduler_table) {
```
ThomasHabets, line 53: Not a great name here either. But more generally: Many (most, probably) of these should live with getting a `const` for readability.

gr-blocks/lib/rotator_cc_impl.h
```diff
@@ -23,7 +23,7 @@ struct phase_inc_update_t {
     double phase_inc;
 };
 
-bool cmp_phase_inc_update_offset(phase_inc_update_t lhs, phase_inc_update_t rhs)
+inline bool cmp_phase_inc_update_offset(phase_inc_update_t lhs, phase_inc_update_t rhs)
```
ThomasHabets, line 26: Unrelated?

gr-filter/lib/qa_firdes.cc
```diff
@@ -21,8 +21,8 @@ using std::vector;
 void print_taps(std::ostream& s, vector<float>& v)
 {
     std::streamsize tmp = s.precision(9);
-    for (unsigned int i = 0; i < v.size(); i++) {
-        s << v[i] << ", ";
+    for (float i : v) {
```
ThomasHabets, line 24: Again a float named i.

gr-iio/lib/attr_sink_impl.h
```diff
@@ -40,7 +40,7 @@ class attr_sink_impl : public attr_sink
                    const std::string& channel,
                    attr_type_t type,
                    bool output);
-    ~attr_sink_impl();
+    ~attr_sink_impl() override;
```
ThomasHabets, line 43: Should be separate PR?

ThomasHabets: (also other instances)

ThomasHabets: Never mind, same PR different commit is good.

gnuradio-runtime/lib/math/qa_fxpt_vco.cc
marcusmueller: agreed. This is just the autoconversion that clang-tidy does. When the original variable was short and the container ends in an "s", the clang-tidy pass tries to guess a singular and uses that (eg `for(int i = 0; i<inputs.size(); ++i)`->`for(auto& input : inputs)`), but here there's no plural-container :)

gnuradio-runtime/lib/top_block_impl.cc
marcusmueller: yeah the heuristic with which the `const`ness is determined is really rather ... conservative. At least it doesn't do any copies as far as I looked through it.

gr-filter/lib/qa_firdes.cc
marcusmueller: yeah, will fix.

marcusmueller:
@ThomasHabets fixed the readability issues where arisen from range-based for-conversion; drive-by modernized other things, as well.

gr-dtv/lib/dvb/dvb_bbscrambler_bb_impl.cc
```diff
@@ -31,210 +32,12 @@ dvb_bbscrambler_bb_impl::dvb_bbscrambler_bb_impl(dvb_standard_t standard,
                                                  dvb_code_rate_t rate)
     : gr::sync_block("dvb_bbscrambler_bb",
                      gr::io_signature::make(1, 1, sizeof(unsigned char)),
-                     gr::io_signature::make(1, 1, sizeof(unsigned char)))
+                     gr::io_signature::make(1, 1, sizeof(unsigned char))),
+      kbch(dvb::constants::kbch_mapping.at(framesize).at(rate)),
```
marcusmueller, line 36: now we just need to do that everywhere and we get very compact source files :D

gr-dtv/lib/atsc/atsc_equalizer_impl.cc
```diff
@@ -27,42 +28,46 @@ atsc_equalizer::sptr atsc_equalizer::make()
     return gnuradio::make_block_sptr<atsc_equalizer_impl>();
 }
 
-static float bin_map(int bit) { return bit ? +5 : -5; }
+constexpr float bin_map(unsigned char bit) { return bit ? +5 : -5; }
 
-static void init_field_sync_common(float* p, int mask)
+template <size_t N>
+constexpr std::array<float, N> init_field_sync_common(bool invert)
```
ThomasHabets, line 34: `std::array` does not have move assignment/constructor, thus every call here will cause `N*4` bytes to be copied (unless RVO). Is that the intention?

ThomasHabets (commented):
Only a couple of outstdanding comments.

gr-digital/lib/ofdm_carrier_allocator_cvc_impl.cc
```diff
@@ -143,8 +143,8 @@ int ofdm_carrier_allocator_cvc_impl::work(int noutput_items,
 
     memset((void*)out, 0x00, sizeof(gr_complex) * d_fft_len * noutput_items);
     // Copy Sync word
-    for (unsigned i = 0; i < d_sync_words.size(); i++) {
-        memcpy((void*)out, (void*)&d_sync_words[i][0], sizeof(gr_complex) * d_fft_len);
+    for (const auto& sync_word : d_sync_words) {
+        memcpy((void*)out, (void*)&sync_word[0], sizeof(gr_complex) * d_fft_len);
```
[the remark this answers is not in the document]
marcusmueller: true! Intuitively wanted to opt for `std::copy_n`, as we have the FFT length, and my *guess* is that a contiguous memory copy is more efficient than having to check for `!= end` after every item. However, `std::copy` has a specialization that seems to call memcpy under the hood; and it seems to kick on current compilers.

gr-dtv/lib/atsc/atsc_equalizer_impl.cc
ThomasHabets: `std::array` is stack (well… technically "auto storage duration") allocated (including the data), so cannot even in theory do a pointer swap for move. It can do per-element move of all its elements, but for floats that's not an optimization.

Your godbolt example shows two differences:
1. it likely triggers RVO, which always means no copy nor move. (I don't think it's [mandatory RVO](https://en.cppreference.com/w/cpp/language/copy_elision))
2. the function is inlined, so not really returned.

[Adding `-fno-inline` takes care of the second, and making RVO impossible](https://gcc.godbolt.org/z/6GzGcn9eo) (for example by calling a second time) causes a call to memcpy (or [in this example](https://gcc.godbolt.org/z/61n6x7rfe) a `rep movsq`).

So if the only correct way to call this function is in the mandatory RVO situations? That may be a bit of a foot gun for non-initialization at caller site. And I don't think as-is this is mandatory RVO since it matches the non-mandatory's section "In a return statement, when the operand is the name of a non-volatile object with automatic storage duration".

I guess since this is constexpr it may actually be mandatory RVO, but still only applies when caller makes it possible (in initialization).

As for `std::move`, despite its name it doesn't move. It just converts lvalue to rvalue, and you can copy an rvalue. But also yes technically `std::array` can be moved, but it's just an elementwise move, which for array of float means a de facto copy.

Also examples like [this](https://gcc.godbolt.org/z/fq8YovMed) show that `std::move()` doesn't force a move, only enables it.

But yes, technically I was wrong by saying `std::array` doesn't "have" a move assign/constructor. It does, it just doesn't do the expected thing, because it can't.

marcusmueller: Oh, wow! First of all: thanks for letting me occupy you so much! I'm thankful, and I'm learning a lot :)

So, the right approach here is instead letting `init_field_sync_common` return a `std::vector` value, making `training_sequence1,2` std::vectors and using `std::vector`s (actually moving) move c'tor in through `training_sequence1(std::move(init_field_sync_common<…>(false))` in the initializer list?

ThomasHabets: Maybe. But at the cost of one indirection when the sequences are accessed.

I'd say there are three options:

1. As this PR currently does: Depending on compiler and options could do a needless copy, but the training data is inlined with the impl object so no indirection.
2. Switch to `std::vector`. No copies on init, but an indirection on access to the data, as `vector`s don't have a "small string optimization".
3. Call init functions, as before, but now with `std::array` instead of C arrays. No performance cost, as init is done in-place, and data inlined. Only downside is that it's a shame to not use the neatness of initializer list.

I've not played with dtv, but I suspect the important bit is to make the data path fast, which means (1) or (3). I'd lean towards (3).

marcusmueller: opted for (3)!

gr-digital/lib/constellation.cc
```diff
@@ -353,21 +355,16 @@ std::vector<float> constellation::soft_decision_maker(gr_complex sample)
...
-            d_im_max = d_constellation[i].imag();
-
-        if (d_constellation[i].real() < d_re_min)
-            d_re_min = d_constellation[i].real();
-        if (d_constellation[i].imag() < d_im_min)
-            d_im_min = d_constellation[i].imag();
+    // Find min/max of constellation for both real and imag, axes.
+    d_re_min = std::numeric_limits<decltype(d_re_min)>::max();
+    d_im_min = std::numeric_limits<decltype(d_re_min)>::max();
+    d_re_max = -d_re_min;
```
ThomasHabets, line 361: I think it's safer and more correct to use `std::numeric_limits<decltype(d_re_max)>::lowest()`.

gr-digital/lib/ofdm_carrier_allocator_cvc_impl.cc
```diff
@@ -137,36 +134,38 @@ int ofdm_carrier_allocator_cvc_impl::work(int noutput_items,
...
+    auto input_cnt = static_cast<std::size_t>(ninput_items[0]);
+    auto end_pointer = out + d_fft_len * noutput_items;
     std::vector<tag_t> tags;
 
-    memset((void*)out, 0x00, sizeof(gr_complex) * d_fft_len * noutput_items);
     // Copy Sync word
-    for (unsigned i = 0; i < d_sync_words.size(); i++) {
-        memcpy((void*)out, (void*)&d_sync_words[i][0], sizeof(gr_complex) * d_fft_len);
+    for (const auto& sync_word : d_sync_words) {
+        std::copy(sync_word.cbegin(), sync_word.end(), out);
```
ThomasHabets, line 145: Looks a bit odd to do `.cbegin()` with `.end()`. `.cend()`?

ThomasHabets (approved):
Aside from those two things: Looks good

gnuradio-runtime/lib/flat_flowgraph.cc
```diff
@@ -315,12 +313,12 @@ void flat_flowgraph::merge_connections(flat_flowgraph_sptr old_ffg)
     }
 
     // Now connect inputs to outputs, reusing old buffer readers if they exist
-    for (basic_block_viter_t p = d_blocks.begin(); p != d_blocks.end(); p++) {
-        block_sptr block = cast_to_block_sptr(*p);
+    for (auto& d_block : d_blocks) {
```
mbr0wn, line 313: The loop variable should not start with `d_` as per our coding guidelines.

gnuradio-runtime/lib/flat_flowgraph.cc
```diff
@@ -413,27 +411,27 @@ void flat_flowgraph::setup_buffer_alignment(block_sptr block)
 std::string flat_flowgraph::edge_list()
 {
     std::stringstream s;
-    for (edge_viter_t e = d_edges.begin(); e != d_edges.end(); e++)
-        s << (*e) << std::endl;
+    for (auto& d_edge : d_edges)
```
mbr0wn, line 411: Same comment (`d_`)

mbr0wn:
> ruthlessness and surprise

Surprises... aren't a great thing to have in a PR.

gnuradio-runtime/include/gnuradio/block.h
```diff
@@ -545,9 +545,9 @@ class GR_RUNTIME_API block : public basic_block
      * \brief Allocate the block_detail and necessary output buffers for this
      * block.
      */
-    void allocate_detail(int ninputs,
-                         int noutputs,
-                         const std::vector<int>& downstream_max_nitems_vec,
+    void allocate_detail(unsigned int ninputs,
```
mbr0wn, line 548: `unsigned int` often bothers me, reminds me of C of yore. Wouldn't `size_t` be the more appropriate type?

gnuradio-runtime/include/gnuradio/block.h
```diff
@@ -545,9 +545,9 @@ class GR_RUNTIME_API block : public basic_block
...
+                         unsigned int noutputs,
+                         const std::vector<unsigned int>& downstream_max_nitems_vec,
```
mbr0wn, line 550: Here, for example, I know it's preference, but I don't like that space in the template param.

mbr0wn (commented):
Can you maybe split this up? Some of these commits aren't even truly modernizing (like the extra inlining). I'm very glad people like you and @ThomasHabets take on the work of de-crufting and modernizing our codebase, but this is pretty big and might be easier to review and merge if you put up multiple PRs.

mormj:
Love the changes, but agree with @mbr0wn on splitting this up - many good changes, but it is tough to review.  Also concerned that the extent of the changes would be questionable to backport - and not backporting would cause an unnecessary code diversion

Closing for now - I think smaller PRs would help alleviate the concerns

PR #5424, opened 2021-12-27
modernization: use using= instead of typedef

marcusmueller:
# Pull Request Details

## Description

It started out with me innocently thinking, hey, let's have `using` in the modtool templating, then I thought it'd be better if that was consistent with the tree, then I checked how consistent the tree was.

Then I ran clang-tidy on the tree, then I fixed what that broke (interestingly, clang-tidy sometimes breaks on std::map and most of the
time with function pointers), and then I made this PR

Quite everything, I daresay.

## Checklist


willcode:
Backport to 3.9 would be useful to keep code similar - we'll see if this is possibe.

ThomasHabets (approved):
\o/

mbr0wn:
Personally, I prefer `using` over `typedef` in C++ (and rarely use `typedef` at all these days). This change is fine. However, I would like to caution us putting work towards these types of cleanup. They are hard not to justify as busywork. Besides, we're not fixing any bug other than people's preferences. `typedef` is perfectly valid C++, and doesn't remove a Boost dependency or anything like that. Keep in mind that if you put up PRs like this, there's always a risk that something breaks, and you suck up some reviewer's time, plus all the people who need to rebase their PRs.

marcusmueller:
Agreed, these are big changesets with (ideally, hopefully, very much) no functional change. I'm painfully aware of the time that goes into these.

The actual syntactic changes were that I had to replace a few `typedef enum {} name;` simply with a freestanding enum (or struct).

What this attempts to do is very similar to why we do `clang-format`, and why we decided to do it in one fell swooooop back then; consistency¹, and only a single pain point. 

I'm really not upset if anyone says "nope, this is too bad on the effort/benefit scale, let's not do this anytime soon" :) Happy if I'm not putting workload on maintenance! The point in time was intentionally chosen (low number of open PRs, mergeable shortly after a release on master without no obligation to backport, though I must admit I forgot how active Jeff is – I just didn't assume he would *want* to do that... so that's a misestimate on my end).

----
¹ being 100% honest, the motivation chain runs a bit deeper: I [added](https://github.com/gnuradio/gnuradio/commit/9d01e324c995ba31bddf2909d86f370fdb78bc53) the input/output types as member types to the modtool block template, which I back then didn't think about and just did `using`. Lately, I worked through a few blocks to fix for-loops old-school index- or iterator-walking through vectors. By converting them to range-based for loops, I noticed I obsoleted some of the `typedefs` (and later decided to attack that loop problem whole-tree). One thing led to another… and basically, if one converts the codebase to `using`, one can let clang-tidy remove the unused former typedefs (not done here to avoid making the changes less "all the same" to review). Consistency with newer code already using `using` is not an afterthought, but it's not how I ended up wanting this.

mormj:
I like these types of syntactical changes - but agree with Martin that it might make sense to hold off until a point where we don't have to maintain backports of them.  Flagged this with GR 4.0 - would be good to accumulate all the good coding practices as we migrate to 4.0 and "start fresh" there.  

There is an in-progress wiki page to summarized features for 4.0 here:
https://wiki.gnuradio.org/index.php?title=GNU_Radio_4.0_Summary_of_Proposed_Features#Coding_Standard
Part of that should be (placeholder now) changes to the coding standard we want to enforce - things that are hard to enforce given the pervasiveness across the current gr codebase

mormj:
Closing as I think we should defer these types of changes to 4.0, or do them in smaller PRs.

PR #5433, opened 2021-12-29
CMake: the Reformatting

marcusmueller:
# Pull Request Details

We have automated formatting, enforced for C++. We have PEP8 for Python.

Why can't we have some assistance with the least beautiful language in tree?

Enter [cmake-format](https://cmake-format.readthedocs.io/en/latest/), which does exactly that. Only needed to write a short config file (actually, a Python module), a minimal amount of glue to bind a few settings in that to what's in .clang-format, and then run

```
cmake-format -c .cmake-format.py \
             -i \
             cmake/cmake_uninstall.cmake.in cmake/Modules/* \
             cmake/Toolchains/arm_cortex_a8_native.cmake cmake/Toolchains/oe-sdk_cross.cmake 
             $(git ls-files | grep CMakeLists.txt )
```

CMake.

## Testing Done

Still builds like a champ!

I don't want to enforce running this on every commit – I have no idea how
stable cmakelang is medium-term. We could add this to a pre-commit hook,
together with flake8 and clang-formatting, though.

## Checklist


mormj:
Are you suggesting this for 3.10 or start this with 3.11?

willcode:
It this all whitespace mods, or does it do re-ordering and other changes?

marcusmueller:
> Are you suggesting this for 3.10 or start this with 3.11?

@mormj I'd be suggesting this for 3.11; not a functional change, though.

> It this all whitespace mods, or does it do re-ordering and other changes?

Whitespace, indents, line breaks, consistent capitalizations (`builtin_function(KEYWORD OthEr_stuff)`); no reordering. It could have reordered argument lists where that's permitted, but we have no instance of that, far as I could find in the tree.

marcusmueller:
Noticed that setting the indent depth to 2 instead of 4 would possibly yield a smaller diff. Must investigate.

CMakeLists.txt
```diff
@@ -42,10 +44,10 @@ message(STATUS "Build date set to ${BUILD_DATE}.")
 include(GrComponent)
 
 # Set the version information here
-SET(VERSION_MAJOR 3)
-SET(VERSION_API   10)
-SET(VERSION_ABI   0)
-SET(VERSION_PATCH 0-rc2)
+set(VERSION_MAJOR 3)
+set(VERSION_API 10)
```
mbr0wn, line 48: I actually preferred the old indenting here.

mbr0wn (approved):
Much nicer.

CMakeLists.txt
marcusmueller: agreed, don't know whether cmake-lang supports that, though. Haven't found an option, so [asking on SO](https://stackoverflow.com/questions/70597446/keep-multiple-space-aligned-setvariable-value-the-way-they-were-during-cmake).

mormj:
Holding off a bit more on this until we are done with emergency patch releases for 3.10.  Maybe should be merged closer to the 3.10.1.0 window

CMakeLists.txt
mbr0wn: Isn't this all you need? https://cmake-format.readthedocs.io/en/stable/format-features.html?highlight=disable#disable-formatting-locally

marcusmueller: Yep! As [someone helpfully pointed out as well](https://stackoverflow.com/questions/70597446/keep-multiple-space-aligned-setvariable-value-the-way-they-were-during-cmake/70846835#70846835) :)

willcode:
Close in favor of #6100?

mormj:
> Close in favor of #6100?

Yes.  6100 has the main commit, then the aftermath of running the script again on the latest `main`

PR #5434, opened 2021-12-30
Typed sync blocks

marcusmueller:
# Pull Request Details

## Description
Many blocks are sync and almost all blocks have known IO types at instantiation. This block type tries to make use of that.

The benefits are:

1. type-safe `work`
2. Less boilerplate code

For this to work out, it was first necessary to make access to the in- and output streams type safe.
So, stream and item iterators had to be implemented, which encompasses three things:

1. `stream_span`, which lets one iterate through individual input streams
2. `item_span`, which lets one iterate through individual items in a stream (so, is pretty close to std::span, but way less powerful;
   sadly, it's not C++20'o'clock yet.)
3. `stream_iter`, which is an internal type of `stream_span`, which creates `item_span`s on the fly for the raw counters it's been given.
4. 

Introduces a new block type (to compete with `gr::sync_block`) to `gnuradio-runtime`.

## Testing Done

Implemented basic C++ QA tests. Seems to work

## Checklist

- [ ] I have updated [the documentation](https://wiki.gnuradio.org/index.php/Main_Page#Documentation) where necessary. :exclamation: 

## Intent

The adoption of new buffer interfaces, and all forward-looking scheduler developments in general, are aiming at a cleaner interface towards the consumer. So, this is something that I hope makes sense in that context – **definitely** not something I think should be in 3.10 (in fact, it's probably better if it isn't in a release anytime soon, before it's undergone some real-world usage).


gnuradio-runtime/lib/qa_typed_sync_block.cc
```diff
@@ -0,0 +1,116 @@
...
+        _left -= to_copy;
+        return { to_copy, !(_left) };
+    }
+    vector_snk(size_t total = 100000)
+        : typed_sink<value_t>("typed_sink", 1, 1, 0, 0), _left(total)
+    {
+        _vector.reserve(total);
+    }
+    const std::vector<value_t>& data() const { return _vector; }
+};
```
marcusmueller, line 91: Just to highlight how `vector_sink` looks in `typed_sync_block`

gnuradio-runtime/lib/qa_typed_sync_block.cc
```diff
@@ -0,0 +1,116 @@
...
+    {
+        for (const auto& output_stream_span : output_streams) {
+            for (auto& item : output_stream_span) {
+                item = _counter++ & ((1 << 12) - 1);
+            }
+        }
+        return { n_items };
+    }
+    int_generator() : int_src("int_generator", 0, 0) {}
+};
```
marcusmueller, line 59: Just for illustration: this is a simple `int`-producing source, in its entirety.

mormj:
@marcusmueller - I like this idea, but going to close it for now.  Similar functionality is planned in dev-4.0, so I would like to add effort there instead of introducing a new series of blocks in 3.x.

PR #5440, opened 2021-12-31
blocks: Remove Fast Multiply Const (just an alias for multiply_const)

marcusmueller:
# Pull Request Details

block_multiply_const_vxx.block.yml contains the same functionality + vector abilities.

Since removal breaks the examples using that block, updated them to use _vxx before, so we have no non-working commits in the history.

## Related Issue

#5439 is the maint-3.10 deprecation target for this

Only gr-dtv used the block in-tree.

## Testing Done

After the change, the gr-dtv examples still generate and run.

## Checklist


drmpeg:
There is a functionality difference between the two blocks. The Fast Multiply Const block uses a scalar constant even when the pins are vectors.

marcusmueller:
@drmpeg oh! Now I'm honestly confused; will need to check that out, but the make line is very similar for both GRC files:

A=Fast
B="normal"; 
`type.fcn` is cc, ff, ii, ss …

```diff
<<<<<<< variant A
    make: blocks.multiply_const_${type.fcn}(${const}, ${vlen})
>>>>>>> variant B
    make: blocks.multiply_const_${ 'v' if context.get('vlen')() > 1 else '' }${type.fcn}(${const})
======= end
```

So, I had to re-check the two `_impl.cc` and you're right, now I have a stinging feeling in my forehead!

PR #5442, opened 2022-01-02
gr-qtgui: Frequency sink freezes the running flowgraph

dl1ksv:
---

# Pull Request Details
With #5405 the segfault of freq_sink_c was fixed, but
if you enable the control panel and change the FFT size to 8192
while the flow graph is running, the whole flowgraph freezes.

## Description
This patch:

- fixes the freeze
- extends the fftsize to 32768 for freq_sink_c and freq_sink_f
- changes the entry for fftsize in the corresponding yml files from integer to enum
  and provides a drop down menu with those values that are provided by the gui
  in the running flowgraph.

The code provides code for setting the fftsize in form_menus.h and in freqcontrolpanel.cc
If you change some signal and slots you can do without the code from form_menus.h.
So this part is removed.

freq_sink_c, freq_sink_f
## Testing Done

Flowgraph:
signal source -> freq_sink 
                      ->gui_sink
While running, change fftsize to each availabe size. No freeze at all.

## Checklist

### DIFF
--- a/gr-qtgui/grc/qtgui_freq_sink_x.block.yml
+++ b/gr-qtgui/grc/qtgui_freq_sink_x.block.yml
@@ -20,8 +20,9 @@ parameters:
     hide: ${ ('none' if len(name) > 0 else 'part') }
 -   id: fftsize
     label: FFT Size
-    dtype: int
+    dtype: enum
     default: '1024'
+    options: ['32','64','128','256','512','1024','2048','4096','8192','16384','32768']
     hide: ${ ('all' if (type == 'msg_complex' or type == 'msg_float') else 'none') }
 -   id: freqhalf
     label: Spectrum Width
--- a/gr-qtgui/grc/qtgui_freq_sink_x.block.yml
+++ b/gr-qtgui/grc/qtgui_freq_sink_x.block.yml
@@ -397,6 +397,9 @@ outputs:
     optional: true
     hide: ${ not showports }
 
+asserts:
+- ${fftsize >= 32 and fftsize <= 32768}
+
 templates:
     imports: |-
         from PyQt5 import Qt
--- a/gr-qtgui/grc/qtgui_freq_sink_x.block.yml
+++ b/gr-qtgui/grc/qtgui_freq_sink_x.block.yml
@@ -20,7 +20,7 @@ parameters:
     hide: ${ ('none' if len(name) > 0 else 'part') }
 -   id: fftsize
     label: FFT Size
-    dtype: enum
+    dtype: int
     default: '1024'
     options: ['32','64','128','256','512','1024','2048','4096','8192','16384','32768']
     hide: ${ ('all' if (type == 'msg_complex' or type == 'msg_float') else 'none') }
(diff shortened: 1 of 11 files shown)


mormj (approved):
This definitely simplifies the freq sink block - I would hope that the dropdown is not problematic for anyone that is changing the parameter programmatically.  But the fixed range simplifies the logic.  This tests out as advertised.  LGTM

gr-qtgui/lib/freq_sink_c_impl.cc
```diff
@@ -129,7 +129,7 @@ void freq_sink_c_impl::initialize()
     if (!d_name.empty())
         set_title(d_name);
 
-    set_output_multiple(d_fftsize);
+    set_output_multiple(32768);
```
mbr0wn, line 132: Can we please have constants somewhere and reuse them?

gr-qtgui/grc/qtgui_freq_sink_x.block.yml
```diff
@@ -20,8 +20,9 @@ parameters:
     hide: ${ ('none' if len(name) > 0 else 'part') }
 -   id: fftsize
     label: FFT Size
-    dtype: int
+    dtype: enum
```
mormj, line 23: If you keep these as int, the dropdown persists, but the value can also be set to a variable

drmpeg (approved):
Tests good here.

dl1ksv:
@mormj my patch nearly includes #5446, but here you can change the const's for the fft range and the values of the combobox are calculated. You don't have to edit the list.

mormj:
Your solution here is much better - mine was a quick hack.

drmpeg:
The error message is a little ugly. It should have a space after "be" and no spaces around the line feed.

Change:
```"FFT size must be>= {} and <= {}. \n So falling back to {}."```
to:
```"FFT size must be >= {} and <= {}.\nSo falling back to {}."
in both ```freq_sink_f_impl.cc and freq_sink_c.impl.cc

drmpeg:
Also, the FFT size needs to be checked in the Qt GUI Sink. Values over 32768 cause this error on exit.

double free or corruption (out)

>>> Done (return code -6)

drmpeg:
Is that Is that ${fftsize >= 32 and fftsize <= 32768} assert in the correct file? Shouldn't it be ```qtgui_sink_x.block.yml```?

mormj:
The freq range was definitely change in freq sink, and this was made to match it:
https://github.com/gnuradio/gnuradio/pull/5442/files#diff-2d20922160ac6b37e729922683c1b8f2eec1dfdf2822b465ee6497e31ef5028aR144

drmpeg:
I thought you were trying to fix my previous comment about the Qt Sink block.

mormj:
No, I didn't touch that - was trying to get the last small pieces of the freq sink in.

mormj:
It's worth doing but wanted to go ahead and get everything else merged first

drmpeg:
Won't that assert make all that code dead?

mormj:
Only if you're strictly in GRC, in which case it is nice to catch it earlier and get your red text.

drmpeg:
Ok.

PR #5443, opened 2022-01-03
network: remove unused d_setlock guards

marcusmueller:
There are no protected accessors in these blocks, and stop() can never be called while work() is running.

# Pull Request Details

## Description

Found that `d_setlock` was guarded in work() – but never actually used anywhere else. Since acquiring a lock is relatively costly, removed that.

UDP sink & source, TCP sink.

## Checklist

### DIFF
--- a/gr-network/lib/tcp_sink_impl.cc
+++ b/gr-network/lib/tcp_sink_impl.cc
@@ -228,8 +228,6 @@ int tcp_sink_impl::work(int noutput_items,
                         gr_vector_const_void_star& input_items,
                         gr_vector_void_star& output_items)
 {
-    gr::thread::scoped_lock guard(d_setlock);
-
     if (!d_connected)
         return noutput_items;
 
--- a/gr-network/lib/udp_sink_impl.cc
+++ b/gr-network/lib/udp_sink_impl.cc
@@ -167,8 +167,6 @@ udp_sink_impl::~udp_sink_impl() { stop(); }
 bool udp_sink_impl::stop()
 {
     if (d_udpsocket) {
-        gr::thread::scoped_lock guard(d_setlock);
-
         if (b_send_eof) {
             // Send a few zero-length packets to signal receiver we are done
             boost::array<char, 0> send_buf;
@@ -220,8 +218,6 @@ int udp_sink_impl::work(int noutput_items,
                         gr_vector_const_void_star& input_items,
                         gr_vector_void_star& output_items)
 {
-    gr::thread::scoped_lock guard(d_setlock);
-
     long num_bytes_to_transmit = noutput_items * d_block_size;
     const char* in = (const char*)input_items[0];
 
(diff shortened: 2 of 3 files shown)


ghostop14:
Hi @marcusmueller !  As long as there's no potential for stop() and work() to be called simultaneously, removing the locks should be fine.

gr-network/lib/tcp_sink_impl.cc
```diff
@@ -228,8 +228,6 @@ int tcp_sink_impl::work(int noutput_items,
                         gr_vector_const_void_star& input_items,
                         gr_vector_void_star& output_items)
 {
-    gr::thread::scoped_lock guard(d_setlock);
```
mbr0wn, line 231: It looks like this does protect some attributes. If I'm reading this Boostery correctly, `accept_handler()` runs async, and `run_listener()` runs in its own thread. The former modifies `d_connected` (could be made `std::atomic_bool`, or maybe even `std::atomic_flag`), `d_tcpsocket` (maybe we can use `d_setlock` just for this). The latter modifies `d_start_new_listener` (could also be properly using atomic types instead of requiring a setlock).

mbr0wn (commented):
Not sure if I'm correct, but I think the tcp_sink changes are potentially unsafe. If I'm right, two ways to fix this are:
- Just don't touch the TCP sink, it's not made for the high-performance use cases anyway, and no one is complaining about locking load when using it.
- Spend some time to minutely check the concurrency accesses in that block, use proper atomic semantics, etc.

The UDP changes look fine. How about you remove the TCP change and deal with that separately?

Unless of course I'm reading code after my brain's prime time in the day.

gr-network/lib/tcp_sink_impl.cc
marcusmueller: `d_setlock` simply is not touched anywhere else but `stop`, so there's simply nothing protected here. gnuradio-runtime itself doesn't use it.

mbr0wn: Yeah you're right, the single lock is nonsense. But unlike the UDP blocks, I think we should be doing some safe concurrent accesses here.

If you remove the lock, then the code is just as good/bad as it is before. So sure, this change won't do any harm. But now that you've opened that can of worms, what about the missing locks in the TCP sink?

I guess you could say "it's unsafe already, and at least it'll be faster". Why not! Or, you could put locks in run_listener() and accept_handler(), too. ("it's safe now, albeit a bit slower"). Are you even supposed to put locks in an accept_handler function? Don't know! Brain still in holidays-mode.

marcusmueller: So, I opened that can of worms, and I stared into it. And the worms stared back: you're right, we *must* protect things from concurrent access from accept_handler / run_listerner. That's an oversight on my end.

mormj:
Closing this draft PR - please reopen if it is revisited.

PR #5444, opened 2022-01-04
cmake: build python deps conditionally

mormj:
## Description

Currently, there is a hard dependence on pybind11 to build gnuradio-runtime
This PR adds additional conditionals so that if the user selects ENABLE_PYTHON=OFF, gnuradio-runtime can be built without error.  Also, if ENABLE_PYTHON=OFF, grc and gr-utils is skipped since those are all python based

Only affects when ENABLE_PYTHON=OFF is set (forced on by default)

## Testing Done
Added an additional test to the CI

## Checklist


gnuradio-runtime/lib/CMakeLists.txt
```diff
@@ -135,14 +140,20 @@ target_link_libraries(gnuradio-runtime PUBLIC
...
+    libunwind::libunwind 
   )
 
+if (ENABLE_PYTHON)
+target_link_libraries(gnuradio-runtime
+  INTERFACE
+  Python::Module
+  PRIVATE
+  Python::Python
+  pybind11::pybind11
```
marcusmueller, line 153: Agree in principle, but isn't this part of the runtime completely independent from Pybind? Why do we need to link it against pybind?

mormj: The only part that is not independent is the gateway blocks that use the python embedding via pybind.  But linking is a good question since pybind11 is header only.  It complains though if it is not there.  Perhaps a good question for @ryanvolz

ryanvolz: When you add a "link" to the `pybind11::pybind11` target, what it actually does is add the include paths for pybind11 and python to your target. No actual linking is added.

marcusmueller (approved):

PR #5451, opened 2022-01-09
cmake: Repair build order dependencies.

drmpeg:


marcusmueller (approved):
Ah! grc needs to be enabled for the gr-xyz/CMakeLists.txt to include gr-xyz/grc/CMakeLists.txt.

PR #5452, opened 2022-01-09
grc: Follow up to #5442

dl1ksv:
fftsize starts in the runtime gui of the f/c sink with 512.
Adjust it to 32 , as in freq_sink_f/c

Introduce fft range check to the sink_x and the waterfall_sink_x block.
Check inside the sink_(f/c) blcok, if the fftsize is valid, to avoid core dump.

---

# Pull Request Details
fftsize starts in the runtime gui of the f/c sink with 512.
Adjust it to 32 , as in it's yml file and in freq_sink_f/c

Introduce fft range check to the sink_x and the waterfall_sink_x block.
Check inside the sink_(f/c) blcok, if the fftsize is valid, to avoid core dump.

## Description
Adds range checks for sink_x and waterfall_sink_x
Completes #5442 

## Checklist

### DIFF
--- a/gr-qtgui/lib/sink_f_impl.cc
+++ b/gr-qtgui/lib/sink_f_impl.cc
@@ -135,8 +135,18 @@ QWidget* sink_f_impl::qwidget() { return d_main_gui.qwidget(); }
 
 void sink_f_impl::set_fft_size(const int fftsize)
 {
-    d_fftsize = fftsize;
-    d_main_gui.setFFTSize(fftsize);
+    if ((fftsize >= d_main_gui.MIN_FFT_SIZE) && (fftsize <= d_main_gui.MAX_FFT_SIZE)) {
+        d_fftsize = fftsize;
+        d_main_gui.setFFTSize(fftsize);
+    } else {
+        GR_LOG_INFO(
+            d_logger,
+            fmt::format("FFT size must be >= {} and <= {}.\nSo falling back to {}.",
+                        d_main_gui.MIN_FFT_SIZE,
+                        d_main_gui.MAX_FFT_SIZE,
+                        d_main_gui.DEFAULT_FFT_SIZE));
+        d_main_gui.setFFTSize(d_main_gui.DEFAULT_FFT_SIZE);
+    }
 }
 
 int sink_f_impl::fft_size() const { return d_fftsize; }
--- a/gr-qtgui/grc/qtgui_freq_sink_x.block.yml
+++ b/gr-qtgui/grc/qtgui_freq_sink_x.block.yml
@@ -399,6 +399,7 @@ outputs:
 
 asserts:
 - ${fftsize >= 32 and fftsize <= 32768}
+- ${not (fftsize & (fftsize -1))}
 
 templates:
     imports: |-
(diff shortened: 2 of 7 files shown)


marcusmueller:
looks good, do we need to check for power-of-two-ness, as well?

dl1ksv:
God point. Indeed I checked it for the freq-sink
https://github.com/gnuradio/gnuradio/blob/3993cfc7f3381c20b8bb82faa6f54c70f8aa3a21/gr-qtgui/lib/freqcontrolpanel.cc#L239-L254
If not a power of two I select the greatest power of  two lower than the value.

This could be done for the sink-x and waterfall-sink. I'll see if I can do it quickly.

dl1ksv:
I'd like to handle the fftsize menu at one place and use it in the different sinks. But this is a bigger task, so I'd handle it in a different pr later.

gr-qtgui/lib/sink_f_impl.cc
```diff
@@ -135,8 +135,18 @@ QWidget* sink_f_impl::qwidget() { return d_main_gui.qwidget(); }
 
 void sink_f_impl::set_fft_size(const int fftsize)
 {
-    d_fftsize = fftsize;
-    d_main_gui.setFFTSize(fftsize);
+    if ((fftsize >= d_main_gui.MIN_FFT_SIZE) && (fftsize <= d_main_gui.MAX_FFT_SIZE)) {
```
mbr0wn, line 138: My expectation would have been to coerce to the closest available value (e.g., if you request FFT size 10000 you still get 32kib). But honestly, I'm not sure if that's better than this, so no complaint here, just an observation.

mbr0wn (changes requested):
@dl1ksv The code changes are fine, but can you please improve the commit messages' subject lines? You aren't modifying grc, and the reference to #5442 is meaningless when viewed with `git log`. My suggestions:

```
qtgui: fft sink: Improve bounds checking
qtgui: fft sink: Add power-of-two check to GRC bindings
```

gr-qtgui/lib/sink_f_impl.cc
dl1ksv: At the moment, the power-of-two check is only implemented in freq_sink_(c/f). But you can only enter a non power-of-two value, if you edit the python code. In this case 10000 would be replaced by 8192. Another possible value could be 16384.

If you use grc, you can't generate a flowgraph wit fftsize of 10000.
I'd like to handle all the fftsize checking and possible replacement in a single class, that could be used by the different sinks. This would make maintaining easier. But this will take some time. So the assertions in the yml files will help to avoid some but not all possible errors.

mormj:
My only concern with limiting to powers of 2 is that if people load old .grc files with a non-power of 2, the behavior is changed - when it doesn't necessarily have to be.  The Freq sink will still work with non-power-of-2 fft size

dl1ksv:
Even the behaviour of the freq sink has changed. With #5442 I introduced a check for power of 2 and changed (!) the value to a power of 2. Before in the freq sink the fftsize was set to 32 if it was not a power of 2 and for the waterfall sink and the (general) sink I don't know which value really was applied. I think it's better to get an error message than an invalid result.

dl1ksv:
But it would be nice if we could assign an error message to the assert statement in yml file that would tell more than 'assertion failed'

PR #5469, opened 2022-01-13
grc, block bindings: enable additional error messages for assertions

dl1ksv:
---

# Pull Request Details

If the validation of an assert statement of a block fails, grc only shows the failed assertion. This may sometimes be a bit cryptic like

Assertion "not (fftsize & (fftsize -1))" failed

which means  "fftsize must be a power of 2".

So it would be nice, to add optional a descriptive message to the assertion.

## Description
Optional extend the asserts statements of a block by a descriptive string
like 

${fftsize >= 32 and fftsize <= 32768;'The FFT Size is out of range 32 - 32768'}

instead of 

${fftsize >= 32 and fftsize <= 32768}

or
- ${not (fftsize & (fftsize -1));'The FFT Size "{}" should be a power of 2'.format(fftsize)}

If the  validation of the assert fails, the descriptive string will be printed as error message in grc. If there is no descriptive string,
the failed assertion will be printed. ( The actual behaviour )

## Checklist

### DIFF
--- a/gr-qtgui/grc/qtgui_freq_sink_x.block.yml
+++ b/gr-qtgui/grc/qtgui_freq_sink_x.block.yml
@@ -398,8 +398,8 @@ outputs:
     hide: ${ not showports }
 
 asserts:
-- ${fftsize >= 32 and fftsize <= 32768}
-- ${not (fftsize & (fftsize -1))}
+- ${fftsize >= 32 and fftsize <= 32768#'The FFT Size is out of range 32 - 32768'}
+- ${not (fftsize & (fftsize -1))#The FFT Size should be a power of 2}
 
 templates:
     imports: |-
--- a/gr-qtgui/grc/qtgui_sink_x.block.yml
+++ b/gr-qtgui/grc/qtgui_sink_x.block.yml
@@ -104,8 +104,8 @@ outputs:
     hide: ${ not showports }
 
 asserts:
-- ${fftsize >= 32 and fftsize <= 32768}
-- ${not (fftsize & (fftsize -1))}
+- ${fftsize >= 32 and fftsize <= 32768#The FFT Size is out of range 32 - 32768}
+- ${not (fftsize & (fftsize -1))#The FFT Size should be a power of 2}
 
 templates:
     imports: |-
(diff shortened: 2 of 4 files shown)


haakov:
I haven't looked at the code yet, but I really like the idea :)

skoslowski:
I also like the idea. Clear and helpful error messages are very useful.
My concern here is the new syntax `asserts: [${assert_expr;message_expr}]`:
* Currently all uses of `${}` (outside of the templates that is) are simple valid python expressions, basically a subset of what mako allows. Adding a semicolon in there makes this something else: two statements - which can't run in `eval()`
* For the same reason this makes all block-wrappers using it backward-incompatible. Outdated GRC versions can;t eval something with a `;` in it.  
* I am not sure, if an evaled message is strictly necessary here.

I have currently no suggestion how to add these message inside the assert list without breaking compat. But if compat can be broken for this, than my preference would be using something that does not add special syntax. e.g. in addtion to a eval-string  the items in the asserts list could be dicts:
```yaml
asserts:
- ${blabla}             # current, simple form
- expr: ${blabla}    # optional, extended form
  message: "blabla failed" 
```
This may be more verbose, but it uses the existing yaml and keeps the rules for `${}` expressions unchanged.

dl1ksv:
I found a simple way to keep backward compatibility.
Just use # as separator. So an assertion would like

\- ${not (fftsize & (fftsize -1))#The FFT Size should be a power of 2}
This is valid code and can be checked by grc.
If the assertion fails you see the comment as addition.
The modified version of grc just extracts the comment and displays it as error message.

At the moment the comment is just handled as string. If in the future we need more complex messages it could be extended to an eval-string.

skoslowski:
This is clever hack. Nice!

However, I still don't like adding "mini-language" here: a comment inside an expansion field inside a string that is gets parsed like a string literal - maybe even fully eval'ed in the future. Seems complicated when there is all that nice YAML around it. 

If you went with sth like my suggestion above, it could immediately be used for in-tree wrappers. OOTs would have to wait a bit before opt-in to the extended syntax. IMO this is a much cleaner (consistent, explainable, maintainable, ...) solution.

mormj (approved):

mormj:
Agree with @skoslowski 's suggestion of not introducing mini-language into the yaml makes a lot of sense, so holding off on this one for now.  I really do like the addition of the message into the assertions though - very useful and hope we can get it worked out.

dl1ksv:
Forget the change in block.py. 
Just add the comment to the asserts, and you get a more instructive error message. This is already valid code, but nobody used it.

mormj:
the key fix here is included in #5763

PR #5494, opened 2022-01-18
global: cleanup name change master to main

### LINKED ISSUE #4963: Rename default branch
Following the move across the open source community as described and executed here: https://github.com/gnuradio/volk/issues/461 for the volk database

This needs to be done at a convenient time so as to not disrupt any impending release - preferable right after our 3.10 release, or if that is deferred then the next set of maint branch releases

As a side note, the github default is `main` - but since we make heavy use of `maint-` this will make tab completion slightly more cumbersome.  Perhaps a `devel` or `default`, though that would depart from the standard `main`

mormj:
## Description
In the change of default branch naming, some things in the code that explicitly refers to `master` that needs to be changed to `main`

## Related Issue
Fixes #4963 

Mostly just docs and CI configurations

### DIFF
--- a/.conda-forge.yml
+++ b/.conda-forge.yml
@@ -47,4 +47,4 @@ skip_render:
   - LICENSE
 # enable uploads to Anaconda Cloud from specified branches only
 # ***** UPDATE THIS FOR MAINTENANCE BRANCHES ********************************
-upload_on_branch: master
+upload_on_branch: main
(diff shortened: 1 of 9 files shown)


ryanvolz:
It looks good, and I just checked that re-rendering the conda CI produces no diff, so you got everything.

mormj:
Thanks!

PR #5495, opened 2022-01-18
docs: redirect master branch

mormj:
Document in `master` branch that it should not be used and how to update


willcode:
@mormj Did you want this on the master branch? It is currently against main.

willcode:
It's the "this branch is left here" that's confusing, since "this" branch is "main".

mormj:
shoot.  already foiled!!!

PR #5505, opened 2022-01-20
SigMF Minimal Blocks

777arc:
## Description

A first cut of the two minimal SigMF blocks that have been discussed in #SigMF within Matrix.  The motivation here is to improve adoption of SigMF, by having at least basic ability to save and read SigMF files within GNU Radio, in Ubuntu 22 specifically, as the freeze is happening soon and gr-sigmf isn't ready.  These blocks are immediately being set to "deprecated" as you can see, they will be removed when gr-sigmf is ready to use.

The SigMF Sink is just a wrapper around File Sink that produces the .sigmf-meta file based on params fed into the block.  The idea was to keep it extremely simple.  

The SigMF Source is literally just a File Source with some extra documentation explaining how to use it to read in SigMF files, so this is targeted at newer folks, and the most important part of this block is that a SigMF Source will come up in the block tree, else we would be leaving people wondering why there is a Sink with no Source.

This is obviously extremely last minute, wrt Ubuntu 22 freeze, so if there is any chance that these changes might break something or lead to poor quality code locked into Ubuntu, we can just nix this PR and focus on the actual gr-sigmf.

If anyone wants to help, I can give you write access to my fork. 

## Testing Done

Only basic testing, done on 3.10 in Ubuntu 20.04.  I should have more time next week to test and throw it weird combos of params, but I didn't want to wait to get it up here due to the time crunch of Ubuntu 22 freeze.

### DIFF
--- a/gr-blocks/grc/blocks.tree.yml
+++ b/gr-blocks/grc/blocks.tree.yml
@@ -38,6 +38,7 @@
   - blocks_file_meta_source
   - blocks_file_meta_sink
   - blocks_tagged_file_sink
+  - blocks_sigmf_sink_minimal
 - IQ Correction:
   - blocks_correctiq_auto
   - blocks_correctiq_man
--- a/gr-blocks/grc/blocks.tree.yml
+++ b/gr-blocks/grc/blocks.tree.yml
@@ -38,6 +38,7 @@
   - blocks_file_meta_source
   - blocks_file_meta_sink
   - blocks_tagged_file_sink
+  - blocks_sigmf_source_minimal
   - blocks_sigmf_sink_minimal
 - IQ Correction:
   - blocks_correctiq_auto
--- a/gr-blocks/grc/blocks_sigmf_sink_minimal.block.yml
+++ b/gr-blocks/grc/blocks_sigmf_sink_minimal.block.yml
@@ -5,7 +5,7 @@ flags: [ python, deprecated ]
 parameters:
 -   id: filename
     label: File Name
-    dtype: string
+    dtype: file_save
 -   id: sample_rate
     label: Sample Rate
     dtype: float
--- a/gr-blocks/grc/blocks_sigmf_sink_minimal.block.yml
+++ b/gr-blocks/grc/blocks_sigmf_sink_minimal.block.yml
@@ -12,6 +12,7 @@ parameters:
 -   id: center_freq
     label: Center Frequency
     dtype: float
+    default: 0
 -   id: author
     label: Author
     dtype: string
@@ -27,10 +28,17 @@ parameters:
 -   id: type
     label: Stream Type
     dtype: enum
-    options: [complex, float]
+    options: [complex, float, short]
     option_attributes:
-        size: [gr.sizeof_gr_complex, gr.sizeof_float]
+        size: [gr.sizeof_gr_complex, gr.sizeof_float, gr.sizeof_short]
     hide: part
+-   id: ishort
+    label: Type of Shorts
+    dtype: bool
+    default: 'True'
+    options: ['False', 'True']
+    option_labels: ['Real Only', 'Interleaved IQ']
+    hide: ${ 'all' if type != 'short' else 'none' }
 
 inputs:
 -   domain: stream
@@ -49,6 +57,7 @@ templates:
             center_freq=${center_freq},
             author=${author},
             description=${description},
-            hw_info=${hw_info})
+            hw_info=${hw_info},
+            ishort=${ishort})
 
 file_format: 1
--- a/gr-blocks/grc/blocks_sigmf_sink_minimal.block.yml
+++ b/gr-blocks/grc/blocks_sigmf_sink_minimal.block.yml
@@ -21,7 +21,8 @@ parameters:
 -   id: center_freq
     label: Center Frequency
     dtype: float
-    default: 0
+    options: [np.nan]
+    option_labels: ['None']
 -   id: author
     label: Author
     dtype: string
@@ -43,7 +44,9 @@ asserts:
 - ${ len(filename) > 0 }
 
 templates:
-    imports: from gnuradio import blocks
+    imports: |-
+        from gnuradio import blocks
+        import numpy as np
     make: |-
         blocks.sigmf_sink_minimal(
             item_size=${type.size},
--- a/gr-blocks/grc/blocks_sigmf_sink_minimal.block.yml
+++ b/gr-blocks/grc/blocks_sigmf_sink_minimal.block.yml
@@ -1,0 +1,54 @@
+id: blocks_sigmf_sink_minimal
+label: SigMF Sink (Minimal)
+flags: [ python, deprecated ]
+
+parameters:
+-   id: filename
+    label: File Name
+    dtype: string
+-   id: sample_rate
+    label: Sample Rate
+    dtype: float
+-   id: center_freq
+    label: Center Frequency
+    dtype: float
+-   id: author
+    label: Author
+    dtype: string
+    default: ''
+-   id: description
+    label: Description
+    dtype: string
+    default: ''
+-   id: hw_info
+    label: Hardware Info
+    dtype: string
+    default: ''
+-   id: type
+    label: Stream Type
+    dtype: enum
+    options: [complex, float]
+    option_attributes:
+        size: [gr.sizeof_gr_complex, gr.sizeof_float]
+    hide: part
+
+inputs:
+-   domain: stream
+    dtype: ${ type }
+
+asserts:
+- ${ len(filename) > 0 }
+
+templates:
+    imports: from gnuradio import blocks
+    make: |-
+        blocks.sigmf_sink_minimal(
+            item_size=${type.size},
+            filename=${filename},
+            sample_rate=${sample_rate},
+            center_freq=${center_freq},
+            author=${author},
+            description=${description},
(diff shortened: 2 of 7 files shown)


gr-blocks/python/blocks/sigmf_sink_minimal.py
```diff
@@ -0,0 +1,83 @@
...
+                "core:version": sigmf_version
+            },
+            "captures": [
+                {
+                    "core:sample_start": 0,
+                }
+            ],
+            "annotations": []
+        }
+        if center_freq:
```
marcusmueller, line 78: hm, what if I explicitly want to set `center_freq` to 0, e.g. in a system where there's more or less well-centered quadrature mixing/close-to-zero-IF, I can see how that'd happen.

Recommendation: give the `center_freq` parameter a `= None` default, check for `is not None`.

marcusmueller (changes requested):

willcode:
Python formatting script is `tools/autoformat_python.sh`, which takes a directory as an arg. It may take a single file too.

willcode:
Since there's no cpp sink and the source is minimal, should the cpp code be removed from the source?

777arc:
> Since there's no cpp sink and the source is minimal, should the cpp code be removed from the source?

done

gr-blocks/python/blocks/sigmf_sink_minimal.py
777arc: Couldn't figure out how to allow None for float type

jacobagilbert: I know it was my suggestion to make this optional, since it is per the spec, but IMO we should just require this be populated if we can't also allow zero.

jacobagilbert: BTW, the way you can do this is sort of complicated:

```diff
--- a/gr-blocks/grc/blocks_sigmf_sink_minimal.block.yml
+++ b/gr-blocks/grc/blocks_sigmf_sink_minimal.block.yml
@@ -21,7 +21,8 @@ parameters:
 -   id: center_freq
     label: Center Frequency
     dtype: float
-    default: 0
+    options: [np.nan]
+    option_labels: ['None']
 -   id: author
     label: Author
     dtype: string
@@ -43,7 +44,9 @@ asserts:
 - ${ len(filename) > 0 }
 templates:
-    imports: from gnuradio import blocks
+    imports: |-
+        from gnuradio import blocks
+        import numpy as np
     make: |-
         blocks.sigmf_sink_minimal(
             item_size=${type.size},
```

Then just check to see if the value is `np.nan` instead of `if frequency:`

jacobagilbert:
@777arc I pushed an update that will only populate configured metadata fields... can revert if we don't like.

777arc:
> @777arc I pushed an update that will only populate configured metadata fields... can revert if we don't like.

This looks great!

jacobagilbert (approved):

mormj:
To make it show up orange as deprecated, we just need to modify the following in grc/core/blocks/block.py
```
            return self.flags.deprecated or any("deprecated".casefold() in cat.casefold()
                       for cat in self.category)
```

mormj:
Might want to make the Sample Rate field default to `samp_rate` like other blocks

mormj:
Maybe the file picker for the sigmf source should filter on sigmf-data?  I see it is documented in the docs tab, but might want to narrow it down further.

777arc:
RE Josh:
- I made samp_rate default
- This addition to grc/core/blocks/block.py feels like it should be its own commit/PR right?
- That's a good idea, to filter for sigmf files, didn't even think about that.  I dug deeper into how the dtype file_open might do that, and in grc/gui/ParamWidgets.py where it calls Gtk.FileChooserDialog, to filter, an object of the gtk.FileFilter can be applied by using the add_filter() method.  But like we would need to pass in the file type to filter, and that would touch SO much grc code, I think we would want it to be its own PR and perhaps not rush it like we are doing with these 2 new blocks =P

gr-blocks/grc/blocks_sigmf_sink_minimal.block.yml
```diff
@@ -0,0 +1,62 @@
...
+-   id: filename
+    label: File Name
+    dtype: file_save
+-   id: type
+    label: Stream Type
+    dtype: enum
+    options: [complex float (fc32_le), real float (rc32_le), complex short (sc16_le), real short (rc16_le)]
+    option_attributes:
+        type: [complex, float, short, short]
+        size: [gr.sizeof_gr_complex, gr.sizeof_float, gr.sizeof_short, gr.sizeof_short]
```
jdemel, line 15: Is complex short and real short the same here?

gr-blocks/grc/blocks.tree.yml
```diff
@@ -38,6 +38,8 @@
   - blocks_file_meta_source
   - blocks_file_meta_sink
   - blocks_tagged_file_sink
+  - blocks_sigmf_source_minimal
+  - blocks_sigmf_sink_minimal
```
jdemel, line 42: It looks like those to files don't get installed.

gr-blocks/python/blocks/sigmf_sink_minimal.py
```diff
@@ -0,0 +1,87 @@
...
+        elif item_size == 2 and is_complex:
+            datatype_str = 'ci16_le'
+        elif item_size == 2 and not is_complex:
+            datatype_str = 'ri16_le'
+        else:
+            raise ValueError
+        
+        if '.sigmf' in filename:
+            filename = filename.rsplit('.', 1)[0]
+        gr.log.info('Generating '+filename+'.sigmf-meta, writing SigMF data to '+filename+'.sigmf-data')
```
jdemel, line 49: f-strings would make this line more readable.
Also, the whole file would benefit from auto-formatting.

jdemel (changes requested):
Just a few minor things.

gr-blocks/grc/blocks_sigmf_sink_minimal.block.yml
jacobagilbert: They actually are since the shorts are interleaved but come one part per item. At some point we might consider making a special type for this to avoid this situation.

gr-blocks/grc/blocks.tree.yml
jacobagilbert: Every blocks*yml in this directory will get installed by cmake, try removing `..../build/gr-blocks/grc` to re-trigger the cmake logic to run.

gr-blocks/python/blocks/sigmf_sink_minimal.py
jacobagilbert: Agree on both

gr-blocks/grc/blocks_sigmf_sink_minimal.block.yml
777arc: Yep the only difference is how it gets annotated in the metafile, and there's no way for the block to know which one it is unless you tell it, so that's what the ishort param is for.

gr-blocks/python/blocks/sigmf_sink_minimal.py
777arc: Done (auto-formatting part)

gr-blocks/grc/blocks.tree.yml
jdemel: That's great. Is this a new default in all OOTs? I remember adding every block file to the install list in grc/

gr-blocks/grc/blocks_sigmf_sink_minimal.block.yml
jdemel: IShort is one of those datatypes that are confusing. They exist somehow but are poorly supported. VOLK defines smth similar to `std::complex<short>` but it seems like it only exists for the sake of completeness. i.e. it is mostly unsupported.

gr-blocks/python/blocks/sigmf_sink_minimal.py
```diff
@@ -0,0 +1,89 @@
...
+            datatype_str = 'ci16_le'
+        elif item_size == 2 and not is_complex:
+            datatype_str = 'ri16_le'
+        else:
+            raise ValueError
+
+        if '.sigmf' in filename:
+            filename = filename.rsplit('.', 1)[0]
+        gr.log.info('Generating '+filename +
+                    '.sigmf-meta, writing SigMF data to '+filename+'.sigmf-data')
```
jdemel, line 51:
```suggestion
        gr.log.info(f'Generating {filename}.sigmf-meta, writing SigMF data to {filename}.sigmf-data')
```
At least for me, f-strings are more readable.

gr-blocks/grc/blocks_sigmf_sink_minimal.block.yml
jacobagilbert: As far as I know it exists primarily as a file format; half the size of fc32 yet has enough dynamic range to support a lot of RF use cases.

gr-blocks/grc/blocks.tree.yml
jacobagilbert: Im not sure if OOTs do this [yet], to they probably should it's two lines of CMake.

mormj (approved):

PR #5506, opened 2022-01-21
digital: put back import of packet_utils

mormj:
Had gotten lost in the autopep8 formatting
Thanks to @drmpeg for finding and bisecting as part of #5188


drmpeg (approved):
Tested and works good.

PR #5509, opened 2022-01-21
ci: make conda build integrated with main workflow

mormj:
## Description
Following a suggestion from @willcode on chat - don't do conda builds unless it passes the formatting checks

none, just github ci

## Testing Done
- [x] check in with commit that doesn't pass clang-format, verify that conda doesn't build
- [ ] remove the commit and push, verify that conda does build

### DIFF
--- a/.github/workflows/conda-build.yml
+++ b/.github/workflows/conda-build.yml
@@ -3,7 +3,8 @@
 # -*- mode: yaml -*-
 
 name: Build conda package
-on: [push, pull_request]
+on: 
+  workflow_call:
 
 jobs:
   build:
--- a/.github/workflows/make-test.yml
+++ b/.github/workflows/make-test.yml
@@ -136,4 +136,4 @@ jobs:
       run: 'cd /build && ctest --output-on-failure'
   call-conda-build:
     needs: [check-formatting, check-python-formatting]
-    uses: gnuradio/gnuradio/.github/workflows/conda-build.yml@main
+    uses: gnuradio/gnuradio/.github/workflows/conda-build.yml@78f1070ed
--- a/.github/workflows/make-test.yml
+++ b/.github/workflows/make-test.yml
@@ -136,4 +136,4 @@ jobs:
       run: 'cd /build && ctest --output-on-failure'
   call-conda-build:
     needs: [check-formatting, check-python-formatting]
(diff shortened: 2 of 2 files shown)


ryanvolz:
Note that this won't survive a re-render with conda-smithy. Or rather, re-rendering will result in the `conda-build.yml` workflow returning and thus being duplicated. There are already some manual steps that happen after re-rendering, to delete files that are not necessary for GNU Radio, and this could be an additional manual step to copy the workflow into `make-test.yml`. TBD how easy that would be to automate in something like #5322. It would be worth adding a note about this step to the comments in `.conda-forge.yml` and `.packaging/conda_recipe/README`.

mormj:
Definitely don't want to make the process more manual - perhaps it is easier then to just copy the formatting steps into the conda-build workflow?

ryanvolz:
I guess that would run the formatting steps twice, but it would be slightly easier to do manually. I'm not too concerned with the manual labor since it doesn't have to happen that often and it amounts to copy and paste. But perhaps there is a cleverer way. Maybe there's a way to rename `conda-build.yml` so that Github doesn't pick it up as a workflow, but then somehow import the `build` key from its yaml file into the `make-test.yml`. Does YAML have include functionality like that? Time to do some Googling.

ryanvolz:
Maybe https://docs.github.com/en/actions/using-workflows/reusing-workflows? Would only involve change the `on:` directive in `conda-build.yml` and then calling the workflow from `make-test.yml`, if I'm reading it correctly.

mormj:
It's worth a shot!  here goes

mormj:
might play with this more later - right now it seems there is a chicken and egg problem - the reusable workflow has to already be checked in to be called by the main workflow.  i just need to check it in somewhere else, but need to come back to it later

ryanvolz:
> the reusable workflow has to already be checked in to be called by the main workflow

That makes it hard to test updating the conda CI build, so I think I'd prefer the manual editing if we can't get it to read the workflow file from the commit that triggered the CI.

mormj:
A bit of trial and error, but all error - doesn't seem to be supported yet per https://github.com/github/feedback/discussions/9050

Closing this for now.  Maybe they will support the feature in the near term

PR #5543, opened 2022-01-31
runtime: downstream buffer size calculation back to 3.9 value

### LINKED ISSUE #5542: Custom Buffers breaks DVB-T receiver flow graph.
### What happened?

A bug happened!

### System Information

OS: Ubuntu 20.04, 21.10

### GNU Radio Version

3.11-git (main)

### Specific Version

tip of main

Run the example flow graph gr-dtv/examples/dvbt_rx_8k.grc. The flow graph fails because it can't allocate enough input items for a block. However, the request is not large and worked fine before.

I've bisected the bug to commit https://github.com/gnuradio/gnuradio/commit/f3c558d88bc68d865f823c31e7d9aa78b3feab59

The test file for dvbt_rx_8k.grc can be downloaded here:

https://www.w6rz.net/adv16.cfile

@mormj @dsorber 

### Relevant log output

```shell
block_executor :error: sched: <block dvbt_demod_reference_signals (9)> is requesting more input data  than we can provide.  ninput_items_required = 2  max_possible_items_available = 1  If this is a filter, consider reducing the number of taps.
```

mormj:
## Description
The calculation of buffer sizes for downstream blocks was changed in the custom buffer implementation - the reason for the change related to history is still unclear.

## Related Issue
Fixes #5542 

All blocks 

## Testing Done
Ran the example flowgraph that was broken as described in #5542.  Needs testing of 
- [ ] Custom Buffers
- [ ] Other flowgraph

## Checklist

### DIFF
--- a/gnuradio-runtime/lib/flat_flowgraph.cc
+++ b/gnuradio-runtime/lib/flat_flowgraph.cc
@@ -119,8 +119,8 @@ void flat_flowgraph::allocate_block_detail(basic_block_sptr block)
             double decimation = (1.0 / dgrblock->relative_rate());
             int multiple = dgrblock->output_multiple();
             int history = dgrblock->history();
-            nitems = std::max(
-                nitems, static_cast<int>(2 * (decimation * multiple + (history - 1))));
+            nitems =
+                std::max(nitems, static_cast<int>(2 * (decimation * multiple + history)));
 
             // Calculate the LCM of downstream reader nitems
 #ifdef BUFFER_DEBUG


willcode (approved):

willcode:
Note that blocks can set history at runtime and there is no realloc. Reverting to the previous behavior makes the alloc good at init, at least.

mormj:
The trick here is going to be figuring out why it was changed to `history - 1` and being able to test those corner cases presumably in the custom buffer use case that precipitated that change

mormj:
@dsorber - what types of tests do you think we need to run to make sure this doesn't break the single mapped buffer implementation?

dsorber:
> @dsorber - what types of tests do you think we need to run to make sure this doesn't break the single mapped buffer implementation?

A while back I started putting together some test cases here: https://github.com/BlackLynx-Inc/gr-blnxngsched/tree/main/python. I know things have moved around a bit since then so I'm not sure how easy it would be to resurrect those tests.

The other thing to do would be to hardcode buffer allocation such that [host_buffer](https://github.com/gnuradio/gnuradio/blob/main/gnuradio-runtime/lib/host_buffer.cc) (which utilizes the `single_mapped_buffer` interface) are used for allocations in place of `buffer_double_mapped`.  Then rerun the QA tests and all aside from a handful should pass.  I know we figured out the best way to do this a few months ago but I cannot recall the details at the moment.  I will poke around a little and see if it comes back to me.

dsorber:
I dug up the conversation, the method was to update `io_signature.{h,cc}` and change all the `buffer_double_mapped::type` default values to `host_buffer::type`.  This should ensure that all blocks will create `host_buffer` instances by default.

mormj:
Also verified that gr-cuda works after this change.  

I'm ok with this change as long as 
1) normal double buffer GR works fine
2) simple custom buffers like gr-cuda work fine

and I think this passes that bar

dsorber:
I see slightly different but overall very similar results on my system (test timeout at 30s):

Before:
```
	 28 - blocks_qa_gr_top_block.cc (Failed)
	112 - qa_fecapi_dummy (Timeout)
	113 - qa_fecapi_ldpc (Timeout)
	114 - qa_fecapi_repetition (Timeout)
	135 - qa_filterbank (Timeout)
	145 - qa_pfb_arb_resampler (Timeout)
	151 - qa_rational_resampler (Failed)
	201 - qa_linear_equalizer (Timeout)
	215 - qa_pfb_clock_sync (Timeout)
	231 - qa_qtgui (Failed)
```

After:
```
	 28 - blocks_qa_gr_top_block.cc (Failed)
	112 - qa_fecapi_dummy (Timeout)
	113 - qa_fecapi_ldpc (Timeout)
	114 - qa_fecapi_repetition (Timeout)
	135 - qa_filterbank (Timeout)
	145 - qa_pfb_arb_resampler (Timeout)
	151 - qa_rational_resampler (Failed)
	201 - qa_linear_equalizer (Timeout)
	215 - qa_pfb_clock_sync (Timeout)
	231 - qa_qtgui (Failed)
```

`blocks_qa_gr_top_block.cc` is an expected failure as it tests specific double mapped buffer behavior.  `qa_qtgui` is also unrelated.  The other tests used to pass, but haven't run them recently.  As you said, this is a separate issue.

dsorber (approved):
This was probably leftover from some experimentation and turns out to not really be needed.

PR #5547, opened 2022-02-01
grc: A generated python hier_block is written to the wrong directory

### LINKED ISSUE #5531: Hier Blocks are not installed to ~/.grc_gnuradio
### What happened?

If you create Hier blocks in GRC and run them the generated python and block description file are generated int he same folder as the .grc file. They are not copied to ~/.grc_gnuradio .

### System Information

OS: Distributor ID:	Ubuntu
Description:	Ubuntu 20.04.3 LTS
Release:	20.04
Codename:	focal

### GNU Radio Version

3.11-git (main)

### Specific Version

v3.10.0.0-5-g3349d483

1. Create Hier block in GRC
2. Run it to install
3. Reload Block list
4. Hier block is not available because it was not copied to ~/.grc_gnuradio . Generated files are in the same folder as the .grc file of the Hier block and cannot be found by GRC.

### Relevant log output

_No response_

dl1ksv:
---

# Pull Request Details

The generated code is written to the directory of the generating grc file
instead to the 'hier_block_lib_dir' directory.

cpp generated code ends up in the correct directory.

This error was introduced with commit af3fbe432fe16efb05f26c24a0fa9724680dc8dd

## Description

Take the directory from the config settings ( as before)

## Related Issue

Fixes #5531

## Testing Done

Tested with different hier blocks

## Checklist

### DIFF
--- a/grc/compiler.py
+++ b/grc/compiler.py
@@ -55,7 +55,7 @@ def main(args=None):
     for grc_file in args.grc_files:
         os.path.exists(grc_file) or exit('Error: missing ' + grc_file)
         Messages.send('\n')
-
+        platform.config.hier_block_lib_dir = output_dir
         flow_graph, file_path = platform.load_and_generate_flow_graph(
             os.path.abspath(grc_file), os.path.abspath(output_dir))
         if not file_path:
--- a/grc/core/generator/hier_block.py
+++ b/grc/core/generator/hier_block.py
@@ -12,7 +12,7 @@ from ..io import yaml
 class HierBlockGenerator(TopBlockGenerator):
     """Extends the top block generator to also generate a block YML file"""
 
-    def __init__(self, flow_graph, output_dir=None):
+    def __init__(self, flow_graph, _):
         """
         Initialize the hier block generator object.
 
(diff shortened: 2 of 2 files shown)


willcode:
Does this still work with grcc? #4871 makes it seem like this change was intended. Should the correct directory be passed in from above instead of ignored?

dl1ksv:
Good point. 

Without my patch
In case test.grc is a hier block with output langage python
grcc -o /tmp/test test.grc
is written to /tmp/test
but if the output language is c++ the code is written ~/.grc_gnuradio/
which is not intended.

After my patch in both cases the code is written ~/.grc_gnuradio.

All other combinations of generation option and output language are not affected by my patch.
To write hier block code to the specified output directory this should be fixed in grcc.
This is done by the following commit.

mormj:
> After my patch in both cases the code is written ~/.grc_gnuradio.

Shouldn't the output given `grcc -o /tmp/test` be to /tmp/test in both cases?

dl1ksv:
Yes, but I wanted to point out, that this was not the case without any patch. After my first commit the code for the hier block is written to ~/.grc_gnuradio  , independently of the the output language. For other types the code is written to the correct directory.
The second commit forces grcc -o /tmp/test to write to  /tmp/test for hier blocks , too.

mormj:
Ah ok - I didn't get the distinction between the two commits. Thanks!

mormj:
Tested and working - sorry for the delay and thank you for the fix.

mormj (approved):

PR #5557, opened 2022-02-07
cmake: fix incorrect backporting of #5444

argilo:
The backport of #5444 copied a reference to spdlog into GNU Radio 3.9. LOG4CPP should be referenced instead.


willcode:
Thanks for catching that!

PR #5559, opened 2022-02-07
Ptest work main

balister:
Updates gr-network and gr-pdu so I can  package the tests and run them on targets.

The timeout patch is needed because I am not build the c++ based tests yet, so the attempt to set properties fails. When I figure out how to deal with the c++ tests, this can be reverted. (I can also carry the patch locally if you like)


mormj (approved):

mormj:
Ready to merge this, just waiting to decide about the CMake formatting - as this will need to be rebased after that, or cause a rebase of that one.

willcode:
Backport note: 3.9 includes 1/3 commits: afaddfe9d109204fa65485a7d59fe9a8c3399300

PR #5566, opened 2022-02-11
qtgui: Add output message ports to range widgets

charles-dyfis-net:
# Pull Request Details

## Description

Previously, the Qt range widget did not support output ports -- contrary to several other Qt widgets (such as the dial control), which does.

This PR brings the widget types into conformance and lets messages be used to convey data entered by the user.

## Related Issue

None filed.

The Qt range block is modified. In theory, performance impact should be minimal.

## Testing Done

Instantiated a flowgraph with a range block, with another GUI widget following its output via variable updates, and a message sink block following outputs over the message port. Validated that both consumers retrieve updates, thus that the traditional callback mechanism was not impacted. Modified the message name, and checked that the value passed to the message debug sink was updated accordingly.

Also directly invoked `python -m gnuradio.qtgui.range` to invoke the block's self-test capabilities.

## Checklist

### DIFF
--- a/gr-qtgui/grc/qtgui_range.block.yml
+++ b/gr-qtgui/grc/qtgui_range.block.yml
@@ -58,16 +58,16 @@ parameters:
     dtype: string
     default: 'value'
     hide: part
--   id: gui_hint
-    label: GUI Hint
-    dtype: gui_hint
-    hide: part
 -   id: showports
     label: Show Msg Ports
     dtype: enum
     options: ['True', 'False']
     option_labels: ['Yes', 'No']
     default: 'False'
+    hide: part
+-   id: gui_hint
+    label: GUI Hint
+    dtype: gui_hint
     hide: part
 value: ${ value }
 
--- a/gr-qtgui/grc/qtgui_range.block.yml
+++ b/gr-qtgui/grc/qtgui_range.block.yml
@@ -53,11 +53,29 @@ parameters:
     dtype: int
     default: '200'
     hide: part
+-   id: outputmsgname
+    label: Message Property Name
+    dtype: string
+    default: 'value'
+    hide: part
 -   id: gui_hint
     label: GUI Hint
     dtype: gui_hint
     hide: part
+-   id: showports
+    label: Show Msg Ports
+    dtype: enum
+    options: ['True', 'False']
+    option_labels: ['Yes', 'No']
+    default: 'False'
+    hide: part
 value: ${ value }
+
+outputs:
+-   domain: message
+    id: out
+    optional: true
+    hide: ${ showports != 'True' }
 
 asserts:
 - ${start <= value <= stop}
@@ -65,7 +83,7 @@ asserts:
 
 templates:
     imports: |- 
-        from gnuradio.qtgui import Range, RangeWidget
+        from gnuradio.qtgui import Range, GrRangeWidget
         from PyQt5 import QtCore
     var_make: self.${id} = ${id} = ${value}
     callbacks:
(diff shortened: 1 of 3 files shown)


777arc:
Is there any way this change would break existing flowgraphs, either grc or python?

charles-dyfis-net:
> Is there any way this change would break existing flowgraphs, either grc or python?

None that I'm aware of -- it's intended to be strictly additive.

If we wanted extra certainty, it would be good to have review from someone with deeper knowledge of Qt (and specifically its Python bindings) for whether there's potential for the way the callback is wrapped to be harmful -- it's conceivable that any function that _isn't_ in fact a Qt slot instance ought to be wrapped with the slot decorator, f/e.

gr-qtgui/grc/qtgui_range.block.yml
```diff
@@ -76,7 +86,9 @@ templates:
             range = 'self._%s_range'%id
         %>\
         ${range} = Range(${start}, ${stop}, ${step}, ${value}, ${min_len})
-        ${win} = RangeWidget(${range}, self.set_${id}, "${no_quotes(label,repr(id))}", "${widget}", ${rangeType}, ${orient})
+        ${win} = GrRangeWidget(${range}, self.set_${id}, "${no_quotes(label,repr(id))}", "${widget}", ${rangeType}, ${orient}, "${outputmsgname}")
```
dl1ksv, line 89: You must change "${outputmsgname}" to "${no_quotes(outputmsgname)}" otherwise the key is quoted and will not be recognized by the receiver

gr-qtgui/grc/qtgui_range.block.yml
```diff
@@ -53,19 +53,29 @@ parameters:
     dtype: int
     default: '200'
     hide: part
+-   id: outputmsgname
```
dl1ksv, line 56: You should add the option to hide the msg outputport

gr-qtgui/grc/qtgui_range.block.yml
```diff
@@ -53,19 +53,29 @@ parameters:
...
+    hide: 'part'
 -   id: gui_hint
     label: GUI Hint
     dtype: gui_hint
     hide: part
 value: ${ value }
 
+outputs:
+-   domain: message
+    id: value
```
dl1ksv, line 69: I'd call the output port out not value, the port name is independent from the message key.

dl1ksv (changes requested):

gr-qtgui/grc/qtgui_range.block.yml
charles-dyfis-net: I'm not sure I understand the guidance here. Do you mean it should be `hide: all`? Under which circumstances?

dl1ksv: Have a look at the qtgui sink:
https://github.com/gnuradio/gnuradio/blob/f44384ae1f061f257603b02288aef4db1b74d4be/gr-qtgui/grc/qtgui_freq_sink_x.block.yml#L116-L120

and 
https://github.com/gnuradio/gnuradio/blob/f44384ae1f061f257603b02288aef4db1b74d4be/gr-qtgui/grc/qtgui_freq_sink_x.block.yml#L380-L398

Even better than using a boll value it would be to use enums, so you could toggle this option instead of entering True or False

charles-dyfis-net: I believe the PR as-amended now takes this into account.

dl1ksv: Principally this looks good. Did you try to hide and show the message ports, when there was a an entry in the gui_hint field, for instance 0,0,1,1 ?
 I noticed sometimes some error messages, when the gui_hint parameter appeared befoe the showports parameter.

See #5665

mormj:
Tested this out, and it does not break existing flowgraphs

PR #5567, opened 2022-02-12
filter: fix polyphase channelizer hier block and example

### LINKED ISSUE #4015: Broken PFB Channelizer flowgraph

777arc:
Fixes https://github.com/gnuradio/gnuradio/issues/4015, the function call shouldn't have those default args included.  The example broke because the bw is set to 0.6 in the flowgraph but it was getting reset to 1.0 which messed up the channelizer.

Signed-off-by: Marc Lichtman [marcll@vt.edu](mailto:marcll@vt.edu)

Before fix - 
Screenshot from 2022-02-12 17-20-56

After fix -

Screenshot from 2022-02-12 17-54-46

### DIFF
--- a/gr-filter/python/filter/pfb.py
+++ b/gr-filter/python/filter/pfb.py
@@ -452,8 +452,7 @@ class channelizer_hier_ccf(gr.hier_block2):
             gr.io_signature(1, 1, gr.sizeof_gr_complex),
             gr.io_signature(len(outchans), len(outchans), gr.sizeof_gr_complex))
         if taps is None:
-            taps = self.create_taps(
-                n_chans, atten=100, bw=1.0, tb=0.2, ripple=0.1)
+            taps = self.create_taps(n_chans, atten, bw, tb, ripple)
         taps = list(taps)
         extra_taps = int(math.ceil(1.0 * len(taps) / n_chans) *
                          n_chans - len(taps))


willcode (approved):

mormj (approved):

mormj:
@777arc - thanks for the fix here!  Can you please sign the commit and prefix it with `filter:`?  Thanks!

777arc:
Oh since it was a small change I used github web GUI for it, but I think that signs it for me right?

image

Also doesn't it use the PR name for the commit msg when you squash/merge?

mormj:
Yeah I can fix the commit message with the github squash - but it doesn't look like it is signed

dkozel:
@777arc GitHub is crypto signing the commit (which rebasing will actually break) but not adding the sign-off in the commit message that we use for the DCO.

777arc:
> @777arc GitHub is crypto signing the commit (which rebasing will actually break) but not adding the sign-off in the commit message that we use for the DCO.

Ah, ok np

mormj:
still not showing as signed

willcode:
It is the text that matters, but it has to be in the commit. GitHub adds its own crypto signature, for commits done via GitHub, to show that it has validated the user.

Volker does commits via GitHub, so maybe https://github.com/gnuradio/gnuradio/pull/5547 will give a clue how this is done.

777arc:
Hmm, I would think if he was also using github web gui it would have the Verified text which is automatic

image

image

ValZapod:
> using github web gui it would have the Verified text which is automatic

That is quite useless in github, BTW, since you usually do jot add your own GPG key, did you? Also the commit will be not merged, like in linux, but rebased, so it will be invalid after that. As for Signed-off-by: thay can be just added in github ui too.

mormj:
> BTW, one liners does not need DCO, do they?

It's a very minimal requirement, so we are going to be strict on it just so we stay on the up and up

Closing this for now, but @777arc, please resubmit when you can sign your commit

PR #5569, opened 2022-02-15
filter: Fix crash in Rational Resampler logging

argilo:
## Description

If the Interpolation and Decimation factors of a Rational Resampler block have common factor, and taps are manually specified, then the block segfaults when it is launched.

This happens because the block declares its own `d_logger` pointer but does not initialize it before use:

https://github.com/gnuradio/gnuradio/blob/3349d483a1f7e5db0eea042815d237810373857b/gr-filter/lib/rational_resampler_impl.h#L33

https://github.com/gnuradio/gnuradio/blob/3349d483a1f7e5db0eea042815d237810373857b/gr-filter/lib/rational_resampler_impl.cc#L124-L130

The `basic_block` class already provides a logger, so there is no need to override it. I've switched to using the provided logger here.

It looks like this bug was introduced in #3952. I've confirmed that it's present in 3.9 and 3.10.

* Rational Resampler

## Testing Done
I used the following flow graph for testing:
Screenshot from 2022-02-14 18-54-38

I also added a QA test (nearly an exact copy of `test_006_interp_decim`, but with an interpolation factor of 6 instead of 3. The test generates a segfault without the code changes, and passes with them in place.

## Checklist

### DIFF
--- a/gr-filter/lib/rational_resampler_impl.cc
+++ b/gr-filter/lib/rational_resampler_impl.cc
@@ -120,7 +120,7 @@ rational_resampler_impl<IN_T, OUT_T, TAP_T>::rational_resampler_impl(
     auto d = GR_GCD(interpolation, decimation);
 
     if (!taps.empty() && (d > 1)) {
-        d_logger->info(
+        this->d_logger->info(
             "Rational resampler has user-provided taps but interpolation ({:d}) and "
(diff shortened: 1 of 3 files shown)


mormj (approved):
LGTM!

argilo:
Rebased.

PR #5583, opened 2022-02-17
blocks: Skip alignment check when the input file is not seekable

argilo:
## Description

If a non-seekable file like /dev/zero is opened in the File Source block, the following warning appears:

`file_source :warning: file size is not a multiple of item size
(9223372036854775807 ≠ N·8)`

This happens because non-seekable files are treated as if their size is INT64_MAX, which is almost never a multiple of the item size.

https://github.com/gnuradio/gnuradio/blob/3349d483a1f7e5db0eea042815d237810373857b/gr-blocks/lib/file_source_impl.cc#L151-L180

To solve the problem, I propose to simply skip the alignment check for non-seekable files.

* File Source

## Testing Done
* Open `/dev/zero` in a File Source block (Output Type = complex) and verify that no warning is printed.
* Open a 1024-byte file in a File Source block and verify that no warning is printed.
* Open a 1023-byte file in a File Source block and verify that a warning is printed.

## Checklist

### DIFF
--- a/gr-blocks/lib/file_source_impl.cc
+++ b/gr-blocks/lib/file_source_impl.cc
@@ -176,7 +176,7 @@ void file_source_impl::open(const char* filename,
     // If length is not specified, use the remainder of the file. Check alignment at end.
     if (length_items == 0) {
         length_items = items_available;
-        if (file_size % d_itemsize) {
+        if (d_seekable && (file_size % d_itemsize)) {
             d_logger->warn("file size is not a multiple of item size ({:d} ≠ N·{:d})",
                            file_size,
                            d_itemsize);


argilo:
The same code path is present in 3.8, 3.9, and 3.10, but I only started noticing this after #5270 appeared on the main branch (because the default logging level changed, perhaps?). So I'm not sure whether it's worth backporting this change.

drmpeg:
I'm pretty sure I've seen that warning when starting Gqrx in the past. But it was easier to ignore when it just said "file size is not a multiple of item size" instead of printing some huge number.

marcusmueller:
Also used to be a warning, right? So, @drmpeg must be right, me extending the message might have been what made this noticable. 

Two options: 

1. This is actually rightfully a *warning*, and not just some informational message, then making it more noticeable would actually kind of be a good thing
2. This thing doesn't need to be a *warning*, then we should probably demote its severity level to `info` or `debug`.

gr-blocks/lib/file_source_impl.cc
```diff
@@ -176,7 +176,7 @@ void file_source_impl::open(const char* filename,
     // If length is not specified, use the remainder of the file. Check alignment at end.
     if (length_items == 0) {
         length_items = items_available;
-        if (file_size % d_itemsize) {
+        if ((file_size != INT64_MAX) && (file_size % d_itemsize)) {
```
marcusmueller, line 179: Personally, I would have preferred if you just used `d_seekable` instead of `file_size != INT64_MAX` here – it describes the intent better.

argilo:
@drmpeg You're correct that this warning has always appeared when launching Gqrx, at least as far back as 3.8. My attention was drawn to it now because of the addition of the enormous number. :smile: Given that, I think it is worth backporting the fix.

mormj (approved):

PR #5584, opened 2022-02-17
Replace all usage of Boost.Math with standard C++

argilo:
## Description
GNU Radio uses Boost.Math in a few places. Replacements are now available in standard C++:

* `boost::math::iround` -> `std::lround` (since C++11)
* `boost::integer::gcd` -> `std::gcd` (since C++17)
* `boost::integer::lcm` -> `std::lcm` (since C++17)

Since GNU Radio requires C++17 (since 3.10) it should be safe to switch away from Boost.

Note: I removed the `integer_math.h` header since it's no longer used within GNU Radio. I suppose OOTs could use it though, so we might not want to backport that to 3.10.

* gnuradio-runtime (buffers, flat_flowgraph)
* gr-digital (Symbol Sync)
* gr-filter (Polyphase Arbitrary Resampler, Rational Resampler)
* gr-qtgui (QT GUI Histogram Sink)

## Testing Done
None yet. It compiles without warnings at least. :smile: 

## Checklist


willcode:
Note that there is a change in transitive includes. Worth the risk for maint-3.10 at this point in my opinion.

argilo:
Agreed that it should be low-risk to remove `integer_math.h`. It's a recent addition which appeared in 3.9: https://github.com/gnuradio/gnuradio/pull/3337

mormj (approved):

PR #5611, opened 2022-02-24
Add in config blocks with validation

jsallay:
Signed-off-by: John Sallay <jasallay@gmail.com>

# Pull Request Details

## Description

It is frequently useful to use a config file to set the parameters for a flowgraph.  In GRC, a variable config block can be used, but there are several shortcomings.  Each config parameters in read in separately and requires a different block to be instantiated.  It also requires that a default value is specified, even if there isn't a sensible default.

I added in a few blocks, that, I believe, add to a cleaner user experience and allow for validation of the config file.  A json schema file can be used to ensure that all required fields are present and have valid values.  The config file is available in GRC as a dictionary, which can be used directly in parameters for other blocks.  i.e specify the Sample Rate as config["samp_rate"].

## Testing Done

I built my branch from source and  manually tested in GRC.  I created sample json, yaml, and schema files.  I created a simple flowgraph and ensured that it worked as expected in GRC and that the python produced was valid and worked.  Note that these blocks could be extended to work in c++, but that would be a lot more effort and introduce new dependencies.  These blocks are presently python only.

## Checklist

### DIFF
--- a/grc/blocks/json_config.block.yml
+++ b/grc/blocks/json_config.block.yml
@@ -1,0 +1,37 @@
+id: json_config
+label: Json Config
+flags: [ show_id, python ]
+
+parameters:
+-   id: config_file
+    label: Config File
+    dtype: file_open
+    default: default
+-   id: schema_file
+    label: Config Schema
+    dtype: file_open
+    default: ""
+# Not my favorite thing because it doesn't explicitly close the file, but
+# it should be okay.  The garbage collector will take care of it.
+value: "${ json.load(open(config_file)) }"
+
+asserts:
+#- ${ schema_file == "" or validate(value, json.load(open(schema_file))) is None }
+- ${ schema_file == "" or validate(json.load(open(config_file)), json.load(open(schema_file))) is None }
+
+templates:
+    imports: |-
+        import json
+        from jsonschema import validate
+    var_make: "with open(${config_file}) as fid:\n\
+              \    self.${id} = ${id} = json.load(fid)\n\
+              self.${id}_schema = ${id}_schema = ${schema_file}\n\
+              if ${id}_schema:\n\
+              \    with open(${id}_schema) as fid:\n\
+              \        validate(${id}, json.load(fid))"
+
+documentation: |-
+    This block represents a json config file, that is read in as a dictionary.
+
+
+file_format: 1
--- a/grc/blocks/json_config.block.yml
+++ b/grc/blocks/json_config.block.yml
@@ -22,6 +22,8 @@ templates:
     imports: |-
         import json
         from jsonschema import validate
+    # Note that yaml makes it really hard to insert spaces at the beginning of the line.
+    # The "\" char below is used to escape the space at the beginning of a line.
     var_make: "with open(${config_file}) as fid:\n\
               \    self.${id} = ${id} = json.load(fid)\n\
               self.${id}_schema = ${id}_schema = ${schema_file}\n\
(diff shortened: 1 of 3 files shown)


777arc:
Cool idea!

mormj:
Love this.

grc/blocks/json_config.block.yml
```diff
@@ -0,0 +1,37 @@
+id: json_config
+label: Json Config
```
777arc, line 2: JSON Config

777arc:
Because all built-in blocks have a link to a wiki page, see below, that's the best spot to stick any usage related docs and example flowgraphs, although the yaml file should still explain what the block does and how any params are used.  I would update the Json > JSON before we create the wiki page though, because the wiki page titles match the label.

Screenshot from 2022-02-26 14-49-03

grc/blocks/json_config.block.yml
```diff
@@ -0,0 +1,37 @@
...
+flags: [ show_id, python ]
+
+parameters:
+-   id: config_file
+    label: Config File
+    dtype: file_open
+    default: default
```
777arc, line 9: probably better off with an empty string here instead

grc/blocks/json_config.block.yml
```diff
@@ -0,0 +1,37 @@
...
+        from jsonschema import validate
+    var_make: "with open(${config_file}) as fid:\n\
+              \    self.${id} = ${id} = json.load(fid)\n\
+              self.${id}_schema = ${id}_schema = ${schema_file}\n\
+              if ${id}_schema:\n\
+              \    with open(${id}_schema) as fid:\n\
+              \        validate(${id}, json.load(fid))"
+
+documentation: |-
+    This block represents a json config file, that is read in as a dictionary.
```
777arc, line 34: First time using the block, some questions pop in my head like 1) is the schema optional 2) is the json supposed to be only 1 level deep 3) you say it's read in as a dictionary but what does that mean in GR sense, like does each field become a variable? 

I think we should present people with the most basic example of using it, so for example:
1 - Make a new file called example.json with the contents:
{
  "amplitude": 0.1
}
2 - Add the JSON Config block and point Config File to that file
3 - Make a signal source, and for Amplitude enter the value json_config_0[ 'amplitude']
4 - Connect it to a Time Sink through a Throttle and you'll see the amplitude show up as 0.1

777arc:
Went ahead and started the wiki page https://wiki.gnuradio.org/index.php?title=JSON_Config it assumes Json gets swapped to JSON for the label

grc/blocks/json_config.block.yml
jsallay: Done

grc/blocks/json_config.block.yml
jsallay: Where should I put this example?  I could put it right in the doc string.  For example,
```
With an input file of 
{
  "samp_rate": 1e6,
  "center_freq": 8e9 
}
```
We can access the data, like `config["samp_rate"]`

777arc: see https://wiki.gnuradio.org/index.php?title=JSON_Config#Example_Flowgraph

777arc: Wouldn't hurt to also have an example json in the yaml as well, and it's fine if it's different example than the wiki

grc/blocks/yaml_config.block.yml
```diff
@@ -0,0 +1,44 @@
+id: yaml_config
+label: Yaml Config
```
777arc, line 2: YAML =)

777arc:
I love the YAML block, in fact I'll personally be more likely to use that than the JSON.  The wiki docs can always follow the merge so we can focus on the code portion for now.  I like the idea of each .yml file in the documentation section explaining what the 2 params do, and have an example json/yaml file and then say how to actually use the values in another block's param, that way the user doesn't have to guess at how it works.

grc/blocks/json_config.block.yml
jsallay: Added in examples

mormj (approved):
Love it - merging!

willcode:
Noticed that this imports from `jsonschema`. On my system, `python3-jsonschema` is only there because `docker-compose` requires it. Have we identified this as a requirement?

mormj:
Good point - probably need to add a check to the cmake and make installation of the grc dependent on the output.  Creating an issue

PR #5615, opened 2022-03-02
grc: add .desktop suffix to the desktop launchable, other metainfo fixes

### LINKED ISSUE #5614: GNU Radio Companion has broken AppStream metadata
### What happened?

GNU Radio Companion has broken AppStream metadata because its metainfo file name is not matching the desktop file name and AppStream parsers have no way to match the correct desktop file. There is a launchable tag to fix this already present, but it seems to miss the .desktop suffix which should be there.

### System Information

OS: Fedora 35 Workstation
GR Installation Method: native distro packages - rpm, fedora

### GNU Radio Version

3.9  (maint-3.9)

### Specific Version

3.9.2.0

1. Open "GNOME Software" and try searching for "GNU Radio". No results are found.
2. Install "gnuradio" manually using dnf, reboot your computer and open "GNOME Software" again. Switch to the "Installed" tab and look at the "GNU Radio Companion" entry. It is displayed now, but its icon is missing.

### Relevant log output

_No response_

AsciiWolf:
Closes #5614.

Also move the tag to a better position (where most other apps have their launchable tags) in the file.

Without this, AppStream is not able to match the correct desktop file leading to missing icon and other issues in metadata.

### DIFF
--- a/grc/scripts/freedesktop/org.gnuradio.grc.metainfo.xml
+++ b/grc/scripts/freedesktop/org.gnuradio.grc.metainfo.xml
@@ -9,6 +9,7 @@
 -->
 <component type="desktop-application">
 <id>org.gnuradio.grc</id>
+<launchable type="desktop-id">gnuradio-grc.desktop</launchable>
 <metadata_license>CC0-1.0</metadata_license>
 <project_license>GPL-3.0+</project_license>
 <name>GNU Radio Companion</name>
@@ -39,7 +40,6 @@
   <category>HamRadio</category>
   <category>DataVisualization</category>
 </categories>
-<launchable type="desktop-id">gnuradio-grc</launchable>
 <screenshots>
   <screenshot type="default">
     <caption>GNU Radio Companion</caption>
--- a/grc/scripts/freedesktop/org.gnuradio.grc.metainfo.xml
+++ b/grc/scripts/freedesktop/org.gnuradio.grc.metainfo.xml
@@ -50,7 +50,6 @@
 <url type="homepage">https://www.gnuradio.org</url>
 <url type="help">https://wiki.gnuradio.org/</url>
 <url type="bugtracker">https://github.com/gnuradio/gnuradio/issues</url>
-<url type="contact">mailto:discuss-gnuradio@gnu.org</url>
 <url type="contact">https://lists.gnu.org/mailman/listinfo/discuss-gnuradio</url>
 <url type="contact">https://chat.gnuradio.org</url>
 <provides>


AsciiWolf:
Oops, I am dumb. :-) I overlooked that there is already the launchable tag in the file. However, it is missing the .desktop suffix that should be there. I will update the PR to reflect that.

AsciiWolf:
I have also removed the contact tag that does not appear to be correct (only http(s) seems to be supported) and added OARS metadata that are required by many AppStream parsers nowadays.

Here is an `appstreamcli validate grc/scripts/freedesktop/org.gnuradio.grc.metainfo.xml --explain` output before the change:
```
I: org.gnuradio.grc:~: content-rating-missing
   This component has no `content_rating` tag to provide age rating information. You can generate
   the tag data online by answering a few questions at https://hughsie.github.io/oars/

E: org.gnuradio.grc:53: web-url-expected mailto:discuss-gnuradio@gnu.org
   A web URL was expected for this value.
```

And after:
```
Validation was successful: pedantic: 1
```

marcusmueller (commented):
seems right.

We might want to rename the file `gnuradio-grc.desktop` to `org.gnuradio.grc.desktop` *separately* (not here), for consistency reasons.

grc/scripts/freedesktop/org.gnuradio.grc.metainfo.xml
```diff
@@ -50,11 +50,11 @@
 <url type="homepage">https://www.gnuradio.org</url>
 <url type="help">https://wiki.gnuradio.org/</url>
 <url type="bugtracker">https://github.com/gnuradio/gnuradio/issues</url>
-<url type="contact">mailto:discuss-gnuradio@gnu.org</url>
```
marcusmueller, line 53: :/ 

I kind of don't... like this change (don't heed me, I'm just a grumpy old guy), because this is something the metainfo schema checker incorrectly marks as error; It's perfectly [legal](https://freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-url) to use a `mailto:` URL, and it'd be honest, because, hey, that's how we as community actually communicate, for an important part. However, it's discouraged, by freedesktop, as Desktop environments seem to have a hard time making it easy for users to send webmail, and now that is a problem of software maintainers ;) I think they might be right though, if not in the stated reasoning, nobody looks at the software description page and then spontaneously wants to send an email.

grc/scripts/freedesktop/org.gnuradio.grc.metainfo.xml
```diff
@@ -50,11 +50,11 @@
...
 <url type="contact">https://lists.gnu.org/mailman/listinfo/discuss-gnuradio</url>
 <url type="contact">https://chat.gnuradio.org</url>
 <provides>
   <binary>gnuradio-companion</binary>
   <binary>grcc</binary>
 </provides>
+<content_rating type="oars-1.1" />
```
marcusmueller, line 59: I'd agree, we're absolutely suited for all audiences. However, this feels *super* unnecessary for a desktop application. Also, it places the burden on the maintainers to check for any feature we might add in the future (e.g. chat integration, flow graph sharing) whether we might need adjust this setting – again, for no obvious (to me) gain.

I'd rather we not do that. Not a strongly held opinion, though

AsciiWolf: Well, the OARS metadata that I added are empty (which usually means "suited for all audiences" vs "unknown" when there are no OARS metadata). :-) As I mentioned, this is just to make the AppStream parsers happy because many of them now require OARS metadata and fail to parse the AppStream file when there are no OARS metadata in it. Many other desktop applications have already added empty OARS metadata to their AppStream files for the same reason.

AsciiWolf:
> We might want to rename the file `gnuradio-grc.desktop` to `org.gnuradio.grc.desktop` _separately_ (not here), for consistency reasons.

Renaming the desktop file to match the AppStream metadata file would be the best solution. Just don't forget to remove the [launchable tag](https://github.com/gnuradio/gnuradio/pull/5615/files#diff-53cc24f60520b15e513b867fda40fe0c79e80baff46e0b60dba1d8df0e07078eR12) from AppStream file afterwards. :-)

AsciiWolf:
@marcusmueller Feel free to let me know if you have any further questions regarding the AppStream metadata file. :-) If you want, we can merge just the launchable tag fix for now.

grc/scripts/freedesktop/org.gnuradio.grc.metainfo.xml
marcusmueller: oh, many? I was aware of only one standard, and that does not mandate these tags. Which parsers fail? I'd rather fix their bugs upstream (namely, not accepting valid data) than adding a persistent maintenance burden downstream. Enforcing meaningless metadata that is likely to outdate is strictly a bug for a metadata parsing ecosystem!

marcusmueller: Don't take this the wrong way, please. The problem I'm having here is that there's a somewhat clear file format standard – adhering to anything stricter than that seems OK. But if tools consuming the kind of file refuse to work with files that adhere to the standard on own, undocumented rules, then that's a problem of these tools and should not be exported to projects contributing metadata :)

AsciiWolf: There is only one standard specification, however OARS metadata are a special case. :-) They are [considered optional](https://freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-content_rating) by the standard, but since there was a lot of discussion and plans to make OARS mandatory in the future, some parsers and validators started requiring OARS. A good example of this is [Flathub](https://github.com/flathub/flathub/issues/158).

AsciiWolf: Anyway, as I said, if you want, I can remove the OARS tag. I would like to at least keep the second part of the commit that removes the mailto url since that one was considered an error by appstreamcli validator and was actually never properly supported, but it is also up to you.

AsciiWolf: By the way, for example [Gqrx](https://gqrx.dk) has the same OARS metadata: https://github.com/gqrx-sdr/gqrx/blob/master/dk.gqrx.gqrx.appdata.xml#L31

AsciiWolf:
Here is an AppStream log from Debian: https://appstream.debian.org/sid/main/issues/gnuradio.html

1. The _no-metainfo_ warning and _gui-app-without-icon_ error are caused by the missing .desktop suffix in launchable.
2. The _asv-web-url-expected_ warning is caused by the mailto in url which is deprecated and does not seem to be supported by most parsers anymore.
3. The _asv-content-rating-missing_ hint is caused by missing OARS metadata.

All should be fixed by this PR. :-)

AsciiWolf:
@marcusmueller Could you please take another look at this PR (see comment https://github.com/gnuradio/gnuradio/pull/5615#discussion_r820154734 and https://github.com/gnuradio/gnuradio/pull/5615#issuecomment-1060599421)? Thanks! :-)

AsciiWolf:
/cc @mormj

I see that you did review the [original PR](https://github.com/gnuradio/gnuradio/pull/3621). Could you please take a look at this one as well? Thanks!

marcusmueller:
Hey, sorry to take this long to get back to you – this required a bit of internal discussion.

So, I'd like to take you up on the offer to remove the OARS addition from this commit – it's really a separate thing :) The rest of your commit *fixes* existing metadata, but that OARS tag adds new metadata.

Then: We find it hard to fit ourselves into the categories that OARS offers. As such, we'd prefer to not be put into the categories, which really don't seem to fit the kind of software GNU Radio is well.

Seeing that the absence of this standard-wise non-mandatory piece of text that we can't reasonably guarantee might exclude us from app stores, I'd ask you make a separate commit that states the software is *unrated* (i.e., `none` in all aspects; it's a design defect in the standard that stating nothing means "we certify everything is fine", IMHO).

PR #5617, opened 2022-03-03
gr-blocks: Squelch debug messages in file sink.

drmpeg:


argilo:
Would it be enough to just change the log level to `debug`? I'm wary of `#if` because conditionally-compiled code tends to rot.

drmpeg:
It still prints with log level ```debug```. I'm open to other ways to address this, but it has to be a default and not some user level setting that has to be added (not even sure how to do that myself).

argilo:
I had a peek and it looks like the problem is that the log level of `file_sink_base`'s `d_base_logger` logger is never set. In other places, logging levels are set with `gr::configure_default_loggers`.

drmpeg:
Thanks for the tip. Updated with ```gr::configure_default_loggers```.

argilo (approved):

PR #5619, opened 2022-03-04
digital: add generic CRC blocks

daniestevez:
## Description

This PR includes a class that computes a generic CRC and two blocks that use this class to compute and append the CRC to a PDU, and check the CRC of a PDU. These are similar to the current crc16_async and crc32_async blocks, but the user can define the CRC parameters.

I have recently added these blocks to gr-satellites, and figured that they would be useful in-tree. I've mainly design the blocks according to my use case for gr-satellites, so the main goal of this PR (besides eventually merging these in-tree) is to gather feedback about the user interface and coding style, and what would be best for broader use cases.

The notation to define the CRC parameters is taken from [this online calculator](http://www.sunshine2k.de/coding/javascript/crc/crc_js.html), which I find quite useful when working with CRCs. Still, I often find slightly confusing some of the CRC parameters, particularly those regarding endianness, so I'm open to suggestions on how we may improve this (documentation and examples would help, for sure).

The CRC calculation is done using a table-driven byte-by-byte algorithm. I have deliberately chosen not to use boost for this in order to reduce the number of things we use boost for.

Something that is missing from this PR is C++ templates in the GRC YAMLs of the blocks. These can be added later in this PR when we are all happy with the blocks API.

The commit message is as follows (given here for convenience):

```
This adds a crc class that implements a generic CRC calculation using
a table-driven byte-by-byte approach. The CRC parameters are set when
instancing the class.

The crc class is used in the CRC Append and CRC Check blocks, which
take a PDU as input, and either compute the CRC of that PDU and append
it, or check the CRC included in the PDU (sending out the PDU in one
of two outputs according as to whether the CRC was correct).

The CRC blocks can optionally ignore a certain amount of "header" bytes
in the CRC calculation (used for packets where the CRC only affects the
"payload" and not the header). Additionally, the CRC Check block can
optionally remove the CRC after checking it.
```

The blocks CRC Append and CRC Check, as well as the class `crc` have been added to gr-digital.

## Testing Done

There is a QA test called `qa_crc.py`. This checks that the CRC calculation is correct for a few common CRC codes, and that the CRC Check block is able to detect whether the CRC is correct or not.

## Checklist

- [ ] I have updated [the documentation](https://wiki.gnuradio.org/index.php/Main_Page#Documentation) where necessary. -> No updates done to the documentation yet. Potentially these blocks should be added to the [list of blocks](https://wiki.gnuradio.org/index.php?title=Category:Block_Docs). This should be done after "freezing" the API for the blocks.


daniestevez:
Something else that needs to be changed before merging is the logging (currently it's the "older" logging framework, to keep it backwards-compatible in gr-satellites).

mormj:
@daniestevez - this is awesome!  Should the crc_16 and crc_32 blocks be marked as deprecated with some guide how to use your general purpose crc blocks?  Or those blocks made to wrap this?

daniestevez:
I think all the functionality of crc16_async_bb and crc32_async_bb is covered by these new blocks, so there's no point in keeping around the C++ code for the old blocks. Regarding the GRC blocks, I don't have a clear idea about what's best. I can think of several possibilities:

* Make the change transparent for GRC users. The old GRC blocks generate code using the new blocks.
* Deprecate the old GRC blocks. The GRC blocks I'm including for these new blocks already have defaults for CRC-32 (but I haven't checked if the endianness is exactly the same as in crc32_async_bb). Document somewhere what parameters can be used with the new blocks to obtain the same results as with crc16_async_bb.
* Extend/add to the GRC blocks for the new blocks so that the user can select a "standard" CRC from a drop down list. In this case, the old crc16 and crc32 would be options in that list, so it's probably best to deprecate the old GRCs. I don't know how many CRC codes is sensible to include in this list. Potentially we could include all the ones from the [online calculator](http://www.sunshine2k.de/coding/javascript/crc/crc_js.html) (this is probably useful for people finding a matching CRC code by trial and error, which I've had to do a few times).

gr-digital/include/gnuradio/digital/crc.h
```diff
@@ -0,0 +1,86 @@
...
+private:
+    std::array<uint64_t, 256> d_table;
+    unsigned d_num_bits;
+    uint64_t d_mask;
+    uint64_t d_initial_value;
+    uint64_t d_final_xor;
+    bool d_input_reflected;
+    bool d_result_reflected;
+
+    uint64_t reflect(uint64_t word);
```
rear1019, line 80: Nit-pick: Add `const` specifier to the member function.

gr-digital/include/gnuradio/digital/crc.h
```diff
@@ -0,0 +1,86 @@
...
+        bool result_reflected);
+    ~crc();
+
+    /*!
+     * \brief Computes a CRC
+     *
+     * \param data the input data for the CRC calculation
+     * \param len the length in bytes of the data
+     */
+    uint64_t compute(const uint8_t* data, std::size_t len);
```
rear1019, line 59: It might be useful to have a stream like interface which doesn’t require *all* of the data to be available, e.g. `push_byte()/push_bytes()` followed by a `finalize()`.

daniestevez:
@rear1019 Thanks for your suggestions. I'm not so savvy on C++, so these are quite helpful.

gr-digital/include/gnuradio/digital/crc.h
daniestevez: I see that it might be useful, and often CRC libraries have this. However, for the intended use case it's not necessary, since all the data is already available at once in a PDU, so I prefer not to add it at the moment. If a new use case for this pops up where not all the data is available at once, then I think it's easy to add the required functions to this class. The `compute()` method would become a wrapper over `push()` followed by `finalize()`.

mormj:
@daniestevez - again, this is great!  Are you all set with this PR?

daniestevez:
@mormj since this has sitting for a couple weeks and no one has complained about the API or UI of this blocks, I think we can go ahead. Before merging, I want to add the C++ templates for the GRC blocks. I'll do this in at most two days, and I'll squash everything together. Then it's ready to go.

daniestevez:
C++ templates added (this was easy). Now I'll squash everything together.

willcode:
Holding off on 3.9 backport because of crc vs crc32 conflict in some of the cmake files and one pybind file. The conflict itself is simple enough. I'm wondering if there is some duplication that needs to be removed too before this is backported. @daniestevez?

daniestevez:
@willcode There was the idea to eventually mark the old CRC-32 and CRC-16 blocks as deprecated, but this is something for the long run and probably not something that would affect 3.9. Other than this, I don't think there is any duplication to remove.

What sort of conflict is there with 3.9? I'm curious about why there was no conflict in 3.10 and master.

willcode:
@daniestevez 3 CMakeLists.txt with
```
<<<<<<< HEAD
    crc32.h
=======
    crc.h
>>>>>>> 1c4ef3dac (digital: add generic CRC blocks)
```
where there is already a `crc32.h` and `crc.h` is added, and
```
<<<<<<< HEAD
void bind_crc32(py::module&);
=======
void bind_crc(py::module&);
>>>>>>> 1c4ef3dac (digital: add generic CRC blocks)
``` in python_bindings.cc.

I wasn't sure whether replacing the `crc32.h` with `crc.h` would mess something up.

willcode:
@daniestevez Leaving both didn't work (see #5694). It appears that `crc` is a module in 3.9. Perhaps this just needs an adjustment in QA? I didn't look too closely.

PR #5628, opened 2022-03-06
pdu: Fix flaky tests

argilo:
## Description
`qa_add_system_time`/`test_002_timing` and `qa_time_delta`/`test_002_normal` routinely fail CI, usually on Windows. Some recent failed runs:

https://github.com/gnuradio/gnuradio/runs/5429110249?check_suite_focus=true
https://github.com/gnuradio/gnuradio/runs/5439304121?check_suite_focus=true

In both cases, the tests fail because the error tolerance is set too low.

None; this is a CI-only change.

## Testing Done
I ran these test locally to verify that they run correctly, and fail if the `delta` value are set very low.

## Checklist

### DIFF
--- a/gr-pdu/python/pdu/qa_add_system_time.py
+++ b/gr-pdu/python/pdu/qa_add_system_time.py
@@ -71,7 +71,7 @@ class qa_add_system_time(gr_unittest.TestCase):
                                         pmt.intern("systime"),
                                         pmt.from_double(0.0)))
         # should be sufficient tolerance
-        self.assertTrue(((t1 - t0) - 1) < 0.05)
+        self.assertAlmostEqual(t0, t1 - 1, delta=0.2)
 
 
 if __name__ == '__main__':
--- a/gr-pdu/python/pdu/qa_time_delta.py
+++ b/gr-pdu/python/pdu/qa_time_delta.py
@@ -88,8 +88,8 @@ class qa_time_delta(gr_unittest.TestCase):
         time_tag = pmt.dict_ref(a_meta, pmt.intern("system_time"), pmt.PMT_NIL)
         delta_tag = pmt.dict_ref(a_meta, pmt.intern(
             "sys time delta (ms)"), pmt.PMT_NIL)
-        self.assertAlmostEqual(tnow, pmt.to_double(time_tag), delta=60)
-        self.assertAlmostEqual(10000, pmt.to_double(delta_tag), delta=10)
+        self.assertAlmostEqual(tnow - 10.0, pmt.to_double(time_tag), delta=1e-6)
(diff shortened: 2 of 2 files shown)


gr-pdu/python/pdu/qa_add_system_time.py
```diff
@@ -71,7 +71,7 @@ def test_002_timing(self):
                                         pmt.intern("systime"),
                                         pmt.from_double(0.0)))
         # should be sufficient tolerance
-        self.assertTrue(((t1 - t0) - 1) < 0.05)
+        self.assertAlmostEqual(t0, t1 - 1, delta=0.2)
```
argilo, line 74: In addition to increasing the tolerance, I also changed the assertion to `assertAlmostEqual` so that the error message will display the two input values and the difference between them.

gr-pdu/python/pdu/qa_time_delta.py
```diff
@@ -88,8 +88,8 @@ def test_002_normal(self):
         time_tag = pmt.dict_ref(a_meta, pmt.intern("system_time"), pmt.PMT_NIL)
         delta_tag = pmt.dict_ref(a_meta, pmt.intern(
             "sys time delta (ms)"), pmt.PMT_NIL)
-        self.assertAlmostEqual(tnow, pmt.to_double(time_tag), delta=60)
-        self.assertAlmostEqual(10000, pmt.to_double(delta_tag), delta=10)
+        self.assertAlmostEqual(tnow - 10.0, pmt.to_double(time_tag), delta=1e-6)
```
argilo, line 91: This value is not subject to timing noise, so `delta` can be decreased drastically once the right comparison is made.

gr-pdu/python/pdu/qa_time_delta.py
```diff
@@ -88,8 +88,8 @@ def test_002_normal(self):
...
+        self.assertAlmostEqual(10000, pmt.to_double(delta_tag), delta=50)
```
argilo, line 92: Recent test failures have a delta of less than 20 ms, so hopefully 50 will be enough to prevent failures.

gr-pdu/python/pdu/qa_add_system_time.py
marcusmueller: Wow, still a lot of tolerance, but yeah, timing here is pretty uncertain...

marcusmueller (approved):

gr-pdu/python/pdu/qa_add_system_time.py
jacobagilbert: yea, these are benchmarking tools so I think this should catch most egregious errors. possibly this should be run many times and the mean error evaluated but this is an improvement.

jacobagilbert (approved):

Issue #5412, opened 2021-12-20
QT GUI Number Control 'value out' emits previous value for 'value in' messages or variable changes


muaddib1984:
### What happened?

In the QT GUI digital number control widget example in GNURadio 3.9, the 'value out' port emits it's previously stored value when either a new input message arrives or (if) a variable is used in the 'value' field and it's changed from another source (i.e. range widget).

### System Information

OS: Ubuntu20.04
GR Installation Method: pybombs

### GNU Radio Version

3.9 (maint-3.9)

### Specific Version

3.9.4.0

### Steps to Reproduce the Problem

Using the Digital Number Control example [here:](https://github.com/gnuradio/gnuradio/blob/master/gr-qtgui/examples/test_digitalnumcontrol.grc)
Start with 90e6 as the initial value of 'freq'  for both the message edit box and the number control, then change the value of 'freq' in the message edit box to 95e6. The number control display will change to 95,000,000 but the message emitted from the 'value out' port will be 90e6. If you change the edit box again, now 95e6 will be emitted from the 'value out' port.

willcode:
The problem has to do with `self.cur_freq` not being set immediately when the frequency is set by the message handler.

PR #5643, opened 2022-03-13
DigitalNumberControl: Always emit new frequency

### LINKED ISSUE #5141: qtgui: Qt digital number control output message lags input message by one
Well, it's hard to find a suitable heading sometimes. Whatever, when I tick the numbers on the GUI, the new frequency is sent on the output port, but If I send a new frequency to the dial, via the input port,  the dial then sends the *old* frequency on its output port. This may be intended behaviour, but feels a bit nonintuitive to me.
Actually, when looking into the code it seems like the intention is to send the new frequency. I suspect this happens because there is a race condition somewhere.

Adding a demo flowgraph. You may have to expand its window when running. See the console log as well.
The TX FREQUENCY dial is there just to show what RX frequency sends.

[dncbug.grc.zip](https://github.com/gnuradio/gnuradio/files/7285827/dncbug.grc.zip)

Initial view:
image

After clicking the red bar:

image

Clicking again:
image

df7cb:
While the DigitalNumberControl widget correctly emitted the new
frequency value when the widget was clicked on, it would report the old
frequency when a new value was set via an incoming message.

Close #5141 #5412

This duplicates the fix from #5592, but the other PR is about something else, and contains this fix only as a side-effect.

## Testing Done

It fixes the use case I have for this widget: being able to change the frequency by solely relying on the widget, while submitting frequency changes from other sources (double clicking on the waterfall or spectrum widgets, or messages from other sources).

## Checklist


willcode:
@df7cb @mormj Let's go with #5592 so we don't have to go back and change that one.

willcode:
Merged #5592

df7cb:
Thanks, glad it's getting fixed!

PR #5626, opened 2022-03-06
runtime: Remove boost::format and modernize logging

argilo:
## Description
This continues the work started in https://github.com/gnuradio/gnuradio/pull/5270. Here I've removed all remaining usage of `boost::format` from gnuradio-runtime, and modernized some more of the logging.

Together with my other pull requests, this removes all usage of `boost::format` from GNU Radio, with the exception of the following helper class in the logger:

https://github.com/gnuradio/gnuradio/blob/3349d483a1f7e5db0eea042815d237810373857b/gnuradio-runtime/include/gnuradio/logger.h#L276-L286

That can be removed as soon as the other pull requests are merged.

## Related Issue
* #5270

## Which blocks/areas does this affect?
* `thread_body_wrapper`
* `basic_block`
* `block_detail`
* `block_executor`
* `rpcpmtconverter`
* `flat_flowgraph`
* `flowgraph`
* `hier_block2_detail`
* `top_block`

## Testing Done
1. Set `debug_level = trace` and `log_level = trace`, ran a sample flowgraph (including a hierarchical block) before & after, and verified that the debug output matched (apart from differing process IDs).
2. Ran Gqrx and verified that trace-level debug output looks sane.

I'd appreciate help with additional testing.

## Checklist
- [x] I have read the [CONTRIBUTING document](https://github.com/gnuradio/gnuradio/blob/main/CONTRIBUTING.md).
- [x] I have squashed my commits to have one significant change per commit. 
- [x] I [have signed my commits before making this PR](https://github.com/gnuradio/gnuradio/blob/main/CONTRIBUTING.md#dco-signed)
- [x] My code follows the code style of this project. See [GREP1.md](https://github.com/gnuradio/greps/blob/main/grep-0001-coding-guidelines.md).
- [ ] ~~I have updated [the documentation](https://wiki.gnuradio.org/index.php/Main_Page#Documentation) where necessary.~~
- [x] ~~I have added tests to cover my changes~~, and all previous tests pass.

PR #5645, opened 2022-03-14
runtime: Remove boost::format support from the logger

argilo:
## Description
Now that `boost::format` has been removed from all GNU Radio modules (in #5626 and others), it should no longer be necessary to support `boost::format` in the logger.

* logging

## Testing Done
I build GNU Radio and tested a few flow graphs with the log levels set to `trace`. Everything still worked as before.

## Checklist

### DIFF
--- a/gnuradio-runtime/include/gnuradio/logger.h
+++ b/gnuradio-runtime/include/gnuradio/logger.h
@@ -42,8 +42,6 @@ using logger_ptr = std::shared_ptr<void>;
 
 #include <spdlog/sinks/dist_sink.h>
 
-#include <boost/format.hpp>
-
 namespace gr {
 using log_level = spdlog::level::level_enum;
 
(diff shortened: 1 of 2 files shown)


marcusmueller:
I (obviously) like this!

marcusmueller (approved):
quite a breaking change :) I wonder whether we should be offering a header containing the converter code) that people can include in their OOTs in case they want to continue using boost::format, but this might just be too much of an appeasement strategy that might not benefit anyone (the number of OOTs that did extensive logging was/is limited to begin with)

argilo:
Oops, I forgot to remove the `#include`. That's fixed now.

argilo:
You're right that this is a breaking change. I suppose it might be worthwhile to check over CGRAN to see whether any public OOTs are using `boost::format` logging. If so, I suppose we could skip backporting the change to 3.10 and make the breaking change in 3.11.

mormj:
> I suppose we could skip backporting the change to 3.10 and make the breaking change in 3.11.

Since boost isn't going anywhere for 3.10, it probably makes sense to not backport this particular change

willcode:
This breaks gr-satellites.

argilo:
I fixed the branch name in https://github.com/gnuradio/gr-recipes/commit/7cc5d8b774657c44129761b2f2d9a0acfd4caab3.

argilo:
I've also updated gr-satellites' logging here: https://github.com/daniestevez/gr-satellites/pull/350

daniestevez:
> This breaks gr-satellites. Ah, the pybombs recipe still calls for master. The main branch works.

I wonder why main works and master doesn't. The rename from master to main in gr-satellites was done on February this year, and I don't see any changes since then that would affect this. In any case, https://github.com/daniestevez/gr-satellites/pull/350 will fix any breakage.

argilo:
> I wonder why main works and master doesn't.

There was some breakage because GNU Radio no longer included `boost/format.hpp`. That was fixed by adding the necessary includes in https://github.com/daniestevez/gr-satellites/pull/345.

Issue #5198, opened 2021-10-14
GRC: Bus Ports are Broken in FEC flowgraphs


mormj:
### What happened?

Bus ports in 
               'ber_curve_gen_ldpc.grc',
               'ber_curve_gen.grc',
               'polar_ber_curve_gen.grc',
               'tpc_ber_curve_gen.grc',
are broken

![image](https://user-images.githubusercontent.com/34754695/137306296-0abd2ebb-0fb1-4592-bf97-3f6aa32b4cd3.png)

### System Information

OS: (Linux Distro, Mac, Windows)
GR Installation Method: (native distro packages - apt, rpm, etc., Conda, Source

### GNU Radio Version

3.9 (maint-3.9)

### Specific Version

3.9.4.0

### Steps to Reproduce the Problem

Open up one of these example flowgraphs:

               'ber_curve_gen_ldpc.grc',
               'ber_curve_gen.grc',
               'polar_ber_curve_gen.grc',
               'tpc_ber_curve_gen.grc',

GRC is unable to use fields from the blocks (num_curves) in the expressions used for other fields. 

### Relevant log output

```shell
Param - Bus_Structure_Sink(bus_structure_sink):
	Value "list(map(lambda b:list(map(lambda a:b * len(esno) * 2 + a, range(len(esno)*2))), range(num_curves)))" cannot be evaluated:
	name 'num_curves' is not defined
```

haakov:
Actually, parameter values are evaluated in the flowgraph's context:
https://github.com/gnuradio/gnuradio/blob/48120c39228ead7941140be310580188c29b61c1/grc/core/params/param.py#L213
In contrast, parameter attributes like hide and multiplicity are evaluated in the block's context and can thus use the block's other parameters. 

The bus_structure_sink parameter for these blocks was likely never evaluated successfully, the error was just masked by the fact that raw parameters weren't validated.

PR #5646, opened 2022-03-14
grc: fails to evaluate a block parameter that references another parameter of the block

### LINKED ISSUE #5627: gr-qtgui /gr-fec gnuradio/gr-fec/examples/ber_curve_gen.grc fails to build
### What happened?

Trying to build the gnuradio/gr-fec/examples/ber_curve_gen.grc fails as the qtgui_ber_sink fails to evaluate.

The error message is:

Param - Bus_Structure_Sink(bus_structure_sink):
	Value "list(map(lambda b:list(map(lambda a:b * len(esno) * 2 + a, range(len(esno)*2))), range(num_curves)))" cannot be evaluated:
	name 'num_curves' is not defined

but num_curves is set to 1 in the example.

### System Information

OS: Fedora 35 , build from scratch

### GNU Radio Version

3.11-git (main)

### Specific Version

0.0.0.0

Just open [ber_curve_gen flowgraph ](https://github.com/gnuradio/gnuradio/blob/main/gr-fec/examples/ber_curve_gen.grc) in grc

### Relevant log output

_No response_

dl1ksv:
# Pull Request Details

## Description

If you have a block with

param1: 3
param2: 3 * param1

the evaluation of param2 fails, as only variables of the flowgraph are used during evaluation.
This pr stores previous evaluated paramters of the block and uses them during evaluation,
so param2 will be correctly evaluated.

## Related Issue

Fixes #5627

## Testing Done

Run
[ber_curve_gen](https://github.com/gnuradio/gnuradio/blob/main/gr-fec/examples/ber_curve_gen.grc)
in grc.
## Checklist

### DIFF
--- a/grc/core/FlowGraph.py
+++ b/grc/core/FlowGraph.py
@@ -235,6 +235,7 @@ class FlowGraph(Element):
         # to get rid of entries of blocks that
         # are no longer valid ( deleted, disabled, ...)
         self.namespace.clear()
+        self._eval_cache.clear()
         # Load imports
         for expr in self.imports():
             try:
@@ -292,8 +293,6 @@ class FlowGraph(Element):
                 log.exception('Failed to evaluate variable block {0}'.format(
                     variable_block.name), exc_info=True)
                 pass
-
-        self._eval_cache.clear()
 
     def evaluate(self, expr, namespace=None, local_namespace=None):
         """
--- a/grc/core/params/param.py
+++ b/grc/core/params/param.py
@@ -216,7 +216,8 @@ class Param(Element):
                         if scale_factor in self.scale:
                             expr = str(float(expr[:-1]) *
                                        self.scale[scale_factor])
-                    value = self.parent_flowgraph.evaluate(expr)
+                    value = self.parent_flowgraph.evaluate(expr, self.parent.namespace, self.parent_flowgraph.namespace)
+                    self.parent.namespace[self.key] = value
                 except Exception as e:
                     raise Exception(
                         'Value "{}" cannot be evaluated:\n{}'.format(expr, e))


haakov:
This topic has been briefly discussed in an earlier issue: https://github.com/gnuradio/gnuradio/issues/5198

What happens if you have a "global" (flowgraph-wide) variable with the same name as a block parameter?

skoslowski (changes requested):
This is not bug! In my opinion, the parameters of a block should not depend on each other. Parameters are not variables. I recommend not to merge this.

Sorry, @dl1ksv for shooting down your idea (again). Here's my reasoning:
- The mental model here is that a block is like a (python) class (because is usually is) and the parameters end up as (kw)args of the init-function. Parameters do not reference each other. If they depend on each other, this is expressed by a variable (to be fair, or explicitly with an inline assigment). 
- How would a user know what the internal names of these Variables are? 
- What about shadowing, eval order, recursive evals?

Maybe I am missing some really compelling use-cases here.

dl1ksv:
@skoslowski I'm a bit surprised, that this should not be a bug, as in an earlier version of gnuradio ( 3.7,3.8, may be even 3.9) this flowgraph could be build, but maybe I'm wrong, as the screenshot does not show the running flowgraph.

See [ber_curve_usage_example](https://wiki.gnuradio.org/index.php?title=File:Ber_curve_gen_fg.png)

You are right, parameter should not depend on each other. I think I used a wrong example.

 The problematic code part is
https://github.com/gnuradio/gnuradio/blob/f44384ae1f061f257603b02288aef4db1b74d4be/gr-qtgui/grc/qtgui_ber_sink_b.block.yml#L314-L318

As bus_structure_sink is hiden, the user can't change the assignment.
In this case:  is bus_structure_sink a parameter or a ( block ) local variable ?

You can make the flowgraph work by adding two variables num_curves and esno. But how would you handle a flowgraph with two  ber_curve sinks with different parameter settings? Modifying the ber_curve sink , to be able to modify the bus_structure_sink parameter?

May be it's a useless discussion, as only ber_curve_sink is malformed

skoslowski:
If GRC really did allow resolving other parameters at some point, I would consider that a bug. Anyway, glad that is not the case anymore.

The bus ports code has always been problematic to say the least. I can see why it would benefit from accessing other parameters. I someone really wanted to refactor (rewrite) the bus ports implementation I would still suggest not to generally allow parameters to depend on each other. I'd make a special (bus ports specific, internal) dtype that behaves as desired (probably by delaying the eval until the ports are (re-)formed)

mormj:
Closing based on discussion so as to not affect general GRC behavior in order to patch bus ports

PR #5647, opened 2022-03-14
runtime: expose mechanism for testing host_buffer

mormj:
## Description
By forcing the default buffer type to host_buffer instead of buffer_double_mapped, we can more thoroughly test the single mapped buffers through all the QA.  This should not be enabled by default

## Related Issue
None

By default should have no impact, only enabled for testing purposes with a #define

## Checklist


dsorber (approved):
Looks good.

dsorber:
This is a good idea.  It would be very useful to be able to run the tests with both host_buffer and buffer_double_mapped back to back, but seeing as that would require compile-time changes that's not really possible.

PR #5649, opened 2022-03-18
CMake: Be less strict about deps checking

mbr0wn:
# Pull Request Details

This lets cmake run even if no boost, MPLIB, or Volk are present. It will disable all targets except for the documentation. This is useful if you have a dedicated machine to build docs.


cmake/Modules/FindMPLIB.cmake
```diff
@@ -11,7 +11,7 @@ if(GMP_FOUND)
     set(MPLIB_PC_ADD_LIBS ${GMP_PC_ADD_LIBS})
 else(GMP_FOUND)
     message(STATUS "GMP not found; this is not a problem if MPIR can be found.")
-    find_package(MPIR REQUIRED)
+    find_package(MPIR)
```
willcode, line 14: Will this fall through if not found, giving the "MPIR found" message and doing all the sets below?

mormj:
With a change like this I would be afraid we will get some confusing build errors from users where it appears to build correctly, but then there is nothing there.  Could this Docs Only build be enabled by a CMake flag, such as `cmake -DGR_DOCS_ONLY`?

cmake/Modules/FindMPLIB.cmake
mbr0wn: Hm, I think I'll clean up this section.

mbr0wn:
> With a change like this I would be afraid we will get some confusing build errors from users where it appears to build correctly, but then there is nothing there. Could this Docs Only build be enabled by a CMake flag, such as `cmake -DGR_DOCS_ONLY`?

You already have to specify CMake flags to not build without runtime, even with this change. If Boost or something equally difficult is missing, the default behaviour is still to fail:
`root@0e7eb0551ca4:/build# cmake ../uhd
-- The CXX compiler identification is GNU 9.4.0
-- The C compiler identification is GNU 9.4.0
-- Check for working CXX compiler: /usr/bin/c++
-- Check for working CXX compiler: /usr/bin/c++ -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Check for working C compiler: /usr/bin/cc
-- Check for working C compiler: /usr/bin/cc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Build type set to Release.
-- Build date set to Fri, 01 Apr 2022 09:57:19Z.
--
-- The build system will automatically enable all components.
-- Use -DENABLE_DEFAULT=OFF to disable components by default.
-- Could NOT find Git (missing: GIT_EXECUTABLE)
-- Looking for pthread.h
-- Looking for pthread.h - found
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD - Failed
-- Looking for pthread_create in pthreads
-- Looking for pthread_create in pthreads - not found
-- Looking for pthread_create in pthread
-- Looking for pthread_create in pthread - found
-- Found Threads: TRUE
CMake Error at /usr/share/cmake-3.16/Modules/FindPackageHandleStandardArgs.cmake:146 (message):
Could NOT find Boost (missing: Boost_INCLUDE_DIR date_time program_options
system regex thread) (Required is at least version "1.69")
/usr/share/cmake-3.16/Modules/FindPackageHandleStandardArgs.cmake:393 (_FPHSA_FAILURE_MESSAGE)
/usr/share/cmake-3.16/Modules/FindBoost.cmake:2179 (find_package_handle_standard_args)
cmake/Modules/GrBoost.cmake:84 (find_package)
CMakeLists.txt:58 (include)


-- Configuring incomplete, errors occurred!
See also "/build/CMakeFiles/CMakeOutput.log".
See also "/build/CMakeFiles/CMakeError.log".``
```

I can easily add another flag if you like, but I don't think people will accidentally add these sparsely (if at all) documented flags to enable just building the manual. At the very least, you need to add `-DGR_BOOST_REQUIRED=OFF` for this to fly. Would you consider that sufficient?

mormj:
No, I don't think an additional flag is necessary if the general use case won't pass silently but without building anything

mbr0wn:
> No, I don't think an additional flag is necessary if the general use case won't pass silently but without building anything if deps are not present. As long as there is some setting that gates this behavior

OK cool :) I take it you approve this general behaviour then?

mormj:
Yes, being able to compile a minimal subset of the code when the deps aren't there, for special purpose, makes sense to me.

willcode (approved):
OK if this is useful for someone. What can you actually build correctly if gnuradio-runtime doesn't build, though?

mbr0wn:
> OK if this is useful for someone. What can you actually build correctly if gnuradio-runtime doesn't build, though?

The manual, which is how I created https://gnuradiosdr.readthedocs.io/en/latest/

PR #5651, opened 2022-03-19
pdu: Replace Boost with standard C++

argilo:
## Description
gr-pdu uses Boost in only a few places. I've replaced them with standard C++ here.

* `boost::this_thread::sleep` -> `std::this_thread::sleep_for`
* `boost::posix_time::microseconds` -> `std::chrono::microseconds`
* `boost::get_system_time` -> `std::chrono::system_clock::now`

I also got rid of `d_epoch`, since `std::chrono::system_clock` uses the Unix epoch in practice. (This is guaranteed to be the case beginning in C++20.)

* Add System Time
* PDU To Stream
* Tags To PDU
* Time Delta

## Testing Done
I have not yet done any testing. I'd appreciate help from someone familiar with gr-pdu.

## Checklist

### DIFF
--- a/gr-pdu/lib/tags_to_pdu_impl.cc
+++ b/gr-pdu/lib/tags_to_pdu_impl.cc
@@ -15,6 +15,7 @@
 #include "tags_to_pdu_impl.h"
 #include <gnuradio/io_signature.h>
 #include <gnuradio/pdu.h>
+#include <chrono>
 
 namespace gr {
 namespace pdu {
@@ -156,8 +157,9 @@ void tags_to_pdu_impl<T>::publish_message()
     d_meta_dict = pmt::dict_add(
         d_meta_dict, metadata_keys::pdu_num(), pmt::from_uint64(d_burst_counter));
     if (d_wall_clock_time) {
-        double t_now((boost::get_system_time() - d_epoch).total_microseconds() /
-                     1000000.0);
+        double t_now(std::chrono::duration<double>(
+                         std::chrono::system_clock::now().time_since_epoch())
+                         .count());
         d_meta_dict = pmt::dict_add(
             d_meta_dict, metadata_keys::sys_time(), pmt::from_double(t_now));
     }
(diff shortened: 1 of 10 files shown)


mormj:
@jacobagilbert - any concerns about this one?

gr-pdu/lib/tags_to_pdu_impl.cc
```diff
@@ -156,8 +157,9 @@ void tags_to_pdu_impl<T>::publish_message()
     d_meta_dict = pmt::dict_add(
         d_meta_dict, metadata_keys::pdu_num(), pmt::from_uint64(d_burst_counter));
     if (d_wall_clock_time) {
-        double t_now((boost::get_system_time() - d_epoch).total_microseconds() /
-                     1000000.0);
+        double t_now(std::chrono::duration<double>(
+                         std::chrono::system_clock::now().time_since_epoch())
+                         .count());
```
mbr0wn, line 162: This assumes that the 'tick' value of the `duration` object is seconds. That makes sense and all, but is that guaranteed? I would have expected something like
```suggestion
                         std::chrono::duration_cast<std::chrono::seconds>(std::chrono::system_clock::now().time_since_epoch())
                         .count();
```

gr-pdu/lib/time_delta_impl.cc
```diff
@@ -84,7 +82,9 @@ void time_delta_impl::handle_pdu(const pmt::pmt_t& pdu)
 
     gr::thread::scoped_lock l(d_mutex);
 
-    double t_now((boost::get_system_time() - d_epoch).total_microseconds() / 1000000.0);
+    double t_now(
```
mbr0wn, line 85: Same comment as above. Also, I wonder if seconds are fine enough resolution here (unrelated to your PR) and if we should have a separate time-helper header with code like this (getting the number of secs since epoch seems useful elsewhere, too)

argilo: > I wonder if seconds are fine enough resolution here

The time is represented as a `double` so the resolution should be on the order of a microsecond.

gr-pdu/lib/tags_to_pdu_impl.cc
mbr0wn: OK I read that page but missed that detail. Thanks!

gr-pdu/lib/time_delta_impl.cc
mbr0wn: OK, I thought that `count()` would literally return the number of tick counts, and always return an integer (and then that gets cast to double). But that's not the case here (also it would break previous functionality... so.... yeah).

mbr0wn (approved):

PR #5653, opened 2022-03-20
blocks: Remove usage of Boost

argilo:
## Description
This pull request removes all usage of Boost from gr-blocks except for Boost.Test in test cases.

In a few cases, `boost::mutex` was used in headers but the implementation used it in `gr::thread::scoped_lock`, so I changed those to `gr::thread::mutex` (which can eventually be implemented as `std::mutex`).

In one case `boost::shared_mutex` was used, but shared locking was not actually used, so I replaced it with `gr::thread::mutex`.

* Ctrlport Probe
* File Sink
* File Source
* Message Strobe
* Message Strobe Random-Delay
* Probe Rate
* Wav File Sink

## Testing Done
I tested the Message Strobe, Message Strobe Random-Delay, and Probe Rate blocks with the following flow graph:
Screenshot from 2022-03-20 15-14-11
All work as before.

## Checklist

### DIFF
--- a/gr-blocks/include/gnuradio/blocks/file_sink_base.h
+++ b/gr-blocks/include/gnuradio/blocks/file_sink_base.h
@@ -13,7 +13,7 @@
 
 #include <gnuradio/blocks/api.h>
 #include <gnuradio/logger.h>
-#include <boost/thread.hpp>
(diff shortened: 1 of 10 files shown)


mormj (approved):
LGTM

PR #5669, opened 2022-03-21
blocks: Add exponential distribution to Message Strobe Random-Delay

argilo:
## Description
The Message Strobe Random-Delay block currently has three options for the distribution of the delays between messages:
* Poisson (default in GRC)
* Gaussian
* Uniform

Poisson is a strange choice, that distribution measures the number of events that occur during a fixed time interval in a Poisson process, not the time delays between them (which follow an Exponential distribution). I think the Exponential distribution would be much more useful as a default choice, and I suspect it's what was intended when this block was created.

To preserve backwards compatibility, I left the Poisson distribution in place and added Exponential as a fourth choice, which will be the default for a newly instantiated block.

I also updated the block's YAML file so that the "standard deviation" parameter is hidden for distributions that don't have a deviation parameter. The label is also dynamically set to indicate that the parameter has different meanings depending on the distribution. This makes the "peculiarities" section of the block documentation unnecessary. I also updated the C++ documentation to make the function of the arguments clearer.

* Message Strobe Random-Delay

## Testing Done
I tested all the distributions, and they appear to work correctly.

## Checklist

### DIFF
--- a/gr-blocks/grc/blocks_message_strobe_random.block.yml
+++ b/gr-blocks/grc/blocks_message_strobe_random.block.yml
@@ -10,16 +10,17 @@ parameters:
 -   id: dist
     label: Distribution
     dtype: enum
-    options: [blocks.STROBE_POISSON, blocks.STROBE_UNIFORM, blocks.STROBE_GAUSSIAN]
-    option_labels: [Poisson, Uniform, Gaussian]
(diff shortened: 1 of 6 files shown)


argilo:
A couple other deficiencies I noticed while working on this, which I'll leave to follow-up PRs:

* This block has no tests.
* Delays values are cast from `double` to `long`, losing precision. This would be especially problematic when the "mean" parameter is set to a small value.

mormj (approved):

PR #5670, opened 2022-03-21
Trickle throttle

marcusmueller:
# Pull Request Details

## Description
This should reduce the amount of chunkiness of signal  processing if used.

Users can now use throttle and get regular, because smaller, item chunks even if (sample rate) << buffer size.

This works now, without the user waiting half an hour for the first message debug output:

throttle_rate_test

To avoid surprise, this is an optional argument defaulting to previous behaviour (i.e., API and ABI compatible and backportable).

gr::blocks::throttle

## Testing Done

Added a test with an infinite sample source, a throttle at a rate of 50, and a max chunk size of 10, stopping the flowgraph after 0.5 s; without the chunkification, this should lead to an output of buffer_size / 2 items after waiting for buffer_size/2/50 (s, so typically 4096 samples  after 40s or so), with, this should lead to 20 or 30 items.

## Checklist

  - [x] Block API docs
  - [ ] Wiki

### DIFF
--- a/gr-blocks/grc/blocks_trickling_throttle.block.yml
+++ b/gr-blocks/grc/blocks_trickling_throttle.block.yml
@@ -1,0 +1,93 @@
+id: blocks_trickling_throttle
+label: Throttle (limited blocking)
+flags: [ throttle, python, cpp ]
+
+parameters:
+-   id: type
+    label: Type
+    dtype: enum
+    options: [complex, float, int, short, byte]
+    option_attributes:
+        size: [gr.sizeof_gr_complex, gr.sizeof_float, gr.sizeof_int, gr.sizeof_short,
+            gr.sizeof_char]
+    hide: part
+-   id: samples_per_second
+    label: Sample Rate
+    dtype: real
+    default: samp_rate
+-   id: vlen
+    label: Vector Length
+    dtype: int
+    default: '1'
+    hide: ${ 'part' if vlen == 1 else 'none' }
+-   id: ignoretag
+    label: Ignore rx_rate tag
+    dtype: bool
+    default: 'True'
+    hide: ${ ('part' if str(ignoretag) == 'True' else 'none') }
+-   id: duration_or_samples
+    label: Limited Quantity
+    dtype: enum
+    options: ['time', 'items', 'neither']
+    option_labels: ["Max. Time", "Max. Items", "No Limit"]
+-   id: maximum
+    label: 'Max. wait[s] or items'
+    dtype: real
+    hide: ${ ('part' if duration_or_samples == 'neither' else 'none') }
+    default: 0.1
+
+inputs:
+-   domain: stream
+    dtype: ${ type }
+    vlen: ${ vlen }
+
+outputs:
+-   domain: stream
+    dtype: ${ type }
+    vlen: ${ vlen }
+    optional: true
+
+asserts:
+- ${ vlen > 0 }
+- ${ samples_per_second > 0 }
+- ${ duration_or_samples == 'neither' or float(maximum) > 0 }
+
+templates:
+    imports: from gnuradio import blocks
+    make: 'blocks.throttle(
+               ${type.size}*${vlen},
+               ${samples_per_second},
+               ${ignoretag},
+               0 if "${duration_or_samples}" == "neither" 
+                   else max(
+                       int(float(${maximum}) * ${samples_per_second})
+                       if "${duration_or_samples}" == "time"
+                       else int(${maximum}),
+                       1)
+                )'
+    callbacks:
+    - set_sample_rate(${samples_per_second})
+
+cpp_templates:
+    includes: ['#include <gnuradio/blocks/throttle.h>', '#include <string_view>', '#include <algorithm>']
+    declarations: |-
+      constexpr unsigned int ${id}_limit(float value) {
+        const std::string_view limited("${duration_or_samples}");
+        if(limited == "time") {
+          return std::max(static_cast<unsigned int>(value * ${samples_per_second}), 1U);
+        } else if (limited == "items") {
+          return std::max(static_cast<unsigned int>(value), 1U);
+        }
+        return 0;
+      }
+      blocks::throttle::sptr ${id};
+
+    make: 'this->${id} = blocks::throttle::make(${type.size} * ${vlen}, ${samples_per_second}, ${ignoretag}, ${id}_limit(${maximum}));'
+    callbacks:
+    - 'set_sample_rate(${samples_per_second})'
+    - 'set_maximum_items_per_chunk(${id}_limit(${maximum}))'
+    translations:
+        'True': 'true'
+        'False': 'false'
+
+file_format: 1
--- a/gr-blocks/lib/throttle_impl.cc
+++ b/gr-blocks/lib/throttle_impl.cc
@@ -1,6 +1,7 @@
 /* -*- c++ -*- */
 /*
  * Copyright 2005-2011 Free Software Foundation, Inc.
+ * Copyright 2022 Marcus Müller
  *
  * This file is part of GNU Radio
  *
@@ -8,12 +9,15 @@
  *
  */
 
+#include "gnuradio/thread/thread.h"
+#include <chrono>
 #ifdef HAVE_CONFIG_H
 #include "config.h"
 #endif
 
 #include "throttle_impl.h"
 #include <gnuradio/io_signature.h>
+#include <algorithm>
 #include <cstring>
 #include <limits>
 #include <thread>
@@ -23,18 +27,25 @@ pmt::pmt_t throttle_rx_rate_pmt(pmt::intern("rx_rate"));
 namespace gr {
 namespace blocks {
 
-throttle::sptr throttle::make(size_t itemsize, double samples_per_sec, bool ignore_tags)
+throttle::sptr throttle::make(size_t itemsize,
+                              double samples_per_sec,
+                              bool ignore_tags,
+                              unsigned int maximum_items_per_chunk)
 {
     return gnuradio::make_block_sptr<throttle_impl>(
-        itemsize, samples_per_sec, ignore_tags);
+        itemsize, samples_per_sec, ignore_tags, maximum_items_per_chunk);
 }
 
-throttle_impl::throttle_impl(size_t itemsize, double samples_per_second, bool ignore_tags)
+throttle_impl::throttle_impl(size_t itemsize,
+                             double samples_per_second,
+                             bool ignore_tags,
+                             unsigned int maximum_items_per_chunk)
     : sync_block("throttle",
                  io_signature::make(1, 1, itemsize),
                  io_signature::make(0, 1, itemsize)),
       d_itemsize(itemsize),
-      d_ignore_tags(ignore_tags)
+      d_ignore_tags(ignore_tags),
+      d_maximum_items_per_chunk(maximum_items_per_chunk)
 {
     set_sample_rate(samples_per_second);
 }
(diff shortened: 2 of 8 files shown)


marcusmueller:
@willcode this *should* be backportable, question is whether you want!

marcusmueller:
Lol, [failing window build](https://github.com/marcusmueller/gnuradio/runs/5630771867?check_suite_focus=true#step:2:780):

```text
  Error: error: cannot lock ref 'refs/remotes/origin/file_meta_sink_py3_vector<uint8_t>': Unable to create 'D:/a/gnuradio/gnuradio/.git/refs/remotes/origin/file_meta_sink_py3_vector<uint8_t>.lock': Invalid argument
```

Windows file system restrictions strike once more! (fixed by deleting the uninvolved `file_meta_sink_py3_vector<uint8_t>` branch)

willcode:
It would be backportable if the `make` in GRC used keywords. Otherwise, older versions of 3.10 would not be able to use newer GRC files.

Is there a good, shorter name for `maximum_items_per_chunk`?

This helps latency through Throttle, but does it do anything for latency in a typical flowgraph? Won't other blocks end up buffering, especially any before the Throttle?

marcusmueller:
> ~It would be backportable if the `make` in GRC used keywords. Otherwise, older versions of 3.10 would not be able to use newer GRC files.~ _no, that wouldn't work either_

Ah, I was hoping GRC would just ignore fields it doesn't know. If that's not the case, we really shouldn't backport it (the only way of doing so would be adding a new block, but that we'd have to then forward-port to main, lest we want 3.10 flowgraphs to not be forward-compatible, right?).

> Is there a good, shorter name for `maximum_items_per_chunk`?

`max_items_per_chunk` would be an option, but I'm not sure it's enough of a shortening? Do you have something in mind?
 
> This helps latency through Throttle, but does it do anything for latency in a typical flowgraph? Won't other blocks end up buffering, especially any before the Throttle?

Well,  anything before throttle gets notified of any finished call to its `work`, so it would then most probably process a likely-sized 
chunk (because that's how much free output space it just got). Of course, that doesn't reduce the size of buffers at all.

The aim here is to enable applications with e.g. 2 kHz rate displayed on a 512-point Qt frequency sink or similar, which would previously have been multiple-seconds-stop-motion things missing most of data, where good visualization was impossible, because the instrumentation sink would get multiple screens worth of data at once, only display the last of these, then wait for a couple of secondsfor the next heap of items, and so on.

mormj:
@marcusmueller - This feature should be very useful, so thank you for adding it.  Since throttle is a very common block, I'm wondering if the top level param should be made more user friendly?  Perhaps rather than specifying the max chunk of items, specifying the max period to produce samples.  

In your example, if you have the throttled output connected to a QT sink, you want to produce at a rate similar to the QT block, set the max period to something similar or smaller

mormj:
@marcusmueller - This feature should be very useful, so thank you for adding it.  Since throttle is a very common block, I'm wondering if the top level param should be made more user friendly?  Perhaps rather than specifying the max chunk of items, specifying the max period to produce samples.  

In your example, if you have the throttled output connected to a QT sink, you want to produce at a rate similar to the QT block, set the max period to something similar or smaller

marcusmueller:
Good point! I'd try to stay "integer counter" internally/for the ctor, but offer alternative `make` signatures, one with the maximum number of items, but another convenience one that converts time & rate to number of samples and expose both to Python.

That way, it remains useful for testing things (we can use this to modulate the input chunk sizes to other block's `work`s!) but also gains usefulness for the practical time-based user.

Regarding the GRC backwards/forwards compatibility problem: 

Maybe I'll do the following:

1. restore the blocks_throttle.block.yml functionally to its original state, so that we don't break existing flow graphs and can still backport
2. add one or two separate "Throttle (Chunk-Limited)"/"Throttle (Latency-Limited)" block definitions that use the `make`s from above, make both available on a maint-3.10-targetting PR
3. on main, change the visible name of the blocks_throttle.block.yml to "Throttle (Unlimited Latency)" or such.

Would that sound reasonable?

mormj:
> Would that sound reasonable?

This sounds very reasonable - sorry for the delay in me getting back to this

dkozel:
Having multiple blocks (in GRC) sounds confusing and unnecessary. I can't quite find the exact naming, but what about adding an optional parameter for "Minimum Frames per Second" with a default value of 1 (existing behavior)? Having the default constructor be a passthrough to the new one with max_items_per_chunk=samp_rate? (If I've understood existing behavior correctly).

777arc:
RE: the "so that we don't break existing flow graphs and can still backport", even if we don't backport, whether this is added to 3.10.3 or 3.11.0 it would be nice to not break all existing flowgraphs that use throttles, which is quite a bit, or am I misunderstanding what would break.

mormj:
My biggest hesitancy at this point is the visible units of the new parameter.  As a user facing parameter, it seems non-intuitive.  It seems like units in seconds would make more sense - since you are already giving the block "samples per second" - specifying the maximum chunk or incurred/excess latency in seconds would make more sense, and not have to back out sample calculations from the sample rate.  If that is a separate constructor, would be fine -- but an integer counter for such a common block seems like a big step for many users.

In the example you give above:
> Added a test with an infinite sample source, a throttle at a rate of 50, and a max chunk size of 10, stopping the flowgraph after 0.5 s; without the chunkification, this should lead to an output of buffer_size / 2 items after waiting for buffer_size/2/50 (s, so typically 4096 samples after 40s or so), with, this should lead to 20 or 30 items.

You could set the maximum chunk (or whatever it should be called) to 0.2 seconds - ensuring that samples are produced (roughly) every 0.2 seconds.

marcusmueller:
@mormj I came at it from the Qt Time Sink side, where you do specify the number of samples. But let me quickly remedy that by simply introducing the steps outlined above.

marcusmueller:
@dkozel it is annoying, but I can't have a single backwards-compatible block; GRC doesn't support ignoring unknown parameters from a GRC file. So, I'll have at least two blocks (classic and trickling), might as well have three.

mormj:
@marcusmueller - thank you for the updates.  This makes more sense to me now:

image

Probably also needs to be added to block.tree.yaml to put it in grc in the right place

Tagged @dkozel as a reviewer

marcusmueller:
This still needs a test case and wiki page before of can be merged. C++ gen is completely untested!

willcode:
A bunch of extra build/formatter stuff got checked in with the last push.

willcode:
Can the id in the yml can go back to being `throttle` instead of `blocks_trickling_throttle`.

It looks like the situations where this will help are
- Simulations, where output widgets/printouts are otherwise very delayed
- Burst-based tx where sob/eob are used

The wiki could use some guidance on how to use, and where in the flowgraph to put, the throttle to get the desired effect. Guessing that it often goes as early as possible so it acts as the choke point?

marcusmueller:
> The wiki could use some guidance on how to use, and where in the flowgraph to put, the throttle to get the desired effect. Guessing that it often goes as early as possible so it acts as the choke point?

Exactly; with the "old" throttle behaviour it doesn't matter as much – it'll tend to generate full output buffers with maximally long waits in between, if it's actually the main bottleneck of the flow graph – but with these changes, it'll make sense to put it as closely before the thing where you benefit most from the "small chunk" behaviour. In other words: if you want to look at the signal in a Qt Time Sink, put it right before that. If you want to consume data from a source block at a relatively constant rate, e.g. to simulate the workload when connected to a hardware receiver, put it right after the source.

gr-blocks/grc/blocks_trickling_throttle.block.yml
```diff
@@ -0,0 +1,81 @@
...
+    label: Ignore rx_rate tag
+    dtype: bool
+    default: 'True'
+    hide: ${ ('part' if str(ignoretag) == 'True' else 'none') }
+-   id: duration_or_samples
+    label: Limited Quantity
+    dtype: enum
+    options: [True, False]
+    option_labels: ["wait duration", "#items per chunk"]
+-   id: maximum_items_per_chunk
```
willcode, line 33: Can the two max's be collapsed into one? Hidden value are a pain (e.g., finding a hidden, invalid value in the asserts below). They're different types, so not sure what to do with that.

willcode (commented):
Is it worth finding a way to warn the user when the block will not do anything useful, or will end up being a bottleneck? There's a limited range of "max" that makes sense. Too big and it's not doing anything anyway, too small and it could kill performance somewhere else in the flowgraph.

What happens if the max items based on time rounds to 0?

marcusmueller:
@willcode 
Looks like this now:

canvas

dialogitems
dialogtime

marcusmueller:
I also added the `deprecated` flag to the "old" throttle – since we have no mechanism to automatically upgrade to the new one, and the new YAML will cause old flow graphs to not work, this seems pretty much the only way forward withut extensive modifications in GRC.

willcode:
Deprecation (indefinite) sounds right. Would we want an Auto setting too, in that case, so users don't need to think about a magic max value for that?

willcode:
On second thought, v3.10 will be around for a while, and having an orange/deprecated throttle block that everyone is still using doesn't seem right. No better idea yet. Let's make any deprecation a separate thing later.

marcusmueller:
Yeah, that's why I made the deprecation a separate commit – no need to deprecate it now, so if backporting, simply omit that commit.

marcusmueller:
Re: auto setting: that's what happens when you leave it at 0 – no limited block duration. Or do you mean it should default to something like 0.1 s?

willcode:
I was thinking that Auto (or Default) would be a separate mode in the GRC yml, and that would set the value to 0 internally. It seems a little non-intuitive that 0.00001 sec means low latency and 0.00000 sec means max latency. 0 might even be an error.

marcusmueller:
ah, so "auto" would mean "unlimited"?

willcode:
Yes. I suppose "No Limit" is a more accurate name.

marcusmueller:
made the suggested additions, made C++ generation work!

willcode (approved):
Seem like a good time to get this some exposure on `main`!

mormj (approved):

willcode:
@marcusmueller I've been trying this out a bit, and have some suggested UI changes:
- Default should be No Limit, since that's the old behavior.
- The limit field should be hidden when No Limit is selected (all vs part).
- The block name takes up a lot of real estate in GRC. We could call the new one "Throttle" and the old one "Throttle (old)" or something like that.
- The id of the block could be changed to "throttle2". The "trickling_throttle" name was good for discussion but makes for an odd variable name in generated code.
- C++ template has a callback for `maximum_items_per_chunk`, but the Python template doesn't have a callback - was that on purpose?
- I was thinking that ms would be easier than sec for the time limit, but I wasn't sure how to make that work in the C++ template.
- Default value of 1 looks better than 0.1 - there's no right answer anyway and 0.1 doesn't make any sense for num items.

Suggestions so far:
```
diff --git a/gr-blocks/grc/blocks.tree.yml b/gr-blocks/grc/blocks.tree.yml
index 81c664dd8..05ac8e32e 100644
--- a/gr-blocks/grc/blocks.tree.yml
+++ b/gr-blocks/grc/blocks.tree.yml
@@ -90,7 +90,7 @@
   - blocks_msg_meta_to_pair
 - Misc:
   - blocks_throttle
-  - blocks_trickling_throttle
+  - blocks_throttle2
   - blocks_vector_source_x
   - blocks_head
   - blocks_skiphead
diff --git a/gr-blocks/grc/blocks_throttle.block.yml b/gr-blocks/grc/blocks_throttle.block.yml
index fef9db7ec..81ffb8ebb 100644
--- a/gr-blocks/grc/blocks_throttle.block.yml
+++ b/gr-blocks/grc/blocks_throttle.block.yml
@@ -1,5 +1,5 @@
 id: blocks_throttle
-label: Throttle (unlimited blocking)
+label: Throttle (old)
 flags: [ throttle, python, cpp, deprecated ]
 
 parameters:
diff --git a/gr-blocks/grc/blocks_throttle2.block.yml b/gr-blocks/grc/blocks_throttle2.block.yml
index e6b348d6a..9aa17678a 100644
--- a/gr-blocks/grc/blocks_throttle2.block.yml
+++ b/gr-blocks/grc/blocks_throttle2.block.yml
@@ -1,5 +1,5 @@
-id: blocks_trickling_throttle
-label: Throttle (limited blocking)
+id: blocks_throttle2
+label: Throttle
 flags: [ throttle, python, cpp ]
 
 parameters:
@@ -25,16 +25,16 @@ parameters:
     dtype: bool
     default: 'True'
     hide: ${ ('part' if str(ignoretag) == 'True' else 'none') }
--   id: duration_or_samples
-    label: Limited Quantity
+-   id: limit
+    label: Limit
     dtype: enum
-    options: ['time', 'items', 'neither']
-    option_labels: ["Max. Time", "Max. Items", "No Limit"]
+    options: ['auto', 'time', 'items']
+    option_labels: ["None", "Max Time (sec)", "Max Items"]
 -   id: maximum
-    label: 'Max. wait[s] or items'
+    label: 'Maximum'
     dtype: real
-    hide: ${ ('part' if duration_or_samples == 'neither' else 'none') }
-    default: 0.1
+    hide: ${ ('all' if limit == 'auto' else 'none') }
+    default: 1
 
 inputs:
 -   domain: stream
@@ -50,7 +50,7 @@ outputs:
 asserts:
 - ${ vlen > 0 }
 - ${ samples_per_second > 0 }
-- ${ duration_or_samples == 'neither' or float(maximum) > 0 }
+- ${ limit == 'auto' or float(maximum) > 0 }
 
 templates:
     imports: from gnuradio import blocks
@@ -58,10 +58,10 @@ templates:
                ${type.size}*${vlen},
                ${samples_per_second},
                ${ignoretag},
-               0 if "${duration_or_samples}" == "neither" 
+               0 if "${limit}" == "auto" 
                    else max(
                        int(float(${maximum}) * ${samples_per_second})
-                       if "${duration_or_samples}" == "time"
+                       if "${limit}" == "time"
                        else int(${maximum}),
                        1)
                 )'
@@ -72,7 +72,7 @@ cpp_templates:
     includes: ['#include <gnuradio/blocks/throttle.h>', '#include <string_view>', '#include <algorithm>']
     declarations: |-
       constexpr unsigned int ${id}_limit(float value) {
-        const std::string_view limited("${duration_or_samples}");
+        const std::string_view limited("${limit}");
         if(limited == "time") {
           return std::max(static_cast<unsigned int>(value * ${samples_per_second}), 1U);
         } else if (limited == "items") {
```

gr-blocks/lib/throttle_impl.cc
```diff
@@ -44,30 +55,44 @@ throttle_impl::~throttle_impl() {}
 bool throttle_impl::start()
 {
     d_start = std::chrono::steady_clock::now();
-    d_total_samples = 0;
```
argilo, line 47: @marcusmueller The removal of this line has broken pause/resume when playing back I/Q files in Gqrx: https://github.com/gqrx-sdr/gqrx/issues/1234

I believe it should have been replaced with `d_total_items = 0`.

PR #5694, opened 2022-03-26
digital: add generic CRC blocks (backport to maint-3.9)

willcode:
This adds a crc class that implements a generic CRC calculation using
a table-driven byte-by-byte approach. The CRC parameters are set when
instancing the class.

The crc class is used in the CRC Append and CRC Check blocks, which
take a PDU as input, and either compute the CRC of that PDU and append
it, or check the CRC included in the PDU (sending out the PDU in one
of two outputs according as to whether the CRC was correct).

The CRC blocks can optionally ignore a certain amount of "header" bytes
in the CRC calculation (used for packets where the CRC only affects the
"payload" and not the header). Additionally, the CRC Check block can
optionally remove the CRC after checking it.

Signed-off-by: Daniel Estévez <daniel@destevez.net>
(cherry picked from commit 1c4ef3dac7a12d4841b9544cc4c6cef0aca4cff0)

Backport https://github.com/gnuradio/gnuradio/pull/5619


daniestevez:
I've taken a detailed look at why this fails. In 3.9, there is a CRC Python module in gr-digital: https://github.com/gnuradio/gnuradio/blob/maint-3.9/gr-digital/python/digital/crc.py

This was removed in 3.10: https://github.com/gnuradio/gnuradio/commit/5b82f925f4af11ba99d6c7e61dbef3fc3c0d33b3

I don't like the idea of touching the existing `digital.crc` module of 3.9, because in theory some 3rd party could be using it. The problem is that this module shadows the Python binding for the `crc` C++ class. But for the purpose of the QA test we can import the Python binding directly from `digital_python`. Just adding
```
from gnuradio.digital.digital_python import crc
```
at the beginning of `qa_crc.py` and then using `crc` instead of `digital.crc` works.

It still makes it a bit awkward for a 3rd party to use the `crc` C++ class from Python (because they will hit this problem again), but I can't think of a much better solution for 3.9.

Do you think we can use this solution here?

willcode:
This isn't going to be in 3.9.6.0, so we've got some time to think about it. I don't know enough about what this might impact, or what non-obviousness we'd be creating.

willcode:
Skipping this in 3.9, for now at least. We can revisit if there's demand.

PR #5698, opened 2022-03-28
GRC: add jsonschema check for JSON/YAML blocks module

### LINKED ISSUE #5696: GRC: json validation blocks need conditional install
### What happened?

https://github.com/gnuradio/gnuradio/pull/5611#issuecomment-1079499969

python3-jsonschema is not a checked requirement, but the json block requires it

### System Information

Any

### GNU Radio Version

3.11-git (main)

### Specific Version

0.0.0.0

Install the json block on a system without python3-jsonschema

### Relevant log output

_No response_

marcusmueller:
# Pull Request Details

## Description

Currently, python3-jsonschema is a "hidden" dependency, and without it, the JSON and YAML config blocks simply don't work.

This PR registers a new component (just for these two blocks), which depends on GRC and the presence of jsonschema. That way, there's feedback to whoever runs CMake that a dependency is missing.

## Related Issue

Fixes #5696 

* JSON Config
* YAML Config

## Testing Done

No tests, presently. We should probably have added such!

## Checklist

  - Interesting point is that this is a dependency that we need to document in our build guides
  - wish I had, but no time


marcusmueller:
CI (at least the doxygen build) correctly disables these blocks: https://github.com/gnuradio/gnuradio/runs/5719601895?check_suite_focus=true#step:4:261


```
-- Configuring JSON/YAML config blocks support...
--   Dependency ENABLE_GRC = ON
--   Dependency JSONSCHEMA_FOUND = FALSE
--   Disabling JSON/YAML config blocks support.
--   Override with -DENABLE_JSONYAML_BLOCKS=ON/OFF
```

marcusmueller:
ah, since I'm touching conda here: pinging @ryanvolz as well

.circleci/config.yml
```diff
@@ -0,0 +1,24 @@
+# This file was generated automatically from conda-smithy. To update this configuration,
```
ryanvolz, line 1: I've been manually deleting this file to reduce clutter since it only has a purpose of the repository is connected to CircleCI.

.github/workflows/automerge.yml
```diff
@@ -0,0 +1,19 @@
+on:
```
ryanvolz, line 1: Same here, manually delete since we have no use for automerge.

.github/workflows/webservices.yml
```diff
@@ -0,0 +1,13 @@
+on: repository_dispatch
```
ryanvolz, line 1: Last one to delete manually.

.packaging/conda_recipe/meta.yaml
```diff
@@ -28,6 +28,7 @@ requirements:
     - ninja
     - pkg-config  # [not win]
     - sysroot_linux-64 2.17  # [linux64]
+    - jsonschema
```
ryanvolz, line 31: This is not needed as a "build" dependency, which is for things that must run natively in a cross-compilation scenario and so includes mainly build tools. Having it in "host" is enough to be present for detection by CMake at build time, and "run" makes it a runtime dependency.

ryanvolz (changes requested):
The changes I'm suggesting shouldn't change the success/failure of the build, but they remove a little excess.

.circleci/config.yml
marcusmueller: We should `git rm` and  .gitignore this file, then?

.packaging/conda_recipe/meta.yaml
marcusmueller: cool! Will do. Question: what's the meaning of "build" then?

.circleci/config.yml
ryanvolz: Yes, although last I checked the .gitignore will have no effect on conda-smithy adding the files as a staged change, unfortunately. I couldn't figure out a way to automatically keep those files out with the current conda-smithy tooling.

marcusmueller: would we be better of if we had a script that ran conda-smithy, then did the commit rather than using `conda-smithy -c auto`?

marcusmueller: (note that we also always have to manually sign these commits, as well)

.github/workflows/webservices.yml
marcusmueller: What about .ci_support/README?

willcode (approved):

willcode:
We'll have to see whether people run into a problem with this as an optional dependency. We'll end up with flowgraphs that use what seems to be a basic, built-in capability, that won't run on some systems. Packagers should make this a hard dep even if we can build without.

PR #5700, opened 2022-03-28
digital: Replace boost::crc with built-in CRC implementation

argilo:
## Description
In #5619 a generic CRC class was added. Existing blocks that use Boost.CRC can switch over to this implementation, thereby eliminating a Boost dependency.

## Related Issue
* #5619

* Async CRC16
* Async CRC32
* Stream CRC32
* Packet Header Generator
* Packet Header Parser

## Testing Done
There are five places where the CRC implementation was replaced:
1. `crc32_bb`: Here I verified that `qa_crc32_bb.py` fails if I corrupt the CRC parameters.
2. `crc16_async_bb`: This block had no tests. I added a basic test, and verified that it passes before & after my changes.
3. `crc32_async_bb`: This block had no tests. I added a basic test, and verified that it passes before & after my changes.
4. `packet_header_default`: Here I verified that `qa_packet_headerparser_b.py` and `qa_pacrket_headergenerater_bb.py` fail if I corrupt the CRC paremeters.
5. `header_format_crc`: This block does not appear to have tests. Unfortunately its Python binding appears broken, so I didn't attemp to create one. However, this class uses the same CRC parameters as `packet_header_default`, so it is unlikely to be broken.

## Checklist

### DIFF
--- a/gr-digital/lib/header_format_crc.cc
+++ b/gr-digital/lib/header_format_crc.cc
@@ -27,7 +27,9 @@ header_format_crc::sptr header_format_crc::make(const std::string& len_key_name,
 
 header_format_crc::header_format_crc(const std::string& len_key_name,
                                      const std::string& num_key_name)
-    : header_format_base(), d_header_number(0)
+    : header_format_base(),
+      d_header_number(0),
+      d_crc_impl(8, 0x07, 0xFF, 0x00, false, false)
 {
     d_len_key_name = pmt::intern(len_key_name);
     d_num_key_name = pmt::intern(num_key_name);
@@ -47,10 +49,11 @@ bool header_format_crc::format(int nbytes_in,
     // for 12-bit representation?
     nbytes_in &= 0x0FFF;
 
-    d_crc_impl.reset();
-    d_crc_impl.process_bytes((void const*)&nbytes_in, 2);
-    d_crc_impl.process_bytes((void const*)&d_header_number, 2);
-    uint8_t crc = d_crc_impl();
+    unsigned char buffer[] = { (unsigned char)(nbytes_in & 0xFF),
+                               (unsigned char)(nbytes_in >> 8),
+                               (unsigned char)(d_header_number & 0xFF),
+                               (unsigned char)(d_header_number >> 8) };
+    uint8_t crc = d_crc_impl.compute(buffer, sizeof(buffer));
 
     // Form 2 12-bit items into 1 2-byte item
     uint32_t concat = 0;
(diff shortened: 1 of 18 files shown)


gr-digital/lib/header_format_crc.cc
```diff
@@ -47,10 +49,11 @@ bool header_format_crc::format(int nbytes_in,
     // for 12-bit representation?
     nbytes_in &= 0x0FFF;
 
-    d_crc_impl.reset();
-    d_crc_impl.process_bytes((void const*)&nbytes_in, 2);
-    d_crc_impl.process_bytes((void const*)&d_header_number, 2);
-    uint8_t crc = d_crc_impl();
+    unsigned char buffer[] = { (unsigned char)(nbytes_in & 0xFF),
```
daniestevez, line 52: I think this is only doing the same thing as the previous code if the platform is little-endian. But I like this version better, as it does the same thing regardless of the endianness.

But then I wonder if the order in which the bytes of the `nbytes_in` and `d_header_number` fields end up in the output depends on the machine endianness. Looking at the code I guess it does depend. So we might end-up having something weird where the CRC isn't calculated over the bytes in the order they end up in the header but in some other order.

Perhaps it's good to take a closer look to this code and try to get a behaviour that is consistent regardless of the endianness and the CRC calculated using the same order as the data in the header?

daniestevez (commented):
Thanks for this. It's a good idea to get rid of boost's CRC.

The code looks good to me, except for my comment/question about the endianness in the packet_header block, which we may or may not want to handle more carefully.

gr-digital/lib/header_format_crc.cc
argilo: > I think this is only doing the same thing as the previous code if the platform is little-endian. But I like this version better, as it does the same thing regardless of the endianness.

I suspect the old code didn't work at all for big-endian platforms, because `nbytes_in` is an `int`, which is usually 4 bytes and therefore would have zeroes as its first two bytes.

argilo: >But then I wonder if the order in which the bytes of the nbytes_in and d_header_number fields end up in the output depends on the machine endianness. Looking at the code I guess it does depend.

Does it? It looks to me like `add_field32` includes endianness conversion (via `volk_32u_byteswap`):

https://github.com/gnuradio/gnuradio/blob/6767475c9312520395cb888997d3348de32b1054/gr-digital/lib/header_buffer.cc#L72-L84

daniestevez: I hadn't look at `add_field32` in detail, but now that I do, I note two things:
* `header_format_crc::format` always calls `add_field32` with `bs = true`, so the `byteswap` is never done.
* I don't really understand the intention of `x = x >> (32 - len);` after the byteswap. I think that if you feed the same number in `data` and use opposite values for `bs` in a little-endian and big-endian system, if you don't do this shift `x = x >> (32 - len);`, then you get the same data appended to `d_buffer`. But if you do the shift, you get different data appended to the buffer (unless `len == 32`, of course).

mormj (approved):

mormj:
@daniestevez @argilo - is this one ready to merge?

mormj:
Thanks for following up here - I'll follow your recommendation to go ahead and merge and address any additional issues on a separate PR

daniestevez:
@argilo thanks for looking at this again. I agree that fixing the big-endian problems will be best done in a separate PR, since I think those problems already existed before replacing boost::crc with the built-in CRC.

PR #5702, opened 2022-03-29
Fix create hier / missing top_block error

bkerler:
# Pull Request Details
Fixes an error on creating hier, which states that "top_block" is missing, thus making the option "Create hier" unusable. Affects gnuradio >= 3.8.

## Description
On selecting several blocks, right click "More -> Create Hier", the expected behaviour should be to create a new hier consisting of the selected blocks and to add pads to the hier block. Instead there is an error stating that "top_block" cannot be found.

## Related Issue
No issue found.

Affects gnuradio-companion on using the menu "Create hier" as suggested in the wiki https://wiki.gnuradio.org/index.php?title=Hier_Blocks_and_Parameters

## Testing Done
The function was tested and seemed to work fine.

## Checklist

### DIFF
--- a/grc/gui/Application.py
+++ b/grc/gui/Application.py
@@ -311,7 +311,7 @@ class Application(Gtk.Application):
             flow_graph.move_selected(coords)
 
             # Set flow graph to heir block type
-            top_block = flow_graph.get_block("top_block")
+            top_block = flow_graph.get_block(Constants.DEFAULT_FLOW_GRAPH_ID)
             top_block.params['generate_options'].set_value('hb')
 
             # this needs to be a unique name


grc/gui/Application.py
```diff
@@ -311,7 +311,7 @@ def flow_graph_update(fg=flow_graph):
             flow_graph.move_selected(coords)
 
             # Set flow graph to heir block type
-            top_block = flow_graph.get_block("top_block")
+            top_block = flow_graph.get_block("default")
```
mormj, line 314: Thank you for tracking this one down.  The only thing I might suggest here is replacing this hardcoded name with Constants.DEFAULT_FLOW_GRAPH_ID

alphafox02: Thank you, applied the fix to 3.8 and now it’s possible to highlight blocks and right click to create a heir block.

bkerler: Adapted it accordingly

alphafox02:
Changing from ‘default’ to Constants.DEFAULT_FLOW_GRAPH does not work in GR 3.8, at least not 3.8.1, it complains there’s a keyerror and no block with that name. However the initial change to default did work.

bkerler:
It seems Constants.DEFAULT_FLOW_GRAPH_ID doesn't exist on GR 3.8, thus a backport to 3.8 wouldn't be possible. On >=3.10 however this is fine.

mormj (approved):

mormj:
Forced a rebase on here - will merge when CI completes

PR #5706, opened 2022-03-30
audio: Replace Boost with standard C++

argilo:
## Description
This pull requests removes all Boost from gr-audio, replacing it with standard C++:

* `boost::this_thread::sleep` -> `std::this_thread::sleep_for`
* `boost::posix_time::milliseconds` -> `std::chrono::milliseconds`
* `boost::lockfree::spsc_queue` -> `std::queue` + `std::mutex`

* ALSA Audio Sink
* Windows Audio Source

## Testing Done
Unfortunately, I don't know how to build GNU Radio on Windows, so I haven't been able to test the Windows Audio Source. (The code compiles, at least. :smile:) I'd appreciate help with testing.

To test the ALSA Audio Source, I used the following flow graph:
Screenshot from 2022-03-29 22-07-40

## Checklist


argilo:
A couple tests I would suggest running on Windows:
* Audio Source (Sample Rate: 48k) -> Throttle (Sample Rate: 24k) -> Null Sink
  * Make sure that `buffer_queue` does not grow indefinitely.
* Audio Source (Sample Rate: 48k) -> Null Sink
  * Check whether CPU utilization is high. Right now the `work` function exits immediately if no samples are available, but I wonder whether it should instead use a condition variable to block and wait for samples to arrive through the callback function.

gr-audio/lib/alsa/alsa_sink.cc
```diff
@@ -81,7 +83,7 @@ alsa_sink::alsa_sink(int sampling_rate, const std::string device_name, bool ok_t
         error = snd_pcm_open(&t, d_device_name.c_str(), SND_PCM_STREAM_PLAYBACK, 0);
         d_pcm_handle.set(t);
         if (error < 0) {
-            boost::this_thread::sleep(boost::posix_time::milliseconds(10));
+            std::this_thread::sleep_for(std::chrono::milliseconds(10));
```
mbr0wn, line 86: All the cool kids do

```suggestion
            std::this_thread::sleep_for(10ms`);
```

:)

(Not a change request, just another C++ thing I kind of like)

argilo:
I was able to test this on Windows using the conda build. Unfortunately there's a deadlock. It seems that `waveInAddBuffer`, which is called by `work` while it holds the `buffer_queue_mutex`, can call the callback function, which also needs the mutex.

I'll put this pull request in draft until that is sorted out.

argilo:
Problem solved. The `work` function doesn't need to keep the `buffer_queue_mutex` locked once the item is popped from the `buffer_queue`, so we can unlock it at that point, and then lock it again before the loop resumes. After this change, the deadlock is gone and the Windows Audio Source appears to work correctly.

argilo:
I also ran the two tests I suggested above (which generate excessive back-pressure or none at all) and didn't observe any problems with high CPU or memory utilization.

That said, it appears there is a bug with the Windows Audio Source, since the audio sounds bad both before and after this change. I'll look into that, but a fix probably belongs in a separate pull request.

gr-audio/lib/windows/windows_source.cc
```diff
@@ -181,11 +183,14 @@ int windows_source::work(int noutput_items,
                 dw_items += buffer_length / 2;
             }
             buffer_queue.pop();
+            lock.unlock();
```
mormj, line 186: For safety, should this section be a scoped_lock?  What happens if there is an exception in those methods?

argilo: What safety issue did you have in mind?

My understanding is that if an exception occurs, then the `unique_lock` destructor well be called, thereby unlocking it.

argilo:
> That said, it appears there is a bug with the Windows Audio Source, since the audio sounds bad both before and after this change. I'll look into that, but a fix probably belongs in a separate pull request.

After some further testing, I noticed the same audio issue is present even when using the PortAudio source. So I suspect this is a problem with my particular audio device, and not the Windows source block.

willcode:
@mormj Do you think this is good for backport, based on the last couple comments here?

mormj:
Seems like there is still a non-zero amount of risk - and the benefit to a maint release (other than limiting code divergence) would not be seen.

PR #5719, opened 2022-03-30
runtime: Replace Boost mutexes with standard C++

argilo:
## Description
Here I've replaced all instances of `boost::mutex`, `boost::shared_mutex`, `boost::mutex::scoped_lock`, `boost::lock_guard`, `boost::unique_lock`, `boost::shared_lock`, and `boost::condition_variable` with standard C++.

* `rpcbufferedget`
* `rpcserver_thrift`
* `gr::thread::condition_variable`
* `gr::thread::mutex`
* `gr::thread::scoped_lock`
* `gr::thread::thread_group`
* PMT
* `top_block_impl`
* `tpb_thread_body`

## Testing Done
Not much yet. I've verified that Gqrx still runs correctly. Help with testing would be appreciated.

## Checklist
- [ ] ~~I have added tests to cover my changes~~, and all previous tests pass.

### DIFF
--- a/gnuradio-runtime/include/gnuradio/basic_block.h
+++ b/gnuradio-runtime/include/gnuradio/basic_block.h
@@ -18,7 +18,6 @@
 #include <gnuradio/runtime_types.h>
 #include <gnuradio/sptr_magic.h>
 #include <gnuradio/thread/thread.h>
-#include <boost/thread/condition_variable.hpp>
 #include <deque>
 #include <functional>
 #include <map>
--- a/gnuradio-runtime/include/gnuradio/rpcbufferedget.h
+++ b/gnuradio-runtime/include/gnuradio/rpcbufferedget.h
@@ -11,8 +11,8 @@
 #ifndef RPCBUFFEREDGET_H
 #define RPCBUFFEREDGET_H
 
-#include <boost/thread/condition_variable.hpp>
-#include <boost/thread/mutex.hpp>
+#include <condition_variable>
+#include <mutex>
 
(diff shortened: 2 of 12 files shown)


argilo:
It looks like the `blocks_qa_gr_top_block` and `blocks_qa_set_msg_handler` tests are timing out, which will need some investigation.

argilo:
I was able to reproduce the problem with `blocks_qa_set_msg_handler` locally. Sometimes there's a deadlock where a null source is waiting for output buffer space to become available, and the top block is waiting for the flow graph to stop.

https://github.com/gnuradio/gnuradio/blob/3349d483a1f7e5db0eea042815d237810373857b/gnuradio-runtime/lib/tpb_thread_body.cc#L141-L147

https://github.com/gnuradio/gnuradio/blob/3349d483a1f7e5db0eea042815d237810373857b/gr-blocks/lib/qa_set_msg_handler.cc#L29-L58

argilo:
I guess the problem is that stopping the flow graph ultimately calls `interrupt()` on all the threads, and `boost::condition_variable::wait()` is an interruption point (but presumably std::condition_variable::wait() is not).

https://www.boost.org/doc/libs/1_78_0/doc/html/thread/thread_management.html#interruption_points

This seems like it will make it difficult to switch away from Boost.

mbr0wn:
> 

Personally, I think interruption points are an awful concept. We did the switch (mostly) from Boost thread stuff to standard C++ in UHD, and Boost interruption points were a major pain point.

PR #5726, opened 2022-03-31
look for x-terminal-emulator (backport to maint-3.10)

maitbot:

Backport #5730

---
name: Pull Request for gnuradio 3.10
about: generating grc.conf
title: 'look for x-terminal-emulator'
labels: ''
assignees: ''

---

# Pull Request Details

## Description

When building gnuradio in a clean build environment, which is great for doing
repeatable builds, the CMake logic for generating the grc.conf file
left nothing for the xterm_executable.

This leads to a confusing error the first time a user runs gnuradio-companion.

This patch uses c-terminal-emulator by default. Which is a nice choice in Debian and Ubuntu.

gnuradio-companion

## Testing Done

This patch has long been in debian/patches
Recently not included, the result was quickly noticed.

I added foot to the search list. It does the right thing when running
gnuradio-companion under Wayland and the sway window manager,
where foot is the default Wayland terminal. So - not X, but works.

## Checklist

### DIFF
--- a/grc/CMakeLists.txt
+++ b/grc/CMakeLists.txt
@@ -102,7 +102,7 @@ endif(CMAKE_INSTALL_PREFIX STREQUAL "/usr")
 
 if(UNIX)
     find_program(GRC_XTERM_EXE
-        NAMES x-terminal-emulator gnome-terminal.wrapper konsole xterm foot
+        NAMES x-terminal-emulator gnome-terminal konsole xterm foot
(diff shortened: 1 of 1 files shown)


willcode:
Fedora has `gnome-terminal` but not `gnome-terminal.wrapper`. We should make sure that one works across platforms, since not everyone seems to install `xterm` these days.

PR #5735, opened 2022-04-01
Add parantheses

eltos:
Add parantheses to the templates, such that entering expressions like `f1+f2` into the parameter fields will produce the expected result.

Phase increment before: `2.0*math.pi*f1 + f2/sample_rate`  
Phase increment now: `2.0*math.pi * (f1+f2) / (sample_rate)`  

Related: https://github.com/gnuradio/gnuradio/pull/3046

### DIFF
--- a/gr-blocks/grc/blocks_freqshift_cc.block.yml
+++ b/gr-blocks/grc/blocks_freqshift_cc.block.yml
@@ -24,9 +24,9 @@ templates:
     imports: |-
         from gnuradio import blocks
         import math
-    make: blocks.rotator_cc(2.0*math.pi*${freq}/${sample_rate})
+    make: blocks.rotator_cc(2.0*math.pi*(${freq})/(${sample_rate}))
     callbacks:
-    - set_phase_inc(2.0*math.pi*${freq}/${sample_rate})
+    - set_phase_inc(2.0*math.pi*(${freq})/(${sample_rate}))
 
 documentation: |-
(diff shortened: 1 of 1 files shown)


willcode (commented):
Does the callback work correctly when f1 or f2 is changed in this case? It looks like it should, just checking. Great idea, since the user would have no idea when parens are required for an internal express. How many other places is something like this needed?

eltos:
> Does the callback work correctly when f1 or f2 is changed in this case? It looks like it should, just checking. Great idea, since the user would have no idea when parens are required for an internal express.

Well that user was me yesterday, took me a while to figure out what was going on. 
After adding the parantheses it obviously worked.

eltos:
> How many other places is something like this needed?

Good question, at least one more:
- [ ] gnuradio/gr-filter/grc/filter_low_pass_xlating_filter.block.yml which has `2*${samp_rate}` in Line 70 and 78

mormj:
> Good question, I ran a query and found simmilar pitfalls in 37 more files. See my second commit in this PR 

Not seeing a second commit here - did you push it?

eltos:
> > Good question, I ran a query and found simmilar pitfalls in 37 more files. See my second commit in this PR
> 
> Not seeing a second commit here - did you push it?

Accidentally pushed to gitlab 🙈  
Now it's here

willcode:
I hadn't thought of people using an expression for `vlen`. Make sense to protect against it just in case.

dl1ksv:
But after sleeping one night: This can only be a temporary solution. Neither can we be sure that we changed all occurrences, or that new yml files will be  setup correctly ? What about OOT's.

So we should try to fix it on the grc level.

mormj:
That is a very good point - probably ok to merge this as a patch as a temporary fix, but definitely exposes a weakness in grc.

PR #5739, opened 2022-04-02
grpython use posix prefix scheme

maitbot:
Recently in Debian's pthon3.10 the default scheme for sysconfig
changed from 'posix_prefix' to 'posix_local'. GrPython.cmake
expects the 'posix_prefix' behavior. So explicity use it.

---
name: grpython use posix prefix scheme
about: GrPython.cmake
title: ''
labels: ''
assignees: ''

---

# Pull Request Details

## Description

Install path changed to /usr/local with python3.10 - even though the desire
is to set the install directories using gnuradio's cmake paths.

## Related Issue

Debian bug:
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1008854

Not only gnuradio itself - but every OOT project that inherits GrPython.cmake.

## Testing Done

Debian builds as part of the python3.10-default transition.

## Checklist

### DIFF
--- a/cmake/Modules/GrPython.cmake
+++ b/cmake/Modules/GrPython.cmake
@@ -159,7 +159,7 @@ except AttributeError: pass
 
 if not install_dir:
     #find where to install the python module
-    install_dir = sysconfig.get_path('platlib')
+    install_dir = sysconfig.get_path('platlib','posix_prefix')
     prefix = sysconfig.get_config_var('prefix')
 
 #strip the prefix to return a relative path


willcode (approved):
Since `prefix_posix` is the default scheme
https://docs.python.org/3/library/sysconfig.html#installation-paths
and we are not currently specifying a scheme, this seems safe.
@ryanvolz can you verify that this is ok for Windows, since there are separate schemes listed for nt.

ryanvolz:
Hmm, clearly there are no issues with the conda builds because the recipe sets GR_PYTHON_PATH (and that's still true for the conda-forge packages), so I don't expect to personally see any problems because of this change. But I do wonder about other theoretical Windows users since this does change the default scheme from "nt" to "posix_prefix" in that case. (It looks like there will be a new scheme option with Python 3.11 that would do what @mait wants but also keep the nt scheme on Windows, "venv", but obviously it would be a while until we could use that.) Maybe a ternary expression like `"posix_prefix" if os.name != "nt" else "nt"` would result in minimal change for users?

drmpeg (approved):
Tested on Ubuntu 22.04.

mormj:
Let's go ahead and merge this on `main` but hold off on backport, given the Windows concerns.

PR #5740, opened 2022-04-03
Fix Mach-kernel timebase

### LINKED ISSUE #5705: Mach kernel/M1 High-res timer, timebase incorrectly used. (No QT-GUI update from signal source (MacOS))
### What happened?

GUI is not updating after drawing the initial waveform. The GUI itself is perfectly responsive, just the signal doesn't seem to update.

### System Information

OS: MacOS, M1, Monterey 12.3
GR Installation Method: brew

### GNU Radio Version

3.9  (maint-3.9)

### Specific Version

3.9.3.0

Just running a simple signal source, throttle and QT Gui Time sink. I have a GUI Range to change frequency of the signal source, but unfortunately the GUI is not updating after drawing the initial waveform. The GUI itself is perfectly responsive, just the signal doesn't seem to update. I know the "simulation" is running, because removing the throttle block indicates a big cpu-spike.

I did implement the fix for the black screen in the companion. But i'm not sure how that would affect the running gnuradio python instance.

### Relevant log output

_No response_

wakass:
## Related Issue
Fixes #5705


willcode:
@wakass Needs a Signed-off-by line in the commit for DCO.
https://wiki.gnuradio.org/index.php/Development#What's_this_Developer_Certificate_of_Origin_(DCO)%3F

willcode:
Whenever you change a .h file that has python bindings, the corresponding pybind cc file needs to be checked. If there is no change in signature, this just involves recomputing the hash of the .h file. There is script for this, once you are sure you haven't change the signature of anything:

gr-utils/bindtool/scripts/binding-hash-hotfixer.zsh gnuradio-runtime/python/gnuradio/gr/bindings/high_res_timer_python.cc

wakass:
Thanks for the explanation!

mormj:
Rebased in #5910

PR #5743, opened 2022-04-04
Fix circular dependency, and clean up code getting around it.

orlando017:
## Description
Clean up code getting around circular dependency.

## Testing Done
I'm trying to build gnuradio with this change with cmake, but getting an error that the .h versions are out of sync; do I need to do something to get this to build?

  Python bindings for basic_block.h are out of sync
  gnuradio-runtime/python/gnuradio/gr/bindings/CMakeLists.txt:86 (GR_PYBIND_MAKE_CHECK_HASH)

## Checklist


willcode:
(1) Whenever you change a .h file that has python bindings, the corresponding pybind cc file needs to be checked. If there is no change in signature, this just involves recomputing the hash of the .h file. There is script for this, once you are sure you haven't change the signature of anything. For example:

gr-utils/bindtool/scripts/binding-hash-hotfixer.zsh gnuradio-runtime/python/gnuradio/gr/bindings/basic_block_python.cc

(2) Commit needs a Signed-off-by line in the commit for DCO.
https://wiki.gnuradio.org/index.php/Development#What's_this_Developer_Certificate_of_Origin_(DCO)%3F

(3) Please quash your commits, unless they are things that you want to be separate.

PR #5744, opened 2022-04-04
blocks: Synchronize access to d_level to prevent race conditions

### LINKED ISSUE #2206: qa: qa_filterbank finds some aligment issues
```
269/360 Test #269: qa_filterbank ......................................***Failed    0.57 sec
gr::buffer::allocate_buffer: warning: tried to allocate
   8 items of size 8000. Due to alignment requirements
   64 were allocated.  If this isn't OK, consider padding
   your structure to a power-of-two bytes.
   On this platform, our allocation granularity is 4096 bytes.
gr::buffer::allocate_buffer: warning: tried to allocate
   8 items of size 8000. Due to alignment requirements
   64 were allocated.  If this isn't OK, consider padding
   your structure to a power-of-two bytes.
   On this platform, our allocation granularity is 4096 bytes.
F
FAIL: test_000 (__main__.test_filterbank_vcvcf)
From: https://ci.gnuradio.org/buildbot/#/builders/49/builds/331/steps/5/logs/stdio
  File "/data/src/pull/2204/master/gr-filter/python/filter/qa_filterbank.py", line 128, in test_000
    self.assertComplexTuplesAlmostEqual(comb_results2, outdata2, 6)
  File "/data/src/pull/2204/master/gnuradio-runtime/python/gnuradio/gr_unittest.py", line 86, in assertComplexTuplesAlmostEqual
    self.assertComplexAlmostEqual (a[i], b[i], places, msg)
  File "/data/src/pull/2204/master/gnuradio-runtime/python/gnuradio/gr_unittest.py", line 54, in assertComplexAlmostEqual
    msg or '%r != %r within %r places' % (first, second, places))
AssertionError: (0.03780283836319677-0.13573173385334286j) != (0.04380534589290619+0.08077553659677505j) within 6 places
Ran 1 test in 0.047s
FAILED (failures=1)
```

I'm pulling this out of #1261 since I couldn't reproduce this on my on machine.
Bash line to test for failures:
```
for i in {1..1000}; do echo "$i. run"; result=$(ctest --output-on-failure -R qa_filterbank 2>&1); if [ $? -ne 0 ]; then echo "$i. run failed\n"; echo $result >> /tmp/error.log; fi; done
```
It's not entirely clear if that's a CI shortcoming or a serious issue.

### LINKED ISSUE #4322: qa_filterbank flakiness
```
133/241 Test #133: qa_filterbank ................................***Failed    0.26 sec
F
FAIL: test_000 (__main__.test_filterbank_vcvcf)
Generates nfilts sets of random complex data.
  File "/__w/gnuradio/gnuradio/gr-filter/python/filter/qa_filterbank.py", line 120, in test_000
    self.assertComplexTuplesAlmostEqual(comb_results2, outdata2, 6)
  File "/__w/gnuradio/gnuradio/gnuradio-runtime/python/gnuradio/gr_unittest.py", line 70, in assertComplexTuplesAlmostEqual
    return all([
  File "/__w/gnuradio/gnuradio/gnuradio-runtime/python/gnuradio/gr_unittest.py", line 71, in <listcomp>
    self.assertComplexAlmostEqual(x, y, places, msg)
  File "/__w/gnuradio/gnuradio/gnuradio-runtime/python/gnuradio/gr_unittest.py", line 32, in assertComplexAlmostEqual
    raise self.failureException(
AssertionError: (-0.14535445023395024-0.020636200107483925j) != (-0.10142534971237183-0.014487081207334995j) within 6 places
```

https://github.com/gnuradio/gnuradio/pull/4298/checks?check_run_id=2013512503

argilo:
## Description
While looking into #4322, I found that the root cause of the issue is that the Probe Signal Vector block is not synchronized: if `level()` is called while the work function is writing to `d_level`, then invalid data will be returned. The same is true of the Probe Signal block. To fix these problems, I did the following:

1. In the Probe Signal block, change `d_level` from `T` to `std::atomic<T>`.
2. In the Probe Signal Vector block, use a mutex to synchronize access to `d_level`.

I also added regression tests which readily fail without these changes, and pass with them.

## Related Issue
* Fixes #2206.
* Fixes #4322.

* Probe Signal
* Probe Signal Vector

## Testing Done
* Verified that `qa_filterbank` no longer fails, even when its `sleep`s are removed.
* Verified that the new regression tests fail without the code changes, and pass with them.

## Checklist

### DIFF
--- a/gr-blocks/lib/CMakeLists.txt
+++ b/gr-blocks/lib/CMakeLists.txt
@@ -222,6 +222,10 @@ IF(MSVC)
     )
 ENDIF(MSVC)
 
+if(MSVC)
+    target_compile_definitions(gnuradio-blocks PUBLIC _ENABLE_ATOMIC_ALIGNMENT_FIX)
+endif()
+
 if(BUILD_SHARED_LIBS)
     if (SNDFILE_FOUND)
         GR_LIBRARY_FOO(gnuradio-blocks SNDFILE)
--- a/gr-blocks/lib/probe_signal_impl.h
+++ b/gr-blocks/lib/probe_signal_impl.h
@@ -13,6 +13,7 @@
 #define PROBE_SIGNAL_IMPL_H
 
 #include <gnuradio/blocks/probe_signal.h>
+#include <atomic>
 
 namespace gr {
 namespace blocks {
@@ -21,7 +22,7 @@ template <class T>
 class probe_signal_impl : public probe_signal<T>
 {
 private:
-    T d_level;
+    std::atomic<T> d_level;
 
 public:
     probe_signal_impl();
--- a/gr-blocks/lib/probe_signal_v_impl.cc
+++ b/gr-blocks/lib/probe_signal_v_impl.cc
@@ -47,6 +47,7 @@ int probe_signal_v_impl<T>::work(int noutput_items,
 {
     const T* in = (const T*)input_items[0];
 
+    gr::thread::scoped_lock guard(d_mutex);
     for (size_t i = 0; i < d_size; i++)
         d_level[i] = in[(noutput_items - 1) * d_size + i];
 
@@ -57,6 +58,13 @@ template class probe_signal_v<std::int16_t>;
 template class probe_signal_v<std::int32_t>;
 template class probe_signal_v<float>;
 template class probe_signal_v<gr_complex>;
+
+template <class T>
+std::vector<T> probe_signal_v_impl<T>::level() const
+{
+    gr::thread::scoped_lock guard(d_mutex);
+    return d_level;
+}
 
 } /* namespace blocks */
 } /* namespace gr */
--- a/gr-blocks/lib/probe_signal_v_impl.h
+++ b/gr-blocks/lib/probe_signal_v_impl.h
@@ -23,12 +23,13 @@ class probe_signal_v_impl : public probe_signal_v<T>
 private:
     std::vector<T> d_level;
     const size_t d_size;
+    mutable gr::thread::mutex d_mutex;
 
 public:
     probe_signal_v_impl(size_t size);
     ~probe_signal_v_impl() override;
 
-    std::vector<T> level() const override { return d_level; }
+    std::vector<T> level() const override;
 
     int work(int noutput_items,
              gr_vector_const_void_star& input_items,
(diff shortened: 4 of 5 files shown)


argilo:
It looks like the bug has existed since the Probe Signal blocks were added in GNU Radio 3.6.0: https://github.com/gnuradio/gnuradio/commit/0cbaf1b350c5ef6fc25f429430f1c892e7844e01

argilo:
The Windows build currently fails to build with this error:

> C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Tools\MSVC\14.16.27023\include\atomic(620): error C2338: You've instantiated std::atomic<T> with sizeof(T) equal to 2/4/8 and alignof(T) < sizeof(T). Before VS 2015 Update 2, this would have misbehaved at runtime. VS 2015 Update 2 was fixed to handle this correctly, but the fix inherently changes layout and breaks binary compatibility. Please define _ENABLE_ATOMIC_ALIGNMENT_FIX to acknowledge that you understand this, and that everything you're linking has been compiled with VS 2015 Update 2 (or later).

argilo:
`_ENABLE_ATOMIC_ALIGNMENT_FIX` sounds OK at least for GNU Radio 3.9+, since 3.9 requires VS 2017.

willcode (approved):
Cool. I haven't used `mutable`, so no opinion there.

argilo:
> Cool. I haven't used `mutable`, so no opinion there.

`mutable` is needed to allow the mutex to be used from a `const` member function (`level()`). That seemed better than removing `const`.

argilo:
I added the `_ENABLE_ATOMIC_ALIGNMENT_FIX` to Visual Studio builds as suggested by the error message, and the Windows build now works.

gr-blocks/lib/CMakeLists.txt
```diff
@@ -222,6 +222,10 @@ IF(MSVC)
     )
 ENDIF(MSVC)
 
+if(MSVC)
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -D_ENABLE_ATOMIC_ALIGNMENT_FIX")
```
ryanvolz, line 226: I'll suggest
```
target_compile_definitions(gnuradio-blocks PUBLIC _ENABLE_ATOMIC_ALIGNMENT_FIX)
```
here instead.

argilo: Done.

ryanvolz (approved):

mormj (approved):

PR #5746, opened 2022-04-05
runtime: Fix flaky qa_uncaught_exception test

argilo:
## Description
The `qa_uncaught_exception` test sometimes fails in CI, for instance: https://github.com/gnuradio/gnuradio/runs/5827362248?check_suite_focus=true

This happens because the flow graph may take more than 2.5 seconds to execute. (The timeout was previously increased from 0.5 seconds to 2.5 seconds in https://github.com/gnuradio/gnuradio/pull/3029, but it's still not enough.)

Here I've increased the timeout to 10 seconds. This doesn't slow down the test in the success case, since the `join` returns as soon as the flow graph finishes.

## Related Issue
* https://github.com/gnuradio/gnuradio/pull/3029

None. This change only affects tests.

## Testing Done
Played with the test locally and verified that it fails when the timeout is too low, and that increasing the timeout doesn't slow down the test.

## Checklist

### DIFF
--- a/gnuradio-runtime/python/gnuradio/gr/qa_uncaught_exception.py
+++ b/gnuradio-runtime/python/gnuradio/gr/qa_uncaught_exception.py
@@ -62,7 +62,7 @@ class test_uncaught_exception(gr_unittest.TestCase):
         p = Process(target=process_func, args=(False,))
         p.daemon = True
         p.start()
-        p.join(2.5)
+        p.join(10.0)
         exit_code = p.exitcode
         self.assertIsNotNone(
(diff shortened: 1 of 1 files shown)


argilo:
I intentionally did not increase the timeout in the `test_exception_throw_caught` test, since that would slow down CI. If the `join` timeout occurs before the handled exception occurs, it does not result in a CI failure.

mbr0wn (approved):
Why not, no harm.... but why does it take so long to execute?

argilo:
> but why does it take so long to execute?

I don't know for sure, but I'd guess it's due to high load on the CI machine.

PR #5748, opened 2022-04-05
Replaced some star imports in __init__.py files

andydazzo7:
# Pull Request Details
In this PR I am attempting to tackle https://github.com/gnuradio/gnuradio/issues/1466, which states that there are far too many star imports in the init files for most of the python packages. 

## Description
I went through the python packages and targeted certain init files that had many star imports. I was thus able to find the files and classes that were actually being imported in these init files, and then changed the star imports into explicit imports.

## Related Issue
https://github.com/gnuradio/gnuradio/issues/1466

## Checklist

### DIFF
--- a/gr-digital/python/digital/__init__.py
+++ b/gr-digital/python/digital/__init__.py
@@ -29,7 +29,7 @@ from .psk import (
     psk_demod,
     psk_constellation,
 )
-from .qam import(
+from .qam import (
     is_power_of_four,
     get_bit,
     get_bits,
--- a/gr-digital/python/digital/__init__.py
+++ b/gr-digital/python/digital/__init__.py
@@ -23,20 +23,38 @@ except ImportError:
     from .digital_python import *
 
 from gnuradio import analog  # just need analog for the enum
-from .psk import *
-from .qam import *
-from .qamlike import *
-from .bpsk import *
-from .qpsk import *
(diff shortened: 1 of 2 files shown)


gr-digital/python/digital/__init__.py
```diff
@@ -23,20 +23,38 @@
...
+    psk_constellation,
+)
+from .qam import(
+    is_power_of_four,
+    get_bit,
+    get_bits,
+    make_differential_constellation,
+    make_non_differential_constellation,
+    large_ampls_to_corners_mapping
+)
```
mbr0wn, line 39: Maybe for stuff like this we just do
```suggestion
from . import qam
```

...and then amend the API calls below. With calls like `get_bits()`, they are pretty ambiguously named, and it might  be nicer to not have them in the main namespace. But that would be an additional change over what you have already.

mbr0wn (commented):
Thanks for your contribution! And welcome here.

I think the changes are fine, but please fix the formatting issues (see CI run) and when you're done, squash the commits and make sure you use the git commit formatting convention, too.

mormj:
Closing this as it does not pass CI, and also, @mbr0wn's suggestion would make things generally cleaner.  Please reopen if this PR is revisited.  Thanks!

PR #5753, opened 2022-04-06
filter: Fix various bugs in Generic Filterbank

argilo:
## Description
The `qa_pfb_channelizer` sometimes fails in CI. For example: https://github.com/gnuradio/gnuradio/runs/5535718772?check_suite_focus=true

Running the test in valgrind reveals two bugs:

* `filterbank_vcvcf_impl::general_work` reads `d_updated` before it is initialized.
* `filterbank_vcvcf_impl::general_work` passes an unaligned buffer (`working`) into `fir_filter::filter`, which requires its input buffer to be aligned.

The second bug can result in `volk_32fc_32f_dot_prod_32fc_a` reading data from outside the `working` buffer, corrupting the filter calculation. (Out-of-bounds values are multiplied by zero, but this can still result in `NaN`, as seen in the CI failure.)

To fix these problems, I did the following:

* Initialize `d_updated` to false in the constructor.
* Allocate `working` using ~~`volk_malloc`~~ `volk::vector`.

I also fixed two other minor issues I noticed along the way:

* The history is set one sample too long, which causes the output to be delayed by one sample.
* The `working` buffer is one sample too long, which wastes a small amount of memory.

* Generic Filterbank
* Hierarchical Polyphase Channelizer (which uses Generic Filterbank under the covers)

## Testing Done
* Verified that `qa_pfb_channelizer` still works correctly and does not produce any valgrind errors.

## Checklist

### DIFF
--- a/gr-filter/lib/filterbank_vcvcf_impl.cc
+++ b/gr-filter/lib/filterbank_vcvcf_impl.cc
@@ -28,16 +28,17 @@ filterbank_vcvcf_impl::filterbank_vcvcf_impl(const std::vector<std::vector<float
     : block("filterbank_vcvcf",
             io_signature::make(1, 1, sizeof(gr_complex) * taps.size()),
             io_signature::make(1, 1, sizeof(gr_complex) * taps.size())),
-      filterbank(taps)
+      filterbank(taps),
+      d_updated(false)
 {
-    set_history(d_ntaps + 1);
+    set_history(d_ntaps);
 }
 
 void filterbank_vcvcf_impl::set_taps(const std::vector<std::vector<float>>& taps)
 {
     gr::thread::scoped_lock guard(d_mutex);
     filterbank::set_taps(taps);
-    set_history(d_ntaps + 1);
+    set_history(d_ntaps);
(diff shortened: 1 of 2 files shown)


gr-filter/lib/filterbank_vcvcf_impl.cc
```diff
@@ -63,7 +64,8 @@ int filterbank_vcvcf_impl::general_work(int noutput_items,
         return 0; // history requirements may have changed.
     }
 
-    std::vector<gr_complex> working(noutput_items + d_ntaps);
+    gr_complex* working = (gr_complex*)volk_malloc(
```
jdemel, line 67: Instead of `volk_malloc` you can use `volk::vector<gr_complex>` that uses `volk_malloc`. You can still pass pointers around via `vec.data()` but you don't need manual memory management anymore.

argilo: Oh, nice! I'll make that change.

argilo:
Looks like I broke `qa_filterbank`. Investigating now.

PR #5754, opened 2022-04-07
runtime: Process system messages before others

### LINKED ISSUE #4348: qa_packet_headerparser_b: Flaky

argilo:
## Description
Various tests involving message passing fail intermittently, for example `qa_packet_headerparser_b`. One or more messages fail to arrive at the `message_debug` block when the flow graph stops processing samples. This occurs due to a race condition in `tpb_thread_body`.

The [main loop of `tpb_thread_body`](https://github.com/gnuradio/gnuradio/blob/3349d483a1f7e5db0eea042815d237810373857b/gnuradio-runtime/lib/tpb_thread_body.cc#L70-L152) iterates over the message queues in an undefined order, reading pending messages from each. One of these queues is the special "system" queue; if a "done" message arrives on this queue, then the corresponding handler sets `d_finished` to true, which will later cause the main loop to exit:

https://github.com/gnuradio/gnuradio/blob/3349d483a1f7e5db0eea042815d237810373857b/gnuradio-runtime/lib/block.cc#L759-L770

The following race condition can occur:

1. The main loop reads from the "foo" message queue, and finds it empty.
2. A message arrives on the "foo" queue.
3. A "done" message arrives on the "system" queue.
4. The main loop reads from the "system" queue, finds the "done" message, and sets `d_finished` to true.
5. The main loop exits, because `d_finished` is true.

The message on the "foo" queue is never processed.

To solve this problem, I've used a custom comparator to order the message queues such that the "system" queue comes first. This guarantees that after a "done" message is processed, all other queues will be processed at least once before the main loop exits.

## Related Issue
Fixes #4348.

Message passing in the thread-per-block scheduler.

## Testing Done
I ran `qa_packet_headerparser_b` in a loop (`while gr-digital/python/digital/qa_packet_headerparser_b_test.sh; do :; done`) to verify that it no longer fails intermittently.

## Checklist

### DIFF
--- a/gnuradio-runtime/include/gnuradio/basic_block.h
+++ b/gnuradio-runtime/include/gnuradio/basic_block.h
@@ -28,6 +28,25 @@
 
 namespace gr {
 
+class msg_queue_comparator
+{
+private:
+    const pmt::pmt_t d_system_port = pmt::intern("system");
+
+public:
+    bool operator()(pmt::pmt_t const& queue_key1, pmt::pmt_t const& queue_key2) const
+    {
+        if (pmt::eqv(queue_key2, d_system_port))
+            return false;
+        else if (pmt::eqv(queue_key1, d_system_port))
+            return true;
+        else {
+            pmt::comparator cmp;
+            return cmp(queue_key1, queue_key2);
+        }
+    }
+};
+
 /*!
  * \brief The abstract base class for all signal processing blocks.
  * \ingroup internal
@@ -49,8 +68,8 @@ private:
     d_msg_handlers_t d_msg_handlers;
 
     typedef std::deque<pmt::pmt_t> msg_queue_t;
-    typedef std::map<pmt::pmt_t, msg_queue_t, pmt::comparator> msg_queue_map_t;
-    typedef std::map<pmt::pmt_t, msg_queue_t, pmt::comparator>::iterator
+    typedef std::map<pmt::pmt_t, msg_queue_t, msg_queue_comparator> msg_queue_map_t;
+    typedef std::map<pmt::pmt_t, msg_queue_t, msg_queue_comparator>::iterator
         msg_queue_map_itr;
 
     gr::thread::mutex mutex; //< protects all vars
--- a/gnuradio-runtime/python/gnuradio/gr/bindings/basic_block_python.cc
+++ b/gnuradio-runtime/python/gnuradio/gr/bindings/basic_block_python.cc
@@ -14,7 +14,7 @@
 /* BINDTOOL_GEN_AUTOMATIC(0)                                                       */
 /* BINDTOOL_USE_PYGCCXML(0)                                                        */
 /* BINDTOOL_HEADER_FILE(basic_block.h)                                             */
-/* BINDTOOL_HEADER_FILE_HASH(53f812404aa54083e64261ba5b5cf26c)                     */
+/* BINDTOOL_HEADER_FILE_HASH(97ef3f809e58b459c57d41bab36bb97d)                     */
 /***********************************************************************************/
 
 #include <pybind11/complex.h>


argilo:
Something is broken, because `set_msg_handler` is now raising this exception when I start up Gqrx:

https://github.com/gnuradio/gnuradio/blob/3349d483a1f7e5db0eea042815d237810373857b/gnuradio-runtime/include/gnuradio/basic_block.h#L386-L394

argilo:
It appears the exception is thrown when trying to register a `system` handler for the `fix_cc` block in gr-iqbal. The problem only occurs some of the time.

argilo:
Never mind. The issue was simply that I needed to rebuild gr-iqbal to pick up the change to `basic_block.h`.

willcode (approved):
Since the order is currently undefine, this shouldn't hurt anything. It makes sense to process system messages first, giving them a higher priority than other messages.

gnuradio-runtime/include/gnuradio/basic_block.h
```diff
@@ -28,6 +28,21 @@
...
+public:
+    bool operator()(pmt::pmt_t const& p1, pmt::pmt_t const& p2) const
+    {
+        std::string s1 = pmt::symbol_to_string(p1);
+        std::string s2 = pmt::symbol_to_string(p2);
+        if (s1 == "system")
+            s1 = "";
+        if (s2 == "system")
+            s2 = "";
+        return s1 < s2;
```
jdemel, line 42: First off, message queues should always be identified via pmts. cf. https://github.com/gnuradio/gnuradio/blob/fa8953de2ae67990ccb052af08affac073e106ae/gnuradio-runtime/lib/basic_block.cc#L76
and the following methods. Also, `is_symbol()` requires that the PMT is a string.
https://github.com/gnuradio/gnuradio/blob/fa8953de2ae67990ccb052af08affac073e106ae/gnuradio-runtime/lib/pmt/pmt_int.h#L44

I suggest to add a private member to `msg_queue_comparator`
```c++
const pmt::pmt_t d_system_queue_key = pmt::mp("system");
```
Then you can compare the inputs `p1` and `p2` via `pmt::equal`. It looks like this is how its done in the `pmt` implementation. It might avoid some corner cases. In case it is not the system queue, I suggest to fall back to `pmt::comparator`.

gnuradio-runtime/include/gnuradio/basic_block.h
```diff
@@ -28,6 +28,21 @@
 
 namespace gr {
 
+class msg_queue_comparator
+{
+public:
+    bool operator()(pmt::pmt_t const& p1, pmt::pmt_t const& p2) const
```
jdemel, line 34: I suggest to give the parameters more descriptive names. e.g. `queue_key0`.

jdemel (changes requested):
Thanks for this PR. I have a few suggestions.

gnuradio-runtime/include/gnuradio/basic_block.h
```diff
@@ -28,6 +28,28 @@
...
+private:
+    const pmt::pmt_t d_system_port = pmt::intern("system");
+
+public:
+    bool operator()(pmt::pmt_t const& queue_key1, pmt::pmt_t const& queue_key2) const
+    {
+        bool is_system1 = pmt::eqv(queue_key1, d_system_port);
+        bool is_system2 = pmt::eqv(queue_key2, d_system_port);
+
+        if (is_system1 && is_system2)
```
willcode, line 42: This is equivalent to
```
if (is_system2) return false;
else if (is_system1) return true;
else ...
```
but perhaps it's more obvious the way you have it.

argilo: I left it the long way since it seemed a bit more understandable. I'll leave it unless there's a strong preference to shorten it.

willcode: Nah, the compiler will take care of it anyway.

jdemel: I suggest to do a quick check with https://godbolt.org/
You never know what your compiler does.

gnuradio-runtime/include/gnuradio/basic_block.h
```diff
@@ -28,6 +28,28 @@
...
+            return false;
+        else if (is_system1 && !is_system2)
+            return true;
+        else if (!is_system1 && is_system2)
+            return false;
+        else
+            return queue_key1 < queue_key2;
```
jdemel, line 49: How is this comparison defined? Would it be an option to use `pmt::comparator`?

jdemel (commented):
Thanks for the updates. I have two more comments. Besides LGTM.

gnuradio-runtime/include/gnuradio/basic_block.h
argilo: The code is actually quite a bit simpler the other way, since the `is_system1` and `is_system2` don't need to be separately defined. I'll rewrite it.

gnuradio-runtime/include/gnuradio/basic_block.h
argilo: I incorrectly assumed it would use the comparator. I'll change the code to explicitly call it.

argilo:
The code is now refactored, and I verified that `qa_packet_headerparser_b_test` still passes reliably.

PR #5755, opened 2022-04-07
blocks: Fix flaky qa_rotator test

argilo:
## Description
The `test_phase_inc_update_out_of_range` test in `qa_rotator` sometimes fails. For instance: https://github.com/gnuradio/gnuradio/runs/5772119971?check_suite_focus=true

This is due to a race condition. The test schedules a phase increment for 32,768 samples in the future, waits for some samples to be processed, and then verifies that the phase increment has not yet occurred. But it can happen (and often does happen with the system is under load) that more than 32,768 samples will be processed between these two steps.

https://github.com/gnuradio/gnuradio/blob/3349d483a1f7e5db0eea042815d237810373857b/gr-blocks/python/blocks/qa_rotator_cc.py#L286-L318

To solve the problem, I've added a throttle block to the flow graph, so that it will take 500 ms for 32,768 samples to be processed. This should be more than enough time for the test to progress from the `while (self.rotator_cc.nitems_written(0) == 0)` step to the `self._assert_tags([], [])` step.

QA tests for the Rotator block.

## Testing Done
* Verified that the test passes even when the system is under heavy load (e.g. `stress -c 32`).

## Checklist

### DIFF
--- a/gr-blocks/python/blocks/qa_rotator_cc.py
+++ b/gr-blocks/python/blocks/qa_rotator_cc.py
@@ -39,12 +39,13 @@ class qa_rotator_cc(gr_unittest.TestCase):
         #
         self.tb = gr.top_block()
         self.source = blocks.vector_source_c(in_samples)
+        self.throttle = blocks.throttle(gr.sizeof_gr_complex, 2**16)
         self.rotator_cc = blocks.rotator_cc(phase_inc, tag_inc_updates)
         self.sink = blocks.vector_sink_c()
         self.tag_sink = blocks.tag_debug(gr.sizeof_gr_complex, "rot_phase_inc",
                                          "rot_phase_inc")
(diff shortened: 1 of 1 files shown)


marcusmueller:
I know this sounds super picky, and it honestly feels very ungrateful, but could you mention the race condition (or just copy&paste the full description you wrote) in the commit message? I see myself scratching my head very much, and me or someone else removing a "superfluous" throttle slowing down our CI in the future. Even just mentioning "blocks: Fix race condition leading to flaky qa_rotator" would be super awesome to document *why* you're adding the throttle, not just *that* you're adding it :)

argilo:
Good point. I've updated the commit message.

777arc (approved):

mormj (approved):

PR #5763, opened 2022-04-12
Additional error messages for assertions (without the error messages)

mormj:
This is a continuation of https://github.com/gnuradio/gnuradio/pull/5469

Keeps the comments in the yml files describing what the assertion values mean, but does not modify grc to parse based on the `#` character


dl1ksv:
I believe, this is a real improvement

PR #5772, opened 2022-04-17
gr-soapy: deactivate stream before closing

willcode:
# Pull Request Details

It was reported that, with the HackRF module, a flowgraph terminated
with a segfault. Most Soapy modules implement deactivateStream(), but
that function is not called from gr-soapy.

Problem reported by Gavin Jacobs on the mailing list.

gr-soapy

## Checklist

### DIFF
--- a/gr-soapy/lib/block_impl.cc
+++ b/gr-soapy/lib/block_impl.cc
@@ -383,6 +383,7 @@ bool block_impl::stop()
 {
     if (d_stream) {
         std::lock_guard<std::mutex> l(d_device_mutex);
+        d_device->deactivateStream(d_stream);
         d_device->closeStream(d_stream);
         d_stream = nullptr;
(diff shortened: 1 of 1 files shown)


willcode:
This needs some testing with various SDR modules before merge. The RTLSDR module calls `deactivateStream()` from `closeStream()`. Most other modules look like they expect us to call `deactivateStream()`.

willcode:
Tested with RTLSDR, HackRF and AirspyHF. The HackRF segv (return code -11) goes away with this patch. The other SDR did not show any different behavior (which is good). So, I think this is safe and correct.

willcode:
@ncorgan Could you take a look at this? Thanks.

ncorgan:
Looks good to me.

mormj (approved):

PR #5774, opened 2022-04-18
gr-qtgui: add gr-display functionallity to gr-qtgui

dl1ksv:
# Pull Request Details

## Description

This pr adds three modules to gr-qtgui:
 - display_image_sink ( QTGUI Image Sink in grc)
 -  display_text_msg    (QTGUI Text Msg Sink in grc)
 - display_text_sink     (QTGUI Text Sink in grc)
 
None
## Testing Done

display_image_sink was tested with an image received from NOAA19.
For testing display_text_sink and display_text_msg see the examples 
qtgui_display_text_sink.grc
qtgui_display_text_msg.grc
## Checklist


mormj:
These blocks seem very useful - any way to add an example flowgraph for the image sink?

dl1ksv:
I have some examples  using the image sink. They all deal with decoding and displaying wav- files received from Noaa weather satellites. The files are about 200 MB. Of course I could only upload the grc file(s) ? How should I proceed ?

mormj:
Having the .grc files would be helpful - there are examples with dtv where the binary files are hosted elsewhere (we need as a project a common place to host these files).  Also, once merged, having an example flowgraph in the wiki showing the result of the flowgraph would be helpful

I just don't have a good way to see what the widget looks like currently

dl1ksv:
@777arc The wav files represent 16 bit stereo samples sampled at 192K. The gnuradio wav file source converts the samples to two streams (I/q) of float. So I'd need a wav to SigMF converter and a SigMF gnuradio source to generate and handle these data.

777arc:
Turns out don't need wav to SigMF, you just need wav to binary IQ, which you can do within GNU Radio (File Sink creates binary IQ).  I can help make them SigMF formatted but it's essentially just renaming the binary IQ file to something.sigmf-data and then creating a something.sigmf-meta that is just json and has info like sample rate and center freq in it, if you are curious check out https://pysdr.org/content/iq_files.html#sigmf-and-annotating-iq-files

But anyway the reason I asked is because we won't be able to put the wav/IQ files in the repo obviously but we are working on collecting a bunch of SigMF example files that could be used to test flowgraphs and various GR functionality, and we are storing them together somewhere (might be Azure blob storage, still TBD).  And since SigMF files are essentially just binary IQ files with a metadata file next to them, all we would have to do is that simple conversion I mentioned.

dl1ksv:
My example flowgraph is capable of using wav files or sigmf data files. You just have to activate/deactivate the appropriate blocks in the flowgraph.

But the sigmf data file is about twice as big as the wav file, as the wav file stores the values as integer while the sigmf data are floats.

mormj:
@dl1ksv - thank you for putting together this PR to bring functionality to the main tree.  I feel this are very useful blocks, but given the dependence on external files for solid examples, it is probably best left as a well-maintained OOT.  This might be worth revisiting, but going to close for now.

PR #5799, opened 2022-05-01
GrPybind: GR_PYBIND_MAKE_CHECK_HASH enable include option

dl1ksv:
# Pull Request Details

In the GR_PYBIND_MAKE_CHECK_HASH the include option is disabled.
So the python bindings of the qt-gui modules can't be rebuild, as the
the include directories of qt5 and qwt are not used at this point.

Therefore in gr-qtgui/python/qtgui/bindings/CMakeLists.txt GR_PYBIND_MAKE
is used instead of GR_PYBIND_MAKE_CHECK_HASH and the binding files *_python.cc
have no header section.
## Description

Now the include option is enabled in GR_PYBIND_MAKE_CHECK_HASH and uses
the variable extra_includes.
To be compatible with the actual calls to GR_PYBIND_MAKE_CHECK_HASH
the argument for the include parameter in bind_intree_file.py may be empty now.
So you can switch from GR_PYBIND_MAKE to GR_PYBIND_MAKE_CHECK_HASH
in gr-qtgui/python/qtgui/bindings/CMakeLists.txt without any problems.

But #5775 should be applied before to get the correct binding for the qwidget() function.

## Testing Done

The second commit  demonstrates how to build the bindings inside the cmake process.
## Checklist


dl1ksv:
The C++ formating error for sink_c_pydoc_template.h will be fixed by #5775

mormj (approved):

mormj:
still seems to be failing the c++ formatting check

dl1ksv:
Yes, that's correct. The file sink_c_pydoc_template.h is not correctly formatted and should be regenerated during the make run. So the c++ formating check must fail in the beginning and should succeed after make. But the formatting check is not setup for this.
Another problem:
If we  force a regen for sink_c.h, perhaps by changing the hash in sink_c_python.cc the build will fail with:

  File "/home/schroer/gnuradiocomponents/gnuradio-volker/gr-utils/bindtool/scripts/bind_intree_file.py", line 3, in <module>
    from gnuradio.bindtool import BindingGenerator
ModuleNotFoundError: No module named 'gnuradio'
make[2]: *** [gr-qtgui/python/qtgui/bindings/CMakeFiles/sink_c_python.cc_regen_bindings.dir/build.make:73: ../gr-qtgui/python/qtgui/bindings}/sink_c_python.cc] Fehler 1
make[1]: *** [CMakeFiles/Makefile2:6828: gr-qtgui/python/qtgui/bindings/CMakeFiles/sink_c_python.cc_regen_bindings.dir/all] Fehler 2
make: *** [Makefile:146: all] Fehler 2
As the module gnuradio will be setup on install and is not available during a fresh build from source. So you need an running gnuradio to do a regen in a build from source. And I don't know who to handle this inside these checks.

But I'll push a correctly formated sink_c_pydoc_template.h.

willcode:
This looks safe for backport, except maybe for the loss of `qwt_symbol.h`, which seems like a fairly obscure transitive dep. Comments?

dl1ksv:
I think, you could backport , but as this patch is only for those of interest, who want to modify and rebuild qtgui blocks, I'd  not backport this.

PR #5810, opened 2022-05-04
runtime: Silence unused parameter warnings

argilo:
## Description
CI build logs show 63 warnings for unused parameters in `rpcregisterhelpers.h`. Adding `[[maybe_unused]]` should silence the warnings.

I also replaced `__GR_ATTR_UNUSED` with C++17's `[[maybe_unused]]` elsewhere in GNU Radio. I left the macro in place, since removing it would be an API change.

## Checklist

### DIFF
--- a/gnuradio-runtime/include/gnuradio/rpcregisterhelpers.h
+++ b/gnuradio-runtime/include/gnuradio/rpcregisterhelpers.h
@@ -802,8 +802,8 @@ struct rpcbasic_register_set : public rpcbasic_base {
      * \param display_     The display mask
      */
     rpcbasic_register_set(const std::string& block_alias,
-                          const char* functionbase,
-                          void (T::*function)(Tto),
+                          [[maybe_unused]] const char* functionbase,
+                          [[maybe_unused]] void (T::*function)(Tto),
                           const pmt::pmt_t& min,
                           const pmt::pmt_t& max,
                           const pmt::pmt_t& def,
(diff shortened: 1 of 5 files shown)


gnuradio-runtime/include/gnuradio/rpcregisterhelpers.h
```diff
@@ -802,8 +802,8 @@ struct rpcbasic_register_set : public rpcbasic_base {
      * \param display_     The display mask
      */
     rpcbasic_register_set(const std::string& block_alias,
-                          const char* functionbase,
-                          void (T::*function)(Tto),
+                          __GR_ATTR_UNUSED const char* functionbase,
```
mormj, line 805: C++17 introduced [[[maybe_unused](https://en.cppreference.com/w/cpp/language/attributes/maybe_unused)]] that we use elsewhere for variables

I see __GR_ATTR_UNUSED in our codebase for functions that are not used, maybe this should become `[[maybe_unused]]`here for consistency

argilo: Oh, nice! I guess I can just convert everything over to `[[maybe_unused]]` in this PR then.

mormj:
I don't think it will be an issue, but this does introduce C++17 specific behavior into a public header - 
This tripped us up in the past with std::any's and pygccxml's lagging c++ compiler support.

If no one objects, I think we should go ahead and merge this, but hold off backport

argilo:
An alternative would be to switch back to `__GR_ATTR_UNUSED` for now. What do you think?

mormj:
It should be fine with the `[[maybe_unused]]` - looking through the code it does exist in other public headers.  I still think we should hold the backport for now though

argilo:
Sounds fine to me. Since it's just silencing warnings, there's no need for it to be backported.

willcode:
Changes would propegate to OOTs via `basic_block.h`. In 3.11, we should say that OOTs must use a compiler with the same capabilities as GR.

marcusmueller:
C++ dictates that unknown `[[ … ]]` attributes be ignored, so backporting this is fine (done so)

PR #5804, opened 2022-05-03
dtv: Fix use of uninitialized memory

argilo:
## Description

Valgrind reports that `poly_pack` reads uninitialized memory, leading to undefined behaviour. This happens because it rounds the length up to the next multiple of 32 before reading the polynomial coefficients. Initializing the coefficients to zero solves the problem.

```
==1539339== Conditional jump or move depends on uninitialised value(s)
==1539339==    at 0x4C3174CE: poly_pack (dvbt2_framemapper_cc_impl.cc:994)
==1539339==    by 0x4C3174CE: gr::dtv::dvbt2_framemapper_cc_impl::bch_poly_build_tables() (dvbt2_framemapper_cc_impl.cc:1032)
==1539339==    by 0x4C31AC60: gr::dtv::dvbt2_framemapper_cc_impl::dvbt2_framemapper_cc_impl(gr::dtv::dvb_framesize_t, gr::dtv::dvb_code_rate_t, gr::dtv::dvb_constellation_t, gr::dtv::dvbt2_rotation_t, int, int, gr::dtv::dvbt2_extended_carrier_t, gr::dtv::dvbt2_fftsize_t, gr::dtv::dvb_guardinterval_t, gr::dtv::dvbt2_l1constellation_t, gr::dtv::dvbt2_pilotpattern_t, int, int, gr::dtv::dvbt2_papr_t, gr::dtv::dvbt2_version_t, gr::dtv::dvbt2_preamble_t, gr::dtv::dvbt2_inputmode_t, gr::dtv::dvbt2_reservedbiasbits_t, gr::dtv::dvbt2_l1scrambled_t, gr::dtv::dvbt2_inband_t) (dvbt2_framemapper_cc_impl.cc:266)
[...]

PR #5816, opened 2022-05-05
dtv: Fix compiler warning

argilo:
## Description

In #5804 I accidentally caused a compiler warning:

```
warning: suggest braces around initialization of subobject [-Wmissing-braces]
```

This change fixes the warning while still initializing the polyout array to zero.

## Related Issue
* #5804

* DVB-T2 Frame Mapper

## Testing Done
I verified that `qa_dtv` still passes and does not produce any valgrind warnings.

## Checklist


drmpeg (approved):
It just wants the initializer to be two dimensional. So ```int polyout[2][200] = { { 0 } };``` would also work if you want to show more intent.

argilo:
> `int polyout[2][200] = { { 0 } };` would also work if you want to show more intent.

I think leaving it empty is fine, and that seems to be the convention elsewhere in GNU Radio.

drmpeg:
Agreed.

PR #5830, opened 2022-05-06
digital: Fix CRC checking in Async CRC16 and Async CRC32 blocks

### LINKED ISSUE #5818: qa_crc16_async_bb is flaky
The `qa_crc16_async_bb` test added in #5700 has already failed three times in CI:

* https://github.com/gnuradio/gnuradio/runs/6312845070?check_suite_focus=true
* https://github.com/gnuradio/gnuradio/runs/6315190144?check_suite_focus=true
* https://github.com/gnuradio/gnuradio/runs/6316453166?check_suite_focus=true

```
 FAIL: test_crc16 (__main__.qa_crc16_async_bb)
  File "$SRC_DIR/gr-digital/python/digital/qa_crc16_async_bb.py", line 48, in test_crc16
    self.assertEqual(dbg_check.num_messages(), 1)
AssertionError: 0 != 1
```

The failure looks similar to the one corrected in #5754, where messages don't make it through before the flow graph shuts down.

So far I haven't been able to reproduce a failure locally. If the test continues to be problematic, it may be best to remove it until the source of the problem is tracked down.

argilo:
## Description
The Async CRC16 and Async CRC32 blocks can perform misaligned loads, as was the case in #5807.

In addition, the Async CRC16 check fails randomly because it reads four bytes from memory instead of two, causing #5818.

I've fixed both problems by using `memcmp` (with the correct number of bytes) to perform the CRC comparison.

I also removed the unnecessary `crc16_async_bb_impl::process_crc` method to make the implementations of Async CRC16 and CRC32 nearly identical.

## Related Issue
* Fixes #5818.

* Async CRC16
* Async CRC32

## Testing Done
I verified that the `qa_crc16_async_bb` and `qa_crc32_async_bb` tests still pass with these changes in place.

## Checklist
- [ ] ~~I have updated [the documentation](https://wiki.gnuradio.org/index.php/Main_Page#Documentation) where necessary.~`

### DIFF
--- a/gr-digital/lib/crc16_async_bb_impl.cc
+++ b/gr-digital/lib/crc16_async_bb_impl.cc
@@ -77,7 +77,7 @@ void crc16_async_bb_impl::check(pmt::pmt_t msg)
     const uint8_t* bytes_in = pmt::u8vector_elements(bytes, pkt_len);
 
     crc = d_crc_ccitt_impl.compute(bytes_in, pkt_len - 2);
-    if (crc != *(unsigned int*)(bytes_in + pkt_len - 2)) { // Drop package
+    if (memcmp(&crc, bytes_in + pkt_len - 2, 2)) { // Drop package
         d_nfail++;
         return;
     }
--- a/gr-digital/lib/crc16_async_bb_impl.cc
+++ b/gr-digital/lib/crc16_async_bb_impl.cc
@@ -55,8 +55,7 @@ void crc16_async_bb_impl::calc(pmt::pmt_t msg)
     const uint8_t* bytes_in = pmt::u8vector_elements(bytes, pkt_len);
     std::vector<uint8_t> bytes_out(2 + pkt_len);
 
-    crc = process_crc(bytes_in, pkt_len);
-
+    crc = d_crc_ccitt_impl.compute(bytes_in, pkt_len);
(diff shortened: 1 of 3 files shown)


mormj (approved):

willcode:
@argilo I did a quick try at backporting to 3.9. Can't remember if this depends on something else - I can look it up but you probably already know if this is supposed to apply to 3.9.

argilo:
I guess the reason the patch doesn't apply cleanly is because #5700 also changed the CRC blocks and is not present in maint-3.9.

willcode:
If you have time, you could probably do it faster. We didn't pull back #5700 because it dropped some includes. I suppose we could backport #5700 but leave the boost includes in place.

@mormj opinion? #5700 applies cleanly. I'll put in a draft with #5700 and this PR together, with the boost includes left in, just to see if it works.

mormj:
backporting with the boost includes remaining seems like a reasonable path

willcode:
OK, I'll do that. @argilo I'll let you know if I run into trouble.

willcode:
@argilo ran into trouble trying to backport 5700, which then led to the "generic CRC" PR, which isn't going to work for 3.9. So, I just made the changes from `!=` to `memcpy` which was really the point of the PR.

Issue #5832, opened 2022-05-06
qa_dtv is flaky


argilo:
The `qa_dtv` test failed recently:

* https://github.com/gnuradio/gnuradio/runs/6276369380?check_suite_focus=true

I haven't yet been able to reproduce the failure locally.

The symptoms look very similar to the bug that was fixed in #5750. (In that instance, an uninitialized variable was packed into the transmitted bit stream.)

argilo:
It appears this test is flaky because the PAPR block ends up processing an OFDM symbol that is symmetric, and therefore has two samples of maximum magnitude. If I reverse the order of the loop below that searches for the maximum sample, then I get the same failure locally:

https://github.com/gnuradio/gnuradio/blob/3349d483a1f7e5db0eea042815d237810373857b/gr-dtv/lib/dvbt2/dvbt2_paprtr_cc_impl.cc#L686-L691

Replace the first line with `for (int n = papr_fft_size - 1; n >= 0; n--) {`, and the test fails.

argilo:
@drmpeg proposed to change the input file, in the hopes that the problematic OFDM symbol won't have samples of equal magnitude anymore. If that doesn't work, an alternative would be to remove the PAPR reduction block from the test.

argilo:
The problematic OFDM symbol is the second-last one, i.e. the 10th out of 11 OFDM symbols.

PR #5835, opened 2022-05-08
gr-dtv: Update DVB-T2 QA test to avoid spurious CI failures.

drmpeg:


argilo:
This fixes #5832.

argilo (approved):
Looks good! The qa_dtv test now passes even if I reverse the order of the loop in the PAPR block that searches for the sample with greatest magnitude.

argilo:
The CI failure is expected. It will be fixed by #5830.

argilo:
I also ran `qa_dtv` in a loop to verify that it is reliable, and verified that there are no valgrind errors.

PR #5845, opened 2022-05-12
blocks: Add symbolinterleaver block

jdemel:
---
name: blocks: Add symbolinterleaver block
about: Add a symbol interleaver to GNU Radio that takes in a vector and produces an interleaved vector.
title: ''
labels: 'blocks', 'interleaver', 'enhancement'
assignees: ''

---

# Pull Request Details
This commit adds a symbol interleaver to gr-blocks. The idea is to specify an interleaver sequence to be able to interleave and deinterleave vectors of constant size. For convenience, the deinterleaver sequence is always derived. Thus, it is not necessary to manually derive the corresponding deinterleaver sequence.

## Related Issue
The discussion for this PR started in #3637.
This block may improve on #4861 because this block can be parametrized to act as a matrix interleaver.

This is an entirely new block.

## Testing Done
At the moment, this is a draft and QA tests are missing. That'll change once the compiler/linker issues are fixed.

## Checklist


jdemel:
So almost all the CI past. Except the conda windows build. The error message says:
```
INTERNAL COMPILER ERROR in 'C:\PROGRA~1\MICROS~2\2022\ENTERP~1\VC\Tools\MSVC\1416~1.270\bin\HostX64\x64\cl.exe'
```

daniestevez:
Hi @jdemel, I think this block is great.

One thing I'm not so sure about is how to best handle the packed bytes option. Even though it would be possible not to handle it in this block, and get the same effect by combining the interleaver with the pack/unpack blocks, i guess that having this functionality included in the interleaver can be useful.

Since the `is_packed` parameter only applies to the `uint8_t` case, maybe it's helpful to hide this option in the GRC block parameters, unless `uint8_t` is selected. I also wonder if it's worth to have a slightly different constructor for the `uint8_t` case, where the `is_packed` parameter is present, and not have the `is_packed` parameter in the constructors of the other cases.

gr-blocks/lib/symbolinterleaver_xx_impl.cc
```diff
@@ -0,0 +1,178 @@
...
+      d_packer(std::make_unique<gr::blocks::kernel::pack_k_bits>(8)),
+      d_interleaver(std::make_unique<symbol_interleaving>(interleaver_indices))
+{
+    if (not std::is_same<T, uint8_t>::value) {
+        is_packed = false;
+        d_is_packed = false;
+    }
+    if (is_packed and d_interleaver->interleaver_length() % 8) {
+        throw std::invalid_argument(
+            fmt::format("Packed Interleaver requires 'interleaver_indices.size()' to be "
```
jaredd, line 72: I can see how a developer might appreciate the detail in this error message basically describing the test function. Could a simpler message be more broadly understood, e.g..

"Interleaver with packed byte input requires the length of interleaver_indices to be divisible by 8. <size> is not divisible by 8."

gr-blocks/lib/symbolinterleaver_xx_impl.cc
```diff
@@ -0,0 +1,178 @@
...
+                        "a multiple of 8 but got modulo8({}) = {}!",
+                        interleaver_indices.size(),
+                        interleaver_indices.size() % 8));
+    }
+    d_unpacked_original.resize(d_interleaver->interleaver_length());
+    d_unpacked_interleaved.resize(d_interleaver->interleaver_length());
+
+    gr::sync_block::set_output_multiple(is_packed
```
jaredd, line 80: Wouldn't this limit the potential size of the interleaver? I have had to implement interleavers exceeding 32K indices.

An unconstrained, pipelined streaming interleaver is a bit more logic, so for this case, perhaps have some check to on interleaver length to flag out-of-range values? Otherwise, I imagine the user will get some confusing message from the runtime.

jaredd:
Hello, thanks for including me in this review, I think it's my first requested review!

I also appreciate the generality of this block. Due to this generality you might consider the block name. SymbolInterleaver seems to suggest this is meant to be used post mapping. Maybe my vocabulary is just wrong. I would suggest something like BlockInterleaver since this clearly operates on defined blocks of data, be it symbols, bits, or otherwise.

For the future, I could imagine a couple extra GRC blocks based off this that fill in the interleaver indices for the user from a modified set of interface parameters. Like a matrix interleaver row, column, or a relative prime block interleaver with block size and step size.

argilo:
> Except the conda windows build.

It looks like it fails because Visual Studio does not support the `and` and `or` alternative tokens for logical operators. Changing those to `&&` and `||` should fix it.

jdemel:
> > Except the conda windows build.
> 
> It looks like it fails because Visual Studio does not support the `and` and `or` alternative tokens for logical operators. Changing those to `&&` and `||` should fix it.

This is surprising. [C++ standard Sec. 5.5 'Alternative tokens'](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2017/n4713.pdf) defines them. Thus, I'd expect them to be available in any C++ compiler nowadays.
How does an unknown keyword lead to an internal compiler error in MSVC?

gr-blocks/lib/symbolinterleaver_xx_impl.cc
jdemel: I have to check how GR handles `set_output_multiple` calls. I hope there's a check in there or some logic that increases the buffer size.
In any case, it'd be useful to communicate to a user why this block might fail.

I thought about supporting interleaving with larger indices then the output multiple. However, a general interleaver must assume that the last sample might be the first on the output and vice versa.
I assume that larger interleavers are designed to take this issue into account. If there's a hard limit, this info can be integrated into the docs.

jdemel:
> One thing I'm not so sure about is how to best handle the packed bytes option. Even though it would be possible not to handle it in this block, and get the same effect by combining the interleaver with the pack/unpack blocks, i guess that having this functionality included in the interleaver can be useful.

I imagine this block is often used right after the FEC block which typically supports packed and unpacked operations too.

> Since the `is_packed` parameter only applies to the `uint8_t` case, maybe it's helpful to hide this option in the GRC block parameters, unless `uint8_t` is selected. I also wonder if it's worth to have a slightly different constructor for the `uint8_t` case, where the `is_packed` parameter is present, and not have the `is_packed` parameter in the constructors of the other cases.

I think I'll re-order the ctor args and have `is_packed` default to `false`. That might make usage more intuitive too. And it'd be possible to hide this option in GRC more easily. 


> I also appreciate the generality of this block. Due to this generality you might consider the block name. SymbolInterleaver seems to suggest this is meant to be used post mapping. Maybe my vocabulary is just wrong. I would suggest something like BlockInterleaver since this clearly operates on defined blocks of data, be it symbols, bits, or otherwise.

I struggled with the name for a while. `Block Interleaver` might be a better name. [The original implementation name is `BitInterleaver`](https://github.com/ant-uni-bremen/gr-symbolmapping/blob/master/include/symbolmapping/bit_interleaver.h) and I want a better name for it because it is misleading. Just `Interleaver` is already used.
I thought about `Frame Interleaver` too because sometimes the FEC encoder output is called a frame, e.g. in the context of "Frame Error Rate". I'm open for suggestions.

> For the future, I could imagine a couple extra GRC blocks based off this that fill in the interleaver indices for the user from a modified set of interface parameters. Like a matrix interleaver row, column, or a relative prime block interleaver with block size and step size.

That's exactly the idea I had. I just want to make sure this block is available before special parametrizations are added.

gr-blocks/lib/symbolinterleaver_xx_impl.cc
jdemel: This is one of the places where `output_multiple` is used.

https://github.com/gnuradio/gnuradio/blob/eaeb3789401137fa3d7f3a30fdf782f07b331cb1/gnuradio-runtime/lib/block_executor.cc#L81

It throws in case that doesn't fit.

This is where buffers are allocated:
https://github.com/gnuradio/gnuradio/blob/eaeb3789401137fa3d7f3a30fdf782f07b331cb1/gnuradio-runtime/include/gnuradio/buffer.h#L55

And this is how a block allocates buffers. It checks `output_multiple`.
https://github.com/gnuradio/gnuradio/blob/eaeb3789401137fa3d7f3a30fdf782f07b331cb1/gnuradio-runtime/lib/block.cc#L448

jdemel: I added a test with
`nbits = int(2 ** 16) + int(2 ** 15)`
Thus, the interleaver handles `98304` items. Let's see how this works out. So far, I can confirm that it works on my machine.

jdemel:
I incorporated all review comments.
Most notably:
- `is_packed` is `False` by default and the third argument.
- `is_packed` is hidden in GRC if it doesn't apply to this type.
- Renamed `symbolinterleaver` to `blockinterleaver`.
- Fixed some issues with logical operations.

There's one comment I haven't full addressed. How to deal with large/huge interleavers. GR checks `output_multiple` during flowgraph construction and tries to allocate buffers accordingly. Surely, there's a limit. However, it seems like there's no hard limit.

gr-blocks/lib/blockinterleaver_xx_impl.cc
```diff
@@ -0,0 +1,178 @@
...
+    : gr::sync_block("blockinterleaver",
+                     gr::io_signature::make(1, 1, sizeof(T)),
+                     gr::io_signature::make(1, 1, sizeof(T))),
+      d_interleave_mode(interleave_mode),
+      d_is_packed(is_packed),
+      d_unpacker(std::make_unique<gr::blocks::kernel::unpack_k_bits>(8)),
+      d_packer(std::make_unique<gr::blocks::kernel::pack_k_bits>(8)),
+      d_interleaver(std::make_unique<block_interleaving>(interleaver_indices))
+{
+    if (not std::is_same<T, uint8_t>::value) {
```
argilo, line 66: This will hopefully fix the broken Windows build.
```suggestion
    if (!std::is_same<T, uint8_t>::value) {
```

jdemel: Thanks! I missed this one. I was searching for `or` and `and`.

jdemel:
@argilo Thanks for your hint! This fixed the last failing test.

I pushed another commit with a GRC example and some GRC block definition fixes.

gr-blocks/lib/symbolinterleaver_xx_impl.cc
jaredd: Cool, well, this was new to me. If nothing else, large interleaver tests can serve as bounds checks for individual installations.

gr-blocks/include/gnuradio/blocks/blockinterleaving.h
```diff
@@ -0,0 +1,57 @@
...
+ */
+
+#ifndef INCLUDED_GR_BLOCKS_BLOCKINTERLEAVING_H
+#define INCLUDED_GR_BLOCKS_BLOCKINTERLEAVING_H
+
+#include <gnuradio/blocks/api.h>
+
+
+namespace gr {
+namespace blocks {
```
mormj, line 18: should `blockinterleaving` be in the `kernel` namespace as well?

mormj, line 18: should `blockinterleaving` be in the `kernel` namespace as well?

jdemel: Well. It is a kernel. I didn't think of it. Sounds reasonable. Just to make sure

`gr::blocks::kernel::blockinterleaving` should be correct?

mormj: that seems like it would be consistent

mormj (approved):

PR #5846, opened 2022-05-14
runtime: Replace boost::endian with standard C++

argilo:
## Description
This is a continuation of the effort to remove Boost from GNU Radio.

In an effort to speed up PMT serialization (#3320, #3369), `boost::endian` was introduced. It seems that most of the performance gain was from other optimizations, though. Switching back to bit shifts for endianness conversion comes at a small performance cost (~25%) but I think it's worth it to be rid of Boost. If there's a way to improve performance further, I'd be happy to try it.

I tested performance using the Python script provided in #3320. Before:
```
	 U8 VECTOR
0.14483952522277832
128
	 U16 VECTOR
0.22285586595535278
128
	 F32 VECTOR
0.18961408734321594
128
	 C32 VECTOR
0.1912863850593567
128
```
After:
```
	 U8 VECTOR
0.1397770643234253
128
	 U16 VECTOR
0.24227461218833923
128
	 F32 VECTOR
0.23737838864326477
128
	 C32 VECTOR
0.2476167380809784
128
```

## Related Issue
* #3320
* #3369

PMT serialization & deserialization.

## Testing Done
I verified that all tests still pass, and that the test script in #3320 does not report any errors.

## Checklist

### DIFF
--- a/gnuradio-runtime/lib/pmt/pmt_serialize.cc
+++ b/gnuradio-runtime/lib/pmt/pmt_serialize.cc
@@ -15,7 +15,6 @@
 #include "pmt/pmt_serial_tags.h"
 #include "pmt_int.h"
 #include <pmt/pmt.h>
-#include <boost/endian/conversion.hpp>
 #include <limits>
 #include <vector>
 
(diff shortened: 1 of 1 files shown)


willcode:
Would something like this be useful? https://mklimenko.github.io/english/2018/08/22/robust-endian-swap/

argilo:
> Would something like this be useful?

Maybe, but it looks like it does an unconditional byte swap. We wouldn't want to do that on big-endian systems.

My approach was inspired by [The byte order fallacy](https://commandcenter.blogspot.com/2012/04/byte-order-fallacy.html) which argues that you should just use bit shifts and not try to determine the system's endianness.

gnuradio-runtime/lib/pmt/pmt_serialize.cc
```diff
@@ -239,10 +270,14 @@ static bool deserialize_untagged_u64_vector(std::vector<uint64_t>& data,
 {
     std::streambuf::traits_type::int_type t;
     data.resize(nitems);
-    t = sb.sgetn((char*)&data[0], nitems * sizeof(uint64_t));
+    std::vector<uint8_t> buf(nitems * sizeof(uint64_t));
+    t = sb.sgetn((char*)&buf[0], nitems * sizeof(uint64_t));
     sb.pubseekoff(sizeof(uint64_t) * nitems, std::ios_base::cur);
     for (size_t i = 0; i < nitems; i++) {
-        boost::endian::big_to_native_inplace(data[i]);
+        data[i] = ((uint64_t)buf[8 * i] << 56) | ((uint64_t)buf[8 * i + 1] << 48) |
```
mormj, line 277: For code cleanliness, should we make this bit of code an inline function?  it is reused several times.

mormj, line 277: For code cleanliness, should we make this bit of code an inline function?  it is reused several times.

argilo: Good idea. I'll factor out the repeated parts.

argilo:
I've rewritten this so that all the bit-shifting operations are encapsulated in inline helper functions.

mormj (approved):
Looks good.  Thanks!

mormj (approved):

PR #5850, opened 2022-05-15
runtime: Replace boost::function0 with std::function

argilo:
## Description
`boost::function0` is used in one place in gnuradio-runtime; I've replaced it with `std::function` here.

Thread groups, which are used by the thread-per-block scheduler.

## Testing Done
I verified that flow graphs still execute after this change.

## Checklist


willcode:
Do we think there could be external users of `thread_group` who would care about an API change (for backport)?

argilo:
I'd be surprised. It seems like something that GNU Radio copied over from Boost for its own use.

`std::function` appears to be compatible with `boost::function*` so I'd imagine any code that does use `thread_group` would still compile properly.

mormj (approved):
LGTM!!

PR #5852, opened 2022-05-15
runtime: Replace Boost with standard C++ in high resolution timer

argilo:
## Description
GNU Radio's high resolution timer uses `boost::posix_time` in a few places. I've replaced those with `std::chrono`.

* Performance counters
* QT GUI sinks

## Testing Done
To force the generic timer implementation to be used, I commented out the lines that check for the various platforms (`#if defined(linux)` etc.) to force `GNURADIO_HRT_USE_GENERIC_CLOCK` to be defined.

I checked the behaviour of the high resolution timer using the following Python code:
```
from gnuradio import gr
gr.high_res_timer_now()
gr.high_res_timer_now_perfmon()
gr.high_res_timer_tps()
gr.high_res_timer_epoch()
```
Also, I built GNU Radio with ControlPort and performance counters enabled (`-DENABLE_CTRLPORT_THRIFT=ON -DENABLE_PERFORMANCE_COUNTERS=ON`) and verified that I could still monitor the performance of a running flow graph with `gr-perf-monitorx`.

Finally, I verified that the QT GUI sinks work as before.

## Checklist

### DIFF
--- a/gnuradio-runtime/include/gnuradio/high_res_timer.h
+++ b/gnuradio-runtime/include/gnuradio/high_res_timer.h
@@ -12,7 +12,8 @@
 #define INCLUDED_GNURADIO_HIGH_RES_TIMER_H
 
 #include <gnuradio/api.h>
-#include <boost/date_time/posix_time/posix_time.hpp>
+#include <chrono>
+#include <ratio>
 
 ////////////////////////////////////////////////////////////////////////
 // Use architecture defines to determine the implementation
@@ -28,7 +29,7 @@
 #define GNURADIO_HRT_USE_CLOCK_GETTIME
 #include <ctime>
 #else
-#define GNURADIO_HRT_USE_MICROSEC_CLOCK
+#define GNURADIO_HRT_USE_GENERIC_CLOCK
 #endif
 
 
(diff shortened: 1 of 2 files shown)


gnuradio-runtime/include/gnuradio/high_res_timer.h
```diff
@@ -123,34 +124,30 @@ inline gr::high_res_timer_type gr::high_res_timer_tps(void)
...
 ////////////////////////////////////////////////////////////////////////
-#ifdef GNURADIO_HRT_USE_MICROSEC_CLOCK
+#ifdef GNURADIO_HRT_USE_GENERIC_CLOCK
 inline gr::high_res_timer_type gr::high_res_timer_now(void)
 {
-    static const boost::posix_time::ptime epoch(boost::posix_time::from_time_t(0));
-    return (boost::posix_time::microsec_clock::universal_time() - epoch).ticks();
+    return std::chrono::duration<gr::high_res_timer_type, std::nano>(
+               std::chrono::steady_clock::now().time_since_epoch())
+        .count();
```
argilo, line 132: I chose `steady_clock` because the timer is meant for performance measurements.

gnuradio-runtime/include/gnuradio/high_res_timer.h
```diff
@@ -123,34 +124,30 @@ inline gr::high_res_timer_type gr::high_res_timer_tps(void)
...
-        gr::high_res_timer_tps() /
-        double(boost::posix_time::time_duration::ticks_per_second());
-    boost::posix_time::time_duration utc =
-        boost::posix_time::microsec_clock::universal_time() -
-        boost::posix_time::from_time_t(0);
-    return gr::high_res_timer_now() - utc.ticks() * hrt_ticks_per_utc_ticks;
+    static const double ticks_per_second = gr::high_res_timer_tps();
+    const double seconds_since_epoch =
+        std::chrono::duration<double>(std::chrono::system_clock::now().time_since_epoch())
+            .count();
```
argilo, line 149: `system_clock` is used here because we need to measure time since the Unix epoch. The documentation states that prior to C++20 most implementations of `system_clock` used the Unix epoch, and as of C++20 implementations are required to do so. GNU Radio already depends on this behaviour in other places.

mormj (approved):

PR #5854, opened 2022-05-16
runtime: Replace Boost with standard C++ in rpcbufferedget

argilo:
## Description
rpcbufferedget uses Boost mutexes & condition variables in a few places. Those can easily be replaced with their standard C++ equivalents.

rpcbufferedget, which is used by ControlPort probes.

## Testing Done
None yet.

## Checklist

### DIFF
--- a/gnuradio-runtime/include/gnuradio/rpcbufferedget.h
+++ b/gnuradio-runtime/include/gnuradio/rpcbufferedget.h
@@ -11,8 +11,8 @@
 #ifndef RPCBUFFEREDGET_H
 #define RPCBUFFEREDGET_H
 
-#include <boost/thread/condition_variable.hpp>
-#include <boost/thread/mutex.hpp>
+#include <condition_variable>
+#include <mutex>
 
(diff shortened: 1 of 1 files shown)


gnuradio-runtime/include/gnuradio/rpcbufferedget.h
```diff
@@ -34,7 +34,7 @@ class rpcbufferedget
         if (!d_data_needed)
             return;
         {
-            boost::mutex::scoped_lock lock(d_buffer_lock);
+            std::unique_lock lock(d_buffer_lock);
```
mormj, line 37: I'm curious why not std::scoped_lock here

argilo: Oh, I suppose this should be a `scoped_lock` since it's not needed for a condition variable.

argilo: Fixed.

mormj (approved):

PR #5856, opened 2022-05-17
grc: Add parentheses on arithmetic expressions

dl1ksv:
# Pull Request Details

If a variable consists of an aritmetic expression and is used in another
expression the expression should be put in parentheses
## Description

See example in #5735

To avoid redundant parentheses check if the expression contains + or - operator
## Related Issue

This is a more general solution than #5735 and does not need changes in yml files and works for python and cpp code generation.

## Checklist

### DIFF
--- a/grc/core/params/param.py
+++ b/grc/core/params/param.py
@@ -308,6 +308,8 @@ class Param(Element):
                 self.evaluate()
             return '[' + value + ']' if self._lisitify_flag else value
         else:
+            if self.dtype in ('int', 'real') and ('+' in value or '-' in value):
+                value = '(' + value + ')'
             return value
 
     def get_opt(self, item):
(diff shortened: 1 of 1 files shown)


grc/core/params/param.py
```diff
@@ -308,6 +308,8 @@ def to_code(self):
                 self.evaluate()
             return '[' + value + ']' if self._lisitify_flag else value
         else:
+            if self.dtype in ('int', 'real') and ('+' in value or '-' in value):
```
mormj, line 311: what about other arithmetic operators?

dl1ksv: While trying to justify my choice I came across, that we need parentheses in some cases for * as

a = b/c*d != a=b/(c*d)

So it's better to add () for +,-,*,/

haakov (approved):
Looks good. We should probably add some tests for these cases too though.

PR #5858, opened 2022-05-18
runtime: Fix broken log format string in set_min_output_buffer

argilo:
## Description
If Minoutbuf is set to a value other than 0 in the "Advanced" tab of a block, then the following logging error occurs:
```
[*** LOG ERROR #0001 ***] [2022-05-18 16:38:00] [source] {invalid type specifier}
```
This happens because `unique_id()` is an integer, but `{:s}` was incorrectly specified in the format string.

This bug was introduced in #5599.

Any block that uses a non-default minimum output buffer size.

## Testing Done
After changing the format specifier to `{:d}`, all is well:
```
source :info: set_min_output_buffer on block source(1) to 1024
```

## Checklist

### DIFF
--- a/gnuradio-runtime/lib/block.cc
+++ b/gnuradio-runtime/lib/block.cc
@@ -365,7 +365,7 @@ long block::min_output_buffer(size_t i)
 void block::set_min_output_buffer(long min_output_buffer)
 {
     d_logger->info(
-        "set_min_output_buffer on block {:s} to {:d}", unique_id(), min_output_buffer);
+        "set_min_output_buffer on block {:d} to {:d}", unique_id(), min_output_buffer);
     for (int i = 0; i < output_signature()->max_streams(); i++) {
         set_min_output_buffer(i, min_output_buffer);
     }


drmpeg (approved):
Tested and works good.

gnuradio-runtime/lib/block.cc
```diff
@@ -365,7 +365,7 @@ long block::min_output_buffer(size_t i)
 void block::set_min_output_buffer(long min_output_buffer)
 {
     d_logger->info(
-        "set_min_output_buffer on block {:s} to {:d}", unique_id(), min_output_buffer);
+        "set_min_output_buffer on block {:d} to {:d}", unique_id(), min_output_buffer);
```
jdemel, line 368: Most of the fmtlib examples don't use type specifiers. Wouldn't it be reasonable to just use `{}`?

argilo: Most of the `boost::format` strings used explicit types, so I copied those over in the transition to fmtlib. C++20 is around the corner and will add type checking, so it's probably fine to leave the type specifiers in.

mormj (approved):

PR #5862, opened 2022-05-19
Fix constructor & destructor template syntax

argilo:
## Description
C++20 is stricter about template syntax, and does not allow redundant template parameter lists in constructors and destructors. I've fixed the two instances that occur in GNU Radio, and verified that compiling in C++20 mode succeeds.

* `buffer_type` class, used in buffer implementations
* Noise Source

## Testing Done
I verified that flow graphs still execute, and that the Noise Source block still works.

## Checklist

### DIFF
--- a/gnuradio-runtime/include/gnuradio/buffer_type.h
+++ b/gnuradio-runtime/include/gnuradio/buffer_type.h
@@ -111,7 +111,7 @@ public:
                                     buf_owner);
     }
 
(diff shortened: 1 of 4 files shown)


argilo:
@jdemel pointed out that `noise_source_impl` does not need to have its destructor explicitly defined, so I removed it.

jdemel (approved):
LGTM.

dsorber (approved):

PR #5889, opened 2022-05-25
c++ generation: Fix some template errors

dl1ksv:
# Pull Request Details

Fix cpp_templates in some filter_tap blocks and other blocks used  in the [filter taps example](https://github.com/gnuradio/gnuradio/blob/main/gr-filter/examples/filter_taps.grc)
## Description

Set cpp flag in low/high_pass_filter_taps block

Replace firdes. by filter::firdes:: in all *_filter_taps blocks

Switch to correct include file in filter_fir_filter_xxx block

Enabling enable_legend and setting spectrum width in qtgui_freq_sink_x block

## Testing Done

To test use the [filter tap example](https://github.com/gnuradio/gnuradio/blob/main/gr-filter/examples/filter_taps.grc)
Set the output language to C++
Disable the len_taps block. This block is not needed for the flowgraph and prevents generating c++ code

Compare the output of the python and c++ generated flowgraph
## Checklist

### DIFF
--- a/gr-filter/grc/variable_band_pass_filter_taps.block.yml
+++ b/gr-filter/grc/variable_band_pass_filter_taps.block.yml
@@ -36,6 +36,9 @@ parameters:
     label: Beta
     dtype: float
     default: '6.76'
+-   id: value
+    dtype: float_vector
+    hide: all
 value: ${ getattr(firdes, str(type))(gain, samp_rate, low_cutoff_freq, high_cutoff_freq,
     width, win, beta) }
 
@@ -51,8 +54,8 @@ cpp_templates:
     includes: ['#include <gnuradio/filter/firdes.h>']
     declarations: 'filter::firdes::sptr ${id};'
     var_make: |-
-        this->${id} = ${id} = firdes.${type}(${gain}, ${samp_rate}, ${low_cutoff_freq}, \
-        ${high_cutoff_freq}, ${width}, ${win}, ${beta});
+        ${id} = filter::firdes::${type}(${gain}, ${samp_rate}, ${low_cutoff_freq}, \
+        ${high_cutoff_freq}, ${width}, ${win.replace('window.','fft::window::')}, ${beta});
     link: ['gnuradio::gnuradio-filter']
 
 documentation: |-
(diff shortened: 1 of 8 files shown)


mormj:
Tested this as described.  LGTM!

mormj (approved):

gr-filter/grc/variable_band_pass_filter_taps.block.yml
```diff
@@ -36,6 +36,9 @@ parameters:
     label: Beta
     dtype: float
     default: '6.76'
+-   id: value
```
mormj, line 39: What is this additional parameter used for?

dl1ksv: Without this entry cpp generation fails with:

  File "/usr/local/gnuradio/lib64/python3.10/site-packages/gnuradio/grc/core/generator/cpp_top_block.py", line 86, in write
    self._variable_types()
  File "/usr/local/gnuradio/lib64/python3.10/site-packages/gnuradio/grc/core/generator/cpp_top_block.py", line 310, in _variable_types
    if var.params['value'].dtype != 'raw':
KeyError: 'value'

dl1ksv: I checked some other blocks that claimed to be a variable block and to be able to generate cpp code.
If there is no value entry cpp code generation fails. Just try the Default Header Format Obj.
We are talking about blocks of 'type variable'. A variable represents a value, so the value parameter is not a dummy parameter, but more important for the cpp generation, it's associated with a type, and this type is used for cpp code generation. This is important. Take the above mentioned filter tap example and enable the len_taps block. Then you can't generate cpp code anymore, as deriviation of the type from the expression len(lp_taps) fails for cpp, though the error message is misleading.
So I believe, it's the right way to add the value parameter.
At the moment there are lot of cpp templates for variable generation, that fail, but could be fixed in this way. But apparently they are not used.

mormj: Ok, I see those now - I guess it makes sense as the way it is worked around.

mormj (approved):

PR #5902, opened 2022-05-27
ci: Add Fedora 36, remove Fedora 34

willcode:
## Description
Fedora 36 is current. Change CI to test Fedora 35/36 instead of 34/35.

## Checklist

### DIFF
--- a/.github/workflows/make-test.yml
+++ b/.github/workflows/make-test.yml
@@ -64,16 +64,19 @@ jobs:
             cxxflags: -Werror -Wno-error=invalid-pch
             ctest_args: '-E "qa_cpp_py_binding|qa_cpp_py_binding_set|qa_ctrlport_probes|qa_qtgui|metainfo_test"'
             ldpath:
-          - distro: 'Fedora 34'
-            containerid: 'gnuradio/ci:fedora-34-3.9'
-            cxxflags: -Werror -Wno-error=invalid-pch -Wno-error=deprecated-declarations
-            ctest_args: '-E "qa_cpp_py_binding|qa_cpp_py_binding_set|qa_ctrlport_probes|qa_qtgui"'
-            ldpath: /usr/local/lib64/
+            pypath:
           - distro: 'Fedora 35'
             containerid: 'gnuradio/ci:fedora-35-3.9'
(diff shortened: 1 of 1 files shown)


marcusmueller (commented):
To me this looks sane, and I don't remember a Fedora 36 changeset entry that said the PYTHONPATH handling differed. I'm honestly a bit surprised pre-36 finds the module, as I've never actually saw my module try /usr/local by default. /usr/local/lib[64] also doesn't appear in the standard ld directories on fedora, so I'm now honestly a bit confused why this used to work? Maybe the docker containers are different than desktop installations? We could see which directories it tries by inserting a `-vv` into the `python3 -c "import gnuradio…`  call.

Anyways, we don't seem to find the installed module. So, hm. Maybe instead of doing the slightly inelegant needs-root, but still not part of the default search space, we could modify our cmake line to contain `DCMAKE_INSTALL_PREFIX=$(python3 -m site --user-base)`. We'd still need to point `LD_LIBRARY_PATH` to `$(python3 -m site --user-base)/lib64`

willcode:
@marcusmueller are you referring to the ldpath? I should have filled in the F35 entry - it was the F34 one that was blank. I don't know exactly what effect this has (apparently none that CI cares about).

Are you also talking about this install location for Python modules on F36? I noticed that they go into local/lib64 while the shared objects go into lib64. Had to update some env vars as a result.

willcode:
This change adds `sysconfig.get_path('platlib')` to the `PYTHONPATH`.

Users are going to be confused. One possibility is that we use `sysconfig.get_path('platstdlib')` (vs `platlib`) in `cmake/Modules/GrPython.cmake`. Someone with cross-platform cmake powers would need to think about that.

willcode:
Previous idea didn't work out. Instead, the PYTHONPATH is set explicitly to `/usr/local/`+`pypath`+`lib64/pythonX.Y/site-packages` where X and Y are pulled from `sys.version_info` and `pypath` is set to `local/` where required. It wouldn't hurt anything to set the extra path for all distros, I suppose.

willcode:
Replaced by #5905

PR #5903, opened 2022-05-29
Add a test case for float_to_int clipping input values slightly greater

michael-roe:
than INT_MAX to INT_MAX.

This adds a test case for float_to_int clipping an input value greater than MAX_INT. It is a regression test for a bug in volk_32f_s32f_convert_32i for which the current work-around is to not use that VOLK kernel.

### DIFF
--- a/gr-blocks/python/blocks/qa_type_conversions.py
+++ b/gr-blocks/python/blocks/qa_type_conversions.py
@@ -193,8 +193,8 @@ class test_type_conversions(gr_unittest.TestCase):
         self.assertFloatTuplesAlmostEqual(expected_data, dst.data())
 
     def test_float_to_int_identity(self):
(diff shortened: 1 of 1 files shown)


marcusmueller (approved):

marcusmueller:
small request: Our commit message format is `component: short description`, i.e., I'd ask you to rewrite your commit subject to 

```
blocks: extend float_to_int test to values slightly > INT_MAX

These values should correctly be clipped to INT_MAX.

Signed-off-by: Michael Roe <youremail@youremailhost.com>
```

michael-roe:
Ok, I'll send a revised PR tomorrow.

marcusmueller:
Thank you so much! Note: it's totally OK to just `git commit --amend` and then `git push --force` to fix this :)

PR #5905, opened 2022-05-30
CI: Add Fedora 36, remove Fedora 34, second try

willcode:
# Pull Request Details
## Description
CI: Add Fedora 36, remove Fedora 34.

In addition, Fedora 36 and Ubuntu 22.04 have changed the Python install
scheme to put modules under <prefix>/local. Our default install location
of /usr/local results in files being install under /usr/local/local. The
change to GrPython.cmake puts default installs back in /usr/local.
(Suggested by dl1ksv).

CI.

## Checklist

### DIFF
--- a/cmake/Modules/GrPython.cmake
+++ b/cmake/Modules/GrPython.cmake
@@ -165,7 +165,7 @@ if not install_dir:
     else:
         scheme = 'posix_prefix'
     install_dir = sysconfig.get_path('platlib', scheme)
-    prefix = sysconfig.get_config_var('prefix')
+    prefix = sysconfig.get_path('data')
 
 #strip the prefix to return a relative path
 print(os.path.relpath(install_dir, prefix))"
--- a/.github/workflows/make-test.yml
+++ b/.github/workflows/make-test.yml
@@ -64,13 +64,13 @@ jobs:
             cxxflags: -Werror -Wno-error=invalid-pch
             ctest_args: '-E "qa_cpp_py_binding|qa_cpp_py_binding_set|qa_ctrlport_probes|qa_qtgui|metainfo_test"'
             ldpath:
-          - distro: 'Fedora 34'
-            containerid: 'gnuradio/ci:fedora-34-3.9'
+          - distro: 'Fedora 35'
+            containerid: 'gnuradio/ci:fedora-35-3.9'
             cxxflags: -Werror -Wno-error=invalid-pch -Wno-error=deprecated-declarations
             ctest_args: '-E "qa_cpp_py_binding|qa_cpp_py_binding_set|qa_ctrlport_probes|qa_qtgui"'
             ldpath: /usr/local/lib64/
-          - distro: 'Fedora 35'
-            containerid: 'gnuradio/ci:fedora-35-3.9'
+          - distro: 'Fedora 36'
+            containerid: 'gnuradio/ci:fedora-36-3.9'
             cxxflags: -Werror -Wno-error=invalid-pch -Wno-error=deprecated-declarations
             ctest_args: '-E "qa_cpp_py_binding|qa_cpp_py_binding_set|qa_ctrlport_probes|qa_qtgui"'
             ldpath: /usr/local/lib64/


cmake/Modules/GrPython.cmake
```diff
@@ -165,7 +165,7 @@ if not install_dir:
     else:
         scheme = 'posix_prefix'
     install_dir = sysconfig.get_path('platlib', scheme)
-    prefix = sysconfig.get_config_var('prefix')
+    prefix = sysconfig.get_path('data')
```
marcusmueller, line 168: probably the righter thing! And since it doesn't break on the debianoids, this seems fine. Let's wait for the conda builds (but these we could fix ourselves, no losses), but this: really LGTM.

marcusmueller (approved):

marcusmueller:
Btw, the new build gets an xxx-unknown version string, due to recent git becoming more careful about repos not owned by the current user. 

```
-- Extracting version information from git describe...
fatal: unsafe repository ('/__w/gnuradio/gnuradio' is owned by someone else)
To add an exception for this directory, call:

	git config --global --add safe.directory /__w/gnuradio/gnuradio
```

dl1ksv (approved):
GR_PYTHON_DIR works as it should.

ryanvolz (approved):

mormj (approved):

willcode:
@dl1ksv This fix was reported not to work on 22.04 with a _custom_ install location. From @drmpeg:

```
With cmake -DCMAKE_INSTALL_PREFIX=/opt/gnuradio-3.11.0git ../
It's installing into /opt/gnuradio-3.11.0git/../lib/python3.10/site-packages/
In other words, /opt/lib
```

ryanvolz:
@willcode I think the thing to try is something like this:
```
scheme = 'venv' if 'venv' in sysconfig.get_scheme_names() else None
install_dir = sysconfig.get_path('platlib', scheme=scheme, vars={'base': prefix})
```
and skip the whole relative path thing or still do it but keep `prefix` set to `CMAKE_INSTALL_DIR` from above and don't overwrite it with a query to sysconfig.

The `venv` scheme should cover the latest Fedora and possibly Debian/Ubuntu that was causing issues in the first place, see https://docs.fedoraproject.org/en-US/fedora/latest/release-notes/developers/Development_Python/#_paths_for_bootstrapping_python_virtual_environments. Passing the prefix as the `base` var should mean that the relative path shenanigans are unnecessary.

willcode:
Since I only speak "tourist" cmake, would it be
```
if(NOT DEFINED GR_PYTHON_DIR)
execute_process(
    COMMAND ${PYTHON_EXECUTABLE} -c "import os
import sysconfig
import site

install_dir = None
prefix = '${CMAKE_INSTALL_PREFIX}'

#use sites when the prefix is already recognized
try:
  paths = [p for p in site.getsitepackages() if p.startswith(prefix)]
  if len(paths) == 1: install_dir = paths[0]
except AttributeError: pass

if not install_dir:
    scheme = 'venv' if 'venv' in sysconfig.get_scheme_names() else None
    install_dir = sysconfig.get_path('platlib', scheme=scheme, vars={'base': prefix})
print(install_dir)"
    OUTPUT_STRIP_TRAILING_WHITESPACE
    OUTPUT_VARIABLE GR_PYTHON_DIR
)
endif()
```

willcode:
`scheme = None` is not valid for `sysconfig.get_path()`, at least on F35 where there is no 'venv'. I think the 'nt' or 'posix_prefix' scheme selection is still needed.

dl1ksv:
On Fedora 36 it leads to 
/usr/lib64/python3.10/site-packages
and this is not what we need.

ryanvolz:
> On Fedora 36 it leads to /usr/lib64/python3.10/site-packages and this is not what we need.

What should it be?

dl1ksv:
/opt/gnuradio-3.11.0git/lib64/python3.10/site-packages/

willcode:
The CI run for F36 installed Python files in /usr/local/lib64/python3.10/site-packages, which is what we want as a default.

I just did a local build with CMAKE_INSTALL_PREFIX=/tmp/gr. It installed most files in /tmp/gr/..., but tried to install Python files under /usr/lib64/python3.10/... which is not good.

@dl1ksv Why do we want to switch to /opt?

dl1ksv:
Not to /opt but to /opt/gnuradio-3.11.0git, as the INSTALL_PREFIX was set to /opt/gnuradio-3.11.0git

willcode:
Ah, OK. Same result I got for a custom prefix, then.

dl1ksv:
I try to understand , what happens. But as I have no access to ubuntu 22.04 , can someone post the output of the following python code:

import sysconfig
sysconfig.get_path('platlib', 'posix_prefix')
sysconfig.get_path('data')

On my fedora 36 I get

'/usr/local/lib64/python3.10/site-packages'
'/usr/local'

mormj:
Ubuntu 22.04 gives: 
'/usr/lib/python3.10/site-packages'
'/usr/local'

changing to sysconfig.get_path('data','posix_prefix') 
gives 
/usr

mormj:
I tried in #5981 to let it use the same scheme for both paths, and it works for custom prefix

ryanvolz:
Try also setting `platbase` to the prefix in the `vars` dictionary, as it seems the `platlib` value is based on that and not `base`.

ryanvolz:
> I tried in #5981 to let it use the same scheme for both paths, and it works for custom prefix

I guess that was what was missing to make that approach work. I'm still worried that making the `install_dir` into a relative path is a brittle approach that could bite again in the future, so my preference would be trying to get #5979 to work still.

willcode:
My local build installed to the correct place with the addition of `platbase` so I'll push that update to #5979.

dl1ksv:
@mormj That works on fedora 36, too. Just tested.

PR #5912, opened 2022-05-31
pdu: Remove --no-as-needed linker option

argilo:
## Description
It appears this workaround was copied from the gr-wavelet module, but it's not needed here because gr-pdu doesn't depend on GSL. Removing it reduces the number of libraries that gr-pdu links to.

* gr-pdu cmake build

## Testing Done
None. If CI is happy, this should be good.

## Checklist

### DIFF
--- a/gr-pdu/lib/CMakeLists.txt
+++ b/gr-pdu/lib/CMakeLists.txt
@@ -53,11 +53,6 @@ target_include_directories(gnuradio-pdu
   )
 
 
-# we need -no-as-needed or else -lgslcblas gets stripped out on newer version of gcc
-if(CMAKE_COMPILER_IS_GNUCC AND NOT APPLE)
-    SET_TARGET_PROPERTIES(gnuradio-pdu PROPERTIES LINK_FLAGS "-Wl,--no-as-needed")
-endif()
-
 if(BUILD_SHARED_LIBS)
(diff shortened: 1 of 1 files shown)


argilo:
Diffing the `ldd` output shows that this change prevents the following extraneous libraries from being linked:

```
libboost_atomic.so.1.74.0
libboost_regex.so.1.74.0
libboost_system.so.1.74.0
libfftw3f.so.3
libfftw3f_threads.so.3
libFLAC.so.8
libgmpxx.so.4
libgnuradio-blocks.so.3.11.0git
libgnuradio-fft.so.3.11.0git
libgnuradio-filter.so.3.11.0git
libicudata.so.70
libicui18n.so.70
libicuuc.so.70
libogg.so.0
libopus.so.0
libsndfile.so.1
libvorbisenc.so.2
libvorbis.so.0
```

argilo:
I wonder whether the workaround is still needed for gr-wavelet. It similarly links to tons of things that it doesn't need to.

argilo:
I'm experimenting with gr-wavelet in #5913.

The fact that so many unneeded libraries are passed to the linker may be a sign that dependencies should be cleaned up in the cmake files.

ryanvolz (approved):
This one seems straightforward, since GSL is not needed at all, and I very much approve of reducing overlinking.

mbr0wn (approved):

PR #5913, opened 2022-05-31
wavelet: Remove --no-as-needed linker option

argilo:
This option causes a lot of extraneous libraries to be linked. It may no longer be needed.

It was added as a workaround in https://github.com/gnuradio/gnuradio/commit/c0b3ce38db0eaaca05fe2f45827fcf6c9184b72b.

I'm opening this draft pull request to see whether CI passes without it.

### DIFF
--- a/gr-wavelet/lib/CMakeLists.txt
+++ b/gr-wavelet/lib/CMakeLists.txt
@@ -44,11 +44,6 @@ target_include_directories(gnuradio-wavelet
   )
 
 
-# we need -no-as-needed or else -lgslcblas gets stripped out on newer version of gcc
-if(CMAKE_COMPILER_IS_GNUCC AND NOT APPLE)
-    SET_TARGET_PROPERTIES(gnuradio-wavelet PROPERTIES LINK_FLAGS "-Wl,--no-as-needed")
-endif()
-
 if(BUILD_SHARED_LIBS)
   GR_LIBRARY_FOO(gnuradio-wavelet)
 endif()


argilo:
When I tested this on Ubuntu 22.04, `ldd` showed that `libgslcblas.so.0` was still linked.

argilo:
It's possible the issue was fixed (at least on Debian-based systems) in this Debian patch: https://sources.debian.org/patches/gsl/2.6+dfsg-2/gsl-cblas-linkage/

ryanvolz (commented):
I'll admit I don't quite understand why GSL is a pain with its linking, but I do know that https://github.com/gnuradio/gnuradio/blob/main/.packaging/conda_recipe/meta.yaml#L61 was at one time necessary. With this change, the conda linux-64 build logs show that `libcblas` is not linked by anything and is an extraneous run dependency, which seems contrary to what @argilo says happens on Debian which has the GSL patch whereas conda-forge doesn't. I'd be curious about what happens if https://github.com/gnuradio/gnuradio/blob/main/.packaging/conda_recipe/meta.yaml#L61 is removed, since it looks like it can be following this change.

My memory is hazy, but I think there was a time after 3.9 where I was able to move `libcblas` from a dependency for all platforms to just a dependency on Linux. So I think something changed where we don't actually need to link `libcblas` anymore, but it was still linking on Linux because of the lines that this PR removes. If https://github.com/gnuradio/gnuradio/blob/main/.packaging/conda_recipe/meta.yaml#L61 is removed and everything is still green, then I think that would confirm the theory and we could go ahead and merge the change.

mormj (approved):

PR #5917, opened 2022-06-02
filter: Remove unused dependency

argilo:
## Description
The gr-filter module has gnuradio-blocks in its `target_link_libraries` even though it doesn't use this module. Removing it will prevent extraneous libraries from being linked.

CMake build for the gr-filter module.

## Testing Done
I verified that GNU Radio still builds correctly, the test suite passes, and Gqrx runs without issue.

## Checklist


argilo:
Oops, I missed that the frequency xlating fir filter depends on the rotator. Strange that this built on my machine.

PR #5952, opened 2022-06-10
qtgui: Add missing <deque> include to spectrumdisplayform.h

ryanvolz:
# Pull Request Details

## Description

This adds a missing `<deque>` include in a header where `std::deque` is used. It seems that this is not usually a problem because of transitive includes, but I just encountered a Windows build failure over at conda-forge for v3.10.3.0-rc1 that this solves. I'm not sure what combination of differences made this necessary there but not on other platforms or on the CI here (maybe not using precompiled headers over there?).

qtgui

## Testing Done

Build failed on Windows on conda-forge before this, and now it doesn't.

## Checklist


ryanvolz:
It would be nice to get this into 3.10 before the release so I don't have to carry a patch.

Issue #5891, opened 2022-05-25
Remove Boost dependency


argilo:
One of the goals of GNU Radio 4.0 is to remove Boost as a dependency. After the open Boost removal PRs (#5706, #5846, #5852, #5853, #5888, #5890) are merged, there will only be a small amount of Boost left. The table below summarizes the remaining work.

Boost component | Used in | LOC | Proposed replacement | Notes
------------------------|------------|-------:|--------------------------------|---------
Boost.Test | C++ test cases | 921 | ???
Boost.Asio | gr-network, rpcserver_booter_thrift | 91 | [Stand-alone ASIO library](https://think-async.com/Asio/)
Boost.Any | PMT | 27 | `std::any` | See #5218, #5311
Boost.Thread | Thread-per-block scheduler | 26 | `std::thread` | Needs refactoring to remove `interrupt()`
Boost.Program_options | `gr::prefs`, gnuradio-config-info, UHD tags_demo, GRC C++ generation | 4 | ???
`file_lock` | gr-fft | 2 | ???
`this_line_location` | `gr::prefix()` | 1 | Copy code from Boost? ([Source](https://github.com/boostorg/dll/blob/develop/include/boost/dll/runtime_symbol_info.hpp))
`math::constants::pi` | GRC C++ generation | 1 | `3.14159265358979323846` or `std::numbers::pi` (C++20)

argilo:
Thanks for the suggestions! I've added them to the table.

> Boost.program_options --> cli11 works well and is a minimal dependency

This will take care of the CLI parts, but `gr::prefs` also uses program_options to parse config files (which use INI format). We'll need a replacement for that as well.

marcusmueller:
I think the most intrusive change is going to be the boost::thread replacement. Had a go at that on two occasions, and I think it leaked into the API. So, yeah, touching the core tpb thread body would be necessary, which might not be much fun (as it's possible to introduce new shutdown bugs), but maybe it's easy, in the end an interrupt point is just a semaphore "should this thread still be running? If not, throw an exception!", and maybe just replacing that with a check against one of the already existing "done" state indicators would actually untangle the code, and force `sched_tpb::stop()` to actually signal shutdown to all threads instead of generally interrupting them, which might or might not wake them, and make them finish their job.

mbr0wn:
I see no value in replacing Boost.Test and Boost.Program_options.

argilo:
[...]

PR #5957, opened 2022-06-16
gr-network: port to standalone asio

japm48:
# Pull Request Details

Port `gr-network` to standalone asio from `boost::asio`.

## Description

This is part of the de-boostification effort. Hopefully this will still work without changes with the standard C++23?/TS? library.
Most of this is naive renaming (remove `boost::` namespace), but there's also some API changes needed (mainly [io_service->io_context](https://github.com/boostorg/asio/issues/110)).
Boost is still required here, as `boost::circular_buffer` (unrelated to Asio) is still used.

## Related Issue

#5891

## Testing Done

QA tests are passing on my computer, examples are also working.

## Checklist


.packaging/debian/control
```diff
@@ -6,9 +6,10 @@ Build-Depends: cmake,
                debhelper (>= 12~),
                dh-python,
                doxygen,
-	            gir1.2-gtk-3.0,
-	            gir1.2-pango-1.0,
-	            graphviz,
+               gir1.2-gtk-3.0,
```
japm48, line 9: Note: there were some scattered tabs over there. I can revert if you prefer...

japm48:
asio is still not in the GR docker images, so most CI tests are meaningless.
Should I post a PR to add it in [gnuradio-docker](https://github.com/gnuradio/gnuradio-docker)?

willcode:
Since I/O is blocking (otherwise `work()` would spin), and blocks are already threaded, why are we doing ASIO in the first place? I think there was some talk about this a while ago. Why not remove the boost dependency by removing ASIO in this case? It's more work to make that change, obviously.

Is there a path to removing boost::circular_buffer if we do this? Would a GR circular buffer work?

japm48:
@willcode Unless I'm mistaken some high-level networking library is still desired; you probably don't want to implement this in raw sockets, with TCP/UDP and IPV4/V6, and dealing with Winsock idiosyncrasies...
Apart from that, it is true that there are two schedulers (GR's and Asio, e.g. see [this comment](https://github.com/gnuradio/gnuradio/blob/d25a59cbb61992237b8d46db85be9c40ed5adf9e/gr-network/lib/udp_sink_impl.h#L45)), so maybe using asio's sync API instead of async makes sense here?

> I think there was some talk about this a while ago.

Is there a record of that?

> Is there a path to removing boost::circular_buffer if we do this? Would a GR circular buffer work?

I'm not really sure if the GR one is suitable here. If it is not, replicating some parts from the [boost implementation](https://github.com/boostorg/circular_buffer/tree/develop/include/boost) should not be too difficult IMO.

willcode:
There was some discussion on matrix (Development) on 6/25 between @mormj, @mbr0wn and @marcusmueller. Raising this same question and saying that the header-only asio library was probably better than using boost.

You're right that we would want a platform-independent layer. I think if we used asio (or some other library) to that end, and got rid of any unnecessary use of async, it would be good. Async is great if you want to avoid the thread-per-connection model for whatever reason. I think the only place we have multiple clients/servers is when we use ZMQ.

Just as a note (not an argument either way), using header-only libraries pretty much eliminates the possibility of ABI-compatible releases. It's already almost impossible to guarantee ABI compatibilty with GR, so this isn't really a factor. And we're looking at stuff like jlohmann-json in the future already.

mormj:
This seems like a logical replacement of boost::asio given that standalone asio is well packaged:

A couple of things we need to consider:
1) Min version checking for ASIO dependency - probably would lock it to whatever is in Ubuntu 20.04, which looks like 1.12. 
2) Need some testing with Thrift Ctrlport

japm48:
> 1. Min version checking for ASIO dependency - probably would lock it to whatever is in Ubuntu 20.04, which looks like 1.12.

Yes, that's what I did (see `build_guide.dox.in`)

> 2. Need some testing with Thrift Ctrlport

I think it already compiles and works (on my computer™), but I'm not sure how to add a test for that.

mormj:
@japm48 - I hadn't even remembered about build_guide, so thanks for adding there.  Just pushed a commit to enforce a min ASIO version through cmake and be a bit more draconian - looks like the CI workers need to be updated with standalone asio for CI to pass now

As far as CtrlPort testing - manual testing is probably the best we can do.  I don't see qa tests for ctrlport really doing anything significant.  Might be an improvement in the future, but for now having a second set of eyes on ctrlport functionality would be good.

mormj:
Added https://github.com/gnuradio/gnuradio-docker/pull/29 which will have to propagate to dockerhub for this to pass CI

mbr0wn:
> Just as a note (not an argument either way), using header-only libraries pretty much eliminates the possibility of ABI-compatible releases. It's already almost impossible to guarantee ABI compatibilty with GR, so this isn't really a factor. And we're looking at stuff like jlohmann-json in the future already.

Everyone always glares at me when I suggest this, but we can always vendor in a version of ASIO if we need to stick to specific versions. That said, aren't there some compiler/linker switches we can throw to hide symbols obtained from ASIO? And then wouldn't that maintain ABI compat?

mbr0wn (approved):
I haven't run this locally, but the code changes are all 100% renaming of `boost::asio` to `asio`, with some minor, correct updates (like using `io_context` instead of `io_service`).

mbr0wn:
@japm48 Since we're currently unsure if we want to kill Boost.Asio in GR 3, I've pulled out most of your PR #6260.

mormj:
@japm48 - thank you for getting everything prepared for standalone ASIO - as mentioned by @mbr0wn, this is incorporated in #6260 so closing

PR #5975, opened 2022-06-29
CI: make F36 builder (gcc12) use 0xFE as memory init pattern

marcusmueller:
New feature in GCC12, helping find reliance on coincidental initialization

---

# Pull Request Details

## Description

The hopes here are that we'll find places were we accidentally rely on memory being initialized to 0.

Build-wide

## Testing Done

None, if all unit tests continue to work, we feel happy.

## Checklist

### DIFF
--- a/.github/workflows/make-test.yml
+++ b/.github/workflows/make-test.yml
@@ -69,9 +69,9 @@ jobs:
             cxxflags: -Werror -Wno-error=invalid-pch -Wno-error=deprecated-declarations
             ctest_args: '-E "qa_cpp_py_binding|qa_cpp_py_binding_set|qa_ctrlport_probes|qa_qtgui"'
             ldpath: /usr/local/lib64/
-          - distro: 'Fedora 36'
+          - distro: 'Fedora 36 (with 0xFE memory initialization)'
             containerid: 'gnuradio/ci:fedora-36-3.9'
-            cxxflags: -Werror -Wno-error=invalid-pch -Wno-error=deprecated-declarations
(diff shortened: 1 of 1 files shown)


.github/workflows/make-test.yml
```diff
@@ -96,7 +96,10 @@ jobs:
     - name: CMake
       env:
         CXXFLAGS: ${{ matrix.cxxflags }}
-      run: 'cd /build && cmake ${GITHUB_WORKSPACE} -DENABLE_DOXYGEN=OFF'
+        EXTRA_CMAKE_FLAGS: ${{ matrix.extra_cmake_flags }}
+      run: 'cd /build && cmake ${GITHUB_WORKSPACE} -DENABLE_DOXYGEN=OFF ${EXTRA_CMAKE_FLAGS}'
```
argilo, line 100: These lines look to be left over from a previous experiment, since there's no `extra_cmake_flags` in the matrix anymore.

.github/workflows/make-test.yml
```diff
@@ -96,7 +96,10 @@ jobs:
...
+    - name: Dump compile commands
+      run: 'cd /build && cat compile_commands.json'
```
argilo, line 102: Should this be removed now that we know the change works correctly?

.github/workflows/make-test.yml
marcusmueller: though it's probably a handy tool, right, this shouldn't be here.

argilo (approved):

willcode:
This would probably be safe to backport to maint-3.10, but since it will already turn up problems on main, there's no need.

PR #5981, opened 2022-06-30
cmake: use same scheme for py rel path calculation

mormj:
Another attempt at making the python relative path calculation consistent between distros


mormj:
Closing in favor of #5979

Issue #5959, opened 2022-06-17
analog/fastnoise_source constructor argument `seed` type inconsistent with docstring


bgottula:
### What happened?

The [docstring](https://github.com/gnuradio/gnuradio/blob/main/gr-analog/include/gnuradio/analog/fastnoise_source.h#L46-L48) for analog/fastnoise_source `seed` argument states:

> seed for random generators. Note that for uniform and Gaussian distributions, this should be a negative number.

This always seemed a strange requirement, but the type was changed in a recent commit from `long` to `uint64_t` so clearly a negative value is not even possible. I'm guessing the docstring is probably wrong, but I don't know the history of why a negative number was recommended previously.

Commit where the argument type was changed: https://github.com/gnuradio/gnuradio/commit/4dd02dac2fcaf45f398c074735a7359330f78d77

Related issue: https://github.com/gnuradio/gnuradio/issues/1559

I also note that the type of the `seed` parameter for the analog/noise_source block is still a `long` and it has the same docstring: https://github.com/gnuradio/gnuradio/blob/main/gr-analog/include/gnuradio/analog/noise_source.h#L44-L48

### System Information

OS: Ubuntu 20.04
GR Installation Method: apt via gnuradio-releases PPA

### GNU Radio Version

3.10 (maint-3.10)

### Specific Version

3.10.2.0

### Steps to Reproduce the Problem

Use negative seed value as documentation suggests when instantiating fastnoise_source. (This produced no error in GNU Radio 3.9)

### Relevant log output

```shell
In [12]: a = analog.fastnoise_source_c(analog.GR_GAUSSIAN, 1, -1, 100000)
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Input In [12], in <cell line: 1>()
----> 1 a = analog.fastnoise_source_c(analog.GR_GAUSSIAN, 1, -1, 100000)

TypeError: __init__(): incompatible constructor arguments. The following argument types are supported:
    1. gnuradio.analog.analog_python.fastnoise_source_c(type: gr::analog::noise_type_t, ampl: float, seed: int = 0, samples: int = 16384)

Invoked with: noise_type_t.GR_GAUSSIAN, 1, -1, 100000
```

marcusmueller:
Chances are high it's me that changed the argument type. Yes, the docstring is incorrect!

[...]

PR #5988, opened 2022-07-04
analog: Fix docstring for noise source seed param

bgottula:
Docstring incorrectly recommended negative seed values.

Signed-off-by: Brett Gottula <bgottula@gmail.com>

---
name: Pull Request Template
title: ''
labels: ''
assignees: ''

---

# Pull Request Details

## Description

Docstring for the `seed` parameter of noise source blocks recommended using a negative value in some cases, but a negative value is not valid since the type for `seed` is `uint64_t` (or will soon be changed to that type).

This ought to be backported to previous releases. I suppose prior back when the `seed` argument had type `long` everywhere a negative value was not invalid, but I'm skeptical that there was ever something special about the sign of the seed to the random number generator.

## Related Issue

Addresses part of #5959 

Docstrings for:
* analog/fastnoise_source
* analog/noise_source

## Testing Done

None. Only documentation has been changed.

## Checklist


bgottula:
A question for reviewers: Is a seed value of 0 special for these blocks? My understanding is that for some noise source blocks in GRC, a seed value of 0 is special because it uses the system clock (or something like that) to pick a seed such that the seed is different on each run of a program, in contrast to non-zero values which produce the same output on every run. I'm not sure if that is true for these blocks, but if it is true that seems relevant to mention in the docstring I'm updating here. (I've been bitten by using `seed=0` in QA flowgraphs, where the tests will pass most of the time but then fail mysteriously later.)

marcusmueller (approved):

PR #5994, opened 2022-07-07
PMT: add string and map/vector<pair<>> interface

marcusmueller:
# Pull Request Details

## Description

PMT dicts are notoriously annoying to generate; with this it gets easier:

```c++
using pmt::mp;
using pmt::pmt_t;
auto dict = dict_from_mapping(std::map<pmt_t, pmt_t>{{mp(foo), mp(12)}, {mp(bar), mp(baz)}});
```

To make the whole generation of symbols less of a copyfest, make symbol
interfaces use `string_view` instead of `const string&`; this allows usage
of PMT symbols without having an owning string.  Goal is to remove
unnecessary copies and allocations for temporary strings.

(while we're at it, remove an unused config.h inclusion)

PMT, and thus all block authors that want to emit dictionaries

## Checklist

### Usage example

https://github.com/gnuradio/gnuradio/blob/66d98ba2adf1c3edc777399b19c07b0bc8b2df64/gr-blocks/lib/probe_rate_impl.cc#L79-L81

### DIFF
--- a/gnuradio-runtime/include/pmt/pmt.h
+++ b/gnuradio-runtime/include/pmt/pmt.h
@@ -12,6 +12,8 @@
 #define INCLUDED_PMT_H
 
 #include <pmt/api.h>
+
+#include <string_view>
 #include <any>
 #include <complex>
 #include <cstdint>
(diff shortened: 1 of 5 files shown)


willcode:
Since `string_view` doesn't control ownership, what happens when a PMT dict is created from some sort of temporary strings? Is this something the dev will need to be aware of?

willcode:
Could you keep both `string` and `string_view` to avoid the API change?

marcusmueller:
pinging @jsallay on this!

willcode:
Since the tests all passed, can we assume that this actually is API compatible with existing `func(const sdt::string &)` calls? If so, then I don't see any problem with more efficient passing of strings.

jsallay (approved):
LGTM

mormj (approved):
Yeah let's go forward with this on `main`

willcode:
@mormj @marcusmueller I think this is actually backportable since const string appears to be promoted to string_view - could you check me on that?

marcusmueller:
@willcode as promised, made backports of this and the depending commits without that API change :)

PR #5996, opened 2022-07-07
Better Probe Rate

marcusmueller:
# Pull Request Details

I needed a better probe_rate block to distinguish between multiple probe rates in one flow graph. So: added a name parameter which is added to the emitted dictionary

probe_rate_message_debug

## Related Issue

Depends on  #5994 

The old format was kept around if name is set to an empty string; this is thus a self-contained change.

## Testing Done

Would love to have done more testing, alas, needs work on message_debug first.

## Documentation Written

- yes. (block docs)
- Adding an example flowgraph to the extended message_debug PR (which hence will depend on this)
- wiki documentation: Will add example flowgraph picture; explain parameter

## Checklist

### DIFF
--- a/gr-blocks/grc/blocks_probe_rate.block.yml
+++ b/gr-blocks/grc/blocks_probe_rate.block.yml
@@ -24,6 +24,9 @@ parameters:
     label: Update Alpha
     dtype: real
     default: '0.15'
+-   id: name
+    label: Name
+    dtype: string
 
 inputs:
 -   domain: stream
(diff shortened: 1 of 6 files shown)


willcode:
If #5994 maintained API compatibility, this would work for 3.10.

willcode:
How does ownership work with `std::string_view`? I see code to hash the string, but I'm missing where it's actually copied. Or is this meant to work only on string constants?

PR #5997, opened 2022-07-07
Message Debug: Add log output

marcusmueller:
# Pull Request Details
## Description

Wanted to work with the probe rate block, ended up having trouble using the output

Decided this thing should be able to log to our logging system instead of the very verbose printf that it does

Allowed people to set log levels.

message_debug

Updated wiki page staging: https://wiki.gnuradio.org/index.php?title=User:MarcusMueller/Message_Debug

## Related Issue
Depends on #5996 

Since the original inputs remain unchanged, this is a self-contained change.

## Testing Done

Only connection testing

## Checklist

### DIFF
--- a/gr-blocks/grc/blocks_message_debug.block.yml
+++ b/gr-blocks/grc/blocks_message_debug.block.yml
@@ -4,13 +4,22 @@ flags: [ python, cpp ]
 
 parameters:
 -   id: en_uvec
-    label: PDU Vectors
+    label: Print PDU contents
     dtype: enum
     default: 'True'
     options: ['True', 'False']
     option_labels: ['On', 'Off']
+-   id: log_level
+    label: Log level
+    dtype: enum
+    default: 'info'
+    options: ['trace', 'debug', 'info', 'warn', 'err', 'critical']
+    option_labels: ['🔬 Trace', '🔎 Debug', '🛈 Info (default)', '⚠ Warning', '💥 Error', '☠ Critical']
 
 inputs:
+-   domain: message
+    id: 'log'
+    optional: true
 -   domain: message
     id: print
     optional: true
@@ -22,14 +31,15 @@ inputs:
     optional: true
 
 templates:
-    imports: from gnuradio import blocks
-    make: blocks.message_debug(${en_uvec})
+    imports: from gnuradio import blocks, gr
+    make: blocks.message_debug(${en_uvec}, gr.log_levels.${log_level})
     callbacks:
     - set_vector_print(${en_uvec})
+    - level(gr.log_levels.${log_level})
 
 cpp_templates:
     includes: ['#include <gnuradio/blocks/message_debug.h>']
     declarations: 'blocks::message_debug::sptr ${id};'
-    make: 'this->${id} = blocks::message_debug::make();'
+    make: 'this->${id} = blocks::message_debug::make(${en_uvec}, spdlog::level::${log_level});'
 
 file_format: 1
--- a/gr-blocks/include/gnuradio/blocks/message_debug.h
+++ b/gr-blocks/include/gnuradio/blocks/message_debug.h
@@ -1,6 +1,7 @@
 /* -*- c++ -*- */
 /*
  * Copyright 2005,2012-2013 Free Software Foundation, Inc.
+ * Copyright 2022,2023 Marcus Müller
  *
  * This file is part of GNU Radio
  *
@@ -13,6 +14,7 @@
 
 #include <gnuradio/block.h>
 #include <gnuradio/blocks/api.h>
+#include <spdlog/common.h>
 
 namespace gr {
 namespace blocks {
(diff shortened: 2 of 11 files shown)


gr-blocks/grc/blocks_message_debug.block.yml
```diff
@@ -4,13 +4,22 @@ flags: [ python, cpp ]
...
     dtype: enum
     default: 'True'
     options: ['True', 'False']
     option_labels: ['On', 'Off']
+-   id: log_level
+    label: Log level
+    dtype: enum
+    default: 'info'
+    options: ['trace', 'debug', 'info', 'warn', 'err', 'critical']
+    option_labels: ['🔬 Trace', '🔎 Debug', '🛈 Info (default)', '⚠ Warning', '💥 Error', '☠ Critical']
```
mbr0wn, line 17: :eyes:

mbr0wn:
I've only skimmed the code, but what I saw seems sensible. Any reason we can't go ahead with this PR?

marcusmueller:
@mbr0wn this PR was basically blocked on #5994 but now the pipeline's been unclogged

gr-blocks/lib/message_debug_impl.cc
```diff
@@ -24,30 +25,55 @@ namespace blocks {
...
-message_debug::sptr message_debug::make(bool en_uvec)
+message_debug::sptr message_debug::make(bool en_uvec, spdlog::level::level_enum log_level)
 {
-    return gnuradio::make_block_sptr<message_debug_impl>(en_uvec);
+    return gnuradio::make_block_sptr<message_debug_impl>(en_uvec, log_level);
 }
 
-message_debug_impl::message_debug_impl(bool en_uvec)
+message_debug_impl::message_debug_impl(
+    bool en_uvec, spdlog::level::level_enum log_level = spdlog::level::info)
```
willcode, line 34: Do we need a default in the impl, or just in the .h?

willcode (commented):
Could you also comment on the API compatibility of std::string_view and String&. If code using String& will compile against the new API without change, I'd rather keep main and maint-3.10 as close as possible.

marcusmueller:
> Could you also comment on the API compatibility of std::string_view and String&. If code using String& will compile against the new API without change, I'd rather keep main and maint-3.10 as close as possible.

That's why both the maint-3.10 and the main version of this are the same API: to keep them as identical as possible :)

willcode:
@marcusmueller Rephrasing the question: I was reading that you can wholesale replace `const std::string&` with `std::string_view` in an API if certain ownership, persistence, and cstr issue are ruled out. It would be useful to keep the 3.10 and "3.11" codebases even closer if possible. Do you have thoughts on using `std::string_view` in the API for 3.10? The only catch I see is null-terminated strings (maybe?).

gr-blocks/include/gnuradio/blocks/message_debug.h
```diff
@@ -73,6 +77,15 @@ class BLOCKS_API message_debug : virtual public block
...
     virtual void set_vector_print(bool en) = 0;
+
+    /*!
+     * \brief get the log level
+     */
+    virtual spdlog::level::level_enum level() const = 0;
+    /*!
+     * \brief set the log level
+     */
+    virtual void level(spdlog::level::level_enum log_level) = 0;
```
willcode, line 88: `set_level(...)` would be good for consistency.

marcusmueller:
> @marcusmueller Rephrasing the question: I was reading that you can wholesale replace `const std::string&` with `std::string_view` in an API if certain ownership, persistence, and cstr issue are ruled out. It would be useful to keep the 3.10 and "3.11" codebases even closer if possible. Do you have thoughts on using `std::string_view` in the API for 3.10? The only catch I see is null-terminated strings (maybe?).

I'm a bit confused: We're talking about the PMT API, not about the API of message debug or probe rate here, right?

In that case: I didn't want to change the PMT API on maint-3.10 this fundamentally. Though I really can't think of a case where the compiler wouldn't tell you you've done something wrong when replacing `const std::string&` with `std::string_view`, it might still break existing code (e.g. when a type has a cast for std::string but not for std::string_view).

willcode:
OK - can you look at the couple of comments I had, and we can get this one in.

gr-blocks/lib/message_debug_impl.cc
marcusmueller: Only in the h, indeed, lemme fix that.

willcode:
Thank. Will merge when CI is done. You'll do the same in the backports?

marcusmueller:
of course!

willcode:
Note that `🛈` did not render on a user's Ubuntu 22.04 system. Is there a way a ensure "typical" installs support all the characters we use? This particular glyph has been around since 2014.

One solution for Ubuntu 22.04 is to install the fonts-noto-core package.

PR #5999, opened 2022-07-07
GRC: numbers: supersede often misled "engineering" representation

### LINKED ISSUE #5986: GRC shouldn't do anything but simple exponent notation
### Feature Description

While well-intended and used for a decade, it's a common problem that people try to enter "1.024k" into GRC fields, because that's how GRC displays 1024. That's wrong on multiple levels (1.024k being longer than 1024, which is still shorter than the precision set for the floating point representation, and nobody in the real world saying "I have an one point o two four kilo-point FFT", aside from the lack of space between number and unit prefix, and unit prefixes usually not being used without a physical unit, and most importantly, 1024 is an integer, precisely, 1.024k might just as well be 1024.49, and albeit we can't get rid of that ambiguity for really large numbers, for 1024 it still is very confusing); it's time to change that.

Let's simply do 

| Number | formatted output | remark |
|------------|------------------------|----------|
| 1 | 1 |
| -1 | -1 |
| 2500 | 2500 |
| 1000000000 | 1e9 |
| 1234567890 | 1.23457e9 | or similar, like 1.23457e+9 |
| 0.2 | 0.2 |
| 0.2020012 | 0.202001 |
| 0.00000123 | 1.23e-6 | 

That would allow for much better readability, less beginner confusion, copy&paste, and loss of duplicate code.

### Feature Urgency

### More Information

After [a discussion](https://matrix.to/#/!BMFENoeifGOSJaeGco:gnuradio.org/$phAFgmkARj_4V-Y-bnIOdgc9CSV-FLqzDlcM5iiMlW0?via=gnuradio.org&via=matrix.org&via=ei8fdb.org) in https://matrix.to/#/#grc-dev:gnuradio.org

marcusmueller:
# Pull Request Details

## Problem being solved

GRC currently displays numbers, regardless of their "integerness", in a kind-of-engineering notation. On one hand, that's fine, for fields like "Carrier Frequency" with values like 2 400 000.2 (that would be displayed as "2.4G"); on the other hand, that's very confusing for values like 1024 (which often appears as count of elements), which under no sensible circumstances should be displayed as "1.024k" or 0.1, which gets displayed as "100m".

Another issue arises because "2.4G" is not actually valid Python, and hence can't be entered into a field. So, a user *sees* a number displayed in a specific format, and tries to use that format (which means they've *intuited* from the UI!), which then promptly fails. We want these fields to evaluate valid python, and I'd rather not invent a new language where we can mix Python and 2.4G-style numbers, so entering "2.4G" has to fail.

For most of us, reading 2.4G isn't nicer than 2.4e9; and unlike the former, the latter is valid python. Let's use that format for display!

## Description

* modest Ints: under no circumstances write 1024 as 1.024k or 1.024e3
* large ints: from (incl) 1 million to (excluding) 10⁹, 1_000_000 is the format of choice (and valid python) (later edit: we could alternatively use spaces to group thousands, and hope people won't try to enter spaces, but spaces look less like a "character I should also use" than "G" or ","; commas, by the way, would also make it hard to localize later on)
* supersized ints: 2³⁰ -> 1.07374e9
* modest floats: 0.12
* very small/large floats: -1.2e-18
* never does the 1.2e`+0`9 that sadly `f"{number:g}"` introduces
 
## Related Issue

Closes #5986 

Display of parameters in GRC (and only there)

## Testing Done

Since visual: Here's a screenshot. In the comments below the blocks, the literal inputs

### Before

before

### After

GRC

## Checklist

### DIFF
--- a/grc/gui/Utils.py
+++ b/grc/gui/Utils.py
@@ -1,5 +1,6 @@
 """
 Copyright 2008-2011,2015 Free Software Foundation, Inc.
+Copyright 2022 Marcus Müller
 This file is part of GNU Radio
 
 SPDX-License-Identifier: GPL-2.0-or-later
@@ -9,8 +10,8 @@ SPDX-License-Identifier: GPL-2.0-or-later
 
 from sys import platform
 import os
-import numbers
 
+import numpy
 from gi.repository import GLib
 import cairo
 
@@ -70,28 +71,35 @@ def align_to_grid(coor, mode=round):
         return align(coor)
 
 
-def num_to_str(num):
+def num_to_str(num) -> str:
     """ Display logic for numbers """
-    def eng_notation(value, fmt='g'):
-        """Convert a number to a string in engineering notation.  E.g., 5e-9 -> 5n"""
-        template = '{:' + fmt + '}{}'
-        magnitude = abs(value)
-        for exp, symbol in zip(range(9, -15 - 1, -3), 'GMk munpf'):
-            factor = 10 ** exp
-            if magnitude >= factor:
-                return template.format(value / factor, symbol.strip())
-        return template.format(value, '')
 
-    if isinstance(num, numbers.Complex):
-        num = complex(num)  # Cast to python complex
-        if num == 0:
-            return '0'
-        output = eng_notation(num.real) if num.real else ''
-        output += eng_notation(num.imag, '+g' if output else 'g') + \
-            'j' if num.imag else ''
-        return output
-    else:
-        return str(num)
+    def clean_e(unclean: str) -> str:
+        """
+        Removes the + from 1.2e+09, and the 0 from both 1.2e-09 and 1.2e09
+        """
+        return unclean\
+            .replace("e-0", "e-")\
+            .replace("e+0", "e+")\
+            .replace("e+", "e")\
+            .strip()
+
+    if numpy.iscomplex(num):
+        intreal = int(numpy.real(num))
+        intimag = int(numpy.imag(num))
+        if intreal == numpy.real(num) and \
+           intimag == numpy.imag(num) and \
+           abs(num) < 2**0.5 * 10**6:
+            if intreal < 10**6 and intimag < 10**6:
+                return f"{intreal:d}{intimag:+d}j"
+            return f"{intreal:_d}{intimag:+_d}j"
+        return clean_e(f"{numpy.real(num):5.4g}{numpy.imag(num):+5.4g}j")
+
+    if int(num) == num and abs(num) < 10**9:
+        if abs(num) < 10**6:
+            return f"{int(num):d}"
+        return f"{int(num):_d}"
+    return clean_e(f"{num:9g}")
 
 
 def encode(value):
--- a/grc/gui/Utils.py
+++ b/grc/gui/Utils.py
@@ -84,6 +84,18 @@ def num_to_str(num) -> str:
             .replace("e+", "e")\
             .strip()
 
+    def thousands_sep(inputstr: str,
+                      replacement: str = "\N{SPACE}",
+                      original_sep: str = "_") -> str:
+        """
+        Python allows us to group thousands with _ , but not with arbitrary
+        characters.
+
+        This function is just a simple str.replace wrapper to keep the default
+        somewhere visible
+        """
+        return inputstr.replace(original_sep, replacement)
+
     if numpy.iscomplex(num):
         intreal = int(numpy.real(num))
         intimag = int(numpy.imag(num))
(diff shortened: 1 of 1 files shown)


grc/gui/Utils.py
```diff
@@ -70,28 +71,35 @@ def align(value):
...
+            .replace("e+0", "e+")\
+            .replace("e+", "e")\
+            .strip()
+
+    if numpy.iscomplex(num):
+        intreal = int(numpy.real(num))
+        intimag = int(numpy.imag(num))
+        if intreal == numpy.real(num) and \
+           intimag == numpy.imag(num) and \
+           abs(num) < 2**0.5 * 10**6:
```
willcode, line 92: Could it just be based on max(real, imaj)? Using abs could end up causing some strange break points. Of course, this only happens when there's a pretty weird complex number in the first place.

noc0lour:
Does the notation also work for C++ flowgraph generation. E.g. 2_000_000 integer representation? Or does this require some line in documentation highlighting the difference?

marcusmueller:
@noc0lour  Just to be clear, this *only* changes how numbers are displayed in GRC, it has no influence on any code generation. To the generated code, numbers are generated as before.

grc/gui/Utils.py
marcusmueller: Yeah, I thought about that, but came down to the point where I said "If the magnitude > 1.4e6, you'll always want _-notation for both parts, both if they are very different in magnitude, or when they are roughly the same". But since this is for integer im, re only, anyways, not so sure it makes too much sense.

Happy to change it if you have strong feelings about it, otherwise would leave it like it is.

777arc:
If this is only for GRC viewing sake than why the underscores for large ints instead of something common like commas?  I've never seen underscores used in place of commas before.  I think people won't make the same mistake of copying commas like they did with k's and m's.

marcusmueller:
Well, but `1_000_000` is valid python, `1,000,000` is also valid python, but denotes a tuple `(1, 0, 0)`, so that's going to be quite confusing in itself, @777arc

marcusmueller:
Tried things out:

| Unicode space type | Look | comment |
|--------------------------|------|----------------|
| EN SPACE | thousands en space separator | Half em |
| PUNCTUATION SPACE | thousands punctuation space separator | "space equal to narrow punctuation of a font" |
| SPACE | thousands space separator | Regular " " space |
| THREE-PER-EM SPACE | thousands three-per-em space separator | 1/3 em |
| FOUR-PER-EM SPACE | thousands four-per-em space separator | ¼ em |
| THIN SPACE | thousands thin space separator | "a fifth of an em (or sometimes a sixth)" |
| SIX-PER-EM SPACE | thousands six-per-em space separator | 1/6 em |
| HAIR SPACE | thousands hair space separator | not to be confused with Hair, the musical |

marcusmueller:
looking at this, I'm tending towards thin or 1/6 em space

777arc:
Thin or 1/6 and I think it looks excellent compared to underscores

muaddib1984:
I like the larger options en space, but I'm viewing this on a small screen. 

Is it just me or does it seem that this is the dawning of the age of Aquarius.

marcusmueller:
> Is it just me or does it seem that this is the dawning of the age of Aquarius.

and a home for the fleas in my hair

dkozel:
1/4th or Thin look good to me. This seems like a good time to ask @ei8fdb for an opinion and/or a quick spot check with the UX group?

marcusmueller:
@dkozel exactly! Got the technical feasibility sorted out, just updating the "what am I solving here" to avoid having to follow links to other issues, let's get @ei8fdb in on this!

ei8fdb:
(I was on hols so getting to this now.)


(Trying to not sound negative!)

The original problem statement was that "GRC currently displays numbers, regardless of their "integerness", in a kind-of-engineering notation." This is a usability issue. This PR won't really fix this usability issue, but it certainly improves the situation. 

To make sure I understnd this PR correctly - it's about how to display, and the spacing between characters in GRC UI?

If this is right, in terms of the spacing between characters, I'd agree with @dkozel for this PR either 1/4em or thin looks fine to me.

The problem is about displaying numbers is a way that is understandable/readable to users.

@marcusmueller @dkozel I'll start a discussion in the Matrix.

marcusmueller:
Discussion with @ei8fdb yielded:

I should do a quick survey, maybe a low-quality, high-count one on twitter, and a more-feedback, potentially-fewer-participants one via email, and/or make this selectable.

Three options: 

1. Display as before, i.e. SI-prefixes, but invalid as input: 2.4e9-> 2.4G, 0.2 -> 200m, 1024 -> 1.024k
2. Display as here, i.e. a compromise between somewhat human-readable and fully valid python: 2.4e9 -> 2.4e9, 0.2 -> 0.2, 1024 -> 1024
3. Display as human readable, but where not valid, impossible to mistake for valid input: 2.4e9 -> 2.4 · 10⁹, 0.2 -> 0.2, 1024 -> 1024

marcusmueller:
TODO: Convert me to issue against GRC-QT,

marcusmueller:
Since not a problem in Qt-GRC, closing this.

PR #6002, opened 2022-07-09
add missing includes to fix build with newer versions of GCC

danielnachun:
---
name: Pull Request Template
title: ''
labels: ''
assignees: ''

---

# Pull Request Details

## Description
On newer versions of GCC, `cmath` is not included by default and must be explicitly added with `include`.  If it is missing the build will fail because of undefined variables.

## Related Issue
Fixes the failure we saw in Homebrew CI on Linux: https://github.com/Homebrew/homebrew-core/pull/105154

This has no runtime effects - it only fixes a compilation error.

## Testing Done
A patch with the same fix was run in https://github.com/Homebrew/homebrew-core/pull/105154 and the build now succeeds.

## Checklist

### DIFF
--- a/gr-qtgui/lib/FrequencyDisplayPlot.cc
+++ b/gr-qtgui/lib/FrequencyDisplayPlot.cc
@@ -16,7 +16,7 @@
 #include <gnuradio/qtgui/qtgui_types.h>
 #include <qwt_scale_draw.h>
 #include <QColor>
-
+#include <cmath>
 
 /***********************************************************************
  * Widget to provide mouse pointer coordinate text
--- a/gr-qtgui/lib/VectorDisplayPlot.cc
+++ b/gr-qtgui/lib/VectorDisplayPlot.cc
@@ -25,6 +25,8 @@
 #include <qwt_legend_label.h>
 #endif /* QWT_VERSION < 0x060100 */
 
+#include <cmath>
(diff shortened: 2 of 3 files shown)


marcusmueller:
Hey @danielnachun! Thank you for working on the homebrew package, that is much appreciated.

I don't think this should have influence on the issue, as I consider the fix "include cmath if you use cmath functions" correct, but: what's the GCC version homebrew is currently using, where this happened?

danielnachun:
We are using GCC 11.  I believe a lot of changes to default includes happened in GCC 10 as I’ve seen similar patches in other software that were needed after GCC 10 was released.

danielnachun:
Yes we do use parallel builds with GCC, so there could be build order differences.  And we are using Ubuntu, which may handle the C++ stdlib a little differently.  In fact I wouldn't be surprised if Fedora even does some patching of `libstdc++` to maintain backwards compatibility with code written for older GCC that assumes more headers are included by default.  Interestingly, Clang + libc++ on macOS doesn't have any of these issues with missing `#include` statements even for very old code, which I suspect is because they wanted to maintain backwards compatibility as well.

l29ah:
Same issue on Gentoo on gcc-12.

mormj (approved):

Issue #5993, opened 2022-07-06
config.h included in very many places, nearly use nowhere


marcusmueller:
### What happened?

for historical reasons, we've got 

```c++
#ifdef HAVE_CONFIG_H
#include <config.h>
#endif
```

all over the place.

And luckily, we nearly never need that. 

So, in the interest of code hygiene, we should remove that

### System Information

n/a

### GNU Radio Version

3.11-git (main)

### Specific Version

main

### Steps to Reproduce the Problem

```
rg -l '#include.*config\.h'
```

### Relevant log output

```shell
gr-network/lib/tcp_connection.cc
gr-fft/lib/fft_v_fftw.cc
gr-fft/lib/qa_fft_shift.cc
gr-fft/lib/goertzel.cc
gr-fft/lib/ctrlport_probe_psd_impl.cc
gr-fft/lib/goertzel_fc_impl.cc
gr-fft/lib/window.cc
gr-network/lib/socket_pdu_impl.cc
gr-network/lib/udp_sink_impl.cc
gr-network/lib/tcp_sink_impl.cc
gr-network/lib/udp_source_impl.cc
gr-network/lib/tuntap_pdu_impl.cc
gr-network/lib/stream_pdu_base.cc
gnuradio-runtime/lib/block.cc
gnuradio-runtime/lib/sync_decimator.cc
gnuradio-runtime/lib/flat_flowgraph.cc
gnuradio-runtime/lib/msg_handler.cc
gnuradio-runtime/lib/qa_vmcircbuf.cc
gnuradio-runtime/lib/block_detail.cc
gnuradio-runtime/lib/qa_io_signature.cc
gnuradio-runtime/lib/realtime.cc
gnuradio-runtime/lib/pdu.cc
gnuradio-runtime/lib/sptr_magic.cc
gnuradio-runtime/lib/sync_block.cc
gnuradio-runtime/lib/logger.cc
gnuradio-runtime/lib/buffer_reader_sm.cc
gnuradio-runtime/lib/top_block.cc
gnuradio-runtime/lib/vmcircbuf.cc
gnuradio-runtime/lib/scheduler.cc
gnuradio-runtime/lib/vmcircbuf_mmap_shm_open.cc
gnuradio-runtime/lib/qa_buffer.cc
gnuradio-runtime/lib/local_sighandler.cc
gnuradio-runtime/lib/tpb_detail.cc
gnuradio-runtime/lib/hier_block2_detail.cc
[...]

PR #6005, opened 2022-07-10
runtime: make a canary that lets compilation fail w/o HAVE_CONFIG_H

marcusmueller:
Don't merge, this is just a canary for #5993 to see whether this applies to all tested platforms.

Tests should fail early.


marcusmueller:
Yeah, fails everywhere: HAVE_CONFIG_H is simply never defined, and all our checks from config.h are always false, no matter whether they've been configured correctly :(

mormj:
Closing this for now - can be reopened when HAVE_CONFIG_H issues are addressed

PR #6012, opened 2022-07-18
global: Initial port of newsched to dev-4.0 branch

mormj:
This includes the base set of functionality porting newsched 0.5 into
the gnuradio dev-4.0 branch

From newsched, the following things have been removed for now:
- qtgui (needs reworking/replacing generally)
- soapy (only a couple blocks implemented)
- distributed runtime (proof of concept-y)
- rpc
- grc (do we want to start with qt version?)


marcusmueller (approved):

blocklib/analog/.gitignore
```diff
@@ -0,0 +1 @@
+meson.build
```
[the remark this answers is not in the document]
mormj: Probably.  These .gitignores get ugly

kernel/lib/analog/agc.cc
```diff
@@ -0,0 +1,36 @@
+#include <gnuradio/kernel/analog/agc.h>
+
+namespace gr {
+namespace kernel {
+namespace analog {
+
+template <typename T>
+T agc<T>::scale(T input)
+{
+    T output = input * _gain;
```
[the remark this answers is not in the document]
mormj: once we settle on that as a coding guideline, will have to go back through and replace all the `d_`'s

blocklib/math/add/add_cpu.cc
```diff
@@ -0,0 +1,67 @@
...
+
+template <class T>
+work_return_code_t add_cpu<T>::work(work_io& wio)
+{
+    auto out = wio.outputs()[0].items<T>();
+    auto noutput_items = wio.outputs()[0].n_items;
+    int noi = d_vlen * noutput_items;
+
+    memcpy(out, wio.inputs()[0].items<T>(), noi * sizeof(T));
+    for (size_t i = 1; i < wio.inputs().size(); i++) {
```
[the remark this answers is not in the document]
mormj: there's a bunch of places like this that need to be cleaned up - straight ports from the old blocks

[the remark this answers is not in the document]
mormj: actually this one is tricky because it starts from `1`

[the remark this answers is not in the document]
marcusmueller: ah we're C++17, right? no std::view that allows us to skip the first element then...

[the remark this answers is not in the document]
mormj: I don't think it would be ridiculous to jump to C++20 given that the min distro we prob need to support is ubuntu 22 which has gcc11

bench/bm_msg_forward.cc
```diff
@@ -0,0 +1,89 @@
...
+#include <iostream>
+#include <thread>
+
+#include <gnuradio/blocks/msg_forward.h>
+#include <gnuradio/flowgraph.h>
+#include <gnuradio/realtime.h>
+#include <gnuradio/runtime.h>
+#include <pmtf/base.hpp>
+
+#include <iostream>
```
jdemel, line 11: `iostream` is included twice. It'd be interesting to replace `iostream` with `fmt`. e.g. use `fmt::print`.

bench/meson.build
```diff
@@ -0,0 +1,56 @@
+# incdir = include_directories('../include')
+
+if (CLI11_dep.found())
+srcs = ['bm_copy.cc']
+executable('bm_nbt_copy', 
+    srcs, 
+    # include_directories : incdir, 
+    link_language : 'cpp',
```
eli-schwartz, line 8: I'm guessing you are overriding to cpp because you don't want to link with the cuda linker?

blocklib/blocks/include/gnuradio/blocks/meson.build
```diff
@@ -0,0 +1,35 @@
...
+
+pkg = import('pkgconfig')
+libs = []     # the library/libraries users need to link against
+h = ['.'] # subdirectories of ${prefix}/${includedir} to add to header path
+pkg.generate(libraries : libs,
+             subdirs : h,
+             version : meson.project_version(),
+             name : 'libgnuradio-blocks',
+             filebase : 'gnuradio-blocks',
+             install_dir : get_option('prefix') / 'lib' / 'pkgconfig',
```
eli-schwartz, line 34: This should already be the default installation directory, you don't need to override it.

blocklib/blocks/include/gnuradio/blocks/meson.build
```diff
@@ -0,0 +1,35 @@
...
+  install_dir : get_option('prefix') / 'lib' / 'cmake' / 'gnuradio',
+  configuration : cmake_conf
+)
+
+pkg = import('pkgconfig')
+libs = []     # the library/libraries users need to link against
+h = ['.'] # subdirectories of ${prefix}/${includedir} to add to header path
+pkg.generate(libraries : libs,
+             subdirs : h,
+             version : meson.project_version(),
```
eli-schwartz, line 31: This is the default version if unspecified.

blocklib/blocks/include/gnuradio/blocks/meson.build
```diff
@@ -0,0 +1,35 @@
...
+install_headers(blocks_headers, subdir : 'gnuradio/blocks')
+
+# TODO - export this as a subproject of gnuradio
+
+conf = configuration_data()
+conf.set('prefix', prefix)
+conf.set('exec_prefix', '${prefix}')
+conf.set('libdir', join_paths('${prefix}',get_option('libdir')))
+conf.set('includedir', join_paths('${prefix}',get_option('includedir')))
+conf.set('LIBVER', '0.0.1')
```
eli-schwartz, line 14: Is this being used for anything? I don't see it referenced, and it looks like it's configuration data for a pkg-config file but you use Meson's generator anyway, which handles this for you.

blocklib/fec/lib/reed-solomon/README
```diff
@@ -0,0 +1,2 @@
+This code is from http://people.qualcomm.com/karn/code/fec
+It is based on reed-soloman-3.1.1 (1 Jan 2002).
```
eli-schwartz, line 2: Maybe this would fit better as subprojects/reed-solomon?

blocklib/filter/python/filter/meson.build
```diff
@@ -0,0 +1,54 @@
...
+
+filter_pybind_sources += gen_filter_pybind
+
+gnuradio_blocklib_filter_pybind = py3_inst.extension_module('filter_python',
+    filter_pybind_sources, 
+    include_directories: ['../../lib'],
+    dependencies : [gnuradio_blocklib_filter_dep, python3_dep, pybind11_dep],
+    link_language : 'cpp',
+    install : true,
+    install_dir : join_paths(py3_inst.get_install_dir(),'gnuradio','filter')
```
eli-schwartz, line 45: You can use `subdir: 'gnuradio/filter'` for this, I think.

meson.build
```diff
@@ -0,0 +1,95 @@
...
+fmt_dep = dependency('fmt', method: 'cmake', modules: ['fmt::fmt'])
+spdlog_dep = dependency('spdlog', method: 'cmake', modules : ['spdlog::spdlog'])
+python3_dep = dependency('python3', required : get_option('enable_python'))
+python3_embed_dep = dependency('python3-embed', required : get_option('enable_python'))
+
+# Import python3 meson module which is used to find the Python dependencies.
+py3_inst = import('python').find_installation('python3')
+py3_mod = import('python3')
+py3 = py3_mod.find_python()
+py3_version = py3_mod.language_version()
```
eli-schwartz, line 32: This module (python3) is deprecated and replaced by the "python" module. As far as I can tell it is only used as a script runner, though? You can use `py3_inst` from the python module as a script runner too.

```suggestion
py3 = py3_inst
py3_version = py3_inst.language_version()
```

subprojects/CLI11.wrap
```diff
@@ -0,0 +1,3 @@
+[wrap-git]
+url = https://github.com/CLIUtils/CLI11
+revision = tags/v2.1.2
```
eli-schwartz, line 3: I recommend you instead use the wrap file installed by:

```
meson wrap install cli11
```

It looks like this:
```[wrap-file]
directory = CLI11-2.2.0
source_url = https://github.com/CLIUtils/CLI11/archive/refs/tags/v2.2.0.tar.gz
source_filename = CLI11-2.2.0.tar.gz
source_hash = d60440dc4d43255f872d174e416705f56ba40589f6eb07727f76376fb8378fd6

[provide]
cli11 = CLI11_dep
```

It works even without git, it securely checksums the external resource...

meson.build
```diff
@@ -0,0 +1,95 @@
...
+pybind11_dep = dependency('pybind11', required : get_option('enable_python'))
+threads_dep = dependency('threads')
+gtest_dep = dependency('gtest', main : true, version : '>=1.10', required : get_option('enable_testing'))
+cuda_dep = dependency('cuda', version : '>=10.1', required : USE_CUDA, modules : ['cufft'])
+
+libpmtf = subproject('pmt')
+pmtf_dep = libpmtf.get_variable('pmtf_dep')
+
+libCLI11 = subproject('CLI11')
+CLI11_dep = libCLI11.get_variable('CLI11_dep')
```
eli-schwartz, line 46: ... and you can change this to
```suggestion
CLI11_dep = dependency('CLI11')
```

subprojects/json.wrap
```diff
@@ -0,0 +1,4 @@
+[wrap-git]
+url = https://github.com/nlohmann/json
```
eli-schwartz, line 2: This is available as `meson wrap install nlohmann_json` and `dependency('nlohmann_json')`

mormj:
@eli-schwartz - thank you for the review and guidance on the meson code - having expert eyes there is super helpful

bench/meson.build
mormj: yes, it was causing issues with the CUDA version I had at the time this was implemented.  Not sure if newer versions will link better.

blocklib/fec/lib/reed-solomon/README
mormj: Perhaps - this was a straight port from GR.  I'll leave it as is for now, but open to a later PR

meson.build
mormj: I'm going to postpone the `meson wrap` changes to a separate PR - this one gets a bit complicated by the fact that PMT uses CLI11 as well - want to fix that one first

bench/meson.build
eli-schwartz: Newer versions of Meson don't change anything here. The cuda linker has a higher priority for autodetecting the "best" linker.

There *is* a change in Meson git master for the cuda linker to avoid manually injecting `-lstdc++` (cuda handles this automatically just like the g++ driver) which may conflict with `-lc++`. But I'm not sure that's relevant here. Also, the change is not yet released.

meson.build
eli-schwartz: FWIW, the first project to configure the subproject "wins".

It's an easy fix anyway, but I found a couple more: https://github.com/gnuradio/pmt/pull/52

mormj:
Since CI is passing, going ahead and merging this so more PRs can be created against `dev-4.0`

Meson still needs some cleanup, but will reserve that for separate PRs.

PR #6018, opened 2022-07-22
WIP: 4.0 - Registry and reflection capability

mormj:
Create a GR registry that maintains knowledge about what modules,
blocks, etc. are available, and has the ability to return generic block
constructors.

Basic usage would be
```
auto list = gr::registry::modules()
// "math" should be in list
list = gr::registry::blocks("math");
// "vector_source_f" should be in list
auto f = gr::registry::factory("math","vector_source_f")
auto blk = f(.... map of string/pmt)
```

Right now, this is all done through auto code generation, with the need being 
`.yml` --> headers, then headers should include reflection capability

This adds a generic factory to each block that can be invoked with `std::map<std::string, pmtf::pmt>' - which is a list of all the parameters in the block.  Since some parameters are not part of the constructor, these will need to be handled through the setters.

There is probably a more elegant solution using `refl-cpp` which can then be wrapped for nicer python bindable API, but this first pass is pretty brute force just generate the code.  I still question whether some of these things can be done with a general purpose tool like refl-cpp, such as some parameters being part of the constructor and some being setters

Still needing to be addressed:

- [ ] python bindings of `registry.h`
- [ ] block parameters and callbacks
- [ ] rest of the things in `block.yml`
- [ ] parameters that are not part of constructor
- [ ] schedulers, buffers, other modular components

### DIFF
--- a/blocklib/blocks/vector_sink/vector_sink_cpu.h
+++ b/blocklib/blocks/vector_sink/vector_sink_cpu.h
@@ -29,7 +29,7 @@ public:
     // private member variable, and pass it out as pmt when queried
     void on_parameter_query(param_action_sptr action) override
     {
-        this->d_debug_logger->debug(
+        this->d_debug_logger->trace(
             "block {}: on_parameter_query param_id: {}", this->id(), action->id());
         pmtf::pmt param = d_data;
         // auto data = pmtf::get_as<std::vector<float>>(*param);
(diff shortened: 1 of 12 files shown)


willcode:
I like the idea. My eyes are going a little buggy on the templates ... have to spend some more time looking at the details.

mormj:
The templates generally in 4.0 are a bit messy and need some refactoring.  

One thing that I just thought about with the templates generally - there is now python in the loop for all build environments because of the jinja based codegen.  You cannot build GR without python, not required at runtime, though.

mormj:
Also, there is a chicken and egg problem with the qa test :/

Need to tell the registry::init method the location of the build .so files somehow - which would come through meson I suppose

PR #6023, opened 2022-07-24
meson: do not require cmake when finding dependencies

eli-schwartz:
fmt and spdlog provide pkg-config files, and these work and compile perfectly fine. Meson can try finding with both pkg-config and cmake, if it is allowed to, which means that dependency lookup now works on systems where cmake is not installed.

Note this is not immediately useful, since later on a cmake subproject is explicitly used, but that can be fixed in the future to make cmake no longer required for a successful full configure.

Even when using cmake for dependency lookup, the modules kwarg is not needed. Both dependencies export a single cmake target, so Meson can figure out the correct one to use without trouble.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -20,8 +20,8 @@ SCRIPTS_DIR=join_paths(meson.project_source_root(),'utils','blockbuilder','scrip
 
 volk_dep = dependency('volk', version : '>=2.2')
 yaml_dep = dependency('yaml-cpp', version : '>=0.6')
-fmt_dep = dependency('fmt', method: 'cmake', modules: ['fmt::fmt'])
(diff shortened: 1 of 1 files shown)


mormj (approved):

mormj:
@eli-schwartz - more general meson question.  Is there a way to properly export the nlohmann_json dependency.  E.g. gnuradio has it as a subproject, but header files that are installed (e.g. block.h) #include<nlohmann/json.h> -- now for modules built on top of GNU Radio, do they also need to have nlohmann_json as a subproject, or is there a way to install it and export the dependency through meson?

eli-schwartz:
The meson.build for nlohmann_json is part of the upstream repo, and it only installs itself `if not meson.is_subproject()`. It sounds like you want it to be installed even as a subproject, so you will probably have to get that patched upstream.

PR #6027, opened 2022-07-25
4.0: Integration of GRC

mormj:
## Description
This is a port of GRC into the `dev-4.0` branch - yes, it is the GTK version of GRC and not the newest QT work, but we can do that transition when it is ready.  There is a lot more I'd like to do with GRC, namely:
1) Modularize it and detach from GNU Radio - (https://github.com/gnuradio/greps/blob/main/grep-0025-grc-out-of-tree.md)
2) Work directly on the block yml files (file_format: 2??) and not autogenerate .grc files

But it seems important to get basic grc functionality in place before major improvements are considered

All blocks as far as the autogenerated grc

## Testing Done
Installed and run flowgraph through GRC

## Checklist


mormj:
Going ahead and merging this even if we have bigger plans for GRC, it is a way to demonstrate some of the new features for now.

PR #6047, opened 2022-07-28
Allow GNU radio to be included in CMake-based projects

wgledbetter:
Signed-off-by: William G Ledbetter <williamgledbetter@gmail.com>

# Pull Request Details
There is currently not a clean way to make a CMake-based project that includes the gnuradio source via `add_subdirectory`. because many cmake commands are referenced off of the project root instead of the relative gnuradio root. Pretty much anything that uses `CMAKE_SOURCE_DIR` is subject to breaking.

This PR fixes that by introducing `GNURADIO_SOURCE_DIR` and `GNURADIO_BINARY_DIR` and doing a find-and-replace for `CMAKE_SOURCE_DIR` and `CMAKE_BINARY_DIR`. On the `SOURCE` side, this makes sure all paths are relative to gnuradio's root rather than any higher-level project root. On the `BINARY` side, it's mostly for cleanliness so that gnuradio binaries don't pollute the higher-level project's outputs.

## Description
I created new CMake variables `GNURADIO_SOURCE_DIR` and `GNURADIO_BINARY_DIR` that point to the source and binary roots of gnuradio.

## Testing Done
All targets build successfully, indicating that include and link paths are correct.

## Checklist


marcusmueller:
Honestly, nice! But I've got questions: Since I've never considered using GNU Radio this way, do you have a demo project that makes use of this? Would it make sense to detect this kind of usage, and disable the "build all components possible" default and instead just compile the components used by the top-level project?

willcode:
Good to backport to 3.10? It looks like the behavior is supposed to be the same.

ryanvolz:
Might I suggest using `PROJECT_SOURCE_DIR` and `PROJECT_BINARY_DIR` in place of introducing `GNURADIO_SOURCE_DIR` and `GNURADIO_BINARY_DIR`? The result should be the same, and this has the benefit of being documented and searchable within the CMake docs and not just a definition local to GNU Radio.

wgledbetter:
@ryanvolz Ah, yeah. That's a much cleaner solution. Changed in b70e97972f102bcfb9cc78623e681e81645b7929

wgledbetter:
@marcusmueller Here's a repo where I just took the dial_tone example and implemented it with gnuradio as a submodule. https://github.com/wgledbetter/gr-example

As for detecting this kind of usage and disabling targets, the top-level project has the option to hide targets via `add_subdirectory(gnuradio EXCLUDE_FROM_ALL)`, and that'll have the desired effect. If we wanted to enforce that ourselves, we could compare `CMAKE_SOURCE_DIR` to `PROJECT_SOURCE_DIR` and conditionally add the `EXCLUDE_FROM_ALL` property to internal targets.

ryanvolz (approved):
This contains some whitespace changes, but since this is touching a lot of lines already I think it's an appropriate time to get those in. I, at least, will be thankful that `GrPybind.cmake` will no longer have that trailing whitespace.

marcusmueller (approved):

marcusmueller:
@wgledbetter Thanks! I *really* like that example, because, including CMake, that's a GNU Radio C++ tool in one file :) So, thank you for the effort you put in.

Perspectively, I'd guess I see the main use case being the situation where you want to ship a specific (potentially patched) version of GNU Radio with your application's source code. Neat!

Issue #6035, opened 2022-07-26
4.0: Move to C++20


mormj:
### Feature Description

By the time 4.0 comes out, all the major distributions should have solid C++20 support.

One particular feature that might be of use is atomic_flags - this might cut down on the mutex waiting time in the concurrent queue in the scheduler.

https://modernescpp.com/index.php/performancecomparison-of-condition-variables-and-atomics-in-c-20

### Feature Urgency

medium (would be nice to have in the near future)

marcusmueller:
Oh I'd love to think we'd have `std::span`; that's really just a thin and cheap wrapper around a contiguous piece of memory, but it would be super handy to pass around parts of e.g. a ring buffer, or a vector. We could extend e.g. `block_work_input` like

```C++
class block_work_input
{
// …
public:
// …
std::span<T> data()
{
    return {static_cast<T>(buffer->read_ptr()), n_items};
}
}
```
so that `for(auto sample : work_io.inputs()[0].data()) {/*…*/}` works.

Then, the whole std::ranges / std::view thing, where I can lazily define the operations to be done on some input range (like a vector, span, initializer list), and then only when I access the result it is computed:

```C++
namespace views = std::views;
auto squared_positives =
    (std::span(work_io.inputs()[0].items(), work_io.inputs()[0].n_items) |  // pipes - such UNIX, much wow
     views::filter([](auto v) { return v > 0; }) |   // filter first
     views::transform([](auto v) { return v * v; })  // square after
    );
```

marcusmueller:
compilation with std=c++20 fails – in logging. I know whom to beat up over that! (in other words, I'll fix that.)

mormj:
Fixed with #6050 - thanks @marcusmueller

PR #6050, opened 2022-08-05
C++20: Fix logging, set std to C++20

marcusmueller:
# Pull Request Details

## Description

In prep for C++20ification, overhauled the logging header: we should actually be doing perfect forwarding. Don't remember why I didn't do that on main – might need to backport.

## Related Issue

#6035 

I hoped this would solve the "spdlog fails with C++20", alas it didn't. However, this is still cleaner, and cleans up the API by removing levels that we were dragging along from 3.7 that are in basically no use.

Lögging.

## Checklist

### DIFF
--- a/blocklib/fileio/file_source/file_source_cpu.cc
+++ b/blocklib/fileio/file_source/file_source_cpu.cc
@@ -200,12 +200,12 @@ void file_source_cpu::open(const std::string& filename,
         if (file_size && file_size != INT64_MAX) {
             if (auto ret = posix_fadvise(
                     fd, start_offset, file_size - start_offset, POSIX_FADV_SEQUENTIAL)) {
-                d_logger->warn("failed to advise to read sequentially, " +
+                d_logger->warn("failed to advise to read sequentially, {}",
                                fadv_errstrings.at(ret));
             }
             if (auto ret = posix_fadvise(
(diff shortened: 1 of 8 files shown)


marcusmueller:
ah we need modern fmt, so ubuntu 20.04 fails.

@mormj should I include spdlog/fmtlib as subproject (whatever of the two suffices to make Ubuntu 20.04 build)? What's the oldest distro we need to support?

mormj:
We can probably bump up to 22.04 for simplicity. There are some c++20 features that are not supported by the gcc in 20.04. 

But as far as subprojects thats the path i took for cppzmq. For header libs it seems to work fine.

mormj:
Re-ran the CI because failures were because of some PMT changes.  I'm ok ignoring the Ubuntu 20.04 failure and removing 20.04 as a worker if we are going to run into more of these issues.  Some other c++20 changes I'd like to do will exacerbate this and there is no real reason to stick with 20.04 as a minimum for a future release

mormj (approved):
I'll remove the Ubuntu 20 worker in a separate commit (and upgrade Ubuntu on my dev machine lol)

mormj (approved):
With Ubuntu 20.04 out of the way, is this one good to go?

marcusmueller:
Hey sorry to keep you hanging there (in the middle of moving flats), yes, it is.

Issue #6052, opened 2022-08-07
Build failures with libfmt-9.0.0: static assertion failed: Cannot format an argument


vowstar:
### What happened?

After upgrading libfmt-9.0.0 and then compiling the latest version of gnuradio [v3.10.3.0](https://github.com/gnuradio/gnuradio/releases/tag/v3.10.3.0) , the compilation fails with: static assertion failed: Cannot format an argument

### System Information

OS: Linux Distro: Gentoo Linux
GR Installation Method: Source from github https://github.com/gnuradio/gnuradio/releases/tag/v3.10.3.0

### GNU Radio Version

3.10 (maint-3.10)

### Specific Version

3.10.3.0

### Steps to Reproduce the Problem

git clone from https://github.com/gnuradio/gnuradio

```bash
cd gnuradio
mkdir build && cd build
cmake ..
make -j
```

### Relevant log output

```shell
https://bugs.gentoo.org/858659


In file included from /usr/include/spdlog/fmt/fmt.h:27,
                 from /usr/include/spdlog/common.h:45,
                 from /var/tmp/portage/net-wireless/gnuradio-3.10.3.0/work/gnuradio-3.10.3.0/gnuradio-runtime/lib/../include/gnuradio/logger.h:36,
                 from /var/tmp/portage/net-wireless/gnuradio-3.10.3.0/work/gnuradio-3.10.3.0/gnuradio-runtime/lib/../include/gnuradio/custom_lock.h:15,
                 from /var/tmp/portage/net-wireless/gnuradio-3.10.3.0/work/gnuradio-3.10.3.0/gnuradio-runtime/lib/../include/gnuradio/buffer.h:15,
                 from /var/tmp/portage/net-wireless/gnuradio-3.10.3.0/work/gnuradio-3.10.3.0/gnuradio-runtime/lib/../include/gnuradio/buffer_double_mapped.h:15,
                 from /var/tmp/portage/net-wireless/gnuradio-3.10.3.0/work/gnuradio-3.10.3.0/gnuradio-runtime/lib/../include/gnuradio/io_signature.h:17,
                 from /var/tmp/portage/net-wireless/gnuradio-3.10.3.0/work/gnuradio-3.10.3.0/gnuradio-runtime/lib/../include/gnuradio/basic_block.h:15,
[...]

PR #6053, opened 2022-08-07
Fix build failures with libfmt-9.0.0

vowstar:
---
name: Pull Request Template
title: ''
labels: ''
assignees: ''

---

# Pull Request Details

## Description

libfmt-9.0.0 require explicit tagging of ARG types. This commit
adds those tags for the types that are currently formatted via fmt.
Detailed issue: https://github.com/gnuradio/gnuradio/issues/6052

Bug: https://bugs.gentoo.org/858659

## Related Issue

Fixes: https://github.com/gnuradio/gnuradio/issues/6052

Build

## Testing Done

After this patch, this issue has been fixed to compile and run correctly.

## Checklist


willcode:
Is there a way to centralize this? It could get out of hand if every block eventually needs this.

vowstar:
> Is there a way to centralize this? It could get out of hand if every block eventually needs this.

One way would be to move these things to something like [gnuradio-runtime/include/gnuradio/logger.h](https://github.com/gnuradio/gnuradio/commit/4cd04154aaaa19419ec3410f024d6e935e0f7eb9#diff-a2cfba3b921585e0b89fb33857a4592753f49ba939cb33808b35aa1b34414f06L276-L287), but the consequence would be to cause ``gnuradio-runtime/include/gnuradio/logger.h`` to refer to all associated types of headers.

mormj:
I think putting these definitions in various places would become unwieldy. I would prefer to address this by not passing pointers of types not supported by libfmt into the logging functions.  E.g. instead of 
```
d_logger->debug("block {}", block)
```
do 
```
d_logger->debug("block {}", block.identifier())
```

vowstar:
> I think putting these definitions in various places would become unwieldy. I would prefer to address this by not passing pointers of types not supported by libfmt into the logging functions. E.g. instead of
> 
> ```
> d_logger->debug("block {}", block)
> ```
> 
> do
> 
> ```
> d_logger->debug("block {}", block.identifier())
> ```

Done.

Thanks for your help, I modified the input parameters not supported by libfmt according to your suggestion.

vowstar:
> I think putting these definitions in various places would become unwieldy. I would prefer to address this by not passing pointers of types not supported by libfmt into the logging functions. E.g. instead of
> 
> ```
> d_logger->debug("block {}", block)
> ```
> 
> do
> 
> ```
> d_logger->debug("block {}", block.identifier())
> ```

All done and test passed

gnuradio-runtime/lib/flat_flowgraph.cc
```diff
@@ -411,24 +411,31 @@ std::string flat_flowgraph::msg_edge_list()
...
 
         int ni = detail->ninputs();
         int no = detail->noutputs();
         for (int i = 0; i < no; i++) {
             buffer_sptr buffer = detail->output(i);
-            d_logger->info("   output {:d}: {}", i, buffer);
+            d_logger->info(
+                "   output {:d}: {}",
+                i,
+                static_cast<std::stringstream&&>(std::stringstream() << buffer).str());
```
willcode, line 428: Traced this through to `operator<<` in `buffer.cc` and it looks sensible. Could you just verify that this previously printed that content and not just a pointer?

gnuradio-runtime/lib/flat_flowgraph.cc
```diff
@@ -64,10 +64,10 @@ void flat_flowgraph::setup_connections()
     // Connect message ports connections
     for (msg_edge_viter_t i = d_msg_edges.begin(); i != d_msg_edges.end(); i++) {
         d_debug_logger->debug("flat_fg connecting msg primitives: ({}, {})->({}, {})\n",
-                              i->src().block(),
-                              i->src().port(),
```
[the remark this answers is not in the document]
vowstar: if changed ``pmt::write_string(i->src().port())`` to ``std::to_string(i->src().port())`` will caused 
```
flat_flowgraph.cc:68:45: error: no matching function for call to ‘to_string(pmt::pmt_t)’
   68 |                               std::to_string(i->src().port()),
```
The ``std::to_string`` does not recognize ``pmt::pmt_t``, resulting in a compilation error.

gnuradio-runtime/lib/flat_flowgraph.cc
vowstar: I was also confused here at first, because the other place with @([this code](https://github.com/gnuradio/gnuradio/pull/6053/files#diff-c21cea5b87f37199c00e6b3de48931e5dd3704d4a98250b7f59e896bf100305bR419)) prints a pointer. Later, after careful analysis and debugging, it was found that the previous behavior here was indeed printing the buffer content.

willcode: Was it printing buffer content, or the result of `operator<<`?

gnuradio-runtime/lib/flat_flowgraph.cc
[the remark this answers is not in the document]
willcode: OK, got it. I have no idea why we're returning a pmt from `port()` but no something to deal with in this PR.

gnuradio-runtime/lib/flat_flowgraph.cc
vowstar: Ah, sorry for the inaccuracy described earlier, its behavior is to print the result of ``operator<<``
E.g.:
```
  sz:    <size>
  nrdrs: <size of reader>
    rd_idx: <read index>
    abs_rd_offset: <read offset>
    rd_idx: <read index>
    abs_rd_offset: <read offset>
    rd_idx: <read index>
    abs_rd_offset: <read offset>
    rd_idx: <read index>
    abs_rd_offset: <read offset>
...
    rd_idx: <read index>
    abs_rd_offset: <read offset>
```

PR #6055, opened 2022-08-08
ci: add E275 to ignored error list

mormj:
Python formatting check stopped working all of a sudden.


willcode:
If we ever do another Great Reformatting, E275 is a good one to put back in.

PR #6056, opened 2022-08-08
4.0: Basic reflection capability

mormj:
Rebase of #6018 

Create a GR registry that maintains knowledge about what modules,
blocks, etc. are available, and has the ability to return generic block
constructors.

Basic usage would be
```
auto list = gr::registry::modules()
// "math" should be in list
list = gr::registry::blocks("math");
// "vector_source_f" should be in list
auto f = gr::registry::factory("math","vector_source_f")
auto blk = f(.... map of string/pmt)
```

Right now, this is all done through auto code generation, with the need being 
`.yml` --> headers, then headers should include reflection capability

This adds a generic factory to each block that can be invoked with `std::map<std::string, pmtf::pmt>' - which is a list of all the parameters in the block.  Since some parameters are not part of the constructor, these will need to be handled through the setters.

There is probably a more elegant solution using `refl-cpp` which can then be wrapped for nicer python bindable API, but this first pass is pretty brute force just generate the code.  I still question whether some of these things can be done with a general purpose tool like refl-cpp, such as some parameters being part of the constructor and some being setters

Still needing to be addressed:

- [ ] python bindings of `registry.h`
- [ ] block parameters and callbacks
- [ ] rest of the things in `block.yml`
- [ ] parameters that are not part of constructor
- [ ] schedulers, buffers, other modular components

### DIFF
--- a/blocklib/blocks/vector_sink/vector_sink_cpu.h
+++ b/blocklib/blocks/vector_sink/vector_sink_cpu.h
@@ -29,7 +29,7 @@ public:
     // private member variable, and pass it out as pmt when queried
     void on_parameter_query(param_action_sptr action) override
     {
-        this->d_debug_logger->debug(
+        this->d_debug_logger->trace(
             "block {}: on_parameter_query param_id: {}", this->id(), action->id());
         pmtf::pmt param = d_data;
         // auto data = pmtf::get_as<std::vector<float>>(*param);
--- a/blocklib/soapy/rtlsdr_source/rtlsdr_source.yml
+++ b/blocklib/soapy/rtlsdr_source/rtlsdr_source.yml
@@ -51,32 +51,32 @@ parameters:
       settable: true
     - id: center_freq
       label: "Center Freq (Hz)"
-      category: RF Options
       dtype: rf32
       grc:
+          category: RF Options
           default: "freq"
       settable: true
     - id: gain
       label: "RF Gain"
-      category: RF Options
       dtype: rf32
       grc:
+          category: RF Options
           default: "20"
           hide: ${'all' if agc else 'part'}
       settable: true
     - id: freq_correction
       label: "Frequency Correction (PPM)"
-      category: RF Options
       dtype: rf32
       default: 0
       grc:
+          category: RF Options
           hide: "part"
     - id: agc
       label: "AGC"
-      category: RF Options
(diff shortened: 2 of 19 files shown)


mormj:
Recent updates here allow the block parameters to be queried from the register:
```c++
auto list = gr::registry::parameters("math", "multiply_const_ff");
auto p = gr::registry::parameter("math", "multiply_const_ff", "k");
```
where p is a struct whose fields match the yaml

Currently not satisfied with 2 things in here:
1) It is very brute force
2) The defaults from the yaml are now specified in multiple places - this is because the json_default piece is not working correctly

RalphSteinhagen:
@mormj thanks a lot for drafting and implementing this GR reflection proposal! :+1: 

For our [our purposes](https://github.com/fair-acc/gr-digitizers/issues/42) this appears to be sufficient, helps, and once merged would simplify our user code a lot!

I primarily looked at the user-API side. Everything needed to initialise a GRC-based flow graph during run-time and only using C++ seems to be there. The unit tests are also quite helpful.

Regarding the library implementation itself:
  * Deriving the RTTI block information through the same YML-file definitions that drive the C++-Python bindings and thus re-using parts of the existing Python-based code generation is a fair and practical solution.
  * Using a singleton pattern (while often frowned upon) for the `gr::registry` seems to be also a valid use case for this particular application. I looked into alternatives and haven't found any other viable/better alternative myself -- short of using C++ compile-time reflection (only targeted for C++26), and interim solutions like, for example, [refl-cpp](https://github.com/veselink1/refl-cpp) which also uses singletons for storing this type of information globally.
  * I cannot assess the Python templates other than that they work as advertised since I am not familiar enough with that topic. 

Thanks again for your proposal and please accept my sincere apologies for the late feedback. Once this is merged we can finally gracefully retire our own [workaround implementation](https://github.com/fair-acc/gr-flowgraph). 

As discussed with you and @dkozel earlier: this makes now indeed a compelling case for us to skip 3.10 entirely and migrate our OOT eco-system directly to GR 4.0! Thanks again to you and others involved for being forthcoming.

mormj:
Thanks for the evaluation Ralph!  Rebased and re-pushed to re-run CI.  If it all passes, I'll merge and then handle additional features and improvements in separate PRs

PR #6057, opened 2022-08-08
4.0: meson: check python module dependencies

mormj:
As brought up by @marcusmueller in chat, there is currently no way in meson to check versions of python deps.  This is a workaround for our limited set of deps by routing the import and version check to an external python script

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -33,6 +33,16 @@ if py3_version.version_compare('< 3.6')
   error('Invalid python version!?')
 endif
 
+# Python module min versions
+py_check_script = join_paths(meson.project_source_root(),'utils','py_module_check.py')
+run_command('python3', py_check_script, 'numpy', '1.22', check:true)
+run_command('python3', py_check_script, 'jinja2', '3.0.2', check:true)
+run_command('python3', py_check_script, 'yaml', '5.3.1', check:true)
+run_command('python3', py_check_script, 'zmq', '22.3.0', check:true)
+run_command('python3', py_check_script, 'jsonschema', '3.2.0', check:true)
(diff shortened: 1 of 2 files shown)


ryanvolz:
From a packaging perspective, it would also be nice to be able to turn these checks off so that runtime dependencies don't have to be installed at build time. That's a small hassle typically, but it can be a larger hassle for situations like cross-compilation (in which case these particular checks would be pointlessly checking the build environment instead of the target environment). Even better than just turning it off would be to have a runtime test that checks these dependencies. And I'm not saying that you shouldn't merge this; I just wanted to put the idea out there.

eli-schwartz:
> there is currently no way in meson to check versions of python deps

To clarify, there's also no way to check names of python deps in Meson either.

Meson supports checking the names of python importable modules -- that's not necessarily the same as package names, though it's often similar.

mormj:
@eli-schwartz - what is the preferred way in meson to check python importable modules?

eli-schwartz:
```diff
-py3_inst = import('python').find_installation('python3')
+py3_inst = import('python').find_installation('python3', modules: ['importname1', 'importname2'])
```

Checking package names and versions would likely tie into future projected enhancements around the creation of python dist-info and wheels. The main issue is that `importlib.metadata` is the correct way to do this, but it requires Meson to drop support for python < 3.8 (currently we support 3.7, and it will be dropped when it is EOL.)

mormj:
>  it would also be nice to be able to turn these checks off so that runtime dependencies don't have to be installed at build time

I think this is very doable - will adjust the script accordingly to specify

mormj:
Closing this PR, as most of it was included in #6171

Issue #6086, opened 2022-08-13
Is async crc16 the right endianness?


michael-roe:
### What happened?

I put the following PDU into an Async CR16 block:
0x64 0x07 0x42 0x00

and got the following output:
0x64 0x07 0x42 0x00 0xdd 0xf9

But ... online CRC calculators for CRC16-CCITT-FALSE (polynomial x16+x12+x5+1,  initial value 0xffff, xorout 0x0000) give a CRC of 0xf9 dd.

.. which is the opposite endianness. Is it the wrong way round?

======

I have a second issue with this block, which is that the CRC I really wanted was x16+x12+x5+1, initial value 0x0, xorout 0xffff, and the block isn't configurable enough to match the diverse range of CRC variants we find in real radio protocols. (This CRC is used by Yaesu System Fusion).

### System Information

OS: Linux
GR Installation Method: Built from guthub source

### GNU Radio Version

3.11-git (main)

### Specific Version

0.0.0.0

### Steps to Reproduce the Problem

-

michael-roe:
from crc16_async_bb_impl.cc, line 60:

memcpy((void*)(bytes_out.data() + pkt_len), &crc, 2);

That's an endianness bug, isn't it? (Computer running the code is probably a little endian x86 architecture machine, network endian is big endian).

michael-roe:
qa_crc16_async_bb will also need fixing if we change the block to fix endianness.

marcusmueller:
fixed as per the above PRs

PR #6096, opened 2022-08-15
Fix CRC16 endianness

michael-roe:
Signed-off-by: Michael Roe <michael-roe@users.noreply.github.com>

Fixes issues #6086
The CRC had the endianness of the host computer -- typically little endian -- rather than network (big) endian.

Modules affected: gr-digital/lib/crc16_async_bb_impl.cc, qa_crc16_async_bb

Tests carried out: "make test" with the new qa_crc16_async_bb

- [X] I have updated [the documentation](https://wiki.gnuradio.org/index.php/Main_Page#Documentation) where necessary.

### DIFF
--- a/gr-digital/lib/crc16_async_bb_impl.cc
+++ b/gr-digital/lib/crc16_async_bb_impl.cc
@@ -57,7 +57,8 @@ void crc16_async_bb_impl::calc(pmt::pmt_t msg)
 
     crc = d_crc_ccitt_impl.compute(bytes_in, pkt_len);
     memcpy((void*)bytes_out.data(), (const void*)bytes_in, pkt_len);
-    memcpy((void*)(bytes_out.data() + pkt_len), &crc, 2);
+    bytes_out.data()[pkt_len] = (crc >> 8) & 0xff;
+    bytes_out.data()[pkt_len + 1] = crc & 0xff;
 
     pmt::pmt_t output = pmt::init_u8vector(
         pkt_len + 2,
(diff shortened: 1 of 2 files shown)


willcode:
How do we determine whether this is compatible with other/legacy apps?

michael-roe:
> How do we determine whether this is compatible with other/legacy apps? It won't be compatible with prior versions of GR.

I was worrying someone would say that ... that the bug has now effectively become part of the specification now that we've released a version of gnuradio containing it.

Personally, I doubt that compatabiliity with old versions of GR is much of an issue with this block. I am more concerned with interoperability with standardized radio protocols that use CRC16 (with the right endianness).

If people really want backlwards compatability, I am open to making the CRC endianess a parameter to the block.

Or we could create an entirely new async CRC block that is just like the old one, only without the endianness bug.

willcode:
A whole separate block would confuse people, so if it's determined that compatibility is important, an endianness param would be good. I don't know if that's required, though. I'll see if anyone on chat has opinions.

willcode:
Note that there are separate CRC append and check blocks, which allow the polynomial, size, and other params to be specified. The CRC16 and CRC32 block yaml in GRC should probably just produce code that uses the more general blocks.

michael-roe:
Ok, so the more generic "CRC Check" block will do what I want, with appropriate parameters.

I still think crc16_async_bb is buggy. It will produce different results when run on  a big-endian machine, and a 16 vbit CRC with the little end first is almost certainly not what the user wants.

willcode:
@argilo concurs on chat with fixing endianness and also mentioned deprecating the CRC16 and CRC32 blocks. We probably need to make the general CRC blocks easier to use, and possibly make the CRC16/32 yml wrap the general blocks as part of the deprecation plan.

argilo (approved):
As mentioned on Matrix chat, I think it's OK to break backwards compatibility in this instance. I doubt the block was used much, since it was recently added (in #4649) and until even more recently had a serious bug (fixed in #5830).

mormj (approved):

willcode:
Calling this one a bug fix, so backporting unless there are any objections.

Issue #6105, opened 2022-08-22
Regression: Blocks with history lose samples on top_block::lock()/top_block::unlock()


vladisslav2011:
### What happened?

Any GNU Radio block with history loses some samples on top_block::lock() followed by top_block::unlock().
Problematic commit: 788827ae116bef871e144abd39b1e4482208eabe

### System Information

OS: Linux
GR Installation Method: Source

### GNU Radio Version

3.11-git (main)

### Specific Version

3.10.3.0

### Steps to Reproduce the Problem

1. Clone GNU Radio repository
2. Check out main
3. Compile and install
4. Compile test program (cpp and makefile below)
5. Run

main.cpp
```
#include <iostream>
#include <cstdint>
#include <gnuradio/block.h>
#include <gnuradio/sync_block.h>
#include <gnuradio/top_block.h>
#include <gnuradio/io_signature.h>

#if GNURADIO_VER == 38
#define GRSPTR boost::shared_ptr
#elif GNURADIO_VER == 37
#define GRSPTR boost::shared_ptr
#else
#define GRSPTR std::shared_ptr
#endif


class cnt_source : virtual public gr::sync_block
{
public:
    typedef GRSPTR<cnt_source> sptr;
    static sptr make(size_t nitems)
    {
        return gnuradio::get_initial_sptr(new cnt_source(nitems));
    }
    cnt_source(size_t nitems)
    : sync_block("cnt_source",
                 gr::io_signature::make(0, 0, 0),
                 gr::io_signature::make(1, 1, sizeof(uint32_t))),
                 d_cnt(0),
                 d_nitems(nitems)
    {
    }
    virtual ~cnt_source() {};
    int work(int noutput_items,
                            gr_vector_const_void_star& input_items,
                            gr_vector_void_star& output_items)
    {
        uint32_t* o = (uint32_t*)output_items[0];
        if(d_cnt >= d_nitems)
            return -1;
        for (size_t n = 0; n < noutput_items; n++)
        {
            if(d_cnt == d_nitems)
                return n;
            o[n] = d_cnt;
            d_cnt++;
        }
        return noutput_items;
    }
    void reset_counter()
    {
        d_cnt = 0;
    }
    uint32_t get_counter()
    {
        return d_cnt;
    }
private:
    uint32_t d_cnt;
    uint32_t d_nitems;
};

class cnt_sink : virtual public gr::sync_block
{
public:
    typedef GRSPTR<cnt_sink> sptr;
[...]

PR #6108, opened 2022-08-22
Fix sample dropouts on top_block::unlock()

vladisslav2011:
## Description
Locking and unlocking the top_block (even without actual reconfiguration) results in a huge sample loss if any block have large history.
This PR fixes this issue by not resetting `buffer_reader::d_read_index` and `buffer::d_write_index` when
resetting absolute offsets.

## Related Issue
Fixes https://github.com/gnuradio/gnuradio/issues/6105

Any block with large history (FIR filters, delays, etc)

## Testing Done
Added a test "t13_unlock_history" to gr-blocks/lib/qa_gr_top_block.cc
The test is kept in a separate commit to make it possible to cherry-pick it onto the branch in question and confirm, that it is failing without a fix.

## Checklist

### DIFF
--- a/gr-blocks/lib/qa_gr_top_block.cc
+++ b/gr-blocks/lib/qa_gr_top_block.cc
@@ -16,11 +16,16 @@
 #include <gnuradio/blocks/nop.h>
 #include <gnuradio/blocks/null_sink.h>
 #include <gnuradio/blocks/null_source.h>
+#include <gnuradio/blocks/throttle.h>
+#include <gnuradio/blocks/vector_sink.h>
+#include <gnuradio/blocks/vector_source.h>
 #include <gnuradio/top_block.h>
 #include <boost/test/unit_test.hpp>
 #include <iostream>
+#include <thread>
 
 #define VERBOSE 1
+
 
 BOOST_AUTO_TEST_CASE(t0)
 {
@@ -312,4 +317,29 @@ BOOST_AUTO_TEST_CASE(t12_release_shared_pointers)
     BOOST_CHECK_EQUAL(1, src.use_count());
     BOOST_CHECK_EQUAL(1, head.use_count());
     BOOST_CHECK_EQUAL(1, dst.use_count());
+}
+
+BOOST_AUTO_TEST_CASE(t13_unlock_history)
+{
+    if (VERBOSE)
+        std::cout << "qa_top_block::t13()\n";
+
+    auto tb = gr::make_top_block("top");
+
+    std::vector<int32_t> data(1000000);
+    auto src = gr::blocks::vector_source<int32_t>::make(data, false, 1);
+    auto thr = gr::blocks::throttle::make(sizeof(int32_t), 1000000);
+    auto dst = gr::blocks::vector_sink<int32_t>::make(1, 0);
+
+    tb->connect(src, 0, thr, 0);
+    tb->connect(thr, 0, dst, 0);
+
+    tb->start();
+    while (!dst->data().size())
+        std::this_thread::yield();
+    tb->lock();
+    tb->unlock();
+    tb->wait();
+
+    BOOST_CHECK_EQUAL(1000000, dst->data().size());
 }
--- a/gnuradio-runtime/include/gnuradio/buffer.h
+++ b/gnuradio-runtime/include/gnuradio/buffer.h
@@ -126,11 +126,7 @@ public:
 
     uint64_t nitems_written() { return d_abs_write_offset; }
 
-    void reset_nitem_counter()
-    {
-        d_write_index = 0;
-        d_abs_write_offset = 0;
-    }
+    void reset_nitem_counter() { d_abs_write_offset = 0; }
 
     size_t get_sizeof_item() { return d_sizeof_item; }
 
--- a/gnuradio-runtime/include/gnuradio/buffer_reader.h
+++ b/gnuradio-runtime/include/gnuradio/buffer_reader.h
@@ -106,11 +106,7 @@ public:
 
     uint64_t nitems_read() const { return d_abs_read_offset; }
 
-    void reset_nitem_counter()
-    {
-        d_read_index = 0;
-        d_abs_read_offset = 0;
-    }
+    void reset_nitem_counter() { d_abs_read_offset = 0; }
 
     size_t get_sizeof_item() { return d_buffer->get_sizeof_item(); }
 
--- a/gnuradio-runtime/python/gnuradio/gr/bindings/buffer_python.cc
+++ b/gnuradio-runtime/python/gnuradio/gr/bindings/buffer_python.cc
@@ -14,7 +14,7 @@
 /* BINDTOOL_GEN_AUTOMATIC(0)                                                       */
 /* BINDTOOL_USE_PYGCCXML(0)                                                        */
 /* BINDTOOL_HEADER_FILE(buffer.h)                                                  */
-/* BINDTOOL_HEADER_FILE_HASH(8e570cc3747473c3774ae5ffef6653aa)                     */
+/* BINDTOOL_HEADER_FILE_HASH(291a90813b3621085dc1eb443ed3e6dc)                     */
 /***********************************************************************************/
 
 #include <pybind11/complex.h>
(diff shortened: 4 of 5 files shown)


vladisslav2011:
This PR may break something. Do not merge without extensive testing, please.

willcode:
The current behavior causes sample loss on lock/unlock. Are there cases where this is desirable? For instance, if the history changes, if there is a rate change somewhere? I can't think of a specific case at the moment. Agree that some thought should be given to this before merge.

vladisslav2011:
I'm thinking about custom buffers from 788827ae116bef871e144abd39b1e4482208eabe that may rely on counters being reset.
Is there any implementation of custom buffers (FPGA accelerated blocks)?

Ubuntu 20.04 CI fails with timeout. I'll setup an Ubuntu 20.04 chroot and try to fix it.

willcode:
There is a CUDA implementation of custom buffers. Not sure about a FPGA one.

vladisslav2011:
Freezing test is fixed (I hope).

willcode:
In `flat_flowgraph.cc`, there is
```
        // Now deal with the fact that the block details might have
        // changed numbers of inputs and outputs vs. in the old
        // flowgraph.
        block->detail()->reset_nitem_counters();
        block->detail()->clear_tags();
```
I'd really like to understand that comment before thinking about changing anything. What are all the things that can happen before a `restart()` that could interact with this change?

Also, why would the read/write indices and the absolute counters treated differently?

vladisslav2011:
The only problem, I see, is tag propagation, that does not work well in case of flowgraph reconfiguration either. Tags are just dropped, instead of being handled correctly.
I think, `buffer_reader` should not have absolute counter at all and should use it's parent's absolute write counter to calculate `nitmes_read()` and difference between read and write pointers. This should solve the problem with newly added `buffer_reader`s and newly added buffers too as `buffer_reader`s would use newly allocated buffers and, so, calculate sample offsets correctly.
I've opened an issue to discuss tag propagation problem: https://github.com/gnuradio/gnuradio/issues/6110

gr-blocks/lib/qa_gr_top_block.cc
```diff
@@ -16,12 +16,71 @@
...
+             gr_vector_void_star& output_items)
+    {
+        uint32_t* o = (uint32_t*)output_items[0];
+        uint32_t* i = (uint32_t*)input_items[0];
+        for (int n = 0; n < noutput_items; n++) {
+            o[n] = i[n + history()];
+        }
+        return noutput_items;
+    }
+};
```
argilo, line 81: I don't think it's necessary to create new blocks just to test this. Here's an example that reproduces the problem without any new blocks (and even without blocks that use history): https://github.com/gnuradio/gnuradio/issues/6105#issuecomment-1311872035

vladisslav2011: Thanks for an example of using vector sink to count samples. I'll update the test soon.

argilo: By the way, I edited the example after initially posting it. The only blocks needed are Vector Source -> Throttle -> Vector Sink. In fact the bug can even be demonstrated without the Throttle block, but adding it helps avoid the test consuming CPU unnecessarily.

vladisslav2011: Updated with test, based on python example.

gr-blocks/lib/qa_gr_top_block.cc
```diff
@@ -16,12 +16,20 @@
 #include <gnuradio/blocks/nop.h>
 #include <gnuradio/blocks/null_sink.h>
 #include <gnuradio/blocks/null_source.h>
+#include <gnuradio/blocks/throttle.h>
+#include <gnuradio/blocks/vector_sink.h>
+#include <gnuradio/blocks/vector_source.h>
+#include <gnuradio/io_signature.h>
+#include <gnuradio/sync_block.h>
```
argilo, line 23: These are not needed anymore.

gr-blocks/lib/qa_gr_top_block.cc
```diff
@@ -16,12 +16,20 @@
...
 #include <gnuradio/top_block.h>
 #include <boost/test/unit_test.hpp>
+#include <atomic>
```
argilo, line 26: This is not needed anymore.

gr-blocks/lib/qa_gr_top_block.cc
vladisslav2011: Thanks. Updated.

gr-blocks/lib/qa_gr_top_block.cc
vladisslav2011: Updated.

vladisslav2011:
I don't think, that this PR is the best possible solution. It just returns old behavior, that results in tags loss and tags going out of sync with sample offsets... And it does not prevent sample loss in case of hardware device as a source block. It should be better to change the reconfiguration logic to keep as much data in old buffers as possible, update read pointers and absolute offsets on newly connected readers, copy tails to resized buffers and prevent stopping and restarting the whole flowgraph to avoid calling `stop()`/`start()` on hardware sources/sinks during reconfiguration when possible. This may be achieved by adding a flag to a block, that indicates, that this block is a subject to reconfiguration. Then the scheduler may be asked to not call `work`/`general_work` on marked blocks and wait for all marked blocks' `work`/`general_work` to exit before returning to caller. After the reconfiguration the scheduler should be asked to rescan the flowgraph and continue calling `work`/`general_work` on all blocks or just resetting the flag may be enough to indicate that thip particular block has finished reconfiguration and is ready `work`.
Another problem: heavy block detail allocation is done after stopping the flowgraph. This results in long reconfiguration times and increased chances of sample loss. I think, we should allocate datails/buffers before stopping the flowgraph and do this only on blocks, that require this (newly added or requiring buffers reconfiguration - changed number of output buffers or changed buffer size to fit larger alignment/history).

Of course, we should always call `block::start()` when a new source/sink is connected and call `block::stop()` when an old source/sink is disconnected, but only on destroyed/created blocks.

gr-blocks/lib/qa_gr_top_block.cc
```diff
@@ -313,3 +318,28 @@ BOOST_AUTO_TEST_CASE(t12_release_shared_pointers)
...
+    tb->connect(thr, 0, dst, 0);
+
+    tb->start();
+    while (!dst->data().size())
+        std::this_thread::yield();
+    tb->lock();
+    tb->unlock();
+    tb->wait();
+
+    BOOST_CHECK((1000000 == dst->data().size()));
```
argilo, line 344:
```suggestion
    BOOST_CHECK_EQUAL(1000000, dst->data().size());
```

vladisslav2011: Thanks.

argilo:
> I don't think, that this PR is the best possible solution. It just returns old behavior, that results in tags loss and tags going out of sync with sample offsets

If this change breaks tag propagation, then maybe it is best not to merge it. Is it possible to eliminate dropped samples without breaking tags?

The other concerns in your comment sound like performance / optimization issues that could be addressed separately. But I think correctness (no dropped samples, proper tag propagation) is most important.

argilo:
Whether or not we make this change, the test will remain useful. Having similar tests for other behaviours (like correct tag propagation) would help ensure that this functionality does not break again in the future (once it is fixed). Flow graph reconfiguration has been very prone to bugs because it is a rarely used feature, and it is almost completely untested in CI.

vladisslav2011:
The tag propagation is broken either with this change or without it...
Tags are always dropped here https://github.com/gnuradio/gnuradio/blob/main/gnuradio-runtime/lib/flat_flowgraph.cc#L363

argilo:
OK. Does this change make the problem worse though? Or does it leave the tag situation the same?

vladisslav2011:
It is hard for me to tell, what is worse: to drop a large block of tags on each reconfiguration or have chances to drop a small block of tags on a reconfiguration and have a small tag offset after reconfiguration.
There is my attempt to fix tag propagation: https://github.com/gnuradio/gnuradio/pull/6113 But it requires more work. New single-mapped buffers are broken...
It may be better to finish #6113 instead of merging this quick fix...

argilo:
To me, introducing a tag offset sounds worse than dropping tags (even if the number of dropped tags is reduced).

I'll try to have a look at #6113 tomorrow.

willcode:
Closing for now due to lack of time.

Issue #6110, opened 2022-08-22
Tags loss on flowgraph reconfiguration


vladisslav2011:
### What happened?

Some tag are lost on flowgraph reconfiguration even on flowgraph branches, that are not changed.
Event `top_block::lock()` followed by `top_block::unlock()` results in tag loss.

### System Information

OS: Linux
GR Installation Method: Source

### GNU Radio Version

3.11-git (main)

### Specific Version

v3.10.0.0-rc1-263-g69c5d33b

### Steps to Reproduce the Problem

1. Clone GNU Radio repository
2. Check out main
3. Compile and install
4. Compile test program (cpp and makefile below)
5. Run

main.cpp
```
#include <iostream>
#include <cstdint>
#include <gnuradio/block.h>
#include <gnuradio/sync_block.h>
#include <gnuradio/top_block.h>
#include <gnuradio/io_signature.h>

#if GNURADIO_VER == 38
#define GRSPTR boost::shared_ptr
#elif GNURADIO_VER == 37
#define GRSPTR boost::shared_ptr
#else
#define GRSPTR std::shared_ptr
#endif


class cnt_source : virtual public gr::sync_block
{
public:
    typedef GRSPTR<cnt_source> sptr;
    static sptr make(size_t nitems)
    {
        return gnuradio::get_initial_sptr(new cnt_source(nitems));
    }
    cnt_source(size_t nitems)
    : sync_block("cnt_source",
                 gr::io_signature::make(0, 0, 0),
                 gr::io_signature::make(1, 1, sizeof(uint32_t))),
                 d_nitems(nitems)
    {
    }
    virtual ~cnt_source() {};
    int work(int noutput_items,
                            gr_vector_const_void_star& input_items,
                            gr_vector_void_star& output_items)
    {
        uint32_t* o = (uint32_t*)output_items[0];
        if(d_cnt >= d_nitems)
            return -1;
        for (size_t n = 0; n < noutput_items; n++)
        {
            if(d_cnt == d_nitems)
                return n;
            o[n] = d_cnt;
            if(d_tag_interval && (d_cnt % d_tag_interval == 0))
            {
                add_item_tag(0, nitems_written(0) + n, pmt::intern("cnt_tag"), pmt::from_long(d_cnt));
                d_tags++;
            }
            d_cnt++;
        }
        return noutput_items;
    }
    void reset_counters()
    {
        d_cnt = 0;
        d_tags = 0;
    }
    uint32_t get_counter()
    {
[...]

PR #6113, opened 2022-08-23
Fix incorrect tag propagation and loss on flowgraph reconfiguration

vladisslav2011:
## Description
Do not use `buffer_reader` internal counter to calculate `nitems_read`
Correctly recalculate tag offsets taking into account possible rate changes.

## Related Issue
Fixes https://github.com/gnuradio/gnuradio/issues/6110 and maybe https://github.com/gnuradio/gnuradio/issues/6105
All blocks that create/handle tags.

## Testing Done
TBD.

## Checklist

### DIFF
--- a/gnuradio-runtime/include/gnuradio/buffer_reader.h
+++ b/gnuradio-runtime/include/gnuradio/buffer_reader.h
@@ -104,13 +104,9 @@ public:
 
     gr::thread::mutex* mutex() { return d_buffer->mutex(); }
 
-    uint64_t nitems_read() const { return d_abs_read_offset; }
+    uint64_t nitems_read();
 
-    void reset_nitem_counter()
-    {
-        d_read_index = 0;
-        d_abs_read_offset = 0;
-    }
+    void reset_nitem_counter() { d_read_index = 0; }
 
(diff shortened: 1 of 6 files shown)


willcode:
Copying tags to propogate them probably isn't going to help performance.

vladisslav2011:
Agree. It may be better to check if the tag offset needs to be changed and copy the tag in this case only.
Anyways this fix needs to be adapted to single-mapped buffers, but, as single mapped buffers do not follow "guard sample" convention https://github.com/gnuradio/gnuradio/blob/main/gnuradio-runtime/lib/buffer_double_mapped.cc#L155-L157 and are using absolute offsets to find out if the buffer is full or empty in case of equal read and write pointers, it may require some changes to single-mapped buffers logic.

As single-mapped buffer logic looks slightly overengineered to me, any comments and suggestions from @dsorber are welcome.

Out-of-tree tests are here: https://github.com/vladisslav2011/gnuradio-buffer-test
I'll add tests with blocks that add some sample delay (with `declare_sample_delay()` and without) and resampling blocks (fixed and variable resampling ratio) as primary processing block soon.

willcode:
Closing for now due to lack of time.

PR #6114, opened 2022-08-24
file source: update d_items_remaining on seek

willcode:
## Description
d_items_remaining was not being updated during seek, so end/repeat
of playback would not happen at the correct time.

File Source

## Testing Done
Problem found and tested using Gqrx playback.

## Checklist


willcode:
I was having trouble thinking of a good way. You'd have to run a flowgraph with something like a 10 sec long source file with repeat turned off, do a couple of seeks, and verify that the flowgraph ended after a certain amount of time. Pretty interactive and not so foolproof.

mormj:
Yeah, seems more error prone than it is worth

mormj (approved):

PR #6117, opened 2022-08-28
gr-soapy: Add rx_time_ns tags to source blocks

### LINKED ISSUE #6106: gr-soapy: Add timestamp tag to source output
### Feature Description

SoapySDR will provide the hardware timestamp of the received buffer in the form of nanoseconds. gr-soapy's source block discards this timetamp `time_ns` when outputting items. 

https://github.com/gnuradio/gnuradio/blob/6c34c6bb795639a7f85ed74421ef1316999da4b8/gr-soapy/lib/source_impl.cc#L51-L97

Can we add a time tag to the output much like gr-uhd does (and possible the frequency and rate as well?)
https://github.com/gnuradio/gnuradio/blob/1dcf8740f4358edee5a6fda7fabd97907817c217/gr-uhd/lib/usrp_source_impl.cc#L623-L630

### Feature Urgency

### More Information

_No response_

NZSmartie:
## Description
Tag the output stream with the hardware based timestamp if populated by the SoapySDR driver.

## Related Issue
Fixes #6106 

gr-soapy - Unsure of performance impact but assumed to be minimal since gr-uhd creates three tags per stream for frequency, time and sample rate. 

## Testing Done
Created a a source gr-soapy source block for the LimeSDR driver. Plugged it into a Tag Debug sink block. Configured the sample rate, frequency and bandwidth and ran the flow graph. the output contained the following snippet:
```
Tag Debug:
Input Stream: 00
  Offset: 200680  Source: n/a     Key: rx_time_ns   Value: 2006800000
```

## Checklist


willcode (changes requested):
Since gr-uhd has time tags, the gr-soapy behavior should match if possible. gr-uhd uses tags `rx_time`, `rx_rate` and `rx_freq`, so `rx_time` would be appropriate here.

Adding a tag to every `work()` call is not necessary. Take a look at how gr-uhd handles this https://github.com/gnuradio/gnuradio/blob/9a698313235c03c250da8444742ee80c24b3e5ad/gr-uhd/lib/usrp_source_impl.cc#L615. A single tag is sufficient as long as the sample stream is continuous. Whenever the stream might be interrupted, new tag(s) can be sent. gr-ud does this for overflows, sample rate changes, frequency changes, etc.

Whether timeNs means anything depends on the Soapy driver. For example, it appears that the SoapyRTLSDR driver initializes time to 0, then increments timeNs based on the sample rate, and ignores any discontinuities. Would GNU Radio consumers of this information need to know whether the time value was actually valid data?

willcode:
@ncorgan What are your thoughts on how best to use SoapySDR's timeNS in gr-soapy?

NZSmartie:
Just realising now, with the `cmd_handler_frequency`, `cmd_handler_hardware_time` and `cmd_handler_samp_rate` happening in `block_impl` it might be worth moving them there.

willcode:
Looks like this is on the right track. I haven't looked carefully to see what should be moved `block_impl`, but anything that can be made common would be great.

SoapySDR (and gr-soapy) support multi-channel sources. As in gr-uhd, tags should be placed on all streams, not just stream 0.

willcode (approved):
I think this is good to go on main so people can test it. I'm not sure we'll put this in the September 3.10 release. Probably December.

willcode:
@ncorgan if you get a chance, could you take a quick look?

PR #6147, opened 2022-09-07
GR 4.0: Add numpy include path to python bindings meson files

wirew0rm:
On some systems (e.g. openSuse and nixOS) python packages are split between different install directories. This leads to include errors where includes like `#include <numpy/arrayobject.h>` cannot be resolved.

This PR fixes this by determining the numpy include path in the main meson.build file and adding it manually to all the relevant meson files as well as the j2 template for the autogenerated block files.

The approach is inspired by the numpy documentation for using numpy from meson [1].

We have tested these changes to build on nixOS unstable and openSUSE, but I do not have an ubuntu or fedora machine which seem to be the platforms that were tested to work before. I do not see why this should break builds there, though and I'll trust in the CI.

I'm also new to meson, but I have some experience with other buildsystems and linux packaging. While this approach works, it relies on importing the installed package at build time, which might be unwanted by some packagers (https://github.com/gnuradio/gnuradio/pull/6057#issuecomment-1208433867) and also might return wrong paths for more complicated situations like cross compiling. Ideally we would specify `dependency = [ numpy ... ] ` and have meson figure out the correct compiler arguments instead of manually adding paths. As I understand from the linked issue this is not modeled in meson at least at the moment.

[1] https://numpy.org/devdocs/f2py/buildtools/meson.html

### DIFF
--- a/blocklib/filter/python/filter/meson.build
+++ b/blocklib/filter/python/filter/meson.build
@@ -30,15 +30,15 @@ gen_filter_pybind = custom_target('gen_filter_pybind',
                             '--blocks', d['blocks'],
                             '--imports', d['imports'],
                             '--module', d['module'],
-                            '--output_file', '@OUTPUT@', 
+                            '--output_file', '@OUTPUT@',
                             '--build_dir', join_paths(meson.build_root())],
-                        install : false)      
+                        install : false)
 
 filter_pybind_sources += gen_filter_pybind
 
 gnuradio_blocklib_filter_pybind = py3_inst.extension_module('filter_python',
-    filter_pybind_sources, 
-    include_directories: ['../../lib'],
+    filter_pybind_sources,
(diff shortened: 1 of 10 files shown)


mormj (approved):
Thank you for providing this PR and the caveats with packaging concerns.  This will be good to address / keep an eye on.  Going ahead and merging to get things running with these other OSes

eli-schwartz:
> On some systems (e.g. openSuse and nixOS) python packages are split between different install directories. This leads to include errors where includes like `#include <numpy/arrayobject.h>` cannot be resolved.

Note the real issue is that the upstream software such as numpy has private install directories, and some systems like Ubuntu have unofficial symlinks.

Anyway, the general tracking ticket for this is https://github.com/mesonbuild/meson/issues/9598

PR #6151, opened 2022-09-08
runtime: make debug logger have separable options

mormj:
## Description
Currently, the logging config looks like
```yaml
[LOG]
log_level = emerg
debug_level = debug
log_file = stdout
debug_file = stderr
```
implying that the debug logger can be logged as specified.  Unfortunately it always gets logged to the console

Underlying logger in all blocks

## Checklist

### DIFF
--- a/gnuradio-runtime/lib/logger.cc
+++ b/gnuradio-runtime/lib/logger.cc
@@ -23,9 +23,27 @@
 #include <spdlog/sinks/stdout_color_sinks.h>
 #include <spdlog/spdlog.h>
 
+#include <iostream>
 #include <memory>
 
 namespace gr {
+
+static void setup_logger(const std::shared_ptr<spdlog::sinks::dist_sink_mt>& backend,
+                         const std::string& filename)
+{
+    if (filename == "stderr") {
+        auto console_sink = std::make_shared<spdlog::sinks::stderr_color_sink_st>();
+        backend->add_sink(console_sink);
+    } else if (filename == "stdout") {
+        auto console_sink = std::make_shared<spdlog::sinks::stdout_color_sink_st>();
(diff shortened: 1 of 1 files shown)


gnuradio-runtime/lib/logger.cc
```diff
@@ -23,9 +23,27 @@
...
+        backend->add_sink(console_sink);
+    } else if (filename == "stdout") {
+        auto console_sink = std::make_shared<spdlog::sinks::stdout_color_sink_st>();
+        backend->add_sink(console_sink);
+    } else if (!filename.empty()) {
+        auto file_sink =
+            std::make_shared<spdlog::sinks::basic_file_sink_st>(filename, true);
+        backend->add_sink(file_sink);
+    }
+}
```
willcode, line 45: else {
   // filename is empty - default to ... ?
}

gnuradio-runtime/lib/logger.cc
```diff
@@ -37,26 +55,12 @@ logging::logging()
     _default_backend->set_level(_default_level);
     _debug_backend->set_level(_debug_level);
 
-
-    auto debug_console_sink = std::make_shared<spdlog::sinks::stderr_color_sink_st>();
-    _debug_backend->add_sink(debug_console_sink);
-
     const auto pref = prefs::singleton();
     const auto default_file = pref->get_string("LOG", "log_file", "");
+    const auto debug_file = pref->get_string("LOG", "debug_file", "");
```
willcode, line 60: Defaulting to "stderr" would make this backward compatible.

mormj:
Haven't come up with a way to get 2 logging backends to work independently without breaking API - and I don't think this is worth breaking API over.  Cleanup in 4.0

PR #6166, opened 2022-09-13
pybind11: when no pygccxml, inherit gr::block

mormj:
## Description
When pygccxml is not installed, the intention is that the user gets a basic functionality with parsing of the block implementation that gets automatically pybinded.  However, this doesn't appear to work at the moment because to have access to base class methods, the generated bindings need to list the base class.  

Prior to this fix, the following error occurs during flowgraph run when pygccxml is not installed and `gr_modtool bind` is called on a block:
```
ValueError: Unable to coerce endpoints: 'gnuradio.asdf.asdf_python.foo_general' object has no attribute 'to_basic_block'
```

This fix adds the minimally necessary gr::block and gr::basic_block to the inheritance chain

OOT modules using gr_modtool only when pygccxml is not installed

## Testing Done
Used modtool to add and bind blocks, then test in GRC

## Checklist

### DIFF
--- a/gr-utils/blocktool/core/parseheader_generic.py
+++ b/gr-utils/blocktool/core/parseheader_generic.py
@@ -286,6 +286,7 @@ class GenericHeaderParser(BlockTool):
             namespace_dict['name'] = "::".join(namespace_to_parse)
             class_dict = {}
             class_dict['name'] = blockname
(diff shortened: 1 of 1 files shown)


dl1ksv (approved):
Looks good. Works with and without pygccxml

Issue #6109, opened 2022-08-22
gnuradio-blocks should not link publicly to libsndfile


ryanvolz:
### What happened?

GNSS-SDR links to gnuradio-blocks, and consuming the existing `gnuradio::gnuradio-blocks` CMake target requires the downstream library (GNSS-SDR) to link against libsndfile and thus have it as a dependency. This shouldn't be the case, because the use of libsndfile is (as far as I can tell) limited to the internal implementation and headers of the WAV source/sink and is not required to consume any public GNU Radio headers.

The fix would be to move `sndfile::sndfile` from the `PUBLIC` to `PRIVATE` link section for the CMake target: https://github.com/gnuradio/gnuradio/blob/main/gr-blocks/lib/CMakeLists.txt#L184-L199. Doing so would not effect the compilation of GNU Radio itself, so testing of the fix would have to involve re-building the changed GR and ensuring that something linking against gnuradio-blocks does not need the sndfile headers available.

### System Information

Ubuntu, but applies to all.

### GNU Radio Version

3.10 (maint-3.10)

### Specific Version

3.10.3.0

### Steps to Reproduce the Problem

Link against gnuradio-blocks using the `gnuradio::gnuradio-blocks` CMake target.

jwmelto:
First observed on Ubuntu 22.04 / GNU Radio 3.10.1
Also present on Fedora rawhide / GNU Radio 3.10.3

Minimizing external linkage is great, but if the linkage is actually required, then the `gnuradio` packages should depend on their distro-specific sources (`libsndfile1-dev` for Ubuntu; `libsndfile-devel` for Fedora).

PR #6172, opened 2022-09-14
gr-fec: More graceful error handling of missing gsl dependency.

velichkov:
Signed-off-by: Vasil Velichkov <vvvelichkov@gmail.com>

## Description
Can't build an OOT that links to gnuradio-fec

## Checklist


ryanvolz (changes requested):
GSL does not appear in gr-fec's (and probably all of GNU Radio's) public headers, so it's not actually required in order for OOTs to link with `gnuradio::gnuradio-fec` except that CMake thinks it is. So I think the better fix is to change GSL from a `PUBLIC` to `PRIVATE` link dependency, and then there's no need to propagate finding GSL to OOTs.

As an aside, this is yet more evidence (like #6109) that we need to audit GNU Radio's publicly linked libraries to figure out what can be privated. It's on my list to get to eventually...

ryanvolz (approved):

willcode:
@velichkov Do you need this in order to package this week's releases? 3.9 is pretty much done. I can do another rc if this is important.

Issue #5437, opened 2021-12-31
"Hilbert" misnamed: block doesn't calculate (only) the Hilber transform, but the analytic signal


marcusmueller:
### What happened?

The API documentation says

> hilbert_fc
> Hilbert transformer.
> real output is input appropriately delayed. imaginary output is hilbert filtered (90 degree phase shift) version of input.
> Constructor Specific Documentation:
> Build a Hilbert transformer filter block.
> Args:
>     ntaps : The number of taps for the filter.
>     window : Window type (see `fft::window::win_type`) to use.
>     param : Parameter value for a Kaiser/Exp/Gaussian/Tukey window.

Now, what input + H(input) for a real input is, it's the analytic signal.

The bug here is merely naming; also, it would seem to be nice to have "Hilbert transformer taps" as separate block, just as we have "low pass filter taps", etc, especially since chances are you'd want your discrete HT to be of significant length, i.e. use an FFT filter.

### System Information

--

### GNU Radio Version

3.10-git (master)

### Specific Version

since the beginning of time (at least 3.6.2)

### Steps to Reproduce the Problem

--

valdaarhun:
Hi. I am new to gnuradio and I would like to work on this issue. I have read through the [PySDR](https://pysdr.org/) tutorial so I believe I have been acquainted with signals and systems, and dsp in general.

valdaarhun:
Hi. I wanted to confirm the changes that are to be done in this issue.

Should the name `hilbert_fc` be changed to something else, for eg: `analytic_signal_fc`? And then should a new `hilbert_fc` block be created that calculates `H(input)`?

marcusmueller:
Hi @valdaarhun, sorry for the long silence.

> Should the name hilbert_fc be changed to something else, for eg: analytic_signal_fc? 

No! That's far too intrusive, as it breaks existing code, only to have better naming. All that needs to change is the name in the .grc Block definition file.

> And then should a new hilbert transformer tap be created that calculates H(input)?

no.

valdaarhun:
Hi. I am sorry for the really late response. I think I have understood what I need to do here. I believe I need to change the description given in the `hilbert_fc.h` file.

valdaarhun:
Hi. I have opened a PR in #6177. Please do let me know if there are changes that need to be incorporated.



[...]

PR #6177, opened 2022-09-17
docs: Fix definition of hilbert transformer in hilbert_fc.h

valdaarhun:
## Description

In [hilbert_fc.h](https://github.com/gnuradio/gnuradio/blob/main/gr-filter/include/gnuradio/filter/hilbert_fc.h), the description given is that of an analytic signal rather than the hilbert transformer.

## Related Issue

Related to #5437 

The change is related to the hilbert transformer block. There shouldn't be a change in performance whatsoever.

## Testing Done

Documentation-related change, so no testing is required.

## Checklist

### DIFF
--- a/gr-filter/include/gnuradio/filter/hilbert_fc.h
+++ b/gr-filter/include/gnuradio/filter/hilbert_fc.h
@@ -25,8 +25,7 @@ namespace filter {
  * \ingroup filter_blk
  *
  * \details
- * real output is input appropriately delayed.
- * imaginary output is hilbert filtered (90 degree phase shift)
+ * the output is hilbert transformed (90 degree phase shift)
  * version of input.
  */
 class FILTER_API hilbert_fc : virtual public sync_block


mbr0wn (changes requested):
The output of this block is in fact the analytic signal to its input (see https://github.com/gnuradio/gnuradio/blob/463c3477549b26b32d9b73eef30044e97c4eee64/gr-filter/lib/hilbert_fc_impl.cc#L44-L56). That means the original documentation is, in fact, accurate.

In practice, a single block that simply phase-shifts can be implemented by other means and would not be super useful by itself. Hence this block that generates an analytic signal to its real input.

I'm guessing that you're thrown off by the name of the block. Given that, I agree though that the documentation could be improved. Please either find a better description of this block, or consider closing the PR again.

valdaarhun:
Hi. I understand the issue here. The description and implementation both point to the generation of an analytic signal. But the name implies that this is a hilbert transformer. Simply changing the name of the file to something more appropriate might have unfavourable project-wide consequences.

According to [this description](https://github.com/gnuradio/gnuradio/issues/5437#issue-1091620730):

> it would seem to be nice to have "Hilbert transformer taps" as separate block, just as we have "low pass filter taps", etc, especially since chances are you'd want your discrete HT to be of significant length, i.e. use an FFT filter.

How do you find the idea of moving the current implementation to a new file that is appropriately named and changing the implementation in `hilbert_fc.h` to something based on the above comment?

mbr0wn:
> > it would seem to be nice to have "Hilbert transformer taps" as separate block, just as we have "low pass filter taps", etc, especially since chances are you'd want your discrete HT to be of significant length, i.e. use an FFT filter.

I think this assessment by @marcusmueller is a bit overkill. How often do people actually use Hilbert transfomers in practice? All SDR hardware generates complex I/Q. If you are using this block in a simulation, then there's no need to optimize this for the last few percentages. And are you really going to create a Hilbert transformer with so many taps that the FFT filter is going to be significantly faster? I doubt many people who run this in real-time applications have ever changed the default number of taps, which is 65. And which does not require an FFT filter.

> How do you find the idea of moving the current implementation to a new file that is appropriately named and changing the implementation in `hilbert_fc.h` to something based on the above comment?

Like I said, I don't agree with @marcusmueller on this being a major issue. I would suggest going back to improving the documentation, maybe adding a better explanation what exactly it is that this block does, maybe a reference or two. You can reference textbooks or wikipedia, and you can certainly reference `firdes::hilbert()` which is used internally. IMO, that would be sufficient to close the associated bug.

Thanks for taking a look at this, @valdaarhun!

mbr0wn:
Thank you!

mbr0wn:
@valdaarhun I'll close this PR until you have a better documentation suggestion. Then, please either re-open the PR or create a new one. This is so the PR queue on github doesn't become too cluttered. We look forward to your contribution!

valdaarhun:
Hi. Sorry for the delay. I have got a new patch ready. I am not sure how one re-opens a PR. I think I'll just open a new PR.

PR #6178, opened 2022-09-19
cmake: Switch suitable `PUBLIC` `target_link_libraries` to `PRIVATE`

### LINKED ISSUE #6109: gnuradio-blocks should not link publicly to libsndfile

ryanvolz:
## Description

Declaring a library as `PUBLIC` in `target_link_libraries` tells downstream linkers of the target to also need to link to the `PUBLIC` libraries. This is necessary when the headers of those libraries are required by the public headers of the target library. There were some instances where linked libraries were declared `PUBLIC` but were not required to use the public headers, so they have been switched to `PRIVATE` declarations. This will allow OOT modules to reduce their linking and dependencies in some cases.

This could use some scrutiny from people more familiar with the various modules that are affected. I made these changes based purely on searching for every `target_link_libraries` call and then searching the public headers to see if the linked libraries were `#include`d somewhere. It's very possible I missed things.

`Boost::program_options` used to be a `PUBLIC` link of gnuradio-runtime, but I think it makes more sense as something that is linked by individual executables that need it, so I moved that where appropriate.

This also should conform to the new CMake formatting, which resulted in some more formatting changes in files that I guess were missed in that earlier PR.

## Related Issue

Fixes #6109.

CMake in many modules.

## Testing Done

None besides CI build testing. The real effects of this are downstream with building OOTs, so it would be good to try building ones that use affected modules. That might be easiest to do if this were merged just to main first to see if anyone notices problems. 

## Checklist

### DIFF
--- a/gnuradio-runtime/apps/CMakeLists.txt
+++ b/gnuradio-runtime/apps/CMakeLists.txt
@@ -9,5 +9,5 @@
 # Setup executables
 ########################################################################
 add_executable(gnuradio-config-info gnuradio-config-info.cc)
-target_link_libraries(gnuradio-config-info gnuradio-runtime)
+target_link_libraries(gnuradio-config-info gnuradio-runtime Boost::program_options)
 install(TARGETS gnuradio-config-info DESTINATION ${GR_RUNTIME_DIR})
--- a/gnuradio-runtime/lib/CMakeLists.txt
+++ b/gnuradio-runtime/lib/CMakeLists.txt
@@ -122,11 +122,12 @@ target_link_libraries(
     PUBLIC
         gnuradio-pmt
         Volk::volk
-        Boost::program_options
         Boost::thread
         spdlog::spdlog
         MPLib::mplib
-    PRIVATE libunwind::libunwind)
+    PRIVATE
+        Boost::program_options
+        libunwind::libunwind)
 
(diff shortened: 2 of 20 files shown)


mormj (approved):
Going to let this one sit on main.  Thanks for all the interface cleanup!

PR #6189, opened 2022-09-23
4.0: Fix some single mapped buffer issues and add test

mormj:
## Description
There were some issues with single mapped buffers and they weren't being tested.  
This adds a buffer type: buffer_cpu_host (copied from GR 3.x) that simulates a host/device buffer type but can be used without specific hardware.  
The default buffer type can be configured in config.yml

## Testing Done
Added host_buffer test
Ran tests with default_buffer_type set to "cpu_host"


mormj:
Closing for now as I need to port more of the 3.10 logic in and can't get to it right away

PR #6191, opened 2022-09-27
4.0: Make gnuradio installation path consistent

dl1ksv:
## Description

On gnuradio 3.x there is only on 'prefix'/lib(64) with  its subdirectories. On 4.0 you can set --libdir. Setting this to lib64, you get 'prefix'/lib64 and
'prefix'/lib
 with subdirectories.

This patch removes the hardcoded lib subdirectory path and forces to install only to one directory ( lib or lib64 ) depending on the --libdir setting.

A similar patch was applied to gnuradio/pmt
## Testing Done

Tested on F36 and F37
## Checklist


mormj:
This looks great!  Thank you for the meson cleanup.  The hardcoded 'lib' in there was pretty ugly, so thanks for fixing it all up.

PR #6199, opened 2022-09-29
gr-blocks: print correct error message for soundfile sf_open

pinkavaj:
The wavfile_sink and wavfile_source  now prints invalid error messages when eg. the existing WAV file is broken.

Now description for `errno` error is printed, but soundfile library does not set `errno` on error, but have its own way to get error message. If sf_open fails, textual description of the error message should be retrieved by calling `sf_strerror(NULL)`, not `strerror(errno)`. But in case of error related for example to opening a file, the decription from `sf_strerror` is pretty crappy and misleading. This patch combines both aproaches to get the best from it.

I have manualy (on my project tested):
 * to create file in nonexisting directory
 * to open non-wav file as wav

Bot returned expected error description.


mormj:
This PR is opened against the `master` which is no longer the default branch.  Please update these changes against `main`

PR #6203, opened 2022-09-30
digital: Apply symbol map on hard variable constellation decodes

ThomasHabets:
## Description

Allow symbol map to actually work, when using hard decision `Constellation Decoder`.

## Related Issue

This is PR and bug report in one.

`Constellation Object`

## Testing Done

No testing done. Currently my building for source is segfaulting, but hoping that the github-triggered testing will be a bit useful.

Marking WIP until I can get it properly tested, and add a unit test.

## Checklist

- [maybe] I [have signed my commits before making this PR] (https://github.com/gnuradio/gnuradio/blob/main/CONTRIBUTING.md#dco-signed)

### DIFF
--- a/gr-digital/lib/constellation.cc
+++ b/gr-digital/lib/constellation.cc
@@ -407,7 +407,11 @@ constellation_calcdist::constellation_calcdist(std::vector<gr_complex> constell,
 // Inefficient.
 unsigned int constellation_calcdist::decision_maker(const gr_complex* sample)
 {
-    return get_closest_point(sample);
+    const auto point = get_closest_point(sample);
+    if (d_apply_pre_diff_code) {
+        return d_pre_diff_code.at(point);
+    }
+    return point;
 }
 
 


marcusmueller (approved):

ThomasHabets:
This is not right yet. Actually I'm not sure what it should be. It looks like *some* uses of `Constellation Object` uses the mapping, and other uses do not.

* [timing error detector thinks it needs to manually map](https://github.com/gnuradio/gnuradio/blob/main/gr-digital/lib/timing_error_detector.cc#L226).
* [constellation decoder thinks that it does not](https://github.com/gnuradio/gnuradio/blob/main/gr-digital/lib/constellation_decoder_cb_impl.cc#L59)
* [constellation soft decoder](https://github.com/gnuradio/gnuradio/blob/main/gr-digital/lib/constellation_soft_decoder_cf_impl.cc#L52) also thinks that it does not, but [in this case it's correct in thinking that](https://github.com/gnuradio/gnuradio/blob/main/gr-digital/lib/constellation.cc#L277)

I think the right way forward is to have the hard decider work like the soft decider, and do the mapping internally in the `Constellation Object`.

Opinions?

ThomasHabets:
... unless this means it gets double-mapped first by the `Symbol Sync` and then by the `Constellation Decoder`. But empirically before this PR setting the map to `[0,0,0,0]` did not map everything to zero in `Symbol Sync` or `Constellation Decoder`.

ThomasHabets:
Other uses: meas_evm_cc_impl.cc, ofdm_equalizer_simpledfe.cc, which both assume they have to do it. Ok, so this may actually be the best way. Updating PR to instead have constellation encoder do the mapping externally.

ThomasHabets:
Hmm... no maybe it's good. I need to think about it some more to understand the intent, here.

ThomasHabets:
I no longer think this is a good fix.

`d_pre_diff_code` looks like it should be used only for "well known" encodings, not for variable encoding. Specifically only `dqpsk`. And no test covers it!

It *is* also used for soft decoding of variable constellation, but not hard decoding. It's not used for any other hard or soft decoding. E.g. even though `qpsk` sets `d_pre_diff_code` it doesn't set `d_apply_pre_diff_code` so it appears to be a no-op. Kinda makes sense since qpsk is not a differential mode. But doesn't make sense since it's a no-op.

And why does the constellation object care if if's a differential mode? Am I missing something, or does the contellation block use this information at any point? 

It seems to me that constellation should either go all-in on the mapping stage, or not.

But then I realized the true intention.

## The `Constellation Encoder/Decoder` never meant to use all data from the `Constellation Object`

Differential? Do the math yourself before the encoder. (though "is differential" is not actually part of the `Constellation Object`. But the code implies that it's differential-aware)

Have a nonstandard encoding? The Encoder and Decoder doesn't care.

It seems that you are meant to put a `Map` block before / after the `Constellation Encoder / Decoder` if you want to map in another way. You can actually do this, by having the `Map` object have a map of `cons.pre_diff_code()`. It's a bad name for it, but there you go.

## Still unknown

So why does the soft decoder use it for DQPSK and optionally for variable constellation? I'll have to dig in more to understand the soft decoder output better.

My guess: It should not be there either.

mormj:
@ThomasHabets - thank you for the analysis!  Closing this for now - please reopen this or a new PR if you address the issues a different way

igorauad:
Hi @ThomasHabets ,

I've been experimenting with the Constellation Soft Decoder block and ran into some issues related to what you are describing in this thread. I cannot say I understand the implementation entirely yet, but I agree with some of your insights above.

In my case, I had the following definitions for a Gray-coded M-PAM constellation:

```python
mpam_const_points = 0.5 * np.arange(-(M - 1), (M + 1), 2)
gray_code_lut = [x^(x>>1) for x in range(0, M)]
gray_coded_mpam_const_points = [mpam_const_points[gray_code_lut[i]] for i in range(0, M)]
```

Namely, `mpam_const_points` defines the possible M-PAM constellation points in order, each element surrounded by its nearest neighbor(s) in Euclidean space. The `gray_code_lut` list contains the binary values of each point in `mpam_const_points`, namely the i-th point in `mpam_const_points` has value `gray_code_lut[i]`.  Finally, the `gray_coded_mpam_const_points` variable holds the constellation points in Gray-coded order.

On the GR flowgraph, I want to have a Chunk to Symbols block mapping input decimal values to the corresponding Gray-coded M-PAM symbol. Hence, the symbol table I use on the Chunk to Symbols block is the `gray_coded_mpam_const_points` list. That is, for input decimal `x`, the Tx sends `gray_coded_mpam_const_points[x]`. Meanwhile, on the Rx side, I want to have the Constellation Soft Decoder outputting the LLRs associated with `x`. So the test flowgraph is as follows:

Soft Decoder Test Flowgraph

The soft decoder block depends on the constellation object, which I experimented configuring in two ways:
- Approach 1:
    - Constellation Points field set to `mpam_const_points`.
    - Symbol map field set to `gray_code_lut`.
- Approach 2:
    - Constellation Points set to `gray_coded_mpam_const_points`.
    - Symbol map field set to nothing (empty list).


Constellation Object 1
Constellation Object 2

Then, I verified that Approach 1 does not work. When the Tx side transmits the decimal symbol `x`, the resulting constellation symbol is `gray_coded_mpam_const_points[x]`. However, the soft decoder outputs the LLRs associated with `gray_code_lut[x]`, not `x`. That happens due to this line: https://github.com/gnuradio/gnuradio/blob/v3.10.5.0/gr-digital/lib/constellation.cc#L277.

In contrast, Approach 2 does work. In this case, there is no pre-diff code, and the constellation symbols are already in Gray-coded order. As a result, when the Tx side transmits the decimal symbol `x`,  the soft decoder correctly outputs the LLRs associated with `x`.

The two solutions are in the attached flowgraphs:
[soft_decoder_test.zip](https://github.com/gnuradio/gnuradio/files/10270817/soft_decoder_test.zip).

I believe the following change on `constellation.cc` would ensure both approaches work, although I haven't had a chance to try this patch yet:
```diff
@@ -267,22 +267,22 @@ std::vector<float> constellation::calc_soft_dec(gr_complex sample, float npwr)
     std::vector<float> s(k, 0);
 
     for (int i = 0; i < M; i++) {
+        if (d_apply_pre_diff_code)
+            v = d_pre_diff_code[i];
+        else
+            v = i;
+
         // Calculate the distance between the sample and the current
         // constellation point.
-        float dist = std::abs(sample - d_constellation[i]);
+        float dist = std::abs(sample - d_constellation[v]);
         // Calculate the probability factor from the distance and
         // the scaled noise power.
         float d = expf(-dist / npwr);
 
-        if (d_apply_pre_diff_code)
-            v = d_pre_diff_code[i];
-        else
-            v = i;
-
         for (int j = 0; j < k; j++) {
             // Get the bit at the jth index
             int mask = 1 << j;
-            int bit = (v & mask) >> j;
+            int bit = (i & mask) >> j;
 
             // If the bit is a 0, add to the probability of a zero
             if (bit == 0)
```

That is, when computing the LLR contributions of the i-th constellation symbol, take the i-th point from the pre-diff mapped set of symbols (i.e., `d_constellation[d_pre_diff_code[i]]`, rather than the i-th point from the unmapped set of symbols (`d_constellation[i]`). At the same time, assume `i` (the index) is the binary value, so the LLRs are representative of the index `i` instead of the pre-diff coded value `d_pre_diff_code[i]`.

Nevertheless, it must be pointed out that my goal is to get the LLRs on the Rx side representing the Gray-demapped decimal symbols, the same decimal values sent on the Tx side. However, I imagine the original implementation had another objective in mind. In any case, maybe it would be useful to have an option to output the LLRs representing the pre-diff decoded binary values.

PR #6207, opened 2022-10-02
dev-4.0:  suppress empty 'outputs:' statements in yml code

dl1ksv:
## Description

On grc startup some errors like
audio_alsa_sink.block.yml  block: error: Value type 'NoneType' for 'outputs' not in valid types
are reported.
This pr avoids empty 'outputs:' statements.

## Testing Done

Start grc without and after applying this patch.
## Checklist

### DIFF
--- a/utils/blockbuilder/templates/blockname.grc.j2
+++ b/utils/blockbuilder/templates/blockname.grc.j2
@@ -86,7 +86,7 @@ parameters:
     hide: 'part'
 
 inputs:
-{% for port in ports -%}{%- if port['direction'] == 'input'%}
+{% for port in ports if port['direction'] == 'input'%}
 -   domain: {{'${impl.domain}' if port['domain'] == 'stream' else port['domain']}}
 {% if type_inst %}
     dtype: ${ type_inst.{{port['type']|grc_type}} }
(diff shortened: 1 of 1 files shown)


utils/blockbuilder/templates/blockname.grc.j2
```diff
@@ -97,16 +97,17 @@ inputs:
...
-{% endif %}{% endfor %}
+{% endfor %}
 -   domain: message
     id: param_update
     label: param
     optional: true
     hide: ${ not showports }
 
+{% for port in ports if port['direction'] == 'output'%}
+{% if loop.first %}
```
mormj, line 108: Should we do the same for input ports?

dl1ksv: This was my first thought, too. But then I noticed that input ports are never empty
https://github.com/gnuradio/gnuradio/blob/64951620164f098d9c16f5bc47bc562219df0606/utils/blockbuilder/templates/blockname.grc.j2#L88
.......
https://github.com/gnuradio/gnuradio/blob/64951620164f098d9c16f5bc47bc562219df0606/utils/blockbuilder/templates/blockname.grc.j2#L101-L105

mormj: oh that's right.  i forgot about the default message ports.

PR #6209, opened 2022-10-03
gr-blocks: print correct error message for soundfile sf_open

pinkavaj:
The wavfile_sink and wavfile_source prints misleading error messages when eg. the existing WAV file is broken.

Description for `errno` error is printed and errors reported directly by `libsndfile` are ignored. The reason to use `errno` is likely because of `libsndfile` returns unhelpful error messages when the error is at the system IO level. This patch combines both approaches to get the best results.

I have manually tested (on my project):
 * to create file in nonexisting directory
 * to open non-wav file as wav

Bot returned expected error description.

### DIFF
--- a/gr-blocks/lib/wavfile_sink_impl.cc
+++ b/gr-blocks/lib/wavfile_sink_impl.cc
@@ -112,8 +112,16 @@ bool wavfile_sink_impl::open(const char* filename)
     if (d_append) {
         // We are appending to an existing file, be extra careful here.
         sfinfo.format = 0;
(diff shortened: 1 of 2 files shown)


drmpeg (approved):
Tested with libsndfile 1.1.0 to force an error on high sample rate. The reported error [1] is now correct and much more informative. This is an excellent update.

[1] ```wavfile_source :error: sf_open failed: /opt/xfer/sine1000samp.wav: Internal error : SF_INFO struct incomplete.```

PR #6211, opened 2022-10-04
4.0: Fix Single Mapped Buffer logic

mormj:
This brings the rest of the code changes in from 3.10 custom buffers

The logic is throughout the entire scheduler and buffer objects

## Testing Done
Added a QA test


mormj:
needs to be squashed

PR #6213, opened 2022-10-04
4.0: meson: move pkgconfig/cmake to lib dir for all modules

mormj:
This way modules can be referenced out of tree - it was done this way for some but not all modules


mormj:
OpenSuse Tumbleweed is unhappy on every PR now :/

PR #6215, opened 2022-10-05
4.0: soapy: fix up the bindings

mormj:
The test soapy application would not run - this brings the base bindings into the python_bindings file.

### DIFF
--- a/blocklib/soapy/.gitignore
+++ b/blocklib/soapy/.gitignore
@@ -1,5 +1,5 @@
 meson.build
 !include/gnuradio/soapy/meson.build
 !lib/meson.build
-!python/soapy/bindings/meson.build
+!python/gnuradio/soapy/bindings/meson.build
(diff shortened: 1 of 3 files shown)


ryanvolz:
I noticed that the `soapy/test/meson.build` is empty, so the test doesn't appear to run during `meson test`. Am I misunderstanding how this works or is that a bug?

Also, I'm guessing this is likely to fix the import error I'm seeing on the Linux conda build?

mormj:
This is likely a bug - there is some inconsistency as to which meson files get .gitignored, and I need to go through and clean it all up.

mormj:
Also, the soapy test `test_soapy.py` doesn't run in QA because it requires hardware to be present.  It is really just a debug script.  Not sure the best place to put these, but have named such files test_ and the ones that run with CI qa_

PR #6221, opened 2022-10-06
qtgui: cmake: Align CMake structure with other modules

mbr0wn:
This adds two changes to the qtgui CMake:

- The structure of lib/CMakeLists.txt now matches that of other
  in-tree modules. This is so that in the future, we can have optional
  sources, e.g. if we make QtOpenGL an optional dependency, and want to
  add blocks based on its existence.
- We find the Qt5 package with QtWidgets as a COMPONENT, rather than
  directly for Qt5Widgets. This allows adding OPTIONAL_COMPONENTS later
  on.

## Testing Done

built yo

## Checklist

### DIFF
--- a/gr-qtgui/CMakeLists.txt
+++ b/gr-qtgui/CMakeLists.txt
@@ -12,7 +12,9 @@ include(GrPython)
 
 # Note: gr-qtgui requires Qt5.
 
-find_package(Qt5Widgets QUIET)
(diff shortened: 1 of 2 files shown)


ryanvolz (approved):
Looks good. Will probably have to update either this or #6178 so the diffs are compatible, but there is no substantive conflict there.

PR #6234, opened 2022-10-09
fec: Prevent integer overflow in path metric calculations

### LINKED ISSUE #5851: Decode CCSDS 27 degrades over time
### What happened?

I encoded a stream of bits with Encode CCSDS 27, conveted 0 and 1 to (float) -1 and +1, added Gaussian noise (amplitide 0.69) and decoded with Decode CCSDS 27.

It works for about the first 10 million bits, and then the bit error rate of the decoded bits gets dramatically worse.

(I have a vague memory that we had a bug like this before).

What I expect to happen is that the bit error rate in the output should remain approximately the same, no matter how many bits have previously been processed.

As a work-around, I can probably use one of the other implementations of the Viterbi algorithm in Gnuradio (there are several). If this is a known limitation, it probably ought to be documented.

### System Information

OS: Ubuntu 22.04
GR Installation Method: compiled from source

### GNU Radio Version

3.11-git (main)

### Specific Version

0.0.0.0

I encoded a stream of bits with Encode CCSDS 27, conveted 0 and 1 to (float) -1 and +1, added Gaussian noise (amplitide 0.69) and decoded with Decode CCSDS 27.

### Relevant log output

_No response_

argilo:
## Description
The Decode CCSDS 27 block fails after processing approximately 4.2 million input bits. This happens because Viterbi path metric calculations are done using signed 32-bit integers, even though they are stored in signed 64-bit integers. Switching the calculations to signed 64-bit integers resolves the issue.

## Related Issue
Fixes #5851.

* Viterbi decoder

## Testing Done
I ran @daniestevez's test script (https://destevez.net/2017/07/degradation-bug-in-gnu-radio-decode-ccsds-27/) and also added a QA test which fails without this change and passes with it.

## Checklist

### DIFF
--- a/gr-fec/lib/decode_ccsds_27_fb_impl.cc
+++ b/gr-fec/lib/decode_ccsds_27_fb_impl.cc
@@ -66,6 +66,14 @@ int decode_ccsds_27_fb_impl::work(int noutput_items,
                 viterbi_get_output(d_state0, out++);
                 // printf("%li\n", *(out-1), metric);
             }
+
+            // Periodically normalize the path metrics to prevent overflow
+            if (d_count % 65536 == 65535) {
+                long diff = d_state0[0].metric;
+                for (int j = 0; j < 64; j++) {
+                    d_state0[j].metric -= diff;
+                }
+            }
         }
 
         d_count++;
--- a/gr-fec/python/fec/qa_ecc_ccsds_27.py
+++ b/gr-fec/python/fec/qa_ecc_ccsds_27.py
@@ -11,6 +11,7 @@
 
 from gnuradio import gr, gr_unittest, blocks
 from gnuradio import fec
+import numpy as np
 
 
 class test_ccsds_27 (gr_unittest.TestCase):
@@ -21,9 +22,9 @@ class test_ccsds_27 (gr_unittest.TestCase):
     def tearDown(self):
         self.tb = None
 
-    def xtest_ccsds_27(self):
-        src_data = (1, 2, 3, 4, 5, 6, 7, 8, 9, 10)
-        expected = (0, 0, 0, 0, 1, 2, 3, 4, 5, 6)
+    def test_ccsds_27(self):
+        src_data = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
+        expected = [0, 0, 0, 0, 1, 2, 3, 4, 5, 6]
         src = blocks.vector_source_b(src_data)
         enc = fec.encode_ccsds_27_bb()
         b2f = blocks.char_to_float()
@@ -35,6 +36,20 @@ class test_ccsds_27 (gr_unittest.TestCase):
         self.tb.run()
         dst_data = dst.data()
         self.assertEqual(expected, dst_data)
+
+    def test_ccsds_27_long(self):
+        src_data = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10] * 100000
+        src = blocks.vector_source_b(src_data)
+        enc = fec.encode_ccsds_27_bb()
+        b2f = blocks.char_to_float()
+        add = blocks.add_const_ff(-0.5)
+        mul = blocks.multiply_const_ff(2.0)
+        dec = fec.decode_ccsds_27_fb()
+        dst = blocks.vector_sink_b()
+        self.tb.connect(src, enc, b2f, add, mul, dec, dst)
+        self.tb.run()
+        dst_data = dst.data()
+        self.assertEqual([0, 0, 0, 0] + src_data[:-4], dst_data)
 
 
 if __name__ == '__main__':
--- a/gr-fec/lib/decode_ccsds_27_fb_impl.h
+++ b/gr-fec/lib/decode_ccsds_27_fb_impl.h
@@ -26,7 +26,7 @@ private:
     struct viterbi_state d_state1[64];
     unsigned char d_viterbi_in[16];
 
-    int d_count;
+    unsigned int d_count;
 
 public:
     decode_ccsds_27_fb_impl();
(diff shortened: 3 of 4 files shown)


gr-fec/python/fec/qa_ecc_ccsds_27.py
```diff
@@ -21,9 +21,9 @@ def setUp(self):
     def tearDown(self):
         self.tb = None
 
-    def xtest_ccsds_27(self):
-        src_data = (1, 2, 3, 4, 5, 6, 7, 8, 9, 10)
-        expected = (0, 0, 0, 0, 1, 2, 3, 4, 5, 6)
+    def test_ccsds_27(self):
+        src_data = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
+        expected = [0, 0, 0, 0, 1, 2, 3, 4, 5, 6]
```
argilo, line 26: This test used to be disabled. The problem was a type mismatch (list vs. tuple), which I've solved by switching the inputs to lists to match the output type of the Vector Sink.

gr-fec/python/fec/qa_ecc_ccsds_27.py
```diff
@@ -36,6 +36,20 @@ def xtest_ccsds_27(self):
         dst_data = dst.data()
         self.assertEqual(expected, dst_data)
 
+    def test_ccsds_27_long(self):
+        src_data = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10] * 100000
```
argilo, line 40: This generates 1,000,000 bytes = 8,000,000 bits of input, which is enough to trigger the bug. The test takes about 650ms to execute on my laptop, which seems acceptable.

gr-fec/python/fec/qa_ecc_ccsds_27.py
```diff
@@ -36,6 +36,20 @@ def xtest_ccsds_27(self):
...
+        enc = fec.encode_ccsds_27_bb()
+        b2f = blocks.char_to_float()
+        add = blocks.add_const_ff(-0.5)
+        mul = blocks.multiply_const_ff(2.0)
+        dec = fec.decode_ccsds_27_fb()
+        dst = blocks.vector_sink_b()
+        self.tb.connect(src, enc, b2f, add, mul, dec, dst)
+        self.tb.run()
+        dst_data = dst.data()
+        self.assertEqual(src_data[-104:-4], dst_data[-100:])
```
argilo, line 51: I'm only comparing the last 100 output bytes here, because `assertEqual` is very slow when comparing long lists.

argilo:
It looks like this fails on Windows. Maybe long is 32 bits there? It might be necessary to reset the path metrics after all.

willcode:
How about `long long`?

argilo:
I think that would be an API change though:

https://github.com/gnuradio/gnuradio/blob/c49548c2e079fa3b50efbd8d1699b5089d523438/gr-fec/include/gnuradio/fec/viterbi.h#L23-L26

argilo:
I'll have a look to see whether other Viterbi decoders in GNU Radio solve this in a clean way.

argilo:
The other implementation I found (cc_decoder) was different enough I didn't find it a useful model to follow. I just added a normalization step that occasionally subtracts out whatever has built up in the path metrics table.

I also changed the type of `d_count` from `int` to `unsigned int`; without that change it would become negative when it overflowed, which would break the modular arithmetic calculations in the `work` function.

willcode:
Is `state[0].metric` always the smallest? Or does it matter if the other metrics go negative?

gr-fec/python/fec/qa_ecc_ccsds_27.py
mbr0wn: I agree. This is important enough.

argilo:
> Is `d_state0[0].metric` always the smallest? Or does it matter if the other metrics go negative?

It won't necessarily be the smallest, but the metrics all grow at approximately the same rate. It shouldn't be a problem if some of them go negative, since they're stored in signed longs.

gr-fec/lib/decode_ccsds_27_fb_impl.cc
```diff
@@ -66,6 +66,14 @@ int decode_ccsds_27_fb_impl::work(int noutput_items,
                 viterbi_get_output(d_state0, out++);
                 // printf("%li\n", *(out-1), metric);
             }
+
+            // Periodically normalize the path metrics to prevent overflow
+            if (d_count % 65536 == 65535) {
```
argilo, line 71: With this interval, the metrics reach approximately 2^24 before being normalized. This keeps them well below overflow, which would occur at 2^31.

gr-fec/python/fec/qa_ecc_ccsds_27.py
argilo: I had a second look at this, and `assertEqual` is actually quite fast in the success case (~40 ms) whereas `np.testing.assert_equal` takes more than 11 seconds. `assertEqual` only becomes slow when the lists are unequal, but that won't happen anyway unless one of the blocks under test is broken. I'll update this to compare the entire list.

daniestevez: That's a bit surprising, but anyhow, since `assertEqual` is fast in the success case, the idea to use it with the full list is good. I suspect that in the failure case it spends too much time trying to find how the inputs differ and how to format the error output.

argilo: The issue is discussed here: https://github.com/python/cpython/issues/63416

It looks like it may be fixed in Python 3.12.

mormj (approved):

PR #6244, opened 2022-10-12
runtime: fix type replication in py_io_signature

### LINKED ISSUE #6243: runtime: py_io_signature - fails to concatenate tuple with list
### What happened?

When writing Python OOT block, if one uses _py_io_signature_ to specify IO port constraints like this:
```
        gr.sync_block.__init__(self,
            name="myblock",
            in_sig=gr.py_io_signature(1, 3, [np.float64]),
            ...)
```
GR fails when calling _py_io_signature.port_types()_ later down the pipe (GRC logs):
```
thread_body_wrapper :error: ERROR thread[thread-per-block[17]: <myblock(4)>]: TypeError: can only concatenate tuple (not "list") to tuple

At:
  /usr/lib/python3.10/site-packages/gnuradio/gr/gateway.py(82): port_types
  /usr/lib/python3.10/site-packages/gnuradio/gr/gateway.py(172): handle_general_work


>>> Done
```
This is because when less data types are given than number of ports, the last type in the list should be replicated to fill for missing types. Current code does it almost correctly, except that it creates a list (not tuple) and then attempts to concatenate it with a tuple, and that fails:
```
return self.__types + [self.__types[-1]] * (nports - ntypes)
# ^^^ in above line, self.__types is a tuple
``` 

Same bug is present on `main` branch, so I'm going to make a PR to fix it (it's a one-line fix)

### System Information

OS: Linux (Gentoo)
GR Installation Method: native - Portage
Python: 3.10

### GNU Radio Version

3.10 (maint-3.10)

### Specific Version

3.10.3.0

Make OOT Python block, use _gr.py_io_signature_ to specify 2 or more inputs (outputs), but give it only one data type, try to run the flowgraph - you will see exception that I've mentioned in the description.

### Relevant log output

_No response_

RomaVis:
## Description
Original version of _py_io_signature.port_types()_ method concatenates tuple with a list - and that fails in modern Python. This PR makes sure everything in that operation is a tuple, so it works correctly.

## Related Issue
Fixes #6243

From what I see, respective part of original code was unusable and always failed (though, perhaps it worked in Python 2? I'm not sure about that...). This patch should have no effect on current blocks or flowgraphs which were running correctly without the patch applied.

## Testing Done
With a patch applied, when py_io_signature is used in a way described in #6243, flowgraph now runs correctly instead of throwing `TypeError: can only concatenate tuple (not "list") to tuple` Python exception. Verified to work with a custom OOT block locally.

## Checklist

### DIFF
--- a/gnuradio-runtime/python/gnuradio/gr/gateway.py
+++ b/gnuradio-runtime/python/gnuradio/gr/gateway.py
@@ -79,7 +79,7 @@ class py_io_signature(object):
             return ()
         if nports <= ntypes:
             return self.__types[:nports]
-        return self.__types + [self.__types[-1]] * (nports - ntypes)
+        return self.__types + (self.__types[-1],) * (nports - ntypes)
 
     def __iter__(self):
         """


willcode (approved):
It did run on Python2 ... when I added this code 5 years ago. Apparently most people are still using the older backward compatible method, and not `py_io_signature`.

PR #6245, opened 2022-10-13
analog: pll blocks pybind inherit control_loop

mormj:
## Description
The python bindings for pll blocks did not have the base class inheritance for control loop, so these methods were not available as callbacks

## Related Issue
None
analog/pll_*

## Testing Done
Added QT range variables for the 3 parameters in the blocks.  before the fix, it caused exceptions when the value was changed, after there is no error

## Checklist


dl1ksv (approved):
pll_carriertracking and pll_freqdet_cf work for me without this patch, but pll_refout_cc fails with
AttributeError: 'gnuradio.analog.analog_python.pll_refout_cc' object has no attribute 'set_loop_bandwidth'

After applying your patch, everything works fine.

dl1ksv:
pll_carriertracking and pll_freqdet implement
https://github.com/gnuradio/gnuradio/blob/032ead82b43eb505925f2ff1e215c14de4838850/gr-analog/include/gnuradio/analog/pll_carriertracking_cc.h#L63-L66
while pll_refout does not.

Should we remove these implementations now?

Issue #5973, opened 2022-06-28
i686 test failures


unspecd:
### What happened?

I am working on updating the GNU Radio from 3.8.0.0 to 3.10.3.0. I'm seeing failures when building on i686.

Here is the build log: https://github.com/void-linux/void-packages/runs/7086379225?check_suite_focus=true#step:7:10646

### System Information

OS: Void Linux (kernel 5.18.7, gcc 10.2.1, python 3.10.5, pybind 2.9.2, pygccxml 2.2.1)

### GNU Radio Version

3.10 (maint-3.10)

### Specific Version

3.10.3.0

### Steps to Reproduce the Problem

```
$ git clone https://github.com/void-linux/void-packages && cd void-packages
$ curl -L https://github.com/void-linux/void-packages/pull/35049.patch | git apply -
$ ./xbps-src binary-bootstrap
$ ./xbps-src -Q pkg gnuradio
```

### Relevant log output

```shell
[0/1] Running tests...
Test project /builddir/gnuradio-3.10.3.0/build
        Start   1: runtime_qa_buffer.cc
  1/261 Test   #1: runtime_qa_buffer.cc .........................   Passed    5.86 sec
        Start   2: runtime_qa_io_signature.cc
  2/261 Test   #2: runtime_qa_io_signature.cc ...................   Passed    0.01 sec
        Start   3: runtime_qa_logger.cc

... skipped ...

        Start  63: qa_moving_average
 63/261 Test  #63: qa_moving_average ............................***Failed    0.37 sec
...F.
======================================================================
FAIL: test_vector_complex (__main__.test_moving_average)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/builddir/gnuradio-3.10.3.0/gr-blocks/python/blocks/qa_moving_average.py", line 150, in test_vector_complex
    self.assertEqual(dut_data, ref_data)
AssertionError: Lists differ: [(705[443 chars].3553466796875+406.9476318359375j), (208.45861[1463 chars]25j)] != [(705[443 chars].35531616210938+406.9476318359375j), (208.4586[1474 chars]75j)]
First differing element 11:
(207.3553466796875+406.9476318359375j)
(207.35531616210938+406.9476318359375j)
Diff is 4376 characters long. Set self.maxDiff to None to see it.
----------------------------------------------------------------------
Ran 5 tests in 0.124s
FAILED (failures=1)
        Start  64: qa_multiply_conjugate
[...]

PR #6248, opened 2022-10-14
digital: Fix constellation soft decision LUT generation & tests

argilo:
## Description
As noted in #5973, `qa_constellation` fails on i686 with the following errors:

```
FAIL: test_soft_qam16_calc (__main__.test_constellation)
  File "/builddir/gnuradio-3.10.3.0/gr-digital/python/digital/qa_constellation.py", line 379, in test_soft_qam16_calc
    self.assertFloatTuplesAlmostEqual(y_cpp_raw_calc, y_cpp_table, 0)
  File "/builddir/gnuradio-3.10.3.0/gnuradio-runtime/python/gnuradio/gr_unittest.py", line 96, in assertFloatTuplesAlmostEqual
    self.assertEqual(len(a), len(b))
AssertionError: 17 != 25
FAIL: test_soft_qpsk_calc (__main__.test_constellation)
  File "/builddir/gnuradio-3.10.3.0/gr-digital/python/digital/qa_constellation.py", line 331, in test_soft_qpsk_calc
    self.assertEqual(y_cpp_raw_calc, y_cpp_table)
AssertionError: Lists differ: [0.0,[16 chars]1.0, 1.0, 0.0, 1.0, 1.0, 0.0, 0.0, 1.0, 0.0, 1[13 chars] 1.0] != [0.0,[16 chars]1.0, 0.0, 1.0, 0.0, 1.0, 0.0, 1.0, 0.0, 0.0, 1[28 chars] 1.0]
First differing element 5:
1.0
0.0
Second list contains 3 additional elements.
First extra element 17:
0.0
  [0.0,
   0.0,
   1.0,
   0.0,
   1.0,
+  0.0,
   1.0,
+  0.0,
+  1.0,
+  0.0,
+  1.0,
+  0.0,
   0.0,
   1.0,
   1.0,
   0.0,
-  0.0,
   1.0,
   0.0,
-  1.0,
-  1.0,
   0.0,
   1.0]
```
These errors occur because `constellation::gen_soft_dec_lut` is in fact broken: the exit conditions in its `while` loops depend on floating point calculation being exact, and when error creeps in an extra row & column may be added to the resulting lookup table, causing lookups to fail. I've fixed this by adding half a step to the end condition instead of a full step; this should allow a floating point error of plus or minus a half step.

In addition, I noticed that the `slicer` function in `qa_constellation` is broken; a misplaced `else` means that it deletes all the `1.0` values that should be in the result, and instead always puts a single `1.0` at the end of the list.

## Related Issue
* #5973

`gr::digital::constellation::soft_decision_maker`, when using the lookup table mode.

## Testing Done
I verified that the CI test runs correctly on both 32-bit and 64-bit machines after these changes.

## Checklist

### DIFF
--- a/gr-digital/lib/constellation.cc
+++ b/gr-digital/lib/constellation.cc
@@ -245,9 +245,9 @@ void constellation::gen_soft_dec_lut(int precision, float npwr)
     float maxd = 1.0f;
     float step = (2.0f * maxd) / (d_lut_scale - 1);
     float y = -maxd;
-    while (y < maxd + step) {
+    while (y < maxd + (step / 2)) {
         float x = -maxd;
-        while (x < maxd + step) {
+        while (x < maxd + (step / 2)) {
             gr_complex pt = gr_complex(x, y);
             d_soft_dec_lut.push_back(calc_soft_dec(pt, npwr));
             x += step;
--- a/gr-digital/python/digital/qa_constellation.py
+++ b/gr-digital/python/digital/qa_constellation.py
@@ -117,10 +117,10 @@ difficult_constellation_info = (
 def slicer(x):
     ret = []
     for xi in x:
-        if(xi < 0):
+        if xi < 0:
             ret.append(0.0)
-    else:
-        ret.append(1.0)
+        else:
+            ret.append(1.0)
     return ret
 
 


willcode (approved):
The original code probably should have iterated using integers - hard to represent `1/(2^n-1)` very well in float. This looks like a good  to `almostLessThanOrEqual(maxd)` :-)

PR #6251, opened 2022-10-15
digital: Rename test to avoid duplicate name

argilo:
## Description
qa_ofdm_frame_equalizer_vcvc.py has two tests named `test_002_static_wo_tags`. The second overwrites the first, preventing it from executing. The duplicate name was introduced in bb01988e75d50d82cbb44c1a49c86c1d08f05665, but that was a large merge commit so I'm not sure exactly what the test was meant to do. But it's different from the other test, so I think both should be kept. I've simply renamed the second test here.

**Edit:** I also removed a duplicate variable definition that @willcode noticed.

QA tests for the OFDM Frame Equalizer block.

## Testing Done
I verified that all 8 tests pass. (Prior to this change, only 7 out of 8 tests executed.)

## Checklist


willcode (approved):
Also, in `test_002_static()`, `channel` is defined twice (the same way, so no foul).

argilo:
Good catch. Since it's in the same test file, I'll add that into this pull request.

argilo:
Done.

PR #6254, opened 2022-10-15
fec: Fix LDPC output size calculation

### LINKED ISSUE #989: qa_fecapi_ldpc.py test when run in a 32bit system can hang indefinitely using 100% of one core
During rpm build on the Mageia Build System, qa_fecapi_ldpc.py hangs the system with one core (of 24) at 100% until that test is killed, after which all the other tests complete (with a few expected fails).
This only affects the 32bit build.
I reproduced this in a chroot locally (on a 4 core machine) with exactly the same outcome.
We have temporarily disabled that test in the 32bit build, but raising a bug here so it may be addressed properly.

argilo:
## Description
As noted in #989 and #5973, `qa_fecapi_ldpc` fails on i686 systems. This happens because the output size of the LDPC encoder is miscalculated. The frame size is multiplied by the code rate in floating point, and the result is cast to an integer without rounding. As a result, the output size may be one too small.

## Related Issue
Fixes #989.

* LDPC Generator Matrix
* LDPC Parity Check Matrix

## Testing Done
I verified that the QA tests pass on i686 and x86_64.

## Checklist

### DIFF
--- a/gr-fec/lib/ldpc_gen_mtrx_encoder_impl.cc
+++ b/gr-fec/lib/ldpc_gen_mtrx_encoder_impl.cc
@@ -55,7 +55,7 @@ bool ldpc_gen_mtrx_encoder_impl::set_frame_size(unsigned int frame_size)
 
     d_frame_size = frame_size;
 
-    d_output_size = static_cast<int>(d_rate * d_frame_size);
+    d_output_size = static_cast<int>(round(d_rate * d_frame_size));
 
     return ret;
 }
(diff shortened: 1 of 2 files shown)


argilo:
I chose this style for consistency with other `set_frame_size` methods in gr-fec:

https://github.com/gnuradio/gnuradio/blob/3349d483a1f7e5db0eea042815d237810373857b/gr-fec/lib/ldpc_decoder.cc#L70

https://github.com/gnuradio/gnuradio/blob/3349d483a1f7e5db0eea042815d237810373857b/gr-fec/lib/ldpc_bit_flip_decoder_impl.cc#L65

PR #6258, opened 2022-10-16
analog: Use approximate equality to check for standard CTCSS tones

argilo:
## Description
The CTCSS Squelch block checks whether one of the standard CTCSS tones is being used, so that it can set its guard bands more accurately in this case. The comparison checks for exact floating point equality, which is not a good idea since some of the CTCSS frequencies do not have an exact floating point representation. Here I've change the check to allow a tolerance of 0.001 Hz.

* CTCSS Squelch

## Testing Done
In addition to running the QA tests, I temporary added `printf` statements and passed in test values like 100.0, 100.0005, and 100.0015 to verify the behaviour of the `find_tone` method.

## Checklist

### DIFF
--- a/gr-analog/lib/ctcss_squelch_ff_impl.cc
+++ b/gr-analog/lib/ctcss_squelch_ff_impl.cc
@@ -36,7 +36,7 @@ ctcss_squelch_ff::make(int rate, float freq, float level, int len, int ramp, boo
 int ctcss_squelch_ff_impl::find_tone(float freq)
 {
     for (int i = 0; i <= max_tone_index; i++)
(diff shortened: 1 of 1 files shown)


willcode (approved):
Curious whether this caused an actual error, since the approximate rep for 2 float (that were specified as exact numbers) should be the same. But, the function should work for any input, so good idea to change.

argilo:
> the approximate rep for 2 float (that were specified as exact numbers) should be the same

Is that true across languages and architectures?

willcode:
Hmm, right, it could have been a double that got passed from Python. I really don't know the answer to that. And technically, the caller could have calculated the freq for some reason, so it can't be assumed to be an exact rep at any point.

PR #6259, opened 2022-10-16
blocks: Use mutex instead of std::atomic for probe synchronization

argilo:
## Description
@balister noticed that `test_003_race_condition_regression_test` (added in #5744) fails on ARM:

```
  File "/usr/lib/gnuradio/ptest/gr-blocks/python/blocks/qa_probe_[signal.py](http://signal.py/)", line 64, in test_003_race_condition_regression_test
    self.assertIn(output, [1 + 2j, 3 + 4j])
AssertionError: (1+4j) not found in [(1+2j), (3+4j)]
```

It's apparent from the error message that the probe level is not being updated atomically, so the I value from one sample and the Q value from another are mixed together. I suspect the problem is that `std::atomic` is not supported on all platforms.

The vector version of the Probe Signal block uses a mutex (since `std::atomic` cannot be used with vectors) and the corresponding test always passes on ARM. We can fix the problem with the scalar version by switching to a mutex. This brings the two implementations closer together, and removes the need to define the `_ENABLE_ATOMIC_ALIGNMENT_FIX` macro (which was needed to work around a `std::atomic` bug in MSVC, and added it #5744).

* Probe Signal

## Testing Done
I ran qa_probe_signal in a loop to verify that the race condition never occurs (at least on x86_64).

## Checklist

### DIFF
--- a/gr-blocks/lib/CMakeLists.txt
+++ b/gr-blocks/lib/CMakeLists.txt
@@ -212,10 +212,6 @@ if(MSVC)
     target_sources(gnuradio-blocks PRIVATE ${CMAKE_CURRENT_BINARY_DIR}/gnuradio-blocks.rc)
 endif(MSVC)
 
-if(MSVC)
-    target_compile_definitions(gnuradio-blocks PUBLIC _ENABLE_ATOMIC_ALIGNMENT_FIX)
-endif()
-
 if(BUILD_SHARED_LIBS)
     if(SNDFILE_FOUND)
         gr_library_foo(gnuradio-blocks SNDFILE)
--- a/gr-blocks/lib/probe_signal_impl.cc
+++ b/gr-blocks/lib/probe_signal_impl.cc
@@ -46,6 +46,7 @@ int probe_signal_impl<T>::work(int noutput_items,
 {
     const T* in = (const T*)input_items[0];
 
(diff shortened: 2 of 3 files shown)


willcode (approved):
There are only 3 other places in the code that use `atomic` and they all protect `bool` flags. Is it only the multi-byte atomic types that causes a problem here?

Do we run on any platform where `atomic<bool>` makes sense anyway?

argilo:
> Is it only the templated atomic type that causes a problem here?

To be honest, I'm not certain what's causing trouble. From what I've read about `std::atomic`, it seems like it should be allowed to use it with `std::complex<float>`. I wonder if @balister's platform might have a bug. Since MSVC had a bug of its own, it seemed prudent to stop using `std::atomic` in this block. The only potential downside I can see is performance.

I'm not sure whether anything should be done about the few places that use `std::atomic<bool>`. Since the value is stored in a single byte, I'd at least expect bugs to be less likely with `bool`.

willcode:
Just guessing that the mutex is a lot faster than any operation that reads the entire input in a typical work function, so the performance difference probably won't be detectable.

mbr0wn (approved):

balister:
I ran this code on a qemu arm (32) machine and running the test in a loop has shown no failures.

PR #6271, opened 2022-10-20
Fix controlport thrift backend on MSVC/Windows

ryanvolz:
## Description

This enables thrift support on Windows in the conda CI builds and tries to make it build/run. I've managed to get it building, but anything that loads gnuradio-runtime seemingly gets caught in a deadlock. I've hit my limit in understanding how that code works, so I'm posting this in case someone else has insights.

In particular, I'm not sure if https://github.com/gnuradio/gnuradio/commit/fa587e6e287be89c796aea6737f0af2387671852 is the right thing to do and I'm just seeing further bugs now that the build completes, or if that's the wrong approach to fix that particular build issue. [That build issue](https://dev.azure.com/conda-forge/84710dde-1620-425b-80d0-4cf5baca359d/_apis/build/builds/591568/logs/100) boils down to this:

- MSVC tries to compile `rpcserver_selector.cc`
- That includes `rpcserver_booter_thrift.h`, which merely declares `class rpcserver_base` and `class rpcserver_thrift` before referencing `thrift_server_template<rpcserver_base, rpcserver_thrift, rpcserver_booter_thrift>`
- The definition for `thrift_server_template` in `thrift_server_template.h` includes the implementation for `i_impl()` which requires an implicit conversion from `rpcserver_thrift` to `rpcserver_base` in the return statement
- MSVC errors [with C2440](https://learn.microsoft.com/en-us/cpp/error-messages/compiler-errors-1/compiler-error-c2440?view=msvc-170#cast-of-undefined-type) at that because it hasn't been told that `rpcserver_base` is a parent class of `rpcserver_thrift`

My solution was to just include the appropriate headers that define `rpcserver_base` and `rpcserver_thrift` rather than declaring them, but I'm not sure why that wouldn't have been done in the first place.

runtime (controlport), CI

## Testing Done

CI checks only

## Checklist


ryanvolz:
/rerender

(see if this works)

PR #6272, opened 2022-10-20
blocks: Increase tolerance in qa_throttle tests

argilo:
## Description
In https://github.com/gnuradio/gnuradio/pull/3149 I added tests for the Throttle block. I noticed that these tests are skipped in Conda builds (OSX & Windows), and the logs show that they do occasionally fail due to execution of the flow graph taking approximately 450 ms. (Conda builds appear to be more sensitive to timing issues because the test suite is executed in parallel.) Here I've increased the tolerance to 600 ms, which I hope will be enough to make the tests reliable. I've also removed `qa_throttle` from the list of skipped tests in Conda builds.

Tests for the Throttle block.

## Testing Done
I verified that the tests still pass reliably on my machine, even when adding CPU stress (with `stress -c 16`).

## Checklist

### DIFF
--- a/.packaging/conda_recipe/bld.bat
+++ b/.packaging/conda_recipe/bld.bat
@@ -35,7 +35,6 @@ qa_agc^
 |qa_file_source^
 |qa_rotator_cc^
 |qa_tcp_server_sink^
-|qa_throttle^
 |qa_wavfile^
 |test_modtool^
 %=EMPTY=%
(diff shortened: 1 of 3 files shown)


argilo:
Welp, it looks like 0.6 wasn't enough:

https://github.com/gnuradio/gnuradio/actions/runs/3357519882/jobs/5563380391

```
FAIL: test_rx_rate_tag (__main__.test_throttle)
  File "%SRC_DIR%\gr-blocks\python\blocks\qa_throttle.py", line 61, in test_rx_rate_tag
    self.assertLess(total_time, 0.6)
AssertionError: 0.6404304 not less than 0.6

```

PR #4463, opened 2021-03-25
Remove things marked as deprecated

mormj:
- [x] digital: equalizers
- [x] analog: sig source freq port
- [x] blocks: tcp/udp
- [x] filter: mmse_interpolator
- [x] digital: simple_framer
- [x] digital: simple_framer_correlator
- [x] digital: correlate_access_code_bb

(do these in a future PR)
- [ ] digital: clock sync blocks
- [ ] digital: psk mod/demod
- [ ] digital: qam mod/demod
- [ ] analog: cpfsk (needs replacement)

More still needs to be done here to qa tests and examples with the replacement blocks
gr-network needs qa for tcp/udp blocks (removed from gr-blocks)

PR #6273, opened 2022-10-20
CI: trim the list of skipped tests in Conda builds

argilo:
## Description
Conda builds skip many CI tests. Some of the tests do still fail, but others now pass reliably because bugs have been fixed. Here I've cleaned up the list of skipped tests:

* `qa_cpp_py_binding`, `qa_cpp_py_binding_set`, `qa_ctrlport_probes` → not executed because ControlPort is not enabled
* `qa_rotator_cc` → fixed in https://github.com/gnuradio/gnuradio/pull/5755
* `qa_tcp_server_sink` → removed in https://github.com/gnuradio/gnuradio/pull/4463
* `test_modtool` → only fails on Windows
* `qa_add_system_time` → fixed in https://github.com/gnuradio/gnuradio/pull/5749

CI tests in Conda builds

## Testing Done
I had a look over recent Conda builds to see which tests are still failing. None of these tests failed in the past five runs.

## Checklist

### DIFF
--- a/.packaging/conda_recipe/bld.bat
+++ b/.packaging/conda_recipe/bld.bat
@@ -33,8 +33,6 @@ qa_agc^
 |qa_file_descriptor_source_sink^
 |qa_file_sink^
 |qa_file_source^
-|qa_rotator_cc^
-|qa_tcp_server_sink^
 |qa_wavfile^
 |test_modtool^
 %=EMPTY=%
--- a/.packaging/conda_recipe/build.sh
+++ b/.packaging/conda_recipe/build.sh
@@ -19,20 +19,11 @@ cmake --build . --config Release --target install
 
 if [[ $target_platform == linux* ]] ; then
     SKIP_TESTS=(
-        qa_cpp_py_binding
-        qa_cpp_py_binding_set
-        qa_ctrlport_probes
         qa_qtgui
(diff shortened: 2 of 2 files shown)


ryanvolz (approved):
I've had this at the back of my mind to do at some point, so thanks!

argilo:
> `test_modtool` → only fails on Windows

I did a bit more digging, and the failure on Windows is `ModuleNotFoundError: No module named 'readline'`. The readline module doesn't exist on Windows, so https://github.com/gnuradio/gnuradio/pull/1529 would have caused that failure.

Whenever that is fixed, the `test_modtool` skip for Windows can be removed at the same time.

/cc @marcusmueller

ryanvolz:
> I did a bit more digging, and the failure on Windows is `ModuleNotFoundError: No module named 'readline'`. The readline module doesn't exist on Windows, so #1529 would have caused that failure.

A cursory search turns up that `pyreadline3` might be a suitable substitute that can be used on Windows (and is available through conda-forge). Or it might be better to skip adding a dependency and just disable readline functionality when it can't be imported (if that can be done reasonably, I haven't looked).

argilo:
Rebased to resolve merge conflicts.

PR #6274, opened 2022-10-21
analog: fix agc3 on windows, logical bugs

marcusmueller:
* Windows implementation had several bugs
  * inverted "is this the first items" logic
  * inverted "1/magnitude finite" logic
  * allocations/deallocations every work()
  * specifying alignment of std::vector header doesn't align data
* non-windows / GNU implementation had bug
  * Variable-length arrays in C++ are GCC extension
* both versions had several bugs, both hard and perf
  * thread-safety of setters of floating point numbers
  * first-run work code's numerical output values depend on workload
    size (but we're in output_multiple mode, so easy to fix)
  * counter-productive cache thrashing to improve vectorization at the
    expense of having VLA/runtime allocation
  * unnecessary division in comparison
* AGC3 test on windows was disabled, which hid the above bugs

By removing one unnecessary hot-loop division, at worst-case identical
performance replace VOLK's invsqrt with a plain sqrt, by which I divide.

Introduced mutable setter lock.

gr-analog's agc3; used in a few examples

## Testing Done

AGC3 test passes under Linux, enabled test for anaconda windows CI
## Checklist

### DIFF
--- a/gr-analog/lib/agc3_cc_impl.cc
+++ b/gr-analog/lib/agc3_cc_impl.cc
@@ -8,16 +8,16 @@
  *
  */
 
-#ifdef HAVE_CONFIG_H
-#include "config.h"
-#endif
-
-#include <cfloat>
-#include <vector>
-
 #include "agc3_cc_impl.h"
+
 #include <gnuradio/io_signature.h>
+#include <spdlog/fmt/fmt.h>
 #include <volk/volk.h>
+#include <volk/volk_alloc.hh>
+#include <algorithm>
+#include <cmath>
+#include <numeric>
+#include <vector>
 
 namespace gr {
 namespace analog {
@@ -40,14 +40,15 @@ agc3_cc_impl::agc3_cc_impl(float attack_rate,
     : sync_block("agc3_cc",
                  io_signature::make(1, 1, sizeof(gr_complex)),
                  io_signature::make(1, 1, sizeof(gr_complex))),
-      d_attack(attack_rate),
-      d_decay(decay_rate),
-      d_reference(reference),
-      d_gain(gain),
-      d_max_gain(65536),
-      d_reset(true),
-      d_iir_update_decim(iir_update_decim)
+      d_reset(true)
 {
+    set_reference(reference);
+    set_attack_rate(attack_rate);
+    set_decay_rate(decay_rate);
+    set_gain(gain);
+    set_max_gain(65536);
+    test_and_log_value_domain(iir_update_decim, "input power sampling stride");
+    d_iir_update_decim = iir_update_decim;
     set_output_multiple(iir_update_decim * 4);
     const int alignment_multiple = volk_get_alignment() / sizeof(gr_complex);
     set_alignment(std::max(1, alignment_multiple));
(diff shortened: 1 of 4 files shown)


marcusmueller:
Re: backporting: uses C++17 to avoid MSVC specifics.

argilo:
`test_006` now fails on several platforms because the last four samples don't make it through to the vector sink. https://github.com/gnuradio/gnuradio/pull/5244 was supposed to solve that problem, but the solution looks kludgy to me. If it's true that there's no guarantee that all samples will pass through the flowgraph, then the test shouldn't try to use a "magic" number of samples to make it more likely that they will pass. It could instead just look at the last 100 samples, i.e. `result = [abs(x) for x in dst_data[-M:]]`. But I wonder if there's actually a scheduler bug here that should be fixed. Why shouldn't we expect all samples to pass through this particular flowgraph?

gr-analog/python/analog/qa_agc.py
```diff
@@ -473,6 +473,8 @@ def test_006(self):
         dst_data = dst1.data()
         M = 100
         result = [abs(x) for x in dst_data[N - M:]]
+        for idx, (x, y) in enumerate(zip(result, M * [ref, ])):
+            self.assertAlmostEqual(x, y, 4, f"failed at pos {idx}")
```
argilo, line 477: This looks like it's duplicating the check on the following line (with a friendlier error message). If that's the case, we could delete the next assertion.

Also, this could be simplified:
```suggestion
        for idx, x in enumerate(result)):
            self.assertAlmostEqual(x, ref, 4, f"failed at pos {idx}")
```

gr-analog/lib/agc3_cc_impl.cc
```diff
@@ -56,81 +56,137 @@ agc3_cc_impl::agc3_cc_impl(float attack_rate,
...
-        d_gain = d_reference * (noutput_items / mag);
+        gain = reference * (iir_update_decim * 4 / magnitude);
 
-        if (d_gain < 0.0)
-            d_gain = 10e-5;
-
-        if (d_max_gain > 0.0 && d_gain > d_max_gain) {
-            d_gain = d_max_gain;
-        }
+        std::clamp(gain, 1e-4f, max_gain);
```
argilo, line 143: Switching to `std::clamp` seems like it's changing the behaviour. For instance, the documentation says that `max_gain = 0` means unlimited, but that will no longer be the case. Also, the previous version only set the gain to `10e-5` in the event that `d_gain` was negative. (I'm not even sure how that could happen, unless `reference` was negative!)

gr-analog/lib/agc3_cc_impl.cc
```diff
@@ -56,81 +56,137 @@ agc3_cc_impl::agc3_cc_impl(float attack_rate,
...
+    {
+        gr::thread::scoped_lock guard(d_setter_mutex);
+        attack = d_attack;
+        decay = d_decay;
+        reference = d_reference;
+        gain = d_gain;
+        max_gain = d_max_gain;
+        iir_update_decim = d_iir_update_decim;
+        d_gain_change = false; // if gain setting happened, we've just used the new gain.
+    }
```
argilo, line 135: Is there really much to be gained by reducing the scope of locking? The requisite `d_gain_change` makes the code more complex.

marcusmueller:
Interestingly, the test passes on my F35 and F36 machines! But it's clearly a logical bug. We do set the output multiple, so not getting a number that's not a multiple of that is OK. Need to fix that.

Also, need to be more awake next time I'm attempting to fix a block.

gr-analog/lib/agc3_cc_impl.cc
argilo: The fact that this change didn't break any tests also makes it clear that the `max_gain = 0` behaviour is untested.

argilo:
> We do set the output multiple, so not getting a number that's not a multiple of that is OK.

Here the output multiple is 4 and the input length is a multiple of 4, so I don't see any reason for samples to disappear.

marcusmueller:
hm, true. I had in mind the default decim was higher, but it's 1. Hmmmmm

gr-analog/python/analog/qa_agc.py
marcusmueller: did the simplification.

gr-analog/lib/agc3_cc_impl.cc
marcusmueller: probably not, you're right. let me uncruftify my code.

gr-analog/lib/agc3_cc_impl.cc
marcusmueller: I'm also not really sure the default max_gain makes any sense, 2¹⁶ seems very arbitrary.

marcusmueller: also, the setting of a gain to 1e-4 in case the reference was negative, um, let's simply get rid of that: 1e-4 might be an enormous gain in some settings, and we'd have the most funky behaviour. A negative reference (power!) makes no sense, so we reject it in the setter.

marcusmueller:
Doing a test run; will need to squash.

marcusmueller:
The testg

> > We do set the output multiple, so not getting a number that's not a multiple of that is OK.
> 
> Here the output multiple is 4 and the input length is a multiple of 4, so I don't see any reason for samples to disappear.

a hint might be that the head comes before the agc block.

argilo:
True, I wonder if the Head block might be the source of the problem.

marcusmueller:
Indeed, putting the head block after the AGC solved the issue..  this warrants separate investigation.

ryanvolz (approved):

gr-analog/lib/agc3_cc_impl.cc
```diff
@@ -56,81 +55,109 @@ agc3_cc_impl::agc3_cc_impl(float attack_rate,
...
+}
+void agc3_cc_impl::set_decay_rate(float rate)
+{
+    gr::thread::scoped_lock guard(d_setter_mutex);
+    d_decay = rate;
+}
+void agc3_cc_impl::set_reference(float reference)
+{
+    if (reference <= 0.0f) {
+        d_logger->error("Can't set non-negative reference power {:g}", reference);
```
argilo, line 97:
```suggestion
        d_logger->error("Can't set non-positive reference power {:g}", reference);
```

argilo:
The code changes look good to me, apart from that one type in the log message.

No more `#ifdef`s! :tada:

I haven't tried executing the block yet, but hopefully I'll be able to take it for a spin tomorrow.

gr-analog/lib/agc3_cc_impl.cc
```diff
@@ -40,97 +40,135 @@ agc3_cc_impl::agc3_cc_impl(float attack_rate,
...
     set_alignment(std::max(1, alignment_multiple));
 }
 
 agc3_cc_impl::~agc3_cc_impl() {}
 
+void agc3_cc_impl::test_and_log_value_domain(float value, std::string_view description)
+{
+    if (value <= 0.0f) {
+        d_logger->error(
+            "Can't set {} power that is not strictly positive: {:g}", description, value);
```
argilo, line 63: I am guessing that "power" was not meant to be here:
```suggestion
            "Can't set {} that is not strictly positive: {:g}", description, value);
```

gr-analog/lib/agc3_cc_impl.cc
```diff
@@ -40,97 +40,135 @@ agc3_cc_impl::agc3_cc_impl(float attack_rate,
...
+}
+void agc3_cc_impl::set_gain(float gain)
+{
+    test_and_log_value_domain(gain, "gain");
+    gr::thread::scoped_lock guard(d_setter_mutex);
+    d_gain = gain;
+}
+void agc3_cc_impl::set_max_gain(float max_gain)
+{
+    test_and_log_value_domain(max_gain, "maximum gain");
```
argilo, line 120: This would block `max_gain = 0` (for unlimited), wouldn't it?

argilo:
A couple small issues still left, but the code is generally looking good.

It's unfortunate that `test_006_001_agc3` tests so little: As written, I expect that a Constant Source block could pass the test. :smile: It would be nice to pass in a more realistic signal that exercises the reset, attack & decay conditions, and verify that all the output samples match (to within some tolerance) the expected values. That would also help to confirm that the old & new code works identically. I don't insist on having that in this PR, but it would be a "nice to have."

argilo:
> It would be nice to pass in a more realistic signal

What I had in mind here was something like a sine wave in three pieces, with the first piece having high amplitude, the second piece having low amplitude, and the third piece having high amplitude again.

Even better, include a piece with zero amplitude, and one with an amplitude low enough to trigger the "max gain" condition.

marcusmueller:
> but it would be a "nice to have."

I disagree, it should be expected that it is there. I did not come around to defining a good test case yet, but I should.

gr-analog/lib/agc3_cc_impl.cc
argilo: I guess there should at least be a test that `max_gain = 0` is accepted, and ideally that it behaves as expected.

mormj:
@marcusmueller - did you want to hold off merging this until an updated test is ready, or add the test as an issue to be addressed separately?

marcusmueller:
@mormj I really wanted to add these tests, but seeing for how long I'm blocking this: let's have them separately.

argilo:
Another bug I noticed while testing this is that `max_gain` is only enforced in the "reset" case and not afterwards. But that bug has existed for as long as the `max_gain` setting has been in AGC3 (5bdd638f56775e50cce23ffe3bef241de53dcad1) so it's probably best to leave that for a separate PR if we want to change the behaviour.

argilo (approved):
The code looks good. I also took this for a spin and all the features work as expected (with the exception of Max Gain, which I think should be considered in its own pull request). Performance is a bit worse (27% CPU at 10 Msps vs 21% CPU on `main`), but I think correctness is more important than performance.

willcode:
Since the `scoped_lock` is locking only one mutex, is it reasonable to backport using `lock_guard` instead?

Is there anything else C++17 in here?

willcode:
Hmm, we already have 399 instances of `scoped_lock` in the maint-3.10 codebase. I don't think a couple more will hurt!

PR #6275, opened 2022-10-21
blocks: Fix file tests to work on Windows

argilo:
## Description
Several file-related tests fail on Windows with "Permission denied" errors:

```
ERROR: test_file_descriptor (__main__.test_file_descriptor_source_sink)
  File "%SRC_DIR%\gr-blocks\python\blocks\qa_file_descriptor_source_sink.py", line 33, in test_file_descriptor
    fhandle0 = open(temp.name, "wb")
PermissionError: [Errno 13] Permission denied: 'C:\\Users\\RUNNER~1\\AppData\\Local\\Temp\\tmpmfjxx9mx'

ERROR: test_file_sink (__main__.test_file_sink)
  File "%SRC_DIR%\gr-blocks\python\blocks\qa_file_sink.py", line 32, in test_file_sink
    snk = blocks.file_sink(gr.sizeof_float, temp.name)
RuntimeError: can't open file

ERROR: setUpClass (__main__.test_file_source)
  File "%SRC_DIR%\gr-blocks\python\blocks\qa_file_source.py", line 26, in setUpClass
    with open(cls._datafilename, 'wb') as f:
PermissionError: [Errno 13] Permission denied: 'C:\\Users\\RUNNER~1\\AppData\\Local\\Temp\\tmp9haek5tv'

```

This happens because the tests use `tempfile.NamedTemporaryFile` but do not close the file before it is opened in other places. The [documentation](https://docs.python.org/3/library/tempfile.html#tempfile.NamedTemporaryFile) states:

> Whether the name can be used to open the file a second time, while the named temporary file is still open, varies across platforms (it can be so used on Unix; it cannot on Windows).

Therefore, it is necessary to close temporary files before they are used in flowgraphs. The `delete=False` argument is necessary to prevent the file from being automatically deleted after being closed. This means that the file must be cleaned up manually (with `os.unlink()`) after the flowgraphs are executed.

With the problematic tests fixed, we can remove them from the list of skipped tests in Conda builds.

Tests for File Source, File Sink, File Descriptor Source, and File Descriptor Sink.

## Testing Done
I verified that the tests still pass on my machine.

## Checklist

### DIFF
--- a/.packaging/conda_recipe/bld.bat
+++ b/.packaging/conda_recipe/bld.bat
@@ -30,9 +30,6 @@ set SKIP_TESTS=^
 qa_agc^
 |qa_dtv^
 |qa_fecapi_ldpc^
-|qa_file_descriptor_source_sink^
-|qa_file_sink^
-|qa_file_source^
 |qa_wavfile^
 |test_modtool^
 %=EMPTY=%
--- a/gr-blocks/python/blocks/qa_file_descriptor_source_sink.py
+++ b/gr-blocks/python/blocks/qa_file_descriptor_source_sink.py
@@ -19,9 +19,13 @@ class test_file_descriptor_source_sink(gr_unittest.TestCase):
     def setUp(self):
         os.environ['GR_CONF_CONTROLPORT_ON'] = 'False'
         self.tb = gr.top_block()
+        temp = tempfile.NamedTemporaryFile(delete=False)
+        temp.close()
+        self._datafilename = temp.name
 
     def tearDown(self):
         self.tb = None
+        os.unlink(self._datafilename)
 
     def test_file_descriptor(self):
         src_data = range(1000)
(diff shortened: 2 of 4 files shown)


argilo:
Note: It may be helpful to ignore whitespace changes when reviewing this pull request, since the indentation level of some blocks of code has changed:

https://github.com/gnuradio/gnuradio/pull/6275/files?w=1

argilo:
Oops, this didn't fix all the test failures on Windows. I'll investigate.

ryanvolz:
Mysterious that it still complains about files still being opened/used by another process. `qa_file_source` passes though, and that's the only one that creates/deletes the temp file in the test setup and teardown methods, so maybe switching to that structure for the other two tests would fix the errors?

argilo:
You're correct. After some investigation I determined that the blocks don't actually close their files until they are destroyed (even if `close()` is called — it seems to close `d_new_fp` rather than `d_fp`, which may be a bug of its own). I made exactly the change you proposed and it seems to work. We'll see how CI goes. :crossed_fingers:

argilo:
I also had to change the file descriptor test to use `os.open` instead of `open`, because a crash occurs when both Python and C++ try to close the same file descriptor. Now the C++ side is responsible for closing the file descriptors (when the blocks are destroyed).

argilo:
Those changes did the trick. The tests now pass on my fork, so they should pass here as well.

This pull request is ready for final review now.

ryanvolz (approved):

argilo:
Rebased to resolve the merge conflict in bld.bat.

argilo:
It looks like the "Check C++ Formatting" job has hung. I re-pushed to kick off a new CI run.

That's the second time this job has hung recently. I'll have a look to see if there's any obvious bugs. At a minimum, the timeout should be lowered from the default 360 minutes.

PR #6290, opened 2022-10-24
WIP: 4.0: Blocks necessary for FM receiver

mormj:
This PR is WIP, and attempting to determine what is the best way to handle the lack of `history()` in the current 4.0 API

Currently the QA tests surrounding filters don't pass because there are differences in how the filter blocks work between 3.x and 4.0

### Implications:

The lack of `history()` comes into play in several places:
1) How should buffers get sized appropriately
 a) If a block has a large history requirement - like many taps in a filter - it can never produce samples in some cases given equal input and output buffers
2) What input/output sizes should be given to the block in the executor (since forecasting also gone)
3) Group delay and phase of the filters
 a) the work function start by receiving the first sample, not `history()-1` samples of 0's - so there are no longer the expected startup transients without putting that manually in the block (first call to `work()` copies zeros then `in` into a temp buff and calls the filter kernel on that) 
 b) the qa tests fail because of this

The current solution in this PR is to have a property similar to history called noconsume - this effectively means history, but the extra samples are not given to the block.  Instead this just informs the scheduler that N samples will not be consumed from the input implying that relative_rate() * noutput_items +noconsume should always be handed in.  
**NOTE: I don't think this is ideal**

### Other questions that arise:
1) Should the scheduler have to ensure *any* of the block API contracts such as `output_multiple()` - can't a block just take the max number of samples available from the scheduler and restrict itself
2) What convenience methods could make forecasting, size constraints, buffer size requirements easier on both the scheduler and the block developer


mormj:
updated in #6306

PR #6302, opened 2022-10-27
uhd: add async msgs port to usrp_source block and publish overflows

ph1t0:
Signed-off-by: ph1t0 <fito@ektocomms.space>

## Description

Overflows occurs when the host does not consume data fast enough. 
It is very useful to have and explicit message (not only a log) from the `usrp_source` block every time an overflow occurs. 
This feature allows the user to properly handle the event.

gr-uhd's usrp_source block.

## Testing Done

I've compiled and installed gnuradio main branch with the changes proposed. 
Then I've executed the following piece of code:

```python
#!/usr/bin/env python3
# -*- coding: utf-8 -*-

from gnuradio import blocks, gr, uhd


class OverflowTest(gr.top_block):
    def __init__(self):
        gr.top_block.__init__(self, self.__class__.__name__, catch_exceptions=True)

        self.sample_rate = 1e6

        # Blocks
        self.uhd_source = uhd.usrp_source("", uhd.stream_args(cpu_format="fc32", channels=range(1)))
        self.uhd_source.set_samp_rate(self.sample_rate)
        self.throttle = blocks.throttle(gr.sizeof_gr_complex*1, self.sample_rate / 10, True)
        self.null_sink = blocks.null_sink(gr.sizeof_gr_complex*1)
        self.print_msg = blocks.message_debug()

        # Connections
        self.connect(self.uhd_source, self.throttle, self.null_sink)
        self.msg_connect((self.uhd_source, 'async_msgs'), (self.print_msg, 'print'))

if __name__ == '__main__':
    tb = OverflowTest()
    tb.start()

    try:
        input('Press Enter to quit: ')
    except EOFError:
        pass
    tb.stop()
    tb.wait()
```
```bash
# python /tmp/test_overflow_msg.py 
[INFO] [UHD] linux; GNU C++ version 12.2.0; Boost_108000; UHD_4.3.0.HEAD-0-g1f8fd345
[INFO] [B200] Detected Device: B200
[INFO] [B200] Operating over USB 3.
[INFO] [B200] Initialize CODEC control...
[INFO] [B200] Initialize Radio control...
[INFO] [B200] Performing register loopback test... 
[INFO] [B200] Register loopback test passed
[INFO] [B200] Setting master clock rate selection to 'automatic'.
[INFO] [B200] Asking for clock rate 16.000000 MHz... 
[INFO] [B200] Actually got clock rate 16.000000 MHz.
[INFO] [B200] Asking for clock rate 32.000000 MHz... 
[INFO] [B200] Actually got clock rate 32.000000 MHz.
Press Enter to quit: Ousrp_source :error: In the last 1111 ms, 1 overflows occurred.
(uhd_async_msg (overflow))
usrp_source :error: In the last 778 ms, 1 overflows occurred.
O******* MESSAGE DEBUG PRINT ********
(uhd_async_msg (overflow))
Ousrp_source :error: In the last 778 ms, 1 overflows occurred.
(uhd_async_msg (overflow))
O******* MESSAGE DEBUG PRINT ********
(uhd_async_msg (overflow))
Ousrp_source :error: In the last 1515 ms, 2 overflows occurred.
(uhd_async_msg (overflow))
Ousrp_source :error: In the last 777 ms, 1 overflows occurred.
(uhd_async_msg (overflow))
O******* MESSAGE DEBUG PRINT ********
(uhd_async_msg (overflow))
Ousrp_source :error: In the last 1515 ms, 2 overflows occurred.
(uhd_async_msg (overflow))
Ousrp_source :error: In the last 777 ms, 1 overflows occurred.
```

## Checklist
- [x] ~~I have added tests to cover my changes, and~~ all previous tests pass.

### DIFF
--- a/gr-uhd/grc/gen_uhd_usrp_blocks.py
+++ b/gr-uhd/grc/gen_uhd_usrp_blocks.py
@@ -119,10 +119,10 @@ outputs:
 % if sourk == 'sink':
 
 outputs:
+% endif
 -   domain: message
     id: async_msgs
     optional: true
-% endif
 
 templates:
     imports: |-
--- a/gr-uhd/lib/usrp_source_impl.cc
+++ b/gr-uhd/lib/usrp_source_impl.cc
@@ -49,6 +49,7 @@ usrp_source_impl::usrp_source_impl(const ::uhd::device_addr_t& device_addr,
 
     _samp_rate = this->get_samp_rate();
     _samps_per_packet = 1;
+    message_port_register_out(ASYNC_MSGS_PORT_KEY);
     register_msg_cmd_handler(cmd_tag_key(),
                              [this](const pmt::pmt_t& tag, const int, const pmt::pmt_t&) {
                                  this->_cmd_handler_tag(tag);
@@ -647,6 +648,12 @@ int usrp_source_impl::work(int noutput_items,
                 std::chrono::duration_cast<std::chrono::milliseconds>(delta).count();
             d_logger->error(
                 "In the last {:d} ms, {:d} overflows occurred.", ms, _overflow_count);
+
+            pmt::pmt_t value = pmt::dict_add(
+                pmt::make_dict(), EVENT_CODE_OVERFLOW, pmt::from_uint64(_overflow_count));
+            pmt::pmt_t msg = pmt::cons(ASYNC_MSG_KEY, value);
+            message_port_pub(ASYNC_MSGS_PORT_KEY, msg);
+
             _overflow_count = 0;
         }
         // ignore overflows and try work again
--- a/gr-uhd/lib/usrp_source_impl.h
+++ b/gr-uhd/lib/usrp_source_impl.h
@@ -21,6 +21,9 @@
 static const pmt::pmt_t TIME_KEY = pmt::string_to_symbol("rx_time");
 static const pmt::pmt_t RATE_KEY = pmt::string_to_symbol("rx_rate");
 static const pmt::pmt_t FREQ_KEY = pmt::string_to_symbol("rx_freq");
+static const pmt::pmt_t ASYNC_MSGS_PORT_KEY = pmt::string_to_symbol("async_msgs");
+static const pmt::pmt_t ASYNC_MSG_KEY = pmt::string_to_symbol("uhd_async_msg");
+static const pmt::pmt_t EVENT_CODE_OVERFLOW = pmt::string_to_symbol("overflows");
 
 namespace gr {
 namespace uhd {


gr-uhd/grc/gen_uhd_usrp_blocks.py
```diff
@@ -119,10 +119,10 @@
 % if sourk == 'sink':
 
 outputs:
+% endif
```
willcode, line 122: Would the `if/endif` would be removed entirely in this case?

gr-uhd/lib/usrp_source_impl.cc
```diff
@@ -639,6 +640,12 @@ int usrp_source_impl::work(int noutput_items,
     case ::uhd::rx_metadata_t::ERROR_CODE_OVERFLOW: {
         _tag_now = true;
         ++_overflow_count;
+
+        const pmt::pmt_t val = pmt::PMT_NIL;
```
willcode, line 644: These messages, if very frequent, could flood the message queue. They should still be limited, same as the log messages.

willcode:
If we end up adding something like this, we should also look at adding it to gr-soapy, to keep the behavior similar. There's a bigger question of whether we want errors of various types to be sent as messages.

mbr0wn:
At first glance, I like this (@willcode's comments notwithstanding). Need to review in more detail.

gr-uhd/grc/gen_uhd_usrp_blocks.py
ph1t0: You still need that `if/endif`. Otherwise there will be two `outputs` labels.
The `stream` port is an input for the sink and an output for the source.

willcode: The UHD yml is being tricky at the expense of readability ... OK, not your problem.

gr-uhd/lib/usrp_source_impl.cc
ph1t0: ok! If that is the case maybe this could work:

```diff
diff --git a/gr-uhd/lib/usrp_source_impl.cc b/gr-uhd/lib/usrp_source_impl.cc
index fbd008df2..8a0f6b44b 100644
--- a/gr-uhd/lib/usrp_source_impl.cc
+++ b/gr-uhd/lib/usrp_source_impl.cc
@@ -641,11 +641,6 @@ int usrp_source_impl::work(int noutput_items,
         _tag_now = true;
         ++_overflow_count;
 
-        const pmt::pmt_t val = pmt::PMT_NIL;
-        pmt::pmt_t value = pmt::dict_add(pmt::make_dict(), EVENT_CODE_OVERFLOW, val);
-        pmt::pmt_t msg = pmt::cons(ASYNC_MSG_KEY, value);
-        message_port_pub(ASYNC_MSGS_PORT_KEY, msg);
-
         auto now = std::chrono::steady_clock::now();
         auto delta = now - _last_log;
         if (delta > _overflow_log_interval) {
@@ -654,6 +649,13 @@ int usrp_source_impl::work(int noutput_items,
                 std::chrono::duration_cast<std::chrono::milliseconds>(delta).count();
             d_logger->error(
                 "In the last {:d} ms, {:d} overflows occurred.", ms, _overflow_count);
+
+            pmt::pmt_t value = pmt::dict_add(pmt::make_dict(),
+                                             EVENT_CODE_OVERFLOW,
+                                             pmt::from_uint64(_overflow_count));
+            pmt::pmt_t msg = pmt::cons(ASYNC_MSG_KEY, value);
+            message_port_pub(ASYNC_MSGS_PORT_KEY, msg);
+
             _overflow_count = 0;
         }
         // ignore overflows and try work again
diff --git a/gr-uhd/lib/usrp_source_impl.h b/gr-uhd/lib/usrp_source_impl.h
index e883c9e3c..8cd986eb8 100644
--- a/gr-uhd/lib/usrp_source_impl.h
+++ b/gr-uhd/lib/usrp_source_impl.h
@@ -23,7 +23,7 @@ static const pmt::pmt_t RATE_KEY = pmt::string_to_symbol("rx_rate");
 static const pmt::pmt_t FREQ_KEY = pmt::string_to_symbol("rx_freq");
 static const pmt::pmt_t ASYNC_MSGS_PORT_KEY = pmt::string_to_symbol("async_msgs");
 static const pmt::pmt_t ASYNC_MSG_KEY = pmt::string_to_symbol("uhd_async_msg");
-static const pmt::pmt_t EVENT_CODE_OVERFLOW = pmt::string_to_symbol("overflow");
+static const pmt::pmt_t EVENT_CODE_OVERFLOW = pmt::string_to_symbol("overflows");
 
 namespace gr {
 namespace uhd {
```

mbr0wn: I think your solution is good -- can you please apply it? I'll unresolve this thread until you do.

mbr0wn: BTW, you can use the github suggestion feature to show changes inline.

mbr0wn (approved):

Issue #6300, opened 2022-10-27
s390x test failures


argilo:
The Debian build logs at https://buildd.debian.org/status/package.php?p=gnuradio show that many tests currently fail on the s390x architecture. Many of the tests exit with segmentation faults, and many others fail with what appear to be endianness bugs.

Although it's unlikely that anyone is running GNU Radio on this architecture, the test failures very likely point to places where GNU Radio depends on undefined behaviour. These bugs could well lead to failures on other architectures in the future, so I think they should be fixed.

Here's a full log of the test run from 3.10.4.0-3 on the s390x architecture:
```
        Start   1: runtime_qa_buffer.cc
  1/261 Test   #1: runtime_qa_buffer.cc .........................   Passed    1.84 sec
        Start   2: runtime_qa_io_signature.cc
  2/261 Test   #2: runtime_qa_io_signature.cc ...................   Passed    0.03 sec
        Start   3: runtime_qa_logger.cc
  3/261 Test   #3: runtime_qa_logger.cc .........................   Passed    0.02 sec
        Start   4: runtime_qa_host_buffer.cc
  4/261 Test   #4: runtime_qa_host_buffer.cc ....................   Passed    0.04 sec
        Start   5: runtime_qa_vmcircbuf.cc
  5/261 Test   #5: runtime_qa_vmcircbuf.cc ......................   Passed    0.10 sec
        Start   6: math_qa_fxpt.cc
  6/261 Test   #6: math_qa_fxpt.cc ..............................   Passed    0.03 sec
        Start   7: math_qa_fxpt_nco.cc
  7/261 Test   #7: math_qa_fxpt_nco.cc ..........................   Passed    0.14 sec
        Start   8: math_qa_fxpt_vco.cc
  8/261 Test   #8: math_qa_fxpt_vco.cc ..........................   Passed    0.13 sec
        Start   9: math_qa_math.cc
  9/261 Test   #9: math_qa_math.cc ..............................   Passed    0.01 sec
        Start  10: math_qa_sincos.cc
 10/261 Test  #10: math_qa_sincos.cc ............................   Passed    0.01 sec
        Start  11: math_qa_fast_atan2f.cc
 11/261 Test  #11: math_qa_fast_atan2f.cc .......................   Passed    0.03 sec
        Start  12: pmt_prims
 12/261 Test  #12: pmt_prims ....................................   Passed    0.49 sec
        Start  13: pmt_unv
[...]

PR #3425, opened 2020-04-28
pmt: add absolute pmt serialization QA

mrjacobagilbert:
currently there is nothing to check the absolute accuracy of PMT serialization, and it is easy for changes to break this provided serialize_str() and deserialize_str() are inverse functions. this provides some absolute reference PMT strings to ensure compatibility is retained.

Note: this version accounts for the serialization compatibility breakage i introduced in #3320. I will backport this onto 3.8 and will use the old serialization strings.

~~Note2: #3389 breaks this again, as dicts now need a different identifier from pairs. as soon as that is pulled in-tree, i will force-push a change to this to update accordingly.~~

Ready to merge.

PR #6307, opened 2022-10-27
runtime: Fix PMT serialization of C32 vectors

argilo:
## Description
#3320 made some changes to the PMT serialization format:

* Prior to the change, 32-bit complex vectors were serialized as (real, imag) big-endian doubles
* After the change (which went into the 3.9.0.0 release), 32-bit complex vectors were serialized as:
  * (imag, real) big-endian floats on little-endian systems
  * (real, imag) big-endian floats on big-endian systems

The platform-dependent behaviour causes the corresponding absolute serialization test (added in #3425) to fail on big-endian systems because it expects the (imag, real) ordering:

```
FAIL: test23_absolute_serialization_float_uvecs (__main__.test_pmt)
  File "/<<PKGBUILDDIR>>/gnuradio-runtime/python/pmt/qa_pmt.py", line 350, in test23_absolute_serialization_float_uvecs
    self.assertEqual(out_str, in_str)
AssertionError: b'\n\n\x00\x00\x00\x02\x01\x00K\x81\x01\x82\xbf\x80\x00\x00@I\x0eVB\xc7\xfa\xe1' != b'\n\n\x00\x00\x00\x02\x01\x00\xbf\x80\x00\x00K\x81\x01\x82B\xc7\xfa\xe1@I\x0eV'

```

I think the platform-dependent ordering (which is out of order on the most commonly used platforms) would be confusing to users, so I'm proposing to change it back to (real, imag) on all platforms.

I've also updated the test, replacing the inscrutable byte arrays with `struct.pack()`.

## Related Issue
* #6300

PMT serialization of 32-bit complex vectors.

## Testing Done
I verified that the tests pass on my amd64 system. ~~I'll try them on an s390x system shortly.~~ They also pass on a s390x system.

## Checklist

### DIFF
--- a/gnuradio-runtime/lib/pmt/pmt_serialize.cc
+++ b/gnuradio-runtime/lib/pmt/pmt_serialize.cc
@@ -528,10 +528,8 @@ tail_recursion:
             for (size_t i = 0; i < npad; i++) {
                 ok &= serialize_untagged_u8(0, sb);
             }
-            // Note that if endianness causes byte swap that real/imag will also be
-            // swapped
-            ok &= serialize_untagged_u64_array(
-                (uint64_t*)&c32vector_elements(obj)[0], vec_len, sb);
+            ok &= serialize_untagged_u32_array(
+                (uint32_t*)&c32vector_elements(obj)[0], vec_len * 2, sb);
             return ok;
         }
 
@@ -729,9 +727,8 @@ pmt_t deserialize(std::streambuf& sb)
             return vec;
         }
         case (UVI_C32): {
-            // Data was serialized as uint64, so do the same here
-            deserialize_untagged_u64_vector(u64v, nitems, sb);
-            pmt_t vec = init_c32vector(nitems, (std::complex<float>*)&u64v[0]);
+            deserialize_untagged_u32_vector(u32v, 2 * nitems, sb);
+            pmt_t vec = init_c32vector(nitems, (std::complex<float>*)&u32v[0]);
             return vec;
         }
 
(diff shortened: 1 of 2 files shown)


willcode:
Backward compatibility opinions on this one? We haven't really addressed runtime data compatibility, and maybe something would already break if a 3.8 flowgraph sent a PMT via ZMQ to a 3.10 flowgraph?

jacobagilbert:
I think the bigger compatibility issue is that earlier 3.10 releases may be runtime incompatible with this... That said this does seem like a problem...

argilo:
> and maybe something would already break if a 3.8 flowgraph sent a PMT via ZMQ to a 3.10 flowgraph?

Yes, 3.8 uses a different format than 3.9+, so there would already be breakage there. So the question is whether we can allow another breakage (for one data type) to occur between 3.9-3.10.4 and 3.10.5+.

argilo:
The alternatives I can see are:

* delay the fix (and compatibility break) to 4.0
* keep the (imag, real) format forever

If we take either of these options, I think we should still change the format on big-endian systems to match little-endian in the 3.10.5 release. Big-endian systems are rare, so the risk of breakage would be very low.

argilo:
My inclination is to bite the bullet. There's no 3.11 on the horizon, and 4.0 is a long way off. Leaving things as they are would leave a lot of time for the oddball format to become entrenched.

That said, I rarely use PMTs, so I'm not certain how common it would be to send them between different versions of GNU Radio.

jacobagilbert (approved):
👍

willcode (approved):
Agree that backporting is the least bad fix.

PR #6308, opened 2022-10-28
digital: Make the add_field functions portable

argilo:
## Description
The `add_field16`, `add_field32`, and `add_field64` functions have undefined behaviour because they copy the machine representation of multi-byte integers directly into the output buffer. As a result, they do not produce the correct output on big-endian systems. Switching to bit shifts will ensure consistent operation across platforms.

The errors on big-endian systems are as follows:
```
/home/argilo/gnuradio/gr-digital/lib/qa_header_format.cc(71): fatal error: in "test_default_format": critical check 0xAA == (int)h0 has failed [170 != 0]
/home/argilo/gnuradio/gr-digital/lib/qa_header_format.cc(154): fatal error: in "test_counter_format": critical check 0xAA == (int)h0 has failed [170 != 0]

/home/argilo/gnuradio/gr-digital/lib/qa_header_buffer.cc(48): fatal error: in "test_add16": critical check (uint8_t)0xAF == header.header()[0] has failed [0xaf != 0x5c]
/home/argilo/gnuradio/gr-digital/lib/qa_header_buffer.cc(87): fatal error: in "test_add32": critical check (uint8_t)0xAF == header.header()[0] has failed [0xaf != 0x54]
/home/argilo/gnuradio/gr-digital/lib/qa_header_buffer.cc(132): fatal error: in "test_add64": critical check (uint8_t)0xAF == header.header()[0] has failed [0xaf != 0x67]
/home/argilo/gnuradio/gr-digital/lib/qa_header_buffer.cc(193): fatal error: in "test_add_many": critical check (uint8_t)0x01 == header.header()[0] has failed [0x1 != 0x67]
```

## Related Issue
* #6300

* `gr::digital::header_buffer` class
* `gr::digital::header_format_*` classes

## Testing Done
I verified that `digital_qa_header_buffer.cc` and `digital_qa_header_format.cc` now pass on both amd64 (little-endian) and s390x (big-endian) machines.

## Checklist

### DIFF
--- a/gr-digital/lib/header_buffer.cc
+++ b/gr-digital/lib/header_buffer.cc
@@ -57,43 +57,22 @@ void header_buffer::add_field8(uint8_t data, int len, bool bs)
 
 void header_buffer::add_field16(uint16_t data, int len, bool bs)
 {
-    int nbytes = len / 8;
-    if (d_buffer) {
-        uint16_t x = data;
-        if (!bs) {
-            volk_16u_byteswap(&x, 1);
-            x = x >> (16 - len);
-        }
-        memcpy(&d_buffer[d_offset], &x, nbytes);
-        d_offset += nbytes;
-    }
+    add_field64(data, len, bs);
 }
 
 void header_buffer::add_field32(uint32_t data, int len, bool bs)
 {
-    int nbytes = len / 8;
-    if (d_buffer) {
-        uint32_t x = data;
-        if (!bs) {
-            volk_32u_byteswap(&x, 1);
-            x = x >> (32 - len);
-        }
-        memcpy(&d_buffer[d_offset], &x, nbytes);
-        d_offset += nbytes;
-    }
+    add_field64(data, len, bs);
 }
 
 void header_buffer::add_field64(uint64_t data, int len, bool bs)
 {
     int nbytes = len / 8;
     if (d_buffer) {
-        uint64_t x = data;
-        if (!bs) {
-            volk_64u_byteswap(&x, 1);
-            x = x >> (64 - len);
+        for (int i = 0; i < nbytes; i++) {
+            int shift = bs ? i : (nbytes - i - 1);
+            d_buffer[d_offset++] = (data >> (8 * shift)) & 0xff;
         }
-        memcpy(&d_buffer[d_offset], &x, nbytes);
-        d_offset += nbytes;
     }
 }
 


argilo:
It turns out that this also fixes `qa_packet_format`, which formerly failed on big-endian systems with the following errors:

```
FAIL: test_packet_format_async_counter (__main__.test_packet_format_fb)
  File "/<<PKGBUILDDIR>>/gr-digital/python/digital/qa_packet_format.py", line 162, in test_packet_format_async_counter
    self.assertEqual(access_code, rx_access_code)
AssertionError: b'\xac\xdd\xa4\xe2\xf2\x8c \xfc' != b'\xfc \x8c\xf2\xe2\xa4\xdd\xac'

FAIL: test_packet_format_async_default (__main__.test_packet_format_fb)
  File "/<<PKGBUILDDIR>>/gr-digital/python/digital/qa_packet_format.py", line 68, in test_packet_format_async_default
    self.assertEqual(access_code, rx_access_code)
AssertionError: b'\xac\xdd\xa4\xe2\xf2\x8c \xfc' != b'\xfc \x8c\xf2\xe2\xa4\xdd\xac'

```

willcode:
The original code had `if !bs then byteswap`, which seems backward. Was that part of the problem?

Was this actually an alignment problem? I would have thought that VOLK would do exactly what the new code here does.

argilo:
The documentation for `header_buffer` says:

> Each field is a specific length of 8, 16, 32, or 64 bits that are to be transmitted in network byte order. We can adjust the direction of the bytes by setting the byte-swap flag, `bs`, to true or false.

The current code assumes that the machine order is little-endian, which means that a byte swap is needed to get to network order. `bs=true` means little-endian, so no swap is needed in that case.

argilo:
I don't believe there were any alignment problems. It was just that the code failed completely on big-endian systems -- it would read bytes from the wrong end of the integer.

willcode:
OK, so the sense of `bs` was reversed, which was strange, maybe originating as you say from a LE point of view. Either way, the caller needs to know the endianness of the machine in order to fill `bs` correctly. Seems like the param should be `byte_order`, should default to BE, and the library should figure out whether a swap is needed based on machine order.

Why not use VOLK swaps? Are you thinking that the generic shifting is faster than the overhead of calling a VOLK kernel for one item?

argilo:
> OK, so the sense of bs was reversed, which was strange, maybe originating as you say from a LE point of view.

Yeah, from my read of the documentation I think the intent was that `bs=false` outputs big-endian, and `bs=true` outputs little-endian.

> Either way, the caller needs to know the endianness of the machine in order to fill bs correctly.

That's true today, but I don't think there's any reason the user should need to know the machine order.

Also, the current `add_field` functions are completely broken on big-endian systems when the `len` argument is used. For instance, `add_field64(0x1234, 16)` returns `0x0000`. This breaks `header_format_default` and `header_format_counter`, which use the `len` argument when writing out `d_access_code`. So it's clear this code has never been tested on a big-endian system.

> Seems like the param should be `byte_order`

Renaming it would be an API change, since the function is exposed in Python.

> should default to BE

Agreed.

> and the library should figure out whether a swap is needed based on machine order.

I don't think the code should know or care what the machine order is.

> Why not use VOLK swaps?

Because doing so would require knowledge of the machine order, and I don't think the code should know or care about that.

This blog pots makes the case that code that needs to know the machine's byte order is misguided: https://commandcenter.blogspot.com/2012/04/byte-order-fallacy.html

argilo:
> > Why not use VOLK swaps?
>
> Because doing so would require knowledge of the machine order, and I don't think the code should know or care about that.

Actually I suppose that's not true -- it would be possible to write a platform-agnostic version that uses VOLK swaps to switch between the `bs=false` and `bs=true` cases. But it didn't seem worth the trouble of hooking into VOLK for such a simple operation.

argilo:
Because of the `len` argument, it's not as simple as just doing a byte swap; the result also needs to be bit-shifted. The code ended up being more complicated when I tried to use VOLK byteswaps.

mormj (approved):
LGTM and nicely cleaned up!

PR #6312, opened 2022-10-28
network: Replace boost::circular_buffer with gr::buffer

argilo:
## Description
This is a continuation of the effort to remove unnecessary Boost from GNU radio. The UDP Source and UDP Sink blocks use `boost::circular_buffer` to buffer data, but this can be replaced with ~~`std::queue`. The only change that's needed is to check the queue size before each `push()` to check whether the oldest element needs to be discarded.~~ `gr::buffer`. This also allows the local buffers and associated `memcpy` calls to be eliminated.

* UDP Source
* UDP Sink

## Testing Done
So far I've only verified that the tests still pass. I'll take the blocks for a spin soon.

## Checklist

### DIFF
--- a/gr-network/lib/udp_sink_impl.cc
+++ b/gr-network/lib/udp_sink_impl.cc
@@ -105,8 +105,6 @@ udp_sink_impl::udp_sink_impl(size_t itemsize,
 
 bool udp_sink_impl::start()
 {
-    d_localbuffer = new char[d_payloadsize];
-
     long max_circ_buffer;
 
     // Let's keep it from getting too big
@@ -182,11 +180,6 @@ bool udp_sink_impl::stop()
         d_io_context.stop();
     }
 
-    if (d_localbuffer) {
-        delete[] d_localbuffer;
-        d_localbuffer = NULL;
-    }
-
     d_localqueue_reader.reset();
     d_localqueue_writer.reset();
 
@@ -248,16 +241,14 @@ int udp_sink_impl::work(int noutput_items,
                 asio::buffer((const void*)d_tmpheaderbuff, d_header_size));
         }
 
-        // Fill the data buffer
-        memcpy(d_localbuffer, d_localqueue_reader->read_pointer(), d_precomp_datasize);
-        d_localqueue_reader->update_read_pointer(d_precomp_datasize);
-
         // Set up for transmit
         transmitbuffer.push_back(
-            asio::buffer((const void*)d_localbuffer, d_precomp_datasize));
+            asio::buffer(d_localqueue_reader->read_pointer(), d_precomp_datasize));
 
         // Send
         d_udpsocket->send_to(transmitbuffer, d_endpoint);
+
+        d_localqueue_reader->update_read_pointer(d_precomp_datasize);
     }
 
     int itemsreturned = blocks_available * d_precomp_data_overitemsize;
--- a/gr-network/lib/udp_sink_impl.cc
+++ b/gr-network/lib/udp_sink_impl.cc
@@ -119,7 +119,8 @@ bool udp_sink_impl::start()
             max_circ_buffer = d_payloadsize * 1500;
     }
 
-    d_localqueue = new boost::circular_buffer<char>(max_circ_buffer);
+    d_localqueue_writer = gr::make_buffer(max_circ_buffer, sizeof(char), 1, 1);
+    d_localqueue_reader = gr::buffer_add_reader(d_localqueue_writer, 0);
 
     d_udpsocket = new asio::ip::udp::socket(d_io_context);
 
@@ -186,10 +187,8 @@ bool udp_sink_impl::stop()
         d_localbuffer = NULL;
     }
 
-    if (d_localqueue) {
-        delete d_localqueue;
-        d_localqueue = NULL;
-    }
+    d_localqueue_reader.reset();
+    d_localqueue_writer.reset();
 
     return true;
 }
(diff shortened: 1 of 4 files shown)


willcode:
Any way to test performance? A quick look at the boost code tells me that the only possible difference is that circular_buffer is a vector where queue is a deque. I'd imagine that the deque is a vector where possible, but that's just an assumption.

argilo:
> Any way to test performance?

Perhaps receive UDP packets with `nc` and pipe them into `pv`?

argilo:
Or make a flowgraph that round-trips out to UDP and back, and time that when processing a fixed amount of data.

argilo:
It occurs to me that if a high-performance circular buffer is required, we have `gr::buffer`. Maybe it would be best to use that instead.

argilo:
OK, I've now rewritten this to use `gr::buffer` instead. That actually allowed some loops and `memcpy` to be eliminated, which should be a performance win.

argilo:
I did a test with the following flowgraph:

Screenshot from 2022-10-29 13-09-48

`in.dat` contains a gigabyte of data from `/dev/urandom`.

On the `main` branch, there are dropped packets (`udp_source :warning: [UDP source:1234] missed  packets: 35`), and CPU utilization is approximately 36% for the UDP sink block and 18% for the UDP source block.

With the changes in place, there are no dropped packets, and CPU utilization is approximately 17% for the UDP sink block and 10% for the UDP source block. :tada:

marcusmueller:
Um, 

> On the main branch, there are dropped packets

should probably be a considered a bug. Since this amounts to data loss, this PR might actually qualify for exceptions in linux distros' "applying patches to released versions" policies.

argilo:
> On the main branch, there are dropped packets

Well, there are still dropped packets after the changes, you just have to push the sample rate higher before it starts happening.

gr-network/lib/udp_source_impl.h
```diff
@@ -53,7 +52,8 @@ class NETWORK_API udp_source_impl : public udp_source
 
     // A queue is required because we have 2 different timing
     // domains: The network packets and the GR work()/scheduler
-    boost::circular_buffer<char>* d_localqueue;
+    gr::buffer_sptr d_localqueue_writer;
```
marcusmueller, line 55: these should be `unique_ptr`, as far as I can tell.

Sadly, `gr::make_buffer` returns a shared pointer (it should really probably just return an object by value, C++17 guarantees return value optimization; the original gr::make_buffer code predates std:: smart pointers in GR, however, I think.).

So, leave as is for now :/

marcusmueller (approved):
Nice!

One thing: I think we're leaking d_udpsocket's object. Might be a place for a std::unique_ptr

marcusmueller:
> > On the main branch, there are dropped packets
> 
> Well, there are still dropped packets after the changes, you just have to push the sample rate higher before it starts happening.

It's a bit sad that we don't have a unit test that checks the data after running e.g. 10 MB through it at reasonable rate.

gr-network/lib/udp_source_impl.cc
```diff
@@ -198,15 +192,15 @@ uint64_t udp_source_impl::get_header_seqnum()
 
     switch (d_header_type) {
     case HEADERTYPE_SEQNUM: {
-        retVal = ((header_seq_num*)d_local_buffer)->seqnum;
+        retVal = ((header_seq_num*)d_localqueue_reader->read_pointer())->seqnum;
     } break;
 
     case HEADERTYPE_SEQPLUSSIZE: {
-        retVal = ((header_seq_plus_size*)d_local_buffer)->seqnum;
+        retVal = ((header_seq_plus_size*)d_localqueue_reader->read_pointer())->seqnum;
```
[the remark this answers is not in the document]
argilo: I think you're right that there will be undefined behaviour due to unaligned reads.

That's on top of the undefined behaviour due to the structure being copied straight from memory into the packet, which will presumably create endianness issues.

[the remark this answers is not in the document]
argilo: Fixed in 0ce76ccaf9b14859ce29e17b33562370e5772465. Does that look right, @marcusmueller?

argilo:
Performance is still approximately the same after the switch to `memcpy` for reading packet headers. That makes sense, since the headers are small in comparison to the packet length.

gr-network/lib/udp_sink_impl.cc
```diff
@@ -224,15 +216,17 @@ int udp_sink_impl::work(int noutput_items,
     const char* in = (const char*)input_items[0];
 
     // Build a long local queue to pull from so we can break it up easier
-    for (int i = 0; i < num_bytes_to_transmit; i++) {
-        d_localqueue->push_back(in[i]);
-    }
+    if (d_localqueue_writer->space_available() < num_bytes_to_transmit)
+        d_localqueue_reader->update_read_pointer(num_bytes_to_transmit -
+                                                 d_localqueue_writer->space_available());
+    memcpy(d_localqueue_writer->write_pointer(), in, num_bytes_to_transmit);
```
willcode, line 222: Bounds check on `num_bytes_to_transmit` here? The original code didn't need one because it woudl just keep wrapping.

gr-network/lib/udp_source_impl.cc
```diff
@@ -267,14 +269,16 @@ int udp_source_impl::work(int noutput_items,
             // local queue in case we read more bytes than noutput_items is asking
             // for.  In that case we'll only return noutput_items bytes
             const char* read_data = asio::buffer_cast<const char*>(d_read_buffer.data());
-            for (int i = 0; i < bytes_read; i++) {
-                d_localqueue->push_back(read_data[i]);
-            }
+            if (d_localqueue_writer->space_available() < bytes_read)
+                d_localqueue_reader->update_read_pointer(
+                    bytes_read - d_localqueue_writer->space_available());
+            memcpy(d_localqueue_writer->write_pointer(), read_data, bytes_read);
```
willcode, line 275: Bounds check? `bytes_read` is not checked before use in `memcpy()`.

gr-network/lib/udp_source_impl.cc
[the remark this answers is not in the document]
marcusmueller: It does!

gr-network/lib/udp_sink_impl.cc
argilo: Good catch. I'll throw in bounds checks.

argilo:
Bounds check were added in https://github.com/gnuradio/gnuradio/pull/6312/commits/a749e477f0b9fcb2e5cf2f39b262bab6c473ee89. They should preserve the original behaviour of skipping the earliest bytes if there's more data than can fit in the queue.

willcode (approved):

argilo:
The test failure on Windows is unrelated: https://github.com/gnuradio/gnuradio/pull/6272#issuecomment-1296459000

PR #3165, opened 2020-02-02
QTGUI: Incorporate new GUI controls

ghostop14:
These updates expand the user interface capabilities of
GNU Radio.  This PR includes all of the controls more fully
documented here: https://github.com/ghostop14/gr-guiextra

PR #6316, opened 2022-10-30
qtgui: Fix code generation for Toggle Button & Toggle Switch

argilo:
## Description
The QT GUI Toggle Button and Toggle Switch blocks generate invalid Python code if the Type is set to Float or String:

```
  File "/home/argilo/git/bso2022/ultrasonic/tx.py", line 193, in <module>
    main()
  File "/home/argilo/git/bso2022/ultrasonic/tx.py", line 171, in main
    tb = top_block_cls()
  File "/home/argilo/git/bso2022/ultrasonic/tx.py", line 98, in __init__
    if real == bool:
NameError: name 'real' is not defined
```

```
  File "/home/argilo/git/bso2022/ultrasonic/tx.py", line 193, in <module>
    main()
  File "/home/argilo/git/bso2022/ultrasonic/tx.py", line 171, in main
    tb = top_block_cls()
  File "/home/argilo/git/bso2022/ultrasonic/tx.py", line 98, in __init__
    if string == bool:
NameError: name 'string' is not defined
```

In addition to being broken, the generated code is rather complex:
```python
        if string == bool:
        	self._button_choices = {'Pressed': bool('foo'), 'Released': bool('baz')}
        elif string == str:
        	self._button_choices = {'Pressed': "'foo'".replace("'",""), 'Released': "'baz'".replace("'","")}
        else:
        	self._button_choices = {'Pressed': 'foo', 'Released': 'baz'}
```
And it will have syntax errors if the choices happen to contain double quotes.

I've fixed this by doing the following:

1. Do the type checking in mako instead of Python.
2. Get rid of the special handling for strings, which doesn't seem to be needed.

* QT GUI Toggle Button
* QT GUI Toggle Switch

## Testing Done
I tested out both blocks with all four input types. When testing the string type, I also tried out special characters (e.g. `foo ' bar " baz`) and everything worked as expected.

## Checklist

### DIFF
--- a/gr-qtgui/grc/qtgui_togglebutton.block.yml
+++ b/gr-qtgui/grc/qtgui_togglebutton.block.yml
@@ -86,15 +86,14 @@ templates:
         <%
             win = '_%s_toggle_button'%id
         %>\
-        if ${type} == bool:
-        	self._${id}_choices = {'Pressed': bool(${pressed}), 'Released': bool(${released})}
-        elif ${type} == str:
-        	self._${id}_choices = {'Pressed': "${pressed}".replace("'",""), 'Released': "${released}".replace("'","")}
-        else:
-        	self._${id}_choices = {'Pressed': ${pressed}, 'Released': ${released}}
-        	
-        ${win} = qtgui.ToggleButton(${ 'self.set_' + context.get('id')() }, ${(label if (len(label) - 2 > 0) else repr(id))}, self._${id}_choices, ${initPressed},"${outputmsgname}".replace("'",""))
-        ${win}.setColors("${relBackgroundColor}","${relFontColor}","${pressBackgroundColor}","${pressFontColor}")
+        % if type == 'bool':
+        self._${id}_choices = {'Pressed': bool(${pressed}), 'Released': bool(${released})}
+        % else:
+        self._${id}_choices = {'Pressed': ${pressed}, 'Released': ${released}}
+        % endif
+
+        ${win} = qtgui.ToggleButton(${ 'self.set_' + context.get('id')() }, ${(label if (len(label) - 2 > 0) else repr(id))}, self._${id}_choices, ${initPressed}, ${outputmsgname})
+        ${win}.setColors("${relBackgroundColor}", "${relFontColor}", "${pressBackgroundColor}", "${pressFontColor}")
         self.${id} = ${win}
 
         ${gui_hint() % win}
--- a/gr-qtgui/grc/qtgui_toggleswitch.block.yml
+++ b/gr-qtgui/grc/qtgui_toggleswitch.block.yml
@@ -91,16 +91,15 @@ templates:
     var_make: self.${id} = ${id} = ${value}
     make: |-
         <%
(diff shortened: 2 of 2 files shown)


argilo:
There were a couple more instances of `.replace("'","")` that I missed; I've now replaced those as well. I also added in some missing spaces between generated argument lists.

willcode (approved):

argilo:
And one more bug: both of these blocks overwrote their value variable with a reference to their GUI window: `self.${id} = ${win}`. This causes weird and wild failures if another variable used in the same expression is changed:
```
  File "/home/argilo/prefix_311/lib/python3.10/site-packages/gnuradio/qtgui/range.py", line 204, in changed
    self.notifyChanged(self.rangeType(val))
  File "/home/argilo/prefix_311/lib/python3.10/site-packages/gnuradio/qtgui/range.py", line 308, in sliderChanged
    self.notifyChanged(self.rangeType(value))
  File "/home/argilo/prefix_311/lib/python3.10/site-packages/gnuradio/qtgui/range.py", line 140, in notifyChanged
    self.__slot(value)
  File "/home/argilo/git/bso2022/ultrasonic/tx.py", line 178, in set_level
    self.blocks_multiply_const_vxx_0.set_k((self.left_on * self.level))
TypeError: unsupported operand type(s) for *: 'ToggleButton' and 'float'
```
I've removed the extra `self.${id} = ${win}` lines to prevent the overwrite from happening.

argilo:
I see that the `self.${id} = ${win}` thing is in many of the GUI blocks added in #3165, so they may have similar bugs. I'll leave that to another PR though.

argilo:
There were also a couple typos in the Toggle Switch, which I've added fixes for:

* The window name variable had a misleading suffix of `_toggle_button`, which I've changed to `_toggle_switch`.
* The documentation referred to the widget as a button in one place. I've changed that to "switch."

argilo:
After some further testing I noticed that deleting `self.${id} = ${win}` (which I did to fix a bug which prevented the value from being used in expressions) has caused another bug: it's no longer possible to connect the message output to another block. (That now fails with `AttributeError: 'int' object has no attribute 'to_basic_block'`.) I'll have to explore what the proper solution is, but in the meantime I'll label this PR as "Don't Merge."

argilo:
Hmm. It seems like there's an inherent conflict between blocks that set variable values, and blocks that have ports. In the first case, the variable value goes in `self.id`, and in the second case the block itself goes in `self.id`.

It looks like until recently, no block did both. But then a message port was added to the QT GUI Range widget in #5566. That broke the variable output, which someone else noticed and fixed in #6144, which broke the message port (which is still broken as of today).

willcode:
Some similarity to https://github.com/gnuradio/gnuradio/issues/6320

argilo:
I put back the `self.${id} = ${win}` lines. This restores the message ports.

I think it's best to leave the issue of variable setting to another PR, since it affects a wide range of blocks and doesn't have an obvious solution.

I've tested both blocks, and the blocks appear to work correctly now. I'll remove the "Don't Merge" label.

PR #6317, opened 2022-10-30
qtgui: Fix code generation for Msg CheckBox & Digital Number Control

argilo:
## Description
After finishing #6316 I noticed a couple other blocks with broken code generation.

QT GUI Msg CheckBox generates invalid syntax:
```
  File "/home/argilo/git/bso2022/ultrasonic/tx.py", line 89
    _left_on_toggle_button = qtgui.MsgCheckBox(self.set_left_on, ''foo'', self._left_on_choices, False, 1, 1,"'value'".replace("'",""))
                                                                 ^^^^^^^^^
SyntaxError: invalid syntax. Perhaps you forgot a comma?
```
Also, when the Boolean type is selected, the value is not coerced to boolean (as it is in the Toggle Button & Toggle Switch blocks).

Finally, the QT GUI Digital Number Control uses `.replace("'","")` to munge a string; this is no longer needed, and by removing it the generated code is correct even if the name happens to contain a double quote.

* QT GUI Msg Checkbox
* QT GUI Digital Number Control

## Testing Done
Made a simple flow graph and tested both blocks with various parameters.

## Checklist

### DIFF
--- a/gr-qtgui/grc/qtgui_msgcheckbox.block.yml
+++ b/gr-qtgui/grc/qtgui_msgcheckbox.block.yml
@@ -68,11 +68,15 @@ templates:
     var_make: self.${id} = ${id} = ${value}
     make: |-
         <%
-            win = '_%s_toggle_button'%id
+            win = '_%s_msgcheckbox_win'%id
         %>\
+        % if type == 'bool':
+        self._${id}_choices = {'Pressed': bool(${pressed}), 'Released': bool(${released})}
+        % else:
         self._${id}_choices = {'Pressed': ${pressed}, 'Released': ${released}}
-        	
-        ${win} = qtgui.MsgCheckBox(${ 'self.set_' + context.get('id')() }, '${no_quotes(label,repr(id))}', self._${id}_choices, ${initPressed}, ${cellalignment}, ${verticalalignment},"${outputmsgname}".replace("'",""))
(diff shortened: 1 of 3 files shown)


argilo:
Hang on, there's more bugs in these blocks still. I'll add fixes to this PR.

argilo:
OK, more bugs fixed:

* `self.${id} = ${win}` was causing the value variable to be overwritten with a reference to the GUI widget. I've removed these extraneous lines.
* A typo in `msgcheckbox.py` caused a crash if the Initial State parameter was set to Checked.

willcode (approved):

mormj (approved):

argilo:
One more quick fix: The Msg CheckBox used a misleading `_toggle_button` suffix for the variable containing its GUI widget. I've changed that to `_msgcheckbox_win`.

argilo:
Just like https://github.com/gnuradio/gnuradio/pull/6316#issuecomment-1298019297, the removal of `self.${id} = ${win}` has broken the ability to connect the message output port. I'll mark this as "Don't Merge" while I sort that out.

argilo:
As with #6316, I've restored the `self.${id} = ${win}` lines so that it's possible to use the output ports of these blocks. I re-tested the blocks, and all appears to be working well.

PR #6319, opened 2022-10-31
blocks: Increase number of samples and tolerance in qa_throttle

argilo:
## Description
Even after the tolerance increase in #6272, `qa_throttle` is still failing on Windows:

https://github.com/gnuradio/gnuradio/actions/runs/3357519882/jobs/5563380391
https://github.com/gnuradio/gnuradio/actions/runs/3360756897/jobs/5570335077

Let's bump it up again to stop the test failures.

## Related Issue
* https://github.com/gnuradio/gnuradio/pull/6272#issuecomment-1296459000

Tests for the Throttle block.

## Checklist

### DIFF
--- a/gr-blocks/python/blocks/qa_throttle.py
+++ b/gr-blocks/python/blocks/qa_throttle.py
@@ -23,9 +23,9 @@ class test_throttle(gr_unittest.TestCase):
         self.tb = None
 
     def test_throttling(self):
-        src_data = [1, 2, 3]
+        src_data = list(range(200))
         src = blocks.vector_source_c(src_data)
-        thr = blocks.throttle(gr.sizeof_gr_complex, 10)
+        thr = blocks.throttle(gr.sizeof_gr_complex, 100)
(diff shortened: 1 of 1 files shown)


willcode:
With such a small number of samples, this test is probably not measuring anything useful.

argilo:
The measured `total_time` values on my laptop are always between 301 and 302 ms, so the tests are at least able to measure something useful in an environment where there's no load.

I agree though that it might make sense to have a higher number of samples and a longer-running test, so that the effects of system load are smaller relative to the duration of the test.

argilo:
OK, I've increased the first test to 200 samples and 100sps, and allowed for a run time of 2 to 3 seconds.

The second test uses 400 samples and 200sps, and also allows a run time of 2 to 3 seconds. (A failure should definitely occur if the tag is not processed, since the test will take 4+ seconds in that case.)

willcode (approved):

PR #6321, opened 2022-11-01
pdu: Increase tolerance in flaky test again

argilo:
## Description
The tolerance for this test was increased in #5764, but that was not enough. Another test failure occurred on Windows:

https://github.com/gnuradio/gnuradio/actions/runs/3364269344/jobs/5578423969

## Related Issue
* #5764

CI tests for the Time Delta block.

## Testing Done
I verified that the test still passes on my machine, even when stressed with `stress -c 128`.

## Checklist

### DIFF
--- a/gr-pdu/python/pdu/qa_time_delta.py
+++ b/gr-pdu/python/pdu/qa_time_delta.py
@@ -89,7 +89,7 @@ class qa_time_delta(gr_unittest.TestCase):
         delta_tag = pmt.dict_ref(a_meta, pmt.intern(
             "sys time delta (ms)"), pmt.PMT_NIL)
         self.assertAlmostEqual(tnow - 10.0, pmt.to_double(time_tag), delta=1e-6)
-        self.assertAlmostEqual(10000, pmt.to_double(delta_tag), delta=250)
+        self.assertAlmostEqual(10000, pmt.to_double(delta_tag), delta=500)
 
 
 if __name__ == '__main__':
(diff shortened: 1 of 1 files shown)


willcode:
How can there be such a large epsilon on that? It's a constant.

argilo:
It's not a constant. The system system time is measured once on line 61, and that time minus 10 seconds is added as a tag. The flow graph then executes, and the Time Delta block measures the difference between the time indicated in that tag, and the current system time. It will measure 10 seconds plus the execution time of the test (well, lines 61-79).

willcode:
I might be missing an operation somewhere, but here are the lines I'm looking at:
```
        e_meta = pmt.dict_add(e_meta, pmt.intern(
            'sys time delta (ms)'), pmt.from_double(10000.0))
```
```
        delta_tag = pmt.dict_ref(a_meta, pmt.intern(
            "sys time delta (ms)"), pmt.PMT_NIL)
```
```
        self.assertAlmostEqual(10000, pmt.to_double(delta_tag), delta=500)
```

argilo:
Note `e_meta` vs `a_meta`.

argilo:
It seems the `e_meta` stuff isn't even used...

willcode:
I still can't tell what line 92 is trying to test or what the extra level of indirection is for. Eyes buggy - will look at it again later.

argilo:
The Time Delta block reads the `system_time` property of an incoming PDU and adds a `sys time delta (ms)` property indicating the difference between the current system time, and the earlier time recorded in the `system_time` property.

Line 91 checks that the `system_time` property is still there and set to the original value, and line 92 checks that `sys time delta (ms)` is approximately 10000 ms. (It will actually be a bit more, because some time will have passed before the difference was measured by the Time Delta block.

willcode (approved):

argilo:
I also threw in a second commit to remove those unused variables. Let me know if you're OK with that.

Issue #5948, opened 2022-06-08
qa_constellation_receiver is flaky


argilo:
The qa_constellation_receiver test recently failed on a maint-3.9 backport (#5945):

https://github.com/gnuradio/gnuradio/runs/6772627220?check_suite_focus=true

The failure occurred on Fedora 36, and the details were as follows:
```
======================================================================
FAIL: test_basic (__main__.test_constellation_receiver)
Tests a bunch of different constellations by using generic
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/__w/gnuradio/gnuradio/gr-digital/python/digital/qa_constellation_receiver.py", line 133, in test_basic
    self.assertTrue(correct > req_correct)
AssertionError: False is not true
----------------------------------------------------------------------
Ran 2 tests in 0.877s
FAILED (failures=1)
Constellation is <gnuradio.digital.digital_python.constellation_expl_rect object at 0x7f37440399b0>. Differential is True.  Required correct is 0.8. Correct is 0.6714975845410628. FAIL.
```

So far I have not been able to reproduce the failure locally.

marcusmueller:
It seems to be deterministically seeded:
https://github.com/gnuradio/gnuradio/blob/main/gr-digital/python/digital/qa_constellation_receiver.py#L87-L92

argilo:
I suspect that the flaky test was due to the Polyphase Clock Sync bug fixed in #7040, so I'll close this off. If another test failure occurs, I'll re-open this.

PR #6335, opened 2022-11-03
QA: Make assertions more specific

argilo:
## Description
While looking into #5948, I noticed that the error message (`AssertionError: False is not true`) is unhelpful; `assertTrue` hides the details about the numbers that were compared.

Here I've changed that assertion to use `assertGreater` and also fixed many other instances where `assertTrue` or `assertFalse` were used where other alternatives would give more useful diagnostic information.

QA tests for various modules.

## Testing Done
I verified that all the tests pass on my machine. (We'll see how things go in CI as well.) I also intentionally caused a few of the assertions to fail, to verify that the error messages are more helpful.

## Checklist

### DIFF
--- a/gr-blocks/python/blocks/qa_file_metadata.py
+++ b/gr-blocks/python/blocks/qa_file_metadata.py
@@ -57,14 +57,9 @@ class test_file_metadata(gr_unittest.TestCase):
 
         handle = open(outfile, "rb")
         header_str = handle.read(blocks.parse_file_metadata.HEADER_LENGTH)
-        if(len(header_str) == 0):
-            self.assertFalse()
+        self.assertGreater(len(header_str), 0)
 
-        try:
-            header = pmt.deserialize_str(header_str)
-        except RuntimeError:
-            self.assertFalse()
-
+        header = pmt.deserialize_str(header_str)
         info = blocks.parse_header(header, False)
 
         extra_str = handle.read(info["extra_len"])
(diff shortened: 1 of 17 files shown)


gr-blocks/python/blocks/qa_file_metadata.py
```diff
@@ -57,26 +57,17 @@ def test_001(self):
...
-        if(len(header_str) == 0):
-            self.assertFalse()
-
-        try:
-            header = pmt.deserialize_str(header_str)
-        except RuntimeError:
-            self.assertFalse()
+        self.assertGreater(len(header_str), 0)
 
+        header = pmt.deserialize_str(header_str)
```
argilo, line 62: No `try` / `except` is required here. If an exception is raised, the test will exit immediately and report that.

argilo: Same goes for three more instances below.

gr-digital/python/digital/qa_pfb_clock_sync.py
```diff
@@ -165,8 +165,6 @@ def test03(self):
         self.tb.stop()
         self.tb.wait()
 
-        self.assertTrue(True)
```
argilo, line 168: As far as I can tell, this is useless, so I removed it. (Same another instance below.)

argilo:
I notice that some of the assertions here are also preceded by:
```python
if not <condition>:
    print("<details of error>")
```
Those should be replaced by `msg=` arguments in the assertions themselves, since standard output is suppressed during test execution.

I'll make those changes here as well.

argilo:
The few instances of `print()` have been replaced with `msg=`.

mbr0wn (approved):
Excellent changeset!

mormj (approved):

PR #6336, opened 2022-11-04
4.0: Make base scheduler functionality common

mormj:
## Description
This moves `thread_wrapper` and `graph_executor` out of `scheduler_nbt` and under `gr/` so that it can be used by other scheduler implementations.  This cleans up some of the methods that were being passed in unused objects such as buffer manager

Under this organization, thread_wrapper could be used by another scheduler - more to figure out as to what methods from this should become virtual and be overridden

## Related Issue
None - based on feedback at GRCon

NBT Scheduler

## Testing Done
QA passes


mormj:
Still need to better figure out the option parsing

PR #6339, opened 2022-11-04
4.0: History and Forecast lookalikes

mormj:
## Description
Since 4.0 doesn't have `history()` or `forecast()`, this PR introduces a few convenience methods to simplify block development and the contract with the scheduler

### protected block methods
- noconsume(), declare_noconsume(): allow blocks to keep track of how many samples they don't plan on consuming
- enforce_constraints(): gets called prior to work (scheduler calls `do_work`) and allows adjustment of the input/output nitems to meet the block requirements (whatever they may be)

### contract with scheduler
Two methods added so the scheduler can know the minimum required output buffer
- `min_required_output_buffer()`: by default just `output_multiple`
- `min_required_upstream_buffer()`: the requirement from a reader based on its output to what the upstream block needs to provide

This also adds in changes to `fir_filter` and `pfb_arb_resampler` to reflect the updates

## Checklist

### DIFF
--- a/blocklib/filter/fir_filter/fir_filter.yml
+++ b/blocklib/filter/fir_filter/fir_filter.yml
@@ -44,6 +44,8 @@ parameters:
     label: Decimation
     dtype: size
     settable: false
+    grc:
+      default: 1
 -   id: taps
     label: Taps
     dtype: TAP_T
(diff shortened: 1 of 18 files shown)


mormj:
This might be more dramatically altered by the new buffer API that is being proposed, but I'm going to go ahead and merge this since it is minor changes and cleans up the block interface a bit

PR #6340, opened 2022-11-04
Remove all embedding/linking of Python outside of bindings

### LINKED ISSUE #6283: Segfault on macOS when running a flowgraph containing a Python block
Many CI tests segfault on macOS, and thus need to be excluded from running in Conda builds. I looked into a few of the failures, and it appears that a segfault occurs whenever a flowgraph containing a Python block is started. The following script demonstrates a crash:
```python
import numpy as np
from gnuradio import gr, blocks

class foo_cc(gr.sync_block):
    def __init__(self):
        gr.sync_block.__init__(
            self,
            name='foo',
            in_sig=[np.complex64],
            out_sig=[np.complex64]
        )

    def work(self, input_items, output_items):
        output_items[0][:] = input_items[0]
        return len(output_items[0])


tb = gr.top_block()

src = blocks.vector_source_c([1, 2, 3])
foo = foo_cc()
dst = blocks.vector_sink_c()

tb.connect(src, foo)
tb.connect(foo, dst)
tb.run()
```
Backtrace:
```
* thread #15, stop reason = EXC_BAD_ACCESS (code=1, address=0x1b0)
  * frame #0: 0x00007ff801b20882 libsystem_pthread.dylib`pthread_mutex_lock + 4
    frame #1: 0x0000000106a57670 Python`take_gil + 76
    frame #2: 0x0000000106ab0270 Python`PyGILState_Ensure + 48
    frame #3: 0x0000000105bab61c libgnuradio-runtime.v3.11.0.0git-257-g83ba4690.dylib`pybind11::detail::get_internals() + 60
    frame #4: 0x0000000105bab55c libgnuradio-runtime.v3.11.0.0git-257-g83ba4690.dylib`pybind11::gil_scoped_acquire::gil_scoped_acquire() + 28
    frame #5: 0x0000000105baabc3 libgnuradio-runtime.v3.11.0.0git-257-g83ba4690.dylib`gr::block_gateway_impl::start() + 35
    frame #6: 0x0000000105b4eb62 libgnuradio-runtime.v3.11.0.0git-257-g83ba4690.dylib`gr::block_executor::block_executor(std::__1::shared_ptr<gr::block>, int) + 322
    frame #7: 0x0000000105ba0847 libgnuradio-runtime.v3.11.0.0git-257-g83ba4690.dylib`gr::tpb_thread_body::tpb_thread_body(std::__1::shared_ptr<gr::block>, std::__1::shared_ptr<boost::barrier>, int) + 87
    frame #8: 0x0000000105b95c68 libgnuradio-runtime.v3.11.0.0git-257-g83ba4690.dylib`gr::tpb_container::operator()() + 88
    frame #9: 0x0000000105b95af3 libgnuradio-runtime.v3.11.0.0git-257-g83ba4690.dylib`gr::thread::thread_body_wrapper<gr::tpb_container>::operator()() + 35
    frame #10: 0x000000010518eba2 libboost_thread-mt.dylib`boost::(anonymous namespace)::thread_proxy(void*) + 146
    frame #11: 0x00007ff801b254e1 libsystem_pthread.dylib`_pthread_start + 125
    frame #12: 0x00007ff801b20f6b libsystem_pthread.dylib`thread_start + 15
```

The crash occurs at this line:

https://github.com/gnuradio/gnuradio/blob/962eb04aaef345bcdc808d98cd4621189237539a/gnuradio-runtime/lib/block_gateway_impl.cc#L59

Unfortunately I don't know much about pybind11, so I'm not sure what's going on here.

ryanvolz:
## Description

The motivation is to fix segfaults on macOS + conda, which happen because

1. gnuradio-runtime links with Python so that libpython is loaded dynamically at runtime
2. Conda's Python interpreter links libpython statically

so that with these together we get multiple conflicting versions of Python symbols when running Python blocks. See the investigation in #6283.

Not linking gnuradio-runtime with Python turned out to be as easy as moving `block_gateway` from `gnuradio-runtime` to the `gr_python` bindings, which to my knowledge is the only place it has been used. In this way the Python symbols can come along with the Python interpreter using the bindings, and there is no conflict. Enabling the once-failing Python block tests on macOS shows that this works as desired. I separated `block_gateway` from `gnuradio-runtime` in the way that I thought made the most sense, but I'm happy to accept feedback on other ways of doing it.

Gr-qtgui also had remnants of linking with Python, but this has not actually been necessary since #4994. So I've removed the last traces of Python from gr-qtgui here.

## Related Issue

Fixes #6283

gnuradio-runtime, gr-qtgui

## Testing Done

All macOS tests are enabled and pass in CI!

## Checklist


gnuradio-runtime/python/gnuradio/gr/bindings/block_gateway_python.cc
```diff
@@ -7,17 +7,200 @@
...
+ * The gateway provides access to all the gr::block routines.
+ */
+typedef enum {
+    GW_BLOCK_GENERAL = 0,
+    GW_BLOCK_SYNC,
+    GW_BLOCK_DECIM,
+    GW_BLOCK_INTERP
+} gw_block_t;
+
+class block_gateway : virtual public gr::block
```
jwmelto, line 34: This is now an implementation detail; no need for virtual inheritance
```suggestion
class block_gateway : public gr::block
```
(unless there's something in the pybind implementation that still needs it?)

gnuradio-runtime/python/gnuradio/gr/bindings/block_gateway_python.cc
```diff
@@ -7,17 +7,200 @@
...
+    py::handle _py_handle;
+
+public:
+    block_gateway(const py::handle& py_handle,
+                  const std::string& name,
+                  gr::io_signature::sptr in_sig,
+                  gr::io_signature::sptr out_sig)
+        : block(name, in_sig, out_sig)
+    {
+        _py_handle = py_handle;
```
jwmelto, line 46: any reason this can't be in the initializer list?

gnuradio-runtime/python/gnuradio/gr/bindings/block_gateway_python.cc
```diff
@@ -7,17 +7,200 @@
...
+{
+private:
+    py::handle _py_handle;
+
+public:
+    block_gateway(const py::handle& py_handle,
+                  const std::string& name,
+                  gr::io_signature::sptr in_sig,
+                  gr::io_signature::sptr out_sig)
+        : block(name, in_sig, out_sig)
```
jwmelto, line 44: it must be an arcane facet of scoping rules that probably isn't worth exploiting, that you are in the `gr` namespace here. Go ahead and be explicit
```suggestion
        : gr::block(name, in_sig, out_sig)
```
(note that historically, the impl classes were also in the `gr` namespace; in moving this, it is no longer the case)

gnuradio-runtime/python/gnuradio/gr/bindings/block_gateway_python.cc
```diff
@@ -7,17 +7,200 @@
...
+
+    /*!
+     * Make a new gateway block.
+     * \param py_handle the pybind11 object with callback
+     * \param name the name of the block (Ex: "Shirley")
+     * \param in_sig the input signature for this block
+     * \param out_sig the output signature for this block
+     * \return a new gateway block
+     */
+    static sptr make(const py::object& py_handle,
```
jwmelto, line 60: as a general rule, I'd recommend against in-line method definitions. I'd rather see the class definition as interface, and the method definitions out of line

gnuradio-runtime/python/gnuradio/gr/bindings/block_gateway_python.cc
```diff
@@ -7,17 +7,200 @@
...
+            throw std::runtime_error(
+                "attempt to set_msg_handler_pybind() on invalid input message port!");
+        }
+        d_msg_handlers_pybind[which_port] = handler_name;
+    }
+
+protected:
+    // Message handlers back into python using pybind API
+    typedef std::map<pmt::pmt_t, std::string, pmt::comparator> msg_handlers_pybind_t;
+    msg_handlers_pybind_t d_msg_handlers_pybind;
```
jwmelto, line 175: why is this protected? Can this class be sub-classed?

FWIW, modern C++ prefers `using`:
```suggestion
    // Message handlers back into python using pybind API
    using msg_handlers_pybind_t = std::map<pmt::pmt_t, std::string, pmt::comparator>;
    msg_handlers_pybind_t d_msg_handlers_pybind;
```

gnuradio-runtime/python/gnuradio/gr/bindings/block_gateway_python.cc
```diff
@@ -7,17 +7,200 @@
...
+        // Is there a handler?
+        if (d_msg_handlers_pybind.find(which_port) != d_msg_handlers_pybind.end()) {
+            // d_msg_handlers_pybind[which_port]->calleval(msg); // Yes, invoke it.
+            py::gil_scoped_acquire acquire;
+            // std::string handler_name(d_msg_handlers_pybind[which_port]);
+            py::object ret =
+                _py_handle.attr(d_msg_handlers_pybind[which_port].c_str())(msg);
+        } else {
+            // Pass to generic dispatcher if not found
+            gr::basic_block::dispatch_msg(which_port, msg);
```
jwmelto, line 197: just curious if this method is defined in `gr::block` and you want to skip that version, and go to the grandparent implementation, or if there's any other reason to scope it so explicitly?

(yes, I realize this code was fork-lifted, but it's an opportunity for re-examination).

My preferred pattern (which you may disagree with) is
```c++
class Foo : public Bar
{
    using Base = Bar;
    ...
    void baz() override
    {
        Base::baz();
    }
};
```

jwmelto (commented):
Removing all these python tentacles is brilliant.

argilo:
I'm not enough of a C++, pybind11, or linking expert to comment on much of this, but I can say that GNU Radio still builds and runs fine on my laptop with these changes, and Gqrx runs as well.

It's nice to see the linkage to libpython gone! The `ldd` tool shows that none of the GNU Radio libraries are linked to libpython anymore, except for libgnuradio-uhd, which links to it indirectly through libuhd. (That's of course not GNU Radio's fault.)

argilo:
I also tried creating a simple flow graph with an Embedded Python Block in GRC, and that worked fine.

gnuradio-runtime/python/gnuradio/gr/bindings/block_gateway_python.cc
```diff
@@ -7,17 +7,163 @@
  *
  */
 
+#include "block_gateway.h"
+
+#include <gnuradio/block.h>
```
jwmelto, line 12: unneeded

gnuradio-runtime/python/gnuradio/gr/bindings/block_gateway_python.cc
```diff
@@ -7,17 +7,163 @@
...
+#include <gnuradio/io_signature.h>
```
jwmelto, line 13: redundant with header include

gnuradio-runtime/python/gnuradio/gr/bindings/block_gateway_python.cc
```diff
@@ -7,17 +7,163 @@
...
+#include <gnuradio/types.h>
```
jwmelto, line 14: redundant with header include

gnuradio-runtime/python/gnuradio/gr/bindings/block_gateway_python.cc
```diff
@@ -7,17 +7,163 @@
...
+#include <pmt/pmt.h>
```
jwmelto, line 15: redundant with header include

gnuradio-runtime/python/gnuradio/gr/bindings/block_gateway_python.cc
```diff
@@ -7,17 +7,163 @@
...
+
 #include <pybind11/complex.h>
 #include <pybind11/pybind11.h>
 #include <pybind11/stl.h>
 
+#include <map>
```
jwmelto, line 21: redundant with header include

gnuradio-runtime/python/gnuradio/gr/bindings/block_gateway_python.cc
```diff
@@ -7,17 +7,163 @@
...
+#include <string>
+
 namespace py = pybind11;
 
-#include <gnuradio/block_gateway.h>
+block_gateway::block_gateway(const py::handle& py_handle,
+                             const std::string& name,
+                             gr::io_signature::sptr in_sig,
+                             gr::io_signature::sptr out_sig)
+    : _py_handle(py_handle), gr::block(name, in_sig, out_sig)
```
jwmelto, line 31: order of initializers .... base class first, then members

gnuradio-runtime/python/gnuradio/gr/bindings/block_gateway.h
```diff
@@ -0,0 +1,115 @@
...
+     * \param out_sig the output signature for this block
+     * \return a new gateway block
+     */
+    static sptr make(const py::object& py_handle,
+                     const std::string& name,
+                     gr::io_signature::sptr in_sig,
+                     gr::io_signature::sptr out_sig);
+
+    // Protected members of gr::block trampolined here for python blocks to use
+    void _add_item_tag(unsigned int which_output, const gr::tag_t& tag);
```
jwmelto, line 63: so the question is, why do we need these delegator methods that call base class methods?

jwmelto (commented):
LGTM

willcode:
I'm assuming (in labeling this for backport) that nothing outside GR would try to use the python gateway block. This is technically an API change, if something would use that from C++. Anyone know of a potential direct use/subclass?

argilo:
In a quick search of GitHub, the only project I could find including `block_gateway.h` was airprobe, and that include appears to be unnecessary:

https://github.com/pcabreracamara/airprobe/blob/a22173958f41a01834b5fc477fc272a7d51bb7fb/gsm-receiver/src/lib/gsm_receiver_cf.h#L27

gnuradio-runtime/python/gnuradio/gr/bindings/block_gateway.h
ryanvolz: The comment explaining it makes sense to me, or am I being dense?

ryanvolz:
> In a quick search of GitHub, the only project I could find including `block_gateway.h` was airprobe, and that include appears to be unnecessary:

And of course that would need much more updating to be compatible with anything after GR 3.7 anyway. It seems like it's a pretty safe API change that's not really an API change in practice. I didn't have any issues ABI-wise in running `gqrx` and `gr-satellites` built against 3.10.4.0 with a recompiled GNU Radio that includes these commits, not that I would have expected problems.

gnuradio-runtime/python/gnuradio/gr/bindings/block_gateway.h
jwmelto: It isn't obvious (anymore) _why_ we need the delegators; Presumably they were to allow a code generator to know about base class methods to generate wrappers. But given this file's location in the Python binding directory, it wasn't obvious to me (and I'm pretty ignorant of pybind11; only slightly less ignorant of SWIG) that a wrapper was being generated over this.

(since Python doesn't support overloading, and some of these are overloaded calls, that's not an obvious purpose)

The other possibility is that they exist to change the scoping of the methods; providing public access to protected methods. If that's what the comment is saying, then you are correct that the comment is correct, if less than clear. "[Trampolined](https://en.wikipedia.org/wiki/Trampoline_(computing))" is not a term I come across regularly. But I also do mostly pure language development, rather than crossing APIs. So maybe it's a very clear comment. As a naive reader of this code, I wanted more.

My point in asking was to revisit whether the construct was still required, in light of the code reorganization. If so, great.

gnuradio-runtime/python/gnuradio/gr/bindings/block_gateway.h
```diff
@@ -0,0 +1,115 @@
...
+                     gr_vector_void_star& output_items) override;
+
+    bool start() override;
+
+    bool stop() override;
+
+    void set_msg_handler_pybind(const pmt::pmt_t& which_port, std::string& handler_name);
+
+private:
+    py::handle _py_handle;
```
jwmelto, line 104: According to [GREP1.md](https://github.com/gnuradio/greps/blob/main/grep-0001-coding-guidelines.md),
> All class data members shall begin with d_foo

I don't know how widely the standards are followed (my personal style is the same as yours), but I thought I'd mention it, for completeness.

gnuradio-runtime/python/gnuradio/gr/bindings/block_gateway.h
ryanvolz: > The other possibility is that they exist to change the scoping of the methods; providing public access to protected methods. If that's what the comment is saying, then you are correct that the comment is correct, if less than clear.

That's what it is, since we're only able to bind public methods. I'll tweak the comment text for clarity when I get a chance.

gnuradio-runtime/python/gnuradio/gr/bindings/block_gateway.h
ryanvolz: I don't have a preference and it's only this way because that's what it was in the former `_impl` class, so I'll probably change it.

ryanvolz:
Many thanks for the reviews @jwmelto!

mormj (approved):
Thank you for all this cleanup of the pybind!

PR #6342, opened 2022-11-06
qtgui: Revert changes to Range widget

argilo:
## Description
In #5566, a message output port was added to the QT GUI Range widget. This caused two new bugs:

First, if the Range variable was used in the same expression as another variable, and that other variable was changed before the Range slider was first moved, then an exception would occur:
```
Traceback (most recent call last):
  File "/home/argilo/prefix_311/lib/python3.10/site-packages/gnuradio/qtgui/range.py", line 204, in changed
    self.notifyChanged(self.rangeType(val))
  File "/home/argilo/prefix_311/lib/python3.10/site-packages/gnuradio/qtgui/range.py", line 308, in sliderChanged
    self.notifyChanged(self.rangeType(value))
  File "/home/argilo/prefix_311/lib/python3.10/site-packages/gnuradio/qtgui/range.py", line 140, in notifyChanged
    self.__slot(value)
  File "/home/argilo/epy_test.py", line 193, in set_ampl
    self.epy_block_0.example_param = self.ampl * self.ampl2
TypeError: unsupported operand type(s) for *: 'float' and 'GrRangeWidget'
```

Second, the uhd_fft and uhd_siggen_gui programs would crash whenever a Range slider was adjusted: #6320. Also, any Python code generated by GRC prior to #5566 has the same issue. These problems are due to changes that were made to the `RangeWidget` class.

The first of these bugs was fixed in #6144, however that change introduced a new bug that made the message output port inoperable. If the port is connected to another block, then the following crash occurs:
```
Traceback (most recent call last):
  File "/home/argilo/prefix_311/lib/python3.10/site-packages/gnuradio/qtgui/range.py", line 204, in changed
    self.notifyChanged(self.rangeType(val))
  File "/home/argilo/prefix_311/lib/python3.10/site-packages/gnuradio/qtgui/range.py", line 308, in sliderChanged
    self.notifyChanged(self.rangeType(value))
  File "/home/argilo/prefix_311/lib/python3.10/site-packages/gnuradio/qtgui/range.py", line 139, in notifyChanged
    self.message_port_pub(self.outputportname_pmt, pmt.cons(self.outputmsgname_pmt, self.value_to_pmt(value)))
AttributeError: 'RangeWidget' object has no attribute 'message_port_pub'
```

[...]

PR #6343, opened 2022-11-06
grc: Continue processing connections after an invalid one occurs

argilo:
## Description
While working on #6342, I noticed the following bug:

When opening a GRC file, if any invalid connection is present, then all subsequent valid connections will fail to be loaded. This occurs because the loop that processes connections exits on the first error.

For example, the following flowgraph was generated in GNU Radio 3.10.4.0, which has a message output in the QT GUI Range block. When attempting to load it in another version of GNU Radio which lacks this port, all of the connections vanish:

Screenshot from 2022-11-06 08-23-36

It appears that this bug was introduced in GNU Radio 3.8, in the following commit: dbf58fa349002d94c26883b130ed8864ead79ea6

Moving the `try` back inside the loop allows the valid connections to be loaded again:

Screenshot from 2022-11-06 08-31-50

It would seem that the motivation behind dbf58fa349002d94c26883b130ed8864ead79ea6 was to swallow up exceptions generated by `data.get('connections', [])`, but I haven't yet found a scenario where an exception would occur there.

Loading flow graphs in GRC.

## Testing Done
I generated a flow graph that includes a QT GUI Range block with its new message output port (added in #5566) connected, then loaded it it a branch where #5566 has been reverted.

I also tried loading a flowgraph with zero connection to see if that would cause problems, but it did not.

## Checklist

### DIFF
--- a/grc/core/FlowGraph.py
+++ b/grc/core/FlowGraph.py
@@ -466,9 +466,9 @@ class FlowGraph(Element):
         had_connect_errors = False
         _blocks = {block.name: block for block in self.blocks}
 
-        try:
-            # TODO: Add better error handling if no connections exist in the flowgraph file.
-            for src_blk_id, src_port_id, snk_blk_id, snk_port_id in data.get('connections', []):
+        # TODO: Add better error handling if no connections exist in the flowgraph file.
+        for src_blk_id, src_port_id, snk_blk_id, snk_port_id in data.get('connections', []):
+            try:
                 source_block = _blocks[src_blk_id]
                 sink_block = _blocks[snk_blk_id]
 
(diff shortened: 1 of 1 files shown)


argilo:
> It would seem that the motivation behind https://github.com/gnuradio/gnuradio/commit/dbf58fa349002d94c26883b130ed8864ead79ea6 was to swallow up exceptions generated by `data.get('connections', [])`, but I haven't yet found a scenario where an exception would occur there.

I can induce a failure by replacing the `connections:` section with something completely invalid like `connections: null` or `connections: "foo"`. Both of those cases produce a `TypeError` though, which wouldn't have been caught by moving the `try` outside the loop.

mormj:
I've run into this one quite a bit using GRC with 4.0 since I'm causing a lot of exceptions with some new constructs.  So this looks like a very helpful fix.

sdh11 (approved):
LGTM. I can't remember if there was a specific reason we originally moved that try out of the loop.

Issue #6217, opened 2022-10-06
4.0: Flowgraph Reconfiguration


mormj:
### Feature Description

For reasons of simplicity, 4.0 has not yet implemented live reconfiguration of flowgraph (lock, unlock, disconnect).  With the modular scheduler, this will have to be thought through carefully as to who owns what and how buffer states are preserved.

### Feature Urgency

medium (would be nice to have in the near future)

Issue #1768, opened 2018-05-08
Circular Buffers - Size Optimization


NotTheMattThatYouKnow:
GR ring buffers are sometimes artificially inflating their required size during allocation.  
This is because the allocated buffer sizes are forced to be: 
	an integer multiple of granularity
	an integer multiple of (nitems * sizeof_item)
	(see gnuradio/gnuradio-runtime/lib/buffer.cc lines 76 and 121)

It seems that the criteria for allocated buffer size could instead be:
	an integer multiple of granularity 
	>= (nitems * sizeof_item)

The issue this would currently cause has to do with how the circular buffers are indexed (see gnuradio/gnuradio-runtime/include/gnuradio/buffer.h starting at line 150):
d_bufsize, d_write_index, d_read_index are all handled as counters of items (of size d_sizeof_item)

I believe these indices could be tracked/handled as numbers of bytes instead of numbers of items.  

As best I can tell all of the changes would be made in buffer.cc and buffer.h.  
Not being familiar enough with the other scheduler/buffer code files, I cannot yet tell if these are the only files that will be affected.  
Can someone with more exposure to this specific code please comment?  If these are the only files requiring mods, I might attempt to make the modification myself.  

My specific problem involved a vector of 102400 samples of complex data.  
My memory granularity size is 65536 bytes.
I tried to change my vector size to 102404 samples, and I started getting the following error:

	gr::buffer::allocate_buffer: warning: tried to allocate
	4 items of size 819232.  Due to alignment requirements
	2048 were allocated.  If this isn't OK, consider padding 
	your structure to a power-of-two bytes.  
	On this platform, our allocation granularity is 65536 bytes.
	gr::vmcircbuf_mmap_createfilenamming: VirtualAlloc: Error 87: The parameter is incorrect
	gr::buffer::allocate_buffer: failed to allocate buffer of size 1638464 bytes.

So this is saying that I NEED ~3.13Mbytes (but that's not an even integer multiple of granularity size).  
GR is helping me out by trying to allocate ~1600Mbytes.
With the recommended change, this need would drop to ~3.19Mbytes (that's 51 * granularity size).

This is not a new issue: 
https://github.com/gnuradio/gnuradio/issues/1541 
https://www.ruby-forum.com/topic/212300

willcode:
[...]

PR #6348, opened 2022-11-08
GR 4.0: new buffer API proposal

RalphSteinhagen:
## ... and corresponding lock-free `buffer_host<T>` POC implementation
This proposes a new type-safe, reusable, and efficient buffer interface definition using more modern C++ paradigms.
Its medium-term goal is to replace the existing [buffer](https://github.com/gnuradio/gnuradio/blob/46d293aa14ed5d9050aa5857457b136d00fbc996/gr/include/gnuradio/buffer.h#L123-L328) definition and related implementations (e.g. [buffer_cpu_vmcirc](https://github.com/gnuradio/gnuradio/blob/46d293aa14ed5d9050aa5857457b136d00fbc996/gr/include/gnuradio/buffer_cpu_vmcirc.h#L20-L50)) in view of a more robust separation-of-concerns by narrowing/sharpening/simplifying the interface. The expressed intention is to:
  * support in addition to fundamental/existing types also templated/user-defined data types (e.g. PMTs, `DataSet<T>`, ...),
  * be reused outside the flow-graph/inter-block data exchange (e.g. for buffers within blocks, thread-safe PMT-based getter/setters, interfaces to external lib APIs, etc.),
  * lowers the threshold for new/better user implementations (e.g. that are optimised for new HW architectures), 
  * improve testability, exchangeability, and thus easier long-term maintenance as well as future refactoring.

The core API design deploys a [single,multiple]-writer-multiple-reader pattern by defining the following `Buffer` C++ concepts:
```cpp
template<class T, typename U1 = T, typename U2 = T>
concept Buffer = requires(T t, const std::size_t min_size) {
    { T(min_size) };
    { t.size() }              -> std::same_as<std::size_t>;
    { t.newReaderInstance() } -> BufferReader;
    { t.newWriterInstance() } -> BufferWriter;
};
```
And `BufferReader` and `BufferWriter`concepts:
```cpp
template<class T>
concept BufferReader = requires(T /*const*/ t, const std::size_t n_items) {
    { t.get(n_items) }     -> std::same_as<std::span<const util::first_template_arg_v<T>>>;
    { t.consume(n_items) } -> std::same_as<bool>;
    { t.available() }      -> std::same_as<std::size_t>;
};

template<class T, typename ...Args>
concept BufferWriter = requires(T t, const std::size_t n_items, Args ...args) {
    { t.publish([](std::span<util::first_template_arg_v<T>> &writableData, Args ...args) {}, n_items, args...) } -> std::same_as<void>;
    { t.tryPublish([](std::span<util::first_template_arg_v<T>> &writableData, Args ...args){}, n_items, args...) } -> std::same_as<bool>;
    { t.available() }         -> std::same_as<std::size_t>;
};
```
The specific names and additional interfaces are open for discussion and subject to change.

### Example code for the writer (for details see [here](https://github.com/RalphSteinhagen/gnuradio/blob/c5d4020dff6d62ae1d3aadebca555199df9b831e/test/qa_buffer.cc#L138-L194)):
```cpp
BufferWriter auto writer = buffer.newWriterInstance();
{ // write-only worker (source)
    auto lambda = [](auto w) { // test writer generating consecutive samples
        static int offset = 1;
        std::iota(w.begin(), w.end(), offset);
        offset += w.size();
    };

    EXPECT_GE(writer.available(), buffer.size());
    writer.publish(lambda, 10);
    EXPECT_EQ(writer.available(), buffer.size() - 10);
}
```
*N.B. the lambda ensures thread safety and a syntax-safer claim/publish guarantee which is especially critical for the writers.*
### Example code for the reader:
```cpp
BufferReader auto reader = buffer.newReaderInstance();
{ // read-only worker (sink)
    std::span<const int32_t> fixedLength = reader.get(3); // explicitly typed for illustration, get '3' samples only
    auto available = reader.get(); // get all available samples
    
    // thread-safe use of data

    // consume data -> allows the corresponding buffer to be overwritten by writer
    // if there are no other reader claiming that buffer segment
    if (!reader.consume(fixedLength.size())) {
        throw std::runtime_error(fmt::format("could not consume data"));
    }
}
```

## Testing Done
There is a `buffer_host<T>` proof-of-concept implementation (following largely the low-level double-mapped memory strategy also used by the `buffer_cpu_vmcirc` i.e. no new black magic), corresponding [unit-tests](https://github.com/RalphSteinhagen/gnuradio/blob/c5d4020dff6d62ae1d3aadebca555199df9b831e/test/qa_buffer.cc) and a simple [bm_buffer](https://github.com/RalphSteinhagen/gnuradio/blob/c5d4020dff6d62ae1d3aadebca555199df9b831e/bench/bm_buffer.cpp) benchmark that can be invoked via 
<details>
  <summary>For details click here</summary>

```text
./bm_buffer --help
App description
Usage: ./bm_buffer [OPTIONS]

Options:
  -h,--help                   Print this help message and exit
  --samples UINT              minimum number of samples
  --veclen INT                vector length (<-> size of bulk operation)
  --nreader INT               number of consumers
  --buffer_type INT           Buffer Type (0:simple, 1:vmcirc, 2:cuda, 3:cuda_pinned, 4: buffer_host - POSIX, 5: portable, 6: malloc/C-Style)
  --throughput                whether to print samples/s rather than time
  --rt_prio                   enable real-time priority
  --scan                      scan producer-consumer parameter space

```
</details>

Some preliminary results indicate that using the proposed new API, there is no regression but even a slight performance improvement. The results where taken on an AMD Zen3 (Ryzen/EPYC) architecture:

bm_buffer_quick_bench1

The variability is due to the high writer-reader contention (i.e. 'produce/consume as fast as possible') and could be further improved depending on the chosen waiting strategy.  N.B. you can guess the number of available cores based on when readers are starting to fight for CPU/memory resources. This is also an issue w.r.t. the general GR 4.0 scheduling paradigm that should allow limiting the number and affinity of threads that are concurrently running to the number of cores (CPU shielding paradigm) to achieve the best possible performance and keeping hot L1/L2/L3 chaches.

Intel i7-8565U results:
bm_buffer_quick_bench2

## Open Points -- Design Choices
This is a draft proposal. The following design choices have been identified that need to be addressed/finalised:
  * [ ] final naming of the interface methods
  * [x] default behaviour w.r.t. 'safety vs. performance' i.e. keep guard-rails for novice/new users and allow GR experts to by-pass these or the other way around
  * [x] templated `t.newReaderInstance<U1>()` and `t.newWriterInstance<U2>()` interface to implicitly convert to user-defined values other than the buffer's `T` (which is potentially unsafe/UB/expert-use only) or to use the new std-lib functions [std::as_byte](https://en.cppreference.com/w/cpp/container/span/as_bytes) and [std::as_writable_bytes](https://en.cppreference.com/w/cpp/container/span/as_bytes).
  * [x] how to handle exceptions thrown in user-code (i.e. ignore/rethrow/log/...) -> has a strong performance impact
  * [x] additional required buffer interfaces
  * [ ] clang-format definitions -- the present default is OK for C-style code but creates sub-optimal outcome for C++
  * [ ] ...

Related, but beyond this PR are the design hierarchy and how buffers tie into the rest of the GR architecture. Generally, this could be more explicitly differentiated between at least the logically separate domains of: 'buffer', 'port', and 'block'. 
  
## Related Issues
The buffer API and implementation are designed to be reusable. The following issues have been identified and may further profit from this improvement:
  * #1768
  * #6339
  * #6217 (N.B. this implementation already supports thread-safe adding of readers)
  * #6312
  * ...

While being a stand-alone and header-only implementation, the general intent is that this will supersede the existing [buffer](https://github.com/gnuradio/gnuradio/blob/46d293aa14ed5d9050aa5857457b136d00fbc996/gr/include/gnuradio/buffer.h#L123-L328) interface.

## Checklist

This has been co-authored with @ivan-cukic.

Open for discussion. Please be kind and open-minded.


gr/include/gnuradio/wait_strategy.hpp
```diff
@@ -0,0 +1,373 @@
...
+template<std::int32_t YIELD_THRESHOLD = 10, std::int32_t SLEEP_0_EVERY_HOW_MANY_TIMES = 5, std::int32_t SLEEP_1_EVERY_HOW_MANY_TIMES = 20>
+class SpinWait {
+    using Clock         = std::conditional_t<std::chrono::high_resolution_clock::is_steady, std::chrono::high_resolution_clock, std::chrono::steady_clock>;
+    std::int32_t _count = 0;
+    static void  spinWaitInternal(std::int32_t iterationCount) noexcept {
+        for (auto i = 0; i < iterationCount; i++) {
+            yieldProcessor();
+        }
+    }
+    static void yieldProcessor() noexcept { asm volatile("rep\nnop"); }
```
marcusmueller, line 251: um, slightly confused, what is assembler doing in a scheduling algorithm?

marcusmueller:
Got a bit of a question with claim_strategy.hpp. Am I mistaken in thinking this isn't used?

gr/include/gnuradio/wait_strategy.hpp
RalphSteinhagen: that's a guaranteed 'NOOP' operation that the compiler cannot remove/optimise away. This is needed for ultra-high update rate when sometimes it's enough to wait for one clock cycle to resolve contention on a shared resource (here: atomics).

RalphSteinhagen:
> Got a bit of a question with claim_strategy.hpp. Am I mistaken in thinking this isn't used? Can't find it included in any other file of this PR

@marcusmueller it's included here:

https://github.com/gnuradio/gnuradio/blob/931f07ecab76a02a2b57958c00c8f5e843ae8f1f/gr/include/gnuradio/buffer_host.hpp#L37

RalphSteinhagen:
Thanks again for your interest, comments, and suggestions during our last GR architecture meeting. As mentioned during the meeting the implementation is intentionally kept generic enough to cover the following three use cases:
image
with the 'fan-out' being the most common use case followed by the 'aggregate' one that supports message passing of arbitrary user types. The latter should ideally be cache-line aligned to fit well within a POSIX memory page size, otherwise, the STD and copy-based implementation yields better results (fast/smaller minimum buffers). I updated the implementation to incorporate changes regarding default safeguards (N.B. unsafe raw-memory access is still possible), and notably added a thread-safe and race-condition-free example of how this could be used for. for example, passing stream tags alongside the main stream data.

The low-level infrastructure for the third 'multi-cascade' use case is already in place but would require some smart/intuitive user-facing API to be designed. In essence, a tuple consisting of '{cursor, {reader}}' `Sequence`s need to be maintained per hop. Should this be an indexed writer/reader group? Explicit reference to previous readers/writers? ... The nominal case is easy (i.e. static construction/destruction of the whole sub-graph using the buffer), but need to spend some thought on how to safely deal with it if only some blocks are being taken out from the sub-graph. This is not (yet) a primary use case within GR, thus I reckon that the solution is good enough and can be used as is -- thus also why I am proposing to open up this PR for public review. Improvements could/should maybe be done in follow-up PRs.

Below are the updated results (5 passes for each n-reader/buffer combination) for a 24-core AMD Zen3 architecture:
BufferPerformance

The following things are kept open because they should IMO be reviewed on a larger GR 4.0 scope:
  * final naming of the interface methods: e.g.
      * naming of `buffer_host` which strictly speaking is generic enough to also cover CUDA buffers etc. 
      * also links to a general concept for naming things...
  * code-formatting and -style/guideline choices, some of the present ones obfuscate the readability and safety of C++ code.

IMHO C++ can be used to produce simpler and much safer code. I hope that this PR is an example that this is not at the expense of performance (e.g. looking at a 'throughput' metric only) but rather that this also enhances performance by writing code that is more flexible, easier to be optimised by compilers, while remaining readable.

gr/include/gnuradio/wait_strategy.hpp
marcusmueller: if we care about that, then the 60ns race condition window we've discussed on call is a *humongous* problem…

marcusmueller: plus, I mean, makes no sense to have this in code that's not going to be compiled for x86, so hm.

mormj:
Based on offline conversations, the plan is to merge this standalone, as in not yet change the overall buffer API (and propagated effects to other APIs) but have the classes available to be used for a universal data sink component.

A couple of things I'd like to see fixed up before we merge:
1) For naming, how about `buffer.hpp` --> `buffer_api.hpp` and `buffer_host.hpp` --> `buffer_cpu_dblmapped.hpp` or something to that effect
2) For code consistency, it would be good to keep `snake_case`
3) There are compile failures on Ubuntu and F35

At least on Ubuntu 22, the only way I found to resolve these are:
in sequence.hpp
+#include <fmt/core.h>
+#include <fmt/ostream.h>

then updating the minimum gtest to 1.11.  Without it there are some linker errors for the print functions for pmr 
Unfortunately this puts another subproject in the repo, but that is probably fine??

The other funny thing with meson and gtest is with 1.11 the `main:true` option seems to not work anymore, so I had to propagate gtest_dep and also gtest_main_dep to all the qa targets.  This includes PMT library

RalphSteinhagen:
> 1. For naming, how about buffer.hpp --> buffer_api.hpp and buffer_host.hpp --> buffer_cpu_dblmapped.hpp or something to that effect

Naming is hard :smile:: 

  * The `buffer.hpp` --> `buffer_api.hpp` is probably fine, if the other interface definitions are also going to be named `_api.hpp`. 
  * The `buffer_cpu_dblmapped.hpp` name isn't ideal, because the implementation works equally for CPU|GPU|... as well as dblmapped and using std allocators. I.e. by just adding < 25 lines for CUDA allocators this would be functionally equivalent and could (if desired) replace: `buffer_cpu_simple`, `buffer_sm`, `buffer_cpu_vmcirc`, `buffer_cuda_...`

RalphSteinhagen:
> 3. There are compile failures on Ubuntu and F35

Thanks for the pointers, I didn't have the means to check on those platforms locally and part of it seemed to be related to the non-standard/common way how how {fmt}, spdlog etc. are packaged/included in the code and other things...

mormj (approved):
Looking forward to seeing this buffer concept/class get used.  Going ahead and merging

Issue #5841, opened 2022-05-08
MacOS: GNU Radio only 1/4 the size of the GTK window


fah:
### What happened?

GTK windows appears with only top left quarter is filled with gnu-radio screen.
![GR_BUG](https://user-images.githubusercontent.com/322496/167308109-fe08431d-007e-4c58-8e75-cf698dfb7461.jpg)

----

I tried the following patch _reverse_:
```
sudo vim /usr/local/lib/python3.9/site-packages/gnuradio/grc/gui/DrawingArea.py

  def _update_size(self):
        w, h = self._flow_graph.get_extents()[2:]
        """
        PATCH maually applied: https://github.com/conda-forge/gnuradio-feedstock/pull/85/files
        OLD:
        self.set_size_request(w * self.zoom_factor + 100, h * self.zoom_factor + 100)
        """
        scale_factor = self.get_scale_factor()
        self.set_size_request(
            w * scale_factor * self.zoom_factor + 100,
            h * scale_factor * self.zoom_factor + 100,
        )

```
Which didn't change anything.

### System Information

Installing GNU Radio via  
[CondaInstall - GNU Radio](https://wiki.gnuradio.org/index.php/CondaInstall)
on an iMac with native Graphic set to normal size. Same happens with other resolutions, too.

Versions:
MacOS 11.6.4
GNU Radio Companion 3.10.2.0 
conda 4.12.0

### GNU Radio Version

3.10 (maint-3.10)

### Specific Version

3.10.2.0

### Steps to Reproduce the Problem

1. Start ```gnuradio-companion``` from commandline
2. GTK windows appears with **only top left quarter is filled with gnu-radio screen**.
3. Click on a menu -> crash with log output.  

### Relevant log output

```shell
Log:

(gnuradio) x@iMac Project % gnuradio-companion
>>> Warning: vocoder_codec2_decode_ps - option_attributes are for enums only, ignoring
>>> Warning: vocoder_codec2_encode_sp - option_attributes are for enums only, ignoring
<<< Welcome to GNU Radio Companion 3.10.2.0 >>>

Block paths:
	/Users/x/.grc_gnuradio
	/Users/x/opt/anaconda3/envs/gnuradio/share/gnuradio/grc/blocks

Loading: "/Users/x/radio_3.9.grc"
>>> Done

Loading: "/Users/x/Downloads/Gnuradio/GNU Radio_samples/wbfm with audio out.grc"
>>> Done

[...]

PR #6352, opened 2022-11-10
grc: Fix cairo assertion failure by not storing reference to context

### LINKED ISSUE #2726: GRC38 OSX +Quartz Runtime Crashing
Original thread on gnuradio-discuss is here: https://lists.gnu.org/archive/html/discuss-gnuradio/2019-08/msg00031.html . The basic issue occurs with GRC in GR 3.8.0.0 on OSX when using Quartz for GUI display, not X11. X11 GUI seems to work correctly. The issue is that dragging a block from the righthand list onto a FG will result in a crash.

### LINKED ISSUE #2938: Resizing crashes: Assertion failed: CAIRO_REFERENCE_COUNT_HAS_REFERENCE (&surface->ref_count), file cairo-surface.c, line 955
Gnuradio 3.8.0.0 installed on latest Windows 10.
When the window is resized, gnuradio-companion crashes:

```
C:\Program Files\GNURadio-3.8\bin>"C:\Program Files\GNURadio-3.8\bin\run_gr.bat" gnuradio-companion.py
setting gnuradio environment
WARNING:gnuradio.grc.core.platform:Slow YAML loading (libyaml not available)
<<< Welcome to GNU Radio Companion 3.8.0.0 >>>

Block paths:
        C:\Program Files\GNURadio-3.8\share\gnuradio\grc\blocks

Loading: "O:\local\test_qt.grc"
>>> Done
Assertion failed: CAIRO_REFERENCE_COUNT_HAS_REFERENCE (&surface->ref_count), file cairo-surface.c, line 955

C:\Program Files\GNURadio-3.8\bin>



```

### LINKED ISSUE #5431: Gnuradio-companion tends to crash on macOS 10.15.7 when connecting blocks
### What happened?

A bug happened!

### System Information

OS: macOS 10.15.7
GR Installation Method: Homebrew

### GNU Radio Version

3.9 (maint-3.9)

### Specific Version

3.9.3.0

Launch gnuradio-companion and any two blocks. Try connecting the output of the first block to the input of the second one. This tends to cause a crash.

Note that this started happening not instantly, but after I was playing with gnuradio-companion for a bit.

### Relevant log output

```
2021-12-29 14:52:12.650 Python[75255:1441851] unlockFocus called too many times. Called on <GdkQuartzView: 0x7f8a2a133c80>.
Segmentation fault: 11
```

### LINKED ISSUE #5734: M1 macOS: UI becomes unresponsive after connecting two nodes
### What happened?

After trying to connect two nodes the UI becomes unresposive.

### System Information

OS: MacOS 12.1 Monterey
Architecture: M1, x86 Emulation on
GR Installation Method: `arch -x86_64 /usr/local/bin/brew install gnuradio`

### GNU Radio Version

3.9  (maint-3.9)

### Specific Version

3.9.3.0

1. Obtain M1 mac
2. Install Homebrew
3. Install GNURadio using `arch -x86_64 /usr/local/bin/brew install gnuradio`
4. Run `gnuradio-companion`
5. Create two nodes
6. Connect them
7. The UI will flicker and will become unresposive. Buttons can still be pressed but the interaction is not represented visually.

### Relevant log output

```shell
Nothing, despite running `gnuradio-companion --log debug`
```

ryanvolz:
## Description

The connection flowgraph element previously kept a reference to the cairo context passed to the draw function in order to be able to use cairo's `in_stroke` function to determine if the mouse cursor was on the path of the curved connection line. As it turns out, this is dangerous because GTK is constantly destroying and creating new cairo contexts and surfaces.

This avoids keeping a reference to the cairo context by initializing a local cairo context with the connection class for the sole purpose of storing the curved path and calculating `in_stroke`. The local context and surface can be very basic because not much is needed for `in_stroke`, and the context can persist and just have its path replaced when it needs to be updated.

On Windows, resizing the GRC window (and particularly the flowgraph canvas) to a larger size than its initial size would cause the underlying cairo surface to be destroyed and a new one created. The cairo context stored for the curved connection line had a reference to that destroyed surface, and closing that context (upon replacing it with a new one) would attempt to decrement the reference count on the surface. This would produce [an assertion failure that crashed GRC stating `Assertion failed: CAIRO_REFERENCE_COUNT_HAS_REFERENCE (&surface->ref_count)`](https://github.com/gnuradio/gnuradio/issues/2938).

On macOS with the Quartz backend, [a related crash and assertion error would manifest when connecting blocks](https://github.com/gnuradio/gnuradio/issues/2726). I only figured out that these two were essentially the same bug when I saw that my fix to the Windows crash worked around the same piece of code as [an older patch](https://github.com/eblot/gnuradio/commit/61be64deb0cdfc6b356d4391151b1134ddee7437) carried by MacPorts and, up until recently until it was dropped because of another bug, the conda package. Considering the other bug that that patch triggered with the conda-forge `cairo` (which has since been fixed), I think this approach is lighter weight and has less potential for future problems.

## Related Issue

Fixes #2726.
Fixes #2938.
Fixes #5431.
Fixes #5734.

GRC

## Testing Done

I found a reliable way to trigger #2938 on Windows, and with this fix I don't see any crashes. Selecting connection lines still works as well.

## Checklist

- [x] ~~I have added tests to cover my changes, and all previous tests pass.~~

### DIFF
--- a/grc/gui/DrawingArea.py
+++ b/grc/gui/DrawingArea.py
@@ -169,10 +169,9 @@ class DrawingArea(Gtk.DrawingArea):
 
     def _update_size(self):
         w, h = self._flow_graph.get_extents()[2:]
-        scale_factor = self.get_scale_factor()
         self.set_size_request(
-            w * scale_factor * self.zoom_factor + 100,
-            h * scale_factor * self.zoom_factor + 100,
+            w * self.zoom_factor + 100,
+            h * self.zoom_factor + 100,
         )
 
     def _auto_scroll(self, event):
--- a/grc/gui/canvas/connection.py
+++ b/grc/gui/canvas/connection.py
@@ -10,6 +10,8 @@ SPDX-License-Identifier: GPL-2.0-or-later
 from argparse import Namespace
 from math import pi
 
+import cairo
+
 from . import colors
 from .drawable import Drawable
 from .. import Utils
@@ -45,8 +47,11 @@ class Connection(CoreConnection, Drawable):
 
         self._rel_points = None  # connection coordinates relative to sink/source
         self._arrow_rotation = 0.0  # rotation of the arrow in radians
-        self._current_cr = None  # for what_is_selected() of curved line
         self._line_path = None
+        # simple cairo context for curved line and computing what_is_selected
+        cr = cairo.Context(cairo.RecordingSurface(cairo.CONTENT_ALPHA, None))
+        cr.set_line_width(cr.get_line_width() * LINE_SELECT_SENSITIVITY)
+        self._line_path_cr = cr
 
     @nop_write
     @property
@@ -125,12 +130,13 @@ class Connection(CoreConnection, Drawable):
             cr.curve_to(*(p2 + p3 + p4))
             cr.line_to(*p5)
             self._line_path = cr.copy_path()
+            self._line_path_cr.new_path()
+            self._line_path_cr.append_path(self._line_path)
 
     def draw(self, cr):
         """
         Draw the connection.
         """
-        self._current_cr = cr
         sink = self.sink_port
         source = self.source_port
 
@@ -192,18 +198,11 @@ class Connection(CoreConnection, Drawable):
         if coor_m:
             return Drawable.what_is_selected(self, coor, coor_m)
 
-        x, y = [a - b for a, b in zip(coor, self.coordinate)]
-
-        cr = self._current_cr
-
-        if cr is None:
+        if self._line_path is None:
             return
-        cr.save()
-        cr.new_path()
-        cr.append_path(self._line_path)
-        cr.set_line_width(cr.get_line_width() * LINE_SELECT_SENSITIVITY)
-        hit = cr.in_stroke(x, y)
-        cr.restore()
+
+        x, y = [a - b for a, b in zip(coor, self.coordinate)]
+        hit = self._line_path_cr.in_stroke(x, y)
 
         if hit:
             return self


ryanvolz:
I now also think that #5111 was a partial, circumstantial workaround for the same bug that this addresses, so I've added a commit to undo that one. Now that I understand better how cairo and GTK work, it's clear that it wasn't fixing any scaling issues and only had the effect of making the flowgraph canvas larger on high DPI displays, which just so happened to prevent triggering this bug for some of those users. That analysis agrees with user experience with actual scaling bugs like #5841, where having that commit or reversing it made no difference.

mormj:
Anyone able to test this out on mac?  (tagging names from #4174)
@geraman21
@NeuroForLunch
@jackdally
@sam-k
@mkierczak

ahooper:
Thanks! This fixed my crash connecting blocks. #5431 GNU Radio Companion 3.10.4.0 MacOS 10.15.7

willcode:
WRT `get_scale_factor()`, this doesn't affect anything for HiDPI? Since the scale factor isn't used anywhere else in GRC, it was affecting only the `DrawingArea` and not any of the content, so I'm pretty sure it was already useless ... just checking.

RobertSwirsky:
Thank you! This bug has been plaguing me for years! I replaced the two python files on my Windows installation, and all is fine (and the fix makes sense, too.)

ryanvolz:
> WRT `get_scale_factor()`, this doesn't affect anything for HiDPI? Since the scale factor isn't used anywhere else in GRC, it was affecting only the `DrawingArea` and not any of the content, so I'm pretty sure it was already useless ... just checking.

Right, it only changes the size of the `DrawingArea` and so is not actually helpful for HiDPI. Before understanding this and knowing that including the `get_scale_factor()` fixed problems for some people, I thought that it might be because the rest of the GUI was scaled on HiDPI systems but the `DrawingArea` wasn't because it worked in native pixels, but that's not true. Everything including the `DrawingArea` gets scaled properly, and since I don't have HiDPI I've tested this on my non-HiDPI system by setting `GDK_SCALE=2` when running GRC and finding that everything does in fact double in size. By including the `get_scale_factor` stuff, the `DrawingArea` just gets doubled in size again with the side effect of making this bug harder to trigger. So now that this actually fixes the bug, the `get_scale_factor` is just undesirable.

ryanvolz:
We have confirmation from a few users now in the various bug reports that this fixes those issues (linked by the PR). I haven't heard back yet about a few of the other issues that are potentially related (e.g. GRC 1/4 window size on macOS), but the reports now cover everything I thought should be fixed.

mormj (approved):
Based on the confirmation of fixes, let's go ahead and merge this one!  Very nice!

dkozel:
This is *huge*. This could be the most impactful bugfix of the year as the combo of issues have been total blockers for huge groups of potential users. Ryan, thank you so much for tracking this down.

cryptik:
I am running GnuRadio v 3.10.4.0 installed from home-brew on Ventura 13.0.1 and an M1 and getting the segfaults when using gnu-radio companion.  It does not segfault right away... but at random times... even if just moving around the widget menu.  How can I tell if this particular fix (#6352) is included in the version that I am running?  Should I build from source?

ryanvolz:
@cryptik This fix won't be in your 3.10.4.0 installation, but it's easy enough to modify it without even reinstalling. Just find where Homebrew has installed the particular Python files that this changes and edit them manually as in this PR.

cryptik:
Hi @ryanvolz, thanks for the quick response.  I was able to modify the files indicated in the PR.  I am not sure if this is the correct place, but in my Homebrew install, it appears the python files are located here (/opt/homebrew/Cellar/gnuradio/3.10.4.0_1/lib/python3.10/site-packages/gnuradio/grc/gui).  Hopefully there are not also copied somewhere else.  After making the changes, GRC is hugely more stable.  I do still get segfaults, but they are very infrequent now.

cryptik:
I might have a different problem.  It seems to still segfault whenever I try and load a .grc file from disk.  It loads the file but as soon as I move the mouse, it segfaults.  Running a .grc file seems to do the same thing, but on a more random basis.  This Mac M1 has been more trouble than it's worth.  I so wish I could return it.

PR #6361, opened 2022-11-15
config files: fix misleading config hints

dl1ksv:
## Description

OOT's build out of tree are not found by grc.
To find these blocks a path entry either in
etc/gnuradio/conf.d/grc.yml ( global setting )
or in ~/.gnuradio/config.yml(user specific ) is necessary.

At the moment the hint points to config.conf wich is only valid for 3.0.

In addition I added an configuration option to set the OOT block path during installation. An empty string keeps the old behaviour.

## Testing Done

Applied this fix and had a look at 
etc/gnuradio/conf.d/grc.yml 

## Checklist

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -156,6 +156,6 @@ summary({'bindir': get_option('bindir'),
         'libdir': get_option('libdir'),
         'datadir': get_option('datadir'),
         'prefix': get_option('prefix'),
+        'Custom blocks path': get_option('custom_blocks_path'),
+        'Python dir': py3_install_dir,
         }, section: 'Directories')
-message('Python dir:')
-message(py3_install_dir)
--- a/grc/grc.yml.in
+++ b/grc/grc.yml.in
@@ -1,10 +1,10 @@
 # This file contains system wide configuration data for GNU Radio.
 # You may override any setting on a per-user basis by editing
-# ~/.gnuradio/config.conf
+# ~/.gnuradio/config.yml
 
 grc:
     global_blocks_path: @blocksdir@
-    local_blocks_path:
+    local_blocks_path: @localblocks@
     default_flow_graph:
     xterm_executable: @GRC_XTERM_EXE@
     canvas_font_size: 8
--- a/grc/gui/Config.py
+++ b/grc/gui/Config.py
@@ -17,8 +17,8 @@ from . import Constants
 HEADER = """\
 # This contains only GUI settings for GRC and is not meant for users to edit.
 #
-# GRC settings not accessible through the GUI are in config.conf under
-# section [grc].
+# GRC settings not accessible through the GUI are in config.yml under
+# section grc:.
 
 """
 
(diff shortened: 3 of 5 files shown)


meson.build
```diff
@@ -156,6 +156,6 @@ summary({'bindir': get_option('bindir'),
         'libdir': get_option('libdir'),
         'datadir': get_option('datadir'),
         'prefix': get_option('prefix'),
+        'OOT blocks dir': get_option('oot_blocks_dir'),
```
mormj, line 159: Is there a distinction between OOT blocks directory and your in-tree blocks directory?

dl1ksv: Yes, if you install your OOT block intree the yml files go to prefix/GRC_BLOCKS_DIR which mostly is
prefix/share/gnuradio/grc/blocks. prefix is the installation prefix of gnuradio. In this case the oot_blocks_dir is ''.

But if you install the OOT blocks out of  ( the gnuradio ) tree,  the yml files go to 
(oot install ) prefix/GRC_BLOCKS_DIR.

In this case the oot_blocks_dir should be set to the oot install prefix, to let grc find the oot blocks.

mormj: what is the equivalent of this line in GR3 cmake?

dl1ksv: There is no equivalent at the moment. 
The starting point in GR3 is the file grc.conf build from grc.conf.in and in GR4 grc.yml build from grc.yml.in.
In both cases there is a reference to
local_blocks_path
which is empty and there is no way  to set this reference during  the build process. So this variable is not reported at the end of the cmake run. 

But was does local_blocks_path mean?
Some time ago I installed my OOT modules into the gnuradio tree. So these modules were found by grc.
But then I decided to keep my OOT modules completely separated from the gnuradio installation. But then you must tell grc where to find the OOT modules. 
I thought the local_blocks_path would be a good choice.
At the moment you can only edit etc/gnuradio/conf.d/grc.conf (GR3) or etc/gnuradio/conf.d/grc.yml (GR4)
So I created the option 'oot_blocks_dir' which can be used to set the local_blocks_path during meson setup.

mormj: So then I think this just comes down to a choice over naming - I'm not big on calling it OOT blocks path because that implies that all OOT modules are going to get installed there.  Perhaps it remains local_blocks_path but that can be set during meson build.  Or maybe alternate_blocks_path?

dl1ksv: Yes, I think we discuss the best choice of name.  I wonder , if this path variable has been used frequently and for what reason. Nevertheless I'd like to keep the key 'local_blocks_path' in the grc.yml and config.yml files.
But I think the name of the option should be different from this key, as it's value is set to
{option}/GRC_BLOCKS_DIR. So alternate_blocks_path would be an acceptable choice for me.
But is there a difference between alternate blocks and OOT blocks ?

mormj: Perhaps it's nitpicky, but I feel like OOT blocks path implies that all OOT blocks will get installed there.  custom_blocks_path?

mormj (approved):

mormj:
thanks for the fixes here!

Issue #6253, opened 2022-10-15
GR 4.0: standardise sample rate, signal-name and other DAQ meta-information using stream tags


RalphSteinhagen:
### Feature Description

Presently, the sample rate, signal names, and other GR source meta-information need to be propagated as block parameters to each block where required. While variable definitions shared across blocks help, these are error-prone for users for cases where the sampling rate changes dynamically and/or due to known decimation in down-stream flow-graph segments and blocks/sinks that require this information during their initialisation.

As a maybe more intuitive approach, this and other signal meta-information could be standardised using stream-tags with documented PMT key-names, that are added to the first emitted sample after the flow-graph is initialised and/or whenever the given information changes during run-time. 

To keep it vendor-neutral and 'nomen-est-omen' the following naming are proposed (`SIG` prefix tbd.):
  1. `SIG_RATE` - the 'uint64_t' encoded sample rate (presently often defined as `samp_rate` global variable)
  1. `SIG_NAME` - a 'std::string' encoded name of the signal (N.B. relevant for MIMO & SDRs  with a high channel count)
  1. `SIG_UNIT` - a 'std::string' encoding the physical unit of the signal (optional)
      (N.B. used for debugging/UI and as a light-weight pre-cursor to #6183)
  1. `SIG_MIN`  - an optional 'float' to encode physical DAQ limits
  1. `SIG_MAX`  - an optional 'float' to encode physical DAQ limits
  1. *...additional user-defined DAQ properties...*
  
Blocks requiring this information would read it during initialisation/with the first sample and blocks not participating in these schemes would merely forward the tag information untouched.

Ideally, the above would be incorporated by all sources and blocks, but would probably be only strictly needed for the `SIG_RATE' and for existing blocks that require the 'samp_rate' info.

### Feature Urgency

high (impacts broad portion of ecosystem)

### More Information

Some usage/implementation examples:
 * a 10x decimating block would change `SIG_RATE' from '1000' to '100' while leaving the others untouched
 * an 'adder' would concatenate the `SIG_NAME` of its inputs, e.g. `SignalA+SignalB`
 * a UI widget would use all of the above tag information to setup the default chart and axis parameters
 * a file/data storage sinks automatically document the meta-information w/o needing to configure block parameters
 
N.B. blocks may choose to use these tags only during their first work function invocation as a run-time performance optimisation.

Open for comments, suggestions, additional thoughts, and applications.

jacobagilbert:
[...]

Issue #6252, opened 2022-10-15
GR 4.0: standardise synchronisation of data streams from different flow-graphs & nodes using tags


RalphSteinhagen:
### Feature Description

This feature request does not require changes to the GR core but proposes standardised conventions for the naming and layout of PMT-based trigger tags and meta-information to facilitate synchronising signal streams across different flow-graphs, nodes, and in- and OOT-source devices that are part of the greater GNU Radio eco-systems. The implementation of the specific synchronisation (blocks) would be initially OOT.

It tackles the real-world use-case in MIMO systems where -- for a large number or physically-distributed signal sources -- not all RX signals can be synchronised by using a single SDR but that require a large and/or distributed DAQ infrastructure.

![image](https://user-images.githubusercontent.com/46007894/195979357-0d6c00ab-e5db-4c06-862b-c091742044fc.png)

This is not a new concept (i.e. already used in some LimeSDR and UHD devices) but merely aims to standardise the existing solutions as vendor-neutral conventions. In the proposed scheme, signal sources should
1. map their external (usually HW-based) or internal(usually SW-based) synchronisation triggers as stream tags that are placed on the sample of the exact trigger occurrence. As a minimum key-value-pair information (`TRIGGER` prefix tbd.):
    * `TRIGGER_NAME` - a 'std::string' (default, e.g. 'PPS' -- to be overwritten by info in 'b)')
    * `TRIGGER_TIME` - a 'uint64_t' UTC- or TAI-based time-stamp in nanoseconds (choice: UTC vs. TAI tbd.)
    * `TRIGGER_OFFSET` - floating-point delay of the sample w.r.t. the trigger in seconds (e.g.compensating analog group delays)
2. provide a 'timing' msg port that receives PMT messages sent ahead of the trigger to
    * change the trigger name, precise time stamp, and offset of the upcoming trigger, for example, based on information from an external high-resolution timing receiver (GPS, White-Rabbit, SW-timing system, ...)
    * add custom meta-information on top of the same trigger tag that is use-case or site-specific (<-> user-extensions)

The main goal is to publicly document/implement this for all in-tree and perhaps also as a recommendation for all OOT signal source blocks to allow participating in and sharing of synchronisation schemes. Blocks not participating in these schemes would be unaffected and merely forward the tag information untouched.

[...]

PR #6362, opened 2022-11-15
adding default stream-tag definitions to GR 4.0

RalphSteinhagen:
This is a follow-up of #6252 and #6253 and provides the following SigMF-inspired definitions and some additional helper methods in `class tag_t` to be used in the context of processing stream tags, e.g.:
```cpp
using namespace gr;
tag_t testTag;

testTag.insert_or_assign(tag::SAMPLE_RATE(4.0f));
//testTag.insert_or_assign(tag::SAMPLE_RATE(5.0)); // type-mismatch -> won't compile
EXPECT_EQ(testTag.at(tag::SAMPLE_RATE), 4.0f);
EXPECT_EQ(tag::SAMPLE_RATE.shortKey(), "sample_rate");

EXPECT_TRUE(testTag.get(tag::SAMPLE_RATE).has_value()); // future: allows for monadic expressions
EXPECT_FALSE(testTag.get(tag::SIGNAL_NAME).has_value());
```
These tags are definitions only and the specific naming be changed if necessary.

These definitions should be replaced by `inlinse constexpr` once PMTF supports `constexpr` constructors. 

N.B. The `pmtf::map` lacks some of the methods compared to STL maps, thus the unusual try-catch and bracket operator usage. Maybe this could be refactored towards the STL contracts?

## Related Issue
  * #6252
  * #6253

No blocks are affected. For the time being, these are just re-usable definitions of recurring stream tag key values.

## Testing Done
added the corresponding unit-tests to the existing ones in `qa_tags.cc`.

## Checklist

### DIFF
--- a/gr/include/gnuradio/tag.h
+++ b/gr/include/gnuradio/tag.h
@@ -1,6 +1,8 @@
 #pragma once
 
 #include <pmtf/wrap.hpp>
+#include <map>
+#include <optional>
 #include <string>
 
 namespace gr {
@@ -32,7 +34,46 @@ public:
     bool operator!=(const tag_t& rhs) const;
 
     void set_offset(uint64_t offset) { _offset = offset; }
-    pmtf::pmt operator[](const std::string& key) const { return _map.at(key); }
+    [[deprecated("unsafe - use at(key) instead")]] pmtf::pmt
+    operator[](const std::string& key) const
+    {
+        return _map.at(key); // implementation violates std::map::operator[] contract
+    }
+
+    pmtf::pmt& at(const std::string& key) { return _map.at(key); }
+    const pmtf::pmt& at(const std::string& key) const { return _map.at(key); }
+
+    [[nodiscard]] std::optional<std::reference_wrapper<const pmtf::pmt>>
+    get(const std::string& key) const noexcept
+    {
+        try { // TODO: missing find function in pmtf::map -> exception based look-up
+            return _map.at(key);
+        } catch (std::out_of_range& e) {
+            return std::nullopt;
+        }
+    }
+    [[nodiscard]] std::optional<std::reference_wrapper<pmtf::pmt>>
+    get(const std::string& key) noexcept
+    {
+        try { // TODO: missing find function in pmtf::map -> exception based look-up
+            return _map.at(key);
+        } catch (std::out_of_range&) {
+            return std::nullopt;
+        }
+    }
+
+    void insert_or_assign(const std::pair<std::string, pmtf::pmt>& value)
+    {
+        // TODO: expose std::map compatible insert & insert_or_assign interface
+        _map[value.first] = value.second;
+    }
+    void insert_or_assign(const std::string& key, const pmtf::pmt& value)
+    {
+        // TODO: expose std::map compatible insert & insert_or_assign interface
+        _map[key] = value;
+        // return _map.insert_or_assign(key, std::move(value));
+    }
+
     uint64_t offset() const { return _offset; }
     pmtf::map map() const { return _map; }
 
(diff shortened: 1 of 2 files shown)


RalphSteinhagen:
Sorry for the noise w.r.t. clang-format checks... it seems that `clang-format version 15.0.3` produces different results compared to the definitions/version used by the CI.

gr/include/gnuradio/tag.h
```diff
@@ -32,7 +34,46 @@ class tag_t
...
+    {
+        // TODO: expose std::map compatible insert & insert_or_assign interface
+        _map[value.first] = value.second;
+    }
+    void insert_or_assign(const std::string& key, const pmtf::pmt& value)
+    {
+        // TODO: expose std::map compatible insert & insert_or_assign interface
+        _map[key] = value;
+        // return _map.insert_or_assign(key, std::move(value));
+    }
```
jsallay, line 75: We are in the process of rewriting the pmts to use variants instead of flatbuffers.  It hasn't been integrated into GR4.0 yet.  But you will have direct access to an std::map, rather than a wrapper class.

gr/include/gnuradio/tag.h
```diff
@@ -44,4 +85,60 @@ class tag_t
...
+inline static const default_tag<pmtf::scalar<float>>    SAMPLE_RATE    = { "sample_rate", "Hz", "signal sample rate"};
+inline static const default_tag<pmtf::scalar<float>>    SIGNAL_RATE    = SAMPLE_RATE;
+inline static const default_tag<pmtf::string>           SIGNAL_NAME    = { "signal_name", "", "signal name"};
+inline static const default_tag<pmtf::string>           SIGNAL_UNIT    = { "signal_unit", "", "signal's physical SI unit"};
+inline static const default_tag<pmtf::scalar<float>>    SIGNAL_MIN     = { "signal_min", "a.u.", "signal physical max. (e.g. DAQ) limit"};
+inline static const default_tag<pmtf::scalar<float>>    SIGNAL_MAX     = { "signal_max", "a.u.", "signal physical max. (e.g. DAQ) limit"};
+inline static const default_tag<pmtf::string>           TRIGGER_NAME   = { "trigger_name"};
+inline static const default_tag<pmtf::scalar<uint64_t>> TRIGGER_TIME   = { "trigger_time", "ns", "UTC-based time-stamp"};
+inline static const default_tag<pmtf::scalar<float>>    TRIGGER_OFFSET = { "trigger_offset", "s", "sample delay w.r.t. the trigger (e.g.compensating analog group delays)"};
+
```
jsallay, line 140: I really like this idea, but I have some notes:
- I really like the idea of having meta data to describe the tag names.  I would pick names that match the sigmf names where possible.  I also have the question of ownership.  Should Gnuradio or Sigmf own these definitions.  (I don't know.)
- Right now, tags are kind of annoying because they aren't documented and you have to know the correct type to use it.  It would be really slick to be able to support a syntax for getting the right type using the key name.  Something like.  `auto sample_rate = tag::SAMPLE_RATE::get_from(testTag);`

gr/include/gnuradio/tag.h
RalphSteinhagen: @jsallay thanks for the info. Regarding this specific PR, should I use a different API member method access?

gr/include/gnuradio/tag.h
RalphSteinhagen: @jsallay how I understood @dkozel: this would be defined within GR, kept compatible in naming, and loosely coupled to SigMF. 

Thus the prefix `gr:` which makes the actual trigger name, for example, `gr:trigger_name` (as also given by `default_tag::key()`. There is also the `default_tag::shortKey()` method that would just yield `trigger_name`.
The actual prefix is arbitrary for now. The main goal is to have a global definition in the `gr::tag` namespace that can be shared across all blocks that choose to use these definitions.

>  It would be really slick to be able to support 

This PR already does: the assignment operator ensures that the correct PMTF type is being used for writing the meta info:
```cpp
testTag.insert_or_assign(tag::SAMPLE_RATE(4.0f));
```
If there is a type-mismatch, the above expression wouldn't even compile. 
Similarly for the get interface which is also strongly typed -- again using the PMTF definition:
```cpp
auto rate = testTag.get(tag::SAMPLE_RATE).value();
```
which returns the specified `pmtf::scalar<float>` type. 

One can of course down-cast/convert these PMTF types to other compatible types if the developer chooses to ...

jsallay: I think this is just me confusing the new variant version with the flatbuffers version.  With the variant, there isn't a need for a `pmtf::scalar<float>` wrapper, because its just as easy to get a reference to the underlying value using `std::get<float>`.  So I  believe this will need to be updated too.

mormj: I think the variant version of the PMT will be *much* better - but we should proceed as if it doesn't exist and refactor later

mormj:
Since the flatbuffer PMTs are all we have as of now, let's go ahead and merge this assuming the syntax can be adjusted when the pmt library is switched over.  Also, I think "sigmf-inspired" standard tags are the right approach given there will not be a 1:1 mapping

Issue #6284, opened 2022-10-23
Subpar logger integration in Python block gateway >= 3.10


marcusmueller:
### What happened?

In the process of rewriting the logging, I did not take care to forward the `d_logger` member of the C++ block to python. 

That leads to ugly constraints: users have to create their own logger and use that.

### System Information

main

### GNU Radio Version

3.11-git (main)

### Specific Version

3.11xxxxgit

### Steps to Reproduce the Problem

make python block

marcusmueller:
closed

PR #6364, opened 2022-11-15
Add self.logger to python block gateway (Rebase of #6299)

mormj:
See #6299 for details


willcode:
Backport note (to self): omit any public additions of `const` to avoid API change.

willcode:
@marcusmueller Reviewing the `const` addtion for `crc_check` ... since the only caller of `msg_handler()` is the runtime, and it's via a lookup in a map, it seems like it would be ok to include this in the backport. I'm confused, though, since `msg_handler_t` does not include `const`, how does this work in the first place?

marcusmueller:
Fixes #6284

PR #6366, opened 2022-11-16
blocks: Fix invalid PCH warning

argilo:
## Description
The following warning occurs when building `file_source_impl.cc`:
```
[ 29%] Building CXX object gr-blocks/lib/CMakeFiles/gnuradio-blocks.dir/file_source_impl.cc.o
cc1plus: warning: /build/gr-blocks/lib/CMakeFiles/gnuradio-blocks.dir/cmake_pch.hxx.gch: not used because `_FILE_OFFSET_BITS' is defined [-Winvalid-pch]
```
This can be fixed by disabling precompiled headers for this file.

File Source

## Testing Done
I verified that it's still possible to open a 5 GB file on 32-bit Linux.

## Checklist

### DIFF
--- a/.github/workflows/make-test.yml
+++ b/.github/workflows/make-test.yml
@@ -64,22 +64,22 @@ jobs:
         include:
           - distro: 'Ubuntu 20.04'
             containerid: 'gnuradio/ci:ubuntu-20.04-3.9'
-            cxxflags: -Werror -Wno-error=invalid-pch
+            cxxflags: -Werror
             ctest_args: '-E ""'
             ldpath:
           - distro: 'Ubuntu 22.04'
             containerid: 'gnuradio/ci:ubuntu-22.04-3.9'
-            cxxflags: -Werror -Wno-error=invalid-pch
+            cxxflags: -Werror
             ctest_args: '-E "qa_polar_..coder_(sc_)?systematic"'
             ldpath:
           - distro: 'Fedora 36 (with 0xFE memory initialization)'
             containerid: 'gnuradio/ci:fedora-36-3.9'
-            cxxflags: -Werror -Wno-error=invalid-pch -Wno-error=deprecated-declarations -ftrivial-auto-var-init=pattern
+            cxxflags: -Werror -Wno-error=deprecated-declarations -ftrivial-auto-var-init=pattern
             ctest_args: '-E ""'
             ldpath: /usr/local/lib64/
           - distro: 'Fedora 37 (with 0xFE memory initialization)'
             containerid: 'gnuradio/ci:fedora-37-3.9'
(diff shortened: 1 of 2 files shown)


argilo:
I expect this warning appeared in #2263.

marcusmueller:
Hm this certainly solves the warning, but is it the right thing to do? We could instead move these extra definitions that we're attaching to one of the file source object file build to a configured header (or seeing that we don't actually make anything configurable, just move that one `#define` into file_source_impl.cc?) and actually get the precompiled headers to be used.

But, let's be honest: I didn't do that back then for *some* reason and instead introduced the warning, as it seems. It would be better if that warning went away, soon, before any of us has the time to look at this.

So: :+1: for the PR; but we should make fixing this an issue.

argilo:
OK, I'll try moving the `#define` into the `.cc` file itself. The idea was proposed here, and it sounds like it worked at the time: https://github.com/gnuradio/gnuradio/pull/2263/files#r243985776

drmpeg:
At the time (long before precompiled headers), I went with the external define because I didn't have a way to test on Windows and thought it had a better chance of working on that platform.

If you try change it, be sure to test on all platforms, 64-bit, 32-bit (especially) and Windows.

argilo:
Hmm, even on the `main` branch I can't open a large file on 32-bit Debian; `fopen` fails with "Value too large for defined data type."

argilo:
I'll mark this as "Don't Merge" until we know what's going on.

argilo:
I've reverted back to the original approach of adding `SKIP_PRECOMPILE_HEADERS ON` to CMake, which results in the correct version of the header getting compiled. I've verified that the File Source block can still open a large file on 32-bit Linux.

argilo:
I suppose if we wanted to use precompiled headers here, we'd have to explicitly use `fopen64` and friends on platforms where they're available.

marcusmueller:
> ~Hmm, even on the `main` branch I can't open a large file on 32-bit Debian; `fopen` fails with "Value too large for defined data type."~
> 
> Update: Things are actually fine on `main`, I just didn't install correctly. But this branch is broken; I think defining `_FILE_OFFSET_BITS` in the C++ file doesn't prevent the precompiled header from being used.

Ah! My hand meets my forehead, and I'm sorry I didn't have that epiphany earlier:

Yeah, the point is that at the point of inclusion of header files, there mustn't be new predefines, otherwise the precompilation cannot be determined to be correct. And that's very much what we want to do here, change the behaviour of the included header by doing that `#define` or `-D`. So, yeah, we'll either have to set that predefine already at the point we precompile the headers, or we'll have to live with the PCHs not matching what we demand, and hence getting correctly ignored.

marcusmueller (approved):

argilo:
I rebased again, since the changes should include the newly-added Fedora 37.

PR #6367, opened 2022-11-17
Update clang format

mormj:
## Description
There is a mismatch between the output of clang-format locally using many distros (ones that have clang-format v14) and that run by the CI.  This updates the github action to use clang-format v14 as well as reformat in-tree to match

Appears to only affect the clang-formatting of the pybind files

## Testing Done
CI should be sufficient

## Checklist

### DIFF
--- a/.github/workflows/make-test.yml
+++ b/.github/workflows/make-test.yml
@@ -15,11 +15,12 @@ jobs:
     timeout-minutes: 15
     steps:
     - uses: actions/checkout@v2
-    - uses: gnuradio/clang-format-lint-action@v0.5-4
+    - uses: DoozyX/clang-format-lint-action@v0.14
       with:
         source: '.'
         exclude: './volk'
         extensions: 'h,hpp,cpp,cc,cc.in'
+        clangFormatVersion: 14
   check-python-formatting:
     name: Check Python Formatting
     runs-on: ubuntu-latest
--- a/gnuradio-runtime/python/gnuradio/gr/bindings/basic_block_python.cc
+++ b/gnuradio-runtime/python/gnuradio/gr/bindings/basic_block_python.cc
@@ -157,24 +157,24 @@ void bind_basic_block(py::module& m)
 
 
         .def("empty_p",
-             (bool (basic_block::*)(pmt::pmt_t)) & basic_block::empty_p,
+             (bool(basic_block::*)(pmt::pmt_t)) & basic_block::empty_p,
              py::arg("which_port"),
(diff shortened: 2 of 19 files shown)


willcode:
Makes sense that using a newer version of clang-format will match systems used by devs. How did you pick the DoozyX version?

mormj:
That seems to be the basis of the clang-format-lint that we are currently using and what comes up in the github action search.  Need to check with @mbr0wn about the best place to grab the action

argilo (approved):

marcusmueller (approved):

willcode (approved):
As long as the action has read-only access, it doesn't really matter what the pedigree is, I guess.

v0.15 is available.

mbr0wn (commented):
If we want to standardize on clang-format 14.0 (which I recommend -- it's the default for Ubuntu 22.04) then I strongly recommend we also update our .clang-format. Yes, it'll make it incompatible with older versions of clang-format, but those don't produce the correct output anyway. Installing clang-format-14 on Ubuntu 20.04 is easy via external repo.

We are currently doing a similar exercise on UHD, and what I did there was run

    clang-format-14 --style=file --dump-config > .clang-format 

to update the file in-tree. It'll read the current file and produce a compatible clang-format 14.0 config file, but with all the settings populated. The resulting file might require some hand-tweaking, though.

My rationale here is that having a more detailed clang-format might reduce the risk of future versions (15, 16, etc.) producing different output. Maybe I'm being too optimistic though.

RalphSteinhagen:
@mbr0wn I concur with your assessment. We had/have similar problems on our projects that rely on clang-format.

I mentioned also to @dkozel, @mormj et al. that the missing rules for, for example, `concept`s, templates, and notably maximum code line length also forced me (and probably others) to either disable clang-format in-code and/or to use obfuscating/shortening type redeclarations which make the C++ code harder (than necessary) to read and write for new/novice but experienced C++ contributors.

I was wondering whether this is an excellent time to also revise some of the code-formatting/-style guidelines.

mormj:
@RalphSteinhagen - agreed that the .clang-format rules fall apart for some of the more modern c++ mechanisms - definitely worth a revisit, especially on `dev-4.0` as they are introduced.

mbr0wn:
> notably maximum code line length

Our current `.clang-format` has a max line length of 90 characters:

https://github.com/gnuradio/gnuradio/blob/24a64c0f050d49af2bb18437503638fade2f3913/.clang-format#L41

PR #6398, opened 2022-11-23
4.0: global: integration with variant based pmt

mormj:

## Description
Latest changes to pmt https://github.com/gnuradio/pmt/pull/67 greatly simplify the interface to be a recursive variant.  This has implications on the GR codebase, but should bring further performance gains

The main changes are
pmtf --> pmtv # not flatbuffer based, but variant based
pmtf::get_as<T> --> pmtv::cast<T> or std::get<T> depending on how strict you want to be
pmtf::scalar,vector,string,map --> pmtv::pmt # no more wrapper classes

Benchmarking results should be updated soon

## Related Issue
https://github.com/gnuradio/pmt/pull/67

ALL OF THEM

## Testing Done
CI tests

## Checklist


mormj:
Note: none of this will work until the `variant` branch is merged on PMT, and also this will require the pmt.wrap in here to update to the head of the PMT repo

RalphSteinhagen:
I like this PMT implementation that inherits from [` rva::variant`](https://github.com/codeinred/recursive-variant) which seems to be [accepted](https://github.com/cplusplus/papers/issues/873) in the upcoming C++23 standard (i.e. inheritance from  `std::variant` STL should officially be safe by then).

Had a bit of time for reading and playing around with it. There is presently this rather long custom definition for:
```cpp
using pmt_var_t = rva::variant<
    std::nullptr_t,
    uint8_t, uint16_t, uint32_t, uint64_t,
    int8_t, int16_t, int32_t, int64_t,
    float, double, std::complex<float>, std::complex<double>,
    std::vector<uint8_t>, std::vector<uint16_t>, std::vector<uint32_t>, std::vector<uint64_t>,
    std::vector<int8_t>, std::vector<int16_t>, std::vector<int32_t>, std::vector<int64_t>,
    std::vector<float>, std::vector<double>,
    std::vector<std::complex<float>>, std::vector<std::complex<double>>,
    std::string,
    std::vector<rva::self_t>,
    std::map<std::string, rva::self_t>>;
 ```

Unless certain combinations are intentionally suppressed (e.g. `std::vector<std::string>`) this definition could be made more generic and the related `std::vector<T>` , `std::complex<T>` and `std::map<std::string, T>` automatically generated for all supported types using a single parameter pack or via types stored in a `std::tuple`:
 ```cpp
using pmt2_var_t = as_pmt_t<rva::variant, uint8_t, uint16_t, uint32_t, uint64_t, int8_t, int16_t, int32_t, int64_t, 
                                           float, double, std::string, rva::self_t>;

// initialisation via type list stored in tuple (N.B. tuple could be extended by user with custom OOT types)
using default_supported_types = std::tuple<uint8_t, uint16_t, uint32_t, uint64_t, int8_t, int16_t, int32_t, int64_t,
                                            float, double, std::string, rva::self_t>;
using pmt3_var_t = as_pmt_t<rva::variant, default_supported_types>; 
```
 The latter tuple-based version would support user-defined types -- if this isn't supposed to be discouraged.  For the corresponding helper templates
<details>
<summary>click here.</summary>

```cpp
namespace detail {
template<template<typename... > class VariantType, typename... Args>
struct as_pmt {
    using type = VariantType<std::nullptr_t, Args..., std::complex<Args>..., std::vector<Args>...,
                            std::vector<std::complex<Args>>..., std::map<std::string, Args>...>;
};

template<template<typename... > class TemplateType, typename ...T>
struct as_pmt<TemplateType, std::tuple<T...>> { using type = as_pmt<TemplateType, T...>::type; };
}

template<template<typename... > class VariantType, class... Args>
using as_pmt_t = detail::as_pmt<VariantType, Args...>::type;
 ```
 </details>
 
 See the corresponding [compiler-explorer example](https://compiler-explorer.com/z/ja3ecoG99) for further details.

**Comment by RalphSteinhagen** [comment]
The `using pmt_var_t` definition above includes `std::nullptr_t` is this intentional/needed somewhere?

File: blocklib/analog/noise_source/noise_source_cpu.cc [hunk 1]
```diff
@@ -38,8 +38,8 @@ work_return_t noise_source_cpu<T>::work(work_io& wio)
     auto out = wio.outputs()[0].items<T>();
     auto noutput_items = wio.outputs()[0].n_items;
 
-    auto type = pmtf::get_as<int>(*this->param_type);
-    auto ampl = pmtf::get_as<float>(*this->param_amplitude);
+    auto type = pmtv::cast<int>(*this->param_type);
```
jsallay, line 41: I would probably use `std::get<int>` unless you have a reason to use cast (here and in all of the other instances of `pmtv::cast`.

jsallay (changes requested):
Overall LGTM.  I would just reconsider the use of `pmtv::cast`

blocklib/analog/noise_source/noise_source_cpu.cc
mormj: there are some instances where pmtv::cast would be appropriate, but more in the case where a block accepts a pmt through a message handler or tag, and it is equally sufficient to provide int, long, etc.  For this case, you are right, the type is fixed through the autogen'd code

jsallay: I agree.  A cast is fine when passing in a pmt.

jsallay (approved):

mormj:
Not sure why CI is failing with this ...  will have to look into it later
```
 Cloning into 'pmt'...
HEAD is now at 4f0b63b meson: Make `pmtf_dep` depend on `pmt_gen_h_dep`. (#66)
```

mormj:
Looks like I didn't have the latest `dev-4.0` locally which had an update to pmt.wrap, then was getting stealthily overwritten by the rebase.

PR #6407, opened 2022-11-29
Add self.logger to python block gateway (backport to maint-3.10)

willcode:
Backport https://github.com/gnuradio/gnuradio/pull/6364

In `gr-digital/lib/crc_check_impl.h`, a `const` is added:

`void msg_handler(const pmt::pmt_t& pmt_msg);`

This is an API change. However, this function is only public so the runtime can call the message handler. If the runtime doesn't care about the change, then I think this is fine.


mormj:
I think that API change is fine and extremely unlikely to impact any external application

mormj (approved):

PR #6426, opened 2022-12-04
dev-4.0: Add alsa source

dl1ksv:
## Description

dev-4.0 contains an alsa sink but no alsa source. For tests it's sometimes helpfull, to use an alsa source.

This pr is a port of the alsa source, available in GR3.
## Testing Done

Tested with different soundcards.

## Checklist

### DIFF
--- a/blocklib/audio/alsa_source/alsa_source.yml
+++ b/blocklib/audio/alsa_source/alsa_source.yml
@@ -1,0 +1,48 @@
+module: audio
+block: alsa_source
+label: ALSA Source
+blocktype: sync_block
+category: '[Core]/Audio'
+
+# Example Parameters
+parameters:
+-   id: sampling_rate
+    label: Sampling Rate
+    dtype: ru32
+    settable: false
+    default: 0
+    grc:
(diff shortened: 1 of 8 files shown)


mormj:
Re-running CI since #6428 was required to avoid recent changes in PMT lib

dl1ksv:
@mormj While trying to find out how to implement the test procedure, I noticed that the location of the test procedure differs.

For instance:

./blocklib/filter/fir_filter/qa_fir_filter.py
./blocklib/filter/test/qa_fir_filter.py

./blocklib/streamops/streams_to_vector/qa_streams_to_vector.py

./blocklib/streamops/test/qa_delay.py

So what's the preferred directory?

mormj:
If the tests correspond to a single block (or mainly to one block) then it should be in the block dir itself.  The meson autogeneration will pick up `qa_blockname.py` as a block test.  This was added later, so some tests ended up in the test/ directory and are manually added in test/meson.build

If it is a more complex scenario, then they should go in test/

mormj (approved):

Issue #6433, opened 2022-12-09
FEC Async Decoder overflows when converting from float to uint8_t


daniestevez:
### What happened?

I have noticed that the FEC Async Decoder block can overflow when doing a conversion from `float` to `uint8_t` if the amplitude of the input is too large. This conversion happens, for instance, if the block is used with a CC Decoder Definition to implement a Viterbi decoder.

These are the problematic conversions:
https://github.com/gnuradio/gnuradio/blob/6efdc9d84e6149e3f4049ee7c558dc01106dfd06/gr-fec/lib/async_decoder_impl.cc#L132

https://github.com/gnuradio/gnuradio/blob/6efdc9d84e6149e3f4049ee7c558dc01106dfd06/gr-fec/lib/async_decoder_impl.cc#L185

This is a potential problem when using this block to perform soft-input Viterbi decoding. If the amplitude of the input is too large for any reason, all the conversions to `uint8_t` will overflow, and the effect is that we will pass the inverted symbol stream to the Viterbi decoder. With a transparent convolutional code (most common codes), the Viterbi decoder will happily output the inverted output. This is how I first noticed the problem.

There are some commented lines in the code with calls to `volk_32f_s32f_convert_8i`. These date back to fbbde8d01ab43b7a3ffd5ca9d65cf50a84e2a7be when Tom Rondeau first implemented the block in 2014. The volk kernels for conversion to integers saturate instead of overflowing, but I don't think that the kernels can be used here because there are no kernels that convert to `uint8` (unless we want to jump through some hoops by using the `int8` kernel and adding the bias to convert to `uint8` later, which is not straightforward, because the bias is specified by `d_decoder->get_shift()` as a float, even though that shift would be `128.0` in most cases).

What I think we should do is to use the same code as in the Float to UChar block, since this is exactly the same problem. This block uses [float_array_to_uchar.cc](https://github.com/gnuradio/gnuradio/blob/main/gr-blocks/lib/float_array_to_uchar.cc), which unfortunately I think it is not accessible from gr-fec.

Perhaps we should add a `volk_32f_s32f_s32f_convert_8u` kernel to Volk that includes scale and bias parameters, even if at the moment this kernel would only contain a generic implementation (similar to float_array_to_uchar.cc, but including the multiplication by the scale and addition of the bias).

Thoughts?

### System Information

All systems are affected.

### GNU Radio Version

4.0-git (dev-4.0)

### Specific Version

0.0.0.0

[...]

PR #6445, opened 2022-12-21
fec: fix tagged decoder frame size for CC_TERMINATED

daniestevez:
## Description

This fixes a bug in the calculation of the frame size used by the tagged decoder that affects the CC decoder in terminated mode.

Currently, tagged_decoder calls the `set_frame_size()` method of the decoder with a size computed as the input size multiplied by the rate.

In the cc_decoder, the frame_size parameter of the `set_frame_size()` method is used to calculate and set `d_veclen`. The value of `d_veclen` is used to determine how many forward steps (butterflies) the Viterbi algorithm should be run. The number of input items that are read in `d_veclen` forward steps is `d_veclen * d_rate` (note that `d_rate` is actually the inverse of the rate; typically, `d_rate == 2`). Therefore, the size of the input buffer should be as large as this value, and in most cases it should be equal, because we want the cc_decoder to look at all the provided input.

The `d_veclen` variable is related to `d_frame_size`, which indicates how many output bits are produced by the decoder. The value of `d_frame_size` is equal to the parameter of `set_frame_size()` in all cases except in terminated mode with byte padding, in which case the byte padding is subtracted when computing `d_frame_size`. The value of `d_veclen` is equal to `d_frame_size + d_k - 1` in terminated and streaming modes, and to `d_frame_size` in truncated mode.

The problem with the current code is that in tagged_decoder the `set_frame_size()` method is called with `ninput_items[0] * rate`. In the streaming and terminated modes this will cause `d_veclen` to be too large. The forward steps (butterflies) of the Viterbi algorithm will read past the end of the input buffer. This is undefined behaviour and can cause some bit errors at the end of the decoded codeword (the effect of reading past the end of the buffer is essentially extending the codeword with 12 symbols of garbage).

This problem does not happen in the async_decoder because there is some code that computes a "diff", which works out to be
```
d_k - 1 + d_padding * rate()
```
in terminated mode and 0 in all other cases. The diff is subtracted from the frame size. This subtraction gives the correct `d_veclen` calculation, so that the butterflies do not read past the input end.

This commit fixes the tagged_decoder when used with the terminated CC decoder by adding the same code as in the async_decoder to calculate and subtract this "diff" value.

Some caveats:

1) This commit changes the size of the output packets of tagged_decoder when used with a terminated CC decoder. Formerly, the output packets were 6 bits longer, as if the tail bits were included in the output. However, having these 6 extra bits was just an artifact of looking 12 symbols past the end of the input. The `cc_decoder_impl::chainback_viterbi()` function is not prepared to extract tail bits in the terminated case. This chainback extracts bits "as they exit the shift register on the right", so tail bits cannot be extracted, because they "never exit the shift register". (Here I have in mind the usual depiction of a convolutional encoder, where input bits are shifted from left to right into a shift register).

The CC_TRUNCATED case in `cc_decoder_impl::generic_work()` has some additional code to extract additional bits which haven't "exited the shift register" yet, because it is mandatory to extract them in this case (they are the last 6 bits of the message).

It think there is no good use case for attempting to extract the tail bits in the terminated case, so this modification seems a good way forward. The new behaviour also matches that of async_decoder, which does not output tail bits.

However, this breaks tag propagation of the tagged_decoder when used with the terminated CC decoder, because the block calls `set_relative_rate()` with the nominal rate (1/2), which is slightly different from the true rate once we take into account the fact that tail bits are dropped. This could be fixed by propagating tags manually in the `work()` function.

2) The streaming mode is still broken, both when used with the async_decoder and with the tagged_decoder. Since for the streaming case diff is 0 but `d_veclen` includes the term `d_k - 1`, the streaming mode reads past the input buffer. It isn't really possible to run the streaming mode properly with the async_decoder or tagged_decoder (it should be run with the decoder_impl.cc), because the streaming mode requires history (some overlapping input items between consecutive calls), which is not available with the async_decoder or tagged_decoder.

3) The code copied from async_decoder has a "watch out" comment stating that it might be over-specializing for the CC decoder in terminated mode. It is true that this code seems specially crafted to cover this very specific case, but as far as I have been able to think, it does not affect negatively any other cases (taking into account other modes of the CC decoder and other FEC decoders).

---

I'm specially interested in feedback regarding point 1 above. If we think that this is a good way forward, then I can implement the manual tag propagation. If we think we don't want to change the output size and still have the 6 tail bits in the output, a different approach would be needed to fix the bug.

Affected blocks in gr-fec: Tagged Decoder and CC Decoder.

## Testing Done

I could provide an ad-hoc test flowgraph that shows the problem if needed.

## Checklist

- [x] ~I have updated [the documentation](https://wiki.gnuradio.org/index.php/Main_Page#Documentation) where necessary.~
- [x] ~I have added tests to cover my changes,~ and all previous tests pass.

### DIFF
--- a/gr-fec/lib/tagged_decoder_impl.cc
+++ b/gr-fec/lib/tagged_decoder_impl.cc
@@ -51,7 +51,9 @@ int tagged_decoder_impl::calculate_output_stream_length(const gr_vector_int& nin
     if ((ninput_items[0] * d_decoder->rate()) > (d_mtu * 8)) {
         throw std::runtime_error("tagged_encoder: received frame is larger than MTU.");
     }
-    d_decoder->set_frame_size(round(ninput_items[0] * d_decoder->rate()));
+    int diff =
+        d_decoder->rate() * d_decoder->get_input_size() - d_decoder->get_output_size();
+    d_decoder->set_frame_size(round(ninput_items[0] * d_decoder->rate()) - diff);
     return d_decoder->get_output_size();
 }
 


gr-fec/lib/tagged_decoder_impl.cc
```diff
@@ -51,7 +51,9 @@ int tagged_decoder_impl::calculate_output_stream_length(const gr_vector_int& nin
     if ((ninput_items[0] * d_decoder->rate()) > (d_mtu * 8)) {
         throw std::runtime_error("tagged_encoder: received frame is larger than MTU.");
     }
-    d_decoder->set_frame_size(round(ninput_items[0] * d_decoder->rate()));
+    int diff =
+        d_decoder->rate() * d_decoder->get_input_size() - d_decoder->get_output_size();
+    d_decoder->set_frame_size(round(ninput_items[0] * d_decoder->rate()) - diff);
```
jdemel, line 56: This is a change in the wrapper. Not the CC decoder itself. Other decoders might return `false` in case the `set_frame_size` method is called.

Maybe the decoder should compute this kind of info internally since it is only required in the CC case but LDPC, polar etc. don't need it.

Or, we need a second method that sets the input size rather than the output size. For the CC case that would probably make more sense. output size == frame size.

daniestevez: This kind of code also appears on `async_decoder`, which is also the wrapper, not the CC decoder:

https://github.com/gnuradio/gnuradio/blob/7f1a32432fd45924a9eb5f3e3cc99d3248b740f7/gr-fec/lib/async_decoder_impl.cc#L86-L93

So I think that whatever solution we deem appropriate here (such as adding a method for the input size) would also need to be applied to the `async_decoder`.

Reviewing the code of the different wrappers and the implementation of the CC decoder, I agree that there are several parts that would improve from a refactor.

Since currently `tagged_decoder` is broken and I think that `async_decoder` is not broken, I think there are two possible ways to go:
1. Merge this PR, keep the idea to improve the code quality in the future.
2. Don't merge this PR. Wait until a refactor/improvement of the code quality that also fixes this problem as a by-product.

I prefer approach 1, because it would fix the problem described in this PR right away and it doesn't get in the way of other refactors/improvements happening afterwards. With approach 2, it might be a long time until this particular problem gets fixed, because the best way to improve the code is not clear at the moment.

The only potential concern that I have about merging this PR is whether it breaks something else. I think it doesn't, but I haven't reviewed the other decoders to make sure that this change is harmless for them.

jdemel: The second approach is effectively an API change. This would require a new release. I'd prefer the second approach but I agree that your approach is the better one for now.

The FECAPI needs some refactoring. i.e. a better understanding of codes and properties. Currently, the abstraction is broken.

jdemel (approved):
After a short discussion, this PR LGTM. However, we need to fix the FECAPI.

daniestevez:
Thanks! Agreed that FECAPI needs a clean-up. Over the last weeks I've been giving more use to some FECAPI features I don't use normally and I've found this bug, #6433, and some problems with the LDPC decoder that I think @VladislavF is looking at in more detail and will report as an issue (quick summary: I was unable to get the LDPC decoder with an alist describing the r=1/2, k=1024 CCSDS code; If I gave the decoder a non-zero codeword with a single bit error, the decoder would output the zero codeword).

PR #6446, opened 2022-12-22
grc: snippets before init is called

mormj:
## Description
There are situations where snippets need to be inserted before block initialization, such as when 
a block constructor makes contact with an external server, it might be handy to initialize the server
or check whether it is ready prior to calling the flowgraph init function that calls the block constructor

The problem with the current construct is snippets can only be injected after the flowgraph is initialized,
since they use `self` as a reference to the flowgraph object.  There is probably a better way to do this
but the easy solution is to pass `None` as what will become self, and snippets that are inserted before
main are not able to make calls to `self`

Notes:
1. Main Before Init is placed second on the list, so Main After Init is the default

I'm wondering if it would be more useful to have the injection point inside the `top_block.__init__` after variables 
are instantiated, then the snippet would have access to variables

## Related Issue
None

Just GRC when snippets are used

## Checklist

### DIFF
--- a/grc/blocks/snippet.block.yml
+++ b/grc/blocks/snippet.block.yml
@@ -6,8 +6,8 @@ parameters:
 -   id: section
     label: Section of Flowgraph
     dtype: string
-    options: ['main_after_init', 'main_after_start', 'main_after_stop' ]
-    option_labels: ['Main - After Init', 'Main - After Start', 'Main - After Stop']
+    options: ['main_after_init', 'main_after_start', 'main_after_stop', 'init_before_blocks' ]
+    option_labels: ['Main - After Init', 'Main - After Start', 'Main - After Stop', 'Init - Before Blocks']
 -   id: priority
     label: Priority
     dtype: int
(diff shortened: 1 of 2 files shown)


mormj:
Perhaps just doing the second commit here is cleaner, which looks like:

image


```
        # Blocks
        snippets_init_before_blocks(self)
        self.blocks_throttle_0 = blocks.throttle(gr.sizeof_gr_complex*1, samp_rate,True)
        self.blocks_null_source_0 = blocks.null_source(gr.sizeof_gr_complex*1)
        self.blocks_null_sink_0 = blocks.null_sink(gr.sizeof_gr_complex*1)
```

willcode:
There's no harm in adding more snippet hooks. There's got to be a better way to insert more complicated user code, though.

In the short term, we could have a block where the user can fill in more than one hook, so things stay more coherent, and keep compatibility.

mormj:
In the short term I think the second commit is safer than the first, and maybe also should have one "init before connections".  But you're right there's got to be a better way to bring it together

willcode (approved):

PR #6450, opened 2022-12-27
pdu: Don't enable blocks and filter

cjmayo:
Not required:
e36f2fab5 ("pdu: Remove unused dependencies", 2022-06-01)

Signed-off-by: Chris Mayo <aklhfex@gmail.com>

## Testing Done
Build 3.10.5.0 with this patch.

## Checklist

### DIFF
--- a/gr-pdu/CMakeLists.txt
+++ b/gr-pdu/CMakeLists.txt
@@ -13,8 +13,7 @@
 # Register component
 ########################################################################
 include(GrComponent)
-gr_register_component("gr-pdu" ENABLE_GR_PDU ENABLE_GNURADIO_RUNTIME ENABLE_GR_BLOCKS
-                      ENABLE_GR_FILTER)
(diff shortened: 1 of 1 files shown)


willcode (approved):
I don't see anything in `gr-pdu` that requires `ENABLE_GR_FILTER`. There is a PDU Filter block, but it filters PDUs, not samples.

ryanvolz (approved):

willcode:
Is `GR_BLOCKS` required by the QA code and examples (it's used there), or do we assume that QA and running of examples is being done on a system where everything is installed? The latter seems reasonable to me.

ryanvolz:
> Is `GR_BLOCKS` required by the QA code and examples (it's used there), or do we assume that QA and running of examples is being done on a system where everything is installed? The latter seems reasonable to me.

Agreed, I think this check should be for what's required to build gr-pdu, which it will be after this change. It's reasonable to have separate requirements for QA and examples, and almost nobody is going to build without gr-blocks anyway.

willcode:
Merging here, but holding backport for a bit, just to make sure ...

PR #6451, opened 2022-12-31
grc: fix debug logging and GTK import order

sdh11:
## Description
Replacement PR for #5220.

This fixes the logger prefix which can change depending on how GRC is launched. This also fixes an issue with importing GTK if multiple versions are installed.

## Related Issue
Fixes the logger prefix in grc (#4628)

## Checklist

### DIFF
--- a/grc/main.py
+++ b/grc/main.py
@@ -41,8 +41,15 @@ def main():
     args = parser.parse_args()
 
     # Enable logging
-    # Note: All other modules need to use the 'grc.<module>' convention
-    log = logging.getLogger('grc')
+    # Note: All other modules need to use the same '<prefix>.<module>' convention.
+    # This should automatically be handled by logging.getLogger(__name__), but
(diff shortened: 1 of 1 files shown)


willcode:
I haven't tested, but it looks like this would fix a `ValueError: Namespace Gtk is already loaded with version 4.0` seen when trying to run `grc` from a source directory, in addition to the logger name change.

willcode (approved):
Just for clarity, something other than `__name__[:-5]` would be good. Otherwise, LGTM.

mormj (approved):

willcode:
@sdh11 did you want to implement `rindex` or something similar? `replace` would almost work, but someone could have a submodule that matches `.main`. Regex would be overkill, so `rindex` and a comment would be good enough.

sdh11:
Yes, I will. I haven't had a chance to do that yet.

sdh11:
@willcode: I've pushed the new commit with the fix

PR #6454, opened 2023-01-05
grc: add a pass if snippet is empty

### LINKED ISSUE #6444: Empty Python Snippet causes invalid rendered python
### What happened?

If I plop down a snippet block, but don't put any code in the block, the rendered python looks like:

```
def snipfcn_snippet_0(self):


def snippets_main_after_init(tb):
    snipfcn_snippet_0(tb)
```

Which will cause a python parsing error.  

GRC should have the logic to put a `pass` in there, or not add the snippet if it is empty

### System Information

OS: (Linux Distro, Mac, Windows)
GR Installation Method: (native distro packages - apt, rpm, etc., Conda, Source

### GNU Radio Version

3.11-git (main)

### Specific Version

0.0.0.0

Drop an empty python snippet block in GRC and render the flowgraph

### Relevant log output

_No response_

mormj:

## Description
See related [issue](https://github.com/gnuradio/gnuradio/issues/6444)

An empty snippet prior to this PR will cause an invalid rendered flowgraph 

## Related Issue
Fixes #6444 

GRC files with snippets

## Testing Done
empty and non-empty snippet. 

## Checklist

### DIFF
--- a/grc/core/FlowGraph.py
+++ b/grc/core/FlowGraph.py
@@ -115,8 +115,11 @@ class FlowGraph(Element):
             d['lines'] = snip.params['code'].value.splitlines()
             d['def'] = 'def snipfcn_{}(self):'.format(snip.name)
             d['call'] = 'snipfcn_{}(tb)'.format(snip.name)
-            if not section or sect == section:
-                output.append(d)
+            if not len(d['lines']):
+                Messages.send_warning("Ignoring empty snippet from canvas")
+            else:
+                if not section or sect == section:
+                    output.append(d)
 
         # Sort by descending priority
         if section:


grc/core/FlowGraph.py
```diff
@@ -113,6 +113,8 @@ def get_snippets_dict(self, section=None):
             d['section'] = sect
             d['priority'] = snip.params['priority'].value
             d['lines'] = snip.params['code'].value.splitlines()
+            if not len(d['lines']):
```
willcode, line 116: Same problem if there's just a comment. No problem at all if there's just a literal. All kinds of ways to make bad code with a snippet. Would a warning somewhere be better than silently ignoring? Can't think of a best way. In any case, this gets rid of the most obvious case.

mormj: Good point on the warning - silently failing is not usually a good thing

willcode (approved):

PR #6457, opened 2023-01-06
fec: fix typo in upper_convolve()

### LINKED ISSUE #5430: Issues in polar code module
After some test of polar code module (mainly using the polar_code_example.grc),I found some problem, maybe they are bugs or not.  
1. I can't set the channel type to AWGN, or it will give me a error:(only show a part of tracking message here)
```
File "/usr/lib/python3/dist-packages/gnuradio/fec/polar/channel_construction_awgn.py", line 197, in upper_convolve
    q[0, idx] = (tpm[0 / i] ** 2 + tpm[1, i] ** 2, 2)
ZeroDivisionError: division by zero
```
2. When I testing the encoder and decoder(SC and SCL)(using the original setting in the example(polar_code_example.grc)), I found that I can't decoding the message correctly, even the amplitude of noise have set to 0(meant no noise in transmitting). BUT, if I replace the encoding(and decoding) module from POLAR to LDPC, this example can give the correct output(by viewing the GUI output of Coded window).<br>
I'm not a researcher in this field, I just want to use the module to finish my own project. So.....sorry, I just know I have some problem. But I can't give a suggestion or some worked solutions for them.
Thank you for your notice of this issue.

willcode:

## Description
Type in `upper_convolve` was generating /0 error, and doesn't look like the intended code. Note that I don't know anything about polar codes!

## Related Issue
Fixes #5430 

## Testing Done
None

## Checklist

### DIFF
--- a/gr-fec/python/fec/polar/channel_construction_awgn.py
+++ b/gr-fec/python/fec/polar/channel_construction_awgn.py
@@ -198,7 +198,7 @@ def upper_convolve(tpm, mu):
     idx = -1
     for i in range(mu):
         idx += 1
-        q[0, idx] = (tpm[0 / i] ** 2 + tpm[1, i] ** 2, 2)
+        q[0, idx] = (tpm[0, i] ** 2 + tpm[1, i] ** 2, 2)
         q[1, idx] = tpm[0, i] * tpm[1, i]
         for j in range(i + 1, mu):
             idx += 1
@@ -219,8 +219,8 @@ def lower_convolve(tpm, mu):
     idx = -1
     for i in range(0, mu):
         idx += 1
-        q[0, idx] = (tpm[0 / i] ** 2, 2)
-        q[1, idx] = (tpm[1 / i] ** 2, 2)
+        q[0, idx] = (tpm[0, i] ** 2, 2)
+        q[1, idx] = (tpm[1, i] ** 2, 2)
         if q[0, idx] < q[1, idx]:
             q[0, idx], q[1, idx] = swap_values(q[0, idx], q[1, idx])
         idx += 1


gr-fec/python/fec/polar/channel_construction_awgn.py
```diff
@@ -198,7 +198,7 @@ def upper_convolve(tpm, mu):
     idx = -1
     for i in range(mu):
         idx += 1
-        q[0, idx] = (tpm[0 / i] ** 2 + tpm[1, i] ** 2, 2)
+        q[0, idx] = (tpm[0, i] ** 2 + tpm[1, i] ** 2, 2)
```
jdemel, line 201: This might look like a simple change but it is part of this whole algorithm.

I wrote this code 7 years ago based on the paper "How to construct polar codes". I tried to validate these results by comparing them to the implementation I had available from Ido Tal. Everything looked fine back then.

For small configurations, e.g. 32 or 64, this should produce the same reliability order as other implementations, e.g. the Bhattacharyya Bounds method.

This needs more investigation. Maybe the referenced papers help.

jdemel (commented):
This would need more investigation.

willcode:
@jdemel this was changed here, after your initial work, and I don't understand why these changes would have been made https://github.com/gnuradio/gnuradio/commit/9e625c4821f4c63421b3d3747c0c4f358fef6c5f. This code must not have been used since 2017.

jdemel:
@willcode Thanks for the hint at https://github.com/gnuradio/gnuradio/commit/9e625c4821f4c63421b3d3747c0c4f358fef6c5f#diff-fe39d3d66240b9d5f30547441eea8bda6b412612f17a59524cc0b58e0d54cd78L194

In this case, the "old" or "original" implementation should be restored. It looks a lot like smth broke.

jdemel (approved):
LGTM. Thanks for the discussion!

Issue #6459, opened 2023-01-10
DTV dvbs2_modulator_bc_impl.cc


FranBoric:
I noticed that one of the constants is actually different than then in DVB-S2X standard. 

It is coded here.
https://github.com/gnuradio/gnuradio/blob/0becf06a397edc1b2dcf440785e40b73f76433ad/gr-dtv/lib/dvbs2/dvbs2_modulator_bc_impl.cc#L500

In this implementation of MOD_4_8_4_16APSK, case C140_180, it says that 
r2 = r1 * 2.6 
and it should be 
r2 = r1 * 2.8 
and as such it produces quite wrong result!

The standard referenced is
https://www.etsi.org/deliver/etsi_en/302300_302399/30230702/01.01.01_20/en_30230702v010101a.pdf

willcode:
Fixed in #6460

PR #6460, opened 2023-01-10
Bugfix dvbs2_modulator_bc_impl.cc

FranBoric:
I have described the issue here

https://github.com/gnuradio/gnuradio/issues/6459

The problem was a constant that is different in standard.


drmpeg:
Good catch. However, you have to sign your commit. Also, a more descriptive commit message would be nice. Something like:

gr-dtv: Fix typo in DVB-S2 modulator block for 4+8+4+16APSK constellation

You can do both with ```git commit -s --amend``` and then ```git push --force```.

FranBoric:
Thank you, sorry, this was my first!

mormj:
@FranBoric - looks like the commit still needs to be signed via the instructions from @drmpeg

drmpeg (approved):
The change is correct. I was just hoping for a better commit message.

mormj:
I can fix the commit message with a squash, but can't merge without a DCO check - @FranBoric - can you please update and sign your commit?

drmpeg:
The SOB seems to be there. Is the CI just missng it? Here's what it looks like to me:

image

mormj:
you're right, this does seem to be a CI issue.  perhaps the email in the signoff doesn't match that in the github account.

willcode:
Merged - signoff email should be of the user's choice.

PR #6462, opened 2023-01-13
runtime: (Re)make block buffer size customizable

theartful:
## Description
Make block buffer size customizable using the environment variable `GR_CONF_DEFAULT_BUFFER_SIZE`

Affects maximum allowable buffer size for all blocks.

## Testing Done
I was testing a fft filter with a lot of taps (>100) which required a huge buffer size, and gnuradio complained. Changing the environment variable `GR_CONF_DEFAULT_BUFFER_SIZE` in previous versions of gnuradio made this possible, and now it is possible again after this change.

## Checklist

### DIFF
--- a/gnuradio-runtime/lib/block.cc
+++ b/gnuradio-runtime/lib/block.cc
@@ -27,7 +27,9 @@ namespace gr {
 // 32Kbyte buffer size between blocks
 #define GR_FIXED_BUFFER_SIZE (32 * (1L << 10))
 
-static const unsigned int s_fixed_buffer_size = GR_FIXED_BUFFER_SIZE;
+static const unsigned int s_fixed_buffer_size =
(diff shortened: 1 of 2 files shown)


theartful:
forced push to change the commit message to comply with the style followed

willcode:
This may not be compatible with custom buffers. @dsorber @mormj ?

mormj:
I don't think it should affect custom buffers, but there are some compile flags where we can run the tests with custom buffers forced on.  I'll kick those off locally.  It's not clear to me why s_fixed_buffer_size would have been defined in flat_flowgraph at all

willcode:
Prior to the rework, the buffer size calculation was done in `flat_flowgraph.cc`. That logic was moved to `block.cc` but reading of the setting was left behind (and has not effect). So, this looks right.

willcode (approved):

mormj (approved):

Issue #6452, opened 2023-01-04
GRC infinite loop and crash: "TypeError: Context.rectangle() takes exactly 4 arguments (0 given)"


paul-k-young:
### What happened?

GRC froze when I tried to add the OOT module "QT GUI Inspector Sink" to a new flowgraph.  Additionally the console was flooded with the following messages.

Traceback (most recent call last):
  File "gnuradio\grc\gui\DrawingArea.py", line 223, in draw
    self._flow_graph.draw(cr)
  File "gnuradio\grc\gui\canvas\flowgraph.py", line 535, in draw
    draw_element(cr)
  File "gnuradio\grc\gui\canvas\block.py", line 280, in draw
    port.draw(cr)
  File "gnuradio\grc\gui\canvas\port.py", line 134, in draw
    cr.rectangle(*self._area)
TypeError: Context.rectangle() takes exactly 4 arguments (0 given)

I didn't see any errors related to the Inspector Sink, just the above errors from GRC.

What I expected:

The component to be added to the flowgraph.  Or, if that's not possible, a graceful failure.  Bonus for a sensible error message :)

### System Information

OS: Windows 11
GR Installation Method: Conda

### GNU Radio Version

3.10 (maint-3.10)

### Specific Version

3.10.4.0

### Steps to Reproduce the Problem

Steps to reproduce:
1) Open GNU Radio Companion
2) Select File > New > QT Gui
3) Search for "inspector" in the components list.
4) Drag "GUI > QT GUI Inspector Sink" onto the flowgraph.

### Relevant log output

```shell
Traceback (most recent call last):
  File "C:\Users\paulk\radioconda\lib\site-packages\gnuradio\grc\gui\DrawingArea.py", line 223, in draw
    self._flow_graph.draw(cr)
  File "C:\Users\paulk\radioconda\lib\site-packages\gnuradio\grc\gui\canvas\flowgraph.py", line 535, in draw
    draw_element(cr)
  File "C:\Users\paulk\radioconda\lib\site-packages\gnuradio\grc\gui\canvas\block.py", line 280, in draw
    port.draw(cr)
  File "C:\Users\paulk\radioconda\lib\site-packages\gnuradio\grc\gui\canvas\port.py", line 134, in draw
    cr.rectangle(*self._area)
TypeError: Context.rectangle() takes exactly 4 arguments (0 given)
```

willcode:
[...]

PR #6465, opened 2023-01-14
grc: Fixes a type issue in the parameters module

paul-k-young:
Fixes the truncation function to handle integer values gracefully.

Fixes gnuradio/gnuradio#6452

Signed-off-by: Paul K Young <paul.k.young@gmail.com>

## Checklist

### DIFF
--- a/grc/gui/canvas/param.py
+++ b/grc/gui/canvas/param.py
@@ -91,6 +91,7 @@ class Param(CoreParam):
     ##################################################
 
     def truncate(self, string, style=0):
+        string = str(string)
         max_len = max(27 - len(self.name), 3)
         if len(string) > max_len:
             if style < 0:  # Front truncate


willcode:
Should the cast happen wherever `get_value()` is called and the value is assumed to be a string? If this is failing in `truncate()` it is probably a symptom of something wrong at a higher level.

paul-k-young:
> If this is failing in `truncate()` it is probably a symptom of something wrong at a higher level.

Agreed.  get_value and set_value are defined in the parent class grc.core.params.Param. The setter properly stringifies the input,  however, there's no attempt to restrict visibility of the underlying value (e.g. by calling it self._value).  Even the parent class doesn't use the setter.

A proper fix would be to change the variable name in the parent class to _value and update all code references to use the getter/setter.  I have no idea how involved that would be, or whether there would be a simpler, 'good enough' fix.

mormj:
Closing this for now - if you find a more robust fix, please reopen.

PR #6474, opened 2023-01-17
qtgui: fix instrumentation sink save dialog

### LINKED ISSUE #5991: QT Instrumentation sink save
### What happened?

When I middle click in a QT Instrumentation sink plot (Time or Frequency) and select save a save file dialog appear. I can give the file a name and navigate to the desired directory. The problem is that there is no save button, but a disabled open button in stead. I can work around this issue by selecting an existing file to overwrite.

### System Information

OS: Arch Linux
save_existing
save
.

### GNU Radio Version

3.11-git (main)

### Specific Version

v3.11.0.0git-184-gf9c41f17

Make s simple flowgraph in gnuradio-companion containing a QT instrumentation sink. Run the flowgraph. Middle click in the QT instrumentation sink plot and select Save. 

### Relevant log output

_No response_

R-ohit-B-isht:
## Description

When middle click in a QT Instrumentation sink plot (Time or Frequency) is pressed and select save a save file dialog appear. We can give the file a name and navigate to the desired directory. The problem is that there is no save button, but a disabled open button instead. 
Added a save  option in place of open.
## Related Issue
Fixes #5991 

gr-qtgui/lib/displayform.cc - Line 303-324
gr-qtgui/lib/eyedisplaysform.cc - Line 339-360
gr-qtgui/lib/numberdisplayform.cc - Line 185-207

## Testing Done
Yes

## Checklist

### DIFF
--- a/gr-qtgui/lib/displayform.cc
+++ b/gr-qtgui/lib/displayform.cc
@@ -303,6 +303,8 @@ void DisplayForm::saveFigure()
     QString filename, filetype;
     QFileDialog* filebox = new QFileDialog(0, "Save Image", "./", types);
     filebox->setViewMode(QFileDialog::Detail);
+    filebox->setAcceptMode(QFileDialog::AcceptSave);
+    filebox->setFileMode(QFileDialog::AnyFile);
     if (filebox->exec()) {
         filename = filebox->selectedFiles()[0];
         filetype = filebox->selectedNameFilter();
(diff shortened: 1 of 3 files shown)


willcode:
@R-ohit-B-isht an extra file `gr-qtgui/python/qtgui/CMakeLists.txt.user` got checked in.

R-ohit-B-isht:
> @R-ohit-B-isht an extra file `gr-qtgui/python/qtgui/CMakeLists.txt.user` got checked in.

My mistake!

willcode:
@R-ohit-B-isht This still contains `CMakeLists.txt.user`.

Automatic addition of file extensions: this seems like a good idea in general, but if the user selects `out.jpg`, won't the output file end up being `out.jpg.jpg`? I'd be interested in comments on how this should work.

R-ohit-B-isht:
> @R-ohit-B-isht This still contains `CMakeLists.txt.user`.
> 
> Automatic addition of file extensions: this seems like a good idea in general, but if the user selects `out.jpg`, won't the output file end up being `out.jpg.jpg`? I'd be interested in comments on how this should work.

Well I removed it in the latest push and its not showing in my repo as well , I'll squish the changes to remove it's existence.

Saving it without adding .jpg or other extension made it invisible to the filter and did not showed the earlier saved files,i.e. if a saved  one as "abc" (actually a jpeg) and then took another shot the file save dialog did not display any previous saved jpeg files on selecting the filter to jpeg files (which could be inconvenient if I accidentally replaced any previous screenshot with same name )
I checked the saved file in terminal it showed name as abc.jpg not abc.jpg.jpg , I think JPEG is only set as property without extension. Although I checked it in linux only don't know if it shows .jpg.jpg in any other system

Issue #6456, opened 2023-01-06
Constellation Encoder, Decoder, and Soft Decoder ignore symbol mapping


VladislavF:
### What happened?

This issue might actually be better stated as only the constellation modulator and receiver respect the symbol mapping defined in the constellation object. 
Essentially for any arbitrary constellation, only the trivial or linear symbol map is guaranteed to work. If you  want to create a grey-coded mapping for instance, the constellation encoder and decoders will not apply the mapping. However the constellation modulator and receiver do in fact respect the mapping. This is very confusing if you've chosen to say manually receive the constellation with discrete RRC filters and sync using the decoder or soft decoder resulting in the incorrect symbols being output even though they were in fact permuted correctly by the transmitter. Worse is if you use both the encoder and decoder, then the symbol map is never applied, and it's not obvious unless you check which symbols are actually sent for which bits.
I've discussed this at some length with @argilo 
In the case of grey-coding, this can be worked around by using the Map block or even permuting the constellation points via something like the following: [points[n ^ n >> 1] for n in range(len(points))]
But this is not proper since mappings can be arbitrary and as they are used by some blocks (Mod&RX), it would make sense for all constellation blocks to respect the definition the same way. This seems to be an issue with the blocks in question (Decode,Soft Decode,Encode) not applying the pre_diff_code() function (please correct me if I'm wrong, recalling from my discussion with Clayton). 

### System Information

OS: WSL
GR Installation Method: pybombs -R gnuradio-default

### GNU Radio Version

3.10 (maint-3.10)

### Specific Version

3.10.4

### Steps to Reproduce the Problem

Connect a flowgraph as such: constellation encoder->constellation decoder.
Observe: The symbol mapping is not applied by either block (assumes mapping: range(len(constellation)))

See my crude screenshot of the output from a flowgraph as such: vector source->constellation mod.->delay->RRC->constellation decode->GUI time sink

![image](https://user-images.githubusercontent.com/9085479/211106179-eaaf74db-0428-4467-8b96-e29aa774a4f6.png)

VladislavF:
Actually It looks like the constellation soft decoder may also be handling constellations with higher orders incorrectly. Currently investigating.

VladislavF:
The issue in my case was not with the constellation but the automatically generated soft decision LUT using the argument 'auto'.
[...]

PR #6478, opened 2023-01-19
fec: LDPC and Soft Decision Fixes

VladislavF:
## Description
This PR seeks to fix some critical bugs and oversights in the gr-fec LDPC code and the soft decision LUT calculations. This is very much a WIP and born of a desire to share my findings with the community. 

Essentially the issues encountered that caused me to open the PR are in two parts.
First, the soft decision LUT. ~~The only way I have found to give a working LUT to a constellation object is using the "digital.soft_dec_table()" function, hence this is what is addressed in my PR.~~ The issue here is that the calculations for the LUT are incorrect/have errors. I have fixed two things here. The first was normalizing the constellation to unit power so that the 'nwpr' argument has a simpler definition as the denominator in S/N where S=1. This allows consistent and accurate SNR based calculations of the LUT for any arbitrary constellation without having to perform manual normalizations. The second change is in the actual point distance calculations. I have modified the expressions to match equation 1. in the DVB-S2 soft decoding paper: https://ietresearch.onlinelibrary.wiley.com/doi/epdf/10.1049/iet-com.2013.0307
Thus point distances are now optimally calculated and 'npwr' can be set based on expected channel SNR given the sent constellation was power normalized. (perhaps this could be an argument?)
I have also updated the LUT to be generated with a "reasonable" 2x padding of the largest point in the normalized constellation: maximum_amp=max(max(real(const_points)), max(imag(const_points)))
This will allow for spreading of the outer points as noise increases without clipping.

The second issue was with the LDPC decoder. The current BP decoder has a habit of converging to zeros under various conditions including, but not limited to: LLR's too high, sigma too low, and apparently some codewords. This has to do with very large and very small numbers propagating through the variables as NaN and Inf eventually poisoning the entire block. Also this block for some reason chooses to linearize LLR's with a multiplicative constant 'sigma' parameter which does not make sense for soft LLR inputs. Since I could not find a detailed source on the algorithm besides a handful of expressions, I decided to re-implement a different one in the same framework based on equations 2. and 3. in the following paper: https://citeseerx.ist.psu.edu/document?repid=rep1&type=pdf&doi=ccb90917786903efded5a5536d67fdd1440929c0
Reduced-Complexity Decoding of LDPC Codes by Jinghu Chen et al.
This has some issues which I will touch on, but the advantage here is that the zero output issue is resolved, the code is based on a readily accessible and documented algorithm, the tanh/atanh can eventually be turned into a performant LUT, and lastly, the gr-fec structure is preserved. Lastly I also added some import fixes (otherwise totally broken) to the LDPC codeword generation script which is very important to prototyping with codes of arbitrary rate and length

## Related Issue
This partially addresses some of my findings that caused the opening of this issue, however it actually does not solve the main mapping issue.
https://github.com/gnuradio/gnuradio/issues/6456

The following blocks have been modified:

-  Constellation Object definition
-  LDPC Decoder
- Constellation soft decoder

## Testing Done
So far I have performed various BER tests with multiple included and self defined constellation objects spanning BPSK through 16QAM and 16APSK. I have also done extensive testing with multiple included and generated LDPC codes of various rates and lengths. I have found that my implementation of the "tanh rule" algorithm lags the existing "sum-product" implementation by about <0.5dB (when it actually produced nonzero outputs). I have corroborated findings of the original bugs with @daniestevez. (the BER performance hit may be resolved as of ffce18c12a3b240e4bfaf6fcd2c740b891c8f8de)

## Summary of Changes

- LDPC sum-product algorithm replaced with tanh rule
- Python constellation LUT generator expressions corrected
- LDPC parity check matrix generator patched for python3
- Constellation object has Noise power argument
- Constellation soft decoder has noise power argument
- Constellation object LUT and soft decision CPP fixed

## Known Issues and Next Steps

- The LDPC Decoder is a bit slower now because std::tanh and std::atanh are implemented inefficiently/expensively.
- The LDPC "tanh rule" requires clamping of 'x' in tanh(x) to improve BER performance. This also helps to avoid more NaN and Inf. issues and should be investigated since the 0.5dB performance gap also likely lies here.
- LDPC block and definitions need to be updated (removal of sigma parameter, perhaps add a switch for both algorithms)
- constellation soft decoder needs a callback for 'npwr'
- Bit depths in range[9-11] (as tested) don't generate a proper LUT on 'auto'. works fine in python.

## Discussion
Again, I'm really opening this issue in the hopes that this gets more attention and that perhaps I can get some help to actually get this into a mergeable state. I would very much like for the GR LDPC code and soft decoders to be useful again. I am aware that a bunch of commits still need to be squashed or even potentially split into a separate PR. for instance, maybe the soft decision LUT and LDPC should be separate PRs, but for now I just want to get my code out there since my work requires that I move onto other problems for the time being.

## Checklist


daniestevez (commented):
Thanks for sharing this. I've read through the code and written a few comments. I think one of them explains what is happening with the `tanh` clamping.

In addition to what Vlad has said comparing the current implementation of the LDCP soft decoder and the implementation in this PR, I will add that the main difference is that the current implementation is based on likelihood ratios instead of log-likelihood ratios. Mathematically the two approaches are equivalent, but my understanding is that in practice log-likelihood ratios are used always because they tend to transform expressions involving products into expressions involving sums, and because they are better suited to working with floats or integers (because likelihood ratios of symbols for which we are very confident tend to very small or very large positive numbers, while with log-likelihood ratios we have the logarithm to help us avoid this growth).

I've been comparing Vlad's code with my go to text for LDPC, which is [the book by Sarah Johnson](https://www.cambridge.org/core/books/iterative-error-correction/7964EAB1B743D56D2E9B58CCA4CE71EB). In Algorithm 2.3 the book gives the sum-product algorithm written with log-likelihood ratios. It seems identical to Vlad's code, except for the way the `tanh` calculations are written (more on this below). I'm a bit rusty on my LDPC decoding theory, but if I remember correctly, the sum-product algorithm was the "best" one, and all the others that use slightly different update rules are faster variants that make some compromises or approximations, trading sensitivity for speed. So it bugs me that I don't understand why this code seems to perform 0.5dB worse than the current one.

A few pages after Algorithm 2.3 (on page 68), this book goes to derive the formula that Vlad is using for the `tanh` calculations, where the signs are taken out and multiplied together. However, it continues further to derive a more efficient formula, where the product is replaced by a sum:
```c++
float beta_sum = 0.0;
for (int iprime = 0; iprime < num_mlist[chk]; iprime++){
    [...]
    beta_sum += phi(std::abs(Q[chk][w]));
    [...]
}
R[chk][v] = sign_prod * phi(beta_sum);
```
where the function `phi` is defined as $\mathrm{phi}(x)=\log\frac{e^x+1}{e^x-1}$, and can be computed with a lookup table. It also mentions that the product of signs can be carried out using integers and modulo 2 arithmetic, which is another performance improvement that we can apply to this code.

gr-fec/lib/awgn_bp.cc
```diff
@@ -99,39 +105,75 @@ void awgn_bp::spa_initialize()
...
+                w = mlist[chk][iprime] - 1;
+                //compute prod(sign(LLR))
+                if (std::signbit(Q[chk][w])){sign=-1;}
+                else {sign=1;}
+                sign_prod = sign_prod * sign;
+                //remove divide by 2 for manual tanh
+                x = std::abs(Q[chk][w])/2.0;
+                //x = std::abs(Q[chk][w]);
+                //clamp tanh input (this is some BS, FIXME)
+                if (x > 2.35){x = 2.35;}
```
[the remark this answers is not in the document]
daniestevez: I just realized that the code is using `double`'s rather than `float`'s. Still I think that my suggestion to use the `float32` machine epsilon is good because the code probably would also work equally well (in terms of sensitivity) with `float`'s instead of `double`'s, because the LLR calculations can be successfully carrier out in integer arithmetic with just a few bits of quantization.

gr-digital/python/digital/soft_dec_lut_gen.py
```diff
@@ -103,11 +103,21 @@ def soft_dec_table(constel, symbols, prec, npwr=1):
...
-    re_min = min(numpy.array(constel).real)
-    im_min = min(numpy.array(constel).imag)
-    re_max = max(numpy.array(constel).real)
-    im_max = max(numpy.array(constel).imag)
+    #normalize constellation to unit power
+    power = 0
+    constel = numpy.array(constel)
+    for point in constel:
+        power += numpy.abs(point)**2
+    power = numpy.sqrt(power/len(constel))
```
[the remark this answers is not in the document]
VladislavF: I have gone ahead and renamed the variable.

gr-fec/lib/awgn_bp.cc
```diff
@@ -99,39 +105,75 @@ void awgn_bp::spa_initialize()
...
-        }
-        for (int i = 0; i < num_mlist[chk]; i++) {
-            v = mlist[chk][i] - 1;
-            _prdct = product / double(2 / (1 + Q[chk][v]) - 1);
-            R[chk][v] = double((1 - _prdct) / (1 + _prdct));
+            //compute product of sign and and tanh LLR's
+            for (int iprime = 0; iprime < num_mlist[chk]; iprime++){
+                //exclude iprime from the product set
+                if (iprime == i){iprime++;}
+                if (iprime >= num_mlist[chk]){break;}
```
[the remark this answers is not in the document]
VladislavF: Also a good suggestion. Done.

gr-fec/lib/awgn_bp.cc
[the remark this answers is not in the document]
VladislavF: Okay this is tricky. I have experimented with this for a while. 2.35 specifically gives the best results. Not more not less. I had the exact same idea with the epsilon. I think it makes sense. I have no idea why it gives inferior results. The clamp did have to be in the internal loop though because that is where the first NaN or Inf values would appear. I'm really not sure wtf is going on with this, and I have not yet figured it out.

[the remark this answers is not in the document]
VladislavF: > where the function `phi` is defined as phi(x)=log⁡ex+1ex−1, and can be computed with a lookup table. 

As far as your first comment on the algo. This was a very interesting find. I went ahead and tried it and it seemed to work just as well as the current code save for a bit of a CPU performance hit due to the extra logs and exponents. For the most part... Turns out, it suffers from the same zero problem that the original algo did no doubt because of the nested log(exp()) precision issues. I tried clamping the **phi** function values too, but I wasn't immediately successful. I imagine that having **phi** as an actual LUT though would probably solve both the zero and performance issues. For now I have reverted to what is working more reliably since I just didn't have time to troubleshoot where the NaNs were coming from this time.

> It also mentions that the product of signs can be carried out using integers and modulo 2 arithmetic, which is another performance improvement that we can apply to this code.

I also investigated this briefly as it was mentioned in the original reference I used as well. This makes great sense, but just turned out to harder than I'd hoped to implement. Will have to revisit this.

VladislavF:
It seems that tweaking the soft decisions has closed the performance gap between the current BP algo and the old one. At least in the very brief testing I did. This no doubt has to do with the tanh/atanh non-linear characteristics actually being taken advantage of. 

Previously, the soft decisions were being clipped. Specifically the LUT always spanned +-re(1) +-im(1) with no regard for varying normalization or gaussian padding. I have made it such that now the LUT is normalized to span +-2*max_num_in_const in both real and imaginary planes.

This also fixes the constellation object allowing it to specify a noise power so that the 'auto' and 'None' options can actually work as intended in the LUT argument. I also added a noise power arg. to the soft decoder block in case 'None' was specified and noise has changed since the const. was defined. This however still needs a callback to update at runtime.

willcode:
@VladislavF I know this is still a WIP, so you probably already know this ... there are a couple of API changes in here without default args. If this is intended for v3.10, the API has to stay the same.

VladislavF:
@willcode 

> there are a couple of API changes in here without default args. If this is intended for v3.10, the API has to stay the same.

Not intentional at all. I am not familiar with your best practices. Please go ahead and point out what I'm missing. I thought I had everything filled with a default arg. but I'm not surprised at all that I missed something. I know I added a default arg. in a few places where it "made sense".

gr-digital/include/gnuradio/digital/constellation.h
```diff
@@ -169,7 +170,7 @@ class DIGITAL_API constellation : public std::enable_shared_from_this<constellat
      *        generating the LUT.
      */
     void set_soft_dec_lut(const std::vector<std::vector<float>>& soft_dec_lut,
```
willcode, line 172: API change

gr-digital/include/gnuradio/digital/constellation_soft_decoder_cf.h
```diff
@@ -40,7 +40,7 @@ class DIGITAL_API constellation_soft_decoder_cf : virtual public sync_interpolat
      * 'constellation'. Use base() method to get a shared pointer to
      * this base class type.
      */
-    static sptr make(constellation_sptr constellation);
+    static sptr make(constellation_sptr constellation, float npwr);
```
willcode, line 43: API change

gr-digital/include/gnuradio/digital/constellation.h
VladislavF: this one has a default argument no? I'm sorry, I'm new to the practices here.

gr-digital/include/gnuradio/digital/constellation_soft_decoder_cf.h
VladislavF: should line 43 include for instance "float npwr=1.0"?

gr-digital/include/gnuradio/digital/constellation.h
willcode: I don't see a default on line 172, where `npwr` was added to `set_soft_dec_lut()`.

gr-digital/include/gnuradio/digital/constellation_soft_decoder_cf.h
willcode: Yes. Anywhere a param is added in a public header, existing code needs to compile against it. Adding a default here would allow that.

gr-fec/lib/awgn_bp.cc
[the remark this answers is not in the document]
daniestevez: > The clamp did have to be in the internal loop though because that is where the first NaN or Inf values would appear. I'm really not sure wtf is going on with this, and I have not yet figured it out.

Am I understanding correctly that `tanh()` is the function that is making the NaN's or Inf's appear in the first place? I'm testing what happens if we evaluate the `tanh()` of a large number, such as `100.0`, and it gives `1.0`: https://godbolt.org/z/TWrsda9bf

gr-digital/lib/constellation.cc
```diff
@@ -266,13 +278,17 @@ std::vector<float> constellation::calc_soft_dec(gr_complex sample, float npwr)
     std::vector<float> tmp(2 * k, 0);
     std::vector<float> s(k, 0);
 
+    //check if noise power was overriden
+    if (npwr != d_npwr && npwr != 1.0){d_npwr = npwr;}
```
willcode, line 282: Why `npwr != 1.0` here and in `calc_soft_dec()`? It looks like the result is to set `d_npwr` to `npwr` in most cases, but if it's `1.0` then use the current `d_npwr`? What if the caller wants to set it to `1.0`?

gr-fec/lib/awgn_bp.cc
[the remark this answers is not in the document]
VladislavF: That was indeed the case. But let me check what happens now.

gr-digital/lib/constellation.cc
VladislavF: @willcode So I agree this is awkward, I just haven't had the time to come up with a better solution. 

The soft decision LUT needs to have a noise power to be calculated. This is now given in the constellation defenition/class. However, there is also a chance that there is no LUT defined and the user would like to compute soft decisions in real-time with an arbitrary npwr using the soft decoder block. This is made possible with the new npwr arg. The issue is that it needs a default value. The way it's built now, any change to the npwr in the soft decoder will override what the constellation was defined with as long as it is different and not the default value (1.0). If the user initialized the constellation object with a different noise power (say 0.5) but would like to set it to 1.0 with the soft decoder, then the workaround is to set it to 1.00001 or something that would evaluate: npwr != 1.0. This catch is necessary to prevent the soft decoder from always overriding what the constellation object was initialized with. Perhaps a better default to check against is -1. I just didn't want to confuse the user on reasonable values or potentially break something by trying to do math with -1. In addition, if there is a LUT defined and the npwr was changed it needs to be recomputed, but this is not currently handled either. still need callbacks.

I'm open to suggestions on this.

gr-fec/lib/awgn_bp.cc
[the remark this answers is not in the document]
VladislavF: if I clamp the value in the outer loop (tanh_prod=0.982 or tanh_prod=1-1e-7) I get bursts of bad BER. I can't explain it.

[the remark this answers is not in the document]
daniestevez: Okay. I don't understand why that happens either, but that's just a problem with my understanding. The code never lies, so there's something at play we're not noticing.

Let's keep the clamping in the inner loop unless someone can figure out how to do it in the outer loop without any BER performance problems.

gr-digital/lib/constellation.cc
daniestevez: I was going to suggest using `-1.0` as a sentinel value for the default parameter, but it has the drawbacks you mention. How about using `std::optional<float>` with a default value of `std::nullopt` instead? I think we can use C++17 in GR 3.10.

gr-digital/lib/constellation.cc
```diff
@@ -239,16 +245,22 @@ void constellation::gen_soft_dec_lut(int precision, float npwr)
...
     float step = (2.0f * maxd) / (d_lut_scale - 1);
     float y = -maxd;
     while (y < maxd + step) {
         float x = -maxd;
         while (x < maxd + step) {
-            gr_complex pt = gr_complex(x, y);
+            gr_complex pt = gr_complex(x*ptscale, y*ptscale);
             d_soft_dec_lut.push_back(calc_soft_dec(pt, npwr));
             x += step;
         }
```
daniestevez, line 266: I suggest rewriting this like so:
```c++
    float step = (2.0f * ptscale) / (d_lut_scale - 1);
    float y = -ptscale;
    while (y < ptscale + step) {
        float x = -ptscale;
        while (x < ptscale + step) {
            gr_complex pt = gr_complex(x, y);
            d_soft_dec_lut.push_back(calc_soft_dec(pt, npwr));
            x += step;
        }
 ```
 I think this empasizes the fact that we're sweeping a grid of `d_lut_scale` points covering the interval `[-ptscale, ptscale]`, with both `x` and `y`. Once we do this, we can remove the variable `maxd`, I think.

File: gr-digital/lib/constellation.cc (hunk 6 above)
Comment (willcode) [reply gr-digital/lib/constellation.cc:282]: A `#define` (or `const` or whatever) to a name like `NPWR_DEFAULT` would be good, no matter how it's implemented. That way devs don't need to find the magic value.

File: gr-digital/lib/constellation.cc [hunk 8]
```diff
@@ -239,16 +245,22 @@ void constellation::gen_soft_dec_lut(int precision, float npwr)
...
+    //check if noise power is overriden. use global if not
+    if (npwr != d_npwr && npwr != 1.0){d_npwr = npwr;}
+    else {npwr = d_npwr;}
 
     // We know we've normalized the constellation, so the min/max
     // dimensions in either direction are scaled to +/-1.
     float maxd = 1.0f;
+    //the above comment isn't really true, but LUT is funky so we'll scale our points inside the +-1 LUT
+    //we will multiply inputs to obtain LLR's for points with the scale ptscale instead of +-1
+    float ptscale = 2.0f*d_maxamp;
```
daniestevez, line 257: I think you've mentioned that one of the problems hurting the BER performance of the LDPC decoder was that the range of `x` and `y` for which the LUT was built was not appropriate, so you've increased it from `[-1, 1]` to `[-2*d_maxamp, 2*d_maxamp]`.

This seems reasonable. `[-1, 1]` is simply too small, specially for constellations whose outermost points have amplitudes much larger than one.

I wonder, is there a better choice than `[-2*d_maxamp, 2*d_maxamp]`? Maybe we can make the range slightly smaller than this to gain better resolution without getting too much clipping problems (I assume that the LUT lookup function just clips to the edge value if the `x` or `y` coordinates are outside of the LUT range). Since we're constructing this for an expected noise power, we can use the normal distribution CDF to help us. For example, if we set
```
float ptscale = d_maxamp + 2.0 * sqrtf(npwr);
```
then I think that only less than 2.2% of the LUT lookups will be outside the range. This follows because only 2.2% of the mass of the normal distribution falls above 2*sigma. So to fall outside of `ptscale`, we would need that the symbol corresponds to one of the outermost constellation points that have amplitude `d_maxamp` and that we are so unlucky that the complex noise points towards the "outside" direction and has a large amplitude (here I'm doing some handwaving/approximations, the amplitude of complex Gaussian noise actually has a Rayleigh distribution).

My choice of using 2*sigma in the above is somewhat arbitrary. We can increase it (3*sigma) or reduce it (1*sigma) depending on how bad we believe (or check) that LUT clipping is.

I'm not sure how my idea compares with your choice of `float ptscale = 2.0*d_maxamp`. Mine seems more inspired in how to avoid clipping, and will make the box larger when there is a lot of noise, and smaller when it can afford to reduce it without clipping because there is less noise. But yours seems a more simple choice. I'm happy to keep either your proposal or mine, since your proposal also probably does quite well in practice.

gr-digital/lib/constellation.cc
```diff
@@ -266,13 +278,17 @@ std::vector<float> constellation::calc_soft_dec(gr_complex sample, float npwr)
...
 
+    //check if noise power was overriden
+    if (npwr != d_npwr && npwr != 1.0){d_npwr = npwr;}
+    else {npwr = d_npwr;}
+
     for (int i = 0; i < M; i++) {
         // Calculate the distance between the sample and the current
         // constellation point.
-        float dist = std::abs(sample - d_constellation[i]);
+        float dist = powf(std::abs(sample - d_constellation[i]), 2.0f);
```
daniestevez, line 288:
```suggestion
        const float dist = std::abs(sample - d_constellation[i]);
        const float dist_sq = dist * dist;
```
Using `powf()` to compute a square seems an overkill. Also, calling `dist` to a variable which is actually the distance squared could be confusing. This fixes both problems. I've also taken the liberty to make these variables `const`.

gr-digital/lib/constellation.cc
```diff
@@ -304,26 +320,32 @@ std::vector<float> constellation::calc_soft_dec(gr_complex sample, float npwr)
...
     if (has_soft_dec_lut()) {
         // Clip to just below 1 --> at 1, we can overflow the index
         // that will put us in the next row of the 2D LUT.
-        float xre = branchless_clip(sample.real(), 0.99);
-        float xim = branchless_clip(sample.imag(), 0.99);
+        // Since we are expecting inputs between +-1. We will scale to
+        // match the constellation+padding defined in the LUT
+        float ptscale = 2.0f*d_maxamp;
+        float xre = branchless_clip(sample.real()/ptscale, 1-1e-7);
+        float xim = branchless_clip(sample.imag()/ptscale, 1-1e-7);
```
daniestevez, line 348: The comment is a bit misleading, since we're actually expecting inputs between `+-ptscale`.
How about rewriting the comment and the code around it to make it more clear?
```c++
        float xre = branchless_clip(sample.real(), ptscale-1e-7);
        float xim = branchless_clip(sample.imag(), ptscale-1e-7);
        // [...]
        float scale = d_lut_scale / (2.0f * ptscale);
        re = floorf((ptscale + xre) * scale);
        xim = floorf((ptscale + xim) * scale);
``` 

Also, I wonder if it would be better to use `roundf()` instead of `floorf()` here. `roundf()` seems better in the sense that it actually gets the nearest point from the LUT, but I haven't thought carefully if it comes with any nasty surprises such as an off-by-one error in the LUT index when we round up at the very end of the LUT (though this wouldn't be too problematic, because then there is a check that the index is in bounds).

gr-digital/lib/constellation.cc
```diff
@@ -377,6 +402,8 @@ void constellation::max_min_axes()
         d_re_min = d_im_min;
     if (d_re_max == 0)
         d_re_max = d_im_max;
+    
+    d_maxamp = std::max(std::max(d_im_max, d_re_max),std::abs(std::min(d_im_min, d_re_min)));
```
daniestevez, line 406: I find the following clearer to read:
```suggestion
    d_maxamp = std::max({d_re_max, -d_re_min, d_im_max, -d_im_min});
```

gr-digital/python/digital/soft_dec_lut_gen.py
```diff
@@ -103,15 +103,27 @@ def soft_dec_table(constel, symbols, prec, npwr=1):
...
+    power = 0
+    constel = numpy.array(constel)
+    for point in constel:
+        power += numpy.abs(point)**2
+    amplitude = numpy.sqrt(power/len(constel))
+    constel = constel/amplitude
+    #print("Max soft constellation amp: ", max(abs(constel)), " pwr:", power)
+
+    #padding will simply be 2x the largest re or imag
+    padding = 2
```
daniestevez, line 116: Just a reminder that if we change any of the C++ code regarding `ptscale` as I have suggested, this Python code probably needs to be updated to do the same thing.

gr-digital/lib/constellation.cc
VladislavF: > How about using `std::optional<float>` with a default value of `std::nullopt` instead? I think we can use C++17 in GR 3.10.
I have no idea how this works, not familiar with this syntax. I'll have to look this up monday

gr-digital/lib/constellation.cc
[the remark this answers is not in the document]
VladislavF: yes, and it depends on which normalization method was chosen by the user. None, amplitude, or power. The cases where the outermost point is 1 are possible, common even (M-PSK), but not guaranteed at all.

gr-digital/lib/constellation.cc
VladislavF: I think this should be fine, I'll give it a try Monday and report/commit if all is good.

gr-digital/lib/constellation.cc
VladislavF: @daniestevez So the idea of selecting N standard deviations to pad with was the first thought I had too. However, the 2x padding ended up being simpler as you guessed. Mainly this came down to debugging issues, so perhaps it can change now that things are working properly. On the other hand, there are a few advantages to 2x padding. Firstly, IIRC some standard like DVB adopted this technique for their soft LUT. Second, it makes determining the resolution loss very simple, the constellation is X-1 bits dense no matter the noise level. This somewhat simplifies the analysis and determination of necessary LUT resolution. Lastly, it almost guarantees no clipping under many cases. 
But again, I'm not too attached here, it just made sense when writing the code.

gr-digital/lib/constellation.cc
VladislavF: so this is also tricky. I think this whole LUT function is evil. this is the whole reason I went with my funny `ptscale` method. I originally intended to modify `maxd` and basically change the range that the indices took on. This proved to be very confusing and I was unsuccessful in getting it to work. Later however, I realized much of my trouble was because of some other bug with resolutions of 9-11 bits, and of course the whole time I was testing this was with 9 bits.
The `ptscale` scalar really just made understanding the indexing here easier for me, and I don't see why changing the index range shouldn't work as long as it's consistent to the generation of the LUT. However, I would really avoid messing with the `floorf()` because that's specifically done to take advantage of the indices being "overflowed" by having 1.0 added after clipping to push into the next index. It's very confusing, I hate it, and I don't want to mess with it now that it works.

gr-digital/lib/constellation.cc
daniestevez: On second thought, maybe `std::optional` gives problems with pybind11. I don't know how does pybind11 handle `std::optional`.

gr-digital/lib/constellation.cc
daniestevez: I'm thinking that you're right in the sense that simplicity in the 2x padding is quite desirable. Users concerned with quantization losses can simply set a larger `precision` to build a larger LUT. So I'm happy to keep the 2x padding.

gr-digital/lib/constellation.cc
VladislavF: @daniestevez Naively copy and pasting this seemed to result in the test FG hanging forever. Not sure what the problem is, but I do not have time to troubleshoot unfortunately.

daniestevez: No worries. I don't understand why my suggestion doesn't work and your code does. I thought them to be equivalent (basically replacing `maxd by `ptscale`). After reading your comment https://github.com/gnuradio/gnuradio/pull/6478#discussion_r1089902370, I think that there might be some subtleties that don't make this replacement straightforward, so I'm happy to leave the code as it is currently, given that it works.

VladislavF:
I'm sorry for leaving this alone for a few weeks now, but I just haven't had the bandwidth to address the remaining concerns. And unfortunately things are only going to get busier as new contracts kick off. However, if it's just a few minor things, I can get them cleaned up now while I have some time.

besides the checklist (squashing commits, signing etc.), the soft decoder needs a solution to the noise power default, and the LDPC definition needs sigma removed. Was there anything else I'm missing to make this mergeable? @willcode and @daniestevez  thoughts?

daniestevez:
I've gone through the open threads to try to remember what remains to be done. There are a bunch of threads which are comments or ideas from my side that didn't work out as expected so we decided to drop them. It seems I can't close these threads as resolved, but that would reduce the clutter.

Still there are a couple of threads that I think it would be good to include in this PR: https://github.com/gnuradio/gnuradio/pull/6478#discussion_r1089333988 (specifically @willcode's suggestion of using a `#define`, since my idea to use `std::optional` might give more problems than advantages in practice), and https://github.com/gnuradio/gnuradio/pull/6478#discussion_r1089765091.

willcode:
@VladislavF I guess we didn't notice before that this PR is against maint-3.10. We do all PRs against main, and the backport them if things work out well on main. Would you resubmit this against main (and close the PR)? Thanks.

willcode:
I changed to target to main. It will still require a rebase and push.

VladislavF:
Okay, I actually don't have a dev environment setup for main. Let me see if I can get one spun up

gr-digital/lib/constellation.cc
VladislavF: @jaredd do you have any suggestions on this?

jaredd: I'm not as familiar with this code as I used to be so all that follows may be ill advised. Hopefully there are some ideas you could use.

It seems to me there are too many places where (d_)npwr is being changed. I would have a separate callback in constellation to set d_npwr at runtime (with appropriate concurrency locks) and then recalc the LUT if it exists. In the symbol decoder you would pass any changes to npwr to the constellation callback.

I think changing d_npwr in the course of another call like calc_soft_dec is a confusing side-effect that should probably be avoided. Comparing npwr and d_npwr at symbol rate seems like an unnecessary performance hit so it would prob be better to avoid passing npwr to calc_soft_dec at all and rely on d_npwr in that method. I don't know if that would break calls from other scripts / blocks though.

If you _must_ have an interface to calc_soft_dec that does take npwr arg, then I would overload separately one w/ the npwr arg and one without (not using default value for npwr arg), and the one without uses d_npwr internally (e.g. just calling calc_soft_dec(..., d_npwr ).

I also agree with @willcode that there shouldn't be any magic numbers, explicitly name the default npwr.

PR #6488, opened 2023-01-24
network: change type of bytes_read UDP Source

### LINKED ISSUE #6485: Segmentation Fault in UDP Source Block
### What happened?

The UDP Source block produces a segmentation fault when any UDP packets are received on the specified port.
GDB Logs and steps to reproduce are included in the ticket.

### System Information

OS: Ubuntu 20.04.02 LTS
GR Installation Method: native distro packages - apt

### GNU Radio Version

3.10 (maint-3.10)

### Specific Version

3.10.5.0

Create a chain consisting of a UDP Source, a Throttle block and a File Sink. Relevant configurations:
UDP Source
Input Type: byte
Port: 30000
Header: None (or any other)
UDP Packet data size: 1472
Notify missed frames: No
Src 0s: No
IPv6: No
Vector Len: 1

Throttle
Type: byte
The rest as default

File Sink:
/home/user/somefile.bin
Type: byte
The rest as default.

-----------------------------------------------------
Execute the flow graph and send some UDP packets to port 30000.
The execution crashes with return code 11.

 

### Relevant log output

```shell
GDB output
----------
(...)
Thread 23 "udp_source1" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7fffc621a700 (LWP 25868)]
__memmove_avx_unaligned_erms ()
    at ../sysdeps/x86_64/multiarch/memmove-vec-unaligned-erms.S:436
436	../sysdeps/x86_64/multiarch/memmove-vec-unaligned-erms.S: No such file or directory.
(...)
(gdb) backtrace
#0  __memmove_avx_unaligned_erms () at ../sysdeps/x86_64/multiarch/memmove-vec-unaligned-erms.S:436
#1  0x00007fffeb739ae9 in gr::network::udp_source_impl::work(int, std::vector<void const*, std::allocator<void const*> >&, std::vector<void*, std::allocator<void*> >&) ()
    at /usr/lib/x86_64-linux-gnu/libgnuradio-network.so.3.10.5
#2  0x00007ffff6c6e87b in gr::sync_block::general_work(int, std::vector<int, std::allocator<int> >&, std::vector<void const*, std::allocator<void const*> >&, std::vector<void*, std::allocator<void*> >&)
    () at /usr/lib/x86_64-linux-gnu/libgnuradio-runtime.so.3.10.5
#3  0x00007ffff6c168dd in gr::block_executor::run_one_iteration() () at /usr/lib/x86_64-linux-gnu/libgnuradio-runtime.so.3.10.5
#4  0x00007ffff6c859a7 in gr::tpb_thread_body::tpb_thread_body(std::shared_ptr<gr::block>, std::shared_ptr<boost::barrier>, int) () at /usr/lib/x86_64-linux-gnu/libgnuradio-runtime.so.3.10.5
#5  0x00007ffff6c6cbdd in  () at /usr/lib/x86_64-linux-gnu/libgnuradio-runtime.so.3.10.5
#6  0x00007ffff6c6da47 in  () at /usr/lib/x86_64-linux-gnu/libgnuradio-runtime.so.3.10.5
#7  0x00007ffff661e43b in  () at /usr/lib/x86_64-linux-gnu/libboost_thread.so.1.71.0
#8  0x00007ffff7d9f609 in start_thread (arg=<optimized out>) at pthread_create.c:477
#9  0x00007ffff7ed9163 in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:95
```

willcode:
## Description
bytes_read, the result of asio::ip::udp::socket::receive_from(), was an int, but the function returns size_t. This resulted in an overflow of the int and a subsequent comparison.

Since the int result of buffer space_available() should never be negative, it is safe to cast the result to size_t. If that result is ever changed to something larger than size_t, this could cause trouble, but that seems unlikely.

## Related Issue
Fixes #6485 

UDP Source

## Testing Done
Ran random data into `mbuffer | nc` to the source's UDP port at various rates and verified data showed up on GUI sinks. This crashed or produced bad data before the change, and works after the change.

## Checklist

### DIFF
--- a/gr-network/lib/udp_source_impl.cc
+++ b/gr-network/lib/udp_source_impl.cc
@@ -251,7 +251,7 @@ int udp_source_impl::work(int noutput_items,
         }
     }
 
-    int bytes_read;
+    size_t bytes_read;
 
     // we could get here even if no data was received but there's still data in
     // the queue. however read blocks so we want to make sure we have data before
@@ -271,13 +271,13 @@ int udp_source_impl::work(int noutput_items,
             const char* read_data = asio::buffer_cast<const char*>(d_read_buffer.data());
 
             // Discard bytes if the input is longer than the buffer
-            long overrun = bytes_read - d_localqueue_writer->bufsize();
-            if (overrun > 0) {
+            if (bytes_read > d_localqueue_writer->bufsize()) {
+                size_t overrun = bytes_read - d_localqueue_writer->bufsize();
                 bytes_read -= overrun;
                 read_data += overrun;
             }
 
-            if (d_localqueue_writer->space_available() < bytes_read)
+            if ((size_t)d_localqueue_writer->space_available() < bytes_read)
                 d_localqueue_reader->update_read_pointer(
                     bytes_read - d_localqueue_writer->space_available());
             memcpy(d_localqueue_writer->write_pointer(), read_data, bytes_read);


argilo:
The conversion from `size_t` (unsigned) to `long` (signed) on line 274 will still have implementation-defined behaviour if `overrun` happens to be negative. Maybe it would be best to compare, then subtract only if there is in fact an overrun:

```c++
            // Discard bytes if the input is longer than the buffer
            if (bytes_read > d_localqueue_writer->bufsize())
                long overrun = bytes_read - d_localqueue_writer->bufsize();
                bytes_read -= overrun;
                read_data += overrun;
            }
```

argilo (approved):
The fix looks good, and the UDP sink now runs without issue.

argilo:
Thanks for taking care of the pull request, @willcode!

PR #6491, opened 2023-01-25
grc: fix Choose Editor menu needs to stay above parent dialog #5151

### LINKED ISSUE #5151: Choose Editor menu needs to stay above parent dialog
In default Ubuntu configurations the GTK dialogs are forcibly locked to the center of their parent window. This means that if someone opens the properties on an Embedded Python Block, clicks "Open in Editor", and then clicks anywhere else other than the Choose Editor dialog it can end up hidden behind the Embedded Python Block properties dialog. This results it it being impossible to close the dialog, or even GRC, as all interactions with the lower layers are prevented while the Choose Editor dialog is open.

The fix is to lock the Choose Editor dialog to stay above the parent dialog (Embedded Python Properties).
https://github.com/gnuradio/gnuradio/blob/1a0be2e6b54496a8136a64d86e372ab219c6559b/grc/gui/Dialogs.py#L400

I believe the Gtk `set_keep_above()` function will solve this.

Screenshot from 2021-10-07 09-49-20

R-ohit-B-isht:
Signed-off-by: R-ohit-B-isht <rbtunes0@gmail.com>

## Description
In default Ubuntu configurations the GTK dialogs are forcibly locked to the center of their parent window. This means that if someone opens the properties on an Embedded Python Block, clicks "Open in Editor", and then clicks anywhere else other than the Choose Editor dialog it can end up hidden behind the Embedded Python Block properties dialog. This results it it being impossible to close the dialog, or even GRC, as all interactions with the lower layers are prevented while the Choose Editor dialog is open.

## Related Issue
Fixes #5151

grc/gui/Dialogs.py line-179,180

## Testing Done
Yes

## Checklist

### DIFF
--- a/grc/gui/Dialogs.py
+++ b/grc/gui/Dialogs.py
@@ -176,6 +176,7 @@ class MessageDialogWrapper(Gtk.MessageDialog):
             self, transient_for=parent, modal=True, destroy_with_parent=True,
             message_type=message_type, buttons=buttons
         )
+        self.set_keep_above(True)
         if title:
             self.set_title(title)
         if markup:


grc/gui/Dialogs.py
```diff
@@ -176,6 +176,8 @@ def __init__(self, parent, message_type, buttons, title=None, markup=None,
             self, transient_for=parent, modal=True, destroy_with_parent=True,
             message_type=message_type, buttons=buttons
         )
+        self.set_keep_above(True)
+        self.set_transient_for(parent)
```
willcode, line 180: `transient_for` is set in the super class constructor above. Is it needed here too?

willcode (commented):
Do any of the other dialog classes need `keep_above`?

grc/gui/Dialogs.py
R-ohit-B-isht: Oops , yeah it's not needed .
No, I tried all the other dialogs they all worked fine without it.

PR #6496, opened 2023-01-30
vmcircbuf logging & string handling & file routines

marcusmueller:
- runtime/vmcircbuf_prefs: use sensible include guard name
- runtime/vmcircbuf: use sensible logging, don't use C string handling in C++

## Checklist


willcode:
Let's let this sit on main for a bit. The qa doesn't test all the failure path, but it looks good to me.

Issue #4030, opened 2020-12-28
How can I get rid of "Warning: failed to XInitThreads()"?


ShnDanYao:
How can I get rid of "Warning: failed to XInitThreads()"?

The generated python file includes those lines:
if __name__ == '__main__':
    import ctypes
    import sys
    if sys.platform.startswith('linux'):
        try:
            x11 = ctypes.cdll.LoadLibrary('libX11.so')
            x11.XInitThreads()
        except:
            print("Warning: failed to XInitThreads()")

Should the "libX11.so"  be "libc.so.6"?
How can I change the template of the python file?

drmpeg:
Depends on your system.

If you have no graphics at all, then you should set the ```Generate Options``` to ```No GUI``` in the ```Options``` block of your flow graph. Then XInitThreads() isn't called at all.

If you do have graphics, then install x11.

sudo apt-get install libx11-dev

If you're SSH'ing into a headless system, you may need dbus-x11 in addition to libx11-dev.

sudo apt-get install dbus-x11

ShnDanYao:
It worked ! Thanks!

tuscanred:
And it worked for me too; thanks!

blueride80:
Worked here as well. Many thanks!

Issue #3779, opened 2020-09-25
packet_tx.grc RuntimeError


sobakaliza:
Hello. When I try to run (on Ubuntu) the official example project uhd_packet_tx.grc which depends on packet_tx.grc, I get the following python error:

> Warning: failed to XInitThreads()
Traceback (most recent call last):
  File "/home/gnuradio/Documents/qam/uhd_packet_tx.py", line 377, in <module>
    main()
  File "/home/gnuradio/Documents/qam/uhd_packet_tx.py", line 353, in main
    tb = top_block_cls()
  File "/home/gnuradio/Documents/qam/uhd_packet_tx.py", line 183, in __init__
    self.packet_tx_a = packet_tx(
  File "/home/gnuradio/.grc_gnuradio/packet_tx.py", line 68, in __init__
    self.digital_burst_shaper_xx_0.set_block_alias("burst_shaper0")
  File "/usr/lib/python3/dist-packages/gnuradio/digital/digital_swig1.py", line 2731, in set_block_alias
    return _digital_swig1.burst_shaper_cc_sptr_set_block_alias(self, name)
RuntimeError: symbol already exists, can not re-use!

>>> Done (return code 1)

I managed to fix that issue by editing the packet_tx.py script, which is created automatically in /home/gnuradio/.grc_gnuradio
I changed the line 68: self.digital_burst_shaper_xx_0.set_block_alias("**burst_shaper0**") to self.digital_burst_shaper_xx_0.set_block_alias("**burst_shaper_a**") and now it works. 

May you kindly handle this issue? Thanks.

csylvain:
i believe the correct way to handle this issue is not to edit the hierarchical file (packet_tx.py) but to be sure to have generated the code from the flowgraph (but not *Run*), where it will be automatically stored for you in /home/<username>/.grc_gnuradio .. the ".py" and ".yml" files will be placed there and ready to be referenced by another flowgraph such as packet_loopback_hier.grc

sobakaliza:
I agree that editing the Python file is not the right solution to this problem. But it is still weird that the original example does not work from the scratch. Though my directories were clean.

csylvain:
it worked for me from scratch, but only after being very deliberate in
generating the code of the hier blocks first before opening the flowgraph
that uses them.

sobakaliza:
Sounds interesting! Maybe there should be some official documentation about how to execute the blocks.

WA1CYB:
[...]

PR #6498, opened 2023-01-30
grc: py template: only XinitThreads if $DISPLAY is set

marcusmueller:

## Description

Currently, on every Linux system, a GRC-generated flowgraph tries to load X11 libraries and then tries to run `XInitThreads`, in an attempt to make our multi-thread drawing threadsafe (this might be a design bug, or not even necessar anymore).

This at least checks whether the famous `DISPLAY` environment variable is set, as it should be on any X11 session, and whether the `WAYLAND_SESSION` environment variable is unset, which would indicate we're running on wayland instead of X.

**HOWEVER** I don't have a wayland system to try with.

**ALSO** zero input from WSL/X11 usage and FreeBSD and other X-based systems.

## Related Issue

#4030, #3779 , and a lot of newbie confusion

GRC-generated Python Flowgraphs

## Checklist

- [-] I have added tests to cover my changes, and all previous tests pass.

### DIFF
--- a/grc/core/generator/flow_graph.py.mako
+++ b/grc/core/generator/flow_graph.py.mako
@@ -25,9 +25,10 @@
 from packaging.version import Version as StrictVersion
 
 if __name__ == '__main__':
-    import ctypes
-    import sys
-    if sys.platform.startswith('linux'):
+    from sys import platform
+    from os import environ as argp
+    if platform.startswith('linux') and "WAYLAND_DISPLAY" not in argp and "DISPLAY" in argp:
+        import ctypes
         try:
             x11 = ctypes.cdll.LoadLibrary('libX11.so')
             x11.XInitThreads()


willcode:
Is there another way to tell if `XInitThreads()` should not be called, since `WAYLAND_DISPLAY` is not always
set for Wayland, as `DISPLAY` is for X11?
```
$WAYLAND_DISPLAY

This tells Wayland clients which Wayland compositor to connect to. If it is not set, clients
will try wayland-0 as a fallback. WAYLAND_DISPLAY is normally the name of a socket in
XDG_RUNTIME_DIR, but it can also be an absolute path to a socket anywhere.
```

willcode:
Possibly check `XDG_SESSION_TYPE`?

willcode:
As discussed on matrix, we may not need `XInitThreads()` at all. It is called in the gdk x11 init code.

willcode:
In Qt, `XInitThreads()` is called in XCB (X11 abstraction layer for Qt)
https://github.com/cedrus/qt/blob/f215ae45c75f7f15cab4f5ad8bda48104725a55e/qtbase/src/plugins/platforms/xcb/qxcbintegration.cpp#L129

Here is a simple test to show that `XInitThreads()` is called internally to Qt. Tested on Fedora 37. Ubuntu ltrace doesn't seem to work at all due to the way things are linked there (but the symbol is found in `libQt5XcbQpa.so`).
```
$ ltrace -e XInitThreads python -c "from PyQt5 import Qt"
libX11.so.6->XInitThreads(3, 0x7ffe0a32b898, 0x7ffe0a32b8b8, 0x7f5e511fa6a0) = 1
```

marcusmueller:
Ah, awesome!

Yeah, then this makes this PR redundant, and a much easier solution becomes viable: remove the libX11-calling code from the templates alltogether. Yay!

PR #6502, opened 2023-02-01
runtime: getpagesize for windows

### LINKED ISSUE #6500: Windows MSVC builds have no way to query the system page size
### What happened?

Report [here](https://github.com/ryanvolz/radioconda/issues/47) of a user seeing the message
```
pagesize :error: no info; setting pagesize = 4096
```
on Windows whenever a GNU Radio flowgraph is executed.

That message comes from [here in pagesize.cc](https://github.com/gnuradio/gnuradio/blob/25ad39f7a1a788c36e2ef712ac17b8f4b4d9891d/gnuradio-runtime/lib/pagesize.cc#L43). Since MSVC Windows builds only have a dummy `unistd.h` without `getpagesize` or `sysconf` defined (`HAVE_GETPAGESIZE` and `HAVE_SYSCONF` are false), they will always follow the path that shows this message.

I guess it would be good to not have this displayed as an error since the flowgraph runs fine. Even better would be to query the page size [using the Windows API](https://learn.microsoft.com/en-us/windows/win32/api/sysinfoapi/ns-sysinfoapi-system_info).

### System Information

OS: Windows
GR Installation Method: conda

### GNU Radio Version

3.10 (maint-3.10)

### Specific Version

3.10.5.1

Run a flowgraph on Windows, actually read the console output.

### Relevant log output

_No response_

marcusmueller:
## Description

Windows users shouldn't be faced with an error when determining the page size – adressing that.

- removed unused prefs include
- moved pagesize determination into separate function, to reduce the ugliness of the #ifdefs
- More specific logging

## Related Issue

Fixes #6500

runtime

## Testing Done

Here's the catch – the usual ctest suite runs through fine, but I don't have a Windows machine to test this with.

## Checklist

### DIFF
--- a/gnuradio-runtime/lib/pagesize.cc
+++ b/gnuradio-runtime/lib/pagesize.cc
@@ -1,6 +1,7 @@
 /* -*- c++ -*- */
 /*
  * Copyright 2003,2013 Free Software Foundation, Inc.
+ * Copyright 2023 Marcus Müller
  *
  * This file is part of GNU Radio
  *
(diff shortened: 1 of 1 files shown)


marcusmueller:
Will fix the clang-formatting once conda/win has run through.

ryanvolz (approved):
Looks good, and I tested the Windows conda package that the CI built and the error message is gone! Now I see
```
pagesize :info: Setting pagesize to 4096 B
```
which is as intended.

PR #6512, opened 2023-02-05
Mait/update codec2 mode enums

maitbot:
gr-vocoder: support newer libcodec2 mode enums

## Description

For a while the FreeDV and Codec2 blocks in gnuradio-companion do
not list all available modes. This updates the list.

## Related Issue

Also, this sets the mode parameters using the enum dtype - so that there is
no longer chatter about unsupported options when starting gnuradio-companion.
This adds Codec2 and FreeDV mode enum names for use in C++,
Python and grc for newer modes found in libcodec2 when building
gr-vocoder.

This patch is prepared in response to a user, and hopefully will get
in to Debian Bookworm.
These changes work for me when tested locally.

## Testing Done

This patch is prepared in response to a user, and hopefully will get
in to Debian Bookworm.
These changes work for me when tested locally.

## Checklist

### DIFF
--- a/gr-vocoder/grc/vocoder_codec2_decode_ps.block.yml
+++ b/gr-vocoder/grc/vocoder_codec2_decode_ps.block.yml
@@ -5,7 +5,7 @@ flags: [ python, cpp ]
 parameters:
 -   id: mode
     label: Bit rate
-    dtype: raw
+    dtype: enum
     default: codec2.MODE_2400
     options: [codec2.MODE_3200, codec2.MODE_2400, codec2.MODE_1600, codec2.MODE_1400,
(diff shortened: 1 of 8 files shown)


willcode:
@maitbot Could you submit this against `main`?

Use of enums does make for nicer code, too. Can you think of any backward/forward compatibility issues that change when going from raw (evaluate value if possible when generating code) to enum (always evaluate value at runtime)?

willcode:
Seem right to me. I just get nervous when we change back and forth between symbols and values. People use old flowgraphs on new GR versions and vice versa. I don't think this breaks anything that wouldn't break just based on the additional options, which someone will run into trying to backport a flowgraph.

Can you resubmit this against `main`.

PR #6513, opened 2023-02-06
update codec2 mode enums

maitbot:
Use newer modes from Codec2 and FreeDV as enums.

For a while the FreeDV and Codec2 blocks in gnuradio-companion do
not list all available modes. This updates the list.

## Description

This adds Codec2 and FreeDV mode enum names for use in C++,
Python and grc for newer modes found in libcodec2 when building
gr-vocoder.

## Related Issue

Also, this sets the mode parameters using the enum dtype - so that there is
no longer chatter about unsupported options when starting gnuradio-companion.

gr-vocoder: Codec2 and FreeDV blocks get a more complete list of modes
available in libcodec2.

## Testing Done

This patch is prepared in response to a user, and hopefully will get
in to Debian Bookworm.
These changes work for me when tested locally.

## Checklist

### DIFF
--- a/gr-vocoder/grc/vocoder_codec2_decode_ps.block.yml
+++ b/gr-vocoder/grc/vocoder_codec2_decode_ps.block.yml
@@ -5,7 +5,7 @@ flags: [ python, cpp ]
 parameters:
 -   id: mode
     label: Bit rate
-    dtype: raw
+    dtype: enum
     default: codec2.MODE_2400
     options: [codec2.MODE_3200, codec2.MODE_2400, codec2.MODE_1600, codec2.MODE_1400,
         codec2.MODE_1300, codec2.MODE_1200, codec2.MODE_700, codec2.MODE_700B,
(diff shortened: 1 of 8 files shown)


gr-vocoder/grc/vocoder_codec2_decode_ps.block.yml
```diff
@@ -5,7 +5,7 @@ flags: [ python, cpp ]
 parameters:
 -   id: mode
     label: Bit rate
-    dtype: raw
```
mormj, line 8: iirc, having this as `raw` allows the value to be set as a variable in GRC but doesn't affect the selectability of it as an enum

maitbot: After sleeping on it, there is a good case to keep it as raw.

But there was a message printed out when starting gnuradio-companion ..." option_attributes are for enums only, ignoring"...
which was annoying and made me think it would prefer it to be a enum.

If we change it back to raw, we should quiet the output from grc/core/blocks/_build.py too.

willcode: It's not just a warning. It deletes `option_attributes` if not `enum` so the C++ code here won't work if `raw`.

mormj (approved):

PR #6514, opened 2023-02-06
Logging for Python Top Blocks, Hier Blocks

marcusmueller:
- runtime: hier_block2.py: remove SWIG reference, reflow to 120 line length
- runtime: add python-land loggers to top_block and hier_block2

## Description

I added loggers to gr::block wrapped to python, but didn't so for hier_block2, top_block.

This fixes that!

The approach here is slightly different, though: in gr::block / block_gateway, I made sure that the logger obtained in Python is *the very same* logger as the d_logger member of the underlying block. Here, I'm doing the opposite: Python Hier/top blocks are logically "independent enough" from the C++ foundation that it makes sense for the Python code setting things up to have their own loggers. 

Now, while that approach means that no changes to basic_block.h nor hier_block2.h, top_block.h were necessary, it meant that I had to make `top_block.__init__` actually call the superclass ctor. In general, this seems to me the wiser approach, but I'm sure that mechanism was there for a reason.

While I was doing that, had to change the "redirection to _impl" method from an overloaded `__getattribute__` to actually adding the relevant members of the dimpl to the owning class – which has the interesting side effect of allowing IDEs/python shells to autocomplete these now.

Python hier blocks, top blocks, so:

gnuradio-runtime

## Testing Done

Test cases for logging from a python top block , and a hier block added.

## Checklist

### DIFF
--- a/gnuradio-runtime/python/gnuradio/gr/hier_block2.py
+++ b/gnuradio-runtime/python/gnuradio/gr/hier_block2.py
@@ -9,7 +9,6 @@
 
 import functools
 
-# from .runtime_swig import hier_block2_swig, dot_graph
 from .gr_python import hier_block2_pb
 
 import pmt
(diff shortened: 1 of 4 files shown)


mormj (approved):
This looks great!  Nice to have the top level and hier block logging.  Any concern about having all the underlying implementation methods exposed publicly?

marcusmueller:
Sorry for the delayed response, @mormj: I don't it's any more critical than it was before!

PR #6515, opened 2023-02-06
qtgui: Add fosphor display widget

gmeyerho:
- This adds OpenGL as an optional Qt5 component dependency. If OpenGL is not available, this block will not be built.
- The new block (fosphor_display) can consume compact histogram data and display it in a nice way. It was originally designed to read data from the fosphor RFNoC block, which it can still do, but it will work with any other source that can generate compact histogram data.

Signed-off-by: Grant Meyerhoff <grant.meyerhoff@ni.com>

## Description
This commit was made originally by Martin Braun and handed off to me, my only adjustments were clang-format enforced and a CMakeLists.txt edits, as well as simplifying the example. This code is mostly taken from the gr-ettus OOT module and adapted for this repo (see https://github.com/EttusResearch/gr-ettus/tree/maint-3.8-uhd4.0/lib).

Since the code was mostly Martin's, I did my review passthrough of the code first before putting it up. That included comparing the files to the corresponding gr-ettus files, most changes involving syntax rather than functionality.

This PR will be followed shortly by a PR adding support for the RFNoC Fosphor Block that utilizes this widget (#6516)

UPDATE: Add commit for fosphor formatter block, which replaces the previous embedded python block in the example which was not performant enough

## Related Issue
New feature to adding RFNoC Fosphor Block support

This will affect the RFNoC Foshpor Block support, not in functionality, but in displaying the data.

## Issues needing to be resolved from comments
- [x] Make example for only fosphor display simpler to increase ease of use outside of RFNoC usage. Potentially creating new block to feed into the fosphor display
- [x] Respond with OpenGL package for CI images and add qa file for testing fosphor display block
- [x] Change naming of block to differentiate from gr-fosphor block

## Testing Done
Ubuntu 20.04.5 LTS
Ran the fosphor display example that uses the Fosphor GL Sink which combines Fosphor Formatter block and Fosphor Display Block

https://user-images.githubusercontent.com/68391340/232878066-2c4b6a62-aa6d-4402-82a3-b70108e00ffb.mp4

## Checklist


willcode:
Missing requirement on WIndows not detected, or something else?
`%SRC_DIR%\gr-qtgui\lib\QFosphorColorMapper.cc(154): error C3861: 'glActiveTexture': identifier not found`
https://github.com/gnuradio/gnuradio/actions/runs/4108894170/jobs/7090106685#step:6:4929

gmeyerho:
> Missing requirement on Window and OSX not detected, or something else? `%SRC_DIR%\gr-qtgui\lib\QFosphorColorMapper.cc(154): error C3861: 'glActiveTexture': identifier not found` https://github.com/gnuradio/gnuradio/actions/runs/4108894170/jobs/7090106685#step:6:4929

Yea, I am not quite sure, I am not that familiar with OpenGL, trying to setup a Windows system to build and debug

willcode:
How will we help users distinguish this block from the `fosphor sink (Qt)` provided by `gr-fosphor`? I can see a lot of confusion, where one block takes a time domain stream and one takes already-processed vectors to display.

dkozel:
Hi Grant, thanks for putting together the PR.

It would be very useful to have a block that can generate the compact histogram vectors from a time domain stream as part of this to allow for testing and example flowgraphs in the absence of RFNoC hardware. Is it within scope for you to write one?

gmeyerho:
Taking out of draft as Windows and Mac build issues are resolved

gmeyerho:
> How will we help users distinguish this block from the `fosphor sink (Qt)` provided by `gr-fosphor`? I can see a lot of confusion, where one block takes a time domain stream and one takes already-processed vectors to display.

Interesting, I was not aware of that separate block.

In the description for the item, it mentions that it is a histogram/waterfall display and that the block can consume one or two vectors of unsigned 8-bit numbers. So that can help the user distinguish, but I do understand a user might not read the description and just think that the gr-fosphor block was ported over.

I know the main use case for this block is with rfnoc, but rfnoc isn't required to use it. A line in the description could be added to specify that the main purpose for this block is with rfnoc hardware, but doesn't require it. Would that help ease concerns for confusion?

gmeyerho:
> Hi Grant, thanks for putting together the PR.
> 
> It would be very useful to have a block that can generate the compact histogram vectors from a time domain stream as part of this to allow for testing and example flowgraphs in the absence of RFNoC hardware. Is it within scope for you to write one?

There is an example within this commit that allows for testing without RFNoC hardware, it isn't a nice one block source->sink, but does that help to ease the testing concern? The scope of my work is focused on getting all UHD RFNoC blocks ported to GNU Radio 3.10, as well as the supporting blocks such as this one.

gr-qtgui/lib/CMakeLists.txt
```diff
@@ -87,10 +101,25 @@ target_include_directories(
...
+        PRIVATE ${PRIVATE_LIBS}
+        PRIVATE "-framework OpenGL"
+    )
+else()
+    target_link_libraries(gnuradio-qtgui
+        PUBLIC ${QTGUI_LIBS}
+        PRIVATE ${PRIVATE_LIBS}
+        PRIVATE opengl32
+    )
+endif()
```
ryanvolz, line 122: Rather than linking the specific OpenGL library/framework for each platform, I think the best way would be to `find_package(OpenGL ...)` when we have `ENABLE_GR_QTGUI_OPENGL` (this could be part of the if statement on line 73) and then add the provided `OpenGL::GL` target to `PRIVATE_LIBS`.

I don't know why this is necessary for macOS and Windows but not Linux. Ideally I'd think that the `Qt5::OpenGL` target would bring along the necessary GL libraries if it requires them. Then again, this block directly calls `gl*` functions and so maybe it *is* proper to have to specifically link with the GL libraries (so why does Linux get away with not doing it?). Anyway, unless someone knows better, the goal is to just get it building, so these links are evidently necessary.

ryanvolz (commented):
I'm only commenting on the CMake parts since that's where I think I can help with my perspective as a packager. Nice work getting it building on all platforms!

gr-qtgui/lib/CMakeLists.txt
gmeyerho: Thanks, that is much cleaner

willcode:
I think we need to add OpenGL support to the CI images. This code does not appear to get built during a CI run. What are the minimum required packages for Debian, Ubuntu and Fedora?

In `gr-qtgui/python/qtgui/qa_qtgui.py`, the various sinks are instantiated just to make sure everything works all the way through to Python. Could you add a separate qa file for this one (that gets enabled only when this sink is built)?

willcode:
Is there a simpler example of non-RFNoC use than the included `test_fosphor_display`? Does the block require the raster generation/repeat stuff to work?

The WF did not show up when I tried that example. Resizing the window was REALLY slow. I also tried to input a real signal at 2400000 samples/sec but couldn't figure out how to make that work. I think if this is going to be sitting next to the other Qt sinks in the block tree, that we're going to have to make it easier to use.

willcode:
Are the QT GUI Time Sink and the QT GUI Fosphor Display supposed to show the same thing? When I turned on the Time Sink, the displays did not match.
image

gmeyerho:
> Is there a simpler example of non-RFNoC use than the included `test_fosphor_display`? Does the block require the raster generation/repeat stuff to work?
> 
> The WF did not show up when I tried that example (the histogram did). Resizing the window was REALLY slow. I also tried to input a real signal at 2400000 samples/sec but couldn't figure out how to make that work. I think if this is going to be sitting next to the other Qt sinks in the block tree, that we're going to have to make it easier to use.

My testing has been focused on the RFNoC-based example in #6516. My testing of this example more basic after branching the commits to my repo to ensure it worked. I am not sure why the waterfall isn't showing up for you, I can see both the histogram and the waterfall in my testing. It seems like the space for it is showing up below the histogram, but something is off with the coloring. (I pasted a picture of the example running on my machine in the PR description)

I am noticing the resizing is much slower than in the RFNoC example. In the RFNoC example, the resizing is smooth.

Let me work on creating a simpler example and discuss this with my team, and I'll investigate the waterfall issue you are seeing. I think ideally this block would be added to the QT blocks tree, but for prioritizing porting all RFNoC blocks, it may make more sense to keep this in the UHD tree, at least initially. That way the intention is clear this is currently only meant to be used within the context of UHD devices.

willcode:
@gmeyerho This PR picked up some other changes from main ... probably need to rebase.

gmeyerho:
> I think we need to add OpenGL support to the CI images. This code does not appear to get built during a CI run. What are the minimum required packages for Debian, Ubuntu and Fedora?

For Ubuntu it is libqt5opengl15-dev, it looks like this package is on Debian as well, but not on Fedora. I am looking into what Fedora requires.

gmeyerho:
I have changed the example to have two python blocks that format the data output from the fft block after converting the complex data to magnitude. This gives the advantage of the user being able to open up the block and see the code, and make tweaks to the behavior if desired. (e.g. they want the white and red line to represent something other than max and average value.

willcode:
I'll give this a try later this week. Are the python blocks you mentioned included in the PR?

gmeyerho:
> I'll give this a try later this week. Are the python blocks you mentioned included in the PR?

They are included in the example, using the Embedded Python Block, rather than a new independent block in the tree

willcode:
Is there a description of the histogram format somewhere? I wasn't able to reverse engineer it quickly from the embedded Python code. What are bin length and bin size?

I'm still getting a blank area for the waterfall. When I disable the waterfall, the flowgraph freezes with a blank window for on the order of a minute before anything at all is drawn, then it freezes again.

The update rate for the histogram is about 1/sec with a 100MS/s rate and target frame rate of 30/sec. Even if numpy isn't keeping up, it seems like something would be displayed faster than that.

In the histogram Python block, there is
`value = array[index] // self.bin_size` in `update_vectors()`
after
`self.update_vectors(input_items[0][index].astype(np.ubyte))` in `work()`
Is this casting float values (<<1.0) to ubyte then doing integer division?

gmeyerho:
> Is there a description of the histogram format somewhere? I wasn't able to reverse engineer it quickly from the embedded Python code. What are bin length and bin size?

There is in the uhd code, but let me elaborate in the first comment in the python file. 

And to explain it here: the histogram format is a grid with frequency bins along the x-axis, the length of the fft vector, and the y-axis is the number of power bins (num_bins would probably be better than bin length, I'll change that). Bin size is how much each bin represents. Each time an fft vector is input, it is sorted into bins (e.g. fft vector at index 7 is has a value of 9 and bin size is 2, then power bin 4 is incremented at index 7, as well as power bins 3, 2, and 1 at index 7). Once that is done for the 66 samples, in this case, then the power bin vectors are sent to the display block alongside a max power vector and an average power vector.

There is nice picture to help illustrate from the uhd code that I will copy to the python block too.

> In the histogram Python block, there is `value = array[index] // self.bin_size` in `update_vectors()` after `self.update_vectors(input_items[0][index].astype(np.ubyte))` in `work()` Is this casting float values (<<1.0) to ubyte then doing integer division?

Yea, nice catch, the ubyte cast shouldn't be there anymore since that is used for sorting, I will fix that.

gmeyerho:
> I'm still getting a blank area for the waterfall. When I disable the waterfall, the flowgraph freezes with a blank window for on the order of a minute before anything at all is drawn, then it freezes again.
> 
> The update rate for the histogram is about 1/sec with a 100MS/s rate and target frame rate of 30/sec. Even if numpy isn't keeping up, it seems like something would be displayed faster than that.

What OS is your system running? I notice in the past picture you posted, only the red line showed up, which would only be the max value vector, and none of the power bin color mapping. It seems like something isn't working properly with OpenGL like it is on my system. I can try testing on the OS you are using to debug that.

Here is a clip of the example running on my system, and what the expected behavior would be.

https://user-images.githubusercontent.com/68391340/222848257-d352fd63-69bc-45fe-b1b7-80c3973d8bf2.mp4

willcode:
Fedora 37 on a recent Dell XPS 17 laptop with NVIDIA GeForce RTX 3060. The OOT gr-fosphor runs nicely.

mbr0wn:
I just ran this on my Dell PC (OptiPlex 7070) which has an integrated display controller:

```
00:02.0 VGA compatible controller: Intel Corporation CoffeeLake-S GT2 [UHD Graphics 630] (rev 02)
```

I guess if it can do UHD Graphics, it can also do RFNoC Graphics :trollface:

mbr0wn:
I also noticed this while running CMake:
```
--   Override with -DENABLE_GR_QTGUI=ON/OFF
CMake Warning (dev) at /usr/share/cmake-3.22/Modules/FindOpenGL.cmake:315 (message):
Policy CMP0072 is not set: FindOpenGL prefers GLVND by default when
available.  Run "cmake --help-policy CMP0072" for policy details.  Use the
cmake_policy command to set the policy and suppress this warning.

FindOpenGL found both a legacy GL library:

OPENGL_gl_LIBRARY: /usr/lib/x86_64-linux-gnu/libGL.so

and GLVND libraries for OpenGL and GLX:

OPENGL_opengl_LIBRARY: /usr/lib/x86_64-linux-gnu/libOpenGL.so
OPENGL_glx_LIBRARY: /usr/lib/x86_64-linux-gnu/libGLX.so

OpenGL_GL_PREFERENCE has not been set to "GLVND" or "LEGACY", so for
compatibility with CMake 3.10 and below the legacy GL library will be used.
gr-qtgui/lib/CMakeLists.txt:86 (find_package)
This warning is for project developers.  Use -Wno-dev to suppress it.
```

Do you know which OpenGL library you're using? @gmeyerho what about you?

mbr0wn:
I just re-ran CMake to link against the "other" library (libOpenGL.so instead of libGL.so) but saw no difference. Works fine in either case.

FYI, in case you didn't know like me, "GLVND" is a "Vendor Neutral Dispatch" library for GL. I guess that's what we want. But in my case both the legacy and non-legacy libraries were both doing vendor-neutral dispatches.

Also, I noticed that the only CPU hog is the histogram formatter block. The fosphor display is pretty smooth.

willcode:
I still see this after rebuilding with that var set in cmake. It got rid of the message, which I was getting too..
image

willcode:
True, see what happens if you run `QT_STYLE_OVERRIDE=adwaita-dark test_fosphor_display.py`.

mbr0wn:
My system didn't have `adwaita` (only `breeze`) but that looks the same for light or dark.

willcode:
I changed to the default theme, rebooted, same thing. I thought it might be blocked on processing. Sample rate 10kS/s is about what the Python block will handle, still no change. I don't know how to debug this any further, unfortunately. Maybe I can find a way to tell GL to use a different driver for one app.

gmeyerho:
Added the Don't Merge label as the added fosphor formatter commit relies on the lambda block commit that is currently in PR. I temporarily added that commit to this branch so that the build doesn't break

willcode:
@mbr0wn @gmeyerho What should be used for `scale` in the formatter in order to get dB correct on the plot. What dB units are being used? If set `scale` to `4`, then I get what I see in Gqrx. The GR freq sink gives yet another answer, but I think that's explainable.

willcode:
Here's a comparison with Gqrx showing Volts and the Fosphor block with 1024x64 and scale=4. Things line up pretty well. When I changed fft size or num bins, the scaling changed and I didn't try to figure out what a good scale would be.
image

gmeyerho:
> @mbr0wn @gmeyerho What should be used for `scale` in the formatter in order to get dB correct on the plot. What dB units are being used?

I am not sure off the top of my head, and wasn't able to find time to look through the original gr-ettus implementation to see what the units are today as some other items came up. Maybe @mbr0wn knows, but I will take a look at this next week.

mbr0wn:
I think it's unspecified (a bit like our QT GUI Frequency Sink). Most useful would be dBm, but that would require integration into our calibration framework. I always thought the "Volts" thing was a bit of a cheat; since we don't have oscilloscopes here and our dBFS scaling is a bit arbitrary in SDRs.

willcode:
The frequency sink has the log10 in it and displays dBV and thought the fosphor formatter did that too. Better the way you have it.

willcode:
@mbr0wn I think there's a factor of fft_size missing/extra? somewhere in the calculations. For a N=1024 FFT, using 4 for the scale param in the formatter block makes everything come out about right. For a N=512 FFT, the output is too low using 4 for the scale param. I spent a little time tracking powers of N through the different blocks but didn't find the error right away.

gr-qtgui/lib/fosphor_formatter_impl.cc
```diff
@@ -0,0 +1,286 @@
...
+      // Init sub-blocks
+      d_s2v(gr::blocks::stream_to_vector::make(sizeof(gr_complex), fft_size)),
+      d_input_decim(
+          gr::blocks::keep_one_in_n::make(sizeof(gr_complex) * fft_size, input_decim)),
+      d_fft(gr::fft::fft_v<gr_complex, true /* forward */>::make(
+          fft_size, gr::fft::window::blackman_harris(fft_size), true /* shift */)),
+      d_c2m(gr::blocks::complex_to_mag_squared::make(fft_size)),
+      d_log(gr::blocks::nlog10_ff::make(calculate_log10_scale(scale),
+                                        fft_size,
+                                        20 * std::log10(static_cast<float>(fft_size)))),
```
willcode, line 261: This may be part of the problem. I think the intent is to divide by N^2, but this is multiplying by N^2. Even after that, I still haven't been able to find a scaling that matches the fosphor visualization against the GR frequency sink or Gqrx.

Changing the sign on "k" here and setting "scale" to 6 puts tones at 0dB at all FFT sizes. However, wideband signals and the noise floor don't match other the frequency sink or Gqrx.

willcode:
@mbr0wn I think the `k` in the `log10()` is supposed to be something more like a fixed value to center the log on the graph. Then `scale` is really a fixed value like `256/100` to scale the log value on the graph. Do you have the math you used around still?
```
diff --git a/gr-qtgui/lib/fosphor_formatter_impl.cc b/gr-qtgui/lib/fosphor_formatter_impl.cc
index f09e2d33a..eacc217d6 100644
--- a/gr-qtgui/lib/fosphor_formatter_impl.cc
+++ b/gr-qtgui/lib/fosphor_formatter_impl.cc
@@ -258,7 +258,7 @@ fosphor_formatter_impl::fosphor_formatter_impl(int fft_size,
       d_c2m(gr::blocks::complex_to_mag_squared::make(fft_size)),
       d_log(gr::blocks::nlog10_ff::make(calculate_log10_scale(scale),
                                         fft_size,
-                                        20 * std::log10(static_cast<float>(fft_size)))),
+                                       127)),
       d_f2byte(gnuradio::make_block_sptr<f2uchar_vfvb_impl>(fft_size)),
 
       d_avg(gr::filter::single_pole_iir_filter_ff::make(alpha, fft_size)),
```

willcode:
Backport notes:
- add type fixes from #6750 
- add #6676

willcode:
@gmeyerho When you get a chance ... it looks like the yml files will be installed even if the block is not enabled.

gmeyerho:
> @gmeyerho When you get a chance ... it looks like the yml files will be installed even if the block is not enabled.

Thanks for pointing this out, addressed in #6753

dl1ksv:
I get a strange behaviour when running the gnuradio test scripts.
Most of the time qa_qtgui_opengl_test fails. To isolate the problem I tried to run 
only this test repeatedly. That's what I get

~/gnuradiocomponents/gnuradio-volker/build$./gr-qtgui/python/qtgui/qa_qtgui_opengl_test.sh
.E
ERROR: test02 (__main__.test_qtgui_opengl.test02)
  File "/home/schroer/gnuradiocomponents/gnuradio-volker/gr-qtgui/python/qtgui/qa_qtgui_opengl.py", line 27, in test02
    self.qtsnk = qtgui.FosphorGlSink(1, 16, 512, 64, 1.0, 0.1, 0.99, 16.0, 1024.0)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/schroer/gnuradiocomponents/gnuradio-volker/gr-qtgui/python/qtgui/fosphorgl_sink.py", line 47, in __init__
    self.fosphor_sink = qtgui.fosphor_display(
                        ^^^^^^^^^^^^^^^^^^^^^^
MemoryError: std::bad_alloc

Ran 2 tests in 0.108s

FAILED (errors=1)
schroer@tux64mobil:~/gnuradiocomponents/gnuradio-volker/build$./gr-qtgui/python/qtgui/qa_qtgui_opengl_test.sh
.E
ERROR: test02 (__main__.test_qtgui_opengl.test02)
  File "/home/schroer/gnuradiocomponents/gnuradio-volker/gr-qtgui/python/qtgui/qa_qtgui_opengl.py", line 27, in test02
    self.qtsnk = qtgui.FosphorGlSink(1, 16, 512, 64, 1.0, 0.1, 0.99, 16.0, 1024.0)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/schroer/gnuradiocomponents/gnuradio-volker/gr-qtgui/python/qtgui/fosphorgl_sink.py", line 47, in __init__
    self.fosphor_sink = qtgui.fosphor_display(
                        ^^^^^^^^^^^^^^^^^^^^^^
MemoryError: std::bad_alloc

Ran 2 tests in 0.108s

FAILED (errors=1)
schroer@tux64mobil:~/gnuradiocomponents/gnuradio-volker/build$./gr-qtgui/python/qtgui/qa_qtgui_opengl_test.sh
..
Ran 2 tests in 0.113s

OK
schroer@tux64mobil:~/gnuradiocomponents/gnuradio-volker/build$

I don't recognise no pattern in failing or succeeding.
Any ideas?

I build gnuradio from source and test on F37 and F38

dl1ksv:
I just did a fresh build from the latest commit ( 7303d30a3ffec2c84ea92f4529cc5d793050fff7 ) and it seems that the occasional std::bad_alloc has gone.
Perhaps due to ca4424104212eba3cd7da54933cbf2bd3f13c5b1 ?

marcusmueller:
@dl1ksv are you on *BSD? Otherwise, you'd not been using the changed code in ca44241

dl1ksv:
@marcusmueller I tested on F37 and run the test 3 or 4 times. It never failed. Of course this isn't really a proof.  It may be just luck.

marcusmueller:
@dl1ksv got you :)

PR #6516, opened 2023-02-06
uhd: rfnoc: Add Fosphor block support

gmeyerho:
## Description
These commits enable the addition of the fosphor block support for rfnoc. 
The main commit is adding the actual Fosphor block support. 
One minor commit because the rx_streamer, when connected to the fosphor block, was not erroring on flush(), and that was a requirement for continued execution. I added a timeout to get around this, as flush() is only called when stopping execution of the block.
The other minor commit adds properties to the FFT block to allow for full fosphor support.

Example output using example from PR #6515 

image

Adds support for Fosphor block, adds properties to the FFT block

## Testing Done
I ran the example added with a known source, and I confirmed that the fosphor display showed the expected frequency.

## Checklist

### DIFF
--- a/gr-uhd/lib/rfnoc_rx_streamer_impl.cc
+++ b/gr-uhd/lib/rfnoc_rx_streamer_impl.cc
@@ -12,6 +12,7 @@
 
 #include "gr_uhd_common.h"
 #include "rfnoc_rx_streamer_impl.h"
+#include <gnuradio/high_res_timer.h>
 #include <gnuradio/io_signature.h>
 #include <uhd/convert.hpp>
 #include <uhd/rfnoc/node.hpp>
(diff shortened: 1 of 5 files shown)


willcode:
Changed to draft because it depends on another PR that is draft.

gmeyerho:
Moving out of draft as the PR this is associated with was also moved out of draft

mbr0wn (approved):

gmeyerho:
Removed the example from this PR and moved to #6515 so that this PR is now not blocked on merging

gr-uhd/lib/rfnoc_rx_streamer_impl.cc
```diff
@@ -178,7 +179,10 @@ void rfnoc_rx_streamer_impl::flush()
...
 
     const size_t itemsize = output_signature()->sizeof_stream_item(0);
-    while (true) {
+    // If we don't get an error, time out after 2 seconds
+    gr::high_res_timer_type end_time =
+        gr::high_res_timer_now() + gr::high_res_timer_tps() * 2;
+    while (gr::high_res_timer_now() < end_time) {
         d_streamer->recv(outputs, nbytes / itemsize / d_vlen, d_metadata, 0.0);
         if (d_metadata.error_code != ::uhd::rx_metadata_t::ERROR_CODE_NONE) {
             break;
```
mbr0wn, line 188:
```suggestion
            return;
```

Can you please add a log-warning after the `while` loop? I think this change is good, but I also think it's not expected/normal so it could help to notify people this is happening.

mbr0wn:
@willcode this is independent of #6515, can we merge this? One less thing in our todo list.

willcode:
Sure. I can backport it now too, if you like.

PR #6523, opened 2023-02-09
grc: remove duplicate import, move imports to template code

willcode:
## Description
In ada1f1c168 "from PyQt5 import Qt" was moved from individual yml files to the Python flowgraph template. This line also existed in the GRC Options block yml, resulting in the line being emitted twice.

This commit removes that duplicate code, and also moves several imports not related to options from the Options block to the template.

Built-in Option block, Python flowgraph mako template

## Checklist

### DIFF
--- a/grc/blocks/options.block.yml
+++ b/grc/blocks/options.block.yml
@@ -137,20 +137,6 @@ asserts:
 - ${ all(i >= 0 for i in placement) }
 
 templates:
-    imports: |-
-        from gnuradio import gr
-        from gnuradio.filter import firdes
-        from gnuradio.fft import window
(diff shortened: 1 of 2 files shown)


marcusmueller (approved):
:+1: 

Yeah, this seems better; less random snippets of Python strewn across the tree.

In general, I think we might want to move the `firdes` and `window` imports into the places they're actually used, but since that would quite likely break quite a few current OOTs' python generation, I think we don't need to touch this now.

willcode:
We can let this sit on main for a while. I don't think it causes any problems, but it's also not critical.

Issue #6526, opened 2023-02-13
grc: some filter blocks fail in building flowgrap


dl1ksv:
### What happened?

After apllying  #6523 filter blocks fail that use a window parameter due to a missing import statement that was removed in the option block.

### System Information

OS: FC37
GR Installation Method: Source

### GNU Radio Version

3.11-git (main)

### Specific Version

v3.11.0.0git-392-gf00dd77c

### Steps to Reproduce the Problem

Signal source -> Low Pass Filter -> Null Sink

This can be fixed by either
-  Adding an import block to the flowgraph with "from gnuradio.fft import window" by the user
-  Adding "from gnuradio.fft import window" to the import section in low_pass_filter.block.yml (This has to be done to each yml file the references a window type)
- Adding  again "from gnuradio.fft import window" to the template section in options.block.yml

What's the preferred way to fix this issue?
 

### Relevant log output

```shell
Param - Window(win):
	Value "window.WIN_HAMMING" cannot be evaluated:
	name 'window' is not defined
```

willcode:
I'll revert that PR. Seemed like a good idea. There are many mysteries hidden in GRC.

PR #6527, opened 2023-02-13
Revert "grc: remove duplicate import, move imports to template code"

willcode:
Reverts gnuradio/gnuradio#6523

The Option block is still a mystery to me. Reverting for now - we can figure out the best way to do this another time.

### DIFF
--- a/grc/blocks/options.block.yml
+++ b/grc/blocks/options.block.yml
@@ -137,6 +137,20 @@ asserts:
 - ${ all(i >= 0 for i in placement) }
 
 templates:
+    imports: |-
+        from gnuradio import gr
+        from gnuradio.filter import firdes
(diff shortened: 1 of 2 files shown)


Renlor:
Is this backported into 3.10?  I am seeing the same error as others have reported in https://github.com/gnuradio/gnuradio/issues/6526 and https://github.com/gnuradio/gnuradio/issues/6527 on the current 3.10.11.0 release.

Thanks!

Renlor:
@willcode Problem appears to predate #6523, I pulled in 3.10.5.1 which predates that merge unless the patch was included somehow other than through a merge into main.

willcode:
That issue was closed 18 months ago. Please create a new issue if you are having a problem. Thanks.

PR #6536, opened 2023-02-20
qtgui: fix QT GUI Histogram Sink produces implausible results #6494

### LINKED ISSUE #6494: QT GUI Histogram Sink produces implausible results
### What happened?

A sequence of 10 equidistant values should produce an even histogram (10 bins, 10 values) => 1 value per bin.

### System Information

OS: Ubuntu 22.04
GR Installation Method: native distro packages

### GNU Radio Version

3.10 (maint-3.10)

### Specific Version

3.10.1.1

1. Create flow graph:

test

10 different source values of same distance (0..9), 10 bins, number of points: 10.

2. Expected behaviour:

All 10 bins should have 1 as count. Distance between bins should be exactly 1.

3. Observed behaviour:

Some bins have more values, some have no values inside. Distribution not even. Distance between bins is not 1:
[0.0, 1.1, 2.2, 3.3, 4.4, 5.5, 6.6, 7.7, 8.8, 9.9, ??]

Screenshot_20230129_150941

### Relevant log output

_No response_

R-ohit-B-isht:
## Description
when the data range is very small, the leftmost bin can be too narrow, which can cause some data points to fall outside of the histogram.

## Related Issue
Fixes #6494 

gr-qtgui/lib/HistogramDisplayPlot.cc: line no. 192-198

## Checklist

### DIFF
--- a/gr-qtgui/lib/HistogramDisplayPlot.cc
+++ b/gr-qtgui/lib/HistogramDisplayPlot.cc
@@ -190,11 +190,11 @@ void HistogramDisplayPlot::_resetXAxisPoints(double left, double right)
         throw std::runtime_error("HistogramDisplayPlot::_resetXAxisPoints left and/or "
                                  "right values are invalid");
 
-    d_left = left * (1 - copysign(0.1, left));
-    d_right = right * (1 + copysign(0.1, right));
+    d_left = left;
+    d_right = right;
     d_width = (d_right - d_left) / (d_bins);
     for (unsigned int loc = 0; loc < d_bins; loc++) {
-        d_xdata[loc] = d_left + loc * d_width;
+        d_xdata[loc] = d_left + (loc + 0.5) * d_width; // center the value in each bin
     }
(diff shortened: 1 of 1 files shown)


willcode:
What does the plot look like with this change? Curious what [0,10] looks like. Does it plot the x axis as 0.5, 1.5, ..., 9.5?

R-ohit-B-isht:
> lot look like with this change? Curious what [0,10] looks like. Does it plot the x axis as 0.5, 1.5, ..., 9.5?

image

willcode:
Thanks. Would you also try an example with [-10,10] to how `lround()` in `plotNewData()` works with this change?

R-ohit-B-isht:
> 



> Thanks. Would you also try an example with [-10,10] to how `lround()` in `plotNewData()` works with this change?
like this?
image

willcode (approved):

mormj (approved):

PR #6537, opened 2023-02-21
fec: fix some bugs and refactor async_decoder

### LINKED ISSUE #6433: FEC Async Decoder overflows when converting from float to uint8_t

daniestevez:
## Description

This fixes a few bugs and shortcomings in `async_decoder` and refactors it to reduce code duplication by merging the `decode_unpacked()` and `decoded_packed()` functions into a single `decode()` function.

The problems solved are the following:

- `decode_packed()` wasn't considering the `diff` (Viterbi tail) properly, unlike `decode_unpacked()`. This caused undefined behaviour when working in packed output mode with the Viterbi decoder in terminated mode. Occasionally this would cause bit errors in the last bits of a packet. See #6445 for a very similar problem affecting `tagged_decoder`.

- `decode_packed()` didn't support multiple blocks for fixed-size decoders (unlike `decode_unpacked()`).

- In both `decode_unpacked()` and `decode_packed()`, the input conversion to `uint8_t` could overflow (see #6433). The behaviour has been changed to clipping, which is less harmful.

Taking into account that some of these problems affected `decode_packed()` and not `decode_unpacked()` because the two functions were quite similar but had some differences, it seems a good idea to reduce the code duplication by merging the two functions into one.

Additionally, there are these improvements:

- The opportunity to use `volk_32f_s32f_x2_convert_8u` (when it is released, see https://github.com/gnuradio/volk/pull/617) for the input conversion to `uint8_t` has been identified in a comment.

- When there is a decoder shift and no input conversion, the shift is applied with `volk_32f_s32f_add_32f` rather than with an element-by-element loop.

## Related Issue

Fixes #6433.

This affects the `async_decoder` block only.

## Testing Done

I have an in-house test case that replicates the problem with the `diff` (Viterbi tail) by failing to decode correctly good packets (At least when using the SSE3 spiral kernel. With the generic kernel I haven't been able to find a test case that does the same). After applying these modifications, the test now passes. I can't share this test, but if necessary I'll see if I can cobble together something similar that I can publish.

Since there aren't any QA tests for `async_decoder`, just in case I broke anything inadvertently, I have run the [gr-satellites AO-73 decoder](https://github.com/daniestevez/gr-satellites/blob/main/python/components/deframers/ao40_fec_deframer.py), which uses `async_decoder` in unpacked mode for Viterbi decoding. This works as expected.

## Checklist

- ~[ ] I have updated [the documentation](https://wiki.gnuradio.org/index.php/Main_Page#Documentation) where necessary.~  -> Not applicable.
- [x] ~I have added tests to cover my changes,~ and all previous tests pass. -> I haven't added new tests. It isn't trivial to come up with tests that check some of the problems addressed here (plus it would be desirable to have some basic tests for `async_decoder` first).

### DIFF
--- a/gr-fec/lib/async_decoder_impl.cc
+++ b/gr-fec/lib/async_decoder_impl.cc
@@ -51,12 +51,7 @@ async_decoder_impl::async_decoder_impl(generic_decoder::sptr my_decoder,
     message_port_register_in(d_in_port);
     message_port_register_out(d_out_port);
 
-    if (d_packed) {
-        set_msg_handler(d_in_port, [this](pmt::pmt_t msg) { this->decode_packed(msg); });
-    } else {
-        set_msg_handler(d_in_port,
-                        [this](pmt::pmt_t msg) { this->decode_unpacked(msg); });
-    }
+    set_msg_handler(d_in_port, [this](const pmt::pmt_t& msg) { this->decode(msg); });
 
     // The maximum frame size is set by the initial frame size of the decoder.
     d_max_bits_in = d_mtu * 8 * 1.0 / d_decoder->rate();
@@ -74,7 +69,26 @@ async_decoder_impl::async_decoder_impl(generic_decoder::sptr my_decoder,
 
 async_decoder_impl::~async_decoder_impl() {}
 
-void async_decoder_impl::decode_unpacked(pmt::pmt_t msg)
+// TODO: replace this function by volk_32f_s32f_x2_convert_8u when it gets into
+// a stable volk release
+inline void async_decoder_impl::convert_32f_to_8u(uint8_t* output_vector,
+                                                  const float* input_vector,
+                                                  const float scale,
+                                                  const float bias,
+                                                  unsigned int num_points)
+{
+    volk_32f_s32f_multiply_32f(d_tmp_f32.data(), input_vector, scale, num_points);
+    if (bias != 0.0) {
+        for (size_t n = 0; n < num_points; n++) {
+            d_tmp_f32[n] += bias;
+        }
+    }
+    for (size_t n = 0; n < num_points; n++) {
+        output_vector[n] = std::clamp<float>(d_tmp_f32[n], 0.0f, UINT8_MAX);
+    }
+}
+
+void async_decoder_impl::decode(const pmt::pmt_t& msg)
 {
     // extract input pdu
     pmt::pmt_t meta(pmt::car(msg));
@@ -110,89 +124,47 @@ void async_decoder_impl::decode_unpacked(pmt::pmt_t msg)
         }
     }
 
-    size_t o0(0);
-    const float* f32in = pmt::f32vector_elements(bits, o0);
-    pmt::pmt_t outvec(pmt::make_u8vector(nbits_out, 0x00));
-    uint8_t* u8out = pmt::u8vector_writable_elements(outvec, o0);
+    size_t vector_len = 0; // used only to get the length of PMT vectors
+    const float* f32in = pmt::f32vector_elements(bits, vector_len);
+    const size_t nitems_out = d_packed ? nbits_out / 8 : nbits_out;
+    auto outvec = pmt::make_u8vector(nitems_out, 0x00);
+    uint8_t* u8out = pmt::u8vector_writable_elements(outvec, vector_len);
+    uint8_t* decoder_out = d_packed ? d_bits_out.data() : u8out;
+    const float shift = d_decoder->get_shift();
 
-    if (strncmp(d_decoder->get_input_conversion(), "uchar", 5) == 0) {
-        volk_32f_s32f_multiply_32f(d_tmp_f32.data(), f32in, 48.0f, nbits_in);
-    } else {
-        memcpy(d_tmp_f32.data(), f32in, nbits_in * sizeof(float));
-    }
+    if (std::string_view(d_decoder->get_input_conversion()) == "uchar") {
+        convert_32f_to_8u(d_tmp_u8.data(), f32in, 48.0f, shift, nitems_out);
 
-    if (d_decoder->get_shift() != 0) {
-        for (size_t n = 0; n < nbits_in; n++)
-            d_tmp_f32[n] += d_decoder->get_shift();
-    }
-
-    if (strncmp(d_decoder->get_input_conversion(), "uchar", 5) == 0) {
-        // volk_32f_s32f_convert_8i(d_tmp_u8, d_tmp_f32, 1, nbits_in);
-        for (size_t n = 0; n < nbits_in; n++)
-            d_tmp_u8[n] = static_cast<uint8_t>(d_tmp_f32[n]);
-
-        d_decoder->generic_work((void*)d_tmp_u8.data(), (void*)u8out);
-    } else {
         for (size_t i = 0; i < nblocks; i++) {
-            d_decoder->generic_work((void*)&d_tmp_f32[i * d_decoder->get_input_size()],
-                                    (void*)&u8out[i * d_decoder->get_output_size()]);
+            d_decoder->generic_work(
+                (void*)&d_tmp_u8[i * d_decoder->get_input_size()],
+                (void*)&decoder_out[i * d_decoder->get_output_size()]);
+        }
+    } else {
+        if (shift != 0.0) {
+            volk_32f_s32f_add_32f(d_tmp_f32.data(), f32in, shift, nbits_in);
+        } else {
+            memcpy(d_tmp_f32.data(), f32in, nbits_in * sizeof(float));
+        }
+
+        for (size_t i = 0; i < nblocks; i++) {
+            d_decoder->generic_work(
+                (void*)&d_tmp_f32[i * d_decoder->get_input_size()],
+                (void*)&decoder_out[i * d_decoder->get_output_size()]);
         }
     }
 
-    static const pmt::pmt_t iterations_key = pmt::mp("iterations");
-    meta = pmt::dict_add(meta, iterations_key, pmt::mp(d_decoder->get_iterations()));
-    message_port_pub(d_out_port, pmt::cons(meta, outvec));
-}
+    // Note: when nblocks > 1, this only gets the iterations for the last
+    // decoder block.
+    meta = pmt::dict_add(meta, ITERATIONS_KEY, pmt::mp(d_decoder->get_iterations()));
 
-void async_decoder_impl::decode_packed(pmt::pmt_t msg)
-{
-    // extract input pdu
-    pmt::pmt_t meta(pmt::car(msg));
-    pmt::pmt_t bits(pmt::cdr(msg));
-
-    size_t o0 = 0;
-    size_t nbits_in = pmt::length(bits);
-    int nbits_out = nbits_in * d_decoder->rate();
-    int nbytes_out = nbits_out / 8;
-
-    // Check here if the frame size is larger than what we've
-    // allocated for in the constructor.
-    if (nbits_in > d_max_bits_in) {
-        throw std::runtime_error(
-            "async_decoder: Received frame larger than max frame size.");
+    if (d_packed) {
+        if (d_rev_pack) {
+            d_pack.pack_rev(u8out, decoder_out, nitems_out);
+        } else {
+            d_pack.pack(u8out, decoder_out, nitems_out);
+        }
     }
-
-    d_decoder->set_frame_size(nbits_out);
-
-    pmt::pmt_t outvec(pmt::make_u8vector(nbytes_out, 0x00));
-    uint8_t* bytes_out = pmt::u8vector_writable_elements(outvec, o0);
-    const float* f32in = pmt::f32vector_elements(bits, o0);
-
-    if (strncmp(d_decoder->get_input_conversion(), "uchar", 5) == 0) {
-        volk_32f_s32f_multiply_32f(d_tmp_f32.data(), f32in, 48.0f, nbits_in);
-    } else {
-        memcpy(d_tmp_f32.data(), f32in, nbits_in * sizeof(float));
-    }
-
-    if (d_decoder->get_shift() != 0) {
-        for (size_t n = 0; n < nbits_in; n++)
-            d_tmp_f32[n] += d_decoder->get_shift();
-    }
-
-    if (strncmp(d_decoder->get_input_conversion(), "uchar", 5) == 0) {
-        // volk_32f_s32f_convert_8i(d_tmp_u8, d_tmp_f32, 1.0, nbits_in);
-        for (size_t n = 0; n < nbits_in; n++)
-            d_tmp_u8[n] = static_cast<uint8_t>(d_tmp_f32[n]);
-
-        d_decoder->generic_work((void*)d_tmp_u8.data(), (void*)d_bits_out.data());
-    } else {
-        d_decoder->generic_work((void*)d_tmp_f32.data(), (void*)d_bits_out.data());
-    }
-
-    if (d_rev_pack)
-        d_pack.pack_rev(bytes_out, d_bits_out.data(), nbytes_out);
-    else
-        d_pack.pack(bytes_out, d_bits_out.data(), nbytes_out);
 
     message_port_pub(d_out_port, pmt::cons(meta, outvec));
 }
--- a/gr-fec/lib/async_decoder_impl.h
+++ b/gr-fec/lib/async_decoder_impl.h
@@ -18,6 +18,8 @@
 namespace gr {
 namespace fec {
 
+static const pmt::pmt_t ITERATIONS_KEY = pmt::mp("iterations");
+
 class FEC_API async_decoder_impl : public async_decoder
 {
 private:
@@ -34,11 +36,16 @@ private:
 
     size_t d_max_bits_in;
     volk::vector<float> d_tmp_f32;
-    volk::vector<int8_t> d_tmp_u8;
+    volk::vector<uint8_t> d_tmp_u8;
     volk::vector<uint8_t> d_bits_out;
 
-    void decode_packed(pmt::pmt_t msg);
-    void decode_unpacked(pmt::pmt_t msg);
+    void decode(const pmt::pmt_t& msg);
+
+    inline void convert_32f_to_8u(uint8_t* output_vector,
+                                  const float* input_vector,
+                                  const float scale,
+                                  const float bias,
+                                  unsigned int num_points);
 
 public:
     async_decoder_impl(generic_decoder::sptr my_decoder,


gr-fec/lib/async_decoder_impl.cc
```diff
@@ -112,88 +107,66 @@ void async_decoder_impl::decode_unpacked(pmt::pmt_t msg)
...
+        for (size_t n = 0; n < nbits_in; n++) {
+            const float in = d_tmp_f32[n];
+            uint8_t* out = &d_tmp_u8[n];
+            if (in > max_val) {
+                *out = UINT8_MAX;
+            } else if (in < min_val) {
+                *out = 0;
+            } else {
+                *out = static_cast<uint8_t>(in);
+            }
```
marcusmueller, line 136: loop content could be replaced by
`d_tmp_u8[n] = std::clamp<float>(d_tmp_f32[n], 0.0f, UINT8_MAX);`

gr-fec/lib/async_decoder_impl.cc
```diff
@@ -112,88 +107,66 @@ void async_decoder_impl::decode_unpacked(pmt::pmt_t msg)
...
+        if (shift != 0.0) {
+            for (size_t n = 0; n < nbits_in; n++) {
+                d_tmp_f32[n] += shift;
+            }
+        }
+        const float min_val = 0.0f;
+        const float max_val = UINT8_MAX;
+        for (size_t n = 0; n < nbits_in; n++) {
+            const float in = d_tmp_f32[n];
+            uint8_t* out = &d_tmp_u8[n];
```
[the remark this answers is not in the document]
marcusmueller: (makes no difference, computationally, but we wouldn't need the dereferencing in `*out = …` below)

gr-fec/lib/async_decoder_impl.cc
```diff
@@ -51,12 +51,7 @@ async_decoder_impl::async_decoder_impl(generic_decoder::sptr my_decoder,
     message_port_register_in(d_in_port);
     message_port_register_out(d_out_port);
 
-    if (d_packed) {
-        set_msg_handler(d_in_port, [this](pmt::pmt_t msg) { this->decode_packed(msg); });
-    } else {
-        set_msg_handler(d_in_port,
-                        [this](pmt::pmt_t msg) { this->decode_unpacked(msg); });
-    }
+    set_msg_handler(d_in_port, [this](pmt::pmt_t msg) { this->decode(msg); });
```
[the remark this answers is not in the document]
daniestevez: Today I learned that `pmt_t` is actually a `shared_ptr`. That suggestion makes a lot of sense. Done.

gr-fec/lib/async_decoder_impl.cc
```diff
@@ -74,7 +69,7 @@ async_decoder_impl::async_decoder_impl(generic_decoder::sptr my_decoder,
 
 async_decoder_impl::~async_decoder_impl() {}
 
-void async_decoder_impl::decode_unpacked(pmt::pmt_t msg)
+void async_decoder_impl::decode(pmt::pmt_t msg)
```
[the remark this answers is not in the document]
daniestevez: Done.

gr-fec/lib/async_decoder_impl.cc
```diff
@@ -112,88 +107,66 @@ void async_decoder_impl::decode_unpacked(pmt::pmt_t msg)
...
     size_t o0(0);
     const float* f32in = pmt::f32vector_elements(bits, o0);
-    pmt::pmt_t outvec(pmt::make_u8vector(nbits_out, 0x00));
+    const size_t nitems_out = d_packed ? nbits_out / 8 : nbits_out;
+    pmt::pmt_t outvec(pmt::make_u8vector(nitems_out, 0x00));
     uint8_t* u8out = pmt::u8vector_writable_elements(outvec, o0);
+    uint8_t* decoder_out = d_packed ? d_bits_out.data() : u8out;
+    const float shift = d_decoder->get_shift();
 
     if (strncmp(d_decoder->get_input_conversion(), "uchar", 5) == 0) {
```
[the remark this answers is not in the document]
daniestevez: I also didn't like this `strncmp` when I saw it. Besides the `"ucharacterless" == "uchar"`, I found more annoying that it's not very readable.

I think it makes sense to change this now, instead of opening a new PR just to change this. I have replaced this by:
`std::string_view(d_decoder->get_input_conversion()) == "uchar"`
It seems to work as intended (note that `get_input_conversion()` returns a `char *`).

gr-fec/lib/async_decoder_impl.cc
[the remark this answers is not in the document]
daniestevez: This makes sense. It seems better C++ style. However, this code has now disappeared (see below).

gr-fec/lib/async_decoder_impl.cc
```diff
@@ -112,88 +107,66 @@ void async_decoder_impl::decode_unpacked(pmt::pmt_t msg)
...
+        // TODO: replace all this by volk_32f_s32f_x2_convert_8u when it gets
+        // into a stable volk release
```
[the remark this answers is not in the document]
daniestevez: https://github.com/gnuradio/gnuradio/issues/6545

daniestevez:
Thanks for the review @marcusmueller, I addressed all your comments and force-pushed.

marcusmueller (approved):

gr-fec/lib/async_decoder_impl.cc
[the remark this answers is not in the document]
jdemel: I'd like to add to the discussion with this:
https://github.com/gnuradio/gnuradio/blob/54d199a2a3d50faeb1bb5c9b1a9f723d479a9c11/gnuradio-runtime/include/pmt/pmt.h#L641
In this example, every `dict_add` operation creates a copy of the dictionary. I couldn't find proof but as far as I remember, PMTs are basically `const` objects. Thus, every operation on them creates a copy.

gr-fec/lib/async_decoder_impl.cc
```diff
@@ -112,88 +107,56 @@ void async_decoder_impl::decode_unpacked(pmt::pmt_t msg)
 
     size_t o0(0);
```
jdemel, line 108: I know it's been there. But could you rename this variable? The name bears no meaning, is easily confused with a literal. It looks a lot like someone read [this](https://github.com/Droogans/unmaintainable-code) or smth similar.

gr-fec/lib/async_decoder_impl.cc
```diff
@@ -112,88 +107,56 @@ void async_decoder_impl::decode_unpacked(pmt::pmt_t msg)
...
-    if (d_decoder->get_shift() != 0) {
-        for (size_t n = 0; n < nbits_in; n++)
-            d_tmp_f32[n] += d_decoder->get_shift();
-    }
-
-    if (strncmp(d_decoder->get_input_conversion(), "uchar", 5) == 0) {
-        // volk_32f_s32f_convert_8i(d_tmp_u8, d_tmp_f32, 1, nbits_in);
-        for (size_t n = 0; n < nbits_in; n++)
-            d_tmp_u8[n] = static_cast<uint8_t>(d_tmp_f32[n]);
+        if (shift != 0.0) {
```
jdemel, line 120: I was wondering why a `shift` variable is of type `float`. Looing at L114, I assume the name `bias` would've been more fitting.
I'm torn between updating the variable name and leaving it as is because of consistency.

gr-fec/lib/async_decoder_impl.cc
```diff
@@ -112,88 +107,56 @@ void async_decoder_impl::decode_unpacked(pmt::pmt_t msg)
...
+            for (size_t n = 0; n < nbits_in; n++) {
+                d_tmp_f32[n] += shift;
+            }
+        }
+        for (size_t n = 0; n < nbits_in; n++) {
+            d_tmp_u8[n] = std::clamp<float>(d_tmp_f32[n], 0.0f, UINT8_MAX);
+        }
```
jdemel, line 127: I suggest to move all these lines to a new function that resembles the `volk_32f_s32f_x2_convert_u8` signature. Makes it easier to understand and easier to update in the future. e.g. it'd be possible to just replace this function with its VOLK version in the future.

gr-fec/lib/async_decoder_impl.cc
```diff
@@ -112,88 +107,56 @@ void async_decoder_impl::decode_unpacked(pmt::pmt_t msg)
...
-                                    (void*)&u8out[i * d_decoder->get_output_size()]);
+            d_decoder->generic_work(
+                (void*)&d_tmp_f32[i * d_decoder->get_input_size()],
+                (void*)&decoder_out[i * d_decoder->get_output_size()]);
         }
     }
 
+    // Note: when nblocks > 1, this only gets the iterations for the last
+    // decoder block.
     static const pmt::pmt_t iterations_key = pmt::mp("iterations");
```
jdemel, line 150: Could you make this a class constant? Or a namespace constant? Other code in GR handles this kind of constants this way. It'd be useful for people who want to figure out how to deal with this info.

gr-fec/lib/async_decoder_impl.cc
```diff
@@ -112,88 +107,56 @@ void async_decoder_impl::decode_unpacked(pmt::pmt_t msg)
...
-
-        d_decoder->generic_work((void*)d_tmp_u8.data(), (void*)d_bits_out.data());
-    } else {
-        d_decoder->generic_work((void*)d_tmp_f32.data(), (void*)d_bits_out.data());
+    if (d_packed) {
+        if (d_rev_pack)
+            d_pack.pack_rev(u8out, decoder_out, nitems_out);
+        else
+            d_pack.pack(u8out, decoder_out, nitems_out);
     }
```
jdemel, line 158: Would you mind to add `{..}` around the `if` and `else` branches? I know this is correct but it is also error prone. Wouldn't be the first time, I'd shoot myself in the foot with this.

gr-fec/lib/async_decoder_impl.h
```diff
@@ -34,11 +34,10 @@ class FEC_API async_decoder_impl : public async_decoder
 
     size_t d_max_bits_in;
     volk::vector<float> d_tmp_f32;
-    volk::vector<int8_t> d_tmp_u8;
+    volk::vector<uint8_t> d_tmp_u8;
     volk::vector<uint8_t> d_bits_out;
 
-    void decode_packed(pmt::pmt_t msg);
-    void decode_unpacked(pmt::pmt_t msg);
+    void decode(const pmt::pmt_t& msg);
```
jdemel, line 40: Following the discussion here, a `pmt::pmt_t` is a `shared_ptr`.
I suggest to change the signature to `void decode(const pmt::pmt_t msg);`. This change would make the signature a bit more obvious.

jdemel (changes requested):
@daniestevez thanks for cleaning this up. I hope that all my comments are just simple minor things.

gr-fec/lib/async_decoder_impl.cc
daniestevez: Yes. In hindsight `bias` seems a better name than `shift`. However, this value comes from `d_decoder->get_shift()`, so unless we want to rename that to `get_bias()` in all the decoders, I think it's better to call the variable `shift`.

daniestevez: On the other hand, due to applying https://github.com/gnuradio/gnuradio/pull/6537#discussion_r1115398262, this variable can disappear (we can put the call to `d_decoder->get_shift()` directly as a function parameter), so matter solved.

gr-fec/lib/async_decoder_impl.h
daniestevez: This is somewhat above my knowledge of copy vs. move in C++, so asking to make sure: if the signature was `void decode(const pmt::pmt_t msg)` instead of `void decode(const pmt::pmt_t& msg)`, wouldn't that cause a copy of the `pmt_t` object? @marcusmueller was concerned that a `void decode(pmt::pmt_t msg)` would cause a needless copy (i.e., increase of the refcount of the `shared_ptr`). Does adding the `const` qualifier already elide the copy?

gr-fec/lib/async_decoder_impl.cc
daniestevez: I also didn't like this, but for some reason I decide not to rename variables too much. I've renamed this to `vector_len` and added a comment about what this variable is for, because the only reason we're using it is because `f32vector_elements` and the like require to output the vector length in a variable. We don't use the value.

gr-fec/lib/async_decoder_impl.cc
daniestevez: I've moved this to an `inline` method called `convert_32f_to_8u`, with the same parameters as the Volk kernel. This needs to be a method instead of a standalone function because it needs to access `d_tmp_f32`.

gr-fec/lib/async_decoder_impl.cc
daniestevez: I've moved this to the `.h` inside the `namespace fec`. I tried to move it into the `private:` for the class declaration and this doesn't work (according to [this](https://www.stroustrup.com/bs_faq2.html#in-class) it seems this will only work for integral types).

I've also changed the name to all caps (`ITERATIONS_KEY`) for consistency with some examples in gr-soapy and gr-uhd. However, I've noticed that these examples declare the keys in the `.h` outside of the namespace. If this would also be preferred here, let me know and I'll change it.

gr-fec/lib/async_decoder_impl.cc
daniestevez: Done. I always write the braces. I think I didn't here because I was moving existing code around.

gr-fec/lib/async_decoder_impl.cc
```diff
@@ -112,88 +107,56 @@ void async_decoder_impl::decode_unpacked(pmt::pmt_t msg)
 
     size_t o0(0);
     const float* f32in = pmt::f32vector_elements(bits, o0);
-    pmt::pmt_t outvec(pmt::make_u8vector(nbits_out, 0x00));
+    const size_t nitems_out = d_packed ? nbits_out / 8 : nbits_out;
+    pmt::pmt_t outvec(pmt::make_u8vector(nitems_out, 0x00));
```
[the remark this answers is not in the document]
daniestevez: Done, using the `auto` alternative.

gr-fec/lib/async_decoder_impl.h
jdemel: I just [checked the documentation `typedef std::shared_ptr<pmt_base> pmt::pmt_t`](https://www.gnuradio.org/doc/doxygen/namespacepmt.html#a94a208b58dc16675b02674c28784d88b).

It does require some refcount accounting. However, [SO discussions on the subject](https://stackoverflow.com/a/8385581) basically point to our example and tell us: "Use the `shared_ptr` not the reference".

gr-fec/lib/async_decoder_impl.cc
jdemel: I had these examples in mind. Intuitively, I'd expect the gr-uhd and gr-soapy keys in their namespaces. But there may be other considerations as well.

gr-fec/lib/async_decoder_impl.h
daniestevez: The comments on that stackoverflow to me say:
* Pass a `shared_ptr` by value, then a copy is made. Pro: the copy keeps the allocation alive. Con: the copy is slightly costly because it involves increasing the refcounter atomically. They don't say what happens if you pass by `const` value, so I guess it's the same?
* Pass a `shared_ptr` as a `const` reference, then a copy is not made. Pro: it is cheap. Con: if whoever owns the `shared_ptr` deallocates the object, then your reference is dangling.

In this case I think that the caller to `decode()` owns the `shared_ptr`, and will keep it alive while `decode()` runs. Therefore, it seems harmless to pass as a `const` reference, and it is slightly cheaper. I think this is what @marcusmueller had in mind (please confirm).

willcode:
Is everyone happy with this PR? Will merge, if so.

daniestevez:
I think it's good to go. The only thing that seems to be open is whether to pass PMTs around as `const pmt::pmt_t&`  or `const pmt::pmt_t`. Currently this uses the former.

willcode:
Since it's internal, we can always change the const vs const ref later.

PR #6538, opened 2023-02-22
blocks: update UI for new throttle

willcode:
Minor UI-oriented changes for #5670

### DIFF
--- a/gr-blocks/grc/blocks_throttle2.block.yml
+++ b/gr-blocks/grc/blocks_throttle2.block.yml
@@ -1,5 +1,5 @@
-id: blocks_trickling_throttle
-label: Throttle (limited blocking)
+id: blocks_throttle2
+label: Throttle
 flags: [ throttle, python, cpp ]
 
 parameters:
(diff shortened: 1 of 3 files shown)


gr-blocks/grc/blocks_throttle2.block.yml
```diff
@@ -25,16 +25,16 @@ parameters:
...
+    options: ['auto', 'time', 'items']
+    option_labels: ["None", "Max Time (sec)", "Max Items"]
 -   id: maximum
-    label: 'Max. wait[s] or items'
+    label: 'Maximum'
     dtype: real
-    hide: ${ ('part' if duration_or_samples == 'neither' else 'none') }
-    default: 0.1
+    hide: ${ ('all' if limit == 'auto' else 'none') }
+    default: 1
```
marcusmueller, line 37: I can definitely live with the rest, but this new default of a 1s-chunkiness seems like an unwise change; the idea was that things like Qt Time sink look "smooth".

marcusmueller (changes requested):

gr-blocks/grc/blocks_throttle2.block.yml
willcode: I know you originally had 2 different values and I though they should be the same field :smile: 
OK, `0.1` is fine, since most people will care about time, not samples. There's really no right value for samples anyway.

marcusmueller:
I'm fine with it in this shape. While I admit `trickling_throttle` is a bit unwieldy, `throttle2` might be a bad choice if we ever decide to deprecate and remove `throttle`; what was the motivation here? Maybe we shouldn't call it "Throttle" alltogether – `limit_throughput` might be an alternative?

PR #6544, opened 2023-02-22
gnuradio-runtime: Remove debug messages.

drmpeg:


willcode:
Would changing the pagesize one to "debug" be sufficient?

On the scheduler one, I would think it should either always print the selected schedule (this seems useful where custom schedulers are used) or never.

drmpeg:
By default, "debug" is enabled. I know it can be changed, but I always forget how to do it. Also, for a given system, page size never changes, Seeing the same value over and over again is just spam. Same for the default scheduler.

If a message for a custom scheduler is desired, then print it when a custom scheduler is selected (which for the majority of users is never).

willcode:
We really should have the default as "info", or the debug logger default to off, or whatever it takes so this kind of stuff can be seen when desired.

marcusmueller:
I'd agree that with effectively at most two different page sizes existing in the wild, the page size message should probably demoted to a debug message. I also agree with Jeff on that if we log that the default scheduler has been selected, we should also log when it isn't.

I also agree that these messages shouldn't pester the user by default. Now, does that call for removal of potentially useful debug output, or for a increase of the default log level to info? I'd think the latter!

@drmpeg  if we increased the default log level to info, would that be an OK solution from your perspective?

drmpeg:
I am fine with increasing the default log level to info.

willcode:
Turns out the current log level was recently changed to "info".

drmpeg:
Doesn't seem to work. I'm still seeing the ```logger->debug("Using default scheduler \"{}\"", name);``` being printed.

willcode:
Check `~/.gnuradio/config.conf`. Mine reads:
```
...
[log]
log_level = info
debug_level = info
...
```
It also works as expected if there's no entry at all. If the `info` is changed to `debug` that message shows up.

drmpeg:
I don't have a ```~/.gnuradio/config.conf``` file at all.

willcode:
It will also load any other `*.conf` file from the prefs directory `{prefix}/etc/gnuradio/conf.d/*.conf`, and `gnuradio-runtime/gr_log_default.conf` still has it set as `debug`, meaning that the default in the code is never used on a typical installation. OK, we need to change that to `info`.

drmpeg:
Okay. I'm gonna close this PR.

drmpeg:
I did a fresh install, and my ```{prefix}/etc/gnuradio/conf.d/``` looks like this:
```
00-grc-docs.conf       gr-audio.conf      gr_log_default.conf
gnuradio-runtime.conf  gr-audio-oss.conf  gr-qtgui.conf
gr-audio-alsa.conf     grc.conf           modtool.conf
```

PR #6546, opened 2023-02-22
runtime: change default log level in conf file to info

willcode:
## Description
The default log level in the code was previously changed to "info" but an installed conf file overrode it to be "debug". Also change the pagesize notice to debug instead of info.

## Related Issue
Discussion in #6544 

Runtime

## Checklist


marcusmueller:
I must to my own horror admit that I've not run the test suite back when working on the log chains with various log levels set in the test-affecting configuration files. If the test suite now fails, I owe you a drink or something.

drmpeg (approved):
Works good with and without ```~/.gnuradio/config.conf```.

PR #6566, opened 2023-03-06
blocks: delay hide message port optionally in grc

mormj:
## Description
Give the delay block the ability to hide the `dly` message port in grc

Ideally, a block with single in, single out and hidden message port should be able to be bypassed, but that is a separate issue

after:
image
image

image
image

delay block

## Testing Done
Tried in GRC 

## Checklist

### DIFF
--- a/gr-blocks/grc/blocks_delay.block.yml
+++ b/gr-blocks/grc/blocks_delay.block.yml
@@ -25,7 +25,14 @@ parameters:
     dtype: int
     default: '1'
     hide: ${ 'part' if vlen == 1 else 'none' }
-
+-   id: showports
+    label: Show Msg Ports
+    dtype: bool
+    default: 'True'
+    options: ['False', 'True']
+    option_labels: ['Yes', 'No']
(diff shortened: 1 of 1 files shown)


willcode (approved):
The reversed logic is a little confusing, but I can't think of a less confusing way to do it easily.

marcusmueller (approved):

Notou:
@willcode 
Could the reversal be made in line 44, with     `hide: ${ not showports }` instead of the enum options?

mormj:
> The reversed logic is a little confusing, but I can't think of a less confusing way to do it easily.

Yeah I just copied it directly from how it is done in other blocks without putting any thought into it

willcode:
Swapped the logic.

willcode:
@mormj could you verify this does the same thing as your original code? It's simple enough, but I don't want to merge before you look at it.

mormj:
Yes, this works just the same - thank you for putting in the simplified logic!

PR #6579, opened 2023-03-16
gr-zmq: flush output in ZMQ stream sources

daniestevez:
## Description

This modifies the policy that the ZMQ stream sources (ZMQ PULL Source, ZMQ REQ Source and ZMQ SUB Source) use to decide when to wait for a new ZMQ message or when to return immediately from work().

Currently, the work() function either:
* Already has enough data from a previous message to fill all the output buffer, and so it returns without waiting for more messages.
* If there is not enough data to fill all the output buffer, then the buffer is partially filled with all the data available and the function waits to receive a new ZMQ message (using it to try to fill the rest of the buffer).

This policy has the following drawback: if the timeout parameter is set to a large value (or -1, which means wait forever), then if the end of a message does not fill the output buffer completely, the work() function will wait to receive a new message instead of returning to the scheduler immediately the rest of the current message.

This is a problem in this use case:
* Messages as received sparingly.
* The timeout is set to -1 or a value larger than the maximum period between messages in order to avoid returning from work() without any output, which introduces a latency on the order of 100ms for the next time that work() will be called.
* The message data should be delivered to downstream blocks as soon as available (this makes most sense if using these blocks to convert messages into a tagged stream, by ussing pass_tags = true).

The modified policy only tries to wait for a ZMQ message if we have not produced any output in this work() call. If at the beginning of the work() call we have pending data, we fill the output buffer with the data we have, and then try to receive more ZMQ messages without waiting before returning.

## Related Issue

No related issue.

ZMQ stream sources: ZMQ PULL Source, ZMQ REQ Source and ZMQ SUB Source.

## Testing Done

Tested the changes with a flowgraph that follows the use case described above on a ZMQ REQ Source with a timeout of -1. Before the changes, the ZMQ REQ Source will never deliver the last bytes of the message to the next block until a new ZMQ arrived. This changes solve the problem.

## Checklist

- [x] ~I have updated [the documentation](https://wiki.gnuradio.org/index.php/Main_Page#Documentation) where necessary~ No changes needed.
- [ ] I have added tests to cover my changes, and all previous tests pass. I haven't added tests. Let me know if a test for this specific case would be required.

### DIFF
--- a/gr-zeromq/lib/pull_source_impl.cc
+++ b/gr-zeromq/lib/pull_source_impl.cc
@@ -54,12 +54,15 @@ int pull_source_impl::work(int noutput_items,
             /* No more space ? */
             if (done == noutput_items)
                 break;
+
+            /* Have some output to return: do not wait for more messages */
+            first = false;
         } else {
             /* Try to get the next message */
             if (!load_message(first))
                 break; /* No message, we're done for now */
 
-            /* Not the first anymore */
+            /* Not the first anymore: do not wait for more messages */
             first = false;
         }
     }
--- a/gr-zeromq/lib/req_source_impl.cc
+++ b/gr-zeromq/lib/req_source_impl.cc
@@ -57,6 +57,9 @@ int req_source_impl::work(int noutput_items,
             /* No more space ? */
             if (done == noutput_items)
                 break;
+
+            /* Have some output to return: do not wait for more messages */
+            first = false;
         } else {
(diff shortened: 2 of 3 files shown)


daniestevez:
Force-pushed to fix clang-format.

willcode (approved):
I like this new behavior. Since a ZMQ sender is sending messages, they would expect them to be processed without the receiver waiting for further messages.

PR #6582, opened 2023-03-17
fec: fix regression in async_decoder and add tests

daniestevez:
## Description

The `async_decoder` was broken in 4e50c99ab38f816d56145133ce6b7cc68e164e90. The reason was a mistake when rewriting the PR to address comments. In the case where the decoder input needs to be converted to `uchar`, `nitems_out` rather than `nbits_in` was wrongly used as the length of the data to be converted. Effectively, we are passing only the first half of the input to a rate=1/2 decoder such as the CC decoder.

Since there are no tests of the async_decoder that would have caught the bug, this adds some tests. A new `_qa_helper_async` `top_block` similar to `_qa_helper` is added. This is used in an additional test case in the `qa_fecapi_* `tests to run the `async_encoder` and `async_decoder` with all the different FECAPI encoders/decoders.

The Async Decoder FECAPI block. This also adds tests for Async Encoder.

## Testing Done

The new tests pass when the bug is fixed. They don't pass if the bug is not fixed.

## Checklist

- [x] ~I have updated [the documentation](https://wiki.gnuradio.org/index.php/Main_Page#Documentation) where necessary.~ => No necessary changes.

### DIFF
--- a/gr-fec/python/fec/qa_fecapi_cc.py
+++ b/gr-fec/python/fec/qa_fecapi_cc.py
@@ -13,6 +13,7 @@ from gnuradio import gr, gr_unittest
 from gnuradio import fec
 
 from _qa_helper import _qa_helper
+from _qa_helper_async import _qa_helper_async
 
 
 class test_fecapi_cc(gr_unittest.TestCase):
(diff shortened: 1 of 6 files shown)


willcode (approved):

willcode:
I'll merge soon so we can get this into the March v3.10.6.0. Looks like an overflow without the change.

gr-fec/python/fec/qa_fecapi_cc.py
```diff
@@ -191,6 +192,28 @@ def test_parallelism1_05(self):
 
         self.assertEqual(data_in, data_out)
 
+    def test_async_00(self):
+        frame_size = 30
+        k = 7
+        rate = 2
+        polys = [109, 79]
+        for mode in [fec.CC_TRUNCATED, fec.CC_TERMINATED, fec.CC_TAILBITING]:
```
jdemel, line 200: You may want to have a look at [unittest subtest](https://docs.python.org/3/library/unittest.html#distinguishing-test-iterations-using-subtests). This should help to have this test run with multiple parameters.

jdemel (commented):
Mostly looks good. Just one suggestion.

gr-fec/python/fec/qa_fecapi_cc.py
daniestevez: I'm using subtest 3 lines below this. Is there a better way to use it? I've never known if there is a way to use subtest to generate different test *methods*. The usage I'm doing here (which is like the example in the docs) bugs me a little because it requires several `for`'s and levels of indentation to iterate over several parameters (I guess I could condense the nested `for`'s in a single `for` with `zip`).

jdemel: That seems to be the way to go. I didn't notice.

jdemel (approved):
LGTM

PR #6588, opened 2023-03-19
grc: Error Open the source of a hierachical block

dl1ksv:
## Description

1. Open a fg that contains a hierachical block
2. Select the hierachical block
3. Open the source of the selected hierachical block from the grc menu

gives the following error in the terminal

Loading: "/home/schroer/gnuradiocomponents/gr-noaa/examples/doppler_test.grc"
>>> Done

Loading: "/home/schroer/gnuradiocomponents/gr-funcube/examples/decodeapt.grc"
>>> Done
  File "/usr/local/gnuradio/lib64/python3.11/site-packages/gnuradio/grc/gui/Application.py", line 782, in _handle_action
    for b in flow_graph.selected_blocks():
  File "/usr/local/gnuradio/lib64/python3.11/site-packages/gnuradio/grc/gui/canvas/flowgraph.py", line 673, in <genexpr>
    return (e for e in self.selected_elements if e.is_block)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
RuntimeError: Set changed size during iteration

Despite the error messages the block is correctly displayed in grc.

I'm not sure why set is changed but according to the python recommendations changing to self.selected_elements.copy() fixes this issue.

## Testing Done

Tested with a complex flowgraph
## Checklist

### DIFF
--- a/grc/gui/canvas/flowgraph.py
+++ b/grc/gui/canvas/flowgraph.py
@@ -670,7 +670,7 @@ class FlowGraph(CoreFlowgraph, Drawable):
         Returns:
             sub set of blocks in this flow graph
         """
-        return (e for e in self.selected_elements if e.is_block)
+        return (e for e in self.selected_elements.copy() if e.is_block)
 
     @property
     def selected_block(self):


willcode:
It's a race condition. Do we care about fixing the race, wherever it is, or just having a self-consistent snapshot of the set?

dl1ksv:
I think we should fix the race

willcode:
I took a quick look but didn't find it.

dl1ksv:
The error does not depend on the complexity of the flowgraph.
Even with an simple fg this error can be reproduced. The error test is a simple hier2 block.

hier2

dl1ksv:
You can add a second hier2 block and select both hier2 blocks simultaneaosly. Without this patch tring to open the source only leads to one sourc opened. With this patch both sources will be opened and displayed in grc.

dl1ksv:
How did you open the source of the hier block ? The hier2 block must not already be opened in grc. You must select the hier2 block in the fg and open the source of  the selected block by clicking the right ouside symbol. ( the spotted arrow down ) in the screenshot above.

willcode:
https://github.com/gnuradio/gnuradio/blob/c36e3b59ea0002f1cb9d178b411c9b1d3d8a5ade/grc/gui/Application.py#L782
When `new_page()` is called the selected blocks would be cleared. How `selected_elements` should be protected isn't really clear. I was sort of thinking that it should be up to the caller (above) to deal with it, but let's go with your fix.

Issue #6559, opened 2023-02-28
ZMQ Enhancements


jwmelto:
### Feature Description

This describes a set of changes to the way ZMQ is implemented in GNU Radio that I think would be beneficial. My experience is primarily with the PUB/SUB stream sinks/sources, so it's possible that my understanding has holes in it.

## Stable Protocol
The over-the-wire format sent between a PUB sink and a SUB source varies, depending upon how the sender is configured. This creates a brittle coupling where the receiver must match the sender's configuration. This is not necessary, if we change how messages are serialized.

I think there may have been a misconception about how ZMQ multi-part messages work that led to the idea that they are expensive and should be avoided if possible. In reality, multi-part messages are managed by the ZMQ layer and are delivered _atomically_ to the application. That is, once the first part of a multi-part message is delivered, **all** parts are available. Furthermore, ZMQ manages the network traffic to maximize efficiency, so we shouldn't be as concerned about it.

Therefore, I propose that **ALL** ZMQ messages follow this multi-part protocol:
1. filter part - I don't know if people have done topic-based publishing in GNU Radio; the requirement of the sink to `bind` to the socket (see #4782) might inhibit multiple publishers doing different filterable things. However, if we are going to support it at all (this could be removed entirely), we should do it consistently. The current processing does put the filter string (if defined!) in a separate message, and simply discards it on read, since ZMQ implements the filtering.
2. tags - today, tags are a configurable feature. I would retain this capability, but dedicate a message part to them. If the block chooses not to send tags, then this part would be empty. The negligible overhead of an empty message part is more than worth the de-coupling of requiring the sender and receiver to agree on tag passing. Today, if tags are enabled in the sink, a special block is pre-pended to the payload. That block would be in this part, separating metadata from data
3. payload - the items, as defined by the stream, as implemented today.

The primary benefit of this stable protocol is that misconfiguration between sinks and sources would be eliminated. A well-defined wire protocol could more easily be used by non-GNU Radio blocks.

I prototyped this approach simply using Python blocks. As expected, the performance was somewhat worse than the current approach. Understanding the performance of a production implementation compared to the existing approach would be a prerequisite to accepting this solution.

## Tags
This deserves a sub-section, because there is some inconsistency in the handling of tags in ZMQ blocks.

[...]

PR #6596, opened 2023-03-22
zmq: user-selectable connect/bind for all zmq blocks

### LINKED ISSUE #4782: gr-zmq blocks are not compatible with zmq devices
## Overview

The [base_sink_impl](https://github.com/gnuradio/gnuradio/blob/master/gr-zeromq/lib/base_impl.cc#L83) enforces that the sink blocks bind their ports. ZMQ does *not* enforce any conventions that certain zmq socket types bind the port.

Similarly, [base_source_impl](https://github.com/gnuradio/gnuradio/blob/master/gr-zeromq/lib/base_impl.cc#L157) enforces that the source blocks connect their ports. ZMQ does *not* require that certain zmq socket types connect to the port. 

## Issues

[ZMQ devices](https://learning-0mq-with-pyzmq.readthedocs.io/en/latest/pyzmq/devices/devices.html) are a useful construct for managing ZMQ networks; the current assumptions on sink/source bind/connect are not compatible with ZMQ devices. 

As an example, consider [ZMQ Forwarders](https://learning-0mq-with-pyzmq.readthedocs.io/en/latest/pyzmq/devices/forwarder.html). These devices act like a pub-sub proxy server that allows multiple publishers to route to multiple subscribers through a stable hub. 

image

In this model, typically the forwarder_device **binds** both the front and backend ports (since it is acting as a server), see code snippet below from [zmq documentation](https://learning-0mq-with-pyzmq.readthedocs.io/en/latest/pyzmq/devices/forwarder.html)

image

When trying to integrate a GnuRadio flowgraph with a larger framework, our ZMQ PUB ports are not compatible with this standard ZMQ device. 

There are analogous devices  for [XREP/XREQ](https://learning-0mq-with-pyzmq.readthedocs.io/en/latest/pyzmq/devices/queue.html) and  [PUSH/PULL](https://learning-0mq-with-pyzmq.readthedocs.io/en/latest/pyzmq/devices/streamer.html) models that are also not compatible. 

## Proposed Solution

Update the base_source_impl and base_sink_impl to **default** to bind/connect (respectively), but allow for configuration that supports either binding or connecting the socket.

willcode:
## Description
Message-oriented ZMQ blocks already have selectable connect/bind. Here, the stream-oriented blocks are given the same option. Some refactoring of the base classes was necessary. The API and default behavior is unchanged.

## Related Issue
Various ZMQ improvements were dicussed in #6559. 
Specifically fixes #4782.

ZMQ stream-oriented blocks.

## Testing Done
PUB (connect) to SUB (bind) tested in a flowgraph.

QA tests and documentation TBD.

## Checklist

### DIFF
--- a/gr-zeromq/grc/zeromq_sub_source.block.yml
+++ b/gr-zeromq/grc/zeromq_sub_source.block.yml
@@ -39,6 +39,13 @@ parameters:
     label: Filter Key
     dtype: string
     default: ''
+-   id: bind
+    label: Connection
+    category: Advanced
+    dtype: enum
+    default: 'False'
+    options: ['True', 'False']
+    option_labels: ['Bind', 'Connect']
 
 outputs:
 -   domain: stream
@@ -48,19 +55,20 @@ outputs:
 templates:
     imports: from gnuradio import zeromq
     make: zeromq.sub_source(${type.itemsize}, ${vlen}, ${address}, ${timeout}, ${pass_tags},
-        ${hwm}, ${key})
+        ${hwm}, ${key}, ${bind})
 
 cpp_templates:
     includes: [ '#include <gnuradio/zeromq/sub_source.h>' ]
     declarations: gr::zeromq::sub_source::sptr ${id};
-    make: this->${id} = gr::zeromq::sub_source::make(${type.itemsize}, 
-        ${vlen}, 
+    make: this->${id} = gr::zeromq::sub_source::make(${type.itemsize},
+        ${vlen},
         const_cast<char *>(${address}${'.c_str())' if str(address)[0] not in '"\'' else ')'},
-        ${timeout}, 
-        ${pass_tags}, 
+        ${timeout},
+        ${pass_tags},
         ${hwm},
-        ${key});
-    link: ['gnuradio::gnuradio-zeromq']      
+        ${key},
+        ${bind});
+    link: ['gnuradio::gnuradio-zeromq']
     translations:
       'True': 'true'
       'False': 'false'
--- a/gr-zeromq/include/gnuradio/zeromq/pub_sink.h
+++ b/gr-zeromq/include/gnuradio/zeromq/pub_sink.h
@@ -50,6 +50,8 @@ public:
      * \param hwm High Watermark to configure the socket to (-1 => zmq's default)
      * \param key Prepend a key/topic to the start of each message (default is none)
      * \param drop_on_hwm Optionally drop samples when high watermark is reached.
+     * \param bind If true this block will bind to the address, otherwise it will
+     * connect; the default is to bind
      */
     static sptr make(size_t itemsize,
                      size_t vlen,
@@ -58,7 +60,8 @@ public:
                      bool pass_tags = false,
                      int hwm = -1,
                      const std::string& key = "",
-                     bool drop_on_hwm = true);
+                     bool drop_on_hwm = true,
+                     bool bind = true);
 
     /*!
      * \brief Return a std::string of ZMQ_LAST_ENDPOINT from the underlying ZMQ socket.
--- a/gr-zeromq/lib/base_impl.cc
+++ b/gr-zeromq/lib/base_impl.cc
@@ -26,8 +26,12 @@ namespace zeromq {
 base_impl::base_impl(int type,
                      size_t itemsize,
                      size_t vlen,
+                     char* address,
                      int timeout,
                      bool pass_tags,
+                     int hwm,
+                     bool sink,
+                     bool bind,
                      const std::string& key)
     : d_context(1),
       d_socket(d_context, type),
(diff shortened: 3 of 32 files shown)


gr-zeromq/grc/zeromq_sub_source.block.yml
```diff
@@ -39,6 +39,13 @@ parameters:
     label: Filter Key
     dtype: string
     default: ''
+-   id: do_bind
```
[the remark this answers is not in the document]
willcode: That was left over from a debugging session I'm embarassed to describe - thanks - changed.

gr-zeromq/lib/base_impl.cc
```diff
@@ -43,6 +47,42 @@ base_impl::base_impl(int type,
...
+            d_socket.set(zmq::sockopt::sndhwm, hwm);
+        } else {
+            d_socket.set(zmq::sockopt::rcvhwm, hwm);
+        }
+#else
+#ifdef ZMQ_SNDHWM
+        if (sink) {
+            d_socket.setsockopt(ZMQ_SNDHWM, &hwm, sizeof(hwm));
+        } else {
+            d_socket.setsockopt(ZMQ_RCVHWM, &hwm, sizeof(hwm));
```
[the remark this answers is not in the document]
willcode: ZMQ_SNDHWM and ZMQ_RCVHWM were introduced at the same time, replacing ZMQ_HWM. Maybe checking version numbers would work, too. I think this is good enough for now since even Ubuntu 18.04 has a newer version.

marcusmueller (approved):

gr-zeromq/include/gnuradio/zeromq/pub_sink.h
```diff
@@ -58,7 +58,8 @@ class ZEROMQ_API pub_sink : virtual public gr::sync_block
                      bool pass_tags = false,
                      int hwm = -1,
                      const std::string& key = "",
-                     bool drop_on_hwm = true);
+                     bool drop_on_hwm = true,
+                     bool bind = true);
```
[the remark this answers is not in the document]
willcode: Done.

gr-zeromq/lib/base_impl.cc
jwmelto, line 71: This section of code is the problematic behavior pulled into the base class. Previously, it was correctly in the source/sink sub-classes, where each class did the appropriate thing. However, for the sake of a single place where you put the bind/connect logic, the base class interface and implementation has become polluted with what is more properly sub-class abstraction.

The challenge is in having polymorphic behavior in a constructor. I think the most straightforward solution would be to move the socket setup out of the ctor and into the `start()` method. Then, this base class could declare the following methods:

```c++
public:
    bool start() override;
protected:
    virtual void set_hwm(...) = 0;
    void link_socket();
```

and implement:
```c++
bool base_impl::start()
{
    set_hwm();
    link_to_socket();

    return gr::sync_block::start();
}

void link_socket()
{
    if (bind) {
         d_socket.bind(address);
     } else {
         d_socket.connect(address);
     }
}

```
This of course, comes at a cost of holding the HWM and the address as member state of the base class.

gr-zeromq/lib/base_impl.cc
```diff
@@ -26,8 +26,12 @@ namespace zeromq {
 base_impl::base_impl(int type,
                      size_t itemsize,
                      size_t vlen,
+                     char* address,
                      int timeout,
                      bool pass_tags,
+                     int hwm,
+                     bool sink,
```
jwmelto, line 33: As an implementation decision, this is one way to do it, but I generally view switching behavior based on a state value as a failure to abstract. Given that you are trying to support multiple ZMQ API versions (with incompatible argument types), this is possibly the lesser of two evils, but it is still evil.

gr-zeromq/lib/base_impl.cc
```diff
@@ -67,32 +107,10 @@ base_sink_impl::base_sink_impl(int type,
...
-
-    /* Set ZMQ_LINGER so socket won't infinitely block during teardown */
-#if USE_NEW_CPPZMQ_SET_GET
-    d_socket.set(zmq::sockopt::linger, LINGER_DEFAULT);
-#else
-    d_socket.setsockopt(ZMQ_LINGER, &LINGER_DEFAULT, sizeof(LINGER_DEFAULT));
-#endif
-
-    /* Bind */
-    d_socket.bind(address);
```
jwmelto, line 95: Alternatively, you could simply leave this code in the sub-classes and replace the explicit "bind" call with a method on the base class, something like:
```c++
void base_impl::link_to_address(std::string const& address, bool bind)
{
    if (bind) d_socket.bind(address);
    else      d_socket.connect(address);
}
```
This is the most minimal change, preserving the integrity of the base class.

gr-zeromq/lib/base_impl.h
```diff
@@ -23,8 +23,12 @@ class base_impl : public virtual gr::sync_block
     base_impl(int type,
               size_t itemsize,
               size_t vlen,
+              char* address,
               int timeout,
               bool pass_tags,
+              int hwm,   // high watermark
+              bool sink, // is sink, else source
```
jwmelto, line 30: The interface problem here is that there is insufficient semantic meaning for this argument. The bind/connect flag is better, because ZMQ sockets are agnostic to bind vs connect (with the exception that binding more than once is an error, but that's enforced by the OS not the library).

What does sink vs not sink mean? Nothing to users of the class; it's a lower-level implementation detail of which socket option to set. These kinds of details should not be promoted to interfaces.

jwmelto (commented):
This looks good enough to me. I'm sure it works, which is the main test of a change. As noted, I think you pulled too much behavior into the base class to the detriment of the interface, but it's a subtle complaint that may fall into the realm of preference.

jwmelto:
Guess I'm too slow 😃

willcode:
@jwmelto Thanks for looking at this. Since it's all private impl, with little risk of anyone writing more subclasses, I may have taken some liberties to get the functionality in. But that also means we can clean up these classes if it's important without affecting anything. The main concern was the public API which I think is correct. I'll spend some time thinking about how to implement your ideas.

PR #6597, opened 2023-03-22
uhd: rfnoc: allow multichannel radio+ddc/duc

gmeyerho:
The TX/RX radio and DDC/DUC blocks had a num_channels parameter however the function calls to set freq, rate, etc. were hardcoded to channel 0, this allows for multichannel functionality

## Description
The current implementation really only allowed for using channel 0 even though it had a num_channels parameter. This moves the .yml file to the python generation that some other blocks used for multiple channels.

There are also several other cosmetic changes intended to mirror some of the newer blocks including hiding channel count from displaying on the block (the ports off the block already show this), and not showing block args, etc. unless the user has set something there. Also allows user to give different settings per channel, another option would be to keep the old look and have changing Center Frequency change it to the same value for every channel.

Old look vs new look:
image

## Related Issue
I realized while adding other blocks that the radio block didn't work correctly with multiple channels. This change also reorders setting frequency during make to be after setting bandwidth, which I had noticed caused the expected signal to not appear until manually setting the frequency after the application started.

gr-uhd radio and conveter blocks

## Testing Done
Ran included examples added for showing multichannel functionality. I also tested that the existing single channel examples still work and this change didn't inadvertently break them.

## Checklist


willcode (approved):

mbr0wn:
> BTW, `GEN_BLOCK_YML()` is defined in both `grc/blocks/CMakeLists.txt` and `gr-uhd/grc/CMakeLists.txt`. Is it needed in `gr-uhd`?

As it stands, the CMake in gr-uhd doesn't have access to that macro. We could move it to `cmake/Modules/Gr$SOMETHING.cmake`, but at this point that seems a small gain.

mbr0wn (approved):
This also seems to work fine picking up .grc files generated before this change.

mbr0wn:
> This change also reorders setting frequency during make to be after setting bandwidth, which I had noticed caused the expected signal to not appear until manually setting the frequency after the application started.

Which USRP did you see this behaviour on?

gmeyerho:
> > This change also reorders setting frequency during make to be after setting bandwidth, which I had noticed caused the expected signal to not appear until manually setting the frequency after the application started.
> 
> Which USRP did you see this behaviour on?

I saw this using the n300, when executing a flow graph with a rx radio block the QT GUI frequency sink would initially show up as there was no signal present, even though I had a signal input at the expected frequency. After setting the frequency to the same frequency in a QT GUI Entry the signal would then show up as expected.

PR #6598, opened 2023-03-22
uhd: rfnoc: fix literal_eval error

gmeyerho:
literal_eval() was throwing an exception on drag and drop because it requires a string containing a literal (e.g. string) rather than just a string itself, it then returns the contained string.

This was only an issue on drag/drop as the block code later does this transformation itself, but not before the first var_make()

## Description
Dragging and dropping a graph block from the right panel caused the block to not show up, and looking at the terminal that launched the gnuradio-companion it threw a Syntax EOF exception. This was because during the initial var_make from drag and drop, literal_eval was being passed something like 'type=n3xx', whereas later var_make passed it "'type=n3xx'". I traced this enclosing to the param rewrite in grc. There is likely also the option of trying to reorder operations in grc to ensure this encapsulating before calling var_make, but it seemed the best option to only be touching one block rather than inadvertently breaking another block.

Only drag/drop was previously broken, and you could workaround by drag/drop then save .grc file, close it then open it up and the block would be there, so starting from an example and reworking it also worked fine.

## Related Issue
Drag/drop of rfnoc graph threw exception

gr-uhd graph block

## Testing Done
Tested dragging and dropping a graph block, ensured it was placed as expected. Also tested running applications and opening applications that already had a graph block.

## Checklist

### DIFF
--- a/gr-uhd/grc/uhd_rfnoc_graph.block.yml
+++ b/gr-uhd/grc/uhd_rfnoc_graph.block.yml
@@ -9,10 +9,25 @@ templates:
       <%
           import ast
           # Sanitize
-          graph_args = ast.literal_eval(dev_addr.strip())
-          dev_args_s = ast.literal_eval(dev_args.strip())
-          clock_source_s = ast.literal_eval(clock_source.strip())
-          time_source_s = ast.literal_eval(time_source.strip())
+          # literal_eval requires a string containing a literal (e.g. string), on drag/drop
+          # the params are just strings, so they cause an exception, call literal_eval again
+          # with params enclosed in a string
+          try:
+              graph_args = ast.literal_eval(dev_addr.strip())
+          except SyntaxError:
+              graph_args = ast.literal_eval("'{}'".format(dev_addr.strip()))
+          try:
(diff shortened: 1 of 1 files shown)


gmeyerho:
> What happens if you drop `literal_eval()` altogether? Or use type `raw`. Not sure whether either of those will work. Since this is the only block doing `literal_eval()`, let's keep changes, as you say, to this block. Is there something special that these blocks need to parse that gr-osmosdr, for instance, doesn't? They use string without eval for device args. Now that I look at their yml ... they use `eval()` for the antenna and bandwidth fields. Eh? Reminder, never run someone else's GRC without scanning it first!

Previously, the block didn't have literal_eval() and simply used join() for var_make(). @mbr0wn's change (#6364)  to account for clock/time source parameters added it, so I felt it was best to leave those in and work around the exception. But I don't know of any reason in particular these parameters need to be checked more than those in other blocks.

mbr0wn:
Superseded by #6602.

@willcode you were right to question the usage of `literal_eval()`. I think Grant was doing the right thing fixing the usage, but with deeper digging it turns out we didn't actually need it.

PR #6602, opened 2023-03-24
uhd: rfnoc: fix literal_eval error

mbr0wn:
literal_eval() was used in the RFNoC graph GRC binding to sanitize
inputs. However, it was throwing an exception on drag and drop because
it requires a string containing a literal (e.g. string) rather than
just a string itself,

This was only an issue on drag/drop as the block code later does this transformation itself, but not before the first var_make()

However, the literal_eval() is not actually necessary, since the objects
passed in are of type TemplateArg, not just
string-or-something-that-evaluates-to-string, and thus we can use
facilities from TemplateArg to properly escape the values before usage.

## Related Issue

Supersedes https://github.com/gnuradio/gnuradio/pull/6598

RFNoC graphs in GRC

## Testing Done

Did some RFNoC graphs in GRC

## Checklist

### DIFF
--- a/gr-uhd/grc/uhd_rfnoc_graph.block.yml
+++ b/gr-uhd/grc/uhd_rfnoc_graph.block.yml
@@ -7,15 +7,14 @@ templates:
       from gnuradio import uhd
   var_make: |
       <%
-          import ast
           # Sanitize
-          graph_args = ast.literal_eval(dev_addr.strip())
-          dev_args_s = ast.literal_eval(dev_args.strip())
-          clock_source_s = ast.literal_eval(clock_source.strip())
-          time_source_s = ast.literal_eval(time_source.strip())
+          graph_args = str(dev_addr()).strip()
+          dev_args_s = str(dev_args()).strip()
+          clock_source_s = str(clock_source()).strip()
+          time_source_s = str(time_source()).strip()
           # Build full dev args
           if dev_args_s:
-              graph_args += f",{clock_source_s}"
+              graph_args += f",{dev_args_s}"
           if clock_source_s:
               graph_args += f",clock_source={clock_source_s}"
           if time_source_s:


gmeyerho (approved):

mbr0wn:
@willcode @ryanvolz Do you recognize this error:
```
In file included from $SRC_DIR/build/gnuradio-runtime/CMakeFiles/common-precompiled-headers.dir/cmake_pch.hxx:5:
In file included from $SRC_DIR/gnuradio-runtime/include/gnuradio/top_block.h:15:
In file included from $SRC_DIR/gnuradio-runtime/include/gnuradio/hier_block2.h:15:
In file included from $SRC_DIR/gnuradio-runtime/include/gnuradio/basic_block.h:15:
In file included from $SRC_DIR/gnuradio-runtime/include/gnuradio/io_signature.h:23:
In file included from $SRC_DIR/gnuradio-runtime/include/gnuradio/buffer_double_mapped.h:15:
In file included from $SRC_DIR/gnuradio-runtime/include/gnuradio/buffer.h:15:
In file included from $SRC_DIR/gnuradio-runtime/include/gnuradio/custom_lock.h:16:
In file included from $SRC_DIR/gnuradio-runtime/include/gnuradio/thread/thread.h:19:
In file included from $PREFIX/include/boost/thread/thread.hpp:12:
In file included from $PREFIX/include/boost/thread/thread_only.hpp:22:
In file included from $PREFIX/include/boost/thread/detail/thread.hpp:41:
In file included from $PREFIX/include/boost/functional/hash.hpp:6:
$PREFIX/include/boost/container_hash/hash.hpp:132:33: error: no template named 'unary_function' in namespace 'std'; did you mean '__unary_function'?
        struct hash_base : std::unary_function<T, std::size_t> {};
                           ~~~~~^
$BUILD_PREFIX/bin/../include/c++/v1/__functional/unary_function.h:46:1: note: '__unary_function' declared here
using __unary_function = __unary_function_keep_layout_base<_Arg, _Result>;
^
1 error generated.
```

? 

It popped up during the OSX build; it is unrelated to the changeset here.

willcode:
Looks like boost and the build process disagree on C++ standard (that function was removed in C++17). Boost on my machine has replacement code, so the boost on the OSX build must be old?

ryanvolz:
This just started cropping up because conda-forge packaged clang 16 which drops that function according to the standard as @willcode mentioned. Boost still relied on it though, so there is a patch in the works to fix what is now a broken Boost package: https://github.com/conda-forge/boost-cpp-feedstock/pull/135. There's a workaround compile definition that could be applied, but it's probably easier to wait for the package to be fixed. So I'd say it's fine to ignore for now, and hopefully the fixed package becomes available soon.

Issue #5650, opened 2022-03-19
Selector block does not consume inputs sufficiently


matt-g-87:
### What happened?

Hello all,
I have been using the (new) Selector block from gr-blocks for a while. In my case I have multiple inputs with a single output. _My inputs do not necessarily produce items at the same rate._ Sometimes I get hangs and the selector block stops performing the general_work function. After adding some debug logs, I have determined that this is caused by the general_work function not consuming all inputs sufficiently. What I mean by this, is that the non-selected inputs can build up large numbers of unconsumed items, which eventually leads to a hang. 

This issue appears to be caused by the use of `consume_each(noutput_items)` at the end of the general_work function in the selector block. This consume_each call, of course, tries to consume an equal number of samples every time from all inputs. This is based on the presumption that all of the inputs to the selector block come in at the same rate. However, since the changes made by the commit for issue #3762 and the solution #3763, the selector should no longer assume that inputs have the same rate of items being supplied to it (if any at all). The primary change of #3763 was that forecasting is only done for the selected input (not all inputs, as before that change).

Likewise, instead of calling consume_each(noutput_items), we should consider:
* Consuming noutput_items ONLY from the selected input
* And then consuming ALL ninput_items from each of the non-selected inputs. This clears out all the items in for the other, non-selected inputs. This prevents them from "building up" if they are provided to the selector faster than the selected input.

Since the intent of the Selector block is to pass along the selected input, and to drop all other inputs... then lets drop the other inputs as fast as possible.

I have tested replacing the `consume_each(noutput_items` call with something like this with good success:
```
for (unsigned i = 0; i < d_num_inputs; i++) {
    if (i == d_input_index)
        consume(i,noutput_items);
    else
        consume(i,ninput_items[i]);
}
```

This will consume all of the irrelevant/non-selected inputs as fast as possible.

### System Information

OS: Ubuntu 20.04 LTS
GR Installation Method: Source

### GNU Radio Version

3.9  (maint-3.9)

### Specific Version

3.9.5.0

### Steps to Reproduce the Problem

* Create a selector block with multiple inputs and one output
* On the "selected" input, provide items from a block at some rate _R_
* On the "non-selected" inputs, provide items from a different block at some rate much faster than _R_
* Observe an eventual hang, since the non-selected inputs aren't consumed fast enough

willcode:
[...]

PR #6616, opened 2023-04-04
blocks: selector: don't blindly consume same amount from all inputs

marcusmueller:
- introduce new sync consumption mode that consumes the same from all
  inputs
- default to mode where as much as possible is consumed from active, and
  at most that from inactive inputs

Since the old mode of operation was buggy, it can't be available. Also,
a dump-as-quickly-as-possible mode will be very surprising.

## Description

@Notou noticed that selector consumes `noutput_items` from all inputs – regardless of item availability.

`consume_each` does no sanity checks. Hilarity ensues.

Selector.

This might unbreak other applications that inexplicably only worked with throttles or similar crutches.

## Testing Done

I haven't been able to come up with good tests.

## Checklist

### DIFF
--- a/gr-blocks/include/gnuradio/blocks/selector.h
+++ b/gr-blocks/include/gnuradio/blocks/selector.h
@@ -22,9 +22,13 @@ namespace blocks {
  * \ingroup misc_blk
  *
  * \details
- * Connect the sink at input index to the source at output index
- * Samples from other input ports are consumed and dumped
- * Other output ports produce no samples
+ * Connect the sink at input index to the source at output index.
+ *
+ * Samples from other input ports are consumed and dumped at a rate
+ * which is limited by (but not necessarily the same as) the rate of
+ * samples copied from the active input to the active output.
+ *
+ * Other output ports produce no samples.
  *
  */
 class BLOCKS_API selector : virtual public block
@@ -32,11 +36,20 @@ class BLOCKS_API selector : virtual public block
 public:
     typedef std::shared_ptr<selector> sptr;
 
+    /*!
+     * Create new selector block and return a shared pointer to it
+     *
+     * \param itemsize size of the input and output items
+     * \param input_index the initially active input index
+     * \param output_index the initally active output index
+     */
     static sptr
     make(size_t itemsize, unsigned int input_index, unsigned int output_index);
 
-    // When enabled is set to false, no output samples are produced
-    // Otherwise samples are copied to the selected output port
+    /*!
+     * When enabled is set to false, no output samples are produced.
+     * Otherwise samples are copied to the selected output port
+     */
     virtual void set_enabled(bool enable) = 0;
     virtual bool enabled() const = 0;
 
--- a/gr-blocks/lib/selector_impl.cc
+++ b/gr-blocks/lib/selector_impl.cc
@@ -33,21 +33,24 @@ selector_impl::selector_impl(size_t itemsize,
             io_signature::make(1, -1, itemsize),
             io_signature::make(1, -1, itemsize)),
       d_itemsize(itemsize),
-      d_enabled(true),
       d_input_index(input_index),
       d_output_index(output_index),
       d_num_inputs(0),
-      d_num_outputs(0)
+      d_num_outputs(0),
+      d_enabled(true)
 {
     message_port_register_in(pmt::mp("en"));
-    set_msg_handler(pmt::mp("en"), [this](pmt::pmt_t msg) { this->handle_enable(msg); });
+    set_msg_handler(pmt::mp("en"),
+                    [this](const pmt::pmt_t& msg) { this->handle_enable(msg); });
 
     message_port_register_in(pmt::mp("iindex"));
     set_msg_handler(pmt::mp("iindex"),
-                    [this](pmt::pmt_t msg) { this->handle_msg_input_index(msg); });
+                    [this](const pmt::pmt_t& msg) { this->handle_msg_input_index(msg); });
+
     message_port_register_in(pmt::mp("oindex"));
-    set_msg_handler(pmt::mp("oindex"),
-                    [this](pmt::pmt_t msg) { this->handle_msg_output_index(msg); });
+    set_msg_handler(pmt::mp("oindex"), [this](const pmt::pmt_t& msg) {
+        this->handle_msg_output_index(msg);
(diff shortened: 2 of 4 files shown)


marcusmueller:
hm, my clang-format insists the current format is right. hmmm...

marcusmueller:
the additional defaultable argument shouldn't break API and the code makes no use of overly modern C++ feautures for better backportability.

willcode:
@marcusmueller Do you think the existing behavior is useful enough to keep as a default, or should we just change it?

marcusmueller:
@willcode since the current behaviour advances read pointers beyond what was written, keeping current behaviour is impossible

willcode:
Sync mode seems like the only predictable mode. Per-port relative rates would be interesting, but a little complicated for the user.

marcusmueller:
I agree that sync mode is the thing that seems somewhat sane. But here's the problem: I presume `selector` is being used to select e.g. the one USRP that is currently streaming from a set of sources. If we enforced syncness, then all selector would freeze unless all inputs are "active".

I didn't plan to implement the `bool sync_consume = false` feature at all at first, but then realized that this change would quite probably break many things out there, so I did and made it the default :( :(

gr-blocks/include/gnuradio/blocks/selector.h
```diff
@@ -32,11 +32,25 @@ class BLOCKS_API selector : virtual public block
...
 
-    static sptr
-    make(size_t itemsize, unsigned int input_index, unsigned int output_index);
+    /*!
+     * Create new selector block and return a shared pointer to it
+     *
+     * \param itemsize size of the input and output items
+     * \param input_index the initially active input index
+     * \param output_index the initally active output index
+     * \param sync_consume consume the exact same amount of items from all inputs (true)
```
mbr0wn, line 41: Does anyone actually need this new mode? Seems a bit unnecessary to me.

The old mode (I think) was trying to do what the new version, with `sync_mode == false` is doing, except it didn't account for the fact that the input/output items may be different in size.

And regardless of this change, you can't switch selection on a particular item index, either.

mbr0wn (commented):
I think the changes in `general_work()` are important fixes, but the addition of `sync_mode` seems an unnecessary change, which makes the API more complicate for no good benefit that I can see.

Notou:
There was actually a discussion about that [a year ago](https://github.com/gnuradio/gnuradio/issues/5650).

And I would tend to agree with its conclusion.
It seems to me that the current buggy mode wanted to consume samples in sync, transferring the backpressure to upstream blocks, without considering inputs with less samples. (And that it ended up creating something closer to the `sync_consume = false` mode)

I believe a mode that doesn't consume anything from non input ports would be quite useful, at least to prevent whole graph section from running and consuming resources if not needed

And the `sync_consume = false` mode could handle situations where an inactive input port starves the rest of the flowgraph.

For sure, this would require users to really think about what rate they want/require the blocks to works at to decide which mode they need to use. But I don't think it's a bad thing in itself (if properly documented).

gr-blocks/include/gnuradio/blocks/selector.h
marcusmueller: > Does anyone actually need this new mode? Seems a bit unnecessary to me.

Imagine this: 4 Wavfile sources with different file lengths, hence potentially different "workload chunks", connected to one audio sink through `selector`. 

Do you want them to stay in sync? in that case, `sync_consume`. Do you just want to consume items as they are available from the inactive inputs, without rate-unlimiting them? `~sync_consume`

I do feel like `sync_consume` is the more "understandable" and useful mode, but it seems there's different perspectives on that!

(the oldold blks2 behaviour would be to connect all inactive inputs to null sinks, which is kind of the opposite of `~sync_consume` mode, but honestly, I'm having a bit of a harder time finding a good use case for *that*. Suddenly, throttles in flowgraphs that should have no throttles become a necessity.)

Notou: I can actually see two different types of non `sync_consume` mode:
- Absolutely no rate limit on inactive inputs. So consume the entirety of the input buffer, as fast as it fills up. -> needs upstream throttling
- Inactive inputs samples are consumed _up to_ the output rate. If more comes up, buffer will fill up. If there is less, samples are consumed as they are available. -> downstream backpressure handles throttling

marcusmueller:
Ha! I didn't know about that issue!
@Notou  so that issue concludes with three usage scenarios:

> 1. Consume Inputs Equally
> 2. Consume Unused Inputs Maximally
> 3. Don't consume unused inputs

So, 1. is the `sync_consume == true` case.  However, `sync_consume == false` of my implementation has no correspondence to these other two cases, and I'd concede, yeah, these seem more useful.

So, we go for an `enum`: `CONSUME_SYNC`, `CONSUME_MAX`, or `CONSUME_NONE` as argument?

Or do we keep selector a one-trick pony (whose trick is `CONSUME_SYNC`) without new arguments, and add a new more versatile block (which we could allow to have different policies for different inputs)?

Notou:
To me, your implementation of `sync_consume == false` corresponds to a good default for a `CONSUME_MAX` mode.
By consuming inputs maximally up to the active rate, thus limiting the need for extra throttling blocks upstream. Reducing potential confusion for unaware users

I would say that an actual `CONSUME_MAX` mode, where input buffers are consumed fully regardless of the state of the output buffer, should live in that more versatile (but harder to correctly param) block with potentially different policies for different inputs.

mbr0wn:
OK looks like folks have thought this through much more than I did in the 1 minute I spent reviewing this :smile:.

marcusmueller:
@Notou OK! So I think this is where I'll take this:

1. As the current state of the block is buggy, it's very hard to predict which kind of usage we repair or break by any pick of default behaviour
2. We pick what my proposed `sync_consume == false` does as what `selector` does, and don't introduce a new parameter does
3. The three other modes `CONSUME_SYNC`, `CONSUME_MAX`, `CONSUME_LIMITED_BY_ACTIVE` are subject for a separate block API (might very well be the same block, the constructor is private anyways) which supports per-input settings.

marcusmueller:
Changes to this PR made. I do feel like it would, however, be wise to deprecate the block as soon as we draft the more versatile one.

willcode (approved):

PR #6627, opened 2023-04-09
hier_block2: fix recursive setters

### LINKED ISSUE #6558: Setting processor affinity of hier_block2 with message ports segfaults
### What happened?

If we have a Python `gr.hier_block2` that uses message ports as outputs, when we try to set the processor affinity of the block by calling its `set_processor_affinity()` method, we get a segmentation fault.

This Python script replicates the problem:

https://gist.github.com/daniestevez/f472b2d9ab41c7561b8f17ee44a54989

A `hier_block2` that only contains a Message Strobe connected to its output is defined. The block is instantiated in a flowgraph together with a Message Debug that prints the messages. If we run the script without any options, it runs correctly. If we run the script using the `--set_affinity` command line option, when the flowgraph is constructed it will call the `set_processor_affinity()` method of the `hier_block2` instance. It segfaults at this point.

Adding some debugging prints to `hier_block2_detail::set_processor_affinity()` we see that the cause for the segfault is the following. The call to `d_fg->calc_used_blocks()` returns the hier block itself as part of the used blocks list. Since `set_processor_affinity()` calls `set_processor_affinity()` for each of the used blocks, we enter an infinite recursion that overflows the stack.

I haven't studied the `calc_used_blocks()` code, but I suspect that the problem is that the `self.msg_connect()` call in the `hier_block2` to connect to an output message port contains `self` as the endpoint of the connection.

### System Information

OS: Arch Linux
GR Installation Method: native distro package / built from source in debug mode and with some extra prints

### GNU Radio Version

3.10 (maint-3.10)

### Specific Version

3.10.5.0/3.10.5.1

See the Python script linked above.

### Relevant log output

_No response_

marcusmueller:
On the way, fix the fact that setting the logger level on a hier block never set the logger level of the block itself, only its contained blocks.

## Description

Setting things on hier blocks is hard, because the setting needs to be done recursively, and the cycle-allowing message port hierarchy offers a wide variety of pitfalls when it comes to enumerating blocks.

To solve that, actually *don't* do the first-semester thing of setting things
recursively; have a queue of blocks containing further blocks, and a set of
blocks you've already encountered. When there's no more blocks on queue, work
through that set.

## Related Issue

Fixes #6558

`hier_block2`. It should be noted that `flowgraph::calc_used_blocks` is questionable in efficiency, but that's an aside.

## Testing Done

Minimized and added Daniel's test from #6558, and verified it fails without, and passes with this change :confetti:.

## Checklist

### DIFF
--- a/gnuradio-runtime/lib/hier_block2_detail.cc
+++ b/gnuradio-runtime/lib/hier_block2_detail.cc
@@ -16,6 +16,8 @@
 #include <gnuradio/io_signature.h>
 #include <gnuradio/prefs.h>
 #include <gnuradio/top_block.h>
+#include <deque>
+#include <set>
 #include <sstream>
 #include <stdexcept>
 
@@ -889,17 +891,25 @@ void hier_block2_detail::unlock()
 
 void hier_block2_detail::set_processor_affinity(const std::vector<int>& mask)
 {
-    basic_block_vector_t tmp = d_fg->calc_used_blocks();
-    for (basic_block_viter_t p = tmp.begin(); p != tmp.end(); p++) {
-        (*p)->set_processor_affinity(mask);
+    // We formerly recursed here, can't do that: a message connection to the hier block
+    // itself leads to an infinite recursion.
+    for (auto basic_block_sptr : all_blocks()) {
+        auto block_sptr = std::dynamic_pointer_cast<block>(basic_block_sptr);
+        if (block_sptr) {
+            block_sptr->set_processor_affinity(mask);
+        }
     }
 }
 
 void hier_block2_detail::unset_processor_affinity()
 {
-    basic_block_vector_t tmp = d_fg->calc_used_blocks();
-    for (basic_block_viter_t p = tmp.begin(); p != tmp.end(); p++) {
-        (*p)->unset_processor_affinity();
+    // We formerly recursed here, can't do that: a message connection to the hier block
+    // itself leads to an infinite recursion.
+    for (auto basic_block_sptr : all_blocks()) {
+        auto block_sptr = std::dynamic_pointer_cast<block>(basic_block_sptr);
+        if (block_sptr) {
+            block_sptr->unset_processor_affinity();
+        }
     }
 }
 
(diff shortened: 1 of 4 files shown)


gnuradio-runtime/lib/hier_block2_detail.cc
```diff
@@ -996,5 +1017,70 @@ void hier_block2_detail::reset_hier_blocks_parent()
...
+        if (hier_block) {
+            d_logger->trace(
+                "all_blocks: Block {} from calculated_blocks is hier. Enqueue.",
+                block->alias());
+            queued_blocks.push_back(hier_block);
+        }
+    }
+
+    while (!queued_blocks.empty()) {
+        const auto dequeed_block = queued_blocks.front();
```
daniestevez, line 1054: Perhaps `dequeued_block` is better than `dequeed_block`?

gnuradio-runtime/lib/hier_block2_detail.cc
```diff
@@ -996,5 +1017,70 @@ void hier_block2_detail::reset_hier_blocks_parent()
     }
 }
 
+std::set<basic_block_sptr> hier_block2_detail::all_blocks()
+{
+    std::set<basic_block_sptr> found_blocks;
+    std::deque<hier_block2_sptr> queued_blocks;
+
+    // Get the blocks from our own flow graph. These can contain hier blocks
```
daniestevez, line 1025: Would it make sense to eliminate this part of the code by adding this block to `queued_blocks` at the beginning and then going directly to the `while (!queued_blocks.empty())` loop?

daniestevez (commented):
Thanks for fixing this! I've reviewed the changes, and although I'm not too familiar with this part of the runtime, the approach looks reasonable. I've put a couple of minor comments.

gnuradio-runtime/lib/hier_block2_detail.cc
marcusmueller: much better!

gnuradio-runtime/lib/hier_block2_detail.cc
marcusmueller: ah wait, no. The problem was that for the current `hier_block2_detail`, I don't have an sptr to the owning hier block, only a raw pointer.

I don't want to change `queued_blocks` to be `container<hier_block2*>`, because that would break the thread-safety of enumerating blocks (the result would still be nonsensical if someone was changing the internal block connections while I'm doing this, but at least it would still hold shared ownership of the contained blocks, which hence continue to live, and thus, this walk would not segfault).

The solution to that might be documenting you shouldn't change flowgraphs while you're setting affinity or log level, but I'd rather keep it safe and non-surprising; but as me forgetting this since Sunday highlights this should be commented on in the code, for it not to fall prey to future refactors.

daniestevez: That makes sense. I read the new comment and it seems good.

willcode:
@marcusmueller @dl1ksv would you deconflict this with #6641?

dl1ksv:
This pr and #6641 solve the problem of recursion  in different ways. I did some tests and found:
This pr fixes #6558 but still seqfaults for
processor_affinity() and log_level()
 while #6641 still segfaults for log_level() and set_log_level().

But both pr's fail to handle hier2 blocks correctly that contain  exactly one block.

marcusmueller:
> But both pr's fail to handle hier2 blocks correctly that contain exactly one block.

Oh! So, I seemingly didn't test that. Need to address that.

>  processor_affinity() and log_level()

oi! this means I missed something fundamental.

PR #6633, opened 2023-04-10
Enable GRC example tests

### LINKED ISSUE #1740: Adding automatic unit testing for .grc grcc compilation

marcusmueller:
## Description

In #4208, the necessary infrastructure was introduced, but promptly disabled – 100% test failure at the time being.

With #6632, the last failing GRC example flow graph (on my machine) gets fixed. Time to enable these

Downsides:

- 85 s additional test time on my machine

Upsides:

- Test coverage for all GRC examples
- Unearth GRC code generation regressions much earlier
- Potentially a free pony for everyone
- Integrate the Block YAMLs into testing by writing an example: good for our example stockpile, good for our block files

## Related Issue

Fixes #1740

- Testing: All of the modules
- GRC

## Testing Done

Is itself a mere enabling of unit tests

## Checklist

### DIFF
--- a/grc/gui_qt/components/window.py
+++ b/grc/gui_qt/components/window.py
@@ -138,13 +138,6 @@ class MainWindow(QtWidgets.QMainWindow, base.Component):
         # actions dynamically connected to class functions. Also, the log
         # functionality should be also allocated
         log.debug("__init__")
-        icon_theme = self.app.qsettings.value("appearance/default_qt_gui_theme", "Papirus-Dark")
-        available_themes = QtWidgets.QStyleFactory().keys()
-        if icon_theme in available_themes:
-            self.app.setStyle(QtWidgets.QStyleFactory().create(icon_theme))
-        else:
-            log.info(f"{icon_theme} not available. Falling back to default theme")
-            log.info(f"Available themes are {available_themes}")
 
         # Add the menus from the view
         menus = self.menus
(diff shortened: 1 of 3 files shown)


marcusmueller:
Interesting! I get neither the failures in fecapi nor in gr-uhd; now the question is, what's different on my machine vs the consistency with which the CI machines detect a problem with these flow graphs?

ryanvolz:
For the conda builds at least, grc_tests is timing out at 120 seconds because that's the timout specified in the build scripts when the tests are run! That will have to be increased and it's probably fine to go pretty high since I don't think hanging tests have ever really been a problem.

marcusmueller:
> I don't think hanging tests have ever really been a problem.

They've been before, especially on gr-digital tests that expect some messages to come in and that nondeterministically might not always happen. But I think argilo got these under control.

willcode:
@marcusmueller @ryanvolz Is this one ready for main, or is it going to make conda fail every time?

ryanvolz:
Raising/eliminating the timeout on the tests in the conda build scripts is still something that should happen before merge, but besides that I don't know about the pass/fail reliability in general.

PR #6641, opened 2023-04-13
hier_block2: handling processor affinity segfaults

### LINKED ISSUE #6558: Setting processor affinity of hier_block2 with message ports segfaults

dl1ksv:
## Description

In some cases set_processor_affinity / unset_processor_affinity / processor_affinity leads to a segfault. One example can be found in #6558.

The reason is an infinite loop in calling these functions.

This pr avoids this loop.
## Related Issue

Fixes #6558

hier_block2 and hier_block2_detail
## Testing Done

Running the example in #6558
## Checklist


willcode (approved):

willcode:
@dl1ksv @marcusmueller would you deconflict this with #6627?

marcusmueller:
That'll be hard! We're solving the same problem two different ways here!

gnuradio-runtime/lib/hier_block2_detail.cc
```diff
@@ -891,24 +891,24 @@ void hier_block2_detail::set_processor_affinity(const std::vector<int>& mask)
...
     }
 }
 
 void hier_block2_detail::unset_processor_affinity()
 {
     basic_block_vector_t tmp = d_fg->calc_used_blocks();
     for (basic_block_viter_t p = tmp.begin(); p != tmp.end(); p++) {
-        (*p)->unset_processor_affinity();
+        if (d_owner->unique_id() != (*p)->unique_id())
+            (*p)->unset_processor_affinity();
```
marcusmueller, line 908: Won't this fail if this hier block contains hier block, whose message output connects to an input of this hier block?

gnuradio-runtime/include/gnuradio/hier_block2.h
```diff
@@ -45,6 +45,7 @@ class GR_RUNTIME_API hier_block2 : public basic_block
      */
     std::unique_ptr<hier_block2_detail> d_detail;
 
+    std::vector<int> d_affinity; // thread affinity proc. mask
```
marcusmueller, line 48: why keep this in the interface instead of in the detail, if it's private anyways?

willcode: I was thinking the same thing ... but there are worse offenders in this file.

dl1ksv: It's the same behaviour as in block.k. See line 728 and 799

dl1ksv: And you need it to save and clear the mask if you call the affinity functions for the hier block.

gnuradio-runtime/lib/hier_block2_detail.cc
dl1ksv: I tested this with the example you added to qa_hier_block2.py and it does not seqfault.

willcode:
@dl1ksv @marcusmueller Discussions in both threads ended in a non-reassuring way :smile: . Any progress on either of these, or any guidance?

dl1ksv:
I now fixed the two segfaults in my proposed solution and added the test additional test from #6627 and completed it with calls for processor_affinity() and log_level().

Both calls fail at the moment #6627 but I think this could easily been fixed.
So I have no idea which solution should be prefered.

dl1ksv:
The qa_logger test seems to be affected by settings in ~/.gnuradio/config.conf. I have to figure that out.

dl1ksv:
The log_level handling of the top_block is buggy in main. Just building a top_block and printing or setting the log level results in seqfault

tb=top_block() 
print(tb.log_level())  # Seqfault !!!
tb.set_log_level('debug') # Seqfault !!!

I would expect at this point either the configuration setting of 
  $CMAKE_INSTALL_PREFIX/etc/gnuradio/conf.d/gnuradio-runtime.conf or
~/.gnuradio/config.conf if exists is used.

But you have to connect blocks to the top block, to make these calls work.

Why does only changing the log level of the first build block affects the log level of the top block and no others?
And it can quickly become confusing , who is the first.
Changing the log level of the top block by set_log_level affects all depend blocks, but not if the it's level is changed by setting the level of the first depending block.

I propose to restrict set_log_level to the block it's applied to and it's depend blocks. So I modified the qa_logger.py  to follow these thoughts.

mbr0wn:
@marcusmueller seems like this is 95% fine...? What's your verdict?

Issue #2666, opened 2019-07-24
Invent new `stateless` block type


marcusmueller:
After a discussion with @osh this sounds like a very good idea:

Blocks that don't actually need any block state shouldn't have a (`general_`)`work` method that's allowed to modify any local block state. Instead, a simple 

`gr::stateless_block::general_work(items_out, const items_in, tags_out) const`

would do.

michaelld:
Love the idea. I wonder if `*_work` is `const` whether buffers get updated correctly, whether the current R/W pointers or tags. Especially tags ... can I add a tag to a stream from a `const` method? Thinking not, but I haven't looked at code in a long time ... so maybe so? Interesting idea!

marcusmueller:
@michaelld a `const` member function can't modify any class instance data. Basically, the `this` pointer becomes `const` (and you'd have to imagine every single member access to be prefixed by `this->`).

So, what definitely works is instead of letting the function call `add_item_tag` is simply passing it an empty (or prefilled? Maybe that's more elegant when tags get propagated?) tag vector by reference. The function then just modifies that argument.

What doesn't work is using methods like `produce` or `set_relative_rate`, and that's the reason why I'd start doing this for the "simplest" type of blocks: sync blocks.

marcusmueller:
In fact, at this point, even things like `nitems_read` would be non-guaranteed to be callable because they're not `const`, but #2665 strives to change that

michaelld:
Maybe we need a `const` version of the most critical methods, which then internally de-`const`-ifies & allows calling the non-`const` method? Hacky, but it might work.

I think it's worth exploring moving various methods to `const` & seeing what compiles successfully and what breaks. We'd learn about GR internals, and maybe come up with better ways to do some required tasks (such as setting actual topology).

marcusmueller:
Well, the good thing is that most of these functions really don't change any state and could really well be made `const` without touching any code, as far as I can see.

michaelld:
> Well, the good thing is that most of these functions really don't change any state and could really well be made `const` without touching any code, as far as I can see.

So true! I'd stay if "we" are going to do this, start with the less-used methods & verify that GR builds correctly with each as `const`. Then iterate working toward more-used methods.

I think that we can divide the methods into 3 categories: (1) A reasonable number that can be made `const` without repercussions (at all); (2) some that can be with minimal repercussions; (3) some that will just critically break things either IT or OOT.

[...]

PR #6644, opened 2023-04-17
blocks: Add lambda_block

mbr0wn:
## Description

This is the C++ equivalent of the embedded Python block. It allows
passing in the code to general_work() as a parameter.

This block can be particularly useful when designing hierarchical blocks
which have a small custom kernel. Previously, this would require first
adding this custom kernel as a custom block, and then building the
hierarchical block. Now, custom work functions can be included in the
hierarchical block itself.

None, it's a new block. However, we use this in the upcoming C++ version of
the fosphor formatter block.

## Testing Done

Includes a unit test, also tested this as part of a hier block in the fosphor
formatter.

## Checklist

### DIFF
--- a/gr-blocks/include/gnuradio/blocks/lambda_block.h
+++ b/gr-blocks/include/gnuradio/blocks/lambda_block.h
@@ -1,0 +1,73 @@
+/* -*- c++ -*- */
+/*
+ * Copyright 2023 Ettus Research, A National Instruments Brand
+ *
+ * This file is part of GNU Radio
+ *
+ * SPDX-License-Identifier: GPL-3.0-or-later
+ */
+
+#ifndef INCLUDED_BLOCKS_LAMBDA_BLOCK_H
+#define INCLUDED_BLOCKS_LAMBDA_BLOCK_H
+
+#include <gnuradio/block.h>
+#include <gnuradio/blocks/api.h>
+#include <gnuradio/io_signature.h>
+#include <functional>
+
+namespace gr {
+namespace blocks {
+
+/*!
+ * \brief Map a generic function object or lambda into a GNU Radio block
+ * \ingroup misc_blk
+ *
+ * This block takes a function object as an argument, and uses that to execute
+ * the general_work() function.
+ *
+ * A neat use case for this block is for hierarchical blocks which require a
+ * kernel of custom processing; using the lambda block they can define their
+ * own partial work function.
+ */
+class BLOCKS_API lambda_block : virtual public block
+{
+public:
+    // gr::blocks::delay::sptr
+    typedef std::shared_ptr<lambda_block> sptr;
+
+    using general_work_func_t = std::function<int(lambda_block*,
+                                                  int,
+                                                  gr_vector_int&,
+                                                  gr_vector_const_void_star&,
+                                                  gr_vector_void_star&)>;
+
+    /*!
+     * \brief Make a delay block.
+     * \param itemsize size of the data items.
+     * \param delay number of samples to delay stream (>= 0).
+     */
+    static sptr make(general_work_func_t&& work_func,
+                     io_signature::sptr in,
+                     io_signature::sptr out,
+                     const std::string& name = "lambda_block");
+
+
+    inline void add_item_tag(unsigned int which_output,
+                             uint64_t abs_offset,
+                             const pmt::pmt_t& key,
+                             const pmt::pmt_t& value,
+                             const pmt::pmt_t& srcid = pmt::PMT_F)
+    {
+        gr::block::add_item_tag(which_output, abs_offset, key, value, srcid);
+    }
+
+    inline void add_item_tag(unsigned int which_output, const tag_t& tag)
+    {
+        gr::block::add_item_tag(which_output, tag);
+    }
+};
+
+} /* namespace blocks */
+} /* namespace gr */
+
+#endif /* INCLUDED_BLOCKS_LAMBDA_BLOCK_H */
--- a/gr-blocks/include/gnuradio/blocks/CMakeLists.txt
+++ b/gr-blocks/include/gnuradio/blocks/CMakeLists.txt
@@ -101,6 +101,7 @@ install(
           interleaved_char_to_complex.h
           keep_m_in_n.h
           keep_one_in_n.h
+          lambda_block.h
           lfsr_32k_source_s.h
           message_debug.h
           message_strobe.h
--- a/gr-blocks/lib/CMakeLists.txt
+++ b/gr-blocks/lib/CMakeLists.txt
@@ -111,6 +111,7 @@ set(BLOCKS_SOURCES
     interleaved_char_to_complex_impl.cc
     keep_m_in_n_impl.cc
     keep_one_in_n_impl.cc
+    lambda_block_impl.cc
     lfsr_32k_source_s_impl.cc
     message_debug_impl.cc
     message_strobe_impl.cc
@@ -227,6 +228,7 @@ if(ENABLE_TESTING)
         qa_gr_hier_block2.cc
         qa_gr_hier_block2_derived.cc
         qa_gr_top_block.cc
+        qa_lambda_block.cc
         qa_rotator.cc
         qa_set_msg_handler.cc)
     list(APPEND GR_TEST_TARGET_DEPS gnuradio-blocks)
--- a/gr-blocks/lib/lambda_block_impl.cc
+++ b/gr-blocks/lib/lambda_block_impl.cc
@@ -1,0 +1,49 @@
+/* -*- c++ -*- */
+/*
+ * Copyright 2023 Ettus Research, A National Instruments Brand
+ *
+ * This file is part of GNU Radio
+ *
+ * SPDX-License-Identifier: GPL-3.0-or-later
+ */
+
+#ifdef HAVE_CONFIG_H
+#include "config.h"
+#endif
+
+#include "lambda_block_impl.h"
+#include <gnuradio/io_signature.h>
+#include <utility>
+
+namespace gr {
+namespace blocks {
+
+lambda_block::sptr lambda_block::make(general_work_func_t&& work_func,
+                                      io_signature::sptr in,
+                                      io_signature::sptr out,
+                                      const std::string& name)
+{
+    return gnuradio::make_block_sptr<lambda_block_impl>(
+        std::move(work_func), in, out, name);
+}
+
+lambda_block_impl::lambda_block_impl(general_work_func_t&& work_func,
+                                     io_signature::sptr in,
+                                     io_signature::sptr out,
+                                     const std::string& name)
+    : block(name, in, out), _work_func(std::move(work_func))
+{
+}
+
+lambda_block_impl::~lambda_block_impl() {}
+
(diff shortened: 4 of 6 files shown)


gr-blocks/include/gnuradio/blocks/lambda_block.h
```diff
@@ -0,0 +1,73 @@
...
+     * \param itemsize size of the data items.
+     * \param delay number of samples to delay stream (>= 0).
+     */
+    static sptr make(general_work_func_t&& work_func,
+                     io_signature::sptr in,
+                     io_signature::sptr out,
+                     const std::string& name = "lambda_block");
+
+
+    inline void add_item_tag(unsigned int which_output,
```
mbr0wn, line 55: The one thing I don't like about this is that these are moved from `protected` to `public` (which is necessary to access them in the lambda).

We could add some protection to make sure they're only called from within the lambda, not otherwise. Does anyone think that's necessary?

gmeyerho (approved):

mbr0wn:
@marcusmueller sez (on chat):

> does that block actually need to have access to add_item_tags? That breaks the functional aspect of it all. I'd rather let the lambda return a vector of tags; it's going to have to allocate these tags anyway, and whether in emplaces them on a vector or directly pushes them through add_item_tag makes little performance difference, I'd guess.
(and for all the lambdas that don't produce tags, fine, return {}; does the trick and usually has zero heap alloc overhead in practice. Might as well also do a std::optional<std::vector<tag>>, but that feels overkill)

Anytime someone uses functional programming as an argument, I'm usually on board. In this case, I think it's not as helpful, but it's partly a usability/ergonomic thing, meaning it's somewhat subjective, and I'm curious what others think. Here's my arguments for keeping it non-functional (aka the way it is):
- You don't only need `add_item_tags()`, you also need a bunch of other methods from the block class in your lambda, such as `nitems_written()`. This means you already need the `self` object no matter how we proceed.
- The way it's currently implemented, the code in the function object you pass in looks exactly like code that you'd put into a `general_work()` function elsewhere. I think that makes it more readable/comparable. If you rename `self` to `this`, you can even copy/paste the same code from a lambda block's lambda to a regular block's `general_work()`.
- To take this to the extreme, we're also not returning the output buffers from the lambda, but are writing to them the same way `general_work()` does.
- There's probably some performance penalty for returning tags from your lambda, and then having the `lambda_block`'s `general_work()` function putting them into the right place. Maybe not a great argument, there's probably some move operation we can use to reduce the penalty.
- Note that the lambda already returns the return value for `general_work()`. So if we would return tags, then we'd have to return a tuple of `<int, std::vector<tag_t>>` or something like that. Again, that's a departure from other code.

The main reason for going with Marcus' approach (other than the aesthetic pleasure of doing things the functional way) is that we don't need to make `add_item_tags()` public on the lambda block. Which I think is a really good reason. Personally, I just think the reasons for making the lambda look like `general_work()` outway the reasons for making it look different.

willcode (commented):
From a technical perspective, I think this is cool, and makes it easier to implement blocks without having to go through the whole make-a-block process. It has a strange feel, though, since the inner workings of blocks doesn't currently show up in user programs. Would it be usefult to limit this to sync blocks, hide the consume_each from the user program (not sure about add_item_tag), etc?

mormj:
> Here's my arguments for keeping it non-functional (aka the way it is)

I agree with your assessment here - the goal of 4.0 is to be more functional and have block work functions to behave as Marcus described.  To change the paradigm within 3.x would be big a departure.  This looks to be consistent with a detached work method that is intertwined the same way with the runtime as a non-lambda block, which I think is the right approach.

mbr0wn:
> From a technical perspective, I think this is cool, and makes it easier to implement blocks without having to go through the whole make-a-block process. It has a strange feel, though, since the inner workings of blocks doesn't currently show up in user programs. Would it be usefult to limit this to sync blocks, hide the consume_each from the user program (not sure about add_item_tag), etc?

Well, the reason I wrote this was for the C++ formatter for the fosphor display. And there, I need it to be a non-sync block.

I did think about making a separate sync lambda block, but I didn't have a use case for it.

mbr0wn:
> Hey :)
> 
> I'm fine with saying "this isn't really the _functional_ functional block, we might implement that later or just in 4.0" (#2666 has been around a while, and I've never managed to write anything upstreamable, something was always too complex, or didn't deal with usual use cases ;-) ).
> 
> Naming-wise, I'm a bit uneasy calling this here the lambda block, as it really doesn't feel like a typical lambda if the first argument is a `this` pointer. (Calling a method does the same – pre-C++11 we'd have written this as a class with a `()` operator or similar instead of accepting a `std::function`.)

The name comes from the fact that most likely, you'd call it with a C++ "lambda" (`[](){}`) as an argument. I'm not married to the name, and will happily change it to something that people think is better.

> I will say that the API is not pretty. From a practical point of view, writing a function or C++ lambda expression that effectively takes a `*this` and four other parameters: `class foobar : public gr::simpleblock {…}` wouldn't be that much more typing and follow established patterns.

API can probably be improved. I'm not sure I understand what you're saying with the second sentence.

> Could you elaborate on why this is the technical solution to the C++ formatter for a fosphor display, and not simply a block that does that? I feel like I'm criticizing something nice because I don't understand what it's for!

You can check out that PR. It's not a *requirement*, it's theoretically possible to not have this lambda block. Like @mormj calls it, it's a block with a detached work function. I've often had this problem where I'm solving something with a hier block, but there's one little functionality that's missing. And then I have to add a new block just for that, and add headers, and then that changes the API of gr-blocks or whatever... I really think the ability to define blocks on the fly is useful. The fosphor formatter block was just a reason for me to finally put this into the tree.

> Code-organization wise, this is closer to a new block type than "just another block to go into the catch-all gr-blocks", so if we like the block, we might want to consider moving it to `gnuradio-runtime` instead, but that's really just an unfounded feeling more than an opinion that I could motivate.

Sure if the majority prefers that, I'll move it there. I would leave the QA in gr-blocks so I can use the vector sink/source.

marcusmueller:
>  I really think the ability to define blocks on the fly is useful.

Absolutely! That's another reason to have it in runtime.

> I'm not sure I understand what you're saying with the second sentence.

Sorry for the lack of clarity: What I meant is that, as soon as you write a function (or a `[](){}`) that takes so many arguments and even a very `this`-like pointer, code gets a bit complicated. 
If you then pass that function to something else, you can only hope that you remembered the API of that function correctly.

C++ already comes with a mechanism to wrap a function in a type-safe manner and also provide a wrapper pointer for state, it's classes with methods ;-) 

So, the "problem" I have with your approach is that it feels like the complexity of writing that lambda or free-standing function that fits the API is about as complex as writing a subclass of `gr::block`¹:

```c++
    auto lambda = gr::blocks::lambda_block::make(
        [](gr::blocks::lambda_block* self,
           int noutput_items,
           gr_vector_int& ninput_items,
           gr_vector_void_star& output_items) -> int {
            const float* in = static_cast<const float*>(input_items[0]);
            float* out = static_cast<float*>(output_items[0]);
            const size_t items_to_process =
                static_cast<size_t>(std::min(noutput_items, ninput_items[0]));
            if (VERBOSE)
                std::cout << "[lambda_block_run] general_work(): Items to process: "
                          << items_to_process << std::endl;
            for (size_t i; i < items_to_process; i++) {
                out[i] = 2 * in[i];
            }
            self->consume_each(items_to_process);
            self->add_item_tag(0, 0, pmt::string_to_symbol("test"), pmt::PMT_T);
            return static_cast<int>(items_to_process);
        },
        gr::io_signature::make(1, 1, sizeof(float)),
        gr::io_signature::make(1, 1, sizeof(float)),
        "my_doubler");
```
vs

https://github.com/marcusmueller/minblock/blob/d45e56ada9d6d66a9bd6fd401f95eb73993b572e/doubler_test.cc#L17-L44
```c++
using ios = gr::io_signature;

class doubler : virtual public gr::block
{
public:
    doubler()
        : gr::block("my_doubler",
                    ios::make(1, 1, sizeof(float)),
                    ios::make(1, 1, sizeof(float))){};
    ~doubler(){};
    int general_work(int noutput,
                     gr_vector_int& ninput,
    {
        static pmt::pmt_t symbol = pmt::mp("test");
        auto in_ptr = reinterpret_cast<const float*>(input_items[0]);
        auto out_ptr = reinterpret_cast<float*>(output_items[0]);
        auto items_to_process = std::min(ninput[0], noutput);
        d_logger->debug("general_work(): Items to process: {}", items_to_process);

        for (auto in = in_ptr; in != in_ptr + items_to_process; ++in) {
            *(out_ptr++) = 2 * *in;
        }

        consume_each(items_to_process);
        add_item_tag(0, 0, symbol, pmt::PMT_T);
        return items_to_process;
    }
};
```

So, I'd say, with a tolerance of 3 lines, the non-lambda_block code is about as complicated.

>>Could you elaborate on why this is the technical solution to the C++ formatter for a fosphor display, and not simply a block that does that?
> You can check out that PR

Sorry again for my lack of communication skill there: What you've done in [`fosphor_formatter_impl::_process_histogram`](https://github.com/gnuradio/gnuradio/pull/6515/files#diff-a02b689b8c29e1c61adf246d767549955f4b16a86464dd91dbe7b6ab01e18ec5R172) can be solved with a nested `gr::block` derivative class, passing a reference to the "parenting" `fosphor_formatter_impl` to its ctor, to access things like `d_maxhold_buf`, `d_fftsize`  in the `general_work`. (Or, and that's also done in a few places, multiple-inheritance, though I must say, I find that confusing at times as well.)
<br>
Instead, I guess on purpose, you inverted that relationship, and pass a `gr::block` reference into a function so that you can access a select and with extra code exposed amount of members² of `gr::block`. The motivation for *that* was what I was hoping you elaborate on!

----

¹actually, now that I think about it, how well does your `general_work`-carrying block work without implementing `forecast`? The non-`overrid`d`e`n `forecast` is effectively that of a `sync_block`.

² I do have further requests for this PR, namely to not even start doing things like `#define VERBOSE true; if(VERBOSE)std::cout << "foobar: << d_n;`  but to expose and use `d_logger` of the `gr::block`. My fear is that this is the start of replication the whole `gr::block` API: today you only needed `add_item_tag`. What about `get_tags_in_range`? That feels natural to have when you have a tag-setting method. So, worst case, we end up replicating the `protected` API of `gr::block`, `gr::basic_block` by copy and paste to the `public` scope of the `lambda_block`? That would seem like an antipattern…

mbr0wn:
> My fear is that this is the start of replication the whole `gr::block` API: today you only needed `add_item_tag`. What about `get_tags_in_range`? That feels natural to have when you have a tag-setting method. So, worst case, we end up replicating the `protected` API of `gr::block`, `gr::basic_block` by copy and paste to the `public` scope of the `lambda_block`? That would seem like an antipattern…

That was the main concern I added up top in the original PR. It's obviously an antipattern, and as mentioned, I also don't think it's great.

I'm going to close this PR; I don't have time to deal with a drawn-out conversation right now. @marcusmueller, your comments are all valid but they are not actionable, and of course there's the dreaded ellipsis at the end of your comment which gives me a lot of space for interpretation.

You're 100% right that we can refactor the formatter PR without the lambda block, so I'll do that to not further derail the discussion (@gmeyerho FYI).

Issue #3423, opened 2020-04-28
QT GUI Message Edit Box sending text on loss of focus


duggabe:
If a user types text into a Message Edit Box but DOES NOT PRESS ENTER, and then clicks on a different screen (such as the GRC screen), the Message Edit Box sends the text as if the enter key had been pressed. Here are two test examples:
* https://github.com/gnuradio/gnuradio/blob/master/gr-qtgui/examples/test_digitalnumcontrol.grc
* https://github.com/gnuradio/gnuradio/blob/master/gr-qtgui/examples/test_qtgui_msg.grc

Tested with Raspian Buster with 3.8.1.0; and Ubuntu 20.04 with 3.9.0.0-git

PR #6645, opened 2023-04-20
qtgui: Replace `QLineEdit::returnPressed()` with `editingFinished()`

ryanvolz:
## Description

This affects the QT GUI Entry, Message Edit Box, and Range widgets which use QLineEdit for text entry. Previously, these text edit boxes would not trigger the callback to update the variable unless Return/Enter was pressed. With this change, the variable is updated both in that case and also when the field loses focus, e.g. by clicking somewhere else on the display. See
https://doc.qt.io/qtforpython-5/PySide2/QtWidgets/QLineEdit.html#detailed-description for a description of the two signals.

I'm not sure if `returnPressed` was chosen intentionally to avoid the behavior of `editingFinished`, but I at least find the `editingFinished` behavior to be what I'd expect. There's no visual indication that Return/Enter is needed to set the corresponding variable, so it is natural to enter some text, change focus to another widget (like, for example, clicking a button to act on the new value in the text box), and assume that the variable has been updated to the value shown in the text box because that is what's being displayed. Maybe there are use cases for the current behavior that I'm not aware of, so I'm open to discussion/rejection.

Find/replace also turned up the use of the Entry widget in two of the UHD apps, so this PR also updates those.

QT GUI Entry, Message Edit Box, and Range widgets

## Testing Done

I tested this locally with a QT GUI Entry by modifying the Python script directly, and the `editingFinished` signal triggered as described. I have not tested the other two widgets or the UHD apps, but there shouldn't be any issues with those either.

## Checklist

### DIFF
--- a/gr-qtgui/grc/qtgui_entry.block.yml
+++ b/gr-qtgui/grc/qtgui_entry.block.yml
@@ -43,7 +43,7 @@ templates:
         ${win}.addWidget(Qt.QLabel("${no_quotes(label,repr(id))}" + ": "))
         self._${id}_line_edit = Qt.QLineEdit(str(self.${id}))
         self._${id}_tool_bar.addWidget(self._${id}_line_edit)
-        self._${id}_line_edit.returnPressed.connect(
+        self._${id}_line_edit.editingFinished.connect(
             lambda: self.set_${id}(${type.conv}(str(self._${id}_line_edit.text()))))
         ${gui_hint() % win}
 
@@ -68,7 +68,7 @@ cpp_templates:
         this->_${id}_tool_bar->addWidget(_${id}_label);
         this->_${id}_line_edit = new QLineEdit(QString::number(this->${id}));
         this->_${id}_tool_bar->addWidget(this->_${id}_line_edit);
-        QObject::connect(this->_${id}_line_edit, &QLineEdit::returnPressed, this, [this] () {this->set_${id}(this->_${id}_line_edit->text().toInt());});
+        QObject::connect(this->_${id}_line_edit, &QLineEdit::editingFinished, this, [this] () {this->set_${id}(this->_${id}_line_edit->text().toInt());});
 
         ${gui_hint() % win}
 documentation: |-
--- a/gr-qtgui/lib/edit_box_msg_impl.cc
+++ b/gr-qtgui/lib/edit_box_msg_impl.cc
@@ -134,7 +134,7 @@ edit_box_msg_impl::edit_box_msg_impl(data_type_t type,
     d_vlayout->addItem(d_hlayout);
     d_group->setLayout(d_vlayout);
 
-    QObject::connect(d_val, SIGNAL(returnPressed()), this, SLOT(edit_finished()));
+    QObject::connect(d_val, SIGNAL(editingFinished()), this, SLOT(edit_finished()));
 
     d_msg = pmt::PMT_NIL;
 
--- a/gr-qtgui/python/qtgui/range.py.cmakein
+++ b/gr-qtgui/python/qtgui/range.py.cmakein
@@ -253,7 +253,7 @@ class RangeWidget(QtWidgets.QWidget):
             self.notifyChanged = slot
 
             self.setMaximumWidth(100)
-            self.returnPressed.connect(self.changed)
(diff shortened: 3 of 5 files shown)


willcode:
Either behavior can be annoying. Agree that with no guidance, having to press return, or tab out (focus policy is another source of pain) is confusing. We should have things consistent across the system, includeding in GRC. I really haven't looked that hard.

mormj:
I agree with @willcode  - I've gotten confused by it in the past in that it's not clear when the value will be applied.  I'd hold on it to make sure we are consistent across all widgets.  If this is the one anomaly, then backport, but if there is broader change, then it should probably be introduced on 3.11 only.

ryanvolz:
> Either behavior can be annoying. Agree that with no guidance, having to press return, or tab out (focus policy is another source of pain) is confusing.

Maybe you already got this, but the phrasing makes me think I might need to clarify. The existing behavior is that return/enter is required to trigger the value to change. The new behavior would be to keep return/enter as a trigger, and *also add* losing focus as a trigger to update the value. Since it is adding a trigger, my concern would be someone who relies on the value remaining unchanged when focus is lost and return/enter is not pressed (I can't think of a use case, just leaving open the possibility), not anyone who just relies on return/enter as a trigger (since that will still work).

Anyway, I completely agree that being cautious and slow to roll out a change like this is wise. We can wait to see if anyone finds it disruptive before considering a backport.

willcode:
Got it - I was thinking that either requiring return, or having something trigger when you weren't really done (e.g., going off to cut some text from somewhere else) have annoyed me in the past.

duggabe:
Does this affect https://github.com/gnuradio/gnuradio/issues/3423 ?

ryanvolz:
> Does this affect #3423 ?

Yes, it does undo #5592 which closed that. I wasn't aware of that particular previous discussion. Sounds like there's more to discuss.

willcode:
I think that's why the Hold Backport label. I'd suggest that any editable text UI element that performs an immediate action should have some sort of `send`. The digital number control may be a little different. It would be good to make `main` and `maint-3.10` behave the same when we can.

duggabe:
Using GNU Radio v3.11.0.0git-528-g9b3183ae / 3.11.0git I have verified that the 'editingFinished' behavior works using
- test_qtgui_msg.grc
- test_digitalnumcontrol.grc

https://github.com/duggabe/gr-morse-code-gen presents a case where 'editingFinished' is not desired.

I recommend adding an optional parameter to choose 'editingFinished' (the default) or 'returnPressed'.

ryanvolz:
> I recommend adding an optional parameter to choose 'editingFinished' (the default) or 'returnPressed'.

I like that idea.

This PR touched 3 blocks:
1. QT Gui Entry
2. Message Edit Box
3. Range widget when text entry is enabled

Do you think all of those would benefit from the option to go to returnPressed instead of editingFinished?

willcode:
The Message Edit Box sounds like the most likely to need this. Typically, the Gui Entry and Range Widget are literals of some sort and it seems less likely focus will be lost before the value is complete. Technically, all 3 have effects when set, but generating a message seems slightly different than changing a variable. Kind of gray, though.

willcode:
@ryanvolz @mormj @duggabe 
I'd like to revert this. Common ways of saying you're done entering a value include tabbing, clicking on another widget, or hitting OK/Apply somewhere. Or the value is evaluated on every edit. Having values applied because the pointer strays off or the user is doing cut/paste is a problem. The Range widget is not really a problem because values tend to be atomic.

Really, a control panel should have an Apply for any values that don't take effect as they're being typed. We just don't have a way of doing that.

ryanvolz:
See #6867 which implements what I think we've settled on here.

> Really, a control panel should have an Apply for any values that don't take effect as they're being typed. We just don't have a way of doing that.

It's actually my attempt at doing something like this that lead to me wanting `editingFinished` behavior for QT GUI Entry. I'm using a "QT GUI Entry" for a user to put in a name paired with a "QT GUI Msg Push Button" to trigger an action that uses that entered name. With `returnPressed`, the user would need to both press "Enter" on the text entry field *and then* click the button. With `editingFinished`, the value that gets used when clicking the button is always the text that is visible, which makes more intuitive sense.

PR #6651, opened 2023-05-04
uhd: pybind: use module_local() for types registered by uhd

gmeyerho:
If a type is registered in pybind by uhd and then again within gr-uhd, an error will be thrown if a user imports uhd and imports uhd from gnuradio since pybind by default makes bindings global.

Add py::module_local() to the uhd types registered within gnuradio to keep those bindings localized

## Description
Adds pybind11::module_local() for types that are defined by uhd so that a user can import from uhd and gnuradio.uhd

## Related Issue
I don't know how common the use-case is, but currently if a user does the following:
```
from gnuradio import uhd
from uhd import usrp
```

It will throw an ImportError as numerous types are registered via pybind in both gr-uhd and gnuradio. This change localizes the  gr-uhd bindings. Now the above code no longer throws the import error

gr-uhd

## Testing Done
Sanity checked a  few of the gr-uhd examples such as rfnoc_replay.grc and rfnoc_radio_ddc_multichan.grc

Also checked that importing uhd from gnuradio and importing from uhd itself no longer throws an error

## Checklist

### DIFF
--- a/gr-uhd/python/uhd/bindings/amsg_source_python.cc
+++ b/gr-uhd/python/uhd/bindings/amsg_source_python.cc
@@ -26,7 +26,7 @@ void bind_amsg_source(py::module& m)
 
 
     py::class_<amsg_source, std::shared_ptr<amsg_source>>(
-        m, "amsg_source", D(amsg_source))
+        m, "amsg_source", D(amsg_source), py::module_local())
 
         .def(py::init(&amsg_source::make),
              py::arg("device_addr"),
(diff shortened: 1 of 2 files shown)


mormj:
I've never tried any of this - the only way I know to use types from external modules is `py::module::import("pmt");` for example - but I'm not sure if this would achieve the same thing.  I think the issue with that is that we would have to ensure that both modules were compiled with the same version of pybind11

PR #6665, opened 2023-05-13
soapy: sdrplay agc and gain fixes

### LINKED ISSUE #6652: Soapy SDRPlay Source fails with AGC=True
### What happened?

I changed the RF Options / AGC to `True` and the source doesn't start. The warning and error in log output do not show with AGC=False

Expected automatic gain to be enabled along with a new parameter to choose a level, AFAIK there are parameters to adjust the AGC.

### System Information

OS: Mac
GR Installation Method: homebrew

### GNU Radio Version

3.10 (maint-3.10)

### Specific Version

3.10.6.0

1. Add Soapy SDRPlay Source
2. Edit Preferences; set Center Freq, and change AGC to `True`
3. Add Null Sink and connect
4. Execute

### Relevant log output

```shell
[INFO] devIdx: 0
[INFO] SerNo: B0002P0001
[INFO] hwVer: 1
[INFO] rspDuoMode: 0
[INFO] tuner: 1
[INFO] rspDuoSampleFreq: 0.000000
[WARNING] Not updating IFGR gain because AGC is enabled
[INFO] Using format CF32.
[ERROR] error in activateStream() - Init() failed: sdrplay_api_Fail



SoapySDRUtil -probe show following:

----------------------------------------------------
-- Device identification
----------------------------------------------------
  driver=SDRplay
  hardware=RSP1
  sdrplay_api_api_version=3.070000
  sdrplay_api_hw_version=1

----------------------------------------------------
-- Peripheral summary
----------------------------------------------------
  Channels: 1 Rx, 0 Tx
  Timestamps: NO
  Other Settings:
     * RF Gain Select - RF Gain Select
       [key=rfgain_sel, default=1, type=string, options=(0, 1, 2, 3)]
     * IQ Correction - IQ Correction Control
       [key=iqcorr_ctrl, default=true, type=bool]
     * AGC Setpoint - AGC Setpoint (dBfs)
       [key=agc_setpoint, default=-30, type=int, range=[-60, 0]]

----------------------------------------------------
-- RX Channel 0
----------------------------------------------------
  Full-duplex: NO
  Supports AGC: YES
  Stream formats: CS16, CF32
  Native format: CS16 [full-scale=32767]
  Antennas: RX
  Corrections: DC removal
  Full gain range: [0, 42] dB
    IFGR gain range: [20, 59] dB
    RFGR gain range: [0, 3] dB
  Full freq range: [0.01, 2000] MHz
    RF freq range: [0.01, 2000] MHz
    CORR freq range:  MHz
  Sample rates: 0.0625, 0.096, 0.125, 0.192, 0.25, ..., 6, 7, 8, 9, 10 MSps
  Filter bandwidths: 0.2, 0.3, 0.6, 1.536, 5, 6, 7, 8 MHz
```

willcode:
## Description
SDRPlay gains are actually attenuations
- grdb: gain reduction of 20 to 59 dB
- lna state: model-specific attenuation steps for LNA

LNA state is independent of AGC. Gain reduction is N/A when AGC is on.

SoapySDR attempts to do automatic distributions of gains if setGain() is called without a gain name. This is what gr-soapy was trying to do and it resulted in invalid gain values.

AGC has a setpoint, which was not being set. This is probably why a crash was happening when AGC was set.

This change:
- Adds the AGC setpoint
- Separates gain and lna state
- Makes gain positive instead of the previous negative range
- LNA state is still an attenuation with device-specific stages
- Sets gain when agc is turned off (not done by driver)
- Defers setting IFGR gain when agc is on, to avoid warnings
- Enables bias tee control
- gain, lna_state, agc, agc_setpoint and bias are runtime settable

## Related Issue
Fixes #6652

Soapy SDRPlay

## Checklist

### DIFF
--- a/gr-soapy/grc/soapy_sdrplay_source.block.yml
+++ b/gr-soapy/grc/soapy_sdrplay_source.block.yml
@@ -35,6 +35,12 @@ parameters:
     default: 'RX'
     hide: part
 
+  - id: bias
+    label: Bias Tee Power
+    dtype: bool
+    default: 'False'
+    hide: part
+
   - id: center_freq
     label: 'Center Freq (Hz)'
     category: RF Options
@@ -55,12 +61,26 @@ parameters:
     default: 'False'
     hide: part
 
-  - id: gain
-    label: 'RF Gain (-47dB - 0dB)'
+  - id: agc_setpoint
+    label: AGC Setpoint (-60dB - 0dB)'
     category: RF Options
-    dtype: real
-    default: '-20'
-    hide: ${'all' if agc else 'part'}
+    dtype: int
+    default: '-30'
+    hide: part
+
+  - id: gain
+    label: 'Gain (0dB - 39dB)'
+    category: RF Options
+    dtype: int
+    default: '20'
+    hide: part
+
+  - id: lna_state
+    label: 'LNA State (atten step)'
+    category: RF Options
+    dtype: int
+    default: '3'
+    hide: part
 
 inputs:
   - domain: message
@@ -74,29 +94,64 @@ outputs:
 
 templates:
   imports: from gnuradio import soapy
-  make: |
+
+  make: |-
       None
+      ## Negative value for agc_setpoint is placed in parens by GRC.
+      ## This code turns it into an int and bounds checks
+      _agc_setpoint = int(${agc_setpoint})
+      _agc_setpoint = max(min(_agc_setpoint, -20), -70)
+
       dev = 'driver=sdrplay'
       stream_args = ''
       tune_args = ['']
       settings = ['']
+
+      ## Intercept the agc callback and restore the gain when agc is
+      ## disabled. The driver does not do this. LNA state is preserved.
+      def _set_${id}_gain_mode(channel, agc):
+          self.${id}.set_gain_mode(channel, agc)
+          if not agc:
+              self.set_${id}_gain(channel, self.gain)
+      self.set_${id}_gain_mode = _set_${id}_gain_mode
+
+      ## Intercept the gain callback. The soapysdrplay driver prints an
+      ## error message every time gain is adjusted with agc on. Defer
+      ## setting IFGR until agc is turned off. Bounds check.
+      def _set_${id}_gain(channel, gain):
+          if not self.agc:
+              self.${id}.set_gain(channel, 'IFGR', min(max(59 - gain, 20), 59))
+      self.set_${id}_gain = _set_${id}_gain
+
+      ## Valid values for LNA state (RFGR here) vary by model, so runtime
+      ## changes may cause errors.
+      def _set_${id}_lna_state(channel, lna_state):
+              self.${id}.set_gain(channel, 'RFGR', min(max(lna_state, 0), 9))
+      self.set_${id}_lna_state = _set_${id}_lna_state
 
       self.${id} = soapy.source(dev, "${type}", 1, ${dev_args},
                                 stream_args, tune_args, settings)
       self.${id}.set_sample_rate(0, ${samp_rate})
       self.${id}.set_bandwidth(0, ${bandwidth})
       self.${id}.set_antenna(0, ${antenna})
-      self.${id}.set_gain_mode(0, ${agc})
       self.${id}.set_frequency(0, ${center_freq})
       self.${id}.set_frequency_correction(0, ${freq_correction})
-      self.${id}.set_gain(0, min(max(-${gain}, 0.0), 47.0))
+      self.${id}.write_setting('biasT_ctrl', ${bias})
+      self.${id}.write_setting('agc_setpoint', ${agc_setpoint})
+      self.set_${id}_gain_mode(0, ${agc})
+      self.set_${id}_gain(0, ${gain})
+      self.set_${id}_lna_state(0, ${lna_state})
+
   callbacks:
     - set_sample_rate(0, ${samp_rate})
     - set_bandwidth(0, ${bandwidth})
     - set_antenna(0, ${antenna})
-    - set_gain_mode(0, ${agc})
     - set_frequency(0, ${center_freq})
     - set_frequency_correction(0, ${freq_correction})
-    - set_gain(0, min(max(-${gain}, 0.0), 47.0))
+    - write_setting('biasT_ctrl', ${bias})
+    - write_setting('agc_setpoint', ${agc_setpoint})
+    - self.set_${id}_gain_mode(0, ${agc})
+    - self.set_${id}_gain(0, ${gain})
+    - self.set_${id}_lna_state(0, ${lna_state})
 
 file_format: 1


gr-soapy/grc/soapy_sdrplay_source.block.yml
```diff
@@ -76,7 +90,7 @@ templates:
   imports: from gnuradio import soapy
   make: |
       None
-      dev = 'driver=sdrplay'
+      dev = 'driver=sdrplay,agc_setpoint=-30'
```
tko, line 93: correct me if I'm wrong, but the `agc_setpoint` parameter above isn't connected to anything as this is hardcoded?

gr-soapy/grc/soapy_sdrplay_source.block.yml
```diff
@@ -55,12 +55,26 @@ parameters:
...
   - id: gain
-    label: 'RF Gain (-47dB - 0dB)'
+    label: 'Gain (0dB - 39dB)'
     category: RF Options
-    dtype: real
-    default: '-20'
-    hide: ${'all' if agc else 'part'}
+    dtype: int
+    default: '20'
+    hide: part
```
tko, line 70: Could hide manual gain control like before when agc is enabled, to be more clear they're mutually exclusive?

willcode:
Right. There's a piece of template code I still have to figure out. Ok, I'll try defaulting the lna setting to something all rsps can do.

gr-soapy/grc/soapy_sdrplay_source.block.yml
willcode: That made it difficult to build a flowgraph with GUI controls for agc and gain, since either setpoint or gain would be hidden based on mode.

willcode:
I figured out the Mako-GRC-foo to set the setpoint, and also restore gain after agc is turned off. @tko could you give this one a try?

gr-soapy/grc/soapy_sdrplay_source.block.yml
willcode: When using GUI elements (radio buttons to select agc) hiding the fields makes it difficult to set up, so I'll keep them all visible.

gr-soapy/grc/soapy_sdrplay_source.block.yml
```diff
@@ -86,17 +106,31 @@ templates:
...
   callbacks:
     - set_sample_rate(0, ${samp_rate})
     - set_bandwidth(0, ${bandwidth})
     - set_antenna(0, ${antenna})
-    - set_gain_mode(0, ${agc})
     - set_frequency(0, ${center_freq})
     - set_frequency_correction(0, ${freq_correction})
-    - set_gain(0, min(max(-${gain}, 0.0), 47.0))
+    - self.set_gain_mode(0, ${agc})
+    - set_gain(0, 'IFGR', min(max(59 - ${gain}, 20), 59))
```
tko, line 133: I don't know how these callbacks work but it'd be nice to _not_ make this call when `agc=True` since we know it'll just spam the console, and when agc gets unset the value should be reapplied.

tko (commented):
I discovered the official(?) recommendation is to start with AGC=True
https://github.com/pothosware/SoapySDRPlay2/issues/44#issuecomment-443496435

gr-soapy/grc/soapy_sdrplay_source.block.yml
```diff
@@ -74,9 +88,15 @@ outputs:
 
 templates:
   imports: from gnuradio import soapy
-  make: |
-      None
-      dev = 'driver=sdrplay'
+
+  var_make: hello
```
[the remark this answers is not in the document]
willcode: Ha, I went through a lot of different tests before figuring out how to do this!

gr-soapy/grc/soapy_sdrplay_source.block.yml
```diff
@@ -86,17 +106,31 @@ templates:
...
-      self.${id}.set_gain(0, min(max(-${gain}, 0.0), 47.0))
+      self.${id}.set_gain_mode(0, ${agc})
+      ##
+      ## Bounds check. Valid values for LNA state (RFGR here)
+      ## vary by model, so runtime changes may cause errors.
+      self.${id}.set_gain(0, 'IFGR', min(max(59 - ${gain}, 20), 59))
+      self.${id}.set_gain(0, 'RFGR', min(max(${lna_state}, 0), 9))
+      ##
+      ## Intercept the agc callback and restore the gain when agc is
+      ## disable. The driver does not do this. LNA state is preserved.
```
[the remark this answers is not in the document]
willcode: Here's a GRC where is works for me ...
https://pastebin.com/QVdHBSnV

gr-soapy/grc/soapy_sdrplay_source.block.yml
willcode: No hello, spam warnings suppressed.

gr-soapy/grc/soapy_sdrplay_source.block.yml
```diff
@@ -74,9 +88,15 @@ outputs:
...
-      dev = 'driver=sdrplay'
+
+  var_make: hello
+
+  make: |-
+      ## Negative value for agc_setpoint is placed in parens by GRC.
+      ## This code turns it into an int and bounds checks
+      _agc_setpoint = int(${agc_setpoint})
+      _agc_setpoint = max(min(_agc_setpoint, -20), -70)
+      dev = f'driver=sdrplay,agc_setpoint={_agc_setpoint}'
```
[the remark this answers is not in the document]
willcode: I tried dev_args first and it didn't work. Maybe I was doing something wrong.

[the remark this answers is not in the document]
willcode: The problem is that the user can specify dev_args. It would only take a coulple lines of code do merge them. This works, so I'll stick with this for now.

gr-soapy/grc/soapy_sdrplay_source.block.yml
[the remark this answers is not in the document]
tko: Ok, everything works nicely. Turns out I had two copies of the yaml and I kept updating the wrong one... 🤦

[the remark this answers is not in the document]
willcode: Yah, that NEVER happens to me 😀

gr-soapy/grc/soapy_sdrplay_source.block.yml
```diff
@@ -86,17 +105,44 @@ templates:
...
       self.${id}.set_frequency(0, ${center_freq})
       self.${id}.set_frequency_correction(0, ${freq_correction})
-      self.${id}.set_gain(0, min(max(-${gain}, 0.0), 47.0))
+      self.${id}.set_gain_mode(0, ${agc})
+      ##
+      ## Bounds check. Valid values for LNA state (RFGR here)
+      ## vary by model, so runtime changes may cause errors. Set
+      ## gain (IFGR) only if agc is off.
+      self.${id}.set_gain(0, 'RFGR', min(max(${lna_state}, 0), 9))
+      if not self.agc:
```
tko, line 116: Putting this source in a separate top block fails because it's trying to access `agc` that does not exist, in that top block.

I think this ought to be `if not ${agc}` -- however I'm unsure how the methods should access `self.gain` and `self.agc`

willcode: "Putting this source in a separate top block" ... do you mean cutting/pasting the source code, or something you can do with GRC?

I'll change to `${agc}` to be more like other params. In the code I'm looking at, though, it doesn't seem like it makes much difference:
`self.agc = agc = True`

We're hitting the limit of what these templates can do!

willcode: I do see that placing the block in a hier block does not work well.

willcode: Now works fine in a hier block.

Issue #6604, opened 2023-03-24
Allow Soapy message commands to set non-channel-specific device settings


ryanvolz:
### Feature Description

I have been bit by the following [TODO](https://github.com/gnuradio/gnuradio/blob/627493488b365aa9ce2464bc059db6fa11757bd3/gr-soapy/lib/block_impl.cc#L1465) for gr-soapy's `block_impl::cmd_handler_setting` function:
```
// TODO: any way to call channel-less version?
void block_impl::cmd_handler_setting(pmt::pmt_t val, size_t channel)
```
Basically I want to send a "setting" command to a Soapy Source block to toggle a bias tee, but that is not currently possible because the setting in question ("biastee") is a device setting that could be set by the [channel-less version of `write_setting`](https://github.com/gnuradio/gnuradio/blob/627493488b365aa9ce2464bc059db6fa11757bd3/gr-soapy/lib/block_impl.cc#L1085) but the command handler only has the ability to call the [channel-specific version of `write_setting`](https://github.com/gnuradio/gnuradio/blob/627493488b365aa9ce2464bc059db6fa11757bd3/gr-soapy/lib/block_impl.cc#L1113-L1115) where "biastee" is not valid. Currently the only way to use these channel-less settings is by specifying them as device arguments during initialization.

Note that not setting a `chan` key in the command causes the current command handler to loop through all active channels and apply that command to each of them, which is where this gets forced into channel-specific code paths in the first place. For other command handlers this works out OK in the end because even though all of the command handlers take a `channel` argument, a lot of them just ignore it as irrelevant. The "setting" command seems somewhat unique in that the Soapy API defines both channel-specific and channel-less settings, although in practice a quick survey of Soapy device modules didn't turn up any that use channel-specific settings.

I can imagine different ways of implementing this:

1. It's Soapy's problem, the channel-specific settings should fall back to channel-less settings at the Soapy API level.
2. It's gr-soapy's problem, the channel-specific `write_settings` should fall back to the channel-less `write_settings` in the gr-soapy block implementation.
3. Explicit is better than implicit and we should avoid fallbacks, so a "setting" command without a channel specified should go to the channel-less `write_settings` and not loop through all of the available channels. This requires changing the message command handler, which has implications for all other commands. This opens up even more possibilities.

Thoughts?

### Feature Urgency

medium (would be nice to have in the near future)

willcode:
I see the iteration over channels only at init. The handler always assumes there is a channel.

[...]

PR #6666, opened 2023-05-13
soapy: rtlsdr bias tee and restore gain after agc

willcode:
## Description

Enable select of bias tee via "bias" param.

Restore previous gain when AGC is disabled.

## Related Issue
Every so slightly addresses #6604 

Soapy RTLSDR Source

## Checklist

### DIFF
--- a/gr-soapy/grc/soapy_rtlsdr_source.block.yml
+++ b/gr-soapy/grc/soapy_rtlsdr_source.block.yml
@@ -47,7 +47,13 @@ parameters:
     category: RF Options
     dtype: real
     default: '20'
-    hide: ${'all' if agc else 'part'}
+    hide: 'part'
+
+  - id: bias
+    label: 'Bias Tee Power'
+    dtype: bool
+    default: 'False'
+    hide: 'part'
 
 inputs:
   - domain: message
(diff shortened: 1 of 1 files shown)


willcode:
@ryanvolz Could you see if this works for you? This doesn't solve the problem we were talking about before (setting via message) but at least users can set the bias tee from the "easy" block.

ryanvolz:
> @ryanvolz Could you see if this works for you? This doesn't solve the problem we were talking about before (setting via message) but at least users can set the bias tee from the "easy" block.

I was trying this with an Airspy but moreover through the Soapy Custom Source block to allow configuring for different devices from a config file. So the message approach is the only thing that makes sense for toggling the bias tee during runtime in that context.

Still, I tried to give this a spin with an RTL-SDR and its bias tee, and that has led me to discover that my `soapysdr-module-rtlsdr` conda-forge package doesn't seem to have the bias tee stuff enabled. So I'll have to go fix that before I can confirm that this works for me.

ryanvolz (approved):
Both the AGC and bias tee changes work for me.

willcode:
@ryanvolz Do you use an AirspyHF? The blocks has gain settings which are not available on my AirspyHF+, so I'm wondering if anyone as an AirspyHF of some sort that requires these settings.

ryanvolz:
> @ryanvolz Do you use an AirspyHF? The blocks has gain settings which are not available on my AirspyHF+, so I'm wondering if anyone as an AirspyHF of some sort that requires these settings.

I have access to one purchased somewhat recently that I could test something with. What do you want to know? I know that it has three gain stages and I've been able to set each of them independently before.

willcode:
Wondering if the gain settings in the Soapy AirspyHF block work. Also the AGC and freq correction params. It's the AirspyHF+, flat package with a single SMA? Mine could be old - I see that there is a recent version with support for more params. I just updated the firmware - no change. I though I might have messed up the yml originally. Maybe not.

willcode:
Bug in the soapy driver
```
#if AIRSPYHF_VER_MAJOR >= 1
#if AIRSPYHF_VER_MINOR >= 1
#if AIRSPYHF_VER_REVISION >= 2
...
```

willcode:
PR for version bug in SoapyAirspyHF: https://github.com/pothosware/SoapyAirspyHF/pull/25

PR #6667, opened 2023-05-15
Include cstdint in gr-fec's alist.h

chris-pcguy:
## Description

GCC 13 will error out otherwise.

gr-fec

## Checklist

### DIFF
--- a/gr-fec/include/gnuradio/fec/alist.h
+++ b/gr-fec/include/gnuradio/fec/alist.h
@@ -23,6 +23,7 @@
 #define ALIST_H
 
 #include <gnuradio/fec/api.h>
+#include <cstdint>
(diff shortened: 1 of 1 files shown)


marcusmueller (approved):
Fair enough – this uses a `uint8_t`, and should include that header.

marcusmueller:
@willcode  probably a candidate for backporting (potentially even to pre-3.10?); I don't think including that header has any expectable detrimental side-effects.

willcode:
Is this causing compile problems for someone? As far as pre-3.10, it would only help people who pull from maint-3.9, since there won't be any more 3.9.X releases.

marcusmueller:
True – older builds with modern GCC will be rare.

PR #6674, opened 2023-05-18
gr-analog: add cpp code generation for analog_quadrature_demod_cf

Akira25:
For an S-NET decoding project, we need to include a GNU-Radio Flowgraph into our application. Preferably all linked statically in C++ (see also email [here](https://lists.gnu.org/archive/html/discuss-gnuradio/2023-05/msg00070.html)).

This is an attempt to add cpp code generation support for a block that we need. As we are not sure, if this change is enough to support the feature we need, this is a draft, as suggested by @marcusmueller. Any comments on it are highly welcome and appreciated. 

## Description

This PR should add cpp-generation support for the Quadrature demodulation block.

gr-analog/analog_quadrature_demod_cf

## Testing Done

I have compile-tested the changes, but wasn't able to test them in gnu-radio itself. It would be great to have a way to test this, without messing up my existing systems installation. Tips on how to do this are very welcome.

## Checklist

### DIFF
--- a/gr-analog/grc/analog_quadrature_demod_cf.block.yml
+++ b/gr-analog/grc/analog_quadrature_demod_cf.block.yml
@@ -1,6 +1,6 @@
 id: analog_quadrature_demod_cf
 label: Quadrature Demod
-flags: [ python ]
+flags: [ python, cpp ]
 
 parameters:
 -   id: gain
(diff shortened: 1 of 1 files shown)


marcusmueller:
This looks pretty good to me! I'll go as far as saying that you can remove the "Draft" status and make this a proper PR.

willcode (approved):
Code looks good. We don't seem to have much C++ QA code.

willcode:
That test failure does not appear to be related. I re-started the test. We can merge this to main and let people try it out.

PR #6676, opened 2023-05-19
qtgui/QFosphorColorMapper: std::sstream+string instead of sscanf("%m")+char*

### LINKED ISSUE #6648: [MSVC] 'sscanf()' and "%m"
### Feature Description

Building **gnuradio-qtgui.dll** on MSVC (with OpenGL support), I noted this important warning:
```c
QFosphorColorMapper.cc(67): warning C4476: 'sscanf' : unknown type field character 'm' in format specifier
QFosphorColorMapper.cc(67): warning C4474: 'sscanf' : too many arguments passed for format string
QFosphorColorMapper.cc(67): note: placeholders and their parameters expect 1 variadic arguments, but 2 were provided
```

A support for `%m` seems Linux specific too me. Ref:  http://linux.die.net/man/3/scanf 

No word on this on [MSDN](https://learn.microsoft.com/en-us/cpp/c-runtime-library/scanf-type-field-characters?view=msvc-170). Not sure if MinGW-w64 supports this. 

For MSVC, that code will crash at runtime; `free()` of an illegal pointer.

I tried this feeble attempt instead:
```diff
--- a/gr-qtgui/lib/QFosphorColorMapper.cc 2023-04-26 15:19:11
+++ b/gr-qtgui/lib/QFosphorColorMapper.cc 2023-05-04 13:47:35
@@ -64,7 +64,13 @@

         /* Are we in a palette ? */
         if (kp < 0) {
+#ifdef _WIN32
+            char buf[41] = { '\0' };
+            name = buf;
+            sscanf(line.constData(), "%d %40s", &kp, buf);
+#else
             sscanf(line.constData(), "%d %ms", &kp, &name);
+#endif

             if (kp)
                 gradient = new QLinearGradient();
@@ -75,8 +81,10 @@

             addPalette(name, pixmap);

+#ifdef _WIN32
             /* Release name & prepare for next */
             free(name);
+#endif
             name = NULL;

             kp = -1;
@@ -96,7 +104,9 @@
                 delete gradient;
                 gradient = NULL;

+#ifdef _WIN32
                 free(name);
+#endif
                 name = NULL;

                 kp = -1;
```

Must be a better way.

### Feature Urgency

high (impacts broad portion of ecosystem)

### More Information

_No response_

marcusmueller:
Fixes #6648

(hopefully)

## Description

#6648 highlights that we introduced GCC extension of the C sscanf to the codebase. That `%ms` is handy when dealing with unknown length string data in C: it avoids having to scan for string length first, then `malloc` a buffer, then do the actual `scanf`. But: That problem is a C-exclusive problem. C++ has `std::string`, and `std::stringstream` (though the latter *could* be nicer).

So, we replace the usage of sscanf and raw C strings with `std::string` altogether and hopefully arrive at easier to read code, as well.

## Related Issue

- #6648, which this is supposed to fix
- #6515, which introduced the code

- qtgui-internal (specifically, Fosphor-internal)

## Testing Done

none, due to hot laptop on hot train

## Checklist

### DIFF
--- a/gr-qtgui/lib/QFosphorColorMapper.cc
+++ b/gr-qtgui/lib/QFosphorColorMapper.cc
@@ -9,6 +9,8 @@
 
 #include "QFosphorColorMapper.h"
 #include <QGLContext>
+#include <memory>
+#include <sstream>
 
 namespace gr {
 namespace qtgui {
@@ -45,8 +47,8 @@ QFosphorColorMapper::QFosphorColorMapper(QObject* parent) : QObject(parent)
 
 int QFosphorColorMapper::loadFromFile(QFile& file)
 {
-    QLinearGradient* gradient = NULL;
-    char* name;
+    std::unique_ptr<QLinearGradient> gradient;
+    std::string name;
     int kp = -1;
 
     /* Make sure it's open */
@@ -54,6 +56,7 @@ int QFosphorColorMapper::loadFromFile(QFile& file)
         file.open(QFile::ReadOnly);
 
     /* Scan until the end */
+    std::stringstream ss;
     while (!file.atEnd()) {
         /* Grab a line */
         QByteArray line = file.readLine().simplified();
@@ -64,10 +67,14 @@ int QFosphorColorMapper::loadFromFile(QFile& file)
 
         /* Are we in a palette ? */
         if (kp < 0) {
-            sscanf(line.constData(), "%d %ms", &kp, &name);
+            ss.str(line.toStdString());
+            ss >> kp;
+            ss >> name;
+            ss.clear();
 
-            if (kp)
-                gradient = new QLinearGradient();
+            if (kp) {
+                gradient = std::make_unique<QLinearGradient>();
+            }
         } else if (kp == 0) {
             /* Create palette */
             QString filename = line;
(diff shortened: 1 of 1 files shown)


willcode (approved):
Code looks good. We need someone report that this works before merging.

marcusmueller:
I think @gmeyerho is the person I'd trust on that – he understands what the code actually does much better than I do! I just really did a "mechanical" modification, he's our expert :)

gmeyerho:
I tested this out, but I am not seeing the color heat map anymore, just the red and white lines for max and average hold. If I remove this commit, I can see the heat map again. I didn't have time to look into this more than building and testing to see why that is. I will try to find time to figure out which part is causing that tomorrow.

marcusmueller:
:sweat: didn't mean to break things! Thanks for having a look!

gr-qtgui/lib/QFosphorColorMapper.cc
```diff
@@ -64,10 +67,13 @@ int QFosphorColorMapper::loadFromFile(QFile& file)
 
         /* Are we in a palette ? */
         if (kp < 0) {
-            sscanf(line.constData(), "%d %ms", &kp, &name);
+            ss.str() = line.toStdString();
```
gmeyerho, line 70: I found this needs to be ss.str(line.toStdString()); instead, it looks like ss.str() just returns a copy of the current contents rather than a reference that can be used to set the contents of the stringstream.

gr-qtgui/lib/QFosphorColorMapper.cc
```diff
@@ -64,10 +67,13 @@ int QFosphorColorMapper::loadFromFile(QFile& file)
...
+            ss >> kp;
+            ss >> name;
```
gmeyerho, line 72: Something about either of these, probably the name one since it's last, is causing an error. I didn't dig into exactly the error case, but making the other change and inserting a ss.clear(); after this line causes the color heat-map to appear like before

marcusmueller: Then let's do that!

gr-qtgui/lib/QFosphorColorMapper.cc
```diff
@@ -76,29 +83,24 @@ int QFosphorColorMapper::loadFromFile(QFile& file)
...
+            name.clear();
 
             kp = -1;
         } else {
-            float f[4];
+            std::array<float, 4> f;
 
             /* Read data point */
-            sscanf(line.constData(), "%f %f %f %f", f, f + 1, f + 2, f + 3);
+            ss.str() = line.toStdString();
```
gmeyerho, line 93: There was one more case of this needing to be ss.str(line.toStdString()); down here, sorry didn't call it out in the above comment. Though the color seems to work without this change, but should probably still make it

gmeyerho (approved):

gr-qtgui/lib/QFosphorColorMapper.cc
marcusmueller: I should have fixed that by myself, you shouldn't have to call out individual instances of systematic problems … sorry.

PR #6677, opened 2023-05-19
ci: add fedora 38

willcode:


willcode:
Blocked on error - `stop()` was redefined in uhd rfnoc replay, flagged by newer compiler.

See #6617

PR #6680, opened 2023-05-20
filter: designer rework and fixes (backport to maint-3.10)

willcode:
Backport #6621 #6629 #6639 #6640 

There is a change in exceptions `pm_remez()`:

> throws std::invalid_argument if args are invalid and std::runtime_error if calculation fails to converge

where previously only `std::runtime_error` was thrown.

Given that so many checks were missing and the risk of bad results, it seems like a good idea to backport this behavior change.


willcode:
@marcusmueller @argilo Backporting with the exception change.

PR #6688, opened 2023-05-25
gr-uhd: fix infinite recursion in the case a device is poorly connected

theartful:
When using a usrp on a shaky ethernet connection, the device might repeatedly report overflow errors which caused an infinite recursion in the work function. This is problematic since this will prevent the gnuradio flowgraph from being stopped since this block will always be busy.

I added a variable for number of retries until quitting, but it has a hard coded value of 10. I didn't want to complicate the interface and make this value changeable from the outside since it doesn't seem likely that tuning this value could cause any benefits.

I also took the liberty to change the recursive mutex to a normal mutex. The mutex is used in only three functions:
1. work
2. start
3. stop

And none of them call each other anymore, so this should be safe.

## Checklist


willcode (commented):
Agree recursion should not be used as a way of handling errors except in special cases. Since `WORK_DONE` is never returned by `work()`, the `-1` might be ok. Typically that's a reserved number and could cause confusion.

mbr0wn (approved):
Thanks! I like these changes.

PR #6690, opened 2023-05-28
CI runner improvements

marcusmueller:
## Description

I'm experimenting with speeding up CI; I'd like to use github `cache` action to archive and recall build artifacts, but it's not clear whether this is generally feasible. While doing so there's some other CI builder improvements to be done; let's see how all this works out.

- Let every worker upload a build_commands.json artifact.
  - This can be used to understand failures as well as to allow language servers (LSP servers for neovim, emacs, vscode) to run with the right options on a machine that didn't do the build
- Change installation target from /usr/local to /tmp/prefix to check relocatability
- roll a tarball of installation
  - this can be used on the same distro as a read-to-run prefix

## Related Issue

n/a

CI, Artifacts

## Testing Done

Testsuite continues to pass

## Checklist

### DIFF
--- a/.github/workflows/make-test.yml
+++ b/.github/workflows/make-test.yml
@@ -14,7 +14,7 @@ jobs:
     runs-on: ubuntu-latest
     timeout-minutes: 15
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v3
     - uses: DoozyX/clang-format-lint-action@v0.14
       with:
         source: '.'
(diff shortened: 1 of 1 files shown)


willcode:
Did ccache improve performance at all?

marcusmueller:
Not yet, but the idea is to use https://github.com/actions/cache to do the following:

1. Load (if exists) .ccache directory for individual builder using `actions/cache/restore@v3`
2. Run build
    - CI workflow was caused by push to `main` or `maint-xx`, make CMake use ccache in RW mode, else
    - if cache existed make CMake use ccache in RO mode, else
    - run cmake without ccache
3. If CI workflow was caused by push to `main`/`maint-xx`
    1. run `ccache -X 12` to compress cache (faster reading, less cache evictions)
    2. update builder cache using `actions/cache/save@v3`

PR #6694, opened 2023-05-31
runtime: thread_group use set instead of list for set of threads

marcusmueller:
## Description

Currently, `thread_group` is using `std::list` to keep a set of thread unique ptrs, which is both inefficient computationally and complicated code-wise.

## Related Issue

Fixes a missed `#include <mutex>`, which was hidden by precompiled headers (or so it seems!) on the way.

Internal to `thread_group`; API doesn't change

## Checklist

### DIFF
--- a/gnuradio-runtime/include/gnuradio/thread/thread_group.h
+++ b/gnuradio-runtime/include/gnuradio/thread/thread_group.h
@@ -3,6 +3,7 @@
  * Copyright (C) 2001-2003 William E. Kempf
  * Copyright (C) 2007 Anthony Williams
  * Copyright 2008,2009 Free Software Foundation, Inc.
+ * Copyright 2023 Marcus Müller
  *
  *  Distributed under the Boost Software License, Version 1.0. (See accompanying
  *  file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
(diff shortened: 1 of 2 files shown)


willcode:
Backport note: I'm assuming that this is only used internally, and that if anyone did use it externally, they weren't checking for exceptions.

willcode:
Curious - does a scoped_lock take effect in the middle of a block by spec or just by convention? I was trying to think of reasons why the compiler can't reorder code. Is it because the mutex param is not const? I've always created another `{...}` scope to show where the lock is expected to be in effect, but never figured out whether that was necessary.

marcusmueller:
> does a scoped_lock take effect in the middle of a block by spec or just by convention?

The effect being that a mutex is modified should happen exactly where the `scoped_lock` is constructed. Of course, C++ (or, rather, a C++ compiler) is allowed to reorder things, as long as they don't break the outcome (but usually, modifying a mutex will act as a barrier – still, if something is a local, side-effect free thing, than it could probably pass "through" the position of mutex modification).

The reason I chose to put the mutex smack in the middle of `::create_thread` was that the actual critical section is relatively low-duration compared to the creation of a thread – so it made sense to keep the thread creation out of the locked region. Also, we save ourselves a lock acquisition/unlocking in case the thread creation fails, e.g., due to being out of memory. (this is really a minor solace.)

PR #6715, opened 2023-06-09
fosphor_formatter updates

willcode:
## Description
Rework the formatter into a single block (vs many blocks in a heir block) and simplify the code.

WIP. Still need to work on:
- Average/IIR
- Scaling

Implementation of scaling is still unclear, since the fosphor display block can not currently be scaled.

QT GUI Fosphor Formatter

## Checklist

### DIFF
--- a/gr-qtgui/lib/fosphor_formatter_impl.cc
+++ b/gr-qtgui/lib/fosphor_formatter_impl.cc
@@ -1,6 +1,7 @@
 /* -*- c++ -*- */
 /*
  * Copyright 2023 Ettus Research, A National Instruments Brand
+ * Copyright 2023 Jeff Long
  *
  * This file is part of GNU Radio
  *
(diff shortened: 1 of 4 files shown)


willcode:
@mbr0wn @gmeyerho This is a WIP reworking the Fosphor Formatter. The Fosphor block scale is fixed at -100dB to 0dB. Are you planning to make that adjustable?

gr-qtgui/lib/fosphor_formatter_impl.cc
```diff
@@ -12,205 +13,223 @@
...
-        }
-        out[i] = r;
-    }
+    // Output is d_num_bins histogram levels, plus average and max rows. Ensure
+    // there is space, in case output is generated.
+    set_min_output_buffer(2 * (d_num_bins + 2));
+
+    // TODO: make sure decims are >0
+
+    d_fft = new gr::fft::fft_complex_fwd(fft_size);
```
mbr0wn, line 67: Shouldn't we put this into a unique_ptr?

gr-qtgui/lib/fosphor_formatter_impl.cc
```diff
@@ -12,205 +13,223 @@
...
-        // Reset hit counter
-        std::fill(d_hit_count.begin(), d_hit_count.end(), 0);
-        return num_items_written;
-    }
+                const int16_t hc = d_hit_count[i];
+                float hv = d_histo_buf_f[i];
 
+                // Subtract from previous histogram values proportional to d_tdecay,
+                // and add in current hit count proportional to d_trise. Take
+                // decimation into account. (This code could be just a bit clearer).
```
mbr0wn, line 172: To be honest, I basically copy&pasted this from gr-fosphor, which is probably optimized for OpenCL.

gr-qtgui/lib/fosphor_formatter_impl.cc
```diff
@@ -12,205 +13,223 @@
...
+        volk_32f_x2_max_32f(
+            d_maxhold_buf.data(), d_maxhold_buf.data(), d_logfft.data(), d_fft_size);
+
+        ++d_histo_count;
+        if (d_histo_count == d_histo_decim) {
+
+            ++n_histo;
+            d_histo_count = 0;
+
+            volk_32f_s32f_x2_convert_8u(
```
mbr0wn, line 154: Is this kernel new? I remember looking for something like this and blanking.

mbr0wn: In fact, `volk` `master` does not have this kernel.

gr-qtgui/lib/fosphor_formatter_impl.cc
mbr0wn: BTW, I think this function (updating a pixels luminescence, or whatever the right word is) could be pulled into its separate function, and then we could have fosphoric eye diagrams, constellation plots, etc.

gr-qtgui/lib/fosphor_formatter_impl.cc
willcode: VOLK uses `main`

gr-qtgui/lib/fosphor_formatter_impl.cc
willcode: Yes!

gr-qtgui/lib/fosphor_formatter_impl.cc
willcode: This is new since v3.0.0, so I guess I can't use it here and will have to use `volk_32f_s32f_convert_8i` or something.

PR #6718, opened 2023-06-11
Qt GUI: Added new Matrix Sink

R-ohit-B-isht:
## Description

This widget will allow for the visualization of radar Doppler/range plane or 2D equalizer taps, making it useful for radar imaging and equalizer analysis.

QT GUI Matrix Sink

## Testing Done

Yes

## Checklist


marcusmueller:
Hi Rohit!

Thanks for opening this PR!
As you can see, a few of the checks from the CI have failed. Click on "Details" next to "Make Test / {Distro} {Version}" to see what's happening.

There's two different failures:

1. When you look forall pipelines where your code builds (that being Debian, Ubuntu, fedora 37), you see that the qa_qtgui fails with a segmentation fault – that usually means you're accessing memory outside of the size of some container, or are accessing a piece of memory belonging to an object that never existed. Sounds like a bug! <br>The interesting first part is why it works on your computer and fails on *all* the machines of the CI! Figuring that out would probably lead you very far to finding a solution.
2. The Fedora 38 builder fails to build your code:

```text
In file included from /__w/gnuradio/gnuradio/gr-qtgui/lib/matrix_display.cc:10:
/__w/gnuradio/gnuradio/gr-qtgui/lib/matrix_display.h:105:10: error: 'mousePressEvent' overrides a member function but is not marked 'override' [-Werror,-Winconsistent-missing-override]
    void mousePressEvent(QMouseEvent* e);
         ^
/usr/include/qt5/QtWidgets/qwidget.h:620:18: note: overridden virtual function is here
    virtual void mousePressEvent(QMouseEvent *event);
                 ^
[ 61%] Building CXX object gr-blocks/python/blocks/bindings/CMakeFiles/blocks_python.dir/tagged_stream_multiply_length_python.cc.o
1 error generated.
make[2]: *** [gr-qtgui/lib/CMakeFiles/gnuradio-qtgui.dir/build.make:1011: gr-qtgui/lib/CMakeFiles/gnuradio-qtgui.dir/matrix_display.cc.o] Error 1
[ 61%] Building CXX object gr-qtgui/lib/CMakeFiles/gnuradio-qtgui.dir/matrix_sink_impl.cc.o
In file included from /__w/gnuradio/gnuradio/gr-qtgui/lib/matrix_sink_impl.cc:10:
In file included from /__w/gnuradio/gnuradio/gr-qtgui/lib/matrix_sink_impl.h:13:
/__w/gnuradio/gnuradio/gr-qtgui/lib/matrix_display.h:105:10: error: 'mousePressEvent' overrides a member function but is not marked 'override' [-Werror,-Winconsistent-missing-override]
    void mousePressEvent(QMouseEvent* e);
         ^
/usr/include/qt5/QtWidgets/qwidget.h:620:18: note: overridden virtual function is here
    virtual void mousePressEvent(QMouseEvent *event);
                 ^
In file included from /__w/gnuradio/gnuradio/gr-qtgui/lib/matrix_sink_impl.cc:10:
/__w/gnuradio/gnuradio/gr-qtgui/lib/matrix_sink_impl.h:59:9: error: 'work' overrides a member function but is not marked 'override' [-Werror,-Winconsistent-missing-override]
    int work(int noutput_items,
        ^
/__w/gnuradio/gnuradio/gnuradio-runtime/lib/../include/gnuradio/sync_block.h:40:17: note: overridden virtual function is here
    virtual int work(int noutput_items,
                ^
2 errors generated.
make[2]: *** [gr-qtgui/lib/CMakeFiles/gnuradio-qtgui.dir/build.make:1027: gr-qtgui/lib/CMakeFiles/gnuradio-qtgui.dir/matrix_sink_impl.cc.o] Error 1
make[2]: Target 'gr-qtgui/lib/CMakeFiles/gnuradio-qtgui.dir/build' not remade because of errors.
make[1]: *** [CMakeFiles/Makefile2:7225: gr-qtgui/lib/CMakeFiles/gnuradio-qtgui.dir/all] Error 2
```
That one is easy to fix: add the word `override` to the end of your `void mousePressEvent(QMouseEvent* e)` declaration in your .h; same for your `work` declaration in your `_impl.h`. I'll leave a review on that below!

By the way, you don't have to close this PR at all – just force-push to the same branch, and it will update this PR.

gr-qtgui/lib/matrix_sink_impl.h
```diff
@@ -0,0 +1,67 @@
...
+    ~matrix_sink_impl();
+
+    void exec_() override;
+    QApplication* d_qApplication;
+    QWidget* qwidget() override;
+
+    // Where all the action really happens
+    int work(int noutput_items,
+             gr_vector_const_void_star& input_items,
+             gr_vector_void_star& output_items);
```
marcusmueller, line 61: add `override` before `;`

gr-qtgui/lib/matrix_display.h
```diff
@@ -0,0 +1,112 @@
...
+    void set_x_axis_label(const std::string& x_axis_label);
+    void set_y_axis_label(const std::string& y_axis_label);
+    void set_z_axis_label(const std::string& z_axis_label);
+    void initialize_mouse_actions();
+    bool isClosed() const;
+
+public slots:
+    void set_data(QVector<double> data);
+    void save_figure();
+    void mousePressEvent(QMouseEvent* e);
```
marcusmueller, line 105: add `override` before `;`

marcusmueller (changes requested):

gr-qtgui/python/qtgui/qa_qtgui.py
```diff
@@ -78,6 +78,11 @@ def test13(self):
     def test14(self):
         self.qtsnk = qtgui.eye_sink_c(1024, 1, 1, None)
 
+    def test15(self):
+        self.qtsnk = qtgui.matrix_sink("Dopler", 2, 4, False,
```
marcusmueller, line 82: Doppler is written with two "p", because the guy who discovered / described the Doppler effect was called Doppler.  :grin:

gr-qtgui/lib/matrix_display.cc
```diff
@@ -0,0 +1,314 @@
...
+    d_plot->setAxisScale(QwtPlot::yRight, d_z_min, d_z_max);
+    d_plot->enableAxis(QwtPlot::yRight, true);
+}
+
+
+void matrix_display::save_figure()
+{
+    QPixmap qpix = this->grab();
+
+    QString types = QString(tr("JPEG file (*.jpg);;Portable Network Graphics file "
```
marcusmueller, line 215: Would it be possible to have PNG be the first choice here? JPEG is lossy and makes ugly artifacts at sharp edges, and isn't true to colors – all properties of an image format that we don't want for the pictures that come out of your code!

gr-qtgui/lib/matrix_display.cc
```diff
@@ -0,0 +1,314 @@
...
+        filename = filebox->selectedFiles()[0];
+        filetype = filebox->selectedNameFilter();
+    } else {
+        return;
+    }
+
+    if (filetype.contains(".jpg")) {
+        qpix.save(filename + ".jpg", "JPEG");
+    } else if (filetype.contains(".png")) {
+        qpix.save(filename + ".png", "PNG");
```
marcusmueller, line 233: Does that mean that if I entered "mypicture.png" as file name, I get "mypicture.png.png" as saved file?

gr-qtgui/python/qtgui/qa_qtgui.py
R-ohit-B-isht: oops , I'll correct that

gr-qtgui/lib/matrix_display.h
R-ohit-B-isht: Okay

gr-qtgui/lib/matrix_display.cc
R-ohit-B-isht: Sure

gr-qtgui/lib/matrix_display.cc
R-ohit-B-isht: I also addressed the same in this issue #6474

Saving it without adding .jpg or other extension made it invisible to the filter and did not showed the earlier saved files,i.e. if a saved one as "abc" (actually a jpeg) and then took another shot the file save dialog did not display any previous saved jpeg files on selecting the filter to jpeg files (which could be inconvenient if I accidentally replaced any previous screenshot with same name )
I checked the saved file in terminal it showed name as abc.jpg not abc.jpg.jpg , I think JPEG is only set as property without extension. Although I checked it in linux only don't know if it shows .jpg.jpg in any other system

marcusmueller: I'm not sure what you mean with "showed"; what is shown in a graphical file list might or might not include file extensions, I guess?

gr-qtgui/lib/matrix_sink_impl.h
R-ohit-B-isht: ok

gr-qtgui/include/gnuradio/qtgui/matrix_sink.h
```diff
@@ -0,0 +1,87 @@
...
+     * \param y_start The starting value of the y axis.
+     * \param y_end The ending value of the y axis.
+     * \param z_max The maximum value of the z axis.
+     * \param z_min The minimum value of the z axis.
+     * \param x_axis_label The label for the x axis.
+     * \param y_axis_label The label for the y axis.
+     * \param z_axis_label The label for the z axis.
+     * \param parent The parent QWidget.
+     */
+    static sptr make(const std::string& name,
```
willcode, line 63: Other widgets also hava a lot of contstructor params, but this is way too many. Let's have defaults for everything that has a setter, then call the setter for each param or set of params.

R-ohit-B-isht: got it

willcode (approved):
Tried this out, works, seems easy enough to use. One minor thing I noticed is that "None" is not an option for interpolation on the on-screen menu, even though it shows up in the block params. "Nearest neighbor" seems to be equivalent to "None". Is there meant to be a difference? If not, remove "None" from the params.

willcode:
Even though `matrix_display` is unlikely to be used outside `matrix_sink`, it would be nice style-wise if the constructor was minimized in the same way. Maybe put that on a list of things to do later if you have other work now.

It would be nice if the example showed something other than noise. Also not a priority, and I don't have a great idea for what to show.

R-ohit-B-isht:
> Tried this out, works, seems easy enough to use. One minor thing I noticed is that "None" is not an option for interpolation on the on-screen menu, even though it shows up in the block params. "Nearest neighbor" seems to be equivalent to "None". Is there meant to be a difference? If not, remove "None" from the params.

Right  , removed "None" .

R-ohit-B-isht:
> Even though `matrix_display` is unlikely to be used outside `matrix_sink`, it would be nice style-wise if the constructor was minimized in the same way. Maybe put that on a list of things to do later if you have other work now.
> 
> It would be nice if the example showed something other than noise. Also not a priority, and I don't have a great idea for what to show.

I'll add it in the cart 👍🏻

R-ohit-B-isht:
> > Tried this out, works, seems easy enough to use. One minor thing I noticed is that "None" is not an option for interpolation on the on-screen menu, even though it shows up in the block params. "Nearest neighbor" seems to be equivalent to "None". Is there meant to be a difference? If not, remove "None" from the params.
> 
> Right , removed "None" .

removing "None" seems to break the windows build 🤔

noc0lour:
I re-scheduled the test on your failed commit and it seems like the Windows test might be a fluke. Try removing the 'none' parameter again from this block description.

noc0lour:
@marcusmueller Since you already did a first review, can you check if this PR is now still missing something?

gr-qtgui/lib/matrix_display.cc
```diff
@@ -0,0 +1,348 @@
...
+
+{
+    d_isclosed = false;
+    d_hLayout = nullptr;
+
+    d_menu = nullptr;
+    d_save_act = nullptr;
+    d_stop_act = nullptr;
+    d_contour_menu = nullptr;
+    d_interpolation_menu = nullptr;
```
marcusmueller, line 49: would put these constant initializations into the class declaration, i.e. in the .h file

gr-qtgui/lib/matrix_display_signal.h
```diff
@@ -0,0 +1,28 @@
...
+
+public:
+    matrix_display_signal() {}
+    ~matrix_display_signal() {}
+
+signals:
+    void data_ready(QVector<double> data);
+};
+
+#endif /* MATRIX_DISPLAY_SIGNAL_H */
```
marcusmueller, line 28: at the end of a header file, please always have a line break

marcusmueller (changes requested):

willcode:
@R-ohit-B-isht since you're going to be doing some refactoring work, let's keep PRs for these new blocks open until the end of your project. I have a feeling you'll be making changes to these blocks once you dig into all the others.

gr-qtgui/lib/matrix_sink_impl.cc
```diff
@@ -0,0 +1,146 @@
...
+                           gr_vector_const_void_star& input_items,
+                           gr_vector_void_star& output_items)
+{
+    auto in = static_cast<const input_type*>(input_items[0]);
+
+
+    QVector<double> qvec;
+    for (unsigned int i = 0; i < d_vlen; i++) {
+        qvec.append(in[i]);
+    }
```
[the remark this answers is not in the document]
R-ohit-B-isht: i tried that but since the incoming data is float so it wouldn't take that , also when casting the "in" from float to double the plot wasn't accurate so had to stick to that.

willcode:
@R-ohit-B-isht It turns out this is the PR that has the Qt version problem with the QVector constructor. Since v3.10 needs to run on Ubuntu 20.04, we need to either fix this or make this conditional on a new enough version of Qt. Ideas? I didn't catch that before merging (GRCon fog). If we can fix this soon, the PR can stay in, otherwise I'll need to back it out because it's tripping up CI.

willcode:
@R-ohit-B-isht Would `QVector::fromStdVector` work for this case, since iteration is from begin to end?
http://man.hubwiz.com/docset/Qt_5.docset/Contents/Resources/Documents/doc.qt.io/qt-5/qvector.html#fromStdVector

R-ohit-B-isht:
> Would `QVector::fromStdVector` work for this case, since iteration is from begin to end

I think it should work , I'll test it quickly

willcode:
Great! You can submit that as a new PR. I put in one as a test, but you can ignore that.

willcode:
The cast from `std::vector<float>` to `QVector<double>` in the initializer didn't work in my quick test ... see what you can come up with.

R-ohit-B-isht:
> QVector<double> qvec;
>     for (unsigned int i = 0; i < d_vlen; i++) {
>         qvec.append(in[i]);
>     }

yeah  i also tried that , so instead of casting and juggling between 2 arrays i was inserting it with a loop earlier which was also working on all

`
 QVector<double> qvec;
    for (unsigned int i = 0; i < d_vlen; i++) {
        qvec.append(in[i]);
    }
`

willcode:
Use reserve() first to avoid realloc, and there should be a std::copy that can replace the explicit loop

willcode:
@R-ohit-B-isht are you going to be able to finish this up?

willcode:
Backport note: need to add applicable chunks of #7065

marcusmueller:
Can't be backported, since older versions of Qt5 do not support instantiating a QVector from a std::vector

marcusmueller:
Backport squash with https://github.com/gnuradio/gnuradio/pull/7813

PR #6722, opened 2023-06-16
grc: show better contrast in dark mode

kdriedger:
Signed-off-by: Kevin Driedger <linuxbox@gmail.com>

## Checklist

### DIFF
--- a/grc/gui/canvas/colors.py
+++ b/grc/gui/canvas/colors.py
@@ -72,25 +72,26 @@ PORT_TYPE_TO_COLOR.update((key, get_color(color))
 #################################################################################
 # param box colors
 #################################################################################
+
 DARK_THEME_STYLES = b"""
-                         #dtype_complex         { background-color: #3399FF; }
-                         #dtype_real            { background-color: #FF8C69; }
-                         #dtype_float           { background-color: #FF8C69; }
-                         #dtype_int             { background-color: #00FF99; }
+                         #dtype_complex         { background-color: #CC6600; }
+                         #dtype_real            { background-color: #007396; }
+                         #dtype_float           { background-color: #007396; }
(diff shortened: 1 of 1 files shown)


kdriedger:
Before Screenshot:
before-patch-Screenshot

After Screenshot:
after-patch-Screenshot

willcode:
Since there haven't been any other comments, I'm going to say that the colors arrived at by subtracting are not very good. These colors need to be manually selected, or derived from some theme.

kdriedger:
Does this PR make the codebase better?  

Does it need to be perfect before being accepted?

marcusmueller:
To me, the after is actually worse to read!

kdriedger:
> To me, the after is actually worse to read!

You can't be serious (maybe that's what the smiley face means?).  Especially the ID field.  It's completely unreadable.

willcode:
@kdriedger would you manually select some nice colors? Yes, the hot pink is more readable than the mint green. A one-time change in colors would be good.

There are 15 colors changed in the PR. Do the rest of them look good?

marcusmueller:
> > To me, the after is actually worse to read!
> 
> You can't be serious. Especially the ID field. It's completely unreadable.

I wish I wasn't, but while I agree that ID is an unquestionable improvement (so much that I didn't even realize I couldn't read it in the before state), Title and Run are now worse.

marcusmueller:
In all honesty, can we just break with the system theme, make the text black, and replace these high-intensity colors with less-saturated, closer-to-white options? I mean, for people having problems discerning colors, anything but nearly-black-on-nearly-white (or the inverse) is hard to read, and for the rest "pinkish-white" is as good a type hint as "garish-pink", right?

marcusmueller:
Color handling now different.

PR #6724, opened 2023-06-17
soapy: fix sdrplay biasT_ctrl/agc/gain settings

tko:
## Description
When using Soapy SDRPLay Source alone without any variables using RSP1 (clone) there are errors:
* `biasT_ctrl` setting is unknown
* `self.agc` does not exist
* `self.gain` does not exist

## Related Issue
Original fixes https://github.com/gnuradio/gnuradio/pull/6665 seem to require extra variables / parameters in the flowgraph.

Soapy SDRPlay Source

## Testing Done
Using gnuradio 3.10.6
1. Run a top block with nothing but Soapy SDRPlay Source and Null Sink
2. Create a custom hier block with nothing but Soapy SDRPlay Source and a Pad Sink
3. Run a top block with nothing but the previous hier block and Null Sink

## Checklist

### DIFF
--- a/gr-soapy/grc/soapy_sdrplay_source.block.yml
+++ b/gr-soapy/grc/soapy_sdrplay_source.block.yml
@@ -38,7 +38,6 @@ parameters:
   - id: bias
     label: Bias Tee Power
     dtype: bool
-    default: 'False'
     hide: part
 
   - id: center_freq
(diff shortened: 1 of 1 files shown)


willcode:
Thanks! You're right, I probably had same-name variables in the flowgraph already and didn't notice they were being used.

willcode:
I think we'll make some changes (later) to the settings logic, to get rid of errors when a setting does not exist. There's no great answer, but I think being too strict causes more errors, and forces client-side logic like this. A little client-side logic is ok, none would be even better.

Issue #5455, opened 2022-01-09
crosscompiling GNU Radio 3.9 (and 3.10?) uses host naming of library rather than target


jmfriedt:
### What happened?

Cross-compiling GNU Radio 3.9 to ARM aarch64 architecture using Buildroot and the
https://github.com/oscimp/oscimp_br2_external/blob/master/configs/raspberrypi4_64_gnuradio39_defconfig
configuration script leads to libraries (bindings?) named x86_64 which is the host architecture, and prevents GNU Radio from loading modules (e.g. ``import gnuradio`` in Python3 on the Raspberry Pi 4 target succeeds but ``from gnuradio import audio`` fails due to missing binding). Renaming all .so files including x86_64 (host architecture) to aarch64 (target architecture) ended up solving the problem. This sound very similar to https://github.com/pybind/pybind11/issues/1330 concluding with "now the pybind takes the correct python directory but still the architecture of the generated .so file is x86_64" but no solution is provided. This issue is also valid for OOT module, e.g. when cross-compiling https://github.com/jmfriedt/gr-rpitx the resulting .so file must be renamed from the host to target architecture.

### System Information

Buildroot 2021.11 with the configuration file mentioned in the bug report

### GNU Radio Version

3.9 (maint-3.9)

### Specific Version

0.0.0.0

### Steps to Reproduce the Problem

Select GNU Radio 3.9 in the Buildroot configuration. After make and running the dd-ed on the RPi4, running in Python3
from gnuradio import audio
will fail due to misnamed library (binding) .so. Renaming (mv) or linking (ln -s) to the same library name but replacing x86_64 with aarch64 solves the issue.

drmpeg:
This was discussed on Matrix/IRC, but I'm going to document it here. The ABI string comes from pybind11. Here are the details.

A special function ```pybind11_add_module()``` in ```gnuradio/cmake/Modules/GrPybind.cmake``` is used to create the Python binding libraries.

https://github.com/gnuradio/gnuradio/blob/master/cmake/Modules/GrPybind.cmake#L9

This function is in ```/usr/lib/cmake/pybind11/pybind11Tools.cmake```. Within that function, a call to ```set_target_properties() is made```.

```set_target_properties(${target_name} PROPERTIES SUFFIX "${PYTHON_MODULE_EXTENSION}")```

[...]

PR #6727, opened 2023-06-20
grPyBind: fix invalid module suffix for cross-compile

trabucayre:
This is required as pybind11 uses host python3 and therefore the host platform suffixes.

## Description

By default, module libraries have a suffix based on cpython version + host architecture: this is fine for a native compile when these libraries are used on the same computer (or similar computers). But when target architecture is not the same python is unable to find libraries due to the wrong suffix and produces unclear errors messages.
This PR, when `CMAKE_CROSSCOMPILING` is set, force libraries suffix extension to resolve this runtime issue.

## Related Issue

Fixe #5455

## Testing Done

Tested with 3 gnuradio 3.10 tags, and *main* in a cross-compilation environment

## Checklist

- [ x] I have read the [CONTRIBUTING document](https://github.com/gnuradio/gnuradio/blob/main/CONTRIBUTING.md).
- [ x] I have squashed my commits to have one significant change per commit. 
- [ x] I [have signed my commits before making this PR](https://github.com/gnuradio/gnuradio/blob/main/CONTRIBUTING.md#dco-signed)
- [ x] My code follows the code style of this project. See [GREP1.md](https://github.com/gnuradio/greps/blob/main/grep-0001-coding-guidelines.md).

### DIFF
--- a/cmake/Modules/GrPybind.cmake
+++ b/cmake/Modules/GrPybind.cmake
@@ -8,6 +8,12 @@ macro(GR_PYBIND_MAKE name updir filter files)
                    ${CMAKE_CURRENT_BINARY_DIR} COPYONLY)
 
     pybind11_add_module(${name}_python ${files})
+    
+    # Use normal .so suffix when crosscompiling
+    # See https://github.com/gnuradio/gnuradio/issues/5455
+    if (CMAKE_CROSSCOMPILING)
+        set_target_properties(${name}_python PROPERTIES SUFFIX ".so")
+    endif()
 
     set(MODULE_NAME ${name})
     if(${name} STREQUAL gr)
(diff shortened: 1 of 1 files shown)


willcode:
Any reason we wouldn't do the same thing for native?

balister:
This sounds like a bug in pybind11 you are using for cross work. Let me look into what OpenEmbedded does. Using the host python is a terrible idea as it may not match the version in the cross environment....

balister:
Also what version of pybind11 do you use?

mormj:
Minimum of 2.4 but it varies from system to system because we don't vendorize it

balister:
Pybind11 current version is 2.10.4, it is possible the issue has been fixed in recent pybind11's. I'd like to know what version the reporter is using.

trabucayre:
Sorry for delay.
I use pybind11 2.10.3.
After a closer look to `pybind11NewTools.cmake` I have discovered `-DPYTHON_MODULE_EXTENSION=".so"`. This option avoid pybind11 to determine/forge suffix by itself.

balister:
> Sorry for delay. I use pybind11 2.10.3. After a closer look to `pybind11NewTools.cmake` I have discovered `-DPYTHON_MODULE_EXTENSION=".so"`. This option avoid pybind11 to determine/forge suffix by itself.

Does this mean we should cancel the pull request?

trabucayre:
I think so.
If it appears for a future pybind11 release this PR is requested I whill create a new one (or reopen this one).

PR #6738, opened 2023-06-30
soapy: fix variable naming in RTL grc

willcode:
## Description
Fix variable names, which just happened to work if a control with the same name already existed. This is the same type of fix Tommi made recently for SDRPlay.

## Checklist

### DIFF
--- a/gr-soapy/grc/soapy_rtlsdr_source.block.yml
+++ b/gr-soapy/grc/soapy_rtlsdr_source.block.yml
@@ -52,7 +52,6 @@ parameters:
   - id: bias
     label: 'Bias Tee Power'
     dtype: bool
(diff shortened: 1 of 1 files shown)


tko (commented):
Looks ok to me ("_value" always assigned before used, `get_gain_mode` returning 0 for manual)

On quick google the support for "biastee" in soapy depends on existence of "rtlsdr_set_bias_tee" function (compile time rather than runtime/hardware feature) -- [the check was added in 2020](https://github.com/pothosware/SoapyRTLSDR/commit/a444e903ba8681d7cbc2e4c4ff4c9a592c429f88) so it could be safe to assume the setting always exists.

I don't have RTL-SDR.

PR #6739, opened 2023-07-03
soapy: more robust handling of bias and agc settings in rtlsdr yml

willcode:
## Description
Check with soapy driver whether biastee is a valid setting, instead of relying on a black field as a default.

Protect bias and agc calls from blank fields. GRC allows a blank for bool. In the templates, this is an actual blank, not an empty string, so it can cause errors.

This is an improvement to #6738.

Soapy RTLSDR Source

## Checklist


willcode:
@tko Since the existance of "biastee" can be detected at runtime, I reworked this a little. Should have waited a few minutes before commiting the previous version! Could you do a quick sanity check again? Thanks.

PR #6748, opened 2023-07-12
fec: LDPC Decoder and Soft Decision Fixes

VladislavF:
This PR seeks to fix some critical bugs and oversights in the gr-fec LDPC code and the soft decision calculations.
Soft decision calculation had mathematical errors, and the LDPC decoder was numerically unstable with high LLR values. My decoder is not an improvement, but it is stable under conditions that caused the old one to fail.
At the request of @willcode I have reopened the previous PR: https://github.com/gnuradio/gnuradio/pull/6478 against main.
Please see the previous PR for more exhaustive details and discussion.

The main changes from the last PR were cleaning up the constellation class architecture to have the noise power argument be a member variable that is updated as needed. This code should be in much better condition now and I have verified that it still works as expected on this branch. I manually moved my changes since rebasing from maint-3.10 proved to be impossible with my git ability.

The following blocks have been modified:

-  Constellation Object definition
-  LDPC Decoder
- Constellation soft decoder

## Summary of Changes

- LDPC sum-product algorithm replaced with tanh rule
- Python constellation LUT generator expressions corrected
- LDPC parity check matrix generator patched for python3
- Constellation object has Noise power argument
- Constellation soft decoder has noise power argument
- Constellation object LUT and soft decision CPP fixed

## Known Issues and Next Steps

- The LDPC Decoder is a bit slower now because std::tanh and std::atanh are implemented inefficiently/expensively.
- The LDPC "tanh rule" requires clamping of 'x' in tanh(x) to improve BER performance. This also helps to avoid more NaN and Inf. issues and should be investigated since the 0.5dB performance gap also likely lies here.


willcode:
I hit "Run" manually on CI. Needs DCO and clang-format.

willcode:
For formatting, run `clang-format -i` on each modified C++ file. There's an in-tree format conf file that will control the formatting.

VladislavF:
> For formatting, run `clang-format -i` on each modified C++ file. There's an in-tree format conf file that will control the formatting. CI stopped on `gr-fec/lib/awgn_bp.cc`.

I used the VSCode clag-format originally, went back and followed your suggestion only to find exactly 1 character change in all the files. awgn_bp.cc had a single missing space on a comment...

VladislavF:
I saw this make test failure on my GitHub CI as well, I cannot explain it since I cannot reproduce the issue on my end. The constellation QA runs fine for me.

daniestevez:
Happy to see that this PR is closer to being merged!

I already reviewed all the algorithmic details in #6478. I had some doubts, but overall we agreed that the code looked good.

In the mean time, I have done my own implementation of a LDPC decoder in Rust (with many arithmetic variants to benchmark them). I was meaning to publish this for quite some time already, but I have rushed and [published it now](https://destevez.net/2023/07/ldpc-toolbox-gets-ldpc-decoding/), as it can be useful to test this PR.

I have wrapped these Rust LDPC encoder and decoder as GNU Radio blocks. I want to check that the BER performance of the decoder in this PR is reasonable by swapping out and comparing with my decoder (which in turn I have been comparing to performance curves in the [CCSDS Green Book](https://public.ccsds.org/Pubs/131x0b4.pdf) ). Hopefully I'll be able to do this over the weekend and report here.

After doing my own decoder implementation I have a better understanding of what can go wrong numerically. Probably I should revisit the comments in #6478 (we were speaking about the need to clip the argument to the arctangent, and similar stuff), since maybe I have new ideas for improvement. But I don't think any of those ideas should block this PR. Any improvement can be done later on.

daniestevez:
I'm comparing the LDPC decoder from this PR with the LDPC decoder from [gr-ldpc-toolbox](https://github.com/daniestevez/gr-ldpc-toolbox) and the Eb/N0 performance of this PR is not so good (maybe I'm doing something wrong that hurts performance).

I'm using the CCSDS r=1/2 k=1024 code and modifying the [gr-ldpc-toolbox example](https://github.com/daniestevez/gr-ldpc-toolbox/blob/main/examples/ldpc_simulation.grc) (the alist file is [included in gr-ldpc-toolbox](https://github.com/daniestevez/gr-ldpc-toolbox/blob/main/python/ldpc_toolbox/ar4ja_1_2_1024.alist)).

This LDPC code is actually a (2560, 1024) code in which the last 512 bits are punctured. Since the LDPC decoder in this PR doesn't have built-in support for puncturing, for simplicity I'm not doing puncturing and using this code as a (2560, 1024) code, so its Eb/N0 performance should be somewhat better than what indicated by the CCSDS Green Book, as we're using it as a r=2/5 code. (I could do puncturing by hand externally to the LDPC decoder if I wanted).

The GR LDPC implementation has its own "peculiar" idea of [which are the systematic bits](https://github.com/gnuradio/gnuradio/blob/main/gr-fec/lib/cldpc.cc#L42) in the codeword. I think that choice only makes sense for the kind of codes introduced in the book by Urbanke and Richardson (which allow low complexity systematic encoding by cleverly choosing which bits are systematic), but it limits many real world use cases (so it would be good to discuss what to do outside of this PR). In this case, the systematic bits are the first 1024 bits. I have modified this function accordingly to return the first 1024 bits:

```diff
diff --git a/gr-fec/lib/cldpc.cc b/gr-fec/lib/cldpc.cc
index 1dce774aa..9dfb9ea42 100644
--- a/gr-fec/lib/cldpc.cc
+++ b/gr-fec/lib/cldpc.cc
@@ -44,9 +44,13 @@ std::vector<uint8_t> cldpc::get_systematic_bits(std::vector<uint8_t> in)
     std::vector<uint8_t> data;
     data.resize(K);
     int index;
-    for (size_t i = 0; i < K; i++) {
+    /*    for (size_t i = 0; i < K; i++) {
         index = permute[i + rank_H];
         data[i] = in[index];
+        }*/
+    // hack: get the first K bits
+    for (size_t i = 0; i < K; i++) {
+        data[i] = in[i];
     }
     return data;
 }
```

The flowgraph I'm using is here: 
[ldpc_test.zip](https://github.com/gnuradio/gnuradio/files/12063696/ldpc_test.zip)


There is an Embedded Python block that generates 1024-bit frames which contain a 32-bit sequence number, a random payload, and a CRC-32. These frames are run through the LDPC encoder, AWGN channel, and LDPC decoder. The CRC of the output frames is checked, and correctly decoded frames are printed. By looking at how the sequence number grows, we can get a rough idea of the frame error rate.

The flowgraph has both the gr-ldpc-toolbox decoder and the decoder from this PR instantiated. By disabling one or the other it is possible to compare them.

With an Eb/N0 of 2 dB, the gr-ldpc-toolbox LDPC decoder seems to decode all the frames (or most of them). With the same Eb/N0, the decoder from this PR has a very high frame error rate (maybe it's decoding 50% of the frames only). At high Eb/N0 both decoders decode all the frames.

There is a slight difference in the data that is passed as input to each LDPC decoder. For the gr-ldpc-toolbox one, the LLRs are computed externally by multiplying the signal by `-2.0 / noise_sigma**2` (the signal is BPSK symbols of amplitude one). The minus sign is because the decoder uses the convention that a positive LLR means that the bit 0 is more likely.

In this PR, it seems to me that the decoder expects LLRs with the convention that a positive LLR means that the bit 1 is more likely. It might seem that the decoder does the conversion to LLRs internally, because it takes the noise sigma as a parameter, however, this parameter is not used internally (see [this function](https://github.com/gnuradio/gnuradio/blob/df9d4fc51a83af08774391ba8636fc68b0a49baa/gr-fec/lib/awgn_bp.cc#L82)). Therefore, to feed this decoder I'm multiplying the signal by `2.0 / noise_sigma**2`. I've also tried multiplying by `1` instead, obtaining similar results.

VladislavF:
okay, I have finally fixed a few issues with the QA testing. I have to say that the way the QA is setup for the constellation is heinously bad. It is comparing output between python and C++. And seeing as these languages treat floats very differently it's asking for and indeed causes massive trouble. I was able to make it work, but this should be revisited as it is not a good test. 

@daniestevez I have just taken a look at your flowgraph and the first thing I notice is that your conversion to LLR is just a multiplication. Maybe there's a shortcut I'm overlooking, but I'd imagine you actually need to take the log for the input to the LDPC decoder to be what it's expecting.

daniestevez:
> @daniestevez I have just taken a look at your flowgraph and the first thing I notice is that your conversion to LLR is just a multiplication. Maybe there's a shortcut I'm overlooking, but I'd imagine you actually need to take the log for the input to the LDPC decoder to be what it's expecting.

The conversion to LLR is supposed to be just a multiplication.

For instance, this is from the beginning of [Sarah Johnson's book](https://www.cambridge.org/core/books/iterative-error-correction/7964EAB1B743D56D2E9B58CCA4CE71EB):

llr1

[...]

llr2

VladislavF:
@daniestevez okay, in that case I would say it probably just is worse. I once again confirmed today that this decoder does perform about 0.25-0.5 dB worse than the original awgn_bp implementation depending on the exact tanh clamp value. I am not attached to this particular code, I would even agree that it's not very good.

However, the current LDPC code is unusable if you have LLR values with unit symbol power and expected SNR that's anywhere above 7dB or so. It is very easy to cause it to produce total garbage. So what my fix provides is increased numerical stability. I would love to make this better, but I unfortunately just do not have any more time to spend on this project. If you would like to roll your implementation into this PR I'm completely open to it. I am even open to just dropping the LDPC part completely. but between the unintuitive sigma value that basically recalculates what it's already been given and numerical instability it doesn't feel right to just leave it.

@willcode apologies for all the formatting errors, my python linter just does not work. It says everything is fine, even making some changes, and then the formatting CI still fails. Is there a command I can run on my end to make sure my python formatting is right? like the clang-format?

daniestevez:
> I once again confirmed today that this decoder does perform about 0.25-0.5 dB worse than the original awgn_bp implementation depending on the exact tanh clamp value.

Do you have an example flowgraph that you're using to test this? I'd like to see how you're passing LLRs to the decoder and to try to compare the performance with my decoder in the same case.

> If you would like to roll your implementation into this PR I'm completely open to it.

At the moment I don't have the bandwidth to rewrite my Rust LDPC decoder in C++, which would be needed for inclusion in this PR. And this doesn't seem the best idea. It would probably be better to understand exactly what are the differences and what might be causing problems. How the Tanner graph is iterated is unlikely to be wrong or cause problems, so there's no benefit in rewriting that. The problems are more likely in subtle things in the check node or variable node updates, or even outside of this block (in what the block expects regarding LLRs, or even in the FECAPI wrapper).

Which brings me to another reason for having an example that shows how the decoder can be used successfully at an Eb/N0 close to the threshold: without this, it will be quite easy to use it incorrectly.

Only now I understood that the `awgn_bp` constructor parameter `sigma` now has the semantics of being the tanh clamp value, which is probably something that the user doesn't have to tweak. For `float32` I already [had good results clamping at `9.0`](https://github.com/daniestevez/ldpc-toolbox/blob/main/src/decoder/arithmetic.rs#L434), and didn't feel the need to make this configurable, nor to clamp at smaller values. In any case, I wouldn't call the clamp value `sigma` in the code, since it's quite possible to mistake it for the noise standard deviation (which is what the old `sigma` was). I would also consider this an API breaking change.

I also realize that the flowgraph that I shared before wasn't the one I intended. This is the correct one. I've also updated the `sigma` parameter of the `awgn_bp` to 2.3, which is what you have in the GRC file.

[ldpc_simulation.zip](https://github.com/gnuradio/gnuradio/files/12134699/ldpc_simulation.zip)

My LDPC decoder does error-free decoding at 2 dB Eb/N0, which is expected. With your decoder I need around 9 to 10 dB Eb/N0 to get error-free decoding. This is a huge difference. I'm not talking about underperforming by 0.5 or 1 dB, which would probably be okay. Since this isn't showing up in your tests, I wonder what the difference between your tests and my tests is. Maybe it's the kind of LDPC codes you're using, or maybe it's something about how I'm computing the input LLR.

daniestevez:
> ldpc_simulation.zip

I'm now testing this flowgraph with the main branch instead of this PR (with my patch to make the first bits of the codeword systematic). I find that I need around 4.5 dB Eb/N0 for error-free decoding. This is even more confusing. The main branch is also underperforming noticeably, but not as much as this PR, and the differences between each of the 3 decoders I've tested are several dB for this particular simulation.

To use the main branch I bypassed the Multiply Const block that converts to LLR, since the LDPC Decoder from main does the conversion internally. Also, I ensured that the parameter `sigma` was correctly set to the noise sigma used in the simulation.

daniestevez:
I've also been reviewing how the code in this `awgn_bp` in this PR compares to my LDPC implementation (using what I calle the `Tanhf32` rule). The variable node update is identical, except for the order of operations (there are not many variations that can be applied to the variable node update).

For the check node update, to generate an outgoing message to a variable `V` this PR does:
1. For each variable node message `Q` except the one coming from `V`, compute `x = abs(Q) / 2`. 
2. Clamp `x` to a maximum of `2.3` (configurable value) to produce `x_clamped`.
3. Compute `tanh(x_clamped)`.
4. Compute the product of all such `x`'s. Call this `prod_tanh`.
5. Compute the product of the signs of all such `Q`'s (still, except the one coming from `V`). Call this `prod_signs`.
6. The outgoing message is `2 * prod_signs * atanh(prod_tanh)`.

In contrast, my code does:
1. For each variable node message `Q` except the one coming from `V`, compute `x = Q / 2`. 
2. Clamp `x` to a maximum of `9.0` and a minimum of `-9.0` (fixed) to produce `x_clamped`.
3. Compute `tanh(x_clamped)`.
4. Compute the product of all such `x`'s. Call this `prod_tanh`.
6. The outgoing message is `2 * atanh(prod_tanh)`.

My rationale for clamping to `9.0` is that `np.tanh(np.float32(9.0)) = 0.99999994`, while `np.tanh(np.float32(10.0)) = 1.0`. The goal is to avoid `prod_tanh` from being equal to `1.0` (or too close) or `-1.0` (or too close), since `np.arctanh(np.float32(1.0)) = inf`. But `np.arctanh(np.tanh(np.float32(9.0))) = 8.66434`, so clamping the `x`'s to `9.0` is fine. If we do this, each `tanh(x_clamped)` will be smaller or equal than `0.99999994`, so the product will also be. In this region, `atanh` is not very precise (as evidenced by `np.arctanh(np.tanh(np.float32(9.0))) = 8.66434`, but we don't care much; we only care about having numerical stability).

The other lesson learned is that the `tahn`'s and `atanh`'s carry the signs without any trouble, so it's not necessary to treat them separately (other update rules really work with absolute values of LLRs and need the signs to be handled separately). But in this case the difference between treating the signs separately or not should only affect CPU performance.

VladislavF:
> My rationale for clamping to `9.0` is that `np.tanh(np.float32(9.0)) = 0.99999994`, while `np.tanh(np.float32(10.0)) = 1.0`. The goal is to avoid `prod_tanh` from being equal to `1.0` (or too close) or `-1.0` (or too close), since `np.arctanh(np.float32(1.0)) = inf`. But `np.arctanh(np.tanh(np.float32(9.0))) = 8.66434`, so clamping the `x`'s to `9.0` is fine. If we do this, each `tanh(x_clamped)` will be smaller or equal than `0.99999994`, so the product will also be. In this region, `atanh` is not very precise (as evidenced by `np.arctanh(np.tanh(np.float32(9.0))) = 8.66434`), but we don't care much; we only care about having numerical stability.

This is interesting, I get very bad BER, when setting the clamp value too high. you should be able to use my FG to see exactly what I'm doing.

> Only now I understood that the `awgn_bp` constructor parameter `sigma` now has the semantics of being the tanh clamp value, which is probably something that the user doesn't have to tweak. For `float32` I already [had good results clamping at `9.0`](https://github.com/daniestevez/ldpc-toolbox/blob/main/src/decoder/arithmetic.rs#L434), and didn't feel the need to make this configurable, nor to clamp at smaller values. In any case, I wouldn't call the clamp value `sigma` in the code, since it's quite possible to mistake it for the noise standard deviation (which is what the old `sigma` was). I would also consider this an API breaking change.

I just did this to help twiddle and play with things. not a change I'm attached to, and your points are valid. However, seeing as it affects performance I put it in for now.

> The other lesson learned is that the `tahn`'s and `atanh`'s carry the signs without any trouble, so it's not necessary to treat them separately (other update rules really work with absolute values of LLRs and need the signs to be handled separately). But in this case the difference between treating the signs separately or not should only affect CPU performance.

This is great, that can remove a couple lines and perhaps improve performance a bit. Thank you for your general description, I think without that sign step our code is almost identical. and really does beg the question of what's up with the tanh clamping. I've attached my one of my test flowgraphs with an alist file. It's more complicated than necessary since it's also testing soft decisions, but it is the exact same flow graph I used to characterize the performance vs. the old code.

VladislavF:
[test_soft_ldpc.zip](https://github.com/gnuradio/gnuradio/files/12134913/test_soft_ldpc.zip)
@daniestevez here's my test flow

daniestevez:
I think I just found a bug (or maybe it's too late here). Shouldn't the variable node update read:

```c++
    //(step 2) of LLR-BP tanh algo
    // iterate over columns of check matrix
    for (int var = 0; var < N; var++) {
        excluded = double(0.0);
        // iterate over nonzero columns of check matrix
        for (int i = 0; i < num_nlist[var]; i++) {
            _sum = rx_lr[var];
            c = nlist[var][i] - 1;
            // compute sum of LLRs excluding current value
            for (int iprime = 0; iprime < num_nlist[var]; iprime++) {
                d = nlist[var][iprime] - 1;
                // save the excluded value
                if (i == iprime) {
                    excluded = R[d][var];
                } else {
                    _sum = _sum + R[d][var];
                }
            }
            // this is fed back to node checks
            Q[c][var] = _sum;
        }
        // decision is made on this value
        lr[var] = _sum + excluded;
```

Note that I've moved `_sum = rx_lr[var]` to an inner for loop. Otherwise it seems to me that we're accumulating into `_sum` too many times.

There's also a typical CPU performance optimization that can be applied here: first add all the check nodes in one for loop to compute `lr[var]`. Then do another for loop in which we calculate `Q[c][var]` by subtraction as `lr[var] - R[d][var]`. This has 2 separate for loops instead of 2 nested for loops. But in any case I'm more concerned with Eb/N0 performance at the moment.

daniestevez:
> Note that I've moved `_sum = rx_lr[var]` to an inner for loop. Otherwise it seems to me that we're accumulating into `_sum` too many times.

I think that this fixes the Eb/N0 problems (plus replacing the 2.3 with 9.0, or even 18.0, since I've realized you're using `double` instead of `float`). I was working on an older version of your branch, and have now updated to have a clean state. I'm rebuilding all GR because of some pydoc changes, so I'll confirm when this finishes building.

If I'm right, then probably this bug was the worst offender in the main branch implementation of the decoder.

daniestevez:
Oh, no, the main branch variable update looks correct to me. It's using the technique of sum everything up, then subtract one by one that I suggested (though with products and divisions because it uses LRs instead of LLRs). It's difficult to read because there are variables called `_sum` and `__sum`.

daniestevez:
Yeah. Moving the `_sum = rx_lr[var]` solved the problems. This branch https://github.com/daniestevez/gnuradio/tree/vlad-ldpc works using my test at 2 dB Eb/N0 and a tanh clamp of `18.0` (note that the branch also includes my hack to get the systematic bits, which shouldn't end up into this PR).

I'd suggest to fix this probem and to hardcode the tanh clamping to `18.0`.

I'll look at your test flow another day, since now it's quite late here.

VladislavF:
> Yeah. Moving the `_sum = rx_lr[var]` solved the problems.

What would I do without you dani, I applied and pushed this fix. What's fascinating to me is that I've been doing just averaged BER tests, and this change doesn't affect them much. However it does significantly change the error distribution. I now find that errors are more bursty when looking at the difference plots between sent and received bits, with more space that's entirely error free than before. This would explain what you were seeing since you were looking at FER not BER which would be significantly worse with a more uniform error distribution.

> There's also a typical CPU performance optimization that can be applied here: first add all the check nodes in one for loop to compute `lr[var]`. Then do another for loop in which we calculate `Q[c][var]` by subtraction as `lr[var] - R[d][var]`. This has 2 separate for loops instead of 2 nested for loops. But in any case I'm more concerned with Eb/N0 performance at the moment.

I would love to incorporate this, but I just barely wrapped my brain around what was going on in this code 4 months ago, and now I've forgotten everything, with a little more detail I would gladly make the change.

VladislavF:
I vote to cancel binding hashes, dag nabbit!

daniestevez:
> I would love to incorporate this, but I just barely wrapped my brain around what was going on in this code 4 months ago, and now I've forgotten everything, with a little more detail I would gladly make the change.

This is what I meant:

```c++
void awgn_bp::update_vars()
{
    double _sum;
    int c;
    // (step 2) of LLR-BP tanh algo
    // iterate over columns of check matrix
    for (int var = 0; var < N; var++) {
        _sum = rx_rx[var];
        // iterate over nonzero columns of check matrix
        for (int i = 0; i < num_nlist[var]; i++) {
            c = nlist[var][i] - 1;
            _sum += R[c][var];
        }
        // decision is made on this value
        lr[var] = _sum;
        for (int i = 0; i < num_nlist[var]; i++) {
            // exclude the current value to generate the message fed back to the nodes
            c = nlist[var][i] - 1;
            Q[c][var] = _sum - R[c][var];
        }
    }
}
```

daniestevez:
I've also given some thought to how the bug of the location of `_sum = rx_rx[var];` affected the algorithm. The effect of this bug is to add multiple times the check node messages, both for the calculation of the variable messages, and for the calculation of the variable LLR on which decision is made.

Intuitively, this will have worse effects for variable nodes with higher degree, since the values computed for `_sum` deviate more from the correct value. The LDPC code you were using has most variable nodes of degree 3, and a few of degree 2. The LDPC code I was using has a mixture of degrees 1, 2, 3 and 6 (degree 3 appears twice as many time, and the other degrees are equally frequent). Therefore, for variable nodes of degree 6 in my code the effect would be much worse.

I also have the impression that the problem is worse for LDPC codes with an irregular variable degree distribution. If the variable degree distribution is regular or almost regular (as in your code), then the calculated values will tend to be off by a certain multiplicative constant, but it will be the same constant for all the nodes (for instance, for a check node of degree 3, we are calculating the decision LLR as `L = 2 * R[0] + 2 * R[1] + 3 * R[2]` instead of the correct `L = R[0] + R[1] + R[2]`, and the check messages as `Q[0] = R[1] + R[2]`, `Q[1] = R[0] + R[1] + 2 * R[2]`, `Q[2] = 2 * R[0] + 2 * R[1] + 2 * R[2]` instead of the correct `Q[0] = R[1] + R[2]`, `Q[1] = R[0] + R[2]`, `Q[2] = R[0] + R[1]`. My point is that if the degree is regular, then multiplicative constant is the same for all the variable nodes, but if not, then the multiplicative constant will depend on the degree of the node, which probably unbalance things. Specially in the case of regular degree, these ill effects are probably compensated somewhat by artificially clamping the values of the check node messages `R[j]`, so that even though the sums we're doing include them multiple times, the final result is similar in magnitude to the correct one. This perhaps explains why you obtained good results by clamping to `2.3` the argument of the tanh, while higher clamping values gave problems.

gr-fec/lib/awgn_bp.cc
```diff
@@ -78,7 +83,7 @@ void awgn_bp::rx_lr_calc(std::vector<float> codeword)
     float y;
     for (int i = 0; i < N; i++) {
         y = codeword[i];
-        rx_lr[i] = exp((-1 * y) / (2 * sigma * sigma));
+        rx_lr[i] = y;
```
daniestevez, line 86: I suggest the following comment, since there is a sign flip lurking in `ldpc_decoder` which has made me trip.

```suggestion
        // The ldpc_decoder performs a sign inversion before passing the LLR data to awgn_bp.
        // The input of awgn_bp is LLRs with the convention that a positive LLR indicates that the
        // bit 0 is more likely.
        rx_lr[i] = y;
```

gr-fec/lib/awgn_bp.cc
```diff
@@ -99,38 +104,70 @@ void awgn_bp::spa_initialize()
...
+                // remove divide by 2 for manual tanh
+                x = Q[chk][w] / 2.0;
+                // x = std::abs(Q[chk][w]);
+                // clamp tanh input (this is some BS, FIXME)
+                if (x > 18) {
+                    x = 18;
+                }
+                if (x < -18) {
+                    x = -18;
+                }
```
daniestevez, line 133:
```suggestion
                x = branchless_clip(x, 18.0);
```

We can use [`branchless_clip()`](https://github.com/gnuradio/gnuradio/blob/main/gnuradio-runtime/include/gnuradio/math.h#L77) here.

gr-fec/lib/awgn_bp.cc
```diff
@@ -234,6 +271,7 @@ std::vector<uint8_t> awgn_bp::decode(std::vector<float> rx_word, int* niteration
                 break;
             }
         }
+        // std::cout << "niters:" << *niteration << std::endl;
```
daniestevez, line 274: Do we need this comment anymore?

daniestevez (commented):
I've gone through the LDPC decoder code and put a few comments. Still I'd need to go through the constellation decoder changes, since I've mostly forgotten about those.

gr-fec/lib/ldpc_decoder.cc
```diff
@@ -24,21 +24,20 @@
...
+ldpc_decoder::ldpc_decoder(std::string alist_file, int max_iterations)
     : generic_decoder("ldpc_decoder")
 {
     if (!std::filesystem::exists(alist_file))
         throw std::runtime_error("Bad AList file name!");
 
     d_list.read(alist_file.c_str());
     d_code.set_alist(d_list);
-    d_spa.set_alist_sigma(d_list, sigma);
+    d_spa.set_alist_sigma(d_list, 1.0);
```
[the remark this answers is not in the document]
VladislavF: yes, I was trying to do that, yet there's some API domino effect I'm not understanding because removing that function caused a segfault. I'll investigate more now

VladislavF:
ah, sorry for the commit spam. Had to rebase due to github not signing the commit when I approved code changes.

gr-digital/lib/constellation.cc
```diff
@@ -269,10 +319,10 @@ std::vector<float> constellation::calc_soft_dec(gr_complex sample, float npwr)
     for (int i = 0; i < M; i++) {
         // Calculate the distance between the sample and the current
         // constellation point.
-        float dist = std::abs(sample - d_constellation[i]);
+        float dist = powf(std::abs(sample - d_constellation[i]), 2.0f);
```
daniestevez, line 322: Probably this generates better asm code, avoiding to compute a square root and then `powf`.
```suggestion
        const gr_complex z = sample - d_constellation[i];
        float dist = z.real() * z.real() + z.imag() * z.imag();
```

gr-digital/lib/constellation.cc
```diff
@@ -269,10 +319,10 @@ std::vector<float> constellation::calc_soft_dec(gr_complex sample, float npwr)
...
         // Calculate the probability factor from the distance and
-        // the scaled noise power.
-        float d = expf(-dist / npwr);
+        // the scaled noise power. (divisor 2.0f for noise amplitude, 1.0f for power?)
+        float d = expf(-dist / (d_npwr * 1.0f));
```
daniestevez, line 325: I'm not sure about this. I would expect a `2.0` here.

The probability density function for receiving `sample` conditioned to the fact that the transmitted symbol was `d_constellation[i]` is `exp(-|sample - d_constellation[i]|^2/(2 * sigma^2))`, where `sigma` is the noise standard deviation.

I'd have to check all the math here to be sure, since the way I'm used to computing LLRs for received symbols from a constellation is to compute `a[i] = Re(sample * conj(d_constellation[i])) / sigma^2`, then max* reduce the `a[i]` for those symbols in which the bit has the value 0 (call the result `b`), max* reduce the `a[i]` for those symbols in which the bit has the value 1 (call the result `c`), and subtract `c - b`.

daniestevez: Looking more at this code (and comparing with some notes I have for how this can be computed in different ways), I see that what the code is doing is to sum the PDFs of `sample` conditioned to `d_constellation[i]` for those `i`'s for which the bit is 1 (call this `one`), sum the PDFs of `sample` conditioned to `d_constellation[i]` for those `i`'s for which the bit is 0 (call this `zero`), and then computing `log(one) - log(zero)`. This is the way to compute the LLR following the definition.

So now I'm convinced that it should be
```c++
float d = expf(-dist / (d_npwr * 2.0f));
```

gr-digital/lib/constellation.cc
```diff
@@ -297,7 +347,16 @@ std::vector<float> constellation::calc_soft_dec(gr_complex sample, float npwr)
...
-        s[k - 1 - i] = (logf(tmp[2 * i + 1]) - logf(tmp[2 * i + 0]));
+        float one = tmp[2 * i + 1];
+        float zero = tmp[2 * i + 0];
+        // clamp input to log to prevent NaN and Inf.
+        if (one < 3.4e-38) {
+            one = 3.4e-38;
+        }
+        if (zero < 3.4e-38) {
+            zero = 3.4e-38;
+        }
```
daniestevez, line 358: I'm curious about were the `3.4e-38` came from. My numpy is fine down to `1e-45`:
```python
>>> np.log(np.float32(1e-45))
-103.27893
```

Also, I wonder whether we hit this in practice. For this to happen, the sum of probabilities (for `one` or for `zero`) would need to be very close to zero. For this to happen, since those probabilities are the PDF of a gaussian, that would happen if the Euclidean distance from `sample` to all the constellation points corresponding to a bit value is larger than about 13 `noise_sigma` (because `exp(-90) = 8.19e-40` and `sqrt(2 * 90) = 13.4`). I guess this isn't impossible if one sets `noise_sigma` to a small value (i.e., if one expects a very high SNR), so it's good to have this check.

gr-digital/grc/digital_constellation_soft_decoder_cf.block.yml
```diff
@@ -6,6 +6,10 @@ parameters:
 -   id: constellation
     label: Constellation Object
     dtype: raw
+-   id: npwr
+    label: Noise Power
+    dtype: float
+    default: '-1'
```
daniestevez, line 12: Is a default noise power of `-1` intended?

gr-digital/lib/constellation.cc
VladislavF: I would tend to agree, I remember looking at this before. The only thing I don't understand is why this change seems to improve BER. I suppose I could be accidentally calculating my noise power incorrectly. My attached flowgraph illustrates this difference. I calculate my noise power as 10^(-SNR_dB/10) which is the input parameter sigma. Maybe there's a missing power of 2 or square root somewhere.

gr-digital/lib/constellation.cc
```diff
@@ -632,6 +707,8 @@ constellation_bpsk::constellation_bpsk()
     d_constellation[1] = gr_complex(1, 0);
     d_rotational_symmetry = 2;
     d_dimensionality = 1;
+    // normalization_t norm = POWER_NORMALIZATION;
+    // normalize(norm);
```
daniestevez, line 711: Does removing this code cause any behavior changes? I've sometimes found that the default normalization of constellations is not what I want and done some changes by hand (such as multiply my signal by a constant). I wonder if I'd need to tweak my changes due to this.

gr-digital/lib/constellation.cc
VladislavF: Yes, this is exactly why I put this check. Even modest snrs above 13db or so will begin producing spurious values. The odd clamp value was the float32 precision limit off the top of my head. Can be changed, just works.

gr-digital/grc/digital_constellation_soft_decoder_cf.block.yml
VladislavF: Yes, this is to allow the value set in the constellation object to take precedence. I.e. if you specified a LUT calculated for some sigma/SNR and you don't want it to be overriden by the decoder block.

gr-digital/lib/constellation.cc
VladislavF: Yes this is very frustrating. I think the default constellation amplitudes make no god damn sense at all which is why you see these lines. The unfortunate thing is there's several QA tests involving equalizers that depend on these bizarre and unscaled constellations. I've just decided that I don't want to open Pandora's box on fixing that and anything else this might break. Even if I think it's the correct solution. Hence I've left it as a comment for now.

VladislavF: So to be clear. These had no normalization by default. I thought this was strange and added it. This broke things that I didn't want to fix, so I've left comments as a reminder for the future.

gr-fec/include/gnuradio/fec/awgn_bp.h
```diff
@@ -36,14 +36,14 @@ class FEC_API awgn_bp
     //! Default constructor
     awgn_bp(){};
 
-    //! A constructor for given GF2Mat and sigma
-    awgn_bp(const GF2Mat X, float sgma);
```
willcode, line 40: API change

gr-fec/include/gnuradio/fec/awgn_bp.h
```diff
@@ -36,14 +36,14 @@ class FEC_API awgn_bp
...
+    //! A constructor for given GF2Mat
+    awgn_bp(const GF2Mat X);
 
-    //! A constructor for given alist and sigma
-    awgn_bp(alist _list, float sgma);
```
willcode, line 43: API change

gr-fec/include/gnuradio/fec/awgn_bp.h
```diff
@@ -36,14 +36,14 @@ class FEC_API awgn_bp
...
+    //! A constructor for given alist
+    awgn_bp(alist _list);
 
-    //! Initializes the class using given alist and sigma
-    void set_alist_sigma(alist _list, float sgma);
```
willcode, line 46: API change

gr-fec/include/gnuradio/fec/ldpc_decoder.h
```diff
@@ -33,7 +33,7 @@ class FEC_API ldpc_decoder : public generic_decoder
 {
 private:
     // private constructor
-    ldpc_decoder(std::string alist_file, float sigma, int max_iterations);
```
willcode, line 36: API change

willcode (commented):
We need a really good justification to change the API for these public functions. For example, nobody else has ever used these functions and they are completely broken and wouldn't work if someone tried to use them.

gr-digital/include/gnuradio/digital/constellation.h
```diff
@@ -127,7 +131,7 @@ class DIGITAL_API constellation : public std::enable_shared_from_this<constellat
      *
      * This is expensive to compute.
      */
-    void gen_soft_dec_lut(int precision, float npwr = 1.0);
```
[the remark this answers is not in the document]
VladislavF: This function does not make mathematical sense without a noise power argument. It's a key part of the calculation. Because it's an API change I've defaulted it to the previously hard coded value.

gr-digital/include/gnuradio/digital/constellation.h
```diff
@@ -148,7 +152,7 @@ class DIGITAL_API constellation : public std::enable_shared_from_this<constellat
      * \param sample The complex sample to get the soft decisions.
      * \param npwr Estimate of the noise power (if known).
      */
-    virtual std::vector<float> calc_soft_dec(gr_complex sample, float npwr = 1.0);
```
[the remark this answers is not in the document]
VladislavF: Same logic as above applies here

gr-digital/include/gnuradio/digital/constellation.h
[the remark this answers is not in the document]
VladislavF: Changing this resolved the need for the conditionals I had in the class previously that checked for the sentinel value from the soft decoder. (Previous PR).
As far as I can tell, adding noise power as an argument was a relatively new change and it was unused anyway.

gr-fec/include/gnuradio/fec/awgn_bp.h
VladislavF: The new LDPC algorithm does not have sigma as a parameter, therefore it is unused and unnecessary. I would also say it never should've been there in the first place since the soft decisions should already take this into account.

gr-fec/include/gnuradio/fec/awgn_bp.h
VladislavF: Necessary by above change

gr-fec/include/gnuradio/fec/awgn_bp.h
VladislavF: Necessary by above change, see first comment

gr-fec/include/gnuradio/fec/ldpc_decoder.h
VladislavF: Necessary by above change

gr-digital/include/gnuradio/digital/constellation.h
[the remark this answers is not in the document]
VladislavF: A user can achieve the same behavior as before by calling set_npwr(npwr) first and then using the LUT or soft dec functions. Both of these functions require information stored in class members anyway, so the API change will ensure their proper use and I think is better architecture.

willcode:
@VladislavF these functions have had these parameters for 4 years. We really can't remove them in 3.10, so there wouldn't be much point in merging to main. What options are there for leaving the params in, even if they are ignored, and would that produce any unexpected effects for the user.

VladislavF:
> @VladislavF these functions have had these parameters for 4 years. We really can't remove them in 3.10, so there wouldn't be much point in merging to main. What options are there for leaving the params in, even if they are ignored, and would that produce any unexpected effects for the user.

@willcode I suppose I could leave them in and just ignore them. That's fine, as long as they are not actually used, there will be no unexpected behavior to the user so long as they set the noise power of the constellation object at some point i.e. via the constructor/block or set_npwr() function. I think the Fedora build seems to complain about unused vars though.

Is this also going to be an issue with the LDPC decoder?

willcode:
I don't know how to determine if anyone is using those params. Sorry I didn't notice this earlier, but whatever we do has to work with any potential current users.

VladislavF:
@willcode what's the typical procedure here then? Because in the case of the LDPC decoder it's just not used at all by this algorithm.

In the case of the constellation, I think I can work around putting it back in defaulted to a sentinel value that those methods can check against and use either the member or argument that way.

willcode:
Leaving a default value with documentation that it is ignored is probably the best thing if it's not used at all. If someone had used it, would anything be unexpected in the behavior?

Sounds right for the constellation, and the behavior would be the same?

VladislavF:
I would argue that the LDPC decoder really was not being used by anyone since the issues with it were so severe as to render it useless in all but some very specific circumstances. It had totally non-functional behavior for many possible and reasonable values of sigma. I can put it back, but I feel that would be confusing to the user.

As for the constellation, I have successfully added the npwr argument back into both functions. the argument is defaulted to -1 which is checked by calc_soft_dec in which case it uses the member variable. If it is positive (negative npwr is not a reasonable input), then it will override the member variable and be used, thus preserving behavior for legacy users. If a legacy user calls the function without the argument and without using set_npwr(), behavior should still be preserved since the class defaults d_npwr=1.0 which is what the calc_soft_dec and gen_soft_dec_lut functions defaulted to previously.

willcode:
@daniestevez do you agree that the LDPC decoder was previously unusable by external users? I don't actually see any internal use of it, and google/bing/github searches don't turn up any instances of `ldpc_decoder::make`. This code was last touched (ignoring formatting, etc.) in 2015.

gr-fec/lib/awgn_bp.cc
```diff
@@ -99,38 +104,70 @@ void awgn_bp::spa_initialize()
...
+            for (int iprime = 0; iprime < num_mlist[chk]; iprime++) {
+                // exclude iprime from the product set
+                if (iprime == i) {
+                    continue;
+                }
+                w = mlist[chk][iprime] - 1;
+                // remove divide by 2 for manual tanh
+                x = Q[chk][w] / 2.0;
+                // x = std::abs(Q[chk][w]);
+                // clamp tanh input (this is some BS, FIXME)
```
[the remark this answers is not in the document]
marcusmueller: Exactly; clipping messages / LLRs is a very common thing in decoder implementations, due to both reasons of stability and of "OK, come on, we're sure to a probability of 1-10<sup>-11</sup> about this; there's no significant difference to being 1-10<sup>-10</sup> (this bit will never change!), so we might as well clip our LLRs at some value"

gr-fec/include/gnuradio/fec/awgn_bp.h
daniestevez: What I can say here is that the previous LDPC decoder which had a `sigma` was technically broken for anything besides BPSK and QPSK. It's not possible to compute LLRs correctly without a knowledge of the constellation. The old LDPC decoder always computed LLRs by using the BPSK/QPSK formula, so it wasn't possible to use it correctly with other constellations. In the new LDPC decoder, the task to compute LLRs has been moved to the constellation soft decoder (and for BPSK/QPSK the LLR can also be computed with a multiply).

I can't think of a way to fix this brokeness without an API change.

daniestevez:
> @VladislavF these functions have had these parameters for 4 years. We really can't remove them in 3.10, so there wouldn't be much point in merging to main.

How about instead of modifying the old LDPC decoder, we add this code as a new separate block and mark the old one as deprecated. This was recently done with the Throttle block. In this way we don't break someones hypothetical application (my take is that the old LDPC decoder has been broken forever, and no one has possibly used it in common real-world scenarios, so I don't know if anyone is actually using the old LDPC decoder), while the new decoder can be merged to 3.10 and people can start using it.

> What options are there for leaving the params in, even if they are ignored, and would that produce any unexpected effects for the user.

We could still keep the parameter `sigma` and use it as in the old decoder, so the old behaviour won't change (if you set `sigma` to the correct value for the noise std and you feed in a BPSK / QPSK signal of amplitude 1, the decoder will correctly compute LLRs for you), but instruct people through documentation that the intended way to use this block is to compute the LLR externally, and in this case `sigma` should be set to `sqrt(2)` so that it has no effect. This solution seems to cause more problems in the long run than it solves, though.

gr-digital/lib/constellation.cc
daniestevez: You're right. I was mistaking noise power and noise sigma for complex Gaussian noise. Since `noise_pwr = 2 * sigma^2`, the denominator needs to have `d_npwr`, not `2.0 * d_npwr`.

daniestevez:
> @daniestevez do you agree that the LDPC decoder was previously unusable by external users?

I think so. My impression is that except in some very particular cases (such as the example flowgraphs), the decoder was broken and produced wrong output. When I attempted to use it with a CCSDS code (which is rather common), it produced silly results (if I passed a codeword with a single bit error, the output was all zeros). I haven't heard of anyone trying it with a DVB-S2 code, which is another common real world family of codes, but I bet it wouldn't work either.

marcusmueller:
If that's the case, the proposal to deprecate (doesn't stop these special users from continuing to use it) and to supply a better (read: universally useful) decoder block is a winning strategy. 

> we add this code as a new separate block and mark the old one as deprecated. This was recently done with the Throttle block

ha! The new Throttle GRC block is the same C++ block underneath, it just offers access to the new, optional parameters.  That way, old Python/C++ code continues to work as before, but new users get "defaulted" to using something that will not potentially block for many seconds :) The old GRC block got renamed from "Throttle" to "Throttle (old)" and marked as deprecated on the GRC side.

willcode:
@marcusmueller Are you suggesting we duplicate the whole block?

gr-digital/lib/constellation.cc
VladislavF: Your change works :)

VladislavF:
@marcusmueller @willcode I know I've made my opinion clear already, but I do want to say that I would really like to avoid creating a whole new block with just 1 parameter removed.
@daniestevez will remember the genesis of this whole thing many moons ago was when I messaged him asking about this accursed LDPC decoder. He described his woes with trying to use it, and the conclusion we both came to then was that nobody uses it. Heck, even the QA code didn't need to be changed at all. the only internal use of 'sigma' was the example flowgraphs which mirrored the default constructor.

willcode:
Yah, I think I agree with just changing the decoder API, given the reports that it doesn't work. Still, holding off for a little bit to see if anyone else comments.

marcusmueller:
@willcode not at all, if we can avoid it! The good thing is we can have as many GRC blocks to set up one C++ block as we want, and we can have as many `make` functions as we want – we can even C++-`[[deprecate` the broken ones.

gr-digital/grc/digital_constellation_soft_decoder_cf.block.yml
VladislavF: @daniestevez just saw your edit, I agree that it's surprising, but this defaults the encoder to not override the noise power set by the constellation object. I feel this is a rational default since it lets the user know that's an option that they may otherwise never know. This allows a LUT to be set without being overridden by a change in noise power for instance. Should probably just add some additional documentation to the block to explain this.

VladislavF:
>The good thing is we can have as many GRC blocks to set up one C++ block as we want, and we can have as many `make` functions as we want – we can even C++-`[[deprecate` the broken ones.

@marcusmueller  Would you like to make a commit suggestion or give more detail on this?  as dumb as it sounds, I do not know how to make multiple blocks for one class

willcode:
The grc yml blocks have template code to make an instance (take a look at the tempates/cpp_templates sections of a yml file. In a new yml, if the id and label of the block are changed and everything else remains the same, there are now two grc blocks that do exactly the same thing.

VladislavF:
@willcode @marcusmueller Okay, I can do this, but what's the desired behavior? For the sigma parameter to be present but ignored? And also what of the set_alist_sigma() function, do we agree to make that change? i.e. it stays removed from the class.

willcode:
I still think that if the C++/Python decoder is absolutely unusable, as in nobody who ever used it would care if it was deleted altogether, then making the API change is OK and makes things simpler.

If sigma is totally broken, but anyone at all might be using the C++/Python code, leave signa in, ignore it, and document.

The duplicated/depricated GRC block is for the second case, where sigma would disappear in the new block. Here, the decision would be based on whether anyone could possibly have a flowgraph (that does anything they consider useful) including this block. If so, duplicate/deprecate.

I don't have the expertise to know whether this block is useful to anyone. Based on a quick search, I'd say it's unused in publicly shared code, but that's the best I can tell.

If we don't know the answer, (1) leave sigma in but ignore it, and (2) here or some time in the future duplicate/deprecate the GRC block.

daniestevez:
> Heck, even the QA code didn't need to be changed at all. the only internal use of 'sigma' was the example flowgraphs which mirrored the default constructor.

To be honest, the QA is not very good if it passes with the original block, which is mostly broken in practice. We'd need better QA, but that's a separate story.

VladislavF:
I think I'm about done toying with this now. if @willcode @daniestevez and @marcusmueller can come to a consensus with the LDPC decoder, I can implement it.

I vote to just make the API change since it is broken and unused (insofar as any of us can tell). Plus this is the cleanest most user-friendly option.

Alternatively, I could agree to adding a new make funtion with sigma that is called by a duplicate (deprecated block) and simply ignore the value. I would still want to drop the set_alist_sigma function since as Dani pointed out, having unused parameters and functions is just confusing to the user.

willcode:
I'll put out a note on the mailing list and chat. If nobody responds, we can make the API change for that one block.

VladislavF:
sorry, for all the CI cycles, I can't help but tweak.

willcode:
@VladislavF the CI failures here are unrelated - Python checker just got more picky - we'll have to look into it.

willcode:
@VladislavF would you please rebase on main, where I have turned off the E721 Python linter check. While you're doing that, could you make the commit message a little more concise, just describing the change and omitting the steps taken during the process? Hopefully, CI will be happy now and we can get this merged to main.

PR #6749, opened 2023-07-13
runtime: c++20 friendly logging

theartful:
when compiling in c++20 mode, fmt checks at compile time that the format string is constant expression. this is not the case with spdlog::string_view, and we get the error: ‘msg’ is not a constant expression so instead use spdlog::format_string_t (which actually defaults to string_view when using older compilers that don't support it)

## Checklist

### DIFF
--- a/gnuradio-runtime/include/gnuradio/logger.h
+++ b/gnuradio-runtime/include/gnuradio/logger.h
@@ -36,6 +36,7 @@ using logger_ptr = std::shared_ptr<void>;
 #include <spdlog/common.h>
 #include <spdlog/fmt/fmt.h>
 #include <spdlog/fmt/ostr.h>
+#include <spdlog/version.h>
 #include <memory>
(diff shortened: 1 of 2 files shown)


theartful:
I'm guessing the "Python bindings for logger.h are out of sync" is nothing I can fix, and is related to CI?

willcode:
@theartful run `gr-utils/bindtool/scripts/binding-hash-hotfixer.zsh` on each `.../bindings/..._python.cc` corresponding to a changed header file. This will update the MD5 hash.

This PR looks like it changes the API. @marcusmueller since this is related to a recent PR, what do you think would be the impact of these changes?

theartful:
forced push to change const ref to perfect forwarding, which is actually required for "format_string" to work.
I accidentally missed it in the PR.

marcusmueller (approved):

PR #6760, opened 2023-07-21
io_signature: API cleanup

### LINKED ISSUE #6758: io_signature exceptions are really bad
### What happened?

Got a user reporting an error when creating an io_signature from Python. I need to read code to learn what

     ValueError: gr::io_signature(3)

means. 

### System Information

3.11git

### GNU Radio Version

3.11-git (main)

### Specific Version

0.0.0.0

```Python
from gnuradio import gr
gr.io_signature.makev(10, 10, [gr.sizeof_gr_complex*1, gr.sizeof_gr_complex*1, gr.sizeof_float*1, gr.sizeof_float*1, gr.sizeof_float*1, 0*1, gr.sizeof_float*1, gr.sizeof_float*1, gr.sizeof_float*1, gr.sizeof_float*1])
```

### Relevant log output

```shell
ValueError                                Traceback (most recent call last)
Cell In[4], line 1
----> 1 gr.io_signature.makev(10, 10, [gr.sizeof_gr_complex*1, gr.sizeof_gr_complex*1, gr.sizeof_float*1, gr.sizeof_float*1, gr.sizeof_float*1, 0*1, gr.sizeof_float*1, gr.sizeof_float*1, gr.sizeof_float*1, gr.sizeof_float*1])

ValueError: gr::io_signature(3)
```

### Remarks

- Error messages for this could actually be self-explanatory. 
- I'm not sure whether changing an exception string would be considered an API break
- Using `fmt`, as available through spdlog anyways, one could have nice error mesages that explicitly say what's wrong
-  Good first issue for someone who knows how to compile GNU Radio!

marcusmueller:
`io_signature` is around the oldest code we have in the API, and its `make` variants make little sense.

Clean these up, deprecate the current API; and then, if consensus for that can be found, remove "bad" `makeX`s from main.

## Related Issue

Fixes #6758 

#6758 illustrated needs for better inspectabilty, and more sensible types, in `io_signature`.

Tree-wide

## Checklist

### DIFF
--- a/gnuradio-runtime/include/gnuradio/buffer_single_mapped.h
+++ b/gnuradio-runtime/include/gnuradio/buffer_single_mapped.h
@@ -12,7 +12,6 @@
 #define INCLUDED_GR_RUNTIME_BUFFER_SINGLE_MAPPED_H
 
 #include <cstddef>
-#include <functional>
 
 #include <gnuradio/api.h>
 #include <gnuradio/buffer.h>
--- a/gnuradio-runtime/include/gnuradio/buffer_type.h
+++ b/gnuradio-runtime/include/gnuradio/buffer_type.h
@@ -16,9 +16,7 @@
 
 #include <cstdint>
 #include <functional>
-#include <mutex>
 #include <string>
-#include <typeinfo>
 #include <vector>
 
 namespace gr {
--- a/gnuradio-runtime/include/gnuradio/io_signature.h
+++ b/gnuradio-runtime/include/gnuradio/io_signature.h
@@ -1,6 +1,7 @@
 /* -*- c++ -*- */
 /*
  * Copyright 2004,2007 Free Software Foundation, Inc.
+ * Copyright 2023 Marcus Müller
  *
  * This file is part of GNU Radio
  *
@@ -11,17 +12,20 @@
 #ifndef INCLUDED_IO_SIGNATURE_H
 #define INCLUDED_IO_SIGNATURE_H
 
+#include <gnuradio/api.h>
+#include <spdlog/fmt/fmt.h>
+#include <string_view>
+#include <memory>
+
+#include <gnuradio/buffer_type.h>
 // For testing purposes, force single mapped buffers to make all QA use them
 // #define FORCE_SINGLE_MAPPED
-
-#include <gnuradio/api.h>
 #ifdef FORCE_SINGLE_MAPPED
 #include <gnuradio/host_buffer.h>
 #else
 #include <gnuradio/buffer_double_mapped.h>
 #endif
-#include <gnuradio/buffer_type.h>
-#include <gnuradio/runtime_types.h>
+
 
 namespace gr {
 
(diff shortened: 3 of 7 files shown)


marcusmueller:
Up to "[runtime: io_signature: Prepare for API improvement](https://github.com/gnuradio/gnuradio/pull/6760/commits/238005f89b1c0d53631df9a9687ad03f9e2379be)" things should be API-stable

willcode:
What GR release are you thinking of for this? The deprecation warnings are going to hit other projects too, not just our CI.

marcusmueller:
> What GR release are you thinking of for this? The deprecation warnings are going to hit other projects too, not just our CI.

- Anything that breaks API: 3.11-git, only
- Deprecating API: 3.11-git & maint-3.10. The whole idea would be to give 3.10 consumers a timely heads up that future minor releases will not contain these functions. I don't think many OOTs will use `make3`, anyways.
- Replacing in-tree usage of now deprecated functions: 3.11 & 3.10. Gotta keep warning-free on -Wall.
- Deleting previously deprecated make2, make3: 3.11-git, possibly with enough delay.

willcode:
Even though main is technically 3.11, we're not really headed toward a 3.11, and holding things like this on main may make it difficult to maintain 3.10. The likelyhood of a 3.11 is a bigger discussion, which we ought to address elsewhere.

marcusmueller:
yep, agreed, I will not usher an age or even a discussion on a 3.11 era! That would be counter-productive.

Thing is, if we do keep a branch around from which we make backports to 3.10, then I'd like at least for that branch the API to low in cruft (as in: any bugfixes, and maybe even features, do not use the old `make.` interface). To allow backporting, we need to settle on one preferred alternative. To me, that'd be `make(int min, int max, vector<size_t> vec_sizes, vector<buffer_type> buftypes=default)`; instead of `makev(min,max,vec_sizes)` + `makev(min, max, vec_sizes, buftypes)`. You'll notice that this is mostly due to my preference of one overloaded function over having N different ones (knowing that C++ is sufficiently type-safe). So, if you say, meh, let's not introduce the vector-`make` and just mentally keep `makev` the desired interface, that's fine with me.

Why I did this was honestly because during #6759 it struck me how much code duplication is in there, including — as far as I can tell — unintended differences in allocated buffertype vectors for the different `make.` paths a developer might choose.

willcode:
Extending the API, and even changing in-tree blocks to use it, is fine. We just can't deprecate the old one. So, it's a question of how much work it is to maintain multiple signatures. Also, as we've seen with the Python signatures, you can add a new/better API but years later people still pick the old one because they're copying existing code. Certainly getting rid of all those ifdefs is a good thing - I like the single `using` ifdef.

marcusmueller:
yeah let me pull out that into a separate PR, that should be backportable separately anyways.

marcusmueller:
> Also, as we've seen with the Python signatures, you can add a new/better API but years later people still pick the old one because they're copying existing code.

100% agreed, which is why this PR main commit ("Prepare…") does the effort to really make all these interfaces actually do the same thing under the hood, and not be slightly different. That, should we ever need to fix bugs, we have to only fix them in one place. We don't have to deprecate, it's more of an option for us to "nudge" C++ devs into the right direction, long before, and only should we ever, release a 3.(10+2). release that doesn't contain the code duplication, by the sheer power of emitting annoying warnings during build ;-)

willcode:
Sounds like a plan. We can just hold the deprecattion PR with a Don't Merge (or just remember to do this later if there is a new minor release?). I'll read through this one again with that in mind.

willcode (approved):
@mormj could you take a quick look at the pybind stuff - I'm not as familiar

mormj:
pybind looks solid

mormj (approved):

willcode:
@marcusmueller This won't rebase for some reason.

marcusmueller:
huh, yeah github shows the same to me, but it's totally already rebased?!

marcusmueller:
I mean, even https://github.com/marcusmueller/gnuradio/tree/deprecate_iosig_factories says it's 4 commits ahead of main, not divergent at all?!

marcusmueller:
I'll add an empty commit, push, remove the commit, force-push to see whether I can get this unstuck, bear with me :(

marcusmueller:
yeah, that seems to have done it.

Issue #6762, opened 2023-07-27
Build fail maint-3.8 with gcc 13


MakiKyouko:
### What happened?

Build of maint-3.8 failed with gcc 13.
due to:
https://gcc.gnu.org/gcc-13/porting_to.html#header-dep-changes

### System Information

OS: archlinux
GR Installation Method: aur package gnuradio-maint38-git

### GNU Radio Version

3.8  (maint-3.8)

### Specific Version

0.0.0.0

### Steps to Reproduce the Problem

Compile source with gcc13.

### Relevant log output

```shell
/home/build/.cache/aurutils/sync/gnuradio-maint38-git/src/gnuradio/gnuradio-runtime/lib/pmt/pmt_serialize.cc: In function ‘bool pmt::serialize(pmt_t, std::streambuf&)’:
/home/build/.cache/aurutils/sync/gnuradio-maint38-git/src/gnuradio/gnuradio-runtime/lib/pmt/pmt_serialize.cc:277:52: error: ‘int32_t’ is not a member of ‘std’
  277 |                     ((i < std::numeric_limits<std::int32_t>::min() ||
      |                                                    ^~~~~~~
/home/build/.cache/aurutils/sync/gnuradio-maint38-git/src/gnuradio/gnuradio-runtime/lib/pmt/pmt_serialize.cc:277:52: note: suggested alternatives:
In file included from /usr/include/stdint.h:34,
                 from /usr/lib/gcc/x86_64-pc-linux-gnu/13.1.1/include/stdint.h:9,
                 from /home/build/.cache/aurutils/sync/gnuradio-maint38-git/src/gnuradio/gnuradio-runtime/lib/pmt/../../include/pmt/pmt.h:27,
                 from /home/build/.cache/aurutils/sync/gnuradio-maint38-git/src/gnuradio/gnuradio-runtime/lib/pmt/pmt_int.h:25,
                 from /home/build/.cache/aurutils/sync/gnuradio-maint38-git/src/gnuradio/gnuradio-runtime/lib/pmt/pmt_serialize.cc:27:
/usr/include/bits/stdint-intn.h:26:19: note:   ‘int32_t’
   26 | typedef __int32_t int32_t;
      |                   ^~~~~~~
/usr/include/bits/stdint-intn.h:26:19: note:   ‘int32_t’
[...]

PR #6763, opened 2023-07-27
runtime: pmt_serialize uses std::int32_t without <cstdint>

marcusmueller:
## Description

Currently, building this on GCC13 only works through "wonky" transitive includes; since PMT ideally works out-of-tree at some point, it's worth fixing this.

Also, fixing this might cheaply restore buildability of maint-3.8 with GCC13 ;)

## Related Issue

Found while looking at #6762 (thanks, @MakiKyouko)!

None, since impl

## Checklist

- [/] I have updated [the documentation](https://wiki.gnuradio.org/index.php/Main_Page#Documentation) where necessary.
- [/] I have added tests to cover my changes, and all previous tests pass.

### DIFF
--- a/gnuradio-runtime/lib/pmt/pmt_serialize.cc
+++ b/gnuradio-runtime/lib/pmt/pmt_serialize.cc
@@ -15,6 +15,7 @@
 #include "pmt/pmt_serial_tags.h"
 #include "pmt_int.h"
 #include <pmt/pmt.h>
+#include <cstdint>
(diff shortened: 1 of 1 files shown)


willcode:
I guess we could backport to the maint-3.8/3.9 trees, but there won't be a release.

marcusmueller:
@willcode  yeah, well aware of that, but meh, better to keep that code in sync across branches and functional, I guess

PR #6767, opened 2023-07-30
Reference added for turbo product code

fkvd:
## Description
Reference for turbo product code is added to make code more clear.

## Related Issue
This PR doesn't related to any issue.

Since the only contribution is comment, no block is affected.

## Checklist

### DIFF
--- a/gr-fec/include/gnuradio/fec/tpc_decoder.h
+++ b/gr-fec/include/gnuradio/fec/tpc_decoder.h
@@ -47,6 +47,8 @@ class FEC_API tpc_decoder : public generic_decoder
     int get_output_size() override;
     int get_input_size() override;
 
+	// Variables are denoted in Channel coding for IEEE 802.16e mobile WiMAX
+	// https://community.wvu.edu/~mcvalenti/documents/ICC2009.pdf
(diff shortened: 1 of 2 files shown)


willcode:
Was the code written based on this presentation, or is that a presentation on the same subject?

fkvd:
> Was the code written based on this presentation, or is that a presentation on the same subject?

Hello, I believe the code is written based on the presentation. Because, I couldn't find any other reference which name the related variables as in the code. Also it took me a long time to find this reference. So, I thought it could be helpful to add reference directly to the code. Also this link can be accessed via web archive.

willcode:
The wiki would be a good place. You could edit
https://wiki.gnuradio.org/index.php?title=TPC_Encoder_Definition
and
https://wiki.gnuradio.org/index.php?title=TPC_Decoder_Definition
adding a link to the presentation at the end of the intro (before Parameters) on each page.

I don't think we want to put a reference in code unless the code author (apparently @osh) says that the code is based on that reference.

fkvd:
Okay, thank you for your interest and guidance. I will edit wiki as soon as I have free time.

PR #6776, opened 2023-08-07
gr-iio: fix deadlock when destroying fmcomms2_source that didn't start

theartful:
## Description
fmcomms2_source creates a thread that monitors overflows. this thread is created on construction, and destroyed when the source is stopped. if an fmcomms2_source is constructed, but never started, the overflow thread will be stuck in an infinite loop. also if the fmcomms2_source is started and stopped multiple times, the overflow thread will be destroyed after the first stop.

this solution creates the thread on each start, and destroys it on each stop.

I could've been more efficient by adding a "paused" state, but I think that this is simpler.

## Checklist

### DIFF
--- a/gr-iio/lib/device_source_impl.cc
+++ b/gr-iio/lib/device_source_impl.cc
@@ -173,7 +173,8 @@ device_source_impl::device_source_impl(iio_context* ctx,
       buf(NULL),
       buffer_size(buffer_size),
       decimation(decimation),
-      destroy_ctx(destroy_ctx)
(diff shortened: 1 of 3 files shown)


willcode:
@adamhorden @tfcollins Would you guys give this a quick look? Seems like a good idea.

tfcollins:
Yes this is fine. @willcode would you want the virtual rename to be split to a separate commit?

willcode:
Adding `override` as we fix things is fine with me, since it doesn't touch the API.

PR #6785, opened 2023-08-14
Improvements for choosing external Python editor from GRC

### LINKED ISSUE #3567: Python Module open in editor error Win 10 64 bit
i install gnu-radio companion under win 10 , when i add python module and try to open in editor , i chose use default , i got error opening an external editor please select different editor ! , i try chose editor , but what should i chose ?? i use win 10 64 bit gnu-radio companion , any help ????

### LINKED ISSUE #6458: change the editor in an easy way
### Feature Description

Currently,
1. in linux, the GRC embedded python block's editor can only be changed by edit config.conf. would you pls try to make a easy way to do this.
2. in windows, I have to chooes the editor each time I open the epy file. Would you pls assign a default editor like linux?

Thanks!

### Feature Urgency

### More Information

_No response_

ryanvolz:
## Description

Last week in the matrix chat a user brought up that, when editing a Python Embedded Block, they needed to choose an external editor for every block and with every GRC restart. This and related issues have actually been reported before in #3567 and #6458. It turns out that a typo meant that the selected editor was never saved to the user's config file, and in digging around I also tried to address the other reported issues.

This PR has ~3~2 commits with progressively more invasive changes, and some of these can be backed off if desired:

~1) Fix saving the chosen external editor to the GNU Radio config file.~ Moved to #6786.
2) Add a button so that the editor can be re-chosen (previously, this was only possible by editing the config file). See the "Choose Editor" button now to the right of the "Open in Editor" button:
new_choose_editor_button
3) Use the GTK App Chooser dialog rather than the File Chooser dialog to select the editor (so users can choose from a list of relevant editor applications rather than having to know what editor they want and hunt down the path). Here's what the GTK App Chooser dialog looks like on my Linux system:
new_choose_editor_dialog

Particularly for that last one, testing on a broader set of systems would be helpful to make sure usability is not negatively impacted.

## Related Issue

Fixes #3567
Fixes #6458

GRC parameters of the PythonEditor type, which are used for things like the Embedded Python Block.

## Testing Done

I tried this out by editing the conda installation on my Linux system and in a Windows VM and verified it works in those two environments.

## Checklist

- ~[ ] I have added tests to cover my changes, and all previous tests pass.~

Wiki documentation (e.g. https://wiki.gnuradio.org/index.php/Embedded_Python_Block) will need to be updated.

### DIFF
--- a/grc/gui/Dialogs.py
+++ b/grc/gui/Dialogs.py
@@ -9,7 +9,7 @@ import sys
 import textwrap
 from shutil import which as find_executable
 
-from gi.repository import Gtk, GLib, Gdk
+from gi.repository import Gtk, GLib, Gdk, Gio
 
 from . import Utils, Actions, Constants
 from ..core import Messages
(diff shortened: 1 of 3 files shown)


willcode:
Could you separate out the first commit ... we'll put that one in right away and backport.

Selecting a new editor sounds great. I'd just like more opinions on the UI, especially placing the select button right next to the open-in-editor button.

ryanvolz:
> Could you separate out the first commit ... we'll put that one in right away and backport.

Done, #6786.

marcusmueller:
Definitely a usability improvement for about any user (I'm sure there's some that preferred to pick their xemacs-RPC-calling shell script, but frankly, they can edit the config file)!

Small problem:

    DeprecationWarning: Gtk.AppChooserDialog.new_for_content_type is deprecated

on Gtk4. Now, we're currently on Gtk3, so that's fine. But I wonder what the replacement is, and why!

ryanvolz:
> #gtk chat says that the Gtk4 replacement is GtkFileLauncher, and afaict, that's not in Gtk3, so, we have no choice here!

Right, it looks like `FileLauncher` is new and only in Gtk4, and it is a different API for the "open this file with a default application or ask the user what to use if necessary" use case that the `AppChooserDialog` handles in Gtk3.

It's maybe also worth noting that the `FileChooserDialog` (what is being used now without this change) has also been deprecated in Gtk4, replaced by `FileDialog`. So Gtk4 would force a change here regardless.

marcusmueller:
yeah, honestly not a fan in the change of "philosphy" there; the intent is to choose an application, not to choose an application for this specific file. But GTK4 as usual knows its users best, so we'll flow with it once that GTK4 bridge comes, I guess.

marcusmueller (approved):

PR #6787, opened 2023-08-15
Add buffer size parameter to RTLSDR Source

argilo:
## Description
Currently, there is no way to set the Buffer Size (`bufflen`) parameter of an RTLSDR source from within GNU Radio Companion. The default value (262144 bytes / 131072 samples) can be excessively large, particularly when operating at low sample rates. For instance, at 240000 sps the buffer is 546 ms long. This causes high latency and very slow update rates for GUI sinks.

Here I've added a USB Buffer Size parameter which allows this value to be adjusted.

Screenshot from 2023-08-15 17-29-38

* Soapy RTLSDR Source

## Testing Done
I tested using a flowgraph with an RTLSDR source connected to a Time Sink & Frequency Sink. With the sample rate set to 240000, the display becomes much more responsive when the buffer size is decreased to 16384.

## Checklist
- [ ] I have updated [the documentation](https://wiki.gnuradio.org/index.php/Main_Page#Documentation) where necessary. **I'll do this if and when this PR is merged.**
- [ ] ~~I have added tests to cover my changes,~~ and all previous tests pass.

### DIFF
--- a/gr-soapy/grc/soapy_rtlsdr_source.block.yml
+++ b/gr-soapy/grc/soapy_rtlsdr_source.block.yml
@@ -55,6 +55,15 @@ parameters:
     default: 'False'
     hide: 'part'
 
+  - id: bufflen
+    label: 'USB Buffer Size'
+    dtype: int
+    default: 16384
+    hide: 'part'
+
+asserts:
+  - ${ bufflen % 512 == 0 }
+
 inputs:
   - domain: message
     id: cmd
@@ -70,7 +79,7 @@ templates:
   make: |
       None
       dev = 'driver=rtlsdr'
-      stream_args = ''
+      stream_args = 'bufflen=${bufflen}'
       tune_args = ['']
       settings = ['']
 


argilo:
The RTL-SDR driver has had a default buffer size of 262144 bytes since 2012, so I decided to leave the block's default value the same to avoid changing the behaviour of existing flow graphs.

There could still be a case for changing it though. The RTL-SDR's large default buffer size has caused trouble for me (stuttering audio, slow refresh rates) on many occasions, even at high sample rates (e.g. https://github.com/gqrx-sdr/gqrx/issues/726#issuecomment-579045228). Since the Soapy blocks are relatively new, there might not be much risk of breakage.

willcode:
I was thinking that users won't generally know a good value to use for block size, or even that that's what should be adjusted, and it should "just work" since there is unlikely to be a performance problem.

argilo:
What do you think about making the default 16384? The RTL-SDR sources say it's best to use a multiple of 16384:

https://github.com/osmocom/rtl-sdr/blob/142325a93c6ad70f851f43434acfdf75e12dfe03/include/rtl-sdr.h#L364-L366

16384 would give a latency of 36 ms at the minimum sample rate of 225001.

There are 15 buffers by default, so there would be 38 ms of buffer at the maximum sample rate of 3.2 Msps, or 51 ms at the more common 2.4 Msps rate, which is the maximum recommended rate to avoid sample loss.

willcode:
Sounds good. Do you think it's clear that the buffer is the RTL buffer, and not a GR buffer? Or should the field be labeled more specifically?

willcode (approved):

PR #6788, opened 2023-08-15
Set the frequency sink's output multiple on start

### LINKED ISSUE #5759: update rate in QT GUI Frequency Sink is unintuitive/could be improved
### What happened?

Running a flowgraph with QT GUI Frequency sink and an update rate of 0.01 at sample rate of 48k with various FFT sizes yields sluggish/jerky response (the opposite of the expected behavior). If the FFT size is changed from the initial value, then changed back, on some systems the behavior is corrected, on others it's not.
After discussion with @argilo on the dev channel it was found that "the current work function implementation is basically 'if at least <fft size> samples have arrived and at least <update rate> time has passed, then update the display'."
This could be improved by a change in implementation similar to GQRX with a timer and sample buffer.

### System Information

GNURadio v3.10.2.0-rc2
Ubuntu 20.04 and 21.04
multiple hardware platforms/core counts

### GNU Radio Version

3.10 (maint-3.10)

### Specific Version

3.10.2.0-rc1

run a flowgraph and set the QT Freq GUI with update rate of 0.01, run at high/low sample rates (2Msps/48ksps)

### Relevant log output

_No response_

argilo:
## Description
The refresh rate of the QT GUI Time Sink can be painfully slow when the input sample rate is low. This was a side effect of #5442, which set the block's output multiple to `MAX_FFT_SIZE` = 32768 in the block's constructor, presumably to ensure that the scheduler allocates a large enough buffer to accommodate FFT size changes made in the GUI at runtime:

https://github.com/gnuradio/gnuradio/blob/0966207dbc5cc0077f7ee6cde0aaa95f7c7a8d99/gr-qtgui/lib/freq_sink_c_impl.cc#L128

If the FFT size is changed at runtime, the refresh rate speeds up because the output multiple is adjusted to the current FFT size:

https://github.com/gnuradio/gnuradio/blob/0966207dbc5cc0077f7ee6cde0aaa95f7c7a8d99/gr-qtgui/lib/freq_sink_c_impl.cc#L415

By doing the same in the `start()` method, the block will have the expected refresh rate, while still ensuring that the scheduler will allocate enough space to accommodate runtime FFT size changes.

## Related issue

Partially resolves #5759.

* QT GUI Frequency Sink

## Testing Done
I tested with the following flowgraph:

Screenshot from 2023-08-15 14-31-52

The frequency sinks have their update rates set to 0.01.

Before the change, the sinks update at 3 fps. After the change, they update at 100 fps.

Changing the FFT size at runtime still works properly, all the way up to the maximum FFT size of 32768.

## Checklist

### DIFF
--- a/gr-qtgui/lib/freq_sink_c_impl.cc
+++ b/gr-qtgui/lib/freq_sink_c_impl.cc
@@ -503,6 +503,12 @@ void freq_sink_c_impl::_test_trigger_norm(int nitems,
     }
 }
 
+bool freq_sink_c_impl::start()
+{
+    set_output_multiple(d_fftsize);
+    return true;
+}
+
 int freq_sink_c_impl::work(int noutput_items,
                            gr_vector_const_void_star& input_items,
                            gr_vector_void_star& output_items)
--- a/gr-qtgui/lib/freq_sink_c_impl.h
+++ b/gr-qtgui/lib/freq_sink_c_impl.h
@@ -166,6 +166,7 @@ public:
     int work(int noutput_items,
              gr_vector_const_void_star& input_items,
              gr_vector_void_star& output_items) override;
+    bool start() override;
 };
 
 } /* namespace qtgui */
--- a/gr-qtgui/lib/freq_sink_f_impl.cc
+++ b/gr-qtgui/lib/freq_sink_f_impl.cc
@@ -506,6 +506,12 @@ void freq_sink_f_impl::_test_trigger_norm(int nitems,
     }
 }
 
+bool freq_sink_f_impl::start()
+{
+    set_output_multiple(d_fftsize);
+    return true;
+}
+
 int freq_sink_f_impl::work(int noutput_items,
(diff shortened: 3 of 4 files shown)


willcode:
Since block.h marks this as virtual, and the description implies that it's meant to be overridden with useful functionality, I'd say there no need to call block::start. Some blocks do, but most don't. Otherwise, LGTM.

argilo:
> Just return `true`.

Done.

argilo:
I expect this will mostly resolve #5759, although there will still be some situations where the refresh rate will be slower than necessary due to the `work` function being called infrequently.

mbr0wn:
Why can't you just change the `set_output_multiple()` call in the constructor (== the one in `initialize()`)? We do already know the fft size at that point.

argilo:
The output multiple in the constructor needs to be large enough to accommodate runtime FFT size increases, otherwise the scheduler won't allocate a large enough buffer.

PR #6791, opened 2023-08-16
qt integration

R-ohit-B-isht:
## Description

Integrates the gr-qtgui flowgraph’s Qt GUI widget component Qt Designer to design the UI components.
wiki page : https://wiki.gnuradio.org/index.php?title=QT_Designer_Integration

flow_graph.py.mako
top_block.py
param.py
Actions.py
Application.py
Bars.py
## Testing Done

yes
## Checklist


cafeclimber:
Hey Rohit, I think improving the way widgets are laid out would be welcomed by many. Ultimately, my question is since we already have PyQtGraph as a dependency, would it make sense to leverage their Dock widgets in addition to something like supporting .ui files? I'm picturing everything wrapped in a [`DockWidget`](https://pyqtgraph.readthedocs.io/en/latest/api_reference/dockarea.html) so the user can drag/drop/resize widgets at runtime. That might make more sense for something like plots, but curious to hear your thoughts

grc/core/generator/flow_graph.py.mako
```diff
@@ -78,23 +82,31 @@ class ${class_name}(gr.top_block, Qt.QWidget):
...
-        self.top_scroll = Qt.QScrollArea()
-        self.top_scroll.setFrameStyle(Qt.QFrame.NoFrame)
-        self.top_scroll_layout.addWidget(self.top_scroll)
-        self.top_scroll.setWidgetResizable(True)
-        self.top_widget = Qt.QWidget()
-        self.top_scroll.setWidget(self.top_widget)
-        self.top_layout = Qt.QVBoxLayout(self.top_widget)
-        self.top_grid_layout = Qt.QGridLayout()
-        self.top_layout.addLayout(self.top_grid_layout)
+        file_path = os.fspath(Path(__file__).resolve().parent / "${class_name}.ui")
```
cafeclimber, line 85: I think you can just stick with using `Path` everywhere. This could be

```python
file_path = (Path(__file__).resolve().parent / "${class_name}.ui").absolute()
```

Then later just use `Path` methods like `if file_path.exists()`

grc/core/params/param.py
```diff
@@ -423,11 +424,12 @@ def collision_detection(row, col, row_span, col_span):
 
             if self.parent_flowgraph.get_option('output_language') == 'python':
                 widget_str = textwrap.dedent("""
-                    self.{layout}.addWidget({widget}, {row}, {col}, {row_span}, {col_span})
-                    for r in range({row}, {row_end}):
-                        self.{layout}.setRowStretch(r, 1)
-                    for c in range({col}, {col_end}):
-                        self.{layout}.setColumnStretch(c, 1)
+                    if not os.path.exists(file_path):
```
cafeclimber, line 427: It's more modern to use f-strings instead of `.format()` (I know much of the rest of the codebase uses `.format()`, but I think we should work to use modern Python techniques where possible!)

grc/core/generator/flow_graph.py.mako
```diff
@@ -28,6 +28,10 @@
 from packaging.version import Version as StrictVersion
 from PyQt5 import Qt
```
cafeclimber, line 29: I've seen this in some other places, but think it's caused some issues at least for me. I believe it's more idiomatic to import e.g. widgets from `QtWidgets`. I was always under the impression the `Qt` namespace was more for constants and enums like `Qt.AlignLeft` etc. When I run the test_examples test, I get loads of AttributeErrors about `QPushButton` not being a part of `PyQt5.Qt`

grc/core/params/param.py
marcusmueller: seconding that; a lot of the Python you find in-tree is from a Python 2.7-heritage (as that was all we had in GNU Radio 3.7), and we instantly bumped the requirement to Python 3.6 when we released GNU Radio 3.8, exactly because we wanted f-strings!

cafeclimber: Yeah I've been trying to work through linter errors/warnings from rough and have found a lot of 2.7-migrated-to-3.1+ style code :P

PR #6794, opened 2023-08-20
QTGUI: Applications show a confusing error message on first start

dl1ksv:
## Description

Create a new qt based flowgraph in grc and run it.
Than you get on the first run:
Qt GUI: Could not restore geometry: restoreGeometry(self, Union[QByteArray, bytes, bytearray]): argument 1 has unexpected type 'NoneType'
but the fg starts.

On the second run this message has gone.

## Testing Done

For testing you can use the gui_sink example in the gr-qtgui/examples directory.

Make sure that no gui_sink.conf exists in the ~/.config/GNU Radio/ directory.
Running the flowgraph shows the mentioned error message and creates gui_sink.conf, so on the second run the fg starts without errors.  But if you delete the config file , you'll get this error again.
 
## Checklist

### DIFF
--- a/grc/core/generator/flow_graph.py.mako
+++ b/grc/core/generator/flow_graph.py.mako
@@ -99,10 +99,11 @@ class ${class_name}(gr.top_block, Qt.QWidget):
         self.settings = Qt.QSettings("GNU Radio", "${class_name}")
 
         try:
-            if StrictVersion(Qt.qVersion()) < StrictVersion("5.0.0"):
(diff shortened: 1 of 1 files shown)


marcusmueller:
yep, we should be testing for presence of that key in flow_graph.py.mako:100 before trying to restore a geometry that isn't there.

marcusmueller:
@dl1ksv  can you test whether #6796 solves the problem?

dl1ksv:
#6796 is a better fix

PR #6796, opened 2023-08-20
grc/qtgui template: check for nonexisting setting; don't check for Qt4

marcusmueller:
We simply do not have any branch with Qt4 support alive

## Description

We check whether we have Qt < 5 (which we don't need, we don't build with Qt4 anymore, for quite some time), but we ignore when we try to extract a QSetting that doesn't exist (and then spam the user with a message on stderr)

## Related Issue

Fixes #6794

GRC flowgraphs generated with the qtgui option

## Checklist

### DIFF
--- a/grc/core/generator/flow_graph.py.mako
+++ b/grc/core/generator/flow_graph.py.mako
@@ -25,7 +25,6 @@
 ##Create Imports
 ########################################################
 % if generate_options == 'qt_gui':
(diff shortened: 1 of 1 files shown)


marcusmueller:
please wait for green light from @dl1ksv

dl1ksv (approved):
This fix works and is better than #6794

PR #6813, opened 2023-08-28
Repeat: itemwise, solve buffer allocation issue

### LINKED ISSUE #6809: Repeat block: buffer overflow for large interpolations
### What happened?

Consider this simple flow graph:
grafik

When changing the value of repeat from 1k to 10k at runtime the data flow stops with the following error:

    thread_body_wrapper :error: ERROR thread[thread-per-block[1]: <block repeat(2)>]: Buffer too small for min_noutput_items

However, if the repeat is already set to 10k by default when the flow graph starts, sufficient buffer is allocated and no error occurs.

I consider this a bug because there is no fundamental reason for the repeat block to allocate such a large buffer in the first place. **Instead of exceeding the allocated buffer size, the block should put back-pressure on the upstream block.** Also, the same functionality can be realized with two subsequent repeat blocks without such issues (limiting the granularity of possible interpolation of course):
grafik

### System Information

OS: Linux Ubuntu
GR Installation Method: Build from source

### GNU Radio Version

3.10 (maint-3.10)

### Specific Version

3.10.7.0

See above

### Relevant log output

_No response_

marcusmueller:
## Description

Repeat is a block that's commonly runtime-reconfigured. However, if the new interpolation is larger than the one specified at instantiation, this can lead to unsatisfiable output buffer requirements.

This change makes it so that the block is not a `sync_interpolator` anymore, but simply keeps track of how many times the last item still has to be repeated.

On the way there:

- Changed block type from `sync_interpolator` to `block`
- Hence, implemented a correct `forecast`
- Constructor and interpolation setter check for interpolation >= 1
- `general_work`: aforementioned handling of partial repeats
- `qa_repeat`: original QA was totally insufficient and ignored most
  corner cases. Also, checking code was very slow.
- Add the PMT formatter for fmt/spdlog that's necessary for setter logging

## Related Issue

Closes #6809

No API changes

## Testing Done

significantly extended the test coverage: more items, more strange combinations

## Checklist

### DIFF
--- a/gr-blocks/lib/repeat_impl.cc
+++ b/gr-blocks/lib/repeat_impl.cc
@@ -1,6 +1,7 @@
 /* -*- c++ -*- */
 /*
  * Copyright 2012 Free Software Foundation, Inc.
+ * Copyright 2023 Marcus Müller
  *
  * This file is part of GNU Radio
  *
@@ -8,12 +9,12 @@
  *
  */
 
-#ifdef HAVE_CONFIG_H
-#include "config.h"
-#endif
-
 #include "repeat_impl.h"
 #include <gnuradio/io_signature.h>
+#include <gnuradio/pmt_fmt.h>
+#include <algorithm>
+#include <cstddef>
+#include <stdexcept>
 
 namespace gr {
 namespace blocks {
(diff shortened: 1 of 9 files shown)


gr-blocks/lib/repeat_impl.cc
```diff
@@ -24,50 +25,134 @@ repeat::sptr repeat::make(size_t itemsize, int interp)
...
+    d_interp = new_interp;
 }
 void repeat_impl::set_interpolation(int interp)
 {
     // This ensures that interpolation is only changed between calls to work
     // (and not in the middle of an ongoing work)
-    _post(pmt::mp("interpolation"),                                   /* port */
-          pmt::cons(pmt::mp("interpolation"), pmt::from_long(interp)) /* pair */
+    if (interp < 1) {
+        d_logger->error("Requested interpolation {} is not strictly positive", interp);
```
willcode, line 66: "Repeat block interpolation set to {}, but must be a positive integer." or similar - don't want to bother user with "strictly positive". Does something in logging already give the user any info on what block may have caused the message? Same kind of message in other cases above would be nice. Can we call set_interpolation from make to avoid doing the check in 2 places?

willcode (changes requested):
Looks good - maybe tweak errors users are most likely to see.

gr-blocks/lib/repeat_impl.cc
marcusmueller: yes, the logging system tells you the block alias :)

Agreed, "strictly positive" is probably overkill. Will rework.

gr-blocks/lib/repeat_impl.cc
```diff
@@ -24,50 +25,142 @@ repeat::sptr repeat::make(size_t itemsize, int interp)
...
+    }
+
+
+    // making sure we get enough input
+    size_t requested_round_up = (static_cast<size_t>(noutput_items) -
+                                 d_left_to_copy) // amount we haven't "prepared"
+                                + d_interp - 1;  // round up to multiple of d_interp
+
+    // only one possible input
+    required[0] = requested_round_up / d_interp;
```
willcode, line 99: Example: nout = 10, left to copy = 3, interp = 4
10 - 3 + 4 - 1 = 10
10 / 4 = 2
Should the answer be 1? Were the parens meant to do something here?

Issue #6800, opened 2023-08-21
SIN and COS Calculations are inconsistent


VladislavF:
### What happened?

The calculations for SIN and COS are not consistent depending on sample rate or even which blocks are used. See the attached flowgraph. Using a signal source set to sine or cosine and a time sink with a number of samples set to the sample rate should yield an identical result for each refresh, at least I'd imagine. This however, is only true if the sample rate is a power of two (like 32e3). 
This issue also manifests if using a complex VCO and multiply together with a frequency shift/rotator block. These are incompatible even though they should produce a mathematically identical result. using a shift->shift is fine, using a VCO+mult.->VCO+Mult. is fine, using a VCO+Mult.->shift has an inaccurate result. This is regardless of sample rate.
The more worrying thing investigating this is that the root cause likely lies with the fact that the rotator block uses VOLK for calculating sin/cos which defaults to using std::sin or std::cos whereas the analog source and VCO use this fxpt function which manually implements some sort of table lookup. These two are not consistent. Architecturally, I don't understand why there is a custom sin/cos implementation at all. Why not use VOLK/std:: everywhere?
This seems like a pretty severe issue given it's such a fundamental calculation.

As far as I can tell, this happens in any version of GR. tested with 3.10.0 and latest 3.11 main.
[sin_test.zip](https://github.com/gnuradio/gnuradio/files/12398356/sin_test.zip)

### System Information

OS: Linux/WSL
GR Installation Method: pybombs

### GNU Radio Version

3.11-git (main)

### Specific Version

3.11.0

### Steps to Reproduce the Problem

Use a frequency shift or rotator together with a complex VCO and multiply block.
Use a signal source and time sink with a sample rate set to anything except for a power of two.

willcode:
There could be cases where there is efficient SIMD method and std is slower than lookup. Getting the same value result (maybe with some noise) from the fxpt implementation is certainly a reasonable requirement.

marcusmueller:
yeah I have had a gr::math fxpt benchmark from the 3.6.5 days (someone had to shift the frequency of a target, and instead of making his Bachelor thesis actually work, he might or might not have done 1000 things around the source code), and honestly, with the compilers back then, I was not extremely impressed. This might have improved. (also, I was inexperienced in running benchmarks back then, so take with a grain of salt)

VladislavF:
[...]

PR #6815, opened 2023-08-29
blocks: re-enable benchmarking tests

marcusmueller:
Were probably disabled since they don't compile with current GR.

## Description

These are non-unit tests, but benchmarks. They still serve the purpose of compile-checking, and take very little time to build.

While we're at it, remove the now defunct vco test, replace it with a polynomial sine approximation. 

Takeaway: approximation through a float polynomial that has less than 3.5e-6 of absolute error over the whole [-pi, pi] range is faster than our fxpt implementation.

## Related Issue

#6800 inspired me to look into the in-tree benchmarks

nöthing

## Checklist


willcode:
How do we use the information from the polynomial vco benchmark if this method isn't used anywhere?

marcusmueller:
> How do we use the information from the polynomial vco benchmark if this method isn't used anywhere?

:D That's the thing, someone might be willing to triage "oscillators broken" as fixable mistake once they know there's a viable alternatives.

willcode:
> > How do we use the information from the polynomial vco benchmark if this method isn't used anywhere?
> 
> :D That's the thing, someone might be willing to triage "oscillators broken" as fixable mistake once they know there's a viable alternatives.

OK, maybe someone will stumble across this and make changes that improve performance and fix 6800.

marcusmueller:
No seriously, I'm getting started doing the triage issues job, and #6800 is kind of severe for an SDR framework. I had this polynomial code lying around, so and I needed to compare it to what GNU Radio does otherwise, so I was kind of happy to find `benchmark_.co` here. Didn't build, so I fixed it. This obviously really isn't useful getting #6800 sorted out itself, but it's kind of infrastructure desirable to test the qualities of these implementations; `qa_fxpt_vco` isn't as good as it should be, so I do plan on migrating code to that once I get around to fixing the issue.

willcode:
As long as this is self-contained, go ahead and merge when you're done. Any chance of dependency or compile problems for unwitting users?

marcusmueller:
Don't think so, but as you noticed, for a change that does little to actual functionality, this must be the one where I had to do the most changes for nothing but windows (and thankfully so, these were real issues that MSVC detected)

PR #6821, opened 2023-09-08
iio: make gr::io_signature for device_source and device_sink dynamic

adisuciu:
This modification enables using devices that have arbitrary channel size or channels with different sizes accross the same device

## Description

Internally we found that when using gr-iio blocks with devices that do not have 16 bit channels, the data received is malformed.
This is due to the signature being hardcoded to multiple "short" channels.

The output type added to the blocks in grc is there to prevent the companion from erroring - has no real effect on building the block. There is still a problem with devices that have heterogenous outputs such as accelerometers:
X - 32 bit signed int
Y - 32 bit signed int
Z - 32 bit signed int
Timestamp - 64bit unsigned int
However these come by pretty rarely when using gnuradio.

iio::device_source and iio::device_sink

## Testing Done

We tested this internally with various IIO devices. 

## Checklist

### DIFF
--- a/gr-iio/grc/iio_device_sink.block.yml
+++ b/gr-iio/grc/iio_device_sink.block.yml
@@ -48,9 +48,16 @@ parameters:
     dtype: string
     hide: ${( 'part' )}
 
+-   id: ch_type
+    label: Channel type
+    dtype: enum
+    options: [ complex, float, int, short, byte]
+    default: short
+    hide: part
+
 inputs:
 -   domain: stream
-    dtype: short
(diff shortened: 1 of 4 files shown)


marcusmueller:
Hi @adisuciu, is there something about this that's 3.10-specific? Otherwise I'd ask you to make the PR against `main`, as well (or if you prefer, exclusively)

willcode:
As Marcus says, PRs are generally against `main` and then we'll backport them as necessary. Please resubmit against `main`.

PR #6822, opened 2023-09-08
Adds a grid to the flowgraph canvas.

### LINKED ISSUE #5672: Add option to display grid in GRC canvas
### Feature Description

Add the ability to display the grid that blocks snap to in GRC. This should be toggleable and saved using a preference and have a menu item in the dropdown menu to enable/disable.

### Feature Urgency

low (just an idea)

### More Information

Thanks to @skoslowski for outlining the steps to implement the drawing of the grid:

> To implement it, you'll want to start here: https://github.com/gnuradio/gnuradio/blob/main/grc/gui/DrawingArea.py#L213
> The grid size is defined in https://github.com/gnuradio/gnuradio/blob/main/grc/gui/Constants.py#L50
> Best add a color in https://github.com/gnuradio/gnuradio/blob/main/grc/gui/canvas/colors.py
> To make it optional grep for some action like TOGGLE_SNAP_TO_GRID and add your own in the same manner. 

Examples of loading/using/saving the preference can be found here.

https://github.com/gnuradio/gnuradio/blob/b2c9623cbd548bd86250759007b80b61bd4a2a06/grc/gui/MainWindow.py#L143

https://github.com/gnuradio/gnuradio/blob/f44384ae1f061f257603b02288aef4db1b74d4be/grc/gui/Config.py#L165-L166

sdh11:
## Description
This adds a grid to the flowgraph canvas which can be optionally shown by the user.

## Related Issue
Fixes #5672. 

GRC

## Testing Done
Did basic testing to make sure the feature worked as expected.

## Checklist


willcode:
I think #5672 (despite the title) is also about snap-to-grid. Does this PR also snap?

sdh11:
Yes, it's driven by the same points as snap to grid, so blocks should be aligned.

willcode:
Updated Python formating on a couple of files using `tools/autoformat_python.sh` to make CI pass.

willcode:
Grid spacing is correct, but the blocks do not line up on the grid.
image
At some zoom levels, they do line up. At zoom 1 on my machine, they do not.

dkozel:
Sounds like a possible additional issue would be to change the block rendering to round up to an integer of the snap grid size? That might improve visual consistency? IIRC the original feature request came from the general chat room and I posted it on their behalf. I do think the grid has some use (defaulting to off) and, aside from the commented out portion, has a small code footprint.

grc/gui/DrawingArea.py
```diff
@@ -210,6 +212,54 @@ def draw(self, widget, cr):
...
+                cr.set_line_width(1.0)
+            
+                for w_i in points_w:
+                    cr.move_to(w_i, 0)
+                    cr.line_to(w_i, height)
+                for h_i in points_h:
+                    cr.move_to(0, h_i)
+                    cr.line_to(width, h_i)
+                    cr.stroke()
+            # elif grid_mode == "dots":
```
mbr0wn, line 237: I think we should either leave this in, or delete these lines. I'm for leaving it in, even if it's not obvious how to change it (I don't think this PR needs to add ways to set this through the GUI, for example).

mbr0wn:

PR #6823, opened 2023-09-08
iio: make gr::io_signature for device_source dynamic

adisuciu:
This modification enables using devices that have arbitrary channel size or channels with different sizes across the same device

## Description

Internally we found that when using gr-iio blocks with devices that do not have 16 bit channels, the data received is malformed.
This is due to the signature being hardcoded to multiple "short" channels. The solution consists in querying IIO for the channel format and setting the gr::io_signature at block consturction.

The output type added to the blocks in the gnuradio-companion is there to prevent the companion from erroring when connecting blocks to the device_source/device_sink that are not short. The output type in the companion has no real effect on building the block.  There is still a problem in the companion with devices that have heterogenous outputs such as accelerometers:
X - 32 bit signed int
Y - 32 bit signed int
Z - 32 bit signed int
Timestamp - 64bit unsigned int
However these come by pretty rarely when using gnuradio-companion with device_source / device_sink and specific blocks can be created if needed. The C++ blocks work correctly for these cases as well, it's just the companion that can get messed up.

iio::device_source and iio::device_sink

## Testing Done

We tested this internally with various IIO devices - Pluto, IIO accelerometers (with heterogeneous inputs), Precision ADCs (with 8 32-bit channels), etc

## Checklist

### DIFF
--- a/gr-iio/grc/iio_device_sink.block.yml
+++ b/gr-iio/grc/iio_device_sink.block.yml
@@ -48,9 +48,16 @@ parameters:
     dtype: string
     hide: ${( 'part' )}
 
+-   id: ch_type
+    label: Channel type
+    dtype: enum
+    options: [complex, int, short, byte]
+    default: short
+    hide: part
+
 inputs:
 -   domain: stream
-    dtype: short
+    dtype: ${ ch_type }
     multiplicity: ${ len(channels) }
 
 asserts:
--- a/gr-iio/lib/device_sink_impl.cc
+++ b/gr-iio/lib/device_sink_impl.cc
@@ -13,6 +13,7 @@
 
 #include "device_sink_impl.h"
 #include "device_source_impl.h"
+#include "iio_priv.h"
 #include <gnuradio/io_signature.h>
 
 #include <string>
@@ -90,6 +91,7 @@ device_sink_impl::device_sink_impl(iio_context* ctx,
       d_len_tag_key(pmt::PMT_NIL)
 {
     unsigned int nb_channels, i;
+    std::vector<int> signature;
 
     /* Set minimum input size */
     set_output_multiple(buffer_size / (interpolation + 1));
@@ -116,6 +118,8 @@ device_sink_impl::device_sink_impl(iio_context* ctx,
 
             iio_channel_enable(chn);
             channel_list.push_back(chn);
+
+            signature.push_back(channel_size(chn));
         }
     } else {
         for (std::vector<std::string>::const_iterator it = channels.begin();
@@ -132,10 +136,13 @@ device_sink_impl::device_sink_impl(iio_context* ctx,
             if (!iio_channel_is_enabled(chn))
                 throw std::runtime_error("Channel not enabled");
             channel_list.push_back(chn);
+
+            signature.push_back(channel_size(chn));
         }
     }
 
     set_params(params);
+    set_input_signature(gr::io_signature::makev(1, -1, signature));
 
     buf = iio_device_create_buffer(dev, buffer_size, cyclic);
     if (!buf)
@@ -215,7 +222,9 @@ int device_sink_impl::work(int noutput_items,
     }
 
     for (unsigned int i = 0; i < input_items.size(); i++)
-        channel_write(channel_list[i], input_items[i], noutput_items * sizeof(short));
+        channel_write(channel_list[i],
+                      input_items[i],
+                      noutput_items * channel_size(channel_list[i]));
 
     ret = iio_buffer_push(buf);
     if (ret < 0) {
--- a/gr-iio/lib/device_source_impl.cc
+++ b/gr-iio/lib/device_source_impl.cc
@@ -12,6 +12,7 @@
 #endif
 
 #include "device_source_impl.h"
+#include "iio_priv.h"
 #include <gnuradio/io_signature.h>
 #include <gnuradio/thread/thread.h>
 
@@ -177,6 +178,7 @@ device_source_impl::device_source_impl(iio_context* ctx,
       thread_stopped(false)
 {
     unsigned int nb_channels, i;
+    std::vector<int> signature;
 
     if (!ctx)
         throw std::runtime_error("Unable to create context");
@@ -200,6 +202,8 @@ device_source_impl::device_source_impl(iio_context* ctx,
 
             iio_channel_enable(chn);
             channel_list.push_back(chn);
+
+            signature.push_back(channel_size(chn));
         }
     } else {
         for (std::vector<std::string>::const_iterator it = channels.begin();
@@ -214,11 +218,14 @@ device_source_impl::device_source_impl(iio_context* ctx,
 
             iio_channel_enable(chn);
             channel_list.push_back(chn);
+
+            signature.push_back(channel_size(chn));
         }
     }
 
     set_params(params);
     set_output_multiple(0x400);
+    set_output_signature(gr::io_signature::makev(1, -1, signature));
 
     message_port_register_out(port_id);
 }
@@ -311,7 +318,8 @@ int device_source_impl::work(int noutput_items,
     unsigned long items = std::min(items_in_buffer, (unsigned long)noutput_items);
 
     for (size_t i = 0; i < output_items.size(); i++)
-        channel_read(channel_list[i], output_items[i], items * sizeof(short));
+        channel_read(
+            channel_list[i], output_items[i], items * channel_size(channel_list[i]));
 
     items_in_buffer -= items;
     byte_offset += items * iio_buffer_step(buf);
--- a/gr-iio/lib/iio_priv.h
+++ b/gr-iio/lib/iio_priv.h
@@ -1,0 +1,30 @@
+/* -*- c++ -*- */
+/*
+ * Copyright 2023 Analog Devices Inc.
+ * Author: Adrian Suciu <adrian.suciu@analog.com>
+
+ * SPDX-License-Identifier: GPL-3.0-or-later
+ *
+ */
+
+#ifndef INCLUDED_IIO_IIO_PRIV_H
+#define INCLUDED_IIO_IIO_PRIV_H
+
+#include <iio.h>
+
+namespace gr {
+namespace iio {
+inline unsigned int channel_size(const iio_channel* chn)
+{
+    // to round up a / b without moving to float, you can do (a + (b-1)) / b
+    // (31 + 7) / 8 = 4 // 31 bits - 4 bytes
+    // (32 + 7) / 8 = 4 // 32 bits - 4 bytes
+    // (33 + 7) / 8 = 5 // 33 bits - 5 bytes
+    const iio_data_format* data = iio_channel_get_data_format(chn);
+    return (data->length + (8 - 1)) / 8;
+}
+
+} // namespace iio
+} // namespace gr
+
+#endif // INCLUDED_IIO_IIO_PRIV_H
--- a/gr-iio/grc/iio_device_source.block.yml
+++ b/gr-iio/grc/iio_device_source.block.yml
@@ -42,9 +42,16 @@ parameters:
     default: packet_len
     hide: part
 
+-   id: ch_type
+    label: Channel type
+    dtype: enum
+    options: [complex, int, short, byte]
+    default: short
+    hide: part
+
 outputs:
 -   domain: stream
-    dtype: short
+    dtype: ${ ch_type }
     multiplicity: ${ len(channels) }
 -   domain: message
     id: msg


willcode:
Please run
`clang-format -i gr-iio/lib/device_sink_impl.cc`
`clang-format -i gr-iio/lib/device_source_impl.cc`
to make sure formatting uses project standards.

Personally, I prefer
`type *var`
over
`type* var`
too, but it doesn't match the rest of the code.

gr-iio/lib/device_sink_impl.cc
```diff
@@ -116,6 +117,11 @@ device_sink_impl::device_sink_impl(iio_context* ctx,
 
             iio_channel_enable(chn);
             channel_list.push_back(chn);
+
+            const iio_data_format *data = iio_channel_get_data_format(chn);
+            int bit_length = (data->length + (8 - 1)) / 8;
+            signature.push_back(bit_length);
+
```
marcusmueller, line 124: please avoid duplicating code. This could very well be a small `constexpr unsigned int bit_length(const iio_channel* data){…}` function in a private header in the lib/ directory. See how the same implementation in `device_source_impl` is better commented! It would be an advantage if that was the code I've read before reading this. But I read this first and scratched my head.

Generally, I'd say this applies to this whole `for` loop, but we're not going to force you to refactor this whole set of constructors now.

gr-iio/lib/device_sink_impl.cc
```diff
@@ -132,10 +138,15 @@ device_sink_impl::device_sink_impl(iio_context* ctx,
             if (!iio_channel_is_enabled(chn))
                 throw std::runtime_error("Channel not enabled");
             channel_list.push_back(chn);
+
+            const iio_data_format *data = iio_channel_get_data_format(chn);
+            int bit_length = (data->length + (8 - 1)) / 8;
+            signature.push_back(bit_length);
```
marcusmueller, line 144: see above regarding code duplication.

gr-iio/lib/device_source_impl.cc
```diff
@@ -200,6 +201,15 @@ device_source_impl::device_source_impl(iio_context* ctx,
...
             channel_list.push_back(chn);
+
+            const iio_data_format *data = iio_channel_get_data_format(chn);
+            // to round up a / b without moving to float, you can do (a + (b-1)) / b
+            // (31 + 7) / 8 = 4 // 31 bits - 4 bytes
+            // (32 + 7) / 8 = 4 // 32 bits - 4 bytes
+            // (33 + 7) / 8 = 5 // 33 bits - 5 bytes
+            int bit_length = (data->length + (8 - 1)) / 8;
+            signature.push_back(bit_length);
+
```
marcusmueller, line 212: See above re code duplication

gr-iio/lib/device_source_impl.cc
```diff
@@ -214,11 +224,16 @@ device_source_impl::device_source_impl(iio_context* ctx,
 
             iio_channel_enable(chn);
             channel_list.push_back(chn);
+
+            const iio_data_format *data = iio_channel_get_data_format(chn);
+            int bit_length = (data->length + (8 - 1)) / 8;
+            signature.push_back(bit_length);
```
marcusmueller, line 230: see above re code duplication

marcusmueller (changes requested):

adisuciu:
I've updated according to the comments. However .. 
I've tested mostly with maint-3.10, and things are working fine. 
When I moved to main however, I got weird results in both the companion and our software. It looks as if buffers are only containing half the data .. 

image
^ that should be a sine wave .. 

Do you know anything that may cause this ?

adisuciu:
We can close this until we get to the bottom of this ..

willcode:
The only change to gr-iio since 3.10.7.0 is a61091f7cd57d76ae312642fbecef1c4267211e3

willcode:
Not enough info to say why you are getting that interesting sine wave. Did you mean you wanted to close this PR? You can do that if you want, or you can mark it as draft (under the Reviewers list).

adisuciu:
Silly me, I didn't add the whole thing .. now it works as expected. Tested with Pluto and AD7124 (32 bit channel signature, 8 channels)

image

gr-iio/lib/iio_priv.h
```diff
@@ -0,0 +1,29 @@
+/* -*- c++ -*- */
+/*
+ * Copyright 2023 Analog Devices Inc.
+ * Author: Adrian Suciu <adrian.suciu@analog.com>
+
+ * SPDX-License-Identifier: GPL-3.0-or-later
+ *
+ */
+
+#ifndef IIO_PRIV_H
```
marcusmueller, line 10: To stay consistent with the other headers in this directory (and to avoid conflict with any IIO headers of the future?), I'd go with `INCLUDED_IIO_IIO_PRIV_H` instead

marcusmueller (approved):

gr-iio/lib/iio_priv.h
adisuciu: Fixed

gr-iio/lib/iio_priv.h
```diff
@@ -0,0 +1,29 @@
...
+ */
+
+#ifndef INCLUDED_IIO_IIO_PRIV_H
+#define INCLUDED_IIO_IIO_PRIV_H
+
+#include <iio.h>
+
+namespace gr {
+namespace iio {
+constexpr unsigned int channel_size(const iio_channel* chn)
```
marcusmueller, line 17: Ah, shoot, my bad, if you do it this way: because `iio_channel_get_data_format` is C and not C++, it can't be `constexpr`, so this whole function cannot be `constexpr`, which MSVC recognizes and hence fails to compile this.

This might not be happening on other platforms since iio.h declares the function `__attribute__(const)` there; I don't think there's an MSVC equivalent to that.

Now, what is a good way forward? Hard one: probably start by dropping the `constexpr` here (annoying, I know).

Maybe upstream fix libiio/iio.h to define their `__cnst` to evaluate to `constexpr` in case of `__cplusplus`, and then add a version check so that GNU Radio uses `constexpr` for `channel_size()` when the system's libiio allows for that.

Speaking of version checks: Note that iio.h doesn't have a `#define LIBIIO_VERSION` or similar, which makes making code compatible unnecessarily hard on people that work with libiio a lot :-)

adisuciu: I removed the constexpr and pushed .. 
I agree with LIBIIO_VERSION in iio.h (especially now that we're moving to libiio v1)
Regarding constexpr ? I'm a little confused, wouldn't __cnst be a macro in GR and used in iio_priv.h ?

gr-iio/lib/iio_priv.h
```diff
@@ -0,0 +1,29 @@
...
+ */
+
+#ifndef INCLUDED_IIO_IIO_PRIV_H
+#define INCLUDED_IIO_IIO_PRIV_H
+
+#include <iio.h>
+
+namespace gr {
+namespace iio {
+unsigned int channel_size(const iio_channel* chn)
```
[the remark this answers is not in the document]
adisuciu: I created iio_priv.cc. Wouldn't that work too ?

gr-iio/grc/iio_device_sink.block.yml
```diff
@@ -48,9 +48,16 @@ parameters:
     dtype: string
     hide: ${( 'part' )}
 
+-   id: ch_type
+    label: Channel type
+    dtype: enum
+    options: [complex, float, int, short, byte]
```
willcode, line 54: `float` and `int` have the same size. Are both types actually available once the signature propagates down to the iio code? Same for the source block.

willcode (changes requested):

gr-iio/grc/iio_device_sink.block.yml
adisuciu: No, just int .. I'm not aware of any devices that provide data in floating point format directly.
On that note, is there any integer 64 bit alternative available in the companion to replace complex with that ?

I'll push the changes tomorrow when I'm in the office.

willcode: GRC defines "Complex Integer 64". Types can be passed to a block constructor explicitly (e.g, as an enum, as a varant of the block type), or by size. A lot of blocks infer type from size because the flowgraph validates connections by matching sizes. This leads to some ambiguity in flowgraph checking, but GRC and blocks can know the exact type.

gr-iio/lib/device_source_impl.cc
```diff
@@ -311,7 +318,8 @@ int device_source_impl::work(int noutput_items,
     unsigned long items = std::min(items_in_buffer, (unsigned long)noutput_items);
 
     for (size_t i = 0; i < output_items.size(); i++)
-        channel_read(channel_list[i], output_items[i], items * sizeof(short));
+        channel_read(
+            channel_list[i], output_items[i], items * channel_size(channel_list[i]));
```
willcode, line 322: Since `channel_size()` calls `iio_channel_get_data_format()` in `work()`, can we assume that the iio call is very low cost?

adisuciu: Yes, it is a simple getter from memory - https://github.com/analogdevicesinc/libiio/blob/main/channel.c#L442-L446 
Alternatively one could get channel size directly from the signature with int gr::io_signature::sizeof_stream_item(int index)

gr-iio/grc/iio_device_sink.block.yml
adisuciu: I couldn't find a list of possible values for types in that can be set in the yml file. Is it possible to add that here ?

gr-iio/lib/device_source_impl.cc
willcode: It's probably fine as is. A couple of pointer indirections per block of samples shouldn't make a difference, and you'd have at least one indirection with a cached vector of sizes.

marcusmueller: @willcode jep, that's why I hoped for `constexpr` (which was not very clever of me!); anyways, the overhead here is that `get_channel_data_format` is a dynamically loaded&linked symbol, so we're actually doing the full call overhead here, which probably is a fair bit more expensive than the pointer derefence involved. It's sensible that libiio has the channel struct as opaque data structure (to avoid people fumbling about in IIO-internal data structures), but unless you can do `constexpr` or `__attribute__((pure))` or `__attribute__((const))` on the actual object. 

(The latter two are supported when built with GCC, so that's why the constexpr worked. I'm *kind of worried* that they actually shouldn't be there, now having slept about that, but way we currently deal with things should be safe, kind of, and it would be an IIO bug, not a gr-iio bug if it weren't: 

https://github.com/analogdevicesinc/libiio/blob/main/include/iio/iio.h#L1498-L1499C34

```C
__api __check_ret __cnst const struct iio_data_format * iio_channel_get_data_format(
		const struct iio_channel *chn);
```

`__cnst` expands to `__attribute__((const))`, but that tells GCC "If you pass in the same pointer (not the identical object pointed to!) twice, you get the same result"; but that's not actually the case, because a pointer to the same address might point to a different object next invocation, so the data format might be different. See [GCC's warning about use with pointers, as well](https://gcc.gnu.org/onlinedocs/gcc/Common-Function-Attributes.html#index-const-function-attribute); will file a bug against libiio)

gr-iio/grc/iio_device_sink.block.yml
willcode: There a couple lists in `grc/core/Constants.py`: `CORE_TYPES` and `PARAM_TYPE_NAMES`. I think the latter is what applies here, and there is no 64-bit int in the list. If there were a 64-bit int, it would have the same size as complex float, too.

Anyway, I was more worried about the ambiguity of `float` and `int`, so you'd remove `float`?

gr-iio/lib/device_source_impl.cc
willcode: I'm going to interpret that as "ok, we'll see how it works".

gr-iio/grc/iio_device_sink.block.yml
adisuciu: Yes, I just pushed .. removed float.

willcode:
@adisuciu @tfcollins I think we'll have to revert this, as it breaks PLUTO.

fmcomms2_source_impl sets the io_signature to the size/type specified in grc, but it also subclasseses device_source, which also sets the io_signature.

device_source after this PR automatically detects the type from the device.

On a pluto, that's complex bytes (size = 2) which does not match the type set in grc. It does not appear that pluto works after the PR.

I think we'll have to revert the PR so we can get a release out.

adisuciu:
It is true that I only tested it with regular iio_device_source, not with the PLUTO blocks. Isn't it possible to override the signature in FMCOMMS2 to complex if needed ?

willcode:
Not sure what you mean.

PR #6832, opened 2023-09-13
blocks: fix repeat block tag propagation

daniestevez:
## Description

The repeat block was refactored as a general block in #6813 (it was a sync_interpolator before). However, tag propagation was missing from the refactor. Manual tag propagation needs to be done if the block is a general block, because the automatic tag propagation preserves the offset of tags, which is not what we want for the repeat block.

Incidentally, before the refactor tag propagation was also somehow broken for interpolation changes at runtime. After an interpolation change, tags would disappear in the output for some time, or appear in the wrong place.

The tag propagation behaviour introduced by this block is the most natural one. Tags from each input item are copied to the first repetition of such input item in the output.

Additionally, two QA tests for tag propagation have been added. One of them does no interpolation changes at runtime, and the other does one interpolation change at runtime. The two tests do not pass with the previous version of the repeat block, but they pass after adding this fix.

Repeat block.

## Testing Done

Added two QA tests that fail with the current `main` but pass with this fix. 

Additionally, I have used the following flowgraph to see the effects of interpolation changes at runtime in the output tags.
[test_repeat.zip](https://github.com/gnuradio/gnuradio/files/12595733/test_repeat.zip)
This flowgraph can be used to see how tag propagation was broken even before #6813.

## Checklist

- [ ] I have updated [the documentation](https://wiki.gnuradio.org/index.php/Main_Page#Documentation) where necessary => *TODO:* The wiki doesn't say anything about how Repeat does propagation. We might want to add a sentence saying that "Tags from each input item are copied to the first repetition of such input item in the output."

### DIFF
--- a/gr-blocks/lib/repeat_impl.h
+++ b/gr-blocks/lib/repeat_impl.h
@@ -1,6 +1,7 @@
 /* -*- c++ -*- */
 /*
  * Copyright 2012 Free Software Foundation, Inc.
+ * Copyright 2023 Daniel Estevez <daniel@destevez.net>
  *
  * This file is part of GNU Radio
  *
@@ -12,6 +13,7 @@
 #define INCLUDED_REPEAT_IMPL_H
 
 #include <gnuradio/blocks/repeat.h>
+#include <cstdint>
 
 namespace gr {
 namespace blocks {
@@ -37,6 +39,19 @@ private:
 
     void msg_set_interpolation(const pmt::pmt_t& msg);
     const pmt::pmt_t c_msg_port;
+
+    //! \brief helper function to propagate tags manually
+    void copy_tags(int consumed, int produced, int num_items)
+    {
+        std::vector<tag_t> tags;
+        const auto start_in = nitems_read(0) + static_cast<uint64_t>(consumed);
+        const auto start_out = nitems_written(0) + static_cast<uint64_t>(produced);
+        get_tags_in_range(tags, 0, start_in, start_in + num_items);
+        for (const auto& tag : tags) {
+            const auto offset = d_interp * (tag.offset - start_in) + start_out;
+            add_item_tag(0, offset, tag.key, tag.value);
+        }
+    }
 };
 
 } /* namespace blocks */
--- a/gr-blocks/python/blocks/qa_repeat.py
+++ b/gr-blocks/python/blocks/qa_repeat.py
@@ -1,6 +1,7 @@
 #!/usr/bin/env python
 #
 # Copyright 2008,2010,2012,2013 Free Software Foundation, Inc.
+# Copyright 2023 Daniel Estevez <daniel@destevez.net>
 #
 # This file is part of GNU Radio
 #
@@ -8,9 +9,11 @@
 #
 #
 
+import time
 
-from gnuradio import gr, gr_unittest, blocks
+from gnuradio import gr, gr_unittest, blocks, pdu
 import numpy as np
+import pmt
 
 
 class test_repeat(gr_unittest.TestCase):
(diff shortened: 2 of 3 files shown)


daniestevez:
An additional thing. I'm wondering about the following comment:

https://github.com/gnuradio/gnuradio/blob/e1bf7314f060816aa158e9a55b235c4defbae6c0/gr-blocks/lib/repeat_impl.cc#L68

The trick there is that the block sends itself a message to process the interpolation change later. Is it true that the `work()` function and a message handler can never run at the same time? I think I've implemented blocks where `work()` blocks waiting for something to happen in a message handler, and this does not result in a deadlock, but perhaps I'm misremembering.

marcusmueller:
Oh dear, I completely forgot about tag prop!

gr-blocks/lib/repeat_impl.h
```diff
@@ -37,6 +39,19 @@ class BLOCKS_API repeat_impl : public repeat
...
+    {
+        std::vector<tag_t> tags;
+        const auto start_in = nitems_read(0) + static_cast<uint64_t>(consumed);
+        const auto start_out = nitems_written(0) + static_cast<uint64_t>(produced);
+        get_tags_in_range(tags, 0, start_in, start_in + num_items);
+        for (const auto& tag : tags) {
+            const auto offset = d_interp * (tag.offset - start_in) + start_out;
+            add_item_tag(0, offset, tag.key, tag.value);
+        }
+    }
```
marcusmueller, line 54: very nice! I'd put it in the `_impl.cc`, but that's more of a question of preference.

gr-blocks/python/blocks/qa_repeat.py
```diff
@@ -34,6 +37,71 @@ def test_001_various_sizes(self):
...
+        self.tb.connect(src, rpt, dst)
+        self.tb.run()
+        tags_out = dst.tags()
+        self.assertEqual(len(tags_in), len(tags_out))
+        for t_in, t_out in zip(tags_in, tags_out):
+            self.assertEqual(t_out.offset, r * t_in.offset)
+            self.assertEqual(t_in.key, t_out.key)
+            self.assertEqual(t_in.value, t_out.value)
+
+    def test_tag_propagation_rate_change(self):
```
marcusmueller, line 60: Excellent idea to test it that way!

marcusmueller (approved):

marcusmueller:
> An additional thing. I'm wondering about the following comment:
> 
> https://github.com/gnuradio/gnuradio/blob/e1bf7314f060816aa158e9a55b235c4defbae6c0/gr-blocks/lib/repeat_impl.cc#L68
> 
> The trick there is that the block sends itself a message to process the interpolation change later. Is it true that the `work()` function and a message handler can never run at the same time? I think I've implemented blocks where `work()` blocks waiting for something to happen in a message handler, and this does not result in a deadlock, but perhaps I'm misremembering.

Yes, that's true: when you look into block_executor, you find a single worker that can only either call general_work, or handle messages.

gr-blocks/lib/repeat_impl.h
daniestevez: My intention in putting it in the `_impl.h` was to have the method inlined in all usages and no "standalone" code for the method. However, this hasn't work as expected. The function doesn't get inlined, and in fact its assembly code is much larger than I expected, so probably it doesn't make much sense inlining it. I have tested what happens if I move the implementation to the `_impl.cc` and declare it `inline`, and the compiler seems to produce the same output. So really no preference from my side here.

daniestevez:
> Maybe you were watching for messages to arrive (and not, them to be handled)?

I think I'm misremembering things. Maybe I was waiting from `work()` with a timeout, just in case stuff arrived (and was handled) to the message handler that I could resume doing processing in `work()` without the need to return to the scheduler. I guess that all those waits would simply timeout, so my solution (if it was like I'm remembering now) was based on a misunderstanding.

PR #6835, opened 2023-09-13
network: improved UDP Source vectorization

### LINKED ISSUE #6826: UDP source block (since 3.9) works weird with "Vector Length" parameter
### What happened?

New UDP Source block (which were added into GR since 3.9) doesn't work with vector length set more than 5. It just doesn't produce any samples at all. In comparison, old UDP Source block (I checked it in GR 3.8.2) worked fine with both vector length set to 1 (stream mode) and with arbitrary vector length.

### System Information

OS: Windows 10
GR Installation Method: radioconda installer

### GNU Radio Version

3.10 (maint-3.10)

### Specific Version

3.10.7.0

To reproduce the problem you need to check attached flowgraphs. They art:
№1 udp_transmit310.grc - flowgraph for sending data to udp 
№2 udp_receive310.grc - flowgraph for receiving data from udp
I tried both transmitting between 2 computers and sending data to localhost, there is no difference, that's on your choice.
№3 udp_receive38.grc - GR 3.8.2 flowgraph for receiving data from udp. You may use it for comparison. There are 2 UDP source blocks there with stream mode and with vector length > 0 and both work fine.

So. You need to set suitable IP's and ports into flowgraphs (at least №1 и №2). Then run №1 and  №2. Try to see data in №2 flowgraph using differen vector length value in UDP Source block. For me, the workable value is 1-4, then block just stops producing anything to its output. (Of course values 2-4 damages the data but at least data is receiving). Why can't I receive entire packet (1472 length in example) into 1 vector with the same length? Idk, but it doesn't work :) 

[udp_troubleshooting.zip](https://github.com/gnuradio/gnuradio/files/12575800/udp_troubleshooting.zip)

### Relevant log output

_No response_

willcode:
## Description
The calculation for set_output_multiple() in did not take vectors into account correctly. Use of vectors is still tricky, as the vector size needs to divide into the payload size evenly. Incorrect values result in corrupted data.

## Related Issue
Fixes #6826

UDP Source

## Testing Done
Sample GRCs provided with issue.

## Checklist

### DIFF
--- a/gr-network/lib/udp_source_impl.cc
+++ b/gr-network/lib/udp_source_impl.cc
@@ -99,10 +99,17 @@ udp_source_impl::udp_source_impl(size_t itemsize,
             "least 8 bytes once header/trailer adjustments are made.");
     }
 
-    d_precomp_data_size = d_payloadsize - d_header_size;
-    d_precomp_data_over_item_size = d_precomp_data_size / d_itemsize;
+    if (d_payloadsize % d_block_size) {
+        d_logger->error("Payload size must be a multiple of item size * vector length.");
 
-    int out_multiple = (d_payloadsize - d_header_size) / d_block_size;
+        throw std::invalid_argument(
+            "Payload size must be a multiple of item size * vector length.");
+    }
+
+    d_precomp_data_size = d_payloadsize - d_header_size;
+    d_precomp_data_over_item_size = d_precomp_data_size / d_block_size;
+
+    int out_multiple = d_precomp_data_over_item_size;
 
     if (out_multiple == 1)
         out_multiple = 2; // Ensure we get pairs, for instance complex -> ichar pairs


daniestevez:
> Use of vectors is still tricky, as the vector size needs to divide into the payload size evenly. Incorrect values result in corrupted data.

Does this mean that `payloadsize % (itemsize * veclen) == 0` must be true for the block to work correctly? If so, could we add a check in the constructor that throws an exception if this is not the case?

willcode:
Good idea, I'll add that. There's an additional constraint that the logic assumes the stream may be ichar or ishort and forces a minimum output multiple of 2. So it's really
`payloadsize % (itemsize * veclen * 2) == 0`
which gets a little hard to explain in an error message.

PR #6852, opened 2023-09-15
Add Black formatter and run over all grc code

cafeclimber:
## Description
Right now, there's no standard code formatting for Python code. Black is a very popular formatter with good defaults, and I think introducing it to the `grc` module is a good first step to establishing it as the standard.

## Related Issue
None

grc

## Testing Done
No new testing added. All previously passed unit tests still pass

## Checklist


willcode:
We do have pep8 formatting (tools/autoformat_python.sh) and checking (CI).

willcode:
It looks like black mostly replaced single quotes with double quotes. This doesn't look worth touching all these files at this point.

cafeclimber:
It's more extensive than autopep8 to my understanding, but totally fine with sticking with autopep8. I do still think it would be valuable to at least introduce a pyproject.yaml to the grc repo to add configuration for autopep8 that could be picked up by folks editors instead of having a script for folks to run :)

PR #6853, opened 2023-09-18
digital: add sps setter to symbol_sync

daniestevez:
## Description

This adds a method to update the samples/symbol of the symbol_sync block at runtime. When the set_sps() method is called, it resets all the internal symbol synchronization loop status and ajusts values for the new samples/symbol, so synchronization is lost and the loop needs to lock again. This is reasonable, because the set_sps() method can only be called asynchronously with respect to symbol rate changes in the input sample stream. A more advanced runtime change could use tags to update the samples/symbol at a specific point without losing loop lock, but this commit does not cover this use case.

## Related Issue

No related issue.

The Symbol Sync block.

## Testing Done

Testing has been done with the following flowgraph, which generates a square-pulse BPSK signal with an integer sps, runs Symbol Sync on it, and allows runtime changes of sps with an QT GUI entry.
[test_symbol_sync.zip](https://github.com/gnuradio/gnuradio/files/12646107/test_symbol_sync.zip)

## Checklist

- ~I have updated [the documentation](https://wiki.gnuradio.org/index.php/Main_Page#Documentation) where necessary.~ -> No documentation updates are probably needed. The user-facing consequences of this change is that now the Samples per Symbol parameter is underlined in GRC, so in a sense, the change is self-documenting.
- [ ] I have added tests to cover my changes, and all previous tests pass. -> I haven't added any unit tests for this. It seems tricky to write a good unit test that performs an sps change at runtime.

### DIFF
--- a/gr-digital/lib/symbol_sync_cc_impl.cc
+++ b/gr-digital/lib/symbol_sync_cc_impl.cc
@@ -1,6 +1,7 @@
 /* -*- c++ -*- */
 /*
  * Copyright (C) 2017 Free Software Foundation, Inc.
+ * Copyright (C) 2023 Daniel Estevez <daniel@destevez.net>
  *
  * This file is part of GNU Radio
  *
@@ -72,6 +73,8 @@ symbol_sync_cc_impl::symbol_sync_cc_impl(enum ted_type detector_type,
       d_inst_output_period(sps / static_cast<float>(osps)),
       d_inst_clock_period(sps),
       d_avg_clock_period(sps),
+      d_sps(sps),
+      d_max_deviation(max_deviation),
       d_osps(static_cast<float>(osps)),
       d_osps_n(osps),
       d_time_est_key(pmt::intern("time_est")),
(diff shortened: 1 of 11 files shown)


willcode (approved):

gr-digital/lib/symbol_sync_cc_impl.cc
```diff
@@ -132,6 +129,48 @@ symbol_sync_cc_impl::symbol_sync_cc_impl(enum ted_type detector_type,
...
+
+    d_inst_output_period = sps / d_osps;
+    d_inst_clock_period = sps;
+    d_avg_clock_period = sps;
+
+    sync_reset_internal_clocks();
+
+    d_inst_interp_period = d_inst_clock_period / d_interps_per_symbol;
+
+    check_interps();
```
mbr0wn, line 155: Do we want this in the regular ctor, too?

willcode: Line 114 ... or are you saying it shouldn't be in the constructor?

daniestevez: The code that is now in `check_interps()` (which didn't exist before) was originally only in the constructor. I wanted to do the same checks in the constructor and in the `set_sps()` method, so I extracted them to a common function.

PR #6858, opened 2023-09-21
gr-uhd: add a GRC block for the RFNoC NullSrcSink block

zacharyr:
## Description

gr-uhd is missing a GRC block for the RFNoC block NullSrcSink. For context, this RFNoC block provides a null source/sink and loopback capability. This addition adds a GRC block to allow the use of this RFNoC block in GRC. In the options for the block, you can toggle Source or Sink to change the associated use of the block.

This addition also adds an example flowgraph as seen below:

Screenshot 2023-09-20 at 4 47 32 PM

This work was performed here at the NASA Jet Propulsion Laboratory, I am required to post the following:

Affiliation: Jet Propulsion Laboratory, California Institute of Technology

Contact Information: Zach Rohde <zachary.s.rohde@jpl.nasa.gov>

Funding Acknowledgment: The research was carried out at the Jet Propulsion Laboratory, California Institute of Technology, under a contract with the National Aeronautics and Space Administration (80NM0018D0004).

* gr-uhd/grc/CMakeLists.txt
* gr-uhd/examples/grc/CMakeLists.txt

## Testing Done

Tested using an RFNoC configuration with the attached flowgraph on a USRP X310.

## Checklist

### DIFF
--- a/gr-uhd/grc/uhd_rfnoc_null_src_sink.block.yml
+++ b/gr-uhd/grc/uhd_rfnoc_null_src_sink.block.yml
@@ -1,0 +1,54 @@
+id: uhd_rfnoc_null_src_sink
+label: RFNoC Null Source Sink Block
+category: '[Core]/UHD/RFNoC/Blocks'
+
+parameters:
+- id: btype
+  label: Block Type
+  dtype: enum
+  options: ['sink', 'source']
+  option_labels: ['Sink', 'Source']
+  default: 'sink'
+  hide: part
+- id: block_args
+  label: Block Args
+  dtype: string
+  default: ""
+  hide: ${ ( 'part' if not block_args else 'none' ) }
+- id: device_select
+  label: Device Select
+  dtype: int
+  default: -1
+  hide: ${ ( 'part' if device_select == -1 else 'none' ) }
+- id: instance_index
+  label: Instance Select
+  dtype: int
+  default: -1
+  hide: ${ ( 'part' if instance_index == -1 else 'none' ) }
+
+inputs:
+- domain: rfnoc
+  dtype: 'sc16'
+  hide: ${ ( True if str(btype) == 'source' else False ) }
+
+outputs:
+- domain: rfnoc
+  dtype: 'sc16'
+  hide: ${ ( True if str(btype) == 'sink' else False ) }
+
+templates:
+  imports: |-
+    from gnuradio import uhd
+  make: |-
+    uhd.rfnoc_block_generic(
+        self.rfnoc_graph,
+        uhd.device_addr(${block_args}),
+        "NullSrcSink",
+        ${device_select},
+        ${instance_index})
+
+documentation: |-
+  RFNoC Null Source Sink Block:
+  Send data to a "bit bucket" sink or receive data from a source.
+
+file_format: 1
--- a/gr-uhd/grc/uhd_rfnoc_null_src_sink.block.yml
+++ b/gr-uhd/grc/uhd_rfnoc_null_src_sink.block.yml
@@ -7,7 +7,7 @@ parameters:
   label: Block Type
   dtype: enum
   options: ['sink', 'source', 'both']
-  option_labels: ['Sink', 'Source']
+  option_labels: ['Sink', 'Source', 'Both']
   default: 'sink'
   hide: part
 - id: block_args
--- a/gr-uhd/grc/uhd_rfnoc_null_src_sink.block.yml
+++ b/gr-uhd/grc/uhd_rfnoc_null_src_sink.block.yml
@@ -6,7 +6,7 @@ parameters:
 - id: btype
   label: Block Type
   dtype: enum
-  options: ['sink', 'source']
+  options: ['sink', 'source', 'both']
   option_labels: ['Sink', 'Source']
   default: 'sink'
   hide: part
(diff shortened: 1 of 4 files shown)


gr-uhd/grc/uhd_rfnoc_null_src_sink.block.yml
```diff
@@ -0,0 +1,54 @@
+id: uhd_rfnoc_null_src_sink
+label: RFNoC Null Source Sink Block
+category: '[Core]/UHD/RFNoC/Blocks'
+
+parameters:
+- id: btype
+  label: Block Type
+  dtype: enum
+  options: ['sink', 'source']
```
mbr0wn, line 9: Can you add a sink&source mode please?

```suggestion
  options: ['sink', 'source', 'both']
```

The reason being: If you wanted to use the same block for sink & source (which is totally fine from an RFNoC perspective) we currently have an issue where you can only access a block so many times from GNU Radio before gr-uhd becomes suspicious. That's something we need to improve elsewhere, but why not simply make this block accessible from both ends at once.

gr-uhd/grc/uhd_rfnoc_null_src_sink.block.yml
```diff
@@ -0,0 +1,54 @@
...
+
+inputs:
+- domain: rfnoc
+  dtype: 'sc16'
+  hide: ${ ( True if str(btype) == 'source' else False ) }
+
+outputs:
+- domain: rfnoc
+  dtype: 'sc16'
+  hide: ${ ( True if str(btype) == 'sink' else False ) }
```
mbr0wn, line 37: If you change the above, this would become something like

```suggestion
  hide: ${ ( True if str(btype) in ('sink', 'both') else False ) } 
```

mbr0wn:
@zacharyr Thanks for adding this! I can think of some minor improvements, but there's nothing fundamentally wrong with this, and I would approve this (modulo that one small change requests I suggested).

gr-uhd/grc/uhd_rfnoc_null_src_sink.block.yml
zacharyr: No problem at all! I have pushed the change.

zacharyr:
@mbr0wn Please re-review when you can, I have made the changes.

willcode:
The open labels need to be updated too.

willcode:
As does the input/output logic. I don't have anything with RFNoC, so can't test.

zacharyr:
I updated the Option Label to include the "both" option. The input/output logic did not need updating as the if-statement works for the "both" case already.

I ran an additional test for this new option, although I had to slightly modify my example to get it to work given our RFNoC configuration on our X310. As you can see, it is null sinking the receiver input, then the null source is correctly being output. For context: the RFNoC null source does not output zeros, it outputs some counter value (I believe).

Screenshot 2023-09-21 at 4 31 35 PM
Screenshot 2023-09-21 at 4 30 51 PM

marcusmueller:
> For context: the RFNoC null source does not output zeros, it outputs some counter value (I believe).

Uh? I'd call that surprising, to say the least. But it seems to be true:

```verilog
  assign src_pyld_tdata  = {NIPC{{~src_line_cnt[ITEM_W/2-1:0], src_line_cnt[ITEM_W/2-1:0]}}};
```

and that `src_line_cnt` is being counted up every `rfnoc_chdr_clk` as long as tvalid&&tready. (which doesn't explain the "jumps" in the waveform. huh!)

If that's the case, is this block simply misnamed? I somewhat understand the implications of renaming anything in the RFNoC verilog code base, but how do we save GRC users from thinking this block outputs zeros?

zacharyr:
Yeah it surprised me too when I first turned on the RFNoC null source a while back, I was expecting zeros like the GR version.

I also cannot explain the jumps. The USRP was not overflowing ("O") or dropping packets ("D") as far as I could tell, which usually explains these kind of weird outputs.

marcusmueller:
@zacharyr the behaviour of the block certainly is out of scope of what you can address in this PR :) So, yeah, we should probably document the hell out of that!

I put up a stub on 
https://wiki.gnuradio.org/index.php?title=RFNoC_Null_Source_Sink_Block . @zacharyr would it be OK if I asked you to fill that with life later on?

zacharyr:
No problem at all @marcusmueller. I’ll update both of the documentation stubs.

mbr0wn:
Yeah the output not being zeros has surprised people in the past. FYI the null source/sink was the first ever RFNoC block, written by the Ettus himself :)

FWIW counters are much more useful than zeros, given you have to carry them all the way from the device out of the streamers in to whatever comes next. I'm not sure where those jumps are coming from, but we can have a pretty good guess at how many samples were dropped, if we assume that it's less than one period. For that reason, we won't change this in UHD. The name has been around for almost 10 years now, too, so we're probably not going to rename it (to... null sink/counter source? null sink/ramp source?). But the docs should reflect this, I agree.

mbr0wn:
BTW I also don't know where those jumps come from. Someone must be droping samples. If it's not the transport/streamer... then who is? Hmmmm....

mbr0wn (commented):
@willcode haha I was just reviewing this as you merged :) Serendipity? Synchronicity? Anyway this looks good. Like I said, I can think of ways to improve it, but that doesn't mean this PR isn't good by itself. Thanks @zacharyr!

PR #6859, opened 2023-09-21
docs: Update RTD config to include build.tools

mbr0wn:
This fixes the following RTD error message:

    Problem in your project's configuration. Invalid configuration option
    "build.tools": At least one item should be provided in 'tools' or
    'commands'

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -8,8 +8,8 @@ version: 2
 # Set the version of Python and other tools you might need
 build:
   os: ubuntu-22.04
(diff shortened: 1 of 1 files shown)


willcode:
@mbr0wn Nothing to CI here, so ok to merge now?

mbr0wn:
> @mbr0wn Nothing to CI here, so ok to merge now?

Please do -- I was tempted to merge straight to `main`, but my consciousness and muscle memory both fought me on that and won :grin:

PR #6862, opened 2023-09-22
Msg Pair To Var: fix attribute error during callback error handling

eltos:
Fixes `Attribute Error: function has no attribute "name"`

### DIFF
--- a/gr-blocks/python/blocks/msg_pair_to_var.py
+++ b/gr-blocks/python/blocks/msg_pair_to_var.py
@@ -39,7 +39,7 @@ class msg_pair_to_var(gr.sync_block):
         try:
             self.callback(new_val)
         except Exception as e:
-            gr.log.error("Error when calling " + repr(self.callback.name()) + " with " +
+            gr.log.error("Error when calling " + repr(self.callback) + " with " +
                          repr(new_val) + " (reason: %s)" % repr(e))
 
     def stop(self):


gr-blocks/python/blocks/msg_pair_to_var.py
```diff
@@ -39,7 +39,7 @@ def msg_handler(self, msg):
         try:
             self.callback(new_val)
         except Exception as e:
-            gr.log.error("Error when calling " + repr(self.callback.name()) + " with " +
+            gr.log.error("Error when calling " + repr(self.callback) + " with " +
```
willcode, line 42: That would print something like `<function f at 0x7fbd745ec040>` for the function name. Would
```
gr.log.error("Error when calling " + self.callback.__name__ + " with " +
```
(with no repr) work?

eltos: I would strongly advise against making any assumptions about what `callback` is, given the fact that this is an error handler where something about `callback` is not as expected. What if for whatever reason `callback` is `None`? Indeed, Attribute error...

Also, in my opinion, the important aspect about error messages is to provide helpful information to identify the cause, not to be pretty. The above provides any reasonable information, e.g. about the class and module (something like `bound method Classname.set_my_var of <my_module.Classname object at 0x7f37369b249` for example) instead of only the plain function name.

willcode:
Alright, safe is good.

Issue #6863, opened 2023-09-23
Improved ZMQ Teardown 


SaikWolf:
### Feature Description

For more complex usage, the ZMQ sources/sinks can hang when trying to shutdown/stop the flowgraphs they're in.

Making sure the destructor properly terminates resources should help fix that.

(Most notable within Python usage)

### Feature Urgency

medium (would be nice to have in the near future)

### More Information

A quick patch for my own use is here: https://github.com/SaikWolf/gnuradio/pull/2

I'm sure I'm missing things needed for full integration, but if anyone finds it useful as a starting point.

willcode:
That looks good. Do you want to submit that against `main`?

SaikWolf:
Anywhere that makes it simple to integrate would be fine.

willcode:
You can either submit a PR against our `main`, or I can pull this from your repo.

vmw:
This commit results in a compilation error when running under Oracle 8. This compilation failure is reproducible using the following Dockerfile:

https://github.com/pervices/pvpkg/blob/f4c6410/Oracle/Oracle8/GnuRadio/Dockerfile

We specifically get the following compilation error:

---

[ 31%] Building CXX object gr-zeromq/lib/CMakeFiles/gnuradio-zeromq.dir/base_impl.cc.o
[...]

PR #6865, opened 2023-09-25
Tweaking gr-zmq to be more friendly to our use cases.

SaikWolf:
zmq: Improving functionality when needing to interrupt sockets

## Description

Mostly housekeeping with ZMQ sockets to better use them in more complex systems.
This allows for better handling in python when sockets need to be interrupted and properly close to prevent hanging.

## Related Issue

Addresses #6863 

ZMQ sources/sinks.

## Testing Done

Ubuntu 20/22.04 improved functionality within python usage.

## Checklist

### DIFF
--- a/gr-zeromq/lib/base_impl.cc
+++ b/gr-zeromq/lib/base_impl.cc
@@ -85,7 +85,12 @@ base_impl::base_impl(int type,
     }
 }
 
-base_impl::~base_impl() {}
+base_impl::~base_impl()
+{
+    d_context.shutdown();
+    d_socket.close();
+    d_context.close();
+}
 
 std::string base_impl::last_endpoint()
 {
--- a/gr-zeromq/lib/rep_msg_sink_impl.cc
+++ b/gr-zeromq/lib/rep_msg_sink_impl.cc
@@ -58,7 +58,12 @@ rep_msg_sink_impl::rep_msg_sink_impl(char* address, int timeout, bool bind)
     message_port_register_in(d_port);
 }
 
-rep_msg_sink_impl::~rep_msg_sink_impl() {}
+rep_msg_sink_impl::~rep_msg_sink_impl()
+{
+    d_context.shutdown();
+    d_socket.close();
+    d_context.close();
+}
 
 bool rep_msg_sink_impl::start()
 {
@@ -79,7 +84,7 @@ void rep_msg_sink_impl::readloop()
     while (!d_finished) {
 
         // while we have data, wait for query...
-        while (!empty_p(d_port)) {
+        while (!empty_p(d_port) && !d_finished) {
 
             // wait for query...
             zmq::pollitem_t items[] = {
--- a/gr-zeromq/lib/pub_msg_sink_impl.cc
+++ b/gr-zeromq/lib/pub_msg_sink_impl.cc
@@ -55,7 +55,12 @@ pub_msg_sink_impl::pub_msg_sink_impl(char* address, int timeout, bool bind)
     set_msg_handler(pmt::mp("in"), [this](pmt::pmt_t msg) { this->handler(msg); });
 }
 
-pub_msg_sink_impl::~pub_msg_sink_impl() {}
+pub_msg_sink_impl::~pub_msg_sink_impl()
+{
+    d_context.shutdown();
+    d_socket.close();
+    d_context.close();
+}
 
 void pub_msg_sink_impl::handler(pmt::pmt_t msg)
 {
--- a/gr-zeromq/lib/pull_msg_source_impl.cc
+++ b/gr-zeromq/lib/pull_msg_source_impl.cc
@@ -59,7 +59,12 @@ pull_msg_source_impl::pull_msg_source_impl(char* address, int timeout, bool bind
     message_port_register_out(d_port);
 }
 
-pull_msg_source_impl::~pull_msg_source_impl() {}
+pull_msg_source_impl::~pull_msg_source_impl()
+{
+    d_context.shutdown();
+    d_socket.close();
+    d_context.close();
+}
 
 bool pull_msg_source_impl::start()
 {
--- a/gr-zeromq/lib/push_msg_sink_impl.cc
+++ b/gr-zeromq/lib/push_msg_sink_impl.cc
@@ -56,7 +56,12 @@ push_msg_sink_impl::push_msg_sink_impl(char* address, int timeout, bool bind)
     set_msg_handler(pmt::mp("in"), [this](pmt::pmt_t msg) { this->handler(msg); });
 }
 
-push_msg_sink_impl::~push_msg_sink_impl() {}
+push_msg_sink_impl::~push_msg_sink_impl()
+{
+    d_context.shutdown();
+    d_socket.close();
+    d_context.close();
+}
 
 void push_msg_sink_impl::handler(pmt::pmt_t msg)
 {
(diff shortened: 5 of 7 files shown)


willcode:
I should have mentioned that you can run
`clang-format -i <filename>`
in the tree on any modified files, and the results will pass CI.

Please squash to a single commit and sign it (`git commit --amend --signoff`) to satisfy the DCO.

willcode:
@SaikWolf Could add a signoff to the commit? I'd like to merge this one now so we can put it in the next release. You can use the command above, or just manually add a line to the end:
`Signed-off-by: Bill Clark <saikou@vt.edu>`
or whatever email address you are using now (I pulled that one from the logs).

gr-zeromq/lib/base_impl.cc
```diff
@@ -85,7 +85,12 @@ base_impl::base_impl(int type,
     }
 }
 
-base_impl::~base_impl() {}
+base_impl::~base_impl()
+{
+    d_context.shutdown();
+    d_socket.close();
+    d_context.close();
```
daniestevez, line 92: The `zmq::socket_t` destructor [already calls `close()`](https://github.com/zeromq/cppzmq/blob/160ac8ec61298508093933a9f53bfedfb6ba0425/zmq.hpp#L2225), and likewise the `zmq::context_t` destructor [already calls `close()`](https://github.com/zeromq/cppzmq/blob/160ac8ec61298508093933a9f53bfedfb6ba0425/zmq.hpp#L832). A double `close()` isn't harmful because on the first `close()` an internal pointer is set to `nullptr`, so the second `close()` does nothing. But I would say that implicitly calling `close()` with the destructors is better RAII style.

The call to `shutdown()` is interesting. It basically calls [`zmq_ctx_shutdown()`](http://api.zeromq.org/master:zmq-ctx-shutdown), which says "Context shutdown will cause any blocking operations currently in progress on sockets open within context to return immediately with an error code of ETERM." This seems to be something we'd like, because if we have a ZMQ block blocked waiting for data, we'd want to be able to terminate the flowgraph normally (I currently have some use cases in which the flowgraph needs to be SIGKILL'ed because of this). However, I wonder if the `shutdown()` here really achieves what is intended. If a call to `zmq::poll()` (or whatever other function) is blocked in the `work()` function, can the destructor actually start running?

gr-zeromq/lib/rep_msg_sink_impl.cc
```diff
@@ -79,7 +84,7 @@ void rep_msg_sink_impl::readloop()
     while (!d_finished) {
 
         // while we have data, wait for query...
-        while (!empty_p(d_port)) {
+        while (!empty_p(d_port) && !d_finished) {
```
daniestevez, line 87: Why is checking `d_finished` necessary here? Some lines below there is
```c++
               if (!ok) {
                    // Should not happen, we've checked POLLIN.
                    d_logger->error("Failed to receive message.");
                    break; // Fall back to re-check d_finished
                }
```
This `break` goes to the `d_finished` check in the outermost `while` loop. Are there other cases in which `d_finished` can become `true` that are not handled by this code?

I find the logic in `readloop()` slightly difficult to follow, so this is why I wonder what are exactly the changes achieved by introducing another `d_finished` check here.

daniestevez (commented):
I've tried to understand in more detail what this PR does, since the commit description and PR description don't really say. I've left a couple comments with specific questions on the code.

daniestevez:
By the way. Here is a minimal example of the kind of flowgraphs that I mentioned that need to be SIGKILL'ed.
test_zmq_req_stop
[test_zmq_req_stop.grc.zip](https://github.com/gnuradio/gnuradio/files/12747312/test_zmq_req_stop.grc.zip)

I'm using a timeout of -1 because I need to have minimal latency in this application (this is the input of a transmitter, which asynchronously gets frames from an external application as a tagged stream using REQ/REP). If I allow the `work()` function of the ZMQ REQ Source to return when there are no ZMQ messages (by setting a timeout), then then next ZMQ message that arrives will need to wait for the runtime to call `work()`, which can take 10s of ms to happen, and so introduces too much latency.

It would be great if this PR allows this flowgraph to terminate normally, but I haven't tested it yet.

gr-zeromq/lib/rep_msg_sink_impl.cc
willcode: It looks like it bypasses the poll, terminating d_timeout ms faster.

gr-zeromq/lib/base_impl.cc
willcode: That's a good point. It was the shutdown I thought was the most interesting. I wonder if the shutdown belongs in `stop()` instead of in the destructor right before the `join()`. @SaikWolf any thoughts on this? Did you find that the shutdown had an effect after the flowgraph was already stopped?

SaikWolf: My reason for it in the destructor over the stop:
 - There might be cases where the flowgraph has been stopped, but any processing messages might still be desired.
 - If the object is being destroyed, it's clear that anything in process can be dropped.

gr-zeromq/lib/rep_msg_sink_impl.cc
SaikWolf: The other reason is 
- if the producer of the messages is sending them faster than block can pull them, there's a chance that the socket/context have been closed before we check if we're finished.
- This condition will cause errors on exit that having the check if finished there will naturally handle.

gr-zeromq/lib/base_impl.cc
SaikWolf: The reason for the explicit close calls is that the context/socket need to be closed in that order. Sometimes I've seen them closed in the wrong order causing a hang. This just makes sure everything wraps up as ZMQ likes.

daniestevez: Given that the class members are declared in the order
```c++
    zmq::context_t d_context;
    zmq::socket_t d_socket;
```
(which makes sense, because to construct `d_socket` we need `d_context), the destructors will run in the reverse order: first for the socket, then for the context, which is the correct order for the `close()`'s. What are the cases in which you observed the wrong order?

SaikWolf: It's usually been in python operations. I could be remembering wrong, and this is overkill.

gr-zeromq/lib/rep_msg_sink_impl.cc
daniestevez: > It looks like it bypasses the poll, terminating d_timeout ms faster.

Okay. That makes sense on a second read.

> there's a chance that the socket/context have been closed before we check if we're finished

How? We only close the socket and context when we destroy the class. This thread is joined in `stop()`. How can the destructor run before the thread finishes?

SaikWolf: Sorry I'm confusing problems.

The problem is that the call to close can hang under the conditions of:
- d_port has more messages
- stop() has been called
- the state of the block is waiting for POLLIN to be set, but the sender is already done.

Under these conditions the inner loop is infinite. Checking d_finished there prevents that's.

daniestevez: That explanation makes total sense. Thanks. Consider this solved (I think I can't mark the thread as solved myself).

PR #3163, opened 2020-02-02
RangeWidget: implement Eng & EngSlider

Sec42:
This is a small addition to the RangeWidget. An input field `Eng` that accepts numbers in Engineering notation (as used by `gnuradio.eng_notation`) and a combo `EngSlider` that is similar to `CounterSlider` but constructed out of `Eng` and `Slider` instead of `Counter` and `Slider`,

PR #6867, opened 2023-09-25
qtgui: Refine use of `QLineEdit::returnPressed()` or `editingFinished()` for text entry

ryanvolz:
## Description

This is a follow-up to #6645 which addresses comments that have arisen since that was merged. Namely:

1. Using `returnPressed` for the Message Edit Box is the preferred behavior, so reverts the prior change and restores that behavior as intended with #5592.
2. Since there are reasonable use cases for either signal with the QT GUI Entry, this provides a GRC option that allows the user to select the update trigger to be either `returnPressed` or `editingFinished`.
3. The range widget's use of `editingFinished` only arises with the `eng` and `eng_slider` variants, and for those the `editingFinished` behavior makes sense. However, in reviewing this, I discovered that neither of those variants has been exposed in GRC since they were introduced in #3163, so this PR includes a commit to make those available.

qtgui

## Testing Done

I verified the new option for QT GUI Entry in GRC does generate the desired code.

## Checklist

### DIFF
--- a/gr-qtgui/grc/qtgui_entry.block.yml
+++ b/gr-qtgui/grc/qtgui_entry.block.yml
@@ -21,6 +21,13 @@ parameters:
     label: Default Value
     dtype: ${ type }
     default: '0'
+-   id: entry_signal
+    label: Update Trigger
+    dtype: enum
+    default: editingFinished
+    options: [returnPressed, editingFinished]
+    option_labels: ["'Enter'", "'Enter' or focus lost"]
+    hide: part
 -   id: gui_hint
     label: GUI Hint
     dtype: gui_hint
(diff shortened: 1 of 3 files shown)


willcode:
Nice, I like the solution. I'll give it a try.

gr-qtgui/grc/qtgui_entry.block.yml
```diff
@@ -21,6 +21,13 @@ parameters:
     label: Default Value
     dtype: ${ type }
     default: '0'
+-   id: entry_signal
+    label: Update trigger
+    dtype: enum
+    default: editingFinished
+    options: [returnPressed, editingFinished]
+    option_labels: ["'Enter' is pressed", "'Enter' is pressed / line edit loses focus"]
```
willcode, line 29: Maybe "Enter", "Enter or focus lost" to make it a little shorter.

Nit: "Update Trigger" (caps) to match other param names.

Issue #2687, opened 2019-07-30
GRC color scheme help


na0b:
I am running GRC on a Live Linux Parrot OS distribution. Many of the text colors are messed up. When editing a block, sometimes the text and the background are of the same color, so you cannot see what you are writing. While I am able to change colors in other applications, nothing seems to work for GRC.
Can someone point me in the right direction?
Thanks,
Juan.

mormj:
Would you be able to post a screenshot to give a better idea what you are seeing?

na0b:
OK.Here is a screen capture. In the ID field of the Variable block, both the text and the background are set to the same color. Thanks.

![grccolorscheme](https://user-images.githubusercontent.com/4021714/62180574-243ada80-b316-11e9-953e-f22b18bdf53f.png)

mormj:
Hmmm .. does changing your color scheme in your linux distro have any effect?

na0b:
This is not the case  with the Parrot distro. I  also posted a question at the Parrot community, but received no answer. I will go back to Ubuntu. Thanks.
What I think would be very useful to schools is a lean live GNURadio distro with networking. If you  know of any, please let me know.

cbdonohue:
You can change the theme in Parrot OS by going to System -> Look and Feel -> Appearance. Then try a theme like "TraditionalOK" and you should be able to see everything in gnuradio. 
![Screenshot from 2019-10-31 14-29-47](https://user-images.githubusercontent.com/1757340/67975790-264d5780-fbeb-11e9-8d2e-020f1b33cac4.png)

nlitsme:
I noticed that I needed to change my OS color scheme from dark to standard, and then back to dark.
Now the foreground text color in GRC changed from white ( unreadable ) to black ( ok ).
While also retaining the general scheme I like for the rest of my apps.

But since GRC is the first app I found which seems to make a wrong choice by default for the text color, I think the problem may be with GRC.

I am running a kde/plasma desktop on debian.

willem

cafeclimber:
I honestly think that's clearer in the long term at least for the GTK version. I personally never even made the connection between the colors and the parameter's type 😅 For Qt, I know you can have different validators on the various fields

marcusmueller:
connections between colors and types: me neither, the pastel goth version of the in/output port colors just doesn't strike me as too significant, and also, there's no "string" input port, and reserving anything reddish for "invalid" seems just right…

marcusmueller:
[...]

PR #6888, opened 2023-10-03
grc: Remove param field colorings by type and replace with labels

cafeclimber:
## Description
Many folks have complained about the garish color scheme used to denote types of parameters in the block parameters dialog. The meaning is not especially clear, and the colors do not adapt to different color schemes well. Instead, this PR adds the field's datatype as a label at the end of the parameter (see below) and removes coloring.

(apologies for the change count...I will try to remember to turn off Black when working on PRs in the future 😖)

grc_params
grc_params_dark

## Related Issue
#2687 

GRC

## Testing Done
pytest was run in `grc/` and all tests passed

## Checklist

### DIFF
--- a/grc/gui/ParamWidgets.py
+++ b/grc/gui/ParamWidgets.py
@@ -25,29 +25,22 @@ def have_dark_theme():
         """
         Check if a theme is dark based on its name.
         """
-        return theme_name and (
-            theme_name in Constants.GTK_DARK_THEMES or "dark" in theme_name.lower()
-        )
+        return theme_name and (theme_name in Constants.GTK_DARK_THEMES or "dark" in theme_name.lower())
 
     # GoGoGo
     config = configparser.ConfigParser()
     config.read(os.path.expanduser(Constants.GTK_SETTINGS_INI_PATH))
     prefer_dark = config.get(
-        "Settings", Constants.GTK_INI_PREFER_DARK_KEY, fallback=None
-    )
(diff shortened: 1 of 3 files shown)


willcode:
I like this idea. I always have to hover over the field to find out what type the color means anyway.

Could you remove the unnecessary changes?

cafeclimber:
K fixed those changes. I do want to revisit an switching to Black if possible. There's so many odd inconsistencies both across and within files. If not that, then adding a pyproject.toml to define autopep8 settings to be consistent across the project.

marcusmueller:
the code is clean (ok, there's still four empty line fixes), and it seems all good to me.

willcode:
Type field should be suppressed for
- `enum` - no point in telling the user that
- `id` - not a type
- `file_save` - not a type
- see if there are any others like this?

Vector types could have the `_` removed. That's the internal type and not relevant for the user.

cafeclimber:
I think I saw `_multiline_python` as well. And this is why I added the ignored types list :D simple fix. 

I could also make a dictionary of type-label if we think that would be better the `int`, `list` etc

cafeclimber:
Ack! I hadn't pushed the commit with the updated label ignore list. Was waiting to see folks thoughts on a dict of labels by the vs just a list of ignored types.

Should I just push the latest code to this branch?

willcode:
Sorry! Just make another PR with the updated list. Thanks.

PR #6889, opened 2023-10-04
GRC: Enable edge/connection properties

mbr0wn:
This is a bunch of modifications that would allow creating edge/connection properties. These properties are not known by the runtime (4.0 style), only by GRC for rendering.

I think it's most obvious what I'm trying to do if you take a look at `rfnoc.domain.yml`.

How to try:
- Install this branch
- Open an RFNoC flow graph (e.g., rfnoc_radio_loopback.grc). Then right-click on one of the green arrows. You will be able to set a Boolean property of this connection.
- When you render the flow graph, it'll have access to the params.

## Related Issue
For RFNoC, I would like the ability to pass arguments to the `connect()` call.

Technically, anything GRC. However, the changes are currently localized to any RFNoC GRC files. Old files get saved and interpreted as before. If someone edits a .grc file that doesn't have edge params, the file gets stored in a backward-compatible way, just like before. Only when people start using this feature, will the .grc file become no longer backward compatible. The flow graph version is bumped to 2.

## Testing Done

Looksy!

https://github.com/gnuradio/gnuradio/assets/508035/919906fd-b08c-4f28-93a8-015e6498a50f

The generated GRC is also correct.

Also tried opening/modifying/saving/closing/reopening unrelated GRC files.

## Checklist

### DIFF
--- a/gr-uhd/grc/rfnoc.domain.yml
+++ b/gr-uhd/grc/rfnoc.domain.yml
@@ -7,4 +7,10 @@ multiple_connections_per_output: false
 
 templates:
 -   type: [rfnoc, rfnoc]
-    connect: self.rfnoc_graph.connect(self.${ source.parent_block.name }.get_unique_id(), ${ source.key }, self.${ sink.parent_block.name }.get_unique_id(), ${ sink.key }, False)
+    connect: self.rfnoc_graph.connect(self.${ source.parent_block.name }.get_unique_id(), ${ source.key }, self.${ sink.parent_block.name }.get_unique_id(), ${ sink.key }, ${ is_back_edge })
+    parameters:
+        - id: is_back_edge
+          label: 'Back-edge'
+          dtype: bool
+          options: ['True', 'False']
+          default: 'False'
--- a/grc/core/Connection.py
+++ b/grc/core/Connection.py
@@ -6,6 +6,7 @@ SPDX-License-Identifier: GPL-2.0-or-later
 
 """
 
+import collections
 
 from .base import Element
 from .Constants import ALIASES_OF
@@ -13,15 +14,23 @@ from .utils.descriptors import lazy_property
 
 
 class Connection(Element):
+    """
+    Stores information about a connection between two block ports. This class
+    knows:
+    - Where the source and sink ports are (on which blocks)
+    - The domain (message, stream, ...)
+    - Which parameters are associated with this connection
+    """
 
     is_connection = True
+    documentation = {'': ''}
 
     def __init__(self, parent, source, sink):
         """
         Make a new connection given the parent and 2 ports.
 
         Args:
-            flow_graph: the parent of this element
+            parent: the parent of this element (a flow graph)
             source: a port (any direction)
             sink: a port (any direction)
         @throws Error cannot make connection
@@ -41,6 +50,16 @@ class Connection(Element):
         self.source_port = source
         self.sink_port = sink
 
+        # Unlike the blocks, connection parameters are defined in the connection
+        # domain definition files, as all connections within the same domain
+        # share the same properties.
+        param_factory = self.parent_platform.make_param
+        conn_parameters = self.parent_platform.connection_params.get(self.type, {})
+        self.params = collections.OrderedDict(
+            (data['id'], param_factory(parent=self, **data))
+            for data in conn_parameters
+        )
+
     def __str__(self):
         return 'Connection (\n\t{}\n\t\t{}\n\t{}\n\t\t{}\n)'.format(
             self.source_block, self.source_port, self.sink_block, self.sink_port,
@@ -56,6 +75,10 @@ class Connection(Element):
 
     def __iter__(self):
         return iter((self.source_port, self.sink_port))
+
+    def children(self):
+        """ This includes the connection parameters """
+        return self.params.values()
 
     @lazy_property
     def source_block(self):
@@ -78,6 +101,18 @@ class Connection(Element):
             true if source and sink blocks are enabled
         """
         return self.source_block.enabled and self.sink_block.enabled
+
+    @property
+    def label(self):
+        """ Returns a label for dialogs """
+        src_domain, sink_domain = [
+            self.parent_platform.domains[d].name for d in self.type]
+        return f'Connection ({src_domain} → {sink_domain})'
+
+    @property
+    def namespace_templates(self):
+        """Returns everything we want to have available in the template rendering"""
+        return {key: param.template_arg for key, param in self.params.items()}
 
     def validate(self):
         """
@@ -113,9 +148,33 @@ class Connection(Element):
         Export this connection's info.
 
         Returns:
-            a nested data odict
+            A tuple with connection info, and parameters.
         """
-        return (
+        # See if we need to use file format version 2:
+        if self.params:
+            return {
+                'src_blk_id': self.source_block.name,
+                'src_port_id': self.source_port.key,
+                'snk_blk_id': self.sink_block.name,
+                'snk_port_id': self.sink_port.key,
+                'params': collections.OrderedDict(sorted(
+                    (param_id, param.value)
+                    for param_id, param in self.params.items())),
+            }
+
+        # If there's no reason to do otherwise, we can export info as
+        # FLOW_GRAPH_FILE_FORMAT_VERSION 1 format:
+        return [
             self.source_block.name, self.source_port.key,
-            self.sink_block.name, self.sink_port.key
-        )
+            self.sink_block.name, self.sink_port.key,
+        ]
+
+    def import_data(self, params):
+        """
+        Import connection parameters.
+        """
+        for key, value in params.items():
+            try:
+                self.params[key].set_value(value)
+            except KeyError:
+                continue
--- a/grc/core/schema_checker/domain.py
+++ b/grc/core/schema_checker/domain.py
@@ -1,9 +1,11 @@
 from .utils import Spec, expand
+from .block import PARAM_SCHEME
 
 DOMAIN_CONNECTION = expand(
     type=Spec(types=list, required=True, item_scheme=None),
     connect=str,
     cpp_connect=str,
+    parameters=Spec(types=list, required=False, item_scheme=PARAM_SCHEME),
 )
 
 DOMAIN_SCHEME = expand(
--- a/grc/gui/Application.py
+++ b/grc/gui/Application.py
@@ -21,6 +21,8 @@ from .MainWindow import MainWindow
 from .PropsDialog import PropsDialog
 
 from ..core import Messages
+from ..core.Connection import Connection
+from ..core.blocks import Block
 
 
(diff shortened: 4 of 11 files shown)


gr-uhd/grc/rfnoc.domain.yml
```diff
@@ -7,4 +7,10 @@ multiple_connections_per_output: false
 
 templates:
 -   type: [rfnoc, rfnoc]
-    connect: self.rfnoc_graph.connect(self.${ source.parent_block.name }.get_unique_id(), ${ source.key }, self.${ sink.parent_block.name }.get_unique_id(), ${ sink.key }, False)
+    connect: self.rfnoc_graph.connect(self.${ source.parent_block.name }.get_unique_id(), ${ source.key }, self.${ sink.parent_block.name }.get_unique_id(), ${ sink.key }, False) \# ${not(params['is_back_edge'].get_value())}
```
mbr0wn, line 10: I would like this to look like
```suggestion
    connect: self.rfnoc_graph.connect(self.${ source.parent_block.name }.get_unique_id(), ${ source.key }, self.${ sink.parent_block.name }.get_unique_id(), ${ sink.key },  ${not is_back_edge}) 
```

haakov:
Interesting!

Are there other properties we might want to associate with connections? 

My initial thought is that connection props probably should be an opt-in feature (through a Preferences window or similar)

skoslowski:
I like this. Your approach is sensible. It is consistent to how its handled on blocks. Off the top of my head, I'd guess
 
* your missing `Connection.import_data()`
* you'll want to provide a namespace of parameter values to the template rendering, see `Block.namespace_templates`. A full fleshed version of this would also be able to make use of (flow graph) variables, which would entail copying more functions from `Block` - or have both classes share them

To make this optional, as Hakon suggested, I'd simply disable the PropsDialog for connections that have no parameters.

mbr0wn:
Thanks for the feedback!

> I like this. Your approach is sensible. It is consistent to how its handled on blocks. Off the top of my head, I'd guess
> 
>     * your missing `Connection.import_data()`

Yup, thanks for the pointer. That solved my settings-getting-lost problem.

>     * you'll want to provide a namespace of parameter values to the template rendering, see `Block.namespace_templates`. A full fleshed version of this would also be able to make use of (flow graph) variables, which would entail copying more functions from `Block` - or have both classes share them

This was also an easy fix. It solved my issue of wanting access to  the `is_back_edge` variable directly in the template. I didn't go the fully fleshed route (yet?).

> To make this optional, as Hakon suggested, I'd simply disable the PropsDialog for connections that have no parameters.

That was a one-liner, luckily.

gr-uhd/grc/rfnoc.domain.yml
mbr0wn: Thanks @skoslowski for helping me resolve this.

mbr0wn:
@skoslowski your comment answered all my questions, and I think this is usable now. I removed the 'Don't Merge' tag.

dkozel:
I could imagine there being properties in a custom buffers implementation that this UI could apply to. But I still have not used custom buffers in ernest so don't have specifics.

grc/core/Connection.py
```diff
@@ -79,6 +103,16 @@ def enabled(self):
         """
         return self.source_block.enabled and self.sink_block.enabled
 
+    @property
+    def label(self):
+        """ Returns a label for dialogs """
+        return 'Connection'
```
skoslowski, line 109: idea: Mention the connection domain here

grc/core/Connection.py
```diff
@@ -113,9 +147,26 @@ def export_data(self):
...
+        con_data = (
             self.source_block.name, self.source_port.key,
-            self.sink_block.name, self.sink_port.key
+            self.sink_block.name, self.sink_port.key,
         )
+        # We only attach params if they exist, this will help with
+        # backward-compatibility. This way, regular graphs without connection
+        # parameters won't touch the connection format.
+        if self.params:
+            con_data += (collections.OrderedDict(sorted(
```
skoslowski, line 160: NIT: consider making con_data a list and append to it here

grc/core/schema_checker/domain.py
```diff
@@ -1,9 +1,25 @@
...
+
+    dtype=str,
+    default=object,
+
+    options=list,
+    option_labels=list,
+    option_attributes=Spec(types=dict, required=False,
+                           item_scheme=(str, list)),
+)
+
```
skoslowski, line 17: IDEA: reuse PARAM_SCHEME from block.py. Looks like, the only difference is `hide`, which would be treated as `required=False`.

grc/gui/Application.py
```diff
@@ -811,11 +834,16 @@ def flow_graph_update(fg=flow_graph):
 
         selected_blocks = list(flow_graph.selected_blocks())
         selected_block = selected_blocks[0] if selected_blocks else None
+        # See if a connection has modifiable parameters or grey out the entry
+        # in the menu
+        selected_connections = list(flow_graph.selected_connections())
+        selected_connection = selected_connections[0] if selected_connections else None
+        selected_conn_has_params = selected_connection and bool(len(selected_connection.params))
```
skoslowski, line 841: IDEA: unless we know, that the clicked connection is at index 0, I suggest to disable the properties. Otherwise a use can't know which one is being edited (from e.g. the Dialog title)

skoslowski (approved):
This great work! I had a couple of minor nits. 

It might be worth discussing, whether modified parameters should be visualized in canvas somehow

grc/gui/Application.py
mbr0wn: Ah yeah, good point. If you've selected multiple connections.

grc/core/Connection.py
mbr0wn: Sure why not.

grc/core/Connection.py
mbr0wn: image

mbr0wn:
@skoslowski Thanks for the feedback! I addressed all the issues.

willcode:
This will require `file_format` (`FLOW_GRAPH_FILE_FORMAT_VERSION`) to be bumped if a grc can not even be loaded by older software versions. Since this generates non-backward compatible files, I don't think we can backport as is.

Is there another way to store edge properties?

skoslowski:
Good catch. ATM connections are stored as list of 4-tuples. This adds an optional 5th item, which AFAIK breaks older versions. 
If we want to avoid bumping the format version, the connections could possibly be either saved as a mapping of  4-tuples to parameter values, or the connection parameters are saved as a complete separate list. Neither option feels clean to me.
If we need to bump the format version anyway, it might be worth replacing the 4-tuple with a mapping. This would be analogous to how blocks are saved and make adding future properties easier.

willcode:
The problem with non-backported features is that they may not be released for a long time. A less-optimal, but backward compatible, format would get this out there. Mapping sounds good for the future.

mbr0wn:
I don't think there's an easy, fully-backward compatible solution here. As @skoslowski says, the old code stores the edge info as a 4-tuple and GRC hardcodes its length.

I made sure that when dealing with flowgraphs that do not change edge properties, the resulting GRC does not include them. That means that most flow graphs will not produce backward-incompatible .grc files.

This means that you cannot open a .grc file in which edge properties are set with an old GRC. However, that seems fine to me: Since the edge property would be ignored, that also means that the old GRC wouldn't interpret the file correctly anyway. But you can still open any other .grc file with an older GRC.

mbr0wn:
I just had another look and between the schema and the actual code, the `connections:` section of a .grc file is a list of 4-tuples, and nothing else. The only way that I can think of to make this backward-compatible is to store the connection metadata separately from the connections, but that means if you open a new .grc file (with properties) in an old GRC and then save, you lose the connection properties. I would much rather fail to open a new file (with edge properties) in an old GRC.

Another fun fact: GRC will happily open .grc files with a `file_format` of 2 or higher.

mbr0wn:
@willcode why is backward compatibility a requirement for backporting? Can't we simply have .grc files that require a certain version of GNU Radio to run?

All that said, I also prefer mappings instead of lists.

mbr0wn:
> @mbr0wn People will create new flowgraphs for OOTs, the wiki, etc. If this only happened on purpose, it wouldn't be a big deal. We'll end up with a bunch of flowgraphs that are otherwise compatible, but can't be opened by older versions.
> 
> Hmm, the fact that version goes unchecked makes it difficult to do this right.

New flowgraphs are backward compatible, unless people use this new feature. In which case they shouldn't be backward compatible.

gr-uhd/grc/rfnoc.domain.yml
```diff
@@ -7,4 +7,10 @@ multiple_connections_per_output: false
 
 templates:
 -   type: [rfnoc, rfnoc]
-    connect: self.rfnoc_graph.connect(self.${ source.parent_block.name }.get_unique_id(), ${ source.key }, self.${ sink.parent_block.name }.get_unique_id(), ${ sink.key }, False)
+    connect: self.rfnoc_graph.connect(self.${ source.parent_block.name }.get_unique_id(), ${ source.key }, self.${ sink.parent_block.name }.get_unique_id(), ${ sink.key }, ${ is_back_edge })
+    parameters:
```
willcode, line 11: Why does `parameters` come out indented under `templates`?

willcode:
If we're going to do this, we might as well emit a dict - only in cases where there are connection params, and detect either form on load. Otherwise, we'll just have another incompatible change in the future. The user experience will be the same - fail to load with no explanation.

Even though there is no check for version currently, we should bump the version to 2 and add a check like the one currenlty done for block yml. It's easier to explain that older GRC forgot to do a version check than it is to have incompatible formats with the same version number.

We could emit version 1 files by default and version 2 only if there are any connection params in the FG. We just have to remember to handle this if more breaking changes are made in the future.

I hope that the feature sees more general use (otherwise RFNoC could just store its info in metadata or something). That would lead to more incompatible flowgraphs, so we have to be prepared for that.

Finally, I think more people need to weigh in on this @mormj @sdh11 @haakov because it's going to have implications for support and user experience.

gr-uhd/grc/rfnoc.domain.yml
mbr0wn: The parameters are specific to this connection type (`[rfnoc, rfnoc]`).

mbr0wn:
> If we're going to do this, we might as well emit a dict - only in cases where there are connection params, and detect either form on load. Otherwise, we'll just have another incompatible change in the future. The user experience will be the same - fail to load with no explanation.

I made this change...

> Even though there is no check for version currently, we should bump the version to 2 and add a check like the one currenlty done for block yml. It's easier to explain that older GRC forgot to do a version check than it is to have incompatible formats with the same version number.

...and also bumped the version to 2, but the compat check should come in a separate commit.

argilo:
It appears this has broken bus connections. Details in #7059.

PR #6890, opened 2023-10-04
analog: PLL frequency detector example: use limited-wait throttle

marcusmueller:
this actively enhances the quality of the example, since changes become
more immediately visible

## Description

While working through things, I opened this example to check whether I still understand PLL freqdet. I did, but the example wasn't great UX-wise.

## Checklist


marcusmueller:
Looks like on the current version of [the wiki block documentation](https://wiki.gnuradio.org/index.php?title=PLL_Frequency_Detector#Simple_Usage_Example) now, so:

fg

willcode:
Would it be better to put the throttle inline, as it would be typically used? Even if it works with no output connection, this seems to require the user to understand more about scheduler operation than necessary.

Issue #6895, opened 2023-10-07
fmcomms2/3/4 Source and Sink port types are reversed in GRC


dkozel:
### What happened?

There's an extra field in the YML for both blocks that's causing a mismatch in the displayed label and the selected type.

### System Information

OS: (Linux Distro, Mac, Windows)
GR Installation Method: (native distro packages - apt, rpm, etc., Conda, Source

### GNU Radio Version

3.11-git (main)

### Specific Version

0.0.0.0

### Steps to Reproduce the Problem

Add either block to GRC, select a data type in the block, observe the color of the data port.

PR #6896, opened 2023-10-07
iio: Remove extra element from fmcomms2 YML

dkozel:
A quick fix to https://github.com/gnuradio/gnuradio/issues/6895

I don't think this fully solves the issues with the IO signature of the Pluto blocks though. But I noticed this while debugging that issue. I need to sign this commit, but posting this for visibility given the 3.10 RC1


willcode:
Bypassed the DCO, will take this now.

PR #6912, opened 2023-10-12
CI: Don't force generic VOLK kernels during CI

argilo:
## Description
In 4ca2456b0120a6d1468665a6afd2094ad2da8998, GNU Radio's test suite was modified so that tests always execute with generic VOLK kernels instead of platform-specific ones (SSE, AVX, NEON, etc). The commit message says that this was to "help control precision issues." I suspect the intention was to get some relief from flaky tests; until recently there were many tests that checked for exact floating point equality, or had tolerances set too low.

Now that flaky tests have been corrected, I think we should consider removing the `VOLK_GENERIC` switch, so that tests will run in a more realistic environment. (End users do not typically run GNU Radio with `VOLK_GENERIC=1`.) Doing so may help catch more VOLK bugs, such as https://github.com/gnuradio/volk/issues/414.

I left this as a draft pull request for now, to see whether this reveals any new test failures. At a minimum, we'll likely need to disable `qa_polar_decoder_sc_systematic_test` on all platforms that have broken `volk_8u_x2_encodeframepolar_8u` VOLK kernels.

**Update:** The following tests fail:

All platforms:
* `qa_polar_decoder_sc_systematic_test`, due to https://github.com/gnuradio/volk/issues/414

Ubuntu 20.04 & Debian 11:
* `qa_fecapi_cc`, due to https://github.com/gnuradio/volk/pull/458, fixed in VOLK 2.5.0
* `qa_polar_decoder_sc*`, due to https://github.com/gnuradio/volk/pull/434, fixed in VOLK 2.5.0

Test suite / CI.

## Testing Done
I ran CI and checked what failed.

## Checklist
- [x] I have ~~added~~ **disabled** tests to cover my changes, and all previous tests pass.

### DIFF
--- a/.github/workflows/make-test.yml
+++ b/.github/workflows/make-test.yml
@@ -66,24 +66,24 @@ jobs:
           - distro: 'Ubuntu 20.04'
             containerid: 'gnuradio/ci:ubuntu-20.04-3.9'
             cxxflags: -Werror
-            ctest_args: '-E ""'
-            # ldpath: 
+            ctest_args: '-E "^(qa_fecapi_cc|qa_polar_decoder_sc|qa_polar_decoder_sc_list|qa_polar_decoder_sc_systematic)$"'
+            # ldpath:
           - distro: 'Ubuntu 22.04'
             containerid: 'gnuradio/ci:ubuntu-22.04-3.9'
             cxxflags: -Werror
-            ctest_args: '-E "qa_polar_..coder_(sc_)?systematic"'
+            ctest_args: '-E "^qa_polar_decoder_sc_systematic$"'
             # ldpath:
           - distro: 'Fedora 37 (with 0xFE memory initialization)'
             containerid: 'gnuradio/ci:fedora-37-3.9'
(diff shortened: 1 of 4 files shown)


argilo:
`qa_fecapi_cc` hangs on Ubuntu 20.04 and Debian 11. I don't yet know why.

argilo:
The `qa_fecapi_cc` hang is due to `assertEqual` being very slow when comparing two long & differing lists. The bug causing incorrect output was almost certainly https://github.com/gnuradio/volk/pull/458, fixed in VOLK 2.5.0.

argilo:
`qa_polar_decoder_sc`, `qa_polar_decoder_sc_list`, and `qa_polar_decoder_sc_systematic` all fail on Ubuntu 20.04 and Debian 11. Most likely due to the bug that was fixed here: https://github.com/gnuradio/volk/pull/434.

argilo:
Sadly, I don't think it's practical to do this yet because too many VOLK bugs have made it out to Linux distributions, so a lot of GNU Radio tests would have to be bypassed. Perhaps we can revisit this in a couple years.

PR #6918, opened 2023-10-16
grc: Fix way-over-backslashing in file paths

argilo:
## Description

aka https://xkcd.com/1638/ :smile: 

Each time File -> Open Recent is used on Windows, the number of backslashes in the filename doubles.

backslash

This happens because a `GLib.Variant` object is passed through `str()`, which produces a quoted (and escaped) string. The quotes are then removed, but the escaping remains. The underlying string can instead be accessed using the `get_string()` method.

The same broken pattern exists in two other places, so I fixed those as well.

GNU Radio Companion

## Testing Done
I verified that opening recent files works correctly (both through the menu and toolbar) on Linux and Windows. I also verified that creating new files still works (both through the menu and toolbar).

## Checklist

### DIFF
--- a/grc/gui/Application.py
+++ b/grc/gui/Application.py
@@ -610,16 +610,14 @@ class Application(Gtk.Application):
             main.new_page()
             args = (GLib.Variant('s', 'qt_gui'),)
             flow_graph = main.current_page.flow_graph
-            flow_graph.options_block.params['generate_options'].set_value(str(args[0])[
-                                                                          1:-1])
+            flow_graph.options_block.params['generate_options'].set_value(args[0].get_string())
             flow_graph.options_block.params['author'].set_value(getuser())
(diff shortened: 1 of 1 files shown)


willcode (approved):

willcode:
@argilo could you change the commit message to something less linky?

argilo:
Sure. I moved the XKCD reference down into the details.

willcode:
LGTM. I'll give the other reviewers some time to look this over.

marcusmueller (approved):
this seems, in any way, the right way to deal with a GVariant, but I'm really not sure why we need to use a Variant here. Anyways, can't test it – no wind\ows.

argilo:
> I'm really not sure why we need to use a Variant here

I'm not sure we have a choice, I think it's just how Gtk stores parameters in its menu system.

PR #6922, opened 2023-10-18
gr-uhd: RFNoC Rx-Streamer: Add start stream options

eltos:
## Description

Exposes the start_stream_cmd option and the time_spec field to GNU Radio. This is used by the start-stream command of RFNoC to initiate streaming.  
The default values keep the present behaviour

| Block | Defaults | Delayed start stream
|-|-|-|
|grafik|grafik|grafik

## Related Issue

For RFNoC loopback flow graphs which require data from the host, and without the time_spec option, one cannot guarantee that the host is ready to provide samples by the time the RFNoC graph is started. The time_spec allows to set a delay which can ensure that.

See the thread on the mailing list for details: https://www.mail-archive.com/usrp-users@lists.ettus.com/msg16664.html

RFNoC Rx Streamer

## Testing Done

Tested with UHD 4.4.0.0 and an X310 with the following flow-graph:

grafik

## Checklist

### DIFF
--- a/gr-uhd/include/gnuradio/uhd/rfnoc_rx_streamer.h
+++ b/gr-uhd/include/gnuradio/uhd/rfnoc_rx_streamer.h
@@ -45,12 +45,16 @@ public:
      * \param vlen Vector length
      * \param issue_stream_cmd_on_start If true, the streamer sends a stream
      *                                  command upstream.
+     * \param start_time_set If true, set start time spec to the stream command
+     * \param start_time The time spec for the stream command if start_time_set is true
      */
     static sptr make(rfnoc_graph::sptr graph,
                      const size_t num_chans,
                      const ::uhd::stream_args_t& stream_args,
                      const size_t vlen = 1,
-                     const bool issue_stream_cmd_on_start = true);
+                     const bool issue_stream_cmd_on_start = true,
+                     const bool start_time_set = false,
+                     const ::uhd::time_spec_t& start_time = ::uhd::time_spec_t(0.0));
 
     //! Return the unique ID associated with the underlying RFNoC streamer
     virtual std::string get_unique_id() const = 0;
--- a/gr-uhd/lib/rfnoc_rx_streamer_impl.cc
+++ b/gr-uhd/lib/rfnoc_rx_streamer_impl.cc
@@ -32,10 +32,17 @@ rfnoc_rx_streamer::sptr rfnoc_rx_streamer::make(rfnoc_graph::sptr graph,
                                                 const size_t num_chans,
                                                 const ::uhd::stream_args_t& stream_args,
                                                 const size_t vlen,
-                                                const bool issue_stream_cmd_on_start)
+                                                const bool issue_stream_cmd_on_start,
+                                                const bool start_time_set,
+                                                const ::uhd::time_spec_t& start_time)
 {
-    return gnuradio::make_block_sptr<rfnoc_rx_streamer_impl>(
-        graph, num_chans, stream_args, vlen, issue_stream_cmd_on_start);
+    return gnuradio::make_block_sptr<rfnoc_rx_streamer_impl>(graph,
+                                                             num_chans,
+                                                             stream_args,
+                                                             vlen,
+                                                             issue_stream_cmd_on_start,
+                                                             start_time_set,
+                                                             start_time);
 }
 
 
@@ -43,7 +50,9 @@ rfnoc_rx_streamer_impl::rfnoc_rx_streamer_impl(rfnoc_graph::sptr graph,
                                                const size_t num_chans,
                                                const ::uhd::stream_args_t& stream_args,
                                                const size_t vlen,
-                                               const bool issue_stream_cmd_on_start)
+                                               const bool issue_stream_cmd_on_start,
+                                               const bool start_time_set,
+                                               const ::uhd::time_spec_t& start_time)
     : gr::sync_block(
           "rfnoc_rx_streamer",
           gr::io_signature::make(0, 0, 0),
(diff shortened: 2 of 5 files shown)


gr-uhd/lib/rfnoc_rx_streamer_impl.cc
```diff
@@ -58,7 +67,9 @@ rfnoc_rx_streamer_impl::rfnoc_rx_streamer_impl(rfnoc_graph::sptr graph,
       d_streamer(graph->create_rx_streamer(num_chans, stream_args)),
       d_unique_id(
           std::dynamic_pointer_cast<::uhd::rfnoc::node_t>(d_streamer)->get_unique_id()),
-      d_issue_stream_cmd_on_start(issue_stream_cmd_on_start)
+      d_issue_stream_cmd_on_start(issue_stream_cmd_on_start),
+      d_start_time_set(start_time_set),
+      d_start_time(start_time)
```
mbr0wn, line 72: Where do these get used? I'm probably staring right at it, but I can't figure it out...

gr-uhd/include/gnuradio/uhd/rfnoc_rx_streamer.h
```diff
@@ -45,12 +45,16 @@ class GR_UHD_API rfnoc_rx_streamer : virtual public gr::sync_block
...
+     * \param start_time_set If true, set start time spec to the stream command
+     * \param start_time The time spec for the stream command if start_time_set is true
      */
     static sptr make(rfnoc_graph::sptr graph,
                      const size_t num_chans,
                      const ::uhd::stream_args_t& stream_args,
                      const size_t vlen = 1,
-                     const bool issue_stream_cmd_on_start = true);
+                     const bool issue_stream_cmd_on_start = true,
+                     const bool start_time_set = false,
```
mbr0wn, line 56: Can we make this redundant? In UHD, a time of 0 often means 'ASAP' (see here: https://github.com/EttusResearch/uhd/blob/master/host/include/uhd/types/time_spec.hpp#L33-L34). Or maybe a `boost::optional`?

mbr0wn (commented):
I like this in principle. It's similar to USRP source and I think it's a great addition. Maybe we can do better on the arguments (i.e., skip the bool).

And maybe the behaviour should be more similar to USRP source (e.g., provide a start time API: https://github.com/mbr0wn/gnuradio/blob/70ec6a0877f01b248babe437044b7c90bdf4833c/gr-uhd/include/gnuradio/uhd/usrp_source.h#L82).

Also, I'm certain you would have to modify `start()`, no? https://github.com/mbr0wn/gnuradio/blob/70ec6a0877f01b248babe437044b7c90bdf4833c/gr-uhd/lib/rfnoc_rx_streamer_impl.cc#L82-L90

gr-uhd/lib/rfnoc_rx_streamer_impl.cc
eltos: These are used in the `start()` method, as you pointed out in your comment above:
https://github.com/mbr0wn/gnuradio/blob/70ec6a0877f01b248babe437044b7c90bdf4833c/gr-uhd/lib/rfnoc_rx_streamer_impl.cc#L82-L90

I did not modify the method, just added the parameters to the constructor, so they can be set from within GNU Radio. What modifications to `start()` do you have in mind?

gr-uhd/include/gnuradio/uhd/rfnoc_rx_streamer.h
eltos: The [`::uhd::stream_cmd_t`](https://files.ettus.com/manual/structuhd_1_1stream__cmd__t.html) has two fields: `stream_now` and `time_spec`, and they are used both inside the [`start()`](https://github.com/mbr0wn/gnuradio/blob/70ec6a0877f01b248babe437044b7c90bdf4833c/gr-uhd/lib/rfnoc_rx_streamer_impl.cc#L82-L90) method depending on the `d_start_time_set` boolean. Actually, you @mbr0wn wrote this piece of code yourself 3 years ago - I guess you had your reasons to do it like this, and not using a time of 0?   
When you say "0 often means 'ASAP'", the "often" does not sound convincing to be honest ;)  
But I don't know, you are the expert ;)

eltos: BTW the [`usrp_source`](https://github.com/mbr0wn/gnuradio/blob/70ec6a0877f01b248babe437044b7c90bdf4833c/gr-uhd/lib/usrp_source_impl.cc#L504-L516) never uses the `stream_now`, but **always** uses a "reasonable_delay" of 0.1s, see https://github.com/mbr0wn/gnuradio/blob/70ec6a0877f01b248babe437044b7c90bdf4833c/gr-uhd/lib/usrp_source_impl.cc#L513

eltos:
> And maybe the behaviour should be more similar to USRP source (e.g., provide a start time API: https://github.com/mbr0wn/gnuradio/blob/70ec6a0877f01b248babe437044b7c90bdf4833c/gr-uhd/include/gnuradio/uhd/usrp_source.h#L82).

This exists in [`rfnoc_rx_streamer_impl.cc`](https://github.com/mbr0wn/gnuradio/blob/748481543fc63f6c166e0dffa74fe956bf55fd46/gr-uhd/lib/rfnoc_rx_streamer_impl.cc#L163-L167), but is not exposed to GNU Radio. I don't see an immediate requirement for it (there's no point in making this runtime-adjustable).

mbr0wn:
@eltos Also have you seen the latest changes to the RX radio (message port for streaming)?

eltos:
> @eltos Also have you seen the latest changes to the RX radio (message port for streaming)?

Yes, if you mean this (from mailing list thread linked above):
> [...] but it seams that the RFNoC RX Radio Block can be triggered to issue a start-stream command via it's message port (I cannot find any documentation about this, but the source code suggest it [1]). However, the RFNoC Rx Streamer Block will always issuing a start-stream command independently, since it is hardcoded.
> [1] https://github.com/gnuradio/gnuradio/blob/5dc195b0195e8a073b7f7c1ffd6c8c36d3bb0bd2/gr-uhd/lib/rfnoc_rx_radio_impl.cc#L164

For which the ability to disable the Rx streamers "start_stream_cmd" is important.

PS: Do you know where I can find the documentation?

gr-uhd/include/gnuradio/uhd/rfnoc_rx_streamer.h
mbr0wn: Hey don't ask me about something I did 3 years ago... or even last week :grin: 

No I think you're right. Having a second look at this (a few days later) I think this is fine.

mbr0wn (approved):
I was initially hesitant, because this looks different yet again from USRP source, and from the RX radio. But the changes are fairly slim, and actually match the code well enough. And it's definitely useful. Let's go for it!

PR #6928, opened 2023-10-24
spdlog: include spdlog system settings header before including spdlog/fmt/

### LINKED ISSUE #6927: 3.10.8.0 build failure
👋 trying to build the latest release, but run into some build issue. The error log is as below:

<details>
<summary>error build log</summary>

```
  /opt/homebrew/include/spdlog/fmt/fmt.h:27:14: fatal error: 'spdlog/fmt/bundled/core.h' file not found
  #    include <spdlog/fmt/bundled/core.h>
               ^~~~~~~~~~~~~~~~~~~~~~~~~~~
  1 error generated.
```

</details>

full build log, https://github.com/Homebrew/homebrew-core/actions/runs/6589047383/job/17902765159
relates to Homebrew/homebrew-core#151928

marcusmueller:
Otherwise, spdlog/fmt/anything.h might include the internal
fmt/anything.h, although spdlog is set up to use an external fmt.

This is addressed upstream in
https://github.com/gabime/spdlog/issues/2922 .

## Description

Just an `#include <spdlog/tweakme.h>` with an explanatory comment on why reordering the include would be a bad idea.

## Related Issue

Hopefully:

Fixes #6927

a couple, sadly including public runtime headers

## Checklist

### DIFF
--- a/gnuradio-runtime/include/gnuradio/io_signature.h
+++ b/gnuradio-runtime/include/gnuradio/io_signature.h
@@ -11,6 +11,9 @@
 
 #ifndef INCLUDED_IO_SIGNATURE_H
 #define INCLUDED_IO_SIGNATURE_H
+/* ensure that tweakme.h is included before the bundled spdlog/fmt header, see
+ * https://github.com/gabime/spdlog/issues/2922 */
+#include <spdlog/tweakme.h>
 
 #include <gnuradio/api.h>
 #include <spdlog/fmt/fmt.h>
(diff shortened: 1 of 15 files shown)


marcusmueller:
> Including tweakme in any code that uses spdlog just doesn't seem right.

it's only that code which uses spdlog/fmt without using spdlog beforehand!
The alternative would have either adding another layer of redirective header, or to include spdlog – something I actively don't want to do in io_signature.h

marcusmueller:
> I'd rather avoid github links in the code.

Sure, I can include a short explanation in place

willcode:
Another level of header would be better than tweakme. That really looks like an internal spdlog/packager thing. Even including spdlog would be better - you don't like that one due to compile performance?

BTW, spdlog headers use `#pragma once`, mostly without `#ifdef` guards, so we're already requiring the compiler to support it!

willcode:
Does this fail on anything other than homebrew, and will homebrew pull the latest fmt with the fix?

chenrui333:
yeah, homebrew just always build whatever latest dependency.

PR #6930, opened 2023-10-25
GRC: add Documentation URLs

marcusmueller:
GRC: enable setting of documentation URLs, relative or absolute in yaml

## Description

This allows OOT module developers to link to online or offline documentation, and allows us to disambiguate in-tree blocks with the same label, as currently, only the label is used to form the URL.

GRC: Optional block field

## Checklist

### DIFF
--- a/gr-uhd/grc/uhd_fpga_fft.block.yml
+++ b/gr-uhd/grc/uhd_fpga_fft.block.yml
@@ -9,4 +9,5 @@ outputs:
 -   domain: stream
     id: port0
 
+doc_url: UHD_FPGA_FFT
 file_format: 1
--- a/grc/core/blocks/_build.py
+++ b/grc/core/blocks/_build.py
@@ -18,9 +18,20 @@ from ._flags import Flags
 from ._templates import MakoTemplates
 
 
-def build(id, label='', category='', flags='', documentation='',
-          value=None, asserts=None,
-          parameters=None, inputs=None, outputs=None, templates=None, cpp_templates=None, **kwargs) -> Type[Block]:
+def build(id,
+          label='',
+          category='',
+          flags='',
+          documentation='',
+          value=None,
+          asserts=None,
+          parameters=None,
+          inputs=None,
+          outputs=None,
+          templates=None,
+          cpp_templates=None,
+          doc_url=None,
+          **kwargs) -> Type[Block]:
     block_id = id
 
     cls = type(str(block_id), (Block,), {})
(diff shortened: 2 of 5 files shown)


marcusmueller:
This also properly aligns the link, and adds a tooltip that shows the URL:
Untitled

marcusmueller:
And here with the custom URL in action:
Untitled2

marcusmueller:
Note that we need to update the wiki on what to put into block yaml should we merge this.

duggabe:
That looks good!

marcusmueller:
The :globe_with_meridians: symbol only appears for URLs that aren't local (so, paths or starting with "file:")

mbr0wn:
This is nice! I wonder if we should introduce namespaces for OOTs? Let's say a popular OOT (like gr-satellites) would like to use the wiki. What if all blocks that link from there have a "gr_satellites" prefix in the URL, without having to specify that for every block?

mbr0wn:
In-tree blocks could maybe specify their own namespace somehow. These "FPGA" blocks have been abandoned, but we do plan to spruce them up. However, they have a special use case (building RFNoC images through GRC), so they're pretty special-case. They would warrant their own namespace.

cafeclimber:
Imo if we were to open up the wiki to OOT modules, we should establish a documentation standard. E.g. all blocks must be documented including all parameters and functionality, etc

marcusmueller:
I'd argue in line with @cafeclimber here; letting OOTs onto our wiki is akin to committing to co-maintaining their documentation. We'd put limits on that, especially because it limits what we can do to the wiki in the future without breaking whole ecosystems :) 

But! I don't think that was your main point, @mbr0wn; it was more whether modules should be able to define their own URL prefix. 

I think they shouldn't! (I was surprised to come to that conclusion, hence the exclamation mark.)

**Scenario**: an OOT with HTML files in docs/ ; or pages on a webserver with commonprefix

The overhead for adding the full URL vs the relative path to a URL prefix is probably negligible. The argument "making the prefix configurable for local installability" is a great one, and it's something that should be done at install time, statically string-replacing a relative path to a HTML file with the full installation path of the HTML file; installation of is the only time we know where the files end up. 

**Proposed solution**: have an 

```cmake
install_block_docs([DOWNLOAD]
                   File1… )
```
CMake command, which 

1. downloads the files from the specified URL when asked (`DOWNLOAD`) for it, 
3. installs the specified files to that target `${CMAKE_INSTALL_PREFIX}/${CMAKE_DATA_DIR}/docs/${CMAKE_PROJECT_NAME}/blocks/`

and amend our GRC block Yaml installation to respect a `[REMOVE_PREFIX prefix]` optional argument, to remove an URL prefix and replace it with `${CMAKE_INSTALL_PREFIX}/${CMAKE_DATA_DIR}/docs/${CMAKE_PROJECT_NAME}/blocks/`.

That way, the OOT maintainer can have absolute URLs pointing at whatever wiki/CMS/BTX page they want to. And it takes a flip of switch to download, install and correctly link to them.

**Side benefit**: same mechanism can be extended to download the in-tree block docs wiki pages.

PR #6932, opened 2023-10-27
grc: use yaml.CSafeLoader for performance if available

willcode:
## Description
This significantly speeds up creation of the block cache, where libyaml is built in to Python yaml. The cache is almost unnecessary.

The previous code noted whether `yaml.__with_libyaml__` was defined, but did not actually use the C loader.

GRC

## Testing Done
Ran both branches, noted speed.

## Checklist

### DIFF
--- a/grc/core/platform.py
+++ b/grc/core/platform.py
@@ -52,9 +52,6 @@ class Platform(Element):
         self._block_categories = {}
         self._auto_hier_block_generate_chain = set()
 
-        if not yaml.__with_libyaml__:
-            logger.warning("Slow YAML loading (libyaml not available)")
-
     def __str__(self):
         return 'Platform - {}'.format(self.config.name)
 
--- a/grc/core/io/yaml.py
+++ b/grc/core/io/yaml.py
@@ -80,5 +80,10 @@ def dump(data, stream=None, **kwargs):
     return yaml.dump_all([data], **config)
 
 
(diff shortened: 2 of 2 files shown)


willcode:
@haakov could you give this a try?

willcode:
This should also speed up flowgraph loading, as `platform.py` imports `yaml` from `.io`. The YAML Config block uses the Python implementation. This would matter only if config data was large.

grc/core/platform.py
```diff
@@ -52,9 +52,6 @@ def __init__(self, *args, **kwargs):
         self._block_categories = {}
         self._auto_hier_block_generate_chain = set()
 
-        if not yaml.__with_libyaml__:
-            logger.warning("Slow YAML loading (libyaml not available)")
```
haakov, line 56: Are you sure we want to remove this warning?

haakov:
Great! I'm getting these speeds on my system:

#### With CSafeLoader
- Building block library:  2 secs
- Building example library (feature-grc-qt branch): 2 secs

#### Without CSafeLoader
- Building block library: 18 secs
- Building example library (feature-grc-qt branch): 9 secs

grc/core/platform.py
willcode: The warning hasn't been printed with slow loading in the past, and if it is printed, there's nothing the average user can do about it, so it just adds noise.

Issue #6799, opened 2023-08-21
Add message to var block


Bob0111:
### Feature Description

Currently, there is a block called message pair to var, but the document says input message pairs are of the form (<name> . <value>). For this block, <name> is ignored. So I think we may create a block that can just change the variable with a simple message instead of a pair. Because in this case, the <name> is ignored. So just send the pmt message to the variable will make writing the embedded python block that can change variables values much simpler.

### Feature Urgency

high (impacts broad portion of ecosystem)

Bob0111:
Input message pairs are of the form (name . value). For this block, <name> is ignored.

Animesh312:
Can I work on this issue

marcusmueller:
@Animesh312 sure! Contributions are always welcome!

So, I propose that instead of a separate "special purpose" *message to variable* block, you rather add a simple *message to dict* block, where the block has
- one parameter, which sets the first element of the message, and
- one message input, and
- one message output

The handler of the input message `msg` creates a dictionary, where the *key* is the parameter passed to the block constructor, and the value is the unmodified input `msg`. 

That way, we can use the currently existing "Message Pair to Var", and it has utility beyond just setting variable through messages.

(There is a bit of a "political" reason to why I'd rather see you spend time on a block that creates messages out of other messages than to build more "message to var" blocks: While the variable mechanism in GRC-generated python is fun for a lot of purposes, it's architecturally very questionable, and, [in some places very broken](https://github.com/gnuradio/gnuradio/issues/6766) (just one example). The general idea of messages was that it allows for much safer, and much more "obvious", modifications of parameters of blocks at runtime. However, a block like I describe above would be universally useful.)

marcusmueller:
(another benefit is that a `message_to_message_pair` block is very easy to write a unit test for!)

PR #6950, opened 2023-11-12
blocks: msg_pair_to_var: Log with block logger, not gr.log

marcusmueller:
This also adds the missing unit test.

## Description

The block was logging using gr.log.warn. It should have been logging using its own logger!

also, no unit tests makes me sad.

## Related Issue

#6799 looked at the block, and so I looked at the block, and was met with the above issues

/

## Checklist

### DIFF
--- a/gr-blocks/python/blocks/qa_msg_pair_to_var.py
+++ b/gr-blocks/python/blocks/qa_msg_pair_to_var.py
@@ -1,0 +1,81 @@
+#!/usr/bin/env python
+#
+# Copyright 2023 Marcus Müller
+#
+# This file is part of GNU Radio
+#
+# SPDX-License-Identifier: GPL-3.0-or-later
+#
+#
+
+
+from gnuradio import blocks
+from gnuradio import gr_unittest
+from gnuradio import gr
+import pmt
+import sys
+import time
+
+
+class test_msg_pair_to_var(gr_unittest.TestCase):
+
+    def setUp(self):
+        self.backend = gr.dictionary_logger_backend()
+        gr.logging().add_default_sink(self.backend)
+        self.tb = gr.top_block(catch_exceptions=False)
+
+    def tearDown(self):
+        self.tb = None
+
+    def test_calling(self):
+        canary = 0xDEAFBEEF
+        self.test_val = 0
+
+        def test_f(value: int = 0):
+            self.test_val = value
+
+        dut = blocks.msg_pair_to_var(test_f)
+        test_msg = pmt.cons(pmt.intern("foo"), pmt.from_long(canary))
+        src = blocks.message_strobe(test_msg, 10)
+        self.tb.msg_connect(*(src, pmt.intern("strobe")),
+                            *(dut, blocks.msg_pair_to_var.IN_PORT))
+        self.tb.start()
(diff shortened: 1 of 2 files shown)


marcusmueller:
https://github.com/marcusmueller/gnuradio/actions/runs/6843384944/job/18605928276#step:6:9194

Only on windows this fails with:

```
   File "%SRC_DIR%\gr-blocks\python\blocks\qa_msg_pair_to_var.py", line 38, in test_calling
    test_msg = pmt.cons(pmt.intern("foo"), pmt.from_long(canary))
                                           ^^^^^^^^^^^^^^^^^^^^^
TypeError: from_long(): incompatible function arguments. The following argument types are supported:
    1. (x: int) -> pmt.pmt_python.pmt_base

Invoked with: 3736059631
```

uuuuuuuh, pretty sure 3736059631==0xdeadbeef is an integer value, and especially a valid python `int`. sure, it's larger than `1<<31`, but why is this failing here… will have to dig up what the correct behaviour is and why we're failing it (and whether PMT is missing a kind of obvious test case).

marcusmueller:
To fix that error, for now, let's `from_uint64_t` instead :/

marcusmueller:
Adding the don't merge until I squash (or squash yourself before merging, if you please!)

willcode:
@marcusmueller Squashed/merged to main. Conda win is giving errors related to deadbeef. I've re-run the test a couple of times, now re-running all the Conda versions. Could you take a look and see if there's still a problem there?

gr-blocks/python/blocks/qa_msg_pair_to_var.py
```diff
@@ -0,0 +1,81 @@
...
+        def test_f(value: int = 0):
+            self.test_val = value
+
+        dut = blocks.msg_pair_to_var(test_f)
+        test_msg = pmt.cons(pmt.intern("foo"), pmt.from_uint64(canary))
+        src = blocks.message_strobe(test_msg, 10)
+        self.tb.msg_connect(*(src, pmt.intern("strobe")),
+                            *(dut, blocks.msg_pair_to_var.IN_PORT))
+        self.tb.start()
+        time.sleep(15e-3)
```
argilo, line 43: The short wait time here readily causes test failures on my system if I add some CPU load with `stress -c 8`.

gr-blocks/python/blocks/qa_msg_pair_to_var.py
```diff
@@ -0,0 +1,81 @@
...
+            if value == nightingale and value != lark:
+                raise Exception("It was the nightingale, and not the lark")
+
+        dut = blocks.msg_pair_to_var(test_f)
+        test_msg = pmt.cons(pmt.intern("birb"), pmt.from_uint64(nightingale))
+        src = blocks.message_strobe(test_msg, 10)
+        self.tb.msg_connect(*(src, pmt.intern("strobe")),
+                            *(dut, blocks.msg_pair_to_var.IN_PORT))
+        self.tb.start()
+        time.sleep(50e-3)
```
argilo, line 65: This also causes failures, although less frequently.

argilo:
I opened #6981 to track the flaky tests.

PR #6963, opened 2023-11-17
project: add tox.ini, so that editors use the same formatting as github CI

marcusmueller:
## Description

I've been a bit annoyed by the fact that my neovim + pylsp would not complain about the same formatting as the CI.

Adding this instructs the pycodestyle as used by pylsp to adopt the same style.

None directly, hopefully future Python quality

## Checklist

### DIFF
--- a/.github/workflows/make-test.yml
+++ b/.github/workflows/make-test.yml
@@ -29,7 +29,10 @@ jobs:
     - uses: actions/checkout@v2
     - uses: quentinguidee/pep8-action@v1
       with:
+        # -----------------------------------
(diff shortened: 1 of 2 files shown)


marcusmueller:
I don't know whether this is enough for the CI and we could remove the manual settings from .github/workflows and use this as single source of truth.

marcusmueller:
> I don't know whether this is enough for the CI and we could remove the manual settings from .github/workflows and use this as single source of truth.

Sadly not. So, we keep notes around to make sure things aren't modified in one place only

PR #6964, opened 2023-11-17
GRC: clean up core Python, type annotate

marcusmueller:
- project: add tox.ini, so that editors use the same formatting as github CI
- grc: no Python type 'long'
- grc: remove Py3 backports
- grc/core/FlowGraph: refactor internal namespace refresh code
- grc/core/FlowGraph: remove unused imports
- grc/core: Type annotations, so I can somewhat sensibly work in my editor
- grc: param.py: refactor internal function name
- grc/core/FlowGraph: pythonify old strings, lists

## Description

I found bugs while trying to understand what GRC does, so I fixed them.

To achieve some understanding, type annotations proved to be very helpful, so I'm keeping them around for future readers

GRC

## Checklist


grc/core/blocks/block.py
```diff
@@ -11,15 +11,15 @@
 import collections
 import itertools
 import copy
-
 import re
-
 import ast
+import typing
```
cafeclimber, line 16: Type annotations can be restricted to "not runtime" with something like:

```python
from __future__ import annotations

from typing import TYPE_CHECKING

if TYPE_CHECKING:
    from typing import Optional, # ... etc
```

grc/core/params/dtypes.py
```diff
@@ -39,15 +46,15 @@ class ValidateError(Exception):
 
 
 @validates('id')
-def validate_block_id(param, black_listed_ids):
+def validate_block_id(param, black_listed_ids: List[str]):
```
cafeclimber, line 49: Can we type hint this? If there's many types, we could return `-> Union[possible type, possible type, ...]` and if necessary, make that a type alias?

grc/core/params/param.py
```diff
@@ -22,6 +23,8 @@
 @setup_names
 class Param(Element):
 
+    EvaluationType = Union[None, str, complex, float, int, bool, List[str], List[complex], List[float], List[int]]
```
cafeclimber, line 26: If something can be None, would it make more sense for the type to be `Union[all possible types except None]`, then later use `Optional[EvaluationType]`? I'm not sure which is most "pythonic"

cafeclimber (commented):
Heck yes types! Some minor questions / comments I had

grc/core/params/dtypes.py
```diff
@@ -7,17 +7,24 @@
...
 from .. import blocks
 from .. import Constants
 
 
 # Blacklist certain ids, its not complete, but should help
-ID_BLACKLIST = ['self', 'gnuradio'] + dir(builtins)
+ID_BLACKLIST = {'self', 'gnuradio'} | set(dir(builtins)) | set(keyword.kwlist)
+# Python >= 3.10 has soft keywords in a list, i.e. words that are reserved in
+# specific contexts, e.g. `match` isn't generally a keyword, but in `case` it is.
+if hasattr(keyword, 'softkwlist'):
```
willcode, line 21: Same as `set(keyword.kwlist)` above?

grc/core/blocks/block.py
marcusmueller: hm, what's the advantage of that? I tried that in a place where annotating with a more detailed type than `Element` was not possible due to circular imports, and it still falls on my feet during precompilation.

grc/core/params/dtypes.py
marcusmueller: We could, but the value is thrown away anyways, as far as I'm aware; maybe we just remove the `return value` from all these `validate_*` and `-> None`?

https://github.com/gnuradio/gnuradio/blob/dc5d108344856c94fd28ab8c1d0170be2b714678/grc/core/params/param.py#L164-L169

grc/core/params/param.py
marcusmueller: Yeah, that would be equivalent. I opted for explicitly listing `None`, because that's one possible evaluation. (Kind of ambiguous what `None` means here)

grc/core/params/dtypes.py
marcusmueller: Oh yes, good spotting!

marcusmueller: Well-spotted

grc/core/params/dtypes.py
marcusmueller: Let me actually quickly try this; will have to squash things here anyways

marcusmueller: Did the annotation: most of these validators weren't returning anything, and for the other two, since they are 100% only used through the `validators` "factory", that return value would never be usable. Hence, removed, and annotated all functions `-> None`.

grc/core/blocks/block.py
cafeclimber: Even with `__future__.annotations`? Ultimately I think preventing the import of things only used for typing can help with runtime speed esp when actually running the GUI (boot times). Of course that might be miniscule and might be premature optimization, though I do also find it personally cleaner to do these typing guards to restrict all typing related code to one place. Of course that could just be personal taste!

marcusmueller: ah, I didn't do that, as afaik that's Py 3.7, and we still are only depending on Py 3.6.5 …

We should probably raise that for `main`, but we can't for `maint-3.10`…

cafeclimber: 3.6?? Even 3.7 is end-of-life with 3.8 EOL coming next October....oof

willcode: Ubuntu 20.04 has Python 3.8, so I think we can relax the version requirement. The Python 3.6 requirement was based on Ubuntu 18.04. We were trying to keep GR 3.9 and GR 3.10 compatible at the time.

@mormj do you remember if we are intentionally allowing Python 3.6? @balister are there embedded platforms running GR 3.10 but Python 3.6?

haakov (approved):

grc/core/blocks/block.py
haakov: Hmm it would be interesting to see how much time the type annotations add to the GRC startup time

willcode:
Just to check - nothing in the current PR depends on anything newer than Python 3.6 anyway, right?

marcusmueller:
I did take care to not let that happen, but let me try a build on Ubuntu 18.04 (Py 3.6.9)

marcusmueller:
@willcode  can't find any platform with Python3 that wouldn't fall flat on multiple other core dependencies (mostly, boost < 1.69), so can't do a master build to test this :(

grc/core/FlowGraph.py
```diff
@@ -384,7 +397,7 @@ def remove_element(self, element):
     ##############################################
     # Import/Export Methods
     ##############################################
-    def export_data(self):
+    def export_data(self) -> OrderedDict[str, str]:
```
marcusmueller, line 400: Found the 3.6.5-incompatible type!

grc/core/blocks/block.py
```diff
@@ -61,10 +61,8 @@ def __init__(self, parent):
         param_factory = self.parent_platform.make_param
         port_factory = self.parent_platform.make_port
 
-        self.params = collections.OrderedDict(
-            (data['id'], param_factory(parent=self, **data))
-            for data in self.parameters_data
-        )
+        self.params: typing.OrderedDict[str, Param] = collections.OrderedDict(
```
marcusmueller, line 64: `OrderedDict` not in 3.6.5 `typing`

marcusmueller (commented):
yep, 3.6.5-unsupported type annotations did make their way into this.

willcode:
I don't mind bumping the Python requirement to 3.7, since GR 3.10 is long term.

marcusmueller:
@willcode  how about this: I make a maint-3.10 PR of this without the usage of `OrderedDict`; that *should* be Python3.6 compatible, even if I can't find a test platform to verify that.

On main, we bump the minimum Python version to Py3.7, 3.8 or 3.9

| distro | default Python version |
|---|----|
| debian 10 (oldoldstable) | 3.7.3 |
| debian 11 (oldstable) | 3.9.2 |
| debian 12 (stable) | 3.11.2 |
| debian testing/unstable | 3.11.4 |
| Ubuntu 18.04 | 3.6.9 |
| Ubuntu 20.04 | 3.8.2 |
| Ubuntu 22.04 | 3.10.4 / 3.10.6 |
| Ubuntu 23.04 | 3.11.4 |
| Ubuntu 24.04 (projection) | 3.11.4 |
| Fedora 38 | 3.11.6 |
| Fedora 39 | 3.12.0 | 
| Fedora 40 | 3.12.0 | 
| ~CentOS 8~ | can't check |
| Rocky8.4 | 3.6.9, 3.11 (through official backport `python3.11` package, but build currently impossible due to PyGObject)|
| Rocky9.2 | 3.9.16, 3.11 |

marcusmueller:
> I don't mind bumping the Python requirement to 3.7, since GR 3.10 is long term.

@willcode hm, bump on maint-3.10? That makes things easier and I could incorporate @cafeclimber's recommendations

grc/core/blocks/block.py
mormj: I'm supposing there could be some people still using RHEL type distros unfortunately - but I haven't seen 3.10 running on them anyhow.  I think the 3.6 decision was made a LONG time ago (like GR3.7-->3.8), and should be reevaulated.

willcode:
Also, I'm not too worried about type hints slowing things down, unless someone has some performance figures.

grc/core/blocks/block.py
marcusmueller: I *think* the decision for 3.6.5 went like 

> OK, how do we break portability to the least platforms, but directly end up on a version of Python3 that fully supports all the `f"An error occurred ({error}), ejecting the humans to space. Sorry Dave."` strings we want

and then we as @willcode said, there was motivation for bringing GR 3.9 to Ubuntu 18.04

cafeclimber:
Just to be clear, to me it's not just performance but makes it clear what the function of those imports are. Of course types from the typing module are clear, but types that aren't called in the file can also be imported for type purposes and putting them in that `TYPE_CHECKING` guard makes that clearer imo

willcode:
No need for `__future__` now that Python is 3.7.2?

cafeclimber:
> No need for `__future__` now that Python is 3.7.2?

I believe the `__future__` import is still required up to at least 3.10: https://docs.python.org/3/whatsnew/3.7.html#pep-563-postponed-evaluation-of-annotations

willcode:
Let's see if we get some testers on `main`.

PR #6968, opened 2023-11-20
Python: minimum version 3.7.2, was 3.6.5

willcode:
Even 3.7 is EOL at this point, so we may bump again. 3.7 allows the usage of certain type checking features.

Required for suggestions for #6964.
[Python release status](https://devguide.python.org/versions)


argilo (approved):
This seems safe, since GR 3.10's target OSes all appear to have Python >= 3.8.

marcusmueller (approved):

marcusmueller (changes requested):
Can we directly go for 3.7.2, please? https://docs.python.org/3/library/typing.html#typing.OrderedDict was added in 3.7.2.

PR #6970, opened 2023-11-21
qtgui: Remove unnecessary imports from templates

### LINKED ISSUE #6914: Adding QT GUI Range crashes GRC
### What happened?

A bug happened!
if i drop a qt gui range widget into the grc, the whole application will crash with no information.

### System Information

OS:Windows10 LTSC X64
GR Installation Method:
radioconda-2023.07.26-Windows-x86_64.exe

### GNU Radio Version

3.10 (maint-3.10)

### Specific Version

0.0.0.0

Open the GNU Radio Companion with "C:\Users\crackest\radioconda\python.exe C:\Users\crackest\radioconda\cwp.py C:\Users\crackest\radioconda C:\Users\crackest\radioconda\Scripts\gnuradio-companion.exe"
Search the QT GUI RANGE
drag it into the GRC.
the all crash....

### Relevant log output

```shell
no..
```

argilo:
## Description
When a QT GUI Range block is added to the canvas in GRC, there is a noticeable delay, which does not occur for other blocks. This happens because `from gnuradio.qtgui import Range, RangeWidget` causes a large number of shared libraries to be loaded. The import is unnecessary, because the flow graph template already imports `gnuradio.qtgui`:

https://github.com/gnuradio/gnuradio/blob/53de3dd32a9fc9ba1fa4e834487749340a6859a3/grc/core/generator/flow_graph.py.mako#L24-L30

The import can also cause a crash on Windows, as noted in #6914.

Note: The QT GUI RFNoC Fosphor Display (Byte Vector) block has the same bug, ~~but I'll fix that in a separate PR to make backporting simpler. (The block is not present in GNU Radio 3.10.)~~ **Update: The QT GUI RFNoC Fosphor Display (Byte Vector) block was in fact already merged to maint-3.10, so I'll include it here as well.**

## Related Issue
Fixes #6914.

* QT GUI Range
* QT GUI RFNoC Fosphor Display (Byte Vector)

## Testing Done
In GRC, I made a flow graph containing a Range widget and verified that the flow graph still executed normally.

I do not have RFNoC-capable hardware, so the only testing I did with the QT GUI RFNoC Fosphor Display (Byte Vector) block was to drop it onto a flow graph and verify that the delay was gone.

## Checklist

### DIFF
--- a/gr-qtgui/grc/qtgui_range.block.yml
+++ b/gr-qtgui/grc/qtgui_range.block.yml
@@ -64,9 +64,7 @@ asserts:
 - ${start <= stop}
 
 templates:
-    imports: |- 
-        from gnuradio.qtgui import Range, RangeWidget
-        from PyQt5 import QtCore
+    imports: from PyQt5 import QtCore
     var_make: self.${id} = ${id} = ${value}
     callbacks:
     - self.set_${id}(${value})
@@ -75,8 +73,8 @@ templates:
             win = 'self._%s_win'%id
             range = 'self._%s_range'%id
         %>\
-        ${range} = Range(${start}, ${stop}, ${step}, ${value}, ${min_len})
-        ${win} = RangeWidget(${range}, self.set_${id}, "${no_quotes(label,repr(id))}", "${widget}", ${rangeType}, ${orient})
+        ${range} = qtgui.Range(${start}, ${stop}, ${step}, ${value}, ${min_len})
+        ${win} = qtgui.RangeWidget(${range}, self.set_${id}, "${no_quotes(label,repr(id))}", "${widget}", ${rangeType}, ${orient})
         ${gui_hint() % win}
 
 documentation: |-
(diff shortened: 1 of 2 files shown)


willcode:
Is there a reason GRC tries to import runtime imports? We don't have any live stuff happening in GRC, so this seem unnecessary.

argilo:
> Is there a reason GRC tries to import runtime imports? We don't have any live stuff happening in GRC, so this seems unnecessary.

Perhaps it's useful for something like Quadrature Demod, which does `include math` and uses `math.pi` in a parameter that GRC would need to evaluate.

willcode:
Yah, it doesn't seem easy to figure out all the uses or figure out how to change GRC, so stripping unneeded imports out of the templates sounds good. It also looks like hier block might need this.

argilo:
The importing is triggered on this line:

https://github.com/gnuradio/gnuradio/blob/53de3dd32a9fc9ba1fa4e834487749340a6859a3/grc/core/FlowGraph.py#L244

Perhaps a GRC expert would know more.

mbr0wn (approved):
This is fine. I copy pasta'd that line from another grc yaml and must have thought it necessary .

PR #6972, opened 2023-11-25
GRC: generate valid C++ string parameters without explicit double quotes

### LINKED ISSUE #2519: GRC doesn't generate valid C++ string parameters unless double-quotes are explicitly added
When a string parameter is empty or does not have quotes (as is the case for many default parameter values), the generated C++ string is surrounded by single quotes, which is valid Python but not C++. If the string is given with single-quotes, the generated C++ string will have single-quotes. The only time a valid C++ string is generated is if the value itself is already surrounded by double-quotes.

[Attached](https://github.com/gnuradio/gnuradio/files/3243256/cpp_string_bug.txt) is a minimal .grc file (with the extension changed to .txt so Github will accept it) that demonstrates the behavior as of 27dd99e49a7d34eeb4c0a8fedfe3b072ad582662 under the following conditions:

Condition|Python Output|C++ Output
-------------|----------|----------
String parameter with no quotes|String in single quotes (valid)|String in single quotes **(invalid)**
String parameter in single quotes|String in single quotes (valid)|String in single quotes **(invalid)**
String parameter is empty|String in single quotes (valid)|String in single quotes **(invalid)**
String parameter is set to empty single-quotes|String in single quotes (valid)|String in single quotes **(invalid)**
String parameter is set to empty double-quotes|String in double quotes (valid)|String in double quotes(valid)
String parameter in double quotes|String in double quotes (valid)|String in double quotes(valid)

### LINKED ISSUE #6551: zmq: C++ template requires explicit double quotes around address and filter key params
### What happened?

Invalid C++ code was generated, where unquoted or single-quoted strings were used as parameters for ZMQ blocks.

### System Information

Codegen

### GNU Radio Version

3.11-git (main)

### Specific Version

3.11.0.0git

Generate a C++ program using a ZMQ Sub Source, using and unquoted `ipc:///tmp/zmq.sock` (for example) for Address, and leave Filter Key empty. This will generate the line:
```
this->zeromq_sub_source_0 = gr::zeromq::sub_source::make(
    sizeof(gr_complex), 1,
    const_cast<char *>('ipc:///tmp/zmq.sock'),
    100,
    false,
    (-1),
    ''
);
```
(indenting added for clarity). The address and filter key params are single-quoted, not double-quoted.
The same thing happens if the params are single-quoted. If double-quotes are used, they carry through, and both C++ and Python code are valid.

### Relevant log output

_No response_

AsriFox:
## Description
When generating C++ code, GRC did not have a standard way to translate Python strings to C++ strings, because Mako uses simple text substitution. To address this, additional translations are added to generators for blocks and parameters/variables with the first commit.

Translations already used in the block definitions were removed with the second commit:
- manual translations with `no_quotes` helper, like in [qtgui::waterfall_sink_x](https://github.com/gnuradio/gnuradio/blob/9b3183aec7b629fd7fbb2b1574f0bb1c87eb53ee/gr-qtgui/grc/qtgui_waterfall_sink_x.block.yml#L329);
- regex translations in `cpp_templates`, like in [audio::source](https://github.com/gnuradio/gnuradio/blob/9b3183aec7b629fd7fbb2b1574f0bb1c87eb53ee/gr-audio/grc/audio_source.block.yml#L48):
```
translations:
  "'": '"'
  True: true
  False: false
```

## Related Issue
Fixes #2519
Fixes #6551

GRC
Block definition files in: gr-analog, gr-audio, gr-blocks, gr-digital, gr-fft, gr-filter, gr-network, gr-qtgui, gr-zeromq

## Testing Done
Generated C++ code from this flowgraph:
[cpp_string_bug_fixed.grc.txt](https://github.com/gnuradio/gnuradio/files/13464281/cpp_string_bug_fixed.grc.txt)

Issue fixed for simple and widely-used cases:
| Condition | GRC | Display | C++ Output |
|-|-|-|-|
| String parameter is empty | `value: ''` | | `""` |
| String parameter is set to empty single-quotes | `value: ''''''` | | `""` |
| String parameter is set to empty double-quotes | `value: '""'` | | `""` |
| String parameter with no quotes | `value: 127.0.0.1` | 127.0.0.1 | `"127.0.0.1"` |
| String parameter in single quotes | `value: '''127.0.0.1'''` | 127.0.0.1 | `"127.0.0.1"` |
| String parameter in double quotes | `value: '"127.0.0.1"'` | 127.0.0.1 | `"127.0.0.1"` |
| Multi-part string in double quotes | `value: '"127.0"".0.1"'` | 127.0.0.1 | `"127.0"".0.1"` |
| Escaped double quotes in quoted strings | `value: '''127.0.\"\"0.1'''` | 127.0"".0.1 | `"127.0\"\".0.1"` |

Problems arise for these cases, because substitution is regex-based:
| Condition | GRC | Display | C++ Output |
|-|-|-|-|
| Multi-part string in single quotes | `value: '''127.0.''''0.1'''` | 127.0.0.1 | `"127.0''.0.1"` |
| Escaped single quotes in quoted strings* | `value: '''127.0.\''\''0.1'''` | 127.0''.0.1 | `"127.0\"\".0.1"` |
| Raw string literals** | `value: R'127.0.0.1'` | 127.0.0.1 | `R"127.0.0.1"` |

\* Escaped single-quotes are translated with regex when used as values for block parameters. Variable/parameter blocks do not use the regex, so they produce the correct result: "127.0.\'\'0.1".

\** C++ raw string literals require delimiters (parenthesis, etc.), while Python raw string literals don't. This C++ output does not compile.

## Checklist

### DIFF
--- a/grc/core/blocks/block.py
+++ b/grc/core/blocks/block.py
@@ -15,7 +15,7 @@ import re
 import ast
 import typing
 
-from ._templates import MakoTemplates
+from ._templates import MakoTemplates, no_quotes
 from ._flags import Flags
 
 from ..base import Element
@@ -101,7 +101,7 @@ class Block(Element):
         try:
             clean_bus_structure = self.evaluate(bus_structure)
             return clean_bus_structure
-        except:
+        except Exception:
             return None
 
     # region Rewrite_and_Validation
@@ -373,7 +373,7 @@ class Block(Element):
         """Gets the block's current source bus structure."""
         try:
             bus_structure = self.params['bus_structure_source'].value or None
-        except:
+        except Exception:
             bus_structure = None
         return bus_structure
 
@@ -382,7 +382,7 @@ class Block(Element):
         """Gets the block's current source bus structure."""
         try:
             bus_structure = self.params['bus_structure_sink'].value or None
-        except:
+        except Exception:
             bus_structure = None
         return bus_structure
 
@@ -441,31 +441,33 @@ class Block(Element):
         self.cpp_templates = copy.copy(self.orig_cpp_templates)
 
         # Determine the lvalue type
-        def get_type(element: str, _vtype: typing.Optional[type] = None) -> str:
+        def get_type(element: str, vtype: typing.Optional[type] = None) -> str:
             evaluated = None
             try:
                 evaluated = ast.literal_eval(element)
-                if _vtype is None:
-                    _vtype = type(evaluated)
+                if vtype is None:
+                    vtype = type(evaluated)
             except ValueError or SyntaxError as excp:
-                if _vtype is None:
+                if vtype is None:
                     print(excp)
-            simple_types = {int: "int", float: "double", bool: "bool", complex: "gr_complex", str: "std::string"}
-            if _vtype in simple_types:
-                return simple_types[_vtype]
-            if _vtype == list:
+            simple_types = {int: "long", float: "double", bool: "bool", complex: "gr_complex", str: "std::string"}
+            if vtype in simple_types:
+                return simple_types[vtype]
+
+            elif vtype == list:
                 try:
                     # For container types we must also determine the type of the template parameter(s)
                     return f"std::vector<{get_type(str(evaluated[0]), type(evaluated[0]))}>"
+
                 except IndexError:  # empty list
                     return 'std::vector<std::string>'
 
-            if _vtype == dict:
+            elif vtype == dict:
                 try:
                     # For container types we must also determine the type of the template parameter(s)
-                    key = list(evaluated)[0]
-                    val = list(evaluated.values())[0]
+                    key, val = next(iter(evaluated.entries()))
                     return f"std::map<{get_type(str(key), type(key))}, {get_type(str(val), type(val))}>"
+
                 except IndexError:  # empty dict
                     return 'std::map<std::string, std::string>'
 
@@ -481,9 +483,19 @@ class Block(Element):
         if 'string' in self.vtype:
             self.cpp_templates['includes'].append('#include <string>')
 
-    def get_cpp_value(self, pyval):
+    def get_cpp_value(self, pyval, vtype: typing.Optional[type] = None) -> str:
+        """
+        Convert an evaluated variable value from Python to C++ with a defined type.
 
-        if type(pyval) == int or type(pyval) == float:
+        Returns:
+            string representation of the C++ value
+        """
+        if vtype is None:
+            vtype = type(pyval)
+        else:
+            assert vtype == type(pyval)
+
+        if vtype == int or vtype == float:
             val_str = str(pyval)
 
             # Check for PI and replace with C++ constant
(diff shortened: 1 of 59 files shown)


willcode:
Check `tools/autoformat_python.sh` and `tools/check_pep8.sh` for Python formatting rules.

If the rules have strayed from the one use by CI, the current args should be
`--max-line-length=120 --ignore E265,E266,E275,E402,E501,E704,E712,E713,E714,E711,E721,E722,E741,W504,W605`

AsriFox:
GRC test was failing because of a check I added in [Block.format_expr](https://github.com/AsriFox/gnuradio/blob/614b3633c43fba79cffa494e4a5249d9e33c7524/grc/core/blocks/block.py#L457). It is now fixed.
However, `int` and `long` type distinction is specific to C++, and #6964 removes it. Maybe I should remove it as well, and simply use `long` for all integers?

grc/core/blocks/block.py
```diff
@@ -451,71 +449,47 @@ def get_type(element, _vtype):
...
-                    '${value}', self.get_cpp_value(evaluated, self.vtype))
-            )
+            self.cpp_templates['var_make'] = self.cpp_templates['var_make'].replace(
+                '${value}', self.get_cpp_value(evaluated))
 
         if 'string' in self.vtype:
             self.cpp_templates['includes'].append('#include <string>')
 
-    def get_cpp_value(self, pyval, vtype: str):
+    def get_cpp_value(self, pyval, _vtype: typing.Optional[type] = None):
```
willcode, line 485: Why the `_` in the parameter name?

grc/core/blocks/block.py
```diff
@@ -442,49 +440,35 @@ def format_expr(self, py_type):
         self.cpp_templates = copy.copy(self.orig_cpp_templates)
 
         # Determine the lvalue type
-        def get_type(element, _vtype):
+        def get_type(element: str, _vtype: typing.Optional[type] = None) -> str:
```
willcode, line 443: Why `_` in `_vtype`?

AsriFox: Left it as it was before me ([source](https://github.com/gnuradio/gnuradio/blob/d7f887227add25431b948d3a18471efd56bb1945/grc/core/blocks/block.py#L443)). Perhaps, it denoted an optional argument?

haakov:
Thank you for the PR! I see that you've marked this as a draft, are there other things you want to add before removing the draft tag?

AsriFox:
I am unsure what to do with irregular strings (multi-part, escaped quotes, raw literals).
When using my example, in generated C++ code there are compilation errors in cases 8 (escaped single-quotes) and 9 (escaped double-quotes); in generated Python code, `zeromq_req_source_9` throws an `Invalid argument` error (escaped double-quotes).

willcode:
Should this be set back to "draft", or have the items above been addressed?

AsriFox:
Removed underscores.

haakov:
I see. Is there a reason why we should expect those irregular strings in flowgraphs? I'm not sure I've seen one.

If not, I'm inclined to merge this as-is. (Once the conflict with grc/core/blocks/block.py has been resolved, that is)

AsriFox:
I expect those irregular strings to be rare. Raw literals and escaping characters might be handled later.

haakov (approved):
Looks good! But these commits should be squashed before merging.

PR #6976, opened 2023-11-26
grcqt: Minor Issues Fixes

### LINKED ISSUE #6957: feature-grc-qt Minor Issues
This is a collection of minor issues or notes found while testing the new Qt version of GRC.

Last checked on 4e444b7ff

- [x] Properties menu does not open when using a block's right click menu
- [x] Opening a GRC file immediately modifies it so it cannot be closed without saving
- [x] Right clicking on an unselected block does not select it and give access to options like enable, disable, properties

AsriFox:
## Description
Fixes for minor issues with the Qt version of GRC:
- [x] Ctrl+Scroll zooms in on the top left corner. Maybe it should zoom in on where the mouse is?
- [x] Zooming with Ctrl +, Ctrl -, Ctrl 0 is not implemented
- [x] Generate fails if the flowgraph hasn't been saved
- [x] Right clicking on an unselected block does not select it and give access to options like enable, disable, properties

Note: "Zoom in" shortcut is set to Ctrl =, because Ctrl + requires Shift to be pressed.

## Related Issue
Fixes #6957 

GRC: Qt GUI

## Checklist


grc/gui_qt/components/canvas/block.py
```diff
@@ -443,6 +443,15 @@ def mousePressEvent(self, e):
...
         self.moveToTop()
 
+    def contextMenuEvent(self, e):
+        if not self.isSelected():
+            self.parent.clearSelection()
+            self.setSelected(True)
+
+        view = self.parent.view
+        contextMenu = view._app().MainWindow.menus["edit"]
+        pos = view.mapFromScene(e.pos())
```
haakov, line 453: `e.pos()` is in block-relative coordinates, so you will have to map to scene coords before you can map to view coords

grc/gui_qt/components/window.py
```diff
@@ -448,6 +445,25 @@ def createActions(self, actions):
         actions["properties"].setEnabled(False)
 
         # View Actions
+        actions["zoom_in"] = Action(
+            Icons("zoom-in"),
+            _("Zoom &in"),
+            self,
+            shortcut="Ctrl+=",  # TODO: Keys.ZoomIn is "Ctrl++"
```
haakov, line 452: I think we should use Keys.ZoomIn, to stay consistent with GRC Gtk and other applications

haakov (changes requested):
Thank you for the PR! 

I only have a few suggestions :smiley:

AsriFox:
There are issues with GRC tests:
- GitHub actions fail because `pytest-qt` is not found in the container;
- some tests fail because of a [typo](https://github.com/gnuradio/gnuradio/blob/82c17682d4c0a95c0cecc94991aeb96d66dc5d23/grc/core/blocks/block.py#L62): there should be `parent_platform`, not `parent.platform`
- tests that involve `add_block_from_query` fail because the query does not produce a block; the text is typed in, but autocompletion doesn't work with qtbot's input (for some reason).

haakov:
Great! Thank you for the typo, just fixed it in the feature-grc-qt branch. I'm moving the other issues you mentioned to https://github.com/gnuradio/gnuradio/issues/6957, so we still have it once this PR is closed.

haakov:
https://github.com/gnuradio/gnuradio/blob/cfd798900fdd78beae8935ac304c083f0aaa256e/grc/gui_qt/components/flowgraph.py#L79

This should probably be set to False, since otherwise GRC will crash if you try to generate immedately after generating a new flowgraph

haakov:
Merging once that's fixed

PR #6991, opened 2023-12-05
GRC: allow variable expressions for enum fields

### LINKED ISSUE #6870: Block properties: allow variable expressions for enum fields
### Feature Description

Block properties with `dtype: enum` are the only properties which can not be controlled dynamically by variables or expressions.
The drop-down menu for enums simply does not allow entering variables names.

Take the [File source](https://github.com/gnuradio/gnuradio/blob/8a55c4f14d7cd06f21e162e586130ce0756f9d89/gr-blocks/grc/blocks_file_source.block.yml#L17-L22) as an example:
Here "Repeat" is even runtime-adjustable, but because it is not possible to assign it via a variable, there is no way to actually change it at runtime.
```yml
-   id: repeat
    label: Repeat
    dtype: enum
    default: 'True'
    options: ['True', 'False']
    option_labels: ['Yes', 'No']
```
grafik

It would be good to have the ability to enter expressions, similar to how it is done for other types which have `option_labels` defined, for example [File sink](https://github.com/gnuradio/gnuradio/blob/8a55c4f14d7cd06f21e162e586130ce0756f9d89/gr-blocks/grc/blocks_file_sink.block.yml#L22C1-L27C33):
```yml
-   id: unbuffered
    label: Unbuffered
    dtype: bool
    default: 'False'
    options: ['False', 'True']
    option_labels: ['Off', 'On']
```
grafik

In this very specific example, one could also change enum to bool, but this feature request is about a general way to solve this constraint.

As a nice side effect, this would also allow creating type-dynamic hier blocks by controlling the blocks "Type" parameter (which is an enum) via a common variable that could be exposed as a (non-runtime adjustable) hier-block parameter.

### Feature Urgency

high (impacts broad portion of ecosystem)

### More Information

/label GRC

AsriFox:
## Description
Block properties with `dtype: enum` can now be controlled dynamically by variables or expressions. The evaluated expression is then checked against the enum's `options`.

## Related Issue
Fixes #6870

GRC

## Testing Done
[dvb_enums.grc](https://github.com/gnuradio/gnuradio/files/13562854/dvb_enums.grc.txt).txt

## Checklist

### DIFF
--- a/grc/core/params/dtypes.py
+++ b/grc/core/params/dtypes.py
@@ -126,3 +126,9 @@ def validate_gui_hint(param, _) -> None:
         param.parse_gui_hint(param.value)
     except Exception as e:
         raise ValidateError(str(e))
+
+
+@validates('enum')
+def validate_enum(param, _):
+    if param.get_evaluated() not in param.options:
+        raise ValidateError(f'Value {param.get_evaluated()} is not in enum options: {param.options}')
--- a/grc/core/params/param.py
+++ b/grc/core/params/param.py
@@ -121,9 +121,9 @@ class Param(Element):
 
     def get_value(self):
         value = self.value
-        if self.is_enum() and value not in self.options:
-            value = self.default
-            self.set_value(value)
+        # if self.is_enum() and value not in self.options:
+        #     value = self.default
+        #     self.set_value(value)
         return value
 
     def set_value(self, value):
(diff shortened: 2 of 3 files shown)


AsriFox:
Some parameters may benefit from not having this dynamic evaluation, like `output_language` or `generate_options` in Options block. This will probably require adding a new hidden dtype.

willcode:
Added Needs Action based on the comment above. Should this be changed to draft?

mbr0wn:

PR #7009, opened 2023-12-14
gr_modtool: fix non-C++ block detection and manipulation

### LINKED ISSUE #7008: gr_modtool: rename only works on C++ blocks
### What happened?

`gr_modtool rename` can't rename blocks, if they don't have a C++ implementation. 

I suspect that [this line](https://github.com/gnuradio/gnuradio/blob/6404f371abb97f282aa1aa270837ffa9f0ddce84/gr-utils/modtool/core/base.py#L37) in `get_block_candidates` should only be applied to `cpp_blocks`.

Other commands are not affected as much, because they can use regular expressions, so they work with Python blocks (e.g. [tx_hier](https://github.com/AsriFox/gr-dvbs2acm/blob/main/python/dvbs2acm/tx_hier.py)), GRC-only blocks (e.g. [modcod_probe](https://github.com/AsriFox/gr-dvbs2acm/blob/main/grc/dvbs2acm_modcod_probe.block.yml)) and header-only "blocks" (e.g. [dvbs2_config](https://github.com/AsriFox/gr-dvbs2acm/blob/main/include/gnuradio/dvbs2acm/dvbs2_config.h)). 

### System Information

OS: Ubuntu 23.10

### GNU Radio Version

3.11-git (main)

### Specific Version

gfac5e1a33

1. Create a Python block;
2. Run `gr_modtool rename` to rename the new block.

### Relevant log output

```shell
~/Documents/gr-dvbs2acm$ gr_modtool rename tx_hier
GNU Radio module name identified: dvbs2acm
ModToolException: Blockname for renaming does not exists!

~/Documents/gr-dvbs2acm$ gr_modtool rename
GNU Radio module name identified: dvbs2acm
Enter name of block/code to rename (without module name prefix): tx_hier
ModToolException: Blockname for renaming does not exists!
```

AsriFox:
## Description
[`get_block_candidates`](https://github.com/gnuradio/gnuradio/blob/6404f371abb97f282aa1aa270837ffa9f0ddce84/gr-utils/modtool/core/base.py#L25-L39) function is used in `gr_modtool` to detect blocks in the OOT module. However, the current implementation only detects C++ blocks, because the pattern for Python files does not check the directory recursively.

Also, the `rename` subcommand does not check for non-C++ blocks and throws an exception if there are no C++ implementation files for the selected block.

This PR presents the following fixes:
- [x] 1. Change the glob for Python block detection (`python/*.py` -> `python/*/*.py`).
- [x] 2. Add the ability to detect GRC-only (`grc/*.block.yml`) and header-only (`include/gnuradio/*/*.h`) "blocks".
- [x] 3. Add checks for the absence of C++ files in `rename` subcommand.

## Related Issue
Fixes #7008

gr_modtool

## Checklist

### DIFF
--- a/gr-utils/modtool/cli/bind.py
+++ b/gr-utils/modtool/cli/bind.py
@@ -52,7 +52,7 @@ def cli(**kwargs):
 def get_pattern(self):
     """ Get the regex pattern for block(s) to be parsed """
     if self.info['pattern'] is None:
-        block_candidates = get_block_candidates()
+        block_candidates = get_block_candidates(self.info['modname'], skip_grc=True, skip_python=True)
         with SequenceCompleter(block_candidates):
             self.info['pattern'] = cli_input(
                 'Which blocks do you want to parse? (Regex): ')
--- a/gr-utils/modtool/cli/disable.py
+++ b/gr-utils/modtool/cli/disable.py
@@ -32,7 +32,7 @@ def cli(**kwargs):
 def get_pattern(self):
     """ Get the regex pattern for block(s) to be disabled """
     if self.info['pattern'] is None:
-        block_candidates = get_block_candidates()
+        block_candidates = get_block_candidates(self.info['modname'])
         with SequenceCompleter(block_candidates):
             self.info['pattern'] = cli_input(
                 'Which blocks do you want to disable? (Regex): ')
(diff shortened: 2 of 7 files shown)


marcusmueller (approved):
Haven't tested, but looks good to me so far.

AsriFox:
The issue with `rename` was not checking if the header exists [here](https://github.com/gnuradio/gnuradio/blob/6404f371abb97f282aa1aa270837ffa9f0ddce84/gr-utils/modtool/core/rename.py#L134). The C++ impl/noblock rename function [here](https://github.com/gnuradio/gnuradio/blob/6404f371abb97f282aa1aa270837ffa9f0ddce84/gr-utils/modtool/core/rename.py#L70-L83) uses functions with checks, so it's fine for now.

marcusmueller (approved):

PR #7016, opened 2023-12-19
grc: Make core.Connection.type not lazy

### LINKED ISSUE #7013: Virtual Source/Sink Regression
### What happened?

When a flowgraph using a virtual source/sink in message domain is reloaded, the virtual source/sink ports are taken to be stream domain rather than message domain. This causes a domain mismatch error in the flowgraph.
```
No connection known between domains "stream" and "message"
No connection known between domains "message" and "stream"
```
The problem was most likely introduced in #6889 / #6977

Reported by Jason Merlo on gnuradio-discuss.

### System Information

From maint-3.10 source (v3.10.9.0-rc1)

### GNU Radio Version

3.10 (maint-3.10)

### Specific Version

3.10.9.0-rc1

[message_test.grc.txt](https://github.com/gnuradio/gnuradio/files/13708163/message_test.grc.txt)

### Relevant log output

_No response_

mbr0wn:
This fixes an issue where virtual blocks can not correctly update their type when connected to message sources/sinks.

Well, it fixes the example in #7013.

Aah, this reminds me of the good old SWIG days, where I would modify .i files without knowing what I'm doing. This is what this feels like, except this time, thought I was pretty decent at Python.

@haakov @sdh11 @skoslowski  if you wouldn't mind taking a look at this. What did I do here? Is this really a fix?

Update: OK, I see now what's going on. Well, mostly. `lazy_property` sets a value once, the first time it's read. Something in the edge-prop PR causes the `domain` attribute of a port to get set before everything is resolved, which is latched into the 'lazy property'.

I don't think `lazy_property` is necessary here, it seems a bit like a premature optimization.

## Related Issue

Fixes #7013.
Well, maybe. 

See #7013.

## Testing Done

Well, I tried running the example from the bug report, and it works now.

## Checklist

### DIFF
--- a/grc/core/Connection.py
+++ b/grc/core/Connection.py
@@ -88,7 +88,7 @@ class Connection(Element):
     def sink_block(self):
         return self.sink_port.parent_block
 
-    @lazy_property
+    @property
     def type(self):
         return self.source_port.domain, self.sink_port.domain
 
--- a/grc/core/utils/descriptors/_lazy.py
+++ b/grc/core/utils/descriptors/_lazy.py
@@ -3,11 +3,40 @@
 #
 # SPDX-License-Identifier: GPL-2.0-or-later
 #
+"""
+Class method decorators.
+"""
 
 import functools
 
+# pylint: disable=too-few-public-methods,invalid-name
 
-class lazy_property(object):
+
+class lazy_property:
+    """
+    Class method decorator, similar to @property. This causes a method that is
+    declared as @lazy_property to be evaluated once, the first time it is called.
+    Subsequent calls to this property will always return the cached value.
+
+    Careful! Not suitable for properties that change at runtime.
+
+    Example:
+
+    >>> class Foo:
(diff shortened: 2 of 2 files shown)


skoslowski (approved):
Sorry to hear this caused headache.... cache invalidation remains one of the "hard problems" :smile: 

Removing the caching should be safe. The alternative would be to do a reset (by deleting the instance attribute); maybe during rewrite, not sure... This property seems to be only used internally by the class, anyway. 

For the docs: Maybe mention the [cached_property](https://docs.python.org/3/library/functools.html#functools.cached_property) that was added in Python 3.8. Much better name (another "hard problem" :smile: )
What is the min. Python version for GRC? Maybe we could even switch to that?

willcode:
@skoslowski Python 3.7.2 as of the upcoming release.

willcode:
Force pushed Python formatting fixes.

PR #7020, opened 2023-12-20
network: Add timeout to prevent qa_tcp_sink from hanging

argilo:
## Description
On the sparc64 architecture, GNU Radio's circular buffers seem to be completely broken. (I'll open an issue for this, and have a look later to see if I can track down the root cause.) This causes most tests to fail. But `qa_tcp_sink` hangs, forcing the Debian build server to wait for 10 hours:

```
        Start 261: qa_tcp_sink
E: Build killed with signal TERM after 600 minutes of inactivity
```

QA test for TCP Sink.

## Testing Done
I verified that the test still passes, and that it exits after 30 seconds if I comment out the final `self.tb.start()` and `self.tb.stop()` lines to induce a hang.

## Checklist

### DIFF
--- a/gr-network/python/network/qa_tcp_sink.py
+++ b/gr-network/python/network/qa_tcp_sink.py
@@ -31,6 +31,7 @@ class qa_tcp_sink (gr_unittest.TestCase):
 
     def test_restart(self):
         serversocket = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+        serversocket.settimeout(30.0)
         serversocket.bind(('localhost', 2000))
         serversocket.listen()
 


argilo:
I opened #7021 for the circular buffer bug.

argilo:
`qa_tcp_sink` was created in https://github.com/gnuradio/gnuradio/pull/5372. So this mess is my fault. 😄

gr-network/python/network/qa_tcp_sink.py
```diff
@@ -31,6 +31,7 @@ def tearDown(self):
 
     def test_restart(self):
         serversocket = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+        serversocket.settimeout(30.0)
```
argilo, line 34: I used a large value (30 seconds) here to make sure that the test doesn't become flaky when running on a heavily loaded machine. The test still completes immediately when it passes, and the 30-second delay will only occur if the test fails.

PR #7028, opened 2023-12-23
pdu: Allow PDU to Tagged Stream to write partial PDUs

### LINKED ISSUE #6628: qa_fecapi_dummy is flaky
A CI run (https://github.com/gnuradio/gnuradio/actions/runs/4651656779/jobs/8231524712) recently failed with the following error:
```
117/265 Test #117: qa_fecapi_dummy ..............................***Failed    0.61 sec
.............
FAIL: test_async_00 (__main__.test_fecapi_dummy) (packed=False, rev_pack=True)
  File "/__w/gnuradio/gnuradio/gr-fec/python/fec/qa_fecapi_dummy.py", line 279, in test_async_00
    self.assertSequenceEqualGR(data_in, data_out)
  File "/__w/gnuradio/gnuradio/gnuradio-runtime/python/gnuradio/gr_unittest.py", line 115, in assertSequenceEqualGR
    self.assertEqual(len(data_in), len(data_out), msg="Lengths do not match")
AssertionError: 15360 != 15120 : Lengths do not match

Ran 14 tests in 0.393s

FAILED (failures=1)
```
I was able to reproduce this on my laptop by running the test in a loop (`while gr-fec/python/fec/qa_fecapi_dummy_test.sh; do :; done`) in parallel with a heavy CPU load (`stress -c 16`).

argilo:
## Description
The `tagged_stream_block` class is a wrapper to help make [Tagged Stream blocks](https://wiki.gnuradio.org/index.php/Tagged_Stream_Blocks). Unfortunately, it causes problems for the scheduler by refusing to produce any output unless the input and output buffers are long enough to contain an entire PDU:

https://github.com/gnuradio/gnuradio/blob/4d3eb9f321bf28d7d6310e304a6a0323cdfa66aa/gnuradio-runtime/lib/tagged_stream_block.cc#L107-L109

https://github.com/gnuradio/gnuradio/blob/4d3eb9f321bf28d7d6310e304a6a0323cdfa66aa/gnuradio-runtime/lib/tagged_stream_block.cc#L112-L116

If the scheduler observes that a block refuses to produce output and all of its upstream blocks are finished (i.e. won't produce any more output), then it concludes that the block is finished and terminates its worker thread.

For the PDU to Tagged Stream block in particular, this is problematic because the scheduler can terminate it while it still has PDUs waiting in its input queue. This causes test failures in the `qa_fecapi_*` tests, as noted in #6628. Failures can readily be reproduced by increasing `rep` to 7 in this test:

https://github.com/gnuradio/gnuradio/blob/4d3eb9f321bf28d7d6310e304a6a0323cdfa66aa/gr-fec/python/fec/qa_fecapi_repetition.py#L161

Prior to #323, the PDU to Tagged Stream block did not have this problematic behaviour; if it had PDUs in its input queue, it would always produce output. If there wasn't enough room to write out a full PDU, it would store the rest for later. Here I've rewritten the block to behave as it did before.

## Related Issue
Fixes #6628.
Reverts #323.

* PDU to Tagged Stream

## Testing Done
I verified that all tests that use the PDU to Tagged Stream block still pass.

## Checklist


argilo:
> the qa_fecapi_repetition test passes even if `rep` or `frame_size` are increased

Note: Increasing them too much can still cause trouble for another block (Tagged Stream to PDU):

`block_executor :error: sched: <block tagged_stream_to_pdu (9)> is requesting more input data  than we can provide.`

This block also inherits the problematic behaviours of `tagged_stream_block`.

argilo:
I suppose a more comprehensive solution (which would fix all Tagged Stream blocks) might be to instead implement input & output PDU buffering in the `tagged_stream_block` class. There would be a bit more risk of performance issues in making a change that affects all Tagged Stream blocks. Let me know what you think.

argilo:
I'll put this into draft while I explore the other option (adding buffering to `tagged_stream_block`) which I suspect may also solve the problem of Tagged Stream blocks being incompatible with custom buffers.

Issue #7021, opened 2023-12-20
Circular buffers are broken on sparc64


argilo:
Debian build logs for the sparc64 architecture show a large number of test failures, because GNU Radio's circular buffers don't work:

https://buildd.debian.org/status/logs.php?pkg=gnuradio&arch=sparc64

For 3.10.9.0~rc1-1 the failure was:
```
RuntimeError: gr::vmcircbuf_sysconfig
```

For 3.7.10.1-2 the failure was:
```
vmcircbuf_createfilemapping: createfilemapping is not available
gr::vmcircbuf_sysv_shm: shmat (3): Invalid argument
gr::vmcircbuf_sysv_shm: shmat (3): Invalid argument
gr::vmcircbuf_sysv_shm: shmat (3): Invalid argument
terminate called after throwing an instance of 'gr::signal'
Aborted
```

argilo:
From the logs on GNU Radio 3.7, it would appear that the third `shmat` call in the `vmcircbuf_sysv_shm` constructor returns `EINVAL`:

https://github.com/gnuradio/gnuradio/blob/cce52e9dfd1ea6e0a081e824754422852e0872b1/gnuradio-runtime/lib/vmcircbuf_sysv_shm.cc#L109

argilo:
From the linux man page, `EINVAL` means:

> Invalid `shmid` value, unaligned (i.e., not page-aligned and `SHM_RND` was not specified) or invalid `shmaddr` value, or can't attach segment at `shmaddr`, or `SHM_REMAP` was specified and `shmaddr` was `NULL`.

argilo:
Tests failed the same way on hppa, prior to Debian bypassing them:

https://buildd.debian.org/status/logs.php?pkg=gnuradio&arch=hppa

argilo:
I installed Debian in qemu-system-sparc64 and built gnuradio-runtime. With debug logging enabled, `runtime_qa_vmcircbuf.cc` produces the following output:
```
$ gnuradio-runtime/lib/runtime_qa_vmcircbuf.cc
Running 1 test case...
standard_tests :info: Testing gr::vmcircbuf_createfilemapping_factory...
pagesize :debug: Setting pagesize to 8192 B
gr::vmcircbuf :info: vmcircbuf_createfilemapping: CreateFileMapping is not available
gr::test_a_bunch :info: Failed to allocate gr::vmcircbuf number 1 of size 8192 (cum = 0)
[...]

PR #7032, opened 2023-12-28
analog: Fix buffer overruns in AGC3

argilo:
## Description
While experimenting with using `host_buffer` as a fallback for `buffer_double_mapped`, I noticed that `qa_agc` fails because `agc3_cc` overruns its input and output buffers: https://github.com/gnuradio/gnuradio/issues/7021#issuecomment-1871212351. This happens because its work function expects `noutput_items` to be a multiple of `iir_update_decim`, but its call to `set_output_multiple` is immediately overwritten by a call to `set_alignment`, which overwrites the output multiple with `volk_get_alignment() / sizeof(gr_complex)` (which could be 1, 2, 4, or 8, depending on the machine type). If `iir_update_decim` does not divide this number, then buffer overruns occur because the `work` function processes `ceil(noutput_items / iir_update_decim) * iir_update_decim` items.

* AGC3

## Testing Done
I ran `qa_agc` inside valgrind, with `#define FORCE_SINGLE_MAPPED` uncommented in io_signature.h, and verified that no overruns occur. I also added debug logging and verified that the scheduler sets `noutput_items` to a multiple of `4 * iir_update_decim`.

## Checklist

### DIFF
--- a/gr-analog/lib/agc3_cc_impl.cc
+++ b/gr-analog/lib/agc3_cc_impl.cc
@@ -17,7 +17,6 @@
 #include <gnuradio/io_signature.h>
 #include <spdlog/fmt/fmt.h>
 #include <volk/volk.h>
-#include <volk/volk_alloc.hh>
 #include <algorithm>
 #include <cmath>
 #include <numeric>
--- a/gr-analog/lib/agc3_cc_impl.cc
+++ b/gr-analog/lib/agc3_cc_impl.cc
@@ -56,8 +56,6 @@ agc3_cc_impl::agc3_cc_impl(float attack_rate,
     test_and_log_value_domain(iir_update_decim, "input power sampling stride");
     d_iir_update_decim = iir_update_decim;
     set_output_multiple(iir_update_decim * 4);
-    const int alignment_multiple = volk_get_alignment() / sizeof(gr_complex);
-    set_alignment(std::max(1, alignment_multiple));
 }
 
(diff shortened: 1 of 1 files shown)


argilo:
Removing `set_alignment` shouldn't have any impact on performance, because VOLK was removed from the main processing loop in #6274.

marcusmueller (approved):

marcusmueller:
`set_alignment` erasing `output_multiple` is a bit of a pitfall!

argilo:
Yeah, that should probably get its own issue since this isn't the first time it's caused trouble.

willcode:
Agree ... also the `set_unaligned()` and `set_alignment()` functions don't appear to interact. I'm not sure what `set_unaligned()` is there for, since it isn't used. Setting of alignment, unalignment (sketchy name) and output multiple should be ironed out.

willcode:
Can the `volk_alloc.hh` include be removed if the volk allocator is not used?

argilo:
> I'm not sure what `set_unaligned()` is there for, since it isn't used.

It is used by the block executor to keep track of how much unaligned data remains if the scheduler was not able to provide enough input items to a block that has called `set_alignment`.

argilo:
> Can the `volk_alloc.hh` include be removed if the volk allocator is not used?

Yes. Done.

willcode:
OK, I missed the non-zero call to `set_unaligned()` in `block_executor.cc`.

argilo:
Issue opened: #7033

PR #7039, opened 2024-01-03
blocks: Write debug samples at the correct time in Peak Detector2

argilo:
## Description
I made some experimental changes to the block executor to detect buffer overruns, and found that Peak Detector2 sometimes produces an extra sample on its "debug" output port without reporting it to the scheduler.

Peak Detector2 has an optional debug output which can be used to view the internal estimated mean (`d_avg`). When a peak is detected at sample `i` in the main loop, the `work` function exits early and reports to the scheduler that it has produced `i` samples (i.e. `optr[0]` through `optr[i-1]` and `sigout[0]` through `sigout[i-1]`). However, it writes all the way up to `sigout[i]` on the debug output, producing a total of `i+1` samples.

When the scheduler then invokes the `work` function a second time and the `d_found == true` loop is executed, the block skips writing `sigout[0]` since that sample was written on the previous invocation.

Producing a sample without reporting it to the scheduler seems like a bad idea. Blocks that produce different numbers of items on each output are supposed to invoke `produce()` and return `WORK_CALLED_PRODUCE`. In Peak Detector2 we can avoid needing to do that by delaying production of the extra debug sample until the `d_found == true` case occurs.

* Peak Detector2

## Testing Done
I verified that the corresponding test still passes, and no longer triggers an error with the strict block executor.

## Checklist

### DIFF
--- a/gr-blocks/lib/peak_detector2_fb_impl.cc
+++ b/gr-blocks/lib/peak_detector2_fb_impl.cc
@@ -84,8 +84,6 @@ int peak_detector2_fb_impl::work(int noutput_items,
     if (d_found == false) {
         for (int i = 0; i < noutput_items; i++) {
             d_avg = d_alpha * iptr[i] + (1.0f - d_alpha) * d_avg;
-            if (output_items.size() == 2)
-                sigout[i] = d_avg;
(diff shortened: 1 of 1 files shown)


willcode:
Not related to this PR: having the debug output once per work function seems hard to document.

PR #7040, opened 2024-01-03
digital: Fix early return logic in Polyphase Clock Sync

argilo:
## Description
I made some experimental changes to the block executor to detect buffer overruns, and found that Polyphase Clock Sync reads uninitialized memory. It also fails to process some input samples.

The Polyphase Clock Sync has an "Output SPS" (`d_osps`) property, which defaults to 1. Its main work loop produces `d_osps` output samples at a time, and it has some special logic to handle the situation where `noutput_items` is not a multiple of `d_osps`, in which case it would split the group of `d_osps` output samples across two `general_work` invocations. This logic is not needed, because the block calls `set_output_multiple`, which guarantees that `noutput_items` will always be a multiple of `d_osps`.

Unfortunately, this logic has an off-by-one error, which causes it to execute on every invocation of `general_work`. This causes the main work loop to exit one iteration early and with `d_out_idx` set to `d_osps`, which causes an output sample to be skipped at the beginning of the following `general_work` invocation. This severely degrades performance at the beginning of each group of samples processed.

By placing a Throttle block just ahead of a Polyphase Clock Sync block and setting its limit to "Max Items: 1", it can be seen that clock synchronization completely fails.

## Related Issue
I suspect that #5948 is due to this bug in Polyphase Clock Sync. The `qa_constellation_receiver` tests use `generic_demod`, which includes a Polyphase Clock Sync block.

* Polyphase Clock Sync

## Testing Done
I used the following flow graph for testing. The lower the Throttle block's "Max Items" parameter, the more distortion becomes apparent in the received QPSK constellation:
pfb_sync_fg

Before:
pfb_sync_before

After:
pfb_sync_after

## Checklist

### DIFF
--- a/gr-digital/lib/pfb_clock_sync_ccf_impl.cc
+++ b/gr-digital/lib/pfb_clock_sync_ccf_impl.cc
@@ -51,8 +51,7 @@ pfb_clock_sync_ccf_impl::pfb_clock_sync_ccf_impl(double sps,
       d_nfilters(filter_size),
       d_max_dev(max_rate_deviation),
       d_osps(osps),
-      d_error(0),
-      d_out_idx(0)
+      d_error(0)
 {
     if (taps.empty())
         throw std::runtime_error("pfb_clock_sync_ccf: please specify a filter.");
@@ -369,8 +368,7 @@ int pfb_clock_sync_ccf_impl::general_work(int noutput_items,
             }
         }
 
-        while (d_out_idx < d_osps) {
-
+        for (int out_idx = 0; out_idx < d_osps; out_idx++) {
             d_filtnum = (int)floor(d_k);
 
             // Keep the current filter number in [0, d_nfilters]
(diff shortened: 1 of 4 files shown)


argilo:
While working on this, I also noticed that Polyphase Clock Sync reads input without any bounds checks. If a signal with large amplitude is passed in, then `d_error`, `d_k`, `d_filtnum` and `count` can all become large, which causes out-of-bounds input samples to be read, resulting in a segfault. This is a separate issue, so I'll leave it to another PR.

daniestevez:
Interesting. I wonder whether any of these bugs are present in the Symbol Sync block as well. When @awalls-cx18 added that block, I think he mentioned that he did some clean up when taking code from Polyphase Clock Sync, and perhaps fixed some bugs as a consequence (but only in Symbol Sync, not in the original Polyphase Clock Sync). This raises the question of why Polyphase Clock Sync is not marked as deprecated (Clock Recovery MM is), since Symbol Sync as intended to be a drop-in replacement with more features and perhaps less bugs.

awalls-cx18:
Regarding the output ssmple stuff:
I did a lot of work with sample rates and clocking in the new symbol sync blocks to ensure correct results.  I won't guarantee "bug free" but you can read my comments in the forecast function here: https://github.com/gnuradio/gnuradio/blob/main/gr-digital/lib/symbol_sync_ff_impl.cc#L410
And look at the general work function.

FWIW, the Symbol Sync blocks are a complete rewrite and do not contain any real borrowing of code from the other blocks, which all had bugs.

The one bug I know of is that tag propagation is only correct for output samples per symbol = 1.  Otherwise it's off.

awalls-cx18:
IMO, incrementally fixing the PFB Clock sync block is a futile endeavor.  The errors in processing under certain conditions won't be eliminated until you eventually converge on an implementation that is equivalent to the Symbol Sync block. (Which is maintaining or tracking 4 separate sample clocks to get the job done.)

argilo:
I opened #7048 to track the missing bounds checks, and #7049 to propose deprecating the block.

PR #7043, opened 2024-01-03
pmt: Check length before calling memcmp

argilo:
## Description
If GNU Radio is built with GCC's Undefined Behaviour Sanitizer, then the following error appears when running qa_pdu_split:

```
/home/argilo/prefix_311/src/gnuradio/gnuradio-runtime/lib/pmt/pmt.cc:889:19: runtime error: null pointer passed as argument 1, which is declared to never be null
/home/argilo/prefix_311/src/gnuradio/gnuradio-runtime/lib/pmt/pmt.cc:889:19: runtime error: null pointer passed as argument 2, which is declared to never be null
```

The line that triggers this is the following:

https://github.com/gnuradio/gnuradio/blob/86af478a88cf0521b023ad2c924c74ea38d9adcf/gr-pdu/python/pdu/qa_pdu_split.py#L96

We can avoid Undefined Behaviour by short-circuiting the `memcmp` if `len_x` is zero.

`pmt::equal`

## Testing Done
I verified that the test passes without UBSAN warnings after the change.

## Checklist
- [ ] ~~I have updated [the documentation](https://wiki.gnuradio.org/index.php/Main_Page#Documentation) where necessary.~`

### DIFF
--- a/gnuradio-runtime/lib/pmt/pmt.cc
+++ b/gnuradio-runtime/lib/pmt/pmt.cc
@@ -886,7 +886,7 @@ bool equal(const pmt_t& x, const pmt_t& y)
         size_t len_x, len_y;
         const void* x_m = xv->uniform_elements(len_x);
         const void* y_m = yv->uniform_elements(len_y);
-        if (memcmp(x_m, y_m, len_x) == 0)
+        if ((len_x == 0) || (memcmp(x_m, y_m, len_x) == 0))
             return true;
 
         return false;


daniestevez (approved):
This looks good to me. On a first read I was wondering what happens in the case where `len_x > 0` but `len_y == 0`, but then I saw that a couple lines above the function returns `false` if the lengths are not equal, so this case can't happen.

PR #7045, opened 2024-01-03
digital: add soft symbol handling to additive scrambler

daniestevez:
## Description

In receivers it is typical to perform additive descrambling with soft symbols before passing these soft symbols to an FEC decoder. For instance, CCSDS protocols work like this.

Currently, the Additive Scrambler block is only able to work with hard symbols, which makes using it to descramble soft symbols rather awkward (but not impossible, since it is possible to feed it zeros, take its output, convert from unipolar to bipolar and then mulitply this by the soft symbols). This commit adds to additive scrambler the capability to work directly with soft symbols.

The additive_scrambler_bb class has been refactored as a template additive_scrambler, which is instantiated for gr_complex, float, int32_t, int16_t and uint8_t. The uint8_t instantiation corresponds to the behaviour of the current additive_scrambler_bb class and is intended to be backwards compatible with no API changes.

The new instantiations all correspond to soft symbols (in different numeric formats). The LFSR output is used to decide whether to invert the sign of each soft symbol or not. For these soft symbols instantiations it is mandatory that bits_per_byte == 1. The constructor will throw an execption otherwise.

A simple QA test for the float instantiation has been added.

The old digital_additive_scrambler_bb GRC block has been marked as deprecated, and a new digital_additive_scrambler_xx GRC block with an option to select the IO Type (defaulting to byte) has been added as a replacement.

## Related Issue

None

Additive Scrambler block, in gr-digital.

## Testing Done

The new QA test passes. Some testing in GRC has been done to check that the new GRC block works correctly.

## Checklist

- [ ] I have updated [the documentation](https://wiki.gnuradio.org/index.php/Main_Page#Documentation) where necessary. -> Updating the [wiki](https://wiki.gnuradio.org/index.php/Additive_Scrambler) is pending on merge.


daniestevez:
I think that this PR doesn't constitute an API change (at least this is the intention), but it would be good that someone reviews this carefully.

willcode:
That looks like a good way to maintain C++/Python API and GRC file backward compatibility. I'll wait until some other people get a chance to look at it though. @marcusmueller maybe?

marcusmueller (approved):
Pretty much the same updating principle I applied for "Throttle" -> "Throttle (old)" / + "Throttle"  (which is the trickle-capable throttle).

Didn't read the rest of the code right now, sorry :(

PR #7053, opened 2024-01-05
analog: Pass through max_gain parameter in agc3_cc::make

### LINKED ISSUE #6955: Power Squelch in combination with AGC3 outputting to a file sink results in files with NaN values.
### What happened?

I have a few flowgraphs that are used as part of an airband streaming setup. It basically works like this:
`[python script generated by grc] --(named pipe)--> [gstreamer] --> [icecast]`
The output of the script is a file sink, which connects to the named pipe.

Usually, I reconfigure them, using grc, whenever ATC decides to switch to alternate frequencies (i.e. they are fixed frequency streams, instead of scanners), and during one of these events, I received reports that the audio was completely muted after the stream was reconfigured.

During my investigation, I discovered that this was caused by an interaction between the AGC3 and the Power Squelch blocks. When the flowgraph outputs to an audio sink (pipewire), no abnormal behavior is observed from the flowgraph and audio is outputted as normal, the squelch silencing the audio during periods of inactivity as expected. However, outputting to a file sink results in files that are full of -NaNs.

If the squelch is permanently opened (by setting too low of a threshold) or disabled, the file output by the flowgraph contains valid audio.

The cause appears to be the following line on the script generated by GRC, related to the instancing of an AGC3 block:
```
self.analog_agc3_xx_0 = analog.agc3_cc((1e-0), (1e-4), 1.5, 0.0025, 1, 65536.0)
```
I noticed that a previous script, generated by an older version of GRC (3.10.4.0), instanced AGC3 in a different way:
```
self.analog_agc3_xx_0 = analog.agc3_cc((1e-0), (1e-4), 1.5, 0.0025, 1)
self.analog_agc3_xx_0.set_max_gain(65536.0)
```
which is how I discovered the cause of this issue. Changing the instancing of AGC3 to this different style results in correct operation of the flowgraph and valid audio is output from the file sink, as expected.

Here is the flowgraph that was used:
[am_rx_121850.grc.gz](https://github.com/gnuradio/gnuradio/files/13356174/am_rx_121850.grc.gz)

### System Information

OS: openSUSE Tumbleweed
GR Installation Method: Package from hardware:sdr repository

### GNU Radio Version

3.10 (maint-3.10)

### Specific Version

3.10.8.0

1. Open attached flowgraph, or any other flowgraph with Power Squelch, AGC3 and file sink, with GRC.
2. Generate python script.
3. Run python script.
4. Open output file in hex editor or audio editing software.
5. File should contain -NaN values only if the squelch was closed during recording.

### Relevant log output

_No response_

argilo:
## Description
In #6624 a `max_gain` parameter was added to AGC3, but `agc3_cc::make` doesn't pass it through to the block constructor. Here I've added the pass-through, and also added a test which verifies that the values make it through.

## Related Issue
Fixes #6955.

* AGC3

## Testing Done
I verified that the test fails without the change, and passes with it.

## Checklist

### DIFF
--- a/gr-analog/lib/agc3_cc_impl.cc
+++ b/gr-analog/lib/agc3_cc_impl.cc
@@ -33,7 +33,7 @@ agc3_cc::sptr agc3_cc::make(float attack_rate,
                             float max_gain)
 {
     return gnuradio::make_block_sptr<agc3_cc_impl>(
-        attack_rate, decay_rate, reference, gain, iir_update_decim);
+        attack_rate, decay_rate, reference, gain, iir_update_decim, max_gain);
 }
 
 agc3_cc_impl::agc3_cc_impl(float attack_rate,
--- a/gr-analog/python/analog/qa_agc.py
+++ b/gr-analog/python/analog/qa_agc.py
@@ -529,6 +529,22 @@ class test_agc(gr_unittest.TestCase):
             self.assertAlmostEqual(x, ref, None,
                                    f"failed at pos {idx} (stride = {stride})", 0.1)
 
+    def test_007_agc3_constructor_arguments(self):
+        attack_rate = 1.1e-3
+        decay_rate = 1.2e-4
+        reference = 1.3
+        gain = 1.4
+        iir_update_decim = 2
+        max_gain = 1.5e4
+
+        agc3 = analog.agc3_cc(attack_rate, decay_rate, reference, gain, iir_update_decim, max_gain)
+
+        self.assertAlmostEqual(agc3.attack_rate(), attack_rate)
+        self.assertAlmostEqual(agc3.decay_rate(), decay_rate)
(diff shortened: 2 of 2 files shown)


marcusmueller:
uff! I really failed at seeing that!

marcusmueller (approved):

argilo:
@willcode It might be worth slipping this into a 3.10.9.2 release so that the regression doesn't get baked into Ubuntu 24.04.

willcode:
@maitbot what would a cutoff date for that be?

argilo:
Ubuntu 24.04 has a Debian import freeze on February 29:

https://discourse.ubuntu.com/t/noble-numbat-release-schedule/35649

willcode:
OK, I can put out 3.10.9.2 in the next couple of weeks. Might as well wait to see if anything else should go in.

PR #7073, opened 2024-01-08
runtime: use XDG_CONFIG_DIR for the userconf path

### LINKED ISSUE #3631: Configuration files should be stored following the XDG Base Directory Specification
For some reason gnuradio creates a folder `~/.gnuradio` and uses it for configuration. That's not in-line with [XDG Base Directory Specification](https://wiki.debian.org/XDGBaseDirectorySpecification) and clutters people's home directories.

gnuradio should follow XDG base directory specification and use `~/.config/gnuradio` for configuration files

mboersch:
This changes the config path from ~/.gnuradio to ~/.config/gnuradio. On windows this maps to %APPDATA%/.config/gnuradio.
This was requested in #3631. 

## Description

Adhering to the XDG specification help reduce clutter of custom dotfile locations in $HOME...

## Related Issue

Fixes #3631

runtime

## Testing Done

CMake build and ctest.

## Checklist

Not sure how I should document this or where to even start. Dito testing :-)

### DIFF
--- a/gnuradio-runtime/include/gnuradio/sys_paths.h
+++ b/gnuradio-runtime/include/gnuradio/sys_paths.h
@@ -11,17 +11,18 @@
 #define GR_SYS_PATHS_H
 
 #include <gnuradio/api.h>
+#include <filesystem>
 
 namespace gr {
 
 //! directory to create temporary files
-GR_RUNTIME_API const char* tmp_path();
+GR_RUNTIME_API std::filesystem::path tmp_path();
 
 //! directory to store application data
-GR_RUNTIME_API const char* appdata_path();
+GR_RUNTIME_API std::filesystem::path appdata_path();
 
 //! directory to store user configuration
-GR_RUNTIME_API const char* userconf_path();
+GR_RUNTIME_API std::filesystem::path userconf_path();
 
 } /* namespace gr */
 
--- a/gnuradio-runtime/lib/prefs.cc
+++ b/gnuradio-runtime/lib/prefs.cc
@@ -65,7 +65,7 @@ std::vector<std::string> prefs::_sys_prefs_filenames()
     // Find if there is a ~/.gnuradio/config.conf file and add this to
     // the end of the file list to override any preferences in the
     // installed path config files.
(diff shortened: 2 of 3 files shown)


willcode:
I like the idea of having the GR conf file use the system scheme. We've got an awful lot of people out there with the conf file in the current location, so I feel like some sort of transition is needed. That might involve the old location having priority, or copying the file to the new location (and removing the old one?) or something else. Then there's the user "where did my file go?" problem. Needs some thought.

marcusmueller:
@mboersch  would you please run clang-format on your changed files?

@willcode I think for `main` a change where ~/.gnuradio doesn't have priority (but is still found, if XDG_CONFIG_HOME doesn't contain config) is OK, even desirable!

The backport to maint-3.10 (which sadly is explicitly necessary, if we want to do it, because these functions are public API…) would probably want to not do anything "automatically surprising" about the file(s)'s location, fully agreed. But we could print a nice log entry, and I wouldn't mind adding a GRC button that does move the configuration, abusing, say, the "Help" menu.

willcode:
Things that are `main`-only may never see the light of day. Don't want to waste peoples' time. Also, this is an area where different behavior between `main` and releases could really confuse people who run multiple versions.

mboersch:
> @mboersch would you please run clang-format on your changed files?

Done. 

Some backwards compatibility seems reasonable -- even though that code would probably live forever.

I can try add a lookup to the old `.gnuradio` to the lookup list for the userconf path. 
I'm not sure if I should put the lookup / filesystem checks into the sys_path.cc -- seems a bit odd to have the PATH code do filesystem checks.
FWIW, introducing a `xdg_userconf_path` and a `legacy_userconf_path` there, and have `userconf_path` check if either directory exists may work. Should I go for it?

willcode:
Marked with "API Change" so we track that we need to think of a way to make this backward compatible.

willcode:
Setting as Draft until backward compatibility is worked out.

marcusmueller:
merged through #7136

PR #7078, opened 2024-01-10
cmake: Stop installing byte-compiled Python modules

### LINKED ISSUE #2944: cmake: Python byte-compilation always uses legacy format filenames
e.g.:
```
/usr/lib/python3.7/site-packages/gnuradio/fft/fft_swig.py
/usr/lib/python3.7/site-packages/gnuradio/fft/fft_swig.pyc
/usr/lib/python3.7/site-packages/gnuradio/fft/fft_swig.pyo
```

I think on >= Python 3.2
https://docs.python.org/3/whatsnew/3.2.html#pep-3147-pyc-repository-directories
these compiled files will be ignored because the source file exists:
https://www.python.org/dev/peps/pep-3147/#case-4-legacy-pyc-files-and-source-less-imports

For packaging on Linux it could even be easier if they weren't compiled via cmake.

cjmayo:
Even with current versions of Python old-style .pyc and .pyo files are being created because GrPython.cmake is generating the byte-compiled filenames itself and forcing py_compile.compile() to create them using the cfile parameter.

Since Python 3.2 [1] .pyc files are ignored when in the same directory as the .py file [2].
Since Python 3.5 .pyo files have been eliminated [3].

[1] https://docs.python.org/3.2/whatsnew/3.2.html#pep-3147-pyc-repository-directories
[2] https://www.python.org/dev/peps/pep-3147/#case-4-legacy-pyc-files-and-source-less-imports
[3] https://docs.python.org/3.5/whatsnew/3.5.html#whatsnew-pep-488

---

Fixes #2944

I've tested that this installs only .py files.

### DIFF
--- a/cmake/Modules/GrPython.cmake
+++ b/cmake/Modules/GrPython.cmake
@@ -213,58 +213,50 @@ function(GR_PYTHON_INSTALL)
             DESTINATION ${GR_PYTHON_INSTALL_DESTINATION}
             ${GR_PYTHON_INSTALL_UNPARSED_ARGUMENTS})
 
-        #create a list of all generated files
-        unset(pysrcfiles)
-        unset(pycfiles)
-        unset(pyofiles)
-        foreach(pyfile ${GR_PYTHON_INSTALL_FILES})
-            get_filename_component(pyfile ${pyfile} ABSOLUTE)
-            list(APPEND pysrcfiles ${pyfile})
+        if(ENABLE_TESTING)
+            # pyc files are generated in the build directory to support testing
 
-            #determine if this file is in the source or binary directory
-            file(RELATIVE_PATH source_rel_path ${CMAKE_CURRENT_SOURCE_DIR} ${pyfile})
-            string(LENGTH "${source_rel_path}" source_rel_path_len)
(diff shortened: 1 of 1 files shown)


willcode:
This makes sense. On installs where the user has write privilege to the install directory, the `__pycache__` logic already has precedence over the pyc files. I'll let someone else check the cmake logic to make sure the install list is still generated correctly and nothing else breaks.

willcode:
That would be a reasonable way to go. Since the Python source is spread across modules, and we don't want to modify the source directories, I can't think of a better way.

willcode:
Now that I think of it, though, generating the pyc files unconditionally tests that the Python code passes a syntax check. The format checker may help with that, but actually "compiling" the files is a good check. So, they just don't need to be copied for install.

willcode:
@ryanvolz would you mind taking a look, since this is cmake stuff?

ryanvolz (approved):
Looks good to me. It also formalizes a patch that I've carried in the conda-forge package since its inception, and it does it the nice way of actually cleaning code up instead of the hacky delete-the-install-lines way that I did it.

PR #7079, opened 2024-01-10
grc: Only bussify blocks that have ports

argilo:
## Description
The "More -> Toggle Source Bus" and "More -> Toggle Sink Bus" options in GRC's context menu are enabled whenever any blocks are selected, regardless whether those blocks actually have any source or sink ports. Selecting "Toggle Source Bus" for a block that doesn't have any source ports will create a useless bus of length zero, and the same for "Toggle Sink Bus".

Screenshot from 2024-01-10 15-14-26

To fix this, I've changed the context menu so that "Toggle Source Bus" is only enabled if at least one of the selected blocks has a source port, and "Toggle Sink Bus" is only enabled if at least one of the selected blocks has a sink port.

Since not all of the selected blocks will necessarily have the required ports, I also changed `Block.bussify` to check whether the current block has at least one port of the required type before converting to a bus.

Bus ports in GRC

## Testing Done
I manually tested GRC, selecting one or more blocks, checking whether the correct context menu items were enabled, and clicking the menu items to see whether ports are correctly bussified.

## Checklist

### DIFF
--- a/grc/core/blocks/block.py
+++ b/grc/core/blocks/block.py
@@ -750,14 +750,16 @@ class Block(Element):
     def bussify(self, direc):
         if direc == 'source':
             ports = self.sources
-            ports_gui = self.filter_bus_port(self.sources)
-            self.bus_structure = self.get_bus_structure('source')
-            self.bus_source = not self.bus_source
+            if ports:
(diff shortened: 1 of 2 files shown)


willcode:
Looks like a simple solution.

PR #7090, opened 2024-01-23
GRC: fix C++ hier block generation

AsriFox:
## Description
This PR fixes two issues I found when generating C++ hier blocks with `grcc`:
1. The `--output` argument had no effect on location of generated files; they were always put in the `$HOME/.grc_gnuradio` directory.
2. The generated `make` function (that creates a shared pointer) was not using any parameters as arguments; that resulted in a compile error, because the generated class constructor was using these parameters as arguments.

Note: new `make` function template does not use default values for arguments yet.

GRC generation / `grcc`

## Testing Done
Minimal example: [cpp_hier.grc](https://github.com/gnuradio/gnuradio/files/14021893/cpp_hier.grc.txt);  
Command used: `grcc cpp_hier.grc -o ./output_dir`

## Checklist

### DIFF
--- a/grc/core/generator/cpp_hier_block.py
+++ b/grc/core/generator/cpp_hier_block.py
@@ -12,24 +12,22 @@ from ..io import yaml
 class CppHierBlockGenerator(CppTopBlockGenerator):
     """Extends the top block generator to also generate a block YML file"""
 
-    def __init__(self, flow_graph, file_path):
+    def __init__(self, flow_graph, output_dir):
         """
         Initialize the hier block generator object.
 
         Args:
             flow_graph: the flow graph object
-            file_path: where to write the py file (the yml goes into HIER_BLOCK_LIB_DIR)
+            output_dir: the path for written files
         """
-        CppTopBlockGenerator.__init__(self, flow_graph, file_path)
         platform = flow_graph.parent
+        if output_dir is None:
+            output_dir = platform.config.hier_block_lib_dir
+        if not os.path.exists(output_dir):
+            os.mkdir(output_dir)
 
-        hier_block_lib_dir = platform.config.hier_block_lib_dir
-        if not os.path.exists(hier_block_lib_dir):
-            os.mkdir(hier_block_lib_dir)
-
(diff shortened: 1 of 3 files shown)


willcode:
`make()` with out default arguments - are you talking about the top block `make(param_str)` at the bottom? If the PR is not done, could you mark it draft?

AsriFox:
When a block is created using `gr_modtool`, in its public header a `static sptr make()` function is generated with arguments sourced from the `--argument-list` string; so the user is responsible for specifying valid argument names, types and default values.
When the hier block is generated from GRC, however, all arguments correspond to Parameter blocks in the flowgraph. These do not have a property like 'optional', so I assumed all parameters should be required. I can change this by adding ${param.templates.render('make')}, if needed.
There is no easy way to translate Python expressions to C++, however, so sometimes rendering this will result in compile errors. Only numbers and double-quoted strings will work normally.

willcode:
Do things other than numbers and double-quoted strings work in GRC with/without this PR? Booleans, lists, etc.

I think you are correct that all specified parameters are provided. However, I'm just assuming that optional parameters are not required in GRC.

willcode:
@haakov could you take a look at this? Just noticed it's your code, so I will stop trying to figure out how this all works.

AsriFox:
Tried changing the `make` [template](https://github.com/AsriFox/gnuradio/blob/97727aee79c5c3ed6db00b595002faa7d2e01838/grc/core/generator/cpp_templates/flow_graph.hpp.mako#L88) to this:
```
static sptr make(${", ".join((param.vtype + " " + param.name + " = " + param.cpp_templates.render('make')) for param in parameters)});
```
And it breaks when using 'complex' (e.g. `1+1j`).

Perhaps, expressions should be evaluated (or converted) before pasting them in the template, like it's done for 'var_make' of vtype 'complex'.

willcode:
It sounds like the param fix is better than the way things are currently, but that it may not be complete. We'll put this on main but hold the backport until people get to try it.

PR #7092, opened 2024-01-23
grcqt: Handle mainwindow closeEvent

dl1ksv:
## Description

At the moment the closeEvent (clicking the X in the right upper corner) is handled by qt, so no changes in the settings - especially the geometry -- are not written to grc.conf.
Redirecting closeEvent exit_triggered solves this issue.
Setting object names for toolbar objects avoids warnings on close.

## Checklist


haakov:
Great, thank you for the PR!

marcusmueller:
oh that QT handles this properly already might be really nice; need to test later whether this enables something else:

In gtk-grc, you can't do things like run GRC with/in the python profiler module, because the profiling information post-processing happens after regular application shutdown, and gtk-GRC just sys.exits from random points, and never actually properly lets the python process finish.

PR #7105, opened 2024-01-30
Added ccache speedup compilation

GermanAizek:
## Description
More effective helped after compilation checkout branches.
More information on https://ccache.dev

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -75,6 +75,13 @@ if(NOT DEFINED BOOST_ROOT)
     set(BOOST_ROOT "${CMAKE_INSTALL_PREFIX}")
 endif()
 
+find_program(CCACHE "ccache")
+if(CCACHE)
+    set(CMAKE_C_COMPILER_LAUNCHER ${CCACHE})
+    set(CMAKE_CXX_COMPILER_LAUNCHER ${CCACHE})
+    set(ENV{CCACHE_SLOPPINESS} pch_defines,time_macros)
+endif(CCACHE)
+
 ########################################################################
 # Import executables from a native build (for cross compiling)
 # http://www.vtk.org/Wiki/CMake_Cross_Compiling#Using_executables_in_the_build_created_during_the_build


willcode:
Please sign your commit (`git commit -S --amend`) and force push, to acknowledge the DCO (contributor agreement).

willcode:
What kind of speedup did you see? Was there any speedup on an initial build, or only on subsequent builds?

GermanAizek:
> What kind of speedup did you see? Was there any speedup on an initial build, or only on subsequent builds?

Yep, more effective recompilation after checkout branches.

willcode:
How much speedup?

Our CI does not have ccache installed, based on the cmake output in the logs. It wouldn't have any effect on CI anyway since that's a single compilation.

marcusmueller:
on most Linux distros where ccache is installed, it does get used automatically:

```shell
git clone https://github.com/gnuradio/gnuradio
mkdir gnuradio/build-defaultconfig
cd gnuradio/build-defaultconfig
cmake ..
grep CMAKE_CXX_COMPILER CMakeCache.txt
```
yields
```cmake
CMAKE_CXX_COMPILER:FILEPATH=/usr/lib64/ccache/c++
CMAKE_CXX_COMPILER_AR:FILEPATH=/usr/bin/gcc-ar
CMAKE_CXX_COMPILER_RANLIB:FILEPATH=/usr/bin/gcc-ranlib
//ADVANCED property for variable: CMAKE_CXX_COMPILER
CMAKE_CXX_COMPILER-ADVANCED:INTERNAL=1
//ADVANCED property for variable: CMAKE_CXX_COMPILER_AR
CMAKE_CXX_COMPILER_AR-ADVANCED:INTERNAL=1
//ADVANCED property for variable: CMAKE_CXX_COMPILER_RANLIB
CMAKE_CXX_COMPILER_RANLIB-ADVANCED:INTERNAL=1
```
without any further ado.

You also want to have a definitive way to disable it, as well. I'm not sure on which systems this is actually needed, so I'm kind of sceptical it's generally a good idea; we'd really need a killswitch to disable it in the face of compiler and linker trouble, or when you just don't want to use ccache, because you know this is going to be a modified one-off build.

marcusmueller:
By the way, @willcode:

> Our CI does not have ccache installed, based on the cmake output in the logs. It wouldn't have any effect on CI anyway since that's a single compilation.

Definitely on my to-do list, we could (after adding ccache to the CI builders) have a CI run that we execute once per push to maint-3.10 or main, which then gets pushed as artifact; all the regular CI runs would first download and extract the ccache before running the build.

willcode:
I've seen it get picked up automatically in the past, too. @GermanAizek ccache is not used automatically on your distro? Which one is that?

I actually stopped using ccache when SSDs first came out and were pretty small, since the cache was always using up a lot of space. Haven't tried it again recently.

mbr0wn:
I have been also using `ccache` for almost a decade now with no CMake mods required. I'm not sure how this changeset will work for cases where `ccache` is available, but not desired (currently, you can simply do `CXX=/usr/bin/g++ cmake .....`). @GermanAizek what's your dev setup? Fedora/Ubuntu users certainly won't need this, but what system are you on? Maybe we're making things unnecessarily complicated for some OSs.

willcode:
Unclear whether this is needed on most systems, and DCO (git signoff) missing, so changing to draft for now.

GermanAizek:
> @GermanAizek what's your dev setup? Fedora/Ubuntu users certainly won't need this, but what system are you on? Maybe we're making things unnecessarily complicated for some OSs.

Debian and FreeBSD

marcusmueller:
Debian? Interesting! So, we're now a bit on a crossroads: 

1. We change nothing, and ccache is opt-in on debian, and opt-out on Fedora, Ubuntu, it seems, through setting `-DCMAKE_CXX_COMPILER_LAUNCHER` (to empty string to opt out and to `ccache` to opt in)
2. We generally adopt the direction of the PR, and fix the PR to at least honor an already defined `CMAKE_CXX_COMPILER_LAUNCHER`, even and especially when set to empty, thus making `ccache` opt-out.

In both cases the current form of the PR isn't going to get us where we want!

marcusmueller (changes requested):
Sorry, won't do; this needs to respect an already defined `CMAKE_{LANGUAGE}_COMPILER_LAUNCHER`

mbr0wn:
This looks like it won't make it in -- closing PR. Thanks for submitting and sparking the conversation!

PR #7135, opened 2024-02-14
uhd/rfnoc python bindings: remove unused hashes from headers

marcusmueller:
This avoids changes when the binding-hash-hotfixer is run.

## Description

Although an unused hash is not a cmake-time error, it still is a bit superfluous, and it becomes an unnecessary change when someone runs binding-hash-hotfixer.zsh to check which binding files need fixing.

only gr-uhd

## Testing Done

-/-

## Checklist


willcode:
What makes these hashes unused, and why would the `BINDTOOL_HEADER_FILE` be there if it was unused?

Issue #1010, opened 2016-09-17
No appropriate QA test for gr::prefs / config file parser


mbr0wn:
There's no QA for gr::prefs _at all_.

Being core functionality, http://gnuradio.org/redmine/issues/918 will have to wait until I implement one.

marcusmueller:
> There's no QA for gr::prefs _at all_.

Not true:
```
    def test_001(self):
        p = gr.prefs()

        # Read some options
        self.assertFalse(p.has_option('doesnt', 'exist'))

        # At the time these tests are run, there is not necessarily a default
        # configuration on the build system, so not much to do with testing here
```

But truth be told, that is not *sufficient* testing.

marcusmueller:
This being a really well-stressed multi-platform-problematic core component, lack of testing *does* constitute a bug, since we'll not notice failures.

There's been at least one acute instance of this biting us (namely, 3.7 < 3.7.13 can't change audio settings since the prefs file handling was broken).

marcusmueller:
remains valid, should be sensibly fixable, long-term maintainability issue.

PR #7136, opened 2024-02-15
sys_paths: use XDG_CONFIG_HOME if present, new API

### LINKED ISSUE #2020: Quadrature Demod: missing FM input end-to-end QA test
So, we need a unit test with FM modulated signals getting decoded correctly in a zero-noise scenario.

extend gr-analog/python/analog/qa_quadrature_demod.py:

1. add a new `def test_fm_end_to_end(self):`
2. in that, use the `analog.frequency_modulator_fc` block with, 
3. implement 
```text 
signal source (cos 1000 Hz) +-> vector_sink "A"
                            |
                            \-> analog.frequency_modulator_fc -> analog.quadrature_demod_cf -> vector_sink "B"
```
4. verify that, first samples aside, A.data() is equal to B.data() after running

### LINKED ISSUE #3631: Configuration files should be stored following the XDG Base Directory Specification

marcusmueller:
## Description
- Introduced `gr::paths::` namespace to keep the different path getters
  together
  - new `cache()` getter, because FFTW wisdom should not clutter the
    home directory, but is also not portable state nor portable config
    (compare
    [`XDG_CACHE_HOME`](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html))
- new path getters return `std::filesystem::path` instead of C string
- deprecated old getters, which now just call the new getters underneath
- Python wrappers
- added a unit test for the paths (was untested so far)
- added prefs path unit test for checking reaction to changing
  environment variables (prefs was completely untested)
- Fixed `vmcircbuf_prefs`, fftw wisdom that assumed existence of paths
- Replace deprecated API from in-tree usage
- As some configuration files move to subdirectories of directories not
  guaranteed to exist, create directories as needed

In this new configuration, `gr::paths::userconf()` (and the legacy,
`gr::userconf_path()`) tries to find the user configuration in the
following order:

1. `$GR_PREFS_PATH` if that environment variable is set
2. `$XDG_CONFIG_HOME/gnuradio`; if `$XDG_CONFIG_HOME` is set, and
   `$XDG_CONFIG_HOME/gnuradio` exists and is a directory
3. `appdata()/.config/gnuradio` (`appdata ~= $HOME`) if that exists and
   is a directory
4. `appdata()/.gnuradio` if that exists and is a directory
5. `$XDG_CONFIG_HOME/gnuradio` if `$XDG_CONFIG_HOME` is set
6. `appdata()/.config/gnuradio`

So, in other words: unless explicitly set to something else, try the new
XDG_CONFIG_HOME path, if that doesn't exist, try the old path, and if
neither exists, use the new one.
## Related Issue

Supersedes #7073

(but still uses mboersch code, so copyright notices include mboersch)

Closes #3631

Looks at #1010, but makes only minor progress

- `gr::userconf_path()`: deprecated
- `gr::tmp_path()`: deprecated
- `gr::appdata_path()`: deprecated
- `gr::fft::fft`: stores wisdom no longer in `~/.gr_fftw_wisdom` but in
  `~/.cache/gnuradio/fftw_wisdom`
- `vmcircbuf_prefs`: instead of hard-coding `appdata()/.gnuradio/prefs/key`,
  uses `userconf()/prefs/key` (which is a bugfix, actually)

## Testing Done

- wrote a sys path test
- wrote prefs tests (long overdue)

## Checklist

## TODO

After merge of this: 

- Document on wiki (@duggabe will like that)

### DIFF
--- a/gnuradio-runtime/include/gnuradio/sys_paths.h
+++ b/gnuradio-runtime/include/gnuradio/sys_paths.h
@@ -1,5 +1,6 @@
 /*
  * Copyright 2011,2013 Free Software Foundation, Inc.
+ * Copyright 2024 mboersch, Marcus Müller
  *
  * This file is part of GNU Radio
  *
@@ -10,18 +11,51 @@
 #ifndef GR_SYS_PATHS_H
 #define GR_SYS_PATHS_H
 
+#include "api.h"
 #include <gnuradio/api.h>
+#include <filesystem>
 
 namespace gr {
+namespace paths {
+/*! \brief directory to create temporary files.
+ *
+ * On UNIX-oid systems, typically /tmp.
+ */
+GR_RUNTIME_API std::filesystem::path tmp();
+
+/*! \brief directory that stores user data; typicall $HOME
+ */
+GR_RUNTIME_API std::filesystem::path appdata();
+
+/*! \brief directory that stores configuration.
+ *
+ * Defaults to $XDG_CONFIG_HOME/gnuradio (fallback: appdata()/.config/gnuradio), but if
+ * that doesn't exist, checks the legacy path, appdata()/.gnuradio
+ */
+GR_RUNTIME_API std::filesystem::path userconf();
+
+/*! \brief directory to store non-portable caches (e.g. FFTW wisdom)
+ *
+ * Defaults to $XDG_CACHE_HOME, falls back to appdata()/cache
+ */
+GR_RUNTIME_API std::filesystem::path cache();
+
+/*! \brief directory to store persistent application state (e.g. window layouts, generated
+ *          GRC hier blocks)
+ */
+/*
+GR_RUNTIME_API std::filesystem::path persistent();
+*/
+} /* namespace  paths */
 
 //! directory to create temporary files
-GR_RUNTIME_API const char* tmp_path();
+[[deprecated("use gr::paths::tmp()")]] GR_RUNTIME_API const char* tmp_path();
 
 //! directory to store application data
-GR_RUNTIME_API const char* appdata_path();
+[[deprecated("use gr::paths::appdata()")]] GR_RUNTIME_API const char* appdata_path();
 
 //! directory to store user configuration
-GR_RUNTIME_API const char* userconf_path();
+[[deprecated("use gr::paths::userconf()")]] GR_RUNTIME_API const char* userconf_path();
 
 } /* namespace gr */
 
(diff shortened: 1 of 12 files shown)


marcusmueller:
@willcode  no API change, only ABI-irrelevant annotations, and additions: designed for backportability.

marcusmueller:
really hoping this works on other platforms as well!

marcusmueller:
oha.

so, situation:

- because the parent directory doesn't inherently exist, need to create ~/.config/gnuradio recursively
- Windows is fun: https://en.cppreference.com/w/cpp/filesystem/path/generic_string ; need to `path::string().c_str()` to get the path in the legacy functions. Oh well.

marcusmueller:
TODO: check whether CI runs all fail during testing or whether something strange is going on; some fail surprisingly fast.

gnuradio-runtime/include/gnuradio/sys_paths.h
```diff
@@ -10,18 +11,51 @@
...
+*/
+} /* namespace  paths */
 
 //! directory to create temporary files
-GR_RUNTIME_API const char* tmp_path();
+[[deprecated("use gr::paths::tmp")]] GR_RUNTIME_API const char* tmp_path();
 
 //! directory to store application data
-GR_RUNTIME_API const char* appdata_path();
+[[deprecated("use gr::paths::tmp")]] GR_RUNTIME_API const char* appdata_path();
```
willcode, line 55: Deprecation note on the last 2 - cut/paste.

gnuradio-runtime/include/gnuradio/sys_paths.h
```diff
@@ -10,18 +11,51 @@
...
 #include <gnuradio/api.h>
+#include <filesystem>
 
 namespace gr {
+namespace paths {
+/*! \brief directory to create temporary files.
+ *
+ * On UNIX-oid systems, typically /tmp.
+ */
+GR_RUNTIME_API std::filesystem::path tmp();
```
willcode, line 24: Does the name change provide enough value to bother people with deprecation warnings?

marcusmueller: yes. The main point is that the return value changes, from a manually (actually, leaking or use after free or concurrently modified in the current API) `const char*` to a `std::filesystem::path`.

willcode:
F39 CI is failing possibly due to the gr-fft portion of this fix. Hold backport until we figure it out.

willcode:
Backport: add #7181

PR #7145, opened 2024-02-21
FFT Window Threadsafety

### LINKED ISSUE #7144: FFT block: thread-unsafe window setting
### What happened?

On the mailing list, someone asked how to update the window of the FFT block at runtime. I was about to just quickly add that callback to the block yaml, but then checked:

None of that setter protects against changing the window while it's being used, including resizing, unsetting it alltogether.

### System Information

n/a

### GNU Radio Version

3.11-git (main)

### Specific Version

v3.11.0.0git-657-g039f0b89a

n/a

### Relevant log output

_No response_

marcusmueller:
## Description

I add locks around window setting/usage.

As a drive-by improvement, let's not use floor, ceil to implement (inherent) integer division semantics.

## Related Issue

Fixes #7144

FFT 

## Checklist

### DIFF
--- a/gr-fft/python/fft/qa_fft.py
+++ b/gr-fft/python/fft/qa_fft.py
@@ -133,7 +133,7 @@ class test_fft(gr_unittest.TestCase):
         self.assertComplexTuplesAlmostEqual2(expected_result, result_data,
                                              abs_eps=1e-9, rel_eps=4e-4)
 
-    def test_forward(self):
+    def test_101_forward(self):
         src_data = tuple([complex(primes[2 * i], primes[2 * i + 1])
                           for i in range(self.fft_size)])
         expected_result = primes_transformed
@@ -148,7 +148,7 @@ class test_fft(gr_unittest.TestCase):
         result_data = dst.data()
         self.assert_fft_ok2(expected_result, result_data)
 
-    def test_reverse(self):
+    def test_102_reverse(self):
         src_data = tuple([x / self.fft_size for x in primes_transformed])
         expected_result = tuple(
             [complex(primes[2 * i], primes[2 * i + 1]) for i in range(self.fft_size)])
@@ -163,7 +163,7 @@ class test_fft(gr_unittest.TestCase):
         result_data = dst.data()
         self.assert_fft_ok2(expected_result, result_data)
 
-    def test_multithreaded(self):
+    def test_201_multithreaded(self):
         # Same test as above, only use 2 threads
         src_data = tuple([x / self.fft_size for x in primes_transformed])
         expected_result = tuple(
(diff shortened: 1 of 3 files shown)


daniestevez (commented):
The code looks good to me, but performance-wise I'm wondering if it wouldn't be better to get the window lock at the beginning of the `work()` function and hold it for the whole duration of `work()`. I'm slightly worried that if the FFT size is small, locking and unlocking the mutex for each individual FFT might introduce too much overhead.

willcode:
CI stopped on Python formatting:
`/gr-fft/python/fft/qa_fft.py:288:76: E226 missing whitespace around arithmetic operator`

gr-fft/python/fft/qa_fft.py
```diff
@@ -273,6 +273,20 @@ def test_reverse_window_shift(self):
...
 
+    def _set_length_and_test(self, fftlength: int, windowlength: int) -> bool:
+        win = [1.0] * windowlength
+        for shift in (True, False):
+            fft_block = fft.fft_vcc(fftlength, forward=True, window=[], shift=shift)
+            return fft_block.set_window(win)
+
+    def test_303_test_window_setter(self):
+        """
+        Teqt whether the window setter correctly returns False for lenghts != FFT length.
```
mbr0wn, line 284:
```suggestion
        Test whether the window setter correctly returns False for lengths != FFT length.
```

mbr0wn (commented):
I tend to agree with @daniestevez, although I don't know about the performance overhead. But what if we flip it around: If we unlock sooner, we get better concurrency. To which I say "so what", what's the value of being able to call `set_window()` more often from a different thread. So if we lock for the duration of `work()`, we get maybe some, maybe very little performance gains, but we lose nothing of value.

daniestevez:
@mbr0wn that's exactly what I had in mind. `set_window()` will typically be called infrequently, and it can wait until `work()` finishes to update the window.

marcusmueller:
I'm fine either way, but I remember I had to be finer in locking granularity in one place, because otherwise I could deadlock our CI, then made it consistent in every place. Will have to check what the precise reason was, and whether it affects work, probably not. Anyways, currently no time, so this becomes a draft for now.

gr-fft/python/fft/qa_fft.py
```diff
@@ -273,6 +273,20 @@ def test_reverse_window_shift(self):
...
+            fft_block = fft.fft_vcc(fftlength, forward=True, window=[], shift=shift)
+            return fft_block.set_window(win)
+
+    def test_303_test_window_setter(self):
+        """
+        Test whether the window setter correctly returns False for lengths != FFT length.
+        """
+        for fftlength in [1, 2, 3] + [2**i for i in range(2, 10)] + list(primes):
+            self.assertTrue(self._set_length_and_test(fftlength, fftlength))
+            self.assertFalse(self._set_length_and_test(fftlength, fftlength+2))
```
StudHamza, line 288: Not sure if will solve `./gr-fft/python/fft/qa_fft.py:288:76: E226 missing whitespace around arithmetic operator` but probability it will, according to [E226](https://www.flake8rules.com/rules/E226.html)
```suggestion
            self.assertFalse(self._set_length_and_test(fftlength, fftlength + 2  ))
```

marcusmueller: that will definitely cause another error due to superfluous whitespace *before* the `)`, @StudHamza  ;)

PR #7147, opened 2024-02-23
grc: add option to show properties field colors

daniestevez:
## Description

In #6888 the colors that indicate the type of each field in the block properties were replaced by labels, because colors can give contrast problems, specially in dark mode. I personally find colors useful, because they help me to identify quickly the field that I want to edit (perhaps I'm just too used to having colors).

This commit adds the field colors back, but only as an optional feature controlled by an entry in the View menu. This setting is saved to the GRC preferences, and it defaults to colors disabled. Users that prefer colors can simply enable the old behavior in this menu entry.

The new labels to the right of each field that indicate the field type are always shown, regardless of whether field colors are enabled. Even when colors are shown, the labels can be a useful reminder of what type each color stands for, and they don't use too much screen real state.

----

**Rationale:** call me old fashioned, change adverse, or whatever, and you'd be right. After a few days of using GNU Radio 3.10.9.2 I've found that the lack of colors in the block properties makes GRC harder to use for me. It turns out that when I'm editing a flowgraph quickly, I often find the field that I want to edit because I know its general location and color (example: if I want to change the number of inputs in the QT GUI Frequency Sink, I quickly search for a green field that is midway in the properties). I wasn't aware that I was doing this until I've started to use GRC without colors. The feeling is similar to reading code without syntax highlighting: the code looks harder to read and duller. I think I'm just too used to having colors in GRC.

I get that there are many reasons why colors can be problematic, but perhaps other users are like me and prefer to have colors. This PR adds the colors back as an optional feature that is disabled by default. The code changes aren't much, so in this way I think we can keep everyone happy.

GRC

## Testing Done

Run GRC and checked that the new entry in the View menu does indeed enable or disable the colors of the block properties. Also tested that this setting is successfully saved to the preferences and available in the next start of GRC. I haven't tested what happens if I click the View menu entry while a block properties window is open because my window manager doesn't let me click on the GRC main window while a block properties window is open.

## Checklist

- [ ] I have updated [the documentation](https://wiki.gnuradio.org/index.php/Main_Page#Documentation) where necessary. --> Probably no documentation changes are needed for this change?
- [x] I have added tests to cover my changes, and all previous tests pass. -> No tests added to cover this change. All the tests pass.


daniestevez:
Tagging @cafeclimber FYI. For some reason Github won't let me add her as a reviewer, which is what I intended to do.

willcode:
The colors in dark mode were just plain awful, so I think if we add colors back in we also have to work on readability and make it look nice.

daniestevez:
I agree that colors in dark mode were bad, but in light mode they're okay. I don't have a good solution for a nice color scheme for dark mode. I saw other people trying, but the attempts weren't too good and the PRs weren't merged. That is why I propose to merge this PR as is. People that are using light mode and want colors can enable it. Waiting until we have a usable solution for colors in dark mode means waiting for an undetermined amount of time.

willcode:
@haakov what would it take to port this to qt? I don't want to merge anything related to grc until qt merges, unless you think it's super easy.

PR #7151, opened 2024-02-25
modtool: New manifest format

haakov:
This PR contains some suggested changes to the OOT Manifest format for increased machine readability:
- Use YAML, move the Markdown description to a YAML field
- Add OOT version as a field
- Add manifest file version as a field
- Add YAML schema

Related PR:
https://github.com/gnuradio/gnuradio/pull/7134


gr-utils/modtool/templates/gr-newmod/MANIFEST.yml
```diff
@@ -0,0 +1,20 @@
...
+version: 1.0
+brief: Short description of gr-howto
+tags: # Tags are arbitrary, but look at CGRAN what other authors are using
+  - sdr
+author:
+  - Author Name <authors@email.address>
+copyright_owner:
+  - Copyright Owner 1
+license:
+gr_supported_version:
```
haakov, line 11: I also think that the list fields should have plural names. Thoughts?

haakov:
Force-pushed formatting fix.

PR #7181, opened 2024-03-10
fft: change path calculation for fftw_wisdom to avoid compiler warning

willcode:


marcusmueller (approved):
I think we're both equally confused as to *why* this is necessary, but this seems sensible

PR #7192, opened 2024-03-14
grc: formatting for qt-gui

willcode:
Ran autopep and checked with same pycodestyle line as used in CI.


grc/core/blocks/__init__.py
```diff
@@ -5,6 +5,9 @@
 #
 
 
+from .virtual import VirtualSink, VirtualSource
+from .embedded_python import EPyBlock, EPyModule
+from .dummy import DummyBlock
```
haakov, line 10: This change breaks all GRC tests on my end:
```
______________________________________________________________ ERROR collecting test_block_flags.py ______________________________________________________________
ImportError while importing test module '/home/hkon/qt_gr/grc/tests/test_block_flags.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
/usr/lib/python3.11/importlib/__init__.py:126: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
test_block_flags.py:2: in <module>
    from grc.core.blocks._flags import Flags
../core/blocks/__init__.py:16: in <module>
    from .virtual import VirtualSink, VirtualSource
../core/blocks/virtual.py:10: in <module>
    from . import Block, register_build_in
E   ImportError: cannot import name 'register_build_in' from partially initialized module 'grc.core.blocks' (most likely due to a circular import) (/home/hkon/qt_gr/grc/core/blocks/__init__.py)
```

willcode: Removed the import reordering.

haakov (approved):

PR #7205, opened 2024-03-21
grc-qt: custom_block_paths set but not used

dl1ksv:
## Description

Preferences contains a custom_block_paths setting. But this setting is not used to find a block.
## Testing Done

Set the custom_block_paths in preferences. Restart grc and have a look at grc.log to see the block_paths.

### DIFF
--- a/grc/gui_qt/Config.py
+++ b/grc/gui_qt/Config.py
@@ -1,6 +1,9 @@
 import os
+from os.path import expanduser, normpath, expandvars, exists
+from collections import OrderedDict
 
 from ..core.Config import Config as CoreConfig
+from qtpy import QtCore
 
 
 class Config(CoreConfig):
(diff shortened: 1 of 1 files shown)


dl1ksv:
The custom_block_paths is part of grc --qt, not part of grc --gtk. It was introduced in https://github.com/gnuradio/gnuradio/blob/0eec37527000670f2068e520690e28c9d523babf/grc/gui_qt/resources/available_preferences.yml#L58-L59

As core/Config.py/Config/ contains parameters that are used together in gtk3 and  qt, we shoud'nt set custom_block_paths there. The custom_block_paths depends on QSettings, so it should go to grc-qtgui/Config.

willcode:
Let's put this in common code somehow. There shouldn't be any non-gui setting that are specific to gtk or qt.

dl1ksv:
I moved the changes to get the value for the custom_block_paths to core/Config.py/Config/Config.py.
That works for both qt and gtk.

willcode:
Is there a reason that Python's `configparser` won't work here? CI wasn't too happy with `import pyqt`.

dl1ksv:
Strange error message. There was a bug in the code. I got this error message running the tests:


`self = <grc.core.Config.Config object at 0x7f904d45de20>
version = 'v3.11.0.0git-680-g54361809', version_parts = ('3', '11', '0')
name = 'GNU Radio Companion Compiler'
prefs = <gnuradio.gr.gr_python.prefs object at 0x7f904edfa070>

    def __init__(self, version, version_parts=None, name=None, prefs=None):
        self._gr_prefs = prefs if prefs else DummyPrefs()
        self.version = version
        self.version_parts = version_parts or version[1:].split(
            '-', 1)[0].split('.')[:3]
        self.enabled_components = self._gr_prefs.get_string(
            'grc', 'enabled_components', '')
        if name:
            self.name = name
    
\>       self.qsettings = QtCore.QSettings(self.gui_prefs_file, QtCore.QSettings.IniFormat)
E       AttributeError: 'Config' object has no attribute gui_prefs_file

../../../grc/core/Config.py:34: AttributeError`

There are to config files grc.conf and grc_qt.conf to be used.
I moved the reference from grc_qtgui/Config.py to core/Config.py and the tests ran without errors.

I don't know why grc stil ran on my system.

dl1ksv:
I tested this Pr on F38 and F39 and the tests work without error. I only get the error messages  if  the package qtpy is missing.  So I replaced qtpy by PyQt5, to see if it works here now.

dl1ksv:
Now the tests pass without error. So pyqt is missing, although pyqt is mentioned at least in the conda part

willcode:
Is the qt config file incompatible with the Python configparser? QSettings (ini mode) was brought in with the qt front end, but I don't think that was the best idea now.

dl1ksv:
The qt config file is compatible. But we use two different files grc_qt.conf for qt and grc.conf for gtk, so they cannot conflict.
If I have correctly understood, the qt version of grc is the preferred version on the long run, so configparser should be replaced by QSettings.
But the coexistence of QSettings and configparser is not the reason for the failing tests. The reason is just the missing package pyqt. I think, we did not come accross this issue , as there was no test where pyqt was used. I'm not familiar how to configure the checks here. So can you advice me , how to install the pyqt package.

willcode:
I asked @haakov for ideas on the GRC Development channel. Are you on matrix? Didn't see you there.

dl1ksv:
Thanks for the hint

Issue #7203, opened 2024-03-18
grc-qt: Blocks rotated 180° are upside down.


drmpeg:
### What happened?

A bug happened!

### System Information

OS: Linux
GR Installation Method: Source

### GNU Radio Version

3.11-git (main)

### Specific Version

v3.11.0.0git-680-g54361809

### Steps to Reproduce the Problem

Rotate a block 180°.

![Screenshot from 2024-03-17 19-28-45](https://github.com/gnuradio/gnuradio/assets/5397196/8561ed7a-a25c-4abc-899f-446a804f2143)

willcode:
Also in #6957. Leaving this one open too, since this is pretty important.

PR #7209, opened 2024-03-23
grc-qt: fix sink ports labels alignment after 180 rotation

vnydv21:
## Description

The PR  #7206 fixes the upside issue and port labels in blocks with only one sink (in) port. 
Unexpectedly, for multiple sinks (in) ports, the port labels are not aligned properly.

Check the left image (before) and right (after) of this PR's commits.

Screenshot from 2024-03-23 04-47-51

## Related Issue

 #7203

gui_qt/components/canvas

## Testing Done

Did UI Test

## Checklist

### DIFF
--- a/grc/gui_qt/components/canvas/port.py
+++ b/grc/gui_qt/components/canvas/port.py
@@ -118,25 +118,32 @@ class GUIPort(QGraphicsItem):
         painter.setPen(pen)
         painter.setBrush(QBrush(self._bg_color))
 
+        # TODO: should boundingRect() be used here ?
         if self.core._dir == "sink":
             rect = QRectF(-max(0, self.width - 15), 0, self.width, self.height)
         else:
             rect = QRectF(0, 0, self.width, self.height)
         painter.drawRect(rect)
 
+        # TODO: Adjustments for finer rotation values if required (eg. 15, 30)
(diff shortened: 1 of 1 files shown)


haakov (changes requested):
Thank you for the PR! Are you able to fix this for the sink ports as well?

msg_strobe

vnydv21:
@haakov, yup. 
All sink port are fixed, the many I checked, a few below.
image

vnydv21:
oops, the PR's name should have, "sink port".
Since, the issue was with sink ports.
Changed.

willcode:
I'm going to put this on main right away since we want people to try out qt and not run into problems like this. But, can we take another pass sometime and make sure there not a simpler way? Seems like we're manually compensating for the rotation with a text translation when there should be a way to rotate the whole thing at once.

vnydv21:
@willcode, yes, you are right. A better approach is indeed to rotate the Rect elements, and the text follows, but I noticed that this would require (maybe) a bit restructuring in certain places, basically to prevent hard-coded snippets. 
The reason is that, by doing so, some other minor GUI issues can also be resolved automatically.

Just an observation, since Qt has significant pre-built tools to simply the GUI.
@haakov, please advise for the same.

Also, I would try to look into it whenever possible.

Issue #5561, opened 2022-02-10
Notify user when GRC file was last opened in older GNU Radio


dkozel:
### Feature Description

Continuing from https://github.com/gnuradio/gnuradio/issues/5521

GRC files used to, and now will have in the future, the version of GRC they were saved with stored in the header. GRC should read this version number and notify the user when it last last opened in a significantly older version of GNU Radio that is expected to have possible changes in behavior (a change in `a` or `b` in the version number `a.b.c.d`).

### Feature Urgency

low (just an idea)

ei8fdb:
@dkozel I'd like to put some design work forward for this notification. Is this a likely scenario -

Bob the radio-astronomer has a grc file, black-hole.grc, that he's created on his desktop using GRC 4.0.

(In the headers GR version=4.0 will be stored in the header.)

Bob wants to test it on his old-trusty radio-astronomy rig which is running a GNU Radio 3.1.

The purpose of the proposed notification is to tell him that the GRC file was written in a much newer Gnuradio version. Because of this, the GRC behaviour might not be what he expects.

Is that correct?

I've got some questions to understand better:

- do we know what Bob's expectations are when he opens the file in GR 3.1?
- is there any advice for users if this happens? i.e. upgrade the grc version? use older blocks?
- are there any ideas what those possible behaviours might be, or be localised to? e.g. filters, math calculations?
- is there a way for them to verify if the behaviour is as expected. i.e. I'm guessing there's not a checksum
- are there particular logs/outputs they should check to verify it's working/or not as it should?

ei8fdb:
A good [usability heuristic when dealing with errors](https://www.nngroup.com/articles/ten-usability-heuristics/
) - helping users recognise, diagnose and recover from them is "Error messages should be expressed in plain language (no error codes), *precisely indicate the problem, and constructively suggest a solution.*"

When I worked on the pip project, we faced a similar issue providing solution advice where users ran into[ dependency resolution problems](https://pip.pypa.io/en/stable/topics/dependency-resolution/?highlight=conflict). Due to the complexity, we couldn't tell the user the exact cause of the conflict but we did provide them with[ some possible solutions](https://pip.pypa.io/en/stable/topics/dependency-resolution/?highlight=conflict#possible-ways-to-reduce-backtracking).

In order to indicate the problem and suggest a solution, I'm trying to list the most possible causes of these "possible changes in behavior" of grc files, and how to solve them (if known).

### Causes for possible change in grc file behaviour 
[...]

PR #7210, opened 2024-03-24
grc:Notify user when GRC file was last opened in older GNU Radio

ZakyHermawan:
## Description

Newer versions of GNU Radio might introduce potential behavioral changes compared to older versions.
This PR will make grc notify user if the grc file was last opened in older GNU Radio Version

## Related Issue

Fixes [#5561](https://github.com/gnuradio/gnuradio/issues/5561)

gtk gui MainWindow: call function to check grc version
gtk gui Notebook: implement grc version checker here
qt gui dialogs component: implement MessageDialog class
qt gui window: implement grc version checker here

note: also refactor few lines in function close_triggered on qt gui window file due to fix python autoformatter problem

## Testing Done

UI Test on gtk and qt gui
how testing is done: 
1. Change grc_version on metadata grc file
2. try to open this file
3. do this on both gui library (gtk and qt)

## Checklist

### DIFF
--- a/grc/gui/MainWindow.py
+++ b/grc/gui/MainWindow.py
@@ -263,6 +263,8 @@ class MainWindow(Gtk.ApplicationWindow):
                 flow_graph=flow_graph,
                 file_path=file_path,
             )
+            page.check_grc_version()
+
             if getattr(Messages, 'flowgraph_error') is not None:
                 Messages.send(
                     ">>> Check: {}\n>>> FlowGraph Error: {}\n".format(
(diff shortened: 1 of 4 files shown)


willcode:
There is `file_format` in the grc, already used (in theory) where there are incompatible changes. This is used, for example, in `core/platform.py/Platform/parse_flowgraph()`. Files will be opened by different versions with no bad effect in most cases, so putting up a warning any time this happens is a bit too aggressive.

The version check logic should be in the core portion of GRC, not duplicated in each GUI.

Where do you see compatibility problems? That would be the thing to address, if possible.

haakov:
I agree with @willcode. I'm also not sure `check_grc_version()` belongs in expr_utils.py

ZakyHermawan:
Thanks for the feedback @haakov @willcode,
I have make some changes to make the notification less aggressive by just put it into console with `log.warning`

can someone suggest me where to put `check_grc_version` ?

PR #7212, opened 2024-03-25
gr-fec: removed inconsistent header and implementation for RS code

### LINKED ISSUE #6985: fec: inconsistent header and implementation for Reed-Solomon int functions
### What happened?

While working in #6984, I have noticed that `_int` functions such as `encode_rs_int()` are declared in `gr-fec/include/gnuradio/fec/rs.h`, but they are never built.

The code that would build these functions is in `gr-fec/include/gnuradio/fec/int.h`, but this file is never included in a compilation unit that gets built (unlike `char.h` and `fixed.h`, which are handled properly). The file `int.h` appears in `init_rs.c`, but I guess that this file is built for `char.h` because `int.h` is not defined.

I see two options:
1. Remove the `_int` function definitions from `rs.h` and the unused code from `fec/` (such as `int.h`).
2. Fix the code in `fec/` so that the `_int` functions get built. The file `init_rs.c` would need to be built twice: once for `char` and another for `int`.

My suggestion is to do 1, since I don't see a use case for the `_int` functions.

### System Information

N/A

### GNU Radio Version

3.11-git (main)

### Specific Version

0.0.0.0

(Look at the code)

### Relevant log output

_No response_

StudHamza:
## Description

Implemented solution 1 that was proposed by @daniestevez:
Remove the `_int` function definitions from `rs.h` and the removed fec/int.h.

## Related Issue

Fixes #6985 

## Testing Done

No errors during build and install.

## Checklist


StudHamza:
forgot to sign my commit, but now I signed it and pushed to main

gr-fec/lib/reed-solomon/rstest.c
```diff
@@ -95,20 +94,8 @@ int main()
...
-                                      Tab[i].prim,
-                                      Tab[i].nroots)) == NULL) {
-                printf("init_rs_int failed!\n");
-                continue;
-            }
-            errs = exercise_int(handle, Tab[i].ntrials);
-#else
             printf("init_rs_init support is not enabled\n");
             exit(1);
-#endif
```
marcusmueller, line 111: Is this proper or do we need to 
```c++
} else {
#if !defined(ALL_VERSIONS)
printf(…);
exit(1);
#endif
}
```

?

StudHamza: You are right, I will fix it then rebuild.

StudHamza:
Building is successful, removed references of `_int` function in `gr-fec/lib/reed-solomon/rs.3` (a man page) and gr-`fec/include/gnuradio/fec/rs.h`, and `gr-fec/lib/reed-solomon/rstest.c`

gr-fec/lib/reed-solomon/rstest.c
daniestevez: Now that we're removing the `_int` functions, we can also remove the test cases in `Tab[]` that have `symsize > 8`, since those are exclusively for the `_int` functions, and also get rid of the `if (Tab[i].symsize <= 8)` and all the `else`.

daniestevez:
@StudHamza there are changes on a QTGUI file that you probably didn't mean to include in the PR.

daniestevez:
There is also code related to the `_int` functions in `exercise.c`. Probably this should be removed too (even though I think it's not being built), since we've also removed it from `rstest.c`.

gr-fec/lib/reed-solomon/rstest.c
StudHamza: what if the the is not `Tab[i].symsize <= 8`, then it will not implement the code in the else statement `else{
#if !defined(ALL_VERSIONS)
            printf("init_rs_init support is not enabled\n");
            exit(1);
#endif
        }`
      I dont think we should remove the `if (Tab[i].symsize <= 8)` and all the `else`.

StudHamza:
I re-based and forced push, plus removed `_int` functions in `exercise.c`

gr-fec/lib/reed-solomon/rstest.c
daniestevez: Just to be clear, my suggestion was as follows:

Amend
```c++
struct {
    int symsize;
    int genpoly;
    int fcs;
    int prim;
    int nroots;
    int ntrials;
} Tab[] = {
    { 2, 0x7, 1, 1, 1, 10 },
    { 3, 0xb, 1, 1, 2, 10 },
    { 4, 0x13, 1, 1, 4, 10 },
    { 5, 0x25, 1, 1, 6, 10 },
    { 6, 0x43, 1, 1, 8, 10 },
    { 7, 0x89, 1, 1, 10, 10 },
    { 8, 0x11d, 1, 1, 32, 10 },
    { 8, 0x187, 112, 11, 32, 10 }, /* Duplicates CCSDS codec */
#ifdef ALL_VESIONS
    { 9, 0x211, 1, 1, 32, 10 },
    { 10, 0x409, 1, 1, 32, 10 },
    { 11, 0x805, 1, 1, 32, 10 },
    { 12, 0x1053, 1, 1, 32, 5 },
    { 13, 0x201b, 1, 1, 32, 2 },
    { 14, 0x4443, 1, 1, 32, 1 },
    { 15, 0x8003, 1, 1, 32, 1 },
    { 16, 0x1100b, 1, 1, 32, 1 },
#endif
    { 0, 0, 0, 0, 0 },
};
```
to
```c++
struct {
    int symsize;
    int genpoly;
    int fcs;
    int prim;
    int nroots;
    int ntrials;
} Tab[] = {
    { 2, 0x7, 1, 1, 1, 10 },
    { 3, 0xb, 1, 1, 2, 10 },
    { 4, 0x13, 1, 1, 4, 10 },
    { 5, 0x25, 1, 1, 6, 10 },
    { 6, 0x43, 1, 1, 8, 10 },
    { 7, 0x89, 1, 1, 10, 10 },
    { 8, 0x11d, 1, 1, 32, 10 },
    { 8, 0x187, 112, 11, 32, 10 }, /* Duplicates CCSDS codec */
    { 0, 0, 0, 0, 0 },
};
```

Then remove the condition
```c++
if (Tab[i].symsize <= 8)
```
since now this will always be true, taking the code inside the `if` outside, so that it always runs. Finally, remove
```c++
        } else{
#if !defined(ALL_VERSIONS)
            printf("init_rs_init support is not enabled\n");
            exit(1);
#endif
        }
```

By the way, this also makes me realize that we aren't building the test with `-DALL_VERSIONS`, and maybe we should.

gr-fec/lib/reed-solomon/exercise.c
```diff
@@ -22,8 +22,6 @@
 #include "fixed.h"
 #define EXERCISE exercise_ccsds
 #elif defined(BIGSYM)
```
daniestevez, line 24: Now that we've removed the `_int` functions, we don't need `BIGSYM` (this part of the `#elif` was now empty, and probably wouldn't build with `-DBIGSYM`).
```suggestion
```

gr-fec/lib/reed-solomon/rs.3
```diff
@@ -62,12 +51,12 @@ and \fBdecode_rs_8\fR.
...
-the _char or _int versions), the user must first
-call \fBinit_rs_int\fR or \fBinit_rs_char\fR as appropriate. The
+the _char ), the user must first
+call \fBinit_rs_char\fR as appropriate. The
 arguments are as follows:
 
 \fBsymsize\fR gives the symbol size in bits, up to 8 for \fBinit_rs_char\fR
-or 32 for \fBinit_rs_int\fR on a machine with 32-bit ints (though such a
+or 32 for on a machine with 32-bit ints (though such a
 huge code would exhaust memory limits on a 32-bit machine). The resulting
```
daniestevez, line 60: This sentence is not correct now. Once we have removed the `_int` functions, the symbol size can only be 8 bits.

gr-fec/lib/reed-solomon/rs.3
```diff
@@ -90,16 +79,15 @@ per code block.
...
+The \fBencode_rs_char\fR functions accept
+the pointer returned by \fBinit_rs_char\fR to
 encode a block of data using the specified code.
 The input data array is expected to
 contain K symbols (of \fBsymsize\fR bits each, right justified
 in each char or int) and \fBnroots\fR parity symbols will be placed
 into the \fBparity\fR array, right justified.
 
-The \fBdecode_rs_char\fR and \fBdecode_rs_int\fR functions correct
+The \fBdecode_rs_char\fR and functions correct
```
daniestevez, line 90:
```suggestion
The \fBdecode_rs_char\fR functions corrects
```

gr-fec/lib/reed-solomon/rs.3
```diff
@@ -121,8 +109,8 @@ order.  This means that the
 array passed through this parameter \fImust\fR have at least \fBnroots\fR
 elements to prevent a possible buffer overflow.
 
-The \fBfree_rs_int\fR and \fBfree_rs_char\fR functions free the internal
-space allocated by the \fBinit_rs_int\fR and \fBinit_rs_char\fR functions,
+The \fBfree_rs_char\fR functions free the internal
```
daniestevez, line 112:
```suggestion
The \fBfree_rs_char\fR function frees the internal
```

gr-fec/lib/reed-solomon/rs.3
```diff
@@ -121,8 +109,8 @@ order.  This means that the
...
+space allocated by the \fBinit_rs_char\fR functions,
 respecitively.
```
daniestevez, line 114:
```suggestion
space allocated by the \fBinit_rs_char\fR function.
```

gr-fec/lib/reed-solomon/rs.3
```diff
@@ -137,7 +125,7 @@ and using the resulting pointer with \fBencode_rs_char\fR and
 \fBdecode_rs_char\fR.
 
 .SH RETURN VALUES
-\fBinit_rs_int\fR and \fBinit_rs_char\fR return a pointer to an internal
+\fBinit_rs_char\fR return a pointer to an internal
```
daniestevez, line 128:
```suggestion
\fBinit_rs_char\fR returns a pointer to an internal
```

StudHamza:
Fixed grammar errors in man page `rs.3`, removed `if (Tab[i].symsize <= 8)` and `#ifdef ALL_VESIONS
    { 9, 0x211, 1, 1, 32, 10 },
    { 10, 0x409, 1, 1, 32, 10 },
    { 11, 0x805, 1, 1, 32, 10 },
    { 12, 0x1053, 1, 1, 32, 5 },
    { 13, 0x201b, 1, 1, 32, 2 },
    { 14, 0x4443, 1, 1, 32, 1 },
    { 15, 0x8003, 1, 1, 32, 1 },
    { 16, 0x1100b, 1, 1, 32, 1 },
#endif`

gr-fec/lib/reed-solomon/rs.3
```diff
@@ -62,13 +51,11 @@ and \fBdecode_rs_8\fR.
 
 The _8 and _ccsds functions do not require initialization.
 To use the general purpose RS encoder or decoder (i.e.,
-the _char or _int versions), the user must first
-call \fBinit_rs_int\fR or \fBinit_rs_char\fR as appropriate. The
+the _char ), the user must first
```
daniestevez, line 54:
```suggestion
the _char version), the user must first
```

gr-fec/lib/reed-solomon/rstest.c
```diff
@@ -76,40 +65,27 @@ int main()
...
-                printf("init_rs_int failed!\n");
-                continue;
-            }
-            errs = exercise_int(handle, Tab[i].ntrials);
-#else
+        } 
+#if !defined(ALL_VERSIONS)
             printf("init_rs_init support is not enabled\n");
             exit(1);
 #endif
```
daniestevez, line 88: What is the intent here?

daniestevez: This piece of code wasn't really doing anything in the original code. I imagine that the intent was that if by accident a test case with `symsize > 8` was declared outside the `#ifdef ALL_VERSIONS`, then the test would print and error and finish, because `init_rs_int()`, which is necessary to test that test case, was only used when `ALL_VERSIONS` was defined. Now that we've removed the `_int()` functions, this piece of code is not only not doing anything, but also it doesn't play any role anymore, since now there should be no test cases with `symsize > 8` defined at all.

By the way, I realized that the original code uses `#ifdef ALL_VESIONS` in one place instead of `#ifdef ALL_VERSIONS`, but this doesn't matter because we've now deleted all this.

daniestevez (approved):
All looks good to me now.

PR #7222, opened 2024-03-26
runtime and fft: use create_directories() instead of iterating

### LINKED ISSUE #7156: Build problem with MinGW (filesystem error)
### What happened?

Change https://github.com/gnuradio/gnuradio/commit/5173f01788a7e3c95e07286dbfca22e8b75d5273 caused an error when compiling with MinGW gcc.
Before this everything was fine.

### System Information

Host OS: Linux Ubuntu based 20.04 amd64
Compiler: GCC 11.4, mingw-w64 11.0.1 (custom toolchain)
GR Installation Method: Cross-compiling

### GNU Radio Version

3.11-git (main)

### Specific Version

v3.11.0.0git-668-gc2d1b5c5

In `gr-fft/lib/fft.cc` I replaced the code blocks wrapped in `_MSC_VER` with `_WIN32`, and the build was successful.
But when trying to use Gnuradio it results in an error:
`filesystem error: cannot create directory: File exists [C:]`

### Relevant log output

```shell
terminate reached from thread id: 124Got std::system_error
filesystem error: cannot create directory: File exists [C:]
```

willcode:
## Description
Iterating while creating subdirs in a path was causing problems for MinGW, an is not necessary. Replace with `create_directories()`.

Also use the same fopen() parameters with or without _MSC_VER.

## Related Issue
Fixes #7156

- runtime vmcircbuf config
- fft

## Checklist

### DIFF
--- a/gnuradio-runtime/lib/vmcircbuf_prefs.cc
+++ b/gnuradio-runtime/lib/vmcircbuf_prefs.cc
@@ -37,17 +37,8 @@ static auto pathname(string_type key)
 static void ensure_dir_path()
 {
     // recursively make sure the directory exists
-    fs::path path;
-    for (const auto& path_component : gr::paths::userconf()) {
-        path /= path_component;
-        if (!fs::exists(path)) {
-            fs::create_directory(path);
-        }
-    }
-
-    path = path / "prefs";
-    if (!fs::exists(path))
-        fs::create_directory(path);
+    fs::path path = gr::paths::userconf() / "prefs";
+    fs::create_directories(path);
 }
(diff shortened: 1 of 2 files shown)


argilo (approved):
I made a test build of Gqrx using this branch:

https://github.com/gqrx-sdr/gqrx/actions/runs/8441795368

Gqrx runs correctly, and I verified that the `.config\gnuradio` and `.cache\gnuradio` directories are recreated if needed.

I also verified that things still work on Ubuntu, and didn't run into any problems.

PR #7239, opened 2024-04-01
gr-trellis: port examples from xml and change the deprecated block (backport to maint-3.10)

willcode:
Backport #7216


willcode:
Mistakenly backported from main to main ... actually backported as part of #7680

PR #7252, opened 2024-04-03
gr-filter: delete gr-filter/grc/filter_low_pass_xlating_filter.block.yml

duggabe:
## Description

Delete an unused and obsolete yml block.

Affects only the subject block which is not used in any GNU Radio version since v3.7.

## Checklist


willcode:
The clarify on "unused", the grc yml file exists, and has even been maintained. But, it has never been listed in `CMakeLists.txt` so it has never been installed.

Deleting this is fine. There is also the option of adding it to `CMakeLists.txt`. We have `Low Pass Filter` which does exactly the same thing, but with a non-xlating FIR.

haakov (approved):

PR #7254, opened 2024-04-06
gr-gtqui: Fixed Qt GUI Dial - weird behaviour

### LINKED ISSUE #5107: Qt GUI Dial - weird behaviour
The Qt GUI Dial default value seems to be the exact value returned when read from other blocks, not the knob position. This works OK if the scale factor is 1.
 
If it is, say 0.01, the knob is initially set to a 1/100th of angle that represents the default value when the scale factor is applied, and the displayed value is also 1/100th of the default value.

Conversely, if the default value is interpreted as the initial knob position, the returned value does not account for the scale factor - until the knob is moved.

It needs to be documented whether the default value refers to the actual value or the knob position and then either divide with the scale factor to set the initial position or apply the scale factor when setting the default value.

StudHamza:
## Description

In issue #5107 the Qt GUI dial default seems to be the exact value returned, if the default value is interpreted as the initial knob position, the returned value does not account for the scale factor - until the knob is moved.
Check the issue for more explaination

## Related Issue

This PR will fix #5107 

Only the QT GUI Dial is affected. There is no effect on the performance. 
## Testing Done

I built and ran the example provided by the issue
previous results are explained in detail here:
https://github.com/gnuradio/gnuradio/issues/5107#issuecomment-932992531

New results:
https://github.com/gnuradio/gnuradio/assets/118675709/9f595841-ce4f-4cae-a21c-1cfd16786aeb

testing case file: 
[qt_gui_dialweirdness.grc.zip](https://github.com/gnuradio/gnuradio/files/7274174/qt_gui_dial_weirdness.grc.zip)

## Checklist

### DIFF
--- a/gr-qtgui/python/qtgui/dialcontrol.py
+++ b/gr-qtgui/python/qtgui/dialcontrol.py
@@ -23,7 +23,7 @@ class LabeledDialControl(QFrame):
                  minsize=100, isFloat=False, scaleFactor=1, showvalue=False,
                  outputmsgname='value'):
         QFrame.__init__(self, parent)
-        self.numberControl = DialControl(minimum, maximum, defaultvalue, backgroundColor,
+        self.numberControl = DialControl(minimum, maximum, defaultvalue / scaleFactor, backgroundColor,
                                          self.valChanged, changedCallback, minsize)
 
         layout = QVBoxLayout()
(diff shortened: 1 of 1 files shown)


StudHamza:
Do i close this pull request?

willcode:
> Do i close this pull request?

You can close PRs if you want to. I've taken a quick look at it, but need to try it out. PRs automatically close when they're merged.

PR #7256, opened 2024-04-07
grc_qt: save panel settings

dl1ksv:
## Description

In View-> Panels four panels can be enabled and disabled. But these settings are not saved and restored.
This is done now

## Testing Done

Run grc --qt
View-> Panels
 Disable some panels and restart

## Checklist

### DIFF
--- a/grc/gui_qt/components/block_library.py
+++ b/grc/gui_qt/components/block_library.py
@@ -106,6 +106,8 @@ class BlockLibrary(QDockWidget, base.Component):
     def __init__(self):
         super(BlockLibrary, self).__init__()
 
+        self.qsettings = self.app.qsettings
+
         self.setObjectName("block_library")
         self.setWindowTitle("Block Library")
 
@@ -204,6 +206,9 @@ class BlockLibrary(QDockWidget, base.Component):
 
         # Dict representing which examples contain various blocks
         self.examples_w_block = {}
+
+        if not self.qsettings.value("appearance/display_blocklibrary", True, type=bool):
+            self.hide()
 
     def createActions(self, actions):
         pass
--- a/grc/gui_qt/components/wiki_tab.py
+++ b/grc/gui_qt/components/wiki_tab.py
@@ -30,7 +30,7 @@ log = logging.getLogger(f"grc.application.{__name__}")
 
 
 class WikiTab(QtWidgets.QDockWidget, base.Component):
-    def __init__(self, argv_enabled=False):
+    def __init__(self):
         super(WikiTab, self).__init__()
 
         self.qsettings = self.app.qsettings
@@ -40,27 +40,11 @@ class WikiTab(QtWidgets.QDockWidget, base.Component):
 
         self.setFloating(False)
 
-        active = None
-        if argv_enabled:
-            active = True
-        else:
-            if self.qsettings.value("appearance/display_wiki", False, type=bool) == True:
-                active = True
-            else:
-                active = False
-
-        if active:
-            try:
-                from qtpy.QtWebEngineWidgets import QWebEngineView
-                self.hidden = False
-            except ImportError:
-                log.error("PyQt QWebEngine missing!")
-                self.hide()
-                self.hidden = True
-                return
-        else:
+        try:
+            from qtpy.QtWebEngineWidgets import QWebEngineView
+        except ImportError:
+            log.error("PyQt QWebEngine missing!")
             self.hide()
-            self.hidden = True
             return
 
         # GUI Widgets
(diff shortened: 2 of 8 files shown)


grc/gui_qt/components/window.py
```diff
@@ -1475,6 +1475,28 @@ def exit_triggered(self):
                 # We cancelled closing a tab. We don't want to close the application
                 return
 
+        # Save the panel settings
```
willcode, line 1478: Can these use `!x.isHidden()` to avoid the if/else?

grc/gui_qt/components/wiki_tab.py
```diff
@@ -40,29 +40,17 @@ def __init__(self, argv_enabled=False):
...
-            try:
-                from qtpy.QtWebEngineWidgets import QWebEngineView
-                self.hidden = False
-            except ImportError:
-                log.error("PyQt QWebEngine missing!")
-                self.hide()
-                self.hidden = True
-                return
-        else:
+        try:
```
willcode, line 43: The wiki widget causes a continuous stream of error messages on the console on Fedora+Wayland, so this was an opt-in feature until we figured that out. I'm ok with the saved setting, but we would just remove the `--wiki` arg and default to off in that case. No need to have a command line way to turn the feature on, but not turn it back off.

grc/gui_qt/components/block_library.py
```diff
@@ -205,6 +207,9 @@ def __init__(self):
         # Dict representing which examples contain various blocks
         self.examples_w_block = {}
 
+        if not self.qsettings.value("appearance/display_blocklibrary", True, type=bool):
```
willcode, line 210: What determines whether something goes under `appearance` or `grc`? It seems to be mix and match. Since there are `show_` vars under `grc`, it seems like these should go there. And they could be renamed `show_blocklibrary`, etc.

grc/gui_qt/components/window.py
dl1ksv: Yes, that can be done. I aggree, it looks much better. I'll fix it

grc/gui_qt/components/block_library.py
dl1ksv: Under appearance there is gui / widget related stuff. blocklibrary, console etc are dockwidgets. The show_..... are block or flowgraph related stuff. So I'd prefer to keep the setting  show/hidden for the dockwidgets under appearance.
By the way what about [grc] in config.conf?

grc/gui_qt/components/wiki_tab.py
dl1ksv: I tested on F39 and got no errors. Sure, if save and restore the panel settings, there is no need for an command line option.
By the way , I just saw that the  Display Wiki Tab option can be modified in the preference editor in the Appearance Tab

grc/gui_qt/components/block_library.py
willcode: I deleted the `[grc]` section from `config.conf` on my machine and it was not added back in when I ran grc. Must have been old stuff.

willcode:
No, I'd say that should be removed from Preferences if the selection in View does exactly the same thing.

grc/gui_qt/components/wiki_tab.py
willcode: I'm running Wayland and the nvidia driver. It could be specific to that.

grc/gui_qt/components/block_library.py
dl1ksv: No, that's not old stuff. config.conf is never written by grc it's  readonly. 
This is a part of my config.conf

[grc]
canvas_default_size=1280, 1024
canvas_font_size=8
default_flow_graph=
editor=gedit
global_blocks_path=/usr/local/gnuradio/share/gnuradio/grc/blocks
local_blocks_path=/usr/local/oot3/share/gnuradio/grc/blocks
xterm_executable=/usr/bin/xterm

My oot's are located in local_blocks_path. If I remove config.conf  no oot's are found in grc --gtk.

If you remove config.conf  some settings may be taken from
/usr/local/gnuradio/etc/gnuradio/conf.d/grc.conf

But neither grc --gtk nor grc --qt has an editor for config.conf.

willcode: So, none of these things will be read from `grc[_qt].conf`? What creates the `[grc]` entries in `config.conf`?

dl1ksv: At least xterm_executable is read from grc_qt.conf.
You have to create the entries in config.conf by yourself using vi or nano.
So I would plead for moving  the grc settings for grc --qt to grc_qt.conf  but keep them in config.conf  for the use in --gtk.
Doing so, we could edit these settings using the ( new ) preference editor.
But I'd like to do it in a new pr

PR #7263, opened 2024-04-10
Add callback to the protocol formatter block

jsallay:
## Checklist


marcusmueller:
@willcode  ideally, yes.

@jsallay the following idea should **not** affect this PR at all; but:<br>
I think we have a total of 4 `header_format_base` implementations (crc, default, ofdm, counter?). Could we add `virtual header_format_base::serialize() -> pmt_t` that impls overload, and a `header_formatter_registry` with a `…_registry::deserialize(pmt_t) -> unique_ptr<header_format_base>`? That way we could do the architecturally nice thing and change formatters using tags and message passing. (As a bonus, we could save and restore protocol state for QA.)

jsallay:
I agree as well. I can add that in

PR #7264, opened 2024-04-12
gr_modtool: remove colon in block yaml label

eshrh:
## Description
Removes the colon in the default block yaml parameter label.  
The presence of this colon causes the default generated block to throw an obscure error: "FlowGraph Error: mapping values are not allowed in this context"

With the colon removed, the FIX ME message renders correctly. Note that the colon itself is unnecessary, since grc renders parameter labels with a colon after it automatically. If you quote the label, then grc shows "FIX ME::".

## Testing Done
Default behavior:
```
gr_modtool add test
<compile the block>
<open grc>
```
And you get `FlowGraph Error: mapping values are not allowed in this context` and the block fails to appear.

After removing the colon it works fine.

## Checklist

### DIFF
--- a/gr-utils/modtool/templates/templates.py
+++ b/gr-utils/modtool/templates/templates.py
@@ -566,7 +566,7 @@ templates:
 #     * default
 parameters:
 - id: parametername_replace_me
-  label: FIX ME:
+  label: FIX ME
   dtype: string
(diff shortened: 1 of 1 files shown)


willcode:
It's not quite a bug, since the developer should be changing these lines, but makes sense to make things build if possible.

Issue #7265, opened 2024-04-12
Documentation for Max block is incorrect


mormj:
### What happened?

![image](https://github.com/gnuradio/gnuradio/assets/34754695/a38470ce-a9b2-489d-a55c-484390ef5ea6)

Documentation on the max block seems to include info from other blocks

### System Information

OS: Ubuntu 22.04
GR Installation Method: Source

### GNU Radio Version

3.11-git (main)

### Specific Version

0.0.0.0

### Steps to Reproduce the Problem

Open the properties of the Max block

ZakyHermawan:
In _update_docs_page method from class PropsDialog, im trying to print docstrings variable right after `docstrings = self._block.documentation.copy()`, here's the output
![image](https://github.com/gnuradio/gnuradio/assets/44158147/eafe32ed-940e-41ad-a04f-c791978fcca1)

As you can see, from the beginning, there is matrix interleaver docs in the docstrings dict, this docs is taken from matrix_interleavers.py

ZakyHermawan:
are there any reason for a block to have docs from yml file and the python file at the same time ? isn't the python file for a block is actually generated from yml file ?

We can quick fix this issue by just return right after yaml file being parsed:
```python
if from_yaml:
  buf.insert(pos, '\n')
  return

```
![image](https://github.com/gnuradio/gnuradio/assets/44158147/b4b29f50-8949-480d-865e-0d96efe52c9f)

ffu:
This bug can be traced to function docstring_guess_from_key in module extract_docs.py. The handling of block names with variable types like blocks_max_xx is incorrect. The function uses a regex to search for matching block names:

    pattern = re.compile(
        '^' + init_name.replace('_', '_*').replace('x', r'\w') + r'\w*$')
        
However, for an init_name like "max_xx", the regex becomes r"^ma\w_*\w\w\w*$". The x in "max" is replaced with "\w" and the underscore is optional.  Therefore, any block that begins with "ma" and contains at least four characters become a matching block for max_xx. This includes magphase_to_complex and matrix_interleaver. Eventually, the documentation for the max block also contains docstrings for magphase_to_complex and matrix_interleaver.

A possible solution is to check if the block name ends with _x, _xx, _xxx, or _vxx. If so, then only replace x with "\w" in the suffix.

PR #7271, opened 2024-04-15
fix issue where the documentation is coming from another block

ZakyHermawan:
## Description

Fix issue where the documentation come from another block
Also add documentation for max block

## Related Issue

Fixes https://github.com/gnuradio/gnuradio/issues/7265

Max block

## Testing Done

max block now have correct documentation
image

docs for another blocks still working properly, here's few examples
image
image
image

## Checklist


willcode:
@ZakyHermawan does the bug exist in the Qt version, and if so is there an equivalent fix?

ZakyHermawan:
Documentation for blocks in Qt version of grc:

image

image

image

willcode:
Would you remove the documentation for the max block and submit that as a separate PR? We can work on the wording a little there. I'll get the code part of this in for this release.

willcode:
This PR causes a failure when opening disable blocks. Let's hold off until the next release. A lot of blocks don't have any documentation in this field anyway and rely on the wiki ... something we need to think about.

ZakyHermawan:
@willcode would you please give an example on how to reproduce the bug ? 

I can open properties of disabled max block
image

Btw I make mistake where i didn't set the textedit for documentation to readonly (i can fix this easily)

Also, I have made a fix on `test_qtbot where the test case failed on test_move_blocks because it tries to compare 270 with -90`. Do you want me to make a separate PR for this ? (if yes, does GR have issue where i could mention in the PR description?)

willcode:
The QT Sink block is the one that didn't work in my test, when disabled. It evenutally caused a crash. If you can't replicate it, I'll look through the rather messy stack trace it caused before crashing.

haakov:
I'm able to reproduce. Here's the trace:

```
  File "/home/hkon/qt_gr/grc/core/blocks/_templates.py", line 77, in render
    return template.render(**namespace)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/mako/template.py", line 438, in render
    return runtime._render(self, self.callable_, args, data)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/mako/runtime.py", line 874, in _render
    _render_context(
  File "/usr/lib/python3.11/site-packages/mako/runtime.py", line 916, in _render_context
    _exec_template(inherit, lclcontext, args=args, kwargs=kwargs)
  File "/usr/lib/python3.11/site-packages/mako/runtime.py", line 943, in _exec_template
    callable_(context, *args, **kwargs)
  File "memory:0x75bb98128890", line 115, in render_body
TypeError: not all arguments converted during string formatting

  File "/home/hkon/qt_gr/grc/gui_qt/components/canvas/block.py", line 423, in mouseDoubleClickEvent
    self.open_properties()
  File "/home/hkon/qt_gr/grc/gui_qt/components/canvas/block.py", line 453, in open_properties
    self.props_dialog = PropsDialog(self.core, self.force_show_id)
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/hkon/qt_gr/grc/gui_qt/components/dialogs.py", line 176, in __init__
    block_constructor = block_templates.render(
                        ^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/hkon/qt_gr/grc/core/blocks/_templates.py", line 79, in render
    raise TemplateError(error, text)
grc.core.errors.TemplateError: (TypeError('not all arguments converted during string formatting'), "<%\n    win = 'self._%s_win'%id\n%>\\\nqtgui.${type.fcn}(\n    ${fftsize}, #fftsize\n    ${wintype}, #wintype\n    ${fc}, #fc\n    ${bw}, #bw\n    ${name}, #name\n    ${plotfreq}, #plotfreq\n    ${plotwaterfall}, #plotwaterfall\n    ${plottime}, #plottime\n    ${plotconst}, #plotconst\n    None # parent\n)\nself.${id}.set_update_time(1.0/${rate})\n${win} = sip.wrapinstance(self.${id}.qwidget(), Qt.QWidget)\n\nself.${id}.enable_rf_freq(${showrf})\n\n${gui_hint() % win}")
```

haakov:
Actually, the error appears with GRC Gtk too! But it doesn't crash the application

ffu:
There are multiple issues in this conversation including:
1. The original issue #7265 
2. GRC Qt does not have a documentation tab
3. An exception occurs when opening a disabled Qt block like qtgui sink

Should they be separated out? I'll submit a pull request for the first issue if there are no objections.

ffu:
The exception occurs in all Qt GUI blocks because of the gui_hint template in the make section of yaml files. They usually end with:
```
 ${gui_hint() % win}
```
However the evaluate method in Param contains the following:
```
elif dtype == 'gui_hint':
    return self.parse_gui_hint(expr) if self.parent_block.state == 'enabled' else ''
```
[Link to param.py
](https://github.com/gnuradio/gnuradio/blob/44aff13ce59255397ab43fa4d088dd6d3021ec89/grc/core/params/param.py#L271-L272)

When the Qt GUI block is disabled, gui_hint() evaluates to an empty string, and during the template call to render, ${gui_hint() % win} cannot be formatted because there is no "%s" in gui_hint().

Preliminary tests show that removing the ternary operator and changing to:
```
elif dtype == 'gui_hint':
    return self.parse_gui_hint(expr)
```
will fix the exception. I'll test a little more to make sure it doesn't break anything.

marcusmueller:
I think the "wrong docs" issue is indeed fixed, and the "show documentation" should probably be done in a separate PR. Closing for now :(

Issue #7269, opened 2024-04-15
grc-qt: AttributeError


dl1ksv:
### What happened?

Double clicking a connection line should remove the connection ( or do nothing ? )

### System Information

Linux, F39 build from source

### GNU Radio Version

3.11-git (main)

### Specific Version

0.0.0.0

### Steps to Reproduce the Problem

Start grc --qt

Build a simple flowgraph 
audio source -> audio sink
The double click on the connection line  and you get the err

### Relevant log output

```shell
Traceback (most recent call last):
  File "/usr/local/gnuradio/lib64/python3.12/site-packages/gnuradio/grc/gui_qt/components/canvas/connection.py", line 163, in mouseDoubleClickEvent
    self.parent.connections.remove(self)
    ^^^^^^^^^^^
AttributeError: 'GUIConnection' object has no attribute 'parent'. Did you mean: 'paint'?
```

dl1ksv:
In grc --gtk double clicking a connection does the same as a simple click, just selecting the line.

willcode:
Needs a deref to the core object.

PR #7278, opened 2024-04-19
fix double click connection to delete for qt gui

ZakyHermawan:
## Description

Fix AttributeError while deleting connection with double click

## Related Issue

Fixes https://github.com/gnuradio/gnuradio/issues/7269

`grc/gui_qt/components/canvas/connection.py`

## Testing Done

- Manual gui test
- Automated gui testing with `pytest grc/tests/test_qtbot.py`

## Checklist

### DIFF
--- a/grc/gui_qt/components/canvas/connection.py
+++ b/grc/gui_qt/components/canvas/connection.py
@@ -159,5 +159,6 @@ class GUIConnection(QGraphicsPathItem):
         painter.drawPath(self._arrowhead)
 
     def mouseDoubleClickEvent(self, e):
-        self.parent.connections.remove(self)
+        connection = self.core
+        connection.parent.connections.remove(connection)
         self.scene().removeItem(self)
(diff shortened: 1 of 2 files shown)


grc/gui_qt/components/canvas/connection.py
```diff
@@ -159,5 +159,6 @@ def paint(self, painter, option, widget):
         painter.drawPath(self._arrowhead)
 
     def mouseDoubleClickEvent(self, e):
-        self.parent.connections.remove(self)
+        connection = self.core
+        connection.source_port.parent_flowgraph.connections.remove(connection)
```
willcode, line 163: Isn't there a `connection.parent`? Or did that not work?

ZakyHermawan: you're right. I think using `connection.parent` is better than `connection.source_port.parent_flowgraph`

willcode:
Thanks. For future PRs, please start the commit message (and PR title) with the module name (and block if applicable), like
`grc-qt: fix thing`
This make it easier to run through the commits and PRs later.

willcode:
@ZakyHermawan see #7284 ... there should be a better way to do this, e.g., delete element. I'm going to hold off backporting this because it leaves the flowgraph in an invalid state but doesn't mark it as such. Take a look at how the select/delete way of removing a connection works.

PR #7283, opened 2024-04-21
grc-qt: fix double click connection to delete (backport to maint-3.10)

willcode:
Backport #7278


willcode:
It looks like there is a better way to do this, so not backporting the current fix from main.

PR #7292, opened 2024-04-24
GRC: Allow for angular connections

### LINKED ISSUE #5152: Re-implement support for straight connections between blocks
Some folks prefer having straight/90 degree lines for the connections between blocks. The math and code for the connection types is still available from GNU Radio 3.7 and could be cleaned up and integrated as an option so users can set a preference for one or the other.

Current curved implementation
https://github.com/gnuradio/gnuradio/blob/fd45d06b8fa4c398ffc5cac53b78e82344be9e95/grc/gui/canvas/connection.py#L103-L124

3.7 straight line implementation
https://github.com/gnuradio/gnuradio/blob/248f50f3460fe5dde2aa5b05b34eeec78e1439ac/grc/gui/Connection.py#L102-L144

marcusmueller:
- **GRC: Angular connection (by choice via Preferences)**
- **GRC: Make connection line width preference-configurable**

## Description

grc_angled

## Related Issue

Closes #5152

- Qt GRC
- Hopefully a selling point for Qt GRC for those who want to go back to the 3.7 days, optically :D

## Testing Done

grc_angled

## Checklist

### DIFF
--- a/grc/gui_qt/components/canvas/connection.py
+++ b/grc/gui_qt/components/canvas/connection.py
@@ -26,6 +26,7 @@ class StyledConnection():
         self._qsettings = QApplication.instance().qsettings
         self._conn_style_str = self._qsettings.value("appearance/connection_style", "CURVED", type=str)
         self._conn_style = self.ConnectionStyle[self._conn_style_str]
+        self._line_width_factor = self._qsettings.value("appearance/line_width_factor", 1.0, type=float)
 
 
 class DummyConnection(QGraphicsPathItem, StyledConnection):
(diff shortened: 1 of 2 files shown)


haakov (approved):
Great! I only have a few things:

- I think the DummyConnection line width should be updated too
- This setting requires re-opening the FG (otherwise, only new connections are affected). The other options in the Appearance tab typically require restart. Maybe there should be a label at the top of the tab saying that? I'm fine with making a Github issue of this and merging this PR as-is.

marcusmueller:
I'll quickly fix the DummyConnection issue. Regarding restart: yeah, I think just putting that warning there once is nicer, and we should maybe do that separately.

marcusmueller:
Oh and maybe we really want to get rid of all GUI constants from the `Constants` module? Gut feeling is they should be moved to a style sheet?

haakov (approved):

PR #7294, opened 2024-04-24
grc-qt: grc --qt filename.grc doesn't open filename.grc

### LINKED ISSUE #7291: gnuradio-companion --qt filename.grc doesn't open filename.grc
### What happened?

Tried to open a GRC flowgraph directly with QT GRC; it doesn't do that, doesn't throw an error. The same file works when opened via menu.

Doesn't work without the --qt if the default GRC variant is set to QT in the preferences, either.

### System Information

OS: linux

### GNU Radio Version

3.11-git (main)

### Specific Version

v3.11.0.0git-716-g77155087

gnuradio-companion --qt filename.grc

### Relevant log output

_No response_

dl1ksv:
## Description

Now the list of files specified on the cl for grc is appended to the list of files to be opened on startup.

fixes #7291

## Checklist

### DIFF
--- a/grc/gui_qt/components/window.py
+++ b/grc/gui_qt/components/window.py
@@ -66,7 +66,7 @@ QStyle = QtWidgets.QStyle
 
 
 class MainWindow(QtWidgets.QMainWindow, base.Component):
-    def __init__(self):
+    def __init__(self, file_path=[]):
         QtWidgets.QMainWindow.__init__(self)
         # base.Component.__init__(self)
 
@@ -161,7 +161,7 @@ class MainWindow(QtWidgets.QMainWindow, base.Component):
         )
         self.setCentralWidget(self.tabWidget)
 
-        files_open = list(self.app.qsettings.value('window/files_open', []))
+        files_open = list(self.app.qsettings.value('window/files_open', [])) + file_path
         grc_file_found = False
         if files_open:
             for file in files_open:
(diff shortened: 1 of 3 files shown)


marcusmueller (approved):
Looks good, tested it and it works!

grc/gui_qt/components/window.py
```diff
@@ -66,7 +66,7 @@
 
 
 class MainWindow(QtWidgets.QMainWindow, base.Component):
-    def __init__(self):
+    def __init__(self, file_path=[]):
```
marcusmueller, line 69: Setting a default `= []` argument *can* be tricky in situations where the function modifies the parameter; the default parameter is only constructed once and then it stays the same object forever. For example:

```python
def print_list(printme = []):
  printme += ["end"]
  print(printme)

print("first call:")
print_list()

print("second call:")
print_list()
```

would print
```text
first call:
['end']
second call:
['end', 'end']
```

marcusmueller: The typical workaround is to use a `None` default parameter:
```python
def better_print_list(printme = None):
  if not printme:
   printme = []
  printme += ["end"]
  print(printme)
```

marcusmueller (commented):
Not a criticism, because the usage here is IMHO absolutely OK (`file_path` would not reasonably ever be changed in `__init__`); really just a remark:

grc/gui_qt/components/window.py
dl1ksv: Thanks for your clarification. I used [] as default, as this is just the value you get from the cli if you enter no grc file.  So I can just add file_path to files_open. And as you mentioned file_path will not be changed in __init__

PR #7324, opened 2024-05-07
CI: conda: Update conda build configuration and scripts

ryanvolz:
## Description

This updates some of the dependencies that the conda CI builds against, and keeps pace with the conda-forge ecosystem. I'm also hoping that this will fix the macOS CI, which seems to think it is a doing a cross-compilation ever since some recent conda-forge infrastructure change.

## Checklist

### DIFF
--- a/.ci_support/linux_64_numpy1.23python3.11.____cpython.yaml
+++ b/.ci_support/linux_64_numpy1.23python3.11.____cpython.yaml
@@ -1,5 +1,5 @@
 alsa_lib:
-- 1.2.10
+- '1.2'
 c_compiler:
 - gcc
 c_compiler_version:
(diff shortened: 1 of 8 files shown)


ryanvolz:
> I'm also hoping that this will fix the macOS CI, which seems to think it is a doing a cross-compilation ever since some recent conda-forge infrastructure change.

A-ha! The GitHub runner macos-latest image started pointing to macos-14, which is arm64 and thus the x64 builds were indeed being cross-compiled! This update does set the GitHub runner to macos-13 explicitly, so that should be fixed.

marcusmueller:
uh, but we could also have native arm64 macOS builds?

ryanvolz:
> oh, but we could also have native arm64 macOS builds?

Not yet, but I imagine that's coming soon.

Issue #7322, opened 2024-05-06
freedesktop warning during cmake (GR3.10)


Ech0xing:
### What happened?

A bug happened!
CMake Warning (dev) at grc/scripts/freedesktop/CMakeLists.txt:13 (configure_file)

After compiling from source, the icon does not show up (not a PATH problem) and unable to type in gnuradio-companion from the CLI.

### System Information

OS: Ubuntu 20.04.6
GR Installation Method: Source

### GNU Radio Version

3.10 (maint-3.10)

### Specific Version

3.10.10

### Steps to Reproduce the Problem

Not sure why this is happening, I have installed all of the dependencies. I looked at line 13 from the CMakeList.txt file and cant make heads or tails of it. I have reloaded the laptop twice and still get the same error. This issue causes where I cant find the icon to click on GNURadio Companion under the App menu, nor can i type in gnuradio-companion from the CLI to open.

xdg-utils is installed

git clone https://github.com/gnuradio/gnuradio.git
cd gnuradio
git checkout maint-3.10
mkdir build
cd build
cmake -DCMAKE_BUILD_TYPE=Release -DPYTHON_EXECUTABLE=/usr/bin/python3 ../

This is where part where the Dev error comes up:

-- Configuring JSON/YAML config blocks support...
--   Dependency ENABLE_GRC = ON
--   Dependency JSONSCHEMA_FOUND = TRUE
--   Enabling JSON/YAML config blocks support.
--   Override with -DENABLE_JSONYAML_BLOCKS=ON/OFF
CMake Warning (dev) at grc/scripts/freedesktop/CMakeLists.txt:13 (configure_file):
  configure_file called with unknown argument(s):

   FILE_PERMISSIONS
   OWNER_READ
   OWNER_WRITE
   OWNER_EXECUTE
   GROUP_READ
   GROUP_EXECUTE
   WORLD_READ
   WORLD_EXECUTE

Attached is a screenshot of the CMakeList.txt file that is in question. The highlighted portion is the line in question.
![GNURadio-ConfigError](https://github.com/gnuradio/gnuradio/assets/71466458/77cbe0f4-3e41-4c93-90e8-8ea65d76d0e8)

### Relevant log output

```shell
-- Configuring JSON/YAML config blocks support...
--   Dependency ENABLE_GRC = ON
--   Dependency JSONSCHEMA_FOUND = TRUE
--   Enabling JSON/YAML config blocks support.
[...]

Issue #7323, opened 2024-05-07
Bump CMake requirement to 3.22 on main


marcusmueller:
### Feature Description

As seen in #7294, we tend to use cmake features not present in the fairly ancient minimum CMake version we demand; in that case, we'd need at least 3.20.

Considering cmake versions:

| distro | release | cmake ver |
|--|--|--|
|debian  | oldstable (bullseye)| 3.18| 
|debian | stable (bookworm) |3.25|
|Ubuntu| 22.04 | 3.22|
|Ubuntu| 23.10 | 3.27|
|Ubuntu| 24.04 | 3.28|
|Ubuntu| 24.10| 3.27|
|Fedora | 39 | 3.27|
|Fedora | 40| 3.28 |

I'd say that even for maint-3.10, cmake 3.18 would be safe (and I think even without #7294 we do use 3.16-ish features, anyway) . For main, we should go to at least 3.22 (if we decide that we want to target Ubuntu 22.04 with maint-3.10), or staright for 3.25.

So, this boils down to deciding:

- maint-3.10: Update at all? Updated to 3.18 or 3.22?
- main: Update to 3.22 or 3.25?

### Feature Urgency

medium (would be nice to have in the near future)

dkozel:
@johnwparent and @mwoehlke-kitware Tagging you both as an FYI of the discussion and to ask for input on what features you're hoping or expecting to use in the installer work.

Thanks!

willcode:
@marcusmueller Ubuntu 20.04 has 3.16 and has EOL in 2025, so we'd have to drop support.

marcusmueller:
@willcode  ah sorry for the late reply: I'm not too worried about main development on Ubuntu 20.04 – people using Ubuntu 20.04 should very likely use the nicely maintained 3.10 release series ;)

willcode:
I still think of `main` as `maint-3.10-dev`, since we don't know if there will every be a 3.11. Non-backported PRs are effectively orphaned on `main` and can potentially cause conflicts.

marcusmueller:
that's more than fine and logical, but that would still mean that for forward development of the maint-3.10 branch we at some point need to drop the Ubuntus we supported on the day 3.10 was released, right. That would arguably be the point where we need to actually do a 3.11… but I mean, speculation.

[...]

PR #7326, opened 2024-05-07
Update FREEDESKTOP CMakeLists.txt

Ech0xing:
A fix for #7322 I updated the CMakeList.txt under 'gnuradio/grc/scripts/freedesktop' 

CMake indeed does not recognize a PERMISSIONS sub-command.

- I've used the file(COPY ... FILE_PERMISSIONS ...) command. This command is capable of copying a file and setting its permissions in one step.
- Made sure the paths are correct and that the file grc_setup_freedesktop.in is present in the source directory.

This should help to set the file permissions correctly after configuring the file with configure_file.

## Checklist

### DIFF
--- a/grc/scripts/freedesktop/CMakeLists.txt
+++ b/grc/scripts/freedesktop/CMakeLists.txt
@@ -13,9 +13,14 @@ if(UNIX AND HAVE_XDG_UTILS AND ENABLE_POSTINSTALL)
     configure_file(
         ${CMAKE_CURRENT_SOURCE_DIR}/grc_setup_freedesktop.in
         ${CMAKE_CURRENT_BINARY_DIR}/grc_setup_freedesktop
+        @ONLY
+    )
+    file(
+        COPY ${CMAKE_CURRENT_SOURCE_DIR}/grc_setup_freedesktop.in
+        DESTINATION ${CMAKE_CURRENT_BINARY_DIR}
         FILE_PERMISSIONS OWNER_READ OWNER_WRITE OWNER_EXECUTE GROUP_READ GROUP_EXECUTE WORLD_READ WORLD_EXECUTE
-    @ONLY)
-   install (
+    )
+    install (
         CODE "execute_process(COMMAND ${CMAKE_CURRENT_BINARY_DIR}/grc_setup_freedesktop install)"
     )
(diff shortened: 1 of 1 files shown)


grc/scripts/freedesktop/CMakeLists.txt
```diff
@@ -13,9 +13,14 @@ if(UNIX AND HAVE_XDG_UTILS AND ENABLE_POSTINSTALL)
     configure_file(
         ${CMAKE_CURRENT_SOURCE_DIR}/grc_setup_freedesktop.in
         ${CMAKE_CURRENT_BINARY_DIR}/grc_setup_freedesktop
+        @ONLY
+    )
+    file(
+        COPY ${CMAKE_CURRENT_SOURCE_DIR}/grc_setup_freedesktop.in
```
marcusmueller, line 19: Hm, that seems to be a bit wrong, in this order, the `file(COPY` should overwrite the `configure_file` result, right?
Wouldn't it make more sense to 

1. `configure_file(${CMAKE_CURRENT_SOURCE_DIR}/grc_setup_freedesktop.in` to `${CMAKE_CURRENT_BINARY_DIR}/grc_setup_freedesktop.tmp` (so, fill in the variables in the .in file, and put the result in build/…)
2. `file(COPY ${CMAKE_CURRENT_BINARY_DIR}/grc_setup_freedesktop.tmp ${CMAKE_CURRENT_BINARY_DIR}/grc_setup_freedesktop FILE_PERMISSIONS  …` to set the permissions?

(instead of doing the copy, it'd be more logical to do a `file(CHMOD…`, but that's a CMake 3.19 addition, and the fact that you don't have a very modern CMake was the original problem)

Ech0xing: There is always a better way. Let me test it and update you with the result.

willcode:
We accept changes against `main`, but not against the maint branches. I'll close this here - please resubmit against `main`.

grc/scripts/freedesktop/CMakeLists.txt
marcusmueller: Re: Ubuntu: there's still a bug with us, if we need newer CMake than we declare that we need; see #7323 

Re: Better way: I think your approach here is wrong, not suboptimal :)

PR #7339, opened 2024-05-15
modtool: replace pkg_resources with import.metadata when available (backport to maint-3.10)

willcode:
Backport #7317


marcusmueller:
superseeded by #7342

PR #7350, opened 2024-05-17
make grc use correct config paths

### LINKED ISSUE #7337: GRC ignores gr::prefs directories
### What happened?

I've got my GNU Radio configuration (as reported by `gnuradio-config-info --userprefsdirs` in ~/.config/gnuradio

Still, starting GRC and making settings creates a ~/.gnuradio folder and puts config files in there.

### System Information

Linux F40 x86

### GNU Radio Version

3.11-git (main)

### Specific Version

v3.11.0.0git-744-g5776fd3c8

1. remove  ~/.gnuradio, have all config in ~/.config/gnuradio
2. start GRC

### Relevant log output

_No response_

marcusmueller:
- **runtime/paths: add persistent() for persistent application data**
- **GRC: GR config & persistent state location instead of ~/.gnuradio || ~/.grc_gnuradio**

## Description

As discussed in [GRC Development](https://matrix.to/#/!BMFENoeifGOSJaeGco:gnuradio.org/$BblQe5QpHvTswR1Ypi42C1ZJ33i6xwfViBwH2KhBKkg?via=gnuradio.org&via=matrix.org&via=ei8fdb.org),
it would be the right thing if GRC configuration ends up in the same tree as
GNU Radio itself stores configuration. (And we're now doing `XDG_CONFIG_HOME`,
unless that doesn't exist, but the legacy `.gnuradio` does.)

Same for generated hier blocks: they should live in a standard persistent state
directory. And for caches from the example browser and block lib, use the
appropriate cache dir.

Luckily, GNU Radio offers these paths by now; `gr.paths` is your friend. Just
need to use them.

We're also including fallbacks, should importing that fail for whatever reason.
## Related Issues

Closes #7337

GRC Qt, GTK, converter

We add (no API breakage) a `persistent` getter to `sys_paths` and export it to python.

## Checklist

### DIFF
--- a/grc/core/platform.py
+++ b/grc/core/platform.py
@@ -133,7 +133,12 @@ class Platform(Element):
         self.cpp_connection_templates.clear()
         self._block_categories.clear()
 
-        with Cache(Constants.CACHE_FILE, version=self.config.version) as cache:
+        try:
+            from gnuradio.gr import paths
+            cache_file = os.path.join(paths.cache(), Constants.GRC_SUBDIR, Constants.CACHE_FILE_NAME)
+        except ImportError:
+            cache_file = Constants.FALLBACK_CACHE_FILE
+        with Cache(cache_file, version=self.config.version) as cache:
             for file_path in self._iter_files_in_block_path(path):
 
                 if file_path.endswith('.block.yml'):
--- a/grc/gui_qt/components/example_browser.py
+++ b/grc/gui_qt/components/example_browser.py
@@ -250,7 +250,12 @@ class ExampleBrowser(QWidget, base.Component):
     def find_examples(self, progress_callback, ext="grc"):
         """Iterate through the example flowgraph directories and parse them."""
         examples_dict = {}
-        with Cache(Constants.EXAMPLE_CACHE_FILE, log=False) as cache:
+        try:
+            from gnuradio.gr import paths
+            cache_file = os.path.join(paths.cache(), Constants.GRC_SUBDIR, Constants.EXAMPLE_CACHE_FILE_NAME)
+        except ImportError:
+            cache_file = Constants.FALLBACK_EXAMPLE_CACHE_FILE
+        with Cache(cache_file, log=False) as cache:
             for entry in self.platform.config.example_paths:
                 if entry == '':
                     log.info("Empty example path!")
--- a/grc/gui_qt/properties.py
+++ b/grc/gui_qt/properties.py
@@ -53,6 +53,7 @@ class Paths(object):
     TOP_BLOCK_FILE_MODE = HIER_BLOCK_FILE_MODE | stat.S_IXUSR | stat.S_IXGRP
 
     # Setup paths
+    # FIXME: this should use ..main.get_config_directory() and get_state_directory(), probably.
     '''
     HIER_BLOCKS_LIB_DIR = os.environ.get('GRC_HIER_PATH',
                                          os.path.expanduser('~/.grc_gnuradio'))
--- a/gnuradio-runtime/apps/gnuradio-config-info.cc
+++ b/gnuradio-runtime/apps/gnuradio-config-info.cc
@@ -1,6 +1,7 @@
 /* -*- c++ -*- */
 /*
  * Copyright 2009 Free Software Foundation, Inc.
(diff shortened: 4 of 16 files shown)


grc/gui_qt/properties.py
```diff
@@ -53,6 +53,7 @@ class Paths(object):
     TOP_BLOCK_FILE_MODE = HIER_BLOCK_FILE_MODE | stat.S_IXUSR | stat.S_IXGRP
 
     # Setup paths
+    # FIXME: this should use ..main.get_config_directory() and get_state_directory(), probably.
```
marcusmueller, line 56: @haakov  I wasn't sure what to do with the commented-out code. Do we get to keep it or should we clean it up?

grc/core/platform.py
```diff
@@ -133,7 +133,12 @@ def build_library(self, path=None):
         self.cpp_connection_templates.clear()
         self._block_categories.clear()
 
-        with Cache(Constants.CACHE_FILE, version=self.config.version) as cache:
+        try:
```
willcode, line 136: How can this fail?

grc/gui_qt/components/example_browser.py
```diff
@@ -250,7 +250,12 @@ def reset(self):
     def find_examples(self, progress_callback, ext="grc"):
         """Iterate through the example flowgraph directories and parse them."""
         examples_dict = {}
-        with Cache(Constants.EXAMPLE_CACHE_FILE, log=False) as cache:
+        try:
+            from gnuradio.gr import paths
```
willcode, line 254: Same comment.

willcode:
I guess that's a general comment ... why are we assuming that an import from our own package could fail?

grc/core/platform.py
marcusmueller: my mental model was that GRC is something we perspectively want to move out of the GR tree and make less tightly version-coupled.

In the wake of that, depending on the very recent addition of `gr.paths` (or gr.paths.persistent) is still sensible (because we're still having GRC in-tree), but falling back gracefully in this place makes it easier to use "new" GRC with "old" GRC in the future.

willcode:
Not for this PR, but a wiki writeup of all the "if here", nope, "look there", hmm, "maybe over here", "copy to new place", etc., logic recently added to GRC would be good.

willcode:
Since caching is just for startup performance, is there really a need to make that part backward compatible? Even switching back and forth between two GRC versions will just cause two caches to be kept up to date.

grc/gui_qt/properties.py
haakov: I kind of want to get rid of this file. We have Config.py, we have Constants.py, and we have Platform.py in the same directory, and I'm not quite sure what goes where. Maybe we can move the contents of this file to the others?

marcusmueller: OK, then let's do that, it's OK enough an effort to be done quickly; but let's keep that out of this PR, then. I guess that especially means that the commented out code on Lines 57ff will just be removed.

marcusmueller:
> Since caching is just for startup performance, is there really a need to make that part backward compatible? Even switching back and forth between two GRC versions will just cause two caches to be kept up to date.

You're right!

marcusmueller:
> Since caching is just for startup performance, is there really a need to make that part backward compatible? Even switching back and forth between two GRC versions will just cause two caches to be kept up to date.

Thinking about this some more: you're right, no need to be backwards-compatible (which is great for backporting, I guess!), but we still shouldn't crash if we are faced with the old API only. So, I don't think there's anything to change here :)

marcusmueller:
squashed

marcusmueller:
@willcode ping!

argilo:
It looks like this broke hierarchical blocks, because they are still loaded from `~/.grc_gnuradio` by default:

https://github.com/gnuradio/gnuradio/blob/2a8b58ca18b5bbd978ca6cd495450c2a9b4ea2b8/grc/core/generator/top_block.py#L157

For the moment, I was able to work around the problem by overriding the path from which hierarchical blocks are loaded:

```
GRC_HIER_PATH=~/.local/state/gnuradio gnuradio-companion
```

PR #7354, opened 2024-05-18
modtool: Remove comments from C++ iosignatures

### LINKED ISSUE #5222: gr_modtool makeyaml produces some invalid fields
### What happened?

gr_modtool makeyaml produces some invalid fields due to the regex parsing of the impl files

### System Information

OS: Ubuntu 20.10

### GNU Radio Version

3.10-git (master)

### Specific Version

395-gc3c411be

1) Create a block with gr_modtool
2) gr_modtool makeyaml myblock

The output yaml has a few issues - see log output

### Relevant log output

```shell
id: cudademo_multiply_const
label: multiply const
category: '[Cudademo]'
templates:
  imports: import cudademo
  make: cudademo.multiply_const(${k})
parameters:
- id: k
  label: K
  dtype: float
inputs:
- label: in
  domain: stream
  dtype: raw
  multiplicity: 1 /* max inputs */
outputs:
- label: out
  domain: stream
  dtype: raw
  multiplicity: 1 /*max outputs */
file_format: 1
```

marcusmueller:
## Description

Fix the case where the comments (which are present in gr-newmod) in iosignatures break YAML gen

## Related Issue

Fixes #5222

(At least for reasonable cases. Comments with nested comment symbols, or strings containing C comments, might lead to interesting results. But these are not likely to occur.

gr-modtool

## Testing Done

https://gist.github.com/marcusmueller/f03ff708e0975c82dc80c06156c1ae3b

(can't do automatic tests before we fix modtool to run non-interactively)

## Checklist

### DIFF
--- a/gr-utils/modtool/tools/parser_cc_block.py
+++ b/gr-utils/modtool/tools/parser_cc_block.py
@@ -21,6 +21,21 @@ def dummy_translator(the_type, default_v=None):
     return the_type
 
 
+def remove_comments(inp: str) -> str:
+    def comment_removal(match: re.match) -> str:
+        groups = match.groups()
+        if len(groups) == 1:
+            return ''
+        return inp
+    try:
+        out = re.sub(r'(/\*[^(\*/)]*\*/)', comment_removal, inp)
+        logger.debug(f"out: {out}")
+        return out
+    except Exception as e:
+        logger.error(f"error removing comments {e}")
+        return inp
+
+
 class ParserCCBlock(object):
     """ Class to read blocks written in C++ """
 
@@ -51,7 +66,7 @@ class ParserCCBlock(object):
             type_match = re.search(r'sizeof\s*\(([^)]*)\)', typestr)
             if type_match is None:
                 return self.type_trans('char')
-            return self.type_trans(type_match.group(1))
+            return self.type_trans(remove_comments(type_match.group(1)))
 
         def _typestr_to_vlen(typestr):
             """ From a type identifier, returns the vector length of the block's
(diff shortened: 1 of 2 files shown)


gr-utils/modtool/tools/parser_cc_block.py
```diff
@@ -78,18 +99,19 @@ def _typestr_to_vlen(typestr):
...
+        except Exception as e:
+            logger.error(f"Error: Can't parse input signature: {e}")
         try:
             iosig['out'] = _figure_out_iotype_and_vlen(iosig_match.group('outcall'),
                                                        iosig_match.group('outtype'))
-            iosig['out']['min_ports'] = iosig_match.group('outmin')
-            iosig['out']['max_ports'] = iosig_match.group('outmax')
+            iosig['out']['min_ports'] = remove_comments(iosig_match.group('outmin'))
+            iosig['out']['max_ports'] = remove_comments(iosig_match.group('outmax'))
         except Exception:
```
willcode, line 115: Make same exception format change as above?

gr-utils/modtool/tools/parser_cc_block.py
```diff
@@ -21,6 +21,27 @@ def dummy_translator(the_type, default_v=None):
     return the_type
 
 
+def remove_comments(inp: str) -> str:
+    def comment_removal(match: re.match) -> str:
+        groups = match.groups()
+        if len(groups) == 1:
+            return ''
+        if len(groups) % 3 == 0:
```
willcode, line 29: What does this do?

marcusmueller: ah, good catch; that was from a more unified time, when I had a single regex that could rework `2 /* channels per polarizations */ * 2 /* polarizations */` into `2  * 2 ` by replacing a 

    (([not comment])*)((comment)*)(([not comment])*)

match by the not-comment parts. That was to enable nested comment starts, `1 /* /* comment */` being correctly removed, and to disable `"/*comment*"` in strings being removed, but then specialties of python regexes and having to implement the C escaping rules for quotes came in the way. I didn't want to maintain half a C++ tokenizer in a regex, so here I pareto'd and accepted that we'll be able to handle 
`3 /* max inputs */` and `2 /* channels per polarizations */ * 2 /* polarizations */`, but not the problem cases mentioned above. 


The group length check and 3-handling is a remnant of that, gimme a second.

gr-utils/modtool/tools/parser_cc_block.py
marcusmueller: :+1:

gr-utils/modtool/tools/parser_cc_block.py
```diff
@@ -21,6 +21,21 @@ def dummy_translator(the_type, default_v=None):
     return the_type
 
 
+def remove_comments(inp: str) -> str:
+    def comment_removal(match: re.match) -> str:
```
willcode, line 25: Minor, but how about dropping this code back into the outer function. The subfunc doesn't serve any purpose any more and makes it harder to read.

marcusmueller: not quite sure what you mean here - the callable needs to see `inp`, so it needs to be local

marcusmueller: Made it simpler; no need for the function aside from really pathological corner cases, so omit.

willcode:
@marcusmueller could you squash and sign?

Issue #7360, opened 2024-05-21
Function Probe is broken due to wrong order of code


nemanjan00:
### What happened?

I used Signal Probe and Probe Function blocks. 

Generated python code fails to run, due to `object has no attribute` error

### System Information

OS: Arch Linux
GR Installation Method: Native distro package (3.10.10.0-2)

### GNU Radio Version

3.10 (maint-3.10)

### Specific Version

3.10.10.0

### Steps to Reproduce the Problem

Use both of mentioned blocks

Generated code:

https://gist.github.com/nemanjan00/058d42e4a878c9f3a91e78461663f8af

Line 87 should be after 119 
Line 76 refers to `self.freq_pro` which is set only on 119

Or 76 can be moved into try block

Looking at the `grc/blocks/variable_function_probe.block.yml`, I do not see any way to make sure code is generated in the proper order

Looking a the `grc/core/generator/top_block.py`, looks like blocks are sorted by name (line 213)

### Explanation

The issue is that Function Probe block is generated before Probe Signal, so when thread starts reading Signal Probe, it is not defined on class...

willcode:
@haakov @dl1ksv if you get a chance, this is a race condition in the python generated by GRC.

willcode:
Additional notes in #7360 about potential solutions.

PR #7361, opened 2024-05-21
Fix issue with out of order creation of code for function probe #7360

nemanjan00:
## Description
I created a way for block to detect if class was initialized

## Related Issue
This fixes issue #7360 

It affects Function Probe block and global template for generating flow graph in python

## Testing Done
I did not test it yet, since I am going to sleep, just wanted to do it so I do not forget it

## Checklist

### DIFF
--- a/grc/blocks/variable_function_probe.block.yml
+++ b/grc/blocks/variable_function_probe.block.yml
@@ -34,6 +34,8 @@ templates:
     make: |+
         def _${id}_probe():
           while True:
+            if self.initialized != True:
+              continue
             <% obj = 'self' + ('.' + block_id if block_id else '') %>
             val = ${obj}.${function_name}(${function_args})
             try:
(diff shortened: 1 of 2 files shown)


willcode:
That creates a busy waiting loop, and `initialized` is not initialized (to False).

In any case, if we do this, it will be with something like `threading.Condition`. The `self.flowgraph_initialized.wait()` would be before the `while True:`, and `self.flowgraph_initialized.notify_all()` would be at the end of the template. Somewere earlier in the template, `flowgraph_initialized` would be created.

A couple of general notes for submissions:
- commit messages start with the module name, e.g., `grc: blah, blah`
- commits must have a git signature, which is the authors acknowledgement of the GNU Radio DCO (see the bottom of https://wiki.gnuradio.org/index.php?title=Development)

willcode:
PRs are accepted against the `main` branch, not directly against maintenance branches. We backport to maint when appropriate, unless the backport is difficult (which is good to avoid) and author assistance is needed.

dl1ksv:
@nemanjan00 Can you provide a simple grc file for testing ?

willcode:
I'm going to close this since it's against maint-3.10 and needs to be changed to use threading synchronization objects. Please resubmit against the main branch.

PR #7377, opened 2024-05-30
Modtool: Non-Interactive Mode

### LINKED ISSUE #4735: gr_modtool command line arguments are incomplete or broken
Using gr_modtool from the command line, we can provide command-line values to skip asking questions.  For example, there is `--add-python-qa` and `--add-cpp-qa` but there is no method by which to select *only* python-qa or *only* cpp-qa.  The reason is that there is a way to ask for the inclusion of tests, but no way to specify the exclusion of tests.  

I would propose either adding `--no-add-python-qa` and `--no-add-cpp-qa` or changing their defaults to be `False` rather than `None`.  I'd be happy to submit a patch, depending on which way we want to go with this.

marcusmueller:
- **modtool: if not interactive, don't assume 'yes'**
- **modtool: default SCM config auto instead of 'no'**
- **SQUASHME: modtool: add option for batch processing**
- **modtool: conversion from info dict to descriptive object: add object skeleton**
- **fixup! modtool: conversion from info dict to descriptive object: add object skeleton**
- **modtool: convert from dictionary to initialized object for tool info**
- **modtool: batch mode; tests; fail if mandatory args missing**

## Description

modtool has the problem of inconsistent argument handling in terms of being automatically enabled / disabled.
This currently makes it impossible to batch-add blocks, which makes command line CI impossible to write.

This PR adds a `--batch` (`-b`) command line option, and a `--no` option. `--no` disables every binary flag (add QA tests etc) unless specifically specified. `--batch` goes one step further and fails if necessary arguments are missing.

## Related Issue

Closes #4735

modtool, test-ability

## Testing Done

added a shell script to test `--batch`.

## Checklist


marcusmueller:
@ryanvolz does this: 

https://github.com/gnuradio/gnuradio/actions/runs/9339079011/job/25702994417?pr=7377#step:3:6624

mean that we're not correctly `#!`-ing the gr_modtool script on Windows?

ryanvolz:
> @ryanvolz does this:
> 
> https://github.com/gnuradio/gnuradio/actions/runs/9339079011/job/25702994417?pr=7377#step:3:6624
> 
> mean that we're not correctly `#!`-ing the gr_modtool script on Windows?

Windows seems fine. This in the Linux log:
```
241/267 Test #244: test_001_modtool_batch_cli ...................***Failed    0.11 sec
$PREFIX/bin/gr_modtool
Check that we fail if make a new module in batch mode w/o name
$PREFIX/bin/gr_modtool: line 10:  A tool for editing GNU Radio out-of-tree modules. : command not found
$PREFIX/bin/gr_modtool: line 13: from: command not found
$PREFIX/bin/gr_modtool: line 19: syntax error near unexpected token `except'
$PREFIX/bin/gr_modtool: line 19: `    except KeyboardInterrupt:'
Make new module
$PREFIX/bin/gr_modtool: line 10:  A tool for editing GNU Radio out-of-tree modules. : command not found
$PREFIX/bin/gr_modtool: line 13: from: command not found
$PREFIX/bin/gr_modtool: line 19: syntax error near unexpected token `except'
$PREFIX/bin/gr_modtool: line 19: `    except KeyboardInterrupt:'
```
which seems to indicate that `gr_modtool` is being run as a bash script, or at least not Python. Installing `gr_modtool` before running it in this test should be turning the shebang into `#!$PREFIX/bin/python`, so I don't know what's going wrong.

Weird that macOS and Windows are happy but Linux is not. Maybe those are just exiting because `command -v gr_modtool` fails?

marcusmueller:
> Maybe those are just exiting because command -v gr_modtool fails?

aargh, never assume that just because a thing isn't skipped on your machine it's not skipped in CI… you're right. Let me fix that.

marcusmueller:
So, I just tried setting up a conda environment in a clean debian container, and there, the gr_modtool binary, as installed, is perfectly marked as to be executed with the conda env's bin/python (which is 3.12.3, in this case). 

So, really don't know why the Linux conda build container tries to run that as shell script, and honestly, out of ideas.

ryanvolz:
Ok, I finally know what's happening. The shebang line is too long thanks to the fact that conda-build purposely uses a long environment path for binary substitution purposes, and the interpreter gives up and assumes the script is shell. This gets fixed when `gr_modtool` is packaged because the long build-time prefix is replaced by the install-time prefix, which is a sane length. And nobody has tried to run any of these Python scripts from the build-time conda environment until now.

The fix I would propose is to rip out [the `GR_PYTHON_INSTALL` logic that substitutes `#!/usr/bin/env python` for the full path to the Python interpreter](https://github.com/gnuradio/gnuradio/blob/975c5d5ec2f06751b99148e5ae8081a3a9d70751/cmake/Modules/GrPython.cmake#L360). Conda would be perfectly happy if the shebang remained `#!/usr/bin/env python`, and I suspect everyone else would be too.

marcusmueller:
wow, what sleuthing! Kudos!

Don't think we can fully rip out that replacement logic (still need to use the set / detected python interpreter e.g. when cross-compiling. Also, I just tried, on a fresh debian 12 with Python3 installed, `python` isn't installed (`python3` is), so `/usr/bin/env python` would fail. Easy fix there  – use `python3`).

>  and I suspect everyone else would be too.

I'm afraid we introduced that logic exactly because people needed to specify the python they're using! So, I'm afraid this really is not an option; we could simply set `-DPYTHON_EXECUTABLE` on the cmake command line of the CI runner to a wrapper that calls `/long/path/to/python3.12`.

Inspired by your effort, I dug deeper:

The systematic problem here is, that when I set up a test (debian12) container, installed miniforge, conda-setup a new env, and installed the `gnuradio` package in that using `mamba`, I found that the same full path to the python interpreter is used by non-gnuradio scripts, such as `pip`: 
```
head -n1 ~/miniforge/envs/gnuradio/bin/pip
#!/root/miniforge3/envs/gnuradio/bin/python3.12
```

Not a problem on my system, because that's shorter than 127 characters, but it basically means that it's no use that we plug this on our end – it's a general conda problem; the only solution to me seems to be making sure the conda prefix isn't longer than 127-2-strlen("/bin/pythonX.XX").

willcode:
Do you guys want to go with the test failure for now, or fix it one way or another before merging?

ryanvolz:
> Not a problem on my system, because that's shorter than 127 characters, but it basically means that it's no use that we plug this on our end – it's a general conda problem; the only solution to me seems to be making sure the conda prefix isn't longer than 127-2-strlen("/bin/pythonX.XX").

Figuring that there's no way that conda has been failing for years for people who install into long prefixes, I did some more digging and found that what conda does is it [shortens shebangs](https://github.com/conda/conda/blob/main/conda/core/portability.py#L258) during the installation step when they would be too long. That only happens once something has been packaged, and then during installation of the package. It doesn't help when you're trying to run something within the build environment.

> Do you guys want to go with the test failure for now, or fix it one way or another before merging?

I don't like having failures, but it's up to @marcusmueller about how much additional effort he wants to put into this. I would propose either adding a CMake flag that would allow the conda build to skip the shebang substitution, or putting in some extra logic that checks for a too-long shebang and keeps it at `#!/usr/bin/env python3` in that case.

marcusmueller:
> Figuring that there's no way that conda has been failing for years for people who install into long prefixes, I did some more digging and found that what conda does is it [shortens shebangs](https://github.com/conda/conda/blob/main/conda/core/portability.py#L258) during the installation step when they would be too long. That only happens once something has been packaged, and then during installation of the package. It doesn't help when you're trying to run something within the build environment.

Oh! Gotta think about that. It feels very inconsistent to do one thing on a long path length, and another on a shorter path length. After all, things will fail (or work) differently if you called python scripts from an environment before you first activate it. 
 
> > Do you guys want to go with the test failure for now, or fix it one way or another before merging?
> 
> I don't like having failures, but it's up to @marcusmueller about how much additional effort he wants to put into this. I would propose either adding a CMake flag that would allow the conda build to skip the shebang substitution, or putting in some extra logic that checks for a too-long shebang and keeps it at `#!/usr/bin/env python3` in that case.

Because the maximum shebang line length is again platform-dependent, and because on non-conda things, the default `env python` often is not what you need, I'd rather not go down that road. 

But I hate failures, too. I think the quickest solution is to copy our python test wrapper generator for a shell script wrapper, setting a `GR_PYTHON_INTERPRETER` or such, so that we can directly call `${CMAKE_CURRENT_SOURCE_DIR}/../gr_modtool`.  I wish the goal here was to actually test just the python module – but it is to run the CLI and check for it being actually non-interactive, so sadly no just importing the `modtool` module.

gr-utils/modtool/core/disable.py
```diff
@@ -100,43 +100,45 @@ def _handle_cc_qa(cmake, fname):
...
-                    blockname_pattern = f"^{self.info['modname']}_{self.info['blockname']}.block.yml$"
+                    blockname_pattern = f"^{self.info.modname}_{self.info.blockname}.block.yml$"
                 if blockname_pattern:
                     filenames = cmake.find_filenames_match(blockname_pattern)
-            elif self.info['pattern']:
-                filenames = cmake.find_filenames_match(self.info['pattern'])
-            yes = self.info['yes']
+            elif self.info.pattern:
+                filenames = cmake.find_filenames_match(self.info.pattern)
+            yes = (not self.info.no) and self.info.yes
```
mbr0wn, line 128: Same here, I think ToolConfig could catch invalid --no --yes settings, and make sure that `info.no` and `info.yes` are valid.

gr-utils/modtool/cli/add.py
```diff
@@ -140,17 +148,17 @@ def get_arglist(self):
 def get_py_qa(self):
     """ Get a boolean value for addition of py_qa """
     if self.add_py_qa is None:
-        if not (self.info['blocktype'] in ('noblock') or self.skip_subdirs['python']):
+        if not (self.info.no or self.info.blocktype in ('noblock') or self.skip_subdirs['python']):
             self.add_py_qa = ask_yes_no(click.style(
                 'Add Python QA code?', fg='cyan'), True)
         else:
-            self.add_py_qa = False
+            self.add_py_qa = bool(self.info.yes)
```
mbr0wn, line 155: Shouldn't `info.yes` always be a bool?

gr-utils/modtool/cli/base.py
```diff
@@ -137,11 +136,15 @@ def common_params(func):
     @click.option('--skip-grc', is_flag=True,
                   help="Don't do anything in the grc/ subdirectory.")
     @click.option('--scm-mode', type=click.Choice(['yes', 'no', 'auto']),
-                  default=gr.prefs().get_string('modtool', 'scm_mode', 'no'),
-                  help="Use source control management [ yes | no | auto ]).")
+                  default=gr.prefs().get_string('modtool', 'scm_mode', 'auto'),
```
mbr0wn, line 139: Did you sneak a non-related behavioural change in here?

gr-utils/modtool/cli/base.py
```diff
@@ -110,6 +108,7 @@ def setup_cli_logger(logger):
     """ Sets up logger for CLI parsing """
     try:
         import colorama
+        colorama  # Make linters happy: uses colorama name, but does nothing
```
mbr0wn, line 111: Which linters are these, and are we mandating them now in GR?

mbr0wn (commented):
At first glance, this looks good. I hope you get those shebang issues resolved.

mbr0wn:
@marcusmueller in the non-conda case, the failures seem unrelated to long paths? Can we just fix the tests for non-conda, and simply skip them for Conda?

mbr0wn:

PR #7382, opened 2024-06-03
grc-qt: Can only connect sink in port to source out port

dl1ksv:
## Description

Run

grc --qt
File -> new
Add a constant source block and a null sink block
Click the source out port and then the sink in port and you should get connection, but nothing happens. Now clicking again the source out port you get a connection.

File ->	new
Add a constant source block and a null sink block
Click the sink in  port and then the source out port and you immediately  get a connection.

Applying this pr you get immediately a connection when clicking the source out port and then the sink in port.

## Testing Done

Run the described examples

## Checklist

### DIFF
--- a/grc/gui_qt/components/canvas/flowgraph.py
+++ b/grc/gui_qt/components/canvas/flowgraph.py
@@ -284,14 +284,22 @@ class FlowgraphScene(QtWidgets.QGraphicsScene, base.Component):
             c_item = g_item.core
             if c_item.is_block:
                 self.moving_blocks = True
+                if event.button() == Qt.LeftButton:
+                    self.mousePressed = True
+                    super(FlowgraphScene, self).mousePressEvent(event)
+                    return
             elif c_item.is_port:
                 new_con = None
                 if len(selected) == 1:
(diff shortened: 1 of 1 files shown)


dl1ksv:
You are right.  It is weird. It's a bit more complicated.

Run grc --qt
File -> new

Add source and sink.

Save the fg and exit from grc.
Now rerun grc --qt

Now try to connect source out with sink in. Nothing happens.
Clicking again on source out setups the connection. But this is just the same as directly clicking sink in and then source out.

Another way is:
Click twice source out. Avoid a double click . Then click sink in.

I what will happen through the restart

haakov:
@dl1ksv It looks to me like this creates some new bugs. I can't unselect a block by clicking in the canvas, and the QRubberband does not work anymore. It works again if I replace https://github.com/gnuradio/gnuradio/blob/2e59108989fb8e06a54abee0f23cba33827e0b2f/grc/gui_qt/components/canvas/flowgraph.py#L313 with `super(FlowgraphScene, self).mousePressEvent(event)`

It also causes some of the tests in test_qtbot to fail

PR #7383, opened 2024-06-04
Draft: gr-blocks: Add variable save restore block

eltos:
## Description

Allows saving multiple variables and restoring their values at a later point. Multiple slots are available to support different sets of values in parallel.

Useful to save the UI state or to quickly switch between different configurations.

Preferably to be used with variables from the following widgets, which reflect value changes on the UI:
- QT GUI Entry
- QT GUI Numeric Entry (#7349)
- QT GUI Chooser
- QT GUI Check Box

Related to the *Variable Config* block, however it is different because *Variable Config* actually creates a (single) variable, so it cannot be a GUI widget variable at the same time and thus not save/restore UI states.

## Related Issue

none

adds a new block to gr-blocks (categorized under "Variables")

grafik

grafik

## Testing Done

In my own OOT. Added an example flow graph to this PR

## Checklist

### DIFF
--- a/gr-blocks/python/blocks/variable_save_restore.py
+++ b/gr-blocks/python/blocks/variable_save_restore.py
@@ -1,0 +1,143 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+#
+# Copyright 2023 Philipp Niedermayer.
+#
+# SPDX-License-Identifier: GPL-3.0-or-later
+#
+
+
+import numpy
+from gnuradio import gr
+import pickle
+import pmt
+import os
+
+
+class variable_save_restore(gr.sync_block):
+    """
+    (see description in this blocks yml file)
+    """
+
+    def __init__(self, top_block=None, slot="default", variables="", save_trigger=False, restore_trigger=False, description_variable=None):
+        gr.sync_block.__init__(self, name="variable_save_restore", in_sig=[], out_sig=None)
+        self.message_port_register_in(pmt.intern("save"))
+        self.set_msg_handler(pmt.intern("save"), self.handle_msg_save)
+        self.message_port_register_in(pmt.intern("restore"))
+        self.set_msg_handler(pmt.intern("restore"), self.handle_msg_restore)
+
+        self.filename = f"~/.config/gnuradio/{top_block.__class__.__name__}.variables.sav"
+        self.filename = os.path.expanduser(self.filename)
+        print(f"[VariableSaveRestore] Config file: {self.filename}")
+        os.makedirs(os.path.dirname(self.filename), exist_ok=True)
+        self.load_storage()
+
+        self.top_block = top_block
+        self.description_variable = description_variable
+        self.set_slot(slot)
+        self.set_variables(variables)
+        self.save_trigger = save_trigger
+        self.restore_trigger = restore_trigger
+
+    def set_slot(self, slot):
+        self.slot = slot
+        if self.description_variable:
+            data = self.storage.get(self.slot, {})
+            info = data.get(self.description_variable, "")
+            self.set_variable_values({self.description_variable: info})
+
+    def set_variables(self, variables):
+        if isinstance(variables, str):
+            variables = [v.strip() for v in variables.split(",") if v.strip()]
+        self.variables = variables
+        if self.description_variable:
+            self.variables.append(self.description_variable)
+
+    def set_save_trigger(self, save_trigger):
+        if save_trigger and not self.save_trigger:
+            self.save()
+        self.save_trigger = save_trigger
+
+    def set_restore_trigger(self, restore_trigger):
+        if restore_trigger and not self.restore_trigger:
+            self.restore()
+        self.restore_trigger = restore_trigger
+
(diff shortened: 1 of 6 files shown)


willcode:
Save/restore is a great idea. I wish we had consistancy across the blocks so users would know what to expect when doing a save/restore.

gr-blocks/python/blocks/variable_save_restore.py
```diff
@@ -0,0 +1,143 @@
...
+    """
+
+    def __init__(self, top_block=None, slot="default", variables="", save_trigger=False, restore_trigger=False, description_variable=None):
+        gr.sync_block.__init__(self, name="variable_save_restore", in_sig=[], out_sig=None)
+        self.message_port_register_in(pmt.intern("save"))
+        self.set_msg_handler(pmt.intern("save"), self.handle_msg_save)
+        self.message_port_register_in(pmt.intern("restore"))
+        self.set_msg_handler(pmt.intern("restore"), self.handle_msg_restore)
+
+        self.filename = f"~/.config/gnuradio/{top_block.__class__.__name__}.variables.sav"
```
willcode, line 29: We now have `gr.paths.userconf()`, which returns `.../.config/gnuradio` or a different path if specified by `GR_PREFS_PATH` or derived from `XDG_CONFIG_HOME`.

gr-blocks/python/blocks/variable_save_restore.py
```diff
@@ -0,0 +1,143 @@
...
+        data = self.storage.get(self.slot, {})
+        data = {k: v for k, v in data.items() if k in self.variables}
+        n = self.set_variable_values(data)
+        print(f"[VariableSaveRestore] Restored {n} of {len(self.variables)} variables from slot {self.slot}")
+
+    def load_storage(self):
+        """Load from disk into memory"""
+        if os.path.exists(self.filename):
+            with open(self.filename, 'rb') as handle:
+                self.storage = pickle.load(handle)
```
willcode, line 132: Since we use YAML elsewhere, we shouldn't use pickle for serialization. This would be the only use.

gr-blocks/python/blocks/variable_save_restore.py
```diff
@@ -0,0 +1,143 @@
...
+        else:
+            self.storage = {}
+
+    def save_storage(self):
+        """Save from memory to disk"""
+        with open(self.filename, 'wb') as handle:
+            pickle.dump(self.storage, handle)
+
+    def work(self, input_items, output_items):
```
willcode, line 141: Does this ever get called if there are no stream inputs or outputs?

willcode (commented):
@haakov @dl1ksv I left a few comments, but this needs some thought. It's a good idea in general. It may be a little confusing, and as noted by the author only certain gui elements update their display on restore. It's also adding to the python features that aren't mirrored in c++.

gr-blocks/python/blocks/variable_save_restore.py
eltos: Thanks, changed

gr-blocks/python/blocks/variable_save_restore.py
eltos: As you like. Changed to yaml safe load

gr-blocks/python/blocks/variable_save_restore.py
eltos: No. We can as well remove it.

gr-blocks/python/blocks/variable_save_restore.py
willcode: Too bad they don't define those tag constants anywhere. For the tuple one, is it important to you that it stays a tuple and not SafeRepresenter's mapping to a list?

eltos: Yes, save/restore must not unexpectedly change data types.

gr-blocks/python/blocks/variable_save_restore.py
```diff
@@ -0,0 +1,155 @@
...
+    def __init__(self, top_block=None, slot="default", variables="", save_trigger=False, restore_trigger=False, description_variable=None):
+        gr.sync_block.__init__(self, name="variable_save_restore", in_sig=[], out_sig=None)
+        self.message_port_register_in(pmt.intern("save"))
+        self.set_msg_handler(pmt.intern("save"), self.handle_msg_save)
+        self.message_port_register_in(pmt.intern("restore"))
+        self.set_msg_handler(pmt.intern("restore"), self.handle_msg_restore)
+
+        self.filename = os.path.join(gr.paths.userconf(), "saverestore", f"{top_block.__class__.__name__}.yml")
+        self.filename = os.path.expanduser(self.filename)
+        print(f"[VariableSaveRestore] Config file: {self.filename}")
```
marcusmueller: this should be a logging call, but I'm not willing to hold off a PR for that

marcusmueller:
we need QA for this!

marcusmueller:
rebase atop #7947 when that's merged and rerun CI

marcusmueller:
Need to merge #8096 into main first

PR #7386, opened 2024-06-06
gr-uhd: Do not require PyQt5 for all python functionality

gmeyerho:
The push-button helper for the replay block uses PyQt5, however using gr-uhd in python as a whole should not be gated by PyQt5 being installed.

## Description
During some recent development, we ran into an error requiring PyQt5 being installed because the added replay block message button helper now added in an import for PyQt5, however everything else in gr-uhd doesn't require this python module.

To work around this, I added a try-except to ignore the module import error. Then I also removed the installation of the replay block message button related items in the cmake files, so that a user doesn't see functionality that they cannot use.

gr-uhd

## Testing Done
Testing importing gnuradio.uhd no longer errors if PyQt5 is not installed, and checked that the related items are not installed (i.e. .py file, grc example, block not available in UHD->RFNoC tree in GRC). Then installed PyQt5 module, re-built and confirmed functionality by running the previously uninstall replay block GRC example.

## Checklist

### DIFF
--- a/gr-uhd/CMakeLists.txt
+++ b/gr-uhd/CMakeLists.txt
@@ -17,6 +17,8 @@ else()
     set(UHD_FOUR_POINT_OH_RFNOC FALSE)
 endif()
 
+gr_python_check_module("PyQt5" PyQt5 True PYQT5_FOUND)
+
 ########################################################################
 # Register component
 ########################################################################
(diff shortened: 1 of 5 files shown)


gmeyerho:
The windows ctest failures appear to be happening in main as well: https://github.com/gnuradio/gnuradio/actions/runs/9436056687

marcusmueller:
yep, that failure is due to reasons outside your control, nevermind that :)

willcode:
@gmeyerho @mbr0wn This looks safe and only affects uhd - should I merge?

mbr0wn (approved):
@willcode yes please! Thanks!

Issue #7347, opened 2024-05-17
grc-qt selecting input and output of the same block creates upstream connection causing one upstream block error


chrisjohgorman:
### What happened?

I selected the input and output of a throttle block connected in a flowgraph.  This adds a connection from the block preceding the throttle block in the flowgraph to the throttle block.  This connection can be deleted by selecting the arrow going into the throttle block and hitting delete.

I expected that no operation would be performed.

### System Information

OS: Arch Linux
GR Installation Method: Source

### GNU Radio Version

3.10 (maint-3.10)

### Specific Version

3.10.10.0

### Steps to Reproduce the Problem

1) make a small flowgraph of three blocks, constant source, throttle and null sink.
2) connect all the blocks (constant source -> throttle -> null sink).
3) click on the input of the throttle and the output of the throttle block.
There should now be two connections to the input of the throttle from the constant source.
Here is a picture of the flowgraph now.
![qt_two_upstream_blocks1](https://github.com/gnuradio/gnuradio/assets/29354995/1ff05374-610a-4bd0-b60b-2ab393cda0b5)
This is the error shown by the errors button.
![qt_two_upstream_blocks](https://github.com/gnuradio/gnuradio/assets/29354995/4b97a4c8-4505-4b4e-b55d-f4b08e881e2e)

willcode:
@chrisjohgorman Are the summary and "steps to reproduce" describing the same problem?

chrisjohgorman:
Hi @willcode Sorry for the delay in replying.  I just noticed this in my inbox this morning.  The summary and "steps to reproduce" should be describing the same problem.  Basically I found that if had a small three element flowgraph, I could make two connections from the first block in the flowgraph to the second block by following the steps to reproduce.  This appropriately causes the error message 'Domain "stream" can only have one upstream block'.  Perhaps I should have described this more clearly in the summary.  Does this explanation more clearly describe the issue?

dl1ksv:
This issue should be fixed with #7402

Issue #7345, opened 2024-05-16
grc-qt deleting block that connects to two or more blocks leaves floating arrow lines


chrisjohgorman:
### What happened?

When replacing a block in the flowgraph then deleting one it leaves connection arrows without any associated pair of blocks.  It is most easily seen in the steps to reproduce the problem.

### System Information

OS: Arch Linux
GR Installation Method: Source

### GNU Radio Version

3.10 (maint-3.10)

### Specific Version

3.10.10.0

### Steps to Reproduce the Problem

1) make a three block flowgraph. (I used constant source, throttle and null sink)
2) add a second (null) sink and connect the throttle to the null sink.
3) now add a second throttle block and connect to the constant source and both null sinks but don't disconnect the first throttle block.  
4) disable the first throttle block.
(1-4) see first attached picture
![qt_arrow_pointer2](https://github.com/gnuradio/gnuradio/assets/29354995/dc08a56c-322d-45db-bcb5-26a87141ae93)
5) now enable the first throttle block and then delete the second throttle block
After 5 we're left with the following flowgraph.  Note the connection arrow to the first null block that is floating without any blocks connected to it.
![qt_arrow_pointer3](https://github.com/gnuradio/gnuradio/assets/29354995/3bd57cb1-4fd9-4422-8358-aa7ff23b0ff6)

PR #7402, opened 2024-06-19
grc-qt: Better solution for #7382

dl1ksv:
## Description

#7382 contains some bugs.
See the comments there
This is an better approach to fix these bugs

## Checklist

### DIFF
--- a/grc/gui_qt/components/block_library.py
+++ b/grc/gui_qt/components/block_library.py
@@ -82,12 +82,6 @@ class LibraryView(QTreeView):
                 QUrl(prefix + label.replace(" ", "_"))
             )
 
-    def handle_clicked(self):
-        if self.isExpanded(self.currentIndex()):
-            self.collapse(self.currentIndex())
-        else:
-            self.expand(self.currentIndex())
-
     def contextMenuEvent(self, event):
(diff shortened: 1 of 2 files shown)


dl1ksv:
I tried to figure out, why there is a difference if you click an output port of a block  that was added directly from the blocklib or if the block was loaded with an existing grc file. I had no success.

But I found that
https://github.com/gnuradio/gnuradio/blob/7a0db9434d38ace044f096ae59e906b1f917d307/grc/gui_qt/components/canvas/flowgraph.py#L302-L304 
leads always to false in
https://github.com/gnuradio/gnuradio/blob/7a0db9434d38ace044f096ae59e906b1f917d307/grc/gui_qt/components/canvas/flowgraph.py#L324

But we need the dummy connection only in a mouse move event. So I changed the code accordingly.

That resulted in another problem. Suddenly double clicking a block in the blocklib added only once the selected block.
Further double clicks were used as two single clicks by handle_clicked.

Removing this function made the double click work as expected without removing the functionality of expanding / collapsing the blocklib view

dl1ksv:
Fixes: #7345

willcode:
Fixes #7347

PR #7412, opened 2024-06-27
fec: Add encode and decode functions to generic encoder bindings

noc0lour:
This allows to use all FEC encoders and decoders by just creating corresponding encoder and decoder objects in Python and without creating a flowgraph

## Description
Adds additional methods to the Python class, to call the generic_work functions with numpy buffers. 

fec/generic_encoder fec/generic_decoder

## Testing Done
I'd like to add some test, but first I'd like to see the opinion on having this kind of change in-tree.

## Checklist

### DIFF
--- a/gr-fec/python/fec/bindings/generic_decoder_python.cc
+++ b/gr-fec/python/fec/bindings/generic_decoder_python.cc
@@ -18,6 +18,7 @@
 /***********************************************************************************/
 
 #include <pybind11/complex.h>
+#include <pybind11/numpy.h>
 #include <pybind11/pybind11.h>
 #include <pybind11/stl.h>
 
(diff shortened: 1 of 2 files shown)


willcode:
Is testing the main use of this? These functions can't be used with a block that is in a flowgraph, since en/decoders can be stateful.

I sort of think we should just mirror C++ functionality in the Python wrappers. The generic en/decoder could get encode/decode methods. It may not change the wrapper if it mostly deals with buffers, but we won't be adding functions in the wrappers.

We could easily justify this as a way to test en/decoders outside a flowgraph. I don't know if we want to encourage non-block use of blocks in general, so it would be a test support function that you could use at your own risk.

noc0lour:
Yes, the main use I implemented this was to bypass flowgraphs to test in-/output of FEC encoders/decoders in Python. If it is desired that the method name matches the C++ method name I can also rename it to "generic_work" since it's not doing much but call this function with a provided input buffer and a newly constructed Numpy output buffer.

This would definitely interfere with blocks which run in flowgraphs, therefore this is only of use for testing/outside of flowgraphs.

willcode:
How about renaming the functions `test_encode` and `test_decode`. That would imply that this isn't a general purpose function, and it could be supported only in the pybind layer.

marcusmueller:
merged into main because I think testing waits on that? Not putting up a backport for this yet, as I'd agree that undocumentedly, these names might be a bit too confusing? Or is there an inheritance reason to use these names?

PR #7419, opened 2024-07-02
Add grc as submodule

ZakyHermawan:
## Description

This is the first step of grc separation,
make GRC as a submodule and change nothing

We need to have a separate gnuradio/grc repo, so the grc development can still be done into that repo.
Either, we fork the submodule repo,
or we make an empty gnuradio/grc repo, and let me make a PR from submodule repo (https://github.com/ZakyHermawan/grc) into gnuradio/grc repo. I prefer the second option, but it's up to you

## Testing Done

compile, build, and run

## Checklist


ZakyHermawan:
@skoslowski @haakov @willcode

ZakyHermawan:
Just fix the CI issues, I will update the PR (change submodule url and squash) once my [PR](https://github.com/gnuradio/gnuradio-companion) is merged

willcode:
I can't merge this to main because we need to be able to backport grc changes from main to maint-3.10. Changing to a submodule on maint-3.10 will mess up too many people. Need ideas.

ryanvolz:
I haven't been following this GSOC project, so take this drive-by comment with a grain of salt.

As a packager, git submodules are usually bad news. It's primarily a developer workflow, and so unless special care is taken, it usually only works well for developers and not users who don't want to have to deal with build systems. The key to "special care" is a strong definition of what the boundary is between the two repos. By that I mean:

1. Are these truly separate projects, with separate releases, and a definite API/ABI so that different versions of each can be used with each other (within reason)? Then they should be treated as separate projects in the packaging/CMake, and git submodules only make sense as a convenience for developers to keep existing workflows going. Here, this would probably mean that the GNU Radio packaging would know nothing about GRC, and the GRC repo would depend on GNU Radio like any other dependency. The only connection within the gnuradio repo would be the submodule and some CMake that detects if the GRC directory/submodule exists and does a build/install of that when a build/install of GNU Radio is done.
2. Are these really the same project, and the split is for some development reason but they are only ever released/used in a tightly coupled manner? In that case, the most convenient thing for packaging would be to re-bundle them in the source tarball and distribute as one project/release.

Again, not having really followed this, I think the situation is closer to (1) here? In that case, I think the CMake work to give packagers the option to really treat these as separate projects is key, in addition to the submodule workflow which lets developers keep building more or less the same as before. I also think it's fine to do that CMake work later, so long as it happens before a release.

willcode:
Note on (1) - the idea is that eventually GRC will not depend on GNU Radio. This will let GRC work for GR3, GR4, or some other workflow.

ZakyHermawan:
Yes, this is like the first step to separate grc from gnuradio codebase, at the end, there will be no grc inside gnuradio codebase, you need to install grc separately, just like volk

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "grc"]
+	path = grc
+	url = https://github.com/ZakyHermawan/grc.git
```
haakov, line 3: This needs to be updated

haakov (changes requested):

haakov:
@willcode I see your argument about being able to backport from main to maint-3.10. But doesn't this PR mean that we have to backport from the gnuradio-companion repo at some point either way?

I do agree that this is too drastic of a change to bring into maint-3.10 though.

@ryanvolz Yes, the situation is closer to (1). This GSoC project comes from [this idea](https://wiki.gnuradio.org/index.php?title=GSoCIdeas#GRC:_Standalone_application_and_pluggable_workflows). As Jeff said, the idea is that GRC will be independent of GNU Radio and ready to use for other applications too.

willcode:
If there's a way to make this so it's all backward compatible, with maybe a few cmake lines difference to change things to completely out of tree, then it could be backported. The dir structure would be the same as currently on main/maint-3.10 with some way of quickly changing it to become out of tree.

This is also tied up in where people want to go from here on main/maint. We aren't necessarily going to stay with the same thing we've been doing the last few years.

PR #7428, opened 2024-07-11
gui_prefs_file does not need to use os.path.join

chrisjohgorman:
When os.path.join is used, it appends an extra directory to the config file path.  For grc the config file is $HOME/.config/gnuradio/grc.conf/grc.conf and for grc qt the file is $HOME/.config/gnuradio/grc.conf/grc_qt.conf.

This patch proposes to change this and remove the grc.conf directory from the path.  This would make the grc config file $HOME/.config/gnuradio/grc.conf and the grc qt config file $HOME/.config/gnuradio/grc_qt.conf.

All that is needed to accomplish this is to remove the use of os.path.join to generate the gui_prefs_file.

## Description

The above change removes the use of os.path.join(get_config_file_path(), 'grc.conf') to set the variable gui_prefs_file in both grc and grc_qt.

This change is required because current method adds an extra grc.conf directory to the config file path.

This patch affects the grc/gui/Config.py and grc/gui_qt/Config.py.

## Testing Done

I moved the .config directory to .config.old.  Then I tested both with and without the patch on both windows msys2 and linux.

I could not see where this would break other areas of code, but I am new to making changes to gnuradio, so it is possible that this change is a mistake, although I wouldn't make a PR unless I thought it warranted. 

## Checklist

### DIFF
--- a/grc/gui/Config.py
+++ b/grc/gui/Config.py
@@ -28,8 +28,7 @@ class Config(CoreConfig):
 
     name = 'GNU Radio Companion'
 
-    gui_prefs_file = os.environ.get(
-        'GRC_PREFS_PATH', os.path.join(get_config_file_path(), 'grc.conf'))
+    gui_prefs_file = os.environ.get('GRC_PREFS_PATH', get_config_file_path())
 
     def __init__(self, install_prefix, *args, **kwargs):
         CoreConfig.__init__(self, *args, **kwargs)
--- a/grc/gui_qt/Config.py
+++ b/grc/gui_qt/Config.py
@@ -11,8 +11,7 @@ class Config(CoreConfig):
 
     name = 'GNU Radio Companion'
 
-    gui_prefs_file = os.environ.get(
-        'GRC_QT_PREFS_PATH', os.path.join(get_config_file_path(), 'grc_qt.conf'))
+    gui_prefs_file = os.environ.get('GRC_QT_PREFS_PATH', get_config_file_path('grc_qt.conf'))
 
     def __init__(self, install_prefix, *args, **kwargs):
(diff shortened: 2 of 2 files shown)


willcode:
Bug - I'd like to backport this before the final v.3.10.11.0, so could I get a quick sanity check from @marcusmueller @haakov or @dl1ksv ?

haakov (approved):

dl1ksv:
Some remarks:
I think, it's a good idea to move grc.conf and grc_qt.conf to ~/.config/gnuradio

But as I come from the main branch ( latest commit) I already have the **directories**
~/.config/gnuradio/grc.conf  
~/.config/gnuradio/grc_qt.conf
so I have to remove those directories manually.

After deleteing  these directories running grc --qt leads to
`Found specification for config path '/home/schroer/.config/gnuradio/grc_qt.conf', but file does not exist. Old default config file path '/home/schroer/.gnuradio/grc_qt.conf' exists; using that. Please consider moving configuration to new location.`

If I now modify the preferences and save them, I'd expect to save the modified preferences to the new location, but the go to the old one.

grc --qtk starts without any hints , but uses the old location and saves any modifications to the old location.
You should get the same message like  above.

I'd expect that in both cases the settings are taken from the old location and  written to the new location on exit..

At the moment you have to create an empty configuration file or copy the old files to the new location.

willcode:
Agree it's a bit complicated, and I don't know whether orphaning or deleting old confs would be a good idea either. If we expected people to do a single upgrade and never fall back, then doing a hard migration would definitely be the best thing. Switching between versions would create all kinds of unexpected behavior if we did that, though.

PR #7429, opened 2024-07-12
Small C++ modernization

gerstrong:
## Description
Just a small change that avoids running into initialization order problems when constructor init lists is used for default values

## Checklist

### DIFF
--- a/gr-video-sdl/lib/sink_s_impl.cc
+++ b/gr-video-sdl/lib/sink_s_impl.cc
@@ -40,16 +40,10 @@ sink_s_impl::sink_s_impl(
                  io_signature::make(0, 0, 0)),
       d_chunk_size(width * height),
       d_framerate(framerate),
-      d_wanted_frametime_ms(0),
       d_width(width),
       d_height(height),
       d_dst_width(dst_width),
-      d_dst_height(dst_height),
-      d_current_line(0),
-      d_screen(NULL),
-      d_image(NULL),
-      d_avg_delay(0.0),
-      d_wanted_ticks(0)
+      d_dst_height(dst_height)
 {
     if (framerate <= 0.0)
         d_wanted_frametime_ms = 0; // Go as fast as possible
--- a/gr-video-sdl/lib/sink_s_impl.h
+++ b/gr-video-sdl/lib/sink_s_impl.h
@@ -46,17 +46,17 @@ protected:
     int copy_plane_to_surface(int plane, int noutput_items, const short* src_pixels);
 
     float d_framerate;
-    int d_wanted_frametime_ms;
+    int d_wanted_frametime_ms = 0;
     int d_width;
     int d_height;
     int d_dst_width;
     int d_dst_height;
-    int d_current_line;
-    SDL_Surface* d_screen;
-    SDL_Overlay* d_image;
+    int d_current_line = 0;
+    SDL_Surface* d_screen = nullptr;
+    SDL_Overlay* d_image = nullptr;
     SDL_Rect d_dst_rect;
-    float d_avg_delay;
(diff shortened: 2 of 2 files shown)


willcode:
Was there a problem changing the order of the initializers?

gerstrong:
No, but whenever you change them you get a warning about initialization order. By removing init variables that get default values anyways you reduce that risk. Usually such warnings arise, when you decide to add another member somewhere in the class declaration and in the constructor init list you put it somewhere, messing up the order.

Why are order in init list important? https://www.geeksforgeeks.org/order-of-execution-in-initializer-list-in-c/

willcode:
Best I can tell, using initializer lists is the "modern" way, initializing in declaration order is important, and that's what the warning is for. For trivial values, there's not much advantaage to initializer lists, but if an expression is used for an initial value, then in-constructor initialization is required and use of an initializer list helps check the order. This is also the way we (try to) do things throughout the code.

If someone else has opinions on this, let me know. Otherwise, we'll stick with constructor lists and fix any broken ordering turned up by warnings.

gerstrong:
Similar discussion already here: https://stackoverflow.com/questions/36600187/whats-the-differences-between-member-initializer-list-and-default-member-initia

I have a preference for default member initialization and keeping constructors as slim as possible, especially if you have to declare different constructors and repeat defining the init lists with the same default variables. Anyway, it is up to you how to decide on this. It still would be nice though to use nullptr instead of NULL, but same here; up to you...

mbr0wn:
I have a tendency towards default member initialization because they tend to keep the code shorter, and the core guidelines recommend it: https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#c48-prefer-default-member-initializers-to-member-initializers-in-constructors-for-constant-initializers

...but not a strong one, and it does get tricky when you mix both styles (one attribute gets initialized up there, another down there).

@gerstrong If you wanted this PR merged, you should fix the commit message according to our guidelines (assuming @willcode  even wants to merge this).

willcode:
Given that the majority of GR uses initializaiton in the constructor (including gr-uhd), I don't really want to start making this change without the intent to update the codebase (huge undertaking). It would be too confusing. Even after this PR, the initialization of this class would have a mix of initialization based on the args to the constructor. This is practical for this particular (the only) constructor, but doesn't help with any kind of code goodness goal.

The nullptr changes are welcome. We're trying to gradually get rid of NULL.

gerstrong:
Well, I will make a new PR then. Feel free to reject this one.Von meinem iPhone gesendetAm 16.07.2024 um 13:15 schrieb Jeff Long ***@***.***>:﻿
Given that the majority of GR uses initializaiton in the constructor (including gr-uhd), I don't really want to start making this change without the intent to update the codebase (huge undertaking). It would be too confusing. Even after this PR, the initialization of this class would have a mix of initialization based on the args to the constructor. This is practical for this particular (the only) constructor, but doesn't help with any kind of code goodness goal.
The nullptr changes are welcome. We're trying to gradually get rid of NULL.

mbr0wn:
> Given that the majority of GR uses initializaiton in the constructor (including gr-uhd), I don't really want to start making this change without the intent to update the codebase (huge undertaking). It would be too confusing. Even after this PR, the initialization of this class would have a mix of initialization based on the args to the constructor. This is practical for this particular (the only) constructor, but doesn't help with any kind of code goodness goal.

Yeah, I agree with that. Thanks for pushing back on cosmetic changes 😄

mbr0wn:
> Well, I will make a new PR then. Feel free to reject this one

@gerstrong Many thanks for your contribution! I'll close this PR to keep the PR queue tidy.

PR #7431, opened 2024-07-14
git hook: add pre-commit and autopep8 git hook

ZakyHermawan:
## Description

Automatically run autopep8 in place, recursively each time you commit based on rules from [autoformat_python.sh](https://github.com/gnuradio/gnuradio/blob/main/tools/autoformat_python.sh),
if there are some errors, your commit will get rejected (no commit) and your code will be modified in place,
just run git add and git commit again.
This pre-commit hook will only modify the `.py` files that you have changed.

Note: To skip commit hooks during a commit, you can use the `--no-verify` option with git commit.
image

## How to use:
1. Install pre-commit package manager with `pip install pre-commit`
2. run `pre-commit install` to set up the git hook scripts
3. Make changes
4. `git add`
5. `git commit`
6. If your commit fails, simply run steps 4 and 5 again

Note: If you don't follow either step 1 or step 2, then the workflow will be same as usual (no git hooks will be run).
image

## Why do we need this ?
One of the problems with GNU Radio codebase is that there is autoformat
script that is needed to run manually to pass format test in git action,
so I think it will be a good idea to add pre-commit git hook to handle the
formatting before commit.

Every `.py` files

## Testing Done

Make a commit that violating our code convention in `.py` files, also make commit to modify non `.py` file

Change non `.py` file
image

Change `.py` file while violating code convention:
image

## Checklist

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,14 @@
+repos:
+-   repo: https://github.com/hhatto/autopep8
+    rev: v2.3.1
+    hooks:
+    -   id: autopep8
+        args: [
+            "--in-place",
+            "--recursive",
+            "--select=E101,E11,E121,E122,E123,E124,E125,E126,E127,\
+            E128,E129,E131,E133,E20,E211,E22,E224,E225,E226,E227,E228,E231,E241,E242,E251,\
+            E252,E26,E265,E266,E27,E301,E302,E303,E304,E305,E306,E401,E402,E501,E502,E701,\
+            E70,E711,E712,E713,E714,E721,E722,E731,W291,W292,W293,W391,W503,W601,W602,W603,\
+            W604,W690"
+        ]


willcode:
I'm not going to be excited about anything that automatically changes code on my machine, even if it is useful.

marcusmueller:
@willcode  to be fair, you don't have to use the `pre-commit` program :)

willcode:
I guess I can't complain too much about code security if the user has to do a `pre-commit install` and it only affects the one repo. It could be left to the dev to decide whether it fits into their workflow.

Next thing: how do we keep the E list sync'd? There are already 2 (script and CI) and now we'll have the list in the hook in the code and the list for the installed hook.

marcusmueller:
yeah the disparity between these two worries me (in CI we *exclude things*, here we *include*, which *does* seem more sensible if we want to normalize). Ideally, we have a script that extracts one from the other (probably pre-commit on install?). I don't know whether that is worth it, or whether we add a comment `If you change this, also change…` to the CI.

ZakyHermawan:
For E list problem, I'm thinking about removing (or deprecating) python autoformatter in tools directory, since it's more convenient of using pre-commit, and you can also do formatting for specific file only (for example: `pre-commit run --files filename.py`)

PR #7434, opened 2024-07-15
grc-qt: Entries in ErrorsDialog are editable

dl1ksv:
## Description

As the ErrorsDialog should only display the errors of the fg, the entries should be readonly and not editable.

## Testing Done

Setup a fg that contains errors. Click on the Error Display Box and there click on an item.
You can edit this selected item and edit it. This has no further effect, but it makes no sense to edit these error messags.

So this pr sets the the whole treeview to readonly. 

## Checklist


willcode:
Can the error dialogs get lost on the screen if they're not modal?

dl1ksv:
If you want a non modal dialog, you have to do much more coding.
If you call err.setModal(False) and err.show() the err window is not visible.

willcode:
Never mind - I read `model` as `modal` in the diff :smile:

PR #7436, opened 2024-07-16
grc: Don't overwrite port domain on auto-type deduction

mbr0wn:
Fixes the following issue: Create a port with automatic type deduction which is *not* in the streaming domain (e.g., create an RFNoC RX streamer block with no changes to settings from the default).

GRC will attempt to guess the dtype of this port (e.g., if it is connected to another block that does sc16, it will now also be sc16). However, due to the order in which objects are evaluated, the first time this is attempted is before the connection objects are generated, which means the auto-type deduction will fail the first time it is attempted.

The issue is, on failure, we re-set the port domain to 'stream' unconditionally. If the port domain was anything else (e.g., as above, 'rfnoc') then it will temporarily disable any domain-specific settings until the connection object is created, which is too late to import connection-and-domain-specific settings.

With this change, we simply don't touch the domain when the auto-type deduction fails.

## Description

See above. However, I'm not sure that the original code wasn't catching any corner case.

## Related Issue

This is easily reproduced by creating an RFNoC flowgraph in GRC with an RX streamer object, e.g., 

Radio -> RX streamer -> Null sink.

The connection from the Radio to the RX streamer will have no "Properties" box. There's a workaround: You can manually fix the type of the RX streamer input to 'sc16', and then there's no problem (after you reload the GRC). With this fix, no workaround is required.

Currently this is most easily tested with RFNoC blocks.

## Testing Done

I used my test case from above to test. However, I don't know which corner cases should be tested to make sure this didn't break other stuff down the line.

## Checklist

### DIFF
--- a/grc/core/ports/port.py
+++ b/grc/core/ports/port.py
@@ -163,7 +163,7 @@ class Port(Element):
             self.set_evaluated('vlen', port.vlen)
             self.domain = port.domain
         except AttributeError:
-            self.domain = Constants.DEFAULT_DOMAIN
+            pass
 
     def add_clone(self):
         """


dl1ksv:
Do we have any flowgraph example without an RFNoC block but with auto port deduction ?

mbr0wn:
> Do we have any flowgraph example without an RFNoC block but with auto port deduction ?

I think this feature might be used by virtual sinks/sources, but I'm not sure.

dl1ksv:
@mbr0wn Thank you for your hint. I found some example flowgraphs that uses virtual sinks/sources

Some examples run without or with this patch. Others fail, but these failures are not provoked by this patch.
So in my opinion this pr could be applied.

willcode:
OK, let's leave it on main for a bit to see if anything turns up.

PR #7437, opened 2024-07-16
grc: Improve check for connection in properties dialog

mbr0wn:
Without this fix, GRC tries to populate a full block-property dialog when opening a property dialog on connections. This improves the checking for the right attributes to correctly identify connections, and removes the following spurious output when opening property dialogs on connections:
```
  File "..../grc/gui/PropsDialog.py", line 210, in update_gui
    self._update_docs_page()
  File "..../grc/gui/PropsDialog.py", line 219, in _update_docs_page
    in_tree = self._block.category and self._block.category[0] == "Core"
AttributeError: 'Connection' object has no attribute 'category'
```

## Description

Opening a dialog before the fix:

image

After the fix:

image

AFAICT, only edges with properties.

## Testing Done

See pics above.

## Checklist

### DIFF
--- a/grc/core/Connection.py
+++ b/grc/core/Connection.py
@@ -23,7 +23,9 @@ class Connection(Element):
     """
 
     is_connection = True
+    category = []
     documentation = {'': ''}
+    doc_url = ''
 
     def __init__(self, parent, source, sink):
         """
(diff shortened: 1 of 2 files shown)


willcode:
Is it possible to initialize those fields instead of doing an existance check?

mbr0wn:
> Is it possible to initialize those fields instead of doing an existance check? Not a big deal. I see we use getattr elsewhere. That seems like something to do if the field name has to be constructed, but for fixed field names if feels more hackish.
> 
> This seems like a good one to get in before the final release.

Yes! That's a much better suggestion. I'll update the PR.

mbr0wn:
I updated the PR to pre-populate the fields with the same defaults as `Block`. The diff also seems much nicer now.

Issue #6989, opened 2023-12-05
gr-iio assumes underlying sample size


jamesoncollins:
https://github.com/gnuradio/gnuradio/blob/4efa6fbb4a3cddf3ad9f983f4730f8116a5df985/gr-iio/lib/device_source_impl.cc#L314

In the line above the underlying sample size of a libiio sample is assumed to be 16-bits to match the output port of the gnuradio block.  But in reality this sample might be 32-bits (and might also be 8-bits, not sure).

Because of the way iio works, especially with network targets, there is no good way to know, programatically, ahead of time what the sample size is.  It should either be left up to the user to pick the appropriate size, and throw a runtime error if they're wrong OR there should be an option to convert all samples to float regardless of the underlying type.

dkozel:
@jamesoncollins Thanks for the report, I think this is a quite reasonable request. Can you give an example of an IIO device which outputs a different item format? It's useful to have a specific example as motivation for the change.

willcode:
I'd vote for letting the user pick the type. Then at runtime either (1) the hardware produces that type, (2) the iio library or gr-iio block converts to the requested type, or (3) a runtime error says that conversion is not possible.

GR flowgraphs can be created on a machine with no hardware, so build-time hardware interrogation is not possible.

For the future, some sort of self-contained client, with controls like in Gqrx, would be nice in GR. It could have its own config file, discover and interact with harware, etc.

jamesoncollins:
How about if the requested size is an integer-type and is larger than the actual type then it transparently gets sign extended and converted, but if its smaller then it's a runtime error.  And if the output is float then its always a conversion without a runtime error?  There should probably be support for complex values in the future as well.

I don't know anything about gqrx so I can't really comment on that.

hennichodernich:
I second this. I had to write an 8-bit version for my selfbuilt hardware:
https://github.com/hennichodernich/gr-iio/blob/master/lib/eightbit_source_impl.cc
Currently I'm struggling to write a version for 24-bit samples padded to 32 bits.

Zach1031:
Hi,

I made a patch that gives the user two options for device_source and device_sink: conversion to float or letting the user pick a sample size. If the user picks and their choice doesn't align with the underlying it's a runtime error. Float conversion is automatic and doesn't rely on the underlying size. The device blocks support 1, 2, and 4 byte sample sizes.
https://github.com/Zach1031/gnuradio

willcode:
[...]

PR #7442, opened 2024-07-23
iio: Port libiio v1 files

Andrei-Fabian-Pop:
## Description

Libiio had some changes to the API with the transition to libiio v1 and it requires modifications to the source files. This PR modifies the gr-iio module to detect the libiio version in cmake and load the source files specific to that version. It also adds the source files related to libiio v1, these have the _v1 suffix, while the v0.X ones have the _v0 suffix.

## Related Issue

None.

The impacted block/area is `gr-iio`.

## Testing Done

The testing was performed using a PlutoSDR and 2 gnuradio flows, one for generating data and one for plotting that data.

## Checklist


willcode:
CMake on Window Conda does not have STRLESS

https://github.com/gnuradio/gnuradio/actions/runs/10058581509/job/27803026550?pr=7442#step:6:1668

pcercuei:
> CMake on Window Conda does not have STRLESS
> 
> https://github.com/gnuradio/gnuradio/actions/runs/10058581509/job/27803026550?pr=7442#step:6:1668

It probably have `VERSION_LESS`, @Andrei-Fabian-Pop maybe try with that?

willcode:
Hmm, it doesn't seem to like that either:
https://github.com/gnuradio/gnuradio/actions/runs/10059658360/job/27807704441?pr=7442#step:6:1734

marcusmueller:
> Our CI will only test v0 since none of the releases have the newer IIO. Any ideas how to get the v1 code path tested in CI?

Hm, I don't know how to "correctly" build and install libiio v1 yet, so maybe @Andrei-Fabian-Pop could make a PR against https://github.com/gnuradio/gnuradio-docker/ first, where he copies `ci/ci-ubuntu-24.04-3.10` (or the fedora 40 one? or the debian 12 one? Whatever is easiest!) to a new directory, removes any installation of libiio from the Dockerfile, and adds a `RUN git clone; mkdir build; cmake; make; make install; rm -rf source_code_and_build` step?

(or if you have have nightly libiio-dev distro packages falling out of your CI, that would of course be even easier)

Then, as soon as we merge that, the new image becomes available for us to build against. Then, this PR here would also need to add another builder to `.github/workflows/make-test.yml` that uses that docker image.

gr-iio/lib/attr_sink_impl_v1.cc
```diff
@@ -0,0 +1,159 @@
...
+
+    keys = pmt::dict_keys(pdu);
+
+    for (k = 0; k < pmt::length(keys); k++) {
+
+        attribute = pmt::symbol_to_string(pmt::nth(k, keys));
+        value =
+            pmt::symbol_to_string(pmt::dict_ref(pdu, pmt::nth(k, keys), pmt::PMT_NIL));
+
+        const struct iio_attr* iio_attribute = nullptr;
```
marcusmueller, line 94: Huh? This is C++, not C; is the `struct` here necessary?

gr-iio/lib/attr_source_impl_v1.cc
```diff
@@ -0,0 +1,317 @@
...
+            }
+        }
+    }
+
+    // Tell runtime system how many output items we produced.
+    return samples_per_update;
+}
+
+} /* namespace iio */
+} /* namespace gr */
```
marcusmueller, line 317: same as for attr_source (i.e., see above).

gr-iio/lib/dds_control_impl_v1.cc
```diff
@@ -0,0 +1,172 @@
...
+    }
+}
+
+/*
+ * Our virtual destructor.
+ */
+dds_control_impl::~dds_control_impl() {}
+
+} /* namespace iio */
+} /* namespace gr */
```
marcusmueller, line 172: This _v1 file is 100% identical to _v0. Undo the split.

gr-iio/lib/device_sink_impl_v1.cc
marcusmueller, line 1: There's very well-contained changes in the constructor, destructor, `channel_write` and `work` function. Please `#if` them instead of copying the whole file. Also, really, use the version number from the define in the IIO header file, detecting whether we're _V1 in Cmake doesn't scale as soon as there's more than two versions to support.

gr-iio/lib/device_source_impl_v1.cc
marcusmueller, line 1: This is the file with the most complex changes between _v0 and _v1. Still, all these changes are so well-localized that we should be using `#if`/`#else` around these to avoid duplicating source code.

gr-iio/lib/pluto_utils_v0.cc
```diff
@@ -9,10 +9,14 @@
 
 #include <gnuradio/iio/pluto_utils.h>
 #include <gnuradio/logger.h>
+#include <stdexcept>
 
+#ifdef LIBIIO_V1
+#include <iio/iio.h>
+#else
 #include <iio.h>
+#endif
```
marcusmueller, line 18: No sense in doing that here if you're only using that in the non-`LIBIIO_V1` case.

gr-iio/lib/pluto_utils_v1.cc
marcusmueller, line 1: This file is actually a full rewrite! It might make sense to keep it separate. But, if we merge all the other files, doing that might actually be more work than not.

marcusmueller (changes requested):
As you might imagine, I'm super thankful libiio finally sees a release 1.0!

But as you'll notice below, I have a strong opinion on having to maintain code copies that didn't have to be copied; the changes in all but one of these files are so small and self-contained that it'd reduce maintenance effort (and bug probabilities) if we only encapsulate these differing sections in preprocessor conditionals rather than having near-verbatim copies of source files in tree.

I'd also prefer if the code used the version number as defined in the iio headers instead of our detection from cmake: that makes things more robust, and the people that link against gr-iio won't have the GNU Radio CMake run, they will just have the GNU Radio, gr-iio and libiio headers, so they would need to get things from the libiio headers, anyways.

gr-iio/lib/CMakeLists.txt
```diff
@@ -9,15 +9,27 @@
...
-add_library(
-    gnuradio-iio
-    attr_sink_impl.cc
-    attr_source_impl.cc
-    attr_updater_impl.cc
-    dds_control_impl.cc
-    device_sink_impl.cc
-    device_source_impl.cc
-    iio_types.cc)
+if(DEFINED LIBIIO_V1 AND NOT DEFINED CACHE{LIBIIO_V1})
```
marcusmueller, line 12: what does that do?

gr-iio/lib/pluto_utils_v1.cc
marcusmueller: If I track this correctly, if we *don't* do the splitting for `pluto_utils`, and use the version number from the libiio header, we can leave the two CMakeLists.txt completely untouched.

gr-iio/lib/attr_sink_impl_v1.cc
```diff
@@ -0,0 +1,159 @@
...
+                    channel.c_str(),
+                    attribute.c_str());
+            throw std::runtime_error(error);
+        }
+    }
+}
+
+
+} /* namespace iio */
+} /* namespace gr */
```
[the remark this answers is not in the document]
pcercuei: When ADI releases Libiio v1.1, the current code will still work, because it won't break ABI/API :)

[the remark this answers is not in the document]
marcusmueller: @pcercuei  just to illustrate why I think this is not just a personal taste thing:

https://github.com/gnuradio/gnuradio/pulls?q=is%3Apr+is%3Aclosed+iio+NOT+backport

is the list of things we changed. Some of these changes have been coming from within ADI, and we're very thankful for that, but most of them haven't. And no matter where the changes come from, the effort of reviewing them is on the maintainer.

So, when now someone notices a bug in gr-iio, and wants to write a fix after you copy all the _impl.cc, they get to do it in two files, one of which they cannot even build, even if it's not related to the differences between libiio versions. We get to review both changes – and then someone gets to decide what to backport to release branches. (Currently, that's @willcode).

I think this just doesn't scale right now, and it won't scale when we don't have a @willcode  around to diligently go through two files to compare the changes.

willcode:
Agree ... conditionals or some kind of refactoring is far better than two nearly identical classes.

gr-iio/lib/attr_sink_impl_v1.cc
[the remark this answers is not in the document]
pcercuei: Oh, no need to argue with me, I agree with you fully.

gr-iio/lib/CMakeLists.txt
Andrei-Fabian-Pop: This helped me when I was playing around with both libiio versions. I didn't always make a clean build and I added this so that gnuradio does not accidentally link to a wrong verion (maybe one what I removed from the system). This does nothing unexpected when making a clean build. If you think this is not necessay or it does not fit in with the rest of the CMakeLists from gnuradio I can take it out and leave only the `if(DEFINED LIBIIO_V1)`

Andrei-Fabian-Pop:
Hello and thank you @marcusmueller and @willcode for all the comments on this PR. I merged back the source files and used `#ifdef` for differentiating between the versions. This way, the only change in the `CMakeLists.txt` is the check for the libiio version and setting the macro for the source code. Regarding the new commit with changes, I will squash it later when most of the changes are implemented :)

willcode:
This at least gets everything into one set of files. I suspect that maintenance would be even easier if some of the legacy functions were reimplemented as wrappers around the newer functions. For example, `iio_channel_attr_read_double` could be a wrapper around `iio_channel_find_attr`. If this could be made to work, then many of the `#ifdef`s would be unneeded.

Andrei-Fabian-Pop:
> This at least gets everything into one set of files. I suspect that maintenance would be even easier if some of the legacy functions were reimplemented as wrappers around the newer functions. For example, `iio_channel_attr_read_double` could be a wrapper around `iio_channel_find_attr`. If this could be made to work, then many of the `#ifdef`s would be unneeded.

Won't the code be less readable this way? For example in `attr_cource_impl.cc` there are switches that could go in a wrapper, but not all of them. Only 2 of the branches per switch would use the `iio_channel_attr_read_double` wrapper, but the next branch has the `iio_device_debug_attr_read_double` function call. I think that this might make the code a bit harder to follow without clear distinctions between versions, but if that is the preferred implementation, then I will add it.

willcode:
I think the single set of files was the main thing. The rest was a suggestion for the people who will be maintaining contents of these files (not me!).

pcercuei:
> This at least gets everything into one set of files. I suspect that maintenance would be even easier if some of the legacy functions were reimplemented as wrappers around the newer functions. For example, `iio_channel_attr_read_double` could be a wrapper around `iio_channel_find_attr`. If this could be made to work, then many of the `#ifdef`s would be unneeded.

Note that Libiio 1.0 already provides an optional and fully functional compatibility layer: https://github.com/analogdevicesinc/libiio/blob/main/compat.c

I guess in the long term it would be beneficial to switch GNU Radio to the Libiio v1.0 API, but in the meantime GNU Radio should work just fine with the compat layer.

Also, while I applaud the work done to port the gr-iio blocks to Libiio v1.0, I do feel like things should be done in order - and the first thing would be to actually have a Libiio v1.0. While the API is pretty much final now, it has yet to be released.

willcode:
I'm going to make this Draft, based on the comments above. We can revisit when the new API is out.

jamesoncollins:
@Zach1031 has this [fork ](https://github.com/Zach1031/gnuradio) to deal with the sample size issues associated with gr-iio according to #6989 .

I'd like to try and get that fork merged in soon, but wanted to make sure you were aware of the effort.

PR #7447, opened 2024-07-26
Donation link and badge within main README

777arc:


willcode:
@mormj Just since this has a link to somewhere people will donate money, can I have a presidential double check :smile:

mormj:
This looks good to me!

PR #7452, opened 2024-07-28
tools: update clang_format script to search for clang-format 14

### LINKED ISSUE #7063: Clang Format helper script looks for the wrong version
### What happened?

Tried running https://github.com/gnuradio/gnuradio/blob/main/tools/clang_format.py and found it was looking for clang-format 7.0. Our CI uses version 14 so this is likely a simple fix to change the version it's looking for.

### System Information

OS: Ubuntu 22.04

### GNU Radio Version

3.10 (maint-3.10)

### Specific Version

0.0.0.0

Run ./tools/clang_format.py

### Relevant log output

_No response_

souponaroom:
## Description
The old clang_format.py script searches for clang-format 7, but the version that the github CI uses is 14

## Related Issue
Fixes #7063

tools

## Testing Done
Verified that error stops when ran on a docker container running the same ubuntu version as the github ubuntu CI (22.04)

## Checklist

- [ ] I have updated [the documentation](https://wiki.gnuradio.org/index.php/Main_Page#Documentation) where necessary. NA
- [ ] I have added tests to cover my changes, and all previous tests pass. NA

### DIFF
--- a/tools/clang_format.py
+++ b/tools/clang_format.py
@@ -49,8 +49,8 @@ if __name__ == "__main__" and __package__ is None:
 #
 
 # Expected version of clang-format
-CLANG_FORMAT_VERSION = "7.0.1"
-CLANG_FORMAT_SHORT_VERSION = "7.0"
+CLANG_FORMAT_VERSION = "14.0.0"
+CLANG_FORMAT_SHORT_VERSION = "14.0"
 
 # Name of clang-format as a binary
 CLANG_FORMAT_PROGNAME = "clang-format"


willcode:
This won't make things any works, but for instance Fedora 40 has `clang-format-17`. It seems like the specific version will hit only occasionally.

@marcusmueller do you think we should just get rid of the search for a specific version?

souponaroom:
What is the intended purpose behind the version matching? If it's to ensure that a certain set of features are available, then it could be refactored to be a minimum version requirement, or a fence between two versions (minimum/maximum).

marcusmueller:
@willcode There's sadly slight differences between clang-format versions in corner cases, so having one place where we say, "hey, this is the golden version" makes sense, and that place should match the CI. (It'd not be a big deal to update the version every mid-sized release, you just don't want to do it all the time, as, as small chance as that has of happening, might break the CI success of existing PRs, and might not be worth it).

willcode:
Alright - it probably won't affect most of the target platforms, but won't hurt anything.

PR #7457, opened 2024-08-03
runtime: Use ABC's in tag_utils

souponaroom:
## Description
Currently, tag_utils detects if an argument in python is an integer/list/whatever by checking its type. This can lead to hard to track bugs if e.g. the programmer attempts to use a numpy np.int_ or an ndarray. These types can be supported by having isinstance check for an abstract base class.

None

## Testing Done
This change is trivial so I have not added a unit test, but all old tests still pass. If I should create a unit test, let me know.
An older codebase I was reformatting was not working due to np.ndarray and numpy int usage, and now it works with this change, so the change seems to do exactly what it ought to do.

## Checklist

- [ X] I have read the [CONTRIBUTING document](https://github.com/gnuradio/gnuradio/blob/main/CONTRIBUTING.md).
- [X ] I have squashed my commits to have one significant change per commit. 
- [ X] I [have signed my commits before making this PR](https://github.com/gnuradio/gnuradio/blob/main/CONTRIBUTING.md#dco-signed)
- [ X] My code follows the code style of this project. See [GREP1.md](https://github.com/gnuradio/greps/blob/main/grep-0001-coding-guidelines.md).
- [ ] I have updated [the documentation](https://wiki.gnuradio.org/index.php/Main_Page#Documentation) where necessary. NA
- [ ] I have added tests to cover my changes, and all previous tests pass. NA

### DIFF
--- a/gnuradio-runtime/python/gnuradio/gr/tag_utils.py
+++ b/gnuradio-runtime/python/gnuradio/gr/tag_utils.py
@@ -1,3 +1,6 @@
+from collections.abc import Mapping, Collection
+from numbers import Integral
+
 import pmt
 
 from . import gr_python as gr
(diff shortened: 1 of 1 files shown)


souponaroom:
Looking back at my request, I have just realized that it's entirely possible for a collections.abc.Collection to not support __get_item__, which is required by the code.

The most broad abc that adequately defines what's required in line 72 is a Sequence, but that type is too restrictive, as ndarrays are not sequences.

I still think that detecting "close-ish" types for arguments should be supported, or at least not silently fail, but I'll need to check the Python features to see how to do so correctly.

PR #7464, opened 2024-08-11
github CI: Mingw64: don't install click-plugins; no longer used

marcusmueller:
Saves us the whole pip install step. Less complexity is nice.
## Description

We removed the click-plugins dependency a while back. (never got used)
The mingw64 builder (seems to only exist on main) still installs them; no need to, so let's remove that.

Hopefully, none, but complexity of our CI script

## Testing Done

n/a

## Checklist


willcode:
Backport along with #7226

Issue #7458, opened 2024-08-03
Numpy 2 is installed on mingwx64 CI runner, and qa_correlate_access_code_XX_ts is affected by NEP 50.


souponaroom:
### What happened?

When the mingwx64 CI environment runs pacboy, it installs numpy 2. This version of numpy no longer automatically promotes integer types, so qa_correlate_access_code_XX_ts fails because it attempts to multiply a np.uint8 by 256.

### System Information

CI Runner mingw64

### GNU Radio Version

3.11-git (main)

### Specific Version

0.0.0.0

### Steps to Reproduce the Problem

re-run the CI runner on main and confirm?

### Relevant log output

```shell
See #6745 actions page.
```

PR #7465, opened 2024-08-11
python: set numpy type propagation policy in gateway

marcusmueller:
## Description

This way, Python code using GNU Radio blocks has the same type propagation behaviour as numpy 2, no matter whether it's using numpy 1 (>=1.24.0) or 2.

Background is that numpy2 makes test cases fail that do something like

```python
a = numpy.ndarray([1,2], dtype=numpy.uint8)
b = a[0] * 256 + a[1]
```

because numpy2's propagation rules convert 256 to the dtype of `a[0]` first, and that fails, because 256 is too large to fit an 8 bit integer.

## Related Issue

#7458 highlights how numpy2 test cases differ from numpy1 test cases

Sadly, everything that imports from gnuradio in Python

## Testing Done

None, this should, if anything, make *more* testcases fail; I expect the same failure we see on the Mingw64 test (using numpy2) to happen everywhere that's got a recent enough python version (Deb12, Ubuntu24, Fedoras)

## Checklist

### DIFF
--- a/gnuradio-runtime/python/gnuradio/gr/gateway.py
+++ b/gnuradio-runtime/python/gnuradio/gr/gateway.py
@@ -1,5 +1,6 @@
 #
 # Copyright 2011-2012, 2018, 2020 Free Software Foundation, Inc.
+# Copyright 2024 Marcus Müller
 #
 # This file is part of GNU Radio
 #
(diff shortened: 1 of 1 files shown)


willcode:
Is this a change you actually want to make, and if so, why no backport?

marcusmueller:
@willcode Well, we need to make that change to make sure that our software and QA works with numpy 2 as it does currently with numpy 1; so yep, I want to do that.

I don't think this should be backported, because it changes the behaviour of numpy for the whole python application that  `from gnuradio import gr`, and that would constitute a breaking change, I'm afraid.

willcode:
I'm going to add the backport tag, just so we don't lose track of this, but leave the no backport tag as well.

marcusmueller:
Fixed by #7499

PR #7471, opened 2024-08-15
Option for creating an Issue that isnt a feature or bug

777arc:
@marcusmueller does this seem OK?  There are occasionally Issues that are not a bug or feature request

### DIFF
--- a/.github/ISSUE_TEMPLATE/03-Misc.yml
+++ b/.github/ISSUE_TEMPLATE/03-Misc.yml
@@ -1,0 +1,13 @@
+name: Other
+description: Questions, comments, or anything else
+labels: []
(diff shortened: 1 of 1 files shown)


marcusmueller:
Please don't. Let's keep the bug tracker for tracking bugs.

777arc:
I'm not removing the bug tracker

marcusmueller:
I wasn't expecting that you'd do, but I'd really think we'd be worse off if we also started doing user support via github issues. Very few, if any, of the "help me" issues reported before we added the "please don't ask about usage here" upfront were of a high quality. Most of them were answered with "please go and ask on the mailing list or in chat".

PR #7473, opened 2024-08-17
blocks: Fix displayed SigMF dataset format strings

argilo:
## Description
The SigMF dataset format strings shown in the SigMF Source & Sink (Minimal) blocks are incorrect. Here I've changed them in accordance with the specification (https://sigmf.org/#sigmf-dataset-format).

* SigMF Source (Minimal)
* SigMF Sink (Minimal)

## Testing Done
I verified that the correct names appear in the drop-down menus, and that existing flow graphs still function after the change.

## Checklist

### DIFF
--- a/gr-blocks/grc/blocks_sigmf_sink_minimal.block.yml
+++ b/gr-blocks/grc/blocks_sigmf_sink_minimal.block.yml
@@ -10,6 +10,7 @@ parameters:
     label: Stream Type
     dtype: enum
     options: [complex float (fc32_le), real float (rc32_le), complex short (sc16_le), real short (rc16_le)]
+    option_labels: [complex float (cf32_le), real float (rf32_le), complex short (ci16_le), real short (ri16_le)]
     option_attributes:
         type: [complex, float, short, short]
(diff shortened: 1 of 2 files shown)


jacobagilbert (approved):
Good catches, thanks!

argilo:
I noticed one unfortunate side effect. Since this changes the type values that are stored in GRC files, existing flow graphs are broken and fall back to the default type (cf32_le).

argilo:
To avoid breaking existing flow graphs, we could instead add `option_labels` to change how the options are presented to the user without changing the GRC format.

argilo:
I've updated this to add `option_labels` instead of changing the `options` list, which avoids breaking existing flow graphs.

777arc:
Good thinking, thanks for fixing it!!

777arc:
At least this part was correct https://github.com/gnuradio/gnuradio/blob/main/gr-blocks/python/blocks/sigmf_sink_minimal.py#L38

777arc (approved):

argilo:
> At least this part was correct

Yeah, I didn't find the incorrect strings anywhere other than the `options` lists.

PR #7479, opened 2024-08-25
grc: add missing developer name tag to metainfo file

AsciiWolf:
## Description
This PR adds the previously missing developer name tag to the GRC AppStream metainfo file.

It replaces the original #7478 PR that also added OARS metadata that should not be added.

It also fixes a small typo in the comment section.

Please note that the current developer name content is just a placeholder for the actual (correct) developer name. If it is incorrect, please let me know and I will change it to the correct one provided by you.

/cc @marcusmueller

None.

## Checklist

### DIFF
--- a/grc/scripts/freedesktop/org.gnuradio.grc.metainfo.xml
+++ b/grc/scripts/freedesktop/org.gnuradio.grc.metainfo.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!--
-    This file was tested against `appstream cli validate` v0.12.7 (Fedora 31)
+    This file was tested against `appstreamcli validate` v0.12.7 (Fedora 31)
     and v0.12.5 (Debian 10) It fails to validate under v0.12.0 (Ubuntu 18.04).
     So, this file becomes valid somewhere between v0.12.1 and v0.12.5.
     Note that metainfo doesn't mandate (even support at this point) schema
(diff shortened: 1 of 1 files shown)


777arc:
Probably better using "GNU Radio" for developer name, right?

marcusmueller:
`Radio project` in any way is definitely wrong. 

@777arc we could use `GNU Radio`, but at that point, let's simply not have the metadata entry? It's meaningless to say "GNU Radio was created by GNU Radio folks". I'd say using the flexibility of the metadata format to *not* give information that isn't relevant/established should be used, instead of inserting dummy values.

grc/scripts/freedesktop/org.gnuradio.grc.metainfo.xml
```diff
@@ -13,6 +13,7 @@
 <metadata_license>CC0-1.0</metadata_license>
 <project_license>GPL-3.0+</project_license>
 <name>GNU Radio Companion</name>
+<developer_name>Radio project</developer_name>
```
marcusmueller, line 16: Definitely wrong; see discussion

marcusmueller (changes requested):

marcusmueller:
Closing because not spending maintainer time on adding deprecated metadata.

AsciiWolf:
This tag is deprecated only in the latest AppStream 1.0 standard - and even there is supported and will be for a very long time because of backwards compatibility. I have intentionally not added the new _developer_ tag because AppStream 1.0 is still not used by most current distributions and parsers and that tag would be ignored by them. Why not ask first before closing the PR that I spent my time on?

I love GNU Radio and use it regularly as a ham radio operator and radio enthusiast - that's the main reason why I want to made it accessible for installation to regular users (not only Linux/*NIX geeks). But I really dislike such attitude.

Anyway, I accept your opinion - if you really do not have a name for your developer community. (I originally thought that it is GNU Project but it looks like it is not.)

willcode:
I think the important part is that we don't have an official developer name, so it's "filling in the blank" to come up with one.

PR #7485, opened 2024-08-27
grc: Fix issue #7484 (channelizer output count incorrect in some situations).

### LINKED ISSUE #7484: GRC: Hierarchical Polyphase Channelizer multiplicity (output port count) is wrong when output channel list is supplied
### What happened?

# Summary

The Hierarchical Polyphase Channelizer GRC component incorrectly produces more output ports on the component than required whenever the "Output Channels" map is non-empty. This makes the signal graph unusable at runtime, crashing in `gnuradio/gr/hier_block2.py` with a ValueError:

```
    self.primitive_connect(*args)
ValueError: port number 2 exceeds max of 1
```

# Background

The Hierarchical Polyphase Channelizer is a signal processing block which splits an incoming signal stream into discrete sub-channels of equal bandwidth. It can be configured to skip some of its channels on output via an output map parameter `outchans`. When an output map is provided, the number of output ports provided by the block is reduced at runtime.

# Fix

The number of output ports for a flexible block like the channelizer is controlled at runtime via the `outputs.multiplicity` item in the block configuration. The bug is due to an incorrect template expression in the YAML for this block (`filter_pfb_channelizer_hier.block.yml`).

The current expression reads:

```yaml
outputs:
-   domain: stream
    dtype: complex
    multiplicity: ${ nchans }
```

Whereas it should read:

```yaml
outputs:
-   domain: stream
    dtype: complex
    multiplicity: ${ nchans if outchans is None else len(outchans) }
```

Which will correctly derive the number of outputs whenever there is an output channel list.

### System Information

OS: Mac OS X
uname -a: Darwin machine.local 23.2.0 Darwin Kernel Version 23.2.0: Wed Nov 15 21:53:18 PST 2023; root:xnu-10002.61.3~2/RELEASE_ARM64_T6000 arm64

GR Installation Method: homebrew

### GNU Radio Version

3.10 (maint-3.10)

### Specific Version

3.10.9.2

# Graphical setup

graph
channelizer-setup

1. From a new GRC flowgraph:
2. Adjust the `samp_rate` variable to 1_000_000
3. Drop in a Noise Source block (Waveform Generators -> Noise Source)
4. Drop in a Hierarchical Polyphase Channelizer block (Channelizers -> Hierarchical Polyphase Channelizer)
5. Drop in three Null Sink blocks (Misc -> Null Sink)
6. Connect the noise source to the channelizer.
7. Configure the channelizer. The default number of channels should be three. Just add an output channel map of only two elements. (Number of Channels: 3, **Output Channels: [0,1]**)
8. **Notice**: The number of output ports remains at three, even though there should now only be two.
9. Connect all three output ports to its own Null Sink.
10. Run the flowgraph.
11. Observe the crash.

### Relevant log output

```shell
Executing: /opt/homebrew/Cellar/gnuradio/3.10.9.2_8/libexec/venv/bin/python -u /Users/bug/Desktop/bug.py

log :warning: Matplotlib is a required dependency to use DistanceRadar and AzElPlot.  Please install matplotlib to use these blocks (https://matplotlib.org/)
  File "/Users/bug/Desktop/bug.py", line 136, in <module>
    main()
  File "/Users/bug/Desktop/bug.py", line 114, in main
    tb = top_block_cls()
         ^^^^^^^^^^^^^^^
  File "/Users/bug/Desktop/bug.py", line 90, in __init__
    self.connect((self.pfb_channelizer_hier_ccf_0, 2), (self.blocks_null_sink_0_1, 0))
  File "/opt/homebrew/Cellar/gnuradio/3.10.9.2_8/lib/python3.12/site-packages/gnuradio/gr/hier_block2.py", line 40, in wrapped
    func(self, src, src_port, dst, dst_port)
  File "/opt/homebrew/Cellar/gnuradio/3.10.9.2_8/lib/python3.12/site-packages/gnuradio/gr/hier_block2.py", line 126, in connect
    self.primitive_connect(*args)
ValueError: port number 2 exceeds max of 1

>>> Done (return code 1)
```

ke6jjj:
Fixes #7484 by checking both the number of channels being produced AND the output map (if provided) to determine the actual number of output ports that the block will produce at runtime.

## Description
Full description is in issue #7484

This issue (and patch) only affects GNURadio Companion.

## Testing Done
I used the example flow graph in issue #7484 to both reproduce the issue and to verify the fix.

## Checklist

### DIFF
--- a/gr-filter/grc/filter_pfb_channelizer_hier.block.yml
+++ b/gr-filter/grc/filter_pfb_channelizer_hier.block.yml
@@ -44,7 +44,7 @@ inputs:
 outputs:
 -   domain: stream
     dtype: complex
-    multiplicity: ${ nchans }
+    multiplicity: ${ nchans if outchans is None else len(outchans) }
 
 templates:
     imports: from gnuradio.filter import pfb


willcode:
Please sign the commit - this says that you accept the Developer Certificate of Origin.

ke6jjj:
I signed the commit. I even see the "Verified" mark on it. Is that not enough?

willcode:
For our purposes, the `Signed-off-by: Your Name <your@e.mail>` tag added by `git commit -s` implies the DCO. Github verification is nice, but the Git repo needs to be standalone outside Github. Take a look at the commit log so see examples. You can sign your existing commit using
`git commit --amend --signoff`
then do a force push.

willcode:
@marcusmueller could you sanity check this one - I missed the DCO being updated a while ago, it looks right.

PR #7486, opened 2024-08-28
video-sdl: Separate render thread

### LINKED ISSUE #7482: video-sdl: Doesn't work with sdl12-compat
### What happened?

The SDL2 output window is pure black and unresponsive with no error logs.

The cause is probably the change in `libsdl1.2debian` package in Ubuntu 24.04 LTS and Debian Testing (trixie), which now uses SDL2 in the background through `sdl12-compat`. The same issue is present in Arch Linux, because the `gnuradio` package depends on `sdl12-compat`.

Fully updating the module to SDL2 will solve this.

### System Information

OS: Ubuntu 24.04 LTS
GR Installation Method: Source, apt

### GNU Radio Version

3.10 (maint-3.10)

### Specific Version

3.10.9.2

1. Install GNU Radio on Ubuntu 24.04 LTS
2. Verify that `apt show libsdl1.2debian` contains "_This package contains sdl12-compat, a binary compatible API for programs built SDL 1.2, but using SDL 2.0_" in the description
3. Run demo flowgraph: https://wiki.gnuradio.org/images/e/ed/Demonstrate_sdl.grc

### Relevant log output

_No response_

AsriFox:
## Description
SDL2 expects that all SDL render calls originate from a single thread [[1]](https://wiki.libsdl.org/SDL2/FAQDevelopment#Can_I_call_SDL_video_functions_from_multiple_threads). However, a block's constructor and `work` function can be (and most of the time are) called from different threads.

Because SDL v1.2 is being replaced with a compatibility library *sdl12-compat* ([Ubuntu 24.04 LTS](https://launchpad.net/ubuntu/noble/+package/libsdl1.2-dev), [Debian Testing (trixie)](https://packages.debian.org/trixie/libsdl1.2-dev), [Arch Linux](https://archlinux.org/packages/extra/x86_64/sdl12-compat)), this presents an issue for newer systems.

Moving all SDL function calls to a single separate thread solves this issue, allowing *gr-video-sdl* to work with *sdl12-compat*.

## Related Issue
Fixes #7482

gr-video-sdl

## Testing Done
https://wiki.gnuradio.org/images/e/ed/Demonstrate_sdl.grc

## Checklist


gr-video-sdl/lib/sink_uc_impl.cc
```diff
@@ -226,67 +119,142 @@ int sink_uc_impl::work(int noutput_items,
...
+            memcpy(&overlay->pixels[2][i * overlay->pitches[2]],
+                   &sink->d_buf_v[i * sink->d_width / 2],
+                   sink->d_width / 2);
+        }
+        SDL_UnlockYUVOverlay(overlay);
+        SDL_DisplayYUVOverlay(overlay, &dstrect);
+    }
+
+    SDL_FreeYUVOverlay(overlay);
+    exit(0);
```
willcode, line 255: Can't exit() in a block.

gr-video-sdl/lib/sink_s_impl.cc
```diff
@@ -232,66 +120,153 @@ int sink_s_impl::work(int noutput_items,
...
+            memcpy(&overlay->pixels[2][i * overlay->pitches[2]],
+                   &sink->d_buf_v[i * sink->d_width / 2],
+                   sink->d_width / 2);
+        }
+        SDL_UnlockYUVOverlay(overlay);
+        SDL_DisplayYUVOverlay(overlay, &dstrect);
+    }
+
+    SDL_FreeYUVOverlay(overlay);
+    exit(0);
```
willcode, line 267: Can't exit() in a block.

AsriFox:
There isn't any synchronization, tearing is not avoided. I'll work on that.

gr-video-sdl/lib/sink_uc_impl.cc
AsriFox: Is there a way to broadcast a "stop" signal not just to the block, but to the whole flowgraph?

willcode: A display wouldn't typically want to terminate the flowgraph. If a work() function returns DONE, then the framework does not expect it to take any more data, and that would result in the flowgraph stopping (eventually?).

AsriFox:
No more tearing, but the image flickers on low framerates. Not sure why.

marcusmueller (approved):
I could not test locally on short notice, but codewise 100% an improvement

AsriFox:
The two threads *should* not access the same buffer, this is ensured by the `d_frame_pending` flag: when it's `false`, the render thread doesn't render; when it's `true`, the main thread skips writing into the overlay.

`SDL_Quit()` is required to destroy the SDL window (at least in SDL 1.2), so it should be executed once the block is destroyed. It also sends the event for `SDL_QuitRequested()` from the block destructor to catch in the render thread.

drmpeg:
Just FYI, i have a full motion test suite here.

https://www.w6rz.net/sdltest2.tar.gz

1,010,079,355 bytes

AsriFox:
@drmpeg Thank you! This test revealed two bugs in my code:
* [x] Color planes have full width and height on input, but must have only half width and height on display due to decimation.
* [ ] If the input sample rate is higher than the desired sample rate (desired framerate * width * height), the display lags increasingly hard. No such issue if the input sample rate is lower than the desired, including the special value "0" (to produce frames as fast as possible).

drmpeg:
That test sequence is actually 60 frames/sec (900 frames = 15 seconds). I'm not sure why I had 30 fps in the test flow graph (probably because the computer I was using at the time couldn't keep up). Here's the source material for comparison.

https://www.w6rz.net/rickmorty4_7.mp4

679,924,385 bytes

drmpeg:
Looks good. However, on a fast CPU the frame rate parameter does not limit the frame rate. The frames are played as fast as possible no matter what the setting of the frame rate is (only limited by the throttle block).

AsriFox:
> Looks good. However, on a fast CPU the frame rate parameter does not limit the frame rate. The frames are played as fast as possible no matter what the setting of the frame rate is (only limited by the throttle block).

Can't reproduce. The frames are displayed with `d_avg_delay`; it's not directly connected to the target framerate, but it's close. However, the separation of threads means that extra incoming frames are simply skipped. Perhaps, SDL sinks should include throttle functionality?

drmpeg:
> Perhaps, SDL sinks should include throttle functionality?

That's the old behavior. If you set the frame rate to 1 fps, it plays all the frames for 1 second each. No skipping.

AsriFox:
On the wiki, framerate is described as follows:
> Number of updates per second. If set to 0, make a new screen every (Input Width) · (Input Height) samples. (Recommended: set to 0, use modern [Throttle](https://wiki.gnuradio.org/index.php?title=Throttle) with a maximum wait time low enough and as sample rate (Input Width) · (Input Height) · (desired frames per second))

So I don't think input rate should be limited by output (display) rate. New behavior allows you to control them independent of each other; to limit the actual input rate, Throttle should be used.

micsthepick:
the above failing build (on windows at least) will be fixed by merging with https://github.com/gnuradio/gnuradio/commit/6fbc354978bb24971c4548f70330ce0e27c4d43d or later

micsthepick:
@AsriFox does this work with the same behavior as before when input width != display width, and same for height?

micsthepick:
Screenshot 2024-10-31 195311
see green bar on the right - it's in greyscale mode... not sure if this is my fault (I partially built just the video sdl module and it's dependencies, and copied over what was needed to make it run on an existing grc install from radioconda on windows)

micsthepick:
[Demonstrate_sdl.zip](https://github.com/user-attachments/files/17585458/Demonstrate_sdl.zip)
The above MVCE replicates it for me, if it doesn't for you, then I'd not worry about that issue

AsriFox:
@micsthepick I don't have a Windows build right now, but on Linux difference in frame sizes does not produce this issue. The input frame is simply stretched to fit the display size.
However, green pixels did appear when I set the input width or height to odd values (31, 511, 2047). The bars were only 1px, but it does suggest that there's a mismatch between the texture size and the size of data written into it - probably a limitation of IYUV color planes, which is out of scope for this PR.

AsriFox:
It seems that using odd width/height with YUV pixel formats is simply not supported in SDL 1.2: libsdl-org/SDL-1.2#445. Existing behavior looks like [this](https://trac.ffmpeg.org/attachment/ticket/1322/ffplay.png) - black pixels and misaligned frame. The [workaround](https://trac.ffmpeg.org/ticket/1322) in old ffplay was to crop the input frame to even size.

The reason it's different with this PR is the use of intermediate buffers between input data and display pixels. Some pixels on U/V planes are not filled, thus they're green. Even `memset(d_image->pixels[1], 128, d_image->pitches[1] * (d_height / 2));` is not enough to fill the overlay. I've not noticed missing pixels in Y plane, so this might only be present in color space. Perhaps, the way to resolve it is to make the YUV overlay even-sized, either cropping 1px or adding 1px black bars.

PR #7497, opened 2024-08-30
blocks: Fix EOF detection in Wav File Source

### LINKED ISSUE #7496: Wav File Source's "Repeat" property does not work for MP3
### What happened?

If the Wav File Source is used to play an MP3 file, and the block's "Repeat" property is set to "Yes", the file does not actually repeat. I did some debugging, and it looks like the problem is that the `SF_INFO.frames` value returned by `sf_open` does not match the actual length of the file, and so the `if (d_sample_idx >= d_h.samples_per_chan)` check never passes. In my test, the MP3 file was 18312192 samples long, but `SFINFO.frames` (and therefore also d_h.samples_per_chan) was 18335223. The following upstream issue describes the problem: https://github.com/libsndfile/libsndfile/issues/1001

One possible workaround would be to detect the end of the input file by checking the return value of `sf_read_float`.

### System Information

OS: Ubuntu 24.04
GR Installation Method: PyBOMBS

### GNU Radio Version

3.10 (maint-3.10)

### Specific Version

3.10.11.0

Read an MP3 file using a Wav File Source block, and set "Repeat" to "Yes". It may be necessary to try multiple MP3 files, since the `SFINFO.frames` appears to be an estimate, and it may not always be larger than the actual file length.

### Relevant log output

_No response_

argilo:
## Description
When reading certain MP3 files (specifically, those that lack an "Info" frame), the Wav File Source block hangs when reaching the end of the input file. This happens because libsndfile returns an approximate length for the input file (because it would be expensive to determine the exact number of samples), while the Wav File Source expects the returned value to be exact.

To solve the problem, I have modified the end-of-file detection to check whether the number of returned samples was smaller than the amount requested. The [libsndfile documentation](https://libsndfile.github.io/libsndfile/api.html#read) indicates that this is a valid way to detect end-of-file:

> The sf_read_XXXX and sf_readf_XXXX functions return the number of items or frames read, respectively. **Unless the end of the file was reached during the read, the return value should equal the number of objects requested. Attempts to read beyond the end of the file will not result in an error but will cause the read functions to return less than the number of objects requested or 0 if already at the end of the file.** When the buffer is not is not completely filled, unused buffer space is filled by zeroes.

## Related Issue
Fixes #7496.

* Wav File Source

## Testing Done
I added two test cases that read an MP3 file, and verified that they fail with the current code, and pass with the updated code. I also made a test flowgraph and verified that the Wav File Source block can read an MP3 file correctly.

## Checklist


argilo:
Oops, the new tests fail on older OSes because MP3 support was only added to libsndfile in version 1.1.0 (released 2022-03-27).

argilo:
I modified the new tests so that they are skipped if MP3 support is missing. But they still fail on Fedora 39 for some other reason. I'll have to investigate that.

argilo:
I suspect it's due to this bug, which was fixed in libsndfile 1.2.1:

https://github.com/libsndfile/libsndfile/pull/943

Fedora 39 has libsndfile 1.1.0.

argilo:
I added unconditional skips for the MP3 tests. We can re-enabled them once we reach the point where all supported OSes have libsndfile >= 1.2.1.

Since the Repeat parameter would be untested until then, I've added in corresponding tests that use WAV files.

argilo:
I just noticed a bug: The old `if (d_sample_idx >= d_h.samples_per_chan)` condition had the side effect of ensuring that `sf_read_float` would never return zero samples, and thus `items` could only be zero on the first iteration of the loop, which ensured that `-1` would only be returned if no samples had been produced.

But that's no longer the case, and `sf_read_float` can now return zero samples at any time. Thus the return value in the `if (!d_repeat)` case needs to be adjusted so that `-1` is returned only when zero samples have been produced. Without this change, files with a multiple of 1024 samples would be truncated. I've added another test case to verify that files of these lengths are handled correctly.

drmpeg (approved):
Looks good. I tested with various file formats and repeat works fine. I also have an MP2 file that is from a broadcast Transport Stream (so no "Info" frame). This file fails with the old code and works with the new.

https://www.w6rz.net/jazzclip.mp2

argilo:
Oops, it looks like one of the new tests isn't playing nicely with MinGW-w64, which causes CI to fail:

```
ERROR: test_read_wav_1024 (__main__.test_wavefile.test_read_wav_1024)
  File "D:/a/gnuradio/gnuradio/gr-blocks/python/blocks/qa_wavfile.py", line 194, in test_read_wav_1024
    with wave.open(tf.name, "w") as wf:
         ^^^^^^^^^^^^^^^^^^^^^^^
  File "D:/a/_temp/msys64/mingw64/lib/python3.11/wave.py", line 633, in open
    return Wave_write(f)
           ^^^^^^^^^^^^^
  File "D:/a/_temp/msys64/mingw64/lib/python3.11/wave.py", line 425, in __init__
    f = builtins.open(f, 'wb')
        ^^^^^^^^^^^^^^^^^^^^^^
PermissionError: [Errno 13] Permission denied: 'D:/a/_temp/msys64/tmp/tmpryg8mql9.wav'
```

argilo:
This should be fixed by #7515. :crossed_fingers:

PR #7515, opened 2024-09-30
blocks: Fix test_read_wav_1024 on Windows

argilo:
## Description
In #7497 I added a new test (`test_read_wav_1024`) which fails on Windows.

According to the [documentation for `NamedTemporaryFile`](https://docs.python.org/3/library/tempfile.html), the temporary file cannot be opened a second time on Windows unless `delete=False` is used. (Elsewhere in GNU Radio's test suite, this approach is used.) Since the file is not automatically deleted, I've also added an `os.unlink()` call to clean up the temporary file.

Tests for Wav File Source.

## Testing Done
Once the test suite runs, we'll see whether the failing test passes across all platforms.

## Checklist

### DIFF
--- a/gr-blocks/python/blocks/qa_wavfile.py
+++ b/gr-blocks/python/blocks/qa_wavfile.py
@@ -190,13 +190,14 @@ class test_wavefile(gr_unittest.TestCase):
 
     def test_read_wav_1024(self):
         for expected_len in range(1024, 32768, 1024):
-            with tempfile.NamedTemporaryFile(suffix=".wav") as tf:
+            with tempfile.NamedTemporaryFile(suffix=".wav", delete=False) as tf:
                 with wave.open(tf.name, "w") as wf:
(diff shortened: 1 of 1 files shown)


argilo:
Now the `os.unlink` fails on Windows. I'll change this PR to draft while I sort it out.

argilo:
Fixed.

Windows doesn't allow a file to deleted while it's open, and Wav File Source doesn't close the file until the block is destroyed, so I added `del src` and `del self.tb` to ensure that happens.

Perhaps the block should have a `close()` method like the File Source.

argilo:
The Conda OSX CI failure is unrelated.

willcode:
Does `with` have any meaning with `delete=False`? Just curious?

argilo:
> Does with have any meaning with delete=False?

It probably doesn't do anything in that case. I just forgot to remove it.

PR #7522, opened 2024-10-15
Apply ITU-T O.150 Standard Polynomials

mcdiarmid:
## Description

If no mask (mask=0) is provided for the GLFSR block, a default will be selected from a look-up-table `s_polynomial_masks`. The current table of default masks contains a valid maximal length sequence (MLS) for degrees 1..32. However, all of these defaults differ to industry standard, so more often than not the user will need to specify the mask they desire. I ran into this problem at my previous job, where all of the commercial-off-the-shelf (COTS) modems that implement the IRIG and CCSDS standards, used differing PN sequence polynomials to the GNURadio defaults.

Where do the current polynomials come from? I came across this link (https://users.ece.cmu.edu/~koopman/lfsr/index.html) while researching standards that define PRBS/PN sequences. All instances of the PN sequence polynomials defined in `s_polynomial_masks` are actually the first/smallest polynomial that defines a viable MLS for the given degree. The first line on
each of the links in the "Text file (First hundred entries)" column corresponds to the entry currently defined in `s_polynomial_masks`.

I propose that the following standards should be implemented for default PN sequence polynomials:

- [ITU-T O.150](https://www.itu.int/rec/T-REC-O.150-199605-I/en) - (9, 11, 15, 19, 20, 23, 29, 31)
- TODO - find a good source for the information contained in [this link](https://www.digikey.co.nz/en/articles/use-readily-available-components-generate-binary-sequences-white-noise), as it aligns with ITU-T O.150 and implements other polynomials for degree 2..32 not found in ITU-T O.150.

Furthermore, perhaps outside of the scope of this MR - the way the integer masks and polynomials are related can be slightly confusing. E.g. the polynomial for PN11 - x^11 + x^9 + 1 is actually represented as (1<<10) | (1<<8). I found this discrepancy slightly confusing when working through understanding how to use this block.
## Checklist


mcdiarmid:
I don't think the workflow failure is due to the changes in this PR, but I don't have permission to re-run to check it's an intermittent failure.

willcode:
> I don't think the workflow failure is due to the changes in this PR, but I don't have permission to re-run to check it's an intermittent failure.

No, we're seeing that one intermittently on various targets lately.

willcode:
@marcusmueller @drmpeg looking for opinions on whether anyone would be using glfsr as a known-sequence scrambler, or just as a PRN source. If the former, people might depend on the sequence. It seems unlikely, but it's a behavior change.

marcusmueller:
I'd say it's most probably more commonly used as a source of a known cyclic sequence than not

mcdiarmid:
If changing default behaviour isn't the correct approach, perhaps either:
- These masks can be made available to import: `from gnuradio.digital.glfsr import PNMasks`...
- A "PN Sequence" Hier block could be added that doesn't have a mask parameter, with degree being used to select one of the proposed masks.

willcode:
I'm personally ok with this because it seems unlikely that someone has matched GR polys to a separate application that won't get updated. It could happen, and we'll probably hear from the one person who has saved data based on the current polys. I'll put out a note on the dev channel and see if we get more opinions.

drmpeg:
Just a note. This is used in pn_correlator_cc_impl.cc also.

willcode:
I think we're going to need the default to be backwards compatible, but an "ITU-compliant" switch could be added.

mcdiarmid:
> Just a note. This is used in pn_correlator_cc_impl.cc also.
Thanks for pointing that out.  Seems like the same logic is applied with defaults if mask=0, I'll keep this in mind.

> I think we're going to need the default to be backwards compatible, but an "ITU-compliant" switch could be added.
I'll add a yes/no drop-down box.   Set to no (default) doesn't change behaviour of how the block is initialised.  Set to yes will hide the mask parameter.  I'll also try expose the new table of polynomials via the Python binding API.

mcdiarmid:
Or, perhaps it just makes sense if I just add a table of these polynomials and mask representation to the GLFSR Block wiki page, and perhaps make reference to this in the block Details pane.  I'll still keep the change to the block yml file though as a hex representation is much nicer for bit-fields.

What do you guys think is more appropriate?

willcode:
From the user point of view, the easiest thing would be a "GR" and "ITU" selection, with "GR" as the default. We don't have any way of knowing whether anyone depends on the existing polys.

This would require a backward compatible change to the API, adding something like `itu=false` as the last param wherever this is used. I see:

- pn_correlator_cc
- glfsr_source_b
- glfsr_source_f

The param needs to be plumbed through the Python/Pybind11/C++ layers.

PR #7523, opened 2024-10-15
grc-qt: some improvements on the PropertyDialog

dl1ksv:
## Description

Multiline text cannot be saved  ( In comments)
This is fixed now

The return key is connected to the accept button. So you have to reopen the dialog, to do further changes. Now this key only finishes the input on an entry and you can modify this entry again, without reopening.

After changing a parameter the gui is now updated.

For instance open a filter block and change the window parameter from Hamming to Kaiser, than the Beta parameter appears.

## Testing Done

Open a highpass filter block.

Change the default window type to kaiser. The Beta parameter should appear.

Go to the "Adanced tab" and try to add a comment and  try to save.

## Checklist

### DIFF
--- a/grc/gui_qt/components/dialogs.py
+++ b/grc/gui_qt/components/dialogs.py
@@ -79,95 +79,17 @@ class PropsDialog(QDialog):
         self.edit_params = []
 
         self.tabs = QTabWidget()
-        ignore_dtype_labels = ["_multiline", "_multiline_python_external", "file_open", "file_save"]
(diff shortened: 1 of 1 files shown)


dl1ksv:
I detected a small bug, that shoul be fixed before merging.

dl1ksv:
test_qtbot.py works with the assumption that the Qt.Key_Enter triggers the accept button. That's not true any more. So the tests must be modified. I'd prefer to do thi in a different pr

PR #7527, opened 2024-10-15
Test clang tidy action

marcusmueller:
This is just to test whether CI will deal correctly with clang-tidy / git refs . **Don't merge**.


marcusmueller:
OK; this needs PCH work. Will come back to this later.

But good to know how buggy actions/checkout is: doesn't use the right user for checking out, so if one wants to do anything with the git repo after, one needs to add an exception for that. They (as of v4.2.1) don't expose the runner's ability to run git as the right user …

marcusmueller:
image
:tada:

marcusmueller:
Will probably want to use clang**d**-tidy, with a custom .clangd file for the action.

PR #7532, opened 2024-10-27
analog: precalculate squelch ramp window values

phyBrackets:
analog: precalculate squelch ramp window values

Optimize squelch ramping calculations by storing precomputed cosine window 
values in a lookup table. Updates occur only when ramp size changes.

## Description
The squelch implementation was performing costly cosine calculations for 
every sample during attack and decay transitions. This change precalculates 
these values once when the ramp size is set, significantly improving 
performance during runtime transitions.

- gr-analog/lib/squelch_base_ff_impl.cc
- Performance improvement in squelch transitions

## Checklist


gr-analog/lib/squelch_base_ff_impl.cc
```diff
@@ -33,13 +33,31 @@ squelch_base_ff_impl::squelch_base_ff_impl(const char* name, int ramp, bool gate
...
+void squelch_base_ff_impl::precalculate_window()
+{
+    if (d_ramp <= 0) {
+        d_window_table.clear();
+        return;
+    }
+
+    d_window_table.resize(d_ramp + 1);
+    for (int i = 0; i <= d_ramp; i++) {
+        d_window_table[i] = 0.5 - std::cos(GR_M_PI * i / d_ramp) / 2.0;
```
marcusmueller, line 50: this is a narrowing conversion from `double` to `float`; and we don't need the double precision here, so while we pre-calculate a table, let's also improve this code, use `0.5f` in place of `0.5`, calculate a `const float increment = static_cast<float>(GR_M_PI) / d_ramp`, and use that, as in `std::cos(i * increment)`.

gr-analog/lib/squelch_base_ff_impl.cc
```diff
@@ -33,13 +33,31 @@ squelch_base_ff_impl::squelch_base_ff_impl(const char* name, int ramp, bool gate
     d_state = ST_MUTED;
     d_envelope = d_ramp ? 0.0 : 1.0;
```
marcusmueller, line 34: ah, this isn't your fault at all, but! 

If someone sets `ramp` to something > 0 in the constructor, then `d_envelope` will take different during operation. So far, so good.

Now the user later on calls `set_ramp(0)`. See the problem? In that case, we need to set `d_envelope = 1.0f`. So, please amend your `set_ramp` to check that case :)

gr-analog/lib/squelch_base_ff_impl.cc
```diff
@@ -100,8 +119,10 @@ int squelch_base_ff_impl::general_work(int noutput_items,
             break;
 
         case ST_DECAY:
-            // FIXME: precalculate window for speed
-            d_envelope = 0.5 - std::cos(GR_M_PI * (--d_ramped) / d_ramp) / 2.0;
+            if (!d_window_table.empty()) {
+                d_envelope = d_window_table[--d_ramped];
+            }
+
             if (d_ramped == 0.0) {
```
marcusmueller, line 126: huh, this code is comparing `int d_ramped` with `double` 0.0; can you fix that while you're at it?

marcusmueller (changes requested):
Thanks! This generally looks pretty good, small, but still relevant, changes necessary.

marcusmueller (approved):

phyBrackets:
Hey @marcusmueller Is it possible for me to merge? Currently Merging is blocked with "The base branch restricts merging to authorized users".

willcode:
@phyBrackets only maintainers can merge.

There is no bound on `d_ramp`. For a fast stream, the table could be very large. True, this is usually used at audio rates and squelch attack (and usually delay) is usually a relatively small number of samples. But, on the other hand, what is the actual performance advantage of making this into a table if it is used only in low rate paths?

I think we'd need to see at test that shows a performance advantage at low rates AND a way of handling high rates via interpolation. Also, there is a complex version of this block that should get any changes we add to the float version.

marcusmueller:
@willcode  bit of historical context: Last time I touched this file, I left the `// FIXME: precalculate window for speed` in there last time I touched the block (it was added with the original version of this block by trondeau ca 12 years ago), as I assumed "short" ramps were to be expected. (also, I probably wasn't paying very much attention.)

You're right, however, the computational benefit is indeed questionable; trig functions aren't really expensive to compute in this context. Do you think it's worth further pursuing this PR? @phyBrackets might come up with good benchmarks, but they will undoubtedly show that starting at some ramp length, the downsides of doubling the memory bandwidth required will outshine the computational savings.

If we say "this is for audio rates, anyways", then the computational optimization makes no sense – not even 12 years ago. However, if that's the case, why do we even *have* the complex variant?

So, I think in the interest of keeping this code maintainable, we'd want essentially the same code in both `squelch_base_ff_impl` and `squelch_base_cc_impl`, and because performance gains are probably modest for short ramps and strongly negative for long ramps, I'd argue that in the interest of not prematurely optimizing something we have no acute need for, we might want to change the direction this PR is taking:

1. Code should be the same (essentially) in `squelch_base_ff_impl` and `squelch_base_cc_impl`, and we only do EITHER tables OR in-place calculation
2. the `// FIXME: ` needs to be removed, either way. Non-critical performance increases would be a "TODO", thinking about it. (that's why I put the "Low" label on this PR, anyways)
3. I'd argue a lot of the things @phyBrackets did here are good and worth keeping – mostly, the avoidance of calculating a double-precision `std::cos(double)` where a single-precision cosine would have done probably really made this fast enough. And the `d_envelope`-related bugfixes and checks are a should-have anyways!

@phyBrackets  do you think that's a direction you might be willing to take this? Not doing the conversion to a table lookup, but simply fixing our ramp handling and doing the conversion to single precision, as you have already done for your table calculation?

willcode:
A good first step would be to show that there is a performance problem. Otherwise, no need to change anything just to be clever.

marcusmueller:
I'd argue that the bugfix (`d_envelope` doesn't get set to 1.0f when ramp gets set to 0) and the no-downsides `std::cos(double)`->`std::cos(float)` conversion would be desirable

PR #7535, opened 2024-11-02
runtime: Fix minoutbuf/maxoutbuf for blocks with unlimited outputs

### LINKED ISSUE #7534: Minoutbuf & Maxoutbuf do not work for blocks with an unlimited number of outputs
### What happened?

Setting the "Minoutbuf" or "Maxoutbuf" advanced properties in GRC has no effect if the block has an unlimited number of outputs (e.g. Stream Demux, Selector, Null Source, Chunks to Symbols).

This happens because GRC generates calls to `set_min_output_buffer` or `set_max_output_buffer` without specifying output port numbers, and these methods do nothing if `output_signature()->max_streams()` is -1, i.e. unlimited.

https://github.com/gnuradio/gnuradio/blob/c6e9d6b6503c63124eca0a88a55b55344c9658ee/gnuradio-runtime/lib/block.cc#L345-L350

https://github.com/gnuradio/gnuradio/blob/c6e9d6b6503c63124eca0a88a55b55344c9658ee/gnuradio-runtime/lib/block.cc#L367-L374

One possible solution might be for GRC to generate one `set_min_output_buffer` or `set_max_output_buffer` call per output port.

### System Information

OS: Ubuntu 24.04
GR Installation Method: PyBOMBS

### GNU Radio Version

3.10 (maint-3.10)

### Specific Version

3.10.11.0

Connect a Null Source to a Null Sink, and set the Null Source's Maxoutbuf to 1. Observe that no warning is generated when running the flow graph.

Replace the Null Source with a Signal Source, and set its Maxoutbuf to 1. Observe that a warning is now produced: "max output buffer set to 512 instead of requested 1"

### Relevant log output

_No response_

phyBrackets:
Modifies set_min_output_buffer and set_max_output_buffer methods to properly
handle blocks that have unlimited output streams (-1). Uses actual number of
connected outputs instead of max_streams when setting buffer sizes.

## Description
This change fixes an issue where setting Minoutbuf or Maxoutbuf advanced properties in GRC had no effect on blocks with unlimited outputs. The problem occurred because the buffer setting methods were using output_signature()->max_streams() directly, which returns -1 for unlimited outputs, causing the buffer settings to be skipped entirely.

## Related Issue
Fixes #7534 

## Checklist


argilo:
In programs generated by GRC, `set_max_output_buffer` is called before connections are made, so the fallback case (`max_streams = 1`) is always used, regardless of the actual number of outputs. So I don't think this will work correctly in most cases.

argilo:
> or starts with a minimum size vector that expands as ports are connected

I'm not sure whether it's a good idea to have `set_max_output_buffer(long max_output_buffer)` touch ports that are created in the future.

The fix I had in mind was to modify GRC so that it generates calls to `set_max_output_buffer(int port, long max_output_buffer)` instead. GRC knows how many output ports are going to be connected.

argilo:
I suppose it could make sense to have `set_max_output_buffer(long max_output_buffer)` control a global setting for all the block's outputs, but we would need to decide what happens if both `set_max_output_buffer(long max_output_buffer)` and `set_max_output_buffer(int port, long max_output_buffer)` are called for the same block, and document that.

PR #7544, opened 2024-11-12
uhd: Update RFNoC image builder GRC bindings

mbr0wn:
*NOTE*: Despite the magnitude of this change, this doesn't actually touch anything regarding standard gr-uhd usage. This only affects the ability to build FPGA bitfiles using GNU Radio Companion. This feature is currently broken, and this PR makes it workable. That said, I don't think people are trying to use it a lot anyway.

## Commit Message

This updates the portion of gr-uhd that deals with generating RFNoC bitfiles from within GRC. For now, this is limited to generating a .grc file with GNU Radio Companion that describes an RFNoC image core; this file must then be passed to `rfnoc_image_builder` with the `-r` argument, the same way that regular image core YAML files are passed with the `-y` argument.

The following changes are included:
- All available image core blocks were updated to match recent UHD
- New image core blocks were added (addsub, CHDR DMA transport adapter, no-connect, fosphor aka f15, keep-one-in-N, null source/sink, replay/capture, stream endpoint, signal generator, split-stream, switchboard).
- The X310 BSP block was replaced with a generic BSP block, which supports all currently available RFNoC devices.
- The domains were updated, and domains for axi4_mm, gpio, and transport adapters were added.
- Two examples for image core files were added: An X3x0 image, which can be used to generate X300_HG, X300_XG, X310_HG, and X310_XG standard bitfiles, and an x410_X4_200 example.
- A README.md is added in gr-uhd/examples/grc to better explain what's going on.

The usage of this to generate RFNoC bitfiles depends on the availability of updated UHD. However, note that with current UHD and GNU Radio, the image builder feature of gr-uhd was not functional.

## Testing Done

I built a bunch of bitfiles for X310 and X410 using the latest version of `rfnoc_image_builder`.

## Checklist


gr-uhd/examples/grc/rfnoc_images/x310_image_core.grc
```diff
@@ -0,0 +1,559 @@
+options:
```
mbr0wn, line 1: As a courtesy this is the GRC file:

x310_image_core

gr-uhd/examples/grc/rfnoc_images/x410_X4_200_image_core.grc
```diff
@@ -0,0 +1,533 @@
+options:
```
mbr0wn, line 1: GRC:
x410_X4_200_image_core

marcusmueller:
Love the fact that you included pictures of the flow graphs added. That is very helpful, and I think it's something we might want to ask people to do whenever they add a new flow graph :heart:

gr-uhd/examples/grc/README.md
marcusmueller, line 1: This is extremely good. We should copy it onto a wiki page (I know, copying is a bit antithetical to a software dev), and link the rfnoc_ and uhd_ block doc pages to it, maybe?

gr-uhd/examples/grc/rfnoc_images/x410_X4_200_image_core.grc
marcusmueller: Excellent, more of a functional question: "Virtual Port"s in GRC work with NoC domain connections, as well, right? So in theory, we could have a `ce_clk` virtual port connecting to the BSP block's output and to all noc shell wrapper block's inputs?

gr-uhd/grc/CMakeLists.txt
marcusmueller, line 1: These entries seem rather randomly ordered? Would it make sense to order them alphabetically (.block and .domain separately)? (because reordering would obfuscate which things you added in the PR, that would preferably be done in a separate commit)

gr-uhd/grc/uhd_fpga_device_bsp.block.yml
marcusmueller, line 1: I like the architectural choice here to allow for explicit definition of clocks in the same block "gizmo" that defines the overall properties of the generated bitstream.

gr-uhd/grc/uhd_fpga_device_none.block.yml
```diff
@@ -0,0 +1,39 @@
...
+    label: Direction
+    dtype: enum
+    default: 'in'
+    options: ['in', 'out']
+    hide: part
+
+inputs:
+-   domain: rfnoc.data
+    id: term_in
+    dtype: fc32
```
marcusmueller, line 22: is the fixed dtype intentional?

gr-uhd/grc/uhd_fpga_fosphor.block.yml
```diff
@@ -0,0 +1,38 @@
...
+    id: hist
+    dtype: fc32
+    multiplicity: ${ NUM_PORTS }
+-   domain: rfnoc.data
+    id: wf
+    dtype: fc32
+    multiplicity: ${ NUM_PORTS }
+
+documentation: |-
+  RFNoC F15 (fosphor) (FPGA Implementation).
```
marcusmueller, line 34: you'll have to explain what the *F15* stands for, somewhere :D

gr-uhd/grc/uhd_fpga_device_none.block.yml
marcusmueller: Ah wait, GRC only sees FC32 for all these interfaces, anyways. so, LGTM

marcusmueller (approved):
Exemplary PR!

gr-uhd/grc/uhd_fpga_null_src_sink.block.yml
```diff
@@ -0,0 +1,51 @@
...
+    id: source
+    dtype: fc32
+    multiplicity: ${ NUM_PORTS }
+-   domain: rfnoc.data
+    id: loop
+    dtype: fc32
+    multiplicity: ${ NUM_PORTS }
+
+documentation: |-
+  RFNoC keep-one-in-n block (FPGA Implementation).
```
zacharyr, line 48: This is not the correct doc stub for this block.

gr-uhd/grc/uhd_fpga_sep.block.yml
```diff
@@ -23,23 +23,40 @@ parameters:
...
+
+  There must always be one streaming endpoint that is configured to carry
+  control traffic. This is typically the first SEP in the flowgraph. For all
+  other SEPs, the data traffic option should be set to "Yes".
+
+  The buffer size should be chosen depending on the Tx (host to device) streaming
+  capabilities that are required. If no streaming to the device is required, it
+  may be set to zero to save FPGA resources. The larger the buffer size, the
+  better the streaming endpoint can cope with bursty traffic and high rates.
+  However, not that the available storage resources on the FPGA are limited. wa
```
zacharyr, line 60: Not sure what "wa" is, did something get cut off here?

gr-uhd/grc/uhd_fpga_device_none.block.yml
mbr0wn: The whole `dtype` thing is suboptimal. Without changing GRC, the best I can do is pick a `dtype` (arbitrarily) for every RFNoC port type. Obviously, `fc32` has nothing to do with `rfnoc.data` but it's something that GRC can work with.

gr-uhd/grc/uhd_fpga_null_src_sink.block.yml
mbr0wn: Thanks for catching -- fixed.

gr-uhd/grc/uhd_fpga_sep.block.yml
mbr0wn: It means "this developer uses vim and sometimes doesn't switch to normal mode before typing commands". I'll fix this.

gr-uhd/examples/grc/rfnoc_images/x410_X4_200_image_core.grc
mbr0wn: That's a good catch, and something that I didn't think of. It won't work, though, because the `rfnoc_image_builder` would require an update (the GRC file is not parsed by GNU Radio, but by UHD for the image building). I think that's a good change to make in UHD.

gr-uhd/grc/CMakeLists.txt
mbr0wn: I probably had some idea about how to sort them, but sure, you're right let's sort them alphabetically.

mbr0wn: OK, I sorted them by category (with comment) and then alphabetically.

gr-uhd/grc/uhd_fpga_device_bsp.block.yml
mbr0wn: Thanks! 😄

gr-uhd/examples/grc/README.md
mbr0wn: Yes, we definitely should do something like this. I haven't made up my mind where the best place would be (GNU Radio wiki vs. Ettus KB), plus, I'm, working on some more updates to GNU Radio to make this even nicer.

gr-uhd/grc/uhd_fpga_fosphor.block.yml
mbr0wn: Honestly, I don't know ☹️ . Sylvain started calling fosphor-in-rfnoc that as a shorthand, and I know we talked about it, but if we did I can't remember. Maybe it's short for fosphor-over-rfnoc (f + 15 letters).

willcode:
@mbr0wn could you just verify that this is meant to be backward compatible and backported? The changes to the tree file caught my eye.

gr-uhd/examples/grc/README.md
marcusmueller: Hm, if we're talking blocks and flow graphs shipped with GNU Radio, my guess is that customers would look at GNU Radio for documentation.

gr-uhd/grc/uhd_fpga_fosphor.block.yml
marcusmueller: @smunaut can you chime in?

smunaut: 15 is the atomic number for Phosphorus

For me it's fine to just use "RFNoC fosphor".  My main concern was in another place where it was just called "fosphor" without the "RFNoC" denomination and that is a different thing.

mbr0wn: Ah that's right! I know remember the conversation. Thanks for the reminder!

And yes, I'll be careful to distinguish this from regular fosphor wherever appropriate.

PR #7547, opened 2024-11-18
gr: improve system resource release: hier blocks and message edges

SaikWolf:
gr: improve system resource release: hier blocks and message edges

Python-passed objects don't currently release their resources, bogging down the process.
This should release the hier_block2 and msg_edges resources

## Description
The pybind11 interface treats resources a little differently and unless the process is closed those resources get bogged down in more complex use cases (many flowgraphs).

These resources were being held in the message edges in the flowgraph and the hier blocks, especially the recursive use of the hier blocks. These changes explicitly address that issue.

## Related Issue
Unknown

All blocks. Enhances teardown.

## Testing Done
The problem was observed in the development of https://github.com/vtnsiSDD/rfrl-gym
By implementing these changes the resource bottleneck was cleared up improving the functionality.

## Checklist

### DIFF
--- a/gnuradio-runtime/lib/hier_block2_detail.cc
+++ b/gnuradio-runtime/lib/hier_block2_detail.cc
@@ -496,9 +496,27 @@ endpoint_vector_t hier_block2_detail::resolve_port(int port, bool is_input)
     return result;
 }
 
+void hier_block2_detail::recursive_disconnect_all(const std::string& caller)
+{
+    d_debug_logger->debug("Disconnect hier_block2 recursive...");
+    std::vector<basic_block_vector_t> paths = d_fg->partition();
+    for(std::vector<basic_block_vector_t>::iterator p = paths.begin(); p != paths.end(); p++){
+        for(basic_block_viter_t bp = (*p).begin(); bp != (*p).end(); bp++){
+            if (this->d_owner->identifier() != caller){
+                hier_block2_sptr hh = std::dynamic_pointer_cast<hier_block2>((*bp));
+                if (hh != 0){
+                    hh->d_detail->recursive_disconnect_all(hh->identifier());
+                }
+            }
+        }
+    }
+    d_debug_logger->debug("Disconnect hier_block2 recursive...finished");
+}
+
 void hier_block2_detail::disconnect_all()
 {
     d_debug_logger->debug("Disconnect all...");
+    recursive_disconnect_all(this->d_owner->identifier());
     reset_hier_blocks_parent();
     d_fg->clear();
     d_blocks.clear();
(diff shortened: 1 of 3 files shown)


marcusmueller:
is it intentional that this was done againt `maint-3.10` instead of against `main`?

marcusmueller:
The formatting problems can be best resolved using `clang-format -i filename.cc` (rather than doing them manually). Many text editors/IDEs have handy plugins or shortcuts to run clang-format on the current source file.

marcusmueller:
I've made rebase on `main` [here](https://github.com/gnuradio/gnuradio/compare/main...marcusmueller:gnuradio:SaikWolf-pr7547?expand=1), to verify it's straightforward. It is!

gnuradio-runtime/lib/hier_block2_detail.cc
```diff
@@ -496,9 +496,27 @@ endpoint_vector_t hier_block2_detail::resolve_port(int port, bool is_input)
     return result;
 }
 
+void hier_block2_detail::recursive_disconnect_all(const std::string& caller)
+{
+    d_debug_logger->debug("Disconnect hier_block2 recursive...");
+    std::vector<basic_block_vector_t> paths = d_fg->partition();
+    for (std::vector<basic_block_vector_t>::iterator p = paths.begin(); p != paths.end(); p++) {
+        for (basic_block_viter_t bp = (*p).begin(); bp != (*p).end(); bp++) {
+            if (this->d_owner->identifier() != caller) {
```
marcusmueller, line 505: Does `this->d_owner` change in these loops? If not, then this should be an outer condition, not an inner condition: "If we are already owned by the caller, don't recursively descend" is better than "recursively descend and while we're doing so, check for every node we find whether we might be owned by the caller"

gnuradio-runtime/lib/hier_block2_detail.cc
```diff
@@ -496,9 +496,27 @@ endpoint_vector_t hier_block2_detail::resolve_port(int port, bool is_input)
...
+                hier_block2_sptr hh = std::dynamic_pointer_cast<hier_block2>((*bp));
+                if (hh != 0) {
```
marcusmueller, line 507: I'd argue that it would be idiomatic C++ here to simply check `if(hh)` :)

gnuradio-runtime/lib/hier_block2_detail.cc
```diff
@@ -496,9 +496,27 @@ endpoint_vector_t hier_block2_detail::resolve_port(int port, bool is_input)
     return result;
 }
 
+void hier_block2_detail::recursive_disconnect_all(const std::string& caller)
```
marcusmueller, line 499: why do we have to identify the caller it by its string identifier? Why can't we simply use the pointer?

marcusmueller (changes requested):
I think this needs to be done against `main`, not `maint-3.10`, but could well be backported.

The formatting problem needs to be solved, though I think my range-based `for` proposal below is an elegant solution to the problem of overly long lines :)

Since no new blocks are allocated during disconnection, we can simply compare the address of the caller with `d_owner.get()` of this detail, instead of doing the string identifier detour.

willcode:
@SaikWolf please close this PR and move discussion to #7549

PR #7548, opened 2024-11-21
GRC: Pluggable Workflows Reprise

mbr0wn:
This PR replaces https://github.com/gnuradio/gnuradio/pull/7445 by @ZakyHermawan. While that PR already had most of the meat, it had a bunch of issues that I tried taking care of here. Most importantly, there was a lot of code duplication in the original PR that would have made maintenance really hard (e.g., duplicating the generator code between hier blocks and top blocks, instead of sharing). I also had some issues just to get things running in the first place through all workflows (Qt GRC, compiler, ...).

Also, I added a new workflow to actually test the feature: A workflow to build RFNoC bitfiles from within GRC. I will remove that commit from the PR before I make it a non-draft, but I'll leave it in for now so people can see how this all works together.

@ZakyHermawan I couldn't add you as a reviewer, but as the original author of much of this code, I appreciate your input.

## Description

Please read the commit messages. High level, this makes workflows in GRC (which is the code that gets generated dependent on output language and generate options) a plugin. We provide all plugins to enable all the same workflows that existed before (C++/Python, Hier Block or Top Block, Qt and no GUI, and a Python-bokeh-gui workflow).

Really, it depends all GRC operations.

## Testing Done

I've generated a lot of flow graphs. I also generated code from GRC with both `main` and this branch and diffed the output.

There is one difference: The imports are ordered differently. I wasn't able to fix that immediately, but because imports were already not in an order that would make linters happy, I decided it wasn't a big deal (at least there are no missing imports, which would have been an issue!).

As for the RFNoC workflow: This meshes well with the recent updates to designing RFNoC bitfiles from within GRC. Without this change, you need to save the .grc file that represents the RFNoC image, and then feed it into the command line tool `rfnoc_image_builder` manually on the command line. Now, the bitfile is generated when you press F6 (Execute).

When you hit F5 (Generate), it will generate the build artifacts that are required to generate a bitfile, if you want to check them out. However, when you do "Execute", it will skip the generation step (this is also a new feature in GRC). That's because the way the image builder works is it always generates the output artifacts before building, and a double-generation would be redundant.

Here's what it looks like from the GTK version:

image

If I hit F5, then you can see in the console how the image builder shows its output from generating:

image

If I hit F6, it skips the generation and goes straight to the build. This can take some hours depending on the device, so we launch it through a terminal:

image

## Checklist>


gr-uhd/examples/grc/rfnoc_images/x310_image_core.grc
```diff
@@ -1,34 +1,37 @@
 options:
```
mbr0wn, line 1: This is the same GRC file as before, only with a different output language. However, that means new generate options, which explains the diff.

x310_image_core

willcode:
CI: `platform.load_and_generate_flow_graph()` returns `None` for `generator`.

mbr0wn:
> CI: `platform.load_and_generate_flow_graph()` returns `None` for `generator`.

Thanks; I saw that too. Funny thing is, I was locally testing on 22.04 (which fails here). But I'll figure it out. I don't think this is a showstopper for reviewing this 😄 .

haakov:
Nice! I've built the branch and will be testing this out tomorrow.

haakov:
GRC starts, but I'm getting this error
```
Error while loading /usr/local/share/gnuradio/grc/blocks/rust.workflow.yml
  File "/usr/local/lib/python3.12/site-packages/gnuradio/grc/core/platform.py", line 171, in build_library
    loader(data, file_path)
  File "/usr/local/lib/python3.12/site-packages/gnuradio/grc/core/workflow.py", line 67, in load_workflow
    wf = Workflow(filepath)
         ^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/gnuradio/grc/core/workflow.py", line 38, in __init__
    self.generate_options = self.workflow_params.pop('generate_options')
                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
KeyError: 'generate_options'
```

haakov:
Works on my machine, but it seems that the CI error is caused by GRC not being able to find the `output_language`: 
`>>> Load Error: /__w/gnuradio/gnuradio/grc/tests/resources/test_cpp.grc: No workflow found for output_language:  and generate_options: qt_gui`

https://github.com/gnuradio/gnuradio/pull/7548/files#diff-acc36795c5e592249b83ecd5860c5c3277f91b7eeb503145d4ff59f841193b39R328

mbr0wn:
I'm a bit optimistic about the tests now....

mbr0wn:
> Works on my machine, but it seems that the CI error is caused by GRC not being able to find the `output_language`: `>>> Load Error: /__w/gnuradio/gnuradio/grc/tests/resources/test_cpp.grc: No workflow found for output_language: and generate_options: qt_gui`

Yeah, I was following that too. The problem (and the reason why it worked for me) was that in a non-installed state, GRC can't finde the workflow files. The error is a bit weird, but I moved them into `grc/blocks/`.

grc/compiler.py
```diff
@@ -56,10 +56,17 @@ def main(args=None):
...
+        flow_graph, generator = platform.load_and_generate_flow_graph(
             os.path.abspath(grc_file), os.path.abspath(output_dir))
-        if not file_path:
+        if flow_graph is None and generator is None:
+            exit('Error: Failed to load ' + grc_file)
+        if generator is None:
+            exit("Error: Failed to create generator")
+        if not generator.file_path:
             exit('Compilation error')
     if file_path and args.run:
```
haakov, line 67: `file_path` is always None

grc/core/blocks/options.py
```diff
@@ -0,0 +1,362 @@
...
+SPDX-License-Identifier: GPL-3.0-or-later
+"""
+
+from . import Block, register_build_in
+from ._templates import MakoTemplates
+from ._build import build_params
+from ..base import Element
+
+from collections import OrderedDict, defaultdict
+import typing
```
haakov, line 11: Unused import?

grc/core/blocks/options.py
```diff
@@ -0,0 +1,362 @@
...
+        tmp_dct.attributes = defaultdict(dict)
+        for generated_options, generated_options_label in list_of_generated_options_pair:
+            tmp_dct[generated_options] = generated_options_label
+        self.params['generate_options'].options = tmp_dct
+
+        # Update the generator_class_name and generator_module parameters based
+        # on the current workflow
+        self.params['generator_class_name'].set_value(self.current_workflow.generator_class)
+        self.params['generator_module'].set_value(self.current_workflow.generator_module)
+        Element.rewrite(self)
```
haakov, line 303: superfluous?

grc/core/blocks/options.py
mbr0wn: yup

grc/core/blocks/options.py
mbr0wn: Yes, good catch. I was debating whether or not to remove this line or the `super().rewrite()`, because we don't use the super-rewrite anywhere else. That said, this does not inherit directly from `Element` so I removed this line like you suggested.

mbr0wn:
@haakov I removed the UHD/RFNoC workflow, so it's down to 1 commit. As mentioned before, I would like some feedback if this meets the bar for backporting -- unless @marcusmueller really wants to release a 3.11 for workgroups before GRCon.

haakov (approved):

marcusmueller:
daaamn, I just mixed up reviews for #7792 and this. Sorry! Could you please look at #7792's review comments, @haakov?

marcusmueller:
Regarding backports: Can a 3.10.0.0-GRC still open a GRC file that was just opened, not functionally modified, then saved with a GRC with this changeset?

haakov:
> Regarding backports: Can a 3.10.0.0-GRC still open a GRC file that was just opened, not functionally modified, then saved with a GRC with this changeset?

Great point. It'll open it, but the parameters will not be set correctly

haakov:
Hmm I suppose ideally GRC should notify the user upon opening the flowgraph that "this flowgraph has a workflow that I don't know about"

marcusmueller:
> Great point. It'll open it, but the parameters will not be set correctly

If I interpret this correctly, we need to bump `metadata->file_format` in our YAML output.
Which is totally OK, if we make changes that mean that new files can't be opened with old GRCs, then we should make that discoverable, and that's what that's for.

(Side note: does GRC *check* `file_format`?)

Unless we make it optional to write the old format, that would be exactly the kind of breaking behaviour we need to do a 3.11 release for. So, maybe this isn't sensibly backportable?

haakov:
GRC core does check `file_format`:
https://github.com/gnuradio/gnuradio/blob/1441c4d6d4e67fa58701bc85d4171aafc4dcf245/grc/core/platform.py#L365-L367


`FLOW_GRAPH_FILE_FORMAT_VERSION` defined here:
https://github.com/gnuradio/gnuradio/blob/1441c4d6d4e67fa58701bc85d4171aafc4dcf245/grc/core/Constants.py#L33-L39

haakov:
About your #7792 comments, I generally agree with all of them, but I think most of them should be fixed in a separate PR. It'll be easier (for me, at least) to divide and conquer them after this PR has been merged

I agree, it makes sense to bump the `file_format` to 3 for this changeset, and at that point it'll be hard to justify backporting this PR. Otherwise, a flowgraph created in 3.10 will not open correctly on all 3.10-versions of GR.

haakov:
PR opened here with fixes for comments that I've :+1:'ed: https://github.com/mbr0wn/gnuradio/pull/6
I think the other comments should be addressed in separate issues/PRs

haakov:
Commit that increases `file_format` for flow graphs here (I haven't had the chance to test it): https://github.com/haakov/gnuradio/commit/cf2a736a1b17fac187beb7e84cb0b6dcf9fceff0

mbr0wn:
> > Regarding backports: Can a 3.10.0.0-GRC still open a GRC file that was just opened, not functionally modified, then saved with a GRC with this changeset?
> 
> Great point. It'll open it, but the parameters will not be set correctly

@haakov can you elaborate on that? I tried doing this, and it worked fine. The .grc file will have a few options stored that the old GRC doesn't understand, but there doesn't seem to be any problem with this. Note that most of the actual `.grc` files is not changed with this PR.

marcusmueller (approved):
OK having verified Python generation even of problematic flow graphs still works, I'm giving this the thumbs up: 
:+1: :+1:  
(emojis discriminate against people with symmetric, instead of identical hands)

haakov:
> OK having verified Python generation even of problematic flow graphs still works, I'm giving this the thumbs up: 👍 👍 (emojis discriminate against people with symmetric, instead of identical hands)

Agreed!

marcusmueller:
:tada:

dkozel:
@haakov Are there some specific checks or features you think should be added to GRC before this is merged to maint-3.10?

Issue #7456, opened 2024-07-30
`wavfile_sink` does not continue after lock/unlock


baskiton:
### What happened?

It's bug or feature? WAV file sinking stopped after lock()/unlock(). So any flowgraph reconfigure are not posible

### System Information

OS: Debian 12
GR Installation Method: Conda (conda-forge)

### GNU Radio Version

3.10 (maint-3.10)

### Specific Version

3.10.10.0

### Steps to Reproduce the Problem

```python
import pathlib
import time

import gnuradio as gr
import gnuradio.analog
import gnuradio.blocks
import gnuradio.gr


def foo(var, sr, fp):
    print(f'test {var!r}')
    tb = gr.gr.top_block('Test')
    src = gr.analog.noise_source_f(gr.analog.GR_GAUSSIAN, 1, 0)
    thr = gr.blocks.throttle(gr.gr.sizeof_float, sr)

    if var == 'file':
        sink = gr.blocks.file_sink(gr.gr.sizeof_float, str(fp))
    else:
        sink = gr.blocks.wavfile_sink(
            str(fp),
            1,
            sr,
            gr.blocks.FORMAT_WAV,
            gr.blocks.FORMAT_FLOAT,
            False
        )

    tb.connect(src, thr, sink)

    tb.start()

    time.sleep(3)
    tb.lock()
    st = fp.stat()
    print(f'mt={st.st_mtime:.3f} sz={st.st_size}')
    tb.unlock()

    time.sleep(3)
    tb.stop()
    tb.wait()
    st = fp.stat()
    print(f'mt={st.st_mtime:.3f} sz={st.st_size}')


if __name__ == '__main__':
    print('GNURadio', gr.gr.version())
    sr = 48000
    fp = pathlib.Path('/tmp/sink_test')
    for var in ('file', 'wav'):
        foo(var, sr, fp)

```

### Relevant log output

```shell
GNURadio 3.10.10.0
test 'file'
mt=1722331004.135 sz=557024
mt=1722331007.207 sz=1114048
test 'wav'
mt=1722331010.287 sz=557104
mt=1722331010.387 sz=557104
```

baskiton:
In small projects this will work, but in larger projects, where wav_sink can be hidden deep in hier-blocks, this will become a huge crutch, which, moreover, is extremely difficult to manage

willcode:
Marked as a bug - hope the workaround helps until someone take a look at the problem.

sladecek:
[...]

PR #7550, opened 2024-11-22
blocks: Fix useless sample size functions in wavfile_sink.

sladecek:
## Description
The current wavfile_sink block API has getter and
setter functions for the sample bit size. However, the setter function's
value is ignored because the sample size is fully determined 
by the file format and subformat. Additionally, the code is inconsistent, 
as the getter returns bytes instead of bits, and there is a risk 
of using uninitialized memory.

This update removes unnecessary operations and
resolves the inconsistency between bits and bytes.

## Related Issue
This is preparation before solving 
https://github.com/gnuradio/gnuradio/issues/7456 

wawfile_sink

## Testing Done
make test

## Checklist

### DIFF
--- a/gr-blocks/lib/wavfile_sink_impl.cc
+++ b/gr-blocks/lib/wavfile_sink_impl.cc
@@ -47,7 +47,7 @@ wavfile_sink_impl::wavfile_sink_impl(const char* filename,
       d_new_fp(nullptr),
       d_updated(false)
 {
-    int bits_per_sample;
+    int bits_per_sample{ 8 };
 
     if (n_channels > s_max_channels) {
         throw std::runtime_error("Number of channels greater than " +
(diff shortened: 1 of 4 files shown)


gr-blocks/lib/wavfile_sink_impl.cc
```diff
@@ -302,13 +301,8 @@ int wavfile_sink_impl::work(int noutput_items,
...
 {
-    gr::thread::scoped_lock guard(d_mutex);
-    set_bits_per_sample_unlocked(bits_per_sample);
-}
-
-void wavfile_sink_impl::set_bits_per_sample_unlocked(int bits_per_sample)
-{
-    d_bytes_per_sample_new = bits_per_sample / 8;
+    d_logger->warn(
+        "set_bits_per_sample() does nothing. Bit size is controlled by the subformat.");
```
mbr0wn, line 305:
```suggestion
        "set_bits_per_sample() does nothing. Sample size is controlled by the subformat.");
```

PR #7554, opened 2024-11-27
network: fix build with Boost 1.87

landryb:
## Description

Build fails with boost 1.87b1, because of some deprecations that were removed from the API.

The PR might not be perfect/to polish, but at least its here so some other packagers trying to update their boost version can reuse the patches.

From my understanding, the `resolve.resolve` now only supports the 'split' arguments, not the `query` version.

not 100% comfy about the `cbegin()` + dereference use, as:
- im not a c++ hacker
- i know nothing to boost
- i never used gnuradio, so im not fully comfortable with the codebase. i just tried poking at it from a packager POV, trying to update our patched 3.8.2 to 3.10.11. Cf https://marc.info/?t=173235521700002&r=1&w=2

gr-network

## Testing Done

Builds fine on OpenBSD, using gnuradio 3.10.11.0 tarball (not built from master/git)

as for runtime testing, none done, i've asked SDR users on OpenBSD for feedback.

## Checklist

### DIFF
--- a/gr-network/lib/tcp_sink_impl.cc
+++ b/gr-network/lib/tcp_sink_impl.cc
@@ -63,10 +63,8 @@ bool tcp_sink_impl::start()
 
         std::string s_port = std::to_string(d_port);
         asio::ip::tcp::resolver resolver(d_io_context);
-        asio::ip::tcp::resolver::query query(
-            d_host, s_port, asio::ip::resolver_query_base::passive);
-
-        d_endpoint = *resolver.resolve(query, err);
+        d_endpoint = *(
+            resolver.resolve(d_host, s_port, asio::ip::tcp::resolver::passive).cbegin());
 
         if (err) {
             throw std::runtime_error(
@@ -159,7 +157,7 @@ void tcp_sink_impl::connect(bool initial_connection)
             d_acceptor = new asio::ip::tcp::acceptor(
                 d_io_context, asio::ip::tcp::endpoint(asio::ip::tcp::v4(), d_port));
     } else {
-        d_io_context.reset();
+        d_io_context.restart();
     }
 
     if (d_tcpsocket) {
@@ -194,7 +192,6 @@ bool tcp_sink_impl::stop()
         d_tcpsocket = NULL;
     }
 
-    d_io_context.reset();
     d_io_context.stop();
 
     if (d_acceptor) {
--- a/gr-network/lib/udp_sink_impl.cc
+++ b/gr-network/lib/udp_sink_impl.cc
@@ -125,11 +125,10 @@ bool udp_sink_impl::start()
     std::string str_port = std::to_string(d_port);
     std::string str_host = d_host.empty() ? std::string("localhost") : d_host;
     asio::ip::udp::resolver resolver(d_io_context);
-    asio::ip::udp::resolver::query query(
-        str_host, str_port, asio::ip::resolver_query_base::passive);
-
     asio::error_code err;
-    d_endpoint = *resolver.resolve(query, err);
+    d_endpoint =
+        *(resolver.resolve(str_host, str_port, asio::ip::tcp::resolver::passive, err)
+              .cbegin());
 
     if (err) {
         throw std::runtime_error(std::string("[UDP Sink] Unable to resolve host/IP: ") +
@@ -177,7 +176,6 @@ bool udp_sink_impl::stop()
         delete d_udpsocket;
         d_udpsocket = nullptr;
 
-        d_io_context.reset();
         d_io_context.stop();
     }
 
--- a/gr-network/lib/socket_pdu_impl.cc
+++ b/gr-network/lib/socket_pdu_impl.cc
@@ -54,9 +54,12 @@ socket_pdu_impl::socket_pdu_impl(std::string type,
         d_tcp_endpoint = asio::ip::tcp::endpoint(asio::ip::tcp::v4(), port_num);
     } else if ((type == "TCP_SERVER") || (type == "TCP_CLIENT")) {
         asio::ip::tcp::resolver resolver(d_io_context);
-        asio::ip::tcp::resolver::query query(
-            asio::ip::tcp::v4(), addr, port, asio::ip::resolver_query_base::passive);
-        d_tcp_endpoint = *resolver.resolve(query);
+        d_tcp_endpoint = *(resolver
+                               .resolve(asio::ip::tcp::v4(),
+                                        addr,
+                                        port,
+                                        asio::ip::resolver_query_base::passive)
+                               .cbegin());
     } else if ((type == "UDP_SERVER") &&
                ((addr.empty()) || (addr == "0.0.0.0"))) { // Bind on all interfaces
         int port_num = atoi(port.c_str());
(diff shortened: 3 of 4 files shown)


landryb:
i see that the CI failed on a test:
```
264/266 Test #261: qa_socket_pdu ................................***Failed    2.77 sec
.thread_body_wrapper :error: ERROR thread[thread-per-block[1]: <block socket_pdu(6)>]: send_to: Address family not supported by protocol [system:97 at $PREFIX/include/boost/asio/detail/reactive_socket_service.hpp:261 in function 'send_to']
```
and that's where i throw my hands in the air, and leave someone more knowledgeable in the codebase to look at that :)

marcusmueller:
@landryb out of interest, is OpenBSD planning to integrate Boost 1.87 anytime soon? I ask because debian experimental, fedora rawhide are on 1.83, on Alpine Linux you can at least get 1.84. But that's a bit away from 1.87; and this PR is basically a very valuable early-warning buildability bug report (keep these coming :D), but to realistically prioritize it, it'd be neat to know how soon this needs to be fixed.

landryb:
> @landryb out of interest, is OpenBSD planning to integrate Boost 1.87 anytime soon?

not 'soon' since so far only a b1 was released upstream for boost, and downstreams need to pickup the deprecations, but we (OpenBSD) try to be proactive on the build fixes, cf the thread at https://marc.info/?l=openbsd-ports&m=173201614532254&w=2

mbr0wn:
@marcusmueller This is also an issue with Homebrew, see https://github.com/Homebrew/homebrew-core/pull/198115. We just went through the effort of doing all of these changes on UHD, too.

gr-network/lib/udp_sink_impl.cc
```diff
@@ -125,11 +125,10 @@ bool udp_sink_impl::start()
     std::string str_port = std::to_string(d_port);
     std::string str_host = d_host.empty() ? std::string("localhost") : d_host;
     asio::ip::udp::resolver resolver(d_io_context);
-    asio::ip::udp::resolver::query query(
-        str_host, str_port, asio::ip::resolver_query_base::passive);
-
     asio::error_code err;
-    d_endpoint = *resolver.resolve(query, err);
+    d_endpoint =
+        *(resolver.resolve(d_host, str_port, asio::ip::tcp::resolver::passive, err)
```
mbr0wn, line 130: Why did you use `d_host` instead of `str_host` here?

gr-network/lib/tcp_sink_impl.cc
```diff
@@ -159,7 +157,7 @@ void tcp_sink_impl::connect(bool initial_connection)
             d_acceptor = new asio::ip::tcp::acceptor(
                 d_io_context, asio::ip::tcp::endpoint(asio::ip::tcp::v4(), d_port));
     } else {
-        d_io_context.reset();
+        d_io_context.restart();
```
mbr0wn, line 160: I had to look this up, but it's exactly according to spec:
image

https://think-async.com/Asio/boost_asio_1_30_2/doc/html/boost_asio/reference/io_context.html

mbr0wn (commented):
I did the same exercise for UHD recently (was much worse there...) and this is OK. The `cbegin()`+ dereference is actually something the ASIO folks do in their official examples (they use regular `begin()` but I'm not sure why, according to their own docs `cbegin()` does what you think it should).

I just think it should read `str_host` in that one spot.

mbr0wn (changes requested):
@landryb I found some issues. The missing `v4` identifiers will fix the `send_to` error.

mbr0wn:
@willcode FYI I also checked all of these changes are compatible with our minimum Boost (1.69), although the CI would have probably caught it.

mbr0wn:
> i see that the CI failed on a test:
> 
> ```
> 264/266 Test #261: qa_socket_pdu ................................***Failed    2.77 sec
> .thread_body_wrapper :error: ERROR thread[thread-per-block[1]: <block socket_pdu(6)>]: send_to: Address family not supported by protocol [system:97 at $PREFIX/include/boost/asio/detail/reactive_socket_service.hpp:261 in function 'send_to']
> ```
> 
> and that's where i throw my hands in the air, and leave someone more knowledgeable in the codebase to look at that :)

@landryb I posted what I think is the right fix for this problem, do you think you have the time/energy to push a fix for that?

landryb:
> @landryb I posted what I think is the right fix for this problem, do you think you have the time/energy to push a fix for that?

applied your suggestions, we'll see what CI thinks about it :)

landryb:
@mbr0wn tests still fail because there's no matching prototype, so [b0be9f1](https://github.com/gnuradio/gnuradio/pull/7554/commits/b0be9f114ead31c19b78089e52d77ea61d1e1465) is wrong

mbr0wn:
> > @landryb I posted what I think is the right fix for this problem, do you think you have the time/energy to push a fix for that?
> 
> applied your suggestions, we'll see what CI thinks about it :)
> 
> **edit** btw will have to triple check that those still please boost 1.87..

I was going to say that should be fine; like I said I did almost the same changes to UHD (and I saw the exact same error message when I forgot one of those `v4()` speficiers).

mbr0wn:
> @mbr0wn tests still fail because there's no matching prototype, so [b0be9f1](https://github.com/gnuradio/gnuradio/pull/7554/commits/b0be9f114ead31c19b78089e52d77ea61d1e1465) is wrong .. and i think its again my mistake

Hah yeah mixing up those tcp's and udp's is an easy mistake to make :)

No worries we're not going to hold it against you since you're doing all this work. When you have a passing CI run, please squash your commits into one before we can merge them.

mbr0wn:
@landryb Alright, CI is happy now!

mbr0wn (approved):

marcusmueller:
solves #7564

gdt:
This PR has been cherry-picked into pkgsrc, too.  Thanks @landryb for posting it.

From a packaging viewpoint, I think GNU Radio, once it decides to depend on boost, should ensure that at all times the most recent release of GNU Radio works with the most recent release of boost.  I realize that means tracking boost as it does things, ideally in CI with boost master, and having to have micro releases under time pressure, but packaging systems just update to releases as they **should** be fine.    I suspect that the real issue  here is that boost is not as stable as it should be, given the large footprint of  things that depend on it.

marcusmueller:
@gdt honestly, boost master is far too volatile to track; also, a git master is explicitly *not* what anyone but a project's developers should interact with: it's a development state, not something meant to be used by library consumers.

Yes, boost does occasionally have breaking changes, and we strive to notice them early on. But we can't maintain a special-purpose bleeding edge rolling release Linux distro; also, what would we learn if something breaks with an unreleased version of any library? They might fix that tomorrow, or we might have to fix it if they don't until the next release. But we'll have to wait for that – no way our medium-sized software project can do the stability and QA for boost.

gdt:
I didn't mean that releases should work with git master.  I just meant that as a packager, the latest gnu radio should work with the latest boost, always.   But this got fixed in GNU Radio master pretty fast, thanks.   It would be good to have a micro so the other packaging systems won't trip over it.

If what you're saying is "boost is too unstable to have as a dependency", then that makes sense, but it logically follows that you'd have a "stop using boost" ticket and be working on it.   I do realize that boost has a life of its own and they aren't going to become API-stable because you ask....

PR #7555, opened 2024-11-28
CI/conda: convert conda logging groups to github logging groups

marcusmueller:
## Description

Conda logs for a single run on Linux are 23 MB (!), which makes them impossible to inspect on the web.

But, a lot of CI failures are found exactly that way, so we should make these logs accessible, if we're already spending so much work and CI time on generating them.

## Related Issue

Tried to figure out what went wrong on a CI build, needed to download the log archive, because githubs web UI was helpless.

CI / conda

## Testing Done

this PR.

## Checklist

### DIFF
--- a/.github/workflows/conda-build.yml
+++ b/.github/workflows/conda-build.yml
@@ -68,7 +68,8 @@ jobs:
           export IS_PR_BUILD="False"
         fi
         echo "::endgroup::"
-        ./.scripts/run_docker_build.sh
+        # run, but convert conda log grouping to github grouping
+        ./.scripts/run_docker_build.sh | sed 's/.*##\[group\]\(.*\)/::group::\1/;s/.*##\[endgroup\].*/::endgroup::/'
 
     - name: Build on macOS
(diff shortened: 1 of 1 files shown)


marcusmueller:
Definitely an improvement.

ryanvolz (approved):
This won't persist through a re-rendering of the conda CI scripts with `conda smithy rerender`, but since I'm the one who does that I can put in a little effort to merge those changes in instead of completely overwriting `.github/workflows/conda-build.yml`.

If you think this would also be useful to other people who might be using conda-smithy-provided CI on Github Actions (which the conda-forge feedstocks themselves don't use, instead using Azure), and you feel up for submitting upstream, then the place to do that would be https://github.com/conda-forge/conda-smithy/blob/main/conda_smithy/templates/github-actions.yml.tmpl.

marcusmueller:
admittedly, was a bit of a hotfix!

marcusmueller:
@ryanvolz looked through that template and then realized that merging this is still a good idea

PR #7556, opened 2024-11-28
blocks: Reopen wavfile_sink after stop().

### LINKED ISSUE #7456: `wavfile_sink` does not continue after lock/unlock

sladecek:
## Description

The flowgraph uses the stop() method of a block 
to indicate that it has been locked for
configuration. I believe the correct behavior for
file-based sinks is to write all data and headers
to disk, flush the file buffers, but avoid closing the file 
(see also #2590).

This behavior is implemented in the
file_sink. Since the file_sink is very simple and
does not require updating headers, flushing the
file is sufficient.

On the other hand, wavfile_sink can write to
several file formats, some of which include
headers. The implementation relies on libsndfile,
which provides a flush function. However, the
source code indicates that this function does not
update headers before flushing.

The current implementation of wavfile_sink calls
the close() method within the stop() method to
ensure the file on disk is correct. However, this
also closes the file.

My solution is to add a flag for automatic
reopening. When the file is opened, its filename
is stored. When the stop() method is called, the
file is closed, and the automatic reopen flag is
set. Later, within the start() method the file is
automatically reopened in append mode.

## Related Issue
Fixes #7456 .

blocks/wavfile_sink

## Testing Done
- make test
- the reproduction steps from #7456

## Checklist

### DIFF
--- a/gr-blocks/lib/wavfile_sink_impl.cc
+++ b/gr-blocks/lib/wavfile_sink_impl.cc
@@ -99,133 +99,143 @@ wavfile_sink_impl::wavfile_sink_impl(const char* filename,
 
 bool wavfile_sink_impl::open(const char* filename)
 {
-    SF_INFO sfinfo;
-
     gr::thread::scoped_lock guard(d_mutex);
+    d_filename = filename;
 
     if (d_new_fp) { // if we've already got a new one open, close it
         sf_close(d_new_fp);
         d_new_fp = nullptr;
     }
 
-    if (d_append) {
-        // We are appending to an existing file, be extra careful here.
-        sfinfo.format = 0;
-        errno = 0;
-        if (!(d_new_fp = sf_open(filename, SFM_RDWR, &sfinfo))) {
-            if (errno) {
-                d_logger->error(
-                    "sf_open(1) failed: {:s}: {:s}", filename, strerror(errno));
-            } else {
-                d_logger->error(
-                    "sf_open(1) failed: {:s}: {:s}", filename, sf_strerror(NULL));
-            }
+    bool isOpen = d_append ? open_file_for_append() : open_file_for_rewrite();
+    return isOpen;
+}
 
-            return false;
-        }
-        if (d_h.sample_rate != sfinfo.samplerate || d_h.nchans != sfinfo.channels ||
-            d_h.format != (sfinfo.format & SF_FORMAT_TYPEMASK) ||
-            d_h.subformat != (sfinfo.format & SF_FORMAT_SUBMASK)) {
-            d_logger->error("Existing WAV file is incompatible with configured options.");
-            sf_close(d_new_fp);
-            return false;
-        }
-        if (sf_seek(d_new_fp, 0, SEEK_END) == -1) {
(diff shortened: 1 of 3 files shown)


gr-blocks/lib/wavfile_sink_impl.cc
```diff
@@ -255,6 +265,20 @@ bool wavfile_sink_impl::stop()
...
+}
+
+bool wavfile_sink_impl::start()
+{
+    gr::thread::scoped_lock guard(d_mutex);
+    if (d_should_reopen) {
+        d_should_reopen = false;
+        if (!open_file_for_append()) {
+            d_logger->error("Failed to reopen output file after reconfiguration.");
+        }
```
marcusmueller, line 280: Can the sink continue to work in this situation?
If not, we should `return false` here!

marcusmueller (commented):
Needs one clarification; other than that, looks

gr-blocks/lib/wavfile_sink_impl.cc
sladecek: I agree, in this situation is more correct to return `false`.

marcusmueller (approved):

PR #7557, opened 2024-11-29
Refactor threads.cc

marcusmueller:
## Description

#7553 showed that the current, very coarse "we copy all code for 1. Linux, 2. Windows, 3. Mac OS (and all the BSDs, they're the same, right?)" doesn't reflect reality well.

Windows is different between MINGW and MSVC, and not all BSDs are MacOS. So, instead of large redundant blocks, pull the platform decision into the functions, and decide as necessary.

Also, add logging. If the user tried to pin a thread to a CPU and we couldn't do that, we might as well tell them…

Hopefully none functionally; I hope that future porting efforts become easier that way (pretty sure MacOS does support a lot of the functionality that we don't implement on that platform, for example).

## Testing Done

ran test suite on Linux. 

## Checklist

### DIFF
--- a/gnuradio-runtime/lib/thread/thread.cc
+++ b/gnuradio-runtime/lib/thread/thread.cc
@@ -8,257 +8,135 @@
  *
  */
 
+#include <gnuradio/logger.h>
 #include <gnuradio/thread/thread.h>
+#include <spdlog/fmt/fmt.h>
+#include <stdexcept>
 
+/* Platform detection & Platform-specific includes */
 #if defined(_WIN32) || defined(__WIN32__) || defined(WIN32)
-
+#define __GR_TARGET_WIN__
 #include <windows.h>
-
-namespace gr {
-namespace thread {
-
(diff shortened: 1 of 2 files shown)


marcusmueller:
@landryb this incorporates your OpenBSD fix, but it would be very appreciated if you could test this, and give feedback.

marcusmueller:
in fact, added setting of thread names on MacOS; just don't have a device to test whether it makes user life better.

landryb:
> @landryb this incorporates your OpenBSD fix, but it would be very appreciated if you could test this, and give feedback.

i can mostly only build-test things by backporting them on 3.10.11 release (which builds within our ports infra, and sets up the configure args/dependencies/etc), so if that's enough for you.. unless ofc you point me at an unit test that would exercise this codepath.

landryb:
fwiw, i had mostly forgot about it, but the PR manually backported on top of 3.10.11 builds fine on OpenBSD.

PR #7561, opened 2024-12-07
Fix #4196 (Python blocks with configurable IO vector size)

### LINKED ISSUE #4196: Embedded Python Block - Cannot make variable input/output length
Background: I'm using GNURadio 8.2.0 on Ubuntu 18.04

I'm trying to create an Embedded Python Block in my GRC flowgraph. I want a variable to control the vector size of the input and output (`vlen` in my example). I whipped up the following code, but **GRC always uses the default of `vlen` as the vector size** of the input and output (1 in my example). I've tried a couple of combinations, but no luck. Any thoughts?

```python
class blk(gr.sync_block):
    def __init__(self, vlen=1):
        gr.sync_block.__init__(
            self,
            name='Simple Example',
            in_sig=[(np.float32, vlen), (np.float32, 1)],
            out_sig=[(np.float32, vlen)]
        )
```

j0hnnybash:
## Description

As discussed in #4196 and as noted on the [wiki](https://wiki.gnuradio.org/index.php?title=Python_Block_with_Vectors#Warning_for_Vector_Length_Mismatches) Embedded Python block IO signature could not depend 

## Related Issue

Fixes #4196.

Embedded Python blocks

## Testing Done

Manual Testing, also extended the epy_block_io.py  example/self-test (`python -m utils.epy_block_io`).

## Checklist


marcusmueller:
Hi @j0hnnybash, nice to have you around: Heads up, we code-format our python with pycodestyle (I don't think that's a great tool, either, but going for something else would cause a lot of friction). No matter which code formatter we use, in your dictionary comprehension, there's a space missing after the `:`; could you fix that?

There's also a few non-canonical empty lines and line breaks, see the errors in the "Check Python formatting" check above :)

marcusmueller:
other than that, your commit messages are very nice! The subject line is too long, though. I'd just replace

    grc/core/blocks/embedded_python: factor out Python Block parameter ma…

with 

    grc: factor out embedded Python Block parameter marking logic

(the precise file location is in the commit content anyways)

j0hnnybash:
- [ ] TODO: remove the section [Warning for Vector Length Mismatches](https://wiki.gnuradio.org/index.php?title=Python_Block_with_Vectors#Warning_for_Vector_Length_Mismatches) from the "Python Block with Vectors" tutorial.

marcusmueller:
Hey, if you need people to try this code, feel encouraged to say so (here, or on matrix in #Development ). Cheers!

j0hnnybash:
> Hey, if you need people to try this code, feel encouraged to say so (here, or on matrix in #Development ). Cheers!

Thanks, I was just trying to come up with a sensible unit test, however I think a simple example `.grc` example file is probably more helpful for trying out the changes. I've attached a simple example with a python block where the configured `vlen` parameter differs from its default: [epy_block_io_example.grc.txt](https://github.com/user-attachments/files/18063452/epy_block_io_example.grc.txt).

The changes are fully working, however sometimes the following exception is thrown when changing the vector length (but the issues fixes itself...)
```
  File "/home/jlrb/Documents/funland/gnuradio_playground/gnuradio/grc/gui/DrawingArea.py", line 251, in draw
    self._flow_graph.draw(cr)
  File "/home/jlrb/Documents/funland/gnuradio_playground/gnuradio/grc/gui/canvas/flowgraph.py", line 542, in draw
    draw_element(cr)
  File "/home/jlrb/Documents/funland/gnuradio_playground/gnuradio/grc/gui/canvas/block.py", line 280, in draw
    port.draw(cr)
  File "/home/jlrb/Documents/funland/gnuradio_playground/gnuradio/grc/gui/canvas/port.py", line 134, in draw
    cr.rectangle(*self._area)
```
I suspect this is the result of a race condition between the flow graph being redrawn and `create_shapes()` for the newly created IO ports. I suspect that this could already be triggered previously by simply changing the IO signature in the python source code, but I'm haven't managed to do that yet, so maybe it's specific to my changes.

TLDR: please feel free to try out the example :)

marcusmueller:
hm such a condition is of course undesirable.

j0hnnybash:
I just added a fix for the race condition issue, with some discussion of alternatives in the commit message.
The fix works perfectly well, but feels a bit like a band-aid and not a proper solution, since from a quick glance other `draw()` functions seem to also expect `create_shapes()` etc. to have been called at least once, but this does not create any issues. So maybe this isn't a race condition after all?

Maybe someone familiar with the GUI model could chime in and explain what the assumptions are for the "initial" drawing of newly created elements? Are there any aspects of drawing and flowgraph modification that can happen concurrently?

Since there aren't more of these "race conditions" I assume that `FlowGraph.update()`  is guaranteed to not run concurrently with `FlowGraph.draw()`, and that the typical flow after the graph was modified should be:

```python
 FlowGraph.update() # called in response to modifications
     rewrite()  # new elements created, here or potentially earlier
     validate()
     update_elements_to_draw()  # this adds new elements to the "drawlist"
     create_labels()
     create_shapes()  # this makes elements ready to be drawn
  FlowGraph.draw() # elements from the "drawlist" are finally drawn
```
For some reason I observe that `Port.draw()` can sometimes be called before `Port.create_shapes()`, which contradicts the above flow.

Are their conditions that elements can be drawn, without/prior to being added to the "drawlist" i.e. before `FlowGraph.update()` has run?

marcusmueller:
Pulling in @haakov on that question

j0hnnybash:
I've done some additional debugging and identified the reason why, and where, `create_shapes()` **isn't** called:

The `PropsDialog` class contains the following handler that is triggered when a parameter change is "applied" (e.g. by switching to another parameter field, by hitting enter or when the dialog window loses focus, not only by pressing the "Apply"/"OK" button!):

```python
    def _handle_changed(self, *args):
        """
        A change occurred within a param:
        Rewrite/validate the block and update the gui.
        """
        self._block.rewrite()
        self._block.validate()
        self.update_gui()
```

which calls `rewrite()` and `validate()` in order to update the error messages displayed at the bottom of the `PropsDialog`.
`create_shapes()` and `create_labels()` will only be called on the next full "FlowGraph.update()" which will happen when the "Apply" or "OK" button is pressed. If however a "draw" operation is triggered at any time before the "Apply"/"OK" button can be pressed, e.g. by moving the mouse pointer over to the main grc window, the bug/race will be triggered.

This also explains why this could never happen before, for example when changing a python blocks IO signature through an external editor and then saving the change: the [code path](https://github.com/gnuradio/gnuradio/blob/6fd698529be476acc69bb8c1d49db703863bba03/grc/gui/Application.py#L602C9-L607C31) handling external edits calls `flow_graph_update()` directly.

While my original fix still works, it would also be possible to add calls to `create_shapes()` and `create_labels()` in the `_handle_changed` function of `PropsDialog` or maybe instead add a call to `flow_graph_update()`/`FlowGraph.update()`.  It might be important to mention that the invocation of `rewrite()` in `_handle_changed()` is the only invocation of a `rewrite()` function within the GTK GUI tree (`grc/gui/...`) besides the one in `FlowGraph.update()`, i.e. for GTK all other GUI triggered rewrites appear to go through `FlowGraph.update()`.

j0hnnybash:
I've also now tested the QT Gui, starting from an empty `.grc` creating a flow graph as in https://github.com/gnuradio/gnuradio/pull/7561#issuecomment-2528323604 and changing block parameters works perfectly well without triggering an exception like in the GTK branch. (this does trigger a UI bug: old connection arrows become disconnected, but don't disappear after changing the vector length, however the same happens when changing the blocks signature directly in the blocks source code so that's probably a preexisting bug/limitation)

However when loading a `.grc` file with an embedded python block the QT UI crashes with
```
Running from source tree
  File "/home/jlrb/Documents/funland/gnuradio_playground/gnuradio/grc/gui_qt/components/canvas/port.py", line 137, in paint
    block_rotation = self.parentItem().rotation()
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'rotation'
Aborted (core dumped)
```

I only had a short glance at `gui_qt/components/canvas/port.py` but the issue seems to be somewhat related to the GTK issue, in that `GUIPort.create_shapes_and_labels()` contains the following:

```
...
if not self.parentItem():
            self.setParentItem(self.core.parent_block.gui)
...
```
This would be similarly easy to fix by simply skipping the `paint()` if the parent isn't yet set, at the same time I don't see a reason why `setParentItem()` cannot already be called by the `GUIPort` constructor (currently this is apparently done for dummy blocks). Maybe someone more familiar with the QT UI (@haakov?) could provide some guidance/background?

j0hnnybash:
@marcusmueller I've come up with a new fix for the "race condition" issue for the GTK GUI (by looking at the equivalent code in the QT GUI): [8a66811](https://github.com/gnuradio/gnuradio/pull/7561/commits/8a66811024fe585f6ae7e964ae2dc2621c1757cf)
This now makes the original fix [a685fe7](https://github.com/gnuradio/gnuradio/pull/7561/commits/a685fe7a2ed2e29d0c337d2edc58b18775dcac66) technically redundant, should I simply remove it, leave it as is, or maybe add a log/debug output?

With regards to the crash for the QT GUI I mentioned in https://github.com/gnuradio/gnuradio/pull/7561#issuecomment-2539355920, commit [d99c0a5](https://github.com/gnuradio/gnuradio/pull/7561/commits/d99c0a588e394c812f76323b719a9ccaeba15310) adds a fix to avoid the crash. However this only covers up the underlying issue, which is a essentially a GUI element "leak" or "zombie": the GUI elements for the ports of python blocks (`GUIPort`) will never be deleted/removed from the QT GUI scene even if the underlying "core" flowgraph `Port` is removed from its block. This issue/leak also happens without my changes to the IO extraction however it won't result in a crash. With my change to the IO extraction a partially initialized `GUIPort` is leaked, while previously all leaked/zombie `GUIPorts` were fully initialized and look like this:
2024-12-16_16:15:11_screenshot
the leak can be observed by first adding a bunch of ports to an embedded python block (by editing its source code),
2024-12-16_16:15:29_screenshot
 and then removing them again.

Fixing the QT GUI issue properly appears to be a lot more involved  (see https://github.com/j0hnnybash/gnuradio/commit/0f33fd3f57edb71802abf2bf12c81a085bc2e06e for an ugly fix), so I'd hope that maybe we could leave the simple fix for now and work on the ground cause in a separate issue/PR.

Besides the QT GUI issue and any further comments from your side notwithstanding I think that this PR is complete (and hopefully mergeable?) :)

cc @haakov , @skoslowski

marcusmueller:
> This now makes the original fix [a685fe7](https://github.com/gnuradio/gnuradio/pull/7561/commits/a685fe7a2ed2e29d0c337d2edc58b18775dcac66) technically redundant, should I simply remove it, leave it as is, or maybe add a log/debug output?

I'd say both removing or adding/converting it to a log entry are good options; I'd leave this up to you: It's an improvement over the status quo either way, far as I can tell. This has been an awesome PR so far!

>  so I'd hope that maybe we could leave the simple fix for now and work on the ground cause in a separate issue/PR.

Yes, please.

haakov:
Thank you so much for taking the time to dig into this!
I agree with @marcusmueller's last comment. 

I've noticed there are several questions in this thread, and I'm not entirely sure which have been answered already, so maybe we can create a new issue titled "Undesirable GRC Quirks" or something? We can track them there with checkboxes so they don't disappear when this PR is merged

grc/core/blocks/embedded_python.py
```diff
@@ -87,7 +87,6 @@ def __init__(self, flow_graph, **kwargs):
         self.states['_io_cache'] = ''
 
         self.module_name = self.name
-        self._epy_source_hash = -1
```
haakov, line 90: I see you're removing the hash entirely, are you planning on bringing it back in a later PR?

j0hnnybash: No; unless you have a specific use for it in mind that I'm missing :). 

(As far as I can tell) before the change to the IO extraction, the source hash mechanism was strictly an optimization to avoid a redundant block instantiation, extraction and port update; without affecting any observable behavior (though I don't believe I have tested this in isolation...). The only use of the `self._epy_source_hash` variable in the source code was to return early from the `rewrite()` function if the hash of the embedded python source code had not change. Since I had to remove this early return in order to allow for the second (final) instantiation/extraction of the python block with the final parameter values the source hash and the code calculating it were now dead code, thus I removed them. (see efa9a3f955cd87333ce926f41c5e9f2c95c40762)

*Pedantic side note:* I believe that the hash check in itself for the purpose of detecting a **lack of change** was technically invalid in the first place: given that the hash function used for Python strings isn't a cryptographic hash function it would typically only be used to quickly **distinguish** objects (matching hashes would result in a full comparison in a hash table). Admittedly it would have been very unlikely to get a hash collision by accident, but that would have been quite the Heisenbug :smiley:.

j0hnnybash:
> I'd say both removing or adding/converting it to a log entry are good options; I'd leave this up to you: It's an improvement over the status quo either way, far as I can tell. This has been an awesome PR so far!

I've thought about it: I'll remove the original fix (https://github.com/gnuradio/gnuradio/pull/7561/commits/a685fe7a2ed2e29d0c337d2edc58b18775dcac66), but add a comment to the `Port._area` variable in the constructor, that it will be initialized by `create_shapes()` in order to somewhat document the convention that `draw()` expects (and rightfully so) `create_shapes()` to be called first.

[ ] TODO remove a685fe7a2ed2e29d0c337d2edc58b18775dcac66 and add a comment on `Port._area`

j0hnnybash:
> I've noticed there are several questions in this thread, and I'm not entirely sure which have been answered already, so maybe we can create a new issue titled "Undesirable GRC Quirks" or something? We can track them there with checkboxes so they don't disappear when this PR is merged

Good idea, will do.

 - [ ] TODO collect various quirks and open questions in separate issue (QT GUIConnections staying arround, QT+GTK connections being removed when changing vector length of connected python block port, ...?) 
  - maybe also collect answered questions? (e.g. `draw()` is not called concurrently with other functions)
 - [ ] TODO open issue regarding QT GUIPort "leak" (see https://github.com/gnuradio/gnuradio/pull/7561#issuecomment-2545951263)

grc/gui/canvas/port.py
```diff
@@ -33,7 +33,7 @@ def __init__(self, parent, direction, **n):
         self._hovering = False
         self.force_show_label = False
 
-        self._area = []
+        self._area = [] # initialized by create_shapes(), must be initialized before calling draw()!
```
haakov, line 36: PEP8 error: _./grc/gui/canvas/port.py:36:24: E261 at least two spaces before inline comment_

haakov (changes requested):
Sorry for taking so long with this. @j0hnnybash I think this is ready for merge, but would you mind removing cd2f226f392f30ba130a224aff2de4b9af7a45d8 and a685fe7a2ed2e29d0c337d2edc58b18775dcac66 (The commit that has been reverted) first?

Also, can you add another space before your comment in port.py so the formatting check passes?

grc/gui/canvas/port.py
marcusmueller: fixed over in #7843

marcusmueller:
@haakov I've cleaned the PR in #7843. Please have a look at it, add it to the Backports-3.10 project if you think it should be backported, and merge it if OK!

marcusmueller:
I'm archiving this PR here, I think @j0hnnybash really did great work, and we should move this on in #7843.

PR #7565, opened 2024-12-16
Fix image ratio in readme

777arc:
(before change) images look stretched

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@
 
 # GNU Radio
 
-<table>
+<table align="center">
     <tr>
         <td>
             <img src="./docs/grc_eg.png" height="200px" />
--- a/README.md
+++ b/README.md
@@ -11,16 +11,16 @@
 
 # GNU Radio
 
-<table align="center">
+<table>
     <tr>
         <td>
-            <img src="./docs/grc_eg.png" height="300px" />
+            <img src="./docs/grc_eg.png" height="200px" />
         </td>
         <td>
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -14,13 +14,13 @@
 <table align="center">
     <tr>
         <td>
-            <img src="./docs/grc_eg.png" height="300px" />
+            <img src="./docs/grc_eg.png" height="200px" />
```
marcusmueller, line 17: what about specifying the width? Something in github's CSS must be force-scaling things – normally in HTML, when you specify one dimension, the other gets chosen to keep the aspect ratio correct. Here, this isn't happening, so I think it would be prudent to specify both dimensions explicitly in order to not depend on the "external" style squashing the images as it sees fit on whatever screen resolution the user has.

777arc:
Yeah I played around with other styling and it looks like github only honors a small fraction of this stuff, it has their own style for most things, even stuff like padding.

777arc:
feel free to merge

marcusmueller:
will go for it. However, where do all these unrelated broken CI runs come from? :fearful:

777arc:
Yeah I started to dive into the osx one, since it looks like a straightforward error `The following argument was not expected: --update-specs` although that part of the code hasnt changed in many months so it's probably some version got bumped behind the scenes

marcusmueller:
ah darn, we should have squashed these two commits. No big loss, but it'd still have been nicer. That's what happens when one doesn't do maintenance for longer :grimacing:

PR #7597, opened 2024-12-27
gr-blocks: Add pdu support to mute and selector blocks 

### LINKED ISSUE #5099: Blocks with raw PMT input values
Almost all blocks which have PMT message inputs use the name:value format. For example: https://wiki.gnuradio.org/index.php/QT_GUI_Digital_Number_Control

However, there are two which use raw values instead of the name:value format. These are:
- https://wiki.gnuradio.org/index.php/Mute
- https://wiki.gnuradio.org/index.php/Selector

I recommend that the code accept either name:value pairs or raw input, so that existing code still works and the conventional name:value pairs work as well.

the-gremlin:
## Description

In this pull request the message handlers of the blocks are modified so they can accept both raw-value messages and PDUs.
 It also modifies the code in the work function of the mute block to hopefully make it more readable and adds tests for floating point numbers to the qa code for the mute block.

## Related Issue

Fixes #5099.

Affects the mute block and selector blocks. The _enable_ port of the selector block and the _set-mute_ port of the mute block can now receive both raw pmt messages and PDU messsages.

## Testing Done

Manual

## Checklist

### DIFF
--- a/gr-blocks/lib/mute_impl.cc
+++ b/gr-blocks/lib/mute_impl.cc
@@ -1,6 +1,6 @@
 /* -*- c++ -*- */
 /*
- * Copyright 2004,2010,2013,2018,2019 Free Software Foundation, Inc.
+ * Copyright 2004,2010,2013,2018,2019,2024 Free Software Foundation, Inc.
  *
  * This file is part of GNU Radio
  *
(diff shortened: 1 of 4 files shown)


the-gremlin:
I'm closing the pull request because I misunderstood how the DCO signing works, I will make a new pull request with the correct sign-offs.

PR #7598, opened 2024-12-27
gr-blocks: Add PDU support to mute and selector blocks

### LINKED ISSUE #5099: Blocks with raw PMT input values

the-gremlin:
## Description

 Changed the message handlers of the mute and selector blocks so they can receive pdu messages.
Also added a qa test for floats for the mute block and simplified the work function of the mute block
in an attempt to make it more readable.

## Related Issue

Fixes #5099 

The _set-mute_ port of the mute block and the _en_ port of the selector block now 
can receive both raw PMT values and PDUs as messages.
The work function of the mute block was simplified as the old version was not easy to read,
specifically the code that handled forwarding the input values.

## Testing Done

Manual

## Checklist

### DIFF
--- a/gr-blocks/lib/mute_impl.cc
+++ b/gr-blocks/lib/mute_impl.cc
@@ -1,6 +1,7 @@
 /* -*- c++ -*- */
 /*
  * Copyright 2004,2010,2013,2018,2019 Free Software Foundation, Inc.
+ * Copyright 2024 Skandalis Georgios
  *
  * This file is part of GNU Radio
  *
@@ -9,6 +10,7 @@
  */
 
 
+#include <pmt/pmt.h>
 #ifdef HAVE_CONFIG_H
 #include "config.h"
 #endif
@@ -35,8 +37,9 @@ mute_impl<T>::mute_impl(bool mute)
       d_mute(mute)
 {
     this->message_port_register_in(pmt::intern("set_mute"));
-    this->set_msg_handler(pmt::intern("set_mute"),
-                          [this](pmt::pmt_t msg) { this->set_mute_pmt(msg); });
+    this->set_msg_handler(pmt::intern("set_mute"), [this](const pmt::pmt_t& msg) {
+        this->mute_message_handler(msg);
+    });
 }
 
 template <class T>
@@ -45,32 +48,27 @@ mute_impl<T>::~mute_impl()
 }
 
 template <class T>
+void mute_impl<T>::mute_message_handler(const pmt::pmt_t& msg)
+{
+    if (pmt::is_pair(msg)) {
+        this->set_mute(pmt::to_bool(pmt::cdr(msg)));
+    } else {
+        this->set_mute(pmt::to_bool(msg));
+    }
+}
+
+template <class T>
 int mute_impl<T>::work(int noutput_items,
                        gr_vector_const_void_star& input_items,
                        gr_vector_void_star& output_items)
 {
-    const T* iptr = (const T*)input_items[0];
-    T* optr = (T*)output_items[0];
-
-    int size = noutput_items;
+    const T* in = (const T*)input_items[0];
+    T* out = (T*)output_items[0];
 
     if (d_mute) {
-        std::fill_n(optr, noutput_items, 0);
+        std::fill_n(out, noutput_items, 0);
     } else {
-        while (size >= 8) {
-            *optr++ = *iptr++;
-            *optr++ = *iptr++;
-            *optr++ = *iptr++;
-            *optr++ = *iptr++;
-            *optr++ = *iptr++;
-            *optr++ = *iptr++;
-            *optr++ = *iptr++;
-            *optr++ = *iptr++;
-            size -= 8;
-        }
-
-        while (size-- > 0)
-            *optr++ = *iptr++;
+        std::copy_n(in, noutput_items, out);
     }
 
     return noutput_items;
--- a/gr-blocks/lib/mute_impl.h
+++ b/gr-blocks/lib/mute_impl.h
@@ -1,6 +1,7 @@
 /* -*- c++ -*- */
 /*
  * Copyright 2004,2013,2018 Free Software Foundation, Inc.
+ * Copyright 2024 Skandalis Georgios
  *
  * This file is part of GNU Radio
  *
@@ -13,6 +14,7 @@
 #define MUTE_IMPL_H
 
 #include <gnuradio/blocks/mute.h>
+#include <pmt/pmt.h>
 
 namespace gr {
 namespace blocks {
@@ -29,7 +31,7 @@ public:
 
     bool mute() const override { return d_mute; }
     void set_mute(bool mute) override { d_mute = mute; }
-    void set_mute_pmt(pmt::pmt_t msg) { set_mute(pmt::to_bool(msg)); }
+    void mute_message_handler(const pmt::pmt_t& msg);
 
     int work(int noutput_items,
              gr_vector_const_void_star& input_items,
--- a/gr-blocks/lib/selector_impl.cc
+++ b/gr-blocks/lib/selector_impl.cc
@@ -1,6 +1,7 @@
 /* -*- c++ -*- */
 /*
  * Copyright 2019 Free Software Foundation, Inc.
+ * Copyright 2024 Skandalis Georgios
  *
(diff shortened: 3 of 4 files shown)


gr-blocks/lib/mute_impl.cc
```diff
@@ -1,6 +1,6 @@
 /* -*- c++ -*- */
 /*
- * Copyright 2004,2010,2013,2018,2019 Free Software Foundation, Inc.
+ * Copyright 2004,2010,2013,2018,2019,2024 Free Software Foundation, Inc.
```
marcusmueller, line 3: ah, while nice, we don't do copyright assignment to the FSF anymore – instead, you retain the copyright to your contributions :-) 

That means you'll want to do something like
```suggestion
 * Copyright 2004,2010,2013,2018,2019 Free Software Foundation, Inc.
 * Copyright 2024 The Gremlin (or their real name)
```

gr-blocks/lib/mute_impl.cc
```diff
@@ -36,41 +37,36 @@ mute_impl<T>::mute_impl(bool mute)
 {
     this->message_port_register_in(pmt::intern("set_mute"));
     this->set_msg_handler(pmt::intern("set_mute"),
-                          [this](pmt::pmt_t msg) { this->set_mute_pmt(msg); });
+                          [this](pmt::pmt_t msg) { this->mute_message_handler(msg); });
```
marcusmueller, line 40: can we make the signature of this `const pmt::pmt& msg` rather than `pmt::pmt_t msg`?

Background is that you're really just "passing through" values, but pass-by-value on `pmt_t` is actually a reference count increase, and that comes with mutex overhead!

(Technical background 2: `pmt_t` is actually a `std::shared_ptr<pmt_base>`, which means it does a lot of cross-thread synchronized book-keeping behind the scenes when it's copied and when a copy is destroyed).

gr-blocks/lib/mute_impl.cc
```diff
@@ -36,41 +37,36 @@ mute_impl<T>::mute_impl(bool mute)
...
 }
 
 template <class T>
 mute_impl<T>::~mute_impl()
 {
 }
 
+template <class T>
+void mute_impl<T>::mute_message_handler(pmt::pmt_t msg)
```
marcusmueller, line 49: see above, `const pmt::pmt_t&`

gr-blocks/lib/mute_impl.cc
```diff
@@ -36,41 +37,36 @@ mute_impl<T>::mute_impl(bool mute)
...
-            *optr++ = *iptr++;
-            *optr++ = *iptr++;
-            *optr++ = *iptr++;
-            *optr++ = *iptr++;
-            size -= 8;
-        }
-
-        while (size-- > 0)
-            *optr++ = *iptr++;
+        std::copy_n(in, noutput_items, out);
```
marcusmueller, line 69: *much* nicer. Let the compiler vectorize. Manual loop unrolling is probably a very bad idea. :+1:

gr-blocks/lib/mute_impl.h
```diff
@@ -1,6 +1,6 @@
 /* -*- c++ -*- */
 /*
- * Copyright 2004,2013,2018 Free Software Foundation, Inc.
+ * Copyright 2004,2013,2018,2024 Free Software Foundation, Inc.
```
marcusmueller, line 3: see above

gr-blocks/lib/mute_impl.h
```diff
@@ -29,7 +30,7 @@ class mute_impl : public mute_blk<T>
 
     bool mute() const override { return d_mute; }
     void set_mute(bool mute) override { d_mute = mute; }
-    void set_mute_pmt(pmt::pmt_t msg) { set_mute(pmt::to_bool(msg)); }
+    void mute_message_handler(pmt::pmt_t msg);
```
marcusmueller, line 33: see above

gr-blocks/lib/selector_impl.cc
```diff
@@ -1,13 +1,14 @@
 /* -*- c++ -*- */
 /*
- * Copyright 2019 Free Software Foundation, Inc.
+ * Copyright 2019,2024 Free Software Foundation, Inc.
```
marcusmueller, line 3: see above

gr-blocks/lib/selector_impl.cc
```diff
@@ -122,11 +123,16 @@ void selector_impl::handle_msg_output_index(const pmt::pmt_t& msg)
...
+        gr::thread::scoped_lock l(d_mutex);
+        d_enabled = en;
+    } else if (pmt::is_pair(msg)) {
+        const bool en = pmt::to_bool(pmt::cdr(msg));
         gr::thread::scoped_lock l(d_mutex);
         d_enabled = en;
     } else {
-        d_logger->warn("handle_enable: Non-PMT type received, expecting Boolean PMT");
+        d_logger->warn(
+            "handle_enable: Non-PMT type received, expecting Boolean PMT or PDU");
```
marcusmueller, line 135: The message is not yours, but it's still wrong (I mean, you can *only* send PMTs around, what they meant to say is that it's an *invalid* PMT message). If you feel like it, you might just fix that :)

gr-blocks/python/blocks/qa_mute.py
```diff
@@ -74,6 +75,18 @@ def test_mute_cc(self):
...
+        expected_result = [1.5, 2.5, 3.5, 4.5, 5.5]
+        op = blocks.mute_ff(False)
+        self.help_ff((src_data,), expected_result, op)
+
+    def test_mute_ff(self):
+        src_data = [1.5, 2.5, 3.5, 4.5, 5.5]
+        expected_result = [0.0, 0.0, 0.0, 0.0, 0.0]
+        op = blocks.mute_ff(True)
+        self.help_ff((src_data,), expected_result, op)
+
```
marcusmueller, line 89: Can you come up with a unit test that actually test the PMT handling abilities?

I know that's super annoying to do *reliably* with asynchronous message passing and synchronous stream that might be over very quickly, but that's why I extended the throttle block to be more useful for tests:

flowgraph would be something like

source (50 constants, e.g. 1.0) -> throttle (rate = 200 per second; max items per chunk: 2) -> mute -> sink

Then you `tb.start(); sleep(0.1); send_message_to_mute(); tb.wait()` and check whether in the data there's 1.0 until some point after the 20th item and from that first zero on only zeros till the end?

marcusmueller (changes requested):
Thank you very much! This is very good. Can you address the copyright issue, and the testing? The zero-copy handling of PMTs is desirable, but not as pressing.

gr-blocks/python/blocks/qa_mute.py
the-gremlin: I tried to add tests for the PMT's but I couldn't figure out this exact issue. Thank you, I'll add them now.

marcusmueller: no, thank you, your work is awesome!

the-gremlin:
Hello and happy new year!!! 

I tried my best to implement sending the messages while the flowgraph was running, however I couldn't get it to work. I hope this way of implementing the tests is acceptable.

marcusmueller (approved):

marcusmueller:
A happy new year to you, too!

The tests are fine, they actually excite the code path you've added :+1:

marcusmueller:
Do you think you could squash these commits into 1 commit for merging?

the-gremlin:
> Do you think you could squash these commits into 1 commit for merging?

Here you go, I assumed you meant the most recent commits since aside the tests the rest were pretty minor. But tell me if you want me to squash all of the commits in the PR into one instead.

marcusmueller:
We'd prefer a single commit, actually, because that ensures that code + code covered by unit tests are always in lockstep during `git bisect`.

PR #7613, opened 2024-12-30
GRC Qt: Prevent redundant undo actions

### LINKED ISSUE #7611: GRC Qt Move actions are duplicated
### What happened?

When moving a block in GRC Qt, two actions are added to the undo stack. In other words, the user has to hit Ctrl+Z twice to undo the move. It looks like the first action contains the entire move, and the second one is empty. **However**, this only happens if you deselect the block after moving it.

### Steps to reproduce

- Move a block
- Click anywhere in the canvas
- Hit Ctrl+Z once (nothing happens)

### Where to start

Place a breakpoint here:
https://github.com/gnuradio/gnuradio/blob/76900adac6ab09ba3344412589fd590cf02a8ed6/grc/gui_qt/components/undoable_actions.py#L62-L64

Sahil-u07:
## Description
Fixes the issue where two undo actions were added when moving blocks in GRC Qt. This occurred because an empty undo action was being added if the block was deselected after moving it.

## Related Issue
Fixes #7611  

- Affects the **MoveAction** functionality in the **undoable_actions.py** file.
- This fix affects the undo/redo behavior in the GRC Qt interface, specifically preventing redundant actions from being added to the undo stack when blocks are moved.

## Testing Done
- Tested by moving blocks in the GRC Qt interface, deselecting the blocks, and using Ctrl+Z to undo the action.
- Verified that only one undo action is added, and that it correctly reverts the block movement.

## Checklist

- [x] I [have signed my commits before making this PR](https://github.com/gnuradio/gnuradio/blob/main/CONTRIBUTING.md#dco-signed).

### DIFF
--- a/grc/gui_qt/components/undoable_actions.py
+++ b/grc/gui_qt/components/undoable_actions.py
@@ -61,6 +61,9 @@ class RotateAction(QUndoCommand):
 
 class MoveAction(QUndoCommand):
     def __init__(self, scene: FlowgraphScene, diff: QPointF):
+        # Prevent empty undo actions if the block hasn't moved
+        if diff.isNull():
+            return
         QUndoCommand.__init__(self)
         log.debug("init MoveAction")
         self.setText('Move')
@@ -87,6 +90,7 @@ class MoveAction(QUndoCommand):
         for g_block in self.g_blocks:
             g_block.move(-self.x, -self.y)
         self.scene.update()
+
 
 
 class EnableAction(ChangeStateAction):
--- a/grc/gui_qt/components/undoable_actions.py
+++ b/grc/gui_qt/components/undoable_actions.py
@@ -92,7 +92,6 @@ class MoveAction(QUndoCommand):
         self.scene.update()
 
 
-
 class EnableAction(ChangeStateAction):
     def __init__(self, scene: FlowgraphScene):
         ChangeStateAction.__init__(self, scene)
--- a/grc/gui_qt/components/canvas/flowgraph.py
+++ b/grc/gui_qt/components/canvas/flowgraph.py
@@ -341,6 +341,7 @@ class FlowgraphScene(QtWidgets.QGraphicsScene, base.Component):
                 self.start_port = None
                 self.end_port = None
         self.mousePressed = False
+        self.moving_blocks = False
         super(FlowgraphScene, self).mouseReleaseEvent(event)
 
     def createActions(self, actions):


Sahil-u07:
@ThomasHabets  please review the changes. The DCO check has passed and the commit has been amended.

Sahil-u07:
@haakov please review the changes. The DCO check has passed and the commit has been amended.

grc/gui_qt/components/undoable_actions.py
```diff
@@ -89,6 +92,7 @@ def undo(self):
         self.scene.update()
 
 
+
```
marcusmueller, line 95: Please avoid adding unrelated line additions. CI might complain about an extra empty line, anyways

marcusmueller (changes requested):

marcusmueller:
Please fix your commit message; ours are generally formatted:

```text
area: summary

More extensive description
```

In your case, something like 
```text
GRC Qt: Prevent redundant undo actions in MoveAction

It's not clear where the empty move action comes from, but we should not insert empty actions into the undo history in any case.
```

haakov:
@Sahil-u07 Thank you for taking this on!

First of all, your fix works, so that's great :)
But we shouldn't really have to handle Move actions with zero diffs, so I'd love to see a change in this pull request that fixes the root cause of this issue.

I tried placing a breakpoint here
https://github.com/gnuradio/gnuradio/blob/d6badf773fb98e973f274b1cb0dcbcc005e4d748/grc/gui_qt/components/undoable_actions.py#L64

and reproducing the bug. It looks like an `itemMoved` signal is emitted here, from `FlowgraphScene::mouseReleaseEvent()`

https://github.com/gnuradio/gnuradio/blob/d6badf773fb98e973f274b1cb0dcbcc005e4d748/grc/gui_qt/components/canvas/flowgraph.py#L333-L334

I believe `self.moving_blocks` should only be True _while_ we're actually holding the mouse button down and moving a block. Maybe adding `self.moving_blocks = False` between the following lines would help?

https://github.com/gnuradio/gnuradio/blob/d6badf773fb98e973f274b1cb0dcbcc005e4d748/grc/gui_qt/components/canvas/flowgraph.py#L343-L344

PS: Also, I agree with @marcusmueller's comments :)

Sahil-u07:
Hi @haakov 

Thank you for the detailed feedback! I'm glad the fix works. I understand the root cause of the issue now, and I agree that we shouldn't have to handle move actions with zero diffs.

I'll proceed with adding `self.moving_blocks = False` between `self.mousePressed = False` and `super(FlowgraphScene, self).mouseReleaseEvent(event)` in `FlowgraphScene::mouseReleaseEvent()`. This should prevent unnecessary `itemMoved `signals from being emitted.

I'll update the PR soon with these changes.

Thanks again for your help!

grc/gui_qt/components/undoable_actions.py
Sahil-u07, line 1: Hi @marcusmueller ,

Thank you for the feedback! I've removed the extra empty line in the undo() method as requested.

This is my first contribution to open source, and I’m really excited to be part of the project. Please let me know if there’s anything else I can improve.

Thanks again!

willcode:
Was the null MoveAction due to moving_blocks not being reset? Do we need the check in MoveAction() now?

haakov:
@Sahil-u07 Thank you for updating the PR! 

I agree with @willcode, you can remove the check in `MoveAction.__init__()`. That way, if a similar problem with "void" MoveActions shows up again, we'll know about it.

The only thing remaining after that is squashing your commits into a single commit, and this PR should be ready for merge

grc/gui_qt/components/undoable_actions.py
```diff
@@ -61,6 +61,9 @@ def undo(self):
 
 class MoveAction(QUndoCommand):
     def __init__(self, scene: FlowgraphScene, diff: QPointF):
+        # Prevent empty undo actions if the block hasn't moved
```
haakov, line 64: Remove this, as mentioned in my other comment :)

haakov (changes requested):

haakov:
@Sahil-u07 let us know if you have questions on how to squash the PR :)

willcode:
> @Sahil-u07 let us know if you have questions on how to squash the PR :)

And remove the check before squashing, as mentioned above.

willcode:
Included in #7673

PR #7626, opened 2025-01-05
Document that tags are already sorted, don't sort them

### LINKED ISSUE #7621: block detail still talks of tags being kept in a deque; they're not
### What happened?

we've moved away from keeping tags in deque for a long time
include/gnuradio/block_detail.h:105

### System Information

n/a

### GNU Radio Version

3.11-git (main)

### Specific Version

v3.11.0.0git-862-ge95299ec5

n/a

### Relevant log output

_No response_

marcusmueller:
## Description

`block::get_tags_in_range` fills the passed vector in a sorted manner. That became the case after we replaced the deque implementation with a `std::multimap<uint64_t offset, tag_t tag>` somewhen after GRCon'14.

`gr-uhd` definitely already depends on that, and a lot of other places do, as well. So, we should document this behaviour.

Also, we should remove the remaining places where we still sorted the vector.

## Related Issue

Closes #7621

Only documentation

## Checklist

### DIFF
--- a/gnuradio-runtime/include/gnuradio/block.h
+++ b/gnuradio-runtime/include/gnuradio/block.h
@@ -852,6 +852,8 @@ protected:
      *
      * Range of counts is from start to end-1.
      *
+     * The vector is sorted ascendingly by the offset of the tags.
+     *
      * Tags are tuples of:
      *      (item count, source id, key, value)
      *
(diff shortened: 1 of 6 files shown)


willcode:
Instead of "... inherently ...", which causes pondering about the nature of such things, how about something like "... returns a vector sorted by ..."

marcusmueller:
@willcode  good call! Thanks! Done.

PR #7634, opened 2025-01-05
blocks: Fix useless sample size functions in wavfile_sink (backport to maint-3.10)

willcode:
Backport #7550


willcode:
DCO override - signoff email mismatch, but appears valid.

PR #7639, opened 2025-01-06
Tagged Stream Block: don't remove item tags from input, don't show them internally instead

### LINKED ISSUE #7627: tagged_stream_block uses deprecated remove_item_tag
### What happened?

tagged stream blocks rely on the default copy_all_to_all tag propagation policy, and remove the length tags using the deprecated `block::remove_item_tag`, which removes the item *from its input*.

The existence of `remove_item_block` is a bit surprising, for two reasons:
1. The "GNU Radio contract", if you explain it to anyone, is that blocks read their input without modifying it. Tags are input. 
2. we iterate over a "is this tag deleted on this input" list in. every. single. call. to `get_tags_in_range`; just so that subclasses of `tagged_stream_block` don't see the tag. Luckily, this vector is always empty, aside if a tag was emitted from a block that is connected to a TSB. 

#### Remarks
- just keeping the tag in the input items is going to break existing TSBs
- maybe simply shadowing `get_tags_in_range` in `gr::tagged_stream_block` is sufficient: in the context of a TSB's `work` function, `get_tags_in_range()` unambigously means `tagged_stream_block::get_tags_in_range` then, and that could just hide these tags from the reader. Where it fails is when someone accesses the block through a pointer to `gr::block`, but I don't see why someone should call `get_tags_in_range` on a different block (you need to be able to access `protected:` of `gr::block`). But then they simply see the length tag, which probably was desired? Don't know, don't think calling `some_other_block_ptr->get_tags_in_range()` is a thing that ever happens.
- worst case, we need to modify `gr::block`'s API, and add a "block::set_tag_visibility_filter(predicate)` or such, which `block::get_tags_in_range` uses.

### System Information

n/a

### GNU Radio Version

3.11-git (main)

### Specific Version

main, maint-3.10, maint-3.9

n/a

### Relevant log output

_No response_

marcusmueller:
## Description

Before, the TSB superclass used to modify the contents input buffer tags
vector, adding itself to the list of blocks whom to not show the tag on
tags with the length tag key.

Now, instead of doing that, it simply "filters" what TSBs see by
(non-virtually) overriding get_tags_in_{range,window}.

This removes a contention point, avoids adding complex data to the
oft-copied blocks, and removes the usage of the deprecated
`block::remove_item_tag`. It was the last consumer of that function, so
that can now be retired.

Fixes a small problem with `qa_ofdm_txrx` (1040.0 is not exactly the
same as 1040), and correctly copies tags from in- to output in
`digital::packet_headergenerator_bb_impl`.

## Related Issue

Closes #7627

TSBs

## Testing Done

Full test suite, covering same functionality, which lead to the fixing of `packet_headergenerator_bb_impl`

## Checklist

  - https://wiki.gnuradio.org/index.php?title=Tagged_Stream_Blocks#A_note_on_tag_propagation currently can't mention `TPP_TSB` yet

### DIFF
--- a/gnuradio-runtime/include/gnuradio/tagged_stream_block.h
+++ b/gnuradio-runtime/include/gnuradio/tagged_stream_block.h
@@ -1,6 +1,7 @@
 /* -*- c++ -*- */
 /*
  * Copyright 2013 Free Software Foundation, Inc.
+ * Copyright 2025 Marcus Müller <mmueller@gnuradio.org>
  *
  * This file is part of GNU Radio
  *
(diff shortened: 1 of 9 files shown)


gnuradio-runtime/lib/tagged_stream_block.cc
```diff
@@ -44,14 +47,74 @@ void tagged_stream_block::forecast(int noutput_items,
     }
 }
 
+void tagged_stream_block::get_tags_in_range(std::vector<tag_t>& v,
```
mbr0wn, line 50: Does this not produce compiler warnings?

gnuradio-runtime/include/gnuradio/tagged_stream_block.h
```diff
@@ -38,6 +39,79 @@ class GR_RUNTIME_API tagged_stream_block : public block
                         gr::io_signature::sptr output_signature,
                         const std::string& length_tag_key);
 
+    /*!
+     * \brief Given a [start,end), returns a vector of all tags in the range.
+     *
+     * \see block::get_tags_in_range
+     *
+     * Does the same as block::get_tags_in_range, but omits the length_tag_key-tagged tags
```
mbr0wn, line 47: I would suggest adding a comment here like:
```suggestion
     * Does the same as block::get_tags_in_range, but omits the length_tag_key-tagged tags
     *
     * This hides `block:get_tags_in_range()`, but does not overload it. When calling get_tags_in_range() from work(), this function will be called.
```

mbr0wn (approved):
Change seems fine (in the context of TSBs).

gnuradio-runtime/lib/tagged_stream_block.cc
```diff
@@ -44,14 +47,74 @@ void tagged_stream_block::forecast(int noutput_items,
...
+                                            uint64_t abs_end)
+{
+    std::vector<tag_t> tmp;
+    gr::block::get_tags_in_range(tmp, which_input, abs_start, abs_end);
+    v.clear();
+    for (const auto& tag : tmp) {
+        if (!pmt::eqv(tag.key, d_length_tag_key)) {
+            v.push_back(tag);
+        }
+    }
```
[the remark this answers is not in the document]
marcusmueller: this is *EXACTLY* what I'm currently preparing to be possible, if the user code happens to be compiled as C++20. 

(background: idea is to add a 
```c++
block::apply_to_tags_in_range(start, end, predicate, bool (*operation)(const tag_t&))
```
 which just internally (essentially) 
```c++
for(auto iter = d_tag_container.lower_bound(start); iter != d_tag_container.end() && iter.offset <= end; ++iter) {
  if(predicate(*iter)){
    if(!operation(*iter)) {
      break;
    }
  }
}
```

Make `predicate=[key&](const auto& tag){return tag.key == key;}` and `operator=std::back_inserter(vec)`, and you get `get_tags_in_range(start, end, vec, key)`. But you could instead also just, say, insert values into a `std::map` or modify filter taps or whatever in `operation`.

This doesn't look like much for a start, but it's really but a small step from making `get_tags` become a *view*.)

gnuradio-runtime/lib/tagged_stream_block.cc
marcusmueller: nope, this is polymorphism, non-dynamic.

gnuradio-runtime/include/gnuradio/tagged_stream_block.h
marcusmueller: it is an overload, just not one that is bendy at runtime :) but yeah, comments are good.

marcusmueller: (hm, not sure I was using the right words there. I think "overloading" in C++ is only applied to "operator"; but what we both seem to mean is "this is inheritance, with the function getting resolved to different things, depending on whether the type of the object handled is an instance of `block` or `tagged_stream_block`.

willcode:
@marcusmueller were you going to make the `const` change above, and good to merge so we can move through the dependent PRs?

marcusmueller:
yes, just didn't have the time; will do now (and rebase other PRs on top :)

PR #7640, opened 2025-01-06
Get rid of block::remove_item_tag and related infrastructure, checks

marcusmueller:
## Description

With TSBs no longer marking tags as deleted in their input buffer, we
can drop the "Will be removed in 3.8"-deprecated
`block::remove_item_tag`.

With that gone, nothing can put blocks on the "am I deleted for this
particular block?" vector of tags. Hence, the functions called from
`block::get_tags_in_range`, `block_detail::…`, and
`buffer_reader::get_tags_in_range` no longer need to check whether a
block is on that list. Hence, remove that check, and the "who is
asking?" `id` argument.

These were all long-deprecated or internal/detail API anyways, so we're
not breaking compatibility.

It'd be more than desirable to remove the deletion list from `tag_t`,
but that's an API breakage (though I **really** hope nobody was using
that public field of the struct), and should be done in a separate
commit.

## Related Issue

Builds upon #7639

All blocks handling tags might now be a tiny bit faster.

This removed a method marked as to be removed in GNU Radio 3.8 (oops).

This removed methods only used runtime-internally.

## Testing Done

all current unit tests pass.

## Checklist

### DIFF
--- a/gnuradio-runtime/include/gnuradio/block.h
+++ b/gnuradio-runtime/include/gnuradio/block.h
@@ -846,40 +846,6 @@ protected:
     void add_item_tag(unsigned int which_output, const tag_t& tag);
 
     /*!
-     * \brief DEPRECATED. Will be removed in 3.8.
-     *
-     * \param which_input an integer of which input stream to remove the tag from
-     * \param abs_offset   a uint64 number of the absolute item number
-     *                     associated with the tag. Can get from nitems_written.
-     * \param key          the tag key as a PMT symbol
-     * \param value        any PMT holding any value for the given key
-     * \param srcid        optional source ID specifier; defaults to PMT_F
-     *
-     * If no such tag is found, does nothing.
-     */
-    inline void remove_item_tag(unsigned int which_input,
-                                uint64_t abs_offset,
-                                const pmt::pmt_t& key,
-                                const pmt::pmt_t& value,
-                                const pmt::pmt_t& srcid = pmt::PMT_F)
-    {
-        tag_t tag;
-        tag.offset = abs_offset;
-        tag.key = key;
-        tag.value = value;
-        tag.srcid = srcid;
-        this->remove_item_tag(which_input, tag);
-    }
-
-    /*!
-     * \brief DEPRECATED. Will be removed in 3.8.
-     *
-     * \param which_input an integer of which input stream to remove the tag from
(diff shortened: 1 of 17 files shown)


marcusmueller:
Note that the `tag_t` change here is 100% backwards compatible – we just replaced two user-implemented default ctors with the default ones, doing the same, and removed an empty destructor.

marcusmueller:
@willcode rebased!

argilo:
FYI @marcusmueller @bastibl gr-foo no longer builds because it uses `remove_item_tag`:

https://github.com/bastibl/gr-foo/blob/0a34f68565863218c3c010e69f6bd6a9bdfd7142/lib/packet_pad_impl.cc#L140

argilo:
gr-foo is a dependency of gr-ieee-802-11 and gr-ieee802-15-4.

willcode:
I suspect that line in `gr-foo` can be removed, since propagation is `TPP_DONT`. Unless there's a problem with double calls. Also that function looks a bit fragile ... if the tag isn't sob or eob, then it assumes there's a second tag?

argilo:
> Unless there's a problem with double calls.

Without the `remove_item_tag`, I suspect an infinite loop would occur if `pad_front` is set to zero.

argilo:
The `general_work` function could certainly be rewritten so as not to depend on `remove_item_tag`. At present, it does an early return any time it needs to transition into a different state. Adding an outer loop would allow it to continue processing samples until the input buffer is consumed or the output buffer is full.

willcode:
Since this is supposed to be an API-compatible release, I'll back out the change on maint-3.10. @marcusmueller I'll let you decide what to do with main. Since gr-foo is probably packaged somewhere, we can't pull the rug out from under it. Even if it is a ratty, beer-stained rug that we said we wanted to get rid of (but didn't).

argilo:
That said, I wonder whether the Packet Pad block is still relevant. The Burst Shaper block was added to GNU Radio in 2015, and provides similar functionality.

willcode:
I'll give it a couple days and see what people think.

argilo:
@bastibl removed the affected "Packet Pad" block from gr-foo, since its functionality had already been replaced by the "Packet Pad2" block.

Of course, there remains the risk that other projects depend on `remove_item_tag`, so it may make sense to delay its removal until 3.11.

PR #7642, opened 2025-01-07
tag_t: clean up marked_deleted

marcusmueller:
## Description

with the work done to wean tagged stream blocks off `remove_item_tag`, we can now remove the `marked_deleted` member.

## Related Issue

Depends on 

- #7639
- #7640

Tags; tags being among the most-handled objects, shaving up some 24-30 B for an unused `std::vector` member should actually be nice for cache locality when walking through the tag tree in gr::buffer (pending moving from a `std::multimap` to a `std::multiset` with the foreign-key `<container>::lower_bound` functionality).

## Checklist


willcode:
API change: well, technically, but it doesn't seem like there's any external use for it.

Issue #7615, opened 2024-12-31
Var.-Editor context menu: Be context-sensitive!


marcusmueller:
as of 37251f139961c9ff4dfd0039d84593842d7478fd:

![Image](https://github.com/user-attachments/assets/c21dfe19-3f00-4aa2-b653-8e1d8f14e3b3)

- Show "enable" only when disabled
- Same, but opposite for "disabled"
- Remove "add import" and "add variable", unless right-clicking on the variables or imports category header.
- Properties dialog doesn't open when clicking "Properties…"

Also: 

- ~~clicking on the "+" next to "Variables" and "Imports" is nonfunctional~~ (fixed)
- ~~clicking on the "-" next to and entry is nonfunctional~~ (fixed)

Harishankar14:
"Hi @marcusmueller, I’d like to take up this issue and improve the context menu behavior and functionality in the Variable Editor. My approach will involve:

Making sure that  the menu options context-sensitive (e.g., showing 'Enable' only for disabled items and vice versa).
Restricting 'Add Import' and 'Add Variable' to category headers only.
Fixing the 'Properties' action to open the dialog correctly.
Addressing the nonfunctional expand/collapse behavior for '+' and '-' buttons.
I'll test all changes thoroughly to ensure everything works as expected. Let me know if there are any additional considerations.

marcusmueller:
Awesome, @Harishankar14!

Please go ahead! Looking forward to your pull request (pro trick: you can open a "draft" PR early on with the first modifications you've done, even if things aren't perfect yet).

the-gremlin:
Hi @Harishankar14, would you mind if I took on the "+" and "-" buttons issue? I could even help you with reviewing your code or by tackling more of the issue.

Harishankar14:
@the-gremlin  go on

ei8fdb:
Hi @Harishankar14 is there anything you need/are waiting for with your part (i.e. - the "+" and "-" buttons handled by @the-gremlin ) of this issue?

Harishankar14:
@ei8fdb would love to work on this 

- Show "enable" only when disabled

- Same, but opposite for "disabled"

- Remove "add import" and "add variable", unless right-clicking on the variables or imports category header.

- Properties dialog doesn't open when clicking "Properties…"

PR #7643, opened 2025-01-07
GRC: Make "+" and "-" icons work in variable editor as expected

the-gremlin:
## Description

Clicking the "-" icon on a variable or an import will remove it.
Clicking the "+" icon on the "Variables" or "Imports" tabs creates a new variable or new import accordingly.

## Related Issue

Fixes a part of #7615 

The variable editor of the GRC QT gui

## Testing Done

Manual testing.

## Checklist

### DIFF
--- a/grc/gui_qt/components/variable_editor.py
+++ b/grc/gui_qt/components/variable_editor.py
@@ -52,11 +52,19 @@ class VariableEditor(QDockWidget, base.Component):
         self.currently_rebuilding = True
         self._model.dataChanged.connect(self.handle_change)
 
+        self.var_add = VariableEditorAction.ADD_VARIABLE
+        self.import_add = VariableEditorAction.ADD_IMPORT
+        self.del_block = VariableEditorAction.DELETE_BLOCK
+        self._tree.itemClicked.connect(self.handle_click)
+
         imports = QTreeWidgetItem(self._tree)
         imports.setText(0, "Imports")
+        imports.setData(2, Qt.UserRole, self.import_add)
         imports.setIcon(2, QtGui.QIcon.fromTheme("list-add"))
+
         variables = QTreeWidgetItem(self._tree)
         variables.setText(0, "Variables")
+        variables.setData(2, Qt.UserRole, self.var_add)
         variables.setIcon(2, QtGui.QIcon.fromTheme("list-add"))
         self._tree.expandAll()
 
@@ -96,13 +104,27 @@ class VariableEditor(QDockWidget, base.Component):
     def keyPressEvent(self, event):
         super(VariableEditor, self).keyPressEvent(event)
 
+    def handle_click(self, item, col):
+        # we only care about the add/remove icons being clicked
+        if (col != 2):
+            return
+
+        action = item.data(2, Qt.UserRole)
+
+        if action is None:
+            log.warn("Item %s does not contain any actions!!", item)
+
+        self.handle_action(action)
+
     def set_scene(self, scene: FlowgraphScene):
         self.scene = scene
         self.update_gui(self.scene.core.blocks)
         self._tree.resizeColumnToContents(0)
         self._tree.resizeColumnToContents(1)
 
-    def handle_change(self, tl, br):  # TODO: Why are there two arguments?
+    # the handler is required by the signal to take two arguments even
+    # if we only use one
+    def handle_change(self, tl, br):
         if self.currently_rebuilding:
             return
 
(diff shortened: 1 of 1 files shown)


the-gremlin:
The gtk gui pops up a sort-of confirmation button when clicking the "-" symbol on the editor that says "Delete <variable name>", I didn't know if it should also be a part of the QT gui so I'm leaving it as a draft until this gets confirmed.

marcusmueller:
my 2 ct:

That operation (deletion) is "undo"-able. I'm a big fan of not asking users for confirmation for actions that can, same effort as giving confirmation, be undone losslessly.

grc/gui_qt/components/variable_editor.py
```diff
@@ -96,6 +97,18 @@ def contextMenuEvent(self, event):
...
+        # we only care about the add/remove icons being clicked
+        if (col != 2):
+            return
+
+        if item.text(0) == "Imports":
+            self.handle_action(VariableEditorAction.ADD_IMPORT)
+        elif item.text(0) == "Variables":
+            self.handle_action(VariableEditorAction.ADD_VARIABLE)
+        else:
+            self.handle_action(VariableEditorAction.DELETE_BLOCK)
```
marcusmueller, line 110: Gut feeling:

Not a fan of using the text displayed on a menu item as "magic value" somewhere else in the code to track which menu item was clicked. That keeps the program untranslatable, and means someone changing appearance unintentionally changes behaviour.

Can we instead have different `handle_click`-like functions, each connected to the `variables` or `imports` object from `__init__`? Don't know whether that's technically possible. If it isn't, one would store `variables` and `imports` in `self`, and compare against the `item`, rather than `item.text(0)`.

the-gremlin:
that makes a lot of sense, iIm also of the opinion to not have the confirmation because in the gtk ui looks very clunky. So I'll mark the PR as ready to review and wait in case there's an argument for the pop-up

grc/gui_qt/components/variable_editor.py
marcusmueller: You're of course right; the items here are just entries in the data model that the tree view displays. My mistake.

I can see the problem with the dangling references. What about we give the items data, i.e., use setData to identify them?

the-gremlin: I hadn't thought of that, but it is definitely the way to go about it. Im stuck on what the data should be but since we're talking about literally two objects, could we get away with it just being strings? 
i.e.
`imports.setData(0, QtGui.UserRole, "imports")`
since strings are immutable in python i think this will be pretty inexpensive.
Otherwise we can assign some constant value to two fields of the class and set that to be the data. 
i.e.
`__init__():`
`...`
`self.import_const = some_value`
`...`
and
`imports.setData(0, QtGui.UserRole, self.import_const)`

But I'm a fan of the strings because we can be pretty explicit and use something like "Imports Tab", making what is going on very clear.

marcusmueller: for something that only happens when a user clicks, you really don't have to think that much about efficiency, by the way :) That becomes important when you need to do something a couple thousand times a second.

I do like the constant idea! But for a different reason: You already have the right constants for that: in the `VariableEditorAction` enum

the-gremlin: Oh wow that's very clever, I'm going to implement the change really quick and then squash and push the changes

grc/gui_qt/components/variable_editor.py
```diff
@@ -52,11 +52,19 @@ def __init__(self):
...
         self._model.dataChanged.connect(self.handle_change)
 
+        self.var_add = VariableEditorAction.ADD_VARIABLE
+        self.import_add = VariableEditorAction.ADD_IMPORT
+        self.del_block = VariableEditorAction.DELETE_BLOCK
+        self._tree.itemClicked.connect(self.handle_click)
+
         imports = QTreeWidgetItem(self._tree)
         imports.setText(0, "Imports")
+        imports.setdata(2, Qt.UserRole, self.import_add)
```
haakov, line 62: I believe `setdata` should be `setData`

haakov (changes requested):

grc/gui_qt/components/variable_editor.py
the-gremlin: you're absolutely right, thank you

haakov (approved):
Thank you for the PR!

PR #7648, opened 2025-01-10
runtime: block gets get_first_tag_in_range(…, predicate) and …, tag)

marcusmueller:
# Description

All to common is a scheme where a consumer of tags needs to do this
dance:

```c++
std::vector<gr::tag_t> tags;
get_tags_in_range(tags, 0, start, end, d_phase_est_key);
if(!tags.empty) {
  float phase = pmt::to_float(tags[0].value);
  if(phase >= -M_PI && phase < M_PI) {
    d_phase = phase;
  }
}
```

but no more with `get_first_tag_in_range`!

```c++
auto tag = get_first_tag_in_range(0, start, end, d_phase_est_key);
if(tag) {
  float phase = pmt::to_float(tag.value.value);
  if(phase >= -M_PI && phase < M_PI) {
    d_phase = phase;
  }
}
```

But wait! It gets even better! You can *filter with a predicate*:

```c++
auto tag = get_first_tag_in_range(0,
                                  start,
                                  end,
                                  [this](const gr::tag_t& tag) {
                                    if(tag.key != d_phase_est_key)
                                      return false;
                                    float phase = pmt::to_float(
                                      tag.value.value);
                                    return phase >= -M_PI
                                      && phase < M_PI;
                                  });
if(tag) {
  float phase = pmt::to_float(tag.value.value);
  if(phase >= -M_PI && phase < M_PI) {
    d_phase = phase;
  }
}
```

This also works in python (in both variants), feel free to use python
callables as predicate, or your usual PMT key.

## Related Issue

Long-standing annoyences

block, block_gateway

block_detail

and their python bindings

## Testing Done

added a Python test case to check functionality.

## Checklist

### DIFF
--- a/gnuradio-runtime/include/gnuradio/block.h
+++ b/gnuradio-runtime/include/gnuradio/block.h
@@ -11,7 +11,9 @@
 #ifndef INCLUDED_GR_RUNTIME_BLOCK_H
 #define INCLUDED_GR_RUNTIME_BLOCK_H
 
+#include <cstdint>
 #include <memory>
+#include <optional>
 
 #include <gnuradio/api.h>
 #include <gnuradio/basic_block.h>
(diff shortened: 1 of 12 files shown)


marcusmueller:
It has finally happened, my clang-format  (v18) does completely different things that our CI. Same for my pycodestyle :(

willcode:
Is there a way to make the interface more symmetric, i.e., not need `.value.value`?

gnuradio-runtime/include/gnuradio/block.h
```diff
@@ -928,6 +930,59 @@ class GR_RUNTIME_API block : public basic_block
...
+     * auto timestamp = get_first_tag_in_range(
+     *     0,                          // which input
+     *     nitems_read(0),             // start index
+     *     nitems_read(0) + something, // end
+     *     [this](const gr::tag_t& tag) {
+     *         return pmt::eqv(tag.key, d_time_tag) && !pmt::is_null(tag.value)
+     *     });
+     * if (timestamp) {
+     *     d_logger->info("got time tag {} at offset {}",
+     *                    timestamp.value.value,
```
willcode, line 954: Is this just `timestamp.value` and `timestamp.offset`? Based on the QA code, I think so.

gnuradio-runtime/include/gnuradio/block_detail.h
```diff
@@ -152,6 +152,44 @@ class GR_RUNTIME_API block_detail
...
+     * auto timestamp = get_first_tag_in_range(
+     *     0,                          // which input
+     *     nitems_read(0),             // start index
+     *     nitems_read(0) + something, // end
+     *     [this](const gr::tag_t& tag) {
+     *         return pmt::eqv(tag.key, d_time_tag) && !pmt::is_null(tag.value)
+     *     });
+     * if (timestamp) {
+     *     d_logger->info("got time tag {} at offset {}",
+     *                    timestamp.value.value,
```
willcode, line 176: Same comment as above.

gnuradio-runtime/include/gnuradio/block.h
marcusmueller: no, it's `timestamp.value.value`, because `timestamp` is a `std::optional`, and you need to access its value (if any, hence the `if` conditional) via its `value` member, which then is a tag, which has a `value` member.

So sadly, no way I can see around this

marcusmueller: In Python, a function doesn't have a fixed return type, so the wrapped function either returns a `NoneType` or `gr.tag_t`; hence, no need for the `.value` from the optional. But that's pretty pythonic, so it feels natural, I'd say.

willcode: That was my point ... the existing functions use `.value` and the new function uses `.value.field`, so it requires the caller to know about `std::optional` vs the old way, making the API a little asymmetric.

willcode: FWIW, I don't have a great solution either, since `std::optional` is easier to use but requires a different coding style. I can just picture people moving code to the new function, and then having to add either the extra `.value` or creating a new variable so the existing code still works. Not a huge deal.

PR #7652, opened 2025-01-11
trellis: don't bring your own quicksort to a stdlib party

### LINKED ISSUE #7625: gr-trellis/interleaver.cc: don't use your own quicksort implementation (please)
### What happened?

gr-trellis contains its own C quicksort implementation, which was modified to work on std::vectors instead of C arrays.

We should not be carrying around our own quicksort; that's nonsense we need to maintain. It's also probably slower than `std::sort`.

### System Information

n/a

### GNU Radio Version

3.11-git (main)

### Specific Version

a83fa369ad8a53825518ca30f6b716923b1c34a0

n/a

### Relevant log output

_No response_

marcusmueller:
## Description

For reasons that are lost to time, gr::trellis::interleaver's random interleaver constructor brings its own quicksort implementation to reorder an index vector according to a (pseudo-)random number vector. Let's decruftify that!

Sadly, there's no unit test covering that constructor, so we will need to start by working on QA.

## Related Issue

Fixes #7625

gr::trellis::interleaver in random mode

## Testing Done

adding QA to see end-to-end (hopefully: still, and not the first time?) correct.

## Checklist


gr-trellis/python/trellis/bindings/interleaver_python.cc
```diff
@@ -14,56 +15,81 @@
...
+                 &interleaver::interleaver_indices,
+                 D(interleaver, interleaver_indices))
+            .def("DEINTER",
+                 &interleaver::deinterleaver_indices,
+                 D(interleaver, deinterleaver_indices));
+        interleaver_bind.def(
+            py::init<unsigned int, std::vector<int, std::allocator<int>> const&>(),
+            py::arg(length_arg),
+            py::arg(interleaver_arg),
+            D(interleaver, interleaver, 2));
```
marcusmueller, line 92: this fails on `-Werror` builds because the deprecation warning gets emitted whether or not we actually bind or not. Need some precision preprocessor programming…

marcusmueller:
Depends on #7653

willcode:
Ok, try the rebase.

marcusmueller:
rebased!

PR #7653, opened 2025-01-11
runtime: add configure-time version.h[.in] with GR_VERSION_{MAJOR,API,MINOR,MAINT}

marcusmueller:
## Description

As discussed in #Development, it seems a common and desirable thing to have a version preprocessor statement.

Short survey of /usr/include suggests that gnuradio/version.h is a good name.

## Related Issue

Needed this to be able to future-remove features in the source code. 

none, unless you want to use it

## Checklist

### DIFF
--- a/cmake/Modules/GrVersion.cmake
+++ b/cmake/Modules/GrVersion.cmake
@@ -64,6 +64,12 @@ endif()
 ########################################################################
 # Use the logic below to set the version constants
 ########################################################################
+## HEADER VARIABLES -- see include/gnuradio/version.h.in
+set(GR_VERSION_MAJOR "${VERSION_MAJOR}")
+set(GR_VERSION_API "${VERSION_API}")
+set(GR_VERSION_MINOR "${VERSION_ABI}")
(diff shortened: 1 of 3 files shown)


willcode:
What is the theory behind the `<<10`, `<<20`, ...?

marcusmueller:
@willcode  I could have just as well used ×1000, ×1,000,000… , but that invites people to just in-head convert 3.10.10.11 into decimal, and that's error-prone (that's why there is the macro). But it's honestly just a preference of mine.

willcode:
So, this is an incrementing number that isn't meant to have any visual meaning? OK.

Assuming a 64-bit value, since the eventual `4<<30` will require it. Unsigned 32-bit works for v3.

marcusmueller:
> Assuming a 64-bit value, since the eventual 4<<30 will require it. Unsigned 32-bit works for v3.

Correct, but as far as I could find, C++ requires the preprocessor integer type to be compatible with `int_max_t`, and that should be at least 64 bit on any thinkable 32 bit or more platform.

PR #7674, opened 2025-01-22
Fix typos in modtool and block (backport to maint-3.10)

willcode:
Backport #7519 
Backport #7664


willcode:
DCO override ... signoff by github users.

PR #7680, opened 2025-01-22
trellis: don't bring your own quicksort to a stdlib party (backport to maint-3.10)

willcode:
Backport #7652 
Backport #7216 

It appears that I backported 7216 from main to main somehow, so this actually does the backport that #7239 meant to.


willcode:
@marcusmueller any idea what's going on with F40/clang here? Did I get a fix out of order somewhere?

PR #7690, opened 2025-01-24
GRC:Fix properties dialog in variable editor 

Harishankar14:
**Description.**

Properties dialog doesn't open when clicking "Properties…"

Related Issue
Fixes a part of https://github.com/gnuradio/gnuradio/issues/7615

Which blocks/areas does this affect?
The variable editor of the GRC QT GUI

This PR fixes an issue where the properties dialog does not open when clicking "Properties" in the variable editor context menu. The changes include:
- Added logic to open the properties dialog when the "Properties" action is triggered.
- Updated the context menu to enable the "Properties" option only when a valid block is selected.

Note: I could not test these changes locally due to issues with building and running GNU Radio from the source. I’ve made the changes based on the code structure and feedback from the community. I would appreciate it if the maintainers could test the changes and provide feedback.

### DIFF
--- a/grc/gui_qt/components/variable_editor.py
+++ b/grc/gui_qt/components/variable_editor.py
@@ -5,7 +5,7 @@ import logging
 from enum import Enum
 
 from qtpy import QtGui
-from qtpy.QtWidgets import QMenu, QAction, QDockWidget, QTreeWidget, QTreeWidgetItem
+from qtpy.QtWidgets import QMenu, QAction, QDockWidget, QTreeWidget, QTreeWidgetItem, QDialog, QVBoxLayout, QLabel, QLineEdit, QDialogButtonBox
 from qtpy.QtCore import Slot, Signal, QPointF, Qt, QVariant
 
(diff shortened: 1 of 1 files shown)


willcode:
Author says this is not tested, so I added "Needs Action" ... won't merge until someone gets a chance to test.

marcusmueller:
@Harishankar14 if you have any trouble building GNU Radio, do reach out via [chat](https://app.element.io/#/room/#gnuradio-space:gnuradio.org) or [mailing list](https://lists.gnu.org/mailman/listinfo/discuss-gnuradio). You should really be able to test your own code changes, anything else would be far too frustrating for you and us!

dl1ksv:
This pr won't run, as in grc scene has not an 'open_block_properties' function as assumed.
In addition we can use the existing open_properties dialog instead of coding an additional dialog.
A  proposal for an alternative solution is #7709

willcode:
Closing, since it's untested code and not the preferred approach.

Issue #7353, opened 2024-05-18
gr-uhd: is_back_edge argument is ignored when connecting UHD blocks.


pamin98:
### What happened?

According to the UHD documentation, graphs consisting of UHD blocks are allowed to have loops if one of the connections is declared a back edge. However, when creating such a design, a runtime error occurs, regardless of whether the boolean is_back_edge argument is set to True. By inspecting the code, one can find the connect() function implementation in gnuradio/gr-uhd/lib/rfnoc_graph_impl.cc:
```
    void connect(const std::string& src_block_id, 
                 const size_t src_block_port,
                 const std::string& dst_block_id,
                 const size_t dst_block_port,
                 const bool is_back_edge) override
    {
        d_logger->debug("Connecting {:s}:{:d} -> {:s}:{:d}",
                        src_block_id,
                        src_block_port,
                        dst_block_id,
                        dst_block_port);
        if (_tx_streamers.count(src_block_id)) {
            if (_rx_streamers.count(dst_block_id)) {
                throw std::runtime_error("Cannot connect RFNoC streamers directly!");
            }
            _graph->connect(_tx_streamers.at(src_block_id),
                            src_block_port,
                            block_id_t(dst_block_id),
                            dst_block_port,
                            _get_adapter_id(src_block_id, src_block_port));
            return;
        }
        if (_rx_streamers.count(dst_block_id)) {
            _graph->connect(src_block_id,
                            src_block_port,
                            _rx_streamers.at(dst_block_id),
                            dst_block_port,
                            _get_adapter_id(dst_block_id, dst_block_port));
            return;
        }

        _graph->connect(block_id_t(src_block_id),
                        src_block_port,
                        block_id_t(dst_block_id),
                        dst_block_port);
    } 
```

To fix the bug, the final call to _graph->connect() should be modified to include the is_back_edge argument, as it is otherwise set by default to False:
```
_graph->connect(block_id_t(src_block_id),
                            src_block_port,
                            block_id_t(dst_block_id),
                            dst_block_port,
                            is_back_edge);
```

### System Information

OS: Ubuntu 
GR Installation Method: Source

### GNU Radio Version

4.0-git (dev-4.0)

### Specific Version

0.0.0.0

### Steps to Reproduce the Problem

Create a simple design containing a loop, using UHD blocks. Generate and run the code.

pamin98:
Is there any update for this issue? Can I open a PR myself to resolve it?

willcode:
[...]

PR #7703, opened 2025-01-29
uhd: Fix connecting back-edges

mbr0wn:
## Description

gr-uhd contained a bug where back-edges were not being properly propagated into UHD (meaning, the setting was not being set in the corresponding UHD call).

In addition, we add a warning when attempting to define an edge to a streamer node as a back-edge. Because these edges are never the cause of a loop, the underlying UHD API does not allow tagging these as back-edges. If someone attempts to declare a streamer-edge as a back-edge, we will now emit a warning on the logger interface.

## Related Issue

Fixes https://github.com/gnuradio/gnuradio/issues/7353.

RFNoC applications with loops.

## Testing Done

I used `rfnoc_radio_loopback.grc` to first reproduce the error, and then test this fix. Both were straightforward. This works best on an X410 or X440, for some reason the default X310 configuration won't let you do a single-radio loopback (not that it would be useful, so that's probably more of a feature than a bug).

The irony: I did all the GRC work to enable edge properties just so that `rfnoc_radio_loopback.grc` actually did what it was supposed to, which includes this very use case. I spent weeks on that feature, so I'm pretty sure I must have tested it at some point. I can only imagine that I eventually didn't stage all the changes for the PR.

## Checklist

### DIFF
--- a/gr-uhd/lib/rfnoc_graph_impl.cc
+++ b/gr-uhd/lib/rfnoc_graph_impl.cc
@@ -49,6 +49,14 @@ public:
             if (_rx_streamers.count(dst_block_id)) {
                 throw std::runtime_error("Cannot connect RFNoC streamers directly!");
             }
+            if (is_back_edge) {
+                d_logger->warn("Back edge detected between streamer {:s}:{:d} and block "
+                               "{:s}:{:d}! Ignoring.",
+                               src_block_id,
+                               src_block_port,
+                               dst_block_id,
+                               dst_block_port);
+            }
             _graph->connect(_tx_streamers.at(src_block_id),
                             src_block_port,
                             block_id_t(dst_block_id),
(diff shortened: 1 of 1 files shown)


mbr0wn:
`qa_tag_utils` failed here. Should I be worried about this?

gr-uhd/lib/rfnoc_graph_impl.cc
```diff
@@ -68,7 +84,8 @@ class rfnoc_graph_impl : public rfnoc_graph
         _graph->connect(block_id_t(src_block_id),
                         src_block_port,
                         block_id_t(dst_block_id),
-                        dst_block_port);
+                        dst_block_port,
+                        is_back_edge);
```
mbr0wn, line 88: FYI, this line is the actual  bug fix. The rest is just cosmetic logging.

PR #7704, opened 2025-01-29
gr-blocks: Unify "Multiply Const" and "Fast Multiply Const" blocks

eltos:
## Description

- Merges the functionality of  "Fast Multiply Const" block into "Multiply Const" block
- Deprecates "Fast Multiply Const"

"Multiply Const" block get's an additional parameter "mode" to select between "vector constant" (default for backwards compatibility) and "scalar constant" (functionality of Fast Multiply Const block).
If vector length is 1 (default), both "Multiply Const" and "Fast Multiply Const" already did the same thing anyways prior to this PR, and in this case the new "mode" parameter is also hidden.
For vector length > 1, the mode parameter allows to select between "vector constant" (keeping the previous behaviour of "Multiply Const" block) and "scalar constant" (the behaviour of "Fast Multiply Const" block)

gr-blocks "Multiply Const" and "Fast Multiply Const" blocks

## Testing Done

- Confirmed that generated output for python and c++ is syntactically identical

## Checklist

### DIFF
--- a/gr-blocks/grc/blocks_multiply_const_vxx.block.yml
+++ b/gr-blocks/grc/blocks_multiply_const_vxx.block.yml
@@ -12,9 +12,16 @@ parameters:
         const_type:  [complex, real, int, int]
         fcn: [cc, ff, ii, ss]
     hide: part
+-   id: mode
+    label: Mode
+    dtype: enum
+    options: [scalar, vector]
+    option_labels: ['Scalar constant', 'Vector constant']
+    default: 'vector'
+    hide: ${ 'part' if vlen > 1 else 'all' }
 -   id: const
     label: Constant
-    dtype: ${ type.const_type if vlen == 1 else type.vconst_type }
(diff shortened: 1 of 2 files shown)


willcode:
Could you add a sentance to the comment up top about why we want this change?

eltos:
I thought that's obvious
- reduce redundancy (both yml's use the same underlying code and do almost the same)
- avoid confusion about the term "Fast", especially because for vlen=1 the behaviour is identical
- clarify and emphasize the actual difference between the blocks: the constant being a scalar or vector

PS: I did sign the commit with, your DCO bot generates a false error.

willcode:
Thanks for the update. I overrode the DCO. It's sensitive to differences in name/email.

drmpeg:
Just a thought. Change "Fast" to "Scalar" in the ```blocks_multiply_const_xx.block.yml``` label field. One line change.

PR #7705, opened 2025-01-29
grc-qt: Right-clicking in the canvas doesn't bring up the context menu

### LINKED ISSUE #7610: GRC Qt right-clicking in the canvas doesn't bring up the context menu
Right-clicking a block brings up the context menu, but if you right-click anywhere else, nothing happens.

### Where to start

Essentially, move this to `grc/gui_qt/components/flowgraph_view.py` or `grc/gui_qt/components/canvas/flowgraph.py` (not 100% which makes the most sense):
https://github.com/gnuradio/gnuradio/blob/76900adac6ab09ba3344412589fd590cf02a8ed6/grc/gui_qt/components/canvas/block.py#L383-L392

dl1ksv:
## Description

Right-clicking in the canvas now brings up a context menu on a selected block. If multiple blocks are selected the context menu is assoziated to the 'first' block in the list. If no block a hint appears.

If the mouse is over a block this block will be selected.

## Related Issue

Fixes #7610

## Checklist

### DIFF
--- a/grc/gui_qt/components/flowgraph_view.py
+++ b/grc/gui_qt/components/flowgraph_view.py
@@ -14,6 +14,7 @@ from .canvas.block import Block
 from .. import base
 from .canvas.flowgraph import FlowgraphScene, Flowgraph
 from .canvas.colors import LIGHT_FLOWGRAPH_BACKGROUND_COLOR, DARK_FLOWGRAPH_BACKGROUND_COLOR
+from .canvas.block import GUIBlock
 
 from ...core.generator import Generator
 
@@ -61,7 +62,16 @@ class FlowgraphView(
         log.debug("Creating actions")
 
     def contextMenuEvent(self, event):
-        super(FlowgraphView, self).contextMenuEvent(event)
+        if self.scene().selected_blocks():
+            block = self.scene().selected_blocks()[0]
+            self.scene().clearSelection()
+            block.setSelected(True)
+            block.context_menu(event.globalPos())
+        else:
+            empty_menu = QtWidgets.QMenu()
+            no_action = QtWidgets.QAction("No block selected", enabled=False)
+            empty_menu.addAction(no_action)
+            empty_menu.exec_(event.globalPos())
 
     def createMenus(self, actions, menus):
         log.debug("Creating menus")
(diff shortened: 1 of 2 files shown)


grc/gui_qt/components/flowgraph_view.py
```diff
@@ -61,7 +62,13 @@ def createActions(self, actions):
...
 
     def contextMenuEvent(self, event):
-        super(FlowgraphView, self).contextMenuEvent(event)
+        if self.scene().selected_blocks():
+            block = self.scene().selected_blocks()[0]
+            self.scene().clearSelection()
+            block.setSelected(True)
+            block.context_menu(event.globalPos())
+        else:
+            super(FlowgraphView, self).contextMenuEvent(event)
```
haakov, line 71: I think it would be great if something happened even if no blocks are selected. That way, a new user will get the idea that the right-click button can be used for something. GRC Gtk opens (a subset of) the Edit menu, which is an option. I tried replacing the `super()` call with something like
```python
empty_menu = QtWidgets.QMenu()
no_action = QtWidgets.QAction("No block selected", enabled=False)
empty_menu.addAction(no_action)
empty_menu.exec_(event.globalPos())
```

I think it looked okay. Open to suggestions here

haakov (changes requested):

grc/gui_qt/components/flowgraph_view.py
dl1ksv: Yes, this works for me. It looks good and acts like a special messages. But if we are right-clicking  in the canvas , we are on the fg level.  So why not open the file menu from the mainwindow as context menu? Here we get some usefull items.
In GRC GTK we get the edit menu with all items but 'more'  disabled and the 'more' submenu only contains disabled items.

So, does the File menu  bring any benefit to the user at this point? Otherwise I'd prefer your solution

haakov: Interesting thought, but I think I'd prefer the empty menu. I don't think I (as a user) would think to right-click the canvas as a shortcut to the File menu

dl1ksv: I added the empty menu.

Issue #7495, opened 2024-08-29
uhd_msg_tune.grc does not update frequency axis


wavewalkerdsp:
### What happened?

The flowgraph gr-uhd/examples/grc/uhd_msg_tune.grc does not update the frequency axis in QT GUI Freq Sink when UHD is retuned. The QT GUI Freq Sink does not have message input enabled, and when it does and connected to retune message, it still does not update.

![image](https://github.com/user-attachments/assets/3d02e4fe-4938-4712-bd7f-f5d3cf8b2bb6)

### System Information

OS: Ubuntu 24.04
GR Installation Method: source

### GNU Radio Version

3.11-git (main)

### Specific Version

v3.11.0.0git-809-g5c30d92b

### Steps to Reproduce the Problem

Run flowgraph, look at the frequency axis when retuning.

CanoRoger:
Most likely related, since the addition of the async_message output in the "UHD: USRP Source block", it is not taking anymore input commands. 

Can't tune frequency (and other requests) on USRP using PMTs. 

This worked in 3.10.1.1 and previous versions.
Not anymore at least tested from 3.10.11.0

On X300, UHD-3.15 and 4.1.

marcusmueller:
@CanoRoger  interesting! Could you open a new issue "USRP source doesn't add tags when tuning via cmd port" or such? I think we need to track this separately :)

PR #7707, opened 2025-01-30
uhd: examples: Fix uhd_msg_tune.grc

mbr0wn:
- Use "variable to message" which is now available
- Individual settings generate individual messages now (e.g., setting gain won't try to set frequency again)
- The frequency GUI output is now used as a message port as well, and the GUI x-axis is now properly updated on tuning

Here's the FG:

image

## Related Issue
Fixes https://github.com/gnuradio/gnuradio/issues/7495.

Just this example.

## Testing Done
Played around with X310 + siggen. Behaves as expected.

## Checklist


gr-uhd/examples/grc/uhd_msg_tune.grc
```diff
@@ -142,36 +133,132 @@ blocks:
...
+  id: parameter
   parameters:
+    alias: ''
     comment: ''
-    value: 2e6
+    hide: none
+    label: Device Arggs
+    short_id: a
+    type: str
+    value: type=x300
```
marcusmueller, line 148: We could remove that, to remove one more obstacle for newbs, but honestly, if you attach a B200 and wonder why a sink configured for type=x300 can't connect to it…

marcusmueller (approved):
love it!

gr-uhd/examples/grc/uhd_msg_tune.grc
mbr0wn: Good catch, removed. I also found a typo that I fixed.

PR #7715, opened 2025-02-04
Revert "Tags updates (backport to maint-3.10)"

willcode:
Reverts gnuradio/gnuradio#7667

Due to an API change and a deadlock, revert these updates.


marcusmueller (approved):
As much as it obviously pains me, let's do this. I'll have to roll a backport PR with the uncritical things myself.

Issue #7732, opened 2025-02-15
C++ wrong code generation: hier block not declared correctly to top code


gcarmix:
### What happened?

In gnuradio companion, when generating C++ code with NO GUI option enabled, if i add a hier block to the top design, then the generated top.cpp and top.hpp are all buggy:
-no declaration of the hier block in the top class in top.hpp file
-hier block constructor is called with wrong syntax in the top.cpp

### System Information

OS: Ubuntu 22.04
GR Installation Method: Source

### GNU Radio Version

3.10 (maint-3.10)

### Specific Version

3.10.11.1

### Steps to Reproduce the Problem

In gnuradio companion, when generating C++ code with NO GUI option enabled, if i add a hier block to the top design, then the generated top.cpp and top.hpp are all buggy:
-no declaration of the hier block in the top class in top.hpp file
-hier block constructor is called with wrong syntax in the top.cpp

### Relevant log output

```shell

```

amanchouhan01:
I can reproduce this issue on current main.
I’m looking into the C++ code generation path for hierarchical blocks and will propose a fix.

haakov:
@amanchouhan01 it looks like there is already a PR for this issue: https://github.com/gnuradio/gnuradio/pull/7740

marcusmueller:
@haakov  sadly not.

PR #7740, opened 2025-02-19
grc: fix wrong path for generated yml file when generating C++ hier b…

ZakyHermawan:
## Description

When generating C++ hier block, the yaml file should be in GRC_HIER_PATH or in the state directory,
instead of on the same directory as the grc file.
With this PR, the hier block now will be searchable by GRC after being generated and the GRC being refreshed.

Also, when you generate a C++ No GUI flow graph with hier block, the generated code will also contains
However, if the No GUI flow graph contains hier block, the generated code still can't be run. We still need to fix the templates and/or the python code that responsible for code generation. 

## Related Issue

Partially fix https://github.com/gnuradio/gnuradio/issues/7732

generator classes for C++

## Testing Done

Make flow graph with C++ Hier Block
Generate, reg
Check ~/.local/state/gnuradio (the YAML file for hier block should exist here)
The generated YAML file should also include cpp flag
```yaml
...
flags:
- cpp
...
```

## Checklist


marcusmueller:
> When generating C++ hier block, the yaml file should be in GRC_HIER_PATH or in the state directory,
instead of on the same directory as the grc file.

No, that's not correct.

PR #7742, opened 2025-02-21
replace 0, 0, 0 default argument with [0, 0, 0] in Vector Insert

jmfriedt:
Vector Insert: replace 0, 0, 0 default argument with [0, 0, 0]


marcusmueller (approved):
Looks good. We should probably grep through all ymls and check for similar issues

marcusmueller:
@jmfriedt thanks!

On a change this small, I could technically later "click away" the "has the committer signed-off their commit" check, but in case you want to beat me to it:

```shell
git commit --amend -s
git push -f
```

PR #7743, opened 2025-02-21
Fix MMSE resampler

### LINKED ISSUE #7253: Phase jump occurring in the MMSE resampler output
### What happened?

In ```mmse_fir_interpolator_cc::interpolate``` and ```mmse_resampler_cc_impl::general_work```

For a resample rate > 1.0, when ```d_mu``` increments above 2.0, the filter index ```imu``` wraps from NSTEPS to 0, the sample position will skip forward 2 samples, missing out two samples.
For a resample rate < 1.0, when  ```d_mu``` decrements below 1.0, the filter index``` imu``` wraps from 0 to NSTEPS, the sample position will skip backward 2 samples, repeating the last two samples.

These wrap around edge conditions cause an unwanted phase jump in the resampled data.

mmse_resampler_cc_impl.cc
The lines 87 and 104 should be
```int incr = 2 - (int)f;```

### System Information

Source code in the main repo

### GNU Radio Version

main (3.x)

### Specific Version

0.0.0.0

Put ```fprintf``` in the code to output the indices and samples.
See the log output for an example wrap around with resample rate < 1.0

### Relevant log output

```shell
index  incr    imu      d_mu    s - f   output           filter taps
2032	1	0	0.0019	0.0014	-251.42	-264.57  00001000
2033	1	0	0.0014	0.0009	-315.82	-10.57   00001000
2034	1	0	0.0009	0.0004	-332.1	223.49   00001000
2035	0	0	0.0004	0.9999	-302.02	421.08   00001000  <-- incr should be 2
2035	1	128	0.9999	0.9994	-332.1	223.49   00010000  <-- two repeated samples here
2036	1	128	0.9994	0.9989	-302.02	421.08   00010000
2037	1	128	0.9989	0.9984	-232.49	570.36   00010000
2038	1	128	0.9984	0.9979	-134.48	665.44   00010000
2039	1	128	0.9979	0.9974	-21.35	706.71
```

marcusmueller:
## Description

`mmse_resampler_{cc,ff}` had two immediate bugs:

1. regularly (as observed!) read input beyond ninput_items, because it was simply only restricting on noutput_items, and not noticing that you need input index + filter length to interpolate. This lead to hard-to-debug numerical errors.
2. at every work() call, reset of the index increment. That lead to phase jumps, if, and only if, the output item number was not a multiple of the filter length. Now, in a typical unit test flow graph, the GNU Radio scheduling meant that the number of output items was a power of two. The number of taps is 8.

To fix this, I first tackled `mmse_resampler_cc`, and then moved the mostly redundant `_cc` and `_ff` code to a template. That necessitated doing the same for `mmse_fir_interpolator`, to be able to use the right type of interpolator.

## Related Issue

Probably fixes #7253

Hopefully nothing.

## Testing Done

`qa_mmse_resampler` and `mmse_fir_interpolator` now pass under heavy load even with ~1000 repeats.

## Checklist


n-west:
By the way I suspect this might be related to #7253

gr-filter/lib/mmse_resampler_impl.cc
```diff
@@ -0,0 +1,142 @@
...
+            d_delta_mu = static_cast<double>(
+                reinterpret_cast<const float*>(input_items[1])[idx_in]);
+        }
+        const double s = d_mu + d_delta_mu;
+        const double f = floor(s);
+        d_mu = s - f;
+        d_delta_idx = (int)f;
+    }
+    this->set_inverse_relative_rate(d_delta_mu);
+    this->consume_each(idx_in);
```
daniestevez, line 104: I suggest to move `idx_in += d_delta_idx` to the end of the loop. This has the following advantages:

1. The loop condition is simplified to `idx_in < max_input_index`,  since now `idx_in` is not updated before being used in `d_resamp.interpolate()`.

2. The call `this->consume_each(idx_in)` is consuming some additional `d_delta_idx` samples that do no need to be preserved for the next `work()` call (since we would be skipping over them anyway).

3. The value of `d_delta_idx` does not need to be saved between `work()` calls (this is evident from how the code gets structured after these changes, since `d_delta_idx` is set before being read in the loop). Effectively, what achieves saving something between `work()` calls is that the value of `idx_in` in the `consume_each()` call is affected by the `d_delta_idx` of the last loop iteration when this change is done.

Comparing with the original code of the block, I see that it does pretty much what I am suggesting (except that the loop condition fails to account for `ninput_items[0]`, as noted in the PR description. Which leads me to say that I don't understand this bug:

> at every work() call, reset of the index increment. That lead to phase jumps, if, and only if, the output item number was not a multiple of the filter length.

Can you clarify what is wrong in the original implementation? My point is that the `incr` is affecting what is the first sample the the next `work()` call sees, since immediately before calling `consume_each(ii)` there is `ii += incr`.

gr-filter/lib/mmse_resampler_impl.cc
```diff
@@ -0,0 +1,142 @@
...
+
+template <typename sample_t>
+int mmse_resampler_impl<sample_t>::general_work(int noutput_items,
+                                                gr_vector_int& ninput_items,
+                                                gr_vector_const_void_star& input_items,
+                                                gr_vector_void_star& output_items)
+{
+    const sample_t* in = (const sample_t*)input_items[0];
+    sample_t* out = (sample_t*)output_items[0];
+    const auto max_input_index = ninput_items[0] - d_resamp.ntaps();
```
daniestevez, line 84: For the two-input case, is there something in the scheduler that ensures that `ninput_items[0] == ninput_items[1]`? If not, then probably this should be `std::min(ninput_items[0], ninput_items[1])` in this case.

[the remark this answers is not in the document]
marcusmueller: true; (mental check: the minimum number of input items to process exactly 1 item is indeed ntaps, not ntaps+1). Let me fix that!

gr-filter/lib/mmse_resampler_impl.cc
```diff
@@ -0,0 +1,142 @@
...
+        if (ninput_items.size() == 2) {
+            d_delta_mu = static_cast<double>(
+                reinterpret_cast<const float*>(input_items[1])[idx_in]);
+        }
+        const double s = d_mu + d_delta_mu;
+        const double f = floor(s);
+        d_mu = s - f;
+        d_delta_idx = (int)f;
+    }
+    this->set_inverse_relative_rate(d_delta_mu);
```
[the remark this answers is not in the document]
marcusmueller: correct.

gr-filter/lib/mmse_resampler_impl.cc
marcusmueller: oooh yes, indeed.

gr-filter/lib/mmse_resampler_impl.cc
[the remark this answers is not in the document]
daniestevez: Ah, good point. I had forgotten about the second input when I said this.

gr-filter/lib/mmse_resampler_impl.cc
marcusmueller: Sorry for the long delay in getting back to you!

explanation:

I can't consume `idx_in` after the loop if that happens to be `> ninput_items[0]`, so I had to move that consumption (and skipping) to the next call to `general_work`.

However, after fixing that, I fixed the filter oob access, rendering that problem obsolete. 

So, agree, will revert the reordering / split state.

marcusmueller:
Note personal nits being appreciated when they lead to increased code quality!

n-west:
My personal nits are on the single line conditionals. I like everything to have a curly brace, but plenty of gr code doesn't have that, this included

marcusmueller:
@n-west could you review https://github.com/gnuradio/gnuradio/pull/7743/commits/a67df0e6c10984c87d45d41963a99dc39cbdef74 ?

n-west:
@marcusmueller it's beautiful

PR #7747, opened 2025-02-22
pybind/building: don't error on deprecated symbols

marcusmueller:
By definition, pybind has to bind even deprecated functions (until we
actually remove them).

So, erroring out when we do that is not an option.

CI on maint-3.10 fails on F41/clang, because that notices how we're binding deprecated trellis/interleaver.

## Checklist


mbr0wn (approved):
This seems reasonable. I was afraid we were making this acceptable for all code, but as it's limited to the Pybind code I think it's OK.

PR #7749, opened 2025-02-23
blocks/annotator_raw: use set instead of vector to avoid sorting

### LINKED ISSUE #7624: annotator_raw: d_queued_tags is std::vector, gets sorted after every insertion, undergoes many element removals: wrong container
### What happened?

<meta>This is a **good first issue** for people who already know C++ and can figure out how to build GNU Radio on their machine.</meta>

checking the abuse of `std::sort` on tag vectors, I came across this. Sorting a vector every time you insert something screams "use a sorted container, not a `std::vector`".

Also, going forwards (n=0,…,N-1) through a N long vector and erasing the n. element causes N-1 copies of length N-1-n to happen, which happens to be in the order of N² copied elements… uhhhh…

In this case, a `std::set d_queued_tags` would do. 

Also, replace the `while(i!= d_queued_tags.end())` loop with a C++11 `for(const auto& tag: d_queued_tags)` loop.

The block has no unit test. This is also not good.

### System Information

n/a

### GNU Radio Version

3.11-git (main)

### Specific Version

f3127fce18f2363c978af63aec686a72da4cb042

n/a

### Relevant log output

_No response_

marcusmueller:
## Description

`annotator_raw` used a vector as queue for randomly inserted tags (i.e., things were re-sorted all the time). Changed that.

Worse, honestly, is that it had no test coverage. Changed that, too.

## Related Issue

Closes #7624

I thought #7624 was a good beginner's issue to fix, especially with how much work I put into describing the necessary changes. It seems it wasn't, really. Sigh.

annotator_raw

## Testing Done

added new unit test that also covers the temporal aspects
## Checklist

### DIFF
--- a/gr-blocks/lib/annotator_raw_impl.cc
+++ b/gr-blocks/lib/annotator_raw_impl.cc
@@ -8,12 +8,10 @@
  *
  */
 
-#ifdef HAVE_CONFIG_H
-#include "config.h"
-#endif
 
 #include "annotator_raw_impl.h"
 #include <gnuradio/io_signature.h>
+#include <cstdint>
 #include <cstring>
 #include <stdexcept>
 
@@ -42,20 +40,20 @@ void annotator_raw_impl::add_tag(uint64_t offset, pmt::pmt_t key, pmt::pmt_t val
     gr::thread::scoped_lock l(d_mutex);
 
     tag_t tag;
-    tag.srcid = pmt::intern(name());
+    tag.offset = offset;
     tag.key = key;
     tag.value = val;
-    tag.offset = offset;
+    tag.srcid = pmt::intern(name());
 
-    // add our new tag
-    d_queued_tags.push_back(tag);
-    // make sure our tags are in offset order
-    std::sort(d_queued_tags.begin(), d_queued_tags.end(), tag_t::offset_compare);
     // make sure we are not adding an item in the past!
-    if (tag.offset > nitems_read(0)) {
-        throw std::runtime_error(
-            "annotator_raw::add_tag: item added too far in the past.");
+    if (tag.offset < nitems_read(0)) {
+        throw std::runtime_error(fmt::format("annotator_raw::add_tag: item added too far "
+                                             "in the past at {}; we're already at {}.",
+                                             tag.offset,
+                                             nitems_read(0)));
     }
+    // add our new tag
+    d_queued_tags.insert(tag);
 }
 
 int annotator_raw_impl::work(int noutput_items,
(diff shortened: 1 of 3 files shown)


gr-blocks/lib/annotator_raw_impl.cc
```diff
@@ -42,20 +40,20 @@ void annotator_raw_impl::add_tag(uint64_t offset, pmt::pmt_t key, pmt::pmt_t val
...
-    // add our new tag
-    d_queued_tags.push_back(tag);
-    // make sure our tags are in offset order
-    std::sort(d_queued_tags.begin(), d_queued_tags.end(), tag_t::offset_compare);
     // make sure we are not adding an item in the past!
-    if (tag.offset > nitems_read(0)) {
-        throw std::runtime_error(
-            "annotator_raw::add_tag: item added too far in the past.");
+    if (tag.offset < nitems_read(0)) {
+        throw std::runtime_error(fmt::format("annotator_raw::add_tag: item added too far "
```
mbr0wn, line 50: Should this even be an exception? Or should we maybe drop those items? (and maybe log-warn?)

gr-blocks/lib/annotator_raw_impl.h
```diff
@@ -13,16 +13,39 @@
...
 namespace blocks {
 
 class annotator_raw_impl : public annotator_raw
 {
 private:
-    const size_t d_itemsize;
-    std::vector<tag_t> d_queued_tags;
+    struct tag_comparator {
+        using is_transparent = std::true_type;
+        //!\brief comparator over all fields in a tag, not just the offset
```
mbr0wn, line 28: Is that necessary? An offset compare is much cheaper than the other compares. The offset compare is clearly important. As for the others, if the `d_queued_tags` object stores tags with identical order in any offset, then that seems fine to me. So what happens if we just compare the offset in the comparator and the offset is the same? Seems to me that the `<` would just return false and handle it appropriately.

marcusmueller: (i.e., if we can't strictly order tags, we can't do things like "all tags with offsets between this and that"; we'd need to create a dummy tag just too look for things, or we again need to linearly search through the container of all tags, which negates the advantages of using a set instead a vector)

gr-blocks/lib/annotator_raw_impl.cc
marcusmueller: Hm, I think throwing here is fine, if you want to add a tag, you want to know whether that succeeded. Returning success or failure would make more sense, I agree, but if I change the behaviour of this block, this change can't be backported. We could make a separate PR to change the API to a more sensible `bool …::add_tag(uint64_t offset, pmt_t key, pmt_t val)` that doesn't throw, but it'd be a breaking change

PR #7754, opened 2025-03-03
gr-digital: fix early_late ted computation depth to match actual depth

### LINKED ISSUE #7752: early-late gate ted should have a computation depth of 3
### What happened?

The `ted_early_late` initializes with `inputs_per_symbol` of 2 and a `error_computation_depth` of 2.
https://github.com/gnuradio/gnuradio/blob/2acc77ab20ff7e8574076f33a859daccdd4eb2a1/gr-digital/lib/timing_error_detector.h#L351

The `error_computation_depth` is only used to create a the buffer of ted inputs (https://github.com/gnuradio/gnuradio/blob/2acc77ab20ff7e8574076f33a859daccdd4eb2a1/gr-digital/lib/timing_error_detector.cc#L208)

The ted_early_late error_computation uses the gardner style (early - late)*prompt https://github.com/gnuradio/gnuradio/blob/2acc77ab20ff7e8574076f33a859daccdd4eb2a1/gr-digital/lib/timing_error_detector.cc#L299

That directly indexes the d_input buffer that is error_computation_depth long. So d_input has a size() of 2, but is indexed with 0, 1, 2 (i.e., 3 elements required). The buffer should be 1 item longer since this dereference will always be out of bounds.

The computation_depth should be 3 to make this buffer the correct length.

### System Information

N/A. I only discovered this while reading code, and haven't observed bad behavior while running. I also haven't tested the ted_early_late to see if it works as is or not.

### GNU Radio Version

3.11-git (main)

### Specific Version

_No response_

I haven't observed whether this is problematic in runtime or not

### Relevant log output

```shell

```

n-west:
Fixes #7752

- [ X ] I have read the [CONTRIBUTING document](https://github.com/gnuradio/gnuradio/blob/main/CONTRIBUTING.md).
- [ X ] I have squashed my commits to have one significant change per commit. 
- [ X ] My code follows the code style of this project. See [GREP1.md](https://github.com/gnuradio/greps/blob/main/grep-0001-coding-guidelines.md).
- [ N/A ] I have updated [the documentation](https://wiki.gnuradio.org/index.php/Main_Page#Documentation) where necessary.
- [ N/A ] I have added tests to cover my changes, and all previous tests pass.

### DIFF
--- a/gr-digital/lib/timing_error_detector.h
+++ b/gr-digital/lib/timing_error_detector.h
@@ -348,7 +348,7 @@ public:
      * \brief Create a Early-Late timing error detector
      */
     ted_early_late()
-        : timing_error_detector(TED_EARLY_LATE, 2, 2, true, false, constellation_sptr())
+        : timing_error_detector(TED_EARLY_LATE, 2, 3, true, false, constellation_sptr())
     {
     }
     ~ted_early_late() override{};


marcusmueller:
Hey @n-west looks good but for the signoff :)

marcusmueller (approved):
https://github.com/gnuradio/gnuradio/blob/2acc77ab20ff7e8574076f33a859daccdd4eb2a1/gr-digital/lib/timing_error_detector.cc#L299

yeah, accessing `[2]` of a deque of size 2 is a bad idea.

n-west:
Added a sign-off. For what it's worth I did actually check and the early-late does synchronize at least in a simple case and doesn't seem substantially worse than gardner. My guess here is 
1) not many people use early-late gate because there is a better TED readily available for most cases and no tutorials I'm aware of are using early-late
2) the cppreference docs on a deque (https://en.cppreference.com/w/cpp/container/deque) make it sound like a linked list but has a feature of "On the other hand, deques typically have large minimal memory cost; a deque holding just one element has to allocate its full internal array (e.g. 8 times the object size on 64-bit libstdc++; 16 times the object size or 4096 bytes, whichever is larger, on 64-bit libc++)." Which sounds like maybe there's some minimum allocation for pointer alignment occurring and the usage here does a set of pointer reuse which makes d_input[2] always some not-what-you-wanted but sorta valid value.

marcusmueller:
> …  Which sounds like maybe there's some minimum allocation for pointer alignment occurring and the usage here does a set of pointer reuse which makes d_input[2] always some not-what-you-wanted but sorta valid value.

Oh the deque is an interesting creature.

Think of it as an nullptr-terminated array of pointers to fixed-size in-memory blocks (confusingly called "map").

There's also a a "first element" pointer pointing to the … first element in the first used block, and a "after last element" pointer pointing to after the last used element in the last block, or if that block is fully used, some sentinel value.

- Want to append an element, and the last block isn't full? move the "behind last" pointer by one element.
- want to append, but last block full? Allocate new block, use first element position in new block, change "behind last" to point to second element in new block
- want to prepend, but starting at beginning of first block? move all the pointers to the right, allocate new block, insert pointer to freshly allocated block at first position, put new element into last slot of that block, make "first element" pointer point to that element.

So, double pointer lookup because you first need to look up the address of the list of blocks, and then look up the address of the block with your element of interest in it, then add `(number%block_size)*sizeof(element)` to that to find the element.

It's thus really at an interesting point in the "fast insert on both ends" vs "efficient lookup" trade-off space. Note that different compilers used to have **wildly** different block sizes, making, IIRC, deque slow on MSVC even for objects as small as complex float, and the cleanup operations on GCC when popping elements somewhat reminescent of the Java-style "wait till garbage collection lets us do work" annoyance. (note that the last time I played around with this was… 2010-ish???? I bet both GCC and Java have gotten less bad.)

PR #7767, opened 2025-03-11
fec: make LDPC GF(2) use uchar matrix, not double matrix

marcusmueller:
## Description

modernize code

Add FIXMEs where an overhaul is due, but can't be done in short time.

A lot of this reads like C89 code, not C++03, which it should have been
at the time of addition.

gsl also has a pretty antique API (and API paradigms), for a modern C
library.

We were doing a lot of stupid stuff, like copying struct definitions
without explicitly saying so (or having any asserts the structs where
identical).

We were also using double-valued matrices to store GF(2) (read: binary)
LDPC matrices. I could replace these with unsigned char matrices, but I
couldn't, on short notice, write a good GF(2) matrix inverse. The
interested reader should feel encouraged to write one as home exercise.
(Seriously, this is GF(2) represented with GF(2⁸) elements. Just
Gaussianizing the matrix might be pretty darn better than finding the
inverse of a real-valued matrix, then casting that (truncatingly) to
int, then discarding all of that int but the least bit, and than saving the result in a
real-valued matrix.)

my sanity?

## Testing Done

Tests pass

We modify a public header, but that just contains implementation details we leak as opaque pointers in Python.

## Checklist


daniestevez:
> Seriously, this is GF(2) represented with GF(2⁸) elements. Just
Gaussianizing the matrix might be pretty darn better than finding the
inverse of a real-valued matrix, then casting that (truncatingly) to
int, then discarding all of that int but the least bit, and than saving the result in a
real-valued matrix.

Without looking at the code, that can't possibly work. If you treat a matrix of 0's and 1's as real-valued an invert it, typically the result will have rational entries, not just integers. And truncating those to int is disastrous. The trick that actually works is:

1. Invert your matrix as if it was real-valued.
2. Multiply the result by the determinant of the original matrix.
3. On paper, you are guaranteed to get a matrix with integer entries. In practice not, due to floating-point arithmetic, but you can round (not truncate!).
4. Reduce the matrix modulo 2. The result is the same as the inverse of the original matrix over GF(2).

This might be what the code is actually doing.

I agree that this is just a trick (I use it with NumPy, for instance), and that the proper way to do this is to implement Gaussian elimination. Here is the [Gaussian elimination implementation from ldpc-toolbox](https://github.com/daniestevez/ldpc-toolbox/blob/main/src/linalg.rs#L8). It's Rust, so we cannot copy & paste it here. I'm mostly leaving it as pseudocode in case someone wants to write the same in C++, and also to demistify the algorithm and show that it's quite simple and you don't need any linear algebra libraries unless you care a lot about CPU performance.

gr-fec/lib/fec_mtrx_impl.cc
```diff
@@ -148,39 +142,43 @@ void write_matrix_to_file(const std::string filename, matrix_sptr M)
 
 matrix_sptr generate_G_transpose(matrix_sptr H_obj)
 {
-    unsigned int k = H_obj->size1;
-    unsigned int n = H_obj->size2;
-    unsigned int row_index, col_index;
+    const unsigned int k = H_obj->size1;
+    const unsigned int n = H_obj->size2;
```
daniestevez, line 146: I'm not familiar with GSL matrix objects. Before, `H_obj` was a `struct` defined by GNU Radio which contained `size1` and `size2` fields, so this code makes sense. I guess that a `gsl_matrix_uchar` also contains these fields under the hood, but is the user supposed to access the fields, or is it better to treat the struct as an opaque object and use getter functions?

gr-fec/lib/fec_mtrx_impl.cc
```diff
@@ -148,39 +142,43 @@ void write_matrix_to_file(const std::string filename, matrix_sptr M)
...
-        for (col_index = 0; col_index < k; col_index++) {
-            int value = gsl_matrix_get((gsl_matrix*)(H_obj.get()), row_index, col_index);
-            gsl_matrix_set(P_transpose, row_index, col_index, value);
+    gsl_matrix_uchar* P_transpose = gsl_matrix_uchar_alloc(n - k, k);
+    for (unsigned int row_index = 0; row_index < n - k; row_index++) {
+        for (unsigned int col_index = 0; col_index < k; col_index++) {
+            gsl_matrix_uchar_set(P_transpose,
+                                 row_index,
+                                 col_index,
+                                 gsl_matrix_uchar_get(H_obj.get(), row_index, col_index));
```
daniestevez, line 157: So despite what the comment says, there isn't a transpose operation being done here (the comment might still make sense depending on how the notation for the blocks of the generator and parity check matrix is defined).

This code is basically copying out a submatrix: `P_transpose[row_index, col_index] = H_obj[row_index, col_index]`. Might there be a way to write this more compactly with GSL functions? From a quick look at the [docs](https://www.gnu.org/software/gsl/doc/html/vectors.html), I don't see it, but it also seems to me that copying into `P_transpose` is not necessary. A "matrix view" might be enough, or even using `H_obj` instead.

gr-fec/lib/fec_mtrx_impl.cc
```diff
@@ -148,39 +142,43 @@ void write_matrix_to_file(const std::string filename, matrix_sptr M)
...
     // Set G transpose matrix (used for encoding)
-    gsl_matrix_set_zero(G_transp);
-    for (row_index = 0; row_index < k; row_index++) {
-        col_index = row_index;
-        gsl_matrix_set(G_transp, row_index, col_index, 1);
+    gsl_matrix_uchar_set_zero(G_transp);
+    for (unsigned int row_index = 0; row_index < k; row_index++) {
+        const auto col_index = row_index;
+        gsl_matrix_uchar_set(G_transp, row_index, col_index, 1);
     }
```
daniestevez, line 166: Nit. I would write this as follows. It makes it even more obvious that we are only accessing the diagonal.
```suggestion
    gsl_matrix_uchar_set_zero(G_transp);
    for (unsigned int row_index = 0; row_index < k; row_index++) {
        gsl_matrix_uchar_set(G_transp, row_index, row_index, 1);
    }
```
Also, it looks like all this thing could be replaced by [`gsl_matrix_set_identity`](https://www.gnu.org/software/gsl/doc/html/vectors.html#c.gsl_matrix_set_identity).

marcusmueller:
@daniestevez yes, invert as if real, scale by determinant, that's what the original code did, but after that:

```c++
for (row_index = 0; row_index < n; row_index++) {
  for (col_index = 0; col_index < n; col_index++) {
    float value = gsl_matrix_get(matrix_inverse, row_index, col_index);
    // take care of mod 2
    int value_cast_as_int = static_cast<int>(value);
    int temp_value = abs(fmod(value_cast_as_int, 2));
    gsl_matrix_set(matrix_inverse, row_index, col_index, temp_value);
  }
}
```

:)

gr-fec/lib/fec_mtrx_impl.cc
```diff
@@ -277,100 +273,111 @@ fec_mtrx_impl::fec_mtrx_impl()
...
-        for (col_index = 0; col_index < cols; col_index++) {
-            int value = gsl_matrix_get(result, row_index, col_index);
-            int new_value = value % 2;
-            gsl_matrix_set(result, row_index, col_index, new_value);
+    for (unsigned int row = 0; row < n_rows_res; row++) {
+        for (unsigned int col = 0; col < n_cols_res; col++) {
+            unsigned char sum = 0;
+            for (unsigned int idx = 0; idx < n_cols1; idx++) {
+                sum += gsl_matrix_uchar_get(matrix1, row, idx) *
+                       gsl_matrix_uchar_get(matrix2, idx, col);
```
daniestevez, line 348: I would suggest some care here, since `sum` can overflow. Probably this is not going to happen for LDPC matrices, because they have few 1's (that what "LD" means), but it can happen for dense generator matrices (I'm not really paying attention to which kinds of matrices actually need to be passed to this function; I'd rather have a multiply function that works for all GF(2) matrices unless they are very large). So I would use a `uint32_t` at least for `sum`.

gr-fec/lib/fec_mtrx_impl.cc
```diff
@@ -396,42 +403,56 @@ gsl_matrix* fec_mtrx_impl::calc_inverse_mod2(const gsl_matrix* original_matrix)
...
-            int value_cast_as_int = static_cast<int>(value);
-            int temp_value = abs(fmod(value_cast_as_int, 2));
+            // FIXME: we cast to int, which will be problematic when something is
+            // x.9999998; then we do an fmod on that int; then we do an abs on the float
+            // that comes out of that, which we then truncate to int.
+            //
+            // Not many ways this could be less efficient, and more numerically
+            // questionable.
+            const int value_cast_as_int = static_cast<int>(value);
+            const int temp_value = abs(fmod(value_cast_as_int, 2));
```
daniestevez, line 419: Can we go and fix this in this PR? I agree that it is not a long-term solution, but this is a bug. It's probably not too difficult to find matrices for which this gives the wrong inverse because of the flooring.

I would do something like `uint8_t temp_value = static_cast<int>(std::round(value)) % 2`, but I never remember how `% 2` (or `& 1`) work for negative numbers. This might need an `abs`.

gr-fec/lib/fec_mtrx_impl.cc
```diff
@@ -396,42 +403,56 @@ gsl_matrix* fec_mtrx_impl::calc_inverse_mod2(const gsl_matrix* original_matrix)
...
+            const unsigned char value = gsl_matrix_get(matrix_inverse, row, col);
+            max_value = std::max(max_value, value);
+            gsl_matrix_uchar_set(matrix_inv_gf2, row, col, value & 1);
+        }
+    }
 
-    int max_value = gsl_matrix_max(matrix_inverse);
     if (!max_value) {
         throw "Error in calc_inverse_mod2(): The matrix inverse found is all zeros.";
     }
```
daniestevez, line 436: Can this even happen? Setting aside rounding problems in the above code, the only way in which I see this happening is if the inverse has come out as all zeros (or something very close to all zeros). And that should never happen. I would expect `gsl_linalg_LU_invert` to complain if it can't invert, rather than producing an all-zero matrix.

gr-fec/lib/fec_mtrx_impl.cc
```diff
@@ -396,42 +403,56 @@ gsl_matrix* fec_mtrx_impl::calc_inverse_mod2(const gsl_matrix* original_matrix)
...
             gsl_matrix_set(matrix_inverse, row_index, col_index, temp_value);
         }
     }
+    gsl_matrix_uchar* matrix_inv_gf2 = gsl_matrix_uchar_alloc(n, n);
+    unsigned char max_value = 0;
+    for (unsigned row = 0; row < n; ++row) {
+        for (unsigned col = 0; col < n; ++col) {
+            const unsigned char value = gsl_matrix_get(matrix_inverse, row, col);
+            max_value = std::max(max_value, value);
+            gsl_matrix_uchar_set(matrix_inv_gf2, row, col, value & 1);
```
daniestevez, line 430: I think the `& 1` here is redundant because the code above should have only produced 0's and 1's.

gr-fec/lib/fec_mtrx_impl.cc
```diff
@@ -396,42 +403,56 @@ gsl_matrix* fec_mtrx_impl::calc_inverse_mod2(const gsl_matrix* original_matrix)
...
         throw "Error in calc_inverse_mod2(): The matrix inverse found is all zeros.";
     }
 
     // Verify that the inverse was found by taking matrix
     // product of original_matrix and the inverse, which should
     // equal the identity matrix.
-    gsl_matrix* test = gsl_matrix_alloc(n, n);
-    mult_matrices_mod2(test, original_matrix, matrix_inverse);
+    gsl_matrix_uchar* test = gsl_matrix_uchar_alloc(n, n);
+    mult_matrices_mod2(test, original_matrix, matrix_inv_gf2);
```
daniestevez, line 442: This verification shouldn't be necessary. Maybe it's here to catch weird cases because of bad rounding, but this check can definitely go away when the inverse over the real field is replaced by an inverse over GF(2) (Gauss reduction, etc.).

gr-fec/lib/ldpc_G_matrix_impl.cc
```diff
@@ -49,83 +49,94 @@ ldpc_G_matrix_impl::ldpc_G_matrix_impl(const std::string filename) : fec_mtrx_im
...
         }
     }
 
     // Check if the identity matrix exists in the right spot.
-    // int test_if_equal = gsl_matrix_equal(identity, I_test);
-    gsl_matrix_sub(identity, I_test); // should be null set if equal
-    double test_if_not_equal = gsl_matrix_max(identity);
+    // int test_if_equal = gsl_matrix_uchar_equal(identity, I_test);
+    gsl_matrix_uchar_sub(identity, I_test); // should be null set if equal
+    double test_if_not_equal = gsl_matrix_uchar_max(identity);
```
daniestevez, line 69: It looks like now this should be `uint8_t` or `auto` rather than `double`.

gr-fec/lib/ldpc_G_matrix_impl.cc
```diff
@@ -24,21 +22,23 @@ ldpc_G_matrix::sptr ldpc_G_matrix::make(const std::string filename)
...
-
     // Read the matrix from a file in alist format
     matrix_sptr x = read_matrix_from_file(filename);
     d_num_cols = x->size2;
     d_num_rows = x->size1;
 
+    // FIXME: totally misled. If we can't read x safely here, then we also can't copy it
+    // savely. Don't make that copy. All we do is checking the first d_n×d_n submatrix for
+    // identity-matrixness, then transposing and storing the result.
+    //
```
daniestevez, line 36: I don't understand this comment accurately, even after having read through the code below. Chances are that the next person that stumbles upon this comment, maybe years from now, will find it a bit cryptic too. Can we try to reword?

What is the problem that the comment tries to describe? Is it that `read_matrix_from_file()` might fail, in which case the function should exit early without trying to do anything? I agree that the check for whether G is of the form [I P] is bad code, but there is a FIXME comment specific to that bad code, and the bad code is technically correct (in the sense that it doesn't have bugs and produces correct results).

gr-fec/lib/ldpc_G_matrix_impl.cc
```diff
@@ -49,83 +49,94 @@ ldpc_G_matrix_impl::ldpc_G_matrix_impl(const std::string filename) : fec_mtrx_im
...
-        for (col_index = 0; col_index < d_k; col_index++) {
-            int value = gsl_matrix_get(P_transpose, row_index, col_index);
-            gsl_matrix_set(H_ptr, row_index, col_index, value);
+    gsl_matrix_uchar* H_ptr = gsl_matrix_uchar_alloc(d_n - d_k, d_n);
+    gsl_matrix_uchar_set_zero(H_ptr);
+    for (unsigned int row_index = 0; row_index < d_n - d_k; row_index++) {
+        for (unsigned int col_index = 0; col_index < d_k; col_index++) {
+            int value = gsl_matrix_uchar_get(P_transpose, row_index, col_index);
+            // FIXME: this could – without hurting legibility – just directly read the
+            // correct values, transposing on the way
```
daniestevez, line 114: I fully agree!

gr-fec/lib/ldpc_H_matrix_impl.cc
```diff
@@ -182,38 +179,39 @@ void ldpc_H_matrix_impl::back_solve_mod2(gsl_matrix* x,
...
+            } else {
+                // x[i] = 1 and U[i][i] = 0 -> a/0 = undef
+                gr::logger("ldpc_H_matrix_impl")
+                    .error("Error in back_solve_mod2, division between x[{0}]={1} and "
+                           "U[{0}]={2} not defined.",
+                           i,
+                           x_i,
+                           U_i_i);
+            }
+        }
```
daniestevez, line 214: To me it seems that this won't catch `0/0`. I would do it as follows:

1. If `U_i_i == 0`, then either log an error or throw an exception. I would need to think more carefully about in which context this function is being run and whether failure to divide should be a fatal error (for instance, can it happen normally as a decoding failure with a codeword with many bit errors, or mathematically it should never happen and if we see it then something has gone deeply wrong?).

2. Do nothing. Technically we are supposed to set `x[i, 0]` equal to `x[i, 0] / U[i, i]`, but since `U[i, i]`, must be one, the value of `x[i, 0]` won't change and we don't even need to call `gsl_matrix_uchar_set()`.

daniestevez (commented):
These changes generally look good. The only thing that I spotted that needs fixing is a potential overflow in the matrix product (this is a regression). Take or ignore the rest of the suggestions as you see fit.

gr-fec/lib/fec_mtrx_impl.cc
marcusmueller: agreed, let's use the dedicated function

gr-fec/lib/fec_mtrx_impl.cc
marcusmueller: I think since all that matters in the end is `sum mod 2`, that's not a problem?

gr-fec/lib/fec_mtrx_impl.cc
marcusmueller: yes, we can; I was avoiding putting it in the same commit, but will fix that this weekend.

gr-fec/lib/fec_mtrx_impl.cc
marcusmueller: yeah, I didn't see a case either; my guess is this is a remnant from developing the mechanism, when some mistake did lead to this, and this check actually helped identify the issue. Didn't want to remove security mechanisms, but seeing that we calculate `max_value` in vain, hm.

I think I'll leave it in for now and add a TODO to remove it when someone actually implements a "native" GF2 matrix inverse.

gr-fec/lib/fec_mtrx_impl.cc
marcusmueller: Yes, definitely calls for a `//TODO: rm when …`

gr-fec/lib/ldpc_G_matrix_impl.cc
marcusmueller: what is done here is copying the input matrix into a temporary, specifically allocated matrix within this function, "because we can't know what someone else might be doing with the pointer-passed matrix". 

Well, if we can't know that the matrix is not modified while this function works on it, we have a severe multithreading bug. Copying the matrix at the beginning reduces the time window when that might manifest as data corruption, but it's not an appropriate mitigation.

gr-fec/lib/ldpc_H_matrix_impl.cc
marcusmueller: I think 2. is the right way regarding what we should do to `x`. 

To explain why I did that change: I looked at the if/elseifs and noticed that it always sets the new value to the old value, unless the the old value `x_i` is 0 and `U_i_i` is 1, in which case it just errors.

So, you're right, we don't ever need to set things; this could handle `x` as an immutable vector. :exploding_head:

gr-fec/lib/fec_mtrx_impl.cc
daniestevez: You are right. But then, why don't we write this as the correct operations in GF(2)? Like so:
```c++
                sum ^= gsl_matrix_uchar_get(matrix1, row, idx) &
                       gsl_matrix_uchar_get(matrix2, idx, col);
```
I don't know if this allows the compiler to perform additional optimizations, but to me just the fact that this makes the intent of operating in GF(2) clear makes this approach preferable.

gr-fec/lib/fec_mtrx_impl.cc
daniestevez: For me the problem is not remembering how 2's complement work, but the fact that C and C++ have nasty surprises whenever signed integers are involved (specially if you start doing bitwise operations with them).

In this case, it seems that `& 1` does the desired thing (just a bitwise AND), while `% 2` does something else: https://godbolt.org/z/PvW3EnGj1

gr-fec/lib/ldpc_G_matrix_impl.cc
daniestevez: I see where you're going, but I still don't understand this (either your comment or the comment in the original code that says `Make an actual copy so we guarantee that we're not sharing memory with another class that reads the same alist file.`). Assuming that the file itself isn't being altered while it is being read, and unless I've missed something. I don't see any potential problems and no need to copy the matrix. `read_matrix_from_file()` allocates a new matrix and then passes ownership of that to the caller by returning a sptr. There is apparent problem if there are concurrent calls to `read_matrix_from_file()` on different threads, whether they read the same alist file or a different one (why would it matter if the alist file is the same? it is just a path that gets read).

gr-fec/lib/ldpc_H_matrix_impl.cc
daniestevez: > So, you're right, we don't ever need to set things; this could handle x as an immutable vector.

Well no, because somewhere else in this code there is the following operation:
```
// x[i, 0] = x[i, 0] + U[i, j] * x[j, 0]
```
We are doing Gauss reduction in place, after all, so `x` needs to be mutated in some way.

The main point of this comment was that division in GF(2) is a no-op, unless you are trying to divide by 0. And you do need to check for division by 0, because that will eventually happen if you try to reduce a matrix that is not invertible (which can happen if a bad alist is given).

gr-fec/lib/ldpc_G_matrix_impl.cc
marcusmueller: exactly! the "Make an actual copy…" is from the original code. Will rephrase that to be more explictly a `FIXME: remove copy". My interpretation of what the original author got wrong about ownership doesn't really matter.

PR #7773, opened 2025-03-20
grc: Accept failures in message sending

mbr0wn:
This fixes a rare corner case seen in unit tests, where multiple tests interfere with one another.

During the execution of unit tests for the pluggable workflow PR, this sometimes happened. I couldn't quite track it down, but when you run multiple GRC-related tasks in quick succession in a unit test, the output messenger would fail to open-and-close the file descriptor.

## Checklist

### DIFF
--- a/grc/core/Messages.py
+++ b/grc/core/Messages.py
@@ -5,8 +5,9 @@
 #
 
 
-import traceback
+import logging
 import sys
+import traceback
 
 #  A list of functions that can receive a message.
 MESSENGERS_LIST = list()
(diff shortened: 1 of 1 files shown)


grc/core/Messages.py
```diff
@@ -40,7 +40,10 @@ def send(message):
         message: a message string
     """
     for messenger in MESSENGERS_LIST:
-        messenger(_indent + message)
+        try:
+            messenger(_indent + message)
+        except:
+            pass
```
marcusmueller, line 46: I feel like we should be at least logging this? Otherwise, we'll have hard-to-debug problems if for any reason in the future, a crucial messenger permanently fails.

marcusmueller:
Just for understanding: which messenger is affected here? stdout.write? Could you add that to your commit msg, please to avoid future confusion :)

grc/core/Messages.py
mbr0wn: You mean, using the regular Python logger instead of the messenger logger? Sure we can add that.

mbr0wn:
> Just for understanding: which messenger is affected here? stdout.write? Could you add that to your commit msg, please to avoid future confusion :)

✔️

marcusmueller (approved):

PR #7775, opened 2025-03-20
grc: Make generate aware of execution

mbr0wn:
When generating a flow graph, we make the generator aware if it's being called directly, or implied by "execute". This can be useful for future workflows.

## Related Issue

In the RFNoC workflow that I built, the "generate" step is pretty expensive. Because calling "execute" always implies calling "generate", it's good to know where that came from. You can see more info by looking at the RFNoC workflow in #7548.

This change by itself doesn't actually change anything, it just lays down groundwork.

## Testing Done

The RFNoC workflow is a good test case, but the other tests and GRC functions still work with this as before.

## Checklist

### DIFF
--- a/grc/core/generator/top_block.py
+++ b/grc/core/generator/top_block.py
@@ -64,7 +64,7 @@ class TopBlockGenerator(object):
         for key in deprecated_block_keys:
             Messages.send_warning("The block {!r} is deprecated.".format(key))
 
-    def write(self):
+    def write(self, _=None):
         """generate output and write it to files"""
         self._warnings()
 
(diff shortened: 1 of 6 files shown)


grc/core/generator/top_block.py
```diff
@@ -64,7 +64,7 @@ def _warnings(self):
         for key in deprecated_block_keys:
             Messages.send_warning("The block {!r} is deprecated.".format(key))
 
-    def write(self):
+    def write(self, _=None):
```
mbr0wn, line 67: In case it's not clear: The implementer of a workflow can make use of the source of generation here, but doesn't have to (and the regular, core workflows are not changed).

haakov (approved):

marcusmueller:
@mbr0wn @haakov I've not looked at backportability; if you think this should be backported and doesn't break API, add to the backports-3.10 project as "todo" please.

haakov:
Added to backports-3.10 as todo.

PR #7809, opened 2025-04-03
blocks: add Reinterpret Cast block

rsekman:
## Description
This pull request implements a **Reinterpret Cast** block, which performs
```cpp
output[i] = reinterpret_cast<output_t*>(input)[i].
```
The motivation for this block can be seen in the example flowgraph:
Example flowgraph: a Socket PDU block is connected to a Stream to PDU Block, to a Reinterpret Cast block changing the byte stream to a short stream, which goes into an IShort to Float block, finally being consumed by a QT Constellation Sink.
With the Socket PDU block, we can only receive a stream of bytes. But suppose that, actually, the data is interleaved shorts, or complex floats. In C++, this is what `reinterpret_cast` is for. With existing blocks, I could not find an ergonomic way to tell GNU Radio to leave the bytes untouched, but treat them as another type.

The block is essentially a Copy block, except it allows for unequal input and output sizes, and does the necessary maths to account for that. 

No previously existing blocks have been changed.

## Testing Done
I have written tests for `reinterpret_cast` for the cases `{char, short, int} <-> {char, short, int}` and `{char, short, int} -> float`.

## Checklist

Some documentation should probably be added for the block. I'd appreciate pointers to the relevant places in the repo and tips for writing it.


gr-blocks/lib/recast_impl.cc
```diff
@@ -0,0 +1,65 @@
...
+    // Ensure we always consume entire input items
+    if (noutput_items < d_input_size / d_output_size)
+        return 0;
+
+    const uint8_t** in = (const uint8_t**)&input_items[0];
+    uint8_t** out = (uint8_t**)&output_items[0];
+    noutput_items = std::min(noutput_items * d_input_size,
+                             (ninput_items[0] * d_input_size) / d_output_size);
+
+    memcpy(out[0], in[0], noutput_items * d_output_size);
```
marcusmueller, line 53: this can't work, if `output_size > input_size`, this reads beyond the end of input

marcusmueller:
I'm afraid this block makes no sense: 

1. GNU Radio 3.x doesn't care about types anyway. If it has the right itemsize, it connects.
2. you just memcopy a region over from input to output. That is fundamentally different than `for(unsigned int idx = 0; idx < total_number; ++idx) out[idx] = whatever_cast<>(input[idx])`.  <br>(make a sketch about what happens when your block gets say 100 input items of `char`, and copies that over to an output buffer with enough space for 100 output items of `short`)

rsekman:
> I'm afraid this block makes no sense:
> 
>  1. GNU Radio 3.x doesn't care about types anyway. If it has the right itemsize, it connects. I must admit I fail to see when I'd need a `reinterpret_cast`-alike functionality anywhere in there: we already have all the type safety of casting to `void*`. 😬

How would you implement the example flowgraph? `socket_pdu` emits only `char` PDU:s, so the Stream to PDU block has to have output size 1. If we want to connect that to something of input size != 1, we need an adaptor block!

>you just memcopy a region over from input to output. That is fundamentally different than
for(unsigned int idx = 0; idx < total_number; ++idx) out[idx] = whatever_cast<output_type>(input[idx]).

`reinterpret_cast` does no modifications to the bit pattern, unlike `static_cast` and `dynamic_cast`. The point is to get the same bits, but *reinterpreted* with a different item size. https://en.cppreference.com/w/cpp/language/reinterpret_cast

> Unlike static_cast, but like const_cast, the reinterpret_cast expression does not compile to any CPU instructions (except when converting between integers and pointers, or between pointers on obscure architectures where pointer representation depends on its type). It is primarily a compile-time directive which instructs the compiler to treat expression as if it had the type target-type. 

Because GNU Radio's architecture is buffer-to-buffer we can't get to no CPU instructions, only the closest thing of `memcpy`.

> (make a sketch about what happens when your block gets say 10 input items of char, and copies that over to an output buffer with enough space for at least 10 output items of short.

Right, the output buffer will contain 10 bytes, which it will see as 5 shorts.

> Then do the same sketch again, but this time general_work is called first with 6 items, and then with the remaining 4 items of the same 10 items input. Compare the output, byte-wise. See your mistake?)

I do not see my mistake, because the tests I wrote pass :)

> or example, this should have been a fixed-rate block (not a block with general_work and forecast), 

The way I understood the class hierarchy a fixed rate block is either an interpolator or a decimator, but this block can be either depending on which of the input and output sizes is larger. I did a draft dynamically returning either one from `make()`, but found that the code just got longer and harder to follow.

> and it should have had a different name, because nowhere in your code you do an actual recast.

The proper name is `reinterpret_cast` but that's a C++ keyword. The user-facing name is Reinterpret Cast.

marcusmueller:
>If we want to connect that to something of input size != 1, we need an adaptor block!

as I said, stream to vector / vector to stream

>reinterpret_cast does no modifications to the bit pattern, unlike static_cast and dynamic_cast. The point is to get the same bits, but reinterpreted with a different item size. https://en.cppreference.com/w/cpp/language/reinterpret_cast

yes, your implementation is broken, as said. 

> Right, the output buffer will contain 10 bytes, which it will see as 5 shorts.

Yes, but no! Where are the bytes with input data in the 10 bytes? 

> I do not see my mistake, because the tests I wrote pass :)

They are not sufficient, because the amount of input samples is too small for it to require multiple calls to `work()`.

> The proper name is reinterpret_cast but that's a C++ keyword. The user-facing name is Reinterpret Cast.

it doesn't do that, really.

rsekman:
> "for each input item, pad it to the larger output item's size"

That is not what the block does,  that is not what `reinterpret_cast` does in C++, and that is not the behaviour I need to make the example flowgraph work. You are describing, if anything, `static_cast`. I want `reinterpret_cast`.

>again, GNU Radio just doesn't know about types, it only knows about sizes)

The companion disagrees! 
image
Sure, I can group the bytes 8 at a time with a Stream to Vector , but at least in the companion that can't then be connected to something that wants complex input.

The only blocks I could find that don't have the same input and output types are  the Type Conversion blocks, e.g. Int to Float. But if I give Int to Float the bytes `0x40490fdb`, the bytes that come out are `0x4e809220`, which is the IEEE 754 representation of the integer `1078530048` (which isn't the same as `0x40490fdb`, but that's because it's large enough to be past where floats have integer resolution) . But what I want is the bytes `0x40490fdb`, which is the IEEE 754 representation of π. I simply cannot find a way to implement the example flowgraph in a way that the companion's type checking accepts.

> For "drop M out of N input item sizes bytes each input item": that can be done using Keep M in N.

I want to *keep every bit*. 

> Yes, but no! Where are the bytes with input data in the 10 bytes?

In the ten first bytes, where they should be! Because `reinterpret_cast` *does not change the bit pattern.* I don't want 10 shorts, I want 5 shorts made up of 10 bytes.

> They are not sufficient, because the amount of input samples is too small for it to require multiple calls to work().

I have used the block to process streams of hundreds of megabytes (granted, only in char -> int mode).

ffu:
Also, don't forget the incredibly useful and flexible PDU Lambda block. The desired conversion can be solved by using this block with parameter lambda x: x.view("short")

This will convert the char pdu to a short pdu.

PR #7810, opened 2025-04-04
socket_pdu: support multicast for UDP Server mode

rsekman:
## Description
Previously the `socket_pdu` only worked in unicast mode. By checking if the address provided in UDP Server mode is a multicast address, and if so joining that multicast group, we can also receive multicast data.

Only `socket_pdu`.

## Testing Done
I have used the following flowgraph and shell script to send and receive data over UDP multicast.
image
```sh
#!/usr/bin/bash
head --bytes 8k </dev/urandom >in.bin
./multicast_test.py < <(sleep 5) >/dev/null &
# Allow the GNU Radio runtime some time to start
sleep 2
nc -q5 -u 239.0.0.10 50500 < in.bin
sha256sum in.bin out.bin
```

## Checklist


marcusmueller:
Hi @rsekman! Thanks!

Stupid question, because I need to figure out whether this can be backported to existing release series: 

If someone enters a multicast IP address right now, it just wouldn't work, because no host can be the owner of that IP address, right?

I ask because otherwise this might be a behavioural change under some circumstances.

rsekman:
Hi @marcusmueller, thanks for taking a look at this so quickly. That depends on what you mean by wouldn't work.

Right now, entering a multicast address means no data can be received. Running the test I described on main results in `out.bin` being an empty file. So the block doesn't work in the sense that it will not receive multicast data. But it does work in the sense that it doesn't throw exceptions.

This is legal:
```c
struct sockaddr_in addr;
...
int s = socket(AF_INET, SOCK_DGRAM, 0);
bind(s, &addr, sizeof(addr));
```
even if `addr` specifies a multicast address. The call to `bind()` succeeds. But without a call like
```c
struct ip_mreq mreq;
...
setsockopt(s, IPPROTO_IP, IP_ADD_MEMBERSHIP, &mreq)
```
with `mreq` specifying the same multicast address, the socket will never receive any data.
In short, for multicast two operations are required to receive data: binding to an address (possibly `INADDR_ANY`), and joining a multicast group; they can occur in either order.  A program that skips joining the multicast group *runs* but likely does not *work as intended*, i.e., has a bug.

So, if someone out there has entered a multicast address, there would be a behavioural change in that with this PR they could start receiving data they wouldn't before. 

I considered gating this behind a bool argument to the block ("Enable multicast? default = false"), but that's an API break, and it would only apply in certain configurations of the block anyway. (It'd have no effect on TCP operations.)

marcusmueller:
Ah an optional argument wouldn't be an API break (not even an ABI break, usually), because a program that doesn't know the new argument still compiles with the new signature.

However, let's put ourselves in the shoes of someone who binds to a multicast address with the old code: stuff's broken. We fix this. That's not a break of functional compatibility! So: adding this to the list of things to backport.

marcusmueller:
I'd have a request though: Could you add both your flowgraph and the multicast_test.py to a new directory, gr-network/examples/multicast_udp, please?

rsekman:
> I'd have a request though: Could you add both your flowgraph and the multicast_test.py to a new directory, gr-network/examples/multicast_udp, please? An illustration of this functionality makes it much more valuable to potential future users :)

Done.

rsekman:
> Ah an optional argument wouldn't be an API break (not even an ABI break, usually), because a program that doesn't know the new argument still compiles with the new signature.

Ah, I wasn't sure about that; my understanding was the compiler only generates one signature and fills in the default arguments at call-points.

> However, let's put ourselves in the shoes of someone who binds to a multicast address with the old code: stuff's broken. We fix this. That's not a break of functional compatibility! So: adding this to the list of things to backport.

Yes, I agree. It's akin to a fixing a bug where a program claims it can read some type of file, but silently fails to.

marcusmueller (approved):

rsekman:
Changes in the added commits:
1. Reworked the example to be fully contained in GRC. We send a sine as PDU:s in one arm of the flowgraph and receive it in the other, plotting it with a QT Time Sink.
image
3. I realised that `udp_source` should also be multicast-capable. However, that requires an API break because `udp_source` needs to know which multicast group, if any, to join. Currently by getting a port-only endpoint from `boost::asio::ip` it binds to `INADDR_ANY`. This also has the effect of allowing the 
user to bind to only one network interface in unicast situations. I included a corresponding example. 
image

PR #7813, opened 2025-04-13
qtgui/matrixdisplay: avoid extraneous copy, work with Qt<5.14

marcusmueller:
## Description

Was backporting #6718, failed, because that didn't include Jeff's fix to work with Qt 5.12.

This is a "best of both worlds" change that avoids the zero-initialization that `QVector(size_t size)` does, but allows for code to work with older versions of Qt.

## Checklist

### DIFF
--- a/gr-qtgui/lib/matrix_sink_impl.cc
+++ b/gr-qtgui/lib/matrix_sink_impl.cc
@@ -1,6 +1,7 @@
 /* -*- c++ -*- */
 /*
  * Copyright 2023 Free Software Foundation, Inc.
+ * Copyright 2025 Marcus Müller
  *
  * This file is part of GNU Radio
  *
@@ -111,9 +112,13 @@ int matrix_sink_impl::work(int noutput_items,
     auto in = static_cast<const input_type*>(input_items[0]);
 
     for (int k = 0; k < noutput_items; k++) {
-        QVector<double> qvec(d_vlen);
-        std::copy(in + k * d_vlen, in + (k + 1) * d_vlen, qvec.begin());
-        emit d_signal->data_ready(qvec);
+#if QT_VERSION < QT_VERSION_CHECK(5, 14, 0)
+        QVector<double> data(d_vlen);
+        std::copy_n(in + k * d_vlen, d_vlen, data.begin());
+#else
+        QVector<double> data(in + k * d_vlen, in + (k + 1) * d_vlen);
+#endif
+        emit d_signal->data_ready(data);
     }
 
     return noutput_items;


mbr0wn (approved):
This seems fine, but where does it save a copy? Or is the QVector ctor with the data somehow not copying? (Genuinely curious)

gr-qtgui/lib/matrix_sink_impl.cc
```diff
@@ -111,9 +112,13 @@ int matrix_sink_impl::work(int noutput_items,
...
 
     for (int k = 0; k < noutput_items; k++) {
-        QVector<double> qvec(d_vlen);
-        std::copy(in + k * d_vlen, in + (k + 1) * d_vlen, qvec.begin());
-        emit d_signal->data_ready(qvec);
+#if QT_VERSION < QT_VERSION_CHECK(5, 14, 0)
+        QVector<double> data(d_vlen);
+        std::copy_n(in + k * d_vlen, d_vlen, data.begin());
+#else
+        QVector<double> data(in + k * d_vlen, in + (k + 1) * d_vlen);
```
mbr0wn, line 119: In fact, are you sure this is not worse than before? How does `QVector` know how much memory to allocate? Wouldn't it potentially resize the vector as it's iterating?

marcusmueller: I was going from memory of std::vector, where in `std::vector(first, last)` first and last have a defined `std::distance(first,last)`, the alloc happens at an appropriate size once, and complexity is linear in that distance. I was assuming that was the reason Qt introduced that same constructor signature: to make use of that feature.

And, indeed, /usr/include/qt5/QtCore/qvector.h:

```c++
template <typename T>
template <typename InputIterator, QtPrivate::IfIsInputIterator<InputIterator>>
QVector<T>::QVector(InputIterator first, InputIterator last)
    : QVector()
{
    QtPrivate::reserveIfForwardIterator(this, first, last);
    std::copy(first, last, std::back_inserter(*this));
}
```

(and `reserveIfForwardIterator` just SFINAEs an attemt to `std::distance(first, last)`)

marcusmueller:
> This seems fine, but where does it save a copy? 

You're right, the code doesn't save a copy, only elides an unnecessary default initialization.

If you build a toy example like [this](https://0x0.st/KBtP.cc/test.cc), you'll see that this actually *can* lead to a copy elision in an optimized build. I doubt that's happening here (in fact I checked through debugging, and it's not eliding a copy, which would be hard: it would need to see that the rest of what's done with that QVector doesn't touch the content, because the input is const, after all), and if anything, it doesn't come from the code but from compileroptimization:

hence, fixed commit msg

gr-qtgui/lib/matrix_sink_impl.cc
mbr0wn: Ah, thanks for the explanation.

grc/gui/canvas/colors.py
mbr0wn, line 52: Bad rebase?

mbr0wn (approved):
The Qt change seems small and OK. Any reason to not merge and reduce clutter in the PR list?

grc/gui/canvas/colors.py
marcusmueller: fixed (rebased on the wrong remote/main), thanks!

Issue #7819, opened 2025-04-26
Deprecate peak_detector


marcusmueller:
### Feature Description

The logic is inherently broken (e.g. signal jumps very high for a single sample, then just keeps skimming the pseudo-average for a million samples), it's unnecessarily convoluted, and the constructor has a parameter not used by the code – about which the API documentation lies. 

We have peak_detector2, which works better.

So, deprecate on maint-3.10, remove from 3.11, and fix docs.

### Feature Urgency

high (impacts broad portion of ecosystem)

ffu:
If the peak_detector is deleted in 3.11, should GNU Radio only have peak_detector2? Seems like peak_detector2 should then be renamed to peak_detector, but I understand if it's not worth the effort.

PR #7822, opened 2025-05-03
Update blocks_peak_detector_xb.block.yml

### LINKED ISSUE #7818: blocks: peak_detector gets wrong peak or fails to find peak
### What happened?

When testing block peak_detector_fb with a raised cosine wave y[n] = 1 + cos(n*2*$\\pi$/10), I expect peaks to occur at indices 10*k, where k is an integer. Occasionally, either the detected peak is off by 1, or no peak is found.

Attached are screenshots of both cases.

Image

Image

Blue line is the peak detector input. Orange line is the peak detector output.

Analysis shows problem occurs in peak_detector_impl.cc:
```
    if (state == 0) {
        // printf("Leave in State 0, produced %d avg %f\n",noutput_items, d_avg);
        return noutput_items;
    } else { // only return up to passing the threshold
        // printf("Leave in State 1, only produced %d of %d %f\n",peak_ind,noutput_items, d_avg);
        return peak_ind + 1;
    }
}
```
During the work function, if the state == 1, then returning peak_ind + 1 will set the output corresponding to the peak index to 0 (no peak) and this local peak in the input buffer is consumed. Therefore, an attempt will be made to find the local peak in the next call to work. In the next call to work, if there are enough samples, another peak could be found at a different location. Sometimes, due to the throttle block, the next noutput_items is 1. In that case, the input is consumed again, and after another call to work, the next sample may be beneath the fall threshold, resulting in no peak output.

This block is short but complicated due to its multi-state implementation while attempting to keep track of the average. I attempted to fix by changing to "return peak_ind". This will solve the peak calculation, but this could also create repeated unconsumed loops, especially when noutput_items is 1 and the work function returns 0. In a similar fashion, if the input is a plateau or step function, that change may cause the block to become stuck. Therefore, a suitable fix may require some significant changes. As it stands, this investigation shows that peak_detector2 may be a better alternative since it truly implements the look_ahead parameter.

### System Information

OS: Linux and Mac
GR Installation Method: Source (Linux) and Homebrew (Mac)

### GNU Radio Version

3.11-git (main)

### Specific Version

g884e8f6dd

```
from gnuradio import blocks, gr, analog
import matplotlib.pyplot as plt
import numpy as np


THROTTLE_RATE = 1e6 # Increase throttle rate to see the off-by-one error
NUMSAMPS = 9000
SAMP_FREQ = 10
tb = gr.top_block()
pkdet = blocks.peak_detector_fb(1.5, 1.5)
cos = analog.sig_source_f(SAMP_FREQ,
                          analog.GR_COS_WAVE,
                          wave_freq=1,
                          ampl=1,
                          offset=1)

snk_peak = blocks.vector_sink_b()
snk_cos = blocks.vector_sink_f()
head = blocks.head(4, NUMSAMPS)
throttle = blocks.throttle(4, THROTTLE_RATE)
tb.connect(cos, throttle, head, pkdet, snk_peak)
tb.connect(head, snk_cos)
tb.run()

out = np.array(snk_peak.data())
inpt = np.array(snk_cos.data())

expected_peak_indices = range(0, len(out), SAMP_FREQ)
errors = [n for n in expected_peak_indices if out[n] != 1]
if errors:
    for e in errors:
        print("No peak found at index", e)
    first_error = errors[0]
    x = np.arange(first_error-20, first_error + 21)
    plt.plot(x, inpt[x])
    plt.plot(x, out[x])
    plt.show()
```

### Relevant log output

```shell

```

Moez-Ahmed:
grc: Mark peak_detector block as deprecated (Issue #7818)

This commit adds the `flags: [deprecated]` property to the
`gr-blocks/grc/blocks_peak_detector_xb.block.yml` file.

This change is the first step in addressing Issue #7818, which
identifies problems with the `peak_detector` block and
recommends its deprecation and eventual removal.

Marking the block as deprecated will:
- Visually indicate in GNU Radio Companion (GRC) that the
  block is no longer recommended for use (likely with an
  orange color or warning icon).
- Inform users that this block may be removed in future
  releases, encouraging them to migrate to alternative
  solutions like `peak_detector2`.

## Description
This pull request implements the initial step towards
deprecating the `peak_detector` block. By adding the
`flags: [deprecated]` property to its GRC YAML file, this
change will signal to users within GNU Radio Companion that
the block is slated for removal due to identified issues
with its logic and API. This aims to guide users towards more
reliable alternatives.

## Related Issue
Fixes #7818

This change primarily affects the `gr-blocks` module,
specifically the GRC integration for the `peak_detector`
block. It impacts the visual representation of this block
within the GNU Radio Companion.

## Testing Done
After modifying the `blocks_peak_detector_xb.block.yml`
file, I rebuilt and reinstalled GNU Radio from source. I
then opened GNU Radio Companion and verified that the "Peak
Detector" block is now marked as deprecated (indicated by
an orange color in the block explorer).

## Checklist
- [x] I have read the [CONTRIBUTING document](
  https://github.com/gnuradio/gnuradio/blob/main/CONTRIBUTING.md).
- [x] I have squashed my commits to have one significant
  change per commit.
- [x] I [have signed my commits before making this PR](
  https://github.com/gnuradio/gnuradio/blob/main/CONTRIBUTING.md#dco-signed)
- [x] My code follows the code style of this project. See [GREP1.md](
  https://github.com/gnuradio/greps/blob/main/grep-0001-coding-guidelines.md).
- [ ] I have updated [the documentation](
  https://wiki.gnuradio.org/index.php/Main_Page#Documentation) where
  necessary. (The wiki update will be done as a separate step.)
- [ ] I have added tests to cover my changes, and all previous
  tests pass. (This change primarily affects GRC metadata and
  doesn't require new runtime tests.)

### DIFF
--- a/gr-blocks/grc/blocks_peak_detector_xb.block.yml
+++ b/gr-blocks/grc/blocks_peak_detector_xb.block.yml
@@ -1,6 +1,6 @@
 id: blocks_peak_detector_xb
 label: Peak Detector
-flags: [ python, cpp ]
+flags: [ python, cpp, deprecated ]
 
 parameters:
 -   id: type


marcusmueller:
Fixes #7819

PR #7825, opened 2025-05-09
OFDM Receiver: add support for even carriers in STF

spukhafte-sys:
## Description
The OFDM Loopback example doesn't work with even carriers in Sync Word 1 (used in the IEEE 802.15.4 SUN PHY), because [OFDM Receiver](https://wiki.gnuradio.org/index.php/OFDM_Receiver) doesn't set the **use_even_carriers** parameter in its call to the Schmidl & Cox sychronizer.

Without this fix OFDM Receiver can't synchronize if **Sync Word 1** (the STF in 802.15.4) uses even carriers. This fix detects whether even or odd carriers are used, complains if both are and sets the **use_even_carriers** parameter in the call to the Schmidl & Cox synchronizer accordingly.

## Related Issue
N/A

- gr-digital/python/digital/ofdm_txrx.py

## Testing Done
Tested with all IEEE 802.15.4 SUN PHY ODFM Options (1-4).

## Checklist

### DIFF
--- a/gr-digital/python/digital/ofdm_txrx.py
+++ b/gr-digital/python/digital/ofdm_txrx.py
@@ -319,11 +319,17 @@ class ofdm_rx(gr.hier_block2):
         if sync_word1 is None:
             self.sync_word1 = _make_sync_word1(
                 fft_len, occupied_carriers, pilot_carriers)
+            even_carriers = any(self.sync_word1[0::2])
         else:
             if len(sync_word1) != self.fft_len:
                 raise ValueError(
                     "Length of sync sequence(s) must be FFT length.")
+
+            even_carriers = any(sync_word1[0::2])
+            if even_carriers and any(sync_word1[1::2]):
+                raise ValueError("Odd and even carriers in Sync Word 1")
             self.sync_word1 = sync_word1
+
         self.sync_word2 = ()
         if sync_word2 is None:
             self.sync_word2 = _make_sync_word2(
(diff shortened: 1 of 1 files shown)


spukhafte-sys:
A corrected commit message is in commit 8945ec7e1.

gr-digital/python/digital/ofdm_txrx.py
```diff
@@ -319,11 +319,17 @@ def __init__(self, fft_len=_def_fft_len, cp_len=_def_cp_len,
...
         else:
             if len(sync_word1) != self.fft_len:
                 raise ValueError(
                     "Length of sync sequence(s) must be FFT length.")
+
+            even_carriers = any(sync_word1[0::2])
+            if even_carriers and any(sync_word1[1::2]):
+                raise ValueError("Odd and even carriers in Sync Word 1")
             self.sync_word1 = sync_word1
+
```
daniestevez, line 332: Two comments. First, on code duplication. It seems that the new code could go after the `if/else` in order to do the same in both cases. The `if/else` is just calling `_make_sync_word1()` to generate a sync word if one was not supplied, or checking the length of the sync word if one was supplied.

Second, maybe it's worth to add a comment to the `ValueError` (or even add it to the error string) that says that synchronization uses Schmidl & Cox and this requires every other carrier to be zero so that the syncword has a period of Tu/2. This tripped me when I was reviewing this code, as I was left wondering what was wrong if the syncword had non-zero elements in all positions and whether these changes would break any previous usage of this block.

daniestevez (commented):
I just have some minor comments. Besides those, the changes look good to me. The code needs to be squashed to a single commit (there are presently 5 commits). And also DCO is missing, as reported by CI.

spukhafte-sys:
I couldn't figure out how to modify this pull request, so I submitted a [new one](https://github.com/gnuradio/gnuradio/pull/7837).

gr-digital/python/digital/ofdm_txrx.py
spukhafte-sys: Yes, it could go after the `if/else`, but I didn't think there was much point checking the `sync_word1` generated by `_make_sync_word1()`.

PR #7834, opened 2025-05-31
digital: fix bugs in band-edge filter design

daniestevez:
## Description

The band-edge filter that is designed in the FLL Band-edge block has two bugs:

- The width of the frequency domain response of the filter is `2 * rolloff`, but it should be `rolloff`.

- The filter is translated in frequency to be centered at `(1 + rolloff) / (2 * sps)`, but it should be centered at `1 / (2 * sps)`.

These problems, and their fixes, are demonstrated in [this Jupyter notebook](https://github.com/daniestevez/jupyter_notebooks/blob/master/Band-edge%20filter.ipynb).

Additionally, this commit changes "Prototype Filter Size" in the GRC YAML to "Filter Size". The band-edge filter is not a Polyphase Filterbank, so using the word "Prototype" here can be misleading.

## Related Issue

No issue.

FLL Band-Edge block in gr-digital

## Testing Done

No testing done directly on the changes. The attached [Jupyter notebook](https://github.com/daniestevez/jupyter_notebooks/blob/master/Band-edge%20filter.ipynb) has a Python translation of the filter design code, and the frequency response of the filter is plotted to check that it is correct after making these changes.

## Checklist

- [ ] I have updated [the documentation](https://wiki.gnuradio.org/index.php/Main_Page#Documentation) where necessary. -> No docs changes necessary.
- [ ] I have added tests to cover my changes, and all previous tests pass. -> No new unit tests. Proper unit test is clearly lacking for this block if a wrong filter design was able to go under the radar for many years.

### DIFF
--- a/gr-digital/grc/digital_fll_band_edge_cc.block.yml
+++ b/gr-digital/grc/digital_fll_band_edge_cc.block.yml
@@ -19,7 +19,7 @@ parameters:
     label: Filter Rolloff Factor
     dtype: real
 -   id: filter_size
-    label: Prototype Filter Size
+    label: Filter Size
     dtype: int
 -   id: w
     label: Loop Bandwidth
--- a/gr-digital/lib/fll_band_edge_cc_impl.cc
+++ b/gr-digital/lib/fll_band_edge_cc_impl.cc
@@ -167,7 +167,7 @@ void fll_band_edge_cc_impl::design_filter()
     const float half_sps_inv = 2.0f / d_sps;
     for (size_t i = 0; i < d_filter_size; i++) {
         const float k = -M + i * half_sps_inv;
-        const float position = d_rolloff * k;
+        const float position = 0.5f * d_rolloff * k;
         const float tap = sinc(position - 0.5f) + sinc(position + 0.5f);
         power += tap * tap;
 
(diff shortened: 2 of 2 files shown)


daniestevez:
Okay. The unit test for FLL Band-Edge is failing (and also for Constellation Receiver, which I guess uses the FLL Band-Edge). I should take a look at that. I guess that the FLL Band-Edge might have been instantiated with wrong parameters to counter the wrong band-edge filter to some extent.

daniestevez:
Update: do not merge this for now. There are other implementation problems in the FLL Band-Edge that might require us to spin up a new block to fix everything without breaking existing flowgraphs. See Development Matrix room for discussion.

daniestevez:
Closing this PR, since I have realized that the band-edge FIR is implemented correctly as intended. The construction of this FIR is described in https://s3.amazonaws.com/embeddedrelated/user/124841/wpmc_2010_be_filters_5%20_21343.pdf and summarized by the following sentences:

> The quarter cycle band edge shapes of the frequency matched filter are discontinuous hence cannot be implemented. Our approach to implement this filter is to continue the spectral shapes past the discontinuity to form a half cycle of the spectral cosine.

In this sense, this band-edge FIR does not try to follow the optimal frequency derivative matched filter, but rather continues it symmetrically to obtain a smooth frequency response, which probably makes for an FIR filter with a lower number of taps that other approaches to try to realize the frequency derivative matched filter.

Still there are problems with how the loop filter in the FLL Band Edge block is handled. I have outlined these in the Matrix Development channel, but mainly they are:
- The gain of the discriminant is not taken into account.
- The loop filter update is `freq += beta * error; phase += freq + alpha * error;`. However, for an FLL the loop update should be `freq += alpha * error; phase += freq;`.

PR #7836, opened 2025-06-01
grc: Tidy up tests (don't run test_compiler and test_cpp)

### LINKED ISSUE #7828: grc_tests opens an xterm window, waits for input
### What happened?

Current main head:
running `ctest` starts grc_tests, which opens a terminal window asking the user to press a button.

That is of course a breakage in automated testing

### System Information

OS: Fedora 42 x86_64 X.org

### GNU Radio Version

3.11-git (main)

### Specific Version

v3.11.0.0git-945-g22bc9e07fb

`cmake …; ninja or make; ctest -R grc_tests`

Image

`xprop` tells me the window-creating command is

`WM_COMMAND(STRING) = { "/usr/bin/xterm", "-e", "/usr/bin/python3", "-u", "/tmp/top_block.py" }`.

### Relevant log output

```shell

```

haakov:
## Description

Don't run test_compiler and test_cpp, just generate. See https://github.com/gnuradio/gnuradio/issues/7828

Note: Maybe we should have a signal generator that generates only a limited amount of samples, so that it'll quit on its own. What are the other modules' tests doing?

## Related Issue

Fixes #7828

GRC/Testing

## Testing Done

Ran both ctest in build directory and pytest in grc/tests

## Checklist

### DIFF
--- a/grc/tests/.gitignore
+++ b/grc/tests/.gitignore
@@ -1,5 +1,5 @@
 .pytest_cache
-resources/horn/
+resources/cpptest/
 resources/test_cpp_hb_nogui_workflow.block.yml
 resources/test_cpp_hb_nogui_workflow/
 resources/test_cpp_hb_qt_gui_workflow.block.yml
--- a/grc/tests/resources/test_cpp.grc
+++ b/grc/tests/resources/test_cpp.grc
@@ -10,7 +10,7 @@ options:
     gen_linking: dynamic
     generate_options: qt_gui
     hier_block_src_path: '.:'
-    id: horn
+    id: cpptest
     max_nouts: '0'
     output_language: cpp
     placement: (0,0)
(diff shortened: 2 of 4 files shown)


marcusmueller:
Bit confused; why does `run=True` cause an xterm window to be opened?

haakov:
`run=True` runs the flowgraph immediately after building/generating it (must have been an oversight back when I added these tests years ago). So I suppose the question is rather why _didn't_ we see an xterm window earlier!

https://github.com/gnuradio/gnuradio/blob/f7628889f6f3fe0b8016b78ea0b2b1adc71dd260/grc/compiler.py#L27-L28

marcusmueller:
that *is* a fair question :) 

But I guess in this case, we really *don't* want to use the (graphical frontend) GRC "run in terminal" capabilities. The follow-up question to that is how, in the presence of unknown Generators, whose constructor argument might or might not feature an `add_xterm` (or what it's called?) parameter, we can run without opening a terminal?

PR #7840, opened 2025-06-04
qtgui: Fix constellation plots to maintain square aspect ratio

### LINKED ISSUE #7780: Qt GUI Constellation sink should (by default) keep aspect ratio
### Feature Description

I've not once in my life looked at a constellation plot and thought "oh it would be nice if this was stretched to fill the space in x and y separately. I'd always then went on and manually modified it such that it becomes "square".

I believe this is the same for most people. It confuses beginners and annoys experienced comms people.

constellation plot from https://ham.stackexchange.com/questions/23438/output-from-constellation-modulator-different-from-demodulator-output

### Feature Urgency

ffu:
Modified constellation display plot such that the canvas width doesn't surpass its height, which can cause undesirable appearances.

Fixes #7780

## Description

Previously, the constellation plots could be stretched to distracting and undesirable dimensions. Unlike time and frequency plots, wide constellation plots distort the symmetry seen in most modulation schemes like QAM and PSK. 

This change prevents constellation plots from becoming too wide and maintains the square aspect ratio in the canvas, the primary plot object. It does, however, allow the canvas to become narrow. Otherwise, testing showed that manual resizing of the plot widget became confusing and unintuitive. On the other hand, the user could always manually stretch a narrow plot to the desired square plot.

This change is a slight work-around because of the limitations of the QwtPlot widget. A direct fix would involve inheriting QwtPlotLayout and completely overriding the activate() method. However, that method performs numerous calculations based on geometries of the title, axes, and legend before calculating the canvas size, and a lot of effort was done to get that working.

Instead, this fix will call the original activate method and observe the calculated canvas size. If the canvas size is not desirable, another call to activate is made with an adjusted QwtPlot size to get the square canvas.

## Related Issue

Fixes #7780 

Constellation Sink, GUI Sink, and BER Curve Sink.

## Testing Done

Ran qa_qtgui.py for regression testing and passed
Ran ctest for regression testing and passed.

Visual inspection of two constellation sinks stacked vertically.

Narrow case:
narrow

Square case:
square

Wide case:
wide

Also visually tested GUI Sink and BER Curve sink.
## Checklist

### DIFF
--- a/gr-qtgui/lib/ConstellationDisplayPlot.cc
+++ b/gr-qtgui/lib/ConstellationDisplayPlot.cc
@@ -14,9 +14,44 @@
 #include <gnuradio/qtgui/ConstellationDisplayPlot.h>
 
 #include <qwt_legend.h>
+#include <qwt_plot_layout.h>
 #include <qwt_scale_draw.h>
 #include <QColor>
 #include <cmath>
+
+
+class SquareCanvasLayout : public QwtPlotLayout
+{
+public:
+    SquareCanvasLayout() {}
+    SquareCanvasLayout(QwtPlotLayout* layout)
+    {
+        const int AXIS_COUNT = 4;
+        for (int axis = 0; axis < AXIS_COUNT; axis++) {
+            setCanvasMargin(layout->canvasMargin(axis), axis);
+            setAlignCanvasToScale(axis, layout->alignCanvasToScale(axis));
+        }
+        setSpacing(layout->spacing());
+        setLegendPosition(layout->legendPosition());
+        setLegendRatio(layout->legendRatio());
+    }
+    // Override to perform layout with square canvas
+    void activate(const QwtPlot* plot,
+                  const QRectF& plotRect,
+                  Options options = Options()) override
+    {
+        // Do initial layout
+        QwtPlotLayout::activate(plot, plotRect, options);
+        QRectF canvas = canvasRect();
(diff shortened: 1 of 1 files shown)


marcusmueller:
would it be possible to keep the aspect ratio 1:1, just make the "visible" part of the complex plane larger?

ffu:
Are you suggesting expanding the x axis scale accordingly, like this:
Screenshot 2025-06-07 at 11 08 21 AM

Most likely yes, it could be done. However, that requires an automatic adjustment of the x-axis relative to the y-axis. That cannot be done if the the scale is manually set to -2 and 2 for the min and max x and y axis. 

Maybe I could change the default to autoscale, and then I would have to change how autoscale works to do this square aspect ratio. But then autoscale may not be ideal if the dimensions of the constellation points keep changing. Yeah, this is why it was easier to shrink the canvas! Let me know your suggestions. Thanks.

marcusmueller:
right! You're on point: the "widescreen" picture you show is (I think) what most users would probably want, but it's not compatible with what they specify if they set bounds for both axes. 

Yeah, seeing that, your approach here is the best we can do without fundamentally changing what the block does.

marcusmueller:
@ffu I see this is a draft PR. Particular reason? If not, please mark it as ready, and we can move on with merging.

marcusmueller:
rebased, let's run CI once more, then merge.

PR #7843, opened 2025-06-07
Continuation of #7561: GRC Py Blk w/ config' IOvec size

marcusmueller:
Supersedes #7561

(can't push over that branch, so python formatting fixes have to happen here)

- **grc: factor out Python Block parameter marking logic magic**
- **grc: use current block parameter values when extracting io signature**
- **grc: don't draw block port GUI element without valid dimensions**
- **grc: add test for epy blocks with parameter dependent IO signature**
- **grc/gui_qt: don't draw block port GUI element without a valid parent**
- **grc/gui: update ports/labels when rewriting block within PropsDialog**
- **Revert "grc: don't draw block port GUI element without valid dimensions"**
- **grc/gui: document draw() requiring create_shapes()**


ghost:
Sorry if this is not the place. Please, move, edit, or delete this comment if applicable.

I use epy blocks to start threads, open sockets, overwrite files, and other _dangerous_ stuff. Some of these actions could be placed in the `__init__`. I carefully choose default values for not doing these things until running time, when all parameters are stable, and it is possible to gracefully close files and sockets, and terminate threads at the end.

To instantiate the epy block in GRC with non-defaults values, every time a parameter is changed (maybe by mistake), with the sole purpose of getting io signatures, could lead to a _disaster_.

Please, warn about this possibility in the doc, and maybe advice against this _bad_ practice (of mine).

Sincerely grateful for your work!

PR #7857, opened 2025-06-08
Remove unused includes from runtime, blocks

marcusmueller:
- **runtime: remove unused transient includes in include/**
- **buffer_single_mapped: locking needs a named object**
- **runtime: removed unused includes in lib/**
- **blocks: removed unused includes**

## Description

Reducing the amount of transient includes in gnuradio-runtime/include is a bug-reducing measure; sadly makes backporting harder (might break OOTs which erroneously got includes that way).

## Testing Done

Ruuuns. 

## Checklist


gnuradio-runtime/include/gnuradio/buffer_double_mapped.h
```diff
@@ -37,9 +37,6 @@ class GR_RUNTIME_API buffer_double_mapped : public buffer
                                    block_sptr link = block_sptr(),
                                    block_sptr buf_owner = block_sptr());
 
-    gr::logger_ptr d_logger;
```
mbr0wn, line 40: FYI other reviewers: These are also in the parent class `buffer`.

mbr0wn (approved):
This is fine, but I do think we don't need to spend that much time on such incremental improvements, unless we think they hugely affect performance or security. These commits make backports etc. a lot more difficult.

I did a full review on this though, to be clear and think it's safe to merge.

marcusmueller:
@mbr0wn agree with the urgency assessment of this, but these were things that have been cluttering my own quality assurance while working on GNU Radio, so this was "I stumbled over this carpet 10 times, might as well take the time and straighten it".

PR #7864, opened 2025-06-08
blocks:                       Add fail_if_exists option to file_sink  (FileSink should have an option to fail if file already exists #1288)

### LINKED ISSUE #1288: FileSink should have an option to fail if file already exists
FileSink has 2 options: append or overwrite.
Sometimes users forget to change the filename but they don't intend to either lose data (overwrite) or modify it (append).
An option to configure FileSink to throw an exception if the file already exists will be useful.

snk-git-hub:
blocks: Add fail_if_exists option to file_sink

## Description
This PR adds a `fail_if_exists` option to the `file_sink` block. When this option is set, the block will throw an exception or fail to initialize if the target file already exists, preventing accidental overwrites.

This change is required to address user requests for improved data safety and to avoid unintentional data loss by overwriting existing files.

## Related Issue
Fixes #1288

- `blocks/file_sink` implementation and its interface
- Unit tests for `file_sink`

## Testing Done
- Added unit tests to verify that the `file_sink` block throws an exception if the file already exists and `fail_if_exists` is set.
- Ran the full test suite locally (Fedora, GNU Radio built from source, GCC version X.Y.Z).
- Confirmed that all tests pass and no existing functionality is broken.

## Checklist


marcusmueller:
hi @snk-git-hub ! Welcome here! Looking forward to your contributions!

just: These checkboxes up there are not for decoration! You need to actually *do* the things that you tick boxes for :) Unchecked those that still need doing!

We'll need you to do a  couple of things

1. `git rebase` this on `main`, so to get rid of the merge commit (I've retargeted this PR onto main – this is a feature PR, and these always go into `main`, and can get backported if compatible to 3.10)
2. add the new parameters the documentation comments in the headers
3. write a unit test (in gr-blocks/python/qa_file_sink.py) that tests that your block works as expected in both settings. 
5. run `clang-format -i` on the changed files
6. use `git commit --amend -s` to add a "signed-off by" text to your commit.
7. When you build this with an empty build directory, you should probably get a complaint about the hash mismatch in the python bindings — and that's correct! You will need to 
   1. add the missing `py::arg`s and
   2. fix the `BINDTOOL_HEADER_FILE_HASH` at the beginning of `gr_blocks/python/blocks/bindings/file_sink_python(_base).cc`

snk-git-hub:
Thank you for the feedback and for retargeting the PR
I will Rebase my branch on main to remove the merge commit.
Amend my commit to add the signed-off-by line
I’ll update the PR as soon as that’s done.

gr-blocks/lib/file_sink_base.cc
```diff
@@ -65,6 +65,13 @@ bool file_sink_base::open(const char* filename)
     gr::thread::scoped_lock guard(d_mutex); // hold mutex for duration of this function
     d_base_debug_logger->debug("opening file {:s}", filename);
 
+    // Fail if file exists , if required
+    if(d_fail_if_exists && std::filesystem::exists(filename)){
```
marcusmueller: instead of checking using `std::filesystem::exists`, which leaves us open to [TOCTOU](https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use), simply add [`O_EXCL`](https://pubs.opengroup.org/onlinepubs/007904875/functions/open.html) to `flags` below – that makes `open` fail if the file already exists.

Something like
```c++
if(d_fail_if_exists) {
  flags |= O_EXCL;
}
```

marcusmueller (changes requested):

gr-blocks/lib/file_sink_base.cc
snk-git-hub: Ill update the code to use O_EXCL in the open flags when fail_if_exists is set instead of checking with std::filesystem::exists. That makes a lot of sense and should handle the race condition issue , I’ll push the updated code shortly.

marcusmueller: No hurries – I don't think I'll get around reviewing things today :) It's getting a bit late!

snk-git-hub: Thanks : ) for letting me know! I’m still pretty new to open source and figuring out how things work, so I really appreciate your patience (and all the detailed feedback). No worries about the review timing.

gr-blocks/lib/file_sink_base.cc
```diff
@@ -12,14 +12,14 @@
...
 #endif
-
 #include <gnuradio/blocks/file_sink_base.h>
 #include <gnuradio/logger.h>
 #include <gnuradio/thread/thread.h>
 #include <fcntl.h>
 #include <sys/stat.h>
 #include <sys/types.h>
 #include <cstdio>
+#include <filesystem>
```
marcusmueller: I don't think we need that include anymore :)

snk-git-hub: Thanks: )  for pointing that out! I’ll remove the unused include
If there’s anything else you notice that should be changed, please let me know

marcusmueller: ah, you'll notice the CI says that your C++ is ill-formatted. Clang-format? 

Same for Python, `pycodestyle`. If that's too much work or you need help, let me know :)

snk-git-hub: yeah i need help : )

marcusmueller: that's OK, your PR is pretty nice. Let me quickly remove the unused include and format the files for you.

snk-git-hub: Thanks a lot! I really appreciate it  : )

marcusmueller: hey, while I'm at it, I'd like to add 
```
* Copyright 2025 shivanandu
```
to the files, is that OK for you?

snk-git-hub: That’s OK with me

marcusmueller:
So, I did the mechanical things!


1. removed the unused `#include <filesystem>`
2. remove the unused `d_fail_if_exists` in `file_sink_impl`: you handle this in `file_sink_base`!
3. added the empty line that the CI complained about (reading the CI output is always helpful!)
4. added the missing documentation to `file_sink.h`; you only added it in one place, but the parameter appears in multiple places
5. added the python arg as requested above
6. fixed the MD5 hashes as described above
7. `cd gnuradio && clang-format -i $(git diff --names-only main)`

However, your change to using `O_EXCL` was half-heartedly. We can only merge this if you fix your unit test! Marking this PR as draft until you find the time :)

snk-git-hub:
I'll update the unit test for the O_EXCL change and mark the PR ready for review once that’s done.

marcusmueller:
Hi, as explained above, you need to sign your commits. Please actually read the output when a CI workflow fails! It links to useful information on how to solve the issue.

marcusmueller:
You added a folder "Temporary". Remove that. Please do your due diligence!  (Adding the `Don't Merge` label, because I've spent now very much time on this PR, and it's not kept clean; this has become a bit of a time sink for maintenance. Please only remove the draft status when the DCO check passes, and you can really confidently check all five checkboxes in the PR template above! You checked the "I signed my commits" checkbox again, although your commit is not signed.)

snk-git-hub:
I’ll make sure not to mark those as complete unless they are fully verified in the future. I appreciate your patience, and I’ll be more diligent and careful going forward to avoid such slip-ups. : )

PR #7867, opened 2025-06-09
CI: try running gh worker with 4 cores

marcusmueller:
I see that tools that count their own cores run with 4 threads. Let's try the same – CI has been **awfully** slow the last couple of days (where I could routinely get 40 min PR builds, I'm now happy when everything is done 1h 30min later).

### DIFF
--- a/.github/workflows/make-test.yml
+++ b/.github/workflows/make-test.yml
@@ -111,7 +111,7 @@ jobs:
       image: ${{ matrix.containerid }}
       volumes:
         - build_data:/build
-      options: --cpus 2 ${{ matrix.containeroptions }}
+      options: --cpus 4 ${{ matrix.containeroptions }}
     steps:
     - uses: actions/checkout@v1
       name: Checkout Project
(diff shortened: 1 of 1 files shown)


marcusmueller:
This does seem to be a bit faster, though it's hard to compare. I'd tend towards keeping this.

argilo:
Is it intentional that `ctest` is left at `-j2`?

argilo:
The only downside I could see to increasing test parallelism is that time-sensitive tests might fail more often. We're already running them with `-j${CPU_COUNT}` in Conda builds though, so I would anticipate that we've already fixed flaky tests due to failures there.

It's too bad GitHub actions doesn't have something like conda's `CPU_COUNT` environment variable, or I would suggest using that.

argilo (approved):
The change looks good, and I'd be fine with or without the change to test parallelism.

marcusmueller:
> Is it intentional that ctest is left at -j2?

yep, exactly as you say, avoiding making time-sensitive test flakier than they need be

>It's too bad GitHub actions doesn't have something like conda's CPU_COUNT environment variable, or I would suggest using that.

oh good point, maybe we should do something like `$(nprocs)` ? not sure that works in a workflow directly

PR #7869, opened 2025-06-10
gr-blocks: fix redundant null check due to previous dereference

odaysec:
https://github.com/gnuradio/gnuradio/blob/f7628889f6f3fe0b8016b78ea0b2b1adc71dd260/gr-blocks/lib/file_meta_sink_impl.cc#L162-L173

https://github.com/gnuradio/gnuradio/blob/f7628889f6f3fe0b8016b78ea0b2b1adc71dd260/gr-blocks/lib/file_meta_source_impl.cc#L274-L279

This rule finds comparisons of a pointer to null that occur after a reference of that pointer. It's likely either the check is not required and can be removed, or it should be moved to before the dereference so that a null pointer dereference does not occur.

[Null Dereference](https://www.owasp.org/index.php/Null_Dereference)

To fix the issue:
1. Move the null check for `fp` to before its dereference on line 167. This ensures that `fp` is not null before it is dereferenced.
2. Remove the redundant null check on line 172, as it serves no purpose after the dereference.
3. Update the logic to correctly handle the case where `fdopen` fails by checking its return value (`*fp`) instead of `fp`.
4. Check if `fp` is non-null before dereferencing it.
5. If `fp` is null, close the file descriptor (`fd`) and return `false` to prevent further operations on an invalid pointer.

The changes will ensure that the code is safe from null pointer dereferences and that the logic for handling `fdopen` failures is correct. ensure that the code does not attempt to dereference a null pointer, eliminating the risk of undefined behavior.

Signed-off-by: Zeroday BYTE <github@zerodaysec.org>

## Checklist

### DIFF
--- a/gr-blocks/lib/file_meta_sink_impl.cc
+++ b/gr-blocks/lib/file_meta_sink_impl.cc
@@ -159,17 +159,16 @@ bool file_meta_sink_impl::_open(FILE** fp, const char* filename)
         return false;
     }
 
-    if (*fp) { // if we've already got a new one open, close it
+    if (fp && *fp) { // if we've already got a new one open, close it
         fclose(*fp);
-        fp = 0;
+        *fp = nullptr;
     }
 
-    if ((*fp = fdopen(fd, "wb")) == NULL) {
+    if (fp && (*fp = fdopen(fd, "wb")) == NULL) {
         d_logger->error("[fdopen] {:s}: {:s}", filename, strerror(errno));
         ::close(fd); // don't leak file descriptor if fdopen fails.
+        ret = false;
     }
-
-    ret = fp != 0;
 
     return ret;
 }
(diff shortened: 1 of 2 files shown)


marcusmueller:
Hi, I'll need you to `-s` (== `--sign-off`) on your commit. Click on the "DCO" check (with the red cross) for details.

EDIT: engaged maintainer override for that DCO check to get the CI to run on this change.

marcusmueller:
so, the good news is that while certainly a good idea to check the non-null-ness of pointers, the only place the protected `_open` was used was with hardcoded arguments, both of which pointing to addresses of class members. So, no real problem in practice.

mbr0wn:
I fixed the clang-formatting and rebased. This passed other tests. But it still needs a DCO (`git commit --amend -s`). Please add the DCO trailer to your commit message, and reopen the PR.

PR #7871, opened 2025-06-10
annotator_raw: give Mac a lot of time. It needs it.

marcusmueller:
It seems the qa_annotator_raw test really needs to be run with lots of tolerances, as there's no guarantee that sleep(0.4) is close to 400 ms of sleep. (I.e., in practice, the test failed, because the insertion happened more than 500ms, i.e., 100 ms too late.)


marcusmueller:
not backporting this, since the Mac CI is currently dysfunctional and can't be tested

PR #7874, opened 2025-06-11
modtool: Add support for rfnoc_block

mbr0wn:
This extends gr_modtool to also generate C++ block controllers for RFNoC blocks:

    gr_modtool add -t rfnoc_block

...will create such a block controller. C++ RFNoC blocks allow control of RFNoC blocks from GNU Radio, even though GNU Radio does not control the streaming through such blocks.

## Testing Done

I tested this on `gr-rfnoc_gain`, and tried to recreate the same code that was manually created (and is heavily vetted).

## Checklist

### DIFF
--- a/gr-utils/modtool/core/add.py
+++ b/gr-utils/modtool/core/add.py
@@ -50,6 +50,7 @@ class ModToolAdd(ModTool):
         "tagged_stream": "Block with input-to-output flow controlled by input stream tags (e.g. packetized streams)",
         "hier": "Hierarchical container block for other blocks; usually can be described by a flowgraph",
         "noblock": "C++ or Python class",
+        "rfnoc_block": "RFNoC block, used with gr-uhd to control blocks on RFNoC-based USRPs",
     }
     language_candidates = ('cpp', 'python', 'c++')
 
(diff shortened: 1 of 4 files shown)


gr-utils/modtool/core/add.py
```diff
@@ -223,8 +236,7 @@ def _run_lib(self):
         if self.info['version'] in ('38', '310'):
             fname_h = self.info['blockname'] + '.h'
             fname_cc = self.info['blockname'] + '.cc'
-            if self.info['blocktype'] in ('source', 'sink', 'sync', 'decimator',
-                                          'interpolator', 'general', 'hier', 'tagged_stream'):
+            if self.info['blocktype'] not in ('noblock',):
```
mbr0wn, line 242: The right is the same as the left, but the left would have gotten even longer by adding `rfnoc_block` so I though this was neater.

marcusmueller:
first off: thanks sooo much for the `--yes` fix.

gr-utils/modtool/core/add.py
marcusmueller: fine with that. We might choose to later refactor that a centralized `set` that we could just subtract `noblock` from, but honestly, the message here is clearer this way. Anything but a noblock gets an impl.

marcusmueller: thinking about that: `GRTYPELIST.keys()` is already such a set

PR #7879, opened 2025-06-16
grc:  Using the Escape key in PropsDialog is buggy

### LINKED ISSUE #7877: Exiting new sink block crashes the app
### What happened?

First:
```
  File "/usr/lib64/python3.13/site-packages/gnuradio/grc/gui/DrawingArea.py", line 251, in draw
    self._flow_graph.draw(cr)
    ~~~~~~~~~~~~~~~~~~~~~^^^^
  File "/usr/lib64/python3.13/site-packages/gnuradio/grc/gui/canvas/flowgraph.py", line 542, in draw
    draw_element(cr)
    ~~~~~~~~~~~~^^^^
  File "/usr/lib64/python3.13/site-packages/gnuradio/grc/gui/canvas/block.py", line 280, in draw
    port.draw(cr)
    ~~~~~~~~~^^^^
  File "/usr/lib64/python3.13/site-packages/gnuradio/grc/gui/canvas/port.py", line 134, in draw
    cr.rectangle(*self._area)
    ~~~~~~~~~~~~^^^^^^^^^^^^^
```
and this on the board:

Image

Then a full crash.

### System Information

OS: Fedora 43
GR Installation Method: native rpm

### GNU Radio Version

3.10 (maint-3.10)

### Specific Version

3.10.11.0

Place "Wav Sink" on the board.
Double click it.
Change the channels to 2.
Press escape.

### Relevant log output

```shell

```

dl1ksv:
See #7877

But the problem is more general and affects blocks with ports. Changing the port number (input or output ) in a block and pressing Esc removes the block area and only the ports can be seen in the flowgraph.

.
## Description

Pressing Esc in the PropsDialog for blocks without ports the key is mapped to accept, but I think the common use should be 'Abort Editing'

So this pr maps Esc to reject
## Related Issue

Fixes #7877

## Testing Done

Tested like described in #7877
## Checklist

### DIFF
--- a/grc/gui/PropsDialog.py
+++ b/grc/gui/PropsDialog.py
@@ -304,6 +304,9 @@ class PropsDialog(Gtk.Dialog):
             insert('\n\n# External Code ({}.py)\n'.format(block.name), src)
 
     def _handle_key_press(self, widget, event):
+        if event.keyval == Gdk.KEY_Escape:
+            self.response(Gtk.ResponseType.REJECT)
+            return True  # Map Escape key to reject button
         close_dialog = (
             event.keyval == Gdk.KEY_Return and
             event.get_state() & Gdk.ModifierType.CONTROL_MASK == 0 and


mbr0wn:
Can you please change the commit message subject line to:

    grc: Map ESC key to 'reject' in PropsDialog

...otherwise this will throw me off whenever I do `git log grc/`.

PR #7757, opened 2025-03-05
qtgui (and others): Qt6 port

haakov:
Draft PR to keep track of the effort related to https://github.com/gnuradio/gnuradio/issues/7708

PR #7883, opened 2025-06-20
Qt5to6

marcusmueller:
rebase of #7757 

trying to sort of Qt include issues.


marcusmueller:
intensely discussed with @mkoop-ni  & @maitbot at GRCon: will need to be redone atop of the CMake rework (which hopefully allows to cherry-pick 90% of this, and make the rest easier)

PR #7890, opened 2025-07-02
digital: improve FLL Band Edge loop filter

daniestevez:
## Description

The loop filter in the FLL Band Edge block has two problems:

1. It acts on the phase, performing an update that looks like:

```
d_freq += d_beta * error;
d_phase += d_freq + d_alpha * error;
```

This is not correct. An FLL should act only on the frequency, since the error that it measures is a frequency error. The loop update should look like:

```
d_freq += d_beta * error;
d_phase += d_freq;
```

2. It does not take into account the gain of the error discriminant.

This PR addresses these two problems by redesigning the loop filter. Since it is not possible to make this change in a backwards-compatible way that preserves the semantics of the `bandwidth` parameter (since `bandwidth` has broken/inconsistent semantics due to the two problems listed above), a new constructor parameter `improved_loop_filter` that defaults to false is added. When `improved_loop_filter` is false, the legacy behavior is preserved. When `improved_loop_filter` is true, the new fixed loop filter is used.

The legacy loop filter uses control_loop() to set the loop coefficients d_alpha and d_beta. This sets

```
d_alpha = (2 * sqrt(2) * bw) / (1 + sqrt(2) * bw + bw**2)
d_beta = (4 * bw**2) / (1 + sqrt(2) * bw + bw**2)
```

The new loop filter forces the following filter coefficients:

```
d_alpha = 0
d_beta = 8 * pi * bw / samps_per_sym
```

Since having a non-zero alpha term mostly contributes additional (unwanted) loop noise, the legacy and new loop filters can be compared by comparing their expressions for beta. This remaks gives that to maintain a behavior similar to the legacy one with a particular bandwidth `bw`, the new loop bandwidth needs to be set to

```
(bw**2 * samps_per_sym) / ((1 + sqrt(2) * bw + bw**2) * 2 * pi).
```

The new loop filter is designed so that the `bandwidth` parameter has the semantics of BL*T, where BL is the loop bandwidth in Hz and T is the sampling period in seconds. It uses the continuous-update approximation for a 1st order loop.

The `improved_loop_filter` constructor parameter has also been added to the `generic_demod()` and `psk_demod()` classes, which use the FLL Band-Edge block, and their QA tests have been updated to set this parameter to true. The loop bandwidths in these tests have been changed to something reasonable.

For more background information and detailed calculations, see [this blog post](https://destevez.net/2025/07/about-flls-with-band-edge-filters/), which is intended as a companion to this PR.

**Questions for reviewers:** I'm interested in how to introduce these changes in the way that best preservers backwards compatibility and doesn't break existing flowgraphs. The `bandwidth` parameter and the loop filter design in the existing FLL Band-Edge is completely broken. There is no reasonable way to maintain the same or similar semantics of the `bandwidth` parameter, so flowgraphs that want to use the fixed loop filter will need to be updated to set a reasonable bandwidth. The old behaviour is maintained by default. Potentially, for GNU Radio 3.11 the `improved_loop_filter = false` case could be removed altogether, since avoiding breaks is not essential.

## Related Issue

No related issue, but this is related to PR #7834 (see specially the last comment on that PR).

FLL Band-Edge block in gr-digital, and other blocks that use FLL Band-Edge, such as PSK Demod.

## Testing Done

See the "GNU Radio legacy loop filter" section in the referenced [blog post](https://destevez.net/2025/07/about-flls-with-band-edge-filters/). These tests shows that with an appropriate choice of the `bandwidth` parameter, the new loop filter achieves the same convergence time with less loop noise.

## Checklist

- [ ] I have updated [the documentation](https://wiki.gnuradio.org/index.php/Main_Page#Documentation) where necessary -> Documentation updates not done yet. This changes the GRC blocks, so documentation updates will be needed. Not doing them just yet in case the PR review affects how GRC blocks look like.
- [x] I have added tests to cover my changes, and all previous tests pass. -> I have updated the loop bandwidth in tests using this block and used the new loop filter. These tests pass. I haven't added new tests.

### DIFF
--- a/gr-digital/grc/digital_fll_band_edge_cc.block.yml
+++ b/gr-digital/grc/digital_fll_band_edge_cc.block.yml
@@ -24,10 +24,6 @@ parameters:
 -   id: w
     label: Loop Bandwidth
     dtype: real
--   id: improved_loop_filter
-    label: Improved Loop Filter
-    dtype: bool
-    default: False
 
 inputs:
 -   domain: stream
@@ -51,7 +47,7 @@ outputs:
 
 templates:
     imports: from gnuradio import digital
-    make: digital.fll_band_edge_cc(${samps_per_sym}, ${rolloff}, ${filter_size}, ${w}, ${improved_loop_filter})
+    make: digital.fll_band_edge_cc(${samps_per_sym}, ${rolloff}, ${filter_size}, ${w})
     callbacks:
     - set_loop_bandwidth(${w})
 
@@ -63,8 +59,7 @@ cpp_templates:
             ${samps_per_sym},
             ${rolloff},
             ${filter_size},
-            ${w},
-            ${improved_loop_filter});
+            ${w});
     link: ['gnuradio::gnuradio-digital']
     callbacks:
     - set_loop_bandwidth(${w})
--- a/gr-digital/grc/digital_fll_band_edge_cc.block.yml
+++ b/gr-digital/grc/digital_fll_band_edge_cc.block.yml
@@ -19,11 +19,15 @@ parameters:
     label: Filter Rolloff Factor
     dtype: real
 -   id: filter_size
-    label: Prototype Filter Size
+    label: Filter Size
     dtype: int
 -   id: w
     label: Loop Bandwidth
     dtype: real
+-   id: improved_loop_filter
+    label: Improved Loop Filter
+    dtype: bool
+    default: False
 
 inputs:
 -   domain: stream
@@ -47,7 +51,7 @@ outputs:
 
 templates:
     imports: from gnuradio import digital
-    make: digital.fll_band_edge_cc(${samps_per_sym}, ${rolloff}, ${filter_size}, ${w})
+    make: digital.fll_band_edge_cc(${samps_per_sym}, ${rolloff}, ${filter_size}, ${w}, ${improved_loop_filter})
     callbacks:
     - set_loop_bandwidth(${w})
 
@@ -59,7 +63,8 @@ cpp_templates:
             ${samps_per_sym},
             ${rolloff},
             ${filter_size},
-            ${w});
+            ${w},
+            ${improved_loop_filter});
     link: ['gnuradio::gnuradio-digital']
     callbacks:
     - set_loop_bandwidth(${w})
--- a/gr-digital/lib/fll_band_edge_cc_impl.cc
+++ b/gr-digital/lib/fll_band_edge_cc_impl.cc
@@ -36,11 +36,10 @@ float sinc(float x)
 fll_band_edge_cc::sptr fll_band_edge_cc::make(float samps_per_sym,
                                               float rolloff,
                                               int filter_size,
-                                              float bandwidth,
-                                              bool improved_loop_filter)
+                                              float bandwidth)
 {
     return gnuradio::make_block_sptr<fll_band_edge_cc_impl>(
-        samps_per_sym, rolloff, filter_size, bandwidth, improved_loop_filter);
+        samps_per_sym, rolloff, filter_size, bandwidth);
 }
 
 static int ios[] = { sizeof(gr_complex), sizeof(float), sizeof(float), sizeof(float) };
@@ -48,8 +47,7 @@ static std::vector<int> iosig(ios, ios + sizeof(ios) / sizeof(int));
 fll_band_edge_cc_impl::fll_band_edge_cc_impl(float samps_per_sym,
                                              float rolloff,
                                              int filter_size,
-                                             float bandwidth,
-                                             bool improved_loop_filter)
+                                             float bandwidth)
     : sync_block("fll_band_edge_cc",
                  io_signature::make(1, 1, sizeof(gr_complex)),
                  io_signature::makev(1, 4, iosig)),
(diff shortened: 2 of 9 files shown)


gr-digital/grc/digital_fll_band_edge_cc.block.yml
```diff
@@ -19,11 +19,15 @@ parameters:
     label: Filter Rolloff Factor
     dtype: real
 -   id: filter_size
-    label: Prototype Filter Size
+    label: Filter Size
```
daniestevez, line 22: Removed 'prototype' because this doesn't really use a polyphase filterbank.

gr-digital/lib/fll_band_edge_cc_impl.cc
```diff
@@ -35,34 +36,66 @@ float sinc(float x)
...
+
     // Value-check samples per symbol
-    if (samps_per_sym <= 0) {
+    if (samps_per_sym <= 0.0f) {
         throw std::out_of_range("fll_band_edge_cc: invalid number of sps. Must be > 0.");
     }
 
     // Value-check rolloff factor
-    if (rolloff < 0 || rolloff > 1.0) {
+    if (rolloff < 0.0f || rolloff > 1.0f) {
```
daniestevez, line 93: Compare against `float` zero, which is cleaner style.

gr-digital/lib/fll_band_edge_cc_impl.cc
```diff
@@ -73,7 +106,7 @@ fll_band_edge_cc_impl::fll_band_edge_cc_impl(float samps_per_sym,
     }
 
     // Value-check bandwidth
-    if (samps_per_sym <= 0) {
+    if (bandwidth <= 0.0f) {
```
daniestevez, line 104: This fixes what seems a typo in the check.

gr-digital/lib/fll_band_edge_cc_impl.cc
```diff
@@ -193,10 +229,6 @@ void fll_band_edge_cc_impl::design_filter()
         d_taps_upper[index] = std::conj(d_taps_lower[d_filter_size - i - 1]);
     }
 
-    d_updated = true;
-
-    // Set the history to ensure enough input items for each filter
-    set_history(d_filter_size + 1);
```
daniestevez, line 199: `d_updated` and history are not being used. Remove them.

gr-digital/lib/fll_band_edge_cc_impl.cc
```diff
@@ -226,10 +258,6 @@ int fll_band_edge_cc_impl::work(int noutput_items,
                                 gr_vector_void_star& output_items)
 {
     gr::thread::scoped_lock lock(d_setlock);
-    if (d_updated) {
-        d_updated = false;
-        return 0; // history requirements may have changed.
-    }
```
daniestevez, line 232: `d_updated` and history are not being used. Remove them.

gr-digital/python/digital/qa_constellation_receiver.py
```diff
@@ -39,7 +39,7 @@
 TIMING_OFFSET = 1.0
 
 # RECEIVER PARAMETERS
-FREQ_BW = 2 * math.pi / 100.0
+FREQ_BW = 2.3e-3
```
daniestevez, line 42: To the old bandwidth of `2 * pi / 100` a new bandwidth of `2.3e-3` corresponds approximately. This comes from the formula `BL*T = B**2 * samps_per_sym / ((1 + sqrt(2) * B + B**2) * 2 * pi)`.

marcusmueller:
> I'm interested in how to introduce these changes in the way that best preservers backwards compatibility and doesn't break existing flowgraphs.

I'd say you're currently doing it the most considerate way possible, with the backwards-compatible "keep the old (broken) behaviour?" switch. Thanks!

Consequently, I'm adding this to the backports-3.10 project, so that I can merge these changes into maint-3.10.

> There is no reasonable way to maintain the same or similar semantics of the bandwidth parameter, so flowgraphs that want to use the fixed loop filter will need to be updated to set a reasonable bandwidth. The old behaviour is maintained by default. Potentially, for GNU Radio 3.11 the improved_loop_filter = false case could be removed altogether, since avoiding breaks is not essential.

Yes, agreed. Feel free to add another commit, either in this PR or separately, to do that removal (none of us is keen on maintaining the broken code); it would even be fair to log a warning on maint-3.10 when $\neg$`improved_loop_filter`, as the old behaviour is so undesirable that it seems likely a user would want to be informed about it.

daniestevez:
@marcusmueller I have implemented your suggestions in the following way:
- I have amended the commit in the PR to add a log warning in the constructor when `improved_loop_filter == false`.
- I have added a new commit that removes `improved_loop_filter` and makes the FLL behave always as if `improved_loop_filter == true`.

The logic is that the two commits can be merged to the `main` branch to replace the old loop filter by the new one (potentially squashing them on merge), while the first commit can be backported to `maint-3.10` in order to maintain the old behaviour by default, but print a warning an allow the use of the new filter.

marcusmueller (approved):
both the new warning and the removal look fine

marcusmueller:
regarding squash on main: generally, yes, but in this case, I think I'll just merge as-is, and then cherry-pick the first commit to maint, as that keeps the 1:1 relationship clear

PR #7893, opened 2025-07-04
PMT: Improve pmt::list* documentation

### LINKED ISSUE #7892: pmt::list* documentation is missleading
### What happened?

The [documentation of the `pmt::list*` functions](https://www.gnuradio.org/doc/doxygen/namespacepmt.html#a8e6609e639c6eab00831cad67fa02df1) states, e.g.
> `list2()`
> Return a list of length 2 containing x1, x2. 
> `list_add()`
> Return list with item added to it. 

In the common sense and from any programming language like Python or C++ one would assume a "List" to be a sequence of items, e.g. `[1,2]`, or `[1,2,3,4]` and "adding an item to a list" to perform something like `[1,2,3]` -> `[1,2,3,4]`.

However, lists in the PMT world are different, as they represent a nested sequence of pairs such as `(1,(2,(3,NIL)))` as I just learned the hard way and can also be seen from [their implementation](https://github.com/gnuradio/gnuradio/blob/25047da6cb92cadb404e95668d19a075340fc89a/gnuradio-runtime/lib/pmt/pmt.cc#L1083):
```cpp
pmt_t list1(const pmt_t& x1) { return cons(x1, PMT_NIL); }
pmt_t list2(const pmt_t& x1, const pmt_t& x2) { return cons(x1, cons(x2, PMT_NIL)); }
pmt_t list_add(pmt_t list, const pmt_t& item){ return reverse(cons(item, reverse(list))); }
```
As an example, when serializing a list in C++ using `list3(mp("a"),mp("b"),mp("c"))`, and then deserializing it in Python using `pmt.to_python`, one ends up with `("a",("b",("c",None)))` where from the documentation one would expect a "List" in the common sense as `["a","b","c"]`.

The [API documentation](https://www.gnuradio.org/doc/doxygen/namespacepmt.html#a8e6609e639c6eab00831cad67fa02df1) about lists here is insufficient and needs to explain this concept of "List" used by PMT. The [Wiki](https://wiki.gnuradio.org/index.php?title=Polymorphic_Types_(PMTs)) also does not explain this behaviour of lists.

### System Information

OS: Yocto 

### GNU Radio Version

3.10 (maint-3.10)

### Specific Version

3.10.11.0

- Read documentation
- Functions don't work as expected from documentation
- Read source code to understand what the functions really do

### Relevant log output

```shell

```

eltos:
## Description

Improve docstrings of pmt::list* functinos, see #7892

## Related Issue

Closes #7892

Documentation of PMT functions only

## Testing Done

NA

## Checklist

### DIFF
--- a/gnuradio-runtime/include/pmt/pmt.h
+++ b/gnuradio-runtime/include/pmt/pmt.h
@@ -860,34 +860,51 @@ PMT_API pmt_t member(pmt_t obj, pmt_t list);
 PMT_API bool subsetp(pmt_t list1, pmt_t list2);
 
 /*!
- * \brief Return a list of length 1 containing \p x1
+ * \brief Return a list of length 1 containing \p x1, that is
+ * cons(x1, NIL)
+ * Note that PMT lists are nested pairs.
+ * Use pmt::make_tuple to generate a list in the common sense.
  */
 PMT_API pmt_t list1(const pmt_t& x1);
 
 /*!
- * \brief Return a list of length 2 containing \p x1, \p x2
+ * \brief Return a list of length 2 containing \p x1, \p x2, that is
+ * cons(x1, cons(x2, NIL))
+ * Note that PMT lists are nested pairs.
+ * Use pmt::make_tuple to generate a list in the common sense.
  */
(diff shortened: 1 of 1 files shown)


mbr0wn:
Formatting passed, merging.

Issue #7695, opened 2025-01-28
tcp_source (from the GRC blks2 time) still bad, still not replaced


marcusmueller:
### What happened?

A problem of a user yesterday reminded me of our tcp source still being bad (Details: https://github.com/gnuradio/gnuradio/issues/1540). When we decommissioned the blks2 hierarchy (which where Python-constructed blocks that shipped with GRC instead of any of the gr-* modules), we just moved the `tcp_sink` to gr-network. (Although we knew it was bad.)

We need to replace it with one that

- doesn't wait for a connection in the constructor, meaning that you can't even *construct*, leave alone *start* a flow graph without a connection being made (and that can, and does, lead to deadlocks)
- has unit tests for all connection directions and starting orders
- is C++
  - has GRC C++ generation
- correctly handles connection loss

desirable:

- `recv`es directly into output buffer
- allows for re-connection (if requested at constructor time)
- allows for clamping the number of bytes to be received in one call
- avoids dependendency on Boost, since this should all be synchronous anyway (sources blocking in work)  and afaict socket.h works on windows just as well (needs verification)

### System Information

n/a

### GNU Radio Version

3.11-git (main)

### Specific Version

v3.11.0.0git-892-g6406da8f68

### Steps to Reproduce the Problem

n/a

### Relevant log output

```shell

```

marcusmueller:
Tagging this bug since complete lack of QA for a block is one in itself; but also, the "needs to have a connection constructor time" is counter-intuitive and a source of application deadlock.

the-gremlin:
I will attempt to _very slowly_ work on this issue. Probably will start with a non-templated version first and then we'll see how it goes :)

PR #7895, opened 2025-07-06
gr-network: Add new tcp source implementation

the-gremlin:
## Description
This pull request attepts to address issue #7695 by providing a pure C++ implementation of the tcp source block that doesn't block in the constructor and doesn't depend on boost/asio.

gr-network

## Testing Done
No testing yet as I havent figured out how to create the python bindings.

## Checklist


the-gremlin:
This is a draft PR because I would like a second opinion on a lot of the decisions I have taken in the code and also help with creating the python bindings and the documentation.

the-gremlin:
Also this is a non-templated implementation when it probably should be, although since I _think_ that i am writing directly into the output buffer there shouldn't be any problems in that regard

gr-network/lib/tcp_source_impl.cc
```diff
@@ -0,0 +1,179 @@
...
+    : d_connection_mode(source_mode), d_block_size(item_size * vec_len), d_veclen(vec_len)
+{
+
+    struct addrinfo hints;
+    struct addrinfo* host_info;
+
+    memset(&hints, 0, sizeof(struct addrinfo));
+    hints.ai_family = AF_UNSPEC;
+    hints.ai_socktype = SOCK_STREAM;
+    hints.ai_flags = AI_PASSIVE;
```
ThomasHabets, line 95: Should `AI_PASSIVE` be used when mode is client?

gr-network/lib/tcp_source_impl.cc
```diff
@@ -0,0 +1,179 @@
...
+    // copy the host data over
+    d_host_addr_len = host_info->ai_addrlen;
+    d_host_addr = (struct sockaddr*)malloc(host_info->ai_addrlen);
+    memcpy(d_host_addr, host_info->ai_addr, host_info->ai_addrlen);
+
+    // Create the socket and also bind to a a port if we're the server
+    if (d_connection_mode == TCP_SOURCE_MODE_SERVER) {
+        d_socket =
+            socket(host_info->ai_family, host_info->ai_socktype, host_info->ai_protocol);
+        bind(d_socket, host_info->ai_addr, host_info->ai_addrlen);
```
ThomasHabets: Check for error returned by `bind()`.

gr-network/lib/tcp_source_impl.cc
```diff
@@ -0,0 +1,179 @@
...
+    } else {
+        d_socket =
+            socket(host_info->ai_family, host_info->ai_socktype, host_info->ai_protocol);
+    }
+
+    if (d_socket == -1) {
+        d_logger->error("Could not create TCP socket: {}", strerror(errno));
```
ThomasHabets: This strerror will be less useful if `bind()` failed because `socket()` failed.

gr-network/lib/tcp_source_impl.cc
```diff
@@ -0,0 +1,179 @@
...
+        throw std::runtime_error("TCP Source encountered an error.");
```
ThomasHabets: Shouldn't the exception also contain the error (also elsewhere)?

gr-network/lib/tcp_source_impl.cc
ThomasHabets: It's often useful to set `REUSEADDR` when binding, to to be stopped by a wait timeout between attempted runs.

gr-network/lib/tcp_source_impl.cc
```diff
@@ -0,0 +1,179 @@
...
+    if (d_connection_mode == TCP_SOURCE_MODE_CLIENT) {
+        int res = connect(d_socket, d_host_addr, d_host_addr_len);
+
+        if (res < 0) {
+            d_logger->error("Could not connect to specified server: {}", strerror(errno));
+        } else {
+            d_logger->info("Successfully connected to server.");
+            d_connected = true;
+            d_exchange_socket = d_socket;
+            return;
```
ThomasHabets, line 230: The return statements in this function need cleanup. This one is not needed, nor is the one at the end of the function.

gr-network/lib/tcp_source_impl.cc
```diff
@@ -0,0 +1,179 @@
...
+        }
+    } else {
+        int res = listen(d_socket, 5);
```
ThomasHabets: Why 5?

ThomasHabets: Also `listen()` belongs with `bind()`, not here, right?

gr-network/lib/tcp_source_impl.cc
```diff
@@ -0,0 +1,179 @@
...
+            socket(host_info->ai_family, host_info->ai_socktype, host_info->ai_protocol);
+    }
+
+    if (d_socket == -1) {
+        d_logger->error("Could not create TCP socket: {}", strerror(errno));
+        throw std::runtime_error("TCP Source encountered an error.");
+    }
+}
+
+void tcp_source_impl::connection_handler()
```
ThomasHabets: This function is not a connection handler. How about `establish_connection()`?

gr-network/lib/tcp_source_impl.cc
```diff
@@ -0,0 +1,179 @@
...
+        int received = recv(d_exchange_socket, out, to_be_received, 0);
+
+        // error encountered
+        if (received == -1) {
+            d_logger->error("Error encountered while attempting to receive data: {}",
+                            strerror(errno));
+
+            // notify the user in case of disconnects when we are the server
+            if (d_connection_mode == TCP_SOURCE_MODE_SERVER &&
+                (errno == ECONNRESET || errno == ETIMEDOUT)) {
```
ThomasHabets: So what should happen if we get these errors while being the client?

gr-network/lib/tcp_source_impl.h
```diff
@@ -0,0 +1,58 @@
...
+namespace gr {
+namespace network {
+
+class NETWORK_API tcp_source_impl : public tcp_source
+{
+
+protected:
+    int d_socket;
+    int d_exchange_socket;
+    struct sockaddr* d_host_addr;
```
ThomasHabets: Instead of dynamic memory, did you consider `sockaddr_storage`?

gr-network/lib/tcp_source_impl.h
```diff
@@ -0,0 +1,58 @@
...
+    size_t d_host_addr_len;
+    int d_connection_mode;
+    int d_block_size;
+    int d_veclen;
+
+    bool d_connected = false;
+
+    void connection_handler();
+    std::pair<std::string, std::string> stringify_addr(struct sockaddr_storage* a);
```
ThomasHabets: What is this? It doesn't seem to need to be a member function, it probably shouldn't take a pointer to `sockaddr_storage` (what would that even mean?), it returns… two strings(?), and is not implemented?

gr-network/include/gnuradio/network/tcp_source.h
```diff
@@ -0,0 +1,56 @@
...
+{
+public:
+    typedef std::shared_ptr<tcp_source> sptr;
+
+    /*!
+     * Build a tcp_source block.
+     */
+    static sptr make(size_t itemsize,
+                     size_t veclen,
+                     const std::string& host,
```
ThomasHabets, line 48: Does gnuradio allow `string_view`? Normally it's a better choice.

gr-network/lib/tcp_source_impl.cc
```diff
@@ -0,0 +1,179 @@
...
+
+    memset(&hints, 0, sizeof(struct addrinfo));
+    hints.ai_family = AF_UNSPEC;
+    hints.ai_socktype = SOCK_STREAM;
+    hints.ai_flags = AI_PASSIVE;
+
+    // Get IP version, port, e.t.c in a workable format
+    int res;
+    if (d_connection_mode == TCP_SOURCE_MODE_CLIENT) {
+        res = getaddrinfo(host.c_str(), port.c_str(), &hints, &host_info);
```
ThomasHabets: `host_info` is a memory leak. Needs `freeaddrinfo()`.

Because exceptions and returns further down, it's likely worth making a RAII wrapper.

gr-network/lib/tcp_source_impl.cc
```diff
@@ -0,0 +1,179 @@
...
+    } else {
+        res = getaddrinfo(NULL, port.c_str(), &hints, &host_info);
+    }
+
+    if (res != 0) {
+        d_logger->error("Error while initializing: {}", gai_strerror(res));
+        throw std::runtime_error("TCP Source encountered an error.");
```
ThomasHabets: Exception messages don't end with a period.

gr-network/lib/tcp_source_impl.cc
```diff
@@ -0,0 +1,179 @@
...
+    }
+
+    // copy the host data over
+    d_host_addr_len = host_info->ai_addrlen;
```
ThomasHabets: `host_info` is a whole list of addresses. This is relevant for server, but even more relevant for client.

You can't just take the first entry in the `getaddrinfo()` result and hope for the best.

gr-network/lib/tcp_source_impl.cc
the-gremlin: That's a very neat thing I found on the manpage. If we use `AI_PASSIVE` but then the `node` argument isn't null, its ignored by getaddrinfo and it assumes we want an address for `connect`, `sendto` and the like.

gr-network/lib/tcp_source_impl.cc
the-gremlin: If i remember correctly, both the output from the logger and the raised exceptions are shown either in console or the companion application, so I chose to not spam the same message twice. But please let me know if I'm wrong!

the-gremlin:
@ThomasHabets  thank you for the extremely thorough review!! I'll do my best to fix everything you've pointed out as soon as I can (meaning, i will hopefully have enough free time to do so)

gr-network/include/gnuradio/network/tcp_source.h
the-gremlin: I'm following the lead of the other block in the gr-network module here and they all used a const string.

gr-network/lib/tcp_source_impl.cc
the-gremlin: Excellent naming choice!

gr-network/lib/tcp_source_impl.cc
the-gremlin: You are absolutely correct about `listen`, and I'm setting the queue length to 1 since I don't really think a _reasonable_ flowgraph would have two separate tcp sinks competing for the same source.

gr-network/lib/tcp_source_impl.cc
the-gremlin: In my head we would probably silently try to reconnect while blocking, but it makes a lot of sense that we should notify the user as the client :sweat_smile:

the-gremlin:
> Also this is a non-templated implementation when it probably should be, although since I _think_ that i am writing directly into the output buffer there shouldn't be any problems in that regard

This is, in fact, false because doing pointer arithmetic on void pointers is not exactly a good idea...

gr-network/lib/tcp_source_impl.cc
```diff
@@ -0,0 +1,321 @@
...
+    }
+
+    if (failed) {
+        d_logger->error("Could not find a usable address, aborting.");
+        throw std::runtime_error(
+            "TCP Source could encountered an error while initializing socket");
+    }
+
+    d_logger->info("Successfully initialized socket.");
+    freeaddrinfo(host_info);
```
ThomasHabets: This should also be called in the exception case. Hence why I said maybe some RAII wrapper would be worth it.

the-gremlin: you're absolutely right, it completely slipped my mind when going over the reviews :sweat_smile:

the-gremlin:
Waiting on the tests right now, but as far as i can tell the non templated version should at least compile. If no problems arise then I'm going to turn it into the generic version, which hopefully should be a simple affair

gr-network/lib/tcp_source_impl.cc
```diff
@@ -0,0 +1,329 @@
...
+{
+#ifdef WINDOWS_SOCKETS
+
+    WSADATA wsaData;
+
+    int wsa_res = WSAStartup(MAKEWORD(2, 2), &wsaData);
+
+    if (wsa_res != 0) {
+        d_logger->error("WSAStartup failed with error code {}.", res);
+        throw std::runtime_error("Error while initializing Windows sockets.");
```
ThomasHabets, line 75: Here and elsewhere, no periods in exception strings.

gr-network/lib/tcp_source_impl.cc
```diff
@@ -0,0 +1,329 @@
...
+        WSACleanup();
+        throw std::runtime_error("Error while initializing Windows sockets.");
+    }
+
+#endif
+
+    struct addrinfo hints;
+    // very ugly but its either that or make an entire class for a single pointer that
+    // gets used a single time
+    std::unique_ptr<addrinfo_t, void (*)(addrinfo_t*)> host_info(new addrinfo_t,
```
ThomasHabets, line 89: [spec](https://en.cppreference.com/w/cpp/memory/unique_ptr/reset.html) says deleter is only called for non-null, so this could be `nullptr` instead of a dummy allocation.

In fact, it may be safer, because `freeaddrinfo()` could quite possibly assume not just a valid pointer, but a valid object.

gr-network/lib/tcp_source_impl.cc
```diff
@@ -0,0 +1,329 @@
...
+#else
+                d_logger->warn("Failed to initialize socket: {}", strerror(errno));
+#endif
+                failed = true;
+                continue;
+            }
+
+            // get rid of "Address already in use" error
+            int yes = 1;
+            setsockopt(d_socket, SOL_SOCKET, SO_REUSEADDR, &yes, sizeof(yes));
```
ThomasHabets, line 139: Always check for errors.

gr-network/lib/tcp_source_impl.cc
```diff
@@ -0,0 +1,329 @@
...
+
+            int r = bind(d_socket, host_info->ai_addr, host_info->ai_addrlen);
+            if (r == -1) {
+#ifdef WINDOWS_SOCKETS
+                d_logger->warn("Failed to bind socket: {}", WSAGetLastError());
+#else
+                d_logger->warn("Failed to bind socket: {}", strerror(errno));
+#endif
+                failed = true;
```
ThomasHabets, line 148: Also close `d_socket`.

Maybe it'd be best to RAII-ify `d_socket` too.

gr-network/lib/tcp_source_impl.cc
```diff
@@ -0,0 +1,329 @@
...
+#ifdef WINDOWS_SOCKETS
+                d_logger->warn("Error while attempting to listen for connections: {}.",
+                               WSAGetLastError());
+
+#else
+                d_logger->warn("Error while attempting to listen for connections: {}.",
+                               strerror(errno));
+#endif
+                failed = true;
+                continue;
```
ThomasHabets, line 165: Here too close on error.

gr-network/lib/tcp_source_impl.cc
```diff
@@ -0,0 +1,329 @@
...
+    if (res != 0) {
+#ifdef WINDOWS_SOCKETS
+        d_logger->erorr("Error while initializing: {}", WSAGetLastError());
+#else
+        d_logger->error("Error while initializing: {}", gai_strerror(res));
+#endif
+        throw std::runtime_error("TCP Source encountered an error while initializing");
+    }
+
+    bool failed = false;
```
ThomasHabets, line 116: Probably easier to instead have `bool success = false`, and only set it just before `break`?

gr-network/lib/tcp_source_impl.cc
```diff
@@ -0,0 +1,329 @@
...
+#else
+                d_logger->warn("Failed to initialize socket: {}", strerror(errno));
+#endif
+                failed = true;
+                continue;
+            }
+
+            failed = false;
+            d_host_addr_len = i->ai_addrlen;
+            memcpy(&d_host_addr, i->ai_addr, d_host_addr_len);
```
ThomasHabets, line 189: And `break`

gr-network/lib/tcp_source_impl.cc
```diff
@@ -0,0 +1,329 @@
...
+#else
+            d_logger->error("Error encountered while attempting to receive data: {}",
+                            strerror(errno));
+
+            // notify the user in case of disconnects
+            if (errno == ECONNRESET || errno == ETIMEDOUT) {
+                d_logger->warn(
+                    "Connection has been shut down, waiting for a new connection.");
+                if (d_connection_mode == TCP_SOURCE_MODE_SERVER) {
+                    close(d_exchange_socket);
```
ThomasHabets, line 296: For neatness, I'd recommend setting `d_exchange_socket` to `-1` after close. It makes accidental double-close safe.

gr-network/lib/tcp_source_impl.cc
```diff
@@ -0,0 +1,329 @@
...
+            } else {
+                d_logger->info("Client has closed the connection and no more data "
+                               "remains. Waiting for new connection.");
+#ifdef WINDOWS_SOCKETS
+                closesocket(d_exchange_socket);
+#else
+
+                close(d_exchange_socket);
+#endif
+                d_connected = false;
```
ThomasHabets, line 317: Likewise, here set both `d_socket` and `d_exchange_socket` to `-1`

gr-network/lib/tcp_source_impl.cc
```diff
@@ -0,0 +1,329 @@
...
+        // block until we succeed
+        this->establish_connection();
+    }
+
+    int* out = (int*)output_items[0];
+
+    int to_be_received = d_block_size * noutput_items;
+
+    // we have connection, good, then we receive.
+    while (to_be_received > 0) {
```
ThomasHabets, line 266: But we should return as soon as we got something, right? Don't want to block until the output buffer is full?

gr-network/lib/tcp_source_impl.h
```diff
@@ -0,0 +1,66 @@
...
+#endif
+
+namespace gr {
+namespace network {
+
+class NETWORK_API tcp_source_impl : public tcp_source
+{
+
+protected:
+    int d_socket;
```
ThomasHabets, line 37: I recommend setting the sockets to `-1`.

gr-network/lib/tcp_source_impl.cc
```diff
@@ -0,0 +1,329 @@
...
+#else
+    close(d_socket);
+#endif
+}
+
+
+void tcp_source_impl::establish_connection()
+{
+    if (d_connection_mode == TCP_SOURCE_MODE_CLIENT) {
+        int res = connect(d_socket, (struct sockaddr*)&d_host_addr, d_host_addr_len);
```
ThomasHabets, line 217: For clients you need to recreate the socket.

You may want to create a test for this, as it's clear a reconnect would never have worked.

gr-network/lib/tcp_source_impl.cc
the-gremlin: I think we do, since we have a scheduler that tells us how many samples we need to produce each call. And since we're one of the sources in the flowgraph I assume its best to follow that to the letter. The only exception for this is if the connection gets terminated, since we can't really magically produce samples out of thin air

gr-network/lib/tcp_source_impl.cc
the-gremlin: Could you explain this in simpler terms? I'm not entirely familiar with a lot of C++ specifics so this is hard for me to understand

gr-network/lib/tcp_source_impl.cc
the-gremlin: That is indeed a lot cleaner, thank you!

gr-network/lib/tcp_source_impl.cc
ThomasHabets: The scheduler tells you how many you *can* produce, the maximum the output has space for.

That's at least my understanding. Though I can't immediately find any documentation saying so, e.g. in https://wiki.gnuradio.org/index.php/BlocksCodingGuide

gr-network/lib/tcp_source_impl.cc
```diff
@@ -0,0 +1,329 @@
...
+        this->establish_connection();
+    }
+
+    int* out = (int*)output_items[0];
+
+    int to_be_received = d_block_size * noutput_items;
+
+    // we have connection, good, then we receive.
+    while (to_be_received > 0) {
+        int received = recv(d_exchange_socket, out, to_be_received, 0);
```
ThomasHabets, line 267: According to the [blocks coding guide](https://wiki.gnuradio.org/index.php/BlocksCodingGuide#Blocking_calls) blocks must be interruptible by boost threads. Is that still true? If so, then that part of the guide says what to do, here.

gr-network/lib/tcp_source_impl.cc
ThomasHabets: `new addrinfo_t` just creates a struct, it'll exist in memory but will have garbage data. E.g. `host_info->ai_next` will be a dangling pointer.

`freeaddrinfo()` may (in fact probably will) try to follow that pointer.

And while you could zero-initialize it by just adding `()` to this call, I would not count on that being valid. The only "constructor" of `addrinfo` is `getaddrinfo()`.

gr-network/lib/tcp_source_impl.cc
```diff
@@ -0,0 +1,329 @@
...
+#ifdef HAVE_CONFIG_H
+#include "config.h"
+#endif
+
+#include "tcp_source_impl.h"
+
+namespace gr {
+namespace network {
+
+typedef struct addrinfo addrinfo_t;
```
ThomasHabets, line 48: This doesn't seem needed. In C++ you can just use `addrinfo` without typing `struct`.

the-gremlin: I tried to do that but clangd kept showing errors when i passed it to functions that specified `struct sockaddr` in their signature

gr-network/lib/tcp_source_impl.cc
the-gremlin: I went off of issue #7695 where in the last point of the "desirable" features for the block it said that all the block's behaviour should be both independent of Boost and completely synchronous, specifically mentioning sources blocking in work. But I would like to request @marcusmueller clarify whether he meant that Boost threads should still be able to interrupt the work function here or if we shouldn't care about it.

gr-network/lib/tcp_source_impl.cc
the-gremlin: I think especially with TCP the question of _how much_ we should receive/produce is a bit hard, since it's essentially a continuous stream of data. In the UDP case what they do is they attempt to receive enough packets to be equal to or under the requested amount. But packets as a notion don't exist in TCP, so my interpretation is that we should always try to max out our output.

gr-network/lib/tcp_source_impl.cc
the-gremlin: Since recreating the socket, especially as client, means we have to go through the pain that is `getaddrinfo` again, would it be a good idea to put the sockets in their own wrapper class so we could abstract a lot of the code away from the work/cunstructor functions, which would even help make the general behavior of the block more understandable?

gr-network/lib/tcp_source_impl.cc
ThomasHabets: I quick glance of the UDP one it looks like it doesn't block, but adds all the packets that are there right now, and then returns.

This makes the data available to downstream blocks ASAP, which is what I mean is desirable, so that the rest of the flowgraph isn't so stop-go.

gr-network/lib/tcp_source_impl.cc
ThomasHabets: Could be.

But yes you bring up a good point. When reconnecting as a client, `getaddrinfo()` should be called again. Programs that don't pick up DNS changes except on restart are annoying.

`getaddrinfo()` blocks, so the blocking answer is relevant here.

gr-network/lib/tcp_source_impl.cc
the-gremlin: That is a very fair point, and since `recv` just passes any available data to the buffer its pretty easy to achieve. The only issue i can think of would be in a case where the data is consumed by the rest of the flowgraph faster than it is sent by the sink, but I really don't know how important that is from the rest of the flowgraph's POV.

ThomasHabets: If the rest of the flowgraph is faster, then that's not an issue. It's just going to have to wait. If the data has not arrived then it's not arrived.

But if it has arrived, it's not stop-go, and processing latency could also be lower. Concurrency could be higher too.

the-gremlin: So it's not an issue either way, good to know

ThomasHabets: Depends what you mean by "either way". A source not providing data fast enough for the rest of the graph is not an issue, if there is no more data to provide.

Stop-go, high latency, and potentially overrun, are issues.

In other words, if the rest of the flowgraph is suddenly given a lot of data, then it may take a while to clear that. Maybe in the mean time new data comes in over TCP, but the TCP Source block doesn't have room for it in its output stream.

So it stops reading from the socket. And the TCP window reduces, and eventually the TCP connection stalls, even if it's just for a micro-stall, it's not great.

the-gremlin: I agree that high-latency and stop-go are problems in the general case, but I don't think its something that we really should care about here. The fact that the receive buffer may be getting filled faster than the data can be consumed by the rest of the graph just means that there will be more time for it to re-fill after every time `recv` is called.
I don't understand how stalling would be a problem for us, since it would mean that the block wouldn't have a lot/any data to produce for a couple calls, which as far as I can tell isn't a problem for the flowgraph, it just means a bit of a delay in the processing/updating.

the-gremlin:
We haven't had a response for a couple of days now, so I'm going to work with the assumption that we should both block without any care for synchronicity and not deal with boost threads and work with that, mostly because I want to continue working on this block and prefer re-working the whole behaviour of the block if it turns out I made the wrong decision.

gr-network/lib/tcp_source_impl.cc
the-gremlin: So, if I understand it correctly instead of `new addrinfo_t` I should use `nullptr` in the constructor, no?

gr-network/lib/tcp_source_impl.cc
ThomasHabets: Alright. I'm not in the approve path for this, but I would not approve it. It's assuming that new data comes in within a reasonable time, and I don't find that to be a safe assumption.

I'll reserve the right to say "I told you so" when people report bugs about stuck data.

gr-network/lib/tcp_source_impl.cc
ThomasHabets: Yes. Strictly speaking I'm not sure if `freeaddrinfo()` is defined to be safe on `nullptr`, so the paranoid thing to do is to wrap it in your own check.

gr-network/lib/tcp_source_impl.cc
the-gremlin: That's very fair, I would like to request any ideas you have on how to mitigate these issues as you are definitely more experienced than me, both in network programming and in general. I don't want my short-sightedness to cause issues that would otherwise be very preventable as that would just mean more headaches for everyone in this project down the line :)

gr-network/lib/tcp_source_impl.cc
mbr0wn: I don't think you have to be Boost-interruptible, but you can't block indefinitely. At first glance, it looks like that's what you're doing, there's no `fcntl()` call in there that is setting the socket to `O_NONBLOCKING`, nor are you using `MSG_DONTWAIT`.

If you're doing a non-blocking `recv()`, things get more complicated because you need to first check if the socket has any data, and that doesn't have a good portable solution (Linux uses epoll ideally, on Windows I'm not sure but in UHD we use `select()` which I suspect is sub-optimal, and on MacOS X and the BSDs you probably want to use `kselect`).

mbr0wn: I also think that if you start putting in timeouts, we can maybe resolve both threads at once. I agree with @ThomasHabets that it is entirely possible that the TCP data source is stalling for whatever reason, and you don't want to stall the `work()` method indefinitely. If you have a timeout, you can always return from `work()` quickly.

gr-network/lib/tcp_source_impl.cc
```diff
@@ -0,0 +1,329 @@
...
+    if (d_connection_mode == TCP_SOURCE_MODE_CLIENT) {
+        res = getaddrinfo(host.c_str(), port.c_str(), &hints, &tmp);
+    } else {
+        res = getaddrinfo(NULL, port.c_str(), &hints, &tmp);
+    }
+    host_info.reset(tmp);
+
+    if (res != 0) {
+#ifdef WINDOWS_SOCKETS
+        d_logger->erorr("Error while initializing: {}", WSAGetLastError());
```
mbr0wn, line 109: This won't compile (typo)

mbr0wn (changes requested):

ThomasHabets:
`select()` should be portable, right? And the main problem with old skool `select()/poll()` is when you have lots of fds. Sure, it's a syscall, but so's anything except `io_uring`.

Not sure how true it is anymore, but I changed arping to use `select()` over `poll()` [back in 2010 for arping](https://github.com/ThomasHabets/arping/commit/b0a754550bb873b4fdb7049bfc394d38bfe3c72b), due to buggy behavior on MacOSX. At least on those file descriptors `select()` has been reliable.

Or are you thinking of some other suboptimality of `select()`, @mbr0wn ?

the-gremlin:
Unfortunately I won't be able to work on this PR for the forseeable future as a few circumstances in my life have changed and have a lot less time available for something so complicated. Please let me know if I should close it or leave it as draft.

mbr0wn:
> `select()` should be portable, right? And the main problem with old skool `select()/poll()` is when you have lots of fds. Sure, it's a syscall, but so's anything except `io_uring`.
> 
> Not sure how true it is anymore, but I changed arping to use `select()` over `poll()` [back in 2010 for arping](https://github.com/ThomasHabets/arping/commit/b0a754550bb873b4fdb7049bfc394d38bfe3c72b), due to buggy behavior on MacOSX. At least on those file descriptors `select()` has been reliable.
> 
> Or are you thinking of some other suboptimality of `select()`, @mbr0wn ?

I forget the exact values, but `select()` scales much worse wrt. the number of open fds than poll does, and epoll is a further improvement. I recently did some experiments (with UDP sockets) where I pitted `select()` against `poll()`, and `select()` fared much worse in CPU usage. I couldn't immediately measure the improvement of `epoll()`, but since `poll()` is already not portable, you might as well go all the way.

But yes -- `select()` is portable and that's a huge advantage. In UHD, we use `poll()` on Linux and `select()` everywhere else (but I'm not saying that's the optimal outcome!).

mbr0wn:
@the-gremlin Thanks for your work, even if you need to step away from this PR. I hope you're fine, and maybe we'll see you back here in the future. I'll close this PR (otherwise we have too many open PRs), but maybe someone will pick it up and take it over the finish line.

ThomasHabets:
@mbr0wn oh yes you're right. `select()` scales with the `fd` *value*, whereas `poll()` scales with number of `fd`s in the call.

PR #7921, opened 2025-09-22
QTGUI: Fix Number Sink display format for integer types

### LINKED ISSUE #7896: Number sink short display
### What happened?

Image

Image

When setting the QT GUI Number sink to 'short', it displays a number with decimals, this seems nonsensical as shorts are always integer numbers. I would expect this to show simply '0' instead of '0.0000...'

### System Information

OS: (Linux Distro, Mac, Windows)
GR Installation Method: (native distro packages - apt, rpm, etc., Conda, Source

### GNU Radio Version

3.10 (maint-3.10)

### Specific Version

3.10.12.0

Create any flowchart with a QT GUI Number Sink set to 'short'.

### Relevant log output

```shell

```

jsrv07:
## Description
This change fixes a display bug in the QT GUI Number Sink where integer data types (like 'short') were being formatted with unnecessary decimals (e.g., `5` was displayed as `5.000`).

This commit updates the display logic to check the item size of the input. If the type is an integer (char or short), the number is now formatted as a whole number, providing a correct representation for non-floating-point types.

### Before:
Screenshot 2025-09-22 220553

### After:
With short:
Screenshot 2025-09-22 224136

With float:
Screenshot 2025-09-22 232031

## Related Issue
Fixes #7896

This is a UI fix that affects the QT GUI Number Sink block. 
It corrects the string formatting for integer types and has no effect on runtime performance.

## Testing Done
I performed manual testing to verify the fix:
Environment: GNU Radio built from source on Windows Subsystem for Linux (WSL).
Verification Steps:
    1.  Created a GRC flowgraph with a Constant Source (value 10) connected to a QT GUI Number Sink.
    2.  Set the QT GUI Number Sink's Type to Short.
    3.  Ran the flowgraph and confirmed the display correctly showed 10 instead of 10.000.
    4.  Changed the type back to Float and confirmed it still correctly displayed decimal values.

## Checklist
- [x] I [have signed my commits before making this PR](https://github.com/gnuradio/gnuradio/blob/main/CONTRIBUTING.md#dco-signed).
- [ ] I have updated [the documentation](https://wiki.gnuradio.org/index.php/Main_Page#Documentation) where necessary. (Not applicable for this change).
- [ ] I have added tests to cover my changes, and all previous tests pass. (Manual UI testing is sufficient for this visual change).


gr-qtgui/include/gnuradio/qtgui/numberdisplayform.h
```diff
@@ -26,9 +26,11 @@ class NumberDisplayForm : public QWidget
     Q_OBJECT
 
 public:
-    NumberDisplayForm(int nplots = 1,
-                      gr::qtgui::graph_t type = gr::qtgui::NUM_GRAPH_HORIZ,
-                      QWidget* parent = 0);
+    // This is the NEW line , added d_itemsize to resolve #7896
```
marcusmueller: let's not have too many of these kinds of comments: From the git history it's clear that this line is new, and from your git commit message, it becomes clear what you fixed there

gr-qtgui/include/gnuradio/qtgui/numberdisplayform.h
```diff
@@ -26,9 +26,11 @@ class NumberDisplayForm : public QWidget
...
+    NumberDisplayForm(int nplots,
+                      gr::qtgui::graph_t type,
+                      size_t itemsize,
```
marcusmueller: could we instead pass some enum that actually describes the type? Going by itemsize, even if done elsewhere in gr-qtgui, is a bad idea, because `sizeof(float) == sizeof(int)` and `sizeof(unsigned char) == sizeof(signed char)` and `sizeof(std::complex<float>) == sizeof(double) == sizeof(int64_t)`…

gr-qtgui/lib/number_sink_impl.cc
```diff
@@ -82,7 +82,9 @@ void number_sink_impl::initialize()
         d_qApplication = new QApplication(d_argc, &d_argv);
     }
 
-    d_main_gui = new NumberDisplayForm(d_nconnections, d_type, d_parent);
+    // This is the NEW line , added d_itemsize to resolve #7896
```
marcusmueller: as said, remove such comments: git tells us this is a change.

gr-qtgui/include/gnuradio/qtgui/numberdisplayform.h
```diff
@@ -26,9 +26,11 @@ class NumberDisplayForm : public QWidget
     Q_OBJECT
 
 public:
-    NumberDisplayForm(int nplots = 1,
-                      gr::qtgui::graph_t type = gr::qtgui::NUM_GRAPH_HORIZ,
-                      QWidget* parent = 0);
+    // This is the NEW line , added d_itemsize to resolve #7896
+    NumberDisplayForm(int nplots,
+                      gr::qtgui::graph_t type,
```
marcusmueller: I'm not happy that we're removing the default here. This introduces unnecessary complexity when porting for people using this block!

gr-qtgui/lib/numberdisplayform.cc
```diff
@@ -218,8 +223,18 @@ void NumberDisplayForm::newData(const QEvent* updateEvent)
 
         for (unsigned int i = 0; i < d_nplots; ++i) {
             float f = d_factor[i] * samples[i];
-            d_text_box[i]->setText(
-                QString("%1 %2").arg(f, 4, ' ').arg(QString(d_unit[i].c_str())));
+
+            // #7896 fix
+            // Check if the data type is an integer (char or short)
```
marcusmueller: strike

marcusmueller (changes requested):

jsrv07:
Thank you for the detailed feedback, @marcusmueller. Your suggestions make a lot of sense. I will work on incorporating these changes.

marcusmueller:
Considering this still is only a very partial fix (you've addressed the number printage, but the axis ticks are still having decimals), I'd say:

Let's not do it this way! Let's approach this in a slightly easier, less breaking way.

1. add a setter to numberdisplaysink.h / .cc, which allows the user to set whether they want floating point or integer display
2. (don't change the constructor of `NumberDisplayForm`, because that's a breaking change and I couldn't release that on the 3.10 series)
3. make the `number_sink_impl` call that setter appropriately.

Idea is the following:

The itemsize approach (again, although we're already doing that elsewhere in the code) is bad, because there's so many types that have the same size. We might one day (hopefully) overhaul all of qtgui just to get rid of that. So, let's not depend more on passing itemsizes down to the Qt widgets than necessary!

Using a setter instead of changing the constructor signature helps us keep compatibility, while still allowing people to get a bugfix.

jsrv07:
Hi @marcusmueller, I’ve updated the PR according to your feedback (added the setter method, kept the constructor unchanged, and cleaned up the formatting logic). All checks are passing now. Could you please take another look when you have some time? Thanks again for the detailed guidance earlier.

marcusmueller:
This looks pretty nice!

PR #7933, opened 2025-10-06
uhd: Bump minimum UHD version to 3.15

mbr0wn:
If we do a GNU Radio 3.11, I want to limit the available UHD versions. Going to UHD 4 would have been nice for me, but I think some people will want to stay on UHD 3 for some reason and there's plenty of gr-uhd that works just fine with that.


marcusmueller:
I like this. On the machines that GR 3.11 builds, nothing older than 3.15 is likely to be convinced to be built, anyways.

PR #7937, opened 2025-10-09
RFNoC: Add PDU Support to Tx Streamer

mbr0wn:
```
commit d820cb27a4a078ca4ff22cccf5a4c9304df229ee (HEAD -> mbr0wn/rfnoc-tx-pdu, mb/mbr0wn/rfnoc-tx-pdu)
Author: Martin Braun <martin.braun@ettus.com>
Date:   Thu Oct 9 15:38:21 2025 +0200

    uhd: examples: Update rfnoc_duc_radio.grc

    This examples demonstrates the standard RFNoC Tx chain. This update
    fixes the Qt-based configuration of the gain and frequency settings, and
    adds a variable for easier channel selection. It also sets a more
    sensible default antenna and adds some comments to improve the
    educational value of the example.

commit f8e573f422182521dc57d2fea9e8065cf0b49886
Author: Martin Braun <martin.braun@ettus.com>
Date:   Wed Oct 8 14:34:59 2025 +0200

    uhd: Add PDU port to RFNoC Tx streamer

    This allows sending PDUs to the RFNoC Tx streamer instead of regular
    streams. In addition to the normal streaming port, the RFNoC Tx Streamer
    block now also has a message port that accepts PDUs. This can be used
    to implement bursty communication, radar bursts, or other things that
    are better mapped to PDUs than continuous streams.

    Messages can include a timestamp for timed transmit. By default, every
    PDU is transmitted with an end-of-burst flag, but that can be disabled
    for back-to-back PDUs.

    Another example (rfnoc_duc_radio_pdu.grc) is added to show off this
    feature with a message strobe.
```

## Description

This is how the PDU Tx example looks like:

image

If you receive the signal, it looks like this (note I also checked the signal on a scope):

image

(bursts are supposed to come 1s apart -> ✔️ )

Here is the updated regular Tx example:

image

Modifies the RFNoC Tx Streamer

## Testing Done
- Added example
- Tested the example with another USRP connected via cable, then see that I get the bursts I'm expecting
- Used the improved existing example to test that I didn't break the original feature set

## Checklist

### DIFF
--- a/gr-uhd/examples/grc/rfnoc_duc_radio.grc
+++ b/gr-uhd/examples/grc/rfnoc_duc_radio.grc
@@ -1,34 +1,30 @@
 options:
   parameters:
+    alias: ''
     author: Martin Braun <martin.braun@ettus.com>
     catch_exceptions: 'True'
-    category: '[GRC Hier Blocks]'
-    cmake_opt: ''
     comment: ''
     copyright: ''
     description: ''
-    gen_cmake: 'On'
     gen_linking: dynamic
     generate_options: qt_gui
+    generator_class_name: PythonQtGuiGenerator
+    generator_module: gnuradio.grc.workflows.python_qt_gui
     hier_block_src_path: '.:'
     id: rfnoc_duc_radio
     max_nouts: '0'
     output_language: python
-    placement: (0,0)
     qt_qss_theme: ''
     realtime_scheduling: ''
     run: 'True'
     run_command: '{python} -u {filename}'
-    run_options: prompt
-    sizing_mode: fixed
     thread_safe_setters: ''
     title: 'RFNoC: Radio -> DDC Example'
-    window_size: (1000,1000)
   states:
     bus_sink: false
     bus_source: false
     bus_structure: null
-    coordinate: [8, 8]
+    coordinate: [8, 12.0]
     rotation: 0
     state: enabled
 
@@ -37,6 +33,7 @@ blocks:
   id: variable_qtgui_entry
   parameters:
     comment: ''
+    entry_signal: editingFinished
     gui_hint: ''
     label: Frequency (Hz)
     type: real
@@ -45,13 +42,14 @@ blocks:
     bus_sink: false
     bus_source: false
     bus_structure: null
-    coordinate: [288, 80.0]
+    coordinate: [472, 12.0]
     rotation: 0
     state: true
 - name: gain
   id: variable_qtgui_entry
   parameters:
     comment: ''
+    entry_signal: editingFinished
     gui_hint: ''
     label: Gain (dB)
     type: int
@@ -60,25 +58,40 @@ blocks:
     bus_sink: false
     bus_source: false
     bus_structure: null
-    coordinate: [184, 80.0]
+    coordinate: [360, 12.0]
+    rotation: 0
+    state: true
+- name: radio_chan
+  id: variable
+  parameters:
+    comment: ''
+    value: '0'
+  states:
+    bus_sink: false
+    bus_source: false
+    bus_structure: null
+    coordinate: [760, 276.0]
     rotation: 0
     state: true
 - name: radio_rate
   id: variable
   parameters:
-    comment: ''
+    comment: 'This variable must match the master clock rate
+
+      of this device.'
     value: 200e6
   states:
     bus_sink: false
     bus_source: false
     bus_structure: null
-    coordinate: [8, 112.0]
+    coordinate: [936, 220.0]
     rotation: 0
     state: true
 - name: samp_rate
   id: variable_qtgui_entry
   parameters:
     comment: ''
+    entry_signal: editingFinished
     gui_hint: ''
     label: Sampling Rate (Hz)
     type: real
@@ -87,13 +100,14 @@ blocks:
     bus_sink: false
     bus_source: false
     bus_structure: null
-    coordinate: [416, 80.0]
+    coordinate: [608, 12.0]
     rotation: 0
     state: true
 - name: tone_freq
   id: variable_qtgui_entry
   parameters:
     comment: ''
+    entry_signal: editingFinished
     gui_hint: ''
     label: Baseband Tone Frequency (Hz)
     type: real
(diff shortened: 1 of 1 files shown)


gr-uhd/lib/rfnoc_tx_streamer_impl.cc
```diff
@@ -46,15 +46,178 @@ rfnoc_tx_streamer_impl::rfnoc_tx_streamer_impl(rfnoc_graph::sptr graph,
...
+        if (cpu_format != expected_format) {                                             \
+            d_logger->warn("PDU contains " #type_check_func                              \
+                           " data but streamer expects cpu_format '{}' not '{}'",        \
+                           cpu_format,                                                   \
+                           expected_format);                                             \
+            return;                                                                      \
+        }                                                                                \
+        const c_type* data_ptr = pmt_get_elements_func(data, data_len);                  \
+        input_items.push_back(data_ptr);                                                 \
+    } else
```
marcusmueller: is there any reason this is not a function but a macro? it really seems rather inelegant to do this in a macro instead of just `||`-chaining bool function invocations.

I'm really not fond of this; it makes debugging hard and it makes for hard to reason about code.

gr-uhd/lib/rfnoc_tx_streamer_impl.cc
```diff
@@ -46,15 +46,178 @@ rfnoc_tx_streamer_impl::rfnoc_tx_streamer_impl(rfnoc_graph::sptr graph,
...
+
+    // Parse metadata for TX metadata fields
+    ::uhd::tx_metadata_t tx_metadata;
+    tx_metadata.start_of_burst = true;
+    tx_metadata.end_of_burst = true;
+    tx_metadata.has_time_spec = false;
+
+    if (pmt::is_dict(metadata)) {
+        // Check for start of burst
+        if (pmt::dict_has_key(metadata, pmt::mp("tx_sob"))) {
```
marcusmueller: really up to your rate assessment, but do note that `pmt::mp` is something you don't want in your fastpath: it allocates new entries, hashes something slowly, might even grow a a container, hence malloc/realloc.

gr-uhd/lib/rfnoc_tx_streamer_impl.cc
```diff
@@ -46,15 +46,178 @@ rfnoc_tx_streamer_impl::rfnoc_tx_streamer_impl(rfnoc_graph::sptr graph,
...
+            }
+        }
+    }
+
+    // Calculate number of items considering vector length
+    size_t num_items_to_send = data_len / d_vlen;
+    if (data_len % d_vlen != 0) {
+        d_logger->warn(
+            "PDU data length {} is not a multiple of vector length {}", data_len, d_vlen);
+        num_items_to_send = data_len / d_vlen; // Send complete vectors only
```
marcusmueller: redundant

gr-uhd/lib/rfnoc_tx_streamer_impl.cc
```diff
@@ -46,15 +46,178 @@ rfnoc_tx_streamer_impl::rfnoc_tx_streamer_impl(rfnoc_graph::sptr graph,
...
+            size_t remaining_items = num_items_to_send;
+
+            while (remaining_items > 0) {
+                // Update input_items pointers to account for already sent data
+                std::vector<const void*> offset_input_items;
+                for (const void* ptr : input_items) {
+                    const char* byte_ptr = static_cast<const char*>(ptr);
+                    offset_input_items.push_back(byte_ptr +
+                                                 (total_items_sent * d_itemsize));
+                }
```
marcusmueller: OK, but you can definitely use `copy_if` if you'd like, or put C++20 as requirement and range/view-solve this. I'm at a point where I thinkn the time has come.

gr-uhd/lib/rfnoc_tx_streamer_impl.cc
```diff
@@ -46,15 +46,178 @@ rfnoc_tx_streamer_impl::rfnoc_tx_streamer_impl(rfnoc_graph::sptr graph,
...
+
+    if (num_items_to_send > 0) {
+        // Send data via streamer
+        try {
+            size_t total_items_sent = 0;
+            size_t remaining_items = num_items_to_send;
+
+            while (remaining_items > 0) {
+                // Update input_items pointers to account for already sent data
+                std::vector<const void*> offset_input_items;
```
marcusmueller:
```suggestion
                std::vector<const void*> offset_input_items;
                std::vector<const void*> offset_input_items.reserve(input_items.size());
```

gr-uhd/lib/rfnoc_tx_streamer_impl.h
```diff
@@ -32,6 +32,9 @@ class rfnoc_tx_streamer_impl : public rfnoc_tx_streamer
              gr_vector_const_void_star& input_items,
              gr_vector_void_star& output_items) override;
 
+    /***** Message Handlers *************************************************/
+    void handle_pdu(pmt::pmt_t msg);
```
marcusmueller:
```suggestion
    void handle_pdu(const pmt::pmt_t& msg);
```
to avoid the unnecessary refcount increase/decrease of the shared_ptr, which is mutexed

gr-uhd/lib/rfnoc_tx_streamer_impl.cc
mbr0wn: In the other USRP blocks, I'm already doing this the right way, and fixing this gave me the chance to align code further.

gr-uhd/lib/rfnoc_tx_streamer_impl.cc
mbr0wn: I replaced this with `std::transform` to keep it C++17 compatible. It has the same number of lines, though.

gr-uhd/lib/rfnoc_tx_streamer_impl.cc
mbr0wn: OK, I changed it to a lambda, but it's a lot longer now.

gr-uhd/lib/rfnoc_tx_streamer_impl.cc
mbr0wn: Actually -- I'll remove this entirely. There is not a single good reason why UHD wouldn't be able to transmit the entire burst in one go. If that happens, there's an error case and we should just report that.

Issue #7942, opened 2025-10-14
[Bug] Ubuntu 22.04 - Potential race condition in tests, random & inconsistent test failure (also causing issues in CI)


SJ-Innovation:
### What happened?

As found on #7925 CI runs, and replicated in isolation, current main branch qa_gr_hier_block2.cc causes a segfault, this also causes issues starting GRC built from current main. 

I'm investigating & replicating, will edit + update this issue with information as I find it. 

maint-3.10 - No issue, tests pass as expected
main - as of dfcfa7b61 - qa_gr_hier_block2.cc segfaults occasionally. This leads me to think its a memory allocation/ UaF somewhere. 

### System Information

Replicated on
Ubuntu 22.04 (Isolated VM), Fresh Conda, gr built from source.

### GNU Radio Version

3.11-git (main)

### Specific Version

0.0.0.0

### Steps to Reproduce the Problem

Configure for build, run unit tests qa_gr_hier_block2.cc fails. This also causes segfaults when initialising GRC (sometimes). Further details to follow. 

### Relevant log output

```shell

```

SJ-Innovation:
Righty ive been doing a bit more triage on this, having found it now a blocker for 2 of my PRs. 

Its always on Ubuntu22.04,  but doesnt seem limited to just hierblock2. 
It seems that a random (to verify) test fails intermittently.

The tests ive seen fail on a clean 22.04 VM, Miniforge3, GR built from source and ctest run with j1 --output-on-failure are:
qa_gr_hier_block2
qa_socket_pdu
qa_message_strobe
qa_message_debug
qa_message & qa_message_strobe

Putting -j32 or some other heavily parallel option seems to hide the issue, so we have a race condition of some kind... Wouldnt surprise me if its in the test framework itself. Any ideas? 

I think a possibility would be to add a --rerun-failed after the first run, because on the second run the offending test seems to pass. This could be a stopgap until the real cause is found?

PR #7947, opened 2025-10-16
gr-modtool: add grcc build time grc compilation

### LINKED ISSUE #7945: modtool: Add ability to create hier blocks with grc in OOT
### Feature Description

I've got some locally developed CMake that takes a .grc in the <module>/grc folder and runs it through grcc during the build process, allowing hier blocks to be produced within grc and distributed as part of OOT modules. 

I'll work out how to get this added into the gr-modtool template and submit a PR. 

### Feature Urgency

### More Information

_No response_

SJ-Innovation:
## Description

Adds ability for grc files placed in /grc folder to be compiled at build time by OOT modules created with gr-modtool. Handy so you can develop & maintain blocks in grc, without having to manually recompile before release.

## Related Issue

Closes #7945 

No blocks affected, only new OOTs created with gr-modtool.

## Testing Done

Test blocks created with gr-modtool, to verify functionality, also I've used this in my own OOTs for a while. 

## Checklist

- [N/A] I have added tests to cover my changes, and all previous tests pass.

### DIFF
--- a/gr-utils/modtool/templates/gr-newmod/grc/CMakeLists.txt
+++ b/gr-utils/modtool/templates/gr-newmod/grc/CMakeLists.txt
@@ -23,6 +23,12 @@ install(FILES DESTINATION share/gnuradio/grc/blocks)
 # set(test2_AUTOGRC_DEPS
 #        test3)
 
+if (WIN32)
+    set(PATH_SEP ";")
+else ()
+    set(PATH_SEP ":")
+endif ()
+
 # Enable AUTOGRC for any listed .grc files
 foreach (GRC_FILE ${AUTOGRC_FILES})
     get_filename_component(GRC_BASENAME ${GRC_FILE} NAME_WE)
@@ -34,7 +40,7 @@ foreach (GRC_FILE ${AUTOGRC_FILES})
     add_custom_command(
             OUTPUT ${GRCC_YML_OUTPUT} ${GRCC_PY_OUTPUT}
 
-            COMMAND ${CMAKE_COMMAND} -E env GRC_HIER_PATH_POST=${CMAKE_CURRENT_SOURCE_DIR}\;${CMAKE_CURRENT_BINARY_DIR} grcc -o ${CMAKE_CURRENT_BINARY_DIR} ${GRC_FILE}
+            COMMAND ${CMAKE_COMMAND} -E env GRC_HIER_PATH_POST=${CMAKE_CURRENT_SOURCE_DIR}${PATH_SEP}${CMAKE_CURRENT_BINARY_DIR} grcc -o ${CMAKE_CURRENT_BINARY_DIR} ${GRC_FILE}
 
             COMMAND ${PYTHON_EXECUTABLE} cmake_find_replace.py ${GRCC_PY_OUTPUT} "from ${GRC_BASENAME}" "from gnuradio.howto.${GRC_BASENAME}"
             COMMAND ${PYTHON_EXECUTABLE} cmake_find_replace.py ${GRCC_YML_OUTPUT} "from ${GRC_BASENAME}" "from gnuradio.howto.${GRC_BASENAME}"
--- a/grc/gui/Notebook.py
+++ b/grc/gui/Notebook.py
@@ -133,8 +133,8 @@ class Page(Gtk.HBox):
 
     def _handle_scroll_window_key_press(self, widget, event):
         is_ctrl_pg = (
-                event.state & Gdk.ModifierType.CONTROL_MASK and
-                event.keyval in (Gdk.KEY_Page_Up, Gdk.KEY_Page_Down)
+            event.state & Gdk.ModifierType.CONTROL_MASK and
+            event.keyval in (Gdk.KEY_Page_Up, Gdk.KEY_Page_Down)
         )
         if is_ctrl_pg:
             return self.get_parent().event(event)
--- a/grc/compiler.py
+++ b/grc/compiler.py
@@ -14,6 +14,7 @@ from gnuradio import gr
 
 from .core import Messages
 from .core.platform import Platform
+from .main import get_state_directory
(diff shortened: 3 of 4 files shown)


marcusmueller:
Seems to complain about missing empty lines, https://github.com/gnuradio/gnuradio/actions/runs/18569433526/job/52949184493?pr=7947#step:4:8

SJ-Innovation:
R2R: Thanks for your initial comments, I was not quite finished, should be appropriately formatted now though!

SJ-Innovation:
Withdrawn from R2R, found a limitation in dependency resolution between grcs, will improve then return to R2R status.

marcusmueller:
thinking about this, not sure, but we might have something similar in gr-digital's examples for the rx and tx flowgraphs, maybe? not sure.

SJ-Innovation:
I did have a look through the codebase before starting this, and have just rechecked gr-digital. We dont seem to have any existing method of compiling grcs as part of the build process, its all just directly installed hier block .ymls. 

The issue im sorting at the moment is when a grc depends on the build of another in its own OOT we need access to the built yml, which is currently incompatible with also accessing .block.ymls from the source directory. Options are either expanding GRC_HIER_PATH to support multiple directories (preferable) or copying the existing .block.yml's to the build directory (dont like the sound of that, hacky). Currently investigating supporting multiple GRC_HIER_PATH, doesnt seem too tricky.

SJ-Innovation:
@sdh11 Thanks for waiting, this should now be ready to review. I've added extra provisions for loading blocks in a more suitable order when compiling OOTs (ensuring the OOT is preferred over any remaining installed blocks which are superceded). I've also taken the liberty of removing an identical property getter which produces the same result in both grcc & grc modes.

SJ-Innovation:
A few more changes incoming to clean up selection between interactive grc runs and grcc runs, soon.

SJ-Innovation:
Short of any bugfixes or convention adherance, this branch is ready. Prior to merge i'm happy to squash the commits though have left seperate for the moment to provide some extra documentation.

grc/gui/Notebook.py
SJ-Innovation, line 1: I'm not a massive fan of having to import from ..main, might be worth moving get_state_directory (and related get_config_file_path) to a new file in grc/core just to keep things sanitary.

grc/gui_qt/Config.py
SJ-Innovation, line 1: The functionality removed here was identical to that in CoreConfig so was just duplicated code, we can just use the CoreConfig inherited without issue.

grc/gui_qt/components/window.py
SJ-Innovation, line 1: Only the GUI called generators should push their generated hier blocks to the user state directory, everything else should go next to the grc file. Its more sensible to make this decision here than deeper in the generator tree.

grc/workflows/common.py
SJ-Innovation, line 1: Making the decision higher up eliminates the need for this, and the associated "injection" higher up in grcc. Makes output_dir accurately reflect where the products are going all the way down the tree.

SJ-Innovation:
#7942 is preventing this PR from CI'ing. We should probably triage this as thats two of my PRs currently CI failing on this exact issue xD

gr-utils/modtool/templates/gr-newmod/grc/CMakeLists.txt
```diff
@@ -6,4 +6,71 @@
...
+
+    set(GRCC_YML_OUTPUT ${CMAKE_CURRENT_BINARY_DIR}/${GRC_BASENAME}.block.yml)
+    set(GRCC_PY_OUTPUT ${CMAKE_CURRENT_BINARY_DIR}/${GRC_BASENAME}.py)
+
+    add_custom_command(
+            OUTPUT ${GRCC_YML_OUTPUT} ${GRCC_PY_OUTPUT}
+
+            COMMAND ${CMAKE_COMMAND} -E env GRC_HIER_PATH_POST=${CMAKE_CURRENT_SOURCE_DIR}${PATH_SEP}${CMAKE_CURRENT_BINARY_DIR} grcc -o ${CMAKE_CURRENT_BINARY_DIR} ${GRC_FILE}
+
+            COMMAND ${PYTHON_EXECUTABLE} cmake_find_replace.py ${GRCC_PY_OUTPUT} "from ${GRC_BASENAME}" "from gnuradio.howto.${GRC_BASENAME}"
```
mbr0wn, line 45: CMake can do this without Python, is there a way we can use that here?

```cmake
file(READ foo.cpp TEXT)
string(REPLACE "foo" "bar" TEXT "${TEXT}")
file(WRITE foo.cpp "${TEXT}")
```

SJ-Innovation: My logic for using the helper script was that file() and string() operators run during the CMake run itself rather than during make/ninja, we need to do the replacement on the build-time products which (correct me if im wrong) CMake cant do.

mbr0wn: You would call a CMake script the same way you now call a Python script. I don't feel super strongly about this, though, and it might end up being the same amount of work. At least everyone can read Python.

SJ-Innovation: Yeah i brushed up a little on the cmake docs in response to this, its nice to see that this would be possible with pure cmake, though if you dont have too strong an opinion on it I would advocate for sticking with the Python version, as you say everyone can read it. Ty for the prompt :)

grc/gui/Notebook.py
marcusmueller: yeah we can do that separately, I think

marcusmueller (approved):

marcusmueller:
squashed, rebased, letting CI run

PR #7954, opened 2025-10-23
Add runtime-changeable min/max frequency support for DigitalNumberControl

vaibhavj2006:
This PR adds runtime-changeable minimum and maximum frequency support to the DigitalNumberControl and LabeledDigitalNumberControl classes in GNU Radio’s Qt GUI module.

Details:

Introduced new methods:

setMinFrequency(min_freq_hz) – sets the minimum allowed frequency at runtime.

setMaxFrequency(max_freq_hz) – sets the maximum allowed frequency at runtime.

setFrequencyRange(min_freq_hz, max_freq_hz) – sets both min and max frequencies simultaneously.

Updated the LabeledDigitalNumberControl to pass these methods through to the inner DigitalNumberControl.

Current frequency is automatically clamped to the new range when updated.

Enables dynamic range adjustment, making the control suitable for multi-band radios.

### DIFF
--- a/gr-qtgui/python/qtgui/digitalnumbercontrol.py
+++ b/gr-qtgui/python/qtgui/digitalnumbercontrol.py
@@ -145,6 +145,25 @@ class DigitalNumberControl(QFrame):
     def setReadOnly(self, b_read_only):
         self.read_only = b_read_only
 
+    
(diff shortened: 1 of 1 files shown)


mbr0wn:
Dupe of #7955

PR #7956, opened 2025-10-24
Add gr-linux-crypto module

Supermagnum:
A GNU Radio module that provides **Linux-specific cryptographic infrastructure integration**, focusing on what's missing from existing crypto modules (gr-openssl, gr-nacl).

### 1. **Kernel Keyring Integration**
- **Unique to Linux**: Direct integration with Linux kernel keyring
- **Secure key storage**: Keys protected by kernel, not user space
- **Key management**: Add, retrieve, link, unlink keys from kernel keyring
- **No duplication**: This is NOT available in gr-openssl or gr-nacl

### 2. **Hardware Security Module Integration**  
- **Nitrokey support**: Hardware-based key storage and operations
- **TPM integration**: Trusted Platform Module support
- **Hardware acceleration**: Use hardware crypto when available
- **No duplication**: This is NOT available in existing modules

### 3. **Kernel Crypto API Integration**
- **AF_ALG sockets**: Direct use of Linux kernel crypto subsystem
- **Hardware acceleration**: CPU crypto instructions via kernel
- **Performance**: Bypass user-space crypto libraries when possible
- **No duplication**: This is NOT available in existing modules

### **Basic OpenSSL Operations**
- **Use gr-openssl instead**: Symmetric encryption, hashing, HMAC
- **Don't duplicate**: AES, SHA, RSA operations are already in gr-openssl
- **Integration only**: Provide kernel keyring as key source for gr-openssl

### **Modern Crypto (NaCl/libsodium)**
- **Use gr-nacl instead**: Curve25519, Ed25519, ChaCha20-Poly1305
- **Don't duplicate**: Public-key crypto, authenticated encryption
- **Integration only**: Provide hardware key storage for gr-nacl
It compiles, and it has been fuzzed.

Fuzzing Results
    18.4+ billion test executions across all components
    469 total edges covered with 100% stability
    Zero security vulnerabilities found
    

marcusmueller:
Hi Supermagnum,
I'm sorry, we're not currently accepting new submodules into the tree, and especially not without discussion on the mailing list or chat, or non-essential OS-specific functionality!

This seems all interesting functionality. Please publish it as an out-of-tree module; I couldn't maintain it at a sufficient quality and cryptographic insight. I feel honored that you want to upload so much code into the project's and in this specific case, my supervision, but it's not something I'm able to, with a good conscience, can offer to maintain. Also, quite honestly, I don't see a downside to keeping it separate. If anything, it keeps maintenance easier.

There seems to be a lot of new infrastructure in here, namely for fuzzing. Is there parts of that are more general-purpose infrastructure which might be desirable to merge upstream?

Supermagnum:
The fuzzing parts is not needed for the functionality of the code. I left it in place because I could not find any guidelines on testing of the code before any merge requests are created.

marcusmueller:
Keeping your test infrastructure in the code is good. That's why I was asking – if you tested your fuzzing methodology well (and not just let some AI "hey add some fuzzing") then that'd be interesting testing infrastructure for the whole of GNU Radio and could considered for separate upstream inclusion. However, having browsed your code so far, it doesn't meet our quality standards, and I think I'll have to very politely retract my request about your fuzzing integration.

PR #7979, opened 2025-12-03
network: fix udp_sink for variable veclen

### LINKED ISSUE #7007: UDP sink block works weird
### What happened?

My flowgraph add UDP-SINK to send spectrum data as vector type，but UDP sending always stops abnormally quickly.
This is true for several versions of GNURadio，and I wonder if the UDP-SINK block is faulty or if I'm using it the wrong way？

### System Information

OS: Windows 10
GR Installation Method: radioconda installer

### GNU Radio Version

3.10 (maint-3.10)

### Specific Version

3.10.8.0 (Python 3.11.6)

1、My gnuradio by radioconda on windows version：
image
2、My flowgraph to get and send spectrum data as vector type over UDP-SINK block：
image
image
3、In fact, after the flowgraph runs, only a few packets of UDP data can be sent, and then the UDP packets stop sending：
image
4、I tried lowering the sample rate or lowering the FFT size etc, but it still didn't work.
5、I noticed the "NOTE" for "UDP-SINK", but how to add the following lines to "/etc/sysctl.conf" on the windows version?
image

Can anyone fix it? Ask for advice! Thanks a lot!

### Relevant log output

```shell
[INFO] [UHD] Win32; Microsoft Visual C++ version 14.2; Boost_108200; UHD_4.5.0.0-release
[INFO] [B200] Detected Device: B205mini
[INFO] [B200] Operating over USB 3.
[INFO] [B200] Initialize CODEC control...
[INFO] [B200] Initialize Radio control...
[INFO] [B200] Performing register loopback test... 
[INFO] [B200] Register loopback test passed
[INFO] [B200] Setting master clock rate selection to 'automatic'.
[INFO] [B200] Asking for clock rate 16.000000 MHz... 
[INFO] [B200] Actually got clock rate 16.000000 MHz.
[INFO] [B200] Asking for clock rate 32.000000 MHz... 
[INFO] [B200] Actually got clock rate 32.000000 MHz.
[INFO] [MULTI_USRP]     1) catch time transition at pps edge
[INFO] [MULTI_USRP]     2) set times next pps (synchronously)
```

ffu:
## Description

Previously, the UDP Sink block would freeze when using a vector length larger than 1. This change fixes the issue.

## Related Issue

Fixes #7007 

UDP Sink block

## Testing Done

Added QA tests for udp sink with different vector lengths. The previous code would fail, but now with the changes it passes. Also tested with the simplified flowgraph in issue #7007 to confirm system testing. For regression testing, performed make test and verified all other tests passed.

## Checklist

- [X] I have updated [the documentation](https://wiki.gnuradio.org/index.php/Main_Page#Documentation) where necessary.

### DIFF
--- a/gr-network/lib/udp_sink_impl.cc
+++ b/gr-network/lib/udp_sink_impl.cc
@@ -93,7 +93,7 @@ udp_sink_impl::udp_sink_impl(size_t itemsize,
     d_block_size = d_itemsize * d_veclen;
 
     d_precomp_datasize = d_payloadsize - d_header_size;
-    d_precomp_data_overitemsize = d_precomp_datasize / d_itemsize;
+    d_precomp_blocksize = d_precomp_datasize / d_block_size;
 
     int out_multiple = (d_payloadsize - d_header_size) / d_block_size;
 
@@ -257,9 +257,9 @@ int udp_sink_impl::work(int noutput_items,
         d_localqueue_reader->update_read_pointer(d_precomp_datasize);
     }
 
-    int itemsreturned = blocks_available * d_precomp_data_overitemsize;
+    int items_returned = blocks_available * d_precomp_blocksize;
 
-    return itemsreturned;
+    return items_returned;
 }
 
 } /* namespace network */
--- a/gr-network/lib/udp_sink_impl.h
+++ b/gr-network/lib/udp_sink_impl.h
@@ -38,7 +38,7 @@ protected:
     bool b_send_eof;
 
     int d_precomp_datasize;
-    int d_precomp_data_overitemsize;
+    int d_precomp_blocksize;
 
     char d_tmpheaderbuff[12]; // Largest header is 10 bytes
 
--- a/gr-network/python/network/qa_udp_sink.py
+++ b/gr-network/python/network/qa_udp_sink.py
@@ -9,18 +9,43 @@
 
 from gnuradio import gr, gr_unittest, blocks, network
 import time
+import socket
+import threading
 
 
-class qa_udp_sink (gr_unittest.TestCase):
+class qa_udp_sink(gr_unittest.TestCase):
     def setUp(self):
         self.tb = gr.top_block()
+        self.thread_stopped = False
+        self.receive_ready = False
 
     def tearDown(self):
         self.tb = None
 
+    def run_socket_receiver_thread(self, skt):
+        self.packets = []
+        while not self.thread_stopped:
+            try:
+                pkt, _ = skt.recvfrom(9000)
+                # Keep receiving packets until timeout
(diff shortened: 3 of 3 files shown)


snydernlyle:
Being new here I'm not certain if this expected but it does look like there a couple of failed checks. Is that expected?

ffu:
Some of the checks I would consider show-stoppers: DCO, formatting, and check hashes. If any of those fail, you should tell the PR writer to correct them. For the various OS and conda package tests, they will fail on occasion, and that's OK as long as there aren't too many of them. 

There are enough test passes here so that you should continue reviewing the code for this PR. Your feedback is welcome.

gr-network/lib/udp_sink_impl.cc
```diff
@@ -257,7 +257,7 @@ int udp_sink_impl::work(int noutput_items,
         d_localqueue_reader->update_read_pointer(d_precomp_datasize);
     }
 
-    int itemsreturned = blocks_available * d_precomp_data_overitemsize;
+    int itemsreturned = blocks_available * d_precomp_blocksize;
```
snydernlyle: I know this is ticky tacky and you didn't name this variable however should itemsreturned not be items_returned?

snydernlyle:
image

I am getting a failure on the test udp_sink test. I've include a screen shot of the verbose output. All tests are passing other than just this one. Obviously this is the important one for this PR though.

gr-network/lib/udp_sink_impl.cc
ffu: For PR's, there is a high priority to make the diffs as small as possible. Therefore, I kept the clunky variable name itemsreturned. If this was the only issue, I might push back because I've gone so far in the process, and argue the name change is not worth it. However, there is the more pressing QA failure to deal with, so I don't have a problem doing as you suggested.

ffu:
OK, I did not expect the qa test to fail as it's been passing in my environment without problems. Is there any way you can help me out with the testing? My humble request is that on qa_udp_sink.py, line 68, you change it from "rxthread.start()" to "rxthread.start(); time.sleep(5)" and see if that fixes the test. My guess is that this is an intermittent timing issue that will be difficult for me to reproduce. If this change fixes the issue, then I'll have to find a more elegant way of passing the test. Obviously I don't want to randomly sleep for a long time.

snydernlyle:
@ffu yeah the delay did the trick. Also my slower laptop never had an issue which just reaffirms things.

ffu:
Ok  thanks. I'm fairly convinced it's a race condition. I'll update the qa code and also your suggestion about variable naming.

asdaraujo:
What's the status of this PR?

PR #7981, opened 2025-12-09
blocks/typecast: create typecasting blocks to and from byte-streams

### LINKED ISSUE #7980: Advanced Block: Raw (C-Style) Casting
### Feature Description

Hi everyone, 

I think a useful set of in-tree blocks to include would be able to take a given item type and cast it directly (ie, C-style) to another. 

This would be useful for people that want to directly encode (say) ints or floats as a stream of bytes which can then be directly sent over a link. We currently have no way of encoding non-byte data into (say) a PDU. (Please educate me if I'm wrong about this!) 

We already have something similar with unpacking bytes into bit/symbol streams using un/pack bits blocks. I'm proposing something similar as un/pack bytes. Endianness will be fun. 

### Feature Urgency

### More Information

I'll write an OOT to workshop the idea. If anyones got any suggestions let me hear them! :)

SJ-Innovation:
Adds blocks to c-style typecast between compound types and byte streams, useful for packing up baseband data into streams / PDUs without having to save & reload from files. 

Closes #7980 

- [N/A ] I have updated [the documentation](https://wiki.gnuradio.org/index.php/Main_Page#Documentation) where necessary.
- [N/A] I have added tests to cover my changes, and all previous tests pass.


SJ-Innovation:
Fixing those errors & adding tests

ffu:
Please see my comment in issue #7980. Can you provide a specific use case where this block would be needed? Also see conversation in PR #7809.

SJ-Innovation:
Switched back to draft while we discuss in #7980 & given that #7809 exists (and looks better than mine!).

SJ-Innovation:
Closing due to better conversation / potential solution found in #7980.

PR #7985, opened 2025-12-14
blocks: Add numeric types to/from byte stream blocks

### LINKED ISSUE #7980: Advanced Block: Raw (C-Style) Casting

SJ-Innovation:
## Description

Adds numeric type to raw byte stream functionality, 
Uses stream->vector & vector->stream blocks internally so just a couple of yml files, as discussed in #7980.

## Related Issue

Closes #7980 

## Testing Done

As this is just yml instanciations, coverage is provided by existing stream->vector & vector->stream block qa. 

## Checklist

- [Covered by existing] I have added tests to cover my changes, and all previous tests pass.

### DIFF
--- a/gr-blocks/grc/blocks_bytes_to_numeric.block.yml
+++ b/gr-blocks/grc/blocks_bytes_to_numeric.block.yml
@@ -1,0 +1,34 @@
+id: blocks_bytes_to_numeric
+label: Bytes to Numeric Type
+flags: [ python, cpp ]
+
+parameters:
+-   id: otype
+    label: Output Type
+    dtype: enum
+    options: [complex, float, int, short, byte]
+    option_attributes:
+        size: [gr.sizeof_gr_complex, gr.sizeof_float, gr.sizeof_int, gr.sizeof_short,
+            gr.sizeof_char]
+    hide: part
+
+inputs:
+-   domain: stream
+    dtype: byte
+
+outputs:
+-   domain: stream
+    dtype: ${ otype }
+
+templates:
+    imports: from gnuradio import blocks
+    make: blocks.stream_to_vector(1, ${otype.size})
+
+cpp_templates:
+    includes: ['#include <gnuradio/blocks/stream_to_vector.h>']
+    declarations: 'blocks::stream_to_vector::sptr ${id};'
+    make: 'this->${id} = blocks::stream_to_vector::make(1, ${otype.size});'
+
+documentation: This block converts a raw byte stream to a numeric type (such as a complex, float, etc). The input byte stream must be ordered as if the block was directly written to memory, with this block acting as a C-style cast from uint8_t to the output type. No scaling or modification is performed on the stream before casting. System-native endianness is used, mostly this will be little-endian.
+
+file_format: 1
--- a/gr-blocks/grc/blocks_numeric_to_bytes.block.yml
+++ b/gr-blocks/grc/blocks_numeric_to_bytes.block.yml
@@ -1,0 +1,34 @@
+id: blocks_numeric_to_bytes
+label: Numeric Type to Bytes
+flags: [ python, cpp ]
+
+parameters:
+-   id: itype
+    label: Input Type
+    dtype: enum
+    options: [complex, float, int, short, byte]
+    option_attributes:
+        size: [gr.sizeof_gr_complex, gr.sizeof_float, gr.sizeof_int, gr.sizeof_short,
+            gr.sizeof_char]
+    hide: part
+
+inputs:
+-   domain: stream
+    dtype: ${ itype }
+
+outputs:
+-   domain: stream
+    dtype: byte
+
+templates:
+    imports: from gnuradio import blocks
+    make: blocks.vector_to_stream(1, ${itype.size})
+
+cpp_templates:
+    includes: ['#include <gnuradio/blocks/vector_to_stream.h>']
+    declarations: 'blocks::vector_to_stream::sptr ${id};'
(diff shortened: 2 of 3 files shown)


gr-blocks/grc/blocks_bytes_to_numeric.block.yml
```diff
@@ -0,0 +1,32 @@
...
+
+templates:
+    imports: from gnuradio import blocks
+    make: blocks.stream_to_vector(1, ${otype.size})
+
+cpp_templates:
+    includes: ['#include <gnuradio/blocks/stream_to_vector.h>']
+    declarations: 'blocks::stream_to_vector::sptr ${id};'
+    make: 'this->${id} = blocks::stream_to_vector::make(1, ${otype.size});'
+
```
ffu, line 31: Add documentation section since there are no python/c++ files to scrape. For reference see xmlrpc_client.block.yml

gr-blocks/grc/blocks_numeric_to_bytes.block.yml
```diff
@@ -0,0 +1,32 @@
...
+    dtype: enum
+    options: [complex, float, int, short, byte]
+    option_attributes:
+        size: [gr.sizeof_gr_complex, gr.sizeof_float, gr.sizeof_int, gr.sizeof_short,
+            gr.sizeof_char]
+    hide: part
+
+inputs:
+-   domain: stream
+    dtype: ${ type }
```
ffu: This should be itype instead of type

gr-blocks/grc/blocks_numeric_to_bytes.block.yml
```diff
@@ -0,0 +1,32 @@
...
+
+templates:
+    imports: from gnuradio import blocks
+    make: blocks.vector_to_stream(1, ${itype.size})
+
+cpp_templates:
+    includes: ['#include <gnuradio/blocks/vector_to_stream.h>']
+    declarations: 'blocks::vector_to_stream::sptr ${id};'
+    make: 'this->${id} = blocks::vector_to_stream::make(1, ${itype.size});'
+
```
ffu, line 31: Add documentation

ffu:
@SJ-Innovation: As I reviewed the changes, I also noticed these blocks were not in any categories. Please also update blocks.tree.yml so these blocks can be in the "Type Converters" category. I'm open to other categories, but that seems most appropriate.

@marcusmueller: please chime in if you can find a better name than "numeric type to bytes". I find it clumsy, but I can't think of anything that much better. Perhaps "cast to bytes" and "cast from bytes". Or maybe "cast type to bytes" and "cast bytes to type". I like the term "cast" because it implies a raw conversion, but that's just me.

gr-blocks/grc/blocks_numeric_to_bytes.block.yml
SJ-Innovation: Docstring added

gr-blocks/grc/blocks_numeric_to_bytes.block.yml
SJ-Innovation: fixed

gr-blocks/grc/blocks_bytes_to_numeric.block.yml
SJ-Innovation: Docstring added

SJ-Innovation:
@ffu Thanks for your review, all comments have been actioned. I also agree that the naming isnt great, though can't think of a better alternative. Being in the "Type Converters" section I'd say should give sufficient context, no?

ffu:
I did a visual inspection of your latest code, and it looks good. Please squash your two commits into one. Search the web for how to do this. You know you are on the right track if you do the command starting with "git rebase -i", followed by a "git push -f" into your forked repo. Once I see the squash, I'll do some manual testing and verification of your changes. Thanks.

gr-blocks/grc/blocks_bytes_to_numeric.block.yml
ffu: Update: no problem with this issue.

Question: did you confirm the documentation when you opened gnuradio companion, created the cast block, double-click on the block, and check the documentation tab? I'm no YAML expert, but I believe most multi-line entries need a "|-" or something similar. I'll also double-check  this when I get the chance.

SJ-Innovation:
Done.

ffu:
Thanks for the squash. I confirmed I could compile and install OK. I ran GRC and confirmed the blocks works. I also checked the block's GRC documentation tab and it was good

About the only remaining item left is the block documentation on the wiki. I'm not sure how the "Visit Documentation Page" hyperlink gets connected to the wiki in general, nor about how exactly to create a new page on the wiki for this block. Feedback welcome.

SJ-Innovation:
Looks like it just uses the block identifier formatted into the URL. Not sure of the order you guys like to do things, do you prefer adding the wiki page _after_ a new block is merged or before ? I'm happy to do either, just let me know!

